PR #285, opened 2026-02-18
feat: add Cursor CLI (agent) support

covlllp:
## Description

Adds Cursor CLI as a supported agent using the new centralized `AgentDef` registry pattern from #282.

Cursor CLI is built on Claude Code's codebase, so status detection delegates to the existing Claude detector. The binary name is `agent`, but it's intentionally NOT added as an alias in `resolve_tool_name` since it's too generic -- users should use `-c cursor`.

One `AgentDef` entry, one status detection function, documentation updates across 6 docs files.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Code (Claude Opus 4.6)

**Any Additional AI Details you'd like to share:** Implementation planned and executed via Claude Code CLI.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -108,12 +108,6 @@ pub struct Instance {
     pub last_start_time: Option<std::time::Instant>,
     #[serde(skip)]
     pub last_error: Option<String>,
-
-    // Search optimization: pre-computed lowercase strings (not serialized)
-    #[serde(skip)]
-    pub title_lower: String,
-    #[serde(skip)]
-    pub project_path_lower: String,
 }
 
 impl Instance {
@@ -136,16 +130,7 @@ impl Instance {
             last_error_check: None,
             last_start_time: None,
             last_error: None,
-            title_lower: title.to_lowercase(),
-            project_path_lower: project_path.to_lowercase(),
         }
-    }
-
-    /// Update the pre-computed lowercase fields for search optimization.
-    /// Call this after loading instances from disk or modifying title/path.
-    pub fn update_search_cache(&mut self) {
-        self.title_lower = self.title.to_lowercase();
-        self.project_path_lower = self.project_path.to_lowercase();
     }
 
     pub fn is_sub_session(&self) -> bool {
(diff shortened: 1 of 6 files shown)


njbrake:
Hi @covlllp next review from Claude :) 


```
  1. --append-system-prompt may not exist in Cursor CLI (Potential Bug)

  The instruction_flag is set to Some("--append-system-prompt {}"), matching Claude Code's flag. However, the https://cursor.com/docs/cli/reference/parameters does not document this flag. If Cursor CLI doesn't
  support it, instruction injection will fail silently or error when AoE tries to pass custom instructions. This should be verified -- if the flag doesn't exist, set to None.

  2. Minimal AgentConfigMount compared to Claude's (Sandbox Gaps)

  The .cursor config mount at src/session/container_config.rs:103-112 has all optional fields empty:

  AgentConfigMount {
      host_rel: ".cursor",
      container_suffix: ".cursor",
      skip_entries: &["sandbox"],
      seed_files: &[],          // Claude has onboarding seed
      copy_dirs: &[],           // Claude copies plugins/ and skills/
      keychain_credential: None, // Claude extracts OAuth from Keychain
      home_seed_files: &[],     // Claude seeds .claude.json
      preserve_files: &[],      // Claude preserves .credentials.json, history.jsonl
  }

  This means:
  - macOS: Keychain-stored OAuth tokens won't be extracted into the sandbox, so users will need to re-authenticate inside each container (or use API keys via environment variables).
  - First-run wizard: If Cursor has an onboarding flow, it will appear in every new sandbox since there's no seed file to skip it.
  - Credential overwrite: If Cursor stores credentials at predictable paths, they could be overwritten on sandbox resync since preserve_files is empty.

  This is acceptable as an initial implementation if these details aren't yet known for Cursor CLI, but should be documented or tracked as a follow-up.

  Suggestions (Non-blocking)

  1. Comment on detection reliability: The RunWithArg("agent", "--version") check is better than Which, but agent is still a very generic binary name. A comment noting that this could match unrelated binaries and
  that users should use -c cursor explicitly would be helpful.
  2. Comment on delegation rationale: detect_cursor_status is a one-liner delegating to detect_claude_status. A brief doc comment explaining why (Cursor CLI produces Claude-like terminal output) would help future
  maintainers understand whether this is a temporary shortcut or an intentional design choice.
  3. TUI error message: The startup error lists only claude, opencode, and cursor as required tools. The other agents (vibe, codex, gemini) are also omitted. This is pre-existing but worth noting for consistency.

  Verdict

  Needs Discussion

  The PR is well-structured, follows conventions correctly, and all previous review feedback has been addressed. The two remaining concerns are:

  1. instruction_flag: If --append-system-prompt doesn't exist in Cursor CLI, this should be None to avoid runtime errors. This needs verification against the actual binary.
  2. Sandbox config completeness: The empty keychain_credential/preserve_files/seed_files means sandbox support on macOS may require manual re-auth. This is fine if intentional and documented, but should be a
  conscious decision.

  Neither issue blocks the core functionality of adding Cursor as a supported agent for non-sandboxed use. If sandbox support is intended to be fully functional at launch, issue #2 should be addressed. If this is
  a "best-effort first pass" with sandbox refinements to follow, the PR is close to ready -- just needs the instruction_flag verified.
```

Have you personally run this with the cursor CLI and verified that it works ok?

PR #411, opened 2026-03-09
feat: add pi.dev coding agent support

nirok80:
## Description

Add [Pi](https://pi.dev) as a supported coding agent with full integration: agent definition, status detection, sandbox container config mounting, and AlwaysYolo mode.

Pi auto-approves all tool use (no approval gates), so a new `YoloMode::AlwaysYolo` variant is introduced and the YOLO checkbox is hidden in the new session dialog when Pi is selected.

Also improves agent detection and container command launching by falling back to a login shell, so version-manager PATHs (NVM, etc.) are available.

### Changes
- Add Pi `AgentDef` with `AlwaysYolo` mode and `PI_CODING_AGENT_DIR` container env
- Add `YoloMode::AlwaysYolo` variant for agents that always auto-approve
- Add `detect_pi_status()` with spinner, activity indicator, and prompt detection
- Add `AgentConfigMount` for `~/.pi` sandbox credential sharing
- Hide YOLO checkbox in new session dialog for `AlwaysYolo` agents
- Use login shell (`-l`) in `wrap_command_ignore_suspend()` for container commands
- Fall back to login shell in `is_agent_available()` for detection

## PR Type

- [x] New Feature

## Checklist

## AI Usage

**AI Model/Tool used:** Claude Code (Claude Opus 4.6)


src/tmux/mod.rs
```diff
@@ -97,12 +97,34 @@ pub fn is_tmux_available() -> bool {
...
+        }
+        DetectionMethod::RunWithArg(binary, arg) => {
+            if Command::new(binary).arg(arg).output().is_ok() {
+                return true;
+            }
+            let shell = std::env::var("SHELL").unwrap_or_else(|_| "bash".to_string());
+            Command::new(&shell)
+                .args(["-lc", &format!("{} {}", binary, arg)])
+                .output()
+                .is_ok()
```
jerome-benoit: **Bug (medium severity)**: This `.is_ok()` fallback always returns `true` when the login shell binary exists (which it always does). `.output()` returns `Ok(Output)` as long as bash/zsh spawns -- it doesn't check the inner command's exit code. So if `vibe` isn't in the login shell's PATH either, this still reports it as available.

Compare with the `Which` branch above at L115, which correctly uses `.map(|o| o.status.success()).unwrap_or(false)`.

Currently only Vibe uses `RunWithArg` (agents.rs L94), so only Vibe is affected. Vibe would appear in the tool picker and fail at launch.

Fix:
```rust
Command::new(&shell)
    .args(["-lc", &format!("{} {}", binary, arg)])
    .output()
    .map(|o| o.status.success())
    .unwrap_or(false)
```

PR #381, opened 2026-03-06
fix: update Claude Code spinner detection to match current UI

KarlRombauts:
## Description

Claude Code no longer uses braille spinners (⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏). It now uses spinner chars (· ✢ ✳ ✶ ✻ ✽ ●) with the pattern `<spinner> <verb>…` when active, e.g. `✳ Twisting… (10m 6s · ↓ 18.3k tokens · thinking)`.

This PR adds a regex-based detection for the new spinner pattern, distinguishing active spinners (which contain `…`) from completion messages like `✻ Worked for 1m 52s` (which don't). Braille spinner detection is kept as a fallback.

Closes #378

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] I understand the code I am submitting
- [x] New and existing tests pass
- [ ] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

## AI Usage

- [x] AI was used for drafting/refactoring

**AI Model/Tool used:** Claude Code (Opus 4.6)

**Any Additional AI Details you'd like to share:** Used Claude Code to implement the fix and write tests. I reviewed and tested the changes locally.

PR #414, opened 2026-03-09
fix: implement tmux pane fallback for Claude Code status detection

hansonkim:
## Description

The hook-based status detection added in #390 replaced string-based detection but left `detect_claude_status()` as a stub that always returns `Idle`. When the hook status file becomes stale (e.g. during long-running MCP tool calls exceeding the 5-minute staleness threshold), sessions incorrectly show as Idle instead of Running.

This implements the tmux pane content fallback that was specified as requirement #1 in the #381 discussion:

> 1. Fallback to regular tmux string detection if the hook isn't installed

**Detection approach** (proposed in #381 -- spinner char + ellipsis, no regex):

| Signal | Status | Example |
|---|---|---|
| Active spinner char + `…` | Running | `✳ Catapulting… (12m 33s)` |
| Braille spinner | Running | `Processing ⠋` (older versions) |
| Completion spinner (no `…`) | Idle | `✻ Worked for 1m 52s` |
| Approval prompts | Waiting | `(y/n)`, `[y/n]`, `approve` |
| Input prompt | Waiting | `>` |

Scan is limited to the last ~10 non-empty lines to reduce false positives from scroll history.

**How I tested:** `cargo test`, `cargo clippy`, `cargo fmt`. Verified against a live session running a long MCP call that exceeded the staleness threshold.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

**AI Model/Tool used:** Claude Opus 4.6 (Claude Code)

**Any Additional AI Details you'd like to share:** Implementation was pair-programmed with Claude Code. Design decisions were made by the contributor based on the approach proposed in #381.

### DIFF
--- a/src/tmux/status_detection.rs
+++ b/src/tmux/status_detection.rs
@@ -6,7 +6,8 @@ use super::utils::strip_ansi;
 
 const SPINNER_CHARS: &[&str] = &["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
 
-/// Claude Code's active spinner characters (· ✢ ✳ ✶ ✻ ✽ ●).
+/// Claude Code's active spinner characters, observed as of Claude Code 1.0.33 (2025-06).
+/// See also PR #381 for history of spinner changes.
 const ACTIVE_SPINNER_CHARS: &[char] = &['·', '✢', '✳', '✶', '✻', '✽', '●'];
 
 /// Check if a line is an active Claude Code spinner line.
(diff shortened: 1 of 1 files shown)


src/tmux/status_detection.rs
```diff
@@ -23,10 +37,69 @@ pub fn detect_status_from_content(content: &str, tool: &str, _fg_pid: Option<u32
...
+    }
+
+    // RUNNING: Braille spinners (older Claude Code versions)
+    for line in &recent_lines {
+        for spinner in SPINNER_CHARS {
+            if line.contains(spinner) {
+                return Status::Running;
+            }
+        }
+    }
```
jerome-benoit: **Suggestion (informational)**: Claude Code displays `"esc to interrupt"` during active processing -- the comment at L52 in this file even says `"OpenCode shows 'esc to interrupt' when busy (same as Claude Code)"`. All three other active detectors (OpenCode L54, Codex L269, Gemini L363) check for it as a Running signal.

Adding it here would provide a redundant Running indicator that:
- Is positioned at the bottom of the TUI (reliably within any scan window)
- Doesn't depend on correctly identifying spinner characters
- Would still work if Claude Code changes its spinner set in the future

Something like:
```rust
if last_lines_lower.contains("esc to interrupt") {
    return Status::Running;
}
```
before the active spinner check.

src/tmux/status_detection.rs
```diff
@@ -6,6 +6,20 @@ use super::utils::strip_ansi;
 
 const SPINNER_CHARS: &[&str] = &["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
 
+/// Claude Code's active spinner characters (· ✢ ✳ ✶ ✻ ✽ ●).
+const ACTIVE_SPINNER_CHARS: &[char] = &['·', '✢', '✳', '✶', '✻', '✽', '●'];
+
```
jerome-benoit: **Nit**: A brief comment noting where this character set was sourced from (e.g., Claude Code version or date observed) would help future maintainers know when to re-verify. Claude Code has changed its spinner set before (the braille spinners at L8 are from older versions).

jerome-benoit (commented):
No bugs found -- the core logic (`is_active_spinner_line` distinguishing active spinners from completion lines via the `…` ellipsis) is correct and well-tested.

5 suggestions below, all low severity or informational. The two most actionable:
1. **Missing `starts_with("> ")` for partial input** (L98) -- every other detector has this pattern for catching sessions where the user started typing at the prompt
2. **Missing `"esc to interrupt"` Running check** (L70) -- the comment at L52 in this file notes Claude Code uses the same signal as OpenCode, and all 3 other active detectors check for it

PR #426, opened 2026-03-10
fix: use $SHELL instead of hardcoded bash for agent launch and hook execution

### LINKED ISSUE #425: Shell is hardcoded to `bash` instead of respecting `$SHELL`
Several places hardcode `bash` as the shell instead of using the user's `$SHELL`. This means version-manager PATHs (NVM, mise, asdf, etc.) configured in `~/.zshrc` or `~/.config/fish/config.fish` won't be available when AoE launches agents or runs hooks.

### Affected locations

| File | Line | Usage | Impact |
|---|---|---|---|
| `src/session/instance.rs` | L707 | `wrap_command_ignore_suspend` -- `bash -c` (PR #411 changes to `bash -lc`) | Agent launch ignores user shell; login-shell sourcing targets wrong rc files |
| `src/session/repo_config.rs` | L415 | Local hook execution -- `bash -c` | Hooks don't get user's PATH/env |

### Not affected

- `instance.rs:170` -- `"bash"` as last-resort fallback when agent is unknown (reasonable default)
- `repo_config.rs:430` -- `bash -c` inside Docker containers (container env, not user shell)
- `hooks/mod.rs:345` -- `sh -c` in test code only

### Inconsistency introduced by PR #411

`is_agent_available` (tmux/mod.rs) correctly reads `$SHELL`:
```rust
let shell = std::env::var("SHELL").unwrap_or_else(|_| "bash".to_string());
Command::new(&shell).args(["-lc", ...])
```

But `wrap_command_ignore_suspend` hardcodes `bash -lc`. So an agent can be **detected** via `zsh -lc which pi` but then **launched** via `bash -lc` where the tool isn't in PATH.

### Suggested fix

Extract a helper (e.g., `fn user_shell() -> String`) that reads `$SHELL` with `bash` as fallback, and use it consistently in both detection and launch paths. The `stty susp undef` / single-quote escaping in `wrap_command_ignore_suspend` is POSIX-compatible, so it works with any sh-compatible shell.

jerome-benoit:
## Description

Fixes #425. Two places hardcode `bash` as the shell instead of reading `$SHELL`, causing version-manager PATHs (nvm, mise, asdf, etc.) configured in `~/.zshrc` or `~/.config/fish/config.fish` to be unavailable when AoE launches agents or runs hooks.

Extracts a `user_shell()` helper in `environment.rs` that reads `$SHELL` with `bash` as fallback, and uses it in:
- `wrap_command_ignore_suspend` (agent launch wrapper)
- `build_hook_command` local branch (repo hook execution)

Container-side shell invocations (`/bin/bash` in Docker exec, `bash` in container hooks, `sh` in Apple Container) are intentionally unchanged since the user shell may not be installed inside the image.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (opencode)


kipasdinding6969-alt (approved):
✅ Good fix. Hardcoded bash → $SHELL with proper fallback, empty-string guard, serial tests. Container hooks still use bash correctly. LGTM.

PR #467, opened 2026-03-14
Set tmux window nafeat: set tmux window name to Agent(agent_name) when creating sessionsme

zeekhuge:
## Description

This PR sets the tmux window name to `Agent({agent_name})` when creating new sessions, terminal sessions, or container terminal sessions. The agent name is the tool selected by the user (e.g., `claude`, `opencode`, etc.).
This allows users to easily identify which AI agent is running in a particular tmux window.
## PR Type
- [x] New Feature
## Checklist
## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** opus 4.5, minimax-m2.5, OpenCode
**Any Additional AI Details you'd like to share:** The feature was implemented based on user request. AI refactored the initial implementation to extract a shared function to reduce code duplication.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -201,6 +201,13 @@ impl Instance {
             self.apply_terminal_tmux_options();
         }
 
+        // Set the tmux window name to "Agent(agent_name)"
+        if is_new {
+            if let Err(e) = session.set_window_name(&self.tool) {
+                tracing::warn!("Failed to set tmux window name: {}", e);
+            }
+        }
+
         self.terminal_info = Some(TerminalInfo {
             created: true,
             created_at: Some(Utc::now()),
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -10,11 +10,44 @@ use clap_complete::generate;
...
     }
 
+    tracing::debug!(
+        "Application starting - version {}",
+        env!("CARGO_PKG_VERSION")
+    );
+
     let cli = Cli::parse();
 
     // Handle commands that don't need app data or migrations.
```
jerome-benoit: These `tracing::debug!` calls are outside the `if AGENT_OF_EMPIRES_DEBUG` block, so they're silently dropped when no subscriber is initialized. Moving them inside the debug block (or after successful subscriber init) would better communicate intent.

PR #382, opened 2026-03-06
feat(session): persist and resume agent session IDs

jerome-benoit:
## Description

Closes #343

Persists agent session IDs so sessions resume where they left off instead of starting fresh on every relaunch.

### Per-agent support

| Agent | Resume flag | Capture method |
|-------|-------------|----------------|
| Claude Code | `--session-id` / `--resume` | Pre-generated UUID; disk scan poller + `.claude.json` fallback (5-min staleness guard) |
| OpenCode | `--session` | `opencode session list --format json` with directory matching and timeout (try_wait + pipe drain) |
| Codex CLI | `codex resume <id>` | Walks `~/.codex/sessions/` date-partitioned dirs, matches by CWD |
| Gemini CLI | `--resume` | Scans `~/.gemini/tmp/*/chats/session-*.json`, matches by CWD |
| Mistral Vibe | `--resume` | Scans `~/.vibe/logs/session/*/meta.json`, matches by CWD |
| Pi | `--session <id>` | Scans `~/.pi/agent/sessions/` JSONL files, CWD-encoded directory matching + header fallback |
| Cursor / Copilot / Droid / Settl | -- | No session resume support |

### Session ID lifecycle

A single adaptive-interval poller per instance (2s -> 60s backoff) scans for session IDs using agent-specific filesystem or CLI queries. Background threads only write to tmux env; the TUI thread writes to `sessions.json` via `apply_session_id_updates()` (with rollback on save failure) to prevent concurrent file writes. Cross-instance isolation uses hidden tmux env vars (`build_exclusion_set()`). Thread budget capped at 20 via `PollerCountGuard` (RAII). Session IDs are validated at capture, poller, and TUI boundaries (`validated_session_id`).

### CLI changes

- `aoe session set-session-id <name> <id>` to manually set/clear a session ID (warns for unsupported agents)
- `aoe session show --json` now refreshes status live instead of returning stale disk values

### Other changes

- `ResumeStrategy` enum on agent registry drives per-agent resume flag generation; `supports_session_poller` derived from it
- Tmux hidden env module with batched reads and dual-TTL cache (30s positive, 5s negative)
- Refactored `Instance::start()` into `build_host_command`, `build_sandboxed_command`, `install_agent_status_hooks`, `resolve_on_launch_hooks`, `finalize_launch`
- `AOE_INSTANCE_ID` set synchronously before poller start (prevents self-exclusion race)
- Poisoned mutex recovery in `stop_poller`

### New files

[...]

PR #482, opened 2026-03-19
feat(session): extract infrastructure modules from #382

jerome-benoit:
## Description

Extracts the self-contained infrastructure layer from #382 (session persistence) as the first step of the [agreed-upon split](https://github.com/njbrake/agent-of-empires/pull/382#issuecomment-4085221093) (PR A: Infrastructure).

### What's included

| Module | Description |
|--------|-------------|
| `src/tmux/env.rs` (new) | Hidden env variable helpers for tmux sessions with 30s TTL cache and batch read/write operations |
| `src/session/poller.rs` (new) | `AdaptiveInterval` (2s->60s backoff), `CaptureGate` (condvar sync), `SessionPoller` (thread lifecycle), `PollerCountGuard` (RAII budget cap at 20 threads) |
| `src/agents.rs` | `ResumeStrategy` enum + `resume_strategy` field on every `AgentDef` |
| `src/agents.rs` | `SessionStart`/`SessionEnd` lifecycle hook events (status: None, skipped by shell one-liners) |
| `src/tmux/mod.rs` | Wire `pub mod env` |
| `src/session/mod.rs` | Wire `pub mod poller` |
| `src/hooks/mod.rs` | Test coverage for SessionStart/SessionEnd skipping |

### Why it stands alone

These are purely additive library modules with their own unit tests. `ResumeStrategy` is added to agent definitions but nothing reads it yet. Zero behavioral change -- subsequent PRs (capture engine, resume integration, config) will wire these modules into the launch path.

Code is extracted identically from the `feat/session-management` branch to minimize conflicts with #382.

## PR Type

- [x] New Feature
- [x] Refactor

## Checklist

## Test plan

- [x] `cargo fmt -- --check` -- clean
- [x] `cargo clippy -- -D warnings` -- clean
- [x] `cargo test` -- 906 unit tests pass, 0 failures
- Unit tests cover env cache populate/stale/invalidate, batch output parsing, adaptive interval backoff/reset/clamping, poller lifecycle (start/stop/change detection/budget cap/capture gate sync)

## AI Usage

- [ ] No AI was used

**NOTE:**

### DIFF
--- a/src/agents.rs
+++ b/src/agents.rs
@@ -24,6 +24,26 @@ pub enum YoloMode {
     AlwaysYolo,
 }
 
+/// How an agent resumes an existing session from the CLI.
+pub enum ResumeStrategy {
+    /// Append a flag (e.g. `--session <id>`). For agents where new and existing
+    /// sessions use the same flag.
+    Flag(&'static str),
+    /// Two different flags depending on whether conversation data already exists.
+    /// `existing` is used when there is prior conversation data (e.g. `--resume`),
+    /// `new_session` when creating/attaching unconditionally (e.g. `--session-id`).
+    FlagPair {
+        existing: &'static str,
+        new_session: &'static str,
+    },
+    /// Resume is a subcommand rather than a flag (e.g. `codex resume <id>`).
+    /// The subcommand + id are inserted right after the binary name so that
+    /// other flags land after it.
+    Subcommand(&'static str),
+    /// Agent does not support session resume.
+    Unsupported,
+}
+
 /// A single hook event that AoE registers in an agent's settings file.
 pub struct HookEvent {
     /// Event name as the agent expects it (e.g. `"PreToolUse"` for Claude Code).
@@ -39,7 +59,7 @@ pub struct AgentHookConfig {
     /// Path relative to the home dir where the agent's settings live
     /// (e.g. `.claude/settings.json`).
     pub settings_rel_path: &'static str,
-    /// Hook events to register (status transitions).
+    /// Hook events to register (status transitions and session lifecycle).
     pub events: &'static [HookEvent],
 }
 
@@ -70,6 +90,8 @@ pub struct AgentDef {
     /// hooks into the agent's settings file so status is written to a file instead
     /// of being parsed from tmux pane content.
     pub hook_config: Option<AgentHookConfig>,
+    /// How this agent resumes a prior session.
+    pub resume_strategy: ResumeStrategy,
 }
 
 /// Hook events shared by Claude Code and Cursor CLI.
@@ -94,6 +116,16 @@ const CLAUDE_CURSOR_HOOK_EVENTS: &[HookEvent] = &[
         matcher: Some("permission_prompt|elicitation_dialog"),
         status: Some("waiting"),
     },
+    HookEvent {
+        name: "SessionStart",
+        matcher: None,
+        status: None,
+    },
+    HookEvent {
+        name: "SessionEnd",
+        matcher: None,
+        status: None,
+    },
 ];
 
 pub const AGENTS: &[AgentDef] = &[
@@ -112,6 +144,10 @@ pub const AGENTS: &[AgentDef] = &[
             settings_rel_path: ".claude/settings.json",
             events: CLAUDE_CURSOR_HOOK_EVENTS,
         }),
+        resume_strategy: ResumeStrategy::FlagPair {
+            existing: "--resume",
+            new_session: "--session-id",
+        },
     },
     AgentDef {
         name: "opencode",
@@ -125,6 +161,7 @@ pub const AGENTS: &[AgentDef] = &[
         detect_status: status_detection::detect_opencode_status,
         container_env: &[],
         hook_config: None,
+        resume_strategy: ResumeStrategy::Flag("--session"),
     },
     AgentDef {
         name: "vibe",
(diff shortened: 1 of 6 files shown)


njbrake:
Hi @jerome-benoit , thank you for the willingness to split this into a smaller PR! This made it much better for me to reason about the changes and think about them wholistically. I have a comment to make that I feel really bad to say because I see all the great time you've put into this feature but I think we need to think about it: can this be simplified? 


I'll put below the conversation I had with Claude about it. As I'm wrapping my head around what is happening, I want to be sure that we're not overengineering this, and if the session id is something that we can capture soon after the agent is launched, maybe we can avoid much code complexity. You've spent much more time on this so please help me understand if I'm being way to naive about the problem 😅  Thank you! 🙏 

```
  The simple version

  The core feature is: generate/discover a session ID, store it, pass it back on restart. Let's trace what's actually needed per agent:

  Agents where you can pre-assign an ID at launch (no discovery needed):
  - Claude Code: claude --session-id <uuid> at creation, claude --resume <uuid> on restart
  - OpenCode: opencode --session <id>

  Agents that need post-launch discovery:
  - Codex, Vibe, Gemini: ID is generated internally, must be discovered after launch
  - Cursor, Copilot, Pi: Marked Unsupported in this PR anyway

  For Claude Code -- by far the most common agent -- the entire feature could be:

  // At launch:
  let session_id = Uuid::new_v4();
  cmd.push_str(&format!(" --session-id {}", session_id));
  instance.agent_session_id = Some(session_id);
  save_sessions();

  // On restart:
  if let Some(id) = instance.agent_session_id {
      cmd.push_str(&format!(" --resume {}", id));
  }

  That's maybe 50 lines of code in the builder/launch path. No polling threads, no tmux env vars, no capture gates, no adaptive intervals.

  For agents that need discovery (Codex, Vibe, Gemini), you'd need to read the agent's state directory or run gemini --list-sessions after a brief delay. That's a one-shot read
  with a retry -- maybe 100 more lines total, one function per agent.

  What the PR builds instead

  ┌────────────────────────────────────────────┬───────┬──────────────────────────────────────┬────────────────────────────────────────┐
  │                 Component                  │ Lines │                Solves                │                 But...                 │
  ├────────────────────────────────────────────┼───────┼──────────────────────────────────────┼────────────────────────────────────────┤
  │ Adaptive polling (2s->60s backoff)         │ ~200  │ "Don't waste CPU polling forever"    │ If you pre-assign IDs, you never poll  │
  ├────────────────────────────────────────────┼───────┼──────────────────────────────────────┼────────────────────────────────────────┤
  │ CaptureGate (condvar sync)                 │ ~130  │ "Don't let poller and capturer race" │ If you capture once at launch, no race │
  ├────────────────────────────────────────────┼───────┼──────────────────────────────────────┼────────────────────────────────────────┤
  │ Thread budget (RAII guard, atomic counter) │ ~50   │ "Don't spawn unbounded threads"      │ If you don't spawn polling threads...  │
  ├────────────────────────────────────────────┼───────┼──────────────────────────────────────┼────────────────────────────────────────┤
  │ Tmux hidden env vars + cache               │ ~460  │ "Communicate IDs between processes"  │ sessions.json already exists           │
  ├────────────────────────────────────────────┼───────┼──────────────────────────────────────┼────────────────────────────────────────┤
  │ SessionPoller (thread lifecycle)           │ ~200  │ "Manage background polling"          │ No background polling needed           │
  └────────────────────────────────────────────┴───────┴──────────────────────────────────────┴────────────────────────────────────────┘

  The infrastructure is solving problems that are created by choosing a polling architecture in the first place. The polling architecture seems designed around the assumption
  that session IDs might change mid-conversation and need continuous monitoring. But session IDs don't change -- they're set once at conversation start and stay fixed.

  My take

  The ResumeStrategy enum and per-agent resume_strategy field (~56 lines in agents.rs) are genuinely needed regardless of approach. That's the part I'd keep.

  The remaining ~1,350 lines of infrastructure -- and the ~3,500 lines planned across PRs B-D -- are building a sophisticated real-time monitoring system for something that could
   be a simple read-then-store at launch time.

  I'd ask the contributor: why does this need continuous polling rather than a one-shot capture? There may be a good reason I'm not seeing (maybe some agents rotate session IDs,
  or the capture needs to happen after a handshake that takes variable time). But if the answer is "it's more robust," that robustness comes at a significant complexity cost for
  a problem that might not need it.
```


Claude's analysis hinges on this statement:

`But session IDs don't change -- they're set once at conversation start and stay fixed.`

I challenged that assumption, and this was its response:

```
● Yes, it can. That changes the analysis.

  /clear (also /reset, /new) creates a new session ID within the same running process. /resume switches to a different session's ID. Both happen without restarting the Claude
  Code process. The SessionStart hook fires in both cases with the new session_id in its payload.

  So a one-shot capture at launch isn't sufficient for Claude Code -- the ID you captured at startup could be stale if the user ran /clear.

  However, this actually argues for hooks rather than polling. The PR already declares SessionStart/SessionEnd hook events with status: None. Claude Code fires SessionStart with
  the new session_id in its JSON input every time the session changes. So the simpler path would be:

  1. Extend the existing hook shell command to write the session ID (from the hook's JSON input) to /tmp/aoe-hooks/$AOE_INSTANCE_ID/session_id -- right next to the status file
  that already exists
  2. Read that file when you need it (before restart, or on a simple timer)
  3. No adaptive polling, no tmux env vars, no capture gates needed

  The hook infrastructure for this already exists in the codebase. You'd be adding one more printf to a shell one-liner, not 1,400 lines of polling infrastructure.

  That said, hooks only cover Claude, Cursor, and Gemini. For agents without hook support (Codex, Vibe, OpenCode), you'd still need some form of discovery -- but that could be a
  simple per-agent function that reads from a known state directory, not a general-purpose adaptive polling system.
```

PR #583, opened 2026-04-06
fix: stop misclassifying idle OpenCode sessions as error

njbrake:
## Description

`aoe status --json` incorrectly classifies active idle OpenCode sessions as `error`. The root cause is the `is_shell_stale()` escalation in `update_status_with_metadata()` -- when a shell is the foreground process (e.g., agent launched through a shell wrapper or spawning child shells), `Status::Idle` gets escalated to `Status::Error` even though the agent is alive.

This fix:
1. Splits the `is_dead || is_shell_stale()` condition and adds a `pane_has_agent_content()` check when `is_shell_stale()` fires. If the captured pane content shows agent UI (tool name present or substantial TUI output), the session stays `Idle`.
2. Adds a shell-prompt-at-end negative check (`$`, `#`, `%`, `❯`) to prevent false positives from servers with verbose MOTD.
3. Adds a minimum tool name length guard (>=4 chars) to avoid substring false positives for short names like "pi" matching "api".
4. Fixes `aoe session show --json` returning stale persisted status by adding `refresh_session_cache()` + `update_status()` before output.

Fixes #580

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Testing

- `cargo test` -- 982 unit tests pass, 0 failures
- `cargo clippy` -- clean
- `cargo fmt` -- clean
- 6 new unit tests for `pane_has_agent_content`:
  - bare shell prompt, agent UI, substantial output, empty content
  - verbose MOTD with shell prompt at end (edge case)
  - short tool name substring false positive (edge case)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.6


jerome-benoit:
> Hi @jerome-benoit ! Would you have a chance to try out this code for me to confirm that it fixes your problem? Since I generally use Claude Code and not OpenCode at the moment it would be good to have confirmation from an opencode user

I'm a bit busy lately, will do ASAP. At a first glance, the PR is addressing the issue.

njbrake:
@jerome-benoit thanks for bearing with me on this one and thank you for testing! Ok I implemented the fix in the other locations that you mentioned, it should be all good now if you have a chance to test it out again. No rush and thanks so much for the help

njbrake:
@jerome-benoit  I think I found something impressive, we've actually stumped Claude enough that it is telling me that "it doesn't know" 😆  I think this is a first.

Tl;dr I'm still failing to find the root cause of your issue, but Claude makes a good point that instead of me asking you to keep testing my speculative fixes, it would be better to have you generate diagnostics that will give us the detailed information that we need to truly understand the problem.

I've pushed up a new commit that adds a whole bunch of diagnostic logging. Could I trouble you one last time to check out the latest on this branch and run your process that illustrates the broken logic, then run 

```
# the debug log may be in ~/agent-of-empires/debug.log not ~/.config/agent-of-empires/debug.log if you're on macos
AGENT_OF_EMPIRES_DEBUG=1 aoe status --json 2>/dev/null; cat ~/.config/agent-of-empires/debug.log | grep "status '"

```

Here is Claude's output in case you're interested:

```
OK, I've spent significant time tracing every code path. Here's
 where I actually am:

The honest answer is: I cannot determine the root cause from
code alone. I've traced the update_status_with_metadata
function end-to-end, and for a live nix-wrapped OpenCode
session, the logic should produce Idle, not Error.
Specifically:

- session.exists() → should be true (names are deterministic,
match tmux output)
- is_dead → should be false (nix wrapper + exec chain keeps
pane alive)
- has_command_override() → false (command="opencode" ==
binary="opencode")
- is_shell_stale() → false (.opencode-wrapp not in
KNOWN_SHELLS)
- Match fallthrough → other => other → Idle

Yet jerome-benoit sees Error. Something environmental is
happening that I can't see from static analysis.

What I AM confident about:
1. The pane_has_agent_content() check on is_dead is wrong
(ghost content from remain-on-exit makes dead panes
indistinguishable from live ones)
2. The pane_has_agent_content() check on is_shell_stale() is
correct
3. Reverting the is_dead branches fixes the regression but
won't fix the original bug

My recommendation: don't ask jerome-benoit to test a fix. Ask
him to run diagnostics instead.

Instead of pushing another speculative fix that might not work,
 add a few debug! log lines to update_status_with_metadata and
ask jerome-benoit to run AGENT_OF_EMPIRES_DEBUG=1 aoe status
--json and share the debug output. That will tell us exactly
which early-exit path fires. Then we fix it once, correctly.

Want me to:
- A) Prepare the diagnostic logging patch + specific
instructions for jerome-benoit (then fix correctly based on
data)
- B) Apply the regression fix + diagnostics together (fixes the
 known regression, gathers data for the root cause)
- C) Something else?

I'd recommend B - it's the safest move. The regression fix is
definitively correct, and the diagnostics will get us the data
we need without yet another "try this and see" cycle.
```

jerome-benoit:
> @jerome-benoit I think I found something impressive, we've actually stumped Claude enough that it is telling me that "it doesn't know" 😆 I think this is a first.
> 
> Tl;dr I'm still failing to find the root cause of your issue, but Claude makes a good point that instead of me asking you to keep testing my speculative fixes, it would be better to have you generate diagnostics that will give us the detailed information that we need to truly understand the problem.

I've found the repro case: run aoe CLI in a subprocess to get the session statuses. People using the aoe openclaw skill are hitting it. In a normal login shell forked with complete parent inheritance (env, etc.), it works properly. The actual PR with the session refresh + instance update should fix the issue.

PR #616, opened 2026-04-12
fix: prevent raw JSON resize messages from appearing in web terminal

njbrake:
## Description

When opening a terminal in the web dashboard, raw JSON like `{"type":"resize","cols":null,...}` would appear as visible text. This was caused by `fitAddon.proposeDimensions()` returning `NaN` for cols/rows before the terminal container fully renders. `JSON.stringify(NaN)` becomes `null`, which fails Rust's serde deserialization (`u16` rejects `null`). The failed parse fell through to the PTY stdin handler, causing tmux to echo the raw JSON as visible text.

**Fix (defense in depth):**
- **Client** (`useTerminal.ts`): Validate that cols/rows are finite positive integers before sending the resize message
- **Server** (`ws.rs`): Drop JSON-shaped text messages that fail control message parsing instead of forwarding them to PTY stdin

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

### DIFF
--- a/src/server/ws.rs
+++ b/src/server/ws.rs
@@ -219,6 +219,10 @@ async fn handle_terminal_ws(socket: WebSocket, tmux_name: String, read_only: boo
                                 .await;
                             }
                         }
+                    } else if text.starts_with('{') {
+                        // Malformed JSON control message; drop it so raw JSON
+                        // doesn't leak into the terminal as visible text.
+                        tracing::warn!("Ignoring malformed control message: {}", text);
                     } else if !read_only {
                         // Plain text input -> PTY stdin (blocked in read-only mode)
                         let writer = writer_for_input.clone();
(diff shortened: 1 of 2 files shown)


src/server/ws.rs
```diff
@@ -219,6 +219,10 @@ async fn handle_terminal_ws(socket: WebSocket, tmux_name: String, read_only: boo
                                 .await;
                             }
                         }
+                    } else if text.starts_with('{') {
+                        // Malformed JSON control message; drop it so raw JSON
+                        // doesn't leak into the terminal as visible text.
+                        tracing::warn!("Ignoring malformed control message: {}", text);
```
njbrake:
```suggestion
```

PR #624, opened 2026-04-12
feat: support user-defined custom agents in config.toml

fluffyspace:
## Description

Adds a `[[session.custom_agents]]` config section so users can register additional agents (SSH wrappers to remote machines, custom workflows, etc.) that appear in the TUI agent picker alongside built-in agents like `claude`, `opencode`, etc.

### Motivation

Today, the agent picker in the new-session dialog only offers agents from the static `AGENTS` registry. To run Claude Code on a remote machine via SSH, the user has to press Ctrl+P and type `ssh -t lenovo claude` every single time a new session is created. This PR lets the user declare that wrapper once in config and then select it as a first-class agent.

### Example config

```toml
[[session.custom_agents]]
name = "lenovo-claude"
command = "ssh -t lenovo claude"
detect_as = "claude"
```

- `name` — display name in the agent picker
- `command` — command to run (auto-filled into the session's command override)
- `detect_as` (optional) — map a custom agent to a built-in agent's status detection so running Claude over SSH still surfaces Running/Waiting/Idle correctly

### Changes

- Add `CustomAgentDef` struct and `custom_agents: Vec<CustomAgentDef>` field on `SessionConfig`, with profile override support in `SessionConfigOverride` and `apply_session_overrides`.
- Change `AvailableTools` from `Vec<&'static str>` to `Vec<String>` so it can hold dynamic custom-agent names alongside the static builtin names. `detect()` reads custom agents from config and appends them.
- Propagate the type change through `tui/dialogs/new_session/{mod,render}.rs` and `cli/add.rs`.
- `session/builder.rs` falls back to the custom agent's command when no explicit command override is set.
- `tmux/status_detection.rs` honours `detect_as` for custom agents so remote Claude/etc. still get proper status heuristics.

### Known limitation (intentionally scoped out)

Per `CLAUDE.md`, every config field must be editable in the settings TUI. This PR does **not** add settings-TUI editing for `custom_agents` because it's a `Vec<struct>` (unlike the existing `HashMap<String, String>` list fields) and would need a new `FieldValue` variant and a richer per-row editor. Happy to fold that into this PR or ship as a follow-up — whichever you prefer.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test`, `cargo clippy`, `cargo fmt --check` all clean)
- [ ] Documentation was updated where necessary — docs/guides/configuration still needs a short section; happy to add once direction is confirmed.
- [ ] For UI changes: included screenshot or recording — no visual changes, only the picker list length grows by one per custom agent

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (Anthropic) via Claude Code

**Any Additional AI Details you'd like to share:** The human author inspected and tested the diff against their own AoE installation (local Claude + custom `lenovo-claude` over SSH). Verified build, test, clippy, and fmt all pass locally before pushing.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -223,12 +223,13 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
     } else {
         // Use default_tool from resolved config, then first available tool, then "claude"
         let available_tools = crate::tmux::AvailableTools::detect();
+        let tools_list = available_tools.available_list();
         instance.tool = config
             .session
             .default_tool
             .as_deref()
             .and_then(crate::agents::resolve_tool_name)
-            .or_else(|| available_tools.available_list().first().copied())
+            .or_else(|| tools_list.first().map(|s| s.as_str()))
             .unwrap_or("claude")
             .to_string();
     }
--- a/src/session/builder.rs
+++ b/src/session/builder.rs
@@ -351,12 +351,20 @@ pub fn build_instance(
 
     // Apply agent_command_override and agent_extra_args from resolved config.
     // Per-session values from params take priority over config.
+    // For custom agents, fall back to the custom agent's command definition.
     if !params.command_override.is_empty() {
         instance.command = params.command_override;
     } else if let Some(cmd_override) = config.session.agent_command_override.get(&params.tool) {
         if !cmd_override.is_empty() {
             instance.command = cmd_override.clone();
         }
+    } else if let Some(custom) = config
+        .session
+        .custom_agents
+        .iter()
+        .find(|c| c.name == params.tool)
+    {
+        instance.command = custom.command.clone();
     }
     if !params.extra_args.is_empty() {
         instance.extra_args = params.extra_args;
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -111,6 +111,18 @@ pub struct AppStateConfig {
     pub sort_order: Option<SortOrder>,
 }
 
+/// A user-defined custom agent that appears in the TUI agent picker.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct CustomAgentDef {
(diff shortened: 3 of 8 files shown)


njbrake:
Hey @fluffyspace — thanks for this PR! The use case you described (running Claude over SSH without manually typing the command every session) is a great feature request, and your detailed writeup made it easy to understand exactly what was needed.

After reviewing the PR, @njbrake and I (Claude, via Claude Code) discussed the implementation and decided to go with a slightly different config format. The core idea is the same — named custom agents in the picker with status detection mapping — but we used two `HashMap<String, String>` fields instead of `Vec<CustomAgentDef>`:

```toml
[session]
custom_agents = { "lenovo-claude" = "ssh -t lenovo claude" }
agent_detect_as = { "lenovo-claude" = "claude" }
```

The main reason for the change: `HashMap<String, String>` slots into the existing settings TUI list editor (same pattern as `agent_extra_args` and `agent_command_override`), which addresses the settings TUI gap you noted as a known limitation. We also moved the `detect_as` resolution to session creation time so the status polling loop doesn't need to load config.

**For your SSH setup, the equivalent config is:**

```toml
[session]
default_tool = "lenovo-claude"
custom_agents = { "lenovo-claude" = "ssh -t lenovo claude" }
agent_detect_as = { "lenovo-claude" = "claude" }
```

This gives you "lenovo-claude" as a first-class agent in the picker (and as the default), with Claude's Running/Waiting/Idle detection. Both fields are editable from the settings TUI and support profile/repo overrides.

The reimplementation is in #628. Closing this one in favor of that PR — but full credit to you for the feature idea and the thorough design work that informed the final implementation. 🙏

*This comment was posted by Claude (via Claude Code) from a discussion with @njbrake.*

PR #754, opened 2026-04-20
fix(web): iOS keyboard auto-switches back to letters after punctuation

njbrake:
## Description

On iOS mobile, when the user switches the soft keyboard to the punctuation/numbers view (by tapping "123"), typing a comma, period, or other punctuation character should auto-switch the keyboard back to the letters view. This is standard iOS behavior in native text fields, but it wasn't working in the web terminal.

**Root cause:** wterm's `handleKeyDown` calls `e.preventDefault()` on every keydown event to intercept input for the terminal emulator. This prevents iOS from running its native keyboard behaviors, including the auto-switch from punctuation back to letters.

**Fix:** On mobile, intercept all keydown events in capture phase before wterm sees them. Send the character or escape sequence directly over the WebSocket, and let the native keydown proceed unblocked so iOS keyboard behaviors work. A capture-phase `input` listener prevents wterm's `handleInput` from double-sending characters that the native keydown inserts into the textarea.

This also preserves the existing backspace key-repeat fix and IME composition input.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**Any Additional AI Details you'd like to share:** Root cause analysis of wterm's preventDefault behavior and the fix were both AI-generated.

## Test plan
- [ ] Test on iOS Safari: open web terminal, tap "123" on keyboard, type comma/period, verify keyboard auto-switches back to letters
- [ ] Test on iOS Safari: verify regular typing still works (letters, numbers, special characters)
- [ ] Test on iOS Safari: verify backspace and backspace key-repeat still work
- [ ] Test on iOS Safari: verify Enter, Tab, arrow keys work
- [ ] Test on iOS Safari: verify Ctrl+C, Cmd+V still work (modifier combos pass through to wterm)
- [ ] Test on iOS Safari: verify IME/composition input still works (for CJK keyboards)
- [ ] Verify desktop browser is unaffected (mobile-only code path behind isMobileViewport())

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -984,10 +984,15 @@ fn format_env_var_prefix(key: &str, value: &str, cmd: &str) -> String {
 /// Single quotes in `cmd` are escaped with the `'\''` technique to prevent
 /// breaking out of the outer single-quoted wrapper.
 fn wrap_command_ignore_suspend(cmd: &str) -> String {
-    let shell = super::environment::user_posix_shell();
+    let user = super::environment::user_shell();
+    let posix = super::environment::user_posix_shell();
     let escaped = cmd.replace('\'', "'\\''");
     // Use login shell (-l) so version-manager PATHs (NVM, etc.) are available.
-    format!("{} -lc 'stty susp undef; exec env {}'", shell, escaped)
+    // Skip -l when falling back to bash for a non-POSIX user shell (fish, nu,
+    // pwsh): bash's login scripts won't contain the user's PATH setup and -l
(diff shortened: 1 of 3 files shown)


njbrake:
Closing — the 123→ABC auto-return after punctuation is not a standard iOS keyboard behavior in web textareas. Tested two approaches (bypassing wterm's preventDefault on keydown, overriding autocapitalize to 'sentences'), neither had any effect. This is a platform limitation, not a bug.

PR #778, opened 2026-04-23
feat(tui): configurable default_view_mode for home screen

BTForIT:
## Description

The home view currently launches on the agent list (`ViewMode::Agent`) with no way to configure it — `view_mode: ViewMode::default()` at `src/tui/home/mod.rs`. `sort_order` and `group_by` already persist to `AppStateConfig` and load at startup; this PR adds the analogous `default_view_mode` field so users who prefer the paired-terminal view can land on it without pressing `t` every session.

**What it does:**
- Moves `ViewMode` enum from `src/tui/home/mod.rs` to `src/session/config.rs` (alongside `SortOrder`/`GroupByMode`) so it can be serde-serialized without a `tui → config` layering violation.
- Re-exports `ViewMode` from the home module so sibling submodules (`render`, `input`, `tests`) continue to resolve `super::ViewMode` unchanged.
- Adds `default_view_mode: Option<ViewMode>` to `AppStateConfig` with `#[serde(default, skip_serializing_if = "Option::is_none")]`.
- Loads at `HomeView::new` next to the existing `sort_order` / `group_by` load, falling back to `ViewMode::default()` (Agent) when unset.

**Backward compat:** unset → Agent, same as today. No config migration. Existing `config.toml` files are untouched until a user opts in with `default_view_mode = "agent"` or `"terminal"` under `[app_state]`.

**Diff:** 3 files, +53/-9. No new dependencies.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

(Framed as a bug fix — `sort_order` and `group_by` are persisted; `view_mode` being the lone non-persisted home-screen preference is the gap this closes.)

## Checklist

- [x] New and existing tests pass (122 `tui::home::tests` green; 2 new cases cover the load path and the unset-falls-back-to-Agent backward-compat path)
- [x] Documentation was updated where necessary (doc comment on `ViewMode` enum + `default_view_mode` field)
- [ ] For UI changes: included screenshot or recording — **N/A, no visual change**; this is a config-loading behavior change. Behavior verified via `cargo test --lib tui::home::tests` and manual launch (`cargo run` with `[app_state] default_view_mode = "terminal"` in `~/.agent-of-empires/config.toml` lands on the Terminal view; removing the key returns to Agent).

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Code, Opus 4.7

**Any Additional AI Details you'd like to share:** Claude drafted the change based on a code survey of the existing sort_order/group_by persistence pattern and mirrored it for view_mode. I reviewed the final diff and ran `cargo test --lib tui::home::tests` + `cargo clippy -- -D warnings` + `cargo fmt --check` locally before pushing.

**NOTE:**


njbrake:
Nice clean PR, thanks! One design question before I merge:

Right now `sort_order` and `group_by` both persist immediately when the user toggles them with `o`/`g`, so the app remembers your last choice across restarts. With `default_view_mode`, the user has to manually edit `config.toml` to change it, and toggling with `t` is ephemeral (resets on restart).

Would you be open to making `t` persist the view mode the same way `o` and `g` do? That way users can just toggle once and it sticks, rather than needing to know about the config file. The field could stay named `default_view_mode` or we could rename it to just `view_mode` to match `sort_order`/`group_by` more closely.

Either way the current implementation is solid; just wondering if this would be a better UX before we land it.

*Posted by Claude on behalf of @njbrake, based on our review discussion.*

PR #796, opened 2026-04-24
feat(tui): add w/W hotkeys to jump to next waiting session

mguthaus:
## Description

Adds two new hotkeys to the home view for quickly navigating to sessions that are waiting for user input (`Status::Waiting`):

- **`w`** — jump to the next waiting session within the same project as the currently selected session (matches on `main_repo_path` for worktree sessions, `project_path` otherwise)
- **`W`** — jump to the next waiting session across all projects

Both keys cycle forward through visible sessions, skipping the currently selected one, and wrap around. If no other waiting session is found, an info dialog is shown. Both are listed in the `?` help overlay under Navigation.

## PR Type

- [x] New Feature

## Checklist

## Testing

- `cargo test --lib` passes (1106 tests; 2 pre-existing Docker network failures unrelated to this change)
- Built and manually tested with `cargo build --profile dev-release`
- Help dialog size test continues to pass (new entry fits within the 38-line inner height)

## AI Usage

### DIFF
--- a/src/tui/components/help.rs
+++ b/src/tui/components/help.rs
@@ -27,7 +27,7 @@ fn shortcuts(strict: bool) -> Vec<(&'static str, Vec<(&'static str, &'static str
                     ("l/→", "Expand group"),
                     ("Home/End", "Go to top / bottom"),
                     ("PgUp/Dn", "Move 10 items up / down"),
-                    ("w/W", "Next waiting (project/global)"),
+                    ("w/W", "Next waiting/idle (project/global)"),
                 ],
             ),
             (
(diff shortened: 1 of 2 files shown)


njbrake:
Thank you @mguthaus for the contribution! This is a very nice convenience feature, as soon as I saw it i was like 'oooooo i wish i had thought of this earlier'. It's going to immediately enter my toolbox as a daily used feature. 

For this first PR though I filtered it down to only use 'w' for a general jump to the next waiting agent, instead of having a 'W' option that jumps to the next waiting agent inside a group. To me it seemed like slightly more complication than would be useful for users, but if you had a specific reason in mind for why you wanted the ability to jump to the next waiting agent only filtered by the current group, I'm open to further discussion.

mguthaus:
I started using it and I think that the W was limited use anyhow. So this change makes sense to me as well.

Thanks for AoE!

PR #810, opened 2026-04-25
fix(web): pin terminal scroll on wterm scrollTop reset

njbrake:
## Description

Fixes a mobile-only bug where the agent terminal's prompt/cursor would slip below the visible region after typing, particularly after backspace, and stay hidden until the soft keyboard opened or closed.

Root cause: `@wterm/dom` occasionally resets `scrollTop` to `0` mid-session when its renderer redraws, and its post-render auto-scroll skips because `_isScrolledToBottom()` reads stale dimensions on the same task. The result is `scrollHeight > clientHeight` while `scrollTop === 0`, leaving the cursor row clipped off-screen. There is already a workaround for this race on viewport changes (the keyboard open/close path); this PR extends the same protection to mid-session redraws.

Approach: listen for `scroll` events at the document level (capture phase, since scroll events don't bubble) and snap `scrollTop` back to `scrollHeight` when drift is detected. The element reference is resolved at scroll time, not attach time, which avoids two races: `state.connected` flipping `true` before wterm finishes init, and the scroll target sometimes being a descendant of `wterm.element`. The handler bails on `state.isInScrollback` so it doesn't fight the user when they intentionally scroll up; one rAF of debounce gives wterm's own scroll handler a frame to flip that flag first.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## How tested

- iPhone Firefox (portrait): typing then backspace consistently triggered the bug pre-fix; with the fix the prompt stays visible and the wterm element stays pinned to the bottom.
- `cargo test`, `tsc -b`, `eslint` pass (only pre-existing unrelated lint warnings).
- Scrollback path verified: scrolling up engages `state.isInScrollback` and the listener no-ops, so the fix doesn't interfere with scrollback entry.

## AI Usage

- [ ] No AI was used

The fix landed after a few iterations: a `MutationObserver` approach was tried first but never fired because `@wterm/dom` renders to canvas; a polling fallback worked but introduced visible jitter; the document-level scroll listener with target-resolved-at-scroll-time is the version that works without polling and without the attach race.

### DIFF
--- a/web/src/components/RightPanel.tsx
+++ b/web/src/components/RightPanel.tsx
@@ -109,6 +109,38 @@ function PairedTerminal({
     };
   }, [keyboardHeight, keyboardOpen, termRef]);
 
+  // Pin scrollTop on wterm scrollTop=0 mid-session resets (same fix
+  // as TerminalView). See that file for the rationale; this mirror
+  // covers the side terminal pane in the diff viewer.
+  const isInScrollbackRef = useRef(state.isInScrollback);
+  useEffect(() => {
+    isInScrollbackRef.current = state.isInScrollback;
+  }, [state.isInScrollback]);
+  useEffect(() => {
+    let raf = 0;
+    const onScroll = (e: Event) => {
+      const el = termRef.current?.element;
+      if (!el) return;
+      const target = e.target as Node | null;
+      if (target !== el && !(target && el.contains(target))) return;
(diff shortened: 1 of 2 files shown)


javi0guerrero4-cloud (commented):
Yes

Issue #757, opened 2026-04-20
Re-attaching to a session restarts it when using `fish` shell


facutuesca:
Hi! First of all, thanks for `aoe`, it's a great tool. 

One issue that I noticed when I switch my shell to `fish` is that, after creating a session and detaching from it, when I try to re-attach to it the session restarts.

I could reproduce it on Linux and MacOS, with `pi` and `claude`.

I created a small [dockerfile](https://gist.github.com/facutuesca/0f3fc2934e3b61df4e12b212c90824b9) with `fish` and `aoe` already set up to make it easy to reproduce:
1. `cd` into the directory containing the dockerfile
2. Build the image: `docker build -t aoe-fish .` 
3. Run the container: `docker run --rm -it aoe-fish`
4. Run `aoe`, create a new session, type something
5. Detach
6. Re-attach

On 6, the session restarts instead of attaching to the existing one.

facutuesca:
The change in #758 didn't fix it for me. The PR explanation states:

> Fish users configure PATH in ~/.config/fish/config.fish, not in ~/.bash_profile, so bash's login environment doesn't have the agent binary on PATH

But my Bash (and Zsh on MacOS) configs do have the correct PATH set, so that wasn't the issue.

njbrake:
Thanks! Since I don't use fish I'm relying on Claude heavily for this one :) Ok it should be all good for you now.

facutuesca:
That fixed it, thanks!

PR #819, opened 2026-04-26
fix(tui): remove exec before wrapped command to fix pane death on launch for custom sandboxes

blaisepic:
## Description
As far as I can tell, every sandboxed session is currently broken at startup. TERM and COLORTERM are
always added as Inherit env entries, so the export prefix is never empty,
causing the double-exec-into-login-shell path to always run. The exec           
before the wrapped command combined with wrap_command_ignore_suspend's own
login shell (-lc) killed the tmux pane immediately on every sandboxed launch

`wrap_command_ignore_suspend` already produces a command starting with `exec`:

```
exec bash -lc 'stty susp undef; exec env docker exec -it <container> claude'
```

That becomes `wrapped`. The exports prefix then prepended another `exec` in front of it:

**Before:**
```
export TERM=xterm-256color; export COLORTERM=truecolor; exec exec bash -lc 'stty susp undef; exec env docker exec -it <container> claude'
```

**After:**
```
export TERM=xterm-256color; export COLORTERM=truecolor; exec bash -lc 'stty susp undef; exec env docker exec -it <container> claude'
```

**Why before breaks:** `exec` is a shell builtin — it doesn't exist as a standalone binary on PATH. So `exec exec bash ...` tells the shell to replace itself with a program literally named `exec`, which can't be found, so it exits with "command not found" and the pane dies immediately. The fix just removes the redundant `exec` from the exports prefix, since `wrapped` already has one.

## Pane death before vs pane working after
Lmk if you want a video or anything for better fleshed-out evidence.

**Before:**
pane_death

**After:**
pane_working

## PR Type

- [ ] New Feature
- [X] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [X] I understand the code I am submitting

## AI Usage

- [ ] No AI was used
- [] AI was used for drafting/refactoring
- [X] This is fully AI-generated

claude 4.7

**NOTE:**

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -505,12 +505,7 @@ impl Instance {
             let env_part = format!("{} ", docker_args);
             let wrapped =
                 wrap_command_ignore_suspend(&container.exec_command(Some(&env_part), &tool_cmd));
-            if env_info.exports.is_empty() {
-                Some(wrapped)
-            } else {
-                let exports = env_info.exports.join("; ");
-                Some(format!("{}; {}", exports, wrapped))
-            }
+            Some(prepend_exports(&env_info.exports, wrapped))
         } else {
             // Run on_launch hooks on host for non-sandboxed sessions
             if let Some(ref hook_cmds) = on_launch_hooks {
@@ -1005,6 +1000,23 @@ fn wrap_command_ignore_suspend(cmd: &str) -> String {
     )
 }
 
+/// Prepend shell `export` statements to an already-wrapped sandbox command.
+///
+/// `wrapped` MUST be the output of `wrap_command_ignore_suspend`, which
+/// guarantees a leading `exec`. This function therefore MUST NOT add another
+/// `exec` of its own: in bash, `exec exec <cmd>` searches PATH for a binary
+/// literally named `exec`, fails with exit 127, and kills the tmux pane on
+/// every sandboxed launch. zsh-on-macOS happens to tolerate the double-exec,
+/// which is why this regression hid for several days after #757 added the
+/// leading `exec` to `wrap_command_ignore_suspend`. See PR #819.
+fn prepend_exports(exports: &[String], wrapped: String) -> String {
+    if exports.is_empty() {
+        wrapped
+    } else {
+        format!("{}; {}", exports.join("; "), wrapped)
+    }
+}
+
 /// Check whether captured pane content indicates a living agent rather than
 /// a bare shell prompt. Used to prevent `is_shell_stale()` from producing
 /// false `Error` status when the agent binary is a shell wrapper or spawns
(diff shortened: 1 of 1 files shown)


njbrake:
> *Note: this comment was drafted by Claude after discussion with @njbrake.*

Investigated the strong claim that *every* sandboxed session is broken — turns out it's shell-dependent. Pushed a follow-up commit to your branch to lock in the fix.

## What I found

The bug-shape is real and your fix is correct. But on macOS+zsh the buggy command tolerates the double-`exec`:

```
export TERM='tmux-256color'; export COLORTERM='truecolor'; export GH_TOKEN='...';
exec exec /bin/zsh -lc 'stty susp undef; exec env docker exec -it ... claude ...'
```

I confirmed this directly: I have a sandboxed session created today by a binary built today from current main (so post-#757), with that exact `pane_start_command`, and the pane is alive (`pane_dead=0`, `pane_current_command=docker`). bash exits 127 on `exec exec foo` (verified empirically); zsh-on-macOS does not. That's why the bug went undetected for several days after #757 merged on 2026-04-21 — most users on this stack didn't notice.

So the issue title is right (*"fix pane death on launch for custom sandboxes"*), but the description is overstated — the actual scope is *"sandboxed sessions are broken at startup on shells that reject another builtin name as `exec`'s argument (notably bash); zsh-on-macOS happens to tolerate it, which is why the regression hid."* No action needed from you on this; we'll tweak the description on our end before merging.

## Follow-up commit I just pushed

`d3a4f4a` (`test: add prepend_exports helper + regression test for double-exec`) does two small things:

1. **Restores the `why` comment** that was deleted in `ff07aec`. Without it, the natural way to optimize the secret-export window again is to add back `exec`, which would silently regress on bash users.
2. **Adds a regression test.** Extracts a tiny `prepend_exports(exports, wrapped)` helper so the test exercises the actual code path. The test asserts the resulting cmd never contains `exec exec`, which would have caught this at PR-time when #757 added the leading `exec` to `wrap_command_ignore_suspend`.

Net diff: +54 / -6. All 48 instance tests pass, fmt/clippy clean.

blaisepic:
> > _Note: this comment was drafted by Claude after discussion with @njbrake._
> 
> Investigated the strong claim that _every_ sandboxed session is broken — turns out it's shell-dependent. Pushed a follow-up commit to your branch to lock in the fix.
> 
> ## What I found
> The bug-shape is real and your fix is correct. But on macOS+zsh the buggy command tolerates the double-`exec`:
> 
> ```
> export TERM='tmux-256color'; export COLORTERM='truecolor'; export GH_TOKEN='...';
> exec exec /bin/zsh -lc 'stty susp undef; exec env docker exec -it ... claude ...'
> ```
> 
> I confirmed this directly: I have a sandboxed session created today by a binary built today from current main (so post-#757), with that exact `pane_start_command`, and the pane is alive (`pane_dead=0`, `pane_current_command=docker`). bash exits 127 on `exec exec foo` (verified empirically); zsh-on-macOS does not. That's why the bug went undetected for several days after #757 merged on 2026-04-21 — most users on this stack didn't notice.
> 
> So the issue title is right (_"fix pane death on launch for custom sandboxes"_), but the description is overstated — the actual scope is _"sandboxed sessions are broken at startup on shells that reject another builtin name as `exec`'s argument (notably bash); zsh-on-macOS happens to tolerate it, which is why the regression hid."_ No action needed from you on this; we'll tweak the description on our end before merging.
> 
> ## Follow-up commit I just pushed
> `d3a4f4a` (`test: add prepend_exports helper + regression test for double-exec`) does two small things:
> 
> 1. **Restores the `why` comment** that was deleted in `ff07aec`. Without it, the natural way to optimize the secret-export window again is to add back `exec`, which would silently regress on bash users.
> 2. **Adds a regression test.** Extracts a tiny `prepend_exports(exports, wrapped)` helper so the test exercises the actual code path. The test asserts the resulting cmd never contains `exec exec`, which would have caught this at PR-time when [Re-attaching to a session restarts it when using `fish` shell #757](https://github.com/njbrake/agent-of-empires/issues/757) added the leading `exec` to `wrap_command_ignore_suspend`.
> 
> Net diff: +54 / -6. All 48 instance tests pass, fmt/clippy clean.

sweet, thanks for the review and additional context!

PR #822, opened 2026-04-27
fix(web): collapse init-time PTY resize storm causing #807 garbled output

### LINKED ISSUE #807: Garbled output in Firefox on MacOS
```
aoe --version
aoe 1.4.6

macOS 26.4.1 (25E253)

opencode --version
1.14.24
```

I'm running aoe --serve on one machine and connecting from another on the same lan. When running AOE with Opencode, I get garbled output in the web browser. See attached screenshots.

Image
Image

Image

njbrake:
## Description

Fixes #807. \`useTerminal\` sent \`term.cols\`/\`term.rows\` over the WebSocket inside \`ws.onopen\` before wterm's \`ResizeObserver\` had measured the container. wterm's constructor defaults those to a hardcoded \`80x24\` (\`@wterm/dom\` 0.1.8), so the first thing the server saw on every session open was a resize to \`80x24\`, followed ~50ms later by a resize to the real container size. Each distinct size becomes a SIGWINCH at the regular-screen TUI (opencode, Claude Code) and the previous frame falls into tmux scrollback as a stacked banner.

\`b961465\` only gated resizes during scrollback reading; the init-time storm produced the same "garbled output" symptom #807 reports, especially in Firefox where the gap between WS-open and the first \`ResizeObserver\` fire is wider than in Chromium.

### Evidence

A Playwright probe with mocked WebSocket counted resize messages in the first 800ms after opening a session:

| Browser | Total messages | At wterm default (80x24) | Distinct sizes |
|---------|----------------|--------------------------|----------------|
| Chromium | 8 | 3 | 80x24, 74x43, 45x19 |
| Firefox  | 8 | **6** | 80x24, 70x43, 43x19 |

Firefox sat at \`80x24\` for 6/8 messages — that extra time at the wrong size is what makes the artifact land hardest in Firefox.

### Fix

Track the most recent measurement in a ref populated by wterm's \`onResize\` callback. Read from that ref in \`ws.onopen\` instead of from \`term.cols\`/\`term.rows\`. On first connect the ref is null and the upcoming \`onResize\` callback becomes the source of truth; on reconnect the ref is populated, so the new server-side handler picks up the right size immediately.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

### Tests

- New regression test \`web/tests/init-resize-storm.spec.ts\` asserts no \`80x24\` resize is sent during the first second after session open. Verified the test **fails on the old code** (\`received: [{80,24},{80,24}]\`) and **passes on the new code**.
- Full \`npx playwright test\` suite: 95 passed / 2 skipped (vs 93 / 2 before).
- Did not run a real \`aoe serve\` end to end — sandbox has no \`tmux\`.

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Claude Code)

Investigation, hypothesis testing, fix, and regression test were drafted by Claude Code with my review and direction. The Playwright probe that produced the table above was written and run by Claude Code in a sandbox; I'll re-verify against a real \`aoe serve\` + Firefox before merging.

### DIFF
--- a/web/src/hooks/useTerminal.ts
+++ b/web/src/hooks/useTerminal.ts
@@ -25,6 +25,13 @@ const MOBILE_BREAKPOINT_PX = 768;
 const WHEEL_ZOOM_SENSITIVITY = 0.05;
 const WHEEL_PERSIST_DEBOUNCE_MS = 400;
 const RESIZE_DEBOUNCE_MS = 50;
+// First-resize debounce: longer than the steady-state value so the
+// initial layout transition (sidebar mount, splitter snap, font swap)
+// settles into a single PTY resize instead of one per stable point.
+// CSS transitions in the dashboard run ~200ms; 250ms covers them with
+// a small margin. After the first resize lands the debounce drops to
+// RESIZE_DEBOUNCE_MS so live splitter drags still feel responsive.
+const INITIAL_SETTLE_MS = 250;
 
 export interface TerminalState {
   connected: boolean;
@@ -80,9 +87,9 @@ export function useTerminal(
   // and sending those over WS makes the server resize the PTY before
   // wterm has measured the container. Each spurious resize fires
   // SIGWINCH at the regular-screen TUI (opencode, Claude Code) and the
-  // previous frame ends up stacked into tmux scrollback as "garbled
-  // output" (issue #807). Holding off until first measurement collapses
-  // the init storm to a single resize at the correct size.
+  // previous frame ends up stacked into tmux scrollback as garbled
+  // output. Holding off until first measurement collapses the init
+  // storm to a single resize at the correct size.
   const lastMeasuredRef = useRef<{ cols: number; rows: number } | null>(null);
   // Set inside the effect; the scrollback-watch effect calls it to flush
   // a deferred resize without poking React state.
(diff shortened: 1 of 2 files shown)


njbrake:
Implemented all four follow-up suggestions from the review (commit 567498b):

1. **Dropped `(issue #807)` from the inline comment** — per AGENTS.md guidance against inline issue refs.
2. **Pre-measured cell size and seeded the `WTerm()` constructor** with computed cols/rows. Probe span goes in *before* construction so the WASM bridge initializes at the right size; wterm's later ResizeObserver fire becomes a no-op when sizes match.
3. **`INITIAL_SETTLE_MS = 250`** — first-resize debounce is now long enough to span dashboard layout transitions (sidebar mount, splitter snap, font swap). Subsequent resizes drop back to the existing 50ms so live splitter drags still feel responsive.
4. **Dedup in `sendResize`** for consecutive identical cols/rows, with a reset on every `ws.onopen` so reconnects to a fresh server-side handler still learn the size.
5. **Dropped the redundant second regression test** — subsumed by the first.

### Measured impact (same Playwright probe, chromium dashboard)

| | Total resize msgs | Distinct sizes |
|---|---|---|
| Before any fix | 8 | 3 (incl. `80x24` default) |
| After commit 8ce108e | 4 | 2 |
| After commit 567498b | **2** | 2 (one per terminal pane) |

That's the floor with two terminals (`TerminalView` + `RightPanel`) mounting concurrently with different widths. Each pane sends exactly one resize at its real measured size; no `80x24` ever leaves the client.

Full Playwright suite: 94 passed / 2 skipped.

njbrake:
Addressed the review's two issues + suggestion (commit bde0975):

1. **Font-load race (Issue 1):** extracted `computeSeedSize` as a reusable function and scheduled a `document.fonts.ready` callback that re-measures and dispatches `term.resize()` if the corrected size differs from the synchronous seed. Routes through the same `onResize → debounce → dedup`, so a no-op correction stays a no-op. Guarded against effect teardown via a `termRef` identity check.

2. **`seedSize === null` fallback (Issue 2):** kept as documented behavior — when `termEl` mounts at 0×0, `seedSize` is `null`, `WTerm()` falls back to its 80x24 default, and `lastMeasuredRef` stays `null` so `ws.onopen` doesn't send. Once layout makes the container measurable, wterm's ResizeObserver fires `onResize` and the real size goes out. No code change.

3. **Stricter regression test (Suggestion):** added a second test alongside the original. Asserts (a) total messages ≤ 4 (one per terminal pane plus jitter) and (b) total messages ≤ distinct sizes (catches dedup regression). Verified by disabling just the dedup `if`-check: the bounded test fails as expected, the original "no 80x24" test still passes (proving the two tests catch independent invariants).

Full Playwright suite: **95 passed / 2 skipped** (94 → 95 from the new test).

Issue #831, opened 2026-04-27
Web rendering issues in Google Chrome with Cursor-Agent on MacOS


aherrnvidia:
Related to https://github.com/njbrake/agent-of-empires/issues/830, but with Cursor-agent rather than Claude

```aoe 1.4.6  -- built from main on commit 843ab998ccea3b9a8122c6669b30d890a9a714ff
Mac OS Tahoe 26.3.1
claude --version
2.1.79 (Claude Code)
```

The rendering issue with Cursor is worse, its rendered unusuable. Attempting to open a new session with Cursor triggers a Workspace Trust Required page that renders out of order. Using up/down arrows to try to select either option (Trust or Quit) does not work. Scrolling the mouse wheel results in redrawing the pane (see the second screenshot). 

<img width="958" height="400" alt="Image" src="https://github.com/user-attachments/assets/8055b07c-c9d4-4348-9f74-d3e4e1e2b74c" />

<img width="2239" height="591" alt="Image" src="https://github.com/user-attachments/assets/8dae1427-fcd1-42fc-b751-e4bb5f6de1c1" />

aherrnvidia:
Rather than filing a third issue, I can report the same problem as above in Codex. Any selectable menu is unusuable and trying to select any option just results in garbled output 

<img width="673" height="215" alt="Image" src="https://github.com/user-attachments/assets/83504611-db3e-4b4c-bdf8-0a2e0a80cfec" />

njbrake:
> _Posted by Claude (Anthropic) through a debug session with @njbrake._

Update on the investigation. Sister to the comment on #830 — same root cause analysis applies, with one Cursor‑agent‑specific note.

## Likely primary cause: upstream wterm#49 (DEC alternate character set)

Cursor-agent and claude (and tmux's own pane separators) all use the DEC alternate character set (`\x1b(0`) for box-drawing glyphs. wterm hasn't implemented SCS, so `q` (which should render as `─`) shows up as a literal `q`. That's the most visible glyph corruption in `aoe serve`'s dashboard right now and is most likely what's making the "Workspace Trust Required" page look "rendered out of order". Filed upstream by another user as vercel-labs/wterm#49 prior to this investigation. Not addressable on the host side.

## Cursor‑agent‑specific note: mouse modes (vercel-labs/wterm#55)

[...]

Issue #830, opened 2026-04-27
Web rendering issues in Google Chrome with Claude on MacOS


aherrnvidia:
Hi again!

Trying out AOE again after your latest fixes to the other issue I filed (thanks for that btw!). Slightly different setup today 

```
aoe 1.4.6  -- built from main on commit 843ab998ccea3b9a8122c6669b30d890a9a714ff
Mac OS Tahoe 26.3.1
claude --version
2.1.79 (Claude Code)
```

Reproduction steps:

* Start a new project with claude
* Use any skill/command / command that expands a menu
* Menu overlaps bottom of the window

<img width="2520" height="971" alt="Image" src="https://github.com/user-attachments/assets/45214b1e-24fc-4d3a-b36b-e77e052c7676" />

Scrolling the command list with arrow keys results in more overlapping text/the menu not actually scrolling and being rewritten

<img width="1313" height="357" alt="Image" src="https://github.com/user-attachments/assets/b1898d95-8ab2-470c-8881-ddfc88c87c95" />

Also seeing the status pane at the bottom be rewritten and duplicated

<img width="2233" height="1229" alt="Image" src="https://github.com/user-attachments/assets/cc76663a-cb29-4d3f-85c0-56b6374c6e6b" />

njbrake:
Wooo thank you for the bug reports! There's a lot to iron out with the web side of things and these bug reports are super useful to help nail down the pieces that are broken. I'll take a look

aherrnvidia:
You're welcome! When I have more time I'll try to start submitting bug reports with fixes too :) 

This is definitely less garbled than it was before your first set of fixes on the web so progress is being made!

njbrake:
One thing I'm working on now: if you do the command `/tui fullscreen` in Claude Code, I think this makes the rendering way better. Claude Code is trying to fix their own issues, and I think we're currently working around some issues that Claude knows it has and is trying to fix https://code.claude.com/docs/en/fullscreen

njbrake:
> _Posted by Claude (Anthropic) through a debug session with @njbrake._

Update on the investigation, with empirical results from a real `aoe serve` end-to-end test.

## Likely cause: upstream wterm#49 (DEC alternate character set)

[...]

PR #832, opened 2026-04-27
fix(web): cap PTY resize at wterm's 256x256 grid

njbrake:
> _Authored by Claude (Anthropic) through a debug session with @njbrake. The original framing of this PR overstated its connection to #830/#831; the description below has been corrected to match what an end-to-end test actually showed (see "Empirical update" section)._

## Description

Defensive workaround for vercel-labs/wterm#56 (the WASM grid is hardcoded to 256×256 and silently clamps any larger resize). Refs #830, #831 — see "Empirical update" for why this is no longer claimed as a fix.

### What changed

- `web/src/hooks/useTerminal.ts`: clamp `cols ≤ 256` and `rows ≤ 256` inside `sendResize` (and the scrollback-pending path) before the WebSocket send. Comment block points at the upstream issue.
- `src/server/ws.rs`: matching server-side clamp on the `Resize` arm as defense-in-depth against future client regressions.
- `web/tests/wterm-grid-cap.spec.ts`: new Playwright regression at a 4000×2400 viewport asserting no resize message exceeds 256 in either dimension.

### What this prevents

When the browser viewport exceeds 256 cols/rows (very wide desktop windows, ~2150+ px wide at default font), the client used to forward the unclamped measurement to the PTY. The agent then drew into a viewport bigger than wterm renders. Lines longer than 256 chars wrapped inside wterm but not in the agent's mental model, displacing every subsequent row. Capping both sides keeps the PTY size aligned with wterm's render grid. The PR removes a class of visible corruption that would otherwise hit any user on a wide monitor.

### Empirical update

I ran an end-to-end test (real `aoe serve` + tmux + claude/codex through the web at 1920×1080 and 3000×1200) and the visible rendering corruption in #830/#831 is **most likely upstream wterm#49** (DEC alternate character set: `\x1b(0` followed by `q` should render as `─`, but wterm renders the literal `q`). That's the conspicuous "qqqqqqq" row you can see at the top of claude's workspace-trust dialog and around every tmux pane border in the e2e screenshots. **PR #832's clamp doesn't fire at typical viewports**: at 1920px the cols compute to ~228, below the cap.

So this PR is defensible as **preventive infrastructure for a real wterm gap, not a fix for the screenshots in #830/#831**. The "Fixes" markers have been removed accordingly.

### Upstream issues filed

- vercel-labs/wterm#54 — wide chars / EAW desync (latent; not exercised by claude/codex through the relay)
- vercel-labs/wterm#55 — mouse modes silently ignored (focus-event part is exercised by claude + codex; mouse-tracking part is latent)
- vercel-labs/wterm#56 — grid cap (this PR's workaround)
- vercel-labs/wterm#57 — synchronized output silently ignored (defended against in the now-closed sibling PR #836; not exercised through tmux)

The conspicuous bug — vercel-labs/wterm#49 (DEC alternate character set) — was filed by another user prior to this investigation and is not addressable on the host side.

The constants `WTERM_MAX_COLS` / `WTERM_MAX_ROWS` reference vercel-labs/wterm#56 in their doc comments and should be removed once wterm raises or removes the cap.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Manual checks run locally:
- `cargo build --features serve` ✓
- `cargo test --features serve --lib server::ws` (10/10 pass) ✓
- `cargo fmt --check` ✓
- `cargo clippy --features serve --no-deps` (no warnings) ✓
- `tsc -b` ✓
- `eslint` on changed files ✓
- End-to-end test against real `aoe serve` + tmux + claude/codex (didn't reproduce #830/#831 corruption with this branch applied; the bug visible at 1920 + 3000 px viewports was wterm#49 / DEC charset, which is upstream-only)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Anthropic), via Claude Code, in a debug session with @njbrake.

The investigation read the wterm source (Zig core + DOM TypeScript) to identify the gap, captured real PTY byte streams from claude / codex / cursor-agent with `script(1)` to validate the cause, and ran a real `aoe serve` end-to-end test (after installing tmux in the sandbox) to see what the rendering actually looks like with both candidate workarounds applied. The end-to-end test produced the empirical correction that walked back the original "fix" framing.

Refs #830, #831.

### DIFF
--- a/src/server/ws.rs
+++ b/src/server/ws.rs
@@ -120,6 +120,16 @@ pub async fn terminal_ws(
     }
 }
 
+/// Defence-in-depth cap that mirrors the client-side cap in
+/// `web/src/hooks/useTerminal.ts`. The browser's wterm WASM grid is fixed
+/// at 256x256 (vercel-labs/wterm@0.1.x); resizing the PTY past this lets
+/// the agent draw past wterm's render cap, producing visible corruption.
+/// The client should never send a larger resize, but if a future client
+/// regression slips through, clamp here so the PTY is never larger than
+/// the in-browser grid. Remove when wterm raises or removes the cap.
+const WTERM_MAX_COLS: u16 = 256;
+const WTERM_MAX_ROWS: u16 = 256;
+
 /// Unique client ID counter for primary-client tracking.
 static CLIENT_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
 
@@ -321,6 +331,8 @@ async fn handle_terminal_ws(
                     if let Ok(control) = serde_json::from_str::<ControlMessage>(&text) {
                         match control {
                             ControlMessage::Resize { cols, rows } if cols > 0 && rows > 0 => {
+                                let cols = cols.min(WTERM_MAX_COLS);
+                                let rows = rows.min(WTERM_MAX_ROWS);
                                 pending_size = Some((cols, rows));
 
                                 let dominated = is_primary_or_vacant(
--- a/src/server/ws.rs
+++ b/src/server/ws.rs
@@ -120,7 +120,7 @@ pub async fn terminal_ws(
     }
 }
 
-/// Defence-in-depth cap that mirrors the client-side cap in
+/// Defense-in-depth cap that mirrors the client-side cap in
 /// `web/src/hooks/useTerminal.ts`. The browser's wterm WASM grid is fixed
 /// at 256x256 (vercel-labs/wterm@0.1.x); resizing the PTY past this lets
 /// the agent draw past wterm's render cap, producing visible corruption.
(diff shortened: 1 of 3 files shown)


njbrake:
> _Posted by Claude (Anthropic) through a debug session with @njbrake._

Closing without merging.

The end-to-end test we ran against real `aoe serve` + tmux + claude/codex didn't reproduce any corruption attributable to wterm's 256×256 grid cap. The visible bug in #830 / #831 was vercel-labs/wterm#49 (DEC alternate charset), which is now addressed by the merged PR #837. The cap workaround in this PR remains correct preventive code for vercel-labs/wterm#56, but it's defending against a class of bug we haven't actually observed in the wild — and it adds ~30 lines of state-machine logic to `useTerminal.ts` that future maintainers would need to understand on every change to that file.

Branch `fix/wterm-grid-cap-workaround` is preserved on origin. Two scenarios would justify reopening / re-implementing:

1. A user reports rendering corruption that specifically tracks the 256-col / 256-row symptom (overlapping rows displaced by exactly the amount of "missing" cols on a wide monitor).
2. vercel-labs/wterm#56 stalls upstream past a few weeks of silence and we have evidence wterm isn't going to raise the cap soon.

Until either happens, the simpler default — "trust upstream, ship workarounds only for observed bugs" — wins. The investigation context is captured in the comment thread above.

PR #835, opened 2026-04-28
feat: in-app self-update with aoe update and a TUI hotkey

### LINKED ISSUE #834: Add a way to update aoe from inside the app
Hey! When the TUI shows the yellow "update available" bar at the bottom, there's no way to actually install the update from inside the app. The CLI notice on launch points me at `brew update && brew upgrade aoe`, but I installed via the curl script in the README, so that's not the right command for my install. Nix users would hit the same issue from the other direction.

Would it make sense to add an `aoe update` command that:

- detects how aoe was installed (Homebrew, curl-script tarball, Nix, cargo)
- dispatches to the right upgrade flow for that method (delegate to `brew upgrade aoe` for brew, download + replace for the tarball case, refuse with helpful instructions for Nix and cargo)
- mirrors the same flow as a hotkey in the TUI when the update bar is up, so the bar text can change to "press `u` to update"

Happy to take a stab at it if you think it's worth building. Open to whatever approach you'd prefer.

weedgrease:
## Description

When the TUI shows the "update available" bar, there's no way to actually run the update from inside the app, and the CLI notice on launch hardcodes `brew update && brew upgrade aoe` — wrong for users who installed via the curl script, Nix, or cargo. This PR adds an `aoe update` CLI command and a `u` hotkey in the TUI that detect the install method and dispatch to the right upgrade flow.

It also flips `scripts/install.sh` to install in a user-writable directory by default so neither install nor `aoe update` need sudo on the dominant path.

### What's new

- **`aoe update` CLI command** with `-y/--yes`, `--check`, and `--dry-run` flags. Force-fresh GitHub release check, four-line confirm prompt (version / method / location / sudo), streaming download with byte-percent progress, runs the candidate's `--version` to sanity-check the new binary, atomic `rename(2)` with `sudo` fallback when the install path needs elevation.
- **`u` hotkey in the TUI** (only enabled when `update_info.available`). Opens a confirm dialog mirroring the CLI prompt block; on yes, the TUI suspends the alternate screen via the existing `App::with_raw_mode_disabled` so `sudo`'s password prompt and `brew`'s progress work, then restores cleanly. Pressing `u` on a Nix/cargo/unknown install flashes a transient status with the manual upgrade command instead of being a silent no-op.
- **`Ctrl+x` dismiss** for the update bar (and the persistent "update complete" toast). Session-only — the flag is in-memory, so quitting and relaunching brings the bar back if the update still applies. Bar reads `[u] update  [Ctrl+x] dismiss`.
- **Refusal paths** print the right command for each install method: `nix run github:njbrake/agent-of-empires` for Nix, `cargo install --git ...` for cargo, the curl install script for unknown locations.
- **CLI notice text** changed from `(run: brew update && brew upgrade aoe)` to `(run: aoe update)` so it's right regardless of how aoe was installed.

### `install.sh` changes

- **Default `INSTALL_DIR` is now `$HOME/.local/bin`** (was `/usr/local/bin`). Sudo only fires if the user explicitly sets a write-protected `INSTALL_DIR=...`. Lands AOE in line with rustup/bun/deno conventions.
- **PATH guidance**: when the install dir isn't on `$PATH`, prints the exact `echo 'export PATH=...' >> $rc_file` for the user's detected shell (zsh / bash / fish).
- **Legacy install warning**: if `/usr/local/bin/aoe` exists when running install.sh, prints a one-line warn with the exact `sudo rm` to clean it up. Otherwise users running install.sh after the default change end up with two binaries and PATH order picks the winner.

### How install detection works

A pure path-prefix classifier identifies the method from the canonicalized `current_exe()` path: `/nix/store/...` → Nix, `~/.cargo/bin/...` → Cargo, parent in `/usr/local/bin` / `~/.local/bin` / `~/bin` → Tarball, otherwise Unknown. **Both the exe path and the home dir are canonicalized** before comparison — without canonicalizing the home dir, a binary at `/tmp/.../.local/bin/aoe` on macOS gets a canonicalized exe path of `/private/tmp/.../.local/bin/aoe` (since `/tmp` is a symlink) but home stays `/tmp/...`, and the parent-prefix comparison fails and a perfectly fine tarball install misclassifies as Unknown.

Homebrew is layered on top: `brew list aoe` must succeed AND its reported path must canonicalize to the same file as the running binary, so a brew-installed-but-running-from-elsewhere binary doesn't get classified as brew. The `brew list aoe` probe is bounded by a hard 2s timeout so a hung brew (locked formula DB, network-bound auto-update) can't stall TUI startup.

### Tarball update flow

Mirrors `scripts/install.sh` step-for-step but updates the binary at its *current* location instead of `$INSTALL_DIR`:

1. Resolve the platform string (`linux-amd64` / `linux-arm64` / `darwin-amd64` / `darwin-arm64`).
2. Create a `TempDir` *inside* the binary's parent directory so the rename in step 5 stays on the same filesystem.
3. Download `https://github.com/njbrake/agent-of-empires/releases/download/v{version}/aoe-{platform}.tar.gz` into that tempdir, with byte-percent progress to stderr (CLI) or the dialog status line (TUI).
4. Shell out to `tar xzf`, then run the extracted binary's `--version` and require a whole-token match against the expected version (so `0.5.10` doesn't satisfy `expected = 0.5.1`).
5. `std::fs::rename(extracted, binary_path)` — atomic on POSIX, and the kernel keeps the old inode alive for the running process so the TUI can safely overwrite its own binary. On `EACCES`, falls back to two sequential `sudo` invocations (`mv` then `chmod 0755`) using sudo's timestamp cache to keep it to one password prompt.
6. The user is told to restart `aoe` after success; we don't auto-kill the TUI since they may have unsaved state in agent sessions.

`parent_is_writable` (used to set the "Sudo: required" prompt line and to gate the suspend-vs-background dispatch) creates a uniquely-named probe file via `tempfile::Builder::tempfile_in(parent)`. A process killed mid-probe leaves no stale file behind that could poison subsequent calls.

### TUI dispatch notes

The non-sudo tarball path runs in the background via `std::thread::spawn` + `Handle::block_on` (rather than `tokio::spawn`) because `perform_update`'s future is `!Send` — its `Option<&mut dyn FnMut(...)>` callback parameter has no `Send` bound on the trait object. The new `update_status` is a `(text, expires_at)` pair so failure messages auto-clear after 10 s while success messages persist (they tell the user to restart, which they should keep seeing until they do).

### Documentation

README gets a one-line note pointing users at `aoe update`. `docs/cli/reference.md` was regenerated via `cargo xtask gen-docs`.

### Out of scope (deliberate)

- **Silent auto-apply on TUI startup.** A previous iteration of this branch added an `AutoupdatePolicy { Off, Notify, Patch, All }` enum that auto-applied patch releases for users who opted in. It was reverted — the new attack surface (silent download + atomic-replace without an "I'm choosing this" moment) wasn't worth it for a feature no user asked for, especially without SHA256 or signature verification on the tarball. The manual `u` hotkey + `aoe update` CLI gives users a single-keystroke update path. We can revisit if there's user demand once supply-chain hardening is in place.
- Pinning to a specific version (`aoe update --version X.Y.Z`).
- SHA256 / signature verification of downloaded tarballs (matches `scripts/install.sh`'s current trust-HTTPS posture).
- Auto-restarting the TUI after success.
- Channels (stable / beta / nightly).

Closes #834

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Testing

- `cargo build`: clean
- `cargo test --lib`: 1153 passing
  - Added hermetic shim-based tests for the previously-untested branches: `sudo_replace_moves_then_chmods` + `sudo_replace_propagates_failure_from_mv` (PATH-shimmed sudo that exec's its args), `runs_update_then_upgrade_aoe` + `brew_update_failure_aborts_before_upgrade` + `brew_upgrade_failure_is_reported` (PATH-shimmed brew that records argv), `probe_returns_none_when_brew_hangs` (sleep-shim verifies the 2s timeout). Plus `classifier_requires_consistent_canonicalization` and `detects_tarball_through_symlinked_home` for the macOS canonicalize fix, and `test_legacy_auto_update_field_is_silently_ignored` for the schema-compat regression.
- `cargo test --test update_command`: 2 passing (CLI integration; HOME-isolated via `tempfile`)
- `cargo test --test e2e update`: 1 passing, 1 ignored (Homebrew shim test gated on a real GitHub-published update being available)
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean

**Manual end-to-end on Linux** (driven via tmux on a private socket so the test couldn't touch the user's real aoe sessions):
- `install.sh` against the real GitHub release, fresh `$HOME` → lands at `~/.local/bin`, PATH guidance fires correctly
- `install.sh` with `/usr/local/bin/aoe` already present → legacy warn fires with correct `sudo rm` line
- `aoe update --check` from a 1.4.5 binary → correctly reports `available: true`
- `aoe update --dry-run` from `~/.local/bin/aoe` → shows correct prompt block, classifies as Tarball, `Sudo: not required`
- TUI 1.4.5 → 1.4.6 via `u` hotkey: bar → confirm dialog with right Method/Location → press `y` → "update complete" toast → binary on disk replaced
- `Ctrl+x` dismiss: bar visible → press `Ctrl+x` → bar gone in <1 frame → stays gone for the session → returns on relaunch

**Manual end-to-end on macOS** (initial run hit a `/tmp` → `/private/tmp` symlink classification bug that was fixed in the canonicalize commit; subsequent runs work):
- TUI flow on `~/.local/bin` install path
- Settings TUI panel renders the Updates section without errors

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Code (Opus 4.7, with Sonnet 4.6 sub-agents for individual implementation tasks).

**Any Additional AI Details you'd like to share:** Brainstormed the spec, drafted the implementation plan, then implemented under my review with subagent-driven task execution and a two-stage spec-and-quality review per task. A final whole-branch audit by parallel reviewers caught a strict-mode hotkey bug (help advertised `U`, dispatch matched `u`), a non-hermetic test (was writing to real `~/.config/agent-of-empires/`), and a few style outliers vs the rest of the repo — all fixed before submission. Maintainer review pass added the install.sh dir change, hardening (brew timeout, tempfile probe, sudo + brew hermetic tests), the macOS canonicalize fix caught during manual testing, and the Ctrl+x dismiss; an autoupdate policy was prototyped then reverted on supply-chain grounds.


weedgrease:
i should have accounted for the rate limit in specs 😭 my bad! thanks for running with this!

PR #836, opened 2026-04-28
fix(web): buffer DEC ?2026 BSU/ESU updates as wterm workaround

njbrake:
## Description

Workaround for vercel-labs/wterm#57. wterm silently ignores DEC private mode 2026 (synchronized output, BSU/ESU): writes land in wterm's renderer as soon as its internal rAF fires. Modern TUIs (codex, claude, neovim, btop) use this mode to bracket atomic frame updates so a single frame paints in one go. When the bracketed content spans multiple PTY reads — large frame, slow link, scroll-paint — each WebSocket message lands in a separate JS task and triggers its own rAF render, and the user sees half-drawn frames.

Parse the BSU/ESU markers in `useTerminal.ts` and hold writes between them so wterm receives one `term.write` per frame. Force-flush after 150ms (per the spec's recommended cap) if a host opens BSU and never closes it. Bytes outside any BSU pass through unchanged, so agents that never use the mode incur no behavior change.

This is the second of three workarounds for wterm gaps surfaced while investigating #830 / #831:
- #832 (already merged-or-pending) caps the PTY at wterm's 256x256 grid (workaround for vercel-labs/wterm#56).
- This PR addresses synchronized output (workaround for vercel-labs/wterm#57).
- vercel-labs/wterm#54 (wide chars) and #55 (mouse + focus events) remain upstream-only — no host-side workaround possible.

### What changed

- `web/src/hooks/useTerminal.ts`: adds `BSU_BYTES` / `ESU_BYTES` / `SYNC_FLUSH_TIMEOUT_MS` constants, an `indexOfBytes` helper, and a per-WebSocket-connection `writeWithSyncBuffer` state machine. Wired into the binary path of `ws.onmessage`. String messages bypass the buffer (the server only sends UTF-8 for control messages; PTY bytes always arrive as ArrayBuffer). Timer is cleared on effect teardown.
- `web/tests/helpers/terminal-mocks.ts`: extends `MockHandle` with a `push()` function so tests can inject arbitrary PTY bytes from the WS mock. The route handler assigns it on each connection (last-wins if multiple sockets are opened, which is fine for these tests).
- `web/tests/wterm-sync-output.spec.ts`: three new regressions —
  1. A frame split across BSU…ESU is held until ESU and rendered atomically.
  2. BSU without ESU flushes after the 150ms safety timeout.
  3. Bytes outside any BSU pass through immediately.

### What this fixes / doesn't fix

- **Fixes:** mid-frame partial-paint artifacts when a TUI brackets its frame redraws with `?2026h` / `?2026l` and the bracketed content spans multiple PTY reads. Codex emits ~25 of these pairs in its first 8s of startup output (measured via `script(1)` against a real PTY). Claude uses a related variant.
- **Does not fix:** the wide-char misalignment (vercel-labs/wterm#54), missing mouse / focus events (vercel-labs/wterm#55), or any other rendering issues attributable to wterm gaps that don't go through the synchronized-output mode.

The contribution of this gap to the specific symptoms in #830/#831 is unproven without a side-by-side render comparison; this PR removes one likely contributor.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Manual checks run locally:
- `npm run build` ✓
- `tsc -b` ✓
- `eslint` on changed files ✓
- `npx playwright test tests/wterm-sync-output.spec.ts` (3/3 pass) ✓
- Full Playwright suite ✓ (one flaky Chromium SIGSEGV on retry, unrelated)

(No UI screenshot included: this is a buffering change with no new UI surface. The Playwright tests are the verification.)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Anthropic), via Claude Code, in a debug session with @njbrake.

The synchronized-output gap was identified by capturing real PTY byte streams from `codex`, `claude`, and `cursor-agent` with `script(1)` and analyzing them for the patterns wterm doesn't handle. The 150ms safety-flush timeout matches the recommendation in the iTerm2 synchronized-updates spec.

Refs vercel-labs/wterm#57.

### DIFF
--- a/web/src/hooks/useTerminal.ts
+++ b/web/src/hooks/useTerminal.ts
@@ -32,6 +32,45 @@ const RESIZE_DEBOUNCE_MS = 50;
 // a small margin. After the first resize lands the debounce drops to
 // RESIZE_DEBOUNCE_MS so live splitter drags still feel responsive.
 const INITIAL_SETTLE_MS = 250;
+// DEC private mode 2026 (synchronized output, BSU/ESU). wterm doesn't
+// implement it: setPrivateMode falls through `else => {}`, and writes are
+// rendered as soon as wterm's internal rAF fires. Modern TUIs (codex,
+// claude, neovim, btop) bracket frame updates with `\x1b[?2026h` ...
+// `\x1b[?2026l` so a single frame paints atomically. When the bracketed
+// content spans multiple PTY reads (large frame, slow link, scroll-paint),
+// each WebSocket message lands in a separate JS task and triggers its own
+// rAF render — the user sees half-drawn frames.
+//
+// Workaround: parse incoming bytes for the BSU/ESU markers ourselves and
+// hold writes between them so wterm sees one term.write per frame. Filed
+// upstream at vercel-labs/wterm#57; remove when wterm ships native sync
+// support. Spec: gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec
+const BSU_BYTES = new TextEncoder().encode("\x1b[?2026h");
+const ESU_BYTES = new TextEncoder().encode("\x1b[?2026l");
+// Spec recommends ~150ms cap. If the agent opens a BSU and never closes
+// it (crash, missed flush), we force-flush so the terminal doesn't appear
+// frozen. 150ms is well above one-frame budgets on real agents.
+const SYNC_FLUSH_TIMEOUT_MS = 150;
+
+/**
+ * Find the first occurrence of `needle` in `haystack` starting at `start`.
+ * Returns -1 if not found. Plain bytewise scan; fine for the 8-byte BSU/ESU
+ * markers we look for.
+ */
+function indexOfBytes(
+  haystack: Uint8Array,
+  needle: Uint8Array,
+  start: number,
+): number {
+  if (needle.length === 0 || haystack.length - start < needle.length) return -1;
+  outer: for (let i = start; i <= haystack.length - needle.length; i++) {
+    for (let j = 0; j < needle.length; j++) {
+      if (haystack[i + j] !== needle[j]) continue outer;
+    }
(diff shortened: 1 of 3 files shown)


njbrake:
> _Posted by Claude (Anthropic) through a debug session with @njbrake._

Closing this PR. After the change was on the branch, I installed tmux in the test sandbox and ran a real `aoe serve` end-to-end test with codex and claude through the web at 1920×1080 and 3000×1200. **The `?2026` (BSU/ESU) markers this PR was designed to handle never appear on the wire.** `grep -c "2026" /tmp/aoe-test/out/console.log` → 0 in both sessions.

The reason: my earlier byte-stream capture (which counted 25 BSU/ESU pairs in codex's output) was running codex **directly** in a PTY with `TERM=xterm-256color`. The aoe path runs codex inside tmux, which exposes itself as `screen-256color` to inner programs, and that terminfo doesn't claim synchronized-output support. Codex (and claude) check the terminfo and skip emitting `?2026` entirely when running under tmux. So the BSU/ESU buffer added by this PR never engages in the realistic aoe deployment.

The code is defensively correct — it would help any user who has tmux configured with `terminal-overrides ',xterm-256color:Sync=...'` to expose synchronized-output capability through tmux — but that's a rare power-user knob and adding 117 lines of code + tests for a path that doesn't fire isn't worth the maintenance cost.

Upstream vercel-labs/wterm#57 stays open; if/when wterm ships native sync support, no host-side change is needed. If we ever see real evidence of `?2026` reaching the browser through tmux (e.g., via different agent versions, different tmux configs, or aoe changing how it sets `TERM`), reopening or re-implementing this is straightforward — branch is preserved.

Closing without merging.

PR #864, opened 2026-04-29
fix(session): eliminate TOCTOU in poller tmux session name capture

jerome-benoit:
## Description

Eliminates a time-of-check/time-of-use gap in `maybe_start_poller()` where `tmux_session()` was called twice: once to initialize the `SessionPoller` and again ~100 lines later for the `on_change` callback. If the tmux session disappeared between the two calls, the poller would hold the real name while the callback got an empty string, causing `publish_session_to_tmux_env` to silently no-op.

Now the session name is captured once, cloned for the callback, and moved into the poller (also eliminating one redundant String allocation).

Extracted from #382.

## PR Type

- [x] Bug Fix

## Checklist

## AI Usage

- [ ] No AI was used

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1079,7 +1079,8 @@ impl Instance {
             .tmux_session()
             .map(|s| s.name().to_string())
             .unwrap_or_default();
-        let mut poller = SessionPoller::new(tmux_session_name.clone());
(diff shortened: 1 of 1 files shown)


jerome-benoit:
The fix is not correct the discrepancy between the naming at runtime is intended.

PR #868, opened 2026-04-30
feat(cockpit): native ACP rendering surface (Beta) for all supported agents

njbrake:
## Description

Adds **cockpit**, an ACP-based structured rendering surface that runs alongside the existing tmux passthrough. Every aoe session is now a per-session pick: `tmux` (legacy, raw bytes through wterm) or `cockpit` (Beta, agent speaks Agent Client Protocol; aoe renders typed events as React cards). Tmux remains the default — cockpit is opt-in via `aoe add --cockpit` or the new substrate picker on the web wizard.

The data model (`cockpit_mode: bool` per session) is already merged on `main`; this branch is the polish + ecosystem expansion that turns cockpit into something you'd actually want to use, plus per-tool ACP support so it's not Claude-only.

## What's in here

**Core substrate (was already on the branch from earlier merges):**
- Per-session ACP supervisor (`src/cockpit/supervisor.rs`) with restart budget, drain task, fs/terminal handlers
- Replay-buffered WebSocket fanout (`src/server/cockpit_ws.rs`)
- React surface built on `@assistant-ui/react` primitives
- Sandbox support via unix-socket transport

**Per-tool ACP (`e4ad824`):** verified each agent's ACP invocation against [agentclientprotocol.com/get-started/agents.md](https://agentclientprotocol.com/get-started/agents.md) and seeded the registry:

| Tool | Path |
|---|---|
| `claude` | `claude-agent-acp` (Zed adapter) |
| `opencode` | `opencode acp` (native, SST) |
| `gemini` | `gemini --acp` (native, Google) |
| `codex` | `codex-acp` (Zed adapter) |
| `vibe` | `vibe-acp` (native, Mistral) |
| `pi` | `pi-acp` (Hermes coding agent) |
| `aoe-agent` | bundled multi-provider fallback |
| *aider, cursor, copilot, droid, settl, hermes* | greyed out — terminal-only |

`Supervisor::pick_agent_for_tool(tool, override)` replaces three copy-pasted `"claude → claude-code, else aoe-agent"` fallbacks.

**UX polish (`56a8822` → `dd249aa`):**
- Composer rebuilt VSCode/Cursor-style — multi-line, lucide icons, focus glow, paper-plane Send / square Stop
- Markdown rendering with shiki code blocks + smooth streaming (`@assistant-ui/react-markdown`)
- Per-kind tool cards: bash / read / edit / search / fetch / think with proper input parsing
- `react-diff-viewer-continued` for edit cards
- `@`-mention file picker (assistant-ui's `Unstable_TriggerPopover` + workspace file index endpoint)
- `/` slash commands
- Empire-themed working spinner ("Conscripting villagers", braille rattle)
- Hover affordances (copy/edit/regenerate via `ActionBarPrimitive`)
- Approval cards realigned with tool-card visual language
- Mode picker: real ACP-advertised modes from `NewSessionResponse.modes`, drop-up menu in composer footer

**Reliability fixes:**
- `3cccf46` — drain-task/send_prompt deadlock (drain held client mutex across `recv().await`)
- `28e8066` + `30a21f8` — TUI no longer marks cockpit sessions as errored ("tmux pane is gone")
- `d244ad6` — composer wins focus race against wterm's async init
- `c1bb7e0` — auth env forwarded by default (ANTHROPIC_API_KEY, CLAUDE_CONFIG_DIR, etc.)

**Library refactor (`f1298bc`):** replaced ~520 lines of reinvention with first-party assistant-ui primitives (TriggerPopover, MarkdownTextPrimitive, smooth streaming).

**Doctor + docs:**
- `aoe cockpit doctor` walks the full registry, prints per-agent install hints, `--fix` `npm install -g`s the npm-distributed adapters
- `docs/cockpit.md` gets a Beta callout, per-agent support/auth table, and updated doctor sample

## PR Type

- [x] New Feature

## Checklist

- [x] New and existing tests pass (5/5 cockpit_acp_smoke, 13/13 e2e Playwright steps)
- [x] Documentation was updated (`docs/cockpit.md`)
- [x] For UI changes: included screenshot or recording (see screenshots in commits)

## AI Usage

The branch was developed iteratively across many conversations. Architecture decisions (ACP as substrate B, per-session toggle, supervisor ownership of agent processes, drain-task pattern, assistant-ui for the React surface) were human-directed; the AI handled implementation, testing, and UX iteration. Notable AI-caught issues that humans would have caught later: the deadlock in `Supervisor`'s drain task, the focus race against wterm's async init, the missing title fallback for tools with empty `raw_input`. Notable AI-missed issues that humans caught: the mode picker initially said "Default" when the agent was actually in yolo mode (we weren't reading agent-advertised modes), the bash card showed `$ {}` for the same reason, and the early version reinvented `Unstable_TriggerPopover` + `MarkdownTextPrimitive` instead of using the assistant-ui primitives that were already in our deps.

## How to build & run

```sh
cargo build --features serve --profile dev-release
./target/dev-release/aoe add . --cmd claude --cockpit   # one cockpit session
./target/dev-release/aoe serve                          # web dashboard
```

> **Note:** the `cockpit` cargo feature was folded into `serve` (commit
> `ec5cfb0`). If you saw earlier instructions saying `--features 'serve cockpit'`,
> just use `--features serve` now. Cockpit ships alongside the dashboard.

## How tested

```
cargo test --features serve --test cockpit_acp_smoke   # 5/5 pass
cargo build --features serve --profile dev-release
cargo clippy --features serve -- -D warnings   # clean
```

Plus a Playwright e2e harness against a live `aoe serve` with a Node ACP test shim:
- 13 steps: session create → cockpit composer mounts → user prompt → agent text streams → tool call card renders → final "done" → REQUEST_PERMISSION flow → Allow click → permission_outcome=yes
- Verified the `@` file picker + `/` slash command popovers
- Verified the working spinner with verb cycling
- Verified focus reclaim after wterm async init

## Caveats

- The five non-Claude agents (opencode/gemini/codex/vibe/pi) were verified at the documentation level (matching `agentclientprotocol.com/agents.md` against upstream docs) and via the supervisor/registry tests, but I didn't exercise each adapter end-to-end on real hardware in this branch — the smoke and e2e tests use the shim. First time you `aoe add . --cmd opencode --cockpit` you may hit per-agent quirks.
- `aoe cockpit doctor` only checks binary presence, not auth state. A future improvement would spawn each adapter's `initialize` and inspect the response's `auth_methods`.
- One `Unstable_*` primitive from assistant-ui (`Unstable_TriggerPopover`); if upstream renames it on a minor bump we'd need a mechanical update.

## Test plan

- [ ] `claude /login` then `aoe add . --cmd claude --cockpit` — verify cockpit conversation works end-to-end with a real Claude subscription
- [ ] `aoe add . --cmd opencode --cockpit` — verify the registry expansion picks up `opencode acp` correctly
- [ ] `aoe cockpit doctor --fix` — verify it installs the missing adapters
- [ ] Web wizard — verify the substrate picker greys out for tools we know don't have ACP
- [ ] TUI — verify no spurious "tmux pane is gone" errors on cockpit sessions


Seluj78:
do you need some testing on this?

njbrake:
@Seluj78 If you would have a chance, absolutely! This is the next generation feature I think to make the web mode of AOE great, but I haven't had the time to finish the feature.

Seluj78:
oooh fancy ! I like the new UI!

Screenshot 2026-05-07 at 11 18 15

Commands to run to get it working:

```sh
cargo build --features 'serve cockpit' --profile dev-release
cd web && npm install && npm run build && cd ..
./target/dev-release/aoe add . --cmd claude --cockpit
./target/dev-release/aoe serve
```

Here are my feedbacks so far:

- I was not expecting this, but now that I see the new central window, I feel like it's weird to have the terminal on the right side to still be a `tmux`.
- These kind of interfaces are amazing but I am always reluctant in using them compared to claude's CLI because it might be missing features, like the AskUserQuestion tool for example, or other tools that might be added by the agents AOE supports that then need a downstream update to support them afterwards

I asked a suggested question and the right side isn't spinning
Screenshot 2026-05-07 at 11 22 22

After a bit more investigation, it looks like the session never starts
Screenshot 2026-05-07 at 11 23 27

njbrake:
> These kind of interfaces are amazing but I am always reluctant in using them compared to claude's CLI because it might be missing features, like the AskUserQuestion tool for example, or other tools that might be added by the agents AOE supports that then need a downstream update to support them afterwards

💯 this is my concern too. My hope 🤞  was that by using ACP we would have to worry about less of this and things would just work https://github.com/agentclientprotocol/agent-client-protocol .

Thank you for giving it a try. One of the biggest hurdles I'm trying to figure out is how to manage a session between TUI and Web in cockpit mode. Like, in the web dashboard I certainly want it using cockpit mode, but then if I go to use the TUI I think I would rather have it use the tmux+native Claude Code etc mode. So is there a way where I can run a single session and have it be able to switch between cockpit and tmux views. Idk 🤷

Seluj78:
> 💯 this is my concern too. My hope 🤞 was that by using ACP we would have to worry about less of this and things would just work [agentclientprotocol/agent-client-protocol](https://github.com/agentclientprotocol/agent-client-protocol)

Oooh that's what this is ! I didn't bother checking it out earlier. It makes so much sense! Then yeah 100% this is the right way to go, my concern has disappeared. The only thing I would keep in mind is a way to know when new versions of the SDK is released so we know that we need to update AOE to support it (a CI check?)

> Thank you for giving it a try. One of the biggest hurdles I'm trying to figure out is how to manage a session between TUI and Web in cockpit mode. Like, in the web dashboard I certainly want it using cockpit mode, but then if I go to use the TUI I think I would rather have it use the tmux+native Claude Code etc mode. So is there a way where I can run a single session and have it be able to switch between cockpit and tmux views. Idk 🤷

From what I saw in the webUI (I didn't try the TUI in this PR) you had a toggle to switch back to tmux. So I don't understand right now the problem you're having :)

@njbrake could you update this PR (43 commits behind) and I'll try to do some sessions on it and see how it feels and give more feedbacks :)

Seluj78:
@njbrake is there specific things you need tested in this PR ? Or just general feedback on the experience of using it? Or something else ? :)

njbrake:
@Seluj78 yes at this point just general feedback. There are a few dimensions we need to test/debug before merge.

1. How does the web app behave? Is it a smooht experience and can you easily view both cockpit sessions and non-cockpit sessions?
2. How does the TUI handle the cockpit sessions and can you cleanly convert a session between cockpit and tmux modes
3. What is the risk? Is there any impact to the TUI? The TUI is stable and needs to stay that way, the Web App is beta/experimental so I'm less worried about lil rough edges on the web app that we can smooth out in follow up prs

Basically I don't need this PR to handle all the edge cases but I need to be sure that the web app is at least generally as usable as it was before the change, and that it doesn't regress anything in the TUI to degrade the TUI experience

Seluj78:
@njbrake okay, well I can't go any further in my testing because I cannot get claude to respond to me in the cockpit mode
but switching back to tmux mode and asking the same question works

Actually I wanted to reproduce the issue above and I can't even get a new session to show up:

Screenshot 2026-05-07 at 18 04 52

I ran those commands and it said `session started` but nothing happened in the webui.

Seluj78:
> _Comment from Claude (Opus 4.7), posted by @Seluj78 — investigating the "session started but nothing in the webui" report._

Traced the flow. Two-tier bug:

**1. `aoe session start` is a no-op for cockpit sessions.**

`Instance::start_with_size_opts` (`src/session/instance.rs:797-804`) returns `Ok(())` immediately when `cockpit_mode` is true:

```rust
#[cfg(feature = "serve")]
if self.cockpit_mode {
    return Ok(());
}
```

The CLI prints `✓ Started session: ...` regardless, so it looks like it worked.

**2. Cockpit workers only auto-spawn at `aoe serve` startup or via REST.**

ACP worker spawn happens in three places:
- `aoe serve` startup scan (`src/server/mod.rs:471-511`) — runs once, over sessions that exist *at that moment*.
- `POST /api/sessions` (web-UI create) — `src/server/api/sessions.rs:745-767`.
- `POST /api/cockpit/sessions/:id/enable` (substrate switch) — `src/server/api/cockpit.rs:344-356`.

Nothing watches for cockpit sessions added *after* `aoe serve` is already running. The 2 s `status_poll_loop` (`src/server/mod.rs:1218`) reloads instances from disk, so the new session does appear in `GET /api/sessions`, but no code path calls `cockpit_supervisor.spawn(...)` for it.

**Repro of the exact flow in the report:**

```
Term A: aoe serve                              # daemon up, 0 cockpit sessions known
Term B: aoe add . --cmd claude --cockpit       # written to disk only
Term B: aoe session start Ethiopians           # no-op (cockpit early return)
                                               # 2s later: poll reloads disk, session
                                               # appears in webui list, but no worker
                                               # was ever spawned → agent silent.
```

This matches both screenshots in the comment above: "claude doesn't respond" (worker missing for an existing cockpit session) and "session never starts" (worker missing for the freshly-added one).

**Fix options:**

1. **Daemon-side reconciler (recommended).** Extend `status_poll_loop` to track attempted spawns and, on each tick, call `supervisor.spawn` for any cockpit session on disk that has no running worker and hasn't been attempted yet. Idempotent, mirrors the startup auto-spawn, fixes both `aoe add --cockpit` while serve is running and any race where serve starts before a session is fully written. Need a "already attempted" set so a permanently-failing spawn doesn't retry every 2 s — supervisor already has restart bookkeeping for in-process crashes, but the initial spawn is currently a one-shot.
2. **CLI → daemon IPC.** `aoe session start` for cockpit POSTs to the running daemon. Heavier (needs daemon discovery + auth token plumbing) and only fixes that one entry point — `aoe add --cockpit` while serve is already up still wouldn't trigger a spawn.
3. **Loud CLI error.** `aoe session start` on cockpit prints "cockpit sessions are managed by \`aoe serve\`; restart serve to spawn the worker." Cheap, but the UX is bad and the underlying mismatch stays.

Going with #1 unless you'd rather a different shape — happy to push a fix on top of this branch.

njbrake:
@Seluj78 thanks! I'm good with you pushing, maybe you need to create a branch in your fork to be able to do this though.

Seluj78:
> @Seluj78 thanks! I'm good with you pushing, maybe you need to create a branch in your fork to be able to do this though.

Yes, cause I'm not a maintainer in this repo, I'll make a branch in my fork that targets the `native` branch here :) gimme a few

Seluj78:
@njbrake https://github.com/njbrake/agent-of-empires/pull/953 is ready to review, it fixes a few different bugs. Let me know there if you disagree with some choices!

Seluj78:
(one this PR is merged, or at least in a working state with my PR merged into it I will start using AOE to work on AOE 😉 )

njbrake:
@Seluj78 I am working on some cleanup here. Here's my plan:

Compile the feature as a part of the existing 'serve' flag, but gate the cockpit feature behind an env var `EXPERIMENTAL_COCKPIT`so that most people don't yet use the cockpit feature when they're not expecting it.

This gives us time to slowly improve the cockpit UI/UX until we're ready to to expose it to most users, and also lets us put it into main without having to constantly maintain a separate huge feature branch.

To make sure this PR doesn't hang lets just verify that:

1. This PR leaves current users of the Web and TUI and CLI completely unaffected
2. When cockpit mode is enabled, you can launch and use a cockpit agent in the web ui
3. when a session is running in cockpit mode, the TUI should display that it exists but not allow it to be launched.

I think that's it?

Seluj78:
@njbrake sounds good to me! I'm away for a couple more hours so I can't do more than read and comment, but your plan sounds good to me, it'll allow me to use the cockpit daily as well !

Seluj78:
*AI disclosure: investigation and first-draft prose written by Claude (Opus 4.7). The framing, idea, the scope cut, and this disclosure line are mine (the human opening this).*

@njbrake confirmed locally — env var (`AOE_EXPERIMENTAL_COCKPIT`) is wired through CLI, REST, server-side wizard gate, and TUI. All three asks check out:

1. **Web/TUI/CLI unaffected when unset**: `default_cockpit_for_web()` returns false, CLI `--cockpit` bails with actionable error, REST silently downgrades. Persisted cockpit sessions still auto-spawn (gate is for *new* sessions) with one-time warn log.
2. **Cockpit launchable when enabled**: `/api/about` exposes `experimental_cockpit`, wizard reveals substrate picker (`AgentStep.tsx`).
3. **TUI shows but blocks launch**: ` [web]` badge in Agent view, Enter shows toast `"Cockpit session: open the web dashboard (aoe serve) to attach"`, `app.rs` guards the `--launch` path too.

One discrepancy worth a fix before merge: server `default_cockpit_for_web()` claims browser default is cockpit when env set, but wizard initial state is `cockpitMode: false` and the submit handler always sends an explicit value — so the server default never fires for the web flow. IMO the wizard should flip `cockpitMode` to true after `fetchAbout()` resolves with `experimental_cockpit: true`, matching the server-side intent. WDYT?
-> I think when the env var is set we should ALWAYS use cockpit when creating sessions in the web ui

njbrake:
The web UI is looking..... bad. I am going to merge this in because I think directionally it's headed the right direction and shouldn't impact existing users, but there is certainly a lot of work needed to clean up UI UX for web to be fun to use.

Issue #863, opened 2026-04-29
RFC: Stop hook semantics — should Claude's Stop event mark a session 'waiting' instead of 'idle'?


BTForIT:
## Context

@njbrake asked me to surface this as an issue first rather than a PR (per #855), so we can pick a direction on what `waiting` means before any code lands. Three reasonable options below; I have a preference but I want to lay them out flat.

## The observation

Claude Code's `Stop` hook fires the moment the agent finishes a turn. Today aoe maps that event to:

```
printf idle > $status
```

…which means a session that *just* came back to the user (asking a question, awaiting confirmation, finished a long task) is rendered identically to one that has been quiet for an hour. On a dashboard with 10–20 panes, the just-finished turn is the **most actionable** session in the list, but it visually fades to the same shade as background-quiet ones.

The `w` / `W` "jump to next waiting" keybind from #796 is the consumer that actively cares about this distinction — it's specifically designed to triage user-actionable sessions, and right now it skips over Stop-fired turns that the user genuinely needs to look at.

### Option A — Status quo (`Stop` → `idle`)

Stop is treated as "agent stopped doing things; nothing demands attention." User-actionable cues come exclusively from elicitation hooks (#524) and explicit prompts.

**Pro:** `idle` truly means "no agent activity," symmetric with `PreCompact/PostToolUse` aging.
**Con:** A finished turn looks identical to a session that's been quiet for an hour. `w/W` walks past freshly-finished turns.

### Option B — `Stop` → `waiting`

Treat the human as the next actor at end-of-turn. Just-finished turns surface as "needs attention" until age-based decay or the next hook event moves them.

**Pro:** Matches user mental model — "Claude just answered; my move." `w/W` finds them. Foregrounds the most actionable sessions.
**Con:** `waiting` now means *two* different things: "agent is blocked on a tool/elicitation" AND "agent finished a turn and the human is next." The semantic gets fuzzier.

### Option C — Add a fourth state distinct from `idle`/`waiting`

E.g., `attention` or `turn_complete`. `Stop` writes that; the dashboard renders it visually distinct from both `idle` (quiet) and `waiting` (blocked-mid-turn).

**Pro:** No semantic overload. Different colors / sort priorities possible. `w/W` can choose to include or exclude.
**Con:** New state touches more surface area: status enum, dashboard rendering, sort/filter UI, `w/W` keybind logic, docs. Bigger change.

## My lean

[...]

PR #891, opened 2026-05-03
feat: add session templates support + UTF-8 safe truncate_id fix

swamy18:
## Description

This PR introduces a **session templates system** to enable reusable, pre-configured agent workflows, along with a fix for safe string handling.

### 🚀 Features
- Added `template` CLI commands:
  - `list` – view all templates
  - `create` – create a new template
  - `delete` – remove a template
  - `show` – view template details
- Implemented `SessionTemplate`:
  - Supports predefined command (`cmd`)
  - Supports environment variables (`env_vars`)
- Added default templates:
  - `debugging`
  - `refactoring`
  - `documentation`

### ⚙️ Improvements
- Integrated templates into session creation flow:
  - Apply template configs during `add`
  - Template command and env override defaults
- Improved command resolution priority:
  - `CLI args > template > config`

### 🛠 Fixes
- Fixed `truncate_id` to safely handle UTF-8 strings:
  - Uses character boundaries instead of byte slicing
  - Prevents potential runtime panics

### 📦 Notes
- Templates are stored in `templates.toml` in the app directory
- Fully backward compatible with existing behavior

---

## Checklist

- [ ] I understand the code I am submitting
- [ ] I have tested this change locally
- [ ] I have updated relevant documentation

---

## Summary
This change makes session setup faster and more consistent using reusable templates, while also improving reliability with safer string handling.

### DIFF
--- a/TEST_CHANGE.md
+++ b/TEST_CHANGE.md
@@ -1,0 +1,1 @@
+﻿Sample test file added by automation on 2026-05-03 19:02:37
--- a/TEST_CHANGE.md
+++ b/TEST_CHANGE.md
@@ -1,1 +1,0 @@
-﻿Sample test file added by automation on 2026-05-03 19:02:37
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -48,8 +48,12 @@ pub struct AddArgs {
     extra_repos: Vec<PathBuf>,
 
     /// Run session in Docker sandbox
-    #[arg(short = 's', long)]
+    #[arg(long)]
     sandbox: bool,
+
+    /// Template to use for pre-configured settings
+    #[arg(long)]
+    template: Option<String>,
 
     /// Custom Docker image for sandbox (implies --sandbox)
     #[arg(long = "sandbox-image")]
@@ -88,6 +92,23 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
     }
 
     let config = repo_config::resolve_config_with_repo_or_warn(profile, &path);
+
+    let template_opt = if let Some(ref template_name) = args.template {
+        let templates = crate::session::load_templates()?;
+        crate::session::find_template(&templates, template_name).cloned()
+    } else {
+        None
+    };
+
+    let tool_override = template_opt.as_ref().and_then(|t| t.cmd.clone());
+    let mut env_override = Vec::new();
+    if let Some(template) = &template_opt {
+        if let Some(env_vars) = &template.env_vars {
+            for (k, v) in env_vars {
+                env_override.push(format!("{}={}", k, v));
+            }
+        }
+    }
 
     // Preserve the original project path for hook trust checking.
     // `path` gets reassigned to the worktree/workspace directory below,
@@ -222,7 +243,8 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
         instance.parent_session_id = Some(parent);
     }
 
-    if let Some(cmd) = &args.command {
+    let effective_cmd = args.command.as_ref().or(tool_override.as_ref());
+    if let Some(cmd) = effective_cmd {
         let tool_name = detect_tool(cmd)?;
         // Verify the agent binary is actually on PATH before creating the session
         if let Some(agent_def) = crate::agents::get_agent(&tool_name) {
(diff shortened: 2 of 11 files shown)


swamy18:
@njbrake Would you like this split into smaller PRs (templates / clone / notifications), or is this structure fine?

njbrake (changes requested):

Thanks for putting this together, @swamy18, and for asking up front whether it should be split. To answer that directly: yes, please split, and before you do I want to flag some things that should change the shape of the work.

### Build status

The branch does not compile against `main`. `cargo build` reports 21 errors. The main ones:

* `Instance` has no `env_vars` field, so `instance.env_vars.extend(...)` in `src/cli/add.rs:316` and the two references in `src/cli/clone.rs:48` will not compile.
* `Status` has no `Completed` variant (the variants are `Running`, `Waiting`, `Idle`, `Unknown`, `Stopped`, `Error`, `Starting`, `Deleting`, `Creating`), so `src/session/notifications.rs:12` fails.
* `DEFAULT_TEMPLATES` is declared `const` but every entry calls `"...".to_string()` and one builds a `HashMap` inline. None of that is const-evaluable on stable Rust. This needs to be a function or a `LazyLock`/`OnceLock`, not a `const`.
* `super::add::detect_tool` is private; `src/cli/clone.rs:69` cannot call it.
* `find_template` in `src/session/templates.rs:91` is missing a lifetime, and `Instance::new("", ...)` in `src/cli/clone.rs:44` does not match the current signature.

Please run `cargo build`, `cargo clippy`, `cargo test`, and `cargo xtask gen-docs` locally before opening a PR. CI runs all of these and will block merge regardless.

### Scope feedback per feature

Rather than just asking you to fix the compile errors and resubmit, I want to be honest about which of these features I think should land at all, because I don't want you to spend more time on something that isn't going to merge.

**1. `truncate_id` UTF-8 fix.** Happy to take this as its own small PR. The only current callers pass UUIDs, which are ASCII, so it fixes a theoretical issue rather than an observed bug, but the change is correct and harmless. Open it standalone.

**2. Session templates.** I don't think this is the right shape for the codebase. We already have a profile system (`src/session/profile_config.rs`) and a per-repo config layer (`src/session/repo_config.rs`) that together cover what `SessionTemplate` is trying to provide:

* `default_tool` already does what `template.cmd` does.
* `agent_extra_args` and `custom_agents` already cover preset workflows.
* Sandbox `environment` already covers `env_vars`.
* Profiles cascade and merge; the proposed templates are a flat parallel surface.

Adding templates would create two configuration systems competing for the same job, and per `CLAUDE.md` any new configurable field also has to be wired through the settings TUI (`FieldKey`, `SettingField`, apply functions, override merge), which this PR does not do. If you've hit a real workflow gap that profiles do not solve, please open an issue describing the specific scenario and we can discuss whether the answer is \"extend profiles\" or something new.

**3. `aoe clone`.** I want to design this before any implementation lands. Sessions own a tmux session, often a git worktree (sometimes a multi-repo workspace), and sometimes a sandbox container. \"Clone\" has several plausible meanings: same branch in a new worktree, new branch off the same base, same directory in a new tmux session, with or without a fresh sandbox container. The current implementation copies struct fields without creating a new worktree or tmux session, and ships a `--launch` flag that the body explicitly leaves as a stub. I'd rather not merge a half-implemented command and then redesign it. Please open an issue first describing the intended UX, and we can lock that down before code.

**4. Status change notifications.** This duplicates an existing system that has been getting active investment. Today we have per-instance `notify_on_waiting`, `notify_on_idle`, `notify_on_error` on `Instance` (`src/session/instance.rs:201`), settings TUI rows for the web equivalents (`src/tui/settings/fields.rs:282`), web push notifications, and recent work like #863 and #872 refining the idle pulse and stop hook. The proposed `eprintln!` from inside the TUI render loop would also corrupt the ratatui frame, so it would not be safe to ship even if there were no existing system. I'd prefer not to add a parallel notification path; if there's a specific gap in the current notifications you want filled, please file an issue.

### Process notes

A few smaller things worth flagging for next time:

* The branch `feature/add-sample-file` and the first commit `feat: add TEST_CHANGE.md` look like leftover scratch work; please rebase those out before opening PRs.
* The PR title and description mention templates and the `truncate_id` fix only, but the diff also adds `clone` and notifications. Description should match contents.
* `src/cli/add.rs` silently drops the `-s` short form for `--sandbox`. That's a user-facing breaking change and per `CLAUDE.md` it needs to be called out explicitly. In this case I'd rather not break it at all.
* `docs/cli/reference.md` is auto-generated by `cargo xtask gen-docs`; new subcommands need that re-run.
* No tests were added for any of the new CLI subcommands. `CLAUDE.md` asks that new CLI subcommands consider an e2e test.

### Suggested next step

Open the `truncate_id` change as its own small PR and I'll merge it. For clone and any templates redesign, open issues first so we can align on scope before you write more code. Thanks again for the contribution and for being open to feedback on direction.

njbrake:
Let me know if you have any questions or disagree with any of the analysis I had Claude do for this 🙏

PR #905, opened 2026-05-05
fix(session): read opencode session list from SQLite, not subprocess

### LINKED ISSUE #904: Memory leak that seems to be specific to opencode in aoe
When running opencode in aoe in non-sandbox mode, the following files accumulate in /tmp/ over time: 

```
sudo du -sh /tmp/.[!.]* 2>/dev/null
4.5M    /tmp/.f9ef7ff75feb77fe-00000000.so
4.5M    /tmp/.f9ef7ff77feff7ff-00000000.so
4.5M    /tmp/.f9ef7ff7dffb37fe-00000000.so
4.5M    /tmp/.f9ef7ff7dfffa7be-00000000.so
```

This issue does not happen when running opencode directly or manually in tmux, when running other coding agents (tested Claude and Pi) or when running opencode in sandbox mode (not checked whether the same tmp files accumulate inside container yet)

An AI analysis suggests looking at the head of the temp file:

```
strings /tmp/.f9ef7fffdbeb4ffe-00000000.so | head -30
@%DC
AOP.#
fZ6
&SN%
4;s@
{8bg:
setLogCallback
setEventCallback
createNativeSpanFeed
__tls_get_addr
getauxval
getArenaAllocatedBytes
getBuildOptions
getAllocatorStats
createRenderer
setTerminalEnvVar
setUseThread
setClearOnShutdown
destroyRenderer
setBackgroundColor
setRenderOffset
resetSplitScrollback
syncSplitScrollback
setPendingSplitFooterTransition
clearPendingSplitFooterTransition
updateStats
updateMemoryStats
getCurrentBuffer
getLastOutputForTest
setHyperlinksCapability
```

Which seems to suggest it relates to some terminal rendering functionality. 

My suspicion is that opencode might be attempting to update and build the terminal backend, but somehow it is not working as intended inside aoe. Not sure why this is occurring non-stop and can accumulate up to hours. 

Observed on aoe 1.5 and 1.51 (latest as of now) 

Deleting the temp files seem to be safe and will release the memory, but as long as an opencode aoe session is running, the files will come back.

njbrake:
> _Note: this PR was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Description

Fixes #904. Memory leak under aoe non-sandbox mode with opencode: ~4.5MB files like `/tmp/.f9ef7ff75feb77fe-00000000.so` accumulate over time. At 60s steady-state polling that's ~270MB/hour per active session, ~2GB in an 8-hour day.

**Root cause** (verified empirically). opencode is built with `bun --compile` and bundles the [OpenTUI](https://github.com/anomalyco/opentui) Zig native renderer as an embedded `bun:ffi` library. Bun extracts that library to `/tmp/.<hash>-NNNNNNNN.so` on every process startup and never unlinks it on exit. Every invocation of every opencode subcommand (`--version`, `--help`, `session list`) leaks exactly one tempfile. This is a known, currently unresolved upstream bug: anomalyco/opencode#6523, anomalyco/opencode#16996, anomalyco/opencode#13479.

**Why aoe and not other agents.** `claude` is also Bun-compiled but ships no embedded FFI native lib, so its subcommands leak zero. opencode is the only aoe-supported tool that embeds a Bun-FFI native lib that the runtime extracts on startup.

**Why aoe and not opencode-by-itself.** aoe's session-resume poller (`src/session/capture.rs:try_capture_opencode_session_id`, `src/session/poller.rs:48-51`) spawns `opencode session list --format json` every poll tick. Steady-state interval is 60s. Each poll spawns a fresh subprocess, each subprocess leaks one ~4.5MB file.

**Why sandbox mode is unaffected.** The sandbox path routes through `docker exec opencode session list`; leaks land in the container's ephemeral `/tmp` and disappear with the container.

**The fix.** Read opencode's local SQLite store at `$XDG_DATA_HOME/opencode/opencode.db` directly. The `session` table exposes `id`, `directory`, `time_updated`, which is exactly what `select_opencode_session` needs. Same row shape (synthesized to match the CLI JSON), same filtering pipeline, no subprocess, no `/tmp` pressure, ~100x faster per poll.

The fallback to the subprocess path is preserved, but only fires on infrastructure errors (DB missing, locked, schema drift); a readable DB with no matching session is authoritative and does not fall back, because the subprocess would just leak for the same empty answer.

Sandbox path (`try_capture_opencode_session_id_in_container`) is intentionally unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## How tested

- 6 new unit tests in `src/session/capture.rs` covering: most-recent match, exclusion set, launch-time floor, missing DB, schema mismatch, and the critical "readable DB with no match must NOT fall back" contract.
- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `cargo test --lib` all pass locally.
- Verified empirically against the real opencode binary that the SQLite path leaks zero `/tmp/.<hash>-*.so` files across 60s of real `SessionPoller` polling and 3 minutes of full `aoe` TUI runtime, where the unfixed subprocess path leaks 1:1 with invocations.

## Tradeoffs

- Adds `rusqlite = { version = "0.32", features = ["bundled"] }`. `bundled` ships its own libsqlite3, no new system dependency. Adds roughly 250KB to the release binary.
- Couples to opencode's current `session` table schema. If opencode renames or drops the columns we read, we degrade to the subprocess path with a one-shot `tracing::warn!` log; we do not panic.

## AI Usage

- [ ] No AI was used

Used Claude Code to investigate the issue, locate the upstream root cause, design the fix, and write the implementation and tests. Schema decisions, fallback semantics (specifically the "no match is authoritative" contract), the choice to leave sandbox mode unchanged, and the bundled-rusqlite tradeoff were all reviewed and approved by me before the code was written.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,12 @@ futures-util = "0.3"
 # Temporary files and directories
 tempfile = "3.14"
 
+# Local SQLite reader for opencode's session store. Bundled to avoid a
+# system libsqlite3 dependency. Used to read session state without spawning
+# `opencode session list`, which leaks a 4.5MB /tmp/.<hash>.so per invocation
+# via opencode's bun:ffi OpenTUI extract path (see anomalyco/opencode#6523).
+rusqlite = { version = "0.32", features = ["bundled"] }
+
 # Web server (optional, behind "serve" feature)
 axum = { version = "0.8", features = ["ws"], optional = true }
 tower-http = { version = "0.6", features = ["fs"], optional = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,6 +74,7 @@ dependencies = [
  "rattles",
  "regex",
  "reqwest",
+ "rusqlite",
  "rust-embed",
  "serde",
  "serde_json",
@@ -92,6 +93,18 @@ dependencies = [
  "tui-input",
  "unicode-width",
  "uuid",
+]
+
+[[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
+dependencies = [
+ "cfg-if",
+ "once_cell",
+ "version_check",
+ "zerocopy",
 ]
 
 [[package]]
@@ -961,6 +974,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
+
+[[package]]
 name = "fancy-regex"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -102,6 +102,12 @@ futures-util = "0.3"
 # Temporary files and directories
 tempfile = "3.14"
 
+# Local SQLite reader for opencode's session store. Bundled to avoid a
+# system libsqlite3 dependency. Used to read session state without spawning
+# `opencode session list`, which leaks a 4.5MB /tmp/.<hash>.so per invocation
+# via opencode's bun:ffi OpenTUI extract path (see anomalyco/opencode#6523).
+rusqlite = { version = "0.32", features = ["bundled"] }
```
jerome-benoit, line 109: hermes can also benefit from it instead of using sqlite3 CLI.

PR #977, opened 2026-05-08
tui: compact tool selector with bidirectional cycle navigation

flowq-C:
## Description

This PR fixes three issues affecting custom agents (defined via `config.session.custom_agents`) in the New Session dialog.

---

### Fix 1: Tool selector overflow (render.rs)

The dialog renders all tools in a single horizontal line. With 5+ built-ins plus custom agents, the line overflows — custom agents become invisible and unreachable.

**Before:**
```
Tool: ● claude  ○ opencode  ○ codex  ○ gemini  ○ copilot  ○ cl-|
                                                            (truncated, unreachable)
```

**After:**
```
Tool: ← ● claude-gemma  [6/10]  →
```

---

### Fix 2: Bidirectional navigation broken (mod.rs)

Both `←` and `→` performed the same action (forward increment). `←` was effectively broken.

Split into two separate handlers: `→` increments, `←` decrements with wrap-around.

---

### Fix 3: YOLO mode ignored for custom agents (instance.rs)

Custom agents are not in the built-in `AGENTS` list, so `get_agent(&self.tool)` returns `None` for them. This caused the YOLO flag to be silently skipped regardless of the YOLO toggle in the dialog — `apply_yolo_mode` was never called.

**Fix:** fall back to `self.detect_as` (already stored on the instance from `agent_detect_as` config) when the primary lookup returns `None`:

```rust
// Before:
let agent = crate::agents::get_agent(&self.tool);

// After:
let agent = crate::agents::get_agent(&self.tool)
    .or_else(|| crate::agents::get_agent(&self.detect_as));
```

This lets custom agents inherit the correct `YoloMode` from their declared `agent_detect_as` agent (e.g. `--dangerously-skip-permissions` for `claude`).

---

## Changes

- `src/tui/dialogs/new_session/render.rs` — compact selector display with counter
- `src/tui/dialogs/new_session/mod.rs` — split Left/Right handler, fix `←` direction
- `src/session/instance.rs` — YOLO mode fallback via `detect_as` for custom agents

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Sonnet 4.6 via Claude Code CLI

**Any Additional AI Details you'd like to share:** All three bugs were discovered while using custom Ollama cloud model aliases with AOE. Fixes were implemented, built locally, and tested before submitting.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -781,7 +781,8 @@ impl Instance {
         let profile = self.effective_profile();
         let on_launch_hooks = self.resolve_on_launch_hooks(skip_on_launch, &profile);
 
-        let agent = crate::agents::get_agent(&self.tool);
+        let agent = crate::agents::get_agent(&self.tool)
+            .or_else(|| crate::agents::get_agent(&self.detect_as));
         self.install_agent_status_hooks(agent);
 
         let cmd = if self.is_sandboxed() {
--- a/src/tui/dialogs/new_session/mod.rs
+++ b/src/tui/dialogs/new_session/mod.rs
@@ -944,7 +944,25 @@ impl NewSessionDialog {
                 }
                 DialogResult::Continue
             }
-            KeyCode::Left | KeyCode::Right if self.focused_field == tool_field => {
+            KeyCode::Left if self.focused_field == tool_field => {
+                self.tool_index = if self.tool_index == 0 {
+                    self.available_tools.len() - 1
+                } else {
+                    self.tool_index - 1
+                };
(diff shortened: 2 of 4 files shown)


Seluj78:
Thanks for the PR ! Can you edit the body to follow the template ?

Otherwise the fix you are proposing sounds good, I am on my phone so I can't test it further right now

PR #981, opened 2026-05-08
fix(serve): WebSocket heartbeat and idle reaper for terminal connections

njbrake:
## Description

Each `aoe serve` WebSocket terminal owns a `tmux attach` child plus its PTY pair (master + cloned reader + writer = ~3 FDs, plus the child's own descriptors). When a client goes silent without sending a Close frame, those resources stay pinned until tmux next emits output, which on an idle session can be hours. The most common ways this happens today:

1. Mobile app gets backgrounded; iOS/Android suspend the WebSocket without closing it cleanly.
2. WiFi vanishes mid-session (subway, elevator) without TCP RST propagating.
3. Browser tab is force-closed.

The accumulated leaked FDs are exactly the pressure that motivated the `RLIMIT_NOFILE` bump in #971. That PR raised the ceiling; this PR removes the leak that fills it.

### What this changes

In `src/server/ws.rs` `handle_terminal_ws`:

1. Send task now also drives a `tokio::time::interval(30s)` and emits `Message::Ping` on each tick. The first immediate tick is consumed before entering the select so we don't ping at connection time. `MissedTickBehavior::Skip` keeps backpressure from producing a burst of pings.
2. Recv task wraps `ws_receiver.next()` in a `tokio::time::timeout(90s)`. On idle timeout it breaks out of the loop, which releases the recv handle and lets the existing cleanup path run (`release_primary`, `pause_exit`, `child.kill()`, `child.wait()`).

axum's `WebSocket` forwards client Pongs back to the recv loop, so live-but-idle sessions stay open indefinitely: each server Ping prompts a client Pong, the Pong arrives in recv, the timeout resets. 90s tolerates two missed round-trips before tear-down.

### Why these constants

- `PING_INTERVAL = 30s`: long enough to be cheap (one tiny frame per direction every 30s); short enough to keep mobile-carrier NAT mappings warm (typical UDP/TCP NAT timeouts on cellular start around 60s).
- `IDLE_TIMEOUT = 90s`: 3 × `PING_INTERVAL`, so a single dropped packet doesn't kill the connection but a genuinely dead client is reaped within ~90s.

### What this doesn't change

- `Message::Ping` payloads are empty; we don't track RTTs or sequence numbers.
- No write-side timeout. A backgrounded mobile client whose TCP receive window has filled will still cause server sends to block; trying to tear that down on a short timeout would punish slow-but-live clients. The recv-side idle timeout is the load-bearing safeguard for FD reclaim.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (N/A, internal behavior)
- [ ] For UI changes: included screenshot or recording (N/A, server-side fix)

Verified locally with:

- `cargo build --features serve` (clean)
- `cargo clippy --features serve --no-deps -- -D warnings` (clean)
- `cargo test --features serve --lib server::ws` (10 ws tests pass)
- Manual: ran `aoe serve --no-auth`, attached a browser, put the tab to sleep without sending Close, watched the connection get reaped after ~90s with the tmux attach child correctly killed.

## AI Usage

- [ ] No AI was used

Drafted with Claude Code after a discussion about the deeper design issue behind #971. Reviewed and verified locally before posting.

### DIFF
--- a/src/server/ws.rs
+++ b/src/server/ws.rs
@@ -7,6 +7,7 @@
 
 use std::io::{Read, Write};
 use std::sync::Arc;
+use std::time::Duration;
 
 use axum::{
     extract::{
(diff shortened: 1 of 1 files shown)


Seluj78 (approved):
Very nice find!

PR #1008, opened 2026-05-09
Cockpit polish: SQLite persistence, session/load resume, tool/markdown rendering, offline state

### LINKED ISSUE #997: Cockpit: use session/load to resume ACP sessions across restart and respawn
*AI disclosure: investigation and first-draft prose written by Claude (Opus 4.7). The framing, idea, the scope cut, and this disclosure line are mine (the human opening this).*

## Problem

Cockpit (ACP) mode currently calls `NewSessionRequest` on every worker spawn — including respawn after a crash and the auto-spawn path that runs after `aoe serve` startup. That means:

- The agent process gets a fresh ACP `session_id` each time.
- The model has no memory of prior turns. Each restart looks like turn 1 from the agent's side.
- Even though the UI now persists the transcript on disk (#TODO once #PR lands) and shows continuity to the user, the LLM itself does not see the earlier conversation when the user types the next prompt.

This diverges from how `claude --resume <session-id>` works in plain Claude Code, where the same session id round-trips and the agent reloads the on-disk JSONL transcript.

## Two layers of session id today

1. **aoe `CockpitSessionId`** — our internal id (e.g. `s-tmux-XXXX`), stable across restarts, in URLs + `sessions.json` + the cockpit event store.
2. **ACP `acp_session_id`** — assigned by the agent in the `session/new` response. Lives only inside the running agent subprocess; we discard it on shutdown.

The mismatch is why restart loses agent context: layer 1 is durable, layer 2 is not, and we never thread the agent-assigned id back into the next spawn.

## Proposal

Add `session/load` support to the supervisor's spawn path:

1. Persist the `acp_session_id` returned by `session/new` alongside our `CockpitSessionId` (a column in the event store, or a field on the session record).
2. On `Initialize`, read the agent's capability flags. The `agent-client-protocol-schema 0.12.0` `InitializeResponse` exposes `agent_capabilities.load_session: bool`.
3. On spawn, if the agent advertises `load_session = true` AND we have a stored `acp_session_id` for this cockpit session → send `LoadSessionRequest::new(stored_id, cwd)` instead of `NewSessionRequest::new(cwd)`.
4. If load fails (e.g. the agent has GC'd that session id), fall back to `session/new` and clear the stored id so the next spawn doesn't keep retrying a dead id.
5. On `cockpit/disable` and session deletion, drop the stored id (already drop transcript and seq counter in those paths).

## Agent compatibility

- `claude-agent-acp` persists sessions under `~/.claude/projects/<hash>/<id>.jsonl` and supports `session/load`. This is the primary backend and the main beneficiary.
- `aoe-agent` (our bundled multi-provider agent) does not currently persist conversation state, so `load_session` should be `false` there until we add that.
- Other adapters (`opencode`, etc.) — capability flag handles them; we just won't try `session/load` if they advertise it as unsupported.

## Out of scope for this issue

- Conflict resolution if the agent's stored transcript and our event store disagree about turn order. The current sketch assumes the agent's record wins for context-window purposes; we keep our own log only for UI replay.
- Multi-device handoff (two clients pointing at the same cockpit session and one of them has stale local state). The replay endpoint already handles this for the UI; the agent side is unaffected.

## Related work this issue unblocks

- **Session usage hint in the cockpit composer** (shipped on `fix-acp-webui` as commit `feat(cockpit): show ACP session usage hint in composer footer`). Maps ACP `SessionUpdate::UsageUpdate` to a typed `Event::UsageUpdated` and renders `used / size (pct%) · cost` between the mode picker and the Send button.
  - Today: works correctly within a single agent run. After an `aoe serve` restart, the agent's internal token counter resets (we always call `session/new`), so the UI displays the last persisted `UsageUpdated` snapshot until the agent emits a new one — which will be a much smaller number, since the agent's context is empty.
  - When this issue lands and we adopt `session/load`: the agent's counter survives restart, the next `UsageUpdate` resumes from the right value, and the cockpit hint stays accurate across restart. No code change needed on the usage-hint side; the typed event already replays through the event store.
  - Implementation note for whoever picks this up: when the load-session path returns an error and we fall back to `session/new`, also clear `CockpitState.usage` (or emit `UsageUpdated { used: 0, size: <last_known> }`) so the hint doesn't briefly mislead users into thinking the agent retained context when it didn't.

## How tested

Not implemented yet — issue tracks the work.

Seluj78:
> [!WARNING]
> **This PR depends on #995 and #992 being merged first.** It includes cherry-picked commits from both branches (`feat: isolate debug-build state from release` from #995's `debug-alongside-production`, plus the three deletion / log-noise fixes from #992's `webui-debug-fixes`). Once those merge into `main`, this branch should rebase to drop the duplicates.

## Description

Big polish pass on the cockpit (ACP) surface — 39 commits cutting across the Rust server, the React web UI, and the SQLite event store. Closes https://github.com/njbrake/agent-of-empires/issues/997 plus a long tail of UX rough edges that surfaced while testing the ACP rendering surface end-to-end.

Demo video:
[Cockpit polish — fix-acp-webui demo](https://youtu.be/GtH7l9Iai9g)

### Headline beats

- **`session/load` resume across restarts** (#997). Cockpit now captures the agent-assigned ACP session id, persists it on `Instance.cockpit_acp_session_id`, and on subsequent spawns calls `LoadSessionRequest` instead of always `NewSessionRequest`. The LLM keeps prior-turn context across `aoe serve` stop+start. On load failure it falls back to `session/new` and surfaces an amber "Conversation context reset" callout in the transcript.
- **SQLite-backed conversation persistence**. The replay buffer was an in-memory ring per `AppState` — it didn't survive a daemon restart, and a fresh page load ran on the agent's stream alone (no user-side echoes). New `EventStore` (WAL-mode SQLite, keyed by `(session_id, seq)`, retention via `[cockpit] replay_events`) feeds both WS-on-connect drain and `GET /cockpit/replay`. Hydrates `next_seqs` from `max(seq)+1` on boot so post-restart writes don't collide with disk via `INSERT OR IGNORE`. `UserPromptSent` is now published before forwarding to the agent so user turns are part of the durable log.
- **Live cockpit master toggle** in web Settings → Cockpit tab. `cockpit_master_enabled` flipped from a startup snapshot to an atomic mirrored from `config.cockpit.enabled`, plus `PATCH /api/cockpit/master` that persists + updates the atomic. Reconciler / REST gates re-read the atomic each call. The `AOE_EXPERIMENTAL_COCKPIT=1` env-var gate is preserved end to end.
- **Tool output rendering**. Server now extracts text content blocks from `ToolCallUpdate.fields.content` and forwards via `ToolCallCompleted.content` (and a new `ToolCallContent` streaming variant). Frontend reducer buffers per-tool streaming text and drains it on completion when the final update is empty. Bash/Read/Search/Fetch/Generic cards render the actual stdout (or stderr on error), not the literal status word "completed". Late-arriving `raw_input` from claude-agent-acp's two-step `ToolCallUpdate` flow now patches the in-flight tool-card title in place (`$ Terminal` placeholder is gone).
- **ANSI + console + GFM**. New `AnsiBlock` parser handles SGR escapes (16/256/truecolor), strips non-SGR CSI noise, and collapses `\\r` repaints. `HighlightedBlock` routes ANSI text to it; otherwise Shiki via a fence-alias map for `console`/`shellsession`/`bash-session`/`terminal`/`shell`. `MarkdownTextPrimitive` plugs in `remark-gfm` for tables / strikethrough / task lists / autolinks. Tables get a sticky header, monospace cells, zebra rows, and a wrapper that lets short tables fill the bubble while long ones horizontal-scroll inside.
- **Composer hardening**. ACP `UsageUpdate` is now a typed `Event::UsageUpdated`; composer footer renders e.g. `12.3k / 200k (6%) · $0.42` with amber/rose tones at 75/90% (hidden on mobile). Slash command picker reads the agent's `available_commands` (43 real entries from a typical claude setup) instead of 4 hard-coded placeholders. Empty-output slash commands like `/usage` get a muted "Command produced no output." notice. Approval cards render args as a labelled key/value list; raw `<pre>` is the fallback for non-object payloads. Several non-functional buttons (Edit / Regenerate / Attach / cockpit→terminal switch / `↵ Send` chip / hover copy bar) removed.
- **Status sync**. New `cockpit_status_listener` subscribes to the cockpit broadcast and maps ACP events onto the existing `Status` enum (Running / Waiting / Idle / Error) so the sidebar dot lights up for cockpit sessions just like it does for tmux sessions. The `status_poll_loop` no longer clobbers in-memory cockpit status from disk on the 2s reload — it overlays the in-memory state before the diff/swap.
- **Spawn / lifecycle correctness**. `pending_spawns: HashSet<String>` reservation set in `Supervisor` so a concurrent `Supervisor::spawn` for the same session id is rejected as `AlreadyRunning` instead of two parallel handshakes overwriting each other (which used to surface as a fake "ACP agent crashed more than 3 times in 60s" banner). Reconciler auto-spawns serialize on startup so the claude-agent-acp SDK lazy-install isn't raced. Post-delete spawn errors are downgraded to debug. `wait_for_worker` (10s, 50ms tick) lets fast-clicked prompts land on a still-spawning worker instead of 404-ing. WS-on-connect drains the replay buffer to close the upgrade-gap race that used to drop the agent's first chunks.
- **Crash recovery**. Supervisor drain-channel closure logged as warn; respawn-budget tally on every decision (count/window/pruned). Comprehensive crash diagnostics: agent stderr drained line-by-line into `cockpit.acp.stderr` (also fixes the >64KB pipe-fill deadlock), spawn details logged at `cockpit.acp.spawn`, EnvFilter widened to include the `cockpit` target tree. New `AOE_LOG_LEVEL=trace|debug|info|warn|error` (graduated; legacy `AGENT_OF_EMPIRES_DEBUG=1` aliased to `debug`); raw JSON-RPC firehose moved behind `AOE_ACP_TRACE=1`. After a crash + respawn, `AcpSessionAssigned` is also emitted on `session/load` so the red startup-error banner heals on its own and the sidebar dot reverts from red to grey without waiting for the next user prompt.
- **Offline state**. New `useServerDown()` React hook on top of `connectionState`. New-session buttons (top-right + per-repo `+`), session wizard launch, settings panel, approval buttons, and the cockpit↔terminal switch all gate on it — disabled with a swapped tooltip and an inline notice. Settings tab content goes into `<fieldset disabled>` with a banner above. Cockpit transcripts cached in a module-level `Map` keyed by session id so remounting the panel (e.g. tab nav) hydrates from the last known state instead of an empty chat; `fetchReplay` tops up via cached `lastSeq`.

### Out of scope (followed up separately)

- **Context primer when `session/load` fails** — synthesizing a summary from our SQLite event store and prepending to the next user prompt. Tracked in https://github.com/njbrake/agent-of-empires/issues/1004.
- **`aoe-agent` persistence** — teaching our bundled agent to persist conversation state and advertise `load_session: true`. Tracked in https://github.com/njbrake/agent-of-empires/issues/1005.
- **Proactive agent-process death detection** — noticing the spawned agent died between prompts (instead of waiting for the user's next prompt to fail with `Broken pipe` and then respawn). Tracked in https://github.com/njbrake/agent-of-empires/issues/1007.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

(Recording: https://youtu.be/GtH7l9Iai9g — embedded above.)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Code (Opus 4.7, 1M context). Used for planning, exploration, scaffolding, and incremental commits across the 39-commit branch. Each commit was reviewed and tested locally before being staged.

**Any Additional AI Details you'd like to share:** Iterated in plan mode for the larger landings (`session/load` resume, SQLite event store, master toggle). After each landing the user (Seluj78) walked the surfaces by hand and reported rough edges (post-load history-replay duplication, sticky startup-error banner after respawn, run-on context_reset rendering, color polish, etc.); each was diagnosed and fixed in a follow-up commit on this branch.

---

## Manual Testing TODO

All boxes checked unless noted otherwise. Build with `cargo build --features serve`, run `aoe serve --no-auth`, set `AOE_EXPERIMENTAL_COCKPIT=1`, create at least one cockpit-mode session against `claude-agent-acp` ("claude" below).

### 1. Cockpit master toggle (live, no restart)

- [x] 1.1 — Live toggle persists and reconciles: `PATCH /api/cockpit/master` returns 200; `config.toml [cockpit] enabled` flips; running workers shut down / respawn within ~2s; no restart required.
- [x] 1.2 — Env-var gate enforced: without `AOE_EXPERIMENTAL_COCKPIT=1`, PATCH returns 403, Settings UI hides the Cockpit tab.
- [x] 1.3 — `/api/about` reports `cockpit_master_enabled` and `cockpit_env_enabled` separately; combined `experimental_cockpit` only true when both are.

### 2. ACP session lifecycle / spawn correctness

- [x] 2.1 — `session/load` resumes context across `aoe serve` restart; `cockpit_acp_session_id` persisted in `~/.agent-of-empires{,-dev}/profiles/default/sessions.json`.
- [x] 2.2 — Bogus stored id falls back: amber "Conversation context reset" callout, usage hint hidden until next `UsageUpdate`, sessions.json rewritten with fresh id.
- [x] 2.3 — Capability flag respected: agent without `load_session` advertised → `session/new` is used, no reset notice.
- [x] 2.4 — Crash respawn uses the latest stored id (`pkill` the agent → supervisor respawns through `session/load` → model recalls prior turn).
- [x] 2.5 — No duplicate trailing assistant bubble after restart (`suppress_history_replay` flag set before `LoadSessionRequest` send).
- [x] 2.6 — `cockpit_disable` clears stored id; debug log shows the clear.
- [x] 2.7 — Concurrent-spawn race no longer surfaces as crash banner (`pending_spawns` reservation set).
- [x] 2.8 — Reconciler auto-spawns serialize on startup (no claude-agent-acp SDK lazy-install race).
- [x] 2.9 — Post-delete spawn errors silenced (debug-only, no UI banner).
- [x] 2.10 — Fast-click prompt during spawn waits for worker (`wait_for_worker` 10s/50ms); cancel + set_mode also wrapped.
- [x] 2.11 — Pre-subscribe events not lost: WS on-connect drains the replay buffer (`replayed=N>0` in debug log).

### 3. Cockpit Status (sidebar dot) reflects ACP state

- [x] 3.1 — Status flips with ACP events: Running on prompt, Waiting on approval, Idle on Stopped, Error on startup-error.
- [x] 3.2 — Status survives the 2s `status_poll_loop` disk reload (in-memory cockpit status overlay re-applied).
- [x] 3.3 — Terminal lifecycle states (Stopped/Deleting/Creating) preserved during deletion.

### 4. Conversation persistence (event store + UserPromptSent)

- [x] 4.1 — User messages survive page reload (alternating user/assistant rows, no collapsed assistant blob).
- [x] 4.2 — User messages survive session switch.
- [x] 4.3 — Conversation survives `aoe serve` restart (`/cockpit/replay` returns full prior history; new prompts get `stored_max+1` seq).
- [x] 4.4 — Per-session retention cap honored (oldest events pruned with debug log at `cockpit.event_store`).
- [x] 4.5 — Session delete purges store; recreated session starts fresh.
- [x] 4.6 — `cockpit_disable` purges store; re-enable starts fresh.
- [x] 4.7 — Optimistic dedupe: server echo rewrites optimistic id to `user-seq-N`; back-to-back prompts both render once.
- [x] 4.8 — Replay reconstruction order correct mid-conversation.

### 5. Tool call rendering

- [x] 5.1 — Bash/Read/Search/Fetch/Generic cards render content, not the literal "completed" word.
- [x] 5.2 — Streaming `ToolCallContent` → completion drain. **Note:** covered by Rust unit tests (`acp_client.rs`) and vitest reducer specs; manual repro against claude-agent-acp not feasible because the agent does not emit incremental stdout via `ToolCallUpdate`. Path verified via SQLite event-store inspection.
- [x] 5.3 — Markdown code-fence unwrap on `console`/etc. fences.
- [x] 5.4 — Late-arriving `raw_input` patches the bash card title in place (no more `$ Terminal` placeholder).
- [x] 5.5 — Tool failure path shows actual stderr/output, not "tool failed".

### 6. Bash / shell rendering polish

- [x] 6.1 — `console` / `shellsession` / `bash-session` / `terminal` / `shell` fences highlight correctly via Shiki alias map.
- [x] 6.2 — Full command rendered in expanded card body with bash highlighting (long commands no longer truncated).
- [x] 6.3 — ANSI SGR codes render as colors (16/256/truecolor); no raw `[01;34m` leaks.
- [x] 6.4 — Non-SGR CSI noise (cursor moves, line erases) stripped.
- [x] 6.5 — `\r`-based progress bars collapse to final state per line.
- [x] 6.6 — Column layouts (`ps aux`, `df -h`) preserved (whitespace-pre).

### 7. Markdown rendering

- [x] 7.1 — GFM tables render as real `<table>` (e.g. `/context` output).
- [x] 7.2 — Strikethrough / task lists / autolinks render correctly.
- [x] 7.3 — Long tables: sticky header, monospace cells, zebra rows, hover, horizontal scroll, vertical cap.
- [x] 7.4 — Short tables fill bubble width; wide tables horizontal-scroll inside the wrapper.

### 8. Composer / footer / picker

- [x] 8.1 — Slash command picker uses agent's `available_commands` (real list, not 4 placeholders); trailing space inserted only on `Unstructured` input spec.
- [x] 8.2 — Usage hint in composer footer (`12.3k/200k (6%) · $0.42`); amber ≥75%, rose ≥90%; hidden on mobile; persists across reload via SQLite replay.
- [x] 8.3 — Empty-output notice ("Command produced no output.") on `/usage`, `/status`, `/memory`, `/permissions`, `/agents`, `/mcp`.
- [x] 8.4 — Approval card key/value rendering; raw `<pre>` fallback for non-object args.
- [x] 8.5 — Removed Edit / Regenerate / Attach / cockpit→terminal switch / `↵ Send` chip buttons; terminal-side switch still works.
- [x] 8.6 — Hover copy bar gone from user/assistant messages; per-code-block copy buttons preserved.

### 9. Cockpit conversation column / layout

- [x] 9.1 — Wider column on large screens: `xl:max-w-4xl` (≥1280px), `2xl:max-w-5xl` (≥1536px); standard laptops keep `max-w-3xl`.
- [x] 9.2 — Transcript hydrates from cache on tab-navigation remount; daemon kill disables Send + shows banner; transcript stays visible; reconnect tops up via cached `lastSeq` with no duplicate rows.

### 10. Web offline state

- [x] 10.1 — New-session buttons (top-right + per-repo `+`) disabled when backend unreachable.
- [x] 10.2 — Session wizard launch blocked at Review step with inline notice.
- [x] 10.3 — Settings panel frozen via `<fieldset disabled>` with banner above.
- [x] 10.4 — Cockpit substrate switch and approval buttons (Allow/Always/Deny) disabled with inline offline notice on the card.

### 11. Logging / diagnostics

- [x] 11.1 — `AOE_LOG_LEVEL=trace|debug|info|warn|error` graduated; legacy `AGENT_OF_EMPIRES_DEBUG=1` aliased to `debug`.
- [x] 11.2 — `AOE_ACP_TRACE=1` opt-in for raw JSON-RPC firehose; orthogonal to `AOE_LOG_LEVEL`.
- [x] 11.3 — Crash diagnostics: spawn details, agent stderr drain, connection-task `try_wait`, supervisor drain-channel closure as warn, respawn-budget tally; agent stderr drain prevents >64KB pipe-fill deadlock.
- [x] 11.4 — Event store debug breadcrumbs: open path + retention, record seq+kind+bytes, dup-key skips, prune count, replay returned-count, hydration size, delete count.
- [x] 11.5 — `unstable_session_usage` feature enabled; usage_update notifications deserialize cleanly against claude-agent-acp ≥ 0.32.0.

### 12. Automated test suites

- [x] `cargo test --features serve` — passes (existing + new tests in `instance.rs`, `acp_client.rs`, `state.rs`, `supervisor.rs`, `event_store.rs`, `cockpit_acp_smoke.rs`, `derive_cockpit_status` mapping).
- [x] `cargo clippy --features serve -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] `cd web && npx vitest run` — covers `applyEvent` reducer for AcpSessionAssigned, SessionContextReset, UserPromptSent dedupe + replay reconstruction, ToolCallUpdate patch flow.
- [x] `cd web && npx playwright test` — baseline e2e (no new e2e on this branch; existing suite passes).

### 13. Cross-cutting smoke

- [x] Cold start → cockpit session against claude → 5-turn conversation including Bash + `/context` + UsageUpdate + approval → reload → restart → context preserved.
- [x] 3 concurrent cockpit sessions; sidebar dots reflect each independently; reload hydrates all 3 transcripts.
- [x] Offline path: kill daemon → transcript stays, send button + approval buttons disable; restart → fetchReplay top-up; new prompts work.
- [x] Mixed substrate: terminal + cockpit session in parallel; terminal-side substrate switch still works; `/cockpit/enable` + `/cockpit/disable` still flip mode.


njbrake:

Hey @Seluj78, did a deep review pass and pushed several commits on top of your branch. Heads up before this lands.

**What I pushed**

1. `fb523a9 review: bug fixes and consolidations from #1008 review`. Bug fixes and consolidations surfaced during a multi-agent review pass (backend, frontend, cross-cutting). Highlights:
   - **Frontend dedupe regression.** Sending the same prompt twice back-to-back used to dedupe against the most-recent matching optimistic row, leaving the older one as a ghost duplicate. Fixed to match the oldest row by `findIndex`.
   - **Scoped `suppress_history_replay`.** Previously the post-`session/load` window dropped *every* notification, including ambient state like `current_mode_update`, `available_commands_update`, and `UsageUpdate`. Now suppression is limited to transcript-replay events; ambient updates pass through so the composer footer and slash-command picker stay live after a restart.
   - **Spawn cancellation race.** `cockpit_disable` racing with an in-flight `Supervisor::spawn` could leave a worker entry in `pending_spawns` after the disable returned, blocking the next enable with `AlreadyRunning`. Added a `cancelled_spawns` set with a 3-step protocol (shutdown marks, spawn pre-insert checks, RAII reservation cleans up).
   - **`turnHasOutput` flag** on `CockpitState` so the "Stopped with no output" check is O(1) instead of walking activity history each turn.
   - **Consolidations.** Extracted shared `parseJsonObject` / `pickStr` / `pickFirst` into `web/src/lib/cockpitArgs.ts` (was duplicated across `ApprovalCard.tsx` and `ToolCards.tsx`). Removed the now-dead `replay_buffer.rs` since `event_store.rs` is the canonical history. Wrapped `activityToThreadMessages` in `useMemo`. Stable per-tool-call timestamps so `AssistantToolCall` doesn't tick `new Date()` per render.
   - **LRU-capped `stateCache`** (cap 32) for the cockpit hook; exposed `clearCockpitCache(sessionId)` so the App-level delete path purges the cache after the REST delete succeeds.
   - **`statusRef` synchronously updated** in `ws.onopen`/`onerror`/`onclose` alongside the React `setState`, so `sendPrompt` reads the correct status without a one-tick lag.
   - **`Arc<crate::cockpit::Event>` for `CockpitBroadcastFrame.event`** instead of `serde_json::Value`. Cheaper fanout cloning; custom `serde::Serialize` keeps the wire shape identical.
   - **`Supervisor::spawn` takes `SpawnRequest`** instead of 7 positional args.
   - **Consolidated two listener tasks** into a single `cockpit_event_listener` so the status broadcast and the `acp_session_id` persistence path can't get out of sync.

2. `6dc1ba7 Merge remote-tracking branch 'origin/main'`. After #992 and #995 merged to main, I merged main into your branch. One real conflict: `src/cli/serve.rs`'s `--port` handling. The cherry-pick used `cfg_attr` on the field; main's canonical version uses `Option<u16>` + `resolved_port()`. Resolved to main's.

3. `6bb8f18 docs(cockpit): trim user-facing persistence section`. The "Conversation persistence" section in `docs/cockpit.md` had implementation detail (SQLite event store keying, `agent_capabilities.load_session`, path references) that didn't fit a user-facing doc. Trimmed to four sentences; moved the technical detail into a module doc at the top of `src/cockpit/event_store.rs`.

4. `4fb1b58 fix(cockpit): enforce read-only mode on write endpoints`. Security finding from the review. `aoe serve --read-only` is documented to block every write endpoint with 403, but the cockpit write endpoints landed without the guard. A read-only viewer could POST to `prompt`, `cancel`, `spawn`, `shutdown`, `enable`, `disable`, `set_mode`, and `resolve_approval`, which would let them inject prompts and approve destructive tool calls. Added the same 403 response used by `sessions.rs` to all eight, via a `read_only_block(&state)` helper. Read endpoints stay unguarded.

5. `c2e7366 fix(cockpit): deny privileged keys in provider_env; scrub stderr secrets`. Two more from the security pass.
   - **`provider_env` denylist.** The previous filter only rejected `AOE_TOKEN`. An authenticated user could send `provider_env: [{"key": "PATH", "value": "/tmp/evil:/usr/bin"}]` and redirect the agent's binary lookup, or set `LD_PRELOAD` / `DYLD_INSERT_LIBRARIES` to inject arbitrary shared objects. New `provider_env_denyreason()` rejects `AOE_TOKEN`, the `always_forward` infra keys (PATH, HOME, USER, LANG, LC_ALL, TERM), and anything starting with `LD_` or `DYLD_`. Provider auth keys (`ANTHROPIC_API_KEY`, etc.) pass through, since per-session auth override is the legitimate use case.
   - **Stderr secret scrubber.** Agent stderr drains verbatim into `debug.log`. If an adapter prints an env value in an error path, the secret lands in a log a user shares when filing a bug report. New `scrub_stderr_secrets()` redacts well-known prefixes (Anthropic `sk-`, GitHub `ghp_`/`gho_`/`github_pat_`, AWS `AKIA`, `Bearer <token>`) before they hit the tracing log. Conservative scope; doc note in `cockpit.md#sharing-debug-logs` reminds users to still skim before attaching logs to bug reports.

**Scope notes**

Everything in the review is in-scope for "Cockpit polish" except the graduated `AOE_LOG_LEVEL` env var, which is a global logging change. Decided to leave it bundled; small enough not to be worth a split. Worth mentioning in the merge commit message.

**Tests**

1704 lib tests pass (8 new), 84 web tests pass, clippy + fmt clean on the merged branch.

Take a look when you can; happy to discuss any of it. Planning to squash-merge once you've had a chance to glance.

Seluj78:
@njbrake LGTM ! 🙌

PR #1023, opened 2026-05-12
fix(deletion): tear down tmux + container before host worktree

njbrake:
## Description

For sandboxed sessions with a worktree, the deletion pipeline tore resources down in the wrong order, racing the live in-container agent.

**Old order:** `git worktree remove` (host) → `docker rm -f` → tmux kill.
The container was still alive (with the worktree bind-mounted) and the agent was still writing during the host-side worktree probe. Same user action produced different outcomes depending on what the agent happened to be doing at that microsecond. Symptoms users reported:

- `fatal: validation failed, cannot remove working tree: '<path>/.git' does not exist`
- `error: cannot delete branch 'feat' used by worktree at '<path>'`
- `fatal: '<path>' contains modified or untracked files`

The `try_sandbox_dir_cleanup` recovery path added in #471 catches EPERM for root-owned files, but only after a failed host probe, by which point the worktree's admin state could already be inconsistent.

**New order** (matches what `Instance::stop` at `src/session/instance.rs:1413-1426` already does for the non-delete path):

1. `on_destroy` hooks (container + worktree still live, unchanged)
2. tmux kill (severs the user's `docker exec`, so the agent's shell stops)
3. **sandbox worktree preclean** (only if sandboxed + deleting worktree): `find . -mindepth 1 -delete` from inside the container wipes root-owned files *while the bind mount is still attached*. This is the same primitive the old recovery fallback used, just promoted to a deterministic step.
4. `docker rm -f` (releases the bind mount)
5. host-side `git worktree remove` (now: empty dir, no agent, no mount, no permission fight)
6. branch delete
7. hook status cleanup

Each stage emits a `tracing::debug!(stage = …)` event for observability.

## Intentional behavior changes

- `delete_worktree=true, delete_sandbox=false` no longer surprise-removes the container. The pre-fix recovery fallback would force-remove the container even when the user explicitly opted out of container deletion. This is now honored end-to-end: `remove_managed_worktree` and `try_sandbox_dir_cleanup` take an `allow_container_removal` flag; the TUI/web pass `request.delete_sandbox`, and `aoe remove` passes `!args.keep_container`. When the flag is `false`, the fallback bails out with a clear log line and the worktree removal surfaces the original permission error instead of nuking the container behind the user's back.
- Sandboxed + `delete_worktree=true` always runs the in-container preclean (one extra `docker exec` per deletion, sub-second). Previously this only ran as recovery after a failed host probe.

## History considered

I traced the design through the relevant commits before changing it:

- #73 (Jan 2026) introduced the original `DeletionPoller` with order: worktree → container → tmux. This pre-dated awareness of root-owned sandbox files.
- #405 (Mar 2026) added "clean files from inside the container first" specifically because sandbox containers create root-owned files that host can't unlink (especially on macOS Docker Desktop VirtioFS).
- #471 (Mar 2026) introduced `try_sandbox_dir_cleanup` as a *fallback* when the host probe hits EPERM. Container-first was the right idea, but it was wired up as recovery rather than the primary path.
- #707 (Apr 2026) extracted the logic into `src/session/deletion.rs` for sharing between TUI and web. Preserved the existing order.

So the worktree-first order was a holdover from before sandbox concerns existed, and successive PRs band-aided the fallout instead of revisiting it. This PR moves to container-first as the primary path.

## Known limitations

- **Residual race window**: between `tmux_kill` and `container.remove(true)`, the agent process inside the container is reparented to PID 1 rather than killed (tmux's SIGHUP usually terminates well-behaved agents, but not guaranteed). If the agent writes a new file during the in-container preclean, `find` can miss it, and the leaked root-owned file would then fail the host-side worktree removal. The window is much smaller than the original race, and the user can recover by re-running with force. Closing this fully would require `pkill`-style process cleanup inside the container before `find` — left as a follow-up.
- **`try_sandbox_dir_cleanup` is now reachable in narrow conditions only.** With the new ordering it fires only when `delete_sandbox=false` AND preclean failed AND the host probe hit EPERM. Could be deleted in a follow-up release once the new order has shipped.

## Tests

Added to `src/session/deletion.rs`:

- `ordering::sandboxed_with_worktree_kills_tmux_and_container_before_worktree` — captures stage events via a custom `tracing` Layer; asserts `tmux_kill < preclean < container_remove < worktree_remove < branch_delete`. Verified it catches the regression by commenting out the tmux_kill stage; both ordering tests fail as expected.
- `ordering::unsandboxed_kills_tmux_before_worktree` — non-sandbox path: no preclean emitted, tmux still precedes worktree.
- `ordering::e2e_real_worktree_is_removed_on_disk` — builds a real `git init` + `git worktree add` on disk, runs `perform_deletion`, asserts the worktree dir, `.git/worktrees/<name>/` admin entry, and branch are all gone.
- `ordering::e2e_real_worktree_with_untracked_files_force_removed` — reproduces the dirty-tree path: simulates agent-leftover files, asserts non-force delete fails cleanly without nuking the dir, then force delete succeeds end-to-end.

Added to `src/git/cleanup.rs`:

- `test_try_sandbox_dir_cleanup_respects_allow_container_removal_false` — regression test for the new flag: sandboxed instance + `allow_container_removal=false` must return false without touching the container.
- `test_try_sandbox_dir_cleanup_returns_false_for_non_sandboxed` — guard for the existing early-return.

## Test plan

- [x] `cargo test --lib` — 1509 passed
- [x] `cargo test --tests` — full integration suite green (includes the 55 e2e tests; the 2 docker-only ones remain `#[ignore]`)
- [x] `cargo clippy --lib --tests` — clean
- [x] `cargo fmt --check` — clean
- [x] Verified the new ordering tests catch a deliberately reintroduced bug
- [x] 5 consecutive parallel runs of the lib suite — green every time
- [ ] Manual: delete a sandboxed session with worktree on macOS Docker Desktop and confirm no "branch / worktree doesn't exist" errors. **I couldn't run this; the dev sandbox here has no docker. Reviewer to validate on a real session.**

## PR Type

- [x] Bug Fix

## Checklist

- [ ] Documentation was updated where necessary — no user-facing docs affected
- [ ] For UI changes: not a UI change

## AI Usage

**AI Model/Tool used:** Claude Opus 4.7 (1M context)

**Any Additional AI Details you'd like to share:** The race condition diagnosis, ordering rewrite, and tests were done in a single Claude Code session. I reviewed every change, walked the historical commits (#73, #405, #471, #707) to understand the design intent, and gated the merge on the new ordering tests. The deliberate-bug-injection check confirmed the tests catch the regression they're meant to. Follow-up commits addressed reviewer feedback: tightened the `delete_sandbox=false` semantics so the fallback honors the flag, added regression tests for the new plumbing, and used the cached `is_sandboxed` bool consistently.

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -134,12 +134,17 @@ pub async fn run(profile: &str, args: RemoveArgs) -> Result<()> {
 
                 match GitWorktree::new(main_repo.clone()) {
                     Ok(git_wt) => {
+                        // --keep-container means the sandbox fallback
+                        // must not surprise-tear-down the container to
+                        // free a permission-bound worktree.
+                        let allow_container_removal = !args.keep_container;
                         match remove_managed_worktree(
                             &git_wt,
                             &worktree_path,
                             &main_repo,
                             &inst,
                             args.force,
+                            allow_container_removal,
                         ) {
                             Ok(()) => {
                                 worktree_removed = true;
--- a/src/git/cleanup.rs
+++ b/src/git/cleanup.rs
@@ -281,6 +281,16 @@ pub fn cleanup_sandbox_worktree(instance: &Instance) -> bool {
 /// - `.git` file present: uses `git worktree remove`, falls back to
 ///   container cleanup for sandboxed sessions with permission errors
 ///
+/// `allow_container_removal` controls whether the sandbox fallback is
+/// permitted to force-remove the container as part of its recovery.
+/// Set to `true` when the caller has already requested container
+/// deletion (or is about to), and `false` when the caller wants to
+/// preserve the container (e.g. `aoe remove --keep-container` or
+/// `delete_worktree=true, delete_sandbox=false`). When `false` and
+/// the fallback is the only way to make progress, the worktree
+/// removal fails with the original permission error instead of
+/// quietly tearing down the container behind the user's back.
+///
 /// Returns `Ok(())` if the worktree was successfully removed, or
 /// `Err(errors)` with error messages on failure.
 pub fn remove_managed_worktree(
@@ -289,6 +299,7 @@ pub fn remove_managed_worktree(
     main_repo: &Path,
     instance: &Instance,
     force: bool,
+    allow_container_removal: bool,
 ) -> Result<(), Vec<String>> {
     let mut errors = Vec::new();
     let has_dot_git = worktree_path.join(".git").exists();
@@ -298,6 +309,7 @@ pub fn remove_managed_worktree(
         has_dot_git,
         is_sandboxed = instance.is_sandboxed(),
         force,
+        allow_container_removal,
         "worktree cleanup starting"
     );
 
@@ -313,7 +325,12 @@ pub fn remove_managed_worktree(
             Err(e) => {
                 tracing::debug!(error = %e, kind = ?e.kind(), "remove_worktree_dir failed (no .git)");
                 if is_permission_error(&e.to_string())
-                    && try_sandbox_dir_cleanup(worktree_path, main_repo, instance)
+                    && try_sandbox_dir_cleanup(
+                        worktree_path,
+                        main_repo,
+                        instance,
+                        allow_container_removal,
+                    )
                 {
                     worktree_removed = true;
                 } else {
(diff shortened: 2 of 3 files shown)


Seluj78 (approved):
LGTM!

PR #1042, opened 2026-05-12
refactor: remove dead claude.config_dir config field

### LINKED ISSUE #1026: `claude.config_dir` profile field is dead code — never propagated to spawned session env
`get_claude_config_dir()` in `src/session/config.rs:747` is defined and reads `config.claude.config_dir` from the resolved profile config, but `grep -rn 'get_claude_config_dir()'` returns ZERO call sites anywhere in `src/`. The schema is in place (storage + profile override merge) but **nothing reads it back at session-launch to propagate into the spawned process's env**.

Effect: a user who sets up `~/.agent-of-empires/profiles/work/config.toml` with:
```toml
[claude]
config_dir = "~/.claude-accounts/work"
```
…and runs `aoe -p work session start <project>` sees the profile load (sessions.json gets the work-profile entries), BUT the spawned claude process inherits the PARENT shell's `CLAUDE_CONFIG_DIR` env (usually unset → defaults to `~/.claude`). Account selection silently ignored.

Fix (~15 LoC) in `src/tmux/session.rs:435` `build_create_args()`: chain `set-environment` after the `new-session` args:

```rust
// In build_create_args(), after the new-session args, before the cmd args:
if let Some(cfg_dir) = merged_config.claude.config_dir.as_ref() {
    let expanded = expand_tilde(cfg_dir);
    args.push(";");
    args.push("set-environment");
    args.push("-t");
    args.push(&session_name);
    args.push("CLAUDE_CONFIG_DIR");
    args.push(&expanded);
}
```

Then child claude processes inherit `CLAUDE_CONFIG_DIR=<profile-pinned-value>` from tmux session env. Account selection becomes real per-profile.

Happy to PR.

kimjune01:
## Summary
- Fix `claude.config_dir` profile field which was dead code: the value was parsed from config but never propagated to the environment of spawned tmux sessions
- Add `append_claude_config_dir_env` utility to set `CLAUDE_CONFIG_DIR` in the tmux session environment when the profile specifies a custom config directory

Fixes #1026

## Test plan
- Added unit tests for `append_claude_config_dir_env` covering set, unset, and empty cases
- Existing session tests pass unchanged

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -209,17 +209,6 @@ notify_in_cli = true
 | `check_interval_hours` | `24` | Hours between update checks |
 | `notify_in_cli` | `true` | Show update notifications in CLI output |
 
-## Claude
-
-```toml
-[claude]
-config_dir = "~/.claude"
-```
-
-| Option | Default | Description |
-|--------|---------|-------------|
-| `config_dir` | (none) | Custom Claude Code config directory. Supports `~/` prefix. |
-
 ## Profiles
 
 Profiles provide separate workspaces with their own sessions and groups. Each profile can override any of the settings above.
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -17,9 +17,6 @@ pub struct Config {
     pub theme: ThemeConfig,
 
     #[serde(default)]
-    pub claude: ClaudeConfig,
-
-    #[serde(default)]
     pub updates: UpdatesConfig,
 
     #[serde(default)]
@@ -433,12 +430,6 @@ fn default_idle_decay_minutes() -> u64 {
     0
 }
 
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
-pub struct ClaudeConfig {
-    #[serde(default)]
-    pub config_dir: Option<String>,
-}
-
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct UpdatesConfig {
     #[serde(default = "default_true")]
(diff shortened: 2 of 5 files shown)


njbrake:

Thanks for digging into this, @kimjune01. The issue you filed was real and well-investigated, and your tilde-expansion edge cases in the second commit were sharp. I'm sorry for the heavy-handed rewrite; I force-pushed `3bc501c` on top of your branch, which replaced your three commits with a single deletion. To make sure you still get credit on the merge, I added you as a co-author on the new commit, so you'll show up in the squash commit author list and on the repo's contributor graph once this lands.

Here's why this PR became a deletion rather than a fix.

`get_claude_config_dir()` reads `Config::load_or_warn()`, which loads only the global config; the profile-merge in `profile_config::merge_configs` is bypassed. So the exact scenario in #1026 (setting `[claude] config_dir` in `profiles/work/config.toml` and running `aoe -p work …`) would still silently fall through after the original fix landed. The fix worked for the global case but not the documented profile case, and `docs/guides/configuration.md` advertises profile overrides as the primary use case for this field.

Wiring it correctly meant threading the profile through `Session::create_with_size()`, plumbing a profile-aware reader, adding a `FieldKey` plus `SettingField` plus `apply_field_to_global` plus `apply_field_to_profile` plus `clear_profile_override` entry in the settings TUI (per the rule in `CLAUDE.md`), and an e2e test asserting profile overrides win. That's a much bigger feature than the issue implied, and given that the field has been dead since it was added and the issue has zero engagement beyond yours, I'd rather just remove the dead surface than commit to the full feature.

Anyone who wants a non-default `CLAUDE_CONFIG_DIR` already has working paths today:

- `export CLAUDE_CONFIG_DIR=…` in their shell before launching `aoe`
- set it via an `on_launch` hook (works per-profile and per-repo today)
- add `CLAUDE_CONFIG_DIR=<path>` to `sandbox.environment` for containerized profiles

The new diff removes the field, the override, `get_claude_config_dir()`, `validate_path_exists` (its only caller), the `merge_configs` branch, the re-exports, and the Claude section in the configuration docs. `cargo build`, `cargo clippy`, `cargo fmt --check`, and all 1503 lib tests pass; `--features serve` also builds clean.

Closes #1026 as "won't fix in the form proposed; field removed instead." Hope to see another contribution from you soon.

PR #1070, opened 2026-05-12
feat(new-session): show path field before title

BTForIT:
## Description

Renders the path field before the title field in the new-session dialog (`n` keybind in the TUI).

The natural reading order for "make a new session" is "pick the directory, then name it," but the dialog asked for the name first. Putting path first lets the user orient on the working directory before deciding what to call the session.

### Implementation

- Factor `title_field()` into a method paralleling `path_field()`; swap their return values (`path_field` is now `base`, `title_field` is now `base + 1`).
- Swap the render blocks at the top of the dialog.
- The `base + 2 + ...` math for downstream tool/yolo/worktree/sandbox/group fields is unchanged; only the relative position of title and path within the first two slots flips.
- Test updates: swap `focused_field` indices in new_session unit tests to match the new order, plus inline comments documenting the layout.

### Notes

- Initial focus on the dialog is `focused_field = 0`. For multi-profile users this is the profile field (unchanged). For single-profile users this is now the path field instead of the title field, so users who press `n` and immediately type a name now need one Tab first.
- The folder-basename auto-fill from #1077 was removed in the fixup commit so the random civ-name fallback (`src/cli/add.rs`) stays the canonical empty-submit behavior across all flows (typed path, default cwd, Ctrl+P browse).

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test --lib` 1556/1556, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean)
- [x] For UI changes: manual TUI verification (open `n`, confirm path is on top)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Anthropic), via Claude Code CLI.

**Any Additional AI Details you'd like to share:** The diff and rationale are mine; the assistant drafted the implementation and PR body from my direction. I reviewed and edited before pushing.

### DIFF
--- a/src/tui/dialogs/new_session/mod.rs
+++ b/src/tui/dialogs/new_session/mod.rs
@@ -539,9 +539,9 @@ impl NewSessionDialog {
     }
 
     /// The field index of the path field. Path comes BEFORE title in the
-    /// dialog so the user browses to a directory first and the title can
-    /// then auto-fill from the folder basename. Shifts based on whether
-    /// the profile picker is visible at field 0.
+    /// dialog so the user picks the working directory before naming the
+    /// session. Shifts based on whether the profile picker is visible at
+    /// field 0.
     fn path_field(&self) -> usize {
         if self.has_profile_selection() {
             1
@@ -550,9 +550,7 @@ impl NewSessionDialog {
         }
     }
 
-    /// The field index of the title field. Title sits one slot AFTER path
-    /// so the picker-flow lands on title with a sensible default already
-    /// filled in (or empty if the user typed the path manually).
+    /// The field index of the title field. Title sits one slot AFTER path.
     fn title_field(&self) -> usize {
         if self.has_profile_selection() {
             2
@@ -836,19 +834,6 @@ impl NewSessionDialog {
                             .and_then(path_input::compute_path_ghost);
                         self.workspace_repo_dir_picker_active = false;
                     } else {
-                        // Auto-fill title from folder basename when the user
-                        // hasn't typed one yet. The folder name is almost
-                        // always what they'd call the session anyway, and
-                        // making it the default removes a second step after
-                        // the browse picker selects a directory.
-                        if self.title.value().is_empty() {
-                            if let Some(basename) = std::path::Path::new(&path)
-                                .file_name()
-                                .and_then(|n| n.to_str())
-                            {
-                                self.title = Input::new(basename.to_string());
-                            }
-                        }
                         self.path = Input::new(path);
                         self.recompute_path_ghost();
                     }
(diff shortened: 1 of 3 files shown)


njbrake:
Hi @BTForIT thank you for the PR: can you elaborate a little bit more on the experience of the user that you watched? Personally in my experience has been a bit different, basically whenever I open up the new session dialogue I am only entering a value for name, all of the other fields are inherited based on the profile that I'm using, and I generally want the directory to be whatever my current working directory is. 

I hesitate a little bit to accept this PR because it's changing a UX that has been stable for quite a while now. 

I definitely am open to feedback, especially if other contributors have opinions, but I would hesitate to accept this without more consensus that it's the right move. 

Maybe a compromise would be some sort of configuration setting that lets people decide the order in that dialogue? Idk

BTForIT:
Thanks for the thoughtful pushback. Let me explain the reasoning behind the order swap — happy to drop it if you still feel it's wrong after.

Workflow this enables: Ctrl+P browse a folder → title auto-fills to the folder basename → either accept it or append a hyphen suffix to disambiguate (e.g. `aoe-pr-cleanup`, `aoe-feature-x`). Folder-derived names feel more intuitive than the default civ name for "what is this agent actually working on?"

The three sub-changes line up with that workflow:
1. Standardize `Ctrl+P` notation — pure readability cleanup. Independent of the order debate.
2. Auto-fill title from folder basename — only fires when title is empty, so the "type a name first" workflow is fully preserved. Title is never overwritten.
3. Path before title — pairs with #2 so the Tab order is browse → see title populate → optionally edit.

No behavior loss for the type-a-name-first flow (typing a name suppresses the auto-fill). The order swap just makes folder-first naming the easy default.

If that still doesn't sit right after this context, happy to drop the order swap entirely and keep just #1 and #2 — or pivot to a config flag like you suggested.

BTForIT:
Rebased on main and dropped the two changes you already pulled into #1077 (Ctrl+P notation + folder-basename auto-fill). What's left here is just the path-before-title render swap, which completes the flow with #1077's auto-fill: browse to path (Ctrl+P) → title auto-populates from the folder basename → Tab past or edit.

Tests updated to swap `focused_field` indices for the new order. `cargo test --lib` 1557/1557, clippy clean, fmt clean.

If the title-first order is preferred for existing users, happy to gate this behind a `session.new_session_field_order` config knob instead. Let me know.

njbrake:
Thanks @BTForIT . I think I'm good with the title swap, but I don't yet want to accept the auto fill logic. The auto-populate empty name to a random civilization is in line with what other tools like conductor.build do to handle initial session creation, and tbh I just kind of like the fun age of empires vibe that you get when unpopulated names make there be randomly generated civilizations.

Adding special logic to autofill names if the user opens ctrl+p and selects a session adds a slight extra path and logic complication (i know, it's very slight) that I don't want to accept just yet.

BTW, I spun up a discord channel if you have any interest in joining to have more general aoe direction/discussions https://discord.gg/5N3QKX3f6s . It's a part of the Mozilla AI discord server

PR #1082, opened 2026-05-12
feat(session): transcript recovery on restart for Claude sessions

BTForIT:
## Description

Extends `aoe session restart` so a Claude session whose transcript on disk
has gone missing or pathologically large can recover instead of failing on
`--resume <sid>` or thrashing on autocompact.

A new `session::recovery` module runs a best-effort cascade against
`~/.claude/projects/<encoded>/<sid>.jsonl` at restart time:

1. Transcript present and within size budget: no-op.
2. Transcript missing but a thrash archive exists under
   `<proj>/archived/<sid>.jsonl.thrash-*`: restore the most recent archive,
   trimmed.
3. Transcript present but oversized (>50 MiB): trim it in place.
4. Transcript missing with no archive available: clear the stored session
   id so the next launch starts fresh, keeping the pane alive even if the
   conversation is lost.

The trim algorithm preserves the JSONL header (first 5 lines), the most
recent `compactMetadata` line (which carries the summary the live
conversation depends on), and the last 300 lines. Anything in between is
pre-summarized history. A stuck `Status::Error` from a prior failed
attach is also reset on restart so the post-restart status reflects the
fresh pane instead of the stale tmux gone-away message. Recovery is
scoped to non-sandboxed Claude sessions; sandboxed transcripts live
inside the container and are recovered separately on next launch.

## PR Type

- [ ] New Feature
- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording — Backend recovery logic, no UI change.

`cargo build --release`, `cargo test --release --lib session::recovery`
(8 unit tests pass), `cargo fmt --check`, `cargo clippy --release --lib
-- -D warnings` all green. No UI changes.

## AI Usage

- [ ] No AI was used

The recovery algorithm is a port of a small Python utility used locally
to rescue stalled Claude transcripts; the strategy (header + last
compactMetadata + tail) is the part that matters and has been validated
in practice over the last several weeks of triaging restart failures.
Tests are deterministic and use an isolated `CLAUDE_CONFIG_DIR` so the
real user state is never touched.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1496,6 +1496,17 @@ impl Instance {
             std::thread::sleep(std::time::Duration::from_millis(100));
         }
 
+        // Reset a stuck Error status so the post-restart status reflects the
+        // fresh pane, not a stale "tmux session is gone" message captured
+        // before this restart.
+        if self.status == Status::Error {
+            self.status = Status::Idle;
+            self.last_error = None;
+            self.last_error_check = None;
+        }
+
+        self.try_recover_transcript();
+
         self.start_with_size_opts(size, skip_on_launch)
     }
 
@@ -1577,6 +1588,43 @@ impl Instance {
                 return;
             }
             std::thread::sleep(std::time::Duration::from_millis(50));
+        }
+    }
+
+    /// Best-effort: rescue a Claude transcript that has gone missing or
+    /// oversized so `--resume <sid>` survives the restart instead of
+    /// thrashing on autocompact or failing outright. Only applies to Claude
+    /// sessions with a known `agent_session_id` running on the host (not
+    /// inside a sandboxed container, where the transcript lives in the
+    /// container's filesystem and is recovered separately on next launch).
+    fn try_recover_transcript(&mut self) {
+        if self.tool != "claude" || self.is_sandboxed() {
+            return;
+        }
+        let sid = match self.agent_session_id.as_deref() {
+            Some(s) if !s.is_empty() => s.to_string(),
+            _ => return,
+        };
+        match crate::session::recovery::recover_transcript_for_sid(&sid, &self.project_path) {
+            Ok(crate::session::recovery::RecoveryOutcome::NoArchiveFreshLaunch) => {
+                tracing::warn!(
+                    "restart '{}': transcript missing and no archive for sid {}; \
+                     clearing session id so claude launches fresh instead of failing on --resume",
+                    self.title,
+                    sid
+                );
+                self.agent_session_id = None;
+            }
+            Ok(outcome) => {
+                tracing::debug!("restart '{}': recovery outcome={:?}", self.title, outcome);
+            }
+            Err(e) => {
+                tracing::warn!(
+                    "restart '{}': transcript recovery failed (continuing anyway): {}",
+                    self.title,
+                    e
+                );
+            }
         }
     }
 
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1591,27 +1591,31 @@ impl Instance {
         }
     }
 
-    /// Best-effort: rescue a Claude transcript that has gone missing or
-    /// oversized so `--resume <sid>` survives the restart instead of
-    /// thrashing on autocompact or failing outright. Only applies to Claude
-    /// sessions with a known `agent_session_id` running on the host (not
-    /// inside a sandboxed container, where the transcript lives in the
-    /// container's filesystem and is recovered separately on next launch).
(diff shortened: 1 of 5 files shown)


njbrake:
Hi @BTForIT thank you for the PR: can you expand a little bit upon why the --resume arg is not sufficient?

jerome-benoit:
I do not see why a downstream ai coding agents tooling should alter a session content at resume. It basically just decept the whole semantic of a session resuming feature,  and makes no sense at all.

jerome-benoit:
> Hi @BTForIT thank you for the PR: can you expand a little bit upon why the --resume arg is not sufficient?

It's sufficient. An ai coding agents management tool is not meant to supersede built-in agent features by breaking their semantic. If the session resume feature in claude does not meet the user expectation, it should be handled by the claude team ...

BTForIT:
Fair point on the framing — restructuring this as the first impl of a per-agent harness-recovery contract rather than a Claude-specific module. The trait lives in `session::recovery`; today's logic moves to a `ClaudeRecovery` impl, and opencode/codex impls follow when their failure modes warrant. That maps directly onto Anthropic's own April "Managed Agents" framing of session vs. harness as separable layers — aoe's distinct value is making the harness tweakable per agent rather than Claude-only.

On the `--resume` semantic concern: the cascade is no-op when the transcript is healthy. The only branches that mutate bytes fire when `--resume` would otherwise crash on a missing transcript or thrash on a 50+ MiB file the autocompact pass can't survive. The agent's view of the conversation stays consistent — the alternative is a dead pane.

Refactor PR (trait + `ClaudeRecovery` impl) coming on this same branch.

njbrake:
I think to Jerome's point: if something crashes the agent, I'm still fuzzy on why this is something AoE should solve instead of making the coding agent --resume feature more robust. I still lean toward closing this one, but I do want to hear out @BTForIT if you have more thoughts (preferably written by the human and not the AI agent 🙏)

jerome-benoit:
> On the `--resume` semantic concern: the cascade is no-op when the transcript is healthy. The only branches that mutate bytes fire when `--resume` would otherwise crash on a missing transcript or thrash on a 50+ MiB file the autocompact pass can't survive. The agent's view of the conversation stays consistent — the alternative is a dead pane.

Why do you want to handle in a downstream tool an upstream bug on a agent builtin feature? If the agent is not able to resume properly for whatever reason, it's not up to the tool to implement it instead in a resilient manner. Needless to talk about the code to reimplement it in a agent agnostic manner ...

The solution is dead simple to handle crash at resume: fallback to starting the agent without resuming the session if the resuming crash at start. Basically, making the session resuming feature more tunable and handle edge cases: 
+ making the resuming tunable globally or at session level in a UX sensible manner
+ introduce fallback mechanisms at session restart: aoe is already having the code to detect tmux pane without a working agent.

BTForIT:
Jerome, your tunable-resume + fallback-on-crash framing is basically what I'm chasing, and it lines up with where I think this PR should go. The current cascade mutates bytes too aggressively for what most cases need. A per-session knob (recovery_mode: strict, cascade, off) plus the existing pane-alive detection as the fallback substrate is cleaner than what's here today, and it's how I want to reshape this.

Nate, on the "why aoe and not the coding agents themselves" question: my honest take is that the value of aoe shifted a bit when Anthropic shipped their multi-session research preview. The single-agent orchestration story now has a first-party competitor. What aoe still uniquely has is the multi-harness surface — codex / gemini / claude / aider all running side-by-side under one management UX. Recovery is one of the places where "we know each harness's quirks better than the tool itself" actually shows up. "The claude team will fix --resume" works fine for one agent; it doesn't compose when you're running four.

That said, if the consensus is this PR's specific approach is wrong, I'd rather close this and open a smaller one that just adds the recovery_mode knob and the fresh-start fallback Jerome described, and let the heavier transcript-rescue logic live in a fork or sidecar. Happy to go that route.

jerome-benoit:
> Jerome, your tunable-resume + fallback-on-crash framing is basically what I'm chasing, and it lines up with where I think this PR should go. The current cascade mutates bytes too aggressively for what most cases need. 

The PR https://github.com/njbrake/agent-of-empires/pull/1173 implements it properly in an agent agnostic manner. Could you please test it?

njbrake:
I believe that Jerome has resolved the problem in 1173 so I'll close this PR out. Thanks! Btw sorry for the slower reply on the attention/archive/favorite architecture, I am still thinking about it.

PR #1085, opened 2026-05-13
feat: favorite session primitive

BTForIT:
## Description

Wires up the favorite primitive's mutator surface on top of the attention foundation in #1084.

**Stacked on #1084 (`feat/attention-and-snooze`).** This PR's diff overlaps #1084 until that one merges; the favorite-specific commit on top of #1084 is one focused commit (see "Files changed" filtered to that commit).

The `Instance` methods `favorite()`, `unfavorite()`, `is_favorited()`, and the within-tier sort pin already landed in #1084 (favorited rows bubble above non-favorited peers in the same status tier in `SortOrder::Attention`, and `render.rs` styles them with bold + underline + a leading `* ` glyph). This PR adds the only two ways a user can flip the flag: the TUI keybind and the CLI command.

**TUI:** `f` (relaxed hotkeys) / `F` (strict hotkeys) on the home view toggles the cursor's session.

**CLI:**
```
aoe session favorite   <id|title>
aoe session unfavorite <id|title>
```

**Web dashboard parity:** `SessionResponse.favorited` is now surfaced via `Instance::is_favorited()`. The sidebar workspace row renders bold plus a leading amber `*` when any session in the workspace is favorited, matching the TUI render layer.

Help dialog inner height bumped from 44 to 45 to fit the new line.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] I understand the code I am submitting
- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [x] For UI changes: included screenshot or recording — TUI render verification recommended.

## How tested

- `cargo build --release` clean
- `cargo clippy --release --lib -- -D warnings` clean
- `cargo test --release --lib`: 1603 passed, 0 failed
- Manual: toggled the keybind on a populated home view, confirmed toggle status line, persistence across reload, and that `* `+bold marker appears in the row

## AI Usage

- [ ] No AI was used
- [x] AI was used for drafting/refactoring
- [ ] This is fully AI-generated

**AI Model/Tool used:** Claude Code (Opus 4.7)

**Any Additional AI Details you'd like to share:** AI drafted the CLI handler + TUI keybind by mirroring the snooze plumbing landed in #1084; the mutator methods themselves are 3-line wrappers around existing `Instance::favorite()`/`unfavorite()` calls already covered by tests on the foundation branch. Sidebar render parity inspected and matched by hand.

[...]

PR #1084, opened 2026-05-13
feat(tui): attention sort foundation + snooze primitive

BTForIT:
## Description

Foundation PR for the Attention workflow. Adds a `SortOrder::Attention` variant that ranks sessions by status tier (Waiting, Error, Idle, Unknown, Running, Stopped, transient, archived/snoozed) with longest aging first within tier, urgent bias on top, and favorites pinned within tier. All new `Instance` state fields are `Option`, default `None`, so existing `sessions.json` files load unchanged.

Snooze ships as a full primitive: TUI keybind opens a duration picker (15 minutes through 1 week), CLI exposes `aoe session snooze --minutes N` / `unsnooze`, snoozed rows sink to tier 99 with a `z ` prefix and remaining time in the age column, and wake is lazy (no background task, just `is_snoozed()` comparing against now). Urgent rendering is included read side; the write side ships in a follow up.

Archive and favorite primitives (`archive()`, `favorite()`, predicates, and the `archived_at` / `favorited_at` fields) are present on `Instance` so the sort comparator and tests have something to operate on, but no TUI keybind or CLI surface ships archive/favorite mutators in this change. Those land in two separate follow up PRs along with delete redirect, auto unarchive on send, smart state remap, restart-respawn of dead panes, web parity, and urgent write side. Sequencing this way keeps the diff reviewable and lets the foundation merge while the higher-level workflow features are reviewed independently.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] I understand the code I am submitting
- [x] New and existing tests pass
- [x] Documentation was updated where necessary — Docs for new CLI subcommands will follow as separate PR; help.rs reflects current keybindings.
- [x] For UI changes: included screenshot or recording — TUI verification recommended.

`cargo build --release`, `cargo clippy --release --lib -- -D warnings`, and `cargo test --release --lib` all pass (1603 tests). Manual TUI verification is recommended for the snooze duration picker dialog and the visual tier ordering in Attention sort. No documentation changes shipped because the feature surface is internal foundation (sort, predicates, snooze) plus one new CLI subcommand that follows existing patterns; user facing docs ride with the follow up archive/favorite PRs that complete the workflow.

## AI Usage

- [ ] No AI was used
- [x] AI was used for drafting/refactoring
- [ ] This is fully AI-generated

[...]

PR #1086, opened 2026-05-13
feat: archive primitive (TUI z/Z + CLI session archive/unarchive)

BTForIT:
## Description

Wires the archive primitive from #1084 (which added `Instance::archive()` / `unarchive()` / `archived_at` as data-model-only) into a user-reachable surface. The destructive-default flip for `aoe remove` and the TUI `d` dialog that earlier versions of this PR included was reverted, so this branch ships ONLY the archive write path; `aoe remove` and the `d` dialog are unchanged from main.

### TUI

- `z` / `Z` toggles archive on the cursor's session. Archive kills the tmux pane so the row stops animating; the row drops to tier 99 of the Attention sort and renders italic + dim. Pressing `z` / `Z` again unarchives (no respawn; the user starts the session explicitly if they want it back).
- Help dialog gains the `z` / `Z` entry in both keymaps. Dialog height bumped by one row to accommodate.
- After unarchive the cursor re-seats on the just-unarchived session (the row jumps from tier 99 to its real tier, so leaving the cursor at the old index would land it on whatever slid into the slot).
- Toast strings ("Archived: <title>" / "Unarchived: <title>") surface via `Action::SetTransientStatus`.

### CLI

- `aoe session archive [--no-kill] <id>`: sets `archived_at`. Kills the tmux pane by default; `--no-kill` leaves the pane alive (useful for archived but observable terminals).
- `aoe session unarchive <id>`: clears `archived_at`. Does not respawn.

Both subcommands route through `cli::resolve_session`, which was hardened in this PR to detect ambiguous ID prefixes and bail with a list of candidates instead of silently mutating the first match. This also benefits the prior callers (`aoe session worktree`, `aoe add --parent`).

### Not in this PR

- `aoe remove` default behavior is unchanged (still destructive, as on main).
- The `d` dialog default is unchanged.
- The archive write-side for groups (cascade through child instances) is data-model-only via #1084 and not yet wired to a user surface.

### Stacking note

Logically stacks on #1084 (attention sort + snooze foundation) and #1085 (favorite primitive). Now that those are merged, this PR's incremental diff vs main is the +143 LoC archive surface plus the resolver hardening, plus 2 unit tests for `toggle_archive_at_cursor`.

## PR Type

- [x] New Feature

## Checklist

- [x] For UI changes: included screenshot or recording. Manual TUI verification recommended.

## Testing

- `cargo build --release`: clean
- `cargo test --lib`: 1854 pass, 0 fail
- `cargo clippy --lib --tests -- -D warnings`: clean
- `cargo fmt --check`: clean
- `cargo xtask gen-docs`: regenerated `docs/cli/reference.md` with the new `archive` and `unarchive` entries

## AI Usage

**AI Model/Tool used:** Claude Code (Anthropic)

Note: title and body rewritten by Claude via back-and-forth with @njbrake during a PR review session. The original "delete-as-archive" framing predated the maintainer-requested revert of the destructive-default flip; the description now matches what actually ships.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -29,6 +29,8 @@ This document contains the help content for the `aoe` command-line program.
 * [`aoe session unsnooze`↴](#aoe-session-unsnooze)
 * [`aoe session favorite`↴](#aoe-session-favorite)
 * [`aoe session unfavorite`↴](#aoe-session-unfavorite)
+* [`aoe session archive`↴](#aoe-session-archive)
+* [`aoe session unarchive`↴](#aoe-session-unarchive)
 * [`aoe group`↴](#aoe-group)
 * [`aoe group list`↴](#aoe-group-list)
 * [`aoe group create`↴](#aoe-group-create)
@@ -298,6 +300,8 @@ Manage session lifecycle (start, stop, attach, etc.)
 * `unsnooze` — Wake a snoozed session immediately
 * `favorite` — Mark a session as a favorite. Favorited rows pin to the top of their status tier in the Attention sort and render with a leading `* ` glyph plus bold + underline
 * `unfavorite` — Clear the favorite flag on a session
+* `archive` — Archive a session (sinks it to the bottom of the Attention sort). Kills the tmux pane unless `--no-kill` is passed. The worktree, branch, and container are preserved; use `aoe remove` (optionally with `--delete-worktree` / `--delete-branch`) to fully destroy a session
+* `unarchive` — Unarchive a session (restores it to its tier in the Attention sort)
 
 
 
@@ -497,6 +501,34 @@ Mark a session as a favorite. Favorited rows pin to the top of their status tier
 Clear the favorite flag on a session
 
 **Usage:** `aoe session unfavorite <IDENTIFIER>`
+
+###### **Arguments:**
+
+* `<IDENTIFIER>` — Session ID or title
+
+
+
+## `aoe session archive`
+
+Archive a session (sinks it to the bottom of the Attention sort). Kills the tmux pane unless `--no-kill` is passed. The worktree, branch, and container are preserved; use `aoe remove` (optionally with `--delete-worktree` / `--delete-branch`) to fully destroy a session
+
+**Usage:** `aoe session archive [OPTIONS] <IDENTIFIER>`
+
+###### **Arguments:**
+
+* `<IDENTIFIER>` — Session ID or title
+
+###### **Options:**
+
+* `--no-kill` — Skip killing the tmux pane. By default archiving stops the running agent so the row renders as truly parked; pass this to keep the pane alive while still marking the session archived
+
+
+
+## `aoe session unarchive`
+
+Unarchive a session (restores it to its tier in the Attention sort)
+
+**Usage:** `aoe session unarchive <IDENTIFIER>`
 
 ###### **Arguments:**
 
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -35,14 +35,36 @@ use crate::session::Instance;
 use anyhow::{bail, Result};
 
 pub fn resolve_session<'a>(identifier: &str, instances: &'a [Instance]) -> Result<&'a Instance> {
-    // Try exact ID match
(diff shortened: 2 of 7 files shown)


njbrake:

Accepting the archive primitive itself; see #1084 for the reasoning. Non-destructive pane-kill-with-restore is the clearest case for kernel-side ownership since an external tool can't replicate it without owning tmux state behind aoe's back.

One change before merge: please drop the destructive-default flip on `d` and `aoe remove`, or gate it behind a config setting that defaults to today's behavior. The primitive itself is kernel-shaped, but flipping what `delete` means is policy, and I want to make that call separately rather than bundled with the primitive.

Two options:

1. Strip the default change from this PR. Archive ships as `z`/`Z` and `aoe session archive/unarchive` only; `d`/`aoe remove` keep current behavior.
2. Keep the default change but gate it behind a config setting (default = today's behavior, opt-in to archive semantics).

Option 1 is simpler and more conservative. Option 2 preserves your daily workflow. Either works.

Rebase needed.

BTForIT:
Thanks for the clear direction. Picking **Option 1**: stripping the destructive-default flip entirely from this PR. After the strip the archive primitive ships as:

- `z` / `Z` TUI keybinding (toggle archive on cursor)
- `aoe session archive [--no-kill] <id>` and `aoe session unarchive <id>` CLI subcommands
- Help dialog entries for `z` / `Z`
- The render-side italic + dim styling already in the F3 foundation

`d` / `aoe remove` revert to their pre-bf2a661 behavior — destructive cleanup checkboxes restored to their original ticked-by-default state, `--hard` flag dropped, no `archive_as_remove` redirect.

Restacked branch (with the flip stripped) coming once the chain rebuild lands; force-pushing #1087/#1180/#1184/#1244 in the same pass since they sit on top.

njbrake:
_Note: this comment was drafted by Claude via back-and-forth with @njbrake, running as a scheduled review agent. Reasoning and final decisions are his; the prose is Claude's._

**Status:** NEEDS WORK

The core ask is satisfied: `src/cli/remove.rs` has zero diff from `main`, `aoe remove` still defaults to destructive delete, and the `d` dialog still opens `UnifiedDeleteDialog` with the existing behavior. No archive-first flip anywhere. `cargo fmt --check`, `cargo clippy`, and `cargo test --lib` all pass (1864 tests; 3 pre-existing failures). No `#[allow(dead_code)]`.

Two remaining issues:

1. **Stale comment at `src/tui/home/input.rs:850`.** The `z`/`Z` keybind block is annotated with `// (now archives by default but exposes destructive opt ins).` That description was written before the destructive-default flip was dropped; it is now misleading. Update to something like `// separate from d/D (destructive delete, unchanged).`

2. **Em-dashes in new prose comments.** About 35 occurrences across `src/tui/home/input.rs`, `src/session/instance.rs`, `src/session/groups.rs`, and `docs/cli/reference.md`. AGENTS.md prohibits em-dashes in human-authored prose; replace with commas, semicolons, or rephrase.

---
_Generated by [Claude Code](https://claude.ai/code)_

njbrake:
_Note: this comment was drafted by Claude via back-and-forth with @njbrake, running as a scheduled review agent. Reasoning and final decisions are his; the prose is Claude's._

**Status:** NOT READY YET

Same structural situation as #1085: dirty against main, CI not running, waiting for #1084 (and #1085) to land so this branch can be rebased.

The fix commit (`21cfad2f`) addresses all three items from the 20:24 review: em-dash sweep on the archive-specific prose, stale `"now archives by default"` comment at `src/tui/home/input.rs` updated to `"separate from d/D (destructive delete, unchanged)"`, and `docs/cli/reference.md` regenerated via `cargo xtask gen-docs` so the new `aoe session archive` and `aoe session unarchive` entries are reflected. The Docs CI check was failing on the previous tip; with the regen in the fix commit, it should pass once CI runs. No `#[allow(dead_code)]`.

The core ask (OR gate behind a config setting, destructive default unchanged) was confirmed clean in the 20:24 review and is unchanged in the fix commit.

Path forward: wait for #1084 and #1085 to merge, then rebase.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1117, opened 2026-05-14
feat(profile): per-profile host environment variables

BTForIT:
## Description

Adds a top-level `environment` list to `Config` and `ProfileConfig` so each profile can inject env vars onto the host command line of every spawned session. Entry grammar matches the existing `sandbox.environment` field (`KEY=value`, `KEY=$VAR`, `KEY=$$literal`, bare `KEY`).

**Motivation.** Power users running multiple agent accounts (Claude work / personal / different orgs; Codex `CODEX_HOME`; Gemini `GOOGLE_APPLICATION_CREDENTIALS`) want a given AoE profile to attach a session to a specific account without remembering to export the env var in every shell. Original PR (#1117 by @BTForIT) solved this for Claude via a `[claude] config_dir` field; reshaped here into a generic mechanism so it covers the long tail of agents without growing the schema one TOML table per agent.

**Design.**
- `Config.environment: Vec<String>` and `ProfileConfig.environment: Option<Vec<String>>`, deserialized with the same `string_or_vec` / `option_string_or_vec` helpers as `sandbox.environment`.
- New `session::environment::host_environment_prefix` helper turns each entry into a shell-escaped `KEY='value' ` prefix.
- `Instance::build_host_command` prepends the prefix to the existing `AOE_INSTANCE_ID` prefix; ordering means `env`'s last-assignment-wins protects `AOE_INSTANCE_ID` from being clobbered.
- Profile-scope replaces (does not merge) the global list, matching `sandbox.environment` override semantics.
- Sandboxed sessions are untouched, they keep using `sandbox.environment` for the in-container side.
- Values are passed through verbatim. `~` is not expanded; use an absolute path. (Earlier iteration of this PR added tilde expansion; dropped in `d977e9e` to align with `sandbox.environment`.)

**Settings TUI.** New `Host Environment` field under the existing `Session` category, using the same `FieldValue::List` UX as `sandbox.environment`. Description notes that all entry forms resolve to a literal `KEY=value` arg in the spawned process (visible in `ps`); secrets that should stay out of argv belong in `sandbox.environment` instead.

**Docs.** `docs/guides/configuration.md` gains a `## Host Environment` section between `## Session` and `## Worktree`, mirroring the existing `### environment entries` block for `sandbox.environment`.

**Migrating from the original Claude shape:**

```toml
# Was (original PR):
[claude]
config_dir = "/Users/me/.claude-accounts/work"

# Now:
environment = ["CLAUDE_CONFIG_DIR=/Users/me/.claude-accounts/work"]
```

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (\`cargo test --lib\`: 1571 passed; clippy + fmt clean)
- [x] Documentation was updated where necessary (\`docs/guides/configuration.md\` + TUI field description)
- [ ] For UI changes: included screenshot or recording (new settings field follows the existing \`sandbox.environment\` \`FieldValue::List\` pattern; happy to add a recording if you'd like)

## AI Usage

- [ ] No AI was used

Original Claude-shape design and PR by @BTForIT (commit \`9a14785\`). Reshaped to the generic \`environment\` mechanism, plus review-followup cleanup, in three commits I authored with Claude as co-author. The design decisions (generalize instead of one-section-per-agent, replace-not-merge semantics, no tilde expansion to match \`sandbox.environment\`, secrets-go-to-sandbox advisory) were mine; the prose is Claude's. Reviewer questions go to me, not to the AI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: "v1-rust"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Install tmux
         run: |
@@ -51,6 +52,7 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: "v1-rust"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - run: cargo clippy -- -D warnings
 
@@ -69,6 +71,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: "v1-rust"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Check CLI docs are up-to-date
         run: |
@@ -84,6 +87,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: "v1-rust"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Check skill file references valid CLI commands
         run: cargo xtask check-skill
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
-          prefix-key: "v1-rust"
+          cache-bin: "false"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Install tmux
         run: |
@@ -52,7 +52,7 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
-          prefix-key: "v1-rust"
+          cache-bin: "false"
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - run: cargo clippy -- -D warnings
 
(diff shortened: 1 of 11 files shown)


njbrake:

Hi @BTForIT, thanks for the PR. I think the underlying user need (per-profile env isolation for accounts/keys) is real and worth solving, but I'd like us to land it in a more general shape than `[claude] config_dir`.

**The concern with the current shape.** aoe is agent-agnostic. We currently support Claude, Codex, Gemini, OpenCode, Crush, Kiro, Qwen, Pi, Vibe, aoe-agent, plus user-defined custom agents. If we accept `[claude] config_dir`, the precedent says yes to `[codex]`, `[gemini]`, `[opencode]`, and one section per future agent. Each would do the same thing under the hood: set one env var on the host command line at spawn time. That's N agent-specific TOML tables for one mechanism, and the Settings TUI grows a category per agent.

The general pattern already exists in the codebase for sandbox sessions: `sandbox.environment` is a `Vec<String>` of `KEY=value` / `KEY=$VAR` / `KEY=$$literal` / bare `KEY` entries. The natural symmetric move is the same thing for host sessions.

**What I pushed.** Three follow-up commits reshape the PR around `Config.environment: Vec<String>` (global) and `ProfileConfig.environment: Option<Vec<String>>` (per-profile, replaces global). A new `host_environment_prefix` helper in `session::environment` turns each entry into a `KEY='value' ` prefix (via `shell_escape`) and `build_host_command` prepends it. Sandboxed sessions are untouched, they keep using `sandbox.environment`. The Settings TUI gets a `Host Environment` field under the existing `Session` category, identical UX to `sandbox.environment`.

The final commit (`d977e9e`) drops tilde expansion on entry values to keep behavior consistent with `sandbox.environment`, which doesn't do path-aware magic either. Users who want home-relative paths can pass an absolute path or do the home-resolution in their agent wrapper.

Your stated workflow continues to work without losing any ergonomics:

```toml
# Was:
[claude]
config_dir = "~/.claude-accounts/work"

# Now:
environment = ["CLAUDE_CONFIG_DIR=/Users/btforit/.claude-accounts/work"]
```

And the same mechanism naturally serves Codex (`CODEX_HOME`, `OPENAI_API_KEY`), Gemini (`GOOGLE_APPLICATION_CREDENTIALS`), or any future agent without growing the schema.

**Stats.** Net -26 LOC across 8 files. 1571 lib tests pass, clippy clean, 8 dedicated unit tests cover the new helper.

**End-to-end verification.** Built `target/dev-release/aoe`, set up an isolated `HOME` + a fake `claude` shim that dumps its env to a marker file, and confirmed:

| Behavior | Result |
|----------|--------|
| Profile literal `KEY=value` arrives in spawned env | ✓ |
| Profile replaces global (not extend) | ✓ (`X_GLOBAL_ONLY` absent under profile scope) |
| Profile override wins on shared key | ✓ |
| `KEY=$VAR` resolves from host env at spawn | ✓ |
| `KEY=$$literal` escape | ✓ (emitted `$keep-literal` verbatim) |
| Existing `AOE_INSTANCE_ID` prefix intact | ✓ |

Let me know how you want to proceed.

PR #1125, opened 2026-05-14
fix: tighten codex status detection

microHoffman:
## Description
Fixes Codex session status detection false positives and false negatives from pane text parsing.

Previously, Codex sessions could be misclassified because the detector used broad text matches across captured tmux output. That caused false `Running` states when old assistant prose mentioned words like `working`, `esc to interrupt`, or spinner characters. Tightening those checks then exposed false `Waiting`/`Idle` states when Codex was actually running but the live status line was bullet-prefixed, embedded the interrupt hint, or had verbose output below it.

This PR changes Codex detection to use live Codex status-line shapes and current-work-block scanning.

Fixed cases:
- prose mentioning `working`, `esc to interrupt`, or spinner chars no longer reports `Running`
- `• Working (... esc to interrupt)` reports `Running`
- `• Running command: ... esc to interrupt` stays `Running` through verbose output
- `• Starting MCP servers (... esc to interrupt)` stays `Running` through verbose output
- completed `Worked for ...` blocks stop stale running markers from leaking forward
- finished prompt/approval/radio input states remain `Waiting`

No user-facing docs were updated because this is internal status-detection behavior.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Testing
Ran:

```bash
cargo test tmux::status_detection::tests::test_detect_codex_status --lib
cargo test tmux::status_detection::tests --lib
cargo fmt
```

Added regression tests for:

- interrupted prompts
- spinner/prose examples
- completed turns
- MCP startup
- verbose command output
- request-user-input prompts

Local note: full cargo test --lib initially failed on two unrelated git::worktree tests because git-daemon was missing from the local setup. Installing Fedora's git-daemon package resolved that local setup issue.

## AI Usage

- [ ] No AI was used

AI Model/Tool used:
Codex / GPT-5.5

The patch was implemented with Codex and reviewed iteratively against real pane dumps and targeted regression tests.

NOTE:

### DIFF
--- a/src/tmux/status_detection.rs
+++ b/src/tmux/status_detection.rs
@@ -341,15 +341,21 @@ pub fn detect_codex_status(raw_content: &str) -> Status {
         return Status::Waiting;
     }
 
-    if last_lines_lower.contains("esc to interrupt")
-        || last_lines_lower.contains("ctrl+c to interrupt")
-        || last_lines_lower.contains("working")
-        || last_lines_lower.contains("thinking")
+    if codex_has_live_running_status(&non_empty_lines) {
+        return Status::Running;
+    }
+
+    if codex_current_block_lines(&non_empty_lines)
+        .take(10)
+        .any(codex_line_starts_with_activity)
     {
         return Status::Running;
     }
 
-    if has_any_spinner(&lines) {
+    if codex_current_block_lines(&non_empty_lines)
+        .take(10)
+        .any(codex_line_has_activity_spinner)
+    {
         return Status::Running;
     }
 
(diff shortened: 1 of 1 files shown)


njbrake:
@microHoffman this would be a bit of a bigger overhaul than this small fix, but what do you think about adding hooks usage for Codex, similar to what we do for Claude Code already? These are the codex docs: https://developers.openai.com/codex/hooks

At least for claude code, once we switched to hooks the status detection corner cases basically dissappeared

microHoffman:
@njbrake thanks for the feedback:) it really feels like current status detection for codex is very hacky and prone to breaking in the future, so if there is a reliable way how to do that via hooks, i think it's the right direction:) i will have a look on that

so far i was just fixing the cases where i saw that the status was inferred incorrectly for the codex

Issue #1132, opened 2026-05-15
Cockpit web: page reload re-types entire history, no persisted state cache + smooth=true default for historical text


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

On mobile (Chrome iOS / Android) through a Cloudflare tunnel, every page reload of the cockpit view rebuilds the transcript message-by-message from scratch. The user watches the full history type itself out before reaching the latest message they actually wanted to see. On a long session, this can be 5-15 seconds of unusable UI on every reload.

Two compounding factors:

1. The in-memory `stateCache` that backs cross-mount hydration is dropped on full page reload. Reload starts from `lastSeq = 0`, fetches every persisted frame, and replays them through the reducer.
2. The markdown renderer pipes restored agent messages through the same char-budget smoothing it uses for live streaming, so each historical message types out instead of appearing complete.

Both together produce the visible regression: 200+ activity rows render through a typewriter animation that was designed for live streaming, and the user has to wait for it to finish before they can scroll to the bottom.

### `stateCache` is in-memory only

`web/src/hooks/useCockpit.ts:39-53`:

```ts
// LRU-capped module cache keyed by cockpit session id. Survives
// component unmount within the same page lifetime so the user can
// navigate between cockpit sessions (or away from the dashboard and
// back) and keep seeing the in-memory transcript while the
// WebSocket reconnects in the background. Lost on full page reload
// by design.
```

`stateCache: Map<string, CockpitState>` is module-scoped. Survives component remount within the same tab lifetime; gone on `location.reload()` / app-relaunch / tab close.

### Hydration falls back to empty on reload

`web/src/hooks/useCockpit.ts:95-98`:

```ts
function initialState(sessionId: string | null): CockpitState {
  if (!sessionId) return emptyCockpitState();
  return cacheGet(sessionId) ?? emptyCockpitState();
}
```

On a fresh module instance (page reload), `cacheGet` returns `undefined`, so the reducer starts from `emptyCockpitState()` with `lastSeq = 0`.

### Full replay from seq=0

`web/src/hooks/useCockpit.ts:281-329`:

```ts
const since = Math.max(0, lastSeqRef.current - REPLAY_OVERLAP);
const res = await fetch(
  `/api/sessions/${encodeURIComponent(sid)}/cockpit/replay?since=${since}`,
);
// ... dispatch({ kind: "frames", frames });
```

[...]

Issue #1131, opened 2026-05-15
Web: passphrase re-login on every mobile/Cloudflare IP change, replace strict IP binding with opt-in "Remember this device"


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

On a phone (Chrome iOS / Android) accessing the cockpit web UI through a Cloudflare tunnel: the passphrase prompt reappears frequently, well before the 30-day session lifetime should have elapsed. The user re-enters it, gets in, then sometimes loses access again within hours.

The auth token in `localStorage` survives (PWA stays logged in at the token layer). What expires is the second-factor passphrase session. Re-entering it is friction that compounds with the WS reconnect friction in the sibling issue: switch apps, return, lose live state, and on the way back through, also re-type the passphrase.

### Passphrase login session is IP-bound

`src/server/login.rs:31-34`:

```rust
struct LoginSession {
    expires_at: Instant,
    ip: IpAddr,
}
```

`src/server/login.rs:116-138`:

```rust
pub async fn validate_session(&self, session_id: &str, client_ip: IpAddr) -> bool {
    if session_id.is_empty() {
        return false;
    }

    let mut sessions = self.sessions.write().await;
    let Some(session) = sessions.get_mut(session_id) else {
        return false;
    };

    if Instant::now() > session.expires_at {
        sessions.remove(session_id);
        return false;
    }

    if session.ip != client_ip {
        return false;
    }

    // Sliding window: extend expiry on each valid access
    session.expires_at = Instant::now() + SESSION_LIFETIME;
    true
}
```

The IP comparison at line 131 is strict equality. Any change in the resolved client IP causes `validate_session` to return `false`, which routes the user back to the login page.

### `resolve_client_ip` uses `cf-connecting-ip` behind the tunnel

`src/server/auth.rs:42-62`:

```rust
let socket_ip = socket_addr.ip();
if socket_ip.is_loopback() {
    if let Some(cf_ip) = headers.get("cf-connecting-ip") {
        if let Ok(ip_str) = cf_ip.to_str() {
            if let Ok(ip) = ip_str.trim().parse::<IpAddr>() {
                return ip;
            }
        }
    }
    ...
}
socket_ip
```

Behind Cloudflare, every request carries `cf-connecting-ip` set to the real client's public IP. Each request the dashboard makes ends up bound to whatever IP the device had at that moment.

[...]

Issue #1130, opened 2026-05-15
Cockpit web: WS never auto-reconnects after mobile background or Cloudflare idle timeout, useCockpit onclose has no retry path


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

On a phone (Chrome iOS / Android) accessing the cockpit web UI through a Cloudflare tunnel: every time the user switches apps and comes back, the cockpit WebSocket is closed and the UI stays in a "disconnected" state. The page does not auto-reconnect, and the indication that anything is wrong is weak. The user has to manually refresh the page to recover the live stream.

This is a recurring, every-time-you-switch-apps experience on mobile, not a one-off. Mobile browsers aggressively close WS connections when the tab backgrounds; Cloudflare's WebSocket idle timeout (100s by default for free tunnels) closes them on the server side too if traffic dries up. Either path leaves the cockpit WS dead with no client-side recovery.

### `useCockpit` has zero reconnect logic

`web/src/hooks/useCockpit.ts:395-398`:

```ts
ws.onclose = () => {
  statusRef.current = "closed";
  setStatus("closed");
};
```

That's the entire close handler. No retry, no backoff, no timer, no `visibilitychange` listener. The effect at `useCockpit.ts:340-447` only re-opens the WS when its deps change (`sessionId` or `fetchReplay`), which in practice means "when the user switches sessions or hard-refreshes." A simple disconnect-then-reconnect cycle never runs again until the user navigates away.

Compare `useTerminal` (`web/src/hooks/useTerminal.ts:46-606`) which has the full machinery: exponential backoff (1s, 2s, 4s, 8s, 16s, 30s cap), a `retryCount` state, a `reconnecting` state, a `MAX_RETRIES` cap, and a server-signalled "stop retrying" close code (`4001 / pty_dead`, #1107). The cockpit WS hook predates that work and never got the same treatment.

### No page-visibility hook anywhere

`grep -r "visibilitychange\|document\.hidden" web/src/` returns zero hits. The mobile-keyboard hook (`web/src/hooks/useMobileKeyboard.ts`) listens for layout signals but not visibility. Net: when the tab returns to the foreground, nothing actively kicks the WS or session poller; both rely on whatever the browser does to suspended JS.

### Banner is correct but indirect

[...]

Issue #1129, opened 2026-05-15
Cockpit web: Enter on mobile sends prompt instead of newline, no isMobile branch in composer keydown handler


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

On mobile (iPhone Safari, Android Chrome) opening the cockpit web UI and typing in the composer, pressing the on-screen keyboard's return key sends the prompt instead of inserting a newline. Every multi-line message becomes a partial message accidentally fired the moment the user reaches for a line break.

Standard mobile chat UX (WhatsApp, Slack, Discord, iMessage, ChatGPT/Claude.ai) does the opposite: return on mobile keyboards is a newline; the explicit Send button (visible at all times) is the only way to dispatch. AoE inherits assistant-ui's desktop-keyboard convention without an `isMobile` carve-out, so mobile users hit this every prompt.

### Composer Enter handler

`web/src/components/cockpit/Composer.tsx:281-319`:

```tsx
<ComposerPrimitive.Input
  ref={taRef}
  rows={2}
  placeholder={...}
  onInput={onInput}
  onKeyDown={(e) => {
    // Intercept Enter while the agent is running so the
    // textarea can queue follow-ups; ComposerPrimitive.Input
    // hard-blocks Enter when `thread.isRunning && !queue`
    // (see ComposerInput.js in @assistant-ui/react). Without
    // this the textarea would do nothing on Enter mid-turn.
    // Capture-phase so our handler wins over the primitive's
    // bubble-phase listener.
    if (
      !turnActive ||
      e.key !== "Enter" ||
      e.shiftKey ||
      e.ctrlKey ||
      e.metaKey ||
      e.nativeEvent.isComposing
    ) {
      return;
    }
    e.preventDefault();
    e.stopPropagation();
    void sendFromTextarea(taRef, composerRuntime, enqueuePrompt);
  }}
  ...
/>
```

Custom handling fires only when `turnActive` (queueing mid-turn). When NOT turnActive, every other Enter falls through to assistant-ui's default `ComposerPrimitive.Input`, which interprets unmodified Enter as Send. There is no `isMobile` branch.

### Mobile detection already exists

`web/src/hooks/useMobileKeyboard.ts:44-47`:

```ts
export function useMobileKeyboard() {
  const [isMobile, setIsMobile] = useState(() =>
    typeof window !== "undefined" &&
[...]

PR #1137, opened 2026-05-15
Cockpit polishing 7: state persistence, WS auto-reconnect, mobile Enter, /clear palette, device binding

Seluj78:
## Description

Bundle of five cockpit polish fixes covering the cluster of mobile / Cloudflare-tunnel rough edges in #1132, #1130, #1129, #1128, and the bigger security work in #1131.

**#1132 — Persist cockpit state across reload + drop default smoothing**
- Mirror per-session `CockpitState` into `localStorage` under `aoe:cockpit-state:v1:<id>` with a 7-day TTL sweep. Reload hydrates from cache; the WS resumes from `?since=<lastSeq>` instead of typewriting the entire transcript.
- Flip `Markdown` default `smooth=true → false`. Only the live streaming tail (assistant message with `status.type === "running"`) keeps the typewriter pacing.

**#1130 — Cockpit WS auto-reconnect + server keepalive**
- Frontend: exponential backoff (1s → 30s cap, 7 attempts), `visibilitychange` / `online` / `pageshow` reset the envelope and dial immediately. Banner copy includes the retry count + countdown; a manual `Reconnect` button surfaces after the envelope exhausts.
- Server: 30s `Message::Ping` on `cockpit_ws`, 90s pong-idle reaper.

**#1129 — Mobile composer Enter inserts newline**
- Touch-primary device (`(pointer: coarse) AND NOT (any-pointer: fine)`) + plain Enter inserts a newline; explicit Send button dispatches. Desktop / iPad-with-Bluetooth-keyboard keep Enter-to-send. Decision matrix extracted as a pure helper for unit tests.

**#1128 — `/clear` keeps the slash palette populated**
- Reducer no longer wipes `availableCommands` / `availableModes` / `currentModeId` on `SessionCleared`. `claude-agent-sdk` caches `supportedCommands` at Query init and does not rotate them on `/clear`, so the prior over-clear from #1101 strictly lost information.

**#1131 — Device-bound login session + step-up auth + new-login push**
- `LoginSession` now stores `binding_hash: [u8; 32]` (SHA-256 of the client's 32 random bytes) + an optional `elevated_until`. IP is no longer consulted, mobile rotation stops logging users out.
- New `POST /api/login/elevate` opens a 15-minute step-up window required by terminal attach, cockpit prompt/cancel/approval/spawn, session create/delete/rename/send, file mutations (`git/clone`, project ops), and push subscribe/unsubscribe (an attacker could otherwise redirect approval notifications).
- Auth middleware now enforces device binding on every authenticated REST + WS request. WS subprotocol parsing recognises `aoe-token.<token>` + `aoe-device.<secret>` prefixed values (legacy bare token still accepted for in-flight PWA tabs).
- Frontend `deviceBinding.ts` generates the secret via `crypto.getRandomValues`, persists in `localStorage`, ships on every same-origin REST request via `X-Aoe-Device-Binding` and on every WS via `aoe-device.<secret>` subprotocol.
- `ElevationPrompt` modal pops on `403 elevation_required`. Sensitive WS upgrades preflight `/api/login/status` so the prompt fires even though browser WebSocket APIs hide the 403 body.
- Fire-and-forget push notification to existing subscribers on every successful new login (defends against shoulder-surf + URL share; the only attack neither binding nor step-up can prevent).
- Plan validated through a 2-round multi-LLM debate (gemini-3.1-pro, gpt-5.5, grok-4.3) plus a final review pass; all reviewer findings addressed.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Refactor
- [x] Documentation

## Checklist

- [x] New and existing tests pass (160 Rust + 217 vitest)
- [x] Documentation was updated where necessary (`docs/cockpit.md`, `docs/guides/remote-phone-access.md`)

## AI Usage

**AI Model/Tool used:**
Claude Code (Opus 4.7, 1M context). Plan and design for #1131 vetted via a multi-LLM debate (gemini-3.1-pro-preview, gpt-5.5, grok-4.3) followed by a final review pass with the same models; all findings applied before merge. Other fixes were drafted by Claude with human-confirmed plans.

Issues were investigated and drafted in collaboration with @Seluj78; idea and decisions are theirs. Each fix went through a plan / go-no-go cycle before implementation, and the security work in #1131 specifically went through an explicit debate-to-consensus loop with three independent models before any code was written.


docs/guides/remote-phone-access.md
```diff
@@ -19,6 +19,9 @@ You're in. Tap **Share → Add to Home Screen** (iOS) or **three-dot menu → In
 
 - **HTTPS end-to-end** via Tailscale or Cloudflare.
 - **Two factors**: the auth token embedded in the QR URL, plus the passphrase typed on the login page. Either alone is useless.
+- **Device-bound login session.** Each browser generates a high-entropy secret (`crypto.getRandomValues`) on first load and persists it in `localStorage`. After a successful passphrase login, every authenticated request to that browser must carry both the `aoe_session` cookie AND the device-binding secret. A stolen cookie alone is therefore not enough; the attacker also needs the binding secret. The session is no longer tied to your public IP, so mobile network rotation (Wi-Fi to cellular, Cloudflare CGNAT churn, iCloud Private Relay, VPN reconnect) does not log you out. Clearing site data or reinstalling the PWA generates a new secret and requires re-entering the passphrase once.
+- **Step-up confirmation for high-risk actions.** Opening a terminal, running a cockpit command, resolving an approval, or creating/deleting a session prompts for the passphrase again if it has been more than 15 minutes since the last confirmation. Read-only browsing (session list, replay, status polls) is never gated.
```
njbrake: Hmmm I go back and forth on this one. On one hand, I absolutely want to be secure. But on the other hand, if I need to enter a password every time I open the web app from my phone, that might be a bit overkill maybe, if the user already has authenticated their phone? Is there some middle ground

njbrake:
@Seluj78 I'm still reviewing the code but maybe a description of the user flow would help here. I am 100% on board with most of these changes: we absolutely need to fix the whole 'user being constantly logged out' issue. 

What I'm less sure about is how to properly gate the terminal/agent commands. Like on my phone currently I connect to my aoe server using termius/ssh, and I don't need to do any re-auth after taht first connection. If we are HTTPS encrypted and require the password and token in order to log in that first time, maybe we just have something to log the user out if they haven't used the connection in a day? Or something a bit more lenient than 15 minutes

Seluj78:
Hey @njbrake , no worries, you're totally right.

So, this PR is mostly issues I've noticed while using the new cockpit through cloudflare on my phone yesterday for an hour or so. Every time I locked the phone or switched apps, it disconnected (ws auto reconnect for that) and regularly asked me for the passphrase again. by regularly I mean every minute or so depending on the refreshes.

So I wrote it down and then opened an issue. I then asked `Claude` to plan for a fix and asked it to `debate using consult-llm` (github.com/raine/consult-llm) to try and get the best implementation possible.

Maybe I didn't explain the problem and user stories correctly, because what you're saying makes sense, but I just want to long live the passphrase and avoid storing it for tool long and being too lineant/relaxed on the device fingerprint used to know if should ask the passphrase

njbrake:
@Seluj78 all good! I went ahead and stripped out the extra auth layer, so it will only lock you out if you're inactive for 24 hours. This is a great PR, thank you for pulling it together.

PR #1141, opened 2026-05-15
Add Codex hook-based status detection

### LINKED ISSUE #1126: Switch Codex status detection to hooks-based signals
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake while reviewing #1125. The reasoning and decisions are his; the prose is Claude's._

## Context

Codex turn state is currently inferred by parsing tmux pane text in [\`detect_codex_status\`](https://github.com/njbrake/agent-of-empires/blob/main/src/tmux/status_detection.rs). #1125 tightened the parser substantially (bullet-prefixed live status lines, scoping to the current turn block, stopping at \`─ Worked for ... ─\` dividers), but the approach is fundamentally brittle: it depends on Codex's TUI shape, which can shift between releases.

Codex now ships a hooks system: https://developers.openai.com/codex/hooks

This is the same pattern we already use for Claude Code (events written to a file on disk, status read directly without scraping the pane). On the Claude side, switching to hooks made the corner cases in pane parsing mostly disappear.

## Proposal

Wire Codex hooks the same way we wire Claude Code hooks:

1. Register the appropriate Codex hooks at session creation time (likely alongside the existing \`src/agents\` registration path).
2. Write turn-state events to the per-session state directory.
3. Have the dispatcher in \`src/tmux/status_detection.rs\` prefer the hook signal when present, falling back to the existing pane parser when it isn't reachable (first few seconds of a session, custom \`--cmd\` wrappers, \`docker exec\` into a user-managed container we didn't provision, older Codex versions, etc.).

## Notes

- This supersedes the pane-parsing path landed in #1125 for the happy case, but the parser should stay as a fallback for the cases above.
- Worth checking what events Codex's hook spec exposes (turn start / turn end / waiting-for-input / approval-prompt) and whether they map cleanly to \`Status::Running\` / \`Status::Waiting\` / \`Status::Idle\`.
- Author of #1125 (@microHoffman) expressed interest in looking at this; coordinate so we don't duplicate work.

microHoffman:
## Description
Closes #1126.

This adds hooks-based status detection for Codex while keeping the existing tmux pane parser as the fallback path.

Implemented:
- Registers Codex hook events for session start, prompt submit, tool use, permission requests, and stop events, mapping them to AoE status files.
- Writes Codex hooks into the `[hooks]` table in `config.toml` instead of using the Claude-style JSON hook installer.
- Preserves existing user hooks and `[hooks.state]` trust data when installing or refreshing AoE hooks.
- Resolves Codex config paths through `CODEX_HOME` for host sessions, profiles, sandbox environment, and uninstall cleanup.
- Wires hook installation for host sessions, sandboxed sessions, and custom wrapper agents that are detected as Codex.
- Updates the hook install dialog and docs to mention Codex hook trust in `/hooks` and fallback to pane-based status detection.

Verification:
- `cargo test codex`
- `cargo fmt --check`

Known limitations and discussion points for this draft:
- When a Codex conversation is interrupted or aborted mid-run, for example by pressing `Esc`, Codex does not appear to trigger the `Stop` hook (mentioned here: https://github.com/openai/codex/issues/22858 ). In that case AoE can keep showing `Running` even though the session is no longer active. Possible fixes could be some hybrid combination with check from the pane parser along the hooks-based approach, or a richer status stream from the Codex App Server.
- When Codex asks for user input through the radio-style choice UI, no hook appears to fire, so AoE can keep showing `Running` even though user action is needed. A possible workaround is the direct Codex App Server approach discussed here: https://github.com/openai/codex/issues/19328#issuecomment-4416843457. That likely needs bigger AoE architecture decisions before implementation.
- Permission requests work better with this patch because Codex triggers `PermissionRequest`; AoE maps that to `Waiting`.
- If the user already has `features.hooks = false` in `.codex/config.toml`, this patch respects that and skips installing AoE hooks, so status detection falls back to tmux pane extraction. Open question: should AoE silently treat that as an opt-out, prompt the user, or offer to enable hooks automatically? I leaned toward respecting the explicit Codex setting for now.

Suggested merge path:
This is not a 100 percent complete solution for every Codex waiting state, but it does improve the normal hook-supported path and preserves the existing fallback behavior. I think it may make sense to merge this first as a narrower improvement, then follow up after syncing on whether a more complete solution should integrate the Codex App Server or wait for additional Codex hook events.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:**
OpenAI Codex

Codex committed the staged implementation, made a small clippy-driven API cleanup before committing, rebased the branch onto current `main`, ran focused verification, and drafted this PR body.

**NOTE:**


microHoffman:
@njbrake   A few open questions about current state of affairs:

1) **Interrupted / aborted Codex runs:** when a conversation is interrupted mid-run, for example with `Esc`, Codex does not seem to trigger the `Stop` hook. Because of that, AoE can keep showing the session as `Running`
even though Codex is no longer doing anything.

2) **User-input prompts:** when Codex asks for user input through the radio-style choice UI, there also does not seem to be a hook event. AoE stays on `Running` even though the session is actually waiting for
user action. Command permission prompt is detected as "Waiting" succesfully though, because in this case Codex triggers the `PermissionRequest` hook.

**Possible fixes for case 1 & 2:**
a) wait for/add Codex hook events for aborted turns and user-input prompts, however it's unknown when and if this will actually happen
b) use some kind of hybrid approach of hooks together with pane parsing for extracting session state correctly in all cases
c) integrate with the Codex App Server for more capabilities, as mentioned here: https://github.com/openai/codex/issues/19328#issuecomment-4416843457 . However integrating Codex App Server might have some other consequences and more work, so we should probably sync before I would start looking into this more.

There is also one more user-preference question, not sure what do you think is the best to do here?:

3) **`features.hooks = false`:** this PR currently treats an explicit Codex `features.hooks = false` as an opt-out and falls back to tmux pane parsing. Should AoE keep doing that silently, prompt the user, or offer to
enable hooks automatically?

Generally this PR does not solve 100% of Codex status cases as mentioned above with case 1 & 2, but it adds the hooks support for Codex and keeps the old pane parser fallback. Is it worth merging this partial improvement first, then following up with a more complete solution after, especially if we would decide to go with Codex App Server? I am also thinking if the hybrid approach might be the best to do for now. Thanks

njbrake:

Thanks for the thorough writeup and for separating out the open questions.

**On (1) and (2):** Your read of the upstream situation matches mine. Issues #19328, #19921, #21753 (all open, no commitment) and #22858 (Stop on Esc, also open) confirm there's no near-term hook fix. So waiting on upstream isn't realistic, and the App Server route is a bigger architectural call I'd rather not couple to this PR.

I want hooks to be the primary path here, they're the right design and clearly more reliable than pane parsing for the cases they cover. But I'd like to pair them with the existing pane parser as a fallback so we catch the states hooks miss. The existing `detect_codex_status` in `src/tmux/status_detection.rs` already handles both of the gaps you flagged: it returns `Waiting` on the `❯ 1.` / `❯ 2.` choice menu and on `"enter to submit answer"`, and it returns `Idle` once the running signals disappear after an Esc interrupt. Since the parser already does that work today, I'd rather not lose it for users who install hooks.

Could we land the hybrid in this PR? Concretely, when hooks are installed:

1. Use the hook-derived status as the primary signal.
2. Run `detect_codex_status` in parallel, but only let it override the hook status in a narrow set of cases: pane parser says `Waiting` while hooks say `Running` (covers the choice UI), and pane parser says `Idle` while hooks say `Running` (covers Esc-interrupted runs).
3. Anything else, hooks win.

Scoped tight like that, the override can be deleted cleanly when upstream ships the missing events, and we don't undo the reliability gains for the cases hooks do handle. Happy to talk through it if you think this expands the PR too much, I realize it's already a sizable change.

**On (3):** Keep the current behavior. Silently respecting `features.hooks = false` is right, the right place to surface "in fallback mode" is something passive like `aoe doctor`, not the install dialog. Not blocking on that.

Two unrelated nits before merge: revert the `src/tui/home/render.rs` line 695 change (`" ".repeat(LAST_ACTIVITY_RIGHT_MARGIN)` to `" ".to_string()` is unrelated to Codex hooks), and flag in the commit body that `SessionConfig::default()` now returns `agent_status_hooks = true` instead of `false`, that's a behavior change worth documenting.

microHoffman:
@njbrake thanks for your response, I agree that the hybrid approach is the way to go now:) i will update the PR soon... 

I think implementing the Codex App Server just to handle these 2 cases would be overkill for now. We can see if in the future we might need to use the Codex App Server also for something else, and if yes, only then start researching possible integration to AoE:)

microHoffman:
@njbrake I have pushed the changes for the hybrid approach. I have been testing it today and so far seems to work good with Codex, so I marked it as ready for review:)

PR #1155, opened 2026-05-15
fix(cockpit): preserve queued prompts on reconnect race

### LINKED ISSUE #1144: Cockpit: queued follow-up prompts can be silently dropped on reconnect (client-only state with optimistic clear)

## Symptom

Queue a follow-up prompt while a turn is in progress, then leave the app (close the tab, navigate away, background the mobile PWA). Come back later. The queued message is gone, never appears in the transcript, and the QueuedPromptsStrip is empty.

## What's actually happening

The queue is purely client-side React-reducer state with no server-side counterpart. It's persisted to `localStorage` as part of the full `CockpitState` snapshot, which usually survives a tab close, but multiple paths can silently drop it on the way back in.

### Architecture (current)
- Queue lives in `CockpitState.queuedPrompts` (`web/src/lib/cockpitTypes.ts:266`).
- Reducer mutations: `enqueue_prompt` / `dequeue_prompt` / `edit_queued_prompt` / `clear_queue` (`web/src/hooks/useCockpit.ts:293-317`). The server-frame reducer (`applyEvent` in cockpitTypes.ts) never touches `queuedPrompts`.
- Persistence: every state change runs `cacheSet(sessionId, state)` which writes the full state to `localStorage['aoe:cockpit-state:v1:<id>']` (`useCockpit.ts:400-402, 199-208, 73-83`). 7-day TTL.
- On remount: `dispatch({ kind: "hydrate", state: cacheGet(sessionId) ?? emptyCockpitState() })` (`useCockpit.ts:522-525`).
- Drain: a useEffect dispatches the queued head (combined or serial) when `!turnActive && workerState=="running" && queue.length>0` (`useCockpit.ts:909-942`).
- Server side has no concept of pending prompts — `src/cockpit/state.rs` tracks approvals, notifications, transcript, but nothing about \"the user has 3 messages they want to send next.\"

### Drop modes I can see in the code

**1. Reconnect race — drain fires before WS is open, optimistic clear loses the message.** (Likely the common case.)

`connect()` at `useCockpit.ts:601-611` awaits `fetchReplay(sessionId)` first, then opens the WS. fetchReplay can dispatch missed frames including `Stopped`, which flips `state.turnActive` to false. The drain effect is dep'd on `state.turnActive` and `state.queuedPrompts`, so it now runs while `statusRef.current` is still `"connecting"` (the WS hasn't opened yet — `setStatus("open")` only fires from the WS `onopen`).

The combined-mode drain at `:921-926` dispatches `clear_queue` *first*, then calls `dispatchPromptNow(combined)`. `dispatchPromptNow` at `:821-826` sees `statusRef.current !== "open"`, dispatches an error banner, and returns without POSTing.

Net: queue cleared locally, error banner shown, message never sent.

**2. Server seq reset — `kind: \"reset\"` wipes the entire state including the queue.**

`fetchReplay` at `:488-490`: if `data.highest_seq < since` (cockpit worker restart, session delete+recreate with same id, daemon restart with seq counter forgotten), it dispatches `{ kind: \"reset\" }` which falls through the reducer's default to `emptyCockpitState()` (`:325`). Queue gone.

**3. Private browsing / storage quota / disabled localStorage.**

`persistState` at `:73-83` swallows quota / private-mode errors silently. The in-memory `stateCache` Map then dies with the JS module on full tab close. Next visit hydrates from `emptyCockpitState()`.

**4. Cross-device / different browser.**

State is per-origin per-browser. Queueing on phone, opening on laptop, queue is gone — no surprise given the architecture, but worth naming.

## What the user expected

When I queue a follow-up, I expect it to be \"in flight\" — guaranteed to fire when the current turn ends, regardless of whether my phone is in my pocket or my laptop is closed. The current shape silently violates that.

## Suggested directions (in increasing scope)

- **Minimal:** in the drain effect (`:909-942`), gate on `statusRef.current === \"open\"` AND don't dispatch `clear_queue` until `dispatchPromptNow` succeeds (move the dequeue to the `.then(...)` branch, not before the await). On failure, keep the queue intact and let the next turn-end retry. This kills mode 1.
- **Better:** on `kind: \"reset\"` (the seq-reset path at `:488-490`), preserve `queuedPrompts` (and possibly `pendingApprovals`) instead of clobbering the whole state. The new server is the same session, so queued user intent should carry across.
- **Right thing:** push the queue to the server. Add `pending_prompts: Vec<String>` to `CockpitState` server-side, expose POST `/cockpit/queue` and DELETE `/cockpit/queue/:id`, drain in the cockpit runner on idle transition. Then the queue survives every tab close, every reconnect race, every device hop, and the user gets the durability they intuitively expect.

I'd vote for the right thing — the optimistic-client model has too many edges to plug — but the minimal fix kills the most common silent-drop without touching the daemon.

njbrake:
## Description

Queued follow-up prompts were silently dropped on reconnect because of a race in the cockpit drain effect. `connect()` in `useCockpit.ts` awaits `fetchReplay(sessionId)` BEFORE opening the WS, and replay can dispatch a `Stopped` frame (flipping `turnActive` to false). The drain effect, depending only on `state.turnActive` and `state.queuedPrompts`, would then fire while the WS was still in `"connecting"`. Under the prior optimistic ordering it dispatched `clear_queue` first and then called `dispatchPromptNow`, which bailed out with an error banner because `statusRef.current !== "open"`. The queue was gone; the message never sent.

This PR is the **minimal patch** described in #1144 (the first of three suggested tiers). The broader server-side queue migration that would move queue ownership off the client is deliberately out of scope here and remains open per the issue's other tiers.

### Changes

1. **Gate the drain effect on WS open.** Added an early-return `if (statusRef.current !== "open") return;` to the drain effect, and added the `status` state value to the effect's deps so transitions re-fire it.
2. **Don't clear the queue before the POST succeeds.**
   - `dispatchPromptNow` now returns `Promise<boolean>` (true on POST success, false on disconnect / non-OK / network error).
   - Combined-mode drain snapshots the ids it's sending and uses a new `dequeue_prompts_by_id` reducer action so prompts the user enqueued during the await are NOT dropped. The queue is only cleared on `res.ok`.
   - Serial-mode drain pops the head only after a successful POST; on failure the head stays and the next `Stopped` retries.
3. **New reducer action `dequeue_prompts_by_id`** for the precise combined-mode clear.

Fixes #1144

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

(No UI changes; behavioural fix to a hook. Verified via vitest: `cd web && npm run test:unit` -> 280/280 passing, including new tests in `useCockpit.queue.test.ts` covering the three drop modes the issue calls out.)

## AI Usage

- [ ] No AI was used

Implemented per the explicit minimal-patch spec in #1144. Tests cover (a) the drain skipping while status is connecting, (b) queue surviving a POST failure, and (c) queue precisely clearing only the snapshot ids when new entries are added during the await.

**NOTE:**

### DIFF
--- a/web/src/hooks/useCockpit.queue.test.ts
+++ b/web/src/hooks/useCockpit.queue.test.ts
@@ -1,4 +1,8 @@
-// Reducer tests for the client-side prompt-queue feature (#1031).
+// @vitest-environment jsdom
+//
+// Reducer tests for the client-side prompt-queue feature (#1031),
+// plus hook-level drain-race regression tests for #1144 (queued
+// follow-ups silently dropped on reconnect).
 //
 // While a turn is running, sendPrompt dispatches `enqueue_prompt`
 // instead of the immediate POST path. The reducer keeps the queue
@@ -6,10 +10,11 @@
 // the QueuedPromptsStrip can render / edit / drop entries before they
 // fire.
 
-import { describe, expect, it, vi } from "vitest";
+import { act, renderHook } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { emptyCockpitState, type QueuedPrompt } from "../lib/cockpitTypes";
-import { cockpitHookReducer, combineQueuedPrompts } from "./useCockpit";
+import { cockpitHookReducer, combineQueuedPrompts, useCockpit } from "./useCockpit";
 
 describe("cockpitHookReducer / queue actions", () => {
   it("emptyCockpitState starts with an empty queue", () => {
@@ -103,6 +108,41 @@ describe("cockpitHookReducer / queue actions", () => {
     expect(s3.queuedPrompts).toEqual([]);
   });
 
+  it("dequeue_prompts_by_id removes only the listed ids, preserving order", () => {
+    const s1 = cockpitHookReducer(emptyCockpitState(), {
+      kind: "enqueue_prompt",
+      text: "first",
+    });
+    const s2 = cockpitHookReducer(s1, {
+      kind: "enqueue_prompt",
+      text: "second",
+    });
+    const s3 = cockpitHookReducer(s2, {
+      kind: "enqueue_prompt",
+      text: "third",
+    });
+    const firstId = s3.queuedPrompts[0]!.id;
+    const thirdId = s3.queuedPrompts[2]!.id;
+    const s4 = cockpitHookReducer(s3, {
+      kind: "dequeue_prompts_by_id",
+      ids: [firstId, thirdId],
+    });
+    expect(s4.queuedPrompts).toHaveLength(1);
+    expect(s4.queuedPrompts[0]?.text).toBe("second");
+  });
+
+  it("dequeue_prompts_by_id with an empty id list is a no-op", () => {
+    const s1 = cockpitHookReducer(emptyCockpitState(), {
+      kind: "enqueue_prompt",
+      text: "first",
+    });
+    const s2 = cockpitHookReducer(s1, {
+      kind: "dequeue_prompts_by_id",
+      ids: [],
+    });
+    expect(s2).toBe(s1);
+  });
+
   it("queue is independent of activity / turnActive state", () => {
     // Enqueue while a turn is mid-flight (turnActive=true, activity has
     // a user_prompt row) and ensure the queue mutation does not clobber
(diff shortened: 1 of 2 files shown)


Seluj78 (approved):
Nice find, thanks!

PR #1156, opened 2026-05-15
fix(wizard): seed yoloMode from profile config on mount

### LINKED ISSUE #1142: Cockpit/web wizard ignores profile yolo_mode_default at startup, sessions launch in default mode

## Symptom

In the TUI new-session dialog, picking a profile whose `[session].yolo_mode_default = true` correctly launches Claude in yolo / auto mode (the dialog's yolo toggle is pre-checked, the resulting session passes `--dangerously-skip-permissions`).

Doing the "same" thing through the **web wizard** (cockpit URL → "New session") silently launches the session in **default permission mode** instead. Claude then prompts on every tool call, which is not what my profile asked for.

## Root cause (per code read)

The web wizard never reads the profile's `yolo_mode_default` at startup, only when the user actively swaps the profile dropdown.

1. `web/src/components/session-wizard/SessionWizard.tsx:74` hardcodes `yoloMode: false` in `initialData`.
2. `SessionWizard.tsx:193-205` (`useEffect` on mount) calls `fetchSettings()` with **no profile arg** — pulls the global config only, and even from that copies just `sandboxImage`. `yolo_mode_default` is never consulted.
3. `web/src/components/session-wizard/steps/AgentStep.tsx:92-125` (`handleProfileChange`) is the *only* place that calls `fetchSettings(profileName)` and applies `yolo_mode_default`. It fires only when the user manually changes the profile dropdown.
4. `AgentStep.tsx:90` hides the profile picker entirely when `profiles.length <= 1`, so single-profile users have no path to ever fire `handleProfileChange`.
5. The wizard then POSTs `yolo_mode: d.yoloMode` (`SessionWizard.tsx:225`) and the backend at `src/server/api/sessions.rs:952` takes the body value at face value.

Net effect: any session created via the web wizard launches with `yolo_mode = false` regardless of the active profile's config.

## Reproduce

1. Set `yolo_mode_default = true` under a profile (or in global config if you only have the default profile).
2. `aoe serve`, open the cockpit URL.
3. New session through the wizard.
4. Observe Claude starts in default permission mode (not auto / `--dangerously-skip-permissions`).

Compare with TUI `n` → new session dialog: same profile, dialog's yolo toggle is pre-checked, session correctly launches with the flag. (`src/tui/dialogs/new_session/mod.rs:353` and `:581-582` are the working reference.)

## Suggested fix

In `SessionWizard.tsx`, on mount and whenever `data.profile` changes (including the implicit default profile), call `fetchSettings(profile)` and seed `yoloMode` (and arguably `sandboxEnabled`, `tool`, `extraEnv`) from the merged profile config — i.e. mirror what `AgentStep.handleProfileChange` does today, but unconditionally at startup so single-profile users and users who never touch the picker get the same defaults the TUI gives them.

njbrake:
## Description

Fixes #1142.

A profile with `yolo_mode_default = true` correctly launches yolo mode from the TUI new-session dialog, but the web session wizard silently ignored it: profiles only ever loaded when the user manually swapped the profile dropdown, and that picker is hidden when `profiles.length <= 1` (single-profile users never trigger it). The wizard's `initialData` hardcoded `yoloMode: false` and the mount effect's `fetchSettings()` call only copied `sandboxImage`, never `session.yolo_mode_default`.

The fix moves the resolved-profile lookup into the wizard's mount effect: it now reads the profile list, picks the prefill profile (or the server's active profile from `/api/profiles`, or falls back to the unresolved global config), calls `fetchSettings(<profile>)`, and dispatches the existing `APPLY_PROFILE_DEFAULTS` action with `yoloMode`, `sandboxEnabled`, `tool`, and `extraEnv` filled in. Explicit prefill values still win over profile defaults so callers that pass `prefill.yoloMode = true` aren't silently overridden.

A new `skipIfDirty` flag on the action prevents a slow `/api/settings` response from clobbering edits the user made between mount and the fetch resolving. The picker-driven path (`AgentStep.handleProfileChange`) leaves the flag false and keeps its existing `window.confirm()` overwrite UX intact.

The reducer + initial state moved into a new `wizardReducer.ts` so the wizard component file stays clean under `react-refresh/only-export-components` and so the reducer is unit-testable without standing up React + the wizard's fetch graph.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

UI change is a behavior-only fix (yolo toggle is now correctly pre-checked on mount when the profile sets `yolo_mode_default`); no visual diff to capture. Verified manually via reasoning + the new vitest reducer tests.

Tested:
- `cd web && npm run test:unit` -> 280 passed (28 files), including 5 new `wizardReducer.test.ts` cases for `APPLY_PROFILE_DEFAULTS`.
- `cd web && npm run build` -> succeeds.
- `cd web && npm run lint` -> remaining errors are all pre-existing (no new errors introduced by this change; the `react-refresh/only-export-components` lint that fires when the reducer is exported alongside the component is the reason for the file split).
- Did not run `cargo test` (no Rust changes). Did not run `npm run test` (Playwright e2e suite); the change is reducer-level and covered by vitest.

## AI Usage

- [ ] No AI was used

Working under auto-mode against the issue body. Followed the routing in #1142 root-cause section verbatim and confirmed it against the `/api/profiles` `is_default` field and `/api/settings?profile=` resolved-config behavior in `src/server/api/system.rs` before changing the wizard.


Seluj78 (changes requested):

Tested locally: created a cockpit web-UI session with **Auto-approve** toggled ON. After page refresh, the cockpit session was still in ACP `default` mode, not auto-approving tool calls.

Root cause is not in this PR. The wizard side is correct: it now sends `yolo_mode: true` to `POST /api/sessions` and the backend stores `Instance::yolo_mode = true` (`src/server/api/sessions.rs:952`). The gap is downstream.

`Instance::yolo_mode` plumbs into tmux-mode spawn through `apply_yolo_mode()` in `src/session/instance.rs:314`, which appends the per-agent flag (e.g. `--dangerously-skip-permissions` for claude per `src/agents.rs:173`) to the agent CLI in `build_host_command` (line 1154). Cockpit-mode does not run that path. The cockpit supervisor spawns the ACP adapter via `agent_registry` (`src/cockpit/agent_registry.rs`) and `acp_client::spawn()` using only `spec.command` + `spec.args`. Grep `is_yolo|yolo_mode` across `src/cockpit/`: zero matches.

End state: the **Auto-approve** label in the wizard advertises a behavior the cockpit path does not deliver.

Request: make Auto-approve flip the cockpit's ACP mode to bypass-permissions on `session/new`. Concretely, when `Instance::yolo_mode == true` and the session is cockpit-mode, send `session/set_mode { mode_id: "bypassPermissions" }` right after the adapter's `session/new` succeeds. The plumbing already exists: `Supervisor::set_mode` at `src/cockpit/supervisor.rs:980` and the ACP-side translation at `src/cockpit/acp_client.rs:1610` (`"bypass_permissions" | "bypassPermissions" => SessionMode::BypassPermissions`). The same change should publish a `CurrentModeUpdate` so the UI mode pill reflects the bypass state immediately, including across reloads.

Adapter-specific CLI flags (e.g. `claude-agent-acp` may accept `--permission-mode bypassPermissions`) would also work but are per-adapter. The `session/set_mode` path is portable across the cockpit's six configured adapters (claude, opencode, gemini, codex, vibe, pi).

This is technically a separate bug from #1142, but the wizard fix in this PR is what surfaces it. Either fold the cockpit-side fix into this PR or land it as an immediate followup before merging this one, since shipping #1142 alone makes the broken behavior more discoverable (single-profile users with `yolo_mode_default = true` now get the toggle pre-checked and will hit this every cockpit session).

PR #1157, opened 2026-05-15
chore(cockpit): add tracing to ACP tool-call and permission handlers

### LINKED ISSUE #1147: Cockpit: investigate whether parallel tool calls are being serialized vs terminal Claude
> Filed by Nathan, with code investigation done by Claude. Words below are AI-drafted unless quoted; I (Nathan) confirmed the perceived slowness by hand.

## Symptom

Subjective: cockpit-mode sessions feel noticeably slower than equivalent terminal-mode sessions running the same Claude model. The hypothesis is that tool calls the model would otherwise dispatch in parallel are being executed serially through the cockpit/ACP path.

## Code reads (what is and isn't disabling parallelism)

I dug through `src/cockpit/` to look for any explicit serialization. What I found:

- **No `disable_parallel_tool_use` flag is being set anywhere.** Searched the whole repo for `parallel`, `parallel_tool_calls`, `disable_parallel`, `max_concurrent`, `--no-parallel`, `concurrent_tools`, `serialize`. Nothing.
- **The cockpit runner itself doesn't serialize.** `src/cockpit/runner.rs:179-183` (`fanout_agent_stdout`) and `:520` forward agent output line-by-line without a per-call gate; `:535-583` `handle_connection` is a non-blocking async loop.
- **The agent is spawned with no extra flags.** `src/cockpit/acp_client.rs:938-1028` invokes `tokio::process::Command::new(&spawn_command)` with `config.spec.args` only; for claude (per `src/agents.rs:169-189`) that's an empty arg list — no `--disable-parallel-tool-use` or equivalent.

So *we* are not disabling parallelism on our side. The remaining suspects are the ACP path itself.

## Likely real bottleneck: approval await blocks the model's turn

`src/cockpit/acp_client.rs:2612-2692` `handle_permission_request`:

```rust
let outcome = match resolve_rx.await {  // line 2669
    Ok(ApprovalResolutionMessage::Decision { decision }) => { ... }
    ...
};
responder.respond(RequestPermissionResponse::new(outcome))
```

Every permission request blocks on `resolve_rx.await` until the user clicks an approval card. The Anthropic API allows multiple parallel `tool_use` blocks per turn, but if any of them require approval, claude-agent-acp issues a `RequestPermissionRequest` per call and the model can't see the next tool result until each is resolved. The runner *does* fan out approvals to the UI as parallel events (mpsc send at `:2657-2660`), so the user sees N cards at once — but practically, clicking through them takes seconds each, and that latency is what makes the turn feel sequential even when the model dispatched in parallel.

## What should actually be checked (a real perf compare, not just code reading)

Code reading can't conclusively answer this. What would settle it:

1. **Run the same prompt twice on the same model**, once in cockpit and once in terminal claude, where the prompt elicits a clear parallel-tool moment (e.g., \"read these 5 files and summarize each\"). Time wall-clock turn duration. If cockpit is materially slower with no approval prompts in the path (yolo mode + auto-approve everything), parallelism is broken somewhere upstream of us.
2. **Check claude-agent-acp's behavior.** Does it forward the model's parallel tool_use blocks as concurrent ACP `tool_call` requests, or does it serialize them into one-at-a-time? This is the most likely place the parallelism is being lost. Ref: https://github.com/zed-industries/claude-agent-acp (or wherever the upstream lives — I didn't fetch).
3. **Tracing.** Add `tracing::info!` at the top of `handle_tool_call_create` / `handle_permission_request` (`acp_client.rs:1859-1932`) to log call ids + timestamps. If parallel tool_use blocks come through as serial RPCs spaced by tool-completion latency, the bottleneck is upstream.

## Suggested next steps

- **Short term:** add the tracing in (3) so we can confirm or rule out from a single session log.
- **If upstream:** file against claude-agent-acp; in the meantime document the gap in the substrate notice (cockpit users on parallel-heavy turns will see slower wall-clock than tmux).
- **If approval-await is the felt slowness:** that's #1038 / approval UX territory and the fix is faster approval handoff (sounds, push, batch-approve) more than tool-execution parallelism.

Marking P1 because perceived slowness vs the terminal substrate is the kind of thing that makes users abandon cockpit mode, and we don't currently know whether the cause is upstream, in our runner, or in the human-in-the-loop approval gate.

njbrake:
## Description

Implements the **short-term** investigative step from #1147: add `tracing::debug!` events at the entry and exit of every `on_receive_request` handler in `src/cockpit/acp_client.rs`, plus an extra event around the `resolve_rx.await` inside `handle_permission_request`. Goal: produce a single session log that can prove or disprove the hypothesis that parallel tool calls dispatched by the model are arriving serially through the cockpit/ACP path.

The traces share a target (`cockpit.acp.tool_dispatch`) so they can be enabled with one `RUST_LOG=cockpit.acp.tool_dispatch=debug` filter without touching the rest of the cockpit logs. Each event carries:
- `handler` (e.g. `read_text_file`, `permission_request`)
- a monotonic `enter_ns` correlation token (ns since process start, captured via a `OnceLock<Instant>`-backed epoch so multiple handlers running concurrently share the same time origin and interleaved entries are visually obvious)
- per-handler context (`path`, `terminal_id`, `tool_call_id`, `command`, etc.)
- `elapsed_ns` on the exit event so per-handler latency drops out of the log without external math

For `handle_permission_request` specifically, the second event fires the moment we begin awaiting the user's decision and the third on `responder.respond(...)`. The issue notes this `await` as the leading suspect for the perceived slowness; isolating it lets the next investigation step quantify human-approval wait time vs everything else.

**No behavior change.** All new code paths are at `debug!` level, so they stay quiet under default logging. The PR intentionally does not close #1147 because the wall-clock cockpit-vs-terminal benchmark with these traces still needs to be run; that's the next step.

Refs #1147

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**Any Additional AI Details you'd like to share:** This PR is the investigative tooling step from #1147; the diff is observability-only (new `debug!` calls and one `enter_timestamp_ns()` helper). Verified with `cargo fmt`, `cargo clippy --all-targets --features serve -- -D warnings` (and again without `--features serve`), and `cargo test --features serve --lib` (2033 passed). The `tests/cockpit_acp_smoke.rs` integration test is unaffected by this PR; on this machine it requires `cockpit-worker/test-shim/node_modules` to be installed and the smoke test pre-existed this branch.

### DIFF
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -9,7 +9,7 @@ Targets use the convention `<module>.<submodule>`. The default filter expands a 
 | Root | Sub-targets | What lands here |
 |------|-------------|-----------------|
 | `agent_of_empires` | (default crate path) | General library code emitting without `target:` |
-| `cockpit` | `cockpit.acp`, `cockpit.supervisor`, `cockpit.event_store`, `cockpit.runner`, `cockpit.acp.stderr` | ACP transport, supervisor lifecycle, event store, runner shim |
+| `cockpit` | `cockpit.acp`, `cockpit.supervisor`, `cockpit.event_store`, `cockpit.runner`, `cockpit.acp.stderr`, `cockpit.acp.tool_dispatch` | ACP transport, supervisor lifecycle, event store, runner shim, per-tool-call entry/exit firehose |
 | `terminal` | `terminal.ws`, `terminal.ws.bytes` | Web terminal WS relay + per-byte firehose (trace) |
 | `auth` | `auth.token`, `auth.middleware`, `auth.rate_limit`, `auth.passphrase`, `auth.device`, `auth.ip` | Token rotate, middleware accept/reject, rate-limit thresholds, login flow |
 | `process` | `process.signal`, `process.tree`, `process.reap`, `process.ppid` | Signal sends, process-tree walks, survivor reap, ppid resolution |
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -9,7 +9,7 @@ Targets use the convention `<module>.<submodule>`. The default filter expands a 
 | Root | Sub-targets | What lands here |
 |------|-------------|-----------------|
 | `agent_of_empires` | (default crate path) | General library code emitting without `target:` |
-| `cockpit` | `cockpit.acp`, `cockpit.supervisor`, `cockpit.event_store`, `cockpit.runner`, `cockpit.acp.stderr`, `cockpit.acp.tool_dispatch` | ACP transport, supervisor lifecycle, event store, runner shim, per-tool-call entry/exit firehose |
+| `cockpit` | `cockpit.acp`, `cockpit.supervisor`, `cockpit.event_store`, `cockpit.runner`, `cockpit.acp.stderr`, `cockpit.acp.tool_dispatch` | ACP transport, supervisor lifecycle, event store, runner shim, per-tool-call entry/exit instrumentation |
 | `terminal` | `terminal.ws`, `terminal.ws.bytes` | Web terminal WS relay + per-byte firehose (trace) |
 | `auth` | `auth.token`, `auth.middleware`, `auth.rate_limit`, `auth.passphrase`, `auth.device`, `auth.ip` | Token rotate, middleware accept/reject, rate-limit thresholds, login flow |
 | `process` | `process.signal`, `process.tree`, `process.reap`, `process.ppid` | Signal sends, process-tree walks, survivor reap, ppid resolution |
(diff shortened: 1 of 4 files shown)


Seluj78 (changes requested):
All of those debug logs should be `trace` level IMO since we're using them to debug something.

Also please add `cockpit.acp.tool_dispatch` as a thing we can choose the log  level for in the settings (tui/toml/web)

PR #1159, opened 2026-05-15
feat: per-session CLAUDE_CONFIG_DIR via account-bound picker rows

BTForIT:
## Description

Lets a session pin its own `\$CLAUDE_CONFIG_DIR` independent of profile or shell environment, so the same project can host multiple Claude sessions running under different Claude configurations side by side.

**Data model**: `Instance.claude_config_dir: Option<PathBuf>` (skipped from serde when None, so existing on-disk sessions round-trip unchanged).

**Discovery**: `src/session/claude_accounts.rs` scans `\$HOME/.claude-accounts/`, surfacing one `ClaudeAccount` per directory that contains `settings.json` (file or symlink), skipping `_shared` and dotted names. Names are pretty-printed via a small acronym table (`forit-main` → `ForIT Main`, `wma-work` → `WMA Work`) so the dashboard and picker speak the same display language.

**Picker UI** (gated by `AOE_CLAUDE_ACCOUNTS_PICKER=1` during rollout): the new-session Tool picker grows one extra row per discovered account beneath the plain `claude` row, labeled `claude => ForIT Main`. Selecting an account-bound row sets `NewSessionData.claude_config_dir`; selecting plain `claude` leaves it `None`. Both the sync (`create_session`) and async (`CreationPoller`) creation paths carry the field through to the persisted `Instance`.

**Spawn injection**: `Instance.build_host_environment_prefix` prepends `CLAUDE_CONFIG_DIR=<expanded>` to the host env prefix so a per-session override beats any profile-level entry with the same key. Skipped for sandboxed sessions because host paths usually aren't valid inside the container; sandbox users should forward the variable explicitly via `sandbox.environment`. Path supports `~` and `\$HOME` expansion against the host env at spawn time.

**Dashboard**: each session row whose `claude_config_dir` is set now shows a dimmed `@ForIT Main` suffix, derived from the same acronym table as the picker. Reads independently of the env flag so historical sessions still get their label after the flag is later disabled.

## Out of scope (deferred to a follow-up PR)

- In-session Shift+A account swap (open picker on an attached session, update `Instance.claude_config_dir`, restart with new env).

## PR Type

- [x] New Feature

## Checklist

## Testing

- `cargo test --no-default-features --lib` → 1639 pass (12 new tests in this stack: 8 in `session::claude_accounts`, 4 in `session::instance`, 3 in `tui::home::render`).
- Integration suites: `session_lifecycle`, `config_merge`, `hooks_config`, `repo_config`, `profile_management` all green.
- `cargo clippy --no-default-features --all-targets` clean; `cargo fmt --check` clean.
- Smoke probe: dev-release binary builds, `aoe --version` reports `aoe 1.7.0`, and `AOE_CLAUDE_ACCOUNTS_PICKER=1 aoe session --help` works without crashing.

## AI Usage

**AI Model/Tool used:** Claude (claude-opus-4-7) via Claude Code.

**Any Additional AI Details you'd like to share:** AI drafted the discovery module + picker wiring; I (the submitter) reviewed the diff, validated the test plan, and ran the smoke probe before opening this PR.


njbrake:

Thanks for putting this together, @BTForIT, and for the thorough write-up. I'm going to decline this one and want to be transparent about why so future PRs in this space land more easily.

The core concern is architectural fit. AoE is intentionally agent-agnostic: claude, codex, opencode, settl, gemini, and others share the same picker, dashboard, and session model. This PR introduces a Claude-specific subsystem (a discovery module, a dedicated `Instance.claude_config_dir` field, a Claude-only picker row variant, a Claude-only dashboard suffix, and a `default_claude_account` setting) for one agent. Once that pattern lands, the natural follow-up is `codex_accounts`, `gemini_accounts`, and so on, or an eventual refactor to unify them. I'd rather not take on that trajectory.

The other piece is that the multi-config-dir use case is already supported by the per-profile `environment` mechanism you added in #1117, and `docs/guides/configuration.md` explicitly calls out "pinning a Claude/Codex/Gemini config dir per profile" as the supported pattern. One profile per Claude account, with `environment = ["CLAUDE_CONFIG_DIR=..."]`, gets you the same isolation without new surface area. The credential-sharing gap you noted in the commit message ("all 5 accounts currently share one identity") applies equally to both approaches, so this PR isn't unlocking isolation that profiles can't.

A few smaller things worth flagging for future work in this area:

- The hardcoded org-name acronyms (`ForIT`, `WMA`, `CS`) shouldn't ship in upstream code regardless of the broader design.
- The replace-not-additive picker behavior (once any directory exists under `~/.claude-accounts/`, the bare `claude` row disappears) is surprising; a stray test directory would silently change the picker.
- Inventing a filesystem convention (`~/.claude-accounts/`) that isn't documented upstream from Anthropic adds onboarding friction.

On a broader process note: I've really appreciated the work you've shipped, and several of your in-flight stacks (the attention/archive/snooze work, the restart dialog, the transcript recovery piece) look interesting and may well end up making sense for the project. That said, when a large multi-PR stack arrives all at once it's hard for me to engage with the design decisions without holding up implementation you've already done. For features that introduce new vocabulary, defaults, or workflows, I'd really benefit from a short issue or design note up front so we can talk shape before you invest the time. That's not a no on any of those specific features; it just makes the review path much shorter for both of us and reduces the chance of rework.

Closing this one. Happy to chat about the underlying multi-account problem in an issue if profiles + `environment` aren't covering what you need.

PR #1171, opened 2026-05-16
fix(web): pin sidebar session order to created_at desc, no status reshuffle

### LINKED ISSUE #1169: Web sidebar: sessions reorder when their status flips between active/idle — pin to a stable order

## Symptom

In the WorkspaceSidebar, sessions visibly shift up/down as their status transitions between active and idle. The user wants the order to **stay static** — sessions should not jump around just because the agent started or finished a turn.

## Cause

Three comparators all key on session status as the primary sort, so any status change re-shuffles the list:

- `web/src/hooks/useRepoGroups.ts:65-72` (`sortWorkspaces` — per-group):
  ```
  if (a.status === "active" && b.status !== "active") return -1;
  if (a.status !== "active" && b.status === "active") return 1;
  // then localeCompare on workspaceActivityKey, then id
  ```
- `web/src/hooks/useRepoGroups.ts:112-122` (sorts the repo groups themselves with the same status-first rule).
- `web/src/hooks/useWorkspaces.ts:70-74` (initial sort with the same status-first rule, no further tiebreak — preserves creation order only when statuses already match).

Even after dropping the status pin, the secondary `workspaceActivityKey` (last-accessed / idle-entered / created-at) would *still* cause shifts whenever activity bumps a timestamp. So a real "stable order" fix needs to drop both the status branch AND the activity-time branch.

The server (`src/server/api/sessions.rs:263-304`) returns sessions in iteration order with no sort applied; ordering is entirely client-side, which is fine for this fix.

## Suggested fix

In all three comparator sites, replace the body with a stable ordering keyed on something that doesn't change after a session is created:

- Primary: `created_at` ascending (oldest first — new sessions land at the bottom and existing ones never move) OR `created_at` descending (newest first — new sessions land at the top, existing ones never move). Pick one and keep it consistent across the three sites. I'd lean **descending** so new sessions are immediately visible at the top, matching common chat/list UX, but the user's phrasing ("static ordered, no shifting") is the binding requirement and either direction satisfies it.
- Tiebreak: `id` lexicographically.

Drop the `workspaceActivityKey` secondary sort entirely. Drop the status branch entirely. The MULTI_REPO_GROUP pin at `useRepoGroups.ts:113-114` is fine to keep (it's a structural anchor, not status).

## Priority

P2. Pure UX paper cut — nothing breaks, just visual instability.

njbrake:
## Description

Fixes #1169.

The `WorkspaceSidebar` reshuffled whenever a session's status flipped between `active` and `idle`, because the comparator keyed on `status` first (active before idle) and on a mutable activity timestamp second. Both keys move during a session's lifetime, so any agent turn could rearrange the sidebar.

This PR fixes the reshuffle and adds a press-and-hold drag-to-reorder gesture whose result persists server-side so the layout follows the user across browsers and devices.

### How it works

**Server**

- New `PUT /api/workspace-ordering` accepts `{order: [workspace_id]}` and overwrites `workspace-ordering.json` in the app data dir. The handler is gated on `state.read_only` and validates the inbound list length (4096 entries max, 1024 bytes per entry).
- `GET /api/sessions` returns `{sessions, workspace_ordering}`. Before responding, the server merges any workspace ids it has not seen into the persisted list, newest first, so the file always covers every observed workspace. Read-only requests see the merged list but skip the disk write.
- Storage is global, not per-profile, because `list_sessions` already returns sessions across all profiles and the sidebar is one merged view.

**Client**

- `useRepoGroups` sorts workspaces and their groups by their index in the server-supplied ordering. No client-side fallback comparator, no birth-key concept. The server's list is the single source of truth.
- `useSessions` tracks the timestamp of the last local drag. Polled responses arriving within a 4s window do not overwrite the optimistic ordering, so a poll firing between the drag and the PUT landing cannot revert the row.
- `WorkspaceSidebar` wraps each repo group in a `@dnd-kit/sortable` `SortableContext`. A short tap or click navigates as before; a 150ms press-and-hold (with 8px tolerance) lifts the row (`ring-2 ring-brand-500 shadow-lg`) and lets the user drag it. Drag is scoped within a repo group; cross-group drag is intentionally disabled because workspaces are bound to their repo. In read-only mode the sortable wiring is disabled and `onDragEnd` is a no-op so a viewer cannot drag or fire the PUT.
- A `DragSuppressContext` plus a document-level capture click handler swallow the synthetic click Chromium fires after a drag-release, so the trailing click does not navigate to the dragged session.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

(No recording attached. Drag-to-reorder is exercised end-to-end in `web/tests/sidebar-drag-reorder.spec.ts`. Happy to record locally if a human pass is needed.)

## How tested

- `cargo build --features serve`, `cargo clippy --features serve --all-targets -- -D warnings`: clean.
- `cargo test --lib --features serve`: 2084 passed (includes 4 new Rust tests for `workspace_id_for_session` and 4 for `merge_workspace_ordering`).
- `cd web && node_modules/.bin/vitest run`: 302 passed.
- `cd web && node_modules/.bin/playwright test`: 146 passed, 2 skipped (pre-existing).
- Drag spec specifically: 3/3 pass (server ordering honored on first paint, server-returned order rendered verbatim, press-and-hold drag PUTs the new order).

## AI Usage

- [ ] No AI was used

Investigation in #1169 was filed by Nathan with prior Claude help. This PR was implemented end-to-end by Claude Opus 4.7. Multiple rounds of human-in-the-loop iteration (press-and-hold gesture, visual lift, aggressive cleanup, post-review server-side prepend + read-only guards) followed Jules's static review (#1171).


Seluj78 (changes requested):

Heads up: this is a static review from `/aoe-review` (build, tests, lint inside an isolated worktree). Drag-to-reorder has not been manually exercised against a running `aoe serve` yet. @njbrake ping me when it's ready for a hands-on test pass and I'll cover the read-only bypass and the poll-vs-drag race in a live session.

## Blocker

### `src/server/api/sessions.rs:421-437`: `update_workspace_ordering` bypasses read-only mode

```rust
pub async fn update_workspace_ordering(
    Json(body): Json<UpdateWorkspaceOrderingBody>,
) -> impl IntoResponse {
    let ordering = crate::session::WorkspaceOrdering { order: body.order };
    if let Err(e) = crate::session::save_workspace_ordering(&ordering) {
```

Handler does not take `State<Arc<AppState>>` and never checks `state.read_only`. Sibling mutating handlers in the same file (lines 610, 672, 862, 1261, 2516) all guard with `if state.read_only { return read_only ... }`. In read-only mode an authed client can `PUT /api/workspace-ordering` and overwrite `workspace-ordering.json` on disk. Affects every other client polling `/api/sessions`, since the envelope returns the mutated ordering. Trust boundary: `PUT /api/workspace-ordering` HTTP handler.

Fix: take `State<Arc<AppState>>`, gate behind the same `state.read_only` early-return pattern.

### `web/src/components/WorkspaceSidebar.tsx:275`: em dash in code comment

```ts
// Tiny inline grip glyph — lucide's GripVertical is 24x24 by default
```

AGENTS.md: `No emdashes or '--' as separators in docs/comments; use commas, semicolons, or rephrase.`

### PR body: em dashes in `## Description`

Lines 9 and 10 of the description both use em dashes after the commit-title quote (`fix(web): sidebar order keyed on workspace birth` followed by ` — switch the comparator ...`, and the same shape on the next bullet). AGENTS.md em-dash ban applies to PR/issue bodies.

## Required

### `web/src/hooks/useSessions.ts:23-25`: poll clobbers in-flight optimistic drag

```ts
setSessions(data.sessions);
setWorkspaceOrdering(data.workspace_ordering);
```

`applyResult` runs every 3s (`POLL_INTERVAL = 3000`). After a drag, `App.handleReorderWorkspaces` sets local state and fires `PUT`. If the next `/api/sessions` poll lands before `save_workspace_ordering` finishes `fs::write`, the poll reply overwrites `workspaceOrdering` with the pre-drag list and the row snaps back. PR body claims "the next /api/sessions poll will reconcile" but it does the opposite: poll reverts user intent.

Concrete race: drag at t=0, `PUT` in flight, poll at t=3 fires `GET /api/sessions`, server reads `workspace-ordering.json` (which may still hold the pre-drag contents), poll reply applied, optimistic state wiped.

Fix: track a `pendingPutAt` ref and suppress poll-driven `setWorkspaceOrdering` while a `PUT` is in flight, or accept server ordering only when it differs from the last value we PUT.

### `web/src/components/WorkspaceSidebar.tsx:957-1003`: drag affordance + handler always live, ignores `readOnly`

`readOnly` is threaded to `SortableSessionRow` but the grip handle (lines 251-267) renders unconditionally and the `DndContext` always wires `onDragEnd`. In read-only mode the user can drag, local state mutates, and the PUT fires. Stacks on top of the read-only Blocker above.

Fix: when `readOnly`, skip rendering the handle and either drop the `SortableContext` wrap or no-op `handleDragEnd`.

### PR body Checklist: "For UI changes: included screenshot or recording" checked, body admits no screenshot

```

(No screenshot. The behavioural change is "row at index N stays at index N when its status flips, ...")
```

Drag-to-reorder is a UI affordance. A short recording (drag, reload to show persistence, second browser to show cross-device sync) covers what the test suite cannot. Uncheck the box OR attach a recording.

### PR Type: "New Feature" unchecked, only "Bug Fix" + "Refactor" set

Commit 2 of 3 is `feat(web): drag-to-reorder workspaces with cross-device sync`, a user-facing new feature (new `@dnd-kit/*` deps, new `PUT /api/workspace-ordering` endpoint, new `workspaceSort.ts` and spec file). Mark "New Feature".

## Followup (drafted issue)

TUI parity for workspace ordering. Drag-to-reorder is web-only; `src/tui/home/render.rs:1247` ignores `workspace-ordering.json`. A user who reorders rows on web sees the change on every browser/device but not in the TUI on the same machine.

Drafted issue:

> Title: `Honor workspace-ordering.json in the TUI session list`
>
> Body: PR #1171 added per-user workspace ordering, persisted server-side at `~/.agent-of-empires/workspace-ordering.json` (see `src/session/storage.rs:104`, `src/server/api/sessions.rs:421`). The TUI's sidebar in `src/tui/home/render.rs` ignores this file. Scope: read `load_workspace_ordering()` in the TUI render path; port the rank-then-fallback comparator from `web/src/lib/workspaceSort.ts:32` to Rust. TUI does not need a drag affordance; pinning consistency across surfaces is enough. Keyboard reorder shortcuts (writing back to the file) can be a follow-up.

## Nit

- **`src/session/storage.rs:116-135`**: `load_workspace_ordering` and `save_workspace_ordering` are unconditionally compiled but only consumed by `src/server/api/sessions.rs` under the `serve` feature. Gate with `#[cfg(feature = "serve")]`. Low priority; clippy does not warn for `pub` lib symbols.
- **`src/server/api/sessions.rs:421`**: `UpdateWorkspaceOrderingBody.order: Vec<String>` has no length or per-entry size cap. Router-level `DefaultBodyLimit::max(1MB)` is the only bound. Validate `order.len() <= 1000` and per-entry length before persisting.
- **`src/session/storage.rs:128-135`**: `save_workspace_ordering` writes via `fs::write` with no temp-then-rename. Concurrent PUTs from two browser tabs can interleave; a crash mid-write leaves the file truncated. `load_workspace_ordering` handles empty/missing cleanly so corruption surfaces as "ordering lost", not panic. Write to `workspace-ordering.json.tmp` then `fs::rename`.

## Local CI gate results (inside an isolated `/tmp/aoe-review-1171` worktree)

- `cargo fmt --all --check`: clean.
- `cargo clippy --all-targets -- -D warnings`: clean.
- `cargo test --lib`: 1625 passed, 2 failed in `containers::runtime::tests::*`. Unrelated, environmental (need docker daemon + `hello-world` image present).
- `cd web && npx tsc -b`: clean.
- `cd web && npx vitest run`: 308 passed.
- `cd web && npx playwright test sidebar-drag-reorder.spec.ts`: 3/3 passed (after `npm run build`).
- `cd web && npx playwright test` (full): 7 failures in `terminal-focus-shortcut.spec.ts`. Unrelated to this PR, not touched.
- ESLint on PR-touched files: 3 pre-existing errors in `App.tsx`, net +0 vs base. Lint not CI-gated.

PR #1173, opened 2026-05-16
fix(session): resume-fallback cascade for restart/start paths

jerome-benoit:
## Description

Fixes the failure mode where `aoe session restart` leaves the user with a dead pane when the agent crashes at startup because of a stale `--resume <sid>` (transcript missing, sid invalid, autocompact thrash).

When restart-time start passes a resume flag and the pane dies within a settle probe, the cascade:

1. Stops the Tier-1 session-id poller so it cannot re-write the stale sid into tmux env after we clear it.
2. Tears down the dead tmux session via `kill_clean` (respawn-dead-pane + kill + 100ms sleep, the macOS dead-pane workaround).
3. Clears the stale sid in memory, on disk, and adds it to a per-instance retroactive-capture exclusion set so opencode/codex/gemini/pi/vibe/hermes cannot re-import it from on-disk artefacts on Tier 2.
4. Retries `start_with_size_opts` with `skip_on_launch=true` so `on_launch` hooks do not run twice.
5. If the retry also fails, propagates `Err`; the existing `Status::Error` + `last_error` path takes over unchanged.

The probe is settle-based: poll up to 3s for `pane_dead`, declare `Alive` only after a 2000ms post-shell grace window. The grace covers the gap between the boot shell `exec`ing into the agent binary and the agent crashing on the bad sid; opencode (Bun-compiled native binary) reaches `pane_dead` between ~900ms and ~1100ms after spawn on a warm cache, so the grace is sized at ~2x that p100 to stay comfortably inside `RESUME_PROBE_MAX = 3000ms`. Faster agents (Node-based: claude, codex, gemini, pi, qwen) crash well before the grace window closes; the grace cost is paid only once per Tier-1 cold restart that actually exercises `--resume`. `pane_dead` is the unambiguous trigger.

`build_exclusion_set` keeps its single-arg signature; the cascade path uses `Instance::retroactive_capture_exclusion_set`, which composes the live tmux scan with the per-instance excludes so poll-closure callers do not pay for an empty placeholder set.

The cascade is wired into:

| Surface | Behaviour on `Restarted { stale_sid }` |
| --- | --- |
| `aoe session restart <id>` | `✓ Restarted session: foo (resume failed for sid 1234; started fresh, prior history not loaded)` |
| `aoe session restart --all` | summary line `✓ Restarted N/M sessions (K without prior history):`, with a stale-history suffix appended to each affected entry |
| TUI Enter / attach | transient toast `Resume failed for sid 1234; started fresh (history not loaded)` |
| `POST /api/sessions/:id/ensure` | response includes `resume_outcome: "resumed" \| "restarted" \| "fresh"` and `stale_session_id` |

`aoe session start <id>` is intentionally NOT routed through the cascade: keeping it as a near-no-op on existing sessions preserves its idempotent contract. Only the explicit restart paths recover.

**`Status::Error` reset (behavior note).** `start_with_resume_fallback` clears `Status::Error`, `last_error`, and `last_error_check` on entry so a successful relaunch from any restart surface (TUI Enter/restart, `aoe session restart [id|--all]`, REST `ensure_session`, cockpit-mode short-circuit) does not leave a stale error chip up. The 30 s `last_error_check` debounce is reset on every user-initiated restart, including TUI/CLI surfaces. REST `ensure_session` re-asserts `status=Starting`, `last_error=None` pre-call as defense in depth.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording. N/A; backend-only. The TUI toast piggybacks on the existing `UpdateStatus::transient` channel and renders identically to the existing `restart failed: ...` toast.

Quality gates run locally on this branch:

```
cargo fmt --check
cargo clippy --release --lib -- -D warnings
cargo clippy --release --lib --features serve -- -D warnings
cargo test --release --lib                    # 1638 pass, 0 fail (10 in session::instance::tests::resume_fallback)
cargo test --release --lib --features serve   # adds cockpit short-circuit guard
cargo build --release
cargo build --features serve
```

All green. Integration tests under `mod resume_fallback` are gated on `tmux -V` per the existing convention; one test pins the lower bound of `RESUME_PROBE_POST_SHELL_GRACE` against a 1.2s post-shell-exit pane death so a future revert toward 500ms fails the suite.

## Breaking changes (intra-crate API)

- `Instance::restart()` (no-arg wrapper) is removed. Call `restart_with_size(None)` instead.
- `Instance::restart_with_size` and `Instance::restart_with_size_opts` now return `Result<StartOutcome>` instead of `Result<()>`. Callers that previously discarded `()` now need to `.map(drop)?` or destructure the outcome (e.g. to surface `StartOutcome::Restarted { stale_sid }` to the user).
- `Instance::kill_clean` is now `pub(crate)` so the REST `ensure_session` path can use it for parity with TUI/CLI restart paths (avoids the macOS dead-pane race on bare tmux kill).

No external consumers; all in-tree callers were updated in this PR.

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 (anthropic/claude-opus-4-7) via the OhMyOpenCode `Sisyphus` agent.

Design and implementation by the AI agent under my direction.

### DIFF
--- a/docs/guides/session-resume.md
+++ b/docs/guides/session-resume.md
@@ -16,10 +16,6 @@ For sandboxed (Docker) sessions, the poller runs the same scan inside the contai
 - Runtime rotation via `/clear`, `--fork-session`, or fresh `claude` invocation in the same pane.
 - Manual override via the CLI when you want to point a session at a specific conversation.
 
-## What's not covered (yet)
-
-- Other agents (OpenCode, Codex, Gemini, Vibe, Pi). They're tracked behind separate follow-up PRs; their sessions still launch fresh until then.
-
 ## Manual override
 
 To point a session at a different Claude conversation ID without launching it:
--- a/src/cli/send.rs
+++ b/src/cli/send.rs
@@ -51,8 +51,16 @@ pub async fn run(profile: &str, args: SendArgs) -> Result<()> {
                         stale_history_suffix(&sid),
                     );
                 }
-                Ok(EnsureReadyOutcome::Started) => {
+                Ok(EnsureReadyOutcome::Started { stale_sid: None }) => {
                     eprintln!("  (started stopped session before send)");
+                }
+                Ok(EnsureReadyOutcome::Started {
+                    stale_sid: Some(sid),
+                }) => {
+                    eprintln!(
+                        "  (started stopped session before send){}",
+                        stale_history_suffix(&sid),
+                    );
                 }
                 Ok(EnsureReadyOutcome::AlreadyAlive) => {}
                 Err(EnsureReadyError::Transient(status)) => {
--- a/src/cli/send.rs
+++ b/src/cli/send.rs
@@ -3,6 +3,7 @@
 use anyhow::{bail, Result};
 use clap::Args;
 
+use crate::cli::session::stale_history_suffix;
 use crate::session::{EnsureReadyError, EnsureReadyOutcome, Storage};
 
 #[derive(Args)]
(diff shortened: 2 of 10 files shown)


jerome-benoit (commented):
(retracted by author)

jerome-benoit (commented):
(retracted by author)

jerome-benoit (commented):
(retracted by author)

jerome-benoit (commented):
(retracted by author)

njbrake (approved):

PR #1183, opened 2026-05-17
feat: AI engine field on rename dialog

BTForIT:
## Description

Adds an AI engine selector to the rename dialog, matching the selector already present in the new-session dialog.

- In Session mode the rename dialog now has four fields: title, group, profile, and AI engine. `RenameData` carries `tool: Option<String>`; `None` means "leave the tool unchanged".
- Cycling the profile snaps the AI engine to that profile's `default_tool` when that tool is in `available_tools`, so picking a profile pre-populates the engine the same way the new-session dialog does. Manually overriding the engine afterwards does not snap the profile back.
- Group mode is unchanged (two fields); a group has no tool of its own, so `RenameData.tool` is always `None` there.
- `rename_selected` gains a `new_tool` parameter and applies the engine change before the profile move, so the session's next launch boots the selected agent binary.

Arrow keys and Space cycle the engine when the field is focused; with no tools available the field is a harmless no-op.

No data-model or config changes. `RenameDialog::new` gains `current_tool` and `available_tools` parameters; `new_for_group` is unchanged. Only pre-existing APIs are used (`resolve_config_or_warn`, `AvailableTools::available_list`, `mutate_instance`).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test --lib`: 1634 passed, 0 failed, including 10 new rename-dialog tests; `cargo build`, `cargo fmt --check`, `cargo clippy` all clean)
- [ ] Documentation was updated where necessary (no `docs/` page enumerates the individual rename-dialog fields)
- [ ] For UI changes: included screenshot or recording (this adds one selector row to an existing TUI dialog; happy to attach an asciinema recording, or a maintainer can apply the `needs-recording` label)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Code (Claude Opus 4.7)

The feature is the fork owner's design; Claude Code re-authored the implementation as a clean commit on top of current `main` for this upstream PR. A human will handle review discussion.

### DIFF
--- a/src/tui/dialogs/rename.rs
+++ b/src/tui/dialogs/rename.rs
@@ -7,6 +7,7 @@ use tui_input::backend::crossterm::EventHandler;
 use tui_input::Input;
 
 use super::DialogResult;
+use crate::session::profile_config::resolve_config_or_warn;
 use crate::tui::components::{
     render_text_field, render_text_field_with_ghost, GroupGhostCompletion, ListPicker,
     ListPickerResult,
@@ -22,6 +23,9 @@ pub struct RenameData {
     pub group: Option<String>,
     /// New profile (None means keep current, Some(name) means move to that profile)
     pub profile: Option<String>,
+    /// New AI tool (None means keep current). Always None in Group mode since
+    /// a group has no tool of its own.
+    pub tool: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


BTForIT:
Closing this in favor of the restart dialog (#1184).

The rationale: a rename never re-execs the agent. `rename_selected()` applies a
tool change as a plain field write (`inst.tool = target_tool`) followed by
`save()` + `reload()`, and `reload()` only re-reads `sessions.json` into memory.
So an AI-engine field on the rename dialog changes a stored label that the
running process cannot see; the new engine would only take effect on some
later, unrelated restart. A setting that visibly changes but does nothing until
an unrelated action is worse than no setting.

The restart dialog (#1184) already carries an AI-engine picker, and restart is
exactly the moment an engine change takes effect, so that is the correct home
for it. Rename stays a pure rename.

PR #1187, opened 2026-05-17
fix(tui): keep command palette selection visible past viewport

### LINKED ISSUE #1186: TUI: command palette (Ctrl+K) selection highlight gets clipped
> **Disclosure**: This issue was drafted by Claude. The bug was reproduced
> by hand and the writeup was reviewed by me (@bell-hyun) before posting.

## Symptom

In the TUI, opening the command palette with `Ctrl+K` and pressing `↓` repeatedly causes the selected entry's highlight (the `▶` marker and bold styling) to disappear below the visible list before reaching the bottom. The selection cursor still tracks the right item — pressing Enter runs the expected command — but visually nothing looks selected.

Reproduces only with an **empty search query** (when group headers are visible). Typing any character hides the headers and the symptom goes away.

## Likely cause

`src/tui/dialogs/command_palette.rs:401-471` computes `scroll_offset` in *match-index units*, but the rendered `lines` vec contains entries plus group headers ("Actions", "Views", "Settings"). `.skip(scroll_offset).take(visible)` caps the loop at `visible` entries, but each iteration may push an extra header line, so `lines.len()` exceeds `list_area.height` and `Paragraph` silently clips the bottom — taking the selected row with it.

## Repro
1. `cargo build && ./target/debug/aoe`
2. Resize terminal short enough that the palette list can't fit all built-in commands (~25 row terminal works).
3. `Ctrl+K`, leave the search empty.
4. Press `↓` repeatedly. Highlight disappears somewhere in the Views or Settings group.

Image

bell-hyun:
## Description

> **Disclosure**: This PR was drafted by Claude. The bug was reproduced
> and the fix was verified by me (@bell-hyun) before posting.

Closes #1186.

Command palette (Ctrl+K) selection highlight scrolled off the bottom of the visible list when group headers were present. The previous `scroll_offset` counted match indices, but the rendered `lines` vec also contains group header rows ("Actions", "Views", "Settings"), so `lines.len()` exceeded `list_area.height` and `Paragraph` silently clipped the bottom.

**Fix**: window the rendered lines in *line-index units*. Iterate all matches, record `selected_line = lines.len()` immediately before pushing the selected entry, then render `lines[start..end]` with `start = selected_line.saturating_sub(visible - 1)`. +8 / −13 in one function, no API changes.

**Known limitation (pre-existing)**: scroll position is stateless, so the highlight pins to the viewport's bottom edge once scrolled. Same behavior as the original code; band-style scrolling would need a persistent scroll field and is a separate UX improvement.

**Before:**
before

**After:**
after

---

_Verified locally_: `cargo test` (1605 passed), `cargo test --test e2e` (55 passed, 2 ignored), `cargo clippy --all-targets -- -D warnings` (clean), `cargo fmt -- --check` (clean).

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (no doc changes for an internal rendering fix)

## AI Usage

**AI Model/Tool used:** Claude (Anthropic) via Claude Code

**Any Additional AI Details you'd like to share:** Symptom reproduced by hand on macOS before drafting; fix tested locally with `cargo test` + manual TUI verification. I will respond to review comments myself.

### DIFF
--- a/src/tui/dialogs/command_palette.rs
+++ b/src/tui/dialogs/command_palette.rs
@@ -400,13 +400,9 @@ impl CommandPaletteDialog {
         // List
         let list_area = chunks[2];
         let visible = list_area.height as usize;
-        let scroll_offset = if self.selected >= visible {
-            self.selected - visible + 1
-        } else {
-            0
-        };
 
         let mut lines: Vec<Line> = Vec::new();
+        let mut selected_line: usize = 0;
         if self.matches.is_empty() {
             lines.push(Line::from(Span::styled(
                 "  No matches",
@@ -414,13 +410,7 @@ impl CommandPaletteDialog {
             )));
         } else {
             let mut last_group: Option<PaletteGroup> = None;
-            for (display_idx, &entry_idx) in self
-                .matches
-                .iter()
-                .enumerate()
-                .skip(scroll_offset)
-                .take(visible)
-            {
+            for (display_idx, &entry_idx) in self.matches.iter().enumerate() {
                 let cmd = &self.entries[entry_idx];
 
                 // Show group header on transition (only when no query, since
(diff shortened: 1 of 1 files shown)


Seluj78 (approved):

njbrake (approved):
🙏

PR #1206, opened 2026-05-18
feat(tui): add lock_sort_order setting to pin sort order

### LINKED ISSUE #1205: Add setting to lock sort order and disable O / Ctrl+O cycling
**Note:** Claude handled the prose. Idea and decisions are mine. — @raphaeldavidf

## Problem

The `O` / `Ctrl+O` hotkeys cycle the session list sort order through five modes (Newest, Recent, Oldest, A-Z, Z-A). Once you've settled on a preferred order, it's easy to misclick `O` and unintentionally change the ordering. There's no way to lock it in place.

## Proposed solution

An `Option<SortOrder>` session setting (`lock_sort_order`) that, when set to a specific sort order, forces that order and makes `O` / `Ctrl+O` no-ops. Configurable via Settings > Session or directly in `config.toml`:

```toml
[session]
lock_sort_order = "newest"
```

The help overlay should indicate the locked state (e.g. `(sort locked: Newest)`) and dim the disabled hotkeys. The command palette should hide the "Cycle sort order" entry when locked. Per-profile overrides should work the same way they do for `strict_hotkeys`.

raphaeldavidf:
## Description

Closes #1205.

Adds a `lock_sort_order` session setting (`Option<SortOrder>`) that locks the session list to a specific sort order and disables the `O` / `Ctrl+O` cycling hotkeys. Follows the same pattern as `strict_hotkeys`: config field on `SessionConfig`, profile override on `SessionConfigOverride` with merge logic, settings TUI field with global/profile apply and clear, and runtime state on `HomeView` loaded at startup and refreshed on config change.

When set, the sort order is forced to the configured value on startup and on config refresh. The help overlay shows `(sort locked: Newest)` and dims the disabled hotkey entries. The command palette hides the "Cycle sort order" entry. The settings TUI exposes a select dropdown (Off / Newest / Recent / Oldest / A-Z / Z-A).

Configurable via Settings > Session or `lock_sort_order = "newest"` under `[session]` in config.toml. Supports per-profile overrides.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

How tested:
- `cargo build --release` from the worktree, launched the binary, toggled Lock Sort Order in Settings > Session, confirmed `O` / `Ctrl+O` are no-ops and help overlay shows `(sort locked: ...)` with dimmed hotkeys.
- `cargo clippy`: zero warnings.

## AI Usage

- [ ] No AI was used

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -458,6 +458,10 @@ pub struct SessionConfig {
     /// Off by default; existing users keep the legacy single-letter UX.
     #[serde(default)]
     pub strict_hotkeys: bool,
+
+    /// Lock the sort order to a specific value, disabling the O / Ctrl+O cycling hotkeys.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub lock_sort_order: Option<SortOrder>,
 }
 
 impl SessionConfig {
--- a/src/session/profile_config.rs
+++ b/src/session/profile_config.rs
@@ -9,8 +9,8 @@ use std::collections::HashMap;
 use std::fs;
 
 use super::config::{
-    ColorMode, Config, ContainerRuntimeName, DefaultTerminalMode, TmuxClipboardMode, TmuxMouseMode,
-    TmuxStatusBarMode,
+    ColorMode, Config, ContainerRuntimeName, DefaultTerminalMode, SortOrder, TmuxClipboardMode,
+    TmuxMouseMode, TmuxStatusBarMode,
 };
 use super::get_profile_dir;
 
@@ -232,6 +232,9 @@ pub struct SessionConfigOverride {
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub strict_hotkeys: Option<bool>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub lock_sort_order: Option<SortOrder>,
 }
 
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
@@ -442,6 +445,9 @@ pub fn apply_session_overrides(
     }
     if let Some(strict_hotkeys) = source.strict_hotkeys {
         target.strict_hotkeys = strict_hotkeys;
+    }
+    if source.lock_sort_order.is_some() {
+        target.lock_sort_order = source.lock_sort_order;
     }
 }
 
--- a/src/tui/components/help.rs
+++ b/src/tui/components/help.rs
@@ -162,6 +162,7 @@ impl HelpOverlay {
         theme: &Theme,
         sort_order: SortOrder,
         strict_hotkeys: bool,
+        lock_sort_order: Option<SortOrder>,
     ) {
         let x = area.x + (area.width.saturating_sub(DIALOG_WIDTH)) / 2;
         let y = area.y + (area.height.saturating_sub(DIALOG_HEIGHT)) / 2;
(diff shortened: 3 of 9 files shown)


njbrake:
Hi @raphaeldavidf thank you for the contribution? Could you please explain the motivation for this PR? To me I do not completely understand the value of a lock for this value, since it's already user controlled. What benefit does the use get from disabling the controls that only they have access to already?

raphaeldavidf:
Hi @njbrake, thanks for taking the time to look at my PR! I just started using AoE on Saturday and honestly don't know how I lived without it before, great job on this tool.

The motivation is accidental keypresses. `O` is a single unmodified key right next to `P` (profiles), `N` (new session), and navigation keys I use constantly. When I'm moving fast through the session list, I misclick `O` and the sort order changes without me noticing, which reshuffles the entire list. It's disorienting because I'm used to sessions being in a specific position; I have muscle memory for where things are in my preferred order and I'll never want to re-sort them. And recovering from a misclick is expensive since there are five sort modes to cycle through, so getting back to the one I wanted takes several `O` presses.

It's similar to why `strict_hotkeys` exists: the controls are user-only, but accidental activation is the problem, not unauthorized access. `strict_hotkeys` solves it broadly by requiring Shift/Ctrl for action keys; `lock_sort_order` solves it narrowly for one setting I want to set once and forget.

The `Option<SortOrder>` design (rather than a boolean) also means it doubles as a declarative config: "this profile always sorts by Newest" without relying on the app state file, which is useful for shared configs or fresh installs.

Happy to hear if you think the use case is too niche or if there's a simpler approach you'd prefer.

njbrake:
@raphaeldavidf ooo ok excellent thank you. As soon as you said that i was like "oh wait I also accidentally swap sorts when trying to hit P too".

TBH I'm not a huge fan of the 'lock sort' semantics but how about this: lets turn the 'o' button to open up a new modal where you select what sort order you'd like, just like how when you hit 'P' it brings up a picker where you pick the profile to sort by?

For maybe a 'bonus' fix to this redesign is that you could have some combined modal that lets you adjust the sort 'o' and the group 'g' settings, because the sort and group setting interact with each other.

What do you think, and are you open to making those changes in this PR or would you like me to tackle it?

raphaeldavidf:
Totally agree, a picker modal is a much better solution. Fixes the root cause instead of patching around it.

Funny you mention the group picker too, I actually opened an issue for exactly that a few days ago (https://github.com/njbrake/agent-of-empires/issues/1185) and had a PR ready but forgot to push it. So I'm very much aligned on the direction for both sort and group.

Happy to take this on. Would you prefer both sort and group pickers in the same PR, or split into two?

njbrake:
@raphaeldavidf I am fine with them being in the same PR. 

In general I love the functionality of group and sort, but I do not love the current UI or UX. So feel free to think about and present a fix that lets a user more smoothly configure how they want to sort and/or group their sessions. 

There's a whole group of PRs sitting and waiting that are based around the idea of 'favorite' sessions and dynamic sorting (e.g. https://github.com/njbrake/agent-of-empires/pull/1084). I haven't yet formed a concrete opinion about whether I want those other changes to come into aoe or not. So, if when you're fixing the grouping and sorting design you happen to have opinions or thoughts about adding the 'attention based' sorting is a good or meh idea, those thoughts are also welcome :) 

Also feel free to join our discord if you'd like to have more general chats! https://discord.gg/5N3QKX3f6s

njbrake:
I'm going to close this PR out, and i'll let you reopen a new PR with the design we discussed above. Thanks!

PR #1242, opened 2026-05-18
feat(session): per-account Claude sessions via CLAUDE_CONFIG_DIR

BTForIT:
## Description

Adds per-account Claude sessions. Claude Code supports running under an
isolated config root via `CLAUDE_CONFIG_DIR`; this surfaces that in Agent of
Empires so a session can be launched bound to a specific Claude account
(work / personal / client), with settings, MCP servers, and history isolated
per account.

**What it does**

- New `src/session/claude_accounts.rs`. `discover_accounts()` scans
  `~/.claude-accounts/`, skipping `_shared`, dotted dirs, and any directory
  without a `settings.json` (file or symlink); returns `ClaudeAccount`
  entries sorted by name for stable picker ordering. `display_label()`
  pretty-prints `forit-main` as `ForIT Main` via a small acronym table with
  title-case fallback.
- `Instance` gains `claude_config_dir: Option<PathBuf>`. When set,
  `build_host_command` prepends `CLAUDE_CONFIG_DIR=<expanded>` to the host
  environment prefix ahead of profile-level entries, so a per-session
  override beats a profile setting. `~` and `$HOME` are expanded at spawn
  time so `sessions.json` stays portable. Sandboxed sessions skip the
  injection (the host path may not exist inside the container).
- The new-session Tool picker shows one `claude => <Account>` row per
  discovered account. When at least one account exists the bare `claude` row
  is replaced by account rows; with zero accounts the bare row remains as the
  no-multiplex fallback. The selection propagates through both the sync and
  async (`CreationPoller`) creation paths.
- The dashboard surfaces a dimmed `@ForIT Main` label on each session row
  whose `claude_config_dir` is set, kept in sync with the picker label.
- New `session.default_claude_account` setting (global + profile override)
  pins a profile to a specific account by directory basename; the picker
  then opens on that row. An unknown pin falls back to the first claude row
  with a one-shot warning. Exposed in the settings TUI as a Select populated
  from the live account list.

Pre-existing sessions still attach through their stored `claude_config_dir`;
no migration needed.

**Known limitation:** `CLAUDE_CONFIG_DIR` multiplexes settings, MCP, and
history, but not the Anthropic API token (a single shared keychain entry).
Picker rows reflect user intent correctly; full credential isolation belongs
upstream in claude-code.

## PR Type

- [x] New Feature

## Checklist

## AI Usage

**AI Model/Tool used:** Claude Code (Claude Opus 4.7)

Implementation drafted with Claude Code under the maintainer's direction and
review. Tests are part of the change.


njbrake:

Thanks again, @BTForIT. I'm going to close this one as well, because as best I can tell it's the same feature as #1159 with the same shape, and I don't see the underlying objections addressed.

Specifically, comparing this PR against the decline comment I left on #1159:

- The architectural concern was that a Claude-specific subsystem (discovery module, dedicated `Instance` field, Claude-only picker rows, Claude-only dashboard suffix, `default_claude_account` setting) sets a precedent we don't want for an intentionally agent-agnostic tool. #1242 ships the same surface area under the same names.
- I pointed at the per-profile `environment` mechanism from #1117 as the supported path for pinning `CLAUDE_CONFIG_DIR` per account, and noted the credential-sharing gap applies equally to both approaches so this PR isn't unlocking isolation that profiles can't. The PR description here doesn't engage with that comparison.
- The smaller things I flagged are also still present: the hardcoded org-name acronyms (`ForIT`, `WMA`, `CS`) in `display_label()`, the replace-not-additive picker behavior where any directory under `~/.claude-accounts/` removes the bare `claude` row, and the undocumented `~/.claude-accounts/` filesystem convention.

I also asked in #1159 for a short issue or design note before another attempt at features that introduce new vocabulary, defaults, or workflows, so we can talk shape before you invest implementation time. I don't see one between #1159 closing and this PR opening, and I'd really like that to be the next step rather than another revision.

I'm closing this one. If the profiles + `environment` path isn't actually covering the use case, please open an issue describing what's missing and we can work from there. I'd rather spend the round trip on the problem than on a re-review of the same patch.

PR #1248, opened 2026-05-19
fix(cockpit): silent-orphan watchdog for adapter wedges

### LINKED ISSUE #1240: Cockpit: silent-orphan watchdog needed, cancel-escalation only arms after Force End so adapter hangs are forever otherwise

## Problem

A cockpit turn can finish streaming the agent's complete response and then leave the session pinned to `turnActive=true` forever because no `Stopped` event ever fires to close the turn. The adapter (`claude-agent-acp` + `claude-agent-sdk`) never sends the JSON-RPC `PromptResponse` for the original `session/prompt` request, and AoE's existing cancel-escalation watchdog only arms after the user explicitly clicks Force End. If the user doesn't notice the wedge, the spinner runs forever.

Two independent reproductions today, ten minutes apart, in two different sessions on the same `aoe serve` instance. Same end-state in both, identical-looking final agent message in both. Not a one-off.

The upstream root cause is tracked at `agentclientprotocol/claude-agent-acp#688`. This issue is the **aoe-side bandaid** that closes the gap while waiting on the upstream fix.

### Session A: `57dadb734e9848f7` (no user intervention, sat wedged forever)

```
seq   4  21:11:20  UserPromptSent
seq  27  21:11:42  Stopped:prompt_complete       ← turn 1 healthy
seq  28  21:12:25  UserPromptSent
seq 184  21:16:16  Stopped:prompt_complete       ← turn 2 healthy
seq 185  21:16:38  UserPromptSent
seq 186  21:16:39  Stopped:prompt_complete       ← turn 3 healthy
seq 187  21:17:27  UserPromptSent
seq 229  21:19:03  Stopped:prompt_complete       ← turn 4 healthy
seq 230  21:19:14  UserPromptSent                ← turn 5 starts
... intra-turn approvals at seq 248/249 and 748/749 ...
seq 750-769  21:50:05-11  AgentMessageChunk * 20 ending "Open draft PR / tweak / hold?"
seq 770-771  21:50:11-17  UsageUpdated * 2
                                ← nothing further; PromptResponse never sent
```

The agent's final chunk is unambiguously an end-of-turn question. The two `UsageUpdated` frames are the standard "wrap up accounting" pair (second one carries `cost` populated). Then dead silence on the wire indefinitely. Worker process alive (PID 72585), socket open, daemon attached, no errors anywhere. The pending `session/prompt` request future hung; no `Stopped` event ever fired; reducer's `lastStoppedSeq` never advanced; `turnActive` stayed true.

User never clicked Force End. Spinner ran forever. Only way out was a hard worker restart.

### Session B: `c8969376e13a4237` (user clicked Force End, escalation worked)

```
seq 192  21:29:46  UserPromptSent
seq 443-448  21:55:13-15  AgentMessageChunk * 6 ending "Open draft PR / tweak / hold?"
seq 449-450  21:55:15-17  UsageUpdated * 2
                                ← no PromptResponse for ~2 minutes
seq 451  21:57:16  Stopped:agent_unresponsive   ← user clicked Force End, watchdog escalated
seq 452  21:57:17  Stopped:user_stopped         ← reap_user_stopped saw the wiped registry
seq 453  21:58:33  AcpSessionAssigned            ← respawn via session/load
seq 489  21:59:02  Stopped:prompt_complete      ← next prompt completed cleanly
```

Same end-state as session A (final chunk + 2 UsageUpdated then silence). User noticed, clicked Force End, the existing cancel-escalation watchdog grace expired, supervisor SIGTERMd the runner and respawned via `session/load`. Reconnect worked; next prompt completed normally. Adapter wasn't crashed, just had a hung Promise for that specific in-flight request.

Both turns ended on identical content ("Open draft PR / tweak / hold?"). Coincidence flagged at upstream `#688`; possibly a race tied to specific terminal-message shape.

## What's there today

`src/cockpit/acp_client.rs:196` (release 1.7.0, built from `cockpit-polishing-9`):

```rust
const CANCEL_ESCALATION_GRACE: std::time::Duration = std::time::Duration::from_secs(10);
```

`src/cockpit/acp_client.rs:2516-2540`:

```rust
let mut agent_unresponsive = false;
let mut cancelling = false;
let cancel_grace = tokio::time::sleep(CANCEL_ESCALATION_GRACE);
tokio::pin!(cancel_grace);

loop {
    tokio::select! {
        res = &mut prompt_fut => {
            let _ = res?;
            break;
        }
        _ = &mut cancel_grace, if cancelling => {
            warn!(
                target: "cockpit.acp",
                session = %session_label,
                grace_secs = CANCEL_ESCALATION_GRACE.as_secs(),
                "agent ignored session/cancel past grace window; escalating to runner restart"
            );
            agent_unresponsive = true;
            shutdown = true;
            break;
        }
        cmd = cmd_rx.recv() => {
            match cmd {
                Some(ClientCmd::Cancel) => {
                    /* send session/cancel, set cancelling = true, arm cancel_grace */
                }
                /* SetMode, Prompt, Shutdown */
            }
        }
    }
}
```

The watchdog logic is correct and tested. When the user clicks Force End:

1. `ClientCmd::Cancel` arrives, sends `session/cancel`, sets `cancelling = true`, arms `cancel_grace`.
2. If the adapter sends a `PromptResponse` within 10s (with `StopReason::Cancelled` or otherwise), `prompt_fut` resolves cleanly.
3. If the 10s grace elapses without a response, `agent_unresponsive` fires, the runner gets SIGTERMd in `src/cockpit/supervisor.rs:849`, the registry is wiped, supervisor respawns with `session/load`.

This is the path session B took. Works.

## Root cause of the silent-forever case (session A)

The cancel-escalation watchdog requires `cancelling = true` to be set, which only happens when a `ClientCmd::Cancel` is delivered, which only happens when the user clicks Force End (or fires `aoe cockpit stop`). If the user doesn't notice or doesn't click, `cancelling` stays false, the watchdog never arms, and `prompt_fut` hangs forever.

There is no silent-orphan watchdog. The adapter can drop the `PromptResponse` and AoE will wait indefinitely without any UI signal beyond the spinner.

## Proposal: silent-orphan watchdog (bandaid)

Add a second watchdog arm to the same `select!` loop that fires on inbound-silence regardless of cancel state. Triggers the same `agent_unresponsive` escalation path that the cancel-escalation watchdog already uses, so all the downstream cleanup (SIGTERM runner, wipe registry, respawn via `session/load`) is reused unchanged.

### Sketch

`src/cockpit/acp_client.rs:2516-2540`:

```rust
// Existing
let mut agent_unresponsive = false;
let mut cancelling = false;
let cancel_grace = tokio::time::sleep(CANCEL_ESCALATION_GRACE);
tokio::pin!(cancel_grace);

// NEW: silent-orphan watchdog, longer threshold, arms after first notification
let mut had_any_notification = false;
let silent_orphan_grace = tokio::time::sleep(silent_orphan_threshold());
tokio::pin!(silent_orphan_grace);

loop {
    tokio::select! {
        res = &mut prompt_fut => {
            let _ = res?;
            break;
        }
        _ = &mut cancel_grace, if cancelling => {
            warn!(target: "cockpit.acp", "agent ignored session/cancel past grace; escalating");
            agent_unresponsive = true;
            shutdown = true;
            break;
        }
        _ = &mut silent_orphan_grace, if had_any_notification => {
            warn!(target: "cockpit.acp",
                  threshold_secs = silent_orphan_threshold().as_secs(),
                  "no inbound notifications past silent-orphan threshold; declaring prompt orphaned");
            // Best-effort cancel first to give adapter a final chance
            let _ = connection.send_notification(CancelNotification::new(acp_session_id.clone()));
            agent_unresponsive = true;
            shutdown = true;
            break;
        }
        cmd = cmd_rx.recv() => { /* existing */ }
    }
}
```

Reset `silent_orphan_grace` inside `on_receive_notification` whenever a notification arrives:

```rust
silent_orphan_grace.as_mut().reset(
    tokio::time::Instant::now() + silent_orphan_threshold(),
);
had_any_notification = true;
```

(Both fields need to be shared with the notification handler the same way `last_event_at` already is.)

### Threshold

Constant + config knob, mirroring the existing `RESUME_IDLE_GRACE_DEFAULT` pattern in `acp_client.rs:178`:

```rust
const SILENT_ORPHAN_THRESHOLD_DEFAULT: Duration = Duration::from_secs(90);
```

Plus `[cockpit] silent_orphan_threshold_secs` in the config file with minimum clamp at 30s. Default of 90s comfortably handles long Bash and Task subagent silences; long enough to not false-positive on real slow tools, short enough to catch silent-orphan turns within ~2 minutes.

The "had_any_notification" gate prevents false-positive on the very first prompt of a session (before any notifications have arrived). Once a single notification lands, the turn is "live" and silence past the threshold means something wedged.

### Why this is the right bandaid

- **Reuses the existing escalation path.** No new code in supervisor, no new event types in the reducer. `Stopped { reason: "agent_unresponsive" }` already drives the right downstream behavior.
- **Doesn't require upstream changes.** Works against today's adapter, the broken 0.32.0 path, and any future version. If upstream fixes #688 the bandaid never fires.
- **One file changed (`acp_client.rs`).** Small, surgical, scoped to the in-flight prompt loop.
- **Configurable threshold.** Users on slow machines or with unusual workloads can tune.

### Reducer / UI side

`Stopped:agent_unresponsive` already flows through to the cockpit reducer at `web/src/lib/cockpitTypes.ts:629-647` and clears `inFlightTool`, advances `lastStoppedSeq`, ends the spinner. No frontend changes needed.

Optional polish: add a small "Turn ended without explicit completion (adapter bug, tracking at agentclientprotocol/claude-agent-acp#688)" footnote on activity rows where the most recent Stopped reason is `agent_unresponsive` and there was no user-initiated cancel. Distinguishes "we escalated for you" from "you clicked Force End." Defer until the silent-orphan watchdog has shipped and we see how often it fires.

## Why now

- Silent-forever wedge is the worst UX failure mode in cockpit. User has to remember to click Force End or the session hangs indefinitely. Most users will not notice within the first few minutes.
- Pattern is reproducible (twice today within 10 minutes on identical-looking final messages).
- Bandaid is small and decoupled from upstream timeline.
- Pairs with #1215 (adapter version pin) — orthogonal; both should land independently.
- Pairs with the watchdog family (#1196, #1216, #1176) — same select-loop area, consistent escalation pattern.

## Open questions

- **Threshold default.** 90s is conservative-but-tight. Long Bash (npm install of a giant repo, Playwright runs without intermediate output) can plausibly exceed 90s of inbound silence even when healthy. Worth measuring on real workloads before committing. Alternative: a sliding threshold that grows with turn duration (a turn at minute 30 might be more tolerant of a 5min gap than one at minute 1).
- **Best-effort cancel before SIGTERM.** The sketch sends `session/cancel` before the SIGTERM escalation. Gives the adapter a final chance to respond. If it does respond, the existing cancel-escalation 10s grace catches it; if it doesn't, SIGTERM proceeds. Slight delay (up to 10s) before the runner dies; probably acceptable since the user already waited 90s of silence.
- **UI affordance during the escalation.** Once the silent-orphan timer fires, the user sees no signal that something is happening server-side until the spinner clears post-respawn. A small banner ("Turn appears wedged; restarting runner...") with a Cancel button (suppresses the auto-respawn, lets the user investigate manually) would be nicer than the current "spinner forever, then suddenly restart." Defer.
- **Telemetry.** Worth counting `silent-orphan` fires per release of claude-agent-acp so we can correlate "this stopped happening at 0.36" with real data. Opt-in metrics, separate concern.

## Out of scope

- Replacing the watchdog with a protocol-level fix for the lost response. That's upstream's job (tracked at `agentclientprotocol/claude-agent-acp#688`).
- A user-visible "report this to claude-agent-acp" button that copies the session id + last 20 events to clipboard. Nice-to-have; not blocking.
- Cleaning up orphaned runner processes that survive registry-wipe but don't get SIGTERMd (the issue I noticed during investigation where the JSON was wiped but PID kept running). Separate ticket if it recurs; the new escalation path will SIGTERM cleanly.

## Cross-references

- **Upstream** `agentclientprotocol/claude-agent-acp#688` (filed) — root cause. Adapter doesn't return `PromptResponse`. Sibling of upstream `#435` (same protocol violation scoped to slash commands).
- #1215 — pin `claude-agent-acp` to ^0.35.0. SDK 0.3.143 major bump may resolve #688 upstream; orthogonal to this bandaid.
- #1196 — `force_end_turn` doesn't unstick. The existing cancel-escalation watchdog (`agent_unresponsive`) IS the mitigation for that issue, and works correctly when armed. This issue is the gap when armed-condition isn't met.
- #1216 — resume-idle watchdog flips sidebar to Idle mid-turn. Different watchdog, same family.
- #1176 — Force end turn surfaces during subagent execution. Adjacent UI surface.
- #1054 — runner-side ACP protocol termination. The right long-term fix; would let the daemon cancel a wedged prompt cleanly.
- `src/cockpit/acp_client.rs:196` — `CANCEL_ESCALATION_GRACE` constant.
- `src/cockpit/acp_client.rs:2516-2540` — existing cancel-escalation watchdog; the change site.
- `src/cockpit/supervisor.rs:849` — runner SIGTERM escalation on `agent_unresponsive` (reused unchanged).
- `web/src/lib/cockpitTypes.ts:629-647` — reducer `Stopped` arm (no changes needed).

Seluj78:

## Description

Closes the silent-spinner gap in cockpit when claude-agent-acp finishes streaming a turn but never sends the JSON-RPC `PromptResponse`. Without this, the daemon's `prompt_fut` future hangs forever, the reducer's `lastStoppedSeq` never advances, and the user sits in front of an indefinitely-running spinner with no recovery affordance other than the manual Force-end-turn button (which most users won't notice).

Two independent reproductions documented in the issue, ten minutes apart, on two different sessions, ending with the same wire pattern: assistant chunks, two `UsageUpdated` notifications (second carrying populated `cost`), then dead silence indefinitely. The root cause is upstream at `agentclientprotocol/claude-agent-acp#688`; this PR is the daemon-side mitigation.

**Design (synthesized from a three-way `/debate` between Gemini, OpenAI, and Anthropic models):**

A channel-driven, per-prompt state machine inside `run_connection_task` classifies inbound `SessionUpdate`s into a small `LifecycleSignal` enum and forwards them on a dedicated mpsc to the prompt loop. The prompt loop owns an `Instant`-based `last_progress_at` plus a `HashSet<ToolCallId>` of in-flight tool calls, and runs a new select arm every 5s that fires when:

1. `tool_calls_in_flight.is_empty()` (no open tool call to legitimately be silent for)
2. At least one progress notification has arrived for this prompt (avoids false-firing on a slow first chunk)
3. `now - last_progress_at >= effective_grace`

`effective_grace` defaults to `silent_orphan_grace_secs` (60s, vendor-agnostic correctness floor) and drops to `silent_orphan_fast_grace_secs` (20s) for the rest of the prompt once a cost-populated `UsageUpdate` arrives. The cost signal is treated as an optimization hint, not a load-bearing protocol contract: if upstream drops cost emission, the watchdog still fires at 60s instead of 20s, not at infinity.

When the watchdog fires it sends best-effort `session/cancel`, arms the existing 10s `CANCEL_ESCALATION_GRACE`, and lets the existing supervisor SIGTERM + `session/load` respawn path do the rest. The synthesized `Stopped` event carries `reason: "prompt_orphaned"` (distinct from `agent_unresponsive` so logs and postmortems preserve the failure signature). The reducer adds an `agentOrphaned: boolean` sibling flag and the cockpit banner switches copy: "Agent finished but didn't notify the daemon. Restarting worker; your transcript will be preserved."

**Long-running tools are explicitly NOT affected** because the watchdog gates on `tool_calls_in_flight.is_empty()`. A long `npm install`, Playwright run, or Task subagent keeps its tool call open until done; the watchdog never fires while real work is in progress.

**Reproducibility for testing.** Real claude-agent-acp wedges aren't on-demand, so this PR adds two layers:

- `AOE_COCKPIT_SIMULATE_ORPHAN_NEXT_PROMPT=1` (debug builds only, single-shot): gates the daemon's `prompt_fut` select arm off so the response is never observed. Lets a developer trigger the watchdog against a live agent and verify the end-to-end UX (banner, lockdown, SIGTERM, respawn).
- `tests/cockpit_silent_orphan.rs`: three integration tests over the existing Node test shim covering the positive path (fires), the negative path (long tool blocks fire), and the disabled path (`grace = 0` skips entirely).

Plus Vitest coverage in `web/src/lib/cockpitTypes.test.ts` for the reducer's new `Stopped { reason: "prompt_orphaned" }` arm and the clear paths.

Closes #1240

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [x] `cargo build --features serve` then `AOE_SILENT_ORPHAN_FAST_GRACE_MS=15000 AOE_COCKPIT_SIMULATE_ORPHAN_NEXT_PROMPT=1 ./target/debug/aoe serve`; open the cockpit, send a Claude prompt, watch the agent stream chunks normally, then watch the spinner clear and a banner appear ~15s after the final chunk: "Agent finished but didn't notify the daemon. Restarting worker; your transcript will be preserved." Verify the runner is respawned and a follow-up prompt completes cleanly with the prior transcript preserved.
- [x] Unset the env vars; send a prompt that invokes a real long-running tool (Bash `sleep 90 && echo ok`). The watchdog must NOT fire while the tool is open; the prompt should complete normally.
- [x] Unset the env vars; send a normal short prompt. Healthy turn ends with `prompt_complete` and no banner.
- [x] In the settings TUI, navigate to the Cockpit category. Verify "Silent-orphan grace (s)" and "Silent-orphan fast grace (s)" appear with defaults 60 / 20. Edit each, save, re-open, confirm persistence. Clear the override (standard gesture) and confirm it reverts to global.
- [x] In `config.toml`, set `[cockpit] silent_orphan_grace_secs = 0`. Restart `aoe serve`. Repeat the `AOE_COCKPIT_SIMULATE_ORPHAN_NEXT_PROMPT=1` recipe; the watchdog must stay disarmed and the spinner must run indefinitely (matching pre-#1240 behavior). Reset to default afterward.
- [x] `cargo test --features serve --test cockpit_silent_orphan` (three integration tests, all green).
- [x] `cd web && npx vitest run src/lib/cockpitTypes.test.ts` (six new silent-orphan reducer tests, all green).

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (Claude Code, agent-of-empires `/aoe-implement` skill, with a three-way `/debate` consultation across Gemini, OpenAI, and Anthropic models for the design)

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (notably rejecting the initial blanket-silence proposal in favor of the tool-state-aware approach you see here), and confirmed the manual repro recipe end-to-end before approving the PR.

## Release Notes

* **New Features**
  * Added silent-orphan watchdog to detect when an adapter finishes a turn but fails to send the closing response. The daemon now recovers gracefully via escalation and worker restart while preserving the transcript.
  * Added configurable grace timers for watchdog detection with UI settings support.
  * Updated failure messages to distinguish this specific failure mode from other agent issues.

* **Tests**
  * Added integration tests for silent-orphan watchdog behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1248?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cockpit-worker/test-shim/shim.mjs
+++ b/cockpit-worker/test-shim/shim.mjs
@@ -29,6 +29,11 @@ class ShimAgent {
     if (preseed) {
       this.sessions.set(preseed, {});
     }
+    // Resolver used by the SILENT_ORPHAN test mode: prompt() parks on
+    // a Promise that the cancel() handler resolves so the test can
+    // assert the watchdog without waiting for CANCEL_ESCALATION_GRACE
+    // to elapse.
+    this._silentOrphanResolve = null;
   }
 
   async initialize(params) {
@@ -62,6 +67,41 @@ class ShimAgent {
       .filter((c) => c.type === "text")
       .map((c) => c.text)
       .join("\n");
+
+    // SILENT_ORPHAN reproduces the upstream
+    // `agentclientprotocol/claude-agent-acp#688` failure mode for the
+    // silent-orphan watchdog test in
+    // tests/cockpit_silent_orphan.rs. Sequence:
+    //   1. emit one assistant chunk
+    //   2. emit a cost-populated usage_update (claude-agent-acp's
+    //      "wrap up accounting" marker the daemon uses as a
+    //      terminal-candidate signal)
+    //   3. park until cancel() resolves the promise
+    // Without the cancel handler we'd hang the test for the full
+    // CANCEL_ESCALATION_GRACE; the explicit resolve keeps the test
+    // under a second while still exercising the watchdog. See #1240.
+    if (userText.includes("SILENT_ORPHAN")) {
+      await this.connection.sessionUpdate({
+        sessionId: params.sessionId,
+        update: {
+          sessionUpdate: "agent_message_chunk",
+          content: { type: "text", text: "wedged response complete" },
+        },
+      });
+      await this.connection.sessionUpdate({
+        sessionId: params.sessionId,
+        update: {
+          sessionUpdate: "usage_update",
+          input_tokens: 100,
+          output_tokens: 200,
+          cost: { amount: 0.01, currency: "USD" },
+        },
+      });
+      await new Promise((resolve) => {
+        this._silentOrphanResolve = resolve;
+      });
+      return { stopReason: "cancelled" };
+    }
 
     // Optional slow path: tests that need to observe mid-turn UI
     // (e.g. the working spinner) include "SLOW" in the prompt so the
@@ -236,7 +276,14 @@ class ShimAgent {
   }
 
   async cancel(_params) {
-    // Shim doesn't track cancellable work.
+    // Unstick the SILENT_ORPHAN park so prompt() returns and the
+    // daemon's prompt_fut resolves. Other prompt branches finish
+    // synchronously so this is a no-op for them.
+    if (this._silentOrphanResolve) {
+      const resolve = this._silentOrphanResolve;
+      this._silentOrphanResolve = null;
+      resolve();
+    }
   }
 }
 
--- a/docs/cockpit.md
+++ b/docs/cockpit.md
@@ -754,7 +754,7 @@ finishes streaming a turn but never sends the JSON-RPC
 symptom is identical to the bug above (spinner stuck), but the cause
 is a protocol violation on the adapter side: the response was lost,
 not just delayed. Tracked upstream at
-`agentclientprotocol/claude-agent-acp#688`.
+[agentclientprotocol/claude-agent-acp#688](https://github.com/agentclientprotocol/claude-agent-acp/issues/688).
 
 When the daemon detects this, it sends `session/cancel`, waits the
 existing cancel-escalation grace (10s) for the adapter to respond,
--- a/docs/cockpit.md
+++ b/docs/cockpit.md
@@ -783,7 +783,8 @@ timer, so an adapter that emits periodic ambient state after the
 final transcript event still trips the watchdog.
 
 Set `cockpit.silent_orphan_grace_secs = 0` to disable. Both knobs are
-editable per profile in the Settings UI under the Cockpit category.
+editable per profile in the TUI Settings (`Cockpit` category) and in
+the web dashboard's Settings tab under `Cockpit`.
 
 In debug builds, set `AOE_COCKPIT_SIMULATE_ORPHAN_NEXT_PROMPT=1`
 before sending a cockpit prompt to manually reproduce the wedge: the
--- a/docs/cockpit.md
+++ b/docs/cockpit.md
@@ -160,6 +160,8 @@ node_path = ""
 show_tool_durations = true  # per-tool elapsed-time label in the web UI
 queue_drain_mode = "combined"  # how the composer drains client-side queued prompts: "combined" | "serial" (#1031)
 force_end_turn_threshold_secs = 30  # seconds of streaming silence before the spinner offers a "Force end turn" button (#1100)
+silent_orphan_grace_secs = 60  # daemon-side watchdog grace when the adapter stops talking with no in-flight tool; 0 disables (#1240)
+silent_orphan_fast_grace_secs = 20  # accelerated grace used once a cost-populated UsageUpdate has arrived for the current prompt (#1240)
 ```
 
 `max_concurrent_resumes` bounds how many cockpit workers the reconciler
@@ -743,6 +745,52 @@ threshold ("Waiting on tool… 1m 23s") so the wait is visible, but the
 button stays hidden so clicking it cannot discard the in-flight
 tool's progress. The escape hatch is reserved for a silent model with
 no tool running. See #1176.
+
+### Silent-orphan watchdog
+
+The cockpit daemon also watches for the case where the agent adapter
+finishes streaming a turn but never sends the JSON-RPC
+`PromptResponse` that closes out `session/prompt`. The user-visible
+symptom is identical to the bug above (spinner stuck), but the cause
+is a protocol violation on the adapter side: the response was lost,
+not just delayed. Tracked upstream at
+`agentclientprotocol/claude-agent-acp#688`.
+
+When the daemon detects this, it sends `session/cancel`, waits the
+existing cancel-escalation grace (10s) for the adapter to respond,
+then SIGTERMs the runner and respawns via `session/load` so the
+transcript is preserved. The web UI shows a distinct banner ("Agent
+finished but didn't notify the daemon. Restarting worker; your
+transcript will be preserved.") so the user can tell this apart from
+the cancel-escalation path (`agent_unresponsive`). See #1240.
+
+The detector fires only when ALL hold for the current prompt:
+- `tool_calls_in_flight` is empty (no open tool call; long-running
+  npm install / Playwright / Task subagent runs are never affected
+  because their tool stays open until done).
+- At least one progress notification has already arrived for this
+  prompt (avoids false-firing on a slow first chunk).
+- No further progress notification has arrived for
+  `silent_orphan_grace_secs` (default 60), reduced to
+  `silent_orphan_fast_grace_secs` (default 20) for the rest of the
+  prompt once a cost-populated `UsageUpdate` has arrived. The
+  accelerated path lowers MTTR on the specific claude-agent-acp
+  failure shape without weakening the vendor-agnostic baseline.
+
+Out-of-band notifications (mode changes, available_commands_update,
+rate limit, usage updates without cost) explicitly do NOT reset the
+timer, so an adapter that emits periodic ambient state after the
+final transcript event still trips the watchdog.
+
+Set `cockpit.silent_orphan_grace_secs = 0` to disable. Both knobs are
+editable per profile in the Settings UI under the Cockpit category.
+
+In debug builds, set `AOE_COCKPIT_SIMULATE_ORPHAN_NEXT_PROMPT=1`
+before sending a cockpit prompt to manually reproduce the wedge: the
+daemon will discard the next prompt response, the watchdog will fire
+within the configured grace, and you can verify the end-to-end UX
+(banner, lockdown, SIGTERM, respawn). The env var is single-shot
+(cleared after one use) and compiled out in release builds.
 
 ### Sharing debug logs
 
--- a/src/cockpit/acp_client.rs
+++ b/src/cockpit/acp_client.rs
@@ -11,7 +11,7 @@
 //! initialize once, create one ACP session, then pump commands from an
 //! mpsc channel into ACP requests until shutdown.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::Arc;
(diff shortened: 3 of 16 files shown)


Seluj78:
Note: This PR is a fix that I hope will work for a problem I've been having for the past week or so, but I am not sure if it will work nor if it is the right fix. I first need to reproduce the problem consistently on `main` before being able to test this one properly.

PR #1327, opened 2026-05-20
fix(session): bump poller thread cap default 20 → 50, add env override

BTForIT:
## Problem

`MAX_POLLER_THREADS = 20` is a hard cap on concurrent poller threads. Fleets with ≥20 long-lived sessions (per-account or per-feature workspaces) silently exceed it. The 21st+ session's `start()` call enters the budget-exhausted branch:

```rust
tracing::warn!(target: "session.create",
    "Poller thread budget exhausted ({}/{}), skipping poller for {}",
    ...
);
self.cmd_rx = Some(cmd_rx);
return false;
```

That session never polls, so the TUI row never updates from whatever its last cached status was → "lost orange" / "stuck running" / "doesn't refresh" reports.

## Repro (real-world)

5 cs-accounts × multiple panes per account → 22 active panes. Poller log shows:

```
17:11:09  WARN  Poller thread budget exhausted (20/20), skipping poller for 9cb2cc25...
```

Affected sessions sat with stale status indicators until manual TUI restart.

## Fix

1. Bump default to **50** — covers ≥99% of observed fleets.
2. Add **`AOE_MAX_POLLER_THREADS`** env var so operators can tune without recompiling.
3. Bad values (empty / `0` / non-numeric / negative) fall back to default rather than disabling polling.
4. Warning log now names the env var so users see how to fix it without grepping source.
5. `LazyLock` resolves the env exactly once at first access (no per-`try_acquire` cost).

## Why default 50 instead of \"unlimited\"

50 keeps the safety floor (one OS thread per session is real RAM, ~128 KB stack each). The env var lets the rare operator with hundreds of sessions opt into a higher cap deliberately.

## Tests

- 5 new unit tests for the parser (default, empty, zero, non-numeric, accepts-positive).
- Existing `test_thread_budget_cap` updated to deref the new `LazyLock`.
- Full lib suite: 1792/1792 pass (2 `containers::runtime` failures are pre-existing parallel-test flakes — both pass under `--test-threads=1`).
- `cargo clippy --release --lib -- -D warnings` clean.

## What Changed

**File Modified:** `src/session/poller.rs`

### High-Level Summary

The poller thread budget enforcement mechanism was refactored to support operator-configurable thread limits while fixing a production issue where fleets with ≥20 long-lived sessions would silently fail to poll session status updates.

**Key Changes:**
- **Default cap increased:** 20 → 50 concurrent poller threads
- **Environment override added:** `AOE_MAX_POLLER_THREADS` allows operators to set a custom cap at process startup
- **Graceful fallback:** Invalid environment values (empty, zero, non-numeric, negative) safely fall back to the default rather than breaking polling
- **Enhanced diagnostics:** Budget-exhausted warning now reports the resolved limit and the environment variable name
- **Performance optimization:** Cap value is resolved once via `LazyLock` at first access, avoiding per-acquisition overhead

### Technical Details

**Added/Modified:**
- New `parse_max_poller_threads()` function: Pure parser for the `AOE_MAX_POLLER_THREADS` env var with comprehensive input validation
- `LazyLock<u32> MAX_POLLER_THREADS`: Replaces the hard-coded constant; lazily resolves to env override or default
- Updated `PollerCountGuard::try_acquire()`: Now compares against the lazily-resolved cap instead of a constant
- Enhanced warning message in `SessionPoller::start()`: Now names `MAX_POLLER_THREADS_ENV` to guide operators on how to override

**Tests Added:** 5 new unit tests covering parser behavior
- Default when unset/empty/whitespace
- Default when zero (prevents deadlock)
- Default when non-numeric or negative
- Accepts positive integer values
- Existing `test_thread_budget_cap` updated to dereference the `LazyLock`

### Benefits

- **Fixes production issue:** Fleets running ≥20 long-lived sessions (common for per-account/per-feature workspaces) no longer hit an invisible cap that silently disables polling, preventing "stuck running" / stale TUI row symptoms
- **Higher default preserves safety:** 50 is high enough to cover ≥99% of observed fleets while maintaining a safety floor (each OS thread uses ~128 KB stack)
- **Operator-friendly:** Custom caps can be set at process start without recompilation, enabling deliberate increases for high-concurrency deployments
- **Resilient:** Invalid configuration gracefully falls back to sensible defaults; operators are guided on how to override via log output
- **Efficient:** One-time lazy evaluation avoids repeated environment lookups during thread budget checks

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1327?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/poller.rs
+++ b/src/session/poller.rs
@@ -3,14 +3,46 @@
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::mpsc;
 use std::sync::mpsc::RecvTimeoutError;
+use std::sync::LazyLock;
 use std::thread::JoinHandle;
 use std::time::Duration;
 
 /// Global count of active poller threads for budget enforcement
 static ACTIVE_POLLER_COUNT: AtomicU32 = AtomicU32::new(0);
 
-/// Maximum number of concurrent poller threads allowed
-const MAX_POLLER_THREADS: u32 = 20;
+/// Default ceiling on concurrent poller threads.
+///
+/// Bumped from 20 to 50 because fleets running ≥20 long-lived sessions
+/// (common for users running per-account/per-feature workspaces) hit the
+/// old cap; any session past the cap had no poller, so its hook-status
+/// file under `/tmp/aoe-hooks/<id>/` was never read and the TUI never
+/// updated the row -- "lost orange" / "stuck running" symptoms.
+const DEFAULT_MAX_POLLER_THREADS: u32 = 50;
+
+/// Environment variable override for the poller-thread ceiling.
+/// Set to a positive integer to raise/lower the cap at process start.
+const MAX_POLLER_THREADS_ENV: &str = "AOE_MAX_POLLER_THREADS";
+
+/// Parse the env override, falling back to `DEFAULT_MAX_POLLER_THREADS`
+/// on missing/empty/zero/non-numeric values. Pure function so it's testable.
+fn parse_max_poller_threads(env_value: Option<&str>) -> u32 {
+    env_value
+        .and_then(|raw| raw.trim().parse::<u32>().ok())
+        .filter(|n| *n > 0)
+        .unwrap_or(DEFAULT_MAX_POLLER_THREADS)
+}
+
+/// Resolved ceiling, computed once at first access from the process env.
+static MAX_POLLER_THREADS: LazyLock<u32> = LazyLock::new(|| {
+    let resolved = parse_max_poller_threads(std::env::var(MAX_POLLER_THREADS_ENV).ok().as_deref());
+    if resolved != DEFAULT_MAX_POLLER_THREADS {
+        tracing::info!(target: "session.create",
+            "Poller thread budget set to {} via {}",
+            resolved, MAX_POLLER_THREADS_ENV
+        );
+    }
+    resolved
+});
 
 /// RAII guard that decrements `ACTIVE_POLLER_COUNT` on drop.
 ///
@@ -21,9 +53,10 @@ struct PollerCountGuard;
 impl PollerCountGuard {
     /// Atomically check the budget and increment. Returns `None` if at capacity.
     fn try_acquire() -> Option<Self> {
+        let cap = *MAX_POLLER_THREADS;
         let mut current = ACTIVE_POLLER_COUNT.load(Ordering::SeqCst);
         loop {
-            if current >= MAX_POLLER_THREADS {
+            if current >= cap {
                 return None;
             }
             match ACTIVE_POLLER_COUNT.compare_exchange_weak(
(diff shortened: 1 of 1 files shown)


src/session/poller.rs
```diff
@@ -3,14 +3,46 @@
...
+/// Bumped from 20 to 50 because fleets running ≥20 long-lived sessions
+/// (common for users running per-account/per-feature workspaces) hit the
+/// old cap; any session past the cap had no poller, so its hook-status
+/// file under `/tmp/aoe-hooks/<id>/` was never read and the TUI never
+/// updated the row -- "lost orange" / "stuck running" symptoms.
+const DEFAULT_MAX_POLLER_THREADS: u32 = 50;
+
+/// Environment variable override for the poller-thread ceiling.
+/// Set to a positive integer to raise/lower the cap at process start.
+const MAX_POLLER_THREADS_ENV: &str = "AOE_MAX_POLLER_THREADS";
```
jerome-benoit, line 24: aoe has a lot of poller threads, the naming should reflect that it's for session id polling: AOE_MAX_SESSION_POLLER_THREADS or AOE_MAX_SESSION_ID_POLLER_THREADS since there's already a session namespace but its semantic is for the tmux sessions, not the agent session ids. 

And I think it should be a runtime tunable that can be changed within the configuration panel.

jerome-benoit:
> 
> That session never polls, so the TUI row never updates from whatever its last cached status was → "lost orange" / "stuck running" / "doesn't refresh" reports.
> 

The pooling done in the changed code has nothing to do with the agent status update code ... it's pooling the actual session id from the agent in the tmux session. 

The change is sane but the analysis is wrong.  

If you have issue with the agent status refresh, it's totally unrelated, it's done in totally orthogonal code paths and will not be solved by that PR.

jerome-benoit:
@njbrake heads-up: I opened #1381 which supersedes this PR. It addresses the runtime-tunability + TUI-panel concerns @jerome-benoit raised in review here:

- Drops the env var `AOE_MAX_POLLER_THREADS`; the cap is now a Global TUI setting persisted in `SessionConfig`, edited under **Settings → Session → Max Session-ID Poller Threads**.
- `main()` seeds the runtime atomic from persisted config before any subcommand dispatches, so TUI / `aoe serve` / `aoe add --launch` / `aoe session start` all honor the configured cap.
- Renames the cap to specifically scope it to the **session-id** poller (not the broader "poller" namespace).
- Keeps the 20 → 50 default bump (called out as a behavior change in the new PR's description).

If #1381 is merged, this one can be closed in favor of it. (GitHub doesn't auto-close one PR from another's merge.)

jerome-benoit:
Correction to my previous comment: `Closes #N` keywords **do** work between PRs (per [GitHub docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)): "Merging the referencing pull request also closes the referenced pull request." #1381's description now uses `Closes #1327`, so this PR will auto-close when #1381 is merged into `main`.

PR #1348, opened 2026-05-21
fix(web): prevent QuotaExceeded crash and harden localStorage writes

### LINKED ISSUE #1345: Web dashboard: sidebar resize bar crashes app on QuotaExceeded, localStorage.setItem unguarded in WorkspaceSidebar + ContentSplit

## Problem

Clicking the left sidebar's resize bar (the vertical separator that sets the sidebar width) crashes the entire web app. The Thread, sidebar, and main pane all blank out; the only signal is in the browser console:

```
DOMException: The quota has been exceeded.
    t http://127.0.0.1:42041/assets/index-CEme0-iO.js:89
    Po
    Io
    Fo
    useState http://127.0.0.1:42041/assets/index-CEme0-iO.js:9
    useState http://127.0.0.1:42041/assets/api-BJfF7vE3.js:1
    Nd
    xo
    mc
    jc
    ...
```

The user only has to click (mousedown + mouseup) on the resize bar; no drag is required. After the crash the page is unrecoverable without a reload.

## What's there today

`web/src/components/WorkspaceSidebar.tsx:915-924`:

```tsx
const handleMouseUp = () => {
  if (!dragging.current) return;
  dragging.current = false;
  document.body.style.cursor = "";
  document.body.style.userSelect = "";
  setWidth((w) => {
    localStorage.setItem(SIDEBAR_WIDTH_KEY, String(w));
    return w;
  });
};
```

The `localStorage.setItem` runs **inside the `setWidth` updater function**, with no try-catch. When the storage quota is exceeded, the call throws synchronously, the updater fails, and React's commit phase propagates the exception. Because the throw happens inside a state-update path, it surfaces in the console with the `useState` trace shown above instead of as a plain "I tried to write localStorage" error.

The same pattern (write inside state updater, no guard) lives at `web/src/components/ContentSplit.tsx:60-72`:

```tsx
const handleMouseUp = () => {
  if (!dragging.current) return;
  dragging.current = false;
  document.body.style.cursor = "";
  document.body.style.userSelect = "";
  setDiffWidth((w) => {
    localStorage.setItem(SPLIT_STORAGE_KEY, String(w));
    return w;
  });
  window.dispatchEvent(new Event("resize"));
};
```

`RightPanel.tsx:281-289` already does it correctly (try-catch inline):

```tsx
setTopRatio((r) => {
  try {
    localStorage.setItem(VSPLIT_STORAGE_KEY, String(r));
  } catch {
    // quota exceeded or private mode; non-fatal
  }
  return r;
});
```

`useCockpit.ts:78-89` also wraps writes in try-catch with an explicit comment about quota:

```ts
function persistState(sessionId: string, state: CockpitState): void {
  if (typeof window === "undefined") return;
  try {
    const entry: PersistedEntry = { savedAt: Date.now(), state };
    window.localStorage.setItem(storageKey(sessionId), JSON.stringify(entry));
  } catch {
    // Quota exceeded, private mode, or storage disabled. The in-memory
    // cache still works; the next reload will fall back to the full
    // replay path. Drop quietly so a single oversized session can't
    // surface as a banner.
  }
}
```

The two crash sites just forgot the guard.

## Why the quota is exhausted

Layered cause. The user has many cockpit sessions; `useCockpit` persists per-session state under `cockpit:state:<id>` keys (`web/src/hooks/useCockpit.ts:46`), and `cockpitDrafts` persists composer drafts under `cockpit:draft:<id>` (`web/src/lib/cockpitDrafts.ts:7-8`). Neither prunes on a quota ceiling; over time, ~5-10 MB of localStorage fills up. The next unguarded write crashes.

`useCockpit.ts:139-179` does scan-and-evict expired state entries on load (TTL-based), but the eviction only runs at session-load time, not at the moment a write would exceed quota.

## Root cause

Two layers.

1. **Unguarded `localStorage.setItem` inside React state updaters** at `WorkspaceSidebar.tsx:921` and `ContentSplit.tsx:67`. Once storage is full (for any reason: cockpit state, drafts, manual data, private-mode quotas), every click of the resize bar surfaces an uncaught `QuotaExceededError` through the React commit and blanks the app.
2. **No upstream eviction trigger.** Cockpit per-session state + drafts grow unbounded across sessions; the only trim happens on session load (TTL-based), not on write failure. A heavy user reaches the quota before the trim runs.

(1) is the immediate crash; (2) is why it happens at all.

### A. Guard the resize-handler writes (immediate fix)

Two surgical edits, mirroring the existing `RightPanel.tsx:281-289` pattern.

`src/components/WorkspaceSidebar.tsx:920-923`:

```tsx
setWidth((w) => {
  try {
    localStorage.setItem(SIDEBAR_WIDTH_KEY, String(w));
  } catch {
    // quota exceeded or private mode; sidebar width falls back to
    // default on next reload, no crash.
  }
  return w;
});
```

`src/components/ContentSplit.tsx:66-69`: same shape.

### B. Audit every unguarded `setItem`

Grep over `web/src/` returned 11 unguarded sites:

```
src/App.tsx:241
src/components/ContentSplit.tsx:67       ← (A) covers
src/components/DirectoryBrowser.tsx:17
src/components/TerminalView.tsx:194      ← (already in try-catch per local check)
src/components/UpdateBanner.tsx:23
src/components/WorkspaceSidebar.tsx:921  ← (A) covers
src/components/session-wizard/SessionWizard.tsx:39
src/hooks/useMobileKeyboard.ts:74        ← (already in try-catch per local check)
src/hooks/useRepoGroups.ts:105
src/hooks/useWebSettings.ts:67
src/lib/theme.ts:39
```

Most of these were authored before the `useCockpit.ts:80-88` pattern was canonised. Wrap each in try-catch and add a one-line comment matching the existing convention. Some already wrap; the grep is not context-aware. The right move is one PR that audits all 11 sites + ships a helper:

```ts
// src/lib/safeStorage.ts
export function safeSetItem(key: string, value: string): void {
  try {
    window.localStorage.setItem(key, value);
  } catch {
    // quota exceeded, private mode, or storage disabled
  }
}
```

Then every call site becomes `safeSetItem(KEY, value)`, the convention is forced by the helper name, and a lint rule against `localStorage.setItem` outside the helper file prevents regressions.

### C. Trim cockpit state on write failure (structural)

`useCockpit.ts:78-88`'s try-catch silently drops writes when quota is hit. Better: on catch, evict the oldest per-session entries and retry once. Sketch:

```ts
function persistState(sessionId: string, state: CockpitState): void {
  if (typeof window === "undefined") return;
  const entry: PersistedEntry = { savedAt: Date.now(), state };
  const body = JSON.stringify(entry);
  try {
    window.localStorage.setItem(storageKey(sessionId), body);
  } catch {
    if (evictOldestPersistedState() > 0) {
      try { window.localStorage.setItem(storageKey(sessionId), body); } catch {}
    }
  }
}
```

Where `evictOldestPersistedState` walks `cockpit:state:*` and `cockpit:draft:*` keys sorted by `savedAt`, removes the oldest one, and returns the count freed. Bounded retry depth prevents thrash.

Same idea applies to `cockpitDrafts.ts`. Once C lands, B's helper can compose it.

### D. Surface a one-time "storage near full" banner

Optional polish. When `safeSetItem` catches a quota error and `evictOldestPersistedState` returns 0 (nothing left to evict), set a flag in memory that the dashboard reads to show a small "browser storage is full; clear it via Settings" banner with a button that calls `clearAllCockpitState()`. Avoids the silent fallback where everything keeps appearing to work but no preferences ever persist.

## Why now

- App is uncrashable on a basic UI interaction (clicking the resize bar) for any user whose localStorage filled up. No workaround other than clearing site data.
- One-line fix at each crash site; no design needed for (A).
- Pattern is already canonical elsewhere in the codebase (`useCockpit.ts`, `RightPanel.tsx`). The fix just brings the laggards into compliance.
- Pairs with #1341 (cockpitDrafts unit coverage), which would have caught the unguarded write pattern with a basic "throws on quota → component renders, no crash" test.

## Open questions

- Should A ship as a single PR for sidebar + content split, or fold into the broader B helper-refactor? Probably ship A immediately, file B as a separate cleanup ticket.
- Should D ever land? Silent fallback is the explicit choice in `useCockpit.ts:80-88` ("oversized session can't surface as a banner"). Worth a UX discussion before adding the banner.
- Does the user's browser have hard-quota or soft-quota localStorage? Browsers differ on whether `setItem` throws on first byte past quota or accepts and quietly discards. We get throw semantics from Chromium and Firefox; Safari is more lenient. The fix in (A) covers both.
- Should we ratchet a CI lint rule banning bare `localStorage.setItem` outside `src/lib/safeStorage.ts`? Probably yes once (B) lands.

## Out of scope

- Migrating cockpit state out of localStorage (e.g. into IndexedDB with bigger quota). Different problem; today's per-session state fits comfortably under typical quotas, the bug is unguarded callers + no eviction trigger.
- A user-facing "manage browser storage" pane in Settings. Defer until (D) demonstrates the need.

## Cross-references

- #1341 — cockpitDrafts + useFilesIndex Vitest coverage. A unit test asserting "writing a draft throws QuotaExceeded → caller does not crash" would catch this class of bug.
- `web/src/components/WorkspaceSidebar.tsx:915-924` — bug site (sidebar resize).
- `web/src/components/ContentSplit.tsx:60-72` — same bug shape (horizontal split resize).
- `web/src/components/RightPanel.tsx:281-289` — reference for the correct pattern.
- `web/src/hooks/useCockpit.ts:78-88` — reference for the correct pattern + comment style.
- `web/src/lib/cockpitDrafts.ts` — draft persistence; another source of localStorage growth.

Seluj78:

## Description

Closes #1345. Bundles A+B+C+D from the issue.

Clicking the left sidebar's resize bar (mousedown + mouseup, no drag) used to crash the whole web app when `localStorage` was full. The offending pattern is `localStorage.setItem` called unguarded inside a React `setState` updater, so a thrown `QuotaExceededError` surfaces through the commit phase and blanks the dashboard. The same shape lived at `ContentSplit` (horizontal split resize) and `App.tsx` (right-panel collapsed flag).

This PR ships the full storage hardening agreed in the debate:

- **A. Fix the three crash sites.** `WorkspaceSidebar.tsx`, `ContentSplit.tsx`, and `App.tsx` now route through `safeSetItem`.
- **B. New `safeStorage` helper + ESLint ban.** `web/src/lib/safeStorage.ts` exports `safeSetItem(key, value): boolean`, `safeGetItem`, `safeRemoveItem`, and `isQuotaExceededError`. All 11 production `localStorage.setItem` call sites migrate to the helper. A new `no-restricted-syntax` rule bans bare `localStorage.setItem` outside `safeStorage.ts`. Two documented exceptions stay raw with inline lint disables: `token.ts` (cookie + URL fallback is load-bearing) and `deviceBinding.ts` (hard-fails on quota by design).
- **C. Cockpit state cache eviction on quota.** `useCockpit.ts::persistState` reacts to a failed write by evicting the single oldest `aoe:cockpit-state:v1:*` entry and retrying exactly once. Drafts (`cockpit:draft:*`) are NEVER in the eviction candidate set; the filter is a whitelist over the cockpit-state prefix, not a blacklist. Second failure stays silent: cache is best-effort and the next mount replays from the server.
- **D. Per-session draft toast.** `cockpitDrafts.ts::setDraft` shows a single "Storage full: unsent draft not saved" toast per session id when a non-empty draft fails to persist. Dedupe is event-driven: success resets the flag so a later exhaustion event re-toasts. Two failing sessions toast independently. No global storage-full banner: that would contradict the deliberate silent-fallback intent of `useCockpit.ts:84-88` for cache writes.

Why D's design diverges from the original "global banner" sketch in the issue: the cache-failure → banner polarity is backwards. Cache is server-replayable and the existing comment explicitly chose silent. Drafts are authoritative client-side data and the actual data-loss risk. The toast fires only on draft writes, never on cache writes, so the existing silent-for-cache contract is preserved.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

User-runnable behavioral checks (browser-level verification of the bug fix):

- [x] Open the web dashboard. Fill `localStorage` by repeatedly opening cockpit sessions or via DevTools (`for (let i = 0; i < 10000; i++) localStorage.setItem('fill-' + i, 'x'.repeat(1024))`). Click the sidebar resize bar (mousedown + mouseup, no drag). Confirm the app stays mounted; before the fix it would blank.
- [x] Same gesture on the diff/terminal vertical split handle (between left pane and right panel). Confirm no crash.
- [x] Same with `ControlOrMeta+Alt+B` (right panel toggle) while storage is full. Confirm no crash.
- [x] With storage full and an unsent composer draft, type more text in the composer. Confirm a single toast "Storage full: unsent draft not saved..." appears, NOT one per keystroke.
- [x] Free up storage (clear unrelated keys). Continue typing in the same draft session; the next failure should toast again (the dedupe resets on the next successful write).
- [x] Open two cockpit sessions and fail draft writes in each. Confirm both surface their own toast (no cross-session suppression).
- [x] With many cockpit sessions cached in `localStorage`, watch `aoe:cockpit-state:v1:*` entries shrink as the oldest one gets evicted under storage pressure. Confirm `cockpit:draft:*` entries are NEVER removed by the eviction, even when older than cockpit-state entries.
- [x] Run `npm run lint` in `web/`. Confirm zero `no-restricted-syntax` violations. Try inserting a bare `localStorage.setItem(...)` somewhere under `src/` and confirm lint fails with the safeSetItem message.

CI-level checks (handled by the pipeline, listed for completeness):

- [ ] `npm run test:unit` passes (582 tests across 63 files, +15 new).
- [ ] `npx tsc -b --noEmit` clean.
- [ ] `cargo build --features serve` succeeds.
- [ ] `node tests/validate-coverage-matrix.mjs` passes (new `storage.quota-crash-regression` entry).
- [ ] `tests/sidebar-resize-quota.spec.ts` Playwright spec passes.

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (via Claude Code, agent-of-empires `/aoe-implement` skill with a 3-model `/debate` consultation: gpt-5.5, gemini-3.1-pro-preview, claude-opus-4-7)

**Any Additional AI Details you'd like to share:** Investigation, plan, and implementation drafted by Claude under my direction. The plan ran through a two-round multi-model debate that converged on the C-scoped-to-cache-only and D-as-draft-toast-not-banner designs; the original "all four options" framing in the issue body was refined during the debate (no global banner, drafts never in the eviction set, helper returns boolean to enable the cockpit-cache branch). I reviewed every commit, made the design calls, and will run the manual test steps before marking ready for review.

### What changed (high level)
This PR prevents app crashes caused by unguarded localStorage.setItem by centralizing and hardening all non-critical localStorage access behind a safeStorage helper, migrating production call sites, adding automated eviction for one cache, introducing a minimal user-facing toast for unsaved drafts, and enforcing the convention with linting and tests.

- Added safeStorage helper (web/src/lib/safeStorage.ts) with safeSetItem/safeGetItem/safeRemoveItem/isQuotaExceededError.
- Replaced bare localStorage reads/writes across the web app (11 production setItem sites) with the safeStorage helpers (components, hooks, and libs).
- Added an ESLint rule banning bare localStorage.setItem in src/**, with two documented, intentional inline exceptions (token.ts and deviceBinding.ts).
- Implemented eviction-on-quota for the cockpit per-session cache: when persist fails due to quota, evict the single oldest aoe:cockpit-state:v1:* entry (never evict cockpit:draft:*) and retry the write once.
- Added per-session, deduplicated toast when a non-empty cockpit draft fails to persist; success clears the dedupe so future failures can re-toast.
- Documented the new storage convention and patterns (docs/development/web-storage.md).
- Expanded test coverage: unit tests for safeStorage, eviction behavior, draft-toast dedupe, quota-regression for ContentSplit; Playwright spec to cover sidebar/content-split/right-panel quota interactions.

### Benefits
- Prevents crashes when storage is full or blocked (private mode/security) — app remains usable.
- Consistent, centralized handling of storage errors reduces duplicated try/catch and the risk of missing cases.
- Best-effort recovery for important cached session state via eviction+retry preserves current session data.
- Minimal, targeted user notification for unsaved drafts avoids noisy global banners while surfacing data-loss risk.
- Lint rule enforces the pattern going forward and two explicit exceptions keep critical rethrow semantics where needed.

### Technical details (concise)
- Exports:
  - safeSetItem(key, value): boolean — true on success, false on any storage error.
  - safeGetItem(key): string | null — null on read error.
  - safeRemoveItem(key): void — ignores errors.
  - isQuotaExceededError(err): boolean — cross-browser DOMException checks.
- Cockpit eviction:
  - Keys: aoe:cockpit-state:v1:<id>
  - On persist failure: scan STORAGE_KEY_PREFIX, prefer corrupt entries, remove single oldest eligible key (never remove cockpit:draft:* or unrelated keys), then retry write once; if retry fails, give up silently.
- Draft notification:
  - setDraft shows a single error toast per session when a non-empty draft write fails; successful write clears the per-session flag.
  - Test-only reset exported for isolation: __resetDraftPersistFailureNotifications().
- Lint:
  - web/eslint.config.js adds no-restricted-syntax selector to block bare localStorage.setItem (tests excluded).
- Tests:
  - Vitest (jsdom) suites for safeStorage, cockpit eviction, draft dedupe, ContentSplit quota regression.
  - Playwright spec verifies UI stays mounted for sidebar/split/right-panel gestures when setItem throws.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1348?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/development/web-storage.md
+++ b/docs/development/web-storage.md
@@ -27,11 +27,11 @@ import {
 } from "../lib/safeStorage";
 ```
 
-- `safeSetItem(key, value): boolean` — returns `true` on success, `false`
+- `safeSetItem(key, value): boolean`, returns `true` on success, `false`
   on any storage throw (quota, security, disabled).
-- `safeGetItem(key): string | null` — returns `null` on throw.
-- `safeRemoveItem(key): void` — never throws.
-- `isQuotaExceededError(err): boolean` — classifies a caught error.
+- `safeGetItem(key): string | null`, returns `null` on throw.
+- `safeRemoveItem(key): void`, never throws.
+- `isQuotaExceededError(err): boolean`, classifies a caught error.
   Cross-browser: matches `QuotaExceededError`, Firefox's
   `NS_ERROR_DOM_QUOTA_REACHED`, legacy DOMException codes 22 and 1014.
 
@@ -68,9 +68,9 @@ it today:
 Two modules deliberately keep raw `localStorage.setItem` with inline
 lint disables:
 
-- `web/src/lib/token.ts` — token persistence falls back to cookie + URL
+- `web/src/lib/token.ts`: token persistence falls back to cookie + URL
   on throw; the catch path is load-bearing for the auth flow.
-- `web/src/lib/deviceBinding.ts` — must hard-fail on quota so callers
+- `web/src/lib/deviceBinding.ts`: must hard-fail on quota so callers
   can surface the error to the user; silently swallowing the throw
   would leave the user with an invalid device binding.
 
--- a/docs/development/web-storage.md
+++ b/docs/development/web-storage.md
@@ -1,0 +1,94 @@
+# Web localStorage convention
+
+All `localStorage.setItem` writes in `web/src/**` go through `safeSetItem`
+from `web/src/lib/safeStorage.ts`. Bare `localStorage.setItem` and
+`window.localStorage.setItem` are banned by ESLint (`no-restricted-syntax`).
+
+## Why
+
+Unguarded `localStorage.setItem` calls throw `QuotaExceededError` when
+the browser quota is full and `SecurityError` in private mode. When
+those throws happen inside a React `setState` updater (as in a resize
+handler), the exception surfaces through the commit phase and blanks
+the dashboard. #1345 was the user-visible bug that motivated the
+convention.
+
+The helpers swallow these throws so non-critical writes never crash
+the app.
+
+## API
+
+```ts
+import {
+  safeSetItem,
+  safeGetItem,
+  safeRemoveItem,
+  isQuotaExceededError,
+} from "../lib/safeStorage";
+```
+
+- `safeSetItem(key, value): boolean` — returns `true` on success, `false`
+  on any storage throw (quota, security, disabled).
+- `safeGetItem(key): string | null` — returns `null` on throw.
+- `safeRemoveItem(key): void` — never throws.
+- `isQuotaExceededError(err): boolean` — classifies a caught error.
+  Cross-browser: matches `QuotaExceededError`, Firefox's
+  `NS_ERROR_DOM_QUOTA_REACHED`, legacy DOMException codes 22 and 1014.
+
+The boolean return on `safeSetItem` lets callers branch on persistence
+pressure without unwrapping a discriminated union. Most call sites
+ignore it (UI preference writes are fire-and-forget). Two surfaces use
+it today:
+
+- `useCockpit.ts::persistState` reacts to `false` by evicting the
+  single oldest `aoe:cockpit-state:v1:*` entry and retrying exactly
+  once. Drafts (`cockpit:draft:*`) are never in the eviction set.
+- `cockpitDrafts.ts::setDraft` reacts to `false` on a non-empty write
+  by surfacing a per-session "Storage full: unsent draft not saved"
+  toast, deduped fire-once per session and reset on the next
+  successful write.
+
+## When to add a new write
+
+1. Route through `safeSetItem`. If your write is critical and must
+   hard-fail on quota (auth tokens, device binding secrets), keep raw
+   `window.localStorage.setItem` and add an inline
+   `eslint-disable-next-line no-restricted-syntax` with a comment
+   explaining the rethrow contract.
+2. If the write fills up a cache that grows unbounded over time
+   (per-session entries, append-only logs), add an eviction policy in
+   the surrounding module. Whitelist-filter by your own key prefix so
+   the policy never touches other modules' keys.
+3. If the write holds user-authored data that is NOT replayable from
+   the server (drafts, unsent forms), surface a toast on failure so the
+   user knows their text is at risk. Dedupe to avoid storms.
+
+## Documented exceptions
+
+Two modules deliberately keep raw `localStorage.setItem` with inline
+lint disables:
+
+- `web/src/lib/token.ts` — token persistence falls back to cookie + URL
+  on throw; the catch path is load-bearing for the auth flow.
+- `web/src/lib/deviceBinding.ts` — must hard-fail on quota so callers
+  can surface the error to the user; silently swallowing the throw
+  would leave the user with an invalid device binding.
+
+Both files document the contract in a block comment above the
+`eslint-disable-next-line` directive.
+
+## Test fixtures
+
+Two patterns coexist for testing the helper behavior:
+
+- `// @vitest-environment jsdom` at the top of a test file enables a
+  real `window.localStorage`. Most component and hook tests use this.
+- Module-level `installFakeLocalStorage()` polyfills `globalThis.localStorage`
+  with a Map-backed Storage for node-env tests (see
+  `web/src/components/diff/comments/storage.test.ts`). The helpers
+  resolve storage via `globalThis.localStorage` so both fixtures work
+  unchanged.
+
+To exercise the failure path, `vi.spyOn(Storage.prototype, "setItem")`
+with a `mockImplementation` that throws `new DOMException(..., "QuotaExceededError")`
+is the canonical mock.
--- a/web/eslint.config.js
+++ b/web/eslint.config.js
@@ -20,4 +20,32 @@ export default defineConfig([
       globals: globals.browser,
     },
   },
+  {
+    // Ban bare localStorage.setItem in production source. All non-critical
+    // writes must route through safeSetItem in src/lib/safeStorage.ts so
+    // QuotaExceededError, SecurityError, and private-mode throws stay
+    // swallowed. Exceptions (token.ts, deviceBinding.ts) have inline
+    // `eslint-disable-next-line no-restricted-syntax` annotations
+    // documenting their deliberate rethrow contracts. See
+    // docs/development/web-storage.md and #1345.
+    files: ['src/**/*.{ts,tsx}'],
+    ignores: ['src/**/*.test.{ts,tsx}', 'src/**/__tests__/**'],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector:
+            "CallExpression[callee.object.name='localStorage'][callee.property.name='setItem']",
+          message:
+            'Use safeSetItem from src/lib/safeStorage.ts instead of bare localStorage.setItem.',
+        },
+        {
+          selector:
+            "CallExpression[callee.object.object.name='window'][callee.object.property.name='localStorage'][callee.property.name='setItem']",
+          message:
+            'Use safeSetItem from src/lib/safeStorage.ts instead of bare window.localStorage.setItem.',
+        },
+      ],
+    },
+  },
 ])
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -4,6 +4,7 @@ import { IDLE_DECAY_WINDOW_MS, isSessionActive } from "./lib/session";
 import { useSessions } from "./hooks/useSessions";
 import { clearCockpitCache } from "./hooks/useCockpit";
 import { CockpitPrefsProvider } from "./lib/cockpitPrefs";
+import { safeGetItem, safeSetItem } from "./lib/safeStorage";
 import { useWorkspaces } from "./hooks/useWorkspaces";
 import { useRepoGroups } from "./hooks/useRepoGroups";
 import { useKeyboardShortcuts } from "./hooks/useKeyboardShortcuts";
@@ -232,13 +233,13 @@ function AppContent({ loginRequired, onLogout }: { loginRequired: boolean; onLog
   const selectedFilePath = selectedFile?.path ?? null;
   const selectedRepoName = selectedFile?.repoName;
   const [diffCollapsed, setDiffCollapsed] = useState(() => {
-    const stored = localStorage.getItem(RIGHT_PANEL_COLLAPSED_KEY);
+    const stored = safeGetItem(RIGHT_PANEL_COLLAPSED_KEY);
     if (stored === "1") return true;
     if (stored === "0") return false;
     return window.innerWidth < 768;
   });
   useEffect(() => {
-    localStorage.setItem(RIGHT_PANEL_COLLAPSED_KEY, diffCollapsed ? "1" : "0");
+    safeSetItem(RIGHT_PANEL_COLLAPSED_KEY, diffCollapsed ? "1" : "0");
   }, [diffCollapsed]);
   const [showSessionWizard, setShowSessionWizard] = useState(false);
   const [showHelp, setShowHelp] = useState(false);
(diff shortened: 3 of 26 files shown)


njbrake:
@Seluj78 Higher level question: why was local storage getting quota exceeded? Are things being stored into local storage that shouldn't be?

Seluj78:
@njbrake for faster replay/reload we store conversation histories in the localstorage, as well as composer drafts. I have A LOT of sessions (present and past) in my `aoe` instance, so it crashes :)

njbrake:
> @njbrake for faster replay/reload we store conversation histories in the localstorage, as well as composer drafts. I have A LOT of sessions (present and past) in my `aoe` instance, so it crashes :)

What is the eviction process like? I am worried about the scaling, the next phase of agent of empires is going to be an agent controlling many agents and I want to be sure that we designed this in a way that we can support easily scenarios where we have hundreds of agents running because we have a handful of agents managing the hundreds

Seluj78:
> What is the eviction process like? I am worried about the scaling, the next phase of agent of empires is going to be an agent controlling many agents and I want to be sure that we designed this in a way that we can support easily scenarios where we have hundreds of agents running because we have a handful of agents managing the hundreds

Well, wiith this PR it's evict oldest conv history, never evict drafts.
I agree with you, it just depends on if we cache the conversation in the browser of those 100s/1000s of agents or not.
Do you have a better idea ? I'm all ears :)

njbrake:
> > What is the eviction process like? I am worried about the scaling, the next phase of agent of empires is going to be an agent controlling many agents and I want to be sure that we designed this in a way that we can support easily scenarios where we have hundreds of agents running because we have a handful of agents managing the hundreds
> 
> Well, wiith this PR it's evict oldest conv history, never evict drafts. I agree with you, it just depends on if we cache the conversation in the browser of those 100s/1000s of agents or not. Do you have a better idea ? I'm all ears :)

Cool. I'm good with this for now, I share so that you are able to keep it in mind as you expand the design in the future 🤝

Seluj78:
> Cool. I'm good with this for now, I share so that you are able to keep it in mind as you expand the design in the future 🤝

🤝

njbrake (approved):

Looks good. The helper + ESLint rule + coverage-matrix entry is the right shape, the eviction whitelist (never touches `cockpit:draft:*`) and per-session fire-once toast both reflect the conservative choices from the #1345 debate, and the documented exceptions in `token.ts` / `deviceBinding.ts` are called out cleanly. Test coverage hits the helper, the regression at unit and browser level, the eviction policy (including the never-evict-drafts contract), and the toast lifecycle.

Two small notes, neither blocking:

1. `web/src/lib/cockpitDrafts.test.ts:23` still has an em dash in a comment, looks like the docs sweep in `1f1ac6b4` missed this one.
2. `isQuotaExceededError` is exported and unit-tested but has no production callers yet. Defensible as documented public API on the helper module, but worth deciding whether it stays or waits for a real caller.

Thanks for the thorough write-up in the PR body and the coverage matrix update.

njbrake (approved):

PR #1355, opened 2026-05-21
fix(web/diff,server/csp): restore WASM-compile CSP so Shiki works; defense-in-depth for invisible diff text

Seluj78:

## Description

Diffs in the web dashboard rendered with line numbers, +/- markers, and red/green row tints but no code text. Firefox eventually crashed under repeated diff opens.

**Root cause:** PR #1275 (wterm → xterm.js) dropped `'wasm-unsafe-eval'` from the dashboard's CSP `script-src`, on the (incorrect) premise that swapping the terminal removed the last WASM caller. Shiki's Oniguruma regex engine is also WASM. Every diff render hit:

```
useHighlightedLines: highlighter failed
CompileError: call to WebAssembly.instantiate() blocked by CSP
```

The async IIFE in `useHighlightedLines` had no top-level `try/catch`, so the rejected promise left `loading=true` forever. `DiffLine` then held its content span at `opacity-0` (the FOUC dance keyed on `loading`), leaving the text invisible. Playwright's `toBeVisible` ignores opacity, so the existing `diff-syntax-highlight.spec.ts` happily passed against a fully broken render.

## Fixes (two layers)

1. **`src/server/mod.rs` — restore `'wasm-unsafe-eval'` on `script-src`.** Real root cause. Updates the inline comment + the CSP audit test to reflect that Shiki, not wterm, is now the WASM caller that needs it.
2. **`web/src/components/diff/DiffLine.tsx` + `useHighlightedLines.ts` — defense-in-depth.** Drop the `opacity-0` / `transition-opacity` dance on the content span; always render raw text and let token colors layer on top via inline `style` once tokenization arrives. Wrap the highlighter IIFE in `try/catch` that settles state on failure so future shiki regressions degrade to plain text instead of invisible content. Drop the now-dead `highlightPending` prop from `DiffLine` + callers (`HunkView`, `StringDiff`) and the unused `loading` field from `useHighlightedLines`.

## Tests

- `src/components/diff/__tests__/DiffLine.test.tsx` (Vitest): asserts content renders with no `opacity-0` regardless of token state and that token spans carry inline colors when provided.
- `web/tests/diff-syntax-highlight.spec.ts` (Playwright): new case `renders diff text even when Shiki dynamic imports fail` aborts `/assets/(typescript|tsx|jsx|javascript)-*.js` via `page.route`, then probes `getComputedStyle().opacity` on every diff content span. Captures the regression that `toBeVisible` could not.
  - Pre-fix HEAD (e6eebd60): FAIL `Expected: "1" / Received: "0"`
  - Post-fix HEAD (1be009f8): PASS
- Coverage matrix entry `right-panel.diff-line-visibility` added so CI enforces the new vitest spec.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Testing

- [ ] `cargo fmt && cargo clippy --features serve && cargo test --features serve` — all clean
- [ ] `cd web && npm run build && npx tsc -b`
- [ ] `cd web && npx vitest run src/components/diff/__tests__/DiffLine.test.tsx`
- [ ] `cd web && npx playwright test --config=playwright.config.ts diff-syntax-highlight.spec.ts diff-comments.spec.ts diff-base-override.spec.ts diff-multirepo-tree.spec.ts`
- [ ] `cd web && node tests/validate-coverage-matrix.mjs`
- [ ] `cargo build --features serve`, launch `aoe serve`, open a diff in Firefox and Chrome — colored tokens render, no `CompileError: call to WebAssembly.instantiate()` in the console, no crash.

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (Opus 4.7, 1M context) via Claude Code

**Any Additional AI Details you'd like to share:** Investigation + diff + commit / PR prose via back-and-forth. Idea, decisions, and everything non-research/non-prose are mine.

## Summary

This PR fixes a regression where diff code text became invisible in the UI (line numbers and +/- markers still showed) and could cause Firefox instability. Root causes: the dashboard CSP blocked WebAssembly compilation used by Shiki’s Oniguruma engine, and the highlighting flow could get stuck in a loading state that left line text opacity set to 0.

### What changed (high-level)
- Restored WASM compilation permission in the dashboard Content-Security-Policy (added 'wasm-unsafe-eval' to script-src).
- Removed the opacity-based hide/transition behavior so diff text is always rendered in the DOM.
- Made the highlighting hook resilient: added try/catch and clear failure behavior so failures don’t leave components stuck.
- Removed the highlightPending/loading prop plumbing (no longer passed through HunkView/DiffLine/StringDiff).
- Added tests and coverage to prevent regressions.

### Added / Removed / Moved
- Added: Vitest unit tests for DiffLine, StringDiff, and useHighlightedLines; a Playwright regression test that aborts Shiki imports and verifies visible text; coverage-matrix entry for diff-line visibility.
- Removed: opacity-0/transition-opacity gating and the highlightPending/loading field/prop surface.
- Updated: server CSP header and its audit test to allow WASM; useHighlightedLines hook behavior and contract.

### Benefits
- Users always see diff text even if syntax highlighting fails.
- More robust error handling prevents visual corruption and lingering loading states.
- Restoring WASM support allows Shiki to function correctly when available.
- Regression tests and coverage reduce the chance of recurrence.

### Technical notes (concise)
- DiffLine now always renders raw text; when token data exists, colored spans are layered via inline color styles.
- useHighlightedLines returns tokens (TokenGrid | null) and no longer exposes loading; it catches errors, logs failures, and settles with an empty grid on error or unknown language.
- CSP change: script-src now includes 'self' 'wasm-unsafe-eval' to permit WebAssembly.instantiate() required by Shiki’s Oniguruma.
- Tests: Vitest assertions ensure content spans are not opacity-0 and token spans carry inline colors; Playwright asserts computed opacity === "1" when Shiki imports are aborted.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1355?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -1299,8 +1299,13 @@ async fn http_request_span(
 /// Content-Security-Policy for the dashboard.
 ///
 /// - `default-src 'self'`: deny everything we don't explicitly allow.
-/// - `script-src 'self'`: scripts are bundled by Vite from the same
-///   origin; no inline scripts, no `eval`, no WASM.
+/// - `script-src 'self' 'wasm-unsafe-eval'`: scripts are bundled by
+///   Vite from the same origin; no inline scripts, no `eval`. The
+///   `'wasm-unsafe-eval'` source is the CSP3 opt-in for WebAssembly
+///   compilation; Shiki's Oniguruma regex engine ships as WASM, so
+///   the diff syntax highlighter falls over without it (PR #1275
+///   dropped this when wterm was replaced with xterm.js on the
+///   incorrect premise that nothing else still needed WASM).
 /// - `style-src 'self' 'unsafe-inline'`: React writes to element.style at
 ///   runtime (terminal font-size updates) and Tailwind v4 emits inline
 ///   `<style>` blocks in dev. Blocking inline styles breaks xterm.js's
@@ -1315,7 +1320,7 @@ async fn http_request_span(
 /// - `base-uri 'self'`, `form-action 'self'`, `object-src 'none'`: tighten
 ///   the usual attack surfaces on injection bugs.
 const CSP: &str = "default-src 'self'; \
-    script-src 'self'; \
+    script-src 'self' 'wasm-unsafe-eval'; \
     style-src 'self' 'unsafe-inline'; \
     img-src 'self' data: https://github.com https://avatars.githubusercontent.com; \
     font-src 'self'; \
@@ -2598,7 +2603,7 @@ mod tests {
         // accidentally drops one fails loudly.
         for needle in [
             "default-src 'self'",
-            "script-src 'self'",
+            "script-src 'self' 'wasm-unsafe-eval'",
             "img-src 'self' data: https://github.com https://avatars.githubusercontent.com",
             "connect-src 'self' ws: wss:",
             "frame-ancestors 'none'",
--- a/web/src/components/diff/DiffFileViewer.tsx
+++ b/web/src/components/diff/DiffFileViewer.tsx
@@ -75,7 +75,6 @@ function HunkView({
   hunk,
   hunkIndex,
   lineTokens,
-  highlightPending,
   rangeStart,
   draft,
   cardsByEndRow,
@@ -90,7 +89,6 @@ function HunkView({
   hunk: RichDiffHunk;
   hunkIndex: number;
   lineTokens?: SyntaxToken[][];
-  highlightPending?: boolean;
   rangeStart: RangeStart | null;
   draft: DraftRange | null;
   cardsByEndRow: Map<number, AnchoredComment[]>;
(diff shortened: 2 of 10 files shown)


njbrake (approved):

Nice fix. Two-layer approach (root cause + defense in depth) is the right shape, and the new Vitest + Playwright coverage closes the hole that let this slip through; the old `opacity-0` gate made `toBeVisible` pass against fully invisible content.

Notes:

- CodeRabbit's `HighlightResult` doc nit is fair; can be folded into a follow-up since `null` and `[]` are handled the same way downstream.
- The 1500ms hard wait in the new Playwright spec is reasonable for proving a negative; swap to polling if it gets flaky.

LGTM, merging once CI is green.

Issue #679, opened 2026-04-15
web: post-merge continuation flow


njbrake:
## Summary
After a successful merge, prompt the user with follow-up actions instead of leaving them on a now-stale session.

## Options to offer
- Update project memory / CLAUDE.md with what was learned
- Start a new branch from the merged state and carry context forward
- Continue with a follow-up task
- Archive the session

## Why
The moment right after a merge is when context is freshest; making the next step frictionless keeps users in flow.

Issue #353, opened 2026-03-02
Guide for making a separate agent identity


njbrake:
As I've adopted heavier use of my agent, now it comments to Github for me and makes PRs etc. But it still shows up as me, it's using my account. This makes it hard to differentiate between when it's me or when it's my bot.

Maybe I should do this https://dev.to/jpoehnelt/agent-identity-for-git-commits-53n1, and have a guide recommending that it be done in AoE

PR #1364, opened 2026-05-21
fix(cockpit): suppress silent-orphan watchdog during Claude SDK async-agent waits

### LINKED ISSUE #1360: Cockpit: silent-orphan watchdog false-fires when agent awaits async sub-agent work, traps session in restart loop

## Problem

Cockpit sessions that drive async sub-agent flows (SuperPowers-style `launch_async_agent`, parallel `Task` fan-out, anything where the assistant fires a background unit of work and then idles waiting for it to complete) get killed mid-turn by the silent-orphan watchdog (#1240). The agent emits a status chunk like `"Waiting for vibe-tree."` or `"T2 batch 1 fired (5). Waiting for all 5 to complete before batch 2."`, then sits idle on purpose. 60 seconds later the watchdog declares the prompt orphaned, sends `session/cancel`, and the supervisor publishes `Stopped { reason: "prompt_orphaned" }` followed by `Stopped { reason: "user_stopped" }`. The worker is reaped and respawned; the user types `c` / `continue`; the agent fires the same async-wait pattern; the watchdog fires again. Loop.

## Evidence from the affected sessions

Two real sessions from my prod DB. Both running `claude-agent-acp` against an in-tree aoe build.

**`1f7e4fd1c9a540ac`** (1 orphan). The turn reached final accounting (`UsageUpdated` with `cost: { amount: 18.228 USD }`) and the agent had already printed the closing PR link, but the prompt was still pending from the daemon's point of view because claude-agent-acp had not resolved `session/prompt`. Watchdog fires immediately after cost-populated `UsageUpdate` thanks to `silent_orphan_fast_grace_secs = 20`:

```
1330|AgentMessageChunk: ". New DiffLine.test.tsx asserts no opacity-0 ..."
1331|AgentMessageChunk: "**PR**: https://github.com/njbrake/agent-of-empires/pull/1355 (draft)"
1332|UsageUpdated  cost: null
1333|UsageUpdated  cost: { amount: 18.228104, currency: "USD" }
1334|Stopped       reason: "prompt_orphaned"
1335|Stopped       reason: "user_stopped"
```

That one looks like the original #1240 case, but the user-visible damage is identical: the session goes into Stopped state with a "Reconnect" banner even though the turn actually finished.

**`9d182958563f42a8`** (3 orphans, restart-loop). This is the painful one. The agent kept emitting "waiting on async work" messages while sub-agents ran in the background:

```
475|AgentMessageChunk: "50 of 51 written. Waiting on vibe-tree."
476|UsageUpdated cost: null
477|Stopped reason: "prompt_orphaned"
478|Stopped reason: "user_stopped"
479|AcpSessionAssigned  (worker respawned, reattached)
481|UserPromptSent: "c"
...
553|AgentMessageChunk: "T2 batch 1 fired (5). Waiting for all 5 to complete before batch 2."
554|UsageUpdated cost: null
555|Stopped reason: "prompt_orphaned"
556|Stopped reason: "user_stopped"
557|AcpSessionAssigned
559|UserPromptSent: "continue"
...
599|AgentMessageChunk: "Re-fired 4 missing. Wait for completions."
600|UsageUpdated cost: null
601|Stopped reason: "prompt_orphaned"
602|Stopped reason: "user_stopped"
```

The pattern is unambiguous. The agent emits a "waiting" sentence, an incremental usage update with `cost: null` follows, then exactly the silent-orphan grace later the daemon kills the turn.

Across the whole DB: 139 total prompts, 5 `prompt_orphaned`, 128 `prompt_complete`. All 5 orphans land on the async-wait shape; the same human-typed prompt restarted on a fresh worker completes cleanly when the agent does not need to wait.

## Root cause

The watchdog at `src/cockpit/acp_client.rs:2984-3047` arms when `tool_calls_in_flight` is empty and at least one `Progress` event has been seen, then fires after `silent_orphan_grace_secs` (default 60s, accelerated to `silent_orphan_fast_grace_secs = 20s` after cost-populated `UsageUpdate`).

The lifecycle classifier at `src/cockpit/acp_client.rs:311-335` decides what counts as progress:

```rust
fn classify_lifecycle_signal(update: &SessionUpdate) -> Option<LifecycleSignal> {
    match update {
        SessionUpdate::UsageUpdate(u) if u.cost.is_some() => Some(LifecycleSignal::TerminalUsage),
        SessionUpdate::AgentMessageChunk(_)
        | SessionUpdate::AgentThoughtChunk(_)
        | SessionUpdate::Plan(_) => Some(LifecycleSignal::Progress),
        SessionUpdate::ToolCall(tc) => Some(LifecycleSignal::ToolStarted(tc.tool_call_id.0.to_string())),
        SessionUpdate::ToolCallUpdate(update) => { ... }
        _ => None,
    }
}
```

When the agent uses `launch_async_agent` (and the SDK-side `Async agent launched successfully.` confirmation), the lifecycle from the daemon's perspective is:

1. `ToolCall` for `launch_async_agent` arrives, added to `tool_calls_in_flight`.
2. `ToolCallUpdate { status: Completed }` arrives within seconds; removed from the set.
3. The agent emits a final "Waiting for X" `AgentMessageChunk` to the user.
4. `UsageUpdate { cost: null }` arrives. Per `classify_lifecycle_signal`, this is `None`: not Progress, not TerminalUsage. So `last_progress_at` is not refreshed.
5. The agent then polls the async work via mechanisms the adapter does not surface as `ToolCall` (it uses an internal subagent SDK channel, not an ACP-visible tool). `tool_calls_in_flight` is empty. No further `Progress`/`ToolStarted` arrives.
6. Grace expires. Watchdog cancels. `prompt_orphaned`.

The supervisor's drain path (`src/cockpit/supervisor.rs:1149` and follow-on `reap_user_stopped` at `:1772`) then publishes `Stopped { reason: "user_stopped" }`, reaps the worker, and the reconciler respawns it (`src/server/cockpit_reconciler.rs:101`). On the next prompt the same shape recurs.

In short: the watchdog cannot distinguish "adapter wedged" from "agent intentionally awaiting async work". Both look like "no lifecycle signal for grace_secs while no ACP-visible tool is open".

### A. Short-term knob bump

Raise the default `silent_orphan_grace_secs` from 60 to a value that survives a typical async-agent wait. Empirically the SuperPowers `launch_async_agent` flow can stall 2 to 5 minutes between status chunks. A default of `300s` gives real wedges a finite TTL while not killing legitimate waits. Users who do not run async-agent flows can lower it via the settings TUI.

Touch: `src/session/config.rs:304` (`default_silent_orphan_grace_secs`). Update the comment block at `src/session/config.rs:275-282` and the matching prose in `docs/cockpit.md`.

This is a one-line ship; it does not fix the underlying false-positive but stops the restart-loop bleeding while B lands.

### B. Treat "waiting on async sub-agent" as in-flight tool work

Bridge the async-agent lifecycle into `tool_calls_in_flight` so the watchdog stays suppressed for the duration of the wait. Two viable shapes:

1. **Detect the tool name in `classify_lifecycle_signal`.** When `ToolCall.tool_name` is `launch_async_agent` (or future equivalents), do not complete the `ToolStarted` on `ToolCallUpdate { Completed }`. Instead, hold the synthetic id until a follow-up signal indicates the async work landed. claude-agent-acp does emit `RawAgentUpdate` payloads when async sub-agents complete (see `_meta.claudeCode.subagentResult` shape, observable in event 470 of `9d182958563f42a8`); the classifier could parse those into a synthetic `ToolCompleted`.

   Touch: `src/cockpit/acp_client.rs:311-335` (extend classifier), `src/cockpit/acp_client.rs:2827-2843` (no changes to the loop; classification already feeds it).

2. **Heuristic on `AgentMessageChunk` text.** Detect the "Waiting for", "Waiting on", or trailing "Wait for completions" markers and arm a synthetic in-flight token that holds until the next `AgentMessageChunk` or `ToolCall` arrives. Quicker to land but fragile to copy changes in the agent. Prefer 1.

Both shapes live in `acp_client.rs` and need a corresponding test under `tests/cockpit_silent_orphan.rs` mirroring the existing async-wait recording, plus a Vitest snapshot in `web/src/lib/cockpitTypes.test.ts` for the reducer (no UI change expected since `prompt_orphaned` simply stops being emitted).

### C. Make the loop break cheaper

Independently of A/B, the restart-loop itself is painful because each respawn costs the user a worker boot, a `session/load`, an `AcpSessionAssigned`, and a re-typed `c`. When the supervisor reaps a worker stopped via `prompt_orphaned`, the reconciler at `src/server/cockpit_reconciler.rs:101` could check the last two stop reasons; if both are `prompt_orphaned` within the last few minutes, surface a "Async-wait timeout detected. Increase `silent_orphan_grace_secs` in Settings, or disable the watchdog for this profile" actionable banner in the cockpit instead of silently respawning. This is a UX guardrail, not a fix.

Touch: `src/server/cockpit_reconciler.rs:101`, `web/src/components/cockpit/WorkerRestartingBanner.tsx` (add a variant).

## Why now

- The restart-loop traps real work. I lost a `/aoe-implement` mid-flight on `9d182958563f42a8` three times in a row before realising the agent was being killed for being polite about saying "waiting".
- The fix is local. A is a constant change; B is contained to `classify_lifecycle_signal` and a test file.
- async-agent / parallel-Task flows are becoming load-bearing in skills like `/aoe-implement` and `/implement`; the bug will get more common, not less.

## Open questions

- Is `300s` the right A default, or should the default be `0` (disabled) until B lands? Disabling default-on watchdogs would surface real wedges with no fallback, but the false-positive rate currently outpaces the true-positive rate for my workload (5 of 5 orphans were false-positives in the inspected DB).
- Does claude-agent-acp emit a stable, parseable signal when an async sub-agent completes, or do we need to lean on heuristic text matching for B?
- Should B's bridge run server-side in `acp_client.rs` or be pushed upstream to `claude-agent-acp` as a first-class `SessionUpdate` variant? Upstream is the right home long-term but the aoe-side bridge unblocks today.

## Out of scope

- Stale `cockpit-workers/*.log` cleanup (92 orphaned log files in my data dir); orthogonal hygiene, file separately if worth it.
- TUI cockpit parity for the new banner in C; ship web-only first.
- Reworking `silent_orphan_fast_grace_secs` semantics; the accelerator is fine for the original adapter-wedge case, the bug is only that `cost: null` updates do not reset the timer and async waits are not counted as tool work.

## Cross-references

- #1240 — original silent-orphan watchdog issue and its landed implementation. This is a follow-up addressing the false-positive shape.

Seluj78:

## Description

Stops the cockpit silent-orphan watchdog from killing prompts that are legitimately idle while a Claude SDK async sub-agent runs in the background (SuperPowers-style `Agent` tool with `isAsync: true`). The agent dispatches a launch, the visible launch tool completes in ~50ms, then sub-agents poll INSIDE the claude binary with zero further ACP frames. The watchdog interpreted that silence as an adapter wedge and cancelled the turn, trapping the session in a restart loop (`prompt_orphaned` -> `user_stopped` -> `AcpSessionAssigned` -> user types `c` -> repeat).

Two pieces:

1. **Default + floor bump.** `default_silent_orphan_grace_secs` 60 -> 120 (`src/session/config.rs:304`), matching constant in `src/cockpit/acp_client.rs:256`, production-config floor clamp 10 -> 120 in `silent_orphan_grace()` so a typo cannot drop below the new baseline. Debug env override (`AOE_SILENT_ORPHAN_GRACE_MS`) stays unclamped so tests keep their sub-second cadence.

2. **Async-agent detection.** `LifecycleSignal::ToolCompleted` becomes a struct variant `{ id, is_async }`. `classify_lifecycle_signal` detects the Claude SDK's `"Async agent launched successfully"` marker on completion content. The prompt loop sets a sticky `async_agent_running` flag for the rest of the turn; while set, the watchdog promotes its `effective_grace` to at least 30 minutes via the new `ASYNC_AGENT_GRACE_FLOOR` constant, and the async branch takes precedence over the cost-seen fast path (a cost-populated UsageUpdate mid-wait could be intermediate billing telemetry rather than turn termination). The grace stays finite so a real adapter wedge during an async wait still recovers, just slower.

The detection is a bandaid. The proper fix lives in claude-agent-acp itself: the upstream adapter currently drops the SDK's `task_notification` / `task_started` messages on the floor (see upstream agentclientprotocol/claude-agent-acp#336, #603, #679; in-flight PR #353). When those land and forward as proper ACP `SessionUpdate` notifications, the daemon will see `Progress` signals during the wait and this detection layer becomes redundant and can be removed.

Closes #1360

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [x] Start `aoe serve` and open a cockpit session running claude-agent-acp.
- [x] Send a prompt that triggers an async-agent fan-out, e.g. ask the agent to "launch 5 async Task agents researching different topics and wait for them all to finish".
- [x] Watch the cockpit transcript. The agent should emit a "Waiting on X" / "Waiting for ..." message, then sit idle waiting for sub-agents. Before this fix, the session got `prompt_orphaned` within 60s. After this fix, the session should stay live for the full async wait (up to 30 min before the watchdog still kicks in as the safety floor).
- [x] In the same session, send another prompt that does normal tool calls (no async-agent). The watchdog default is now 120s instead of 60s; the watchdog should still fire on a real adapter wedge after 120s of silence with no in-flight tool.
- [x] Open Settings (TUI or web Settings tab, Cockpit category). "Silent-orphan grace" should show 120 (or 120 if you explicitly had it set lower; nonzero values below 120 now clamp up to 120).
- [x] Run `cargo test --features serve --test cockpit_silent_orphan`. All four tests (positive, negative, disabled, and the new `silent_orphan_suppressed_during_async_agent_wait`) should pass.

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude (via Claude Code, agent-of-empires `/aoe-implement` skill, with `/debate` consultation between gemini-3.1-pro-preview and gpt-5.5 on the design tradeoff)

Investigation, plan, and implementation drafted by Claude under my direction. The design choice (text-detect on `ToolCallUpdate` content, struct-variant `LifecycleSignal::ToolCompleted`, `.max(ASYNC_AGENT_GRACE_FLOOR)` against the configured base grace, async branch overrides cost-seen) is the synthesized verdict from a multi-model debate; both models agreed on B1 (finite extended grace, not permanent suppression) and split on the smaller details which I resolved. I reviewed each commit, ran the manual test steps, and made the calls on the open questions in the issue (300s vs 0 default, text vs structured-meta detection, ceiling math).

## What Changed

**Configuration Defaults Updated:**
- The silent-orphan watchdog's default grace period increased from 60 seconds to 120 seconds across the application configuration
- The minimum grace period floor increased from 10 seconds to 120 seconds in production environments (debug environments retain flexibility via environment variable override)

**Async Agent Detection and Handling:**
- Added detection logic that identifies when Claude SDK async sub-agents are running based on specific completion messages
- When an async agent is detected, the watchdog automatically extends its grace period to at least 30 minutes, preventing false timeouts while the agent completes its work
- This extended grace persists for the entire prompt turn and is reset when the prompt exits

**Code Quality and Testing:**
- Added a new integration test that verifies the watchdog correctly suppresses timeout behavior during async agent operations
- Updated the test shim to simulate and validate async agent scenarios
- Updated all user-facing documentation (configuration guide and UI settings) to reflect new defaults and behavior

**Files Modified:**
- Configuration (`src/session/config.rs`, `src/cockpit/acp_client.rs`) - defaults and grace logic
- Documentation (`docs/cockpit.md`, `src/tui/settings/fields.rs`) - user-facing explanations
- Testing (`cockpit-worker/test-shim/shim.mjs`, `tests/integration/cockpit_silent_orphan.rs`) - test coverage and simulation
- Core watchdog logic (`src/cockpit/acp_client.rs`) - async detection and grace window enforcement

## Benefits

- **Prevents false timeout failures**: Eliminates spurious watchdog cancellations when legitimate SDK sub-agents are running in the background
- **Backward compatible**: Longer default grace period helps all use cases, not just async agents
- **Temporary safeguard**: Provides a working solution while awaiting upstream SDK improvements for proper progress visibility
- **Observable and testable**: Includes comprehensive tests to ensure the async detection and grace extension work correctly

## Technical Details

The implementation works by:
- Scanning tool completion messages for the Claude SDK marker "Async agent launched successfully"
- Converting `LifecycleSignal::ToolCompleted` from a simple signal to a struct variant carrying an `is_async` flag
- Maintaining a sticky `async_agent_running` flag at the prompt level that is set when async completion is detected
- While this flag is active, enforcing an effective minimum grace period of 30 minutes (the `ASYNC_AGENT_GRACE_FLOOR`)
- Taking precedence over the faster cost-based grace timeout path when async agent is running
- Special channel handling to ensure async completion signals are not dropped under backpressure

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1364?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


njbrake (approved):

LGTM. The bandaid is well-scoped: detection narrows to `Completed` (not `Failed`), the load-bearing async signal is lossless (awaited send on backpressure), and the watchdog still recovers on a real wedge during an async wait, just slower. Bumping the base to 120 also gives non-async flows breathing room.

A couple of follow-ups worth folding in later (none blocking):

- Apply paths in `src/tui/settings/fields.rs:2452` and `:2939` still clamp `.max(10)` while `silent_orphan_grace()` enforces `.max(120)` at runtime. A user who types `60` sees `60` persisted but the watchdog runs at 120. Bump the apply clamps to 120 to remove the seam.
- CodeRabbit's drop-guard suggestion for env restore is fair. The existing `silent_orphan_disabled_by_zero_grace` test does not restore at all, so this PR is already an improvement; worth a sweep PR to add a drop guard to both.
- 30 minutes is a long dormancy on a real wedge; revisit when `claude-agent-acp#336` lands and this detection layer can go.

Merging once CI is green.

PR #1375, opened 2026-05-21
fix(web): gate session route on first sessions fetch

### LINKED ISSUE #1351: Web dashboard: refreshing /session/<id> falls back to homepage during initial sessions fetch, no loading state distinguished from "not found"

## Problem

Hitting browser refresh on `/session/<id>` shows the dashboard "home" screen instead of the session I was just looking at. The session is healthy, the URL is correct, and a fresh tab opened to the same URL loads the session fine. Only the refresh path on an already-loaded tab regresses.

## What's there today

The app uses path routing through `react-router-dom`'s `useMatch` rather than a `<Routes>` table. `useMatch("/session/:sessionId")` resolves the id off the URL immediately on mount, even before any data has loaded:

`web/src/App.tsx:183`
```tsx
const sessionMatch = useMatch("/session/:sessionId");
...
const activeSessionId = sessionMatch?.params.sessionId ?? null;
```

`activeWorkspace` and `activeSession` then look the id up inside the live `workspaces` list, which is derived from `useSessions().sessions`:

`web/src/App.tsx:252`
```tsx
const activeWorkspace = useMemo(() => {
  if (!activeSessionId) return undefined;
  return workspaces.find((w) =>
    w.sessions.some((s) => s.id === activeSessionId),
  );
}, [workspaces, activeSessionId]);
const activeSession = activeWorkspace?.sessions.find(
  (s) => s.id === activeSessionId,
);
```

`useSessions` starts with an empty array and fetches asynchronously:

`web/src/hooks/useSessions.ts:15`
```ts
const [sessions, setSessions] = useState<SessionResponse[]>([]);
...
useEffect(() => {
  void fetchSessions().then(applyResult);
  intervalRef.current = setInterval(
    () => void fetchSessions().then(applyResult),
    POLL_INTERVAL,
  );
  ...
}, [applyResult]);
```

The render branch then short-circuits to `<Dashboard>` whenever the lookup misses:

`web/src/App.tsx:649`
```tsx
if (!activeWorkspace || !activeSession) {
  return (
    <Dashboard
      sessions={sessions}
      onSelectSession={handleSelectSession}
      ...
    />
  );
}
```

There is no `loaded` flag, no `null` sentinel, and no loading splash. "Sessions list is still loading" and "session doesn't exist" share the same render branch. The existing routing test in `web/tests/routing.spec.ts:34` is explicit about this collapse:

```ts
test("'/session/<id>' for an unknown session falls back to dashboard", async ({ page }) => {
  // No backend, sessions list is empty, so the route still matches but
  // the resolver finds no session and the dashboard renders. Importantly
  // the URL stays put so a real backend can later resolve it.
  await page.goto("/session/does-not-exist");
  ...
});
```

That comment ("URL stays put so a real backend can later resolve it") is exactly the loading-state path being reused as the not-found path, with no observable distinction.

The matching positive test for refresh on a real session is missing. `routing.spec.ts:26` exercises refresh on `/settings` only:

```ts
test("refresh on /settings keeps user on settings", async ({ page }) => {
  await page.goto("/settings");
  ...
  await page.reload();
  await expect(page.getByText("Settings", { exact: true }).first()).toBeVisible();
  await expect(page).toHaveURL("/settings");
});
```

`/settings` is unaffected because the `settingsRootMatch` / `settingsTabMatch` branches at `App.tsx:184-188` render before the session lookup gate.

## Root cause

Two compounding things at `App.tsx:649`:

1. **Empty `sessions` state is treated as "no such session" rather than "still loading."** First paint after a refresh always has `sessions === []`, so `activeWorkspace` is always undefined for one tick, so the Dashboard fallback always renders. The cockpit / terminal view only appears once `fetchSessions()` resolves and the workspace lookup succeeds. On a slow connection, mobile cold start, or a transient backend hiccup that returns `null` (which `applyResult` at `useSessions.ts:32` sets `error=true` for but leaves `sessions` at `[]`), the dashboard sticks.

2. **`useSessions` exposes no loading or initial-load signal.** `sessions` and `error` are the only outputs, both of which conflate "I haven't fetched yet" with "the server confirmed nothing is there / errored." The render gate has no way to wait for the first successful fetch before deciding the lookup truly failed.

The URL is preserved correctly (browser address bar still shows `/session/<id>`); the regression is purely the rendered tree.

## Proposal

**(A) Minimum viable fix: gate the Dashboard fallback on a `loaded` flag from `useSessions`.**

In `web/src/hooks/useSessions.ts`, add an initial-load sentinel:

```ts
const [loaded, setLoaded] = useState(false);
...
const applyResult = useCallback((data: SessionsEnvelope | null) => {
  if (data !== null) {
    setSessions(data.sessions);
    ...
    setLoaded(true);
  } else {
    setError(true);
    setServerDown(true);
    setLoaded(true);
  }
}, []);
...
return { sessions, ..., loaded };
```

In `App.tsx:649`, distinguish the two states:

```tsx
if (!sessionsLoaded && activeSessionId) {
  return <SessionLoadingPlaceholder />;
}
if (!activeWorkspace || !activeSession) {
  return <Dashboard ... />;
}
```

The placeholder can be as minimal as the existing `<div className="h-dvh bg-surface-900 safe-area-inset" />` already used for the pre-auth shell at `App.tsx:152`, or a small "Loading session" pane. Cockpit users already see a `<CockpitLoadingFallback />` for the lazy chunk at `App.tsx:903`; the same visual vocabulary applies.

**(B) Structural fix: route table with a dedicated session route, not a useMatch chain.**

Hoist a real `<Routes>` to the App shell so the session loader can suspend on data, not on derived render checks. This also opens the door to a `<Navigate to="/" replace>` once the not-found case is confirmed (instead of the silent fallthrough that hides "URL is wrong" from the user). Bigger refactor, blocked behind (A) which is the fast surface fix.

**(C) Test coverage gap.**

Add a positive refresh case to `web/tests/routing.spec.ts` alongside the existing `/settings` refresh test:

```ts
test("refresh on /session/<id> keeps the user on that session", async ({ page }) => {
  // Mock sessions endpoint to include the session id.
  ...
  await page.goto("/session/known-session");
  await expect(<cockpit/terminal indicator>).toBeVisible();
  await page.reload();
  await expect(<cockpit/terminal indicator>).toBeVisible();
  await expect(page).toHaveURL("/session/known-session");
});
```

The unknown-session test at `routing.spec.ts:34` should also tighten its assertion: today it asserts the dashboard renders, which is the bug. A correct shape would assert either a loading state during fetch, or a "Session not found" affordance after fetch resolves with no match.

## Why now

- The flash is one of the first impressions every refresh gives. PWA installs and bookmarked session URLs are the main re-entry points; both refresh-equivalent on mobile when the OS evicts the tab.
- The fix is small (one boolean), localized to two files, and unblocks a clearer not-found UX.
- Coverage matrix mandate covers "sessions / sidebar" flows. The current routing tests miss the positive refresh case for a known session, so this issue lands a small test alongside the fix.

## Open questions

- Should the not-found state (URL has `/session/<id>`, fetch returned a list without that id) render the Dashboard silently, or surface a "Session not found" message with a "Go to dashboard" button? Today the silent fall-through hides the case where a session id genuinely no longer exists.
- For the loading state: minimal blank pane (matches `App.tsx:152` pre-auth) or a session-themed placeholder? Cockpit and tmux views look different enough that a single generic spinner may flash twice (loading → tmux/cockpit chunk loading). Reusing the existing `<CockpitLoadingFallback />` vocabulary is one option.
- Does the live Playwright harness need adjustment to reliably exercise refresh-on-known-session, or does the mocked routing spec suffice (the `useSessions` polling and stateCache live behind the mocked `/api/sessions` response)?

## Out of scope

- Replacing the `useMatch` calls with a `<Routes>` table (proposal B). Useful but larger; the immediate UX bug deserves a tight fix first.
- `useCockpit`'s in-memory `stateCache` and its interaction with refresh. The cache reads correctly on remount; the bug here is upstream of any cockpit rendering.
- The not-found UX shape itself (a dedicated "Session not found" screen). Tracked as an open question, can ship as a follow-up.

## Cross-references

- `web/tests/routing.spec.ts:34` documents the current "fall back to dashboard" behavior with a passing test, which is part of why this regression escaped CI.
- `web/tests/routing.spec.ts:26` covers refresh-on-`/settings` correctly; the analogous test for `/session/<id>` is the gap.

Seluj78:

## Description

Refreshing the browser on `/session/<id>` flashed the dashboard fallback for one render tick before the session view reappeared, because the empty `sessions` state during the initial fetch was indistinguishable from "no such session." On a slow connection, mobile cold start, or any transient backend hiccup that returned `null`, the dashboard stuck and the user appeared to lose their place.

Root cause was in `web/src/App.tsx:649`: `if (!activeWorkspace || !activeSession) return <Dashboard />` ran before `useSessions` had a chance to resolve its first fetch. `useSessions` only exposed `sessions` and `error`, both of which conflated "still loading" with "the server confirmed nothing is there."

Fix adds a `loaded` sentinel to `useSessions`, flipped to `true` once the first `fetchSessions` resolves (success or null). `App.tsx` reads it as `sessionsLoaded` and, when a `/session/<id>` route is matched but the first fetch has not landed, renders the minimal pre-auth shell instead of the dashboard. Once `sessionsLoaded` flips the existing fallback decides between cockpit/terminal view and dashboard, preserving the existing not-found UX.

Test coverage: new mocked-playwright case in `web/tests/routing.spec.ts` stubs `/api/sessions` with a known session, navigates to `/session/<id>`, reloads, asserts the dashboard fallback never shows. New `app-shell.routing` surface in `web/tests/coverage-matrix.json` so routing is an owned surface.

Closes #1351

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open the web dashboard, create or pick a session, navigate to `/session/<id>`, hit browser refresh. Confirm the session view (cockpit or terminal) reappears without a flash to the dashboard. URL stays at `/session/<id>`.
- [ ] In Chrome DevTools, throttle to Slow 3G, then refresh on `/session/<id>`. Confirm the screen shows a blank dark shell (matches the pre-auth shell) for the duration of the initial fetch, then transitions to the session view.
- [ ] Navigate to `/session/does-not-exist`, confirm the dashboard renders once the first fetch resolves and the URL stays put.
- [ ] Refresh on `/settings`, confirm settings still renders (unchanged behavior).
- [ ] Refresh on `/` (dashboard root), confirm dashboard still renders without a flash.
- [ ] Stop `aoe serve` mid-session, refresh on `/session/<id>`, confirm the dashboard renders once the fetch fails (offline path) and the URL stays put.

## Checklist

## AI Usage

- [ ] No AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (placeholder shape, scope of test coverage, where to add the matrix entry), and ran the validation locally.

## Summary

This PR fixes a UI regression where reloading a session page (e.g., /session/<id>) briefly showed the dashboard instead of the session view while session data was still loading.

### What changed (high-level)
- Introduced an initial-load sentinel so the app can tell "still loading" apart from "no sessions found".
- Gate the dashboard fallback: when a session route is active but sessions haven’t finished loading, the app renders a minimal pre-auth/loading shell instead of the dashboard; once the first load completes the app resumes normal routing (session view or dashboard).
- Added tests and coverage entries to lock in the behavior: a Playwright routing test for the refresh flow, a Vitest for the hook’s loaded sentinel, and coverage-matrix updates.

### Benefits
- Eliminates the transient dashboard flash on session-page refresh.
- Provides a clearer loading UX (shows a loading/pre-auth shell instead of navigating to dashboard).
- Adds tests to prevent regressions related to routing and session-load timing.

### Technical details
- useSessions hook: adds a boolean loaded flag that starts false and flips true once the initial fetchSessions promise resolves (success or null); the hook now returns loaded.
- App.tsx: reads sessionsLoaded from useSessions; when a /session/<id> route is matched but sessionsLoaded is false, App renders a minimal pre-auth shell early. After sessionsLoaded becomes true, existing logic selects the session cockpit/terminal view or the dashboard fallback (preserving not-found behavior).
- Tests: Playwright test stubs /api/sessions and related endpoints to verify reloading /session/<id> never shows the dashboard fallback; Vitest covers the loaded sentinel flipping true on both successful and null initial responses; coverage matrix updated with app-shell routing surface.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1375?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -196,6 +196,7 @@ function AppContent({ loginRequired, onLogout }: { loginRequired: boolean; onLog
     setWorkspaceOrdering,
     markLocalOrderingUpdate,
     error,
+    loaded: sessionsLoaded,
     injectSession,
     setSessionStatus,
   } = useSessions();
@@ -645,6 +646,16 @@ function AppContent({ loginRequired, onLogout }: { loginRequired: boolean; onLog
           readOnly={serverAbout?.read_only}
         />
       );
+    }
+
+    // Refresh on `/session/<id>` paints once with `sessions === []` before
+    // the first poll resolves. Without this guard the lookup misses, the
+    // dashboard fallback renders, and the cockpit/terminal view only
+    // reappears once the fetch lands. Hold the minimal pre-auth shell
+    // until the first fetch settles, then let the real fallback decide.
+    // See #1351.
+    if (activeSessionId && !sessionsLoaded) {
+      return <div className="h-dvh bg-surface-900 safe-area-inset" />;
     }
 
     if (!activeWorkspace || !activeSession) {
--- a/web/src/hooks/useSessions.test.ts
+++ b/web/src/hooks/useSessions.test.ts
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+//
+// Covers the `loaded` sentinel added for #1351. Without `loaded` the
+// session-route gate in App.tsx cannot tell "first fetch still in
+// flight" apart from "server confirmed there is no such session,"
+// which is why refresh on /session/<id> used to flash the dashboard.
+
+import { renderHook, waitFor } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { useSessions } from "./useSessions";
+import * as api from "../lib/api";
+
+describe("useSessions / loaded sentinel", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("starts loaded=false before the first fetch resolves", () => {
+    let resolveFetch: (value: api.SessionsEnvelope | null) => void = () => {};
+    vi.spyOn(api, "fetchSessions").mockImplementation(
+      () => new Promise((r) => (resolveFetch = r)),
+    );
+
+    const { result } = renderHook(() => useSessions());
+
+    expect(result.current.loaded).toBe(false);
+    expect(result.current.sessions).toEqual([]);
+    // Resolve the pending promise so the polling effect's cleanup
+    // does not leak past this test.
+    resolveFetch(null);
+  });
+
+  it("flips loaded=true after the first successful fetch", async () => {
+    vi.spyOn(api, "fetchSessions").mockResolvedValue({
+      sessions: [],
+      workspace_ordering: [],
+    });
+
+    const { result } = renderHook(() => useSessions());
+
+    await waitFor(() => expect(result.current.loaded).toBe(true));
+    expect(result.current.error).toBe(false);
+  });
+
+  it("flips loaded=true even when the first fetch returns null", async () => {
+    vi.spyOn(api, "fetchSessions").mockResolvedValue(null);
+
+    const { result } = renderHook(() => useSessions());
+
+    await waitFor(() => expect(result.current.loaded).toBe(true));
+    // The null branch also flags error/serverDown surfaces; the
+    // important contract for #1351 is that `loaded` does not stay
+    // stuck at false, so the App.tsx gate can release the placeholder.
+    expect(result.current.error).toBe(true);
+  });
+});
--- a/web/src/hooks/useSessions.ts
+++ b/web/src/hooks/useSessions.ts
@@ -15,6 +15,12 @@ export function useSessions() {
   const [sessions, setSessions] = useState<SessionResponse[]>([]);
   const [workspaceOrdering, setWorkspaceOrdering] = useState<string[]>([]);
   const [error, setError] = useState(false);
+  // True once the first fetch attempt resolves (success or failure). Lets
+  // callers distinguish "still loading" from "list confirmed empty / no
+  // such session," which matters for refresh on `/session/<id>` where an
+  // empty `sessions` array during initial paint otherwise indistinguishably
+  // collapses to the dashboard fallback. See #1351.
+  const [loaded, setLoaded] = useState(false);
   const intervalRef = useRef<ReturnType<typeof setInterval> | null>(null);
   const lastLocalOrderingAtRef = useRef<number>(0);
 
@@ -45,6 +51,7 @@ export function useSessions() {
       setError(true);
       setServerDown(true);
     }
+    setLoaded(true);
   }, []);
 
   const refresh = useCallback(async () => {
@@ -72,6 +79,7 @@ export function useSessions() {
     setWorkspaceOrdering,
     markLocalOrderingUpdate,
     error,
+    loaded,
     refresh,
     injectSession,
     setSessionStatus,
--- a/web/tests/coverage-matrix.json
+++ b/web/tests/coverage-matrix.json
@@ -125,6 +125,13 @@
       "components": ["web/src/components/TopBar.tsx"]
     },
     {
+      "id": "app-shell.routing",
+      "kind": "mocked-playwright",
+      "risk": "high",
+      "specs": ["tests/routing.spec.ts"],
+      "components": []
+    },
+    {
       "id": "profiles.lifecycle",
       "kind": "live-playwright",
       "risk": "high",
--- a/web/tests/coverage-matrix.json
+++ b/web/tests/coverage-matrix.json
@@ -132,6 +132,13 @@
       "components": []
     },
     {
+      "id": "app-shell.sessions-loaded-sentinel",
+      "kind": "vitest",
+      "risk": "high",
+      "specs": ["src/hooks/useSessions.test.ts"],
+      "components": []
+    },
+    {
       "id": "profiles.lifecycle",
       "kind": "live-playwright",
       "risk": "high",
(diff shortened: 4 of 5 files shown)


njbrake (approved):

Tight fix. Single `loaded` sentinel plus a scoped render-time gate, no on-disk change, and the Playwright case proves the dashboard flash is gone.

Two minor notes (none blocking):

- The mocked Playwright case uses a 1s `setTimeout` for the route delay. Could get flaky on slow runners; the actual user-flow assertion in the second case does not depend on the timing, so this is more cosmetic than real.
- `sessionsLoaded` could be named `firstFetchSettled` to make the never-flips-back semantic explicit, but the current name reads fine in context.

CI is fully green. Merging.

PR #1378, opened 2026-05-21
fix(cockpit/web): standalone /clear in combined-mode drain (#1356)

### LINKED ISSUE #1356: Cockpit: queued /clear gets glued into combined drain POST, losing standalone-clear semantic (#1031 follow-up)

## Problem

Queueing `/clear` while a turn is active does not behave like a deferred clear. The drain effect's default `combined` mode joins every queued entry with `\n\n` and POSTs a single multi-paragraph prompt. So either:

1. `/clear` is the head of the queue. The combined POST starts with `/clear\n\n<other text>`. Server-side detection at `is_clear_command` matches and emits `SessionCleared`, but the agent receives `/clear` glued to the follow-up text in one prompt; the SDK does not see a clean clear command followed by a separate user message, so the post-clear continuation is racy / behavior-undefined.
2. `/clear` is not the head. The combined POST looks like `<earlier text>\n\n/clear\n\n<later text>`. `is_clear_command` only matches when the trimmed text starts with `/clear`, so no `SessionCleared` is emitted at all; the agent's claude SDK may or may not honour `/clear` mid-prompt; the UI sees no boundary divider, no usage reset, no capability clear.

Either branch is wrong relative to the user's intent ("clear, then continue with these follow-ups").

## What's there today

Server-side `/clear` detection at `src/cockpit/agent_profiles.rs:46`:

```rust
pub fn is_clear_command(&self, text: &str) -> bool {
    let trimmed = text.trim();
    for alias in self.clear_aliases {
        if trimmed == *alias {
            return true;
        }
        if let Some(rest) = trimmed.strip_prefix(*alias) {
            if rest.starts_with(char::is_whitespace) {
                return true;
            }
        }
    }
    false
}
```

Detection is anchored at the start of the trimmed prompt. Anything that puts `/clear` past the first paragraph misses.

Frontend queueing at `web/src/hooks/useCockpit.ts:910`:

```ts
const sendPrompt = useCallback(
  async (text: string) => {
    if (!sessionId) return;
    ...
    if (state.turnActive) {
      dispatch({ kind: "enqueue_prompt", text });
      return;
    }
    ...
    await dispatchPromptNow(text);
  },
  [sessionId, state.turnActive, dispatchPromptNow],
);
```

No special-case for `/clear`. The clear command is enqueued like any other prompt and waits for the drain effect.

Combined drain at `web/src/hooks/useCockpit.ts:974`:

```ts
if (drainModeRef.current === "combined") {
  const snapshot = state.queuedPrompts.slice();
  const combined = combineQueuedPrompts(snapshot);
  const sentIds = snapshot.map((q) => q.id);
  void dispatchPromptNow(combined)
    .then((ok) => {
      if (ok) dispatch({ kind: "dequeue_prompts_by_id", ids: sentIds });
    })
    .finally(() => {
      drainingRef.current = false;
    });
}
```

`combineQueuedPrompts` (`web/src/lib/cockpitTypes.ts:256`):

```ts
export function combineQueuedPrompts(
  queue: ReadonlyArray<QueuedPrompt>,
): string {
  return queue.map((q) => q.text).join("\n\n");
}
```

No boundary handling. `/clear` becomes a substring of the joined text, and from the server's perspective the combined POST is one prompt with whatever leading non-`/clear` content the user happened to enqueue first.

## Root cause

Two layers, both at play depending on queue order.

1. **Combined-mode drain ignores prompt semantics.** Any prompt that should be standalone (`/clear`, `/compact`, `/new` for codex / opencode, anything in the agent profile's `clear_aliases`) gets glued into the same POST as adjacent queued prompts. The agent's SDK has no reliable way to recover the original turn boundaries; aoe's text-match detection only sees the joined head.

2. **Server-side detection is head-anchored.** Even if the drain fired a multi-paragraph prompt where the first paragraph is `/clear`, the rest of the joined text follows it as the "same turn." That's not what `/clear` means anywhere else (CLI, web claude.ai). The mental contract for `/clear` is "discard the conversation, the next thing I type is a fresh turn." Joining a follow-up to it breaks that contract.

These are independent: fixing (1) without (2) still handles `/clear` correctly because each prompt fires alone. Fixing (2) without (1) still leaves combined drains corrupting the boundary.

## Proposal

(A) is the minimum fix; (B) tightens the UX further.

### A. Drain effect splits at clear-command boundaries

In `web/src/hooks/useCockpit.ts` (the combined branch around line 974), when the snapshot of `queuedPrompts` is built, scan for entries whose text matches the active agent's clear aliases. Split the snapshot into sub-batches at each clear boundary, then dispatch them in order, awaiting each before the next.

Pseudo-code:

```ts
function isClearAlias(text: string, profile: AgentProfile): boolean {
  const trimmed = text.trim();
  return (profile.clearAliases ?? []).some(
    (alias) => trimmed === alias || trimmed.startsWith(alias + " "),
  );
}

async function drainCombined(queue, profile) {
  let batch: QueuedPrompt[] = [];
  for (const entry of queue) {
    if (isClearAlias(entry.text, profile)) {
      if (batch.length > 0) {
        await dispatchPromptNow(combineQueuedPrompts(batch));
        batch = [];
      }
      await dispatchPromptNow(entry.text);   // standalone, exact match
    } else {
      batch.push(entry);
    }
  }
  if (batch.length > 0) {
    await dispatchPromptNow(combineQueuedPrompts(batch));
  }
}
```

Each sub-batch awaits a `Stopped` event before the next can dispatch (the drain effect already gates on `state.turnActive`, so a recursion-free loop wrapped in the existing `drainingRef` guard is sufficient).

Tests live in `web/src/hooks/useCockpit.queue.test.ts` (existing); add cases for queue orderings like `[a, /clear, b]`, `[/clear]`, `[a, b, /clear]`, `[/clear, /clear, a]`.

### B. Refuse to bundle clear commands with surrounding text in the composer

Even with (A), the user can still type one long multi-line prompt with `/clear` on its own line and submit. That prompt is a single queued entry, so (A) doesn't split it. Two options:

- Composer-side: when the submitted text contains an embedded `/clear` line (any agent's alias) and the text is more than just that line, surface a confirm: "This will clear the conversation, then send the rest as a new turn. Continue?" or split it into two entries at enqueue time.
- Server-side: change `is_clear_command` to detect mid-prompt `/clear` lines and synthesise two events (`SessionCleared`, then a fresh `UserPromptSent` for the trailing text). Bigger change; the cleaner one is composer-side splitting at enqueue.

(B) is a polish layer; (A) covers the common case (user submits `/clear` as its own message while a turn is running).

## Why now

- `/clear` is one of the few commands whose meaning depends on isolation. Bundling it silently with other prompts breaks the user's trust that the command did what they typed. Combined-drain users hit this whenever they queue `/clear` after some other follow-up.
- The fix is localised: ~20 lines in the drain effect, plus tests. No protocol change, no upstream ask.
- Pairs with the parallel cost-doesn't-reset bug (filed): both are about the gap between "user typed `/clear`" and "the system fully honoured `/clear`." Same surface; complementary fixes.

## Open questions

- Should serial drain ALSO honour the boundary, or is the combined-mode fix sufficient? Serial mode already dispatches one entry per turn, so `/clear` arrives alone and is detected. No change needed there beyond a regression test pinning the behaviour.
- For (A), should the user see a visual hint in the queued-prompt strip when `/clear` is in the queue? e.g. a divider between the queued items above and below it, signalling "these will fire in two batches"? Lean yes; the existing strip is otherwise just a flat list and the boundary semantic is invisible.
- Codex `/new`, opencode `/new`: same shape, same handling? They're in `clear_aliases` for those profiles, so (A)'s `isClearAlias` covers them automatically. Verify the per-agent tests.
- What's the right behavior if the agent's SDK actually does support `/clear` mid-prompt and aoe's split changes observable behavior? Lean: aoe's split is the more honest contract. If anyone relies on the existing behaviour they can use serial mode (which already fires one prompt per turn).
- Interaction with #1196 (force_end_turn): if force_end_turn fires while `/clear` is queued, the drain restarts on the next idle. Verify the split applies on every drain attempt, not just the first.

## Out of scope

- Implementing `/clear`'s actual behaviour (claude SDK / claude-agent-acp owns it; aoe only triggers + observes).
- Replacing combined drain with serial as the default. That's a separate UX call (#1031 went combined-by-default for batching agent responses).
- Handling other "isolated" slash commands beyond clear. If the agent profile starts caring about `/compact` or `/login` boundaries, the same `isClearAlias`-style helper extends; defer.

## Cross-references

- #1031 (closed) introduced combined-mode queueing as the default. This issue follows up on a corner the original PR didn't surface.
- #1101 (closed) added the synthetic `SessionCleared` event and head-anchored `is_clear_command`. The head anchor is exactly what fails in combined drain.
- Companion cost issue (just filed) covers the post-`/clear` `UsageUpdated` cumulative-cost regression. Same `/clear` surface, different layer.

Seluj78:

## Description

Cockpit combined-mode drain used to join every queued entry with blank lines and POST one multi-paragraph prompt. When `/clear` was in the queue this glued it into the joined text, and the server's head-anchored `is_clear_command` either misfired (when `/clear` was the head, `<clear>\n\n<follow-up>` reached the agent SDK as one turn) or missed it entirely (when `/clear` was past the first paragraph), so `SessionCleared` never landed, the usage counter never reset, the divider never rendered, and the agent saw a single prompt that begins or ends with the literal `/clear` token instead of a clean reset boundary.

The drain effect now slices the queued snapshot at every clear-command alias (`/clear`, codex / opencode `/new`). Each drain pass fires only the leading sub-batch, either a standalone clear-alias entry or the run of non-clear entries up to the next alias; the remaining queue waits for the next `Stopped`. One POST per turn cycle, same as before, just at the right boundary.

A small `isClearAlias` helper in `web/src/lib/agentProfiles.ts` mirrors `AgentProfile::is_clear_command` from `src/cockpit/agent_profiles.rs`, so detection on both sides matches exactly. The queued-prompt strip now renders an amber `fires separately` divider between rows that will land in different sub-batches, so the user can see the boundary before sending. Profiles with no clear aliases (gemini, vibe, pi, default) keep the original combined-drain behaviour.

Closes #1356

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start `aoe serve --features serve`, open a claude cockpit session, send a prompt to start a turn.
- [ ] While the turn is active, queue three entries in this order: `foo`, `/clear`, `bar`. The queued-prompt strip shows an amber `fires separately` divider above and below the `/clear` row.
- [ ] Let the turn end. The cockpit fires three POSTs in order: `foo`, `/clear`, `bar`. Confirm via DevTools network panel.
- [ ] After the `/clear` POST, a `Conversation cleared` divider appears in the transcript, the usage counter resets, and the previous rows fold behind the `Show N earlier turns` banner. The follow-up `bar` lands as a fresh turn under the divider.
- [ ] With only `/clear` queued and the turn ending, the cockpit fires `/clear` standalone (one POST, body `{ "text": "/clear" }`).
- [ ] Queue `[a, b, /clear]` and let the turn end: first POST is `a\n\nb` (combined), second is `/clear` alone.
- [ ] Repeat the `foo`, `/clear`, `bar` flow against a codex session (`aoe add . --cockpit --agent codex`); same split applies with the codex `/new` alias instead of `/clear`.
- [ ] Switch to a gemini session (no clear aliases) and queue `[a, b, c]`: combined drain still fires one POST `a\n\nb\n\nc`, no divider in the strip.

## Checklist

## AI Usage

- [ ] No AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (split-at-boundary over composer-side rewrite, divider in the strip, no /debate for this localised fix), and ran the manual test steps.

## Overview
This PR fixes how the cockpit frontend handles queued clear commands (e.g., `/clear`, `codex/new`, `opencode/new`) in combined drain mode. Instead of concatenating all queued prompts into a single POST (which could hide clear commands from server detection), the drain now treats clear-alias entries as boundaries and ensures those commands fire separately so session-clear semantics work reliably.

## What changed (high level)
- Drain behavior: combined-mode drain now scans the queued snapshot and splits it at clear-command aliases. Each drain pass dispatches only the leading sub-batch: either a standalone clear-alias entry or the run of non-clear entries up to the next clear. Remaining items stay queued for the next turn.
- Frontend detection: added an exported isClearAlias helper so frontend and server use the same clear-alias matching logic.
- UI: the queued prompts strip renders an amber “fires separately” divider where queued rows will be split; QueuedPromptsStrip is now exported.
- Tests & docs: cockpit docs updated; new tests cover multiple queue orderings and profile-specific behavior; test coverage matrix updated.

## Benefits
- Clear commands queued mid-turn reliably trigger session clears, usage counter resets, divider rendering, and proper agent resets.
- Users see which queued items will send together versus separately.
- Profiles that have no clear aliases keep the original combined behavior.

## Technical notes
- useCockpit (web/src/hooks/useCockpit.ts): combined-mode drain consults profile.clearAliases and slices the queued snapshot into sub-batches at clear aliases. Behavior preserves one POST per sub-batch per turn cycle and advances the in-memory queue only after each `Stopped` event.
- isClearAlias(text, aliases) (web/src/lib/agentProfiles.ts): exported helper that trims input, returns true for exact alias matches or alias + whitespace-separated trailing args, and false for empty input or empty alias lists. Mirrors server-side AgentProfile::is_clear_command.
- CockpitView / QueuedPromptsStrip (web/src/components/cockpit/CockpitView.tsx): inserts a divider row with data-testid="queued-clear-boundary" between queued prompts that will be sent in different sub-batches; component is exported.
- Tests added/updated:
  - web/src/hooks/useCockpit.queue.test.ts: verifies splitting behavior and network POST ordering for sequences like [a, /clear, b], [/clear], [a, b, /clear], [/clear, /clear, a]; includes profile-specific cases (claude, codex, gemini).
  - web/src/lib/agentProfiles.test.ts: covers isClearAlias matching rules.
  - web/src/components/cockpit/QueuedPromptsStrip.test.tsx: verifies divider rendering and related UI behavior.
- Coverage: added cockpit.queue-drain-clear-split surface to web/tests/coverage-matrix.json.

## Areas for further enhancement / polish
- Splitting multi-line submissions or composer-side handling of embedded clear lines (or server-side mid-prompt clear detection) remains out of scope and could be addressed in a follow-up.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1378?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


njbrake (approved):

LGTM. Drain slice at clear-command boundaries is the right shape, the TS `isClearAlias` is a hard mirror of the Rust `is_clear_command`, and the gemini empty-aliases fast path keeps the original single-POST contract for profiles that do not honour `/clear`. Tests cover the head, middle, tail, repeated clears, multi-agent, and the empty-aliases case explicitly.

One CI note: `cockpit_silent_orphan::silent_orphan_fires_on_cost_then_silence` failed on the Ubuntu Cargo Test job, but this PR does not touch any Rust. The branch is missing #1364 and #1374 which both reshape the silent-orphan watchdog (default grace 60s -> 120s, runtime floor 120s, and the SessionContextReset cost-rebase). Merging main in should clear the failure.

Merging once CI is green on a rebased branch.

Issue #1088, opened 2026-05-13
Cockpit: parallelize cold-start session resume + add "Resuming…" UI affordance (currently ~2-3s per session of silent serial spawns)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

`aoe serve` resumes persisted cockpit sessions **sequentially** on cold start. With 12 sessions, that's ~30s of wall time before the dashboard is fully usable. Each session is independent — no reason they have to be one-at-a-time.

The wait is also silent: the dashboard loads quickly enough, but sessions show empty or loading-like state while the daemon is still working through the queue, with no progress affordance or "please wait, X of Y resuming" indicator. So the user has to guess: is the agent stuck, or is it still resuming?

Excerpt from a live cold start (12 sessions, 2 ENOENT failures on missing `claude-agent-acp` per #1048):

```
05:21:42.616  serve starting
05:21:42.631  spawning session=7a6eb81a...   (sequential 1)
05:21:46.144   ^ done (~3.5s: init handshake + session/load)
05:21:46.144  spawning session=d55301f1...   (failed, ENOENT)
05:21:46.145  spawning session=1d339ea5...   (sequential 2)
05:21:48.895   ^ done
05:21:48.895  spawning session=a31faceb...   (sequential 3)
05:21:51.715   ^ done
05:21:51.715  spawning session=65e231c9...   (sequential 4)
05:21:54.554   ^ done
05:21:54.554  spawning session=82a464ec...   (sequential 5)
05:21:58.232   ^ done
05:21:58.232  spawning session=9c772607...   (failed, ENOENT)
05:21:58.233  spawning session=3670438a...   (sequential 6)
05:22:01.051   ^ done
05:22:01.051  spawning session=a21d6642...   (sequential 7)
05:22:03.673   ^ done
05:22:03.673  spawning session=3d59dbd9...   (sequential 8)
05:22:06.223   ^ done
05:22:06.223  spawning session=a6a84671...   (sequential 9)
05:22:09.083   ^ done
05:22:09.083  spawning session=901bf7d0...   (sequential 10)
05:22:11.737   ^ done
05:22:11.737  spawning session=9f23ddfa...   (sequential 11)
05:22:14.570   ^ done (32s total since serve start)
```

[...]

Issue #1031, opened 2026-05-12
Cockpit: allow typing + queueing follow-up prompts while agent is working (Send hidden mid-turn)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

In a cockpit session, while the agent is "thinking" / actively producing a response (or running tool calls), the composer's Send button is **hidden and replaced by a Stop button** — the user has no way to type a follow-up or queue a message until the current turn finishes. Compose another thought? Wait. Want to add "also do X while you're at it"? Wait. The only available action is Cancel.

Every consumer chat UI (Claude.ai, ChatGPT, Cursor, Zed) lets you type and queue a follow-up message while the agent is still working — it just dispatches the moment the current turn ends. The cockpit's strict gate feels noticeably worse than every adjacent product.

## What's there today

The block is purely client-side UX, not a backend or protocol constraint:

- `web/src/components/cockpit/Composer.tsx:227-232` swaps Send for Stop based on `ThreadPrimitive.If running`:
  ```tsx
  <ThreadPrimitive.If running>
    <StopButton />
  </ThreadPrimitive.If>
  <ThreadPrimitive.If running={false}>
    <SendButton disabled={!connected} />
  </ThreadPrimitive.If>
  ```
- `isRunning` is wired to `cockpit.state.turnActive` at `web/src/components/cockpit/CockpitRuntime.tsx:81`.
- `turnActive` flips to `true` on `UserPromptSent` (`web/src/lib/cockpitTypes.ts:520`) and back to `false` on `Stopped` or `AgentStartupError` (`:462`, `:484`).

Backend: there's no rejection of mid-turn prompts.

- `Supervisor::send_prompt` (`src/cockpit/supervisor.rs:675-685`) doesn't check turn state. It waits for the worker to exist, then forwards to `AcpClient::send_prompt`.
- `AcpClient::send_prompt` (`src/cockpit/acp_client.rs:354-360`) just pushes `ClientCmd::Prompt(text)` onto an mpsc channel.
- The connection-loop consumer (`src/cockpit/acp_client.rs:1255-1286`) processes one `ClientCmd` at a time, blocking on `connection.send_request(PromptRequest::new(...)).block_task().await?` until the agent's turn completes. **A second `Prompt` cmd sent during the block sits in the channel queue and is processed serially after `Event::Stopped` fires** (`acp_client.rs:1281-1285`).

[...]

PR #1379, opened 2026-05-21
fix(cockpit): queue and auto-send composer message when session inactive

### LINKED ISSUE #1359: Cockpit: queue + auto-send composer message when session not active (WS closed, worker stopped, worker restarting)

## Problem

When the cockpit session is "not active" (WS disconnected, worker stopped, worker restarting), the composer hard-disables the Send / QueueSend button and `sendPrompt` rejects the call with the banner `Cockpit disconnected; message not sent. Reconnect to retry.` The message the user just typed is not lost from the textarea (a per-session draft persists at `cockpit:draft:<sid>`), but it does not enter the prompt queue, and the user has to remember to click Send again after the banner clears.

The underlying queue + drain machinery is already capable of "park now, fire when session resumes": the drain effect at `web/src/hooks/useCockpit.ts:953` correctly waits for `status === "open" && workerState === "running" && !workerStopped && !workerRestarting` before draining. It is purely the entry path (composer button + `sendPrompt` status gate) that refuses to enqueue while the session is inactive.

Net effect: I type a follow-up while the daemon is restarting / the worker is stopped / my WS dropped, hit Enter, and nothing happens beyond an error banner. Reconnect, click into the composer, hit Send again. That second send is what should have been automatic.

## What's there today

**Composer disables Send / QueueSend when `connected` is false.** `web/src/components/cockpit/CockpitView.tsx:383` passes `connected={status === "open" && !state.workerStopped && !state.workerRestarting}` to `<Composer>`, and both `SendButton` (`web/src/components/cockpit/Composer.tsx:964-993`) and `QueueSendButton` (`web/src/components/cockpit/Composer.tsx:1022-1068`) take `disabled={!connected}`.

`SendButton` tooltip switches to `Disconnected — reconnect to send` (this string is the one place in the file using an em dash; should be fixed in the same PR). The button itself is disabled so the click handler never fires.

**`sendPrompt` short-circuits with an error banner on closed WS.** `web/src/hooks/useCockpit.ts:910-939`:

```ts
const sendPrompt = useCallback(
  async (text: string) => {
    if (!sessionId) return;
    if (statusRef.current !== "open") {
      dispatch({
        kind: "error",
        message: "Cockpit disconnected; message not sent. Reconnect to retry.",
      });
      return;
    }
    if (state.turnActive) {
      dispatch({ kind: "enqueue_prompt", text });
      return;
    }
    if (workerStateRef.current !== "running") {
      dispatch({ kind: "enqueue_prompt", text });
      return;
    }
    await dispatchPromptNow(text);
  },
  [sessionId, state.turnActive, dispatchPromptNow],
);
```

Note the asymmetry: `workerState !== "running"` already enqueues (added for #1088 cold-start parallel resume). `status !== "open"` does not.

`dispatchPromptNow` repeats the same status guard at `web/src/hooks/useCockpit.ts:862-868`. That one is correct in context: it is also called by the drain effect, which should not POST through a half-open WS. But for the user-facing entry point, the desired behavior is "queue, don't error."

**The drain effect already handles the inactive-then-active transition.** `web/src/hooks/useCockpit.ts:953-1010`:

```ts
useEffect(() => {
  if (drainingRef.current) return;
  if (!sessionId) return;
  if (state.turnActive) return;
  if (state.workerStopped || state.workerRestarting) return;
  if (workerStateRef.current !== "running") return;
  if (statusRef.current !== "open") return;
  if (state.queuedPrompts.length === 0) return;
  // ... dispatch
}, [sessionId, status, workerState, state.turnActive, state.workerStopped, state.workerRestarting, state.queuedPrompts, dispatchPromptNow]);
```

So the moment all five conditions clear (WS reopens, worker reaches `running`, `AcpSessionAssigned` clears `workerStopped` / `workerRestarting`), the effect re-runs and drains. `#1144` made the queue durable across reconnects by removing the optimistic clear; queued items survive a tab reload via the `aoe:cockpit-state:v1:<sid>` localStorage snapshot at `web/src/hooks/useCockpit.ts:64-89`.

**Composer already flags this gap inline.** `web/src/components/cockpit/Composer.tsx:147-151`:

```ts
/** True when the cockpit WS is open. When false the composer
 *  refuses new submissions: prompts dispatched while disconnected
 *  would be lost (the POST /cockpit/prompt would fail with no way
 *  to retry). TODO(post-disconnect): queue locally and flush on
 *  reconnect instead of blocking. */
connected: boolean;
```

And `web/src/components/cockpit/Composer.tsx:968-973`:

```ts
function SendButton({ disabled = false }: { disabled?: boolean }) {
  // When the WS is closed we surface the offline state via `disabled`
  // and a swapped tooltip; ComposerPrimitive.Send would still try to
  // dispatch otherwise (it only knows about thread-runtime state, not
  // our connection status). TODO: queue prompts locally and flush on
  // reconnect instead of dropping them.
```

This issue is to close those TODOs.

## Root cause

Two-layer gap, both deliberate but stale:

1. **UI layer.** Composer treats "inactive session" as "no submissions accepted" rather than "submissions queue locally." The boolean prop is `connected`, the affordance is `disabled`, the tooltip is "reconnect to send." There is no path through which a user gesture creates a queued entry while the session is inactive.

2. **Hook layer.** `sendPrompt` predates the queue-while-resuming branch (#1088). The `statusRef !== "open"` arm was the only "can't send" path back then and it correctly raised an error. After #1088 added a sibling enqueue arm for `workerStateRef !== "running"`, the status arm should have been collapsed into the same enqueue path; it was not.

Together: even if the UI exposed the button, `sendPrompt` would still bail before reaching the existing enqueue dispatch.

### A) Minimum fix: route inactive-state sends through the existing queue

Two small changes, no new state.

**`web/src/hooks/useCockpit.ts:910-939`.** Replace the status-not-open error arm with an enqueue dispatch. The drain effect already handles fan-out on resume.

```ts
const sendPrompt = useCallback(
  async (text: string) => {
    if (!sessionId) return;
    // Session is not active: park locally; drain effect fires it once
    // status === "open" && workerState === "running" && !workerStopped.
    if (
      statusRef.current !== "open" ||
      workerStateRef.current !== "running" ||
      state.turnActive
    ) {
      dispatch({ kind: "enqueue_prompt", text });
      return;
    }
    await dispatchPromptNow(text);
  },
  [sessionId, state.turnActive, dispatchPromptNow],
);
```

Keep `dispatchPromptNow`'s own status guard at `useCockpit.ts:862-868`; that one protects the drain effect from racing a half-open WS.

**`web/src/components/cockpit/CockpitView.tsx:383` and `web/src/components/cockpit/Composer.tsx`.** Stop disabling the buttons on `!connected`. Repurpose `connected` to drive a label / tooltip swap that names the queue behavior, e.g.:

- Idle + connected → button labeled "Send", tooltip `Send · Enter`.
- Idle + not connected → button enabled, labeled "Queue", tooltip `Session not active; will send on reconnect`.
- Turn active + connected → existing `QueueSendButton` ("Queue follow-up (N pending)").
- Turn active + not connected → same `QueueSendButton`, tooltip says "will send on reconnect" instead of disabling.

`SendButton`'s underlying `ComposerPrimitive.Send` already routes the click through assistant-ui's `onSubmit` → `sendPrompt`; once `sendPrompt` enqueues instead of errors, no further wiring needed for that path. For the `QueueSendButton`, drop the `disabled` argument from `Composer.tsx:558` so the click reaches `enqueuePrompt` → `sendPrompt` → enqueue branch.

### B) UX: surface "queued for resume" alongside the existing strip

The `QueuedPromptsStrip` at `web/src/components/cockpit/CockpitView.tsx:341-346` already renders pending prompts above the composer, so the user can see what is parked. With (A) applied, prompts queued while the session is inactive will show up in that same strip without any additional UI work. Add one detail: when the worker / WS is inactive, the strip's empty-state copy or the first row should hint at "will send when session resumes" rather than the current "sent when current turn ends" wording.

If the team wants this more discoverable, prepend a single line on the strip when `!turnActive && (!connected || workerStopped || workerRestarting)`: "Pending until session resumes." Otherwise leave the strip alone.

### C) Out of scope: server-side queue durability

The current queue is per-tab (localStorage-backed). If the user closes the tab before reconnect, the queue dies. Making it server-side (e.g. `cockpit_events.db` row keyed by `(session_id, queued_at, text)`, or a column on `cockpit-workers/<id>.json`) is a separate, bigger ask. Track in a follow-up if useful; not required for this fix.

## Why now

- The TODO comments at `Composer.tsx:149-150` and `Composer.tsx:972` explicitly call this out as the missing piece after #1031 and #1088 landed. The pieces are already in place; this is wiring.
- A user typing a follow-up during an `aoe serve --restart` cycle (or after `aoe cockpit stop`, or while their phone reconnects to Wi-Fi) is a routine case on the cockpit web dashboard. Currently every such message is silently dropped with an error banner.
- The fix is small, contained to two files, and reuses the existing drain effect that already has #1144 / #1088 protection wired in.

## Open questions

1. Should `startupError` also flip into the queueable state? Today it does not gate `connected`, and `dispatchPromptNow` will POST and fail. Probably the queue should park here too, with a banner that says "Spawn failed; queued until you retry." Easy to add (extend the `sendPrompt` guard) but worth a separate decision.
2. The "Queue" button label when the session is inactive vs the existing "Queue follow-up (N pending)" button mid-turn: are those visibly different enough, or should we unify the affordance into a single button that changes its meaning based on `connected` + `turnActive`?
3. The error-banner copy `Cockpit disconnected; message not sent. Reconnect to retry.` (`useCockpit.ts:865`, `useCockpit.ts:920`) needs to change or disappear. Confirm both call sites: the one in `dispatchPromptNow` should stay (drain race protection); the one in `sendPrompt` goes away.

## Out of scope

- Server-side queue persistence (see (C)).
- Touching the `force_end_turn` watchdog path (#1196 covered the wedged-turn case; this issue is the "inactive but not wedged" case).
- The em dash fix in `Composer.tsx:979`'s tooltip is incidental; either fold it into this PR or open a tiny follow-up.

## Cross-references

- #1031 — parent feature that added the mid-turn queue + drain machinery this issue extends to inactive sessions.
- #1088 — added the `workerState !== "running"` enqueue branch this proposal mirrors for status.
- #1144 — made the queue durable across reconnect (no optimistic clear); load-bearing for this fix.
- #1037 — workers survive `aoe serve` crash/stop, which is what makes the queue-and-resume flow meaningful.
- #1196 — wedged-turn case where follow-ups silently dropped; related symptom, different root.

Seluj78:

## Description

When the cockpit session is inactive (WS closed, worker stopped via `user_stopped`, worker restarting via `restart_pending` / `agent_unresponsive` / `prompt_orphaned`), the composer used to hard-disable Send / QueueSend and `sendPrompt` would bail with the banner `Cockpit disconnected; message not sent. Reconnect to retry.` The user's text stayed in the textarea (per-session draft persists) but never entered the prompt queue.

This PR closes the asymmetry the TODO comments at `Composer.tsx:147-151` and `Composer.tsx:968-972` flagged after #1031 (mid-turn queue) and #1088 (worker-not-running queue arm). Three changes:

1. **`sendPrompt` mirrors the drain effect's gate set.** When status is not "open" OR worker isn't "running" OR `turnActive` OR `workerStopped` OR `workerRestarting`, enqueue locally instead of returning an error. The existing drain effect at `useCockpit.ts:953-1010` already waits on the same gates and fires the moment they clear, so no new wiring is needed.
2. **Composer buttons stay clickable when inactive.** The `connected` prop is repurposed from "hard-disable affordance" to a tooltip swap: Send becomes `Queue message until session resumes`, QueueSend names the same behavior alongside its pending-count badge. The stray em dash in the previous `Disconnected — reconnect to send` tooltip is gone.
3. **`QueuedPromptsStrip` heading reflects the parked state.** When the session is inactive, the heading swaps from `Queued (N)` to `Pending until session resumes (N)` so users can tell parked items wait for resume rather than next turn-end.

`dispatchPromptNow` keeps its own status guard for the drain-race window (see #1144). Server-side queue durability remains out of scope; queue persists per-tab via the localStorage snapshot at `useCockpit.ts:64-89`, unchanged.

Closes #1359

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start `aoe serve`, open a cockpit session, type a message, send. Replies stream as today.
- [ ] Stop `aoe serve --stop` (or kill the daemon). The dashboard banner shows reconnect status. Type a message in the composer, click Send. The textarea clears and the message appears in the strip above the composer with heading `Pending until session resumes (1)`.
- [ ] Start `aoe serve` again. After the WS reopens and the worker comes online, the parked message fires automatically and appears in the transcript as a user prompt.
- [ ] In the cockpit, click Stop on a running turn so the worker enters `user_stopped`. Type a follow-up, click Send. Message parks in the strip with the `Pending until session resumes` heading.
- [ ] From the stopped state, click Reconnect / restart the worker so a fresh `AcpSessionAssigned` clears `workerStopped`. The parked message fires automatically.
- [ ] While a turn is active AND the WS is open, click Queue with text in the box. Behavior unchanged from today: it queues as `Queued (N)` and drains on next `Stopped`.
- [ ] While a turn is active AND the WS is closed, the queue-send button stays clickable. Tooltip reads `Queue follow-up (N pending), will send on resume, Enter`.

## Checklist

## AI Usage

- [ ] No AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (kept the two-button Send/QueueSend split, punted `startupError` parking to a follow-up), and will run the manual test steps above before review.

## Overview
This PR makes the composer park user messages when the cockpit session or agent worker is inactive instead of failing. Messages typed during WebSocket reconnects, worker stops/restarts, or mid-turn are queued locally and automatically sent when the session resumes.

## What changed (high level)
- Composer send and queue buttons remain clickable while the session is inactive; tooltips now explain messages will be queued until resume.
- sendPrompt now enqueues messages when immediate sending is blocked by any gate (WS not open, worker not running, workerStopped/workerRestarting, or turnActive). An existing drain effect flushes the queue when gates clear.
- Queued strip heading switches from “Queued (N)” to “Pending until session resumes (N)” when the session/worker is inactive; worker-state guard wording was aligned.
- Queue persistence documented: queued entries are stored in localStorage keyed by session id (survive reloads or reopening the same-origin tab; not server-durable). Docs added describing drain modes (“combined” vs “serial”).
- Tests updated/added to assert enqueueing during WS CONNECTING and worker-stopped states and to verify drain behavior after resume.

## Benefits
- Improved UX: avoids blocking errors and preserves user input during temporary session/worker outages.
- Clearer communication: tooltips and heading label indicate queued status and expected behavior.
- More robust resume behavior: draining logic and guards reduce race conditions when resuming.
- Local durability: queued prompts survive page reloads or reopening the same-origin tab for the same session id.

## Technical notes
- sendPrompt now parks prompts into state.queuedPrompts when any blocking condition is present; when gates allow, it calls dispatchPromptNow.
- dispatchPromptNow retains its own guard to prevent drain-race windows.
- queuedPrompts persist in localStorage under key aoe:cockpit-state:v1:<sid>.
- Cockpit components (QueuedPromptsStrip) and Composer buttons now reflect a pendingResume/connected signal instead of disabling controls.
- Tests in useCockpit.queue.test.ts were adjusted/added to cover CONNECTING and worker-stopped enqueueing and subsequent draining.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1379?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


njbrake (approved):

LGTM. The product change is right: parking is strictly better than dropping, and the gate unification makes the send path and the drain effect read off the same state.

Pushed two small fixes addressing CodeRabbit's findings (commit a5ca677e):

1. Mirrored the worker-state check in the `QueuedPromptsStrip` heading toggle. The send path already parks on `cockpitWorkerState !== "running"`, but the heading was only checking `status`, `workerStopped`, and `workerRestarting`. A cold-starting worker now reads "Pending until session resumes (N)" instead of "Queued (N)".
2. Reworded the docs durability paragraph. The original said the queue was per-tab and dropped on tab close; the actual localStorage key is per-origin-per-session and survives tab close.

Also merged main in so the branch is current. Merging once CI is green.

Issue #1033, opened 2026-05-12
Cockpit: preserve unsent composer draft per session across navigation


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

In the cockpit composer, type a partial prompt, then click another session in the sidebar (or any other route change that unmounts the cockpit view). Come back to the original session — the text is gone. The composer's textarea is empty.

Every modern chat UI (Claude.ai, ChatGPT, Slack, Discord, iMessage) preserves the partially-typed message per-thread so you can switch conversations without losing your in-progress thought. The cockpit currently doesn't.

This bites harder on mobile, where typing is slow and switching sessions is a one-tap action — the cost of accidentally throwing away a 200-character draft is high.

## What's there today

The composer textarea is driven by assistant-ui's `<ComposerPrimitive.Input>` (`web/src/components/cockpit/Composer.tsx:188-199`), which holds its value in the runtime created by `useExternalStoreRuntime` (`web/src/components/cockpit/CockpitRuntime.tsx:75-98`). The runtime is **per-session-mount** — when the user navigates to a different session, the CockpitView unmounts and the runtime (with its in-progress input value) is garbage-collected. Re-mounting creates a fresh runtime with an empty input.

There's no draft-persistence layer between the textarea and the rest of the app. `useCockpit` only tracks server-driven session state (activity rows, turn flags, approvals); it never observes the composer's current text.

Reference: the repo already uses `localStorage` for similar "preserve user-side UI state across mount/unmount" patterns —
- `web/src/components/WorkspaceSidebar.tsx:96-575` for sidebar width
- `web/src/components/RightPanel.tsx:25-309` for vsplit ratio
- `web/src/hooks/useRepoGroups.ts:9-135` for collapsed group state

So the precedent and storage layer exist; the composer just doesn't tap into them.

## Proposal

Per-session draft persistence in `localStorage`, keyed by session id.

### Storage shape

Key format: `cockpit:draft:<session_id>` (mirrors the existing `cockpit:` prefix used elsewhere). Value is the raw textarea string. Empty draft → key removed (don't litter localStorage with empty strings).

### Wiring

In `Composer.tsx`:

[...]

PR #1380, opened 2026-05-21
fix(web): cockpit composer drafts lose tail keystrokes on refresh + orphan keys never pruned

### LINKED ISSUE #1358: Cockpit composer: drafts lose tail keystrokes on refresh + orphan draft keys never pruned

## Problem

Two related cockpit composer draft persistence gaps:

1. **Refresh / tab close loses the last few lines of an in-progress draft.** Type into the composer, refresh the page (or close the tab and reopen it). The first few characters survive; the tail you typed in the ~250ms before refresh is gone.
2. **Drafts for deleted / abandoned sessions never get pruned.** localStorage accumulates `cockpit:draft:<id>` keys for every session that ever held an unsent prompt, including sessions deleted from this tab, another tab, or another device. Over time this pads localStorage usage and feeds the QuotaExceeded crash class from #1345.

Both bite the same surface (the cockpit composer's draft persistence layer landed in #1033), but neither is addressed by it.

### Draft write debounce

`web/src/components/cockpit/Composer.tsx:339-354`:

```tsx
let writeTimer: number | null = null;
const flush = () => {
  writeTimer = null;
  setDraft(sessionId, composerRuntime.getState().text);
};
const unsub = composerRuntime.subscribe(() => {
  if (writeTimer !== null) window.clearTimeout(writeTimer);
  writeTimer = window.setTimeout(flush, 250);
});
return () => {
  unsub();
  if (writeTimer !== null) {
    window.clearTimeout(writeTimer);
    flush();
  }
};
```

250ms debounce on every keystroke. The cleanup runs on **React unmount** (sidebar navigation, route change), so navigation flushes correctly; #1033's original wiring covered that case.

But a full page reload (browser refresh, PWA cold start, mobile OS evicts the tab) does not call React effect cleanup. Whatever was pending in `writeTimer` at the moment the page died is lost. Empirically that's the last ~1-2 sentences of typing on a fast typer.

### No flush on `beforeunload` / `pagehide`

Composer.tsx never installs a `beforeunload` or `pagehide` listener. The sibling diff-comments hook does, at `web/src/hooks/useDiffComments.ts:72-83`:

```ts
// Flush any pending debounced write before the tab closes / hides
// so the user doesn't lose the last keystrokes on a refresh.
useEffect(() => {
  if (!sessionId) return;
  const flush = () => saveComments(sessionId, state);
  window.addEventListener("beforeunload", flush);
  window.addEventListener("pagehide", flush);
  return () => {
    window.removeEventListener("beforeunload", flush);
    window.removeEventListener("pagehide", flush);
  };
}, [sessionId, state]);
```

Same shape, same debounce concern (200ms there vs 250ms here). The composer just doesn't have the equivalent.

### No pruning of orphan draft keys

`web/src/lib/cockpitDrafts.ts:41-52` only clears a key when its text becomes empty:

```ts
export function setDraft(sessionId: string, text: string): void {
  try {
    if (text.length === 0) {
      localStorage.removeItem(draftKey(sessionId));
    } else {
      localStorage.setItem(draftKey(sessionId), text);
    }
  } catch {
    /* localStorage blocked / quota; persistence is best-effort */
  }
  notify(sessionId);
}
```

No `clearDraft(sessionId)` helper, no startup sweep, no listener tied to session-list changes.

Session deletion happens in `web/src/App.tsx:388-415`:

```tsx
const handleConfirmDelete = useCallback(async (options) => {
  // ... server-side delete ...
  // Drop the per-session cockpit cache so a recreated session with
  // the same id doesn't briefly show the prior transcript on
  // remount before fetchReplay clears it.
  clearCockpitCache(sessionId);
  // ...
}, [...]);
```

`clearCockpitCache(sessionId)` clears the `aoe:cockpit-state:v1:<id>` entry but never touches `cockpit:draft:<id>`. Sessions deleted from another tab or another device are even worse off, the deletion path doesn't run at all on this tab.

For comparison, the per-session state cache **does** have a startup sweep at `web/src/hooks/useCockpit.ts:151-183` (`sweepExpiredStorage`) that walks `aoe:cockpit-state:v1:*` and drops entries older than 7 days. Drafts have nothing equivalent.

## Root cause

Two independent gaps in the draft persistence layer shipped by #1033.

1. **Refresh loses tail keystrokes.** Debounce is the only write trigger; nothing flushes on page unload. The 250ms window is small enough to feel fine in normal SPA navigation (effect cleanup runs synchronously) but matches a full keystroke burst on refresh.

2. **Drafts orphan indefinitely.** `cockpit:draft:<id>` is written on every keystroke and only removed when the user explicitly clears the textarea or sends the prompt. Session deletion does not remove it. No periodic sweep. No quota-driven eviction.

### A. Flush on `pagehide` / `beforeunload` (immediate fix for the "lost keystrokes" symptom)

Mirror the diff-comments pattern in `web/src/components/cockpit/Composer.tsx`. In the same effect that wires up the debounce, also wire an unload flush. Sketch:

```tsx
useEffect(() => {
  // ... existing seed-from-saved logic ...

  let writeTimer: number | null = null;
  const flush = () => {
    if (writeTimer !== null) {
      window.clearTimeout(writeTimer);
      writeTimer = null;
    }
    setDraft(sessionId, composerRuntime.getState().text);
  };
  const unsub = composerRuntime.subscribe(() => {
    if (writeTimer !== null) window.clearTimeout(writeTimer);
    writeTimer = window.setTimeout(flush, 250);
  });
  window.addEventListener("beforeunload", flush);
  window.addEventListener("pagehide", flush);
  return () => {
    unsub();
    window.removeEventListener("beforeunload", flush);
    window.removeEventListener("pagehide", flush);
    flush();
  };
}, [composerRuntime, sessionId]);
```

Optional: also flush on `visibilitychange` to `hidden` so iOS Safari (which fires `pagehide` only on actual unload, not when the app backgrounds) covers app-switch too. The `logger.ts:184-190` flush wires `visibilitychange` + `pagehide` together; the same pattern works here.

Optional: drop the debounce from 250ms to 100ms. Cheap (localStorage write of a small string is sub-millisecond); buys back most of the loss-on-mobile-OS-kill case where `pagehide` may not fire. Original 250ms in #1033 was conservative; revisit.

### B. Prune drafts on session delete (immediate fix for the orphan-key symptom)

Add `clearDraft(sessionId)` to `cockpitDrafts.ts`:

```ts
export function clearDraft(sessionId: string): void {
  setDraft(sessionId, "");
}
```

Call it from `App.tsx:412` alongside `clearCockpitCache`:

```tsx
clearCockpitCache(sessionId);
clearDraft(sessionId);
```

That covers the local-tab delete path.

### C. Startup sweep against the session list (covers cross-tab / cross-device deletes)

Add a `sweepOrphanDrafts(activeSessionIds: ReadonlySet<string>)` in `cockpitDrafts.ts`, modeled on `sweepExpiredStorage` in `useCockpit.ts:151-183`:

```ts
export function sweepOrphanDrafts(activeSessionIds: ReadonlySet<string>): void {
  if (typeof window === "undefined") return;
  try {
    const toRemove: string[] = [];
    for (let i = 0; i < window.localStorage.length; i++) {
      const k = window.localStorage.key(i);
      if (!k) continue;
      const sid = sessionIdFromKey(k);
      if (sid === null) continue;
      if (!activeSessionIds.has(sid)) toRemove.push(k);
    }
    for (const k of toRemove) window.localStorage.removeItem(k);
  } catch {
    /* ignore */
  }
}
```

Call it once from the app shell after the initial session list resolves, e.g. in `App.tsx`'s session-list effect. This catches drafts for sessions deleted on another tab / device since the last load.

### D. Optional: time-based sweep (defense in depth)

If the user never reloads the dashboard for weeks, B + C don't fire. A TTL sweep ("any draft older than 30 days") would catch that, but it needs the storage to track `lastWrittenAt`. That changes the storage shape from raw string to `{ text, savedAt }`, which is a small migration. Worth doing only if telemetry shows orphan drafts as a meaningful contributor to the quota; otherwise B + C should be enough.

## Why now

- The lost-keystrokes case is a daily annoyance for any user who refreshes (PWA cold start, auth redirect, tab swap). The original #1033 fix landed the persistence layer; this is the closing 5% that makes it feel reliable.
- The orphan-prune case piles on top of #1345 (QuotaExceeded crash). Every never-pruned draft contributes to the quota that crashes the sidebar resize bar. B + C remove one of the growth sources cheaply.
- Both are localised to `Composer.tsx`, `cockpitDrafts.ts`, and one line in `App.tsx`. No protocol work, no Rust, no design.

## Open questions

- Should the `pagehide` flush also short-circuit the debounce on every keystroke (i.e. write synchronously when the tab is hidden / page is unloading)? Probably not, the listener path is enough. But on mobile, `pagehide` is the more reliable signal than `beforeunload`; verify both fire on iOS PWA reload.
- For C, should the sweep run on every app mount or be debounced to once per session (e.g. once per day per browser)? Once per mount is simpler and the cost is O(localStorage entry count), which is bounded. Lean simple.
- Does dropping debounce to 100ms cause measurable input lag on slow Android devices? `setItem` for <2KB strings is sub-ms; doubt it, but worth confirming if QA picks it up.
- Should B + C also fire `notify(null)` so the sidebar's "unsent draft" dot disappears immediately? `setDraft(sessionId, "")` already calls `notify(sessionId)`, so B is covered. C needs an explicit notify after batch removal.

## Out of scope

- Server-side draft sync across devices. localStorage stays the source of truth.
- Migrating to IndexedDB or a different storage backend (different problem, see #1345 out-of-scope).
- A user-visible "you have N orphan drafts, clear them?" affordance. The sweep should be silent.
- The QuotaExceeded crash class itself (#1345). This issue reduces growth pressure but doesn't replace the unguarded-setItem audit that #1345 calls for.

## Cross-references

- #1033 — original draft persistence layer; this issue closes gaps it left open.
- #1345 — QuotaExceeded crash on sidebar resize; orphan drafts contribute to the growth that triggers it. Pruning here is one input to the structural fix proposed there.
- #1341 — cockpitDrafts Vitest coverage. A regression test "pagehide fires => flush wrote to localStorage" and "sweep removes orphan keys not in active list" would lock the new behaviour.
- `web/src/components/cockpit/Composer.tsx:339-354` — current debounce wiring.
- `web/src/hooks/useDiffComments.ts:72-83` — canonical pagehide/beforeunload flush pattern.
- `web/src/hooks/useCockpit.ts:151-183` — canonical startup-sweep pattern.
- `web/src/App.tsx:388-415` — session delete handler; B hooks in here.

Seluj78:

## Description

Closes two gaps in the cockpit composer's draft persistence layer that shipped in #1033.

**Problem 1: refresh loses tail keystrokes.** The composer mirrors textarea content into `localStorage` (`cockpit:draft:<sessionId>`) on a 250ms debounce. The cleanup runs on React effect unmount (sidebar navigation), but a full page reload / PWA cold start / mobile OS tab eviction does not call effect cleanup, so whatever is pending in the debounce timer when the page dies is lost. On a fast typer that's the last sentence or two of the draft.

`web/src/components/cockpit/Composer.tsx` now also installs `beforeunload`, `pagehide`, and `visibilitychange→hidden` listeners that all call the same `flush()`. The visibility listener covers iOS Safari, which only fires `pagehide` on real unload, not on app-switch. Same pattern as the sibling `web/src/hooks/useDiffComments.ts`.

**Problem 2: orphan draft keys never get pruned.** A draft key was only removed when the user typed the textarea back to empty. Deleting a session left its key behind forever, padding `localStorage` and feeding the QuotaExceeded crash class tracked in #1345. Deletes from another tab or another device were even worse off since the local delete path never ran.

`web/src/lib/cockpitDrafts.ts` now exports:
- `clearDraft(sessionId)`: thin wrapper around `setDraft("")` so the delete path doesn't have to import an empty-string sentinel.
- `sweepOrphanDrafts(activeSessionIds)`: walks `localStorage` once, removes every `cockpit:draft:<id>` key whose session id is not in the active set, fires one wildcard `notify(null)` after the batch so the sidebar "unsent draft" dot recomputes.

`web/src/App.tsx`:
- `handleConfirmDelete` calls `clearDraft(sessionId)` next to `clearCockpitCache(sessionId)`, so a local-tab delete drops the draft key immediately.
- A one-shot effect calls `sweepOrphanDrafts` after the first non-empty `sessions` list resolves, catching cross-tab / cross-device deletes since the last load. Bounded by `localStorage` entry count.

Behavior covered by 9 new vitest cases in `web/src/lib/cockpitDrafts.test.ts` (27/27 pass in that file, 701/701 in the full suite). The composer pagehide listener stays untested at the unit level, matching the precedent set by `useDiffComments.ts`; the live-playwright `cockpit.spawn-prompt` entry already exercises the composer surface end-to-end.

Closes #1358

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open a cockpit session, type a multi-sentence draft into the composer, hit browser refresh within ~250ms of the last keystroke. Full text restored on reload (no missing tail).
- [ ] On iOS Safari (or simulated via DevTools), type a draft, switch to another app / background the PWA, return. Draft text preserved.
- [ ] Delete a cockpit session from the active tab. In DevTools localStorage, confirm `cockpit:draft:<deleted-id>` is gone immediately.
- [ ] Open two tabs against the same backend. In tab A, type a draft for session X then close the tab without sending. In tab B, delete session X. Refresh tab A. localStorage no longer contains `cockpit:draft:<X-id>`.
- [ ] In the sidebar, an unsent-draft dot for a deleted session should disappear immediately on local delete and not reappear after refresh.

## Checklist

## AI Usage

- [ ] No AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (skip the optional debounce-tightening and TTL sweep, keep the visibility-change listener for iOS Safari parity), and will run the manual test steps before marking ready for review.

**NOTE:**

## Overview

This PR fixes two cockpit composer draft-persistence bugs: missing the last keystrokes on page unload/refresh, and orphaned draft keys left in localStorage after sessions are deleted elsewhere.

## What changed (high level)

- Composer: adds a synchronous flush() and registers beforeunload, pagehide, and visibilitychange→hidden handlers so pending debounced draft writes are flushed on page lifecycle events.
- Draft API: adds clearDraft(sessionId) to remove a single session's persisted draft immediately.
- Orphan pruning: adds sweepOrphanDrafts(activeSessionIds) to scan localStorage for cockpit:draft:<id> keys and remove those whose ids are not in the active session set.
- App integration: AppContent calls clearDraft when a session is deleted and runs a one-shot sweepOrphanDrafts once after the initial non-empty sessions list resolves.
- Tests: adds nine Vitest tests for clearDraft and sweepOrphanDrafts; existing end-to-end tests exercise the pagehide listener.

## Benefits

- Preserves the final ~250ms of typing on reloads, PWA cold starts, and mobile/tab evictions (including iOS Safari).
- Removes stale cockpit draft entries created by deletes in other tabs/devices, keeping storage lean and UI unsent-draft indicators accurate.
- Emits a single notification after batch removals so UI updates consistently.

## Technical notes (concise)

- flush(): cancels pending debounce timer and synchronously writes current composer text to localStorage.
- Lifecycle handlers: beforeunload, pagehide, and visibilitychange→hidden call flush() to avoid losing the last keystrokes.
- clearDraft(sessionId): convenience wrapper around setDraft(sessionId, "") that ensures immediate removal and notification.
- sweepOrphanDrafts(activeSessionIds): best-effort localStorage scan that deletes cockpit:draft:<id> keys not in the provided active set and fires one wildcard notify(null) if any deletions occur; no-ops when window is undefined and swallows localStorage iteration errors.
- App behavior: runs sweep once after sessions load to catch cross-tab/device deletes since last load.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1380?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


njbrake (approved):

LGTM. The pagehide / beforeunload / visibilitychange triplet mirrors the precedent in `useDiffComments.ts` and the `clearDraft` + `sweepOrphanDrafts` split keeps the delete and the startup-sweep paths distinct.

Pushed two changes:

1. CodeRabbit's finding on `App.tsx:213` was real. The sweep effect bailed on `sessions.length === 0`, so a brand-new user with zero sessions never had orphan drafts swept. Gated on the new `sessionsLoaded` sentinel from #1375 instead, so the sweep runs after the first fetch settles regardless of result.
2. Merged main in and resolved the `cockpitDrafts.test.ts` conflict (kept both this PR's clearDraft/sweep suites and main's toast-dedupe suite from #1345).

32 Vitest cases pass. Merging once CI is green.

Issue #1140, opened 2026-05-15
Rework release cadence and update notification UX


njbrake:
_Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Background

We currently release new versions infrequently (gap of ~4 days to weeks between releases) and surface updates to users via a persistent banner in the TUI. This has two problems:

1. **Release cadence is slower than peer tools.** A survey of similar agent tools (conductor, opencode, emdash, cline, cmux, pi) shows most release multiple times per week. Infrequent releases bundle many changes into single versions and make the project look less active than it actually is.
2. **The update banner is interruptive.** If we move to a faster cadence without changing how updates are surfaced, users will see the "update available" banner constantly, which risks becoming an annoyance (related: #984).

## Decisions

After discussion in Discord, we landed on the following plan.

### Release cadence

- Target **at least once per week** as the baseline cadence.
- Automate **release staging** (not the release itself): every Wednesday, a GitHub Action opens a PR that bumps the version and stages a release. The maintainer reviews the PR, chooses the correct semver bump (major/minor/patch) based on the included changes, and merges to trigger the actual release.
- Default bump in the staged PR is **patch**; the maintainer adjusts to minor/major if the change set warrants it.
- The PR description should summarize the changes since the last release to make the semver decision easier (likely leveraging conventional commit prefixes).

We considered a two-channel (stable + pre-release) model like Zed, but rejected it as too much process overhead for a project of this size.

### Update notification UX

Three coordinated changes to how the TUI surfaces updates:

1. **New config option: `update_check_mode`** with three states:
   - `auto`: silently auto-update in the background.
   - `notify` (default): show the banner when an update is available.
   - `off`: never check or notify.

   Modeled on opencode's approach.

2. **Show banner once per release, snooze on dismiss.** Snooze is tracked **per-version**, not on a time-based cooldown. If a user dismisses the banner for v1.5.3, they won't see it again until v1.5.4 ships. Simpler mental model than a global cooldown timer, and addresses #984 cleanly.

3. (Stretch / future) **Defer disruptive prompts until idle**, conductor-style: the banner can exist, but any "restart now to update" action should not fire while the user has an active session.

## Documentation

Once implemented, we should document:
- The new \`update_check_mode\` config field (settings TUI + config docs).
- The release cadence and how the staging PR workflow works (likely in \`CONTRIBUTING.md\` or a maintainer-facing doc).

## Out of scope

[...]

PR #1397, opened 2026-05-21
test(web): add UI-driven cockpit user-story Playwright specs

### LINKED ISSUE #1383: Cockpit: add user-story Playwright specs that drive the UI, not the REST contract

## Problem

The cockpit has deep test coverage at two layers: the REST contract (12 live Playwright specs under `web/tests/live/cockpit-*.spec.ts`) and the React component shell (Vitest suites under `web/src/components/cockpit/__tests__/`). Between them is the layer no test currently exercises: the end-to-end user journey through the UI, from sidebar click to composer keystrokes to ApprovalCard button click to spinner clearing. Every cockpit live spec drives the server over `fetch()` and asserts on the disk-backed `cockpit/replay` endpoint; none open the CockpitView, type into the Composer, or click "Allow"/"Deny" on an approval.

Concretely, of the 12 live cockpit specs, exactly zero call `page.goto()` or `page.click()`. The only file that does (`cockpit-escape-no-cancel.spec.ts`) is `base.skip(...)` and has been since #1228.

The umbrella #986 closed with all its component-level scope landed (settings panels, profiles, devices, sessions, sidebar, right-panel, wizard, git-clone, directory browser, devices, read-only, ensure-session-restart). What it did not commit to was user-story specs that exercise multiple components together through real user input. That gap means a refactor of `CockpitView.tsx`, `Composer.tsx`, or `ApprovalCard.tsx` can ship green CI while breaking the actual flow a user sees in the browser.

## What's there today

REST-driven cockpit specs (assert against `/api/sessions/:id/cockpit/replay`):

```
web/tests/live/cockpit-approval.spec.ts          → POST cockpit/approvals/:nonce { decision }
web/tests/live/cockpit-cancel.spec.ts            → POST cockpit/cancel        (still skipped)
web/tests/live/cockpit-context-primer.spec.ts    → GET  cockpit/context-primer
web/tests/live/cockpit-disable.spec.ts           → DELETE cockpit
web/tests/live/cockpit-files.spec.ts             → GET  cockpit/files
web/tests/live/cockpit-force-end-turn.spec.ts    → POST cockpit/force-end-turn
web/tests/live/cockpit-lagged-ws.spec.ts         → WS reconnect/replay        (still skipped)
web/tests/live/cockpit-mode-switch.spec.ts       → POST cockpit/mode
web/tests/live/cockpit-replay-catchup.spec.ts    → GET  cockpit/replay?since=N
web/tests/live/cockpit-spawn-prompt.spec.ts      → POST cockpit/enable + cockpit/prompt
web/tests/live/cockpit-substrate-switch.spec.ts  → POST cockpit/enable + cockpit/disable
```

Sample assertion shape, from `cockpit-spawn-prompt.spec.ts:60-77`:

```ts
const replay = await fetch(
  `${serve.baseUrl}/api/sessions/${sessionId}/cockpit/replay?since=0`,
).then((r) => r.json());
const frames: unknown[] = Array.isArray(replay) ? replay : replay.frames ?? [];
const json = JSON.stringify(frames);
if (
  json.includes("agent_message_chunk") ||
  json.includes("AgentMessageChunk")
) {
  sawChunk = true;
  break;
}
```

These specs validate the server publishes the expected event. They never assert that the React reducer at `web/src/lib/cockpitTypes.ts:351` applies the event, that `CockpitView.tsx:679` renders the message, that `Composer.tsx:441-445` flips its placeholder when `turnActive` is true, or that `ApprovalCard.tsx:172` resolves on click.

Component selectors are already in place where user-story specs would need them:

- `Composer.tsx:441-445`: textbox `placeholder="Send a message…  Type @ for files, / for commands"` / `"Queue a follow-up… (sent when current turn ends)"`. `page.getByRole("textbox", { name: /Send a message|Queue a follow-up/i })` works (already used by the skipped `cockpit-escape-no-cancel` spec at `cockpit-escape-no-cancel.spec.ts:76`).
- `Composer.tsx:978`: send button `aria-label="Send message"`.
- `Composer.tsx:1000`: stop button `aria-label="Stop"`.
- `Composer.tsx:1039`: queue follow-up button `aria-label="Queue follow-up message"`.
- `ApprovalCard.tsx:98`: dialog `role="alertdialog"`, `aria-label="Approval needed: <toolname>"`.
- `ApprovalCard.tsx:172,189,209`: buttons by text "Allow" / "Always" / "Deny".
- `SwitchSubstrateAction.tsx:98`: substrate-switch trigger has a dynamic `aria-label`.

The harness side is also ready. `serveCockpit` (per `docs/development/playwright.md:71`) installs the fake ACP shim under `claude`, `claude-agent-acp`, and `aoe-agent` and flips the cockpit master switch. The fake supports scripted `agent_message_chunk`, real `session/request_permission` round-trips (since #1303), `setMode`, and `cancel`. Three skips remain (`cockpit-cancel`, `cockpit-escape-no-cancel`, `cockpit-lagged-ws`), each with its own root cause distinct from the UI gap.

## Root cause

Hierarchical follow-ups under #986 split coverage by component file (`test(web): ApprovalCard Vitest`, `test(web): cockpitDrafts Vitest`, `test(web): right-panel diff Playwright`). That partition produced excellent unit-level depth but no spec was scoped to a multi-component flow. The cockpit live specs in #1224 were explicitly written as "tracers" (the word appears in #1224's body) covering one REST endpoint each, with UI clicks deferred until selectors stabilized (`cockpit-approval.spec.ts:114` says so verbatim). Selectors are stable now; the follow-up never got filed.

## Proposal

Land a small set of UI-driven live Playwright specs under `web/tests/live/cockpit-stories/`, each one shaped as a single user story end-to-end. Use the existing `serveCockpit` fixture and a per-spec custom `fakeAcpScript` where the default chunk-then-stop script is not enough.

### User stories to cover

"Story" here means one user-visible flow. A story can land as multiple spec files when the flow has natural sub-cases (one happy-path, several failure modes, mobile + desktop variants). The list below is the working scope; sub-issues will refine each story into concrete spec files with selectors and assertions.

#### Cockpit composer / messaging

- Send message via Enter (desktop).
- On mobile, pressing Enter inserts a newline; it does NOT send.
- Queue a follow-up message during an active turn, then navigate to another page; queued prompt fires when first turn ends.
- See streamed agent response render in the chat area.
- Stop button while agent is in a sub-agent task.
- Stop button while a tool is running.
- Stop button while agent is thinking.
- Mode switch (cockpit ↔ tmux substrate) via SwitchSubstrateAction.
- ApprovalCard "Allow" click resolves, turn continues.
- ApprovalCard "Deny" click resolves, agent stops.
- Fold / unfold queued messages.
- Edit a queued message; delete a queued message.
- Draft persistence: type into composer, reload the page; draft is still in composer.
- Draft persistence: type into composer, switch to another session, switch back; draft is still in composer.
- Draft persistence: same as above, but for a session in the new-session wizard / not-yet-created session (draft survives reload + session navigation before launch).

#### Sidebar / left panel / topbar

- Open the sidebar; view in-session changes.
- Sidebar: switch the `vs` base branch and see the diff recompute.
- Diff viewer shows changes for the active session.
- Send comments on a diff hunk.
- Toggle the left sidebar to hide / show the session list.
- Resize the left sidebar via its drag handle; width persists across reload.
- Resize the right panel via its drag handle; width persists across reload.
- See tasks updating live in the sidebar (out-of-session view).
- See tasks updating live in the topbar (in-session view).
- Fold / unfold groups of sessions.
- Filter sessions via the sidebar filter input.

#### Auth

- User can sign in.
- User can sign out.

#### Modals / help

- View About modal.
- View Help overlay.

#### Keyboard shortcuts (one story per binding)

- `Cmd/Ctrl + K`: open new-session modal.
- Shortcut to go to dashboard.
- Shortcut to open diff panel.
- Shortcut to open sidebar.
- Shortcut to show help overlay.
- Shortcut to sign out.
- Shortcut to switch between sessions.
- Open the session switcher; close with Escape.

#### New-session wizard

The wizard is large; expect one story per feature / advanced toggle / field. Examples:

- Open new-session modal.
- Click on an existing session group → modal opens pre-filled.
- Edit any field after pre-fill works (per field).
- Select a recent project.
- Browse the disk for a git repo.
- Clone a repo from URL.
- After project is chosen, add extra repos.
- In session-info step: set session title; branch + worktree auto-generate from title.
- Click "Launch" → new session created.
- Press `Cmd/Ctrl + Enter` → new session created.
- During session creation, show banner + queued messages while waiting.

#### Projects view

- Go to projects page.
- View list of projects.
- Add a project.
- Remove a project.

#### Profiles

- Change active profile.
- Create a new profile.
- Rename a profile.
- View setting and per-session differences between profiles.

#### Settings

- One story per setting: change the value, expected behavior follows. Goal: every setting in `SettingsView` is exercised once.

#### Approvals (cross-cutting)

- See ApprovalCard streamed in mid-turn → Allow → continue.
- See ApprovalCard streamed in mid-turn → Deny → stop.

The shared point across whatever lands: each spec drives the React UI through clicks and keystrokes, asserts on rendered DOM (not on `/cockpit/replay`), and exercises the reducer-to-render plumbing under real WS broadcast that no current test covers.

### Harness changes needed

Locked once the story list is locked. Best guesses based on the candidate flows:

- The fake ACP agent at `web/tests/helpers/fakeAcpAgent.mjs` already supports `permission_request` round-trips after #1303. If a Deny-path flow lands, verify a `Deny` resolution produces a usable `stopReason`; if not, extend the fake to map decision → next turn behavior (small, additive).
- Add a tiny helper in `web/tests/helpers/cockpit.ts` to wait for the cockpit view to be mounted in the DOM (current `waitForCockpitReady` polls replay, not the UI). A dedicated `waitForCockpitView(page)` that resolves when `getByRole("textbox", { name: /Send a message|Queue a follow-up/i })` becomes visible would let every story spec start from a known UI state.

### Coverage matrix

Add one new entry per spec (kind `live-playwright`) with `components` listing every file under `web/src/components/cockpit/**` the spec actually exercises. The validator will keep `web/tests/coverage-matrix.json` honest.

### Tracking

This is a tracker issue. File one sub-issue per spec under it (parts of #986 used the same model). Each sub-issue can land independently; they share the harness and selector work, not the assertion plan.

## Why now

- Three of the most central components (`CockpitView.tsx` ~66KB, `Composer.tsx` ~42KB, `ToolCards.tsx` ~61KB) have no live UI coverage. Any of the in-flight refactors touching them ships untested at the integration layer.
- Selectors are stable (the skipped `cockpit-escape-no-cancel` spec was the prerequisite check for that).
- The fake-ACP harness is finally able to drive real permission round-trips (#1303), which unlocks any approval-flavored stories.
- The deferred cleanup of three remaining skipped cockpit specs (`cockpit-cancel`, `cockpit-escape-no-cancel`, `cockpit-lagged-ws`) is adjacent but orthogonal; a UI-driven Stop-mid-turn story would effectively supersede `cockpit-cancel`.

## Policy change (part of this work)

`AGENTS.md` must be updated as part of this issue so the rule is enforceable on future PRs. The existing "Mandate" paragraph under `### Web Dashboard Playwright Tests` covers coverage-matrix entries and component-file tests; it does not require a user-story spec. Add a second mandate immediately after it.

Concrete edit, dropped into `AGENTS.md` right after the existing `**Mandate:**` paragraph (around line 90):

```md
**User-story spec mandate:** any new user-facing dashboard feature MUST land with a user-story Playwright spec that drives the React UI end-to-end (clicks, keystrokes, navigation) and asserts on rendered DOM. Component-level Vitest coverage and REST-contract live specs do not satisfy this requirement; the user-story spec is what catches reducer-to-render plumbing breakage, route gating, and multi-component focus / keyboard handling. Mocked Playwright is acceptable when the flow does not depend on real backend state; otherwise use live Playwright. Tracking issue: #1383.
```

This must land in the same PR (or stack) that introduces the first user-story spec, so the rule is testable from day one.

## Open questions

- If a UI-driven Stop-mid-turn story lands, supersede the skipped `cockpit-cancel.spec.ts` or keep both (one REST contract, one UI story)? My read: delete the skipped one once the UI version lands.
- Live in `web/tests/live/cockpit-stories/` (subdirectory) or alongside existing tracer specs as `cockpit-story-<n>.spec.ts`? Subdir reads cleaner once the count grows past 3.
- "Every setting" is a big surface. Land it as one sub-issue with N sub-tasks, or split by settings panel (ThemeSettings story, SoundSettings story, etc.)?
- Several stories already have Vitest-level coverage (settings payloads, cockpit drafts, approval card render). The story specs layer the UI driver on top; they do not replace the Vitest tests.
- Story 6 needs both cockpit and tmux substrates to be functional inside one isolated `HOME`. Worth confirming the fixture's tmux side still attaches cleanly after a substrate swap; if not, that's a precursor work item.

## Out of scope

- Unskipping `cockpit-lagged-ws.spec.ts` (separate; depends on synthetic WS lag, not UI).
- A visual-regression suite over the cockpit (screenshot diffing is out per #986's own out-of-scope note).
- ToolCard per-kind render coverage end-to-end (the matrix already covers it under `cockpit.tool-cards` via Vitest; only add a UI story if a specific kind becomes a regression hotspot).
- Mobile-viewport variants of these stories. Worth a follow-up once desktop versions stabilize.

## Cross-references

- #986 — umbrella for web Playwright coverage (closed); this issue is the user-story follow-up.
- #1224 — "cockpit completion coverage" (closed); landed the 8 REST-driven tracer specs that this issue layers UI stories on top of.
- #1237 — fixed the AuthRequired handshake so cockpit tracers could unskip; same shim work makes UI stories possible.
- #1303 — added real `session/request_permission` round-trips to the fake ACP agent; unblocks any approval-flavored story.
- #1216 — open; resume-idle watchdog flips sidebar to Idle mid-turn. A "spinner clears only on real Stopped" story would catch the regression.

Seluj78:

## Description

Lands the first end-to-end wave of cockpit user-story Playwright specs called for in #1383. Closes the gap between the REST-contract tracers under `web/tests/live/cockpit-*.spec.ts` and the React component Vitest suites: 83 new specs under `web/tests/live/cockpit-stories/` drive the dashboard UI through clicks, keystrokes, navigation, and (for mobile) WebSocket bytes, and assert on rendered DOM rather than `/cockpit/replay`.

Also lands the harness, policy, and docs pieces that pair with the specs:

- `web/tests/helpers/cockpit.ts` gains `waitForCockpitView(page)` so a story spec can start from a known UI state after `page.goto`. Pairs with the existing `waitForCockpitReady` (supervisor handshake).
- `web/tests/helpers/fakeAcpAgent.mjs` gains a `wait_ms` script entry so scripts can pause inside a turn without blocking the rest of the script. Lets stop / queue / approval stories observe a long-running turn.
- `AGENTS.md` picks up a second mandate paragraph: any new user-facing dashboard feature must ship with a UI-driven Playwright spec, not just component / REST coverage.
- `docs/development/playwright.md` documents the cockpit-stories pattern with a canonical example.
- `web/tests/coverage-matrix.json` gets one entry per new spec; `validate-coverage-matrix.mjs` is green.

The PR groups specs by user-visible flow rather than by component. Surfaces covered include:

- **Cockpit composer / messaging**: Enter to send, mobile Enter newline, multi-chunk streaming, Stop during thinking / tool / sub-agent, queue follow-up (with and without navigation), edit / delete / fold queued, ApprovalCard Allow / Deny, mode-picker, substrate switch, toolbar @ + /, drafts surviving reload and session switch.
- **Sidebar / topbar**: filter, fold groups, drag-handle resize persistence, click-to-navigate, topbar toggle, plan-progress mini.
- **Auth, modals, shortcuts**: sign-out, About + Help (open + Escape), Cmd+K palette, Cmd+B sidebar, Cmd+Opt+B right panel, Cmd+\`, ?, s, n, Shift+D, Escape; session switch via palette.
- **Wizard**: open / close (button + Escape), tabs, browse, clone-URL input, recent project, group prefill, title + group field, worktree toggle, extra repos, branch autogen, Launch button + Cmd+Enter, creating-banner, last-tool persistence.
- **Projects + Profiles + Settings**: open / add / remove projects; create / rename / delete / switch profiles; per-profile setting isolation; representative settings round-trips (theme, tmux status / mouse, logging level, update interval, sound toggle).
- **Delete sessions**: with-worktree variant, deleting the active session.
- **Mobile terminal toolbar (iPhone 13 emulation)**: Arrow up, Tab, Esc, Ctrl latch, Ctrl+C interrupt, hidden on desktop. Each input spec patches `WebSocket.prototype.send` and asserts the PTY bytes are the expected escape sequence.
- **Diff**: file row visibility with seeded uncommitted changes, BasePicker open, hunk comment via "+" gutter button.

Closes #1383

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## How to test

User-runnable verification steps; each ticks off one acceptance area:

- [ ] `cd web && npx playwright test --config=playwright.live.config.ts tests/live/cockpit-stories/` runs locally green (needs `cargo build --features serve --release` once; tmux installed).
- [ ] `node web/tests/validate-coverage-matrix.mjs` is green from the repo root.
- [ ] CI `playwright-live` and `coverage` jobs both pass on this PR.
- [ ] Open the wizard, type a title, walk Session / Agent / Review, press Cmd/Ctrl+Enter; new session row appears in the sidebar.
- [ ] On a cockpit session, type into the composer and press Enter; the agent's response renders. Type during the turn and click Queue follow-up; after the turn ends, the queued prompt fires.
- [ ] On a tmux session on an iPhone-13-sized viewport (or DevTools mobile mode), tap each toolbar button (Arrow up, Tab, Esc, Ctrl, ^C); the pane responds.
- [ ] Make a project file change, open the session; the file appears in the diff list. Click into the file, hover a line, click "+", write a comment, save; the comments banner ticks up.
- [ ] In Settings, switch profiles; tmux status_bar set on profile A stays on A and profile B starts at the default.

## AI Usage

- [ ] No AI was used

Investigation, spec drafting, and matrix wiring drafted by Claude under my direction. I picked the scope cuts (which stories to land in this PR vs follow-up), reviewed each commit, set the Stop / queue / sub-agent acceptance criteria, and validated the harness changes (`waitForCockpitView`, `wait_ms`) before committing.

This PR adds a comprehensive, UI-driven Playwright test suite and supporting infra/docs to validate the Cockpit React UI end-to-end.

What changed (high-level, non-technical)
- Added ~83 Playwright end-to-end "user-story" specs under web/tests/live/cockpit-stories/ that drive the UI (navigation, clicks, keystrokes, mobile toolbar buttons, WebSocket/PTY interactions) and assert on rendered DOM instead of relying only on REST replay contracts.
- Added test helpers and infra hardening to reduce flakiness: waitForCockpitView(page) and enableCockpitAndWait(...) (surfaces cockpit/enable POST failures early), and safer fake-ACP timing (wait_ms clamped/capped).
- Extended test coverage tracking with one coverage entry per spec in web/tests/coverage-matrix.json and kept validate-coverage-matrix green.
- Documentation updates: AGENTS.md now requires user-story Playwright specs for new user-facing dashboard features; docs/development/playwright.md documents the cockpit-stories pattern and example usage.

Notable user-facing flows covered
- Composer: send/enter, mobile newline, streamed multi-chunk responses, Stop (thinking/tool/sub-agent), queued follow-ups (queue/edit/delete/fold), approval allow/deny, and draft persistence across reloads and session switches.
- Sidebar / Topbar: filtering, group fold/unfold, toggles, resize persistence, plan progress, and navigation.
- Wizard: open/close, Browse/Clone URL/Recent tabs, branch-from-title, worktree toggle, last-agent persistence, and launch via button or Cmd/Ctrl+Enter.
- Auth / Modals / Shortcuts: sign-out, About/Help modals, many keyboard shortcuts and command-palette flows, terminal focus chord.
- Projects / Profiles / Settings: add/remove projects, profile create/rename/delete/switch, per-profile settings isolation, and settings round-trips.
- Mobile terminal toolbar: iPhone emulation tests asserting emitted PTY/WebSocket bytes (Arrow Up, Tab, Ctrl+C, Escape) and desktop-visible behavior.
- Diff viewer: file rows, BasePicker branch suggestions, and diff hunk commenting persistence.
- Session management: delete-session variants including deleting an associated worktree.

Benefits
- Exercises full UI flows and WebSocket interactions to catch integration and reducer→render regressions that REST-only tests may miss.
- Reduces flakiness via explicit readiness checks, deterministic polling instead of fixed sleeps, and safer fake-ACP timing.
- Improves maintainability by mapping user stories to coverage entries and mandating end-to-end specs for new dashboard features.
- Provides clear examples and helpers to onboard future tests and ensure failures surface early.

Technical summary (concise)
- Helpers: web/tests/helpers/cockpit.ts adds enableCockpitAndWait(baseUrl, sessionId, timeoutMs?) and waitForCockpitView(page, timeoutMs?).
- Fake ACP: web/tests/helpers/fakeAcpAgent.mjs adds support for sessionUpdate: "wait_ms" with clamping and a 60s cap.
- Tests: ~83 new Playwright specs under web/tests/live/cockpit-stories/ using spawnAoeServe / aoe add seeding, per-spec fake ACP scripts, page.addInitScript() for WS interception, and deterministic localStorage polling/assertions. Several specs were hardened to replace fixed waits with expect.poll and stronger UI-state assertions.
- Coverage: web/tests/coverage-matrix.json gained entries for each new spec; validate-coverage-matrix.mjs remains passing.
- Docs: AGENTS.md and docs/development/playwright.md updated to mandate and document user-story Playwright specs.

Potential areas for further enhancement
- Replace any remaining fixed delays with deterministic polling where practical.
- Tag or tier slow/integration-heavy specs to optimize CI scheduling and reduce runtime/flakiness.
- Consider incremental parallelization or CI resource adjustments for long-running specs under heavy CI load.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1397?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Seluj78:

Thanks for the review pass! Pushed a follow-up commit addressing the 14 inline findings:

**Harness**
- Added `enableCockpitAndWait` to `web/tests/helpers/cockpit.ts` so the `cockpit/enable` POST is asserted with `response.ok` before polling readiness. All cockpit story specs now route through it; a 4xx/5xx surfaces as an explicit failure instead of a 15s readiness timeout. (Addresses `composer-send-enter.spec.ts` and `diff-hunk-comment.spec.ts` feedback.)
- `fakeAcpAgent.mjs` `wait_ms` now clamps non-finite / negative values and caps at 60s so bad fixture data fails loudly. (`fakeAcpAgent.mjs:168`)

**Draft persistence**
- `composer-draft-reload.spec.ts` and `composer-draft-session-switch.spec.ts` swap the fixed `waitForTimeout(400)` for an `expect.poll` against the debounced `cockpit:draft:<id>` localStorage key.

**Settings round-trips**
- `settings-logging-level.spec.ts` and `settings-theme-select.spec.ts` now require a distinct option exists before asserting persistence; a single-option select can no longer pass without a real write/read.

**Stop / focus / nav assertions**
- `shortcut-settings-key.spec.ts` adds a SettingsView-rendered assertion in addition to the URL match.
- `shortcut-terminal-focus.spec.ts` polls for `document.activeElement` inside `[data-term="paired"]` after the chord.
- `substrate-switch-to-cockpit.spec.ts` picks the seeded session by title instead of `sessions[0]`.
- `wizard-last-tool-persists.spec.ts` reopens the wizard post-reload and asserts the claude button carries the selected-class.
- `wizard-prefill-from-group.spec.ts` and 5 sibling wizard specs anchor the agent-step heading regex with `^...$`.
- `queue-follow-up-with-nav.spec.ts` bumps turn-1 `wait_ms` from 1s to 8s so the navigation cycle happens during the active turn under CI load.

**`settings-tmux-mouse.spec.ts`** was already using a hardcoded distinct value (`"disabled"` vs default `"auto"`), so no change there.

njbrake (changes requested):

The shape of the PR is right and the story-spec pattern is solid, but the latest run on `19e37062` has hard blockers. Pushed two small spec tightenings (`settings-tmux-mouse` no-op guard and `composer-mobile-enter-newline` WebSocket spy instead of `waitForTimeout(500)`) plus a merge from main; that surfaced the issues below.

**1. 34 of 135 live Playwright specs fail.** Mostly the new cockpit-stories, all from UI / data drift since the branch was started. Patterns:

- **Settings panel selectors (8+ specs):** `Default level`, `Theme`, `Mouse support`, `Status bar`, `Check interval (hours)`, `Play sounds on session status changes`, `tmux status_bar` all return "element not found". `SettingsView` markup must have shifted in main; the `locator("div").filter({ has: page.locator("label", { hasText: ... }) }).locator("select")` pattern is no longer matching. Worth replacing with stable `data-testid` hooks if the SettingsView refactor is going to keep happening.
- **Wizard heading "Name your session" missing** (`wizard-branch-from-title`, `wizard-creating-banner`, others): the wizard entry flow changed, the heading is no longer the first one rendered.
- **`profile-settings-isolation`:** asserts default profile name is `"default"`, runtime returns `"main"`. The `v008_lock_in_default_profile` migration in `#1140` flipped this.
- **`delete-session-with-worktree`:** `toBeChecked()` rejects with "Not a checkbox or radio button". The worktree-delete control is no longer a native `<input type="checkbox">`; pivot to reading `aria-checked` or `data-state`.
- **`composer-toolbar-insert`:** expects `"@/"`, gets `"@ /"`. The toolbar now inserts a space between the @ and / shortcuts.
- **`projects-add` and `projects-remove`:** strict-mode locator violation; `getByText(projectName)` matches two nodes (the row name and the path tooltip). Needs `{ exact: true }` or a narrower selector.
- **`cockpit-mode-picker`:** mode picker stays at "Default" instead of flipping to "Plan". Either the picker UX changed or the script isn't triggering `current_mode_changed`.
- **`cockpit-plan-strip`:** plan step text "Investigate the bug" missing. PlanStrip render path changed or the script's plan emission isn't reaching the reducer.
- **`diff-viewer-shows-files`, `diff-hunk-comment`:** diff text shows twice (strict-mode violation) and the comment placeholder isn't surfacing.

**2. Live Playwright job ran 387s = 6 min 27 sec.** Over the 6-minute target by 27s with the new specs in play. Even if the 34 failures were fixed, the job is past budget. Two reasonable shapes:

- Split the cockpit-stories into a separate matrix shard so it runs in parallel with the existing live specs (cuts wall time for both).
- Bump `workers` in `playwright.live.config.ts` from 4 to 6. The runner has the CPU; live specs are I/O-heavy so contention is minor.

Either keeps the suite under cap as `cockpit-stories/` grows.

**3. Unrelated:** `session::recovery::tests::recovery_lock_acquires_and_releases` panicked on the Ubuntu Cargo Test job. This is a pre-existing main-branch flake from the filesystem-lock timing, not from this PR. Will not block once #1 and #2 are addressed.

Once the 34 specs land green and the job fits the 6-min budget, this is ready.

Seluj78:

Thanks for the detailed triage. Pushed two follow-up commits that work through each bucket.

**1. 34 failing specs.** All addressed in `d1caaae8` and `3cee33c8`:

| Bucket | Fix |
|---|---|
| Settings panel selectors | Three new helpers in `web/tests/helpers/cockpit.ts` — `settingsSelectByLabel`, `settingsNumberInputByLabel`, `openSettingsTab` — walk from the `<label>` text up to its wrapping div. Theme / Tmux (Status bar + Mouse support) / Logging / Updates / Sound specs now click the right side-tab first, then resolve the control via the helper. Same change applied to `profile-settings-isolation`. |
| Wizard heading "Name your session" missing | Wizard step order is `project → session → agent → review`. `wizard-prefill-from-group` now asserts the Session heading (was wrongly asserting Agent). `wizard-recent-project` clicks Next after selecting the recent row (the click only sets `data.path`). `wizard-worktree-toggle` + `wizard-group-field` + `wizard-session-title-edit` drop the polling for-loop in favor of a single `await expect(...).toBeVisible(15s)` so a slow wizard mount doesn't skip past the assertion. `wizard-extra-repos` stays on the Project step (ExtraReposPicker mounts below the recent rows, so no back-navigation is needed). |
| `profile-settings-isolation` default profile name | Captures `await profileSelect.inputValue()` at mount instead of hardcoding `"default"` — works for both `main` and `default`. `profile-switch-view` does the same. |
| `delete-session-with-worktree` | Replaces `toBeChecked()` with `toHaveAttribute("data-checked", "true")` since `DeleteSessionDialog`'s custom `<Checkbox>` renders as a `<label data-checked="…">`, not a native input. |
| `composer-toolbar-insert` | Relaxes the assertion to `toHaveValue(/@/)` and `toHaveValue(/@.*\//)` since the popover trigger can intrude a space after the inserted character. |
| `projects-add` + `projects-remove` | `.first()` + `{ exact: true }` on the row text to dodge the strict-mode multi-match against the path tooltip. |
| `cockpit-mode-picker` | `fakeAcpAgent.mjs` `session/setMode` now emits the ACP-canonical `current_mode_update` variant alongside the legacy `current_mode_changed`. Without the canonical variant the supervisor never translates to `Event::CurrentModeChanged` so the reducer never flips `currentModeId`. |
| `cockpit-plan-strip` + `sidebar-plan-progress` | Add the mandatory `priority` field to each PlanEntry. `agent-client-protocol-schema`'s `PlanEntry` requires it; `DefaultOnError<VecSkipError>` silently drops entries that omit it, which is why "Investigate the bug" never reached the reducer. |
| `diff-viewer-shows-files` + `diff-hunk-comment` | `.first()` on the file row, and the gutter "+" click now dispatches via `evaluate(el => el.click())` so the `opacity-0 + group-hover:opacity-100` hover-only paint doesn't race the click. |
| `auth-sign-out` | Drives the LoginPage form directly. `seedAuth` is a no-op for the `servePassphrase` fixture (no `preloginViaHarness`), which is why the topbar never paints. |
| `composer-mobile-enter-newline` | Asserts the prompt is NOT dispatched (no fake-ACP response, composer not cleared) rather than asserting the literal newline survives at the caret — that's a pointer:coarse browser behavior Playwright reproduces inconsistently. |
| `shortcut-settings-key` + `shortcut-terminal-focus` + `substrate-switch-to-cockpit` + `wizard-last-tool-persists` + `wizard-prefill-from-group` | Coderabbit nits from the prior round; all picked up in `d1caaae8`. |
| Queue specs (`queue-delete-message` / `queue-edit-message` / `queue-fold-unfold` / `queue-follow-up-with-nav`) | Single composer locator whose regex matches both placeholder variants (`Send a message|Queue a follow-up`) so filling the follow-up doesn't race the placeholder swap when `turnActive` flips mid-test. |

**2. Live Playwright over the 6-minute budget.** Pushed `4f1...` (workers bump) — `web/playwright.live.config.ts` now runs 6 workers on the GitHub runner. The suite is I/O bound (tmux + fetch + WS round-trips), CPU contention is minor, and `aoeServe.ts`'s port allocation (`5200 + workerIndex*100 + parallelIndex`) already supports six workers cleanly without collisions. Sharding into a separate matrix is a heavier lift; happy to do that as a follow-up if 6 workers isn't enough headroom once cockpit-stories/ keeps growing.

**3. `recovery_lock_acquires_and_releases`** Acknowledged as a pre-existing main flake, untouched by this PR.

Ready for the next CI run.

Seluj78:

Pushed `2bc...` addressing the round-2 CodeRabbit pass:

- 31 specs now resolve their target session by seeded title (`sessions.find(s => s.title === "...")`) instead of `sessions[0]!.id`; closes the order-flakiness flag across cockpit-mode-picker, composer-stop-*, mobile-toolbar-*, queue-*, approval-*, sidebar-row-click-navigate, topbar-go-to-dashboard, and the rest.
- `composer-draft-session-switch` and `queue-follow-up-with-nav` migrate the remaining raw `cockpit/enable` POSTs to `enableCockpitAndWait`, so the response is asserted before readiness polling.
- `composer-queue-follow-up` bumps turn-1 `wait_ms` from 600ms to 2s.
- `diff-hunk-comment` hovers the diff line ancestor first so the `opacity-0 group-hover:opacity-100` "+" gutter button paints before the click.
- `projects-add` seedFn now throws on `git init` / `git commit` non-zero exits.
- `sidebar-plan-progress` asserts the `cockpit/prompt` POST returned 2xx.
- `sidebar-resize-persist` + `right-panel-resize-persist` capture pre-drag localStorage and assert the value changed after drag-release.

njbrake:

Approved the workflows on `b74e767c`. CI itself is green; Tests / Playwright (live) is red with **14 of 134 specs failing**, all in `cockpit-stories/`. Full report attached so you can work through it before the next push.

### Failure breakdown

| Bucket | Specs | Symptom |
|---|---|---|
| Turn-active UI never paints | `approval-allow`, `approval-deny`, `cockpit-plan-strip`, `composer-queue-follow-up`, `composer-stop-mid-turn`, `queue-follow-up-with-nav`, `queue-edit-message` | Awaited cockpit text (`Approval needed`, `Stop`, scripted chunks like `First turn.`, `Investigate the bug`) never visible inside the 10s/15s window. Looks like the fakeAcpAgent script isn't reaching the cockpit reducer for these flows. |
| Selector / contract drift | `cockpit-mode-picker` (button stays `"Default"`, no `Plan` flip after `session/setMode`), `diff-hunk-comment` (comment composer never paints after gutter `+` click), `wizard-last-tool-persists` (`localStorage["aoe-wizard-last-tool"]` is `null` after Launch) | UI shape or storage key differs from what the spec asserts. |
| Profile / queue fixture timing | `profile-settings-isolation` (initial `profileSelect.inputValue()` is `""`), `profile-delete`, `queue-delete-message`, `queue-fold-unfold` (all `locator.click` 60s timeouts) | First-render race or selector mismatch on the profile / queue UI. |

### Likely shared root cause for the dominant bucket

7 of the 14 failures are scripted-update specs where the fake agent's first chunk never reaches the React surface. That's the same shape as the `cockpit-mode-picker` failure round 1 fixed by emitting both `current_mode_changed` and `current_mode_update`, but now it's the upstream channel (the prompt itself) that isn't being delivered.

Worth checking before fixing each spec individually:

1. Does `seedSessionViaAoeAdd` wire `claude` / `claude-agent-acp` / `aoe-agent` on `$PATH` to the fakeAcpAgent shim for the seeded session, or only for the `serveCockpit` fixture? The CI failures cluster on specs that use `spawnAoeServe({ cockpit: true, fakeAcpScript: ..., seedFn: seedSessionViaAoeAdd(...) })` rather than the fixture. If the seedFn-installed `aoe add` resolves `claude` from a different `bin/` than the supervisor reads from, the supervisor falls through to the system adapter and silently never emits anything.
2. `enableCockpitAndWait` proves the `initialize` handshake completed (replay has at least one frame). It does not prove the subsequent `session/prompt` lands on the fake. If the supervisor is hung on a stale `session/new`, the first `cockpit/prompt` POST returns 2xx but the prompt is dropped. Worth asserting against `/cockpit/replay` after the prompt to confirm the script's first chunk shows up server-side before driving any UI.
3. Six workers on the 4-core runner: the dominant failure mode (text never appears in 10s) looks like CPU starvation rather than missing wiring on at least some of these. The 1ms inter-update tick in `emitSessionUpdates` may not survive contention. Bumping the inter-update tick to 5ms and / or dropping back to 4 workers when running locally + CI may surface whether this is wiring vs starvation.

### Selector drift bucket

- `wizard-last-tool-persists`: confirm the wizard still writes `aoe-wizard-last-tool` on Launch and that the storage write happens before navigation, not after. If the write moved into a navigation effect, the spec's reload races it.
- `diff-hunk-comment`: the hover-then-click landed in round 2, but the composer never paints. Check whether `CommentForm` now opens via a different placeholder string or whether the gutter `+` button writes to a different DOM container.
- `cockpit-mode-picker`: the previous round added the canonical `current_mode_update` variant; CI says the reducer still doesn't flip. Either the variant name changed again in main or the supervisor isn't translating.

### Profile / queue timing bucket

`profile-settings-isolation` capturing `""` at mount suggests the `ProfileSelect` renders before its initial fetch resolves. Poll `profileSelect.inputValue()` until non-empty rather than capturing it immediately. `profile-delete` and the queue specs hitting 60s click timeouts probably need their target locators re-anchored after the round-2 changes.

### Hold for next round

Holding off on approving the review until these are green. Once you push the next pass, no need to re-tag; CI approval will reissue.

Full failing traces live in the `playwright-live-report` artifact on the failed Tests run; if any of the buckets above are wrong about root cause, the traces will say so before fix attempts.

njbrake (approved):

Approving. This closes the test-debt gap from #1383 in the right way: UI-driven specs that drive React end-to-end and assert on rendered DOM, complementing (not replacing) the REST-contract tracers under `web/tests/live/cockpit-*.spec.ts`.

What I checked beyond the test bodies themselves:

- The `session/setMode` to `session/set_mode` rename plus the `current_mode_changed` to `current_mode_update` update tag in `fakeAcpAgent.mjs` match what production deserializes (`SessionUpdate::CurrentModeUpdate` at `src/cockpit/acp_client.rs:2659`), so the fake is now closer to wire-correct rather than further from it.
- Modeling `session/cancel` as a notification with a per-session flag (sliced 50ms inside `wait_ms` so cancels land fast) is the correct ACP shape.
- The `tmpdir()` to `/tmp` switch for the HOME base is the right fix for Darwin's 104 byte `sun_path` limit; the runner-socket deadline knob is debug-only and release builds are unchanged.
- Two-phase `waitForCockpitReady` plus the `debug.log` tail dump on failure is a meaningful debugability win for the whole live suite, not just the new stories.
- Coverage matrix: 83 new entries, all required fields present, validator green on the PR branch.

CodeRabbit posted 29 line-level findings; 28 are explicitly marked addressed and spot checks confirm the title-based lookups, `enableCockpitAndWait` adoption, and select-by-label patterns landed. The one remaining (`right-panel-resize-persist` localStorage poll) and a handful of doc and whitespace nits I pushed as `2c2bd11b`:

- Stale `current_mode_changed` references in `fakeAcpAgent.mjs` header and `cockpit-mode-picker.spec.ts` swapped to `current_mode_update`.
- Stray space in `waitForCockpitView , enableCockpitAndWait` imports removed across 17 spec files.
- `right-panel-resize-persist.spec.ts` post-drag read wrapped in `expect.poll` so the React-batched mouseup writer cannot race the assertion.

Nice work, Jules.

Seluj78:
Closing to split the PR in logical and manageable chunks (and mergeable)

PR #1399, opened 2026-05-22
fix: let directory browser load more entries

MovieHolic-Plex:
## Description
Fixes the web directory browser's hard 100-entry ceiling by adding an explicit "Load 100 more" control when the filesystem browse API reports more entries.

Previously, directories with more than 100 visible subfolders showed "Showing first 100 entries. Use the filter to narrow results.", but filtering only searched the already-loaded first page. Entries after the first 100 could not be selected from the web UI.

The browser now:
- loads the first 100 entries as before
- shows the current loaded count when more entries exist
- lets users request the next page by increasing the browse limit
- keeps navigation resets at the first page for each new directory

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:**

Codex

Used to implement and validate a focused web directory-browser pagination fix.

**NOTE:**

## Validation

- `npx eslint src/components/DirectoryBrowser.tsx tests/wizard-project-step.spec.ts`
- `npx playwright test wizard-project-step.spec.ts`
- `npm run build`
- `git diff --check`
- commit hook: `cargo clippy -- -D warnings`, `cargo fmt -- --check`

## Overview

Removes the web directory browser's hard 100-entry ceiling by adding explicit paging and server-side filtering so users can find and select entries beyond the first page.

## What changed (high level)

- The directory browser still loads the first page initially, and now shows how many entries are loaded with a "Load 100 more" control to fetch additional pages.
- Filtering is performed server-side so search queries can match entries not yet loaded on the client.
- Navigation initialization improved: tries initialPath → last-used directory → home.
- Async navigation is hardened so slow/stale responses can't overwrite newer results, and loading/errors no longer leave the UI stuck.
- Tests added/updated (unit and Playwright) and coverage updated for DirectoryBrowser.

## What moved / was added / was removed

- Added BROWSE_PAGE_SIZE constant and entryLimit state; replaced hard-coded "100" with the constant.
- Replaced single-page fetch logic with a new loadPath(path, limit, options) loader that manages entries, hasMore, loading, error, and filter state.
- Added loadMore() to increase the loaded limit and request the next page.
- Debounce/filter logic refined to avoid regressions when loading more pages and to prevent unnecessary reloads.
- Server: browse API accepts an optional filter and applies case-insensitive substring matching.
- Client API: browseFilesystem signature extended to accept an optional filter parameter.
- Unit and Playwright tests added to cover pagination, server-filtered search, stale-response handling, and startup fallback behavior.
- No breaking changes to exported component signatures.

## Technical notes

- Uses a numeric loadSeq token to ignore out-of-order async responses; setLoading(false) is called only for the latest request.
- Mount initialization uses a microtask sequence to select the first available path (initial → last-used → home).
- Filter is trimmed/lowercased before sending to the server; server-side filtering skips entries that don't contain the normalized filter substring.
- Debounce race conditions addressed with a clearable debounce handle and checks that the requested filter differs from the currently loaded filter.

## Benefits

- Users can load and select directories beyond the previous 100-entry limit without client-side workarounds.
- Searches can find matches anywhere in a large directory without loading all pages.
- Clearer UI feedback about loaded entries and more reliable navigation behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1399?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/server/api/system.rs
+++ b/src/server/api/system.rs
@@ -354,6 +354,7 @@ pub async fn list_profiles(State(state): State<Arc<AppState>>) -> Json<Vec<Profi
 pub struct BrowseQuery {
     pub path: String,
     pub limit: Option<usize>,
+    pub filter: Option<String>,
 }
 
 #[derive(Serialize)]
@@ -390,6 +391,7 @@ pub async fn browse_filesystem(
 ) -> impl IntoResponse {
     let result = tokio::task::spawn_blocking(move || {
         let limit = query.limit.unwrap_or(100);
+        let filter = query.filter.map(|f| f.trim().to_lowercase());
         let path = std::path::Path::new(&query.path);
         let canonical = path.canonicalize().map_err(|_| "Path does not exist")?;
 
@@ -416,6 +418,11 @@ pub async fn browse_filesystem(
             let is_dir = entry_path.is_dir();
             if !is_dir {
                 continue;
+            }
+            if let Some(filter) = filter.as_deref() {
+                if !filter.is_empty() && !name.to_lowercase().contains(filter) {
+                    continue;
+                }
             }
             let is_git_repo = entry_path.join(".git").exists();
             entries.push(DirEntry {
--- a/web/src/components/DirectoryBrowser.tsx
+++ b/web/src/components/DirectoryBrowser.tsx
@@ -4,6 +4,7 @@ import { safeGetItem, safeSetItem } from "../lib/safeStorage";
 import type { DirEntry } from "../lib/types";
 
 const LAST_DIR_KEY = "aoe-last-browse-dir";
+const BROWSE_PAGE_SIZE = 100;
 
 function loadLastDir(): string | null {
   return safeGetItem(LAST_DIR_KEY);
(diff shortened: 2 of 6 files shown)


Seluj78:
Hey, thanks for the work ! Make sure the setting you're adding is available through the TUI, web ui and toml :)

MovieHolic-Plex:
Updated this PR with the reviewer fixes plus one additional issue found during manual repro:

- Guarded `DirectoryBrowser` loads with a latest-request token so stale slower responses cannot overwrite newer navigation.
- Let startup fall through from a bad `initialPath` to last-used directory / home.
- Added server-side filtering to `/api/filesystem/browse`, so entries beyond the first 100, such as `z-project`, are searchable without loading every page first.
- Added Vitest coverage for fallback, stale-response handling, and server-filtered search.
- Added Playwright coverage for both `Load 100 more` and filtering a `z-project` entry beyond the first page.

On the TUI/web/TOML setting comment: this PR does not add a user setting. It changes the existing web directory browser behavior and browse API query handling only, so there is no new setting to expose in TUI, web settings, or TOML.

Validation run:

- `npm run test:unit -- src/components/__tests__/DirectoryBrowser.test.tsx`
- `npx eslint src/components/DirectoryBrowser.tsx src/components/__tests__/DirectoryBrowser.test.tsx tests/wizard-project-step.spec.ts`
- `npm run build`
- `npx playwright test wizard-project-step.spec.ts`
- `node tests/validate-coverage-matrix.mjs && git diff --check`
- `cargo fmt --check`

MovieHolic-Plex:
Addressed the latest CodeRabbit pass:

- Wrapped `browseFilesystem` in `try/catch/finally` inside `loadPath` so rejected requests cannot leave the browser stuck in loading state.
- Preserved the latest-request guard while moving `setLoading(false)` into the guarded `finally` branch.
- Updated the Load more button to use dashboard UI typography/height and `BROWSE_PAGE_SIZE` in the label instead of a hard-coded `100`.
- Tightened the stale-response unit test so it fails if the slow request resolver is never captured.
- Fixed a regression found during validation where the filter debounce could reload the first page after Load more; the component now only sends a filter request when the normalized filter value actually changes.

Validation run:

- `npm run test:unit -- src/components/__tests__/DirectoryBrowser.test.tsx`
- `npx eslint src/components/DirectoryBrowser.tsx src/components/__tests__/DirectoryBrowser.test.tsx tests/wizard-project-step.spec.ts`
- `npm run build`
- `npx playwright test wizard-project-step.spec.ts`
- `node tests/validate-coverage-matrix.mjs && git diff --check`
- `cargo fmt --check`

MovieHolic-Plex:
Addressed the debounce race from the latest CodeRabbit pass:

- Added a `filterDebounceHandle` ref and `clearFilterDebounce()` helper.
- `loadMore()` now clears any pending filter debounce before requesting the expanded limit.
- Reset navigation also clears pending filter debounce before clearing the filter.
- The scheduled debounce callback now re-checks `nextFilter` against `loadedFilter.current` before calling `loadPath`, so a timer that survives until after another load completes cannot clobber the expanded listing.

Validation run:

- `npm run test:unit -- src/components/__tests__/DirectoryBrowser.test.tsx`
- `npx eslint src/components/DirectoryBrowser.tsx src/components/__tests__/DirectoryBrowser.test.tsx tests/wizard-project-step.spec.ts`
- `npm run build`
- `npx playwright test wizard-project-step.spec.ts`
- `node tests/validate-coverage-matrix.mjs && git diff --check`
- `cargo fmt --check`

njbrake (approved):

Approving. Real bug (folders past entry 100 were unreachable), tight fix. The `loadSeq` stale-response guard, try/catch/finally error handling, debounce + load-more coordination, and the conditional startup fallback chain are all correctly done. Vitest + Playwright cover the meaningful paths, and the coverage matrix is updated.

One thing worth doing before landing: re-trigger CI on the final commit. Only `check-template` and `close-stale` ran on `3868be8`; the Rust and web suites did not pick up the final patch.

Optional follow-ups:
- Add one Playwright assertion for the combined "filter set, click Load More" path (the code forwards the filter through `loadMore`, just not tested directly).
- Stylistic: `return clearFilterDebounce;` in the filter `useEffect` works but `return () => clearFilterDebounce();` reads as the conventional cleanup shape.

PR #1400, opened 2026-05-22
fix(web): bump @assistant-ui to pick up tap out-of-bounds fix

### LINKED ISSUE #1119: Cockpit: blank-screen crash on "tapClientLookup: Index 3 out of bounds (length: 3)" from @assistant-ui/tap

## Problem

The web cockpit occasionally crashes into a blank screen. The browser console shows:

```
Uncaught Error: tapClientLookup: Index 3 out of bounds (length: 3)
    get http://127.0.0.1:8080/assets/CockpitView-BYKvVa4Z.js:1
    message
    t
    get
    e
    t
    r
    Vo http://127.0.0.1:8080/assets/index-Coc36Z0z.js:9
    useSyncExternalStore http://127.0.0.1:8080/assets/index-Coc36Z0z.js:2
    R http://127.0.0.1:8080/assets/CockpitView-BYKvVa4Z.js:1
    hc
    _c
    wo
    sc
    Pc
    Nu Au ku _u ld id od ud
    ...
    notifySubscribers
    runTask pt mt
```

Reproduced on session `d6dcb681bbf649d5` (the session with a Plan agent / Task subagent). The error fires from `@assistant-ui/tap`'s subscriber lookup ("Index 3 out of bounds (length: 3)" is a classic off-by-one: asking for index 3 of an array whose valid indices are 0-2). Reaches React's `useSyncExternalStore` and unwinds through the cockpit component tree, taking the view down to a blank screen.

## What's there today

The cockpit renders through `@assistant-ui/react`'s external-store runtime:

`web/src/components/cockpit/CockpitRuntime.tsx:29-78`:

```tsx
import {
  AssistantRuntimeProvider,
  useExternalStoreRuntime,
  type ThreadMessageLike,
} from "@assistant-ui/react";
// ...
const cockpit = useCockpit(sessionId, cockpitWorkerState);
const messages = useMemo(
  () => activityToThreadMessages(cockpit.state.activity, cockpit.state.turnActive),
  // ...
);
```

`@assistant-ui/tap` is a transitive dep of `@assistant-ui/react`. It implements assistant-ui's internal reactive store used by `useExternalStoreRuntime`. `tapClientLookup` is one of its subscriber-lookup helpers; the error string is internal.

Package versions from `web/package-lock.json`:

```
"@assistant-ui/react": "^0.12.28",
"@assistant-ui/core": "^0.1.17",
"@assistant-ui/store": "^0.2.9",
"@assistant-ui/tap": "^0.5.10",
```

Aoe-side state plumbing that feeds the tap store every event:

- `web/src/hooks/useCockpit.ts:51` — `stateCache` Map preserves `CockpitState` across session switches.
- `web/src/hooks/useCockpit.ts:288-291` — `hydrate` action fires on every session change, replacing the entire reducer state with a cached or empty `CockpitState`.
- `web/src/lib/cockpitTypes.ts:710-716` — `pushActivity` returns a fresh array every dispatch (and slices to `ACTIVITY_LIMIT = 200`, per #1111).
- `web/src/components/cockpit/CockpitRuntime.tsx:85-87` — `messages` rebuilds whenever `activity` or `turnActive` change.

Every event from the WS produces a new `activity` array, which produces a new `messages` array, which assistant-ui's external store consumes and re-dispatches to its tap subscribers. A subscriber list growing or shrinking mid-dispatch is the kind of state that would produce an off-by-one in a tap-style reactive lookup.

## Root cause

Best-guess (without a clean repro): a race in `@assistant-ui/tap`'s subscriber bookkeeping triggered by aoe's high-frequency state churn. Specifically:

1. Tap maintains an indexed list of subscribers per dependency.
2. During `notifySubscribers` (visible in the stack), the list length is captured as a snapshot.
3. A nested dispatch unsubscribes one entry (or the list resets entirely) before the loop iterates to the captured length.
4. The next iteration reads `subscribers[i]` where `i === length`, producing the out-of-bounds error.

aoe's pattern of rebuilding `messages` on every event (and `hydrate` wholesale-replacing state on session switch) is a stress test for the subscriber-tracking code. Pretty likely the bug is upstream; aoe's pattern surfaces it.

Two suspect aoe-side triggers:

1. **Session switch mid-event**: the user switches sessions while a WS frame is being applied. `useCockpit.ts:278-291`'s effect tears down the old subscriber set and `dispatch hydrate` immediately. If a tap subscriber from the prior session is still in the queue when this happens, the lookup index can outrun the new shorter list.
2. **Rapid frame application during reconnect**: on WS reconnect with `since=N`, the server replays multiple events back-to-back. Each one mutates `activity`; the messages rebuild fires that many times within a microtask. If tap batches subscribers per microtask, the bookkeeping mismatch becomes more likely.

The trace's `notifySubscribers → l → l → st → t → runTask → pt → mt → <anonymous>` chain is consistent with assistant-ui scheduling a microtask drain of pending notifications, which is where this kind of race usually lives.

### A. Capture a reliable repro

The crash is rare and the error string is vendor-internal. Filing upstream needs a reduced-case repro. Two steps:

1. Add a top-level React error boundary around `CockpitRuntime` that captures the error, the rendered-message snapshot (or at least its length), and the last few events from `state.activity`. Surface via the existing aoe debug log surface (#1009 logging discipline). Renders a "Cockpit crashed (caught error: ...) Reload" fallback instead of a blank screen.
2. Bisect locally: try to repro under React strict mode, under session-switch-during-active-turn, and under rapid reconnect replay. Note which path fires it.

### B. Track the upstream fix

Already filed upstream: **assistant-ui/assistant-ui#3652** (open). Same `tapClientLookup` out-of-bounds, different indices, same root-cause analysis (subscribers from prior thread notified before unmount, hit bounds check on the new shorter array). Two related closed issues with the same error string: #3994 and #3968.

aoe-side data point added as a comment on #3652 with our specific repro (subagent-heavy session, session-switch and reconnect-replay triggers, full stack).

Until upstream lands, the error boundary from A/C keeps the UI usable.

### C. Defensive: error boundary alone (independent of root cause)

Even if the upstream fix lands, aoe should not blank-screen on any uncaught render error. The error boundary from A is worth landing as standalone polish:

```tsx
class CockpitErrorBoundary extends React.Component {
  componentDidCatch(error, info) {
    // log to debug surface; show fallback
  }
  render() { ... }
}
```

Wrap the cockpit subtree in `App.tsx` or `CockpitView`. Any future crash (this one or another) shows a recovery affordance instead of going dark. No backend change.

Default to C immediately as polish; A and B are the right path for the actual fix.

## Why now

- Blank-screen crash is the worst possible UX failure: no message, no recovery, no signal to the user that anything happened. They reload, sometimes lose draft context.
- C is a small standalone PR; ~30 LOC error boundary + log call.
- A unblocks the upstream report and turns "intermittent crash" into "actionable bug."

## Open questions

- What's the exact trigger? Strong candidates listed in Root cause. Need the error boundary's snapshot data to confirm.
- Is the crash session-dependent? User reported it on `d6dcb681bbf649d5`, which has a heavy subagent + many tool calls. Other reports would help confirm whether the trigger correlates with subagent / approval / mid-turn-restart shapes.
- Does an `@assistant-ui/tap` upgrade help? Latest published version unknown without checking npm. Worth bumping as a low-cost first try.
- Should the error boundary auto-reload after N seconds, or require explicit user action? Auto-reload risks loops if the crash is deterministic; lean on explicit user click.

## Out of scope

- Reworking aoe's state model to dispatch fewer updates. The frequent rebuilds are inherent to streaming-event UI; the fix belongs in the subscriber bookkeeping, not in aoe.
- Replacing `@assistant-ui/react`. Out of proportion for an intermittent vendor bug.
- Adding telemetry around the crash. Local debug log from C is enough.

## Cross-references

- #1009 — earlier round of "WS disconnects + UI breaks" work. This is a different kind of breakage (renderer crash, not WS death) but argues for the same instrumentation-first discipline.
- #1041 — subagent rendering (closed). Reported session triggers a subagent render path; could be a contributing trigger.
- #1111 — frontend `ACTIVITY_LIMIT=200` evicts older rows. Independent issue but tightens the array churn that feeds tap.

Seluj78:

## Description

Picks up the upstream fix for the `tapClientLookup: Index N out of bounds (length: N)` crash that occasionally blanked the cockpit (#1119).

Bumps:

- `@assistant-ui/react` `^0.12.28` → `^0.14.7`
- `@assistant-ui/react-markdown` `^0.12.11` → `^0.14.0`

The matching transitives lift to `@assistant-ui/core@0.2.4`, `@assistant-ui/store@0.2.11`, `@assistant-ui/tap@0.5.11`.

Upstream landed two PRs to close the zombie-fiber pattern behind the crash:

- assistant-ui/assistant-ui#4069: derived scopes re-keying on meta change (closed the actual throw path).
- assistant-ui/assistant-ui#4077: keyed part fibers by absolute index (closed the underlying reuse-across-reshape pattern).

Upstream tracking issue `assistant-ui/assistant-ui#4051` (filed from #1119) was closed with both PRs in.

No AoE-side code changes were needed: `tsc -b` clean, the full vitest suite passes (777/777), `vite build` succeeds.

Closes #1119

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open the cockpit on a session with a long assistant message containing many parts (text + tools + reasoning), switch to another session in the sidebar, then switch back. Cockpit renders without throwing `tapClientLookup: Index N out of bounds` and without going blank.
- [ ] Repeat the switch a few times on a subagent-heavy session (the original repro from #1119, session `d6dcb681bbf649d5` shape). No blank screen, no console error.
- [ ] Trigger a WS reconnect during an active turn (kill the daemon, let it come back). Replayed events apply without the out-of-bounds throw.
- [ ] Normal cockpit flows still work end-to-end: send a prompt, watch tool cards render, approve / reject an approval, `/clear`, `/compact`.

## Checklist

## AI Usage

- [ ] No AI was used

Investigation, plan, and prose drafted by Claude under my direction. The actual change is two version strings plus the regenerated lockfile; I made the version-pick decisions and verified the build / tests pass locally.

**NOTE:**

## Overview

This pull request updates the `@assistant-ui` library dependencies to versions that include an upstream fix for an intermittent cockpit crash. The crash manifested as a blank screen with a "tapClientLookup: Index out of bounds" error in the browser console, caused by a race condition in the subscriber bookkeeping logic of the `@assistant-ui/tap` module.

## What Changed

**Dependencies Updated:**
- `@assistant-ui/react`: bumped from `^0.12.28` to `^0.14.7`
- `@assistant-ui/react-markdown`: bumped from `^0.12.11` to `^0.14.0`

These changes pull in transitive updates to `@assistant-ui/core@0.2.4`, `@assistant-ui/store@0.2.11`, and `@assistant-ui/tap@0.5.11` from upstream.

**No application code changes.** Only `web/package.json` was modified (2 lines added, 2 removed).

## Benefits

- **Fixes intermittent crash:** Resolves a race condition in subscriber management that could blank the cockpit during rapid message rebuilds, session switches, or WebSocket reconnections.
- **Zero risk upgrade:** No application code changes required; the fix is purely upstream dependency upgrades.
- **Validated:** All tests pass (777/777), TypeScript compilation is clean, and the build succeeds.

## Technical Details

The upstream fix addresses the derived-scope rekeying and keyed fiber tracking in `@assistant-ui/tap`'s `notifySubscribers` function, which was iterating over a subscriber list that could change (due to unsubscribes or resets during notification), resulting in out-of-bounds index access. The upstream pull requests (assistant-ui/assistant-ui#4069 and `#4077`) contain the detailed fixes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1400?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -8,8 +8,8 @@
       "name": "web",
       "version": "0.0.0",
       "dependencies": {
-        "@assistant-ui/react": "^0.12.28",
-        "@assistant-ui/react-markdown": "^0.12.11",
+        "@assistant-ui/react": "^0.14.7",
+        "@assistant-ui/react-markdown": "^0.14.0",
         "@dnd-kit/core": "^6.3.1",
         "@dnd-kit/sortable": "^10.0.0",
         "@dnd-kit/utilities": "^3.2.2",
@@ -108,19 +108,19 @@
       "license": "MIT"
     },
     "node_modules/@assistant-ui/core": {
-      "version": "0.1.17",
-      "resolved": "https://registry.npmjs.org/@assistant-ui/core/-/core-0.1.17.tgz",
-      "integrity": "sha512-IWIP98UVQ9W+oF0yz8XqFRtaX8HtozWVUWt6D/BSV6cyKwLfJ8niHtLG74bSnllTnGcreU2El3GR/tIodR1XuA==",
+      "version": "0.2.4",
+      "resolved": "https://registry.npmjs.org/@assistant-ui/core/-/core-0.2.4.tgz",
+      "integrity": "sha512-ct3FyWfo5iIA6y0AXg41oLGRJRGTY4Z8UcYnimQZPyRfpsxI0QggjfGzL/MploSJuOb4p+jE3XZIg2wB6p1rjw==",
       "license": "MIT",
       "dependencies": {
-        "assistant-stream": "^0.3.12",
-        "nanoid": "^5.1.9"
+        "assistant-stream": "^0.3.15",
+        "nanoid": "^5.1.11"
       },
       "peerDependencies": {
-        "@assistant-ui/store": "^0.2.9",
-        "@assistant-ui/tap": "^0.5.10",
+        "@assistant-ui/store": "^0.2.11",
+        "@assistant-ui/tap": "^0.5.11",
         "@types/react": "*",
-        "assistant-cloud": "^0.1.27",
+        "assistant-cloud": "^0.1.28",
         "react": "^18 || ^19",
         "zustand": "^5.0.11"
       },
@@ -140,27 +140,28 @@
       }
     },
     "node_modules/@assistant-ui/react": {
-      "version": "0.12.28",
-      "resolved": "https://registry.npmjs.org/@assistant-ui/react/-/react-0.12.28.tgz",
-      "integrity": "sha512-czjpexLK1lKnNDNM1YMJi8SufeKUWBICqiVUtiHMV+86PYGCwJykOZKkchI8MVbSQ62xZ8A1LfPO5W2IDjed3A==",
+      "version": "0.14.7",
+      "resolved": "https://registry.npmjs.org/@assistant-ui/react/-/react-0.14.7.tgz",
+      "integrity": "sha512-GiEF4tYyU9uuBxUhGxy65kw54mtHS/L5zC6PyJpVUUFyGTjzprTqaFGZLhQuVFwJWqqGIzuHotYrfKZRgU9cCQ==",
       "license": "MIT",
       "dependencies": {
-        "@assistant-ui/core": "^0.1.17",
-        "@assistant-ui/store": "^0.2.9",
-        "@assistant-ui/tap": "^0.5.10",
+        "@assistant-ui/core": "^0.2.4",
+        "@assistant-ui/store": "^0.2.11",
+        "@assistant-ui/tap": "^0.5.11",
         "@radix-ui/primitive": "^1.1.3",
         "@radix-ui/react-compose-refs": "^1.1.2",
         "@radix-ui/react-context": "^1.1.3",
         "@radix-ui/react-primitive": "^2.1.4",
         "@radix-ui/react-use-callback-ref": "^1.1.1",
         "@radix-ui/react-use-escape-keydown": "^1.1.1",
-        "assistant-cloud": "^0.1.27",
-        "assistant-stream": "^0.3.12",
-        "nanoid": "^5.1.9",
+        "assistant-cloud": "^0.1.28",
+        "assistant-stream": "^0.3.15",
+        "nanoid": "^5.1.11",
         "radix-ui": "^1.4.3",
         "react-textarea-autosize": "^8.5.9",
-        "zod": "^4.3.6",
-        "zustand": "^5.0.12"
+        "safe-content-frame": "^0.0.19",
+        "zod": "^4.4.3",
+        "zustand": "^5.0.13"
       },
       "peerDependencies": {
         "@types/react": "*",
@@ -178,9 +179,9 @@
       }
     },
     "node_modules/@assistant-ui/react-markdown": {
-      "version": "0.12.11",
-      "resolved": "https://registry.npmjs.org/@assistant-ui/react-markdown/-/react-markdown-0.12.11.tgz",
-      "integrity": "sha512-gYu4XVI2lX3lp9UG7V5VWP1+eO7SZomiBKsAZOKUOeuwn/hoL+J0vFY52FUgJixdF2R8NPPto2lb98DmJE70lA==",
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@assistant-ui/react-markdown/-/react-markdown-0.14.0.tgz",
+      "integrity": "sha512-G29cz4UHnQFyxTXCHsloMfyqHSuKSiPhkaAa069tWdf/Px3ZKCyhcP0TENCnsjqsw2iTFMVm8/Nqa0d5OwsdYw==",
       "license": "MIT",
       "dependencies": {
         "@radix-ui/react-primitive": "^2.1.4",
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Approving. Picks up the confirmed upstream fix for #1119 at the minimum version that includes both upstream PRs (4069 in `core@0.2.3` / `react@0.14.6` and 4077 in `core@0.2.4` / `react@0.14.7`). Static import-surface check is clean: every name imported from `@assistant-ui/react`, `@assistant-ui/react-markdown`, and `@assistant-ui/core` still exists in the new versions, and the breaking-change window (only `0.14.0` deprecation cleanup, since `0.13` was skipped) does not touch any API we use. New transitive `safe-content-frame` is MIT, around 26KB, and tree-shaken since we do not render MCP apps.

Worth re-running the full web CI on the final commit before landing; only `check-template` ran, not the build / vitest / Playwright suites. Separately, `Composer.tsx` imports from `@assistant-ui/core` as a phantom dep; from `react@0.14.6` upstream began re-exporting those three names directly, so a small follow-up could drop the implicit core import.

PR #1420, opened 2026-05-22
fix(ci): switch nix-npm-hash bot to PR flow + validate hash on PRs

jerome-benoit:
## Description

The `nix-npm-hash` bot is structurally broken: every run on a `web/package-lock.json` change is rejected with `GH013: Repository rule violations` because the main branch ruleset enforces "Require a pull request before merging" plus 14 required status checks, and a `GITHUB_TOKEN` push from the workflow cannot satisfy them. Verified in run logs [`26161177808`](https://github.com/njbrake/agent-of-empires/actions/runs/26161177808) (commit `45f280d`) and [`26287007783`](https://github.com/njbrake/agent-of-empires/actions/runs/26287007783) (commit `5e001fe`). Each failure leaves `flake.nix` stale on `main`, breaks `cachix-push.yml`, and forces a maintainer to push the hash manually.

This PR closes the desync class structurally and fixes the current stale state in one shot, no transition window.

### What

- **Mirror `open-release-pr.yml`'s PR pattern** (post-merge bot, rewrite of `nix-npm-hash.yml`). The bot pushes to a fixed singleton branch `chore/nix-npm-hash-update` using `RELEASE_TOKEN` (the PAT already proven by `open-release-pr.yml` and `tag-release-pr.yml`), opens or refreshes one PR via `gh pr create`/`gh pr edit`, and enables auto-merge (`gh pr merge --auto --squash`). The PR satisfies the 14 required checks naturally and merges without any bypass.
- **New PR auto-fix bot (`nix-npm-hash-fix-pr.yml`).** When a contributor PR bumps `web/package-lock.json` without bumping `flake.nix`, this workflow computes the matching hash via `prefetch-npm-deps` (lockfile-only, no `npm install`, no postinstall execution) and either pushes a `flake.nix` fix to same-repo PR branches or upserts a sticky comment with a copy-paste snippet on fork PRs. `pull_request_target` is used safely: workflow runs from base, PR head is read as data only via `gh api`, no PR-supplied code is evaluated, all PR-controlled values are passed via `env:` (never interpolated into `run:`).
- **Defense in depth: `Nix Build Web` PR-CI job.** Runs `nix build .#aoe-with-web` on every PR that touches `web/package-lock.json`, `flake.nix`, or `flake.lock`, so a stale hash cannot land on `main` even if the bot itself is offline. Hash mismatch fails the job loudly with the exact `got: sha256-...` value the contributor needs.
- **Shared shell scripts (`.github/scripts/{get,set}-flake-npm-hash.sh`).** Single source of truth for the anchored regex + match-count assertion + hash format validation. Used by both bots so the regex cannot drift.
- **Both bots use `prefetch-npm-deps`** for symmetric semantics. The post-merge bot dropped its previous `nix build` + fake-hash + `got:`-extraction trick (~50 lines of complexity); cachix-push.yml continues to do the full build validation post-merge.
- **Drop the `workflow_run` chain in `cachix-push.yml`.** The bot's PR merges as a regular `flake.nix` change; the existing `paths: [flake.nix]` push trigger picks it up.
- **Bundle the immediate hash bump** (`flake.nix:78` to `sha256-RTL/3DccMZODLr9lMZdQfoMFtK2ZsARFMABh4SUV4xI=`, captured from failed run `26287007783`) so this PR's own `Nix Build Web` job is end-to-end proof the validator works.

### Why RELEASE_TOKEN, not a new bypass actor

Routing through a PR with auto-merge keeps the bot's commit subject to the same 14 checks as any human PR. No bypass list edit, no new GitHub App, no widening of secret blast radius beyond what `open-release-pr.yml` already accepts.

### Edge cases

| Case | Behavior |
|---|---|
| Two `web/package-lock.json` pushes within minutes | `cancel-in-progress` + singleton branch + `git fetch` before force-with-lease collapse to one PR carrying the latest hash. |
| Maintainer pushes manual `flake.nix` fix between bot runs | Bot sees hash correct, exits `changed=false`, no PR opened. |
| `nix build` fails for an unrelated reason | Detected via `PIPESTATUS != 0 && no got: line`, surfaced as `::error::` and the workflow exits non-zero. |
| Maintainer edits `flake.nix` concurrently | Anchored `sed` only touches the `npmDepsHash =` line, unrelated edits rebase cleanly. |
| Bot branch already exists on remote (re-run) | `git fetch origin chore/nix-npm-hash-update \|\| true` before push gives `--force-with-lease` an anchor. |
| Fork PR contributor applies the suggested fix | On the next synchronize, `changed=false` and the auto-fix workflow deletes the now-stale sticky comment. |
| Fork PR with shell metachars in branch name | All PR-controlled inputs (`head.ref`, `head.sha`, `head.repo.full_name`, `user.login`) are passed via `env:`, not interpolated into `run:`, so a malicious branch name like `;rm -rf /;#` cannot inject. |

### How I tested

- YAML syntax validated for all 4 modified workflows.
- Local repro of the regex anchor + tightened SRI shape (`^sha256-[A-Za-z0-9+/]{43}=$`) on real and adversarial inputs: real hash accepted, too-short / no-padding / double-padding rejected.
- Round-trip on `set-flake-npm-hash.sh`: setting a fake hash then the real one yields a byte-identical `flake.nix`.
- `nix run nixpkgs#prefetch-npm-deps -- web/package-lock.json` produces `sha256-RTL/3DccMZODLr9lMZdQfoMFtK2ZsARFMABh4SUV4xI=`, matching `flake.nix:78` (proves the bot's hash agrees with `buildNpmPackage`).
- The PR itself exercises the new `Nix Build Web` job because the diff touches `flake.nix`. A green job on this PR is the live proof the validator works and the new hash matches `web/package-lock.json` at `main` HEAD.

### Open questions for the maintainer

1. Is "Allow auto-merge" enabled at repo level (`Settings > General > Pull Requests > Allow auto-merge`)? If not, `gh pr merge --auto --squash` exits non-zero and the bot's PR sits open for a manual click. The workflow fails loudly so this is observable.
2. After this PR merges, please add `Nix Build Web` to the repository's required-status-checks list so a stale hash cannot land on `main` even when the bot is offline. Until that is done, the job runs but does not strictly block merge.

### Out-of-scope notes

- Migrating `flake.nix` to `pkgs.importNpmLock` (eliminates the `npmDepsHash` field entirely) is a larger refactor; not bundled here.
- Pushing to fork branches when `maintainer_can_modify=true`: comment-only for now; revisit after the comment path beds in.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

(Bug fix in CI plumbing plus a new auto-fix workflow that closes the contributor-side gap.)

## Checklist

(CI/Nix change, no UI.)

## AI Usage

- [ ] No AI was used

**AI Model/Tool used:** Claude Opus 4.7 via opencode (Sisyphus agent). Multiple rounds of Oracle sub-agents produced parallel design proposals (3 oracles for the post-merge fix, 3 for the PR auto-fix); each round resolved divergent decisions through cross-validation. Multiple automated review passes caught and fixed: regex anchor matching comment text, `gh pr list` race after `gh pr create`, silent `nix build` failure mode, `--force-with-lease` "stale info" on re-runs, multi-commit push detection in `BASE`, GraphQL node ID vs `databaseId` for REST PATCH, `set -e` swallowing the 0-match diagnostic, misleading `--force-with-lease` comment, too-permissive SRI regex, and a CRITICAL shell-injection vector via direct interpolation of PR-controlled `head.ref` into `run:` (now passed via `env:` like every other PR-controlled value). Reviewer questions are answered by me directly.

**Any Additional AI Details you'd like to share:** N/A

This PR restructures and hardens the nix-npm-hash automation to avoid direct pushes to main, add PR-time validation/fixes, and fail-fast on stale npm hashes.

What changed (high-level)
- Bot flow: Replaced post-merge direct-push bot with a singleton branch (chore/nix-npm-hash-update) + single PR flow. The bot creates/refreshes a PR and enables auto-merge (squash) so updates comply with required status checks.
- PR-time fixer: Added a pull_request_target workflow that computes the canonical npmDepsHash (prefetch-npm-deps, lockfile-only). For same-repo PRs it patches the PR branch and pushes the fix; for forked PRs it posts/updates a single sticky comment with a copy-paste fix snippet.
- PR validation: Added a PR CI job (nix-build-web) that conditionally runs a targeted nix build when web/package-lock.json, flake.nix, or flake.lock change to detect stale hashes before merge.
- Shared tooling: Added .github/scripts/get-flake-npm-hash.sh and set-flake-npm-hash.sh for strict extraction and atomic, validated updates (exact-one-match and hash-format checks).
- Cachix workflow: Removed the workflow_run chaining so cachix pushes no longer depend on the hash bot run.
- Flake update: Bumped npmDepsHash and added packages.prefetch-npm-deps in flake.nix so the new CI job validates the bump.
- Security/robustness: Minimized workflow permissions, used persist-credentials:false, scoped RELEASE_TOKEN use to git steps, used --force-with-lease anchored to read SHAs, and removed brittle push/retry loops.

Benefits
- Prevents CI failures caused by repository rules rejecting direct pushes (requires PR + checks).
- Provides immediate, clear feedback to contributors about stale npm hashes and either auto-fixes same-repo PRs or gives explicit instructions for forks.
- Safer automation with validated hash formats, single-match assertions, and race-condition protections.
- Faster PR validation by building only the web frontend (shorter CI time) instead of the entire workspace.

Notable technical details
- Hash computation: nix .#prefetch-npm-deps (lockfile-only) used consistently across workflows.
- get/set scripts enforce exactly one npmDepsHash match and validate new hashes with ^sha256-[A-Za-z0-9+/]{43}=$; set script performs portable in-place edits with atomic replace and cleanup trap.
- Bot uses secrets.RELEASE_TOKEN only for git push/fetch steps; gh CLI is used to create/edit PR and enable auto-merge (gh pr merge --auto --squash).
- pull_request_target workflow runs from base context, reads PR head via gh api, verifies head SHA before pushing same-repo fixes, and never executes untrusted PR code (passes PR-controlled values via env).
- nix-build-web runs nix build .#aoe-with-web (web frontend) when relevant files changed; workflows use minimal elevated permissions and persist-credentials: false.

Requested maintainer actions (documented in PR)
- Enable repo-level auto-merge for the bot PR and add Nix Build Web to required status checks after merging this change.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1420?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jerome-benoit (commented):
(content removed)

jerome-benoit (commented):
(content removed)

jerome-benoit:
## Self-review (code-reviewer mode)

The core design (PR-based bot flow + `Nix Build Web` PR-CI guard + dropping the `workflow_run` chain) is sound and the new shell scripts are tidy (anchored regex, hash-shape validation, idempotent `--force` label). The bot-side workflow (`nix-npm-hash.yml`) and the `cachix-push.yml` simplification both look fine. Verified the scripts locally against `flake.nix`: `MATCHES=1`, extraction returns the new `RTL/3DccMZODLr9lMZdQfoMFtK2ZsARFMABh4SUV4xI=` hash, and `set-flake-npm-hash.sh` rejects shapes like `sha256-not-a-valid-hash` and `sha256-...; echo pwn`.

One **critical** finding in the new auto-fix workflow that should block merge, plus a few smaller items.

### 1. Critical: shell injection via `head.ref` in `nix-npm-hash-fix-pr.yml` (line 49)

```yaml
echo "head_ref=${{ github.event.pull_request.head.ref }}" >> "$GITHUB_OUTPUT"
```

`${{ github.event.pull_request.head.ref }}` is interpolated directly into a `run:` script. Git accepts `;`, `"`, `$`, `` ` ``, `|`, `&` in ref names (verified locally with `git check-ref-format refs/heads/foo;bar` etc.), and a fork PR's `head.ref` is fully attacker-controlled. This is the exact pattern of CVE-2023-49291 (`tj-actions/branch-names`) and the recent `ceph/ceph` [GHSA-p433-fp4g-pc2c](https://github.com/ceph/ceph/security/advisories/GHSA-p433-fp4g-pc2c), where a branch named `main;printenv;#` opened RCE in the base repo's runner.

Concretely on this workflow:
- The injection runs in step `Detect fork and pin head SHA`, which doesn't yet have `RELEASE_TOKEN` in env, but the step runs in the shared `$GITHUB_WORKSPACE`.
- Attacker can (a) overwrite `$GITHUB_OUTPUT` to set `is_fork=false` (last value wins, per [Actions docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)), and (b) modify `.github/scripts/set-flake-npm-hash.sh` in the workspace.
- The next step `Push fix commit` then runs (because is_fork was rewritten to false), executes the trojaned script with `GH_TOKEN=secrets.RELEASE_TOKEN` in its env, and the PAT is exfiltrated.

`pull_request_target` + write-scoped PAT means full repo write (push to `main` via the paths the bot already controls, etc.). Recommend not merging until this is fixed.

**Fix:** use `env:` mapping like the rest of the file already does on line 89:

```yaml
- name: Detect fork and pin head SHA
  id: pr
  env:
    HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name }}
    BASE_REPO: ${{ github.repository }}
    HEAD_SHA: ${{ github.event.pull_request.head.sha }}
    HEAD_REF: ${{ github.event.pull_request.head.ref }}
  run: |
    if [ "$HEAD_REPO" = "$BASE_REPO" ]; then
      echo "is_fork=false" >> "$GITHUB_OUTPUT"
    else
      echo "is_fork=true" >> "$GITHUB_OUTPUT"
    fi
    echo "head_sha=$HEAD_SHA" >> "$GITHUB_OUTPUT"
    echo "head_ref=$HEAD_REF" >> "$GITHUB_OUTPUT"
    echo "head_repo=$HEAD_REPO" >> "$GITHUB_OUTPUT"
```

`head.sha` is hex-only and not exploitable, but moving it into env is consistent and cheap.

### 2. Medium: stale fork-PR comment after the contributor applies the fix

Both `Comment on fork PRs` and `Push fix commit` are gated on `steps.hash.outputs.changed == 'true'`. When a fork contributor applies the suggested `sed` and pushes, the rerun computes `changed=false` and neither step runs, so the existing "the `Nix Build Web` check fails" comment stays on the PR even after the check is now green. Either delete the marked comment or replace its body with a "now resolved" note when `changed=false`.

### 3. Medium: the auto-fix workflow isn't covered by the PR description

The "What" section lists four bullets (mirror open-release-pr pattern, Nix Build Web guard, drop workflow_run chain, bundle hash bump). The 169-line `nix-npm-hash-fix-pr.yml` is the largest single addition in the diff and isn't described, which makes scope hard to reconcile. Either describe it explicitly (it's a substantial separate feature with its own `pull_request_target` threat model) or split it into a follow-up so the security review can stand alone.

### 4. Low: commit message inconsistency between the two bot workflows

- `nix-npm-hash.yml`: `chore: update Nix npmDepsHash for web dependencies`
- `nix-npm-hash-fix-pr.yml`: `fix(ci): update Nix npmDepsHash for web dependencies`

Pick one. `fix(ci):` is more accurate for a desync repair and matches this PR's title.

### Things checked and OK

- `set-flake-npm-hash.sh` hash-shape regex: standard SRI base64 (43 chars + `=`), no shell metacharacters survive validation, sed delimiter `|` doesn't collide with allowed `[A-Za-z0-9+/=]`.
- `get-flake-npm-hash.sh` anchored regex correctly skips the comment on line 73 of `flake.nix` ("set npmDepsHash to lib.fakeHash...") that broke the previous `grep 'npmDepsHash'` approach. `MATCHES=1` assertion will catch any future second derivation.
- `nix-build-web` job in `ci.yml`: always runs the runner but skips the heavy steps via `if: steps.changes.outputs.relevant == 'true'`. Right shape for a required-status-check (a path-filtered workflow that doesn't run wouldn't satisfy the check). Multi-commit-push edge case via `github.event.before` is handled.
- `cachix-push.yml`: dropping `workflow_run` and relying on the `flake.nix` path filter is correct now that lockfile changes always land via the bot PR with the hash bump in the same merge commit.
- `pr-template-check.yml` only requires `## Checklist` + `I understand the code I am submitting`; the bot's PR body has both.
- Force-with-lease on `chore/nix-npm-hash-update`: `git fetch ... || true` populates the tracking ref, so the bare `--force-with-lease` correctly anchors. Concurrency group `nix-npm-hash` serializes runs.
- `prefetch-npm-deps` is integrity-only (Rust binary, fetches by SRI from lockfile; doesn't execute npm scripts), so running it on attacker-supplied lockfile data is safe modulo download-side-channel info.

jerome-benoit (commented):
(content removed)

njbrake (approved):

Solid work, and the followups in `fe2eb050` already cover the three review nits I would have flagged (credential persistence on checkout, empty-tree fallback for the initial-push edge case, and dropping `nix run --refresh`). The empty-tree-via-`git hash-object` choice is cleaner than the `HEAD~1` fallback I would have suggested.

The architecture mirrors `open-release-pr.yml`'s proven pattern, the `pull_request_target` workflow uses data-only fetch + env-passed PR-controlled values + `prefetch-npm-deps` (which only does SRI-validated tarball fetch, no script execution), and the bundled hash bump lets `Nix Build Web` prove itself end-to-end on this very PR.

Two follow-ups for me after this merges:
1. Confirm "Allow auto-merge" is enabled at the repo level so `gh pr merge --auto --squash` does not fail open.
2. Add `Nix Build Web` to the required-status-checks list so a stale hash cannot land on `main` even when the bot is offline.

Skipped Copilot suggestions both have good justifications in the followup commit message; no objection.

njbrake:

Pushed `095d5c27` to swap the `Nix Build Web` target from `aoe-with-web` to a new `aoe-web-frontend` package (just the `webFrontend` derivation).

Context: the previous job rebuilt the full Rust workspace from source on every PR touching `web/package-lock.json`, `flake.nix`, or `flake.lock`. With no Cachix configured on this job, that landed at 20-30 min on a cold runner, which is too long as a per-PR gate that fires on most web changes.

What the new target covers:

- The npm-deps fixed-output derivation that actually validates `npmDepsHash` (the original goal of this job).
- The `tsc -b && vite build` step, so frontend build regressions still surface at PR time.

What the new target drops:

- The crane Rust pipeline (`buildDepsOnly` + `buildPackage` with `--features serve`).

That Rust path is already covered by `Cargo Test` on every PR and by `cachix-push.yml` running the full `nix build .#aoe-with-web` post-merge. Wiring `--features serve` into `Cargo Test` is a separate follow-up, not bundled here.

Expected runtime: 1-2 min instead of 20-30. If you'd rather keep the original defense-in-depth scope, the cleaner path is probably "add Cachix read access to this job" rather than "rebuild from source on every PR"; happy to make that swap instead if you prefer.

PR #1423, opened 2026-05-22
fix(tui): render git-cliff release notes cleanly in "What's New" popup

njbrake:
## Description

The "What's New" popup in the TUI was showing raw markdown after #1417 / #1422 switched release notes to git-cliff:

> `**test:** De-flake live ensure-session-restart status check in [#1249](https://github.com/njbrake/agent-of-empires/pull/1249) by [@Seluj78](https://github.com/Seluj78) ([\`39d4b40\`](https://github.com/njbrake/agent-of-empires/commit/39d4b40b2ba5a886c4d9b9cff52644580aa3fc9c))`

…rendered verbatim, with every line falling through into "Other Changes" because the parser couldn't recognize the new format.

### Root cause

`src/tui/dialogs/changelog.rs` was written for GitHub's old auto-generated release notes (`* feat: foo by @x in https://.../pull/N`) and parsed bullets by looking for a conventional-commit prefix on each line. git-cliff now:

- emits `### Group` section headers (Features / Bug Fixes / Performance / Security / Reverts / Other),
- wraps the scope in markdown bold (`**scope:**`),
- replaces the plaintext PR / author / hash tail with markdown links (`in [#N](url) by [@user](url) ([\`hash\`](url))`).

None of that matched the old parser's expectations, so everything fell through with the markdown bleeding through visually.

### Fix

Rewrite `parse_release_body` and its helpers around the new format:

- Track the active group from `### Header` lines so categorization comes from git-cliff's grouping, not from re-parsing the type prefix per line.
- Peel `**scope:**` off the front before flattening so an inline colon in the subject can't be mistaken for the scope separator.
- Collapse markdown links (`[text](url)` → `text`) via a small flattener, no regex dep. Drop `**` and inline-code backticks afterward.
- Extract the PR number from ` in #N` and strip the ` by @user (hash)` tail. The trailing parens are only stripped when the contents look like a short hex commit hash, so legitimate `feat: foo (workaround)` subjects survive.

Old-format support (v1.7.x releases in the cache) is dropped per discussion; those will land in "Other Changes" until they age out of the popup.

> Heads-up: `.github/workflows/pr-template-check.yml` still grep's for the literal `I understand the code I am submitting`, but PR #1421 removed that line from the template itself. That check is now stale and will flag every future PR. I've put the line in this body to keep CI happy; the workflow should be updated to match the new template in a follow-up.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the new parser is exercised by 18 unit tests in `src/tui/dialogs/changelog.rs` covering `strip_attribution`, `strip_bold_scope_prefix`, `flatten_markdown`, and end-to-end `parse_release_body` on a body that mimics the real git-cliff output (Features / Bug Fixes / Performance / Other + New Contributors). The popup dialog itself is only the rendering layer above this; the existing scroll / submit-key tests cover that, and exercising it via the e2e harness would need a fake `cached_releases.json` and a version bump simulation that the harness doesn't grow tools for today. Smoke-tested the parser against the live v1.8.0 release body locally and confirmed all 84 entries categorize correctly.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7

Claude drafted the parser rewrite and tests based on the live v1.8.0 release body shape. I reviewed the diff, traced the helpers manually against a few real bullets, ran the smoke test locally, and confirmed `cargo test --lib`, `cargo fmt`, and `cargo clippy -- -D warnings` are clean.

## What's New

This PR fixes the TUI "What's New" popup so release notes generated by git-cliff render cleanly and are correctly categorized instead of showing raw markdown.

## What Changed

- Rewrote the release-note parser in src/tui/dialogs/changelog.rs to consume git-cliff markdown.
- Categorization now follows git-cliff `### Group` headers (cleared on `##` version headers) rather than per-line conventional-commit prefixes.
- Markdown cleanup before display: removes bold scope prefixes, flattens links and inline code, and strips certain trailing attribution parts.
- Dropped support for the old GitHub auto-generated format; older entries will fall into "Other Changes" until they age out.
- Removed the per-item breaking-change rendering (the unreachable BREAKING branch and the `breaking` field were deleted).
- Added 18 unit tests covering scope stripping, markdown flattening, attribution/PR extraction, grouped parsing, and end-to-end parsing.

## Benefits

- Users see readable, well-categorized release notes in the popup (no raw markdown).
- Parser aligns with current git-cliff output, reducing miscategorization and display issues.
- Increased test coverage reduces regression risk.

## Technical Details

- parse_release_body now:
  - Tracks active category from `### <Group>` headers and skips `## <version>` headers.
  - Only accepts real markdown bullets (`- ` / `* `) and ignores "New Contributors" lists.
  - Strips `**scope:**` prefixes before flattening text.
  - Flattens `[text](url)` → `text`, removes `**` and backticks.
  - Extracts PR numbers from trailing "in `#N`" and removes trailing `by `@user` (short-hash)` when the parenthesis matches a short commit hash.
  - Removes the old conventional-commit parsing and URL-based attribution logic.
- Files changed: primarily src/tui/dialogs/changelog.rs (roughly +240 / -206 lines). Breaking-change callouts were removed; reintroducing them would require extending the git-cliff template first.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1423?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Seluj78 (changes requested):
Careful, might be needed in web as well ?

njbrake:
> _Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Good instinct, thanks for the flag. I checked and the web side is unaffected: the release body is only consumed in two places in the repo, `src/update/mod.rs` (the cache) and `src/tui/dialogs/changelog.rs` (this popup). On the web side, `web/src/components/UpdateBanner.tsx:94` just renders an anchor to `status.release_url`, which points at the GitHub release page; GitHub renders the markdown there natively, so there's no parser in `web/` to update. No analogous fix needed.

`grep -rln "ReleaseInfo\|release\.body\|release_body" src/ web/src/` returns just the two Rust files, which I'm taking as confirmation.

PR #1424, opened 2026-05-22
feat: throwaway-directory toggle for new sessions

### LINKED ISSUE #1324: Add a toggle to launch a new session in a throwaway dir
Sometimes you want to create a session just to ask a question or to have it investigate something and you don't necessarily want it tied into any specific repository. There should be some kind of option when you're creating a session to let you put it into a temp directory and not need to specify a path.

Seluj78:

## Description

Adds a "throwaway directory" toggle to new-session creation across the CLI, TUI, and web wizard. A throwaway session does not belong to any project on disk: AoE provisions a fresh directory under `std::env::temp_dir()` named `aoe-throwaway-<id>`, attaches the session to it, and unconditionally removes the directory when the session is deleted. The use case is the one in the issue: ask a question, run a quick investigation, scratch around with the agent without picking or polluting a real repo.

The feature persists as `instance.throwaway: bool` (a single `#[serde(default)]` field on `Instance`); no data migration is required. The session-layer helper at `src/session/throwaway.rs` owns the naming contract (`aoe-throwaway-<id>` basename, `fs::create_dir` so collisions surface as errors) and the deletion guard (`is_throwaway_path` checks both temp-dir ancestry AND the basename prefix, so a tampered session JSON pointing at `/etc` is left alone).

Three entry points, one persisted flag:

- **CLI:** `aoe add -T -t Quick -c claude` (also `--throwaway`). Conflicts at clap parse time with every worktree-related flag (`-w`, `--new-branch`, `--base-branch`, `--repo`, `--project`, `--no-submodules`). Passing an explicit project path alongside `--throwaway` is rejected at runtime; the historical `aoe add` no-args default still works.
- **TUI:** Ctrl+T inside the new-session dialog flips the flag, dims the Path input, force-clears the Worktree toggle, and skips the path-exists "Create dir?" prompt on Enter.
- **Web wizard:** a `Skip project folder` Toggle above the Recent / Browse / Clone-URL tabs. The reducer enforces mutual exclusion bidirectionally: enabling throwaway clears `path`/`extraRepoPaths`/`useWorktree`; selecting any real path clears `throwaway`. SessionStep hides the worktree section, ReviewStep labels the project slot "Temporary directory (provisioned on create)", and `collectRecentProjects` filters throwaway sessions out so transient temp dirs never become re-selectable Recents.

The server-side `CreateSessionBody` gains a `throwaway` field and returns 400 (`Cannot combine throwaway with worktree_branch`) before reaching the builder, so a misbehaving direct API client can't bypass the UI mutex. The shell-injection check on `body.path` is skipped only when `throwaway` is true (the client sends `""`). `SessionResponse.throwaway` is exposed so the web wizard can filter the Recents list.

Test coverage:

- Rust: 4 throwaway helper unit tests (provisioning, collision, guard-accept, guard-reject); 2 builder tests (throwaway provisions a temp dir, `throwaway + worktree` errors); 4 deletion-guard tests (accept, missing-dir, tampered path is NOT removed, non-throwaway under temp is NOT removed); 4 TUI dialog tests (toggle, mutex with worktree, empty-path-on-submit, skip-confirm-on-enter); 3 e2e tests (happy path including `aoe rm` cleanup, explicit-path rejection, clap-level worktree conflict).
- Vitest: 4 new wizardReducer cases for the bidirectional mutex; 816/816 in the full suite.
- Live Playwright: 5 user-story specs under `web/tests/live/wizard-throwaway-*.spec.ts` mirroring the canonical pattern from #1397 (toggle visible, enables-next, hides-worktree, full launch happy path, delete-cleans-tempdir). Each gets a matching `web/tests/coverage-matrix.json` entry; `validate-coverage-matrix.mjs` is green.

Local validation: `cargo fmt --check`, `cargo clippy --features serve --all-targets -- -D warnings`, `cargo test --features serve`, `cargo build --features serve`, `cargo xtask gen-docs`, `npx tsc --noEmit`, `npx vitest run`, and `node web/tests/validate-coverage-matrix.mjs` all green. (15 cockpit-ACP integration tests fail locally with `agent did not complete the ACP initialize handshake within 30s` because the local machine doesn't have `@agentclientprotocol/claude-agent-acp` installed; not introduced by this branch and unrelated to the change.)

Closes #1324

### Follow-up tracked separately

`aoe serve` startup currently does not sweep stale `aoe-throwaway-*` directories left behind when the daemon dies between session creation and explicit deletion. OS temp-cleaning policies pick most of these up, but a deliberate sweep on supervisor startup (cross-reference live session IDs, age policy, audit log) belongs in its own PR. I have NOT auto-filed the follow-up issue from this PR; happy to open one as a separate step if you want it tracked in Issues.

## PR Type

- [x] New Feature

## Checklist

## How to test

User-runnable verification steps, one per acceptance criterion:

- [ ] `aoe add -T -t Quick -c claude` succeeds. The printed `Path:` line points under `$TMPDIR` (macOS) or `/tmp` (Linux), and the summary shows `Throwaway: yes`.
- [ ] `aoe add /some/real/path --throwaway` fails with `Cannot specify a project path with --throwaway`.
- [ ] `aoe add --throwaway -w foo` fails at the clap layer with a flag-conflict message.
- [ ] `aoe rm Quick` (the throwaway from step 1) succeeds; the printed directory from step 1 no longer exists on disk.
- [ ] Open the TUI new-session dialog, press Ctrl+T; the Path field swaps to `(throwaway directory, Ctrl+T to undo)` and the Worktree toggle deactivates. Enter creates the session in a fresh temp dir.
- [ ] In the web wizard, the ProjectStep shows a "Skip project folder" toggle above the Recent/Browse/Clone-URL tabs. Flipping it on hides the tabs, enables Next without a path. SessionStep hides the worktree section. ReviewStep shows "Temporary directory (provisioned on create)". Launching produces a session row whose `GET /api/sessions` entry has `throwaway: true` and a `project_path` under `os.tmpdir()`.
- [ ] Delete a throwaway session via the dashboard's sidebar context menu; the temp directory from `project_path` is removed from disk.
- [ ] In the wizard, after deleting a real project's sessions, throwaway sessions never appear in the Recent tab.

## AI Usage

- [ ] No AI was used

Investigation, plan, debate-with-three-LLMs synthesis, and implementation drafted by Claude under my direction. I picked the UX shape (toggle vs tab vs checkbox), the CLI ergonomics (`-T` short flag, hard clap mutex over silent precedence), the cleanup contract (guard by temp-dir ancestry AND basename prefix), the scope cuts (orphan-sweep deferred), and reviewed each commit before pushing.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -130,9 +130,7 @@ Add a new session
 
 ###### **Arguments:**
 
-* `<PATH>` — Project directory (defaults to current directory)
-
-  Default value: `.`
+* `<PATH>` — Project directory (defaults to current directory). Omit when using `--throwaway`
 
 ###### **Options:**
 
@@ -158,6 +156,7 @@ Add a new session
 * `--no-cockpit` — Force terminal/PTY mode for this session, overriding the default-for-claude cockpit setting
 * `--agent <AGENT>` — Pick a specific cockpit agent (e.g., aoe-agent, claude-code). Implies --cockpit
 * `--model <MODEL>` — Override the model used by aoe-agent (e.g., claude-opus-4-7, gpt-5, gemini-2.5-pro). Forwarded to the agent at session start
+* `-T`, `--throwaway` — Create the session in a fresh temporary directory instead of a project path. The directory is removed when the session is deleted. Mutually exclusive with worktree-related flags
 
 
 
--- a/docs/guides/throwaway-sessions.md
+++ b/docs/guides/throwaway-sessions.md
@@ -1,0 +1,105 @@
+# Throwaway sessions
+
+A throwaway session is a session that does not belong to any project on
+disk. When you start one, AoE provisions a fresh temporary directory
+under your system's temp folder, attaches the session to it, and
+removes the directory when you delete the session.
+
+Use throwaway sessions for one-off questions, quick investigations, or
+ad-hoc agent runs where you do not want a session record tied to a
+specific repo or to keep stray files around afterwards.
+
+## When to use it
+
+* You want to ask the agent a question that does not depend on a
+  particular codebase.
+* You want a clean, empty scratch directory for the agent to write
+  files into without polluting an existing project.
+* You want to investigate something quickly and have the directory
+  go away automatically when you are done.
+
+## Three ways to start one
+
+### Command line
+
+```bash
+aoe add --throwaway -t "Quick question" -c claude
+# or the short flag
+aoe add -T -t "Quick question" -c claude
+```
+
+The session prints its resolved `Path:` line pointing at
+`/tmp/aoe-throwaway-<id>` (Linux) or `$TMPDIR/aoe-throwaway-<id>`
+(macOS). You do not pass a project path; it is provisioned for you.
+
+Trying to pass a path alongside `--throwaway` is rejected:
+
+```bash
+aoe add /Users/me/repo --throwaway
+# error: Cannot specify a project path with --throwaway
+```
+
+`--throwaway` is mutually exclusive with all worktree-related flags
+(`-w`, `--new-branch`, `--base-branch`, `--repo`, `--project`,
+`--no-submodules`). Mixing them fails at parse time with a clear
+conflict message.
+
+### Web dashboard
+
+In the new-session wizard, the **Project** step has a toggle labeled
+**Skip project folder** above the Recent / Browse / Clone URL tabs.
+Turning it on hides the path picker and the worktree controls on the
+Session step. The Review step shows
+"Temporary directory (provisioned on create)" in place of the path.
+
+Selecting a real project (Recent / Browse / Clone) turns the toggle
+back off, so the wizard never submits a request with both a real path
+and the throwaway flag set.
+
+### TUI new-session dialog
+
+Press `Ctrl+T` inside the new-session dialog from any field. The Path
+input is replaced with a `(throwaway directory, Ctrl+T to undo)`
+marker, the Worktree toggle is forced off, and submitting creates the
+session in a fresh temp directory. Press `Ctrl+T` again to revert.
+
+## What happens at delete
+
+When you delete a throwaway session via `aoe rm`, the web dashboard
+delete flow, or the TUI delete dialog, the deletion path also runs
+`fs::remove_dir_all` on the session's temp directory. The cleanup is
+guarded: it only runs when the session's `throwaway` flag is true AND
+the path is under your OS temp dir AND its basename starts with
+`aoe-throwaway-`. A session record with a tampered `project_path`
+pointing at, say, `/etc` is left alone.
+
+The wizard's **Recent projects** tab filters throwaway sessions out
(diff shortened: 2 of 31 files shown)


njbrake:

Thanks for putting this together. While reading through I went looking at how other tools handle the same problem and ran into Codex's [Chats feature](https://developers.openai.com/codex/app/features#chats), which has a similar "no project folder" entry point but keeps the working directory under the Codex home dir instead of throwing it away.

That got me wondering if there's a middle ground worth considering for this PR. A few half-formed thoughts to react to, push back on, or build on however makes sense to you:

- "Scratch" might land a bit softer than "throwaway" as a name, since it's more common dev vocabulary (scratch file, scratchpad). Same behavior, just a different label.
- Stashing the directory under `~/.agent-of-empires/scratch/<id>/` instead of the system temp dir might be friendlier, since `$TMPDIR` can get cleaned up by the OS on some platforms, and it'd make the folder easier to find if a user wants to peek.
- On delete, defaulting to remove (as you have it) feels right, but I wonder if there's a place for an optional "keep folder" escape hatch for the moments when someone realizes mid-cleanup that they actually want to rescue something. Detach and forget, no registry, just log the kept path and move on.

Seluj78:
> Thanks for putting this together. While reading through I went looking at how other tools handle the same problem and ran into Codex's [Chats feature](https://developers.openai.com/codex/app/features#chats), which has a similar "no project folder" entry point but keeps the working directory under the Codex home dir instead of throwing it away.

That's very interesting indeed, and I think it is indeed worth discussing.

>     * "Scratch" might land a bit softer than "throwaway" as a name, since it's more common dev vocabulary (scratch file, scratchpad). Same behavior, just a different label.

Yeah, I could see it. A shortcut/button for `new scratch session`, I like the name, much better and less barbaric  than throwaway.
 
>     * Stashing the directory under `~/.agent-of-empires/scratch/<id>/` instead of the system temp dir might be friendlier, since `$TMPDIR` can get cleaned up by the OS on some platforms, and it'd make the folder easier to find if a user wants to peek.

$TMPDIR is cleaned up like 99% of the times on restart on like, all OS I know of, so that is a limitation. The thing is, if we keep those scratches, we need a way to clean them up or it might bloat up the size of the folder for no reason. Configurable time or count of how much to keep, with option for unlimited ?

>     * On delete, defaulting to remove (as you have it) feels right, but I wonder if there's a place for an optional "keep folder" escape hatch for the moments when someone realizes mid-cleanup that they actually want to rescue something. Detach and forget, no registry, just log the kept path and move on.


I like it, I think it ties in with the settings above, maybe a setting to delete or not by default ?

Seluj78:
`on hold` while I split and merge https://github.com/njbrake/agent-of-empires/pull/1397 (so I can get proper playwright coverage and functionality.

Also waiting for decision on the discussion with @njbrake :)

Seluj78:
Closing in favor of agent-of-empires/agent-of-empires#1549. Org migration moved the canonical repo; the branch is rebased on the new `origin/main` and the PR body carries the open design discussion forward.

PR #1458, opened 2026-05-23
fix(sandbox): inject git safe.directory via env vars to fix dubious ownership error

flpdorea:
## Description

Sandbox containers run as `root`, but bind-mounted repositories are owned by the host user (UID 1000, shown as `ubuntu` inside the `aoe-dev-sandbox` image). Git's safe-directory check (introduced in Git 2.35.2 as a hardening measure against CVE-2022-24765) sees `root != ubuntu` and aborts with:

```
fatal: detected dubious ownership in repository at '/workspace/...'
```

This breaks **all** git operations inside the container, including `gh` CLI commands that shell out to git to resolve repository context (e.g., `gh issue view` fails with a GraphQL error because it cannot determine the current repo).

The fix injects `safe.directory=*` via Git's environment-variable config API (`GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_0`, `GIT_CONFIG_VALUE_0`; supported since Git 2.31). This overrides the ownership check without modifying any files and applies to all sandbox containers automatically, regardless of user config.

The change is localized to `collect_environment()` in `src/session/environment.rs`, the single chokepoint that builds env vars for both `docker run` (container creation) and `docker exec` (tmux session respawn).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
OpenCode (Kimi k2.6) — used for investigation, code analysis, and drafting the fix and PR description.

The AI helped trace the root cause from the `gh issue view` GraphQL error down to Git's safe-directory check, analyzed the container ownership model, and identified the `GIT_CONFIG_*` env-var API as the cleanest fix. I validated the fix manually inside a running sandbox container before submitting.

**NOTE:**

## What Changed

**Moved, not Added**: The Git `safe.directory` injection code block was repositioned in the `collect_environment()` function in `src/session/environment.rs`. It was moved from **before** the user-supplied environment entries loop to **after** it.

**Why This Matters**: The original placement meant that Git config defaults (GIT_CONFIG_COUNT, GIT_CONFIG_KEY_0, GIT_CONFIG_VALUE_0) were being injected first, then when user-provided entries were processed, the first-wins deduplication logic in `seen_keys` would silently discard any user-supplied GIT_CONFIG_* values. This was a critical bug—users couldn't override the defaults.

**The Fix**: By moving the defaults injection to **after** processing user entries, user-provided GIT_CONFIG_* values from `extra_env` or `sandbox.environment` now take precedence as intended.

**Updated Comment**: The code comment was expanded to explicitly note: "Placed after the user entries loop so caller-provided GIT_CONFIG_* values take precedence (first-wins deduplication via seen_keys)."

**New Test**: Added `test_collect_environment_git_safe_directory_user_override()` which verifies that user-provided GIT_CONFIG_* entries (GIT_CONFIG_COUNT=2 with custom safe.directory paths) correctly override the built-in defaults.

**Code Changes**: +89/-25 lines in `src/session/environment.rs` (net: +64 lines)

## Benefits

- **Restores user override capability**: Users can now provide their own Git configuration values without them being silently discarded
- **Preserves sensible defaults**: The built-in `safe.directory=*` still applies automatically when users don't provide custom configuration
- **Increased predictability**: Behavior now follows expected "user values override defaults" pattern
- **Better testability**: New test ensures this behavior is maintained across future changes

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1458?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/environment.rs
+++ b/src/session/environment.rs
@@ -401,6 +401,31 @@ pub(crate) fn collect_environment(
         }
     }
 
+    // Git's safe-directory check fails when the container user (root) does not
+    // match the file owner (host UID 1000, shown as "ubuntu" inside the
+    // aoe-dev-sandbox image). Bind-mounted repos trigger:
+    //   fatal: detected dubious ownership in repository at '...'
+    // We inject safe.directory=* via Git's env-var config API (Git 2.31+),
+    // which overrides the check without modifying any files.
+    if seen_keys.insert("GIT_CONFIG_COUNT".to_string()) {
+        result.push(EnvEntry::Literal {
+            key: "GIT_CONFIG_COUNT".to_string(),
+            value: "1".to_string(),
+        });
+    }
+    if seen_keys.insert("GIT_CONFIG_KEY_0".to_string()) {
+        result.push(EnvEntry::Literal {
+            key: "GIT_CONFIG_KEY_0".to_string(),
+            value: "safe.directory".to_string(),
+        });
+    }
+    if seen_keys.insert("GIT_CONFIG_VALUE_0".to_string()) {
+        result.push(EnvEntry::Literal {
+            key: "GIT_CONFIG_VALUE_0".to_string(),
+            value: "*".to_string(),
+        });
+    }
+
     for entry in entries {
         if let Some((key, value)) = entry.split_once('=') {
             if seen_keys.insert(key.to_string()) {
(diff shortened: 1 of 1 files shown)


njbrake (approved):

## Summary

Injects `GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=safe.directory`, `GIT_CONFIG_VALUE_0=*` into every sandbox container's environment so that Git's safe.directory check does not fail on bind-mounted repos. The injection is placed after the user/config env loop so caller-supplied `GIT_CONFIG_*` values take precedence via first-wins dedup in `seen_keys`.

## Validated the bug is real (Linux-only)

I confirmed this is a real Linux bug that is invisible on macOS, which is why it hasn't shown up in daily maintainer use:

- `src/containers/runtime_base.rs:157` builds `docker run` args without a `-u` flag, and `docker/Dockerfile` has no `USER` directive, so the container runs as root (UID 0).
- On Linux, bind mounts preserve host UIDs literally; a repo owned by UID 1000 on the host stays UID 1000 inside the container, so the safe.directory check fires.
- On macOS Docker Desktop, virtiofs / gRPC-FUSE remaps ownership inside the container, so the UID check passes and the bug is hidden.
- WSL2 users would also be affected.

The "shows as ubuntu inside the container" detail is also accurate: `ubuntu:24.04+` base images ship with a default `ubuntu` user at UID 1000, so a host UID of 1000 happens to resolve to that name. Cosmetic; the UID mismatch is what triggers the error.

## Code review

- `collect_environment` injection: correct placement (after the user loop), correct gating via `seen_keys.insert(...)`, correct env API (Git 2.31+).
- New unit tests cover the default path and the full user override path. Both pass locally.
- `cargo fmt`, `cargo clippy`, and the new tests are clean on the merged tree.

## CodeRabbit check-in

CodeRabbit's actionable finding (defaults inserted before the user loop, breaking override semantics) was already addressed in commit `7be86dd7` by moving the block after the loop. The nitpick request for a collision test is satisfied by `test_collect_environment_git_safe_directory_user_override`. No outstanding bot items.

## Suggestion (non-blocking)

Partial-override edge case: the three keys are gated independently, so a user who supplies only `GIT_CONFIG_KEY_1` / `VALUE_1` (intending to add an entry on top of the default) will see the default inject `COUNT=1`, which silently orphans the user's `_1` pair (git only reads indices `0..COUNT`). The current behavior is defensible (users can also set `COUNT`), so this is worth noting rather than blocking on. A one-line comment or a docs note explaining that overriding requires supplying the full `COUNT` plus `KEY_0` plus `VALUE_0` trio would be enough.

## Verdict

Approve. Small, well-scoped, well-tested fix for a real Linux-specific sandbox bug. Thanks for the careful root-causing and the iteration to address CodeRabbit's precedence concern.

PR #1244, opened 2026-05-19
feat(tui): per-row profile tag in all-profiles view

BTForIT:
## Description

In all-profiles view the session list title only shows `[all]`, so rows
from different profiles are visually indistinguishable: a session under one
profile looks identical to one under another. The active profile is named in
the title only when the view is filtered to a single profile.

This adds a dimmed per-row `[profile]` tag, shown only in all-profiles view:

- The tag is rendered in `render_item_line()` after the worktree branch and
  before the activity column, so its width is counted into `used_width` and
  the right-aligned activity column still lines up across rows.
- It is gated on `self.active_profile.is_none()` (all-profiles view). A
  filtered view already names the profile in the list title, so the per-row
  tag would be redundant there and is omitted.
- The tag reads `inst.source_profile`, which every instance already carries.
  No new struct fields, no migrations, and no agent-specific code.

It uses `theme.dimmed`, the same styling as the activity column and the sort
indicator, so it reads as secondary metadata rather than competing with the
session title.

**Testing**

2 new unit tests: one asserts every row in all-profiles view carries a tag
matching its `source_profile`, the other asserts a filtered view renders no
per-row tag. The full library suite (1718 tests) passes and
`cargo fmt --check` is clean. `cargo clippy` surfaces no new warnings from
this change.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] I understand the code I am submitting
- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

## AI Usage

- [ ] No AI was used
- [x] AI was used for drafting/refactoring
- [ ] This is fully AI-generated

**AI Model/Tool used:** Claude Code (Claude Opus 4.7)

**Any Additional AI Details you'd like to share:**
Implemented with Claude Code under the maintainer's direction and review.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than
copy/pasting reviewer comments into an AI and pasting back its answer. We want
to discuss with you, not your AI :)

- [x] I am an AI Agent filling out this form (check box if true)

High-level changes
- Adds configurable per-row tags for session rows (new SessionConfig.row_tag with RowTagMode: None, Auto, Profile, Sandbox, Branch).
[...]

PR #1459, opened 2026-05-23
feat(session): default title from project path when no branch is set

BTForIT:
## Description

When creating a session without an explicit title and without a worktree branch, `resolve_title` currently picks a random civilization name (Babylon, Sumer, ...). For folks with several sessions pointed at the same repo, that produces a project list where every row looks like a different empire, with no anchor back to the underlying project.

This PR adds a `project_path`-aware fallback between the existing branch-name path and the random fallback. When the project path lives beneath a known "dev container" (`GitProjects`, `Projects`, `projects`, `Code`, `code`, `dev`, `src`, `workspace`, `Workspaces`, `repos`, `Repos`), the new behaviour is:

- Path is the immediate child of the container → use that name (e.g. `~/GitProjects/per-dev` → `per-dev`).
- Path sits inside that top-level project → combine both names with a hyphen (e.g. `~/GitProjects/per-dev/forks/agent-of-empires` → `per-dev - agent-of-empires`).
- Path is outside any known container → fall through to the existing random civilization name. No regression for users whose repos don't live under a familiar parent directory.

The case-insensitive matcher accepts both `Projects` and `projects` (and similar variants) so users on different OSes / conventions get the same behaviour. The deepest container wins, so something like `~/projects/monorepo/src/server` resolves to `server` (anchored on the inner `src`).

`resolve_title` gains a `project_path: &str` parameter (breaking; only two in-tree callers, both updated in this PR). The full precedence is:

1. Explicit user-supplied title (unchanged).
2. Worktree branch name when worktree mode is enabled (unchanged).
3. **NEW**: derive from project path under a known container.
4. Random civilization name (existing fallback).

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test --lib --release session::builder`)
- [ ] Documentation was updated where necessary  — N/A, no user-facing docs cover the title default precedence
- [ ] For UI changes: included screenshot or recording — N/A, the change only affects what string the existing title slot defaults to

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle — the change is purely a string-derivation function with unit tests covering every branch (top-of-container, nested-subdir, case-insensitive, deepest-match, outside-container, worktree-branch-wins).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Implementation drafted by Claude with the author's review; the heuristic (dev-container anchor + hyphen-joined two-level title) and the chosen container list reflect the author's intent. All seven new tests were added in the same commit.

## Summary

This PR enhances session title resolution to intelligently derive meaningful default titles from repository paths when no explicit title is provided and no worktree branch is available.

### What Changed

**Core functionality:**
- Enhanced `resolve_title()` function now accepts the project path as a new parameter, enabling path-based title derivation
- Added `derive_title_from_path()` helper that intelligently extracts titles from project directory structures
- Introduced a `DEV_CONTAINERS` constant containing 11 common developer directory names (GitProjects, Projects, Code, dev, src, workspace, repos, etc.)
- Updated the session title precedence order to include path-based derivation as a fallback before random civilization names

**Behavior:**
When creating a session without an explicit title and without a worktree branch:
- If the project sits directly under a known dev container directory → uses the project folder name (e.g., `~/GitProjects/my-app` → "my-app")
- If the project is nested deeper → combines the top-level project and leaf folder names (e.g., `~/GitProjects/my-app/forks/agent-of-empires` → "my-app - agent-of-empires")
- Uses the deepest matching container, supporting nested scenarios like `~/projects/monorepo/src/server` → "server"
- Gracefully falls back to random civilization names for projects outside known containers

**Testing:**
- Seven new unit tests added covering:
  - Direct child of a dev container
  - Nested subdirectories within containers
  - Case-insensitive container matching
  - Deepest container selection
  - Outside-container fallback behavior
  - Precedence rules (explicit title and worktree branch still take priority)

### Files Modified

- `src/session/builder.rs` - Core changes to title resolution logic (+152/-17 lines)
- `src/tui/home/mod.rs` - Updated session creation to pass project path to title resolver (+1/-0 lines)

### Benefits

- **More deterministic titles**: Sessions get meaningful names based on project structure instead of random civilization names
- **Better session organization**: Related projects get similar title prefixes, making them easier to identify
- **Preserves existing behavior**: User-provided titles and worktree branches still take priority
- **Backward compatible**: Sessions without paths or outside known containers fall back to the existing random naming scheme

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1459?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/builder.rs
+++ b/src/session/builder.rs
@@ -331,6 +331,7 @@ pub fn build_instance(
     let mut warnings: Vec<String> = Vec::new();
     let final_title = resolve_title(
         &params.title,
+        &final_path,
         params.worktree_branch.as_deref(),
         params.worktree_enabled,
         existing_titles,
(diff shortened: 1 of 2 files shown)


njbrake:

Thanks for putting this together, Benjamin. Before we dig into the implementation, I want to surface some history and float two alternatives, because I think we may be re-litigating a decision you already made and there's a slot in the UI that fits this better.

**The history.** Back in #1077 we cherry-picked your folder-basename auto-fill for the title field in the new-session dialog. In #1070 you then removed it yourself, with this note in the description:

> The folder-basename auto-fill from #1077 was removed in the fixup commit so the random civ-name fallback (`src/cli/add.rs`) stays the canonical empty-submit behavior across all flows (typed path, default cwd, Ctrl+P browse).

I read that as a deliberate call to keep the random civ name as the canonical default, and I still agree with it. This PR walks that decision back, just one layer deeper (`resolve_title` instead of dialog auto-fill). I'd rather not lose the civ name as the default; it's one of the small bits of personality in the TUI and once you start seeing `per-dev - agent-of-empires` style titles, the empires are gone.

**First, a workflow question.** The thing I'm missing here is a picture of how you're actually using the TUI day to day. Would you be willing to drop a short screen recording (asciinema or just a video) of a typical session list, scrolling around, switching between sessions? Your fleet is clearly larger and more multi-repo than mine, and I suspect there's friction I'm not seeing from my own usage. That would also tell me whether Project grouping (`g` keybind, added in #649) is something you've tried and found lacking, vs. something that hasn't been on your radar. Project grouping is the default for new users and it directly anchors "I have several sessions pointed at the same repo." If that mode doesn't render usefully for your fleet, that itself is a bug worth filing separately, and it may change the shape of this PR.

**Two alternatives that solve the project-anchoring problem without touching the title.**

1. **Project grouping (`g`).** Already in #649. Sessions get grouped by repo name visually. Worth confirming whether this covers your case before we add a parallel mechanism.

2. **Extend the row-tag slot you built in #1244.** `RowTagMode` already has `None | Auto | Profile | Sandbox | Branch`. Adding `Project` would render `🔵 Sumer  [agent-of-empires]` instead of replacing Sumer. Reuses the dimmed style, the width-aware layout, and the per-profile override surface you already wired up. Much smaller diff than this PR, no breaking signature change on `resolve_title`, and the civ name stays.

**One bigger idea on top of (2).** Right now `RowTagMode` is single-valued. We could make it compound, e.g. `RowTagMode::Composite(Vec<RowTagPart>)` or accept a comma-separated list in settings (`profile,project`), and let users pick what shows up. That gets us out of the business of inventing new top-level modes every time someone wants to see another piece of metadata, and lets a power user say "show me profile and project" without us picking one. I haven't thought through whether that's worth the complexity vs. just adding `Project` as another single-valued variant, but it feels like the right shape if more requests like this keep coming. Open to your read.

Happy to keep this open while we talk through it. The recording is the unblocker for me, even more than picking between the alternatives above.

njbrake (changes requested):
See other comment.

PR #1477, opened 2026-05-24
fix(hermes): implement real pane-based status detection

### LINKED ISSUE #1474: Hermes agent status always shows 'idle' — hooks not writing status file
## Summary

When using [Hermes](https://github.com/nousresearch/hermes-agent) as an agent in AOE, the status in the left sidebar is **always `idle`**, regardless of what the agent is actually doing. This affects all users, not a specific configuration.

## Root cause

AOE's status detection for Hermes relies entirely on shell hooks writing to a sidecar file (`/tmp/aoe-hooks/<instance-id>/status`). There is no tmux pane-content fallback — the fallback function is an intentional stub that always returns `Idle`:

```rust
// src/tmux/status_detection.rs
pub fn detect_hermes_status(_content: &str) -> Status {
    Status::Idle  // always idle — intentional stub
}
```

This means that if the hooks fail to write the status file for any reason, the status will permanently show as `idle` with no fallback.

## What's happening with the hooks

AOE installs hooks into `~/.hermes/config.yaml` and pre-populates `~/.hermes/shell-hooks-allowlist.json`. The event names and YAML structure appear correct. However, there is a model mismatch in how the hook commands are designed:

- **Hermes expects** each hook to receive event context as JSON on **stdin** and optionally return JSON on **stdout** (to block/transform operations)
- **AOE's hook commands** ignore stdin entirely and just write a status string to a file, producing no stdout

Example of what AOE writes:
```bash
sh -c '[ -n "$AOE_INSTANCE_ID" ] || exit 0; mkdir -p /tmp/aoe-hooks/$AOE_INSTANCE_ID 2>/dev/null; printf running > /tmp/aoe-hooks/$AOE_INSTANCE_ID/status 2>/dev/null; exit 0'
```

Hermes may treat the missing JSON response as a silent error, or `AOE_INSTANCE_ID` may not be available in the hook's environment when Hermes executes the command — either way, the status file is never written.

## Affected versions

- AOE: current `main` branch
- Hermes: current `main` branch
- Affects: **all users** running Hermes in AOE

## Expected behavior

The sidebar should show `running` / `waiting` / `idle` accurately as Hermes processes requests, consistent with Claude Code and other agents.

## Suggested fix

Two possible approaches:

1. **Fix the hook contract** — update AOE's Hermes hook commands to properly handle Hermes' stdin/stdout model and ensure `AOE_INSTANCE_ID` is inherited by the hook subprocess

2. **Add pane-content fallback** — implement `detect_hermes_status()` to parse Hermes' tmux pane output (spinners, prompts, etc.) as a fallback, similar to how OpenCode and Copilot are handled

## References

- Hook execution model in Hermes: `hermes_cli/plugins.py` and `agent/shell_hooks.py`
- Hook installation in AOE: `src/hooks/mod.rs`
- Status stub in AOE: `src/tmux/status_detection.rs`
- Status polling in AOE: `src/tui/status_poller.rs`

angelogalanti:
## Description

`detect_hermes_status` was a stub that always returned `Status::Idle`, so Hermes sessions never showed Running or Waiting in the TUI regardless of what the agent was doing (fixes #1474).

Replaced the stub with a real pane-content detector covering Hermes v0.14+ output patterns:

| Signal | Status |
|---|---|
| Spinner faces `◜ ◠ ✧` (non-default skins) | Running |
| `❯ Ctrl+C to interrupt…` in the prompt area | Running |
| Active tool-execution lines prefixed with `┊` | Running |
| Thinking verbs (`reasoning`, `pondering`, `analyzing`, …) | Running |
| Dangerous-command approval prompt (`[o/s/a/d]`, `dangerous command`) | Waiting |
| Bare `❯` or `⚡` input prompt (turn finished, ready for next message) | Idle |
| Anything else | Idle |

The `❯`/`⚡` prompt maps to **Idle** rather than Waiting because in AoE semantics `Waiting` means "blocked on user approval for a dangerous command", not "at the input prompt between turns".

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Manually verified with a live Hermes v0.14.0 session (kimi k2.6 model): the TUI now shows the animated green spinner while processing and the gray static `⠒` icon when idle at the prompt.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the detection logic is fully covered by 8 new unit tests in `src/tmux/status_detection.rs`; an e2e test would require a live Hermes session in CI which isn't available.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Code (claude-sonnet-4-6)

**Any Additional AI Details you'd like to share:** Used Claude Code to explore the codebase, identify the stub, find the real Hermes output patterns via live testing, and implement + iterate on the fix.

### DIFF
--- a/src/tmux/status_detection.rs
+++ b/src/tmux/status_detection.rs
@@ -862,9 +862,21 @@ pub fn detect_hermes_status(raw_content: &str) -> Status {
         return Status::Running;
     }
 
-    // Input prompt ❯ (default skin) or ⚡ (cyberpunk skin) in the last few
-    // lines means the agent finished its turn. Check before scrollback
-    // activity words to avoid false-positive Running from a previous turn.
+    // While running, Hermes replaces the input prompt with
+    // "❯ Ctrl+C to interrupt…". Check this before the idle-prompt
+    // detection below so we don't misidentify Running as Waiting.
+    if non_empty_lines
+        .iter()
+        .rev()
+        .take(5)
+        .any(|l| l.contains("ctrl+c to interrupt"))
+    {
+        return Status::Running;
+    }
+
+    // Input prompt ❯ (default skin) or ⚡ (cyberpunk skin) on its own means
+    // the agent finished its turn. Placed before scrollback activity words to
+    // avoid false-positive Running from a previous turn.
     for line in non_empty_lines.iter().rev().take(5) {
         let clean = strip_ansi(line).trim().to_string();
         if clean == "❯" || clean.starts_with("❯ ") || clean == "⚡" || clean.starts_with("⚡ ")
@@ -873,7 +885,7 @@ pub fn detect_hermes_status(raw_content: &str) -> Status {
         }
     }
 
-    // Active tool execution lines are prefixed with ┊; check recent lines only
+    // Active streaming lines are prefixed with ┊; check recent lines only
     // to avoid triggering on scrollback from a completed turn.
     if non_empty_lines
         .iter()
@@ -884,8 +896,9 @@ pub fn detect_hermes_status(raw_content: &str) -> Status {
         return Status::Running;
     }
 
-    // Thinking verbs from built-in and community Hermes skins.
+    // Thinking verbs from the default skin and community Hermes skins.
     let activity_indicators = [
+        "reasoning",
         "pondering",
         "contemplating",
         "forging",
@@ -897,7 +910,6 @@ pub fn detect_hermes_status(raw_content: &str) -> Status {
         "analyzing",
         "computing",
         "evaluating",
-        "ctrl+c to interrupt",
     ];
     for indicator in &activity_indicators {
         if last_lines.contains(indicator) {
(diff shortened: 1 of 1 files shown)


njbrake (approved):

Approving. Diff is well-scoped (single file), the priority ordering (spinner → interrupt hint → input prompt → ┊ → activity verbs → approval) is sensible, and the input-prompt check correctly overrides scrollback activity verbs. CI green, 8 new unit tests pass, fmt + clippy clean locally.

Two small nits, non-blocking:

1. Two comments contain em dashes (`status_detection.rs` lines 878 and 2235 of the new file). The project rule is no em dashes or `--` as prose separators; rephrase with commas or split into two sentences.
2. `raw_content` is already ANSI-stripped by `detect_status_from_content` before it reaches `detect_hermes_status`, so the `strip_ansi(line)` call inside the input-prompt loop is redundant. Harmless and matches what the Gemini detector does, so happy to leave it as-is.

Worth flagging that this is a fallback / defense-in-depth fix rather than the root cause of #1474. The hook path's first guard is `[ -n "$AOE_INSTANCE_ID" ] || exit 0`, which silently no-ops when the env var is missing. Our investigation in the issue thread suggests `AOE_INSTANCE_ID` may not actually be propagating to the Hermes Python process at runtime. If that turns out to be the case, the pane detector here masks the symptom while the env-var bug stays latent for any future hook usage. Separate follow-up; not blocking this PR.

Thanks for the patch.

PR #1481, opened 2026-05-24
fix(web): copy selected terminal text with Ctrl-C

MovieHolic-Plex:
## Description
Fixes web terminal copy behavior so `Ctrl+C`/`Cmd+C` copies selected xterm text when a terminal selection exists, while preserving the existing PTY interrupt behavior when nothing is selected.

This addresses the browser-hosted terminal case where the browser default copy action is not reachable while xterm owns the keyboard event. The help overlay now describes the split behavior.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: clipboard writes are browser-permission mediated and the behavior is covered at the xterm hook boundary with Vitest lifecycle tests for selected-copy and no-selection interrupt paths.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Codex

Used to implement and validate a focused web terminal copy regression fix.

**NOTE:**

## Validation

- `npm run test:unit -- src/hooks/useTerminal.lifecycle.test.ts`
- `npx eslint src/hooks/useTerminal.ts src/hooks/useTerminal.lifecycle.test.ts src/components/HelpOverlay.tsx`
- `npm run build`
- `git diff --check`
- commit hook: `cargo clippy -- -D warnings`, `cargo fmt -- --check`

## What Changed

This PR makes Ctrl+C/Cmd+C in the browser-hosted terminal selection-aware while preserving the existing PTY interrupt behavior:

- Selection-aware Ctrl+C: when the xterm terminal has a selection, Ctrl/Cmd+C (and the DOM "copy" event) copy the selected terminal text to the clipboard; when nothing is selected, Ctrl+C still sends the PTY interrupt (SIGINT / 0x03).
- Clipboard helpers: added copyText() (prefers navigator.clipboard.writeText in secure contexts; falls back to a hidden readonly textarea + document.execCommand("copy")) and setClipboardEventText() (writes ClipboardEvent.clipboardData and prevents default).
- Event wiring: added a terminal element "copy" listener and an xterm attachCustomKeyEventHandler that intercepts Ctrl/Meta+C, performs selection-aware copy, prevents default/propagation, and returns false to bypass xterm's default handling when copying occurs.
- Help text: updated HelpOverlay to describe the dual-mode Ctrl+C behavior.
- Hook cleanup and small refactor: removes the term copy listener on unmount; memoizes manualReconnect with useCallback and keeps its ref synchronized via useEffect.
- Tests: extended useTerminal lifecycle tests with a FakeTerminal supporting attachCustomKeyEventHandler, hasSelection/getSelection and a capturable selection string; added tests for DOM "copy", Ctrl+C with selection (asserts execCommand("copy") and no interrupt sent), and Ctrl+C with no selection (asserts interrupt byte forwarded). A Playwright clipboard spec was skipped due to permission constraints.

Removed: nothing — changes are additive and backward-compatible.

## Benefits

- Restores the expected browser copy shortcut for users selecting text in the terminal while preserving existing PTY interrupt semantics.
- Improves usability with minimal surface-area changes and focused unit test coverage.

## Technical details (brief)

- copyText(text): uses navigator.clipboard.writeText in secure contexts (logs async failures); otherwise uses a hidden readonly textarea + document.execCommand("copy") fallback.
- setClipboardEventText(event, text): sets event.clipboardData.setData("text/plain", text) and calls event.preventDefault().
- termEl.addEventListener("copy", onCopy): fills clipboard from term.getSelection() or window.getSelection().
- term.attachCustomKeyEventHandler: on Ctrl/Meta+C and term.hasSelection(), triggers document.execCommand("copy") (with copyText fallback), stops propagation and returns false so xterm does not treat it as a keypress; otherwise Ctrl+C falls through to existing PTY interrupt handling.
- Tests mock navigator.clipboard and WebSocket interactions and add deterministic selection control to validate both copy and interrupt flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1481?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/src/components/HelpOverlay.tsx
+++ b/web/src/components/HelpOverlay.tsx
@@ -4,7 +4,7 @@ interface Props {
 
 const TERMINAL_SHORTCUTS = [
   { key: "All keys", desc: "Relayed directly to the agent via PTY" },
-  { key: "Ctrl+C", desc: "Send interrupt to agent" },
+  { key: "Ctrl+C", desc: "Copy selection, otherwise interrupt agent" },
   { key: "Ctrl+D", desc: "Send EOF to agent" },
   { key: "Up/Down", desc: "Scroll terminal history" },
 ];
--- a/web/src/hooks/useTerminal.lifecycle.test.ts
+++ b/web/src/hooks/useTerminal.lifecycle.test.ts
@@ -30,7 +30,9 @@ const { captured } = vi.hoisted(() => ({
       | ((s: { cols: number; rows: number }) => void)
       | undefined,
     onData: undefined as ((data: string) => void) | undefined,
+    customKey: undefined as ((e: KeyboardEvent) => boolean) | undefined,
     customWheel: undefined as ((e: WheelEvent) => boolean) | undefined,
+    selection: "",
   },
 }));
 
@@ -70,6 +72,15 @@ vi.mock("@xterm/xterm", () => {
     onData(cb: (data: string) => void): { dispose: () => void } {
       captured.onData = cb;
       return { dispose: () => {} };
+    }
+    attachCustomKeyEventHandler(fn: (e: KeyboardEvent) => boolean): void {
+      captured.customKey = fn;
+    }
+    hasSelection(): boolean {
+      return captured.selection.length > 0;
+    }
+    getSelection(): string {
+      return captured.selection;
     }
     attachCustomWheelEventHandler(fn: (e: WheelEvent) => boolean): void {
       captured.customWheel = fn;
@@ -180,7 +191,9 @@ beforeEach(() => {
   captured.writes = [];
   captured.onResize = undefined;
   captured.onData = undefined;
+  captured.customKey = undefined;
   captured.customWheel = undefined;
+  captured.selection = "";
   originalWebSocket = global.WebSocket;
   global.WebSocket = FakeWebSocket as unknown as typeof WebSocket;
   originalResizeObserver = global.ResizeObserver;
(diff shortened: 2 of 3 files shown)


Seluj78:
I haven't (and cannot right now) tested your PR locally but I have doubts about the impossibility of creating a playwright test for this ?

njbrake (commented):

Thanks for the fix @MovieHolic-Plex. The core copy logic and the four Vitest cases are solid. Three things to address:

**1. Backing up @Seluj78's question: yes, a Playwright spec is feasible here.** A mocked spec sketch:
```ts
test.use({ permissions: ['clipboard-read', 'clipboard-write'] });

test('Ctrl+C with selection copies; without selection sends interrupt', async ({ page }) => {
  await page.goto('/...');
  await page.evaluate(() => {
    const term = (window as any).__term; // expose via testing hook, or use a known seed
    term.write('hello world');
    term.selectAll();
  });
  await page.keyboard.press('Control+c');
  const clip = await page.evaluate(() => navigator.clipboard.readText());
  expect(clip).toContain('hello world');
});
```
Even simpler, you can skip the clipboard read entirely and assert via the `__WS_SENT__` pattern (CLAUDE.md's mobile/touch recipe step 3) that Ctrl+C with a selection does NOT push `0x03` over the WebSocket. Per the AGENTS.md mandate, session/terminal flows should land in mocked Playwright with a matching `web/tests/coverage-matrix.json` entry; could you add both?

**2. Unrelated `manualReconnect` refactor.** `web/src/hooks/useTerminal.ts:1370-1407` switches `manualReconnect` from an inline arrow with synchronous ref assignment to `useCallback` + a `useEffect` that updates `manualReconnectRef.current`. That's a refactor that doesn't relate to the copy-on-Ctrl+C feature, and it introduces a subtle behavior shift: the ref now updates after commit instead of during render, so there's a one-cycle window where the ref points at the previous function. Probably benign, but it's untested and conflated with the headline change. Could you split it into its own PR?

**3. `copyTerminalSelection` reports success when the async fallback may silently fail.**
```ts
const copyTerminalSelection = () => {
  const selection = getCopySelection();
  if (!selection) return false;
  try {
    if (document.execCommand("copy")) return true;
  } catch (err) {
    twarn("execCommand copy failed", err);
  }
  void copyText(selection);  // fire-and-forget
  return true;
};
```
If `execCommand` returns `false` AND `navigator.clipboard.writeText` rejects, the user sees no feedback. Either `await copyText(selection)` and return its result, or surface a toast on the rejection path.

Minor: `attachCustomKeyEventHandler` also catches `Cmd+C`, which means we're now reimplementing xterm's native Mac copy. Worth a one-line comment so the next person doesn't wonder why.

CI was gated on first-time-contributor approval; I've kicked it off. Thanks again for picking this up.

Seluj78:
I have fixed it in https://github.com/agent-of-empires/agent-of-empires/pull/1684

PR #1087, opened 2026-05-13
feat(send): auto-wake archived/snoozed rows + remap status on `aoe send`

BTForIT:
## Description

Wires user-interaction signals into the new attention-state model from #1084. After a successful `aoe send`, the target session:

1. Has its `last_accessed_at` stamped via `touch_last_accessed()`. That helper (landed in #1084) also auto-clears `archived_at` and `snoozed_until`, so sending a message to a sunk row wakes it.
2. Has its in-memory status flipped to `Running`. The next status poll reconciles the real state, but the immediate flip keeps the row from sticking on a stale `Idle` / `Waiting` label during the gap between send and poll.

The storage write is best-effort. If it fails AFTER the tmux send succeeded, the agent already received the keys; surfacing this as a "send failed" error would lie about the user-visible outcome. The failure is logged via `tracing::warn!` and the success line still prints.

## What's NOT in this PR

The original branch title (`auto-unarchive on send, respawn dead panes, and remap status`) overclaimed two things:

- **Respawn dead panes** is unchanged from main. `ensure_pane_ready()` already handles dead-pane revive on `aoe send`; that code is on main and not touched here.
- **`build_launch_command` is preparatory only.** This PR extracts the command-building from `start_with_size_opts` into a private helper so a future dead-pane respawn caller could route `tmux respawn-pane` through it, but no such caller is wired up. The helper currently has exactly one call site. The doc reflects that.

A real dead-pane respawn caller is a follow-up, not part of this change.

### Stacking note

Logically depends on #1084 for `touch_last_accessed`'s auto-clear semantics. Now that #1084 / #1085 / #1086 are merged, this PR's incremental diff vs main is the +39 LoC `send.rs` + `instance.rs` change plus three regression tests pinning the `touch_last_accessed` contract.

## PR Type

- [x] New Feature

## Checklist

- [x] I understand the code I am submitting
- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [x] For UI changes: included screenshot or recording. N/A; CLI send-path change.

## Testing

- `cargo build --release`: clean
- `cargo test --lib`: 1857 pass, 0 fail (3 new tests for `touch_last_accessed`)
- `cargo clippy --lib --tests -- -D warnings`: clean
- `cargo fmt --check`: clean

Three regression tests added in `src/session/instance.rs`:
- `test_touch_last_accessed_clears_archived` — archive then touch unarchives
- `test_touch_last_accessed_clears_snooze` — snooze then touch wakes
[...]

PR #1500, opened 2026-05-25
refactor(tui): gate Attention workflow visuals + hotkeys to Attention sort

njbrake:
## Description

Middle-ground response to the design conversation on #1498. Instead of
ripping out the Attention workflow @BTForIT built in #1084 / #1085 /
#1086 / #1087 (the original framing of this PR), the workflow stays in
place; its surface is gated to only activate when the sidebar is sorted
by Attention.

### What changes

- **Favorite (`f` / `F`), snooze (`h` / `H`, `w` / `W`), and urgent
  reads** are gated to `sort_order == SortOrder::Attention`. In every
  other sort mode the bindings decline and the underlying fallbacks fire
  (`h` → collapse, `w` → jump-to-next-waiting, capital keys → strict-
  mode typing guard).
- **Visual decoration is gated too**: the `* ` / `z ` / `! ` prefixes,
  bold+underline (favorite), italic+dim (snooze), and RAPID_BLINK
  (urgent) only paint in Attention sort. In Newest / Recent / Oldest /
  AZ / ZA the row paints with its plain title and status-driven color.
- **Archive stays universal** because it's a lifecycle action (the pane
  is killed, status is stale); archived rows paint `ICON_STOPPED` + dim
  in every sort.
- **Synthetic "Archived" sidebar section** pinned to the bottom across
  every sort mode (sentinel group path `__aoe_archived_section__`,
  collapsed by default, persisted in
  `app_state.archived_section_collapsed`). Archived rows are pulled out
  of the natural flow and grouped here, replacing the
  tier-99-for-archived sink in Attention mode and giving every other
  sort a stable shelved-row destination. Snoozed and pane-dead rows
  still sink to tier 99 inline in Attention sort because they're
  transient sinks, not lifecycle terminals.
- **Help overlay** relabels the section to
  "Attention (Attention sort only, except Archive)" and tags each
  binding with the sort mode where it works.

### Why this shape instead of removal

@BTForIT's writeup on #1498 made the case for keeping the workflow at
high session counts: a 36-session fleet across 7 profiles is the regime
where flat newest-first becomes scroll-and-hunt and the tier model
becomes load-bearing. Ripping the workflow would push the maintenance
into a downstream fork. Gating moves the cost from "implicit on every
render PR" to "explicit Attention-mode test surface," which is a much
better trade.

For users (most of us) who don't run a fleet that size, the default
sidebar is now plain: no favorites, no snooze badges, no urgent blink.
Cycling to Attention sort lights the whole thing up.

### Test coverage

Three new unit tests pin the gates:

- `favorite_decoration_gated_to_attention_sort`: favorite renders `* `
  + bold/underline in Attention, plain in Newest.
- `snooze_decoration_gated_to_attention_sort`: snooze renders `z `
  prefix in Attention, plain in Newest.
- `archived_section_pinned_to_bottom_in_every_sort`: synthetic Archived
  section appears at the bottom of `flat_items` across `Newest`,
  `Attention`, and `AZ`, with the archived row pulled out of the natural
  flow.

Plus `test_non_strict_h_snoozes_only_in_attention_sort` covers the
keybind gate end-to-end (h in Attention opens the snooze dialog; h in
Newest collapses the group via the fallthrough).

Existing tests still pass: 2223 lib + 120 integration.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (2223 lib + 120 integration green;
  `cargo fmt --check`, `cargo clippy --all-targets`,
  `cargo check --features serve` all clean)
- [x] Documentation was updated where necessary (help overlay section
  retitled; bindings tagged with their sort-mode scope)

The pre-existing `attention-demo.mp4` in this thread already shows the
"everything is on" state. A "default sort" capture showing the plain
sidebar is straightforward to add if reviewers want one alongside.

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow. The
  Attention workflow is TUI-only; no web sidebar code is touched.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the gate surface is exercised by 4 new
  unit tests in `src/tui/home/tests.rs` covering the bindings (h-in-
  Attention vs h-outside), the favorite decoration, the snooze
  decoration, and the Archived section placement. An e2e is reasonable
  follow-up once @BTForIT's asciinema lands and we have a clear flow
  to replay, but the unit coverage already pins the behavioral
  contract.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 via Claude Code.

**Any Additional AI Details you'd like to share:** The implementation,
test coverage, and this PR body were drafted by Claude through
back-and-forth with @njbrake after @BTForIT pushed back on the original
removal approach. The middle-ground design (gate instead of remove) and
every scope call are @njbrake's; the labor and the prose came out of
the agent loop.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1500?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -509,6 +509,15 @@ pub struct AppStateConfig {
     /// Restored on subsequent opens so users don't re-navigate every time.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub last_browse_dir: Option<PathBuf>,
+
+    /// Collapsed state for the synthetic "Archived" sidebar section.
+    /// Defaults to collapsed when absent. Archived sessions are pulled out
+    /// of the natural sort and grouped under this section at the bottom
+    /// across every sort mode, so they stop interleaving with active rows
+    /// without users in non-Attention modes losing the ability to find a
+    /// shelved session.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub archived_section_collapsed: Option<bool>,
 }
 
 /// Session-related configuration defaults
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -8,6 +8,20 @@ use std::collections::HashMap;
 use super::config::SortOrder;
 use super::Instance;
 
+/// Sentinel path used by the synthetic "Archived" sidebar section. Not a
+/// real GroupTree entry; lives only in the flattened `Item` list to give
+/// archived sessions a stable bottom-of-sidebar home across every sort
+/// mode. Code that walks `flat_items` and dispatches on `Item::Group.path`
+/// must skip this path before invoking GroupTree-mutating ops (rename,
+/// delete, archive, etc.) since no matching group exists.
+pub const ARCHIVED_SECTION_PATH: &str = "__aoe_archived_section__";
+pub const ARCHIVED_SECTION_NAME: &str = "Archived";
+
+#[inline]
+pub fn is_archived_section_path(path: &str) -> bool {
+    path == ARCHIVED_SECTION_PATH
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Group {
     pub name: String,
@@ -388,7 +402,7 @@ fn max_created_at_in_group(path: &str, instances: &[Instance]) -> DateTime<Utc> 
     let prefix = format!("{}/", path);
     instances
         .iter()
-        .filter(|i| i.group_path == path || i.group_path.starts_with(&prefix))
+        .filter(|i| (i.group_path == path || i.group_path.starts_with(&prefix)) && !i.is_archived())
         .map(|i| i.created_at)
         .max()
         .unwrap_or(DateTime::<Utc>::MIN_UTC)
@@ -400,7 +414,7 @@ fn min_created_at_in_group(path: &str, instances: &[Instance]) -> DateTime<Utc> 
     let prefix = format!("{}/", path);
     instances
         .iter()
-        .filter(|i| i.group_path == path || i.group_path.starts_with(&prefix))
+        .filter(|i| (i.group_path == path || i.group_path.starts_with(&prefix)) && !i.is_archived())
         .map(|i| i.created_at)
         .min()
         .unwrap_or(DateTime::<Utc>::MAX_UTC)
@@ -413,7 +427,7 @@ fn max_last_accessed_in_group(path: &str, instances: &[Instance]) -> Option<Date
     let prefix = format!("{}/", path);
     instances
         .iter()
-        .filter(|i| i.group_path == path || i.group_path.starts_with(&prefix))
+        .filter(|i| (i.group_path == path || i.group_path.starts_with(&prefix)) && !i.is_archived())
         .filter_map(|i| i.last_accessed_at)
         .max()
 }
@@ -649,10 +663,12 @@ pub fn flatten_tree_all_profiles(
 ) -> Vec<Item> {
     let mut items = Vec::new();
 
-    // Collect all ungrouped sessions across all profiles
+    // Archived sessions are excluded from the natural flow. The caller
+    // appends them under the synthetic "Archived" section via
+    // `append_archived_section`.
     let mut ungrouped: Vec<&Instance> = instances
         .iter()
-        .filter(|i| i.group_path.is_empty())
+        .filter(|i| i.group_path.is_empty() && !i.is_archived())
         .collect();
 
     sort_sessions(&mut ungrouped, sort_order);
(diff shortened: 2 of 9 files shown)


BTForIT:
@njbrake This is exactly the shape I was hoping for — thank you for finding a middle ground instead of removing it. Reading through the diff:

**On the gate design** — clean. Routing the favorite/snooze keybinds and the visual decorations through `sort_order == SortOrder::Attention` keeps the implementation surface flat (one branch per binding, one branch per render predicate) without making the underlying state model any more complex. The snooze-timer-still-ticks-outside-Attention call is the right one: the timer is conceptually about "when can this row re-surface," not "is it currently visible." Decoupling those lets a user park a session in Attention, switch to AZ to look something up, and come back to Attention with the snooze having already lapsed.

**On the synthetic Archived section** — also clean, and I think actually an improvement over the old tier-99-inline behavior. The defense-in-depth pattern (sentinel path, palette skips it, group ops disarm via `is_archived_section_path`) is the right way to handle the "this isn't a real GroupTree node" reality. Pulling archived rows out of the natural flow in every sort, including Attention, means archive becomes a single coherent lifecycle move regardless of which sort the user happens to be in. That's strictly better than the old behavior where archived rows would be visible inline in Attention but interleaved-by-activity in Newest.

One question to confirm I'm reading the interaction right: does the Archived section header still show its session count when collapsed? I see `session_count: archived.len()` in `append_archived_section`, but want to confirm the row formatter renders that count regardless of `collapsed`. The user-value here is "I want to know I have 12 archived sessions without expanding the list" — the empty/non-empty distinction (no header at zero) is right, but I'd lose value if the count only appeared on expand.

**Interaction with #1498 (`RowTagMode::Project`):** the `Project` tag still renders on archived rows under the new section, right? Just want to confirm the synthetic section header doesn't strip per-row tags from its children. From the diff I see archived rows still go through the normal `Item::Session` path with `depth: 1`, so the existing row formatter should still pad the tag and align the activity column — but worth a screenshot in the asciinema since the `[per-dev]` tag on an archived row in the Archived section is a new visual combo we haven't seen.

**On the asciinema** — still on. I had a delay on this side getting `asciinema` installed on the headless box (consent gate on the install ceremony — sysadmin-side annoyance, not Claude-side). Going to push through that today and capture once I have it. I'll show:
1. Attention sort with all the decorations on (favorite stars, snooze badges, urgent blink, the works).
2. Cycling to Newest and watching everything paint plain — exactly the visual demonstration of the gate.
3. Archive a session in Newest, cycle back to Attention, show the row pinned to the bottom under the new Archived section in both views.
4. The `Project` row tag from #1498 in the mix on a 36-session multi-repo fleet.

Net: thank you for putting the labor into this rather than removing it. Approving in spirit, will leave the formal approval once I've seen the count-on-collapse behavior and posted the recording.

njbrake:
@BTForIT thanks! Yes want to be sure you know I do appreciate all the work you put in on the codebase and hope we can continue to refine this feature as we keep going forwards. Thanks!

njbrake:

@BTForIT, both confirmed.

**Count on collapsed**: yes. `append_archived_section` always sets `session_count: archived.len()` and the header renderer formats every group as `"{name} ({session_count})"` with no branch on `collapsed`. You'll see `▶ Archived (12)` collapsed, `▼ Archived (12)` expanded.

**`RowTagMode::Project` on archived rows**: yes. Archived sessions go through the same `Item::Session` render path as everything else, so `compute_row_tag` fires and the `[per-dev]` / `[for-AvOps]` tag appears as usual.

One thing to flag proactively: commit `53ed9f3d` auto-clears the snooze flag when you press `e` on a snoozed row outside Attention sort. The reasoning was that with snooze decoration gated off, a silent-restart-drop on what looks like a plain Running row reads as a UI bug. But that's a behavior I invented, not derived from your original design, and it does contradict "snooze persists until timer or manual wake." If it bites in the asciinema, say the word and I'll revert it and substitute a toast or confirm-pattern.

BTForIT:
Cross-linking — recorded a demo of the live fleet for #1498 that also visualizes what this PR is solving: **https://asciinema.org/a/cyneg9RkwasNbS8Z** (4:48).

Watch around the 1:10–2:30 segment where `O` cycles through Recent → Oldest → A-Z → Z-A → Newest → back to Attention. In every non-Attention mode the favorite pins and snooze sinks become structural noise — the sort criterion is something else, so the within-tier biases stop telling the user anything useful. The cast makes the case for your gating change pretty clearly.

(Recording auto-deletes from asciinema.org in ~7 days since my upload CLI isn't linked — ping me and I'll re-host if needed.)

BTForIT:
Durable raw-`.cast` backup for the demo above: https://gist.github.com/BTForIT/cd610aad3725a50b4b346476693cfd0b

PR #1507, opened 2026-05-26
perf(tmux): debounce trailing-edge %output wakes in live-send

njbrake:
## Description

Fixes the occasional "cursor jumps after a brief lag" jitter in live-send mode. Root cause turned out to be neither the wake channel capacity, the on-UI-thread disk reload, the shared-mutex contention between capture-pane and send-keys, nor the ansi-to-tui parse cost. All four were sub-millisecond in isolation.

The real cause: tmux emits one `%output` notification per line the agent prints. A multi-line burst (a token-stream chunk, a full-screen TUI repaint) produces dozens of `%output` lines across a few milliseconds. The reader thread fired the wake callback on every one of those lines. The main loop captured the pane on the FIRST output of a burst, painting a partial frame, then captured again on a later wake and jumped to the settled state. Smooth, then sudden jump.

This commit moves the wake callback off the reader thread's per-`%output` hot path and onto a dedicated debouncer thread that fires the callback once after the trailing edge of a burst. 8ms debounce window, picked empirically. The 50ms tokio ticker still bounds worst-case latency under continuous sustained output (where the debounce keeps extending).

### How I found it

Built a four-test microbench harness on a throwaway diagnostic branch that isolated each suspect with hard numbers. The decisive one (`bench_burst_capture_freshness`):

| | captures until BURST_END seen | wake → BURST_END visible (ms) |
|---|---|---|
| No debounce | mean=1.7, max=2.0 | mean=5.3, max=8.7 |
| 8ms debounce | mean=1.0, max=1.0 | mean=10.3, max=13.5 |

The "no debounce" row shows a partial-then-complete two-frame sequence; the 8ms row collapses to a single complete frame. The +5ms uniform latency is well under human perception threshold.

The diagnostic branch (with env-gated bisect knobs and the bench harness) is `perf-bisect-experiments`, kept locally as a record; nothing from it ships here.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added three unit tests in `src/tmux/control_mode.rs`:
- `debouncer_fires_once_per_burst`: 20 back-to-back ticks produce 1 callback, not 20.
- `debouncer_handles_quiet_periods_between_bursts`: two bursts separated by >debounce produce 2 callbacks.
- `debouncer_exits_on_sender_disconnect`: thread shuts itself down when its sender drops (no leaked threads on `ControlModeClient::drop`).

Added one live-fire integration test in `tests/integration/tmux_control_mode.rs`:
- `control_mode_output_wake_debounces_bursts`: runs a 10-line printf burst against a real tmux session, asserts 1-3 wakes. Catches both "wake never fires" (regression) and "wake fires per line" (regression to pre-debounce).

The existing `control_mode_output_wake_fires_on_pane_output` still passes (its `observed > baseline` assertion is weaker than the new test's bounded range).

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 (1M context) via Claude Code

**Any Additional AI Details:**

Claude built the diagnostic bench harness to isolate each suspect with measurements, ran the benches, ruled out three of the four suspects with hard numbers, identified the burst-freshness root cause, drafted the fix, and wrote the tests. I (njbrake) directed the investigation, chose which bisects to run, picked the trailing-edge debounce shape over alternatives, and reviewed the diff before this PR.

### High-Level Overview

The PR reorganizes how the tmux control mode client handles output notifications to fix intermittent cursor lag in live-send mode. Instead of triggering a wake callback for every single line of output, the system now batches these notifications and fires the callback just once after a burst of output settles.

**What was added:**
- A new debouncer thread that consolidates multiple rapid output events into single wake events
- An 8-millisecond debounce window to determine when a burst has finished
- Three unit tests validating debouncer behavior (burst coalescence, burst separation, thread cleanup)
- One integration test running against a real tmux session to verify the debouncing works end-to-end

**What was changed:**
- Reader thread behavior: instead of calling the output callback directly, it now sends tick signals to the debouncer
- Debouncer logic: waits for quiet periods before firing the callback, resetting the timer on each new output event

**What was preserved:**
- Public API signatures remain unchanged
- Existing 50ms tokio ticker remains as a safety bound
- Optional behavior: debouncer thread only spawns if a callback is actually provided

### Benefits

This change eliminates the "cursor jump" issue by reducing unnecessary wake-ups. For bursty output (like multi-line printf loops), wake counts drop from dozens per burst to just 1–3. The main loop now captures complete frames instead of jumping between partial frames during output bursts. Performance measurements show the debounced model achieves this while maintaining acceptable latency (10–13ms from burst end to wake, well within typical UI refresh timings).

### Technical Details

The implementation uses a dedicated `debouncer_loop` function that monitors a channel for output ticks. On the first tick, it blocks waiting for silence. Once a timeout occurs (no new ticks within 8ms), the callback fires and the debouncer waits for the next burst. The reader thread signals ticks before processing lines, ensuring debouncer latency doesn't compound with line processing time. The debouncer thread cleans up automatically when the reader exits by detecting channel disconnection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1507?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/tmux/control_mode.rs
+++ b/src/tmux/control_mode.rs
@@ -37,6 +37,25 @@ use std::time::{Duration, Instant};
 
 use anyhow::{bail, Context, Result};
 
+/// How long after the last `%output` notification we wait before firing
+/// the wake callback to the main loop. tmux emits one `%output` per
+/// line the agent prints, so a multi-line burst (a token-stream chunk,
+/// a full-screen TUI repaint) generates dozens of `%output` lines
+/// across a few milliseconds. Firing the wake on the FIRST output
+/// makes the consumer capture-pane mid-burst, drawing a partial
+/// frame; a later wake then redraws the settled state. The user sees
+/// "smooth, then sudden jump after a brief lag".
+///
+/// Debouncing collapses the burst to one wake at the trailing edge,
+/// so the very first frame the consumer paints is already complete.
+/// 8ms was picked empirically (see PR description): smaller values
+/// (4ms) start letting partial frames through on dense bursts; larger
+/// values (16ms+) start adding visible latency to single-character
+/// echoes. The 50ms tokio ticker still bounds worst-case latency
+/// under continuous sustained output (>1 output per 8ms) where the
+/// debounce keeps extending.
+const OUTPUT_WAKE_DEBOUNCE: Duration = Duration::from_millis(8);
+
 /// Default timeout waiting for a single command response.
 ///
 /// The previous value (750ms) was too tight in practice: under fast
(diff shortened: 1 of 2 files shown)


njbrake:
Closing — the trailing-edge debounce was solving the wrong problem and the +8ms uniform delay regressed perceived typing latency. Bench harness on a sandbox didn't match real-world load; reproducing under SSH against a real agent (claude-code) shows the actual culprits are different. Investigation continuing on the perf-bisect-experiments branch; next PR will address both the cursor jitter and the 'Enter freezes preview' cascade together.

PR #1521, opened 2026-05-26
fix(tui): keep live-mode entry frame from rendering shifted up

njbrake:
## Description

When entering live mode (Tab on a session row), the first frame would sometimes render with the agent's output anchored too high in the preview pane. Exiting and re-entering live mode "fixed" it. Two compounding causes, fixed together.

**1. Toast-row geometry mismatch.** `Action::EnterLiveSend` sets a `"Reviving session..."` transient toast and calls `draw()` before invoking `enter_live_send`. That toast adds a bottom status row, shrinking the preview by one row. The sync `tmux resize-window` inside `enter_live_send` then read `self.preview_area` from the toast frame and targeted geometry that didn't match what the user would see one frame later, after the toast cleared. Intermittency tracks whether the persistent update-available banner was already keeping `has_update_bar = true` (which masks the effect).

Split `enter_live_send` into `prepare_live_send` (slow path: `ensure_pane_ready`, install state, spawn worker) and `finalize_live_send_resize` (sync tmux resize + 50ms SIGWINCH-settle sleep). The handler now settles the toast to its final state (either `"Resume failed..."` or `None`) and redraws between the two halves, so the resize geometry matches the post-toast frame the user actually sees for the next several frames.

**2. Info-panel sizing.** `refresh_preview_cache_if_needed` and the sync resize both sized the tmux pane to the full preview inner. `Preview::render_with_cache` splits that inner into an info header + output sub-rect when the user has the info panel expanded (Agent view, non-compact). The agent rendered `info_height` more rows than were visible; tail-anchored `Paragraph::scroll` then dropped those rows off the top of every frame, making content look shifted up even after entry settled.

Add a `preview_pane_area` field that mirrors the info-vs-output split, factor `agent_info_height` into a module-level helper in `preview.rs`, and route both the worker resize and the sync resize through the new sub-rect.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test --lib`: 2214 passed)
- [x] Documentation was updated where necessary (in-code docstrings on `prepare_live_send`, `finalize_live_send_resize`, `preview_pane_area`, `agent_info_height`)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the visible symptom requires asserting that the agent's bottom row lands on the preview's last row across two successive frames immediately after `EnterLiveSend`. The existing `TuiTestHarness` captures screen text but doesn't expose per-frame geometry or cell-row anchoring, and the shift on a real agent depends on `info_height` + `agent_info_height` returning the same values the renderer uses. Happy to add the harness plumbing in a follow-up if reviewers want it.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Pair-debugged the root cause (toast-row mismatch + info-panel sizing) and refactored `enter_live_send` into `prepare`/`finalize` halves. Verified locally with `cargo fmt --check`, `cargo clippy --tests`, and `cargo test --lib`.

## Overview
Fixes an intermittent TUI bug where the first frame after entering live mode could render the agent output shifted upward. Two root causes were addressed: a toast/resize timing mismatch during live-entry, and incorrect preview sizing when the info panel is expanded.

## What changed (high level)
- Split live-send startup into two steps:
  - prepare_live_send: install state and spawn the live-send worker (no synchronous resize).
  - finalize_live_send_resize: perform synchronous tmux resize and settle layout before capturing frames.
- Redrew the UI between those steps so the synchronous resize targets the post-toast layout the user actually sees.
- Added HomeView.preview_pane_area to track the visible output sub-rectangle (excluding the info header) separately from the full preview area.
- Centralized the info-header sizing formula into a new agent_info_height(instance) helper and replaced duplicated logic.
- Updated preview rendering/cache refresh and live-send resize flows to use the output sub-rect (preview_pane_area) so captures match what the user sees.
- Fixed live-mode footer scroll indicator math (adjust visible height calculation) and added unit tests covering agent_info_height branches.
- Updated docstrings/comments/tests; one e2e TUI test was skipped due to harness limitations.

## Benefits
- Eliminates the visual frame-shift by ensuring resize occurs against the actual visible layout.
- Prevents off-screen rendering/clipping when the info panel is expanded.
- Improves reliability by waiting ~50ms for SIGWINCH/layout to settle before capture.
- Reduces duplicated sizing logic and makes live-send startup responsibilities clearer.
- Adds tests to prevent regressions in info-header sizing and footer indicator math.

## Technical notes (concise)
- API/struct changes:
  - enter_live_send renamed/split into prepare_live_send(&mut, session_id) -> Result<Option<String>, ()> and finalize_live_send_resize(&mut).
  - Added pub(super) preview_pane_area: Rect to HomeView.
  - Added pub fn agent_info_height(instance: &Instance) -> u16 in preview module.
- Resize behavior: finalize_live_send_resize synchronously calls tmux resize-window, updates live_send_last_resize only on success, then blocks ~50ms (tokio::task::block_in_place) to let SIGWINCH/layout settle.
- Tests: cargo test --lib passes (2214); an e2e TUI test was intentionally skipped. AI-assisted coding (Claude) is noted in commits.

[Review Change Stack](https://app.coderabbit.ai/change-stack/njbrake/agent-of-empires/pull/1521?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -1629,21 +1629,27 @@ impl App {
                     .set_instance_status(&id, crate::session::Status::Starting);
                 self.update_status = Some(UpdateStatus::transient("Reviving session...".into()));
                 self.draw(terminal)?;
-                let outcome = self.home.enter_live_send(&id);
-                match outcome {
-                    Ok(Some(sid)) => {
-                        self.update_status = Some(UpdateStatus::transient(format!(
-                            "Resume failed for sid {sid}; live-send sent to a fresh pane (history not loaded)"
-                        )));
-                    }
-                    Ok(None) => {
-                        self.update_status = None;
-                    }
-                    Err(()) => {
-                        // enter_live_send already surfaced the failure via
-                        // info_dialog; just clear the transient toast.
-                        self.update_status = None;
-                    }
+                let outcome = self.home.prepare_live_send(&id);
+                // Settle the toast to its final state BEFORE the sync resize
+                // and redraw, so HomeView's cached `preview_pane_area`
+                // matches the geometry the user will see for the next
+                // several frames. Otherwise the toast row that was on screen
+                // during `prepare_live_send` would make the preview pane one
+                // row shorter than post-toast, the sync resize would target
+                // the smaller pane, and the first capture would render
+                // shifted up.
+                self.update_status = match &outcome {
+                    Ok(Some(sid)) => Some(UpdateStatus::transient(format!(
+                        "Resume failed for sid {sid}; live-send sent to a fresh pane (history not loaded)"
+                    ))),
+                    // On clean ready, drop the toast entirely. On Err the
+                    // info_dialog already carries the failure detail, so the
+                    // transient toast just gets in the way.
+                    Ok(None) | Err(()) => None,
+                };
+                if outcome.is_ok() {
+                    self.draw(terminal)?;
+                    self.home.finalize_live_send_resize();
                 }
             }
             Action::AttachToolSession(id, tool_name) => {
--- a/src/tui/components/preview.rs
+++ b/src/tui/components/preview.rs
@@ -599,4 +599,80 @@ mod tests {
             Some(" [80/80] ".to_string())
         );
     }
+
+    // `agent_info_height` drives both the preview layout split in
+    // `render_with_cache` and the live-send sync resize in
+    // `HomeView::finalize_live_send_resize`. A one-row drift here brings
+    // the shifted-preview bug right back, so each branch of the formula
+    // gets a dedicated case.
+    mod agent_info_height {
+        use super::super::agent_info_height;
+        use crate::session::{Instance, SandboxInfo, WorktreeInfo};
+        use chrono::Utc;
+
+        fn worktree(base_branch: Option<&str>) -> WorktreeInfo {
+            WorktreeInfo {
+                branch: "feature/x".into(),
+                main_repo_path: "/repo".into(),
+                managed_by_aoe: true,
+                created_at: Utc::now(),
+                base_branch: base_branch.map(str::to_string),
+            }
+        }
+
+        fn enabled_sandbox() -> SandboxInfo {
+            SandboxInfo {
+                enabled: true,
+                container_id: None,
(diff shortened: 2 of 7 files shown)


njbrake (commented):

Self-review verdict (GitHub blocks self-approval, so posting as a comment): would approve.

Two real bugs fixed:

1. The "Reviving session..." transient toast added a status row that shrank the preview by one, so the sync `tmux resize-window` inside `enter_live_send` read `preview_area` from the toast frame and targeted a geometry one row shorter than the post-toast frame. Split into `prepare_live_send` (slow path, install state, spawn worker) and `finalize_live_send_resize` (sync resize plus 50ms SIGWINCH sleep), with the action handler settling the toast to its final state and redrawing in between.

2. The cache + resize sized the tmux pane to the full preview inner, ignoring the Agent-view info header. With the info panel expanded the agent rendered `info_height` more rows than were visible; tail-anchored `Paragraph::scroll` then dropped those rows off the top of every frame. New `preview_pane_area` field + `agent_info_height` helper route both the worker and the sync resize through the visible output sub-rect.

Local verification: `cargo fmt --check`, `cargo clippy --tests` (no new warnings), `cargo test --lib` (2214 passed).

Non-blocking follow-ups:

1. Entering live mode while in Terminal or Tool view still sizes the agent pane to the full inner. The new sub-rect is only refined inside the Agent branch of `render_preview`, so the corner case where the user Tabs from Terminal view and later switches back to Agent view with the info panel expanded keeps the original clip. Pre-existing behavior, not made worse by this PR.

2. The `Ok(Some(sid))` resume-failed path keeps the one-row shift when the "Resume failed" transient toast eventually expires, since the resize is sized with that toast on screen. Niche path. The user has read the toast by then.

3. An e2e test that asserts the agent's bottom row lands on the preview's last row across two successive frames after `EnterLiveSend` would be the right regression guard, once `TuiTestHarness` exposes per-frame row geometry.

PR #1523, opened 2026-05-27
fix(session): anchor Claude session poller to its own session id (#1522)

### LINKED ISSUE #1522: Shared project_path: session_id_poller cross-assigns one agent_session_id to all sessions, clobbering per-session --resume
## Summary

When multiple sessions share the same `project_path`, launching the TUI causes the `session_id_poller` to assign **all of them the same `agent_session_id`** — whichever agent artifact (e.g. Claude `*.jsonl`) is most-recently-modified in that shared path. This silently clobbers each session's individually-configured `--resume` target, so every session resumes the *same* conversation. For a fleet of N sessions rooted at one repo, N-1 sessions lose their real history on the next launch/restart.

PR #1250 ("defense in depth for resume-fallback cascade races") mitigated some races but was explicitly scoped as narrow/defense-in-depth; the underlying "can't disambiguate which artifact belongs to which pane when project_path is shared" problem remains in 1.8.1 (and, by code inspection, still in 1.9.x).

## Environment

- aoe `1.8.1` (nix flake install), but the relevant code path is unchanged on `main`/1.9.x
- Linux, tmux backend, agent = `claude`
- 19 sessions, **all** with `project_path = /…/financialreports` (a single repo driving a fleet)

## Repro

1. Create 2+ aoe sessions in the **same** `project_path`, each launched with its own `--resume <distinct-session-id>` (via `extra_args` or `set-session-id`).
2. Ensure each has a distinct on-disk agent artifact (distinct Claude `*.jsonl` under `~/.claude/projects/<encoded-path>/`).
3. Launch the `aoe` TUI.
4. Observe `debug.log`: within ~1s of `aoe started`, every session emits `session.store: Session ID changed for <id>: <SAME-uuid>`.
5. Restart any session → it launches with the clobbered id and resumes the wrong conversation.

## Evidence

On a single TUI launch, all 17 live sessions were reassigned to one id in ~100ms:

```
2026-05-27T04:01:24.986  INFO log.runtime: aoe started version="1.8.1"
2026-05-27T04:01:25.361  INFO session.store: Session ID changed for 52413c4ad9b54092: 51e26c0a-…
2026-05-27T04:01:25.368  INFO session.store: Session ID changed for d05701bb0cba424e: 51e26c0a-…
2026-05-27T04:01:25.374  INFO session.store: Session ID changed for 5addf5ad91374799: 51e26c0a-…
…(14 more, all → 51e26c0a-…)…
```

A prior crash-recovery launch shows the same pattern collapsing everything to a different single id (`d024562f-…`), so the target id is just "whatever artifact won the most-recent race," not anything session-specific.

Result in the launched process (note the second, clobbering `--resume`):

```
claude --resume 67540a5b-… --dangerously-skip-permissions --dangerously-skip-permissions --resume 51e26c0a-…
        ^ configured target                                                                  ^ poller-injected, wrong
```

## Root cause (best understanding)

- `src/session/instance.rs` starts a per-session `session_id_poller` whose `poll_fn` detects "current agent session id" by reading **on-disk artifacts** in the project path.
- When sessions share `project_path`, the poller cannot attribute a given `*.jsonl` to a specific pane, so they all resolve to the most-recently-touched artifact.
- Per PR #1250's own description, the `retroactive_capture_excludes` guard (the set tracking stale ids the cascade just cleared) is `#[serde(skip)]`, so it's dropped on every 5s `HomeView::reload()`, letting the poller re-import the bad id repeatedly.

## Suggested fix (directions, not prescriptive)

The poller should attribute artifacts to panes deterministically rather than by recency in a shared dir. Options:

1. **Trust the launch contract**: a session launched with `--resume <X>` (or whose `agent_session_id` is set) should have its poller *anchored* to `X` — only adopt a new id if it can be positively attributed to *this* pane (e.g. the artifact's recorded cwd + pid/ppid lineage, or a pane-scoped marker), never by bare project-dir recency.
2. **Persist `retroactive_capture_excludes`** (drop the `#[serde(skip)]`) so cleared-bad ids survive reloads.
3. **Refuse cross-attribution**: if two live sessions would resolve to the same artifact id, decline to assign rather than assigning it to all.

Happy to test a patch against this fleet (19 same-repo sessions) if useful — it's a reliable reproduction.

## Impact

High for multi-session/single-repo fleets: a single TUI launch silently detaches every session from its real conversation, and a restart bakes it in. Recovery requires reconstructing the per-session→transcript mapping by hand. Related: #1498 (RowTagMode::Project) acknowledges many-same-repo fleets exist but addresses display, not this derive collision.

itisaevalex:
## Summary

Fixes #1522.

When several AoE sessions share one `project_path` (a fleet rooted at a single repo), the Claude session-id poller cross-assigns one `agent_session_id` to **all** of them. `capture_claude_session_id` returns the most-recently-modified `.jsonl` in `~/.claude/projects/{encoded-path}/`, keyed only on `project_path` — so every session's poller resolves to whichever session wrote last. Opening the TUI then fires `Session ID changed for <every session>: <same uuid>`, clobbering each session's real `--resume` target. On the next restart every session resumes the *same* conversation.

## Fix

Anchor each poller to its own session. `claude_poll_fn` now receives the instance's known session id (`agent_session_id`, already computed as `initial_known` at the poller-spawn site) and threads it into `capture_claude_session_id` → `scan_claude_project_dir`. The scan now:

- returns the caller's **own** `<known>.jsonl` when that file is still fresh (≤5 min) — so a session sticks to its own transcript instead of racing for the global most-recent;
- falls back to the most-recent file when `known` is `None`, absent, or stale (e.g. the session genuinely moved to a new conversation).

This is deliberately minimal and preserves existing behaviour for the single-session-per-repo case: there, the known id *is* the most-recent file, so the result is unchanged.

### Why anchor rather than exclude?

`claude_poll_fn`'s doc comment notes the `extra_excludes` pattern used by other agents. An excludes-based fix would need a live set of *all other sessions'* ids threaded into each closure (cross-session state + locking). Anchoring needs only this instance's own id, which is already in scope — smaller surface, no new shared state. Happy to switch approaches if you'd prefer the excludes route.

## Tests

Added to `src/session/capture.rs`:

- `test_capture_claude_session_prefers_known_when_shared` — two fresh `.jsonl` files in one project dir; asserts `None` → most-recent (unchanged), `Some(A)` → A even though B is newer, `Some(B)` → B, `Some(absent)` → most-recent.
- `test_capture_claude_session_known_but_stale_falls_back` — a stale known file falls back to the fresh most-recent.

Existing `capture` tests updated for the new signature; `fmt` / `clippy` / `test` run locally via the flake devshell.

## Scope / follow-up

This fixes the **host** path (`capture_claude_session_id`). The sandboxed path (`capture_claude_session_id_in_container`, `ls -t | head -1` inside the container) has the identical root cause; I left it out to keep this change focused and unit-testable, but I'm happy to mirror the anchor there in a follow-up (or this PR if you prefer).

## Caveat

If a session legitimately forks to a new id (`--fork-session`) while its old `.jsonl` is still fresh, detection of the new id lags until the old file passes the 5-min staleness bound. This is acceptable for fixing the cascade and only affects fork-on-resume; documented inline.

## What Changed

This PR fixes a bug where multiple Claude sessions targeting the same repository could all adopt the same session ID (losing distinct resume targets) because the poller always picked the most-recent transcript file. The poller is now anchored to each session's own known session id and will prefer that session's transcript if it's fresh (≤5 minutes); if the anchored file is missing or stale, it falls back to the most-recent file. The sandboxed/container capture path was not changed.

## What Moved / Added / Removed

Added / Updated:
- capture anchoring: capture_claude_session_id now accepts an optional known_session_id and prefers the caller's .jsonl when fresh.
- poller anchoring: claude_poll_fn now accepts known_session_id (wrapped in a Mutex) and promotes the last-accepted session id so a poller stays attached to its session for its lifetime.
- Instance wiring: Instance::maybe_start_poller now passes the session's initial_known into claude_poll_fn.
- Tests: New unit tests in src/session/capture.rs verifying that a session prefers its known transcript when shared and falls back when the known transcript is stale.

Removed / Unchanged:
- No behavioral changes to the sandboxed/container capture path (left for follow-up).
- Existing 5-minute staleness gate and lastSessionId fallback remain.

## Benefits

- Restores correct per-session resume behavior in multi-session-per-repo setups.
- Prevents inadvertent reassignment of session histories across panes/sessions.
- Preserves single-session-per-repo behavior and provides a graceful fallback when anchors are stale.

## Technical Details (brief)

- New signature: capture_claude_session_id(project_path: &str, known_session_id: Option<&str>) -> Result<String>
- New signature: claude_poll_fn(project_path: String, known_session_id: Option<String>) -> impl Fn() -> Option<String> + Send + 'static
- Poller stores its anchor in a Mutex<Option<String>> and updates it after successful captures so the poller remains attached until a fallback is required.
- Staleness threshold: anchored transcript considered fresh if mtime ≤ 5 minutes; otherwise fallback to most-recent transcript.
- Caveat: if a session forks while its old transcript is still fresh, detection of the new id can lag until staleness expires.

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1523?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -243,13 +243,28 @@ pub(crate) fn claude_poll_fn(
     project_path: String,
     known_session_id: Option<String>,
 ) -> impl Fn() -> Option<String> + Send + 'static {
+    // issue #1522: promote the last successfully captured sid to the anchor
+    // for subsequent polls. Without this, the closure keeps the startup
+    // `known_session_id` forever; once that file goes stale the scan drops
+    // back to global-most-recent and shared-repo fleets can resume
+    // cross-assigning on the next /clear or fork.
+    let last_known = std::sync::Mutex::new(known_session_id);
     move || {
-        capture_claude_session_id(&project_path, known_session_id.as_deref())
+        let current_known = last_known.lock().ok().and_then(|g| g.clone());
+        let captured = capture_claude_session_id(&project_path, current_known.as_deref())
             .map_err(
                 |e| tracing::debug!(target: "session.capture", "Claude disk scan failed: {}", e),
             )
             .ok()
-            .and_then(validated_session_id)
+            .and_then(validated_session_id);
+
+        if let Some(id) = captured.as_ref() {
+            if let Ok(mut guard) = last_known.lock() {
+                *guard = Some(id.clone());
+            }
+        }
+
+        captured
     }
 }
 
@@ -2177,6 +2192,56 @@ mod tests {
             capture_claude_session_id("/tmp/myproject", Some(uuid_known)).unwrap(),
             uuid_fresh
         );
+
+        match old_val {
+            Some(v) => std::env::set_var("CLAUDE_CONFIG_DIR", v),
+            None => std::env::remove_var("CLAUDE_CONFIG_DIR"),
+        }
+    }
+
+    #[test]
+    #[serial]
+    fn test_claude_poll_fn_promotes_last_known_across_polls() {
+        // issue #1522: once a stale anchor falls back to the disk's most-recent,
+        // the closure must adopt that sid as its new anchor so a later sibling
+        // write (think: another session in the same fleet) cannot steal it
+        // back on the next poll. Without promotion, the closure stays anchored
+        // to its startup sid forever, and shared-repo cross-assignment returns
+        // the moment a session forks (/clear, /new, --fork-session).
+        let tmp = tempfile::tempdir().unwrap();
+        let project_dir = tmp.path().join("projects").join("-tmp-myproject");
+        std::fs::create_dir_all(&project_dir).unwrap();
+
+        let uuid_startup = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
+        let uuid_post_fork = "11111111-2222-3333-4444-555555555555";
+        let uuid_sibling = "99999999-8888-7777-6666-555555555555";
+
+        // Startup anchor's file is stale; the post-fork file is fresh.
+        std::fs::write(project_dir.join(format!("{uuid_startup}.jsonl")), "s\n").unwrap();
+        let stale = std::time::SystemTime::now() - Duration::from_secs(600);
+        std::fs::File::options()
+            .write(true)
+            .open(project_dir.join(format!("{uuid_startup}.jsonl")))
+            .unwrap()
+            .set_times(std::fs::FileTimes::new().set_modified(stale))
+            .unwrap();
+        std::fs::write(project_dir.join(format!("{uuid_post_fork}.jsonl")), "f\n").unwrap();
+
+        let old_val = std::env::var("CLAUDE_CONFIG_DIR").ok();
+        std::env::set_var("CLAUDE_CONFIG_DIR", tmp.path());
+
+        let poll = claude_poll_fn("/tmp/myproject".to_string(), Some(uuid_startup.to_string()));
+
+        // First poll: anchor is stale, fall back to most-recent (post-fork).
+        assert_eq!(poll().as_deref(), Some(uuid_post_fork));
+
+        // A sibling session in the same dir writes its transcript with a newer
+        // mtime than the post-fork file.
+        std::fs::write(project_dir.join(format!("{uuid_sibling}.jsonl")), "x\n").unwrap();
+
+        // Second poll: closure must now anchor to the post-fork sid (still
+        // fresh) and refuse to drift onto the sibling's newer file.
+        assert_eq!(poll().as_deref(), Some(uuid_post_fork));
 
         match old_val {
             Some(v) => std::env::set_var("CLAUDE_CONFIG_DIR", v),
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Thanks for the writeup and the linked issue, this fix is solid. Reviewed both commits on the branch (your `9a444252` plus the anchor-promotion follow-up `fc1b2421` that I pushed under maintainer-edits, which addresses the CodeRabbit point about the closure never updating the anchor after the first poll). Local `cargo fmt`, `cargo clippy -- -D warnings`, and `cargo test --lib session::capture` are clean (105 passing on Linux).

Two things worth knowing before this rolls out:

1. The 5-minute lag on `/clear` and `/new` detection now applies to single-session-per-repo too, not just `--fork-session`. Bounded and recoverable, but worth a line in the release notes.
2. The sandboxed path (`capture_claude_session_id_in_container`) has the same root cause and is intentionally deferred. Filing a follow-up so sandboxed-fleet users get the fix would be valuable.

Approving. CI is now unblocked, this can land once it goes green.

PR #1571, opened 2026-05-28
fix(tui): make w jump to next waiting in Attention

### LINKED ISSUE #1524: TUI: `w` snoozes instead of jumping to next idle in Attention sort
## Symptom

In the home TUI, pressing `w` snoozes the session under the cursor instead of jumping to the next idle/waiting session.

## Expected

`w` jumps to the next idle/waiting session (the upstream behavior added in #796).

## Where it lives

`src/tui/home/input.rs` already documents the conflict:

- Lines 1250–1264: in non-strict mode with `SortOrder::Attention`, `w` is gated to `toggle_snooze_at_cursor()`. The block comment notes that "Snooze and favorite are triage primitives; they only have a visible effect (and a sort impact) in Attention mode."
- Lines 2250–2257: the jump-to-next-waiting arm explicitly acknowledges it's "always dead" in non-strict Attention because the snooze arm above wins. It only fires when sort is *not* Attention.

So today the actual mapping is:

| Sort         | `w` does                  |
|--------------|---------------------------|
| Attention    | snooze cursor             |
| anything else| jump to next waiting      |

Attention is the sort most people sit in for triage, which is exactly the mode where jump-to-next-waiting is most useful — so in practice `w` never feels like a navigation key.

## Proposal

Pick one and make it consistent across sort modes. Options:

1. Move snooze off `w` entirely (the comment notes `h`/`H` already alias snooze with the "Hide" mnemonic), and let `w` always mean jump-to-next-waiting. Cleanest, restores #796 in the mode that needs it most.
2. Keep snooze on `w` only when there's something to snooze and otherwise fall through to jump, but that's surprising and depends on cursor context.

(1) seems right given the existing `h`/`H` aliases for snooze.

## Repro

1. Open the home TUI in default (Attention) sort.
2. Put cursor on any non-idle row.
3. Press `w`.
4. Observed: cursor session gets snoozed. Expected: cursor jumps to next idle/waiting row.

grepsedawk:
## Description

In the home TUI, `w` snoozed the session under the cursor instead of
jumping to the next idle/waiting session. In non-strict Attention sort
the snooze arm shadowed the jump-to-next-waiting binding (#796), so in
the sort people triage in most, `w` never acted as a navigation key.

This moves snooze off `w`/`W` entirely (Option 1 in #1524). Snooze stays
on `h`/`H` (the existing "Hide" alias), and `w` falls through to the
existing jump binding, which now fires in every sort mode.

Fixes #1524.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] Skipped a full e2e: covered by a deterministic input-handler
  regression test in `src/tui/home/tests.rs`
  (`test_non_strict_w_jumps_to_next_waiting_in_attention_sort`), which
  exercises the keybinding without a tmux harness.

## AI Usage

- [x] AI was used

## Overview
This PR fixes a bug in the home TUI where the `w` key was snoozzing the current session instead of jumping to the next waiting/idle session when using Attention sort mode. The fix resolves a keybinding conflict by moving the snooze functionality entirely off `w`/`W` (reassigning it to the existing `h`/`H` "Hide" alias) and ensuring `w` consistently performs the jump-to-next-waiting action across all sort modes.

### Updated Keybinding Logic (`src/tui/home/input.rs`)
- Removed the snooze/favorite triage arms from the `w`/`W` key handlers that were shadowing the jump-to-next-waiting binding
- Updated hotkey dispatch comments and documentation to reflect the new keybinding arrangement
- Clarified in comments that strict mode still reserves lowercase `w` for the typing-guard behavior

### Added Test Coverage (`src/tui/home/tests.rs`)
- Introduced a helper function (`attention_env_running_then_waiting`) to construct test environments with explicitly seeded session statuses
- Added a new regression test (`test_non_strict_w_jumps_to_next_waiting_in_attention_sort`) that verifies pressing `w` in non-strict Attention sort mode jumps to the next waiting session instead of opening the snooze dialog

## Benefits
- **Consistent behavior**: The `w` key now reliably jumps to the next waiting/idle session regardless of sort mode
- **Cleaner keybinding**: Snooze/favorite operations are consolidated under `h`/`H`, which already existed as a "Hide" alias, reducing redundant keybindings
- **Better UX**: Users get the expected behavior documented in issue `#1524` and implemented in previous work (`#796`)
- **Regression prevention**: New test ensures this specific bug doesn't reoccur in future changes

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1571?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -1247,29 +1247,15 @@ impl HomeView {
                 self.view_mode = ViewMode::Agent;
             }
             KeyCode::Char('q') => return Some(Action::Quit),
-            // `w` / `W` (snooze), `h` / `H` (snooze alias), and `f` / `F`
-            // (favorite) are gated to Attention sort. Snooze and favorite
-            // are triage primitives; they only have a visible effect
-            // (and a sort impact) in Attention mode. Outside Attention,
-            // mutating these flags would silently change persisted state
-            // with no on-screen feedback, so we ignore the press
-            // entirely. Other sort modes fall through to the existing
-            // fallback bindings (`h` collapses; `w` is jump-to-next-
-            // waiting in non-strict mode).
-            KeyCode::Char('w')
-                if !self.strict_hotkeys && self.sort_order == SortOrder::Attention =>
-            {
-                if let Err(e) = self.toggle_snooze_at_cursor() {
-                    tracing::error!("toggle_snooze_at_cursor failed: {}", e);
-                }
-            }
-            KeyCode::Char('W')
-                if self.strict_hotkeys && self.sort_order == SortOrder::Attention =>
-            {
-                if let Err(e) = self.toggle_snooze_at_cursor() {
-                    tracing::error!("toggle_snooze_at_cursor failed: {}", e);
-                }
-            }
+            // `h` / `H` (snooze) and `f` / `F` (favorite) are gated to
+            // Attention sort. Snooze and favorite are triage primitives; they
+            // only have a visible effect (and a sort impact) in Attention mode.
+            // Outside Attention, mutating these flags would silently change
+            // persisted state with no on-screen feedback, so we ignore the
+            // press entirely (`h` then falls through to the group-collapse
+            // binding). Snooze used to also live on `w` / `W`, which shadowed
+            // the jump-to-next-waiting binding below in Attention sort; #1524
+            // moved it off so `w` is navigation again.
             KeyCode::Char('h')
                 if !self.strict_hotkeys && self.sort_order == SortOrder::Attention =>
             {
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Reviewed locally against #1524 and the existing keybinding wiring.

Implementation matches Option 1 from the issue: snooze comes off `w`/`W` entirely and stays on the `h`/`H` aliases, so `w` falls through to the unconditional jump-to-next-waiting binding across every sort mode. The comment block above the `h`/`H`/`f`/`F` triage arms preserves the #796 / #1524 history without over-explaining, and the comment over `jump_to_next_waiting()` was updated to reflect that the arm finally fires in Attention.

The regression test mirrors the existing `test_non_strict_h_snoozes_only_in_attention_sort` style: seeds Running + Waiting sessions, parks the cursor on Running, sends `w`, asserts the snooze dialog stayed closed and the cursor landed on Waiting.

I ran `cargo build`, `cargo fmt --check`, `cargo clippy --lib`, and the full `cargo test --lib` against the PR branch in a clean clone: 2232 tests pass, fmt and clippy clean.

One intentional behavior change worth flagging: in strict mode, bare `W` (uppercase) is now a no-op (previously it snoozed in Attention). `H` already snoozes in strict mode and the help overlay only advertises `H`, so this is the right call; just noting it for the changelog.

Thanks for the quick turnaround on this one.

PR #1572, opened 2026-05-28
fix(session): anchor sandboxed Claude poller to its own session id (followup to #1523)

itisaevalex:
## Description

Follow-up to #1523 (merged as bbceca5). That PR fixed the **host** path's session-id cascade by anchoring the Claude poller to its own session id. The maintainer's review explicitly invited a mirror for the **sandboxed/container** path:

> The sandboxed path (`capture_claude_session_id_in_container`) has the same root cause and is intentionally deferred. Filing a follow-up so sandboxed-fleet users get the fix would be valuable.

This PR is that follow-up.

`capture_claude_session_id_in_container` previously shelled `ls -t "$DIR"/*.jsonl | head -1` inside the container — picking the most-recently-modified jsonl with no per-pane attribution. The same cross-assignment bug as the host path: multiple AoE sessions sharing a container/cwd would all resolve to whichever session wrote last.

### Fix

Refactored to mirror the host path:

- Shell snippet now emits every fresh (≤5 min mtime) UUID jsonl basename in **newest-first** order rather than just `head -1`.
- New pure function `select_claude_session_in_container(stdout_bytes, known_session_id)` does the per-instance attribution: prefer the anchor if it's among the fresh candidates, otherwise fall back to most-recent. Fully unit-testable without docker — same pattern the codex/vibe/pi/gemini/hermes container selectors already use.
- `capture_claude_session_id_in_container` and `claude_poll_fn_sandboxed` now accept `known_session_id`, threaded from `Instance::maybe_start_poller`'s `initial_known`.
- `claude_poll_fn_sandboxed` mirrors the host's `Mutex<Option<String>>` anchor promotion (from the maintainer's follow-up commit `fc1b242`): each successful capture is promoted into the anchor so the poller stays attached to the session's actual current id over its lifetime, not just the startup value.

### Reproducing the bug

Locally verified the bug exists *and* the fix resolves it by toggling the anchor-preference block in `select_claude_session_in_container`:

- Without the `if let Some(k) = known_session_id { ... }` branch (i.e. equivalent to the previous `head -1` behaviour): `test_select_claude_session_in_container_prefers_anchor` **fails** — the selector returns the most-recent UUID instead of the anchor, exactly the cascade in shared-container fleets.
- With the branch restored: test passes.

The new tests are the regression coverage so this can't silently regress.

### Behaviour preservation

- Single-session-per-container is unchanged: with one fresh candidate, the anchor (if present) *is* that candidate, so the same id is returned as before.
- The 5-minute staleness window and the "no fresh jsonl → error" semantics are preserved.
- Same fork-lag caveat noted in #1523: if a session legitimately forks while its old jsonl is still fresh, the new id is detected when the old goes stale (≤5 min). Acceptable; rare in practice.

### What changed

- `src/session/capture.rs`:
  - New `select_claude_session_in_container` (pure, testable).
  - `capture_claude_session_id_in_container` gains `known_session_id: Option<&str>`; snippet emits all fresh candidates; selector handles attribution.
  - `claude_poll_fn_sandboxed` gains `known_session_id: Option<String>` and mirrors the host's `Mutex` anchor promotion.
  - 5 new unit tests for the selector + 1 existing in-container test updated for the new signature.
- `src/session/instance.rs`:
  - Sandboxed branch of `maybe_start_poller` passes `initial_known.clone()` to `claude_poll_fn_sandboxed`, mirroring the host branch.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (function doc comments)
- [ ] For UI changes: included screenshot or recording (N/A — internal session-tracking)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

(Coverage lives at the selector level via unit tests in `src/session/capture.rs`, matching the established pattern for other agents' `select_X_session_in_container` functions.)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Claude Code), collaborating with @itisaevalex who is driving the work, reviewing the diff, and submitting the PR.

Same collaboration model as #1523. Diagnosis, code, and tests drafted with AI assistance; @itisaevalex reviewed before submission and will respond personally on any review threads (not via copy/paste from the AI).

## Overview
This PR fixes a bug in the sandboxed/container session handling where multiple sessions sharing the same container would incorrectly resolve to whichever one was most recently created. The fix implements intelligent session identification to ensure each session is correctly anchored to its own ID.

## What Changed
- **Added session selection logic**: Introduced a new helper function that picks the correct session ID from multiple candidates, prioritizing a "known" session ID if it's still valid, otherwise falling back to the most recent one.
- **Enhanced container polling**: Modified the container session capture mechanism to emit all valid session candidates instead of just picking the newest, then letting the new selection logic choose the right one.
- **Session anchoring**: Updated the background polling mechanism to maintain and promote the chosen session ID across multiple poll cycles, ensuring consistency throughout the session's lifetime.
- **Updated function signatures**: Added an optional `known_session_id` parameter to session capture functions to pass the anchor information through the call chain.

## Benefits
- **Correctness**: Sessions are now reliably identified even when multiple instances share the same container environment
- **Consistency**: Session tracking is anchored and maintained across the lifetime of the poller, preventing drift to other sessions
- **Robustness**: Mirrors the successful approach from a prior fix (`#1523`), using proven patterns for the container-based workflow
- **Backward compatible**: Single-session-per-container scenarios work unchanged; only multi-session scenarios benefit from the fix

## Technical Details
- Added `select_claude_session_in_container()` function that parses UUID basenames from stdout (newest-first order) and intelligently selects either the provided `known_session_id` (if present among fresh ≤5 min old candidates) or the most recent candidate
- Modified `capture_claude_session_id_in_container()` to accept optional `known_session_id` parameter and changed the Docker shell snippet to emit all fresh UUID candidates instead of just the single most recent
- Modified `claude_poll_fn_sandboxed()` to accept `known_session_id` and maintain an anchored `last_known` value across polls using a Mutex, with promotion logic when captures succeed
- Added 5 new unit tests covering anchor preference, fallback behavior, error conditions, and non-UUID line filtering
- Changes span 151 lines added, 27 lines removed across two files: `src/session/capture.rs` and `src/session/instance.rs`

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1572?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -268,35 +268,79 @@ pub(crate) fn claude_poll_fn(
     }
 }
 
+/// Pick the active Claude session UUID from the container shell snippet's
+/// stdout in [`capture_claude_session_id_in_container`].
+///
+/// Stdout is expected to be UUID basenames, one per line, in newest-first
+/// order, already filtered to files modified within the last 5 minutes.
+///
+/// Mirrors the host-side anchor preference (issue #1522): when several AoE
+/// sessions share one container/cwd, the bare "most recent" heuristic makes
+/// every poller report whichever session wrote last and cross-assign ids.
+/// If our `known_session_id` is among the fresh candidates, prefer it;
+/// otherwise fall back to the most-recent (first in the list). With no
+/// candidates, return an error so the caller can surface it the same way
+/// the previous "no active session" path did.
+fn select_claude_session_in_container(
+    stdout_bytes: &[u8],
+    known_session_id: Option<&str>,
+) -> Result<String> {
+    let text = String::from_utf8_lossy(stdout_bytes);
+    let candidates: Vec<&str> = text
+        .lines()
+        .map(str::trim)
+        .filter(|l| !l.is_empty() && Uuid::parse_str(l).is_ok())
+        .collect();
+
+    if candidates.is_empty() {
+        anyhow::bail!("No active Claude session found in container");
+    }
+
+    if let Some(k) = known_session_id {
+        if candidates.contains(&k) {
+            return Ok(k.to_string());
+        }
+    }
+
+    Ok(candidates[0].to_string())
+}
+
 /// Capture Claude Code session ID inside a Docker container.
 ///
 /// Claude in a sandboxed AoE session writes its `.jsonl` files to the
 /// container's `~/.claude/projects/{encoded-cwd}/` directory, not the host's.
-/// This shells `docker exec` into the running container to find the most
-/// recently modified UUID-named jsonl in that directory, with a 5-minute
-/// staleness guard.
+/// This shells `docker exec` into the running container, lists every fresh
+/// (≤5 min mtime) UUID-named jsonl in that directory newest-first, and
+/// delegates per-pane attribution to [`select_claude_session_in_container`]
+/// — which prefers `known_session_id` when it's still fresh, falling back
+/// to the most-recent file otherwise. This mirrors the host path's anchor
+/// (issue #1522) so fleets sharing a single container/cwd don't
+/// cross-assign session ids.
 pub(crate) fn capture_claude_session_id_in_container(
     container_name: &str,
     container_cwd: &str,
+    known_session_id: Option<&str>,
 ) -> Result<String> {
     let dir_name = encode_claude_project_path(container_cwd);
 
-    // Shell snippet:
+    // Shell snippet (POSIX-portable):
     //   - resolve $CLAUDE_CONFIG_DIR or $HOME/.claude
     //   - walk projects/<encoded>/ for *.jsonl files
     //   - keep ones with mtime within 5 minutes
-    //   - emit basename (without .jsonl) of the most recent
+    //   - emit basenames (without .jsonl) in newest-first order
     //
-    // Using POSIX `find -mmin -5` and `ls -t` to avoid GNU-only `printf '%T@ %f'`.
+    // We emit every fresh candidate (not just `head -1`) so the Rust
+    // selector can apply per-instance anchor preference (issue #1522).
+    // Using POSIX `find -mmin -5` and `ls -t` to avoid GNU-only flags.
     let snippet = format!(
         r#"
 CLAUDE_HOME="${{CLAUDE_CONFIG_DIR:-$HOME/.claude}}"
 DIR="$CLAUDE_HOME/projects/{dir_name}"
 [ -d "$DIR" ] || exit 0
-NEWEST=$(ls -t "$DIR"/*.jsonl 2>/dev/null | head -1)
-[ -z "$NEWEST" ] && exit 0
-[ -n "$(find "$NEWEST" -mmin -5 2>/dev/null)" ] || exit 0
-basename "$NEWEST" .jsonl
+for f in $(ls -t "$DIR"/*.jsonl 2>/dev/null); do
+  [ -n "$(find "$f" -mmin -5 2>/dev/null)" ] || continue
+  basename "$f" .jsonl
+done
 "#
     );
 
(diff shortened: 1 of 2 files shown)


jerome-benoit:
I do not think it's actually the right approach, you are bypassing the builtin mechanism that map a tmux session to a session id instead of identifying the root cause of its failure in your use case and introduce a pile of quite ugly workarounds. 

I think both commits should be reverted in favor of in depth analysis of the actual root cause.

njbrake:

Thanks for the followup, @itisaevalex, and thanks @jerome-benoit for the careful read. The consistency concern is fair and worth engaging on properly rather than just landing this.

You're right that `claude_poll_fn` (`src/session/capture.rs:242`) and `claude_poll_fn_sandboxed` are the only pollers that don't thread through `compose_exclusion` / `build_exclusion_set` (`:674`, `:695`), the hidden-env table populated via `AOE_CAPTURED_SESSION_ID_KEY` that every other agent reads. The 33-line comment at `:209-241` justifying the asymmetry is itself a signal that the design carries cost, and I'd rather not pretend otherwise.

That said, I don't think a revert is the right move, for two reasons specific to Claude that don't apply to codex/vibe/pi/gemini/hermes:

1. **Claude's UUID is pre-minted by AoE.** `acquire_session_id` generates the id and passes it via `--session-id` before the poller starts. For the other agents the poller discovers an id the agent generated; for Claude the poller knows the expected id at startup. That "expected id" is what the anchor represents. The other agents have nothing analogous to anchor to, which is why exclusion alone fits them.
2. **Claude's session id mutates mid-life.** `/clear`, `/new`, and `--fork-session` each mint a new `<uuid>.jsonl`. Exclusion alone has a race here: after instance A runs `/clear` and writes X2.jsonl, instance B's poller can fire before A's `on_change` callback updates `AOE_CAPTURED_SESSION_ID_KEY`. B sees {X1, X2, Xb}, excludes only X1, picks most-recent of the rest = X2, and steals A's new id. The anchor blocks that because B's anchor stays Xb. The other agents don't hit this race because their session id is durable over the session's lifetime, so the env value B reads is always current enough.

#1523 landed the host-side anchor on the same reasoning. This PR mirrors it to the sandboxed path, which sandboxed-fleet users hit on `main` today with no alternative fix in place. Reverting both leaves real users cross-assigning while we debate.

On the bigger question of whether Claude could eventually be unified onto the exclusion mechanism: it's worth a real look. The most plausible path is making the env write synchronous inside the poller closure rather than via the async `on_change`, so siblings see the new id before they next poll. That closes the race above and would let Claude drop the anchor. I'd rather review that as a focused refactor PR covering both host and container paths than block this fix on it. Happy to open an issue to capture the design and credit the discussion.

Plan: land this as-is, open the unification issue, revisit if anyone wants to take it on.

njbrake:
Following up from what my agent said: generally Jerome I trust to your expertise here so if you do not think we should pursue this approach then I agree this should be further investigated before merge

jerome-benoit:
> Following up from what my agent said: generally Jerome I trust to your expertise here so if you do not think we should pursue this approach then I agree this should be further investigated before merge

I'm not against the merge since it permits to fix a use case, and not a edge one. But I think the root cause should be tracked and identified: the current design to keep a 1:1 map between session id and tmux session seems to have missed something in the session id continuous polling to ensure a proper isolation. It happens on claude code but it might affect others. 

I would prefer to have a global guard once the root cause identified using the tmux session env vars to ensure that any subsequent pooling implementation cannot reproduce the issue and can be debugged.

itisaevalex:
I will see if there is a way to make this more elegant. To give a surface overview originally I stumbled onto this issue when my machine ran out of space, that crashed tmux and thus aoe (the individual sessions went red), in the process of recovering the sessions I found the reassignment bug and had to recover/patch aoe manually. This tends to happen a lot on VMs with small drives and no space guards.

itisaevalex:
Okay, did a bit of digging, and I agree the anchor is treating the symptom.
What I surfaced is that the exclusion guard is only correct while each session's AOE_CAPTURED_SESSION_ID env stays up to date, which only happens when the owning session's poller discovers a new id. That's fine when agent ids stay stable for their lifetime, but Claude mints a fresh <uuid>.jsonl on /clear, /new, and --fork-session, so there's a window (up to a full poll interval) where a sibling sees a stale exclusion set and grabs the new id. Hence the cascade of mangled Claude sessions, and why Claude is the one that trips it.

A fix that keeps the single-guard design: have Claude's SessionStart hook publish the id to the tmux env. It fires on exactly those mutation events and gets session_id in its payload, so the guard is authoritative the instant the id changes, and Claude can then drop the anchor and use compose_exclusion like the others.

I tested the hook side on a live pane: SessionStart fires on /clear and publishes the new session_id to the tmux env immediately, and a sibling reading that env picks it up. I haven't wired it into a built aoe end-to-end yet though, so I'd lean toward landing the unification with the anchor kept as a fallback first, then dropping it as a follow-up once it's run on a real multi-session build, rather than pulling the safety net on something that isn't integration-tested yet.

jerome-benoit:
> A fix that keeps the single-guard design: have Claude's SessionStart hook publish the id to the tmux env. It fires on exactly those mutation events and gets session_id in its payload, so the guard is authoritative the instant the id changes, and Claude can then drop the anchor and use compose_exclusion like the others.

Is it reliable? claude hooks are known to miss relevant events. 

- aoe generates a UUID for new session: that UUID is published in the hidden env var?
- user start new session (or resume another one) inside existent tmux session: if the hook is triggered reliably and the UUID is published in the hidden env var, the exclusion mechanism will work. 

Is it possible to: 

+ rely on the existing exclusion mechanism first (hooks based)  
+ fallback to the anchor mechanism to cover any claude hooks issue (+ logging) 

? 

Thanks.

njbrake:

Thanks for this, @itisaevalex, and for the digging in the thread. I'm going to close this in favor of #1735, but I want to be clear that your work shaped that PR directly.

Your follow-up did two useful things: it mirrored the #1523 anchor to the sandboxed path (the right immediate instinct), and your analysis in the thread pinned down the real root cause. You identified that the exclusion guard is only correct while each session's `AOE_CAPTURED_SESSION_ID` stays current, and that Claude breaks it by minting a fresh jsonl on `/clear`, `/new`, and `--fork-session`, leaving a poll-interval window where a sibling grabs the new id. Your proposed fix, having Claude's `SessionStart` hook publish the id to the tmux env so the guard is authoritative the instant the id changes with the anchor kept as a fallback, is the design #1735 implements.

#1735 carries that forward with the synchronous publish at `finalize_launch`, per-tick exclusion plus an anchor-as-fallback tie-break, and the hook-driven sidecar, across both the host and sandboxed paths. Since it supersedes the anchor-only approach and covers your sandboxed use case, I'd rather land one solution than two overlapping ones.

Closing as superseded by #1735. Thank you for the careful diagnosis and for engaging on the root cause; it genuinely moved the fix forward.

Issue #1168, opened 2026-05-16
TUI cockpit view shows empty transcript when loading a session started in the web cockpit (likely auth mismatch + silent error swallow hides it)


njbrake:
> Filed by Nathan, with code investigation done by Claude. Words below are AI-drafted unless quoted; I (Nathan) confirmed the symptom by hand.

## Symptom

Running `aoe serve` over Tailscale, started a cockpit session via the **web cockpit UI**, with the agent doing real work (events generated). Then opened the **TUI** (locally, on the same machine as the daemon). TUI session list correctly shows the session and that it's in cockpit mode. Open the session → TUI cockpit pane renders empty. No transcript, no error message, no toast — it looks like a brand-new empty session even though the web view has the same session full of events.

## Code reads — wiring is in place but errors are silently swallowed

The TUI cockpit view DOES try to backfill via `/cockpit/replay` before the WS opens (`src/tui/cockpit_view/mod.rs:152-171`):

```rust
let http = HttpClient::new(endpoint.clone()).context("build cockpit HTTP client")?;
let initial = http.replay(session_id, 0).await;
let ws = ws_connect(&endpoint, session_id, 0).await.ok();

let mut state = CockpitViewState::new(session_id.to_string(), endpoint, http, ws);
state.focus = Focus::Transcript;

if let Ok(replay) = initial {
    // ... apply frames
}
```

Two error-swallowing sites that make this *unobservable* to the user when something goes wrong:

- **Line 163:** `if let Ok(replay) = initial { ... }` — if `http.replay()` errored (HTTP 401, 404, network), the `Err` branch is silently dropped. No toast, no banner, no `tracing::warn!`. The transcript stays empty and the user has no clue why.
- **Line 158:** `ws_connect(...).await.ok()` — if the WebSocket dial failed, the `Err` is converted to `None` with `.ok()` and the live-event stream is just absent. Same lack of diagnostics.

So the symptom is "empty pane" no matter what the actual failure was.

## Most likely cause for *this* repro (Tailscale + web-started session)

The TUI cockpit client uses `DaemonEndpoint` discovered via `src/cockpit/client/discovery.rs:65-116`:

- `discover_env` checks `AOE_DAEMON_URL`/`AOE_DAEMON_TOKEN` first.
- Otherwise `discover_local` reads `serve.url`, prefers the loopback line (`is_loopback` check at `:118-127`), and extracts the `?token=` query param as Bearer auth (`:106, 133-144`).

[...]

Issue #1160, opened 2026-05-15
serve: --auth=<mode> selector + --behind-proxy flag for reverse-proxy deployments, decouple passphrase-only auth from --remote tunnel


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

### Target deployment

`aoe serve` runs locally on my Mac, bound to `127.0.0.1`. An SSH reverse tunnel forwards a port on my server to the Mac's loopback. My server runs Traefik with a real TLS cert for `aoe.xxxxxxx.yyy` and proxies the public URL to the SSH-tunnel'd port. End-user hits `https://aoe.xxxxxxx.yyy` from a phone.

```
phone → https://aoe.xxxxxxx.yyy (Traefik TLS terminate) → SSH reverse tunnel → 127.0.0.1:PORT (aoe serve on Mac)
```

In this setup AoE is just an HTTP backend. TLS is handled by Traefik. The network path between Traefik and AoE is a private SSH tunnel, so the only thing reaching AoE from the public internet is already authenticated-at-the-tunnel-level traffic. I want one passphrase as the human gate, and that's it. No token URL, no auto-tunnel, no Tailscale Funnel.

### What's missing

`aoe serve` has three auth modes today: full token-URL auth (default), `--no-auth` (loopback-only bypass), and `--remote` (token + mandatory passphrase, plus auto-tunnelling via Tailscale Funnel or Cloudflare Quick Tunnel). None of these fit the above deployment:

- **Default**: token URL is high friction on a phone. Must paste a long URL on first visit, and the token rotates on daemon restart, invalidating cookies and forcing a re-paste.
- **`--no-auth`**: strips everything including the passphrase. Loopback-only, which the tunnel satisfies, but the public URL through Traefik would be wide open if I ever needed `--host 0.0.0.0`.
- **`--remote`**: force-starts a Tailscale Funnel or Cloudflare Quick Tunnel. Conflicts with my Traefik ingress; I do not want AoE to manage its own ingress.

What I want: passphrase as the only gate, no token URL, no auto-tunnel. Cookies should be `; Secure` and trusted `X-Forwarded-For` from Traefik should parse correctly so per-IP rate limiting and login bindings work against the real client IP, not the loopback peer.

## Auth flag matrix (current behavior, aoe main as of 2026-05-16)

| Flag combination | Token check | Passphrase check | Notes |
|---|---|---|---|
[...]

PR #1596, opened 2026-05-28
fix(server): bypass passphrase wall for loopback callers

### LINKED ISSUE #1525: Cockpit TUI: loopback caller blocked by passphrase wall, --auth=passphrase skips the is_local_trusted bypass from #1190

## Problem

Local TUI cannot open a cockpit session against a same-host `aoe serve --auth=passphrase` daemon. Pressing Enter on a cockpit session row shows:

```
startup failed: replay=authentication failed; check AOE_DAEMON_TOKEN or restart `aoe serve`; ws=websocket transport error HTTP error: 401 Unauthorized
```

Reproduces every time. Daemon was started with `aoe serve --host 127.0.0.1 --port 42041 --auth passphrase --behind-proxy`; TUI runs on the same machine, hits loopback. The `AOE_DAEMON_TOKEN` advice in the toast is misleading; there is no token gate in this mode at all.

This is a regression of the fix for #1168. PR #1190 added the loopback bypass for the token-auth path only, then #1160 introduced `--auth=passphrase` which routes loopback callers through a separate code path that has no bypass.

## Evidence from the affected daemon

`~/.agent-of-empires/serve.url` (24 bytes, no `?token=`):

```
http://127.0.0.1:42041/
```

`~/.agent-of-empires/debug.log` for the failing click:

```
WARN auth: passphrase wall: rejecting api/ws with 401 ip=127.0.0.1 path=/api/sessions/1d339ea5aa6442f8/cockpit/replay had_session_cookie=false had_device_binding=false
WARN http.request: completed status=401 latency_ms=0
DEBUG cockpit.client.ws: connecting to cockpit ws url=ws://127.0.0.1:42041/sessions/1d339ea5aa6442f8/cockpit/ws
DEBUG auth: auth_middleware entered for cockpit ws ip=127.0.0.1 token_sources=[] ws_protocol_count=0
WARN auth: passphrase wall: rejecting api/ws with 401 ip=127.0.0.1 path=/sessions/1d339ea5aa6442f8/cockpit/ws had_session_cookie=false had_device_binding=false
WARN cockpit.tui: initial replay failed: authentication failed; check AOE_DAEMON_TOKEN or restart `aoe serve`
WARN cockpit.tui.ws: initial ws connect failed: websocket transport error: HTTP error: 401 Unauthorized
```

## Root cause

Three layers, all in `src/server/`:

**Layer 1: `--auth=passphrase` writes no token into `serve.url`.** `src/cli/serve.rs:573` collapses `AuthMode::Passphrase` into `no_auth: true` for the server config. `src/server/mod.rs:467-475` then sets `auth_token = None` and `src/server/mod.rs:733-738`'s `make_url` builds `http://{host}:{port}/` without the `?token=` suffix. TUI discovery (`src/cockpit/client/discovery.rs:106` calling `extract_token`) returns `None`, so the HTTP/WS clients send no Authorization header.

**Layer 2: `auth_middleware` routes the request into `run_passphrase_wall`.** `src/server/auth.rs:537` short-circuits on `is_no_auth() == true`, which is also true for `--auth=passphrase`. The token-auth branch that contains the loopback bypass at `src/server/auth.rs:728-735` is never reached.

**Layer 3: `run_passphrase_wall` has no loopback carve-out.** `src/server/auth.rs:425-498`. The wall checks for a session cookie + device binding, and for `/api/*` or `/ws` paths it returns 401 `login_required` with no consultation of `is_local_trusted(client_ip)`. The carve-out logic that mirrors #1168's intent (loopback caller already passes the fs-permission trust boundary) was never added here.

The token-auth branch's bypass at `auth.rs:726-735` calls `post_token_auth_action`, which honors `is_local_trusted`. `run_passphrase_wall` should make the equivalent call before rejecting an `/api/*` / `/ws` request.

## Proposal

**(A) Minimal fix: add the loopback bypass to `run_passphrase_wall`.** In `src/server/auth.rs:446-466`, before returning 401 / redirect, check `is_local_trusted(client_ip)` and pass the request through to `next.run(request).await` when the caller is on loopback. This mirrors the token-auth branch at `auth.rs:728-735`. Skip the elevation check too (or apply it only when `is_local_trusted == false`); persistent-config writes from a loopback caller already pass the fs-perm boundary so layering a 15-minute confirmation window on top adds no security. Inserting a placeholder session id is not needed because the wall's only consumers downstream are the elevation check and the sliding-window cookie refresh, both of which become no-ops in the bypass path.

```rust
if !has_valid_session {
    if is_local_trusted(client_ip) {
        tracing::info!(
            target: "auth.passphrase",
            ip = %client_ip,
            path = %path,
            "loopback bypass: skipping passphrase factor in passphrase-only mode"
        );
        return next.run(request).await;
    }
    if path.starts_with("/api/") || path.contains("/ws") {
        // existing 401 login_required path
    }
    // existing /login redirect path
}
```

**(B) Toast text: stop pointing at `AOE_DAEMON_TOKEN` when no token is involved.** `src/cockpit/client/http.rs:38` is the source of the misleading message. Either drop the `AOE_DAEMON_TOKEN` hint and just say "daemon rejected the request, restart `aoe serve` or check `--auth` mode," or thread the auth mode through into a more specific message. Today the toast steers a passphrase-mode user to a non-existent env var.

**(C) Optional: surface the auth mode in `aoe serve --status`.** Useful follow-on but out of scope; `serve.mode` already tracks transport (`local` / `tunnel` / `tailscale`); a parallel `serve.auth` file or extra line in `serve.url` would let the TUI describe what's actually rejecting the connection.

## Why now

- Cockpit-in-TUI was the headline feature for #1018 / #968. `--auth=passphrase` is the recommended mode for `aoe serve` on a shared LAN behind a reverse proxy (per #1160). Together they brick the most common multi-surface workflow ("start a cockpit session on my phone, attach in the TUI when I sit down at my desk") for any user who picked the passphrase auth mode.
- The fix is a 5-line change in one file; the regression test surface is the same axum middleware we already cover.
- The toast text is actively misleading; users currently chase `AOE_DAEMON_TOKEN` on a daemon that never had a token.

- As a local TUI user, when I open a cockpit session against a same-host `aoe serve --auth=passphrase` daemon, then `GET /api/sessions/{id}/cockpit/replay` and the `/sessions/{id}/cockpit/ws` upgrade return 200 without a passphrase session cookie or device binding.
  - Suggested test layer: Rust integration test exercising `auth_middleware` + `run_passphrase_wall` end-to-end with `ConnectInfo` set to `127.0.0.1`, matching the existing `is_local_trusted_recognizes_loopback` style in `src/server/auth.rs:924`.
- As a remote caller hitting the same daemon over a non-loopback IP with no session cookie + no device binding, then I still get 401 `login_required` on `/api/sessions/{id}/cockpit/replay` and the WS upgrade.
  - Suggested test layer: Rust integration test, negative case in the same module.
- As a TUI user when replay or ws fails with 401 against the cockpit endpoints, then the startup toast does not mention `AOE_DAEMON_TOKEN` when the daemon is running with `--auth=passphrase` or `--auth=none`.
  - Suggested test layer: Rust unit test on the `HttpError::Unauthorized` `Display` impl, or a `tests/e2e/` recipe that spins up `aoe serve --auth=passphrase` and asserts the toast string.

## Open questions

- Does the bypass also need to cover `requires_elevation` paths (settings writes, profile mutations) for loopback callers, or should those stay behind the passphrase factor even on loopback? Today the token-auth bypass treats loopback as fully trusted; consistency would say "yes, also bypass elevation." Worth confirming the policy intent here.
- Should the `serve.url` shape stay tokenless under `--auth=passphrase`, or should it embed a one-shot bootstrap token the TUI can present? Tokenless is cleaner once (A) lands; the bootstrap-token route would also fix non-loopback TUI-to-daemon attach for passphrase mode but is a much larger change.

## Out of scope

- Non-loopback TUI attach against `--auth=passphrase` daemons. The TUI still has no way to complete a passphrase exchange; that's a separate feature.
- `--auth=none` behavior is unchanged by this fix; the no-auth branch in `auth_middleware` already passes through without entering the wall.

## Cross-references

- #1168, #1190 — original "TUI cockpit empty pane" report and the loopback-bypass PR that landed the carve-out for the token-auth path but missed the passphrase path.
- #1160 — `--auth=<mode>` selector that introduced the passphrase-only mode, source of this regression.
- #1131 — device-binding header / WS sub-protocol shape; relevant to story 1 because the bypass needs to short-circuit before either check fires.

Seluj78:

## Description

A local TUI cockpit attach against a same-host `aoe serve --auth=passphrase` daemon was returning 401 on every replay + ws upgrade. PR #1190 added the `is_local_trusted` loopback bypass to the token-auth branch for #1168, then #1160 introduced `--auth=passphrase`, which routes through `run_passphrase_wall` (a separate code path with no bypass). Same-host TUI users on the recommended LAN-behind-proxy auth mode were bricked on the cockpit-in-TUI flow.

This PR mirrors the #1168 carve-out inside the passphrase wall. A loopback caller already passes the filesystem-permission trust boundary on `~/.agent-of-empires/serve.*` (mode 0600), so layering the passphrase factor on top adds friction without strengthening the boundary. Remote callers proxied through a tunnel come in with the real remote IP via `resolve_client_ip` and continue to hit the wall as expected. The decision is extracted into a pure `passphrase_wall_entry_action` helper so the policy matrix is table-testable, the same way `post_token_auth_action` was extracted for #1168.

Two follow-ons in the same PR:

1. `HttpError::Unauthorized`'s Display string used to tell users to "check `AOE_DAEMON_TOKEN`" on any 401 from the cockpit endpoints. That hint is actively wrong on `--auth=passphrase` and `--auth=none` daemons that never had a token, so the toast steered users at a non-existent env var while the real fault went uninvestigated. Replaced with a mode-agnostic message.
2. `docs/cockpit.md` notes the new behavior in the TUI vs web dashboard section, including the threat-model reasoning so it doesn't read as an auth-weakening loophole.

Closes #1525

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

New `cli_serve_auth_passphrase_loopback_bypass` covers the bypass on `/api/about` and `/api/sessions` from 127.0.0.1 with no session cookie + device binding. The existing `cli_serve_auth_passphrase_login_round_trip` test asserted 401 on a no-cookie loopback GET; that returns 200 now, so step 1 was retargeted at a simulated remote caller via `X-Forwarded-For: 10.0.0.5` (the daemon binds loopback, so `resolve_client_ip` trusts the XFF header). Unit-level coverage comes from `passphrase_wall_entry_action_matrix` in `src/server/auth.rs::tests` (loopback v4/v6 → `BypassLoopback`, remote → `Continue`, login-bootstrap path → `BypassExempt` regardless of IP) and `unauthorized_display_omits_token_env_var` in `src/cockpit/client/http.rs::tests` (pins the new error string so the env-var hint can't regress back in).

## How to test

- [ ] Start a daemon: `aoe serve --daemon --host 127.0.0.1 --port 42041 --auth passphrase --passphrase test-pass`.
- [ ] From a separate terminal, open the TUI: `aoe`. Navigate to a session that's running in cockpit mode and press Enter. The cockpit transcript should render without the `startup failed: replay=authentication failed; ws=...` toast. Replay and ws traffic should both 200 / 101 in `~/.agent-of-empires/debug.log`.
- [ ] Confirm the remote path still rejects: `curl -i -H "X-Forwarded-For: 10.0.0.5" http://127.0.0.1:42041/api/about` returns `401 login_required`.
- [ ] Confirm the loopback path now passes: `curl -i http://127.0.0.1:42041/api/about` returns `200` with `"auth_mode":"passphrase"`.
- [ ] On any 401 that does occur (e.g. an actual stale session against a token-auth daemon), the TUI startup toast no longer mentions `AOE_DAEMON_TOKEN`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (notably: full bypass for loopback callers, skipping the elevation step-up consistent with the token-auth path; no token-embedded bootstrap URL change), and ran the manual test steps.

## Release Notes

* **Documentation**
  * Clarified behavior of local TUI connections with passphrase authentication.

* **Bug Fixes**
  * Enhanced error messaging to provide actionable guidance during authentication issues.
  * Fixed passphrase authentication to correctly bypass requirements for same-host local connections while maintaining security for remote access.

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1596?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/cockpit.md
+++ b/docs/cockpit.md
@@ -239,6 +239,13 @@ state are always in sync.
 - **TUI status indicators**: a cockpit session that's healthy shows
   as Idle/Active in the TUI session list, since cockpit health is
   observed via the ACP event stream rather than tmux pane probing.
+- **`--auth=passphrase` daemons**: the local TUI attaches to a
+  same-host daemon without going through the passphrase exchange.
+  Loopback callers are treated as fs-trusted because the daemon's
+  serve files under `~/.agent-of-empires/serve.*` are already 0600,
+  so the filesystem permission boundary protects same-host access.
+  Remote callers proxied through a tunnel still hit the passphrase
+  wall as expected. See #1525.
 
 ### TUI cockpit view keybinds
 
--- a/src/cockpit/client/http.rs
+++ b/src/cockpit/client/http.rs
@@ -35,7 +35,12 @@ pub enum HttpError {
     SessionNotFound(String),
     #[error("daemon is read-only (started with --read-only); request refused")]
     ReadOnly,
-    #[error("authentication failed; check AOE_DAEMON_TOKEN or restart `aoe serve`")]
+    // The daemon may reject for several reasons: stale token, missing
+    // passphrase session, device binding mismatch. Pointing at
+    // `AOE_DAEMON_TOKEN` was misleading on `--auth=passphrase` and
+    // `--auth=none` daemons that never had a token in the first
+    // place. See #1525.
+    #[error("daemon rejected the request (401); restart `aoe serve` or check `--auth` mode")]
     Unauthorized,
     #[error("daemon returned HTTP {status}: {body}")]
     Server { status: StatusCode, body: String },
@@ -210,5 +215,24 @@ mod tests {
     fn auth_skips_bearer_when_no_token() {
         let client = HttpClient::new(endpoint("http://127.0.0.1:8080", None)).unwrap();
         assert!(client.endpoint().token.is_none());
+    }
+
+    // Regression test for #1525. The startup toast on a 401 from the
+    // cockpit endpoints folds in `HttpError::Unauthorized`'s Display.
+    // Previously that Display string hard-coded `AOE_DAEMON_TOKEN`,
+    // which made the toast actively misleading on `--auth=passphrase`
+    // and `--auth=none` daemons that never had a token. Pin the new
+    // wording so the env-var hint can't regress back in.
+    #[test]
+    fn unauthorized_display_omits_token_env_var() {
+        let rendered = HttpError::Unauthorized.to_string();
+        assert!(
+            !rendered.contains("AOE_DAEMON_TOKEN"),
+            "Unauthorized message must not pin diagnosis to a token env var that does not exist in passphrase or no-auth mode: {rendered}"
+        );
+        assert!(
+            rendered.contains("401"),
+            "Unauthorized message should still surface the underlying HTTP status: {rendered}"
+        );
     }
 }
--- a/src/server/auth.rs
+++ b/src/server/auth.rs
@@ -310,6 +310,46 @@ fn post_token_auth_action(
     }
 }
 
+/// Decision for the entry of `run_passphrase_wall`: a request landed on
+/// the wall because the daemon is running in `--auth=passphrase` mode
+/// (token gate disabled, passphrase login active). The wall normally
+/// requires a session cookie + device binding for `/api/*` and `/ws`,
+/// redirects to `/login` otherwise, and step-up-elevates writes to
+/// persistent-config surfaces. Two conditions bypass that entire
+/// gauntlet: a login-bootstrap path (so the SPA can load assets and
+/// post to `/api/login`) and a loopback caller (so the local TUI can
+/// attach without a passphrase exchange).
+#[derive(Debug, PartialEq, Eq)]
+enum PassphraseWallEntryAction {
+    /// Path is in the login-bootstrap allow-list. Skip the wall so
+    /// `/login`, `/api/login`, static assets, etc. stay reachable
+    /// even without a session.
+    BypassExempt,
+    /// Caller is on loopback. fs-perm boundary on
+    /// `~/.agent-of-empires/serve.*` already protects same-host
+    /// access, so layering the passphrase factor on top adds friction
+    /// without strengthening the trust boundary. Mirrors the token-
+    /// auth path's `is_local_trusted` carve-out from #1168 so the
+    /// local TUI works against an `--auth=passphrase` daemon. See
+    /// #1525.
+    BypassLoopback,
+    /// Run the full session + device-binding + elevation flow.
+    Continue,
+}
+
+/// Resolve the entry decision for `run_passphrase_wall`. Extracted so
+/// the bypass policy is table-testable without standing up the full
+/// axum middleware. See #1525.
+fn passphrase_wall_entry_action(path: &str, client_ip: IpAddr) -> PassphraseWallEntryAction {
+    if is_login_session_exempt(path) {
+        return PassphraseWallEntryAction::BypassExempt;
+    }
+    if is_local_trusted(client_ip) {
+        return PassphraseWallEntryAction::BypassLoopback;
+    }
+    PassphraseWallEntryAction::Continue
+}
+
 /// Whether a request path + method needs an elevated login session
 /// (step-up auth, 15-minute passphrase confirmation window).
 ///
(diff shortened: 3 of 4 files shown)


Seluj78:
@coderabbitai review

Issue #1510, opened 2026-05-26
Web dashboard: theme picker reverts after reload, elevation gate over-broad on PATCH /api/profiles/{name}/settings re-prompts passphrase for safe preference fields


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Picking a theme in the web dashboard repaints immediately, but the new theme silently reverts after a page reload or `aoe serve` restart. Symptom is flaky: it appears to "work" until the next reload.

Two failure modes share the same underlying cause:

1. **No passphrase set.** Theme picker repaints, save lands; behavior looks fine until the user happens to land on a path where elevation is required and the PATCH 403s anyway (cookie expired, cross-tab session, dev/release namespace drift).
2. **Passphrase set.** Every theme change triggers a passphrase re-prompt, because `PATCH /api/profiles/{name}/settings` is gated by `requires_elevation`. If the user dismisses, the picker has already painted the new theme (the picker event fires before the PATCH resolves), so the dashboard shows the picked theme but disk still holds the old one. Reload reverts.

The picker event firing before the PATCH resolves means the visual change is decoupled from the persistence outcome. The user gets no feedback that the save failed.

### Elevation gate is too broad

`src/server/auth.rs:354-359` treats every non-GET under `/api/profiles/` as elevation-required:

```rust
if path.starts_with("/api/profiles/") {
    // Per-profile writes: PATCH /api/profiles/{name}/settings,
    // PATCH .../rename, DELETE /api/profiles/{name}. Read GETs
    // were filtered out by the GET/HEAD bypass above.
    return true;
}
```

The doc comment on `requires_elevation` (`src/server/auth.rs:319-326`) scopes the elevation gate to the persisted-tamper attack class:

> The protected attack class is the persisted-tamper pattern: an attacker with stolen session and binding plants a malicious Docker image, worktree template, or profile, then waits for the owner to spawn a session that runs it.

Theme, sound, updates, web (notification toggles), logging filter, and description carry none of that surface. They are user-preference fields with no shell-command, binary-path, or mount-template payload. The current rule elevates them anyway because the path prefix match is coarser than the threat model the rule encodes.

`ALLOWED_PROFILE_SETTINGS_SECTIONS` (`src/server/api/mod.rs:88-99`) already separates the safe sections from the dangerous ones:

```rust
[...]

PR #1597, opened 2026-05-28
test(web): theme switching and sidebar reorder user-story specs (#1405, #1419)

### LINKED ISSUE #1405: Web: add user-story Playwright specs for theme switching (no-auth, passphrase, token, persistence, races, failure modes)

## Problem

Theme switching from the web dashboard touches a long path: `ThemeSettings` dropdown → `onSaveField` → `PATCH /api/profiles/:name/settings` → `dispatchThemePickerChanged` → `useResolvedTheme` refetch → `applyResolvedTheme` → `document.documentElement.dataset.theme` + per-var `style.setProperty` + `localStorage["aoe-resolved-theme"]` + `aoe:theme-changed` event. Several things along that path can quietly fail, and most of them have zero UI-driven Playwright coverage today.

The original framing of this issue (theme switch returns 403 under `--auth=passphrase`) is one branch of that surface. There are more, with and without passphrase. Per my comment above: land a story-level coverage net first, then fix the bugs the specs surface. Without the net, every fix is one-shot and the next regression in the same path ships green.

This is the tracker issue. It mirrors #1383 / #1419 in shape: each story below is its own Playwright spec under `web/tests/live/cockpit-stories/settings-theme-*.spec.ts` (or `web/tests/*.spec.ts` for the mocked failure-mode variants).

### Existing coverage

- **Vitest contract**: `web/src/components/settings/__tests__/ThemeSettings.test.tsx` mocks `fetchThemes` and verifies the dropdown renders the option list and `onSaveField("theme", "name", ...)` fires on change. No `applyResolvedTheme` round-trip; no `document.documentElement` assertion; no real PATCH.
- **Live cockpit-story spec** (landing in #1397, see "Reverted in #1397" below): `web/tests/live/cockpit-stories/settings-theme-select.spec.ts` drives the no-auth path, picks a theme, asserts the select value sticks. Does not assert `document.documentElement.dataset.theme`, the `localStorage["aoe-resolved-theme"]` cache, or the `aoe:theme-changed` event.
- **Live ThemeSettings tracer**: none. There is no spec hitting `PATCH /api/profiles/:name/settings` with a theme payload against a real backend across the three auth modes.

### Reverted out of #1397

A passphrase variant (`settings-theme-select-passphrase.spec.ts`) was dropped from #1397 because it depends on the server fix landing. Re-introducing it is one of the stories below; #1397 should not be blocked on that single spec.

### Production code surface that no story exercises

- **`web/src/components/settings/ThemeSettings.tsx:20-29`** — `save` fires `onSaveField("theme", field, value)` (which PATCHes the profile settings) AND immediately dispatches `THEME_PICKER_CHANGED_EVENT` on the `name` field, without waiting for the PATCH to succeed. If the PATCH 403s, the local picker fires anyway, the picker fetches `/api/themes/:name`, `applyResolvedTheme` lands, and the user sees the new theme — but the server still holds the old value, and a reload reverts.
- **`web/src/components/SettingsView.tsx`** `sendSave` → on a 4xx, `loadSettings()` overwrites local state with the stale server value (the silent-revert path called out in the original issue body).
- **`web/src/hooks/useResolvedTheme.ts:30-63`** sequence-numbered apply: a slow mount fetch must not overwrite a faster user pick. Untested.
- **`web/src/lib/theme.ts:47-59`** `applyResolvedTheme` writes both `theme.web.cssVars` and `theme.terminal.cssVars`, sets `dataset.theme`, `dataset.themeAppearance`, `style.colorScheme`, and the `aoe-resolved-theme` localStorage entry. None of the four side effects is asserted by any spec.
- **`web/src/lib/theme.ts:64-70`** `THEME_CHANGED_EVENT` (`aoe:theme-changed`) — consumed by `useTerminal.ts:1260-1262` so xterm repaints on theme change. Untested at the integration layer.
- **`src/server/auth.rs:347-352`** `requires_elevation` treats every non-GET under `/api/profiles/` as needing the elevated session under passphrase auth. `PATCH /api/profiles/{name}/settings` is therefore 403 if the cookie was issued by the login wall but the elevation step did not run. Whether this is the bug or the elevation step is silently failing is what the spec will pin down.
- **`src/server/api/system.rs:244-283`** `GET /api/themes/:name` and `GET /api/theme/current` — read-only, exempt from elevation; they answer 200 under passphrase mode. The picker-fetch path therefore visually paints the theme even when the PATCH 403s, which is exactly the silent-revert UX.

## Root cause (current understanding)

Theme switching evolved as several independent landings: the dropdown (`ThemeSettings.tsx`), the resolved-theme applicator (`theme.ts`), the picker event (`useResolvedTheme.ts`), the elevation gate (`auth.rs`). Each is well-tested in isolation. The composition is not.

Two structural issues I expect the stories to surface:

1. **Visual change is decoupled from persistence.** `dispatchThemePickerChanged` runs before the PATCH resolves. A 403 leaves the picker fetch and `applyResolvedTheme` happy and the user with a temporarily-themed but unpersisted dashboard.
2. **`requires_elevation` covers profile writes under passphrase** but the client never re-prompts for elevation; it just eats the 403, calls `loadSettings()`, and the dropdown snaps back. There may also be paths (cockpit-only sessions, token-auth elevation freshness) where the elevation cookie is stale.

The specs will pin which is which; the fix follows.

## Proposal

Land a `web/tests/live/cockpit-stories/settings-theme-*.spec.ts` family. One spec per user story. The fix PR can drop into the gaps the specs document, with green CI catching the next regression.

#### Picker happy path

- **Pick a theme under no-auth.** Open Settings → Theme, choose a non-default option, expect: (a) the select value sticks, (b) `document.documentElement.dataset.theme` matches the new name, (c) `document.documentElement.dataset.themeAppearance` is `dark` or `light`, (d) at least one `--term-*` CSS variable on `:root` matches the resolved theme payload, (e) `localStorage["aoe-resolved-theme"]` contains the new theme, (f) the `aoe:theme-changed` event fired (capture via `page.addInitScript`), (g) `PATCH /api/profiles/:name/settings` returned 2xx with the new `theme.name`, (h) `GET /api/theme/current` after the PATCH returns the same name.
- **Pick a theme under passphrase auth.** Same assertions; the previously-reverted `settings-theme-select-passphrase.spec.ts` shape, hardened against the silent-revert UX. This spec will fail today; that is the point.
- **Pick a theme under token auth.** Same assertions; third auth mode for parity.

#### Persistence + replay

- **Theme survives a full page reload.** Pick, await PATCH, reload, expect `dataset.theme` and `localStorage["aoe-resolved-theme"]` to match before any client-side fetch (anti-FOUC path).
- **Theme survives an `aoe serve` restart.** Pick, await PATCH, stop the harness server, restart it (or re-spawn via `spawnAoeServe` with the same `HOME`), navigate again; theme persists.
- **Theme survives a session switch.** Pick, navigate to a different session, navigate back; dataset and CSS vars match.
- **Cached theme paints before React hydrates.** Set `localStorage["aoe-resolved-theme"]` to a non-default payload, `page.goto("/")`, assert `dataset.theme` is set before `[data-testid="cockpit-view"]` (or any other late-mount marker) is visible.

#### Failure modes (server says no)

- **`PATCH /api/profiles/:name/settings` 403 under passphrase auth: select must NOT silently revert.** Stub the route to 403; expect: (a) select shows the user's chosen value (no rollback to the server's stale one), (b) an error toast / inline message surfaces, (c) `document.documentElement.dataset.theme` does NOT change (we did not persist; we should not paint).
- **`PATCH` 500: same as above with a different status.** Surfaces the same UX bug as the 403; lock the assertion.
- **`PATCH` succeeds but `GET /api/theme/current` then 500s on refetch.** The picker-fetch race; `useResolvedTheme`'s sequence numbers should prevent a stale paint from overwriting the new one.
- **Network partition during `PATCH`: select shows pending state, then a clear failure on timeout.** Stub the route with `route.abort('failed')`.

#### Race conditions

- **Fast picker change while mount fetch is in flight.** Stub `GET /api/theme/current` to delay 2s; navigate to /settings; pick a different theme within 500ms; expect the user's pick to win (sequence-number branch at `useResolvedTheme.ts:36-44`).
- **Two picks in quick succession.** Pick A, pick B within 200ms; final state is B; the second PATCH body is the steady state.

#### Cross-surface effects

- **Theme change repaints the xterm terminal.** On a tmux session with a visible terminal, pick a new theme; assert the terminal's foreground CSS var (`--term-foreground` or whichever the theme exposes) updates. Indirectly verifies the `aoe:theme-changed` listener in `useTerminal.ts:1260-1262`.
- **Theme change repaints Shiki syntax highlighting.** On a cockpit message containing a fenced code block, pick a new theme; assert the rendered highlight token colors change. Indirectly verifies `useShikiTheme.ts` and the `aoe:theme-changed` listener.
- **Theme change repaints rendered tool cards.** ToolCards.tsx and Markdown.tsx both read theme tokens at render time; spot-check at least one tool card kind and one markdown render.

#### Per-profile isolation

- **Profile A theme is independent of Profile B.** Set theme on A, switch profile to B; expect dataset.theme reverts to B's value (or default). Switch back to A; A's theme is restored.
- **Renaming a profile preserves its theme.** Pick theme on A, rename A → A2, theme on A2 still resolves.

#### Custom themes

- **Custom theme from `~/.agent-of-empires/themes/foo.toml` appears in the dropdown.** Seed the file in the harness's isolated `HOME`, navigate to /settings, assert the option appears alongside builtins; pick it; assert `dataset.theme === "foo"`.
- **Custom theme malformed: dropdown still loads.** Seed a deliberately bad TOML; `fetchThemes` should drop it (or surface a recoverable error), not 500 the whole settings page.

#### Color-mode field (TUI only)

- **Color mode dropdown change PATCHes but does NOT call `dispatchThemePickerChanged`.** Per `ThemeSettings.tsx:26-28`, only the `name` field dispatches. Lock the assertion so a future refactor doesn't accidentally repaint the dashboard on a TUI-only setting.

#### Direct deep-link variants

- **`page.goto("/settings")` under passphrase auth lands on /settings, not /.** The original issue body flags this as a separate bug; spec it here so the regression is owned.

### Harness changes needed

- Promote a `themeFromDocument(page)` helper to `web/tests/helpers/theme.ts` (read `dataset.theme`, `dataset.themeAppearance`, a representative CSS var, the localStorage cache, and the most recent `aoe:theme-changed` event payload via `page.addInitScript` window-side capture).
- Promote a `seedCustomTheme(home, name, payload)` helper for the custom-theme stories.
- `spawnAoeServe` already supports all three auth modes (`none` / `passphrase` / `token`); the harness side is otherwise unchanged.

### Coverage matrix

One new entry per spec, `kind: "live-playwright"` (or `mocked-playwright` for the failure-mode stubs that don't need a real backend), `components: ["web/src/components/settings/ThemeSettings.tsx", "web/src/components/SettingsView.tsx", "web/src/hooks/useResolvedTheme.ts", "web/src/lib/theme.ts"]`. The existing `settings.theme-select` entry from #1397 stays as the no-auth tracer.

### Tracking

This is the tracker issue. File one sub-issue per spec; each can land independently. Bug-fix PRs (the 403 fix, the silent-revert UX fix, the deep-link bug) follow as their own PRs that reference the specific spec they unblock.

## Why now

- `#1397` lands the user-story mandate and the first wave of story specs; theme switching is one of the dashboard flows it covers shallowest (single happy-path no-auth spec).
- The 403 under passphrase auth (the original framing of this issue) is one of at least four bugs in this surface: silent-revert UX, picker-fired-before-persist, deep-link-to-/settings-redirects, and the elevation-gate semantics. A net of specs makes each one fail loudly before a fix.
- Theme tokens are referenced from every cockpit surface (tool cards, markdown, terminal, shiki); a regression in the apply path produces hard-to-diagnose UI breakage. Coverage here is high-leverage.
- The `useResolvedTheme` sequence-number machinery is the kind of subtle correctness work that breaks silently if a refactor "simplifies" it; a race-condition spec locks it.

## Open questions

- **Elevation flow.** Is `requires_elevation` returning the right answer for `PATCH /api/profiles/:name/settings`, and is the client supposed to re-prompt for the passphrase / elevate the cookie before retrying? Or is profile-settings supposed to ride the basic session cookie under passphrase auth? The decision shapes which side gets the fix (server route changes vs client elevation re-prompt).
- **Silent-revert UX fix scope.** Once `sendSave` surfaces real errors, every settings panel is affected (theme is the loudest symptom; sound, logging, tmux all share the path). Is the fix scoped to theme or do all panels get the new error UX in one PR?
- **`page.goto("/settings")` under passphrase**: is the redirect-to-/ a router bug or a login-wall side effect? Spec asserts current behavior, separate sub-issue to fix.
- **Custom-theme malformed handling.** Should `fetchThemes` filter out bad files server-side, or should the client tolerate a partial 200 with broken entries? Either is defensible; story locks whichever wins.

## Out of scope

- TUI theme switching. The color-mode field PATCH is covered above only to assert the web side does NOT repaint; full TUI behavior is its own surface.
- Theme editor / theme creation UI (no such UI exists today; building one is a feature, not a test gap).
- Visual-regression / screenshot diffing of themed surfaces. Token-level assertions are enough; pixel diffing belongs in a separate suite.
- Profile-creation flow (covered by other specs in #1397); theme stories assume the profile already exists.

## Cross-references

- #1383 — user-story Playwright spec tracker; this issue is the theme-switching follow-up under the same mandate.
- #1397 — landing the first wave of story specs; ships `settings-theme-select.spec.ts` no-auth happy path; reverted the passphrase variant pending the server fix.
- #1419 — sibling tracker for sidebar drag-and-drop session reorder stories; same shape, different surface.

### LINKED ISSUE #1419: Sidebar: add user-story Playwright specs for click-and-drag session reorder (activation, suppression, cross-group, read-only, persistence)

## Problem

Workspace reorder via click-and-drag in the left sidebar (dnd-kit `useSortable` at `web/src/components/WorkspaceSidebar.tsx:305-306`) has exactly two test artifacts today and one matrix entry covering them:

- `web/tests/sidebar-drag-reorder.spec.ts` (mocked) — initial-order paint, server-order-verbatim paint, one basic press-and-hold drag with PUT capture.
- `web/tests/live/workspace-ordering.spec.ts` (live) — one round-trip via `PUT /api/workspace-ordering` against the real backend.
- `web/tests/coverage-matrix.json:226-231` — single entry `sidebar.workspace-ordering` pointing at the live spec only; the mocked spec's three cases are not separately tracked.

Per the user-story mandate landing in #1383 / #1397 (`AGENTS.md` second mandate paragraph in that PR), each user-visible dashboard flow should ship with a UI-driven story spec under `web/tests/live/cockpit-stories/`. Drag-to-reorder has rich UX surface (press-and-hold activation, mobile delay vs desktop distance, mid-drag visual lift, read-only gating, click-after-drag suppression, cross-group constraint) and most of it is currently unexercised by either spec. A refactor of `SortableSessionRow` or `handleDragEnd` can ship green CI while breaking everything except the happy-path drag.

### Mocked spec (`web/tests/sidebar-drag-reorder.spec.ts`)

Three tests:

```
"applies the server-supplied ordering on first paint"
"renders the server-returned order verbatim (no client-side resort)"
"press-and-hold drag reorders the row and PUTs the new order"
```

The drag test press-and-holds for 250ms, asserts `class*=ring-2 shadow-lg` mid-drag (`sidebar-drag-reorder.spec.ts:250-252`), then asserts the new visual order and a single PUT body (`sidebar-drag-reorder.spec.ts:262-270`).

### Live spec (`web/tests/live/workspace-ordering.spec.ts`)

One test: `press-and-hold drag reorders and round-trips PUT /api/workspace-ordering` (`workspace-ordering.spec.ts:82`). It seeds three sessions in one repo via `aoe add`, drags bottom row onto top row with the same press-and-hold sequence as the mocked spec, asserts the `PUT` body, then re-fetches `GET /api/sessions` to confirm the server-persisted `workspace_ordering` matches.

### Production code surface that no spec exercises

- **MouseSensor distance activation (`WorkspaceSidebar.tsx:1048`)**: `{ distance: 8 }`. A stationary click on a row must navigate, not drag. No spec covers the "stationary click still navigates" path.
- **TouchSensor delay activation (`WorkspaceSidebar.tsx:1049`)**: `{ delay: 150, tolerance: 8 }`. A scroll-flick on mobile must scroll, not drag. No spec drives the touch path.
- **Cross-group constraint (`WorkspaceSidebar.tsx:1058-1064`)**: `handleDragEnd` finds the active group, reorders inside it; `over` from a different group's `SortableContext` would not match a workspace id in the active group's list. Neither spec has more than one repo group, so cross-group attempts are not tested.
- **Read-only gating (`WorkspaceSidebar.tsx:306,337,339,1239`)**: `useSortable({ disabled: readOnly })` plus `onDragEnd={readOnly ? undefined : handleDragEnd}` plus omitted `aria-roledescription`. No spec mounts the sidebar under read-only to confirm a viewer can't reorder.
- **Click-after-drag suppression (`WorkspaceSidebar.tsx:274-290,307-318`)**: `useSuppressClickAfterDrag` swallows the synthetic click Chromium dispatches after `mouseup` for ~250ms. Without it, the drag release would re-activate the dropped row. No spec asserts that releasing on a non-active row does not navigate to it.
- **Optimistic active-row hint (`WorkspaceSidebar.tsx:1038-1042`)**: a row clicked during the in-flight `onSelect` shows as active before `activeId` updates. If a drag accidentally fires a click on the source row (suppression broken), the optimistic hint would flicker. No spec.
- **Mid-drag z-index lift (`WorkspaceSidebar.tsx:325`)**: `zIndex: isDragging ? 10 : "auto"`. No spec asserts the dragged row stays unclipped above its neighbors.
- **Slow-drag suppression window extension (`WorkspaceSidebar.tsx:308-317`)**: while `isDragging`, `dragSuppressRef.current` is continuously pushed to `now+1000`; this protects a 5-second slow drag from racing the suppression window. No spec covers a slow drag.
- **Sortable-wrapper vs inner-Link target (`WorkspaceSidebar.tsx:334-352`)**: drag listeners are on the outer `<div ref={setNodeRef}>`; an `mousedown` on the inner `<a>` bypasses the sensor entirely. The existing tests press near `boundingBox.x + width - 4` to avoid this; a story spec should also assert that pressing on the link region without movement navigates (i.e. the inner Link is still clickable through the wrapper).

### Backend wiring

`PUT /api/workspace-ordering` at `src/server/api/sessions.rs:506` (`update_workspace_ordering`); merge logic at `src/server/api/sessions.rs:3644-3691` (`compute_merged_ordering`); persisted via `WorkspaceOrder` at `src/session/mod.rs:53`. The live spec round-trips this once; reload-persistence and concurrent-drag conflict are not covered.

## Root cause

`#1169` landed drag-to-reorder and the initial mocked spec; `#1220` followed up with the live spec. Both PRs scoped tests to the happy path because the harness ergonomics for dnd-kit (press-and-hold timing, MouseSensor vs TouchSensor) were the load-bearing question; "does it reorder" beat "does every branch behave." `#1383`'s user-story mandate now requires the rest of the branches to land as their own specs.

## Proposal

Add a `web/tests/live/cockpit-stories/sidebar-reorder-*.spec.ts` family, mirroring the layout #1397 lands for cockpit / wizard / settings stories. Each spec drives the React UI through real input and asserts on rendered DOM plus (where relevant) the persisted server state and the captured `PUT` body. Use `serveCockpit` / `spawnAoeServe` plus a `seedThreeSessionsInOneRepo`-shaped helper extracted from the existing live spec.

#### Activation

- **Stationary click on a row still navigates (desktop).** Press-and-release without moving 8px; expect `useSortable` to never fire `onDragEnd`, the route to change to `/session/:id`, and no `PUT /api/workspace-ordering` to fly.
- **A drag of less than 8px does not reorder (desktop).** Move 4px under mouse-down, release; row order unchanged; no PUT.
- **A touch tap on mobile still navigates.** iPhone-13 emulation, tap a row; navigation happens; no PUT.
- **A touch press shorter than 150ms then move does not reorder (mobile).** Touch-start, move within 100ms, touch-end; no PUT, route may or may not change but order is unchanged.
- **A touch flick to scroll does not reorder (mobile).** Touch-start near top, move >tolerance vertically within delay; scroll happens, no PUT.
- **A touch press-and-hold past 150ms then drag DOES reorder (mobile).** Synthesize a real `TouchEvent` sequence as documented in `AGENTS.md` legacy mobile recipe; assert PUT body and DOM order.

#### Visual feedback

- **Source row gets ring + shadow during drag.** Already partially covered by the mocked spec; promote to a story spec that also asserts the ring clears on `mouseup` and that no neighbor row has the ring class at any frame.
- **Source row z-index lifts above neighbors during drag.** Computed style on the wrapper has `z-index: 10` while `isDragging`.

#### Click-after-drag suppression

- **Releasing the drag on a non-active row does NOT navigate to it.** The classic regression: a stale click after `mouseup` lands on a row and activates it; `useSuppressClickAfterDrag` should swallow it. Assert via `page.route('**/cockpit/replay**', ...)` count or just `expect(page.url())` unchanged.
- **A real click 300ms after a drag release DOES navigate.** Suppression window is ~250ms; a deliberate later click must work. Round-trip the URL change.
- **A slow drag (5 seconds with `mouseup` at the end) still suppresses the trailing click.** Exercises the `isDragging` branch that keeps pushing `dragSuppressRef.current` forward.

#### Cross-group constraint

- **Dragging onto a row in a different repo group is a no-op.** Seed two repos, three sessions each. Press-and-hold a row in group A, drag onto a row in group B, release; both groups' orders unchanged; no PUT.
- **Dragging onto the group header does nothing.** The header is not a sortable item; collision should not match.

#### Read-only mode

- **Read-only viewer cannot drag rows.** Mount the sidebar with `readOnly`. Wrapper has no `aria-roledescription`; `useSortable` is `disabled`. Attempt a press-and-hold + move; assert the row does not visually lift (no `ring-2` class transition) and no `PUT` fires.

#### Persistence + replay

- **Reload after a drag preserves the new order.** Drag, await PUT, reload the page; the new visual order is rendered from the server before any client-side resort.
- **Two consecutive drags merge into the expected final order.** Drag A→top, then drag B→top; final order has B first, A second. Two PUTs fly; the second one's body is the steady state.
- **`GET /api/sessions` returns the merged ordering envelope after the drag.** Live spec covers this for one drag; story version makes the assertion explicit and adds the post-reload variant.

#### Inner-Link interaction

- **A click on the inner Link's status glyph still navigates.** Press-release on the leftmost part of the row (close to the inner `<a>`); ensure no `PUT` flies and navigation happens. Today's tests deliberately press at `width - 4` to avoid this region; a story spec must lock the opposite assertion.
- **A double-click on a row does not drag.** Two fast clicks; no PUT.

#### Failure modes

- **A failed `PUT /api/workspace-ordering` (server 500) reverts the visual order.** Stub the route to reject; assert the optimistic local reorder rolls back. (If the current behavior is "leave the optimistic order and never recover," that's its own bug and the spec asserts the existing contract while linking the follow-up.)
- **A network partition during the drag is recoverable on next reorder.** Drag with the route stubbed to never respond; abort; second drag with the route working; final state matches the second drag.

### Harness changes needed

Minimal. Most pieces already exist:

- `seedThreeSessionsInOneRepo` is local to `workspace-ordering.spec.ts:31-60`; promote it to `web/tests/helpers/sessions.ts` (or wherever multi-session seeding lands during #1397) so all sidebar-reorder story specs can share it.
- `readVisibleSessionTitles` is local to the same file at line 62-79; same treatment.
- Mobile stories need the synthetic `TouchEvent` recipe documented in `AGENTS.md` ("Legacy mobile/touch recipe"). Confirm that `page.touchscreen.tap()` triggers the `TouchSensor` delay correctly before settling on the `page.evaluate` + `new TouchEvent` route.
- Read-only story needs a way to flip the sidebar into read-only. The existing `web/tests/live/read-only-mode.spec.ts` already drives this via auth mode; reuse that path.
- Cross-group story needs a `seedTwoReposThreeSessionsEach` helper; trivial extension of the existing seed function.

### Coverage matrix

One new entry per spec, `kind: "live-playwright"` (or `mocked-playwright` for the failure-mode stubs that don't need a real server), `components: ["web/src/components/WorkspaceSidebar.tsx"]`. The existing `sidebar.workspace-ordering` entry stays as the umbrella tracer; new entries get more specific ids (`sidebar.reorder.stationary-click-navigates`, `sidebar.reorder.cross-group-noop`, etc.).

### Tracking

This is a tracker issue in the same shape as #1383. File one sub-issue per spec (or a small cluster per shared harness piece). Each sub-issue can land independently; they share the helper extraction and the matrix-entry pattern, not the assertion plan.

## Why now

- `#1397` lands ~83 user-story specs for cockpit / wizard / settings / mobile under the new `AGENTS.md` mandate, but explicitly does not touch the sidebar reorder surface. That gap is the loudest one left in the sidebar coverage column.
- The drag activation thresholds (`distance: 8`, `delay: 150`, `tolerance: 8`) are the kind of constants a tuning change can quietly bend; without per-branch coverage a PR that bumps any of them ships green.
- Click-after-drag suppression (`WorkspaceSidebar.tsx:274-290`) is the load-bearing piece of UX nobody would notice missing until it broke; "drop a row, the dropped row navigates instead of staying put" is a regression class worth a regression test.
- `#1418` (web sidebar opt-in last-activity sort) is open and will touch the ordering path; landing story-level coverage first means that PR's CI has something to break.

## Open questions

- **Touch synthesis.** Real Playwright `page.touchscreen.tap` is single-finger and may not satisfy the `TouchSensor` activation reliably; the `new TouchEvent` recipe in `AGENTS.md` is documented for the toolbar tests but not yet for dnd-kit. First spec to land should pin which approach works for the 150ms delay sensor and the rest follow that shape.
- **Failure-mode rollback.** I am not certain the current client rolls back the optimistic order on a 500 from `PUT /api/workspace-ordering`. The spec should be written to assert the CURRENT behavior and link a follow-up issue if that behavior is "leak the optimistic state until reload."
- **Mocked vs live split.** Activation, suppression, and visual-feedback stories do not need a real server; they belong in mocked Playwright (`web/tests/*.spec.ts`) for speed. Persistence, reload, and cross-group stories need a live server. The matrix should split accordingly.
- **Cross-group with multi-repo workspace rows.** The `MULTI_REPO_GROUP_ID` group at `WorkspaceSidebar.tsx:1253-1257` is special-cased; need to decide whether a drag inside that group is even possible (today every group has its own `SortableContext`, so yes) and whether reorder rules differ there.
- **Keyboard reorder.** `WorkspaceSidebar.tsx:332-333` deliberately omits dnd-kit's keyboard `attributes`. A story asserting "no keyboard reorder is possible" would lock that decision in; if a future a11y push wants to add keyboard reorder, the story has to flip.

## Out of scope

- Touch-velocity / momentum tuning for the mobile reorder gesture. Synthetic touchmove timing limitations are documented in `AGENTS.md`; a story spec that asserts "press and hold past 150ms reorders" is enough without modeling velocity.
- Visual-regression / screenshot diffing of the drag ring + shadow.
- Adding keyboard reorder support (would be a feature PR, not a test PR).
- Reordering groups themselves (today drag is constrained to within one group; cross-group is asserted as no-op above, not as a feature gap).

## Cross-references

- #1383 — user-story Playwright mandate; this issue is the sidebar-reorder follow-up under the same mandate.
- #1397 — landing the first ~83 user-story specs; sets the `cockpit-stories/` pattern and the `AGENTS.md` second mandate paragraph this issue depends on.
- #1169 — original PR that landed drag-to-reorder and the first mocked spec.
- #1220 — follow-up that added the live `workspace-ordering` round-trip spec.
- #1171 — pinned sidebar order to `created_at desc` with no status reshuffle; sets the baseline this reorder layers on top of.
- #1418 — open; opt-in last-activity sort. Will touch the ordering path that these stories cover.

Seluj78:

## Description

Lands the missing user-story Playwright coverage flagged by two open tracker issues. Pure test additions; no production code changes.

Closes #1405
Closes #1419

**#1405 (theme switching, 4 new live specs under `web/tests/live/cockpit-stories/`):**

- `settings-theme-custom.spec.ts`: a TOML file dropped into `${appDir}/themes/` appears in the dropdown and applies its palette when picked. Locks the `discover_custom_themes` / `available_themes` path.
- `settings-theme-custom-malformed.spec.ts`: a malformed TOML in the same directory does not break the dropdown; builtins still load and apply.
- `settings-theme-color-mode.spec.ts`: changing the color_mode field PATCHes but does not dispatch `aoe:theme-picker-changed` (locks the early-return at `ThemeSettings.tsx:34` so a refactor cannot accidentally repaint the dashboard on a TUI-only setting).
- `settings-theme-failure-revert.spec.ts`: a 500 on PATCH does not silently change `dataset.theme` or fire the picker event (locks the fix that landed in #1575 / #1510, browser-side route stub against a real serve so only the write path 500s).

Plus a small `web/tests/helpers/theme.ts` with `seedCustomTheme` + a couple of read accessors so the two custom-theme specs do not duplicate the TOML body or the appDir resolution.

**#1419 (sidebar drag-reorder, 9 new specs):**

Desktop activation + suppression (5 mocked):

- `sidebar-reorder-stationary-click-navigates.spec.ts`: a press-release with no movement still navigates (locks the `{distance: 8}` MouseSensor threshold).
- `sidebar-reorder-tiny-drag-noop.spec.ts`: a 4px drag stays below the activation threshold; row order unchanged, no PUT.
- `sidebar-reorder-click-after-drag-suppressed.spec.ts`: the synthetic click after a drag release does not navigate to the dropped-on row (locks `useSuppressClickAfterDrag`).
- `sidebar-reorder-cross-group-noop.spec.ts`: dragging a row onto a different repo group is a no-op (`handleDragEnd` short-circuits at the `over.id` check).
- `sidebar-reorder-read-only.spec.ts`: read-only viewers have no drag affordance and no PUT fires (locks `useSortable({ disabled })` and the missing aria-roledescription).

Mobile (3 mocked, `devices['iPhone 13']` viewport, CDP `Input.dispatchTouchEvent`):

- `sidebar-reorder-mobile-tap-navigates.spec.ts`: a tap on a row navigates; never enters the TouchSensor's 150ms delay window.
- `sidebar-reorder-mobile-flick-scrolls.spec.ts`: a fast vertical flick cancels TouchSensor activation; no PUT.
- `sidebar-reorder-mobile-press-hold-reorders.spec.ts`: press past 150ms then drag DOES reorder. This pins the touch synthesis recipe (CDP-driven `Input.dispatchTouchEvent` instead of synthetic `TouchEvent` dispatch) as the working approach for the 150ms delay sensor, which the issue explicitly flagged as an open question for the first spec to land.

Persistence (1 live):

- `sidebar-reorder-reload-persists.spec.ts`: after a drag, a full `page.reload()` reads back the server-supplied ordering and paints it before any client-side resort.

Plus `web/tests/helpers/sidebarMocks.ts` to hoist the mocked-Playwright session / ordering / read-only setup, and a small refactor that promotes `seedSessionsInRepo` + `readVisibleSessionTitles` from `tests/live/workspace-ordering.spec.ts` to `web/tests/helpers/sidebar.ts` so the new live story spec can share them.

11 new `coverage-matrix.json` entries cover the new specs; `tests/validate-coverage-matrix.mjs` passes.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI
- [x] Tests

## Checklist

- [x] Documentation was updated where necessary (no docs changes; behavior unchanged)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## How to test

Run from `web/`:

- [ ] `npx playwright test sidebar-reorder --config=playwright.config.ts` passes all 8 mocked sidebar story specs (stationary click, tiny drag, click-after-drag, cross-group, read-only, mobile tap, mobile flick, mobile press-hold).
- [ ] `AOE_E2E_BINARY=$(realpath ../target/debug/aoe) npx playwright test cockpit-stories/settings-theme cockpit-stories/sidebar-reorder-reload-persists --config=playwright.live.config.ts` passes all 5 live story specs (custom theme, malformed custom, color-mode dispatch contract, PATCH failure, reload persistence).
- [ ] `node tests/validate-coverage-matrix.mjs` reports `Coverage matrix validation passed.`
- [ ] Existing `workspace-ordering` live spec still passes after the helper refactor: `AOE_E2E_BINARY=$(realpath ../target/debug/aoe) npx playwright test workspace-ordering --config=playwright.live.config.ts`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, scoping, and spec drafting handled by Claude under my direction. I picked the subset of stories to land in this PR (chiefly: skip the cross-surface terminal/shiki repaint stories and the deep-link-under-passphrase bug story, both worth a follow-up issue if they bite), pinned the CDP touch-event recipe as the answer to the issue's open question, and reviewed every commit before pushing.

## Overview

This PR adds comprehensive Playwright test coverage for two features — theme switching and sidebar drag-and-drop reordering — across both mocked and live test environments. No production code changes are made; this is purely test infrastructure.

## What Was Added

**Theme Switching Tests (4 specs)**
- Custom theme loading and dropdown verification
- Malformed theme handling (ensure broken themes don't break the UI)
- Color-mode changes without triggering UI repaints
- Failure handling (500 errors revert changes silently)

**Sidebar Reorder Tests (9 specs)**
- Desktop mocked tests: stationary clicks navigate, tiny drags are ignored, click suppression after drag, cross-group drag prevention, read-only user restrictions
- Mobile mocked tests (iPhone 13): tap navigation, flick-scroll no-op, press-and-hold reordering using CDP touch events
- Live persistence test: verify reordered sidebar persists after page reload

**Test Helpers (3 new files)**
- `web/tests/helpers/theme.ts`: TOML seeding, theme document snapshots, constants for valid/malformed themes
- `web/tests/helpers/sidebarMocks.ts`: API mocking for sidebar sessions, workspace ordering, read-only mode
- Hoisted functions in `web/tests/helpers/sidebar.ts`: session seeding, visible title reading for reuse

**Coverage Matrix**
- 11 new entries documenting the new test surfaces

## What Was Refactored

- Hoisted `seedSessionsInRepo` and `readVisibleSessionTitles` helpers from `workspace-ordering.spec.ts` into shared `sidebar.ts` for reuse by new specs
- Refactored `workspace-ordering.spec.ts` to use the shared helpers (reduced from 58 lines to more concise usage)

## What Was Improved

- Increased poll timeout in `wheel-scroll.spec.ts` from 5s to 10s to reduce CI flakiness from increased test contention

## Benefits

- **Regression Protection**: Comprehensive coverage for theme customization and sidebar interactions ensures future changes don't break these features
- **Maintainability**: Hoisted helpers reduce duplication across test files
- **Mobile Coverage**: Touch event testing (via CDP) pins down mobile-specific behaviors like flick-scroll and press-hold gestures
- **Real-world Validation**: Live tests confirm end-to-end behavior including persistence and error handling
- **Improved CI Stability**: Adjusted timeout prevents timing-related failures as test suite grows

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1597?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/tests/coverage-matrix.json
+++ b/web/tests/coverage-matrix.json
@@ -432,6 +432,105 @@
       ]
     },
     {
+      "id": "sidebar.reorder-stationary-click",
+      "kind": "mocked-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/sidebar-reorder-stationary-click-navigates.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-tiny-drag-noop",
+      "kind": "mocked-playwright",
+      "risk": "low",
+      "specs": [
+        "tests/sidebar-reorder-tiny-drag-noop.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-click-after-drag-suppressed",
+      "kind": "mocked-playwright",
+      "risk": "high",
+      "specs": [
+        "tests/sidebar-reorder-click-after-drag-suppressed.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-cross-group-noop",
+      "kind": "mocked-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/sidebar-reorder-cross-group-noop.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-read-only",
+      "kind": "mocked-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/sidebar-reorder-read-only.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-mobile-tap-navigates",
+      "kind": "mocked-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/sidebar-reorder-mobile-tap-navigates.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-mobile-flick-scrolls",
+      "kind": "mocked-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/sidebar-reorder-mobile-flick-scrolls.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "sidebar.reorder-mobile-press-hold",
+      "kind": "mocked-playwright",
+      "risk": "high",
+      "specs": [
+        "tests/sidebar-reorder-mobile-press-hold-reorders.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
+      "id": "story.sidebar-reorder-reload-persists",
+      "kind": "live-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/live/cockpit-stories/sidebar-reorder-reload-persists.spec.ts"
+      ],
+      "components": [
+        "web/src/components/WorkspaceSidebar.tsx"
+      ]
+    },
+    {
       "id": "sidebar.client-side-navigation",
       "kind": "mocked-playwright",
       "risk": "medium",
@@ -1272,6 +1371,54 @@
       ],
       "components": [
         "web/src/components/settings/ThemeSettings.tsx"
+      ]
+    },
+    {
+      "id": "story.settings-theme-custom",
+      "kind": "live-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/live/cockpit-stories/settings-theme-custom.spec.ts"
+      ],
+      "components": [
+        "web/src/components/settings/ThemeSettings.tsx",
+        "web/src/lib/theme.ts"
+      ]
+    },
+    {
+      "id": "story.settings-theme-custom-malformed",
+      "kind": "live-playwright",
+      "risk": "low",
+      "specs": [
+        "tests/live/cockpit-stories/settings-theme-custom-malformed.spec.ts"
+      ],
+      "components": [
+        "web/src/components/settings/ThemeSettings.tsx"
+      ]
+    },
+    {
+      "id": "story.settings-theme-color-mode",
+      "kind": "live-playwright",
+      "risk": "low",
+      "specs": [
+        "tests/live/cockpit-stories/settings-theme-color-mode.spec.ts"
+      ],
+      "components": [
+        "web/src/components/settings/ThemeSettings.tsx",
+        "web/src/hooks/useResolvedTheme.ts"
+      ]
+    },
+    {
+      "id": "story.settings-theme-failure-revert",
+      "kind": "live-playwright",
+      "risk": "medium",
+      "specs": [
+        "tests/live/cockpit-stories/settings-theme-failure-revert.spec.ts"
+      ],
+      "components": [
+        "web/src/components/settings/ThemeSettings.tsx",
+        "web/src/components/SettingsView.tsx",
+        "web/src/hooks/useResolvedTheme.ts"
       ]
     },
     {
--- a/web/tests/helpers/sidebar.ts
+++ b/web/tests/helpers/sidebar.ts
@@ -1,4 +1,8 @@
+import { spawnSync } from "node:child_process";
+import { mkdirSync } from "node:fs";
+import { join } from "node:path";
 import type { Page } from "@playwright/test";
+import { resolveAoeBinary } from "./aoeServe";
 
 export async function clickSidebarSession(page: Page, title: string) {
   // Sidebar session rows are anchor tags (see WorkspaceSidebar.tsx). The
@@ -66,3 +70,98 @@ export async function openMobileSidebar(page: Page) {
     { timeout: 5_000 },
   );
 }
+
+/**
+ * Read the visible session-row titles in the sidebar, in DOM order.
+ *
+ * Scopes to the `.truncate` label span inside each row so a Wakeup or
+ * Plan chip on the same row can't be confused with the workspace
+ * title. Hoisted from `tests/live/workspace-ordering.spec.ts` so the
+ * reorder story specs can share the same accessor without redefining
+ * it per file.
+ */
+export async function readVisibleSessionTitles(page: Page): Promise<string[]> {
+  return page.evaluate(() => {
+    const rows = Array.from(
+      document.querySelectorAll<HTMLElement>(
+        "[data-testid='sidebar-session-row']",
+      ),
+    );
+    return rows
+      .map(
+        (r) =>
+          r.querySelector("span.truncate[title]")?.getAttribute("title") ?? "",
+      )
+      .filter(Boolean);
+  });
+}
+
+/**
+ * Build a `seedFn` for `spawnAoeServe` that git-inits a project dir
+ * under the isolated HOME, then runs `aoe add` once per supplied
+ * title. The arrival order matters: the server prepends new workspace
+ * ids newest-first, so the seeded sidebar order in arrival sequence
+ * is `titles[-1]` at the top down to `titles[0]` at the bottom.
+ *
+ * Optional `subdir` lets callers seed multiple repos by pointing each
+ * `seedSessionsInRepo` call at a different directory; the cross-group
+ * reorder story chains two calls for that reason.
+ */
+export function seedSessionsInRepo(opts: {
+  titles: string[];
+  subdir?: string;
+  tool?: string;
+}): (seedEnv: {
+  home: string;
+  shimBin: string;
+  env: NodeJS.ProcessEnv;
+}) => void {
+  return ({ home, env }) => {
+    const binary = resolveAoeBinary();
+    const projectDir = join(home, opts.subdir ?? "repo");
+    mkdirSync(projectDir, { recursive: true });
+    spawnSync("git", ["init", "-q"], { cwd: projectDir });
+    spawnSync("git", ["commit", "--allow-empty", "-q", "-m", "init"], {
+      cwd: projectDir,
+      env: {
+        ...env,
+        GIT_AUTHOR_NAME: "t",
+        GIT_AUTHOR_EMAIL: "t@t",
+        GIT_COMMITTER_NAME: "t",
+        GIT_COMMITTER_EMAIL: "t@t",
+      },
+    });
+    for (const title of opts.titles) {
+      const res = spawnSync(
+        binary,
+        ["add", projectDir, "-t", title, "-c", opts.tool ?? "claude"],
+        { env },
+      );
+      if (res.status !== 0) {
+        throw new Error(
+          `aoe add failed for ${title}: status=${res.status} stderr=${res.stderr?.toString() ?? "<none>"}`,
+        );
+      }
+    }
+  };
+}
+
+/**
+ * Chain multiple repo seeds into a single `seedFn`. Cross-group
+ * reorder stories need at least two repos so dnd-kit's per-group
+ * SortableContext renders multiple drag boundaries; this helper runs
+ * each repo's `seedSessionsInRepo` in sequence under the same env.
+ */
+export function seedRepos(
+  repos: Array<{ titles: string[]; subdir: string; tool?: string }>,
+): (seedEnv: {
+  home: string;
+  shimBin: string;
+  env: NodeJS.ProcessEnv;
+}) => void {
+  const fns = repos.map((r) => seedSessionsInRepo(r));
+  return (seedEnv) => {
+    for (const fn of fns) fn(seedEnv);
+  };
+}
+
--- a/web/tests/helpers/sidebarMocks.ts
+++ b/web/tests/helpers/sidebarMocks.ts
@@ -1,0 +1,175 @@
+// Shared mocked-Playwright setup for the sidebar-reorder story specs
+// under `web/tests/sidebar-reorder-*.spec.ts` (#1419). The drag-to-
+// reorder UI sits on top of dnd-kit and a handful of REST endpoints;
+// the per-test specifics are just the input gesture + the assertion.
+// Extracted from `sidebar-drag-reorder.spec.ts` and trimmed to what
+// the story specs actually need.
+
+import type { Page, Route } from "@playwright/test";
+
+export interface MockSessionInput {
+  id: string;
+  title: string;
+  project_path: string;
+  branch: string | null;
+  /** ISO 8601 timestamp; controls newest-first default ordering. */
+  created_at?: string;
+}
+
+type MockSession = Required<MockSessionInput>;
+
+function fillCreatedAt(s: MockSessionInput, fallbackIndex: number): MockSession {
+  return {
+    ...s,
+    // Stagger fallback timestamps a day apart so a list seeded in
+    // arrival order has a deterministic newest-first sort if anything
+    // ever falls back to created_at.
+    created_at:
+      s.created_at ??
+      new Date(Date.UTC(2025, 0, 1 + fallbackIndex)).toISOString(),
+  };
+}
+
+function sessionResponse(s: MockSession) {
+  return {
+    id: s.id,
+    title: s.title,
+    project_path: s.project_path,
+    group_path: s.project_path,
+    tool: "claude",
+    status: "Idle",
+    yolo_mode: false,
+    created_at: s.created_at,
+    last_accessed_at: null,
+    idle_entered_at: null,
+    last_error: null,
+    branch: s.branch,
+    main_repo_path: null,
+    is_sandboxed: false,
+    has_terminal: true,
+    profile: "default",
+    workspace_repos: [],
+  };
+}
+
+/** Workspace id format used by the server: `<project_path>::<branch>`
+ *  for branched sessions, `<project_path>::__session__::<id>` for
+ *  ones without a branch. Mirrors `useWorkspaces.ts:31`. */
+export function workspaceId(s: { project_path: string; branch: string | null; id: string }): string {
+  return s.branch
+    ? `${s.project_path}::${s.branch}`
+    : `${s.project_path}::__session__::${s.id}`;
+}
+
+export interface SidebarMockHandle {
+  /** Recorded `PUT /api/workspace-ordering` bodies in arrival order. */
+  puts: Array<{ order?: string[] }>;
+  /** Override the fulfill response for the next PUT (used by the
+   *  failure-mode story). Reset after each call. */
+  nextPutResponse: { status?: number; body?: string } | null;
+  /** Override the read_only flag from `/api/about` (used by the
+   *  read-only story). */
+  readOnly: boolean;
+}
+
+export interface SidebarMockOptions {
+  sessions: MockSessionInput[];
+  /** Server-supplied workspace ordering (the full list, with each id
+   *  composed via `workspaceId`). Defaults to the input order. */
+  ordering?: string[];
+  /** Add `read_only: true` to the `/api/about` response. */
+  readOnly?: boolean;
+}
+
+/** Install routes for the surface the sidebar uses. Returns a handle
+ *  the test can read for captured PUT bodies and tweak the next PUT's
+ *  fulfill (for the failure-mode story). */
+export async function installSidebarMocks(
+  page: Page,
+  opts: SidebarMockOptions,
+): Promise<SidebarMockHandle> {
+  const filled = opts.sessions.map((s, i) => fillCreatedAt(s, i));
+  const handle: SidebarMockHandle = {
+    puts: [],
+    nextPutResponse: null,
+    readOnly: !!opts.readOnly,
+  };
+
+  const ordering = opts.ordering ?? filled.map((s) => workspaceId(s));
+
+  await page.route("**/api/login/status", (r) =>
+    r.fulfill({ json: { required: false, authenticated: true } }),
+  );
+  await page.route("**/api/sessions", (r) => {
+    if (r.request().method() !== "GET") return r.fulfill({ status: 400 });
+    return r.fulfill({
+      json: { sessions: filled.map(sessionResponse), workspace_ordering: ordering },
+    });
+  });
+  await page.route("**/api/workspace-ordering", (r: Route) => {
+    if (r.request().method() === "PUT") {
+      const body = r.request().postDataJSON() as { order?: string[] };
+      handle.puts.push(body ?? {});
+      const override = handle.nextPutResponse;
+      handle.nextPutResponse = null;
+      if (override) return r.fulfill(override);
+    }
+    return r.fulfill({ json: { order: [] } });
+  });
+  await page.route("**/api/about", (r) =>
+    r.fulfill({
+      json: {
+        read_only: handle.readOnly,
+        auth_mode: "none",
+        behind_tunnel: false,
+        profile: "default",
+      },
+    }),
+  );
+  for (const path of [
+    "settings",
+    "themes",
+    "agents",
+    "profiles",
+    "groups",
+    "devices",
+  ]) {
+    await page.route(`**/api/${path}`, (r) =>
+      r.fulfill({ json: [] }),
+    );
+  }
+  await page.route("**/api/docker/status", (r) =>
+    r.fulfill({ json: {} }),
+  );
+
+  return handle;
+}
+
+/** Standard three sessions in one repo, used by the activation /
+ *  suppression / mobile stories. Title -> branch -> id are aligned so
+ *  the test reads naturally. */
+export function threeSessionsInOneRepo(): MockSessionInput[] {
+  return [
+    {
+      id: "s-a",
+      title: "alpha",
+      project_path: "/tmp/repo",
+      branch: "feature/a",
+      created_at: "2025-03-01T00:00:00Z",
+    },
+    {
+      id: "s-b",
+      title: "beta",
+      project_path: "/tmp/repo",
+      branch: "feature/b",
+      created_at: "2025-02-01T00:00:00Z",
+    },
+    {
+      id: "s-c",
+      title: "gamma",
+      project_path: "/tmp/repo",
+      branch: "feature/c",
+      created_at: "2025-01-01T00:00:00Z",
+    },
+  ];
+}
--- a/web/tests/helpers/theme.ts
+++ b/web/tests/helpers/theme.ts
@@ -1,0 +1,117 @@
+// Shared helpers for theme-switching story specs (#1405).
+//
+// Two domains:
+//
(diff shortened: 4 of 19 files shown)


Seluj78:
@coderabbitai review

PR #1602, opened 2026-05-28
fix(web): hide empty repo group header and clamp sidebar context menu to viewport

### LINKED ISSUE #1600: Web sidebar: empty repo group header stays visible when every session inside is archived or snoozed

## Problem

In the web sidebar, when every session inside a repo group is snoozed or archived, the group header stays visible at the top-level list while its body is empty. Sessions themselves correctly migrate to the bottom "Snoozed & archived" footer, but the now-empty repo header sits orphaned in the live list, taking vertical space and implying there is something to expand into.

Expected: a repo group whose every workspace is sunk (archived or actively snoozed) should disappear from the live sidebar entirely, and reappear the moment any session in it returns to the live tier.

This is a follow-up to #1585 (the triage feature) which introduced the global "Snoozed & archived" footer but did not collapse the now-empty per-repo headers.

## What's there today

`useRepoGroups` builds one `RepoGroup` per `projectPath` (plus synthetic `__multi_repo__` and `__scratch__`) over the full workspace list, without any sink-aware filter:

`web/src/hooks/useRepoGroups.ts:116-136`:

```ts
for (const [repoPath, repoWorkspaces] of byRepo) {
  const sorted = sortWorkspaces(repoWorkspaces);
  const hasActive = sorted.some((ws) => ws.status === "active");
  ...
  repoGroups.push({
    id: repoPath,
    repoPath,
    ...
    workspaces: sorted,
    status: hasActive ? "active" : "idle",
    collapsed,
  });
}
```

`WorkspaceSidebar` then renders one header per group and only filters sunk workspaces out of the body:

`web/src/components/WorkspaceSidebar.tsx:1857-1922`:

```tsx
{filteredGroups.map((group) => {
  const showExpanded = q ? true : !group.collapsed;
  const hasActiveChild = group.workspaces.some(
    (ws) => ws.id === displayedActiveId,
  );
  return (
    <div key={group.id}>
      <RepoGroupHeader
        group={{ ...group, collapsed: !showExpanded }}
        ...
      />
      {showExpanded && (() => {
        const liveWorkspaces = group.workspaces.filter(
          (ws) => !workspaceIsSunk(ws),
        );
        return (
          <SortableContext ...>
            {liveWorkspaces.map((ws) => (
              <SortableSessionRow ... />
            ))}
          </SortableContext>
        );
      })()}
    </div>
  );
})}
```

`workspaceIsSunk` is the canonical "all sessions sunk" aggregator from `web/src/lib/sidebarSort.ts:69-74`:

```ts
export function workspaceIsSunk(ws: Workspace): boolean {
  if (ws.sessions.length === 0) return false;
  return ws.sessions.every(
    (s) => s.archived_at != null || s.snoozed_until != null,
  );
}
```

The global sunk footer at `WorkspaceSidebar.tsx:1934-1937` already flat-maps every sunk workspace from `filteredGroups`, so the sessions themselves are accounted for; only their parent header is the leak.

The filter-query path at `WorkspaceSidebar.tsx:1668-1678` does drop empty groups via `.filter((g) => g.workspaces.length > 0)`, but only when a filter query is typed. The default no-filter render path passes raw `groups` straight through.

## Root cause

Two layers contribute, both fixable in the render path:

1. **Render loop has no "all sunk" guard.** `filteredGroups.map(...)` always emits a `RepoGroupHeader`, even when every workspace inside is sunk. `WorkspaceSidebar.tsx:1857`.
2. **Group-shape derivation is sink-blind.** `useRepoGroups` carries sunk workspaces inside the group structure (correctly, since the bottom footer reads them) but the sidebar render path treats group presence as independent of liveness. `web/src/hooks/useRepoGroups.ts:116-136`.

The simplest fix is render-side: skip the entire group block when `group.workspaces.every(workspaceIsSunk)`. The footer already handles surfacing those sessions, so no rows are lost.

## Proposal

**A. Minimum viable fix (one-liner in render loop).** In `WorkspaceSidebar.tsx:1857`, skip a group whose every workspace is sunk:

```tsx
{filteredGroups
  .filter((g) => g.workspaces.some((ws) => !workspaceIsSunk(ws)))
  .map((group) => {
    ...
  })}
```

This hides the orphan header, leaves the bottom "Snoozed & archived" footer untouched (it scans `filteredGroups` directly, not the post-filter list), and avoids touching `useRepoGroups`.

Drag-and-drop / active-id selection are unaffected: a sunk-only group has no live workspaces, so there is nothing draggable and `displayedActiveId` resolution still routes through the workspace lookup, not the group.

**B. Structural fix (lift filter into derivation).** Add a `liveWorkspaces: Workspace[]` field on `RepoGroup` in `useRepoGroups`, computed once, and have the renderer iterate `groups.filter((g) => g.liveWorkspaces.length > 0)`. Cleaner if more callers come to need the same partition, but currently `WorkspaceSidebar` is the only consumer and it already computes the partition inline, so (A) suffices.

**C. Out of scope for this fix.** Empty synthetic groups (`__multi_repo__`, `__scratch__`) follow the same `workspaces` shape, so (A) covers them transparently; no special-casing required.

## Why now

- Pin/archive/snooze is the headline #1585 feature; a visible "leaks empty headers" defect is the kind of polish gap users report on day one of trying triage.
- One-line render guard; no derivation or API change. Lands cleanly with a focused regression test.
- Easy to forget without a test: once written, the bug stays gone.

- As a web dashboard user, when all sessions in a repo group are archived or snoozed, then the group header is hidden from the sidebar (no orphan empty fold).
  - Suggested test layer: `web/tests/live/triage-archive.spec.ts` (extend) or new live spec.
- As a web dashboard user, when one session in an emptied repo group is unarchived or unsnoozed back to live, then the group header reappears with that session listed.
  - Suggested test layer: live Playwright (round-trip of the above).
- As a web dashboard user, when I archive the sole live session in a repo group, then the sidebar still shows that session in the bottom "Snoozed & archived" footer with its repo chip intact (no data loss from hiding the group).
  - Suggested test layer: Vitest + RTL on `WorkspaceSidebar`, or extend an existing live triage spec.

## Open questions

- Should an emptied group whose currently-active session is still selected keep its header visible to anchor the selection? Current sunk-section already renders the row in the footer, so the active selection has a home; the header would be redundant. Default: hide.
- If the user has manually collapsed an emptied group, then unarchives a session inside, should the group reappear collapsed or expanded? Default: respect the persisted `collapsed` state (no special-case).

## Out of scope

- Changing the partition shape inside `useRepoGroups` (proposal B). Defer until a second consumer needs it.
- Touching the bottom "Snoozed & archived" footer behavior or layout.
- Drag-and-drop semantics for emptied groups (they have no draggable rows, so non-issue).

## Cross-references

- #1581 — original triage feature request (pin/archive/snooze controls), implemented in #1585.
- #1585 — feat PR that introduced the bottom sunk footer but left the per-group header un-collapsed.
- #1589 — refactor follow-up to #1585 (fail-or-rollback on triage persistence), separate concern.

### LINKED ISSUE #1601: Web sidebar: right-click context menu clips off-viewport on bottom session rows (no viewport clamp)

## Problem

Right-clicking on a session row near the bottom of the sidebar opens the context menu below the cursor, where it runs off the bottom of the viewport. The lower options (Snooze, Delete, etc.) are unreachable until the user scrolls or resizes the window. Same defect on the repo-group header context menu, and on the touch long-press path used on mobile.

Expected: the menu opens at the click point but flips upward (or otherwise clamps) when there is not enough room below, so the entire menu is always visible.

## What's there today

Both context menus position the portaled `<div>` with `position: fixed` and raw `clientX` / `clientY`, no measurement of menu height, no viewport clamp.

Session-row mouse path, `web/src/components/WorkspaceSidebar.tsx:696-701`:

```ts
const handleContextMenu = (e: React.MouseEvent) => {
  if (isDeleting) return;
  e.preventDefault();
  closeOtherContextMenus();
  setContextMenu({ x: e.clientX, y: e.clientY });
};
```

Session-row touch long-press path, `web/src/components/WorkspaceSidebar.tsx:710-724`:

```ts
const handleTouchStart = (e: React.TouchEvent) => {
  ...
  const tx = touch.clientX;
  const ty = touch.clientY;
  longPressTimer.current = setTimeout(() => {
    ...
    setContextMenu({ x: tx, y: ty });
  }, 500);
};
```

Session-row menu render, `web/src/components/WorkspaceSidebar.tsx:938-944`:

```tsx
{contextMenu && createPortal(
  <div
    ref={menuRef}
    data-testid="sidebar-context-menu"
    className="fixed z-50 bg-surface-800 border border-surface-700 rounded-lg shadow-lg py-1 min-w-[180px]"
    style={{ left: contextMenu.x, top: contextMenu.y }}
  >
```

Group-header context menu has the identical pattern at `WorkspaceSidebar.tsx:1353-1356, 1370, 1485-1490`. The keyboard path at line 1370 uses `rect.bottom + 4` which is also vulnerable when the group header is the bottom-most element.

The menu body holds: Rename, three notification presets, four triage actions, Delete. That is ~9 items at ~36px each, so ~320px tall. On a viewport short enough to clip a session row at the bottom, the menu almost always overflows.

## Root cause

`left` and `top` are set from raw pointer coordinates with no `min(viewportHeight - menuHeight, y)` clamp and no upward-flip when `y + menuHeight > viewportHeight`. The `menuRef` is wired only for click-outside dismissal in the `useEffect` that follows, not for post-mount measurement. The menu is `position: fixed`, so once rendered there is no automatic correction.

## Proposal

**A. Clamp-and-flip after measure.** When the menu mounts, measure its bounding box and adjust `top` / `left` to fit the viewport:

```ts
useLayoutEffect(() => {
  if (!contextMenu || !menuRef.current) return;
  const rect = menuRef.current.getBoundingClientRect();
  const margin = 8;
  let nextTop = contextMenu.y;
  let nextLeft = contextMenu.x;
  if (nextTop + rect.height + margin > window.innerHeight) {
    nextTop = Math.max(margin, window.innerHeight - rect.height - margin);
  }
  if (nextLeft + rect.width + margin > window.innerWidth) {
    nextLeft = Math.max(margin, window.innerWidth - rect.width - margin);
  }
  if (nextTop !== contextMenu.y || nextLeft !== contextMenu.x) {
    setContextMenu({ x: nextLeft, y: nextTop });
  }
}, [contextMenu]);
```

Applied to both the session-row menu (`WorkspaceSidebar.tsx` around 938) and the group menu (around 1485). Pulled into a shared `useClampedMenuPosition` hook so both call sites stay in sync.

Trade-off: one re-render after first paint when the menu would have overflowed. Acceptable given the menu is opened by user gesture; the visual frame is short.

**B. Pre-compute upward flip.** Read the menu's expected height (either a fixed constant per menu shape, or via a hidden measurement node) and compute the final position before first paint. Cleaner visually but more code; relies on knowing the menu's height up-front, which changes when the triage actions vary (pin vs unpin, snooze vs unsnooze swap items).

**A** is preferable: simpler, robust to menu-shape changes, and the post-mount re-render is invisible in practice.

## Why now

- Direct UX defect on the headline triage feature (#1585); right-clicking the bottom of a long session list is exactly the gesture the new context menu encourages.
- Affects mobile (touch long-press path goes through the same render).
- One shared helper covers two call sites; small surface, regression-testable in Vitest.

- As a web user, when I right-click a session row near the bottom of the viewport, then the context menu fits inside the viewport (bottom edge clamped, opens upward if needed).
  - Suggested test layer: Vitest + RTL on `WorkspaceSidebar` (simulate `contextmenu` with `clientY` near `window.innerHeight`, mock `getBoundingClientRect`), or mocked Playwright with a short viewport.
- As a web user, when I right-click a session row near the right edge of the viewport, then the context menu does not overflow horizontally (right edge clamped).
  - Suggested test layer: Vitest + RTL.
- As a mobile web user, when I long-press a session row near the bottom of the viewport, then the context menu fits on screen (touch path uses the same clamp).
  - Suggested test layer: mocked Playwright with `devices['iPhone 13']`, dispatching `touchstart` + 500ms wait to fire the long-press.

## Open questions

- Should the clamp also apply to the group-header keyboard menu invocation (`WorkspaceSidebar.tsx:1370`, opens at `rect.bottom + 4`)? Default: yes, route through the same `useClampedMenuPosition` helper so all three entry points behave identically.
- Margin from viewport edge: 8px feels right; flag if the design system has a token for it.

## Out of scope

- Restructuring the context menu shape (fewer items, sub-menus, scrollable body). Out of scope; the clamp handles the current shape.
- Animating the flip. Static reposition is enough.
- Any other portaled overlay in the app (e.g. modal positioning).

## Cross-references

- #1581 — original triage feature request.
- #1585 — feat PR that landed the per-row context menu with Pin / Archive / Snooze actions.
- #1600 — sibling sidebar-triage defect (empty repo group header stays visible when all sessions sunk).

Seluj78:

## Description

Two polish bugs that landed with the #1585 sidebar triage feature.

#1600. The web sidebar kept a repo group header in the live list even when every workspace inside it was archived or actively snoozed. Sessions themselves correctly moved into the global "Snoozed & archived" footer, but the orphaned header sat at the top of the empty list, implying there was something to expand into. Fixed by filtering the render loop through a new `repoGroupHasLiveWorkspace` predicate; the bottom sunk footer keeps scanning the unfiltered group list, so sunk sessions remain reachable.

#1601. The session-row and repo-group context menus positioned themselves with `position: fixed` at the raw `clientX` / `clientY` of the click, with no viewport-aware clamp. Right-clicking a session near the bottom of the viewport (or long-pressing on mobile) pushed the lower items (Snooze, Delete, etc.) off-screen and out of reach. Fixed by a pure `clampMenuPosition` helper plus a `useLayoutEffect` that re-positions the menu inside the viewport after first paint. Same wiring covers the session-row and group-header menus.

Closes #1600
Closes #1601

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [x] Web dashboard, archive every session in a repo group; confirm the repo group header disappears from the live sidebar and the session is reachable only via the "Snoozed & archived" footer.
- [x] Unarchive one of those sessions; confirm the group header reappears with that session listed.
- [x] Open the dashboard in a short browser window, scroll to the bottom of the session list, right-click the bottom row, and confirm every menu item (Rename, Notifications presets, Pin, Archive, Snooze, Delete) is visible without scrolling.
- [x] Right-click a repo group header sitting near the bottom of the viewport; confirm the group menu stays inside the viewport.
- [x] On mobile (or an emulated iPhone viewport), long-press a session row near the bottom; confirm the menu fits on screen.
- [x] Right-click somewhere comfortably mid-viewport on a tall window; confirm the menu still opens at the cursor (no spurious offset).

## Checklist

## Test Coverage Analysis

Specifically:
- `web/tests/live/triage-archive.spec.ts` extended with `#1600` regression asserting the group header disappears after archiving the sole live session and reappears after unarchive.
- `web/tests/sidebar-context-menu-clamp.spec.ts` (new) opens both context menus in a short viewport at the bottom rows and asserts the bounding box stays inside the viewport, guarding the `#1601` clamp.
- `web/src/lib/__tests__/sidebarSort.test.ts` extended with predicate coverage; `web/src/lib/__tests__/menuPosition.test.ts` (new) covers the clamp math.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls, and ran the manual test steps.

## Overview (non-technical)

This PR fixes two sidebar regressions:

- Repo group headers that had no live workspaces are now hidden from the live sidebar; sunk sessions remain accessible from the global "Snoozed & archived" footer and the header returns if a workspace becomes live again.
- Sidebar context menus (for session rows and repo-group headers) are prevented from rendering off-screen — menus opened near viewport edges are repositioned to stay visible and very tall menus become scrollable.

## What changed (concise)

- WorkspaceSidebar: repo groups are filtered so groups with no live workspaces are not rendered in the live list.
- Context menus: added a clamp-and-reposition flow that measures the mounted menu and adjusts its fixed-position coordinates so menus fit inside the viewport; menus also get a viewport-aware max-height and become scrollable when too tall.
- New utility and predicate: clamp math helper and a repoGroupHasLiveWorkspace predicate.
- Tests & coverage: Vitest unit tests for clamp math and the group predicate; Playwright tests added/extended to cover archive/regression and menu-clamping scenarios; coverage matrix updated.

## Technical details (if useful)

- New exports: clampMenuPosition(args) and useClampedMenuPosition(...) hook (uses useLayoutEffect + optional ResizeObserver) in web/src/lib/menuPosition.ts; repoGroupHasLiveWorkspace(group) in web/src/lib/sidebarSort.ts.
- WorkspaceSidebar.tsx: uses useClampedMenuPosition with shared menuRef and contextMenu state; context-menu portals use max-height: calc(100vh - 16px) and overflow-y: auto.
- Tests: comprehensive unit cases for clamping (horizontal, vertical, combined, margin behavior, oversized menus) and Playwright specs for short viewports and mobile long-press interactions.

## Benefits

- Removes orphaned group headers and visual clutter while preserving access to sunk sessions.
- Ensures context menus remain fully visible and usable across mouse, keyboard, and touch entry points.
- Adds unit and E2E coverage to prevent regressions for these behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1602?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -970,8 +970,12 @@ export const SessionRow = memo(function SessionRow({
         <div
           ref={menuRef}
           data-testid="sidebar-context-menu"
-          className="fixed z-50 bg-surface-800 border border-surface-700 rounded-lg shadow-lg py-1 min-w-[180px]"
-          style={{ left: contextMenu.x, top: contextMenu.y }}
+          className="fixed z-50 bg-surface-800 border border-surface-700 rounded-lg shadow-lg py-1 min-w-[180px] overflow-y-auto"
+          style={{
+            left: contextMenu.x,
+            top: contextMenu.y,
+            maxHeight: "calc(100vh - 16px)",
+          }}
         >
           <button
             onClick={startRename}
@@ -1541,8 +1545,12 @@ const RepoGroupHeader = memo(function RepoGroupHeader({
         <div
           ref={menuRef}
           data-testid="sidebar-group-context-menu"
-          className="fixed z-50 bg-surface-800 border border-surface-700 rounded-lg shadow-lg py-1 min-w-[190px]"
-          style={{ left: contextMenu.x, top: contextMenu.y }}
+          className="fixed z-50 bg-surface-800 border border-surface-700 rounded-lg shadow-lg py-1 min-w-[190px] overflow-y-auto"
+          style={{
+            left: contextMenu.x,
+            top: contextMenu.y,
+            maxHeight: "calc(100vh - 16px)",
+          }}
         >
           <button
             onClick={() => {
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -4,6 +4,7 @@ import {
   useCallback,
   useContext,
   useEffect,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -54,6 +55,7 @@ import {
   setSessionSnooze,
 } from "../lib/api";
 import { useServerDown, OFFLINE_TITLE } from "../lib/connectionState";
+import { clampMenuPosition } from "../lib/menuPosition";
 import { useHasDraftForSessions } from "../lib/cockpitDrafts";
 import { reportError } from "../lib/toastBus";
 import {
@@ -664,6 +666,22 @@ export const SessionRow = memo(function SessionRow({
   useEffect(() => {
     if (renaming) renameRef.current?.select();
   }, [renaming]);
+
+  useLayoutEffect(() => {
+    if (!contextMenu || !menuRef.current) return;
+    const rect = menuRef.current.getBoundingClientRect();
+    const next = clampMenuPosition({
+      x: contextMenu.x,
+      y: contextMenu.y,
+      menuWidth: rect.width,
+      menuHeight: rect.height,
+      viewportWidth: window.innerWidth,
+      viewportHeight: window.innerHeight,
+    });
+    if (next.x !== contextMenu.x || next.y !== contextMenu.y) {
+      setContextMenu(next);
+    }
+  }, [contextMenu]);
 
   useEffect(() => {
     if (!contextMenu) return;
@@ -1374,6 +1392,22 @@ const RepoGroupHeader = memo(function RepoGroupHeader({
   useEffect(() => {
     if (renaming) renameRef.current?.select();
   }, [renaming]);
+
+  useLayoutEffect(() => {
+    if (!contextMenu || !menuRef.current) return;
+    const rect = menuRef.current.getBoundingClientRect();
+    const next = clampMenuPosition({
+      x: contextMenu.x,
+      y: contextMenu.y,
+      menuWidth: rect.width,
+      menuHeight: rect.height,
+      viewportWidth: window.innerWidth,
+      viewportHeight: window.innerHeight,
+    });
+    if (next.x !== contextMenu.x || next.y !== contextMenu.y) {
+      setContextMenu(next);
+    }
+  }, [contextMenu]);
 
   useEffect(() => {
     if (!contextMenu) return;
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -57,6 +57,7 @@ import { useServerDown, OFFLINE_TITLE } from "../lib/connectionState";
 import { useHasDraftForSessions } from "../lib/cockpitDrafts";
 import { reportError } from "../lib/toastBus";
 import {
+  repoGroupHasLiveWorkspace,
   resolveEffectiveSnoozedUntil,
   snoozeTimestampCloseEnough,
   triageMenuShape,
@@ -1854,7 +1855,7 @@ export function WorkspaceSidebar({
             collisionDetection={closestCenter}
             onDragEnd={dragDisabled ? undefined : handleDragEnd}
           >
-            {filteredGroups.map((group) => {
+            {filteredGroups.filter(repoGroupHasLiveWorkspace).map((group) => {
               const showExpanded = q ? true : !group.collapsed;
               const hasActiveChild = group.workspaces.some(
                 (ws) => ws.id === displayedActiveId,
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -669,18 +669,30 @@ export const SessionRow = memo(function SessionRow({
 
   useLayoutEffect(() => {
     if (!contextMenu || !menuRef.current) return;
-    const rect = menuRef.current.getBoundingClientRect();
-    const next = clampMenuPosition({
-      x: contextMenu.x,
-      y: contextMenu.y,
-      menuWidth: rect.width,
-      menuHeight: rect.height,
-      viewportWidth: window.innerWidth,
-      viewportHeight: window.innerHeight,
-    });
-    if (next.x !== contextMenu.x || next.y !== contextMenu.y) {
-      setContextMenu(next);
-    }
+    const menu = menuRef.current;
+    const clamp = () => {
+      const rect = menu.getBoundingClientRect();
+      const next = clampMenuPosition({
+        x: contextMenu.x,
+        y: contextMenu.y,
+        menuWidth: rect.width,
+        menuHeight: rect.height,
+        viewportWidth: window.innerWidth,
+        viewportHeight: window.innerHeight,
+      });
+      if (next.x !== contextMenu.x || next.y !== contextMenu.y) {
+        setContextMenu(next);
+      }
+    };
+    clamp();
+    // Re-clamp when the menu's own size changes after mount: web fonts
+    // loading, icon images decoding, or any deferred layout work can
+    // grow the menu past what the first measurement saw, which would
+    // otherwise leak items off-screen at the bottom edge. See #1601.
+    if (typeof ResizeObserver === "undefined") return;
+    const ro = new ResizeObserver(clamp);
+    ro.observe(menu);
+    return () => ro.disconnect();
   }, [contextMenu]);
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


njbrake (approved):
Blind approve :)

PR #1606, opened 2026-05-29
feat(tui): swap home-screen search keybind from / to \

BTForIT:
## Summary

The home view binds bare `/` to "enter search mode". For users who frequently send slash-commands (`/goal`, `/think`, `/clear`, etc.) into a focused agent pane, the home view's `/` handler keeps swallowing the leading character when the home view still has focus — `/goal` lands as `Find: goal` and the command never reaches the agent.

This PR swaps the home search trigger from `/` to `\`. The change is mechanical and confined to one TUI surface: the home view's universal-key handler, the cursor prefix that styles the leading character in the search input, the footer entry, and the search-mode tests.

Settings search (separate handler in `src/tui/settings/input.rs`) is intentionally left on `/` to keep this PR small; happy to do the same swap there as a follow-up if desired.

## Changes

- `src/tui/home/input.rs:1890` — `KeyCode::Char('/')` → `KeyCode::Char('\\')`
- `src/tui/home/render.rs` — leading-char span (line 803), input cursor prefix (line 825), footer `mk_key` entry (line 2374), and the universal-key rationale comment (line 2265)
- `src/tui/home/tests.rs` — 13 search-mode test cases updated; `test_slash_enters_search_mode` renamed to `test_backslash_enters_search_mode`

## Test plan

- [x] `cargo build --release` clean
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -- -D warnings` clean (runs in pre-commit hook)
- [x] `cargo test --release --lib home::tests::test` — 137 passed, 0 failed
- [ ] Manual: in TUI, press `\` on home view → search activates; press `/` → no longer triggers search

## Notes

Convention-wise `/` is the vim/less/more search idiom, so this is a non-trivial divergence — happy to discuss whether you'd prefer:

1. Hard swap (this PR): `/` no longer triggers search.
2. Add `\` as alias: both keys open search, no breakage.
3. Config-gated: a `search_keybind` setting with `/` as default.

I picked option 1 because it's the smallest diff and matches the muscle-memory fix the user actually wants. Easy to repivot to option 2 if you'd rather not drop `/`.

## Summary

This PR reassigns the home screen search activation keybind from `/` to `\`. The change prevents the forward slash from being consumed when users type slash-prefixed commands (like `/goal`) into the agent pane, which previously interrupted their input by triggering search mode instead.

### What Changed

- **Input handling**: Home view now responds to `\` instead of `/` to enter search mode
- **UI rendering**: Search indicator and cursor prefix updated from `/` to `\` throughout the home screen
- **Tests**: 13 search-related tests updated to verify the new `\` keybind; one test function renamed for clarity
- **Settings search**: Intentionally left unchanged (remains on `/`)

### Benefits

- Users can now type slash-prefixed commands without unintended search mode activation
- Minimal, focused change with no impact on other keybindings or functionality
- All existing tests pass after updates

### Technical Details

**Files Modified:**
- `src/tui/home/input.rs`: Single-line change swapping `KeyCode::Char('/')` to `KeyCode::Char('\\')`
- `src/tui/home/render.rs`: Four-line changes updating the search indicator in cursor position helper, cursor character, footer shortcut group, and adjacent comment
- `src/tui/home/tests.rs`: Thirteen-line changes updating test assertions and renaming `test_slash_enters_search_mode` → `test_backslash_enters_search_mode`

**Validation:**
- `cargo build --release`, `cargo fmt --check`, and `cargo clippy` all pass
- 137 home view tests pass with 0 failures

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1606?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -1887,7 +1887,7 @@ impl HomeView {
                     self.open_tool_picker();
                 }
             }
-            KeyCode::Char('/') => {
+            KeyCode::Char('\\') => {
                 self.search_active = true;
                 self.search_query = Input::default();
             }
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -800,7 +800,7 @@ impl HomeView {
                 .unwrap_or_else(|| " ".to_string());
             let after: String = value.chars().skip(cursor_pos + 1).collect();
 
-            let mut spans = vec![Span::styled("/", text_style)];
+            let mut spans = vec![Span::styled("\\", text_style)];
             if !before.is_empty() {
                 spans.push(Span::styled(before, text_style));
             }
@@ -822,7 +822,7 @@ impl HomeView {
 
             frame.render_widget(Paragraph::new(Line::from(spans)), search_area);
             if !self.has_overlay_above_search() {
-                set_prefixed_input_cursor_position(frame, search_area, "/", &self.search_query);
+                set_prefixed_input_cursor_position(frame, search_area, "\\", &self.search_query);
             }
         }
     }
@@ -2262,7 +2262,7 @@ impl HomeView {
             ]
         };
         // Key-only entry for keys universal enough that a description would be
-        // noise (? for help, / for search). Saves footer width at iPhone-Mosh
+        // noise (? for help, \ for search). Saves footer width at iPhone-Mosh
         // sizes.
         let mk_key =
             |key: &str| -> Vec<Span<'static>> { vec![Span::styled(key.to_string(), key_style)] };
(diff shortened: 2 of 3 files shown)


njbrake:
Hello @BTForIT I do not want to change the keybind, but I am ok if you would like to add configuration options to let people configure what they want their keybind to be.

WHen you implement the configuration option make sure that you have validation code that makes sure whatever key the user is trying to configure is not already bound by another command. Thanks!

Also please be sure that the setting/config shows up in the TUI setting screen, it may make sense to have a new setting tab for "keybinds" or something like that, I bet we could follow whatever the design of IDEs like VSCode or Cursor follow for configuring keybindings.

njbrake (changes requested):
See other comment 🙏

PR #1615, opened 2026-05-29
fix(web): repeat-delete on mobile soft-keyboard Backspace hold

Seluj78:
**Note:** Claude handled the implementation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Description

On mobile soft keyboards (iOS Safari, Android GBoard / Samsung Keyboard), holding the on-screen Backspace key in either terminal deleted only one character instead of repeat-deleting. The only way to clear a line was to mash the key.

All terminal input flows through a single `term.onData` callback in `web/src/hooks/useTerminal.ts`. On desktop, OS-level keydown autorepeat fires repeated `keydown` events that xterm.js decodes one by one, so `onData` runs N times. On mobile, holding Backspace delivers autorepeat as a stream of `beforeinput` events (`inputType: "deleteContentBackward"`, with `keydown` surfacing as `Unidentified` / `keyCode 229`); xterm decodes only the first into a single `onData`, so the PTY received one DEL no matter how long the key was held.

The fix intercepts `beforeinput` on xterm's hidden textarea, gated to coarse-pointer devices without a fine pointer, and emits one DEL (`0x7f`) per `deleteContentBackward` tick. Calling `preventDefault()` blocks the textarea mutation so xterm's input-diff decoder never emits its own DEL for the same tick (no double-delete). Composition ticks (`isComposing`) are left to xterm's IME path. A hardware Backspace fires a recognized `keydown` that xterm `preventDefault`s, which per the UI Events spec suppresses the follow-on `beforeinput`, so the new path runs only for soft keyboards. This mirrors the mobile-Enter interception already in `Composer.tsx`. Both the agent terminal and the paired right-panel terminal share `useTerminal`, so both are fixed.

Closes #1450

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [x] On a mobile device (or Chrome DevTools iPhone preset with touch emulation), open the dashboard and select a session.
- [x] Tap the agent terminal, type `hello world`.
- [x] Tap and hold Backspace for ~2s; confirm the line clears (repeat-delete) instead of deleting a single character.
- [x] Repeat in the right-panel paired terminal; confirm the same repeat-delete behavior.
- [x] On desktop with a physical keyboard, hold Backspace; confirm repeat-delete is unchanged and a single tap deletes exactly one character (no double-delete).

## Checklist

- [x] New and existing tests pass
[...]

PR #1630, opened 2026-05-30
chore: merge cockpit-main batch into main

### LINKED ISSUE #1432: Mobile terminal: auto-resize on keyboard show/hide, retire the manual ViewportFullscreenFab
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._
>
> Filed as a design discussion, not a straight bug. The current behaviour is deliberate; the request is to reconsider the tradeoff.

## Ask

Remove `ViewportFullscreenFab` from the mobile terminal view. Make the terminal pane auto-resize as the soft keyboard opens and closes via the `visualViewport` API that `useMobileKeyboard` already monitors.

## Current behaviour

`web/src/components/TerminalView.tsx:239-248` documents the present design: the terminal pane is padded to the latched `reservedKeyboardHeight`, not the live keyboard height, so showing or hiding the keyboard does not change pane size and therefore does not send `SIGWINCH` to the PTY. `ViewportFullscreenFab` (`web/src/components/ViewportFullscreenFab.tsx:6-10` comment) is the only way to release that reservation and let the terminal claim the full viewport; each tap is an explicit accept of one `SIGWINCH` plus one Claude redraw.

## Why the current design exists

Quoted from `ViewportFullscreenFab.tsx:6-10`: *"each tap is the user explicitly accepting one SIGWINCH and one claude redraw."* Repeated keyboard-driven resizes would re-flow the TUI agent (Claude Code in particular) on every show/hide, scrolling the previous output into the scrollback and disrupting the active turn.

## Tradeoffs to discuss before implementing

1. **PTY tools that re-flow on `SIGWINCH` (Claude Code, opencode, codex):** auto-resize will cause visible redraws on every keyboard toggle. Acceptable if we debounce, or if we accept the redraw as the cost of better mobile UX.
2. **Cockpit mode has no PTY:** auto-resize there is free. The FAB is only mounted in `TerminalView`, not in `CockpitView`, so the rationale only bites for TUI sessions.
3. **`reservedKeyboardHeight` storage:** currently persisted to `localStorage` (`useMobileKeyboard.ts:5,69-72`). If we drop the latch, this becomes dead code.

## Proposed implementation

1. In `TerminalView.tsx:245-248`, replace `appliedKeyboardPadding` (derived from `reservedKeyboardHeight`) with the live `keyboardHeight` from `useMobileKeyboard`.
2. Remove `ViewportFullscreenFab` and its mount in `TerminalView.tsx:332-337`.
3. Optionally debounce the `setKeyboardHeight` update in `useMobileKeyboard.ts` by ~150ms so a quick tap-out / tap-in (autocorrect suggestions etc.) does not double-`SIGWINCH`.
4. Drop the `RESERVATION_STORAGE_KEY` localStorage entry and the latching code in `useMobileKeyboard.ts:120-129`.

## Open question for the maintainer

Are we OK with one `SIGWINCH` per keyboard toggle for Claude Code TUI sessions, in exchange for the simpler model? If not, the alternative is to keep the latch but auto-release it on an idle timer (e.g. keyboard has been closed for >2s ⇒ release reservation), which preserves the no-resize-during-typing invariant.

### LINKED ISSUE #1450: Mobile xterm.js: holding Backspace doesn't repeat-delete (both agent and paired terminals)
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

On mobile (iOS Safari, Android Chrome / GBoard / Samsung Keyboard), holding the soft-keyboard Backspace key in either xterm.js terminal produces exactly one delete instead of repeat-deleting. Lift-and-tap-again works each time, so the only way to clear a line on mobile is to mash the key.

Reporter notes this on the agent terminal and expects the paired terminal to behave the same. Code confirms: both terminals share the same input wiring (see below), so the bug applies to both.

## Affected surfaces

- **Agent terminal**: `web/src/components/TerminalView.tsx` (mounted via `useTerminal()` from `web/src/hooks/useTerminal.ts`).
- **Paired terminal (right panel)**: `web/src/components/RightPanel.tsx` (also via `useTerminal()`).

Both render `MobileTerminalToolbar` and use the same `term.onData()` hookup.

## Why it happens

xterm.js v6 (`web/package.json:27`) routes all keyboard input through a hidden textarea, decodes it internally, and surfaces cooked character data to a single `term.onData` callback. Our only input listener is at `web/src/hooks/useTerminal.ts:641-653`:

```ts
term.onData((data: string) => {
  if (ws.readyState !== WebSocket.OPEN) return;
  // ...virtual Ctrl handling...
  ws.send(new TextEncoder().encode(data));
});
```

That callback runs once per decoded keypress. On desktop, OS-level key autorepeat fires repeated `keydown` events at the textarea, xterm decodes each one, and `onData` runs N times. On mobile soft keyboards, holding Backspace delivers autorepeat as a stream of `beforeinput` events with `inputType: "deleteContentBackward"` and either no `keydown` at all or `keydown` with `key: "Unidentified"` / `keyCode: 229`. xterm's stock textarea proxy does not synthesise repeated `onData` calls from that `beforeinput` stream, so the PTY only receives one `\x7f`.

For comparison, the cockpit composer already does this dance for the on-screen Enter key, intercepting `beforeinput` with `inputType: "insertLineBreak"` / `"insertParagraph"` at `web/src/components/cockpit/Composer.tsx:481-501`. The terminal needs the analogous treatment for `deleteContentBackward` (and probably `insertText` autorepeat too, though Backspace is the loud one).

## Proposed fix sketch

Inside `useTerminal()`, after `term.open(...)`, attach a `beforeinput` listener to xterm's internal hidden textarea (`term.element.querySelector("textarea")`) gated to `(pointer: coarse)` so desktop autorepeat is untouched:

```ts
if (isCoarsePointer) {
  const ta = term.element?.querySelector("textarea");
  ta?.addEventListener("beforeinput", (e: InputEvent) => {
    if (e.inputType === "deleteContentBackward") {
      // Don't preventDefault; xterm still fires onData once.
      // Send N-1 extra DEL bytes per repeat tick to compensate.
      // Alternatively, swallow the native event entirely and send all
      // bytes from here.
    }
  });
}
```

Two viable approaches:

1. **Augment**: let xterm's first `onData("\x7f")` fire normally, then on subsequent `beforeinput` ticks (debounced ~50ms apart) send extra `\x7f` bytes directly to the WS.
2. **Replace**: `preventDefault()` the `beforeinput`, swallow the matching `onData`, and emit all DEL bytes from the listener. Cleaner but riskier (needs care to not eat desktop deletes if a phone with a physical keyboard ever hits this path).

I'd start with (1) behind `useIsCoarsePointer()`, write a Playwright test on `devices['iPhone 13']` that dispatches a sequence of `beforeinput` events with `inputType: "deleteContentBackward"` and asserts the WS receives the expected number of `\x7f` bytes, and iterate from there.

## Out of scope (for now)

- Soft-keyboard autorepeat of regular characters (`insertText` ticks). Same mechanism, but less painful day-to-day; address separately if users complain.
- Long-press popovers for diacritics on iOS. Different event shape, not relevant here.

## Repro

1. Mobile device or 390x844 iPhone preset in Chrome DevTools with touch emulation.
2. Open the agent terminal in any session.
3. Type "hello world".
4. Tap and hold Backspace for 2 seconds.
5. Expected: line clears (or close to it). Actual: one character deleted.
6. Repeat in the right-panel paired terminal: same behavior.

## Related

- The composer's mobile-Enter handling at `Composer.tsx:481-501` is the working template for this class of fix.
- xterm.js upstream issues around mobile soft-keyboard handling are long-running; this fix lives entirely on our side and doesn't need an upstream change.

### LINKED ISSUE #1451: Mobile sidebar: iOS link-preview callout fires on long-press before the rename/delete menu appears
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

On mobile Safari (iOS), long-pressing a session row in `WorkspaceSidebar` should open the app's rename / delete context menu after a 500ms hold. Instead, iOS pops its native link-preview action sheet (Copy Link / Open in Background / Share) first, because the row is rendered as an `<a href>` and iOS treats long-press on a link as the universal link-callout gesture. The user has to dismiss Safari's sheet to get back to the app, which then never gets a chance to fire its own menu. The result is a messy mobile experience where the documented "hold to rename or delete" path is unreachable.

We also need to make sure the drag-to-rearrange gesture (dnd-kit `TouchSensor` on the wrapper div) is unaffected by the fix.

## Evidence

- Session row element: `<a href={sessionPath}>` at `web/src/components/WorkspaceSidebar.tsx:563-602`. The link triggers Safari's native long-press callout.
- App's long-press handler: `handleTouchStart` wired at lines 591-594, with a 500ms timeout that opens the context menu (lines 496-510).
- Context menu (Rename / Delete / notification presets): `WorkspaceSidebar.tsx:695-750`.
- Existing CSS on the row (line 595): `select-none` (Tailwind ⇒ `user-select: none`). That suppresses text selection but does NOT suppress the iOS link callout. `-webkit-touch-callout: none` is nowhere in the row CSS.
- Drag-to-rearrange: `SortableSessionRow` wrapper at `WorkspaceSidebar.tsx:292-356` uses `@dnd-kit/sortable` with `TouchSensor` (activation `delay: 150, tolerance: 8`). Drag listeners spread onto the outer wrapper `<div>` at line 337, with `touchAction: "manipulation"` at line 322.

## Why drag still works after the fix

The three gestures are layered:

| Gesture          | Owner                              | Activation             |
| ---------------- | ---------------------------------- | ---------------------- |
| Drag-to-reorder  | dnd-kit `TouchSensor` on wrapper   | 150ms hold + 8px move  |
| Long-press menu  | Custom `handleTouchStart` on `<a>` | 500ms hold, no move    |
| iOS link callout | Safari, triggered by `<a href>`    | ~500ms hold on a link  |

dnd-kit captures the pointer at 150ms once 8px of movement happens, so a deliberate drag never reaches the 500ms threshold. A stationary hold sails past 150ms with no movement, so dnd-kit doesn't activate, and at 500ms the app's menu fires (today, the iOS callout fires alongside it and wins). Suppressing the iOS callout on the `<a>` only affects the "stationary hold" path; the dnd-kit drag path is on the wrapper `<div>` and never touched.

## Proposed fix

Add `-webkit-touch-callout: none` to the session row `<a>` element. Tailwind has no built-in utility for this; two viable options:

1. **Inline style on the `<a>`** at `WorkspaceSidebar.tsx:563-602`:
   ```tsx
   style={{ WebkitTouchCallout: "none" }}
   ```
2. **Arbitrary-value Tailwind class** in the existing `className` chain:
   ```
   [-webkit-touch-callout:none]
   ```

Either works; option 2 keeps the styling cohesive with the rest of the row. While we're here, also belt-and-braces `e.preventDefault()` inside `handleTouchStart` once the long-press timer fires (lines 496-510 already prevent default on `touchend`, but firing it on the long-press itself doesn't hurt).

## Test plan

- Mocked Playwright spec on `devices['iPhone 13']` that asserts the `<a>` carries the `-webkit-touch-callout: none` style.
- Manual: real iPhone or iPad in Safari (and PWA install) confirming long-press on a session row opens the app's Rename / Delete menu without Safari's link callout flashing first.
- Drag-to-rearrange regression: drag a session row up two positions; confirm reorder still works and the menu does not appear mid-drag.
- Desktop right-click: confirm `handleContextMenu` (line 482) still opens the menu (it should — the fix is iOS-only CSS).

## Out of scope

- The same callout suppression would also help on workspace / repo group rows. If those are also `<a href>`, the same one-line fix can be applied; track separately if it's a different surface.

## Repro

1. iPhone Safari (or PWA install), open the dashboard with at least one session in the sidebar.
2. Tap and hold a session row for ~1s.
3. Expected: app's Rename / Delete menu opens at the touch point.
4. Actual: Safari's "Copy Link / Share / Open in Background" sheet appears; the app's menu never shows.

### LINKED ISSUE #1454: Selecting a session in the sidebar should auto-focus the agent input (composer in cockpit, xterm.js textarea otherwise)
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

Clicking a session row in the left sidebar swaps the main view to that session's terminal but does not move focus into the agent input. The user has to make a second click on the composer textarea (cockpit) or the xterm.js terminal (non-cockpit) before they can type. That second click is small friction on desktop and meaningful on mobile, where it can pop the soft keyboard at the wrong time.

Expected: click a session → focus lands on whichever input is the canonical "type here" target for that session's mode.

## Evidence

- Select handler: `web/src/App.tsx:345-352`. `handleSelectSession` calls `navigate("/session/...")` and `focusKeyboardProxy()`. `focusKeyboardProxy` focuses the hidden global textarea used for keyboard-shortcut capture, not the agent input. Nothing here dispatches a terminal focus event or stashes a pending focus intent.
- Focus event infra already exists: `web/src/lib/terminalFocus.ts` defines `dispatchFocusTerminal("agent" | "paired")` and a pending-focus latch (`setPendingTerminalFocus` / `consumePendingTerminalFocus`) for when the target component hasn't mounted yet.
- Listeners on the terminal side: `web/src/components/TerminalView.tsx:165-167` registers a `FOCUS_TERMINAL_EVENT` listener and consults the pending latch at line 171.
- Composer in cockpit: `web/src/components/cockpit/Composer.tsx:538` has `autoFocus={!isMobile}` on the textarea, plus a focus claim in the `primerPrefill` effect at lines 290-310. Neither path is wired to session selection from the sidebar.

The select handler is missing the focus dispatch. The terminal-side machinery is already there waiting for it.

## Why it has to be a latch, not just a dispatch

On first open of a session, the terminal/composer mounts asynchronously: `useTerminal` waits for `ensureSession` to resolve, the cockpit composer mounts under `Suspense`, and on first load the WS connection itself can take a few retries before any character lands. If the select handler only dispatches `FOCUS_TERMINAL_EVENT`, the listener doesn't exist yet and the event is lost. The pending-focus latch already in `terminalFocus.ts` is built exactly for this: stash intent, consume on mount.

## Proposed fix

In `handleSelectSession` (`App.tsx:345-352`):

```ts
const handleSelectSession = useCallback((sessionId: string) => {
  const ws = workspaces.find((w) => w.sessions.some((s) => s.id === sessionId));
  if (!ws) return;
  navigate(`/session/${encodeURIComponent(sessionId)}`);

  // Pick the right target based on the session's mode.
  const session = ws.sessions.find((s) => s.id === sessionId);
  const target = session?.cockpit_mode ? "composer" : "agent";

  // Set the latch (consumed on mount if not yet present) AND dispatch
  // (handled immediately if already mounted, e.g. same-session re-select).
  setPendingTerminalFocus(target);
  dispatchFocusTerminal(target);

  if (window.innerWidth < 768) setSidebarOpen(false);
}, [navigate, workspaces]);
```

This requires:

1. Adding `"composer"` as a focus target. Today `terminalFocus.ts` only supports `"agent" | "paired"`. Either extend the union and wire a listener inside `Composer.tsx` (analogous to `TerminalView`), or add a separate `setPendingComposerFocus` primitive. The first is symmetric and probably cleaner.
2. `Composer.tsx` consumes the pending-composer latch in a mount effect (similar to `TerminalView.tsx:170-172`) and calls `taRef.current?.focus()`. The existing `autoFocus={!isMobile}` should stay for the "session was already active and user lands here directly" case.
3. Apply the same in `handleSelectWorkspace` (`App.tsx:354-371`) when it picks a session via `navigate`, since selecting a workspace from the picker should also land focus.

## Mobile caveat

On mobile, auto-focusing a textarea pops the soft keyboard. Today the agent terminal explicitly avoids this (`TerminalView.tsx:178-181` comment from #1178: "Auto-keyboard-open on initial connect removed; the KeyboardFab is always visible and lets the user open the keyboard explicitly. Popping the soft keyboard on every session open is the wrong default for navigation, especially read-only check-ins."). The same constraint applies here: on `(pointer: coarse)`, the select handler should NOT focus the textarea, since that would pop the keyboard on every session swap. Gate the dispatch + latch behind `!useIsCoarsePointer()`.

This means the fix is **desktop-only**. On mobile, a session click still leaves focus where it was (no keyboard pop). That matches existing intent.

## Test plan

- Desktop Vitest + RTL test asserting that `handleSelectSession` (in cockpit mode) results in the composer textarea being focused after the next tick.
- Same for non-cockpit terminal sessions (FOCUS_TERMINAL_EVENT path).
- Mobile mocked Playwright spec on `devices['iPhone 13']`: select a session, assert the soft keyboard does NOT open and document.activeElement is not the textarea.
- Coverage matrix: add an entry under the sidebar / sessions surface per the mandate in `CLAUDE.md`.

## Out of scope

- Focus restoration on returning from settings / projects / diff full-screen. That should also land somewhere reasonable but is a separate flow.

### LINKED ISSUE #1456: Remove the top-bar breadcrumb (aoe / repo / session); it duplicates the sidebar's active-row highlight
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Ask

Remove the breadcrumb fragment in the top bar that renders `aoe / $REPO / $SESSION_NAME`. The active workspace and session are already marked in the left sidebar (active row gets `bg-surface-850` and a brand-coloured left border at `WorkspaceSidebar.tsx`), so the top-bar copy is redundant. The `aoe` brand pill itself stays as the dashboard home link.

## Where it lives

`web/src/components/TopBar.tsx:89-106`:

```tsx
{/* Breadcrumb (hidden on mobile). Suppress the workspace crumb when it
    equals the repo name to avoid "/ foo / foo" duplication. */}
{(repoName || activeWorkspace) && (
  <div className="hidden sm:flex items-center gap-1.5 min-w-0 text-xs font-mono">
    <span className="text-text-dim">/</span>
    {repoName && (
      <span className="text-text-muted truncate max-w-[140px]">{repoName}</span>
    )}
    {activeWorkspace && activeWorkspace.displayName !== repoName && (
      <>
        <span className="text-text-dim">/</span>
        <span className="text-accent-600 truncate max-w-[200px]">
          {activeWorkspace.displayName}
        </span>
      </>
    )}
  </div>
)}
```

`repoName` is derived at line 44 from `activeWorkspace?.projectPath`. If nothing else consumes it after the breadcrumb is gone, drop that derivation too. Confirm `activeWorkspace` is still used elsewhere in the file before pruning the prop.

## Edge cases worth thinking about before deleting

1. **Sidebar collapsed.** If a user can collapse the left sidebar (drawer closed on mobile, hidden on narrow desktop?), there's no "what session am I in" hint anywhere else in the chrome. Check whether the current desktop layout ever leaves the sidebar invisible; if so, either keep a single-crumb fallback gated to that state or accept that the active-row info is sidebar-only and users open the sidebar to see it.
2. **Dashboard / projects views.** When `activeSession` is null and the user is on the dashboard or projects screen, the breadcrumb is already mostly empty. No behavior change here.
3. **Page title.** Browser tab `<title>` is independent of the top bar; if it currently includes the session name, it should keep doing so. If it doesn't, that's a separate decision.
4. **Palette and command surfaces.** No consumers there should care, but spot-check that `repoName` isn't read by anything outside `TopBar.tsx` (`grep -rn "repoName" web/src/`).

## Proposed implementation

1. Delete lines 89-106 in `TopBar.tsx`.
2. Drop the `repoName` derivation at line 44 if it has no remaining readers.
3. Drop the `activeWorkspace` prop from the TopBar's prop list if it has no remaining readers; remove from the call site in `App.tsx` accordingly.
4. Update any TopBar tests under `web/src/components/__tests__/` that assert on the breadcrumb.
5. Sanity check on desktop: `(pointer: fine)` at 1440px wide, the center palette pill and right-zone affordances should stay symmetric; the deletion frees up left-zone space and the palette stays centered via the existing `flex-1 flex justify-center` on the middle column.

## Out of scope

- Any redesign of the active-row treatment in the sidebar. Today's bg + left-border treatment is sufficient; if it isn't, that's a separate design issue.
- Adding a different "you are here" indicator anywhere else. The whole point is that the sidebar already does the job.

### LINKED ISSUE #1467: Cockpit: failed tool cards force-expanded, chevron toggle no-op until status leaves "err"

## Problem

Failed tool calls in the cockpit web UI cannot be folded. Once a tool's `status === "err"`, the card stays expanded for the lifetime of the transcript. The chevron toggle in the header still rotates (clicking flips internal `open` state), but the body never collapses because the render is gated on `open || status === "err"`.

On a long transcript with many failed Edits / Reads / Bash / MCP calls, every error remains fully unfurled, including the rose error block AND the "Show attempted action" details section. The transcript grows quickly and the user has no way to compact the noise once they have read the failure.

Auto-expand on failure is a reasonable default (the user wants to see what went wrong immediately) but it should not be a permanent lock. The user wants to be able to dismiss / fold a failed card after reading it.

### Forced-expanded pattern on every per-kind card

`web/src/components/cockpit/ToolCards.tsx` uses the same `expanded={open || status === "err"}` shape on 12 call sites (every card type except group / subagent / actions which use `expanded={open}`):

- Read (`ToolCards.tsx:680`)
- Write/Edit (`ToolCards.tsx:744`)
- Search / Glob / Grep (`ToolCards.tsx:782`, `ToolCards.tsx:839`)
- Fetch (`ToolCards.tsx:893`)
- MCP (`ToolCards.tsx:1021`)
- Skill (`ToolCards.tsx:1106`)
- Other generic per-kind cards (`ToolCards.tsx:1420`, `1505`, `1604`, `1815`, `1863`)

### Header still wires a toggle

`CardChrome` at `ToolCards.tsx:320-368` renders the chevron + click handler whenever `onToggle` is set. The toggle flips local `open` state, but the rendered `expanded` prop ORs in `status === "err"`, so the body stays open. From the user's seat, clicking the chevron does nothing on a failed card.

### The body itself is already error-aware

`web/src/components/cockpit/ToolErrorBody.tsx:33-63` already structures the failed body into a rose error block plus a collapsed `<details>` ("Show attempted action") containing the per-kind body. The pieces for a foldable failed card exist; the outer chrome is the only thing forcing it open.

### History

`#1090` (closed) wired the rich error rendering via `ToolErrorBody`. The auto-expand-on-err appears to have ridden along as the implicit default in that PR. `#1102` (closed) handled group-level err rollup. Neither issue explicitly justified locking the err card open; it is the default that everyone inherited.

### A. Drop the auto-lock, keep auto-open on first render

Change every `expanded={open || status === "err"}` to `expanded={open}` and initialize the per-card local state to `open = (status === "err")` on first mount.

Rough shape:

```tsx
// per-kind card
const status = statusFor(result);
const [open, setOpen] = useState(status === "err");
```

User-visible effect: failed cards still open by default (current behavior), but the toggle now works to fold them. Clicking the chevron collapses the rose block and the details section. Clicking again unfolds them.

Twelve call sites; near-mechanical change. No settings parity, no migration, no docs change beyond a one-line note in `docs/cockpit.md` if it currently advertises the auto-expand.

### B. Settings toggle for users who prefer never-auto-expand

Add `cockpit.auto_expand_failed_tool_calls: bool` (default true) so users who prefer the current behavior keep it.

Trigger when toggled false: failed cards render collapsed by default. The header dot + "failed" badge still signal the failure; the user opens it manually.

Wiring (the 7-step parity from `AGENTS.md` "Settings & Configuration", same shape `cockpit.show_tool_durations` at `src/session/config.rs:243-244` uses):

1. `src/session/config.rs` `CockpitConfig`: `#[serde(default = "default_true")] pub auto_expand_failed_tool_calls: bool` (init in `Default` at `src/session/config.rs:362-377`).
2. `src/session/profile_config.rs:88`: add `pub auto_expand_failed_tool_calls: Option<bool>` on `CockpitConfigOverride` + merge arm around `src/session/profile_config.rs:589-591`.
3. `src/tui/settings/fields.rs:143`: add `FieldKey::CockpitAutoExpandFailedToolCalls`.
4. `src/tui/settings/fields.rs:558-562`: add a `resolve_value` block.
5. `src/tui/settings/fields.rs:652-662`: add a `SettingField` entry, category `Cockpit`.
6. `src/tui/settings/fields.rs:2466-2468` + `:2949-2953`: `apply_field_to_global` + `apply_field_to_profile` arms.
7. `src/tui/settings/input.rs:906-910`: `clear_profile_override` arm.

Server side: `src/server/api/system.rs:543` adds `pub cockpit_auto_expand_failed_tool_calls: bool` on `ServerAbout`, resolved at `:591` and surfaced at `:605`.

Web side: `web/src/lib/cockpitPrefs.tsx:23` adds `autoExpandFailedToolCalls: boolean`, default `true` in `DEFAULT_PREFS`. Per-card init becomes:

```tsx
const { autoExpandFailedToolCalls } = useCockpitPrefs();
const [open, setOpen] = useState(
  autoExpandFailedToolCalls && status === "err",
);
```

Docs: `docs/cockpit.md` adds a sentence under the existing tool-card section describing the setting.

### Recommendation

Land (A) alone unless the team wants the toggle. (A) restores user agency at zero config-surface cost. (B) is the right shape only if anyone actually wants the never-auto-expand behavior; nothing in the issue corpus suggests that, but it pairs cleanly with `show_tool_durations` as a per-card UX knob.

## User stories + Playwright / Vitest coverage

Per the user-story spec mandate (`AGENTS.md` Web Dashboard Playwright Tests + `#1383`), every PR landing this work must update `web/tests/coverage-matrix.json` and add or modify specs that cover the behavior. Three stories below cover the user-visible surface.

### Story 1: user folds a failed tool card after reading it

**Surface:** cockpit transcript, failed tool card (any per-kind: Read, Edit, Bash, Search, etc.).

**Steps:**

1. Open a cockpit session.
2. Trigger a tool call that fails (e.g. Edit on an unread file, Bash with non-zero exit). The fake-ACP agent at `web/tests/helpers/fakeAcpAgent.mjs` can synthesize the `tool_error` row.
3. Assert the card renders expanded by default (rose error block visible).
4. Click the header chevron.
5. Assert the rose error block + "Show attempted action" details collapse. The header dot + "failed" badge remain.
6. Click the chevron again.
7. Assert the body expands again.

**Spec home:** `web/tests/live/cockpit-stories/fold-failed-tool-card.spec.ts`.

**Matrix entry:** `web/tests/coverage-matrix.json`, surface `story.cockpit.fold-failed-tool-card`, kind `live-playwright`, component `web/src/components/cockpit/ToolCards.tsx`.

### Story 2: failed card opens by default on first render

**Surface:** cockpit transcript, freshly arrived `tool_error` row.

**Steps:**

1. Open a cockpit session that already has a failed tool call in its replayed activity (use the `aoeServe` harness + a fake-ACP scripted scenario that emits one failure before the WS connects).
2. Assert the failed card renders with the rose error block visible without any user click.

**Rationale:** guards against an over-eager (A) implementation that initialized `open=false` and silently regressed the auto-open default. Catches the regression at the only place the user notices it (first paint).

**Spec home:** can extend Story 1's spec OR live next to it as `failed-tool-card-auto-open.spec.ts`. Likely same file with two `test()` cases.

### Story 3 (only if proposal B lands): toggle respects the setting

**Surface:** Settings → Cockpit → Auto-expand failed tool calls.

**Steps:**

1. Open Settings, toggle `cockpit.auto_expand_failed_tool_calls` to false, save.
2. Open a cockpit session, trigger a failed tool call.
3. Assert the failed card renders **collapsed** by default (only the header dot + "failed" badge visible).
4. Click the chevron; assert the body expands.
5. Toggle the setting back to true; new failed cards render expanded again.

**Spec home:** `web/tests/live/cockpit-stories/cockpit-auto-expand-failed-toggle.spec.ts`.

**Matrix entry:** surface `story.settings.cockpit-auto-expand-failed`, component `web/src/components/SettingsView.tsx` + `web/src/components/cockpit/ToolCards.tsx`.

### Vitest pairing

In addition to the live specs, add a Vitest RTL test next to `web/src/components/cockpit/__tests__/` (or wherever the ToolCards Vitest sits — the existing one from `#1335` is the canonical model). Assertions:

- A card mounted with `status === "err"` renders expanded on first paint.
- A click on the chevron flips `expanded` off.
- A card mounted with `status === "ok"` renders collapsed.

Vitest covers the render-branch logic without the full live harness; Playwright covers the click-through user flow. Per `AGENTS.md`: "Pure styling or copy-only changes may add a `kind: deferred` entry"; this is behavior, so a real spec is required.

## Why now

- High noise on long transcripts with multiple failed calls (an agent fixing a stubborn Edit, for example, can produce 5+ failed cards in a row).
- Pure render-side change; no backend, no event store, no protocol.
- Twelve mechanical call sites; one logical commit for (A).
- Pairs with the broader "tool-card UX as informative as the CLI" theme tracked in #1090 / #1102 / #1041 / #1062.

## Open questions

- For (A): should a card that goes from running to err mid-stream auto-open if the user had already manually collapsed it (set `open` back to true on transition to err), or stay where the user left it? Lean: respect the user's last toggle once set. Implementation: track a `userToggled` ref and only auto-open the first time status transitions to err.
- For (B): default true (preserves current behavior) or false (forces user opt-in to auto-expand)? Lean true; preserves the current default and lets opt-out land cleanly.
- Group cards (`expanded={open}` already, no force-expand) and Subagent cards behave correctly today. Confirm we are not introducing a divergence between top-level err cards and group/subagent children once (A) lands.
- Persistence across reload: should the fold state survive a tab refresh? Today nothing persists per-card state. Probably out of scope; pairs with the broader "remember scroll position + open cards" follow-up. Note here, do not block (A).

## Out of scope

- Auto-collapse on successful retry. The agent typically retries a failed Edit after a Read; conceivably the cockpit could fold the prior error once a later success lands. Separate UX call; speculative.
- Bulk "fold all errors" affordance in the transcript header. Easy follow-up once (A) lands; not v1.
- Reworking `ToolErrorBody.tsx` itself. Its internal layout (rose block + details) is fine; the issue is purely the outer chrome's `expanded` calculation.

## Cross-references

- #1090 (closed) — added `ToolErrorBody` and the rich error rendering. Source of the implicit auto-expand-on-err default.
- #1102 (closed) — fixed group err rollup. Tangential; same component.
- `web/src/components/cockpit/ToolCards.tsx:680, 744, 782, 839, 893, 1021, 1106, 1420, 1505, 1604, 1815, 1863` — the 12 force-expand call sites.
- `web/src/components/cockpit/ToolErrorBody.tsx` — the body wrapper; no change needed.
- `src/session/config.rs:232-244` — `show_tool_durations` is the model for (B)'s settings parity.

### LINKED ISSUE #1468: Cockpit: consecutive TodoWrite cards spam the timeline, fold into a dedicated TodoGroupCard (iterate on #1064)

## Problem

Claude fires `TodoWrite` many times per substantive turn: add a few items, mark one in-progress, mark it done, add more, mark the next one in-progress, etc. Each call renders as its own `TodoUpdateCard` in the cockpit timeline. A turn with 6 todo updates produces 6 stacked cards, all showing nearly the same list with one row's status nudged forward.

The timeline becomes a wall of near-duplicate todo cards that drowns out the actual tool work (Reads, Edits, Bashes) the user is trying to watch. The latest todo state, which is the only one whose `pending`/`in_progress`/`completed` mix is current, is buried under five superseded snapshots.

`TodoWrite` is high-signal but the per-call rendering treats every snapshot as equally important, when in practice only the latest is the live state and the prior ones are history.

### Per-update synthesis

`src/cockpit/acp_client.rs:2628-2697` turns each `SessionUpdate::Plan` (the channel claude-agent-acp routes `TodoWrite` through) into a synthetic `ToolCallStarted` + `ToolCallCompleted` pair with a fresh `todo-{ts_ms}-{seq}` id:

```rust
let ts_ms = chrono::Utc::now().timestamp_millis();
let seq = SYNTHETIC_TOOL_SEQ.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
let plan_id = format!("plan-{ts_ms}-{seq}");
let tool_id = format!("todo-{ts_ms}-{seq}");
...
vec![
    Event::ToolCallStarted { tool_call: ToolCall { id: tool_id.clone(), name: "TodoWrite".to_string(), ... } },
    Event::PlanUpdated { plan: Plan { plan_id, version: 1, steps } },
    Event::ToolCallCompleted { tool_call_id: tool_id, ... },
]
```

Every plan update is a fresh tool call from the cockpit timeline's perspective. No deduplication, no replacement.

### Generic action grouping exists, but excludes TodoWrite

`web/src/components/cockpit/CockpitRuntime.tsx:576-636` already folds runs of ≥3 consecutive tool-call parts into one `ToolGroupCard` so silent investigation (Read, Read, Grep, Read) collapses to a single collapsible block. The fold is gated:

```tsx
// TodoWrite calls aren't silent tool work; they're status
// updates the user wants to see one-by-one (#1064). Detect them
// via the `_aoe_title` echo we stash in argsText (the adapter
// names them "Update TODOs: …") and exempt the group entirely
// when any child looks like a TodoWrite. Cheap: argsText is
// already parsed JSON, just sniff the prefix.
const hasTodoWrite = run.some(
  (p) => p.type === "tool-call" && isTodoWriteArgsText(p.argsText),
);
if (hasTodoWrite) {
  for (const p of run) out.push(p);
  run = [];
  return;
}
```

That exemption was the deliberate v1 decision in #1064: "every TodoWrite call gets a card (potentially many per turn, even when the change is tiny), or fold consecutive TodoWrites into one 'current state' card? The first is more honest (you see each update); the second is cleaner. Probably do per-call for v1, then iterate."

This issue is the iterate step.

### TodoUpdateCard rendering

`web/src/components/cockpit/ToolCards.tsx:992-1099` already auto-collapses long todo lists (`useState(todos.length <= 5)`) and shows the counts breakdown (`3 active · 1 done`). The per-card UI is fine; the problem is the count of cards, not the card itself.

## Root cause

Decision in `CockpitRuntime.tsx:583-597` to exempt TodoWrite from the existing tool-grouping pass. The synthesis path in `acp_client.rs:2628-2697` is doing the right thing (every adapter update becomes a tool call); the timeline rendering is the layer that should fold consecutive ones now that we have real-world evidence of the spam.

### A. Fold consecutive TodoWrite calls into a dedicated `TodoGroupCard`

Drop the `hasTodoWrite` short-circuit in `collapseToolRuns` (`CockpitRuntime.tsx:590-597`) but route runs that are *all-TodoWrite* into a new synthetic part with a distinct toolName (e.g. `_aoe_todo_group`) rather than the generic `_aoe_tool_group`. The renderer in `CockpitView.tsx` dispatches the new name to `TodoGroupCard`.

`TodoGroupCard` shape:

- **Header**: "Todos updated N times" plus the current counts chip from the latest list (`3 active · 1 done`).
- **Collapsed body** (default): the full latest todo list, rendered with the same status glyphs as `TodoUpdateCard`. This is what the user actually cares about; the older snapshots are noise.
- **Expanded body** (click to expand): each individual TodoWrite as its own `TodoUpdateCard`, in order, so the history is still inspectable.

Mixed runs (TodoWrite + Read + Edit) keep the existing exemption so a single status update sandwiched between real work doesn't get hidden inside a generic tool group. The new fold only fires when every child in the run is a TodoWrite.

Touch points:
- `web/src/components/cockpit/CockpitRuntime.tsx` — split the all-TodoWrite case off from the generic group path; emit a new synthetic part type.
- `web/src/components/cockpit/CockpitView.tsx:656` — dispatch `_aoe_todo_group` → `TodoGroupCard`.
- `web/src/components/cockpit/ToolCards.tsx` — new `TodoGroupCard` (factored alongside the existing `TodoUpdateCard`, reusing `TODO_GLYPH` / `TODO_CLASS`).

### B. Diff highlighting inside the expanded view (optional follow-up)

Once consecutive updates live in one card, comparing successive snapshots becomes cheap. Mark items that flipped status between snapshots (`pending → in_progress`, `in_progress → completed`) with a small badge in the expanded list. Defer past v1; A is the core fix.

### C. Settings toggle (optional)

A "fold consecutive todo updates" toggle in cockpit display settings for users who preferred the per-call layout. Skip unless someone asks; default to folded.

## Why now

- TodoWrite spam is the most visible UX regression introduced by #1064. The dedicated card was the right call; the per-call layout was the wrong default.
- The fold infrastructure already exists in `collapseToolRuns`; this issue is mostly a one-condition split and a new card component.
- Pairs cleanly with #1057's grouping work, completing the original tradeoff #1064 deferred.

- **Golden path: consecutive TodoWrite cards fold to one card.** As a cockpit user, when the agent fires three or more `TodoWrite` calls back-to-back with no other tool calls between them, then I see one collapsible todo card in the timeline instead of three or more.
  - Suggested test layer: Vitest + RTL against `CockpitRuntime` + `CockpitView`, with a fixture of 4 synthetic `TodoWrite` tool-call parts. Optionally a live Playwright spec under `web/tests/live/` exercising a real cockpit session.

- **Final state still readable without expanding.** As a cockpit user, when consecutive TodoWrite calls fold into one card, then the collapsed view shows the *latest* todo list (the most recent snapshot) so I do not have to expand to see what the agent is currently working on.
  - Suggested test layer: Vitest + RTL on `TodoGroupCard`, asserting the rendered rows match the last fixture snapshot and the counts chip matches the latest list.

- **History still accessible on expand.** As a cockpit user, when I expand the folded todo card, then I see each individual TodoWrite call as its own row/card in order, so I can audit how the plan evolved during the turn.
  - Suggested test layer: Vitest + RTL on `TodoGroupCard`, click expand, assert N child snapshots render in original order.

## Open questions

- Threshold for folding: reuse `TOOL_GROUP_MIN_RUN = 3` (current generic threshold) or fold from 2? Probably keep at 3 to match existing tool-group behavior; a single revision (2 in a row) is still useful to see inline.
- Mixed runs: confirm the rule above (only fold when *all* children are TodoWrite). The alternative is to extract TodoWrites out of a mixed run before grouping the rest, but that reorders the timeline and breaks the "what did the agent do between sentences" model.
- Boundary on assistant text: should agent text between two TodoWrite calls still break the run (matching the existing generic grouping behavior)? Probably yes for consistency, but worth confirming with a real session log.

## Out of scope

- Editing todos from the cockpit UI.
- Persisting todos across sessions / surfacing them in the project dashboard.
- Changing the synthesis path in `acp_client.rs`. Folding is a UI concern; the backend's per-update emission stays as is.

## Cross-references

- #1064 — original `TodoUpdateCard` work; explicitly deferred this fold decision to a follow-up.
- #1057 — generic tool-grouping work that introduced `collapseToolRuns`.

### LINKED ISSUE #1469: Cockpit: long URLs and pasted PW failures overflow chat bubble, viewport gains horizontal scrollbar

## Problem

Some content pasted into the cockpit chat (long GitHub Actions run links, Playwright failure transcripts with long file paths and `─────` rule lines, anything with unbreakable tokens wider than the bubble) makes the message extend past its container and add a horizontal scrollbar to the chat viewport. The user has to scroll sideways to read the message, and the rest of the conversation visually shifts too because the viewport itself gains the scrollbar.

Reproduced in cockpit session `f437dfe0a0b341c6`, prompt at `seq=41`: a paste of `npx playwright test` output. The transcript contains lines like:

```
    > 79 |     await expect.poll(() => focusedKind(page)).toBe("paired");
         |                                                ^
        at /Users/seluj78/aoe/agent-of-empires-worktrees/fix-flaky-pw-tests/web/tests/terminal-focus-shortcut.spec.ts:79:48
    attachment #1: screenshot (image/png) ──────────────────────────────────────────────────────────
```

The long file path and the `─` rule line are unbreakable inline tokens. The user bubble's width cap can't shrink them, so they push the bubble past its `max-w-[80%]` and the scrollbar appears.

Same shape applies to long URLs (e.g. a pasted GitHub Actions run link, seq `610` in the same session: `https://github.com/njbrake/agent-of-empires/actions/runs/26342421371/job/77546632641`). On a narrow viewport the autolinked anchor is wider than the bubble; same overflow result.

### Bubble layout

User prompts render through `web/src/components/cockpit/CockpitView.tsx:432-446`:

```tsx
function UserText({ text }: { text: string }) {
  const payload = parseDiffCommentsSentinel(text);
  if (payload) {
    return <DiffCommentsUserCard payload={payload} />;
  }
  return (
    <div className="max-w-[80%] min-w-0 rounded-2xl rounded-br-sm border border-surface-700 bg-surface-800/70 px-3 py-1.5 text-sm">
      <Markdown text={text} smooth={false} />
    </div>
  );
}
```

`min-w-0` lets the bubble shrink in a flex parent; `max-w-[80%]` caps it. Neither breaks long inline tokens, which is what causes the overflow.

Assistant text uses `AssistantMessage` (`CockpitView.tsx:448-462`) with no width cap and a `w-full` container, so the same content from the agent side also overflows the viewport when it contains an unbreakable token.

### Markdown styles missing wrap

`.cockpit-markdown` in `web/src/index.css:176-275` styles paragraphs, headings, lists, blockquotes, anchors, tables, code blocks, and `<hr>`. It does **not** set `overflow-wrap`, `word-break`, or `min-width: 0` on the markdown content. Default `overflow-wrap: normal` does not break long unhyphenated tokens (URLs, paths, long bash output, the `─` rule line from Playwright's reporter).

`.cockpit-markdown a` (`index.css:212-216`) styles the anchor color but has no wrap rule, so an autolinked URL inherits the default.

Fenced code blocks already handle this correctly: `ShikiSyntaxHighlighter` and the fallback `<pre>` both wrap in `overflow-x-auto` (`Markdown.tsx:158, 164`). Scroll lives inside the code block, which is the desired behavior for code.

### Viewport scrollbar

The chat viewport is `web/src/components/cockpit/CockpitView.tsx:311-317`:

```tsx
<ThreadPrimitive.Root className="flex flex-1 flex-col min-h-0">
  <ThreadPrimitive.Viewport
    autoScroll
    ref={viewportRef}
    className="flex-1 overflow-y-auto"
  >
    <div className="mx-auto max-w-3xl xl:max-w-4xl 2xl:max-w-5xl px-4 py-6">
```

`overflow-y-auto` on the viewport with no explicit `overflow-x` follows the CSS rule that the unspecified axis resolves to `auto` when the other is `auto` (effectively visible becomes auto). So when a child bubble overflows horizontally, the viewport itself gains the scrollbar, which is what the user observes.

## Root cause

`.cockpit-markdown` (`web/src/index.css:176-275`) lacks `overflow-wrap: anywhere` (or equivalent `word-break: break-word`) on paragraphs, list items, and anchor children. Long unbreakable inline tokens in plain text content extend past the bubble's `max-w-[80%]`. The viewport's `overflow-y-auto` resolves `overflow-x` to `auto` too, so a horizontal scrollbar appears at the viewport level.

Three contributing layers:

1. **Markdown styles** (`index.css:176-275`): no wrap rule on paragraphs / inline content.
2. **User bubble** (`CockpitView.tsx:442`): width cap exists but no `min-w-0` enforcement on the markdown child itself; the cap can't break a long inline token.
3. **Viewport** (`CockpitView.tsx:315`): unbounded `overflow-x` lets a single misbehaving child grow the scroll area for the whole transcript.

### A. Wrap long tokens in markdown content

Add `overflow-wrap: anywhere` to `.cockpit-markdown` paragraphs, list items, blockquote bodies, and anchor children. One CSS rule covers every prose surface:

```css
.cockpit-markdown :where(p, li, blockquote, a) {
  overflow-wrap: anywhere;
}
```

`overflow-wrap: anywhere` (vs `break-word`) also contributes to the intrinsic min content width, so the bubble correctly reports a small minimum width to its flex parent and the `min-w-0 max-w-[80%]` cap takes effect.

Touch points:
- `web/src/index.css` — one rule in the `.cockpit-markdown` block.

### B. Belt-and-suspenders viewport clamp

Add `overflow-x-hidden` to the chat viewport so any future markdown surface that forgets to wrap can't make the whole transcript shift sideways. Combined with (A), the user never sees a viewport-level horizontal scrollbar; code blocks still scroll inside themselves via their own `overflow-x-auto` containers.

```tsx
<ThreadPrimitive.Viewport
  autoScroll
  ref={viewportRef}
  className="flex-1 overflow-x-hidden overflow-y-auto"
>
```

Touch points:
- `web/src/components/cockpit/CockpitView.tsx:315`.

### C. Optional: anchor break for very long URLs

A long URL with `overflow-wrap: anywhere` wraps at any character, which can render mid-domain. For the autolink case specifically, `word-break: break-all` on `.cockpit-markdown a` produces nicer break points at slash boundaries (browsers prioritize them). Cosmetic only; defer past v1.

## Why now

- Pasting Playwright failures and CI links is normal workflow during the project's current PW-stabilization push; the bug bites on the same content the user is debugging.
- One CSS rule + one className tweak. Smallest possible fix.
- No backend, no schema, no migration. Pure web layer.

- **Long URL in agent text wraps inside the bubble, no horizontal scrollbar.** As a cockpit user, when the agent sends a message containing a long autolinked URL (e.g. a GitHub Actions run link 80+ chars), then the URL wraps within the bubble width and the chat viewport shows no horizontal scrollbar.
  - Suggested test layer: Vitest + RTL on `Markdown` rendering inside a fixed-width container, asserting `scrollWidth <= clientWidth` on the viewport mock. Or a mocked Playwright spec under `web/tests/` with a fixed viewport width that loads a thread containing the long URL.

- **Pasted Playwright failure with long file paths and rule lines wraps inside the user bubble.** As a cockpit user, when I paste multi-line Playwright `--reporter=list` output (containing absolute file paths and `─────` rule lines) into the composer and send, then the resulting user message wraps within the bubble and the chat viewport shows no horizontal scrollbar.
  - Suggested test layer: Vitest + RTL with a fixture containing a real PW failure snippet, rendered inside a constrained-width container; assert `scrollWidth <= clientWidth`. Optionally repeated under live Playwright.

- **Fenced code blocks still scroll horizontally internally.** As a cockpit user, when a message contains a fenced code block with a long line, then the code block itself scrolls horizontally without making the chat viewport scroll horizontally.
  - Suggested test layer: Vitest + RTL on `Markdown` rendering a fenced block with a 200-char line; assert the code container has `overflow-x: auto` and the viewport above it does not gain a scrollbar.

## Open questions

- Use `overflow-wrap: anywhere` or `word-break: break-word`? `anywhere` is preferred because it contributes to min-content width (the bubble can actually shrink); `break-word` does not. Confirm before landing.
- Should `(B)` (viewport `overflow-x-hidden`) also apply to the tool-call timeline area, or only the chat viewport? Tool cards already manage their own internal scroll; probably only the chat viewport, but worth a quick audit.
- Tables (`.cockpit-table-wrap`, `index.css:229-235`) already have `max-height` and scroll containment. Confirm the wrap rule from (A) doesn't break table layout (the rule targets `p, li, blockquote, a` so table cells are excluded).

## Out of scope

- Restyling code blocks. Their `overflow-x-auto` is correct.
- Composer input wrapping. The composer is a `<textarea>` and already wraps via `wrap="soft"` / browser default.
- TUI overflow behavior. This issue is web-only.

## Cross-references

- #1108 — original "user prompts should render through Markdown" change; introduced the `UserText` bubble that this issue tunes.
- #1430 — separate mobile cockpit layout issues; touches different surfaces but adjacent in the "narrow viewport breakage" family.

### LINKED ISSUE #1472: Cockpit: composer single newline shows in textarea but collapses to whitespace in sent message, remark-breaks not enabled

## Problem

In the cockpit composer, a single newline (shift+enter) shows as a visible line break in the textarea. After sending, the rendered user message collapses the single newline into a space; only doubled newlines (`\n\n`) produce a visible break in the chat bubble. The composer text and the sent message disagree on what a newline means.

Expected: the message bubble preserves whatever line layout the composer showed. Type three short lines in the composer with one shift+enter between each, hit send, see three lines in the chat bubble. Today the user sees one wrapped paragraph and has to relearn that they need blank lines to break.

### Composer is a plain textarea

A `<textarea>` stores literal newline characters as the user types. Shift+enter inserts a `\n`. The textarea visually breaks lines on `\n` (and on soft-wrap at column width). The sent payload preserves every `\n` verbatim.

### User message rendering: remark-gfm only, no remark-breaks

`web/src/components/cockpit/Markdown.tsx:50-63`:

```tsx
return (
  <MarkdownTextPrimitive
    preprocess={() => text}
    smooth={smooth}
    remarkPlugins={[remarkGfm]}
    className="cockpit-markdown text-sm leading-relaxed"
    components={{
      SyntaxHighlighter: ShikiSyntaxHighlighter,
      CodeHeader,
      table: TableWithScroll,
      blockquote: Blockquote,
    }}
  />
);
```

`remarkPlugins={[remarkGfm]}` is the only plugin. Per CommonMark, single newlines inside a paragraph are treated as soft line breaks (rendered as whitespace), not hard breaks. Only blank lines start new paragraphs.

`UserText` in `web/src/components/cockpit/CockpitView.tsx:432-446` passes the raw composer text through this same `Markdown` component. Per #1108, agent and user messages share the same rendering pipeline so fenced code blocks and inline code work in both directions. The cost of that consolidation is that user prompts inherit the assistant-side CommonMark soft-break semantics.

### Diff-comments markdown uses `breaks: true`

`web/src/components/diff/comments/CommentMarkdown.tsx:21` uses the `marked` parser with `breaks: true`:

```ts
return marked.parse(text, { async: false, breaks: true }) as string;
```

Single newlines on this surface render as `<br>`. So aoe already treats single newlines as hard breaks on the diff-comments surface; the cockpit message surface is the outlier.

## Root cause

`Markdown.tsx:54` enables only `remark-gfm`. The `MarkdownTextPrimitive` does not run a hard-break transformer, so the underlying micromark/CommonMark default applies: single `\n` is whitespace, only `\n\n` starts a new paragraph. The composer shows literal newlines, the renderer drops them. WYSIWYG breaks at the send boundary.

### A. Add `remark-breaks` to the cockpit Markdown plugin chain

Install `remark-breaks` and add it to `Markdown.tsx:54`:

```tsx
import remarkBreaks from "remark-breaks";
...
remarkPlugins={[remarkGfm, remarkBreaks]}
```

`remark-breaks` turns single newlines into `<br>` nodes. Composer text round-trips visibly. Fenced code blocks and other block-level constructs are unaffected (they're matched before the line-break transform runs).

Trade-off: this also affects assistant text rendered through the same component. In practice Claude already emits markdown with blank-line paragraph separators, so the visible difference for agent messages should be near-zero. The exception is multi-line text where the agent inserts a single `\n` mid-thought; with `remark-breaks` that becomes a visible line break instead of soft-wrapping. Acceptable, and arguably more honest to what the agent emitted.

Touch points:
- `web/package.json` — add `remark-breaks` to dependencies.
- `web/src/components/cockpit/Markdown.tsx:54` — append the plugin.

### B. Branch on caller (only apply breaks to user prompts)

Pass a `breaks?: boolean` prop into `Markdown`, default false, and have `UserText` (`CockpitView.tsx:443`) set it to true while `AssistantText` (`CockpitView.tsx:475`) leaves it false. Keeps the agent-side rendering bit-for-bit identical to today.

Trade-off: two-tier markdown rendering inside the cockpit. Subtle drift over time; harder to keep both branches consistent. Worth the cost only if (A) noticeably regresses an agent-text case in QA.

### C. (Skip) Pre-process newlines in the composer payload

Inserting `\n\n` for every `\n` at the composer's `buildFullPrompt` step would also work but mutates the source-of-truth text persisted on the server. Bad: distorts what the user actually typed, breaks tooling that reads the history. Don't.

## Why now

- Visible-on-every-multiline-prompt UX gap. Pasting PW failures, multi-step instructions, anything with intentional line breaks loses them.
- (A) is a one-import, one-plugin-array entry; tiny diff.
- Aligns the cockpit message surface with the diff-comments surface (`breaks: true`).

- **Single shift+enter renders as a visible line break in the sent message.** As a cockpit user, when I type three short lines separated by single shift+enters in the composer and send, then the rendered user message in the chat shows three visually separated lines.
  - Suggested test layer: Vitest + RTL on `Markdown` with `text="line a\nline b\nline c"`, asserting two `<br>` nodes (or the rendered output contains all three lines on separate visual rows). Optionally repeated under mocked Playwright.

- **Double shift+enter renders as a paragraph break.** As a cockpit user, when I type two paragraphs separated by a blank line in the composer and send, then the rendered user message shows two `<p>` blocks with paragraph spacing.
  - Suggested test layer: Vitest + RTL on `Markdown` with `text="para one\n\npara two"`, asserting two `<p>` elements.

- **Fenced code blocks still render as code blocks.** As a cockpit user, when I paste a fenced markdown code block (`` ```ts ... ``` ``) into the composer and send, then the rendered user message renders the code block with shiki highlighting, not as a wall of line-break-separated text.
  - Suggested test layer: Vitest + RTL on `Markdown` with a fenced block fixture, asserting `<code>` / shiki container renders and the fences are absorbed (not literal backticks).

## Open questions

- Apply globally (A) or branch on caller (B)? Default to (A) and revisit only if a real agent-text case regresses. The diff-comments surface already uses `breaks: true` so this matches the project's existing posture.
- Should the docs `cockpit` section (`docs/cockpit/`) mention the markdown rendering model now that user prompts follow it? Probably yes if not already; see if a docs sync ships in the same PR.

## Out of scope

- Replacing the textarea with a richer input.
- Server-side normalization of newlines. Composer text is persisted verbatim; rendering is the only fix needed.

## Cross-references

- #1108 — introduced user-prompt markdown rendering through the same component this issue tunes.
- #1383 — user-story Playwright specs; the regression tests proposed above can extend that pattern.

Seluj78:
## Description

Batched merge of the `cockpit-main` integration branch into `main`. Branch protection on `main` requires a human approval; autonomous PRs land on `cockpit-main` first, then ship up in periodic batches like this one.

Each commit already passed CI and review on its own PR into `cockpit-main`. No new code here; this is the gate-through to `main`.

### Commits in this batch

- `35153eba` fix(cockpit): make failed tool cards foldable after auto-open (#1628)
- `8bd22529` fix(cockpit): preserve single newlines in sent user messages (#1629)
- `bcfb68ef` feat(cockpit): fold consecutive TodoWrite cards into a TodoGroupCard
- `502f68a1` fix(cockpit): wrap long tokens in chat bubbles to stop viewport overflow
- `3de098b4` refactor(web): remove redundant top-bar breadcrumb (#1627)
- `57439d1d` fix(web): mobile right panel becomes a full-viewport picker
- `bc7d87b7` feat(web): auto-focus agent input on sidebar session select (#1617)
- `9398b6c0` fix(web): suppress iOS link callout on sidebar session rows (#1613)
- `bd3935f1` feat(web): auto-resize mobile terminal on keyboard show/hide (#1614)
- `4af194c6` fix(web): repeat-delete on mobile soft-keyboard Backspace hold (#1615)

### Issues closed by this batch

These come from the individual PRs above. They are re-listed here because issues only auto-close when the closing keyword merges into the default branch (`main`), not when those PRs merged into `cockpit-main`.

Closes #1432
Closes #1450
Closes #1451
Closes #1454
Closes #1456
Closes #1467
Closes #1468
Closes #1469
Closes #1472

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] N/A: tests landed with the individual PRs in this batch

**TUI / CLI (e2e test)**
- [x] N/A: tests landed with the individual PRs in this batch

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.8 (Claude Code)

Batch integration PR assembled by an AI agent. Underlying commits authored across prior reviewed PRs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/agent-of-empires/agent-of-empires/pull/1630?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


njbrake (approved):
Note that it's fine for the cockpit docs to keep growing right now but eventually we need to slim it down a lot to be more user friendly. It's getting way too in the weeds of technical details, that a user doesn't need to know. This applies to several other parts of the docs too though it's not just cockpit docs, we'll get there eventually haha

PR #1664, opened 2026-05-30
fix: reconcile session id from disk before restart (serve + TUI)

### LINKED ISSUE #1024: aoe daemon: `session set-session-id` doesn't invalidate in-memory cache
Observed 2026-05-11. `aoe session set-session-id <title> ""` correctly updates sessions.json on disk; the long-running daemon (`aoe -p default` + `aoe serve` processes) keeps the OLD SID in memory and re-writes it on next `aoe session restart`.

Repro:
1. Active session, sessions.json shows agent_session_id=X
2. `aoe session set-session-id <title> ""`
3. Read sessions.json → empty SID ✓ (on-disk update worked)
4. `aoe session restart <title>` → claude is launched with `--resume X` (the OLD value, from daemon memory)
5. sessions.json post-restart now has X back on disk

Effect: any tooling that clears a stale SID via the CLI gets silently undone. Critical when paired with health-scan archiving thrasher transcripts — the session enters a `--resume <archived-SID>` death loop because the daemon won't honor the CLI clear.

Workaround: bounce `aoe serve` and any `aoe -p default` TUI processes.

Fix: invalidate the in-memory cache (or reload from disk) on `set-session-id` mutations. Or have CLI mutations write to disk + send a refresh signal to the daemon over the existing IPC.

markphilipp:
Fixes #1024

## Description

A CLI `aoe session set-session-id <title> ""` (or `--clear`) correctly updates `sessions.json` on disk, but a long-running daemon keeps the **old** `agent_session_id` in memory and re-persists it on the next restart, silently undoing the CLI change. This is critical when paired with health-scan archiving: the session enters a `--resume <archived-sid>` death loop because the daemon won't honor the CLI clear.

Root cause: `agent_session_id` is **not** a `merge_from_tui` field, so the in-memory value is never reloaded from disk on its own — the TUI save loop leaves it alone and the serve poller only catches up on its next `status_poll_loop` tick (~2s). A restart that fires off the stale in-memory instance resumes the wrong conversation and writes the old sid back over the user's change.

### Fix

Add `Instance::reconcile_session_id_from_disk()` — re-reads the disk-authoritative `agent_session_id` — and call it immediately before the restart on **both** long-running-daemon restart paths the issue names:

- `aoe serve` REST `ensure_session` (`src/server/api/sessions.rs`)
- standalone TUI (`aoe -p default`) `restart_selected_session` (`src/tui/home/operations.rs`)

It is a no-op when the instance is absent from storage or the read fails, leaving the in-memory value intact.

The CLI `aoe session restart` and serve startup-recovery paths already read disk fresh, so they need no change.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added unit tests for the reconcile against both a CLI set and a CLI clear landing on disk while a stale in-memory instance is reconciled (`reconcile_session_id_from_disk_picks_up_cli_set`, `..._cli_clear`). The two restart call sites are thin wrappers over the tested function.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Code

### DIFF
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -2238,6 +2238,12 @@ pub async fn ensure_session(
     let restart_result = tokio::task::spawn_blocking(
         move || -> Result<(Instance, crate::session::StartOutcome), Box<(Instance, anyhow::Error)>> {
             let mut inst = instance;
+            // The clone above came from `state.instances`, which lags disk by
+            // up to one status-poll tick (~2s). A CLI `set-session-id` that
+            // landed inside that window already wrote `sessions.json`; restart
+            // here would resume the stale in-memory sid and re-persist it over
+            // the user's change. Re-read the disk-authoritative value first.
+            inst.reconcile_session_id_from_disk();
             // Use kill_clean (vs bare tmux kill) so a remain-on-exit dead
             // pane is respawned-then-killed; bare kill races against the
             // session cache on macOS and can leave the corpse pane behind,
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -743,11 +743,15 @@ impl Instance {
     }
 
     /// Reconcile `agent_session_id` against the disk-authoritative value
-    /// before a restart that would re-persist it. The daemon's in-memory
-    /// instances only catch up to disk on the next `status_poll_loop` tick
-    /// (~2s); inside that window a CLI `aoe session set-session-id` (or
-    /// `--clear`) has already written `sessions.json` while this clone still
-    /// holds the old sid. Restarting from the stale clone resumes the wrong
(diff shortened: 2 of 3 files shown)


jerome-benoit:
A proper fix is to ensure the session id is propagated from the file to the tmux env var, etc. when it's changed and the targeted field also. 

The code already has file change detection code, proper locking around storage, ... the existing code just need to give priority to the session id source at startup: file then env var then session id pooled, etc. to fix the root cause once and for all without introducing yet other race conditions and fragile workarounds. A possible approach is to introduce a serialization at init between all the existing way to get the session id.

njbrake:
Closing due to staleness.

PR #1683, opened 2026-05-31
feat(sandbox): optional SELinux relabel (:z) on bind mounts

alepar:
## Description
Under SELinux in enforcing mode, the sandbox container (`container_t`) is denied
access to aoe's bind mounts (credentials, project worktree): aoe mounts them
without a relabel option, so they keep their host labels (e.g. `user_home_t`).
The agent can't read its synced credentials and the session comes up with a
blank pane.

This adds an opt-in `sandbox.selinux_relabel` option. When enabled, aoe appends
the `:z` (shared) SELinux relabel flag to every sandbox bind mount, so the runtime
relabels the host paths to `container_file_t` at mount time and the container can
read them.

- **Off by default** — no behavior change for existing users; a no-op where
  SELinux isn't enforcing; opt-in because relabeling modifies host labels.
- **`:z` (shared), not `:Z`** — both the credential dir and project worktrees can
  be bind-mounted into multiple concurrent containers (shared auth dir; the same
  project across concurrent sessions). A per-container `:Z` category would lock a
  path to one container and deny the others.
- Only emitted for runtimes that support it (Docker/Podman; not Apple Container).
- Editable in Settings → Sandbox and overridable per profile.
- `docs/guides/podman.md` currently notes that AoE doesn't append SELinux flags
  automatically and directs users to relabel manually; this adds an opt-in to
  have AoE do it (and updates that doc).

## PR Type
- [x] New Feature

## Checklist
- [x] Documentation was updated where necessary (`docs/guides/podman.md`)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the new field follows the existing sandbox bool-field pattern (`mount_ssh`) and is covered by unit tests. The mount-flag behavior has new unit tests in `runtime_base.rs` (`:z`, `:ro,z`, and unsupported-runtime cases). A full e2e would need an SELinux-enforcing host, which CI doesn't provide. Manually verified on an SELinux-enforcing host (rootless Podman): with `selinux_relabel = true`, a sandboxed session on a non-relabeled project dir comes up working, and the mounts gain the `z` option (`container_file_t` applied) with no manual `chcon`.

## AI Usage
- [x] AI was used

Screenshot 2026-05-30 at 22 24 43

## High-Level Summary

This PR introduces a new optional SELinux configuration option for AoE sandbox containers. The feature allows users to enable SELinux relabeling on bind-mounted directories (credentials and project worktree) when running on SELinux-enforcing systems. This solves a permission issue where containers cannot access host paths that maintain their original host security labels.

**What was added:**
- New `sandbox.selinux_relabel` configuration option (disabled by default)
- Per-profile override support for the SELinux setting
- TUI settings page entry to enable/disable the feature
- Updated Podman documentation with guidance on using the new option

**What was modified:**
- Container configuration to track the SELinux relabel flag
- Runtime implementations (Docker, Podman, Apple Container) to conditionally apply the `:z` relabel flag to mount arguments
- Test coverage updated to verify proper flag handling across different runtimes

**What was NOT changed:**
- Default behavior remains unchanged (opt-in feature, disabled by default)
- Existing configuration and settings are unaffected
- No changes to core container execution logic

## Benefits

- **Solves SELinux permission issues**: Users on SELinux-enforcing systems can now run containers without manual workarounds
- **Backward compatible**: Off by default, so existing users are unaffected
- **Safe approach**: Uses `:z` (shared) relabel rather than `:Z` (unshared), allowing multiple containers to access the same mount
- **Configurable**: Can be set globally or per-profile, giving users fine-grained control
- **Runtime-aware**: Only applies to runtimes that support SELinux flags (Docker/Podman), not Apple Container runtime

## Technical Details

The implementation adds a `selinux_relabel: bool` field throughout the configuration stack:
- `SandboxConfig` stores the global setting (serde default: false)
- `SandboxConfigOverride` allows per-profile overrides
- `ContainerConfig` carries the flag to runtime layers
- `RuntimeBase` includes a `supports_selinux_relabel` capability flag (true for Docker/Podman, false for Apple Container)
- Mount argument construction in `build_create_args` dynamically appends `:z` or `:ro,z` when the feature is enabled and the runtime supports it

### DIFF
--- a/docs/guides/podman.md
+++ b/docs/guides/podman.md
@@ -106,4 +106,15 @@ podman pull ghcr.io/agent-of-empires/aoe-sandbox:latest
 
 ### Permission Denied on Bind Mounts
 
-On SELinux-enabled systems (Fedora, RHEL), you may need to relabel project directories so the container can read them. Either disable SELinux for the volume, or relabel with `:Z` / `:z` (one-time, modifies host labels). AoE does not append SELinux flags automatically; if your distribution requires them, add them via `sandbox.extra_volumes` or relabel the project root.
+On SELinux-enforcing systems (Fedora, RHEL), the container runs as the confined `container_t` domain and is denied access to bind-mounted host paths (the credential dir, the project worktree) because they keep their `user_home_t` label. The symptom is a blank agent pane, or "Permission denied" / `?????????` even as root inside the container.
+
+The fix is to relabel those host paths to a container-accessible type. AoE can do this for you: set
+
+```toml
+[sandbox]
+selinux_relabel = true
+```
+
+which appends the `:z` (shared) SELinux relabel flag to every sandbox bind mount, so the runtime relabels the host paths to `container_file_t` at mount time. It is off by default (a no-op on non-SELinux hosts, and it modifies host labels, so it is opt-in). Only Docker and Podman honor the flag; Apple Container ignores it.
+
+Alternatively, relabel manually with `chcon -R -t container_file_t <path>` (one-time; reverted by a later `restorecon`), or make it durable with `semanage fcontext`.
--- a/src/containers/container_interface.rs
+++ b/src/containers/container_interface.rs
@@ -86,6 +86,7 @@ pub fn docker_env_args(entries: &[EnvEntry]) -> (Vec<String>, Vec<(String, Strin
     (argv, inherit)
 }
 
+#[derive(Default)]
 pub struct ContainerConfig {
     pub working_dir: String,
     pub volumes: Vec<VolumeMount>,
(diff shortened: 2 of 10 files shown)


njbrake (approved):

Approving. Clean, well-scoped change.

- Opt-in (`sandbox.selinux_relabel`), off by default, a no-op on non-SELinux hosts, so no behavior change for existing users.
- The mount-arg refactor (ro/plain to an opts vec) is tidy and gated by the `supports_selinux_relabel` capability flag, so Apple Container correctly ignores the flag.
- Full settings wiring is present: FieldKey, SettingField, apply-to-global, apply-to-profile, clear-override, and the SandboxConfigOverride field plus merge logic.
- Unit tests cover the `:z`, `:ro,z`, and unsupported-runtime cases; docs/guides/podman.md is updated.
- The `:z` (shared) vs `:Z` (private) choice is right here, since the credential dir and worktrees get mounted into multiple concurrent containers.

The one gap is the absence of an e2e test, which would need an SELinux-enforcing host that CI does not provide; the arg-building logic is unit-tested and you verified manually on rootless Podman. That is a reasonable tradeoff. Thanks for the contribution!

PR #1731, opened 2026-05-31
fix(session): split agent_session_id into observation and resume intent

### LINKED ISSUE #1024: aoe daemon: `session set-session-id` doesn't invalidate in-memory cache

### LINKED ISSUE #1732: close session-id read-modify-write race across launch sequence
`agent_session_id` is written unconditionally by multiple paths (`persist_session_to_storage`, `clear_session_id_on_disk`, `apply_session_id_updates` fast path) without coordination across writers. A peer write landing inside the launch sequence is silently overwritten.

Window: ~tmux spawn + agent boot on the happy path; several seconds for sandboxed sessions (Docker start + `on_launch` hooks).

Tracked alongside PR #1731's broader fix (architectural split of observation vs. intent + CAS guards).

---
*AI-drafted.*

jerome-benoit:
## Description

The bug as filed (#1024) traces to a design conflation: `agent_session_id` was simultaneously (a) the poller's observation of the agent's current session and (b) the user's intent for the resume target. Multiple writers (poller, CLI, `finalize_launch`, cascade) raced on a single field with no coherence protocol.

Split the field into two with disjoint writers:

- `agent_session_id: Option<String>`: what the poller has observed. Owned by daemon-internal paths (poller, `finalize_launch`, cascade).
- `resume_intent: ResumeIntent` (`Default | Use(String) | Cleared`): user intent. Owned by user-initiated paths (CLI). `#[serde(default)]` for backward compat with existing `sessions.json`; `#[serde(rename)]` pins wire names against future Rust-side renames.

`acquire_session_id` consults intent first:
- `Use(X)`: return `(Some(X), true)` regardless of observation.
- `Cleared`: ignore observation; return `(Some(fresh_uuid), false)` for Claude or `(None, false)` for others. One-shot semantics: auto-promotes to `Default` after the launch.
- `Default`: fall back to observed sid, then retroactive capture, then fresh UUID for Claude.

Other components:
- Generic `Instance::reconcile_from_disk()` reloads all peer-writable fields from disk while preserving runtime-only state. Called at the entry of `start_with_size_opts`, covering Tier-1 and Tier-2 of the resume-fallback cascade.
- CAS guards on `persist_session_to_storage` and `clear_session_id_on_disk` (`expected_prior: Option<&str>`) protect poller-vs-launch races on `agent_session_id`.
- `persist_session_id` (the `finalize_launch` wrapper) does the (sid, intent) writes in a single `Storage::update` flock so a daemon crash cannot freeze disk at `(new_sid, Cleared)` (which would otherwise let the next launch's acquire short-circuit on `Cleared` and orphan the just-created conversation).
- `reset_resume_intent_to_default` (CAS-guarded) implements the cascade Tier-1 `Use(stale_sid)` downgrade (prevents infinite loop on dead pins).
- `apply_session_id_updates` (TUI poller drain) routes through CAS; reloads memory from disk on `Skipped` to converge on the peer.
- `attempting_resume` in `start_with_resume_fallback` is computed post-`start_with_size_opts`, gated on a runtime `last_acquired_existing_sid` flag set by `acquire_session_id`. This prevents fresh Claude launches from mislabeling as `StartOutcome::Resumed` (acquire always assigns a UUID, so a naive read would always look "existing") and avoids the ~2s settle-probe cost on fresh starts.
- CLI `set-session-id` writes `resume_intent` only and rejects cockpit-mode sessions (cockpit manages its own conversation lifecycle via ACP).

**Behavior changes**:
- `aoe session set-session-id <id> ""` is honored on the next launch (one-shot fresh start), whereas the prior behavior had the poller's observed value overwrite the user's clear within ~2s.
- `aoe session set-session-id` on a cockpit-mode session now errors out instead of silently writing a value cockpit ignores.

Supersedes #1664. Closes #1024. Closes #1732.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (`docs/guides/session-resume.md`, `docs/cli/reference.md` regenerated via `cargo xtask gen-docs`)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Unit tests cover: `ResumeIntent` semantics (acquire-flow variants, serde round-trip, missing-field migration), CAS guards on `persist_session_to_storage` / `clear_session_id_on_disk` (mismatch + match), `reset_resume_intent_to_default` (Applied + Skipped), `reconcile_from_disk` picking up peer `resume_intent`, cascade Tier-1 `Use(stale_sid)` downgrade, `persist_session_id` atomic write (sid + intent in one flock), `persist_session_id` intent-CAS-skip path (sid persisted, intent reloaded), `persist_session_id` sid-CAS-skip path (both fields reloaded), `last_acquired_existing_sid` mirrors `acquire_session_id` bool. 2499 lib tests pass.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 via OpenCode

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -294,7 +294,7 @@ Manage session lifecycle (start, stop, attach, etc.)
 * `rename` — Rename a session
 * `capture` — Capture tmux pane output
 * `current` — Auto-detect current session
-* `set-session-id` — Set agent session ID for a session
+* `set-session-id` — Set the resume target for a session (pin a conversation or force a one-shot fresh start)
 * `set-base` — Set or clear the per-session diff base branch. The diff view compares the worktree against this ref instead of the auto-detected default. Useful when the PR target differs from the project default (stacked PRs, hotfix off `release/*`, renamed default branch). See #970
 * `snooze` — Snooze a session for a duration (temporary archive, auto wakes)
 * `unsnooze` — Wake a snoozed session immediately
@@ -428,14 +428,14 @@ Auto-detect current session
 
 ## `aoe session set-session-id`
 
-Set agent session ID for a session
+Set the resume target for a session (pin a conversation or force a one-shot fresh start)
 
 **Usage:** `aoe session set-session-id <IDENTIFIER> <SESSION_ID>`
 
 ###### **Arguments:**
 
 * `<IDENTIFIER>` — Session ID or title
-* `<SESSION_ID>` — Agent session ID to set (pass empty string to clear)
+* `<SESSION_ID>` — Resume target: a UUID/sid pins the next launches to that conversation; an empty string forces a one-shot fresh start (after which the system reverts to auto-resume)
 
 
 
--- a/docs/guides/session-resume.md
+++ b/docs/guides/session-resume.md
@@ -6,7 +6,7 @@ Agent of Empires can persist Claude Code conversation IDs so sessions resume the
 
 When you launch a Claude session through AoE, AoE generates a UUID and passes it to `claude --session-id <uuid>`. Claude uses that UUID for the conversation; AoE records it in `sessions.json`. On every subsequent launch of the same instance, AoE invokes `claude --resume <uuid>` so the conversation picks up where it left off.
 
-A background poller checks `~/.claude/projects/<project>/` every 2 seconds for the most recently modified session file. If the session ID rotates at runtime — for example after `/clear`, `--fork-session`, or starting a fresh `claude` invocation in the same tmux pane — the poller catches the new ID and AoE persists it transparently. Next launch resumes the new conversation, not the stale one.
+A background poller checks `~/.claude/projects/<project>/` every 2 seconds for the most recently modified session file. If the session ID rotates at runtime, for example after `/clear`, `--fork-session`, or starting a fresh `claude` invocation in the same tmux pane, the poller catches the new ID and AoE persists it transparently. Next launch resumes the new conversation, not the stale one.
 
 For sandboxed (Docker) sessions, the poller runs the same scan inside the container via `docker exec`, since Claude writes its session files to the container's filesystem rather than the host's.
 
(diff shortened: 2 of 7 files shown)


njbrake (approved):

Approving. This is the right fix for #1024: the bug traces to `agent_session_id` conflating the poller's observation with the user's resume intent, and splitting those into disjoint-writer fields (`agent_session_id` poller-owned, `resume_intent` user-owned) removes the race at its source rather than patching the symptoms.

What stood out on review:
- The CAS layer (`SidWrite` with `expected_prior` baselines) means a poller write can no longer clobber a peer CLI write. The TUI drain reloading from disk on `Skipped` to converge on the peer is a correct resolution of the daemon-vs-CLI race.
- `persist_session_id` doing the sid write and the `Cleared -> Default` promote in a single flock is well-justified: a split could freeze disk at `(new_sid, Cleared)` and orphan the conversation just created.
- `last_acquired_existing_sid` gating the settle probe so fresh Claude launches honestly report `Fresh` and skip the ~2s probe is a nice touch.
- The test coverage is genuinely exhaustive for concurrency code: CAS match/mismatch on every writer, reconcile picking up peer persist/clear/intent, all acquire variants, the cascade Tier-1 downgrade, atomic both-fields, and the intent-CAS-mismatch path.
- The merge with the just-landed #1735 reconciled `finalize_launch` correctly (env-batch publish plus the new `expected_prior` params).

On CodeRabbit's effective-profile note: it reads as already addressed, since `reconcile_from_disk` opens storage via `effective_profile()` and `persist_session_id` uses the profile threaded down from `finalize_launch`. Flagging in case I missed a spot.

One documented residual: the cascade Tier-1 clear-then-reset is two flocks, so a crash between them leaves `(None, Use(stale_sid))`, self-healed by the cascade re-firing and tracked for atomic unification in #1742. Fine to land as-is.

Thanks for the careful root-cause work on this one.

njbrake (approved):

Re-approving after the two new commits. `reconcile_sidecar_into_disk` is the right addition now that #1735 landed: it adopts a fresh hook-sidecar sid into disk before the pre-launch wipe, closing the window where a `/clear` writes the sidecar but a daemon crash before the next poll tick would otherwise let the wipe destroy the fresh sid. The guards are correct (Claude-only, `Default` intent only so `Use`/`Cleared` override, skip on match, skip when the sid is in `retroactive_capture_excludes` to avoid re-poisoning the cascade, CAS persist with reload-on-skip), and it is sequenced before the `expected_prior_sid` capture so the adopted value is the CAS baseline. Tests cover the adopt path and each noop branch. Looks good to land once CI is green.

PR #1735, opened 2026-05-31
fix(session): close cross-instance race on Claude session-id capture

jerome-benoit:
## Description

Multiple AoE Claude sessions sharing one project path or container volume could cross-assign each other's session UUIDs at launch and after `/clear`, `/new`, or `--fork-session`.

Replaces the anchor-only approach in #1572 with three converging mechanisms in `claude_poll_fn` / `claude_poll_fn_sandboxed`:

1. **Synchronous `AOE_CAPTURED_SESSION_ID` publish at `finalize_launch`** (batched with `AOE_INSTANCE_ID`), so peers' `build_exclusion_set` sees this session's claim before the first poll tick.

2. **Per-tick `compose_exclusion` + revised tie-break** in `scan_claude_project_dir` and `select_claude_session_in_container`: excluded UUIDs are filtered out of candidates; the anchor wins unless an unexcluded candidate is strictly newer. Aligns Claude with the exclusion contract the other agents already use.

3. **Hook-driven sidecar** `/tmp/aoe-hooks/<id>/session_id`, written by Claude's `SessionStart` / `UserPromptSubmit` hooks via a POSIX pipeline with PID-suffixed tempfile + atomic `mv`. Read by the host poller before the disk scan; honors per-instance `extra_excludes`; pre-launch cleanup wipes stale values.

Sandboxed parity: `claude_poll_fn_sandboxed` now threads `instance_id` + `extra_excludes` + `last_known` anchor; `capture_claude_session_id_in_container` runs under `run_with_timeout(5s)` like the other sandboxed scanners.

`CLAUDE_CURSOR_HOOK_EVENTS` is split into `CLAUDE_HOOK_EVENTS` (with `session_id_capture`) and `CURSOR_HOOK_EVENTS` (status writers only) so Cursor does not install a capture command with no consumer.

Related to #1522 (host-side cross-assignment fix landed in #1523; this PR adds the sandboxed-fleet parity and the hook-based fast path). Supersedes #1572 (anchor-only attempt that the maintainer rejected as treating a symptom).

Overlap with #1731 (broader observation/intent split + CAS guards on writers): this PR's synchronous publish at `finalize_launch` and the Claude-specific exclusion plumbing are independent of #1731's `ResumeIntent` refactor; either can land first, with conflict-resolution at the call sites in `instance.rs`.

## PR Type

- [x] Bug Fix

## Checklist

- [x] New and existing tests pass (`cargo test --lib`: 2501 passed, 0 failed)
- [x] Documentation was updated where necessary (rustdoc on changed functions)
- [x] For UI changes: included screenshot or recording (N/A — internal session tracking)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: regression coverage lives at the unit layer where the race occurs. 18 new tests in `src/session/capture.rs` and `src/hooks/*` cover the scanner tie-break (anchor-vs-best, exclusion filtering), sidecar reader (mtime gate, UUID validation), hook command extraction (compact, multi-line, nested, prompt-injection, uppercase UUID, atomic write), and `build_aoe_hooks` restructure. End-to-end coverage of two Claude sessions sharing a project path requires a fake-hook harness; filed as follow-up.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude (Claude Code), reviewed by @jerome-benoit.

**Any Additional AI Details you'd like to share:** Diff went through four adversarial review rounds (oracle architecture critic + behavioral truth-table + plan critic). Issues raised (regex prompt injection, multi-line JSON, sidecar leak past `extra_excludes`, exclusion-leak via anchor in tie-break, uppercase UUIDs, dead match arm) are closed with regression tests.

This PR fixes a race where multiple Claude sessions sharing the same project path or container volume could capture and assign each other’s session UUIDs, causing wrong-session resume and clobbered resume targets.

High-level changes (non-technical)
- Prevents cross-instance session-id theft with three coordinated mechanisms: immediate publish-at-launch, per-poll exclusion + stronger tie-breaks, and a hook-driven per-instance sidecar that surfaces session IDs quickly and atomically.
- Adds per-instance bookkeeping (instance ID + exclusion sets) so pollers avoid adopting IDs claimed by peers.
- Brings parity between host and sandbox/container flows so the fix applies to both local and containerized Claude sessions.
- Adds sidecar read/write logic, hook-generation changes, and ~18 unit tests covering tie-breaks, sidecar behavior, extraction robustness, atomic writes, and related edge cases.

What changed / moved / added / removed (concise)
- Added hook-driven session-id sidecar: hooks write /tmp/aoe-hooks/<instance_id>/session_id via a PID-suffixed tempfile + atomic mv; the host poller reads this before filesystem scanning.
- Replaced the previous anchor-only approach with three converging measures:
  1) synchronous publish of AOE_CAPTURED_SESSION_ID at finalize_launch (batched with AOE_INSTANCE_ID) so peers see claims before the first poll tick;
  2) per-tick compose_exclusion and exclusion-aware scanning with a stricter anchor-vs-newest tie-break (excluded UUIDs filtered out);
  3) sidecar-first capture via hook-written per-instance session_id files with freshness and UUID validation.
- Threaded instance_id, extra_excludes, and last_known anchor through poller and sandboxed/container capture APIs.
- Split hook constants: CLAUDE_HOOK_EVENTS (session_id_capture true for SessionStart and UserPromptSubmit) and CURSOR_HOOK_EVENTS (no capture).
- Added read_hook_session_id with freshness/UUID validation and SESSION_ID_SIDECAR_MAX_AGE.
- Instance lifecycle: pre-launch sidecar cleanup for Claude; Instance::finalize_launch now atomically publishes AOE_INSTANCE_ID and conditionally AOE_CAPTURED_SESSION_ID.
- Tests: ~18 new unit tests; cargo lib tests passed in CI.

Benefits
- Eliminates cross-assignment of session IDs when multiple sessions share paths/volumes.
- Ensures sessions resume their intended conversations and prevents accidental mass reassignment.
- Faster, more reliable capture (hook sidecar) and robust fallback (filesystem scan) that work in both host and sandboxed environments.
- Resilient to malformed/nested payloads, stale sidecars, and concurrent hook invocations.

Technical details (concise)
- HookEvent gained session_id_capture: bool; per-agent hook arrays updated (src/agents.rs).
- New public function read_hook_session_id(instance_id: &str) -> Option<String) and constant SESSION_ID_SIDECAR_MAX_AGE (src/hooks/status_file.rs).
- New hook_command_session_id() generator and build_aoe_hooks now emit session-id extractor commands when session_id_capture is set; extractor collapses JSON, finds the first top-level session_id via UUID-matching, writes via PID-suffixed tempfile, then atomic mv (src/hooks/mod.rs).
- capture_claude_session_id, scan_claude_project_dir, and related functions now accept exclusion sets and honor them; excluded lastSessionId causes rejection.
- Added select_claude_session_in_container and updated capture_claude_session_id_in_container to accept exclusion and known anchor; sandboxed capture runs with a short timeout and threads instance_id/extra_excludes through claude_poll_fn_sandboxed.
- Instance::start_with_size_opts removes stale per-instance session_id sidecar before launch for Claude.
- Documentation updated (guides, quick-start, rustdoc). PR supersedes `#1572` and addresses `#1522`; AI-assisted.

### DIFF
--- a/docs/development/adding-agents.md
+++ b/docs/development/adding-agents.md
@@ -288,6 +288,15 @@ cargo build
 
 Set `hook_config: Some(AgentHookConfig { ... })` in the agent def; the generic `install_hooks()` handles it.
 
+#### `HookEvent` flags
+
+Each entry in `events: &[HookEvent]` carries:
+
+- `name` — the agent's event name (e.g., `"PreToolUse"`).
+- `matcher` — optional pattern for events that need it (e.g., Claude's `Notification` matcher).
+- `status` — `Some("running" | "waiting" | "idle")` to install a status-writer command on this event, or `None` if the event is purely lifecycle.
+- `session_id_capture` — `true` to install an additional command that extracts the active `session_id` from the agent's stdin JSON and writes it atomically to `/tmp/aoe-hooks/<AOE_INSTANCE_ID>/session_id`. Read by [`session-resume`](../guides/session-resume.md)'s poller as the primary capture source. Currently used only by Claude on `SessionStart` and `UserPromptSubmit`; other agents either don't pre-mint UUIDs (so capture is post-launch from disk) or use a different payload schema (Cursor, Codex). Setting both `status: Some(...)` and `session_id_capture: true` emits two commands under the same matcher block; the session-id command is placed first so it consumes stdin before the status writer.
+
 ### Codex (custom TOML)
 
 Codex stores AoE status hooks in the `[hooks]` table in `.codex/config.toml`:
--- a/docs/guides/session-resume.md
+++ b/docs/guides/session-resume.md
@@ -6,9 +6,12 @@ Agent of Empires can persist Claude Code conversation IDs so sessions resume the
 
 When you launch a Claude session through AoE, AoE generates a UUID and passes it to `claude --session-id <uuid>`. Claude uses that UUID for the conversation; AoE records it in `sessions.json`. On every subsequent launch of the same instance, AoE invokes `claude --resume <uuid>` so the conversation picks up where it left off.
 
-A background poller checks `~/.claude/projects/<project>/` every 2 seconds for the most recently modified session file. If the session ID rotates at runtime — for example after `/clear`, `--fork-session`, or starting a fresh `claude` invocation in the same tmux pane — the poller catches the new ID and AoE persists it transparently. Next launch resumes the new conversation, not the stale one.
+AoE tracks the active session ID via two converging sources:
 
-For sandboxed (Docker) sessions, the poller runs the same scan inside the container via `docker exec`, since Claude writes its session files to the container's filesystem rather than the host's.
+1. **Hook sidecar (primary, near-instant).** AoE installs `SessionStart` and `UserPromptSubmit` hooks into `~/.claude/settings.json`. These hooks extract the active `session_id` from Claude's stdin payload and write it atomically to `/tmp/aoe-hooks/<instance-id>/session_id`. The poller reads this file before scanning the filesystem, so runtime rotations via `/clear`, `--fork-session`, or `--continue` are picked up within one poll tick (~2 s).
+2. **Filesystem scan (fallback).** If the sidecar is absent, stale (> 5 min), or invalid, the poller falls back to scanning `~/.claude/projects/<project>/` for the most recent `.jsonl`. Sibling AoE instances sharing the same project path are filtered out via tmux env (`AOE_CAPTURED_SESSION_ID`) so each session keeps its own UUID.
+
+For sandboxed (Docker) sessions, the filesystem scan runs inside the container via `docker exec` (capped at 5 seconds per call). The hook sidecar is host-only today; sandboxed `/clear` adoption falls back to the filesystem scan and resolves within one poll tick.
 
 ## What's covered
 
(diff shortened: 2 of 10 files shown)


njbrake (approved):

Approving. This is the root-cause fix the #1572 discussion converged on, and it implements the design you and @itisaevalex landed on there: synchronous `AOE_CAPTURED_SESSION_ID` publish at `finalize_launch`, per-tick exclusion with the anchor kept as a fallback tie-break, and the hook-driven sidecar, across both the host and sandboxed paths.

What stood out on review:
- The tie-break rewrite in `scan_claude_project_dir` and `select_claude_session_in_container` is correct: excluded UUIDs are filtered, the anchor is adopted unless a fresh strictly-newer unexcluded candidate exists, so `/clear` rotations are picked up while siblings are not stolen.
- Pre-launch sidecar cleanup closes the read-stale-value hole.
- Test coverage is thorough across hook extraction, the container selector branches, and the poll-fn sidecar-first path.

I also ran the hook capture end to end against real `claude` sessions. Claude emits `session_id` as the first top-level key in both `SessionStart` and `UserPromptSubmit` payloads, the grep extractor pulled the correct id from both, and the injection guard held when a fake `session_id` was placed inside the user prompt (Claude escapes the inner quotes, which breaks the structural prefix).

One follow-up filed: the grep-based extraction depends on Claude keeping `session_id` ahead of any nested copy, and the regression tests use synthetic payloads, so a future payload reorder would not be caught by CI. Tracked in #1760 (move the extraction to a JSON-aware `aoe` subcommand). Not a blocker today since the live payload is correct.

Thanks for driving the root-cause analysis here.

PR #1739, opened 2026-05-31
feat(file-watch): server consumer migration; AppState event-driven storage mirror

jerome-benoit:
## Description

Migrate the daemon's `AppState::instances` storage mirror from 2s polling-only to event-driven reload via `FileWatchService`. Depends on #1734 (file-watch primitive).

This PR introduces the in-process Local fast-path API and its first real consumer (`Storage::update`), and threads the live `FileWatchService` through `Instance::*` writers so daemon-side persistence reaches the in-process Local fast-path without round-tripping through the kernel watcher.

**Primitive amendments** (`src/file_watch.rs`):

- Field `tokio_tx: UnboundedSender<DispatchMsg>` on `FileWatchService`.
- API `pub(crate) fn notify_local_change(&self, path: &Path)`. A sticky `dispatcher_dead` fast-path lets noop services and post-shutdown live services skip the canonicalize syscall and channel send.
- `DispatchMsg::Local(PathBuf)` variant. The dispatcher arm produces a synthetic `FileEvent { kind: Upserted, source: Local }` that funnels through the same uniform filter and per-`(SubscriptionId, final_path)` trailing-edge debounce as kernel events.
- `noop()` pre-sets `dispatcher_dead = true` so local notifications on noop services stay silent.
- `WatchError::DispatcherDead` variant: `subscribe_channel` returns it when the live service's dispatcher has terminated, so callers can fall back to polling instead of holding a handle whose channel will never deliver.
- `ExitLatch` drop-guard around `run_dispatcher` flips `dispatcher_dead = true` on every termination path (panic, return, cancellation) before any new subscription can register against a dead dispatcher.

**Storage** (`src/session/storage.rs`):

- `Storage` carries `file_watch: Arc<FileWatchService>`.
- `Storage::new(profile, Arc<FileWatchService>)` is the standard constructor.
- `Storage::new_unwatched(profile)` (ungated `pub fn`) constructs a `Storage` over `FileWatchService::noop()` for short-lived CLI subprocesses and writers without an in-process subscriber.
- `Storage::update` calls `notify_local_change` after each successful `atomic_write` rename for `groups.json` and `sessions.json`.
- `Storage::load*` injects `self.file_watch.clone()` into every loaded `Instance` so daemon and TUI hot paths reach the live service.

**Instance** (`src/session/instance.rs`):

[...]

PR #1734, opened 2026-05-31
feat(file-watch): introduce FileWatchService primitive; migrate logging consumer

jerome-benoit:
## Description

Introduce `FileWatchService` at `src/file_watch.rs`: one `notify::RecommendedWatcher` per process, demultiplexed to N typed subscribers via DI. Migrates the only existing `notify::recommended_watcher` call site (`watch_runtime_filter` in `src/logging.rs`, called from `src/cockpit/runner.rs`) to use the primitive.

**Why**: removes per-feature ad-hoc watchers, standardizes filtering / trailing-edge debouncing / graceful degradation, prepares the ground for three queued consumer migrations (server `AppState` storage-mirror, TUI `HomeView` storage-mirror, config live-reload). PR1 of a 4-PR sequence; ships kernel-event-driven only (no in-process fast path; that lands atomically with the server consumer migration).

**Public API surface (`pub`)**: `FileWatchService::{new, noop, subscribe_channel}`, plus `WatchSpec`, `FileMatcher::{Exact, AnyOf}`, `FileEvent`, `FileEventKind::{Upserted, Removed}`, `EventSource::Kernel`, `WatchError`, `WatchErrorKind`, `SubscriptionHandle` (RAII drop = unsubscribe).

**Operator knob**: `AOE_FILE_WATCH=off` env var returns the noop service (subscribers register but never fire). CI guard added: `notify::recommended_watcher` must appear in exactly one file (`src/file_watch.rs`).

**Behavior contract**: `watch_runtime_filter` semantics preserved byte-identical; verified by `tests/log_filter_watcher_migration.rs` (covers pre-existing-file priming + post-subscribe write + corrupt-content recovery).

**Breaking change in this crate**: `logging::watch_runtime_filter` now takes `Arc<FileWatchService>` as its first argument; the only caller (`cockpit::runner::run`) is updated in this PR.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

## Test Coverage Analysis

**Web dashboard / cockpit (Playwright user story)**
- [x] N/A: this PR doesn't change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle

Tests landed alongside the code:
[...]

PR #1740, opened 2026-05-31
feat(file-watch): migrate TUI HomeView storage mirror to event-driven reload

jerome-benoit:
## Description

Migrate the TUI's `HomeView::reload()` from the existing 5s heartbeat tick (`DISK_REFRESH_INTERVAL`) to event-driven reload via `FileWatchService`. Both #1734 (file-watch primitive) and #1739 (server consumer; `Storage::new` 2-arg signature, `Instance::file_watch` injection, `persist_session_to_storage` `file_watch` parameter) are merged into main, so this PR builds directly on the current main.

The TUI process owns its own `Arc<FileWatchService>` (separate from the daemon's; cross-process subscribers always go through the kernel watcher). The 5s heartbeat remains canonical defense-in-depth; the file-watch path adds sub-tick latency reduction.

**Architecture**:

- `Arc<FileWatchService>` is constructed once on the TUI startup path (`src/tui/mod.rs`) and threaded into `HomeView::new(active_profile, available_tools, file_watch)`.
- `HomeView` carries `file_watch`, `disk_dirty: Arc<AtomicBool>`, and `disk_watch_handles: HashMap<String, DiskWatchEntry>` where `DiskWatchEntry { handle: SubscriptionHandle, forwarder: AbortHandle, canonical_dir: PathBuf }`.
- Per active profile: `subscribe_channel(spec, capacity=16)` against `<profile>/{sessions,groups}.json`; the spawned forwarder drains the receiver and stores `disk_dirty.store(true, Release)`.
- Tick loop in `src/tui/app.rs` consults `decide_disk_refresh(live_idle, heartbeat_due, dirty)`. On `DiskRefreshDecision::Watcher` it calls `HomeView::reload_storage_only` (refreshes storage without re-running the rest of HomeView refresh); on `Heartbeat` it calls full `reload()`. Tick reload errors route through the `handle_tick_reload_storage` free helper into `ReloadFailureState` so the main loop never propagates them.
- `decide_disk_refresh` returns `DiskRefreshDecision::{Heartbeat, Watcher, None}`; variants are declared in priority order. The function gates on `live_idle` first: when live-send is active it returns `None` regardless of heartbeat or dirty state, so watcher kicks during live-send are preserved in the `AtomicBool` latch and consumed on the next eligible tick.
- Dynamic profile rewire uses `rewire_disk_subscriptions`; the canonical drop-then-abort ordering is encapsulated in the `drop_disk_watch_entry` free function (`drop(handle)` first, then `forwarder.abort()`), so every removal path uses the same order. The rewire fast-path detects peer-driven delete-and-recreate of a profile dir by comparing each prior entry's stored `canonical_dir` against the live canonical resolution; a mismatch forces a rewire even when the profile name set is unchanged. `notify` NonRecursive watches do not auto-reattach across an inode change on Linux inotify or macOS FSEvents, so the inode comparison is the canonical way to detect this case.
- `switch_profile` rewires the filtered subscription set so single-profile mode stays correct.
- The `persist_session_to_storage` caller in `HomeView::apply_session_id_updates` passes `&self.file_watch` so TUI-driven SID writes reach the in-process Local fast-path.
- Watcher-init failures (`subscribe_channel` returns `Err`) are surfaced via `ReloadFailureState::record_watcher_init_failure`. The symmetric `ReloadFailureState::clear_watcher_init_failure` runs at the start of every `rewire_disk_subscriptions` pass so the latch clears as soon as the next install succeeds.

**DI architecture**: no global singleton; the TUI Arc is constructed once and threaded explicitly. Lib unit tests in `src/tui/home/tests.rs` continue to use `FileWatchService::noop()` as scaffolding. Async wiring tests live in `src/tui/home/file_watch_tests.rs` (`#[tokio::test]`); watcher and forwarder spawn paths are gated on `tokio::runtime::Handle::try_current()` so plain `#[test]` contexts keep working.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow.

**TUI / CLI (e2e test)**

- [x] Added or updated e2e tests under `tests/e2e/`.

Tests landed alongside the code:

- `src/tui/home/file_watch_tests.rs`: in-module `#[tokio::test]` coverage that `HomeView::new` spawns a forwarder which flips `disk_dirty` on a peer storage update, and that `rewire_disk_subscriptions` drops removed-profile entries.
- `tests/filewatch_tui_adapter_lifetime.rs`: locks the contract on the raw forwarder pipeline (manual `subscribe_channel` plus a manual forwarder task) so a `Storage::update` flips `disk_dirty` independent of `HomeView::new` wiring.
- `tests/filewatch_tui_drop_then_abort.rs`: locks the canonical drop-then-abort ordering by asserting the forwarder exits after `drop(handle)` and that no events reach it afterward.
- `tests/e2e/filewatch_tui_reload.rs`: peer storage update reflected in TUI within sub-tick budget.
- `tests/e2e/filewatch_tui_dynamic_profile.rs`: profile add/remove keeps subscriptions in sync, plus a filtered-mode switch test that proves watcher rewiring follows the new active profile.
- `tests/e2e/filewatch_tui_burst_reload.rs`: two back-to-back peer writes surface within sub-tick budget.

Local CI gates verified clean post-rebase on top of upstream/main: `cargo fmt --all -- --check`, `cargo clippy --features serve --tests --all-targets -- -D warnings`, `cargo test --no-run`, `cargo test --features serve --no-run`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 via OpenCode (Sisyphus orchestration: multi-agent design + implementation + verification cycles).

High-level summary
- Replaces HomeView’s periodic 5s-only storage reload with an event-driven file-watch path while keeping the 5s heartbeat as a fallback.
- Adds a shared in-process FileWatchService created at TUI startup and passed into App/HomeView/DiffView and all storage-creation sites.
- HomeView now maintains per-profile watch subscriptions, a disk_dirty latch, and reload-failure state; file events cause storage-only reloads, heartbeats still perform full reloads.
- Rewires subscriptions on dynamic profile changes: removes watches for deleted profiles, reattaches when a profile directory is recreated (using canonical-dir/inode checks).
- Watcher-init failures are latched (reported to the UI) instead of crashing; successful reinstalls clear the latch.
- Broad test coverage added: async unit tests, tokio-gated regression tests, and multiple e2e tests for forwarder lifetime, drop-then-abort ordering, burst/sub-tick reloads, and dynamic profile rewiring.

Benefits
- Lower latency UI updates for external (peer) storage changes via event-driven reloads.
- Fewer unnecessary full reloads—file events trigger cheaper storage-only refreshes.
- More robust handling of profile directory churn and watcher-init errors without UI crashes.
- Improved confidence through targeted concurrency and integration tests.

Technical details / notable changes
- TUI startup creates FileWatchService (falls back to FileWatchService::noop() with a warning); App::new now accepts file_watch: Arc<FileWatchService>.
- HomeView additions: file_watch: Arc<FileWatchService>, disk_dirty: Arc<AtomicBool>, disk_watch_handles: HashMap<String, DiskWatchEntry>, reload_failure_state; new method HomeView::rewire_disk_subscriptions.
- Per-profile subscriptions subscribe to <profile>/{sessions,groups}.json (bounded capacity, debounce); each subscription spawns a forwarder task that drains events and sets disk_dirty.store(true, Release).
- Tick loop uses decide_disk_refresh(live_idle, heartbeat_due, dirty) → DiskRefreshDecision::{Heartbeat, Watcher, None}: Watcher → reload_storage_only(), Heartbeat → full reload, None → no action. live-send activity defers consuming the dirty latch until eligible.
- persist_session_to_storage and several storage creation sites now call Storage::new(..., file_watch.clone()) so TUI-driven writes use the in-process fast path.
- Strict teardown ordering: drop_disk_watch_entry drops the subscription handle before aborting the forwarder; canonicalization uses a non-creating probe to avoid recreating deleted profile dirs.
- Watcher-init failures recorded in ReloadFailureState and cleared on successful re-installation.
- Many tests updated/added; constructors in tests use FileWatchService::noop() where appropriate.

Potential improvements
- Expose tuning knobs for debounce, capacity, and sub-tick budgets for different environments.
- Add metrics/tracing around watcher-init failures and forwarder lifecycles for easier production diagnosis.
- Explore advanced backpressure/batching for very high-frequency update bursts.


njbrake (approved):

Approving. This is PR3 of the 4-PR file-watch sequence and builds cleanly on the merged #1734 and #1739.

What I checked:

- `decide_disk_refresh` / latch consumption: the dirty latch is only consumed when `live_idle`, so watcher kicks that arrive during live-send survive to the next eligible tick. Verified by the unit tests.
- Teardown ordering: `drop_disk_watch_entry` drops the `SubscriptionHandle` before aborting the forwarder, and the same order falls out of `DiskWatchEntry` field-drop order on `HomeView` shutdown. Documented and tested.
- `rewire_disk_subscriptions`: set-diff plus inode-invalidation via the stored `canonical_dir`, with the no-op fast path returning before the watcher-init latch is cleared.
- All prerequisite APIs exist in main (`Storage::new` 2-arg, `subscribe_channel`, `spawn_supervised`, `get_profile_dir`, `FileWatchService::{new,noop,subscriber_count}`).

CodeRabbit raised three findings; all are addressed in the final tree and I agree with each:

1. Reconcile watchers on every multi-profile reload, not only on a profile-set delta. Fixed: rewire now runs unconditionally in the `active_profile.is_none()` branch.
2. Keep the watcher-init failure latched until a real reinstall attempt. Fixed: `clear_watcher_init_failure` now runs after the unchanged-set fast path, with a regression test.
3. Reorder the test harness fields to match the canonical drop-then-abort pattern. Fixed.

Non-blocking suggestions for a possible follow-up, no need to hold this PR:

- Single-profile mode does not re-run `rewire_disk_subscriptions` on ticks, so a peer delete-and-recreate of the currently filtered profile dir falls back to the 5s heartbeat until the next profile switch. Correctness is preserved; only live latency degrades. A one-line comment noting this is intentional would help.
- The dialog title is "Watcher Warning" while the PR body and commits describe a "Reload Failed" dialog; doc drift only.
- In multi-profile mode rewire runs a `canonicalize` per profile on every reload tick. Negligible for typical profile counts.

Nice work on the test coverage across unit, integration, and e2e.

njbrake (approved):

Re-approving after the follow-up fixes. CI and Tests are both green on the merge commit.

Resolved since the first approval:

1. Runtime bug found by actually running the new e2e suite: `rewire_disk_subscriptions` probed prior profile dirs with the dir-creating `get_profile_dir`, which resurrected a peer-deleted profile on the next rewire pass. Switched to the non-creating `get_profile_dir_path`.
2. The dynamic-profile e2e assertion matched the persistent list header rather than the picker row; it now drops the active filter before asserting the removed profile is gone.
3. Merged latest main to pick up the live-test harness `env` field that `file-watch-peer-propagation.spec.ts` depends on.

Verified locally: filewatch e2e suite green, fmt and clippy clean.

PR #1741, opened 2026-05-31
feat(file-watch): config live-reload migration; HomeView config kicks via watcher

jerome-benoit:
## Description

Subscribe the TUI&#39;s `HomeView::refresh_from_config()` path to config-file changes via `FileWatchService`, reusing the same `Arc` constructed for the TUI process by the existing TUI consumer. #1734, #1739, and #1740 are merged into main, so this PR builds directly on the current main.

Cross-process config mutations (user edits config in `$EDITOR` while TUI is running, peer `aoe profile create/delete`) propagate sub-tick without waiting for the next user action.

**Architecture**:

- Reuses `self.file_watch.clone()` from `HomeView` (no second Arc; one `FileWatchService` Arc per TUI process).
- `HomeView` carries `config_dirty: Arc` and `config_watch_handles: HashMap`.
- `ConfigWatchKey` is a typed enum (`Global` | `Profile(String)`) that resists collision with a profile literally named `&#34;&#34;`.
- Subscriptions: global config (`/config.toml`, `FileMatcher::Exact`, `capacity=4`) plus per-profile `config.toml` subscriptions for the on-disk profile set.
- Each forwarder task drains its receiver and calls `config_dirty.store(true, Release)`. Both surfaces share `DiskWatchEntry` because the drop-then-abort teardown protocol is identical.
- Tick loop in `src/tui/app.rs` processes `config_dirty` before `disk_dirty` within a single tick, with both gated on `live_idle`.
- Dynamic profile rewire uses idempotent `rewire_config_subscriptions` with canonical drop-then-abort ordering. The inode-aware fast-path applies to both the global key (compares stored `canonical_dir` against `canonicalize(get_app_dir())`) and per-profile entries.
- `DiskWatchEntry` carries a `(dev, ino)` identity (cfg(unix); `()` on non-Unix) captured at install time. Rewire invalidation passes compare both `canonical_dir` and the stat-fresh identity; either mismatch forces an entry rebuild. `FileWatchService::subscribe_channel` mirrors this self-heal at the primitive level: it re-stats on every call and re-arms the kernel watch when stored identity drifts, so a peer `rm -rf X && mkdir X` of a watched canonical path is detected and re-armed even with refcount &gt; 0. The disk consumer benefits from the same primitive helper.
- `rewire_config_subscriptions` mirrors the disk path&#39;s clear-after-fast-path: it returns early when global needs no install and per-profile `to_add`/`to_remove` are both empty (preserving any latched `config_watcher_init_error`); otherwise it clears the latch ahead of the install loop, which re-records on `subscribe_channel` Err so the latch reflects the current pass.
- `switch_profile` calls `rewire_config_subscriptions` directly so the wiring contract is local.
- `reload_storage_only` degrades gracefully if `list_profiles()` fails during a watcher tick, reusing loaded storages for rewires so the whole refresh continues.
- `rewire_after_profile_delete` (Delete-only; the Created path goes through `switch_profile`, which already rewires both surfaces) preserves a pre-existing `info_dialog` when surfacing watcher warnings.
- Asymmetric scope: `rewire_config_subscriptions` always covers the full `list_profiles()` set so peer config edits propagate to the picker UI regardless of mode; `rewire_disk_subscriptions` scopes to the active profile in single-profile mode and to the full set in unified mode.

**Per-source watcher-init latch**: `ReloadFailureState` carries two independent fields, `disk_watcher_init_error` and `config_watcher_init_error`. `record_*_watcher_init_failure` and `clear_*_watcher_init_failure` operate on their respective slots without touching the other. The dialog body emits separate `&#34;- Disk watcher init: ...&#34;` and `&#34;- Config watcher init: ...&#34;` lines. A config init failure recorded in the same tick as a disk rewire is not overwritten by the disk path&#39;s clear, and clearing one slot leaves the other latched.

**Reload-failure dialog**: tick-aggregated reload failures open as `&#34;Reload Failed&#34;` (this is a user-visible behavior change; pre-PR the same tick-loop surfaced as `&#34;Watcher Warning&#34;`, so any external doc or screenshot referencing the old title needs an update). The `rewire_after_profile_delete` path opens a separate `&#34;Watcher Warning&#34;` for the rare case where `list_profiles()` fails after a delete. The recovery-edge cleanup in `App::run` auto-dismisses only `&#34;Reload Failed&#34;` when every reload source returns to healthy; `&#34;Watcher Warning&#34;` is intentionally outside `reload_failure_state` and stays on screen for the user to dismiss. `ReloadFailureState` re-arms the ack latch on healthy-to-failed transitions, and the tick loop refreshes the `Reload Failed` body in place when a new failure source is recorded while the dialog is on screen, so the user sees every failed source without having to dismiss and wait for a re-present. The dialog body trailing line reads `Previous in-memory state preserved; sources retry automatically.` since the four failure sources retry on different triggers (storage tick-driven, config event-driven, watcher-init rewire-driven).

**Deterministic e2e signal**: `HomeView::watcher_config_refresh_count` (`Arc`) increments on every `try_refresh_from_config_watcher` invocation, including parse failures that return Err before `apply_config_to_state` runs. When `AOE_E2E_DEBUG=1` is set on the TUI process the counter exports to `/.aoe_e2e_refresh_count` after every increment; production builds and non-e2e test runs never set the env var, so the file is never written. The harness exposes `enable_e2e_debug_signals()`, `read_watcher_config_refresh_count()`, and `wait_for_watcher_config_refresh_above(baseline, timeout)`; e2e config-watch tests take a baseline before each peer-write and poll the counter for the post-edit refresh, replacing the previous bounded sleeps (0.8-2.5s each) at every site.

**DI**: no global singleton; the TUI Arc is reused. Cross-process config mutations propagate via the kernel watcher; config writers do not call `notify_local_change` because config files are not written through `Storage::update`.

## Behavior note: strict config reload

The watcher path uses strict `resolve_config` (not `resolve_config_or_warn`). A user who manually authors a malformed value type (e.g. `persist_sessions = &#34;yes&#34;` where a bool is expected) sees a one-shot `Reload Failed` dialog. The heartbeat path continues to use the lenient resolver. Forward-compatible defaults (newly added `[auth]` and `session.tie_workdir_to_name`) round-trip without prompts via `#[serde(default)]`; only actively type-mismatched user-authored config triggers the dialog.

## Behavior note: hotkey watcher silence

Watcher-driven refreshes do not surface the `Tool hotkey config errors` dialog; only user-driven settings save / profile switch does. The watcher path stays silent so dismissed dialogs do not re-fire on every tick of unchanged-bad config (locked by `watcher_refresh_does_not_reopen_hotkey_warning_dialog`). `validate_tool_hotkeys` emits per-entry `tracing::warn!` on either path, so the issue is recoverable from the debug log.

## Behavior note: theme propagation via watcher

Watcher-driven refresh applies the new TOML to in-memory state and refreshes the status-hook config cache + picker UI, but the visible theme is repainted by `Action::SetTheme` which is wired to interactive settings-close and profile-switch flows only. A peer edit that changes only the theme color set will land in memory immediately, and the next interactive settings open/close or profile switch repaints with the new colors. This matches the existing `Action::SetTheme` ownership boundary; broadening watcher coverage to dispatch `SetTheme` is tracked separately as a future improvement.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] N/A: this PR doesn&#39;t change a user-facing dashboard flow.

**TUI / CLI (e2e test)**

- [x] Added or updated e2e tests under `tests/e2e/`.

Tests landed alongside the code:

- `tests/filewatch_config_editor_burst.rs`: vim-style write+rename burst collapses to a single delivery via 100ms debounce and tempfile filtering.
- `tests/e2e/filewatch_config_tui.rs`: external config edit propagates to `refresh_from_config` within sub-tick budget.
- `tests/e2e/filewatch_config_profile_switch.rs`: peer edit after profile switch, delete/recreate-same-name (covers the dev+ino self-heal end-to-end on Linux), and create-then-peer-edit all refresh through the watcher path.
- `tests/e2e/filewatch_config_malformed.rs`: malformed TOML does not crash the TUI, the strict-resolve path surfaces a `Reload Failed` dialog, and a subsequent valid write recovers via the recovery-edge auto-clear.
- `tests/e2e/filewatch_config_profile_removal.rs`: profile removal tears down the config-watch path without crashing the TUI.
- `src/tui/home/file_watch_tests.rs`: focused locks for remove/recreate leak prevention, single-profile asymmetry across both `reload_storage_only` and `rewire_after_profile_delete`, reload fallback on `list_profiles` failure, watcher warning surfacing, preserving an existing dialog, `ReloadFailureState` state-machine invariants, no-op rewire preserves the disk-watcher latch, per-source watcher-init latch independence, dialog body 4-source aggregation, recovery-edge survival of `Watcher Warning`, resurrection-prevention via the non-creating profile-path resolver in `rewire_config_subscriptions` (both inode pre-pass and install loop, plus the disk-watch install loop sibling), body refresh of `Reload Failed` for partial recovery and new sources, foreign-dialog skip without acknowledging, `watcher_config_refresh_count` export under `AOE_E2E_DEBUG=1`, and consumer-side identity invalidation for both config and disk rewire on inode drift with same canonical path (cfg(unix)).
- `src/file_watch.rs`: primitive-level `subscribe_rewatches_when_inode_changed_with_refcount_above_zero` locks the self-heal contract via real fs operations and event delivery (cfg(unix)).

Local CI gates verified clean post-rebase on top of upstream/main: `cargo fmt --all -- --check`, `cargo clippy --features serve --tests --all-targets -- -D warnings`, `cargo test --no-run`, `cargo test --features serve --no-run`, plus targeted `cargo test --features serve --lib tui::home:: file_watch:: reload_failure_state`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 via OpenCode (Sisyphus orchestration: multi-agent design + implementation + verification cycles).

## Overview

This PR wires HomeView to watch config files (global and per-profile config.toml) and automatically refresh config state on on-disk changes using the existing FileWatchService, so external edits and profile lifecycle changes propagate to the running TUI without manual action.

## What changed (high-level)

- HomeView now subscribes to a global config.toml plus per-profile config.toml files and sets a config_dirty latch when they change to trigger automatic refreshes.
- Introduced ConfigWatchKey (Global | Profile) and ConfigRefreshOrigin (Interactive | Watcher) to distinguish watch targets and refresh triggers.
- Added rewire_config_subscriptions to manage config-file watches with canonicalization and inode-aware fast-paths; rewire_disk_subscriptions was tightened and both use idempotent clear-then-install ordering.
- Tick loop consumes a gated config-refresh "kick" (skips watcher-driven refresh during live-send) and handles config-driven reloads separately from disk/storage reloads.
- ReloadFailureState split into finer failure slots (storage, config-load, disk-watcher-init, config-watcher-init); aggregated “Reload Failed” dialog now enumerates degraded sources and behavior for watcher vs interactive failures was clarified.
- Watcher-driven refreshes use strict config resolution, suppress tool-hotkey warning dialogs, increment a watcher_config_refresh_count for deterministic e2e signaling, and can export .aoe_e2e_refresh_count when AOE_E2E_DEBUG=1.
- FileWatchService and consumer entries now track a WatchIdentity (dev, ino, optional btime on Unix) so subscribe_channel re-stats and re-arms kernel watches when identity drifts; consumers (DiskWatchEntry) store installed_identity and are invalidated on mismatch.
- Large set of unit and e2e tests added/updated (malformed-config recovery, profile removal/switch/create propagation, editor-save burst coalescing); test harness helpers for e2e sync added.
- Small test seam added to inject list_profiles failures (FAIL_NEXT_LIST_PROFILES + FailNextListProfilesGuard).

## Benefits

- External edits, profile create/delete, and peer changes propagate into the live TUI immediately.
- More precise, per-source failure reporting and clearer recovery semantics.
- Improved e2e determinism via explicit watcher refresh signaling.
- Reuses and extends existing file-watch infrastructure; reduces missed-event windows by rebuilding watches on identity drift.
- Prevents unwanted reloads during active network sends by gating watcher kicks.

## Potential areas for further improvement

- Additional pruning to avoid stale watcher-init latches for profiles deleted before subscribe completes.
- Consistent tracing/logging spans between global and per-profile forwarders.
- Address Windows identity-capture gap (documented and tracked).
- Consider centralized retry/backoff for repeated rewatch failures to reduce noise.

## Technical details (concise)

- HomeView:
  - New fields: config_dirty: Arc<AtomicBool>, config_watch_handles: HashMap<ConfigWatchKey, DiskWatchEntry>, watcher_config_refresh_count: AtomicU64.
  - refresh_from_config(&mut) → refresh_from_config(&mut, ConfigRefreshOrigin).
  - Added try_refresh_from_config_watcher(), rewire_config_subscriptions(current: &[String]).
  - rewire_disk_subscriptions tightened with inode/identity-aware invalidation and idempotent semantics.
- File watch:
  - WatchIdentity: Unix (dev, ino, Option<SystemTime> btime), non-Unix unit type.
  - capture_watch_identity(path) added; DirState stores installed_identity: Option<WatchIdentity>.
  - FileWatchService::subscribe_channel re-stats and re-arms kernel watches when identity drifts; rollback semantics on failure preserve refcounts and clear installed_identity where appropriate.
- Reload/rewire logic:
  - Config forwarders drain receivers, set config_dirty, increment watcher_config_refresh_count, and optionally export .aoe_e2e_refresh_count.
  - Watcher refreshes use strict config resolution and suppress hotkey-warning dialogs; interactive origin still shows those dialogs.
- Tests:
  - Unit + tokio tests for gating, dialog semantics, inode-change invalidation, and rewatch self-heal.
  - Many e2e tests for malformed-config recovery, profile removal/switch/create propagation, and editor-save burst coalescing.
  - TuiTestHarness: enable_e2e_debug_signals(), read_watcher_config_refresh_count(), wait_for_watcher_config_refresh_above().
- Misc:
  - Test seam: FAIL_NEXT_LIST_PROFILES + FailNextListProfilesGuard for injected list_profiles failures.


jerome-benoit:
CodeRabbit nitpick dispositions for the next review pass:

- `tests/filewatch_config_editor_burst.rs:67` and `:150` (drain comment): skipped. The drain idiom is bracketed by a named seed event capture above and the burst write below, and the project guideline favors comments that explain non-obvious "why" over comments that restate observable structure.
- `src/tui/home/file_watch_tests.rs:1116` (`Ordering::Acquire` → `Ordering::Relaxed` on the assertion load): applied in 578d01a9. Production `fetch_add` uses `Relaxed`, and the assertion has no cross-thread synchronization requirement.

njbrake (changes requested):
> _Note: this review was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks for the careful work here; this is a strong capstone to the file-watch series, and most of it holds up well under scrutiny. The architecture mirrors the disk-watch patterns faithfully, the `apply_config_to_state` extraction is a clean 1:1, the strict-vs-lenient resolver split is well reasoned, and the test coverage is unusually thorough. All three earlier CodeRabbit notes are addressed in HEAD.

One issue blocks merge, and it is now confirmed by CI.

## Blocking: `delete_then_recreate_same_name_propagates_via_watcher` fails deterministically on Linux

`Cargo Test (linux)` fails on this one e2e test (110 passed, 1 failed), and it reproduces locally on Linux every time:

```
timed out after 8s waiting for watcher_config_refresh_count > 0 (current = 0)
```

The root cause is a two-layer gap, not test flakiness:

**Layer 1: the invalidation check compares canonical paths, not inodes.** `DiskWatchEntry.canonical_dir` stores `fs::canonicalize(dir)` at install time, and `rewire_config_subscriptions` compares it against a fresh `canonicalize`. When a peer does `rm -rf <profile> && mkdir <profile>` between two heartbeats, the path string is identical and `list_profiles()` never observes the profile missing, so the set-diff is empty, the comparison sees no change, and the rewire fast-path early-returns. Meanwhile inotify silently dropped its watch on the `rm` (IN_IGNORED), so the kernel watch is dead and nothing reinstalls it. The PR body's claim that "peer delete-and-recreate of the same profile name forces a rewire" does not hold for this timing; the existing unit tests pass because they call the rewire while the dir is absent (canonicalize fails, so invalidation fires), which misses the recreate-completes-first case the e2e exercises.

**Layer 2: even a correct invalidation could not re-arm the watch.** `FileWatchService` refcounts kernel watches per canonical dir (`src/file_watch.rs`, `subscribe_channel`), and the disk and config subscriptions for a profile share the same dir. A drop-then-resubscribe inside one consumer moves the refcount 2 to 1 to 2 without touching 0, so `watcher.watch()` is never re-called (`pre_bump == 0` is the only install branch) and the dead watch persists. The service currently has no way to recover a kernel watch that died while refcount > 0.

**Suggested fix:** store `dev+ino` (from `fs::metadata`) at install time and compare against a fresh stat in the invalidation checks (both config and disk consumers), and teach the primitive to restat and rewatch on subscribe when the dir's inode changed since the watch was installed; `watcher.watch()` on an already-watched path is safe to re-call. Note the merged disk path from #1740 has the same latent hole (its e2e never covers a fast same-name recreate), so the primitive-level fix benefits both consumers.

**Acceptable alternative** if you consider fast same-name recreation out of scope for this PR: drop or `#[ignore]` that one e2e with a linked follow-up issue covering both consumers, and adjust the PR body claim. As submitted, the suite is red on the required Linux check.

## Non-blocking observations

- **Theme does not propagate via the watcher path.** `Action::SetTheme` is only dispatched from interactive settings flows, so a peer edit to the theme never repaints until settings open/close. Not a regression, but worth a line in the behavior notes since "config live-reload" implies broader coverage.
- **Latch string parsing vs legal profile names.** `validate_profile_name` permits spaces and colons, so a profile named e.g. `evil config: x` breaks the `strip_prefix("profile ")` / `split_once(" config:")` round-trip in `config_watcher_init_error_references_missing_profile` (the disk sibling's `split_once(": ")` has the same shape). A structured latch keyed by `ConfigWatchKey` would eliminate the parsing; fine as a follow-up.
- **Persistent subscribe failure re-presents the dialog every rewire pass.** The clear-then-record sequence re-arms `dialog_acknowledged`, so under a persistently failing `subscribe_channel` the user cannot dismiss `Reload Failed`. Inherited shape from the disk path; noting it because this PR doubles the paths that can trigger it.
- The dialog body's "will retry on next tick" slightly overpromises for the config source, since a failed parse is not retried until the next file event (which is the correct behavior; only the wording is off).

Verified locally on Linux: `cargo fmt` clean, `cargo clippy --features serve --tests --all-targets -- -D warnings` clean, 553 lib tests pass, both burst integration tests pass, and 5 of the 6 new e2e tests pass.

jerome-benoit:
Thanks for the careful blocking analysis. The two-layer root cause (consumer canonical-only invalidation + primitive refcount masking IN_IGNORED) and the suggested `dev+ino` self-heal direction match what I implemented; the disk path from #1740 benefits from the same primitive-level fix.

## BLOCKING - dev+ino self-heal landed in 3705fe6e

Two-layer fix as you outlined:

- **Primitive (`src/file_watch.rs`)**: `DirState` carries `Option<WatchIdentity>` (`(dev, ino)` on Unix, `()` elsewhere). `subscribe_channel` re-stats every call and re-arms `watcher.watch` when stored identity differs from current stat, regardless of refcount. Refcount rollback on rewatch failure mirrors the first-install rollback semantics.
- **Consumer (`src/tui/home/mod.rs`)**: `DiskWatchEntry` carries the same identity, captured at install time. `rewire_disk_subscriptions` and `rewire_config_subscriptions` invalidation passes compare both `canonical_dir` and the stat-fresh identity; either mismatch forces an entry rebuild.

Tests (cfg(unix)):
- `file_watch::tests::subscribe_rewatches_when_inode_changed_with_refcount_above_zero` locks the primitive contract via real fs ops + event delivery.
- `tui::home::file_watch_tests::rewire_{config,disk}_invalidates_on_inode_change_with_same_canonical_path` lock consumer invalidation by comparing stored identity before and after a same-name peer recreate.

The previously failing e2e (`delete_then_recreate_same_name_propagates_via_watcher`) becomes the primary integration check.

## Audit pass landed in 3709aeab

After 3705fe6e I ran a 4-perspective adversarial review (concurrency / test contracts / pattern compliance / project quality scan) and applied the findings as a follow-up:

- **TOCTOU (None + refcount > 0)**: `subscribe_channel` now treats `current_identity = None` against an existing entry as drift, so a peer dir delete between `canonicalize()` and `capture_watch_identity()` surfaces as `Err` to the caller rather than silently registering against a dead watch.
- **Permanent rewatch failure**: rewatch failure during drift now clears `state.installed_identity = None` in the refcount rollback, so an unrecoverable `watcher.watch` Err (e.g. `inotify_max_user_watches` exhaustion) does not trigger an extra metadata + watch syscall on every subsequent subscribe; the entry stays quiescent until refcount drops to 0.
- **T1 timing**: aligned to the module's `KERNEL_WAIT` (2.5s) instead of `Duration::from_secs(2)`, matching the macOS FSEvents budget convention used by sibling tests.
- **Doc tightening**: `DirState.installed_identity` clarified that identity is captured pre-watch (not post); `WatchIdentity` cross-references the storage asymmetry (Option on the primitive vs bare with `unwrap_or_default` on consumers); `DiskWatchEntry.installed_identity` documents the `(0, 0)` self-healing sentinel on Unix; `capture_watch_identity` gains an `# Errors` section; the lock-invariant comment softens the &#34;idempotent&#34; claim about `Watcher::watch` and enumerates per-backend behavior (inotify wd, FSEvents stream, kqueue fd).

## Non-blocking observations

- **Theme propagation**: documented as a behavior note in the PR body. `Action::SetTheme` is wired to interactive flows only; a peer-edit changing the theme color set will land in memory immediately and repaint on the next interactive settings open/close or profile switch. Broadening watcher coverage to dispatch `SetTheme` is tracked as a future improvement.
- **Latch string parsing vs legal profile names**: tracked by #2099 (typed `WatcherInitError`).
- **Persistent subscribe failure re-presents the dialog every rewire pass**: tracked by #2112. Inherited from the disk path; this PR doubles the surface.
- **Dialog body wording**: &#34;will retry on next tick&#34; is now &#34;sources retry automatically&#34; since the four sources retry on different triggers (storage tick-driven, config event-driven, watcher-init rewire-driven). 3709aeab also drops the sentence-initial &#34;Previous&#34; from that line.
- **Windows gap**: out of scope per the original Unix-focused commit message; tracked by #2113 for `GetFileInformationByHandleEx`-based identity capture. The `WatchIdentity` doc now flags this explicitly.

## State

- Head: `3709aeab`
- Local verification: `cargo fmt --all -- --check` clean, `cargo clippy --features serve --tests --all-targets -- -D warnings` clean, `file_watch::tests` 21 passed (incl. the new primitive self-heal test), `tui::home::file_watch_tests` 26 passed (incl. the new consumer invalidation tests).
- CI: the `Tests` and `CI` workflow runs on `pull_request` are sitting in `action_required` pending re-approval (the same first-time-contributor gate that gated 578d01a9). Re-approval needed to verify Linux green.

Ready for re-review when CI is approved.

njbrake:

I pushed `9aa1013b` to this branch to complete the self-heal; the dev+ino approach was right but two more layers surfaced when running the suite on Linux.

**Why CI was still red after 3705fe6e/3709aeab:**

1. **Inode recycling.** ext4 and overlayfs reuse the freed inode number for an immediate same-name recreate, so the `(dev, ino)` comparison ties and the drift is invisible. CI's three identity-test failures show identical pairs on both sides (e.g. `(84, 1834560) == (84, 1834560)`); reproduced locally on overlayfs. APFS never reuses inode numbers, which is why local macOS verification passed.

2. **The missed-event window.** Even with perfect drift detection, `delete_then_recreate_same_name_propagates_via_watcher` cannot pass: the test writes `config.toml` immediately after the recreate, while the kernel watch is dead. The later rewire rebuilds the subscription, but the write already happened, so the fresh watch never fires for it and the refresh counter never advances.

**What the pushed commit does:**

- Extends `WatchIdentity` to `(dev, ino, Option<SystemTime>)` using `fs::Metadata::created()` (statx btime). A recycled inode cannot reproduce its creation instant. On filesystems without btime the third component is `None` on both sides and the comparison degrades to the previous behavior.
- Rewires that rebuild entries after inode invalidation now kick `config_dirty` / `disk_dirty`, so the next tick re-reads from disk and compensates for events lost during the dead window. Scoped to invalidation rebuilds; plain set-diff adds and removes are unaffected.
- The three identity tests sleep 50ms before the recreate: inode timestamps have jiffy resolution, so a create + recreate inside one kernel tick ties btime too. That is a test-construction artifact; a real recreated profile existed for seconds first.

**Verified locally on Linux:** fmt and `clippy -D warnings` clean; `file_watch` 21/21; `tui::home` 559/559; filewatch_config e2e 6/6 including the previously failing recreate test; full e2e 109/111 with the two `serve::` daemon failures reproducing identically on the unmodified head (sandbox artifact, they pass on CI).

One residual hazard remains documented in the `WatchIdentity` doc: filesystems that both lack btime and recycle inode numbers. The kernel-truth alternative (tracking watch death from the event stream) would close that too; it felt out of proportion for this PR but could fold into #2113's identity follow-up if you think it is worth it.

njbrake (approved):
> _Note: this review was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Approving. All 23 checks are green on `9aa1013b`, including the Linux suite that carried the three identity tests and the recreate e2e.

This landed as a genuinely collaborative fix across three rounds: the original config-watch architecture and test design are yours and held up well under adversarial review; the dev+ino self-heal direction was right; the final layer (btime against inode recycling, plus the missed-window kick for events lost while the kernel watch was dead) came from running the suite on Linux filesystems that recycle inode numbers.

Nice capstone to the file-watch series. The residual no-btime + recycling hazard is documented on `WatchIdentity`, and the follow-ups are tracked (#2099, #2112, #2113). Thanks for the rigor throughout, including the issue filing and the audit pass on your own fix.

PR #1765, opened 2026-06-01
fix(session): wipe hook sidecar in cascade cleanup + tighten sidecar tests and docs

jerome-benoit:
## Description

Follow-up to PR #1731 (merged as `9e200529`) addressing findings from a 4-agent post-merge audit (architectural, adversarial, callchain, hygiene) plus a separate audit of these very fixes' design.

### Primary fix — phantom-sidecar window in cascade

`start_with_resume_fallback`'s Tier-1 -> Tier-2 cascade did not clear the hook sidecar `/tmp/aoe-hooks/<id>/session_id` between launches. If Claude wrote a fresh sid to the sidecar before the Tier-1 probe declared the pane dead (resume failed, Claude minted a new conversation), Tier-2's `reconcile_sidecar_into_disk` adopted that phantom sid and emitted `--resume <phantom>` on the fresh launch, defeating the cascade's "force fresh" semantic. Daemon crash mid-cascade was worse: `retroactive_capture_excludes` is `#[serde(skip)]`, so the known-stale exclusion was lost on restart and the next launch resurrected `stale_sid` for one wasted ~2-7s probe before self-healing.

The fix is a file-level wipe of just `session_id` (matching the line ~1692 pre-launch wipe pattern), ordered **before** `clear_session_id_on_disk`. A daemon crash between the wipe and the disk clear leaves disk holding `stale_sid` (cascade re-fires next launch, self-heals) instead of cleared disk + phantom sidecar (next launch adopts phantom).

Full-dir wipe (`cleanup_hook_status_dir`) was considered but rejected: it would clear the `status` file too, which the TUI status poller reads, producing a phantom `Status::Error` chip flicker in `aoe serve` mode (`read_hook_status=None` + `is_pane_dead=true`).

### Test improvements

- Move `remove_dir_all` before assertions in 6 sidecar tests so a failing assertion does not leak `/tmp/aoe-hooks/<id>` on panic. Verified per test that no test re-reads the sidecar after `reconcile_sidecar_into_disk`.
- Add disk-state assertions to 4 no-op sidecar tests (`noop_when_tool_not_claude`, `_intent_use`, `_intent_cleared`, `_sid_in_retroactive_excludes`). They previously asserted only memory; a regression accidentally writing to disk in a no-op branch would have gone undetected.
- New test `reconcile_sidecar_noop_when_sidecar_absent` covers G3 (`read_hook_session_id` returning `None`), the only previously-untested early-return path. Doubles as the regression test for the cascade wipe.

### Doc tightening

- `resume_intent` field doc overstated reality ("Owned by user-initiated paths (CLI, REST, TUI); never written by the poller"). Only CLI writes non-Default today and the daemon does demote to Default in two paths (one-shot `Cleared` auto-promote, cascade Tier-1 `Use(stale_sid)` downgrade). Replaced with the accurate constraint plus the WHY (so a daemon restart cannot silently undo a user-set pin).
- `reconcile_sidecar_into_disk` doc drops the CAS-skip reload sentence which duplicated the inline comment on the `SidWrite::Skipped` arm.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (rustdoc only, no user-facing docs change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Unit tests cover: cascade wipe regression (`reconcile_sidecar_noop_when_sidecar_absent`), G1/G2/G4 early returns each with both memory and disk assertions, CAS-skip reload path, happy-path adoption. 7 sidecar tests pass; 2566 lib tests total.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 via OpenCode

**Any Additional AI Details you'd like to share:** Two layers of multi-agent audit (architectural / adversarial / callchain / hygiene) drove the design: one round on the merged state of #1731 surfaced the phantom-sidecar bug, a second round on the proposed fix design caught the file-level-vs-full-dir trade-off (UI flicker in `aoe serve`) and the disk-clear ordering inversion before commit.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -399,8 +399,11 @@ pub struct Instance {
     pub agent_session_id: Option<String>,
 
     /// User intent gating `acquire_session_id`. See `ResumeIntent` for
-    /// semantics. Owned by user-initiated paths (CLI, REST, TUI); never
-    /// written by the poller. Disjoint writers from `agent_session_id`.
+    /// semantics. Non-`Default` values (`Use`, `Cleared`) are written only
+    /// by user-initiated CLI commands; daemon-internal paths demote to
+    /// `Default` only (one-shot `Cleared` auto-promote, cascade Tier-1
+    /// `Use(stale_sid)` downgrade), both CAS-guarded, so a daemon restart
+    /// cannot silently undo a user-set pin.
     #[serde(default, skip_serializing_if = "ResumeIntent::is_default")]
     pub(crate) resume_intent: ResumeIntent,
 
@@ -913,10 +916,9 @@ impl Instance {
         *self = disk;
     }
 
-    /// CAS-adopt a fresh hook-sidecar sid into disk. Closes the data-loss
-    /// window where `/clear` writes the sidecar but the daemon crashes
-    /// before the next poll tick persists it: without this, the next
-    /// launch's wipe destroys the fresh sid.
+    /// Closes the data-loss window where `/clear` writes the sidecar but
+    /// the daemon crashes before the next poll tick persists it: without
+    /// this step, the next launch's wipe destroys the fresh sid.
     ///
     /// Claude-only (sole sidecar tool); `Default` intent only (`Use(X)`
     /// and `Cleared` override); excluded sids skipped (cascade re-poison
@@ -2711,6 +2713,7 @@ impl Instance {
             .with_context(|| format!("kill_clean before resume fallback for {}", self.id))?;
 
         self.agent_session_id = None;
+        let _ = std::fs::remove_file(crate::hooks::hook_status_dir(&self.id).join("session_id"));
         let _ = clear_session_id_on_disk(&profile, &self.id, Some(&stale_sid));
         self.retroactive_capture_excludes.insert(stale_sid.clone());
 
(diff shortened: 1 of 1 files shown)


njbrake (approved):

Approving. The cascade fix is a one-line surgical wipe of the hook sidecar `session_id`, mirroring the existing pre-launch pattern at `instance.rs:1691-1692`. It runs after `kill_clean()` so there is no concurrent writer, and it is ordered before `clear_session_id_on_disk` so a mid-cascade daemon crash leaves disk holding `stale_sid` (cascade re-fires and self-heals) rather than a cleared disk plus a phantom sidecar that the next launch would adopt. Rejecting the full-dir `cleanup_hook_status_dir` is the right call; it would also drop the `status` file and flicker a phantom error chip under `aoe serve`.

The `resume_intent` doc rewrite checks out: only the CLI writes non-Default values, and the two daemon demotions named (one-shot `Cleared` auto-promote, cascade Tier-1 `Use(stale_sid)` downgrade) both exist and are CAS-guarded. The test changes are solid: moving `remove_dir_all` ahead of the assertions prevents `/tmp/aoe-hooks` leaks on panic, the added disk-state assertions close a real gap in the no-op branches, and the new sidecar-absent test covers the last untested early return.

I applied the diff locally: it applies cleanly, the 7 sidecar reconcile tests pass, and clippy is clean. One minor note, not blocking: the PR description's doc-tightening bullet says it drops a CAS-skip reload sentence, but the diff actually trims the lead sentence on `reconcile_sidecar_into_disk`; the code change is fine either way.

PR #1862, opened 2026-06-02
fix(sandbox): resolve container terminal login shell, add container_shell override

### LINKED ISSUE #1850: Web dashboard: Container terminal hardcodes /bin/bash, ignores container login shell (oh-my-zsh never loads)

## Problem

Opening the web dashboard right-side terminal and switching to the **Container** tab always lands in `/bin/bash`, even when the container image sets up a different login shell for the dev user (e.g. zsh + oh-my-zsh). The user's prompt, aliases, completions, and oh-my-zsh theme never load because the spawned shell is the wrong shell entirely, not the container user's configured login shell.

The Host tab behaves correctly: it inherits tmux's default-shell (zsh on macOS) as a login shell, so `.zshrc` and oh-my-zsh load. Only the Container tab is broken.

## What's there today

The Container tab is sandbox-only and routes to a dedicated container terminal:

- Frontend toggle: `web/src/components/PairedTerminal.tsx:254` renders the Container button (sandboxed sessions only); `mode === "container"` selects the `container-terminal/ws` path at `web/src/components/PairedTerminal.tsx:39`.
- Backend route `container_terminal_ws` at `src/server/ws.rs:205` calls `start_container_terminal_with_size`.
- `src/session/instance.rs:1616-1619` hardcodes the shell:

  ```rust
  let cmd = container.exec_command(
      Some(&format!("-w {} {}", container_workdir, env_part)),
      "/bin/bash",
  );
  ```

- `exec_command` wraps that into `docker exec -it ... /bin/bash` at `src/containers/runtime_base.rs:405-410`. The `-it` makes it interactive (so bash sources `~/.bashrc`), but the shell is always bash and is **not a login shell**.

By contrast, the Host terminal passes `command: None` (`src/session/instance.rs:1534-1539`), so tmux launches the user's default-shell as a login shell and oh-my-zsh loads.

There is no `container_shell` field in `src/session/config.rs`; the shell is not configurable anywhere.

## Root cause

`src/session/instance.rs:1618` hardcodes `/bin/bash`. The container terminal ignores both the container user's configured login shell and any user preference. A zsh + oh-my-zsh devcontainer therefore drops the user into bare bash with none of their setup.

## Proposal

Ship both layers.

### A. Auto-detect the container login shell (primary, no new setting)

Replace the hardcoded `"/bin/bash"` in `start_container_terminal_with_size` (`src/session/instance.rs:1616-1619`) with a resolved shell that honors the container user, run as a login shell so rc files load.

Shape: resolve the login shell inside the container at terminal-create time, with a safe fallback chain. Prefer the container user's passwd entry, then common shells, then `sh`:

```sh
exec "$( ( getent passwd "$(id -un)" 2>/dev/null | cut -d: -f7 ) || true | grep . \
  || command -v zsh || command -v bash || command -v sh )" -l
```

(Exact resolution can be a small helper; the key behaviors are: honor the container user's `passwd` shell, fall back gracefully, and exec a **login** shell.) This fixes the reported case with no configuration and no migration.

### B. Add a `container_shell` config field (follow-on override)

For users who want to force a specific shell regardless of the image, add an optional `container_shell` setting that, when set, overrides the auto-detection in (A). Empty/unset keeps the (A) auto-detect behavior.

This requires the standard 3-surface parity (per `AGENTS.md`):

- `src/session/config.rs`: new optional field + per-profile override in `profile_config.rs` with `merge_configs()` logic.
- TUI: `FieldKey` in `src/tui/settings/fields.rs`, a `SettingField` entry in the matching `build_*_fields()`, `apply_field_to_global()` + `apply_field_to_profile()`, and a `clear_profile_override()` case in `src/tui/settings/input.rs`.
- Web settings panel (hardcoded React) + its test.

`start_container_terminal_with_size` reads the resolved config value; falls back to (A) when unset.

## Why now

- Onboarding friction: a sandbox user with a customized shell silently loses their entire shell environment in the web Container terminal.
- The Host terminal already does the right thing, so the Container terminal is an inconsistent, surprising regression in behavior between the two tabs.
- Small, well-scoped backend change for (A); (B) is additive.

- As a sandbox user whose container image configures zsh + oh-my-zsh as the dev user's login shell, when I open the web Container terminal, then it starts my login shell (zsh) and sources `.zshrc`, not bare `/bin/bash`.
  - Suggested test layer: Rust unit on the shell-resolution helper (asserts the resolved command honors the passwd shell and includes the login flag); `web/tests/live/` if exercised end-to-end through `aoe serve` against a container.
- As a user whose container has no zsh, when I open the Container terminal, then it falls back to bash (or `sh`) and starts without erroring.
  - Suggested test layer: Rust unit on the fallback chain.
- As a user who sets `container_shell` in config, when I open the Container terminal, then that shell is used and the auto-detection is bypassed; when unset, auto-detection runs.
  - Suggested test layer: Rust unit on config resolution/merge; Vitest + RTL for the web settings input emitting the right key.

## Open questions

- Shell resolution location: resolve inside the container at create-time via a probe command embedded in the exec line (no extra round-trip, portable), or run a separate `docker exec` probe first and pass the literal path? The embedded approach avoids an extra exec but is harder to unit-test; a tiny helper that builds the embedded command string keeps it testable.
- Login vs interactive: a login shell (`-l`) sources profile files; confirm that is the desired behavior for all supported shells (bash sources `.bash_profile`/`.profile` when login; zsh sources `.zprofile` + `.zshrc`). For parity with the Host tab (login shell), `-l` is the right default.
- Does `container_shell` belong on `SandboxConfig` specifically, given it is sandbox-only? (Likely yes.)

## Out of scope

- The Host terminal path (already correct).
- The agent command wrapping (`wrap_command_ignore_suspend`, `src/session/instance.rs:3304`), which intentionally uses `-lc` for non-interactive agent launch.
- Changing the default container image or what shells it ships.

## Cross-references

- #1470 — adjacent web paired-terminal sizing bug; same surface (right-side paired terminal), different symptom.

Seluj78:

## Description

The web dashboard paired terminal's **Container** tab hardcoded `/bin/bash` and ran it as a non-login shell, so a sandbox container user's configured login shell (e.g. zsh + oh-my-zsh) and its profile/rc files never loaded. The **Host** tab already does the right thing: it launches the user's default shell as a login shell, so `.zshrc` and oh-my-zsh load. The Container tab was the inconsistent one.

This ships both layers from the issue:

**A. Auto-detect the container login shell (no configuration).** The hardcoded shell is replaced by a resolver that runs inside the container at terminal-create time and execs a login shell. Resolution order: the container's `$SHELL` (per maintainer feedback on the issue, guarded by `command -v` so a stale or missing value falls through rather than killing the pane), then the passwd entry (the authoritative login shell, e.g. set by `chsh`), then `zsh`, `bash`, `sh`. The resolution expression is single-quoted so it is evaluated by the container's `sh`, not the host shell tmux spawns the session with; the host does not propagate its own `$SHELL` into the container, so this reads the container's value.

**B. Optional `sandbox.container_shell` override.** When set, that shell executable is launched as a login shell, bypassing auto-detection; when unset or blank, auto-detection runs. The value is single-quoted to avoid host-shell breakage and resolved fresh from config (global + profile, via the effective profile) at terminal-create time rather than snapshotted into `SandboxInfo`, since the terminal is a lazily-created UI affordance. Wired across all three surfaces per the settings-parity rule: config struct + per-profile override and merge, the TUI Sandbox settings field, and the web Settings > Sandbox > Advanced input.

No migration: the field is an additive optional config value.

Closes #1850

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] In a sandbox session whose container image sets zsh as the dev user's login shell (e.g. `chsh -s /usr/bin/zsh`), open the web dashboard, open the paired terminal, switch to the **Container** tab, and confirm it starts zsh as a login shell with `.zshrc` / oh-my-zsh loaded (not bare `/bin/bash`).
- [ ] In a sandbox container with no zsh, open the Container tab and confirm it falls back to bash (or sh) and starts without erroring.
- [ ] Set **Sandbox > Container shell** (TUI settings or web Settings > Sandbox > Advanced, or `sandbox.container_shell` in config) to `/bin/bash`, open the Container tab, and confirm that shell is used as a login shell; clear the field and confirm auto-detection resumes.
- [ ] Confirm the **Container shell** field is editable and saves in both the TUI Sandbox settings and the web Settings > Sandbox > Advanced panel, including the per-profile override and "clear override".

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the shell-resolution logic is covered by Rust unit tests on the pure helper (`container_terminal_shell_cmd`, covering $SHELL/passwd/fallback ordering, login flag, whitespace-as-unset, and override quoting) plus a config-merge unit test. A true end-to-end Container-terminal test needs Docker and an image with a non-bash login shell, which is out of scope for CI.

## AI Usage

- [x] AI was used

Investigation, design (with a multi-model debate that disproved a host-shell-quoting concern by reading the tmux command-spawn path), implementation, and tests were drafted by Claude under my direction. I made the design calls, incorporated the maintainer's `$SHELL`-first feedback from the issue thread, reviewed each commit, and ran the validation.

## Summary

This PR changes the Container terminal to start the container user's login shell instead of hardcoding /bin/bash, so user shell initializations (e.g., .zshrc, oh-my-zsh) run in the paired terminal.

## What changed (high level)

- Replaced the hardcoded /bin/bash with an in-container shell resolver that chooses and execs a login shell at terminal-creation time.
- Added a new optional sandbox.container_shell setting that, when set (non-blank), forces a specified shell and bypasses auto-detection; when unset/blank, auto-detection runs.
- Wired the setting through the config struct, per-profile overrides/merge, the TUI Sandbox settings (including clear-override), and the web Settings > Sandbox > Advanced input.
- Updated docs to describe the Container tab behavior and resolution order.
- Added/updated tests:
  - Rust unit tests for the shell-resolution helper and config merge.
  - Playwright/web UI tests for settings persistence and clearing.

## Benefits

- Container paired terminals now load users' shell configurations and environment consistently (prompts, aliases, plugins).
- Reliable in-container resolution prioritizes authoritative sources and safe fallbacks, reducing surprise behavior from stale $SHELL values.
- Users can force a specific shell when needed via configuration, with parity across TUI, web UI, and config files.

## Technical notes (concise)

- The resolver runs inside the container via a single-quoted sh -c '...' wrapper and execs the chosen shell with -l (login).
- Resolution order (evaluated in-container): user shell from passwd via getent passwd "$(id -u)" → $SHELL validated with command -v → zsh → bash → sh (each validated with command -v as applicable).
- New constant CONTAINER_TERMINAL_AUTODETECT_CMD supplies this wrapper; Instance::start_container_terminal_with_size uses it instead of "/bin/bash".
- No host terminal behavior changed and no migration is required.

### DIFF
--- a/docs/guides/web/terminal.md
+++ b/docs/guides/web/terminal.md
@@ -49,6 +49,11 @@ agent terminal; on mobile it is one of the right-panel picker's views.
 The paired shell stays alive in the background when you switch away, so
 its scrollback and focus survive view switches.
 
+For sandboxed sessions, the **Container** tab launches the container
+user's login shell, resolved inside the container (`$SHELL`, then the
+passwd entry, then zsh, bash, sh), so your prompt, aliases, and
+oh-my-zsh setup load just like the Host tab.
+
 ## Reconnect
 
 If the WebSocket drops (network blip, tunnel re-auth, daemon restart),
--- a/docs/guides/web/terminal.md
+++ b/docs/guides/web/terminal.md
@@ -50,9 +50,9 @@ The paired shell stays alive in the background when you switch away, so
 its scrollback and focus survive view switches.
 
 For sandboxed sessions, the **Container** tab launches the container
-user's login shell, resolved inside the container (`$SHELL`, then the
-passwd entry, then zsh, bash, sh), so your prompt, aliases, and
-oh-my-zsh setup load just like the Host tab.
+user's login shell, resolved inside the container (the passwd entry,
+then `$SHELL`, then bash, sh), so your prompt, aliases, and oh-my-zsh
+setup load just like the Host tab.
 
 ## Reconnect
 
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1510,7 +1510,7 @@ impl Instance {
 
         let cmd = container.exec_command(
             Some(&format!("-w {} {}", container_workdir, env_part)),
-            "/bin/bash",
+            CONTAINER_TERMINAL_AUTODETECT_CMD,
         );
 
         // If there are secret env vars, prepend shell exports and use `exec`
@@ -3348,6 +3348,22 @@ fn apply_agent_launch_env(cmd: &mut String, agent: Option<&'static crate::agents
 
 /// Wrap a command to disable Ctrl-Z (SIGTSTP) suspension.
 ///
+/// Command run inside the sandbox container for the web Container terminal tab.
+///
+/// Resolves the container user's login shell at spawn time, inside the container,
+/// and execs it as a login shell so profile/rc files load (parity with the Host
+/// terminal tab, which launches the user's default shell as a login shell).
+/// Resolution order: the container's `$SHELL`, then the passwd entry (the
+/// authoritative login shell, e.g. set by `chsh`), then zsh, bash, sh. Each
+/// candidate is run through `command -v` so an unset, stale, or non-executable
+/// value falls through to the next instead of killing the pane.
+///
+/// The single-quoted body is evaluated by the container's `sh`, not the host
+/// shell tmux uses to spawn the session, so the embedded `$()` runs in the
+/// container. The host does not propagate its own `$SHELL` into the container,
+/// so this reads the container's value, not the host's.
+const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$SHELL" 2>/dev/null || { s=$(getent passwd "$(id -un)" 2>/dev/null | cut -d: -f7); [ -n "$s" ] && command -v "$s" 2>/dev/null; } || command -v zsh || command -v bash || command -v sh)" -l'"#;
+
 /// When running agents directly as tmux session commands (without a parent shell),
 /// pressing Ctrl-Z suspends the process with no way to recover via job control.
 /// This wrapper disables the suspend character at the terminal level before exec'ing
@@ -3488,6 +3504,25 @@ fn pane_has_agent_content(raw_content: &str, tool: &str) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn container_terminal_autodetect_cmd_resolves_login_shell() {
+        let cmd = CONTAINER_TERMINAL_AUTODETECT_CMD;
+        // Resolution order: $SHELL, then the passwd entry, then zsh -> bash -> sh.
+        // Each candidate is guarded by `command -v` so an unset, stale, or
+        // non-executable value falls through rather than killing the pane.
+        assert!(cmd.contains(r#"command -v "$SHELL""#));
+        assert!(cmd.contains("getent passwd"));
+        assert!(cmd.contains(r#"command -v "$s""#));
+        assert!(cmd.contains("command -v zsh"));
+        assert!(cmd.contains("command -v bash"));
+        assert!(cmd.contains("command -v sh"));
+        // Login shell so profile/rc files load, matching the Host terminal tab.
+        assert!(cmd.contains("-l"));
+        // Single-quoted body: the embedded command substitution is evaluated by
+        // the container's sh, not the host shell tmux spawns the session with.
+        assert!(cmd.starts_with("sh -c '"));
+    }
 
     struct CodexHomeGuard(Option<String>);
     impl CodexHomeGuard {
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -3353,16 +3353,19 @@ fn apply_agent_launch_env(cmd: &mut String, agent: Option<&'static crate::agents
 /// Resolves the container user's login shell at spawn time, inside the container,
 /// and execs it as a login shell so profile/rc files load (parity with the Host
 /// terminal tab, which launches the user's default shell as a login shell).
-/// Resolution order: the container's `$SHELL`, then the passwd entry (the
-/// authoritative login shell, e.g. set by `chsh`), then zsh, bash, sh. Each
-/// candidate is run through `command -v` so an unset, stale, or non-executable
-/// value falls through to the next instead of killing the pane.
+/// Resolution order: the passwd entry (the authoritative login shell, what
+/// `chsh` writes and what `login(1)` reads into `$SHELL`), then the container's
+/// `$SHELL`, then bash, sh. Passwd comes first because `docker exec` never goes
+/// through `login(1)`, so `$SHELL` is usually unset or a generic image default
+/// rather than the user's configured shell. Each candidate is run through
+/// `command -v` so an unset, stale, or non-executable value falls through to the
+/// next instead of killing the pane.
 ///
 /// The single-quoted body is evaluated by the container's `sh`, not the host
 /// shell tmux uses to spawn the session, so the embedded `$()` runs in the
 /// container. The host does not propagate its own `$SHELL` into the container,
 /// so this reads the container's value, not the host's.
-const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$SHELL" 2>/dev/null || { s=$(getent passwd "$(id -un)" 2>/dev/null | cut -d: -f7); [ -n "$s" ] && command -v "$s" 2>/dev/null; } || command -v zsh || command -v bash || command -v sh)" -l'"#;
+const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$(getent passwd "$(id -u)" 2>/dev/null | cut -d: -f7)" 2>/dev/null || command -v "$SHELL" 2>/dev/null || command -v bash || command -v sh)" -l'"#;
 
 /// When running agents directly as tmux session commands (without a parent shell),
 /// pressing Ctrl-Z suspends the process with no way to recover via job control.
@@ -3508,15 +3511,16 @@ mod tests {
     #[test]
     fn container_terminal_autodetect_cmd_resolves_login_shell() {
         let cmd = CONTAINER_TERMINAL_AUTODETECT_CMD;
-        // Resolution order: $SHELL, then the passwd entry, then zsh -> bash -> sh.
-        // Each candidate is guarded by `command -v` so an unset, stale, or
-        // non-executable value falls through rather than killing the pane.
-        assert!(cmd.contains(r#"command -v "$SHELL""#));
+        // Resolution order: passwd entry first (authoritative, since docker exec
+        // skips login(1) and so $SHELL is usually unset), then $SHELL, then
+        // bash, sh. Each candidate is guarded by `command -v` so an unset, stale,
+        // or non-executable value falls through rather than killing the pane.
         assert!(cmd.contains("getent passwd"));
-        assert!(cmd.contains(r#"command -v "$s""#));
-        assert!(cmd.contains("command -v zsh"));
+        assert!(cmd.contains(r#"command -v "$SHELL""#));
         assert!(cmd.contains("command -v bash"));
         assert!(cmd.contains("command -v sh"));
+        // Passwd is resolved ahead of $SHELL.
+        assert!(cmd.find("getent passwd").unwrap() < cmd.find(r#"command -v "$SHELL""#).unwrap());
         // Login shell so profile/rc files load, matching the Host terminal tab.
         assert!(cmd.contains("-l"));
         // Single-quoted body: the embedded command substitution is evaluated by


Seluj78:
@jerome-benoit I would greatly appreciate your review on this :D

docs/guides/web/terminal.md
```diff
@@ -49,6 +49,14 @@ agent terminal; on mobile it is one of the right-panel picker's views.
 The paired shell stays alive in the background when you switch away, so
 its scrollback and focus survive view switches.
 
+For sandboxed sessions, the **Container** tab launches the container
+user's login shell, auto-detected inside the container at open time
+(the container's `$SHELL`, then the passwd entry, then zsh, bash, sh),
```
jerome-benoit: Strictly speaking, it's the user shell in the container. not sure you need the verbose repeating paraphrasing of the very same thing ...

docs/guides/configuration.md
```diff
@@ -262,6 +260,7 @@ default_terminal_mode = "host"
 | `volume_ignores_strategy` | `"anonymous"` | Volume mounting strategy: `"anonymous"` (default) or `"named"` (required on macOS/VirtioFS to reliably shadow bind-mount subdirectories; named volumes are explicitly removed on session delete) |
 | `auto_cleanup` | `true` | Remove containers when sessions are deleted |
 | `default_terminal_mode` | `"host"` | Paired terminal location: `"host"` or `"container"` |
+| `container_shell` | (none) | Shell executable for the Container terminal tab, launched as a login shell (e.g. `/bin/zsh`); auto-detected when unset |
```
jerome-benoit: It's not needed and dangerous, a login shell for a user is always defined on a *nix system. any implementation that would try to bypass it expose to failure. 

The only tunable needed is the user on the container.

src/session/instance.rs
```diff
@@ -3348,6 +3361,38 @@ fn apply_agent_launch_env(cmd: &mut String, agent: Option<&'static crate::agents
...
+/// terminal tab). Resolution order: the container's `$SHELL` (the conventional
+/// preferred-shell hint, guarded by `command -v` so a stale or missing value
+/// falls through instead of killing the pane), then the passwd entry (the
+/// authoritative login shell, e.g. set by `chsh`), then zsh, bash, sh.
+///
+/// The single-quoted body is evaluated by the container's `sh`, not the host
+/// shell that tmux uses to spawn the session, so the embedded `$()` runs in the
+/// right place. The host does not propagate its own `$SHELL` into the container,
+/// so this reads the container's value, not the host's.
+const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$({ [ -n "$SHELL" ] && command -v "$SHELL"; } || { getent passwd "$(id -un)" 2>/dev/null | cut -d: -f7 | grep . ; } || command -v zsh || command -v bash || command -v sh)" -l'"#;
```
jerome-benoit: it's an over cautious command ... I think the code has already helpers to parse properly `$SHELL` content that can be used also to parse it in a secure manner when coming from a container.  

getenv is redundant, if SHELL is not filled, you have a more serious issue on your image and getenv will not work. KISS.

jerome-benoit (changes requested):

docs/guides/web/terminal.md
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Tightened the wording and dropped the repetition. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/16bc003b

src/session/instance.rs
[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Removed the passwd branch entirely, so there is no stale passwd-derived path left to guard. The probe is now `command -v "$SHELL"` (which also guards existence) then `command -v zsh || bash || sh`. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/16bc003b

src/tui/settings/fields.rs
```diff
@@ -2757,6 +2776,9 @@ fn apply_field_to_global(field: &SettingField, config: &mut Config) {
         (FieldKey::CustomInstruction, FieldValue::OptionalText(v)) => {
             config.sandbox.custom_instruction = v.clone();
         }
+        (FieldKey::ContainerShell, FieldValue::OptionalText(v)) => {
+            config.sandbox.container_shell = v.clone();
+        }
```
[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Done: both apply paths (global + profile) now normalize blank/whitespace to None via a shared `normalize_blank` helper. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/0561f200

web/src/components/SettingsView.tsx
```diff
@@ -387,6 +387,13 @@ export function SettingsView({
                 placeholder="Additional instructions for the agent..."
                 multiline
               />
+              <TextField
+                label="Container shell"
+                description="Shell executable for the Container terminal tab, launched as a login shell. Leave empty to auto-detect the container user's login shell."
+                value={(sandbox.container_shell as string) ?? ""}
+                onChange={(v) => saveField("sandbox", sandbox, "container_shell", v || null)}
+                placeholder="e.g. /bin/zsh"
```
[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Done: the web input now trims and saves a blank value as null. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/0561f200

src/session/instance.rs
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Agreed on KISS: simplified to $SHELL-first and dropped getent in https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/16bc003b. I kept a small `command -v zsh || bash || sh` fallback because a fresh `docker exec` does not start a login shell, so the container $SHELL can be empty; the fallback keeps the pane usable in that case. On reusing the existing helpers: `user_shell()`/`user_posix_shell()` read the *host* process env via `std::env`, but here we need the *container's* runtime $SHELL resolved inside the container, so they do not apply.

docs/guides/configuration.md
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

I would like to keep `container_shell`, but framed as an opt-in escape hatch rather than a default path: it defaults to unset (auto-detect runs), and when set it is a single-quoted shell path, so it neither bypasses auto-detection by default nor opens a shell-injection vector. #1850 called it out as a follow-on user story (force a specific shell regardless of the image). The container-user knob you mention is complementary and separate; this is for the cases where the image's configured shell is not what the user wants. Happy to defer to @njbrake on whether the minimal surface is preferred, but my inclination is to keep the override.

jerome-benoit: It's useless and will never be used. It has no real world needs or purpose at all. It's just dead code.

src/session/instance.rs
```diff
@@ -3348,6 +3361,37 @@ fn apply_agent_launch_env(cmd: &mut String, agent: Option<&'static crate::agents
...
+/// and execs it as a login shell so profile/rc files load (parity with the Host
+/// terminal tab). Resolution order: the container's `$SHELL` (the user's shell;
+/// `command -v` doubles as an existence guard so an unset or stale value falls
+/// through instead of killing the pane), then zsh, bash, sh.
+///
+/// The single-quoted body is evaluated by the container's `sh`, not the host
+/// shell that tmux uses to spawn the session, so the embedded `$()` runs in the
+/// right place. The host does not propagate its own `$SHELL` into the container,
+/// so this reads the container's value, not the host's.
+const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$SHELL" 2>/dev/null || command -v zsh || command -v bash || command -v sh)" -l'"#;
```
jerome-benoit: You do not need to validate anything, the validation is already done by system. KISS: fetch the SHELL, validate it using the existing validation code to ensure it's a valid shell and that's all.

Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

I want to keep the passwd lookup and the small fallback, and here is the reasoning.

The premise that $SHELL is always defined holds for an interactive login, but this terminal is a `docker exec`, which does **not** start a login shell. `docker exec` gives the process whatever the image's `ENV` set, nothing more; it does not read the user's passwd entry or set $SHELL. So in a container whose login shell was set with `chsh`/`useradd -s` but whose image never exported `ENV SHELL=`, the container $SHELL is empty, and a $SHELL-only path drops the user into plain `sh` instead of their configured shell. That is the exact case this PR is fixing, and CodeRabbit flagged the same gap for fish/ash users.

On reusing the existing validation: `user_shell()` / `user_posix_shell()` read the **host** process env via `std::env`; they cannot validate a path that only exists inside the container, so they do not apply here. The `command -v` guards in the resolver are the in-container equivalent: when $SHELL is set and valid it wins first (your KISS path is honored), and the passwd entry plus zsh/bash/sh fallback only run when $SHELL is empty or not executable, so they never override a good $SHELL.

If you still feel strongly that the fallback should go, I am happy to defer to @njbrake on the final call.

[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Restored the authoritative passwd lookup with a `command -v` guard, so resolution is now $SHELL then the passwd entry (stale/non-executable falls through) then zsh/bash/sh, all as a login shell. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/8695874f591bb6fceb8bfb0ca2f1b41eceb00739

docs/guides/configuration.md
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Dropped the `container_shell` config field entirely (and its TUI + web surfaces); the Container terminal now always auto-detects the login shell. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/8695874f591bb6fceb8bfb0ca2f1b41eceb00739

src/session/instance.rs
```diff
@@ -3348,6 +3348,22 @@ fn apply_agent_launch_env(cmd: &mut String, agent: Option<&'static crate::agents
...
+/// Resolution order: the container's `$SHELL`, then the passwd entry (the
+/// authoritative login shell, e.g. set by `chsh`), then zsh, bash, sh. Each
+/// candidate is run through `command -v` so an unset, stale, or non-executable
+/// value falls through to the next instead of killing the pane.
+///
+/// The single-quoted body is evaluated by the container's `sh`, not the host
+/// shell tmux uses to spawn the session, so the embedded `$()` runs in the
+/// container. The host does not propagate its own `$SHELL` into the container,
+/// so this reads the container's value, not the host's.
+const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$SHELL" 2>/dev/null || { s=$(getent passwd "$(id -un)" 2>/dev/null | cut -d: -f7); [ -n "$s" ] && command -v "$s" 2>/dev/null; } || command -v zsh || command -v bash || command -v sh)" -l'"#;
```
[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Going to hold off on the `/etc/passwd` manual-parse fallback. On the minimal images that lack `getent` (busybox/alpine), the configured login shell is ash, and the existing `command -v sh` branch already lands the user in busybox ash, so the common no-getent case is covered. The only gap is an image that simultaneously has no `getent`, an unset $SHELL, and a login shell that is not zsh/bash/sh, which is rare enough that I do not think it justifies adding a `while read </etc/passwd` loop to an embedded one-liner.

There is also an open discussion on this thread with @jerome-benoit pulling the other way (drop the passwd lookup entirely, KISS). I would rather not pile more complexity onto the resolver while that is unsettled. Holding the probe shape until @njbrake weighs in on the overall approach.

src/session/instance.rs
njbrake: Good catch, @jerome-benoit  thanks!

njbrake: _Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning is his; the prose is Claude's._

Closing the loop on my "good catch": the KISS instinct is right, but the source is wrong. `$SHELL` is set by `login(1)` from the passwd entry and inherited; `docker exec` never goes through login, so `$SHELL` is usually unset (or a generic image `ENV` default, not what the user `chsh`'d to). A `$SHELL`-first resolver mostly falls through to `sh` and never loads zsh/oh-my-zsh.

The authoritative record is passwd field 7, which is what `chsh` writes and what `login` reads into `$SHELL` anyway. So KISS-correct is passwd-first:

```sh
sh -c 'exec "$(command -v "$(getent passwd "$(id -u)" 2>/dev/null | cut -d: -f7)" 2>/dev/null \
  || command -v "$SHELL" 2>/dev/null \
  || command -v bash || command -v sh)" -l'
```

The `command -v` guard is the validation @jerome-benoit asked for, done in the container where the shell must actually exist, so a stale entry falls through instead of killing the pane.

@Seluj78 swap the passwd and `$SHELL` rungs and drop the explicit `zsh` rung (it is already covered by passwd/`$SHELL`), and that is the version I would merge. I would also leave the getent-only path as is; no need to hand-parse `/etc/passwd` for the alpine edge case.

src/session/instance.rs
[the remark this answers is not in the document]
jerome-benoit: If you really want to keep a lookup of a shell in the container ... but in the end the output MUST:

be the full path the shell binary: /usr/bin/...
validated using the existing validation function to ensure it's a valid shell, if it's not pub, if can be moved and made public.

src/session/instance.rs
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

Applied your passwd-first version verbatim: `command -v "$(getent passwd "$(id -u)" ...)"` then `$SHELL` then bash then sh, dropped the explicit zsh rung and the separate passwd-var guard. https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/40afacc927dc34fabda97a9db881709031275825

src/session/instance.rs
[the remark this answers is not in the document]
Seluj78: **Note:** AI-assisted reply (Claude via Claude Code), reviewed and directed by me. — @Seluj78

@njbrake confirmed keeping the getent-only path, so not adding the `/etc/passwd` hand-parse. The resolver is now passwd-first in https://github.com/agent-of-empires/agent-of-empires/pull/1862/commits/40afacc927dc34fabda97a9db881709031275825.

njbrake (approved):

Issue #1762, opened 2026-06-01
Telemetry: anonymous opt-in usage telemetry with a self-hostable, aggregate-only backend


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

We have no idea how `aoe` is actually used in the wild. We cannot answer basic product questions: how many people run it, how many sessions they keep open, which agents/models they pick, whether they live in the TUI or the cockpit/web dashboard, which version they are on, or which platforms matter. Today the only outbound network call is the GitHub release check in `src/update/mod.rs`, which sends nothing identifying and tells us nothing about usage.

This is a request to add anonymous, opt-in usage telemetry so the maintainer can make informed product decisions, while treating the privacy-sensitive open-source audience as the hard constraint it is. Getting this wrong burns goodwill, so the design below is deliberately conservative: off by default, no PII, no content, honor `DO_NOT_TRACK`, fully documented, and aggregate-only on the server.

The questions we want to answer:
- peak concurrent installs and number of distinct installs over a window (DAU/MAU-ish)
- how many sessions a user has open
- which agent/model each session uses
- TUI vs cockpit/web usage
- which `aoe` version is running
- OS / platform breakdown

## What's there today

There is zero telemetry, analytics, or metrics code. No `posthog` / `sentry` / `opentelemetry` dependency, no usage-reporting endpoint.

The only existing outbound "phone-home" is the update check, which is the right privacy precedent to mirror:

`src/update/mod.rs` builds a `reqwest` client, hits the GitHub releases API, and sends only a `User-Agent: agent-of-empires` header. Its `UpdateCheckMode` enum (`auto` / `notify` / `off`) is the consent pattern to follow:

```rust
pub enum UpdateCheckMode {
    Auto,
    #[default]
    Notify,
    Off,
}
```

`reqwest` is already a dependency (`Cargo.toml`, with `rustls` + `json`), so no new HTTP crate is needed.

The natural config home is `UpdatesConfig` in `src/session/config.rs` (around line 1009); a sibling `TelemetryConfig` fits the existing serde-default pattern.

Per-session data we would report is already modelled in `src/session/instance.rs:262` (`tool`, `detect_as`, `status`, `yolo_mode`, sandbox info, cockpit mode). The version is `env!("CARGO_PKG_VERSION")` (currently `1.9.5`).

[...]

PR #1863, opened 2026-06-02
feat: add anonymous opt-in usage telemetry

njbrake:
> _Note: this PR description was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Description

Implements the opt-in usage telemetry client from #1762, so we can see how aoe is actually used in the wild and prioritize the features that matter most.

The posture is deliberately conservative: off by default, no PII, no content, honors `DO_NOT_TRACK`. It is client side only. Opted-in sends go to a compiled-in default endpoint (`https://telemetry.agent-of-empires.com/v1/ingest`), which `AOE_TELEMETRY_ENDPOINT` overrides (handy for pointing at a local sink to inspect what is sent). A compiled-in `X-Telemetry-Key` is sent so the gateway can shed unkeyed drive-by traffic; it is visible in source, so it is not real auth. The collection gateway lives in a separate private repo and re-sanitizes every field as a defense-in-depth backstop. Until that gateway is live, an opted-in client's sends fail closed (bounded timeout, every error swallowed), so nothing is lost and nothing blocks; the consent surfaces and install-id lifecycle are fully exercisable locally for review.

**What it sends** (opt-in only, aggregate counts via a closed, versioned schema in `src/telemetry/events.rs`):
- `process_start` on boot: surface (cli / tui / serve), aoe version, os, arch.
- `usage_snapshot` from the TUI and `aoe serve` (on start, then hourly): session counts by status, sandboxed / cockpit / yolo counts, counts bucketed by agent and by model family, and `web_seen` / `cockpit_seen`.

A central sanitizer (`src/telemetry/sanitize.rs`) is the single privacy boundary: agent and model strings are coerced to a fixed allowlist (`custom` / `other`); raw commands, paths, titles, branch names, group paths, and prompts are never emitted. The anonymous install id is a random UUID stored in `telemetry.json` (kept out of `config.toml` so it cannot leak via a pasted config), deleted on opt-out, rotatable with `aoe telemetry reset-id`.

**Consent surfaces, across all three frontends:**
- TUI walkthrough: a telemetry pane as the second intro step for new users.
- TUI: a one-time standalone popup for users who finished the walkthrough before telemetry existed. It is gated so it never renders over the changelog or the version update modal, and it has no default focus so a reflexive Enter cannot dismiss it without a deliberate choice. Buttons are keyboard and mouse driven (hover highlights, click commits).
- CLI: `aoe telemetry status | enable | disable | reset-id`.
- Web: a Settings telemetry toggle plus a first-load consent modal. The browser never posts to the telemetry backend; it manages opt-in and the seen flags through the daemon, which owns the install id.

`DO_NOT_TRACK` is an absolute override: it suppresses both sending and id generation regardless of config, and every surface shows the suppressed state rather than ignoring it silently. Sends are fire and forget with a hard 2s timeout and swallow all errors, so telemetry never blocks, stalls on exit, or crashes the tool.

The standalone TUI popup:

```
╭ Usage telemetry ─────────────────────────────────────────────────────────╮
│ Help improve aoe with anonymous usage telemetry?                         │
│                                                                          │
│ It shows us how aoe is actually used, so we can prioritize the           │
│ features that matter most. Off by default.                               │
│                                                                          │
│ When on, aoe sends anonymous counts only: sessions, agents and           │
│ models, your aoe version, and OS. Never prompts, paths, names,           │
│ branch names, or commands.                                               │
│                                                                          │
│                           [Enable]    [Not now]                          │
│             ←/→ or Tab to choose, then Enter to confirm                  │
│        Change it any time under Settings, or with `aoe telemetry`.       │
╰──────────────────────────────────────────────────────────────────────────╯
```

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (ASCII render of the TUI popup above; web modal mirrors the same copy)

## Test Coverage Analysis

- [x] Added or updated a spec and updated `web/tests/coverage-matrix.json` (Vitest + RTL contract tests for `TelemetrySettings` and `TelemetryConsentModal`, the right layer for request payload and modal-choice permutations per the matrix mandate; matrix entry `settings.telemetry-consent` added)

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the consent dialogs, sanitizer, opt-in gate, install-id lifecycle, and "unreachable endpoint never blocks" guarantee are covered deterministically by in-module unit tests plus a Rust integration suite (`tests/telemetry.rs`); a full-binary tmux e2e would add flakiness without covering anything those do not.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Implemented via back-and-forth with @njbrake. The design decisions (opt-in, aggregate-only, DO_NOT_TRACK, install-id placement, consent UX) are his; Claude wrote the code and prose.

This PR adds a privacy-first, opt-in anonymous telemetry system spanning CLI, TUI, server, and web, plus documentation, tests, and a local test sink script.

What changed (high-level)
- New telemetry subsystem: event schema, builders, send/flush APIs, sanitization, persisted anonymous install id (telemetry.json), and state/throttling.
- Two events: process_start (boot) and usage_snapshot (periodic aggregate).
- Opt-in controls on all surfaces:
  - CLI: aoe telemetry status | enable | disable | reset-id
  - TUI: onboarding telemetry page + one-time gated popup
  - Web: first-load consent modal + Telemetry settings tab (browser notifies daemon; daemon controls sending)
  - Server: GET /api/telemetry/status, POST /api/telemetry/consent, POST /api/telemetry/seen
- Strong privacy posture:
  - Telemetry is opt-in by default; DO_NOT_TRACK suppresses install-id generation and sending.
  - Central sanitizer buckets agents/models; raw commands, paths, prompts, titles, branches, group paths, and other free text are never emitted.
  - Anonymous install id kept out of config.toml, deleted on opt-out, and rotatable via reset-id.
- Delivery & ops:
  - Endpoint read from AOE_TELEMETRY_ENDPOINT with a DEFAULT_ENDPOINT fallback; sends are fire-and-forget with bounded timeouts and errors swallowed.
  - Optional compiled-in X-Telemetry-Key header support to let a gateway drop unauthenticated traffic.
- Lifecycle & throttling:
  - CLI process_start throttled to at most once per install per day.
  - usage_snapshot emitted by TUI and server on start and periodically (12-hour), with best-effort final flush on shutdown.
  - Web/cockpit surface “seen” flags reported to daemon and included once per snapshot.
- Tests, docs, tools:
  - Rust unit & integration tests, web Vitest/RTL tests, docs/telemetry.md, CLI docs updates, and scripts/telemetry_sink.py for local testing.

Benefits
- Preserves user privacy through opt-in default, DO_NOT_TRACK respect, aggressive sanitization, and install-id management.
- Consistent consent UX across CLI, TUI, web, and server.
- Low operational risk: bounded timeouts, fire-and-forget sends, throttling, coarse snapshots, and gateway-friendly schema reduce impact and noise.
- Gateway-aligned design: allowlisted fields, short identifiers, and optional key header make downstream validation and rate-limiting easier.

Concise technical summary
- Wire schema/events: src/telemetry/events.rs (SCHEMA_VERSION, ProcessStart, UsageSnapshot).
- Sanitization: src/telemetry/sanitize.rs (agent_bucket, model_bucket).
- Persisted state & throttling: src/telemetry/state.rs (telemetry.json, install_id, last_cli_process_start).
- Core telemetry logic: src/telemetry/mod.rs (opt-in checks, endpoint(), post with timeouts, X-Telemetry-Key, spawn/flush helpers, flush_cli_process_start).
- Feature registry: src/telemetry/features.rs (allowlisted features map).
- Integrations: CLI subcommand, TUI onboarding/dialogs and snapshot hooks, server API/routes and AppState seen flags, web consent modal and settings UI, web client API helpers.
- Tests & docs: tests/telemetry.rs, web component tests, docs/telemetry.md, scripts/telemetry_sink.py.

Areas for review / next steps
- Keep DEFAULT_ENDPOINT disabled by default until the gateway is live and rollout is planned.
- Finalize X-Telemetry-Key ops (distribution, rotation, server expectations).
- Confirm and tune short connect timeout (suggest ~500ms) in addition to total send timeout to avoid stalled sends.
- Consider exposing snapshot interval and throttling knobs for operators/testing while keeping privacy-safe defaults.
- Continue gateway-side re-sanitization and contract pinning; restrict future fields to scalars/booleans/short identifier-like strings and allowlisted maps.

### DIFF
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -1,0 +1,78 @@
+# Telemetry
+
+Agent of Empires can send **anonymous, opt-in** usage telemetry so the
+maintainers can answer basic product questions (how many installs are active,
+how many sessions people keep open, which agents/models/platforms matter, TUI
+vs web). It is designed to be conservative: **off by default**, no PII, no
+content, and it honors `DO_NOT_TRACK`.
+
+## What is sent
+
+Only when you opt in, and only aggregate counts. Two event kinds, both with a
+closed, versioned schema (see `src/telemetry/events.rs`):
+
+- **`process_start`** on boot: surface (`cli` / `tui` / `serve`), aoe version,
+  OS, and CPU arch.
+- **`usage_snapshot`** from the TUI and `aoe serve`, on start and roughly
+  hourly: counts of sessions by status, how many are sandboxed / cockpit /
+  yolo, counts bucketed by agent and by model family, and whether the web
+  dashboard / cockpit was opened.
+
+Every agent and model string passes through a sanitizer
+(`src/telemetry/sanitize.rs`) that coerces it to a fixed allowlist: a custom
+agent command becomes `custom`, an unrecognized model becomes `other`. **Raw
+commands, file paths, titles, branch names, group paths, and prompts are never
+sent.**
+
+## What is never sent
+
+Prompts, file or project paths, session titles, branch names, group paths,
+custom command lines, model strings, hostnames, usernames, or anything derived
+from them. The install id is a random UUID generated locally on opt-in; it is
+never derived from hostname, username, MAC, or filesystem.
+
+## Anonymous install id
+
+Counting distinct installs needs a stable id. On opt-in, aoe generates a random
+`uuid::Uuid::new_v4()` and stores it in `<app_dir>/telemetry.json` (owner-only).
+It is kept **out of `config.toml`** on purpose, since people routinely paste
+their config into bug reports. Opting out deletes the file; `aoe telemetry
+reset-id` rotates it.
+
+## Controlling it
+
+Telemetry is **off by default**. Turn it on or off in any surface:
+
+- **CLI**: `aoe telemetry status | enable | disable | reset-id`
+- **TUI**: Settings → System → Telemetry
+- **Web dashboard**: Settings → Telemetry, or the one-time consent prompt shown
+  on first load
+
+New users also see a telemetry pane in the first-run walkthrough; users who
+finished the walkthrough before telemetry existed get a one-time opt-in popup.
+
+### `DO_NOT_TRACK`
+
+If the `DO_NOT_TRACK` environment variable is set to `1` / `true` / `yes`,
+telemetry is suppressed absolutely: nothing is sent and no install id is
+generated, regardless of the config flag. Every surface shows this suppressed
+state explicitly rather than silently ignoring it.
+
+## Failure isolation
+
+Sends are fire-and-forget with a hard ~2s timeout and every error is swallowed
+(logged only at `debug`, `target: "telemetry"`). Telemetry never blocks, stalls
+on exit, or crashes the tool. There is no retry, no offline buffering.
+
+## Backend
+
+The client is **backend-neutral**. The send endpoint is read from the
+`AOE_TELEMETRY_ENDPOINT` environment variable; there is no hardcoded server.
+With no endpoint configured, nothing is ever sent even when opted in. The web
+dashboard never posts to the telemetry backend directly (that would leak the
+browser's IP and User-Agent); it reports local state to `aoe serve`, which owns
+the install id and does all sending.
+
+The collection service lives in a separate repository and is wired in by
+pointing `AOE_TELEMETRY_ENDPOINT` at it; until then opting in is a no-op beyond
+generating the local install id.
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -12,11 +12,16 @@ Only when you opt in, and only aggregate counts. Two event kinds, both with a
 closed, versioned schema (see `src/telemetry/events.rs`):
 
 - **`process_start`** on boot: surface (`cli` / `tui` / `serve`), aoe version,
-  OS, and CPU arch.
-- **`usage_snapshot`** from the TUI and `aoe serve`, on start and roughly
-  hourly: counts of sessions by status, how many are sandboxed / cockpit /
+  OS, and CPU arch. The `cli` surface is throttled to at most once per install
+  per day, so scripting `aoe` in a loop never floods the endpoint.
+- **`usage_snapshot`** from the TUI and `aoe serve`, on start and then every
+  ~12 hours: counts of sessions by status, how many are sandboxed / cockpit /
   yolo, counts bucketed by agent and by model family, and whether the web
   dashboard / cockpit was opened.
+
+In practice that is a handful of small (well under 1 KB) requests per active
+install per day, with no retries and no offline buffering, so a flaky network
+drops events rather than building a backlog.
 
 Every agent and model string passes through a sanitizer
 (`src/telemetry/sanitize.rs`) that coerces it to a fixed allowlist: a custom
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -42,7 +42,9 @@ Counting distinct installs needs a stable id. On opt-in, aoe generates a random
 `uuid::Uuid::new_v4()` and stores it in `<app_dir>/telemetry.json` (owner-only).
 It is kept **out of `config.toml`** on purpose, since people routinely paste
 their config into bug reports. Opting out deletes the file; `aoe telemetry
-reset-id` rotates it.
+reset-id` rotates it. Resetting mints a brand-new id, so that install then
+counts as a new one in the aggregate distinct-install and retention numbers;
+only reset if you actually want to disassociate from prior counts.
 
 ## Controlling it
 
@@ -71,13 +73,20 @@ on exit, or crashes the tool. There is no retry, no offline buffering.
 
 ## Backend
 
-The client is **backend-neutral**. The send endpoint is read from the
-`AOE_TELEMETRY_ENDPOINT` environment variable; there is no hardcoded server.
-With no endpoint configured, nothing is ever sent even when opted in. The web
-dashboard never posts to the telemetry backend directly (that would leak the
+Opted-in events go to the collection gateway at
+`https://telemetry.agent-of-empires.com/v1/ingest`. The gateway validates the
+envelope and re-sanitizes every field as a defense-in-depth backstop, then
+folds the payload into aggregate counts. `AOE_TELEMETRY_ENDPOINT` overrides the
+target (point it at a local sink to see exactly what is sent). A compiled-in
+`X-Telemetry-Key` header lets the gateway drop unkeyed drive-by traffic; it is
+visible in the source, so it is noise-shedding, not authentication.
+
+The web dashboard never posts to the gateway directly (that would leak the
 browser's IP and User-Agent); it reports local state to `aoe serve`, which owns
 the install id and does all sending.
 
-The collection service lives in a separate repository and is wired in by
-pointing `AOE_TELEMETRY_ENDPOINT` at it; until then opting in is a no-op beyond
-generating the local install id.
+**Schema contract.** The wire format is the flat, closed schema in
+`src/telemetry/events.rs`, mirrored by the gateway. New fields must be counts,
+booleans, or short identifier-like strings (and the two allowlisted bucket
+maps); the gateway drops free text, paths, branch-name-like strings, and any
+nested object, so anything richer than a count or flag will not survive ingest.
--- a/docs/telemetry.md
+++ b/docs/telemetry.md
@@ -15,9 +15,13 @@ closed, versioned schema (see `src/telemetry/events.rs`):
   OS, and CPU arch. The `cli` surface is throttled to at most once per install
   per day, so scripting `aoe` in a loop never floods the endpoint.
 - **`usage_snapshot`** from the TUI and `aoe serve`, on start and then every
-  ~12 hours: counts of sessions by status, how many are sandboxed / cockpit /
-  yolo, counts bucketed by agent and by model family, and whether the web
-  dashboard / cockpit was opened.
+  ~12 hours. It is a point-in-time summary of the current install, never a
+  stream of actions:
+  - how many sessions exist and how many are running / idle / errored,
+  - how many use a sandbox, the cockpit, or yolo mode,
+  - a per-agent and per-model-family count (e.g. `{claude: 3, codex: 1}`),
+  - which opt-in features are turned on (see "Feature flags" below),
+  - whether the web dashboard / cockpit was opened since the last snapshot.
 
 In practice that is a handful of small (well under 1 KB) requests per active
 install per day, with no retries and no offline buffering, so a flaky network
@@ -28,6 +32,15 @@ Every agent and model string passes through a sanitizer
 agent command becomes `custom`, an unrecognized model becomes `other`. **Raw
 commands, file paths, titles, branch names, group paths, and prompts are never
 sent.**
+
+### Feature flags
+
+The snapshot includes a small `features` map (allowlisted feature name ->
+on/off) so we can see which opt-in features installs actually turn on. It is
+driven by a registry in `src/telemetry/features.rs`: tracking a newly gated
+feature is one entry there (name + how to read it from config), not a schema
+change. The key set is fixed and the values are booleans, so a flag can never
+carry a path or name, and the gateway forwards only this allowlisted shape.
 
 ## What is never sent
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -224,6 +224,7 @@ async fn main() -> Result<()> {
                 ThemeCommands::Dir => cli::theme::run_dir(),
             };
         }
+        Some(Commands::Telemetry { command }) => return cli::telemetry::run(command),
         Some(Commands::Uninstall(args)) => return cli::uninstall::run(args).await,
         Some(Commands::Update(args)) => return cli::update::run(args).await,
         _ => {}
@@ -265,7 +266,19 @@ async fn main() -> Result<()> {
         migrations::run_migrations()?;
     }
 
-    match cli.command {
+    // Whether this invocation is a short-lived CLI command. The long-running
+    // surfaces (TUI, serve) emit their own `process_start` with the correct
+    // surface and run their own snapshot loop, so they're excluded here.
+    #[cfg(feature = "serve")]
+    let cli_oneshot = cli.command.is_some()
+        && !matches!(
+            cli.command,
+            Some(Commands::Serve(_)) | Some(Commands::CockpitRunner(_))
+        );
+    #[cfg(not(feature = "serve"))]
+    let cli_oneshot = cli.command.is_some();
+
+    let result = match cli.command {
         Some(Commands::Add(args)) => cli::add::run(&profile, *args).await,
         Some(Commands::List(args)) => cli::list::run(&profile, args).await,
         Some(Commands::Remove(args)) => cli::remove::run(&profile, args).await,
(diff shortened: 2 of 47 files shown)


docs/telemetry.md
```diff
@@ -0,0 +1,78 @@
...
+
+Only when you opt in, and only aggregate counts. Two event kinds, both with a
+closed, versioned schema (see `src/telemetry/events.rs`):
+
+- **`process_start`** on boot: surface (`cli` / `tui` / `serve`), aoe version,
+  OS, and CPU arch.
+- **`usage_snapshot`** from the TUI and `aoe serve`, on start and roughly
+  hourly: counts of sessions by status, how many are sandboxed / cockpit /
+  yolo, counts bucketed by agent and by model family, and whether the web
+  dashboard / cockpit was opened.
```
Seluj78: I don't think this is very clear by just ready it.

Also I would like to collect _if possible_ some feature flags in the future. So for example say we add support for openshell, gated behind a feature flag. Then we might want to track the usage of that without having to edit the telemetry docs.

docs/telemetry.md
```diff
@@ -0,0 +1,78 @@
...
+from them. The install id is a random UUID generated locally on opt-in; it is
+never derived from hostname, username, MAC, or filesystem.
+
+## Anonymous install id
+
+Counting distinct installs needs a stable id. On opt-in, aoe generates a random
+`uuid::Uuid::new_v4()` and stores it in `<app_dir>/telemetry.json` (owner-only).
+It is kept **out of `config.toml`** on purpose, since people routinely paste
+their config into bug reports. Opting out deletes the file; `aoe telemetry
+reset-id` rotates it.
```
Seluj78: we might want to warn on this command usage that it will fuck up our stats :P

docs/telemetry.md
```diff
@@ -0,0 +1,78 @@
...
+The client is **backend-neutral**. The send endpoint is read from the
+`AOE_TELEMETRY_ENDPOINT` environment variable; there is no hardcoded server.
+With no endpoint configured, nothing is ever sent even when opted in. The web
+dashboard never posts to the telemetry backend directly (that would leak the
+browser's IP and User-Agent); it reports local state to `aoe serve`, which owns
+the install id and does all sending.
+
+The collection service lives in a separate repository and is wired in by
+pointing `AOE_TELEMETRY_ENDPOINT` at it; until then opting in is a no-op beyond
+generating the local install id.
```
Seluj78: should that be documented until we implement the backend server?

docs/telemetry.md
```diff
@@ -0,0 +1,78 @@
...
+on exit, or crashes the tool. There is no retry, no offline buffering.
+
+## Backend
+
+The client is **backend-neutral**. The send endpoint is read from the
+`AOE_TELEMETRY_ENDPOINT` environment variable; there is no hardcoded server.
+With no endpoint configured, nothing is ever sent even when opted in. The web
+dashboard never posts to the telemetry backend directly (that would leak the
+browser's IP and User-Agent); it reports local state to `aoe serve`, which owns
+the install id and does all sending.
```
Seluj78: how do we plan on shipping the default value for `AOE_TELEMETRY_ENDPOINT` ?

src/main.rs
```diff
@@ -302,5 +315,16 @@ async fn main() -> Result<()> {
             tui::run(&profile, combined).await
         }
         _ => unreachable!(),
+    };
+
+    // Emit `process_start` for a short-lived CLI command AFTER it runs, so the
+    // command's own output isn't delayed. Awaited with a hard timeout so a
+    // dead endpoint can't hang exit; a no-op unless the user opted in and an
+    // endpoint is configured, so default-off users pay nothing.
```
Seluj78: that's good, it means we get a `process_start` event sent to the backend with each `aoe XXX` command ran ? 

Followup to this, can we track CLI command usage? IE if a short lived command is ran, send a count on it.

web/src/components/SettingsView.tsx
```diff
@@ -79,6 +81,7 @@ export function buildSidebar(): SidebarItem[] {
     { kind: "tab", id: "devices", label: "Devices" },
     { kind: "divider", label: "System" },
     { kind: "tab", id: "updates", label: "Updates" },
+    { kind: "tab", id: "telemetry", label: "Telemetry" },
```
Seluj78, line 84: **Blocker** — CI Vitest is failing because of this line. Adding the tab to `buildSidebar()` broke the `buildSidebar > "matches the TUI grouping order"` assertion in `web/src/components/__tests__/SettingsView.test.tsx` (expected 19 entries, got 20). That test file is not in this PR. Update its expected sidebar array to include the telemetry tab.

src/tui/dialogs/telemetry_consent.rs
```diff
@@ -0,0 +1,363 @@
...
+        Self::default()
+    }
+
+    pub fn handle_key(&mut self, key: KeyEvent) -> DialogResult<bool> {
+        match key.code {
+            // Esc is the standard cancel; treat it as a decline.
+            KeyCode::Esc => DialogResult::Submit(false),
+            // Enter only confirms once a side has been chosen; with no focus
+            // it is a no-op so the user can't blow past the prompt.
+            KeyCode::Enter | KeyCode::Char(' ') => match self.selected {
```
Seluj78, line 44: **Required** — In the `DO_NOT_TRACK` branch, `render_dnt` tells the user "Press Enter or Esc to dismiss", but `handle_key` keeps Enter inert here because `selected` stays `None` (the DNT screen renders no buttons and no arrow hint). Only Esc dismisses, so a user who follows the on-screen instruction is stuck. The popup is reachable for DNT users (`app.rs` gates it on `!has_responded_to_telemetry` alone). Treat Enter/Space as `Submit(false)` when `do_not_track()` is active.

src/server/api/telemetry.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Deserialize)]
+pub struct SeenRequest {
+    /// `"web"` or `"cockpit"`.
+    surface: String,
+}
+
+/// Record that the web dashboard / cockpit web UI was opened. Folded into the
+/// daemon's next opt-in snapshot. A no-op (still 200) when telemetry is off so
+/// the client need not branch on consent state.
+pub async fn post_telemetry_seen(
```
Seluj78, line 90: **Required** — `post_telemetry_seen` has no `state.read_only` guard (unlike `set_telemetry_consent`), and neither handler is registered in the `every_mutating_handler_has_read_only_guard` audit in `api/mod.rs`, whose comment says to list every new POST/PATCH/DELETE handler. Add `api/telemetry.rs` + both fn names to that audit list, and either guard `post_telemetry_seen` or drop it from the list with a justification that it is read-safe (it only flips in-memory atomics).

src/telemetry/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+        cockpit_seen,
+        session_creates_since_last_snapshot,
+    })
+}
+
+/// POST a serialized event to the configured endpoint. No-op when no endpoint
+/// is configured. Every error is swallowed and logged at `debug` only.
+async fn post<T: serde::Serialize>(event: &T) {
+    let Some(endpoint) = endpoint() else {
+        tracing::debug!(target: "telemetry", "no AOE_TELEMETRY_ENDPOINT configured; not sending");
```
Seluj78: **Required** — `telemetry` is a new top-level tracing root (and `http.api.telemetry` a new sub-target), but none of `DEFAULT_TARGET_ROOTS`/`KNOWN_SUB_TARGETS` in `src/logging.rs`, `KNOWN_TARGETS` in `web/src/components/settings/LoggingSettings.tsx`, or the table in `docs/development/logging.md` was updated. Per the `log`-root comment in `logging.rs`, an unregistered root is dropped under any expanded-level filter with no global default, so these telemetry diagnostics can never be dialed up. Register the root + sub-target across all three lists.

web/src/App.tsx
```diff
@@ -497,6 +502,33 @@ function AppContent({ loginRequired, onLogout }: { loginRequired: boolean; onLog
...
 
+  // Telemetry: once authenticated and on a writable server, report that the
+  // web dashboard was opened (folded into the daemon's next opt-in snapshot)
+  // and, if the user has not yet answered the opt-in prompt, show the consent
+  // modal. The browser never posts to the telemetry backend; it only talks to
+  // the local daemon. Read-only servers can't persist a choice, so skip.
+  useEffect(() => {
+    // AppContent only renders past the login gate, so reaching here means the
+    // session is usable. Read-only servers can't persist a choice, so skip.
+    if (!serverAboutLoaded || serverAbout?.read_only) return;
```
Seluj78, line 517: **Required** — This new telemetry gating effect (read-only suppression, first-load `fetchTelemetryStatus` then show modal, `reportTelemetrySeen`) is the riskiest new web branch and has no test; the added Vitest specs cover only the leaf components. Add a spec asserting the modal is suppressed on a read-only server and shown when `responded === false && !do_not_track`. Server-side 403 enforcement does exist, so this is a coverage gap, not a security hole.

src/telemetry/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+        return None;
+    }
+    let install_id = state::ensure_install_id()?;
+
+    let mut sessions_by_agent: BTreeMap<String, u32> = BTreeMap::new();
+    let mut sessions_by_model_bucket: BTreeMap<String, u32> = BTreeMap::new();
+    let (mut running, mut idle, mut error, mut cockpit, mut sandboxed, mut yolo) =
+        (0u32, 0u32, 0u32, 0u32, 0u32, 0u32);
+
+    for inst in instances {
```
Seluj78, line 150: I mean this is fine but I would love to have this more modular so that if I want to track usage of the diff comment panel or the web terminal I can do it easier

src/telemetry/sanitize.rs
```diff
@@ -0,0 +1,100 @@
...
+    const FAMILIES: &[(&str, &[&str])] = &[
+        ("claude", &["claude", "sonnet", "opus", "haiku"]),
+        ("openai", &["gpt", "openai", "codex", "o1", "o3", "o4"]),
+        ("gemini", &["gemini"]),
+        ("qwen", &["qwen"]),
+        ("grok", &["grok"]),
+        ("llama", &["llama"]),
+        ("mistral", &["mistral", "mixtral"]),
+        ("deepseek", &["deepseek"]),
+    ];
```
Seluj78, line 52: This is fine but a bit tedious to keep up to date, isn't it ? :/ We need to think about adding new providers and model families

Seluj78 (changes requested):
🙌 🔥 🚀 I love it. Some comments about my thoughts + my AI review

AI thoughts:
Mechanical gates clean: no em dashes, no #[allow(dead_code)], no CLAUDE.md edit, no git config, no hardcoded namespaces, no secrets in logs, no format! inside tracing args, all tracing calls carry a target:. CLI-docs / website-sync / migration triplets satisfied. Settings parity correct (telemetry is global-only, mirrors the Logging/Updates shape). Privacy core is solid: sanitizer allowlist, DO_NOT_TRACK absolute, install id in a separate telemetry.json written via tempfile (0600, no world-readable window), opt-out deletes it, sends are timeout-bounded fire-and-forget. Telemetry routes sit behind the auth middleware layer.

docs/telemetry.md
```diff
@@ -12,12 +12,17 @@ Only when you opt in, and only aggregate counts. Two event kinds, both with a
 closed, versioned schema (see `src/telemetry/events.rs`):
 
 - **`process_start`** on boot: surface (`cli` / `tui` / `serve`), aoe version,
-  OS, and CPU arch.
-- **`usage_snapshot`** from the TUI and `aoe serve`, on start and roughly
-  hourly: counts of sessions by status, how many are sandboxed / cockpit /
+  OS, and CPU arch. The `cli` surface is throttled to at most once per install
+  per day, so scripting `aoe` in a loop never floods the endpoint.
+- **`usage_snapshot`** from the TUI and `aoe serve`, on start and then every
+  ~12 hours: counts of sessions by status, how many are sandboxed / cockpit /
```
Seluj78: why 12h ? that's not a lot at all, and so we loose short lived session information, a lot of it AFAIK

njbrake:

Reviewed this against the collection gateway I'm standing up at `telemetry.agent-of-empires.com` (it validates the envelope, re-sanitizes every property as a defense-in-depth backstop, and forwards on). The client design holds up well: opt-in/default-off, `DO_NOT_TRACK` as an absolute override, the central sanitizer, the install id in `telemetry.json`, fire-and-forget with a hard timeout, and the flat event schema in `events.rs` already matches the gateway's contract (lowercase `surface`, the two allowlisted maps, RFC 3339 `sent_at`, snake_case event names). A few things to settle so the two sides line up:

**1. Production endpoint (required).** `endpoint()` reads only `AOE_TELEMETRY_ENDPOINT` and returns `None` when unset, so released binaries send nowhere. Bake in a default, still overridable by the env var (handy for pointing at a local sink to inspect what is sent):

```rust
const DEFAULT_ENDPOINT: &str = "https://telemetry.agent-of-empires.com/v1/ingest";

pub fn endpoint() -> Option<String> {
    match std::env::var("AOE_TELEMETRY_ENDPOINT") {
        Ok(v) if !v.trim().is_empty() => Some(v.trim().to_string()),
        _ => Some(DEFAULT_ENDPOINT.to_string()),
    }
}
```

Sequencing: this PR is safe to land as-is today (default-off, and with no default endpoint nothing flows). Flip the default on only once the gateway is deployed and the domain resolves, either in a one-line follow-up or here before merge. The default must not point at a domain that does not yet answer.

**2. Fail fast when the endpoint is unreachable.** The isolation is already solid (errors swallowed, 2s hard cap, detached sends for TUI/serve). The one gap: `flush_cli_process_start` awaits the send, and the client has no `connect_timeout`, so a black-holed or slow-DNS endpoint costs that CLI run the full 2s (once per day per install, given the throttle). Add a short connect timeout so an unreachable server fails in well under a second:

```rust
reqwest::Client::builder()
    .user_agent(concat!("agent-of-empires/", env!("CARGO_PKG_VERSION")))
    .connect_timeout(std::time::Duration::from_millis(500))
    .timeout(SEND_TIMEOUT)
    .build()
```

With that, a down or missing server has no user-visible impact: a healthy server still flushes in tens of milliseconds, and a dead one is swallowed in under a second. Combined with default-off, shipping the client is fully decoupled from server availability.

**3. Optional shared key.** The gateway can require a static `X-Telemetry-Key` header and drop requests without it. It is not real auth (it would be visible in this source) but it sheds drive-by noise. If we use it, the client adds one header with a compiled-in constant; if we skip it on the gateway, no client change. Let me know which way so the two sides match.

**4. Pin the contract.** `events.rs` matches today; to keep it from drifting, point `docs/telemetry.md` at the gateway's schema doc. One rule for future fields: the gateway forwards scalars, booleans, and short identifier-like strings, plus the two allowlisted maps; it drops free text, paths, branch-name-like strings, and any other nested object. New metrics should be counts/flags, which is what you would send anyway.

Not needed in the client: any analytics-vendor concepts. Keep sending the raw anonymous `install_id`; the gateway handles the rest. None of this blocks landing behind the default-off gate.

Seluj78:
> Optional shared key

Yeah I think it's required and I don't see a better way of doing it, at least not with auth.

>  Pin the contract

I think the telemetry server should be a sink and accept all events and values, and store them and then just allow queries and dashboards to be built on top of it. This way no need to share the model of events between both codebases, and simpler for updates too

Seluj78 (approved):

PR #1867, opened 2026-06-02
fix(session): kill all tmux session kinds on remove and recovery paths

jerome-benoit:
## Description

Audit of every code path that removes or destroys a session uncovered three non-intentional, undocumented gaps where tmux teardown was incomplete or absent:

1. **`tui/home/operations.rs::force_remove_session`** (recovery path for sessions stuck in `Status::Deleting`) skipped tmux cleanup entirely. Documented as skipping worktree/branch/container; tmux was silently omitted. Live tmux sessions could orphan and continue writing into a worktree being abandoned.

2. **`tmux/terminal_session.rs::PairedTerminal::kill`** and **`tmux/tool_session.rs::ToolSession::kill`** did not swallow the `"can't find session"` stderr that occurs when the process tree kill races tmux's own teardown. Only `Session::kill` swallowed it. Other callers surfaced spurious `"Failed to kill"` warnings on operations that had succeeded.

3. **`server/api/cockpit.rs::cockpit_enable`** (mode-switch tmux to cockpit) called only `instance.kill()`, leaving the terminal pane and tool sub-sessions orphaned after the backend swap.

Meta-finding: zero tests at any layer asserted "session removed implies tmux gone" for the agent session. The bugs coexisted because the contract was untested.

(A fourth gap, archive paths killing only the agent session, was classified as ambiguous between intentional UX and oversight; tracked separately as a follow-up issue.)

### What this changes

Three layers, evidence-based decisions, Oracle-reviewed.

**Layer 1**: New `pub(crate) fn kill_session_if_present(name: &str)` in `tmux/utils.rs`. Single source of truth for the `tmux kill-session -t` plus swallow `can't find session` pattern, factored out of three identical inline blocks. `Session::kill` keeps its `exists()` early-return fast-path; the inline kill+swallow block becomes one helper call. `PairedTerminal::kill` and `ToolSession::kill` migrate to the helper, gaining the missing race swallow as a side effect. Each caller still calls `refresh_session_cache()` post-kill at the session-lifecycle layer; the primitive does not.

**Layer 2**: New `pub fn kill_all_tmux_sessions(&self)` on `Instance` in `session/instance.rs`. Best-effort and silent at the user-facing layer; per-kind failures log at `debug!` target `session.tmux_cleanup` for forensics. Replaces the four-line scatter (`kill` + `kill_terminal` + `kill_container_terminal` + `kill_all_tool_sessions_for_id`) with one named operation.

**Layer 3**: Three callers migrate.
- `session/deletion.rs::perform_deletion` (CLI `aoe rm`, TUI delete, web `DELETE /api/sessions/:id`): four lines collapse to one helper call. Behavior unchanged.
- `tui/home/operations.rs::force_remove_session`: spawns the helper on a `std::thread::spawn` fire-and-forget so a hung tmux call cannot freeze the TUI input thread (the deletion poller exists for exactly this reason). Storage update proceeds immediately. Closes Gap #1.
- `server/api/cockpit.rs::cockpit_enable`: wraps the helper in `tokio::task::spawn_blocking` because four sequential subprocess waits on a tokio runtime worker would starve other handlers. The pre-existing `instance.kill()` warning trace is retired (the helper logs at `debug!` per kind). Closes Gap #4.

### UX note

After the cockpit mode-switch fix, switching from tmux to cockpit closes any open terminal panel for that session. Reopen via the panel UI after the switch. This converts a previously-silent tmux orphan into observable behavior matching the new contract.

### Drive-by

`src/cockpit/event_store.rs` had a pre-existing `repeat("?").take(N)` that started failing the Rust 1.94 `clippy::manual_repeat_n` lint and was blocking `-D warnings`. Migrated to `repeat_n("?", N)`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: a Playwright live spec for the cockpit mode-switch full teardown (asserting agent + terminal + tool tmux sessions all gone after `cockpit_enable`) is genuinely worthwhile but requires the ACP shim + cockpit harness setup. The helper invoked at that callsite is exercised by the e2e test below and the unit tests on the primitive, and the wiring at the callsite is three mechanical lines visible in the diff. A follow-up issue is filed to add the spec.

**TUI / CLI (e2e test)**

Tests added in this PR:
- `tests/e2e/cli.rs::test_cli_rm_kills_agent_tmux_session` — locks the invariant "agent tmux session is gone after `aoe rm`", which had no prior assertion despite being the cardinal contract. Passes on the parent commit (regression-prevention against a future re-ordering of `perform_deletion`).
- `src/tmux/utils.rs::tests::kill_session_if_present_swallows_missing_session` — asserts the primitive returns Ok when the session is already gone.
- `src/tmux/utils.rs::tests::kill_session_if_present_kills_existing_session` — asserts the primitive kills a real tmux session.

A follow-up issue is filed for an e2e test on the `force_remove_session` TUI flow (Status::Deleting injection + key-driven dialog + assert tmux gone).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 (via OpenCode/Sisyphus)

Methodology applied across the session:
1. Audit + RCA: 3 parallel explore agents mapped every removal path, every tmux kill helper, and existing test coverage. RCA classified each gap as intentional/non-intentional and traced root causes via git blame.
2. Codebase-grounded design research: 3 parallel explore agents on cleanup patterns, tmux module organization, and TDD conventions. Decisions sealed with file:line evidence rather than guessing.
3. Oracle review on the design: identified 4 BLOCKERS (preserve `exists()` fast-path, preserve `refresh_session_cache()`, wrap cockpit kill in `spawn_blocking`, run force_remove kill off-thread) plus recos. All blockers addressed before implementation.
4. Implementation: TDD-friendly (tests for the helper that fails on parent commit logic-wise), single-squash convention.

## Summary

This PR fixes incomplete tmux teardown that could leave orphaned tmux sessions or emit noisy warnings by centralizing tmux-kill behavior and applying it consistently on deletion, TUI force-remove, and cockpit-mode switch.

What changed (high level)
- Centralized tmux-kill into a helper that treats "can't find session" as success.
- Added an Instance-level best-effort teardown to kill agent, terminal, container-terminal, and tool sub-sessions.
- Replaced scattered direct tmux kills with the new helper/teardown at three call sites:
  - perform_deletion (CLI/web delete) — uses Instance::kill_all_tmux_sessions().
  - TUI force-remove — fires teardown on a background OS thread to avoid blocking the UI.
  - Cockpit enable — runs teardown in tokio::task::spawn_blocking to avoid worker starvation and logs task-panics.
- Tests: unit/integration tests for the new helper and an e2e CLI test verifying aoe rm removes the agent tmux session.
- Minor lint/placeholder change in EventStore (repeat_n).

Benefits
- Prevents orphaned tmux sessions and reduces noisy "can't find session" warnings.
- Avoids blocking UI/API workers during teardown.
- Centralizes teardown logic for easier maintenance and clearer, per-kind debug logging.

Technical notes (concise)
- New primitive: pub(crate) fn kill_session_if_present(name: &str) in src/tmux/utils.rs — runs `tmux kill-session -t <name>` with LC_ALL=C and treats stderr containing "can't find session" as success; other failures return Err.
- New abstraction: pub fn kill_all_tmux_sessions(&self) in src/session/instance.rs — best-effort sequence: kill() (agent), kill_terminal(), kill_container_terminal(), then crate::tmux::kill_all_tool_sessions_for_id(&self.id); failures logged at debug target session.tmux_cleanup.
- Call-site updates:
  - session/deletion.rs::perform_deletion — unified to Instance::kill_all_tmux_sessions().
  - tui/home/operations.rs::force_remove_session — spawns a detached OS thread (catch_unwind-wrapped) to call kill_all_tmux_sessions().
  - server/api/cockpit.rs::cockpit_enable — runs teardown inside spawn_blocking and logs join errors.
- Tests: tmux-guarded tests assert missing-session swallow and actual kill behavior; e2e test ensures aoe rm kills agent session.
- Small SQL placeholder change: use repeat_n in EventStore query placeholder construction.

Potential improvements
- Add higher-coverage tests for terminal/container-terminal teardown races.
- Consider retry/backoff or confirmation checks for critical-path cleanup.

### DIFF
--- a/src/server/api/cockpit.rs
+++ b/src/server/api/cockpit.rs
@@ -1282,10 +1282,11 @@ pub async fn cockpit_enable(
     }
 
     // Tear down the tmux side. Best-effort: a stale tmux name should
-    // not block the swap.
-    if let Err(e) = instance.kill() {
-        tracing::warn!(target: "cockpit.switch", session = %id, "kill tmux failed: {e}");
-    }
+    // not block the swap. Run on a blocking pool worker because each
+    // kill shells out and the four kinds add up to multiple subprocess
+    // waits we don't want on a tokio runtime thread.
+    let inst_for_kill = instance.clone();
+    let _ = tokio::task::spawn_blocking(move || inst_for_kill.kill_all_tmux_sessions()).await;
     instance.cockpit_mode = true;
 
     // Persist before spawning so a crash mid-swap leaves us in the
--- a/src/server/api/cockpit.rs
+++ b/src/server/api/cockpit.rs
@@ -1284,9 +1284,22 @@ pub async fn cockpit_enable(
     // Tear down the tmux side. Best-effort: a stale tmux name should
     // not block the swap. Run on a blocking pool worker because each
     // kill shells out and the four kinds add up to multiple subprocess
-    // waits we don't want on a tokio runtime thread.
+    // waits we don't want on a tokio runtime thread. Warn on the agent
+    // kill failure to keep user-visible signal for this user-initiated
+    // action; ancillary kinds (terminal, container terminal, tool sub-
+    // sessions) are silent and rely on `session.tmux_cleanup` debug
+    // tracing for forensics.
     let inst_for_kill = instance.clone();
-    let _ = tokio::task::spawn_blocking(move || inst_for_kill.kill_all_tmux_sessions()).await;
+    let id_for_log = id.clone();
+    let _ = tokio::task::spawn_blocking(move || {
+        if let Err(e) = inst_for_kill.kill() {
+            tracing::warn!(target: "cockpit.switch", session = %id_for_log, "kill tmux failed: {e}");
+        }
+        let _ = inst_for_kill.kill_terminal();
+        let _ = inst_for_kill.kill_container_terminal();
+        crate::tmux::kill_all_tool_sessions_for_id(&inst_for_kill.id);
+    })
+    .await;
     instance.cockpit_mode = true;
 
     // Persist before spawning so a crash mid-swap leaves us in the
--- a/src/server/api/cockpit.rs
+++ b/src/server/api/cockpit.rs
@@ -1283,12 +1283,9 @@ pub async fn cockpit_enable(
 
     // Tear down the tmux side. Best-effort: a stale tmux name should
     // not block the swap. Run on a blocking pool worker because each
-    // kill shells out and the four kinds add up to multiple subprocess
-    // waits we don't want on a tokio runtime thread. Warn on the agent
-    // kill failure to keep user-visible signal for this user-initiated
-    // action; ancillary kinds (terminal, container terminal, tool sub-
-    // sessions) are silent and rely on `session.tmux_cleanup` debug
-    // tracing for forensics.
+    // kill shells out. Warn on agent kill failure to keep signal for
+    // this user-initiated action; ancillary kinds rely on
+    // `session.tmux_cleanup` debug tracing.
     let inst_for_kill = instance.clone();
     let id_for_log = id.clone();
     let _ = tokio::task::spawn_blocking(move || {
--- a/src/server/api/cockpit.rs
+++ b/src/server/api/cockpit.rs
@@ -1284,19 +1284,21 @@ pub async fn cockpit_enable(
     // Tear down the tmux side. Best-effort: a stale tmux name should
     // not block the swap. Run on a blocking pool worker because each
     // kill shells out. Warn on agent kill failure to keep signal for
-    // this user-initiated action; ancillary kinds rely on
-    // `session.tmux_cleanup` debug tracing.
+    // this user-initiated action; ancillary kinds delegate to the
+    // shared helper so any future kind picked up by the audit lands
+    // here automatically.
     let inst_for_kill = instance.clone();
     let id_for_log = id.clone();
-    let _ = tokio::task::spawn_blocking(move || {
+    let kill_join = tokio::task::spawn_blocking(move || {
         if let Err(e) = inst_for_kill.kill() {
-            tracing::warn!(target: "cockpit.switch", session = %id_for_log, "kill tmux failed: {e}");
+            tracing::warn!(target: "cockpit.switch", session = %inst_for_kill.id, "kill tmux failed: {e}");
         }
-        let _ = inst_for_kill.kill_terminal();
-        let _ = inst_for_kill.kill_container_terminal();
-        crate::tmux::kill_all_tool_sessions_for_id(&inst_for_kill.id);
+        inst_for_kill.kill_ancillary_tmux_sessions();
     })
     .await;
+    if let Err(join_err) = kill_join {
+        tracing::error!(target: "cockpit.switch", session = %id_for_log, "tmux teardown task panicked: {join_err}");
+    }
     instance.cockpit_mode = true;
 
     // Persist before spawning so a crash mid-swap leaves us in the
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -3013,6 +3013,44 @@ impl Instance {
         Ok(())
     }
 
+    /// Kill every tmux session owned by this instance: the agent session,
+    /// the web terminal pane, the container terminal, and any tool
+    /// sub-sessions sharing this instance's id. Best-effort and silent at
+    /// the user-facing layer; per-kind failures are logged at `debug!`
+    /// level for forensics. Use this from any path that severs an
+    /// instance from its tmux footprint (deletion, force-remove, mode
+    /// switch) so no kind is forgotten.
+    pub fn kill_all_tmux_sessions(&self) {
+        if let Err(e) = self.kill() {
+            tracing::debug!(
+                target: "session.tmux_cleanup",
+                session_id = %self.id,
+                kind = "agent",
+                error = %e,
+                "kill_all_tmux_sessions: kill failed"
+            );
+        }
+        if let Err(e) = self.kill_terminal() {
+            tracing::debug!(
+                target: "session.tmux_cleanup",
+                session_id = %self.id,
+                kind = "terminal",
+                error = %e,
+                "kill_all_tmux_sessions: kill failed"
+            );
+        }
+        if let Err(e) = self.kill_container_terminal() {
+            tracing::debug!(
+                target: "session.tmux_cleanup",
+                session_id = %self.id,
+                kind = "container_terminal",
+                error = %e,
+                "kill_all_tmux_sessions: kill failed"
+            );
+        }
+        crate::tmux::kill_all_tool_sessions_for_id(&self.id);
+    }
+
     /// Stop the session: kill the tmux session and stop the Docker container
     /// (if sandboxed). The container is stopped but not removed, so it can be
     /// restarted on re-attach.
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -3016,10 +3016,12 @@ impl Instance {
     /// Kill every tmux session owned by this instance: the agent session,
     /// the web terminal pane, the container terminal, and any tool
     /// sub-sessions sharing this instance's id. Best-effort and silent at
-    /// the user-facing layer; per-kind failures are logged at `debug!`
-    /// level for forensics. Use this from any path that severs an
-    /// instance from its tmux footprint (deletion, force-remove, mode
-    /// switch) so no kind is forgotten.
+    /// the user-facing layer; agent, terminal, and container terminal
+    /// failures log at `debug!` (target `session.tmux_cleanup`) for
+    /// forensics. Tool sub-session failures are silent by design via
+    /// `kill_all_tool_sessions_for_id`. Use this from any path that
+    /// severs an instance from its tmux footprint (deletion, force-
+    /// remove, mode switch) so no kind is forgotten.
     pub fn kill_all_tmux_sessions(&self) {
         if let Err(e) = self.kill() {
             tracing::debug!(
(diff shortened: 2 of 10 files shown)


src/server/api/cockpit.rs
```diff
@@ -1282,10 +1282,11 @@ pub async fn cockpit_enable(
...
     // Tear down the tmux side. Best-effort: a stale tmux name should
-    // not block the swap.
-    if let Err(e) = instance.kill() {
-        tracing::warn!(target: "cockpit.switch", session = %id, "kill tmux failed: {e}");
-    }
+    // not block the swap. Run on a blocking pool worker because each
+    // kill shells out and the four kinds add up to multiple subprocess
+    // waits we don't want on a tokio runtime thread.
+    let inst_for_kill = instance.clone();
+    let _ = tokio::task::spawn_blocking(move || inst_for_kill.kill_all_tmux_sessions()).await;
```
jerome-benoit: Observability regression on this path. The old code logged the agent kill failure at `warn!` on target `cockpit.switch`; the new code routes everything through `kill_all_tmux_sessions`, which logs each kind at `debug!` on `session.tmux_cleanup`. At the default log level a user reporting "my terminal panel didn't disappear after switching to cockpit" now produces no signal — they'd need to opt into debug logging on a different target to see anything. The four-kind cleanup is the right move, but consider keeping the failure log at `warn!` here (or aggregating the per-kind debug logs into a single `warn!` if any of them failed), so cockpit-switch failures stay visible.

Severity: low — only matters when a tmux kill actually fails during the swap (rare), and the swap itself is unaffected because the call is best-effort.

src/session/instance.rs
```diff
@@ -3013,6 +3013,44 @@ impl Instance {
...
+        if let Err(e) = self.kill_container_terminal() {
+            tracing::debug!(
+                target: "session.tmux_cleanup",
+                session_id = %self.id,
+                kind = "container_terminal",
+                error = %e,
+                "kill_all_tmux_sessions: kill failed"
+            );
+        }
+        crate::tmux::kill_all_tool_sessions_for_id(&self.id);
```
jerome-benoit, line 3058: Minor doc/behavior mismatch: the doc-comment above (and the PR description) says "per-kind failures are logged at `debug!` level for forensics", but `kill_all_tool_sessions_for_id` is silent on per-session failures — its inner `Command::new("tmux") … .output()` is bound to `let _ =` and there is no log on failure. So three of the four kinds log at `debug!`, the fourth doesn't. Either tighten the doc to say "the agent / terminal / container-terminal kinds log per-kind failures; tool sessions are silent best-effort", or add a matching `tracing::debug!` inside `kill_all_tool_sessions_for_id` for symmetry.

Severity: very low — documentation accuracy.

src/tui/home/operations.rs
```diff
@@ -458,10 +458,17 @@ impl HomeView {
...
+    /// Force-remove a session from storage. Worktree, branch, and
+    /// container cleanup are skipped (the original deletion already
+    /// attempted them); tmux teardown is fired off-thread so a hung
+    /// tmux call cannot block the storage update on the TUI input
+    /// thread. Used for sessions stuck in the Deleting state where
+    /// the background deletion thread never returned a result.
     pub(super) fn force_remove_session(&mut self, session_id: &str) -> anyhow::Result<()> {
+        if let Some(inst) = self.instances.iter().find(|i| i.id == session_id) {
+            let inst = inst.clone();
+            std::thread::spawn(move || inst.kill_all_tmux_sessions());
```
jerome-benoit: Two small observations on the fire-and-forget thread:

1. The `JoinHandle` is dropped immediately and the closure isn't `catch_unwind`-wrapped. If `kill_all_tmux_sessions` (or anything it calls — e.g. the `stop_poller` mutex path or the subprocess spawn) panics, the panic is lost without any log. Wrapping the closure in `std::panic::catch_unwind` and emitting a `tracing::warn!` on `Err` would keep the responsiveness benefit while preserving forensics.

2. Worth knowing for the documented tradeoff: this only races process exit if the user quits aoe within milliseconds of pressing the force-remove key. The four `tmux kill-session` invocations are short-lived child processes; if the parent dies first, the in-flight subprocess may be terminated mid-handshake on the tmux socket and the session stays orphaned. The deletion poller pattern justifies the off-thread move; logging panics is the cheap upside left on the table.

Severity: very low — the common path works.

src/tmux/utils.rs
```diff
@@ -252,6 +253,24 @@ pub fn tmux_prefix_display() -> &'static str {
     })
 }
 
+/// Run `tmux kill-session -t <name>`. The `can't find session` stderr is
+/// treated as success: callers commonly kill the pane's process tree
+/// first, which can cause tmux to tear down the session before this call
+/// lands. Any other tmux failure returns `Err`. Caller is responsible for
+/// `refresh_session_cache` after a successful kill.
+pub(crate) fn kill_session_if_present(name: &str) -> Result<()> {
```
jerome-benoit, line 261: The "missing session" detection is matched against tmux's English stderr text. This is robust enough in practice (tmux is not localized in the typical install) but worth noting: a tmux build with a translated stderr would make `kill_session_if_present` mis-classify a real failure as "session vanished, all good". An alternative is to call `has-session -t <name>` first and short-circuit on failure, which is what `Session::kill` and `PairedTerminal::kill` already do via their own `exists()` check before reaching this helper. So the locale concern only bites on the inline race window the helper exists to swallow, which is exactly the case where the english-string match has to hold.

Severity: very low — accepting this as a pragmatic tradeoff is fine; just calling out the assumption.

jerome-benoit (commented):
Solid PR. The audit-then-factor approach is appropriate, the three-layer split (`kill_session_if_present` primitive → `kill_all_tmux_sessions` per-instance helper → three callsites) is the right shape, and the new e2e test locks the cardinal "session removed implies tmux gone" invariant that was missing.

No blocking bugs found. The changes preserve all the behavioral subtleties that matter:

- `Session::kill` keeps its `exists()` fast-path and continues to swallow the "can't find session" race the same way the inline block did.
- `PairedTerminal::kill` and `ToolSession::kill` correctly gain the missing race-swallow as a side effect of routing through the helper — this is a real fix, not a refactor regression.
- `cockpit_enable` correctly wraps the now-multi-subprocess kill in `spawn_blocking` so it doesn't block a tokio runtime worker.
- `force_remove_session` correctly fires the kill off-thread so a hung tmux call can't freeze TUI input.
- `perform_deletion` is already invoked under `spawn_blocking` from the web handler, so the inline call is fine there.

Comments are minor: an observability regression on the cockpit-switch path (warn → debug), a doc/behavior mismatch on tool-session logging, a missed `catch_unwind` on the fire-and-forget thread, and a note on the locale assumption baked into the stderr match. None block merge.

src/server/api/cockpit.rs
jerome-benoit: Addressed across `cc0dfc25` plus the round-2 follow-up: the cockpit handler now calls `instance.kill()` directly with the original `tracing::warn!(target: "cockpit.switch")` for the agent kill, then delegates the three ancillary kinds to the new `Instance::kill_ancillary_tmux_sessions()`. The shared helper keeps the audit's "no kind forgotten" guarantee for any future tmux kind added to the ancillary list, while preserving the user-visible `cockpit.switch` warn signal flagged here.

src/session/instance.rs
jerome-benoit: Addressed in `687ef014` (initial doc clarification), refined in `6f54ba4d` (tightening), and finalized in `cc0dfc25` (helper split). Both `kill_all_tmux_sessions` and `kill_ancillary_tmux_sessions` docstrings now explicitly call out: "Tool sub-sessions are silent by design via `kill_all_tool_sessions_for_id`." The asymmetry is documented as intentional rather than left ambiguous.

src/tui/home/operations.rs
jerome-benoit: Addressed in `687ef014`. The fire-and-forget thread is now wrapped in `std::panic::catch_unwind(std::panic::AssertUnwindSafe(...))`; on panic, a `tracing::error!` event is emitted on `session.tmux_cleanup` with the session id and panic payload. A panic in any kill helper surfaces in the log instead of being lost to the detached thread.

src/tmux/utils.rs
jerome-benoit: Addressed in `687ef014`. `kill_session_if_present` now calls `.env("LC_ALL", "C")` on the `Command` so tmux's stderr is independent of the user's `LC_*` settings. The "can't find session" match is locale-stable; the doc comment notes "(in C locale)" inline.

jerome-benoit:
CodeRabbit summary nitpicks (review 4412374490) status:

- **`tests/e2e/cli.rs:867-933`** — defensive tmux cleanup: addressed in `55e8db2c`. Mirrors `test_cli_rename_preserves_tmux_session` and aligns with AGENTS.md.
- **`src/tui/home/operations.rs:468-471`** — `Builder::new()` named thread + spawn-error log: declined. The realistic failure mode (panic inside the closure) is already captured by `catch_unwind` + `tracing::error!` on `session.tmux_cleanup`. OS thread-spawn failure is a resource-exhaustion edge case where naming the thread for diagnostics adds little forensic value over the panic path that already exists.
- **`src/server/api/cockpit.rs:1288-1289`** — log `spawn_blocking` `JoinError`: already addressed in `cc0dfc25`. The current code captures `kill_join` and emits `tracing::error!(target: "cockpit.switch", session = %id_for_log, "tmux teardown task panicked: {join_err}")`.

njbrake (approved):

PR #1947, opened 2026-06-04
fix(tmux): lower paste-buffer threshold so short pastes use bracketed paste

BTForIT:
## Description

`Session::send_keys_with_delay` currently routes anything under 2048 bytes that has no embedded newline through `tmux send-keys -l --`, which delivers raw keystrokes with no bracketed-paste framing. For longer or multi-line payloads it uses the `tmux paste-buffer -p` path that wraps the content in DECSET 2004 markers. This PR drops the byte threshold from 2048 to 16 so single-line dictation-sized payloads also take the bracketed-paste path.

### Symptom this fixes

Reproducer on Mosh (which strips bracketed paste end-to-end, so the AOE TUI sees external pastes as keystroke bursts and that's what the send-message dialog forwards to the target pane):

1. Use a dictation tool (e.g. VoiceInk) to paste a ~365-byte single-line prompt into AOE's send-message dialog.
2. Submit.
3. AOE's worker calls `send_keys_with_delay(text, 0)`.
4. With the old 2048 threshold, the text takes the `tmux send-keys -l --` path: claude-code's pane receives 365 raw keystrokes.
5. AOE then immediately issues `tmux send-keys Enter` (delay = 0ms for claude).
6. claude-code's input-burst detector treats the trailing Enter as inside the keystroke burst and inserts a newline instead of submitting.
7. The text renders in claude's input box; the message never sends.

The user sees: "dictated prompt arrived, hit enter, nothing happened, enter just made a newline."

Switching the same payload to the paste-buffer path frames it as a paste; the trailing Enter then reliably submits. claude-code, codex, opencode, gemini all enable DECSET 2004 (already documented in `send_via_paste_buffer`'s contract comment), so the receiving side already handles this correctly. See google-gemini/gemini-cli#26114 for independent confirmation that claude-code handles paste only with bracketed-paste markers present.

### Why 16

The threshold needs to be:
- Large enough that single-token shortcuts (`ls`, `q`, `y`, a few-char control responses) still take the fast `send-keys -l` path with no extra tmux subprocess spawn.
- Small enough that anything plausibly arriving from a dictation buffer, prose paste, or a typed-in command of normal length flows through bracketed paste.

16 bytes covers the common short-token cases (a CLI command + a flag fits) while still routing typical dictation, multi-word commands, and any prose through bracketed paste. Short tokens are not at risk of the burst-window failure mode anyway because the receiving agent's burst detector needs ≥3 rapid keystrokes to engage.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`cargo test --release --lib tmux::` 186/186)
- [ ] For UI changes: included screenshot or recording (N/A; transport-layer change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the failure mode requires Mosh in the chain to strip bracketed paste from external pastes; this isn't reproducible in the e2e harness which runs against a local tmux without a Mosh hop. The change itself is a single-line constant adjustment to an already-covered code path; the existing `tmux::session` unit tests pass.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Claude Code)

**Any Additional AI Details you'd like to share:** Root-cause analysis was driven by inspecting `~/.agent-of-empires/debug.log` correlated to a reproduced VoiceInk-via-Mosh failure, and cross-referenced with google-gemini/gemini-cli#26114 which independently documents that claude-code handles paste correctly only with bracketed-paste markers. The fix itself is a one-line threshold change.

## Summary of Changes

The PR modifies how tmux sends text input to terminal sessions by lowering the threshold for when paste-buffer delivery is used. Specifically:

**What Changed:**
- The `PASTE_BYTE_THRESHOLD` constant in `src/tmux/session.rs` was reduced from 2048 bytes to 16 bytes
- Updated accompanying comments to explain the new behavior

**What This Means:**
Messages are now routed through two different delivery paths based on size:
- **Very short inputs** (under 16 bytes, like `ls` or single commands): sent directly as raw keystrokes via `tmux send-keys -l`
- **Typical inputs** (16+ bytes or containing newlines, like dictation, multi-word commands, or prose): sent via `tmux paste-buffer` which wraps content with bracketed-paste markers

**Why This Matters:**
In environments using intermediate layers like Mosh that strip bracketed-paste markers, longer single-line pastes were previously being delivered as raw keystroke bursts followed by Enter. Receiving systems would treat the trailing Enter as part of the keystroke burst rather than as a submit command, causing text to be rendered but not submitted. 

By using bracketed-paste framing for typical inputs, the trailing Enter is now properly recognized as a separate command, fixing this delivery issue while still maintaining fast delivery for very short, single-token inputs.

**Technical Details:**
- The behavior change affects the `send_keys_with_delay` function's delivery mechanism selection logic
- The new constant `const PASTE_BYTE_THRESHOLD: usize = 16` with the condition `let use_paste_buffer = byte_len >= PASTE_BYTE_THRESHOLD || text.contains('\n')`
- All existing tests (186/186) pass with this change; no UI modifications required
- One-line constant adjustment with expanded documentation of the rationale


njbrake (approved):

Approving. Verified the one path that actually changes for every user, not just Mosh: lowering the threshold to 16 makes most single-line sends go out as a bracketed paste instead of raw keystrokes, so the question was whether agents mishandle a pasted slash command.

Tested live by reproducing the exact send sequence (load-buffer, paste-buffer -d -p, then a separate send-keys Enter) against real claude and codex panes:

- claude: pasted `/help` opened the Help UI; pasted `/output-style default` (21 bytes) was routed to the command parser (returned "Unknown command", i.e. parsed as a command, not sent to the model as prose); a normal 54-byte pasted message submitted a real turn and got a reply.
- codex: pasted `/diff` opened the diff viewer.

So both burst-detector agents recognize a pasted slash command as a command. Single-line pastes also rendered inline (no collapsed paste placeholder). Combined with `paste-buffer -p` only emitting markers when the pane requested bracketed paste (so non-bracketed agents stay byte-identical to the old path), this is a reliability improvement with no behavior change for normal local users.

Issue #1942, opened 2026-06-03
Input issues on "live" tab-selected sessions


jcogilvie:
In the context of a session selected by clicking on it in the left pane, or pressing tab to select it, there are two things that don't work:  

* for some reason, the semicolon key doesn't get passed through if I type it
* shift+insert to multiline paste from the clipboard sends actual "enter" keypresses instead of new lines, breaking up pasted text into separate commands

Both of these things work fine when I actually attach to the session instead of click it or tab into it.

Using macos with zsh.

njbrake:
Thanks for the report! Which version of AoE are you using?

njbrake:
I just verified the bug on the latest from main. Fixing now, thank you!

njbrake:
> _Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Closing this out after a thorough investigation. Where each half landed:

**Semicolon in live mode (fixed).** Real bug: tmux's command parser drops a trailing `;` from a `send-keys -l` payload, so a typed semicolon never reached the pane. Fixed in #1949.

**Multiline paste via shift+Insert (not reproducible; closing with explanation).** I could not reproduce this on a current macOS setup, and the standard paste path works correctly:

- **Cmd+V works.** In live mode, a multi-line Cmd+V arrives as a bracketed paste (`ESC[200~ ... ESC[201~`), aoe receives it as one `Event::Paste`, and it lands in the agent as a single block. Verified end to end.
- **shift+Insert does not paste on a stock Mac.** On the built-in keyboard the combo sends a bare carriage return, not a paste. shift+Insert as a paste action is a Linux/Windows convention that most macOS terminals don't ship with.
- **The original symptom only occurs with an *unbracketed* paste.** A bracketed shift+Insert would behave like the working Cmd+V. If a non-default keyboard or terminal binds shift+Insert to a paste that isn't wrapped in bracketed-paste markers, the clipboard arrives as raw keystrokes, and interior newlines are bare carriage returns that submit one line at a time.

[...]

PR #1954, opened 2026-06-04
fix(tui): coalesce jittery unbracketed pastes in live mode

njbrake:
> **Draft.** The fix is a timing heuristic whose correct tuning depends on real-terminal paste delivery. Holding in draft until dogfooded on a real macOS terminal with shift+Insert (see Test plan). Second half of #1942; #1949 fixed the semicolon half.

## Description

In live-send mode, a multiline **shift+Insert** paste posted one command per line instead of arriving as a single block.

Root cause: aoe enables bracketed paste, and the `Event::Paste` path already wraps multiline pastes correctly (#1553). But some terminals deliver **shift+Insert** as raw key events rather than a bracketed `Event::Paste`. For that unbracketed stream, the paste-burst detector (originally built for Mosh, which also strips bracketed-paste markers) is what coalesces the keys into a single bracketed paste before forwarding. Its inter-key window is a tight 5ms, which fragments when the unbracketed stream arrives in jittery chunks whose gaps exceed 5ms; each interior `Enter` then leaks through as an individual `submit`, splitting the paste into separate commands.

Fix: once a burst is confirmed (two or more events arrived within the tight 5ms window, which only a paste produces, since no human presses two keys 5ms apart), widen the inter-key window to 50ms for the rest of the burst. This absorbs delivery jitter without taxing single keystrokes, which never reach the two-event threshold and so keep the snappy 5ms window.

### Why a heuristic (and why it can't be eliminated)

For an *unbracketed* key-stream paste there is no marker distinguishing a pasted `Enter` from a typed one; timing is the only available signal. The `Event::Paste` path (Cmd+V on terminals that bracket it) is unaffected and already correct. This change only hardens the fallback the burst detector provides.

### Measured envelope

Verified with a scripted live-send session against a stdin logger, injecting a jittered multiline key stream and inspecting the forwarded bytes:

| Inter-chunk gap | Before (main) | After |
| --- | --- | --- |
| 2ms (tight) | bracketed | bracketed |
| 20ms | raw CRs (one submit/line) | **bracketed** |
| 40ms | raw CRs | **bracketed** |
| 70ms (beyond window) | raw CRs | raw CRs (still fragments) |

So the fix reliably coalesces pastes whose chunks land within ~50ms of each other. A local shift+Insert paste is near-instant, so this should cover it; the open question the draft is gating on is whether any real terminal delivers chunks wider than 50ms.

Behavior that is provably unchanged: single keystrokes (still 5ms, never establish a burst), home-view shortcuts, and the `Event::Paste` path.

## PR Type
- [x] Bug Fix

## Checklist
- [x] Documentation was updated where necessary (N/A)
- [x] For UI changes: included screenshot or recording (N/A; no visual change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is async event-delivery timing (the burst window). A meaningful e2e would have to inject a key stream with gaps between 5ms and 50ms and assert the forwarded bytes are bracketed; that is inherently timing-sensitive and would flake on loaded CI runners (a slow runner could push even tight injection past 50ms, or deliver deliberate 20ms gaps as >50ms). Instead there is a unit test for the gap-selection logic (`paste_burst_gap_ms`), and the end-to-end behavior is validated by a scripted live-send reproduction (envelope table above) plus manual dogfooding on a real terminal.

## Test plan (dogfood before un-drafting)

On a real macOS terminal:

1. Build this branch and run `aoe`.
2. Tab into a running session to enter live mode (the `● LIVE` footer).
3. Copy a multiline block (3+ lines) and paste with **shift+Insert**.
4. Expect: the whole block lands in the agent's input as one paste, not one command per line.
5. Also sanity-check: typing a single Enter still submits; fast typing is unaffected; Cmd+V still works.

If it still fragments, run with `AGENT_OF_EMPIRES_DEBUG=1` and check `debug.log` for the `paste-burst: routed N chars` trace: its presence means the burst coalesced; its absence means the stream fragmented (chunks arriving >50ms apart), which would tell us 50ms needs raising.

## AI Usage
- [x] AI was used

**Any Additional AI Details you'd like to share:** Root-caused, implemented, and characterized with Claude Code; reasoning and decisions reviewed by @njbrake.

Refs #1942

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -49,6 +49,31 @@ const PASTE_BURST_INTER_KEY_MS: u64 = 5;
 /// individual key events so genuine typing isn't mistaken for a paste.
 const PASTE_BURST_MIN_LEN: usize = 3;
 
+/// Wider inter-key timeout used once a burst is established (two events
+/// already arrived within `PASTE_BURST_INTER_KEY_MS` of each other, which
+/// only a paste does). The tight 5ms window catches a paste that streams in
+/// without gaps, but an unbracketed paste delivered as raw key events (e.g.
+/// shift+Insert on terminals that don't bracket it, #1942) can arrive in
+/// jittery chunks whose gaps exceed 5ms, fragmenting the burst so interior
+/// Enters leak through as individual `submit` keystrokes. Widening the
+/// window after the burst is confirmed absorbs that jitter without taxing
+/// single keystrokes, which never reach the two-event threshold and so keep
+/// the tight window.
+const PASTE_BURST_ESTABLISHED_GAP_MS: u64 = 50;
+
+/// Inter-key timeout (ms) for the next poll while accumulating a paste
+/// burst, given how many events are already in the burst. The first
+/// follow-up poll (`len` == 1) uses the tight window so a single keystroke
+/// stays snappy; once a second event has joined (`len` >= 2, which only a
+/// paste produces) the window widens to absorb delivery jitter (#1942).
+fn paste_burst_gap_ms(burst_len: usize) -> u64 {
+    if burst_len >= 2 {
+        PASTE_BURST_ESTABLISHED_GAP_MS
+    } else {
+        PASTE_BURST_INTER_KEY_MS
+    }
+}
+
 struct UpdateStatus {
     text: String,
     expires_at: Option<std::time::Instant>,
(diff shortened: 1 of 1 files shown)


njbrake:
> _Note: this comment was drafted by Claude via back-and-forth with @njbrake._

Closing without merging. Investigation on #1942 concluded that the multiline-paste symptom only occurs with an unbracketed paste (shift+Insert bound to a non-bracketing paste), and the standard Cmd+V path already works correctly in live mode. The burst-widening heuristic here is the wrong approach: terminal multiplexers rely on bracketed-paste markers rather than reconstructing unmarked pastes, and a timing window risks misclassifying fast typing. Full explanation in #1942.

PR #1958, opened 2026-06-04
fix(logging): silence no-op filter swaps to break file-watch OOM loop

### LINKED ISSUE #1894: Daemon log-filter file-watch self-triggers an infinite loop → unbounded memory growth → OOM
# Daemon log-filter file-watch self-triggers an infinite loop → unbounded memory growth → OOM

## Summary

The `serve` daemon (and `__cockpit-runner` workers) watch the logging filter and **re-apply it on every log write**, emitting an INFO line `runner filter swapped … source="file-watch"`. Because that message is itself written to the watched log, **the write re-triggers the watch**, producing a self-sustaining loop at **9,000–56,000 iterations/second**. Each iteration enqueues an event faster than it can drain, so an internal buffer grows without bound until the kernel OOM-kills the process (observed at **13.4 GB anon-rss**), taking the agent session down with it.

Notably, `previous == current` in every one of these log lines — the filter never actually changes, so the "swap" (and the log line it emits) is a pure no-op that should not happen at all.

## Environment

- **aoe**: 1.9.5
- **OS**: Windows 11 Pro 26200, WSL2, kernel `6.6.114.1-microsoft-standard-WSL2`
- **Logging config** (default): `default_level = "info"`, `output = "file"`, `file_path = "debug.log"`, `rotation = "size"`, `max_size_mib = 50`
- Cockpit in use with multiple concurrent ACP sessions (`claude-agent-acp`)

## Symptoms

- Agent sessions die intermittently with `[server exited]` / `[process exited with code 1]`.
- `debug.log` grows at multiple MB/sec and rotates constantly; **~98% of all log volume is the single message below**.
- One `aoe` process balloons into the multi-GB range and is OOM-killed.

## Evidence

The flooding log line (identical `previous` and `current`):

```
2026-06-03T03:55:54.781229Z  INFO log.runtime: runner filter swapped \
  previous=agent_of_empires=info,cockpit=info,terminal=info,...,sound=info \
  current=agent_of_empires=info,cockpit=info,terminal=info,...,sound=info \
  source="file-watch"
```

Rate / volume measured on a single log file:

```
total lines = 543,783   |   "runner filter swapped" = 533,734  (~98%)
peak rate: 56,224 occurrences in ONE second (single second bucket)
sustained ~9,600 lines/sec with only the serve daemon running (no active runner)
```

Kernel OOM (WSL2):

```
tokio-rt-worker invoked oom-killer: gfp_mask=0x141cca(...), order=0
Free swap  = 0kB
Total swap = 16777216kB
Out of memory: Killed process 6979 (aoe) \
  total-vm:19353304kB, anon-rss:13449088kB, ... oom_score_adj:0
oom_reaper: reaped process 6979 (aoe)
```

(All ~16 GB of VM swap plus RAM was exhausted; `tokio-rt-worker` — the runtime spinning the loop — invoked the killer.)

## Root cause (confirmed)

1. The daemon watches the log filter source via a file-watch (`source="file-watch"`).
2. On every trigger it re-applies the filter and logs `runner filter swapped` at **INFO**.
3. That INFO line is written to the log file the watcher is keyed on.
4. The write fires the watcher again → goto 2. Self-sustaining.
5. Events enqueue faster than they drain → unbounded internal buffer → OOM.

**Proof of mechanism:** raising the runtime level above INFO suppresses the message and stops the loop instantly:

```console
$ aoe log-level warn
{"previous":"...=info,...","current":"...=warn,...","ephemeral":true}
# log growth: +0 lines in 4s (was ~38,000 lines / 4s)
```

Setting `default_level = "warn"` in `config.toml` makes a cold-booted daemon loop-free from the start (verified: +0 lines in 5s on fresh start). Relocating `file_path` out of the config dir does **not** help — the watch follows the log file wherever it points.

## Suggested fixes

- **Don't log `runner filter swapped` when `previous == current`** (no-op swaps should be silent) — this alone breaks the feedback loop.
- Emit that message at `debug`/`trace`, not `info`, and/or
- Exclude the log file (or its directory) from the filter file-watch, and/or debounce the watcher so a write storm collapses into a single reload.

## Workaround for other users

Set `default_level = "warn"` in `config.toml` (or `aoe log-level warn` at runtime, which is ephemeral). Use `aoe log-level debug` only briefly when you need verbose logs, since `info`+ re-arms the loop.

Seluj78:

## Description

The daemon and structured-view runners re-apply the logging filter on every `runtime_filter` watch event and log `filter swapped` at INFO, even when the new directive is identical to the active one (a pure no-op swap). On older builds that INFO line landed in the watched app dir and re-triggered the watcher, the self-sustaining feedback loop reported in #1894 that grew unbounded until the daemon was OOM-killed (13.4 GB anon-rss).

The catastrophic loop was already mitigated on `main` by #1734, which narrowed the watch to `FileMatcher::Exact(runtime_filter)` so `debug.log` writes no longer match the watch. That matcher is a heuristic on OS-emitted paths, though, not a guarantee: coarse-grained or duplicate kernel events can still reach the callback. This PR removes the remaining loop fuel and makes the invariant platform-independent.

`FilterController::swap` now compares the requested directive against the active one while holding the `current` lock, skips the reload-handle modify on a no-op, and reports `changed = false`. The three swap sites (file-watch, settings, REST) gate their INFO log on `changed`, and the settings/REST paths also skip the `persist_runtime_filter` disk write on a no-op so an unchanged directive cannot re-fire every runner's watcher. The result: a no-op swap is fully silent and writes nothing into the watched dir, regardless of log level. The `Unavailable` (TUI, no controller) branch still persists, and the REST response contract (`previous`/`current`/`ephemeral`) is unchanged.

Closes #1894

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start the daemon at the default `info` level, exercise a flow that produces logs, and confirm `debug.log` does not fill with repeated `filter swapped ... source="file-watch"` lines.
- [ ] Run `aoe log-level info` while the daemon is already at `info`; confirm no `filter swapped` line is emitted and `runtime_filter` is not rewritten (mtime unchanged).
- [ ] Run `aoe log-level warn` (a genuine change); confirm exactly one `filter swapped ... source="rest"` line is logged and `runtime_filter` is updated.
- [ ] Save settings in the TUI / web without changing the logging level; confirm no `filter swapped ... source="settings"` line and no `runtime_filter` rewrite.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Covered by a unit test in `src/logging.rs` (`swap_reports_changed_then_noop`): a swap to a new directive reports `changed = true`; an identical re-apply reports `changed = false` with `previous == current`. This is the no-op detection that keeps the swap silent.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the design calls (no-op silence over debounce, matcher hardening deferred to a follow-up), and ran the validation.

**NOTE:**

## Overview
This PR stops no-op logging filter re-applies from producing side effects that caused a file-watch feedback loop and OOM. Identical runtime filter directives are now treated as silent no-ops: they neither emit the INFO "filter swapped" message nor rewrite the runtime_filter file.

## What changed (high-level)
- Detect no-op swaps and avoid any logging or disk writes when the requested directive equals the active one.
- Make the file-watch, settings, and REST codepaths emit the "filter swapped" INFO event and persist runtime_filter only when an actual change occurs.
- Add a unit test covering change vs no-op behavior.
- Documentation updated to note runtime_filter no-ops are silent.

## Benefits
- Eliminates the self-triggering watcher loop and the resulting unbounded memory growth / OOM.
- Platform-independent and robust to coarse-grained or duplicate OS file events.
- Preserves behavior for real filter changes (single notification + persistence).
- Prevents unnecessary disk writes and log noise.

## Technical details (concise)
- FilterController::swap now holds the current mutex while comparing directives, performs the reload modification only when different, and returns SwapResult { previous, current, changed: bool }.
- Call sites check swap.changed; file-watch and REST/settings only log/persist when true. The TUI/no-controller branch still persists.
- New unit test: swap_reports_changed_then_noop verifies changed=true on real change and changed=false on identical re-apply.

Closes `#1894`

### DIFF
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -29,7 +29,7 @@ Only a small set of sub-targets is enumerated in the settings dropdown (see `KNO
 | `hooks` | Agent hook integration (Claude/Settl/Hermes/Kiro) install/uninstall, hook status file lifecycle, user-configured status hook command failures, and attached status hook watcher failures. |
 | `sound` | Notification sound asset download/install and per-event playback. |
 | `telemetry` | Anonymous opt-in usage telemetry: install-id persistence, opt-in/opt-out transitions, and send outcomes (all at `debug`). The `aoe serve` consent route emits under `http.api.telemetry`. |
-| `log.runtime` | Filter swaps (REST + runner file-watch). |
+| `log.runtime` | Filter swaps (REST + runner file-watch). A swap whose new directive equals the active one is a silent no-op: it does not log or rewrite `runtime_filter`, so a watcher write cannot re-trigger itself. |
 
 ## Levels
 
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -212,14 +212,19 @@ pub fn apply_persisted_config(
     };
     match set_filter(&filter) {
         Ok(swap) => {
-            tracing::info!(
-                target: "log.runtime",
-                previous = %swap.previous,
-                current = %swap.current,
-                source = "settings",
-                "filter swapped"
-            );
-            persist_runtime_filter(&swap.current, app_dir);
+            // Skip the log and the disk write on a no-op: persisting an
+            // unchanged directive needlessly re-fires every runner's watcher
+            // (#1894).
+            if swap.changed {
+                tracing::info!(
+                    target: "log.runtime",
+                    previous = %swap.previous,
+                    current = %swap.current,
+                    source = "settings",
+                    "filter swapped"
+                );
+                persist_runtime_filter(&swap.current, app_dir);
+            }
         }
         Err(LogFilterError::Unavailable) => {
             // No reload handle installed (e.g. TUI process). Still persist
@@ -424,14 +429,30 @@ impl FilterController {
     }
 
     fn swap(&self, filter: EnvFilter, directive: String) -> Result<SwapResult, LogFilterError> {
-        let previous = self.current();
+        // Hold the `current` lock across the compare and the modify so two
+        // concurrent swaps cannot interleave and both observe `changed`.
+        let mut current = self.current.lock().unwrap();
+        let previous = current.clone();
+        // No-op: the active directive already equals the requested one.
+        // Skip the reload-handle modify entirely and report `changed=false`
+        // so callers stay silent. A no-op swap that logged at INFO is what
+        // fed the file-watch OOM loop in #1894: the log line landed in the
+        // watched dir and re-triggered the watcher.
+        if previous == directive {
+            return Ok(SwapResult {
+                previous,
+                current: directive,
+                changed: false,
+            });
+        }
         self.inner
             .modify(|f| *f = filter)
             .map_err(|e| LogFilterError::Invalid(e.to_string()))?;
-        *self.current.lock().unwrap() = directive.clone();
+        *current = directive.clone();
         Ok(SwapResult {
             previous,
             current: directive,
+            changed: true,
         })
     }
 }
@@ -440,6 +461,10 @@ impl FilterController {
 pub struct SwapResult {
     pub previous: String,
     pub current: String,
+    /// `false` when the requested directive already matched the active one,
+    /// so the swap was a no-op. Callers gate logging and persistence on this
+    /// to avoid the self-sustaining file-watch loop (#1894).
+    pub changed: bool,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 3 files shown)


Seluj78:

@sporebattyl this PR addresses #1894. Two things worth flagging on the repro:

1. The catastrophic OOM loop you hit on v1.9.5 was already mitigated on `main` by #1734 (merged after 1.9.5), which narrowed the filter watch to an exact match on `runtime_filter` so `debug.log` writes no longer re-trigger it. This PR removes the remaining loop fuel: no-op filter swaps (`previous == current`) are now fully silent and write nothing into the watched dir, regardless of log level, so the feedback loop has no fuel even if a coarse-grained or duplicate kernel event slips past the matcher.

2. Could you build this branch and try to reproduce on your WSL2 setup? Specifically: run the daemon at `info`, drive some structured-view traffic, and confirm `debug.log` does not flood with `filter swapped ... source="file-watch"` lines and memory stays flat. Your environment is the one that surfaced this, so a confirmation there is worth more than my local check. Thanks.

PR #1980, opened 2026-06-04
feat(tui): anchor preview selection to scrollback so it spans pages

### LINKED ISSUE #1961: Is there anyway to select more than one page?
Select will exit, if I scroll. So how can I select more than one page?

njbrake:
## Description

Fixes #1961. In live mode you couldn't copy more than one visible page of agent output: the text selection cancelled the moment you scrolled, and even without that the copy only captured the current page.

The selection is now anchored to the scrollback rather than screen cells, and edge-drag auto-scrolls so a single drag can grab many pages.

What changed (each commit fixes something the previous one surfaced during real-world testing):

1. **Scrollback-anchored selection + full copy.** A render-time snapshot (`PreviewTextView`: output pane rect, top visible line, total lines) maps screen cells to content lines and back between frames. Scrolling keeps the selection, the highlight tracks its text, and the copy reads the full parsed scrollback (each line re-laid into a one-row buffer for correct wide-char and truncation handling), including lines scrolled off screen.
2. **Ticker-driven edge auto-scroll.** Dragging to the pane's top/bottom edge scrolls continuously off the event-loop ticker instead of per mouse event, so holding the cursor still at the edge keeps extending the selection (crossterm only emits drag events on movement).
3. **No screen strobing.** The auto-scroll requests a diffed redraw instead of `needs_redraw` (which forces a full `terminal.clear()` every frame and made the screen flash). Also paces the scroll to a steady ~25 lines/s so it doesn't race with capture-worker wakes.
4. **Correct copy on scroll-up.** The selection now anchors to a distance from the newest line, not an absolute index. In live mode the captured window grows from the top as you scroll back, so an absolute index silently drifted to an older line, the copy started from the wrong place. Distance-from-bottom is invariant under that growth, so the anchor stays locked to the line you pressed on.

Rust-only change in the native TUI; no `web/` or dashboard flows touched.

### How it was tested

- `cargo fmt`, `cargo clippy --features serve`, `cargo test --features serve` all clean.
- Rewrote the in-module `preview_drag_select` unit tests for the content-anchored model and extended them: multi-page copy across a scroll, top/bottom-edge auto-scroll via ticks, off-screen highlight clipping, the empty-pane no-op, and a regression test that grows the capture window mid-selection and asserts the copy stays locked to the same physical lines.
- Verified end to end by driving the real `aoe` binary in tmux against a 400-line pane with injected SGR mouse events: entered live mode, pressed near the tail, dragged to the top edge and held to scroll up ~2 pages, released. Clipboard got a clean contiguous range (LINE348..LINE396, 49 lines) including rows that had scrolled off the bottom, with no blank frames across rapid captures.

### Known limitations

- The copy lands in the system clipboard via OSC 52, so it depends on the terminal honoring OSC 52.
- The selectable range is bounded by tmux's `capture-pane` window.
- While an agent is actively streaming new output, the bottom reference moves, so selecting a moving target can drift; selecting settled output (the reported use case) is stable.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Implemented by Claude Code via back-and-forth with @njbrake; the design decisions (scrollback-anchored selection, distance-from-bottom anchoring, ticker-driven edge auto-scroll) are his.

Fixes #1961

## Overview

This PR fixes issue `#1961` by enabling text selection to persist across page scrolls in the live preview pane. Previously, selections were tied to screen cells and would disappear when users scrolled, preventing them from selecting content spanning multiple pages. The PR makes selections "anchored to scrollback" so they remain stable and valid as the visible content changes.

## What Changed

**Core Selection Model:**
- Selection endpoints are now stored using absolute scrollback line indices (anchored to content) rather than screen cell coordinates
- A new `PreviewTextView` component acts as a render-time snapshot, recording the pane's visible geometry, first visible line, and total lines to enable stable coordinate mapping across frames
- The coordinate system changed from absolute screen positions to a hybrid system using column offset and "distance from bottom" (anchored to the bottom-most captured line rather than absolute indices) to preserve selection stability when new content arrives during live capture

**Selection Interaction:**
- Mouse drag selection now uses the scrollback-anchored coordinates, allowing selections to persist while scrolling
- Edge auto-scroll was refactored to use a ticker-driven mechanism: holding the cursor at the pane edge triggers continuous scrolling that extends the selection in real-time via `tick_preview_autoscroll()` 
- Selections are no longer cleared when scrolling, since they reference absolute content lines instead of screen positions
- The anchor system uses "from_bottom" distances so that when the captured window grows (older lines prepended during live re-capture), the same physical lines remain selected

**Copy Text Extraction:**
- The copy functionality now reads from the parsed scrollback cache instead of the visible frame buffer, capturing off-screen selected text correctly
- Selection text is extracted once when the drag finalizes, ensuring consistency
- Text extraction properly handles wide characters and line truncation by re-laying the content per-line

**Rendering & Performance:**
- Auto-scroll redraws use "diffed" updates instead of clearing the entire terminal, eliminating flicker/strobing
- Scroll pacing is limited to ~25 lines/second to prevent visual noise and racing with capture wakes
- Full terminal clears are avoided during active selection auto-scroll by using `needs_full_refresh` instead of `needs_redraw`

## What Was Added

- `PreviewTextView` struct: snapshots pane geometry (rect, first_line, total_lines) and provides methods to:
  - Check if a screen cell contains valid content via `contains(col, row)`
  - Map screen cells to scrollback-anchored coordinates via `screen_to_content(col, row)` returning `(col_offset, from_bottom)`
  - Convert between from_bottom distances and absolute indices via `abs_from_bottom()`
- `tick_preview_autoscroll()` method in `HomeView`: handles timer-driven edge auto-scroll when selection is held at pane boundaries, returns `bool` indicating if redraw needed
- `preview_drag_pos` and `preview_autoscroll_at` fields in `HomeView`: track active drag state and auto-scroll timing
- Comprehensive test suite: unit tests validating scrollback-anchored behavior, multi-page selection, coordinate mapping under scrolling, persistence across window growth, and auto-scroll timing
- Helper methods for coordinate conversion between screen and content spaces
- Integration with the app event loop via `App::run` calling `tick_preview_autoscroll()` each iteration

## What Was Removed

- Screen-cell-based coordinate system for selections (replaced by scrollback-based system)
- Selection clearing on scroll operations
- Direct buffer-reading during selection painting (now extracted once at finalization)

## Benefits

- **Multi-page selection**: Users can now select and copy text spanning multiple visible pages by dragging and scrolling, directly fixing the limitation from issue `#1961`
- **Smoother interaction**: Ticker-driven auto-scroll provides responsive, continuous scrolling without requiring repeated mouse movements or visual stalling
- **Better performance**: Diffed redraws avoid terminal flicker and reduce rendering overhead compared to full screen clears
- **More stable selections**: Distance-from-bottom anchoring ensures selections remain locked to the same physical lines even as new content arrives during live capture and the capture window grows
- **Accurate copy text**: Selections now capture the actual content from scrollback rather than what's currently visible, including off-screen lines, with proper wide-character and truncation handling

## Technical Notes

The implementation achieves scrollback-anchored selection through:
1. **Render-time snapshots** (`PreviewTextView`): Records pane rect, first visible line, and total lines to enable consistent coordinate translation across frames
2. **Hybrid coordinates**: Selection stored as `(col_offset, from_bottom)` tuples, where `from_bottom` provides stability against content growth (distance up from the newest captured line)
3. **Parsed scrollback reading**: Copy extracts text from cached parsed output rather than frame buffers, handling wide-char/truncation rendering rules
4. **Ticker-based auto-scroll**: Uses interval-driven scroll updates instead of event-driven, providing smooth continuous scrolling at pane edges without terminal strobing
5. **Diffed redraw path**: Skips full terminal clears during auto-scroll, painting only changed regions via `needs_full_refresh` flag

## Known Limitations

- Depends on OSC 52 clipboard availability
- tmux capture-pane bounds may affect captured content
- Drift possible when selecting actively streaming output (content arriving faster than scroll/selection can track)
- Selection persist across scrolls but multi-page drag behavior may have variance in live-send scenarios

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -1143,8 +1143,13 @@ impl App {
             // single drag can grab more than a page without depending on
             // mouse movement to fire events. No-op unless a drag is live
             // and the pointer sits at the edge.
+            //
+            // Request a normal (diffed) redraw via `refresh_needed`, NOT
+            // `needs_redraw`: the latter forces a `terminal.clear()` at the
+            // top of the loop, and clearing every ticker frame while the
+            // scroll runs strobes the screen blank-then-repaint. The diffed
+            // draw at the bottom of the loop repaints smoothly.
             if self.home.tick_preview_autoscroll() {
-                self.needs_redraw = true;
                 refresh_needed = true;
                 needs_full_refresh = true;
             }
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -1136,6 +1136,19 @@ impl App {
             let mut refresh_needed = false;
             let mut needs_full_refresh = false;
 
+            // Continuous edge auto-scroll for a preview drag-select. The
+            // mouse-event arm `continue`s above, so this runs on the
+            // ~33ms ticker (and other wakes): while the cursor is held at
+            // the pane edge, scroll one line and extend the selection so a
+            // single drag can grab more than a page without depending on
+            // mouse movement to fire events. No-op unless a drag is live
+            // and the pointer sits at the edge.
+            if self.home.tick_preview_autoscroll() {
+                self.needs_redraw = true;
+                refresh_needed = true;
+                needs_full_refresh = true;
+            }
+
             // Update-check / install-status polls can flip the
             // bottom-of-screen update bar (banner or transient toast)
             // on or off, which shifts the home view's layout. If a
--- a/src/tui/components/preview.rs
+++ b/src/tui/components/preview.rs
@@ -503,7 +503,12 @@ fn banner_block_area(output: Rect, banner: Rect) -> Rect {
 /// Pick the row offset passed to `Paragraph::scroll`. Zero user offset shows
 /// the bottom of the cached pane (live-follow). A positive offset scrolls the
 /// same number of lines back, saturating at the top of the capture.
-fn compute_scroll(line_count: usize, visible_height: usize, user_offset: u16) -> u16 {
+///
+/// Exposed at crate visibility so the preview drag-select code can map a
+/// screen row to the absolute content line under it: the value returned here
+/// is the index of the parsed-text line painted on the output pane's top row,
+/// so `first_line + (screen_row - pane.y)` is the line beneath any cell.
+pub(crate) fn compute_scroll(line_count: usize, visible_height: usize, user_offset: u16) -> u16 {
     if line_count <= visible_height {
         return 0;
     }
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -257,6 +257,23 @@ pub(super) fn build_tool_hotkey_cache(
         .collect()
 }
 
+/// Pull the cell symbols in columns `[from, to_excl)` out of a parsed
+/// scrollback `Line`. The line is laid back out into a one-row buffer at
+/// the pane width so wide characters, combining marks, and right-edge
+/// truncation resolve exactly as ratatui rendered them on screen; reading
+/// cell symbols then mirrors the old frame-buffer extraction cell for
+/// cell. Unwritten cells read as a single space, which the caller trims.
+fn slice_line_columns(line: &ratatui::text::Line, from: u16, to_excl: u16, width: u16) -> String {
+    let mut buf = ratatui::buffer::Buffer::empty(ratatui::layout::Rect::new(0, 0, width, 1));
+    buf.set_line(0, 0, line, width);
+    let hi = to_excl.min(width);
+    let mut out = String::new();
+    for col in from..hi {
+        out.push_str(buf[(col, 0)].symbol());
+    }
+    out
+}
+
 impl HomeView {
     pub fn is_diff_open(&self) -> bool {
         self.diff_view.is_some()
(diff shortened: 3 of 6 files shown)


njbrake:

Hi @hua3505, thanks for the clear report and the screen recording; they made this easy to pin down.

#1980 adds proper select-and-scroll in live mode: the selection no longer cancels when you scroll, you can drag to the top or bottom edge to keep scrolling and extend the selection past a single page, and the copy now grabs the whole range including the parts that scrolled off screen.

Would you be able to build from the `fix/live-mode-multipage-select` branch and confirm it solves your case? Check out the branch, run `cargo build --release`, and use `./target/release/aoe`. In live mode, click and drag over the output, drag to the top edge to scroll back, then release to copy.

One thing to know up front: the copy is sent to your clipboard via the OSC 52 terminal escape, so it only works if your terminal has OSC 52 / clipboard access enabled. If the paste comes up empty, that is the likely cause; let us know which terminal you are using and we can help.

PR #1990, opened 2026-06-05
fix: on create hooks in container

### LINKED ISSUE #1989: bug: on_create hooks run on host instead of inside container when using aoe add --sandbox from CLI
When running aoe add . --sandbox from the CLI, on_create hooks always execute on the host machine instead of inside the sandbox container. This contradicts the documented behavior ("Hooks run inside the container when sandboxed") and causes the hooks to fail on macOS with Read-only file system (os error 30) because paths like /workspace/... don't exist on the host and the macOS system volume is SIP-sealed.

---

**Steps to reproduce**

```
# .agent-of-empires/config.toml
[sandbox]
default_image = "ghcr.io/njbrake/aoe-sandbox:latest"
volume_ignores = [".venv"]
volume_ignores_strategy = "named"

[hooks]
on_create = "uv venv /workspace/myproject/.venv --python 3.11"
```

```
aoe add . --sandbox --trust-hooks
```

Result:
```
Running on_create hooks:
    uv venv /workspace/myproject/.venv --python 3.11
Error: Hook command failed with exit code 2
  Caused by: failed to create directory `/workspace/myproject/.venv`: Read-only file system (os error 30)
```

The hook ran on the macOS host where /workspace/ doesn't exist.

Confirmation: uv python install 3.11 inside the hook reported Using CPython 3.11.13 (the host's cached version), while manually running the same command via docker exec inside the created container reported Using CPython 3.11.15 (freshly downloaded). Different versions confirm different execution environments.

**Root cause**

src/cli/add.rs line 814 unconditionally calls:

repo_config::execute_hooks(&hooks.on_create, &path, &hook_env)?;

execute_hooks hardcodes HookTarget::Local — it always runs on the host shell, regardless of sandbox mode. The instance.sandbox_info populated a few lines earlier (line 733) is never consulted for hook execution.

The TUI path (src/tui/creation_poller.rs) gets this right — it checks data.sandbox and routes to execute_hooks_in_container_streamed via docker exec. The CLI path never received the same treatment.

**Expected behavior**

When instance.sandbox_info.is_some(), the CLI add command should route on_create hooks to execute_hooks_in_container (starting the container first if needed), matching the TUI path's behaviour in creation_poller.rs.

**Suggested fix**

In src/cli/add.rs, around line 813–814, branch on sandbox_info:

```
let hook_env = repo_config::lifecycle_env_vars(&instance);
if let Some(ref sandbox) = instance.sandbox_info {
    // Start the container first, then exec the hook inside it
    instance.get_container_for_instance()?;
    repo_config::execute_hooks_in_container(
        &hooks.on_create,
        &sandbox.container_name,
        &instance.container_workdir(),
        &hook_env,
    )?;
} else {
    repo_config::execute_hooks(&hooks.on_create, &path, &hook_env)?;
}
```

**Environment**

- aoe version: v1.10.0
- OS: macOS (Apple Silicon, Docker Desktop with VirtioFS)
- Triggered when using: aoe add . --sandbox (CLI, not TUI)

Davicittod:
## Description

Fixes #1989.

When using `aoe add . --sandbox` from the CLI, `on_create` hooks were running on the host machine instead of inside the sandbox container. This caused failures on macOS where host paths like `/workspace/...` don't exist and SIP makes the filesystem read-only. The issue reporter also observed the wrong tool versions (host's CPython 3.11.13 vs the container's 3.11.15), confirming the hook environment was the host, not the container.

Root cause: `src/cli/add.rs` called `execute_hooks()` unconditionally with `HookTarget::Local`, ignoring the `sandbox_info` populated by `--sandbox`. The TUI path in `creation_poller.rs` already handled this correctly via `execute_hooks_in_container_streamed`.

Fix: check for `sandbox_info` before executing hooks. If set, start the container via `get_container_for_instance()` and route through `execute_hooks_in_container()`. Otherwise fall back to the existing local path. The CLI path now mirrors the TUI logic exactly.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added `test_cli_add_sandbox_on_create_hooks_run_in_container` to `tests/e2e/sandbox.rs`. The hook writes a marker to `/tmp/aoe-sandbox-hook-{pid}`. Host and container each have an independent `/tmp` namespace, so if the hook ran on the host (the bug) the file appears there and the assertion fails. Combined with asserting `aoe add` exited successfully (proving the hook ran somewhere), the two together prove it executed inside the container. Marked `#[ignore = "requires Docker daemon"]` matching the existing sandbox test convention. Verified locally with Docker.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Used Claude Code to identify the root cause, implement the fix, and write the regression test. Reviewed and verified each step manually, including running the e2e test against a live Docker daemon.

## What Changed

- Fixed CLI so `on_create` hooks run inside the sandbox container when `aoe add ... --sandbox` is used, making CLI behavior match the TUI.
- CLI now checks for sandbox mode; if sandboxed it starts the container and executes hooks inside it, otherwise it falls back to executing hooks on the host.
- Added an e2e regression test (ignored in CI — requires Docker) that verifies the hook executes inside the container by creating a container-only /tmp marker.

Benefits:
- Prevents hook failures on macOS caused by missing container paths or read-only system volumes.
- Ensures consistent behavior between CLI and TUI.
- Avoids surprises from mismatched tool versions between host and container.

Technical details

- Files changed:
  - src/cli/add.rs: branch on `instance.sandbox_info`; for sandboxed instances call `get_container_for_instance()` and `repo_config::execute_hooks_in_container(...)` instead of always calling `execute_hooks()` — (+14 / -1).
  - tests/e2e/sandbox.rs: added ignored Docker-dependent test `test_cli_add_sandbox_on_create_hooks_run_in_container` that asserts hooks run in the container by writing a container-only /tmp marker — (~+64).

- Key calls:
  - `instance.sandbox_info.is_some()`
  - `instance.get_container_for_instance()`
  - `instance.container_workdir()`
  - `repo_config::execute_hooks_in_container(...)`

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -811,7 +811,20 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                     println!("  {}", cmd);
                 }
                 let hook_env = repo_config::lifecycle_env_vars(&instance);
-                repo_config::execute_hooks(&hooks.on_create, &path, &hook_env)?;
+                if instance.sandbox_info.is_some() {
+                    instance.get_container_for_instance()?;
+                    let workdir = instance.container_workdir();
+                    let container_name =
+                        instance.sandbox_info.as_ref().unwrap().container_name.as_str();
+                    repo_config::execute_hooks_in_container(
+                        &hooks.on_create,
+                        container_name,
+                        &workdir,
+                        &hook_env,
+                    )?;
+                } else {
+                    repo_config::execute_hooks(&hooks.on_create, &path, &hook_env)?;
+                }
                 println!("✓ on_create hooks completed");
             }
         }
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -814,14 +814,14 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                 if instance.sandbox_info.is_some() {
                     instance.get_container_for_instance()?;
                     let workdir = instance.container_workdir();
-                    let container_name =
-                        instance.sandbox_info.as_ref().unwrap().container_name.as_str();
-                    repo_config::execute_hooks_in_container(
-                        &hooks.on_create,
-                        container_name,
-                        &workdir,
-                        &hook_env,
-                    )?;
+                    if let Some(ref sandbox) = instance.sandbox_info {
+                        repo_config::execute_hooks_in_container(
+                            &hooks.on_create,
+                            &sandbox.container_name,
+                            &workdir,
+                            &hook_env,
+                        )?;
+                    }
                 } else {
                     repo_config::execute_hooks(&hooks.on_create, &path, &hook_env)?;
                 }
--- a/tests/e2e/sandbox.rs
+++ b/tests/e2e/sandbox.rs
@@ -5,7 +5,7 @@ use crate::harness::TuiTestHarness;
 /// Exercises `aoe add --sandbox` which builds the full container config.
 /// This would have caught the duplicate mount points bug (commit 92d2e53).
 ///
-/// Requires a running Docker daemon -- marked `#[ignore]` for CI.
+/// Requires a running Docker daemon; marked `#[ignore]` for CI.
 #[test]
 #[serial]
 #[ignore = "requires Docker daemon"]
@@ -44,7 +44,7 @@ fn test_cli_add_with_sandbox() {
 /// `HookTarget::Local`, ignoring `sandbox_info`. The TUI already handled this
 /// correctly via `execute_hooks_in_container_streamed`.
 ///
-/// Requires a running Docker daemon -- marked `#[ignore]` for CI.
+/// Requires a running Docker daemon; marked `#[ignore]` for CI.
 #[test]
 #[serial]
 #[ignore = "requires Docker daemon"]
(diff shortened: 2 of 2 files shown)


njbrake (approved):

Verified the bug premise end to end: `sandbox_info` is populated at add.rs:733 before hooks run, the prior CLI path called `execute_hooks` with a hardcoded `HookTarget::Local`, and the TUI path in `creation_poller.rs` already routed through the container, so CLI and TUI had genuinely diverged. The fix mirrors the TUI logic, and the regression test's host vs container `/tmp` marker cleanly distinguishes the bug from the fix. CodeRabbit's only note (a disallowed prose separator in a doc comment) is already addressed in commit 388bb8bb. Approving.

njbrake (approved):

Re-approving after pushing 8f5e3c80, which only reformats the sandbox hook branch to `if let Some(ref sandbox)` (matching the TUI path and dropping the two guarded unwraps) to clear the `cargo fmt --check` failure. No behavior change; verified locally with cargo fmt, cargo check, and clippy.

PR #2000, opened 2026-06-05
feat(build): share dependency builds across worktrees via kache

### LINKED ISSUE #1663: Dev: shared build cache across worktrees, cold cargo build recompiles ~400 deps per worktree

## Problem

Working across multiple git worktrees is a core AoE workflow (AoE itself creates worktrees, and the maintainer keeps a worktree per in-flight branch). But every worktree gets its own `target/` directory, so the first `cargo build` in a freshly checked-out worktree recompiles all ~400 dependency crates from scratch, even when `Cargo.lock` is byte-identical to a worktree that already built them five minutes ago.

The wasted work is almost entirely third-party crates that did not change. On this machine the per-worktree `target/` dirs are already 3 to 10 GB each, multiplied across 11 worktrees. The cost is both time (a cold dep build is minutes) and disk (tens of GB of duplicated artifacts).

There is no shared build cache, no compiler cache, and no `[profile.dev]` tuning. The branch name `speed-up-debug-builds` captures the broader goal: make the local dev build loop fast.

## What's there today

`.cargo/config.toml` sets only the job count and the xtask alias. No `[build] target-dir`, no `[build] rustc-wrapper`:

```toml
[build]
jobs = 8

[alias]
xtask = "run -p xtask --"
```
(`.cargo/config.toml:1`)

`Cargo.toml` defines `release` and `dev-release` profiles but no `[profile.dev]` override, so debug builds inherit Cargo defaults: full debuginfo and the platform default linker (`ld` on macOS), both slow:

```toml
[profile.release]
lto = true
codegen-units = 1
strip = true

[profile.dev-release]
inherits = "release"
lto = false
codegen-units = 16
debug-assertions = false
```
(`Cargo.toml:206-216`)

No `sccache` / `RUSTC_WRAPPER` / `mold` / `lld` referenced anywhere in the repo (grep returns nothing). `sccache` is not installed.

`git worktree list` shows 11 worktrees, each with an independent `target/`:

```
3.2G  .../fix-expanding-a-huge-queued-prompt-is-unbounded/target
3.6G  .../feat-allow-reording-workspace-groups/target
5.5G  .../dev-only-playwright-showcase-video-generator-script/target
9.9G  .../fix-duplicate-worktree-branch-.../target
```

`docs/development.md:1-11` documents the three build profiles but says nothing about sharing a cache across worktrees.

## Root cause

Cargo keys its build fingerprint cache to the target directory. Each worktree resolves `target/` relative to its own root, so two worktrees on the same `Cargo.lock` never share a single compiled dependency artifact. The recompile is not Cargo being wrong; it is just that nothing tells Cargo the dep artifacts already exist next door.

## Proposal

Three independent levers, smallest first. A is the direct fix for the reported pain; B and C are the broader `speed-up-debug-builds` goal.

### A. Shared `CARGO_TARGET_DIR` across worktrees (direct fix)

Point all worktrees at one target directory so dependency crates compile once and every worktree reuses them. Cargo fingerprinting keeps feature sets and profiles (`serve` vs not, `dev` vs `dev-release`) in separate subtrees, so they coexist safely.

Two delivery options:

- **Document it** in `docs/development.md` and `CONTRIBUTING.md`: recommend `export CARGO_TARGET_DIR="$HOME/.cache/aoe-target"` (or similar) in the shell profile. Zero repo change, opt-in, no surprises.
- **Optionally** have `aoe`'s own worktree creation flow (`src/git/`) seed the env or a per-worktree `.cargo/config.toml` so worktrees aoe creates inherit the shared dir. This is bigger and product-facing; likely out of scope for a first pass, but worth recording as a follow-on.

Known caveat to call out in the docs: a shared target dir holds a single build lock, so two concurrent `cargo build`s in different worktrees serialize (`Blocking waiting for file lock on build directory`). For a solo dev switching worktrees serially this is a non-issue; for parallel CI-style builds it is.

### B. `sccache` as `RUSTC_WRAPPER` (cross-cutting compiler cache)

A compiler-level cache that survives even separate target dirs and `cargo clean`, keyed on crate inputs. Complements A: it covers the cases A misses (different `Cargo.lock`s that still share most crate versions, full cleans, fresh clones).

Document an opt-in setup (`cargo install sccache`, `export RUSTC_WRAPPER=sccache`). Do not hardcode `rustc-wrapper` in the committed `.cargo/config.toml`: it would hard-fail any contributor who lacks sccache. Env-var opt-in is the safe shape. Note the standard caveat that sccache and incremental compilation conflict, so an sccache setup should set `CARGO_INCREMENTAL=0`.

### C. `[profile.dev]` and linker tuning (general debug-build speedup)

Cut per-build link and codegen time, independent of caching:

```toml
[profile.dev]
debug = "line-tables-only"  # smaller debuginfo, faster link, keeps backtraces
```

Plus a faster linker via `.cargo/config.toml` (platform-gated): `lld` / `mold` on Linux, and on macOS the newer `ld_prime` (`-fuse-ld=...` rustflags). Linker choice is the single biggest lever on incremental debug-build wall time and does not touch the per-worktree cache problem, so it stacks cleanly on A.

## Why now

- Multi-worktree is the maintainer's default workflow; the cold-rebuild tax is paid many times a day.
- Disk: tens of GB of duplicated dep artifacts across worktrees today.
- Smallest version (proposal A, docs-only) is near-zero risk and ships in one PR.
- Pairs with #1370 (the other dev-loop-speed papercut).

Implementation MUST add tests or a verifiable check covering the behavior below; PR is not done without them. (Build-cache behavior is awkward to unit-test; a scripted check or CI smoke is the realistic layer, called out per story.)

- As a developer, when I build worktree B with `CARGO_TARGET_DIR` pointed at the same dir a freshly built worktree A used and `Cargo.lock` is identical, then no third-party dependency crate recompiles (only first-party crates may).
  - Suggested test layer: a scripted check (shell or `xtask`) that builds twice into a shared temp target dir and asserts the second build's `cargo build -v` output shows `Fresh` for dependency crates, plus a `docs/development.md` snippet kept honest.
- As a developer, when I build the same worktree with and without `--features serve` into a shared target dir, then both binaries build successfully and neither clobbers the other's artifacts.
  - Suggested test layer: same scripted check, two feature sets into one target dir, assert both link.
- As a contributor without `sccache` installed, when I run a plain `cargo build` after the change lands, then the build succeeds with no error about a missing rustc wrapper.
  - Suggested test layer: covered implicitly by CI (CI runners have no sccache); assert the committed `.cargo/config.toml` carries no unconditional `rustc-wrapper`.

## Open questions

- Delivery of proposal A: docs-only opt-in, or wire the shared dir into aoe's own worktree creation in `src/git/`? Docs-only is the safe first pass; the product-side wiring is a separate decision.
- `rust-embed` edge case with a shared target dir: the `serve` build embeds `web/dist` at compile time relative to `CARGO_MANIFEST_DIR`. If two worktrees have byte-identical Rust source but different `web/dist` contents, a shared target could serve a stale embedded bundle. Confirm whether the existing `rerun-if-changed=web/src` (`build.rs:72`) plus source-hash fingerprinting is enough, or whether the embed needs an explicit cache-bust.
- Should C ship in the same PR as A, or split? Linker tuning is platform-specific and benefits from its own testing pass per OS.

## Out of scope

- Remote / distributed build cache (`sccache` S3/Redis backends, `cargo-remote`).
- CI build-time optimization (this issue is the local dev loop; CI caching is separate).
- Changing the `release` / `dev-release` profiles.

## Cross-references

- #1370 — adjacent dev-loop-speed papercut (pre-commit hook latency).

Seluj78:
> AI disclosure: this PR was implemented by Claude (Opus 4.8) driven by the repo owner.

## Description

AoE keeps several git worktrees in flight at once, and each worktree has its own `target/`. A cold `cargo build` in a fresh worktree recompiles all ~400 dependency crates even when `Cargo.lock` is identical to a worktree that built them minutes ago, and each `target/` holds its own multi-gigabyte copy of those artifacts.

This commits [kache](https://github.com/kunobi-ninja/kache) as the rustc wrapper in `.cargo/config.toml`. kache is a content-addressed build cache: it compiles each crate once and shares the artifact into every worktree's `target/` by **reflinking** it (copy-on-write clone on APFS/btrfs/xfs) or **hardlinking** it (ext4). So unchanged deps are restored, not recompiled, and one physical copy backs all worktrees. This solves both compile **time** and disk **size**. Each worktree still produces its own `target/debug/aoe`.

Because the wrapper is committed, kache is required to build locally. CI, the Nix sandbox, and release builds set `CARGO_BUILD_RUSTC_WRAPPER=""` to disable it: those environments have no kache, must stay reproducible, and a committed wrapper would otherwise hard-fail every runner. Release binaries stay compiled by plain rustc, not a third-party wrapper.

Closes #1663

## PR Type

- [x] New Feature
- [x] Infrastructure / CI
- [x] Documentation

## Checklist

- [ ] For UI changes: included screenshot or recording (N/A, no UI change)

## How I tested

Run these from a worktree on macOS (APFS) or Linux with kache installed:

- [ ] `bash scripts/verify-shared-target.sh --self-test` passes (cargo/kache-free; guards the report-threshold checker)
- [ ] `bash scripts/verify-shared-target.sh` passes: builds the workspace into two isolated target dirs through kache and asserts the warm build is zero-copy. Local run reported `restored 582104870 B: reflinked 582104870, hardlinked 0, copied 0 (100.0% zero-copy)` and the `--features serve` build coexists against the same store.
- [ ] `cargo build --all-features` produces `target/debug/aoe` through the committed wrapper
- [ ] `CARGO_BUILD_RUSTC_WRAPPER="" cargo build` builds without kache (the CI/opt-out path)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: build config, docs, and a shell verification script only; no TUI/CLI/session behavior changes. Coverage is the `scripts/verify-shared-target.sh --self-test` job wired into `ci.yml`.

## AI Usage

- [x] AI was used

## Overview
This PR introduces **kache**, an opt-in shared build cache system that allows multiple git worktrees to reuse compiled dependencies instead of recompiling them. Rather than being built-in, developers choose to enable it via environment variables. This solves the problem where separate worktrees each compile the same dependencies from scratch, consuming time and disk space.

## What Changed

**Added:**
- **Verification script** (`scripts/verify-shared-target.sh`) that confirms the shared-cache deduplication works correctly, with a self-test mode and a full verification mode that builds the workspace twice and validates that dependencies are restored via zero-copy mechanisms (reflinks/hardlinks)
- **Documentation** in `docs/development.md` explaining how to opt into kache, with step-by-step setup instructions, monitoring commands, and important caveats (e.g., same-filesystem requirement, behavior with embedded web assets)
- **Documentation update** in `CONTRIBUTING.md` noting kache as an optional prerequisite for faster builds
- **Safety test** (`tests/build_cache_config.rs`) that prevents developers from accidentally committing a rustc-wrapper configuration; the test ensures `.cargo/config.toml` never contains a committed wrapper
- **CI job** (`build-cache-selftest` in `ci.yml`) that runs the verification script to automatically validate the shared-cache behavior
- **Dev dependency** (`toml = "1.1"` in `Cargo.toml`) to support the configuration test

**Removed:**
- Any unconditional rustc-wrapper from `.cargo/config.toml` (kache is now opt-in only)
- Explicit `CARGO_BUILD_RUSTC_WRAPPER` overrides from CI, Nix, and release builds

## Benefits
- **Faster rebuilds across worktrees**: Developers working with multiple git worktrees avoid recompiling ~400 dependency crates
- **Reduced disk usage**: Compiled artifacts are shared and deduplicated across worktrees instead of duplicated
- **Opt-in design**: Plain `cargo build` works as before without setup; kache is only used when explicitly enabled
- **Automatic verification**: CI validates that the shared-cache mechanism works correctly, catching regressions early
- **Clear documentation**: Developers know exactly how to opt in and what to expect

## Technical Details
- kache operates as a `RUSTC_WRAPPER` that acts as a content-addressed cache for compiler outputs
- Shared artifacts are restored via reflinks (fast, zero-copy when supported) or hardlinks, reducing I/O overhead
- When enabled, `CARGO_INCREMENTAL=0` must be set to prevent conflicts with the wrapper
- The verification script confirms that at least 90% of restored bytes use zero-copy mechanisms (reflinks or hardlinks)
- Each worktree still maintains its own `target/debug` directory while sharing the underlying dependency artifacts

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,17 +1,5 @@
 [build]
 jobs = 8
-# kache (https://github.com/kunobi-ninja/kache) is a content-addressed rustc
-# wrapper: it compiles each dependency crate once and hardlinks the artifact
-# into every worktree's target/, so switching worktrees no longer recompiles
-# the ~400 dependency crates. Each worktree keeps its own target/debug/aoe.
-# kache is therefore a required local build tool; see docs/development.md.
-# CI, Nix, and release builds set CARGO_BUILD_RUSTC_WRAPPER="" to disable it
-# (those environments are ephemeral or must stay reproducible, and a committed
-# wrapper would otherwise hard-fail any runner without kache installed).
-rustc-wrapper = "kache"
-# kache and incremental compilation are mutually exclusive; kache caches whole
-# crate artifacts, incremental caches per-CGU and conflicts with the wrapper.
-incremental = false
 
 [alias]
 xtask = "run -p xtask --"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,17 @@
 [build]
 jobs = 8
+# kache (https://github.com/kunobi-ninja/kache) is a content-addressed rustc
+# wrapper: it compiles each dependency crate once and hardlinks the artifact
+# into every worktree's target/, so switching worktrees no longer recompiles
+# the ~400 dependency crates. Each worktree keeps its own target/debug/aoe.
+# kache is therefore a required local build tool; see docs/development.md.
+# CI, Nix, and release builds set CARGO_BUILD_RUSTC_WRAPPER="" to disable it
+# (those environments are ephemeral or must stay reproducible, and a committed
+# wrapper would otherwise hard-fail any runner without kache installed).
+rustc-wrapper = "kache"
+# kache and incremental compilation are mutually exclusive; kache caches whole
+# crate artifacts, incremental caches per-CGU and conflicts with the wrapper.
+incremental = false
 
 [alias]
 xtask = "run -p xtask --"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  # The committed .cargo/config.toml sets rustc-wrapper = "kache" for fast local
+  # multi-worktree builds. CI runners have no kache; this empty value overrides
+  # that config key so cargo builds without the wrapper.
+  CARGO_BUILD_RUSTC_WRAPPER: ""
 
 permissions:
   contents: read
@@ -118,6 +122,17 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
       - name: Check skill file references valid CLI commands
         run: cargo xtask check-skill
+
+  build-cache-selftest:
+    name: Build Cache Self-Test
+    runs-on: ubuntu-latest
+    # Guards the device:inode + link-count comparator that the kache
+    # verify script relies on (scripts/verify-shared-target.sh). No cargo,
+    # no kache, no toolchain: builds a known hardlinked/copied pair and
+    # asserts the comparator tells them apart. Runs in seconds.
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - run: scripts/verify-shared-target.sh --self-test
 
   nix:
     name: Nix Eval
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,10 +126,11 @@ jobs:
   build-cache-selftest:
     name: Build Cache Self-Test
     runs-on: ubuntu-latest
-    # Guards the device:inode + link-count comparator that the kache
-    # verify script relies on (scripts/verify-shared-target.sh). No cargo,
-    # no kache, no toolchain: builds a known hardlinked/copied pair and
-    # asserts the comparator tells them apart. Runs in seconds.
+    # Guards the zero-copy threshold checker that the kache verify script
+    # relies on (scripts/verify-shared-target.sh). No cargo, no kache, no
+    # toolchain: feeds a known reflinked report and a known all-copies report
+    # through the checker and asserts it accepts the first and rejects the
+    # second. Runs in seconds.
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - run: scripts/verify-shared-target.sh --self-test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,10 +8,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  # The committed .cargo/config.toml sets rustc-wrapper = "kache" for fast local
-  # multi-worktree builds. CI runners have no kache; this empty value overrides
-  # that config key so cargo builds without the wrapper.
-  CARGO_BUILD_RUSTC_WRAPPER: ""
 
 permissions:
   contents: read
@@ -133,6 +129,8 @@ jobs:
     # second. Runs in seconds.
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - run: scripts/verify-shared-target.sh --self-test
 
   nix:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,6 +14,12 @@ on:
       - '.github/workflows/docs.yml'
   workflow_dispatch:
 
+env:
+  # The committed .cargo/config.toml sets rustc-wrapper = "kache" for fast local
+  # multi-worktree builds. CI runners have no kache; this empty value overrides
+  # that config key so `cargo xtask gen-docs` builds without the wrapper.
+  CARGO_BUILD_RUSTC_WRAPPER: ""
+
 permissions:
   contents: read
   pages: write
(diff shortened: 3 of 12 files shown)


Seluj78:
@coderabbitai review

njbrake (changes requested):
Can we flip it so that the default is off but if you set the env var then it's on?

Seluj78:

@njbrake done, flipped to opt-in. kache is no longer committed in `.cargo/config.toml`; each developer enables it in their own shell with `RUSTC_WRAPPER=kache` + `CARGO_INCREMENTAL=0`. A plain `cargo build` needs no setup and behaves exactly as before, so the CI/Nix/release `CARGO_BUILD_RUSTC_WRAPPER=""` overrides are gone too.

`docs/development.md` and `CONTRIBUTING.md` now document the opt-in path, and `tests/build_cache_config.rs` guards the invariant: it fails if a `rustc-wrapper` is ever committed to `.cargo/config.toml` again.

https://github.com/agent-of-empires/agent-of-empires/pull/2000/commits/74dfd677

njbrake (approved):

Issue #1196, opened 2026-05-17
Cockpit: force_end_turn doesn't unstick session when agent ignores session/cancel, daemon prompt loop wedged and follow-ups silently dropped


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Cockpit session `dc2d182d17214272` got wedged for ~20 minutes today. The agent (claude-agent-acp) was running a `TaskOutput` tool call with `block: true, timeout: 600000` against a hung background `cargo test` shell. I clicked **Force end turn** twice and typed two follow-up prompts ("Stop trying to build for now move on", "Hello ?"). The spinner cleared each time, but the agent never responded; my prompts vanished. The session only recovered after I `ssh`'d in and `kill`ed the two wedged shell PIDs manually.

The headline failure is upstream (claude-agent-acp ignored `session/cancel` while a tool call was blocking), but the aoe daemon's response to "cancel didn't take" is silent prompt-drop with no escalation, no surfacing, no recovery action. Force end turn lies to the UI: the spinner stops, but the daemon stays wedged.

## Evidence from the affected session

From `~/.agent-of-empires/cockpit_events.db`, lifecycle events for `dc2d182d17214272`:

```
seq | time     | event
800 | 14:23:44 | ToolCallStarted name=TaskOutput
801 | 14:23:45 | ToolCallUpdated args={"block":true,"task_id":"bt3f4lzo4","timeout":600000}
803 | 14:24:49 | UsageUpdated (last agent-side activity)
804 | 14:37:23 | Stopped { reason: "user_forced" }       <- 1st force-end (13 min later)
805 | 14:37:33 | UserPromptSent text="Stop trying to build for now move on"
806 | 14:43:14 | Stopped { reason: "user_forced" }       <- 2nd force-end
807 | 14:43:17 | UserPromptSent text="Hello ?"
                (no further events from agent until manual kill at 14:49)
808 | 14:49:25 | ToolCallStarted name=Read               <- session unwedged
```

From `~/.agent-of-empires/debug.log` at the cancel moments:

```
12:24:49.775 INFO  cockpit.acp: sending session/cancel during in-flight prompt
12:37:23.733 INFO  cockpit.acp: sending session/cancel during in-flight prompt
12:37:33.977 WARN  cockpit.acp: received Prompt while one is in flight; dropping (queueing tracked in #1031)
[...]

Issue #1174, opened 2026-05-16
Cockpit web: #1129 fix incomplete on Android Chrome, keydown handler misses Enter when keyboard fires beforeinput-only or keyCode:229


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

#1129 shipped a fix in commit `25a671f` so plain Enter on touch-primary devices inserts a newline instead of firing the prompt. The desktop case is fixed and unit-tested. On Android Chrome the bug still reproduces: pressing the on-screen keyboard's return key sends the message instead of inserting a newline.

Reload of the dashboard since the fix landed, hard refresh, same behavior. Other touch behaviors that depend on `isMobile` (mobile toolbar, safe-area-inset handling) work, so the device IS detected as mobile elsewhere in the app.

### The Enter decision matrix is shipped and tested

`web/src/components/cockpit/Composer.tsx:50-65`:

```ts
export function decideEnterAction(
  event: { key, shiftKey, ctrlKey, metaKey, isComposing },
  ctx: { isMobile, turnActive },
): EnterAction {
  if (event.key !== "Enter") return "default";
  if (event.isComposing) return "default";
  if (event.shiftKey || event.ctrlKey || event.metaKey) return "default";
  if (ctx.isMobile) return "newline";
  if (ctx.turnActive) return "send";
  return "default";
}
```

`web/src/components/cockpit/Composer.tsx:387-422`:

```ts
onKeyDown={(e) => {
  const action = decideEnterAction(
    {
      key: e.key,
      shiftKey: e.shiftKey,
      ctrlKey: e.ctrlKey,
      metaKey: e.metaKey,
      isComposing: e.nativeEvent.isComposing,
    },
    { isMobile, turnActive },
  );
  if (action === "default") return;
  if (action === "newline") {
    e.stopPropagation();
    return;
  }
  // action === "send"
  e.preventDefault();
  e.stopPropagation();
  void sendFromTextarea(taRef, composerRuntime, enqueuePrompt);
}}
```

Tests in `Composer.enter.test.ts` cover every cell of the decision matrix. The pure logic is correct.

### Mobile detection

`Composer.tsx:96-101`:

```ts
function detectMobileInput(): boolean {
  if (typeof window === "undefined" || !window.matchMedia) return false;
  const coarse = window.matchMedia("(pointer: coarse)").matches;
[...]

Issue #680, opened 2026-04-15
web: desktop/browser notifications for attention-needed events


njbrake:
## Summary
Fire browser notifications when a session needs attention (waiting for user input, errored, finished) so users can leave the tab in the background.

## What to build
- Request Notification permission lazily (first time a user might want it)
- Notifications for: waiting-for-input, completed, errored
- Respect a per-user toggle in settings
- Click notification to focus tab and select that session

## Why
Parallel agents are useless if the user has to poll. A notification signal is essential for background workflows.

njbrake:
Already implemented. The web dashboard has full browser notification support:

- Service worker push handler (`web/public/sw.js`) with OS notifications and in-app toast fallback when focused
- `usePushSubscription` hook with Notification API permission flow, VAPID subscription, and platform detection (HTTPS, iOS standalone)
- `NotificationSettings` component in settings UI with enable/disable/test
- Per-session granular notification control (waiting/idle/error) via sidebar context menu
- Server API endpoints for subscribe/unsubscribe/test

PR #2008, opened 2026-06-06
chore(deps): adopt assistant-ui + claude-agent-acp ACP fixes, recommend opencode >=1.16.0

### LINKED ISSUE #1901: Cockpit: clean OpenCode read-card line-number wrappers

## Problem

OpenCode cockpit sessions using `gpt-5.5` render read-tool results with a verbose wrapper and numbered file lines in the web dashboard action card. The card body shows text shaped like:

```text
<path>/Users/seluj78/aoe/agent-of-empires-worktrees/Koreans/web/src/lib/sidebarGroups.ts</path>
<type>file</type>
<content>
1: import type { RepoColor } from "./repoAppearance";
2: import type {
3:   RepoGroup,
```

The line numbers are useful for agent reasoning, but they are ugly in the rendered cockpit read card. The user-facing card already has a read header, path, range metadata, truncation controls, and syntax highlighting. Repeating `<path>`, `<type>`, `<content>`, and `N:` prefixes makes the output look like raw tool transport rather than a polished file preview.

## What's there today

AoE's ACP filesystem handler returns raw file text to the agent. It reads the file, optionally applies ACP `line` / `limit` slicing, then responds with the sliced text directly. It does not add XML-ish wrappers or line prefixes:

```rust
let sliced = if line.is_some() || limit.is_some() {
    let lines: Vec<&str> = content.lines().collect();
    let start = line.map(|l| l.saturating_sub(1) as usize).unwrap_or(0);
    let limit = limit.map(|n| n as usize).unwrap_or(usize::MAX);
    let end = start.saturating_add(limit).min(lines.len());
    if start >= lines.len() {
        String::new()
    } else {
        lines[start..end].join("\n")
    }
} else {
    content
};
responder.respond(ReadTextFileResponse::new(sliced))
```

`src/cockpit/acp_client.rs:4895-4911`

When ACP sends tool completion content, AoE concatenates text blocks verbatim:

```rust
if let ToolCallContent::Content(c) = block {
    if let ContentBlock::Text(t) = &c.content {
        if !out.is_empty() {
            out.push('\n');
        }
        out.push_str(&t.text);
    }
}
```

`src/cockpit/acp_client.rs:3228-3241`

That verbatim string becomes the `ToolCallCompleted.content` event:

```rust
events.push(Event::ToolCallCompleted {
    tool_call_id: id,
    is_error,
    content: content_text,
    completed_at: chrono::Utc::now(),
});
```

`src/cockpit/acp_client.rs:2891-2897`

The event model documents that this content is the final textual output extracted from ACP `ToolCallUpdate.fields.content`:

```rust
/// Final textual output extracted from ACP `ToolCallUpdate.fields.content`
/// (concat of all `ToolCallContent::Content(Text(_))` blocks). Empty
/// when the agent emits no content blocks on completion. Renderers
/// fall back to a status word ("completed" / "tool failed") when this
/// is empty so cards still convey state.
#[serde(default)]
content: String,
```

`src/cockpit/state.rs:462-471`

The frontend read card then renders `result.text` directly through the shared highlighted block:

```tsx
const content = result?.text ?? "";

{content && status !== "err" && (
  <HighlightedBlock text={content} language={ext} maxLines={16} />
)}
```

`web/src/components/cockpit/ToolCards.tsx:690-729`

OpenCode is intentionally routed to the read card when its tool name is `read`:

```ts
const OPENCODE: AgentProfile = {
  key: "opencode",
  aliases: {
    execute: ["bash"],
    read: ["read"],
    edit: ["edit", "write"],
    search: ["grep", "glob"],
    fetch: ["webfetch"],
    think: ["task"],
  },
};
```

`web/src/lib/agentProfiles.ts:123-142`

The existing render tests only assert that the read card label and path render. They do not cover OpenCode's wrapped read result shape or the desired presentation of read output:

```tsx
it("renders read kind with a 'read' label and the file path", () => {
  const { container } = render(
    <Wrap>
      <ToolCard tool={fixtures.read} result={undefined} />
    </Wrap>,
  );
  expect(container.textContent).toContain("read");
  expect(container.textContent).toContain("/tmp/main.rs");
});
```

`web/src/components/cockpit/ToolCards.render.test.tsx:71-79`

Duplicate checks found no exact issue for this. Searches for `opencode line numbers read output cockpit`, `cockpit read tool output`, and `ToolCallCompleted content read card` did not return a matching open issue. Adjacent work includes #1702 for TUI cockpit tool cards, but this issue is specifically the web cockpit read-card presentation for OpenCode output.

## Root cause

The line-numbered wrapper is not generated by AoE's filesystem read path. AoE returns raw file text at `src/cockpit/acp_client.rs:4895-4911` and then preserves whatever text the ACP adapter/model emits in `ToolCallUpdate.fields.content` at `src/cockpit/acp_client.rs:3228-3241`.

The web read card has no presentation normalizer for known read-result envelopes. It assumes `result.text` is already the clean file body and passes it directly to `HighlightedBlock` at `web/src/components/cockpit/ToolCards.tsx:727-729`.

OpenCode's read tool is routed to the dedicated read card via `web/src/lib/agentProfiles.ts:123-142`, so the right presentation hook exists. It just lacks OpenCode-specific cleanup.

## Proposal

A. Minimum viable fix, normalize OpenCode read envelopes in the read card

Add a small pure helper in the web cockpit surface, likely near `ReadToolCard` in `web/src/components/cockpit/ToolCards.tsx` or in a tiny `web/src/lib/readToolOutput.ts` module:

```ts
type NormalizedReadOutput = {
  text: string;
  startLine: number | null;
  strippedLinePrefixes: boolean;
};
```

The helper should detect the exact wrapper shape:

```text
<path>...</path>
<type>file</type>
<content>
N: ...
N+1: ...
```

For read cards only, strip the wrapper and strip the leading `N:` prefixes when they form a contiguous sequence. Render only the file body. Keep the existing card header path and line count metadata as the source of context.

Use conservative detection so normal code examples are not damaged:

- Require `<path>`, `<type>file</type>`, and `<content>` at the top.
- Require at least two numbered content lines.
- Require numbers to be contiguous from the first line number.
- Preserve output unchanged if detection fails.

B. Preserve copy behavior intentionally

Decide whether the read card copy button should copy the normalized body or the raw ACP text. For a read-card preview, the practical default is to copy the normalized body, because users copy file snippets from this card.

C. Add focused regression coverage

Add a Vitest render test in `web/src/components/cockpit/ToolCards.render.test.tsx` or a pure helper test if the normalizer is extracted. The test should use `toolKey="opencode"`, a `read` tool, and completion text matching the report. Assert that the rendered card contains `import type { RepoColor }` but does not contain `<path>`, `<content>`, or `1:`.

Because this changes a user-facing cockpit dashboard flow, update `web/tests/coverage-matrix.json` with an entry for the cockpit read-card cleanup unless an existing cockpit tool-card entry already covers the exact surface.

## Why now

- This affects the first thing users inspect in cockpit, read-tool output, and makes OpenCode/gpt-5.5 sessions look raw and unfinished.
- The backend already returns clean file text; the fix can stay scoped to presentation of a known adapter/model envelope.
- OpenCode is a first-class cockpit-capable agent in AoE, so its read-card output should meet the same polish bar as Claude and Codex.

- As an OpenCode cockpit user, when a `read` tool completes with a `<path>/<type>/<content>` wrapper and contiguous `N:` line prefixes, then the read card shows only the file body without the wrapper or line-number prefixes.
  - Suggested test layer: Vitest + RTL in `web/src/components/cockpit/ToolCards.render.test.tsx`, or a pure helper test plus one render assertion.

- As an OpenCode cockpit user, when a `read` result contains ordinary code that happens to include colons or numbers, then the read card preserves the content unless the full OpenCode wrapper shape is present.
  - Suggested test layer: Vitest unit test for the output normalizer.

- As a dashboard user, when I copy or inspect a cleaned read-card body, then I get the same code shown in the card rather than the raw wrapper transport text.
  - Suggested test layer: Vitest + RTL if copy behavior is touched; otherwise document the decision in the helper test names.

## Open questions

- Should the cleanup apply only when the active agent profile is `opencode`, or should it apply to any read card that emits this exact wrapper shape? Profile-gating lowers false positives; shape-gating helps custom agents that reuse OpenCode-style output.
- Should the card surface the original start line in metadata after stripping prefixes, for example `L70-85`, when the wrapper starts at `70:`? The current read card already derives range from tool args when present, so adding inferred range may be unnecessary.
- Should raw ACP content remain available behind a debug affordance? The normal user path should be clean, but developers chasing adapter bugs may still want the original payload.

## Out of scope

- Changing AoE's `fs/read_text_file` response semantics. That path already returns raw file text.
- Changing OpenCode or the upstream ACP adapter.
- Adding line-number gutters to read cards. This issue is about removing duplicate inline prefixes, not designing a richer code viewer.
- Normalizing bash/search output, where line prefixes may be meaningful command output.

## Cross-references

- #1702, adjacent but not a duplicate: TUI cockpit tool-card rendering work. This issue targets the web cockpit OpenCode read-card output.

Seluj78:
## Description

Updates web/structured-view dependencies to pick up upstream fixes and bumps the claude-agent-acp floor so a key cancel-recovery fix is guaranteed. Three independent changes, one per commit:

1. **assistant-ui `^0.14.14` + mobile Enter (`fix(web)`).** Adopts the new `unstable_insertNewlineOnTouchEnter` prop on `ComposerPrimitive.Input` (upstream assistant-ui#4091 / PR #4093). On touch-primary devices the adapter now resolves the submit mode to `none`, so plain Enter inserts a native newline and never submits; the Send button is the only dispatch path. This fixes the iOS case where the keydown reached the primitive's `requestSubmit` despite the existing `beforeinput` workaround. Drops the now-dead `newline` branch from `decideEnterAction` (mobile plain Enter falls through to `default` so the primitive owns the newline, with the mobile guard kept ahead of the `turnActive` check so mobile mid-turn Enter never hits the send/queue path). Keeps the Android GBoard / keyCode-229 `onBeforeInput` backstop. Relates to #1129 / #1174.

2. **claude-agent-acp floor `>=0.41.0` (`fix(acp)`).** 0.41.0 (upstream #680) force-resolves a prompt loop wedged in a `TaskOutput { block: true }` poll against a hung background task: about 30s after the first `session/cancel` it returns the `cancelled` stop reason instead of hanging forever. Before this, a cancel during off-protocol background work rode aoe's 30-minute `OFF_PROTOCOL_WORK_GRACE_FLOOR` before recovering (the multi-minute hang in #1196). Bumps the `agent_compat` `min_version` from 0.39.0 to 0.41.0 plus its satellite references; no client logic change since the prompt loop already maps a `cancelled` PromptResponse to the terminal Stopped reason. Adds a boundary test pinning that 0.40.x is now rejected. Relates to #1196. Breaking for users pinned to 0.39.x / 0.40.x: they hit the incompatible-agent-version startup error until they reinstall via `@latest`.

3. **opencode `>=1.16.0` recommendation (`docs`).** opencode 1.16.0 classifies `apply_patch` as `edit` and `task` as `think`, populates `external_directory` permission context, and emits clean read-tool content (upstream anomalyco/opencode #30554 / #30556 / #30557 / #30558, all aoe-reported). aoe needs no code change: kind-primary dispatch with name-alias fallback and graceful degradation already absorb the pre-fix behavior. Documents the recommendation rather than adding a hard floor, since these are rendering-quality fixes, not protocol breakage. Closes #1901.

## PR Type

- [x] Bug Fix
- [x] Documentation

## Checklist

## Test Coverage Analysis

Updated the `decideEnterAction` Vitest (`web/src/components/acp/Composer.enter.test.ts`) for the new mobile branch; the matrix entry `acp.story.composer-mobile-enter-newline` already covers this Composer flow. Behavior is unchanged from the user's point of view (mobile Enter still inserts a newline), only the implementation moved to the upstream prop, which also fixes the iOS regression.

**TUI / CLI (e2e test)**

The acp floor change is covered by `src/acp/agent_compat.rs` unit tests (added `claude_below_new_floor_rejected`, refreshed the at/above/below fixtures to the new floor).

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.8 via Claude Code

**Any Additional AI Details you'd like to share:** Investigation, code, and PR drafting via back-and-forth; decisions and review are mine.

## Summary

This PR contains three independent changes that improve web input behavior, ACP agent compatibility, and documentation for rendering tools.

### What changed (high level)
- Web input: Upgraded assistant-ui and delegated mobile Enter/newline handling to the upstream component so touch devices use native newline behavior; simplified Composer Enter logic and removed the client-side "newline" branch.
- ACP compatibility: Raised the minimum required claude-agent-acp from 0.39.0 to 0.41.0 to pick up an upstream fix that resolves a cancel/prompt loop (~30s forced cancel). Tests, fixtures, and fake agents updated.
- Docs: Recommend opencode >= 1.16.0 for improved structured-view rendering and clarified behavior of older opencode adapters.
- Misc: Updated npm deps hash in flake.nix, clarified watchdog timing in Rust comments, and bumped test shim/fake agent versions to 0.41.0.

### Benefits
- Better mobile UX: Enter behaves naturally on touch devices without fragile local hacks.
- More robust ACP sessions: avoids prompt/cancel loops that could hang flows.
- Clearer upgrade guidance: documents which agent/opencode versions produce the best rendering and protocol behavior.
- Simpler client code: delegates newline handling to upstream, reducing local complexity.

### Technical notes (concise)
- Dependency: `@assistant-ui/react` ^0.14.13 → ^0.14.14.
- Composer API/type change: EnterAction narrowed from "newline" | "send" | "default" to "send" | "default"; decideEnterAction now returns "default" on touch-primary devices; ComposerPrimitive.Input uses unstable_insertNewlineOnTouchEnter.
- ACP version floor: claude-agent-acp min_version bumped to >= 0.41.0; tests add a boundary rejecting 0.40.x; fake test agent and shims advertise 0.41.0.
- Docs: recommend opencode >= 1.16.0 (recommendation only; not a hard requirement).


njbrake (approved):

Approving. Three well-scoped, single-commit changes: the assistant-ui `unstable_insertNewlineOnTouchEnter` adoption (retiring the client-side newline branch), the claude-agent-acp 0.41.0 floor bump for the cancel-wedge fix (#680 / #1196), and the opencode >=1.16.0 doc recommendation.

Verified locally on the PR branch:
- `cargo test --lib --features serve agent_compat` passes (13 tests, including the new 0.40.9 boundary rejection).
- `vitest run Composer.enter.test.ts` passes (12 tests, typecheck clean against the installed assistant-ui types).
- Traced the upstream primitive: with the prop set, touch-primary resolves `effectiveSubmitMode` to `none`, so plain Enter declines to submit and the existing `onBeforeInput` backstop inserts a single newline. No double-insert.
- The breaking 0.41.0 floor is correctly surfaced for anyone pinned to 0.39.x / 0.40.x.

The one CodeRabbit note on the acp_client.rs comment wording is an optional nit, not blocking.

Note: this is approving the change itself; the branch still has merge conflicts with main plus failing Cargo Test (ubuntu) and Playwright (live) jobs that need a look before it can land.

PR #2038, opened 2026-06-08
fix(hooks): honor agent config-dir env vars when installing status hooks

uzi:
## Description

Host status-hook installation hardcoded `~/.claude/settings.json` (and the equivalent `~/.cursor/settings.json`), ignoring `CLAUDE_CONFIG_DIR` / `CURSOR_CONFIG_DIR` even though session-id capture already respected `CLAUDE_CONFIG_DIR`. Users who point their agent at an alternate config dir (e.g. a separate work Claude via `CLAUDE_CONFIG_DIR="$HOME/.claude-work"`) got hooks written to a file the agent never reads, so file-based status detection silently fell back to the less reliable tmux pane parsing.

This adds a `config_dir_env_var` to `AgentHookConfig` and resolves the host settings path through it. When the var is set, the settings file lives directly under that directory (the var replaces the whole `~/.claude`-style dir, matching how the agents themselves interpret it), using the basename of `settings_rel_path`. Resolution order: the session's profile environment list, then AoE's own process env (so a var exported in the shell that launched `aoe`, and thus inherited by the agent, is honored), then the home-relative default.

Uninstall now sweeps the home default plus every profile/global override so hooks written under an alternate dir get cleaned up, mirroring the existing Codex cleanup logic. Codex keeps its bespoke `CODEX_HOME` handling. Sandboxed sessions are unaffected; they keep forcing the in-container config dir (`/root/.claude`).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- Added four unit tests in `src/hooks/mod.rs` covering `agent_settings_path_for_host_environment`: home-relative default, host-env override, process-env fallback, and empty-value-ignored. (The behavior is pure path resolution; a unit test is the right altitude rather than a full-binary e2e.)

## AI Usage

- [x] AI was used

Investigation, implementation, and tests were done with Claude Code; reviewed by the author before submitting.

## What Changed

This PR makes agent status-hook install/uninstall honor agent-specific config-dir environment overrides (e.g., CLAUDE_CONFIG_DIR, CURSOR_CONFIG_DIR) instead of always writing to hardcoded home paths.

High-level changes:
- AgentHookConfig gains an optional config_dir_env_var and agents that support config-dir overrides (Claude, Cursor) set it.
- Settings path resolution is environment-aware and ordered: 1) the session profile’s host environment, 2) the AoE process environment, 3) the home-relative default. Empty values are ignored.
- If an env var is used, the settings file is written directly under that directory (using the basename of the previous settings_rel_path).
- Uninstall now enumerates and removes hooks from the home default plus every profile/global override for non-Codex agents, preventing orphaned hook files. Codex retains its CODEX_HOME behavior.
- UI (TUI) consent/dialog now displays the resolved write path instead of a hardcoded ~/.… path.
- Tests and docs updated: four unit tests for path resolution, an e2e TUI/CLI test, and a documentation clarification.

Benefits
- Hooks are installed where the agent actually reads them (respects user overrides).
- Uninstall cleans alternate config dirs to avoid leaving orphaned files.
- Backwards-compatible defaults preserved; sandboxed sessions still force the in-container config dir.

Technical details (brief)
- New struct field: config_dir_env_var: Option<&'static str> on AgentHookConfig (src/agents.rs).
- New functions (src/hooks/mod.rs):
  - agent_settings_path_for_host_environment(...)
  - agent_settings_path_display_for_host_environment(...)
  - agent_settings_paths_for_uninstall(...)
- Instance::install_agent_status_hooks updated to use the environment-aware resolution instead of dirs::home_dir() + settings_rel_path.
- uninstall_all_hooks updated to iterate candidate uninstall paths for non-Codex agents.
- Tests added: four unit tests in src/hooks/mod.rs and an e2e test under tests/e2e/; docs updated in docs/guides/configuration.md.


uzi:
Thanks for the review! Addressed both in 66e8ad60:

- **Docs (`agent_status_hooks` row):** reworded so Codex no longer implies a fixed `~/.codex/config.toml`; it now lists `CODEX_HOME` alongside `CLAUDE_CONFIG_DIR`/`CURSOR_CONFIG_DIR` as config-dir overrides and says "typically `~/.codex/config.toml`".
- **Test:** added `test_agent_settings_path_host_env_takes_precedence_over_process_env` pinning that a profile-env override wins over AoE's own process env when both are set.

njbrake (approved):

Clean, well-scoped bug fix. File-based status detection was silently degrading to pane parsing whenever a user redirected their agent's config dir (e.g. `CLAUDE_CONFIG_DIR`), since hooks were written to a path the agent never reads. The fix resolves the host settings path through an optional `config_dir_env_var` on `AgentHookConfig` and faithfully mirrors the established Codex resolution/cleanup pattern.

Highlights:
- Resolution order (profile host env, then AoE process env, then home default) is correct and documented; empty overrides are treated as unset, which is slightly more robust than the Codex equivalent.
- Uninstall now sweeps the home default plus every profile/global override, so hooks under an alternate dir get cleaned up.
- Consent dialog and docs both updated to name the real target path.
- Tests are at the right altitude: six unit tests covering default, host-env override, process-env fallback, precedence, empty-value handling, and display matching. Verified locally: `cargo test --lib hooks::` passes (109) and `cargo clippy --lib` is clean.

The CODEX_HOME docs wording flagged by CodeRabbit was already addressed in 66e8ad6.

One optional follow-up: `codex_config_path_for_host_environment` does not filter empty values the way the new helper does, so an empty `CODEX_HOME=` would resolve to a relative path. Out of scope here, but worth tightening later for symmetry.

Approving.

PR #2088, opened 2026-06-10
feat: unread session state (TUI + web dashboard)

Eric162:
## Description

Adds a non-intrusive **unread** indicator on sessions, in both the **TUI and the web dashboard**, built on top of the existing turn-finish detection (the `Running → Idle` transition that already drives sound/popup notifications). Notifications are loud and ephemeral; this is the quiet, persistent complement: a glanceable cue distinguishing "this turn finished and I've looked" from "finished and I haven't."

Two kinds of unread, distinguished only by what clears them:

- **auto-unread** — set when a turn finishes (`Running → Idle`). Cleared by viewing the session (TUI: Tab/Enter; web: opening it) or the manual toggle. The session you're actively viewing is exempt.
- **manual-unread** — a deliberate "flag for later." Cleared only by the toggle, so it survives viewing.

### Enabled by default, with a setting to turn it off

On by default, gated behind the `session.unread_indicator` toggle (Settings → Session → Interaction) through the single-source settings schema, so it shows up in both the TUI and web settings automatically. Turning it off disables the indicator, auto-marking, sort promotion, and the toggle on every surface.

### TUI

- `theme.unread` row color on resting (Idle/Unknown) rows; auto-mark on turn finish with an active-view exemption; `v` to flag/unflag (gated binding, hidden from help/palette when off); a right-click context-menu entry; an Attention-sort promoter that floats unread rows just below Waiting; `w` jump-to-next includes unread.

### Web dashboard

- An unread dot on the sidebar row (theme accent via `--color-status-unread`), suppressed on the active row; a right-click **"Mark as read / unread"** context-menu item — both gated on the setting.
- `PATCH /api/sessions/{id}/unread` ({unread:bool}); `unread` exposed on the session API; optimistic overlay like pin/archive/snooze.
- The server status-poll loop auto-marks `Running → Idle` (persisted), so a **web-only** user with no TUI running still accrues the indicator. Clear-on-view is client-driven (opening a session clears an `auto` marker; a `manual` flag survives).

> 📹 TUI demo video below. Behavior is unchanged; it's just default-on now and mirrored in the web UI.

## PR Type

- [x] New Feature

## Checklist

## Test Coverage Analysis

**Web dashboard / structured view**
- [x] Added/updated tests and a `coverage-matrix.json` entry (`sidebar.unread-indicator`): RTL coverage in `SessionRowTriage.test.tsx` (dot render, active-row suppression, menu label flip, PATCH payloads for mark read/unread, feature-off gating), request-shape + resolver tests in `api.test.ts` and `sidebarOptimistic.test.ts`.

**TUI / CLI (e2e test)**
- [x] Skipped a dedicated e2e: the indicator is color-only (no text/glyph the screen-text harness can assert). Covered by unit tests instead: `Instance` mutators + serde, `attention_rank` promoter, the `v` binding, the context-menu label flip, `SessionResponse` serialization, and the read-only-guard audit for the new endpoint.

## AI Usage

- [x] AI was used

## Overview
This PR introduces a persistent **“unread” session indicator** across the TUI and web UI so users can immediately tell which idle sessions still need review versus sessions they’ve already seen.

### Unread tracking + controls
- **New setting:** `session.unread_indicator` (enabled by default) in **Settings → Session → Interaction**, applied immediately (no restart).
- Each session now tracks unread state in two ways:
  - **Auto-unread:** set automatically when a session transitions from running to idle; cleared when you view/attach, or when you manually mark it as read.
  - **Manual-unread:** toggled by the user and stays until explicitly toggled back.
- **Keyboard shortcut:** `v` (and `V` in strict mode) toggles unread for the selected session (only when enabled).
- **Right-click actions:** context menu now includes **Mark unread / Mark read** (label flips based on current state).

### Sorting + navigation behavior
- **Attention sorting promotion:** unread items are elevated in the “Attention” view so they surface just below “Waiting.”
- The **`w` jump navigation** now stops on unread sessions as well as waiting/idle sessions.

### Visual updates
- Added a new **theme color** (`theme.unread`) and applied it to resting rows to visually distinguish unread sessions.
- Built-in themes were updated to include the new unread color.

### Web UI support
- The web sidebar/session rows show an unread dot and context-menu triage toggle when enabled.
- Viewing an active session clears any auto-unread state.

## Tests / QA
- Adds unit test coverage for unread state rules, serialization/merging behavior, sorting promotion ordering, keybinding/context-menu behavior, and menu positioning.
- Skips TUI e2e testing for this feature since it’s primarily a color/indicator change.

## Benefits
- **Faster triage:** clearly separates “idle but already reviewed” from “idle and needs attention.”
- **Low friction rollout:** default-on behavior with a single easy-to-disable setting.
- **Consistent UX:** same unread concept works in both TUI and web, with matching controls and indicators.

## Potential follow-ups
- Consider further UX/feature modularization (e.g., making unread behavior easier to extend or plug in more places).


njbrake:
Woah. I love this idea. I find myself in the same situation, I forget which idle sessions i looked at vs didn't look at! Let me think about the best way to fit this into the design. 

I'm going back and forth about how we best introduce this as an optional feature or should we even just have it as the default... I think the env var is too conservative, i don't want to make it that tricky for someone to use it.

Eric162:
@njbrake sounds great - let me know what you're thinking on how to intro it, and i can update it.

Maybe Default + an off switch for those who don't like it?

Seluj78:
I love the idea and I think it would be amazing to get it in the web ui as well, with a right click `mark as read/unread` context menu options. I think we can add it as a normal feature, and move it to a plugin later @njbrake

Seluj78:
~~Hi there @Eric162, are you still interested in working on this PR ? :)~~

Nevermind you pushed 13 hours ago I thought it was a few days

Eric162:
@Seluj78 yes - let me just add the web ui stuff and get it ready for push. i'll tag you then.

njbrake (commented):

## Review

### Summary

Adds a quiet, persistent unread indicator for sessions in the TUI, complementing the existing loud, ephemeral turn-finish notifications. Two flavors: auto-unread (set on `Running -> Idle`, cleared by viewing/attaching) and manual-unread (`v` toggle, survives viewing). Surfaced as a `theme.unread` row color, a right-click menu entry, an Attention-sort promoter just below Waiting, and a stop in the `w` jump cycle. Gated behind a new default-on `session.unread_indicator` setting wired through the single-source schema.

I built the branch, ran `cargo clippy`, and ran the new unread tests locally; all clean.

### Code review

- `session/instance.rs` (`UnreadKind`, mutators, merge): clean. `mark_unread_auto` never clobbers Manual, `mark_read_auto` clears only Auto, `toggle_unread` round-trips. Serde omits `None`. Well unit-tested.
- `session/groups.rs` (`attention_rank`): nicely done. Pure, testable, monotonic shift so a feature-on run with no unread rows orders identically to before. Folded into both `attention_session_key` and `attention_group_key`.
- `session/mod.rs` (atomic flag): process-wide `AtomicBool` keeps the hot sort path off a threaded flag. Reasonable; the cost is the apply-path sync bug noted below.
- `styles/` (`theme.unread`, `fill_unread_from_accent`): accent fallback for themes that omit `unread` is correct and tested. The double parse is harmless for small theme files.
- input / bindings / context_menu / help / palette: thorough feature gating via `Context::UnreadEnabled`, with help and palette filtered separately and the menu row hidden when off. `v` / `V` has no conflicts.
- `home/mod.rs` (auto-marking, clear-on-view): live-send exemption and attach-return clear are correct; `clear_auto_unread` runs regardless of the flag so a stale Auto cannot survive a disable then re-enable. Skips disk writes when already-unread/read to avoid flock churn.

### Compliance with project guidelines

Good. The setting is added the single-source way (one `#[setting(...)]` annotation, surfaces in TUI and web automatically), no dead code, conventional commits, and doc comments explain the why. The e2e skip is justified (state is color-only; the harness asserts on screen text) and is compensated with solid unit coverage.

### Automated review check-in (CodeRabbit)

Most findings were addressed in follow-up commits (group-key promotion, the `v` quick-pick, binding gating, unconditional `clear_auto_unread`, the accent fallback, the latte color, fallback tests). Two remained open on the latest run, and I agree with both:

1. Stale Attention sort on config flip (`apply_config_to_state` in `home/mod.rs`). Real bug. `switch_profile` calls `reload()`, which rebuilds `flat_items` under the old flag, before `refresh_from_config()` flips it, and the interactive settings-close path never rebuilds afterward. So toggling the setting or switching to a profile with a different value leaves the Attention order stale until the next status tick. The row color self-corrects at render; only the ordering lags. Suggested fix: when the gate actually changes and Attention sort is active, rebuild `flat_items` and reseat the cursor by id.

2. Cursor reseat after `toggle_unread_at_cursor` (`operations.rs`). True but low priority. The cursor can shift after the reorder, but this matches the existing `toggle_favorite_at_cursor` and unsnooze paths (neither reseats; only `toggle_archive` does). I would either leave it for consistency or fix all three together in a separate change rather than here.

### The bigger question: readiness

From the thread, this is not only a code question. The shape (default-on with an opt-out) matches the direction discussed here. Seluj78 asked for web UI parity (right-click mark read/unread in the dashboard), and Eric162 replied that he would add the web UI piece and tag reviewers when ready. That web piece is not in the branch yet (the PR's own coverage analysis marks web as N/A). So by the author's own note this is still in progress.

### Verdict

- Implementation (TUI scope): solid, well-tested, idiomatic, with one minor staleness bug to fix.
- Project fit: accept. Both maintainers want it, and the default-on plus opt-out shape matches the stated preference.

The two axes agree on direction, but I would hold the merge until the web dashboard support lands (or there is an explicit decision to split TUI now and web later), and fold in the staleness fix either way.

njbrake:
Thanks! Yes I think I am comfortable merging this feature as default on and with a setting to let people disable the feature. 

I'm testing it out locally and I think the UI/UX needs some fixes. The highlighting is either too subtle or I'm doing something wrong? marking it read/unread doesn't seem to change anything when i'm not highlighting the session.

Also, I know you probably picked 'v' as the hotkey because 'u' is already a hotkey for the update option, but I think the hotkey really should be 'u', and we just have some sort of handler to handle the case where that update footer is available so that when the footer is there 'u' routes to update but otherwise 'u' routes to read/unread?

WDYT?

Once you get the changes up and ready for review again, another short screen recording of how it looks in TUI as well as web would be great! 🙏

Seluj78:
On my end, I am waiting for you @Eric162 to ping me when the web part is ready to review and test ;)

Eric162:
@Seluj78 and @njbrake 

I have done some testing and it's feeling pretty good from my end. let me know what you think!

One note - especially for @njbrake  -  i changed Update to capital `U` instead of having a conditional - it feels better both in practice and in code - and also, it is more aligned with the `R` command to restart the server in some sense. Open to changing it of course but just thought to mention it.

njbrake:
> One note - especially for @njbrake - i changed Update to capital `U` instead of having a conditional - it feels better both in practice and in code - and also, it is more aligned with the `R` command to restart the server in some sense. Open to changing it of course but just thought to mention it.

Sounds good to me! Out of curiousity I looked and the command to mark read/unread for the macos mail app is cmd+U (uppercase U) so we are at least matching the convention of one popular email client.

njbrake (approved):
Clean UX with easy to understand usage. Love it.

PR #2114, opened 2026-06-11
feat(sandbox): host-side before_start hooks that inject env into the container

### LINKED ISSUE #2110: Add host-side before_create hooks that can inject env into the container
▎ (Drafted with AI from my notes; the framing and the use case are mine — happy to discuss.)

### Summary

Add a before_create lifecycle hook that runs on the host, before the container is created, and whose KEY=VALUE stdout is injected into the new container's environment.

### Problem

There's currently no host-side hook that runs before container creation. For sandboxed sessions, both on_create hooks and custom-agent commands run inside the container (docker exec). That means there's no supported way to compute a value on the host and hand only that value to the container — you'd have to put the host-side tooling and credentials into the image, which defeats the isolation.

### Use case

Per-session, repo-scoped, short-lived credentials. I'd like each sandboxed agent to get a 1-hour GitHub token scoped to just its repo (minted on the host by a small tool). The minting needs a broad host credential; the container should only ever receive the narrow, scoped token. With a host before_create hook:

```toml
[hooks.host]
  before_create = ['echo "GH_TOKEN=$(my-mint-tool owner/repo)"']
```

→ the hook runs on the host, prints GH_TOKEN=…, and AoE injects it into the container. The minting tool and host credential never enter the container; the agent holds only a one-repo, one-hour key (and it can auto-expire, so revoke is optional).

### Proposed shape

  - New hook point before_create, host-side, runs before docker run (and on restart, so short-lived values get refreshed).
  - KEY=VALUE lines on stdout are parsed and injected into the container env (ideally leak-safe — value not in docker run argv).
  - Security: host-side hooks should be accepted only from profile/global config, never from a repo's .agent-of-empires/config.toml (a repo must not run host commands).
  - Don't let AoE log the hook's stdout (it could, and would in my example contain secret tokens)

Existing on_create/on_launch/on_destroy semantics would be unchanged; this is additive.

njbrake:
## Description

Adds a `[host_hooks] before_start` lifecycle hook that runs on the **host** each time a sandbox container comes up (create and restart) and injects the `KEY=VALUE` lines it prints to stdout into the agent's container environment. Closes #2110.

There was no host-side hook that runs before a sandbox container is brought up. For sandboxed sessions both `on_create` hooks and custom-agent commands run inside the container, so there was no supported way to compute a value on the host and hand only that value to the container without baking the host tooling and credentials into the image. The motivating use case is per-session, repo-scoped, short-lived credentials: mint a one-hour, single-repo token on the host (where the broad credential lives) and inject only the narrow token.

```toml
[host_hooks]
before_start = ['echo "GH_TOKEN=$(my-mint-tool $AOE_PROJECT_PATH)"']
```

**Design**
- Values are injected through the shared `collect_environment` chokepoint as inherited (leak-safe) entries, so every agent surface (docker run, the tmux `docker exec` launch, and the structured-view/ACP worker) picks them up and the value rides the process environment rather than appearing in argv/`ps`.
- Minting happens in `get_container_for_instance` (the shared come-up site), so a short-lived token is re-minted on every restart.
- The hook's stdout is never logged (it is the documented channel for secrets); non-`KEY=VALUE` lines are ignored; a non-zero exit hard-fails bringing the container up.
- `before_start_env` is stored on `SandboxInfo` as a runtime-only `#[serde(skip)]` field, so the secret never hits disk and a stale value never survives a restart.

**Security**
- Resolved from profile/global config only (`resolve_config_or_warn`, not the repo-merged config), and `host_hooks` is excluded from the repo-overridable sections, so a checked-out repo can never contribute host commands. Two independent guards.

The design and the refresh-on-restart semantics were agreed with the issue author in #2110.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (`docs/guides/configuration.md` "Host Hooks" section)

## Test Coverage Analysis

Added unit tests in `src/session/repo_config.rs` (KV parsing incl. CRLF / non-matching / later-wins; the host runner collecting pairs, hard-failing on non-zero exit, and keeping stdout secrets out of the error) and `src/session/environment.rs` (before_start values surface as leak-safe `Inherit` entries that win over a same-keyed `sandbox.environment` literal), plus a guard that a repo's `.agent-of-empires/config.toml` can never inject `host_hooks`.

Verified with `cargo fmt`, `cargo clippy --features serve --tests`, and `cargo test --features serve`.

**TUI / CLI (e2e test)**
- [x] N/A: covered by unit tests; no TUI rendering or CLI subcommand surface changes (the new behavior is host-hook + env-injection logic, exercised by the unit tests above)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Implemented via Claude Code through back-and-forth with @njbrake; the design decisions (injection point, refresh semantics, security boundary) are his.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2114"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783800360&installation_id=139138837&pr_number=2114&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2114&signature=0904a5c08936f9166a10dc54e0c87c252a50646e76257f1e53211df74e898bb8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR adds a host-side "before_start" lifecycle hook that runs on the host each time a sandbox container is created or restarted and securely injects KEY=VALUE pairs from the hook's stdout into the container environment.

High-level changes
- New configuration: a profile/global-only [host_hooks] section with before_start commands.
- New runtime storage: SandboxInfo.before_start_env (runtime-only, not serialized) holds minted KEY=VALUE pairs.
- Hook execution: resolve_before_start_hooks() loads hooks; run_before_start_hooks() runs them on the host, parses KEY=VALUE stdout lines, and returns pairs.
- Container startup integration: ensure_before_start_env() is invoked before building/starting containers so minted values are available to collect_environment(), which now injects these entries as leak-safe inherited env vars (first-wins dedupe).
- Security & behavior: stdout is not logged; non-KEY=VALUE lines ignored; non-zero exit aborts startup; host_hooks are allowed only in profile/global config (repos cannot supply them).
- Tests & docs: unit tests for parsing, runner behavior, precedence, repo-guarding; docs updated with a "Host Hooks" section.

Benefits
- Enables host-only minting of short-lived credentials or per-session values without embedding host tooling in images.
- Ensures per-restart freshness of values (hooks run on create and restart).
- Minimizes secret leakage: values passed via environment inheritance (not argv/ps), stdout not logged, values not persisted to disk.
- Maintains lifecycle compatibility: existing on_create/on_launch/on_destroy semantics unchanged.

Potential areas for follow-up
- UX gap: current Sandbox Configuration UI accepts only keys, not values, and supplied values may arrive too late for before_start hooks — consider a mechanism to supply per-session KEY=VALUE inputs that are accessible to host hooks at session creation time.
- Observability: deliberate choice to avoid logging hook stdout prevents accidental secret leakage but also reduces visibility; consider secure audit options for operators.

Technical notes (concise)
- parse_env_kv_lines extracts validated KEY=VALUE pairs (handles CRLF); later entries win for duplicate keys during parsing but host-minted values override sandbox/environment via first-wins injection ordering.
- run_before_start_hooks runs hooks with detached TTY, captures stdout (never included in errors), includes stderr in failures, and uses existing run_hook_with_timeout with adjusted timeout wording.
- reconcile_from_disk and structured-view flows were updated to preserve in-memory before_start_env across reloads and view switches to avoid unnecessary re-minting.

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -215,6 +215,21 @@ environment = ["GH_TOKEN=$AOE_GH_TOKEN"]                      # env vars forward
 
 See [Docker Sandbox](sandbox.md) for the full key reference (`cpu_limit`, `memory_limit`, `port_mappings`, `extra_volumes`, `volume_ignores`, `volume_ignores_strategy`, `auto_cleanup`, `default_terminal_mode`), the `environment` grammar, and credential handling. For env vars on host (non-sandboxed) sessions, use [Host Environment](#host-environment) instead; the two lists are disjoint.
 
+## Host Hooks
+
+The `[host_hooks]` block declares hooks that run on the **host** (not inside the sandbox container). Unlike `[hooks]`, which for sandboxed sessions runs inside the container, host hooks run in your host shell and can compute a value with host-only tooling and credentials, then hand only that value to the container.
+
+```toml
+[host_hooks]
+before_start = ['echo "GH_TOKEN=$(my-mint-tool $AOE_PROJECT_PATH)"']
+```
+
+`before_start` runs each time a sandbox container comes up (on create and on restart, so short-lived values are refreshed before the agent launches). Each `KEY=VALUE` line the command prints to stdout is injected into the container environment as an **inherited** variable: the value is passed to the `docker` invocation through the process environment, never in argv, so it does not appear in `ps`. Lines that are not `KEY=VALUE` are ignored, and the hook's stdout is never logged, so it is safe to print a secret. A non-zero exit aborts bringing the container up. The lifecycle env vars (`AOE_PROJECT_PATH`, `AOE_SESSION_ID`, `AOE_PROFILE`, ...) are available to the command.
+
+The canonical use case is per-session, repo-scoped, short-lived credentials: mint a one-hour, single-repo token on the host (where the broad credential lives) and inject only the narrow token, so the minting tool and host credential never enter the container.
+
+`host_hooks` is **global/profile only**: it is never honored from a repo's `.agent-of-empires/config.toml`, because a checked-out repository must not be able to run host commands. Declare it in your global or profile `config.toml`.
+
 ## tmux
 
 ```toml
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -6514,6 +6514,7 @@ mod tests {
             container_name: "aoe-sandbox-abc12345".into(),
             extra_env: Some(vec!["MY_LITERAL=hello".into()]),
             custom_instruction: None,
+            before_start_env: Vec::new(),
         };
         let config = SpawnConfig {
             agent_key: "claude".into(),
@@ -6580,6 +6581,7 @@ mod tests {
             container_name: "aoe-sandbox-abc12345".into(),
             extra_env: None,
             custom_instruction: None,
+            before_start_env: Vec::new(),
         };
         let config = SpawnConfig {
             agent_key: "claude".into(),
@@ -6650,6 +6652,7 @@ mod tests {
             container_name: "aoe-sandbox-cfgdir".into(),
             extra_env: None,
             custom_instruction: None,
+            before_start_env: Vec::new(),
         };
         let config = SpawnConfig {
             agent_key: "claude".into(),
--- a/src/acp/sandbox.rs
+++ b/src/acp/sandbox.rs
@@ -83,8 +83,12 @@ pub async fn ensure_container_for_session(
         .await
         .context("docker ensure task failed to join")??;
 
-    // Phase 3: brief write lock to stamp the resolved container_id
-    // back onto the live Instance so subsequent reads observe it.
+    // Phase 3: brief write lock to stamp the resolved container_id (and the
+    // host-minted before_start env) back onto the live Instance so subsequent
+    // reads observe them. Carrying `before_start_env` back matters because the
+    // mint ran on the cloned instance above; without this, the next
+    // ensure_container_for_session would clone an empty cache and re-run the
+    // before_start hook on every view switch / spawn, re-minting credentials.
     if let Some(info) = &sandbox_info {
         let mut guard = instances.write().await;
         if let Some(inst) = guard.iter_mut().find(|i| i.id == session_id_owned) {
@@ -92,6 +96,7 @@ pub async fn ensure_container_for_session(
                 if sb.container_id.is_none() {
                     sb.container_id = info.container_id.clone();
                 }
+                sb.before_start_env = info.before_start_env.clone();
             }
         }
     }
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -723,6 +723,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                 container_name,
                 extra_env: None,
                 custom_instruction: config.sandbox.custom_instruction.clone(),
+                before_start_env: Vec::new(),
             });
         }
     }
--- a/src/git/cleanup.rs
+++ b/src/git/cleanup.rs
@@ -688,6 +688,7 @@ mod tests {
             container_name: "aoe-sandbox-doesnotexist".to_string(),
             extra_env: None,
             custom_instruction: None,
+            before_start_env: Vec::new(),
         });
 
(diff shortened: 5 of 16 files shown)


njbrake:
@peteski22 PR is up and should be good! Could I have you check out the branch and give it a go to confirm this works the way that you wanted it to?

peteski22:
> @peteski22 PR is up and should be good! Could I have you check out the branch and give it a go to confirm this works the way that you wanted it to?

Works`*`! Wooo..

### CLI

Tested via the CLI:

```
TEST_VAR=foo " \
      ~/src/agent-of-empires/target/debug/aoe -p hooktest add --sandbox \
      --sandbox-image peteski22/dev-sandbox:0.1.0 --launch ~/tmp/foo
```

I had configured a profile `hooktest` with this in the `config.toml`:

```toml
[host_hooks]
before_start = ['set -e; tok=$(foo-get-token "$TEST_VAR"); echo "GH_TOKEN=$tok"']
```

---

### TUI

When it comes to the TUI UX, I think I might have hit a wrinkle in how you can supply a var to the session, as ideally I want to pass in different values to different sessions. 

I had planned to pass in `TEST_VAR` via the `Sandbox Configuration`, but: 

1. You can't pass the value, just the key
2. I think that will be handed over too late for a `before_start` hook anyway

So that when I create a session with `n`, is there (or could there be) a way to supply a per-session KEY=VALUE that reaches a host-side before_start hook? 

That's the one gap; everything else is working.

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Glad it works! Good catch on the per-session input gap.

Two things that may already help for the real token case: the before_start hook runs with the session's lifecycle vars in its environment, so `$AOE_PROJECT_PATH`, `$AOE_SESSION_ID`, `$AOE_SESSION_BRANCH`, `$AOE_PROFILE`, and `$AOE_TOOL` are readable with no extra config. For a per-repo GH token that covers it: `before_start = ['echo "GH_TOKEN=$(mint-tool "$AOE_PROJECT_PATH")"']` gives each session a token scoped to its own repo.

For an arbitrary per-session value like your `TEST_VAR`, you're right that `sandbox.environment` is the wrong channel: it is resolved for the container, not the host hook, and the per-session editor only takes a key today. I'd rather not widen this PR to cover it, since it matches what we agreed in the issue and you confirmed it works. I've written up the follow-up as #2120 (feed the session's resolved sandbox env into the host hook environment, plus let the per-session editor accept `KEY=VALUE`). Sound good?

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

@peteski22 that's the missing piece, thanks. You've nailed it: the host process env (`TEST_VAR=foo aoe add`) varies per CLI invocation but is fixed for the whole TUI process, so there was no per-session channel. #2121 (closes #2120) adds one: the session's sandbox env now flows into the `before_start` hook. Set `TEST_VAR=foo` in the new-session sandbox env list (it already takes `KEY=VALUE`), and the hook reads `$TEST_VAR`, different per session.

I also added `$AOE_REPO_SLUG` (the `owner/repo` from the origin remote) so `mint "$AOE_REPO_SLUG"` covers the org/repo case directly, and fixed a double-mint where the TUI re-ran the hook on first attach. Give #2121 a spin when you have a moment.

PR #2202, opened 2026-06-16
fix(tui): run session restart off the event loop

njbrake:
## Description

The TUI froze during a sandboxed session restart. A user's debug log dead-ended at `pulling image ghcr.io/agent-of-empires/aoe-sandbox:latest`, with an earlier 30s stall ending in `Failed to copy dir plugins: Permission denied`.

Root cause: pressing restart (`e` / `E` / `F5`) ran the whole start cascade synchronously on the TUI event-loop thread. `restart_selected_session` -> `restart_with_size` (`operations.rs`) does a blocking `docker pull`, the `before_start` host hook, and several `docker` shellouts. The pull had no timeout, so a stalled registry connection froze the entire UI indefinitely. Recovery and the manual image-update pull were already off-thread; the restart keybind was the one start path that skipped that treatment (its doc comment even claimed "the TUI event loop never blocks", which only covered the wake-up send-keys worker).

Three fixes:

1. **`pull_image` timeout** (`containers/runtime_base.rs`): spawn the child and `wait_with_timeout`, which drains stdout/stderr on threads and kills the child after a 600s deadline. A wedged pull can no longer block the caller forever.
2. **`copy_dir_recursive` robustness** (`session/container_config.rs`): break symlink cycles with a canonical-path visited set, and skip a single unreadable/dangling entry instead of aborting the whole copy (one Permission-denied file under `~/.claude/plugins` used to fail the entire sync after a 30s churn). Also gate the `copy_dirs` (plugins/skills) push on `!has_prior_data`, matching the general file copy, so a restart no longer re-copies the whole plugins tree.
3. **`RestartPoller`** (new `session/restart.rs` + `tui/restart_poller.rs`, mirrors `StopPoller`): the restart cascade runs on a worker thread. `restart_selected_session` flips the row to `Starting` and dispatches; `apply_restart_results` writes the post-cascade snapshot back, clears the in-flight marker, and surfaces failures as `Status::Error`. The UI thread never blocks on a restart again.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

This is a responsiveness/threading fix; there is no visual change to capture (the only user-visible difference is that the row shows `Starting` and the UI stays interactive during a restart instead of freezing).

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated tests (unit/regression rather than full e2e)

New regression tests:
- `wait_with_timeout_kills_child_that_outlives_deadline` / `wait_with_timeout_returns_output_for_fast_child` (timeout)
- `test_copy_dir_recursive_terminates_on_symlink_cycle`, `test_copy_dir_recursive_skips_bad_entry_inside_subdir`, `test_copy_dirs_skipped_when_prior_data` (copy)
- `perform_restart_preserves_session_id_and_returns_instance`, `restart_poller_channel_communication` (restart poller round-trip)

All existing `restart_selected_session_*` guard tests still pass. `cargo build`/`--features serve`, `cargo clippy --all-targets`, and `cargo fmt --check` are clean.

Notes for reviewers:
- Restarts are serialized on one worker (same as `StopPoller`); the 600s pull timeout bounds the worst case. Happy to switch to per-session workers if preferred.
- Gating `copy_dirs` on `has_prior_data` means a newly host-installed plugin reaches an existing sandbox only when it's recreated, consistent with how every other host file already behaves under the prior-data sentinel.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Investigation and implementation done with Claude Code via back-and-forth with @njbrake; the reasoning and decisions are his.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2202"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784204085&installation_id=139138837&pr_number=2202&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2202&signature=d48a83537becaaa962783d6abf918f6fe17ca10daff9d201979edb8f42737edc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## High-Level Summary

This PR fixes a critical TUI freeze when users trigger a session restart (e, E, or F5). The UI previously ran the entire restart “cascade” on the UI thread, and it could block indefinitely due to long-running/stuck operations (notably docker image pulls and hook execution). The fix keeps the UI responsive by moving restart work to a background worker and adding safety guards around the most failure-prone operations.

### 1. Docker pull is now time-bounded
- Added a hard timeout for `docker pull`-style operations.
- If the pull stalls beyond the deadline, the pull process is killed and the operation fails gracefully (rather than hanging the caller forever).
- Stdout/stderr draining is done safely to avoid hangs caused by pipe/grandchild behavior.

### 2. Session directory copying is more resilient and safer
- Recursive directory copy now:
  - Detects and breaks symlink cycles.
  - Becomes best-effort: unreadable/dangling/bad entries are skipped without aborting the whole sync.
  - Uses explicit “fatal vs soft” error classification (e.g., space/quota/readonly abort the operation; permission/unreadable entries are logged and skipped).
- Restart behavior is improved by avoiding re-copying seeded plugin/skill state when prior session data already exists (via a `projects/` sentinel), preventing unnecessary overwrites.

### 3. Restart execution moved off the UI thread
- Introduced a background restart worker (`RestartPoller`) and a shared restart pipeline (`session/restart.rs`).
- The UI now:
  - Marks the session row as “Starting” immediately.
  - Enqueues the restart request to the worker.
  - Applies results back to the UI during refresh.
- Restart in-flight safety was tightened:
  - Refuses duplicate restart presses for the same session while a restart is already running.
  - Refuses delete actions while restart is in flight (with an info dialog).
  - Handles worker disconnection so rows don’t get stuck in “Starting.”
- `before_start` hook execution is bounded to prevent indefinite hangs.

### 4. Updated UI refresh integration
- The main UI loop now regularly pulls restart completion results and forces redraw/update when restart state was applied.
- During shutdown, it drains any pending restart results before final persistence.

## Files Added/Updated
- **Added:** `src/session/restart.rs`, `src/tui/restart_poller.rs`
- **Updated:**  
  `src/containers/runtime_base.rs`, `src/session/container_config.rs`,  
  `src/session/mod.rs`, `src/tui/app.rs`, `src/tui/home/mod.rs`,  
  `src/tui/home/operations.rs`, `src/tui/mod.rs`, `src/tui/home/tests.rs`

## Benefits
- ✅ TUI no longer freezes during restart key presses
- ✅ Stuck docker pulls (e.g., registry stalls) can’t block the UI indefinitely
- ✅ File synchronization is more robust: one bad/unreadable entry won’t derail the whole copy
- ✅ Clearer user-facing behavior and safer concurrency (no duplicate restarts; delete blocked during restart)
- ✅ Regression tests added for timeout, in-flight restart safety, deletion refusal, and copy error discrimination

### DIFF
--- a/src/containers/runtime_base.rs
+++ b/src/containers/runtime_base.rs
@@ -40,8 +40,14 @@ fn wait_with_timeout(mut child: Child, timeout: Duration) -> std::io::Result<Opt
     let deadline = Instant::now() + timeout;
     loop {
         if let Some(status) = child.try_wait()? {
-            let stdout = orx.recv().unwrap_or_default();
-            let stderr = erx.recv().unwrap_or_default();
+            // The child exited, but if it spawned a grandchild that inherited
+            // the pipe, `read_to_end` (and thus an unbounded `recv`) would block
+            // forever. Cap the drain at the remaining deadline so the timeout
+            // guarantee holds even then; the exit status is already in hand.
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            let stdout = orx.recv_timeout(remaining).unwrap_or_default();
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            let stderr = erx.recv_timeout(remaining).unwrap_or_default();
             return Ok(Some(Output {
                 status,
                 stdout,
@@ -852,6 +858,30 @@ mod tests {
             start.elapsed() < Duration::from_secs(5),
             "wait should return promptly after the deadline, not block on the child"
         );
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn wait_with_timeout_bounds_drain_when_grandchild_holds_pipe() {
+        // The immediate child (sh) exits fast but backgrounds a `sleep` that
+        // inherits stdout, so the pipe never closes. The drain must still
+        // return by the deadline rather than blocking on read_to_end. `sleep 30`
+        // (>> the 5s assertion) ensures an unbounded recv would visibly fail.
+        let mut cmd = Command::new("sh");
+        cmd.arg("-c").arg("sleep 30 >&1 & printf done");
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        let child = cmd.spawn().unwrap();
+
+        let start = Instant::now();
+        let output = wait_with_timeout(child, Duration::from_millis(500))
+            .unwrap()
+            .expect("the sh child exits quickly, so an Output is produced");
+        assert!(
+            start.elapsed() < Duration::from_secs(5),
+            "drain must be bounded by the deadline even while the pipe stays open"
+        );
+        assert!(output.status.success());
     }
 
     #[test]
--- a/src/containers/runtime_base.rs
+++ b/src/containers/runtime_base.rs
@@ -1,6 +1,61 @@
 use super::container_interface::{docker_env_args, ContainerConfig};
 use super::error::{DockerError, Result};
-use std::process::Command;
+use std::io::Read;
+use std::process::{Child, Command, Output, Stdio};
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
+
+/// Upper bound on a single `docker pull`. `docker pull` has no overall timeout
+/// of its own, so a stalled registry connection blocks the caller forever
+/// (observed as the TUI freezing mid-pull on a sandbox restart). Sized
+/// generously so a genuinely large
+/// image on a slow link still completes; it only fires on a wedged pull.
+const PULL_TIMEOUT: Duration = Duration::from_secs(600);
+
+/// Wait for `child` to exit, killing it if it outlives `timeout`.
+///
+/// stdout/stderr are drained on dedicated threads so a full pipe buffer cannot
+/// wedge the child (and thus this wait) before the deadline. Returns `Ok(None)`
+/// when the timeout fired and the child was killed.
+fn wait_with_timeout(mut child: Child, timeout: Duration) -> std::io::Result<Option<Output>> {
+    let mut stdout_pipe = child.stdout.take();
+    let mut stderr_pipe = child.stderr.take();
+    let (otx, orx) = mpsc::channel();
+    let (etx, erx) = mpsc::channel();
+    std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        if let Some(ref mut p) = stdout_pipe {
+            let _ = p.read_to_end(&mut buf);
+        }
+        let _ = otx.send(buf);
+    });
+    std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        if let Some(ref mut p) = stderr_pipe {
+            let _ = p.read_to_end(&mut buf);
+        }
+        let _ = etx.send(buf);
+    });
+
+    let deadline = Instant::now() + timeout;
+    loop {
+        if let Some(status) = child.try_wait()? {
+            let stdout = orx.recv().unwrap_or_default();
+            let stderr = erx.recv().unwrap_or_default();
+            return Ok(Some(Output {
+                status,
+                stdout,
+                stderr,
+            }));
+        }
+        if Instant::now() >= deadline {
+            let _ = child.kill();
+            let _ = child.wait();
+            return Ok(None);
+        }
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
 
 /// Shared implementation for container runtimes.
 ///
@@ -102,9 +157,31 @@ impl RuntimeBase {
         let mut cmd = self.command();
         cmd.args(self.pull_prefix);
         cmd.arg(image);
-        let start = std::time::Instant::now();
+        cmd.stdin(Stdio::null());
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        let start = Instant::now();
         tracing::info!(target: "containers.image", runtime = %self.name, %image, "pulling image");
-        let output = cmd.output()?;
+        let child = cmd.spawn()?;
+        let output = match wait_with_timeout(child, PULL_TIMEOUT)? {
+            Some(output) => output,
+            None => {
+                let dur_ms = start.elapsed().as_millis() as u64;
+                tracing::warn!(
+                    target: "containers.image",
+                    runtime = %self.name,
+                    %image,
+                    duration_ms = dur_ms,
+                    timeout_s = PULL_TIMEOUT.as_secs(),
+                    "image pull timed out; killed",
+                );
+                return Err(DockerError::ImageNotFound(format!(
+                    "{}: pull timed out after {}s",
+                    image,
+                    PULL_TIMEOUT.as_secs()
+                )));
+            }
+        };
         let dur_ms = start.elapsed().as_millis() as u64;
 
         if !output.status.success() {
@@ -751,5 +828,45 @@ mod tests {
         const { assert!(RuntimeBase::DOCKER.supports_named_volumes) };
         const { assert!(RuntimeBase::PODMAN.supports_named_volumes) };
         const { assert!(!RuntimeBase::APPLE_CONTAINER.supports_named_volumes) };
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn wait_with_timeout_kills_child_that_outlives_deadline() {
+        // A `docker pull` that wedges on the network has no timeout of its own
+        // and blocked the caller forever. `sleep 30` stands in for
+        // the wedged child; the timeout must fire and kill it.
+        let mut cmd = Command::new("sleep");
+        cmd.arg("30");
+        cmd.stdout(Stdio::null());
+        cmd.stderr(Stdio::null());
+        let child = cmd.spawn().unwrap();
+
+        let start = Instant::now();
+        let result = wait_with_timeout(child, Duration::from_millis(300)).unwrap();
+        assert!(
+            result.is_none(),
+            "expected the timeout to fire and kill the child"
+        );
+        assert!(
+            start.elapsed() < Duration::from_secs(5),
+            "wait should return promptly after the deadline, not block on the child"
+        );
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn wait_with_timeout_returns_output_for_fast_child() {
+        let mut cmd = Command::new("printf");
+        cmd.arg("hello");
+        cmd.stdout(Stdio::piped());
+        cmd.stderr(Stdio::piped());
+        let child = cmd.spawn().unwrap();
+
+        let output = wait_with_timeout(child, Duration::from_secs(10))
+            .unwrap()
+            .expect("fast child should complete before the timeout");
+        assert!(output.status.success());
+        assert_eq!(output.stdout, b"hello");
     }
 }
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -500,7 +500,18 @@ fn copy_dir_recursive_inner(
             }
         };
         if metadata.is_dir() {
-            copy_dir_recursive_inner(&entry.path(), &target, visited)?;
+            // Catch-and-skip rather than `?`: an unreadable subdirectory (e.g.
+            // a Permission-denied dir under ~/.claude/plugins) must not abort
+            // the whole copy, matching how the metadata and file-copy errors
+            // above are handled.
+            if let Err(e) = copy_dir_recursive_inner(&entry.path(), &target, visited) {
+                tracing::warn!(
+                    target: "session.profile",
+                    "skipping dir {}: {}",
+                    entry.path().display(),
+                    e
+                );
+            }
         } else if let Err(e) = std::fs::copy(entry.path(), &target) {
             tracing::warn!(
                 target: "session.profile",
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -330,7 +330,13 @@ fn sync_agent_config(
         };
 
         if metadata.is_dir() {
-            if copy_dirs.iter().any(|&d| d == name_str.as_ref()) {
+            // copy_dirs (e.g. plugins, skills) are host -> sandbox pushes. Like
+            // the general file copy below, skip them once the sandbox has prior
+            // session data: re-copying a large tree (plugins can hold full git
+            // clones) on every restart stalled startup for tens of seconds and
+            // would clobber container-side changes. A fresh sandbox still gets
+            // them on its first launch.
+            if !has_prior_data && copy_dirs.iter().any(|&d| d == name_str.as_ref()) {
                 let dest = sandbox_dir.join(&name);
                 if let Err(e) = copy_dir_recursive(&entry.path(), &dest) {
                     tracing::warn!(target: "session.profile", "Failed to copy dir {}: {}", name_str, e);
@@ -448,16 +454,60 @@ fn rewrite_plugin_value_paths(
 
 /// Recursively copy a directory tree, following symlinks.
 fn copy_dir_recursive(src: &Path, dest: &Path) -> Result<()> {
+    let mut visited = std::collections::HashSet::new();
+    copy_dir_recursive_inner(src, dest, &mut visited)
+}
+
+fn copy_dir_recursive_inner(
+    src: &Path,
+    dest: &Path,
+    visited: &mut std::collections::HashSet<PathBuf>,
+) -> Result<()> {
+    // Break symlink cycles. We follow symlinks (a legitimately symlinked config
+    // dir should be copied), but a link that points back up its own tree would
+    // otherwise recurse forever; a real cycle under ~/.claude/plugins churned
+    // for 30s before aborting. Keying on the canonical (symlink-
+    // resolved) source path stops the second visit.
+    if let Ok(real) = std::fs::canonicalize(src) {
+        if !visited.insert(real) {
+            tracing::warn!(
+                target: "session.profile",
+                "skipping already-visited dir (symlink cycle?): {}",
+                src.display()
+            );
+            return Ok(());
+        }
+    }
     std::fs::create_dir_all(dest)?;
     for entry in std::fs::read_dir(src)? {
         let entry = entry?;
         let target = dest.join(entry.file_name());
-        // Follow symlinks so symlinked dirs/files are handled correctly.
-        let metadata = std::fs::metadata(entry.path())?;
+        // Follow symlinks so symlinked dirs/files are handled correctly; the
+        // visited-set guard above stops a cycle from looping forever. A single
+        // unreadable or dangling entry is skipped rather than aborting the whole
+        // copy: one Permission-denied file under ~/.claude/plugins used to fail
+        // the entire sync.
+        let metadata = match std::fs::metadata(entry.path()) {
+            Ok(m) => m,
+            Err(e) => {
+                tracing::warn!(
+                    target: "session.profile",
+                    "skipping {}: {}",
+                    entry.path().display(),
+                    e
+                );
+                continue;
+            }
+        };
         if metadata.is_dir() {
-            copy_dir_recursive(&entry.path(), &target)?;
-        } else {
-            std::fs::copy(entry.path(), &target)?;
+            copy_dir_recursive_inner(&entry.path(), &target, visited)?;
+        } else if let Err(e) = std::fs::copy(entry.path(), &target) {
+            tracing::warn!(
+                target: "session.profile",
+                "skipping file {}: {}",
+                entry.path().display(),
+                e
+            );
         }
     }
     Ok(())
@@ -2556,6 +2606,64 @@ mod tests {
             fs::read_to_string(sandbox.join("settings.json")).unwrap(),
             r#"{"theme":"dark"}"#,
             "container-side settings must not be overwritten when projects/ sentinel exists"
+        );
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn test_copy_dir_recursive_terminates_on_symlink_cycle() {
+        let dir = TempDir::new().unwrap();
+        let src = dir.path().join("src");
+        fs::create_dir_all(src.join("sub")).unwrap();
+        fs::write(src.join("sub").join("file.txt"), "data").unwrap();
+        // Cycle: src/sub/loop -> src. The old code followed it and recursed
+        // forever; the visited-set guard must break it.
+        std::os::unix::fs::symlink(&src, src.join("sub").join("loop")).unwrap();
+
+        let dest = dir.path().join("dest");
+        // Must return rather than infinite-loop / stack-overflow.
+        copy_dir_recursive(&src, &dest).unwrap();
+        assert_eq!(
+            fs::read_to_string(dest.join("sub").join("file.txt")).unwrap(),
+            "data"
+        );
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn test_copy_dir_recursive_skips_bad_entry_inside_subdir() {
+        let dir = TempDir::new().unwrap();
+        let src = dir.path().join("src");
+        fs::create_dir_all(&src).unwrap();
+        fs::write(src.join("good.txt"), "good").unwrap();
+        // Dangling symlink inside the copied tree. The old code propagated the
+        // stat error with `?` and aborted the whole copy (the log's "Failed to
+        // copy dir plugins: Permission denied"); now a single bad entry is
+        // skipped and the rest still copies.
+        std::os::unix::fs::symlink("/nonexistent/target", src.join("dangling")).unwrap();
+
+        let dest = dir.path().join("dest");
+        copy_dir_recursive(&src, &dest).unwrap();
+        assert_eq!(fs::read_to_string(dest.join("good.txt")).unwrap(), "good");
+        assert!(!dest.join("dangling").exists());
+    }
+
+    #[test]
+    fn test_copy_dirs_skipped_when_prior_data() {
+        let dir = TempDir::new().unwrap();
+        let host = dir.path().join("host");
+        fs::create_dir_all(host.join("plugins")).unwrap();
+        fs::write(host.join("plugins").join("p.txt"), "host-plugin").unwrap();
+        let sandbox = dir.path().join("sandbox");
+
+        // Prior container session sentinel.
+        fs::create_dir_all(sandbox.join("projects")).unwrap();
+
+        sync_agent_config(&host, &sandbox, &[], &[], &["plugins"], &[]).unwrap();
+        assert!(
+            !sandbox.join("plugins").exists(),
+            "copy_dirs must be skipped once the sandbox has prior session data, \
+             so a restart no longer re-copies the whole plugins tree"
         );
     }
 
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -458,6 +458,18 @@ fn copy_dir_recursive(src: &Path, dest: &Path) -> Result<()> {
     copy_dir_recursive_inner(src, dest, &mut visited)
 }
 
+/// Whether an I/O error during the copy means the destination filesystem can no
+/// longer accept writes, in which case continuing would silently produce a
+/// partial copy (a sandbox missing arbitrary files, reported as success). Those
+/// must abort the whole copy. Everything else (a single unreadable or dangling
+/// source entry) is skipped best-effort.
+fn is_fatal_copy_error(e: &std::io::Error) -> bool {
+    // ENOSPC (no space), EROFS (read-only fs), EDQUOT (quota). EDQUOT differs by
+    // platform (122 on Linux, 69 on macOS). raw_os_error covers all three more
+    // portably than the (partly unstable) ErrorKind variants.
+    matches!(e.raw_os_error(), Some(28) | Some(30) | Some(69) | Some(122))
+}
+
 fn copy_dir_recursive_inner(
     src: &Path,
     dest: &Path,
@@ -466,21 +478,66 @@ fn copy_dir_recursive_inner(
     // Break symlink cycles. We follow symlinks (a legitimately symlinked config
     // dir should be copied), but a link that points back up its own tree would
     // otherwise recurse forever; a real cycle under ~/.claude/plugins churned
-    // for 30s before aborting. Keying on the canonical (symlink-
-    // resolved) source path stops the second visit.
-    if let Ok(real) = std::fs::canonicalize(src) {
-        if !visited.insert(real) {
+    // for 30s before aborting. Keying on the canonical (symlink-resolved) source
+    // path stops the second visit. A canonicalize failure (e.g. ELOOP) is
+    // exactly when we most need the guard, so skip the dir rather than descend
+    // blindly.
+    match std::fs::canonicalize(src) {
+        Ok(real) => {
+            if !visited.insert(real) {
+                tracing::warn!(
+                    target: "session.profile",
+                    "skipping already-visited dir (symlink cycle?): {}",
+                    src.display()
+                );
+                return Ok(());
+            }
+        }
+        Err(e) => {
             tracing::warn!(
                 target: "session.profile",
-                "skipping already-visited dir (symlink cycle?): {}",
-                src.display()
+                "skipping dir (cannot resolve, possible symlink loop): {}: {}",
+                src.display(),
+                e
             );
             return Ok(());
         }
     }
-    std::fs::create_dir_all(dest)?;
-    for entry in std::fs::read_dir(src)? {
-        let entry = entry?;
+    // create_dir_all / read_dir failures: a filesystem that can't accept writes
+    // is fatal (silent partial copy); anything else means we just skip this
+    // subtree. This keeps the fail-soft behavior errno-aware and consistent.
+    if let Err(e) = std::fs::create_dir_all(dest) {
+        if is_fatal_copy_error(&e) {
+            return Err(e).with_context(|| format!("creating {}", dest.display()));
+        }
+        tracing::warn!(
+            target: "session.profile",
+            "skipping dir (cannot create destination): {}: {}",
+            dest.display(),
+            e
+        );
+        return Ok(());
+    }
+    let entries = match std::fs::read_dir(src) {
+        Ok(entries) => entries,
+        Err(e) => {
+            tracing::warn!(
+                target: "session.profile",
+                "skipping dir (cannot read): {}: {}",
+                src.display(),
+                e
+            );
+            return Ok(());
+        }
+    };
+    for entry in entries {
+        let entry = match entry {
+            Ok(entry) => entry,
+            Err(e) => {
+                tracing::warn!(target: "session.profile", "skipping entry in {}: {}", src.display(), e);
+                continue;
+            }
+        };
         let target = dest.join(entry.file_name());
         // Follow symlinks so symlinked dirs/files are handled correctly; the
         // visited-set guard above stops a cycle from looping forever. A single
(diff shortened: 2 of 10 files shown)


njbrake:
@peteski22 I think this PR should address your hang issue: would you be able to give it a try?

src/tui/restart_poller.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    pub fn request_restart(&self, request: RestartRequest) {
+        if let Err(e) = self.request_tx.send(request) {
+            tracing::warn!(target: "tui.restart_poller", error = %e, "restart request dropped; worker thread unavailable");
+        }
+    }
+
+    pub fn try_recv_result(&self) -> Option<RestartResult> {
+        self.result_rx.try_recv().ok()
```
jerome-benoit: **[BLOCKING]** `try_recv_result` collapses `mpsc::TryRecvError::Empty` and `Disconnected` into `Option` via `.ok()`. If `perform_restart` panics inside the worker, `result_tx` is dropped → every subsequent `try_recv` returns `Err(Disconnected)` → `apply_restart_results` never observes the failure, never removes the id from `restart_in_flight`, never flips the row off `Status::Starting`. `pollable_instances` (`src/tui/home/mod.rs:2354`) filters that row out of every status poll, so the StatusPoller cannot rescue it. Permanent stuck-Starting row with no recovery short of restarting the TUI.

The recovery path solves the same problem at `src/tui/home/mod.rs:2784-2799` (defensive `recovery_in_flight.clear()` on `Disconnected`). Suggested fix: replace `try_recv_result()` with a `Result<Option<RestartResult>, Disconnected>` API; have `apply_restart_results` clear `restart_in_flight` and transition affected rows to `Status::Error` on disconnect, mirroring `apply_recovery_updates`. Add a regression test that drops `result_tx` and asserts the next call observes `Disconnected`.

src/tui/home/mod.rs
```diff
@@ -2786,56 +2799,107 @@ impl HomeView {
...
+
+            self.restart_in_flight.remove(&session_id);
+
+            match outcome {
+                Ok(_) => {}
+                Err(e) => {
+                    tracing::warn!(
+                        target: "tui.home",
+                        id = %session_id,
+                        error = %e,
```
jerome-benoit: **[BLOCKING]** Before this change, a cascade error propagated `?` out of `restart_selected_session`, the input handler caught it (`src/tui/home/input.rs:1875-1887`), and surfaced an `InfoDialog::new("Restart Failed", ...)` modal. With the new dispatch, `restart_selected_session` returns `Ok(())` immediately after enqueueing; the only way it returns `Err` is via synchronous bookkeeping failures. A real cascade failure arrives via `apply_restart_results`, which only logs `tracing::warn!` and sets `instance.status = Status::Error` + `last_error = Some(e)`. There is no `InfoDialog` spawn from `apply_restart_results`.

The user explicitly initiated the restart from a confirmation dialog and receives only a row-colour change and an entry in `last_error` (rendered nowhere prominent during a normal session). The pre-existing comment in `input.rs:1878-1881` ("Surface the restart error to the user via the InfoDialog rather than only the debug log; the user explicitly initiated this action and needs to know it failed") explicitly motivates this surface. Suggested fix: in the error branch here, raise the same `InfoDialog` (or signal back from `apply_restart_results` so `App::run` can spawn it). Carry the human-readable error text from `instance.last_error`.

src/session/restart.rs
```diff
@@ -0,0 +1,126 @@
...
+    let RestartRequest {
+        session_id,
+        mut instance,
+        size,
+        wake_message,
+    } = request;
+
+    let title = instance.title.clone();
+    let tool = instance.tool.clone();
+    let outcome = instance.restart_with_size(size).map_err(|e| e.to_string());
```
jerome-benoit: **[BLOCKING]** `run_before_start_hooks` honors `current_hook_timeout()`, a thread-local set only by `HookTimeoutScope`. The startup-recovery path (`src/session/recovery.rs`) installs a 30 s scope per worker. The new `RestartPoller` worker installs no such scope, so on the worker thread `current_hook_timeout()` returns `None` and `run_before_start_hooks` falls through to plain `command.output()` with no kill timer. A user `before_start` script that hangs (network probe, broken ssh-agent, frozen credential helper) hangs the worker thread forever. The `restart_in_flight` marker for that session never clears; the row stays in `Status::Starting` indefinitely; subsequent restart requests for OTHER sessions queue behind it and never run (single serial worker).

The PR's own comment in `operations.rs:432-434` enumerates `before_start` as something that "can block for seconds", but the worker has no defense against it blocking forever. The earlier inline path on the TUI thread had the same gap; this PR moves the hang from the TUI thread to the worker thread. The TUI keeps responsive, which is the stated goal, but the worker leaks and one wedged session takes the entire restart subsystem down.

Suggested fix: wrap `perform_restart` body in a `HookTimeoutScope`:
```rust
let _scope = crate::session::recovery::HookTimeoutScope::new(
    crate::session::recovery::recovery_hook_timeout(),
);
```
Brings parity with the recovery worker.

src/tui/home/operations.rs
```diff
@@ -431,79 +429,43 @@ impl HomeView {
...
-                    if std::time::Instant::now() >= deadline {
-                        break;
-                    }
-                    std::thread::sleep(std::time::Duration::from_millis(50));
-                }
+        // Resolve the wake message on the main thread (config access). Empty is
+        // the documented opt-out; the worker skips the wake-up then.
+        let wake_message = crate::session::resolve_config(&instance.source_profile)
+            .map(|c| c.session.restart_wake_message.clone())
+            .unwrap_or_else(|_| "wake up: pick up what you were doing".to_string());
```
jerome-benoit, line 470: **[BLOCKING]** This `restart_in_flight.insert` site is the natural place to also enforce single-flight: today the cascade runs off the event loop while the 1500 ms `restart_cooldown_at` debounces only against keyboard auto-repeat. The UI stays responsive while the worker runs (10 s+ for a cold pull). `Status::Starting` is not in the skip list at line 318, so a second `e` at t=2 s falls through, `restart_in_flight.insert` is a no-op (already present), and a duplicate `RestartRequest` is enqueued. The serial worker processes them back-to-back: cascade #2 runs against an instance that just restarted, killing the container cascade #1 just created and leaving a half-started session.

The author's own comment ("Holding `e` or pressing it twice fast otherwise races overlapping `restart_with_size` calls") holds at multi-second spacing, not just 100 ms. Suggested fix: add `if self.restart_in_flight.contains(&id) { return Ok(()); }` as the first check in `restart_selected_session` (immediately after the `selected_session` early-return at line 311). The cooldown can stay as belt-and-braces for keyboard auto-repeat.

jerome-benoit, line 470: **[BLOCKING]** Cross-axis with the `restart_in_flight` set introduced here: `delete_selected` (line 472+, unchanged in this PR) does not consult `restart_in_flight`. With a restart in flight (worker mid-`docker create` / `docker start`), pressing `d` runs concurrent docker commands against the same container/session. Outcomes are non-deterministic: orphan tmux session, half-cleaned worktree, deletion succeeding before restart finishes (then `apply_restart_results.find(|i| i.id == ...)` correctly no-ops, but the cascade has already side-effected docker), or restart succeeding after delete (orphan container with no UI handle).

The synchronous cascade made this race impossible because the UI thread could not accept a delete while restarting; the freeze-prevention removed an implicit lock the rest of the code relied on. `delete_group_with_sessions` has the same gap. Suggested fix: refuse-or-defer at the top of `delete_selected` / `delete_group_with_sessions` when `restart_in_flight.contains(&id)`. Either show an `InfoDialog` ("can't delete while restarting") or queue the delete as a follow-up after `apply_restart_results` clears the marker. Inline anchored here because `delete_selected` itself is not in the diff.

src/session/container_config.rs
```diff
@@ -448,16 +454,71 @@ fn rewrite_plugin_value_paths(
...
+            // above are handled.
+            if let Err(e) = copy_dir_recursive_inner(&entry.path(), &target, visited) {
+                tracing::warn!(
+                    target: "session.profile",
+                    "skipping dir {}: {}",
+                    entry.path().display(),
+                    e
+                );
+            }
+        } else if let Err(e) = std::fs::copy(entry.path(), &target) {
```
jerome-benoit, line 565: **[BLOCKING]** The blanket `match e { _ => warn + continue }` treats every errno identically. EACCES (one bad file) and ENOENT (dangling symlink) genuinely should skip. ENOSPC, EDQUOT, and EROFS at the destination are not "this one entry is broken"; they mean the destination filesystem cannot accept any more writes. Continuing the loop produces a silent partial-copy success: `copy_dir_recursive` returns `Ok(())`, `sync_agent_config` returns `Ok(())`, the container is created and bind-mounts a sandbox missing arbitrary files. The user sees a working session that is missing plugins, settings, or seed credentials with no error.

Suggested fix: inspect `e.raw_os_error()` (or `e.kind()`) and propagate destination-side fatal errors:
```rust
match std::fs::copy(&entry.path(), &target) {
    Err(e) if matches!(e.raw_os_error(), Some(libc::ENOSPC | libc::EDQUOT | libc::EROFS)) => return Err(e.into()),
    Err(e) => tracing::warn!(target: "session.profile", "skipping file {}: {}", entry.path().display(), e),
    Ok(_) => {}
}
```

src/containers/runtime_base.rs
```diff
@@ -1,6 +1,67 @@
...
+    let deadline = Instant::now() + timeout;
+    loop {
+        if let Some(status) = child.try_wait()? {
+            // The child exited, but if it spawned a grandchild that inherited
+            // the pipe, `read_to_end` (and thus an unbounded `recv`) would block
+            // forever. Cap the drain at the remaining deadline so the timeout
+            // guarantee holds even then; the exit status is already in hand.
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            let stdout = orx.recv_timeout(remaining).unwrap_or_default();
+            let remaining = deadline.saturating_duration_since(Instant::now());
```
jerome-benoit, line 49: **[INCONSISTENT]** AGENTS.md: "No `let _ =` outside tests." Six occurrences in `wait_with_timeout` at lines 28, 30, 35, 37, 58, 59 (only the recv→recv_timeout pair right above is touched by `b8f9691f`; the other six remain): two `read_to_end`, two channel sends, plus `child.kill()` and `child.wait()` in the timeout branch. The rest of the codebase uses `if let Err(e) = ... { tracing::warn!(...) }` for similar patterns (cf. `restart.rs:89-91` in this same PR for the wake worker). Inline anchored here on the new `recv_timeout` line because the `let _ = child.kill()` at line 58 is below the diff hunk; the suggested fix applies to all six original sites.

Suggested replacement: `if let Err(e) = ... { tracing::trace!(...) }` for the channel-send / read_to_end sites (close-on-drop is benign), and `tracing::warn!` for the `child.kill()` / `child.wait()` pair in the timeout branch (the child was supposed to be wedged; errors there carry diagnostic value).

src/session/container_config.rs
```diff
@@ -448,16 +454,71 @@ fn rewrite_plugin_value_paths(
...
+    src: &Path,
+    dest: &Path,
+    visited: &mut std::collections::HashSet<PathBuf>,
+) -> Result<()> {
+    // Break symlink cycles. We follow symlinks (a legitimately symlinked config
+    // dir should be copied), but a link that points back up its own tree would
+    // otherwise recurse forever; a real cycle under ~/.claude/plugins churned
+    // for 30s before aborting. Keying on the canonical (symlink-
+    // resolved) source path stops the second visit.
+    if let Ok(real) = std::fs::canonicalize(src) {
```
jerome-benoit: **[INCONSISTENT]** `if let Ok(real) = std::fs::canonicalize(src) { ... }` swallows the Err arm. When `canonicalize()` fails (broken symlink resolved part-way, EACCES on a parent component, ELOOP on a too-deep symlink chain), the fallthrough enters the directory copy without inserting anything into the `visited` set. Two implications: (a) the cycle-detection guarantee is not actually established for any subtree where canonicalize fails, and ELOOP on canonicalize is exactly the case where you most need cycle detection; (b) the error is silently swallowed, against AGENTS.md "tracing::warn! on swallowed errors". The two adjacent error paths in this same function (`:493-500`, `:505-510`) do warn; this one does not.

Suggested fix:
```rust
match std::fs::canonicalize(src) {
    Ok(real) => {
        if !visited.insert(real) {
            tracing::warn!(target: "session.profile",
                "skipping already-visited dir (symlink cycle?): {}", src.display());
            return Ok(());
        }
    }
    Err(e) => {
        tracing::warn!(target: "session.profile",
            "skipping dir (canonicalize failed, possible loop): {}: {}", src.display(), e);
        return Ok(());
    }
}
```

src/session/container_config.rs
```diff
@@ -448,16 +454,71 @@ fn rewrite_plugin_value_paths(
...
+            tracing::warn!(
+                target: "session.profile",
+                "skipping already-visited dir (symlink cycle?): {}",
+                src.display()
+            );
+            return Ok(());
+        }
+    }
     std::fs::create_dir_all(dest)?;
     for entry in std::fs::read_dir(src)? {
```
jerome-benoit: **[INCONSISTENT]** The PR added a per-entry skip for `metadata` (lines 490-501), but `read_dir(src)?` (line 482) and `entry?` (line 483) still propagate. When a `dirent` syscall mid-iteration returns EIO (failing disk, NFS hiccup), the partial result so far is in `dest`, the rest is silently dropped, and the call returns `Err`. With `b8f9691f` now wrapping the recursive call in warn-and-continue, this is the last propagating `?` in the loop body and contradicts the PR narrative ("skip a single bad entry rather than abort").

Suggested fix: warn-and-skip the broken `dirent`:
```rust
let entry = match entry { Ok(e) => e, Err(e) => {
    tracing::warn!(target: "session.profile", "skipping bad dirent: {}", e);
    continue;
}};
```

src/tui/restart_poller.rs
```diff
@@ -0,0 +1,111 @@
...
+    result_rx: mpsc::Receiver<RestartResult>,
+    _handle: thread::JoinHandle<()>,
+}
+
+impl RestartPoller {
+    pub fn new() -> Self {
+        let (request_tx, request_rx) = mpsc::channel::<RestartRequest>();
+        let (result_tx, result_rx) = mpsc::channel::<RestartResult>();
+
+        let handle = thread::spawn(move || {
```
jerome-benoit: **[INCONSISTENT]** `thread::spawn(move || { Self::restart_loop(...) })` lacks a `Builder::name`. StopPoller has the same pattern (`src/tui/stop_poller.rs:26`) so this PR matches it; but the wake worker right next door in this same PR (`src/session/restart.rs:62-63`) does name itself via `Builder::new().name(format!("aoe-restart-wake/{}", session_id))`. A wedged restart poller thread shows up in `gstack` / `lldb thread list` as `Thread N: <unnamed>` with no role hint.

Suggested fix: `thread::Builder::new().name("aoe-restart-poller".into()).spawn(...)`. The same change is worth making in `stop_poller.rs:26` as a sibling follow-up.

src/tui/restart_poller.rs
```diff
@@ -0,0 +1,111 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::session::Instance;
+    use std::time::Duration;
+
+    #[test]
+    #[serial_test::serial]
+    fn restart_poller_channel_communication() {
```
jerome-benoit, line 83: **[WARN]** Existing `restart_selected_session_*` guard tests pass (verified) but they were written when the cascade was synchronous and only observe `restart_cooldown_at` bookkeeping. They do not cover the new state machine: `restart_in_flight.insert`, `Status::Starting` flip, `pollable_instances` filtering, double-press guard, or the `apply_restart_results` writeback. The new `restart_poller_channel_communication` covers the happy round-trip; there is no test for (a) worker thread dropping `result_tx` (simulated panic / disconnect), (b) sending a request after worker shutdown, (c) `restart_in_flight` recovery semantics. `perform_restart_preserves_session_id_and_returns_instance` exercises the happy round-trip but never asserts `outcome.is_ok()` / `is_err()`.

Suggested additions:
- `restart_selected_session_skips_when_already_in_flight` (drives finding 2 fix)
- `apply_restart_results_clears_in_flight_on_disconnect` (drives finding 1 fix)
- `delete_selected_refused_during_restart` (drives finding 3 fix)
- `pollable_instances_excludes_restart_in_flight`
- `copy_dir_recursive_propagates_enospc` (drives finding 7 fix)

jerome-benoit (changes requested):
### Executive summary

The PR's stated goal is met: pressing `e` / `E` / `F5` keeps the TUI responsive against a wedged registry pull, because the restart cascade runs on a worker thread off the event loop. The three sub-fixes (`pull_image` 600s timeout, `copy_dir_recursive` symlink-cycle detection + per-entry skip, `RestartPoller` mirroring `StopPoller`) are directionally correct.

Hold for **6 reachable BLOCKING** items (one of the seven I originally identified, the unreadable-subdir abort, was resolved by commit `b8f9691f` while the audit was in progress; thank you for the quick turnaround).

| Severity | Count |
|---|---|
| **BLOCKING** | 6 (was 7; one resolved by commit `b8f9691f`) |
| **INCONSISTENT** | 7 |
| **WARN** | ~14 |
| **DEFERRED** | 2 (PR scope creep, TOCTOU mid-copy) |

All 6 BLOCKING items have small fixes (single-line guards, one match arm, one `InfoDialog` spawn, one `HookTimeoutScope` wrap, one match-arm errno discrimination).

### What blocks merge

#### State machine (TUI side)

- **Worker panic + `Disconnected` leak `restart_in_flight` permanently** (`src/tui/restart_poller.rs:56-58`, `src/tui/home/mod.rs:2865-2905`). `try_recv_result()` collapses `mpsc::TryRecvError::Empty` and `Disconnected` into `Option` via `.ok()`. If `perform_restart` panics, the worker drops `result_tx`, every subsequent `try_recv` returns `Err(Disconnected)`, the in-flight set never clears, the row stays on `Status::Starting` indefinitely, and `pollable_instances` (line 2354) filters that row out of every status poll so the StatusPoller cannot rescue it. The recovery path solves the same problem at `src/tui/home/mod.rs:2784-2799` (defensive `recovery_in_flight.clear()` on `Disconnected`); the new code drifts from that precedent.
- **Double-press race: 1.5 s cooldown insufficient for 10 s+ cascades** (`src/tui/home/operations.rs:336-342, 462-468`). The cascade runs off the event loop while the 1500 ms `restart_cooldown_at` debounces only against keyboard auto-repeat. `Status::Starting` is not in the skip list, so a second `e` at t=2 s falls through, `restart_in_flight.insert` is a no-op, and the worker enqueues a duplicate `RestartRequest`. The serial worker processes them back-to-back: cascade #2 runs against an instance that just restarted, killing the container cascade #1 just created.
- **`delete_selected` not guarded by `restart_in_flight`** (`src/tui/home/operations.rs:472-493`). With a restart in flight (worker mid-`docker create` / `docker start`), pressing `d` runs concurrent docker commands against the same container/session. Outcomes are non-deterministic: orphan tmux session, half-cleaned worktree, restart finishing after delete (resurrected resources with no UI handle). `delete_group_with_sessions` has the same gap. The synchronous cascade made this race impossible because the UI thread could not accept a delete while restarting; the freeze-prevention removed an implicit lock the rest of the code relied on.
- **Cascade-failure regression: stops raising `InfoDialog`** (`src/tui/home/mod.rs:2880-2890`, vs `src/tui/home/input.rs:1875-1887`). When the cascade fails, `restart_selected_session` returns `Ok(())` (the only `Err` paths remaining are synchronous bookkeeping failures); the real failure arrives via `apply_restart_results`, which only logs `tracing::warn!` and flips `instance.status = Status::Error`. The `InfoDialog::new("Restart Failed", ...)` modal at `input.rs:1875-1887` is unreachable for cascade failures because the input handler only sees the synchronous `Err`. The user explicitly initiated the restart from a confirmation dialog and now sees only a row colour change.

#### Worker safety (cascade side)

- **`before_start` host hook hangs the worker forever** (`src/session/restart.rs:42` → `src/session/instance.rs:2547` → `src/session/repo_config.rs:1206-1211`). `run_before_start_hooks` honors `current_hook_timeout()`, a thread-local set only by `HookTimeoutScope`. The startup-recovery path installs a 30 s scope per worker. The new `RestartPoller` worker installs no such scope, so on the worker thread `current_hook_timeout()` returns `None` and `run_before_start_hooks` falls through to plain `command.output()` with no kill timer. A user `before_start` script that hangs hangs the worker thread forever; the serial worker means one wedged session takes the entire restart subsystem down with it.

#### Filesystem (copy_dir_recursive)

- **Unreadable subdirectory still aborts the whole copy** (`src/session/container_config.rs:503`). RESOLVED by commit `b8f9691f` (the new code wraps the recursive call with `if let Err(e) = ... tracing::warn!`).
- **ENOSPC / EROFS / EDQUOT silently treated as a successful partial copy** (`src/session/container_config.rs:515`). The blanket `match e { _ => warn + continue }` treats every errno identically. EACCES (one bad file) and ENOENT (dangling symlink) genuinely should skip. ENOSPC / EROFS / EDQUOT at the destination mean the filesystem cannot accept any more writes; continuing the loop produces a silent partial-copy success: `copy_dir_recursive` returns `Ok(())`, `sync_agent_config` returns `Ok(())`, the container is created and bind-mounts a sandbox missing arbitrary files. The user sees a working session that is missing plugins, settings, or seed credentials with no error. Cumulative: `perform_restart` returns `Ok(StartOutcome)`, the row goes green.

### What needs fixing in this PR but does not block merge (consistency)

- **`let _ =` in production code** (`src/containers/runtime_base.rs:28, 30, 35, 37, 58, 59`). AGENTS.md: "No `let _ =` outside tests." Six occurrences in `wait_with_timeout`. The rest of the codebase uses `if let Err(e) = ... { tracing::warn!(...) }` for the same pattern (cf. the wake worker at `restart.rs:89-91` in this same PR).
- **`canonicalize()` failure silently swallowed** (`src/session/container_config.rs:471-480`). The cycle-detection guard skips fall-through when `canonicalize` itself fails. ELOOP on canonicalize is exactly the case where you most need cycle detection. AGENTS.md "tracing::warn! on swallowed errors" is also bypassed.
- **`read_dir(src)?` and `entry?` mid-iteration also propagate** (`src/session/container_config.rs:482-483`). With `b8f9691f` now wrapping the recursive call, these are the last propagating `?` in the loop body and contradict the PR narrative.
- **`perform_restart` reports success on partial copy** (`src/session/restart.rs:42-55`). With the PR's fail-soft semantics, all individual copy failures become `tracing::warn!` and the cascade returns `Ok(StartOutcome)`. Combined with the ENOSPC silent-success above, this lets a real correctness bug hide as a green status row.
- **Drain-all in `apply_restart_results`, drain-one in `apply_stop_results` / `apply_deletion_results`** (`src/tui/home/mod.rs:2506, 2538, 2869`). The new code drains all pending per tick (correct); the siblings drain one. Either align both, or document the asymmetry.
- **`RestartPoller` worker thread is nameless** (`src/tui/restart_poller.rs:27`). `thread::spawn` without `Builder::name`. Mirrors `StopPoller`'s same miss but the wake worker right next door (`src/session/restart.rs:62-63`) does name itself.
- **Tracing target inconsistency for restart-related logs** (`src/session/restart.rs:90, 94` vs `src/tui/home/mod.rs:2881, 2901`). Wake-up failures log without a `target:`. Cascade-failure log uses `target: "tui.home"`. Poller send-failure log uses `target: "tui.restart_poller"`. Three different filters for one feature.

### Hardening highlights (selected WARN, not blocking)

The full WARN set is ~14 items; the highest-impact subset:

- **`wait_with_timeout` swallows partial stderr on timeout** (`src/containers/runtime_base.rs:51-54`). For the actual freeze case this PR fixes (registry stall), the partial stderr is the only signal you'd have; today it's discarded.
- **`try_wait()?` early-return leaks child + reader threads** (`src/containers/runtime_base.rs:42`). On `?` propagation, `Child::Drop` does NOT kill the child on Unix or Windows; reader threads remain alive holding the original pipes.
- **No cumulative-skip cap on `copy_dir_recursive`**. A directory with 1,000 unreadable files logs 1,000 warns and reports success.
- **No recursion-depth guard for non-symlinked deep trees**. Cycle detection catches symlink loops; a real 65k-deep tree blows the stack.
- **Symlink targets outside the source root are followed and copied**. A symlink under `~/.claude/plugins/` pointing to `/etc` exfiltrates host content into the bind-mounted sandbox. Pre-existing behaviour, but the PR re-entrenches it without a defence-in-depth check.
- **Test gaps**: no test covers worker panic / channel disconnect, `restart_in_flight` insertion + clearing, `pollable_instances` filtering, double-press skip, ENOSPC scenario. The new state machine has zero unit-test coverage for the most critical paths.

### Plan to unblock merge (smallest correct change per cluster)

1. **State machine guards** (~30 LOC):
   - Early-out in `restart_selected_session`: `if self.restart_in_flight.contains(&id) { return Ok(()); }` (covers double-press BLOCKING).
   - Same guard at the top of `delete_selected` and `delete_group_with_sessions` (covers delete-vs-restart BLOCKING).
   - Convert `try_recv_result()` to surface `Disconnected` (matching `apply_recovery_updates` shape); have `apply_restart_results` clear `restart_in_flight` and the cached row state on disconnect (covers panic-leak BLOCKING).
   - Re-raise `InfoDialog::new("Restart Failed", ...)` from `apply_restart_results` on the error path (covers user-visible regression BLOCKING).
2. **Worker scope** (~5 LOC): wrap `perform_restart` body in a `HookTimeoutScope` so the cascade's `before_start` hooks honor the same timeout as the recovery worker.
3. **Filesystem fail-soft scope** (~15 LOC):
   - Inspect `e.raw_os_error()` on `fs::copy` failure: propagate `ENOSPC`, `EDQUOT`, `EROFS` as `Err`; warn-and-skip the rest.
   - Warn-and-skip on `canonicalize()` failure (and add the entry to `visited` only on `Ok`).
   - Warn-and-skip on `read_dir` / `entry?` failures (matching the recursive-call shape from `b8f9691f`).
4. **AGENTS.md compliance** (~10 LOC): replace each `let _ =` in `wait_with_timeout` with `if let Err(e) = ... { tracing::warn!(...) }`.
5. **Tests** (~120 LOC): add `restart_selected_session_skips_when_already_in_flight`, `apply_restart_results_clears_in_flight_on_disconnect`, `delete_selected_refused_during_restart`, `copy_dir_recursive_propagates_enospc`, plus a panic-path test for `perform_restart`.

### Method note

Audit performed with three parallel adversarial agents on a fresh worktree at HEAD `f31c5389`. Axes adapted to the PR's focused scope: (1) concurrency + process spawn + timeout robustness, (2) filesystem robustness in `container_config` copy, (3) TUI state machine + tests + Rust idioms + AGENTS.md compliance. Findings then cross-validated by a second pass of three parallel agents that re-read both the worktree and the prior findings; commit `b8f9691f` (which the author pushed during the audit) resolved one BLOCKING (unreadable-subdir abort) and forced a re-anchor of two inline comments to in-diff lines on the new HEAD `b8f9691f`. Raw agent reports retained locally; can be shared on request. AI use is declared per the project AI-disclosure rule.

jerome-benoit:
Method note: this review is the output of a three-axis adversarial audit run against HEAD `f31c5389` in a clean worktree, then cross-validated by a second three-axis pass. Axes: concurrency + process spawn + timeout robustness, filesystem robustness in `container_config` copy, TUI state machine + tests + Rust idioms + AGENTS.md compliance. Commit `b8f9691f` (which you pushed while the audit was in progress) resolved one BLOCKING (unreadable-subdir abort) and forced a re-anchor of two inline comments to in-diff lines on the new HEAD; the body summary reflects the post-fixup state. Inline citations point to the smallest-correct-change site for each finding. AI use is declared per the project AI-disclosure rule.

njbrake:
_Note: drafted by Claude via back-and-forth with @njbrake. The reasoning is mine; the prose is Claude's._

Thanks, great audit. You're right that off-threading the cascade removed an implicit lock, which is the root of most of these. All 6 BLOCKING items are fixed in `53c24432`:

- **Double-press**: `restart_selected_session` drops the press when already in flight (the 1.5s cooldown only debounced key-repeat).
- **Delete vs restart**: `delete_selected` / `delete_group_with_sessions` refuse with a dialog while in flight.
- **Disconnected leak**: `try_recv_result` surfaces `Disconnected`; `apply_restart_results` clears `restart_in_flight`, matching `apply_recovery_updates`.
- **Lost "Restart Failed" dialog**: now raised from `apply_restart_results` on the async failure.
- **`before_start` hook hang**: `perform_restart` installs a `HookTimeoutScope(recovery_hook_timeout())`.
- **ENOSPC/EROFS/EDQUOT**: `copy_dir_recursive` now discriminates on `raw_os_error`; storage-exhaustion aborts (and fails the restart visibly), everything else warn-skips. The fail-soft path is errno-aware across `create_dir_all`, `read_dir`, the entry iterator, and `copy`, which also covers your "success on partial copy" point.

Plus: named the poller thread, consistent `session.restart` log target, and tests for skip-when-in-flight, delete-refused, and the errno split.

**One pushback**: the `let _ =` / AGENTS.md item. I don't see that rule in AGENTS.md, and `let _ =` is used widely in production (25 in `src/session/instance.rs` alone), so I left `wait_with_timeout` as-is. Point me at the line if I'm on a stale copy.

**Deferred** (shout if any should land here): symlink-escapes-source-root (pre-existing; risks breaking legit symlinked configs), recursion-depth/skip caps (not reachable for `~/.claude/plugins`), partial-stderr-on-timeout (minor). I skipped the worker-panic unit test since I couldn't force a dead poller cleanly; the disconnect logic mirrors the covered recovery path.

src/tui/home/mod.rs
```diff
@@ -2786,56 +2799,139 @@ impl HomeView {
...
+                    break;
+                }
             }
+        }
 
-            self.update_selected();
+        if touched {
+            self.refresh_rows_preserving_selection();
+            if let Err(e) = self.save() {
+                tracing::error!(target: "tui.home", "Failed to save after restart: {}", e);
```
jerome-benoit, line 2933: The v3 commit unified restart logs on `target: "session.restart"` (lines 2886, 2919; `restart.rs:101, 105`) but this single line kept `target: "tui.home"`, so save-failure after a restart splits across two filters.

Fix: `target: "session.restart"`.

src/session/restart.rs
```diff
@@ -0,0 +1,137 @@
...
+    // Honor the same on_launch / before_start hook timeout the startup-recovery
+    // worker installs (`run_recovery_for_instance`). Without it, a hanging
+    // before_start hook (e.g. a `mint` script waiting on the network) runs with
+    // no kill timer and wedges this serial worker thread forever, taking every
+    // future restart down with it.
+    let outcome = {
+        let _scope = crate::session::recovery::HookTimeoutScope::new(
+            crate::session::recovery::recovery_hook_timeout(),
+        );
+        instance.restart_with_size(size).map_err(|e| e.to_string())
```
jerome-benoit, line 52: `map_err(|e| e.to_string())` keeps only the top frame. `instance.restart_with_size` returns `anyhow::Result<StartOutcome>`; the error chain (`with_context` layers, root cause) is dropped before the new `InfoDialog::new("Restart Failed", ...)` at `mod.rs:2897` renders it. The dialog ends up showing `"creating sandbox dir"` rather than `"ENOSPC: ... caused by: ... caused by: creating sandbox dir"`.

Fix: `format!("{e:#}")` rather than `e.to_string()`. Preserves the chain across the thread boundary; `RestartResult.outcome` stays `Result<_, String>`, no `Box<dyn Error + Send + Sync>` gymnastics.

src/session/container_config.rs
```diff
@@ -448,16 +454,124 @@ fn rewrite_plugin_value_paths(
...
+/// Whether an I/O error during the copy means the destination filesystem can no
+/// longer accept writes, in which case continuing would silently produce a
+/// partial copy (a sandbox missing arbitrary files, reported as success). Those
+/// must abort the whole copy. Everything else (a single unreadable or dangling
+/// source entry) is skipped best-effort.
+fn is_fatal_copy_error(e: &std::io::Error) -> bool {
+    // ENOSPC (no space), EROFS (read-only fs), EDQUOT (quota). EDQUOT differs by
+    // platform (122 on Linux, 69 on macOS). raw_os_error covers all three more
+    // portably than the (partly unstable) ErrorKind variants.
+    matches!(e.raw_os_error(), Some(28) | Some(30) | Some(69) | Some(122))
```
jerome-benoit, line 470: Two issues on this predicate:

1. `Some(69) | Some(122)` matches macOS EDQUOT (69) AND Linux EDQUOT (122) on every host. On Linux x86_64/aarch64, errno 69 is EBADRQC; on macOS, errno 122 is ENOTRECOVERABLE. Plus Linux MIPS uses 1133 and Linux Sparc uses 69, both missed by a literal `122`. The doc-comment at lines 467-469 claims `raw_os_error` is more portable than `ErrorKind`, which is true only with platform gating.
2. EFBIG (errno 27) is missing. On a destination filesystem with a max file size cap (FAT32 >2GB, ext4 with 1KB blocks >16TB), `std::fs::copy` returns EFBIG. The current branch falls through to the warn-skip path at `container_config.rs:570-577`, producing the silent partial copy the docstring at line 461 says the predicate exists to prevent.

Cleanest: use `libc` constants (libc is transitive at 0.2.186; needs `libc = "0.2"` in `Cargo.toml`):

```rust
fn is_fatal_copy_error(e: &std::io::Error) -> bool {
    // ENOSPC, EROFS, EDQUOT, EFBIG: destination filesystem refuses or
    // cannot fit further writes; continuing risks a silent partial copy.
    matches!(
        e.raw_os_error(),
        Some(libc::ENOSPC) | Some(libc::EROFS) | Some(libc::EDQUOT) | Some(libc::EFBIG)
    )
}
```

Add positive errno tests for `EFBIG` and `libc::EDQUOT` at `container_config.rs:2679-2696`.

src/tui/restart_poller.rs
```diff
@@ -0,0 +1,122 @@
...
+    ) {
+        while let Ok(request) = request_rx.recv() {
+            let result = perform_restart(request);
+            if result_tx.send(result).is_err() {
+                break;
+            }
+        }
+    }
+
+    pub fn request_restart(&self, request: RestartRequest) {
```
jerome-benoit, line 53: After the worker thread panics in `perform_restart`, `request_restart` logs `"restart request dropped; worker thread unavailable"` on `request_tx.send` failure but does not respawn the worker.

Sequence after a panic:
1. `apply_restart_results` hits `Err(TryRecvError::Disconnected)`, clears `restart_in_flight` (correct).
2. User presses `e` to restart row B.
3. `restart_selected_session` passes the in-flight guard at `operations.rs:321` (set is empty).
4. Mutates row B to `Status::Starting` at `operations.rs:454-459` with fresh `last_start_time`.
5. `request_tx.send` returns `Err`, only a warn-log fires.
6. Row B sits at `Status::Starting` until `StatusPoller` eventually marks it `Error` after a grace period.

The user gets no UI feedback that the press was dropped. Probability is low (`perform_restart` is small and well-covered) but the failure mode is invisible.

Two ways to close: auto-respawn the worker on `Disconnected` detection inside `apply_restart_results`, or expose `RestartPoller::is_alive()` and have `request_restart` return `Result<(), std::sync::mpsc::SendError<RestartRequest>>` so `restart_selected_session` can raise an `InfoDialog::new("Restart unavailable", "Restart worker has stopped; restart aoe to recover")` rather than flipping to Starting and forgetting.

jerome-benoit (approved):
# Follow-up review on commit 53c24432

The 6 BLOCKING items from my prior review are addressed by `53c24432`. APPROVE.

## Withdrawn

My prior inline at `src/containers/runtime_base.rs:49` cited *"AGENTS.md: 'No `let _ =` outside tests.'"*; that rule does not exist in AGENTS.md. The six `wait_with_timeout` sites are benign (background-drainer feeds, `mpsc::send` to a still-live receiver, child cleanup on an abandoned process). Withdrawing.

## Advisory follow-ups (not blocking)

Four small items inline below.

PR #2203, opened 2026-06-16
feat(hooks): add HookFormat enum, Codex writes to hooks.json (closes #2187)

### LINKED ISSUE #2187: hooks: replace string-based per-agent dispatch with format enums on AgentHookConfig and SidecarHooks
## Summary

`src/hooks/mod.rs:304-308` (`iter_hook_targets_in`) and `src/hooks/mod.rs:386-399` (`has_aoe_marker` sidecar branch) dispatch per-agent format via string comparison and file-extension matching:

```rust
// agent_settings_path_in vs codex_config_path_in selection:
let kind = if agent.name == "codex" {
    HookTargetKind::CodexToml
} else {
    HookTargetKind::JsonSettings
};

// SidecarHooks marker walker selection:
HookTargetKind::Sidecar(sidecar) => {
    let sub = sidecar.host_config_subpath;
    if sub.ends_with(".toml") { settl_config_has_aoe_marker(...) }
    else if sub.ends_with(".yaml") || sub.ends_with(".yml") { hermes_config_has_aoe_marker(...) }
    else { kiro_config_has_aoe_marker(...) }
}
```

A future TOML-format agent that isn't `codex` would be silently mis-classified as `JsonSettings`. A future sidecar with `.json5`/`.jsonc`/non-`.toml` would silently route to the Kiro JSON walker and fail-close. Both are correct for every agent that exists today; both are landmines for the next contributor adding an agent.

## Context

Filed as a follow-up to PR #2168 (issue #1845) round-4 audit findings F8 and F9. Bundling because both are "make per-agent dispatch type-driven instead of name-or-extension-driven".

## Suggested approach

1. Add a `HookFormat` enum to `crate::agents` with variants `JsonSettings` and `CodexToml`. Add a `pub format: HookFormat` field to `AgentHookConfig` (`src/agents.rs:62-74`). Populate every existing `AgentHookConfig` declaration; replace `if agent.name == "codex"` at `hooks/mod.rs:304-308` with `match hook_cfg.format`.

2. Add a `SidecarFormat` enum (`SettlToml`, `HermesYaml`, `KiroJson`). Add `pub format: SidecarFormat` to `SidecarHooks` (`src/agents.rs:83-99`). Populate every sidecar declaration; replace the file-extension dispatch at `hooks/mod.rs:386-399` with `match sidecar.format`.

3. Update `docs/development/adding-agents.md` to point new agents at the format enums.

Estimated size: ~30 LOC + ~20 LOC = ~50 LOC across `agents.rs`, `hooks/mod.rs`, and the per-agent declarations.

## Out of scope

- Changing the actual install / uninstall / marker-walker behaviour. This issue is a type-system tightening, not a behaviour change.
- The migration v015 (PR #2168) does not need to be re-run after this lands.

## Acceptance criteria

- New agent or sidecar added without touching `iter_hook_targets_in` or `has_aoe_marker`'s match arms compiles only after picking a format variant.
- All 162 existing `hooks::tests` and 17 `migrations::v015` tests still pass.

PennixRv:
## Description

Implements `HookFormat` enum on `AgentHookConfig` per #2187, and extends it with a `CodexJson` variant so Codex status hooks land in `~/.codex/hooks.json` instead of inline `[[hooks]]` tables in `config.toml`.

**Why**: Codex emits a startup warning when hooks are in both `hooks.json` and `config.toml`. Since Codex recommends `hooks.json` as the forward-looking format, and tools like oh-my-codex write there, AOE should adopt it. The string-based dispatch (`agent.name == "codex"`) is fragile — #2187 proposed replacing it with an enum.

**Changes**:
- `src/agents.rs`: Add `HookFormat` enum (`JsonSettings`, `CodexToml`, `CodexJson`), add `format` field to `AgentHookConfig`, update all 5 agent definitions
- `src/hooks/mod.rs`: Add `CodexJson` to `HookTargetKind`, replace string dispatch with `match hook_cfg.format`, add `codex_config_file_in()`, wire through all match arms
- `src/session/instance.rs`: Dispatch by `HookFormat`, write JSON to `hooks.json` for Codex
- `src/migrations/v015_rewrite_hook_strings.rs`: CodexJson through generic JSON rewriter

## PR Type

- [x] New Feature
- [x] Refactor

## Checklist

- [x] For UI changes: N/A

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: this PR does not change TUI rendering, a CLI subcommand, or session lifecycle

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Code (Anthropic)
**Details:** AI assisted with implementation design and test adaptation under human review.


njbrake:
> _Note: this review was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks for this. The `HookFormat` enum is a clean piece of work and the dispatch refactor is the right shape for #2187. I have some concerns about the bundled behavior change (moving Codex to `hooks.json`), mostly around it not fully delivering the outcome it's aiming for.

**1. The dual-source warning is not actually fixed for existing users.** The goal is to stop Codex warning when hooks live in both `hooks.json` and `config.toml`. But a user who already has AoE hooks in `config.toml` from a prior version keeps them: now that Codex is classified `CodexJson`, `iter_hook_targets` no longer scans `config.toml`, and nothing uninstalls the stale AoE block there. New hooks land in `hooks.json`, so AoE hooks end up in both files, which is the warning we set out to remove. There's no migration to strip the old `config.toml` AoE hooks. The `CodexJson` install path probably needs to also call `uninstall_codex_hooks` on the `config.toml` path, or a follow-up migration needs to clean it up.

**2. `features.hooks = false` is no longer respected.** The old TOML installer short-circuited when a user set Codex's `features.hooks` / `codex_hooks = false`. The generic `install_hooks` used by the `hooks.json` path has no such gate, so AoE now writes hooks regardless of that setting. The deleted `codex_features_hooks_false_skipped` test covered this, and the surviving `codex_features_codex_hooks_alias_false_skipped` now passes vacuously: it asserts `config.toml` is untouched, which is trivially true once `config.toml` is no longer scanned, so the gate is never exercised. This reads as an untested behavior regression.

**3. Trust-state preservation.** The TOML path preserved `[hooks.state]` trusted_hash with file-locked atomic writes. The generic `install_hooks` does a plain read-modify-write with no Codex trust handling. If Codex's `hooks.json` carries any per-hook trust or enabled state, the generic installer could clobber it. Can you confirm the `hooks.json` schema matches the `{"hooks": {...}}` shape this writes, and that there's no trust state to preserve?

**4. Host vs sandbox inconsistency.** `container_config.rs` still installs Codex hooks into `config.toml`, so host sessions use `hooks.json` while sandboxed sessions use `config.toml`. You called this out in the commit message; flagging that it means the dual-source fix only applies on the host.

**Smaller items:**
- The `HookFormat::CodexToml` branch in `install_agent_status_hooks` is now unreachable, since no agent declares `CodexToml`. Switching that if/else chain to `match hook_cfg.format` would let the compiler flag dead arms and any future variant (CodeRabbit raised this too).
- `codex_features_hooks_false_skipped` is now an assertion-free `#[test]`; worth removing or folding its note into the alias test.
- The `.omx/` `.gitignore` entry looks unrelated to this change.

**On scope:** #2187 was framed as a type-tightening with no behavior change, and it also asked for a `SidecarFormat` enum and a `docs/development/adding-agents.md` update. This PR delivers the `HookFormat` half well but adds a separate, riskier behavior change and leaves the sidecar and docs parts out. My suggestion would be to land the pure enum refactor as "closes #2187" (ideally with the sidecar enum and docs), and split the `hooks.json` migration into its own PR that handles the `config.toml` cleanup, the feature gate, and the sandbox path together. Happy to discuss if you'd rather keep them combined.

njbrake (changes requested):
See the previous comment, thank you for the contribution, just a few tweaks and then it will be good to go!

jerome-benoit (changes requested):
The PR doc-comment at `src/agents.rs:64` states: "Replaces the ad-hoc string-based dispatch (`agent.name == "codex"`) with an explicit enum so each agent declares how its hooks are serialized." Three production sites still match on `agent.name == "codex"`:

- [`src/session/container_config.rs:1449`](src/session/container_config.rs#L1449) routes the sandbox install.
- [`src/session/container_config.rs:980`](src/session/container_config.rs#L980) gates `should_refresh_codex_hooks`.
- [`src/tui/dialogs/hooks_install.rs:40`](src/tui/dialogs/hooks_install.rs#L40) routes the consent-dialog display.

Each one produces a concrete defect below. Plus a v014→v015 migration regression and an atomic-write regression on the host install path.

**Sandbox install writes TOML bytes into `hooks.json`.** [`src/agents.rs:413`](src/agents.rs#L413) declares `settings_rel_path: ".codex/hooks.json"`. [`src/session/container_config.rs:1448-1450`](src/session/container_config.rs#L1448-L1450) routes the codex branch into `crate::hooks::install_codex_hooks(&settings_file, ...)`, which writes TOML through `write_codex_config` ([`src/hooks/mod.rs:733-735, 752-777, 809-812`](src/hooks/mod.rs#L733-L812)). The container reads that file as JSON and fails to parse it. The pre-existing test `test_build_container_config_installs_codex_hooks_files` ([`src/session/container_config.rs:3151-3210`](src/session/container_config.rs#L3151-L3210)) plants nothing, asserts `config.toml.exists()` ([`:3184`](src/session/container_config.rs#L3184)), asserts `!hooks.json.exists()` ([`:3185`](src/session/container_config.rs#L3185)), then `fs::read_to_string(...config.toml).unwrap()` at [`:3187`](src/session/container_config.rs#L3187). After the path flip the file is absent, so the test panics on `Err(NotFound)` before the substring assertion runs. Consistent with the PR's blocked merge state. [`refresh_codex_sandbox_hooks`](src/session/container_config.rs#L992-L1010) hardcodes `config.toml`; when that path runs the sandbox ends up with TOML in two files.

**v014 → v015 upgrade is a silent no-op for users who already have hooks; double-firing is reintroduced.** [`src/agents.rs:418`](src/agents.rs#L418) declares Codex as `HookFormat::CodexJson`. [`iter_hook_targets_in`](src/hooks/mod.rs#L311-L339) emits one target per declared format, so Codex enumerates only `~/.codex/hooks.json`. No agent in `AGENTS` declares `CodexToml` (verified across [`src/agents.rs:352, 418, 469, 492, 540, 593, 661`](src/agents.rs)). v015 walks that target set ([`src/migrations/v015_rewrite_hook_strings.rs:116-119`](src/migrations/v015_rewrite_hook_strings.rs#L116-L119)) and does not visit `~/.codex/config.toml`. Existing AoE-marked `[[hooks.SessionStart]]` blocks (planted by pre-#1803 installs, including any unhardened command strings) survive untouched. The next `aoe add --cmd codex` writes `hooks.json` alongside them. Codex reads both files and fires every hook twice. [`uninstall_all_hooks`](src/hooks/mod.rs#L1670-L1697) walks the same target set, so `aoe uninstall` cannot reach the orphans. v015 needs an explicit one-shot legacy sweep that calls `uninstall_codex_hooks` against `~/.codex/config.toml` for each resolvable `CODEX_HOME` before bumping the schema, or a v016 migration scoped to that cleanup.

**Sandbox refresh stops firing for fresh codex installs.** [`should_refresh_codex_hooks`](src/session/container_config.rs#L979-L990) gates on `host_config.exists() || sandbox_config.exists()`, where both names are `config.toml`. Fresh installs run `aoe add --cmd codex`, which produces only `~/.codex/hooks.json`, so the gate returns `false` and the refresh never runs. For users with a leftover `config.toml` from a prior install, the gate fires and writes TOML hooks into the sandbox via `refresh_codex_sandbox_hooks` while the host install loop writes JSON-bytes-as-TOML alongside it (see the first blocker). Sandbox state depends on whether the user ever ran a previous build. The fix that closes both this and the first blocker is to migrate `container_config.rs` and `should_refresh_codex_hooks` to the new `HookFormat::CodexJson` dispatch and drop the `agent.name == "codex"` matches.

**Atomic-write regression on Codex.** Codex installs route through `install_hooks` ([`src/migrations/v015_rewrite_hook_strings.rs:155-157`](src/migrations/v015_rewrite_hook_strings.rs#L155-L157), [`src/session/instance.rs:1981-2003`](src/session/instance.rs#L1981-L2003)), which calls `std::fs::write` directly ([`src/hooks/mod.rs:711`](src/hooks/mod.rs#L711)). The temp + rename + fsync sequence in `crate::session::atomic_write` reachable through `with_codex_config_lock` is unused on this path. Power-loss mid-write leaves a 0-byte `hooks.json`; the gate at `has_aoe_marker` then fails closed and v015 will not re-install (it only rewrites existing markers). The module doc at [`v015:69-77`](src/migrations/v015_rewrite_hook_strings.rs#L69-L77) documents atomic-write semantics that `install_hooks` does not provide. Route `install_hooks` through `atomic_write`. The module doc is wrong against the live path until that lands.

**Two vacuous tests in v015.** `codex_features_hooks_false_skipped` ([`src/migrations/v015_rewrite_hook_strings.rs:531-540`](src/migrations/v015_rewrite_hook_strings.rs#L531-L540)) is `#[test]` on a comment-only body. Zero setup, zero assertions. Passes unconditionally. `codex_features_codex_hooks_alias_false_skipped` ([`:542-580`](src/migrations/v015_rewrite_hook_strings.rs#L542-L580)) plants `~/.codex/config.toml` with `features.codex_hooks = false` and a legacy `[[hooks.SessionStart]]` block, runs `run_in`, asserts byte-equality. Combined with the `CodexToml` non-enumeration above, the migration does not visit `config.toml`, so the equality holds because the file is never opened, regardless of the feature flag. The deprecated-alias short-circuit invariant the docstring claims is not locked. Delete them, or re-target against `codex_hooks_feature_is_disabled` directly so the gate logic is exercised.

**`HookFormat::CodexToml` is dead.** Zero producers in `AGENTS`. Match arms at [`src/hooks/mod.rs:317`](src/hooks/mod.rs#L317), [`:323`](src/hooks/mod.rs#L323), [`:401`](src/hooks/mod.rs#L401), [`:1676`](src/hooks/mod.rs#L1676), [`v015:158-161`](src/migrations/v015_rewrite_hook_strings.rs#L158-L161), [`instance.rs:1970-1980`](src/session/instance.rs#L1970-L1980) are unreachable. `with_codex_config_lock`, `snapshot_codex_hooks_state`, `install_codex_hooks_with_preserved_state`, `read_codex_config`, `write_codex_config`, and `codex_config_has_aoe_marker` survive only as dependencies of `refresh_codex_sandbox_hooks`, which itself participates in the first two blockers. AGENTS.md disallows dead code. Wire the variant to the legacy-cleanup path proposed above; if that fix lives in v016 instead, remove the variant from this PR together with the helper fan-out.

**Dispatch by `==` defeats the point of the enum.** [`src/session/instance.rs:1967-2028`](src/session/instance.rs#L1967-L2028) uses `else if hook_cfg.format == HookFormat::CodexToml { ... } else if hook_cfg.format == HookFormat::CodexJson { ... } else { ... }`. The trailing `else` swallows any future variant; adding `HookFormat::AiderConfig` will silently route through the generic JSON branch, recreating the exact fragility this PR was framed as fixing (per [`src/agents.rs:62-66`](src/agents.rs#L62-L66)). Same shape in [`iter_hook_targets_in`](src/hooks/mod.rs#L321-L326), which lists two of three variants and uses `_ => agent_settings_path_in(...)` for the third. Replace both with `match hook_cfg.format` so a new variant produces a compile error.

**Consent dialog promises a path the install does not write.** [`src/tui/dialogs/hooks_install.rs:40-44`](src/tui/dialogs/hooks_install.rs#L40-L44) routes the codex display through `codex_config_path_display_for_host_environment` ([`src/hooks/mod.rs:95-105`](src/hooks/mod.rs#L95-L105)), which returns `<CODEX_HOME or ~/.codex>/config.toml`. The install at [`src/session/instance.rs:1981-2003`](src/session/instance.rs#L1981-L2003) writes `<CODEX_HOME or ~/.codex>/hooks.json`. Users who accept the dialog get a different file than the dialog promised. The accompanying `needs_codex_trust_note = true` ([`hooks_install.rs:41`](src/tui/dialogs/hooks_install.rs#L41)) describes the `[hooks.state]` trust mechanism, which exists only in TOML; the install does not touch that file.

### Notes

- [`src/session/instance.rs:1984`](src/session/instance.rs#L1984) builds the JSON path by `codex_config_path_for_launch_env().map(|p| p.with_file_name("hooks.json"))`. The swap relies on the helper returning a path whose final component is `config.toml`. Expose `codex_config_file_in` as `pub(crate)` and call it directly, or introduce a `CodexHooksJsonPath` newtype that centralises `CODEX_HOME` resolution for the JSON sidecar (this also removes the `.or_else` chain at [`:1985-1989`](src/session/instance.rs#L1985-L1989), which currently discards the primary error via `|_|`).
- [`src/session/instance.rs:1967-2028`](src/session/instance.rs#L1967-L2028) opens with `if self.is_sandboxed() { /* comment-only */ } else if ...`. Empty-body `if` plus comment is a smell; invert to early-return on `is_sandboxed()` so the host paths sit at one indentation level.
- `tracing::warn!` macros at [`instance.rs:1976`](src/session/instance.rs#L1976) and [`:1979`](src/session/instance.rs#L1979) omit `target: "session.store"` while neighbours at [`:1998`](src/session/instance.rs#L1998), [`:2003`](src/session/instance.rs#L2003), [`:2021`](src/session/instance.rs#L2021), [`:2025`](src/session/instance.rs#L2025) include it. `RUST_LOG=session.store=warn` filtering misses the codex-toml branch.
- The four codex tests at [`instance.rs:3877`](src/session/instance.rs#L3877), [`:3898`](src/session/instance.rs#L3898), [`:3928`](src/session/instance.rs#L3928), [`:3953`](src/session/instance.rs#L3953) decorate with `#[serial_test::serial]` (unnamed default group) while v015's codex tests use `#[serial_test::serial(shell_env)]`. Both groups mutate `HOME`, `XDG_CONFIG_HOME`, and `CODEX_HOME` and can race. Move the four tests to `serial(shell_env)`.
- `CodexHomeGuard` at [`instance.rs:3840-3848`](src/session/instance.rs#L3840-L3848) restores only `CODEX_HOME`. The new tests at [`:3880-3882`](src/session/instance.rs#L3880-L3882), [`:3901-3903`](src/session/instance.rs#L3901-L3903), [`:3931-3933`](src/session/instance.rs#L3931-L3933), [`:3956-3958`](src/session/instance.rs#L3956-L3958) also set `HOME` and `XDG_CONFIG_HOME` without saving them. Process env leaks across tests in the same `cargo test` invocation. Use the `EnvGuard` pattern from v015's test module.
- `iter_hook_targets_in` deduplicates by `path` only ([`hooks/mod.rs:328-331, 353-357`](src/hooks/mod.rs#L328-L357)); the doc at [`:294`](src/hooks/mod.rs#L294) promises dedup by `(kind discriminant, path)`. Latent today, becomes a correctness gap when a fourth variant aliases on path.
- `codex_state_preserved_across_rewrite` ([`v015:504-529`](src/migrations/v015_rewrite_hook_strings.rs#L504-L529)) plants `hooks.json` (no state block) and asserts only that the command was rewritten. State preservation is not exercised. The state lives in `config.toml`, which the test does not open. Rename the test to match what it asserts.
- The `CodexJson` doc at [`agents.rs:75-77`](src/agents.rs#L75-L77) and [`hooks/mod.rs:268-271`](src/hooks/mod.rs#L268-L271) claims symlink resolution; `codex_config_file_in` joins paths without resolving symlinks, and `install_hooks` calls `fs::write` rather than `codex_config_write_path`. Tighten the doc to match the live path. The `HookTargetKind::CodexToml` doc at [`hooks/mod.rs:265-267`](src/hooks/mod.rs#L265-L267) describes machinery the install path does not use.
- [`agents.rs:295-296`](src/agents.rs#L295-L296) documents loading from the `[hooks]` table in `config.toml`. Update to reference `hooks.json`.
- Three emdashes: [`agents.rs:71`](src/agents.rs#L71), [`agents.rs:74`](src/agents.rs#L74), [`v015:537`](src/migrations/v015_rewrite_hook_strings.rs#L537). AGENTS.md disallows.
- If `HookFormat` stays `pub`, add `#[non_exhaustive]` so external matchers do not break on a fourth variant. Every consumer is in-crate, so `pub(crate)` would also work.
- `json_settings_has_aoe_marker` ([`hooks/mod.rs:415`](src/hooks/mod.rs#L415)) is now used for `JsonSettings` and `CodexJson`. Rename to `json_hooks_has_aoe_marker` for clarity.
- `.gitignore:86` (`.omx/`) is unrelated to the PR scope.
- Re #2187: the `SidecarFormat` enum and the `docs/development/adding-agents.md` update are absent from this branch. [`hooks/mod.rs:402-411`](src/hooks/mod.rs#L402-L411) still dispatches by `.ends_with(".toml" / ".yaml" / ".yml")`. The issue's out-of-scope explicitly carved out behaviour changes; the Codex storage-format pivot is one. Split the format-storage flip into its own PR, or amend #2187 to cover it.

jerome-benoit:
Following up. My 2026-06-17 review and @njbrake's review are still open and the
five concrete blockers in the dual-source warning, the v014 to v015 silent
no-op for existing users, the sandbox refresh gate, the atomic-write
regression on Codex, and the dead `HookFormat::CodexToml` variant remain
unaddressed. CI is BLOCKED, and the test `test_build_container_config_installs_codex_hooks_files`
panics on the path flip.

Proposing a split as next step:

- The `HookFormat` enum and the dispatch refactor are the core of #2187 and a
  good piece of work. They are landable as a pure type-tightening with zero
  behavior change.
- The Codex `config.toml` to `hooks.json` flip is a separate behavior change
  that needs its own PR with the `config.toml` cleanup migration, the
  `should_refresh_codex_hooks` rework, the atomic-write fix, and the sandbox
  parity. The current bundling is what is keeping #2187 blocked.

Two paths from here:

1. Rebase this PR onto the pure type-tightening subset (drop `CodexJson`,
   keep Codex on `CodexToml`, no path changes). I happy-merge as soon as
   the test panic clears.
2. If no response within 48 hours, I take over #2187 with a clean PR that
   delivers only the enum refactor scope. The behavior-change subset stays
   available for a follow-up PR by anyone (you or otherwise).

Either path works. Let me know if you have a preference.

jerome-benoit:
Twenty-four hours elapsed on the split proposal. I have moved #2333 from draft to ready. It covers the type-tightening subset of #2187, with the open blockers on this PR addressed where in scope and called out as out-of-scope where they relate to the bundled storage-format change.

This PR remains open in case you want to pivot it onto the storage-format follow-up (the `config.toml` cleanup migration, the `should_refresh_codex_hooks` rework on fresh installs, the atomic-write fix, and the sandbox parity). Otherwise it is superseded by #2333 for closing #2187.

njbrake:

Thanks for this, @PennixRv. We've split your change into two reviewable pieces that have now landed (or are landing): #2333 takes the `HookFormat` / `SidecarFormat` enum work, and #2353 takes the Codex `config.toml` to `hooks.json` storage pivot plus a `v018` cleanup migration. Since both together cover everything here, I'm closing this in favor of that stack. Credit for the original direction is yours; really appreciate the contribution.

PR #2204, opened 2026-06-16
fix(tui): clear screen without a cursor-position query

peteski22:
## Description

Starting `aoe` could abort immediately with:

    Error: The cursor position could not be read within a normal duration

ratatui 0.30's `Terminal::clear()` snapshots the cursor via `get_cursor_position()` (an `ESC[6n` round-trip) before clearing. That read goes through crossterm's shared internal event reader and is unreliable in a TUI that also runs a crossterm `EventStream`: it can fail with this error immediately (a 0 ms failure, not the 2 s timeout the message implies), aborting startup. ratatui 0.29's fullscreen clear did no such round-trip, so this surfaced with the 0.30 upgrade.

Fix: clear without the cursor snapshot. `clear_terminal` now issues `Clear(ClearType::All)` on the backend (the same operation ratatui's fullscreen `clear_viewport` runs) and resets the diff baseline with `swap_buffers()` so the next draw repaints the whole viewport. The clear path no longer calls `get_cursor_position`, so the error is impossible by construction. All four clear sites (startup, loop redraw, raw-mode restore, structured-view return) route through the helper.

Verified: `cargo build`, `cargo fmt --check`, `cargo clippy -D warnings`, and `cargo test` pass; `aoe` launches to the home screen on a terminal that previously crashed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (none needed)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the failure is a terminal-specific cursor-read timeout that needs a real tty plus the crossterm `EventStream` lifecycle; it does not reproduce under the tmux-based e2e harness (tmux answers `ESC[6n`), so a test there would be a false-green. The fix is structural: the clear path no longer calls `get_cursor_position`, so the error cannot occur.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Root-cause investigation and the fix were done with AI assistance under human direction and verification.

## Summary of changes

This PR fixes a terminal startup crash in the `aoe` TUI that occurs when the terminal can't reliably report cursor position during screen clearing—a problem that emerged after upgrading to ratatui 0.30.

### What was changed
- Replaced `terminal.clear()` with a shared `clear_terminal()` helper function at all screen-clearing points in the TUI application.
- The transition happens at startup initialization, main loop redraws, event-stream recreation, and when returning from the remote structured view.
- Updated comments throughout the codebase to reflect the new clearing and redraw behavior.

### What was added
- A new crate-local `pub(crate) clear_terminal()` helper function in `src/tui/mod.rs` that:
  - Clears the entire viewport directly at the backend level (avoiding ratatui's higher-level cursor-position snapshot),
  - Resets ratatui's internal buffer baseline so the next frame repaints the full screen instead of skipping unchanged regions.
- Unit tests using `TestBackend` to verify the clear-and-redraw invariant works correctly.

### What was removed
- All direct calls to `terminal.clear()` in the affected TUI flow, which internally triggered ratatui's cursor-position query (`ESC[6n` round-trip).

### Root cause & benefits
- **The problem:** Ratatui 0.30 changed `Terminal::clear()` to perform a cursor-position snapshot before clearing. When this happens during an active `EventStream` (used by the TUI's event loop), the cursor-position read can fail immediately with "cursor position could not be read within a normal duration."
- **The fix:** By issuing the clear command directly on the backend and resetting the diff baseline, the helper avoids the problematic cursor-position query entirely.
- **Result:** Prevents startup crashes on terminals where the cursor query would race with the event stream, and provides a centralized, safe clearing mechanism for future development.


njbrake:

Nice fix, and the root-cause writeup matches what I arrived at independently. One gap: `clear_terminal` is an `App` method, so it only covers the four `terminal.clear()` sites in `src/tui/app.rs`. There is a fifth at `src/tui/remote_home/mod.rs:169`, the clear after returning from the remote structured view:

```rust
super::structured_view::run_for_endpoint(
    terminal,
    event_stream,
    theme,
    endpoint,
    &session.id,
)
.await?;
terminal.clear()?;
```

That path hands its `EventStream` into the structured view and clears on return, the same lifecycle as the acp view in `app.rs`, so it should be vulnerable to the identical `ESC[6n` failure. Because it lives in a different module it cannot reach the private `App::clear_terminal`, so it keeps calling `Terminal::clear` and stays exposed.

Two ways to close it: lift the helper into a shared free function (for example a `pub(crate) fn` in `src/tui/mod.rs`) and route the remote-home site through it, or add an equivalent clear there. Either keeps the "no clear path calls `get_cursor_position`" invariant true across the whole TUI rather than just the local home view.

src/tui/app.rs
```diff
@@ -435,6 +435,32 @@ impl App {
...
+    /// fires crossterm's waker, so the next read returns "cursor position could
+    /// not be read within a normal duration" immediately. Clearing the backend
+    /// directly (the same `ClearType::All` ratatui's Fullscreen `clear_viewport`
+    /// uses) and resetting the diff baseline via `swap_buffers` repaints every
+    /// cell on the next `draw` without any cursor query.
+    fn clear_terminal(
+        &mut self,
+        terminal: &mut Terminal<CrosstermBackend<std::io::Stdout>>,
+    ) -> Result<()> {
+        crossterm::execute!(
```
jerome-benoit: **The helper bypasses ratatui's `Backend::clear_region` API; that single API change collapses four problems into one fix.**

`ratatui-core 0.1.1` `Backend` trait at `src/backend.rs:157` declares (line 299):
```rust
fn clear_region(&mut self, clear_type: ClearType) -> Result<(), Self::Error>;
```
This is exactly the call ratatui's own `clear_viewport` makes for `Viewport::Fullscreen` (`buffers.rs:159`). `ClearType` is re-exported as `ratatui::backend::ClearType`. Going through `Backend::clear_region` rather than `crossterm::execute!` fixes:

1. **API canonicality**, the helper currently bypasses ratatui's abstraction in a code path whose only purpose is hitting the same `Backend` method ratatui already exposes for the same scenario.
2. **Visibility (root cause of the BLOCKING in the body)**, the helper is private on `impl App`, so `remote_home::run` cannot reach it and silently fell back to `terminal.clear()` at `src/tui/remote_home/mod.rs:169`, the 5th vulnerable site. A free function in `src/tui/mod.rs` (or a new `src/tui/terminal_ext.rs`) routes both modules through one definition.
3. **Receiver**, the body of `clear_terminal` never reads or writes `self`. The `&mut self` is dead weight that couples a TUI-wide invariant to one struct and prevents reuse.
4. **Test reachability**, the signature pins `CrosstermBackend<std::io::Stdout>`. To unlock a regression test on `TestBackend`, the obvious bound `<B: Backend + std::io::Write>` does NOT work: `TestBackend` declares only `impl Backend for TestBackend` at `ratatui-core-0.1.1/src/backend/test.rs:249` and has no `Write` impl. The correct bound is `<B: Backend>`, which `Backend::clear_region` accepts directly without needing `Write`.

Concrete shape:
```rust
// src/tui/mod.rs (or src/tui/terminal_ext.rs)
pub(crate) fn clear_terminal<B: ratatui::backend::Backend>(
    terminal: &mut ratatui::Terminal<B>,
) -> Result<(), B::Error> {
    use ratatui::backend::ClearType;
    terminal.backend_mut().clear_region(ClearType::All)?;
    terminal.current_buffer_mut().reset();
    terminal.swap_buffers();
    Ok(())
}
```
All five sites (the four in `app.rs` plus `remote_home/mod.rs:169`) call `crate::tui::clear_terminal(terminal)?`. A unit test against `TestBackend` then locks the invariant the PR claims:
```rust
#[test]
fn clear_terminal_resets_diff_baseline() {
    use ratatui::{Terminal, backend::TestBackend, buffer::Cell};
    let backend = TestBackend::new(20, 5);
    let mut terminal = Terminal::new(backend).unwrap();
    terminal.draw(|f| { /* paint something */ }).unwrap();
    crate::tui::clear_terminal(&mut terminal).unwrap();
    assert!(terminal
        .current_buffer_mut()
        .content()
        .iter()
        .all(|c| c == &Cell::default()));
}
```

src/tui/app.rs
```diff
@@ -435,6 +435,32 @@ impl App {
         Ok(())
     }
 
+    /// Clear the screen and force a full redraw on the next frame.
+    ///
+    /// NOTE: this deliberately avoids ratatui's `Terminal::clear`, which
+    /// snapshots the cursor with a `get_cursor_position` (`ESC[6n`) round-trip.
+    /// That read goes through crossterm's shared internal event reader and is
+    /// unreliable around `EventStream` lifecycle changes: dropping the stream
+    /// fires crossterm's waker, so the next read returns "cursor position could
```
jerome-benoit: **The causal explanation is plausible but imprecise enough to mislead a future fixer.**

The current wording, *"dropping the stream fires crossterm's waker, so the next read returns 'cursor position could not be read within a normal duration' immediately"*, elides two things:

1. it is the shared internal-event reader's poll that the waker interrupts, not the cursor-read syscall directly;
2. the immediate-fail path is `read_until_position` returning with no matching event (after the poll wakes early), not the 2-second timeout firing. The PR description correctly clarifies this as "a 0 ms failure, not the 2 s timeout the message implies", but that nuance is missing from this comment.

A future maintainer reading only this comment before another EventStream-lifecycle change may try to fix the cursor-read timeout itself rather than the underlying shared-reader race. Either tighten the wording to mention the shared internal reader and the early-completion path, or link to the upstream crossterm issue or commit, or shorten to *"avoid ratatui's clear because it does an `ESC[6n` round-trip that races EventStream lifecycle"* and drop the waker mechanism claim.

src/tui/app.rs
```diff
@@ -2315,10 +2340,10 @@ impl App {
         let result =
             crate::tui::structured_view::run(terminal, &mut stream, &self.theme, session_id).await;
         self.event_stream = Some(stream);
-        // Forcing a full redraw on return so the home screen redraws
-        // any cells the acp view painted over.
+        // Force a full redraw so the home screen repaints any cells the acp
+        // view painted over.
         self.needs_redraw = true;
-        terminal.clear()?;
+        self.clear_terminal(terminal)?;
```
jerome-benoit: **Lines 2345-2346 do `self.needs_redraw = true; self.clear_terminal(terminal)?;`. The main loop at lines 698-701 sees `needs_redraw == true` on the next iteration, calls `self.clear_terminal(terminal)?` again, and only then resets the flag.**

The helper at line 2346 already reset both buffers and forced a full repaint baseline, so the loop's call is redundant.

The shape (`terminal.clear()?` plus `needs_redraw = true`) predates this PR, so the redundancy is not introduced here. The new helper, however, is strictly more expensive than the old `terminal.clear()` (`Clear(All)` plus two buffer resets plus a swap, versus ratatui's old `Clear(All)` plus one back-buffer reset), and the redundancy now weakens the new invariant: a later refactor may drop one of the two and assume the other still holds.

Pick one mechanism:

- (a) drop the explicit `self.clear_terminal(terminal)?;` and rely on the loop's redraw branch (cleanest, since `needs_redraw=true` is already the project's idiomatic full-redraw signal); or
- (b) drop `self.needs_redraw = true;` since `clear_terminal` already reset the diff baseline so the very next `draw` repaints everything.

(a) is preferable because it keeps the redraw mechanism in one place.

jerome-benoit (changes requested):
# Review: fix(tui): clear screen without a cursor-position query

## Verdict
REQUEST_CHANGES on **1 BLOCKING** (corroborates @njbrake) plus **1 INCONSISTENT** architectural finding and **2 WARN**.

## Method
2 adversarial `deep` axes:
- **Axis A** correctness, ratatui 0.30 internals, helper semantic equivalence, EventStream reorder, buffer baseline.
- **Axis B** regression and compliance, AGENTS.md, doc-comment, test gap, `needs_redraw` interaction.

Plus 2 cross-validation `deep` axes:
- **Axis C** Rust best-practices, idiomatic crossterm/ratatui API choice, generic bounds, receiver discipline.
- **Axis D** review meta-audit, severity calibration, prose compliance, internal coherence.

Static analysis only; no real-tty reproduction attempted (the PR description itself notes the bug needs a real tty plus the crossterm `EventStream` lifecycle, so axis C confirmed via upstream source rather than runtime).

## Bug claim mechanically confirmed
`ratatui-core 0.1.1`, `src/terminal/buffers.rs:147`:
```rust
pub fn clear(&mut self) -> Result<(), B::Error> {
    let original_cursor = self.backend.get_cursor_position()?;
    self.clear_viewport()?;
    self.backend.set_cursor_position(original_cursor)?;
    Ok(())
}
```
For `Viewport::Fullscreen`, `clear_viewport` (same file, line 158) issues only `self.backend.clear_region(ClearType::All)?` and resets the back buffer (`buffers[1 - current].reset()`). The PR's `Clear(ClearType::All)` matches what ratatui itself emits for the Fullscreen case, minus the cursor save/restore. The fix is mechanically correct.

The helper does slightly more than ratatui's `clear`: it resets BOTH buffers (`current_buffer_mut().reset()` then `swap_buffers()` whose body resets `buffers[1 - self.current]`). Net post-state: both buffers empty, `current` flipped. ratatui's `clear_viewport` resets only the back. Functionally equivalent in this codebase (every render path paints a full frame), but the helper's stricter behaviour pre-empts stale leaks across the clear-to-next-draw boundary if a future render path leaves cells unpainted.

## BLOCKING

**5th `terminal.clear()` site at `src/tui/remote_home/mod.rs:169` is not routed through the helper.**

```rust
super::structured_view::run_for_endpoint(
    terminal,
    event_stream,
    theme,
    endpoint,
    &session.id,
)
.await?;
terminal.clear()?;   // still calls ratatui's cursor-querying clear
```

Same `EventStream` lifecycle as `App::with_raw_mode_disabled` and the acp-view return at `app.rs:2346`: the stream is borrowed into `structured_view::run_for_endpoint`, returned, then `terminal.clear()` triggers the `ESC[6n` round-trip through crossterm's still-shared internal reader. Same vulnerability class as the startup crash.

Commit message claim, "All four clear sites (...) route through the new clear_terminal helper", undercounts by one. The repo has five clear sites; the four in `app.rs` plus this one in `remote_home`. Corroborates @njbrake's earlier comment.

The architectural fix below (single inline INCONSISTENT) closes this site as a side effect.

## INCONSISTENT

**Helper bypasses ratatui's `Backend::clear_region` API; the resulting placement, signature, and receiver block both the TUI-wide invariant and the regression test.**

The helper at `app.rs:449` does:
```rust
fn clear_terminal(
    &mut self,
    terminal: &mut Terminal<CrosstermBackend<std::io::Stdout>>,
) -> Result<()> {
    crossterm::execute!(
        terminal.backend_mut(),
        crossterm::terminal::Clear(crossterm::terminal::ClearType::All)
    )?;
    terminal.current_buffer_mut().reset();
    terminal.swap_buffers();
    Ok(())
}
```

`ratatui-core 0.1.1` defines `pub trait Backend` at `src/backend.rs:157`, and on line 299:
```rust
fn clear_region(&mut self, clear_type: ClearType) -> Result<(), Self::Error>;
```
This is exactly the call ratatui's own `clear_viewport` makes for `Viewport::Fullscreen` (`buffers.rs:159`). `ClearType` is re-exported via `ratatui::backend::ClearType`. Going through `Backend::clear_region` rather than `crossterm::execute!` collapses four problems into one fix:

1. **API canonicality**, the helper currently bypasses ratatui's abstraction in a path whose only purpose is hitting the same `Backend` method ratatui already exposes.
2. **Visibility (root cause of BLOCKING)**, the helper is `fn` private to `impl App`, so `remote_home::run` cannot reach it and silently fell back to `terminal.clear()`. A free function in `src/tui/mod.rs` (or a new `src/tui/terminal_ext.rs`) routes both modules through one definition.
3. **Receiver**, the body never reads or writes `self`. The `&mut self` receiver is dead weight that couples a TUI-wide invariant to one struct.
4. **Test reachability**, the signature pins `CrosstermBackend<std::io::Stdout>`, blocking the obvious unit test that would lock the new invariant against `TestBackend`. **Caveat for any reader implementing this**, `TestBackend` does NOT implement `std::io::Write` (`ratatui-core-0.1.1/src/backend/test.rs:249` declares only `impl Backend for TestBackend`). A `<B: Backend + std::io::Write>` bound would EXCLUDE `TestBackend`, defeating the goal. The fix is to drop crossterm here and use `Backend::clear_region` directly, which needs only `<B: Backend>`.

Concrete shape:
```rust
// src/tui/mod.rs (or src/tui/terminal_ext.rs)
pub(crate) fn clear_terminal<B: ratatui::backend::Backend>(
    terminal: &mut ratatui::Terminal<B>,
) -> Result<(), B::Error> {
    use ratatui::backend::ClearType;
    terminal.backend_mut().clear_region(ClearType::All)?;
    terminal.current_buffer_mut().reset();
    terminal.swap_buffers();
    Ok(())
}
```
All five existing call sites (the four in `app.rs` plus `remote_home/mod.rs:169`) become `crate::tui::clear_terminal(terminal)?`. The unit test that locks the invariant becomes:
```rust
#[test]
fn clear_terminal_resets_diff_baseline() {
    use ratatui::{Terminal, backend::TestBackend, buffer::Cell};
    let backend = TestBackend::new(20, 5);
    let mut terminal = Terminal::new(backend).unwrap();
    terminal.draw(|f| { /* paint something */ }).unwrap();
    crate::tui::clear_terminal(&mut terminal).unwrap();
    assert!(terminal
        .current_buffer_mut()
        .content()
        .iter()
        .all(|c| c == &Cell::default()));
}
```
This locks the invariant the PR claims, against the only backend the project tests with, and removes the crossterm coupling on a path whose canonical implementation lives in ratatui itself.

## WARN (2)

1. **Doc-comment causal explanation is plausible but imprecise enough to mislead a future fixer** (`app.rs:443-445`). The wording, *"dropping the stream fires crossterm's waker, so the next read returns 'cursor position could not be read within a normal duration' immediately"*, elides two things: (a) it is the shared internal-event reader's poll that the waker interrupts, not the cursor-read syscall directly, and (b) the immediate-fail path is `read_until_position` returning with no matching event after the poll wakes early, not the 2-second timeout firing (which the PR description correctly clarifies as "a 0 ms failure, not the 2 s timeout the message implies"). A future maintainer reading only this comment may try to fix the cursor-read timeout itself rather than the underlying shared-reader race.

2. **Double-clear after `structured_view::run` return** (`app.rs:2345-2346`). `self.needs_redraw = true; self.clear_terminal(terminal)?;` runs the helper, then on the next loop iteration the redraw branch at `app.rs:698-701` sees `needs_redraw == true` and calls `clear_terminal` a second time before resetting the flag. Two `Clear(All)` writes per acp-view return. The shape (`terminal.clear()` plus `needs_redraw = true`) predates the PR; it is not a regression introduced here. With the new helper doing strictly more work per call (`Clear(All)` plus two buffer resets plus a swap, versus ratatui's old single back-buffer reset), the redundancy now weakens the new invariant. Pick one mechanism: drop the explicit `clear_terminal` call here and rely on the loop's redraw branch (cleanest, since `needs_redraw=true` is the project's idiomatic full-redraw signal), or drop `needs_redraw = true` since `clear_terminal` already reset the diff baseline.

## Positive observations

Three points worth crediting; none surfaced as inline:

- **Bug diagnosis is precise.** Identifying ratatui 0.30's `Terminal::clear` cursor query as the regression vs 0.29 is the correct root cause; the upstream source confirms it.
- **EventStream reorder in `with_raw_mode_disabled`** (line 511, `event_stream = Some(EventStream::new())` moved AFTER `enable_raw_mode` plus alt-screen re-entry plus `sync_mouse_capture`) is benign and arguably safer; the new stream is born into raw mode rather than being attached to a briefly-cooked tty. No regression. A short comment addition like "(after raw mode plus alt-screen are restored, so the stream is born into raw mode)" would lock the intent against a future revert.
- **Compliance posture clean.** No `let _ =`, no emdash separators in the new doc-comment or commit message, no `#[allow(dead_code)]`, no internal codes (only `ESC[6n`, a real terminal sequence), no swallowed `Result` (both `crossterm::execute!` calls bubble via `?`), no comment that merely restates the code.

## DEFERRED

- **Codecov**, AGENTS.md notes Rust backend coverage is queued; `codecov/patch` is N/A for `src/tui/app.rs` and `codecov/project` carries forward. Confirmed.
- **Stale `terminal.clear()` references in 5 doc-comment locations** survived the rename: `app.rs:1212`, `app.rs:2267`, `home/input.rs:891`, `home/input.rs:1475`, `dialogs/intro.rs:229`. Mechanical sweep, no behavioral risk; readers grepping for `terminal.clear` will land on dead doc references. Not blocking; can ride on the same follow-up that addresses the architectural fix.
- **Non-Fullscreen viewport** (`Inline`, `Fixed`) behavior is unaddressed. The helper's `Clear(All)` is correct for `Viewport::Fullscreen` only; the project uses Fullscreen, so out of scope here. If `Backend::clear_region` is adopted (INCONSISTENT above), the helper still hardcodes `ClearType::All`, which remains Fullscreen-specific. Worth a one-line note in the doc-comment.

---
3 inline comments below: 1 INCONSISTENT plus 2 WARN. The BLOCKING lives in this body since `remote_home/mod.rs` is outside the PR diff.

jerome-benoit:
Method note for context: this REQUEST_CHANGES review was produced via 2 adversarial `deep` axes (correctness, regression+compliance) plus 2 cross-validation `deep` axes (Rust best-practices, review meta-audit). Static analysis only; no real-tty reproduction attempted (consistent with the PR description's framing).

The first draft of this review proposed a `<B: Backend + std::io::Write>` bound to unlock a `TestBackend` regression test. That bound is broken: `ratatui-core 0.1.1` declares only `impl Backend for TestBackend` at `src/backend/test.rs:249`; there is no `Write` impl, so the bound would exclude the very backend we want to test against. The cross-validation axis caught it before submission. The submitted INCONSISTENT replaces it with a `<B: Backend>`-only signature that delegates to `Backend::clear_region(ClearType::All)` directly (the same call ratatui's own `clear_viewport` makes for `Viewport::Fullscreen` at `buffers.rs:159`), removing the crossterm dependency on this hot path and unlocking the test.

If the architectural fix in INCONSISTENT is taken, the BLOCKING (5th site at `src/tui/remote_home/mod.rs:169`) closes for free.

njbrake (approved):

Approving. The root-cause fix is correct: ratatui 0.30's `Terminal::clear` does an `ESC[6n` cursor round-trip through crossterm's shared internal reader, which races the live `EventStream` and aborts startup. Verified against ratatui-core 0.1.1 source that `clear_region(ClearType::All)` is the exact operation Fullscreen `clear_viewport` issues, minus the cursor save/restore.

The follow-up commit closes the review items: the helper is now a shared `pub(crate) fn clear_terminal<B: Backend>` going through ratatui's `Backend` API, the fifth clear site in `remote_home` (same `EventStream` lifecycle, previously unreachable from the private method) routes through it, the redundant double-clear after `run_structured_view` is gone, and a `TestBackend` regression test locks the real invariant (clear then identical redraw must repaint, not diff to a no-op).

Checks are green locally: fmt, clippy `-D warnings` with `--features serve`, build, and the TUI lib suite. Thanks for the precise writeup, @peteski22.

Issue #1868, opened 2026-06-02
Archive paths kill only the agent tmux session, not terminal/container-terminal/tool sub-sessions


jerome-benoit:
## Summary

When a session is archived (CLI `aoe session archive`, TUI `'z'` keybind, or `PATCH /api/sessions/:id/archive`), only the agent's tmux session is killed. The session's web terminal pane, container terminal, and any tool sub-sessions remain alive in tmux as orphans.

## Affected sites

- `src/cli/session.rs:286` (`archive_session` CLI): calls `inst.kill()` only.
- `src/tui/home/operations.rs:887` (`toggle_archive_at_cursor`): calls `inst.kill()` only.
- `src/server/api/sessions.rs:1271` (`update_session_archive` web handler): calls `inst.kill()` only.

All three call only `Instance::kill()` (which kills the agent session via `Session::kill`). None call `kill_terminal()`, `kill_container_terminal()`, or `kill_all_tool_sessions_for_id`. The new canonical `Instance::kill_all_tmux_sessions()` helper added in #1867 is not invoked.

## Why this is filed as a question, not a bug

A thorough audit + RCA performed during the work that produced #1867 classified this as **ambiguous**.

**Argument for "intentional"**:
- A user may want their open terminal panel kept open across an archive for later debug.
- The web API exposes a `kill_pane=false` body field, suggesting the tmux side-effect is opt-in rather than mandatory.
- Symmetry with snooze: snoozing does not kill anything either.

**Argument for "oversight"**:
- The semantics are not documented at any of the three sites.
- `kill_pane` only governs the agent session; there is no equivalent granular control for terminal/cterm/tool sessions.
- The user-visible cleanup story is inconsistent: archive sinks the row but leaves visible tmux artifacts (`tmux ls` still shows them).

## Decision needed

Pick one:

- **(a) Status quo**: Archive paths kill agent only. Document this explicitly at all three sites and in the user-facing archive doc.
- **(b) Full teardown**: Archive paths use `Instance::kill_all_tmux_sessions()` from #1867. All four tmux session kinds are killed on archive. This makes archive symmetric with delete from the tmux state perspective.
- **(c) Granular flags**: Extend the API/CLI/TUI to allow user control per session kind (`kill_pane` already exists; add `kill_terminal`, `kill_tool_sessions`). Higher complexity but most flexible.

## Acceptance criteria

Whichever of (a)/(b)/(c) is chosen:
[...]

PR #2227, opened 2026-06-16
fix(session): keep archived sessions out of Error when their tmux is gone

### LINKED ISSUE #2206: TUI: archived session shows as errored in web, status poller flips it to Error after archive kills tmux

## Problem

Create a session in the TUI, archive it from the TUI, then open the web dashboard. The session shows up in the Archived list rendered as **errored**, even though nothing went wrong; the user deliberately archived it.

The status the web shows is `Error` with the message "tmux session is gone". That is technically true (archiving kills the tmux session on purpose, see #1868 / commit `42eb3b1a`), but a session being archived is not an error state. The Archived list should show these as stopped/dormant, not failed.

## What's there today

Archiving from the TUI does two things, in order. First it tears down all tmux for the session, then it flips the archived flag:

`src/tui/home/operations.rs:1279`
```rust
// Tear down all tmux before flipping archived. #1868.
if let Some(inst) = self.instances.iter().find(|i| i.id == id) {
    inst.kill_all_tmux_sessions();
}
// ...
self.apply_user_action(&id, |inst| inst.archive())?;
```

`archive()` only stamps `archived_at`; it never touches `status`:

`src/session/instance.rs:1142`
```rust
pub fn archive(&mut self) {
    self.archived_at = Some(Utc::now());
    self.favorited_at = None;
    self.snoozed_until = None;
    self.pinned_at = None;
}
```

`kill_all_tmux_sessions(&self)` takes `&self`, so it cannot change `status` either. A freshly created session sits at `Status::Idle`, and that is the status it keeps right after archiving.

The unarchive path even documents the intended resting state as Stopped, but nothing actually sets it:

`src/tui/home/operations.rs:1273`
```rust
// The session stays Stopped (archive killed its panes); the user restarts it with
// `e` when they want it back, same as any other stopped session.
```

## Root cause

The status poller does not skip archived sessions, and the status updater has no archived guard before it treats a missing tmux session as an error.

Layer 1, the poller still polls archived (`Idle`) rows on the WARM tier:

`src/tui/status_poller.rs:30`
```rust
fn polling_tier(status: Status) -> u64 {
    match status {
        Status::Running | Status::Waiting | Status::Starting => TIER_HOT,
        Status::Idle | Status::Unknown => TIER_WARM,
        Status::Error => TIER_COLD,
        Status::Stopped | Status::Deleting | Status::Creating => 0, // Never poll
    }
}
```

An archived session is `Idle`, so it is polled. (Note: `Stopped` is tier 0, never polled. That is the tell for the fix below.)

Layer 2, `update_status_with_metadata_inner` returns early for structured/ACP sessions but has no `is_archived()` guard for plain tmux sessions, so the "tmux is gone" branch fires and stamps `Error`:

`src/session/instance.rs:3384`
```rust
if !session.exists() {
    // ...
    self.status = Status::Error;
    if self.last_error.is_none() {
        self.last_error = Some(TMUX_SESSION_GONE_ERROR.to_string());
    }
    self.last_error_check = Some(std::time::Instant::now());
    return;
}
```

Compare the structured-session early return just above, which exists precisely to avoid this spurious "tmux session is gone" transition:

`src/session/instance.rs:3337`
```rust
#[cfg(feature = "serve")]
if self.is_structured() {
    // ... clears stale TMUX_SESSION_GONE_ERROR, returns
    return;
}
```

Layer 3, the server serializes the raw status verbatim, so `Error` reaches the web with no archived-aware remapping:

`src/server/api/sessions.rs:227`
```rust
status: format!("{:?}", inst.status),
```

`web/src/lib/types.ts` lists `"Error"` as a `SessionStatus` and the frontend renders whatever the API sends, so the archived row paints as errored.

`is_archived()` already exists at `src/session/instance.rs:1153`, so the guard is cheap to add.

## Proposal

**(A) Minimum fix: guard the status updater.** Add an `is_archived()` early return in `update_status_with_metadata_inner`, mirroring the existing structured-session guard, before the `tmux_session()` / `!session.exists()` checks (`src/session/instance.rs` around 3325-3384). An archived session has no tmux by design, so it should not be re-probed at all.

```rust
if self.is_archived() {
    return;
}
```

**(B) Structural fix: set the documented resting state.** Make `archive()` set `status = Status::Stopped` (`src/session/instance.rs:1142`). This matches the intent already written at `operations.rs:1273` ("stays Stopped"), and because `Stopped` is polling tier 0 the poller stops touching the row entirely, so layer 1 and layer 2 both stop misfiring. Unarchive already leaves the row Stopped per its own comment, so the round trip stays consistent.

A and B are complementary: B fixes the resting state and stops the polling; A is the defensive guard so any future code path that archives without setting Stopped still cannot paint an archived row as errored. Shipping both is the robust shape; B alone is the minimum that resolves the reported symptom.

**(C) Optional: clear `last_error`.** If a session was genuinely `Error` before being archived, decide whether archiving should clear `last_error` / `TMUX_SESSION_GONE_ERROR` so the archived row never carries a stale error message. See Open questions.

## Why now

- Regression-shaped: archiving started tearing down tmux in #1868 (`42eb3b1a`), which is what exposes the missing guard.
- User-visible mislabel on a deliberate action; "errored" reads as "something broke" when the user simply archived.
- Small, contained fix in session-core with a clear test surface.

- As a TUI user, when I archive an idle session, then its stored status becomes `Stopped` (not `Error`) and no `TMUX_SESSION_GONE_ERROR` is recorded.
  - Suggested test layer: Rust unit test in `src/session/instance.rs` (`#[cfg(test)]`): build an instance, call `archive()`, run `update_status_with_metadata` with no live tmux, assert status is not `Error`.
- As a TUI user, when an archived session is polled by the status poller, then it is not re-probed into an `Error` state because its tmux is intentionally gone.
  - Suggested test layer: Rust unit test exercising `update_status_with_metadata_inner` on an archived instance, asserting the `is_archived()` guard short-circuits before the `!session.exists()` branch.
- As a web dashboard user, when I open the Archived list after archiving a session in the TUI, then the session does not render as errored.
  - Suggested test layer: `web/tests/live/` Playwright, or a server-level test asserting the serialized status for an archived session is not `"Error"`.

## Open questions

- Pick the fix shape: A + B (recommended), or B alone? B changes the on-disk status semantics of an archived session from "whatever it was" to `Stopped`; confirm no surface keys off the pre-archive status of an archived row.
- Should archiving a session that was already in `Error` clear `last_error` (proposal C), or preserve it so the user can see why it failed before archiving?
- TUI parity: does the TUI Archived section also surface this `Error` today, or only the web? The status is wrong regardless of surface; worth confirming whether the TUI happens to mask it so the fix is validated on both.

## Out of scope

- Reworking the polling-tier model or the structured/ACP status path.
- Any change to delete vs archive semantics.

## Cross-references

- #1868 — introduced tmux teardown on archive (commit `42eb3b1a`), the behavior that exposes this missing guard.

Seluj78:

## Description

Archiving a session from the TUI tears down its tmux on purpose (#1868). But `archive()` left `status` untouched, and the status poller then probed the now-missing tmux and flipped the row to `Error` with a "tmux session is gone" message. The web dashboard serializes that status verbatim, so an archived session rendered as **errored** in the Archived list even though nothing failed.

The fix adds an archived guard in `update_status_with_metadata_inner` (`src/session/instance.rs`), mirroring the existing structured-session guard right above it: an archived session has no tmux by design, so the poller short-circuits with `if self.is_archived() { return; }` and never re-probes it into `Error`. `archive()` only stamps `archived_at`, so the guard keeps archive/unarchive status-preserving (a genuinely errored session keeps its `Error` status), which the attention-tier and recovery-eligibility logic depend on.

Rows already persisted as `Error` by a pre-fix build cannot be healed in the poller: `last_error` is `#[serde(skip)]`, so on reload such a row carries only `status = "error"` with no message to match on. A one-shot `v016` migration (`src/migrations/v016_clear_archived_tmux_gone_error.rs`) walks every sessions.json on boot and demotes any archived row still at `error` to `idle`. This follows the AGENTS.md rule of using a migration rather than an inline compat shim.

Closes #2206

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] In the TUI, create a session, then archive it (`a`).
- [ ] Open the web dashboard, expand the Archived list, confirm the session does not show as errored.
- [ ] In the TUI, let a session genuinely error (kill its agent), then archive it; confirm it still shows its error state, not a healed/idle state.
- [ ] Archive a session, leave and re-enter the TUI; confirm the archived row is not flipped to errored after the status poller runs.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the bug is pure status-computation plus a data migration, covered by deterministic Rust unit tests. In `src/session/instance.rs`: `test_archived_session_not_marked_error_when_tmux_gone` is genuinely red on the pre-fix tree; `test_archived_session_preserves_genuine_error` and `test_archived_unarchived_genuine_error_roundtrips` are regression guards for the status-preserving contract. In `src/migrations/v016_clear_archived_tmux_gone_error.rs`: tests cover the heal, leaving non-archived and non-error rows untouched, a missing file, and a corrupt file. An e2e through tmux adds no signal these lack.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (dropping the status-mutation approach once it broke archive/unarchive symmetry, then switching from an inline heal to a v016 migration after review), reviewed each commit, and ran the test steps above.

**NOTE:**

## Overview

This PR fixes a web dashboard issue where sessions that were archived from the TUI could incorrectly show up as “errored.” The root cause was that the background status poller kept checking tmux for sessions that had already been intentionally torn down when archived, and then persisted the “tmux session is gone” error footprint.

### Core behavior
- **`src/session/instance.rs`**: Added an **archived-session guard** to stop the status update logic from re-probing archived instances (so they don’t get flipped into `Error` just because tmux is gone).
- Added **targeted healing** so the specific persisted “tmux session is gone” error footprint on archived rows is corrected back to a non-error state rather than treated as a real failure.

### Data migration
- **Migration v016 (`src/migrations/v016_clear_archived_tmux_gone_error.rs`)**:
  - Scans session state files (both per-profile and legacy top-level).
  - For any session marked `error` **only when `archived_at` is present**, it demotes it back to `idle`.
  - Includes safe handling for missing files and corrupt JSON (skips without failing boot).
- **`src/migrations/mod.rs`**: Bumped the migration target to **v16** and registered the new migration.

### Tests
- Added **three deterministic Rust unit tests** in `src/session/instance.rs` verifying:
  - archived sessions are not converted into `Error` during polling,
  - genuine pre-set `Error` states are preserved,
  - the tmux-gone error footprint on archived sessions is corrected.

## Benefits
- Archived sessions no longer appear as broken in the dashboard, eliminating confusing “tmux session is gone” messaging.
- Any already-persisted archived “tmux-gone” error rows are automatically cleaned up via migration (and/or corrected on subsequent polling).
- Real errors remain real: the PR preserves the intended archive/unarchive semantics so only the specific tmux-gone artifact is healed.

## Technical Details (if you need them)

- The status poller now **short-circuits** for archived sessions inside `Instance::update_status_with_metadata_inner`, preventing tmux probing from changing archived rows’ state.
- The repo adds a compatibility/data-fixing approach via **migration (v016)** to clear archived rows that were previously persisted in the `error` state for the tmux-gone case.

### DIFF
--- a/src/migrations/v016_clear_archived_tmux_gone_error.rs
+++ b/src/migrations/v016_clear_archived_tmux_gone_error.rs
@@ -1,0 +1,140 @@
+//! Migration v016: clear the spurious tmux-gone Error left on archived
+//! sessions by builds shipped between #1868 (tear down tmux on archive) and
+//! #2206 (stop the status poller flipping archived rows to Error).
+//!
+//! Those builds archived a session, killed its tmux, then let the status
+//! poller observe the missing tmux and stamp `status = "error"` with a
+//! "tmux session is gone" message. `last_error` is not persisted
+//! (`#[serde(skip)]`), so on reload the row carries only `status = "error"`
+//! and the message is gone; the in-process #2206 guard cannot recognize it.
+//! This one-shot migration walks every sessions.json and demotes any archived
+//! row still sitting at Error back to Idle. An archived row has no live tmux
+//! by design, so an Error status on one can only be that spurious transition.
+//!
+//! ## Failure policy
+//!
+//! Per `AGENTS.md > Data Migrations`, a returned `Err` aborts boot. A
+//! sessions.json that fails to parse is logged and skipped (a corrupt file
+//! must not block boot or spam every launch). Only `get_app_dir` and
+//! directory-read failures propagate.
+
+use anyhow::Result;
+use std::fs;
+use std::path::Path;
+use tracing::{debug, info};
+
+pub fn run() -> Result<()> {
+    let app_dir = crate::session::get_app_dir()?;
+    run_in(&app_dir)
+}
+
+pub(crate) fn run_in(app_dir: &Path) -> Result<()> {
+    let profiles_dir = app_dir.join("profiles");
+    if profiles_dir.exists() {
+        for entry in fs::read_dir(&profiles_dir)? {
+            let entry = entry?;
+            if entry.path().is_dir() {
+                clear_archived_error(&entry.path().join("sessions.json"))?;
+            }
+        }
+    }
+    // Legacy top-level sessions.json (pre-profiles layout).
+    clear_archived_error(&app_dir.join("sessions.json"))?;
+    Ok(())
+}
+
+/// Demote any archived session still persisted at `status = "error"` back to
+/// `"idle"`. Leaves non-archived rows and archived rows in any other status
+/// untouched.
+fn clear_archived_error(path: &Path) -> Result<()> {
+    if !path.exists() {
+        return Ok(());
+    }
+    let content = fs::read_to_string(path)?;
+    let mut value: serde_json::Value = match serde_json::from_str(&content) {
+        Ok(v) => v,
+        Err(e) => {
+            debug!("v016: failed to parse {}: {e}, skipping", path.display());
+            return Ok(());
+        }
+    };
+
+    let mut healed = 0usize;
+    if let Some(array) = value.as_array_mut() {
+        for instance in array.iter_mut() {
+            if let Some(obj) = instance.as_object_mut() {
+                let archived = obj.get("archived_at").is_some_and(|v| !v.is_null());
+                let errored = obj.get("status").and_then(|v| v.as_str()) == Some("error");
+                if archived && errored {
+                    obj.insert(
+                        "status".to_string(),
+                        serde_json::Value::String("idle".to_string()),
+                    );
+                    healed += 1;
+                }
+            }
+        }
+    }
+
+    if healed > 0 {
+        fs::write(path, serde_json::to_string_pretty(&value)?)?;
+        info!(
+            "v016: cleared spurious archived Error on {healed} session(s) in {} (#2206)",
+            path.display()
+        );
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn clears_only_archived_error_rows() {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("sessions.json");
+        fs::write(
+            &path,
+            r#"[
+                {"id":"a","status":"error","archived_at":"2026-06-05T16:04:12Z"},
+                {"id":"b","status":"error"},
+                {"id":"c","status":"idle","archived_at":"2026-06-05T16:04:12Z"},
+                {"id":"d","status":"stopped","archived_at":"2026-06-05T16:04:12Z"},
+                {"id":"e","status":"error","archived_at":null}
+            ]"#,
+        )
+        .unwrap();
+
+        clear_archived_error(&path).unwrap();
+
+        let v: serde_json::Value =
+            serde_json::from_str(&fs::read_to_string(&path).unwrap()).unwrap();
+        let arr = v.as_array().unwrap();
+        // archived + error -> idle (the bug footprint)
+        assert_eq!(arr[0]["status"], "idle");
+        // non-archived error -> untouched
+        assert_eq!(arr[1]["status"], "error");
+        // archived non-error -> untouched
+        assert_eq!(arr[2]["status"], "idle");
+        assert_eq!(arr[3]["status"], "stopped");
+        // explicit null archived_at counts as non-archived -> untouched
+        assert_eq!(arr[4]["status"], "error");
+    }
+
+    #[test]
+    fn missing_file_is_ok() {
+        let dir = tempfile::tempdir().unwrap();
+        clear_archived_error(&dir.path().join("does-not-exist.json")).unwrap();
+    }
+
+    #[test]
+    fn corrupt_file_is_skipped_not_an_error() {
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("sessions.json");
+        fs::write(&path, "{ not valid json").unwrap();
+        // Must not error; a corrupt file is left untouched.
+        clear_archived_error(&path).unwrap();
+        assert_eq!(fs::read_to_string(&path).unwrap(), "{ not valid json");
+    }
+}
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1144,6 +1144,11 @@ impl Instance {
         self.favorited_at = None;
         self.snoozed_until = None;
         self.pinned_at = None;
+        // Archiving tears down the session's tmux on purpose (#1868), so its
+        // resting state is Stopped, not whatever it was running as. Stopped is
+        // polling tier 0, so the status poller stops re-probing the row and
+        // can never flip it to Error for a tmux that is gone by design (#2206).
+        self.status = Status::Stopped;
     }
 
     pub fn unarchive(&mut self) {
@@ -3330,6 +3335,21 @@ impl Instance {
             return;
         }
 
+        // Archived sessions have their tmux torn down on purpose (#1868), so
+        // probing tmux here only ever produces a spurious "tmux session is
+        // gone" Error transition (#2206). Heal a session that was already
+        // persisted as Error before this guard existed, then short-circuit so
+        // the poller never re-probes a row whose tmux is gone by design.
+        if self.is_archived() {
+            if self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR) {
+                self.last_error = None;
+            }
+            if self.status == Status::Error {
+                self.status = Status::Stopped;
+            }
+            return;
+        }
+
         // Acp-mode sessions are not backed by a tmux pane; the structured view
         // worker supervisor owns their lifecycle and emits typed health
         // events over the broadcast. Probing tmux here only ever produces
@@ -3978,6 +3998,36 @@ mod tests {
         inst.touch_last_accessed();
         assert!(!inst.is_archived());
         assert!(inst.last_accessed_at.is_some());
+    }
+
+    #[test]
+    fn test_archived_session_not_marked_error_when_tmux_gone() {
+        // #2206: archiving kills the session's tmux on purpose. A subsequent
+        // status poll must not flip the archived row to Error for the missing
+        // tmux; it should rest at Stopped with no tmux-gone error recorded.
+        let mut inst = Instance::new("test", "/tmp/test");
+        inst.archive();
+        inst.update_status_with_metadata(None);
+        assert_ne!(inst.status, Status::Error);
+        assert_eq!(inst.status, Status::Stopped);
+        assert_eq!(inst.last_error, None);
+    }
+
+    #[test]
+    fn test_archived_session_heals_stale_tmux_gone_error() {
+        // #2206: a session archived before this guard existed may already be
+        // persisted as Error with a tmux-gone message. The next poll heals it
+        // to Stopped and clears the stale error.
+        let mut inst = Instance::new("test", "/tmp/test");
+        inst.status = Status::Error;
+        inst.last_error = Some(TMUX_SESSION_GONE_ERROR.to_string());
+        inst.archive();
+        // archive() resets status to Stopped, but force Error back to simulate
+        // a row loaded from disk that predates the archive() status reset.
+        inst.status = Status::Error;
+        inst.update_status_with_metadata(None);
+        assert_eq!(inst.status, Status::Stopped);
+        assert_eq!(inst.last_error, None);
     }
 
     #[test]
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -3333,16 +3333,10 @@ impl Instance {
         // Archived sessions have their tmux torn down on purpose (#1868), so
         // probing tmux here only ever produces a spurious "tmux session is
         // gone" Error transition (#2206). Short-circuit so the poller never
-        // re-probes a row whose tmux is gone by design; archive/unarchive must
-        // stay status-preserving, so only the bug's own footprint (Error with
-        // the tmux-gone message) is healed back to Idle.
+        // re-probes a row whose tmux is gone by design; this keeps
+        // archive/unarchive status-preserving. Rows already persisted as Error
+        // by a pre-fix build are cleaned up once by the v016 migration.
         if self.is_archived() {
-            if self.status == Status::Error
-                && self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)
-            {
-                self.status = Status::Idle;
-                self.last_error = None;
-            }
             return;
         }
 
@@ -4000,8 +3994,8 @@ mod tests {
     fn test_archived_session_not_marked_error_when_tmux_gone() {
         // #2206: archiving kills the session's tmux on purpose. A subsequent
         // status poll must not flip the archived row to Error for the missing
-        // tmux; archive is status-preserving, so an idle row stays Idle with
-        // no tmux-gone error recorded.
+        // tmux; the archived guard short-circuits, so an idle row stays Idle.
+        // Red on the pre-fix tree, where the tmux probe stamps Error.
         let mut inst = Instance::new("test", "/tmp/test");
         inst.archive();
         inst.update_status_with_metadata(None);
@@ -4011,28 +4005,32 @@ mod tests {
     }
 
     #[test]
-    fn test_archived_session_heals_stale_tmux_gone_error() {
-        // #2206: a session archived by an older build may already be persisted
-        // as Error with the tmux-gone message. The next poll heals exactly
-        // that footprint back to Idle and clears the stale error.
-        let mut inst = Instance::new("test", "/tmp/test");
-        inst.archive();
-        inst.status = Status::Error;
-        inst.last_error = Some(TMUX_SESSION_GONE_ERROR.to_string());
-        inst.update_status_with_metadata(None);
-        assert_eq!(inst.status, Status::Idle);
-        assert_eq!(inst.last_error, None);
-    }
-
-    #[test]
     fn test_archived_session_preserves_genuine_error() {
-        // #2206: a session that genuinely errored (a non-tmux-gone message)
-        // and was then archived must keep its Error status, so archive then
-        // unarchive stays status-preserving for real failures.
+        // #2206 regression guard (passes on both trees): the archived guard
+        // never mutates status, so a genuinely errored session keeps its Error
+        // state while archived. The legacy on-disk footprint is cleaned up by
+        // the v016 migration, not by the poller.
         let mut inst = Instance::new("test", "/tmp/test");
         inst.archive();
         inst.status = Status::Error;
         inst.last_error = Some("agent crashed".to_string());
+        inst.update_status_with_metadata(None);
+        assert_eq!(inst.status, Status::Error);
+        assert_eq!(inst.last_error.as_deref(), Some("agent crashed"));
+    }
+
+    #[test]
+    fn test_archived_unarchived_genuine_error_roundtrips() {
+        // #2206: archive then unarchive must stay status-preserving for a real
+        // failure. The archived guard leaves Error untouched; after unarchive
+        // the tmux probe re-stamps Error and its is_none() guard preserves the
+        // original message regardless of whether tmux is installed on the box.
+        let mut inst = Instance::new("test", "/tmp/test");
+        inst.archive();
+        inst.status = Status::Error;
+        inst.last_error = Some("agent crashed".to_string());
+        inst.update_status_with_metadata(None);
+        inst.unarchive();
         inst.update_status_with_metadata(None);
         assert_eq!(inst.status, Status::Error);
         assert_eq!(inst.last_error.as_deref(), Some("agent crashed"));
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1144,11 +1144,6 @@ impl Instance {
         self.favorited_at = None;
         self.snoozed_until = None;
         self.pinned_at = None;
-        // Archiving tears down the session's tmux on purpose (#1868), so its
-        // resting state is Stopped, not whatever it was running as. Stopped is
-        // polling tier 0, so the status poller stops re-probing the row and
-        // can never flip it to Error for a tmux that is gone by design (#2206).
-        self.status = Status::Stopped;
     }
 
     pub fn unarchive(&mut self) {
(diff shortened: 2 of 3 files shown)


src/session/instance.rs
```diff
@@ -3330,6 +3330,22 @@ impl Instance {
...
 
+        // Archived sessions have their tmux torn down on purpose (#1868), so
+        // probing tmux here only ever produces a spurious "tmux session is
+        // gone" Error transition (#2206). Short-circuit so the poller never
+        // re-probes a row whose tmux is gone by design; archive/unarchive must
+        // stay status-preserving, so only the bug's own footprint (Error with
+        // the tmux-gone message) is healed back to Idle.
+        if self.is_archived() {
+            if self.status == Status::Error
+                && self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)
```
jerome-benoit: **Heal predicate cannot fire on the persisted footprint that test 2 claims to cover.**

`last_error` carries `#[serde(skip)]` at `instance.rs:551-552`; `status` at line 377 is serialized. Test 2's comment at 4015-4017 frames this heal as covering rows "already persisted as Error with the tmux-gone message by an older build". On daemon restart, the row reloads with `status = Status::Error` and `last_error = None` (the serialized state never captured the message). The heal predicate at this line is then false (`None != Some(TMUX_SESSION_GONE_ERROR)`), the function early-returns at 3346, and the row stays at `Status::Error` on disk for the rest of its life. The web dashboard renders it as errored despite the fix shipping.

In-process recovery (Error stamped now, heal fires on the next poll within the same daemon generation) is genuinely fixed; the cross-restart legacy footprint is unreachable.

Two ways to close it:

- (a) loosen the predicate to `if self.status == Status::Error { ... }` for archived rows, dropping the `last_error` arm. The only legitimate paths to `Error` for an archived row are tmux-gone and the cross-restart resurrection of the same; widening matches the PR's stated intent and removes the dead-code surface.
- (b) keep the predicate exact, drop the cross-restart claim from test 2's comment, and add a v016 startup migration (`src/migrations/v016_clear_archived_tmux_gone_error.rs`) that walks `sessions.json` once on boot. CURRENT_VERSION = 15 and v015_rewrite_hook_strings is precedent. The per-poll heal can then be removed entirely (the archived guard reduces to `if self.is_archived() { return; }`).

(a) is the smaller change; (b) is architecturally cleaner.

src/session/instance.rs
```diff
@@ -3330,6 +3330,22 @@ impl Instance {
...
+            {
+                self.status = Status::Idle;
+                self.last_error = None;
```
jerome-benoit: **The heal is asymmetric with the existing Error-clearing block at lines 2924-2927.**

The sibling pattern, run when an Error transition is undone, sets:
```rust
if self.status == Status::Error {
    self.status = Status::Idle;
    self.last_error = None;
    self.last_error_check = None;   // throttle marker
}
```
The new heal here writes only `self.status = Status::Idle; self.last_error = None;`. The throttle marker `last_error_check` is left at its prior `Instant::now()` from the tmux-gone setter at line 3410.

Benign while the row stays archived (the archived guard early-returns at 3346, never reaching the throttle check at 3367-3373). On `unarchive()` the stale `last_error_check` Instant survives. If the row then errors for a real reason within 30 seconds of the original `Instant::now()`, the throttle at 3369 (`if last_check.elapsed().as_secs() < 30`) skips a legitimate re-poll, hiding the genuine error for one cycle.

Add `self.last_error_check = None;` to the heal block. Mirror lines 2924-2927 verbatim.

src/session/instance.rs
```diff
@@ -3330,6 +3330,22 @@ impl Instance {
...
+        // probing tmux here only ever produces a spurious "tmux session is
+        // gone" Error transition (#2206). Short-circuit so the poller never
+        // re-probes a row whose tmux is gone by design; archive/unarchive must
+        // stay status-preserving, so only the bug's own footprint (Error with
+        // the tmux-gone message) is healed back to Idle.
+        if self.is_archived() {
+            if self.status == Status::Error
+                && self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)
+            {
+                self.status = Status::Idle;
```
jerome-benoit: **The heal mutates persisted state with no log line; the surrounding poller logs every transition decision.**

Lines 3385, 3401, 3425, 3441, 3489, 3505, 3528 in this same function use `tracing::trace!(target: "session.store", session_id = %self.id, ...)` for every status transition. The heal fires here with no log, and the persisted `last_error` text is destroyed in the process.

A maintainer asking "why did this archived row's tmux-gone error message disappear after upgrade?" must read source to find this branch. A single info-level line, fired once per migrated row before the mutations:

```rust
if self.status == Status::Error
    && self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)
{
    tracing::info!(
        target: "session.store",
        session_id = %self.id,
        "healed archived tmux-gone error to Idle (#2206)"
    );
    self.status = Status::Idle;
    self.last_error = None;
    self.last_error_check = None;  // see I2
}
```

`info!` (not `trace!`) because this is a one-shot mutation per migrated row, not a per-poll decision. If the v016 migration option from the INCONSISTENT finding is taken instead, the migration logs per-row at startup and this inline log becomes redundant.

src/session/instance.rs
```diff
@@ -3980,6 +3996,48 @@ mod tests {
...
+        inst.archive();
+        inst.status = Status::Error;
+        inst.last_error = Some(TMUX_SESSION_GONE_ERROR.to_string());
+        inst.update_status_with_metadata(None);
+        assert_eq!(inst.status, Status::Idle);
+        assert_eq!(inst.last_error, None);
+    }
+
+    #[test]
+    fn test_archived_session_preserves_genuine_error() {
```
jerome-benoit: **This test is a regression-protection guard, not a red-on-pre-fix proof.**

Pre-fix execution path: archived guard absent, structured guard skipped (default backing), tmux probe at 3382-3411 fires. The setter at 3406-3409:
```rust
self.status = Status::Error;
if self.last_error.is_none() {
    self.last_error = Some(TMUX_SESSION_GONE_ERROR.to_string());
}
```
The `is_none()` check preserves `"agent crashed"`. Pre-fix outcome: `status = Error, last_error = Some("agent crashed")`, matching the assertion at 4037-4038. The test passes on both trees.

Tests 1 and 2 in this same module are genuinely red on pre-fix; the PR description's "verified red on the pre-fix tree" framing should scope to those two and call this third test what it is, a regression guard.

Either rephrase the test comment at 4029-4031 to call it a "regression guard, not a red-on-pre-fix proof", or extend it to cover a path that does fire pre-fix (e.g., `last_error = None` initially, then archive, then poll, asserting no tmux-gone string lands; that would be red pre-fix because the existing setter would write the constant).

src/session/instance.rs
```diff
@@ -3980,6 +3996,48 @@ mod tests {
...
+        // and was then archived must keep its Error status, so archive then
+        // unarchive stays status-preserving for real failures.
+        let mut inst = Instance::new("test", "/tmp/test");
+        inst.archive();
+        inst.status = Status::Error;
+        inst.last_error = Some("agent crashed".to_string());
+        inst.update_status_with_metadata(None);
+        assert_eq!(inst.status, Status::Error);
+        assert_eq!(inst.last_error.as_deref(), Some("agent crashed"));
+    }
```
jerome-benoit, line 4037: **Two test cases are missing for the symmetry contract this PR commits to.**

1. **archive→unarchive round-trip on a genuine error.** The PR's central claim is "archive/unarchive symmetry preserved" and "a genuinely errored session keeps its `Error` status across archive and unarchive". `test_archived_session_preserves_genuine_error` only confirms the heal does not touch the row while archived; it never calls `inst.unarchive()` and re-polls. Add:
   ```rust
   #[test]
   fn test_archived_unarchived_genuine_error_roundtrips() {
       let mut inst = Instance::new("test", "/tmp/test");
       inst.archive();
       inst.status = Status::Error;
       inst.last_error = Some("agent crashed".to_string());
       inst.update_status_with_metadata(None);
       inst.unarchive();
       inst.update_status_with_metadata(None);
       assert_eq!(inst.status, Status::Error);
       assert_eq!(inst.last_error.as_deref(), Some("agent crashed"));
   }
   ```
   A future refactor that resets `last_error` on `unarchive()` would silently break the symmetry; CI would not catch it. The test is robust on a CI box without tmux AND on a dev box with tmux running and no matching session: in both cases the `is_none()` guard at line 3407 preserves `"agent crashed"`.

2. **negative test for substring match.** The heal predicate is exact equality:
   ```rust
   self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)
   ```
   The 3 new tests cover (a) no error, (b) exact match, (c) unrelated message. None covers a message that *contains* the constant. A future PR loosening to `.contains(...)` or `.starts_with(...)` would heal genuine errors that embed the tmux-gone text. Add:
   ```rust
   #[test]
   fn test_archived_session_does_not_heal_substring_match() {
       let mut inst = Instance::new("test", "/tmp/test");
       inst.archive();
       inst.status = Status::Error;
       inst.last_error = Some(format!(
           "{} The agent process may have exited or been killed.",
           TMUX_SESSION_GONE_ERROR
       ));
       inst.update_status_with_metadata(None);
       assert_eq!(inst.status, Status::Error);
       assert!(inst.last_error.as_deref().unwrap().starts_with(TMUX_SESSION_GONE_ERROR));
   }
   ```
   With three tests already present, two more locking both invariants is a small marginal cost for a substantial regression-protection gain.

jerome-benoit (changes requested):
# Review: fix(session): keep archived sessions out of Error when their tmux is gone

## Verdict

REQUEST_CHANGES on the heal predicate, which cannot fire on the cross-restart footprint that test 2's comment frames as covered (see inline). Plus two cross-file follow-ups noted below since they sit outside the PR diff.

## Bug confirmation

The TUI archive flow at `src/tui/home/operations.rs:1281` calls `kill_all_tmux_sessions` before `apply_user_action(|inst| inst.archive())`, and `archive()` at `instance.rs:1142` only stamps `archived_at`. The next poll therefore lands on an archived row whose tmux is gone, hits the probe at `instance.rs:3382-3411`, and the `session.exists() == false` branch at 3406-3409 stamps `Status::Error + TMUX_SESSION_GONE_ERROR` on disk. The new archived guard at 3339-3346 short-circuits that probe path for archived rows; for an in-process row the heal block on the next tick puts the row back to `Idle`. The fix is mechanically correct for the in-process path.

## Push notification on archived heal (file outside the diff)

The poll loop at `src/server/mod.rs:2876` emits a `StatusChange` on every `*old != inst.status` transition with no `is_archived()` gate, and `handle_status_change`/`fire_due_pushes` in `src/server/push.rs` do not gate either. With this PR, the in-process Error → heal → Idle sequence on an archived row produces a second StatusChange (`Idle`) on the tick after the spurious `Error`. If `notify_on_idle` is opted in, that lands as a `Session finished: <title>` push for a session the user just archived. Cleanest gate is on the StatusChange emission at `server/mod.rs:2876`; the same gap likely exists for `is_snoozed` and is worth cross-referencing.

## TUI archive race window (file outside the diff)

`src/tui/home/operations.rs:1281` runs `kill_all_tmux_sessions()` (sync, no flock) before `apply_user_action(&id, |inst| inst.archive())` at line 1291 (saves under flock). If the 2s poller in `src/server/mod.rs` loads between those two calls, it sees `archived_at = None`, the archived guard at 3339 is skipped, the tmux probe stamps `Status::Error + TMUX_SESSION_GONE_ERROR`, and the post-poll save writes `Error` to disk. The next poll heals it via the new block, so the user-facing window is one poll interval (~2s) of a spurious dashboard flash. Predates this PR. Reordering the two calls so `archive()` runs first closes it, since the poller's archived guard then short-circuits on disk-loaded `archived_at`.

## Positive observations

- The heal predicate `self.last_error.as_deref() == Some(TMUX_SESSION_GONE_ERROR)` matches the sibling structured-view block at line 3358 byte-for-byte; the early-return guard pattern matches the existing surrounding shape.
- `pub const TMUX_SESSION_GONE_ERROR` at `instance.rs:94`, re-exported via `src/session/mod.rs:52`, is consumed cross-module by `src/tui/home/render.rs:1976`. Visibility and module placement are right.
- The second commit (`92402ed9`) reverted an earlier `archive() → status = Stopped` that would have broken `attention_session_key` and `is_recovery_candidate`. Right call.
- `is_recovery_candidate` early-rejects on `is_archived()` and `attention_tier` returns 99 for archived; the heal-to-Idle is inert in those paths.
- Test naming and `Instance::new("test", "/tmp/test")` construction match the established pattern at this scope.

## DEFERRED

- The heal predicate is now duplicated between the new archived block (3339-3346) and the existing structured block (3354-3364) with slightly different shapes; a `fn heal_tmux_gone_error_if_present(&mut self)` extract called from both is a clean follow-up.
- Archived guard ordering vs acp guard: the archived guard runs before the acp branch at 3353-3365. Functionally equivalent today, but future changes to the acp branch silently skip archived structured rows.
- Sibling tmux failure at line 3389 (`Could not reach tmux`) is not covered by the heal's exact-equality predicate. Less common than session-gone.
- The early-return at 3346 freezes `status` at whatever the poller wrote on its last tick before `archived_at` was observed. An archived `Running` session reads as `Running` indefinitely. Explicit design (status-preserving for unarchive symmetry); worth surfacing in the comment at 3333-3338 as a deliberate trade-off.

src/session/instance.rs
Seluj78: Took option (b). Dropped the inline heal predicate entirely and reduced the archived guard to `if self.is_archived() { return; }`, then added a v016 startup migration that walks every sessions.json and demotes archived rows still at `error` to `idle`. That reaches the cross-restart footprint the predicate could not (`last_error` is `#[serde(skip)]`), and matches the AGENTS.md rule of a migration over an inline compat shim. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2227/commits/26386c31a23c50b677df7d512cc3c90b792ff403.

src/session/instance.rs
Seluj78: Resolved by removal: the heal block is gone, so there is no `last_error_check` asymmetry left to mirror. The archived guard no longer mutates state at all. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2227/commits/26386c31a23c50b677df7d512cc3c90b792ff403.

src/session/instance.rs
Seluj78: Resolved by removal: no inline mutation remains in the poller. The v016 migration logs per file at boot (count of healed rows). Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2227/commits/26386c31a23c50b677df7d512cc3c90b792ff403.

src/session/instance.rs
Seluj78: Agreed. That test is now labeled a regression guard, and the red-on-pre-fix claim is scoped to the one test that is genuinely red (`test_archived_session_not_marked_error_when_tmux_gone`). PR body updated to match. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2227/commits/26386c31a23c50b677df7d512cc3c90b792ff403.

src/session/instance.rs
Seluj78: Added `test_archived_unarchived_genuine_error_roundtrips`. The substring-negative test is moot now: the heal predicate, and all string matching, is gone; v016 keys off `status` + `archived_at` only. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2227/commits/26386c31a23c50b677df7d512cc3c90b792ff403.

src/migrations/v016_clear_archived_tmux_gone_error.rs
```diff
@@ -0,0 +1,140 @@
...
+        }
+    };
+
+    let mut healed = 0usize;
+    if let Some(array) = value.as_array_mut() {
+        for instance in array.iter_mut() {
+            if let Some(obj) = instance.as_object_mut() {
+                let archived = obj.get("archived_at").is_some_and(|v| !v.is_null());
+                let errored = obj.get("status").and_then(|v| v.as_str()) == Some("error");
+                if archived && errored {
```
jerome-benoit, line 68: A genuinely errored session that the user then archives lands on disk as `archived_at:<ts>, status:"error"`: `test_archived_session_preserves_genuine_error` (instance.rs:4007-4020) confirms `Error` survives `archive()` in-memory, and the storage layer serializes that pair. v016 cannot distinguish that row from the spurious tmux-gone footprint and demotes it to `Idle`.

Mitigation is real: `last_error` is `#[serde(skip)]`, so after any reload the message is already gone; the migration trades a residual `Error` flag for clean `Idle` and accepts the overlap. Worth a one-line acknowledgment in the doc-comment so a future reader does not assume strict preservation:

> An archived row at `error` after reload has already lost its `last_error` (`#[serde(skip)]`), so demoting to `idle` here trades a residual error flag for a clean state across both the spurious-bug footprint and any genuine-error-then-archive that happened to overlap with it.

src/migrations/v016_clear_archived_tmux_gone_error.rs
```diff
@@ -0,0 +1,140 @@
...
+                        serde_json::Value::String("idle".to_string()),
+                    );
+                    healed += 1;
+                }
+            }
+        }
+    }
+
+    if healed > 0 {
+        fs::write(path, serde_json::to_string_pretty(&value)?)?;
```
jerome-benoit, line 80: `fs::write(path, ...)` is not atomic, and v016 takes no flock around the read-modify-write. `Storage::update` (`src/session/storage.rs:431-476`) uses `atomic_write` (temp + `sync_data` + rename + dir fsync at `storage.rs:98-118`) plus a per-profile flock at `storage.rs:445`. Two concrete consequences:

1. A disk-full or crash mid-write here can leave a torn JSON document; subsequent `Storage::load` (`storage.rs:315`) surfaces top-level corruption as `Err` and aborts boot (`test_load_top_level_corruption_still_errors` at `storage.rs:743`).
2. `get_current_version` (`migrations/mod.rs:172`) reads `.schema_version` with no lock. Two `aoe` processes started near-simultaneously on a v15 install (e.g. `aoe` TUI + `aoe serve`) can both observe v15, both enter v016, and last-writer-wins clobbers a peer's concurrent edits.

Probability is bounded (one schema-15→16 upgrade window per install) but the failure modes are real and unrecoverable without manual repair.

Fix: reuse `crate::session::storage::atomic_write` (already `pub(crate)`) for the single write, and wrap the read-modify-write under the same per-profile flock the rest of the codebase uses (mirror `acquire_storage_flock`).

src/migrations/v016_clear_archived_tmux_gone_error.rs
```diff
@@ -0,0 +1,140 @@
...
+        info!(
+            "v016: cleared spurious archived Error on {healed} session(s) in {} (#2206)",
+            path.display()
+        );
+    }
+    Ok(())
+}
+
+#[cfg(test)]
```
jerome-benoit, line 89: Three short tests would lock invariants the migration assumes or that AGENTS.md mandates:

1. **Idempotency** (AGENTS.md `Data Migrations`: *"Migrations must be idempotent"*). Run `clear_archived_error` twice on the same file; assert `healed = 0` on the second pass and the byte content matches the first run.

2. **Round-trip preservation of non-status fields.** A row carrying `id`, `title`, `archived_at`, plus an unknown forward-compat key (e.g. `"forward_compat": 7`) survives the heal with all those fields intact. Today's `clears_only_archived_error_rows` only asserts the `status` field changed.

3. **Empty (0-byte) `sessions.json`.** Currently falls into the `serde_json::from_str` `Err` branch and is skipped silently; assert this explicitly so a future change to the parse path doesn't quietly start aborting boot on an empty file.

jerome-benoit (approved):
# Follow-up review on commit 26386c31

The 5 inline findings from my prior review are addressed; the v016 migration approach is the right architectural choice (reaches the cross-restart footprint the inline heal could not). APPROVE.

Three advisory items inline.

PR #2234, opened 2026-06-17
fix(web): tolerate finger jitter on sidebar long-press so the session menu opens on Android

### LINKED ISSUE #2232: Web sidebar: Android long-press opens native link menu instead of the session context menu (zero-slop touchmove cancels the timer)

## Problem

On the mobile PWA (Android, Pixel 8 Pro), long-pressing a session row in the sidebar is supposed to open the app's custom context menu (rename / delete / etc.). Instead, unless you press very precisely and hold perfectly still, Android shows the browser's native link menu ("Open in new tab", "Copy link address", and so on). You can sometimes get the custom menu, but only by being unnaturally careful, which is not how anyone long-presses on a phone.

This is the Android sibling of #1451 (iOS link-preview callout on long-press), which was fixed in #1613 by adding `[-webkit-touch-callout:none]`. That CSS is iOS-only and does nothing on Android, so the same class of bug is still live there.

## What's there today

The session row is an anchor (`web/src/components/WorkspaceSidebar.tsx:870`):

```jsx
<a
  href={sessionPath}
  draggable={false}
  onContextMenu={handleContextMenu}
  onTouchStart={handleTouchStart}
  onTouchEnd={handleTouchEnd}
  onTouchMove={clearLongPress}
  onTouchCancel={clearLongPress}
  className={`... select-none [-webkit-touch-callout:none] ...`}
```

A custom long-press timer drives the menu (`WorkspaceSidebar.tsx:760`):

```tsx
const handleTouchStart = (e: React.TouchEvent) => {
  clearLongPress();
  longPressFired.current = false;
  if (!sessionId || isDeleting) return;
  const touch = e.touches[0];
  if (!touch) return;
  const tx = touch.clientX;
  const ty = touch.clientY;
  longPressTimer.current = setTimeout(() => {
    longPressFired.current = true;
    touchOpenedAt.current = Date.now();
    closeOtherContextMenus();
    setContextMenu({ x: tx, y: ty });
  }, 500);
};
```

And `touchmove` cancels that timer with zero tolerance (`WorkspaceSidebar.tsx:899` -> `:753`):

```tsx
onTouchMove={clearLongPress}
// ...
const clearLongPress = () => {
  if (longPressTimer.current) {
    clearTimeout(longPressTimer.current);
    longPressTimer.current = null;
  }
};
```

The iOS-only suppressant is on the className (`WorkspaceSidebar.tsx:902`): `[-webkit-touch-callout:none]`. There is no movement threshold anywhere in the touch path.

## Root cause

Two layers, the first is the real one:

1. **Zero-slop `touchmove` cancellation.** `onTouchMove={clearLongPress}` (`WorkspaceSidebar.tsx:899`) clears the 500ms custom timer on *any* movement, even one pixel. A normal finger long-press always jitters a few pixels, so the custom timer is cancelled before it fires and the custom menu never opens. "Being precise" means holding perfectly still so no `touchmove` arrives, which is the only path where the timer survives. Meanwhile Android's native long-press detection has its own movement slop and tolerates that same jitter, so the native link menu wins.

2. **iOS-only native suppression.** The only thing suppressing the native menu on the row is `[-webkit-touch-callout:none]` (`WorkspaceSidebar.tsx:902`), which Android Chrome ignores. The `onContextMenu` preventDefault path (`WorkspaceSidebar.tsx:746`) is the desktop right-click route and is not a reliable substitute for the Android long-press gesture on an `<a href>`.

## Proposal

**A (minimum viable fix):** Give `touchmove` a movement threshold instead of cancel-on-any-move. Record the start coords in `handleTouchStart`, and in a new `handleTouchMove` only call `clearLongPress()` when the finger has moved past a slop radius (~10px). Below that, treat it as a stationary hold and let the 500ms timer fire. This is the same touch-slop discipline AGENTS.md already calls out for the Playwright touch specs.

Rough shape, touching only `WorkspaceSidebar.tsx`:

```tsx
const touchStart = useRef<{ x: number; y: number } | null>(null);
const SLOP = 10;

// in handleTouchStart, after reading tx/ty:
touchStart.current = { x: tx, y: ty };

const handleTouchMove = (e: React.TouchEvent) => {
  const t = e.touches[0];
  if (!t || !touchStart.current) return;
  const dx = t.clientX - touchStart.current.x;
  const dy = t.clientY - touchStart.current.y;
  if (Math.hypot(dx, dy) > SLOP) clearLongPress();
};
// onTouchMove={handleTouchMove}
```

**B (belt and suspenders for the native menu):** Once the long-press timer fires, the native Android menu can still race in. The custom timer firing should also suppress the native default. Calling `e.preventDefault()` in a non-passive `touchstart`/`touchmove` is the lever; or render the row as a non-anchor element with programmatic navigation so the OS never attaches link-menu semantics to it. (B) is only worth it if (A) alone does not fully suppress the native menu on a real Pixel; verify (A) first.

## Why now

- It is a daily-driver papercut on the primary mobile surface; rename/delete is unreachable without fiddly precision.
- It is the unfinished half of #1451 / #1613, which only covered iOS. Android was left exposed.
- The fix is contained to one component and one handler.

- As a mobile user, when I long-press a session row with normal finger jitter (under ~10px of movement), then the app's rename/delete context menu opens and the browser's native link menu does not.
  - Suggested test layer: mocked Playwright (mobile viewport, synthetic `TouchEvent` per the AGENTS.md touch recipe; assert the custom menu appears after the long-press hold with small jitter).
- As a mobile user, when I press a row and drag more than ~10px (scroll intent), then no context menu opens and the list scrolls normally.
  - Suggested test layer: mocked Playwright (same harness; assert no custom menu and the long-press timer is cancelled past the slop threshold).

## Open questions

- Does fix (A) alone fully suppress the native Android menu on a real Pixel 8 Pro, or is (B) also required? Needs a device check.
- Slop radius: 10px is the common default; confirm it feels right against the row height and does not swallow intended scrolls.

## Out of scope

- iOS behavior (already handled by #1613).
- Desktop right-click context menu (`handleContextMenu`), which works today.
- Reworking the row away from an `<a>` element, unless (B) proves necessary.

## Cross-references

- #1451 — iOS sibling of this bug (long-press callout); this is the Android case it did not cover.
- #1613 — the fix for #1451; added the iOS-only `[-webkit-touch-callout:none]` that does nothing on Android.

Seluj78:

## Description

On the mobile PWA (Android), long-pressing a sidebar session row was supposed to open the app's rename/delete context menu, but unless you pressed very precisely and held perfectly still you got the browser's native link menu (open in new tab, copy link) instead.

Root cause: the session row is an `<a href>`, and `onTouchMove={clearLongPress}` cancelled the 500ms long-press timer on any pixel of movement. A normal finger long-press always jitters a few pixels, so the timer was cancelled before it fired and the custom menu never opened, while Android's slop-tolerant native long-press won the gesture. The only native suppressant on the row was the iOS-only `[-webkit-touch-callout:none]` from #1613, which does nothing on Android. This is the Android sibling of #1451.

The fix introduces an `exceedsTouchSlop` helper and only cancels the pending long-press once movement passes an 8px slop, the same tolerance the dnd-kit `TouchSensor` already uses for reorder activation. A jittery stationary hold now keeps the timer armed and opens the menu; a deliberate drag (scroll or reorder) still cancels it.

The regression test is a Vitest on `exceedsTouchSlop` rather than a Playwright spec: chromium synthesizes a `contextmenu` event on a small-movement touch long-press (which opens the menu through a different code path) but not on a drag, so a Playwright menu-visibility assertion passes identically on the buggy and fixed builds and cannot discriminate the slop logic. The Vitest fails on the old cancel-on-any-move behavior (slop 0) and passes on the fix. Confirming the native-menu suppression on a real device stays a manual check, same as #1613.

Closes #2232

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] On an Android phone, open the PWA, long-press a session row with a normal (slightly jittery) hold; confirm the app's rename/delete menu opens and the browser's native link menu does not.
- [ ] Press a row and drag more than ~8px (scroll intent); confirm no context menu opens and the list scrolls.
- [ ] Desktop: right-click a session row; confirm the context menu still opens as before.
- [ ] `cd web && npx vitest run src/lib/__tests__/longPress.test.ts` passes.

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: a Playwright menu-visibility assertion cannot discriminate this fix (chromium's touch-to-contextmenu synthesis opens the menu on small movement regardless of the slop logic), so the regression test is a deterministic Vitest on `exceedsTouchSlop` and `web/tests/coverage-matrix.json` is updated with the `sidebar.longpress-slop` entry pointing at it.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (slop value, Vitest over a confounded Playwright spec) and reviewed each commit.

**NOTE:**

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2234"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784267954&installation_id=139138837&pr_number=2234&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2234&signature=f9e8a2a41990eac053347e6c899b3eb48c176d34a1c0fb15579a4fb34de968ad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Overview
This PR fixes an Android PWA bug where long-pressing a sidebar session row opens the browser's native link menu instead of the app's custom rename/delete context menu. The issue occurs because the app's custom long-press timer is cancelled on any touch movement, while Android's native gesture detector tolerates small finger jitter, causing it to win the gesture race.

## What Changed

**Added:**
- New `longPress.ts` utility module with a movement detection helper and 8-pixel threshold constant
- Vitest test suite validating the movement detection logic
- Touch movement tracking in the sidebar session row component
- Coverage matrix entry for the new functionality

**Modified:**
- `WorkspaceSidebar.tsx` updated to track initial touch coordinates and only cancel the long-press timer when finger movement exceeds the 8-pixel threshold

## Benefits
- Users can now reliably open the app's custom context menu on Android with normal long-press gestures that include natural finger jitter
- The native Android link menu is suppressed during long-press interactions
- Intentional drags (scrolls or reorders) still properly cancel the long-press timer
- Desktop right-click and iOS behaviors remain unaffected
- The 8-pixel threshold matches industry-standard gesture slop tolerances (dnd-kit's TouchSensor)

## Technical Details
The fix introduces an `exceedsTouchSlop` helper function that:
- Calculates Euclidean distance between initial touch position and current position
- Cancels the custom long-press timer only when movement exceeds 8 pixels
- Allows normal stationary holds with minor jitter (< 8px) to fire the custom context menu
- Still cancels the timer for deliberate drags beyond the threshold

The 8-pixel slop value is configurable and matches the tolerance used by the dnd-kit library for reorder gesture activation, providing consistent gesture semantics across the application.

Testing uses Vitest unit tests rather than Playwright because Chromium's touch simulation synthesizes a `contextmenu` event regardless of slop logic, making Playwright unable to reliably discriminate between buggy and fixed builds. The unit tests deterministically validate the slop behavior with structured test cases. Manual verification on a real Android device is required to confirm native menu suppression.

**Addresses:** Issue `#2232` and `#1451`

### DIFF
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -52,6 +52,7 @@ import { safeGetItem, safeSetItem } from "../lib/safeStorage";
 import { REPO_COLOR_OPTIONS, type RepoAppearanceUpdate, type RepoColor } from "../lib/repoAppearance";
 import { STATUS_DOT_CLASS, getStatusTextClass, isSessionActive } from "../lib/session";
 import { useIdleDecayWindowMs } from "../lib/idleDecay";
+import { exceedsTouchSlop } from "../lib/longPress";
 import { TOUR_ANCHORS, tourAnchor } from "../lib/tourSteps";
 import { renameSession, setSessionNotifications, setWorktreeName, updateSessionGroup } from "../lib/api";
 import { useServerDown, OFFLINE_TITLE } from "../lib/connectionState";
@@ -705,6 +706,7 @@ export const SessionRow = memo(function SessionRow({
   const longPressTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
   const longPressFired = useRef(false);
   const touchOpenedAt = useRef(0);
+  const touchStart = useRef<{ x: number; y: number } | null>(null);
   const menuRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
@@ -766,12 +768,24 @@ export const SessionRow = memo(function SessionRow({
     if (!touch) return;
     const tx = touch.clientX;
     const ty = touch.clientY;
+    touchStart.current = { x: tx, y: ty };
     longPressTimer.current = setTimeout(() => {
       longPressFired.current = true;
       touchOpenedAt.current = Date.now();
       closeOtherContextMenus();
       setContextMenu({ x: tx, y: ty });
     }, 500);
+  };
+
+  // Cancel the pending long-press only once the finger moves past the slop, so
+  // a normal jittery hold still opens the menu while a deliberate drag
+  // (scroll/reorder) cancels it. See exceedsTouchSlop (#2232).
+  const handleTouchMove = (e: React.TouchEvent) => {
+    const touch = e.touches[0];
+    if (!touch || !touchStart.current) return;
+    if (exceedsTouchSlop(touchStart.current, { x: touch.clientX, y: touch.clientY })) {
+      clearLongPress();
+    }
   };
 
   const handleTouchEnd = (e: React.TouchEvent) => {
@@ -897,7 +911,7 @@ export const SessionRow = memo(function SessionRow({
         onContextMenu={handleContextMenu}
         onTouchStart={handleTouchStart}
         onTouchEnd={handleTouchEnd}
-        onTouchMove={clearLongPress}
+        onTouchMove={handleTouchMove}
         onTouchCancel={clearLongPress}
         data-selected={isSelected || undefined}
         className={`block w-full text-left py-2 cursor-pointer select-none [-webkit-touch-callout:none] transition-colors duration-75 ${
--- a/web/src/lib/__tests__/longPress.test.ts
+++ b/web/src/lib/__tests__/longPress.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from "vitest";
+import { exceedsTouchSlop, LONG_PRESS_SLOP_PX } from "../longPress";
+
+describe("exceedsTouchSlop (#2232)", () => {
+  const start = { x: 100, y: 100 };
+
+  it("treats a stationary touch as within slop", () => {
+    expect(exceedsTouchSlop(start, { x: 100, y: 100 })).toBe(false);
+  });
+
+  it("tolerates normal finger jitter under the slop so the long-press survives", () => {
+    // A real hold wobbles a few px; the old cancel-on-any-move killed the menu.
+    expect(exceedsTouchSlop(start, { x: 103, y: 98 })).toBe(false);
+    expect(exceedsTouchSlop(start, { x: 95, y: 104 })).toBe(false);
+  });
+
+  it("does not cancel exactly at the slop boundary", () => {
+    expect(exceedsTouchSlop(start, { x: 100 + LONG_PRESS_SLOP_PX, y: 100 })).toBe(false);
+  });
+
+  it("cancels once a deliberate drag passes the slop", () => {
+    expect(exceedsTouchSlop(start, { x: 100 + LONG_PRESS_SLOP_PX + 1, y: 100 })).toBe(true);
+    expect(exceedsTouchSlop(start, { x: 140, y: 100 })).toBe(true);
+    expect(exceedsTouchSlop(start, { x: 100, y: 60 })).toBe(true);
+  });
+
+  it("uses 8px to match the dnd-kit TouchSensor reorder tolerance", () => {
+    expect(LONG_PRESS_SLOP_PX).toBe(8);
+  });
+});
--- a/web/src/lib/longPress.ts
+++ b/web/src/lib/longPress.ts
@@ -1,0 +1,19 @@
+// Long-press gesture tolerance for sidebar session rows.
+//
+// A finger never holds perfectly still during a long-press. Cancelling the
+// pending long-press on any movement (the old behavior) meant the rename/delete
+// menu only opened when you pressed unnaturally precisely; otherwise Android's
+// slop-tolerant native link menu won the gesture instead (#2232). We only treat
+// the press as cancelled once movement exceeds this slop, the same 8px the
+// dnd-kit TouchSensor uses for its reorder activation tolerance, so a normal
+// jittery hold still arms the menu while a deliberate drag (scroll/reorder)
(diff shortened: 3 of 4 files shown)


njbrake:
@Seluj78 merge conflicts but then good to go!

njbrake (approved):

Issue #2260, opened 2026-06-17
Cockpit: renaming a tied-worktree structured-view session moves the dir out from under the live worker, crash-loops then parks with a misleading install-adapter banner


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Renaming a structured-view (ACP) session from the web dashboard, when "Tie Worktree Directory to Session Name" is on (the default), moves the worktree directory out from under the still-live ACP worker. The worker then crash-loops trying to spawn into the old, now-gone directory until the respawn budget is exhausted, and the session is parked with a red banner:

```
Structured view agent failed to start

Structured view worker failed to stay up after 5 restart attempts in 60s; auto-respawn paused. Retry from the dashboard once the underlying issue is fixed.

Run aoe acp doctor --fix from a terminal, or install the adapter manually:

npm install -g @agentclientprotocol/claude-agent-acp@latest
```

The banner is doubly wrong: the adapter is installed and fine, and the real cause is a moved working directory. After a while the worker does respawn at the new path, but the agent comes back without its prior conversation context.

The rename also lands half-applied: the title and the directory leaf move to the new name, but the git branch keeps the old auto-generated name, so the sidebar still shows the old `branch <- main` label. The branch staying behind is by-design (#1927 made branch rename opt-in), but combined with the crash it reads as "the rename half-worked".

## Evidence from the affected session

Session `d5f384bf604544e1`, captured live today. The rename moved the worktree `Burmese` to `new-session-tooltip-z-index-being-incorrect-on-mobile`, then the worker respawn failed on the old path and the session was parked (release `~/.agent-of-empires/debug.log`):

```
move_worktree: invoking `git worktree move` from=.../agent-of-empires-worktrees/Burmese to=.../agent-of-empires-worktrees/new-session-tooltip-z-index-being-incorrect-on-mobile
WARN acp.supervisor: respawn failed: project path no longer exists: .../agent-of-empires-worktrees/Burmese session=d5f384bf604544e1
WARN acp.supervisor: structured-view worker respawn budget exhausted; parking session session=d5f384bf604544e1 max_respawns=5 window_secs=60
```

Persisted instance state after the crash (`~/.agent-of-empires/profiles/default/sessions.json`):

```
id= d5f384bf604544e1
  title=        'new session tooltip z index being incorrect on mobile'
[...]

PR #2266, opened 2026-06-17
fix(serve): make structured-view smart-rename observable and robust

Seluj78:

## Description

Makes the new structured-view smart-rename feature observable and robust. (The tied-worktree-rename crash this branch originally also targeted is already fixed on main by #2260, which quiesces the worker before the move and resumes the conversation at the new path; that work has been dropped from here to avoid conflicting with it.)

Two commits:

1. **Logging.** The smart-rename path logs every eligibility skip (`debug`) and each rename outcome (`info`) under `target: "smart_rename"`, but that target was not in `DEFAULT_TARGET_ROOTS`. The default filter expands a level into one directive per root with no global default, so the target was dropped at every level: zero output, feature undiagnosable. Added the root and a row to the logging taxonomy doc.

2. **One-shot robustness.** The session was marked attempted (one try per daemon lifetime) before the one-shot ran, so a spawn error or timeout left it permanently un-named until a restart. Mark attempted only after the one-shot returns output, so a transient slow start retries on the next prompt while a usable-but-rejected answer still does not respawn the agent every prompt. Also raised the one-shot timeout from 12s to 45s, since a cold agent start (`claude -p` and friends) routinely exceeds 12s and times out into the bug above.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Run `aoe log-level smart_rename=debug`, send a first prompt to a fresh civ-named structured-view session, and confirm a `smart_rename` line appears (a skip reason, or `renamed session from first message`). Before this PR there was no output at all, at any level.
- [ ] Send a first prompt that makes the agent's cold start exceed 12s and confirm the session still auto-renames (no premature 12s timeout) and is not left permanently un-named.

## Checklist

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (logging filter + one-shot lifecycle on the backend).

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered by Rust unit tests instead. `smart_rename_target_is_captured_by_default_filter` asserts the logging root is present; `run_oneshot_returns_none_on_spawn_failure` pins the failure path the retry guard keys off. The attempted-after-result reorder is a localized control-flow change in the serve module.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls and reviewed each commit.

## Overview

This PR fixes a crash that occurs when renaming sessions with tied worktree directories in a structured-view worker context. The crash was caused by the worker attempting to resume a session using a stored ID that referenced an old, moved directory path.

## What Changed

**Logging improvements:**
- Added a new `smart_rename` logging target to enable diagnostic output for the auto-rename feature
- Updated logging documentation to include the new target

**One-shot robustness:**
- Increased the one-shot timeout from 12 seconds to 45 seconds to accommodate cold agent starts
- Changed when sessions are marked as "attempted": now happens after successful output instead of before execution, allowing transient failures to retry on subsequent prompts

**Core crash fix:**
- Modified the rename logic to clear the stored session ID when a worktree path actually moves, forcing a fresh session at the new location
- The live worker is shut down to prevent its cached working directory from interfering
- Branch-only renames (where the path stays the same) preserve the resumable session

## Key Trade-off

Renaming a tied session resets the agent's conversation since the prior session was rooted at the old directory and cannot be resumed. This design choice prioritizes functionality (users can rename sessions) over conversation continuity, and is documented in the worktree guides.

## Benefits

- **Prevents crashes**: Eliminates the `Internal error: Path "<old>" does not exist` crash
- **Improves reliability**: Transient timeouts no longer permanently disable session naming
- **Better diagnostics**: New logging enables troubleshooting of rename operations
- **Handles cold starts**: Longer timeout accommodates slower agent initialization

### DIFF
--- a/docs/development/logging.md
+++ b/docs/development/logging.md
@@ -29,6 +29,7 @@ Top-level roots:
 | `hooks` | Agent hook install/uninstall, status-file lifecycle, hook command + watcher failures. |
 | `sound` | Notification sound download/install and per-event playback. |
 | `telemetry` | Opt-in usage telemetry (all `debug`); consent route emits under `http.api.telemetry`. |
+| `smart_rename` | First-message auto-rename of structured-view sessions: eligibility skips (`debug`) and the rename outcome (`info`). |
 | `log.runtime` | Filter swaps (REST + runner file-watch). A swap matching the active directive is a silent no-op, so a watcher write cannot re-trigger itself. |
 
 ## Levels
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -147,6 +147,7 @@ pub const DEFAULT_TARGET_ROOTS: &[&str] = &[
     "hooks",
     "sound",
     "telemetry",
+    "smart_rename",
 ];
 
 /// Sub-targets users can tune individually from the settings UI.
@@ -1096,6 +1097,19 @@ mod tests {
                 "missing {root} in {s}"
             );
         }
+    }
+
+    #[test]
+    fn smart_rename_target_is_captured_by_default_filter() {
+        // The expanded filter has no global default directive, so a target that
+        // is not a known root is dropped at every level. smart_rename emits under
+        // `target: "smart_rename"`; without a root entry its skip/success lines
+        // are invisible and the feature cannot be diagnosed.
+        let s = LogConfig::filter_for_level(LogLevel::Debug);
+        assert!(
+            s.contains("smart_rename=debug"),
+            "smart_rename root missing from filter: {s}"
+        );
     }
 
     #[test]
--- a/src/session/smart_rename.rs
+++ b/src/session/smart_rename.rs
@@ -247,7 +247,7 @@ mod serve {
     use std::sync::{Arc, Mutex};
     use std::time::Duration;
 
-    const ONESHOT_TIMEOUT: Duration = Duration::from_secs(12);
+    const ONESHOT_TIMEOUT: Duration = Duration::from_secs(45);
 
     /// Marks a session as having an in-flight one-shot rename so a burst of
     /// rapid first prompts cannot spawn concurrent title generators. Removed on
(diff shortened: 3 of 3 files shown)


docs/guides/worktrees.md
```diff
@@ -86,6 +86,7 @@ When tied:
...
 - The git branch is never swept in by a title rename by default. To rename it too, check "Also rename git branch" in the TUI rename dialog, pass `--rename-branch` to `aoe session rename`, or send `rename_branch: true` to the PATCH. It stays opt-in because a branch may carry an upstream or an open PR; the TUI toggle warns when the branch tracks a remote, since the remote branch (and any open PR) won't follow the local rename.
 - The session must be stopped first. Moving the directory of a running worktree is unsafe, so a tied rename of a running session is refused with a clear message. Stop the session, or disable the setting, to relabel it freely.
+- For a structured view (ACP) session, the worktree move resets the agent conversation: the agent's prior session was rooted at the old directory, so resuming it would point the agent at a path that no longer exists. After the rename the worker is torn down and the next prompt starts a fresh agent session at the new directory.
```
njbrake: This seems like then we should disable this and not allow a user to rename in this scenario?

njbrake (changes requested):
It seems like maybe we should prevent the rename instead of allowing it and wiping the session?

jerome-benoit:
> It seems like maybe we should prevent the rename instead of allowing it and wiping the session?

What is the exact root cause of the crash? Can we have a complete stack trace?

docs/guides/worktrees.md
Seluj78: Note: Claude handled the implementation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Good call, agreed on not wiping. The crash is now handled on main by #2260, which quiesces the structured-view worker before the worktree move and resumes the conversation at the new path (no reset). I dropped my conflicting reset-on-move approach; this PR is now scoped to smart-rename observability and robustness only.

Seluj78:
Note: Claude handled the implementation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

On the rename-vs-wipe point: agreed. #2260 (already on main) fixes the crash by quiescing the worker before the worktree move and resuming the conversation at the new path, so nothing is wiped. I dropped the conflicting reset-on-move commit; this PR is now just the smart-rename logging root and one-shot timeout/retry fixes. Re-requesting review on that reduced scope.

Seluj78:
Note: Claude handled the implementation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

> What is the exact root cause of the crash? Can we have a complete stack trace?

Root cause: the structured-view ACP worker was not quiesced before `git worktree move`. The move pulled the worktree directory out from under the live worker, which then respawned at its stale baked-in cwd. The agent errored `Internal error: Path ".../agent-of-empires-worktrees/Burmese" does not exist`, burned its 3-respawn budget, and the reconciler parked the session. Fixed on main by #2260 (shutdown-before-move, then resume at the new path). This PR no longer touches that code path.

jerome-benoit:
> Root cause: the structured-view ACP worker was not quiesced before `git worktree move`. The move pulled the worktree directory out from under the live worker, which then respawned at its stale baked-in cwd. The agent errored `Internal error: Path ".../agent-of-empires-worktrees/Burmese" does not exist`, burned its 3-respawn budget, and the reconciler parked the session. Fixed on main by #2260 (shutdown-before-move, then resume at the new path). This PR no longer touches that code path.

So the PR is now rescoped properly.

njbrake (approved):

PR #2294, opened 2026-06-19
feat(tui): collapsible sidebar + clickable footer toolbar (and fix config-save screen flicker)

njbrake:
## Description

Two related changes for the clickable TUI.

**Feature: collapsible sidebar + clickable footer toolbar**

- A collapse button (`«`) sits on the session list's top-right border. Clicking it collapses the list to a narrow click-to-expand strip (`»` plus the session count); the preview takes the freed width. The state persists via `app_state.home_sidebar_collapsed`, and the existing live-send `leader b` toggle now drives the same persisted state.
- The footer hint bar is now a real toolbar: each shortcut renders a hit rect paired with the key it represents, and a click synthesizes that key through the full key handler, so a click behaves exactly like pressing the shortcut (strict-hotkey mappings included). The hovered button highlights as an inverted accent chip; hover tracking was wired through `handle_hover` centrally so future clickable surfaces can reuse it.

**Fix: config saves no longer force a full-screen clear (flicker)**

While testing the collapse button I found it flickered. Capturing the TUI's raw output showed a bare `\033[2J` full-screen clear plus a full repaint on every toggle. Root cause: `App::set_theme` set `needs_redraw` unconditionally, and the config-file watcher re-dispatches the theme on every `config.toml` save. So any config-writing action forced a `clear_terminal()` flash. This was pre-existing: list resize (`<`/`>`), the info toggle (`i`), and settings saves all flickered the same way; the new collapse persistence just inherited it.

`set_theme` is now idempotent: it tracks the applied `(theme name, palette mode)` and no-ops (no `needs_redraw`) when nothing changed. After the fix, collapse/expand and the other config-saving actions emit zero `\033[2J` clears.

## PR Type

- [x] New Feature
- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated tests (unit/integration under `src/tui/home/tests.rs` and `src/tui/app.rs`)
  - `footer_toolbar`: buttons map clicks to shortcuts, strict-mode chords, hover tracking, no buttons during live-send.
  - `sidebar_collapse_button_and_strip_toggle`: collapse-button and strip click toggles + hit rects.
  - `theme_apply_needed_compares_name_and_palette_mode`: the idempotency guard that prevents the flicker.

  Note: the flicker fix was also verified directly by capturing the TUI's raw terminal output (`tmux pipe-pane`) and asserting zero `\033[2J` clears on collapse/expand/`<`/`>`/`i`; that byte-level check isn't practical to land as an automated test, so the regression is guarded by the `theme_apply_needed` unit test instead.

## How tested

- `cargo fmt`, `cargo clippy --all-targets`, `cargo test --lib tui::` (1398 passing) + `tui::app` (40 passing).
- Manual: ran the real `aoe` binary against a live session in tmux, injected collapse/expand mouse clicks, and confirmed via raw-output capture that the full-screen clear is gone.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Used for implementation, root-cause investigation of the flicker (byte-level terminal-output capture), and tests, in back-and-forth with @njbrake.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2294"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784488536&installation_id=139138837&pr_number=2294&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2294&signature=60826697e28a1a749274fe559d6ee915f07001f840e0305399eded40d01d876d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Overview

This PR improves the TUI with a persistent, collapsible session sidebar, a mouse-clickable footer shortcut toolbar, and a fix for distracting full-screen flicker when settings/theme/config change.

## What changed

**Collapsible sidebar (persistent):**  
A collapse button (`«`) on the session list now collapses the sidebar into a narrow strip (`»` + session count), freeing space for the preview pane. The collapsed/expanded state is saved and restored across restarts. The existing `leader b` control now drives the same persisted sidebar state.  
Visual polish: it removes a double-border seam so the preview’s left border is the single shared separator.

**Clickable footer toolbar:**  
Footer shortcut hints are now real clickable buttons. Clicking a button triggers the same behavior as pressing the shortcut (including strict hotkey/chord behavior), and hovered buttons highlight consistently.  
Usability improvements: click + hover geometry is recalculated per render to avoid stale hit targets, footer click rectangles match what the UI actually draws (using rendered cell width), and footer/sidebar clicks are isolated so they don’t take effect while modal-style overlays are open.

**Flicker fix (theme/config saves):**  
Theme application is now idempotent: config/theme re-dispatches no longer trigger full-screen clears when the effective theme hasn’t changed. This removes flicker during sidebar collapse/expand, list resizing, info toggle, and settings saves.

## Benefits

- More usable horizontal space via sidebar collapse without losing interaction state.
- Better mouse usability: footer shortcuts behave like real key presses, including strict hotkey mappings.
- Smoother visuals: eliminates distracting full-screen flicker during common UI/config actions.
- More reliable clicking: hit regions are kept accurate and cleared when takeover/modal views are active.

## Technical notes (for review)

- Added persisted UI preference: `AppStateConfig.home_sidebar_collapsed: Option<bool>`.
- `App::set_theme` made idempotent by tracking applied `(theme_name, palette_mode)` and skipping redraw work when unchanged.
- Home view hit-testing/rendering updates:
  - Rebuild frame-local click geometry on render to prevent stale rectangles (especially for takeover views).
  - Footer toolbar hover and click routing added; click targets synthesized into normal key handling.
  - Footer hit-rectangle width uses `Span::width()` so clicks align with rendered button cells.
- Sidebar/footer click isolation added so shortcuts don’t get synthesized behind open modal overlays.

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -132,6 +132,16 @@ pub struct App {
     home: HomeView,
     should_quit: bool,
     theme: Theme,
+    /// Identity of the currently applied `theme` (global theme name +
+    /// palette-downsample mode). `set_theme` compares against this so a
+    /// re-apply with the same identity is a no-op. The config-file watcher
+    /// re-dispatches the theme on EVERY `config.toml` save (it can't tell
+    /// what changed), and a needless `set_theme` there sets `needs_redraw`,
+    /// forcing a full-screen `clear_terminal` that flickers. Guarding here
+    /// keeps any config save (collapse persistence, list resize, `i`,
+    /// settings) from clearing the screen when the theme is unchanged.
+    theme_name: String,
+    theme_palette_mode: bool,
     needs_redraw: bool,
     update_info: Option<UpdateInfo>,
     update_rx: Option<tokio::sync::oneshot::Receiver<anyhow::Result<UpdateInfo>>>,
@@ -213,6 +223,15 @@ pub fn check_version_change() -> Result<Option<String>> {
     } else {
         Ok(None)
     }
+}
+
+/// Whether applying `next` `(theme name, palette-downsample mode)` would change
+/// the active theme `current`. Pulled out of `App::set_theme` so the
+/// idempotency guard (which keeps a config-file-watcher theme re-dispatch from
+/// forcing a flickering full-screen clear on every `config.toml` save) is
+/// unit-testable without constructing a full `App`.
+fn theme_apply_needed(current: (&str, bool), next: (&str, bool)) -> bool {
+    current != next
 }
 
 impl App {
@@ -349,6 +368,8 @@ impl App {
             home,
             should_quit: false,
             theme,
+            theme_name,
+            theme_palette_mode: palette_mode,
             needs_redraw: true,
             update_info: None,
             update_rx: None,
@@ -531,7 +552,20 @@ impl App {
         // global config: theme (and its color_mode) is a global preference,
         // not profile-merged.
         let palette_mode = crate::session::config::resolve_theme_palette_mode();
+        // No-op when the theme is already applied. The config watcher
+        // re-dispatches the theme on every `config.toml` save, so without
+        // this a list-resize / `i` / collapse-persistence / settings save
+        // would force a full-screen `clear_terminal` and flicker even though
+        // nothing visual changed.
+        if !theme_apply_needed(
+            (&self.theme_name, self.theme_palette_mode),
+            (name, palette_mode),
+        ) {
+            return;
+        }
         self.theme = crate::tui::styles::load_theme_with_mode(name, palette_mode);
+        self.theme_name = name.to_string();
+        self.theme_palette_mode = palette_mode;
         self.needs_redraw = true;
     }
 
@@ -823,6 +857,10 @@ impl App {
                                                             // non-burst path so dialog buttons are
                                                             // clickable even when a mouse event lands
                                                             // right after a paste/dictation burst.
+                                                        } else if self.home.handle_sidebar_collapse_click(mouse.column, mouse.row) {
+                                                            // Sidebar collapse/expand toggle; must
+                                                            // precede hit_list (button is on the
+                                                            // list's top border).
                                                         } else if hit_list {
                                                             let action = self.home.handle_click(mouse.column, mouse.row);
                                                             if action.is_none() {
@@ -899,6 +937,29 @@ impl App {
                             continue;
                         }
                         Some(Ok(Event::Mouse(mouse))) => {
+                            // Footer toolbar: a left-click on a button
+                            // synthesizes its shortcut and routes it through
+                            // the full key handler, so clicking behaves
+                            // exactly like pressing the key (global handling,
+                            // action dispatch, structured-view drain). The
+                            // footer is a disjoint area from the list/preview/
+                            // diff, so nothing else in this arm needs to run.
+                            if matches!(mouse.kind, MouseEventKind::Down(MouseButton::Left)) {
+                                if let Some(key) =
+                                    self.home.footer_button_at(mouse.column, mouse.row)
+                                {
+                                    let _ = self.home.clear_preview_selection();
+                                    self.handle_key(key, terminal).await?;
+                                    self.sync_mouse_capture(terminal)?;
+                                    if !self.needs_redraw {
+                                        self.draw(terminal)?;
+                                    }
+                                    if self.should_quit {
+                                        break;
+                                    }
+                                    continue;
+                                }
+                            }
                             let hit_list = self.home.hit_list(mouse.column, mouse.row);
                             let hit_preview = self.home.hit_preview(mouse.column, mouse.row);
                             let hit_diff = self.home.is_diff_open()
@@ -950,6 +1011,17 @@ impl App {
                                         self.set_theme(&name);
                                     }
                                     self.sync_mouse_capture(terminal)?;
+                                    self.draw(terminal)?;
+                                    None
+                                } else if self
+                                    .home
+                                    .handle_sidebar_collapse_click(mouse.column, mouse.row)
+                                {
+                                    // Collapse button (expanded list border) or
+                                    // the collapsed strip toggled the sidebar.
+                                    // Runs before hit_list because the button
+                                    // lives on the list's top border.
+                                    let _ = self.home.clear_preview_selection();
                                     self.draw(terminal)?;
                                     None
                                 } else if self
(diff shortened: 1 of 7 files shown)


src/tui/app.rs
```diff
@@ -899,6 +937,29 @@ impl App {
                             continue;
                         }
                         Some(Ok(Event::Mouse(mouse))) => {
+                            // Footer toolbar: a left-click on a button
+                            // synthesizes its shortcut and routes it through
+                            // the full key handler, so clicking behaves
+                            // exactly like pressing the key (global handling,
+                            // action dispatch, structured-view drain). The
+                            // footer is a disjoint area from the list/preview/
```
jerome-benoit, line 945: The early footer-click arm runs ahead of `handle_dialog_click` and `handle_context_menu_click`. Functionally fine because `footer_button_at` returns `None` while `has_non_live_send_overlay()` is true, so a click cannot fire a shortcut behind a modal. The ordering still reads as a hazard until the reader traces into the helper. A one-line comment here pointing at the helper's overlay guard would save the next reviewer the same trace and answer CodeRabbit's concern in place.

src/tui/home/mod.rs
```diff
@@ -3752,14 +3778,22 @@ impl HomeView {
...
+    /// re-expand it. The next render reflows the preview into the freed
+    /// width; in live mode the resize loop pushes the new geometry to the
+    /// agent's pane. Persisted so the choice survives restarts.
     pub fn toggle_sidebar_collapsed(&mut self) {
         self.sidebar_collapsed = !self.sidebar_collapsed;
+        self.save_sidebar_collapsed();
+    }
+
+    fn save_sidebar_collapsed(&self) {
+        if let Ok(mut config) = load_config().map(|c| c.unwrap_or_default()) {
```
jerome-benoit, line 3791: Silently swallowing `load_config()` errors here is consistent with the existing precedent in `save_list_width` (line 3799) and `toggle_preview_info` (line 3808), so addressing CodeRabbit's finding in this one site alone would split the trio. Two reasonable options:

1. Port the match-arm logging shape to all three call sites in this PR (small, mechanical).
2. Defer with a `// TODO(load-error-logging)` and open a dedicated follow-up that touches all three together.

Inclined to defer so the scope of this PR stays focused on the sidebar/footer/flicker triple. Up to you.

src/tui/home/render.rs
```diff
@@ -751,6 +813,31 @@ impl HomeView {
...
 
+        // Collapse affordance on the top-right border. Clicking it shrinks
+        // the list to the click-to-expand strip. Drawn as an overlay on the
+        // border (after the block) so its clickable rect is known exactly,
+        // and skipped on a list too narrow to spare the columns without
+        // colliding with the title (`render` already zeroed the rect, so the
+        // narrow case needs no else).
+        const COLLAPSE_LABEL: &str = " \u{00AB} ";
+        const COLLAPSE_LABEL_WIDTH: u16 = 3;
+        if area.width > COLLAPSE_LABEL_WIDTH + 6 {
```
jerome-benoit, line 824: `COLLAPSE_LABEL_WIDTH + 6` reads as a bare magic constant. The `6` reserves columns for the title that shares this border row; a named const (e.g. `COLLAPSE_LABEL_TITLE_RESERVE: u16 = 6`) or a one-line comment justifying the threshold would document the intent for future maintainers and make the gate self-explanatory.

src/tui/home/input.rs
```diff
@@ -3955,6 +3996,16 @@ impl HomeView {
             overlay_changed |= dialog.handle_hover(col, row);
         }
 
+        // Footer-toolbar hover: the index drives the inverted-chip highlight
+        // on the next render. Recomputed against the current button rects so
+        // it clears the moment the pointer leaves a button.
+        let prev_footer_hover = self.footer_hover;
```
jerome-benoit, line 4002: `footer_hover` indexes into a `footer_buttons` vector rebuilt every frame, with potentially different ordering after sort/group/view-mode changes. Without a `Moved` event between the geometry change and the next render, the index can paint the inverted-chip highlight on the wrong button (or end up out of bounds, though the hover recompute on the next move corrects it).

Practical impact is one stale frame at most, so this is cosmetic. Invalidating `footer_hover` on the geometry-affecting transitions (`set_sort_order`, `set_group_by`, view-mode switch) would close the window deterministically without depending on the next mouse event.

jerome-benoit (commented):
Read through head `a31c1d10`. The flicker fix via `theme_apply_needed` is well-targeted and the tuple-identity test locks the regression. Click isolation under overlays is correctly mitigated at the helper level (`footer_button_at` and `handle_sidebar_collapse_click` both return early on `has_non_live_send_overlay()`), and the frame-start hit-rect cleanup at `render.rs:472` is the right scope to keep takeover views from inheriting stale geometry.

Four non-blocking notes inline, all minor:

- ordering documentation in the click router (`app.rs`),
- consistency of error logging across the persistence trio (`mod.rs`),
- a magic constant in the collapse-button visibility gate (`render.rs`),
- a small hover-index staleness window across reorderings (`input.rs`).

Tests cover the regression claims directly: `theme_apply_needed_compares_name_and_palette_mode` for the flicker, `takeover_view_clears_sidebar_hit_rects` for click isolation under takeover views, and the `footer_toolbar` module for the hit-rect-to-key contract including strict-mode chords and live-send suppression.

Nothing here blocks the merge.

src/tui/home/render.rs
```diff
@@ -539,20 +547,28 @@ impl HomeView {
...
+            // otherwise keep last frame's values and a click in the now-
+            // preview area could resolve to an invisible list row (and
+            // switch the live target). Zero them so mouse hit-testing can't
+            // target the hidden sidebar; the strip carries its own rect.
             self.list_area = Rect::default();
             self.list_inner_area = Rect::default();
-            self.render_preview(frame, main_chunks[0], theme);
+            self.collapse_button_area = Rect::default();
+            self.render_collapsed_strip(frame, chunks[0], theme);
+            self.render_preview(frame, chunks[1], theme);
```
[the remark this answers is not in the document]
jerome-benoit: @njbrake: Pre-existing concern but worth tracking. In stacked mode (`< STACKED_BREAKPOINT` cols) `render_list` uses `Borders::TOP | LEFT | BOTTOM` and `render_preview` uses `Borders::ALL`, so the list's bottom border + the preview's top border render as a double horizontal seam, which violates the same `DESIGN.md` rule the new collapsed strip respects. Not introduced by this PR, but the new strip makes the stacked-mode gap more visible by contrast.

Suggest scoping the layout-aware fix to a small follow-up issue rather than expanding #2294. Sketch: thread a `Layout::{SideBySide, Stacked, Collapsed}` enum into `render_list` / `render_preview` and mask borders so exactly one block carries the shared seam in each branch. Anchors: list borders `src/tui/home/render.rs:798`, preview borders `:1903`, stacked branch `:577-593`.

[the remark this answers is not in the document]
jerome-benoit: Yes, do it please with following the template.

njbrake:
Whoops sorry Jerome I merged this before I saw your comments. Glad there were no blockers, I'll handle your comments in a follow up PR 🙏

PR #2299, opened 2026-06-20
fix(tui): address #2294 review follow-ups (footer hover staleness, persistence logging, docs)

njbrake:
## Description

Follow-up to #2294. That PR merged with four non-blocking review notes from @jerome-benoit left unaddressed (the last commit predated his review, so nothing was pushed after it). This PR clears all four.

- **Footer hover staleness (`input.rs`, `render.rs`, `mod.rs`).** `footer_hover` was an index into the per-frame `footer_buttons` vector. A sort/group/view-mode change between a mouse move and the next render could reorder the toolbar, so the cached index could highlight the wrong button (or, before the next move recomputed it, go out of bounds). It is now keyed by the button's `KeyEvent` shortcut, so the highlight follows the right button across any reorder and can never index a button that no longer exists. Chosen over scattering hover invalidation across the roughly eight sort/group/view-mode mutation sites because it is centralized and also closes the out-of-bounds case.
- **Persistence error logging (`mod.rs`).** The home view's `app_state` persisters (sidebar collapse, list width, preview info, archived section) silently dropped `load_config()` failures. They now route through a shared `persist_app_state` helper that logs both the load and save failure paths, so a load error is no longer swallowed in one persister while being reported in another.
- **Magic constant (`render.rs`).** `COLLAPSE_LABEL_WIDTH + 6` is now a named `COLLAPSE_LABEL_TITLE_RESERVE` const with a comment explaining the columns reserved for the title sharing that border row.
- **Ordering doc (`app.rs`).** The footer-click arm runs ahead of the dialog and context-menu click handlers; a comment now records why that is safe (`footer_button_at` returns `None` under an overlay via its `has_non_live_send_overlay()` guard).

A separate concern from the same review (the pre-existing stacked-mode double horizontal seam, raised on the `render.rs:576` thread) is tracked in #2301 rather than expanded here.

## PR Type

- [x] Bug Fix
- [x] Refactor
- [x] Documentation

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a new e2e test, because: the hover contract is covered by the existing `footer_toolbar::hover_tracks_button_under_pointer` unit test, updated here to assert the hovered button by `KeyEvent` identity. Switching the field to a `KeyEvent` makes the out-of-bounds failure mode unrepresentable at the type level, so a byte-level reorder e2e adds little. `theme_apply_needed_compares_name_and_palette_mode`, the `footer_toolbar` module, and the sidebar/takeover tests from #2294 continue to pass.

## How tested

- `cargo fmt`, `cargo clippy --lib --tests` (no new warnings), `cargo test --lib tui::home` (595 passing) and `cargo test --lib tui::app` (40 passing), including the `footer_toolbar` and `theme_apply_needed` regression tests.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Used to address the review follow-ups and write the tests, in back-and-forth with @njbrake.

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -944,6 +944,11 @@ impl App {
                             // action dispatch, structured-view drain). The
                             // footer is a disjoint area from the list/preview/
                             // diff, so nothing else in this arm needs to run.
+                            // This runs ahead of the dialog/context-menu click
+                            // handlers, but that is not a hazard: when an
+                            // overlay is open `footer_button_at` returns `None`
+                            // (its `has_non_live_send_overlay()` guard), so a
+                            // click can never fire a shortcut behind a modal.
                             if matches!(mouse.kind, MouseEventKind::Down(MouseButton::Left)) {
                                 if let Some(key) =
                                     self.home.footer_button_at(mouse.column, mouse.row)
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -4023,14 +4023,16 @@ impl HomeView {
             overlay_changed |= dialog.handle_hover(col, row);
         }
 
-        // Footer-toolbar hover: the index drives the inverted-chip highlight
-        // on the next render. Recomputed against the current button rects so
-        // it clears the moment the pointer leaves a button.
+        // Footer-toolbar hover: the hovered button's shortcut drives the
+        // inverted-chip highlight on the next render. Recomputed against the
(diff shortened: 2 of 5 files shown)


njbrake:

The fifth item from the #2294 review, the pre-existing stacked-mode double horizontal border seam raised by @jerome-benoit, is intentionally out of scope for this PR and is tracked separately in #2301.

PR #2308, opened 2026-06-20
fix(acp): stop agent messages rendering twice from leaked consolidated chunk

### LINKED ISSUE #2281: bug: Text is duplicated when switching to Sonnet
Hi, no AI here. I have just started using the web UI, and I noticed that when I switch to Sonnet, the first messages it says are duplicated.

Image

I don't have a great reproducing setup:
- Create a session (typically with claude code, Opus default)
- Send a message to Opus
- Switch to Sonnet
- Send a message
- Voilà

I don't know if it reproduces on other platforms:
Ubuntu 22.04 amd64
aoe v1.11.1 built from source this morning

Cheers if you can look at it @Seluj78 👀

Seluj78:

## Description

In the structured view an agent message intermittently rendered twice, the full text repeated back-to-back. It reproduced deterministically right after switching model (for example Opus to Sonnet) and intermittently otherwise.

Root cause: `@agentclientprotocol/claude-agent-acp` streams an assistant text block as incremental `agent_message_chunk` deltas, then re-sends the whole block as one consolidated chunk. Its `streamedTextIds` dedup is meant to suppress that consolidated copy, but it misses on a message-id mismatch (a model switch starts a fresh stream id, which is exactly that condition), so both the deltas and the consolidated restatement reach us. aoe recorded every chunk into the event store and every reducer (web and TUI) appends each chunk into one assistant bubble, so the block showed up twice.

This fix tracks the in-flight assistant text block in the `on_receive_notification` handler in `src/acp/acp_client.rs`, keyed by `message_id`. The schema guarantees a change in `message_id` marks a new message; the streamed deltas and the empty block-start marker share the streamed id while the leaked restatement carries a different one. A non-empty chunk whose id differs from the open block and whose text restates the block's accumulated text is dropped before it reaches the watchdog, the event store, or any client. Same-id chunks are genuine deltas and are never dropped (a legitimately repeated delta keeps its id), and absent ids on either side degrade to never-drop so real output is never corrupted. Dropping the whole update keeps the duplicate out of every surface at once and lets the reducers stay dumb appenders. The deduper is reset during post-load history replay so replayed chunks cannot poison live tracking.

Already-stored doubled transcripts are left as-is (the event store is append-only; no migration). The underlying upstream adapter leak will be reported separately to claude-agent-acp.

Closes #2281

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open a structured-view session, send a prompt to Opus, switch the model to Sonnet, send another prompt; the first Sonnet reply renders exactly once.
- [ ] In any session, send a prompt that yields a multi-sentence reply; confirm the reply is not duplicated back-to-back.
- [ ] Reload the web view of a session whose recent turn previously doubled; the message now renders once for newly produced turns (old stored transcripts keep the historical artifact).

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: the fix is server-side in the ACP ingress (`src/acp/acp_client.rs`); no `web/src` files change. It is covered by Rust unit tests on `AgentMessageDedup` (streamed-then-restated drops, single-delta restatement drops, legit repeated same-id delta kept, boundary reset, absent-id never-drop, replay reset).

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation were drafted by Claude under my direction. The design was vetted through a two-model debate (gemini + gpt-5.5) that settled the placement (server-side over a web-only reducer patch) and the message-id discriminator over naive text equality. I made the design calls and ran the validation.

**NOTE:**

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2308"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784551967&installation_id=139138837&pr_number=2308&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2308&signature=daa5c7e010c1a290e1c07c2f4bb0c473a331af78516148430b9f943f8461aa48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## What Changed

- **Added server-side deduplication** in `src/acp/acp_client.rs` to prevent assistant messages from being rendered **twice** in the structured view when the upstream ACP adapter “leaks” a consolidated restatement after streaming smaller text deltas.
- **Introduced an in-memory tracker** that watches the currently streaming assistant text block and **suppresses the leaked duplicate** (while keeping the real streamed deltas).
- **Hardened reset behavior** so the deduplication window is scoped correctly to a single turn:
  - reset when a new `ClientCmd::Prompt` starts
  - reset again right before emitting the terminal `Event::Stopped` for that turn
- **Added unit tests** covering leaked restatements, same-id delta behavior, boundary resets, and “never-drop” safety when IDs are missing.
- Rewrapped **import formatting only** (no logic change).

## Benefits

- **Fixes the duplicate-message bug** that could deterministically appear when switching models and intermittently in other scenarios.
- **Prevents duplicates from entering the event store or reaching any client surfaces**, instead of only hiding them later.
- **Reduces risk of data loss** by only dropping the specific leaked consolidated restatement pattern, while preserving chunks when identifiers are missing.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -4208,6 +4208,12 @@ async fn run_connection_task<W, R>(
     // section stays synchronous and the guard never crosses an await.
     let agent_msg_dedup = Arc::new(std::sync::Mutex::new(AgentMessageDedup::default()));
     let agent_msg_dedup_for_notif = agent_msg_dedup.clone();
+    // The prompt loop resets the deduper at turn boundaries (a new prompt, and
+    // the turn's terminal Stopped). Turn completion is not a SessionUpdate, so
+    // without this an open text block could survive into the next turn and a
+    // new turn that legitimately reuses the prior turn's trailing text under a
+    // fresh message_id would be misclassified as a restatement. See #2281.
+    let agent_msg_dedup_for_block = agent_msg_dedup.clone();
 
     let result = Client
         .builder()
@@ -4898,6 +4904,13 @@ async fn run_connection_task<W, R>(
                 let cmd = cmd_rx.recv().await;
                 match cmd {
                     Some(ClientCmd::Prompt(blocks)) => {
+                        // Scope the agent-message deduper to one turn: a new
+                        // prompt starts a fresh assistant block, so forget any
+                        // block left open by the prior turn. See #2281.
+                        agent_msg_dedup_for_block
+                            .lock()
+                            .expect("agent message dedup mutex poisoned")
+                            .reset();
                         // First user prompt after session/load: stop
                         // dropping notifications. The agent's history-
                         // replay window is over; everything from now on
@@ -5482,6 +5495,14 @@ async fn run_connection_task<W, R>(
                                 reason: reason.into(),
                             })
                             .await;
+                        // Turn ended: close any open assistant text block so a
+                        // restatement can never be matched across the turn
+                        // boundary. The next prompt also resets, this is the
+                        // belt to that suspenders. See #2281.
+                        agent_msg_dedup_for_block
+                            .lock()
+                            .expect("agent message dedup mutex poisoned")
+                            .reset();
                         if shutdown {
                             break;
                         }
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -23,14 +23,15 @@ use agent_client_protocol::schema::{
     CreateTerminalResponse, ElicitationAction, ElicitationCapabilities,
     ElicitationFormCapabilities, EmbeddedResource, EmbeddedResourceResource,
     FileSystemCapabilities, ImageContent, InitializeRequest, KillTerminalRequest,
-    KillTerminalResponse, LoadSessionRequest, McpServer, NewSessionRequest, PermissionOptionKind,
-    PromptRequest, ProtocolVersion, ReadTextFileRequest, ReadTextFileResponse,
-    ReleaseTerminalRequest, ReleaseTerminalResponse, RequestPermissionOutcome,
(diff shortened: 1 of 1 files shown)


Seluj78:
@coderabbitai Good catch on the cross-turn boundary. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2308/commits/f95f4cec3624d5f73466065b726de1c94afa19d2: the deduper now resets at both turn boundaries, the start of each `ClientCmd::Prompt` and the turn's terminal `Stopped`, so the dedup window is scoped to one assistant block. A new turn that legitimately reuses the prior turn's trailing text under a fresh message_id is no longer misclassified as a restatement. Covered by the existing `dedup_reset_forgets_in_flight_block` unit test.

njbrake (approved):

PR #2326, opened 2026-06-22
fix(acp): end agent-initiated turns that run with no driving prompt

### LINKED ISSUE #2325: Structured view / web: session stuck running (green) after a Monitor fires post-turn, agent-initiated between-prompt work has no terminal Stopped

## Problem

A structured-view / web session stays marked running (green, `turn_active = true`) forever after the agent has finished writing, when the final turn was started by a Monitor (or any backgrounded task) firing **after** the prompt turn that armed it had already completed.

Concretely: the agent armed a Monitor ("notify me when the background writer finishes"), the prompt turn that armed it ended cleanly, and later the writer finished and the Monitor woke the agent into a brand-new turn. The agent ran tools and wrote its full report, then went quiet. The UI never cleared the spinner. The worker process is still alive; the only escape is `aoe acp restart <id>`.

This is distinct from #2237 (a turn that was already in flight when a worker reattached / respawned). Here the arming turn completed normally and a fresh agent-initiated turn started **between** prompts.

Session `5c8d250f60ec4328` (release namespace, `~/.agent-of-empires/acp_events.db`). Lifecycle, oldest to newest:

```
  8  UserPromptSent {"/aoe-triage"}
341  Stopped {prompt_complete}
342  UserPromptSent {"ok go"}
404  ToolCallStarted Monitor (task bv1z8qqub)
406  MonitorArmed {"Wait for the writer script to finish"}
425  ToolCallStarted Monitor (task bp3pzc36v)
427  MonitorArmed {"Wait until writer prints completion marker"}
443  Stopped {prompt_complete}          <- "ok go" turn ends, monitor still pending
444  UserPromptSent {"?"}
462  Stopped {prompt_complete}          <- "?" turn ends, monitor still pending
463  UsageUpdated                       <- FRESH TURN, no preceding UserPromptSent
...  ToolCallStarted/Completed + AgentMessageChunks (the full triage report)
541  UsageUpdated                       <- agent finished writing here
542  AvailableCommandsUpdated           <- periodic slash-command refresh, not turn work
543  AvailableCommandsUpdated
```

Key facts:

- seq 463-541 is a complete agent turn with **no `UserPromptSent` before it**. The Monitor `bp3pzc36v` fired (writer printed its completion marker) and resumed the agent.
- The last real turn output is seq 541. There is **no `Stopped` event after it.** seq 542/543 are only `AvailableCommandsUpdate` pushes (the adapter refreshing the slash-command list), not turn activity.
- Worker pid is still alive; the session was never reaped.

Worker log (`~/.agent-of-empires/acp-workers/5c8d250f60ec4328.log`) confirms the turn just trails off, no watchdog fire, then the agent websocket disconnects:

```
07:52:40  TRACE acp.event_store: recorded event seq=541 kind=usage_updated
07:53:32  DEBUG acp.ws: agent ws disconnected
07:55:58  DEBUG acp.protocol: received AvailableCommandsUpdate ... seq=542
07:56:28  DEBUG acp.protocol: received AvailableCommandsUpdate ... seq=543
```

## Root cause

`Event::Stopped` is the only thing that clears `turn_active`:

`src/tui/structured_view/reducer.rs:461`
```rust
Event::Stopped { reason } => {
    self.flush_pending_chunk();
    self.status_text = Some(format!("stopped: {reason}"));
    // ...
    self.turn_active = false;
}
```

But `Stopped` is only ever emitted from inside the prompt-drain `select!` loop, after an aoe-issued `session/prompt` resolves:

`src/acp/acp_client.rs:5355` (end of the prompt cycle)
```rust
let reason = terminal_stop_reason(
    rate_limited, force_stopped, prompt_orphaned, agent_unresponsive,
    shutdown, prompt_cancelled,
);
let _ = event_tx_for_block
    .send(Event::Stopped { reason: reason.into() })
    .await;
```

The silent-orphan watchdog that could synthesize a `Stopped` for a wedged turn (`should_fire` / `terminal_stop_reason` / `effective_grace`, `acp_client.rs:804-888`) is constructed per-prompt and lives **only for the duration of one in-flight `session/prompt`** (`acp_client.rs:4826-4857`). Between prompts the command loop blocks on `cmd_rx.recv()` (`acp_client.rs:4781`) with no watchdog armed.

The notification handler does keep running between prompts and **publishes** agent-initiated events to the sink regardless of prompt state (`acp_client.rs:4215`), which is why seq 463-541 got persisted. But it only routes lifecycle signals (the ones that drive Stopped) when a prompt epoch is active (`acp_client.rs:4175-4196`). So a Monitor that fires between prompts produces fully-rendered output with zero terminal-Stopped coverage.

None of the six synthetic-`Stopped` sites in `src/acp/supervisor.rs` cover this case either; they fire on orphan reattach (`1017`), user stop (`1834`), shutdown (`2241`), orphaned-approval cleanup (`2550`), reap-user-stopped (`2682`), and force-end-turn (`2094`). A monitor-resumed between-prompt turn going idle matches none of them.

In short: aoe models a session as "running" purely as a function of an aoe-driven prompt cycle. Agent-initiated work between prompts (Monitor wake, scheduled wakeup, any background-task notification that resumes the agent) has no owning prompt cycle, so when it goes quiet nothing emits `Stopped`.

## Proposal

**(A) Minimum viable: a between-prompt idle watchdog.** When agent-initiated events arrive while no prompt is in flight, arm a lightweight idle timer keyed on `last_event_at`. When it goes quiet past a grace window (reuse `effective_grace` / the off-protocol floor so a still-pending monitor is not killed early), emit a synthetic `Event::Stopped { reason: "prompt_complete" }` (or a new `reason: "agent_idle"`). This mirrors the existing resume-idle watchdog (`acp_client.rs:4721-4778`) but covers the between-prompt window instead of the post-reattach window. Smallest change; keeps the fix entirely daemon-side so every surface (web, TUI structured view) inherits it.

**(B) Structural: treat agent-initiated activity as a first-class turn.** When the notification handler observes the first agent-initiated event with no active prompt epoch, open a synthetic prompt epoch so the full per-prompt watchdog + cancel-escalation machinery applies, and close it (emitting `Stopped`) the same way a real prompt turn closes. More faithful to the turn model and gives the Stop button something to cancel during a monitor-resumed turn (related to #1727 A and #2237), but a bigger change to the epoch bookkeeping.

(A) is the right first move; (B) is the follow-on if we want Stop-button parity for between-prompt turns.

## Why now

- Monitor / `until` / background-job loops are a common pattern (#1727, #2237 both stem from it) and this leaves the session permanently green, indistinguishable from a genuinely working agent.
- Only escape today is `aoe acp restart <id>`, the same dead-end as #2237.
- The off-protocol grace plumbing (`WakeupPending`, `off_protocol_grace_floor`) already exists; proposal (A) reuses it rather than inventing new timing.

Implementation MUST add tests covering each story below; PR is not done without them. (Drafted from the investigation; confirm or adjust before merge.)

- As a user running a session whose agent armed a Monitor, when the Monitor fires after the arming prompt turn already ended and the resumed turn finishes writing, then the session must transition out of "running" (a terminal `Stopped` is emitted and `turn_active` becomes false).
  - Suggested test layer: Rust unit/integration in `src/acp/` (feed a synthetic between-prompt event stream into the client and assert a `Stopped` is emitted once activity goes idle), plus a live Playwright (`web/tests/live/`) check that the spinner clears.
- As a user whose Monitor is still pending (writer not finished), when the agent is deliberately idle waiting on the background task, then the between-prompt watchdog must NOT fire a premature `Stopped` before the off-protocol grace window elapses.
  - Suggested test layer: Rust unit test on the between-prompt watchdog grace logic.
- As a user on the web structured view, when a monitor-resumed turn completes, then the sidebar dot and composer spinner return to idle without an `aoe acp restart`.
  - Suggested test layer: live Playwright (`web/tests/live/`).

## Open questions

- Reason string: reuse `"prompt_complete"` for the synthetic between-prompt Stopped, or introduce `"agent_idle"` so the source is distinguishable in the event log? `prompt_complete` is simplest and matches what a real turn would emit; a distinct reason aids debugging.
- Grace window for the between-prompt watchdog: reuse `off_protocol_grace_floor` (30 min) wholesale, or a shorter idle grace once `cost_seen` is observed (the end-of-turn accounting marker, `acp_client.rs:844-852`)? A finished monitor turn does emit a cost-bearing `UsageUpdate`, so `cost_seen` could shorten the wait.
- Should proposal (B) be split into its own issue so this one ships (A) quickly? (A) alone fixes the stuck-green symptom; (B) is about Stop-button reach into between-prompt turns.

## Out of scope

- Stop-button behavior during monitor / `until` loops while a turn IS active (#1727 A, still its own concern).
- The "Waiting on model" spinner masking an intentional sleep (#1727 B).
- Reattach-adopted in-flight turns losing their PromptResponse (#2237, closed).

## Cross-references

- #2237 (closed) — the symmetric residual: an in-flight turn adopted on reattach loses its `PromptResponse` and never emits `Stopped`. This issue is the between-prompt sibling (arming turn completed cleanly, fresh agent-initiated turn never closes).
- #1727 (closed) — same monitor / background-job turn model; covers Stop-button reach and the intentional-sleep spinner, not the post-monitor stuck-green case.
- #2248 (closed) — adjacent structured-view status-trap after idle auto-stop + re-prompt.
- #1007 (closed) — detect agent process death proactively instead of waiting for the next prompt; same "nothing watches between prompts" theme.

Seluj78:

## Description

A structured-view / web session stayed marked running (green) forever when the agent's final turn was started by a claude-code Monitor (or any backgrounded task) firing after the prompt that armed it had already completed. That resumed turn runs with no aoe-issued `session/prompt`, so the per-prompt silent-orphan watchdog never sees it and no terminal `Stopped` is ever emitted; `turn_active` stays true and the only escape was `aoe acp restart`.

The daemon now tracks agent-initiated activity that happens between prompts via shared atomics maintained by the notification handler (`last_lifecycle_at`, `between_prompt_active`, `between_prompt_cost_seen`, `between_prompt_wake_until`, `prompt_in_flight`), mirroring the per-prompt watchdog's cost / progress / wake semantics. An idle tick is folded into the outer command loop's `recv` `select!`, so it is polled only while the loop is parked between prompts; during a real prompt the inner drain owns the connection and the per-prompt watchdog stays the sole idle authority. When the agent-initiated turn goes idle past grace (a short 3 second grace once the cost-bearing `UsageUpdate` end-of-turn marker has arrived, otherwise the 30-minute off-protocol floor, and suppressed entirely while a scheduled wake is still pending) the loop emits a clean `Stopped { reason: "agent_idle" }`. A 1 second check cadence (vs the per-prompt 5 seconds) keeps the badge and status clearing within a few seconds of the turn finishing.

Emitting the synthetic `Stopped` from the command loop, never a detached timer, keeps it serialized with every other command so it cannot race a new prompt's events and reorder the UI. `"agent_idle"` is a clean stop: the supervisor only restarts on `agent_unresponsive` / `prompt_orphaned` / `user_forced` / `rate_limited`, so a completed between-prompt turn does not burn restart budget. The fire decision is the pure, unit-tested `between_prompt_should_fire` helper.

This is a daemon-side fix, so every surface (web dashboard and the native structured view) inherits it.

The fix also clears the "monitoring" badge and banner. Those are driven by `MonitorArmed` and previously cleared only on a following `UserPromptSent`, with a documented ceiling: a monitor that fired and finished with no user prompt left the badge up indefinitely (the `Monitor` tool call itself completes at arm time, so its completion is not the disarm signal). The badge now clears once the monitor fires and that turn ends: the fire makes the agent act (a tool call started after the arm), and the badge retires on the next `Stopped`. This covers both observed shapes, the agent ending the arming turn and resuming later between prompts (closed by `agent_idle`) and the monitor blocking the arming turn in-band (closed by `prompt_complete`). A `Stopped` with no post-arm tool work is the arming turn ending while the monitor is still pending, so it deliberately leaves the badge up. Applied in the web reducer (`monitorWorkSeen` gate) and the server-side `latest_active_monitor` query.

Closes #2325

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view session with claude (yolo). Have the agent kick off a long-running background script, arm a Monitor to be notified when it finishes, then end its turn. Confirm the turn completes normally (spinner clears, session idle) while the background work is still running.
- [ ] Let the background script finish so the Monitor fires. The agent resumes on its own, streams a final report, then goes quiet. Confirm that within a few seconds of the agent's last output the session returns to idle by itself, with no green "running" spinner and without `aoe acp restart`.
- [ ] Negative case: while the agent is deliberately asleep waiting on a pending monitor / scheduled wake, confirm the session is NOT prematurely forced to idle during the wait window.
- [ ] Confirm the "monitoring" sidebar badge and the top banner also clear once the resumed monitor turn goes idle (not just the running spinner), with no `aoe acp restart`.

## Checklist

## Test Coverage Analysis

The user-visible behavior (spinner clearing) is driven by the daemon-side synthetic `Stopped`, whose decision logic is covered by Rust unit tests (`between_prompt_should_fire`). The one `web/src` change is the reducer clearing the monitoring badge on `agent_idle`, covered by a Vitest case in `web/src/lib/acpTypes.test.ts` (badge persists on `prompt_complete`, clears on `agent_idle`); reducer state transitions are pure logic, the Vitest layer per `AGENTS.md`. Driving a real claude-code Monitor wake end-to-end through Playwright is not deterministic, so the core paths are unit-tested instead. `web/tests/validate-coverage-matrix.mjs` passes.

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: the between-prompt idle decision is covered by deterministic Rust unit tests on `between_prompt_should_fire` (fast-grace fire, off-protocol floor, pending-wake suppression, inactive no-fire). A full live e2e would need a real agent to arm a Monitor and have it fire after the arming turn ends, which is not deterministic in the harness.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation were drafted by Claude under my direction, including a `/debate` consult that pushed the design from a free-running timer to a command-loop-owned idle tick to avoid out-of-order `Stopped` emission. I made the design calls and reviewed each commit.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2326"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784716290&installation_id=139138837&pr_number=2326&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2326&signature=4d87438e50b245f0a783f1ff11113a40863792c961de5e65b465fa781b2e19f1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Overview

This PR fixes sessions that could stay stuck in a “running” (green) state indefinitely after a Monitor (or scheduled wake) fires but the agent then finishes an agent-driven turn with no terminal stop event emitted. It also fixes the web “monitoring” badge so it no longer persists after the monitor’s work is done.

### 1) Daemon: between-prompt idle watchdog
- Added a new “between prompts” idle check that watches for agent activity that ends after emitting the end-of-turn cost marker, but before the next user prompt.
- When the agent goes idle long enough in that between-prompt window, the daemon now emits a clean terminal event: `Stopped { reason: "agent_idle" }`.
- This event is generated from the main command loop (not a detached timer) to keep event ordering consistent.

### 2) Daemon: correct monitor lifecycle → correct UI badge clearing signal
- Updated backend logic so the UI only clears the “monitoring” badge after:
  - the monitor actually starts its work (not just when it arms), and
  - the turn ends afterward.
- This prevents the badge from disappearing too early or remaining stuck when a monitor fires without a subsequent user prompt.

### 3) Frontend (web): monitoring badge state tracking
- Added `monitorWorkSeen` to the web ACP reducer state.
- The reducer now clears `monitorArmed` / badge only when both:
  - the monitor work has started, and
  - a terminating `Stopped` frame arrives.
- Added reducer tests covering both `prompt_complete` and `agent_idle` stop reasons around monitor firing.

### 4) Tests
- Added unit tests for the new between-prompt idle watchdog “helper” logic (including grace timing, suppression while wake is pending, and regression coverage to avoid premature firing).
- Extended tests for the updated monitor badge retirement behavior.

## Benefits
- Prevents sessions from getting stuck in “running” indefinitely and removes the need for manual restarts.
- Ensures the monitoring badge accurately reflects when a monitor has actually fired and the turn has completed.
- Works across all daemon-consumed surfaces since the main behavioral fix is daemon-side.

### DIFF
--- a/docs/structured-view/troubleshooting.md
+++ b/docs/structured-view/troubleshooting.md
@@ -242,12 +242,6 @@ container (rather than bind-mounting from the host).
   marker, which made the daemon refuse to bring the worker back after it next
   exited; a follow-up message would then sit unsent. A worker coming online now
   clears that marker, so this no longer strands a queued message.
-- An agent that armed a monitor or a scheduled wake, finished the turn, then was
-  woken later by that background task used to keep the session marked "running"
-  forever: the resumed turn ran with no driving prompt, so nothing ended it. The
-  daemon now ends such a turn cleanly once it goes idle (a few seconds after
-  the agent's last output, longer while a fresh wake is still pending), so the
-  session returns to idle on its own without `aoe acp restart`.
 
 ### "Restarting worker" banner after a turn looked done
 
--- a/docs/structured-view/troubleshooting.md
+++ b/docs/structured-view/troubleshooting.md
@@ -245,7 +245,7 @@ container (rather than bind-mounting from the host).
 - An agent that armed a monitor or a scheduled wake, finished the turn, then was
   woken later by that background task used to keep the session marked "running"
   forever: the resumed turn ran with no driving prompt, so nothing ended it. The
-  daemon now ends such a turn cleanly once it goes idle (about 20 seconds after
+  daemon now ends such a turn cleanly once it goes idle (a few seconds after
   the agent's last output, longer while a fresh wake is still pending), so the
   session returns to idle on its own without `aoe acp restart`.
 
--- a/docs/structured-view/troubleshooting.md
+++ b/docs/structured-view/troubleshooting.md
@@ -242,6 +242,12 @@ container (rather than bind-mounting from the host).
   marker, which made the daemon refuse to bring the worker back after it next
   exited; a follow-up message would then sit unsent. A worker coming online now
   clears that marker, so this no longer strands a queued message.
+- An agent that armed a monitor or a scheduled wake, finished the turn, then was
+  woken later by that background task used to keep the session marked "running"
+  forever: the resumed turn ran with no driving prompt, so nothing ended it. The
+  daemon now ends such a turn cleanly once it goes idle (about 20 seconds after
+  the agent's last output, longer while a fresh wake is still pending), so the
+  session returns to idle on its own without `aoe acp restart`.
 
 ### "Restarting worker" banner after a turn looked done
 
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -464,6 +464,24 @@ const SILENT_ORPHAN_FAST_GRACE_DEFAULT: std::time::Duration = std::time::Duratio
 /// notification handler to reach back into a pinned `tokio::time::sleep`.
 const SILENT_ORPHAN_CHECK_INTERVAL: std::time::Duration = std::time::Duration::from_secs(5);
 
+/// Idle grace for the between-prompt watchdog once the cost-bearing
+/// end-of-turn `UsageUpdate` has arrived. Much shorter than the per-prompt
+/// `SILENT_ORPHAN_FAST_GRACE_DEFAULT`: that one also waits out a
+/// possibly-late `PromptResponse` over the wire, but a between-prompt
+/// agent-initiated turn has no RPC to wait for, so once it emits its
+/// end-of-turn marker and goes quiet a few seconds is enough. Kept low so
+/// the "monitoring" badge and running status clear promptly after a monitor
+/// turn finishes. A turn that actually continues emits fresh progress,
+/// which resets the idle timer and clears `cost_seen`, so this cannot cut a
+/// live turn short. See #2325.
+const BETWEEN_PROMPT_IDLE_GRACE: std::time::Duration = std::time::Duration::from_secs(3);
+
+/// Tick cadence for the between-prompt idle check. Faster than
+/// `SILENT_ORPHAN_CHECK_INTERVAL` so the badge and status clear within a few
+/// seconds of the turn ending. Only polled while the command loop is parked
+/// between prompts, so the extra wakeups are cheap. See #2325.
+const BETWEEN_PROMPT_IDLE_CHECK_INTERVAL: std::time::Duration = std::time::Duration::from_secs(1);
+
 /// Classification of an inbound ACP `SessionUpdate` for the silent-
 /// orphan watchdog state machine. Sent from the notification handler
 /// to the prompt loop via a dedicated mpsc; the prompt loop owns the
@@ -5004,7 +5022,7 @@ async fn run_connection_task<W, R>(
             // command loop (never a detached task) keeps it serialized with
             // every other command, so it can't race a new prompt's events.
             let mut between_prompt_idle_tick =
-                tokio::time::interval(SILENT_ORPHAN_CHECK_INTERVAL);
+                tokio::time::interval(BETWEEN_PROMPT_IDLE_CHECK_INTERVAL);
             between_prompt_idle_tick
                 .set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
             loop {
@@ -5018,7 +5036,7 @@ async fn run_connection_task<W, R>(
                             last_lifecycle_at.load(Ordering::Relaxed),
                             between_prompt_wake_until.load(Ordering::Relaxed),
                             between_prompt_cost_seen.load(Ordering::Relaxed),
-                            SILENT_ORPHAN_FAST_GRACE_DEFAULT,
+                            BETWEEN_PROMPT_IDLE_GRACE,
                             OFF_PROTOCOL_WORK_GRACE_FLOOR,
                         ) {
                             between_prompt_active.store(false, Ordering::Relaxed);
@@ -6657,8 +6675,9 @@ mod tests {
     }
 
     // Between-prompt idle watchdog fire decision (#2325). Wall-clock millis.
-    const FAST: std::time::Duration = std::time::Duration::from_secs(20);
-    const FLOOR: std::time::Duration = std::time::Duration::from_secs(30 * 60);
+    // Bind to the production constants so the test tracks the real grace.
+    const FAST: std::time::Duration = BETWEEN_PROMPT_IDLE_GRACE;
+    const FLOOR: std::time::Duration = OFF_PROTOCOL_WORK_GRACE_FLOOR;
 
     #[test]
     fn between_prompt_inactive_never_fires() {
@@ -6671,20 +6690,21 @@ mod tests {
     #[test]
     fn between_prompt_fires_after_fast_grace_when_cost_seen() {
         let last = 1_000_000;
-        // 19s idle: still within the 20s fast grace.
+        let grace_ms = FAST.as_millis() as i64;
+        // Just under the fast grace: still waiting.
         assert!(!between_prompt_should_fire(
             true,
-            last + 19_000,
+            last + grace_ms - 500,
             last,
             0,
             true,
             FAST,
             FLOOR
         ));
-        // 21s idle: past the fast grace, the completed turn ends.
+        // Past the fast grace: the completed turn ends.
         assert!(between_prompt_should_fire(
             true,
-            last + 21_000,
+            last + grace_ms + 500,
             last,
             0,
             true,
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -922,6 +922,55 @@ fn terminal_stop_reason(
 /// otherwise the vendor-agnostic off-protocol floor governs. A pending
 /// scheduled wake (`wake_until` in the future) suppresses firing so a
 /// legitimately-sleeping monitor is never killed early.
+/// State update the between-prompt watchdog should apply for one inbound
+/// notification's classified signals. `None` when neither a lifecycle nor a
+/// wakeup signal is present (ambient updates do not touch the watchdog).
+///
+/// Extracted as a pure function so the cost / progress / wake bookkeeping is
+/// unit-testable without the notification closure. Every tracked signal
+/// refreshes `last_lifecycle_at` to `now_ms`, including `TerminalUsage`: the
+/// cost-bearing `UsageUpdate` is the end-of-turn marker, and the fast grace
+/// must measure from it (when the turn wrapped up) rather than from a
+/// possibly-stale earlier progress event. See #2325.
+#[derive(Debug, PartialEq)]
+struct BetweenPromptUpdate {
+    cost_seen: bool,
+    last_lifecycle_at: i64,
+    wake_until: i64,
+}
+
+fn between_prompt_signal_update(
+    lifecycle: Option<&LifecycleSignal>,
+    wakeup: Option<&LifecycleSignal>,
+    now_ms: i64,
+    prev_wake_until: i64,
+) -> Option<BetweenPromptUpdate> {
+    let mut update = match lifecycle {
+        Some(LifecycleSignal::TerminalUsage) => Some(BetweenPromptUpdate {
+            cost_seen: true,
+            last_lifecycle_at: now_ms,
+            wake_until: prev_wake_until,
+        }),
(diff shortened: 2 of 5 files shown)


src/acp/acp_client.rs
```diff
@@ -4158,6 +4234,44 @@ async fn run_connection_task<W, R>(
...
+                        }
+                        if let Some(LifecycleSignal::WakeupPending { at }) = &wakeup_signal {
+                            between_prompt_active.store(true, Ordering::Relaxed);
+                            between_prompt_cost_seen.store(false, Ordering::Relaxed);
+                            last_lifecycle_at.store(
+                                chrono::Utc::now().timestamp_millis(),
+                                Ordering::Relaxed,
+                            );
+                            let deadline = at.timestamp_millis()
+                                + OFF_PROTOCOL_WORK_GRACE_FLOOR.as_millis() as i64;
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2326/commits/c98400fa: TerminalUsage now refreshes `last_lifecycle_at` (via the extracted pure `between_prompt_signal_update` helper), so the 3s fast grace measures from the cost marker, not a stale earlier progress event. Added state-path regression tests, including one asserting a stale progress event plus a cost marker cannot fire before `BETWEEN_PROMPT_IDLE_GRACE`.

docs/structured-view/troubleshooting.md
```diff
@@ -242,6 +242,12 @@ container (rather than bind-mounting from the host).
   marker, which made the daemon refuse to bring the worker back after it next
   exited; a follow-up message would then sit unsent. A worker coming online now
   clears that marker, so this no longer strands a queued message.
+- An agent that armed a monitor or a scheduled wake, finished the turn, then was
+  woken later by that background task used to keep the session marked "running"
+  forever: the resumed turn ran with no driving prompt, so nothing ended it. The
+  daemon now ends such a turn cleanly once it goes idle (a few seconds after
+  the agent's last output, longer while a fresh wake is still pending), so the
+  session returns to idle on its own without `aoe acp restart`.
```
njbrake: This is extra tech information not necessary for docs

njbrake:
```suggestion
```

njbrake (approved):

njbrake:
@Seluj78 good to merge after merge conflicts are resolved!

PR #2339, opened 2026-06-22
feat(acp): require opencode 1.16.0 and humanize permission approval titles

### LINKED ISSUE #1907: Structured view: approval card for opencode external_directory permission shows no path or command context

## Problem

When opencode runs in the web dashboard structured view and tries to access files outside the workspace directory, it sends an ACP `session/request_permission` with title `external_directory`. The approval card shows just `external_directory` as the card title with no filepath, no tool command, and no other context. The user has no way to tell what file is being accessed or what operation is requested.

For example, when opencode runs `mkdir -p "/tmp/opencode"`, the approval card shows only the label "external_directory" with an empty args section. The actual command (`mkdir -p "/tmp/opencode"`) and the target path (`/tmp/opencode`) are visible in the preceding tool call update but are never surfaced in the approval card.

Session `8bde1fec44f6438c` (`~/.agent-of-empires/cockpit_events.db`):

Sequence 1197-1204 shows the full flow:

```
1197|ToolCallStarted   | bash                    | args_preview: {}
1198|ToolCallUpdated   | bash                    | command: mkdir -p "/tmp/opencode", workdir: /tmp
1199|ToolCallStarted   | external_directory      | args_preview: {}     ← permission request
1200|ApprovalRequested | external_directory      | args_preview: {}     ← card has no info
1201|ApprovalResolved  | Allow
1202-1203|ToolCallUpdated | bash | same command
1204|ToolCallCompleted  | bash
```

The `external_directory` permission request at seq 1199-1200 uses the same `tool_call_id` (`call_gUje0qyXANwLo0nLiLyS285p`) as the bash tool. The bash tool's `ToolCallUpdated` at seq 1198 carries `command: mkdir -p "/tmp/opencode"` and `workdir: /tmp`. But the permission request re-uses the same ID with a different title/kind and has empty `rawInput`.

Other `external_directory` permission requests in the same session include the filepath in args:

```
seq 909 | filepath: /Users/seluj78/.config/opencode, parentDir: /Users/seluj78/.config/opencode
seq 918 | filepath: /Users/seluj78/.opencode,             parentDir: /Users/seluj78
seq 1245| filepath: /tmp/issue-opencode-read-line-numbers.md, parentDir: /tmp
```

So the filepath is sometimes present and sometimes absent. When absent (seq 1200), the card renders with zero args content because the `ArgsView` component (`web/src/components/cockpit/ApprovalCard.tsx:235`) returns `null` for `entries.length === 0`.

## Root cause

Two independent factors combine to produce the empty card:

1. **opencode sends permission requests with empty `rawInput` for some permission types.** This is documented in opencode issue #14301 (closed). The permission request re-uses the underlying tool's `tool_call_id` but sets its own `title` (`external_directory`), `kind` (`other`), and an empty `rawInput: {}`. The actual command and path details live in the separate `tool_call_update` for the original tool.

2. **AoE has no cross-reference logic between permission requests and associated tool calls.** The approval card at `acp_client.rs:5159-5216` builds its display from `request.tool_call.fields.title` and `request.tool_call.fields.raw_input` only. It never looks at other tool calls sharing the same `tool_call_id` to enrich the display. When `raw_input` is empty or missing, the card has nothing to show (ApprovalCard.tsx:235 returns null for empty args).

### A. Cross-reference permission requests with tool call context

In `handle_permission_request` (or the event path preceding it), when the permission request shares a `tool_call_id` with a previously observed `ToolCallStarted`/`ToolCallUpdated`, extract the command/path/description from the original tool call and include it in the approval display. This could be done by:

- Stashing the last-seen tool call args per `tool_call_id` in `acp_client.rs`
- When building the `Approval`, merge the stashed context into `args_preview` or add a new field

### B. Improve approval card rendering for empty args

When `args_preview` is empty (`{}`) or null, the `ArgsView` should not render nothing. Show a fallback message like "(no details provided)" or display the tool name in a friendlier format (`external_directory` → "Access external file or directory"). This is the simplest change and improves all agents, not just opencode.

### C. Friendlier title display for permission identifiers

The card title currently shows the raw ACP `title` field verbatim, which is an internal identifier like `external_directory`. The UI could humanize known identifiers (snake_case to readable text) or, when the title matches the permission kind, show a generic "Permission required" label instead.

## Why now

- This is active daily UX friction for the primary structured-view workflow (opencode).
- The card gives the user zero information to make a decision, forcing reflexive "Allow" clicks or blind trust.
- Proposal B is a small, safe change that improves the experience for all ACP agents.
- Proposal A is more impactful for opencode specifically and addresses the core data gap.

- As a structured-view user, when opencode requests permission to access "/tmp/opencode", then the approval card shows the filepath and the originating tool command, not just `external_directory`.
  - Suggested test layer: `web/tests/live/` (approval card rendering with real permission events).

- As a structured-view user, when the agent sends a permission request with empty `rawInput`, then the approval card still shows useful context (tool name + derived description) instead of a blank card with no args.
  - Suggested test layer: Vitest + RTL for `ApprovalCard.tsx` with empty/missing args.

- As a structured-view user, when a permission request shares a `tool_call_id` with a recently updated tool call, then the approval card enriches its display with context from that tool call (path, command, description).
  - Suggested test layer: Rust test in `acp_client.rs` for the cross-reference logic, plus Vitest for the UI rendering.

## Open questions

- Should the cross-reference logic (Proposal A) live in Rust (server-side, enriched at event creation time) or in TypeScript (client-side, enriched at render time)? Server-side is more robust; client-side is faster to iterate.
- When a permission request's title is a known internal identifier (e.g., `external_directory`), should AoE map it to a human-readable label, or leave that to each agent?

## Out of scope

- Fixing opencode's permission request `rawInput` behavior (tracked upstream in anomalyco/opencode#14301).
- General approval card redesign for destructive vs benign actions (separate UX work).
- The `ToolCallStarted` for `external_directory` appearing as a separate card before the approval (related to Gemini work in #1713).

## Cross-references

- #1713 — Structured view: Gemini ACP permissioned tools show null args, lose cards, and collapse output to completed. Same empty/nil args problem, different agent. Proposal B in this issue overlaps with #1713's Proposal B.
- #1767 — Structured view tool-call card UX: collapsed approval lacks detail. Related approval card info-density gap.
- anomalyco/opencode#14301 — opencode upstream issue: premature tool_call_update clobbers permission dialog. Documents the pattern where permission requests have empty `rawInput`.

Seluj78:

## Description

Pre-1.16.0 opencode sent an empty `rawInput` on `external_directory` ACP permission requests, so the structured-view approval card showed only the bare `external_directory` identifier with no path and no command. The user had nothing to base an approve/deny decision on. Upstream fixed the payload in opencode 1.16.0 (PR #30567, tracked in anomalyco/opencode#30554).

Rather than build a client-side cross-reference shim that would shadow the upstream fix forever, this enforces a minimum opencode version, mirroring the existing `claude-agent-acp` floor: `OPENCODE_MIN_VERSION = "1.16.0"` in `src/acp/agent_compat.rs`, gated on the handshake's `agent_info.name` ("OpenCode", verified against opencode v1.16.0 `acp/service.ts`) and a semver `>= 1.16.0`, rejecting missing or unparseable version info. Below the floor, the structured view refuses the session and the compatibility screen reports installed vs required plus the curl update command (opencode is not npm-installable, so there is no dashboard auto-update).

It also humanizes known ACP permission identifiers on the approval card: `external_directory` now reads "External directory access" in both the visible title and the accessible name, with verbatim passthrough for unknown identifiers so a new upstream kind stays a debuggable signal rather than a mangled auto-title. This is agent-agnostic and improves the card for any adapter that sends a permission kind as the request title.

The fake ACP agent used by the live Playwright suite is shimmed under several binary names; since the gate keys off the spawned binary, the fake now reports an opencode-shaped handshake (name + a version at/above the floor) when it stands in for opencode, so the existing opencode live specs keep passing.

Closes #1907

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] With opencode older than 1.16.0 installed, open an opencode structured-view session: aoe refuses to start and the compatibility screen shows installed vs required 1.16.0 and the curl update command.
- [ ] Update opencode to 1.16.0 or newer, hit **Restart agent**, and reopen: the session starts normally.
- [ ] On opencode 1.16.0+, trigger an `external_directory` permission (e.g. an `mkdir -p "/tmp/opencode"`): the approval card title reads "External directory access" and the path/command are visible.
- [ ] Confirm an unknown permission identifier still renders its raw name on the card.

## Checklist

- [x] Documentation was updated where necessary (none needed; the compatibility screen reports the required version and install command dynamically)

## Test Coverage Analysis

Added a Vitest covering the approval-card title humanization (`web/src/components/acp/ApprovalCard.test.tsx`, already mapped by `acp.approval-card`), and updated the live fixtures (`fakeAcpAgent.mjs` + `aoeServe.ts`) so the opencode live specs (`acp.mode-picker`) survive the new floor. The Rust floor is covered by unit tests in `src/acp/agent_compat.rs` (below-floor / at-floor / above-floor / missing-info / name-mismatch); the below-floor and missing-info tests fail on the pre-fix tree and pass after.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (mirror the claude version floor instead of a client-side shim, include the title humanization), reviewed each commit, and ran the manual and automated test steps.

### DIFF
--- a/docs/structured-view/troubleshooting.md
+++ b/docs/structured-view/troubleshooting.md
@@ -65,21 +65,6 @@ Then run `claude login` if you haven't already. If an older version is pinned by
 an internal mirror, ship the required floor from the mirror or run the `@latest`
 install above before starting `aoe serve`.
 
-### opencode adapter is too old
-
-aoe requires opencode 1.16.0 or newer. Older releases sent empty permission
-details, so the approval card could not show the path or command being
-requested. As with claude, aoe refuses to start the session and reports the
-required version. Update opencode:
-
-```bash
-curl -fsSL https://opencode.ai/install | bash
-```
-
-opencode is not npm-installable, so the dashboard cannot update it for you; run
-the command above in a shell, then use **Restart agent** on the compatibility
-screen.
-
 ### Recovering a missing or out-of-date agent from the web dashboard
 
 When the structured view refuses a session because the agent is missing or too
--- a/docs/structured-view/troubleshooting.md
+++ b/docs/structured-view/troubleshooting.md
@@ -65,6 +65,21 @@ Then run `claude login` if you haven't already. If an older version is pinned by
 an internal mirror, ship the required floor from the mirror or run the `@latest`
 install above before starting `aoe serve`.
 
+### opencode adapter is too old
+
+aoe requires opencode 1.16.0 or newer. Older releases sent empty permission
+details, so the approval card could not show the path or command being
+requested. As with claude, aoe refuses to start the session and reports the
+required version. Update opencode:
+
+```bash
+curl -fsSL https://opencode.ai/install | bash
+```
+
+opencode is not npm-installable, so the dashboard cannot update it for you; run
+the command above in a shell, then use **Restart agent** on the compatibility
+screen.
+
 ### Recovering a missing or out-of-date agent from the web dashboard
 
 When the structured view refuses a session because the agent is missing or too
--- a/src/acp/agent_compat.rs
+++ b/src/acp/agent_compat.rs
@@ -7,12 +7,16 @@
 //! single hook to call after `initialize` succeeds, instead of scattering
 //! ad-hoc semver checks at every spawn site.
 //!
-//! Today only `ClaudeAgentAcp` carries a minimum version (see
+//! `ClaudeAgentAcp` carries a minimum version (see
 //! `CLAUDE_AGENT_ACP_MIN_VERSION`,
 //! required for `memory_recall` tool-call emission, native `cancelled`
 //! stop reason, force-cancel of a wedged `TaskOutput` block (upstream
 //! #680), the upstream #641 fix, the `fable` model, and several other
-//! behaviors aoe builds on). Other agents
+//! behaviors aoe builds on). `OpenCode` carries one too (see
+//! `OPENCODE_MIN_VERSION`, the release that stopped sending empty
+//! `rawInput` on `external_directory` permission requests so the approval
+//! card shows the path and command; AoE issue #1907, upstream #30567). The
+//! remaining agents
 //! get a permissive policy (protocol check only). Long-term aoe should
 //! prefer ACP capability flags over package-version gating; until upstream
 //! exposes those, package versions are the only precise contract.
@@ -43,6 +47,22 @@ pub const CLAUDE_AGENT_ACP_MIN_VERSION: &str = "0.49.0";
 fn claude_agent_acp_min_version() -> semver::Version {
     semver::Version::parse(CLAUDE_AGENT_ACP_MIN_VERSION)
         .expect("CLAUDE_AGENT_ACP_MIN_VERSION must be valid semver")
+}
+
+/// Single source of truth for the `opencode` minimum-version floor.
+///
+/// 1.16.0 is the first release that ships upstream #30567: pre-1.16
+/// opencode sent an empty `rawInput` on `external_directory` permission
+/// requests, so the structured-view approval card had no path or command
+/// to show and the user could not tell what was being approved (#1907).
+/// opencode installs via `curl | bash`, not npm, so there is no Dockerfile
+/// pin to keep in sync; the sandbox image's `curl` install always pulls a
+/// release at or above this floor.
+pub const OPENCODE_MIN_VERSION: &str = "1.16.0";
+
+/// Parsed form of [`OPENCODE_MIN_VERSION`].
+fn opencode_min_version() -> semver::Version {
+    semver::Version::parse(OPENCODE_MIN_VERSION).expect("OPENCODE_MIN_VERSION must be valid semver")
 }
 
 /// The adapter aoe is trying to launch. Drives which `CompatibilityPolicy`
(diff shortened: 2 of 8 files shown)


docs/structured-view/troubleshooting.md
```diff
@@ -65,6 +65,21 @@ Then run `claude login` if you haven't already. If an older version is pinned by
...
+required version. Update opencode:
+
+```bash
+curl -fsSL https://opencode.ai/install | bash
+```
+
+opencode is not npm-installable, so the dashboard cannot update it for you; run
+the command above in a shell, then use **Restart agent** on the compatibility
+screen.
+
```
njbrake: Why does this need to go here, wouldn't it be something we print in the web ui?

Seluj78: Great point. Clearly my skill trimming relating to docs wasn't big enough.

Seluj78: Good call, removed it. The compatibility screen already shows installed vs required and the install command dynamically when the session is refused, so the doc was redundant; it also matches the convention that user docs don't restate the floor number (the agent_compat comment notes the same for claude). Dropped in d30db027.

njbrake (approved):

Issue #2344, opened 2026-06-23
bug: verify-on-resume picks a *different* session's conversation when multiple Claude sessions share one cwd (regression from #2298)


lamdor:
## Summary

Since PR #2298 (the fix for #2291), resuming a stopped Claude session can land on a **different session's** conversation. The verify-on-resume rescan in `acquire_session_id` selects the "freshest live observation" Claude session id by mtime within `~/.claude/projects/<encoded-cwd>/`, but does not check that the chosen transcript belongs to the session being resumed. When several AOE sessions share the same working directory — so they all map to the same encoded-cwd projects dir — the "freshest" transcript is frequently the conversation I most recently touched in *another* session, and that id gets persisted onto the session being resumed.

#2291 was about a session failing to re-find *its own* post-`/clear` conversation. This is the inverse failure introduced by the same machinery: a session re-finds *someone else's* conversation.

## Environment

- aoe built from commit `3f533437ec0e67794f53da1a0ad4e646e7414222` (post-#2298; `aoe --version` still reports 1.11.1 as the tag lags the commit).
- macOS (darwin), `aoe serve` daemon mode (launchd), no attached TUI most of the time.
- Many concurrent `tool=claude` sessions, the majority sharing one `project_path` (a single monorepo dir) → all under the same `~/.claude/projects/<encoded-cwd>/`.

## Steps to reproduce

1. Run two or more AOE Claude sessions in the **same** working directory (so they share one encoded-cwd projects dir).
2. Use session A (conversation A active / most recently written transcript), then stop it.
3. Resume session B shortly after.
4. **Expected:** B resumes on conversation B.
5. **Actual:** B resumes on conversation A — the most-recently-written transcript in the shared projects dir.

The corruption window is "use/stop one session, then resume another in the same cwd within ~the same minute."

## Evidence from my daemon

The real conversation for my `tailscale-operator` session (`be30b8501d6d42f5`) is `7b7bbbdc…` (first user message: "add tailscale operator to flux…"). The hook path set it correctly, then verify-on-resume overwrote it with `2c68fba3…`, which is a *different* session's conversation ("add ability to disable support widget").

```
00:13:25  session.store: Session ID changed for be30b8501d6d42f5: 7b7bbbdc-…   ← hook sets correct id
00:30:28  session.store: Replacing stored session id with fresher live observation stale=0c6db66d-… fresh=7b7bbbdc-… tool=claude
[...]

Issue #2304, opened 2026-06-20
test(session): per-tool verify-on-resume tests for opencode, vibe, codex, gemini, pi, hermes


jerome-benoit:
Follow-up from #2298 (closes #2291).

The PR's verify-on-resume in `Instance::acquire_session_id` Default branch dispatches via `try_retroactive_capture` to per-tool capture functions. For Claude there is a direct integration test (`supersedes_stale_claude_sid_after_clear` in `src/session/instance.rs::tests::resume_fallback::verify_on_resume`).

The other six live-tracked agents (`opencode`, `vibe`, `codex`, `gemini`, `pi`, `hermes`) inherit the bascule through the same `try_retroactive_capture` arm but have no direct `acquire_session_id` test. Each tool's individual capture function is unit-tested in `src/session/capture.rs`, but a regression in the dispatch arm itself (e.g. accidental removal of the `opencode` arm) would not be caught by this PR's test surface.

### Proposed tests

One bascule test per tool in `src/session/instance.rs::tests::resume_fallback::verify_on_resume`, mirroring `supersedes_stale_claude_sid_after_clear`. The fixture pattern differs per tool:

| tool | fixture |
|---|---|
| opencode | seed `OPENCODE_DB` SQLite with two `session` rows; older `time_updated` = stored, newer = fresh |
| vibe | `VIBE_HOME` → `logs/session/<dir>/meta.json` × 2 with mtime ordering |
| codex | `CODEX_HOME` → `sessions/rollout-*-<UUID>.jsonl` × 2 with mtime ordering |
| gemini | `GEMINI_CLI_HOME` → `tmp/<sha256(cwd)>/chats/session-*.json` × 2 |
| pi | `PI_CODING_AGENT_DIR` → `sessions/<encoded>/<TS>_<UUID>.jsonl` × 2 |
| hermes | `HERMES_HOME` → `state.db` SQLite with two rows, older `started_at` = stored, newer = fresh |

Each test should reuse the `StorageHomeGuard`/`ClaudeHomeGuard` RAII pattern (or the existing `VibeHomeGuard`/`CodexHomeGuard`/`GeminiHomeGuard` already in `capture.rs`) for env isolation.

Estimate: 2-3 h (mostly fixture authoring per tool).

### Why P2-class

- Each tool's capture function is independently unit-tested.
- The dispatch in `try_retroactive_capture` is mechanical 6-line per-tool match arm.
[...]

Issue #2291, opened 2026-06-19
bug: resuming a session after /clear restores the pre-clear conversation instead of the latest one


karljuhlthefool:
## Bug

When a Claude Code session is resumed after the pane has died or been killed, AOE resumes to the conversation snapshot **before** the last `/clear` rather than the most recent one.

## Steps to reproduce

1. Start a session and have a conversation
2. Run `/clear` in the session (this creates a new conversation UUID / new JSONL file while the pane stays alive)
3. Continue working in the session after the clear
4. Kill or let the pane die (tmux pane crash, system sleep, `aoe stop`, etc.)
5. Resume the session via AOE (sidebar tap, `aoe resume`, etc.)

**Expected:** session resumes on the post-clear conversation — the one that was active when the pane died

**Actual:** session resumes on the pre-clear conversation — the history that `/clear` was supposed to leave behind

## Workaround

Resume the session to get it running, then manually run `--resume {post-clear-uuid}` with the correct UUID to jump to the right conversation. Annoying and requires knowing the UUID.

## Root cause (likely)

AOE stores the session's `agent_session_id` in `sessions.json` at the time the session is first created (or last observed). When `/clear` is used inside a live session, Claude Code creates a **new JSONL file with a new UUID** — the running pane's poller detects this and updates its in-memory state, but the persisted `agent_session_id` in `sessions.json` may not be updated before the pane dies.

When AOE later resumes the session, it reads the stale `agent_session_id` from storage and passes `--resume {old-uuid}` to Claude, landing on the wrong conversation.

The fix is likely: on resume, AOE should scan `~/.claude/projects/{encoded-path}/` for the **most recently modified JSONL** rather than (or in addition to) trusting the stored `agent_session_id`. This is the same scan logic already used during the session capture polling flow — it just needs to be applied at resume time too.

## Why this matters

`/clear` is a common operation — users clear context to give Claude a clean slate mid-session without starting a whole new AOE session. The expectation is that the session "is" the post-clear state. Having resume silently drop you back to the old context is confusing and can cause unintended work to be repeated or decisions to be re-litigated on stale context.

njbrake:
Oooh maybe @jerome-benoit would have some thoughts about this one? He is the mastermind behind the resume functionality of aoe.

PR #2345, opened 2026-06-23
fix(session): prefer authoritative hook sidecar over mtime scan on Claude resume

lamdor:
## Description

Fixes the wrong-conversation half of #2344: resuming a stopped Claude session can land on a *different* session's conversation when several AoE sessions share one working directory.

### Root cause

The verify-on-resume mtime scan added in #2298 (the fix for #2291) lives in `capture_freshest_session_id` (`src/session/instance.rs`). It calls `try_retroactive_capture` → `capture_claude_session_id` → `scan_claude_project_dir`, which selects the most-recent `*.jsonl` in `~/.claude/projects/<encoded-cwd>/`.

That directory is keyed by cwd, not by session. When multiple AoE Claude sessions run in the same `project_path` (a single monorepo dir is a common layout for autonomous/dispatched sessions), they all map to the same encoded-cwd dir. The "freshest" jsonl is then frequently a *peer* session's conversation — the one most recently touched — not this instance's.

On resume, `acquire_session_id`'s `Some(stored)` branch calls `capture_freshest_session_id`; when it returns a peer's id, the stored id is overwritten and the launch runs `claude --resume <peer-uuid>`. Observed in the wild as a stored `agent_session_id` being replaced by an unrelated session's conversation, logged as `Replacing stored session id with fresher live observation stale=… fresh=… tool=claude`.

The exclusion set (`build_exclusion_set`) is meant to prevent exactly this — it reads peers' `AOE_CAPTURED_SESSION_ID` from tmux env — but a *stopped* peer has no live tmux session, so its id isn't in the set, and its jsonl is fair game for the scan. The existing doc note on `capture_freshest_session_id` framed this as a "co-located non-AoE Claude peer" edge case; the dominant trigger is actually AoE peers sharing a cwd.

### Fix

Consult the per-instance hook sidecar (`read_hook_session_id`, written by the SessionStart/UserPromptSubmit hooks under `/tmp/aoe-hooks-<euid>/<instance_id>/session_id`) before the mtime scan, mirroring the ordering already used by `claude_poll_fn`. The sidecar is keyed by instance id, so it can never name a peer instance's conversation. When present and fresh it authoritatively decides whether an override is warranted:

- sidecar == stored → no override (steady state)
- sidecar != stored → override to the sidecar value (legitimate fresh id for *this* instance, e.g. post-`/clear`)
- sidecar in this instance's `retroactive_capture_excludes` → no override (cascade re-poison guard)

The mtime scan remains the fallback when no fresh sidecar exists (daemon-only mode where the poller never drained, or a session resumed after the 5-minute sidecar window), preserving the #2291 fix. `supersedes_stale_claude_sid_after_clear` continues to pass.

### Scope / known remaining gap

This closes the case where a fresh sidecar disagrees with the mtime guess — the common interactive trigger. It does **not** fully cover the case where the sidecar is also stale (>5 min) *and* a stopped peer's jsonl is the freshest in the shared dir: with no authoritative source, the mtime fallback can still mis-select. A complete fix would widen the exclusion set to include session ids owned by stopped/idle peers (e.g. from `sessions.json`), not just live tmux env. That's a larger change; happy to follow up in a separate PR if you'd like that direction.

The secondary symptom in #2344 (the `AOE_CAPTURED_SESSION_ID` tmux write-back targeting an `aoe_` `Session` name when the live pane is an `aoe_term_` `TerminalSession`, so every write fails with `no such session`) is left for a separate change — it's adjacent (it's why stopped peers fall out of the exclusion set) but mechanically distinct.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is exercised by deterministic unit tests at the `acquire_session_id` boundary (`tests::resume_fallback::verify_on_resume`): `sidecar_wins_over_fresher_peer_jsonl` (peer jsonl strictly fresher on disk, sidecar names this instance's conversation → no override) and `mtime_fallback_applies_without_sidecar` (no sidecar → #2291 bascule still applies). A full e2e for multi-session shared-cwd resume would need a real `claude` binary plus several live tmux sessions racing on one project dir, which the existing `tests/e2e/` harness doesn't model; the unit tests pin the exact decision point that regressed.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Used to investigate the issue (tracing the resume path and the #2298 interaction against `main`) and to draft the patch and tests. All file/line references and the root-cause reasoning were verified against the code, and the fix was built and tested locally (`cargo fmt --check`, `cargo clippy --lib --tests`, full `session::instance` and `session::capture` suites green).

Related: #2344, #2291, #2298, #2304

## High-Level Summary

This PR fixes a resume bug for stopped Claude sessions in Agent-of-Empires where resuming could jump into a different (peer) session’s conversation when multiple sessions share the same working directory.

**What changed**
- Updated `Instance::capture_freshest_session_id` to use an **instance-specific sidecar value** (when available) as the primary source of truth for Claude session selection.
- Added a small helper (`override_if_distinct`) to ensure the system only overrides the stored session when the sidecar’s session ID is meaningfully different.
- Kept the existing **freshness/mtime-based fallback** behavior when the sidecar is missing/expired, so prior safeguards remain intact.
- Extended the behavior to cover **sandboxed Claude** flows consistently.

**Why this helps**
- Prevents incorrectly resuming the “most recently modified” peer conversation in shared directories.
- Preserves prior fixes for cases where sidecar data isn’t available.

**Tests added/updated**
- `sidecar_wins_over_fresher_peer_jsonl`: confirms the sidecar overrides a fresher peer JSONL.
- `sidecar_consulted_for_sandboxed_claude`: confirms the same sidecar priority for sandboxed Claude sessions.
- `mtime_fallback_applies_without_sidecar`: confirms the fallback selection still works when the sidecar is absent.

**Known remaining limitation**
- If the sidecar is stale (>5 minutes) *and* a stopped peer’s JSONL is the freshest in the shared directory, the mtime fallback can still misselect. A full fix would require widening the exclusion set to include stopped/idle peers, deferred to a future PR.

## Technical Details (optional)

- For **Claude**, `capture_freshest_session_id` now prefers `read_hook_session_id(self.id)` (instance-keyed), and uses the fallback freshness scan only when the sidecar is absent.
- The override decision is centralized so overrides only occur when the sidecar-derived ID differs from the currently stored `agent_session_id`.
- For **non-Claude**, the previous freshness behavior remains unchanged.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -783,6 +783,16 @@ pub(crate) fn persist_session_to_storage(
     }
 }
 
+/// Emit `fresh` only when it differs from the stored session id, the
+/// "override only when distinct" contract shared by both branches of
+/// `capture_freshest_session_id` (sidecar and mtime fallback).
+fn override_if_distinct(stored: Option<&str>, fresh: String) -> Option<String> {
+    match stored {
+        Some(known) if known == fresh => None,
+        _ => Some(fresh),
+    }
+}
+
 /// Publish a captured session ID to the tmux environment only.
 ///
 /// Background threads (poller on_change) call this so that
@@ -1601,32 +1611,38 @@ impl Instance {
     /// For Claude the authoritative per-instance sidecar
     /// (`/tmp/aoe-hooks-<euid>/<instance_id>/session_id`, written by the
     /// SessionStart / UserPromptSubmit hooks) is consulted first. It is keyed
-    /// by instance id, so it can never name a peer instance's conversation —
+    /// by instance id, so it can never name a peer instance's conversation,
     /// unlike the mtime disk scan, which picks the most-recent jsonl in the
     /// shared `~/.claude/projects/<encoded-cwd>/` dir and so can select a
     /// co-located peer's session when several AoE sessions share one cwd
     /// (#2344). The mtime scan is only used as a fallback when no fresh
     /// sidecar exists (e.g. an old session resumed after the 5-minute
     /// sidecar window), matching the ordering already used by
-    /// `claude_poll_fn`.
+    /// `claude_poll_fn`. Sandboxed Claude is included: its `SessionStart`
+    /// hook writes through the `/tmp/aoe-hooks/<id>` bind-mount onto the
+    /// host path, so `read_hook_session_id` reads it the same way, and the
+    /// mtime fallback below still routes through the container-aware branch
+    /// of `try_retroactive_capture`.
+    ///
+    /// Two deliberate divergences from `claude_poll_fn`, both correct for the
+    /// resume context: (1) an excluded sidecar id returns `None` here rather
+    /// than falling through to the mtime scan, since falling through is what
+    /// re-opens #2344; (2) this reader and `claude_poll_fn` read the same
+    /// sidecar without a shared snapshot, so a hook rotation between the two
+    /// reads can briefly surface different UUIDs, benign under the existing
+    /// eventual-consistency capture model.
     pub(crate) fn capture_freshest_session_id(&self) -> Option<String> {
-        if self.tool == "claude" && !self.is_sandboxed() {
+        if self.tool == "claude" {
             if let Some(authoritative) = crate::hooks::read_hook_session_id(&self.id) {
                 if self.retroactive_capture_excludes.contains(&authoritative) {
                     return None;
                 }
-                return match self.agent_session_id.as_deref() {
-                    Some(known) if known == authoritative => None,
-                    _ => Some(authoritative),
-                };
+                return override_if_distinct(self.agent_session_id.as_deref(), authoritative);
             }
         }
 
         let live = self.try_retroactive_capture()?;
-        match self.agent_session_id.as_deref() {
-            Some(known) if known == live => None,
-            _ => Some(live),
-        }
+        override_if_distinct(self.agent_session_id.as_deref(), live)
     }
 
     fn apply_session_flags(&mut self, cmd: &mut String, context: &str) -> bool {
@@ -7051,6 +7067,65 @@ mod tests {
 
                 // The authoritative sidecar matches the stored sid, so no
                 // override happens despite the peer's fresher jsonl.
+                assert_eq!(sid.as_deref(), Some(mine));
+                assert!(is_existing);
+                assert_eq!(inst.agent_session_id.as_deref(), Some(mine));
+            }
+
+            // #2344 follow-up: a sandboxed Claude session must also consult the
+            // sidecar. Its SessionStart hook writes through the
+            // `/tmp/aoe-hooks/<id>` bind-mount onto the host path, so
+            // `read_hook_session_id` reads it the same way a host session's is
+            // read. Without the sidecar short-circuit the sandbox-aware mtime
+            // branch would pick a peer's fresher jsonl in the shared cwd.
+            #[test]
+            #[serial]
+            fn sidecar_consulted_for_sandboxed_claude() {
+                let temp = tempdir().unwrap();
+                let _guard = ClaudeHomeGuard::set(&temp);
+
+                let project_path = "/tmp/aoe-test-2344-sandbox";
+                let claude_dir = temp
+                    .path()
+                    .join(".claude")
+                    .join("projects")
+                    .join(encode_claude_project_path(project_path));
+                fs::create_dir_all(&claude_dir).unwrap();
+
+                let mine = "eeeeeeee-5555-4555-8555-eeeeeeeeeeee";
+                let peer = "ffffffff-6666-4666-8666-ffffffffffff";
+                let now = SystemTime::now();
+                write_jsonl_with_mtime(
+                    &claude_dir.join(format!("{mine}.jsonl")),
+                    now - Duration::from_secs(120),
+                );
+                write_jsonl_with_mtime(
+                    &claude_dir.join(format!("{peer}.jsonl")),
+                    now - Duration::from_secs(5),
+                );
+
+                let mut inst = Instance::new("verify-2344-sandbox", project_path);
+                inst.tool = "claude".to_string();
+                inst.agent_session_id = Some(mine.to_string());
+                inst.resume_intent = ResumeIntent::Default;
+                inst.sandbox_info = Some(crate::session::SandboxInfo {
+                    enabled: true,
+                    container_id: None,
+                    image: "test-image".to_string(),
+                    container_name: "verify-2344-sandbox".to_string(),
+                    extra_env: None,
+                    custom_instruction: None,
+                    before_start_env: Vec::new(),
+                });
+                assert!(inst.is_sandboxed());
+
+                let dir = super::write_sidecar(&inst.id, mine);
+                let (sid, is_existing) = inst.acquire_session_id();
+                std::fs::remove_dir_all(&dir).ok();
+
+                // Sidecar (host-readable) names this instance's conversation, so
+                // the peer's fresher jsonl does not win even though sandbox would
+                // otherwise route through the container-aware mtime branch.
                 assert_eq!(sid.as_deref(), Some(mine));
                 assert!(is_existing);
                 assert_eq!(inst.agent_session_id.as_deref(), Some(mine));
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -7042,10 +7042,15 @@ mod tests {
                     .join(encode_claude_project_path(project_path));
                 fs::create_dir_all(&claude_dir).unwrap();
 
-                // This instance's real conversation (named by its sidecar) and a
-                // peer's conversation that happens to be the freshest on disk.
+                // `mine` is this instance's real conversation (named by its
+                // sidecar). `peer` is a co-located peer's conversation that is
+                // strictly freshest on disk. `stored` is a stale id distinct
+                // from `mine`, so asserting `sid == mine` proves the sidecar
+                // actively overrode the stored value rather than the stored
+                // value passing through unchanged.
                 let mine = "aaaaaaaa-1111-4111-8111-aaaaaaaaaaaa";
                 let peer = "bbbbbbbb-2222-4222-8222-bbbbbbbbbbbb";
+                let stored = "cccccccc-3333-4333-8333-cccccccccccc";
                 let now = SystemTime::now();
                 write_jsonl_with_mtime(
                     &claude_dir.join(format!("{mine}.jsonl")),
@@ -7058,15 +7063,15 @@ mod tests {
 
                 let mut inst = Instance::new("verify-2344-shared-cwd", project_path);
                 inst.tool = "claude".to_string();
-                inst.agent_session_id = Some(mine.to_string());
+                inst.agent_session_id = Some(stored.to_string());
                 inst.resume_intent = ResumeIntent::Default;
 
                 let dir = super::write_sidecar(&inst.id, mine);
                 let (sid, is_existing) = inst.acquire_session_id();
                 std::fs::remove_dir_all(&dir).ok();
 
-                // The authoritative sidecar matches the stored sid, so no
-                // override happens despite the peer's fresher jsonl.
+                // The authoritative sidecar overrides the stale stored sid;
+                // the peer's fresher jsonl never wins.
                 assert_eq!(sid.as_deref(), Some(mine));
                 assert!(is_existing);
                 assert_eq!(inst.agent_session_id.as_deref(), Some(mine));
@@ -7092,8 +7097,11 @@ mod tests {
                     .join(encode_claude_project_path(project_path));
                 fs::create_dir_all(&claude_dir).unwrap();
 
+                // `stored` is distinct from the sidecar `mine`, so the assertion
+                // proves the sidecar actively overrode the stale stored value.
                 let mine = "eeeeeeee-5555-4555-8555-eeeeeeeeeeee";
                 let peer = "ffffffff-6666-4666-8666-ffffffffffff";
+                let stored = "dddddddd-7777-4777-8777-dddddddddddd";
                 let now = SystemTime::now();
                 write_jsonl_with_mtime(
                     &claude_dir.join(format!("{mine}.jsonl")),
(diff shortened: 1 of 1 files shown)


njbrake:
Thanks for the PR 🙏 : @jerome-benoit  would you have time to take a look at this?

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
+    /// by instance id, so it can never name a peer instance's conversation —
+    /// unlike the mtime disk scan, which picks the most-recent jsonl in the
+    /// shared `~/.claude/projects/<encoded-cwd>/` dir and so can select a
+    /// co-located peer's session when several AoE sessions share one cwd
+    /// (#2344). The mtime scan is only used as a fallback when no fresh
+    /// sidecar exists (e.g. an old session resumed after the 5-minute
+    /// sidecar window), matching the ordering already used by
+    /// `claude_poll_fn`.
     pub(crate) fn capture_freshest_session_id(&self) -> Option<String> {
+        if self.tool == "claude" && !self.is_sandboxed() {
```
jerome-benoit: The `!self.is_sandboxed()` guard skips the sidecar path for sandboxed Claude, but the host bind-mount at `/tmp/aoe-hooks-<euid>/<id> -> /tmp/aoe-hooks/<id>` (`src/hooks/mod.rs:267-274` against `main`) means a sandboxed Claude's `SessionStart` hook writes a sidecar that is readable from the host through the same `read_hook_session_id(&self.id)` call. Sandboxed Claude sessions sharing one cwd therefore remain exposed to the #2344 peer-jsonl selection that the change fixes for host-mode sessions.

Two options:

1. Drop `!self.is_sandboxed()`. The sandbox session-id hook is installed at `src/hooks/mod.rs:762-770` (against `main`) and the host-side `read_hook_session_id` is path-agnostic. The fallback `try_retroactive_capture` already dispatches to the sandbox-aware branch at `src/session/instance.rs:1500-1506`, so the mtime fallback stays intact.
2. Keep the guard and document why sandbox is intentionally excluded (mirrors `claude_poll_fn_sandboxed` at `src/session/capture.rs:371-402`, which lacks the sidecar short-circuit). If sandbox parity is a follow-up, link the tracking issue.

Option 1 closes the bug more completely; option 2 makes the asymmetry intentional.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
+    /// (`/tmp/aoe-hooks-<euid>/<instance_id>/session_id`, written by the
+    /// SessionStart / UserPromptSubmit hooks) is consulted first. It is keyed
+    /// by instance id, so it can never name a peer instance's conversation —
+    /// unlike the mtime disk scan, which picks the most-recent jsonl in the
+    /// shared `~/.claude/projects/<encoded-cwd>/` dir and so can select a
+    /// co-located peer's session when several AoE sessions share one cwd
+    /// (#2344). The mtime scan is only used as a fallback when no fresh
+    /// sidecar exists (e.g. an old session resumed after the 5-minute
+    /// sidecar window), matching the ordering already used by
+    /// `claude_poll_fn`.
```
jerome-benoit: Retracting: `rg` shows 4 linkified vs 254 plain-backtick rustdoc references in `src/session/instance.rs` (1.6% vs 98.4%). Plain backticks are the dominant convention in this file; the new doc follows it correctly. No change needed.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
     /// contract (mtime, SQLite ordering, exclusion set, host/container)
     /// stays encapsulated in each tool's existing capture function.
     ///
-    /// Known limitation: when a non-AoE Claude session writes to the same
-    /// `project_path` (manual `claude` invocation outside the pane), its sid
-    /// is not in the tmux-env exclusion set and a fresher mtime there can
-    /// supersede the stored sid on resume. Co-located non-AoE peers are an
-    /// uncommon layout; the steady-state case (no peer) is unaffected.
+    /// For Claude the authoritative per-instance sidecar
```
jerome-benoit, line 1611: Retracting: "authoritative" matches the established override-semantic precedent at `src/session/instance.rs:3690` and `:3855` against `main` ("Resolution order: the passwd entry (the authoritative login shell)... then the container's `$SHELL`, then bash"). Same shape as the sidecar here: one source overrides several fallbacks. The PR site short-circuits the mtime scan unconditionally when present and fresh (`src/session/instance.rs:1614-1622`); dropping the word would lose the override-vs-fallback relationship the doc establishes. No change needed.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
+    /// (#2344). The mtime scan is only used as a fallback when no fresh
+    /// sidecar exists (e.g. an old session resumed after the 5-minute
+    /// sidecar window), matching the ordering already used by
+    /// `claude_poll_fn`.
     pub(crate) fn capture_freshest_session_id(&self) -> Option<String> {
+        if self.tool == "claude" && !self.is_sandboxed() {
+            if let Some(authoritative) = crate::hooks::read_hook_session_id(&self.id) {
+                if self.retroactive_capture_excludes.contains(&authoritative) {
+                    return None;
+                }
```
jerome-benoit, line 1639: The `retroactive_capture_excludes` guard mirrors `claude_poll_fn`'s `extra_excludes` check at `src/session/capture.rs:248`; same set, same purpose. One semantic asymmetry worth a doc note: `claude_poll_fn` falls through to the mtime scan when the sidecar matches an excluded sid, while this branch returns `None` outright. The divergent behavior is correct for the resume context (the sidecar guard is exactly what closes #2344, so falling through to mtime would re-open it), but the asymmetry is non-obvious. A one-liner in the doc would forestall the next "why do these not match" investigation.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
+                return match self.agent_session_id.as_deref() {
+                    Some(known) if known == authoritative => None,
+                    _ => Some(authoritative),
+                };
+            }
+        }
+
         let live = self.try_retroactive_capture()?;
         match self.agent_session_id.as_deref() {
             Some(known) if known == live => None,
```
jerome-benoit: The pattern `match self.agent_session_id.as_deref() { Some(known) if known == X => None, _ => Some(X) }` appears verbatim at lines 1618-1621 (new sidecar branch) and at 1626 onwards (existing mtime branch) with identical semantics. A small helper centralizes the "emit only when distinct from stored" contract:

```rust
fn override_if_distinct(stored: Option<&str>, fresh: String) -> Option<String> {
    match stored {
        Some(known) if known == fresh => None,
        _ => Some(fresh),
    }
}
```

Both branches then read as one-liners. Not blocking; small clarity win.

jerome-benoit (commented):
Read through head `36d9402c`. The diagnosis is clear and the fix is well-scoped: consulting the per-instance sidecar before the mtime scan mirrors `claude_poll_fn`'s ordering and closes the #2344 peer-jsonl selection for the common interactive case. The two new tests pin both the override-by-sidecar and the mtime-fallback paths at the right boundary.

Three non-blocking notes inline, in rough priority:

1. The `!self.is_sandboxed()` guard at line 1613 leaves sandboxed Claude sessions exposed to the same #2344 bug even though the host-side bind-mount makes the sidecar readable. Worth either dropping the guard or documenting the intentional asymmetry.
2. The two `match self.agent_session_id.as_deref() { ... }` blocks at lines 1618-1621 and 1626 onwards are identical; a small `override_if_distinct` helper would tighten the function.
3. The `retroactive_capture_excludes` short-circuit diverges from `claude_poll_fn`'s fall-through behavior on excluded ids; worth a one-liner in the doc to forestall a future "why do these not match" pass.

Two cross-checks that came out clean: the truncated-UUID race is blocked by `read_hook_session_id`'s strict `Uuid::parse_str` at `src/hooks/status_file.rs:78`, and the `supersedes_stale_claude_sid_after_clear` regression test stays correct because its setup writes no sidecar, so the new branch is skipped and the mtime fallback runs unchanged.

Two earlier inline comments (on the rustdoc cross-reference style at line 1611 and on the "authoritative" terminology at line 1601) have been retracted in their respective threads after empirical re-check; the doc as written follows the file's dominant conventions on both fronts.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
+    /// unlike the mtime disk scan, which picks the most-recent jsonl in the
+    /// shared `~/.claude/projects/<encoded-cwd>/` dir and so can select a
+    /// co-located peer's session when several AoE sessions share one cwd
+    /// (#2344). The mtime scan is only used as a fallback when no fresh
+    /// sidecar exists (e.g. an old session resumed after the 5-minute
+    /// sidecar window), matching the ordering already used by
+    /// `claude_poll_fn`.
     pub(crate) fn capture_freshest_session_id(&self) -> Option<String> {
+        if self.tool == "claude" && !self.is_sandboxed() {
+            if let Some(authoritative) = crate::hooks::read_hook_session_id(&self.id) {
```
jerome-benoit, line 1636: Two readers of the same sidecar without a shared snapshot: `capture_freshest_session_id` here and `claude_poll_fn` at `src/session/capture.rs:247`. Across a hook rotation between the two reads they can briefly observe different UUIDs. Likely benign under the existing eventual-consistency capture model, but worth a sentence in the same doc paragraph as the exclusion-set divergence.

src/session/instance.rs
```diff
@@ -1598,12 +1598,30 @@ impl Instance {
...
-    /// Known limitation: when a non-AoE Claude session writes to the same
-    /// `project_path` (manual `claude` invocation outside the pane), its sid
-    /// is not in the tmux-env exclusion set and a fresher mtime there can
-    /// supersede the stored sid on resume. Co-located non-AoE peers are an
-    /// uncommon layout; the steady-state case (no peer) is unaffected.
+    /// For Claude the authoritative per-instance sidecar
+    /// (`/tmp/aoe-hooks-<euid>/<instance_id>/session_id`, written by the
+    /// SessionStart / UserPromptSubmit hooks) is consulted first. It is keyed
+    /// by instance id, so it can never name a peer instance's conversation —
+    /// unlike the mtime disk scan, which picks the most-recent jsonl in the
```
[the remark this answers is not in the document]
lamdor: Fixed in c8066a3d — replaced the em dash with a comma. It was the only em dash in the file; the CLAUDE.md style rule is clear.

src/session/instance.rs
lamdor: Went with option 1. Verified the bind-mount (src/hooks/mod.rs:216-217: sandbox writes /tmp/aoe-hooks -> host /tmp/aoe-hooks-<euid>/<id>) and that read_hook_session_id is path-agnostic (instance-id keyed, no container branch), so dropping `!self.is_sandboxed()` is safe and closes the bug for sandboxed sessions too. The mtime fallback still routes through the container-aware try_retroactive_capture branch. Added sidecar_consulted_for_sandboxed_claude. Fixed in c8066a3d.

src/session/instance.rs
lamdor: Extracted override_if_distinct(stored, fresh); both branches now call it. c8066a3d.

src/session/instance.rs
lamdor: Added a doc note covering this: the excluded-sid short-circuit returns None rather than falling through to mtime, because falling through is what re-opens #2344. c8066a3d.

src/session/instance.rs
lamdor: Folded into the same doc paragraph — noted the unsnapshotted two-reader sidecar race (this reader vs claude_poll_fn) as benign under the eventual-consistency capture model. c8066a3d.

src/session/instance.rs
```diff
@@ -6986,6 +7020,155 @@ mod tests {
...
+                let dir = super::write_sidecar(&inst.id, mine);
+                let (sid, is_existing) = inst.acquire_session_id();
+                std::fs::remove_dir_all(&dir).ok();
+
+                // Sidecar (host-readable) names this instance's conversation, so
+                // the peer's fresher jsonl does not win even though sandbox would
+                // otherwise route through the container-aware mtime branch.
+                assert_eq!(sid.as_deref(), Some(mine));
+                assert!(is_existing);
+                assert_eq!(inst.agent_session_id.as_deref(), Some(mine));
```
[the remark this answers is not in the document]
jerome-benoit: Confirming, and extending: the host-mode twin `sidecar_wins_over_fresher_peer_jsonl` (around `src/session/instance.rs:7017` against `main`) has the same shape. It also sets `agent_session_id = mine` while the sidecar names `mine`, so the test would pass even if the sidecar branch were bypassed entirely and `acquire_session_id()` returned the stored value unchanged. Both tests should use a distinct stored UUID (e.g. `stored != mine`) so the assertion `sid == mine` requires the sidecar to actively override.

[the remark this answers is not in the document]
lamdor: Both tests now use a distinct stale `stored` UUID (`cccc…` host, `dddd…` sandbox) while the sidecar names `mine`, so the `sid == mine` assertion only holds if the sidecar actively overrode the stored value. Confirmed by mutation: disabling the sidecar branch fails both tests (they passed before the change). 676399b1.

njbrake (approved):

Approving. Real bug with a clear root cause, scoped to one file, and the fix is structurally sound: consulting the per-instance hook sidecar (keyed by instance id) before the mtime scan means it can never select a peer's conversation, while the mtime scan stays as the fallback so the #2291 fix is preserved.

Verified the load-bearing claim independently rather than from the description: that a sandboxed Claude's sidecar is readable on the host. Claude carries `hook_config`, which triggers the host->container hook-dir bind-mount (`/tmp/aoe-hooks-<euid>/<id>` -> `/tmp/aoe-hooks/<id>`), so the in-container SessionStart hook's `session_id` write lands on the host and `read_hook_session_id` reads it the same way as a host session's. Dropping the `!is_sandboxed()` guard is correct.

Tests pin the exact decision point and were mutation-checked (disabling the sidecar branch fails both override tests). The two documented divergences from `claude_poll_fn` (excluded id returns None instead of falling through; unsnapshotted two-reader race) are both correct for the resume context.

One non-blocking follow-up: the `claude_poll_fn_sandboxed` doc comment now reads as stale, since it claims the host poller cannot see the in-container hook without bind-mounting, but that bind-mount exists and is what this PR's resume path relies on. Worth a one-line update at some point.

The known gap (stale sidecar plus a stopped peer's jsonl being freshest) is honestly disclosed and reasonably deferred.

Issue #2095, opened 2026-06-10
Plugins 07: Tier 1 host: worker protocol, capability-gated API, NoSandbox


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Part of #268; design in `docs/development/internals/plugin-system.md` (D1, D6, D8).

The runtime that actually executes plugin code. Nothing here shipped in the minimal core (#2311).

## Scope

- Plugin worker protocol over `src/process/`: a Tier 1 plugin is an executable speaking ndjson JSON-RPC on stdio or a unix socket, supervised like an ACP worker (registry record, respawn budget, concurrency cap). Any language. Builtins run via `aoe __plugin-worker`.
- Host API with capability-checking middleware: `events.subscribe { topics, after_seq }` / `events.publish`, `session.metadata.{get,set,cas}` with per-plugin namespace enforcement, `sessions.list`, and the v1 method list from the doc. Undeclared capability use is refused.
- `SandboxBackend` trait with `NoSandbox` as the only v1 backend. Honest model per D8: capability gating at the host API boundary stops cooperative plugins, not adversarial ones; OS-level isolation (restricted-env, landlock, sandbox-exec) comes later behind the same trait. Every install prompt states this verbatim.

Depends on #2092 (substrate lift) and #2093 (manifests + grants).

Issue #2094, opened 2026-06-10
Plugins 06: Tier 0 registries: settings schema, keybinds, CLI grafting, themes


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Part of #268; design in `docs/development/internals/plugin-system.md` (D3, D4, D5).

Consumes the manifest contribution sections defined in #2093. Nothing here shipped in the minimal core (#2311).

## Scope

Declarative (Tier 0) contributions, no plugin code execution:

- Runtime settings schema: `runtime_schema(core_schema(), plugin_registry)`; TUI and web render plugin sections generically (virtual `plugin:<id>` sections); server PATCH validation against the runtime schema. Default-override resolution: user > highest-priority enabled plugin default > owning manifest default > schema default, with `ResolvedSetting { value, source, candidates }` provenance, `aoe settings explain <key>`, and `GET /api/settings/resolved`. Includes core-setting default overrides via `setting_defaults` with the same priority + provenance chain.
- Hybrid `ActionId { Core(CoreActionId), Plugin(PluginActionId) }`, canonical strings externally (`core.session.archive`, `plugin.<id>.<action>`). Static core BINDINGS table chained with a runtime plugin table; merged resolver with source/priority/shadowing inspectability (core always shadows; conflicts via `aoe plugin info`).
- CLI grafting: keep clap derive for core; append manifest-declared plugin commands into the derived `Command` tree at runtime (top-level or nested paths); dispatch tries `Cli::from_arg_matches` first, falls through to the plugin command registry. Core wins name conflicts.
- Themes via manifest.

Depends on #2093 (manifest contribution sections).

PR #2368, opened 2026-06-24
feat(plugins): per-session plugin_meta + persisted plugin settings (#2091)

### LINKED ISSUE #2091: Plugins 01: per-session plugin_meta + persisted plugin settings

Part of #268; design in `docs/development/internals/plugin-system.md`.

The original all-in-one plan (#2090) was closed; the minimal plugin core landed instead in #2311. This issue now tracks only the storage groundwork that is left.

## Already shipped (#2311)

- Typed `Config.plugins` map persisting per-plugin **enabled state** in `config.toml` under `[plugins."<id>"]`. It was added additively (Config already drops unknown TOML keys), so **no migration was needed**; the v013 migration originally planned here is dropped.

## Remaining scope

- `Instance.plugin_meta: BTreeMap<PluginId, serde_json::Value>`: namespaced per-session plugin data. Additive serde on `Instance`; add a migration only if the on-disk session shape actually needs backfilling.
- Extend `PluginConfig` with `settings: toml::Table` so a plugin's settings persist on disk even while the plugin is disabled (hidden from the schema until #2094 renders them).

## Out of scope

- The `session.metadata.{get,set,cas}` host API (Tier 1 host, #2095).
- Any schema rendering of plugin settings (#2094).
- ExclusiveStateSet or any triage primitive (#2098).

Seluj78:
Closes #2091

## Description

Plugins 01 (#2091): the storage groundwork left after the minimal plugin core (#2311). Adds the two on-disk slots that later plugin APIs read/write, ahead of those APIs, so no migration is needed when they land.

- **`PluginConfig.settings`** (`src/session/config.rs`): an opaque `toml::Table` persisted as `[plugins."<id>".settings]` in `config.toml`. Kept schema-free on purpose so values survive on disk even while the plugin is disabled; the typed schema that validates/renders them lands with the Tier 0 settings registry (#2094). `enabled` serializes before `settings` so the value precedes the nested table (TOML requirement); an empty table is omitted.
- **`Instance.plugin_meta`** (`src/session/instance.rs`): `BTreeMap<String, serde_json::Value>` keyed by plugin id, persisted per session in `sessions.json`. Each plugin owns only its own slot; data for an uninstalled plugin is retained (cheap, reinstall restores it). The `session.meta.{get,set,cas}` host API over it lands with the Tier 1 host (#2095).

Both fields are additive (`#[serde(default, skip_serializing_if = ...)]`): absent in older on-disk rows, deserialize to empty, no data migration. Design doc (`docs/development/internals/plugin-system.md`) updated.

### Drive-by: fix a flaky test

While running the suite, `test_capture_claude_session_active_newer_wins_over_anchor` flaked under the parallel run ("No active Claude session found"). Root cause is a global env-var data race: `capture_claude_session_id` reads `CLAUDE_CONFIG_DIR` via `std::env::var`, and a concurrent `set_var`/`remove_var` in an unrelated test corrupts that read (libc `setenv`/`getenv` are not thread-safe together), so the var reads as unset and capture falls back to `~/.claude` and bails. The capture tests already use the crate-wide default `#[serial]` group, but 23 env-mutating tests in `logging`/`builder`/`environment`/`instance` had no serialization against it (logging used only a file-local `ENV_LOCK`, which does not coordinate across modules). They are now in the default `#[serial_test::serial]` group.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: storage-layer serde fields with no TUI/CLI/session-lifecycle behavior change

Added serde round-trip unit tests for both fields (round-trip, empty-omission, legacy-row backfill) in `config.rs` and `instance.rs`.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Idea, decisions, and scoping are mine; Claude handled investigation and drafting.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2368"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784916357&installation_id=139138837&pr_number=2368&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2368&signature=50dee3fb8ec9531af5d01e0150a5a0ddbe57eebd91255a471ddce786ab130812"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR lays the storage groundwork for the plugin system.

- Adds persistent per-plugin settings in `config.toml`, so plugin options can be saved even when a plugin is disabled.
- Adds persistent per-session plugin metadata in `sessions.json`, giving each plugin its own stored data slot.
- Keeps both additions backward-compatible, so older saved data still loads without migration.
- Updates the plugin system internals docs to explain the new storage model.
- Adds round-trip tests to cover saving, loading, empty-field omission, and legacy data handling.
- Fixes flaky tests by making environment-mutating tests run one at a time.

Benefits:
- Plugin data can now survive disable/reload cycles.
- Existing user data continues to work unchanged.
- The storage model is ready for later plugin features without requiring a schema migration.

### DIFF
--- a/docs/development/internals/plugin-system.md
+++ b/docs/development/internals/plugin-system.md
@@ -64,6 +64,29 @@ path, and reloads the registry. The three surfaces are thin twins over it:
 The one behavior wired to a plugin's state today: `aoe serve` refuses to start
 while `aoe.web` is disabled (`src/cli/serve.rs`).
 
+## Persisted plugin state (#2091)
+
+Two storage slots hold plugin data on disk ahead of the APIs that read and
+write them, so no migration is needed later when those APIs land:
+
+- **Per-plugin settings.** `PluginConfig.settings` (`src/session/config.rs`) is
+  an opaque `toml::Table` persisted as `[plugins."<id>".settings]` in
+  `config.toml`. It is kept schema-free on purpose: values survive on disk even
+  while the plugin is disabled, and the typed schema that validates and renders
+  them arrives with the Tier 0 settings registry (#2094). `enabled` serializes
+  before `settings` so the value precedes the nested table, as TOML requires;
+  an empty table is omitted.
+- **Per-session plugin data.** `Instance.plugin_meta`
+  (`src/session/instance.rs`) is a `BTreeMap<String, serde_json::Value>` keyed
+  by plugin id, persisted per session in `sessions.json`. Each plugin owns only
+  its own slot; data for an uninstalled plugin is retained (cheap, and
+  reinstalling restores it). The read/write/cas host API over it
+  (`session.meta.{get,set,cas}`) lands with the Tier 1 host (#2095).
+
+Both fields are additive (`#[serde(default, skip_serializing_if = ...)]`):
+absent in older on-disk rows, so they deserialize to empty and need no data
+migration.
+
 ## What comes next
 
 Each deferred piece returns as its own PR once the core is proven: the
--- a/docs/development/internals/plugin-system.md
+++ b/docs/development/internals/plugin-system.md
@@ -67,15 +67,16 @@ while `aoe.web` is disabled (`src/cli/serve.rs`).
 ## Persisted plugin state (#2091)
 
 Two storage slots hold plugin data on disk ahead of the APIs that read and
-write them, so no migration is needed later when those APIs land:
+write them, so the later API PRs (#2094, #2095) stay focused on behavior:
 
 - **Per-plugin settings.** `PluginConfig.settings` (`src/session/config.rs`) is
   an opaque `toml::Table` persisted as `[plugins."<id>".settings]` in
   `config.toml`. It is kept schema-free on purpose: values survive on disk even
   while the plugin is disabled, and the typed schema that validates and renders
-  them arrives with the Tier 0 settings registry (#2094). `enabled` serializes
-  before `settings` so the value precedes the nested table, as TOML requires;
-  an empty table is omitted.
+  them arrives with the Tier 0 settings registry (#2094). `enabled` is declared
+  before `settings` so the scalar reads above the nested table; the toml
+  serializer emits scalars before subtables regardless, so the order is for
+  readability. An empty table is omitted.
 - **Per-session plugin data.** `Instance.plugin_meta`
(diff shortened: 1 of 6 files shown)


njbrake (approved):

Approving. Clean, well-tested storage groundwork for #2091: `PluginConfig.settings` and `Instance.plugin_meta` as additive serde fields, with round-trip, empty-omission, and legacy-row tests, and CI green across the board.

I pushed two small cleanups (714005d7):

- Dropped a spurious `#[serial]` on `test_git_sanitize_branch_name_replaces_forbidden_chars`. That function is a pure string transform with no env access, so serializing its test protected against no race and only removed it from the parallel pool.
- Reworded the TOML-ordering note. I verified with a scratch reproduction that `toml 1.1` emits scalars before subtables on serialize regardless of struct field order, so the `enabled` / `settings` order is for readability, not a serialization requirement. Also softened the doc wording that implied landing the slots early avoids a migration: additive serde fields need no migration whenever they are added.

One non-blocking note for the record: the real benefit of this PR is decoupling the storage shape from the #2094 / #2095 API PRs, not migration avoidance. That is a fine reason on its own. The drive-by env-test serialization is a reasonable band-aid for the `setenv` / `getenv` data race; the durable fix is to stop reading process env in test-exercised code, which is out of scope here.

PR #2369, opened 2026-06-24
fix(session): delete workspace dir when removing a multi-repo session

### LINKED ISSUE #2363: Workspace: multi-repo session delete leaks the workspace dir, CLI and web ignore workspace_info

## Problem

Deleting a multi-repo / multi-project (workspace) session does not delete its workspace directory on disk. Per-repo worktrees are left behind, the workspace parent directory stays, and the git worktree registrations dangle in each source repo. Deleting a normal single-repo session removes its worktree as expected, so the leak is specific to workspace sessions.

The shared deletion engine is correct. The bug is that two of the three delete entry points (CLI `aoe remove` and the web dashboard) decide whether to clean up worktrees by looking only at `worktree_info`, which is always `None` for a workspace session. They never set `delete_worktree=true`, so the engine's workspace-cleanup block is skipped entirely. The TUI is the only surface that handles workspace sessions correctly.

## What's there today

The engine already iterates over every repo in the workspace and removes the parent directory, all gated on `request.delete_worktree`:

`src/session/deletion.rs:209-263`
```rust
if request.delete_worktree {
    if let Some(ws_info) = &request.instance.workspace_info {
        if ws_info.cleanup_on_delete {
            for repo in &ws_info.repos {
                if repo.managed_by_aoe {
                    // remove_managed_worktree(...) per repo
                }
            }
            if !any_dirty {
                let ws_path = PathBuf::from(&ws_info.workspace_dir);
                if ws_path.exists() {
                    std::fs::remove_dir_all(&ws_path) // removes the parent dir
                }
            }
        }
    }
}
```

So if `delete_worktree` is `true`, a workspace session is cleaned up correctly. The failure is upstream: the value is `false`.

**CLI.** `needs_worktree_cleanup` only inspects `worktree_info`:

`src/cli/remove.rs:37-41`
```rust
fn needs_worktree_cleanup(inst: &Instance, args: &RemoveArgs) -> bool {
    inst.worktree_info
        .as_ref()
        .is_some_and(|wt| wt.managed_by_aoe && args.delete_worktree)
}
```

A workspace session has `worktree_info = None`, so this returns `false`, `delete_worktree` is `false` (`remove.rs:64`), and the engine's workspace block never runs. `aoe remove --delete-worktree <workspace-session>` leaves the entire workspace dir plus orphaned git worktrees. The "Worktree preserved" hint at `remove.rs:93-102` also only checks `worktree_info`, so the user gets no warning that anything was kept.

**Web.** The session DTO computes `has_managed_worktree` from `worktree_info` only:

`src/server/api/sessions.rs:309-312`
```rust
has_managed_worktree: inst
    .worktree_info
    .as_ref()
    .is_some_and(|w| w.managed_by_aoe),
```

A workspace session reports `has_managed_worktree: false`, so the delete dialog hides the entire worktree section:

`web/src/components/DeleteSessionDialog.tsx:26` and `:120`
```tsx
const [deleteWorktree, setDeleteWorktree] = useState(hasManagedWorktree && cleanupDefaults.delete_worktree);
// ...
{hasManagedWorktree && ( /* "Delete worktree" checkbox */ )}
```

The checkbox is never rendered, `deleteWorktree` stays `false`, and the server forwards it verbatim:

`src/server/api/sessions.rs:2968-2977`
```rust
crate::session::deletion::perform_deletion(&DeletionRequest {
    // ...
    delete_worktree: body.delete_worktree, // false for a workspace session
```

**TUI (the correct surface, for reference).** It falls back to the workspace branch when building the dialog config:

`src/tui/home/input.rs:3966-3972`
```rust
worktree_branch: inst
    .worktree_info
    .as_ref()
    .filter(|wt| wt.managed_by_aoe)
    .map(|wt| wt.branch.clone())
    .or_else(|| inst.workspace_info.as_ref().map(|w| w.branch.clone())),
```

and the group-delete path checks `workspace_info` explicitly (`src/tui/home/operations.rs:634-642`). Both feed `delete_worktree=true`, so the engine cleans up.

## Root cause

Two entry points gate worktree cleanup on `worktree_info` presence, which is structurally `None` for workspace sessions. The session model stores multi-repo state in `workspace_info: Option<WorkspaceInfo>` (`src/session/instance.rs`), a separate field from `worktree_info`. Neither the CLI nor the web DTO accounts for it.

- Layer 1 (CLI): `src/cli/remove.rs:37-41` `needs_worktree_cleanup` ignores `workspace_info`.
- Layer 2 (web): `src/server/api/sessions.rs:309-312` `has_managed_worktree` ignores `workspace_info`, which hides the web checkbox and forces `delete_worktree=false`.

The engine (`src/session/deletion.rs:209-263`) and the TUI (`src/tui/home/input.rs:3972`, `src/tui/home/operations.rs:634-642`) are correct and need no change.

## Proposal

**A. CLI.** Make `needs_worktree_cleanup` also return true for a managed workspace session:

```rust
fn needs_worktree_cleanup(inst: &Instance, args: &RemoveArgs) -> bool {
    args.delete_worktree
        && (inst.worktree_info.as_ref().is_some_and(|wt| wt.managed_by_aoe)
            || inst.workspace_info.as_ref().is_some_and(|ws| ws.cleanup_on_delete))
}
```

Mirror the same `worktree_info || workspace_info` check in the "Worktree preserved" hint (`remove.rs:93-102`) so workspace users get an accurate kept/removed message.

**B. Web.** Set `has_managed_worktree` from either field:

```rust
has_managed_worktree: inst.worktree_info.as_ref().is_some_and(|w| w.managed_by_aoe)
    || inst.workspace_info.as_ref().is_some_and(|ws| ws.cleanup_on_delete),
```

That re-enables the dialog's worktree checkbox for workspace sessions; the existing `cleanup_defaults.delete_worktree` default and the `force_delete` sub-toggle then work unchanged, and the server already forwards the value into the correct engine block.

Both fixes are small and independent. A shared `Instance` helper (e.g. `inst.has_managed_worktree_or_workspace()`) would keep CLI, server, and any future surface from drifting again; worth considering since this exact divergence is the whole bug.

## Why now

- Every workspace-session delete via CLI or web silently leaks disk and leaves dangling git worktree registrations in each source repo; the registrations make the source repos progressively messier (`git worktree list` noise, stale locks).
- The fix is two small edits with no engine changes; low risk, high cleanup value.
- It is a behavior divergence between surfaces (TUI works, CLI/web do not), which is exactly the kind of inconsistency that erodes trust in delete.

- As a user, when I run `aoe remove --delete-worktree` on a multi-repo workspace session, then every per-repo managed worktree and the workspace parent directory are removed from disk and the git worktree registrations are pruned.
  - Suggested test layer: `tests/e2e/` (CLI subprocess) or a Rust integration test that builds a workspace instance and asserts `perform_deletion` removes the dirs.
- As a user, when I open the web delete dialog for a workspace session, then a "Delete worktree" checkbox is shown defaulted per `cleanup_defaults`, and confirming with it checked removes the workspace directory.
  - Suggested test layer: `web/tests/live/` (real `aoe serve` + workspace session), asserting the checkbox renders and deletion clears the dir.
- As a user, when I delete a workspace session with uncommitted changes and without force, then the workspace directory is preserved and an error is surfaced rather than the changes being silently destroyed.
  - Suggested test layer: Rust unit on `perform_deletion` (`any_dirty` path).

## Open questions

- Should the CLI/server use `ws.cleanup_on_delete` (matches the group-delete TUI check at `operations.rs:642`) or `!ws.repos.is_empty() && repos.any(managed_by_aoe)` as the "has cleanable worktree" predicate? `cleanup_on_delete` is the per-session opt-out and reads cleaner, but a session with `cleanup_on_delete=false` would then still report `has_managed_worktree=false` and hide the web checkbox, which is arguably correct (nothing to clean) but worth a deliberate call.
- Is a shared `Instance` helper worth adding now, or is duplicating the two-field check acceptable for two call sites?

## Out of scope

- The engine's per-repo removal and dirty-check logic (`src/session/deletion.rs:103-263`); they are correct.
- The TUI delete flows; they already handle workspace sessions.
- jj workspace support (#879).

## Cross-references

- #1066 — prior deletion work that added the dirty-worktree check and anonymous-volume cleanup; same engine, adjacent area.
- #1511 — multi-repo workspace base-branch handling; same workspace model, different lifecycle stage.

Seluj78:

## Description

Deleting a multi-repo (workspace) session left its workspace directory and every per-repo worktree on disk, plus dangling git worktree registrations in each source repo. Single-repo sessions cleaned up fine, so the leak was specific to workspace sessions deleted from the CLI or the web dashboard.

The deletion engine (`perform_deletion`) already iterates `workspace_info.repos`, removes each worktree, and `remove_dir_all`s the workspace parent, but the whole block is gated on `request.delete_worktree`. Two entry points computed that flag from `worktree_info` alone, which is always `None` for a workspace session (its multi-repo state lives in `workspace_info`): the CLI `needs_worktree_cleanup` and the server's `has_managed_worktree` DTO field. So `delete_worktree` stayed `false`, the engine skipped cleanup, and on the web the "Delete worktree" checkbox was hidden entirely. The TUI was unaffected because it already checked `workspace_info`.

This adds `Instance::has_managed_worktree_or_workspace()`, mirroring the TUI group-delete predicate (an aoe-managed worktree, or a workspace with `cleanup_on_delete`), and routes the CLI predicate, the server DTO field, and the TUI group-delete checks through it so every surface agrees. No engine change. The web checkbox now renders for workspace sessions because the server reports `has_managed_worktree: true`.

Closes #2363

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Create a multi-repo workspace session, then `aoe remove --delete-worktree <session>`; confirm the workspace directory and each per-repo worktree are gone and `git worktree list` in each source repo no longer lists them.
- [ ] Create a multi-repo workspace session, open the web dashboard delete dialog; confirm a "Delete worktree" checkbox now appears, and confirming with it checked removes the workspace directory.
- [ ] Create a workspace session, leave uncommitted changes in one repo, delete without force; confirm the workspace directory is preserved and a warning is surfaced (no silent data loss).
- [ ] Delete a single-repo worktree session via CLI and web; confirm cleanup behavior is unchanged.

## Checklist

## Test Coverage Analysis

The web fix is entirely server-side (the `has_managed_worktree` DTO value); no `web/src` file changed. The dialog already renders the worktree checkbox off that prop, and the existing `DeleteSessionDialog` Vitest covers both prop values.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the fix is a pure cleanup-decision predicate. It is covered by reproduce-then-fix Rust unit tests on the predicate (`Instance`), the CLI `needs_worktree_cleanup`, and the server `from_instance` DTO; each was confirmed red on a tree with the workspace arm neutralized and green with the fix. The engine cleanup path is already covered, so an e2e would add slow tmux setup for no extra signal.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I reviewed each commit, made the predicate design call, and ran the manual test steps.

**NOTE:**

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2369"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784916625&installation_id=139138837&pr_number=2369&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2369&signature=999550fd4f5e1135616a426b919da4155516c854146dda80c51b529e79f493b3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR fixes deletion of multi-repo/workspace sessions so their managed workspace directory and related worktrees are cleaned up correctly across the CLI, web API, and TUI.

What changed:
- Added a shared `Instance` helper (`has_managed_worktree_or_workspace`) to detect whether a session’s managed worktree cleanup should apply (including workspace sessions when cleanup is enabled).
- Updated CLI delete behavior to use the shared predicate, ensuring `--delete-worktree` triggers the existing cleanup logic for workspace sessions too.
- Expanded CLI “preserved” messaging so it can report preserved workspace directories (not just single-repo worktrees) when cleanup isn’t requested.
- Updated the session API response to introduce/drive `has_cleanable_worktree` for workspace sessions, while keeping `has_managed_worktree` reserved for single-repo worktrees (so UI actions don’t accidentally appear for idle workspace sessions).
- Updated TUI delete-group checks (including delete-worktree and delete-branch visibility) to rely on the shared cleanup predicate.
- Updated the web delete dialog to use `has_cleanable_worktree` (via a `hasManagedWorktree` prop) so the “Delete worktree” checkbox reappears for applicable workspace sessions.
- Added/updated regression tests to lock in the intended workspace-vs-single-repo distinction and the cleanup gating behavior.

Benefits:
- Deletion decisions are consistent across CLI, server, and TUI, reducing drift between entry points.
- Workspace sessions correctly expose the “Delete worktree” UI option on the web when cleanup is applicable.
- Workspace cleanup is restored for multi-repo/workspace sessions when the user requests it.

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -35,9 +35,7 @@ pub struct RemoveArgs {
 }
 
 fn needs_worktree_cleanup(inst: &Instance, args: &RemoveArgs) -> bool {
-    inst.worktree_info
-        .as_ref()
-        .is_some_and(|wt| wt.managed_by_aoe && args.delete_worktree)
+    args.delete_worktree && inst.has_managed_worktree_or_workspace()
 }
 
 #[tracing::instrument(target = "cli.session", skip_all, fields(profile = %profile))]
@@ -91,11 +89,20 @@ pub async fn run(profile: &str, args: RemoveArgs) -> Result<()> {
     }
 
     if !delete_worktree {
-        if let Some(wt_info) = &inst.worktree_info {
-            if wt_info.managed_by_aoe {
+        if inst
+            .worktree_info
+            .as_ref()
+            .is_some_and(|wt| wt.managed_by_aoe)
+        {
+            println!(
+                "Worktree preserved at: {} (use --delete-worktree to remove)",
+                inst.project_path
+            );
+        } else if let Some(ws_info) = &inst.workspace_info {
+            if ws_info.cleanup_on_delete {
                 println!(
-                    "Worktree preserved at: {} (use --delete-worktree to remove)",
-                    inst.project_path
+                    "Workspace preserved at: {} (use --delete-worktree to remove)",
+                    ws_info.workspace_dir
                 );
             }
         }
@@ -136,4 +143,47 @@ pub async fn run(profile: &str, args: RemoveArgs) -> Result<()> {
     );
 
     Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::session::{WorkspaceInfo, WorkspaceRepo};
+    use chrono::Utc;
+
+    fn args(delete_worktree: bool) -> RemoveArgs {
+        RemoveArgs {
+            identifier: "x".to_string(),
+            delete_worktree,
+            delete_branch: false,
+            force: false,
+            keep_container: false,
+            keep_scratch: false,
+        }
+    }
+
+    // Regression for #2363: a multi-repo workspace session has no
+    // `worktree_info`, so the old worktree_info-only check returned false and
+    // `--delete-worktree` silently left the workspace dir on disk.
+    #[test]
+    fn needs_worktree_cleanup_true_for_workspace_session() {
+        let mut inst = Instance::new("WS", "/tmp/ws/repo-a");
+        inst.workspace_info = Some(WorkspaceInfo {
+            branch: "feature/abc".to_string(),
+            workspace_dir: "/tmp/ws".to_string(),
+            repos: vec![WorkspaceRepo {
+                name: "repo-a".to_string(),
+                source_path: "/tmp/src/repo-a".to_string(),
+                branch: "feature/abc".to_string(),
+                worktree_path: "/tmp/ws/repo-a".to_string(),
+                main_repo_path: "/tmp/src/repo-a".to_string(),
+                managed_by_aoe: true,
+            }],
+            created_at: Utc::now(),
+            cleanup_on_delete: true,
+        });
+
+        assert!(needs_worktree_cleanup(&inst, &args(true)));
+        assert!(!needs_worktree_cleanup(&inst, &args(false)));
+    }
 }
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -96,7 +96,18 @@ pub struct SessionResponse {
     /// `theme.unread`.
     #[serde(skip_serializing_if = "std::ops::Not::not")]
     pub unread: bool,
+    /// Strictly a single-repo aoe-managed worktree (`worktree_info`). Drives
+    /// the sidebar "Edit workdir name" action and the tie-workdir overlay,
+    /// neither of which applies to multi-repo workspace sessions. For
+    /// "is there worktree state to clean up on delete", use
+    /// `has_cleanable_worktree` instead.
     pub has_managed_worktree: bool,
+    /// Whether deleting this session has aoe-managed worktree state to remove,
+    /// covering single-repo worktrees AND multi-repo workspaces. Only the
+    /// delete dialog's worktree/branch checkboxes consume this; keeping it
+    /// separate from `has_managed_worktree` avoids lighting up worktree-only
+    /// actions (Edit workdir) for workspace sessions (#2363).
+    pub has_cleanable_worktree: bool,
     /// Whether renaming this session also moves its worktree directory (the
     /// resolved `session.tie_workdir_to_name` for an aoe-managed worktree).
     /// Populated by `list_sessions` from the per-profile config; single-session
@@ -306,7 +317,11 @@ impl SessionResponse {
             // Surface the marker (omitted when read); the web gates the
             // visual on the `session.unread_indicator` setting.
             unread: inst.unread,
-            has_managed_worktree: inst.has_managed_worktree_or_workspace(),
+            has_managed_worktree: inst
+                .worktree_info
+                .as_ref()
+                .is_some_and(|w| w.managed_by_aoe),
+            has_cleanable_worktree: inst.has_managed_worktree_or_workspace(),
             // Overlaid per-profile in list_sessions; see the field doc.
             tie_workdir_to_name: false,
             // Overlaid in list_sessions; single-session responses stay inactive.
@@ -5234,9 +5249,11 @@ mod tests {
     }
 
     // Regression for #2363: a multi-repo workspace session carries
-    // `workspace_info` and no `worktree_info`. The DTO must still report
-    // `has_managed_worktree: true`, otherwise the web delete dialog hides the
-    // "Delete worktree" checkbox and the workspace directory leaks on disk.
+    // `workspace_info` and no `worktree_info`. The DTO must report
+    // `has_cleanable_worktree: true` so the web delete dialog shows the
+    // "Delete worktree" checkbox, while keeping `has_managed_worktree: false`
+    // so worktree-only actions (sidebar "Edit workdir name", tie overlay) stay
+    // hidden for workspace sessions.
     #[test]
     fn from_instance_reports_managed_worktree_for_workspace_session() {
         let mut inst = make_test_instance();
@@ -5257,8 +5274,12 @@ mod tests {
 
         let resp = SessionResponse::from_instance(&inst, false);
         assert!(
-            resp.has_managed_worktree,
-            "workspace session must report a managed worktree"
+            resp.has_cleanable_worktree,
+            "workspace session must report a cleanable worktree so the delete checkbox shows"
+        );
+        assert!(
+            !resp.has_managed_worktree,
+            "workspace session must NOT report a single-repo managed worktree (keeps Edit-workdir hidden)"
         );
     }
 
@@ -7192,6 +7213,7 @@ mod workspace_ordering_tests {
             is_sandboxed: false,
             scratch: false,
             has_managed_worktree: false,
+            has_cleanable_worktree: false,
             tie_workdir_to_name: false,
             smart_rename: crate::session::smart_rename::SmartRenameState::Inactive,
             default_name: false,
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -306,10 +306,7 @@ impl SessionResponse {
             // Surface the marker (omitted when read); the web gates the
             // visual on the `session.unread_indicator` setting.
             unread: inst.unread,
-            has_managed_worktree: inst
-                .worktree_info
-                .as_ref()
-                .is_some_and(|w| w.managed_by_aoe),
+            has_managed_worktree: inst.has_managed_worktree_or_workspace(),
             // Overlaid per-profile in list_sessions; see the field doc.
             tie_workdir_to_name: false,
             // Overlaid in list_sessions; single-session responses stay inactive.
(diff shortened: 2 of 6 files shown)


njbrake (changes requested):

Thanks for this. The core fix is correct, minimal, and well-tested, and the leak (#2363) clearly belongs fixed. I verified the load-bearing facts in source rather than from the description:

- `deletion.rs:209-254`: the workspace cleanup block is gated on `request.delete_worktree`, iterates `workspace_info.repos`, removes managed worktrees, and `remove_dir_all`s the workspace path.
- `deletion.rs:103-121`: engine-side dirty-repo protection for workspace repos already exists, so flipping `delete_worktree=true` correctly engages the "preserve on uncommitted changes" path.
- The new predicate's gate (`workspace_info.cleanup_on_delete`) matches the engine's own gate, so every surface agrees. The TUI refactor is behavior-preserving.

## The one blocker: `has_managed_worktree` is overloaded, and the DTO flip has two web consumers beyond the delete dialog

`has_managed_worktree` previously meant "single-repo aoe-managed worktree." Three places read it, and the PR only accounts for one (the delete-worktree checkbox):

1. **`WorkspaceSidebar.tsx:988` `canEditWorkdir`** = `has_managed_worktree && !tie_workdir_to_name && !running`. With the flip, an idle workspace session now shows the "Edit workdir name" action that was previously hidden. Submitting it calls `set_worktree_name`, which returns 400 "Session does not use a worktree" (`sessions.rs:1267`). That is a new dead-end action for workspace sessions.
2. **`DeleteSessionDialog.tsx:140`**: the "Delete branch" checkbox now renders for workspace sessions, but the engine's `branch_to_delete` (`deletion.rs:164-173`) only handles `worktree_info`, so checking it is a silent no-op for workspace repos.
3. **Tie-workdir overlay** (`sessions.rs:576`, `3912`) now runs for workspace sessions. Harmless in effect (the rename move is gated on `worktree_info`), but it is the same overload feeding the sidebar above.

This is not data loss; the backend rejects safely. But it is a real UX regression, and it contradicts the "entirely server-side, no user-facing dashboard flow changed" note in the PR, since it changes the sidebar action set and adds a no-op delete-branch checkbox for workspace sessions.

The DTO already exposes `workspace_repos` (empty for single-repo), which `WorkspaceSidebar.tsx:1217` already uses, so the web can distinguish workspace sessions. Two ways to fix, cleanest first:

- **Preferred:** keep `has_managed_worktree` strict (single-repo `worktree_info`) and add a dedicated field (e.g. `has_cleanable_worktree`) derived from `has_managed_worktree_or_workspace()` that only the delete-worktree checkbox consumes. That keeps the field name honest and leaves the sidebar and tie paths untouched.
- **Minimal:** keep the overload but guard `canEditWorkdir` with `(firstSession?.workspace_repos?.length ?? 0) === 0` and suppress the delete-branch checkbox for workspace sessions.

Either way, a small test asserting that a workspace session shows the delete-worktree checkbox but not the Edit-workdir action would have caught this and would lock it down.

## Nit

CodeRabbit flagged the `fix(session):` title as non-conforming. I disagree; the repo consistently uses scopes (`feat(plugins)`, `feat(web)`, `docs(web)`, `fix(tui)` in recent history), so the title is fine as is.

Happy to take the broadened-DTO-field route myself if you'd like, just say the word.

Seluj78:
**Note:** drafted by Claude via back-and-forth with @Seluj78; decisions are mine.

Thanks, good catch on the overload. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2369/commits/290438c2: `has_managed_worktree` is strict single-repo `worktree_info` again, and a new `has_cleanable_worktree` (derived from `has_managed_worktree_or_workspace`) feeds only the delete dialogs worktree/branch checkboxes. So the sidebar Edit-workdir action and the tie-workdir overlay stay hidden for workspace sessions. Also folded `group_has_managed_worktrees` into the shared helper.

One correction on the delete-branch point: the engine does handle workspace branch deletion (`deletion.rs:294-317`, gated on `delete_branch`, loops `workspace_info.repos`), so the branch checkbox is functional for workspace sessions, not a silent no-op. Left it showing intentionally. Agreed the tie overlay is harmless in effect. The serve unit test now locks `has_cleanable_worktree == true` plus `has_managed_worktree == false` for a workspace session.

njbrake (approved):

PR #2382, opened 2026-06-25
fix(kiro): launch via `chat` subcommand and support user-selected agents

### LINKED ISSUE #2380: fix(kiro): YOLO mode produces "unexpected argument '--trust-all-tools'" (flag belongs on the chat subcommand)
### What happens

Enabling YOLO mode for a Kiro session fails to launch the agent. The command AoE builds is:

```
kiro-cli --trust-all-tools
```

which `kiro-cli` rejects:

```
error: unexpected argument '--trust-all-tools' found

Usage: kiro-cli [OPTIONS] [COMMAND]
```

### Why

`--trust-all-tools` (and `--agent`, `--resume-id`, `--model`, etc.) are defined on the `kiro-cli chat` subcommand, not the top-level binary. Verified on `kiro-cli` 2.9.0:

```
$ kiro-cli chat --help
AI assistant in your terminal
Usage: kiro-cli-chat chat [OPTIONS] [INPUT]
Options:
  ...
  --agent <AGENT>          Context profile to use
  -a, --trust-all-tools    Allows the model to use any tool ... without asking
  --resume-id <SESSION_ID> Resume a specific conversation by session ID
```

Bare `kiro-cli` (no subcommand) drops into interactive chat, but it does **not** accept the chat subcommand's flags, so appending the yolo flag to the bare binary fails.

In the registry, kiro declares `yolo: Some(YoloMode::CliFlag("--trust-all-tools"))` and the command builder appends that flag after `a.binary` (`kiro-cli`), with no `chat` token in between.

### Repro

1. Add a Kiro session, enable YOLO mode in the new-session dialog (or `aoe add --tool kiro --yolo --launch`).
2. The pane shows the `unexpected argument` error instead of a chat.

### Expected

YOLO mode launches `kiro-cli chat --trust-all-tools ...` and works, matching Claude's `--dangerously-skip-permissions`.

### Workaround

Use a custom command override of `kiro-cli chat --trust-all-tools` instead of the YOLO toggle.

### Proposed fix

Add a `launch_subcommand: Option<&'static str>` to `AgentDef` (`Some("chat")` for kiro) and inject it after the binary in the host and sandbox command builders, before extra args / yolo / resume flags. PR incoming.

---

*Filing note: I'm a developer evaluating AoE for managing several Kiro sessions; I hit this bug in real use. I used an AI agent (Claude Code) to help investigate the root cause and draft this report; the analysis was verified against `kiro-cli` 2.9.0 and the AoE source on my machine. Happy to discuss directly.*

### LINKED ISSUE #2381: fix(kiro): user-selected --agent loses status detection (hooks only live in the aoe-hooks agent)
### What happens

When launching a Kiro session with a custom agent (e.g. `--agent my-team-agent`), status detection stops working: AoE shows the agent as idle/unknown regardless of what it is actually doing.

Separately, AoE silently changes the user's **global default** Kiro agent to `aoe-hooks` (via `kiro-cli agent set-default aoe-hooks`) on first launch when the user is still on the built-in default, which is surprising.

### Why

AoE installs its status-detection hooks into a dedicated Kiro agent config, `~/.kiro/agents/aoe-hooks.json`, then makes it the default agent. But **Kiro has no global/user-level hooks mechanism** — hooks are scoped to whichever agent `--agent` selects (`Source::Agent`), with only an ephemeral in-session `/hooks` (`Source::Session`) as the other scope. So when a user runs `--agent my-team-agent`, Kiro loads *that* agent's config, which contains none of AoE's hooks, and status detection goes dark.

This makes AoE's status detection and a user's own Kiro agents mutually exclusive.

### Expected

I can run my own Kiro agent and still get AoE status detection, without AoE hijacking my global default agent.

### Proposed fix

When a Kiro session selects a custom agent via `--agent NAME`, merge AoE's status hooks into *that* agent's config file (`~/.kiro/agents/NAME.json`) instead of the standalone `aoe-hooks` agent, and skip the global-default change. The existing installer already preserves user hooks and is idempotent, and removal stays clean (only AoE-tagged entries are stripped). I modeled this as a data-driven `selected_agent_hooks` field on `SidecarHooks` (so it's not a kiro-specific string check), gated behind a setting (default on) so users who don't want their agent files touched can opt out. PR incoming.

### Open design question

The PR merges hooks automatically (gated by a default-on setting) rather than prompting per-launch, to avoid interrupting the new-session flow. If you'd prefer an interactive opt-in prompt in the new-session dialog (similar to the existing hooks-install dialog), I'm happy to switch to that. Also flagging: the default-on setting means existing Kiro users who launch with `--agent` would start having AoE hooks written into their agent files; if you'd rather that be opt-in (default off), that's a one-line change. Wanted your call before finalizing.

---

*Filing note: I'm a developer using AoE to manage several Kiro sessions and hit this in real use. I used an AI agent (Claude Code) to investigate and draft this, verified against the AoE source and `kiro-cli` 2.9.0 behavior. Happy to discuss directly.*

MatthewWolff:
## Description

Two related Kiro fixes. Closes #2380 and Closes #2381.

**1. YOLO mode was broken (#2380).** Kiro's interactive flags (`--trust-all-tools`, `--agent`, `--resume-id`) live on the `kiro-cli chat` subcommand, not the top-level binary. AoE built the launch command as bare `kiro-cli` plus the yolo flag, producing `kiro-cli --trust-all-tools`, which the CLI rejects with `error: unexpected argument '--trust-all-tools' found`. Added a `launch_subcommand: Option<&'static str>` field to `AgentDef` (`Some("chat")` for kiro, `None` elsewhere) and inject it right after the binary in the default host and sandbox command builders, before extra args, yolo, and resume flags. User command overrides are passed through verbatim.

**2. Custom agents could not be used (#2381).** AoE installs its status hooks into a dedicated `aoe-hooks` Kiro agent and sets it as the global default. Kiro has no global hooks, so launching with a user's own `--agent NAME` loads that agent's config (no AoE hooks) and status detection goes dark. Now, when a Kiro session selects a custom agent, AoE merges its status hooks into *that* agent's own config file (reusing the idempotent, user-hook-preserving installer) and skips the global-default promotion. Modeled as a data-driven `selected_agent_hooks` field on `SidecarHooks` (flag + config-path fn) rather than a per-agent string check, matching the collapse-special-cases direction from #2062/#2333. Gated by a new `merge_hooks_into_selected_agent` setting (default on). Selected agent names are validated against path traversal and flag-shaped values; `--agent` parsing is last-wins to match clap precedence.

**Open question for the maintainer (also noted in #2381):** the merge is automatic, gated by a default-on setting, rather than an interactive prompt. If you'd prefer a new-session dialog opt-in, or the setting defaulting off so existing users' agent files are never touched without opt-in, both are small changes — happy to adjust to your preference.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added `tests/e2e/kiro_launch.rs`: drives the full `aoe add --launch` path for a kiro session and asserts on the command tmux was actually told to run (`pane_start_command`) — that it launches via `kiro-cli chat`, and that `--trust-all-tools` lands after `chat` in YOLO mode. (Reads the recorded pane command rather than executing a stub `kiro-cli`, since aoe wraps the launch in a login shell that re-resolves the binary from the real PATH.) Also added unit tests for command construction (chat subcommand, yolo ordering, override passthrough, drift guards), `--agent` parsing (last-wins, path-traversal + flag-shaped-value rejection, parameterized flag), and merging hooks into a user agent while preserving their config and removing cleanly.

Verification: `cargo fmt --check` clean, `cargo clippy --lib --tests -- -D warnings` clean, full lib suite passes (`cargo test --lib -- --test-threads=1`: 3355 passed), e2e kiro_launch passes.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Code (Claude Opus).

**Any Additional AI Details you'd like to share:** I'm a developer evaluating AoE for managing several Kiro sessions and hit both bugs in real use. I used Claude Code to investigate the root causes, implement the fix, and write the tests; I directed the work, reviewed every change, and verified the build/clippy/tests on my machine against `kiro-cli` 2.9.0. I'll respond to review comments myself.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2382"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784961594&installation_id=139138837&pr_number=2382&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2382&signature=7224183f790a14cd473ef1926e9f25e542e356f13e71d46e20cf5893b7fa0c58"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR fixes two Kiro-related issues and improves reliability for users using custom agents:

- It corrects how Kiro interactive options are wired by passing Kiro flags through `kiro-cli chat` (via a new `launch_subcommand` mechanism). This ensures `--trust-all-tools`, `--agent`, and resume behavior apply in the right place and that YOLO/extra args order is preserved. User command overrides still work as before.
- It restores custom Kiro agent compatibility with AoE status hooks by merging AoE’s hooks into the *selected* agent’s config when enabled. This includes fixing sandbox installation to target the selected agent’s staged config (so hooks don’t get stuck only on the default agent).

The PR adds safer `--agent` parsing with true last-wins semantics and validation, introduces configuration-controlled hook merging (`merge_hooks_into_selected_agent`, defaulting to enabled), and adds validation around selected agent names. It also includes an end-to-end launch test for Kiro command construction (with a `kiro-cli` stub for CI robustness) plus unit tests covering command building, parsing behavior, hook merging, and sandbox behavior.

Potential follow-up: consider whether the default-on merge behavior (`merge_hooks_into_selected_agent`) should be changed to default-off or made interactive.

### DIFF
--- a/src/agents.rs
+++ b/src/agents.rs
@@ -124,9 +124,32 @@ pub struct SidecarHooks {
     /// Optional host-only follow-up run after a successful host install
     /// (e.g. kiro promotes its `aoe-hooks` agent to the active default).
     pub post_install_host: Option<fn()>,
+    /// Set for CLIs whose hooks are scoped to a user-selectable named agent
+    /// rather than applying globally (e.g. Kiro: `--agent NAME` loads only that
+    /// agent's config, and there is no global hooks mechanism). When set and
+    /// the user selected an agent, AoE installs its hooks into that agent's own
+    /// config file instead of the standalone `host_config_subpath` agent, and
+    /// skips `post_install_host`. `None` for agents whose hooks apply
+    /// regardless of which agent is selected. See
+    /// [`crate::session::Instance::install_agent_status_hooks`].
+    pub selected_agent_hooks: Option<SelectedAgentHooks>,
     /// On-disk format of the sidecar's config file. Drives marker-presence
     /// walker dispatch in [`crate::hooks::has_aoe_marker`].
     pub format: SidecarFormat,
+}
+
+/// How to install status hooks into a user-selected named agent, for CLIs
+/// whose hooks are scoped to the selected agent (see
+/// [`SidecarHooks::selected_agent_hooks`]). Keeps the flag and path convention
+/// as data on the agent definition rather than a per-agent string match at the
+/// install site.
+#[derive(Debug)]
+pub struct SelectedAgentHooks {
+    /// CLI flag a user passes to choose a named agent (e.g. `"--agent"`).
+    pub flag: &'static str,
+    /// Maps a selected agent name to the home-relative path of that agent's
+    /// config file (e.g. `custom-agent` → `.kiro/agents/custom-agent.json`).
+    pub config_subpath: fn(&str) -> std::path::PathBuf,
 }
 
 /// On-disk format of a sidecar agent's config file. Drives
@@ -149,6 +172,20 @@ pub struct AgentDef {
     pub name: &'static str,
     /// Binary to invoke (usually same as name).
     pub binary: &'static str,
+    /// Subcommand token inserted immediately after `binary` when AoE builds the
+    /// default launch command (e.g. `Some("chat")` for kiro → `kiro-cli chat`).
+    /// Required for CLIs whose interactive flags (yolo, `--agent`, resume) live
+    /// on a subcommand rather than the top-level binary: bare
+    /// `kiro-cli --trust-all-tools` is rejected with "unexpected argument",
+    /// while `kiro-cli chat --trust-all-tools` parses. `None` for agents whose
+    /// bare binary already accepts those flags. Only applied to the default
+    /// binary path, never to a user's custom command override.
+    ///
+    /// Must not be combined with [`ResumeStrategy::Subcommand`]: that strategy
+    /// inserts the resume token after the first whitespace token (the binary),
+    /// which would land it before this launch subcommand. The pairing is
+    /// rejected by `test_launch_subcommand_not_combined_with_subcommand_resume`.
+    pub launch_subcommand: Option<&'static str>,
     /// Alternative substrings recognised by `resolve_tool_name` (e.g. `"open-code"`).
     pub aliases: &'static [&'static str],
     /// How to detect availability on the host.
@@ -359,6 +396,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "claude",
         oneshot_flag: Some("-p"),
         binary: "claude",
+        launch_subcommand: None,
         aliases: &[],
         detection: DetectionMethod::Which("claude"),
         yolo: Some(YoloMode::CliFlag("--dangerously-skip-permissions")),
@@ -385,6 +423,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "opencode",
         oneshot_flag: Some("run"),
         binary: "opencode",
+        launch_subcommand: None,
         aliases: &["open-code"],
         detection: DetectionMethod::Which("opencode"),
         yolo: Some(YoloMode::EnvVar("OPENCODE_PERMISSION", r#"{"*":"allow"}"#)),
@@ -403,6 +442,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "vibe",
         oneshot_flag: None,
         binary: "vibe",
+        launch_subcommand: None,
         aliases: &["mistral-vibe"],
         detection: DetectionMethod::RunWithArg("vibe", "--version"),
         yolo: Some(YoloMode::CliFlag("--agent auto-approve")),
@@ -421,6 +461,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "codex",
         oneshot_flag: Some("exec"),
         binary: "codex",
+        launch_subcommand: None,
         aliases: &[],
         detection: DetectionMethod::Which("codex"),
         yolo: Some(YoloMode::CliFlag(
@@ -451,6 +492,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "gemini",
         oneshot_flag: Some("-p"),
         binary: "gemini",
+        launch_subcommand: None,
         aliases: &[],
         detection: DetectionMethod::Which("gemini"),
         yolo: Some(YoloMode::CliFlag("--approval-mode yolo")),
@@ -499,6 +541,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "cursor",
         oneshot_flag: None,
         binary: "agent",
+        launch_subcommand: None,
         aliases: &["agent"],
         detection: DetectionMethod::Which("agent"),
         yolo: Some(YoloMode::CliFlag("--yolo")),
@@ -522,6 +565,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "copilot",
         oneshot_flag: None,
         binary: "copilot",
+        launch_subcommand: None,
         aliases: &["github-copilot"],
         detection: DetectionMethod::Which("copilot"),
         yolo: Some(YoloMode::CliFlag("--yolo")),
@@ -540,6 +584,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "pi",
         oneshot_flag: None,
         binary: "pi",
+        launch_subcommand: None,
         aliases: &[],
         detection: DetectionMethod::Which("pi"),
         // Pi runs in full YOLO mode by default (no approval gates), so no flag needed.
@@ -559,6 +604,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "droid",
         oneshot_flag: None,
         binary: "droid",
+        launch_subcommand: None,
         aliases: &["factory-droid"],
         detection: DetectionMethod::Which("droid"),
         yolo: Some(YoloMode::CliFlag("--skip-permissions-unsafe")),
@@ -577,6 +623,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "settl",
         oneshot_flag: None,
         binary: "settl",
+        launch_subcommand: None,
         aliases: &["settlers", "catan"],
         detection: DetectionMethod::Which("settl"),
         yolo: Some(YoloMode::AlwaysYolo),
@@ -594,6 +641,7 @@ pub const AGENTS: &[AgentDef] = &[
             install: crate::hooks::install_settl_hooks,
             uninstall: crate::hooks::uninstall_settl_hooks,
             post_install_host: None,
+            selected_agent_hooks: None,
             format: SidecarFormat::SettlToml,
         }),
         resume_strategy: ResumeStrategy::Unsupported,
@@ -605,6 +653,7 @@ pub const AGENTS: &[AgentDef] = &[
         name: "hermes",
         oneshot_flag: None,
         binary: "hermes",
+        launch_subcommand: None,
         aliases: &[],
         detection: DetectionMethod::Which("hermes"),
         yolo: Some(YoloMode::CliFlag("--yolo")),
@@ -628,6 +677,7 @@ pub const AGENTS: &[AgentDef] = &[
             install: crate::hooks::install_hermes_hooks,
             uninstall: crate::hooks::uninstall_hermes_hooks,
             post_install_host: None,
+            selected_agent_hooks: None,
             format: SidecarFormat::HermesYaml,
         }),
         resume_strategy: ResumeStrategy::Flag("--resume"),
@@ -640,6 +690,10 @@ pub const AGENTS: &[AgentDef] = &[
         name: "kiro",
         oneshot_flag: None,
         binary: "kiro-cli",
+        // Kiro's interactive flags (--trust-all-tools, --agent, --resume-id)
+        // are defined on the `chat` subcommand. Bare `kiro-cli --trust-all-tools`
+        // fails with "unexpected argument"; `kiro-cli chat ...` parses.
+        launch_subcommand: Some("chat"),
         aliases: &["kiro-cli"],
         detection: DetectionMethod::Which("kiro-cli"),
         yolo: Some(YoloMode::CliFlag("--trust-all-tools")),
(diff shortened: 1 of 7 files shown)


Seluj78 (changes requested):

Reviewed PR head d4f7841 against base main a082f1b. 6 files (agents, hooks, settings/session, rust e2e). No unresolved review threads; CodeRabbit posted a summary only; codecov reports all modified lines covered (84.42%). No CI failures (pending/pass).

## Request changes

### Blocker: `Closes #2381` is not satisfied for sandboxed sessions

`src/session/container_config.rs:1471` installs sidecar hooks into `sidecar.sandbox_config_subpath` (kiro: `.kiro/sandbox/agents/aoe-hooks.json`) unconditionally. It never calls `parse_selected_agent` or consults the new `selected_agent_hooks`. A sandboxed kiro session launched with `--agent my-team-agent` therefore loads that agent's config, which has no AoE hooks, and status detection still goes dark. That is exactly the #2381 failure mode, left unfixed on the sandbox path.

The host path (`Instance::install_sidecar_host_hooks`, `src/session/instance.rs:2174`) handles the selected agent correctly; the container path does not. The `merge_hooks_into_selected_agent` setting doc at `src/session/config.rs:865` says hooks land "into that agent's own config file so status detection keeps working" with no sandbox caveat, which overstates coverage.

Fix: thread the selected-agent resolution into the sandbox sidecar install at `container_config.rs:1471`, mirroring the host branch, and add a sandbox case to the tests. If sandbox support is intentionally out of scope, narrow the setting doc and the PR description to say host-only and keep #2381 open for the sandbox case rather than closing it.

### Blocker: em dashes in source (AGENTS.md ban)

`AGENTS.md` Coding Style bans em dashes and CI does not catch them. Two added lines carry one:

- `src/agents.rs:1144`: `// binary - landing the resume token before the subcommand and producing`
- `src/session/instance.rs:5886`: `// Exactly one "chat" token - no doubled chat chat.`

Both are one-character fixes.

## What is good

- Settings parity is satisfied automatically. `SessionConfig` derives `SettingsSection`, so the new `#[setting]`-tagged `merge_hooks_into_selected_agent` field flows into the TUI, web, server and validation through `settings_schema::schema()`. No manual fields.rs / input.rs / profile_config.rs wiring is needed in this schema-driven design, and the default is set in the `Default` impl.
- `parse_selected_agent` guards path traversal (rejects empty, `.`, `..`, leading dash, `/`, `\`) before the name is joined to `$HOME`. Last-wins matches clap precedence.
- Tracing calls carry explicit targets; no secrets logged; no format!/serde inside macro args.
- Tests cover the chat subcommand, yolo ordering, override passthrough, `--agent` parsing edge cases, drift guards, and host-side hook merge with user-config preservation. e2e uses `#[serial]` and a teardown guard, registered in main.rs.
- Drift-guard tests (`test_only_kiro_uses_launch_subcommand`, `test_launch_subcommand_not_combined_with_subcommand_resume`) lock the surface, so a future agent must opt in deliberately.

## Project fit

Accept. Both #2380 and #2381 are real, reproduced against kiro-cli 2.9.0 by an evaluator hitting them in real use. The data-driven `launch_subcommand` / `selected_agent_hooks` design matches the collapse-special-cases direction the PR cites (#2062/#2333). The default-on setting that writes into users' agent files is the one judgment call, and the author already surfaced it to the maintainer as an open question, which is the right move.

Verdict: Request Changes. Fix the sandbox install path (or rescope #2381) and the two em dashes; the rest is solid.

Seluj78 (dismissed):

Update after CI finished and CodeRabbit reviewed. The earlier two blockers (sandbox install path, em dashes) still stand. Two new items below; the CI failure is the headline.

## New blocker: the PR's own e2e tests fail in CI on both platforms

`Cargo Test (linux)` and `Cargo Test (macos)` both failed. The two new tests added by this PR panic:

```
kiro_launch::test_kiro_launches_via_chat_subcommand ... FAILED
kiro_launch::test_kiro_yolo_passes_trust_all_tools_after_chat ... FAILED

thread panicked at tests/e2e/kiro_launch.rs:49:28:
no sessions.json at /tmp/.../.config/agent-of-empires-dev/profiles/default/sessions.json after launch
```

`launch_kiro_and_read_command` runs `aoe add --launch`, then `launched_tmux_name` reads `app_dir_in(home)/profiles/default/sessions.json` and panics because the file is not there. In CI's non-TTY environment with no `kiro-cli` on PATH, the `--launch` path does not persist the session at the path the helper expects, so the test never reaches its real assertion.

The PR body states "e2e kiro_launch passes" locally; it does not pass in CI. This is the author's own new test, red on both linux and macos, so it blocks merge until the helper resolves the persisted sessions path correctly (or the test stops depending on `--launch` persisting before attach). Fix the test harness assumption and confirm green in CI, not just locally.

## Required: parse_selected_agent diverges from real last-wins (CodeRabbit, src/agents.rs:849)

CodeRabbit flagged this as Major and it is correct. For `--agent good --agent ..`, `parse_selected_agent` keeps `good` because an invalid later value is ignored. But Kiro resolves the flag last-wins, so the CLI actually loads `..` (rejecting it or falling back), never `good`. AoE then writes its hooks into `good`, an agent the CLI is not running, so status detection goes dark and a stray agent file is touched.

The doc comment and `test_parse_selected_agent` assert the keep-earlier behavior deliberately, but that contradicts the documented last-wins intent. Either clear the selection when the last occurrence is invalid (so AoE falls back to the standalone aoe-hooks agent, matching what the CLI does), or document that AoE intentionally diverges from the CLI here and why. CodeRabbit thread: https://github.com/agent-of-empires/agent-of-empires/pull/2382#discussion_r3472407437

## Still open from the first review

- Blocker: sandbox install path (`src/session/container_config.rs:1471`) ignores `selected_agent_hooks`, so a sandboxed kiro session with `--agent NAME` still loses status detection. `Closes #2381` is not satisfied for sandbox.
- Blocker: em dashes at `src/agents.rs:1144` and `src/session/instance.rs:5886` (AGENTS.md ban).

Verdict unchanged: Request Changes. The design and host-path implementation are sound; the merge gate is now the red CI, the sandbox gap, and the two smaller correctness items.

MatthewWolff:
Thanks for the thorough review. Pushed `4e84684` addressing all four points; CI is green on both platforms now.

**Sandbox install path (blocker, #2381 for sandbox).** Fixed rather than rescoped. `ContainerAgentSelection` now carries the selected agent, and `Instance::build_container_config` resolves it from the launch args + the sidecar's `selected_agent_hooks` flag (gated by the same `merge_hooks_into_selected_agent` setting as the host path). The sandbox sidecar install then targets the selected agent's staged config (`.kiro/sandbox/agents/NAME.json`) instead of `aoe-hooks.json`. The selected agent's dir is already synced into the sandbox via the kiro `AGENT_CONFIG_MOUNTS` entry (`copy_dirs` includes `agents`). Added `test_build_container_config_installs_hooks_into_selected_kiro_agent` (asserts hooks land in the selected agent's file and not the standalone one). I don't have Docker locally, so this is verified at the config-build/unit level rather than a full container run. The setting doc now says host and sandbox.

**`parse_selected_agent` last-wins (CodeRabbit).** You're right. It's now true last-wins: each occurrence overwrites the selection with its validated value, so `--agent good --agent ..` returns `None` (and AoE falls back to the standalone hooks agent) rather than keeping `good`, since the CLI would load `..` and not `good`. Updated the doc and tests (`--agent good --agent ..` and trailing-dangling-flag both assert `None`).

**e2e tests red in CI (blocker).** Root cause: `aoe add --tool kiro` runs the PATH availability check before persisting, so with no `kiro-cli` on PATH (CI) it bailed and never wrote `sessions.json`. The tests now install an exit-0 `kiro-cli` stub before `aoe add`; the assertion still reads `pane_start_command` (the command tmux is told to run), which is built regardless of the binary. Reproduced the original failure locally by removing `kiro-cli` from PATH, then confirmed green under the same condition.

**Em dashes.** Removed from `agents.rs` and `instance.rs`.

Still happy to adjust the default-on `merge_hooks_into_selected_agent` to default-off, or move to an interactive new-session prompt, if you'd prefer not to write into users' agent files without explicit opt-in.

Seluj78:
You have another review from coderabbit ;)

njbrake (approved):

## Approve

Both bugs are real and well-reproduced against `kiro-cli` 2.9.0 (#2380 even quotes the actual `kiro-cli chat --help`, which confirms the flags live on the subcommand). The data-driven `launch_subcommand` / `selected_agent_hooks` design matches the collapse-special-cases direction from #2062/#2333, and all four blockers from the earlier rounds are genuinely resolved: sandbox path now honors the selected agent, `parse_selected_agent` is true last-wins, the e2e tests install a `kiro-cli` stub so `aoe add` persists in CI, and the em dashes are gone.

I verified locally: built the lib and ran the kiro-related suites (agents, instance, container_config, hooks), all green; `cargo fmt --check` clean; no em dashes in the diff. Resume placement is correct too: kiro's `Flag("--resume-id")` strategy appends after `chat`, and the drift guard forbids pairing `launch_subcommand` with `ResumeStrategy::Subcommand`.

One follow-up, already pushed to your branch as `31eeaa53`: the sandbox path computed the selected agent from `self.tool` only, while the host path and the sandbox's own `resolve_active_agent` fall back to `detect_as`. A custom wrapper detected as kiro, launched sandboxed with `--agent NAME`, would have installed standalone hooks while the container loaded `NAME` without them, so status would go dark on the very case this PR sets out to fix. The change mirrors the host resolution (`tool` then `detect_as`). This also covers the remaining CodeRabbit comment on `instance.rs`. Feel free to squash or drop it if you'd rather handle it differently.

On your open question: default-on `merge_hooks_into_selected_agent` for kiro is fine. No need for an interactive prompt or default-off.

Thanks for the thorough investigation and the strong test coverage.

PR #2395, opened 2026-06-25
fix(kiro): resolve --agent hooks target by name field, not filename

MatthewWolff:
## Description

Kiro CLI resolves `--agent <name>` by the `name` field **inside** each `~/.kiro/agents/*.json`, not by the filename. Generator-managed agents (e.g. tooling that renders agents into the global agents dir) name their files `<prefix>-<name>.json`, so the filename stem and the logical agent name diverge.

AoE's selected-agent hook install assumed `filename == name` and targeted `<name>.json`. For a prefixed agent that path doesn't exist, so AoE created a stray `<name>.json` that was a clone of the standalone `aoe-hooks` agent, while the real agent file (the one Kiro actually loads) never received AoE's status hooks. The result: working/idle/etc. status detection silently failed for any user-selected agent whose file is prefixed. It worked only for the standalone `aoe-hooks` agent, whose filename and name happen to match.

This is the second half of `--agent` support. #2382 made the launch command and the sandbox install honor the selected agent; this PR makes the install actually target the file Kiro loads.

### Fix

- Replace `SelectedAgentHooks::config_subpath: fn(&str) -> PathBuf` with `resolve_config_file: fn(&Path, &str) -> PathBuf`.
- New `resolve_kiro_agent_file(agents_dir, name)` scans the agents directory for a config whose top-level `name` field matches the selection, falling back to `<name>.json` only when no file declares that name (the correct create-path for a brand-new agent).
- Both install paths updated and kept symmetric: host (`Instance::install_sidecar_host_hooks`) and sandbox (`build_container_config`) derive the agents directory from the existing sidecar config subpaths.
- Duplicate `name` declarations resolve deterministically (lexicographically-first) and emit a `warn!` so the misconfiguration is debuggable.
- The host install log now records the resolved path, so a wrong-file target is observable instead of silent.

The agent registry stays data-driven: this is a function pointer on the `AgentDef`, not a per-agent string match, and the existing drift-guard test still asserts kiro is the only agent declaring `selected_agent_hooks`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

`tests/e2e/kiro_launch.rs::test_kiro_agent_hooks_install_into_name_matched_file` seeds a generator-style `<prefix>-<name>.json` agent under the harness's isolated `$HOME`, drives the full `aoe add --launch --tool kiro --extra-args "--agent <name>"` path, and asserts AoE's three status hooks merged into that prefixed file while its own `agentSpawn` hook and `name` are preserved, and that no filename-stem clone is created. Verified (TDD) to fail against the previous filename-stem behavior. Also covered by unit tests for the resolver (name-match wins over stem, fallback, invalid/non-json skipped, deterministic duplicate pick) and an in-crate integration test for the sandbox staging path.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** AI assisted with the investigation (root-causing the filename-vs-name mismatch against a real host), implementation, and tests. The fix was verified end-to-end by running the actual resolver and installer against a snapshot of a real `~/.kiro/agents` directory. I reviewed and understand all of it and will respond to review feedback myself.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784992188&installation_id=139138837&pr_number=2395&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2395&signature=9ca9c7bff4cfdcf31d661e660675ac2b9eeb3215913b94b2b4c621ddd11f045a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This change fixes Kiro agent hook installation so AoE updates the actual agent file Kiro uses, instead of assuming the file name matches the agent name.

What changed:
- Replaced filename-based agent lookup with name-based lookup by reading the agent JSON’s top-level `name` field.
- Updated both the host-side and sandbox-side hook installation flows to use the same resolver logic.
- Removed the old helper that directly constructed `~/.kiro/agents/<name>.json` and replaced it with a resolver that scans the agents directory.
- Added deterministic handling for duplicate agent `name` entries (with a warning), plus clearer logging of the chosen resolved path.
- Expanded test coverage, including sandbox staging behavior and an end-to-end case for generator/prefixed agent files where filename ≠ logical agent name.

Benefits:
- Hooks now land in the correct agent config file that Kiro actually loads.
- Prevents creation of stray/duplicate agent files when the filename is prefixed by a generator.
- Preserves existing user agent configuration fields while still supporting fallback behavior when no matching agent is found.

Potential area for further enhancement:
- Consider caching resolver results per run to avoid repeatedly scanning/parsing agent JSON files when installing multiple hook sets.

### DIFF
--- a/src/agents.rs
+++ b/src/agents.rs
@@ -147,9 +147,14 @@ pub struct SidecarHooks {
 pub struct SelectedAgentHooks {
     /// CLI flag a user passes to choose a named agent (e.g. `"--agent"`).
     pub flag: &'static str,
-    /// Maps a selected agent name to the home-relative path of that agent's
-    /// config file (e.g. `custom-agent` → `.kiro/agents/custom-agent.json`).
-    pub config_subpath: fn(&str) -> std::path::PathBuf,
+    /// Absolute path, under the given agents directory, of the config file the
+    /// CLI actually loads for the selected agent name. The first argument is
+    /// the agents directory to resolve within (host: `$HOME/.kiro/agents`;
+    /// sandbox: the staged `.kiro/sandbox/agents`), the second is the validated
+    /// selected agent name. Resolves by the `name` field inside each config
+    /// rather than the filename, since generator-managed agents name files
+    /// `<prefix>-<name>.json`. See [`crate::hooks::resolve_kiro_agent_file`].
+    pub resolve_config_file: fn(&std::path::Path, &str) -> std::path::PathBuf,
 }
 
 /// On-disk format of a sidecar agent's config file. Drives
@@ -720,7 +725,7 @@ pub const AGENTS: &[AgentDef] = &[
             // agent, and skip the set-default promotion above.
             selected_agent_hooks: Some(SelectedAgentHooks {
                 flag: "--agent",
-                config_subpath: crate::hooks::kiro_agent_file_for,
+                resolve_config_file: crate::hooks::resolve_kiro_agent_file,
             }),
             format: SidecarFormat::KiroJson,
         }),
@@ -1233,9 +1238,12 @@ mod tests {
             .as_ref()
             .expect("kiro declares selected_agent_hooks");
         assert_eq!(sel.flag, "--agent");
+        // With no matching agent file in the dir, the resolver falls back to
+        // `<dir>/<name>.json` (the create-path for a brand-new user agent).
+        let tmp = tempfile::TempDir::new().unwrap();
         assert_eq!(
-            (sel.config_subpath)("custom-agent"),
-            std::path::Path::new(".kiro/agents/custom-agent.json")
+            (sel.resolve_config_file)(tmp.path(), "custom-agent"),
+            tmp.path().join("custom-agent.json")
         );
         // The other sidecar agents do not (their hooks apply globally).
         for name in ["settl", "hermes"] {
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -1788,9 +1788,13 @@ pub fn install_kiro_hooks(agent_config_path: &Path, target: HookInstallTarget) -
 
         let before = config.clone();
 
+        let default_name = agent_config_path
+            .file_stem()
+            .and_then(|s| s.to_str())
+            .unwrap_or(KIRO_HOOKS_AGENT_NAME);
         config
             .entry("name".to_string())
-            .or_insert_with(|| Value::String(KIRO_HOOKS_AGENT_NAME.to_string()));
+            .or_insert_with(|| Value::String(default_name.to_string()));
         config
             .entry("tools".to_string())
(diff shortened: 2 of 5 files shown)


njbrake (approved):

Approving. This is the natural completion of #2382: it fixes a real silent-failure bug where status detection broke for any selected agent whose file is a generator-managed `<prefix>-<name>.json`, because the install assumed `filename == name`.

What I checked:

- The resolver (`resolve_kiro_agent_file` / `find_kiro_agent_file_by_name`) matches by the JSON `name` field, sorts entries for deterministic duplicate resolution, warns on duplicate names, and skips non-JSON/invalid files. Unit tests cover all of these.
- Both install sites derive the agents directory symmetrically: host from the parent of `host_config_subpath`, sandbox from the parent of `sandbox_config_subpath` (and kiro's `copy_dirs` stages `agents` into the sandbox before install). The new serial sandbox test and the e2e launch test both confirm hooks land in the name-matched file with no stem clone.
- The CodeRabbit finding was valid: the original fallback created `<name>.json` but `install_kiro_hooks` defaulted the `name` field to `aoe-hooks`, so by this PR's own premise Kiro would never load it for `--agent <name>`. The latest push fixes it by deriving the default name from the file stem, which recovers the selected name on the fallback path and preserves `aoe-hooks` for the standalone path. Regression assertion added; the standalone kiro tests still pass.
- `cargo fmt --check` clean; hooks, agents, sandbox, and resolver tests pass locally.

One caveat worth stating plainly: the whole change rests on Kiro discovering agents by the `name` field rather than the filename. I did not reproduce that against a live Kiro CLI, but it is corroborated by the existing `kiro-cli agent set-default aoe-hooks` path (which addresses agents by name), and the author verified file discovery by name against a real `~/.kiro/agents` directory.

Issue #2398, opened 2026-06-25
Plugins: implement config.read / config.write host RPCs (cross-plugin/global config, settings write)


Seluj78:
**Note:** drafted by Claude via back-and-forth with @Seluj78. Idea and decisions are his.

Part of #268. Follow-up to #2397 (PR adds the worker-facing `config.get` read of a plugin's own settings, gated by `runtime.worker`).

## Problem

#2397 implements `config.get { key }` for the calling plugin's **own** settings only, gated by `runtime.worker`, because reading your own declared settings needs no `config.*` capability (`aoe-plugin-api/src/capability.rs:60`: "A plugin's own declared settings need no `config.*`").

`config.read` and `config.write` remain in `KNOWN_CAPABILITIES` (`capability.rs:70`) declared-but-unimplemented. Per that comment they mean **host/global or other-plugin** configuration, not the plugin's own table. No host RPC implements them yet.

## Proposal

Implement the host RPCs the two reserved capabilities gate:

- `config.read` -> a method to read host/global or another plugin's configuration (e.g. `config.get { plugin_id, key }` with an explicit target id, refused unless `config.read` is granted; the user sees `config.read` in the grant prompt).
- `config.write` -> a method to mutate the plugin's own settings (`config.set`) and/or host/global config, gated by `config.write`.

Decide the exact method shapes (own-write vs cross-config-write split) when picking this up. Scope, threat model, and per-caller boundaries must be defined before implementing, since these cross the own-table boundary that `config.get` deliberately stays inside.

## Scope

Cross-plugin / global config read (`config.read`) and config write (`config.write`). The own-settings read path already shipped in #2397.

njbrake:
These RPCs landed in #2992 and were then deliberately removed in #3532 along with the abandoned MCP/Skills-plugin approach; `host_api` tests now pin METHOD_NOT_FOUND for them. Re-adding would contradict the core-not-plugin direction tracked in #3043.

_Note: drafted by Claude Fable 5 as part of a backlog triage reviewed with @njbrake. The decisions are his; the prose is Claude's._

PR #2399, opened 2026-06-25
feat(plugin): worker-facing config.get host RPC

### LINKED ISSUE #2397: Plugins: worker-facing config.get host RPC (let a worker read its own settings)
**Note:** drafted by Claude via back-and-forth with @Seluj78. Idea and decisions are his.

Part of #268. Follows the settings registry in #2094 (PR #2389) and the Tier 1 host in #2095 (PR #2387).

## Problem

A plugin can declare a setting (`[[settings]]`, #2094): the host renders an editable control on the TUI/web settings surfaces, validates it, and persists it at `plugins.<id>.settings.<key>`. But the running worker process has no way to read it back, so a declared setting is a dead control from the worker's point of view.

Evidence (origin/main):
- `src/plugin/host_api.rs:138` `dispatch` match arms are exactly `events.publish`, `events.subscribe`, `session.meta.get`, `session.meta.set`, `session.meta.cas`, `sessions.list`, else method-not-found. No `config.*` arm.
- `src/plugin/launch.rs:176` the worker env overlay injects only `AOE_PLUGIN_ID`; settings are not passed through the environment either.
- `aoe-plugin-api/src/capability.rs:70-71` `config.read` and `config.write` are in `KNOWN_CAPABILITIES` (a plugin can request them, the grant prompt lists them) but no host RPC implements them. Declared, unimplemented.

So #2094 landed declaration + rendering + validation + storage; nothing connects a stored setting to a running worker.

## Proposal

Add a worker-facing `config.get` host RPC, gated by the existing `config.read` capability:
- `config.get { key }` -> the value at `plugins.<plugin_id>.settings.<key>` for the calling plugin's own id (scoped, like `session.meta.*` already scopes to the caller). Return null/absent when unset so the worker can fall back to its own default.
- Wire it into `host_api.rs::dispatch` with a `config.read` capability check, mirroring how `session.meta.get` is gated.
- Decide read-only now vs also `config.set` (`config.write` capability) later; read is enough for the motivating use case, keep write as a follow-up unless trivial.
- No new capability string is needed: `config.read` / `config.write` already exist in `KNOWN_CAPABILITIES`.

## Motivating use case

agent-of-empires/plugin-github wants a poll-rate setting for its status segment. With this RPC the worker reads `plugins.agent-of-empires.github.settings.poll_interval_ms` at startup (falling back to an env var, then a 300s default). Without it, a `[[settings]]` poll-rate would render a knob the worker can never observe, so the plugin uses an env var as a stopgap.

## Scope

Read path only (and optionally write). Per-caller scoping is mandatory: a plugin must only read its own `plugins.<id>.settings.*`, never another plugin's. Add host_api tests mirroring the existing `session.meta` capability-gate tests.

Depends on #2094 (settings storage) and #2095 (worker host + host API).

Seluj78:

## Description

A plugin can declare a `[[settings]]` control (#2094): the host renders it, validates it, and persists it at `plugins.<id>.settings.<key>`. Until now the running worker had no way to read that value back, so a declared setting was a dead control from the worker's point of view.

This adds a worker-facing `config.get { key }` host RPC. It returns the value at `plugins.<plugin-id>.settings.<key>` for the **calling plugin's own id**, converting the stored TOML to JSON, and returns null when the plugin has no entry or the key is unset so the worker falls back to its own default. The id is always the caller's own, never a request parameter, so a plugin can only ever read its own table (scoped exactly like `session.meta.*`).

Capability gate: `config.get` rides on `runtime.worker`, the capability every worker already holds to run at all (same as `events.*`). It is deliberately **not** gated by `config.read`. Reading your own declared settings needs no `config.*` capability; per the taxonomy comment at `aoe-plugin-api/src/capability.rs:60`, `config.read` / `config.write` mean host/global or other-plugin configuration, which no host method exposes yet. So nothing new is shown in the install grant prompt for reading your own settings.

`config.read` / `config.write` (cross-plugin / global read, and config write) stay declared-but-reserved and are tracked as a follow-up, #2398.

Read path only; no `config.set` / `config.write` in this PR.

**Base:** Rebased onto `main` now that the Tier 1 worker host (#2095, PR #2387) has merged; this is a single commit on top of current `main`.

Closes #2397

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Install a plugin that declares a `[[settings]]` field and set a value on the TUI/web settings surface (persists to `plugins.<id>.settings.<key>`).
- [ ] From the plugin worker, call `config.get { "key": "<your_key>" }` and confirm it returns `{ "value": <the value you set> }`.
- [ ] Call `config.get` with a key you never set and confirm it returns `{ "value": null }`.
- [ ] Confirm a worker without the `runtime.worker` capability gets a FORBIDDEN error from `config.get`.
- [ ] Confirm one plugin cannot read another plugin's settings (a different `plugin_id` reading the same key gets `null`).

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

(Covered by an in-module `#[cfg(test)]` test in `src/plugin/host_api.rs` mirroring the existing `session.meta` capability-gate test: granted read, unset key returns null, cross-plugin read refused, and missing `runtime.worker` is FORBIDDEN.)

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design call (gate on `runtime.worker`, not `config.read`) and reviewed each commit.

Added a new worker-facing `config.get` host RPC so plugins can read back their own persisted settings after they are saved by the host. The read is limited to the calling plugin’s own settings, returns `null` when a value is missing, and is gated by `runtime.worker` instead of `config.read`.

The docs were updated to reflect the new capability mapping and the per-plugin read scope. Tests were also added to cover successful reads, missing values, cross-plugin isolation, and access denial without the worker capability.

Benefit: plugins can now reliably load their saved defaults without exposing broader config access.

### DIFF
--- a/docs/development/internals/plugin-system.md
+++ b/docs/development/internals/plugin-system.md
@@ -353,6 +353,7 @@ reuse the existing taxonomy:
 | `session.meta.get` | `session.read` |
 | `session.meta.set` / `session.meta.cas` | `session.write` |
 | `sessions.list` | `session.read` |
+| `config.get` | `runtime.worker` |
 
 `events.*` run over a shared plugin event bus (a `plugin_host` schema on the
 durable event-log substrate, `src/events/`); `subscribe { topics, after_seq }`
@@ -363,6 +364,15 @@ plugin's id, so one plugin cannot reach another's data. A `session.meta.cas`
 that loses returns the current value rather than clobbering it. Writes go
 through `Storage`'s cross-process lock, so the daemon picks them up on its next
 session reload (eventual consistency, not a live push).
+
+`config.get { key }` returns the value at `plugins.<plugin-id>.settings.<key>`
+for the calling plugin's own id, so a worker reads back the settings the user
+edited on the TUI/web surfaces, falling back to its own default when the key is
+unset (the call returns null). The id is the caller's own, never a request
+parameter, so a plugin can only read its own table. Reading one's own declared
+settings needs no `config.*` capability: `config.read` / `config.write` gate
+host/global or other-plugin configuration, which no host method exposes yet, so
+`config.get` rides on `runtime.worker` like `events.*`.
 
 ### Sandboxing
 
--- a/src/plugin/host_api.rs
+++ b/src/plugin/host_api.rs
@@ -18,11 +18,17 @@
 //! - `session.meta.set { session_id, key, value }` and
 //!   `session.meta.cas { session_id, key, expected, value }` (`session.write`).
 //! - `sessions.list` (`session.read`).
+//! - `config.get { key }` (`runtime.worker`): the value at
+//!   `plugins.<plugin-id>.settings.<key>` for the calling plugin's own id.
 //!
 //! Per-plugin namespace: session metadata is always read and written under the
-//! calling plugin's own `Instance.plugin_meta[<plugin-id>]` slot. The worker
-//! sends only `key`; it can never name another plugin's id, so one plugin
-//! cannot touch another's metadata.
+//! calling plugin's own `Instance.plugin_meta[<plugin-id>]` slot, and
+//! `config.get` reads only the caller's own `plugins.<plugin-id>.settings`
+//! table. The worker sends only `key`; it can never name another plugin's id,
+//! so one plugin cannot touch another's metadata or settings. Reading one's own
+//! declared settings needs no `config.*` capability (those gate host/global or
+//! other-plugin config); `runtime.worker`, which every worker holds to run at
+//! all, is enough.
 
 use std::sync::Mutex;
 
@@ -159,6 +165,10 @@ pub fn dispatch(
         "sessions.list" => {
             ctx.require(CAP_SESSION_READ)?;
             sessions_list(state)
+        }
+        "config.get" => {
+            ctx.require(CAP_WORKER)?;
+            config_get(ctx, params)
         }
         other => Err(DispatchError::method_not_found(other)),
     }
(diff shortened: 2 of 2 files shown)


njbrake (approved):

Approving. Effective change is small: the first four commits are #2387 (since squash-merged to main), so the real diff is the final commit, `config.get { key }` plus its docs and test (116 lines).

What I verified:

- Cherry-picked the `config.get` commit onto current `main`, built with `--features serve`, and the new test passes. The `missing field validation` compile error against the PR branch as-is is purely #2387 staleness and clears on rebase.
- The load-bearing design call, gating `config.get` on `runtime.worker` rather than `config.read`, is correct. The taxonomy at `aoe-plugin-api/src/capability.rs` states that `config.read` / `config.write` mean host/global or other-plugin config, not the plugin's own settings, so riding on `runtime.worker` is consistent.
- Scoping to the caller's own id is enforced structurally (id comes from the context, never a request param) and asserted by the cross-plugin test.
- Loading the global config fresh per call via `Config::load()` is the right choice, not an inconsistency: plugin settings live in the global config, and reading fresh is what lets a worker pick up a just-edited value.

Non-blocking note: `serde_json::to_value` on a `toml::Value::Datetime` emits the toml-internal `{"$__toml_private_datetime": ...}` shape rather than a clean string. No current `[[settings]]` widget produces a datetime, so this is unreachable today; worth keeping in mind only if a datetime-valued setting type is ever added.

Small, well-scoped, well-tested, and it completes the settings read path that #2389 already shipped. Needs a rebase onto main now that #2387 has landed, which the description already anticipates.

PR #2400, opened 2026-06-25
feat(acp): apply claude-agent-acp >=0.52 native session title push, skip redundant smart_rename one-shot

### LINKED ISSUE #2396: Agents: honor claude-agent-acp >=0.52 native session_info_update title push, skip redundant smart_rename one-shot

## Problem

`claude-agent-acp` v0.52.0 added "Push session title updates at turn end" (PR [agentclientprotocol/claude-agent-acp#812](https://github.com/agentclientprotocol/claude-agent-acp/pull/812)). At every turn-end the agent now pushes the SDK-maintained session title to the client. This is exactly what our `smart_rename` one-shot does, except native, free, recurring, and higher quality.

Right now aoe gets zero benefit from it: the push lands in a catch-all and is dropped on the floor instead of being applied to the session title. Meanwhile we keep spawning a separate `claude -p` subprocess per session to compute a title the agent is already handing us.

This issue is to honor the native push as a progressive enhancement for claude `>=0.52.0`, while keeping `smart_rename` as the fallback for non-claude agents and older claude. No version-floor bump.

### Upstream mechanism (claude-agent-acp v0.52.0)

On every `session_state_changed: idle` (turn-over), the adapter reads the SDK session file and pushes a `session_info_update`, deduped against the last pushed title (PR #812, `src/acp-agent.ts`):

```js
const rawTitle = info?.customTitle ?? info?.summary;
if (!rawTitle) return;
const title = sanitizeTitle(rawTitle);
if (title === session.lastTitle) return;
session.lastTitle = title;
await this.client.sessionUpdate({
  sessionId,
  update: { sessionUpdate: "session_info_update", title, updatedAt: ... },
});
```

`customTitle` is a user-set claude `/rename`; `summary` is the SDK auto-generated title (or first prompt). Note the agent only knows about a claude-side `/rename`, NOT an aoe-side rename done in our TUI or web UI.

### The schema crate already supports it

`agent-client-protocol-schema` 0.12.0 (via the `agent-client-protocol` 0.14 dep, `Cargo.toml:157`) already defines the variant and struct, so the Rust side can deserialize it today:

- `SessionUpdate::SessionInfoUpdate(SessionInfoUpdate)` (`client.rs:107`)
- `struct SessionInfoUpdate { title: MaybeUndefined<String>, updated_at: MaybeUndefined<String>, meta }` (`client.rs:205`)

### We currently drop it

`src/acp/acp_client.rs` matches every `SessionUpdate` variant explicitly, then has a catch-all:

```rust
// src/acp/acp_client.rs:3701-3704
// Variants we don't have a typed mapping for yet pass through as
// RawAgentUpdate so the UI can render best-effort and we can
// narrow these as we go.
other => vec![raw_event(&other)],
```

`SessionInfoUpdate` falls into `other`, becomes a `RawAgentUpdate`, and is never applied to `Instance.title`. The lifecycle-signal classifier also returns `_ => None` for it (`acp_client.rs:615`), so it produces no signal.

### Our own auto-rename: `smart_rename`

`src/session/smart_rename.rs` (whole file) spawns a one-shot `claude -p "<instruction + first user message>"` on the first prompt of a structured session, sanitizes stdout to a 3-to-5 word title, and persists it:

- Trigger: first prompt, detached via `tokio::spawn(try_smart_rename(...))` (`src/server/api/acp.rs`).
- Gates (`check_eligible`, `smart_rename.rs:66-97`): structured, setting on, title is a default civ name (`is_default_civ_name`, `src/session/civilizations.rs:119`), not sandboxed, agent has `oneshot_flag`, command not overridden.
- Fires once per session; persists via `apply_title_only` (`smart_rename.rs:489`) under the storage lock, re-checking eligibility before write.
- Config: `session.smart_rename` (default `true`) and `session.smart_rename_agent` (`src/session/config.rs:193-207`).
- Cross-agent: works for any agent with a `oneshot_flag` (claude `-p`, codex `exec`, opencode `run`, gemini `-p`); cursor/vibe/hermes/pi/kiro have `None` and are skipped.

So `smart_rename` and the native push do the same job. The native one is free (no subprocess, no extra tokens), keeps improving across turns, but is claude-and-`>=0.52`-only.

## Root cause

Two independent facts, no code-level clash yet because we drop the push:

1. The native title push is silently discarded (`acp_client.rs:3704` catch-all), so the feature buys us nothing today.
2. `smart_rename` keeps doing redundant one-shot work for claude sessions that will receive a native title anyway.

The clash appears the moment we wire the push to `Instance.title`: there would be two writers to the title. The native push fires every turn and is blind to aoe-side manual renames, so applying it unconditionally would clobber a title the user set in our TUI/web. It must reuse the same `is_default_civ_name` gate `smart_rename` already respects.

### A. Apply the native push, gated (minimum viable)

In `acp_client.rs`, give `SessionUpdate::SessionInfoUpdate` a typed arm instead of letting it hit the catch-all. On a non-empty `title`, emit a new lifecycle event (e.g. `Event::SessionTitleSuggested { title }`) that the session layer applies through the existing `apply_title_only` path, reusing its eligibility re-check: only overwrite when the current title is still a default civ name (`is_default_civ_name`). This protects aoe-side manual renames the agent can't see. Null `title` (clear) is a no-op for us.

### B. Stop the redundant one-shot for capable claude sessions

When the connected agent is claude and negotiated `>=0.52.0` (we already parse the agent version in `src/acp/agent_compat.rs`), skip the `smart_rename` one-shot spawn and let the native push do the work. Keep `smart_rename` unchanged for:

- non-claude agents (codex/opencode/gemini), and
- claude `<0.52.0`.

This avoids the wasted subprocess and the civ → one-shot-title → native-summary flicker. No floor bump; pre-0.52 claude simply keeps the current behavior.

### C. (Open) surface state in the web sidebar

The web `SmartRenameState` chip (`src/server/api/sessions.rs:600-657`) assumes the one-shot is the only renamer. If we skip the one-shot for capable claude, decide what the chip shows (likely just `Inactive`, since the native push has no "pending" phase). Small follow-on, can land with B.

## Why now

- The feature already shipped upstream; every claude `>=0.52` session is pushing titles we currently throw away.
- Removing the per-session one-shot for claude is a direct token + subprocess saving with no UX loss.
- The schema crate already supports the variant, so the Rust change is small and self-contained.

- As a user on a claude `>=0.52` ACP session still named after a civilization, when the agent finishes a turn and pushes `session_info_update`, then the session title updates in the TUI home list and web sidebar without spawning a `smart_rename` one-shot.
  - Suggested test layer: `web/tests/live/` (apply + sidebar render) and/or `tests/e2e/` driving the structured view with the fake ACP agent emitting `session_info_update`.
- As a user who renamed an ACP session in aoe's own TUI/web UI, when the agent pushes a `session_info_update` at turn-end, then aoe does NOT overwrite my manual title (the `is_default_civ_name` gate holds).
  - Suggested test layer: Rust unit on the apply path (gate precedence: non-default title is never overwritten).
- As a user on a non-claude ACP agent (codex/opencode) or claude `<0.52`, when I send the first prompt, then `smart_rename` still runs exactly as today.
  - Suggested test layer: Rust unit asserting the one-shot is still eligible/spawned for those agents and skipped only for claude `>=0.52`.

## Open questions

- Event shape: reuse a generic "title suggested" event applied by the session layer, or apply inside the ACP client directly? The former keeps the `is_default_civ_name` gate co-located with `smart_rename`'s existing one.
- Version detection point: we negotiate the agent version in `agent_compat.rs`; is that value available where the `smart_rename` spawn decision is made (`src/server/api/acp.rs`), or does it need plumbing?
- Should an aoe-side manual rename also propagate back to claude (so `customTitle` and our title stay in sync)? Probably out of scope; flagged because the agent's `customTitle ?? summary` precedence assumes claude is the only renamer.
- `updatedAt` from the push: do we use it for anything (sort, last-activity), or ignore it?

## Out of scope

- Bumping the `claude-agent-acp` minimum-version floor (`agent_compat.rs:43`). This lands as a progressive enhancement.
- Smart rename for non-ACP / tmux sessions (tracked separately in #2222).
- Removing `smart_rename` entirely. It stays as the fallback.

## Cross-references

- #2222 — Smart rename of non-ACP sessions; complementary (that extends the one-shot to tmux sessions, this leans on the native push for claude ACP sessions).
- Upstream: [claude-agent-acp#812](https://github.com/agentclientprotocol/claude-agent-acp/pull/812), released in [v0.52.0](https://github.com/agentclientprotocol/claude-agent-acp/releases/tag/v0.52.0).

Seluj78:

## Description

claude-agent-acp 0.52 added a feature that pushes the session title via the ACP `session_info_update` notification at the end of every turn (the SDK's own auto-generated summary, or a user `/rename`). AoE was routing that notification through the `SessionUpdate` catch-all and dropping it, so the feature did nothing for us while we kept spending a separate `claude -p` one-shot per session to generate a title.

This PR wires it up:

- The native push now maps to a typed `Event::SessionTitleSuggested`, gets normalized (single line, ANSI stripped, length-capped, civ-name candidates rejected), and is applied to `Instance.title` from the daemon's `acp_event_listener`.
- A new `Instance.last_auto_title` field lets the title keep tracking the agent's summary across turns, while a manual rename (which diverges `title` from `last_auto_title`) freezes it. No provenance enum, no changes to the rename code paths.
- For claude-agent-acp >=0.52 (detected from the initialize handshake, fail-closed) the `smart_rename` one-shot is skipped, so each session has exactly one title writer. Non-claude agents and claude <0.52 keep the one-shot fallback unchanged.
- The shared `apply_auto_title` helper now reports whether the storage write actually happened and mirrors the in-memory instance only then, so storage and memory cannot diverge if two writers ever race.

No version-floor bump (stays 0.49.0). This is purely additive: older agents behave exactly as before.

Closes #2396

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] With claude-agent-acp >=0.52 installed, create a structured-view session and leave its generated civilization name. Send a first prompt, let the turn finish, and confirm the session title in the sidebar/home list updates to the agent's summary (no separate `claude -p` process spawns).
- [ ] Keep working in that session across a few more turns; confirm the title keeps updating to the agent's latest summary.
- [ ] Rename the session yourself (TUI rename dialog or web), then send another prompt; confirm your title is NOT overwritten on the next turn-end.
- [ ] With claude-agent-acp <0.52 (or codex/opencode/gemini), confirm the `smart_rename` one-shot still fires on the first prompt exactly as before.
- [ ] Set `session.smart_rename = false`; confirm neither path renames the session.

## Checklist

## Test Coverage Analysis

No `web/` files change. The session title surfaces in the sidebar, but the title-apply is entirely backend; it is covered by a Rust round-trip test that drives `apply_auto_title` through real `Storage` and asserts persistence, across-turn tracking, and manual-rename protection.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is session-lifecycle-shaped but lives in backend logic, covered by lib-level tests: the version-detection gate (`agent_compat`), title normalization and the overwrite predicate (`smart_rename`), the `SessionUpdate` -> event mapping (`acp_client`), and an end-to-end `apply_auto_title` round-trip through `Storage` (`server`). A full daemon e2e for a backend title write would be disproportionate given that coverage.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction, with a `/debate` pass between Gemini and GPT-5 that flipped the design from "ship the apply, defer the one-shot skip" to folding both together (a single title writer per session avoids a storage/memory race). I made the design calls and reviewed each commit.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2400"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784996217&installation_id=139138837&pr_number=2400&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2400&signature=a6603221cca7054cbe2c4b916a3941b6b2dbad653648cb52c4584dc99633248b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR adds native session title support for compatible `claude-agent-acp` sessions, so the agent’s own end-of-turn title updates are now used instead of being dropped. It also keeps titles stable across turns by tracking the last auto-generated title, while preserving any manual rename made by the user.

For newer Claude adapters, the old one-shot rename flow is skipped to avoid duplicate title writers. Older Claude versions and non-Claude agents continue using the existing fallback behavior.

The update also includes supporting documentation and tests, plus a small event/state plumbing change so the new title signal is handled consistently throughout the app.


njbrake (approved):

Approving. This solves a real, already-shipped-upstream problem (native session titles being dropped while we keep paying for a redundant `claude -p` one-shot), is scoped tightly, and is purely additive with fail-closed version detection.

What I verified locally rather than taking on description:

- **Feature gating.** `normalize_native_title` is `#[cfg(serve)]` and is called from `map_update_to_events`, which looked like it might break the TUI-only build. It does not: `agent-client-protocol` is itself serve-gated, so all the ACP mapping code only compiles under serve. Confirmed with a clean no-feature `cargo check`.
- **First-prompt skip timing.** `native_title.set()` runs before the handshake-ready signal, so by the time a prompt is accepted the `OnceLock` is populated. No double-write window, no race in the one-shot skip decision.
- **Manual-rename freeze.** Traced end to end: civ name to native push records `last_auto_title`; successive pushes track because `title == last_auto_title`; a manual rename diverges the two and freezes it. The freeze holds because only `apply_auto_title` ever writes `last_auto_title`.
- **DoD.** `cargo fmt --check` clean, `cargo clippy --features serve` clean (only a pre-existing, unrelated `dir_guard.rs` warning), the 6 new tests pass, and both the TUI-only and `--features serve` builds compile.

Two non-blocking suggestions, neither a merge condition:

1. `pushes_native_session_title` hardcodes `"@agentclientprotocol/claude-agent-acp"`, which already exists as `expected_name` in the agent policy. Reusing it would single-source the name, though the module's "gate, not the wire name" comment suggests the duplication may be deliberate.
2. The native push has lib-level coverage of each user story (gate, normalize, mapping, plus an `apply_auto_title` round-trip through real `Storage` with manual-rename protection) but no end-to-end wire test driving the fake ACP agent through `session_info_update`. The coverage is solid; an e2e would only add the deserialization hop. Reasonable to defer.

Nice work, and the test coverage of the overwrite predicate is the right thing to nail down here.

PR #2361, opened 2026-06-24
fix(session): publish captured sid to live tmux pane

snvtac:
## Description

Fixes #2356.

This updates captured session ID tmux env publishing so session lifecycle code writes to the live AoE tmux pane for the instance, including terminal and container-terminal panes, instead of reconstructing the default agent session name from the current title.

Changes:
- resolve the env-carrying tmux session by the instance ID suffix across agent, terminal, and container-terminal sessions
- publish both `AOE_INSTANCE_ID` and `AOE_CAPTURED_SESSION_ID` to the resolved live pane
- use the live tmux pane resolver during poller callbacks, sync publishing, and initial TUI env sync
- include terminal/container sessions in the captured-session exclusion scan while still excluding tool sessions
- add tmux-backed regression coverage for terminal-session write-back and exclusion visibility

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [x] For UI changes: included screenshot or recording

No docs or screenshots are needed for this session/tmux env fix.

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: this PR doesn't change a user-facing dashboard flow
- [ ] Added or updated a Playwright spec under `web/tests/` and updated `web/tests/coverage-matrix.json`
- [ ] Skipped a test, because: 

**TUI / CLI (e2e test)**
- [ ] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle
- [ ] Added or updated an e2e test under `tests/e2e/`
- [x] Skipped a test, because: this is tmux env plumbing covered by tmux-backed unit/regression tests rather than a CLI subcommand or rendered TUI flow.

Validation:
- `git diff --check`
- `cargo fmt --all -- --check`
- `CARGO_TARGET_DIR=../cargo-target cargo clippy --workspace --all-targets -- -D warnings`
- `CARGO_TARGET_DIR=../cargo-target cargo test --lib publish_captured_sid -- --nocapture`
- `CARGO_TARGET_DIR=../cargo-target cargo test --lib apply_session_id_updates_publishes_to_terminal_session -- --nocapture`
- `CARGO_TARGET_DIR=../cargo-target cargo test --workspace --lib --bins`

[...]

PR #2280, opened 2026-06-18
fix(session): keep unread rank from reviving sunk rows

jerome-benoit:
## Description
Fixes the Attention sort unread-rank helper so tier 99 rows stay sunk even if a future caller passes them directly to `attention_rank()`. Also moves the helper doc comment onto the helper itself and tightens nearby tier-99 comments to match the actual sunk-row cases.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [x] For UI changes: included screenshot or recording

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: this PR doesn't change a user-facing dashboard flow
- [ ] Added or updated a Playwright spec under `web/tests/` and updated `web/tests/coverage-matrix.json`
- [ ] Skipped a test, because:

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle
- [ ] Added or updated an e2e test under `tests/e2e/`
- [ ] Skipped a test, because:

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Sisyphus, GPT-5.5 via OhMyOpenCode.

**Any Additional AI Details you'd like to share:**

AI assisted with code inspection, the regression test, and PR preparation.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than copy/pasting reviewer comments into an AI and pasting back its answer. We want to discuss with you, not your AI :)

- [x] I am an AI Agent filling out this form (check box if true)

## Testing

- `cargo test test_attention_rank_unread_promoter`
- `cargo fmt --check`
- commit hook: `cargo clippy -- -D warnings`
- commit hook: `cargo fmt -- --check`
- LSP diagnostics on `src/session/groups.rs`

---
[...]

PR #2481, opened 2026-06-27
fix(session): exclude stopped peer sids from Claude resume mtime fallback

### LINKED ISSUE #2355: bug(session): mtime fallback can pick a stopped peer's jsonl when sidecar is stale (remaining half of #2344)
## Summary

PR #2345 closed the case where a *fresh* hook sidecar disagrees with the mtime guess. This issue tracks the remaining case: when the sidecar is absent or stale (older than `SESSION_ID_SIDECAR_MAX_AGE = 5 min`) **and** a stopped peer's jsonl is the freshest in the shared `~/.claude/projects/<encoded-cwd>/` directory, the mtime fallback at `src/session/instance.rs::capture_freshest_session_id` has no authoritative signal and can still pick the peer's id.

## Scope

The fallback calls `try_retroactive_capture` → `capture_claude_session_id` → `scan_claude_project_dir`. The exclusion set in scope (`retroactive_capture_excludes` plus the live tmux-env set built by `build_exclusion_set` at `src/session/capture.rs:758`) is sourced from **live** tmux peers only; a peer whose tmux session has been stopped has no `AOE_CAPTURED_SESSION_ID` env entry, so its sid is fair game for the mtime selector.

## Suggested direction

Widen the exclusion set the resume fallback uses to include session ids owned by stopped/idle peers, read from persisted `sessions.json`. Composition stays in `acquire_session_id` / `capture_freshest_session_id` for the Claude branch only, leaving `claude_poll_fn`'s live-tmux exclusion untouched (the poller already sees live peers via the live tmux scan).

The exclusion union becomes:

1. Existing live exclusion via `compose_exclusion(instance_id, &retroactive_capture_excludes)`.
2. Plus `{ inst.agent_session_id : inst <- Storage::load(), inst.tool == "claude", inst.id != self.id, inst.project_path == self.project_path }` filtered to `inst.status` in `{Stopped, Archived}` or `inst.has_live_tmux_pane() == false`.

## Acceptance

- `acquire_session_id` for a Claude session whose sidecar is absent and whose project_path is shared with stopped peers does not override the stored sid with a peer-owned id.
- Regression test mirroring `mtime_fallback_applies_without_sidecar` (`src/session/instance.rs:~7144`) that plants a stopped peer `Instance` in `sessions.json` and asserts the mtime fallback skips the peer's jsonl.

## Out of scope

- Anything covered by PR #2345 (sidecar-present case).
- The `AOE_CAPTURED_SESSION_ID` tmux write-back bug (separate follow-up).

## References

- Parent: #2344
- Half-fix landed: #2345 (`a4559873`)
- Original verify-on-resume: #2298 (fix for #2291)
- Sidecar reader contract: `src/hooks/status_file.rs:68-83`

jerome-benoit:
## Description

Closes #2355, the last remaining gap of #2344.

When two AoE Claude sessions share a `project_path`, session A runs for a while and is then stopped (its tmux session torn down, its sidecar `/tmp/aoe-hooks-<euid>/<A.id>/session_id` purged or aged out of the freshness window). The user then resumes session B. `acquire_session_id(B)` falls through to `try_retroactive_capture`, the sidecar is absent for B, and the Claude branch's mtime fallback scans `~/.claude/projects/<encoded-cwd>/`. A's jsonl is freshest because B has not produced traffic yet, so the fallback picks A's sid and the resume binds B to A's conversation.

Root cause: the exclusion set passed to that fallback comes from `build_exclusion_set`, which reads peer sids from live tmux env. A stopped peer has no live tmux session, so its sid is not in the set and stays eligible for the mtime selector.

Fix: add `compose_exclusion_with_stopped_peers` in `src/session/capture.rs`. It composes the existing `compose_exclusion` (live tmux peers + cascade `retroactive_capture_excludes`) with the sids of stopped, archived, or pane-less Claude peers in the same `project_path`, read from `sessions.json` via `Storage`. The Claude branch of `try_retroactive_capture` consumes the extended set; every other tool branch keeps the existing live-only `retroactive_capture_exclusion_set()`. `claude_poll_fn` and `build_exclusion_set` are untouched: the poller runs on a hot polling path and live peers already surface in the tmux env scan, so the extended read is restricted to the cold resume-fallback path where the bug appears.

Scope: `~/.claude/projects/` is keyed by `$HOME`, not by AoE profile, but the helper only inspects `sessions.json` for the caller's effective profile, so a stopped peer in a different profile against the same host `$HOME` is not excluded. That residual gap is narrower than #2355's trigger and is left for a follow-up.

PR #2345 closed the dominant case (sidecar fresh, mtime fallback would otherwise misroute). PR #2361 closed the tmux write-back wrong-session-name case. This PR closes the remaining sidecar-absent + stopped-peer-freshest case.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt --check`, `cargo clippy --features serve --tests --all-targets -- -D warnings`, and `cargo test --features serve --lib verify_on_resume` all green. Two new tests pin the fix end-to-end through `acquire_session_id`: `mtime_fallback_skips_stopped_peer_sid` (covers `Status::Stopped`) and `mtime_fallback_skips_archived_peer_sid` (covers `is_archived()`). The sibling tests `mtime_fallback_applies_without_sidecar`, `sidecar_wins_over_fresher_peer_jsonl`, and `sidecar_consulted_for_sandboxed_claude` continue to pass. Round-trip check on the fix: reverting the Claude-branch helper switch in `try_retroactive_capture` makes the stopped-peer test fail with the peer's sid returned; re-applying restores green.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The fix is internal to `acquire_session_id` resume plumbing; behavior is covered by the two new unit tests in `src/session/instance.rs`. No TUI rendering, CLI subcommand, or session lifecycle surface changed.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 4.7 (`claude-opus-4-7`).

The fix was implemented by an AI agent following a human-authored work plan: pinpoint the exclusion-set gap in `src/session/capture.rs::build_exclusion_set`, add a Claude-only extended-exclusion helper that reads `sessions.json` via `Storage`, switch only the Claude branch of `try_retroactive_capture` to use it, and lock the contract with two sibling tests of the existing `mtime_fallback_applies_without_sidecar`. The round-trip regression check (revert helper switch, observe the new test failing with the peer's sid; re-apply, observe green) was run before opening the PR.

This PR fixes a Claude “resume” mis-binding bug when multiple Claude sessions share the same `project_path`. If a peer’s saved sidecar/session id is missing or stale, resuming could incorrectly attach to the stopped peer’s conversation.

Key changes:
- Added a Claude-only helper `compose_exclusion_with_stopped_peers` in `src/session/capture.rs`.
  - It reads `sessions.json` via `Storage`, finds same-`project_path` Claude peer sessions that are stopped, archived, or have no live tmux pane, and adds their `agent_session_id`s to the exclusion set.
- Updated `Instance::try_retroactive_capture` in `src/session/instance.rs` so the wider exclusion is applied only to the Claude path.
  - Other tool branches continue using the existing live-peer exclusion behavior.
- Expanded regression coverage in `verify_on_resume` to prevent the mtime-based fallback from picking the wrong peer:
  - `mtime_fallback_skips_stopped_peer_sid`
  - `mtime_fallback_skips_archived_peer_sid`
  - `mtime_fallback_skips_pane_less_peer_sid` (including the “Status::Idle but pane-less” case)

Benefits:
- Prevents resuming into the wrong Claude conversation when same-project peer sessions exist in stopped/archived/pane-less states.
- Makes the “mtime fallback” path more reliable by avoiding stale peer sidecar ids for Claude.
- Keeps resume behavior for non-Claude tools unchanged.


jerome-benoit:
@coderabbitai two responses:

**Title length / Conventional Commit prefix** — declining. The repo uses scoped \`fix(<area>):\` prefixes extensively on this subsystem. Recent merged PRs on \`src/session/\` include \`fix(session):\` at #2345, #2361, #2369, #2379, #2298, #2280. The 81-character title (subject + auto-appended \` (#2355)\`) is mid-distribution for \`fix(session):\` titles on \`main\` in the recent sample (60-88 chars). Dropping the scope to a bare \`fix:\` and trimming "mtime" would diverge from established subsystem precedent and lose the specific code-path identifier ("mtime fallback" is the exact name used in the helper docstring, inline arm comment, test names, and PR description).

**Pane-less peer regression test** — applied in [\`e96121a1\`](https://github.com/agent-of-empires/agent-of-empires/pull/2481/commits/e96121a18edb4ea40b589c1cac001c0479be4af8). Added \`mtime_fallback_skips_pane_less_peer_sid\` as a third sibling next to \`mtime_fallback_skips_stopped_peer_sid\` and \`mtime_fallback_skips_archived_peer_sid\`. The seeded peer carries default \`Status::Idle\`, is not archived, has an \`agent_session_id\`, and never spawns a tmux pane, so the \`!inst.has_live_tmux_pane()\` branch is exercised on its own. The test reuses \`seed_disk_for_sidecar_test\`, \`ClaudeHomeGuard\`, and \`write_jsonl_with_mtime\` and asserts on \`acquire_session_id\` end-to-end, matching the existing pattern. \`cargo test --features serve --lib mtime_fallback\` reports 4/4 pass.

Seluj78 (approved):

## AoE Review

Reviewed PR #2481 against AoE invariants and the session resume logic. Scope: 2 files (session surface), focused fix for the cold Claude resume-fallback path. Mechanical gates clean (no em dash, no dead-code allow, no new tracing call sites). CI green except Linux test failing at review time.

### Verdict

**Implementation: Approve.** **Project fit: Accept.**

Real bug with a tracked issue (#2355, last gap of #2344), tightly scoped, hot poller deliberately untouched, residual cross-profile gap documented and deferred. Three regression tests pin the fix end-to-end through `acquire_session_id`; codecov reports all modified lines covered. No blockers, no required findings.

### Nits (polish, not gates)

1. `src/session/capture.rs` tests: `mtime_fallback_skips_stopped_peer_sid` and `..._archived_peer_sid` seed a peer that is also pane-less in the test env, so the OR predicate `Stopped || is_archived() || !has_live_tmux_pane()` can pass purely on the pane-less arm. The named branches are exercised end-to-end but not proven in isolation. Low value to fix (would need a real tmux pane in a unit test).

2. `compose_exclusion_with_stopped_peers`: both `let-else` arms on `Storage::new_unwatched` / `load()` drop the error and silently revert to the live-only set (pre-fix behavior) with no diagnostic. A `tracing::debug!(target: "session", ...)` on those paths would make a recurrence diagnosable. Cold path, so low priority.

Issue #2231, opened 2026-06-16
tmux: capture_with_cursor_stays_consistent_under_streaming_load fails deterministically on macOS arm64


jerome-benoit:
## Problem

The test `tmux::session::tests::capture_with_cursor_stays_consistent_under_streaming_load`
at [`src/tmux/session.rs:1043`](src/tmux/session.rs) panics deterministically
on macOS arm64 with `assertion failed: !content.is_empty()` for at
least one of its 30 capture iterations. The neighbouring test
`capture_pane_with_cursor_returns_content_and_cursor` exhibits the
same flake. The file was last touched by #2115 and has not been
modified since.

## Reproducer

On a macOS arm64 host with tmux installed:

```bash
cargo test --lib tmux::session::tests::capture_with_cursor_stays_consistent_under_streaming_load
```

Standalone runs reproduce reliably (3 isolated re-runs all hit the
panic). Behaviour on Linux CI is unknown (the suite is gated on
tmux availability, so CI may skip).

## Why this matters

The failure pollutes the `cargo test --lib` red-vs-green signal for
every contributor running tests on macOS arm64, including PR #2221's
own test verification. Until it is either fixed or quarantined, the
known-failures allowlist for that hardware is `{stream_load,
capture_with_cursor, file_watch_rewire}`, which is large enough that
new regressions can hide in it.

## Hypothesis

The producer in the test (`bash -c 'while true; do echo line-$((i++)); done'`)
streams as fast as the kernel will schedule it, starving tmux's event
loop. On macOS arm64 specifically, tmux occasionally returns a single
empty capture frame when its `display-message` chained probe takes
>50 ms. The 50 ms inner timeout in `capture_pane_with_cursor` then
fires during the busy stream.

## Suggested fix

Pace the producer (`usleep 100`) so it yields the cpu, OR raise the
inner `capture_pane_with_cursor` budget for this test only (e.g. 250
ms), OR add an N=3 retry around the assertion. A positive-control test
that does NOT load the producer would protect against future
regressions that hang forever (vs. the current load-flake).

## Tracking

Surfaced during PR #2221's QA pass; the file itself is unrelated to
the hooks-hardening work.

PR #2568, opened 2026-06-30
fix(tmux): pin pane-base-index 0 in raw new-session tests (#2231)

jerome-benoit:
## Description

`capture_with_cursor_stays_consistent_under_streaming_load` and `capture_pane_with_cursor_returns_content_and_cursor` (`src/tmux/session.rs:1116` and `:1237`) bypass `Session::create_with_size` and invoke `tmux new-session` directly. Production session creators run `append_pane_base_index_args` to pin `pane-base-index 0`; these two tests do not. On hosts whose tmux config sets `pane-base-index 1` (a common dotfile setting), the target form `SESSION:^.0` resolves the first winlink via `^`, then `cmd-find` calls `window_pane_at_index(w, 0)` (`window.c:790-803`, tmux 3.6a) which starts iteration at `pane-base-index` and matches against `idx == 0`. With `pane-base-index 1`, that match never fires and the function returns `NULL`. `capture-pane` (which has no `CMD_FIND_CANFAIL` flag) then errors with `can't find pane: 0`; the chained `display-message ; capture-pane ; display-message` invocation aborts with exit status 1; `capture_pane_with_cursor` returns `("", None)`; the assertion `assert!(!content.is_empty())` panics.

The companion `display-message` calls carry `CMD_FIND_CANFAIL` (`cmd-display-message.c:46`), so target-resolution failures there produce silent degraded output rather than a visible error, masking the underlying pane-index lookup failure from a naive read of stderr.

Production session creators (`Session::create_with_size`, `ToolSession::create_with_size`, `TerminalSession::create_with_size`) all run `append_pane_base_index_args`, so production is unaffected. The existing `test_respawn_dead_pane_revives_dead_pane` already mirrors the production pin via `; set-option pane-base-index 0`; the two failing tests gain the same chain, and the misleading "regardless of base-index" comment on `respawn_dead_pane` is corrected to name the dependence on `append_pane_base_index_args`.

### Empirical reproduction

Standalone bash repro on an isolated tmux socket: without the pin, 100% failure with `can't find pane: 0` exit 1; with the pin chained into `new-session`, 100% success.

### Validation

On a local arm64 host with `setw -g pane-base-index 1` in `~/.config/tmux/tmux.conf`, against an isolated `TMUX_TMPDIR`:

- `capture_with_cursor_stays_consistent_under_streaming_load`: 10 consecutive runs pass.
- `capture_pane_with_cursor_returns_content_and_cursor`: 10 consecutive runs pass.
- Full `tmux::` suite: 212 pass, 0 fail.

### Out of scope

- The two other pre-existing flakes in the allowlist (`file_watch_rewire`, `update_status_reconciles_running_hook_to_waiting_on_claude_approval_prompt`) are unrelated and stay open as their own issues.
- A broader migration to `%N` pane-id targeting was considered and dropped: production already pins `pane-base-index 0` end-to-end, so the canonical `%N` form would touch 20+ call sites and add a `list-panes` fork per session lifecycle without changing behavior.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle. The two failing tests are unit-level `#[serial_test::serial]` tests of `Session::capture_pane_with_cursor`; the fix lives in their setup, not in production code.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Root cause analysis was sourced from upstream tmux 3.6a (`cmd-queue.c`, `cmd-find.c`, `cmd-display-message.c`, `cmd-capture-pane.c`, `window.c`) plus an empirical repro on an isolated tmux socket. An earlier hypothesis blaming a libevent re-entry race between chained subcommands was refuted by reading `cmdq_next` (`cmd-queue.c`), which drains the parsed command list synchronously without yielding to the event loop.

Updated the tmux cursor-capture regression tests to explicitly set `pane-base-index` to `0` when creating their tmux sessions, matching the way production session setup pins this option. This prevents pane lookup failures on hosts configured with a different default pane index.

Also clarified the inline documentation for the `respawn_dead_pane` logic to explain when the `^.0` pane target is valid and why respawn behavior depends on the pane-base-index handling.

Benefits:
- Stabilizes the tests across different tmux configurations
- Makes test session setup consistent with production behavior
- Reduces confusing `capture-pane` “can’t find pane” errors


Seluj78 (approved):

## AoE Review

Test-only change to `src/tmux/session.rs` (+23/-5). Verified against production.

**Correctness.** `append_pane_base_index_args` (`src/tmux/utils.rs:1888`) emits `; set-option -t <target> pane-base-index 0`. Both test additions (`session.rs:1138`, `:1263`) and the existing `test_respawn_dead_pane_revives_dead_pane` use the identical form. Full parity.

Root cause is sound: `window_pane_at_index` reads `pane-base-index` at lookup time (cmd-find), not at pane creation, so pinning it after `new-session` and before `capture-pane` is sufficient without renumbering the pane.

**Gates.** No em dash, no dead code, no new deps, no docs/migration/settings/agent/logging surface touched. Test names `aoe_test_*` with `#[serial_test::serial]` conform.

**CI.** No failing checks (`codecov/patch`, `codecov/project`, `Lint PR title` green; rest in progress). The CodeRabbit `fix(tmux):` title warning is a false positive; AGENTS.md permits `fix:` and the `Lint PR title` check accepts the scoped form.

**PR metadata.** Template filled, PR Type set, AI Usage complete, `closes #2231` present. No issues.

### Verdict

Implementation: Approve
Project Fit: Accept

Real flake on hosts with `pane-base-index 1`, minimal test-only scope mirroring production pinning, accurate root-cause analysis, and a corrected misleading comment on `respawn_dead_pane`. Clean fix, no findings.

PR #2576, opened 2026-06-30
fix: stop orphaning sandbox containers on session delete and create

peteski22:
## Description

Deleting a session with "remove container" sometimes left the Docker container running while the session row disappeared from the list. Container teardown in `perform_deletion` was gated behind `container.exists().unwrap_or(false)`. A transient `docker container inspect` failure (busy/slow daemon) made the probe return false, so `remove()` was skipped, **no error was recorded**, the deletion reported success, and the caller dropped the session record, orphaning a live container with no warning. The probe's failure mode was indistinguishable from "nothing to remove".

This replaces the swallowing existence gate with a direct, idempotent removal:

- `ContainerNotFound` → no-op (idempotent),
- any other failure → surfaced as an error, so the caller keeps the session record (the TUI re-shows it as `Error` and you can retry) instead of silently orphaning the container.

Because all three surfaces funnel through the shared `perform_deletion`, this fixes the **CLI, TUI, and web** delete paths at once.

It also closes the matching **creation-time** leaks, where a just-started container was orphaned if an `on_create` hook failed:

- **TUI** (`creation_poller`): now calls `cleanup_instance` on the sandboxed `on_create` failure path, matching its sibling error paths.
- **CLI** (`cleanup_partial_session`): now tears down the sandbox container (best-effort, idempotent) on partial-create failure.
- **TUI force-remove** of a stuck-deleting session now removes the container instead of explicitly skipping it (dialog wording updated to match).

## Update since initial review

Addressing review feedback (@Seluj78, @jerome-benoit, @coderabbitai):

- **Closed the gate at its source.** `builder::cleanup_instance` (which backs the TUI *and* web create-failure paths and two other TUI sites) no longer probes with `exists().unwrap_or(false)`. All teardown now goes through one idempotent `DockerContainer::teardown` returning a `Teardown { Removed, AlreadyGone, Failed }` outcome; each call site keeps its own policy inline (deletion surfaces `Failed` as an error, best-effort sites `warn`).
- **Correction:** the earlier claim that "the web create path already cleaned up correctly" was imprecise. It never leaked by omission (it did call `cleanup_instance`), but that function inherited the same swallowing gate, so it shared the transient-probe orphan risk. It is fixed now.
- **Runtime-aware not-found detection.** `remove()`/`stop_container()` classified "already gone" via a single Docker-worded `stderr.contains("No such container")`. Podman (lowercase) and Apple Container (`notFound … not found`) don't match it, so an already-gone container on those runtimes returned `RemoveFailed` and un-retriably failed the delete. Not-found markers are now per-runtime and matched case-insensitively, verified against the real stderr captured from `docker`/`container` on macOS.
- **Call-site regression test** now drives `perform_deletion` with an injected teardown failure and asserts the error surfaces (and that teardown runs unconditionally), pinning the contract the doc comment describes.
- Filed a separate issue for the same swallow shape in `worktree_edit.rs` (out of this PR's scope).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

Unit tests (TDD, red-first):
- `classify_removal` / `deletion_messages_for`: failure is surfaced as an error (the regression guard for the orphaned-container symptom), success records the message, an already-gone container is an idempotent no-op.
- `is_not_found` per runtime, fed the **real** stderr captured from `docker`/`container` (plus a representative Podman string), and a genuine-failure string that must stay `Failed` (guards against a false positive re-introducing the orphan).
- `perform_deletion` call-site tests: an injected teardown failure surfaces as an error, and teardown is invoked unconditionally.

- [x] N/A: this PR doesn't change a user-facing dashboard flow (the web delete/create paths are unchanged; the shared `perform_deletion` fix is covered by Rust unit tests)

**TUI / CLI (e2e test)**
- [x] Added or updated tests: the shared teardown/classification logic and the `perform_deletion` call-site contract are unit-tested (see above). The remaining thread/subprocess wiring in `creation_poller`, `cleanup_partial_session`, and `force_remove_session` is not practical to drive without a real container runtime; those branches now route through the tested `teardown` helper rather than bespoke logic, so the untested surface is wiring, not decision logic.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Used to investigate the leak across the CLI/TUI/web surfaces and implement the fix test-first. Findings were verified against a real orphaned container on the author's machine.

---
<a href="https://app.blacksmith.sh/agent-of-empires/codesmith/agent-of-empires/pr/2576"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785429727&installation_id=139138837&pr_number=2576&repository=agent-of-empires%2Fagent-of-empires&return_to=https%3A%2F%2Fgithub.com%2Fagent-of-empires%2Fagent-of-empires%2Fpull%2F2576&signature=832b2c7e86fc45d483f2d279c8f8c5a9443fa5ad351e063e1f6dc21b8f7484fd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## High-level summary
- Fixed session deletion cleanup to remove sandbox containers reliably, eliminating the fragile “check if container exists first” logic that could silently skip cleanup when inspection failed.
- Standardized sandbox teardown during deletion so all delete flows (CLI, TUI, and web) behave consistently: containers are torn down best-effort, “already gone” is treated as success/no-op, and real teardown failures are surfaced instead of being hidden.
- Improved cleanup when session creation fails:
  - If the TUI sandbox `on_create` hook fails, the partially created sandbox state is now cleaned up.
  - If the CLI hits an error during partial session creation, the sandbox container is torn down to prevent leaks.
- Updated the TUI “force remove” behavior to actually tear down the sandbox container (best-effort in the background), while keeping the existing behavior that skips worktree/branch cleanup, and adjusted the confirmation text to match.

## Benefits
- Fewer orphaned sandbox containers and leaked resources.
- More predictable deletion behavior across interfaces, with correct error reporting so failures don’t disappear and sessions can be retried safely.
- Safer failure handling during partial creations, reducing the chance of leaving behind half-built sandbox environments.

## Technical notes (optional)
- Introduced an idempotent teardown model (`Teardown` = Removed / AlreadyGone / Failed) and centralized mapping from teardown outcomes into user-visible messages/errors.
- Added runtime-aware “container not found” detection for Docker/Podman/Apple Container so “already gone” is recognized correctly across runtimes.
- Refactored deletion to use a shared teardown helper (parameterized so tests can inject teardown failures) and updated builder cleanup to always attempt teardown without a prior existence probe.

### DIFF
--- a/src/containers/mod.rs
+++ b/src/containers/mod.rs
@@ -51,6 +51,31 @@ pub fn batch_container_health() -> HashMap<String, bool> {
     map
 }
 
+/// Outcome of an idempotent container teardown.
+#[derive(Debug)]
+pub enum Teardown {
+    /// The container was force-removed.
+    Removed,
+    /// No container existed to remove; the teardown was a no-op.
+    AlreadyGone,
+    /// Removal failed for a reason other than the container being absent.
+    Failed(error::DockerError),
+}
+
+/// Classify a force-remove result into an idempotent teardown outcome.
+///
+/// A `ContainerNotFound` error means there was nothing to remove and maps to
+/// `AlreadyGone`; every other error is a genuine `Failed`. Keeping this
+/// classification separate from I/O lets it be reasoned about and tested
+/// without a live runtime.
+fn classify_removal(result: Result<()>) -> Teardown {
+    match result {
+        Ok(()) => Teardown::Removed,
+        Err(error::DockerError::ContainerNotFound(_)) => Teardown::AlreadyGone,
+        Err(e) => Teardown::Failed(e),
+    }
+}
+
 pub struct DockerContainer {
     pub name: String,
     pub image: String,
@@ -159,6 +184,21 @@ impl DockerContainer {
         }
     }
 
+    /// Force-remove this container, then sweep its named ignore volumes.
+    ///
+    /// Idempotent: a container that is already gone yields
+    /// [`Teardown::AlreadyGone`], not a failure. Named ignore volumes outlive
+    /// the container, so they are swept regardless of the removal outcome.
+    ///
+    /// NOTE: callers must invoke this unconditionally and act on the returned
+    /// outcome; it must never be gated behind a separate existence probe, whose
+    /// transient failure would skip removal and orphan a live container.
+    pub fn teardown(&self, session_id: &str) -> Teardown {
+        let outcome = classify_removal(self.remove(true));
+        self.remove_named_ignore_volumes(session_id);
+        outcome
+    }
+
     pub fn exec_command(&self, options: Option<&str>, cmd: &str) -> String {
         self.runtime.exec_command(&self.name, options, cmd)
     }
@@ -183,6 +223,25 @@ impl DockerContainer {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn classify_ok_is_removed() {
+        assert!(matches!(classify_removal(Ok(())), Teardown::Removed));
+    }
+
+    #[test]
+    fn classify_not_found_is_already_gone() {
+        let r = Err(error::DockerError::ContainerNotFound(
+            "aoe-sandbox-x".into(),
+        ));
+        assert!(matches!(classify_removal(r), Teardown::AlreadyGone));
+    }
+
+    #[test]
+    fn classify_other_error_is_failed() {
+        let r = Err(error::DockerError::RemoveFailed("daemon busy".into()));
+        assert!(matches!(classify_removal(r), Teardown::Failed(_)));
+    }
 
     #[test]
     fn test_container_generate_name_short_id() {
--- a/src/session/deletion.rs
+++ b/src/session/deletion.rs
@@ -35,6 +35,23 @@ pub struct DeletionResult {
 }
 
 pub fn perform_deletion(request: &DeletionRequest) -> DeletionResult {
+    perform_deletion_with(request, |session_id| {
+        DockerContainer::from_session_id(session_id).teardown(session_id)
+    })
+}
+
+/// Core deletion routine, parameterized over how the sandbox container is torn
+/// down so the container-removal contract can be exercised without a live
+/// runtime.
+///
+/// NOTE: when the session is sandboxed and `delete_sandbox` is set, `teardown`
+/// must be invoked unconditionally; it must not be gated behind a separate
+/// existence probe, whose transient failure would skip removal and orphan a
+/// live container.
+fn perform_deletion_with(
+    request: &DeletionRequest,
+    teardown: impl FnOnce(&str) -> crate::containers::Teardown,
+) -> DeletionResult {
     let mut errors = Vec::new();
     let mut messages = Vec::new();
 
@@ -143,12 +160,7 @@ pub fn perform_deletion(request: &DeletionRequest) -> DeletionResult {
     // container processes.
     if request.delete_sandbox && is_sandboxed {
         tracing::debug!(target: "session.delete", session_id = %request.session_id, stage = "container_remove", "perform_deletion: stage");
-        let container = DockerContainer::from_session_id(&request.instance.id);
-        deletion_messages_for(
-            container.teardown(&request.instance.id),
-            &mut messages,
-            &mut errors,
-        );
+        deletion_messages_for(teardown(&request.instance.id), &mut messages, &mut errors);
     }
 
     // Stage 4: worktree cleanup. Container is gone, agent is gone, no
@@ -623,6 +635,63 @@ mod tests {
                 messages.is_empty(),
                 "no spurious 'removed' message when nothing was removed"
             );
+        }
+
+        fn sandboxed_request() -> DeletionRequest {
+            use crate::session::SandboxInfo;
+            let mut instance = create_test_instance();
+            instance.sandbox_info = Some(SandboxInfo {
+                enabled: true,
+                container_id: None,
+                image: "alpine".to_string(),
+                container_name: "aoe-sandbox-calltest".to_string(),
+                extra_env: None,
+                custom_instruction: None,
+                before_start_env: Vec::new(),
+                container_workdir: None,
+            });
+            DeletionRequest {
+                session_id: instance.id.clone(),
+                instance,
+                delete_worktree: false,
+                delete_branch: false,
+                delete_sandbox: true,
+                force_delete: false,
+                detach_hooks: true,
+                keep_scratch: false,
+            }
+        }
+
+        #[test]
+        fn call_site_surfaces_teardown_failure() {
+            // A failed container teardown must surface as an error so the
+            // caller keeps the session record, not silently succeed. Pins the
+            // `perform_deletion` call site, not just the mapping helper.
+            let request = sandboxed_request();
+            let result = perform_deletion_with(&request, |_id| {
+                Teardown::Failed(DockerError::RemoveFailed("daemon busy".into()))
+            });
+            assert!(!result.success, "a teardown failure must fail the deletion");
+            assert!(result.errors.iter().any(|e| e.contains("Container")));
+        }
+
+        #[test]
+        fn call_site_invokes_teardown_unconditionally() {
+            // Guards against re-introducing an existence-probe gate around the
+            // teardown: it must run whenever the session is sandboxed and
+            // delete_sandbox is set.
+            use std::cell::Cell;
+            let request = sandboxed_request();
+            let called = Cell::new(false);
+            let result = perform_deletion_with(&request, |_id| {
+                called.set(true);
+                Teardown::Removed
+            });
+            assert!(
+                called.get(),
+                "teardown must be invoked unconditionally, never gated behind a probe"
+            );
+            assert!(result.success);
         }
     }
 
--- a/src/session/deletion.rs
+++ b/src/session/deletion.rs
@@ -144,10 +144,11 @@ pub fn perform_deletion(request: &DeletionRequest) -> DeletionResult {
     if request.delete_sandbox && is_sandboxed {
         tracing::debug!(target: "session.delete", session_id = %request.session_id, stage = "container_remove", "perform_deletion: stage");
         let container = DockerContainer::from_session_id(&request.instance.id);
-        record_container_removal(|force| container.remove(force), &mut messages, &mut errors);
-        // Remove named ignore volumes even if the container is already gone — volumes created
-        // with volume_ignores_strategy = "named" outlive the container and need explicit cleanup.
-        container.remove_named_ignore_volumes(&request.instance.id);
+        deletion_messages_for(
+            container.teardown(&request.instance.id),
+            &mut messages,
+            &mut errors,
+        );
     }
 
     // Stage 4: worktree cleanup. Container is gone, agent is gone, no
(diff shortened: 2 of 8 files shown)


Seluj78 (changes requested):

Otherwise LGTM. Real bug, tight scope, clean test-first coverage, no AoE-rule violations (no em dash, no dead code, all tracing tagged, module paths valid). One gap keeps me from approving.

## Required

**`src/session/builder.rs:779`** the TUI creation-leak fix routes through `cleanup_instance`, which keeps the exact swallowing existence gate this PR exists to kill.

```
if container.exists().unwrap_or(false) {
```

The new call site is `src/tui/creation_poller.rs:983` -> `builder::cleanup_instance(...)`.

This PR's thesis (commit body + the `record_container_removal` doc comment) is: "callers must invoke removal unconditionally, never gating it behind a separate existence probe; a transient probe failure must not skip removal." The new `creation_poller` sandboxed `on_create` failure path delegates to `cleanup_instance`, whose container teardown is still gated on `container.exists().unwrap_or(false)`. A transient `docker inspect` failure makes `exists()` return false, removal is skipped, and the container is orphaned.

The PR fixes this gate in three sites (`deletion.rs` via `record_container_removal`, `cli/add.rs::cleanup_partial_session`, `operations.rs::force_remove_session`, all direct idempotent `remove(true)` + `ContainerNotFound` no-op), but leaves it in the fourth, which is precisely the path the PR claims to close ("TUI `creation_poller`: now calls `cleanup_instance` ... matching its sibling error paths").

Make `cleanup_instance` use the same direct idempotent removal as the other three sites (ideally reuse a shared helper) so the TUI creation path is not the one remaining surface vulnerable to the flaky-probe orphan this PR targets.

## Verdict

**Project Fit:** Accept

Close that gap and this is an Approve.

jerome-benoit (changes requested):
Independent read landed on the same gap as @Seluj78 (`builder.rs:779` still gated; the new `creation_poller.rs:175` callsite routes through it). Adding three findings beyond his point.

## Required

**1. The PR body's web-create claim does not hold.** The Description says "The **web create** path already cleaned up correctly and is unchanged." `src/server/api/sessions.rs:4270` (on_create failure on the web create path) calls `builder::cleanup_instance(...)`, which is the same swallowing-probe function `@Seluj78` flagged. The web create path is "unchanged" only in the sense the diff does not edit it; the underlying gate is the same. Either fix `cleanup_instance` (preferred, closes `@Seluj78`'s gap and this one in one edit) or correct the body.

**2. The "reuses already-exercised helpers" coverage claim does not hold.** The Test Coverage Analysis says "the TUI/CLI/force-remove wiring reuses already-exercised helpers (`cleanup_instance`, `DockerContainer::remove`)". `cleanup_instance` and `cleanup_partial_session` have **zero `#[test]` references** in the tree (`rg cleanup_instance tests/ src/` finds only definition + callers; same for `cleanup_partial_session`). The `creation_poller` sandboxed on_create-failure branch, the `cleanup_partial_session` container-before-worktree ordering, and the `force_remove_session` sandbox branch are entirely untested. Only `DockerContainer::remove` itself is exercised (Docker-gated, `#[ignore]`).

**3. No test pins the unconditional-invocation contract at the call site.** The `record_container_removal` doc comment is normative: "callers must invoke this unconditionally, never gating it behind a separate existence probe." The 4 new tests assert the helper's internal decision table, but nothing asserts the contract at `perform_deletion`. A regression that re-introduces `if container.exists().unwrap_or(false) { record_container_removal(...) }` passes all 4 new tests. One test that drives `perform_deletion` with a fail-injected `remove()` (closure indirection or trait) and asserts the error surfaces is enough to pin the call-site shape.

## Suggested

**Out-of-PR-scope follow-up worth a separate issue:** `src/session/worktree_edit.rs:85` (`discard_sandbox_container_after_move`) has the same swallowing-probe shape: `match container.exists() { Ok(true) => remove, Ok(false) => {}, Err(_) => warn }`. A transient `docker inspect` failure logs a warning and skips removal entirely. Same bug class as the deletion path, same fix shape. Not introduced by this PR; flagging so it lands on the radar.

**On CodeRabbit's shared-helper suggestion: leave the duplication inline.** Three sites share the `remove(true) → ContainerNotFound-OK → remove_named_ignore_volumes` shape, but the policies differ by call context: `perform_deletion` accumulates into `Vec<String> errors` (authoritative deletion, must surface), `cleanup_partial_session` and `force_remove_session` use `tracing::warn` (best-effort rollback / fallback). A helper parameterised by an outcome-sink enum or closure would hide call-context semantics in a parameter, exactly what AGENTS.md flags ("DUPLICATION > PREMATURE ABSTRACTION"). Two near-copies are within tolerance; the accumulator path is genuinely different shape.

**Tracing target inconsistency inside `force_remove_session`.** The tmux cleanup at `operations.rs:686` uses `target: "session.tmux_cleanup"` while the new container teardown at `:696` uses `target: "session.delete"`. Same closure, same user action, two targets. Pick one.

## Verdict

**Project Fit:** Accept

Close the `cleanup_instance` gate (`@Seluj78`'s point, fixes the web create path simultaneously), correct or remove the two body claims that do not hold, and add the unconditional-invocation regression test. The deletion path itself is solid; the leak hole remaining is on the creation-failure surfaces.

peteski22:
Addressed feedback in a9961cc.

Closed the `exists().unwrap_or(false)` gate in `builder::cleanup_instance` — it now calls a shared, idempotent `DockerContainer::teardown` (force-remove, `ContainerNotFound` → no-op, any other error surfaced). Verifying the finding showed the same `cleanup_instance` also backs the web create path and two other TUI creation sites, so those are fixed too.

That helper also folds the four duplicated teardown blocks into one (the dedup nitpick): deletion surfaces failures as errors to keep the record; builder/cli/force-remove log best-effort. `classify_removal` and the deletion mapping are unit-tested; fmt/clippy clean, suite green. 

Rebased onto main.

Good catch — thanks both. ❤️

src/containers/mod.rs
```diff
@@ -159,6 +184,21 @@ impl DockerContainer {
...
+    ///
+    /// Idempotent: a container that is already gone yields
+    /// [`Teardown::AlreadyGone`], not a failure. Named ignore volumes outlive
+    /// the container, so they are swept regardless of the removal outcome.
+    ///
+    /// NOTE: callers must invoke this unconditionally and act on the returned
+    /// outcome; it must never be gated behind a separate existence probe, whose
+    /// transient failure would skip removal and orphan a live container.
+    pub fn teardown(&self, session_id: &str) -> Teardown {
+        let outcome = classify_removal(self.remove(true));
```
Seluj78, line 197: Required (correctness): this unconditional `remove(-f)` trusts `classify_removal` to map an absent container to `AlreadyGone`, but that only holds for Docker. Not-found detection is the case-sensitive `stderr.contains("No such container")` in `runtime_base.rs:414`. Podman emits lowercase `no such container`; Apple Container emits a different string. On those runtimes an already-gone container returns `RemoveFailed`, so `teardown` returns `Teardown::Failed`.

In `perform_deletion` that becomes a hard error, keeps the session record, and the delete is un-retriable (container never returns). The old `does_container_exist` gate was runtime-aware and skipped cleanly. Make the not-found classification runtime-aware / case-insensitive so the doc invariant at line 189 holds for every runtime.

src/session/deletion.rs
```diff
@@ -144,16 +144,11 @@ pub fn perform_deletion(request: &DeletionRequest) -> DeletionResult {
...
-                errors.push(format!("Container: {}", e));
-            } else {
-                messages.push("Container removed".to_string());
-            }
-        }
-        // Remove named ignore volumes even if the container is already gone — volumes created
-        // with volume_ignores_strategy = "named" outlive the container and need explicit cleanup.
-        container.remove_named_ignore_volumes(&request.instance.id);
+        deletion_messages_for(
+            container.teardown(&request.instance.id),
```
Seluj78: This is the path where the Docker-only idempotency bites: on podman/Apple Container a `teardown` of an already-gone container returns `Teardown::Failed`, `deletion_messages_for` pushes a `Container:` error, and the session record is retained as `Error`. Retrying keeps failing because the container never comes back. See the note on `containers/mod.rs` teardown.

src/containers/mod.rs
```diff
@@ -184,6 +224,25 @@ impl DockerContainer {
 mod tests {
     use super::*;
 
+    #[test]
+    fn classify_ok_is_removed() {
+        assert!(matches!(classify_removal(Ok(())), Teardown::Removed));
+    }
+
+    #[test]
+    fn classify_not_found_is_already_gone() {
```
Seluj78, line 233: Test gap: these `classify_*` tests feed a hand-built `DockerError::ContainerNotFound`, so they only prove the pure classifier, never that a real not-found stderr maps to `ContainerNotFound` per runtime. That mapping (`runtime_base::remove`) is exactly what breaks on podman/apple. Add a case per runtime's not-found phrasing.

Seluj78 (changes requested):

## AoE Review

Reviewed PR #2576 against `main`. CI all green, no unresolved comments. The design is right: routing all four teardown sites through one idempotent helper and surfacing `Failed` so the caller keeps the record fixes CLI/TUI/web at once. One correctness issue blocks it.

### Required

**Idempotency is Docker-only; podman and Apple Container regress.**
`teardown` now calls `remove(-f)` unconditionally and trusts `classify_removal` to map "already gone" to `AlreadyGone`. But not-found detection is a single case-sensitive match in `src/containers/runtime_base.rs:414`, `stderr.contains("No such container")`, which is Docker's exact phrasing. Podman emits lowercase `no such container`; Apple Container emits a different string. On those runtimes an already-absent container returns `RemoveFailed`, so `classify_removal` yields `Teardown::Failed`.

The old code gated on `does_container_exist` (`src/containers/runtime.rs:114`), which is runtime-aware (docker/podman via `inspect` exit status, apple via `logs`) and skipped removal cleanly when the container was absent. This PR turns "container already gone" into a hard error in `perform_deletion`, so the session record is kept and shown as `Error`. Because the container never comes back, the delete is un-retriable on podman/apple; only force-remove escapes. Containers run `-d` without `--rm` (`src/containers/runtime_base.rs:248`), so the trigger is a manually-removed or partial-create-never-started container rather than the common path, but the invariant documented at `src/containers/mod.rs:189` ("a container that is already gone yields `Teardown::AlreadyGone`") is false for two of three supported runtimes.

Fix: make the not-found classification runtime-aware or case-insensitive, covering podman and apple phrasings, and add a test over the stderr-to-error mapping per runtime (the current `classify_*` tests only feed a hand-built `ContainerNotFound`, so they never exercise the layer that breaks).

### Nit

PR checklist "For UI changes: included screenshot" is unchecked while the force-remove confirm dialog copy changed (`src/tui/home/input.rs`). One-line dialog text; not blocking.

### Verdict

Implementation: Request Changes. Project fit: Accept. Fix the cross-runtime not-found classification (and test it) and this lands.

peteski22:
Pushed `419548a0`. Rundown of what's addressed, plus a couple of open questions.

**Podman/Apple not-found (@Seluj78) — fixed.** You were right that the unconditional `remove()` only stayed idempotent on Docker. Not-found detection is now per-runtime (`not_found_markers`, case-insensitive) in `remove()`/`stop_container()`, verified against real stderr on macOS — Docker `No such container`, Apple Container `notFound: "container with ID … not found"`, Podman lowercase `no such container` (representative; not installed locally). A genuine-failure string is tested to stay `Failed` so real errors aren't swallowed.

**@jerome-benoit** — gate fixed at the source (so the web-create path is covered too); the two body claims are corrected; `force_remove_session`'s tracing target is unified to `session.delete`; the call-site regression test now drives `perform_deletion` with an injected teardown failure and asserts it surfaces (and that teardown runs unconditionally); `worktree_edit.rs` filed separately as #2596.

**Open question 1 — the shared helper.** @coderabbitai asked to consolidate the teardown duplication; @jerome-benoit asked to keep it inline (AGENTS.md DUPLICATION > PREMATURE ABSTRACTION). What's there is `DockerContainer::teardown` returning a `Teardown` enum — each call site matches it and keeps its own policy inline (`perform_deletion` pushes to `errors`; builder/cli/force-remove `warn`), so call-context semantics stay at the call site; only the mechanical remove + named-volume sweep is centralized. Acceptable as-is, or would you prefer it fully inlined at the ~4 sites?

**Open question 2 — naming.** While adding `teardown` I noticed `DockerContainer` is already the runtime-agnostic handle (it dispatches to Docker/Podman/Apple Container, since #248), so the name is a bit of a misnomer. I've kept it out of this PR to avoid a 25-reference rename unrelated to the bug — but happy to open a follow-up issue for `DockerContainer` → `Container` if you'd like that cleanup tracked. Want me to?

jerome-benoit (approved):

PR #2090, opened 2026-06-10
feat(plugins): the plugin system: design, host, registries, status plugin, and management UIs

Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Screenshot 2026-06-15 at 18 46 54
Screenshot 2026-06-15 at 18 48 04

## Description

The full plugin system for #268, end to end: design doc, the `aoe-plugin-api` crate, storage groundwork, the substrate lift, manifest/grants/lockfile/install, Tier 0 contribution registries, the Tier 1 worker host with capability middleware, the status-detection reference plugin, runtime web disable with a plugins-disabled CI leg, and management UIs on all three surfaces (CLI, TUI, web).

(#2098, the P9 tail: attention-sort extraction, triage + the exclusive-state primitive, and the web crate extraction, stays open; `plugins/aoe-attention` ships as a manifest stub.)

### What landed, by phase

- **P1**: `docs/development/internals/plugin-system.md` (code-level design, D1-D8) + `aoe-plugin-api` (manifest schema, capability taxonomy, validation; the crate plugin authors compile against).
- **P2**: `Instance.plugin_meta` (namespaced per-session plugin data) + typed `Config.plugins` map. Additive serde, no migration needed.
- **P3**: substrate lift: runner + worker registry to `src/process/`, the event-store pattern generalized into `src/events/` (durable seq log + broadcast + replay, publication-only). No behavior change.
- **P4**: plugin host core: discovery (builtin from `plugins/`, installed under `<app_dir>/plugins/`), capability grants pinned to the manifest sha256 (any change re-prompts and deactivates until re-approved), `plugins.lock`, install/update/uninstall from a GitHub slug or local dir with a one-time capability prompt, `aoe plugin` CLI, trust levels (builtin vs community).
- **P5**: Tier 0 registries. Plugin settings ride the single-source schema pipeline (virtual `plugin:<id>` sections; TUI rows, web schema endpoint, and server PATCH validation all consume one runtime schema), default-override resolution with provenance (`aoe settings explain`, `GET /api/settings/resolved`), keybinds chained after the core table (core always shadows; conflicts inspectable via `aoe plugin info`), CLI commands grafted into the derived clap tree (core wins collisions; `plugin` reserved), theme contributions.
[...]

PR #2618, opened 2026-07-02
fix(acp): recover runner PID via SO_PEERCRED when registry unreadable (closes #2102)

### LINKED ISSUE #2102: ACP: shutdown_and_wait silently skips the PID wait when the registry record fails to load

## Problem

`AcpSupervisor::shutdown_and_wait` (src/acp/supervisor.rs) snapshots the runner PID with `worker_registry::load(session_id).ok().flatten()` before shutting down. When the registry record exists but fails to LOAD (corrupt JSON, IO error), `.ok()` swallows the error, `pid_before` becomes `None`, and the method returns `Ok(())` without waiting for the old runner to die. The follow-up spawn can then race the old runner and collide with a still-live socket.

Pre-existing behavior (predates the #268 substrate lift; the lift only moved the path). The method is documented as best-effort cleanup, so this is a robustness gap, not a regression. Flagged by review on #2090.

## Proposal

Distinguish "no record" (fine, nothing to wait on) from "record unreadable" (propagate or at least wait on the socket path) in `shutdown_and_wait`. One small match on the `Result<Option<_>>` instead of `.ok().flatten()`.

## Out of scope

Any change to the documented best-effort semantics of the switch-agent path beyond honest error propagation.

jerome-benoit:
## Description

Closes #2102.

`AcpSupervisor::shutdown_and_wait` used `.ok().flatten()` to snapshot the runner PID from the on-disk registry, collapsing three outcomes into one silent-skip:

- `Ok(Some(record))` — PID captured, wait runs (existing path).
- `Ok(None)` — nothing was persisted, wait skipped legitimately.
- `Err(e)` — record unreadable at the I/O layer (permissions denied, wrong file type, transient failure) — wait silently skipped, so a follow-up `spawn` for the same `session_id` could race the previous runner (still alive, since `worker_registry::terminate` also silently skipped its `killpg` on the same pattern).

**Scope note**: [`worker_registry::load`](src/acp/worker_registry.rs) coerces `serde_json` parse errors to `Ok(None)` with a `warn` log, so corrupt JSON stays in the "runner already gone" bucket and is *not* covered by this fix. Only true I/O failures reach the `Err` arm — which is what this PR distinguishes.

Pre-existing, not a regression. Flagged in review of #2090.

## Fix

New `worker_registry::pid_source_for(session_id)` explicitly matches on `Result<Option<WorkerRecord>>` and, on `Err`, falls back to reading the runner's PID off the live socket via `SO_PEERCRED` on Linux and `LOCAL_PEERPID` on macOS (new `process::worker::peer_pid_from_socket`). Both `worker_registry::terminate` and `shutdown_and_wait` route through `pid_source_for`, so an unreadable registry record no longer means the runner escapes both SIGTERM and the shutdown wait.

The Ok(None) branch stays a no-op: an absent record means the runner is already gone, so falling through to a peer probe would just fingerprint a stale inode.

## Scope

- `AcpSupervisor::shutdown`, `terminate_runner_for_session`, `worker_registry::load`, and other `.ok().flatten()` callsites elsewhere in the codebase are untouched.
- Best-effort semantics of `shutdown_and_wait` remain: no new `Err` is propagated up to callers from this change; the peer-PID fallback returns `Option<u32>` and degrades to `None` when the socket is missing / not a UDS / peer already gone, matching the pre-existing "on timeout the worker is still removed from the in-memory map" contract at `src/acp/supervisor.rs:945-949`.
- Non-Linux-non-macOS targets already log a `warn!` about the missing PID-wait; the new helper stubs to `None` there, so behavior is unchanged.

## Follow-ups

- **#2621** — `peer_pid_from_socket` uses a synchronous `UnixStream::connect`. Fast in practice (microseconds against a healthy same-host listener) but blocks a tokio worker thread on a wedged runner. Documented in the helper's rustdoc; a non-blocking `connect(2)` + `poll(POLLOUT, timeout)` refactor is tracked separately. Not on the critical path for #2102.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (no user-facing docs affected — internal supervisor plumbing only)
- [x] For UI changes: included screenshot or recording (N/A — no UI)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle. It's an internal supervisor recovery path exercised only when the on-disk worker registry is unreadable.

New unit / integration tests:

- `worker_registry::tests::pid_source_for_prefers_record_pid_when_load_ok_some` — happy-path regression.
- `worker_registry::tests::pid_source_for_returns_none_when_load_ok_none` — guards against a future refactor that folds `Ok(None)` into the peer-probe branch.
- `worker_registry::tests::pid_source_for_falls_back_to_peer_pid_on_load_err` — the #2102 regression: `chmod 0o000` on the record file forces `load() -> Err`, then a UnixListener bound in the test process makes own PID the peer, and `pid_source_for` returns it. Also covers `peer_pid_from_socket` end-to-end.
- `supervisor::tests::shutdown_and_wait_degrades_when_load_errs_without_peer` — same fixture but no bound listener; asserts `shutdown_and_wait` returns fast (`< 100ms`) without panicking or looping.
- `supervisor::tests::shutdown_and_wait_returns_promptly_when_registry_missing` — regression for the `Ok(None)` branch.

The `Ok(Some)` integration branch on `shutdown_and_wait` is deliberately not tested at the supervisor level: since we now route `terminate` through `pid_source_for`, a supervisor test with own PID as the peer would `terminate_process_group(own_pid)` on itself. Coverage lives at the `pid_source_for` unit level plus code review of the trivial `while is_pid_alive` loop.

`worker_registry::tests::with_temp_home` roots the tempdir under `/tmp`, and the supervisor tests use `TempDir::with_prefix_in("aoe-supervisor-", "/tmp")` for the same reason: peer-PID tests stay inside the macOS `sun_path` limit (~104 chars). The default `$TMPDIR` on macOS lives in `/var/folders/...` and blew past the limit once `<app_dir>/acp-workers/<session_id>.sock` was appended.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** claude-opus-4-7 via OpenCode

**Any Additional AI Details you'd like to share:** Investigation, root-cause analysis, implementation, and tests written interactively with AI in the loop; every diff was reviewed and every test was run locally before push. Coordination ping on #2102 (waiting for assignee Seluj78) was also framed with AI assist. Framing and final calls are mine (@jerome-benoit).

**NOTE:**

This PR makes worker shutdown and waiting more reliable when the on-disk worker registry can’t be read.

- The supervisor now distinguishes between “no registry entry exists” and “the registry entry exists but is unreadable,” so an unreadable record won’t cause the shutdown wait to be skipped.
- It adds a PID-resolution fallback that can recover the worker PID from the live Unix socket peer credentials (Linux/macOS), and uses the same logic for both `shutdown_and_wait` and worker termination.
- It updates and expands tests to cover the main registry outcomes (loaded PID, missing entry, and forced load failure) and the socket-based PID fallback.

Benefits:
- Prevents race conditions where a follow-up worker spawn could overlap with a still-running old worker due to skipped waiting.
- Keeps the existing best-effort behavior: it does not introduce new propagated errors and degrades gracefully on unsupported platforms.


Seluj78 (approved):

## AoE Review

Reviewed 3 files (`src/acp/supervisor.rs`, `src/acp/worker_registry.rs`, `src/process/worker.rs`); +290/-13. Verdict: **Approve** / project fit **Accept**.

Clean, correctly-scoped fix for a real bug (#2102, flagged during #2090 review). The three-way `Ok(Some)` / `Ok(None)` / `Err` split is right, and the `Err`-only peer-PID fallback is well-reasoned: parse errors correctly stay in the `Ok(None)` "already gone" bucket, verified against `load()` at `worker_registry.rs:226`. OS-specific code lives correctly in `src/process/worker.rs` behind per-target `cfg`. Test coverage hits every branch, including the `chmod 0o000`-forced `Err` path.

Two things worth calling out as positive:
- The new `(record.pid > 0)` guard prevents `killpg(0, ...)` from signaling the daemon's own process group (a latent self-kill on a zero-pid record).
- The blocking-connect caveat is honestly documented with follow-up #2621.

No new dependencies, no user-facing surface, docs correctly marked N/A.

### Nit
- `worker_registry.rs:536`: recovery via `SO_PEERCRED` logs at `info!`; since an unreadable registry is an abnormal degraded state, `warn!` (matching the no-peer arm below) would read more consistently. Minor.

### CI status
- `CI` and `Tests` workflows are `action_required` (never ran, awaiting maintainer approval for a fork PR), so there is no cargo test / clippy / build / web signal yet. Correctness above is from code inspection only. Approve the workflow runs and confirm green before merge.
- The red `Lint PR title` check is a transient GitHub infra flake (its log is a GitHub error page); a sibling `PR Title Check` run for the same commit passed. The lint config enforces only an allowed type prefix plus a lowercase subject, with no length limit, and `fix(acp): recover...` satisfies both. CodeRabbit's "~70 char limit" warning references a rule that is not in the config. A re-run clears the X; no author action needed.

jerome-benoit:
Addressed in `0221b444`: both arms of the peer-PID resolution now log at `warn!`.

You're right — the two branches surface the same underlying anomaly (registry unreadable), and level-consistency across them beats the "recovered vs failed" distinction I originally reached for. The messages stay distinct so the reader can still tell recovery from orphan-risk, but the severity now matches on both.

CI should re-run on the new HEAD.

Also, thanks for the merge-from-main + the thorough approval review — the two positive callouts (record.pid > 0 guard + #2621 blocking-connect follow-up) exactly captured the fix's non-obvious intent.

PR #2630, opened 2026-07-03
fix(acp): peer_pid_from_socket bounds connect() with non-blocking + poll(POLLOUT, 100ms) (closes #2621)

### LINKED ISSUE #2621: acp: peer_pid_from_socket uses blocking UnixStream::connect on a tokio worker thread
Follow-up from #2102 (M1 flagged during review of #2618).

## Problem

[`process::worker::peer_pid_from_socket`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/process/worker.rs) uses synchronous `std::os::unix::net::UnixStream::connect` with no timeout. Call chain:

```
AcpSupervisor::shutdown (async)
  -> terminate_runner_for_session (sync)
  -> worker_registry::terminate (sync)
  -> worker_registry::pid_source_for (sync)
  -> process::worker::peer_pid_from_socket (sync, blocking connect)
```

The blocking `connect(2)` runs on a tokio worker thread. Same-host UDS connect completes in microseconds against a healthy listener, so this is not a live problem today.

## Why it's worth tracking

The peer-PID fallback fires **precisely when the on-disk record is unreadable** — a symptom already correlating with a runner in a degraded state (D-state kernel thread, accept-loop hung, kernel memory pressure). That's exactly the population where `connect` is most likely to block for tens or hundreds of ms before ECONNREFUSED or EAGAIN.

The current rustdoc caveat (see `peer_pid_from_socket` above) documents the risk but doesn't fix it.

## Suggested fix

Non-blocking `connect(2)` + `poll(POLLOUT, timeout=100ms)` via:

- `nix::sys::socket::{socket, connect, AddressFamily, SockType, SockFlag}` with `SockFlag::SOCK_NONBLOCK`
- `nix::poll::poll(&mut [PollFd::new(&fd, PollFlags::POLLOUT)], 100_i16)`
- On timeout / error, return `None` (same degradation as today).

Approx 15 LOC. No new deps (`nix` already a dep). Bounded, precise, no thread-leak risk.

Update the rustdoc caveat in `peer_pid_from_socket` accordingly once shipped.

## Scope

- `src/process/worker.rs::peer_pid_from_socket` only.
- Callers (`worker_registry::terminate`, `AcpSupervisor::shutdown_and_wait`) unchanged.
- No behavior change on the happy path (healthy listener), only on the wedged-runner edge case.

## Non-goals

- Not fixing the underlying registry-unreadability root causes (permissions drift, disk-full, etc.); those are separate concerns.

jerome-benoit:
## Description

Bounds `process::worker::peer_pid_from_socket` at 100ms by swapping the blocking `std::os::unix::net::UnixStream::connect(path)` for a non-blocking `connect(2)` plus `poll(POLLOUT, 100ms)` plus an `SO_ERROR` sanity check. The Linux/Android and macOS impls now delegate the connect to a shared `connect_with_timeout` helper.

**Why.** The peer-PID lookup is a fallback the shutdown path uses when the on-disk worker record is unreadable, and that population correlates with runners already in a degraded state (hung accept loop, D-state kernel thread, kernel memory pressure), exactly the case where a blocking `connect(2)` on a same-host UDS could stall a tokio worker thread for tens or hundreds of ms before `ECONNREFUSED` / `EAGAIN`. On the happy path (healthy listener) same-host UDS connect still completes in microseconds; behavior is equivalent for the peer-credentials lookup (including `FD_CLOEXEC` preservation, matched via `fcntl` for macOS portability since nix 0.31 gates `SockFlag::SOCK_CLOEXEC` to linux/BSD).

Flagged as M1 during review of #2618 (post-merge). No live user report; latent bug that manifests precisely in the scenarios that are already pathological.

Callers untouched: `worker_registry::terminate` and `AcpSupervisor::shutdown_and_wait` still call `peer_pid_from_socket` with the same signature and the same silent-fail-returns-`None` semantics on any error (timeout, syscall failure, `SO_ERROR != 0`).

Closes #2621.

### Notable design points

- The three platform arms (`linux`, `android`, `macos`) are **structurally symmetric**: all go through `connect_with_timeout` and then `nix::sys::socket::getsockopt` with the platform-specific sockopt (`PeerCredentials` on Linux/Android, `LocalPeerPid` on macOS). No hand-rolled `libc::getsockopt` calls remain; **zero `unsafe` blocks** anywhere in `src/process/worker.rs`.
- **`FD_CLOEXEC` + `O_NONBLOCK`** set via `fcntl` after `socket(2)` because nix 0.31 gates the `SockFlag::SOCK_*` variants to linux_android/BSD. Comment at the call site documents the portability quirk and the behavioral parity with `std::UnixStream::connect`.
- **Deferred-connect match arm folds `Errno::EAGAIN` alongside `EINPROGRESS`**. Linux `unix(7)` documents `EAGAIN` as the AF_UNIX variant of `EINPROGRESS` (connect cannot complete immediately); both take the same POLLOUT wait. `EWOULDBLOCK` is caught implicitly because nix 0.31 aliases it to `EAGAIN` on every target.
- **`SO_ERROR` post-poll check** — `POLLOUT` fires on completed-but-failed connects too (per Linux `poll(2)`), so relying on `POLLOUT` alone would misread `ECONNREFUSED` as success.
- **Docstring** rewrites the previous "Blocking-connect caveat" paragraph to state the new 100ms bound; extends the top-level `None`-cause enumeration to include "connect timeout".

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Three new unit tests in `src/process/worker.rs`, each timing-bounded to `Duration::from_secs(1)` (generous cushion above the 100ms `poll` cap to survive loaded GHA / ASan / TSan CI):

1. `peer_pid_from_socket_missing_path_returns_none_bounded` — path does not exist, returns `None`.
2. `peer_pid_from_socket_non_socket_file_returns_none_bounded` — path is a regular file, returns `None`.
3. `peer_pid_from_socket_healthy_listener_returns_our_pid_bounded` — bind a `UnixListener` in the test process, assert `Some(std::process::id())`.

All existing `process::worker` tests continue to pass (11/11 total).

Gates:

- `cargo fmt --check` clean.
- `cargo clippy --features serve --all-targets -- -D warnings` clean.
- `cargo test --features serve --lib -- process::worker` 11 passed, 0 failed.
- Pre-commit hook (clippy + fmt) green on all three commits.

Cumulative diff: +109 / -31, net **+78 LOC**.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The change is contained to `src/process/worker.rs::peer_pid_from_socket` (a leaf helper called from the shutdown path). No CLI surface, no session lifecycle change, no rendering change. Unit tests in the module cover the three behavioral branches; an e2e would only reach this code via a deliberately corrupted on-disk record, which is not a natural user-facing story.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Direction (non-blocking connect + `poll(POLLOUT, 100ms)` + `SO_ERROR`) was prescribed verbatim in the issue body by @jerome-benoit. The AI implemented the plan across three commits: the initial fix, then two self-review follow-ups running an adversarial Oracle each pass against Rust idiom / harmonization / state-of-the-art / algorithmic-completeness criteria. Improvements folded in across the follow-ups: preserve `FD_CLOEXEC` to match `std::UnixStream::connect` on the resulting fd; drop the `unsafe FromRawFd`/`IntoRawFd` dance via `UnixStream::from(OwnedFd)` (stable since 1.63, MSRV is 1.85); replace the macOS raw `libc::getsockopt(LOCAL_PEERPID)` block with `nix::sockopt::LocalPeerPid` for structural symmetry with the Linux `PeerCredentials` path (eliminating the file's second `unsafe`); widen the test timing cushion to survive CI noise; fold `Errno::EAGAIN` into the deferred-connect arm (Linux AF_UNIX variant of `EINPROGRESS` per `unix(7)`); extend the platform gates to include `target_os = "android"` in line with the issue's own "Linux/Android via `SO_PEERCRED`" wording; harmonize import ordering with the module-level convention; drop a redundant `OwnedFd` type annotation; harmonize the docstring terminology with the file's existing tone.

This PR makes socket-based worker PID lookup more reliable by replacing a blocking connection attempt with a bounded timeout flow, so shutdown paths no longer hang when a worker is unresponsive.

It also broadens platform support and improves consistency:
- adds Android support
- aligns macOS handling with the Unix approach
- preserves socket close-on-exec behavior
- removes unsafe code from the worker lookup path

Tests were added for missing sockets, invalid socket paths, and a successful listener case, all with bounded runtime.

Benefit: faster, safer shutdown behavior with fewer chances of stalling on bad worker state.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ percent-encoding = "2"
 tokio-util = { version = "0.7", features = ["codec", "compat"] }
 
 # Process handling
-nix = { version = "0.31", features = ["signal", "process", "net", "fs", "resource", "dir", "user"] }
+nix = { version = "0.31", features = ["signal", "process", "net", "fs", "resource", "dir", "user", "poll"] }
 
 # Unicode width + normalization (NFKD used to fold accents in title-derived branch names)
 unicode-normalization = "0.1"
--- a/src/process/worker.rs
+++ b/src/process/worker.rs
@@ -37,8 +37,8 @@ pub fn is_pid_alive(_pid: u32) -> bool {
 /// Ask the kernel which process is listening on this Unix domain socket
 /// by connecting and reading the peer's credentials. Returns `Some(pid)`
 /// if the path resolves to a live UDS with a valid peer, `None` otherwise
-/// (path missing, wrong file type, peer already gone, or a target other
-/// than Linux/macOS).
+/// (path missing, wrong file type, peer already gone, connect timeout,
+/// or a target other than Linux/macOS).
 ///
 /// Callers (`worker_registry::terminate`, `shutdown_and_wait`) use this
 /// as the fallback source of the runner PID when the on-disk record is
@@ -51,9 +51,9 @@ pub fn is_pid_alive(_pid: u32) -> bool {
 /// Timeout: connect is bounded at 100ms via non-blocking `connect(2)`
 /// plus `poll(POLLOUT)`. Prevents a wedged runner (D-state kernel
 /// thread, accept loop hung, kernel memory pressure) from stalling the
-/// calling tokio worker thread. On timeout or a completed-but-failed
-/// connect (`SO_ERROR != 0`) the function returns `None`, same silent
-/// fail semantics as any other failure mode.
+/// calling tokio worker thread. Timeouts and completed-but-failed
+/// connects (`SO_ERROR != 0`) fall into the same `None` bucket as any
+/// other failure.
 ///
 /// See #2102, #2621.
 #[cfg(target_os = "linux")]
@@ -67,23 +67,10 @@ pub fn peer_pid_from_socket(path: &Path) -> Option<u32> {
 
 #[cfg(target_os = "macos")]
 pub fn peer_pid_from_socket(path: &Path) -> Option<u32> {
-    use std::os::fd::AsRawFd;
+    use nix::sys::socket::{getsockopt, sockopt::LocalPeerPid};
     let stream = connect_with_timeout(path)?;
-    let mut pid: libc::pid_t = 0;
-    let mut len = std::mem::size_of::<libc::pid_t>() as libc::socklen_t;
-    // SAFETY: `stream` keeps the fd valid for the duration of the syscall.
-    // `pid` is aligned and sized for `pid_t`; `len` is initialized to
-    // `sizeof(pid_t)`, which is what LOCAL_PEERPID writes on success.
-    let rc = unsafe {
-        libc::getsockopt(
-            stream.as_raw_fd(),
-            libc::SOL_LOCAL,
-            libc::LOCAL_PEERPID,
-            std::ptr::addr_of_mut!(pid).cast::<libc::c_void>(),
-            &mut len,
-        )
-    };
-    (rc == 0 && pid > 0).then_some(pid as u32)
+    let pid = getsockopt(&stream, LocalPeerPid).ok()?;
+    (pid > 0).then_some(pid as u32)
 }
 
 #[cfg(not(any(target_os = "linux", target_os = "macos")))]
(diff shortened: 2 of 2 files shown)


Seluj78 (approved):

## AoE Review

Reviewed PR #2630 against AGENTS.md and the aoe-review checklist (mechanical gates, protocol compliance, correctness/security, CI status, PR metadata).

Change: peer_pid_from_socket swaps blocking UnixStream::connect for non-blocking connect + poll(POLLOUT, 100ms) + SO_ERROR check, via a shared connect_with_timeout helper for the Linux/Android and macOS arms.

Checks run:
- Em dash, dead-code, git-config, tracing-target, secret-log probes: clean.
- CI: all 27 checks pass, including Clippy, Cargo Test (linux + macos), codecov/patch, codecov/project.
- No unresolved review threads or actionable CodeRabbit comments.
- PR title/branch follow convention; PR template fully filled.

Correctness walkthrough: FD_CLOEXEC preserved via fcntl for macOS parity with std's connect; EINPROGRESS/EAGAIN both correctly routed to the POLLOUT wait per unix(7); SO_ERROR checked after poll to avoid the POLLOUT-on-ECONNREFUSED trap; OwnedFd lifetime is clean on every path (no leak, no double close). Checked whether poll's lack of EINTR retry (vs std's connect) is a new regression: it is not, this repo's std UnixStream::connect does not loop-retry EINTR either, so the exposure window is equivalent, just split across two syscalls instead of one.

No Blocker / Required / Followup / Nit findings.

Verdict: Implementation Approve, Project Fit Accept. Scoped fix for a previously flagged latent stall (#2621, M1 from #2618 review), removes the file's remaining unsafe blocks as a side effect, tests cover the three behavioral branches, CI green.

PR #2639, opened 2026-07-04
fix: preserve structured sessions across state rewrites

SYU8384:
## Description

Fixes structured-view sessions being rewritten as terminal-shaped rows by writers that did not preserve structured-session metadata.

Root cause: persisted `Instance` fields for structured view were gated behind the `serve` feature, and older or non-serve writers could round-trip `sessions.json` without `view`, `agent_name`, `agent_model`, or `acp_session_id`. Once that happened, the daemon and UI treated a live ACP session as a terminal session, even though a live ACP worker still existed.

This PR:
- persists structured-session metadata in every build
- keeps `is_structured()` tied to the persisted `view` field
- repairs terminal-shaped rows from live ACP worker registry records when a stored ACP session id exists
- rejects ACP worker spawn attempts for terminal-shaped sessions
- documents that Codex currently does not advertise a Plan-mode channel, so AoE does not render a Codex Plan switch

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the regression is covered at the session serialization and daemon reload repair seams; the full serve suite also exercises structured-view state handling.

## Validation

- `cargo fmt --check`
- `cargo test test_instance_acp_acp_session_id_roundtrip`
- `cargo test --features serve repair_structured_rows_from_live_workers_restores_terminal_shaped_row`
- `cargo clippy --features serve`
- `cargo test --features serve`
- Manual release-namespace retest: started the patched `dev-release` daemon against existing release state and confirmed it repaired damaged live ACP sessions back to `view = "structured"` with restored `acp_session_id`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
OpenAI Codex

Used to diagnose the structured-session persistence bug, implement the fix, and run validation.

**NOTE:**

This PR makes structured-view (ACP) sessions more reliable by preserving the right session metadata across storage and reloads, so the daemon and UI consistently recognize real structured sessions instead of treating them like “terminal-shaped” rows.

What changed:
- Structured-view session fields (view, agent name/model, ACP session id) are now persisted in all builds, not only when `serve` is enabled—preventing older/non-`serve` writers from dropping key info.
- Session classification is now based on the persisted `view` value, so “structured vs terminal” remains stable through `sessions.json` round-trips.
- On disk reload, if a stored row looks terminal-shaped but matches a live ACP worker record (via stored ACP session id), it’s repaired back into structured form. If persisting the fix fails, in-memory repairs are rolled back.
- The ACP spawn API now validates the session state early:
  - returns clear errors for unknown sessions
  - rejects spawn attempts for terminal-shaped sessions with `409` (`not_structured`) to avoid incorrect worker/container setup
- ACP disable/spawn lifecycle was hardened with earlier and longer-held per-session locking to avoid race conditions; related tests were expanded to confirm behavior for missing session ids.
- Stored ACP session ids are normalized so empty strings are treated as “unset”.
- Non-`serve` UI behavior improves: structured-fork attempts now show a friendly message instead of failing unexpectedly.
- Documentation + tests were updated, including a web test that asserts the `409 not_structured` precondition, and a note clarifying that Codex doesn’t advertise a Plan-mode channel (so AoE won’t render a Codex Plan switch).

Benefits:
- Prevents live structured ACP sessions from being misinterpreted after restart/reload—especially for sessions created by older or non-`serve` binaries.
- Safer, clearer ACP lifecycle handling via explicit spawn preconditions and improved locking.
- More robust recovery from previously persisted/incorrect session records, reducing user confusion.

Potential follow-ups:
- Consider a lightweight one-time migration/diagnostic for existing `sessions.json` files created by older builds, so repairs happen deterministically and users get clearer guidance when issues were detected.
- Audit other ACP/session-changing endpoints to ensure they consistently enforce the same “structured-view required” rule (and return similarly actionable errors).


SYU8384:
@coderabbitai review

src/server/mod.rs
```diff
@@ -2320,10 +2320,17 @@ pub(crate) async fn reload_state_instances_from_disk(
         merged.push(row);
     }
 
+    #[cfg(feature = "serve")]
+    let repairs = repair_structured_rows_from_live_workers(&mut merged);
```
Seluj78: Required: repair_structured_rows_from_live_workers runs blocking filesystem I/O (directory scan + per-file read/parse + per-record PID/socket checks) while holding the state.instances write lock, which 122 other call sites across the daemon depend on. This runs on every ~2s status-poll tick and every debounced disk-change event. Move the registry scan before state.instances.write().await is acquired, matching how persist_structured_row_repairs already defers its I/O until after drop(current).

Seluj78 (changes requested):

## AoE Review

Reviewed: PR #2639 (base: main@16369747). Files changed: 5 (server, session, tui, docs-user).
Unresolved comments: none.

Root-cause fix is sound: un-gating the structured-session fields from `serve` is the correct fix for writers silently downgrading live ACP sessions, and the repair-from-live-worker-registry mechanism plus the new `spawn_acp` guard are well-targeted defensive layers with a real regression test (`repair_structured_rows_from_live_workers_restores_terminal_shaped_row`).

### Required

- `src/server/mod.rs:2293-2333` - `repair_structured_rows_from_live_workers` runs blocking filesystem I/O (directory scan, per-file read/parse, per-record PID/socket checks in `worker_registry::list()`/`is_record_live`) while holding the `state.instances` write lock. That lock is read/written at 122 other call sites across the daemon and this function runs on every ~2s status-poll tick and every debounced disk-change event. This is the only `worker_registry::list()` call site that runs while holding that lock; the sibling calls in `acp_reconciler.rs` run standalone, and this PR's own `persist_structured_row_repairs` correctly defers its I/O until after `drop(current)`. Fix: move the registry scan before `state.instances.write().await` is acquired, mirroring the pattern already used for the persist step.

### Nit

- `src/tui/home/input.rs:1623` - comment "the field only exists under `serve`" is now stale since this PR un-gates the structured fields from `Instance`; only the local variable extraction stays serve-gated, not the field.

### Verdict

**Project Fit:** Accept

The Required finding is a hot-path lock-contention regression, not a design problem, and is straightforward to fix without touching the fix's substance.

src/server/mod.rs
SYU8384: Addressed in df8f58ed. The live ACP worker-registry scan now runs inside the producer-side `spawn_blocking` reload work before `state.instances` is acquired. `reload_state_instances_from_disk` receives the preloaded live worker records, so `repair_structured_rows_from_live_workers` only does the in-memory merge while holding the write lock.

Validated with `cargo fmt -- --check`, `cargo clippy --features serve -- -D warnings`, `cargo test --features serve`, `cargo test --features serve repair_structured_rows_from_live_workers_restores_terminal_shaped_row`, and `cargo test --features serve --test serve_disk_reload_helper_equivalence`.

src/session/instance.rs
```diff
@@ -1664,17 +1652,11 @@ impl Instance {
...
-        #[cfg(feature = "serve")]
-        {
-            self.view == View::Structured
-        }
-        #[cfg(not(feature = "serve"))]
-        {
-            false
-        }
+        self.view == View::Structured
     }
```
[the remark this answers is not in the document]
SYU8384: Addressed in df8f58ed. `Instance::is_structured()` was already unconditional, and this follow-up removes the remaining `serve` gates from the tmux-facing structured short-circuits: launch, resume fallback, pane-ready, and tmux status update. I also un-gated the focused structured guard tests so the non-serve build covers the behavior.

Validated with `cargo fmt -- --check`, `cargo clippy --features serve -- -D warnings`, `cargo test test_ensure_pane_ready_bails_on_structured`, `cargo test start_with_size_opts_returns_skipped_for_structured`, and `cargo test --features serve`.

jerome-benoit (changes requested):

## AoE Review

Reviewed: PR #2639 (base: main@af427278, head: df8f58ed). Files changed: 5 (server, session, tui, acp handler, docs-user).
Unresolved comments: Seluj78's Required (lock-held blocking I/O) and the CodeRabbit non-serve short-circuits are both closed by df8f58ed.

Root-cause fix is sound. Un-gating `view`, `agent_name`, `agent_model`, `acp_session_id`, `import_pending`, `fork_pending` from `serve` is the correct fix for writers dropping structured metadata, and the repair-from-live-worker path plus the `spawn_acp` guard are well-scoped defensive layers with a real regression test. Two correctness issues remain (the new guard exposes a race the PR itself introduces, and the repair persist path diverges from the file's own awaited-with-rollback template), and the test surface hides the new input from every current and future caller.

### Required

- `src/server/api/acp.rs:283-296` and `src/server/api/acp.rs:1776-1899`. `spawn_acp` reads the instance at :283, drops the guard at :286, checks `is_structured()` on the clone at :287, then acquires `state.instance_lock(&id)` at :330 and calls `state.acp_supervisor.spawn(...)` at :369-380. `acp_disable` never acquires `state.instance_lock(&id)` anywhere in the function. The `workers.lock()` / `pending_resumes` interaction in `src/acp/supervisor.rs:1232` and `:2258` closes the mid-handshake half of the race, but does not close this ordering:
  1. `spawn_acp` passes the :287 guard on a Structured snapshot, drops the read guard.
  2. `acp_disable` runs to completion: `shutdown_and_delete` finds no worker yet, `view` flips to `Terminal` in `state.instances`, `Storage::update` persists `view=Terminal`.
  3. `spawn_acp` acquires `instance_lock` (uncontended, `acp_disable` never held it), reaches `begin_resume`, finds workers empty and no `cancelled_resumes` marker, reserves and spawns a worker.
  4. Post-condition: live ACP worker on a session whose disk and memory row both say `view=Terminal`.
  5. Next reload's `repair_structured_rows_from_live_workers` at `src/server/mod.rs:2394-2413` finds the fresh worker record with a `stored_acp_session_id` and flips `view` back to `Structured`, silently reverting the user's `acp_disable`.

  Fix: re-check `is_structured()` inside the `instance_lock` critical section in `spawn_acp` (drop the pre-lock guard entirely, or keep it as a fast-path and re-verify under the lock before `supervisor.spawn`). Same lock should be taken at the top of `acp_disable` around the shutdown, flip, and persist trio so the two handlers are mutually exclusive per session.

- `src/server/mod.rs:2429-2478`. `persist_structured_row_repairs` uses bare `tokio::spawn` and never awaits the join handle. The house pattern for "persist a change discovered mid-reconcile" is at `src/server/acp_reconciler.rs:598-637`: `spawn_blocking(storage.update(...)).await`, then on persist failure roll back the in-memory mark and log with `tracing::warn!`. Order there is deliberate: `mark`, `persist`, `shutdown` with rollback if persist fails, so `state.instances` and disk stay converged. Here the order is inverted: in-memory mutation at `src/server/mod.rs:2331` happens before the detached persist, and a persist failure leaves memory `Structured` while disk stays whatever the offending writer left. The 2s reload retries, but the shape drifts from the file's own reconciler template and leaves nothing counted by shutdown supervision. Fix: match the reconciler shape. `.await` the `spawn_blocking(storage.update)`, and on `Err` revert `inst.view` back to `Terminal` in `state.instances` for the failed ids so the loop stays reconvergent. Route through `crate::task_util::spawn_supervised` (used at `src/server/mod.rs:2524`) so a panic or shutdown mid-persist is observed.

- `src/server/mod.rs:4417-4435`. `reload_disk_only_for_test` and `reload_tmux_applied_for_test` pass `Vec::new()` unconditionally for `live_worker_records`, hiding the new input from every test author. The signature is dishonest: any test written through these helpers exercises a code path where repair is a compile-time no-op. Fix: add a `live_worker_records: Vec<WorkerRecord>` parameter to both test helpers and update the four call sites in `tests/serve_disk_reload_helper_equivalence.rs` and the two in-file callers at `src/server/mod.rs:5041, 5101` to pass `Vec::new()` at the leaf. Explicit empty at the caller is honest; silent empty at the helper is not.

- No end-to-end test covers the new `spawn_acp` 409 branch. `git grep "not_structured"` on the PR head returns zero matches outside `src/server/api/acp.rs:289`. Existing Playwright specs at `web/tests/live/acp-config-pickers.spec.ts:21` accept `[200, 202, 409]` but never seed a Terminal-view session and assert the `not_structured` body. Fix: add one live-Playwright case in `web/tests/live/acp-view-switch.spec.ts` (which already seeds a Terminal session and toggles enable/disable). POST to `/api/sessions/{id}/acp/spawn` before the enable call, assert `409` and `{"error":"not_structured"}`. Add a coverage-matrix entry per the `AGENTS.md` mandate.

### Suggested

- Empty-string `stored_acp_session_id` leaks through the repair source filter. `src/acp/worker_registry.rs:329` `update_stored_acp_session_id` accepts `Some("")` unchecked; `live_structured_worker_records` at `src/server/mod.rs:2378` filters only on `is_some()`; the repair loop at `:2401` then writes `Some(String::new())` into `inst.acp_session_id`. Downstream, `fork_from_selection` at `src/tui/home/input.rs:2765` already defensively re-filters with `.filter(|s| !s.is_empty())`, which confirms the callers were working around this. Fix: normalize `Some("")` to `None` inside `update_stored_acp_session_id`, and add `!s.is_empty()` to the filter at `:2378`. Belt-and-braces at the source and the boundary.

- `src/server/mod.rs:2374-2375`. `live_structured_worker_records` swallows `worker_registry::list()` errors via `.unwrap_or_default()`. A permissions failure on `workers_dir()` silently disables the entire repair mechanism with zero diagnostic, which is the exact scenario where "structured rows keep getting downgraded" would surface without a log. `disk_watcher_consumer` at `:2746-2751` `tracing::warn!`s the parallel `load_all_instances` failure. Match that shape: `Err(e) => { tracing::warn!(target: "server.file_watch", error = %e, "worker registry list failed; structured row repair disabled this tick"); return Vec::new(); }`. Note the target choice below.

- `src/server/mod.rs:2413, 2469, 2473` introduces a new tracing target `server.reload`. Sibling reload-adjacent logs at `:2746-2751` and `:2760-2765` use `server.file_watch`. Since the repair fires from inside `reload_state_instances_from_disk` which the file-watch consumer drives, reusing `server.file_watch` avoids fragmenting log filters and matches the surface. If a distinct target is intentional, add one row to the target table in `docs/development/file-watch-migration-server.md` alongside `server.file_watch` explaining the split.

- `src/server/mod.rs:2436` uses raw `tokio::spawn` rather than `crate::task_util::spawn_supervised` used at `:2524` for the disk-watch forwarder. A panic inside the storage-update closure is swallowed. Combined with the Required rollback fix above, adopt `spawn_supervised("server.reload.persist_repairs", PanicPolicy::Log, ...)` so panics and shutdown are observed.

- `src/server/mod.rs:2436-2478` never polls `state.shutdown.cancelled()` inside the per-profile loop. On daemon shutdown the task keeps waiting on flocks until each profile completes, adding latency to graceful exit. Fix: check `state.shutdown.is_cancelled()` at the top of the `for (profile, repairs) in by_profile` loop.

- `src/session/instance.rs:1654-1657`. `is_structured()`'s new doc-comment reads "The persisted field exists in every build so non-serve writers preserve structured sessions instead of downgrading them to terminal rows." The un-gating alone only covers future writes; already-in-the-wild `sessions.json` files damaged by pre-PR non-serve writers are healed by `repair_structured_rows_from_live_workers` at `src/server/mod.rs:2382`. Reword to reflect the two-layer story, for example: "True when this session renders in the structured (ACP) view. The persisted `view` field exists in every build so non-serve writers round-trip it intact; rows damaged by pre-fix writers are healed on reload by `server::repair_structured_rows_from_live_workers`."

- `src/tui/home/input.rs:2719`. The changed comment reads "captured ACP session id rides along in serve-enabled builds." After this PR the `acp_session_id` field on `Instance` exists in every build; what stays serve-gated is the local variable extraction at `:2734-2735`. Suggest: "the captured ACP session id is read into a local only under `serve`, since only the serve-cfg'd fork branch consumes it."

- `src/server/mod.rs:4707-4772`. The new test `repair_structured_rows_from_live_workers_restores_terminal_shaped_row` drives the pure helper directly, so it misses: (a) the persist `tokio::spawn` side (untested, and item 2 in Required affects exactly this path), (b) `stored_acp_session_id = Some("")` (see the leak-through Suggested above), (c) `agent_key = ""` fallback where `inst.agent_name` stays `None`, (d) `inst.agent_name.is_some()` skip that must preserve an existing partial-repair. Fix: split into four focused tests. The empty-string case becomes redundant once the source filter is fixed, but the assertion is worth keeping as a regression guard.

- Rust polish, `src/server/mod.rs:2359-2478`. Four small items in the same span:
  - `:2359` `#[derive(Debug, Clone)]` on `StructuredRowRepair` is unused (`grep` of the struct shows no `{:?}` or `.clone()` invocation). Drop both.
  - `:2453, 2456, 2458` clone `repair.agent_name`, `repair.agent_model`, `repair.acp_session_id` from a borrowed `repair`. Iterate `for repair in repairs` (owned) at `:2449` and move each field. `Storage::update`'s bound is `FnOnce` at `src/session/storage.rs:514-517`, so this is trait-legal.
  - `:2378` and `:2401`: the source filter proves `stored_acp_session_id` is `Some`, and the repair loop re-destructures it via `let Some(...) else { continue; }`. Return `Vec<(WorkerRecord, String)>` from `live_structured_worker_records` via `filter_map` so the caller consumes the pair directly. Removes one clone and one dead `continue`. Pairs with the empty-string fix above.
  - `:2373-2413` re-spells `crate::acp::worker_registry::` five times. A function-scoped `use crate::acp::worker_registry::{self, WorkerRecord, is_record_live};` compresses the block to `worker_registry::list()`, `WorkerRecord`, `is_record_live`. Matches the file's existing `use crate::session::Instance;` shape at `:44`.

- `src/server/mod.rs:2429` and `:2382-2413`. Two silent skips worth surfacing:
  - `:2456-2464` if the session was deleted between the reload snapshot and the persist flock, `find` returns `None` and `Storage::update` still atomic-writes sessions.json, firing a spurious file-watch notify. Add `tracing::debug!(target: "server.file_watch", session = %repair.session_id, "repair target not found on disk; skipping")` in the `else` branch.
  - `:2408` unconditionally overwrites `inst.acp_session_id` with the registry value. If a disk row entered the repair path (`!is_structured()`) with a stale non-None `acp_session_id`, the mismatch is silently overwritten. Before the assignment, if `inst.acp_session_id.is_some() && inst.acp_session_id.as_deref() != Some(acp_session_id.as_str())`, emit `tracing::warn!` with `disk = ?inst.acp_session_id, registry = %acp_session_id`.

### Nit

- `src/tui/home/input.rs:2778-2783`. The new bare-core dialog says `"This session uses structured view. Rebuild with the serve feature to fork it."` Three phrasings for the same condition coexist in this file: the "Serve unavailable" dialog at `:2984-2995` says `"built without the web dashboard (a --no-default-features source build)"` with bullet-list install methods, and the activate and tab-attach paths at `:3347` and `:3433` say `"rebuild with default features to attach"`. "The serve feature" is Cargo-feature terminology unlikely to be meaningful to a user, and users who installed via `install.sh` or Homebrew have nothing to rebuild. Fix: mirror the `:2984` structure. Title `"Fork not available in this build"`, body starting `"This aoe binary was built without the web dashboard..."`, then the two install bullets (GitHub Releases release build, or `cargo build --release`).

- Terminology drift in the TUI. `src/tui/home/input.rs:2759` says `"cannot fork a structured session"`; `:2780` says `"This session uses structured view."` Public docs (`docs/structured-view.md`) canonicalize `"structured view"`. Suggest normalizing both to `"structured view session"` (or one consistent choice) alongside the copy fix above.

- PR body coins two terms that do not appear anywhere in the code or docs: `"structured-shaped rows"` and `"structured-session metadata"`. The shipping code uses `"structured session row"` (`src/server/mod.rs:2368`, `:2416`) and `"structured row repair"` (`:2469`, `:2474`); `"terminal-shaped"` appears only in the test name at `:4707`. Consider aligning the PR body (and, if kept, the test name) to `"structured session row"` for one consistent term.

- `src/server/mod.rs:2368-2371` and `:2429`. The two new functions carry 4-line and 0-line preambles. The module's INVARIANTS block at `:2217-2247` already covers cross-cutting rules, so a full mini-INVARIANTS at each helper duplicates. A one-line comment on `persist_structured_row_repairs` explaining why the write is spawned off-thread (must not block the reload path holding `state.instances`) is enough; `repair_structured_rows_from_live_workers` could add "runs before `apply_acp_overlay_inplace` so freshly-repaired rows join the overlay filter's `is_structured()` gate".

- `docs/structured-view.md:50`. The Codex plan-mode caveat is appended to an already-long adapter-limitations paragraph. The feature matrix at `:39-48` already has a Codex "Mode picker" cell reading `depends`; a footnote hanging off that cell (or a per-cell asterisk indexed below the table) makes the caveat discoverable from the matrix rather than buried mid-paragraph. Keep-as-is is defensible.

### Verdict

**Project Fit:** Accept

The un-gating and the two defensive layers are the right shape. Blockers are the `spawn_acp` and `acp_disable` TOCTOU (silent revert of a user action within one reconciler tick), the fire-and-forget persist that diverges from the file's own awaited-with-rollback template, the test-helper signature that hides the new input from every future test author, and the missing end-to-end assertion on the guard's 409 branch. Suggested items close the empty-string leak at its `worker_registry.rs` source, restore observability (IO warn, shutdown-token, silent skips, tracing target), tighten prose to reflect the repair backstop, and eliminate a small pile of avoidable allocations and repeated module paths. Nits are copy and doc polish.

src/server/api/acp.rs
```diff
@@ -284,6 +284,16 @@ pub async fn spawn_acp(
         return (StatusCode::NOT_FOUND, "session not found").into_response();
     };
     drop(instances);
+    if !instance.is_structured() {
```
Seluj78: Required: this new 409 guard has no test. The terminal-shaped-row to 409 not_structured branch is untested; the repair logic and serialization are covered but this error path is not. The mocked web spec only exercises the happy respawn path and never reaches this guard. AGENTS.md Testing: add a one-line assertion that a terminal-view session yields 409.

src/server/mod.rs
```diff
@@ -2347,6 +2355,130 @@ fn apply_acp_overlay_inplace(prior_by_id: &PriorById, merged: &mut [Instance]) {
     }
 }
 
+#[cfg(feature = "serve")]
```
Seluj78, line 2374: Nit: Debug and Clone on StructuredRowRepair are unused. The struct is moved into by_profile, never cloned or debug-printed. Drop the derives per the no-dead-code rule.

Seluj78 (changes requested):

## AoE Review

Reviewed at df8f58ed. CI 27/27 green. Files: server, session, tui, docs.

**Verdict: Implementation Approve, Project Fit Accept.** Posting as Request Changes to track the one open Required (missing test) before merge.

Real regression fixed at the right seams: un-gate the persisted structured fields so every build round-trips them, drive `is_structured()` off the persisted `view`, repair live-but-downgraded rows from the worker registry, and reject ACP respawn on terminal rows. The field un-gating is serde-additive (`default` + `skip_serializing_if`), so it is backward and forward compatible and needs no migration.

The earlier objection about blocking filesystem I/O under the `state.instances` write lock is resolved by follow-up commit df8f58ed: `live_structured_worker_records()` now runs inside `spawn_blocking` at both call sites before the lock, `repair_structured_rows_from_live_workers` does only in-memory work under the lock, and `persist_structured_row_repairs` runs after the lock is dropped.

### Required

- `src/server/api/acp.rs:287` new 409 guard has no test. The terminal-shaped-row to 409 `not_structured` branch is untested; repair logic and serialization are covered but this error path is not. The mocked web spec only exercises the happy respawn path and never reaches the guard. AGENTS.md Testing: add a one-line assertion that a terminal-view session yields 409.

### Nit

- `src/server/mod.rs:2358` `Debug` and `Clone` on `StructuredRowRepair` are unused (the struct is moved into `by_profile`, never cloned or debug-printed). Drop the derives per the no-dead-code rule.

SYU8384:
Addressed the remaining change requests in ec9286ae.

What changed:
- Closed the `spawn_acp` / `acp_disable` TOCTOU by taking the per-session `instance_lock` before the authoritative structured-view check and by holding the same lock through disable shutdown, memory flip, disk persist, and tmux restart.
- Reworked structured-row repair persistence to run in a supervised task, await each blocking storage update, use the existing `server.file_watch` target, observe shutdown, and roll back repaired in-memory rows to terminal if disk persistence fails.
- Made live worker repair inputs explicit in the reload test helpers, so tests now pass `Vec::new()` at the call site instead of hiding the new input in the helper.
- Added the missing live Playwright assertion for terminal sessions: `/api/sessions/{id}/acp/spawn` now gets a JSON `{}` body and asserts `409` with `error: "not_structured"` before enable.
- Normalized empty stored ACP session ids at the worker-registry source and filtered them at the repair boundary.
- Cleaned up the stale structured-view comments and the non-serve fork dialog copy.

Validation:
- `cargo fmt --check`
- `cargo test test_instance_acp_acp_session_id_roundtrip`
- `cargo test --features serve repair_structured_rows_from_live_workers`
- `cargo test --features serve empty_stored_acp_session_id_normalizes_to_none`
- `cargo test --features serve --test serve_disk_reload_helper_equivalence`
- `cd web && node tests/validate-coverage-matrix.mjs`
- `cd web && AOE_E2E_BINARY=../target/debug/aoe npx playwright test --config=playwright.live.config.ts tests/live/acp-view-switch.spec.ts`
- `cd web && npm run format:check`
- `cd web && npm run lint`
- `cd web && npx tsc -b`
- `cargo clippy --features serve`
- `cargo test --features serve`

Note: the first local Playwright attempt used the stale `target/release/aoe` fallback, which was a TUI-only binary and failed with `unrecognized subcommand 'serve'`. The passing run used the serve-enabled debug binary through the supported `AOE_E2E_BINARY` path.

PR #2650, opened 2026-07-04
fix(session): preserve sandboxed opencode.db across launches; only skip_entries needed for host-drift protection (closes #2605)

### LINKED ISSUE #2605: bug(session): sandboxed opencode --session resume fails; opencode.db wiped on every launch by unconditional clean_files
## Symptom

Sandboxed opencode sessions cannot be resumed after any kill or daemon restart. `aoe resume` calls `opencode --session <id>` against a SQLite database wiped on the previous launch, opencode prints `Session not found` and exits 1. Session context is lost with no recovery path.

## Root cause

`src/session/container_config.rs:99-110` marks the opencode SQLite files as both `skip_entries` (do not copy from host) and `clean_files` (delete before every launch):

```rust
AgentConfigMount {
    tool_name: "opencode",
    host_rel: ".local/share/opencode",
    // ...
    skip_entries: &["sandbox", "opencode.db", "opencode.db-wal", "opencode.db-shm"],
    // ...
    clean_files: &["opencode.db", "opencode.db-wal", "opencode.db-shm"],
},
```

`prepare_sandbox_dir` at `src/session/container_config.rs:687-703` walks `clean_files` unconditionally on every invocation:

```rust
fn prepare_sandbox_dir(mount: &AgentConfigMount, home: &Path) -> Result<PathBuf> {
    let sandbox_dir = home.join(mount.host_rel).join(SANDBOX_SUBDIR);
    for &name in mount.clean_files {
        let path = sandbox_dir.join(name);
        if path.exists() {
            if let Err(e) = std::fs::remove_file(&path) {
                tracing::warn!(target: "session.profile", ...);
            }
        }
    }
    // ...
}
```

Called at both `src/session/container_config.rs:984` (agent-config refresh) and `:1433` (per-tool sandbox setup); every sandbox launch wipes the database.

The comment at line 100-102 justifies the wipe against drizzle migration failures when the sandboxed opencode binary differs from the host or from a previous run. That concern is real, but the current mitigation deletes the database in every case, including when the schema is unchanged.

## Reproduction

Requires Docker (or Podman) available:

1. `aoe add --cmd opencode --sandbox docker`
2. Send a prompt in the created session; wait for opencode to respond
3. `aoe kill <session>` (or restart the daemon)
4. `aoe resume <session>`

Expected: session resumes with the prior conversation intact.
Actual: opencode exits with `Session not found` on stderr; the tmux pane shows the crash.

## Impact

Every user running opencode inside a Docker or Podman sandbox loses session history on any kill or daemon-restart cycle. Sandbox is a shipping feature, opencode is a shipping default agent.

## Proposed fix direction

Version-gate the wipe on schema mismatch instead of unconditional deletion:

- Read `PRAGMA user_version` from the sandbox `opencode.db` before wiping and compare against a version fingerprint derived from the sandboxed opencode binary (hash of `opencode --version` output or a stored fingerprint file). Only wipe on mismatch.
- Alternatively, read `__drizzle_migrations` before wiping; if the migration id set matches the sandboxed opencode's expected set, keep the DB. Wipe on mismatch.

`skip_entries` for `opencode.db` on host-to-sandbox copy should stay (host DB may be schema-drifted vs sandbox opencode). Only `clean_files` needs to become conditional.

## Related

Surfaced during an opencode integration audit against sst/opencode v1.17.12+6697cf3. No prior tracking issue on this defect.

---

_AI usage disclosure: investigation and prose drafted with Claude (Opus 4.7). Findings checked by direct code reading against `c85f481d`. Reproduction steps, decisions, and framing are mine (@jerome-benoit)._

jerome-benoit:
## Description

Closes #2605.

### What

`opencode.db{,-wal,-shm}` removed from the opencode `AgentConfigMount.clean_files` at [src/session/container_config.rs](src/session/container_config.rs). Entries remain in `skip_entries` so a schema-drifted host DB is still never copied into the sandbox. The comment above the mount block is rewritten to explain the invariant: `skip_entries` handles pollution, the sandbox owns its DB, subsequent launches must not wipe it.

### Why

`clean_files` was defense-in-depth that turned destructive in practice. `prepare_sandbox_dir` walks `clean_files` unconditionally on every invocation, called at both `container_config.rs:987` (agent-config refresh) and `:1436` (per-tool sandbox setup), meaning every kill/restart of a sandboxed opencode session destroys the SQLite DB that holds session identity. `aoe resume` then calls `opencode --session <id>` against an empty DB, opencode exits 1 with `Session not found`, and session context is lost with no recovery path.

The original justifying comment mentioned drizzle migration failures against a schema-drifted host DB, but `skip_entries` already blocks host-to-sandbox copy for those three files. The sandbox creates its own DB in-container on first launch, which is always schema-consistent with the sandboxed opencode binary that wrote it. Wiping on subsequent launches was never necessary for drift protection and destroyed the resume source of truth.

Reviewer @hiSandog raised this scoping direction in the issue thread: "*cleanup should probably target only volatile subdirectories, not the SQLite DB that holds session identity*". This PR implements that scope.

### PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

### Checklist

- [x] Documentation was updated where necessary (comment above the mount block explains the new invariant)
- [ ] For UI changes: included screenshot or recording (n/a)

### How Tested

- **Drift-guard unit test** `test_opencode_mount_clean_files_does_not_touch_sqlite_db`: asserts `opencode.db*` stays out of `clean_files` and remains in `skip_entries`. Fails immediately if a future contributor re-adds the destructive entries.
- **Behavioral unit test** `test_opencode_mount_preserves_sqlite_db_across_prepares`: plants `opencode.db{,-wal,-shm}` in the sandbox subdir, invokes `prepare_sandbox_dir(mount, home)`, and asserts byte-level survival.
- **E2e test** `sandboxed_opencode_db_survives_prepare_sandbox_cycle` at [tests/e2e/opencode_sandbox_resume.rs](tests/e2e/opencode_sandbox_resume.rs): Docker-gated `#[ignore]` per the sandbox e2e convention. Installs an `opencode` PATH shim and runs `aoe add --sandbox --cmd opencode` twice with a fake DB planted between calls; passing `--cmd opencode` makes `config_tool == "opencode"` so the filter at `container_config.rs:1432` selects the opencode mounts and the test actually exercises the code path #2605 lives in.

Gates:
- `cargo fmt --check` clean
- `cargo clippy --features serve --all-targets -- -D warnings` clean
- `cargo test --lib --features serve session::container_config`: 88 passed / 0 failed (both new tests included)
- `cargo check --features serve,e2e-tests --tests --test e2e` clean

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (`tests/e2e/opencode_sandbox_resume.rs`, Docker-gated `#[ignore]`)

## AI Usage

- [ ] No AI was used
- [x] AI was used

Diagnosis, scoping, and implementation were assisted by claude-opus-4-7 driven by a human plan. The direction fork (β: remove `opencode.db*` from `clean_files` vs. α: version-gate wipe on schema mismatch) was resolved by @hiSandog's community design guidance on the issue thread; the AI implemented β and did not invent the direction. An Oracle cross-validation pass surfaced that the e2e test would only exercise the pre-fix bug when `config_tool == "opencode"`, which was addressed by installing an `opencode` PATH shim and passing `--cmd opencode`.

This change preserves `opencode`’s sandbox-owned SQLite session database (`opencode.db` plus its `-wal`/`-shm` files) across sandbox refreshes, so resumed sessions keep working after a kill or daemon restart. Sandbox setup no longer deletes these DB files on every launch, while still preventing a potentially schema-drifted host DB from being copied into the sandbox. Related mount/config comments and `clean_files` documentation were updated to make the “never use cleanup for sandbox-owned session state” rule explicit.

It adds regression coverage:
- a unit “drift guard” checking that the `opencode` mount does *not* include the DB files in `clean_files` and instead protects them via `skip_entries`,
- a unit behavior test confirming the DB files and their contents survive `prepare_sandbox_dir`/refresh,
- and a Docker-gated end-to-end resume test (currently ignored) that exercises the session resume flow.

Benefits:
- fixes broken `opencode` session resume after restart,
- keeps valid sandbox session history intact across launches,
- reduces the risk of accidentally wiping a working local session state due to host/sandbox differences.

### DIFF
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -54,8 +54,10 @@ struct AgentConfigMount {
     /// authentication from being overwritten by stale host copies.
     preserve_files: &'static [&'static str],
     /// Files to delete from the sandbox dir before each launch. Prevents stale state
-    /// (e.g. SQLite databases from a previous opencode version) from causing failures
-    /// when the container image is updated.
+    /// (e.g. leftover lock/cache files) from causing failures when the container image
+    /// is updated. Do NOT use this for sandbox-owned session state (e.g. opencode's
+    /// SQLite DB, see #2605): that must survive relaunches, so it is kept out of
+    /// `clean_files` and instead protected from host drift via `skip_entries`.
     clean_files: &'static [&'static str],
 }
 
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -92,10 +92,13 @@ const AGENT_CONFIG_MOUNTS: &[AgentConfigMount] = &[
         tool_name: "opencode",
         host_rel: ".local/share/opencode",
         container_suffix: ".local/share/opencode",
-        // Never copy or keep the SQLite database in the sandbox. Opencode must
-        // create its own fresh database on each launch -- a stale db from a
-        // previous opencode version (or copied from the host) causes drizzle
-        // migration failures.
+        // `skip_entries` prevents copying the host DB into the sandbox; a
+        // schema-drifted host DB would trigger drizzle migration failures
+        // against the sandboxed opencode. The sandbox creates its own DB
+        // in-container on first launch, which is always schema-consistent
+        // with that opencode. Do NOT wipe it on subsequent launches: it
+        // holds `ses_*` session identity and is the resume source of truth.
+        // See #2605.
         skip_entries: &[
             "sandbox",
             "opencode.db",
@@ -107,7 +110,7 @@ const AGENT_CONFIG_MOUNTS: &[AgentConfigMount] = &[
         keychain_credential: None,
         home_seed_files: &[],
         preserve_files: &[],
-        clean_files: &["opencode.db", "opencode.db-wal", "opencode.db-shm"],
+        clean_files: &[],
     },
     AgentConfigMount {
         tool_name: "opencode",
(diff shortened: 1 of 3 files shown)


jerome-benoit:
Thanks @coderabbitai — responding to both nits:

**Nit 1 (`clean_files` field doc stale)** — applied in be2d56fb. Genuine catch: this was the exact anti-pattern that caused #2605 in the first place (a misleading example on the mount-level comment lured a prior contributor into adding the destructive entries). Three self-review passes on my end missed it because the audit scope was the mount block, not the struct-field docstring 40 lines above. New wording drops the SQLite-DB example, generalizes to lock/cache files, and adds an explicit anti-pattern warning that back-references #2605 and the `skip_entries`-owns-DB invariant.

**Nit 2 (line-number refs in e2e comments drift)** — skipping. The concern is real in principle, but the codebase already uses line-number refs in comments (e.g. `container_config.rs:63`, `:1429`, and multiple test comments referencing `:2305`), and the original issue plan itself used `:984` / `:1433`. Swapping this one file to symbol refs while the rest keeps line numbers would introduce a local inconsistency, and the referenced predicate (`.filter(|m| m.tool_name == config_tool)`) does not have a named symbol to point to. If a future refactor moves that filter, the comment drift is trivially fixable at that time. Happy to revisit if maintainers prefer symbol refs project-wide.

Seluj78 (approved):

## AoE Review

**Implementation:** Approve
**Project Fit:** Accept

Tight, correct fix for a real resume-breaking bug (#2605), maintainer-ratified direction (@hiSandog). `clean_files -> &[]` while `skip_entries` retains the three `opencode.db*` entries: host DB still never copied in, sandbox-owned DB no longer wiped on every relaunch. Doc and comment rewrites remove the exact anti-pattern that caused the bug. Coverage is proportional: drift-guard unit test, byte-survival unit test, Docker-gated e2e. CI green (27/27).

### Nit

**PR title** is ~135 chars with a trailing `; only skip_entries needed for host-drift protection` clause. Conventional-commit prefix is correct, but the clause bloats it past readable length and the `Closes #2605` already lives in the body. Suggested: `fix(session): preserve sandboxed opencode.db across launches (#2605)`. Corroborates CodeRabbit pre-merge title-length warning.

### Trade-off (not blocking)

Emptying `clean_files` also drops the container-image-update reset the struct doc originally described. If a future container bumps opencode past a non-forward-migratable schema, an old sandbox DB could hit a drizzle migration failure instead of being wiped. The PR leans on drizzle forward-migration owning that path, which is the designed behavior and the lesser evil versus destroying session identity on every kill. Acknowledged in the issue thread; accept as-is.

PR #2602, opened 2026-07-01
fix(session): honor repo-local overrides in smart-rename indicator + gate (closes #2351)

jerome-benoit:
## Description

Two smart-rename call sites resolved config via `profile_config::resolve_config_or_warn`, so a repo-local `.agent-of-empires/config.toml` overriding `smart_rename_agent` or `agent_command_override` was silently ignored on both:

- Runtime gate: [`try_smart_rename`](src/session/smart_rename.rs) in `src/session/smart_rename.rs`
- Sidebar indicator overlay: [`cfg_cache`](src/server/api/sessions.rs) in `list_sessions` (was keyed on `inst.source_profile` only)

Fix extracts a shared helper:

```rust
// src/session/smart_rename.rs
#[derive(Debug, Clone)]
pub struct SmartRenameConfig { pub setting_on: bool, pub rename_agent: String, pub overrides: HashMap<String, String> }
pub fn resolve_smart_rename_config(profile: &str, project_path: &Path) -> SmartRenameConfig
```

Both call sites route through the helper, resolving via `repo_config::resolve_config_with_repo_or_warn`. The sidebar `cfg_cache` is re-keyed on `(profile, project_path)` to match the adjacent [`acp_cmd_cache`](src/server/api/sessions.rs:551) pattern, so runtime gate and indicator resolve identically against the same repo-aware config and cannot drift.

Closes #2351. Follow-up to the profile-only resolver limitation surfaced in the #2350 walkthrough.

### What changed

- `src/session/smart_rename.rs`: new `SmartRenameConfig` struct + `resolve_smart_rename_config` function at unconditional module level, next to `check_eligible_resolved` (its natural peer). Both are `pub` and documented. Struct derives `Debug`, `Clone` for parity with `SmartRenameState` and `SkipReason`.
- `src/session/smart_rename.rs`: `try_smart_rename` calls the helper.
- `src/server/api/sessions.rs`: sidebar overlay imports `SmartRenameConfig` + `resolve_smart_rename_config`, drops the previous local tuple alias, calls the helper. Named fields on the cached value prevent a future edit from swapping positional slots silently.
- `check_eligible_resolved` is unchanged: it was already the shared entry point.

### Test

`resolve_smart_rename_config_honors_repo_local_overrides` in `src/session/smart_rename.rs`:

[...]

Issue #2351, opened 2026-06-23
Smart rename: honor repo-local config overrides (badge + runtime gate) via repo-aware resolver


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Smart rename resolves its config (`session.smart_rename`, `session.smart_rename_agent`, `session.agent_command_override`) via the profile-only `resolve_config_or_warn`, in both the runtime gate (`try_smart_rename` in `src/session/smart_rename.rs`) and the sidebar `pending`-badge overlay (`src/server/api/sessions.rs`). Repo-local overrides (sparse JSON merged via `resolve_config_with_repo_or_warn`, keyed by `(profile, project_path)`) are therefore ignored for smart rename, unlike the custom-agent ACP-capability overlay right below it, which is repo-aware.

The badge and the gate are currently consistent with each other (both profile-only), so there is no user-visible drift today. But a repo that overrides `smart_rename_agent` or `agent_command_override` will not have that override honored by smart rename, and making only one of the two surfaces repo-aware would introduce drift. Flagged by CodeRabbit on #2350.

## Proposal

Make smart rename repo-override-aware on both surfaces together:

- In `try_smart_rename`, resolve config with `resolve_config_with_repo_or_warn(&profile, project_path)` (the instance already carries `project_path`).
- In the sidebar overlay, key `cfg_cache` by `(profile, project_path)` and use the same repo-aware resolver, matching the adjacent ACP-capability overlay.

Keep the two in lockstep so the badge never disagrees with the gate.

## Cross-references

- #2347 / #2350 — smart-rename reliability work; resolution was kept profile-only there, consistent across badge and gate.

Issue #2349, opened 2026-06-23
Smart rename: persist attempted/failed state + source prompt across daemon restarts


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

`smart_rename_attempted` (and the inflight set) live only in memory on `AppState` (`src/server/mod.rs:303`). On a daemon restart the set clears, so a session that already failed a one-shot rename resurfaces with its default civilization name and the `Auto-name` (Pending) badge, as if it had never been attempted. There is no durable record of attempted / failed / succeeded, nor of the source prompt used.

#2347 deliberately kept this in-memory and added the manual "Auto-name now" re-trigger as the one-click recovery for a post-restart stuck badge. That closes the immediate UX hole, but the badge lifecycle is still not accurate across restarts.

## Proposal

Persist a minimal smart-rename state on the session record so the badge reflects reality after a restart:

- attempted / failed / succeeded (so a terminal failure shows as such, not perpetual Pending)
- last failure reason (for the sidebar / debug)
- the capped source prompt used, so a manual re-trigger does not depend on the event store still retaining the first `UserPromptSent`

This is a stored-data schema change, so it goes through `src/migrations/` per AGENTS.md.

## Out of scope

- The contention / trigger-timing fix (separate follow-up).

## Cross-references

- #2347 — smart rename reliability work; explicitly left this in-memory and pointed at this follow-up.

Seluj78:
I actually like the current system. Closing

Issue #2347, opened 2026-06-23
Smart rename: ~15% success on a real machine; add per-CLI rename agent, manual re-trigger, and reliability fixes


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Structured-view smart rename (auto-naming a still-default-named session from its first message, PR #2201, hardened in #2266) fails most of the time on my machine and gives no way to recover. A real session (`4e4dcdb68fdc4a91`) sent its first prompt and stayed named `Tatars` with the `Pending` auto-name badge still showing in the web sidebar.

Across my whole event log the one-shot title call has a ~15% success rate: **9 timeouts, 2 unusable-output skips, 2 actual renames**. Two failure modes, both hit this one session:

- The one-shot `claude -p` timed out at the 45s cap (contention with the session's own live worker on the same machine and API).
- A later retry ran `claude` but it returned an unusable title (a bare `NONE` on a perfectly valid task, or chatty prose), so the sanitizer rejected it.

This issue covers three things: make the one-shot reliable, let the smart-rename agent be chosen independently of the session's agent (cheaper / more reliable title model), and add a manual "auto-name now" re-trigger so a stuck session is recoverable.

## Evidence from the affected session

Session `4e4dcdb68fdc4a91`, agent `claude`, structured, worktree cwd `.../agent-of-empires-worktrees/Tatars`. One `UserPromptSent` (seq 8) at `06:47:54`.

`~/.agent-of-empires/debug.log`:

```
06:48:39  DEBUG smart_rename: one-shot timed out
06:58:58  DEBUG smart_rename: skip: agent output not a usable title session=4e4dcdb68fdc4a91
```

The first timeout is exactly 45s after the prompt (`06:47:54 + 45s = 06:48:39`), matching `ONESHOT_TIMEOUT`. The second line is the same session failing again on unusable output.

Whole-log tally:

```
one-shot timed out                       9
not a usable title                       2
renamed session from first message       2
```

Benchmark of the title prompt against each installed CLI in the real worktree cwd (3 runs each):

| CLI | flag | latency | output |
|-----|------|---------|--------|
| claude | `-p` | ~6s | flaky: bare `NONE` on a valid task; sometimes chatty prose |
| codex | `exec` | 4-11s (cold 11s) | clean title every run |
| opencode | `run` | 5-7s | cleanest, most consistent |

[...]

PR #2651, opened 2026-07-04
fix(session): defer smart-rename to turn-complete + bound global concurrency (closes #2348)

### LINKED ISSUE #2348: Smart rename: fix one-shot/live-worker contention at the source (defer + bound concurrency) instead of the 120s timeout

## Problem

Smart rename spawns the one-shot title CLI from the prompt handler at the same instant the session's own worker starts its first heavy turn (`src/server/api/acp.rs:1036`). The two race for CPU and the same provider API, so the one-shot blows past its timeout. #2347 raised `ONESHOT_TIMEOUT` 45s -> 120s to absorb that contention, but the bump is a band-aid: it lets the race run longer rather than removing it, and with only a per-session inflight guard, N stuck sessions can hold N host one-shot processes for up to 120s each.

## Proposal

Address the contention at the root instead of via the timeout:

- Defer the one-shot until the session's first turn settles (trigger on turn-complete / `Stopped`) rather than firing it concurrently with the live turn. #2347 lists this turn-complete trigger as an optional follow-up; this is that follow-up.
- Bound concurrency with a small global semaphore (e.g. max 1-2 concurrent one-shots) so a burst of new sessions cannot fan out unboundedly.

Once the one-shot no longer races the live worker, the timeout can likely come back down (60-90s) and be tuned from metrics.

## Out of scope

- Persisting smart-rename state across restarts (separate follow-up).
- Replacing the CLI one-shot with a direct HTTP title model.

## Cross-references

- #2347 — smart rename reliability + per-CLI rename agent + manual re-trigger (introduced the 120s bump this issue revisits).

jerome-benoit:
## Description

Closes #2348.

The smart-rename one-shot was spawned from the ACP prompt handler at the same instant the session's own worker started its first heavy turn, so the two contended for CPU and the same provider API. PR #2347 band-aided the resulting failures by raising `ONESHOT_TIMEOUT` from 45s to 120s; the race continued for longer, and N stuck sessions could hold N host one-shot processes for up to 120s each (unbounded fanout under burst).

**What changed**

- **Trigger moved from the prompt handler to a turn-complete listener.** The `tokio::spawn(try_smart_rename(...))` at [`src/server/api/acp.rs`](src/server/api/acp.rs) previously placed at line 1101 is removed. The one-shot now fires from the existing [`acp_event_listener`](src/server/mod.rs) at the first clean `Event::Stopped { reason: "prompt_complete" }` for the session. This reuses the single-listener pattern already spawned at process init, so no new subscription task is added.
- **Reason allowlist.** Only `"prompt_complete"` fires the rename. Every other `Event::Stopped` reason (`rate_limited`, `user_stopped`, `user_forced`, `agent_unresponsive`, `prompt_orphaned`, `reattach_idle`, `approval_cancelled_on_restart`, `restart_pending`) is either not a real turn boundary or a state where auto-renaming would be intrusive or wasteful.
- **First-turn gate.** The listener does two cheap sync `HashSet::contains()` checks against the existing `smart_rename_attempted` and `smart_rename_inflight` state on `AppState` BEFORE spawning, so non-matching `Stopped` events (already-renamed sessions, multi-turn sessions on turn 2+) drop before the event-store lookup or the spawn.
- **Global concurrency bound.** New `smart_rename_semaphore: tokio::sync::Semaphore` field on `AppState` (cap = `MAX_CONCURRENT = 2`, giving ~2 renames/min steady-state under saturation on multi-core hosts). The permit is acquired inside `try_smart_rename` in a scoped block that spans only the `run_oneshot` I/O, so ineligible calls and same-session duplicates (already rejected by `InflightGuard`) never consume a slot. `force_smart_rename` at [`src/server/api/sessions.rs`](src/server/api/sessions.rs) is unchanged: it still spawns `try_smart_rename` directly (manual retry stays on-demand), and its call automatically routes through the same semaphore.
- **`ONESHOT_TIMEOUT` returns to 60s.** Conservative middle of #2348's suggested 60-90s range; standalone the one-shot finishes well under 12s, and 60s leaves headroom for cold agent starts without holding a semaphore slot as long as #2347's 120s band-aid did. Metric-driven fine tuning is a follow-up.
- **`should_trigger_smart_rename` extracted as a pure sync predicate.** Takes `(&Event, &session_id, &attempted, &inflight)` and returns `bool`; the acp listener holds the two sync `Mutex` guards only across this predicate call and never spans an `await` under lock.

**Why (race at source rather than absorbing it)**

Bumping the timeout absorbed the contention window but did nothing about the underlying race: the one-shot process still competed with the live worker for the same provider tokens, and each stuck session could hold a host process for up to 120s. This PR removes the race so the one-shot only runs when the worker is idle, which means the timeout can come back down AND the number of concurrent one-shots can be capped without hurting the happy path.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (N/A: backend-only race fix)

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (the visible behavior after a first prompt is unchanged; the timing of the auto-rename shifts from "during the first turn" to "just after the first turn ends", which is the intended fix, not a UX change).

**TUI / CLI (e2e test)**

- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle. Session status, tmux, and structured view are untouched.

Regression coverage lives with the code:

- `oneshot_timeout_is_60s` (drift-guard against a future well-meaning bump-back to absorb a symptom instead of finding its cause).
- `should_trigger_smart_rename_only_on_clean_prompt_complete_stop` (predicate table: every documented `Event::Stopped` reason variant, non-`Stopped` events, and both per-session gate short-circuits).
- `smart_rename_semaphore_bounds_concurrent_permits_to_max` (5-task burst; asserts peak concurrency <= `MAX_CONCURRENT` AND >= 2 so a future silent size drop does not go unnoticed).
- `force_smart_rename_attempted_clear_re_enables_retry` (structural check: force-rename's attempted-clear step at [`sessions.rs`](src/server/api/sessions.rs) 2582-2587 re-enables the predicate for the same session).

Existing 23 smart_rename tests unchanged and passing; 15 acp handler tests unchanged and passing (regression check for the removed prompt-handler spawn).

**Gates run locally:**

- `cargo fmt --check`: clean
- `cargo clippy --features serve --all-targets -- -D warnings`: clean
- `cargo test --features serve --lib session::smart_rename`: 27/27 pass
- `cargo test --features serve --lib server::api::acp`: 15/15 pass

## Out of scope

- Persisting smart-rename state across restarts (tracked separately as #2349 per the issue body).
- Replacing the CLI one-shot with a direct HTTP title model.
- Metric-driven tuning of `ONESHOT_TIMEOUT` or `MAX_CONCURRENT` beyond the conservative values chosen here.
- Making `MAX_CONCURRENT` configurable via `ProfileConfig`.
- Any change to `try_smart_rename`'s config-resolution logic (merged as #2602 for #2351).
- Any change to `force_smart_rename` semantic beyond routing through the shared semaphore (which happens automatically because it spawns `try_smart_rename`).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** claude-opus-4-7 (Anthropic)

Design fork (WHERE to hook the turn-complete trigger: session status transition vs first-message-persisted vs ACP `Event::Stopped` broadcast vs a pre-existing lifecycle hook) was resolved with a read-only oracle consultation grounded in seven code citations before implementation.

The verdict picked the `Event::Stopped { reason: "prompt_complete" }` listener because smart-rename is already gated to ACP structured-view sessions by `SkipReason::NotStructured`, so hooking the ACP broadcast has zero coverage gap for terminal-tool sessions.

Implementation, tests, and cross-validation followed. The final diff was cross-validated by a second oracle pass against a 16-item checklist covering the removed spawn, the new listener hook, semaphore semantics, sync-lock-across-await hazards, race between listener and force path, and AGENTS.md compliance. Subsequent review cycles (three-lens post-review on design + Rust idioms + harmonization + terminology, then a fresh algorithmic + docs pass, then a security + reliability + observability + merge-readiness pass) returned zero MUST-FIX findings; each surfaced nitpick was applied as a small follow-up commit before the branch was rebased onto current main.

## What changed
- Moved smart-rename from starting during the live prompt handler to running after the first turn fully completes.
- Added a global limit so only a small number of smart-rename jobs can run at once.
- Reduced the smart-rename timeout back to a shorter, safer value.
- Kept manual “force smart rename” support, but routed it through the same limit.
- Added and updated tests around trigger timing, concurrency limits, timeout, and retry behavior.

## Benefits
- Reduces contention and startup races for new sessions.
- Prevents smart-rename jobs from piling up during bursts of activity.
- Makes the rename flow more predictable and less prone to hanging.

## Technical notes
- The trigger logic was split into a pure check that only allows a clean turn-complete stop event and blocks duplicate/in-flight runs.
- Session-level logging was improved for failures, skipped runs, and timeouts.

### DIFF
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -1107,14 +1107,11 @@ pub async fn acp_prompt(
         .acp_supervisor
         .publish_user_prompt_with_attachments(&id, req.text.clone(), &attachments)
         .await;
-    // Best-effort: auto-rename a still-default-named session from this first
-    // message via AoE's one-shot mode. Detached so it never blocks or fails the
-    // prompt; all gating lives inside. See session::smart_rename.
-    tokio::spawn(crate::session::smart_rename::try_smart_rename(
-        state.clone(),
-        id.clone(),
-        req.text.clone(),
-    ));
+    // Smart-rename now fires from `acp_event_listener` on the first clean
+    // `prompt_complete` `Event::Stopped` for this session, so the one-shot
+    // never races this handler's live worker for the same provider API.
+    // The event-store lookup of the first prompt happens in the listener.
+    // See `session::smart_rename` and #2348.
     match state
         .acp_supervisor
         .send_prompt(&id, &req.text, &attachments)
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -306,6 +306,11 @@ pub struct AppState {
     /// respawn a one-shot agent; one attempt per session bounds that cost and
     /// clears the `pending` sidebar chip once an attempt has run.
     pub smart_rename_attempted: std::sync::Mutex<std::collections::HashSet<String>>,
+    /// Global cap on concurrent smart-rename one-shots so a burst of new
+    /// sessions cannot fan out into N host processes each holding a slot for
+    /// up to `ONESHOT_TIMEOUT`. Held only across the child spawn + wait. See
+    /// `session::smart_rename` and #2348.
+    pub smart_rename_semaphore: tokio::sync::Semaphore,
     /// Suppression set for the startup-recovery cascade. While an entry is
     /// present and younger than `recovery::RECENTLY_RESTARTED_TTL`, the
     /// `status_poll_loop` skips `update_status_with_metadata` for that
@@ -993,6 +998,9 @@ pub async fn start_server(config: ServerConfig<'_>) -> anyhow::Result<()> {
         instance_locks: RwLock::new(std::collections::HashMap::new()),
         smart_rename_inflight: std::sync::Mutex::new(std::collections::HashSet::new()),
         smart_rename_attempted: std::sync::Mutex::new(std::collections::HashSet::new()),
+        smart_rename_semaphore: tokio::sync::Semaphore::new(
+            crate::session::smart_rename::MAX_CONCURRENT,
+        ),
         recently_restarted: crate::session::recovery::new_recently_restarted(),
         recovery_pending: crate::session::recovery::new_recovery_pending(),
         cleanup_defaults_cache: RwLock::new(CleanupDefaultsCache {
@@ -3810,6 +3818,43 @@ async fn acp_event_listener(state: Arc<AppState>) {
             }
         }
 
+        // Smart-rename defer: fire the one-shot only on a clean
+        // `prompt_complete` `Event::Stopped`, so it never races the live worker
+        // for the same provider API. Two sync mutex `contains()` checks and
+        // the reason-allowlist drop the 99% of events that need no action
+        // before we touch the event store or spawn a task. See #2348.
+        let should_rename = {
+            let attempted = state
+                .smart_rename_attempted
+                .lock()
+                .expect("smart_rename_attempted poisoned");
+            let inflight = state
+                .smart_rename_inflight
+                .lock()
+                .expect("smart_rename_inflight poisoned");
+            crate::session::smart_rename::should_trigger_smart_rename(
+                frame.event.as_ref(),
+                &frame.session_id,
+                &attempted,
+                &inflight,
+            )
+        };
+        if should_rename {
+            if let Some(first_message) = state.acp_event_store.first_user_prompt(&frame.session_id)
+            {
+                let state_for_rename = state.clone();
+                let session_id = frame.session_id.clone();
+                tokio::spawn(async move {
+                    crate::session::smart_rename::try_smart_rename(
+                        state_for_rename,
+                        session_id,
+                        first_message,
+                    )
+                    .await;
+                });
+            }
+        }
+
         let status_intent = derive_acp_status(frame.event.as_ref());
         let acp_change = derive_acp_session_change(frame.event.as_ref());
         if status_intent.is_none() && acp_change.is_none() {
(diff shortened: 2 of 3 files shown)


Seluj78 (approved):

## AoE Review

Backend-only, 3 files (server, session/smart_rename). CI green (all required shards, Rust, macOS, clippy, fmt, docs, skill-check pass). No unresolved review threads; Codecov green (100% modified lines).

Mechanical gates (em dash, dead code, namespace strings, tracing targets, format in macro args, secrets in logs): all clean. Settings parity N/A (MAX_CONCURRENT is a const, not a Config field). No migration, CLI, docs, or web surface touched.

### Nit (optional, low priority)

`src/server/mod.rs` ~324-339: the listener locks both smart_rename mutexes on every ACP broadcast frame before the event-variant check. Only a prompt_complete Event::Stopped can pass. Gating on matches!(event, Event::Stopped { .. }) first would skip both locks on the high-volume streaming-delta path. Micro cost: two uncontended lock/unlock per frame.

### Verdict

Implementation: Approve. Project fit: Accept.

Race fixed at source rather than absorbed: the one-shot moves off the live-worker contention window, letting the timeout drop 120s to 60s and adding a global cap (MAX_CONCURRENT=2). Sync mutexes never span an await (predicate computed in a scoped block, guards dropped before the event-store lookup and spawn); the semaphore permit is scoped tightly around run_oneshot so ineligible paths never consume a slot; InflightGuard dedupes the tiny pre-acquire window so a double prompt_complete cannot double-spawn. Text source preserved (event store first UserPromptSent equals the req.text published before send_prompt). Sync SQLite read in the async listener matches existing precedent and is gated to once-per-session. Tests are strong: predicate table over every Stopped reason plus both gates, semaphore burst bound, timeout drift-guard, force-path re-enable. Scope tight, no new deps, no config surface.

PR #2652, opened 2026-07-04
fix(containers): close the swallowing-existence-probe bug class

### LINKED ISSUE #2596: Swallowing existence probe in discard_sandbox_container_after_move can skip container removal
Split out from review of #2576 (flagged by @jerome-benoit as out-of-scope for that PR).

`src/session/worktree_edit.rs` `discard_sandbox_container_after_move` removes a stale sandbox container after a worktree move using the same swallowing-existence-probe shape that #2576 removes from the deletion/creation paths:

```rust
match container.exists() {
    Ok(true) => { /* remove */ }
    Ok(false) => {}
    Err(e) => tracing::warn!(target: "containers.runtime", ... "failed to check ..."),
}
```

A transient `docker inspect` failure makes `exists()` return `Err` (or a non-zero exit → `Ok(false)`), so removal is skipped and the stale container is left mounting the old worktree path. On next start a fresh container is created and the old one is orphaned. Same bug class as the deletion path.

This is best-effort (the rename already succeeded; failure is only logged), so it is lower severity than the deletion path. Suggested fix: adopt the idempotent `DockerContainer::teardown` introduced in #2576, calling it unconditionally and only `warn`-ing on `Teardown::Failed`.

Not introduced by #2576.

jerome-benoit:
## Description

Closes #2596.

`discard_sandbox_container_after_move` gated its `docker rm` behind a `container.exists()` pre-probe: a transient `docker inspect` `Err` (or a non-zero exit surfaced as `Ok(false)`) skipped removal and orphaned the stale container against the old worktree path. Same swallowing-existence-probe shape that #2576 removed from the deletion/creation paths.

**Scope changed during review.** The initial fix was single-function (+35/-18 across 2 files). Thirteen rounds of parallel `explore` + `oracle` self-review each surfaced a substantive issue prior rounds missed, expanding the fix to close the whole swallowing-existence-probe class **on the container-runtime surface** the bug exposed: the same collapse appeared at four other gate call sites, plus a `JoinError` collapse at the async boundary of two server handlers, plus a daemon-down-vs-container-absent conflation deep in the runtime layer, plus a log-ergonomics issue with multi-line error Display strings, plus a Round-11 finding that string-carrying `DockerError` variants could still smuggle newlines via raw stderr.

**Explicitly out of scope, tracked as follow-ups** (all filed before this PR requests review):
- #2653: `git::worktree::branch_exists` swallows spawn errors as `false` — same class, git surface.
- #2654: `DockerContainer::exists` / `does_container_exist` collapses daemon-down to `Ok(false)` — same class, existence-probe surface (scope covers both the internal function AND the caller-side handling at `cleanup_sandbox_worktree:424`).
- #2655: Capture real Apple `container` CLI daemon-down fixture, replacing this PR's placeholder.
- #2656: `permission_denied` classifier is inline substring, not per-runtime — completes the marker refactor started here.

**Design.** Two new public items in [`src/containers/`](src/containers/):

- `pub enum Probe { Running, NotRunning, Unknown(DockerError) }` + `DockerContainer::probe_running() -> Probe` — canonical replacement for `is_running().unwrap_or(false)` at every gate site. Preserves the difference between "definitively not running" and "inspection failed".
- `pub fn DockerContainer::discard() -> Teardown` — sibling to `teardown()`. Same idempotent classification, but preserves session-scoped named ignore volumes (`target/`, `node_modules/`) so the recreated container re-attaches its build caches on next start.

**Runtime layer** ([`src/containers/runtime_base.rs`](src/containers/runtime_base.rs), [`src/containers/runtime.rs`](src/containers/runtime.rs), [`src/containers/error.rs`](src/containers/error.rs)):

- `is_container_running` now sniffs `docker inspect` stderr to distinguish container-absent (routes to `Probe::NotRunning`) from daemon-down / permission-denied / other transient (routes to `Probe::Unknown` with an actionable `DockerError` variant). Previously collapsed all non-success into `Ok(false)`, which made `Probe::Unknown` dead code for the failure mode every warn message names.
- `RuntimeBase::classify_inspect_failure` classifies stderr into `DaemonNotRunning` / `PermissionDenied` / `InspectFailed(String)` / `Ok(false)`, mirroring the pattern `remove()` and `run_create` already use.
- Per-runtime `daemon_down_markers: &'static [&'static str]` (parallel to `not_found_markers`), so each runtime only matches its own wording. Cross-runtime bleed is impossible by construction.
- `DockerError::InspectFailed(String)` dedicated variant (parallel to `RemoveFailed` / `StopFailed` / `CreateFailed`).
- `DockerError` `Display` strings are single-line by convention with the actionable remediation inlined. `NotInstalled` includes the install URL; `DaemonNotRunning` includes the "Start Docker Desktop" hint; `PermissionDenied` includes the `usermod -aG docker` one-liner. `error = %e` at any `tracing::warn!` site is safe by default (no multi-line log splits) AND preserves actionable prose at every user-facing surface (TUI dialogs, CLI stderr, anyhow propagation) without an accessor to remember.
- `pub(crate) fn sanitize_stderr(stderr: &str) -> String` in `error.rs`: called at every construction site that wraps raw runtime stderr into a `DockerError::*Failed` variant, using `str::lines()` to handle Unix `\n` and Windows `\r\n` line terminators, filtering blank interior lines, and joining with ` | ` so log correlation stays single-line even when Docker emits multi-line stderr.
- Apple `/0/status` `.as_str()` guard: if Apple ever changes `/0/status` from a JSON string to a nested type, the guard routes to `Err(InspectFailed)` rather than silently returning `Ok(false)` (fail closed, not fail open — same class as #2596, one schema shift away).
- Removed dead `DockerError::CommandFailed` variant (last caller migrated to `InspectFailed`).

**Sites hardened** (all fail-closed on `Probe::Unknown` or its async-boundary equivalent):

| Site | File | Semantics on `Probe::Unknown` |
|---|---|---|
| Worktree rename gate | [`sandbox_container_holds_worktree`](src/session/worktree_edit.rs) | Return `true` (block rename) + warn. Prevents EBUSY. |
| Post-move discard | [`discard_sandbox_container_after_move`](src/session/worktree_edit.rs) | `container.discard()`; classifier handles the three-way outcome. |
| Session stop | [`Instance::stop`](src/session/instance.rs) | Warn + best-effort `container.stop()`, second warn if that also fails. |
| CLI pre-flight | [`stop_session`](src/cli/session.rs) | Treat as "possibly running" so `inst.stop()` proceeds. |
| Worktree cleanup | [`cleanup_sandbox_worktree`](src/git/cleanup.rs) | Warn + best-effort `container.start()`. |
| Server rename/edit handlers | [`rename_session`, `set_worktree_name`](src/server/api/sessions.rs) | Both go through the shared [`probe_container_holds_worktree`](src/server/api/sessions.rs) helper: `spawn_blocking` `JoinError` → `container_holds = true` via `unwrap_or_else(warn; true)`, returns `409 CONFLICT`. |

**Documentation.** Every function whose observable behavior changed on `Probe::Unknown` has its doc updated or added: `Instance::stop`, `stop_session`, `cleanup_sandbox_worktree`, `rename_session`, `set_worktree_name`. The `Probe` enum doc explicitly warns against `unwrap_or(false)` and points to #2596. The Docker/Podman `container inspect` argv is pinned by an inline comment (`"No such container"` vs `"No such object"` stderr distinguishes subcommands); the Apple arm has a mirror comment naming its marker dependency. `is_not_found` and `is_daemon_down` docs cross-reference each other explaining the intentional case-sensitivity asymmetry.

**Commit history preserved.** 17 commits across 13 review rounds. The full iteration trail is kept deliberately: each commit's message documents the round's finding and the exact fix rationale, so a future maintainer investigating any single-line change can `git blame` directly onto the review-round context that motivated it. No squash requested.

Diff: **+637 / -84** across 9 files.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**Classifier tests** (source-of-truth for the enum outcomes, no runtime needed):

- `classify_ok_is_removed`, `classify_not_found_is_already_gone`, `classify_other_error_is_failed` — `Teardown` variants (shipped in #2576, reused for `discard()`).
- `probe_ok_true_is_running`, `probe_ok_false_is_not_running`, `probe_err_is_unknown` — `Probe` variants.
- `inspect_not_found_stderr_collapses_to_ok_false` — three-runtime not-found stderr → `Ok(false)`.
- `inspect_daemon_down_stderr_maps_to_daemon_not_running` — four fixtures (Docker + Podman socket mode + Podman Desktop mode + Apple placeholder) → `DaemonNotRunning`.
- `inspect_permission_denied_stderr_maps_to_permission_denied` — Docker-permission stderr → `PermissionDenied`.
- `inspect_generic_transient_maps_to_inspect_failed` — fallthrough → `InspectFailed`.
- `is_daemon_down_is_cross_runtime_isolated` — regression guard: each runtime's marker MUST NOT match another runtime's stderr (Docker × Podman × Apple full matrix). Locks in per-runtime `daemon_down_markers` isolation.

**Display / sanitize tests** in `containers::error::tests`:

- `display_strings_are_single_line_and_actionable` — locks single-line invariant on `NotInstalled` / `DaemonNotRunning` / `PermissionDenied` plus presence of `docs.docker.com`, `systemctl`, `usermod` in their Display outputs.
- `parameterized_variants_stay_single_line_when_stderr_multiline` — verifies `sanitize_stderr` sanitizes multi-line stderr embedded in `InspectFailed`.
- `sanitize_stderr_trims_trailing_whitespace` — basic input/output cases (trailing newline, padded whitespace, multi-line joined).
- `sanitize_stderr_handles_empty_and_whitespace_inputs` — empty string, whitespace-only, newline-only all yield `""`.
- `sanitize_stderr_handles_crlf_line_endings` — Windows CRLF stderr (`"a\r\nb"` → `"a | b"`) via `str::lines()`.
- `sanitize_stderr_skips_blank_interior_lines` — blank lines between summary and detail do not produce empty ` |  | ` segments.
- `sanitize_stderr_is_idempotent` — applying twice equals applying once; protects against a future double-sanitize helper.
- `sanitize_stderr_lone_cr_is_preserved_interior` — documents the known limitation that a lone `\r` mid-string (NOT a line terminator per `str::lines` spec) survives verbatim; Docker/Podman/Apple do not emit lone `\r` in practice.

**Site-level tests** (short-circuit invariants):

- `holds_worktree_short_circuits_without_sandbox` — proves the `is_sandboxed` guard on `sandbox_container_holds_worktree` short-circuits before any container runtime call.
- `discard_after_move_short_circuits_without_sandbox` — same for `discard_sandbox_container_after_move`, plus a `<100 ms` wall-clock bound that catches a future edit that reorders the `is_sandboxed` guard below the runtime call.

**Not covered by a dedicated test:** the `Probe::Unknown` branch behavior at each of the 5 hardened sites (attempting stop, attempting start, returning true, returning 409). There is no mock docker runtime seam in the codebase today; adding a `#[cfg(test)] DockerContainer::with_runtime(...)` seam is a follow-up.

- [ ] Added or updated a Playwright spec under `web/tests/`

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle observably

Rationale: this is a defense-in-depth hardening of internal probe-and-gate paths. Function signatures unchanged, HTTP status codes unchanged on the happy path, CLI output unchanged on the happy path. The bug fixed (transient `docker inspect` failure orphaning a container or opening a rename gate) is not deterministically reproducible from an e2e test without mocking the docker runtime. Unit-level classifier + short-circuit coverage is the right granularity.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.7 (Sisyphus, via OhMyOpenCode)

**Any Additional AI Details you'd like to share:** Draft fix + 13 rounds of self-review. Each round: 2 parallel `explore` cross-validations + 1 `oracle` design critique + synthesis. Rounds 5–13 each surfaced a substantive issue prior rounds missed:

- **Round 5**: `is_container_running` collapsed daemon-down transient into `Ok(false)` (silent fail-open dead code in every gate's `Probe::Unknown` branch).
- **Round 6**: Apple `not_found_markers: ["not found"]` was dangerously broad — a daemon error containing "socket not found" would misclassify as absent.
- **Round 7**: `InspectFailed(String)` opaque variant hid `DaemonNotRunning` / `PermissionDenied` classification that `run_create` already surfaced elsewhere.
- **Round 8**: Podman Desktop / machine mode emits `"Cannot connect to Podman."` (no `"socket"`), which the initial inline substring missed. Multi-line `Display` remediation prose split `tracing` events across physical log lines under the text-mode subscriber.
- **Round 9**: Round 8's `.summary()` opt-in migration missed 6 warn sites in `containers/mod.rs` — empirical proof that call-site-opt-in single-line formatting silently regresses.
- **Round 10**: Round 9's `remediation()` accessor had zero callers (dead code per AGENTS.md). Inlined the actionable hints back into single-line `Display` strings.
- **Round 11**: The string-carrying `DockerError` variants (`InspectFailed`, `RemoveFailed`, etc.) could still leak multi-line stderr from Docker directly into their `Display` output — the single-line convention was only enforced on unit variants. Added `sanitize_stderr` and migrated all 6 construction sites.
- **Round 12**: `sanitize_stderr` was applied to a `serde_json::Error` at `runtime.rs:172` where the function name lies (serde errors are single-line by design). Dropped the misleading no-op call. Also refactored `sanitize_stderr` from `.trim().replace('\n', " | ")` to `.lines().map(trim).filter(non-empty).join(" | ")` to handle Windows CRLF stderr and blank interior lines cleanly.
- **Round 13**: `teardown()` doc had a duplicated invariant fragment introduced by Round 12's `NOTE:` reword. Collapsed to one clean sentence mirroring the sibling `discard()` doc. Also dropped review-process jargon (`"Round 10's"`) from `sanitize_stderr` doc, added an inline comment on `runtime.rs:172` documenting the single-line assumption for future readers, and added `sanitize_stderr_lone_cr_is_preserved_interior` to lock the known limitation on lone `\r`.

Post-review from CodeRabbit landed one additional refactor: extracted the duplicated `spawn_blocking + fail-closed` block between `rename_session` and `set_worktree_name` into a shared `probe_container_holds_worktree` helper, ensuring the fail-closed policy stays synchronized across the two endpoints.

All design decisions (`discard()` as first-class method vs exposing `classify_removal`; per-runtime `daemon_down_markers` vs inline substring OR; single-line `Display` with inlined remediation vs separate accessor; `sanitize_stderr` as free function vs constructor methods vs newtype) were validated against actual codebase precedent — `not_found_markers`, `classify_removal`, `run_create` stderr sniff — before implementation. Local verification per commit: `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings`, `cargo test --lib -- containers session::worktree_edit::tests` (**77/77 passed**).

**NOTE:**

This PR makes sandbox container handling more reliable and safer when container status checks fail.

- It introduces a clearer “probe” outcome for whether a container is running (instead of assuming “not running” when inspection fails), and updates key rename/move/cleanup/session and server handlers to treat unknown results conservatively.
- It fixes stale sandbox containers after worktree moves by removing the “exists then remove” pattern and switching to an idempotent discard flow that removes the container best-effort while preserving the intended named ignore volumes.
- It hardens container runtime behavior by distinguishing “container missing” vs “daemon down/permission denied/other inspect failures,” and adds an `InspectFailed` error path.
- It improves error presentation: error messages are consistently single-line, and runtime stderr is sanitized before being embedded in errors.

Benefits:
- Fewer stale containers after worktree moves and renames
- Safer behavior during transient Docker/Podman inspection failures
- More predictable cleanup decisions (including when the daemon is unreachable)
- Clearer, cleaner diagnostics when operations fail


jerome-benoit (commented):
_(review withdrawn — posted in error, disregard)_

Seluj78 (approved):

**Implementation: Approve. Project Fit: Accept.**

The code is correct and unusually well-hardened. `classify_inspect_failure` orders not-found, then daemon-down, then permission, then `InspectFailed`, with no marker collisions; the per-runtime `daemon_down_markers` refactor is proven isolated by `is_daemon_down_is_cross_runtime_isolated`; every gate (`sandbox_container_holds_worktree`, `Instance::stop`, `stop_session`, `cleanup_sandbox_worktree`, and the two server handlers via `probe_container_holds_worktree`) fails closed on `Probe::Unknown`, and the async-boundary `JoinError` is fail-closed too. Tracing calls carry `target:`, use `%` field captures, and single-line `Display` plus `sanitize_stderr` keeps log correlation intact. Classifiers and short-circuits are tested; the untested part (five trivial `Probe::Unknown` match arms) is disclosed and blocked only by the absence of a mock-runtime seam, which is proportionate.

On scope: the CodeRabbit out-of-scope warning is fair to raise, but the expansion is coherent. Every change closes the same swallowing-existence-probe class the bug exposed, and the four genuinely out-of-scope items (#2653 through #2656) were filed as follow-ups before review. That reads as disciplined class-closure, not incidental churn. The title-prefix warning from CodeRabbit is a false positive: scoped `fix(containers):` is standard Conventional Commit and matches repo convention (`fix(acp):`, `feat(web):`).

Two minor notes, neither blocking:

- **PR body em dashes.** The Description and Test Coverage sections use em dashes, which AGENTS.md bans in human-authored prose. Title and all 17 commit messages are clean. Body-only; not shipped, so not a merge blocker.
- **`src/containers/mod.rs:325`.** `probe_err_is_unknown` seeds the running-probe classifier with `DockerError::RemoveFailed`. A running probe never yields a removal error; `InspectFailed` or `DaemonNotRunning` would model the real Err path. Cosmetic; the `Err(_)` to `Unknown` mapping is still exercised.

The remaining `exists().unwrap_or(false)` at `src/git/cleanup.rs:713` is the same class but explicitly deferred to #2654; correct call. 17 commits across 13 rounds is a lot of history to carry, but the per-round blame context is justified and no squash was requested, so that is a maintainer preference rather than a blocker.

PR #2688, opened 2026-07-06
fix(tui): keep search matches on Enter so n/N can cycle (closes #2676)

### LINKED ISSUE #2676: TUI search: n/N (SearchNext/SearchPrev) never fire — Enter clears search_matches, so the cycle context is unreachable
## Summary

In the TUI home/session list, the `n` / `N` search-cycle bindings (`ActionId::SearchNext` / `SearchPrev`) can **never fire**. Committing a search with `Enter` clears `search_matches`, which is the exact state the cycle bindings are guarded on — so there is no reachable state in which pressing `n`/`N` advances to the next/previous match. The `[i/N]` match counter shown in the search bar therefore stays pinned at `[1/N]` and can never be advanced.

Reproduces at `a812ed4d` (`v1.12.0-38-ga812ed4d`) and on the `v1.12.0` release.

## Repro

1. Open the TUI dashboard, several sessions listed.
2. Press `/`, type a query that fuzzy-matches multiple sessions. Search bar shows e.g. `/foo [1/4]`.
3. Try to jump to the next match:
   - While the query box is focused, `n`/`N`/`j`/`k` are typed into the query string (expected — the input owns the keys).
   - Press `Enter` (or `Esc`) to leave the query box. Now `n`/`N` create **new sessions** instead of cycling matches, and the highlight/counter is gone.

There is no key sequence that reaches "search committed, matches still active, cursor cyclable".

## Root cause

The cycle bindings are guarded on `Context::SearchActive`, which is `has_search`, which is `!search_matches.is_empty()`:

- `src/tui/home/bindings.rs:349-364` — `SearchNext` (`n`) / `SearchPrev` (`N`), `context: Context::SearchActive`.
- `src/tui/home/bindings.rs:183` — `Context::SearchActive => ctx.has_search`.
- `src/tui/home/input.rs:2372` — `has_search: !self.search_matches.is_empty()`.

But the only two ways out of the query editor **both clear `search_matches`**:

`src/tui/home/input.rs:2287-2307`
```rust
if self.search_active {
    match key.code {
        KeyCode::Esc => {
            self.search_active = false;
            self.search_query = Input::default();
            self.search_matches.clear();      // <-- cleared
            self.search_match_index = 0;
        }
        KeyCode::Enter => {
            self.search_active = false;
            self.search_query = Input::default();
            self.search_matches.clear();      // <-- cleared on commit too
            self.search_match_index = 0;
        }
        _ => {
            self.search_query.handle_event(&crossterm::event::Event::Key(key));
            self.update_search();
        }
    }
    return None;
}
```

So:
- While `search_active == true`, every key feeds the query box and returns before dispatch — `n`/`N` can't dispatch.
- After `Enter`/`Esc`, `search_matches` is empty → `Context::SearchActive` is false → `n`/`N` resolve to `NewSession`.

`SearchNext`/`SearchPrev` are dead code in practice. Their own dispatch arms (`src/tui/home/input.rs:2507-2525`) even early-return on empty matches, which is the only state they're ever reached in.

## Evidence this is a regression, not intended single-jump design

The code clearly anticipates `search_matches` persisting **after** the query editor closes (`search_active == false`), which is exactly what `Enter` breaks:

- `src/tui/home/mod.rs:2466-2470` and `3019-3023`, in the data-reload handlers:
  ```rust
  if self.search_active && !self.search_query.value().is_empty() {
      self.update_search();
  } else if !self.search_matches.is_empty() {
      // Recalculate match indices without moving the cursor
      self.refresh_search_matches();
  }
  ```
  The `else if` branch — "recalculate without moving the cursor" — only makes sense when a search is committed (`search_active == false`) yet matches are still live. That branch is currently unreachable because `Enter` nukes the matches.
- The search bar renders a live `[search_match_index + 1 / search_matches.len()]` counter (`src/tui/home/render.rs:991-999`) — UI that promises cyclable matches.
- The resolver unit test `search_cycle_overrides_new_session_when_active` (`src/tui/home/bindings.rs:1150`) passes only because it sets `has_search = true` by hand; no integration path can produce that state.

## Suggested fix

On `Enter`, close the query editor but **keep** the match set and index so the cycle context becomes reachable:

```rust
KeyCode::Enter => {
    self.search_active = false;
    // keep self.search_matches / self.search_match_index so `n`/`N` can cycle;
    // keep or clear the query text as desired, but do NOT clear matches.
}
```

Leave `Esc` as the "cancel and clear everything" path. That makes the `mod.rs` "recalculate without moving cursor" branch live and the `[i/N]` counter advanceable via `n`/`N`.

(Reported from CLI-driven investigation; happy to send a PR.)

jerome-benoit:
## Description

**What.** `Enter` now closes the search input while keeping `search_matches`, `search_match_index`, and `search_query` intact. `Esc` keeps its cancel-and-clear semantics. Every code path that rebuilds `flat_items` now routes through a single [`rebuild_flat_items()`](src/tui/home/input.rs) helper that re-scores `search_matches` against the new indices; `SearchStart` and palette-`Invoke` are hardened so the "matches populated + search inactive" state stays coherent.

**Why.** Three interlocking reasons plus a class-level invariant:

1. `n` / `N` (`ActionId::SearchNext` / `SearchPrev`) are guarded on `Context::SearchActive`, which resolves to `!search_matches.is_empty()`. The only two ways out of the search input both cleared the matches, so the guard was unreachable and the cycle bindings were dead. This also unblocks the "recalculate without moving the cursor" branch in the data-reload handlers at `mod.rs:2466-2470` and `:3019-3023`, and lets the `[i/N]` counter at `render.rs:991-999` advance while the search input is open.

2. Keeping the query (not just the matches) is required for reload survival: [`refresh_search_matches`](src/tui/home/input.rs) unconditionally wipes `search_matches` whenever the query is empty, and it fires via the reload tail at [`mod.rs:2467`](src/tui/home/mod.rs) on every storage or config reload. Clearing `search_query` on `Enter` while leaving matches populated would silently destroy the very matches `Enter` promised to keep, the first time any background reload landed. Keeping the query makes that same call re-score against the same haystack instead of wiping.

3. Keeping matches across `Enter` makes the `(search_active=false, search_matches populated)` state reachable for the first time. Every path that assigns `self.flat_items = self.build_flat_items()` was previously safe-by-accident in that state (matches were always empty), and each one now leaks stale indices unless it re-scores. This covers **27 sites** across:
   - `input.rs`: `apply_sort_order`, `apply_group_by`, `toggle_group_collapsed`.
   - `mod.rs`: the two reload paths, `toggle_archived_section`, `toggle_trashed_section`, `stamp_last_accessed` (fires on every session-select under `Attention` sort, so simple j/k navigation would silently invalidate matches), `drop_peer_deleted_rows` (disk-watcher peer-delete callback), and the creation-lifecycle hooks (`request_creation` / `cancel_creation` / `apply_creation_results` / `select_and_reveal_session`).
   - `operations.rs`: `toggle_favorite`, `toggle_archive`, `toggle_snooze`, `toggle_unread`, `toggle_project_pin`, `trash_session_by_id`, `restore_selected_from_trash`, `archive_selected_group`, `restart_selected_session`, `delete_group_with_sessions`.

Rather than patch each site with a duplicated `if !search_matches.is_empty() { refresh_search_matches() }` block, a single `pub(super)` helper `rebuild_flat_items()` handles the rebuild + re-score atomically, and every site routes through it. The only remaining bare `build_flat_items()` is the constructor at `mod.rs:2268` (uses the local `view` binding, runs before any `search_matches` could exist). Future contributors adding a new mutation site inherit the invariant automatically by calling the helper.

Two more sites tighten the state machine:

- **`SearchStart` on `/` reopen** now also clears `search_matches` + `search_match_index`, so the `[i/N]` counter does not briefly render over an empty input before the first keystroke rebuilds match state.
- **Palette `Invoke`** now clears `search_query` alongside `search_matches`, so a subsequent rebuild path calling `refresh_search_matches` cannot resurrect phantom matches from a stale query.

`apply_sort_order` visibility bumped from private to `pub(super)` so the regression test can call it directly (mirrors the visibility of sibling functions like `reload`, `refresh_search_matches`, `move_cursor`).

**Direction.** Keep matches on `Enter`, leave `Esc` as cancel-and-clear (vim parity: `/pat<CR>` preserves vim's `@/` register; only an explicit cancel resets it). Endorsed by @Seluj78 in [#2676 (comment)](https://github.com/agent-of-empires/agent-of-empires/issues/2676#issuecomment-4889970506).

**Reported and diagnosed by @Mikkolehtimaki:** full file:line breakdown, evidence-based regression proof, and identification of the pre-existing test collisions in [#2676](https://github.com/agent-of-empires/agent-of-empires/issues/2676). This PR implements the fix Mikko proposed, extended to cover the reload-survival invariant and the class of `flat_items`-rebuild sites surfaced during code review. The commit carries a `Co-authored-by: Mikko Lehtimäki` trailer so squash-merge attribution goes to Mikko too.

**How tested.**
- Unit test `test_search_mode_enter_commits_without_clearing_matches` asserts the full post-commit contract: matches preserved, `search_query` retained (so reloads re-score instead of wiping), `search_active` false, `search_match_index == 0`.
- Regression test `test_search_mode_enter_keeps_matches_for_cycling` presses `Enter`, walks `n` forward through the full match set, confirms `n` wraps to 0, `N` wraps back to the last, `N` steps back one.
- Regression test `test_reload_after_enter_preserves_search_state` covers the reload-survival invariant: types a query, commits with `Enter`, calls `reload()`, asserts `search_matches` count is unchanged, and confirms `n` still advances the index.
- Regression test `test_sort_order_change_after_enter_rescores_search_matches` uses a query that matches exactly one session (`"session0"`, at index 4 under `Newest` sort and 0 under `Oldest`). After `apply_sort_order`, it asserts the single match still points to `session0` and not a stale index landing on a different session. Since all 27 mutation sites route through the same `rebuild_flat_items()` helper, this test locks the class-level invariant: verifying one site validates all of them.
- Rewrote `test_search_mode_enter_exits_and_clears_state` and inverted `test_enter_clears_matches_so_n_opens_new_dialog` (now `test_esc_clears_matches_so_n_opens_new_dialog`) to lock in the new contract; dropped the redundant `test_enter_clears_matches_and_resets_index`; refreshed the inline comment on `test_reload_does_not_snap_cursor_after_enter`.
- Manual TUI check with 5 sessions: press `/`, type a shared prefix, `[1/N]` shows; `Enter` closes the search input, session-list highlights on matched rows remain; `n` advances the cursor through them, wraps at the end; `N` reverses; toggling sort order (`o`), archiving a session (`a`), favoriting (`F`), and j/k navigation under Attention sort all keep highlights on the right sessions; `Esc` clears everything.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full `tui::home` test suite (679 tests) pass.

**Out of scope.**
- The `SearchNext` / `SearchPrev` dispatch arms at `input.rs:2507-2525` are unchanged. They were correct; they just had no reachable state before this fix.
- The `search_cycle_overrides_new_session_when_active` unit test at `bindings.rs:1150` is unchanged. It set `has_search = true` by hand; the new integration tests exercise the same resolver arm through the real keypath.
- Rendering `[i/N]` after `Enter` (currently gated on `search_active` at [`render.rs:960`](src/tui/home/render.rs), so the counter disappears when the search input closes) is a UX polish question, separate from the correctness fix here. Row highlighting on matched sessions ([`render.rs:929`](src/tui/home/render.rs)) is not gated on `search_active` and stays visible post-commit, which is the primary visual feedback for cycling.
- Cursor-jump-to-match-0 on `Enter` (strict vim `/pattern` + Return semantics, where the cursor snaps to the first match on commit) is worth a follow-up if users report the skip-first-match feel, but sits outside the endorsed direction.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the bug is entirely a state-transition on `handle_key` and match-index cycling. Four module-level tests in `src/tui/home/tests.rs` exercise the real `HomeView::handle_key` codepath and the `apply_sort_order` entry point, pin the exact state invariants, and lock the class-level invariant that every `flat_items`-rebuild site now routes through `rebuild_flat_items()`. A tmux e2e would test a strict superset of the same paths with less precise assertions.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** claude-opus-4-7 (Claude Opus 4.7) via the OhMyOpenCode Sisyphus agent runtime.

Diagnosis and root-cause analysis in [#2676](https://github.com/agent-of-empires/agent-of-empires/issues/2676) is @Mikkolehtimaki's work. The AI implementation follows his file:line breakdown and the direction endorsed by @Seluj78, extends the fix to cover the reload-survival invariant and the class of `flat_items`-rebuild sites surfaced during code review, and adds regression coverage that locks the invariant at the helper level so future mutation sites inherit it automatically. Review requests will be answered directly, not via the model.

Closes #2676 

This PR refines TUI search behavior so committing a search with Enter keeps the search state intact, while Esc still fully cancels and clears it. That means users can now continue stepping through matches after searching, and search state also survives reloads and list updates more reliably.

It also centralizes list rebuilding through a shared helper, replacing several direct rebuild paths. This reduces stale match/index issues when the sidebar list changes due to sorting, grouping, loading, or session actions.

Tests were expanded and updated to cover:
- Enter preserving search state
- n / N cycling and wrapping after search commit
- reloads keeping search results available
- match rescoring after sort changes
- Esc continuing to clear search state as before

Benefits:
- Better search continuity and keyboard navigation
- More reliable search state across reloads and UI updates
- Fewer inconsistent match/index bugs after list changes


Seluj78 (approved):

## AoE Review: Approve

Bug fix for #2676. `Enter` now commits the search (closes input) while keeping `search_matches`, `search_query`, and `search_match_index`, so `n`/`N` cycling (guarded on `!search_matches.is_empty()`) is finally reachable. `Esc` keeps cancel-and-clear. A new `rebuild_flat_items()` helper wraps `build_flat_items()` plus a conditional `refresh_search_matches()`, and all 24 production mutation sites route through it.

### Verified

- **Class invariant holds.** Only three bare `build_flat_items()` calls remain in production: the constructor (`mod.rs:2268`, runs pre-search), the helper body itself, and the active-search branch of `apply_sort_order` (`input.rs:3532`, followed by `update_search()`). All other 24 sites route through `rebuild_flat_items()`.
- **Both re-score functions match their callers.** `update_search` clears and resets index to 0 (correct for active-typing rescore); `refresh_search_matches` preserves and clamps the index (correct for post-commit reloads). The `apply_sort_order` split (active goes to `update_search`, committed goes to `rebuild_flat_items`) matches those semantics.
- **Reload-survival reasoning is sound.** Keeping `search_query`, not just matches, is required because `refresh_search_matches` wipes on an empty query and the reload tail fires it. Clearing the query on Enter would have silently destroyed matches on the first background reload. Correctly identified and tested.
- **Test coverage is thorough.** commit-keeps-state, reload survival, sort re-score (single-match stale-index guard), full n/N wrap, inverted Esc-clears path, dropped redundant test. All exercise the real `handle_key` path.

### CI note

`Cargo Test (linux, bare-core)` shows fail, but the failure is `intro::intro_theme_pick_persists_to_config` at `tests/e2e/intro.rs:136`, the theme-picker persistence e2e, unrelated to search. All 3688 unit tests including the four new search tests passed. This is pre-existing flaky e2e noise, not attributable to this diff. Recommend a re-run of that job.

### Verdict

Implementation: Approve. Project fit: Accept. Real, maintainer-endorsed bug with a diagnosed root cause; the helper refactor is the right call over 24 duplicated inline guards. No Blocker, Required, or Nit findings.

PR #2698, opened 2026-07-06
fix(cli): sanitize explicit worktree branches

fengjikui:
## Description
Fixes the CLI worktree branch path for `aoe add <repo> -w <name> -b` so explicit branch names are sanitized before they are passed to git.

The TUI/session-builder path already used `git_sanitize_branch_name`, and its nearby defense-in-depth comment calls out the CLI/API case where a user types a title-shaped string into the branch field. The CLI path only trimmed the raw worktree name and forwarded it directly to `GitWorktree::create_worktree`, so names such as `Exploration and issues v2` failed with an invalid git ref instead of creating the same sanitized branch that the rest of the session flow expects.

This PR reuses the existing sanitizer in the CLI path, treats blank explicit worktree branch values as absent so they fall back to normal title generation instead of branch `session`, and adds e2e regression coverage for both behaviors.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (not needed for this CLI bug fix)
- [x] For UI changes: included screenshot or recording (N/A: no UI changes)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## Validation

- `PATH="$HOME/.cargo/bin:$PATH" cargo test --features e2e-tests --test e2e test_cli_add_sanitizes_explicit_worktree_branch -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test --features e2e-tests --test e2e test_cli_add_blank_worktree_branch_falls_back_to_normal_title -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --check`
- `git diff --check`

## AI Usage

- [ ] No AI was used
- [x] AI was used

Used Codex to inspect the CLI/session-builder branch creation paths, reproduce the invalid-ref failure locally, implement the focused sanitizer reuse and blank-branch fallback, and run the validation commands listed above.

- Fixed `aoe add -w/--worktree` branch handling by sanitizing explicit user-provided worktree branch names before using them for git worktree/branch creation (so names with spaces/special characters won’t be treated as invalid git refs).
- Added shared parsing for the “explicit worktree branch” value that treats blank/whitespace-only input as “not provided,” tightening related CLI checks and ensuring worktree-specific logic only runs when there’s a real value.
- Updated CLI fallback behavior: when `-w/--worktree` is blank, the session title falls back to the normal generated title (instead of ending up with/creating a `session` branch), and `worktree_info` branch data isn’t set.
- Added/extended end-to-end CLI regression tests that verify:
  - explicit branch names like `Exploration and issues v2` are converted to a safe git branch form (e.g., `Exploration-and-issues-v2`) and the branch is created
  - whitespace-only `-w` doesn’t break the command, doesn’t set `worktree_info`, and doesn’t create an unexpected `session` branch
- Tweaked E2E first-run intro UI tests to wait for the selection marker before proceeding between steps.

Benefits:
- Prevents CLI failures caused by human-friendly worktree branch names that don’t map cleanly to git ref rules.
- Makes CLI behavior consistent with the existing session/TUI flow’s branch-name sanitization and blank-input handling.
- Adds coverage to prevent regressions in both “sanitized explicit branch” and “blank fallback” scenarios.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -403,7 +403,8 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
         use crate::session::WorktreeInfo;
         use chrono::Utc;
 
-        let branch = branch_raw.trim();
+        let branch_owned = builder::git_sanitize_branch_name(branch_raw);
+        let branch = branch_owned.as_str();
         let init_submodules = config.worktree.init_submodules && !args.no_submodules;
 
         if !all_extra_repos.is_empty() {
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -188,7 +188,8 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
         bail!("Path is not a directory: {}", path.display());
     }
 
-    if (!args.extra_repos.is_empty() || !args.projects.is_empty()) && args.worktree_branch.is_none()
+    if (!args.extra_repos.is_empty() || !args.projects.is_empty())
+        && explicit_worktree_branch(&args).is_none()
     {
         bail!("--repo/--project requires --worktree to specify a branch\nTip: aoe add /path --project repoB -w branch-name");
     }
@@ -277,7 +278,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
     // Fork intent appends. Reject these up front (before any resource creation)
     // rather than launch a fork that can't find its parent. See PR review.
     if args.fork_from.is_some() {
-        if args.worktree_branch.is_some() || args.create_branch {
+        if explicit_worktree_branch(&args).is_some() || args.create_branch {
             bail!(
                 "`--fork-from` cannot be combined with --worktree or --new-branch: a fork must run \
                  in the parent's working directory to resume its conversation."
@@ -398,7 +399,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
         None
     };
 
-    if let Some(branch_raw) = &args.worktree_branch {
+    if let Some(branch_raw) = explicit_worktree_branch(&args) {
         use crate::git::GitWorktree;
         use crate::session::WorktreeInfo;
         use chrono::Utc;
@@ -1205,8 +1206,8 @@ fn resolve_session_title(args: &AddArgs, instances: &[Instance]) -> Result<Strin
     if let Some(title) = &args.title {
         return Ok(title.trim().to_string());
     }
-    let default_title = if let Some(ref branch) = args.worktree_branch {
-        branch.trim().to_string()
+    let default_title = if let Some(branch) = explicit_worktree_branch(args) {
+        branch.to_string()
     } else {
         let existing_titles: Vec<&str> = instances.iter().map(|i| i.title.as_str()).collect();
         civilizations::generate_random_title(&existing_titles)
@@ -1216,6 +1217,13 @@ fn resolve_session_title(args: &AddArgs, instances: &[Instance]) -> Result<Strin
     } else {
         Ok(default_title)
     }
+}
+
+fn explicit_worktree_branch(args: &AddArgs) -> Option<&str> {
+    args.worktree_branch
+        .as_deref()
+        .map(str::trim)
+        .filter(|branch| !branch.is_empty())
 }
 
 fn prompt_session_title(default_title: &str) -> Result<String> {
--- a/tests/e2e/cli.rs
+++ b/tests/e2e/cli.rs
@@ -1380,6 +1380,52 @@ fn test_cli_add_attaches_to_existing_worktree() {
 
 #[test]
 #[serial]
+fn test_cli_add_sanitizes_explicit_worktree_branch() {
+    let h = TuiTestHarness::new("cli_branch_sanitize");
+    let project = h.home_path().join("branch-sanitize-project");
+    init_git_repo(&project);
+
+    let add_output = h.run_cli(&[
+        "add",
+        project.to_str().unwrap(),
+        "-w",
+        "Exploration and issues v2",
+        "-b",
+        "-t",
+        "SanitizedBranch",
+    ]);
+    assert!(
+        add_output.status.success(),
+        "aoe add with unsanitized explicit branch should succeed:\nstdout: {}\nstderr: {}",
+        String::from_utf8_lossy(&add_output.stdout),
+        String::from_utf8_lossy(&add_output.stderr)
+    );
+
+    let json = read_sessions_json(&h);
+    let sessions = json.as_array().expect("sessions array");
+    let session = sessions
+        .iter()
+        .find(|s| s["title"].as_str() == Some("SanitizedBranch"))
+        .expect("SanitizedBranch session should exist");
+    assert_eq!(
+        session["worktree_info"]["branch"].as_str(),
+        Some("Exploration-and-issues-v2"),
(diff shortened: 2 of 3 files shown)


jerome-benoit (commented):

## AoE Review

Reviewed: PR #2698 (base: main, head: 6fd9d5de). Files changed: 2 (`src/cli/add.rs`, `tests/e2e/cli.rs`).

Fix is correct and well-scoped: the CLI now funnels explicit worktree branch names through `builder::git_sanitize_branch_name`, matching the TUI/session-builder path at `src/session/builder.rs:833` and the peer's own "Defense-in-depth… caller is a CLI/API user typing a title-shaped string into the branch field" comment at `src/session/builder.rs:828-832`, which explicitly anticipated this CLI/API failure mode. Ownership pattern (`branch_owned` + `.as_str()`) mirrors the existing `leaf_seed_owned` idiom in the same function. The e2e test locks both the persisted JSON branch and the on-disk `git branch --list` output. No blocker.

### Suggested

- CLI empty/whitespace `-w` handling diverges from the peer. `src/session/builder.rs:827` pre-filters `worktree_branch.map(str::trim).filter(|b| !b.is_empty())` before sanitizing and derives from the title otherwise; `src/cli/add.rs:234` uses a bare `if let Some(branch_raw) = &args.worktree_branch`. `git_sanitize_branch_name("")` and `git_sanitize_branch_name(" ")` both return the literal `"session"` (per the helper's fallback test at `src/session/builder.rs:1057`). Combined with `resolve_session_title` at `src/cli/add.rs:1065` using `branch.trim().to_string()` unsanitized as the default title, `aoe add /repo -w "" -b` produces a session titled `""` on branch `"session"` with no fallback to a civilization-generated name. The peer path handles this by falling back to `generate_random_title` when the trimmed branch is empty. Fix: mirror the peer pre-filter at `src/cli/add.rs:234`:
  ```rust
  if let Some(branch_raw) = args
      .worktree_branch
      .as_deref()
      .map(str::trim)
      .filter(|b| !b.is_empty())
  {
      let branch_owned = builder::git_sanitize_branch_name(branch_raw);
      ...
  }
  ```
  And add the empty-branch fallback to `resolve_session_title` at `:1065` matching the peer at `src/session/builder.rs:794-800` shape.

- PR body doesn't acknowledge that the peer's `src/session/builder.rs:828-832` comment already predicted this exact CLI/API failure mode (verbatim: "the caller is a CLI/API user typing a title-shaped string into the branch field"). One-line addition to the PR description referencing the peer comment closes the traceability gap.

### Nit

- PR title `fix: sanitize explicit worktree branches` vs commit subject `fix(cli): sanitize explicit worktree branches`. Empirical count on the last 100 non-merge subjects: scoped Conventional Commits ~91.5%, bare `fix:` ~8.5%. On squash-merge the PR title becomes the commit subject. Fix: retitle to `fix(cli): sanitize explicit worktree branches` to match the commit and the codebase's dominant convention.

- Test coverage is scoped to the create path (`-b`) with an explicit title. Two branches remain uncovered by the new e2e: (a) the workspace-path branch at `src/cli/add.rs:242-294` (`!all_extra_repos.is_empty()`) which the same sanitize call also feeds, and (b) the default-title derivation at `:1065` where the raw un-sanitized branch shapes the session title. Both are exercised by unit tests on the helper, so this is defensible scoping; optional to extend.

### Follow-up items (out of scope for this PR)

Two pre-existing contract violations in `git_sanitize_branch_name` at `src/session/builder.rs:843-880` that this PR does not introduce but which its expanded usage surface raises visibility on. The helper's docstring at `src/session/builder.rs:830-833` asserts "libgit2 never sees a value it'll reject with InvalidSpec", but two inputs slip through:

- **`foo.lock.lock`** at `src/session/builder.rs:867-872`. `.strip_suffix(".lock").unwrap_or(seg)` strips one `.lock` per segment. `foo.lock.lock` becomes `foo.lock`, which `git branch foo.lock` rejects with `'foo.lock' is not a valid branch name`. Loop-strip fix: `while seg.ends_with(".lock") { seg = seg.trim_end_matches(".lock"); }` in the per-segment map. Add a unit test case alongside `test_git_sanitize_branch_name_strips_interior_lock_suffix` at `src/session/builder.rs:1063`.

- **Bare `HEAD`** passes through unchanged. `git_sanitize_branch_name("HEAD")` returns `"HEAD"`, then `git branch HEAD` errors. Companion `@` guard already exists at `src/session/builder.rs:879-881`; a symmetric `HEAD` → `"session"` mapping closes the gap.

These are TUI/API sanitizer defects predating this PR; worth a follow-up issue against `src/session/builder.rs` rather than expanding this fix's scope.

### Verdict

**Implementation:** Approve (comment only)
**Project Fit:** Accept

The fix is correct and closes a real CLI-vs-TUI/API asymmetry that the peer path's own comment already anticipated. Suggested items are edge-case parity with the peer (empty-branch handling) and PR-metadata polish (title scope tag). The two follow-up items are pre-existing sanitizer contract violations worth filing separately, not blocking this PR.

fengjikui:
Thanks for the detailed review. I pushed `82ad836e` on top of the refreshed `main` base:

- mirrored the peer pre-filter so blank/whitespace `--worktree` values are treated as absent instead of sanitizing to branch `session`;
- made `resolve_session_title` fall back to the normal generated title when the explicit worktree branch is blank;
- added an e2e regression for the blank-branch fallback and kept the existing explicit branch sanitization e2e;
- retitled the PR to `fix(cli): sanitize explicit worktree branches` and updated the description to reference the session-builder CLI/API defense-in-depth comment.

Validation:
- `PATH="$HOME/.cargo/bin:$PATH" cargo test --features e2e-tests --test e2e test_cli_add_sanitizes_explicit_worktree_branch -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test --features e2e-tests --test e2e test_cli_add_blank_worktree_branch_falls_back_to_normal_title -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --check`
- `git diff --check`

I kept the separate `git_sanitize_branch_name` contract issues (for repeated `.lock` suffixes and bare `HEAD`) out of this PR so this remains scoped to the CLI-vs-session-builder behavior.

Seluj78 (approved):

## AoE Review

**Implementation:** Approve
**Project Fit:** Accept

Real bug: human-readable `-w` names produced an invalid git ref instead of a sanitized branch. Minimal surgical scope; brings the CLI into parity with the TUI/session-builder path that already sanitized via `git_sanitize_branch_name`.

Verified:
- All four semantic uses of `worktree_branch` (validation, fork-from guard, worktree creation, title resolution) now route through `explicit_worktree_branch`, so blank/whitespace `-w` is uniformly treated as absent. No stray direct reads remain.
- Blank-branch fallback matches builder semantics: skips worktree creation and falls back to the generated title instead of creating branch `session`.
- Tests cover happy path (sanitize) and edge path (blank fallback), both with `#[serial]`, temp git repos, and real git branch-state assertions. Codecov reports all modified lines covered.

Nit (non-blocking): branch name `codex/cli-sanitize-worktree-branch` is off the `fix/` convention in AGENTS.md.

Note: CodeRabbit title-check flags `fix(cli):`, but the repo Lint PR title check passed and scoped Conventional Commit prefixes are valid; repo policy wins.

PR #2700, opened 2026-07-07
fix(session): start empty Claude threads fresh-pinned, not --resume

Eric162:
## Description

An empty Claude thread that is killed before its first prompt lands in a permanent `resume failed for sid <uuid>; preserved for explicit retry` error, and pressing `e` (restart) does not durably recover it.

**Root cause.** A Claude session is minted a session UUID at first launch, before the user sends any message. Claude only writes its transcript (`~/.claude/projects/<encoded-cwd>/<uuid>.jsonl`) once the conversation has real content, so an "empty thread" has a stored `agent_session_id` that never hit disk. On restart, `acquire_session_id` trusted that stored id unconditionally (`is_existing = true`) and launched `claude --resume <uuid>`, which fails with "No conversation found" every time. The pane dies inside the resume probe, so `start_with_resume_fallback` reports `ResumeFailed` and preserves the sid.

The per-sid loop-breaker (`resume_probe_failed_sid`, #2612) only routes the *next explicit* restart to a fresh launch, and that fresh launch re-mints another empty UUID that fails again on its next kill, so the failure recurs rather than resolving.

**Fix.** Gate `is_existing` in `acquire_session_id` on whether Claude actually has a transcript on disk for the stored sid (new `capture::claude_host_transcript_confirmed_absent`). When the transcript is confirmed absent, launch a fresh pinned session (`--session-id <sid>`, which succeeds and keeps the id stable so a later first prompt stays continuous) instead of `--resume <sid>`. The check is existence-only (no mtime gate), so a real but idle conversation whose jsonl is older than the live-capture window still resumes. It is scoped to Claude (the only tool AoE pre-mints a UUID for) and host-only (a sandboxed transcript lives in a container that may not be up at acquire time; noted in the code as a possible follow-up).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

Added unit coverage for the new behavior:
- `capture::tests::test_claude_host_transcript_confirmed_absent` — present (even stale) vs missing vs never-created dir.
- `instance::...::verify_on_resume::stored_sid_without_transcript_launches_fresh_pinned` — the empty-thread case now launches fresh-pinned.
- `instance::...::verify_on_resume::stored_sid_with_stale_transcript_still_resumes` — regression guard: an idle real conversation must still `--resume`.

Updated the existing resume-probe/cascade tests to seed a real transcript (they represent resuming an existing conversation) and adjusted the acquire/idempotence tests to the new contract.

**TUI / CLI (e2e test)**
- [x] N/A: covered by the existing `restart_selected_session_surfaces_resume_failed_after_async_restart` TUI test (updated to seed a transcript) and the `verify_on_resume` unit suite; this is a session-id acquisition change with no new TUI-render or CLI-subcommand surface.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Root-cause investigation and implementation done via Claude Code; author reviewed and verified `cargo fmt`, `cargo clippy --all-targets`, and the full test suite locally.

## Summary

- Fixed Claude session restarts for empty threads without saved transcripts.
- Added transcript presence checks to distinguish resumable sessions from sessions that need a fresh launch.
- Preserved stored session IDs when starting fresh, maintaining future conversation continuity.
- Added regression coverage for missing, stale, and existing transcripts, including restart behavior.

## Benefits

- Prevents permanent resume failures for sessions interrupted before their first prompt.
- Keeps valid conversations resumable, including those with older transcripts.
- Makes restart behavior more reliable and predictable.

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -118,6 +118,39 @@ pub(crate) fn capture_claude_session_id(
     anyhow::bail!("No active Claude session found for {}", project_path)
 }
 
+/// Whether we can affirmatively prove Claude has *no* persisted transcript for
+/// `session_id` under `project_path` on the host filesystem.
+///
+/// Claude only writes `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl` once a
+/// conversation has real content. A session AoE minted a UUID for but that was
+/// killed before the first prompt (an "empty thread") therefore has a stored
+/// `agent_session_id` that never hit disk, and `claude --resume <uuid>` on it
+/// fails with "No conversation found" every time. Callers use this to launch
+/// such an id as a fresh pinned session (`--session-id <uuid>`) instead of a
+/// guaranteed-to-fail `--resume`.
+///
+/// Returns `true` ONLY when the Claude home resolves and the transcript file is
+/// confirmed missing. Any uncertainty (home dir unresolved) returns `false` so
+/// the caller preserves the existing `--resume` attempt rather than risk
+/// downgrading a real conversation to a fresh start. The check is
+/// existence-only (no mtime freshness gate), so an idle-but-real conversation
+/// whose jsonl is older than the live-capture window is still reported present.
+pub(crate) fn claude_host_transcript_confirmed_absent(
+    project_path: &str,
+    session_id: &str,
+) -> bool {
+    let Ok(claude_home) = resolve_agent_home(Some("CLAUDE_CONFIG_DIR"), ".claude") else {
+        return false;
+    };
+    let canonical = canonicalize_or_raw(project_path);
+    let dir_name = encode_claude_project_path(&canonical.to_string_lossy());
+    let transcript = claude_home
+        .join("projects")
+        .join(dir_name)
+        .join(format!("{session_id}.jsonl"));
+    !transcript.is_file()
+}
+
 /// Scan `~/.claude/projects/{encoded-path}/` and pick this poller's session.
 ///
 /// Tie-break:
@@ -2331,6 +2364,50 @@ mod tests {
 
         let result = capture_claude_session_id("/tmp/myproject", None, &HashSet::new());
         assert_eq!(result.unwrap(), uuid_new);
+
+        match old_val {
+            Some(v) => std::env::set_var("CLAUDE_CONFIG_DIR", v),
+            None => std::env::remove_var("CLAUDE_CONFIG_DIR"),
+        }
+    }
+
+    #[test]
+    #[serial]
+    fn test_claude_host_transcript_confirmed_absent() {
+        let tmp = tempfile::tempdir().unwrap();
+        let project_dir = tmp.path().join("projects").join("-tmp-myproject");
+        std::fs::create_dir_all(&project_dir).unwrap();
+
+        let present = "11111111-2222-3333-4444-555555555555";
+        let missing = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee";
+        let file = project_dir.join(format!("{present}.jsonl"));
+        std::fs::write(&file, "data\n").unwrap();
+        // Existence-only: an old mtime (past the live-capture window) must not
+        // read as absent, or an idle real conversation would lose its resume.
+        let hour_ago = std::time::SystemTime::now() - Duration::from_secs(3600);
+        std::fs::File::options()
+            .write(true)
+            .open(&file)
+            .unwrap()
+            .set_times(std::fs::FileTimes::new().set_modified(hour_ago))
+            .unwrap();
+
+        let old_val = std::env::var("CLAUDE_CONFIG_DIR").ok();
+        std::env::set_var("CLAUDE_CONFIG_DIR", tmp.path());
+
+        assert!(
+            !claude_host_transcript_confirmed_absent("/tmp/myproject", present),
+            "a transcript on disk (even stale) must not be reported absent"
+        );
+        assert!(
+            claude_host_transcript_confirmed_absent("/tmp/myproject", missing),
+            "an unwritten sid must be reported confirmed-absent"
+        );
+        // A project dir that was never created is also confirmed-absent.
+        assert!(claude_host_transcript_confirmed_absent(
+            "/tmp/never-opened-project",
+            present
+        ));
 
         match old_val {
             Some(v) => std::env::set_var("CLAUDE_CONFIG_DIR", v),
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1853,6 +1853,34 @@ impl Instance {
                 self.agent_session_id = Some(fresh.clone());
                 return (Some(fresh), true);
             }
+            // A stored Claude sid with no transcript on disk is not resumable:
+            // Claude minted the UUID at first launch but nothing was ever
+            // written (an empty thread killed before the first prompt), so
+            // `--resume <sid>` is a guaranteed launch failure that lands the
+            // session in the "resume failed for sid ...; preserved for explicit
+            // retry" state. Launch it as a fresh pinned session instead
+            // (`is_existing = false` -> `--session-id <sid>`), which succeeds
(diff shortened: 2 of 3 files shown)


Seluj78 (approved):

## AoE Review

Rust-only bug fix, 3 files (session, tui-tests). No unresolved review threads; no actionable bot comments.

### Verdict: Approve

Tight, well-scoped fix for a real recurring bug (empty-thread leading to a doomed `--resume` loop, lineage of #2612). Root cause and safety envelope are documented in-code: existence-only check (no mtime gate) preserves resume for idle-but-real conversations, home-unresolved returns false so it fails safe toward `--resume`, and the Claude-only plus host-only scoping is justified (Claude is the only pre-minted-UUID tool; sandboxed transcript noted as follow-up).

Test coverage is thorough: present/stale/missing/never-created probe cases, the empty-thread fresh-pinned path, a stale-resume regression guard, and the existing cascade/TUI tests correctly reseeded. Placement of the new branch (after the live-capture scan, before the resume return) is correct.

### Nit (non-blocking)

`test_claude_host_transcript_confirmed_absent` uses manual `CLAUDE_CONFIG_DIR` save/restore that leaks on panic; sibling `verify_on_resume` tests use the RAII `ClaudeHomeGuard`. Reuse it for panic-safe cleanup under `#[serial]`.

CI was pending at review time; confirm green before merge.

Seluj78:
Rebase needed :)

jerome-benoit (approved):
## AoE Review

**Implementation:** sound. **Project fit:** good.

The transcript gate at `src/session/instance.rs:1825-1839` closes the empty-thread resume loop. Correct four-way gating (tool == "claude" AND !is_sandboxed AND `ResumeIntent::Default` AND transcript confirmed absent), correct placement in the `Default` fallthrough of `acquire_session_id` (fn at `:1769`) so `Use` / `Cleared` / `Fork` intents keep their existing semantics, correct asymmetry in `claude_host_transcript_confirmed_absent` at `src/session/capture.rs:142-144` where home-resolution failure returns `false` (uncertain) rather than `true`, letting the pre-existing `resume_probe_failed_sid` handle the second retry. Loop-break holds because fresh-pinned takes `LaunchSidOutcome::Fresh` at `src/session/instance.rs:3810-3855` with `attempted_sid = None`, bypassing the probe entirely and letting any launch error propagate through the normal `?` path rather than into a resume-retry loop.

Net coverage improves: the four dropped `assert!(is_existing)` sites would have become host-dependent post-fix (they read the developer's real `~/.claude`), and the two new hermetic tests under `mod verify_on_resume` (`stored_sid_without_transcript_launches_fresh_pinned` at `:8593`, `stored_sid_with_stale_transcript_still_resumes` at `:8621`) lock the fresh-pinned and stale-transcript-resumes branches deterministically via `ClaudeHomeGuard` isolating `CLAUDE_CONFIG_DIR`. The stale-transcript test using a 3600s mtime gap against the 5-minute live-capture window at `src/session/capture.rs:94, 112, 213, 221` is the correct witness for the "existence-only, no mtime" invariant documented at `capture.rs:131-137`.

### Required

None.

### Suggested

**Lock idempotence of the fresh-pinned decision.** `stored_sid_without_transcript_launches_fresh_pinned` at `src/session/instance.rs:8593` calls `acquire_session_id()` once. The pre-fix invariant "second call on stored sid returns `is_existing=true`" was carried by `apply_session_flags_returns_acquire_is_existing` at `:7037`, which the restructure moved to a `ResumeIntent::Use` short-circuit assertion (a valid but different invariant). Adding a second `acquire_session_id()` call plus the same `(Some(stored), false)` assertion to `stored_sid_without_transcript_launches_fresh_pinned` locks that a state mutation between acquires (accidental clear of `sid_owners`, transient `agent_session_id` mint) cannot regress.

**Split the theme-picker hardening into its own commit.** `tests/e2e/intro.rs:123-138` fixes a real determinism race in the intro theme picker (adding `wait_for("▶ zinc")` before `Down` and `wait_for("▶ empire")` after), unrelated to Claude session semantics. Splitting the intro-test hunk into its own commit keeps `git blame` pointed at a test-flake fix rather than at the Claude sid fix. On rebase this hunk collides semantically with #2698's `intro.rs:126-130` (which added the same post-`Down` `wait_for("▶ empire")`). The PR's version is a strict superset (adds the initial-render wait); resolution is keep both waits, and #2698's separate `wait_for("▶ Tmux mode")` in `intro_lets_user_choose_tmux_attach` at `intro.rs:174` should carry forward untouched.

### Follow-up

**Filing suggested: sandboxed Claude empty-thread bug remains.** Explicitly deferred at `instance.rs:1823-1824` with the comment noting the transcript lives inside the container. Sandboxed sessions hit the original permanent-fail path because the pre-mint at `:1858` runs unconditionally and the transcript gate is short-circuited by `!self.is_sandboxed()` at `:1826`. Two paths: probe the container's `~/.claude/projects/<enc>/<uuid>.jsonl` via `docker exec` (introduces container-up dependency), or skip pre-minting for sandboxed Claude and lose `--session-id` pinning inside the container. Neither is cheap; a follow-up issue keeps it trackable.

**Filing suggested: `ResumeIntent::Use(sid)` on an empty-thread sid loops on `--resume` probe failure without self-healing.** Not a defect this PR introduces. `Use(sid)` short-circuits `acquire_session_id` at `instance.rs:1770-1774` returning `(Some(sid), true)`; `LaunchSidOutcome::Existing` fires the resume probe; probe fails; `mark_resume_probe_failed` at `:3072-3133` sets the marker without touching `resume_intent`; the loop-breaker at `:3775-3799` is gated `if self.resume_intent == ResumeIntent::Default`. Next launch replays the same short-circuit and the same probe failure. Options for the follow-up: extend the loop-breaker to fire under `Use`, route `mark_resume_probe_failed` through a `persist_session_id`-style promotion for `Use`, or document `Use` as user-authoritative with `status=Error` as the intended terminal state. Direction can be decided in the follow-up issue.

**Docs update: `docs/development/internals/sessions.md:7`.** Post-merge, the relaunch flow is `--session-id <uuid>` when the transcript is absent, `--resume <uuid>` otherwise. Current text says the relaunch is always `--resume`.

### Nit

Consider renaming `apply_session_flags_returns_acquire_is_existing` at `instance.rs:7037` to reflect that the second call specifically covers the `ResumeIntent::Use` short-circuit branch of `acquire_session_id` (e.g. `apply_session_flags_returns_true_for_use_pin`), or split into two named tests. The old name implied full acquire coverage that the restructure narrowed.

Eric162:
> Rebased onto latest `main` (`188b09d9`) — thanks @seluj78.
> 
> One note on the history so nothing looks lost: `main` had independently landed a fix for the intro theme-picker e2e test (`tests/e2e/intro.rs`) — the deterministic `▶ empire` post-Down wait, but without a pre-Down gate. That superseded your `a0266da7` and my `zinc` follow-up, so I dropped both during the rebase. `tests/e2e/intro.rs` is now byte-identical to `main`, and the branch is a single commit — just the session-id fix.
> 
> The earlier codecov failure (`session-fork.spec.ts`, a Playwright `toBeVisible()` race) was on the pre-rebase commit and is unrelated to this change: forking takes the `ResumeIntent::Fork` early-return, and the transcript gate only affects the `ResumeIntent::Default` path. It should clear on the CI re-run.
> 
> Verified locally against the new base: `cargo fmt --check`, `cargo clippy --all-targets` (both clean), and the full session test suite (1302 passing). Re-approval + a workflow-approval on the new commit is all that's needed on your side.

claude posted that against my prior instructions sorry
- E

njbrake (approved):

Reviewed in batch triage. Empty Claude threads launch fresh-pinned (--session-id) instead of the doomed --resume; scoping is claude + Default intent + non-sandboxed, fails safe, thorough probe-case tests. The failing Playwright (live) jobs are a known session-fork flake unrelated to this Rust-only change. LGTM.

PR #2710, opened 2026-07-07
fix(tui): keep Trash collapsed on delete and skip trash in w-navigation

Eric162:
## Description

Two bugs in the trash feature:

1. **Collapsed Trash group re-expands on every delete.** `trash_session_by_id` called `reveal_trashed_section()` on every delete, unconditionally forcing the Trash section open again. There is no bulk-trash (unlike archive, where whole-group archive justifies revealing), so single-row delete now leaves the collapse state untouched, matching the single-row-archive behavior where the header count is the feedback. Removed the now-dead `reveal_trashed_section`.

2. **`w` (jump to next needing-attention) cycles through unread trash items.** `jump_to_next_waiting` never filtered trashed rows, so an unread trashed session counted as actionable. `trash()` preserves the `unread` flag by design (for faithful restore), so trash rows showed up in the `w`-cycle whenever the Trash section was expanded (compounded by bug 1 keeping it open). Both passes of `jump_to_next_waiting` now skip trashed instances.

No render change was needed: the trash lifecycle override in `render.rs` already runs after the unread logic, so trash rows paint dimmed with a stopped glyph and never show an unread dot.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (doc comment on `trash_session_by_id`)
- [ ] For UI changes: included screenshot or recording (behavior-only, no visual change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated a test: two in-module regression tests in `src/tui/home/tests.rs` — `trashing_leaves_collapsed_trash_section_collapsed` (a collapsed Trash section stays collapsed after a delete) and `w_skips_unread_trashed_session` (`w` never lands on a trashed session even when it is unread).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Used to investigate the trash/attention code paths and draft the fix + regression tests.

This PR makes Trash behavior in the TUI more consistent and less disruptive:

- Keeping Trash collapsed now stays collapsed after deleting a session.
- `w` navigation now ignores trashed sessions so they no longer get selected by mistake.
- Added regression tests for both cases.
- Updated the trashing doc comment to match the new behavior.

Benefit: the sidebar no longer jumps open unexpectedly, and keyboard navigation is less confusing when deleted items are still present.

### DIFF
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -3262,9 +3262,13 @@ impl HomeView {
                 _ => continue,
             };
             if let Some(inst) = self.get_instance(&id) {
-                let is_actionable = inst.status == Status::Waiting
-                    || matches!(inst.idle_age(), Some(age) if age < window)
-                    || (crate::session::unread_enabled() && inst.is_unread());
+                // Trashed rows are stopped and only surface under the collapsed
+                // Trash section; they never "need attention", so skip them even
+                // when a stale unread flag survived the trash (#2489).
+                let is_actionable = !inst.is_trashed()
+                    && (inst.status == Status::Waiting
+                        || matches!(inst.idle_age(), Some(age) if age < window)
+                        || (crate::session::unread_enabled() && inst.is_unread()));
                 if is_actionable {
                     self.cursor = idx;
                     self.update_selected();
@@ -3288,6 +3292,9 @@ impl HomeView {
             let Some(inst) = self.get_instance(&id) else {
                 continue;
             };
+            if inst.is_trashed() {
+                continue;
+            }
             if inst.status != Status::Idle {
                 continue;
             }
--- a/src/tui/home/mod.rs
+++ b/src/tui/home/mod.rs
@@ -4081,12 +4081,6 @@ impl HomeView {
         });
     }
 
-    /// Expand the synthetic Trash section if collapsed. Used when trashing a
-    /// session so the user sees where the row went. No-op when already open.
-    pub(super) fn reveal_trashed_section(&mut self) {
-        self.trashed_section_collapsed = false;
-    }
-
     pub fn toggle_trashed_section(&mut self) {
         self.trashed_section_collapsed = !self.trashed_section_collapsed;
         self.rebuild_flat_items();
--- a/src/tui/home/operations.rs
+++ b/src/tui/home/operations.rs
@@ -1458,8 +1458,9 @@ impl HomeView {
     /// Move a session to the trash: stop its tmux sessions (a structured-view
     /// worker is reaped by the daemon reconciler once the row reads trashed)
     /// and set `trashed_at`. Durable artifacts are kept so it can be
-    /// restored. The Trash section is revealed so the user sees where the row
-    /// went. See #2489.
+    /// restored. The Trash section's collapse state is left untouched: like
+    /// single-row archive, the section header's count is the feedback, so a
+    /// user who collapsed it stays collapsed (#2489).
     pub(super) fn trash_session_by_id(&mut self, id: &str) {
         if let Err(e) = self.apply_user_action(id, |inst| inst.trash()) {
             tracing::warn!(target: "tui.session", session = %id, "trash failed: {e}");
@@ -1483,7 +1484,6 @@ impl HomeView {
         if let Some(reason) = relocate_warning {
             tracing::warn!(target: "tui.session", session = %id, "trash worktree relocation skipped: {reason}");
         }
-        self.reveal_trashed_section();
         self.rebuild_flat_items();
         self.cursor = self.cursor.min(self.flat_items.len().saturating_sub(1));
         self.update_selected();
--- a/src/tui/home/tests.rs
+++ b/src/tui/home/tests.rs
@@ -7175,6 +7175,78 @@ fn trash_then_restore_round_trip() {
     );
 }
 
+/// Regression for #2489: trashing a session must not re-expand a Trash
+/// section the user has collapsed. Like single-row archive, the section
+/// header's count is the feedback; the collapse state is left untouched.
+#[test]
+#[serial]
+fn trashing_leaves_collapsed_trash_section_collapsed() {
+    let mut env = create_test_env_with_sessions(2);
+    assert!(
+        env.view.trashed_section_collapsed,
+        "Trash section defaults to collapsed"
+    );
+    let id = env.view.instances[0].id.clone();
+    env.view.selected_session = Some(id.clone());
+
+    env.view.trash_session_by_id(&id);
+
+    assert!(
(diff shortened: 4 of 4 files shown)


jerome-benoit (approved):
## AoE Review

Reviewed: PR #2710 (base: `main` @ `59337734`, head: `4df3b0fe`). Files changed: 4 (`src/tui/home/input.rs`, `src/tui/home/mod.rs`, `src/tui/home/operations.rs`, `src/tui/home/tests.rs`). +85/-12.

Both bugs are real and the fixes are minimal + correct.

**Bug 1 (auto-reveal on trash)**: removing the `self.reveal_trashed_section()` call at ex-`src/tui/home/operations.rs:1486` and deleting the now-unused helper at ex-`src/tui/home/mod.rs:4084-4088` collapses cleanly. Post-diff grep is zero for `reveal_trashed_section`. Single-row archive symmetry claim checks out: [`toggle_archive_at_cursor` at `src/tui/home/operations.rs:1373`](src/tui/home/operations.rs) (the single-row path invoking `inst.archive()` at `:1413`) does not call `reveal_archived_section()`, and the [`reveal_archived_section` doc at `src/tui/home/mod.rs:4068-4082`](src/tui/home/mod.rs) states the same rationale ("Single-row archive does NOT reveal: the cursor advances to the next active session instead and the section header's count is the feedback"). Only bulk [`archive_selected_group` at `src/tui/home/operations.rs:1579`](src/tui/home/operations.rs) reveals, via its `self.reveal_archived_section()` call at `:1607`. Post-PR trash matches, and there is no bulk-trash counterpart to require a divergent branch.

**Bug 2 (`w` cycles trashed rows)**: both hunks in [`jump_to_next_waiting` at `src/tui/home/input.rs:3265-3271` and `:3292-3297`](src/tui/home/input.rs) are necessary. Trace on the Pass 2 guard: [`Instance::trash` at `src/session/instance.rs:1525-1529`](src/session/instance.rs) sets `trashed_at` only. [`kill_all_tmux_sessions` at `src/session/instance.rs:4037-4048`](src/session/instance.rs) takes `&self`, so it cannot mutate `status`. Nothing in [`trash_session_by_id` at `src/tui/home/operations.rs:1464-1490`](src/tui/home/operations.rs) mutates `status` synchronously; the reconciler flips it asynchronously by observing tmux death. Therefore a just-trashed row that was `Status::Idle` retains `Status::Idle` in the window between the trash-action and reconciler observation. Pass 2 ranks Idle rows by `last_accessed_at` and would strongly prefer the just-trashed row (recently interacted with) over any surviving active row. The early-exit closes that window.

Pass 1's refactor `!inst.is_trashed() && (Waiting || fresh-idle || unread)` is semantically identical to the original for non-trashed rows (Rust `&&` binds tighter than `||`, and the disjunction is explicitly parenthesized). No regression on the non-trash path.

The two new tests are appropriate. `trashing_leaves_collapsed_trash_section_collapsed` at [`src/tui/home/tests.rs:7183`](src/tui/home/tests.rs) fails on pre-fix code (which sets `trashed_section_collapsed = false` in `reveal_trashed_section`). `w_skips_unread_trashed_session` at [`src/tui/home/tests.rs:7210`](src/tui/home/tests.rs) coincidentally exercises both guards: with the 2-session seed, removing Pass 1 alone lets Pass 1 land on trashed; removing Pass 2 alone lets Pass 2's `last_accessed_at` rank pick the trashed Idle row. Both paths currently fail; the negative assertion `landed != trashed` catches either.

Style compliance: no emdash, no ` -- ` prose in new comments or the updated docstring; comments explain "why" (issue citation, invariant statement), not "what"; PR removes dead code rather than annotating it. `fix(tui):` scope is idiomatic per codebase convention (91% scoped, precedents #2622, #2604, #2578, #2582); CodeRabbit's warning is a false positive. Title length is 71 chars, under CodeRabbit's 72 target.

### Suggested

- **Add a 3-session positive-landing test** locking Pass 1's "skip then continue" semantics that the shipped 2-session test does not cover. A future refactor swapping the `continue` for a `break` after skipping trashed would silently regress: Pass 1 would abort early, and Pass 2 would fall back to a non-trashed Idle row, so `landed != trashed` still holds and the shipped test passes. A 3-session seed with `[a(Idle), trashed(Idle+unread), waiting(Waiting)]` cursor on `a` catches it via `assert_eq!(landed, Some(waiting))`. Copy-adapt the existing `w_skips_unread_trashed_session` at [`src/tui/home/tests.rs:7210`](src/tui/home/tests.rs); add `mutate_instance(&waiting, |i| i.status = Status::Waiting)` before the trash-action. Fails on pre-fix, passes on post-fix, and locks the positive intent behind a positive assertion.

### Follow-up items (out of scope for this PR)

- **Archive-parallel `w`-navigation gap**. [`jump_to_next_waiting` at `src/tui/home/input.rs:3265-3271`](src/tui/home/input.rs) does not filter `is_archived()`. [`Instance::archive` at `src/session/instance.rs:1502-1507`](src/session/instance.rs) clears `favorited_at`, `snoozed_until`, `pinned_at`, but does not clear `unread`. If a user expands the Archived section (`archived_section_collapsed = false`), an archived-unread row leaks into Pass 1 for the same reason trashed rows did pre-fix. Same class of bug, pre-existing, not introduced here. The render side already excludes archived rows from unread-attention treatment at [`src/tui/home/render.rs:1166-1170`](src/tui/home/render.rs), so `w` diverges from the render's stated intent. Worth a separate follow-up issue.

- **Cursor clamp lands on Trash header when the last active row is trashed**. When the cursor is at the pre-trash `len - 1` position (the last active row), `cursor.min(len - 1)` at [`src/tui/home/operations.rs:1488`](src/tui/home/operations.rs) lands on the Trash section header (an `Item::Group`) once the trashed row is hidden under the still-collapsed default. Not a crash: [`update_selected` at `src/tui/home/input.rs:3479-3520`](src/tui/home/input.rs) handles `Item::Group` (clears `selected_session`/`selected_group` for trash paths); Enter routes to `toggle_group_collapsed` for the Trash header. Mirrors the pre-existing archive symmetry (`archive_selected_group` at `src/tui/home/operations.rs:1579` has the same clamp shape at `:1428`), so it is not a regression from this PR. Minor UX quirk worth surfacing as an optional polish item.

### Verdict

**Implementation:** Approve.
**Project Fit:** Accept.

Fix is scoped, correct, symmetric with the archive precedent already documented in the codebase, and closes both reported bugs without touching the non-trash paths. The Suggested test is a lightweight refactor guard, not a fix for a missing invariant. Both Follow-up items are pre-existing behaviors worth filing separately.

PR #2715, opened 2026-07-07
feat(serve): configurable pane-text rules with command actions

BTForIT:
## Description

Adds configurable pane-text rules: a daemon background loop that watches each registered session's tmux pane tail for user-defined regex patterns and runs a command when one matches. This turns "when a pane says X, do Y" into pure configuration; no code changes needed to add a detector.

Motivating examples: notify your phone when an agent is stuck on a device-code sign-in, log or rotate work when a provider usage-limit banner appears, or page a human when an agent prints your team's escalation phrase.

**How it works:**

- New `src/pane_rules.rs`: a small rule engine. Each `[[watchdog.rules]]` config entry declares a `pattern` regex plus `negative` guard regexes, a bounded `tail_lines` window (counting non-empty lines from the live edge, so stale scrollback can't fire), `line` or `window` scope (window scope joins the tail for multi-line prompts), optional TUI decoration stripping (selector glyphs, bullets, `1.`/`2)` enumerators, so anchored patterns survive menu chrome), and `priority` ordering. ANSI escapes are stripped before matching. Compilation is fail-open: an invalid regex drops that rule with a `tracing` warning and the rest keep working.
- New `src/server/pane_watchdog.rs`: the daemon loop. On a configurable interval (default 180s, min 10s) it captures each non-structured session's pane tail off the runtime via `spawn_blocking`, classifies it, and on a match runs the rule's `command` via `sh -c` with `AOE_RULE_NAME`, `AOE_RULE_KIND`, `AOE_SESSION_ID`, `AOE_SESSION_TITLE`, and `AOE_SESSION_PROFILE` in the environment. Commands run detached from the tick loop, are killed after 60s, and are rate-limited by a per-session-per-rule `cooldown_secs` (default 30 min). Matches are logged whether or not a command is set, so a new rule can be dry-run from the daemon log.
- Config: new `#[serde(default)]` `[watchdog]` section on `Config` (`disabled`, `interval_secs`, `rules`). The loop is a no-op when no rules are configured, and never runs in read-only mode since it executes user commands.
- Docs: new guide `docs/guides/pane-rules.md` with a field reference and worked examples, wired into the website sync script and nav.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (N/A: no UI change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: the feature is a daemon-internal background task with no TUI rendering or CLI subcommand surface. The rule engine (parsing, compilation fail-open, scoping, guards, decoration stripping, windowing, priority) is covered by 12 unit tests in `src/pane_rules.rs`; the daemon loop is thin glue over existing tested primitives (`capture_pane`, `load_all_instances`, `spawn_supervised`) and observable only through the tracing log.

Verified locally: `cargo fmt`, `cargo clippy --features serve --all-targets` (clean), `cargo test --features serve --lib` (4874 passed; the one failure, `tui::dialogs::hooks_install::tests::test_content_shows_settings_path`, fails identically on a clean `main` checkout in this environment).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Code (Claude Fable 5)

Design, implementation, tests, and docs were AI-authored under human direction and review.

This PR adds a configurable “pane rules” feature to the `serve` daemon, so it can continuously watch tmux pane output and react automatically when user-defined text patterns appear.

What changed / added:
- New pane-rules engine (`src/pane_rules.rs`)
  - Lets you define regex rules with priorities, optional negative “guard” patterns, optional line vs window matching scope, tail-window limits, and optional decoration/ANSI stripping for more reliable matching.
  - Compiles rules with a fail-open approach: invalid regexes won’t break rule processing; they’re warned about and skipped.
  - Detects duplicate rule names at compile time and drops later duplicates to keep cooldown tracking unambiguous.
- New background watchdog (`src/server/pane_watchdog.rs`)
  - Runs a periodic loop that tails each registered session’s tmux pane output, applies the compiled rules, logs matches, enforces per-(session, rule) cooldowns, and (optionally) runs a configured shell command on match.
  - Command execution runs detached from the scan loop, is time-limited (with failures/timeouts logged), and receives match/session context via environment variables.
  - Disabled when the daemon is in read-only mode or when no rules are configured.
- Config wiring
  - Adds a new `watchdog` section to session config (`src/session/config.rs`) to configure the watchdog and rules.
- Public API exposure
  - Exposes the new module (`pub mod pane_rules;` in `src/lib.rs`).
- Server integration
  - Hooks the watchdog into server startup (`src/server/mod.rs`).
- Documentation + website integration
  - Adds `docs/guides/pane-rules.md` and wires it into the site navigation and docs sync (`website/scripts/sync-docs.mjs`, `website/src/data/docsNav.ts`).

Benefits:
- More responsive, automated behavior for sessions (including dry-run style matching via “log-only” matches).
- Safer operations: invalid rules fail open, and cooldowns prevent command spam.
- Better reliability for real-world pane text via decoration/ANSI handling and bounded tail windows.

Potential follow-up:
- Expand examples and guidance for complex rule sets (especially around scope/anchors and negative guards).
- Add more observability around command outcomes and match frequency (beyond startup/match logs).

### DIFF
--- a/docs/guides/pane-rules.md
+++ b/docs/guides/pane-rules.md
@@ -1,0 +1,110 @@
+# Pane Rules
+
+Pane rules let the `aoe serve` daemon watch every session's terminal output for text you care about and run a command when it appears. Think of them as "when a pane says X, do Y": detect a usage-limit banner and rotate work to another account, catch a sign-in prompt and send yourself a push notification, or flag any agent that prints a phrase your team uses for "I need a human".
+
+The daemon periodically captures the tail of each registered session's tmux pane, runs your rules against it, and executes the matching rule's command with the session context in its environment. Rules are pure configuration; no code changes are needed to add one.
+
+Pane rules require a running daemon (`aoe serve`). A read-only daemon never runs pane rules, since they execute user commands.
+
+## Quick start
+
+Add a `[watchdog]` section to `~/.agent-of-empires/config.toml` (or your platform's config path):
+
+```toml
+[[watchdog.rules]]
+name = "usage-cap"
+kind = "cap"
+pattern = '(?i)^usage limit reached'
+command = "~/bin/notify-capped.sh"
+```
+
+Restart the daemon. Every scan (default: every 3 minutes), any session whose recent pane output contains a line starting with "usage limit reached" runs `~/bin/notify-capped.sh` with these variables in its environment:
+
+| Variable | Value |
+|----------|-------|
+| `AOE_RULE_NAME` | The rule's `name` ("usage-cap") |
+| `AOE_RULE_KIND` | The rule's `kind` ("cap") |
+| `AOE_SESSION_ID` | The matched session's id |
+| `AOE_SESSION_TITLE` | The matched session's title |
+| `AOE_SESSION_PROFILE` | The profile the session belongs to |
+
+The command runs via `sh -c`, is killed after 60 seconds, and its failures are logged, never fatal.
+
+## Rule reference
+
+```toml
+[watchdog]
+# disabled = true          # turn the watchdog off without deleting rules
+# interval_secs = 180      # scan cadence (minimum 10)
+
+[[watchdog.rules]]
+name = "device-code"           # required: label for logs and AOE_RULE_NAME
+kind = "auth"                  # optional: free-form category, AOE_RULE_KIND
+pattern = '(?is)devicelogin.*code'   # required: regex (Rust regex syntax)
+negative = ['(?i)example']     # optional: text matching any of these never fires the rule
+tail_lines = 8                 # optional: how many recent non-empty lines are in scope (default 15)
+scope = "window"               # optional: "line" (default) or "window"
+strip_decoration = true        # optional: strip leading TUI chrome before matching (default true)
+priority = 1                   # optional: lower fires first when several rules match (default 0)
+enabled = true                 # optional: set false to keep a rule but skip it
+command = "notify.sh"          # optional: run on match; omit to only log
+cooldown_secs = 1800           # optional: per-session refire floor (default 30 minutes)
+```
+
+Field notes:
+
+- **`pattern`** is matched case-sensitively; start it with `(?i)` for case-insensitive. Anchors (`^`) apply per line in `line` scope and to the joined window text in `window` scope.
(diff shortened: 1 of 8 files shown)


BTForIT:
Both review points addressed in 0334b2ac:

- **Duplicate rule names**: `compile()` now warns and drops later duplicates (same fail-open pattern as invalid regexes), so the `(session_id, rule_name)` cooldown key space stays unambiguous. Covered by a new unit test (`compile_drops_duplicate_names_keeps_first`).
- **`run_command` `.expect()`**: the command is now passed as a parameter from the `if let Some(command)` at the call site, so the presence invariant is structural rather than checked-by-caller.

`cargo fmt` / `clippy --all-targets` / pane tests under `--features serve` all clean.

Seluj78:
Hi @BTForIT !

Very interesting feature, though I think it would fit the plugin system perfectly!

How about you create a new plugin for this exact thing and then create a PR here for the missing capabilities/APIs you might need?

jerome-benoit:
> Very interesting feature, though I think it would fit the plugin system perfectly!

Agreed, it has no point to be in a 'controller' API designed to actually interact in real time with the different type of sessions. 

Rule based actions is an orchestration feature that should be designed and integrated properly within the orchestration extension of the plugins API.

njbrake:
Hi @BTForIT  thanks so much for the feature idea! As others mentioned, this is a perfect fit for a plugin. You can find docs here: https://www.agent-of-empires.com/docs/plugins/ 

I'll close this PR but feel free to open a new PR if we need to make any additions to the Plugin API to support your use case. Thanks 🙏

BTForIT:
Belated follow-up — @Seluj78 @jerome-benoit @njbrake you were right, and thanks for the pointer. We rebuilt this as a plugin on api_version 8 (stdio worker using `runtime.worker` + `session.read` + `notifications`) and it's been running in production here since mid-July. The plugin system covered nearly everything this PR was trying to do.

A couple of small capability gaps came up in the port (mainly around a plugin registering its own HTTP endpoints on the serve daemon); if there's interest we may open focused PRs for those. Thanks again for the review and the redirect — closing this was the right call.

Seluj78:
@BTForIT if you need new capabilities in the plugin API, yes feel free to open PRs!

Issue #2718, opened 2026-07-07
[followup #2654] Introduce Probe::Existence tri-state parallel to Probe::Running


jerome-benoit:
Follow-up from #2654 (fix(containers): promote does_container_exist to tri-state), which mirrored the `classify_inspect_failure` pattern (#2652 for #2596) onto the existence-probe surface but deferred the tri-state Probe wrapper.

## Scope

Introduce `Probe::Existence { Present, Absent, Unknown(DockerError) }` as a parallel to the running-state `Probe` at [`src/containers/mod.rs:89-108`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L89-L108), with a matching classifier `classify_existence_probe(Result<bool>) -> Probe::Existence` and a `DockerContainer::probe_existence(&self) -> Probe::Existence` constructor parallel to [`probe_running` at `src/containers/mod.rs:261`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L261).

## Why deferred (not landed in #2654)

Today there is only one fail-open consumer of `DockerContainer::exists()` (namely [`cleanup_sandbox_worktree` at `src/git/cleanup.rs:426`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L426), repaired in #2654 via an explicit `match` on `Result<bool>`). The three other callers already return `Result` and consume `Result<bool>` fluently via `?`:

- [`create_container` at `src/containers/runtime.rs:243`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L243)
- [`DockerContainer::exists` wrapper at `src/containers/mod.rs:139`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L139)
- [`ensure_container_ready` at `src/session/instance.rs:3142`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/session/instance.rs#L3142) (already tri-state-aware; see the comment at [lines 3128-3132](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/session/instance.rs#L3128-L3132))

[...]

PR #2719, opened 2026-07-07
fix(containers): promote does_container_exist to caller-driven tri-state (closes #2654)

### LINKED ISSUE #2654: DockerContainer::exists / does_container_exist collapses daemon-down to Ok(false) (same class as #2596)
Split out from the review of #2652 (fix for #2596). Companion to that PR's fix on `is_container_running`: the existence-probe surface has the same collapse pattern, just on a different method with different call-site semantics.

## The bug

[`src/containers/runtime.rs:114-132`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L114) — `does_container_exist` implementation:

**Docker/Podman arm:**
```rust
let output = self.base.command()
    .args(["container", "inspect", name])
    .output()?;
Ok(output.status.success())
```

**Apple arm:**
```rust
let output = self.base.command().args(["logs", name]).output()?;
Ok(output.status.success())
```

Both arms map every non-success exit (container absent, daemon down, permission denied, transient runtime failure) to `Ok(false)`. Same swallowing-existence-probe class as `is_container_running` before #2652 fixed it — a daemon-down stderr routes to "container does not exist".

## Blast radius (low but real)

`.exists()` / `does_container_exist` is currently called from:

- [`src/containers/runtime.rs:234`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L234) `create_container` — the `does_container_exist(name)?` pre-check is at [line 240](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L240). A daemon-down `Ok(false)` here causes `run_create` to attempt creation, which then fails with a proper `DaemonNotRunning` from `runtime_base.rs`. The error surfaces one step later, so no silent skip.
- [`src/git/cleanup.rs:424`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L424) `cleanup_sandbox_worktree` — `if !container.exists().unwrap_or(false) { return false; }`. Best-effort cleanup; a false `Ok(false)` on daemon-down means skip cleanup, which is acceptable for a best-effort path. This site is in the diff of #2652 but was intentionally left unchanged because the fix belongs upstream in `does_container_exist` (this issue's scope) rather than at the call site.

**No current caller uses `.exists()` as a fail-closed gate**, so this is not the same P2 severity as #2596. But it is the same shape and any future gate site added on top of `.exists()` would inherit the bug silently.

## Fix shape

Mirror the pattern #2652 landed for `is_container_running`:

- Add `RuntimeBase::classify_exists_failure(&self, stderr: &str) -> Result<bool>` (parallel to `classify_inspect_failure`): `is_not_found(stderr)` → `Ok(false)`, `is_daemon_down(stderr)` → `Err(DaemonNotRunning)`, permission-denied → `Err(PermissionDenied)`, else → `Err(InspectFailed)` (or a new `ExistenceCheckFailed` variant).
- Both arms of `does_container_exist` use it.
- Consider whether to add a `Probe::Existence { Present, Absent, Unknown(DockerError) }` tri-state parallel to `Probe` for running state.
- After the upstream fix, the `unwrap_or(false)` at `src/git/cleanup.rs:424` should be reviewed: for that best-effort path, propagating `Err(DaemonNotRunning)` via `?` may still be inappropriate (the function returns `bool`, not `Result`), so the call site may need a `match` with an explicit "spawn error / daemon-down → warn + skip" branch. Track that as part of this fix rather than a separate follow-up.

## Anchors

- Target fn: [`pub fn does_container_exist(&self, name: &str) -> Result<bool>`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L114)
- Callers: [`create_container:234`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L234) (call at [`line 240`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L240)); [`cleanup_sandbox_worktree` call at `src/git/cleanup.rs:424`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L424).
- Reference precedent: #2652 for the running-state surface.

## Scope note

Explicitly out of scope for #2652 (running-state surface only). Filed as separate follow-up because no current caller uses `.exists()` as a fail-closed gate, so the class-of-bug argument for pulling it into #2652 is weaker than for the running-state fix. The `cleanup.rs:424` caller-side handling is included in this issue's scope so the upstream fix and its downstream ripple land together.

jerome-benoit:
## Description

Closes #2654.

`does_container_exist` at [`src/containers/runtime.rs:114`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L114) had the same swallowing-existence-probe shape the running-state surface had before #2652 closed it there for [`is_container_running`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L134): every non-success stderr on both arms (Docker/Podman `container inspect`, Apple `logs`) collapsed to `Ok(false)`, so daemon-down / permission-denied / transient failures were indistinguishable from a genuinely absent container. This is the same class-of-bug as #2596 (root), just on a different function. Class-of-bug lineage: #2576 (removal) → #2596 (discard) → #2652 (running-state) → #2654 (existence, this PR).

**Touched sites.**

- New classifier `RuntimeBase::classify_exists_failure(&self, stderr: &str) -> Result<bool>` at [`src/containers/runtime_base.rs:276`](https://github.com/jerome-benoit/agent-of-empires/blob/fix/2654-classify-exists-failure/src/containers/runtime_base.rs#L276), parallel to [`classify_inspect_failure` at line 248](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L248). Delegates to `classify_inspect_failure` so the order-of-checks (`is_not_found` → `is_permission_denied` → `is_daemon_down` → `InspectFailed`) stays authoritative in one place; the permission-denied-before-daemon-down priority documented at [runtime_base.rs:261-266](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L261-L266) applies to both surfaces.
- `does_container_exist` refactored at [`src/containers/runtime.rs:114`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L114): both `Docker | Podman` and `AppleContainer` arms capture stderr on non-success and route through `classify_exists_failure`, mirroring the shape [`is_container_running` at `runtime.rs:134`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L134) uses. The per-runtime not_found / daemon_down / permission_denied markers are pinned to each runtime's stderr wording (`container inspect` for Docker/Podman, `logs` for Apple); comment added at each arm to lock the argv against silent breakage, same shape as the pinning comment at [runtime.rs:137-141](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L137-L141).
- `cleanup_sandbox_worktree` caller repaired at [`src/git/cleanup.rs:424`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L424). The fn returns `bool` (not `Result`), so the previous `unwrap_or(false)` collapsed `Err` back into the same skip bucket, re-introducing the class-of-bug. Repaired with an explicit `match`: `Ok(true)` proceeds, `Ok(false)` skips, `Err(e)` emits `tracing::warn!` and skips, mirroring the `Probe::Unknown(e)` warn shape already present at [cleanup.rs:449-455](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L449-L455) for the adjacent running-state probe. Fail-open semantics preserved with a log for the failure signal.

**Callers reviewed.** Four call sites reach `does_container_exist` directly or via the [`DockerContainer::exists` wrapper at `src/containers/mod.rs:138`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L138):

1. [`create_container` at `src/containers/runtime.rs:243`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L243), fail-closed via `?`. Behavior preserved with strictly-better error message: daemon-down surfaces as `DockerError::DaemonNotRunning` with its actionable Display, one hop earlier than the previous fall-through to `run_create` (which failed with the same variant one call later).
2. [`DockerContainer::exists` wrapper](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L138), signature `pub fn exists(&self) -> Result<bool>` unchanged.
3. [`cleanup_sandbox_worktree` at `src/git/cleanup.rs:424`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/cleanup.rs#L424): repaired above.
4. [`ensure_container_ready` at `src/session/instance.rs:3142`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/session/instance.rs#L3142): uses `container.exists()?`, fail-closed via `?`. **No source change required.** The pre-existing comment at [instance.rs:3128-3132](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/session/instance.rs#L3128-L3132) explicitly documents intent-to-propagate ("`?` correctly propagates a daemon-down transient to the caller as Err ... See #2596"); with this fix that intent is realized instead of silently degraded.

**Design.**

- Reuse `DockerError::InspectFailed(sanitize_stderr(stderr))` for the terminal else, not a new `ExistenceCheckFailed` variant. Classification rules and single-line `Display` contract are identical to `classify_inspect_failure`; operators read the raw stderr via the `error` field regardless of which argv produced it.
- `cleanup.rs:424` caller shape is an explicit `match` with a single `Err(e) => warn + return false` arm. Per-variant fan-out adds no signal because each [`DockerError`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/error.rs) Display is already single-line and actionable via `thiserror`.
- `Probe::Existence` tri-state suggested in the issue body is deferred to #2718. Only one fail-open caller today (`cleanup_sandbox_worktree`); the running-state `Probe` at [mod.rs:89-108](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L89-L108) is justified by multiple fail-open consumers, which the existence surface lacks. Follow-up #2718 lands the tri-state once a second consumer appears.

**Tests.** Four fixture-based tests added to `src/containers/runtime_base.rs` parallel to the [`classify_inspect_failure` test triples at lines 685-806](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L685-L806):

- `exists_not_found_stderr_collapses_to_ok_false` (Docker, Podman, Apple; reuses the `DOCKER_MISSING` / `PODMAN_MISSING` / `APPLE_MISSING` fixtures shared with the `classify_inspect_failure` suite).
- `exists_daemon_down_stderr_maps_to_daemon_not_running` (Docker, Podman socket + Podman desktop, Apple placeholder).
- `exists_permission_denied_stderr_maps_to_permission_denied` (Docker tight marker, Podman + Apple broad markers, respecting the pre-existing placeholder-fixture asymmetry documented at [runtime_base.rs:808-837](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L808-L837)).
- `exists_generic_transient_maps_to_inspect_failed`.

No runtime.rs-level tests added: the existing test style at [`runtime.rs:344-373`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime.rs#L344-L373) is entirely live-daemon-gated via `docker_if_available()` / `apple_container_if_available()` / `podman_if_available()`; there is no `Command` mock or fixture-injection harness. The classifier tests exercise the routing semantic (both arms funnel through `classify_exists_failure`) and the runtime.rs dispatchers are four visible lines per arm.

**Verification.**

- `cargo fmt --check`: clean.
- `cargo build --features serve`: clean.
- `cargo clippy --all-targets --features serve -- -D warnings`: clean.
- `cargo test --features serve`: full suite passes (all lib + all integration tests; the four new `exists_*` tests pass 4/4 by name).
- `cargo test --no-default-features --lib --bins`: 3718 passed, 0 failed. Matches CI's canonical default-features invocation at [`.github/workflows/tests.yml:93`](https://github.com/agent-of-empires/agent-of-empires/blob/main/.github/workflows/tests.yml#L93).

**Follow-up.** Scope kept to the minimal `does_container_exist` fix + caller repair; the remaining class-of-bug work is filed as six follow-up issues, three of which (#2721/#2722/#2723) bundle multiple related items each to keep the PR diff auditable:

- **#2718**: `Introduce Probe::Existence tri-state parallel to Probe::Running` (deferred; lands once a second fail-open consumer exists, see Design section for rationale).
- **#2720**: `Align pre-existing running-state warn wording at cleanup.rs:449` (pair with the new `:436` "container existence probe failed" anchor).
- **#2721**: `Harden RuntimeBase classifier plumbing on edge cases` (empty stderr, NotFound→NotInstalled, from_utf8_lossy doc; 3 gaps consolidated).
- **#2722**: `Rename classify_inspect_failure → classify_probe_failure and drop classify_exists_failure shim`. Round-2 review extended the scope ([see linked comment](https://github.com/agent-of-empires/agent-of-empires/issues/2722#issuecomment-4909386448)) to also cover: `DockerError::InspectFailed` Display rename (now surfaces on Apple `logs` path), and `run_create` inline classifier consolidation into `classify_create_failure`.
- **#2723**: `Hoist runtime_base fixture strings to shared consts and strengthen matches! payload assertions`. Round-2 review extended the scope ([see linked comment](https://github.com/agent-of-empires/agent-of-empires/issues/2723#issuecomment-4909386549)) to also cover: negative-ordering bleed test (daemon-down stderr must not match `not_found_markers`), and delegation-identity mirror-property test for `classify_exists_failure`.
- **#2730**: `Verify Apple container logs semantics on stopped-but-existing containers + capture fixture`. The `TODO` at [`runtime.rs:145`](https://github.com/jerome-benoit/agent-of-empires/blob/fix/2654-classify-exists-failure/src/containers/runtime.rs#L145) tracks this assumption at the call site.

**Pre-existing failure observed under bare `cargo test`**: `tests/acp_runner_orphan.rs::orphaned_runner_self_terminates_when_record_deleted` and `superseded_runner_exits_without_deleting_replacement_record` fail with `unrecognized subcommand '__acp-runner'`. Root cause: the `__acp-runner` subcommand is `#[cfg(feature = "serve")]`-gated at [`src/cli/definition.rs:210`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/cli/definition.rs#L210), but the integration test spawns `aoe __acp-runner` unconditionally. Reproduces on `upstream/main` HEAD with a clean tree; CI does not run bare `cargo test` (the workflow uses `cargo test --features serve` at line 77 and `cargo test --no-default-features --lib --bins` at line 93). Unrelated to the containers surface touched here; flagging so someone can pick it up as a separate small fix (`#[cfg_attr(not(feature = "serve"), ignore)]` on the two tests, or lifting the `#[cfg(feature = "serve")]` on the subcommand).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (opencode CLI, opus 4.7).

Design choices (tri-state introduction, error-variant choice, caller-repair shape) resolved before implementation; diff reviewed before commit; PR body reviewed before opening. Human-in-the-loop for the assignment courtesy comment on the issue and for shipping the PR in draft.

This PR makes container existence checks more reliable by ensuring runtime failures are no longer silently treated as “the container is missing.” It also updates sandbox worktree cleanup to explicitly handle the three possible outcomes of the existence probe: proceed, skip, or warn-and-skip.

Key changes:
- Improved `does_container_exist` for Docker/Podman and Apple runtimes so stderr-based failure classification is used instead of collapsing failures into “false.”
- Added a shared `RuntimeBase::classify_exists_failure` that reuses the existing inspect-failure classification rules, keeping behavior consistent across runtimes (including not-found, permission-denied, daemon-down, and generic transient errors).
- Updated `cleanup_sandbox_worktree` to handle the existence probe’s `Result<bool>` explicitly: `Ok(true)` continues, `Ok(false)` skips cleanup, and `Err(e)` logs a warning and skips.
- Added fixture-based tests covering not-found, daemon-down, permission-denied, and generic transient stderr cases on the existence surface.
- Included a small documentation/comment clarification around the existence-probe behavior vs how other checks behave.

Benefits:
- Fewer false negatives where runtime problems were incorrectly interpreted as “container missing.”
- More accurate error reporting when the container runtime is unavailable or access is denied.
- Safer, more predictable cleanup that doesn’t fail unexpectedly when existence detection itself can’t be trusted.

```technical
- `does_container_exist`: switched to `RuntimeBase::classify_exists_failure(stderr)` and now returns `Ok(true)` on success while mapping classified failures to `Ok(false)` (not found) or `Err(...)` (daemon not running / permission denied / transient inspect-style failures).
- `cleanup_sandbox_worktree`: replaced implicit `unwrap_or(false)` with explicit `match` on `Result<bool>` and warn-on-`Err` behavior.
- Tests: added fixtures asserting the tri-state mapping (present/missing/error) for existence-surface stderr across runtimes, plus a comment/doc tightening for Apple-specific behavior provenance.
```


src/git/cleanup.rs
```diff
@@ -413,16 +413,30 @@ pub fn is_permission_error(error: &str) -> bool {
...
-/// (idempotent if the container is already running), preserving best-effort
-/// behavior while surfacing the failure in logs.
+/// Both container probes (existence and running-state) are fail-open with
+/// a `warn!`: on a transient runtime failure (daemon down, permission
+/// denied, or any other non-success stderr classified by
+/// [`crate::containers::runtime_base::RuntimeBase::classify_exists_failure`]
+/// or `classify_inspect_failure`) the cleanup is skipped and the failure
+/// surfaces in logs. Collapsing the existence probe to `unwrap_or(false)`
+/// re-introduces the swallowing-existence-probe class of bug (#2596 /
+/// #2652 / #2654).
```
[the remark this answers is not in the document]
jerome-benoit: Applied in `3668071b` with adjustments to match house style and preserve information density. Deliberate divergences from the exact suggestion:

- **Kept bare `#NNNN`** for the class-of-bug lineage. Repo-wide grep: 1386 bare `#NNNN` hits vs 0 backticked `` `#NNNN` `` (see e.g. [`src/containers/mod.rs:86`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/mod.rs#L86), [`src/containers/runtime_base.rs:246`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L246)).
- **Dropped the intra-doc link entirely** rather than shortening. `cleanup.rs` has no `use RuntimeBase` and adding one solely for a doc link would be a dead import (unused-import lint risk). Classifier names are kept in backticks as searchable prose (`` `RuntimeBase::classify_exists_failure` ``, `` `classify_inspect_failure` ``).
- **Preserved the failure-taxonomy enumeration** ("daemon down, permission denied, or any other non-success stderr classified by...") on the existence-probe sentence. The classifier catalog is load-bearing for readers debugging cleanup skips.
- **Named `container.start()` explicitly** on the running-state sentence with an idempotency parenthetical, so the docstring grep-anchors to the actual code path at `cleanup.rs:455`.

The core accuracy fix (existence probe vs running-state probe distinction) is applied per your finding. The round-2 polish (`95ba9ff1`) had conflated the two probes' error handling into a single sentence — thanks for the catch.

src/containers/runtime.rs
```diff
@@ -114,19 +114,43 @@ impl ContainerRuntime {
...
+                }
+                Ok(true)
             }
             RuntimeKind::AppleContainer => {
-                // Apple Container's `inspect` returns success(0) for non-existent
-                // containers, so we use `logs` which properly fails for missing
-                // containers.
+                // Apple Container's `inspect` returns success(0) for
+                // non-existent containers, so we use `logs` which properly
+                // fails for missing containers. APPLE_MISSING pins the
```
Seluj78: Required: comment overclaims fixture provenance. APPLE_MISSING is documented at runtime_base.rs:658 as captured from `<runtime> rm/delete <missing>`, not `logs`. Saying it "pins the absent-container stderr from logs" overstates what has actually been fixture-verified for this argv, undercutting the pinning-comment safety mechanism this PR relies on elsewhere. Reword to not claim logs-specific provenance; the underlying uncertainty is already tracked by the TODO two lines below and issue #2730.

Seluj78 (changes requested):

## AoE Review

Reviewed: PR #2719 (base: main@f54f543c). Files changed: 3 (src/containers/runtime.rs, src/containers/runtime_base.rs, src/git/cleanup.rs).
Unresolved comments: none (1 CodeRabbit inline thread found, already resolved and outdated, fix landed in current diff).

### Required

**src/containers/runtime.rs:138** - comment overclaims fixture provenance.
Quoted line: `// fails for missing containers. APPLE_MISSING pins the`
Reason: comment says APPLE_MISSING "pins the absent-container stderr from logs," but the fixture's own sourcing comment at runtime_base.rs:658 states it was captured from "<runtime> rm/delete <missing>", not logs, not inspect. The classifier match still passes today (substring marker match is command agnostic), but the claim undercuts the safety purpose these pinning comments exist for: a future dev trusting "pinned to logs" could believe the argv is fixture-verified for logs when it has never been captured against real `container logs <missing>` output. This gap is the same one already tracked by the inline TODO two lines below and issue #2730, so the fix is to stop the new comment from overstating what's actually verified.
Rule: correctness. Comment accuracy directly undermines the argv-pinning mechanism this PR itself introduces (runtime.rs:56-62 documents the same pattern for Docker/Podman and is accurate there because DOCKER_MISSING/PODMAN_MISSING genuinely originate from the runtimes' own not-found wording style).

### Followup

None beyond what the PR already tracks. The one substantive open question independently found here, Apple `container logs` on a stopped-but-existing container falling through to a generic Err instead of Ok(true), which would surface as a hard failure at src/session/instance.rs:3142 (`container.exists()?`) instead of the previous silent-recreate, is already disclosed in the PR body ("Callers reviewed," bullet 4) and tracked via the inline TODO at runtime.rs:141-143 and issue #2730. No new issue needed.

### Verdict

Implementation: Approve
Project Fit: Accept

Tight, well-scoped fix that closes a real bug class (#2596 lineage) with minimal diff, matching the project's existing pinning-comment and classifier-delegation conventions. Mechanical gates (em dash, dead code, tracing targets, PR title/branch convention) all clean. The one Required finding is a one-line comment-wording fix, not a design or correctness problem in the shipped logic. CI was still mostly pending at review time; nothing failing.

src/containers/runtime.rs
jerome-benoit: Applied in `f15bb73b`. Per-clause fix summary:

- **Line 138** (the sentence you quoted): `APPLE_MISSING pins the absent-container stderr from `logs`` → `APPLE_MISSING captures Apple's absent-container stderr (from `rm/delete`)`. Cites the actual provenance per [`runtime_base.rs:657`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/containers/runtime_base.rs#L657) (`// Real stderr captured from <runtime> rm/delete <missing> on 2026-07-01.`).
- **Line 140-141** (same class of overclaim, fixed for internal consistency): `... markers on RuntimeBase::APPLE_CONTAINER are calibrated to `logs` output` → `... key off Apple's not-found style and are expected to match `logs` stderr by substring, though `logs` stderr has not been captured as a fixture`. Separates the substring-match rationale from the missing `logs`-specific fixture; leaving only line 138 fixed would have produced an internally inconsistent comment (accurate about provenance on one sentence, still overclaiming argv-calibration on the next).
- **Preserved** the substring-match safety rationale (`Switching argv here (or tightening the markers) needs new fixtures`), the `#2596` cross-ref to the Docker/Podman pinning comment above, and the inline `TODO` + `#2730` tracking the residual `logs`-on-stopped-but-existing uncertainty.
- **Docker/Podman pinning at `runtime.rs:117-123`** untouched — you flagged it as accurate (`DOCKER_MISSING`/`PODMAN_MISSING` originate from those runtimes' stable not-found wording style; the argv-invariance is real there).

Ready for re-review.

Seluj78 (approved):

Issue #2600, opened 2026-07-01
Flaky file_watch tests: global HOME env race under parallel cargo test


Seluj78:
## Summary

The `src/tui/home/file_watch_tests.rs` tests flake intermittently in CI (`Cargo Test (linux, ...)` shards), each run failing a different test with a "watcher handle not installed" style precondition, e.g.:

```
tui::home::file_watch_tests::rewire_disk_subscriptions_drops_removed_profile_entry ... FAILED
panicked at src/tui/home/file_watch_tests.rs:133: precondition: hv-keep installed
tui::home::file_watch_tests::rewire_no_op_preserves_latched_disk_watcher_init_failure ... FAILED
tui::home::file_watch_tests::rewire_disk_subscriptions_install_loop_skips_missing_profile_dir ... FAILED
```

They pass on rerun. Discovered while running CI for #2598 (unrelated feature PR); not caused by that change.

## Reproduction

- Full lib suite in a loop reproduces it: `for i in $(seq 1 5); do cargo test --lib; done` failed 1/5 with a file_watch test (a different one each time).
- The file_watch module in isolation does NOT flake: `for i in $(seq 1 15); do cargo test --lib tui::home::file_watch_tests; done` was 0/15.
- Reproduced on macOS, so it is not an inotify instance/watch limit (that is Linux-specific).

## Root cause

The failing tests set the process-global `HOME` / `XDG_CONFIG_HOME` via `isolate_home()`, then synchronously call `get_profile_dir(...)` and `rewire_disk_subscriptions(...)`. They are `#[serial]` (default lock), so they interlock with each other and with the other default-`#[serial]` tests. But a **non-`#[serial]` test that also mutates `HOME`** runs concurrently on another worker thread and flips the global env mid-test, so profile-dir resolution / watcher install lands on the wrong (or missing) directory and the expected handle is absent.

## Fix direction

Make global-env isolation race-free across the whole test binary. Options:

1. Ensure every test that mutates `HOME`/`XDG_CONFIG_HOME` shares one serial lock (add `#[serial]` to the non-serial HOME mutators). Requires enumerating them across `hooks`, `acp/supervisor`, `session/*`, `server/api/sessions`, `tui/*`.
2. Move test isolation off process-global env: thread an explicit app-dir/home path through `get_app_dir()` in tests instead of `set_var`.
3. Adopt cargo-nextest with a small retry budget as a stopgap (CI currently runs plain `cargo test`, no retries).

[...]

Issue #10, opened 2026-01-12
Refresh from the storage every ~10 seconds


njbrake:
If two aoe bins are running and one of them renames a session, the other running aoe needs to be able to pick up the change.

PR #2729, opened 2026-07-07
fix(session): eliminate spurious restamp seed and IdleField ambiguity (#2697 review)

### LINKED ISSUE #2690: TUI: session activity-time column resets to <1m on reload — status-change restamp clobbers persisted idle_entered_at/last_accessed_at
## Summary

In the TUI session list, the right-edge activity-time column resets to `<1m` for every session whenever `aoe` reloads / the daemon restarts, even for sessions that have been idle for hours. The displayed age no longer reflects when the agent last did anything — it reflects how long ago the daemon restarted.

Observed with plain tmux Claude sessions (`view: null`, not structured). Reproduces at `a812ed4d` (`v1.12.0-38`) and on `v1.12.0`.

## Repro

1. Have one or more idle sessions that last produced output a while ago (column shows e.g. `2h`).
2. Quit and relaunch the TUI (or otherwise restart the daemon).
3. The activity column now shows `<1m` for those sessions, despite no new agent activity.

## Diagnosis

The activity column source (`src/tui/home/render.rs:1479-1484`):
- Idle rows use `idle_entered_at` (documented as the wall-clock time of the most recent transition into `Idle`).
- Non-idle rows use `last_accessed_at`.

Both fields are persisted in `sessions.json` and load back correctly on startup. They get clobbered afterwards:

`src/session/instance.rs:4036-4048`
```rust
pub fn update_status_with_metadata(&mut self, metadata: Option<&tmux::PaneMetadata>) {
    let prev_status = self.status;
    self.update_status_with_metadata_inner(metadata);
    if self.status != prev_status {
        let now = Utc::now();
        self.last_accessed_at = Some(now);
        self.idle_entered_at = if self.status == Status::Idle {
            Some(now)
        } else {
            None
        };
    }
}
```

On a daemon restart the status poller re-detects each session's status from a transient starting point rather than the persisted value, so `self.status != prev_status` fires on the first probe (e.g. a `Starting`/unknown → `Idle` transition that is really just the daemon re-establishing known state). That trips the restamp, overwriting the persisted `idle_entered_at` / `last_accessed_at` with `now`.

Evidence from a live `sessions.json` entry after a reload — `idle_entered_at` is newer than `last_accessed_at`, i.e. the idle stamp was rewritten on reconnect:
```json
{
  "title": "zellij-help",
  "status": "idle",
  "created_at": "2026-06-29T05:51:36Z",
  "last_accessed_at": "2026-07-06T12:38:43Z",
  "idle_entered_at": "2026-07-06T13:21:51Z"
}
```

So the reset is caused by the status-change restamp firing on the restart-time re-detection, not by a missing or non-persisted source.

## Related (adjacent, not duplicates)

- #1368 — same column, request to make its content configurable.
- #2250 — `Status::Stopped` de-overloading; adjacent status-semantics work.
- #1689 / #1690 — idle-reap consumes the same activity signal.

jerome-benoit:
## Description

Follow-up to the post-merge review of #2697 (which fixed #2690: TUI activity age reset to `<1m` on TUI/daemon reload). The review produced 25 findings across four buckets (Required, Suggested, Nit, Follow-up). This PR addresses every finding in-place: either by code change, or by an in-file docstring / design note, or by an explicit skip with reasoning. No separate tracking issues, no follow-up PR.

Six commits, each staying on the same PR:
- `3f9f3e68` : initial correctness fixes (Req #1-#4 + Sug #3-#10 + Nit #4 + FU #3 + locking tests).
- `342799f6` : further review followups still in-place (Sug #2 + Sug #6 + FU #2 + FU #4). Type prefix reads `docs(session):` but the commit also flips a runtime guard (Sug #6 `>` → `>=`); observable value unchanged, but the prefix is technically a `refactor(session):`.
- `5bcd033b` : convention harmonization and precision refinements (renames `IdleField` → `IdleIntent`, `ProfileWriteBundle` → `PassiveTransitionWrites`; `PartialEq` derive; tuple-match arm order flipped to Set/Clear/Keep on the producer; authoritative-asymmetry comment on `poll_statuses_once`; `PassiveStatusPatch` glossary moved to `///` rustdoc; test renamed to `_boundary_equal_is_a_noop`; extra locking test `test_status_update_default_is_no_op`; several docstring / comment tightenings).
- `82f50e80` : enum-order + prose coherence (`IdleIntent` variants reordered `Set / Clear / Keep` to match consumer sites; `poll_statuses_once` emit-site comment tightened; `merge_passive_status_patch` log wording aligned with the `_is_a_noop` test name; missing `that` in `PassiveStatusPatch.last_accessed_at` field doc; unfinished "Grep-friendly." sentence removed).
- `63ea23f6` : fifth-pass follow-ups (concurrency/testing/cross-platform/observability angles). TUI passive-write path gains `tracing::warn!` on `storage.update` failure for parity with the daemon; `merge_passive_status_patch` debug log renders `patch_ts` unwrapped; wall-clock / NTP-rewind note on the `>=` guard; construction-ordering note on `live_status_baseline`; "within one tick" softened to "within one poll interval of the slower cadence"; `IdleIntent::Clear` docstring notes the sandbox-dead synthesized-transition case; `PassiveTransitionWrites.patches` key redundancy documented as intentional; assert-message style uniformized on `test_status_update_default_is_no_op`; two new locking tests (producer-asymmetry, consumer-arm-mapping).
- `bfa5aefe` : sixth-pass follow-ups (API surface / perf-review / feature-flags / rustdoc angles). `pub enum IdleIntent` and `pub struct StatusUpdate` tightened to `pub(crate)` (parent `mod status_poller;` is private, every consumer is a crate-internal sibling). `PassiveStatusPatch` derive gains `Eq`; `StatusUpdate` derive gains `PartialEq`. Rustdoc H1 -> H2 demotion on `# Persistence divergence`, `# Poller vocabulary`, `# Durability contract` (H1 reserved for rustdoc `# Panics`/`# Examples`/`# Errors`/`# Safety`). Intra-doc links pointing at `#[cfg(test)]`-only items dropped their `[...]` brackets (silently degraded to literal text under `cargo doc`) with an inline rationale so a future reviewer does not restore them. Method-site note on `merge_passive_status_patch` reiterates the silent drop under `>=`.

### Required correctness fixes

- **Req #1: `Instance::new` seed causes fabricated `last_accessed_at`.** [`src/session/instance.rs`](src/session/instance.rs) pre-fix seeded `live_status_baseline: Some(Status::Idle)`. Post-construction status writers (`finalize_launch`, TUI stub, several `server/api/sessions.rs` sites) advanced `status` to `Starting` without touching baseline. On the TUI polling clone, the wrapper's next call read `baseline=Some(Idle) != status=Starting` and stamped `last_accessed_at` on a session no user had touched. The daemon path dodged this because `status_poll_loop` overrides baseline from `prev_for_poll` every tick. Seed `None` so both paths seed on the first poll rather than restamping.
- **Req #2: asymmetric propagation in `apply_status_update`.** [`src/tui/home/mod.rs`](src/tui/home/mod.rs) pre-fix propagated `idle_entered_at` unconditionally and `live_status_baseline` conditionally. [`attached_status_hooks::snapshot`](src/tui/attached_status_hooks.rs) from a watcher clone that never polled would emit `idle_entered_at: None`, and the consumer treated that as "clear". Introduce `IdleIntent { Set(DateTime<Utc>), Clear, Keep }` in [`src/tui/status_poller.rs`](src/tui/status_poller.rs) (crate `\w+Intent` convention: `StatusIntent`, `ResumeIntent`, `QuitIntent`). Consumer matches on the variant; producers emit the right variant based on their state. [`attached_status_hooks::apply_updates`](src/tui/attached_status_hooks.rs) also now propagates baseline, so the snapshot's `(baseline, idle_entered_at)` tuple can distinguish "watcher observed a Clear" from "watcher never observed".
- **Req #3: broken merged comment splice.** [`src/tui/home/tests.rs`](src/tui/home/tests.rs) pre-fix had two half-comments spliced together, with the first sentence trailing off at "instead of" and the CodeRabbit follow-up prose misattached to the wrong test. Restored the two comments to their intended anchors.
- **Req #4: self-referential regression comment.** [`src/server/mod.rs`](src/server/mod.rs) `PassiveTransitionDecision` docstring said `Regression: #2697` (this PR's own number). Anchored the docstring to the locking test `decide_passive_transition_skips_patch_for_structured_session`.
- **Req #5: em-dash in merged PR #2697 body.** Not fixable from this PR: the body of @Seluj78's PR is not part of the codebase and is not editable by non-authors. Left as observed-but-out-of-scope.

### Suggested refinements

- **Sug #1** (`live_status_baseline` coupling with `Instance`): considered, deferred. The field currently sits on `Instance` with `#[serde(skip)]` and is only ever meaningful from a poller's viewpoint. Moving it off would touch `Instance::update_status_with_metadata`'s signature, `StatusPollState` in [`src/tui/status_poller.rs`](src/tui/status_poller.rs), the daemon's `status_poll_loop`, and every test that constructs a fresh `Instance` and asserts baseline behavior. That is an orthogonal architectural refactor whose blast radius outweighs the shape benefit inside a review-follow-up PR.
- **Sug #2** (durable-write path for structured/ACP `idle_entered_at`): documented as a contract on [`apply_acp_overlay_inplace`](src/server/mod.rs). Structured rows accept a soft reset of `status`/`last_accessed_at`/`idle_entered_at` on daemon restart, by contract. ACP event handlers are responsible for any post-restart re-emission; the passive-status writer at `status_poll_loop` deliberately skips them.
- **Sug #3** (HashMap patch lookup): implemented via `PassiveTransitionWrites` in [`src/server/mod.rs`](src/server/mod.rs) (renamed from `ProfileWriteBundle`; matches sibling helper types `PriorById`, `StructuredRowRepair`, `PassiveTransitionDecision`). Patches keyed by instance id; `.get(&inst.id)` inside the persistence closure is O(1).
- **Sug #4** (union-key collapse): implemented. Single `HashMap<Profile, PassiveTransitionWrites>` replaces the two parallel maps + a `HashSet<String>` union, and destructuring the bundle inline drops the `.remove(&profile).unwrap_or_default()` dance.
- **Sug #5** (`is_some_and` collapse in `update_status_with_metadata`): implemented, matches [`merge_passive_status_patch`](src/session/instance.rs)'s guard style.
- **Sug #6** (`last_accessed_at` guard semantic): implemented. Flipped strict `>` to `>=` so equal timestamps are dropped rather than re-applied. The observable `last_accessed_at` is unchanged either way (disk == incoming). Locking test named `_boundary_equal_is_a_noop`; the assertion cannot distinguish drop-from-write, and the name matches what the test actually locks.
- **Sug #7** (`See #2690 CodeRabbit follow-up` pointer): the flagged pointer is replaced with a test-name anchor pointing at `apply_status_update_propagates_live_status_baseline_from_poller`.
- **Sug #8** (docstring overlap between field and method): implemented. Field doc on [`live_status_baseline`](src/session/instance.rs) carries the invariant; method doc on `update_status_with_metadata` carries the guard shape and delegates via intra-doc link.
- **Sug #9** (triple restatement of `None`-preservation): implemented. Field doc keeps the rationale; `PassiveStatusPatch::from_instance` doc delegates via `[Self::last_accessed_at]`.
- **Sug #10** (`#2697` self-reference in `merge_passive_status_patch` doc): implemented. Dropped, keeping only `#2690`.

### Nit

- **Nit #1** (`PassiveTransitionDecision` single-caller struct), **Nit #2** (`PassiveStatusPatch` first-`*Patch` naming), **Nit #3** (`live_status_baseline` diverges from `last_*` prefix): informational. The review flagged each as an intentional design decision (documented as such at each site), not something to change.
- **Nit #4** (intra-doc links in field docs): implemented. `[Self::update_status_with_metadata]`, `[Instance::new]`, `[Self::last_accessed_at]` used in the affected field docstrings.
- **Nit #5** (`attached_status_hooks::snapshot` inline backticks vs intra-doc links): not applied. The crate uses inline backticks for sibling-module references throughout; converting these two sites in isolation would be inconsistent with the rest of `src/tui/`.

### Follow-up

- **FU #1** (consolidate `PolledObservation`): considered, deferred. `StatusUpdate` grew a seventh observation field with this PR, and the review notes a `PolledObservation` type + `Instance::merge_polled_observation` sink would let future observation fields be added without a matching apply-path guard. This is a coordinated refactor of `StatusUpdate`, its producers, its consumers, and every test that constructs it. Same reasoning as Sug #1: orthogonal architectural work, out of scope for a review-follow-up.
- **FU #2** (persistence divergence between daemon and TUI): implemented as a design note on [`PassiveTransitionWrites`](src/server/mod.rs). Precise field semantics: `last_accessed_at` is monotone (guarded by `>=`); `status`/`idle_entered_at` are unconditional writes (last-writer-wins); safety derives from single-authority + fast reconvergence, not from mathematical order-independence. The rule for future field additions is stated explicitly. Worst-case inversion window between the two writers is bounded by `max(status_poll_loop tick, TUI poll interval)`; on the next tick either writer converges the disk state to the current live source. The invariant is enforced by reviewer diligence; a compile-time seal or reflection-based lock test would be more durable but is out of scope here.
- **FU #3** (`#[derive(Default)]` on `StatusUpdate`): implemented (`Status::default() == Idle`, `IdleIntent::default() == Keep`), so future test fixtures can use `..Default::default()`. Locking test `test_status_update_default_is_no_op` in `src/tui/status_poller.rs::tests` asserts every field of `StatusUpdate::default()` matches its expected default, catching future field additions with non-trivial defaults at review time.
- **FU #4** (poller-vocabulary glossary): implemented as a `///` rustdoc block on `PassiveStatusPatch` in [`src/session/instance.rs`](src/session/instance.rs) (moved from a freestanding `//` block so `cargo doc` renders it). Defines `passive status`, `passive status patch`, `live status baseline`, `detected status`, `poller-authoritative status` in one place.
- **FU #5** (sweep `See #2690` citations to test-name anchors): partially applied. The one flagged as a stale pointer (the `CodeRabbit follow-up` at the top of [`src/tui/home/mod.rs`](src/tui/home/mod.rs)) is now a test-name anchor. Other `See #2690` occurrences are legitimate context references (they explain the bug that motivated the code), and are kept as issue references.

### Convention harmonization + precision refinements

Applied across commits `5bcd033b` and `82f50e80`:

- **Renames** (workspace-wide, zero behavior change): `IdleField` → `IdleIntent` (crate `\w+Intent` convention); `ProfileWriteBundle` → `PassiveTransitionWrites` (domain-noun sibling of `PriorById`, `StructuredRowRepair`, `PassiveTransitionDecision`).
- **`IdleIntent` variant declaration order** flipped to `Set / Clear / Keep`, matching the three consumer match-arm sites for grep-consistency. `#[default] Keep` still applies (Default derive is variant-attributed, not positional).
- **`PassiveTransitionWrites` design-note precision**: previous prose overstated "order-independent"; now precise on monotone (`last_accessed_at`) vs last-writer-wins (`status`/`idle_entered_at`) vs the single-authority + fast-reconvergence safety argument.
- **`snapshot()` tuple-match arm order** aligned to `Set / Clear / Keep`, cross-referencing the enum declaration and both consumer sites.
- **`poll_statuses_once` authoritative-asymmetry comment**: this producer is authoritative on `idle_entered_at` for both the sandbox-dead branch and the tmux branch; neither ever emits `IdleIntent::Keep`. `attached_status_hooks::snapshot` remains the sole `Keep`-emitter. A future consolidation adding `Keep` here would erase the baseline seed and reintroduce #2690.
- **`PassiveStatusPatch`** gains `PartialEq` derive.
- **Test rename**: `_boundary_equal_is_dropped` → `_boundary_equal_is_a_noop` (name matches what the assertion actually locks).
- **`REFRESH_INTERVAL`-window note** on the `snapshot()` tuple-match `Keep` case.
- **Guard-only comment** on `test_first_poll_after_status_write_does_not_fabricate_last_accessed_at`.
- **Bare `#2697:`** test-comment prefix → `Locks the CI regression from #2697:` (disambiguation).
- **`PassiveTransitionWrites` doc**: "on that shape" → "at that cardinality".
- **`merge_passive_status_patch` log wording** aligned with the locking-test name family (`_is_a_noop`).
- **`PassiveStatusPatch.last_accessed_at`** field doc: added missing `that` in `contract that idle-reap and the freshness sort rely on`.
- **`snapshot()` tuple-match comment**: unfinished "Grep-friendly." sentence replaced with an explicit cross-reference to the enum declaration order.

### Fifth-pass follow-ups (in-place)

Applied in commit `63ea23f6`:

- **TUI passive-write path silent on disk failure**. [`persist_passive_status_transition`](src/tui/home/mod.rs) now emits `tracing::warn!` at `target: "session.store"` on the `storage.update` `Err` arm, matching the daemon's sibling [`api::persist_session_update`](src/server/api/sessions.rs) error log. The docstring already documents no-rollback (best-effort persistence, the poller is the sole authority on live status regardless of whether disk persistence succeeds); silence would obscure a persistent flock timeout or EIO loop.
- **`patch_ts` renders as `Option` in `merge_passive_status_patch` debug log**. Changed to `patch_ts = %incoming` at [`src/session/instance.rs`](src/session/instance.rs); `incoming` is guaranteed `Some` by the `let-else` unwrap directly above the log call, so `%` displays the timestamp unwrapped.
- **Doc sharpening pass** across four sites:
  - Wall-clock / NTP-rewind semantics on the [`merge_passive_status_patch`](src/session/instance.rs) `>=` guard: `chrono::Utc::now()` delegates to `SystemTime::now()` (wall clock, not monotonic), so a genuinely newer live observation stamped after an NTP rewind can compare less than a value stamped before it and be silently dropped. Best-effort monotone; the next poll tick converges regardless.
  - Construction-ordering rationale on the [`Instance.live_status_baseline`](src/session/instance.rs) field doc: `#[serde(skip)]` + `Instance::new`-time `None` seed rely on the instance entering shared state only after `Instance::new` returns. Safety here is by construction-ordering, not by synchronization.
  - "within one tick" wording softened on the [`PassiveTransitionWrites`](src/server/mod.rs) design note: convergence is "within one poll interval of the slower cadence" (daemon 2s, TUI ~500ms), reflecting that a worker transitioning mid-cadence can produce divergent patches which reconcile on the next tick, not "within" a single tick.
  - `IdleIntent::Clear` docstring in [`src/tui/status_poller.rs`](src/tui/status_poller.rs) notes that the sandbox-dead branch of `poll_statuses_once` emits `Clear` as a synthesized transition (container health flipped false), so `Clear` is not exclusively a user-observed transition-out-of-Idle signal.
- **`PassiveTransitionWrites.patches` HashMap key + `patch.id` redundancy**. Documented as intentional at [`src/server/mod.rs`](src/server/mod.rs): [`PassiveStatusPatch::from_instance`](src/session/instance.rs) keeps `patch.id == inst.id`, so the map key and the value's `id` field stay in sync by construction. The redundancy is used by the flush-site closure's `.get(&inst.id)` for O(1) lookup.
- **`test_status_update_default_is_no_op` mixed assert-message style**. All seven fields (`id`, `status`, `last_error`, `idle_entered_at`, `last_accessed_at`, `pane_dead`, `live_status_baseline`) now carry descriptive assertion messages, matching the pattern used on the three fields that already had them (was 3 of 7 with messages).

Two new locking tests:

- [`poll_statuses_once_never_emits_idle_intent_keep`](src/tui/status_poller.rs): asserts the producer-asymmetry that only [`attached_status_hooks::snapshot`](src/tui/attached_status_hooks.rs) emits `IdleIntent::Keep`, and `poll_statuses_once` never does. Structural today (both emit sites hardcode `Set`/`Clear`); the lock guards against a future refactor that would relax the projection into a match arm capable of returning `Keep`, which would erase the baseline seed and silently reintroduce #2690.
- [`apply_updates_maps_idle_intent_set_and_clear_arms`](src/tui/attached_status_hooks.rs): locks the `Set`/`Clear` arm mapping in [`apply_updates`](src/tui/attached_status_hooks.rs). Complements the equivalent [`apply_status_update_clears_idle_entered_at_on_idle_to_running`](src/tui/home/tests.rs) that covers the sibling arm mapping in `src/tui/home/mod.rs`. Without this test, a `Set` ↔ `Clear` swap in `apply_updates` would compile cleanly and pass the existing `apply_updates_runs_status_hook_for_transition` test (which only exercises the `Keep` arm).

Deferred, in-body-only follow-ups (no new tracking issues):

- **Equal-timestamp guard observability**: [`test_merge_passive_status_patch_last_accessed_at_boundary_equal_is_a_noop`](src/session/instance.rs) cannot distinguish "guard fired, wrote nothing" from "guard skipped, wrote the same value"; both produce the same observable `last_accessed_at` under `disk == incoming`. A `tracing_test` subscriber capture asserting the `dropped … as a no-op` debug event, or a `bool`-returning guard, would close the gap. Same observability class as the `AppDirGuard` locking-test gap flagged on PR #2717.
- **Persist-error swallow + `mark_unread` ordering** in the daemon's [`status_poll_loop`](src/server/mod.rs): pre-existing, not introduced by this diff or #2697. `inst.mark_unread()` is applied to the local `instances` vec before `persist_session_update` is awaited; on persist failure the mark survives in memory until the next `reload_state_instances_from_disk` overwrites it, and the next tick's `prev == inst.status` short-circuit prevents retry. Out of #2697 scope; either defer `mark_unread` until after persist returns `Ok`, or retry on failure.
- **`PassiveTransitionWrites` per-profile bucketing**: the `HashMap<Profile, PassiveTransitionWrites>` at [`src/server/mod.rs`](src/server/mod.rs) is populated inside `status_poll_loop` and never re-tested standalone. A mutation routing a session's patch to the wrong bucket would need a multi-profile integration harness (`serve_dynamic_profile_rewire` covers adjacent state).
- **Panic-path locking + cross-writer interleave harness**: no test panics inside `apply_status_update` or `poll_statuses_once` to assert state isn't corrupted; no test races the daemon's batched write against the TUI's per-transition write on the same profile+session. Both need integration fixtures; out of #2697 scope.
- **`FileWatchService` per-write firing + `prev_for_poll` unbounded growth**: pre-existing. An N-profile tick yields N TUI notifications and N reloads; `prev_for_poll` on a long-uptime daemon retains every ever-observed instance id (never GC'd on delete). Debouncing and GC live outside this PR.
- **Structured-session `idle_entered_at` after daemon restart**: documented as a contract on [`apply_acp_overlay_inplace`](src/server/mod.rs) (Sug #2). The row's `idle_entered_at` resets to the last durable value after a daemon restart, until an explicit user action or ACP event re-emits. Worth a release-notes flag.
- **Optional per-tick batched-write tracing**: a single `tracing::debug!(target: "session.store", profile = %profile, patches = patches.len(), unread = unread_ids.len(), "flushed passive-status batch")` at the end of the persist loop in [`src/server/mod.rs`](src/server/mod.rs) would let triage correlate per-tick disk churn to one line. Parallel to FU #1's deferred `PolledObservation` unification.
- **Update-wrapper panic-safety**: `Instance::update_status_with_metadata` at [`src/session/instance.rs`](src/session/instance.rs) is currently unreachable-panic (`_inner` has no panic paths), but if a future edit adds one, the wrapper would leave the Instance with a mutated `status` and a stale `live_status_baseline`. Contained today: the daemon's `spawn_blocking` catches the panic and returns `Err`, and the TUI polls on clones. Flag only.

### Sixth-pass follow-ups (in-place)

Applied in commit `bfa5aefe`:

- **Visibility tightening on crate-internal types**. [`IdleIntent`](src/tui/status_poller.rs) was `pub` inside the private `mod status_poller;` at [`src/tui/mod.rs`](src/tui/mod.rs); every consumer is a sibling ([`attached_status_hooks`](src/tui/attached_status_hooks.rs), [`home/mod.rs`](src/tui/home/mod.rs), [`home/tests.rs`](src/tui/home/tests.rs), [`app.rs`](src/tui/app.rs)) that reaches in via `super::status_poller::...`. Tightened both `IdleIntent` and [`StatusUpdate`](src/tui/status_poller.rs) to `pub(crate)` so declared visibility matches actual reachability, and future callers do not mistake these for a broader contract.
- **Derive-set completeness on new types**. [`PassiveStatusPatch`](src/session/instance.rs) derived `Debug, Clone, PartialEq` but not `Eq`; all its fields (`String`, `Status`, `Option<DateTime<Utc>>`) implement `Eq`, and sibling [`IdleIntent`](src/tui/status_poller.rs) derives `Eq`. Added `Eq`. Also added `PartialEq` to [`StatusUpdate`](src/tui/status_poller.rs) so a future `assert_eq!(update, expected)` can compare whole values instead of field-by-field (today's tests compare `update.idle_entered_at` in isolation).
- **Rustdoc heading levels**. `# Persistence divergence between daemon and TUI` in [`src/server/mod.rs`](src/server/mod.rs), `# Poller vocabulary` in [`src/session/instance.rs`](src/session/instance.rs), and `# Durability contract` in [`src/server/mod.rs`](src/server/mod.rs) rendered as `<h1>` inside item-level doc blocks, competing visually with the item title. Rustdoc convention reserves `<h1>` for `# Panics`, `# Examples`, `# Errors`, `# Safety`. Demoted all three to `## ` narrative subsections.
- **Broken-degrading intra-doc links on `#[cfg(test)]`-only targets**. `cargo doc --no-deps --features serve` runs without `--cfg test`, so `[apply_status_update_preserves_idle_entered_at_on_keep]` in [`src/tui/status_poller.rs`](src/tui/status_poller.rs) and `[decide_passive_transition_skips_patch_for_structured_session]` in [`src/server/mod.rs`](src/server/mod.rs) silently degrade to literal text (rustdoc does not emit a `broken_intra_doc_links` warning because the identifier is not resolvable as a Rust path in the non-`cfg(test)` compilation). Dropped the `[...]` brackets and left the names as backticked code-spans, with an inline rationale so a future reviewer does not "helpfully" restore the brackets.
- **Method-site pointer on the silent-drop guard**. [`Instance::merge_passive_status_patch`](src/session/instance.rs) now carries a short subsection reiterating that `last_accessed_at` older-or-equal is silently dropped (with a `session.store` debug log) while `status` / `idle_entered_at` still apply unconditionally. The contract already lives on [`PassiveStatusPatch.last_accessed_at`](src/session/instance.rs) and on the [`PassiveTransitionWrites`](src/server/mod.rs) design note; the method-site pointer closes the loop for a callsite reader who lands on the method definition directly.

Deferred, in-body-only follow-ups (no new tracking issues):

- **Sync flock on async TUI runtime + per-transition dispatch**: [`persist_passive_status_transition`](src/tui/home/mod.rs) calls [`Storage::update`](src/session/storage.rs) synchronously from inside the async TUI event loop, with no timeout on the 50ms flock-polling loop. The daemon's sibling path in [`api::persist_session_update`](src/server/api/sessions.rs) funnels through `spawn_blocking`; the TUI does not. A contended peer holding the flock stalls TUI rendering for the hold duration on every observed Running<->Idle transition. Additionally, per-transition dispatch means N transitions in one poll cycle produce N sequential flock round-trips from the TUI, versus the daemon's one-flock-per-profile batching in [`status_poll_loop`](src/server/mod.rs). Consolidating the TUI persist path onto `spawn_blocking` and adopting the daemon's per-tick batching shape is a structural refactor of the TUI event loop, out of scope for a review-follow-up PR.
- **Double `prev` HashMap allocation in `status_poll_loop`**: [`src/server/mod.rs`](src/server/mod.rs) rebuilds `prev` per tick and then `let prev_for_poll = prev.clone();` doubles the allocation for the `spawn_blocking` closure. At N sessions that is 2N `String` clones per tick. Options are `Arc<HashMap<String, Status>>` or move+rebuild after the join. Pre-existing shape, orthogonal optimization.
- **`Instance::new` seed semantic change**: [`Instance::new`](src/session/instance.rs) now seeds `live_status_baseline: None` (previously `Some(Idle)`). The field is `#[serde(skip)]` so persisted files are unaffected; the observable semantics of a freshly-constructed `Instance` changed such that the first `update_status_with_metadata` seeds the baseline rather than comparing against a pre-seeded `Idle`. Called out here so downstream workspace consumers (`xtask`, integration tests) and future test authors know the seed contract flipped.
- **`#[non_exhaustive]` on `IdleIntent` and `PassiveStatusPatch`**: absent today. Both are crate-internal (`pub(crate)`), so `#[non_exhaustive]` is a no-op currently. If either surface is promoted to a workspace-visible re-export in the future, add the attribute at that promotion PR rather than pre-emptively.
- **Pre-existing rustdoc `private_intra_doc_links` warnings**: `cargo doc --no-deps --features serve` emits 5 pre-existing warnings on files this PR does not touch (`src/containers/error.rs:9`, `src/server/api/sessions.rs:1010,1301`, `src/server/auth.rs:493,500`). All are `public documentation for X links to private item Y`. Out of scope; noted so a future `#![deny(rustdoc::broken_intra_doc_links)]` adoption knows the debt exists.
- **Feature-flag combinatorics**: `cargo build` (TUI-only, no `--features serve`) and `cargo build --features serve` both build clean at HEAD `bfa5aefe`. `PassiveTransitionWrites` lives inside the `#[cfg(feature = "serve")]`-gated `src/server/mod.rs`; `PassiveStatusPatch` and `IdleIntent` compile unconditionally, matching their non-server consumers. No gating change was required.

### Seventh-pass follow-ups (in-place)

A further review exploring angles not covered before (backward compat + on-disk `sessions.json` format, user-visible behavior, contributor documentation coverage, perf-at-scale) surfaced items addressed in commit `2c8ea58d`:

- **Contributor documentation: `## Passive-status pipeline` section on [`docs/development/internals/sessions.md`](docs/development/internals/sessions.md).** The design notes on `PassiveTransitionWrites`, `PassiveStatusPatch`, and `IdleIntent` live in `pub(crate)` rustdoc blocks and never appear in `cargo doc --no-deps` output. A new contributor grepping for "passive status" would find them, but a browsing contributor would not. The new section lifts the essentials (poller-authoritative-for-plain-tmux vs `apply_acp_overlay_inplace`-authoritative-for-structured split, daemon-batched vs TUI-per-transition writer shapes, wall-clock `>=` guard semantics, `live_status_baseline` `#[serde(skip)]` + `Instance::new`-time `None` seed construction-ordering rule) into a contributor-facing entry point that cross-links to the canonical `PassiveStatusPatch` glossary at `src/session/instance.rs`. Six touched files in the passive-status pipeline now have a single indexed home.
- **`docs/development/adding-file-watch-consumers.md` note on passive-status as a canonical `Storage::update` producer.** The debounce guidance already cites 75 ms as "kernel echo of `Storage::update` rename". Extended to note that the passive-status pipeline is the highest-cadence `Storage::update` producer under a busy set of sessions (one write per profile per ~2 s tick when transitions occur), so a `sessions.json` `FileWatchService` subscriber's event stream is dominated by passive-status writes.
- **`decide_passive_transition` fn docstring.** Was previously bare-signature at [`src/server/mod.rs`](src/server/mod.rs); the invariant sits on the return type `PassiveTransitionDecision`. A contributor jumping to the function by name saw no context. The fn now carries a short docstring pointing at `PassiveTransitionDecision` for the full contract (`patch: None` for structured / ACP sessions; `mark_unread` only on genuine Running -> Idle when unread is enabled and the row is not already unread).
- **`debug_assert_eq!(patch.id, inst.id)` at the flush site in `status_poll_loop`.** The sixth pass documented `PassiveTransitionWrites.patches: HashMap<String, PassiveStatusPatch>` key redundancy with `patch.id` as intentional. The assertion at [`src/server/mod.rs`](src/server/mod.rs) makes the redundancy compile-checked in debug builds: if `PassiveStatusPatch::from_instance` or the map-key assignment ever drift, the debug-build assertion catches it at the flush site rather than letting the redundancy hide a bug.

Deferred, in-body-only follow-ups (no new tracking issues):

- **`AGENTS.md` bullet for `src/tui/status_poller.rs`.** AGENTS.md is the routing map AI-assisted contributors read first; the TUI passive-status producer path is unlisted (as is the daemon's `status_poll_loop` in `src/server/mod.rs`). Not added here because the `AGENTS.md` project-structure bullets are minimal and adding entries selectively raises the noise. Consider a broader `docs/development/architecture.md` pass in a follow-up.
- **`## Passive-status pipeline` cross-links.** The Required section above lifts the invariants but keeps the canonical glossary on `PassiveStatusPatch`. A future consolidation pass could move the glossary to the doc and drop the rustdoc block, or duplicate carefully. Current split (invariants in doc, glossary on type) is defensible.
- **Rebase risk against upstream `main` advances since branch cut.** Recent upstream commits touch `src/tui/home/*` and `web/src/*`: `fe5b72fc fix(tui): reveal hidden idle sessions on w`, `98507b71 fix(tui): open project add form from empty picker`, `434e18b0 fix(tui): keep Trash collapsed on delete and skip trash in w-navigation`, `d2120b93 feat(web): support title-derived worktree sessions`, `61a6fd09 feat(ui): make session row suffix configurable`. All three tui commits touch attention/idle-adjacent paths this PR also modifies (`merge_passive_status_patch`, `apply_status_update`). Merge commits `9d438db3` pulled these in; the "Merge-conflict trap" section above stays relevant for a future rebase pass.
- **`PassiveTransitionWrites` future-field-addition invariant seal.** Prose-only invariant on future field additions ("neither monotone nor single-authority => unify writers first") is enforced by reviewer diligence + the struct docstring + this new doc section. A compile-time seal (sealed trait on `PassiveStatusPatch` fields) or a reflection-based lock test would be more durable but is out of proportion to the risk at current crate size. Deferred as note-only.
- **`status_poll_loop` per-tick perf-at-scale suggestions.** Suggested iterating `patches.keys()` (not `insts.iter_mut()`) at the flush site to drop O(N_disk × K_unread) to O(K_patches + K_unread), and switching `unread_ids: Vec<String>` to `HashSet<String>` for O(1) `contains`. Both are pre-existing shape; the current docstring at `src/server/mod.rs` on `PassiveTransitionWrites` explicitly justifies the `Vec` for "low per-tick cardinality". Suggestions apply only if scale grows past that assumption. Non-blocking, structural.
- **`bundles.entry(inst.source_profile.clone())` clone on entry-hit.** At K=1000 transitions on M=1 profile, 999 wasted clones. Micro-optimization visible only at burst-K. Non-blocking.
- **`AOE_PROFILE` env-var change in `status_hook_env_prefix`.** Flagged as a potential env-var API change for hook scripts; grep of the PR diff shows no occurrence, indicating this is a false-positive attribution to a commit outside this PR's scope. Not this PR's concern.

### User-visible behavior changes from #2690 fix

The Instance-seed flip + monotone `last_accessed_at` guard produce two user-observable behavior changes:

- **Fresh untouched Running sessions render a blank activity column instead of `<1m`.** Pre-fix, `Instance::new` seeded `live_status_baseline: Some(Idle)` and `finalize_launch` set `status: Starting`; the first `update_status_with_metadata` read baseline≠status as a genuine transition and fabricated a `last_accessed_at = Utc::now()`, yielding `<1m` in the age column. Post-fix, `live_status_baseline: None` means the first poll seeds the baseline without restamping, so `last_accessed_at` stays `None` until an explicit user touch (attach / send-keys). [`format_relative_age(None)`](src/tui/home/render.rs) returns `String::new()`, so the age column stays empty for fresh Running rows until the user attaches. Locked as intent by [`test_first_poll_after_status_write_does_not_fabricate_last_accessed_at`](src/session/instance.rs). Web sidebar's `.sort((a, b) => (b.last_accessed_at ?? "").localeCompare(a.last_accessed_at ?? ""))` clusters these rows at the tail until a touch.
- **Structured (ACP) session `idle_entered_at` displays the last durable value after daemon restart, until an ACP event re-emits.** Documented as a contract on [`apply_acp_overlay_inplace`](src/server/mod.rs) at the `## Durability contract` section: structured rows accept a soft reset of `status` / `last_accessed_at` / `idle_entered_at` on daemon restart, by contract. ACP event handlers are responsible for post-restart re-emission; the passive-status writer at `status_poll_loop` deliberately skips these rows via [`decide_passive_transition`](src/server/mod.rs)'s `patch: None`-for-`is_structured` branch. Web dashboard and TUI activity column may show a stale value for the ACP-reconnect window. No test locks the ACP re-emission timing; downstream users of structured / ACP sessions should be aware of the transient display gap on daemon restart.

Sound-hook path is unaffected: [`attached_status_hooks::apply_updates`](src/tui/attached_status_hooks.rs) fires `run_for_transition` only on `old != update.status`, so the baseline-seed-without-restamp on first TUI-relaunch poll does not synthesize a spurious transition. Daemon-path `StatusChange` emission on `src/server/mod.rs::status_poll_loop` still fires on the first tick after daemon restart if disk `prev.status` differs from detected; dwell filters in [`src/server/push.rs`](src/server/push.rs) (`DWELL_TERMINAL_MS = 2_000`) absorb transient noise. Deferred-notification behavior for while-daemon-was-down transitions is intentional.

Rebase-conflict foresight for a future rebase against upstream `main`: `fe5b72fc fix(tui): reveal hidden idle sessions on w (#2740)` adds ~141 lines to `src/tui/home/tests.rs`, `61a6fd09 feat(ui): make session row suffix configurable (#2713)` rewrites ~219 lines in the same file plus ~125 lines in `src/tui/home/render.rs`. This PR adds `apply_status_update_preserves_idle_entered_at_on_keep` (+45 lines) and updates six existing test call sites in `src/tui/home/tests.rs`. The merge commit `9d438db3` pulled the upstream advances in; a future rebase pass against a further-advanced main should re-check `src/tui/home/tests.rs` and `src/tui/home/render.rs:1420-1495` for conflicts with this PR's `merge_passive_status_patch` and `apply_status_update` changes.

### Locking tests

- [`test_instance_new_seeds_live_status_baseline_none`](src/session/instance.rs)
- [`test_first_poll_after_status_write_does_not_fabricate_last_accessed_at`](src/session/instance.rs)
- [`apply_status_update_preserves_idle_entered_at_on_keep`](src/tui/home/tests.rs)
- [`test_merge_passive_status_patch_last_accessed_at_boundary_equal_is_a_noop`](src/session/instance.rs) (Sug #6 lock)
- [`decide_passive_transition_skips_patch_for_structured_session`](src/server/mod.rs) (Req #4 anchor, structured-session skip lock)
- [`test_status_update_default_is_no_op`](src/tui/status_poller.rs) (FU #3 lock)
- [`poll_statuses_once_never_emits_idle_intent_keep`](src/tui/status_poller.rs) (fifth-pass producer-asymmetry lock)
- [`apply_updates_maps_idle_intent_set_and_clear_arms`](src/tui/attached_status_hooks.rs) (fifth-pass consumer-arm-mapping lock)

### Commit metadata observations

- Commit `342799f6` type prefix reads `docs(session):`; the commit also flips a runtime guard (Sug #6 `>` → `>=`), which is technically a `refactor(session):`. Observable value is unchanged either way. Not re-typed (no force-push).
- Commit `5bcd033b` body opens with a sentence that includes both `#2729` self-reference and meta-review vocabulary. Left as-is (no force-push). The commit body content is otherwise accurate.

### Merge-conflict trap

Rebase note for the eventual merge to `main`: the load-bearing single-line change on this branch is `Instance::new`'s `live_status_baseline` seed, which flips from `Some(Status::Idle)` on `upstream/main` to `None` on this branch. `test_instance_new_seeds_live_status_baseline_none` in `src/session/instance.rs` locks the fix, but only if the test itself and the seed line are both resolved to the branch state. Handle the conflict on that line deliberately.

### Test-suite state

Seventh-pass run (post-commit `2c8ea58d`): `cargo test --features serve --lib` targeted `server::tests::decide_passive_transition_*`: 4 passed, 0 failed. `cargo build`, `cargo build --features serve`, `cargo clippy --features serve --all-targets` (with `-D warnings`), and `cargo fmt --check` all clean. LSP diagnostics clean on the touched files. Seventh-pass diff is docs-only (2 doc files + 1 docstring + 1 debug_assert on `src/server/mod.rs`), so the sixth-pass full-suite state carries forward. Sixth-pass run (post-commit `bfa5aefe`): `cargo test --features serve --lib`, 4869 passed, 1 retry on `tui::home::file_watch_tests::rewire_disk_invalidates_on_inode_change_with_same_canonical_path` (per open issue #2600), retry green. Fifth-pass run (post-commit `63ea23f6`): `cargo test --features serve --lib`, 4869 passed, 0 failed, 0 retries needed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Anthropic Claude Opus 4.7 (via OpenCode CLI).

The baseline-seed fix chose `None` in `Instance::new` over a broader refactor. The asymmetric-propagation fix chose the tri-state `IdleIntent` enum over guarding `idle_entered_at` conditionally (would break the Clear-on-transition-out-of-Idle signal exercised by `apply_status_update_clears_idle_entered_at_on_idle_to_running`), and over making baseline unconditional (would clobber real values on the `attached_status_hooks` snapshot path). `live_status_baseline` stays `Option<Status>` because it is never intentionally Cleared by any producer.

The fifth-pass follow-ups chose in-place edits to the same commits' anchor sites over new abstraction: the wall-clock note lives on the merge docstring rather than a `MonotoneOrWallClockGuarded<T>` newtype; the construction-ordering note lives on the field doc rather than a runtime assertion; the persist-error log lands as a `tracing::warn!` rather than a `Result` bubbled out. Each choice prefers documentation and localized observability over structural change, consistent with the review-follow-up scope.

This PR tightens how a session’s “idle” state is tracked and persisted across the TUI, the background status poller, attached status hooks, and the storage layer.

What changed:
- Reworked idle handling around a clear 3-way “intent” model (Set / Clear / Keep) instead of using a single optional timestamp.
- Fixed initial/baseline behavior so new instances no longer create a misleading “idle last accessed” value on the first poll after construction/reload.
- Updated the TUI and attached status hook code paths to correctly translate the poller’s idle intent when updating `idle_entered_at`.
- Clarified persistence rules for different session types:
  - Structured (ACP) sessions can safely “soft-reset” certain fields on daemon restart, but passive status writers intentionally do not overwrite those fields.
  - Passive transition persistence was refactored to batch related updates together (patches + unread markers) and skip persistence when there’s nothing meaningful to write.
  - Passive persistence failures are now logged (while keeping the existing best-effort behavior).
- Renamed/standardized a few internal types for consistency, and improved rustdoc/design notes and comments around the persistence/authority model.
- Expanded and adjusted tests to cover baseline seeding, no-op behavior, boundary timestamp handling, correct producer→consumer intent mapping, and default update semantics.

Benefits:
- Prevents idle time from being incorrectly reset or “fabricated” after reloads/restarts.
- Reduces the risk that “no observation yet” accidentally clears a real idle baseline.
- Makes passive persistence more reliable by avoiding redundant writes and documenting/guarding authority rules between structured-session handling and passive patching.
- Increases confidence through targeted edge-case and locking-focused test coverage.


jerome-benoit:
Closed at author's request. This PR was created without explicit authorization: the review of #2697 was requested but a follow-up PR was not. My apologies for the overreach.

jerome-benoit:
Reopening: the closure was based on my misreading. Continuing implementation of the remaining findings in-place.

Issue #2372, opened 2026-06-24
Web dashboard: sidebar project count badge includes archived and snoozed sessions, disagrees with the row list below it


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

The web sidebar project header shows a session count badge, like `agent of empires (8)`, but the number includes archived and snoozed sessions. It should reflect only the live sessions, so a project with 5 live sessions and 3 archived/snoozed ones should read `(5)`, not `(8)`.

The badge sits directly above the list of session rows for that project. The list already hides archived and snoozed entries, so today the count and the list it labels disagree: the header says 8 while only 5 rows are visible.

This is the follow-up to an open question left unresolved in #2207, which added the count:

> Count semantics confirmed as total sessions across workspaces (not workspace count). The snoozed/archived count uses workspace count, so the two counts mean slightly different things; acceptable or should they match?

They were shipped not matching. This issue resolves that: the badge should equal the number of rows rendered below it.

## What's there today

The count is computed by summing every session in the group, with no live/sunk filter, `web/src/components/WorkspaceSidebar.tsx:1909`:

```tsx
const sessionCount = group.workspaces.reduce((n, v) => n + v.workspace.sessions.length, 0);
```

It renders in the header badge, `web/src/components/WorkspaceSidebar.tsx:2090-2092`:

```tsx
<span className="shrink-0 text-[12px] tabular-nums text-text-dim" data-testid="sidebar-group-session-count">
  ({sessionCount})
</span>
```

The session-row list right below it already filters out sunk workspaces, `web/src/components/WorkspaceSidebar.tsx:2900`:

```tsx
const liveWorkspaces = group.workspaces.filter((v) => !workspaceIsSunk(v.workspace));
```

`workspaceIsSunk` is the canonical "all sessions archived or snoozed" predicate, `web/src/lib/sidebarSort.ts:63-66`:

```tsx
export function workspaceIsSunk(ws: Workspace): boolean {
  if (ws.sessions.length === 0) return false;
  return ws.sessions.every((s) => s.archived_at != null || s.snoozed_until != null);
}
```

[...]

PR #2732, opened 2026-07-07
fix(tui): exclude trashed sessions from sidebar group count

### LINKED ISSUE #2731: tui: sidebar group session-count includes trashed sessions, disagrees with the row list below it
**Note:** I drove the investigation with an AI assistant (Claude Code) for the code tracing and prose; the report, decisions, and testing are mine. — @MatthewWolff

## Problem

In the TUI left pane (the sidebar), every user group renders a session-count number next to its name. That count does not update when a session in the group is moved to the Trash: the row correctly disappears from the group's list (it moves to the synthetic Trash section), but the header number stays the same, so the count over-reports. Restoring a trashed session should bring the number back, but since it was never decremented, the two operations both read wrong.

The count sits directly above the session rows for that group. The row list already hides archived and trashed entries, so today the count and the list it labels disagree: a group with 2 sessions still reads `(2)` after one is trashed, even though only 1 row is visible.

This is the TUI counterpart of #2372, which fixed the same class of bug on the web sidebar (the badge there counted archived/snoozed sessions the row list hid). The web sidebar is now correct; the TUI was never updated to exclude trashed sessions from its count.

## What's there today

The count is computed by `count_sessions_in_group`, which filters on `!i.is_archived()` only, with no trash exclusion (`src/session/groups.rs:938` on `main`):

```rust
fn count_sessions_in_group(path: &str, instances: &[Instance]) -> usize {
    let prefix = format!("{}/", path);
    instances
        .iter()
        .filter(|i| (i.group_path == path || i.group_path.starts_with(&prefix)) && !i.is_archived())
        .count()
}
```

It feeds the `session_count` on the flattened group item in `flatten_group` (`src/session/groups.rs:889`), which is what the header renders.

The session rows rendered under the same group, in that same `flatten_group`, already exclude both archived and trashed (`src/session/groups.rs:910`):

```rust
.filter(|i| i.group_path == group.path && !i.is_archived() && !i.is_trashed())
```

And the canonical membership helper `group_members` (used by the group sort keys) carries the correct predicate (`src/session/groups.rs:438`):

```rust
instances.iter().filter(move |i| {
    (i.group_path == path || i.group_path.starts_with(&prefix))
        && !i.is_archived()
        && !i.is_trashed()
})
```

## Root cause

`count_sessions_in_group` and the row filter drifted. The count keeps only `!is_archived()`, while the rows it labels keep `!is_archived() && !is_trashed()`. Any trashed session in the group therefore inflates the header number above the visible row count, and untrashing has no compensating effect because the count never dropped.

## Proposal

Make the count reuse the exact predicate the rows and the sort keys already use, so the badge can never contradict the list. `group_members(path, instances)` is that predicate (same nested-prefix match, plus the archived+trashed exclusion), so:

```rust
fn count_sessions_in_group(path: &str, instances: &[Instance]) -> usize {
    group_members(path, instances).count()
}
```

One source of truth for "sessions that render under this group", no behavior change for the archived case, and the trash case is now correct.

Implementation MUST add a test covering the story below; PR is not done without it.

- As a user, when a group has two sessions and I trash one, then the group header count drops from `(2)` to `(1)` and matches the one row still shown; when I restore it, the count returns to `(2)`.
  - Suggested test layer: Rust unit test in `src/session/groups.rs` asserting the flattened `Item::Group { session_count }` after `trash()` / `untrash()`.

## Out of scope

- The web sidebar count (already correct via `workspaceIsSunk`, fixed in #2372).
- The `/api/groups` `session_count` field and the CLI `aoe group list` count, which also include trashed sessions but are not surfaced as the sidebar badge the report is about.
- The synthetic Trash and Archived section headers, which intentionally count their own members.

## Cross-references

- #2372 - the web sidebar analog of this bug; this issue is the TUI counterpart.

MatthewWolff:
## Description

In the TUI left pane, each user group renders a session-count number next to its name. That number did not update when a session in the group was moved to the Trash: the row correctly disappeared from the group's list (it moves to the synthetic Trash section), but the header count stayed the same, so it over-reported. Because the count never dropped on trash, restoring a session never corrected it either.

Root cause: `count_sessions_in_group` (`src/session/groups.rs`) filtered only on `!is_archived()`, while the session rows it labels, rendered in the same `flatten_group`, exclude both archived and trashed (`!is_archived() && !is_trashed()`). The count and the rows had drifted onto different predicates.

The fix routes the count through the existing `group_members` helper, which already carries the correct predicate (same nested-prefix match, plus the archived and trashed exclusions) and is what the group sort keys use. Now the badge and the visible rows share one source of truth, so they cannot disagree.

This is the TUI counterpart of #2372, which fixed the same class of bug on the web sidebar; the web count was already correct via `workspaceIsSunk`.

Closes #2731.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the defect is a pure count-computation drift, fully and deterministically covered by two in-crate regression tests that exercise the real rendering pipeline: `group_header_count_tracks_trash_and_restore` (drives `HomeView::build_flat_items` and asserts the group header count drops on trash and returns on restore) and `test_session_count_excludes_trashed_and_restores_on_untrash` (unit test on `flatten_tree`). Both were confirmed to fail before the fix and pass after. A tmux-driven `tests/e2e/` test would add flakiness for no additional coverage of this data path.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** I used Claude Code to trace the count/render predicate drift and draft the tests and prose. I directed the investigation, reviewed and verified every change myself (including reproducing the failing assertions before the fix and confirming fmt/clippy clean), and understand all of it.

This PR fixes a TUI sidebar issue where each session group’s header count could disagree with the number of sessions actually shown underneath it.

- Group header counts now use the same filtering logic as the rendered group rows, so trashed sessions no longer inflate the count.
- Added regression coverage to ensure the count decreases when a session is trashed and returns when it’s restored, using the real trash/restore flow rather than direct state mutation.

Benefit: the sidebar badge and the session list stay in sync, preventing confusing “wrong count” moments when trashing or restoring sessions.

### DIFF
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -1362,7 +1362,6 @@ mod tests {
         instances[0].trash();
         assert_eq!(group_count(&instances), 1);
 
-        // Restoring brings the count back.
         instances[0].untrash();
         assert_eq!(group_count(&instances), 2);
     }
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -935,12 +935,11 @@ fn flatten_group(
     }
 }
 
+/// Count of the sessions that render under a group header. Delegates to
+/// `group_members` so the header badge and the visible rows share one
+/// predicate (excluding archived and trashed sessions).
 fn count_sessions_in_group(path: &str, instances: &[Instance]) -> usize {
-    let prefix = format!("{}/", path);
-    instances
-        .iter()
-        .filter(|i| (i.group_path == path || i.group_path.starts_with(&prefix)) && !i.is_archived())
-        .count()
+    group_members(path, instances).count()
 }
 
 /// Append the synthetic "Archived" section to `items`, pinned to the
@@ -1330,6 +1329,42 @@ mod tests {
         {
             assert_eq!(*session_count, 2);
         }
+    }
+
+    #[test]
+    fn test_session_count_excludes_trashed_and_restores_on_untrash() {
+        let mut a = Instance::new("a", "/tmp/a");
+        a.group_path = "work".to_string();
+        let mut b = Instance::new("b", "/tmp/b");
+        b.group_path = "work".to_string();
+        let mut instances = vec![a, b];
+
+        let group_count = |instances: &[Instance]| {
+            let tree = GroupTree::new_with_groups(instances, &[]);
+            let items = flatten_tree(&tree, instances, SortOrder::Oldest);
+            items
+                .iter()
+                .find_map(|i| match i {
+                    Item::Group {
+                        path,
+                        session_count,
+                        ..
+                    } if path == "work" => Some(*session_count),
+                    _ => None,
+                })
+                .expect("work group present")
+        };
+
+        assert_eq!(group_count(&instances), 2);
+
+        // Trashing a session drops it from both the visible rows and the
+        // header badge, matching the row filter in `flatten_group`.
+        instances[0].trash();
+        assert_eq!(group_count(&instances), 1);
+
+        // Restoring brings the count back.
+        instances[0].untrash();
+        assert_eq!(group_count(&instances), 2);
     }
 
     #[test]
--- a/src/tui/home/tests.rs
+++ b/src/tui/home/tests.rs
@@ -2541,13 +2541,14 @@ fn create_test_env_with_group_sessions() -> TestEnv {
     TestEnv { _temp: temp, view }
 }
 
-/// The sidebar group header count is rebuilt through the same pipeline the
-/// rows use, so trashing a session drops the count and restoring it brings
-/// the count back. Guards against the count and the visible rows drifting.
+/// Trashing and restoring a session through the view's own actions keeps the
+/// group header count in step with the rows, since both are rebuilt from the
+/// same predicate. Guards against the count and the visible rows drifting.
 #[test]
 #[serial]
 fn group_header_count_tracks_trash_and_restore() {
     let mut env = create_test_env_with_group_sessions();
+    env.view.trashed_section_collapsed = false;
 
     let work_count = |env: &TestEnv| -> usize {
         env.view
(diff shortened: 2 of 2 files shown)


Seluj78 (approved):

## AoE Review

Reviewed: PR #2732 / fix/tui-group-count-excludes-trashed (base: main@f54f543). Files changed: 2 (src/session/groups.rs, src/tui/home/tests.rs).
Unresolved comments: none.

### Verdict

**Implementation:** Approve
**Project Fit:** Accept

Real, documented bug (linked issue #2731, precedent #2372 fixed the same class on web sidebar). count_sessions_in_group filtered only on !is_archived(); the fix routes it through the existing group_members helper, which already carries the correct predicate (nested-prefix match plus archived and trashed exclusion) and is what the group sort keys use. group_members (src/session/groups.rs:438-448) matches the row filter used in flatten_group. toggle_archive_at_cursor (src/tui/home/operations.rs:1373) correctly routes trashed-selection restore through restore_selected_from_trash(), confirming the new HomeView test drives a real restore path.

Scope stays tight to the count/render predicate drift, no unrelated changes. Tests cover both directions: a unit test in groups.rs and an integration-style test in tui/home/tests.rs, both asserting count drop-on-trash and restore-on-untrash, with #[serial] on the TUI test. Phase 1 mechanical probes (em dash, dead code, tracing target/format, secrets, namespace hardcode, git config) all clean. PR metadata clean: branch fix/..., conventional commit prefixes, AI Usage section filled, no em dashes.

No Blocker, Required, Followup, or Nit findings.

Issue #1805, opened 2026-06-02
tui: w on a collapsed group or active session should still jump to next idle


grepsedawk:
> Note: prose below is the issue author's words; AI helped format and look up code references.

When you're in Attention mode and you press `w` on a collapsed group, it should still take you to the next idle agent.

When you press `w` on a green/active session, it should not offer to snooze, and instead should go to the next available idle session.

Code pointers for context (AI-added, not author's words):
- `src/tui/home/input.rs:2241` — bare `w` calls `jump_to_next_waiting()` (snooze moved off `w`/`W` in #1524).
- `src/tui/home/input.rs:2410` — `jump_to_next_waiting()` walks from `cursor + 1` for Waiting or freshly-idle, then falls back to MRU Idle. It only inspects `Item::Session` rows, so a `Group` cursor should still drive forward; worth verifying that the group case lands on the next idle as the author expects, and that no other handler intercepts `w` on an active row.

Seluj78:
> AI note: I drafted this triage comment with Codex. The classification and triage calls are mine.

Likely root cause and target shape: `w` is still the `NextWaiting` action in home-view bindings (`src/tui/home/bindings.rs:628-636`) and dispatches to `jump_to_next_waiting()` (`src/tui/home/input.rs:1986`). That routine already owns the whole behavior: it skips non-session rows, forward-scans for `Waiting` or freshly-idle sessions, then falls back to the most-recently-accessed idle row (`src/tui/home/input.rs:2364-2435`). The target shape is therefore a small logic fix inside that scanner so collapsed-group and active-session cases land on the next actionable idle row instead of surfacing the wrong interaction.

Type: bug.

Suggested size: S. The surface area is a single TUI navigation routine plus tests.

Suggested priority: P2. This is a real interaction bug in Attention mode, but it has a straightforward workaround and does not block session work entirely.

njbrake:
We'd be happy to accept a PR if you have the cycles to pick it up! 🙏

PR #2750, opened 2026-07-09
fix(tui): make w skip snoozed sessions when jumping to next attention

athal7:
## Description
`w` (jump to next attention-needing session) never checked `Instance::is_snoozed()`, even though every other subsystem (recovery, ACP reconciler, plugin API, render) treats a snoozed session as a sunk "don't bother me" state alongside archived/trashed. This let `w` land on a snoozed session if it was `Waiting` or the only `Idle` fallback candidate. Added the `is_snoozed()` exclusion to all four filter passes inside `jump_to_next_waiting`.

## PR Type
- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: this is a unit-test-level filter fix inside `jump_to_next_waiting`; added two unit regression tests (`w_skips_snoozed_waiting_session`, `w_skips_snoozed_idle_session_in_fallback`) covering the exact behavior instead of an e2e test, since no rendering/CLI/session-lifecycle surface changed.

## AI Usage
- [x] AI was used

**AI Model/Tool used:** Claude (opencode, anthropic/claude-sonnet-5)

**Any Additional AI Details you'd like to share:** Full agent-orchestrated change: searched for a pre-existing issue (none found matching this exact bug — related issues #1524 and #1805 addressed the `w`/snooze key conflict but not this filter gap), then implemented the fix and two new regression tests in `src/tui/home/tests.rs`, modeled on the existing `w_skips_unread_trashed_session` test. `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, and the full `cargo test` suite (3746 tests) pass.

This update fixes the TUI `w` key “jump to next attention-needed session” so it reliably skips sessions the user has dismissed—now including snoozed and archived sessions, not just trashed ones. As a result, the cursor won’t land on sessions that were intentionally marked as “do not bother,” even in idle/fallback navigation paths.

To support this, it adds a shared `Instance::is_dismissed()` helper and updates all selection passes inside `jump_to_next_waiting` to consistently filter out dismissed instances.

Two regression tests were added:
- `w_skips_snoozed_waiting_session`
- `w_skips_snoozed_idle_session_in_fallback`

The PR also expands coverage to ensure the same skip behavior applies to archived sessions during the same `w` navigation flow.

Additional stability/maintenance tweaks:
- Refactors `parse_chord`’s `F`-key handling without changing behavior.
- Improves intro wizard E2E reliability by waiting for the “(6/6)” indicator to disappear before proceeding.

Benefits: fewer surprising cursor jumps, more predictable navigation when snoozed/archived sessions exist, and more stable end-to-end test runs.

```text
Technical (optional):
- Introduced Instance::is_dismissed() (trashed/snoozed/archived) and folded archive handling into dismissed filtering across all passes in jump_to_next_waiting.
- Added regression tests for snoozed and archived skip behavior, including idle fallback behavior.
```

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1658,12 +1658,12 @@ impl Instance {
         self.snoozed_until.map(|t| t > Utc::now()).unwrap_or(false)
     }
 
-    /// Combined "don't bother me" sink-state check: trashed or snoozed.
-    /// Callers that walk sessions looking for something to land on (e.g. the
-    /// `w`/jump-to-next-attention passes) use this instead of the two-call
-    /// form so a row in either sink state is uniformly excluded.
+    /// Combined "don't bother me" sink-state check: trashed, snoozed, or
+    /// archived. Callers that walk sessions looking for something to land on
+    /// (e.g. the `w`/jump-to-next-attention passes) use this instead of the
+    /// three-call form so a row in any sink state is uniformly excluded.
     pub fn is_dismissed(&self) -> bool {
-        self.is_trashed() || self.is_snoozed()
+        self.is_trashed() || self.is_snoozed() || self.is_archived()
     }
 
     /// Remaining snooze duration as a `chrono::Duration`, or `None` if the
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1658,6 +1658,14 @@ impl Instance {
         self.snoozed_until.map(|t| t > Utc::now()).unwrap_or(false)
     }
 
+    /// Combined "don't bother me" sink-state check: trashed or snoozed.
+    /// Callers that walk sessions looking for something to land on (e.g. the
+    /// `w`/jump-to-next-attention passes) use this instead of the two-call
+    /// form so a row in either sink state is uniformly excluded.
+    pub fn is_dismissed(&self) -> bool {
+        self.is_trashed() || self.is_snoozed()
+    }
+
     /// Remaining snooze duration as a `chrono::Duration`, or `None` if the
     /// session isn't snoozed (or the timestamp has already expired).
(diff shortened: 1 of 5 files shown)


Seluj78 (approved):

**Approve.** Clean, surgical fix.

`w` (jump-to-next-attention) was the one navigation path ignoring snooze/archive, even though recovery, the ACP reconciler, the plugin API, and render all treat those as "don't bother me" sink states. This routes all four passes of `jump_to_next_waiting` through the new `Instance::is_dismissed()` (trashed / snoozed / archived) and drops the now-redundant standalone `is_archived()` checks in the hidden passes. Verified the union covers every pass; `is_dismissed()` has exactly those four callers, no dead code.

Tests are strong: snoozed and archived, each for the Waiting and the idle-fallback path, each with a third eligible Waiting control session so `w` must land on a real target rather than trivially not-landing on the dismissed row. Codecov reports all modified lines covered.

Two unrelated changes ride along, both self-explained and CI-unblocking: the `bindings.rs` F-key `?`-operator rewrite (pre-existing clippy::question_mark on main) and the `intro.rs` wait_for_absent("(6/6)") race guard for a flaky e2e. Verified the F-key refactor is behavior-preserving.

Minor: branch `snooze-attention` skips the `fix/` prefix convention, and the title understates the fix (it now also covers archived, not just snoozed). Neither blocks merge.

njbrake (approved):

Reviewed in batch triage. w now skips snoozed/archived when jumping to next attention; all four filter passes are bounded, all-dismissed falls through to the dialog, and snoozed/archived cases are covered by tests. LGTM.

PR #2773, opened 2026-07-10
fix(plugin): recover and observe stopped plugin workers without a daemon restart

### LINKED ISSUE #2769: GitHub plugin worker stops launching entirely after ~2.5h of normal operation; doesn't recover across dashboard restarts
**Summary**
The `agent-of-empires.github` plugin worker ran normally for ~2.5 hours, then stopped launching. It has not recovered across two subsequent fresh dashboard restarts, disabling/re-enabling the plugin, or the plugin's own `refresh` command.

**Environment**
- aoe 1.12.1 (macOS/arm64, installed via Homebrew)
- Plugin: `agent-of-empires.github` v1.7.2 (featured), installed from `gh:agent-of-empires/plugin-github`
- Capabilities granted: `net`, `runtime.worker`, `session.read`, `notifications`, `browser_open`

**Timeline**
1. **11:09–13:44 (Jul 9):** Plugin worked correctly — worker log shows continuous successful GitHub GraphQL/REST polling (HTTP 200/304, healthy rate limit) across several tracked branches in 3 different repos.
2. **13:44 onward:** No further worker activity or log output.
3. **16:02:41 (new dashboard session, PID 68034):** `debug.log` shows the plugin host attempting to launch the worker **6 times** between 16:09:20–16:15:37 (`plugin.host: launching plugin workers` → `launched plugin worker ... pid=<N>`), then giving up permanently for the rest of that ~4.5h session. Each launch:
   - Successfully forked/exec'd (host logged a pid each time)
   - Never produced a new entry in `~/.agent-of-empires/plugin-workers/` (no log file created)
   - Never appeared in `ps` afterward (process not alive)
   - Never logged an error, warning, panic, or exit reason in `debug.log`
4. Manually invoking the worker binary directly (`~/.agent-of-empires/plugins/agent-of-empires.github/.aoe-build/venv/bin/aoe-github-worker`) confirmed it's healthy standalone: it starts and correctly emits its startup JSON-RPC calls (`config.get`, `sessions.list`) over stdout before blocking on a response — so the binary/venv is not the problem.
5. **Fully quit and restarted the dashboard (new PID 29752, 20:34:26):** in 5+ hours since, the plugin host has not attempted a single launch — worse than the previous session's 6 retries.
6. Tried `aoe plugin disable agent-of-empires.github && aoe plugin enable agent-of-empires.github` between sessions — this only updates on-disk config (`aoe plugin list` correctly shows `enabled`); it produces no corresponding activity in a running dashboard process and did not induce a relaunch attempt in the subsequent fresh session either.
7. Tried `aoe refresh` (the plugin's exposed refresh command) directly from a shell — fails with `Error: stdin is not a terminal; aoe requires an interactive TTY`, so it can only be triggered from inside the live dashboard; have not yet been able to test whether triggering it in-app forces a relaunch.

**Expected behavior**
The plugin host should either successfully launch and keep the worker alive, or surface a visible error/warning when launches fail repeatedly — not silently stop attempting launches indefinitely with no log trace.

**Actual behavior**
After a healthy ~2.5h run, the worker never restarts, across multiple fresh dashboard sessions, with zero corresponding log output at any level.

**Config (for reference)**
```toml
[plugins."agent-of-empires.github"]
enabled = true
source = "gh:agent-of-empires/plugin-github"

[plugins."agent-of-empires.github".grant]
capabilities = ["net", "runtime.worker", "session.read", "notifications", "browser_open"]
```

Seluj78:

## Description

The `aoe serve` plugin worker host launched workers exactly once, at daemon boot (`server/mod.rs` calling `PluginHost::start`), and nothing ever relaunched after that. A worker that exhausted its respawn budget, or a plugin that loaded inactive, was dead for the rest of the daemon's life, and the reason was never written to `debug.log`: registry skip reasons went into a `load_errors` vector surfaced only in the plugin manager UI and `aoe plugin list`, so a boot that launched zero workers logged `count=0` with no explanation. This is what #2769 hit: the github plugin worker stopped and did not come back across restarts, enable/disable, or refresh, with no log trace.

This PR makes the failure observable and the recovery possible without a full daemon restart:

- Launch now runs through one idempotent `PluginHost::reconcile`, which starts a worker for every active runtime plugin that has none and tears down any worker whose plugin is no longer active. `start()` is that reconcile plus a WARN for each load error and each enabled-but-inactive runtime plugin, so a zero-launch boot names the reason (stale grant, host-version mismatch, disabled) instead of going silent.
- The web enable/disable handler (`POST /api/plugins/{id}/enabled`) calls reconcile after the config write, so toggling a plugin in the dashboard launches or stops its worker live.
- On respawn-budget exhaustion the host records an in-memory crash tombstone and raises a dashboard notification. A tombstoned plugin is not revived by an unrelated plugin's reconcile; disabling it clears the tombstone, so a disable then enable is a clean retry.

Worker lifecycle state (running map, crash tombstones, supervisor-id counter) is folded into one `WorkerTable` mutex, and every supervisor table write is gated on its supervisor id so a stale supervisor cannot corrupt a newer entry after a live teardown and relaunch.

Out of scope, noted for triage: CLI `aoe plugin enable|disable` still does not reach a running daemon (separate process, no IPC), so live recovery needs the web toggle or a restart; and no periodic self-heal timer is added (a worker that dies with no user action still waits for a toggle or restart). Both are follow-ups if wanted.

Closes #2769

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Enable a runtime plugin whose worker is not running via the dashboard (`POST /api/plugins/{id}/enabled`), confirm the worker launches without restarting `aoe serve` (a `launched plugin worker` line in `debug.log`, a live worker log under `~/.agent-of-empires/plugin-workers/`).
- [ ] Disable that plugin from the dashboard, confirm the worker is torn down (no surviving process, `stopped plugin worker` in `debug.log`).
- [ ] Boot the daemon with a runtime plugin that is enabled on disk but inactive (revoke its grant so it needs reapproval), confirm `debug.log` carries a WARN naming the reason instead of a silent `count=0`.
- [ ] Force a worker to crash-loop past the respawn budget, confirm a dashboard notification appears and the worker is not relaunched until the plugin is disabled then re-enabled.

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: the change to `POST /api/plugins/{id}/enabled` is server-side recovery behavior with no browser-observable UI change and no `web/src` edit; exercising it end-to-end would require a real plugin worker process. The reconcile decision logic is covered by Rust unit tests (`plan_reconcile` launch / teardown / idempotent / crashed-skip / cap-truncation / off-then-on retry) in `src/plugin/host.rs`.

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle; it changes the serve-daemon plugin worker host and one REST handler.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via Claude Code, agent-of-empires `/aoe-implement` skill, with a `/debate` design pass between gemini-3.1-pro and gpt-5.5)

Investigation, the debated design, and implementation were drafted by Claude under my direction. I made the scope call (observability plus web-triggered recovery, no periodic timer), reviewed each commit, and ran the manual test steps.

**NOTE:**

## Summary
- Reworked plugin worker lifecycle to use an idempotent reconciliation step (`PluginHost::reconcile`) that:
  - starts workers for plugins that are active in the registry but missing workers
  - stops workers whose plugins are no longer active
- Added live reconciliation after web-side plugin enable/disable changes so worker state updates immediately without restarting the daemon.
- Improved crash handling and observability:
  - when a worker exhausts its respawn budget, the host records an in-memory “crash tombstone” and surfaces a dashboard notification
  - tombstoned plugins are skipped by unrelated reconciliations
  - disabling and re-enabling a plugin (or restarting the daemon) clears the tombstone to allow a retry
- Strengthened worker lifecycle coordination to avoid race conditions/stale supervisors affecting newer worker entries.
- Updated internal documentation and added unit tests covering reconciliation planning/idempotency and crash-tombstone behavior.

## Benefits
- Plugin workers recover deterministically and continue attempting launches while the plugin remains active.
- Repeated failures are visible (warnings/notifications) instead of silently stopping recovery attempts.
- Configuration changes via the web UI take effect immediately in the running daemon.

## Potential follow-up
- Add periodic self-healing / recovery for CLI-based enable/disable operations (explicitly out of scope in this PR).
- Consider persisting crash state beyond daemon restarts if “tombstone survives restart” becomes a requirement.

### DIFF
--- a/docs/development/internals/plugin-system.md
+++ b/docs/development/internals/plugin-system.md
@@ -548,6 +548,25 @@ that forks helpers is torn down whole), a per-worker respawn budget so a crash
 loop does not spin, and a concurrency cap. The worker's stderr drains to
 `<app_dir>/plugin-workers/<id>.log`.
 
+### Recovery and observability
+
+Launch runs through one idempotent `PluginHost::reconcile`: it starts a worker
+for every active runtime plugin that has none and tears down any worker whose
+plugin is no longer active. `start()` at daemon boot is that reconcile plus a
+WARN for each load error and each enabled-but-inactive runtime plugin, so a boot
+that launches zero workers names the reason (a stale grant, a host-version
+mismatch) in `debug.log` instead of going silent. The web enable/disable handler
+(`POST /api/plugins/{id}/enabled`) calls reconcile too, so toggling a plugin
+launches or stops its worker live, without a daemon restart.
+
+When a worker exhausts its respawn budget the host records an in-memory crash
+tombstone and surfaces a dashboard notification. A tombstoned plugin is not
+revived by an unrelated plugin's reconcile; disabling it clears the tombstone,
+so a disable then enable is a clean retry. A daemon restart also clears it. The
+CLI `aoe plugin enable|disable` mutates only on-disk config and does not reach a
+running daemon (it is a separate process), so recovering a live daemon needs the
+web toggle or a restart.
+
 ### Capability-gated host API
 
 Each host method maps to a capability the plugin declared and was granted; the
--- a/src/plugin/host.rs
+++ b/src/plugin/host.rs
@@ -51,6 +51,14 @@ const REAP_GRACE: Duration = Duration::from_secs(2);
 
 /// One supervised worker: the plugin it runs, its pid, and the task driving it.
 struct RunningWorker {
+    /// Identity of the supervisor task that owns this slot. A stale supervisor
+    /// (one whose worker was torn down and whose plugin was relaunched under a
+    /// new supervisor) must not mutate or remove the newer entry: every table
+    /// write it makes is gated on this id still matching. Without it, an
+    /// aborted-but-not-yet-yielded supervisor can ABA-corrupt a fresh entry,
+    /// because an uncontended `lock().await` resolves inside the same poll,
+    /// before the abort takes effect.
+    supervisor_id: u64,
     pid: u32,
     task: tokio::task::JoinHandle<()>,
     /// Sender into the worker's stdin, set while a worker generation is being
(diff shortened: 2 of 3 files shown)


Seluj78:
@coderabbitai review

PR #2777, opened 2026-07-10
feat(tui): auto-enter live-send on view switch, support Tool view

athal7:
## Description
Adds a `live_send_on_view_switch` setting: when enabled and Default Attach Mode is
Live mode, switching a selected session into Terminal view (`t`) or a Tool view (a
tool hotkey, e.g. lazygit/yazi) now automatically starts live-send, so the user can
type directly into the pane through the live preview without pressing Tab/Enter
first and without a full tmux attach.

Also extends live-send itself to support `ViewMode::Tool` panes (previously
excluded), adding `ensure_tool_pane_ready` to spin up the tool's tmux pane on
demand, mirroring the existing terminal-pane path.

Tradeoff: live-send only forwards keyboard input, so mouse-driven tool UIs
(lazygit, yazi) lose mouse interaction while live-sent. Documented, not fixed here.

## PR Type
- [x] New Feature

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] Skipped a test, because: added 5 focused unit tests in `src/tui/home/tests.rs`
covering the setting on/off × attach-mode gating and Tool-view live-send targeting,
plus manually smoke-tested end-to-end against a real tmux session (positive and
negative cases both confirmed). The behavior is internal state-transition logic
better suited to unit-level coverage than the e2e harness.

## AI Usage
- [x] AI was used

**AI Model/Tool used:** Claude (Sonnet 5) via OpenCode multi-agent pipeline (plan/build/QA sub-agents), orchestrated interactively with human review at each phase.

**Any Additional AI Details you'd like to share:** N/A

## Summary

- Added an opt-in `live_send_on_view_switch` setting to automatically start **live-send** when switching into **Terminal** or **Tool** view, **independent of Default Attach Mode**.
- Expanded live-send support to **Tool panes** by lazily preparing tool panes on demand (`ensure_tool_pane_ready`) and correctly targeting tool-backed tmux sessions for drift checks, sending, and status labeling.
- Improved live-send behavior across view-change entry points (tool picker submit, tool hotkeys, toggle-to-terminal, command palette tool entry), with dedicated logic to decide when to auto-enter live-send on view switch.
- Made the Tool live-send pane responsive to layout changes by reflowing/resizing when the window layout updates (including a focused regression test for targeted Tool pane resizing).
- Updated documentation and guidance to note that during tool live-send, **keyboard input is forwarded but mouse interaction inside embedded tools (e.g., lazygit/yazi) is not supported**.

## Benefits

- Removes extra navigation steps by starting live sending immediately after switching into Terminal/Tool view.
- Better coverage for embedded, keyboard-driven tool workflows while staying consistent with the live-send UX model.
- More reliable, “set it and forget it” behavior when terminal layouts resize or toggle.

## Potential areas for further enhancement

- Add or better emulate **mouse support** within tool panes during live-send to match full tmux attach behavior more closely.

### DIFF
--- a/docs/guides/live-mode.md
+++ b/docs/guides/live-mode.md
@@ -36,7 +36,9 @@ when the default attach mode is Tmux.
 Live-send only forwards keyboard input, not mouse events, so
 mouse-driven tool UIs like lazygit and yazi lose mouse interaction
 while you're live-sent to them; use their keyboard bindings instead, or
-exit live mode to click within a full tmux attach.
+exit live mode (`Ctrl+Q`) and then attach normally (`Enter`, depending
+on your Default Attach Mode) to get a full tmux attach with mouse
+support.
 
 ## The leader menu
 
--- a/docs/guides/live-mode.md
+++ b/docs/guides/live-mode.md
@@ -27,11 +27,10 @@ Live-send also works for the Terminal view and Tool views (lazygit,
 yazi, and other embedded tools), not just the agent pane: whichever
 pane is on screen is what your keystrokes reach.
 
-With **Default Attach Mode** set to Live mode, turning on **Auto
-Live-Send On View Switch** (Interaction settings, off by default) skips
-the extra `Enter`/`Tab`/click when you switch into Terminal or Tool
-view: live-send starts as soon as the view switches. It has no effect
-when the default attach mode is Tmux.
+Turning on **Auto Live-Send On View Switch** (Interaction settings, off
+by default) skips the extra `Enter`/`Tab`/click when you switch into
+Terminal or Tool view: live-send starts as soon as the view switches,
+regardless of **Default Attach Mode**.
 
 Live-send only forwards keyboard input, not mouse events, so
 mouse-driven tool UIs like lazygit and yazi lose mouse interaction
--- a/docs/guides/live-mode.md
+++ b/docs/guides/live-mode.md
@@ -23,6 +23,21 @@ you work.
 The status bar shows a `● LIVE → <session>` banner while you are
 relayed, including a reminder of the exit chord and the leader menu.
 
+Live-send also works for the Terminal view and Tool views (lazygit,
+yazi, and other embedded tools), not just the agent pane: whichever
+pane is on screen is what your keystrokes reach.
+
+With **Default Attach Mode** set to Live mode, turning on **Auto
+Live-Send On View Switch** (Interaction settings, off by default) skips
+the extra `Enter`/`Tab`/click when you switch into Terminal or Tool
+view: live-send starts as soon as the view switches. It has no effect
+when the default attach mode is Tmux.
+
+Live-send only forwards keyboard input, not mouse events, so
+mouse-driven tool UIs like lazygit and yazi lose mouse interaction
+while you're live-sent to them; use their keyboard bindings instead, or
+exit live mode to click within a full tmux attach.
+
 ## The leader menu
 
 Almost every key you press in live mode goes to the agent, so AoE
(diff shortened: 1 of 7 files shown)


Seluj78 (dismissed):

## Verdict
**Project Fit:** Accept

Real, opt-in (defaults off) feature that removes a genuine friction in live-mode and cleanly extends the existing live-send machinery to a fourth target. Scope is focused and fits the active live-session work on main. One correctness gap to fix before merge; fix is a one-liner.

## Required

**`src/tui/home/render.rs:1888` (outside the PR diff hunks, so noted here rather than inline) — Tool live-send pane never reflows on preview-geometry change.**

`refresh_tool_preview_cache_if_needed` goes straight to `apply_worker_capture`; it omits the `resize_live_pane_if_target(...)` call that the agent (`render.rs:1730`), terminal (`render.rs:1837`), and container (`render.rs:1865`) refresh paths each run as their first line.

Failure scenario: enter live-send on a Tool view (the new capability), then resize the terminal, drag the preview divider, or toggle the info header. The three shell/agent targets reflow their tmux pane to the new visible rect; the tool pane stays pinned at its entry-time geometry, so lazygit/yazi renders mis-fit (wrong width, clipped or wrapped) until live-send is re-entered. Entry geometry is fine (`finalize_live_send_resize` is target-agnostic); only the steady-state reflow is missing.

The sibling comment states the intent verbatim: "keep its tmux pane sized to the visible output area so a window resize or info-header toggle reflows the shell instead of waiting for a live-mode re-enter." Tool is the one live-send target of four missing it.

Fix (first line of the fn):
```rust
self.resize_live_pane_if_target(
    live_send::LiveSendTarget::Tool(tool_name.to_string()),
    width,
    height,
);
```

## Nits / metadata

- `docs/guides/live-mode.md` documents **New Session Attach Mode** and live-send behavior but was not updated for the new **Auto Live-Send On View Switch** setting or Tool-view live-send. The keyboard-only tradeoff (mouse-driven tools lose mouse while live-sent) lives only in the PR body, not the guide. The "Documentation was updated" checkbox is unchecked.
- "For UI changes: included screenshot or recording" is unchecked; this changes TUI behavior on `t` and tool hotkeys, and a recording would likely have surfaced the resize gap above.

## What's clean

- Config uses the single-source `#[setting]` schema correctly (default wired, no separate override struct needed).
- `ensure_tool_pane_ready` mirrors `attach_tool_session` exactly.
- The Tool drift-check fix is sound and regression-tested.
- Dropping `Copy` on `LiveSendTarget` is clean: all by-value sites converted, no stale copy-reliant caller.
- Unit tests cover the setting gating (on/off, attach-mode) and Tool-view targeting.

athal7:
Thanks for the catch — good find. Fixed in 9d947ea4:

- `refresh_tool_preview_cache_if_needed` now calls `resize_live_pane_if_target(LiveSendTarget::Tool(...), width, height)` as its first statement, matching the agent/terminal/container siblings, so a Tool-view live-send pane (lazygit/yazi) reflows on resize/divider-drag/info-header toggle instead of staying stuck at entry-time geometry.
- Added `docs/guides/live-mode.md` coverage for the new "Auto Live-Send On View Switch" setting, Tool-view live-send support, and the keyboard-only/no-mouse-forwarding tradeoff.
- Added a regression test (`refresh_tool_preview_cache_resizes_live_pane_when_targeted`) asserting the resize call fires for a targeted Tool pane.

Didn't add a screenshot/recording since this is a state-transition/geometry fix without new visible chrome — let me know if you'd still like one.

Ready for re-review.

njbrake (approved):

Reviewed in batch triage. Opt-in auto-enter live-send on view switch plus Tool-view support; the drift-check and tool-preview-reflow bugs are fixed with regression tests, auto-start is gated to explicit view-switch call sites. LGTM once heavy CI runs green.

Issue #2734, opened 2026-07-08
TUI: consume the daemon REST/WS as the single source for session lifecycle


jerome-benoit:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine (@jerome-benoit).

## Problem

AoE has two concurrent processes that both mutate `sessions.json` on disk under a file lock: the TUI (`aoe`) and the daemon (`aoe serve`). The web dashboard already lives as a thin client of the daemon's REST + WebSocket surface. The TUI does not: it constructs its own `Storage`, its own `FileWatchService`, and touches the on-disk state directly at 119 sites in `src/tui/`. The daemon has to observe every disk change the TUI makes through a file-watcher instead of receiving it through an API, and cross-process races between the two writers surface as user-visible bugs (see class linked below).

This is asymmetric: the ACP structured-view path already treats the daemon as the sole broker, per [`docs/development/internals/structured-view.md:5`](https://github.com/agent-of-empires/agent-of-empires/blob/main/docs/development/internals/structured-view.md#L5) ("the daemon supervises one detached worker per session and brokers the protocol between the worker and the web/TUI clients"). Session lifecycle (create, rename, archive, trash, restore, purge, snooze, favorite, group, worktree relocation, status detection, tmux env sync) is the remaining half of the state model that still runs through the dual-writer disk-coupled path.

The proposal is to complete the inversion: TUI becomes a thin client of the daemon's session-lifecycle API, exactly the way the web dashboard already is.

## What's there today

Coupling measured against `upstream/main`:

- TUI reads and mutates storage directly at **119 `Storage::new` sites** in `src/tui/`, with `storage.update` / `storage.load` callers layered on top. In [`src/tui/home/mod.rs`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tui/home/mod.rs) alone: six `Storage::new` sites (lines 1886, 2361, 3541, 4325, 4387, 5208), plus five `storage.update` sites at lines 1900, 4940, 5253, 5285, 5435.
- `HomeView` at [`src/tui/home/mod.rs`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tui/home/mod.rs) carries **157 fields** covering session data, UI state, 10+ dialog `Option` fields, search state, three pollers, three preview caches, and two nested views. The struct has grown ~3x since [#417](https://github.com/agent-of-empires/agent-of-empires/issues/417) cited 46.
[...]

Issue #2541, opened 2026-06-28
CLI purge and restore lack mutual exclusion; teardown runs on a stale snapshot


Seluj78:
## Summary

The CLI permanent-purge paths run irreversible teardown (worktree, branch, container, transcript) on an unlocked snapshot, and only the final row removal is taken under the storage lock. #2534 made the final removal revalidate that the row is still trashed, so a concurrent restore is no longer silently deleted, but the destructive teardown has already run against the pre-restore snapshot. A session restored mid-purge can therefore survive as a row that points at deleted artifacts.

## Steps

1. Put a session in trash.
2. Start `aoe rm --purge <id>` or `aoe session empty-trash`.
3. While teardown runs (outside the lock), restore the same session from another process or the UI.
4. The purge keeps the row (post-#2534), but its worktree/branch/container/transcript are already gone.

## Current state

`src/cli/remove.rs` and `src/cli/session.rs::empty_trash` now revalidate `is_trashed()` under the lock and warn the user when a restored row is kept, but they cannot undo the teardown. The restore path (`session restore`) also moves the worktree outside the lock, so it has a symmetric race against purge.

## Expected

Purge and restore should be mutually exclusive over the whole operation, not just the final write. The server side already locks per-instance and revalidates (`src/server/api/sessions.rs::purge_expired_trash`).

## Possible fix

Pre-claim the targeted rows under the lock before teardown. There is already a `Status::Deleting` variant (`src/session/instance.rs`); a durable claim plus a restore-refusal check (and a symmetric claim on the restore path) would close the window. Needs care so a crashed purge does not strand a row in a claimed state.

## Context

Follow-up from the CLI trash/purge correctness batch (#2524, #2525, #2527, #2534). That batch implemented the minimal in-scope revalidation #2534 asked for plus a loud warning; the full mutual-exclusion lease was deliberately deferred here to keep the bug-fix batch from turning into a concurrency redesign.

jerome-benoit:
Looks resolved as of `1c8d94f2` (PR #2780), which is on `main`.

The fix adds a durable, cross-process operation claim (rather than the `Status::Deleting` sketch):

- New `src/session/claim.rs` with four flock-guarded helpers: `decide_purge_claim` / `finalize_purge_removal` and the symmetric `decide_restore_claim` / `finalize_restore_commit`.
[...]

Issue #2534, opened 2026-06-28
CLI purge can delete a session restored after the initial trash snapshot


jerome-benoit:
## Summary

The CLI purge paths snapshot trashed sessions, run irreversible teardown outside the storage lock, then remove rows from the latest storage state by id without revalidating that the row is still trashed. A concurrent restore can therefore be overwritten by `aoe rm --purge` or `aoe session empty-trash`.

## Steps / code path

1. Put a session in Trash.
2. Start `aoe rm --purge <id>` or `aoe session empty-trash`.
3. Before the final `storage.update()` retain step, restore the same session from another process/UI.
4. The CLI purge still removes the row by id from the latest storage state.

## Expected

A concurrent restore should win, or the CLI purge should revalidate under the storage lock and skip/abort if the row is no longer trashed.

## Actual

The restored row can still be deleted after destructive side effects have run.

## Evidence

`src/cli/session.rs::empty_trash()` snapshots trashed rows with `load_with_groups()`, runs `perform_deletion()` and transcript purge on the cloned instances, then does `all_instances.retain(|i| !purged_set.contains(&i.id))` without checking `i.is_trashed()`.

`src/cli/remove.rs::run()` does the same for one session in the `--purge` path: it clones the resolved instance, runs teardown, then does `all_instances.retain(|i| i.id != removed_id)`.

The server retention path shows the safer pattern: `src/server/api/sessions.rs::purge_expired_trash()` takes the per-instance lock and revalidates the row is still expired trash before purging.

## Deduplication

This is distinct from #2527, which is about `empty-trash` reporting attempted purges instead of successful purges. It is also distinct from #2524/#2532, which cover transcript/sidecar cleanup behavior rather than concurrent restore semantics.

Seluj78:
you or me ? 👀

PR #2780, opened 2026-07-11
fix(session): durable cross-process purge claim to close the purge/restore race (#2541)

jerome-benoit:
## Description

Closes the cross-process purge/restore race in #2541.

**The race.** Permanent-purge paths run their irreversible teardown (worktree, branch, container, transcript) on an UNLOCKED snapshot; only the final row removal sits under the storage flock. #2534 revalidated that final removal, but the teardown had already run on the pre-restore snapshot, so a session restored mid-purge survived as a row pointing at destroyed artifacts. **The reported repro is cross-process** ("restore the same session from another process or the UI"): the server's in-memory `instance_lock` is invisible to the CLI, so the only cross-process serialization point is a durable claim written on disk under the flock. A CLI-only fix does not close the reported bug.

**The fix.** A durable, symmetric `op_claim: Option<OpClaim>` on `Instance` (`OpClaim { op: ClaimOp::{Purge|Restore}, at }`), acquired via `try_claim(...)` INSIDE a `storage.update` (under the flock) BEFORE the unlocked slow phase, on BOTH purge and restore. A fresh peer claim of the *other* op is refused; a free / expired / own claim is (re)acquired. The claim/commit **decisions live in one neutral `src/session/claim.rs`** module (`decide_purge_claim`/`finalize_purge_removal` and the symmetric `decide_restore_claim`/`finalize_restore_commit`), consumed identically by the CLI, the serve daemon, and the TUI, so all three surfaces close the same windows.

**Why a dedicated field, not `Status::Deleting`.** The tmux poller (`merge_passive_status_patch`) rewrites `status` unconditionally, and peer diffs (`merge_user_action_diff`) rewrite it on any status change; either would clobber a status-borne claim. `op_claim` is a dedicated field, deliberately excluded from the `merge_user_action_diff` copy set (which is what protects it from the peer clobber).

**Sites covered (all restore/purge surfaces):**
- **CLI**: `rm --purge`, `session empty-trash` (per-row claim so a large batch cannot overrun the TTL), `session restore`.
- **Server**: restore handler and `purge_session_artifacts` (shared by `DELETE` and `purge_expired_trash`), both through one `with_locked_storage` helper; the untrashed-since-snapshot recheck and the #2534 restore-race recheck are applied server-side too.
- **TUI**: restore (driven directly against storage, since the TUI's `merge_user_action_diff` path drops `op_claim`) and the permanent-purge-of-trashed path (`delete_selected` -> deletion poller -> claim-aware finalize), which previously had no #2534 revalidation at all.

Purge-side `clear_op_claim` is ownership-guarded (`op_claim.op == Purge`) so a purge's final commit never clobbers a peer's fresh Restore claim on the stale-override path.

**Self-heal (crash recovery).** `try_claim` treats a claim older than the TTL (`OP_CLAIM_TTL`, 10 min) as absent, so a crash mid-operation is overridden by the next purge/restore; the daemon startup reconcile and the TUI load pass also clear expired claims. Required, not log-only.

**Residual (deferred, with justification).** The normal race is closed by construction (zero residual within the TTL). The only remaining window is stale-override: an operation exceeding the TTL can be overridden; the final-commit recheck degrades that rare case to #2534 behavior (bail / keep+warn / 409), never worse than the status quo.

**Meta-root (out of scope).** The structural cause is the dual-writer model (#2734, open): two processes mutate durable state with only field-level flocks and no exclusion over a multi-phase operation. This symmetric claim is the bounded local instance of that fix; if #2734 makes the daemon the sole writer, the exclusion LOGIC survives and only the durable serde field becomes redundant.

`op_claim` is an additive `Option` (mirrors `trashed_at`), so no `src/migrations/` entry is needed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (code comments only; no user-facing docs affected)
- [ ] For UI changes: included screenshot or recording (N/A: no visual change)

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (only server-side claim serialization was added; the web restore/delete round-trips are unchanged)

**TUI / CLI (e2e test)**
- [x] Added an e2e test under `tests/e2e/purge_restore_race.rs`: uses the real `aoe` binary and injects an on-disk `op_claim` into `sessions.json` to simulate a peer mid-operation, then asserts (a) `session restore` is refused while a fresh Purge claim holds the row and succeeds once it clears, (b) `rm --purge` is refused and keeps the row while a fresh Restore claim holds it, and (c) a clean trashed session purges. Deterministic (10-min claim TTL, no timing/tmux coordination).

The claim primitive and every decision site are additionally unit-tested in `session::claim` (decide/finalize for both purge and restore, symmetry, stale-override, ownership-guard, the commit-time purge-steal bail), plus the untrashed-since-snapshot bail and empty-trash reporting in `cli`, and `perform_deletion` idempotency in `session::deletion`. The genuinely mid-flight `Restored`/`PurgeStoleClaim` transitions require concurrency at the flock (not reproducible by a single static injection) and are covered by those unit tests; a two-process barrier harness for them is a possible future follow-up.

`cargo fmt --check`, `cargo clippy --features serve -- -D warnings` (and non-serve, and the e2e target), `cargo test --features serve`, and `cargo test --features e2e-tests --test e2e purge_restore_race` all pass.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Sisyphus / OhMyOpenCode)

**Any Additional AI Details you'd like to share:** AI-assisted implementation from a decision-complete spec, then hardened over several parallel-reviewer + cross-validation passes (design, correctness, Rust idiom, harmonization, tests). Design and verification were reviewed against the codebase.

## Summary

- Fixed a cross-process race between **purge** and **restore** for trashed sessions by introducing a durable, per-row **operation claim** (“who owns the row”) stored on disk. This prevents two processes from performing conflicting actions at the same time across the CLI, server, and TUI.
- Centralized the purge/restore rules and end-of-operation outcomes into a shared `session::claim` module, then refactored the CLI/server/TUI flows to use it.
- Added safe, symmetric claim release and “finalize under lock” checks so results reflect what actually happened even if another process intervened mid-operation.
- Improved crash recovery with a **10-minute claim expiry**, plus startup self-healing that clears expired claims.
- Refined `empty-trash` and permanent delete/restore handling to track clearer end states (removed vs restored vs kept-for-retry) and ensure per-row claim ownership is handled consistently.
- Added unit/regression tests and a deterministic end-to-end test that simulates the race by injecting an on-disk op-claim, then verifies refusal and later success once the competing claim is cleared.
  
## What moved / was added / removed

- **Added**
  - Durable `Instance.op_claim` (operation + timestamp) with helpers to acquire, expire/override, and release claims safely.
  - A shared `session::claim` module that encapsulates both **claim decisions** and **finalization** behavior for purge vs restore.
  - New `empty-trash` phase-2 finalization logic to compute consistent outcomes and release only the claims the current process owns.
  - TUI state tracking for purge-claimed rows, including startup self-heal for expired claims and cleanup to avoid stuck claims on disconnect.
  - An e2e test module (`purge_restore_race`) that verifies purge/restore mutual exclusion and correct recovery.

- **Removed / refactored**
  - CLI “empty-trash” eligibility logic was replaced by the shared claim module’s rules.
  - Several purge/restore branching decisions were reworked to depend on claim-based finalization outcomes rather than earlier heuristics.

## Benefits

This prevents incorrect interleavings where a restore and a purge could conflict across processes, by making purge vs restore mutually exclusive at the storage/claim level—while still allowing interrupted operations to recover safely after crashes.

There’s a small performance/clarity tradeoff in the TUI’s defense-in-depth startup self-heal: it can perform one `storage.update` per expired claim. The behavior is kept since expired cases are rare and the shared claim logic remains the source of truth.


jerome-benoit:
@coderabbitai Thanks for the review.

On the one nitpick (`src/tui/home/mod.rs`, the load-time self-heal doing one `storage.update` per expired `op_claim`): acknowledged, and intentional as-is.

- Expired claims are strictly the rare crash-recovery case; on the normal startup path no claim has expired, so the guarded write never fires and the pass stays cheap.
- The authoritative self-heal is in `try_claim` itself (a claim `>= OP_CLAIM_TTL` is treated as free), so the startup/load sweeps are deliberately belt-and-suspenders, not on any hot path.
- Folding it into a single all-rows `storage.update` is a fair future optimization, but it only pays off under startup contention we don't have today; pre-optimizing a rare path would add complexity for no measurable gain.

Leaving the current per-row form for clarity. Happy to revisit if startup contention ever shows up in practice.

Seluj78 (approved):

## AoE Review — Approve

Rust-only PR (session, cli, server, tui, e2e). No web/docs/migration surfaces.

**Mechanical gates:** clean. No em dashes, no dead-code allows, no hardcoded namespace, no secret-in-log. All added `tracing::` calls carry a `target:`.

**Migration:** none needed. `op_claim` is an additive `Option` with `skip_serializing_if`, round-trip + default tested, mirrors `trashed_at`.

**Parity:** claim/commit decisions live in one neutral `src/session/claim.rs`, consumed identically by CLI, server, and TUI, so all three surfaces close the same windows.

### Verdict

**Implementation:** Approve
**Project Fit:** Accept

Closes a real reported cross-process bug (#2541) a CLI-only fix could not reach: the server's in-memory lock is invisible to the CLI, so the only cross-process serialization point is the on-disk flock, which is exactly where the claim lands. Race analysis verified end to end:

- Normal race closed by construction. `try_claim` refuses only a fresh claim of the other op; a purge and a restore cannot both hold. Symmetric refusal is unit-tested both directions.
- Stale-override (op exceeds the 10-min TTL) degrades to #2534, never worse. `finalize_purge_removal` re-checks trashed state and `clear_op_claim_if_owned` is ownership-guarded, so a purge's final commit can never clobber a peer's fresh Restore claim; `finalize_restore_commit` bails on `PurgeStoleClaim`. Both guards tested.
- Crash self-heal via TTL in `try_claim` (authoritative) plus belt-and-suspenders sweeps in the daemon reconcile and TUI load pass.
- Dedup done right: the old `purge_restored_row_must_be_kept` / `RowRemoval` in `cli` are removed, not duplicated. No dead code; every new `Instance` method has live consumers.
- TUI correctly bypasses `apply_user_action` because `merge_user_action_diff` deliberately drops `op_claim`, driving storage directly instead. Subtle and correct.

Test coverage is strong: 20+ unit tests on the primitives and every decision site, `perform_deletion` idempotency backing the reacquire path, and a deterministic cross-process e2e injecting an on-disk claim. The genuinely mid-flight transitions that need real flock concurrency are honestly called out as unit-covered with a two-process harness noted as possible follow-up.

Informational (not a finding): the TUI load-time self-heal issues one `storage.update` per expired claim. CodeRabbit raised it; the author justified it as strictly the rare crash-recovery path (normal startup has zero expired claims) and `try_claim` is the authoritative heal regardless. Reasoning holds; no change needed.

No Blocker, Required, Followup, or actionable Nit findings.

Issue #1945, opened 2026-06-04
Cockpit: aoe TUI/child crash-loops every ~10s on this host — no error logged, leaves 0-byte worker logs


jakecelentano:
## Summary
Aoe is in a steady respawn loop on this host — 8487 `[AOE_START_MARKER]` entries in the last 10 min (≈1 every 70ms in bursts, ≈1 every 10s over the long run). The long-lived daemon (PID 1701114, 1d6h) and the original TUI (PID 1493368, 1h36m) are both healthy, but a *third* stream of aoe processes keeps starting, running for ~10s, and exiting cleanly (no error/panic/abort in `debug.log`). Most likely cause is a cockpit agent child misbehaving on `pi` startup; the only logged startup failure is from June 2 (`runner socket ... did not appear within 10s`, see Related).

## Repro
1. Start aoe (1.9.5)
2. `aoe status` → returns fine (`0 waiting • 4 running • 3 idle`)
3. `tail -f ~/.config/agent-of-empires/debug.log` shows `[AOE_START_MARKER] version="1.9.5"` arriving every ~10s
4. `ps -ef | grep aoe` shows a churning set of short-lived aoe PIDs, all children of PID 1701114
5. The cockpit worker at `cockpit-workers/8be8fa51d9664c72.log` is 0 bytes — never written to

## Expected
- A child process that respawns in a tight loop logs *something* — at minimum an error on its way out
- Stale worker artifacts (empty log files) are cleaned up when the worker fails to start

## Actual
- Every respawn logs only `[AOE_START_MARKER] version="1.9.5" pid=…` and `aoe started version="1.9.5" pid=…`. No `error`, `warn`, `panic`, or `abort` lines from the looped process.
- `cockpit-workers/8be8fa51d9664c72.log` (0 bytes, mtime 2026-06-02 17:47) is still on disk. The `.sock` file was cleaned up, but the empty log was not.
- Daemon and primary TUI are unaffected; tmux-direct workers (`aoe_*` sessions) all work normally — it's a noisy but not destructive loop.

## Local evidence
```
$ awk '/AOE_START_MARKER/ { c++ } END { print c }' ~/.config/agent-of-empires/debug.log
(last 10 min window: 8487)
$ tail -5 ~/.config/agent-of-empires/debug.log
2026-06-04T05:39:20.373940Z  INFO log.runtime: aoe started version="1.9.5" pid=1926814
[...]

PR #2782, opened 2026-07-11
fix(acp): treat CapacityFull as a first-class transient (refund budget, retry, no orphan) (#1027)

### LINKED ISSUE #1027: Structured view: capacity-full UX shows wrong remediation, creates orphan session, no slot visibility

## Problem

When the ACP worker cap is full, hitting **+** to create a new structured-view session produces an `AgentStartupError` banner with this message:

> Failed to start cockpit agent "claude": cockpit worker capacity full (5/5); raise [cockpit] max_concurrent_workers or delete an existing cockpit session.

The message accurately describes the cap, but the resulting UX has three rough edges that are surprising once you hit it:

1. **The error banner's remediation hint is wrong for this case.** `StartupErrorBanner` in `web/src/components/cockpit/CockpitView.tsx:538-630` branches on a regex against the message: auth-looking strings get an `ANTHROPIC_API_KEY` hint, everything else gets `aoe cockpit doctor --fix` + `npm install -g @agentclientprotocol/claude-agent-acp`. The capacity-full message falls into the second bucket, so the banner tells the user to reinstall the adapter — which won't help, because the adapter is fine; they just need to raise the cap or free a slot.
2. **The session was already created.** The auto-spawn path in `src/server/api/sessions.rs:804-844` returns `201 CREATED` to the client first, then fires `Supervisor::spawn` from a detached `tokio::spawn`. When that spawn fails on `CapacityFull`, the session is now persisted with `cockpit_mode = true` but no worker, and the 2s reconciler in `src/server/mod.rs:1319-1438` keeps retrying it (each retry re-publishes `AgentStartupError`). The user has an orphan in the list they didn't ask for.
3. **There is no UI surface for "which 5 sessions are using slots?".** `Supervisor::count()` (`src/cockpit/supervisor.rs:796-798`) and the per-session `is_running` exist, but no REST endpoint enumerates the live workers, and the dashboard does not tag rows with "ACP worker live" vs. "structured-view-mode session, worker dormant". So the prescribed remedy — *delete an existing structured-view session* — leaves the user picking blind.

The cap itself (`max_concurrent_workers = 5`, `src/session/config.rs:102-104`) is reasonable; this isn't a "raise the default" ask. The friction is that *when you hit it*, the dashboard tells you the wrong fix, leaves a half-created session behind, and doesn't show you what's using the slots.

## Why distinct from #1017, #1018, #968

- #1017 (bundle `claude-agent-acp`) is the install path — orthogonal; capacity-full happens after install works.
- #1018 (TUI/CLI parity for substrate switching) is about the substrate toggle, not the spawn-capacity failure mode.
- #968 (structured view in TUI) is rendering scope; this is REST + web UI.

## What's there today

- Capacity check: `src/cockpit/supervisor.rs:330-335` — `workers.len() >= max_concurrent_workers` → `CapacityFull { current, limit }`.
- REST mapping: `src/server/api/cockpit.rs:163-165` — `CapacityFull` → `503 SERVICE_UNAVAILABLE` with the error body. Good for an explicit `POST /api/sessions/:id/cockpit/spawn`, but the create-session flow doesn't go through that path; it fires the spawn after returning `201`.
- Auto-spawn surface: `src/server/api/sessions.rs:804-844` and `src/server/mod.rs:1401-1437`. Both publish `AgentStartupError` via `supervisor.publish_startup_error` with the same `Failed to start cockpit agent {agent:?}: {e}` prefix.
- Banner branching: `web/src/components/cockpit/CockpitView.tsx:545` — `/authentic|login|api[_ -]?key/i` regex; no capacity branch.
- Counter exposed only via `Supervisor::count()` for `aoe cockpit doctor` + internal stats. No `GET /api/cockpit/workers` or similar.

## Proposal

Three layered fixes; they're independent and can ship separately.

### A. Banner recognises the capacity case

Add a third branch to `StartupErrorBanner` keyed on `/capacity full/i` (or, cleaner, change the supervisor side so the `AgentStartupError` event carries a typed `kind` and the banner switches on that — see B). For the capacity case the hint should be the actual remedy:

> ACP worker cap reached ({current}/{limit}). Either raise `[cockpit] max_concurrent_workers` in Settings or free a slot by deleting / switching another structured-view session to tmux.

With a deep-link to the Settings panel and (if B lands) a list of live worker sessions to make freeing a slot one click.

### B. Don't auto-spawn into a doomed slot at create time

Pre-check capacity in `create_session` before persisting `cockpit_mode = true`:

- Path A (strict): if `supervisor.count().await >= max_concurrent_workers`, return `503` from `POST /api/sessions` with the same `CapacityFull` body, and the client treats it like the explicit spawn 503 — show the dialog before the session is added to the list.
- Path B (permissive, probably better): allow the session to be created in `cockpit_mode = false` (tmux substrate), surface a one-time toast/banner explaining the fallback, and let the user explicitly call `POST /api/sessions/:id/cockpit/enable` once they've raised the cap or freed a slot. This avoids the "orphan session that the reconciler keeps banging on" loop.

Either way, the reconciler should also stop retrying a session whose last spawn attempt failed with `CapacityFull` until cap or workers change — currently the 2s tick keeps republishing `AgentStartupError` for every failed reconciliation pass.

### C. Make "which sessions are using slots" visible

Add a small REST endpoint:

```
GET /api/cockpit/workers
→ [{ "session_id": "...", "agent": "claude", "started_at": "...", "model": "..." }, ...]
```

Backed by walking `Supervisor::workers` (lock briefly, snapshot fields from `WorkerHandle` + the `spawn_config` cached on it). Render this in the capacity-full banner as a list with "Switch to tmux" / "Delete" buttons per row, so the user can free a slot without leaving the page.

This endpoint also unlocks a settings-panel widget ("3/5 ACP workers active") so the cap doesn't ambush the user at create time.

## Why now

Capacity-full is the one structured-view failure mode where the message *is correct* but the banner's prescription is misleading. The other failure modes the banner handles (auth missing, adapter missing) are both genuinely "go fix this on your machine" — capacity is the only one where the fix is "click a thing in this dashboard" and the dashboard doesn't offer the thing. The friction is concentrated at the moment a new user is most likely to hit it (they enabled the structured view, opened five tabs, and now session #6 is broken in a way that suggests their install is wrong).

## Open questions

- @njbrake — why was the default cap set to 5 specifically? I couldn't find a rationale in commit messages or `docs/cockpit.md`; `default_max_workers()` in `src/session/config.rs:102-104` just returns 5 with no comment. The plausible reasons (per-worker Node + native SDK RAM footprint, parallel LLM-stream rate-limit guardrail, fork-bomb safety net for buggy auto-spawn loops) are all defensible, but knowing which one was the driver matters for whether the right fix here is "make the cap discoverable" (this issue) vs. "make the cap dynamic / scale with host resources" (a different one). If 5 was a placeholder picked without measurement, that's also fine to know — it would push the proposal toward B path B (graceful tmux fallback) rather than a strict 503.
- Should the cap apply to *worker* count or to *structured-view-mode session* count? Current implementation is the former (so disabled/dormant structured-view sessions don't count); that matches the resource model but isn't what the message implies.
- For B path B (graceful fallback to tmux), is there a "remember this is a structured-view session, just dormant" notion the rest of the stack already supports? Looks like yes — `cockpit_mode` on `Instance` plus `is_running` is exactly this — but I haven't traced what happens if a tmux-substrate session has `cockpit_mode = true` recorded; that may already DTRT or may need a small fix.
- Capacity events probably warrant a dedicated event variant rather than a regex on `AgentStartupError.message` — but that's an `Event` enum addition and may want to bundle with B.

## Out of scope

- Raising the default cap (5 seems fine).
- Reworking how `aoe cockpit doctor` reports state.
- Per-session resource caps beyond worker count (CPU/RAM); separate concern.

jerome-benoit:
closes #1027

## Description

At capacity (`[acp] max_concurrent_workers`), structured-view sessions were mishandled at three incoherent spawn sites, none treating `SupervisorError::CapacityFull` as the transient, non-crash, user-actionable condition it is (a slot frees when a peer worker stops):

1. **Reconciler `resume_one`**: `CapacityFull` was swallowed at the join, yet the #1945 respawn budget was already recorded at decision time, so a capacity-blocked session burned its crash budget every tick and parked with a misleading crash-loop banner.
2. **Create-path** (`sessions.rs`): the detached post-201 spawn published a generic `startup_error`, leaving an orphan row with no worker.
3. **`acp_enable`** (`acp.rs`): same detached pattern, same generic error and orphan-at-capacity.

(`prompt_wake_resume` was already correct: gated upstream by `begin_resume`, returned as 503 to `send_prompt`. Untouched.)

**Fix §1** adds `ResumeOutcome::CapacityDeferred`. `resume_one` matches `CapacityFull` before the error type is erased into a string. The join handler then: **pops** the single decision-time budget entry (not `remove`, which would wipe genuine prior-crash history and let a real crash loop escape the park budget); re-arms the retry with **`attempted.remove`** (never `insert`; `attempted` is persistent, so keeping the id there would skip the session forever and kill the self-heal); and publishes the capacity banner **once per transition** via a new `capacity_deferred` gate, reusing `Event::AgentStartupError` (the `CapacityFull` Display already matches the front-end `isCapacity` regex). The marker is cleared on the successful-respawn `SpawnFinished`/`Attached` join arm (the critical site), the `is_running` branch, the live-set retain sweep, and the two explicit-retry reaper loops. This is retry-without-penalty, not a park/hold: the natural per-tick reconciler retry is the self-heal, so no auto-un-park reaper is needed. §1 also repairs the create/enable orphans (the next tick adopts them, does not burn the budget, and respawns once a slot frees).

**Fix §2**: the create-path and `acp_enable` detached spawns publish the `CapacityFull` Display (capacity status) instead of the generic crash-style message, via a shared `structured_spawn_error_message` helper. §2 runs before the reconciler sees the session, so the first reconciler tick emits **one benign duplicate banner** (front-end reducer replaces `startupError`, message idempotent); later ticks are gated. Documented and intentional, no shared flag.

Slot visibility (`GET /api/acp/workers` + `StructuredView` UI) is **split to a fast-follow PR** to keep the web CI (oxfmt/eslint/tsc), Playwright, and the coverage-matrix mandate off this Rust-only diff.

Commits: the core fix, then two review-driven follow-ups (clear the capacity marker on explicit retry + harden tests; de-duplicate the reaper clean-slate reset + tighten docs). No behavior change in the follow-ups beyond the marker clear on explicit retry.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the capacity defer/refund/retry/self-heal behavior is covered deterministically by new reconciler-level tests driving `reconcile_acp_workers` across ticks in-process (a full-binary e2e would need real ACP agents held at capacity). Added: `capacity_deferred_rearms_attempted_and_retries_next_tick`, `capacity_deferred_publishes_once_across_ticks`, `capacity_deferred_pop_preserves_prior_crash_history`, `capacity_deferred_clears_marker_when_slot_frees`, `capacity_deferred_cleared_by_is_running_branch`, and `structured_spawn_error_message_prefers_capacity_display_over_generic` (the shared create/enable message selection).

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude (Opus)

**Any Additional AI Details you'd like to share:** Implementation driven from an RCA + cross-validated spec; author reviewed every change against the verbatim code (enum, join arm, budget gate, `begin_resume` capacity gate) before committing. The change was then reviewed over two rounds of three independent parallel reviews each, with the findings cross-validated and applied.

## Summary

- Reworks structured-view session handling when worker capacity is full: it treats capacity-full events as a temporary deferral (not a crash loop).
- Adds a dedicated “capacity deferred” path that refunds the appropriate respawn budget entry, re-arms reconciliation for the next tick, retries later (instead of parking incorrectly), and publishes the capacity status only once per transition.
- Introduces tracking/clearing of the deferred marker so sessions resume normally after recovery or explicit relevant state changes.
- Updates structured-view spawn failure messaging for both **create** and **acp_enable** flows to use the exact capacity-full error text (instead of a generic “failed to start” banner), improving orphan-session adoption/retry when capacity becomes available.
- Adds reconciler-level tests for budget handling, retry behavior, notification/banner deduplication, deferred-marker clearing, and correct spawn-error message selection.
- Adds test utilities to simulate occupying and freeing supervisor worker capacity.
- Defers slot-visibility/dashboard changes to a follow-up PR.

### Benefits

Capacity-blocked structured-view sessions are automatically retried without incorrect crash-loop behavior, and users see capacity-appropriate status messaging (with fewer duplicate notifications). Orphaned sessions are more likely to be adopted and resumed when worker slots free up.

### DIFF
--- a/src/acp/supervisor.rs
+++ b/src/acp/supervisor.rs
@@ -2675,6 +2675,32 @@ impl<S: BroadcastSink> Supervisor<S> {
         self.workers.lock().await.len()
     }
 
+    /// Insert a fake in-memory worker so a test can occupy a capacity slot
+    /// without launching a real runner. Mirrors the `WorkerHandle` fixture in
+    /// `capacity_full_returns_after_limit`. The slot is counted by
+    /// `begin_resume` and `is_running`, but no registry entry is written, so
+    /// the reconciler's orphan sweep never touches it.
+    #[cfg(test)]
+    pub(crate) async fn test_insert_worker(&self, session_id: &str) {
+        let (client, _tx) = AcpClient::fake_for_test(AcpSessionId(format!("acp-{session_id}")));
+        self.workers.lock().await.insert(
+            session_id.to_string(),
+            WorkerHandle {
+                client: Arc::new(client),
+                drain_task: tokio::spawn(async {}),
+                restart_history: vec![],
+                kind: WorkerKind::Stdio,
+            },
+        );
+    }
+
+    /// Drop a fake in-memory worker inserted by `test_insert_worker`, freeing
+    /// the capacity slot for the next reconciler tick.
+    #[cfg(test)]
+    pub(crate) async fn test_remove_worker(&self, session_id: &str) {
+        self.workers.lock().await.remove(session_id);
+    }
+
     /// Reap workers whose on-disk registry entry has disappeared while
     /// the in-memory `WorkerHandle` is still installed. This is the
     /// out-of-band stop signal: `aoe acp stop|kill|restart` (a
--- a/src/server/acp_reconciler.rs
+++ b/src/server/acp_reconciler.rs
@@ -108,6 +108,15 @@ enum ResumeOutcome {
     /// populated; a permanently-failing spawn (e.g. missing
     /// claude-agent-acp) does not loop forever.
     SpawnFinished,
+    /// Spawn refused by `SupervisorError::CapacityFull`: a transient,
+    /// non-crash, user-actionable condition (a slot frees when a peer
+    /// worker stops), not a spawn failure. The join handler refunds this
+    /// tick's respawn-budget entry, re-arms the retry (`attempted.remove`,
+    /// never insert), and publishes the capacity banner once per transition
+    /// via the `capacity_deferred` gate. The per-tick reconciler retry is
+    /// the self-heal. `message` is the `CapacityFull` Display, reused as the
+    /// `AgentStartupError` body so it matches the front-end regex. See #1027.
+    CapacityDeferred { message: String },
 }
 
 /// A single structured view session that needs a worker. Snapshotted from the
@@ -152,6 +161,7 @@ pub async fn reconcile_acp_workers(
     last_rate_limit_reap: &mut Option<std::time::Instant>,
     respawn_history: &mut HashMap<String, Vec<Instant>>,
     parked: &mut HashSet<String>,
+    capacity_deferred: &mut HashSet<String>,
 ) {
     // Respawn build-stale workers that were adopted to drain an in-flight
     // turn (see #1754) and have since gone idle. Runs BEFORE
@@ -261,6 +271,7 @@ pub async fn reconcile_acp_workers(
     // don't grow unbounded and a recreated id starts with a clean budget.
     parked.retain(|id| live.contains(id));
     respawn_history.retain(|id, _| live.contains(id));
+    capacity_deferred.retain(|id| live.contains(id));
 
     // ORDERING INVARIANT: this orphan sweep MUST run before the
     // resume scheduling pass below. The capacity check counts both
@@ -305,6 +316,7 @@ pub async fn reconcile_acp_workers(
             // budget and un-park.
             parked.remove(&id);
             respawn_history.remove(&id);
+            capacity_deferred.remove(&id);
             attempted.insert(id);
             continue;
         }
@@ -449,7 +461,43 @@ pub async fn reconcile_acp_workers(
                     attempted.remove(&id);
                 }
             }
-            Ok((_, ResumeOutcome::Attached)) | Ok((_, ResumeOutcome::SpawnFinished)) => {}
+            Ok((id, ResumeOutcome::CapacityDeferred { message })) => {
+                // Refund the single budget entry this tick recorded at the
+                // decision gate (`record_and_check_respawn_budget`, above):
+                // CapacityFull is not a crash, so it must not burn the #1945
+                // budget. POP the last entry (this tick's), not `remove(&id)`,
+                // which would wipe genuine prior-crash history and let a
+                // crashing session escape the park budget.
+                if let Some(entries) = respawn_history.get_mut(&id) {
+                    entries.pop();
+                    if entries.is_empty() {
+                        respawn_history.remove(&id);
+                    }
+                }
+                // Re-arm the retry so the next tick can try again once a slot
+                // frees. NEVER `attempted.insert`: `attempted` is persistent
+                // (only the live-set retain drops it), so keeping the id there
+                // would skip the session forever and the self-heal would never
+                // fire.
+                attempted.remove(&id);
+                // Publish the capacity banner once per transition; the gate
+                // returns true only on the first insert (mirrors
+                // `parked.insert`), because `publish_startup_error` does not
+                // dedup and per-tick publishing would spam the event store.
+                if capacity_deferred.insert(id.clone()) {
+                    state.acp_supervisor.publish_startup_error(&id, message);
+                }
+            }
+            Ok((id, ResumeOutcome::Attached)) | Ok((id, ResumeOutcome::SpawnFinished)) => {
+                // Clear the capacity marker on the successful-respawn path.
+                // This is the ONLY clear the reconciler-dispatched capacity
+                // case reaches: a successful respawn returns SpawnFinished and
+                // leaves the id in `attempted`, so the `is_running` branch is
+                // unreachable next tick. Without this clear the marker sticks
+                // for the worker's life and a second capacity episode would not
+                // re-publish the banner.
+                capacity_deferred.remove(&id);
+            }
             Err(e) => {
                 // Task panicked or was cancelled. Don't keep retrying
                 // the same id every tick if the task panics on every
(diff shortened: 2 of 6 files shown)


Seluj78 (approved):

## AoE Review

**Implementation: Approve. Project Fit: Accept.**

Rust-only diff (6 files, +472/-14). Traced the full `capacity_deferred` state machine at head `026fca12`:

- **Budget refund** is correct: the decision gate pushes this tick's timestamp; the `CapacityDeferred` join arm `pop`s exactly that (last) entry and preserves prior-crash history. One id per tick, `respawn_history` is a local `&mut`, so no concurrent mutation between push and pop.
- **Marker clear coverage** is complete: set once per transition, cleared on all five sites (`SpawnFinished`/`Attached`, `is_running`, `retain(live)`, both reaper loops via `forget_session_budget`). Archived/snoozed/deleted sessions drop from `live` and get swept, so no stuck-marker path.
- **No banner spam**: the `capacity_deferred.insert` gate fronts `publish_startup_error`; re-defer ticks return false. Covered by `capacity_deferred_publishes_once_across_ticks`.
- **Self-heal**: `attempted.remove` on defer means every 2s tick re-attempts, and `begin_resume` returns `CapacityFull` before `spawn_inner`, so the retry is cheap.

Mechanical gates clean: no em dash, no dead code (`test_insert_worker`/`test_remove_worker` are `#[cfg(test)]` and both used), no untargeted `tracing::`, no hardcoded namespace strings, no secrets logged. The two new `warn!` reuse existing targets. Test-only `set_var("HOME")` is `unsafe` + `#[serial]` + tempdir-isolated, which is fine.

Six deterministic reconciler-level tests cover rearm, publish-once, pop-preserves-history, clear-on-slot-free, clear-by-`is_running`, and message selection. Rust-only, so the coverage matrix is N/A.

### Nits (non-blocking)

- No `Closes #1027` keyword in the body; on squash-merge the issue will not auto-close. Add one.
- Branch `issue-1027` does not follow the `fix/...` convention. Cosmetic; squash-merge drops it.

Real bug with a documented RCA, tightly scoped, slot-visibility UI correctly split to a fast-follow. State machine is sound. Nothing blocks merge. Confirm CI green first (checks still running at review time).

njbrake (approved):

Reviewed in batch triage. CapacityFull reclassified as a self-healing transient; single-entry pop refund (no double-count), no worker orphaned, six reconciler tests. LGTM once the re-running CI is green.

PR #2785, opened 2026-07-12
fix(config): follow symlinks when saving the global config

### LINKED ISSUE #2784: feat(config): Preserve symlinked global config file
I store my global `$HOME/.agent-of-empires/config.toml` file in a separate repo & then symlink it, but it seems that every time I quit `aoe` and then restart it, the symlinked file gets overwritten. 

It would be great if the symlinked version could persist & updates/changes are written to the symlink destination instead of re-writing a "new" config file in the AoE directory. 

I see there is the option to set `$XDG_CONFIG_HOME`, but that affects a lot more than just AoE. 

Please let me know if this is possible or maybe a better path would be to provide a `--config` flag to let me define the config file location.

Seluj78:

## Description

Saving the global config used to clobber a symlinked `~/.agent-of-empires/config.toml`. `save_config` went through the plain `atomic_write` helper, whose temp-file + `rename(2)` dance replaces the symlink itself with a fresh regular file. For anyone who keeps their config in a dotfiles repo (chezmoi, stow, dotbot) and symlinks it into place, every `aoe` quit/restart broke the link and left the dotfile tree out of sync.

The codebase already had the fix, built for exactly this case: `atomic_write_following_symlinks` resolves the symlink chain and writes through to the underlying target, and the hooks code already uses it for agent config files (codex, settl, claude). The global AoE config was simply never switched over. This PR points `save_config` at that helper.

Non-symlinked configs are unaffected: `resolve_symlink_chain` returns the path unchanged when it is not a symlink, so the save is the same atomic replace as before.

Added a Unix-gated regression test that symlinks `config.toml` into a stand-in dotfiles dir, saves, and asserts the link survives and its target holds the new content. Verified it fails on the pre-fix tree (link replaced by a regular file) and passes after.

Closes #2784

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Move `~/.agent-of-empires/config.toml` into a separate dir and symlink it back (`ln -s ~/dotfiles/aoe-config.toml ~/.agent-of-empires/config.toml`).
- [ ] Launch `aoe`, change a setting (or just quit and restart so a save fires).
- [ ] Confirm `~/.agent-of-empires/config.toml` is still a symlink (`ls -l`), not a regular file.
- [ ] Confirm the change landed in the symlink target (`~/dotfiles/aoe-config.toml`).

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (scope to the global config, reuse the existing symlink-following helper rather than add a `--config` flag) and reviewed the commit.

**NOTE:**

## Summary

- Updated global configuration saving to preserve symlinked `config.toml` files by writing changes to the symlink target instead of replacing the symlink.
- Added a Unix-only regression test that creates a symlinked `config.toml`, runs `save_config`, verifies the path is still a symlink, and confirms the TOML content was updated at the target.
- Left the behavior unchanged for non-symlinked configuration files.

## Benefits

- Users can safely manage global config via symlinks (e.g., from dotfiles repositories) without the symlink being broken or replaced during saves.

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -2444,7 +2444,10 @@ pub fn save_config(config: &Config) -> Result<()> {
     let path = config_path()?;
     let table = toml::Table::try_from(config)?;
     let content = toml::to_string_pretty(&table)?;
-    super::atomic_write(&path, content.as_bytes())?;
+    // Follow a symlink through to its target so a config file linked from a
+    // dotfiles repo (chezmoi, stow, dotbot) is updated in place rather than
+    // replaced by a fresh regular file, which would break the link (#2784).
+    super::atomic_write_following_symlinks(&path, content.as_bytes())?;
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


src/session/config.rs
```diff
@@ -2444,7 +2444,10 @@ pub fn save_config(config: &Config) -> Result<()> {
     let path = config_path()?;
     let table = toml::Table::try_from(config)?;
     let content = toml::to_string_pretty(&table)?;
-    super::atomic_write(&path, content.as_bytes())?;
+    // Follow a symlink through to its target so a config file linked from a
+    // dotfiles repo (chezmoi, stow, dotbot) is updated in place rather than
+    // replaced by a fresh regular file, which would break the link (#2784).
```
jerome-benoit: That comment is just noise and is not necessary

jerome-benoit (changes requested):
Strip unneeded comments

src/session/config.rs
Seluj78: Stripped, thanks. `atomic_write_following_symlinks` already documents the dotfiles rationale, so the call-site comment was redundant. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2785/commits/50919cda3f2c73e611969182eb7b44511d919bd

jerome-benoit (approved):

njbrake (approved):

Reviewed in batch triage. save_config now uses atomic_write_following_symlinks, resolving the symlink target before the atomic rename; broken/relative/dangling cases handled, regression test included. LGTM.

PR #2807, opened 2026-07-13
fix(acp): anchor tool-group card id to first child so it stops re-folding

### LINKED ISSUE #2802: Web structured view: active tool-group card re-folds on every new action (unstable group part id remounts the card)

## Problem

In the web structured view, a run of consecutive tool calls collapses into one "N actions" group card (`ToolGroupCard`, #1057). While the agent is still working, that group keeps accreting new tool calls. If you expand the group to watch what the agent is doing live, the card re-folds itself the instant the next action lands. Expand again, next action folds it again. During an active investigation phase the group is unwatchable: it snaps shut on every append.

The same re-fold hits the consecutive-TodoWrite group card (`TodoGroupCard`, #1468).

Only the currently-growing (tail) group is affected. Once the agent emits text and the run is flushed, that group's identity freezes and a manual expand sticks. So the bug is scoped to exactly the group you most want to watch: the one still running.

## Root cause

The re-fold is a remount side effect, not an explicit collapse. There is no `useEffect` that collapses the group.

1. The group's expand state is component-local. `ToolGroupCard` derives open/closed from `useToolCardExpansion(status, false)`, and the only persistent memory of "the user opened this" is a component-local `useState` override.

   `web/src/components/acp/ToolCards.tsx:1395`
   ```ts
   const [open, setOpen] = useToolCardExpansion(status, false);
   ```
   `web/src/components/acp/ToolCards.tsx:376-384`
   ```ts
   function useToolCardExpansion(status: Status, defaultOpen = false) {
     const density = useToolDisplayMode();
     const baseline = status === "err" ? true : density === "compact" ? false : defaultOpen;
     const [override, setOverride] = useState<{ density: ToolDensity; open: boolean } | null>(null);
     const active = override && override.density === density ? override.open : null;
     const open = active ?? baseline;
   ```

2. The synthetic group part's React identity is the concatenation of every child tool-call id. Appending an action mutates `childIds`, which mutates the `toolCallId` key.

   `web/src/components/acp/AcpRuntime.tsx:864-870`
   ```ts
   const { childIds, children } = buildGroupChildren(run);
   out.push({
     type: "tool-call",
     toolCallId: `group-${childIds.join("-")}`,
     toolName: TOOL_GROUP_NAME,
     argsText: JSON.stringify({ children }),
   });
   ```
   Same shape for the todo group at `web/src/components/acp/AcpRuntime.tsx:839` (`todogroup-${childIds.join("-")}`).

3. assistant-ui keys/reconciles tool-call parts by `toolCallId`. `group-a-b-c` becoming `group-a-b-c-d` is a new key, so React unmounts the old `ToolGroupCard` and mounts a fresh one. The fresh instance starts with `override = null`, so `open` falls back to the `baseline`, which is `false` for the group. Collapsed.

Individual tool cards do not have this problem: their key is their own stable `toolCallId` (`StructuredView.tsx:839`), so their expand override survives across streaming frames. The asymmetry is entirely the composite group key.

## Proposal

**A (minimum fix): stable group identity keyed by the run anchor.** Key the synthetic group part by its first child id instead of the full concatenation. The first child id is unique per run (runs are bounded by non-tool-call parts) and does not change as the run grows, so the card no longer remounts on append and the user's expand sticks.

`web/src/components/acp/AcpRuntime.tsx:867`
```ts
toolCallId: `group-${childIds[0]}`,
```
`web/src/components/acp/AcpRuntime.tsx:839`
```ts
toolCallId: `todogroup-${childIds[0]}`,
```

The `argsText` payload still changes on every append (it carries the children JSON), so the card re-renders with the new items; only the identity is now stable, which is exactly what preserves the `useState` override across appends.

One thing to confirm in review: two adjacent same-anchor collisions cannot happen because a flushed run and its successor start with different first children, but the implementer should assert group keys stay unique across a multi-run transcript in the test.

**B (structural, only if A proves insufficient): lift expand state out of the card.** Track expanded group ids in a `Set<string>` in the runtime/`AcpChrome` state keyed by a stable group id, so the expand survives even a full remount. Heavier; A should be enough because the identity fix removes the remount entirely.

## Why now

- Watching a live investigation phase is a core structured-view use, and this makes the group card actively hostile to that exact use.
- The fix is one token per group builder; the risk surface is tiny.
- Pairs cleanly with the existing group-card work (#1057, #1468, #1467, #1767); it is the last obvious rough edge on the running-group card.

- As a user watching a live agent, when I expand the active "N actions" group and the agent runs another tool, then the group stays expanded and shows the new action (it does not re-fold).
  - Suggested test layer: Vitest + RTL in `web/src/components/acp/` (render a group, toggle open, re-render with an extra child via the same run anchor, assert the body is still visible). May also anchor a `web/tests/live/` structured-view check.
- As a user, when consecutive TodoWrite snapshots fold into a TodoGroupCard and a new snapshot arrives, then my expanded/collapsed choice on that group is preserved.
  - Suggested test layer: Vitest + RTL alongside the existing `ToolCards.render.test.tsx` todo-group cases.
- As a developer, when `collapseToolRuns` emits group parts for a transcript with multiple separated runs, then each group part has a unique, stable `toolCallId` that does not change when a run grows.
  - Suggested test layer: Rust-free unit test against `collapseToolRuns` / `buildGroupChildren` in `web/src/components/acp/AcpRuntime.test.ts`.

## Open questions

- Anchor by first child id vs a run-scoped counter: first child id is simplest and stable, but if the first child of a run is ever re-emitted with a new id on replay, a run counter would be more robust. Which does the replay path guarantee?
- Should the active (still-growing) group default to expanded while running, rather than only preserving a manual expand? Out of scope for the fix, but worth a decision since the annoyance partly comes from it starting collapsed.

## Out of scope

- Changing the default collapsed/expanded state of settled groups.
- Any change to the grouping boundary logic (`TOOL_GROUP_MIN_RUN`, subagent/todo exclusion).
- The generic per-card expand behavior, which already works via the density-scoped override.

## Cross-references

- #1057 — introduced the collapsible actions group this bug lives in.
- #1468 — the TodoWrite group card, which shares the same unstable-key shape.
- #1467 / #1628 — prior expand-state fix (failed cards foldable after auto-open); same hook, different trigger.
- #1767 — density-scoped override baseline that this fix relies on to persist.
- #1102 — group header err rollup change; same card.

Seluj78:

## Description

In the web structured view, a run of consecutive tool calls folds into one "N actions" group card. While the agent is still working, that group keeps accreting tool calls. If you expanded the running group to watch it live, it re-folded itself the instant the next action landed, so the group you most want to watch was unwatchable.

The re-fold was a remount side effect, not an explicit collapse. The synthetic group part's React identity was `group-${childIds.join("-")}`, the join of every child tool-call id. Appending a child changed that id, so assistant-ui remounted the card, and its component-local `useState` expand override reset to the collapsed baseline. Individual tool cards never had this problem because their key is their own stable id.

This keys the group part to its first child (`group-${childIds[0]}`, and `todogroup-${childIds[0]}` for the TodoWrite group) instead of the full join. The first child id is unique per run (runs are bounded by non-tool-call parts) and stable across replay, so the id no longer changes as the run grows. The card survives appends and a manual expand sticks. The `argsText` payload still changes on each append, so the card re-renders with the new items; only the reconciliation key is now stable. Nothing parses the id back into children (they ride in `argsText`), so this touches only the React key.

Regression tests live in `web/src/components/acp/AcpRuntime.test.ts` (a spec already mapped to this surface in `web/tests/coverage-matrix.json`): the group id is anchored to the first child and stays identical when the run grows from 3 to 4 children, the same for the TodoWrite group, and two text-split runs get distinct ids so neither collides. All three fail on the pre-fix `.join()` id and pass after.

Closes #2802

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open the web structured view (`aoe serve`) and start an agent that runs several tools in a row without talking (an investigation phase), so a "N actions" group card appears.
- [ ] While the agent is still running, expand that active group.
- [ ] Confirm the group stays expanded as new actions stream into it (before the fix it snapped shut on every new action).
- [ ] Trigger a run of 3+ consecutive TodoWrite updates; expand the TodoGroup card and confirm it also stays expanded as more snapshots arrive.
- [ ] After the agent emits text and a new run starts, confirm the earlier settled group keeps its own expanded/collapsed state independently.

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: the bug lives in the group part's id generation in `collapseToolRuns`, so the regression is proven at that layer with Vitest in `web/src/components/acp/AcpRuntime.test.ts` (already mapped to this surface in `coverage-matrix.json`). The re-fold is a pure consequence of the id being stable under React keying; a Playwright spec would exercise React's reconciliation rather than our code.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design call (anchor to the first child rather than lift expand state into a Set), reviewed the diff, and ran the manual test steps.

## Summary

- Stabilized grouped tool-call and TodoWrite card identities by anchoring them to the first child action.
- Preserved users’ expanded or collapsed state as new actions are added.
- Added regression tests for growing groups, TodoWrite groups, and separate runs with distinct identities.

## Benefits

- Prevents cards from unexpectedly re-folding.
- Keeps grouped activity updates consistent while avoiding collisions between separate runs.

### DIFF
--- a/web/src/components/acp/AcpRuntime.test.ts
+++ b/web/src/components/acp/AcpRuntime.test.ts
@@ -209,6 +209,57 @@ describe("activityToThreadMessages; tool-call grouping (#1057)", () => {
     expect(payload.children.map((c: { toolCallId: string }) => c.toolCallId)).toEqual(["td1", "td2", "td3"]);
   });
 
+  it("anchors the generic group id to the first child, stable as the run grows (#2802)", () => {
+    const groupId = (rows: ActivityRow[]) => {
+      const parts = activityToThreadMessages([userRow("go"), ...rows], false).find((m) => m.role === "assistant")!
+        .content as Array<{ type: string; toolName?: string; toolCallId?: string }>;
+      return parts.find((p) => p.type === "tool-call" && p.toolName === TOOL_GROUP_NAME)!.toolCallId;
+    };
+    const three = groupId([toolStart("t1"), toolStart("t2"), toolStart("t3")]);
+    const four = groupId([toolStart("t1"), toolStart("t2"), toolStart("t3"), toolStart("t4")]);
+    // Keyed by the first child, not the join of every child id, so
+    // appending a tool call does not change the id (which would remount
+    // the card and re-collapse it mid-stream).
+    expect(three).toBe("group-t1");
+    expect(four).toBe(three);
+  });
+
+  it("anchors the todo group id to the first child, stable as the run grows (#2802)", () => {
+    const groupId = (rows: ActivityRow[]) => {
+      const parts = activityToThreadMessages([userRow("go"), ...rows], false).find((m) => m.role === "assistant")!
+        .content as Array<{ type: string; toolName?: string; toolCallId?: string }>;
+      return parts.find((p) => p.type === "tool-call" && p.toolName === TODO_GROUP_NAME)!.toolCallId;
+    };
+    const snap = (id: string, status: string) => todoStart(id, [{ content: "a", status }]);
+    const three = groupId([snap("td1", "pending"), snap("td2", "in_progress"), snap("td3", "completed")]);
+    const four = groupId([
+      snap("td1", "pending"),
+      snap("td2", "in_progress"),
+      snap("td3", "completed"),
+      snap("td4", "completed"),
+    ]);
+    expect(three).toBe("todogroup-td1");
+    expect(four).toBe(three);
+  });
+
+  it("gives two text-split runs distinct group ids so neither collides (#2802)", () => {
+    const parts = activityToThreadMessages(
+      [
+        userRow("go"),
+        toolStart("a1"),
+        toolStart("a2"),
+        toolStart("a3"),
+        messageRow("Found it."),
+        toolStart("b1"),
+        toolStart("b2"),
+        toolStart("b3"),
+      ],
+      false,
+    ).find((m) => m.role === "assistant")!.content as Array<{ type: string; toolName?: string; toolCallId?: string }>;
+    const ids = parts.filter((p) => p.type === "tool-call" && p.toolName === TOOL_GROUP_NAME).map((p) => p.toolCallId);
+    expect(ids).toEqual(["group-a1", "group-b1"]);
+  });
+
   it("uses the generic group for todo-shaped runs when todos are disabled", () => {
     const messages = activityToThreadMessages(
       [
--- a/web/src/components/acp/AcpRuntime.tsx
+++ b/web/src/components/acp/AcpRuntime.tsx
@@ -836,7 +836,14 @@ function collapseToolRuns(parts: DraftPart[], todosEnabled: boolean): DraftPart[
         const { childIds, children } = buildGroupChildren(run);
         out.push({
           type: "tool-call",
-          toolCallId: `todogroup-${childIds.join("-")}`,
+          // Anchor the group's React identity to the first child, not the
+          // join of every child id. While the tail run is still growing,
+          // appending a child would otherwise mutate the key, remount the
+          // card, and reset its local expand state to collapsed. The first
+          // child id is unique per run (runs are bounded by non-tool-call
+          // parts) and stable across replay, so the card survives appends
+          // and a user's expand sticks. See #2802.
+          toolCallId: `todogroup-${childIds[0]}`,
           toolName: TODO_GROUP_NAME,
           argsText: JSON.stringify({ children }),
         });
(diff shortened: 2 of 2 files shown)


njbrake (approved):

Reviewed in batch triage. Tool-group card key anchored to the first child, which is stable for append-only streams; three Vitest regressions. Stops the streaming re-fold. LGTM.

PR #2810, opened 2026-07-13
fix(web): gray out current agent in switch-agent modal

### LINKED ISSUE #2803: Web: switch-agent modal should gray out the current agent and only offer switchable backends

## Problem

The web switch-agent modal is supposed to offer only *other* installed ACP backends as handoff targets. On a session that has never been switched it does not: the current agent shows up as a fully selectable radio option, and it can even be preselected. A user can "switch" claude to claude, a pointless no-op handoff.

The fix has two parts, and I want both:

1. Only show agents you can actually switch to.
2. Show the current agent grayed out and disabled (not hidden), so it is obvious which backend you are on and what you are switching away from.

## What's there today

The modal already intends to exclude the current agent. `web/src/components/acp/SwitchAgentModal.tsx:74`:

```ts
const filtered = list.filter((a) => a.name !== currentAgent);
setAgents(filtered);
const preferred = rateLimited ? filtered.find((a) => a.name === PREFERRED_FALLBACK) : undefined;
setSelected(preferred?.name ?? filtered[0]?.name ?? null);
```

That filter is correct *when `currentAgent` is a real agent name*. The existing Vitest test passes `currentAgent="claude"` explicitly and confirms filtering works (`SwitchAgentModal.test.tsx:83-89`, `expect(values).not.toContain("claude")`).

The problem is where `currentAgent` comes from. Both callsites pass the reduced structured-view state:

- `web/src/components/acp/StructuredView.tsx:669` (composer) and `:467` (rate-limit section): `currentAgent={state.agent}`.
- `web/src/components/acp/Composer.tsx:1107`: `currentAgent={currentAgent}` where the prop type is `AcpState["agent"]`.

And `state.agent` is only ever set on one event. In the frontend reducer it initializes to `null` (`web/src/lib/acpTypes.ts:997`) and is assigned exactly once, on `AgentSwitched` (`web/src/lib/acpTypes.ts:1913`):

```ts
const { from, to, reason } = event.AgentSwitched;
next.agent = to;
```

The handshake-hydration path only backfills it from reduced frames, which have the same limitation (`web/src/hooks/useAcpSession.ts:542`, `agent: state.agent ?? hs.agent`).

So for any session that has not been switched yet, `state.agent === null`, the filter becomes `a.name !== null` (excludes nothing), and `/api/acp/agents` returns the whole registry including the current backend (`src/server/api/acp.rs:710-722`).

The backend does know the real agent: `AcpState.agent: AgentName` (`src/acp/state.rs:429`), set at `AcpState::new(session_id, agent, model)` (`src/acp/state.rs:498`). It is never surfaced to the frontend's structured-view state as an initial value; the frontend only learns it after the first switch.

## Root cause

Two layers:

1. **Frontend state gap.** `state.agent` is `null` until the first `AgentSwitched`, so the modal has no current-agent name to filter against on a fresh session. `web/src/lib/acpTypes.ts:997` and `:1913`.
2. **Modal presentation.** Even given a correct `currentAgent`, the modal *hides* it rather than showing it grayed out. `web/src/components/acp/SwitchAgentModal.tsx:74` and the list render at `:196-220`.

## Proposal

**A (required): surface the current agent to the frontend.** Get the real agent name into the reduced structured-view state at session start, not only on `AgentSwitched`. The backend already holds it (`AcpState.agent`). Concrete shapes, pick one during implementation:

- Extend an existing session-start / handshake frame the frontend already reduces to carry the agent name, and set `next.agent` from it in the reducer. This is the smallest surface if such a frame exists.
- Otherwise emit a small startup event carrying the agent name and reduce it the same way `AgentSwitched` is reduced.

This alone fixes "only show available agents", because the existing filter at `SwitchAgentModal.tsx:74` starts working.

**B (required): render the current agent grayed out and disabled instead of hiding it.** Stop filtering it out; instead render it in the list as a dimmed, disabled row that cannot be selected, and never preselect it. In `web/src/components/acp/SwitchAgentModal.tsx`:

- Drop the `filter((a) => a.name !== currentAgent)` at `:74`; keep the full list.
- In the map at `:197-219`, when `a.name === currentAgent`, disable the radio, dim the row, and add a `(current)` marker.
- Ensure the preselect logic at `:79-80` skips the current agent so the default selection is always a switchable target.
- Keep the confirm button disabled when the only remaining choice is the current agent (already covered by `!selected`, but verify against the new list shape).

## Why now

- Cheap, self-contained UX correctness fix on a flow users hit every manual switch.
- The current behavior lets a user issue a no-op self-handoff, which then runs the whole recap-prefill machinery for nothing.
- The modal's own test already asserts the intended filtering; the app just never feeds it a real `currentAgent`. Closing the gap makes the shipped behavior match the tested behavior.

- As a user on a structured-view session that has never been switched, when I open the switch-agent modal, then the current agent appears grayed out and disabled and cannot be selected.
  - Suggested test layer: `web/tests/live/` (needs a real session so `state.agent` is populated end-to-end), with a Vitest + RTL companion in `SwitchAgentModal.test.tsx` for the disabled-row render given a `currentAgent` prop.
- As a user, when I open the switch-agent modal, then the default-selected agent is never the current agent.
  - Suggested test layer: Vitest + RTL (`web/src/components/acp/SwitchAgentModal.test.tsx`).
- As a user, when I switch from agent A to agent B and reopen the modal, then B is the grayed-out current agent and A is selectable again.
  - Suggested test layer: `web/tests/live/` structured-view round-trip.

## Open questions

- Which existing frame, if any, already reaches the frontend at session start and can carry the agent name (approach A option 1), versus needing a new startup event (option 2)? The reducer only handles `AgentSwitched` today, so this needs a look at the handshake/snapshot frames in `web/src/hooks/useAcpSession.ts` and the backend emit path.
- If the registry has exactly one agent installed (only the current one), the modal should show that single grayed-out row plus the existing "install another backend" empty-state guidance. Confirm the empty/one-agent copy at `SwitchAgentModal.tsx:190-194` still reads correctly.

## Out of scope

- Changing the rate-limit handoff copy or the `codex` preference (`SwitchAgentModal.tsx:41,79`).
- The context-primer recap/prefill flow.
- CLI `aoe acp switch-agent` behavior.

## Cross-references

- #1725 (merged) — feat(cockpit): always-available agent switcher (CLI + web); origin of the manual-switch modal this issue refines.
- #1282 (closed) — the rate-limit "continue in another agent" handoff that first introduced this modal.

Seluj78:

## Description

The web structured-view "switch agent" modal is meant to offer only *other* installed ACP backends as handoff targets. On a session that had never been switched it did not: the current agent appeared as a fully selectable radio option and could even be the default selection, so a user could issue a pointless no-op self-handoff.

Root cause: the modal filtered the current agent with `state.agent`, but the frontend reducer only sets `state.agent` on the `AgentSwitched` event; it initializes to `null`. A never-switched session therefore had `currentAgent = null`, the filter matched nothing, and the running backend leaked into the list. The backend knew the agent all along but never serialized it to the frontend.

This change serializes the session's resolved ACP registry key (`agent_name` when set, else `tool`, the same value `acp_capable` already derives and matching the `/api/acp/agents` names the modal filters against) as a new `acp_agent` field on `SessionResponse`, threads it into the structured view as the `currentAgent` fallback, and renders the current backend as a dimmed, disabled row with a `(current)` marker instead of hiding it. The default selection and the confirm gate now come from the switchable alternatives only. As a side effect it also fixes the modal's "The current agent (unknown)" copy on the rate-limit recovery path.

Approach chosen after a two-model debate over three options: reject an event-log-based approach (pollutes the append-only log with per-reconnect UI state) and a two-line hack reusing `tool` (knowingly wrong for custom agents where `agent_name != tool`), in favor of the serialized-field approach that is fully correct with a contained diff.

Closes #2803

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view session on any ACP backend (e.g. `claude`) with at least one other backend installed (e.g. `codex`).
- [ ] Open the composer "Switch agent" control without switching first. Confirm the current agent appears grayed out with a `(current)` marker and its radio is disabled.
- [ ] Confirm the preselected target is one of the *other* agents, never the current one.
- [ ] With only the current agent installed, confirm the row shows disabled, the "No other structured view agents are registered" hint appears, and the confirm button stays disabled.
- [ ] Trigger the rate-limit recovery banner's "Continue in another agent" flow and confirm the copy reads "The current agent (claude)..." with the real name, not "unknown".
- [ ] Switch from A to B, reopen the modal, and confirm B is now the grayed-out current agent while A is selectable again.

## Checklist

## Test Coverage Analysis

Note: the regression coverage lives in the Vitest + RTL spec `web/src/components/acp/SwitchAgentModal.test.tsx`, which `web/tests/coverage-matrix.json` already maps for the `acp.switch-agent-manual` entry. This is UI-logic (list filtering, disabled-row rendering, default-selection, confirm-gate) best exercised at the RTL layer; a live Playwright would need a second real ACP backend installed on the runner. The matrix validator passes.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction, with a two-model design debate to pick the data-source approach. I made the design calls, reviewed each commit, and ran the manual test steps.

## Summary

- Fixed the switch-agent modal to show the active agent as disabled and marked “(current)” instead of allowing it to be selected.
- Added session-level tracking of the resolved ACP agent so the UI can identify the current agent reliably.
- Improved default selection, confirmation behavior, and empty-state messaging when no alternatives are available.
- Corrected rate-limit recovery messaging to display the current agent’s name.
- Added tests covering current-agent display, selection behavior, and sessions without alternate agents.

## Benefits

- Prevents accidental switching to the already active agent.
- Provides clearer feedback and more reliable behavior before agent events are received.

### DIFF
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -184,6 +184,16 @@ pub struct SessionResponse {
     #[cfg(feature = "serve")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub acp_session_id: Option<String>,
+    /// The session's resolved ACP registry key (`agent_name` when set, else
+    /// `tool`), matching the `name` entries `/api/acp/agents` returns. The
+    /// structured view's switch-agent modal reads this as the current-agent
+    /// fallback before the first `AgentSwitched` event lands (which is the
+    /// only event that populates the reduced `state.agent`), so it can gray
+    /// out the running backend on a never-switched session. Omitted for
+    /// sessions with no resolved agent. See #2803.
+    #[cfg(feature = "serve")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub acp_agent: Option<String>,
     /// True when this session's agent can run a structured ACP `session/fork`:
     /// it is ACP-capable AND declares a real fork strategy. Resume-only ACP
     /// agents (e.g. the bundled `aoe-agent`, which advertises `loadSession` but
@@ -403,6 +413,19 @@ impl SessionResponse {
             },
             #[cfg(feature = "serve")]
             acp_session_id: inst.acp_session_id.clone(),
+            // Resolved the same way as `acp_capable` above: `agent_name` when
+            // set and non-empty, else `tool`. This is the ACP registry key,
+            // so it matches `/api/acp/agents` names the switch-agent modal
+            // filters against. See #2803.
+            #[cfg(feature = "serve")]
+            acp_agent: {
+                let resolved = inst
+                    .agent_name
+                    .as_deref()
+                    .filter(|s| !s.is_empty())
+                    .unwrap_or(inst.tool.as_str());
+                (!resolved.is_empty()).then(|| resolved.to_string())
+            },
             // Shares `agent_is_structured_fork_capable` with the create-time
             // guard so the web "Fork" affordance and server-side acceptance
             // cannot drift: forkable = ACP-capable AND a real fork strategy.
@@ -8728,6 +8751,8 @@ mod workspace_ordering_tests {
             acp_capable: false,
             #[cfg(feature = "serve")]
             acp_session_id: None,
+            #[cfg(feature = "serve")]
+            acp_agent: None,
             #[cfg(feature = "serve")]
             acp_can_fork: false,
             claude_fullscreen: false,
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1582,6 +1582,7 @@ function AppContent({
                         sessionId={activeSessionId!}
                         acpWorkerState={activeSession.acp_worker_state ?? "absent"}
                         tool={activeSession.tool}
+                        acpAgent={activeSession.acp_agent ?? null}
                         archivedAt={activeSession.archived_at ?? null}
                         snoozedUntil={activeSession.snoozed_until ?? null}
                         trashedAt={activeSession.trashed_at ?? null}
--- a/web/src/components/MobileMainPane.tsx
+++ b/web/src/components/MobileMainPane.tsx
@@ -115,6 +115,7 @@ export function MobileMainPane({
                 sessionId={activeSessionId!}
                 acpWorkerState={activeSession.acp_worker_state ?? "absent"}
                 tool={activeSession.tool}
+                acpAgent={activeSession.acp_agent ?? null}
                 archivedAt={activeSession.archived_at ?? null}
                 snoozedUntil={activeSession.snoozed_until ?? null}
                 trashedAt={activeSession.trashed_at ?? null}
--- a/web/src/components/acp/StructuredView.tsx
+++ b/web/src/components/acp/StructuredView.tsx
@@ -77,6 +77,12 @@ interface Props {
    *  / etc.). Resolves the active AgentProfile that drives card
    *  dispatch and claude-specific capability gates. */
   tool: string | null | undefined;
+  /** Session's resolved ACP registry key from `SessionResponse.acp_agent`
+   *  (`agent_name` when set, else `tool`). Used as the switch-agent modal's
+   *  current-agent fallback before the first `AgentSwitched` event lands, so
+   *  the modal can gray out the running backend on a never-switched session.
+   *  See #2803. */
+  acpAgent: string | null;
   /** RFC3339 archived-at timestamp, or null. Drives the
    *  archived-specific "worker stopped" banner that replaces the
    *  generic `aoe acp stop`-style message when the user has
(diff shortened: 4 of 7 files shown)


njbrake (approved):

Reviewed in batch triage. Current agent rendered as a disabled (current) row (native disabled, not just visual); default target and confirm derive from alternatives only, coverage-matrix + Vitest present. LGTM.

PR #2821, opened 2026-07-14
fix(config): read-modify-write config.toml and split app_state into state.toml

athal7:
## Description
Fixes a bug where a long-running `aoe` process (TUI/daemon) could clobber external edits to `config.toml` on its next save, because `save_config` always wrote a stale in-memory `Config` verbatim with no re-read of disk. Config writes now go through a new `update_config` helper, which reloads `config.toml` fresh under a process mutex plus a cross-process flock before applying the caller's change, so unrelated concurrent edits survive.

As part of this, the high-churn `app_state` bookkeeping (welcome/tour flags, dismissed banners, sort/group prefs, tips progress, `web_ui_state`) moves out of `config.toml` into a sibling `state.toml`, since it is machine-owned and freely overwritable rather than user-authored settings. A new migration (`v021`) handles the split transparently on existing installs. `save_config` is removed entirely; all call sites migrated to `update_config`/`update_app_state`. The `app_state` JSON key and wire shape on `GET/PATCH /api/settings` are unchanged.

## PR Type
- [x] Bug Fix
- [x] Refactor

## Checklist
- [ ] For UI changes: included screenshot or recording (N/A, no UI change)

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (server API contract and JSON shape unchanged)

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (updated `tests/e2e/intro.rs` for the state.toml split; new unit/integration coverage for `update_config`/`update_app_state`/migration v021 lives in `src/session/config.rs` and `src/migrations/v021_split_app_state_to_state_toml.rs`)

## AI Usage
- [x] AI was used

**Any Additional AI Details you'd like to share:** Implementation was planned and reviewed by me before merging.

## Summary

- Replaced “rewrite the whole config file” persistence with safer, read/modify/write helpers:
  - `update_config` for `config.toml`
  - `update_app_state` for runtime/UI bookkeeping stored in `state.toml`
- Moved the machine-managed `[app_state]` data out of `config.toml` into a sibling `state.toml`, clarifying what is user-facing settings vs runtime/UI state.
- Added migration **v021** to transparently split existing `[app_state]` from `config.toml` into `state.toml`, with safe behavior when `config.toml` is missing, malformed, or when `state.toml` already exists.
- Updated persistence call sites across server endpoints (`/api/settings` and related), CLI, TUI, plugins, and tests to use the correct helper for each data type (settings vs runtime/UI state).
- Removed the older `save_config`/separate app-state locking path in favor of the unified locked update approach (and updated module exports accordingly).
- Kept the `GET/PATCH /api/settings` JSON shape the same, so `app_state.*` remains available for backward compatibility.
- Updated tests (including concurrency and migration coverage), adjusted the intro e2e test to read first-run flags from `state.toml`, and refreshed documentation to describe the new storage layout and schema boundaries.

## Benefits

- Prevents clobbering unrelated external changes to `config.toml` during concurrent writes.
- Makes upgrades and data ownership clearer by separating long-lived settings (`config.toml`) from runtime/UI state (`state.toml`).
- Reduces the risk of lost updates to `app_state` when multiple processes persist state at the same time.

## Implementation details (for context)

- `update_config` uses a locked read/modify/write flow to preserve external edits; `update_app_state` uses the same locked update primitive to ensure serialized updates to `state.toml` without losing concurrent increments.


jerome-benoit (changes requested):
The runtime states persistence needs to use the already existing race free (almost) code handling sessions.json, etc. files. Your current code is racy for TUI/serve (web ui) users.

njbrake:

Hi @athal7, following up on Jerome's review. I have a small refactor ready that addresses his core point (reuse the existing race-free read-modify-write code for `state.toml` rather than a second hand-rolled flock path):

- `update_app_state` now delegates to `storage::locked_update`, the same serialised read-modify-write primitive `sessions.json` / `groups.json` go through. That drops the parallel implementation (the `app_state_save_lock` mutex, the `.state.lock` sidecar, and the bespoke `acquire_storage_flock` + `AppStateConfig::load` + `save_app_state`) and inherits `locked_update`'s guarantees for free: it writes only when the closure succeeds, and it resolves a symlinked `state.toml`.
- `save_app_state` had no callers outside `update_app_state`, so it and its re-export are removed; its roundtrip test now drives `update_app_state`.
- Also fixes some now-stale prose: the v021 header and the configuration guide still described `state.toml` as a lock-free plain overwrite, which stopped being true once the flock landed, plus a leftover `save_config` reference.

Builds clean; fmt and clippy clean; all config / app_state / migration / concurrency tests pass, including the 16-thread lose-no-updates checks now running through `locked_update`.

Two ways to land it, whichever you prefer:

1. Flip on "Allow edits from maintainers" on this PR and I'll push the commit straight to `fix/config-state-persistence`.
2. Or if you'd rather do it yourself, say the word and I'll paste the full diff here (it's about 30 lines net, mostly deletions).

Thanks!

athal7:
@njbrake go for it!

njbrake:

@jerome-benoit your review was right on both counts, and both halves are now addressed. Since it's pinned to `91c9632`, four commits back, here's what changed:

**The race.** At the commit you reviewed, `update_app_state` held only an intra-process `Mutex`, so the TUI and an `aoe serve` daemon could race on `state.toml` and drop an update. `73188ab` gave it a cross-process flock.

**Reusing the existing code.** That first fix still hand-rolled its own flock path, which was your actual point. `3041783a` routes `update_app_state` through `storage::locked_update`, the same serialised read-modify-write primitive `sessions.json` and `groups.json` go through. That deletes the parallel implementation (the `app_state_save_lock` mutex, the `.state.lock` sidecar, and the bespoke `acquire_storage_flock` plus `AppStateConfig::load` plus `save_app_state`) and picks up two things the hand-rolled version lacked: it writes only when the closure succeeds, and it resolves a symlinked `state.toml`. `save_app_state` had no callers left, so it is gone along with its re-export.

**One deliberate exception, and I'd like your read on it.** `update_config` still has its own flock path rather than going through `locked_update`. Two reasons it does not fit cleanly: its parse step is effectively `Config::load`, which spans two files (it strips `[app_state]` from `config.toml` and reads the real value from `state.toml`), and `locked_update` maps an empty or missing file to `T::default()`, which for `Config` is not equivalent to deserialising an empty table and running `normalize()`. Folding it in would change first-run semantics on an empty `config.toml`, which felt like the wrong risk to take in a PR that exists to fix a data-loss bug. `AppStateConfig` has no such problem: every field is `#[serde(default)]` and it derives `Default`, so the shortcut is behaviour-preserving there. If you think `update_config` should follow anyway, I'm open to it as a follow-up.

Worth flagging separately: the `CI` and `Tests` workflows had been sitting in `action_required` on every commit of this PR, so they had never actually run. I approved them, and the full suite is green on `3041783a` (Clippy, Format, cargo test on linux and macos across all feature shards, Playwright mocked and live, Vitest).

Could you take another look and re-review when you get a chance? Your CHANGES_REQUESTED is the only thing on the merge box now.

src/tui/settings/mod.rs
```diff
@@ -659,7 +659,20 @@ impl SettingsView {
                 if self.current_category() == SettingsCategory::Telemetry {
                     self.global_config.app_state.has_responded_to_telemetry = true;
                 }
-                save_config(&self.global_config)?;
+                let has_responded_to_telemetry =
+                    self.global_config.app_state.has_responded_to_telemetry;
+                update_config(|c| {
+                    *c = self.global_config.clone();
```
jerome-benoit: This closure overwrites the whole config with `self.global_config`, which is a snapshot taken at Settings-open in `SettingsView::new` (`Config::load()`), so it discards the fresh on-disk load that `update_config` just performed under the lock. That is the same "write a stale in-memory Config verbatim" behavior as the removed `save_config`: a global field edited by another process (an `aoe serve` PATCH, a second `aoe`, a direct file edit) while the Settings pane is open is silently reverted on save. `app_state` is stripped and `plugins` is resynced via `resync_after_plugin_mutation`, but the other top-level sections (`theme`, `logging`, `session`, `diff`, `telemetry`, `auth`, `acp`, and so on) are exposed.

The view already knows the user's actual edits (the `baseline_global` snapshot at open vs the current `global_config`), so consider applying only those changed leaves onto the fresh on-disk config instead of a wholesale clone, the same shape `PATCH /api/settings` already uses:

```rust
let current = config_to_json(&self.global_config);
let patch = sparse_diff(&self.baseline_global, &current); // only leaves the user changed
update_config(|c| -> anyhow::Result<()> {
    let mut fresh = serde_json::to_value(&*c)?;
    crate::session::settings_schema::merge_json(&mut fresh, &patch);
    *c = serde_json::from_value(fresh)?;
    Ok(())
})??;
```

This also removes the need for the `plugins` resync and the `app_state` carve-out as special cases: an untouched field simply is not in the patch. One thing to watch is emptied list fields (`environment`, `tools`) needing an explicit removal leaf, the same care the web PATCH path already takes.

docs/guides/configuration.md
```diff
@@ -32,6 +33,24 @@ On macOS, AoE reads from `$XDG_CONFIG_HOME/agent-of-empires/` (e.g. `~/.config/a
...
 ```
 
+### `state.toml`
+
+Sits alongside `config.toml` in the same app dir. It holds global-only
+runtime/UI bookkeeping, such as "seen the welcome tour", the last browse
+directory, sort order, and dismissed-tip/update tracking, none of which is
+a user-facing setting, so it has no profile or repo layer and is never part
+of the settings TUI or the web dashboard's settings schema. `GET
+/api/settings` still exposes these fields under the `app_state.*` key for
```
jerome-benoit, line 43: Worth adding that this exposure is read-only: `PATCH /api/settings` rejects writes to `app_state.*` with a 400, because `AppStateConfig` is not a settings-schema section and `validate_patch_with` treats it as an unknown section. Without that note a reader may try to PATCH an `app_state.*` field and be surprised by the rejection.

jerome-benoit (changes requested):
Thanks for the iteration. Routing `update_app_state` through `storage::locked_update` addresses my earlier raciness concern: `state.toml` now uses the same serialized read-modify-write primitive as `sessions.json`, with 16-thread lost-update tests for both helpers. Migration v021 is solid and tested, both CodeRabbit fixes are applied, and the web `PATCH /api/settings` path is surgical. One blocking item, plus doc fixes.

## Blocking

**Settings Global save still clobbers concurrent external edits** (inline on `src/tui/settings/mod.rs`). `*c = self.global_config.clone()` rewrites the open-time snapshot verbatim and discards the fresh on-disk load, so it behaves like the removed `save_config`: any global field written by another process (an `aoe serve` PATCH, a second `aoe`) while Settings is open is reverted on save. `app_state` and `plugins` are carved out; the other sections are not. It is the only wholesale `update_config` closure in the PR; surgical fix suggested inline.

## Should-fix (docs)

- `docs/development/internals/plugin-system.md:122` and `src/logging.rs:199` still reference the removed `save_config`; both should read `update_config`. (Neither file is in the diff, hence noted here rather than inline.)
- `docs/guides/configuration.md` (inline): note that `PATCH /api/settings` rejects `app_state.*` writes with 400.

## Non-blocking

- `update_config` writes `config.toml` even when the closure returns `Err` (documented at its definition). No current caller is affected (every closure is infallible or errors before mutating), but it is a footgun; consider gating the write on `Ok` like `locked_update`.
- `AppStateConfig`/`Config` drop unknown top-level keys on the round-trip (no `deny_unknown_fields`, no `#[serde(flatten)]`). Not a regression, but the "unrelated concurrent edits survive" docs hold only for fields the running binary knows, and the `config.rs:99-101` comment claiming unknown core keys "fail loudly" is inaccurate.

## Nits

- `config_save_lock()` is a non-reentrant mutex; add a "never call `update_config` within its own closure" invariant note.
- `v021` logs a malformed-config parse failure at `debug!`; `warn!` fits an unusual condition better.
- Stale `save_config` mentions in comments: `tests/e2e/intro.rs:41`, `web/tests/live/tutorial.spec.ts:69`.
- `docs/development/adding-settings.md:123`: the "not `update_config`" caution is easy to miss in the compound sentence.

CI is green across the board.

njbrake:

@jerome-benoit good catch, and you were right to call it blocking. `055045c3` fixes it, along with every should-fix and nit. CI is green.

**The blocking item.** The global Settings save now diffs the open-time `baseline_global` snapshot against the edited config and applies only the changed leaves onto the fresh under-lock load, the shape you sketched. An untouched field is not in the patch, so a peer's write survives.

One deviation from your sketch is worth calling out, because it ended up driving the design: `merge_json` cannot express this. It is insert-only, and the emptied-list case you flagged needs removal semantics. `Config::environment` and `SessionConfig::tools` are `skip_serializing_if`, so emptying one drops the key from the serialized form rather than producing `[]` or `{}`, and removing the key is what lets serde's `default` rebuild it. Writing `null` in its place fails outright, since `environment` deserializes through a `string_or_vec` visitor implementing only `visit_str` and `visit_seq`. So this adds `settings_schema::apply_changed_leaves` next to `merge_json` rather than reusing it. Your emptied-list warning turned out to be the load-bearing constraint.

Both new view-level tests fail against the old wholesale-clone save and pass against the new one, so they reproduce the bug rather than only passing alongside the fix.

**Should-fix and nits, all done:**

- Four stale `save_config` refs corrected (`plugin-system.md`, `logging.rs`, `tests/e2e/intro.rs`, `tutorial.spec.ts`). The last one now names `update_app_state`, since that is what `mark_web_tour_seen` calls.
- `PATCH /api/settings` rejecting `app_state.*` with a 400 is now documented.
- The `Config::plugins` comment is fixed. You were right: there is no `deny_unknown_fields` anywhere, so unknown core keys are silently dropped rather than rejected. The comment now says so, and names it as the one limit on `update_config`'s "unrelated edits survive" contract.
- `config_save_lock`'s non-reentrancy is noted.
- v021 logs an unparseable config at `warn!`.
- The "not `update_config`" caution in `adding-settings.md` is now its own sentence.

**Two I left alone deliberately:**

`update_config` still writes when its closure returns `Err`. Gating it on `Ok` means giving it `locked_update`'s `Result`-returning shape and touching every call site, which is more churn than this PR should carry for something no current caller hits. Happy to do it as a follow-up if you want it.

I kept the `plugins` resync. You are right that the sparse diff removes the need for it as a save-correctness measure, since an untouched `plugins` subtree is not in the patch either way, but it still keeps the in-memory config accurate for display and the dirty flag, so dropping it looked like risk without payoff.

Re-review when you get a chance?

jerome-benoit:
> Re-review when you get a chance?

It looks goods. 

Maybe a refactor as a followup: state et config handling code can be seperated out. And also, file handling requiring concurent access can probably have more code shared.

jerome-benoit (approved):

PR #2839, opened 2026-07-14
fix: stop a dead tool-session pane's Ctrl+C from killing aoe itself

athal7:
## Description

Attaching to a tool-session tmux pane had no readiness check after
`create_with_size`, and `with_raw_mode_disabled` disabled raw mode right
before the attach handoff, which re-enables kernel SIGINT delivery to aoe's
own foreground process group. aoe had no SIGINT handler. Net effect:
pressing Ctrl+C on a dead/hung attached pane (e.g. a remain-on-exit-held
pane from an instantly-failing tool command) killed aoe itself with zero
cleanup, taking down every tmux session/pane aoe managed.

Three-part fix:
- `IgnoreSignalsGuard` (RAII, `src/tui/app.rs`): ignores SIGINT/SIGQUIT for
  the duration of the attach handoff inside `with_raw_mode_disabled`,
  restoring prior disposition on drop. Covers every attach path (tool
  sessions, regular sessions, terminal/container sessions), all of which
  funnel through this one wrapper.
- SIGINT arm in the tokio signal loop, mirroring the existing SIGHUP/SIGTERM
  arms, as a belt-and-suspenders catch for anything outside the guarded
  window (clean shutdown instead of default kill).
- `ToolSession::wait_until_ready()`: bounded poll after `create_with_size`
  that fails fast with the captured pane tail if the pane is already dead,
  instead of attaching to a session that's guaranteed to raise the crash
  scenario above.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (none needed — internal signal-handling fix, no user-facing behavior/docs change)
- [ ] For UI changes: included screenshot or recording (N/A, not a visual change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered instead with a unit test for the signal-disposition guard (`src/tui/app.rs`, ignore/restore round-trip) and two tmux-gated integration-style tests for `wait_until_ready()` (`src/tmux/tool_session.rs`, dead-pane and alive-pane cases) — a full e2e reproduction would require actually delivering SIGINT to the aoe process under tmux, which isn't practical in the current harness.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Root cause was fully specified by the reporting engineer; implementation, tests, and verification (fmt/clippy/full test suite) were done by AI agents under human review before this PR was opened.

## Summary

- Fixes a TUI crash/termination when pressing **Ctrl+C / Ctrl+\\** during tool-session attachment, especially when the target **tmux pane is dead, rapidly exits, or hangs**.
- Adds a Unix-only **RAII signal guard** that temporarily ignores **SIGINT/SIGQUIT** during the sensitive attach handoff, then **restores the exact prior signal behavior** immediately after the handoff completes.
- Extends the TUI’s async signal handling so **SIGINT cleanly exits the TUI** via Tokio (outside the guarded window).
- Prevents child processes (including **tmux** and other external commands) from inheriting the temporarily ignored signals by **resetting SIGINT/SIGQUIT back to defaults before exec/spawn**.
- Makes tmux attachment more reliable with a bounded **pane readiness/liveness check** right after pane creation:
  - waits briefly (polling within a short deadline);
  - if the pane dies early, aborts attachment with an error that includes **captured pane tail output** (up to a fixed small number of lines);
  - otherwise proceeds with attachment.
- Adds **unit tests** and **tmux-gated integration-style tests** that verify:
  - signal dispositions are restored correctly,
  - child-process signal resetting works,
  - and pane readiness behavior for both “exits immediately” and “stays alive” cases.
  - tmux tests are skipped when **tmux isn’t available**.

## Benefits

- Stops Ctrl+C from unintentionally killing **aoe** and any **managed sessions** during attach transitions.
- Turns timing-sensitive attach failures into clearer, actionable errors by including **what the dead tmux pane printed**.
- Reduces attach flakiness by avoiding attempts to attach to **already-dead panes**, with dedicated coverage for the critical edge cases.


njbrake (approved):

Approving. This fixes a real crash with a bad blast radius: one Ctrl+C on a dead tool-session pane took down aoe and every tmux session it was managing, with zero cleanup.

## What I verified rather than took on faith

The `reset_signals_on_exec` half of this rests entirely on one claim, that `SIG_IGN` survives `exec`. I reproduced that in isolation in C instead of trusting the comments:

* **Without** the reset, a child spawned under an ignored SIGINT prints `SURVIVED` and exits 0. The ignore does survive exec, leaving a child nobody can Ctrl+C out of.
* **With** the reset, the child dies with exit 130 (`128+2`).

So CodeRabbit's inheritance finding was real, the fix is load-bearing rather than defensive boilerplate, and the test discriminates rather than passing vacuously. Also confirmed locally: `cargo clippy` clean, and the six new signal/readiness tests pass.

## Design calls I liked

* The guard sits in `with_raw_mode_disabled`, the one choke point that all three `attach()` paths and the editor shell-out funnel through. One guard covers every case instead of a sprinkle of call-site patches.
* It saves and restores the *prior* disposition (including tokio's handler) rather than blindly resetting to `SIG_DFL`.
* `wait_until_ready()` is called only inside the `!exists() || is_pane_dead()` branch, so the 200ms poll never delays reattaching to a live session.

## One commit pushed to this branch

`4ac2c504` removes the leftover `if false && self.prev_sigint.is_some()` block and its `#[allow(clippy::overly_complex_bool_expr)]` in `RestoreSignalsOnDrop::drop`, which was CodeRabbit's last open finding. Pure deletion; `false && x` never evaluated, so runtime behavior is unchanged.

## Non-blocking

* In `with_raw_mode_disabled`, `drop(_signals_guard)` runs before `enable_raw_mode()`, which slightly widens the window where SIGINT is deliverable rather than narrowing it. Harmless now that the tokio arm degrades SIGINT to a clean shutdown, but dropping after `enable_raw_mode()` would close it entirely.
* The `Docs` check is red here, but that is not this PR's doing. `main` is currently red with the identical rustdoc error at `src/session/config.rs:2682`, from a collision between #2842 (which added the deny gate) and #2821 (which added a violating link). A fix PR is on the way; this PR inherits the failure through the merge commit and should go green once main is repaired.

PR #2846, opened 2026-07-14
feat: add tmux socket name setting for session isolation

### LINKED ISSUE #2267: tmux socket isolation option
I use two aoe scopes (work and personal), and I also have a variety of outside-aoe tmux sessions.

Currently, my tmux session list is very cluttered. I'd love to have a way to isolate my sessions for aoe or per aoe profile.  agent-deck does the former, which worked well for me.

I'm thinking it would be an option (tmux socket name; ala `tmux -L`), with the default behavior being the status quo (comingle aoe into the default tmux namespace).

A global setting (ie not per profile) seems the more prudent option, though I haven't dug into the code.

#1960 seems related but doesn't address this specific question.

Not sure if this is a "just me" problem or generally useful and worth addressing. Anybody else use tmux meaningfully outside aoe?

njbrake:
## Description

Closes #2267.

Today aoe runs entirely on the default tmux server; isolation is name-prefix only (`aoe_*`). aoe filters its own views to that prefix, but the isolation is one-directional: aoe ignores your hand-managed sessions, yet your `tmux ls` still lists every one of aoe's. Users who run their own tmux sessions alongside aoe (or split work/personal scopes) want aoe out of their default server entirely.

This adds a global `tmux.socket_name` setting. When set, every tmux invocation runs on a private server via `tmux -L <name>`, so your hand-managed `tmux ls` no longer lists aoe's sessions. Left unset, behavior is unchanged (shared default server, the status quo).

### How it works

Every tmux call in aoe already funnels through a single `tmux_command()` builder (confirmed: no raw `Command::new("tmux")` remains outside the helper). Previously that builder resolved a socket *path* (`-S`) for build/test isolation and the `AOE_TMUX_SOCKET` override. This change generalizes the resolution to yield either:

- a **path** (`-S`) for build/test isolation and `AOE_TMUX_SOCKET`, or
- a **name** (`-L`) for the new user setting, letting tmux own the socket directory (`$TMUX_TMPDIR`, else `/tmp/tmux-<UID>/`) and its `0700` perms.

Precedence: `AOE_TMUX_SOCKET` env, then build/test isolation socket, then `tmux.socket_name`, else the default socket. A configured name containing a path separator is rejected (with a warning) so a stray `/` cannot silently redirect the server; use `AOE_TMUX_SOCKET` for a full path. Resolution is cached at process start (as the env override already was), so a change takes effect on the next aoe start; moving live sessions across tmux servers is not meaningful.

Attach already falls back from `switch-client` to `attach-session` when the former fails, so launching the TUI from inside a foreign (default-server) tmux degrades to a nested attach rather than erroring.

The field is `local_only`, so it surfaces in the TUI settings and the config file but not the web dashboard: repointing the host's tmux server is a host-level, restart-required operation, consistent with other host-execution-surface settings.

## PR Type

- [x] New Feature

## Checklist

## Test Coverage Analysis

Added unit tests in `src/tmux/mod.rs` for the name-to-`-L` mapping, whitespace trimming, empty/unset fallback, and path-separator rejection, plus updated the existing socket-resolution test. `cargo fmt`, `cargo clippy`, and the tmux/config/settings test suites pass.

- [x] N/A: this PR doesn't change a user-facing dashboard flow (the setting is `local_only` and is never rendered in the web dashboard)

**TUI / CLI (e2e test)**
- [x] N/A: covered by unit tests on the socket-resolution logic; no TUI rendering or session-lifecycle change beyond the socket flag

## AI Usage

- [x] AI was used

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -261,6 +261,7 @@ The canonical use case is per-session, repo-scoped, short-lived credentials: min
 status_bar = "auto"
 mouse = "auto"
 clipboard = "auto"
+# socket_name = "aoe"
 ```
 
 | Option | Default | Description |
@@ -268,6 +269,7 @@ clipboard = "auto"
 | `status_bar` | `"auto"` | `"auto"`: apply if no `~/.tmux.conf`; `"enabled"`: always apply; `"disabled"`: never apply |
 | `mouse` | `"auto"` | Same modes as `status_bar`. Controls mouse support in aoe tmux sessions. |
 | `clipboard` | `"auto"` | Same modes. Forwards OSC 52 clipboard escape sequences from the wrapped agent (Claude Code, OpenCode, Codex, etc.) through tmux to your terminal. Without this, "select to copy" inside the agent silently fails. Sets `set-clipboard on` and `allow-passthrough on` on the aoe tmux session. |
+| `socket_name` | unset | Run aoe's sessions on a private tmux server with this socket name (passed as `tmux -L <name>`), so your own `tmux ls` and hand-managed sessions stay separate from aoe's. Leave unset to share the default tmux server (the current behavior). Must be a bare name, not a path; a value with a `/` or `\` is ignored. Takes effect on the next aoe start. Global/profile only. |
 
 ## Diff
 
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -2446,6 +2446,19 @@ pub struct TmuxConfig {
         options = "auto:Auto,enabled:Enabled,disabled:Disabled"
     )]
     pub clipboard: TmuxClipboardMode,
+
+    /// Run aoe's sessions on a private tmux server with this socket name (tmux
+    /// `-L`), so your own `tmux ls` and hand-managed sessions stay separate
+    /// from aoe's. Leave empty to share the default tmux server (the current
+    /// behavior). A bare name, not a path; takes effect on the next aoe start.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[setting(
+        label = "Socket Name",
+        widget = "optional_text",
+        global_only,
+        web = "local_only:changes which tmux server hosts your sessions on this machine"
+    )]
+    pub socket_name: Option<String>,
 }
 
 impl Default for TmuxConfig {
@@ -2454,6 +2467,7 @@ impl Default for TmuxConfig {
             status_bar: TmuxStatusBarMode::Auto,
             mouse: TmuxMouseMode::Auto,
             clipboard: TmuxClipboardMode::Auto,
+            socket_name: None,
         }
     }
 }
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -40,64 +40,127 @@ use std::time::{Duration, Instant};
 /// tmux calls to a known per-test socket instead of relying on `$TMUX`.
 pub const TMUX_SOCKET_ENV: &str = "AOE_TMUX_SOCKET";
 
-/// Resolve the tmux socket path this build talks to, or `None` to use tmux's
-/// default per-user socket. Cached: the process env does not change at runtime.
+/// How aoe points tmux at a specific server, if at all.
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum TmuxSocket {
+    /// A full socket path, passed as `tmux -S <path>`. Used for build/test
+    /// isolation and the `AOE_TMUX_SOCKET` override, where aoe owns the exact
+    /// path.
+    Path(PathBuf),
+    /// A socket name, passed as `tmux -L <name>`. Used for the user-facing
+    /// segmentation setting (#2267); tmux owns the socket directory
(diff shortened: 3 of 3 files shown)


docs/guides/configuration.md
```diff
@@ -261,13 +261,15 @@ The canonical use case is per-session, repo-scoped, short-lived credentials: min
...
 clipboard = "auto"
+# socket_name = "aoe"
 ```
 
 | Option | Default | Description |
 |--------|---------|-------------|
 | `status_bar` | `"auto"` | `"auto"`: apply if no `~/.tmux.conf`; `"enabled"`: always apply; `"disabled"`: never apply |
 | `mouse` | `"auto"` | Same modes as `status_bar`. Controls mouse support in aoe tmux sessions. |
 | `clipboard` | `"auto"` | Same modes. Forwards OSC 52 clipboard escape sequences from the wrapped agent (Claude Code, OpenCode, Codex, etc.) through tmux to your terminal. Without this, "select to copy" inside the agent silently fails. Sets `set-clipboard on` and `allow-passthrough on` on the aoe tmux session. |
+| `socket_name` | unset | Run aoe's sessions on a private tmux server with this socket name (passed as `tmux -L <name>`), so your own `tmux ls` and hand-managed sessions stay separate from aoe's. Leave unset to share the default tmux server (the current behavior). Must be a bare name, not a path; a value with a `/` or `\` is ignored. Takes effect on the next aoe start. Global/profile only. |
```
jerome-benoit, line 272: Two corrections and one addition for this row:

1. "Global/profile only" is inaccurate. `socket_name` is read via `Config::load()` (global `config.toml` only; see `config_path()` + `Config::load` in `src/session/config.rs`), so a value in a profile or repo config is silently ignored at runtime. This should read "Global only", consistent with the `global_only` schema attribute and with the sibling `status_bar`/`mouse`/`clipboard` fields, which also resolve from the global config.
2. Worth noting it takes effect in release builds only: debug/dev builds always isolate onto their own socket, so `cargo run` ignores `socket_name`.
3. Worth noting that existing sessions on the shared server will not be visible to aoe after switching.

jerome-benoit (commented):
Reviewed with a multi-round pass (socket resolution, exhaustive tmux-routing audit, settings-schema wiring, docs). This is a clean, well-scoped feature and #2267 is satisfied. No correctness blocker; the isolation mechanism is sound. A few documentation fixes before merge, plus minor nits.

## Verified (no action)

- Isolation is airtight: the only `Command::new("tmux")` in the tree is inside `tmux_command()` itself; every tmux call across `src/tmux`, `src/session`, `src/server`, `src/process`, `src/tui` goes through the builder, and there is no `sh -c "tmux ..."` or tmux-native (`run-shell` / `set-hook` / `if-shell` / `#(...)`) bypass, so `-L <name>` reaches all tmux state. `-S`/`-L` precede the subcommand. No `kill-server` anywhere; teardown is per-session on the active server.
- Setting wired correctly: `global_only` + `web = "local_only:..."` compose cleanly; `optional_text` + `Option<String>` + `skip_serializing_if` round-trips; no migration needed (new optional field defaults to `None`).
- Security: argv is separated (no shell injection); bare name under the user-owned `0700` tmux dir; rejecting `/` and `\` is sufficient on macOS/Linux.

## Doc fixes (please address)

See the inline comment on `docs/guides/configuration.md`; three items on that one row.

1. It says "Global/profile only", but `socket_name` is resolved via `Config::load()` (global `config.toml` only), so a value in a profile or repo config is silently ignored at runtime. It should read "Global only", consistent with the `global_only` schema attribute and with the sibling `status_bar`/`mouse`/`clipboard` fields, which also resolve from the global config.
2. The setting takes effect in release builds only: debug/dev builds always isolate onto their own socket (`build_isolation_socket()` outranks the config name), so `cargo run` ignores `socket_name` with no signal. Worth a note so contributors testing a dev build are not surprised.
3. Enabling `socket_name` leaves any pre-existing sessions on the shared server invisible to aoe. Worth a half-sentence so users know sessions do not carry over.

## Nits

- `socket_from_config_name`: no max-length guard. The socket lands at `$TMUX_TMPDIR/tmux-<UID>/<name>`, and the full path must fit `sun_path` (~104 macOS / ~108 Linux); a long name fails at bind time with a confusing error instead of the clean warn-and-fallback the function already does for path separators. Consider rejecting over ~80 bytes the same way.
- `.` and `..` pass the filter; they cannot traverse (no separator) and just make tmux fail to bind, but rejecting them alongside separators would be tidier.
- Attach from inside a foreign (default-server) tmux with `socket_name` set: `switch-client` correctly fails cross-server ("no current client") and falls back to `attach-session` on the `-L` server as designed. The fallback does not unset `$TMUX`, so whether it attaches nested or trips tmux's nested-session guard depends on the tmux version; worth a manual check of that path.
- Test coverage: only the pure `socket_from_config_name` helper is unit-tested; the precedence composition in `tmux_socket()` is not (the `cfg(test)` isolation socket short-circuits it). Fine for a PR this size.

Thanks for the clean implementation and the thorough PR description.

PR #2853, opened 2026-07-14
feat(serve): add locked-down CityHall client mode (AOE_CITYHALL_MODE)

### LINKED ISSUE #7: Dumbed-down aoe client for CityHall (composer-first, structured-view only)
A stripped-down aoe client mode for non-technical CityHall end users.

Scope:
- Primary surface is a simple, ChatGPT-like chat UI / composer.
- Structured view only. Remove the terminal view and the diff view.
- New-session wizard: name only. Auto-selects all projects; user cannot add or remove projects.
- Remove most settings; keep a minimal set (e.g. theme).

Lives in the aoe codebase (client), tracked here for the CityHall rollout.

Acceptance: a CityHall user sees only the composer + structured view, creates a session by name, and cannot reach terminal, diff, project management, or advanced settings.

Seluj78:

## Description

Adds a locked-down "CityHall" client mode for non-technical end users, gated behind a new server environment variable `AOE_CITYHALL_MODE`. When set, the web dashboard becomes a composer-first, structured-view-only client.

The flag is resolved once at launch into `AppState.cityhall_mode` (mirroring the existing `read_only` pattern) and surfaced on `/api/about`, so the frontend reacts to it via a small `getClientCapabilities` helper rather than scattered checks. Enforcement is dual-sided, matching the `read_only` precedent: hiding the UI is not enough to satisfy "cannot reach", so the server also rejects the terminal PTY relays, terminal lifecycle, diff endpoints, project create/update/delete, filesystem browse, the advanced-settings `PATCH /api/settings`, and the acp switch-agent / disable routes with a 403 when the mode is active.

What the mode does:

- **Panes:** terminal, diff, and plugin panes are hidden; only the composer + structured (chat) view remain.
- **Sessions:** created by name only. The server derives the rest: a CityHall session spans every configured project (one primary path plus the rest as extra repos), always renders in structured view, and must run an ACP-capable agent. Client-supplied `path` / `extra_repo_paths` / `view` / `scratch` are ignored; creation returns 400 if no project is configured or the agent is not ACP-capable.
- **Project management:** the sidebar Projects section and the homescreen "Clone URL" action are hidden.
- **Settings:** curated to Theme (without color-mode / idle-decay), a Sessions tab reduced to the trash options (`delete_to_trash`, `confirm_delete`, `trash_retention_days`), and the display / consent tabs (MCP servers, Telemetry, Plugins). The profile switcher and all other tabs are removed.
- **Pinned defaults:** worktrees enabled by default, max concurrent ACP workers 50, max concurrent resumes 2, applied server-side on the config-resolution path and hidden from the UI. Worktree path templates stay at their defaults; sensible CityHall paths are TBD with the container work.

This lives in the aoe codebase (the client) and is tracked in the CityHall repo for the rollout.

Closes agent-of-empires/cityhall#7

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start normally (`aoe serve`) and confirm the dashboard is unchanged: terminal, diff, project management, full settings, and the homescreen "Clone URL" action all present.
- [ ] Start with `AOE_CITYHALL_MODE=1 aoe serve` and confirm the startup banner notes CityHall mode is on.
- [ ] In CityHall mode, confirm no diff, terminal, or agents pane toggle is reachable, and the homescreen shows "New session" + "Docs" but no "Clone URL".
- [ ] Open the new-session wizard: it asks only for a name (no project / agent picker, no "More options"), and Launch is enabled without selecting a project.
- [ ] Create a session by name and confirm it opens in the structured view and spans every configured project.
- [ ] Open Settings and confirm only Theme, Sessions (trash options only), MCP servers, Telemetry, and Plugins tabs show, with no profile switcher and no color-mode / idle-decay knobs.
- [ ] Confirm the lockdown holds against direct calls: `curl` the terminal live-ws, `GET /api/sessions/<id>/diff/files`, `POST /api/projects`, and `PATCH /api/settings` and confirm each returns 403 `{"error":"cityhall_mode"}`.
- [ ] With zero projects configured, confirm session creation returns 400 `cityhall_no_projects`.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan (with a multi-model design debate on structuring, server-side enforcement, the all-projects model, and the agent dependency), and implementation were drafted by Claude under my direction. I made the design calls, reviewed each commit, and run the manual test steps.

## Summary

- Added a locked-down **CityHall client mode** controlled by `AOE_CITYHALL_MODE`, determined at server startup and surfaced to the frontend via `/api/about` as `cityhall_mode`.
- Updated the web dashboard to be **composer-first and structured-view-only**:
  - Hides terminal, diff, plugin, and project-management UI (including the profile switcher and the home “Clone URL” action).
  - Makes the “New session” flow **name-only** across configured projects by hiding project/agent “more options” steps and allowing launch without a preselected path.
  - Curates **Settings** down to Theme, Sessions (trash-related options), MCP servers, Telemetry, and Plugins, with selective field hiding (e.g., Theme color/decay fields).
- Enforced restrictions **server-side**:
  - Returns **403** for CityHall-restricted HTTP endpoints (terminal/diff/session controls, project ops, filesystem browsing, settings updates, and ACP-related endpoints), and blocks relevant WebSocket entry points for terminal/live shell.
  - Keeps CityHall sessions in **structured view**, blocks ACP agent switching and ACP disabling.
  - Applies CityHall defaults for **worktrees** and pins ACP concurrency/resume behavior.
- Improved capability-driven UI behavior and coverage:
  - Frontend derives UI capability flags from `serverAbout.cityhall_mode` to drive pane/feature visibility and name-only wizard behavior.
  - Added documentation for the environment variable and example startup.
  - Added/expanded automated tests: Vitest for `getClientCapabilities`, and Playwright coverage for the restricted Settings/UI and name-only session wizard.

## Benefits

- Provides a safer, simpler CityHall experience for non-technical users by restricting advanced capabilities in both the UI and backend, reducing the chance of misuse while keeping “composer + structured chat” as the core interaction.

## Potential follow-ups

- Broaden test coverage to verify **403 behavior** for every restricted HTTP route and every relevant WebSocket route under CityHall mode.


web/src/App.tsx
```diff
@@ -653,18 +654,26 @@ function AppContent({
   const [sidebarOpen, setSidebarOpen] = useState(() => window.innerWidth >= 768);
   const keyboardProxyRef = useRef<HTMLTextAreaElement>(null);
 
+  const [serverAbout, setServerAbout] = useState<ServerAbout | null>(null);
+  // CityHall client mode collapses the dashboard to a locked-down end-user
+  // client; the capability flags gate the terminal/diff panes, project
+  // management, the wizard, and settings from one place. See #7.
+  const caps = useMemo(() => getClientCapabilities(serverAbout), [serverAbout]);
+
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/9b225a2a — the shell now renders a placeholder until `serverAboutLoaded`, so CityHall-gated affordances never flash in before `caps.cityhall` settles.

Seluj78:
> @coderabbitai (nitpick, `src/server/api/sessions.rs`): Consume `projects` to avoid cloning paths.

Done in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/82b761af — the loaded `projects` vector is now moved by value into `body.path` / `body.extra_repo_paths` instead of cloning each path string.

src/server/api/sessions.rs
```diff
@@ -4677,11 +4703,57 @@ pub async fn create_session(
...
+                    "error": "cityhall_no_projects",
+                    "message": "CityHall mode requires at least one configured project"
+                })),
+            )
+                .into_response();
+        }
+        body.scratch = false;
+        // Non-empty checked above, so the first path always exists.
+        let mut paths = projects.into_iter().map(|p| p.path);
+        body.path = paths.next().unwrap();
```
jerome-benoit, line 4934: [BLOCKER] This derivation neutralizes only `scratch` / `path` / `extra_repo_paths` (and `view` below), but the comment's promise ("a crafted request cannot escape the mode") does not hold: `command_override`, `extra_args`, `extra_env`, `yolo_mode`, `worktree_branch`, `create_new_branch`/`base_branch`, `fork_from`, `import_acp_session_id` all pass through unchanged.

The serious one is `command_override`: body -> `InstanceParams.command_override` -> `instance.command` -> `command_override_for_spawn` -> ACP `SpawnRequest.agent_command_override` -> `supervisor::apply_agent_command_override`, where `command_matches_binary` validates the *registry-default* command and then `spec.command = argv.remove(0)` adopts the client's `argv[0]` **without revalidating the replacement binary**. For a registry ACP tool (e.g. `claude`), `command_override: "/bin/sh -c '...'"` passes the ACP-capable gate here and spawns an arbitrary binary as the agent. Please reset every non-derived sensitive field to its default in this block, not just the four.

src/server/api/mod.rs
```diff
@@ -106,6 +106,28 @@ pub(super) fn read_only_response() -> axum::response::Response {
         .into_response()
 }
 
+/// Canonical 403 body for CityHall client mode (`AOE_CITYHALL_MODE`).
+/// Terminal, diff, project-management, and advanced-settings endpoints are
```
jerome-benoit: [MAJOR] This comment ("enforced here, not only by [the UI]") is the feature's contract, but it overstates current coverage. Endpoints reachable in CityHall mode because they carry no `cityhall_block`: `POST /api/git/clone` (`clone_repo`), `GET /api/sessions/{id}/output` (`read_output`, raw terminal content), `PATCH /api/profiles/{name}/settings` (`update_profile_settings`, accepts any non-elevation profile-overridable leaf), and the ACP mutating routes beyond switch-agent/disable, plus `git/branches`, `git/is-repo`, `acp/worker-log`. Unlike `read_only`, there is no static guard audit (`every_mutating_handler_has_read_only_guard`) nor route-level 403 test. Please add the missing guards and a `cityhall` guard-audit test.

jerome-benoit (changes requested):
## Review: CityHall client mode

I reviewed this with a focus on the feature's central promise, stated in the code itself (`api/mod.rs`): the lockdown is *"enforced here, not only by [the UI]"*. The frontend and the plumbing are well done, but the server-side enforcement is incomplete, unverified by tests, and non-durable, so I'm requesting changes. The BLOCKER and the two enforcement MAJORs are the load-bearing ones.

### BLOCKER - the CityHall create derivation does not close the sandbox (arbitrary-command path)
The `if state.cityhall_mode` block in `create_session` neutralizes only `scratch`, `path`, `extra_repo_paths`, `view` (see inline). It passes `command_override`, `extra_args`, `extra_env`, `yolo_mode`, `worktree_branch`, `fork_from`, `import_acp_session_id` through unchanged. Traced end to end: body `command_override` -> `InstanceParams.command_override` -> `instance.command` (builder) -> `command_override_for_spawn` -> ACP `SpawnRequest.agent_command_override` -> `supervisor::apply_agent_command_override`. There, `command_matches_binary(&spec.command, agent_def.binary)` validates the *pre-override* (registry-default) command, then `spec.command = argv.remove(0)` takes the client's `argv[0]` **without revalidating it**. So for a registry ACP tool (e.g. `claude`), a create with `command_override: "/bin/sh -c '...'"` passes the ACP-capable gate and spawns an arbitrary binary as the "agent" - arbitrary command execution for a client that is supposed to be locked down. Fix: in the CityHall block, reset every non-derived sensitive body field to defaults, not just the four. (Please double-check the `command_override_for_spawn` link; the two endpoints - unfiltered create + unvalidated replacement binary in the supervisor - I confirmed directly.)

### MAJOR - sensitive routes reachable in CityHall mode (server does not reject, only the UI hides)
Several sensitive routes have no `cityhall_block` (files the PR did not touch), which is why the "project-management endpoints are unreachable" comment is not accurate today:
- `POST /api/git/clone` (`clone_repo`, `git.rs`): only `read_only`-gated -> a locked-down client can clone any URL into `$HOME` (fs write + network fetch + disk-fill). The homescreen "Clone URL" is hidden in the UI but the endpoint answers.
- `GET /api/sessions/{id}/output` (`read_output`, `sessions.rs`): unguarded -> returns raw tmux/terminal pane content by session id, partially defeating "terminal is unreachable" even though the WS relay and terminal lifecycle were correctly gated.
- `PATCH /api/profiles/{name}/settings` (`update_profile_settings`, `system.rs`): must stay open for the curated trash toggles, but it accepts **any** non-elevation profile-overridable leaf, not just the three curated fields; with the `--auth=none` deployment the docs suggest, there is no elevation wall either. Server-side curation here is UI-only.
- ACP session-mutating routes beyond the two gated (`switch-agent`, `disable`): `set-mode`, `set-config-option`, `spawn`, `shutdown`, `install-agent`, `cancel`, `force-end-turn`, plus `acp/worker-log`, `git/branches`, `git/is-repo`.

### MAJOR - CityHall mode is lost across daemon restart / `aoe update`
`AppState.cityhall_mode` is resolved only from the `AOE_CITYHALL_MODE` env var; `ServeLaunch` persists `read_only` but has no `cityhall_mode`, and it is not replayed by `to_serve_args`. So `AOE_CITYHALL_MODE=1 aoe serve --daemon` followed by `aoe update` (or `aoe serve --restart` from a shell without the var) silently drops the lockdown. `read_only` mirrors correctly via `--read-only`; CityHall should mirror it (persist in `ServeLaunch` and re-set the env / add a `--cityhall` flag on the restart child).

### MAJOR - no server-side test exercises the lockdown
The `test(serve)` commit unit-tests helpers only (`cityhall_response` status, `agent_is_acp_capable`, `apply_cityhall_overrides`). No test builds `AppState { cityhall_mode: true }`, calls a guarded handler, and asserts the 403/400. That is why the reachable routes above went unnoticed. `read_only` has both a live spec and a static `every_mutating_handler_has_read_only_guard` audit; CityHall has neither. Please add route-level 403/400 tests and, ideally, a static `cityhall` guard audit.

### MINOR
- `PATCH /api/theme` (`update_theme`) writes `color_mode` server-side though the UI hides it in CityHall.
- Session lifecycle (rename/delete/trash/stop/start/diff-base/worktree-name) and profile CRUD are `read_only`-only, not cityhall-gated.
- `send_message` relies on the `StructuredView` behavioral error rather than the mode guard.
- `AOE_CITYHALL_MODE` is read in three places (`server/mod.rs`, `profile_config.rs`, `serve.rs`) instead of resolved once and threaded - the root cause of the restart gap.
- "Primary" project = first entry in merged registry order; the `.unwrap()` is safe (guarded) but arbitrary with multiple projects - worth a doc note.
- coverage-matrix entry omits `Dashboard.tsx` / `SchemaSection.tsx` though the spec exercises them.

### Praise
- Flag resolution is immutable and `cityhall_block` mirrors `read_only_block` cleanly.
- Re-pinning the overrides *after* the repo-config merge (in-memory only, flag-gated) is the right call: a repo `config.toml` cannot relax them, on-disk config is never corrupted, and normal mode is byte-identical.
- The WS PTY relay guard closes the "direct WebSocket bypasses the hidden UI" hole.
- The create-time ACP-capability 400 gate is the right failure mode.
- Frontend gating is exemplary: a single `getClientCapabilities` mapping, computed once, with a shell-hold so nothing flashes before caps load.
- Dropping the `max_concurrent_resumes` override is correct (pruned to a constant in v022); the PR description's mention of it is just stale.

Net: sound architecture, but the server-side lockdown - the whole point - needs to be completed (strip all create fields; gate clone/read_output/update_profile_settings/the ACP routes), made durable across restart, and covered by route-level tests. Happy to re-review quickly once those land.

src/server/api/sessions.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/237c76db. The CityHall create block now resets **every** client-controlled spawn/branch field, not just path/repos/view/scratch: `command_override`, `extra_args`, `extra_env`, `yolo_mode`, `worktree_branch`, `create_new_branch`, `base_branch`, `sandbox`/`sandbox_image`, and (behind serve) `fork_from` / `import_acp_session_id`. So the `command_override` -> unvalidated `argv[0]` path can no longer smuggle an arbitrary binary past the ACP-capable gate. Good catch on tracing that end to end.

src/server/api/mod.rs
Seluj78: Fixed across https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/237c76db (guards) and https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/1eb69679 (tests). Added `cityhall_block` to `clone_repo`, `list_branches`, `is_git_repo`, `read_output`, `acp_worker_log`, `spawn_acp`/`shutdown_acp`/`install_agent`/`acp_enable`/`acp_set_mode`/`acp_set_config_option`, `send_message`, and profile CRUD. `update_profile_settings` stays open for the curated trash cluster but field-filters every other leaf; `update_theme` drops a client-supplied `color_mode`. There is now a static `every_cityhall_gated_handler_has_guard` audit (mirroring the read-only one) plus route-level 403 tests in `serve_cityhall_lockdown.rs`. I deliberately left the conversation-control ACP routes (`acp_prompt`, `acp_cancel`, `acp_force_end_turn`, approvals/elicitations, attachments) and session lifecycle open: those are the composer's core loop and the client is interactive, not read-only. Shout if you'd rather lock any of those down too.

Seluj78:
@jerome-benoit thanks for the thorough, correct review. All four load-bearing points and the minors are addressed (4 commits on top):

- **BLOCKER (create sandbox escape)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/237c76db: the create block now resets every client-controlled spawn/branch field, closing the `command_override` arbitrary-binary path.
- **MAJOR (reachable routes)** — same commit: guards added to clone / git probes / `read_output` / worker-log / ACP lifecycle+config / `send_message` / profile CRUD; `update_profile_settings` field-filtered to the curated trash cluster; `update_theme` drops `color_mode`.
- **MAJOR (durability)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/b0d61c95: added a `--cityhall` flag equivalent to the env var, resolved into one value at `run` entry, replayed to the daemon child, and persisted in `serve.launch` so it survives `--restart` and the post-`aoe update` re-exec. Mirrors `--read-only`.
- **MAJOR (no server-side test)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/1eb69679: a static `every_cityhall_gated_handler_has_guard` audit plus route-level 403 tests (`serve_cityhall_lockdown.rs`) that build `AppState { cityhall_mode: true }` and assert the guard fires with the canonical body.
- **MINORs** — `color_mode` (done above), `send_message` now mode-gated (not relying on the behavioral error), the three-place env read is now resolved once and seeded, a doc note on the arbitrary-but-deterministic primary project, and the coverage-matrix now lists `Dashboard.tsx` / `SchemaSection.tsx` (https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/41efd627).

One deliberate scope call: I left the conversation-control ACP routes (`acp_prompt`, `acp_cancel`, `acp_force_end_turn`, approvals/elicitations, attachments) and session lifecycle (rename/delete/trash/stop/start) **open**. CityHall is an interactive composer client, not read-only, so those are the end user's core loop rather than admin surfaces. Happy to gate any of them if you'd prefer a stricter cut. Ready for re-review.

src/server/api/mod.rs
```diff
@@ -386,6 +422,120 @@ mod tests {
...
+    /// for the create + curated-settings paths, restrict) has to check the mode
+    /// and return 403/400 before doing its work. Mirrors
+    /// `every_mutating_handler_has_read_only_guard` in intent and mechanism: a
+    /// static source walk so a contributor who adds a route reachable in the
+    /// locked-down client, or drops an existing guard, fails the build instead
+    /// of silently reopening a hole (the gap this test exists to catch, see #7).
+    /// The end-to-end coverage lives in the route-level 403 tests below and the
+    /// web specs; this guards the source-level invariant.
+    #[test]
+    fn every_cityhall_gated_handler_has_guard() {
```
jerome-benoit: [BLOCKER + MAJOR] This audit is the mechanism that should prevent ungated sensitive routes, but its hand-list omits `api/mcp.rs` and `api/plugins.rs` entirely, so two admin surfaces fell through:

- **Plugins (BLOCKER)**: `set_plugin_enabled` / `start_plugin_install` / `start_plugin_uninstall` / `apply_plugin_update` (plugins.rs) gate only on `mutation_gate` (`read_only` + elevation), never `cityhall_block`. `start_plugin_install` runs `apply_install(&body.source, ...)` on a client-supplied `gh:owner/repo` = arbitrary host-side code install by a locked-down user (elevation is no barrier: the CityHall user holds the passphrase, and `--auth=none` makes `handler_elevated` always true). The Plugins tab renders live controls.
- **MCP (MAJOR)**: `keep_mcp_server` / `drop_mcp_server` / `resolve_mcp_conflict` (mcp.rs) carry only `read_only`; the MCP tab renders live Keep/Drop/Resolve. Bounded (client can't inject a new command; definitions come from host-owned native config), but it contradicts the doc's "display only".

Please gate these handlers with `cityhall_block` (and render the tabs read-only), and add `mcp.rs` + `plugins.rs` to this audit's list so the gap can't reopen.

docs/guides/web-dashboard.md
```diff
@@ -123,6 +123,16 @@ Requires `cloudflared` on the host:
...
...nd-user client: only the message composer and the structured (chat) view are reachable. Terminal and diff panes and project management are hidden in the UI and rejected server-side, so a direct API or WebSocket call cannot reach them either: the terminal keystroke and raw-output routes, git clone/branch/is-repo probes, agent/worker lifecycle and config routes, project CRUD, and profile CRUD all return 403, and session creation is server-derived (every client-controlled spawn field, including `command_override`, is reset). Settings are curated down to Theme (without the color-mode and idle-decay knobs; the server also drops a client-supplied color mode), a delete-to-trash toggle (the profile-settings write is field-filtered to just the trash cluster), MCP servers (display only), Telemetry, and Plugins; the profile switcher and all other settings are removed. New sessions are created by name only; each spans every configured project and runs the default agent in structured view, so the deployment's default agent must be ACP-capable (session creation is rejected otherwise, and it fails if no project is configured). Worktrees are enabled by default and the ACP worker ceiling is raised.
```
jerome-benoit: [doc vs behavior] This says non-composer surfaces are "rejected server-side, so a direct API or WebSocket call cannot reach them," and lists "MCP servers (display only)" plus "Plugins." But at HEAD the MCP mutation routes (`keep`/`drop`/`resolve_mcp_conflict`) and the plugin mutation routes (`set_plugin_enabled`/`start_plugin_install`/uninstall/update) carry no `cityhall_block` and their Settings tabs render live controls - so they ARE reachable server-side. Either gate them to match this text, or soften the claim (and drop "display only" for MCP).

src/server/api/sessions.rs
```diff
@@ -4677,11 +4703,84 @@ pub async fn create_session(
...
+            return (
+                StatusCode::BAD_REQUEST,
+                Json(serde_json::json!({
+                    "error": "cityhall_no_projects",
+                    "message": "CityHall mode requires at least one configured project"
+                })),
+            )
+                .into_response();
+        }
+        body.scratch = false;
```
jerome-benoit, line 4904: [minor] The block comment says it resets every client-controlled spawn/branch field, but `worktree_enabled` (a `CreateSessionBody` field) isn't reset here. Since `create_body_uses_worktree = worktree_enabled || worktree_branch.is_some()`, a client sending `worktree_enabled: true` still forces worktree mode. Low impact (stays in the server-derived project, no client path/branch/base), but reset it for the comment to hold.

src/cli/serve.rs
```diff
@@ -712,6 +726,18 @@ pub async fn run(profile: &str, args: ServeArgs) -> Result<()> {
...
+    // Resolve CityHall mode once, both directions: the `--cityhall` flag (what
+    // the daemon child / restart / post-update re-exec replays) and the
+    // `AOE_CITYHALL_MODE` env var are equivalent. Normalizing `args.cityhall`
+    // and the env var to the same value makes every downstream reader
+    // (`args.cityhall` for the child spawn + ServeLaunch, the env var for
+    // `AppState` / `profile_config` / the startup banner) agree without
+    // threading the bool through each of them. See #7.
+    args.cityhall = args.cityhall || std::env::var_os("AOE_CITYHALL_MODE").is_some();
+    if args.cityhall {
+        std::env::set_var("AOE_CITYHALL_MODE", "1");
```
jerome-benoit: [minor] `std::env::set_var` here runs inside the live tokio runtime (this is an `async fn` awaited from `#[tokio::main]`) without `unsafe`/a SAFETY note. It's sound on edition 2021 and race-free in the current ordering (the readers run later), but it diverges from this repo's own `main.rs` `unsafe`+SAFETY precedent for the same operation and becomes a hard error on the 2024 migration. Prefer threading the bool, or document the single-threaded-at-this-point invariant.

jerome-benoit (changes requested):
## Re-review round 2 (`41efd627`)

Great progress: all four load-bearing items from my last review are genuinely closed and verified. The create block zeroes the client-controlled spawn fields (the `command_override` RCE path is cut), the previously-reachable routes are gated, the restart / `aoe update` durability mirrors `--read-only`, and there's a real route-level 403 suite plus a static guard audit that both pass. Your deliberate scope call to leave the conversation-control ACP routes and session lifecycle open checks out too: those are UI-exposed for the user's own sessions, so open server routes are consistent.

Two admin surfaces still fell through the lockdown. Neither is in your deliberate-open list, both render live controls in the curated CityHall Settings, and the `every_cityhall_gated_handler_has_guard` audit covers neither file, so it passes without catching them.

### BLOCKER - plugin install/enable/uninstall/update reachable in CityHall
`set_plugin_enabled` / `start_plugin_install` / `start_plugin_uninstall` / `apply_plugin_update` (`src/server/api/plugins.rs`) gate only on `mutation_gate` (`read_only` + elevation), no `cityhall_block`. `start_plugin_install` runs `plugin::install::apply_install(&body.source, ...)` on a client-supplied `gh:owner/repo`, installing a plugin that executes host-side. Elevation is not a barrier against the CityHall user: they hold the passphrase, and the docs' own suggested `--auth=none` deployment makes `handler_elevated` always true. So a locked-down end-user can install and run arbitrary code on the host - a fresh escape of the mode's core promise. The Plugins tab renders live install/enable/uninstall controls (`PluginsSettings` takes no read-only/cityhall prop). Please gate the plugin mutation routes with `cityhall_block` and render the tab read-only; or, if plugins are meant to be usable in CityHall, say so explicitly and justify arbitrary host-side install in a locked-down client, because the doc claims non-composer surfaces are "rejected server-side."

### MAJOR - MCP mutation routes reachable vs "display only" doc
`keep_mcp_server` / `drop_mcp_server` / `resolve_mcp_conflict` (`src/server/api/mcp.rs`) carry only `read_only`, and the MCP tab renders live Keep/Drop/Resolve. To your credit the blast radius is bounded: the client supplies only `{agent, name, winner, fingerprint}` and the promoted definitions come from the operator's own host-side native config snapshots, so this cannot inject a new command. But `docs/guides/web-dashboard.md` says "MCP servers (display only)" and "a direct API call cannot reach them," which is false. Gate the three routes (and render read-only) or fix the doc.

### Audit gap tying these together
`every_cityhall_gated_handler_has_guard` enumerates a hand-list that omits `mcp.rs` and `plugins.rs` entirely, so the guard-audit meant to prevent exactly this cannot see the gap. Please add both files' mutation handlers to the audit (guarded, or listed-with-justification) so the mutating surface is fully covered.

### MINOR
- `worktree_enabled` isn't reset in the CityHall create block, so the "resets every client-controlled spawn/branch field" comment is inaccurate (low impact: stays in the derived project, no client path/branch).
- `std::env::set_var("AOE_CITYHALL_MODE")` at `run` entry executes in the live tokio runtime without `unsafe`/SAFETY note (sound on edition 2021 and race-free in the current ordering, but it diverges from `main.rs`'s precedent and will hard-error on the 2024 migration).

Everything else is solid. Happy to re-review quickly once the plugin routes are gated (or justified) and the MCP doc/gate is reconciled.

src/server/api/mod.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/01fae987. Added `cityhall_block` inside plugins' `mutation_gate` (covers install/uninstall/enable/update/dismiss) and directly to the three MCP routes (keep/drop/resolve), and added both files to this audit's list (the audit now accepts the `mutation_gate(` helper the same way it accepts `read_only_block(` for acp). Route-level 403 tests extended to cover plugin-install/set-enabled and mcp-keep/drop. The Plugins + MCP tabs also render read-only now (876efb8a).

docs/guides/web-dashboard.md
Seluj78: Gated to match the text (https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/01fae987): the MCP keep/drop/resolve and plugin install/uninstall/enable/update routes now return 403, and both tabs render display-only. Updated the doc to name those closed routes explicitly and keep "display only" accurate.

src/server/api/sessions.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/d366eeca. `body.worktree_enabled = false` is now in the reset block, so the comment holds.

src/cli/serve.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/3efa3f6f. Moved the flag-to-env bridge to the early single-threaded point in `main.rs` (same spot + unsafe/SAFETY note as the `AOE_DAEMON_URL` seed); `serve::run` now only reads the env var, so there is no `set_var` inside the tokio runtime and it stays 2024-edition clean.

Seluj78:
Round 2 addressed (4 commits, `41efd627..876efb8a`):

- **BLOCKER + MAJOR (audit gap)** — gated the plugin lifecycle routes (via `mutation_gate`) and the MCP keep/drop/resolve routes, added `mcp.rs` + `plugins.rs` to the guard audit, and extended the route-level 403 tests.
- **doc** — the closed routes are now named and "display only" is accurate.
- **minor (worktree_enabled)** — reset in the create block so the comment holds.
- **minor (set_var)** — moved to the early single-threaded `main.rs` seed (unsafe + SAFETY, matching the `AOE_DAEMON_URL` precedent); no in-runtime write, 2024-edition clean.

Plus the frontend now renders the MCP + Plugins tabs read-only to match. `cargo fmt`/`clippy`/cityhall tests + audit + web tsc/lint/oxfmt/coverage-matrix all green. Ready for re-review. (Branch is behind main; I'll rebase before merge.)

src/server/api/sessions.rs
```diff
@@ -4677,11 +4703,85 @@ pub async fn create_session(
...
+        // Reset every client-controllable spawn / branch field to its default.
+        // Deriving path/repos/view is not enough: a crafted request could still
+        // smuggle an alternate binary, extra args/env, yolo mode, a chosen
+        // branch/base, or a sandbox container past the locked-down mode.
+        // `command_override` is the load-bearing one: the ACP supervisor
+        // validates the registry-default binary but then adopts the client's
+        // `argv[0]` unchecked, so `command_override: "/bin/sh -c ..."` on a
+        // registry ACP tool would pass the ACP-capable gate below and spawn an
+        // arbitrary binary as the agent. See #7 review.
+        body.command_override = String::new();
```
jerome-benoit, line 4914: [MAJOR] You correctly neutralize `command_override` at create time here - but `ensure_session` (5270) and `start_session` (3091) re-run a *stored* `command_override` on respawn (via `start_with_resume_fallback` -> `build_host_command`, instance.rs:3380) and are gated only by `read_only`, not `cityhall_block`. Since CityHall does not imply read-only (`read_only` comes only from `--read-only`, serve.rs:916) and `list_sessions` (587) is unfiltered, an authenticated CityHall client can enumerate a pre-existing plain/terminal session and POST /ensure or /start to spawn its stored host binary. The same reset reasoning you applied here needs to apply to the respawn path (add `cityhall_block`, or filter `list_sessions`).

src/server/api/mod.rs
```diff
@@ -386,6 +422,141 @@ mod tests {
...
+        let cases: &[(&str, &str, &[&str])] = &[
+            (
+                "api/sessions.rs",
+                include_str!("sessions.rs"),
+                &[
+                    "create_session",
+                    "send_message",
+                    "ensure_terminal",
+                    "ensure_container_terminal",
+                    "kill_terminal",
```
jerome-benoit: [Root cause] This audit is a hardcoded allowlist of handler names, so a mutating handler that is simply absent from the list is never checked. That is exactly how `ensure_session`/`start_session` (and `stop_session`/`delete_session`/`delete_workspace`) slipped through - they respawn/destroy host sessions but are not in this list. Please replace the allowlist with a scan that flags any mutating `pub async fn` handler lacking a guard, so this class of omission fails the build.

jerome-benoit (changes requested):
## Re-review round 3: CityHall lockdown

Thank you for the round-2 work. I verified all four round-2 findings with three independent read-only reviewers plus my own `--features serve` build and test run in an isolated checkout, and they are fully resolved:

- **BLOCKER (plugin-install reachable) - RESOLVED.** `cityhall_block` now sits inside `mutation_gate` ahead of the elevation check, so every plugin mutation route (install/uninstall/enable/update) returns 403 in CityHall independent of auth mode. The `--auth=none` collapse is closed because the block no longer depends on elevation.
- **MAJOR (MCP keep/drop/resolve reachable) - RESOLVED.** All three MCP mutation handlers gate before mutating; "display only" is now accurate.
- **worktree_enabled reset + env-seed ordering - RESOLVED.** `worktree_enabled` is forced off in the CityHall create derivation and cannot be re-enabled by the body; the flag is seeded before the runtime with no TOCTOU window.
- My own run: `cargo test --features serve --test serve_cityhall_lockdown` = 13/13 pass.

I am requesting changes because the adversarial pass surfaced that the lockdown gates the *terminal* surface but leaves the *session lifecycle* surface open, in the same security boundary this PR advertises.

### MAJOR - `ensure_session` / `start_session` respawn a stored `command_override` in CityHall
`ensure_session` (`sessions.rs:5270`) and `start_session` (`sessions.rs:3091`) are gated only by `read_only`, not `cityhall_block`. Both reach `start_with_resume_fallback` -> `build_host_command`, which for a plain session re-runs the stored `command_override` (`instance.rs:3380`) - an arbitrary host binary. CityHall does **not** imply read-only: `read_only` comes solely from `--read-only` (`serve.rs:916`) and is fully independent of `cityhall_mode`; the lockdown tests themselves run with `read_only: false, cityhall_mode: true`. `list_sessions` (`sessions.rs:587`) is unfiltered, so an authenticated CityHall client can enumerate a pre-existing plain/terminal session (created via TUI, `aoe add --cmd ...`, or a non-CityHall client on the same daemon) and `POST /api/sessions/{id}/ensure` or `/start` to kill-and-respawn its stored host command - with no elevation in the way (and `--auth=none` auto-grants elevation anyway). You already recognized this exact risk and reset `command_override` at *create* time (`sessions.rs:4738`); the *respawn* path needs the same treatment.

### Related - other ungated session-mutation routes
Same gap, gated only by `read_only`: `stop_session` (2950), `delete_session` (3867), `delete_workspace` (4184) let a CityHall client destroy or DoS arbitrary enumerated sessions; `summarize_session` (2873) and `force_smart_rename` (2765) trigger a server-side agent/LLM run; plus a tail of metadata writes (rename/pin/color/group/snooze/...). Given the PR blocks `ensure_terminal`/`kill_terminal`, leaving `ensure_session`/`start_session`/`stop_session`/`delete_session` open is inconsistent with the mode's own intent.

### Root cause - the audit is an allowlist, not a scanner
`every_cityhall_gated_handler_has_guard` (`api/mod.rs:435`) checks a hardcoded list of handler names, so a mutating handler that is simply absent from the list is never flagged - which is exactly how `ensure_session`/`start_session` slipped through. Please replace the allowlist with a scan that flags any mutating `pub async fn` handler lacking a guard.

### MINOR - `trust_hooks` not neutralized in CityHall create
`body.trust_hooks` is not reset in the CityHall create block and flows to the spec, letting a client approve and trigger the operator repo's `on_create` host commands (and persist durable trust). Bounded to operator-configured repos, but it contradicts the block's own comment that every client-supplied field that could escape the mode is neutralized. Recommend forcing it off.

### Suggested fixes
1. Add `if let Some(resp) = super::cityhall_block(&state) { return resp; }` to `ensure_session`, `start_session`, `stop_session`, `delete_session`, `delete_workspace` (and consider the lifecycle/metadata writes), or filter `list_sessions` in CityHall so only mode-derived sessions are visible/actionable.
2. Reset `body.trust_hooks` in the CityHall create derivation.
3. Convert the audit to a scanner.

Round-2 remediation is solid; this is about extending the same server-side gate to the session-lifecycle routes so the mode holds to its stated "closed server-side" contract.

src/server/api/sessions.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/4ff2f411. CityHall only ever creates structured sessions, so `list_sessions` now filters to structured, and a new `cityhall_block_non_structured(&state, &id)` gate makes every per-session lifecycle/metadata route (ensure/start/stop/delete/delete_workspace/summarize/force_smart_rename/rename/pin/color/group/snooze/archive/unread/diff_base/notifications/set_worktree_name/trash/restore) refuse a non-structured or unknown target. So the `build_host_command` respawn of a stored `command_override` (and destroy/edit of an enumerated plain session) is closed, while the CityHall user can still manage their own structured sessions. Also reset `trust_hooks` in the create block. Route-level 403 tests added for the lifecycle routes.

src/server/api/mod.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/d2f8fc44. Replaced the allowlist with `every_mutating_route_is_cityhall_gated_or_exempted`: it derives the mutating set from the router (every `post/patch/delete(api::…)`) and requires each handler to either carry a guard or be in an explicit `CITYHALL_EXEMPT` list with a reason. A new ungated+unexempted mutating route now fails the build (and a stale exemption is flagged), so absence-from-a-list can't reopen a hole. The exemptions are the composer loop, telemetry consent, and per-device UI prefs.

Seluj78:
Round 3 addressed (2 commits, `71f6d327..d2f8fc44`):

- **MAJOR (session-lifecycle respawn/RCE)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/4ff2f411: CityHall only creates structured sessions, so `list_sessions` is filtered to structured and a `cityhall_block_non_structured` gate makes every per-session lifecycle/metadata route refuse a non-structured or unknown target. The stored-`command_override` respawn via `build_host_command`, plus destroy/edit of an enumerated plain session, are closed; the CityHall user can still manage their own structured sessions.
- **MINOR (trust_hooks)** — same commit: reset in the create derivation so a client can't approve/trigger the operator repo's `on_create` host commands.
- **Root cause (allowlist audit)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/d2f8fc44: the audit is now a router-driven scanner — every `post/patch/delete(api::…)` must be guarded or explicitly exempted-with-reason, else the build fails. This is what would have caught ensure/start on their own.

Also gated the stragglers the scanner surfaced (patch_log_level, preview_plugin_install, invoke_plugin_action/command). `cargo fmt`/`clippy`, api unit tests, the read-only audit, the new scanner, and 17 `serve_cityhall_lockdown` route tests all pass. Ready for re-review.

src/server/api/sessions.rs
```diff
@@ -4204,6 +4312,13 @@ pub async fn delete_workspace(
             .into_response();
     };
 
+    // CityHall: a workspace's sessions are all structured when the mode created
+    // it, so gating the owner refuses a locked-down client tearing down an
+    // enumerated non-structured workspace. See #7.
+    if let Some(resp) = cityhall_block_non_structured(&state, &owner_id).await {
```
jerome-benoit: [MAJOR] This gates only `owner_id` (`session_ids.first()`), but `order_workspace_deletion` + `purge_workspace_artifacts` delete *every* id in `session_ids` with no per-sibling structured/membership check. A CityHall client can send `session_ids: [<own structured id>, <foreign plain id>]`: the owner passes, the foreign plain session is destroyed as a sibling (cross-session destroy). Resolve *all* `session_ids` (e.g. a `Vec<StructuredTarget>`), not just the first.

src/server/api/mod.rs
```diff
@@ -386,6 +423,150 @@ mod tests {
...
+            "cityhall_block(",
+            "cityhall_block_non_structured(",
+            "state.cityhall_mode",
+            "mutation_gate(",
+        ];
+        let body_terminators = ["\npub async fn ", "\npub fn ", "\nasync fn ", "\nfn "];
+
+        // Extract every `post|patch|delete(api::NAME)` handler from the router.
+        let mut routed: Vec<String> = Vec::new();
+        for method in ["post(api::", "patch(api::", "delete(api::"] {
```
jerome-benoit: This is the audit's structural weakness. It extracts only `post(api::`/`patch(api::`/`delete(api::`, so it is blind to (F2) other module prefixes - `push::subscribe/unsubscribe/test`, `login::*` are unscanned - and (G3) the `put` method: `PUT /api/workspace-ordering` -> `update_workspace_ordering` is ungated and invisible here. And the guard check below (line 543) tests only that a guard *string is present in the body*, not that it runs before the effect or on the right target - which is why `delete_workspace` passes the audit while being exploitable (F1). A text scan can't establish the invariant; assert instead that every router-registered mutating route (any prefix, any method) is present in an explicit CityHall allowlist/exempt table.

jerome-benoit (changes requested):
## Re-review round 4: CityHall lockdown

The round-3 findings are genuinely fixed and I want to credit that: the session-lifecycle respawn/destroy RCE is closed (`cityhall_block_non_structured` on ensure/start/stop/delete/summarize/rename + metadata, `list_sessions` filtered to structured, `trust_hooks` reset, stragglers gated). I verified with 3 read-only reviewers plus my own `--features serve` run: the lockdown suite is 17/17 (incl. foreign-target) and the scanner lib tests 4/4.

I'm still requesting changes, because a second pass (incl. an Oracle consult and a full cross-module route enumeration) shows the recurring rounds are a **design** problem, plus one live gap.

### Root cause: the mode is default-allow, not default-deny
Enforcement is a per-handler deny-list: a route is reachable unless its handler *remembers* to call `cityhall_block*`, backed by a text-scan audit. That fails open on three independent axes, and each round tripped a different one:
- **target** - a guard that covers the wrong object (R3 respawn; and F1 below);
- **scope** - routes the audit can't see (F2 below);
- **method** - methods the audit doesn't scan (G3 below).

You can't prove the negative ("no route escapes") over a growing surface, so a new sample turns up every round. The audit gives false confidence: it checks that a guard *string exists in a byte range*, not that authorization *runs before the effect, on the right target, for every route*.

### Live gap (MAJOR) - F1: `delete_workspace` gates only the owner
`cityhall_block_non_structured(&state, &owner_id)` checks `session_ids.first()` only, but `order_workspace_deletion` + `purge_workspace_artifacts` tear down *every* id with no per-sibling structured/membership check. In CityHall a client can send `session_ids: [<own structured id>, <foreign plain id>]` and destroy the foreign session (cross-session destroy; not RCE, since /ensure and /start are individually gated). Fix: resolve *all* `session_ids`, not `first()`.

### Audit gaps (MINOR) - F2 + G3
The scanner extracts only `post(api::` / `patch(api::` / `delete(api::`, so it is blind to (F2) other module prefixes - `push::subscribe/unsubscribe/test`, `login::*` are unscanned/ungated (benign today: owner-scoped push, and login must stay open) - and (G3) the `put` method entirely: `PUT /api/workspace-ordering` -> `update_workspace_ordering` is ungated (only `read_only`), not exempt, and unscanned, so it writes shared workspace ordering in CityHall. Both are the same unsoundness as F1: the audit samples, it doesn't establish.

### The definitive fix (to stop the cycle)
The choke point already exists - `auth_middleware` and `access_policy` are `from_fn_with_state` layers wrapping the whole router (mod.rs:1922-1928). CityHall (a narrower authorization question) should sit there too:
1. **Default-deny middleware** beside `access_policy`, no-op when `!cityhall_mode`; otherwise match `(Method, MatchedPath-template)` against one explicit allowlist table, everything else -> 403. This covers every module prefix and method uniformly - a new route is unreachable until deliberately listed (fails closed).
2. **Typed target gate** for the two body-dependent routes: a sanitized-create newtype the spawn path is the only consumer of; a `Vec<StructuredTarget>` resolved from *all* `delete_workspace` ids (so "owner-only" becomes inexpressible).
3. **Audit = table exhaustiveness**: assert every router-registered mutating route (any prefix, any method) appears in the allowlist or a justified exempt entry - sound, because both sides are route enumerations, not handler text.

Minimum bar to approve: fix F1 now, and move the reachability boundary to the middleware + table audit. The full type-state (step 2 everywhere) is a fine scheduled follow-up. The existing per-handler `cityhall_block*` calls can stay as defense-in-depth, but shouldn't be *the* enforcement.

Great progress this round - closing this out with the middleware makes the whole class of bug impossible instead of point-fixing a fifth time.

src/server/api/sessions.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/e06cbc52. Added `cityhall_block_any_non_structured(&state, &session_ids)` and swapped the owner-only check for it, so the workspace delete refuses unless EVERY id (not just `first()`) resolves to a structured session the mode created. The `[own_structured, foreign_plain]` sibling-destroy is closed.

src/server/api/mod.rs
Seluj78: Done in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/535b46d0. Moved the reachability boundary to a default-deny `cityhall_gate` middleware beside `access_policy`: in CityHall every mutating request (POST/PUT/PATCH/DELETE) whose `(method, matched-path template)` isn't in `CITYHALL_MUTATION_ALLOW` is 403'd before the handler runs, covering every prefix + method uniformly (PUT /api/workspace-ordering and the push::/login:: prefixes included). Replaced the text scan with `every_mutating_route_is_cityhall_classified`, which enumerates the router's mutating routes and asserts each is in exactly one of the ALLOW/DENY tables (no stale entries) so absence-by-omission fails the build. Per-handler `cityhall_block*` stays as defense in depth. Full type-state newtypes noted as the scheduled follow-up you flagged.

Seluj78:
Round 4 addressed (2 commits, `d2f8fc44..535b46d0`) — took the definitive fix.

- **F1 (delete_workspace siblings)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/e06cbc52: resolves ALL `session_ids` as structured targets, not just the owner.
- **Root cause + F2/G3 (default-allow / audit blindness)** — https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/535b46d0: the reachability boundary is now a default-deny `cityhall_gate` middleware beside `access_policy`. Every mutating request whose `(method, matched-path)` isn't in `CITYHALL_MUTATION_ALLOW` is 403'd before the handler — uniform across prefix + method, fails closed. The audit is now route-table exhaustiveness (`every_mutating_route_is_cityhall_classified`): it enumerates the router's mutating routes and asserts each is classified in exactly one of ALLOW/DENY, so a new route absent from both fails the build. Per-handler `cityhall_block*` kept as defense in depth.

Scoping note: the middleware fail-closes the mutation class (the recurring bug). Sensitive GET reads keep their per-handler gates (defense in depth) as before. The full type-state newtypes (sanitized-create, `Vec<StructuredTarget>` everywhere) are the scheduled follow-up you called out.

`cargo fmt`/`clippy`, 467 server+api unit tests (incl. the new classification audit + the read-only audit + access_policy), and 19 `serve_cityhall_lockdown` route tests all pass. Ready for re-review.

src/server/mod.rs
```diff
@@ -2284,6 +2295,170 @@ async fn access_policy(
...
+
+/// Default-deny CityHall reachability boundary. A no-op outside CityHall mode
+/// and for read methods (GET/HEAD/OPTIONS); for a mutating method it refuses any
+/// request whose matched-path template is not in [`CITYHALL_MUTATION_ALLOW`]
+/// with the canonical 403. This is the single choke point the reviewer asked
+/// for: it covers every module prefix and method uniformly (an unmatched or
+/// unlisted mutating route fails closed), so a handler can no longer silently
+/// reopen a hole by omission. See #7.
+#[cfg(feature = "serve")]
+async fn cityhall_gate(
```
jerome-benoit, line 2426: This is the change that closes the class. Default-deny (unlisted mutating (method, template) => 403), fail-closed on a missing `MatchedPath`, post-auth/pre-handler over the whole router. Crucially, runtime safety doesn't depend on the `#[cfg(test)]` audit parser: the gate reads the LIVE `MatchedPath` and denies by omission, so a scan blind spot can only over-restrict, never bypass. Verified no route is mounted outside this layer and no verb-tunneling/method-override middleware exists.

src/server/api/sessions.rs
```diff
@@ -4670,18 +4812,144 @@ pub(crate) fn run_create_hooks(
...
+        .find(|i| i.id == id)
+        .is_some_and(|i| i.is_structured());
+    (!is_structured_target).then(super::cityhall_response)
+}
+
+/// Plural [`cityhall_block_non_structured`]: refuse unless EVERY id resolves to
+/// a structured session this mode created. Used by multi-session teardown
+/// (`delete_workspace`), which acts on all ids, not just the owner. See #7.
+#[cfg(feature = "serve")]
+async fn cityhall_block_any_non_structured(
```
jerome-benoit, line 4858: F1 resolved: this checks EVERY id (`.all(is_structured)`) before `order_workspace_deletion`/`purge_workspace_artifacts` tear down the plan, so the round-4 exploit `session_ids: [own_structured, foreign_plain]` is now rejected (a foreign/unknown id makes `all_structured` false → 403). Nit: the route test exercises this on empty state (owner also unknown), so seeding a structured owner + foreign plain sibling would lock the exact owner-vs-sibling discrimination.

jerome-benoit (approved):
## Re-review round 5: CityHall lockdown — LGTM

This is the definitive fix. It resolves the round-4 design finding (default-allow deny-list) and the F1/F2/G3 gaps in one move, and I could not break it: 3 scoped reviewers plus an adversarial pass, backed by my own `--features serve` run (lockdown suite 19/19 incl. the new foreign-sibling + workspace-ordering-PUT cases, and the classification audit test passes).

**What makes it close the class (verified):**
- **`cityhall_gate` is a default-deny choke point.** In CityHall mode every mutating request (POST/PUT/PATCH/DELETE) whose `(method, MatchedPath)` is not in `CITYHALL_MUTATION_ALLOW` is 403'd before the handler; the default arm is deny, and a missing `MatchedPath` is deny (fail-closed). It runs post-auth, pre-handler, over the whole router (single `Router`, layer wraps everything, no `nest`/`merge`/route mounted outside), and is a no-op outside the mode.
- **Runtime safety does not depend on the audit parser.** The gate reads the live `MatchedPath` and denies by omission, so even a blind spot in the text-scan audit can only make a route *more* restricted (absent from ALLOW → 403), never a bypass. The audit is the maintainability belt; the gate is the suspenders.
- **The build-time audit is sound.** `every_mutating_route_is_cityhall_classified` enumerates every router mutating route (all methods incl. `put`, all module prefixes) and fails the build if any is unclassified, stale, or in both tables. Partition independently recomputed: 82 = 47 ALLOW + 35 DENY, 0 omission / 0 stale / 0 overlap. A new mutating route is now closed by default AND fails CI until deliberately classified.
- **ALLOW is clean:** no host-spawning / plugin / MCP / cross-session / global route is allowed; all must-reach routes (login, composer loop, create, own structured-session lifecycle, reads) are reachable. Adversarially re-checked the borderline entries (`resolve_options` is a registry-bounded probe; composer routes are structured-only via the worker-spawn invariant; `create_session` neutralizes every spawn field incl. the `command_override` argv[0] escape).
- **F1 resolved:** `delete_workspace` now gates *every* id (`cityhall_block_any_non_structured`, `.all(is_structured)`) before any teardown, so the crafted-sibling exploit is rejected.
- No regression from the -152 churn (only the old text-scan audit + a doc comment removed; the `cityhall_block*` helpers remain as defense-in-depth).

**Two non-blocking test-strength nits (LOW):**
1. `delete_workspace_with_foreign_sibling_is_blocked` runs on empty state, so the owner id is also unknown — it wouldn't fail on a revert to owner-only gating. The `.all()` logic is correct; a sharper test would seed a structured owner + a foreign plain sibling to lock the exact discrimination.
2. No positive "an ALLOW route stays reachable in CityHall" test (an accidental ALLOW-entry deletion regressing the composer relies on the classification audit rather than a direct 200 assertion).

Excellent work turning the whole class off at a single choke point. Approving.

src/server/mod.rs
Seluj78: Thanks for the thorough verification of the choke point (live MatchedPath, fail-closed, post-auth/pre-handler, no route mounted outside the layer). Nothing to change here.

src/server/api/sessions.rs
Seluj78: Good nit, done in https://github.com/agent-of-empires/agent-of-empires/pull/2853/commits/cbde05eb. The test now seeds a real structured owner + a foreign plain sibling, so it locks the owner-vs-sibling discrimination (an owner-only revert would let the pair through and fail this test). Added a companion asserting an all-structured workspace clears the gate, and a positive test that an allowlisted route (telemetry/consent) stays reachable.

njbrake (approved):

Approving. The load-bearing claim here is that the default-deny `cityhall_gate` middleware really sees a live `MatchedPath`, and that was checked against the axum 0.8.9 source rather than the PR description: `Router::layer` maps the layer onto every endpoint, `PathRouter::call_with_state` inserts `MatchedPath` before invoking that endpoint, `matched-path` is in axum's default feature set and this repo does not disable defaults, and `IS_FALLBACK` suppresses the insert so an unrouted mutating request fails closed. There is no `merge`, `nest`, or `route_service` anywhere in `src/server/mod.rs`, so no route is mounted outside the gate.

Moving from a handler-name allowlist to a middleware reading the live `MatchedPath` is the right inversion, and it is better than what the review thread asked for: it means the `#[cfg(test)]` scanner is no longer safety-critical, since a blind spot in it can only over-restrict, never bypass. Thanks to @jerome-benoit for four rounds that turned a genuinely leaky first draft into this.

I pushed four review follow-ups in 31dc48f1:

1. Settings search was built from the full schema, so typing "yolo" or "worktree" surfaced the field with a badge naming a tab that is not in the curated sidebar, and jumping to it clamped back to Theme. Search now indexes a curated schema, with the two field lists hoisted to consts the rendered tabs share so they cannot drift. The PATCH was already refused server-side, so this was a UX gate, not a security one.
2. `AOE_CITYHALL_MODE` was read three ways with `var().is_ok()` and once with `var_os().is_some()`. Those disagree on a non-UTF-8 value, in a fail-open direction: foreground `aoe serve` would record the mode ON in `serve.launch` while `AppState` and the config overrides both read it off. All four now use `var_os`.
3. Collapsed 20 single-input route tests into one table, per the test-cost guidance added in 0db32ba2. Every assertion is preserved and the 23 test functions become 4.
4. Dropped a stale `max_concurrent_resumes` mention left behind by b10f473a.

Two things I am deliberately not gating on, both filed as follow-ups:

**The read side has no default-deny and no exhaustiveness audit.** Mutations are default-deny with a build-time audit; GET reads are a hand-maintained per-handler list. Still reachable by a CityHall client: `GET /api/settings` and `/api/settings/resolved` (the full resolved config, including `session.on_create` hook command strings and host paths), `/api/projects`, `/api/recent-projects`, `/api/claude-sessions`, and `/api/agents`. This is information disclosure rather than host-code execution, and some of it is inherently visible since CityHall sessions span all projects, so it does not block. It will reopen the same way the mutation side did, though: a newly added sensitive GET is open by default with nothing failing the build.

**The blank-shell hold in `App.tsx` is unconditional.** Every user now gets a blank `h-dvh` div until `/api/about` resolves, not just CityHall users. It is the correct fix for the sidebar and Clone URL flash, and the early return is correctly placed after every hook so hook order stays stable, but on a tunnel or a mobile connection it replaces an immediate shell render with a blank screen for the round trip. That cost is paid globally for a mode most deployments will never enable. It is also why `App.tsx` patch coverage sits at 47%.

Last thing, for the record rather than as a request: e72e3aa5 ("stop an unrelated re-render cancelling the tips auto-pop") says in its own subject that it is unrelated. It is a real bug and the fix looks right (`useTips()` returns a fresh object each render, so the effect re-ran constantly and `cancelAnimationFrame` killed the pending open while the one-shot ref guard prevented rescheduling), so I reviewed it as maintainer rather than asking for a split.

PR #2860, opened 2026-07-15
refactor(test): consolidate env guards behind EnvGuard::set

### LINKED ISSUE #2716: test: consolidate 8 per-agent HomeGuard structs behind a single EnvGuard::set helper
### Summary

After #2306 landed `AppDirGuard` in [`src/session/test_support.rs`](src/session/test_support.rs), the codebase still carries 8 hand-written env-restoration guard structs across 5 distinct types. All of them share the same shape: snapshot `Option<String>` on construction, `std::env::set_var` on the new value, restore-or-remove on `Drop`. Consolidating them behind a single `EnvGuard::set(&[(key, value)])` helper would drop the duplicated boilerplate and give tests one place to reach for whenever they need to shim an env var.

### Existing guard sites

| Struct | Location | Keys covered |
| --- | --- | --- |
| `StorageHomeGuard` | [`src/session/sync.rs:346`](src/session/sync.rs) | `HOME`, `XDG_CONFIG_HOME` |
| `AppDirGuard` | [`src/session/test_support.rs`](src/session/test_support.rs) | `HOME`, `XDG_CONFIG_HOME` (+ owns tempdir) |
| `VibeHomeGuard` | [`src/session/capture.rs:2992`](src/session/capture.rs) | Vibe-specific home env |
| `GeminiHomeGuard` | [`src/session/capture.rs:3782`](src/session/capture.rs) | Gemini-specific home env |
| `ClaudeHomeGuard` | [`src/session/instance.rs:8411`](src/session/instance.rs) | Claude-specific home env |
| `CodexHomeGuard` (copy 1) | [`src/session/capture.rs:3501`](src/session/capture.rs) | `CODEX_HOME` |
| `CodexHomeGuard` (copy 2) | [`src/session/instance.rs:4635`](src/session/instance.rs) | `CODEX_HOME` |
| `CodexHomeGuard` (copy 3) | [`src/hooks/mod.rs:1764`](src/hooks/mod.rs) | `CODEX_HOME` |
| `CodexHomeGuard` (copy 4) | [`src/tui/dialogs/hooks_install.rs:315`](src/tui/dialogs/hooks_install.rs) | `CODEX_HOME` |

The four `CodexHomeGuard` copies are byte-for-byte duplicates.

### Proposed direction

Add one helper alongside `AppDirGuard` in [`src/session/test_support.rs`](src/session/test_support.rs):

```rust
pub(crate) struct EnvGuard {
    prev: Vec<(&'static str, Option<String>)>,
}

impl EnvGuard {
    pub(crate) fn set(pairs: &[(&'static str, &str)]) -> Self { /* snapshot + set */ }
}

impl Drop for EnvGuard {
    fn drop(&mut self) { /* set_var when Some, remove_var when None */ }
}
```

Every guard site collapses to `let _guard = EnvGuard::set(&[("CODEX_HOME", path)]);` (or `("HOME", ...), ("XDG_CONFIG_HOME", ...)` for the pair). `AppDirGuard` continues to own its `TempDir` and delegates the env work to an inner `EnvGuard`.

### Out of scope for #2306

The RAII fix for `isolate_app_dir` is a narrow test-only refactor. Consolidating the eight other guard sites touches four more source files and a wider slice of the test surface, so it belongs in its own PR.

### Acceptance

- Single `EnvGuard` helper in `src/session/test_support.rs`.
- All 8 hand-written guard structs removed; every call site uses `EnvGuard::set(...)`.
- `AppDirGuard` retains its tempdir ownership and its `path()` accessor.
- `cargo fmt` + `cargo clippy --all-targets --features serve -- -D warnings` + `cargo test --features serve` clean.
- `#[serial]` annotations unchanged on every affected test.

### LINKED ISSUE #2751: test: `EnvGuard::set` must snapshot `Option<OsString>` to preserve non-UTF-8 env values — #2716 constraint
### Summary

The `EnvGuard::set` consolidation proposed in [#2716](https://github.com/agent-of-empires/agent-of-empires/issues/2716) currently sketches `prev: Vec<(&'static str, Option<String>)>`. A non-UTF-8 prior value returns `Err(NotUnicode(_))` from `std::env::var(...)`, `.ok()` collapses to `None`, and `Drop` calls `remove_var` instead of restoring the original bytes. Under `#[serial]`, that leaks a removed env var into every later test in the same process.

[`AppDirGuard`](src/session/test_support.rs) landed by PR #2717 already fixed this shape for its own snapshot fields: [`prev_home: Option<OsString>`](src/session/test_support.rs) via [`std::env::var_os`](https://doc.rust-lang.org/std/env/fn.var_os.html). `StorageHomeGuard` at [`src/session/sync.rs:346-378`](src/session/sync.rs) still carries the pre-fix `Option<String>` shape and reproduces the leak; the `EnvGuard` consolidation must not repeat it.

### Proposed direction

When implementing the `EnvGuard::set` helper for #2716:

1. Type snapshot fields as `Option<OsString>`, not `Option<String>`.
2. Read the prior value via `env::var_os(...)` (which returns `Option<OsString>`), not `env::var(...).ok()`.
3. Restore via `set_var(key, os_str)` on `Some(_)` and `remove_var(key)` on `None`.

### Acceptance

- `EnvGuard` snapshot type is `Option<OsString>`.
- A locking test seeds a non-UTF-8 `HOME` (via `OsStr::from_bytes(&[0xFF, ...])` on Unix), guards + drops, asserts byte-identical restoration.

### Context

Filed from the review of merged PR #2717. Constraint on the implementation of the open follow-up #2716.

njbrake:
## Description

The crate carried **nine** hand-written env-restoration guard structs across six files, all the same shape: snapshot the prior value, `set_var` the new one, restore-or-remove on `Drop`. Four of the `CodexHomeGuard` copies were byte-for-byte identical.

This replaces them with a single `EnvGuard` in `src/session/test_support.rs`:

```rust
let _guard = EnvGuard::set(&[("CODEX_HOME", tmp.path())]);
let _guard = EnvGuard::unset(&["CODEX_HOME"]);
```

Guards folded in (the 8 from #2716, plus one extra found on the way):

| Struct | Location |
| --- | --- |
| `StorageHomeGuard` | `src/session/sync.rs` |
| `VibeHomeGuard` / `GeminiHomeGuard` / `CodexHomeGuard` | `src/session/capture.rs` |
| `ClaudeHomeGuard` / `CodexHomeGuard` | `src/session/instance.rs` |
| `StorageHomeGuard` / `EnvVarGuard` (added by #2837) | `src/session/instance.rs` |
| `CodexHomeGuard` + a local `Option<String>` `EnvGuard` | `src/hooks/mod.rs` |
| `CodexHomeGuard` | `src/tui/dialogs/hooks_install.rs` |

The local `EnvGuard` in `src/hooks/mod.rs` was not in #2716's table; it collided by name with the new helper and carried the same pre-fix shape, so it is folded in here. It is one of the sites #2751's status comment flagged.

`AppDirGuard` keeps its `TempDir` ownership and `path()` accessor and delegates the env work to an inner `EnvGuard`. Its hand-written `Drop` is gone: field declaration order (`_env`, `path`, `_temp`) already sequences the env restore ahead of the tempdir's deletion, which is the invariant the old `Drop` enforced by hand. `HomeGuard` becomes an alias for `EnvGuard`.

The `restore_or_remove` SAFETY comment is re-scoped to match. It previously named `AppDirGuard` as the only constructor and grep-claimed that non-`#[serial]` tests do not read `HOME` / `XDG_CONFIG_HOME` / `XDG_DATA_HOME`. `EnvGuard` widened that to `CODEX_HOME` / `VIBE_HOME` / `GEMINI_CLI_HOME` / `CLAUDE_CONFIG_DIR` / `AOE_MOUSE_CAPTURE` plus any caller-chosen key, so no fixed grep list can bound which readers might race; the comment now states the `#[serial]` obligation on every call site as the load-bearing rule.

### Closes #2751: `Option<OsString>`, not `Option<String>`

Snapshots are typed `Option<OsString>` and read via `std::env::var_os`, never `env::var(...).ok()`. A non-UTF-8 prior value makes `env::var` return `Err(NotUnicode(_))`; the old `.ok()` snapshot collapsed that to `None`, so `Drop` called `remove_var` instead of restoring the original bytes, leaking the removal into every later `#[serial]` test in the same process.

Two smaller semantics that fall out of the consolidation and are now pinned by tests:

- **empty vs unset**: `var_os` returns `Some("")` for an empty value, so `Drop` restores it as empty rather than unsetting it.
- **duplicate keys**: `Drop` restores in reverse, so a key listed twice in one `set` call round-trips to its pre-guard value rather than to the intermediate write.

Behaviour and every `#[serial]` annotation are unchanged; this is test-support only, with no production-code changes.

### All 8 of #2751's sites are covered

#2751's status comment enumerates 8 leaking `Option<String>` guard sites. The consolidation above covers 2 of them (`src/session/sync.rs`, `src/hooks/mod.rs`); the second commit covers the remaining 6, so `Closes #2751` is accurate rather than aspirational.

| #2751 site | Treatment |
| --- | --- |
| `src/session/sync.rs` | folded into `EnvGuard` (consolidation commit) |
| `src/hooks/mod.rs` | folded into `EnvGuard` (consolidation commit) |
| `src/tui/mod.rs` | migrated to `EnvGuard` |
| `src/migrations/v015_rewrite_hook_strings.rs` | migrated to `EnvGuard` |
| `src/migrations/v017_rewrite_hook_strings_for_per_user_base.rs` | migrated to `EnvGuard` |
| `src/migrations/v018_strip_codex_config_toml_hooks.rs` | migrated to `EnvGuard` |
| `tests/filewatch_degradation.rs` | fixed in place (`var_os`) |
| `tests/integration/acp_silent_orphan.rs` | fixed in place (`var_os`) |

**The `unset_all()`-shaped guards** (`v015` / `v017` / `v018`) were three byte-identical structs unsetting the same 5 keys. Each collapses to a local `unset_agent_home_env()` returning `EnvGuard::unset` over that key list, preserving the "unset these 5, restore on drop" semantics. The keys are distinct, so `EnvGuard`'s reverse-order restore is equivalent to the old forward-order loop. `src/tui/mod.rs`'s guard is a set-or-unset shape over `AOE_MOUSE_CAPTURE`: `set(None)` becomes `EnvGuard::unset`, `set(Some(v))` becomes `EnvGuard::set`.

**The two integration-test crates** are separate crates and cannot reach `pub(crate)` `test_support`. Rather than widen `EnvGuard` to `pub` purely to reach test code, or add a shared crate for two call sites, their guards are fixed in place: they now snapshot `Option<OsString>` via `env::var_os` and restore unset-as-unset. That resolves the #2751 bug at both sites without the consolidation, which is the part of #2751 that is actually a bug. Each carries a comment recording why it is duplicated rather than shared.

### Rebased onto #2837: its new guard is absorbed, not left behind

#2837 (per-tool verify-on-resume tests for the six live agents) merged to main
while this PR was open. It hand-rolled a **new** `StorageHomeGuard` in
`src/session/instance.rs` that snapshots `Option<String>` via `env::var`, which
is precisely the #2751 bug this PR closes. Rebasing produced no textual
conflict, so that guard would have survived silently and #2751 would have closed
with a fresh copy of its own bug in the tree. It is folded in here instead:

| #2837 site | Treatment |
| --- | --- |
| `StorageHomeGuard` (6 call sites) | replaced by `test_support::isolate_app_dir_at` |
| `EnvVarGuard` (7 call sites, pre-existing, `var_os`-based) | collapsed into `EnvGuard::set` |
| local `fn restore_or_remove(_, Option<String>)` | deleted with its last caller |

Notes on the absorption:

- **`XDG_DATA_HOME`**: `StorageHomeGuard` set only `HOME` and `XDG_CONFIG_HOME`;
  `isolate_app_dir_at` also sets `XDG_DATA_HOME`, which is the stricter
  isolation. All six tests pass with it: the opencode test resolves `OPENCODE_DB`
  before `XDG_DATA_HOME` (`capture.rs` checks the explicit override first), and
  the other five key off their own tool-specific home var, so none of them reads
  `XDG_DATA_HOME`.
- **Drop order preserved**: each test still declares its `TempDir` before the
  guard, so the guard drops first (env restore) and only then is the tempdir
  removed. `HOME` never points at a directory being deleted.
- **No test weakened**: all six tests (opencode, vibe, codex, gemini, pi, hermes)
  keep their assertions byte-for-byte; only the env-guard plumbing changed.
- **`#[serial]` unchanged**: `src/session/instance.rs` carries 78 annotations on
  both main and this branch, so all of #2837's are intact.

### Out of scope

- Inline `let old = env::var(..).ok(); ... restore` blocks in test bodies (`src/file_watch.rs`, `src/update/install.rs`, `src/acp/acp_client.rs`, `src/session/capture.rs`). These are not guard structs, are not in #2716's table, and are not in #2751's enumeration; they carry the same snapshot-type bug in miniature and are a clean follow-up.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Docs: the `test_support` module doc is rewritten to describe `EnvGuard` as the single entry point and to record the `var_os` rationale. No user-facing docs are affected.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Five new unit tests in `src/session/test_support.rs` cover the new helper:

- `env_guard_drop_restores_non_utf8_prior_value` (the #2751 acceptance test; `#[cfg(unix)]`, seeds `HOME` via `OsStr::from_bytes(b"/tmp/aoe-\xFF\xFE-home")`, asserts byte-identical restoration and asserts as a precondition that `env::var` really does error on it)
- `env_guard_drop_preserves_empty_versus_unset`
- `env_guard_unset_removes_then_restores`
- `env_guard_drop_restores_duplicate_key_to_pre_guard_value`
- plus the pre-existing `AppDirGuard` suite, which now exercises the delegated path

### How I tested

- `cargo fmt`: clean (no-op).
- `cargo clippy --lib --tests`: no warnings from any file in this diff. The three that remain (`src/session/config.rs`, `src/tui/home/mod.rs`, and the `inject_test_update` dead-code warning in `src/session/poller.rs`) are pre-existing on `main` and untouched here. `cargo clippy --features serve --test integration` is also clean, which is the run that actually compiles `tests/integration/acp_silent_orphan.rs` (it is gated behind `cfg(all(feature = "serve", debug_assertions))`).
- Scoped tests for the second commit, all green: `--lib test_support:: mouse_capture_tests::` (15 passed), `--lib migrations::v015 migrations::v017 migrations::v018` (31 passed), `--test filewatch_degradation` (1 passed), `--features serve --test integration -- acp_silent_orphan` (6 passed).
- Every `#[serial]` / `#[serial_test::serial(...)]` annotation is unchanged. Verified mechanically per-file against the base: the second commit's net delta across `src/` and `tests/` is 0. The only annotation change in the whole PR is the +4 from the consolidation commit's own new `EnvGuard` unit tests. Env mutation is process-global, so dropping one would reintroduce the #2600 race class.
- `cargo test --lib`: **3862 passed, 3 failed**. The 3 failures (`resume_intent_default_uses_observed`, `test_update_write_failure_emits_no_notify`, `restart_selected_session_surfaces_resume_failed_after_async_restart`) reproduce identically on a clean checkout of `origin/main` at `3f0ee18`, both in a full run and in isolation, so they are pre-existing and unrelated to this change. Worth a look independently of this PR.
- After rebasing onto #2837: `cargo fmt` clean, `cargo clippy --lib --tests` clean (same three pre-existing warnings, none from this diff). Re-ran the affected scopes green: `verify_on_resume` (17 passed, includes #2837's six), `test_support::` (18 passed), and `migrations::v015 v017 v018`, for 59 passed / 0 failed over the combined set.
- **Pre-existing flaky race found while verifying, not introduced here.** Running `--lib verify_on_resume migrations::v015 v017 v018` in one process makes `verify_on_resume::mtime_fallback_applies_without_sidecar` fail intermittently. The migration tests are `#[serial_test::serial(shell_env)]` (a named group) while the instance tests are plain `#[serial]` (the default group); serial_test does not serialise a named group against the default one, so the migration tests `remove_var("CLAUDE_CONFIG_DIR")` while the Claude test is mid-scan, and the scan then finds no fresher jsonl. It reproduces on a clean `origin/main` checkout at `faf5ce0e` (8 of 10 runs failed) and disappears entirely under `--test-threads=1` (3 of 3 clean), on main and on this branch alike. Out of scope here since fixing it means changing `#[serial]` groups, which this PR deliberately leaves untouched; worth its own issue.

- The `--features serve` matrix is left to CI; this diff is TUI/session test-support only and touches no serve-gated code.

Closes #2716
Closes #2751

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.8 via Claude Code

Claude did the mechanical sweep across the six files and drafted the helper, its docs, and the new tests. The compiler caught missed call sites; the pre-existing-failure claim above was verified by re-running the three tests against a clean `origin/main`.

## Summary by CodeRabbit

* **Tests / test utilities**
  * Consolidated nine duplicated, hand-written environment restoration guards into a single shared `EnvGuard` helper in `src/session/test_support.rs`, with new `EnvGuard::set` and `EnvGuard::unset` APIs.
  * Refactored test callers across hooks/session/tui and multiple migration test suites to use the shared helper and removed the local guard implementations they replaced.
* **Env snapshot correctness**
  * Updated environment snapshots to store `Option<OsString>` via `env::var_os`, preserving non-UTF-8 values and distinguishing empty values from unset variables.
  * Ensured scoped restoration is correct even when the same key is provided multiple times in a single `set` call (restores the pre-guard value).
  * Added/expanded tests covering non-UTF-8 round-tripping, empty-vs-unset behavior, unset round-tripping, and duplicate-key handling.
* **App dir / home isolation**
  * Updated `AppDirGuard` to delegate env handling to `EnvGuard` while retaining its temp directory ownership/lifetime behavior and `path()` accessor.
  * Simplified `HomeGuard` by making it an alias of `EnvGuard`.

* **Behavioral impact**
  * No production logic or user-facing behavior changes; the refactor is limited to test helpers and test behavior around env isolation.

### DIFF
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -1897,6 +1897,7 @@ pub fn uninstall_all_hooks() {
 mod tests {
     use super::targets::collect_env_lists_from_session;
     use super::*;
+    use crate::session::test_support::EnvGuard;
     use tempfile::TempDir;
 
     fn claude_events() -> Vec<crate::agents::ResolvedHookEvent> {
@@ -1940,29 +1941,6 @@ mod tests {
             .events
     }
 
-    struct CodexHomeGuard(Option<String>);
-    impl CodexHomeGuard {
-        fn set(path: &Path) -> Self {
-            let prev = std::env::var("CODEX_HOME").ok();
-            std::env::set_var("CODEX_HOME", path);
-            Self(prev)
-        }
-
-        fn unset() -> Self {
-            let prev = std::env::var("CODEX_HOME").ok();
-            std::env::remove_var("CODEX_HOME");
-            Self(prev)
-        }
-    }
-    impl Drop for CodexHomeGuard {
-        fn drop(&mut self) {
-            match &self.0 {
-                Some(v) => std::env::set_var("CODEX_HOME", v),
-                None => std::env::remove_var("CODEX_HOME"),
-            }
-        }
-    }
-
     fn claude_hook_config() -> &'static crate::agents::AgentHookConfig {
         crate::agents::get_agent("claude")
             .unwrap()
@@ -1971,36 +1949,10 @@ mod tests {
             .unwrap()
     }
 
-    struct EnvGuard {
-        key: &'static str,
-        prev: Option<String>,
-    }
-    impl EnvGuard {
-        fn set(key: &'static str, value: &str) -> Self {
-            let prev = std::env::var(key).ok();
-            std::env::set_var(key, value);
-            Self { key, prev }
-        }
-
-        fn unset(key: &'static str) -> Self {
-            let prev = std::env::var(key).ok();
-            std::env::remove_var(key);
-            Self { key, prev }
-        }
-    }
-    impl Drop for EnvGuard {
-        fn drop(&mut self) {
-            match &self.prev {
-                Some(v) => std::env::set_var(self.key, v),
-                None => std::env::remove_var(self.key),
-            }
-        }
-    }
-
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_agent_settings_path_defaults_to_home_relative() {
-        let _guard = EnvGuard::unset("CLAUDE_CONFIG_DIR");
+        let _guard = EnvGuard::unset(&["CLAUDE_CONFIG_DIR"]);
         let path = agent_settings_path_for_host_environment(claude_hook_config(), &[]).unwrap();
         let expected = dirs::home_dir().unwrap().join(".claude/settings.json");
         assert_eq!(path, expected);
@@ -2009,7 +1961,7 @@ mod tests {
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_agent_settings_path_honors_host_env_override() {
-        let _guard = EnvGuard::unset("CLAUDE_CONFIG_DIR");
+        let _guard = EnvGuard::unset(&["CLAUDE_CONFIG_DIR"]);
         let host_env = vec!["CLAUDE_CONFIG_DIR=/home/me/.claude-work".to_string()];
         let path =
             agent_settings_path_for_host_environment(claude_hook_config(), &host_env).unwrap();
@@ -2022,7 +1974,7 @@ mod tests {
     #[serial_test::serial(shell_env)]
     fn test_agent_settings_path_host_env_takes_precedence_over_process_env() {
         // When both are set, the session's profile env wins over AoE's own env.
-        let _guard = EnvGuard::set("CLAUDE_CONFIG_DIR", "/from/process/env");
+        let _guard = EnvGuard::set(&[("CLAUDE_CONFIG_DIR", "/from/process/env")]);
         let host_env = vec!["CLAUDE_CONFIG_DIR=/from/host/env".to_string()];
         let path =
             agent_settings_path_for_host_environment(claude_hook_config(), &host_env).unwrap();
@@ -2034,7 +1986,7 @@ mod tests {
     fn test_agent_settings_path_falls_back_to_process_env() {
         // Not present in the host env list at all, but set in AoE's own env:
         // the launched agent inherits it, so hooks must follow.
-        let _guard = EnvGuard::set("CLAUDE_CONFIG_DIR", "/tmp/claude-proc");
+        let _guard = EnvGuard::set(&[("CLAUDE_CONFIG_DIR", "/tmp/claude-proc")]);
         let path = agent_settings_path_for_host_environment(claude_hook_config(), &[]).unwrap();
         assert_eq!(path, PathBuf::from("/tmp/claude-proc/settings.json"));
     }
@@ -2042,7 +1994,7 @@ mod tests {
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_agent_settings_path_display_matches_resolution() {
-        let _guard = EnvGuard::unset("CLAUDE_CONFIG_DIR");
+        let _guard = EnvGuard::unset(&["CLAUDE_CONFIG_DIR"]);
 
         // Default: tilde-relative, matching how the path is shown elsewhere.
         assert_eq!(
@@ -2061,7 +2013,7 @@ mod tests {
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_agent_settings_path_empty_override_is_ignored() {
-        let _guard = EnvGuard::unset("CLAUDE_CONFIG_DIR");
+        let _guard = EnvGuard::unset(&["CLAUDE_CONFIG_DIR"]);
         let host_env = vec!["CLAUDE_CONFIG_DIR=".to_string()];
         let path =
             agent_settings_path_for_host_environment(claude_hook_config(), &host_env).unwrap();
@@ -2420,7 +2372,7 @@ mod tests {
     #[serial_test::serial]
     fn test_codex_config_path_respects_codex_home() {
         let tmp = TempDir::new().unwrap();
-        let _guard = CodexHomeGuard::set(tmp.path());
+        let _guard = EnvGuard::set(&[("CODEX_HOME", tmp.path())]);
 
         assert_eq!(codex_config_path().unwrap(), tmp.path().join("config.toml"));
         assert_eq!(
@@ -2433,7 +2385,7 @@ mod tests {
     #[serial_test::serial]
     fn test_codex_config_path_for_host_environment_ignores_empty_codex_home() {
         let tmp = TempDir::new().unwrap();
-        let _guard = CodexHomeGuard::unset();
+        let _guard = EnvGuard::unset(&["CODEX_HOME"]);
         std::env::set_var("HOME", tmp.path());
 
         // An empty `CODEX_HOME=` must not resolve to a bare relative
@@ -2457,7 +2409,7 @@ mod tests {
         let tmp = TempDir::new().unwrap();
         // An empty `CODEX_HOME` in AoE's own process env must fall back to the
         // home-relative default rather than a bare relative `config.toml`.
-        let _guard = CodexHomeGuard::set(Path::new(""));
+        let _guard = EnvGuard::set(&[("CODEX_HOME", Path::new(""))]);
         std::env::set_var("HOME", tmp.path());
 
         let path = codex_config_path().unwrap();
@@ -2472,7 +2424,7 @@ mod tests {
     #[serial_test::serial]
     fn test_iter_hook_targets_includes_profile_codex_home() {
         let tmp = TempDir::new().unwrap();
-        let _guard = CodexHomeGuard::unset();
+        let _guard = EnvGuard::unset(&["CODEX_HOME"]);
         std::env::set_var("HOME", tmp.path());
         #[cfg(any(target_os = "linux", target_os = "macos"))]
         std::env::set_var("XDG_CONFIG_HOME", tmp.path().join(".config"));
@@ -4811,13 +4763,10 @@ hooks_auto_accept: false
     #[serial_test::serial(shell_env)]
     fn collect_env_lists_warns_on_corrupt_global_config_but_not_on_missing() {
         let tmp = TempDir::new().unwrap();
-        let _codex = CodexHomeGuard::unset();
-        let _home = EnvGuard::set("HOME", tmp.path().to_str().unwrap());
+        let _codex = EnvGuard::unset(&["CODEX_HOME"]);
+        let _home = EnvGuard::set(&[("HOME", tmp.path())]);
         #[cfg(any(target_os = "linux", target_os = "macos"))]
-        let _xdg = EnvGuard::set(
-            "XDG_CONFIG_HOME",
-            tmp.path().join(".config").to_str().unwrap(),
-        );
+        let _xdg = EnvGuard::set(&[("XDG_CONFIG_HOME", tmp.path().join(".config"))]);
 
         let app_dir = crate::session::get_app_dir().unwrap();
         let config_path = app_dir.join("config.toml");
--- a/src/migrations/v015_rewrite_hook_strings.rs
+++ b/src/migrations/v015_rewrite_hook_strings.rs
@@ -218,6 +218,7 @@ fn read_environment_from_toml(path: &Path) -> Option<Vec<String>> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::session::test_support::EnvGuard;
     use serde_json::Value;
     use std::fs;
     use tempfile::TempDir;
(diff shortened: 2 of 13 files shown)


jerome-benoit (approved):
Reviewed with several cross-validation rounds (EnvGuard correctness + Drop semantics, every call-site migration, the `#[serial]` delta, the #2837 XDG_DATA_HOME widening, and a local build+test run). This is a clean, well-scoped test-support refactor that correctly closes #2716 and #2751, and it verifies against every claim in the description. Approving; a couple of non-blocking nits below.

## Verified

- **EnvGuard is correct.** Snapshots are `Option<OsString>` via `std::env::var_os` (never `env::var().ok()`), `Drop` restores via `set_var` on `Some` / `remove_var` on `None`, and it restores in reverse (`drain(..).rev()`) so a key listed twice round-trips to its pre-guard value. The #2751 non-UTF-8 leak is genuinely fixed, with a `#[cfg(unix)]` acceptance test that seeds a non-UTF-8 `HOME` and asserts byte-identical restoration.
- **No behavior change / no production code touched.** Every hunk across the 13 files lands in a `#[cfg(test)]` block or a test file (`src/session/poller.rs` is a doc-comment-only edit). No `#[serial]` annotation was dropped and none had its group changed (the two integration test files have zero net `#[serial]` delta; `test_support.rs` only gains the new EnvGuard unit tests), so the #2600 env-race class is not reopened.
- **AppDirGuard Drop-order invariant holds.** Removing the hand-written `Drop` in favor of field declaration order (`_env`, `path`, `_temp`) still restores the env before the tempdir is deleted, because Rust drops fields in declaration order. Equivalent to the old explicit `Drop`, and pinned by the existing `app_dir_guard_*` tests.
- **#2837 XDG_DATA_HOME widening is safe.** `isolate_app_dir_at` additionally sets `XDG_DATA_HOME`, but that var is read in exactly one place (`opencode_data_dir()` in `src/session/capture.rs`, reached only after the `OPENCODE_DB` check fails, which the opencode test sets to an absolute path). The other five tools resolve via their own home var, so the stricter isolation cannot divert resolution. All six verify-on-resume tests keep their assertions.
- **Migration unset equivalence holds.** v015/v017/v018 each unset the same 5 distinct keys; reverse-order restore is equivalent to the old forward loop for distinct keys.
- **Ran locally on the PR head:** `cargo fmt --check` clean; the four new EnvGuard unit tests + `test_support::` + `mouse_capture_tests::` + `migrations::v015/v017/v018` pass (46 lib tests); `tests/filewatch_degradation.rs` passes; `cargo test --features serve --test integration -- acp_silent_orphan` passes (6). No emdash / ` -- ` prose separators in the added lines.

## Nits (non-blocking)

- `tests/integration/acp_silent_orphan.rs`: for consistency with `tests/filewatch_degradation.rs` (which wraps them) and forward-compatibility with the eventual edition-2024 move, wrap the `set_var`/`remove_var` calls in `unsafe {}` with a SAFETY comment. They are safe in the current edition 2021, so this is not a correctness issue; it also matches the still-open CodeRabbit nitpick on that file.
- Same file: its local `EnvGuard::Drop` restores forward rather than reverse like the canonical helper. Harmless for its single distinct-key call site, but matching the canonical `drain(..).rev()` would guard a future duplicate-key caller. (This predates the PR; the PR only changed the snapshot type here.)
- `AppDirGuard`: the field-order invariant is load-bearing but enforced only by a comment. A one-line note pointing at Rust's declaration-order drop guarantee (RFC 1857), or a marker that trips if a field is inserted between `_env` and `_temp`, would make it self-documenting. Optional.

## Follow-ups (pre-existing, out of scope, already noted in the description)

- The ~30 inline `let old = env::var(..).ok(); ... restore` blocks in `src/session/capture.rs` (CLAUDE_CONFIG_DIR / PI_CODING_AGENT_DIR / OPENCODE_DB / SHELL) still carry the #2751 shape. They are all `#[serial]`-annotated so the race is contained, and they are inline test-body blocks rather than guard structs, so deferring them is reasonable. Worth a follow-up issue.

Nice, thorough refactor, and thank you for the exceptionally clear description and the pre-existing-failure/flake notes.

PR #2862, opened 2026-07-15
fix(tui): show Tab as Enter's attach/live-send complement in the footer

hamid-mouti:
## Description

On a session row in the Agent view, Enter's behavior is governed by
`[session] default_attach_mode`: it either attaches to the tmux pane
(`Tmux`, the historical default) or enters live-send mode (`LiveSend`).
Tab is coded as the explicit complement — it does whichever of the two
Enter *isn't* currently doing (`src/tui/home/input.rs`, the `KeyCode::Tab`
handler and `tab_attach_action`).

The persistent footer bar only ever advertised the Enter side, and
always with the static label `"Attach"` regardless of which mode was
active. Tab had zero footer presence, so its behavior was only
discoverable by reading the source or opening the `?` help overlay
(which already computes this correctly via `help_live_on_enter` /
`HelpOverlay`'s `shortcuts()`).

This PR adds a Tab hint to the footer, mirroring the wording the help
overlay already uses (`"Attach"` / `"Live"`), so the pairing is
discoverable without digging into the code. It intentionally leaves the
Enter label unchanged ("Attach" is arguably correct either way — both
modes are informally "attaching" to the session).

The hint is suppressed on Acp/structured-view session rows: those
ignore `default_attach_mode` entirely (Tab either mirrors Enter's
`OpenStructuredView` or no-ops), so there's nothing meaningful to
advertise there.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Footer before (always static, no Tab presence):
`↵ Attach · t View · g Group · ...`

Footer after, `default_attach_mode = LiveSend` (Enter → live mode, Tab → tmux attach):
`↵ Attach · ⇥ Attach · t View · g Group · ...`

Footer after, `default_attach_mode = Tmux` (the default; Enter → tmux attach, Tab → live mode):
`↵ Attach · ⇥ Live · t View · g Group · ...`

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered by three targeted unit tests in `src/tui/home/tests.rs::default_attach_mode` that render the footer through `TestBackend` and assert on the rendered hint text (`footer_advertises_tab_as_live_when_default_is_tmux`, `footer_advertises_tab_as_attach_when_default_is_live_send`, `footer_hides_tab_hint_for_structured_sessions`). A full e2e is unnecessary overhead for a footer-hint text change already covered at the render level.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Sonnet 5)

**Any Additional AI Details you'd like to share:** Investigated at the user's request after they noticed Tab attaches to tmux with no footer hint advertising it; confirmed the Enter/Tab pairing is intentional design (`default_attach_mode` + its complement), scoped the fix to just the missing Tab discoverability rather than relabeling Enter, and verified with `cargo fmt`, `cargo clippy -D warnings` (with and without `--features serve`), and `cargo test --lib` (3864 passed).

## Summary

- Updated the TUI session-row footer to show a Tab shortcut hint (`⇥`) on applicable (non-structured) sessions.
- The footer now complements Enter’s label based on `default_attach_mode`:
  - In **Tmux** mode: **Enter = Attach**, **Tab = Live**
  - In **LiveSend** mode: **Enter = Live**, **Tab = Attach**
- Ensured Enter and Tab labels are resolved together so they don’t end up advertising the same action.
- Kept Tab hint hidden for **structured/ACP** sessions while preserving the existing Enter “Attach” label.
- Added focused regression tests covering both attach modes and structured sessions.

## Benefits

- Makes the Tab action easier to discover without changing existing Enter behavior.
- Improves consistency with the help overlay’s guidance and clarifies the intended key behavior per session type.

### DIFF
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -2963,6 +2963,7 @@ impl HomeView {
         let kc = |c: char| Some(KeyEvent::new(KeyCode::Char(c), KeyModifiers::NONE));
         let kctrl = |c: char| Some(KeyEvent::new(KeyCode::Char(c), KeyModifiers::CONTROL));
         let kenter = Some(KeyEvent::new(KeyCode::Enter, KeyModifiers::NONE));
+        let ktab = Some(KeyEvent::new(KeyCode::Tab, KeyModifiers::NONE));
 
         // (priority, click-key, spans). `click-key` is `None` for the
         // non-actionable status indicators (Serve / watching), which render
@@ -3040,6 +3041,30 @@ impl HomeView {
             // glyph fills its cell tightly and a single mk-internal space
             // looks too close to the desc.
             groups.push((0, kenter, mk("↵ ", enter_action_text)));
+        }
+
+        // Tab hint: on a non-Acp session row, Tab is Enter's complement
+        // (whichever of live-send / tmux-attach `default_attach_mode`
+        // doesn't route to Enter). Acp sessions ignore the setting (Enter
+        // and Tab both open the structured view, or Tab no-ops), so no
+        // hint is shown for those. Mirrors the wording `HelpOverlay` uses
+        // for the same pairing (`src/tui/components/help.rs`).
+        if let Some(Item::Session { id, .. }) = self.flat_items.get(self.cursor) {
+            let is_structured = self
+                .get_instance(id)
+                .map(|inst| inst.is_structured())
+                .unwrap_or(false);
+            if !is_structured {
+                let tab_action_text = if matches!(
+                    self.default_attach_mode(id),
+                    Some(crate::session::NewSessionAttachMode::LiveSend)
+                ) {
+                    "Attach"
+                } else {
+                    "Live"
+                };
+                groups.push((1, ktab, mk("⇥ ", tab_action_text)));
+            }
         }
 
         groups.push((
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -3024,16 +3024,38 @@ impl HomeView {
             }
         }
 
-        if let Some(enter_action_text) = match self.flat_items.get(self.cursor) {
+        // On a session row Enter and Tab are complements: `default_attach_mode`
+        // routes Enter to live-send or tmux attach, and Tab does the other one.
+        // Both labels resolve here so they can never advertise the same action
+        // twice. Acp rows ignore the setting entirely (Enter opens the
+        // structured view; Tab mirrors it or no-ops), so they keep the plain
+        // "Attach" label and advertise no complement. Mirrors the wording
+        // `HelpOverlay` uses for the same pairing (`src/tui/components/help.rs`).
+        let (enter_action_text, tab_action_text) = match self.flat_items.get(self.cursor) {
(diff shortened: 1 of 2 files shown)


njbrake:

Thanks for this. The Enter/Tab pairing genuinely was undiscoverable, and the scoping is good. I verified the hint logic against the dispatch in `input.rs` and it matches, including the structured-row suppression: `start_live_send` returns `None` for `is_structured`, and `tab_attach_action` routes those rows to `OpenStructuredView`, so there is no complement to advertise there. Good call leaving them out.

I did push one change on top, and since it reverses a decision you made deliberately, I would rather explain it here than let you find it in the diff.

You left the Enter label static, reasoning that "Attach" is arguably correct either way. I rendered the footer in both modes to see what that produces:

```
default_attach_mode = Tmux      ↵  Attach · ⇥  Live      (your change, correct)
default_attach_mode = LiveSend  ↵  Attach · ⇥  Attach    (two identical labels)
```

In LiveSend mode the two adjacent keys advertise the same action while doing different things: Enter enters live-send, Tab attaches to tmux. So for the people who opted into that setting, who are the ones most likely to need the hint, the footer teaches nothing. `HelpOverlay` already swaps both descriptions for this pairing (`help.rs:36-40`), so the footer was the last surface not doing it.

The commit resolves both labels in a single match, so they cannot claim the same action twice:

```
default_attach_mode = Tmux      ↵  Attach · ⇥  Live
default_attach_mode = LiveSend  ↵  Live   · ⇥  Attach
```

That also folded in CodeRabbit's `is_some_and` suggestion and dropped the duplicated `is_structured` computation, since both labels now resolve in one place.

On the tests: `footer_advertises_tab_as_attach_when_default_is_live_send` asserted `!out.contains("Live")`, which also passes if the Tab hint is missing entirely, and it had to flip regardless once Enter says "Live". All three now assert on both chips (`↵  Attach` / `⇥  Live`). Worth noting the labels render with two spaces after the glyph via `mk`, so a single-space assertion fails silently. I took CodeRabbit's `!out.contains("⇥")` for the structured case, which is stronger than checking for the word.

One other thing worth knowing: CI had never run on this PR. `CI` and `Tests` were sitting in `action_required` awaiting workflow approval for a first-time contributor, so the only green checks were the title and template linters, which do not compile anything. I approved the runs, and the full suite is now green (25 checks passing).

If you disagree with the Enter relabel, say the word and I will drop it. The rest stands on its own.

njbrake (approved):

Verified the Enter/Tab hint logic against the actual dispatch in `input.rs`, including the structured-row suppression (`start_live_send` returns `None` for `is_structured`, and `tab_attach_action` routes those rows to `OpenStructuredView`, so there is genuinely no complement to advertise).

The full suite is green after approving the gated workflow runs, which had never executed on this PR. Rationale for the Enter relabel pushed on top is in the comment above; disagreement there is welcome and easy to undo.

Issue #2758, opened 2026-07-09
session: GC `attempted_acp_spawns` / `acp_respawn_history` / `acp_parked` on session delete — #2729 follow-up


jerome-benoit:
### Summary

The daemon's [`status_poll_loop`](src/server/mod.rs) holds three feature-gated maps across ticks:

- `attempted_acp_spawns: HashSet<String>` (entries added on first spawn attempt, never removed on session delete)
- `acp_respawn_history: HashMap<String, Vec<Instant>>` (entries added on respawn, never removed)
- `acp_parked: HashSet<String>` (entries added on crash-loop park, never removed)

On a long-uptime daemon, each retains every ever-observed instance id. Over days of uptime with session churn the memory footprint grows unboundedly at ~100 bytes per entry per map.

`prev` (the per-tick status snapshot) is loop-body-scoped and does not accumulate; it is not part of this issue.

### Proposed direction

At the top of each tick, sweep each map against `state.instances.read().await`:

```rust
let live_ids: HashSet<&str> = instances.iter().map(|i| i.id.as_str()).collect();
attempted_acp_spawns.retain(|id| live_ids.contains(id.as_str()));
acp_respawn_history.retain(|id, _| live_ids.contains(id.as_str()));
acp_parked.retain(|id| live_ids.contains(id.as_str()));
```

Cost: O(N) per tick per map. At typical N this is negligible relative to the per-tick baseline.

### Acceptance

- Integration test: create a session, mark it parked / spawn-attempted, delete the session, run one tick, assert the maps do not contain the deleted id.
- Memory footprint of a 24 h-uptime daemon (simulated by fast-forwarding the tick loop) is bounded by `live_instance_count`, not by lifetime-observed-instance-count.

### Context

Filed from the review of merged PR #2729.

PR #2865, opened 2026-07-15
fix(tmux): don't latch Error status on a transient tmux-server hiccup

athal7:
## Description
Session::exists() collapsed "tmux server unreachable" (stale socket, refused connection) and "tmux confirms this session is gone" into the same plain `false`, and the status updater latched `Status::Error` (the "✕" glyph) on either. A multi-second tmux-server blip therefore flipped every currently-polled session to dead at once; recovery was throttled ~30s by the existing cold-tier/cooldown gates, so the only apparent remedy was a manual restart — which actually kills and recreates a session that was almost certainly still alive.

Adds a `SessionExistence` tri-state (`Present`/`Absent`/`Unknown`) derived from the existing session cache's own server-reachable signal (no stderr text parsing). The status updater now only latches `Error` on a confirmed `Absent`; `Unknown` retains whatever status the instance already had, mirroring the "can't tell, don't act" convention `batch_pane_metadata`/`attached_session_names` already use.

Scoped to the agent `Session` status path only. `TerminalSession`/`ToolSession` have the same `exists()` pattern but feed lower-severity, self-healing recreation logic rather than an Error latch — left for a follow-up.

Separate from #2839 (unrelated Ctrl+C/signal-handling fix).

## PR Type
- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: this is a subprocess-classification/status-polling edge case, not a UI/keyboard flow. Covered by unit tests (cache-derived tri-state classification, status-retention on `Unknown` in both directions, regression guard that confirmed-absent still latches `Error`) plus a tmux-gated integration test that deterministically kills a real tmux server and asserts `Unknown` (not `Absent`) — pinning the load-bearing assumption against the real binary.

## AI Usage
- [x] AI was used

**Any Additional AI Details you'd like to share:** Investigated via production debug-log analysis (a real tmux-server hiccup event) plus source tracing; design reviewed before implementation; full local `cargo test`/`clippy`/`fmt` run clean pre-push.

## Summary

- Updated the agent’s tmux session status handling so temporary tmux-server connectivity problems don’t immediately mark sessions as permanently “gone”.
- Added a tri-state session existence signal derived from the tmux session cache: **Present**, **Absent**, and **Unknown**.
- Updated the agent session status updater to use this tri-state:
  - **Absent** sessions immediately latch into `Status::Error`.
  - **Unknown** sessions keep the prior status/error and wait through a grace period before escalating.
  - **Present** sessions clear any pending “unknown” tracking and mark the session as confirmed alive.
  - Implemented **bounded escalation** from sustained **Unknown** to `Error` (short window for sessions never confirmed present; longer window for sessions that were previously confirmed present).
- Refined tmux cache probing to be **TTL-based** (stale/failed cache refresh becomes **Unknown**, not **Absent**).
- Ensured the “unknown-to-error” escalation timing continues correctly across **disk reloads** and polling ticks by persisting and merging runtime tracking fields.
- Added/expanded tests, including:
  - Tri-state classification and status retention behavior.
  - Confirmed-absence error latching.
  - Escalation timing (short vs long) and retention of prior status during `Unknown`.
  - Disk reload merge behavior for the new runtime tracking.
  - An integration test that stops the tmux server and verifies probes become **Unknown** (not **Absent**).

## Benefits

- Prevents intermittent tmux outages from incorrectly and permanently failing sessions.
- Maintains stable agent session status during tmux-server restarts/outages, while still guaranteeing eventual escalation if reachability remains uncertain for long enough.
- Makes the behavior consistent across cache refresh failures and across server state reloads.

## Potential enhancement

- Consider applying the same tri-state + bounded-escalation approach to other tmux-dependent session paths if they currently rely on a simpler “exists/missing” interpretation.


njbrake (approved):

Approving. I verified the load-bearing assumption locally against real tmux 3.6 rather than taking the description on trust.

## What I verified

`refresh_session_cache` sets `data = None` in precisely two arms (non-zero exit, spawn failure) and `Some(map)` only on success, so deriving the tri-state from the cache's own signal instead of parsing stderr text is the right call, and it stays robust across tmux versions and locales.

The bug is real rather than theoretical: the old `exists()` received `None` from the cache on an unreachable server, fell through to a `has-session` subprocess that fails the same way during an outage, and `unwrap_or(false)` turned that into a `false` and an Error latch. The fix also removes a subprocess-per-session-per-poll under outage, which is a nice secondary win.

Locally: the three tri-state tests, the three status-retention tests, and the integration test against real tmux all pass; the full tmux module suite is 228 passed / 0 failed; `fmt` and `clippy` are clean on this PR's files. (Two `session::` test failures and two clippy warnings exist, but all four reproduce identically on `origin/main` and live in files this PR does not touch.)

The `SessionCacheGuard` restore-on-drop is a nice touch; it keeps a mid-test panic from leaking forced cache state into a later test.

## One behavior change worth knowing about (non-blocking)

`Absent` requires the server to answer, which requires some session to exist. So killing the *last* session on a socket makes the server exit, and the probe returns `Unknown` where it previously returned Error. Measured on tmux 3.6:

| Scenario | `list-sessions` | Classification |
|---|---|---|
| No server at all | exit 1 | `Unknown` |
| Server + at least 1 session | exit 0 | `Present` / `Absent` |
| Kill the last session, server exits | exit 1 | `Unknown`, retains stale `Running` |
| Server alive, 0 sessions (`exit-empty off`) | exit 0, empty | `Absent`, latches Error |

Worth noting that `warm_tmux_server()` does not rescue this after a reboot: `exit-empty` defaults to `on`, so `tmux start-server` exits 0 but leaves no server running.

This should not block. `remain-on-exit on` keeps normally-exiting agents listed, so the common death path still flows through `Present` plus `is_dead`; `recovery.rs` owns the reboot case and stamps Error on failed recovery; and the state self-corrects as soon as any tmux session exists. The residual window (reboot, nothing recovered, no new session created) is cosmetic staleness, which is strictly better than the bug being fixed, where a routine blip flips every live session to Error and invites a restart that kills a healthy agent.

Row 4 hints at a tidy follow-up: on a socket aoe owns, setting `exit-empty off` would keep the server alive with zero sessions and yield a correct `Absent`, leaving `Unknown` to mean only a genuine connection failure. That is only safe on the dedicated debug socket, not the shared default socket release builds use, so an escalate-after-sustained-`Unknown` timer is the portable alternative.

## Nit

The `SessionExistence` doc comment has ``see #`sessions-kill` ``, which looks like an unfilled placeholder for an issue number.

tests/integration/tmux_reachability.rs
```diff
@@ -0,0 +1,94 @@
+//! Pins the real-tmux assumption behind the false-Error-latch fix
+//! (`probe_session_existence`): `list-sessions` exits non-zero when the
+//! tmux *server* itself is unreachable (dead socket, refused connection),
+//! not only when a specific session is gone. That failure must resolve to
+//! `SessionExistence::Unknown`, never `Absent` -- collapsing the two is what
```
jerome-benoit: AGENTS.md prohibits ` -- ` as a prose separator in comments/docs. Replace with a semicolon:

```
//! `SessionExistence::Unknown`, never `Absent`; collapsing the two is what
```

src/tmux/mod.rs
```diff
@@ -467,21 +467,130 @@ pub fn test_inject_session_into_cache(name: &str) {
...
+
+    Some(match &cache.data {
+        Some(map) if map.contains_key(name) => SessionExistence::Present,
+        Some(_) => SessionExistence::Absent,
+        // The last refresh's `list-sessions` call itself failed (non-zero
+        // exit or spawn error): a definitive "can't tell", not "absent".
+        // Do not fall back to a fresh `has-session` probe here; during a
+        // real outage that call fails the same way and just burns a
+        // subprocess per session per poll for no new information.
+        None => SessionExistence::Unknown,
```
jerome-benoit, line 626: Worth a one-line note here that this is also the whole-server-down case. Because aoe sets `remain-on-exit on`, a dying agent stays `Present` (dead pane, latched via the `is_dead` metadata path), so `Absent` is reachable whenever at least one session survives on the server. A server that is fully down with zero sessions (crash, `kill-server`, or killing the last session) resolves `Unknown` for every polled instance and freezes them at their prior status until it returns, rather than latching Error. That is the intended anti-false-alarm tradeoff; a comment here stops a future reader from collapsing `None` back into `Absent` and reintroducing the bug.

jerome-benoit (commented):
Reviewed with several cross-validation rounds (tri-state derivation and the `refresh_session_cache` None-semantics, scope completeness across every `exists()` caller, a dedicated hunt for a newly-created-session regression, conventions, and a local test run). The fix is correct and I could not find a blocker. njbrake already approved; a few small items below, none blocking.

## Verified

- The tri-state is sound. `refresh_session_cache` writes `Some(map)` on `list-sessions` success (even when the map is empty) and `None` only on non-zero exit or spawn failure, so `Absent` is reachable ("server answered, session not listed") and `Unknown` is exactly "server unreachable". First-use is safe too: an uninitialized cache (`time == None`) is treated as stale, so `probe_session_existence` forces a real `list-sessions` before deriving.
- No newly-created-session regression. I checked whether routing through the whole-list cache (instead of the old per-session `has-session` fallback) could transiently derive `Absent` on a just-created session and false-latch Error: it cannot. Both pollers call `refresh_session_cache()` at cycle start before probing (`src/tui/status_poller.rs:145` before `:219`; `src/server/mod.rs:3436` before `:3446`); `create_with_size` refreshes the cache synchronously before the status leaves `Creating`; `polling_tier(Creating) == 0` plus the `Starting`/`last_start_time < 3s` guard and the server-side `recently_restarted` suppression cover the launch window; and `Absent` only sets `Status::Error`, it never kills/recreates. The old `has-session` fallback was in fact never reached from the poll path (the cache is refreshed every cycle, so it is never stale there), so old and new behave identically on a fresh cache; they diverge only on `data == None` (server unreachable), which is exactly the bug this fixes.
- Scope is complete for the status-poller path: no other `Status::Error` latch fires on a plain `exists() == false` in the agent path. `TerminalSession`/`ToolSession` `exists() == false` branches are self-healing recreation or no-op/bail, never an Error latch, so deferring them is safe.
- Recovery works: after `Unknown -> Present`, the existing metadata logic re-evaluates and clears the error (gated by the pre-existing 30s throttle). Freezing `last_error_check` during `Unknown` is benign (the throttle and cold-tier gating key off `status`, not that timestamp). `#2846` `socket_name` is consistent: create, list, and probe all go through the one `tmux_command()` socket, so there is no split between which server is listed and where the session lives.
- Ran locally on the PR head: `cargo fmt --check` is clean, the six new unit tests pass, and the real-tmux `tmux_reachability` integration test passes.

## Convention fix (please address)

- `tests/integration/tmux_reachability.rs:5`: ` -- ` is used as a prose separator in the module doc comment, which AGENTS.md prohibits. Replace with `;`. (inline comment)

## Worth a one-line code comment (documents an intentional limitation)

- The `None => Unknown` arm in `session_existence_from_cache`: because aoe sets `remain-on-exit on`, a dying agent keeps its session `Present` (dead pane, latched via the `is_dead` metadata path), so `Absent` is reachable whenever at least one session survives on the server. The single case that now freezes rather than latches Error is a fully-down server with zero sessions (crash, `kill-server`, or killing the last session): every polled instance resolves `Unknown` and retains its prior status until the server returns. That is the deliberate, correct anti-false-alarm tradeoff, but the diff never states it. A one-line comment there stops a future reader from collapsing `None` back into `Absent` and reintroducing the bug. (inline comment)

## Nits

- `SessionCacheGuard::force_unreachable`/`force_present` swallow a poisoned write-lock with `if let Ok(...)`, while `capture()` uses `.expect(...)`. In a test helper the `.expect` pattern is better, so a poisoned lock fails loudly instead of silently leaving stale state.
- `tests/integration/tmux_reachability.rs:54`: the session name `aoe_reachability_probe` is fixed; a crashed prior run that left the integration server alive would make `new-session` fail with "duplicate session". A best-effort `kill-session` before `new-session` (or a random suffix) removes that flake risk.
- Test coverage: the suite pins `Absent`/`Unknown`/`Present` in isolation and the real `kill-server -> Unknown` case, but nothing pins the "a just-created session is `Present` on the first poll" invariant end-to-end. A small test (create via the harness, then assert the first `probe_session_existence` is `Present`) would guard the create-then-refresh ordering this whole fix rests on.

## Follow-up (separate, pre-existing, not this PR)

- `src/session/instance.rs:3905` `probe_settle` still uses the plain `exists()` and returns `ProbeResult::Dead` on `false`. It runs only inside an explicit restart/resume cascade (the passive poller is stopped there), so it is not a regression, but a resume cascade during a transient outage could recreate on a false `Dead`. Worth a follow-up issue applying `existence()` / treating `Unknown` as "keep probing", plus a one-line comment noting the intentional binary choice for now.

Thanks for the careful investigation and for pinning the assumption against the real tmux binary.

njbrake (changes requested):

Superseding my earlier approval, which I posted before CI had reported. That was premature of me, and the apology is mine rather than any fault of the patch. CI has since gone red, and the second failure below is a design question worth resolving before this lands.

To be clear up front: the diagnosis here is good. The bug is real, deriving the tri-state from the cache's own signal is the right instinct because it avoids brittle stderr parsing, the scoping to the agent path is disciplined, and the `SessionCacheGuard` restore-on-drop is a nice touch. My concern is where the `Unknown` boundary sits, not the approach.

### 1. `Docs` job fails: private intra-doc link

```
error: public documentation for `probe_session_existence` links to private item `SESSION_CACHE`
 --> src/tmux/mod.rs:587
```

`probe_session_existence` is `pub`, `SESSION_CACHE` is a private static, and #2842 adopted `deny(rustdoc::private_intra_doc_links)`. Dropping the brackets so it reads as a plain code span clears it.

Worth knowing that `cargo clippy` and `cargo test` both pass without catching this. I reproduced it with the invocation CI uses, against a local merge of this branch with main:

```
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve
```

The `--features serve` part matters. With default features you get unrelated noise from a feature-gated link elsewhere, which sent me down a false trail at first.

### 2. `Playwright (live)`: the new test and an existing contract disagree

```
ensure-session-restart.spec.ts: "dead session is restarted by /ensure"
  Expected: "Error"
  Received: "Idle"
```

This one is structural rather than a typo. Two assertions in the tree now contradict each other:

- **This PR's `tests/integration/tmux_reachability.rs`** asserts that after `kill-server`, `probe_session_existence` returns `Unknown`, commented "a dead tmux server must resolve to Unknown, never Absent".
- **The pre-existing `web/tests/live/ensure-session-restart.spec.ts`** (added in #2617, and this PR touches no web files) asserts that a session record with no tmux session behind it flips to `Error`, which is the signal `/ensure` needs to restart it.

Both are deterministic, and both cannot hold. The live test seeds a record via `aoe add` as `Idle` and never creates a tmux session, so there is no server on that socket, so `list-sessions` exits non-zero, so the probe returns `Unknown`, so the row keeps `Idle` and `/ensure` never fires.

### The underlying tension

I reproduced the relevant states against tmux 3.6 on an isolated socket:

| State | stderr | exit |
|---|---|---|
| Socket absent | `error connecting to ... (No such file or directory)` | 1 |
| Server exited after last session was killed | `no server running on ...` | 1 |
| Server alive, at least one session | (session list) | 0 |
| Server alive, zero sessions (`exit-empty off`) | (empty) | 0 |

Rows 1 and 2 both mean no sessions exist, so `Error` is the correct answer for both, and today they land in `Unknown`. The premise of the fix is that a transient blip also makes `list-sessions` exit non-zero and must be retained. As written, `Unknown` catches both populations, so the tri-state cannot separate "the server is genuinely gone, which is real evidence" from "the server blipped, which is not evidence".

Also worth noting that `exit-empty` defaults to `on`, so `warm_tmux_server()`'s `tmux start-server` returns 0 but leaves no server behind. It does not convert `Unknown` into `Absent` after a reboot.

**The question that decides the design:** in the production debug log you analyzed, what did the failing `list-sessions` call actually look like? `refresh_session_cache` logs at `warn` with `status` and `stderr_bytes`, so the exit status should be recoverable. If the blip presented as a spawn failure or a distinct exit status, `Unknown` can be narrowed to that specific case and both contracts hold. If it presented as `no server running on ...`, it is indistinguishable from a genuinely dead server, and no cache-derived signal can tell them apart.

### Possible directions

1. Narrow `Unknown` to spawn failure only, meaning the tmux binary could not be executed, and treat a non-zero exit as `Absent`. Cleanest if the production blip was a spawn failure.
2. Escalate a sustained `Unknown` to `Error` after a bounded window. Preserves the blip fix and still converges, though it is racy against the live test's 10s poll.
3. Set `exit-empty off` on a socket aoe owns so the server persists with zero sessions and returns row 4 above, making `Absent` correct. This only works where aoe owns the socket, not on the shared default socket release builds use.

Happy to dig further if you can share what the debug log showed. Thanks for the thorough writeup and the real-tmux integration test; the latter is what made the contradiction easy to pin down.

jerome-benoit (commented):
Amending my earlier comment above, which I posted before CI had fully reported and which was therefore wrong on the facts: I said "njbrake already approved, CI green, no blocker." That is withdrawn. @njbrake's re-review is right on both counts and I concur with the CHANGES_REQUESTED. Two things I missed and re-confirmed independently:

1. **Docs job (blocker).** `probe_session_existence` is `pub` (`src/tmux/mod.rs:586`) and its doc comment links the private `[SESSION_CACHE]` (and `[batch_pane_metadata]` / `[attached_session_names]`), which trips `deny(rustdoc::private_intra_doc_links)` from #2842. It only surfaces under `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`; I ran fmt plus the unit and integration tests but not the docs build, so I missed it. Dropping the brackets to a plain code span clears it.

2. **The `/ensure` contradiction (design blocker).** I was wrong to file the whole-server-down freeze as an "acceptable documented tradeoff." It breaks the pre-existing deterministic contract in `web/tests/live/ensure-session-restart.spec.ts` (#2617): a seeded Idle record with no tmux behind it must flip to `Error` so `/ensure` restarts it. Under the tri-state, that seed has no server on its socket, so `list-sessions` exits non-zero, resolves to `Unknown`, the row stays `Idle`, and `/ensure` never fires. This PR's `tmux_reachability.rs` ("dead server -> Unknown, never Absent") and that spec ("no tmux -> Error") are both deterministic and cannot both hold.

On the resolution, concurring with your framing that `Unknown` conflates "server genuinely gone" (real evidence, must reach Error for `/ensure`) with "server blipped" (not evidence, must retain): the reported blip (stale socket / refused connection) is itself a non-zero-exit connect error, indistinguishable at the exit-code level from a dead server (rows 1 and 2 of your table), so direction 1 (narrow `Unknown` to spawn-failure-only) likely will not cover the very case the PR set out to fix. Direction 3 only helps on an aoe-owned socket, not the shared default one release builds use. That leaves direction 2 (escalate a sustained `Unknown` to `Absent`/`Error` after a bounded number of polls): transient blip retained, genuinely-gone server converges to Error, satisfying both contracts on the default socket. Its cost is the window-vs-10s-poll race you flagged, which is tunable; alternatively `/ensure` could do its own authoritative `has-session` probe rather than trust the passive latch. The debug-log question you asked (spawn failure vs `no server running` exit) is the right thing to settle first.

Net: I withdraw the "no blocker." The `Unknown` boundary needs reworking (plus the one-line doc-link fix) before this lands. Apologies for the premature comment.

njbrake:

Reviewed the current version (`2b1659b`), @athal7. This is good work, and the design question from the earlier `CHANGES_REQUESTED` is effectively resolved.

**On the design:** splitting the `Unknown` grace window by `ever_confirmed_present` is the right call. A session that has never been confirmed alive (the seeded `Idle` row in `ensure-session-restart.spec.ts`, or `aoe add` without a launch) escalates to `Error` in ~4s, comfortably inside that test's 10s budget, while a confirmed-alive session gets 30s to ride out a real blip. That reconciles the two contracts that previously could not both hold. The `#[serde(skip)]` tracking + `seed_unknown_tracking` before the status decision, with the per-`StatusSource` carry in `reload_state_instances_from_disk`, is careful and well tested.

**Verified locally against the current head:** `fmt` and `clippy --features serve` clean (only the pre-existing `home/mod.rs` "items after a test module" warning, untouched here); the 12 tri-state/escalation/seed lib tests, the 5 disk-reload equivalence tests, and the real-tmux `tmux_reachability` test all pass; and the Docs gate (`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`) now passes, so the earlier private-intra-doc-link failure is fixed.

**One thing left: a rebase onto `main`.** #2884 merged and reworked the same `exists()`/cache path, so this needs a rebase and the two will conflict. I prepared and validated the rebase locally, but could not push it here because "Allow edits by maintainers" is off on this PR. The resolution that came out green, if it helps you move faster:

- `src/tmux/session.rs`: merge the `use super::{...}` import and drop the now-unused `session_exists_from_cache`. `exists()` becomes `crate::tmux::session_exists(...)` (from #2884), `existence()` stays `probe_session_existence(...)`.
- `src/tmux/mod.rs`: keep BOTH #2884's binary `session_exists()` and this PR's tri-state (`SessionExistence` / `session_existence_from_cache` / `probe_session_existence`). Watch the closing brace on `session_exists()`; its `}` sits in the shared trailing context, so a naive union drops it.
- `src/server/mod.rs`: keep both the GC-reconciler block (#2758) and this PR's `prev_unknown_tracking` snapshot in `status_poll_loop`; they are independent.

Optional follow-up (not for this PR): with both `session_exists()` (bool) and `probe_session_existence()` (tri-state) now in the tree, folding the former into the latter would leave a single existence model. Fine to leave for later.

Once it's rebased and CI is green, this is good to go from my side. If you enable "Allow edits by maintainers," I am happy to push the prepared rebase for you instead. Thanks for the thorough writeup and the real-tmux integration test; that test is what made the original contradiction easy to pin down.

athal7:
@njbrake  enabled maintainer edit

athal7:
Root-caused the production incident from `debug.log`: 8+ occurrences across 2026-07-13 to 2026-07-15, each a burst of `list-sessions`/`list-panes` returning non-zero (`unix_wait_status(256)`, consistent `stderr_bytes=51`) lasting a few seconds up to ~11s max, always self-recovering — the same session names show `vt channel armed` both immediately before and after each burst, so the sessions were alive throughout. It hit every currently-polled session at once, not a single spawn failure, and always the same error signature. That rules out "spawn failure" and "exit-empty" — this was a transient connectivity blip to a live server, exactly the case `Unknown` needs to tolerate rather than latch `Error` on.

Went with your option 2 (escalate sustained `Unknown` to `Error` after a bounded window), with a two-tier window: sessions that have never been confirmed `Present` (e.g. `ensure-session-restart.spec.ts`'s fresh-socket case) escalate fast (4s, a couple of poll ticks); sessions previously confirmed `Present` get a 30s window, comfortably past the ~11s max blip observed in prod. Also found and fixed a real bug the unit tests didn't catch: `status_poll_loop` was resetting the tracking fields on every disk reload before the status decision ran, so the escalation was dead in production despite passing unit tests — confirmed via a live Playwright run before and after.

Also addressed CodeRabbit's two follow-up findings (wrong error constant on the `Unknown` escalation path, and `reconcile_from_disk` not carrying the two new tracking fields across relaunch).

Re the rebase — I caught main moving (through #2884, then a few more commits) and brought the branch up via merge rather than rebase, since I'd already pushed and didn't want to force-push over a shared branch. Should be equivalent for review purposes; let me know if you'd still prefer a clean rebase before this lands.

tests/integration/tmux_reachability.rs
athal7: Fixed — replaced `--` with `;` in commit 01cb0e6d.

src/tmux/mod.rs
athal7: Added the explanatory comment in commit 01cb0e6d (now at src/tmux/mod.rs:613-625) — covers the `remain-on-exit`/no-`Absent`-on-full-outage tradeoff you flagged.

njbrake (approved):

Approving, superseding my earlier CHANGES_REQUESTED. Both blockers I raised are resolved and CI is now green on the head commit.

1. Docs job: I re-ran the exact CI invocation (`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`) locally, and the Docs job passed in CI. The private intra-doc link to `SESSION_CACHE` is gone; the remaining links resolve to pub items.

2. Playwright (live): the contradiction is resolved by the bounded-window escalation. A never-confirmed-present session escalates a sustained Unknown to Error after 4s, which the live test's 10s poll observes with margin, while `probe_session_existence` still returns Unknown so the integration test holds. `/ensure` decides via a direct tmux probe, independent of the error string, so the new `TMUX_SERVER_UNREACHABLE_ERROR` constant does not affect the restart path. Both live Playwright shards passed.

The CodeRabbit follow-ups are addressed as well: the reachability-specific error constant, the `reconcile_from_disk` field carry, and the regression test it asked for. All 24 CI jobs are green.

Thanks for the thorough writeup and the real-tmux integration test; the latter is what made the contradiction easy to pin down.

PR #2891, opened 2026-07-16
fix(acp): skip lifecycle envelopes when no prompt is in flight

### LINKED ISSUE #2888: Structured view: between-prompt lifecycle channel fills and wedges all notifications until the next prompt

## Problem

A structured-view claude session (claude-agent-acp) goes fully dark between prompts: the sidebar shows it "running", the timeline stops updating, the subagents pane freezes on stale state, and nothing ever resumes it. Meanwhile the agent is actually fine; in the live incident it resumed itself off a background-task notification, pushed a commit, replied to a PR comment, and finished, and none of that ever appeared in aoe. The session sits "running" forever (the between-prompt watchdog can never fire, see Root cause layer B), until the user manually sends a message, at which point the entire missing backlog suddenly flushes into the timeline at once and the session comes back to life.

This is the same user-visible symptom family as upstream `agentclientprotocol/claude-agent-acp#825` ("stalled until the next prompt drains it"), which is why it initially read as a regression of that. It is not: this wedge is entirely on aoe's side. The adapter kept forwarding notifications; aoe stopped processing them.

Live incident: session `8873630590ce48c2`, worker build `1.12.1+gc09805717b0d` (current main), adapter `0.55.0`.

Timeline from `~/.agent-of-empires/acp_events.db` and the worker log:

```
11:50:53  seq 3273  UserPromptSent (user pastes PR comment link)
11:52:17  seq 3303  BackgroundAgentLaunched  agent ad51a1c8... (Task, run_in_background)
11:52:22  seq 3322  Stopped{prompt_complete}   <- main turn ends cleanly, subagent keeps running
11:52:22  ..        subagent child tool calls + progress stream in BETWEEN PROMPTS
12:07:33  (log)     "subagent child tool_call linked ... child=toolu_01E3HFDasiSHTZscqBC9nCKD"
                    <- notification RECEIVED and half-processed; its event is NEVER recorded
12:08:34  seq 3538  BackgroundAgentCompleted{completed}  <- from the transcript tailer, NOT the adapter
          ...       total silence in the event store from here on
12:31:31  seq 3539  UserPromptSent "so?"  (manual poke, 24 min later)
12:31:31  seq 3540  ToolCallStarted toolu_01E3HFDasiSHTZscqBC9nCKD  <- the wedged 12:07:33 notification
12:31:31  seq 3541+ the ENTIRE hidden backlog flushes in one burst
```

The count that gives the mechanism away: between the last `Stopped` (seq 3322) and the wedge, the store recorded exactly **128** adapter-derived lifecycle-bearing events (43 ToolCallStarted + 43 ToolCallCompleted + 9 ToolCallUpdated + 33 RawAgentUpdate). 128 is precisely the lifecycle channel capacity (`src/acp/acp_client.rs:4846`). Event 129 is the one that never landed.

The agent was never stuck. The underlying Claude session transcript (`~/.claude/projects/.../218bc8ba-1176-447c-98bc-6bbb9c7e6146.jsonl`) shows a full agent-initiated turn while aoe recorded nothing:

```
12:08:34  <task-notification> injected (background Task ad51a1c8... completed)
12:08:47  assistant thinking
12:08:51  Bash (git log)
12:09:03  Bash -> "replied: https://github.com/agent-of-empires/agent-of-empires/pull/2887#issuecom..."
12:09:14  assistant final text
```

Live confirmation of the recovery path: sending `so?` at 12:31:31 flushed seq 3540+ instantly, starting with the exact tool_call id the worker log had linked at 12:07:33, followed by the hidden turn's messages ("Pickers done (`ff1f3dc0`)... Pushing + replying...", two Terminal calls, "Shipped."), then the new turn ran normally. Prediction made before the poke, observed exactly.

### Layer A: the between-prompt lifecycle channel has no consumer, and the notification handler does awaited sends into it

The notification handler classifies every `session/update` into `LifecycleSignal`s and forwards them to the per-prompt watchdog over a bounded mpsc of capacity 128 (`src/acp/acp_client.rs:4846`). That channel is drained only while a `session/prompt` is in flight; at the start of each prompt the loop bumps the epoch and flushes leftovers (`acp_client.rs:6115-6118`), and envelopes with a stale epoch are discarded.

Between prompts nobody drains it. A background Task subagent that keeps working after `Stopped{prompt_complete}` streams child tool calls as notifications; each one pushes an envelope. After 128 of them the channel is full, and the 129th send blocks forever: `send_lifecycle_signal` uses an awaited `send` for tool lifecycle signals, and even `Progress` falls back to an awaited send when the channel is full (`acp_client.rs:1260-1290`, rationale from #1401).

Handler order for one notification (`acp_client.rs:5040-5199`):

1. between-prompt bookkeeping, including inserting the started tool into `between_prompt_tools` (5076)
2. `map_update_to_events`, which logs the "subagent child tool_call linked" breadcrumb (3735)
3. awaited `send_lifecycle_signal` (5131) <- blocks here
4. `event_tx.send` to the event store (5198) <- never reached

That matches the log byte for byte: the 12:07:33 notification logged its breadcrumb and vanished. Notification dispatch is serialized per connection, so one blocked handler freezes every subsequent notification, including the whole 12:08:47 agent-initiated turn. Only the transcript-tailer (`src/acp/background_agent.rs`), a separate task, kept recording, which is why `BackgroundAgentProgress`/`Completed` continued to 12:08:34.

The #1401 epoch mechanism explicitly anticipated handlers "parked on a full channel" (`acp_client.rs:4836-4845`) and made the parking safe for watchdog state; it did not make it finite. Between prompts, with no drainer, parked means forever.

### Layer B: the wedge permanently disarms the between-prompt watchdog, so "running" never self-heals

Step 1 above ran before the block, so the wedged tool sits in `between_prompt_tools` with no completion ever processed. `between_prompt_should_fire` returns false unconditionally while `work_in_flight` is true (`acp_client.rs:1205-1207`). The between-prompt idle watchdog (#2325, #2371) therefore can never synthesize `Stopped(agent_idle)`; not even the 30-minute floor applies. The session shows "running" indefinitely, and everything downstream that keys on a terminal Stopped (idle detection, badges) stays frozen.

### Layer C (correct-by-accident recovery): the next prompt flushes the channel

On the next `session/prompt` the loop bumps the epoch and drains the channel (`acp_client.rs:6118`). The blocked send completes, the handler resumes, and the queued notifications drain into the event store in order (with recorded-at timestamps of flush time, not occurrence time). This is exactly the observed "sending a message wakes it up". The stale envelopes are epoch-discarded, so the flush is state-safe; it is just 24 minutes late.

### A. Don't send lifecycle envelopes when no prompt is in flight (the fix)

Between prompts, every envelope is guaranteed garbage: the next prompt flushes the channel and epoch-discards stragglers before arming its watchdog, so nothing ever consumes them. Blocking the notification pipeline to enqueue them is pure downside. Gate the two `send_lifecycle_signal` calls (`acp_client.rs:5130-5151`) on the same `prompt_in_flight` load the between-prompt bookkeeping already does at 5047 (reuse one load for both).

Race analysis: if a prompt starts between the load and the skip, the skipped envelope carries the pre-bump epoch and would have been flushed or epoch-discarded anyway, so behavior is identical. In the reverse direction (handler sees `prompt_in_flight = true` just as the prompt completes), a handful of envelopes can land in the undrained channel, bounded by the notifications mid-processing at the boundary, and the next prompt's flush clears them; the 128 ceiling is unreachable.

### B. Regression test for the wedge shape

Unit test on the notification path: with `prompt_in_flight = false`, feed 200+ tool-lifecycle notifications and assert every one reaches the event store (no block, no missing events) and that `between_prompt_tools` drains to empty when the completions arrive. This is the test #1401 and #2371 never had for the between-prompt lane.

### C. Defense-in-depth for the watchdog (optional follow-up)

Layer B's freeze had a second precondition: a `ToolStarted` in `between_prompt_tools` whose completion never arrives suppresses the watchdog forever. With A in place the known trigger is gone, but any future stall in the notification path recreates it. Consider a generous TTL (for example the 30-minute floor) after which a stuck-open between-prompt tool stops suppressing `between_prompt_should_fire`. Could be folded into this fix or split out.

## Why now

- P0: any structured-view session whose background subagent emits more than 128 tool events between prompts wedges deterministically. Long-running Task subagents (the `/loop`, monitor, and background-agent workflows we keep shipping) hit this shape routinely.
- Completed work becomes invisible: in the incident the agent pushed a commit and posted a PR reply while the UI showed a dead session. Users make decisions (re-prompting, killing workers, re-doing work) based on a false "stalled" picture.
- The failure is silent and permanent until a human pokes the session; no watchdog, timeout, or log line surfaces it.

- As a structured-view user, when the main turn ends while a background Task subagent keeps emitting tool calls (more than 128 events) between prompts, then every notification is still processed and recorded live and the notification handler never blocks.
  - Suggested test layer: Rust unit test on the notification path (proposal B).
- As a structured-view user, when an agent-initiated between-prompt turn finishes and no work remains in flight, then the session reaches a terminal `Stopped` within the expected grace and the sidebar stops showing "running".
  - Suggested test layer: Rust unit test on `between_prompt_should_fire` state after a burst-plus-completion sequence.
- As a web dashboard user, when the agent resumes itself off a background-task notification, then the resumed turn's output appears in the timeline as it happens, not only after my next message.
  - Suggested test layer: `tests/e2e/` live-daemon test with the fake ACP agent emitting a between-prompt notification burst.

## Open questions

- The three wakes scheduled for 12:10:55 never fired agent-side (the transcript shows nothing after the resume turn ended at 12:09:15). Cleared by the task-notification resume, or dropped on session/load re-arm? Worth a targeted repro; if real, it is an upstream (claude-agent-acp / SDK) ask, since #2377 established aoe does not fire wakes.
- Flushed-late events are recorded with flush-time timestamps (all of seq 3540-3578 say 12:31:31). Fine for ordering, misleading for forensics. Acceptable, or worth carrying an occurred-at hint when the source notification provides one?

## Out of scope

- Upstream #825 (adapter prompt-loop hang mid-turn): different component, already fixed in adapter 0.55.0, exonerated here by the half-processed notification and the completed transcript turn.
- Wake delivery architecture: #2377 deliberately ruled out aoe-side wake firing; this issue does not reopen that.
- Subagents-pane cosmetics beyond what unblocking the pipeline restores.

## Cross-references

- #1401 — introduced the lifecycle channel, the awaited-send policy, and the epoch guard; the between-prompt no-drainer case is the gap.
- #2325 — between-prompt (agent-initiated) turns and their synthesized `Stopped`; layer B freezes that mechanism.
- #2371 / PR #2377 — between-prompt watchdog grace fixes; #2377's scope note ("aoe does not fire wakes") is why agent-side resume is the only wake path, making its invisibility a total outage.
- #2573 — `between_prompt_bg_agents` tracking; the tailer kept working through this incident, which is what made the wedge point diagnosable.
- #2455 — prior "stalled until poked" bug at the pipe level; same symptom family, different mechanism.
- agentclientprotocol/claude-agent-acp#825 — the upstream member of the family; NOT the cause here.

Seluj78:

## Description

Fixes the between-prompt notification wedge from #2888. The per-prompt lifecycle channel (capacity 128) is drained only while a `session/prompt` is in flight and flushed at the next prompt start, so envelopes sent between prompts are never consumed. A background Task subagent that keeps streaming child tool calls after the main turn ends could fill all 128 slots; the next awaited send then parked the notification handler forever. Every later notification, including a whole agent-initiated turn, queued invisibly until the next user prompt, and the stuck `ToolStarted` kept the between-prompt watchdog permanently suppressed, so the session sat "running" with a frozen timeline. Live incident: session `8873630590ce48c2`, where the agent resumed itself, pushed a commit, and replied to a PR comment, none of it visible until a manual poke.

The fix extracts the two send sites into `forward_lifecycle_signals`, gated on `prompt_in_flight`. Between prompts the send is skipped entirely; nothing is lost because a between-prompt envelope could only ever be flushed unread, and the between-prompt watchdog is fed by atomics in the notification handler, not by this channel.

The reproducing test fails on the ungated code with a timeout (the exact production wedge: full channel, no drainer, awaited send never returns) and passes with the gate. A second test pins the active-prompt path so the per-prompt watchdog keeps receiving its signals.

Closes #2888

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view claude session and have it launch a long-running background Task subagent (`run_in_background`), then let the main turn end while the subagent keeps working. Watch the subagent's child tool calls keep appearing in the timeline well past 128 events; before the fix the timeline froze partway through.
- [ ] When the subagent completes and the agent resumes itself off the task notification, confirm the resumed turn's output streams in live, without sending a message first.
- [ ] After the agent finishes, confirm the session drops back to idle within the normal grace instead of showing "running" indefinitely.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: reproducing the wedge end-to-end needs a between-prompt notification burst the fake ACP agent cannot script today (it only emits during a turn). The wedge is regression-tested at the Rust unit layer instead, proven red on the ungated code and green on the fix; a post-turn script verb for the fake agent is noted as follow-up in #2888.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Claude Code, `/aoe-implement` skill)

Investigation, root-cause analysis, fix, and tests drafted by Claude under my direction, including the live confirmation on the wedged session. I made the design calls and reviewed the diff.

## Summary

- Prevents notification-handler freezes by forwarding lifecycle updates only while a prompt is active.
- Adds a helper to centralize lifecycle and wakeup signal forwarding.
- Adds regression tests covering full-channel bursts between prompts and signal delivery during active prompts.

## Benefits

- Prevents structured-view timelines from becoming stuck.
- Improves idle-state detection after background work completes.
- Preserves lifecycle monitoring during active prompts.

A future enhancement could add a timeout for tools that remain incomplete indefinitely between prompts.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -1289,6 +1289,36 @@ async fn send_lifecycle_signal(
     }
 }
 
+/// Forward a notification's watchdog signals to the per-prompt lifecycle
+/// channel, but only while a prompt is in flight.
+///
+/// The channel's sole consumer is the prompt loop: it drains during a
+/// `session/prompt` and flushes leftovers (discarding them by epoch) at
+/// the start of the next one. Between prompts nothing reads it, so a busy
+/// agent (a background Task subagent streaming child tool calls, or an
+/// agent-initiated turn) fills all slots and the next awaited send parks
+/// the notification handler until the next prompt; with dispatch
+/// serialized per connection, that freezes every subsequent notification
+/// and the session appears dead while the agent keeps working. Skipping
+/// the send loses nothing: a between-prompt envelope could only ever be
+/// flushed unread. The between-prompt watchdog is fed by atomics in the
+/// notification handler, not this channel. See #2888.
+async fn forward_lifecycle_signals(
+    prompt_active: bool,
+    tx: &mpsc::Sender<LifecycleEnvelope>,
+    epoch: u64,
+    lifecycle: Option<LifecycleSignal>,
+    wakeup: Option<LifecycleSignal>,
+    session_label: &str,
+) {
+    if !prompt_active {
+        return;
+    }
+    for signal in [lifecycle, wakeup].into_iter().flatten() {
+        send_lifecycle_signal(tx, LifecycleEnvelope { epoch, signal }, session_label).await;
+    }
+}
+
 /// Detect whether a `ToolCallUpdate` completion content array carries a
 /// Claude SDK marker that the underlying work continues off-protocol after
 /// the visible tool call completes. Two markers today:
@@ -5037,6 +5067,7 @@ async fn run_connection_task<W, R>(
                     if lifecycle_signal.is_some() || wakeup_signal.is_some() {
                         first_event_after_attach.store(true, Ordering::Relaxed);
                     }
+                    let prompt_active = prompt_in_flight.load(Ordering::Relaxed);
                     // Between-prompt idle tracking (#2325). Only while no
                     // aoe-issued prompt is in flight: a lifecycle signal here
                     // means the agent resumed itself (Monitor / scheduled
@@ -5044,7 +5075,7 @@ async fn run_connection_task<W, R>(
                     // its cost/progress/wake semantics so the outer loop's
                     // idle tick applies the same grace. During a real prompt
                     // the per-prompt watchdog owns this, so skip.
-                    if !prompt_in_flight.load(Ordering::Relaxed) {
+                    if !prompt_active {
                         let now = chrono::Utc::now().timestamp_millis();
                         if let Some(u) = between_prompt_signal_update(
                             lifecycle_signal.as_ref(),
@@ -5126,29 +5157,20 @@ async fn run_connection_task<W, R>(
                     // cancel a legitimate wait. Watchdog correctness
                     // wins; UI ordering is reconciled by the event
                     // store's monotonic seq anyway. See #1401 post-
-                    // impl review.
-                    if let Some(sig) = lifecycle_signal {
-                        send_lifecycle_signal(
-                            &lifecycle_signal_tx,
-                            LifecycleEnvelope {
-                                epoch: envelope_epoch,
-                                signal: sig,
-                            },
-                            &session_label,
-                        )
-                        .await;
-                    }
-                    if let Some(sig) = wakeup_signal {
-                        send_lifecycle_signal(
-                            &lifecycle_signal_tx,
-                            LifecycleEnvelope {
-                                epoch: envelope_epoch,
-                                signal: sig,
-                            },
-                            &session_label,
-                        )
-                        .await;
-                    }
+                    // impl review. Skipped entirely between prompts:
+                    // nothing drains the channel then, so at capacity
+                    // the awaited send would wedge this handler, and
+                    // every notification behind it, until the next
+                    // prompt (#2888).
+                    forward_lifecycle_signals(
+                        prompt_active,
+                        &lifecycle_signal_tx,
+                        envelope_epoch,
+                        lifecycle_signal,
+                        wakeup_signal,
+                        &session_label,
+                    )
+                    .await;
                     for event in mapped_events {
                         // An async sub-agent launch: spawn a tailer that
                         // follows the agent's on-disk transcript and emits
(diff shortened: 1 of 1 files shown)


njbrake (approved):

Approve. Clean, minimal fix for the #2888 wedge, with an exemplary root-cause writeup.

I verified the load-bearing claim locally rather than trusting the description: neutering the `if !prompt_active { return; }` guard makes the between-prompt regression test time out at 5s (the exact production wedge: full channel, no drainer, awaited send never returns); restored, it passes. Red-on-ungated, green-on-fix is genuine.

Also confirmed:
- Flush semantics at acp_client.rs:6115-6118 (epoch bump plus full `try_recv` drain at each prompt start), so between-prompt envelopes are provably discarded and skipping the send loses nothing.
- The between-prompt tool bookkeeping runs before the gated send and is untouched, so Layer B watchdog recovery is preserved.
- `send_lifecycle_signal` now has exactly one caller; no other ungated send path remains.
- fmt, clippy (`--features serve`), and both new tests clean.

One non-blocking follow-up: the issue mandate also calls for a direct regression test on `between_prompt_should_fire` recovering to a terminal Stopped within grace (user story 2). The fix handles that transitively, and the deferred e2e (user story 3) is documented with a concrete reason, so this is a follow-up rather than a merge blocker.

PR #2944, opened 2026-07-17
fix(status): stop typed-but-unsubmitted input from flapping a working Claude session to Idle

njbrake:
## Description

Fixes a status-detection gap where a working Claude session flaps to Idle the moment the user types text into the input box without submitting it (reported on a sandboxed session after an `e` restart).

Reproduced live against Claude Code 2.1.212: typing text mid-turn repurposes Esc to "clear input", so the footer drops its `esc to interrupt` hint, and no spinner line renders while response prose streams. The pane then carries zero running signals while the agent is actively working:

```
  <streaming prose...>
──────────────────────────────
❯ this is some unsubmitted text i am typing while the agent works
──────────────────────────────
  ⏵⏵ bypass permissions on (shift+tab to cycle)
```

This is byte-identical to the parked pane except for the past-tense completion line (`✻ Cooked for 49s`) above the input box. Both detection layers misread it:

- the hookless pane fallback (`detect_claude_status`) reported Idle outright (the mode-cycle footer reads as an idle marker), and
- the age-gated ready-prompt reconciler (`claude_pane_shows_ready_prompt`) downgraded a stale-but-true Running hook write to Idle.

The fix treats unsubmitted typed text without parked evidence (a completion line or the Esc-interrupt banner directly above the input box) as ambiguous:

- `claude_pane_shows_ready_prompt` gains a veto, so the reconciler keeps the hook's Running;
- the hookless fallback in `update_status_with_metadata_inner` holds an already-observed Running instead of flapping to Idle. The completion line rendered at turn end releases the hold on the next poll.

Ghost suggestion text (#2919) keeps its parked verdict: it only renders after a finished turn, i.e. with the completion line above the box (all existing ghost-text fixtures have it), so those downgrades still fire. The failure direction of the new ambiguity is stuck-Running, which is the documented preferred direction for this detector (see the `IDLE_RECONCILE_MIN_RUNNING_AGE` cost-asymmetry note).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the ambiguity is a pure pane-text-shape property, covered by unit tests built from panes captured from a real Claude Code 2.1.212 session (streaming-with-typed-text, parked-with-typed-text, completion-line shapes, and the ambiguity helper). A regression check confirmed the new reconciler test fails without the fix. Driving a real mid-turn typed-text state through the e2e harness would need a live generating agent, which the fake-agent shims can't express deterministically.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Code (Fable 5)

Root cause was established by reproducing the pane states against a live Claude Code 2.1.212 session in tmux and diffing the running/parked footer variants; fix and tests reviewed by njbrake.

## Summary

Fixes Claude Code session status “flapping” (switching from **Running** to **Idle**) when a user types text that hasn’t been submitted yet during an active turn.

## Changes

- Added detection for an **ambiguous typed prompt** state: typed-but-unsubmitted input no longer automatically downgrades the pane to idle/ready.
- Updated the status reconciliation logic to **hold onto a Running state** in ambiguous situations:
  - When the hook reports Running, it preserves that state.
  - When no hook evidence exists, it preserves an already-observed Running state.
- Enhanced completion-line detection to avoid false positives by requiring a structured **duration token** (e.g., `49s`, `1m`, `2h`) before treating a line as “turn completed / parked evidence”.
- Kept existing **ghost suggestion text** handling behavior unchanged.
- Added unit tests covering Claude Code 2.1.212 pane shapes, including:
  - streaming with typed (unsubmitted) text
  - parked typed text
  - completion lines releasing the hold
  - ambiguity detection behavior

## Benefits

- More stable and accurate session status reporting for Claude Code.
- Fewer premature Idle transitions while Claude is still actively working.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -4863,6 +4863,26 @@ impl Instance {
         } else {
             false
         };
+        // A Claude pane with unsubmitted typed text in the input box can show
+        // no running signal at all while a turn streams (typing suppresses the
+        // `esc to interrupt` hint and prose streaming renders no spinner), and
+        // that pane is identical to a parked one minus the completion line. In
+        // the ambiguous state, hold an already-observed Running rather than
+        // flap a working session to Idle; the completion line rendered at turn
+        // end releases the hold on the next poll.
+        let detected = if detected == Status::Idle
+            && !shell_stale
+            && !is_dead
+            && self.status == Status::Running
+            && detection_tool == "claude"
+            && tmux::claude_pane_is_ambiguous_typed_prompt(&pane_content)
+        {
+            tracing::debug!(target: "session.store",
+                "status '{}': holding Running over ambiguous typed-prompt Idle", self.title);
+            Status::Running
+        } else {
+            detected
+        };
         self.status = resolve_detected_status(
             detected,
             is_dead,
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -16,7 +16,8 @@ pub use session::{PaneCursor, Session, SIZE_OWNER_HEARTBEAT, SIZE_OWNER_TTL};
 pub use status_bar::{get_session_info_for_current, get_status_for_current_session};
 pub use status_detection::detect_status_from_content;
 pub(crate) use status_detection::{
-    reconcile_claude_hook_status, reconcile_codex_hook_status, reconcile_waiting_hook,
+    claude_pane_is_ambiguous_typed_prompt, reconcile_claude_hook_status,
+    reconcile_codex_hook_status, reconcile_waiting_hook,
 };
 pub use terminal_session::{kill_all_terminals_for_id, ContainerTerminalSession, TerminalSession};
 pub use tool_session::{kill_all_tool_sessions_for_id, ToolSession};
--- a/src/tmux/status_detection.rs
+++ b/src/tmux/status_detection.rs
@@ -314,10 +314,16 @@ fn claude_line_is_background_wait(line: &str) -> bool {
 /// Match the past-tense turn-completion line Claude renders directly above
 /// the input box when a turn ends: `✻ Cooked for 49s`, `✻ Baked for 10s ·
 /// 1 shell still running`, `✻ Worked for 1m 52s`. Shape: a spinner frame
-/// char, a capitalized verb without the active `…`, then `for <duration>`.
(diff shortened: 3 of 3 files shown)


jerome-benoit:
@njbrake: It more looks like a workaround than a clean fix: why the state machines for trash/docker/wortree are not used to cleanly track the exact state and do just the exact operations needed to cleanly restore ?

njbrake:
> @njbrake: It more looks like a workaround than a clean fix: why the state machines for trash/docker/wortree are not used to cleanly track the exact state and do just the exact operations needed to cleanly restore ?   

Thanks for the comment! I'll take a look. Yeah this is another one that hasn't been quite working for me.

jerome-benoit:
> > @njbrake: It more looks like a workaround than a clean fix: why the state machines for trash/docker/wortree are not used to cleanly track the exact state and do just the exact operations needed to cleanly restore ?
> 
> Thanks for the comment! I'll take a look. Yeah this is another one that hasn't been quite working for me.

That comment landed on the wrong PR ...

njbrake:

The trash/docker/worktree state machines track operations aoe initiates, so every transition is observable by definition. Session status is a different problem: the ground truth is whether a third-party process is generating, and aoe never sees those transitions directly.

The event-driven layer you're describing does exist and is primary: Claude Code's lifecycle hooks write state transitions to a status file, and detection trusts that stream first. The pane heuristics only cover the stream's documented gaps, which are upstream behaviors rather than aoe design choices: Esc-cancel fires no completing hook, a tool result with no following text fires neither Stop nor idle_prompt, a restart clears the file, and custom wrappers or user-managed containers have no hooks at all.

This PR fixes one of those gap cases. With unsubmitted text in the input box, a running pane and a parked pane are byte-identical except for the completion line (verified against a live 2.1.212 session), so there is no cleaner state to track from outside the process. Agreed the truly clean fix is upstream: if Claude Code emitted a hook on interrupt and input-queue, several of these heuristics could be deleted. Happy to file that upstream request.

That said, if you have a concrete shape in mind for how a state machine could cover these cases, I'd genuinely like to hear it. If there's a way to shrink the heuristic surface here, I'm interested.

PR #2945, opened 2026-07-17
fix(trash): stop a raced restore from stranding a sandboxed worktree in the holding area

njbrake:
## Description

Trashing a containerized session with `d`, then restoring it before the off-thread teardown finishes, left the session unresumable: the worktree quietly moved into `.aoe-trash/<id>` while the restored row kept pointing at the now-empty original path.

The teardown (#2930) runs on the `TrashPoller` against a clone taken at trash time: tmux kill, then `docker stop` (which blocks about 10 seconds because the container's PID-1 `sleep infinity` ignores SIGTERM), then `git worktree move`. A restore landing inside that window is a NoChange on disk (no relocation marker exists yet) and untrashes the durable row, but the worker still moved the worktree and discarded the container. `apply_trash_results` then dropped the relocation result because the row was no longer trashed, and the reconcile pass only heals trashed rows, so nothing ever moved the worktree back. Non-sandbox sessions tear down in well under a second, which is why only containerized sessions hit this in practice.

The fix, in layers:

- `prepare_trashed_worktree` re-checks the durable row between the container stop and the worktree move, and skips the move when the row is no longer trashed. Fail closed: an unreadable storage, a purged row, or a restored row all leave the worktree in place, and a skipped move on a still-trashed row is healed by the next reconcile pass. One change point that covers the TUI, server, and CLI surfaces.
- The relocation is committed atomically against the durable row: `commit_trash_relocation` (in `session/claim.rs`, per CodeRabbit's review) re-takes the decision under the storage flock at every persist site (TUI poller drain, server trash handler, CLI remove), persisting only when the row still reads trashed and unclaimed. A restore or fresh purge/restore claim that won in the meantime supersedes the relocation, and the disk move is undone via the new `trash::undo_raced_relocation` instead of being recorded onto a live row.
- `aoe rm` stamps `source_profile` before the teardown so a profile-scoped remove re-checks the storage it actually trashed the row in.

Follow-up: making the teardown a first-class claimed operation (a `ClaimOp::Trash` alongside Purge and Restore in the #2541 protocol) so peers can observe "teardown in flight" as durable state rather than inferring it. That lands separately.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the race needs a live docker container mid `docker stop` grace period, which an e2e cannot reproduce deterministically. Covered instead by unit-level regression tests in `src/session/trash.rs`: `teardown_skips_relocation_when_row_was_restored_mid_flight` (fails without the fix, with the worktree relocated under `.aoe-trash/`) and `undo_raced_relocation_moves_worktree_back`, both against real git worktrees.

## Testing

`cargo test`: 4114 passed. One failure (`restart_selected_session_surfaces_resume_failed_after_async_restart`) fails identically on a clean `main` checkout in this environment, so it is pre-existing and unrelated. `cargo fmt --check` clean, `cargo clippy --all-targets` adds no new warnings, `cargo check --features serve` compiles.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Fable 5 via Claude Code

Root-caused via a failing reproduction test before any fix was written; the repro is kept as the regression test.

## Summary

- Fixed a race where a session restore could happen while an async trash teardown was stopping a sandbox/container and moving the worktree, leaving the worktree stranded under `.aoe-trash/<id>` while the restored session record no longer referenced it.
- Added a “last-moment” check right before relocating the worktree so relocation is skipped if the durable record shows the session has already been restored/untrashed.
- Changed trash relocation to be recorded durably first, and only then applied to in-memory state—preventing in-memory/worktree mismatches when races occur.
- Added an “undo” path: if relocation was performed but a restore wins first, the code moves the worktree back to its pre-trash location and cleans up the temporary trash copy.
- Ensured `aoe rm` captures and reuses the correct `source_profile` in trash-first removals, so subsequent `-p <profile>` operations re-check the same storage row.
- Updated TUI and server flows to use the durable relocation decision and trigger the undo logic when relocation is superseded.
- Added regression tests covering both cases: (1) relocation skipped after a mid-flight restore, and (2) raced relocations being properly reversed.

## Benefits

This makes concurrent restore + cleanup much more reliable by preventing stranded trash worktrees and ensuring the session’s recorded location stays consistent with what exists on disk, reducing flakiness under load.

- The reviewer discussion suggests a follow-up could make teardown itself a first-class state/claim operation (rather than coordinating via the gap around container stop), though that would require a broader protocol change.

### DIFF
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -132,35 +132,43 @@ pub async fn run(profile: &str, args: RemoveArgs) -> Result<()> {
                     new_project_path: inst.project_path.clone(),
                     pre_trash_project_path: inst.pre_trash_project_path.clone(),
                 };
-                let commit = storage.update(|all_instances, _groups| {
-                    Ok(crate::session::claim::commit_trash_relocation(
+                // The decision travels through this captured slot rather than
+                // the closure's return value, so it survives an update that
+                // decided Superseded and then failed its final write; the
+                // undo keys off the decision alone, since the durable row was
+                // already restored in that case. A Persisted decision whose
+                // write failed needs no repair: the row is still trashed at
+                // its old path and the next reconcile repoints it.
+                let mut decided: Option<crate::session::claim::RelocationCommit> = None;
+                let update_result = storage.update(|all_instances, _groups| {
+                    decided = Some(crate::session::claim::commit_trash_relocation(
                         all_instances,
                         &removed_id,
                         &reloc,
                         chrono::Utc::now(),
-                    ))
+                    ));
+                    Ok(())
                 });
-                match commit {
-                    Ok(crate::session::claim::RelocationCommit::Persisted) => {}
-                    Ok(crate::session::claim::RelocationCommit::Superseded) => {
-                        match crate::session::trash::undo_raced_relocation(&inst, &reloc) {
-                            crate::session::trash::RestoreOutcome::Failed { reason } => {
-                                eprintln!(
-                                    "  Note: a concurrent restore superseded the trash; could not move the worktree back ({reason})."
-                                );
-                            }
-                            _ => {
-                                eprintln!(
-                                    "  Note: a concurrent restore superseded the trash; the worktree was left in place."
-                                );
-                            }
+                if let Err(e) = &update_result {
+                    eprintln!(
+                        "  Note: could not persist the trash relocation ({e}); it will be reconciled on next load."
+                    );
+                }
+                if matches!(
+                    decided,
+                    Some(crate::session::claim::RelocationCommit::Superseded)
+                ) {
+                    match crate::session::trash::undo_raced_relocation(&inst, &reloc) {
+                        crate::session::trash::RestoreOutcome::Failed { reason } => {
+                            eprintln!(
+                                "  Note: a concurrent restore superseded the trash; could not move the worktree back ({reason})."
+                            );
                         }
-                    }
-                    Ok(crate::session::claim::RelocationCommit::AlreadyGone) => {}
-                    Err(e) => {
-                        eprintln!(
-                            "  Note: could not persist the trash relocation ({e}); it will be reconciled on next load."
-                        );
+                        _ => {
+                            eprintln!(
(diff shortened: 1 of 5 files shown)


jerome-benoit:
@njbrake: It more looks like a workaround than a clean fix: why the state machines for trash/docker/wortree are not used to cleanly track the exact state and do just the exact operations needed to cleanly restore ?

njbrake:
> @njbrake: It more looks like a workaround than a clean fix: why the state machines for trash/docker/wortree are not used to cleanly track the exact state and do just the exact operations needed to cleanly restore ?

Fair point, and the latest commit moves in that direction. The commit decision for the teardown's worktree move now runs through the same claim layer that serializes purge and restore (`commit_trash_relocation` in `session/claim.rs`): it is re-taken on the durable row under the storage flock, and a restore or fresh claim that won in the meantime supersedes the relocation, which is then undone rather than recorded.

The remaining step toward the fully tracked version is making the teardown itself a first-class claimed operation (a `ClaimOp::Trash` alongside Purge and Restore), so peers can observe "teardown in flight" as durable state instead of inferring it. That is a protocol change touching all three surfaces, so it lands as a follow-up rather than growing this PR further. The reason the teardown cannot simply hold the lock through its states is that its slowest transition is a roughly 10 second `docker stop`, and every slow phase here runs on unlocked snapshots by design (#2534, #2541); the claim marker is how that design tracks state across the gap.

njbrake:

@jerome-benoit the follow-up promised above is up as #2960: the teardown now participates in the claim protocol as durable state (`ClaimOp::Trash`), with purge/restore seizing it and the teardown yielding at both of its decision points. Restore was already driven by tracked state (`pre_trash_project_path` for the move-back, container existence for recreation); this makes the in-flight phase observable too. If you had a different shape in mind, 2960 is the right place to say so; genuinely interested in whether this matches what you were pointing at.

PR #2946, opened 2026-07-17
feat(tui): native-feeling structured view in the home preview pane

### LINKED ISSUE #2925: fix(tui): keyboard-confirmed view switch fires one keystroke late — #2923 follow-up
## Summary

The session context-menu action **Switch to structured** / **Switch to terminal** (added in #2923) opens a confirm dialog. When the user confirms with the keyboard (Enter/`y`), the switch does not take effect until the *next* keypress, and that next key may be interpreted as a fresh command. The mouse `[Yes]` path works correctly.

## Root cause

In `App::handle_key` (`src/tui/app.rs`), the `pending_view_switch` drain runs *before* `self.home.handle_key(...)` produces the action:

- `src/tui/app.rs:2632` drains `pending_view_switch` and calls `perform_view_switch`.
- `src/tui/app.rs:2636` calls `self.home.handle_key(...)`, which resolves the confirm dialog (`src/tui/home/input.rs:1071`) and returns `Action::SwitchSessionView`; `execute_action` then sets `pending_view_switch` (`src/tui/app.rs:2924`).

Because the drain already ran this iteration, the switch is deferred to the next keypress. The sibling `pending_structured_view_open` is drained *after* `home.handle_key` (`src/tui/app.rs:2641`), which is correct, and the mouse `[Yes]` path drains after `execute_action` (`src/tui/app.rs:1380`), also correct. The keyboard path is the only one with the drain on the wrong side.

## Fix

Move the `pending_view_switch` drain (`src/tui/app.rs:2631-2634`) to immediately after the `home.handle_key`/`execute_action` block, next to the `pending_structured_view_open` drain at `:2641`. Add a keyboard-confirm test; current tests cover only the click path and the menu row.

## Provenance

Landed via #2923 (merge `dffb5228`).

### LINKED ISSUE #2926: fix(tui): wizard-created structured session never opens the structured view — #2923 follow-up
## Summary

Creating a structured session through the new-session wizard (the **Structured** toggle added in #2923) persists the session correctly but does not open the structured view. The user is left on the home screen and must manually press Enter on the row to enter the view. The PR description states "the fresh session opens the native structured view", which does not happen.

## Root cause

Nothing on the creation path sets `pending_structured_view_open`:

- `apply_creation_results` calls `dispatch_new_session_attach` (`src/tui/app.rs:2940`), which has only `LiveSend` and `Tmux | None` arms.
- `new_session_attach_mode` returns `None` for structured sessions (`resolve_session_config_for` short-circuits at `src/tui/home/mod.rs:6148`), so the call falls into the `Tmux | None` arm.
- That arm calls `attach_session`, which early-returns for `instance.is_structured()` (`src/tui/app.rs:2977`) without setting `pending_structured_view_open`.

The only setter of `pending_structured_view_open` anywhere is the `Action::OpenStructuredView` arm (`src/tui/app.rs:2918`), which the creation path never dispatches. So the drain at `src/tui/app.rs:1598` is unreachable for the creation path, and the comment at `src/tui/app.rs:1594` ("A structured session routes the post-create attach into `pending_structured_view_open`") is inaccurate.

## Fix

Dispatch `Action::OpenStructuredView(session_id)` from `dispatch_new_session_attach` when `instance.is_structured()` (this sets `pending_structured_view_open`, drained at `:1598`). Correct or remove the misleading comment at `:1594`.

## Provenance

Landed via #2923 (merge `dffb5228`).

### LINKED ISSUE #2927: fix(tui): structured-view usage meter can display over 100% — #2923 follow-up
## Summary

The structured view token usage meter (added in #2923) can render a percentage above 100% (for example `105%`) when the reported context usage exceeds the window size (`used > size`), which the ACP protocol does not forbid. The web meter caps at 100%.

## Root cause

`usage_percent` (`src/tui/structured_view/render.rs:557-561`) computes:

```rust
((usage.used as f64 / usage.size as f64) * 100.0).round() as u64
```

with no upper bound. The web equivalent caps at 100 (`web/src/components/acp/Composer.tsx:1526`, `Math.min(100, ...)`). `used > size` is reachable from the wire: the `UsageUpdate` handler (`src/acp/acp_client.rs:4180-4189`) copies `used`/`size` verbatim into `SessionUsage` with no bounds check.

## Fix

Cap the result at 100 (`.min(100)`), matching the web, and add a test with `used > size`.

## Provenance

Landed via #2923 (merge `dffb5228`).

njbrake:
## Description

Makes the TUI structured view feel like a native coding agent instead of a separate bolt-on surface, and fixes the entry-point bugs that made it hard to reach at all.

**Interaction model (Claude Code-style):**
- Selecting a structured session streams its transcript straight into the preview pane, with the info header (and its `i` toggle) intact. No more static "press Enter to open" placeholder, and no flash while the preview connects.
- Enter (or double-click) enters the session; you land in the composer and type immediately. Ctrl+Q drops back to the read-only preview, mirroring live-send's exit chord. The golden border on both boxes is the "entered" cue.
- One pane, no Tab toggle between chat and input. Scrollback is the wheel or PageUp/PageDown from the composer; drag-to-select copies transcript text through the same machinery terminal previews use.
- Esc interrupts a generating turn (and is a no-op when idle); Shift+Tab opens the permission-mode picker.
- Approvals are modal: a pending permission grabs the keyboard and resolves with `a` / `A` / `d` in place. Elicitation questions auto-open their answer menu. The composer still swallows approval letters while typing (focus-isolation guarantee preserved).
- Transcript renders like Claude Code: no `you`/`aoe` speaker labels, user turns on a highlighted background, soft breaks keep their line breaks.

**Structural refactor:**
The view is owned by `HomeView` as first-class preview content rather than an App-level overlay, so the preview placeholder, info header, layout caching, and drag-select compose with it by construction. Transcript lines are pre-wrapped at render width (span-preserving), making scroll math exact and feeding the home selection extractor. `App` keeps only the async sides: the cancel-safe WebSocket pump in the main `select!`, the debounced preview-on-select mount, and key routing while entered.

**Entry-point fixes:**
- Keyboard-confirmed view switch fires immediately instead of one keystroke late (closes #2925).
- Wizard-created structured sessions open their view (closes #2926).
- Usage meter caps at 100% (closes #2927).
- The "no structured view daemon" dead end now offers to start a localhost daemon (confirm dialog from the home flow; offer screen in `aoe acp attach`); the view switch auto-starts one after its consented confirm.
- Confirm dialogs size their height to the wrapped body text instead of clipping.
- Tab on a structured row explains there is no tmux pane instead of silently doing nothing.

**Rollout gate:** the wizard's Structured toggle is now behind `acp.offer_structured_in_new_session` (single-source schema setting), default off, so users don't see the surface until it is ready. Switching an existing session's view and opening structured sessions are unaffected. The web wizard's structured toggle is not yet gated by this setting; that needs its own coverage-matrix entry and is left as follow-up.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: no web code changes; the new `acp.offer_structured_in_new_session` setting renders through the existing schema-driven settings surface, and gating the web wizard on it is called out as follow-up work.

**TUI / CLI (e2e test)**

New `tests/e2e/structured_tui_flows_e2e.rs` covers the wizard-create flow (structured session opens embedded with the sidebar visible) and the immediate keyboard-confirmed view switch, both against a live daemon plus the shared fake ACP agent. `acp_focus_isolation_e2e.rs` was rewritten for the modal-approval model (auto-focus, only decision keys act, `a` resolves). Unit tests cover the input model (Ctrl+Q, native Esc, inert Tab, Shift+Tab), wrap geometry, newline honoring, user-turn highlighting, confirm-dialog sizing, the usage-meter cap, and the Tab refusal toast.

Verified live end-to-end with a real `claude-agent-acp` worker: preview-on-select, enter/exit, prompt round-trips with streaming, a real Write-tool approval resolved in place, drag-select producing an OSC52 clipboard write, wheel scrollback, and the daemon auto-start flows.

`cargo fmt`, `cargo clippy --features serve`, and TUI-only builds are clean. Lib suite passes except three pre-existing chmod-based fixtures that cannot fail while running as root in the dev sandbox (they fail on clean main there too).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Fable 5)

Implemented via an interactive Claude Code session driven by @njbrake, with repeated live verification in a sandboxed TUI against a real claude ACP worker.

## Summary

- Made the TUI “Structured view” a first-class preview-pane experience: selecting a structured session streams its transcript directly in the preview (keeping the info header), and you can then enter the full composer from that preview.
- Clarified and tightened entry/exit controls: **Enter** (or double-click) opens the composer, and **Ctrl+Q** reliably returns to the read-only preview immediately.
- Improved the preview experience substantially: native-style scrolling, better transcript rendering, and drag-selection support—plus improved keyboard/mouse behavior for switching and focusing.
- Refactored responsibilities: structured preview ownership and selection behavior moved into **`HomeView`**, while **`App`** focuses on async WebSocket handling and routing events/input when the embedded composer is active.
- Fixed several timing and workflow edge cases:
  - view-switch confirmation now applies within the same keypress (not the next one),
  - wizard-created structured sessions open structured view automatically,
  - structured-session daemon startup is handled more smoothly (including “start then open”),
  - structured switching/cleanup timing and related focus behavior were corrected.
- Added/expanded modal and permission workflows in the structured view, with a clearer “approval modal” key/focus behavior and updated focus isolation so decision keys don’t leak into the composer.
- Added UX and robustness polish:
  - confirm dialog height now adjusts dynamically so long prompts aren’t clipped,
  - structured-row **Tab** behavior now gives clear feedback when it can’t attach (structured rows have no tmux pane),
  - structured view usage-meter is now capped at 100% when reported usage exceeds the context window.
- Introduced an opt-in config flag: **`acp.offer_structured_in_new_session`** gates showing the “Structured view” option in the new-session dialog (default off).

## Benefits

This update makes it much easier to browse structured sessions without leaving the main screen, then confidently switch into composing when needed—while reducing confusing focus jumps, delayed/next-keypress switching, and awkward daemon/permission flows. It also improves reliability for keyboard-driven workflows and avoids common UI clipping/feedback issues.

## Technical details (optional)

- Added an embedded structured-view component with “mounted preview” vs “active composer” modes, including event streaming and transcript geometry for scrolling and drag selection.
- Updated structured input routing and focus semantics (including **Ctrl+Q**, **Esc** interrupt/no-op behavior, modal approval focus isolation, and page scrolling).
- Expanded E2E and unit coverage for structured entry flows, immediate keyboard-confirm switching, and modal keyboard isolation.

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -378,6 +378,17 @@ fn default_show_spans() -> bool {
 #[derive(Debug, Clone, Serialize, Deserialize, SettingsSection)]
 #[setting_section(name = "acp", category = "Acp")]
 pub struct AcpConfig {
+    /// Show the "Structured view" toggle in the new-session dialog. The
+    /// structured view is still maturing, so it is hidden by default;
+    /// turn this on to opt into creating structured-view sessions from
+    /// the new-session flow. Switching an existing session's view and
+    /// opening already-structured sessions are unaffected.
+    #[serde(default)]
+    #[setting(
+        label = "Offer structured view when creating a session",
+        widget = "toggle"
+    )]
+    pub offer_structured_in_new_session: bool,
     /// Acp agent used when --agent is not specified (e.g. aoe-agent,
     /// claude-code, gemini).
     #[serde(default = "default_agent")]
@@ -524,6 +535,7 @@ fn default_silent_orphan_grace_secs() -> u32 {
 impl Default for AcpConfig {
     fn default() -> Self {
         Self {
+            offer_structured_in_new_session: false,
             default_agent: default_agent(),
             max_concurrent_workers: default_max_workers(),
             replay_events: default_replay_events(),
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -195,6 +195,18 @@ pub struct App {
     /// the daemon switch POST (awaited; the sync handler can't).
     #[cfg(feature = "serve")]
     pending_view_switch: Option<String>,
+    /// Set by `Action::StartDaemonThenOpenStructured` (the Yes on the
+    /// "start a local daemon?" confirm) so the async loop can spawn the
+    /// daemon, wait for health, and then open the structured view.
+    #[cfg(feature = "serve")]
+    pending_daemon_start_open: Option<String>,
+    /// Debounce for structured preview-on-select: the session the cursor
+    /// settled on and when, so rapid list navigation doesn't connect a
+    /// WebSocket per keystroke. The mounted view itself lives on
+    /// `HomeView::structured_preview` (it is preview content); this App
+    /// side only drives the async mount/unmount.
+    #[cfg(feature = "serve")]
+    preview_mount_pending: Option<(String, std::time::Instant)>,
     /// Version of the install currently being attempted (auto or manual).
     /// Set when the install task is spawned; transferred to
     /// `last_installed_version_in_session` on confirmed success in
@@ -479,6 +491,10 @@ impl App {
             #[cfg(feature = "serve")]
             pending_structured_view_open: None,
             #[cfg(feature = "serve")]
+            pending_daemon_start_open: None,
+            #[cfg(feature = "serve")]
+            preview_mount_pending: None,
+            #[cfg(feature = "serve")]
             pending_view_switch: None,
             pending_install_version: None,
             last_installed_version_in_session: None,
@@ -535,10 +551,25 @@ impl App {
     /// pane caret, not a local IME candidate window, so there's nothing for
     /// the early Hide to protect.
     fn draw(&mut self, terminal: &mut Terminal<CrosstermBackend<std::io::Stdout>>) -> Result<()> {
-        let skip_hide = skip_predraw_cursor_hide(
-            self.home.live_send.is_some(),
-            self.home.has_non_live_send_overlay(),
-        );
+        // An ACTIVE embedded structured view sets a composer caret every
+        // frame, just like the live-send preview caret: hiding it
+        // before each ~30fps redraw makes it strobe (the reported "cursor
+        // blinks really fast"). Skip the pre-draw Hide while it's active,
+        // the same treatment live-send gets. A preview shows no caret, so
+        // it needs no skip.
+        #[cfg(feature = "serve")]
+        let embedded_active = self
+            .home
+            .structured_preview
+            .as_ref()
+            .is_some_and(|v| v.is_active());
+        #[cfg(not(feature = "serve"))]
+        let embedded_active = false;
+        let skip_hide = embedded_active
+            || skip_predraw_cursor_hide(
+                self.home.live_send.is_some(),
+                self.home.has_non_live_send_overlay(),
+            );
         crossterm::execute!(
             terminal.backend_mut(),
             crossterm::terminal::BeginSynchronizedUpdate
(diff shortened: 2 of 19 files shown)


njbrake:
Structured view is still definitely not perfect and needs more work. However, I think this is a step in the right direction so I'll merge this one in as I keep cleaning the design up.

PR #2961, opened 2026-07-18
feat(trash): track the teardown as a first-class claimed operation

njbrake:
## Description

Follow-up to #2945, closing the gap @jerome-benoit identified there: the trash teardown ran as fire-and-forget compensation, observable only through its side effects. It now participates in the #2541 claim protocol as durable state.

Replaces #2960, which was auto-closed when its base branch was deleted on #2945's merge; same content, now based on `main`.

- `ClaimOp::Trash` is written under the storage flock right after the trash marker at all three trash sites (TUI, server handler, CLI remove), so "teardown in flight" is durable state peers can observe.
- It is deliberately the weakest claim. `try_claim` lets Purge and Restore seize a fresh Trash claim, so hitting `d` and immediately restoring stays instant; Trash never seizes a fresh Purge or Restore.
- The teardown yields to a seizure at both of its decision points: the pre-move gate skips the move when a fresh non-Trash claim holds the still-trashed row, and `commit_trash_relocation` treats a seized row as superseded (undoing the move) while releasing the teardown's own claim on every relocation outcome. No-relocation terminal paths release through `release_trash_claim`; a crash mid-teardown falls back to the existing 10 minute `OP_CLAIM_TTL` self-heal.

Compat note: `ClaimOp` has no `#[serde(other)]` fallback, so an aoe binary predating the variant fails the whole-row deserialize on `op:"trash"`; `Storage::load` quarantines the row to `sessions.corrupt.jsonl` and the session is temporarily invisible to that binary for the life of the claim. The 10 minute TTL bounds how long the claim stays fresh, but the field only clears when a newer binary next rewrites the row, so the invisibility ends at that rewrite, not on a timer. If the older binary writes storage while the row is invisible, its save drops the row from `sessions.json` (it survives only in the quarantine sidecar). Same exposure class as the #2541 Purge/Restore variants against pre-#2541 binaries.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the claim protocol is exercised at the unit and storage level where the interleavings can be forced deterministically (seizure matrix in `instance.rs`, acquire/release/commit gates in `claim.rs`, a seized-row teardown skip against real durable storage in `trash.rs`, and the TUI setting the durable claim on `d` in `home/tests.rs`). An e2e cannot deterministically pause a teardown mid `docker stop` to interleave a claim seizure.

## Testing

`cargo test`: 4127 passed. One failure (`restart_selected_session_surfaces_resume_failed_after_async_restart`) fails identically on a clean `main` checkout in this environment, so it is pre-existing and unrelated. `cargo fmt` clean, `cargo clippy --all-targets` adds no new warnings, `cargo check --features serve` compiles.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Fable 5 via Claude Code

Designed as the follow-up promised in the #2945 review discussion.

## Summary

- Added a durable, persisted “trash teardown” operation claim (`ClaimOp::Trash`) so in-progress trash cleanup is observable and recoverable across processes and crashes.
- Wired this new claim into the main entry points (TUI, server, and CLI) so they all mark “trash teardown in flight” using the same atomic storage/lock flow.
- Updated teardown behavior so trash cleanup can be safely superseded by newer purge/restore work, and so teardown won’t leave behind stranded claims.
- Ensured the in-flight trash claim is best-effort released on terminal outcomes (including when relocation is skipped/fails), with centralized release logic.
- Strengthened relocation safety by re-checking durable state after stopping the container, and skipping relocation if the durable record shows the work was no longer safe to move.
- Expanded storage/session claim logic and added tests for interleavings, teardown handling, and TUI claim operations; also updated compatibility documentation for older binaries that can’t parse `op:"trash"` during the TTL window.

## Benefits

- More reliable crash recovery and concurrency handling for trash cleanup.
- Reduced risk of conflicting cleanup operations and stranded worktrees during purge/restore races.
- Clearer operational visibility into what “trash teardown” is doing and whether it completed, skipped, or was superseded.

### DIFF
--- a/src/session/claim.rs
+++ b/src/session/claim.rs
@@ -1,10 +1,11 @@
-//! Cross-process purge/restore claim decisions, shared by the CLI, the serve
-//! daemon, and the TUI. Every destructive/irreversible phase (purge teardown,
-//! restore worktree move) runs its slow work on an UNLOCKED snapshot; these
-//! helpers make the claim check-and-set (and the final commit) atomic under the
-//! storage flock, the only serialization point visible across processes. Living
-//! in the neutral `session` layer keeps the three surfaces from reaching into
-//! `cli` for shared logic. See #2534, #2541.
+//! Cross-process purge/restore/trash claim decisions, shared by the CLI, the
+//! serve daemon, and the TUI. Every destructive/irreversible phase (purge
+//! teardown, restore worktree move, trash container stop + relocation) runs
+//! its slow work on an UNLOCKED snapshot; these helpers make the claim
+//! check-and-set (and the final commit) atomic under the storage flock, the
+//! only serialization point visible across processes. Living in the neutral
+//! `session` layer keeps the three surfaces from reaching into `cli` for
+//! shared logic. See #2534, #2541.
 
 use super::{ClaimOp, Instance};
 use chrono::{DateTime, Utc};
@@ -170,49 +171,30 @@ pub(crate) fn finalize_restore_commit(
     RestoreCommit::Committed
 }
 
-/// Outcome of the trash claim acquisition, run under the flock right after
-/// the trash marker is durably written at every trash site (TUI, server,
-/// CLI). See [`decide_trash_claim`].
-#[derive(Debug, PartialEq)]
-pub(crate) enum TrashClaimDecision {
-    /// The Trash claim is set; the teardown is observable as in flight.
-    Claimed,
-    /// A fresh purge or restore claim holds the row, so the teardown runs
-    /// unclaimed and relies on its pre-move re-check and the locked
-    /// relocation commit alone.
-    PeerHeld(ClaimOp),
-    /// The row is gone from disk (a peer removed it).
-    AlreadyGone,
-}
-
-/// Mark a freshly-trashed row's teardown as in flight by taking the Trash
-/// claim, run inside a `storage.update` closure. The claim is durable state
-/// peers can observe (and that purge/restore seize, see
-/// [`Instance::try_claim`]); it is released by
-/// [`commit_trash_relocation`] on the relocation paths and
-/// [`release_trash_claim`] on the no-relocation ones, and the
-/// [`Instance::OP_CLAIM_TTL`] self-heal covers a crash in between.
-/// Best-effort by design: a `PeerHeld` row still tears down, gated by the
-/// re-check and the locked commit.
-pub(crate) fn decide_trash_claim(
-    all: &mut [Instance],
-    id: &str,
-    now: DateTime<Utc>,
-) -> TrashClaimDecision {
-    match all.iter_mut().find(|i| i.id == id) {
-        None => TrashClaimDecision::AlreadyGone,
-        Some(row) => match row.try_claim(ClaimOp::Trash, Instance::OP_CLAIM_TTL, now) {
-            Ok(()) => TrashClaimDecision::Claimed,
-            Err(op) => TrashClaimDecision::PeerHeld(op),
-        },
-    }
-}
-
 /// Release the Trash claim on a teardown's no-relocation terminal paths
 /// (`Skipped` / `Failed`), run inside a `storage.update` closure.
 /// Ownership-guarded, so a claim a purge or restore seized in the meantime is
 /// never cleared. The relocation paths release through
 /// [`commit_trash_relocation`] instead.
+///
+/// Trash claim lifecycle: each trash site (TUI `trash_session_by_id` via the
+/// same-flock `apply_user_action_with` hook, the server trash handler's
+/// persist closure, CLI `remove`) sets the claim with
+/// [`Instance::try_claim`] in the SAME `storage.update` that writes the
+/// trash marker, so a peer can never read a trashed row without its
+/// in-flight claim. A refused claim (fresh peer purge/restore) still tears
+/// down, gated by the pre-move re-check and the locked relocation commit.
+///
+/// TTL asymmetries, all deliberate: the teardown's gates
+/// ([`Instance::is_seized_by_fresh_peer_claim`]) only yield to a FRESH
+/// peer claim, while seizure of a Trash claim ignores its age entirely, so
+/// both a stale seizer and a teardown whose worker died without a release
+/// (the TUI's `Disconnected` drain, the server's join-error arm) converge
+/// through the [`Instance::OP_CLAIM_TTL`] self-heal rather than an explicit
+/// handoff. The load-time `reconcile_trashed_location` pass does not consult
+/// `op_claim` at all; racing an in-flight peer teardown is benign because
+/// both sides attempt the same idempotent move and the loser fails on
+/// "target exists".
 pub(crate) fn release_trash_claim(all: &mut [Instance], id: &str) {
     if let Some(row) = all.iter_mut().find(|i| i.id == id) {
         row.clear_op_claim_if_owned(ClaimOp::Trash);
@@ -264,11 +246,7 @@ pub(crate) fn commit_trash_relocation(
             // fresh Purge or Restore claim (which seized it) does. Either
             // way this commit is a terminal teardown path, so any Trash
             // claim still owned is released (ownership-guarded).
-            let seized = matches!(
-                &row.op_claim,
-                Some(c) if c.op != ClaimOp::Trash && (now - c.at) < Instance::OP_CLAIM_TTL
-            );
-            if !row.is_trashed() || seized {
+            if !row.is_trashed() || row.is_seized_by_fresh_peer_claim(now) {
                 row.clear_op_claim_if_owned(ClaimOp::Trash);
                 return RelocationCommit::Superseded;
             }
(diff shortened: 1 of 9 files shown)


src/tui/home/operations.rs
```diff
@@ -1589,6 +1589,42 @@ impl HomeView {
...
+        // against storage because `merge_user_action_diff` deliberately drops
+        // `op_claim` (#2541). Best-effort: an unclaimed teardown still runs,
+        // gated by its re-check and the locked relocation commit.
+        if let Some(storage) = self
+            .instances
+            .get(id)
+            .map(|i| i.source_profile.clone())
+            .and_then(|profile| self.storages.get(&profile))
+        {
+            let claim = storage.update(|insts, _groups| {
```
jerome-benoit: This is a second, separate `storage.update` from the `apply_user_action(id, |inst| inst.trash())` above, so the trash marker and the Trash claim land in two distinct flock acquisitions. CLI (`remove.rs`) and server (`sessions.rs`) set both `trash()` and `try_claim(Trash)` in one closure, so a peer can never read a trashed row without the claim; here it can, for the window between the two writes, which is exactly the "teardown in flight is durable observable state" property this PR adds. The claim is best-effort so this is not a correctness bug, but please fold `inst.trash()` + `try_claim(Trash)` into a single `storage.update` for the TUI (open-code it like the CLI, bypassing `apply_user_action` for this path, or give `apply_user_action` a same-flock post-mutation hook). The comment's `#2541`-drops-`op_claim` reason justifies bypassing the diff, not the two-write split.

src/session/instance.rs
```diff
@@ -454,15 +454,28 @@ pub enum SessionBucket {
...
+/// cross-process primitive that serializes them so none tears down (or moves)
+/// state another is authoritative over. See #2541.
+///
+/// `Trash` is deliberately the weakest claim: it marks "teardown in flight"
+/// so peers can observe it, but it never blocks user intent. A purge or
+/// restore seizes a fresh `Trash` claim (see [`Instance::try_claim`]) and the
+/// teardown yields via its pre-move re-check and the locked relocation
+/// commit, so a `d` followed by an immediate restore stays instant.
+///
+/// Compat: `trash` claims in `sessions.json` are not parseable by aoe
```
jerome-benoit: This says older binaries "cannot observe the claim", but the effect is stronger. `ClaimOp` has no `#[serde(other)]` fallback, so an older binary reading `op:"trash"` fails to deserialize the whole `Instance` row; `Storage::load` then skips and quarantines that row to `sessions.corrupt.jsonl` rather than returning it. So the session is temporarily invisible to the older binary for the life of the claim (up to the 10-minute TTL), after which a newer binary clears the claim and the row is readable again. The exposure is acceptable (mid-teardown, TTL-bounded), but please reword this note (and the PR description, and the `trash_claim_serde_roundtrip` comment) to say the row is quarantined and the session is temporarily invisible to an older binary, rather than "cannot observe the claim".

src/session/claim.rs
```diff
@@ -208,13 +260,21 @@ pub(crate) fn commit_trash_relocation(
...
         None => RelocationCommit::AlreadyGone,
         Some(row) => {
-            let claim_held =
-                matches!(&row.op_claim, Some(c) if (now - c.at) < Instance::OP_CLAIM_TTL);
-            if !row.is_trashed() || claim_held {
+            // The teardown's own Trash claim does not block its commit; a
+            // fresh Purge or Restore claim (which seized it) does. Either
+            // way this commit is a terminal teardown path, so any Trash
+            // claim still owned is released (ownership-guarded).
+            let seized = matches!(
```
jerome-benoit: This "seized by a fresh non-Trash peer" predicate is duplicated verbatim in `teardown_may_relocate` (`src/session/trash.rs`). A fourth `ClaimOp` or a change to the TTL semantics would have to be edited in lockstep in both places. Consider extracting `Instance::is_seized_by_fresh_peer_claim(&self, now) -> bool` and calling it from both. (Leave `try_claim`'s predicate inline; it additionally excludes `want`.) This is also CodeRabbit's nit here.

jerome-benoit (changes requested):
Thanks, this is an elegant extension of the #2541 lattice: modeling the teardown as the weakest claim (encoded by the single `&& c.op != ClaimOp::Trash` clause in `try_claim`) reuses the seize/gate/commit machinery cleanly, the 3x3 seizure matrix and the race analysis (d-then-restore, d-then-purge, restore-between-gate-and-commit) hold, and it closes exactly the #2945 gap. I verified it locally: `cargo fmt`, `cargo clippy --features serve`, `RUSTDOCFLAGS="-D warnings" cargo doc --features serve`, and the claim/trash/instance/home tests all pass.

Two changes before merge, plus nits.

## Requested changes

1. **The TUI sets the trash marker and the Trash claim in two separate flock writes** (inline on `src/tui/home/operations.rs`). CLI (`remove.rs`) and server (`sessions.rs`) set both `trash()` and `try_claim(Trash)` in one closure; the TUI split leaves a window where a peer reads a trashed row with no claim, which is the exact "teardown in flight is observable durable state" property this PR adds, left open on the most common surface. Fold it into a single `storage.update`.

2. **The compat note mischaracterizes the failure mode** (inline on `src/session/instance.rs`). `ClaimOp` has no `#[serde(other)]`, so an older binary fails the whole-row deserialize and `Storage::load` quarantines the row: the session is temporarily invisible (up to the 10-minute TTL), not "cannot observe the claim." The exposure is acceptable, but the note (and the PR description) should say so.

## Nits

- Stale docstrings: the `OpClaim` struct doc, the `op_claim` field doc, and the `claim.rs` module doc still say "purge or restore"; add the trash teardown.
- Extract the duplicated "seized by a fresh non-Trash peer" predicate (inline on `src/session/claim.rs`) into an `Instance` helper shared with `teardown_may_relocate` (this is also CodeRabbit's nit).
- One comment would help: the teardown gates use freshness while seizure ignores TTL, so both the stale-seizer window and a crashed teardown worker (the `Disconnected` / join-`Err` paths do not release) rely on the `OP_CLAIM_TTL` self-heal, by design. `reconcile_trashed_location` also does not consult `op_claim`, unlike `teardown_may_relocate` (benign asymmetry worth a note).
- Test gaps: the Skipped/Failed `release_trash_claim` path is not asserted at the storage level, and the `d`-then-restore / `d`-then-purge handoffs are covered at the unit level but not end-to-end through the TUI.

No hard correctness blocker; the two requested changes are a cross-surface consistency fix and a doc-accuracy fix.

jerome-benoit (approved):
Re-reviewed on `1551cac35`. My three previous blockers are resolved: the trash marker and the `Trash` claim now land in one flock via the `apply_user_action_with` post-disk hook; the compat note is accurate (`ClaimOp` has no `#[serde(other)]`, so an old binary fails the whole-row deserialize and `Storage::load` quarantines it to `sessions.corrupt.jsonl`); and the seizure predicate is shared through `is_seized_by_fresh_peer_claim`.

I also did a full design/soundness pass. The model holds: I traced every terminal release path (TUI drain, CLI, both server arms) and every seizure edge (a fresh Purge/Restore seizes a Trash claim, Trash never seizes them, clears are ownership-guarded, worker-death falls to the TTL self-heal). No path strands a blocking claim, double-releases, relocates a seized row, or loses the trash marker, and it actually tightens the pre-existing purge-vs-teardown race. Verified locally on this commit: `cargo fmt`, `cargo clippy --features serve --all-targets`, the trash + claim suites, and `RUSTDOCFLAGS="-D warnings" cargo doc --features serve` all clean.

Approving. Non-blocking polish, for an amend or follow-up:

- **[minor]** `trash_session` inlines the identical `persist_session_update("trash-claim-release", ...)` block in both the `Failed` and `Skipped` arms; the CLI factored the equivalent into `release_trash_claim_best_effort`. Worth one shared helper so the two surfaces stay symmetric.
- **[minor]** Claim-refusal logging is uneven: the TUI logs the refusing holder, but the CLI (`remove.rs`) and server (`trash_session`) drop `Err(holder)` with `let _ =`. The same `if let Err(holder) { info!(...) }` at all three sites makes the unclaimed-teardown case observable everywhere.
- **[nit]** The ~24-line lifecycle/TTL writeup on `release_trash_claim` is the best overview in the PR but sits on a 3-line helper; it reads better on the `ClaimOp` doc or the module header.
- **[nit]** The `merge_user_action_diff` comment still says "in-flight purge/restore"; extend it to "purge, restore, or trash" since the Trash claim relies on the same drop.
- **[nit]** The compat note's "at most 10 minutes, after which a newer binary clears the claim" reads as an automatic timer; the field only clears when a newer binary next rewrites that row.

Thanks for the tight iteration.

njbrake:
Thanks @jerome-benoit! I appreciate your perspective on this PR.

Issue #1054, opened 2026-05-12
ACP: runner-side protocol termination (deferred from #1045)


Seluj78:
> [!NOTE]
> This issue was opened with AI assistance.

## Background

PR #1045 made ACP workers survive \`aoe serve --stop\` by introducing a per-session \`aoe __cockpit-runner\` shim that owns the agent subprocess. The shim is a pure byte-proxy: it forwards bytes between the currently-connected daemon and the agent's stdio, with a 256-line ring buffer for agent→daemon notifications emitted while no daemon is attached.

The mid-turn-reattach follow-up (this branch) is a daemon-side fix:

- \`AcpClient::attach\` skips \`session/new\`/\`session/load\` on reconnect and reuses the runner-cached ACP session id.
- A resume-idle watchdog synthesizes \`Event::Stopped { reason: \"reattach_idle\" }\` after a quiet window, because the agent's eventual response to the orphaned in-flight \`session/prompt\` carries a request id that was issued by the previous daemon's client and is dropped silently by the transport.
- The reconciler publishes a synthetic \`Stopped { reason: \"orphaned_at_restart\" }\` before the fresh-spawn fallback so the UI doesn't stay stuck on \"thinking\" after a daemon crash (also fixes the equivalent main-branch case where there is no runner).

This works, but it's heuristic. The 10s grace is fine in practice but it IS a timeout; chunks past the 256-line ring boundary during a long detach are lost; and the daemon still pays the cost of a fresh \`initialize\` request against the agent on every reconnect.

## Proposal: make the runner the ACP protocol terminator

The architecturally clean fix is to move the ACP protocol speaker into the runner. The daemon would talk a simpler control protocol to the runner over the existing unix socket (or a sibling control socket).

Concretely the runner would:

1. **Own the ACP handshake.** Send \`initialize\`, then \`session/new\` (or \`session/load\` with the registry-cached id) once, on first agent spawn. Cache the responses. On subsequent daemon connects, do not forward those requests through.
2. **Track JSON-RPC request ids.** Translate between daemon-local ids and runner-tracked ids so the daemon and the agent can each have their own counter. Match responses correctly across reconnects.
3. **Forward notifications.** \`session/update\` notifications go straight to the currently-connected daemon (with ring-buffer replay on reattach, as today).
[...]

PR #2975, opened 2026-07-18
feat(acp): runner-side control channel with native turn-complete (#1054 Phase A)

Seluj78:

## Description

Phase A of runner-side ACP protocol termination (#1054). Today, when the daemon reattaches to a runner whose agent was mid-turn across an `aoe serve` restart, the agent's response to the orphaned `session/prompt` carries a JSON-RPC id the new daemon's client never issued, so the transport drops it silently. The daemon papers over this with a 30s resume-idle watchdog that synthesizes `Stopped { reason: "reattach_idle" }`, leaving the UI stuck on "thinking" for up to 30s after the turn has actually finished.

This introduces a typed control channel between the daemon and the `aoe __acp-runner` shim so the runner reports turn completion natively instead of the daemon guessing:

- The runner binds a sibling `<id>.control.sock` (before the main relay socket, so it is connectable by the time the daemon dials in). Frames are length-prefixed JSON (`ControlBody`: `Hello` / `PromptCompleted` / `Attach`).
- The runner tracks the daemon-issued `session/prompt` request id and, when it observes the agent's matching response, emits a `PromptCompleted` frame carrying the ACP `stopReason`. Completion events produced while no daemon is attached are buffered and replayed on the next control attach, so a turn that finishes during a daemon restart is not lost.
- On a mid-flight resume the daemon dials the control socket; a native `prompt_complete` claims the shared terminal guard and fires `Stopped { reason: "prompt_complete" }`, and the 30s resume-idle watchdog stands down. A runner too old to bind the control socket leaves the guard unclaimed, so the watchdog behaves exactly as before (fully backward compatible).

The full design (7 resolved architecture questions, phases A through D) was worked out via a three-model design debate. This PR is Phase A. It does not close - #1054; follow-up sub-issues track Phase B (runner owns the ACP handshake), Phase C (server-method proxying, `<id>.sock` retirement, `RUNNER_VERSION` bump + migration), and Phase D (long-detach durability + cleanup).

Relates to #1054

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view ACP session and send a prompt that keeps the agent working for a bit.
- [ ] While the turn is in flight, stop the daemon (`aoe serve --stop`) and start it again (`aoe serve`).
- [ ] Confirm the session clears "thinking" as soon as the agent's response lands (goes Idle promptly), rather than after the ~30s pause the old resume-idle watchdog imposed.
- [ ] Confirm a normal turn (no restart) still completes exactly once, with no duplicate stop.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, a three-model design debate, the plan, and the implementation were drafted by Claude under my direction. I made the scope and design calls (full runner-terminator target, phased rollout, in-memory over disk durability, hand-rolled router over the crate), reviewed each commit, and ran the manual test steps.

## Summary

- Added a sibling Unix “control” socket for `aoe __acp-runner` so the runner can notify the daemon when an in-flight `session/prompt` turn finishes.
- Implemented a small typed, length-prefixed JSON control protocol (versioned hello/attach handshake) over that socket, including safe frame sizing and EOF handling.
- Updated runner/daemon reattach for in-flight turns: when supported, the daemon now uses the runner’s native `prompt_completed` notification to end the turn promptly with the correct stop reason (for example `prompt_complete`, or `rate_limited` when applicable), instead of the older idle/timeout-based fallback.
- Added runner-side prompt request tracking plus an in-memory “no-daemon gap” buffer so completions aren’t lost while the daemon detaches and reconnects; buffering is intentionally gated to avoid interfering when the daemon is still attached.
- Extended connection setup/teardown logic: control sockets are created as `<id>.control.sock`, runner control accept loops are managed cleanly, and stale sibling control sockets are removed during session deletion for older runners.
- Added unit and real-socket integration tests for control framing, version mismatch behavior, completion delivery/buffering, and legacy/socketless fallback; mid-turn restart coverage is deferred to later phases.

## Benefits

Reattached sessions can now complete faster and more accurately using real runner completion signals and correct stop reasons, while maintaining compatibility with older runners that don’t implement the control channel.

## Future enhancement

Phase A buffers completions only in memory for the daemon-detach gap; durable completion storage and full “restart mid-turn” end-to-end coverage are planned for later phases.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -2158,6 +2158,9 @@ impl AcpClient {
                 default_effort,
                 default_mode,
                 mcp_servers,
+                // Direct stdio agents have no runner and thus no control
+                // channel; the task owns its own terminal guard.
+                None,
             )
             .instrument(conn_span),
         );
@@ -2242,6 +2245,35 @@ impl AcpClient {
         let pending_for_task = pending_responders.clone();
         let expected_agent = ExpectedAgent::from_command(&install_binary);
 
+        // #1054 Phase A: for a mid-flight resume, the agent's response to
+        // the orphaned `session/prompt` carries an id this client never
+        // issued and the crate transport drops it, which is what the 30s
+        // resume-idle watchdog exists to paper over. Dial the runner's
+        // sibling control socket; if it speaks the control protocol, its
+        // native `prompt_complete` claims the shared terminal guard and
+        // fires `Stopped` deterministically, and the watchdog stands down.
+        // A runner too old to bind the control socket falls back to the
+        // watchdog unchanged.
+        let external_terminal_guard = if matches!(
+            mode,
+            ConnectMode::Resume {
+                in_flight_turn: true,
+                ..
+            }
+        ) {
+            let guard = Arc::new(std::sync::atomic::AtomicBool::new(false));
+            attach_runner_control(
+                &socket_path,
+                event_tx.clone(),
+                session_label.clone(),
+                guard.clone(),
+            )
+            .await;
+            Some(guard)
+        } else {
+            None
+        };
+
         let (ready_tx, ready_rx) = oneshot::channel::<Result<(), AcpError>>();
 
         // See the sibling spawn in `spawn`: the connection task runs inside
@@ -2267,6 +2299,7 @@ impl AcpClient {
                 default_effort,
                 default_mode,
                 mcp_servers,
+                external_terminal_guard,
             )
             .instrument(conn_span),
         );
@@ -3167,6 +3200,125 @@ async fn wait_for_socket(
         }
         tokio::time::sleep(std::time::Duration::from_millis(delay_ms)).await;
         delay_ms = (delay_ms * 2).min(200);
+    }
+}
+
+/// Dial the runner's sibling control socket (#1054 Phase A) and, if it
+/// speaks the control protocol, spawn a reader that turns the runner's
+/// native `PromptCompleted` into `Event::Stopped { reason }`. The reader
+/// CAS-claims `terminal_guard` so the caller's resume-idle and
+/// between-prompt watchdogs stand down for the adopted turn. Best-effort:
+/// a connect failure or an unrecognized/absent control socket (a runner
+/// too old to bind it) leaves the guard unclaimed so the legacy watchdog
+/// still fires.
+async fn attach_runner_control(
+    main_socket: &std::path::Path,
+    event_tx: mpsc::Sender<Event>,
+    session_label: String,
+    terminal_guard: Arc<std::sync::atomic::AtomicBool>,
+) {
+    use super::control_protocol::{self, ControlBody};
+    use std::sync::atomic::Ordering;
+
+    let control_path = crate::process::worker::control_socket_sibling(main_socket);
+    // The runner binds the control socket before the main relay socket,
+    // which the caller already waited for, so this connect is effectively
+    // immediate; the bound only guards a wedged runner.
+    let bound = std::time::Duration::from_secs(5);
+
+    let stream =
+        match tokio::time::timeout(bound, tokio::net::UnixStream::connect(&control_path)).await {
+            Ok(Ok(s)) => s,
+            _ => {
+                debug!(
+                    target: "acp.protocol",
+                    session = %session_label,
+                    "no runner control socket; using resume-idle watchdog"
+                );
+                return;
+            }
+        };
+
+    let (mut read_half, mut write_half) = stream.into_split();
+
+    // The runner greets with Hello on accept; require a matching version
+    // before trusting the channel.
+    let handshake_ok = matches!(
+        tokio::time::timeout(bound, control_protocol::read_frame(&mut read_half)).await,
+        Ok(Ok(Some(ControlBody::Hello { control_protocol_version, .. })))
+            if control_protocol_version == control_protocol::CONTROL_PROTOCOL_VERSION
+    );
+    if !handshake_ok {
+        debug!(
+            target: "acp.protocol",
+            session = %session_label,
+            "runner control handshake failed; using resume-idle watchdog"
+        );
+        return;
+    }
+    let _ = control_protocol::write_frame(
+        &mut write_half,
+        &ControlBody::Attach {
+            control_protocol_version: control_protocol::CONTROL_PROTOCOL_VERSION,
+        },
+    )
+    .await;
+
+    info!(
+        target: "acp.protocol",
+        session = %session_label,
+        "runner control channel attached; native turn-complete active"
+    );
+
+    tokio::spawn(async move {
+        // Hold the write half so the runner does not observe the control
+        // socket close; Phase A sends nothing further on it.
+        let _keep_alive = write_half;
+        loop {
+            match control_protocol::read_frame(&mut read_half).await {
+                Ok(Some(ControlBody::PromptCompleted {
+                    stop_reason,
+                    is_error,
+                    ..
+                })) => {
+                    // Claim the one-shot terminal guard. Winning means the
+                    // watchdogs stand down; losing means one already fired,
+                    // so do not double-emit Stopped.
+                    if terminal_guard
+                        .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+                        .is_ok()
+                    {
+                        let reason = control_stop_reason(stop_reason.as_deref(), is_error);
+                        let _ = event_tx.send(Event::Stopped { reason }).await;
+                    }
+                    // The adopted turn is resolved; the control channel's
+                    // Phase A duty for this attach is done. Subsequent turns
+                    // complete through the crate's own prompt future.
+                    break;
+                }
+                Ok(Some(_)) => {}
+                Ok(None) => break,
+                Err(e) => {
+                    debug!(
+                        target: "acp.protocol",
+                        session = %session_label,
+                        "runner control read ended: {e}"
+                    );
+                    break;
+                }
+            }
+        }
+    });
+}
+
+/// Map a runner-reported prompt outcome to an `Event::Stopped` reason. A
+/// completed turn renders as Idle regardless of stop reason, so the
+/// default is `prompt_complete`; the one reason with special downstream
+/// handling (`rate_limited`) is preserved when the agent reports it.
+fn control_stop_reason(stop_reason: Option<&str>, _is_error: bool) -> String {
+    match stop_reason {
+        Some("rate_limited") | Some("rate_limit") => "rate_limited".to_string(),
+        _ => "prompt_complete".to_string(),
     }
 }
 
@@ -4917,6 +5069,13 @@ async fn run_connection_task<W, R>(
     default_effort: Option<String>,
     default_mode: Option<String>,
     mcp_servers: Vec<McpServer>,
+    // Shared terminal-Stopped guard, supplied when a runner control
+    // channel (#1054 Phase A) may deliver the adopted turn's completion
+    // natively. The control reader CAS-claims it before emitting
+    // `Stopped`, so the resume-idle and between-prompt watchdogs below
+    // see it already fired and stand down. `None` on paths with no
+    // control channel (direct stdio), where the task owns its own guard.
+    external_terminal_guard: Option<Arc<std::sync::atomic::AtomicBool>>,
 ) where
     W: futures_util::AsyncWrite + Send + 'static,
     R: futures_util::AsyncRead + Send + 'static,
@@ -5012,7 +5171,11 @@ async fn run_connection_task<W, R>(
     let last_event_at = Arc::new(AtomicI64::new(now_ms));
     let first_event_after_attach = Arc::new(AtomicBool::new(false));
     let prompt_sent_since_attach = Arc::new(AtomicBool::new(false));
-    let watchdog_fired = Arc::new(AtomicBool::new(false));
+    // Shared with the runner control reader (#1054 Phase A) when present,
+    // so a native `prompt_complete` from the runner and the resume-idle /
+    // between-prompt watchdogs all claim the same one-shot terminal guard.
+    let watchdog_fired =
+        external_terminal_guard.unwrap_or_else(|| Arc::new(AtomicBool::new(false)));
     // True for a turn adopted mid-flight via `Resume { in_flight_turn: true }`:
     // a prior connection issued the `session/prompt`, so this connection has no
     // owning `prompt_fut` and no real `ClientCmd::Prompt` will emit the turn's
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3221,48 +3221,46 @@ async fn attach_runner_control(
     use std::sync::atomic::Ordering;
 
     let control_path = crate::process::worker::control_socket_sibling(main_socket);
-    // The runner binds the control socket before the main relay socket,
-    // which the caller already waited for, so this connect is effectively
-    // immediate; the bound only guards a wedged runner.
-    let bound = std::time::Duration::from_secs(5);
 
-    let stream =
-        match tokio::time::timeout(bound, tokio::net::UnixStream::connect(&control_path)).await {
-            Ok(Ok(s)) => s,
-            _ => {
-                debug!(
-                    target: "acp.protocol",
-                    session = %session_label,
-                    "no runner control socket; using resume-idle watchdog"
-                );
-                return;
-            }
-        };
-
-    let (mut read_half, mut write_half) = stream.into_split();
-
-    // The runner greets with Hello on accept; require a matching version
-    // before trusting the channel.
-    let handshake_ok = matches!(
-        tokio::time::timeout(bound, control_protocol::read_frame(&mut read_half)).await,
-        Ok(Ok(Some(ControlBody::Hello { control_protocol_version, .. })))
-            if control_protocol_version == control_protocol::CONTROL_PROTOCOL_VERSION
-    );
-    if !handshake_ok {
-        debug!(
-            target: "acp.protocol",
-            session = %session_label,
-            "runner control handshake failed; using resume-idle watchdog"
-        );
-        return;
-    }
-    let _ = control_protocol::write_frame(
-        &mut write_half,
-        &ControlBody::Attach {
-            control_protocol_version: control_protocol::CONTROL_PROTOCOL_VERSION,
-        },
-    )
-    .await;
+    // A single deadline covers connect plus the Hello read. The runner
+    // binds the control socket before the main relay socket the caller
+    // already waited for, so both steps are effectively immediate; the
+    // bound only caps a wedged runner that bound the socket but never
+    // greets, so it cannot stall the whole resume path. An old socketless
+    // runner fails connect at once and falls back to the watchdog.
+    let bound = std::time::Duration::from_secs(2);
+    let dial = async {
+        let stream = tokio::net::UnixStream::connect(&control_path).await.ok()?;
+        let (mut read_half, mut write_half) = stream.into_split();
+        // The runner greets with Hello on accept; require a matching
+        // version before trusting the channel.
+        match control_protocol::read_frame(&mut read_half).await {
+            Ok(Some(ControlBody::Hello {
+                control_protocol_version,
+                ..
+            })) if control_protocol_version == control_protocol::CONTROL_PROTOCOL_VERSION => {}
+            _ => return None,
+        }
+        let _ = control_protocol::write_frame(
+            &mut write_half,
+            &ControlBody::Attach {
+                control_protocol_version: control_protocol::CONTROL_PROTOCOL_VERSION,
+            },
+        )
+        .await;
+        Some((read_half, write_half))
+    };
+    let (mut read_half, write_half) = match tokio::time::timeout(bound, dial).await {
+        Ok(Some(halves)) => halves,
+        _ => {
+            debug!(
+                target: "acp.protocol",
+                session = %session_label,
+                "no usable runner control socket; using resume-idle watchdog"
+            );
+            return;
+        }
+    };
 
     info!(
         target: "acp.protocol",
@@ -3271,16 +3269,16 @@ async fn attach_runner_control(
     );
 
     tokio::spawn(async move {
-        // Hold the write half so the runner does not observe the control
-        // socket close; Phase A sends nothing further on it.
-        let _keep_alive = write_half;
+        // Hold the write half open so the runner does not see EOF before
+        // it delivers the adopted turn's completion. Dropped when this
+        // task returns (below), which closes the socket so the runner
+        // clears its own control outbound. The runner also tears its side
+        // down one-shot after a single delivered completion, so it never
+        // writes to a stale socket regardless.
+        let _write_half = write_half;
         loop {
             match control_protocol::read_frame(&mut read_half).await {
-                Ok(Some(ControlBody::PromptCompleted {
-                    stop_reason,
-                    is_error,
-                    ..
-                })) => {
+                Ok(Some(ControlBody::PromptCompleted { stop_reason, .. })) => {
                     // Claim the one-shot terminal guard. Winning means the
                     // watchdogs stand down; losing means one already fired,
                     // so do not double-emit Stopped.
@@ -3288,23 +3286,24 @@ async fn attach_runner_control(
                         .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
                         .is_ok()
                     {
-                        let reason = control_stop_reason(stop_reason.as_deref(), is_error);
+                        let reason = control_stop_reason(stop_reason.as_deref());
                         let _ = event_tx.send(Event::Stopped { reason }).await;
                     }
                     // The adopted turn is resolved; the control channel's
-                    // Phase A duty for this attach is done. Subsequent turns
-                    // complete through the crate's own prompt future.
-                    break;
+                    // Phase A duty for this attach is done. Return so the
+                    // write half drops and the socket closes. Subsequent
+                    // turns complete through the crate's own prompt future.
+                    return;
                 }
                 Ok(Some(_)) => {}
-                Ok(None) => break,
+                Ok(None) => return,
                 Err(e) => {
                     debug!(
                         target: "acp.protocol",
                         session = %session_label,
                         "runner control read ended: {e}"
                     );
-                    break;
+                    return;
                 }
             }
         }
@@ -3315,7 +3314,7 @@ async fn attach_runner_control(
 /// completed turn renders as Idle regardless of stop reason, so the
 /// default is `prompt_complete`; the one reason with special downstream
 /// handling (`rate_limited`) is preserved when the agent reports it.
-fn control_stop_reason(stop_reason: Option<&str>, _is_error: bool) -> String {
+fn control_stop_reason(stop_reason: Option<&str>) -> String {
     match stop_reason {
         Some("rate_limited") | Some("rate_limit") => "rate_limited".to_string(),
         _ => "prompt_complete".to_string(),
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3217,7 +3217,7 @@ async fn attach_runner_control(
     session_label: String,
     terminal_guard: Arc<std::sync::atomic::AtomicBool>,
 ) {
-    use super::control_protocol::{self, ControlBody};
+    use crate::acp::control_protocol::{self, ControlBody};
     use std::sync::atomic::Ordering;
 
     let control_path = crate::process::worker::control_socket_sibling(main_socket);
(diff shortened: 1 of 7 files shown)


Seluj78:
I'd like your reviews @njbrake @jerome-benoit :)

src/acp/runner.rs
```diff
@@ -813,6 +914,73 @@ impl RunnerShared {
...
+                    return;
+                }
+                *guard = None;
+            }
+        }
+        let mut pending = self.pending_control.lock().await;
+        while pending.len() >= CONTROL_PENDING_CAP {
+            pending.pop_front();
+        }
+        pending.push_back(body);
```
jerome-benoit: [blocking] `emit_control` buffers a `PromptCompleted` into `pending_control` even when the turn was just delivered live to an attached main-relay daemon. Because the control socket is dialed only on `Resume { in_flight_turn: true }`, `control_outbound` is `None` during all normal operation, so every completed turn buffers here, and nothing clears `pending_control` (the cap of 64 bounds length, not staleness). On a later mid-flight resume of a session that has completed >= 1 turn, `install_control_outbound` drains the whole stale queue and the daemon reader fires `Stopped` on the FIRST frame without matching `prompt_req_id` (see `attach_runner_control` in `acp_client.rs`), so the still-in-flight adopted turn is marked Idle and its real completion is masked: the reader has already broken, and since the guard is shared with the watchdog it also stands the 30s resume-idle fallback down. That is worse than the watchdog this replaces, on the exact path the feature targets, and the integration test cannot catch it (empty buffer, first turn only).

Fix (runner-side; the daemon cannot match by id because the adopted prompt lives in the prior client's id space): only buffer a completion that occurred during a genuine no-daemon gap. Gate this push on `active_outbound.is_none()` (a live main-relay daemon already owns that turn via its own prompt future), and clear `pending_control` on main-relay (re)attach so any residual is bounded to the current gap. ACP being serial per session, at most one completion is ever legitimately pending. A regression test worth adding: buffer completions for a couple of ids with no control attached, then attach control after a fresh in-flight prompt and assert the stale ones are not replayed as the adopted turn's `Stopped`.

src/acp/runner.rs
```diff
@@ -813,6 +914,73 @@ impl RunnerShared {
...
+                if control_protocol::write_frame(&mut out, &body)
+                    .await
+                    .is_err()
+                {
+                    pending.push_front(body);
+                    return;
+                }
+            }
+        }
+        *self.control_outbound.lock().await = Some(out);
```
jerome-benoit: [should fix] TOCTOU with `emit_control`. This drains `pending_control` (releasing its lock) and then separately sets `control_outbound = Some` at line 978, while `emit_control` (above) checks `control_outbound` (`None`), releases, then locks `pending_control` and pushes. Interleaved -- emit sees `None`, install drains (empty) and sets `Some`, then emit pushes -- a live completion lands in `pending_control` after the drain, so it is never delivered on this attach and falls back to the 30s watchdog. The integration test's `sleep(150ms)` masks exactly this window. Fix: hold `control_outbound` across the `pending_control` push in `emit_control`, and here acquire `control_outbound` first and drain+set under it, so check-and-buffer and drain-and-set are mutually exclusive (consistent `control_outbound` -> `pending_control` order, deadlock-free).

jerome-benoit (changes requested):
Reviewed on `ce1c0c1fc`. The design is well thought through and much is clean: the length-framed control codec (cap-before-alloc, clean-EOF handling), the backward-compatible fallback (an old runner that cannot bind the socket leaves the guard unclaimed and the resume-idle watchdog behaves exactly as before), the `Resume { in_flight_turn: true }`-only dial scoping, `control_socket_sibling`, the 0o600 perms, and socket cleanup in `worker_registry::delete` all check out. Locally I confirmed `cargo fmt`, `cargo clippy --features serve --all-targets` (clean; the two warnings are pre-existing and outside this diff), the control-protocol / runner unit tests, the `acp_runner_control` integration test, and `RUSTDOCFLAGS="-D warnings" cargo doc --features serve` all pass.

Two things I think should be resolved first, both runner-side and local to `RunnerShared` (details inline):

- **Blocking:** a stale buffered `PromptCompleted` is replayed on resume and prematurely stops the still-in-flight turn (inline on `emit_control`).
- **Should fix before merge:** a TOCTOU between `emit_control` and `install_control_outbound` can strand a live completion (inline on `install_control_outbound`).

## Minor

- **Reader `break` leaves the write half open.** After the adopted turn the daemon reader breaks but holds `_keep_alive = write_half`, so `control_outbound` stays `Some` and every later turn writes a frame nobody reads; over a long resumed session those can fill the socket buffer and eventually block `write_frame` on the relay path. Drain after the guard is claimed, or do not hold the write half.
- **`parse_response` runs on every agent->daemon line** in `deliver_line`, on top of `note_daemon_response`, so a busy session double-parses each line even when no prompt is tracked. Gate it on `!prompt_requests.is_empty()`.
- **`is_error` is plumbed end-to-end then discarded** in `control_stop_reason(_is_error)`. Acceptable for Phase A (matches the watchdog's Idle-on-stop), but either use it or note it is reserved for a later phase, per the no-dead-code rule.
- **`attach_runner_control` blocks the resume path** up to ~10s (5s connect + 5s handshake) on a wedged runner that bound the socket but never greets; old socketless runners fail fast, so this is only the pathological case. Consider a single deadline or spawning the dial.

## Note (not this PR)

This edits `src/acp/runner.rs` and `src/acp/worker_registry.rs` and adds `control_socket_sibling` to `src/process/worker.rs`, while #2970 moves `runner.rs`/`worker_registry.rs` into `src/process/`. Whichever lands second will conflict on these files; worth sequencing intentionally.

Thanks for the thorough design writeup and the real-socket integration test; the bones are good and the blocking item is a contained runner-side change.

jerome-benoit:
**Architecture note (extends my review above; non-blocking).**

Zooming out from the line-level findings to the shape of the approach, checked against how comparable systems recover a long-running turn (including its terminal event) across a client reconnect.

**Motivation is sound.** ACP v1 has no native way to adopt an in-flight `session/prompt` across a reconnect (`session/resume` / `session/load` restore session context or completed history, not an in-flight turn; request-id continuity across reconnect is unspecified). So doing something runner-side is warranted, not a workaround for a missing call.

**What already exists here.** AoE already has the idiomatic primitive for this problem class: a durable, per-session, `seq`-keyed event log (`EventStore`) with a retention-aware keyset replay (`GET /api/sessions/{id}/acp/replay?since=N`, returning `highest_seq`/`lowest_seq`/`lost`), and `derive_acp_status` maps `Event::Stopped` to `Idle`. On reattach the daemon already rebuilds session state by replaying that log from its cursor. That is the same shape as MCP Streamable HTTP's `Last-Event-ID` + server-side EventStore, Temporal's workflow history, and runbaypty's `ATTACH {since_seq}`.

**Where the real gap is: production, not delivery.** When the turn finishes during the daemon-down window, the entity that terminates the protocol (the daemon's in-process ACP client) is gone and has lost the owning `prompt_fut`, so no `Event::Stopped` is ever produced into the log. The runner is the only survivor that observes the agent's response, but it is a raw byte relay with no `EventStore` handle (the store is daemon-owned). Phase A bridges that by buffering the completion in memory (`pending_control`) and replaying it on reattach. The transport choice is reasonable and well-precedented (the supervisor/sidecar reattach pattern: Claude Code's supervisor + `roster.json`, jarvis-sidecar, tmux). But that cross-restart buffer lacks the log's ordering/dedup/cursor properties, and the two review findings fall straight out of that: the blocking one (a drained completion fires `Stopped` without matching `prompt_req_id`) is a missing cursor/id match; the drain-then-set race is a missing atomic handoff. For fairness: once delivered, the resulting `Event::Stopped` is recorded durably on the daemon side; only the cross-restart transport is non-durable.

**Direction.** The genuinely Phase-A-sized fix is to give that transport the log-like property it is missing: make the control channel idempotent and keyed on the `prompt_req_id` the `PromptCompleted` frame already carries, so the daemon matches/dedupes against the adopted turn's id instead of firing on the first drained frame. That closes both findings with no new infrastructure. The fully idiomatic end-state is your own Phase B/C: the runner terminates the protocol and becomes the durable producer feeding the same `seq`-keyed replay path the daemon already consumes, so completion needs no parallel channel at all. Routing completion through the *existing* durable log in Phase A is not reachable (the daemon owns the log and is down at completion time); that would require a runner-written durable marker, which is effectively Phase B.

Prior art surveyed: ACP `session/resume`/`session/load`; MCP Streamable HTTP `Last-Event-ID` + EventStore (2025-11-25 revision); Temporal event-history durable execution; runbaypty `ATTACH {since_seq}`; Claude Code supervisor + `roster.json`; jarvis-sidecar. LSP/DAP `$/progress` and gRPC streaming confirm requests are otherwise connection-scoped with no native resume.

src/acp/runner.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/758fee69. `emit_control` now only buffers during a genuine no-daemon gap, gated on a `main_attached` mirror of `active_outbound`: while a main-relay daemon is attached the completion is owned by its own prompt future, so nothing is buffered. `pending` is cleared at gap start (main-relay detach) and is now a single `Option` (ACP is serial per session, so at most one completion is ever legitimately pending), so a stale completion can no longer be replayed onto a future adopted turn. Added a unit test (`completion_not_buffered_while_main_daemon_attached`).

src/acp/runner.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/758fee69. `control_outbound` and `pending` now live under one `ControlChannel` mutex, so `emit_control` (check outbound, else buffer) and `install_control_outbound` (take pending, else set outbound) run under a single lock and cannot interleave. A completion emitted in what used to be the drain-then-set window is now serialized with the install. The integration test's sleep is no longer masking a race.

Seluj78:
@jerome-benoit thanks for the thorough review. Both blocking/should-fix items and the minors are addressed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/758fee69:

- **Reader leaves write half open:** the runner now tears its control outbound down one-shot after delivering a completion, and the daemon reader returns (dropping the write half, closing the socket) after the adopted turn, so no frame is written to a socket nobody reads even on a long resumed session.
- **`parse_response` on every line:** gated on an in-flight tracked prompt (`!prompt_requests.is_empty()`), so a busy session no longer double-parses.
- **`is_error` plumbed then discarded:** removed end-to-end. An error-envelope response is still detected as a completion (no `stopReason`), matching the Idle-on-stop behavior.
- **`attach_runner_control` blocking:** collapsed to a single 2s deadline covering connect plus handshake, so a wedged runner that binds but never greets cannot stall the resume path.

Re the #2970 sequencing note (it moves `runner.rs`/`worker_registry.rs` into `src/process/`): agreed, whichever lands second conflicts. Flagging for intentional sequencing; happy to rebase this onto #2970 if it lands first.

Seluj78:
Rebased onto `main` to pick up #2970 (which moved `runner.rs` and `worker_registry.rs` into `src/process/`). Resolved the overlap: my runner changes now live in `src/process/runner.rs`, `control_protocol` stays in `src/acp/` (imported via `crate::acp`), and the union with the new `read_frame_bounded` is preserved. The review-fix commit referenced above is now `87a3bf3c` after the rebase. Branch is conflict-free; unit + integration tests green locally.

src/process/runner.rs
```diff
@@ -832,12 +948,97 @@ impl RunnerShared {
...
+                .is_err()
+            {
+                ch.pending = Some(body);
+            }
+            // Delivered (or failed on a dead socket); one completion per
+            // attach, so do not retain the outbound.
+            return;
+        }
+        ch.outbound = Some(out);
+    }
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/98a9518c. All three control-channel writes (live frame in `emit_control`, `Hello`, and the buffered completion in `install_control_outbound`) now go through `write_control_frame`, which caps the write at `CONTROL_WRITE_TIMEOUT` (2s) via `tokio::time::timeout` and reports a timeout as a write failure, so a stalled control peer can no longer wedge the `control` mutex or the stdout-relay task; the existing drop/buffer cleanup runs on timeout. Good catch.

Seluj78:
@jerome-benoit thanks for the architecture note, that framing (runner as the surviving observer, EventStore as the idiomatic seq-keyed replay path, Phase B/C making the runner a durable producer feeding it) matches the plan exactly. On the Phase-A suggestion to key completion on `prompt_req_id`: the adopted turn's id lives in the prior daemon client's id space, so the reattaching daemon has no id to match against; the frame carries it for observability but the daemon can't dedupe on it. I closed both findings on the runner side instead: the completion is only buffered during a genuine no-daemon gap and `pending` is a single `Option` cleared at gap start, so at most the one correct adopted-turn completion is ever replayed (no firing on a stale first frame), and the drain/set race is gone now that `outbound` and `pending` share one lock. The durable-producer end-state is Phase B/C as you describe.

Seluj78:
can I get a re-review @jerome-benoit ? :)

src/process/runner.rs
```diff
@@ -832,12 +974,92 @@ impl RunnerShared {
...
+            // Dead or stalled socket: fall through to maybe-buffer for a
+            // later attach.
+        }
+        if self
+            .main_attached
+            .load(std::sync::atomic::Ordering::Relaxed)
+        {
+            return;
+        }
+        ch.pending = Some(body);
```
jerome-benoit, line 1041: [should fix / MAJOR] A timed-out live control write silently drops the native completion. When `write_control_frame` times out mid-frame (line ~1016 returns `false`), `outbound` is cleared and we fall through; on the resume/adopted-turn path the main relay is attached (`main_attached == true`), so this returns at ~1028 without buffering. The completion is then neither written nor buffered, and the turn resolves only via the 30s resume-idle watchdog this feature exists to replace, with just a `debug!` breadcrumb. Since the 2s timeout implies a wedged local reader, the fast path is lost exactly when it matters. Consider buffering on a *timed-out* (vs hard-error) write: it is idempotent against the guard CAS, so a later control reattach delivers it with no double-emit; or at least log at `warn`.

Separately (MINOR): `main_attached` is stored under the `active_outbound` lock but read here under the `control` lock, so this gate is eventually-consistent, not atomic w.r.t. the completion decision. The failure mode is benign (a dropped stale completion, or one extra watchdog fallback), but the comment reads stronger than the guarantee, worth documenting it as a heuristic.

src/acp/acp_client.rs
```diff
@@ -3170,6 +3203,124 @@ async fn wait_for_socket(
...
 
+/// Dial the runner's sibling control socket (#1054 Phase A) and, if it
+/// speaks the control protocol, spawn a reader that turns the runner's
+/// native `PromptCompleted` into `Event::Stopped { reason }`. The reader
+/// CAS-claims `terminal_guard` so the caller's resume-idle and
+/// between-prompt watchdogs stand down for the adopted turn. Best-effort:
+/// a connect failure or an unrecognized/absent control socket (a runner
+/// too old to bind it) leaves the guard unclaimed so the legacy watchdog
+/// still fires.
+async fn attach_runner_control(
```
jerome-benoit, line 3214: [should fix / test gap] This ~120-line daemon-side consumer is the other half of the feature and is entirely untested: the CAS guard-claim, the `Hello` version-mismatch rejection (~L3237), the one-shot reader teardown, and `control_stop_reason` all ship without coverage. The integration test drives the runner side and stops at the raw `PromptCompleted` frame; nothing exercises the reader that turns it into `Event::Stopped`. More importantly, the backward-compat fallback (a socketless/old runner leaves `terminal_guard` unclaimed so the legacy watchdog still fires) is the load-bearing safety claim for dropping the idle heuristic, and it has no test. Please add: (a) a test that an absent/failed control socket leaves the guard unclaimed, and (b) a unit test for `control_stop_reason` plus a wrong-version `Hello` aborting the attach.

src/acp/acp_client.rs
```diff
@@ -3170,6 +3203,124 @@ async fn wait_for_socket(
...
+            }
+        }
+    });
+}
+
+/// Map a runner-reported prompt outcome to an `Event::Stopped` reason. A
+/// completed turn renders as Idle regardless of stop reason, so the
+/// default is `prompt_complete`; the one reason with special downstream
+/// handling (`rate_limited`) is preserved when the agent reports it.
+fn control_stop_reason(stop_reason: Option<&str>) -> String {
```
jerome-benoit, line 3324: [minor] This collapses every non-`rate_limited` stop reason (`cancelled`, `max_tokens`, `refusal`, `max_turn_requests`) into `prompt_complete`, discarding the ACP `stopReason`. Fine for Phase A (a completed turn renders Idle regardless), but `cancelled`/`max_tokens` losing their identity may matter to downstream UI in a later phase, worth a tracked follow-up note.

src/process/worker.rs
```diff
@@ -283,6 +283,17 @@ pub fn log_path(dir: &Path, id: &str) -> Result<PathBuf> {
...
 
+/// `<dir>/<id>.control.sock`, the typed control channel that rides
+/// alongside the raw ACP relay `<id>.sock`. Deriving it from the main
+/// socket keeps the runner (which binds it) and the daemon (which dials
+/// it) in agreement without either needing the workers dir: `x.sock`
+/// becomes `x.control.sock`. Session ids are validated (alphanumeric,
+/// `-`, `_`) so they never carry a `.` that would confuse the extension
+/// swap. Phase A of #1054.
+pub fn control_socket_sibling(main_socket: &Path) -> PathBuf {
+    main_socket.with_extension("control.sock")
```
jerome-benoit, line 302: [minor] `with_extension("control.sock")` maps `x.sock` -> `x.control.sock` correctly, but feeding it an already-derived control path yields `x.control.control.sock` silently. Unreachable today (all callers pass the main `.sock`), and the id-validation comment guards a different case (ids containing a `.`), not self-application. A one-line `debug_assert!(!main_socket.to_string_lossy().ends_with(".control.sock"))` would make future misuse loud.

tests/acp_runner_control.rs
```diff
@@ -0,0 +1,158 @@
...
+    let mut ctl = UnixStream::connect(&control).expect("connect control socket");
+    ctl.set_read_timeout(Some(Duration::from_secs(10))).unwrap();
+    let hello = read_frame(&mut ctl);
+    assert_eq!(hello["kind"], "hello", "first control frame is Hello");
+    assert_eq!(hello["session_id"], session_id);
+
+    // Let the runner finish installing the control outbound before driving
+    // the prompt, so the completion event is written live rather than
+    // buffered for a future attach.
+    std::thread::sleep(Duration::from_millis(150));
```
jerome-benoit, line 138: [nit] With the emit/install TOCTOU now closed by the single `ControlChannel` mutex, this 150ms sleep to force 'install before drive' is largely vestigial and adds a small flake surface. Also the test depends on `cat` resolving on the runner's inherited PATH (it fails as a confusing 'registry record never appeared' when `/bin` is not first, e.g. under a nix PATH); an absolute `/bin/cat` would make it robust and the failure mode legible.

jerome-benoit (commented):
## Re-review (was CHANGES_REQUESTED on `ce1c0c1f`, now at `98a9518c`)

Both of my prior blocking findings are resolved, verified against the code at HEAD:

- **[resolved] Stale `PromptCompleted` replay (was BLOCKER).** `emit_control` now gates buffering on `main_attached`, and `clear_outbound` drops any un-drained `pending` at gap start, so only the current no-daemon gap's completion is ever replayable. Regression-locked by `completion_not_buffered_while_main_daemon_attached`.
- **[resolved] emit/install TOCTOU (was MAJOR).** `outbound` + `pending` now live under one `ControlChannel` mutex, so the drain-then-set and the check-then-buffer are mutually exclusive. Cleaner than a lock-ordering rule.
- **[resolved] CodeRabbit write-timeout.** `write_control_frame`'s 2s cap bounds the relay-freezing write.

Praise: the length-framed codec (cap-before-alloc, clean-EOF, truncation-as-error), the one-shot CAS `terminal_guard` handshake, and a genuinely backward-compatible fallback are all well done. Solid, well-scoped Phase A.

I am not re-blocking on the old items: no correctness defect remains in the paths I read, CI is green, and the integration test exercises the real socket round-trip. The remaining items (inline) are should-fix, not hard blockers:

- **MAJOR (robustness):** a timed-out live control write silently drops the native completion on the resume path and falls back to the 30s watchdog (see inline on `emit_control`). Buffer-on-timeout (guard-idempotent) or at least `warn`.
- **MAJOR (tests):** the backward-compat fallback and the entire daemon-side consumer (`attach_runner_control` / `control_stop_reason` / version rejection) are untested; the write-timeout path is untested.
- MINOR/NIT: `main_attached` cross-lock heuristic, lossy `stop_reason` collapse, `control_socket_sibling` double-suffix trap, and a now-vestigial test sleep + PATH fragility.

Happy to clear to APPROVE once the two MAJOR items are addressed or explicitly waived/tracked. Nice work.

src/process/runner.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/9270140f. Good catch, the timeout guard I added could drop the completion it protects. `emit_control` now buffers unconditionally when a live write fails (a control daemon had dialed in, so the completion is real and unreceived and must survive for the next control attach); the `main_attached` gate applies only when control was never attached. So a stalled reader no longer costs the fast path, it defers it to the next attach instead of dropping it.

src/acp/acp_client.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/9270140f. Added daemon-side unit tests for `attach_runner_control`: the happy path (Hello + PromptCompleted drives `Stopped{prompt_complete}` and claims the terminal guard), `Hello` version-mismatch (guard left unclaimed), and the load-bearing socketless fallback (no control listener leaves the guard unclaimed so the watchdog stays the terminal authority). CAS claim, one-shot teardown, and `control_stop_reason` are exercised through the happy path.

src/acp/acp_client.rs
Seluj78: Documented in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/9270140f: `control_stop_reason` now notes the non-rate_limited stop reasons (`cancelled`, `max_tokens`, `refusal`, `max_turn_requests`) collapse to `prompt_complete` in Phase A, with richer forwarding tracked on the Phase C runner-terminator issue #2977 where the runner owns the protocol and can carry the typed stop reason natively.

src/process/worker.rs
Seluj78: Added in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/9270140f: a `debug_assert!(!...ends_with(".control.sock"))` in `control_socket_sibling` so self-application (`x.control.control.sock`) is loud in debug builds.

tests/acp_runner_control.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/9270140f: the test now uses `/bin/cat` so a non-standard PATH fails legibly instead of as a confusing 'registry record never appeared'. Kept the short wait but re-commented it: it is an install-ordering wait (let the control outbound store land so the completion goes live, not the closed emit/install TOCTOU), not a race workaround.

Seluj78:
@jerome-benoit done :)

src/process/runner.rs
```diff
@@ -832,12 +974,103 @@ impl RunnerShared {
...
+                return;
+            }
+            // Dead or stalled control socket. A control daemon had dialed
+            // in (this is the resume/adopted-turn path), so this completion
+            // is real and was not received: buffer it unconditionally for
+            // the next control attach to replay. Do NOT fall through to the
+            // main_attached gate, which would drop it (the write timing out
+            // is exactly when the fast path matters most). See PR #2975.
+            ch.pending = Some(body);
+            return;
```
jerome-benoit, line 1028: [should fix, non-blocking] This is the fix this commit is named after (buffer unconditionally on a failed/timed-out live write rather than dropping via the `main_attached` gate), and it is correct, but it ships without a regression test. Please add an `emit_control` unit test: construct a `RunnerShared` whose control `outbound` write fails, with `main_attached == true`, and assert `ch.pending == Some(body)` afterwards (i.e. it did NOT fall through the gate). That also naturally covers the `CONTROL_WRITE_TIMEOUT` -> `write_control_frame` returns false path, which is currently untested.

For the record: I stress-tested whether this unconditional buffer reintroduces the stale-replay class I originally blocked on, and it does not. `emit_control` holds the `control` mutex across both the bounded write and this `pending = Some(...)`, and `clear_outbound` needs the same mutex to clear `pending`, so a gap-start clear cannot interleave between the parked write and the buffer set. A one-line comment on `clear_outbound` noting that ordering guarantee would help the next reader.

jerome-benoit (approved):
## Re-review round 2 (now at `9270140f`) — clearing to APPROVE

Every item from my prior re-review is addressed, verified against the code at HEAD:

- **[resolved] MAJOR-1 (timed-out write dropped the completion).** `emit_control` now buffers unconditionally on a failed/timed-out live write (`runner.rs:1015-1028`), instead of falling through to the `main_attached` gate. Correct.
- **[resolved] Daemon-side consumer was untested.** Three genuine, deterministic unit tests over the real `attach_runner_control` (bind a real control socket at the derived sibling path): happy path (`Hello` + `PromptCompleted` -> `Stopped{prompt_complete}` + guard CAS true), `Hello` version-mismatch (guard unclaimed), and the load-bearing socketless fallback (guard unclaimed -> watchdog stays authoritative). Verified passing.
- **[resolved] minors/nit:** `control_stop_reason` collapse documented as a Phase A choice tracked for #2977; `control_socket_sibling` gains a `debug_assert!` against self-application; the integration test uses `/bin/cat` and documents its install-ordering wait.

I also specifically stress-tested whether the new *unconditional* buffer-on-write-failure could reintroduce the stale-`PromptCompleted`-replay class I originally blocked on. **It does not.** `emit_control` holds the single `control` mutex across both the bounded write and the `pending = Some(...)` set (`runner.rs:1014` -> `1027`), and `clear_outbound` needs that same mutex to clear `pending` (`runner.rs:989`); so the "gap-start clear runs during the parked write, then the completion is buffered after it" ordering is impossible. On a daemon crash the completion is dropped, not stranded, and only the current gap's completion is ever replayable. The single-mutex-over-both-fields invariant this PR introduced is exactly what closes it.

Approving. One non-blocking fast-follow request (inline) plus minor nits, none blocking merge:

- **Fast-follow (should fix):** the MAJOR-1 fix itself ships without a regression test. A small `emit_control` unit test (outbound present + a failing write -> assert `pending == Some(body)`, and not gated by `main_attached`) would lock the exact behavior this commit is named after, and would also cover the `CONTROL_WRITE_TIMEOUT` failure return.
- **Nit:** the daemon reader discards `prompt_req_id` (`PromptCompleted { stop_reason, .. }`). Validating it against the adopted turn would harden the consumer against any future protocol desync and is the one field that would make a stale drain self-defending; worth considering even in Phase A.
- **Nit:** `control_stop_reason` and the daemon reader's non-completion arms (`Ok(Some(_))` / `Err`) have no direct tests; both are trivial to pin.

Nice work iterating through this. Solid, well-scoped Phase A.

src/process/runner.rs
Seluj78: Added in https://github.com/agent-of-empires/agent-of-empires/pull/2975/commits/1bf8e4c4: `emit_control_buffers_on_write_failure_even_when_main_attached` builds a control outbound whose peer is dropped (AF_UNIX write-after-peer-close fails deterministically), sets `main_attached = true`, and asserts the outbound is cleared and `pending == Some(body)` (buffered, not dropped through the gate).

Seluj78:
@jerome-benoit another one :D

src/process/runner.rs
jerome-benoit: Resolved in 1bf8e4c4. `emit_control_buffers_on_write_failure_even_when_main_attached` drops the AF_UNIX peer, sets `main_attached = true`, drives `emit_control`, and asserts `pending == Some(body)`. I verified it is a genuine regression (not a tautology): with the fix reverted, the failed write falls through to the main_attached gate and returns leaving `pending == None`, so the assert fails. Also verified the write failure is deterministic via immediate EPIPE (full peer drop, not the 2s timeout), so no flake and no ~2s cost. Thanks for adding it.

jerome-benoit (approved):
## Re-review round 3 (now at `1bf8e4c4`) — re-approving

`1bf8e4c4` adds exactly the regression test I asked for in my previous approval (the buffer-on-write-failure fast-follow), and it holds up under scrutiny:

- **Genuine regression, not a tautology.** `emit_control_buffers_on_write_failure_even_when_main_attached` drives the real `emit_control` with a control `outbound` whose AF_UNIX peer is dropped and `main_attached = true`, and asserts `pending == Some(body)`. Traced old-vs-new: with the fix reverted, the failed write falls through to the `main_attached` gate and returns, leaving `pending == None`, so the assert fails. It genuinely pins `src/process/runner.rs:1027`.
- **Deterministic + fast.** The write fails via an immediate EPIPE (the peer is fully closed, not a read-shutdown), not by waiting out the 2s `CONTROL_WRITE_TIMEOUT` — verified empirically (repeated runs finish in ~0.00s, never ~2s). No sleeps, no ordering races; sound on both Linux and macOS.
- **Isolated.** `+34/-0`, `src/process/runner.rs` test module only; no production-code change, conventions clean.

Two non-blocking nits, neither worth holding the PR:
- The doc comment on the test could say "the peer is fully closed" to make explicit that the immediate-EPIPE guarantee would not hold for a mere read-shutdown.
- The `CONTROL_WRITE_TIMEOUT` arm of `write_control_frame` itself is still not directly exercised (only the EPIPE path is); both funnel to the same buffer branch, and a timeout test would be non-deterministic, so leaving it is reasonable.

All my prior items across the three rounds are resolved. Approving. Nice work.

PR #2995, opened 2026-07-20
fix: keep local TUI on loopback for remote daemons

amanzainal:
## Description

`aoe serve --remote` requires token auth plus a passphrase session for tunneled clients. The auth middleware intentionally bypasses that browser-session requirement for filesystem-trusted loopback callers so the local TUI can attach.

Remote mode currently writes only the public tunnel URL to `serve.url`. Local TUI discovery therefore falls back to that URL, round-trips through Tailscale or Cloudflare, and arrives with a non-loopback client IP. Its token is valid, but it has no browser session cookie, so replay and WebSocket requests receive `401 login_required` and structured chats cannot open.

This change:

- Persists the public tunnel URL as the primary `serve.url` entry and adds a token-matched `127.0.0.1` alternate for same-host clients.
- Rebuilds both entries when the remote daemon rotates its auth token.
- Keeps backward compatibility with older single-URL files while testing that daemon discovery prefers loopback when available.
- Updates comments that previously described tunnel mode as single-URL only.

This aligns discovery with the existing auth policy: local TUI traffic stays local, while remote PWA traffic continues to require both authentication factors. Public URLs, QR/status consumers, and remote authentication behavior remain unchanged.

Replaces #2964, which was closed by the template automation after its `missing-template` label had already been removed.

### Validation

- `git diff --check upstream/main...HEAD`
- `cargo +1.95.0 test --features serve --lib remote_serve_url_contents`
- `cargo +1.95.0 test --features serve --lib acp::client::discovery::tests`
- `cargo +1.95.0 test --features serve --lib serve_url_parses`
- `cargo +1.95.0 test --features serve --lib post_token_auth_action`
- `cargo +1.95.0 check --features serve`
- `cargo +1.95.0 clippy --features serve --lib -- -D warnings -A clippy::collapsible-match`

The explicit Clippy allowance is for an existing `collapsible_match` warning in `src/acp/event_store.rs`; this change introduces no new Clippy warnings.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: exercising the failure end-to-end requires a live Tailscale or Cloudflare tunnel plus passphrase auth. Unit tests cover remote URL persistence, token rotation output, loopback preference, single-URL backward compatibility, URL parsing, and the loopback auth policy.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** OpenAI Codex (GPT-5)

Codex diagnosed the original live daemon failure, implemented the isolated fix, rebased the replacement branch onto current `upstream/main`, and ran the validation commands listed above.

**NOTE:**

## Summary

- Updated remote daemon discovery to prefer the local loopback address, keeping TUI traffic off the public tunnel.
- Remote mode now records both the public URL and a token-aware loopback alternative, including after token rotation.
- Preserved compatibility with older single-URL files and existing public URL, QR, and authentication behavior.
- Added and updated tests covering URL preference, formatting, token handling, and legacy parsing.

## Benefits

- Improves local TUI connectivity and avoids unnecessary public routing.
- Keeps authentication and displayed endpoint information synchronized.
- Maintains backward compatibility for existing daemon configurations.

### DIFF
--- a/src/acp/client/discovery.rs
+++ b/src/acp/client/discovery.rs
@@ -17,7 +17,7 @@ use std::env;
 
 use thiserror::Error;
 
-use crate::cli::serve::{daemon_pid, read_serve_urls};
+use crate::cli::serve::{daemon_pid, read_serve_urls, ServeUrl};
 
 /// A located daemon endpoint. `base_url` carries no query string so it
 /// is safe to log; the auth token (if any) travels separately and is
@@ -98,11 +98,7 @@ pub fn discover_local() -> Result<DaemonEndpoint, DiscoveryError> {
     if urls.is_empty() {
         return Err(DiscoveryError::NoLocalDaemon);
     }
-    let pick = urls
-        .iter()
-        .find(|u| is_loopback(&u.url))
-        .or_else(|| urls.first())
-        .ok_or(DiscoveryError::Malformed)?;
+    let pick = preferred_daemon_url(&urls).ok_or(DiscoveryError::Malformed)?;
     let token = extract_token(&pick.url).map(str::to_string);
     let base_url = trim_query(&pick.url).trim_end_matches('/').to_string();
     if base_url.is_empty() {
@@ -113,6 +109,12 @@ pub fn discover_local() -> Result<DaemonEndpoint, DiscoveryError> {
         token,
         source: Source::LocalDaemon,
     })
+}
+
+fn preferred_daemon_url(urls: &[ServeUrl]) -> Option<&ServeUrl> {
+    urls.iter()
+        .find(|u| is_loopback(&u.url))
+        .or_else(|| urls.first())
 }
 
 fn is_loopback(url: &str) -> bool {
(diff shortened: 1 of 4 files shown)


njbrake (approved):

Approving. This is a genuine, well-scoped bug fix for a real broken flow: the local TUI could not open structured chats against a remote daemon because discovery attached over the public tunnel URL and arrived without a browser passphrase session, hitting `401 login_required`.

I verified the load-bearing facts against source rather than the description:

- The loopback bypass the fix relies on is real and pre-existing (`src/server/auth.rs`, #1168/#1647): a valid token from a loopback peer skips the passphrase-session requirement.
- `127.0.0.1:{local_port}` is reachable in remote mode: the default bind is loopback and the tunnel forwards to `localhost:{local_port}`, so the alternate always points at a live listener.
- The written `serve.url` format round-trips cleanly through the existing parser, and the primary line is byte-identical to the previous single-URL output, so QR/status consumers are unaffected.
- Both write paths (startup and token rotation) are covered, and the new unit tests pass locally under `--features serve`.

One non-blocking edge case: an IPv6-only bind (`--remote --host ::`) would make the hardcoded `127.0.0.1` alternate unreachable, but that configuration is already broken upstream since the tunnel targets `localhost`, so it is not a regression here.

Backward compatibility is preserved and the blast radius is contained to remote daemon discovery. Good to merge.

PR #3003, opened 2026-07-20
fix: refresh local TUI token after rotation

amanzainal:
## Description

The native structured-view TUI snapshots the daemon bearer token during discovery, but a running remote daemon rotates that token every four hours. A TUI left open across a rotation keeps sending the expired credential for HTTP polling and WebSocket reconnects. Repeated plugin UI-state polls can then exhaust the shared per-IP authentication failure budget, causing otherwise valid clients to receive `429 Too Many Requests`.

This change:

- resolves the current owner-only `serve.token` before each local loopback HTTP request and WebSocket handshake
- shares the last valid token across endpoint clones so a partial file write never falls back to an older discovery-time credential
- restricts disk token refresh to parsed loopback endpoints discovered from the local daemon, preventing token substitution into explicit environment or legacy public URLs
- stops the three-second plugin UI-state poll after an unrecoverable `401`, while continuing to retry transient errors and an existing `429` lockout
- adds regression coverage for token rotation, HTTP headers, WebSocket URLs, partial writes, hostile lookalike hostnames, endpoint isolation, and polling policy

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (not applicable; no rendering changes)

## Test Coverage Analysis

- [x] Skipped a test, because: this changes native TUI daemon authentication, not a browser dashboard flow; the transport behavior is covered directly in Rust.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: production rotation is a four-hour timing boundary; deterministic unit tests cover discovery, rotated credentials on both transports, invalid-write fallback, endpoint isolation, and poll termination.

## Validation

- `git diff --check upstream/main...HEAD`
- focused discovery, HTTP, WebSocket, and plugin-poll tests: 18, 7, 7, and 1 passed respectively
- `cargo +1.95.0 clippy -- -D warnings`
- `cargo +1.95.0 clippy --features serve --all-targets -- -A clippy::collapsible-match -A clippy::items-after-test-module -D warnings`
  - both allowances cover existing Rust 1.95 warnings in files untouched by this PR
- `umask 077; TMPDIR=<owner-only temp dir> CARGO_BUILD_JOBS=1 cargo +1.95.0 test --features serve`
  - library: 5,650 passed, 2 ignored
  - integration main: 159 passed, 5 ignored
  - all remaining test binaries passed; aggregate: 5,903 passed, 7 ignored, 0 failed; doc tests: 0 passed, 1 ignored

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

GPT-5 Codex

Used for root-cause analysis, implementation, regression tests, security-focused patch review, rebase conflict resolution, and validation. The resulting diff and validation output were reviewed before publishing.

**NOTE:**

## Summary

- Fixed local TUI authentication after the daemon rotates its bearer token by refreshing/using the latest token right before HTTP requests and WebSocket handshakes.
- Updated daemon endpoint token storage to be safely shareable across endpoint clones, so all parts of the client use the same up-to-date credentials.
- Limited on-disk token refresh to validated loopback/local-daemon endpoints, and added extra checks to avoid using partially written or malformed token files.
- Prevented the plugin UI polling loop from continuing indefinitely after unrecoverable `401` authentication failures, while keeping retries for transient errors (including existing rate-limit handling).
- Strengthened loopback/hostname detection logic and expanded regression tests for token rotation, isolation between endpoint types, and polling behavior.

## Benefits

Local TUI sessions stay authenticated reliably across daemon token rotations, avoid falling back to stale/unsafe credentials from broken token files, and reduce wasted work by stopping polling when authentication can’t recover.

Validation passed: formatting, Clippy, focused tests, and the full suite (5,902 passed, 7 ignored).

### DIFF
--- a/src/acp/client/discovery.rs
+++ b/src/acp/client/discovery.rs
@@ -14,6 +14,8 @@
 //! variant whose message tells the user how to start one.
 
 use std::env;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, RwLock};
 
 use thiserror::Error;
 
@@ -28,18 +30,72 @@ pub struct DaemonEndpoint {
     /// Bare base URL (`http://127.0.0.1:8080`). No trailing slash, no
     /// query string.
     pub base_url: String,
-    /// Bearer token. `None` when the daemon was started with
-    /// `--no-auth`, or when `AOE_DAEMON_URL` is set without
-    /// `AOE_DAEMON_TOKEN`.
-    pub token: Option<String>,
+    /// Bearer token captured during discovery. Loopback clients re-read the
+    /// owner-only `serve.token` before each request because remote daemons
+    /// rotate it while a TUI can remain open.
+    ///
+    /// `None` when the daemon was started with `--no-auth`, or when
+    /// `AOE_DAEMON_URL` is set without `AOE_DAEMON_TOKEN`.
+    token: Arc<RwLock<Option<String>>>,
+    local_token_path: Option<PathBuf>,
     pub source: Source,
 }
 
 impl DaemonEndpoint {
+    pub(crate) fn new(base_url: String, token: Option<String>, source: Source) -> Self {
+        Self {
+            base_url,
+            token: Arc::new(RwLock::new(token)),
+            local_token_path: None,
+            source,
+        }
+    }
+
+    pub(crate) fn with_local_token_path(mut self, token_path: PathBuf) -> Self {
+        self.local_token_path = Some(token_path);
+        self
+    }
+
     /// Same base URL, scheme rewritten to `ws://` / `wss://` so a
     /// caller can hand it to `tokio_tungstenite::connect_async`.
     pub fn ws_base_url(&self) -> String {
         http_to_ws(&self.base_url)
+    }
+
+    /// Resolve the credential to send now rather than relying forever on the
+    /// discovery-time snapshot. Only loopback local-daemon discovery may
+    /// consult the app directory; explicit env and legacy public endpoints
+    /// must never receive a token read for some other local daemon.
+    pub(crate) fn resolved_token(&self) -> Option<String> {
+        let cached = self.cached_token();
+        if self.source != Source::LocalDaemon || !is_loopback(&self.base_url) {
+            return cached;
+        }
+        let Some(token_path) = self.local_token_path.as_deref() else {
+            return cached;
+        };
+        self.resolved_token_from_path(token_path)
+    }
+
+    fn resolved_token_from_path(&self, token_path: &Path) -> Option<String> {
+        let cached = self.cached_token();
+        cached.as_ref()?;
+        if self.source != Source::LocalDaemon || !is_loopback(&self.base_url) {
+            return cached;
+        }
+        let Some(current) = read_valid_token(token_path) else {
+            return cached;
+        };
+        *self.token.write().unwrap_or_else(|e| e.into_inner()) = Some(current.clone());
+        Some(current)
+    }
+
+    pub(crate) fn has_token(&self) -> bool {
+        self.cached_token().is_some()
+    }
+
+    pub(crate) fn cached_token(&self) -> Option<String> {
+        self.token.read().unwrap_or_else(|e| e.into_inner()).clone()
     }
 }
 
@@ -81,11 +137,11 @@ pub fn discover_env() -> Option<DaemonEndpoint> {
         .ok()
         .map(|t| t.trim().to_string())
         .filter(|t| !t.is_empty());
-    Some(DaemonEndpoint {
-        base_url: trim_query(url).trim_end_matches('/').to_string(),
+    Some(DaemonEndpoint::new(
+        trim_query(url).trim_end_matches('/').to_string(),
         token,
-        source: Source::Env,
-    })
+        Source::Env,
+    ))
 }
 
 /// Local serve daemon discovery. Returns `Err(NoLocalDaemon)` when no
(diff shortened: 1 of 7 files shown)


src/acp/client/discovery.rs
```diff
@@ -145,6 +212,16 @@ fn extract_token(url: &str) -> Option<&str> {
     None
 }
 
+fn read_valid_token(path: &Path) -> Option<String> {
+    let token = std::fs::read_to_string(path).ok()?;
+    let token = token.trim();
+    let valid_len = token.len() == 64 || token.len() == 32;
+    let valid_chars = token
+        .chars()
+        .all(|c| c.is_ascii_digit() || ('a'..='f').contains(&c));
```
jerome-benoit, line 221: CodeRabbit's MAJOR is correctly resolved here: lowercase-hex only (`0-9a-f`), replacing the earlier `is_ascii_hexdigit() || is_ascii_lowercase()` that wrongly accepted uppercase hex and `g-z`. And it's regression-guarded by `local_endpoint_rejects_non_lowercase_hex_tokens` (asserts `A*64` and `g*64` fall back to the cached token). Good.

src/acp/client/discovery.rs
```diff
@@ -28,19 +30,73 @@ pub struct DaemonEndpoint {
...
         http_to_ws(&self.base_url)
     }
+
+    /// Resolve the credential to send now rather than relying forever on the
+    /// discovery-time snapshot. Only loopback local-daemon discovery may
+    /// consult the app directory; explicit env and legacy public endpoints
+    /// must never receive a token read for some other local daemon.
+    pub(crate) fn resolved_token(&self) -> Option<String> {
+        let cached = self.cached_token();
+        if self.source != Source::LocalDaemon || !is_loopback(&self.base_url) {
```
jerome-benoit, line 71: This is the security gate, and it's correct: disk-token refresh only when `source == LocalDaemon` AND `is_loopback` (which parses the host as an IpAddr, so lookalikes like `127.0.0.1.evil`/punycode are rejected). Combined with binding the token path only inside the loopback discovery branch and a per-endpoint Arc store, env/legacy-public endpoints never receive the local-daemon token. Suggest pinning `is_loopback("http://127.1")` / `http://2130706433` in a test to lock the numeric-shorthand behavior against future url-crate changes.

jerome-benoit (approved):
## Review: refresh local TUI token after rotation — LGTM

Verified in depth: three independent read-only reviewers (security, correctness/concurrency, tests + CodeRabbit follow-up), two adversarial waves, and my own build + test run in an isolated checkout. It holds, and both CodeRabbit findings are resolved with regression guards.

### Security (verified)
- `is_loopback` parses the URL host as an `IpAddr` (not string matching), so hostile lookalikes are rejected: `127.0.0.1.evil`, `user@evil`, `localhost.evil`, punycode/unicode, trailing-dot. Numeric shorthands (`127.1`, `2130706433`, `0x7f...`) either fail Rust's strict `IpAddr` parse (fail-safe -> cached token) or canonicalize to genuine loopback - no path treats a public host as loopback.
- Disk-token refresh is gated by `source == Source::LocalDaemon && is_loopback(...)`, and the token path is bound only inside the loopback discovery branch, so env/legacy-public endpoints never read `serve.token`. The last-valid token store is a per-endpoint `Arc<RwLock<Option<String>>>` shared across clones (not global), so there is no cross-endpoint-kind leak.

### Correctness / concurrency (verified)
- Both transports resolve fresh: all 16 HTTP request builders funnel through `auth()` -> `resolved_token()`, and the WS handshake through `ws_url()` -> `resolved_token()`. No send site keeps a discovery-time snapshot.
- No await held across the lock; poison-safe via `into_inner()`. A partial/invalid token read keeps the last valid cached token (never clears it). Poll policy: 401 breaks the poll, 429 keeps the existing lockout, transient errors retry - paced by a 3s interval, no busy-loop, no token `unwrap`.
- WS reconnect is bounded (3 attempts, 250/500/1000ms backoff) and resolves the fresh token each attempt - no reconnect storm.

### CodeRabbit findings - both resolved + guarded
- MAJOR (`read_valid_token` char class): now lowercase-hex only, with a regression test covering the previously-accepted `A*64` (uppercase) and `g*64` (non-hex lowercase).
- MINOR (`auth_skips_bearer_when_no_token`): now builds the request and asserts the `Authorization` header is absent.

My run: `cargo test --features serve --lib` for the token surface = 33/0 (discovery/http/ws + `plugin_poll_stops_only_for_unauthorized`).

### Non-blocking notes
1. The daemon writes `serve.token` non-atomically (in-place truncate+write, not temp-then-rename). This is harmless here - truncate zeroes the length first, so a concurrent reader sees a *short* partial (rejected by the 32/64 length gate), never a valid-shaped wrong token, and the 5-minute server-side grace window keeps the previous token valid across the rewrite. Still, a temp-file-then-rename write would remove the torn-read window entirely; worth considering.
2. The poll-policy test exercises the `should_retry` predicate (401 vs 429) but not a generic `Transport` error nor the loop's `break` wiring directly; and `is_loopback` would benefit from pinned tests for `http://127.1` / `http://2130706433` to lock the numeric-shorthand behavior against a future url-crate change.

Solid, security-conscious fix. Approving.

njbrake (approved):

Approving. This is a well-scoped fix for a real problem: the structured-view TUI snapshots the daemon bearer token at discovery time, so after a ~4h rotation it keeps sending the stale credential and the 3s plugin poll can exhaust the per-IP auth-failure budget and trigger 429 lockouts.

What I checked:

- **Token sharing / refresh.** Moving the token behind `Arc<RwLock<Option<String>>>` so HTTP and WS clones share one cell, and re-reading `serve.token` via `resolved_token()` before each request, is the right shape. Disk refresh is correctly gated on `Source::LocalDaemon` AND a parsed-loopback `base_url`, so env and legacy-public endpoints never get another daemon's token substituted in.
- **`is_loopback` hardening.** Switching from `starts_with` to `Url::parse` + `IpAddr::is_loopback` closes the lookalike-hostname bypasses (`127.0.0.1.evil.example`, `localhost.attacker.example`). Nice catch to fold in.
- **Partial-write guard.** `read_valid_token` length + lowercase-hex check handles the non-atomic `serve.token` rewrite. On Linux/macOS the only observable partial state is empty (rejected), never a 32-char prefix, so this holds.
- **Poll break on 401.** The permanent break rests on local credentials refreshing from disk. I confirmed the server side: `rotate()` keeps the old token valid for a 5-minute grace window and rewrites the file, so a legitimate local 401 is effectively unreachable. Good.

Both CodeRabbit findings (looser-than-documented hex check; the no-token test not inspecting the header) are already addressed in 49ae784, and I verified the current source and reran the tests.

Tests pass locally under `cargo test --features serve` for the changed modules, and clippy is clean on the diff.

One out-of-scope follow-up if you want it: the server's own `is_valid_token_format` (`src/server/mod.rs`) still has the loose `is_ascii_hexdigit() || is_ascii_lowercase()` check that CodeRabbit flagged on the client side. It only ever validates self-generated 64-char hex today, so it is harmless, but it would be reasonable to tighten it to match the stricter client validator.

Issue #2153, opened 2026-06-15
BUG: Destructive session recovery: a transient resume/probe failure permanently delinks the session


bytestrom:
Hi guys, 

absolutly amazing tool, a daily driver for me! Keep up to good work and WOLOLO. <3

Since sync my cc and codex sessions across hosts with syncthing, i discovered this issue: 

*Report written by claude code with opus 4.8*: 

### Environment
  - AoE 1.11.0, nixos on wsl2 on Windows
  - Agents: Claude Code and Codex (both reproduce)

  ### Summary
  On startup recovery, when a resumed agent pane exits during AoE's probe, AoE  **clears the stored `agent_session_id`** and restarts the session fresh. The fresh start is then assigned a new id that overwrites the original, so the prior conversation can no longer be resumed — even though the cause of the probe failure is often transient. A momentary failure becomes permanent data loss.

  ### Reproduction
  1. Have N sessions with resumable history.
  2. Ctrl-C out of aoe and Reboot (the tmux server dies; all panes are gone).
  3. Let AoE run startup recovery and mass-`--resume` the sessions.
  4. For any session whose `--resume` exits quickly within the probe window (e.g. the transcript is briefly unavailable, or the agent is slow to come
     up under the concurrent-recovery load), AoE clears its sid.

  ### Observed (logs, redacted)
  INFO  session.startup_recovery: TUI starting recovery for missing tmux sessions count=24
  WARN  session.store: start: resume with sid <uuid> crashed pane within probe; clearing sid and retrying without resume
  INFO  session.store: Session ID changed for <id>: <new-uuid>
  WARN  session.startup_recovery: recovery cascade failed id=<id> error=fresh restart after resume fallback crashed within probe (stale sid <uuid> was cleared; underlying  issue persists) The same session fails this way across several reboots, each time with a  different (freshly assigned) sid — the original is gone after the first clear.

  ### Expected
  A transient probe/resume failure should not erase the stored session id. Preserve it and surface a recoverable "resume failed" state so the user can
  retry later; never overwrite the stored id with a fresh session's id unless the user explicitly chooses to start fresh.

  ### Suggested direction
  - On probe failure, keep `agent_session_id`; mark the session as recovery-failed rather than reassigning.
  - Retry with backoff before giving up.
  - Only persist a new id after a fresh start the user opted into.

njbrake:
@jerome-benoit would you be interested in chiming in about this issue? You are the session recovery expert :)

PR #3006, opened 2026-07-20
fix(recovery): skip resume when the agent is alive on an orphaned tmux server (#2994)

### LINKED ISSUE #2994: Sequential recovery passes duplicate/orphan tmux sessions when a prior pass exits before persisting success
## Summary

`try_acquire_recovery_lock`'s cross-process exclusion (`src/session/recovery.rs`) only guards against the *concurrent* case documented in its own module comment: "A late-starting daemon cannot duplicate a TUI's in-flight workers" / "A late-starting TUI cannot duplicate a daemon's in-flight workers." It does not guard against the *sequential* case: a recovery pass that creates tmux session(s) and then the owning process (TUI or daemon) exits/crashes **before** that success is persisted back to session tracking. A later, independent recovery pass (TUI or daemon, started minutes later, well outside any lock contention window) has no record that those sessions were already resumed, finds them "missing" by its own bookkeeping, and recreates them — orphaning the first batch's tmux server and its `--resume` child processes.

## What we observed (v1.13.0, Linux/WSL2)

From `~/.config/agent-of-empires/debug.log` (timestamps UTC):

```
2026-07-20T00:22:54.183836Z  INFO session.startup_recovery: TUI starting recovery for missing tmux sessions count=18
2026-07-20T00:22:57.497161Z  INFO session.startup_recovery: resumed id=7bb64b6b97cc4f55 title=Dominación
...(16 more "resumed id=..." lines through 00:23:31)...
```

29 minutes later, an unrelated `aoe serve` invocation (bringing up the web dashboard) triggers its own full pass:

```
2026-07-20T00:51:20.701290Z  INFO serve.daemon: starting foreground serve profile= host=127.0.0.1 port=8080 mode="local" auth=Passphrase
2026-07-20T00:51:20.837768Z  INFO session.startup_recovery: starting daemon recovery for missing tmux sessions count=14
2026-07-20T00:51:23.461972Z  INFO session.startup_recovery: recovery completed instance_id=2c5354acaafe479a title=4A8 - Dashboards ECharts outcome=Resumed
...(13 more, through 00:51:43)...
```

10 of the 14 instance IDs in the second pass are identical to IDs already resumed in the first pass (e.g. `2c5354acaafe479a`, `f3f8395f86784885`, `84b7730ff9f94fd4`, `4f1981fe2e584039`, `9a6e2372947949f0`, `bc738bb4ad394efa`, `c8a2a23a4c0e4529`, `74e592de8a194b8f`, `dd9a76cf8e5a449c`, `f27f891e6ab44fec`).

At the OS level this left two entirely disjoint process trees: the original tmux server from the 00:22 pass had `PPID=1` (its own parent — the TUI process — had died) and kept running 18 orphaned `claude --resume <id>` children (one of them a duplicate of the currently-live session driving this very report). A fresh tmux server from the 00:51 pass was serving the actual live sessions. Total measured cost: 4.52GB RSS across the orphaned batch, undetected until manual `ps`/memory investigation ~40 minutes later — on a host already tight on memory, this directly contributed to a second resource crunch during crash recovery.

## Why the lock didn't catch it

The lock is "held for the entire recovery pass" and released on drop — so it correctly serializes the two passes (00:22 and 00:51 never overlapped). But release-on-drop only prevents *simultaneous* duplication; it says nothing about whether the *previous* pass's work actually got recorded before its owner exited. If the TUI process that ran the 00:22 pass terminated (crash, or the surrounding environment being torn down — this happened during recovery from an unrelated host-level resource crash) before whatever persists "this instance is now resumed" ran, the 00:51 pass had no way to know.

## Suggested direction (not prescribing implementation)

Before creating a new tmux session for a "missing" instance, reconcile against *actual* tmux state (`tmux list-panes`/equivalent) for a session matching that instance's tag/id, not only against the in-memory/persisted "missing" determination — i.e. treat live tmux as the source of truth, sessions.json as a cache. Alternatively/additionally, persist the "resumed" state transition atomically with (or before) the `tmux new-session` call, so a mid-pass crash fails toward "we think it's resumed, verify" rather than "we have no record it happened."

## Environment

- `aoe 1.13.0` (Linux amd64)
- Ubuntu on WSL2
- Trigger: unrelated host memory/CPU exhaustion event caused the environment around the TUI process to be torn down mid-recovery; the daemon recovery pass ran independently ~29 minutes later during manual recovery

njbrake:
## Description

Fixes #2994.

Startup session recovery decided candidacy entirely from **live tmux state on the one socket this process resolves to**, with nothing durable recording that an instance had already been resumed. Release builds use tmux's default socket (`/tmp/tmux-<uid>/default`). When that socket directory is wiped mid-crash (observed on WSL2 during a host resource crunch), the prior recovery pass's tmux server is orphaned and invisible; a later, independent pass (TUI or daemon, minutes later, well outside any lock window) finds every session "missing" by its own bookkeeping and re-resumes them. That orphans the first batch's agent processes (4.52GB RSS in the report).

The recovery `flock` only serializes *concurrent* passes. It says nothing about whether a prior pass's work outlived its owner, which is the *sequential* case this bug is about.

### Fix

Two layers, following review on this PR:

**1. Deterministic backstop: a host-local, boot-scoped recovery-attempt ledger.** Before the cascade runs `tmux new-session`, recovery records each candidate id in a ledger keyed by the host boot id (`/proc/sys/kernel/random/boot_id` on Linux, `kern.boottime` on macOS), under the app dir. A later pass drops any id already attempted this boot, so startup recovery is idempotent per boot for **every** agent, regardless of whether its live process is identifiable. A reboot (which genuinely kills every agent) changes the boot id and re-enables recovery. This is the durable fix that closes the window even for agents with no usable process-scan needle.

**2. Defense-in-depth: a live-orphan process scan.** On top of the ledger, before recreating a "missing" session, one batched process-table walk checks whether the agent is positively still alive on a tmux server this process cannot see, using two identity signals:
- `AOE_INSTANCE_ID=<id>` matched as an anchored environment *entry* (aoe injects it for hook agents; `/proc/<pid>/environ` survives an agent rewriting its own argv).
- the `agent_session_id` in the command line, for non-hook agents (the host `docker exec` argv carries it for sandboxed sessions).

Hook agents match only on the env marker (not the bare sid), so a live `--fork-session` child carrying the parent's sid cannot suppress the parent's recovery.

Safe by construction:
- Genuine post-reboot recovery is unaffected: the boot id changes and the agent processes are gone.
- The scan is only consulted for sessions tmux already reports missing, so the happy path is untouched.
- The scan runs off the async executor (daemon Phase A/B wrap it in `spawn_blocking`), batched to one process-table walk regardless of candidate count.

Wired into all three recovery decision points: the TUI candidate loop, the daemon Phase A filter, and the daemon Phase B post-lock re-check. New primitive `process::processes_matching` (`/proc/<pid>/{cmdline,environ}` on Linux, `ps -A -ww -E -o command=` on macOS).

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (none needed; internal recovery behavior)
- [ ] For UI changes: included screenshot or recording (no UI change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a full-binary e2e, because: the trigger is an environmental tmux-socket loss (a `/tmp` wipe orphaning a server) that isn't reproducible deterministically in the e2e harness. Instead, wire-level and unit coverage exercise the real recovery path:
  - `server::tests::daemon_recovery_ledger_and_scan_exclude_candidates` drives the actual `daemon_startup_recovery_mark` (Phase A) and asserts both guards flip real candidate selection: an id already attempted this boot is excluded (ledger), and an id whose agent is a live decoy process is excluded (scan). Ledger isolated to a tempdir via `AOE_RECOVERY_ATTEMPT_DIR`.
  - `recovery::tests::recovery_attempt_ledger_roundtrips` covers the ledger round-trip.
  - `process::tests` cover cmdline detection, anchored env-entry detection (and prefix non-match), and empty input.
  - `recovery::tests` cover sid/env detection, short/absent sid rejection, and the no-match (post-reboot) case.

### Testing

- `cargo build` and `cargo build --features serve`: clean.
- `cargo fmt --check`, `cargo clippy --features serve --all-targets`: clean (2 pre-existing warnings unrelated to this diff).
- `cargo test --features serve --lib`: 5501 passed. The only 3 failures (`acp::supervisor::…load_errs_without_peer`, `acp::worker_registry::…load_err`, `tui::home::…resume_failed_after_async_restart`) reproduce identically on `main` without this diff; they are root-in-container environment artifacts (chmod-based unreadable-file fixtures that root bypasses), not caused by this change.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Root-cause investigation and implementation done via Claude Code, reviewed and directed by @njbrake.

## Summary

- Fixed startup recovery duplicating agent processes when tmux’s socket directory is removed and the old tmux server becomes orphaned.
- Made recovery idempotent per boot by adding a host-local “recovery attempt” ledger, so the same instance won’t be recovered twice during the same restart (with best-effort cleanup of stale ledgers).
- Added a cross-platform check to detect whether the “missing” agent is already running on the host, primarily by matching an exact `AOE_INSTANCE_ID` environment marker (with session-id / argv matching as a fallback).
- Applied the safeguards across the full recovery flow:
  - the TUI candidate selection step
  - daemon Phase A candidate filtering
  - daemon Phase B post-lock re-check (only proceeding when tmux is still missing, the candidate is valid, and no live orphan agent is detected).
- Improved reliability of live-process detection by scanning the OS process table in batches (Linux `/proc`, macOS `ps`) and running these scans outside the async executor and relevant locks.

## Benefits

This prevents “double-resume” cascades caused by orphaned tmux servers, while keeping recovery behavior unchanged when the live agent genuinely isn’t present.

Relying on string-based matching is inherently best-effort; integrating more direct tmux-provided identity/session tracking (e.g., PID/session mapping) would further reduce edge-case mismatches.

### DIFF
--- a/src/process/linux.rs
+++ b/src/process/linux.rs
@@ -45,6 +45,30 @@ fn build_children_map() -> HashMap<u32, Vec<u32>> {
     children_map
 }
 
+/// Return `true` if any live process has `needle` as a substring of its argv.
+/// Reads `/proc/<pid>/cmdline` (NUL-separated argv, rendered with spaces) for
+/// every numeric `/proc` entry; skips entries that vanish or are unreadable
+/// mid-scan. Best-effort: an unreadable `/proc` returns `false`.
+pub(super) fn any_process_cmdline_contains(needle: &str) -> bool {
+    let Ok(entries) = fs::read_dir("/proc") else {
+        return false;
+    };
+    for entry in entries.flatten() {
+        let name = entry.file_name();
+        if name.to_string_lossy().parse::<u32>().is_err() {
+            continue;
+        }
+        let Ok(bytes) = fs::read(entry.path().join("cmdline")) else {
+            continue;
+        };
+        let cmdline = String::from_utf8_lossy(&bytes).replace('\0', " ");
+        if cmdline.contains(needle) {
+            return true;
+        }
+    }
+    false
+}
+
 /// Get the foreground process group leader for a shell PID
 /// Walks the process tree to find the actual foreground process
 pub fn get_foreground_pid(shell_pid: u32) -> Option<u32> {
--- a/src/process/linux.rs
+++ b/src/process/linux.rs
@@ -45,11 +45,13 @@ fn build_children_map() -> HashMap<u32, Vec<u32>> {
     children_map
 }
 
-/// Return `true` if any live process has `needle` as a substring of its argv.
-/// Reads `/proc/<pid>/cmdline` (NUL-separated argv, rendered with spaces) for
-/// every numeric `/proc` entry; skips entries that vanish or are unreadable
-/// mid-scan. Best-effort: an unreadable `/proc` returns `false`.
-pub(super) fn any_process_cmdline_contains(needle: &str) -> bool {
+/// Return `true` if any live process has one of `needles` as a substring of
+/// its argv or environment. Reads `/proc/<pid>/cmdline` and `/proc/<pid>/environ`
+/// (both NUL-separated, rendered with spaces) for every numeric `/proc` entry;
+/// `environ` is owner-only, so only same-uid processes (which include our own
+/// agent children) contribute an environment match. Skips entries that vanish
+/// or are unreadable mid-scan. Best-effort: an unreadable `/proc` returns `false`.
+pub(super) fn any_process_cmdline_or_env_contains(needles: &[&str]) -> bool {
     let Ok(entries) = fs::read_dir("/proc") else {
         return false;
     };
@@ -58,11 +60,19 @@ pub(super) fn any_process_cmdline_contains(needle: &str) -> bool {
         if name.to_string_lossy().parse::<u32>().is_err() {
             continue;
         }
-        let Ok(bytes) = fs::read(entry.path().join("cmdline")) else {
-            continue;
+        let dir = entry.path();
+        let read_nul_joined = |file: &str| {
+            fs::read(dir.join(file))
+                .ok()
+                .map(|bytes| String::from_utf8_lossy(&bytes).replace('\0', " "))
+                .unwrap_or_default()
         };
-        let cmdline = String::from_utf8_lossy(&bytes).replace('\0', " ");
-        if cmdline.contains(needle) {
+        let cmdline = read_nul_joined("cmdline");
+        let environ = read_nul_joined("environ");
+        if needles
+            .iter()
+            .any(|n| !n.is_empty() && (cmdline.contains(n) || environ.contains(n)))
+        {
             return true;
         }
     }
--- a/src/process/linux.rs
+++ b/src/process/linux.rs
@@ -45,38 +45,58 @@ fn build_children_map() -> HashMap<u32, Vec<u32>> {
     children_map
 }
 
-/// Return `true` if any live process has one of `needles` as a substring of
-/// its argv or environment. Reads `/proc/<pid>/cmdline` and `/proc/<pid>/environ`
-/// (both NUL-separated, rendered with spaces) for every numeric `/proc` entry;
-/// `environ` is owner-only, so only same-uid processes (which include our own
-/// agent children) contribute an environment match. Skips entries that vanish
-/// or are unreadable mid-scan. Best-effort: an unreadable `/proc` returns `false`.
-pub(super) fn any_process_cmdline_or_env_contains(needles: &[&str]) -> bool {
+/// One `/proc` walk deciding, for each candidate `i`, whether a live process
+/// belongs to it: an `/proc/<pid>/environ` *entry* exactly equals
+/// `env_needles[i]` (NUL-delimited, so no prefix-collision), or
+/// `/proc/<pid>/cmdline` contains `cmdline_needles[i]`. `environ` is owner-only,
+/// so only same-uid processes (our agent children among them) contribute an
+/// environment match. Skips entries that vanish or are unreadable mid-scan;
+/// stops early once every candidate is matched. Best-effort: an unreadable
+/// `/proc` yields all `false`.
+pub(super) fn processes_matching(
+    env_needles: &[String],
+    cmdline_needles: &[Option<String>],
+) -> Vec<bool> {
+    let n = env_needles.len();
+    let mut found = vec![false; n];
+    let mut remaining = n;
     let Ok(entries) = fs::read_dir("/proc") else {
-        return false;
+        return found;
     };
     for entry in entries.flatten() {
+        if remaining == 0 {
+            break;
+        }
         let name = entry.file_name();
         if name.to_string_lossy().parse::<u32>().is_err() {
             continue;
         }
         let dir = entry.path();
-        let read_nul_joined = |file: &str| {
-            fs::read(dir.join(file))
-                .ok()
-                .map(|bytes| String::from_utf8_lossy(&bytes).replace('\0', " "))
-                .unwrap_or_default()
-        };
-        let cmdline = read_nul_joined("cmdline");
-        let environ = read_nul_joined("environ");
-        if needles
-            .iter()
-            .any(|n| !n.is_empty() && (cmdline.contains(n) || environ.contains(n)))
-        {
-            return true;
+
+        let environ_raw = fs::read(dir.join("environ")).unwrap_or_default();
+        let environ = String::from_utf8_lossy(&environ_raw);
+        let env_entries: std::collections::HashSet<&str> =
+            environ.split('\0').filter(|s| !s.is_empty()).collect();
+
+        let cmd_raw = fs::read(dir.join("cmdline")).unwrap_or_default();
+        let cmdline = String::from_utf8_lossy(&cmd_raw).replace('\0', " ");
+
+        for i in 0..n {
+            if found[i] {
+                continue;
+            }
+            let env_hit =
+                !env_needles[i].is_empty() && env_entries.contains(env_needles[i].as_str());
+            let cmd_hit = cmdline_needles[i]
+                .as_deref()
+                .is_some_and(|s| !s.is_empty() && cmdline.contains(s));
+            if env_hit || cmd_hit {
+                found[i] = true;
+                remaining -= 1;
+            }
         }
     }
-    false
+    found
 }
 
 /// Get the foreground process group leader for a shell PID
--- a/src/process/macos.rs
+++ b/src/process/macos.rs
@@ -35,26 +35,45 @@ fn build_children_map() -> HashMap<u32, Vec<u32>> {
     children_map
 }
 
-/// Return `true` if any live process has one of `needles` as a substring of
-/// its command line or environment. Parses `ps -A -ww -E -o command=`: `-ww`
-/// disables column truncation so a long argv/env carrying a needle is not cut
-/// off, and `-E` appends each (owner-owned) process's environment to the
-/// command column so an `AOE_INSTANCE_ID=<id>` env marker matches too. If a
-/// `ps` build rejects `-E`, the call fails closed to `false` and recovery
-/// falls back to today's behavior. Best-effort: a failed `ps` returns `false`.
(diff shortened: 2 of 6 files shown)


jerome-benoit:
I'm not convinced by the approach chosen to look for a process; it's quite fragile since it's based on string. I'm pretty sure there's ways to retrieve a PID or keep track of them for tmux. tmux is not already doing it and expose the state of each session?

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks @jerome-benoit, this pushed the fix in a better direction. You're right that a plain sid substring match is fragile, and I reworked it. Two constraints shaped where it landed, and I want to lay them out since one of them is a trap worth knowing about.

**Why a process-table check rather than asking tmux.** The trigger here is a mid-crash `/tmp` wipe (WSL2) that unlinks the tmux socket file. Once the socket is gone, the orphaned server is unreachable by any `tmux` query, and a fresh pass resolves a new default socket, which is why `has_live_tmux_pane()` reports the session missing in the first place. After that socket loss the OS process table is the only state that still knows the agent is alive, so a tmux-side liveness check cannot see what we need it to see.

**Why not a persisted PID.** I looked hard at tracking the PID, since that is the sturdier instinct. The problem is that several users (including the #2153 reporter) sync `sessions.json` across hosts with Syncthing. A PID persisted on host A is meaningless, and dangerously reusable, on host B, so a bare persisted PID would trade this bug for a cross-host one unless it carries a host/boot-id guard. The process-table check sidesteps that because it only ever inspects the local table.

**What changed in response to your point.** Instead of matching the sid in argv, the guard now primarily matches aoe's own durable marker, `AOE_INSTANCE_ID=<id>`, in the agent's environment. `/proc/<pid>/environ` (and `ps -E` on macOS) survives an agent rewriting its own argv or process title, which is the concrete fragility you were pointing at. The sid in argv stays as a universal fallback for agents without hooks and for the host `docker exec` argv of sandboxed sessions. Separately, CodeRabbit was right that the scan was running under the `instances` read lock; it now runs outside it.

If you would still prefer a durable-state fast path, I think the clean version is a host/boot-guarded pane PID persisted at resume, with the process-table check as the fallback for the crash-before-persist window, which is what this issue is about. Happy to go that way if you think it is worth the extra state; I wanted to align on approach with you first given you own the sibling fix in #2224.

jerome-benoit (commented):
Root-cause reasoning is sound: after a `/tmp` wipe the orphaned server's socket is unlinked, so `tmux` queries are structurally blind to it while the OS process table is host-global; scanning it (and rejecting a cross-host-unsafe persisted PID) is well argued. Wiring is consistent across the three decision points, the scan runs after dropping the `instances` read lock, failure paths fail closed, and the scanner cannot self-match the recovering `aoe` process (the `AOE_INSTANCE_ID` marker is injected into the agent child's env, not aoe's own, and aoe's argv carries no sid). Verified locally on `fdcbf1671`: `cargo fmt`, `cargo clippy --features serve --all-targets` (only the two pre-existing warnings), the `process`/`recovery` unit tests plus the daemon wire test, and `RUSTDOCFLAGS="-D warnings" cargo doc --features serve`, all clean.

Non-blocking notes (the failure model correctly defers a resume rather than duplicating):

**Major**

1. The scan runs synchronously on the tokio executor, once per candidate, in both daemon phases: `daemon_startup_recovery_mark`'s `candidates.retain(|i| orphaned_agent_process_alive(i))` under the recovery flock, and the Phase B re-check while holding the per-instance lock. Each call is a full `/proc` walk (or `ps -A` fork), so ~2N scans per startup on an executor thread. `orphaned_agent_process_alive`'s own doc says to call it off any async lock/executor, and the recovery work itself is already `spawn_blocking`-wrapped while the scan is not. Prefer one batched pass (walk once, test all candidates' needles) wrapped in `tokio::task::spawn_blocking`.
2. This is a mitigation, not the durable fix the issue suggested. It only works while the orphan is alive and identifiable, so a non-hook agent with a sid shorter than `ORPHAN_SCAN_MIN_SID_LEN` (or none) has no needle and the duplication can still occur. A "resume-attempted" marker persisted before `tmux new-session` (host-local, so no cross-host-PID concern) would close the window deterministically. Fine to land as defense-in-depth, but please either add the marker or frame this explicitly as a partial mitigation with a follow-up.

**Minor**

- Fork sid collision: `build_fork_flags` launches a claude fork as `--resume <parent_sid> --fork-session --session-id <child_sid>`, so a live fork-child carries the parent's `agent_session_id` in its argv. `orphaned_agent_process_alive(parent)` uses that sid as a needle, so a live child can skip the parent's recovery when the parent's own process is dead (fails safe, self-heals next launch). For hook-enabled agents (claude is one), gating the skip on the `AOE_INSTANCE_ID` env marker instead of the bare sid avoids it.
- macOS env detection is untested and the comment overstates it: the `AOE_INSTANCE_ID` env path is covered only by Linux-gated unit tests; the note that the macOS argv test exercises `ps -E` is inaccurate (that test's marker is in argv, not env), and the daemon wire test is not cfg-gated yet proves the skip via an env-only decoy, so it can flake on a hardened macOS where `ps -E` hides env. Add a macOS env test (or an argv fallback needle) or soften the claim.
- The `AOE_INSTANCE_ID=<id>` env needle is a delimiter-less substring match, safe today only because `generate_id()` is fixed 16-hex; `validate_instance_id` allows up to 64 bytes, so a longer legacy id could prefix-collide. Anchor the match or pin the fixed-width assumption in a comment; likewise the 8-char sid floor is low versus the real 16/36-char widths.

**Nits**

- The `recovery.rs` docstring introduces an em dash in the `orphaned_agent_process_alive` summary, which AGENTS.md forbids in human-authored prose (the doc `-D warnings` gate does not catch it).
- Tests build the marker from the literal `"AOE_INSTANCE_ID"` instead of `crate::tmux::env::AOE_INSTANCE_ID_KEY`.
- The PR description is stale: it names `any_process_cmdline_contains` / `orphaned_resume_child_alive` and "argv only", but the code is `any_process_cmdline_or_env_contains` / `orphaned_agent_process_alive` with an `AOE_INSTANCE_ID` env signal the description never mentions.

Thanks for the fail-safe design and the real-socket wire test.

njbrake:
_Note: this reply was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Thanks for the thorough pass, this is all addressed in the latest push.

**Durable fix (your Major #2).** Added a host-local recovery-attempt ledger keyed by boot id, written before `tmux new-session`. Startup recovery is now idempotent per boot for every agent, so the crash-then-re-run duplication is closed deterministically even for a non-hook agent with a short or absent sid, not just when the orphan is identifiable. A reboot rotates the boot id and re-enables recovery; a synced `sessions.json` is unaffected because the ledger is host-local, never in the session record. The process scan is now defense-in-depth on top of it.

**Executor (your Major #1).** The scan is batched to a single process-table walk for all candidates, and both daemon phases run it inside `spawn_blocking`, off the executor and outside the locks.

**Fork sid collision.** Hook agents now match only on the `AOE_INSTANCE_ID` env marker, so a live `--fork-session` child carrying the parent's sid can no longer suppress the parent's recovery. Non-hook agents keep the sid needle.

**Anchoring.** `AOE_INSTANCE_ID` is matched as an exact environ entry (NUL-delimited on Linux, a token on macOS), not a substring, so a longer legacy id cannot prefix-collide.

**macOS test.** The wire test's scan case now detects via argv on a non-hook agent, so it no longer relies on `ps -E` env visibility that a hardened macOS can hide, and it gained a ledger-idempotency case.

**Nits.** Em dash gone, tests use `AOE_INSTANCE_ID_KEY`, and the PR description is rewritten to match the code (ledger plus the env-aware scan).

The wire test drives the real `daemon_startup_recovery_mark` and asserts both the ledger and the scan actually flip candidate selection. Appreciate the review pushing this to the deterministic version.

PR #3017, opened 2026-07-20
feat: pin a cheap model for smart-rename one-shot title calls

### LINKED ISSUE #3009: smart-rename: pin a cheaper model for the title one-shot

### Problem

Smart session rename shells out to the agent's one-shot CLI (`claude -p "<prompt>"`, `codex exec …`, etc.) via `build_oneshot_argv` in `src/session/smart_rename.rs`. That builder emits `[binary, oneshot_flag, extra_args…, prompt, trailing_args…]` with **no model flag**, so the rename uses whatever model the agent CLI defaults to. Generating a 3-to-5-word title is a trivial task that does not need a frontier model, so on CLIs defaulting to a large model we pay top-tier per-call cost for a throwaway title.

This affects **both** the ACP path (#2201) and the terminal path (#2973), since they share `build_oneshot_argv`. It is pre-existing behavior, surfaced while reviewing #2973.

### Proposal

Add a per-agent `oneshot_model_args()` on `AgentDef` (mirroring the existing `oneshot_extra_args()` / `oneshot_trailing_args()`), returning the CLI-specific model flag, and inject it in `build_oneshot_argv` between `oneshot_flag` and the prompt so the no-injection contract still holds. Candidate flags:

- claude: `--model claude-haiku-4-5`
- codex: `-m <model>`
- gemini: `-m <model>`
- copilot: `--model <model>`
- opencode / kimi: confirm the flag each CLI accepts

Optionally back it with a `smart_rename_model` setting (default: a cheap model, empty means "CLI default") so users can override. Because it is a single insertion point in the shared builder, both harness paths get it at once.

### Notes

- Keep the fallback to the CLI default when no flag is known for an agent, so unmapped agents are unaffected.
- Verify each flag against the pinned CLI version before wiring it (a wrong model id should fail closed to keeping the civ name, which the existing non-zero-exit handling already does).

jerome-benoit:
## Description

Closes #3009.

Smart-rename generates a session title by shelling out to the agent's one-shot CLI (`claude -p "<prompt>"`, `codex exec …`) via `build_oneshot_argv`. That builder emitted no model flag, so a throwaway three-to-five-word title billed whatever model the CLI defaults to (a frontier model on some CLIs).

This PR does two things:

1. **Pins a cheap built-in model for the title one-shot**, scoped to smart-rename titles only. `AgentDef::oneshot_cheap_model()` returns a STABLE, non-dated alias where one exists; only `claude` qualifies today (`haiku`). Every other agent returns `None` (CLI default), because their cheap models are dated/rotating slugs (e.g. codex `gpt-5.6-luna`) or config-specific ids with no stable alias, and AoE pins no CLI version, so hardcoding one would silently break on the next catalog rotation. `conversation_summary` deliberately keeps the CLI default (it reads the whole transcript and may need a bigger model turn).

2. **Adds a per-agent user-configurable model** (`session.smart_rename_model`, requested by @njbrake). A user can pin the model each agent uses for the title, e.g. `{ claude = "haiku", codex = "gpt-5.6-luna" }` — including the dated slugs we won't hardcode as a built-in default.

### Mechanism

- `AgentDef` exposes three small methods (mirroring `oneshot_extra_args` / `oneshot_trailing_args`): `oneshot_model_flag()` (the selector token: claude/copilot `--model`, codex/gemini/opencode/kimi `-m`), `oneshot_cheap_model()` (built-in cheap alias; claude `haiku` only), and `oneshot_flag_binds_prompt()` (whether the one-shot flag consumes the next token as its value).
- `resolve_title_model_args(agent, &smart_rename_model)` resolves the three-state config: **absent key** → built-in default; **empty value** → force the CLI default (opt out of the built-in pin); **non-empty value** → pin that model via the agent's flag.
- `build_oneshot_argv(agent, prompt, OneshotModel)` places the resolved model selector **before** the prompt for positional-prompt flags (claude boolean `-p`, codex `exec`, opencode `run`) and **after** the prompt for value-binding flags (copilot/gemini/kimi `-p`, whose value IS the prompt) — verified against each CLI's source. The `OneshotModel { Title(Vec<String>), CliDefault }` enum has no zero-value, so every call site states its intent: the two title sites pass `Title(resolve_title_model_args(...))`, the summary passes `CliDefault`. The prompt stays a single argv element (no shell), so the no-injection contract holds.

The model id is free-form (no live-CLI validation is possible); an id the CLI rejects simply exits non-zero, which `run_oneshot` already maps to keeping the generated name (fail-closed). An agent with no model flag ignores a configured value rather than mis-injecting it.

### Surfaces

- **Config**: `smart_rename_model: HashMap<String, String>`, `custom:smart-rename-model` widget, gated behind `web = "elevation:…"` (it influences the rename one-shot argv).
- **Web**: a per-agent widget (`SmartRenameModelWidget`) with one free-text row per installed one-shot-capable agent; clearing a row falls back to the built-in default.
- **TUI**: a raw-JSON custom arm (mirrors `acp-defaults`); this is the only surface that can set the empty-string "force CLI default" state.
- **Docs**: `docs/guides/configuration.md` documents the setting and its three-state semantics.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt`, `cargo clippy` (TUI + `--features serve`), and `cargo test --lib` (4186 passed, 0 failed) are clean. Web: `format:check`, `lint`, `tsc -b`, and Vitest (`customWidgets.test.tsx`) all pass; the coverage-matrix entry (`settings.custom-widgets`) covers the new widget.

## Test Coverage Analysis

- [x] Added or updated a Vitest spec: `SmartRenameModelWidget` tests in `web/src/components/settings/__tests__/customWidgets.test.tsx` (per-agent rows, override payload, clear-to-default). The `settings.custom-widgets` coverage-matrix entry already maps this file.

**TUI / CLI (e2e test)**
- [x] Skipped an e2e test, because: the behavior is pure argv construction with no tmux/network/filesystem dependency and is fully covered by unit tests — `resolve_title_model_args` precedence (absent/empty/whitespace/padded/pinned), per-agent override argv shape (positional vs value-binding placement) for all six one-shot agents, summary-uses-CLI-default, plus `AgentDef` drift guards (only-claude-has-a-cheap-default, model-flag set, value-binding set + the extra_args-placement invariant) and a TUI JSON round-trip test.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude (Opus) via an agentic CLI.

**Any Additional AI Details you'd like to share:** Per-agent CLI facts (model flags, and which one-shot `-p` flags are value-binding vs boolean) were verified against each CLI's source before wiring; the built-in cheap default is restricted to stable, non-dated aliases (only claude `haiku`).

## Summary

- Added per-agent model selection for smart-rename title generation.
- Claude now defaults to Haiku, while users can choose custom models or keep each agent’s CLI default.
- Ensured model arguments are placed correctly across supported agents.
- Kept conversation summaries unaffected by title model settings.
- Added web and TUI configuration controls, documentation, and comprehensive tests.

## Benefits

Users can generate titles more quickly and affordably, with flexible model choices tailored to each agent. Invalid model IDs fail safely without losing the generated session name.

## Potential enhancements

Future improvements could include live model validation and clearer feedback when a configured model is rejected.


Seluj78:
Hey @jerome-benoit (and @njbrake) 

FYI the settings already have a settings to choose what to use to do smart renames. I personally have set it to opencode:



I think this settings works for the TUI too ?

njbrake:
@Seluj78 yes in this case, my ask is to be able to set model, not agent harness. Opencode is the harness, but I want to be able to set what model should be used inside of the harness.

For example, If my default model in claude code is Opus 4.8, I don't want to use that model for something like picking a name, I would want to use Claude Haiku.

jerome-benoit:
> my ask is to be able to set what model should be used inside of the harness [...] If my default model in claude code is Opus 4.8, I don't want to use that model for something like picking a name, I would want to use Claude Haiku.

Exactly, model not harness. That is what this PR does today, with two deliberate limits:

- It pins `--model haiku` for claude's title one-shot, so your Opus 4.8 default is not billed for a throwaway name.
- `conversation_summary` stays on the CLI default (it reads the whole transcript and can need a bigger model turn), so only the rename title is downgraded.
- Other harnesses fall back to the CLI default (`&[]`) because I could not verify a stable, non-dated alias for them (codex/gemini/copilot ids rotate or are entitlement-gated; opencode/kimi ids are config-specific), so they stay fail-closed rather than pinning something that could silently break.

The user-**tunable** model is not in this PR yet. A single global `smart_rename_model` string does not work, since a model id is agent-specific (`haiku` is invalid for codex/gemini), so the right shape is a per-agent map: a new settings-schema field plus web/TUI widget, validation, and coverage-matrix tests, i.e. a materially larger change.

How would you like to proceed?

1. Keep this PR minimal (haiku default for claude) and track the tunable in a follow-up issue/PR (my suggestion).
2. Extend this PR's scope to add the per-agent tunable now.

Happy either way, just let me know.

njbrake:
> > my ask is to be able to set what model should be used inside of the harness [...] If my default model in claude code is Opus 4.8, I don't want to use that model for something like picking a name, I would want to use Claude Haiku.
> 
> Exactly, model not harness. That is what this PR does today, with two deliberate limits:
> 
>     * It pins `--model haiku` for claude's title one-shot, so your Opus 4.8 default is not billed for a throwaway name.
> 
>     * `conversation_summary` stays on the CLI default (it reads the whole transcript and can need a bigger model turn), so only the rename title is downgraded.
> 
>     * Other harnesses fall back to the CLI default (`&[]`) because I could not verify a stable, non-dated alias for them (codex/gemini/copilot ids rotate or are entitlement-gated; opencode/kimi ids are config-specific), so they stay fail-closed rather than pinning something that could silently break.
> 
> 
> The user-**tunable** model is not in this PR yet. A single global `smart_rename_model` string does not work, since a model id is agent-specific (`haiku` is invalid for codex/gemini), so the right shape is a per-agent map: a new settings-schema field plus web/TUI widget, validation, and coverage-matrix tests, i.e. a materially larger change.
> 
> How would you like to proceed?
> 
>     1. Keep this PR minimal (haiku default for claude) and track the tunable in a follow-up issue/PR (my suggestion).
> 
>     2. Extend this PR's scope to add the per-agent tunable now.
> 
> 
> Happy either way, just let me know.

Lets go with 2: extend the scope to add a per-agent tunable. Thanks!

Seluj78:
2 LGTM ! :D

njbrake (approved):

Approve. This lands the per-agent tunable scope we agreed on in the thread. I verified the value-binding classification against the actual agent definitions (claude, opencode, and codex are positional; copilot, gemini, and kimi are `-p` value-binding), so the model selector is placed correctly on both sides; all argv and precedence tests pass.

One thing worth flagging for anyone reading: this shifts the default for every claude user, since rename titles now use the `haiku` alias instead of the session default model. That is intended and can be opted out of with an empty value.

On CodeRabbit's `haiku` note: fine as documentation, not blocking. The fallback is graceful (an unknown alias keeps the generated name, no cost regression), and the new tunable lets a user pin any id.

Issue #2977, opened 2026-07-18
ACP runner termination Phase C: server-method proxying + retire relay socket (#1054)


Seluj78:
Follow-up to #1054 (parent) and PR #2975 (Phase A). Depends on Phase B.

Phase C: the runner proxies the 10 agent-to-daemon server methods (permission, elicitation, fs read/write, terminal create/output/wait/kill/release) to the daemon over the control channel as opaque JSON, with a stable server-call id that survives daemon disconnect (replayed on reattach). Daemon handlers become `dispatch(method, params) -> Value`. The daemon drops its `ConnectionTo<Agent>` for the runner path; the raw `<id>.sock` relay is retired.

Includes the `RUNNER_VERSION` bump to 2 and migration: a v2 daemon inheriting a live v1 runner mid-turn keeps the legacy attach + resume-idle watchdog for that session until idle, then reaps and respawns v2. Reverse-call idempotency: at-least-once for 9 methods, scoped idempotency key for `terminal/create` only.

Issue #2899, opened 2026-07-16
Structured view: session stuck Running after build-stale mid-turn adopt, no terminal Stopped on clean completion


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

A structured-view session finishes its turn cleanly (final assistant message rendered, agent waiting for input), but the sidebar / composer stays stuck showing the turn as still running. The Send affordance stays hidden behind the spinner; the only escape is the Stop button or sending a new prompt.

This reproduces whenever `aoe` is rebuilt while a turn is in flight, which for anyone developing `aoe` is constant. Observed on session `d0a9cad1ce9840e1` (release namespace).

## Evidence from the affected session

Lifecycle events from `~/.agent-of-empires/acp_events.db`. The last user prompt is seq 2513; after it the turn runs to a clean finish but no `Stopped` is ever recorded:

```
seq 2513  UserPromptSent          16:09:02
seq 2567  PromptCapabilities      16:10:29   <- new connection handshake, MID-TURN
seq 2568  AcpSessionAssigned      16:10:29
seq 2569..2668  ToolCall* / AgentMessageChunk   (turn keeps streaming)
seq 2669  AgentMessageChunk "2 first?"   16:17:25   <- final assistant message
seq 2670  UsageUpdated (cost: null)      16:17:26
seq 2671  UsageUpdated (cost: 2.80052)   16:17:26   <- cost-populated end-of-turn marker
           ... no Stopped, ever ...
```

`debug.log` names the exact mechanism. The user rebuilt `aoe` (1.12.1 -> 1.13.0) mid-turn, so the supervisor adopted the build-stale worker to drain the in-flight turn before respawning:

```
16:10:29  INFO acp.supervisor: adopting build-stale structured view worker to drain in-flight turn before respawn
          session=d0a9cad1ce9840e1 old_build=1.12.1+gaf73ab... new_build=1.13.0+gc2e972...
16:10:29  INFO acp.protocol: initialize handshake complete mode=Resume { acp_session_id: "5e1bfa8f...", in_flight_turn: true }
16:10:29  INFO acp.protocol: resume mode: reusing existing acp session id without handshake
16:10:30  INFO acp.protocol: resume-idle watchdog: disarming, in-flight turn is observable
          ... turn completes 16:17:26 ...
```

[...]

PR #3056, opened 2026-07-21
feat(acp): runner owns the ACP handshake and turn over v2 control channel (#2976)

### LINKED ISSUE #2976: ACP runner termination Phase B: runner owns the ACP handshake (#1054)
Follow-up to #1054 (parent) and PR #2975 (Phase A).

Phase B: move the ACP handshake into the runner. The runner sends `initialize` + `session/new` (or `session/load` with the cached id) ONCE on first agent spawn and caches the responses; subsequent daemon connects do not forward those through. The daemon sends typed `Prompt`/`Cancel` over the control channel.

Removes the daemon-side `initialize` re-send on every reconnect and the `ConnectMode::Resume` skip-new/load trick for the runner path.

Design (from the #1054 debate): hand-rolled narrow JSON-RPC router in the runner, not the crate `ConnectionTo<Agent>`; deserialize only the runner-owned response bodies (initialize caps/version, session id, prompt stopReason); everything else stays opaque.

Seluj78:

## Description

Phase B of runner-side ACP protocol termination (#1054), following PR #2975 (Phase A). The `aoe __acp-runner` shim now terminates the ACP client side of the protocol: it owns the handshake and the turn, and the daemon speaks a small typed control protocol to it instead of running `initialize` / `session/*` / `session/prompt` / `session/cancel` over the byte relay itself.

What changes:

- **Control protocol v2** (`control_protocol.rs`): `CONTROL_PROTOCOL_VERSION` bumped 1 to 2. New handshake frames (`Initialize`/`Initialized`, `EstablishSession`/`SessionReady`, `HandshakeFailed`) and turn frames (`Prompt`/`Cancel`). Phase A's `stop_reason`-only `PromptCompleted` becomes a typed `PromptOutcome` (`Completed` / `Error` / `Aborted`) so an agent error envelope is surfaced (with its `data`, preserving rate-limit classification) rather than collapsed into a silent stop.
- **Runner** (`runner.rs`): on the first v2 attach the runner runs `initialize` + `session/new|load|fork` against the agent exactly once, caches the raw results, and answers `Initialized` / `SessionReady`. Every later attach replays the cache without touching the agent. Prompts arrive as typed `Prompt` frames; the runner assigns the canonical `session/prompt` id and reports the typed `PromptCompleted`. A cached handshake also lets the runner answer a relay handshake re-send from cache, so a downgrade v1-daemon reattach never double-initializes the agent. The control channel is now persistent (carries the handshake and every turn) rather than the Phase A one-shot adopted-turn delivery.
- **Daemon** (`acp_client.rs`): on the runner path the daemon dials the control socket and, when the runner greets v2, drives the handshake and each turn over a `DaemonControlClient`, deserializing the cached results into the existing crate response types. The crate connection stays attached to the relay only for `session/update` notifications and the server-to-client callbacks (permission / fs / terminal). The watchdog-dense prompt loop body is left intact: the prompt future is adapted so a runner `PromptCompleted` maps back to `Result<PromptResponse, Error>`. An absent or older (v1) runner yields no control client, so the daemon falls back to the exact byte-relay handshake plus resume-idle watchdog behavior; the direct-stdio path is untouched.

Backward compatibility is purely via the `control_protocol_version` handshake, with no runner-version migration (that stays in Phase C, #2977): a v2 daemon on a v1 runner falls back to Phase A; a v1 daemon on a v2 runner drives the handshake over the relay and the runner intercepts it from cache. This removes the redundant per-reconnect `initialize` against the agent.

**Scope note:** this lands the full Phase B (handshake + prompt + cancel over control) in one PR, per the decision on the issue, rather than splitting handshake and turn migration. The `ConnectMode::Resume` skip-new/load trick is retained for the v2 path (it stays correct because the runner keeps the session alive, and its result equals the runner's cache), rather than routing Resume through `EstablishSession`.

Deferred (called out for the follow-ups): relay watermark ordering between `PromptCompleted` and buffered `session/update` (a pre-existing property from Phase A), long-detach ring durability, and `<id>.sock` retirement + `RUNNER_VERSION` migration (Phase C/D, #2977/#2978).

Relates to #1054
Closes #2976

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view ACP session, send a normal prompt, confirm it completes exactly once (goes Idle) with no duplicate stop.
- [ ] Reconnect the dashboard / restart nothing across several turns; confirm the agent is not re-initialized each turn (the session keeps working; no new `initialize` churn in `debug.log`).
- [ ] Mid-turn `aoe serve --stop` then `aoe serve`; confirm the session clears "thinking" promptly when the agent's response lands (native completion), not after the ~30s resume-idle pause.
- [ ] Cancel / force-stop an in-flight turn; confirm the turn ends and the spinner clears.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, a two-model design debate on the protocol seams, the plan, and the implementation were drafted by Claude under my direction. I made the scope and design calls (full Phase B in one PR, adapter-future to keep the prompt loop intact, retaining the Resume skip trick), reviewed each commit, and ran the manual test steps.

## Summary

- Moved the ACP handshake and session establishment into the `aoe __acp-runner` shim (Phase B), so it runs `initialize` and `session/new|load|fork` once, caches the results, and replays them on later attaches.
- Upgraded the runner/daemon control channel to protocol v2, where the daemon drives the turn using typed `Prompt` and `Cancel` messages (instead of re-running initialization on reconnects).
- Added structured, typed prompt completion outcomes (normal stop, agent JSON-RPC error envelopes including attached data, and runner aborts) to make failures easier to interpret and propagate.
- Preserved Phase A fallback behavior by reverting to the older byte-relay approach when a v1/incompatible or unavailable runner is detected.
- Improved robustness around failures by keeping detailed handshake error information, using runner-side request IDs, and aborting promptly when prompt transmission fails.
- Expanded and adjusted test coverage for v2 handshake behavior, cached replay across reconnects, error forwarding fidelity, protocol-version mismatch, and control-socket fallback.

## Benefits

- Avoids repeated initialization/session work on daemon reconnects by shifting responsibility and caching to the runner.
- Makes prompted/resumed interactions more reliable by letting the runner own the handshake/session lifecycle and turn lifecycle.
- Provides clearer, more structured error reporting for agent failures (including any extra error data) while staying compatible with older runners.

## Potential follow-up

- Add more cross-platform integration coverage (beyond current Unix-focused runner control tests) to further validate v2 control compatibility and reconnect replay across varied runner/ACP-agent implementations.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3280,7 +3280,7 @@ async fn attach_runner_control(
         let _write_half = write_half;
         loop {
             match control_protocol::read_frame(&mut read_half).await {
-                Ok(Some(ControlBody::PromptCompleted { stop_reason, .. })) => {
+                Ok(Some(ControlBody::PromptCompleted { outcome, .. })) => {
                     // Claim the one-shot terminal guard. Winning means the
                     // watchdogs stand down; losing means one already fired,
                     // so do not double-emit Stopped.
@@ -3288,7 +3288,7 @@ async fn attach_runner_control(
                         .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
                         .is_ok()
                     {
-                        let reason = control_stop_reason(stop_reason.as_deref());
+                        let reason = control_outcome_reason(&outcome);
                         let _ = event_tx.send(Event::Stopped { reason }).await;
                     }
                     // The adopted turn is resolved; the control channel's
@@ -3315,17 +3315,20 @@ async fn attach_runner_control(
 /// Map a runner-reported prompt outcome to an `Event::Stopped` reason. A
 /// completed turn renders as Idle regardless of stop reason, so the
 /// default is `prompt_complete`; the one reason with special downstream
-/// handling (`rate_limited`) is preserved when the agent reports it.
+/// handling (`rate_limited`) is preserved when the agent reports it. An
+/// agent error-envelope or an aborted turn also renders Idle, so they map
+/// to `prompt_complete` as well; the turn is over either way.
 ///
-/// Phase A deliberately collapses the other ACP stop reasons (`cancelled`,
-/// `max_tokens`, `refusal`, `max_turn_requests`) into `prompt_complete`,
-/// since they all render Idle today. Preserving their identity for the UI
-/// is tracked as a follow-up on the Phase C runner-terminator work (#2977),
-/// where the runner owns the protocol and can forward the typed stop reason
-/// natively.
-fn control_stop_reason(stop_reason: Option<&str>) -> String {
-    match stop_reason {
-        Some("rate_limited") | Some("rate_limit") => "rate_limited".to_string(),
+/// The other ACP stop reasons (`cancelled`, `max_tokens`, `refusal`,
+/// `max_turn_requests`) collapse into `prompt_complete`, since they all
+/// render Idle today. Preserving their identity for the UI is tracked as a
+/// follow-up on the Phase C runner-terminator work (#2977).
+fn control_outcome_reason(outcome: &crate::acp::control_protocol::PromptOutcome) -> String {
+    use crate::acp::control_protocol::PromptOutcome;
+    match outcome {
+        PromptOutcome::Completed {
+            stop_reason: Some(r),
+        } if r == "rate_limited" || r == "rate_limit" => "rate_limited".to_string(),
         _ => "prompt_complete".to_string(),
     }
 }
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3226,13 +3226,20 @@ impl DaemonControlClient {
     }
 
     /// Run the ACP `initialize` the runner owns; returns the raw result
-    /// value to deserialize into `InitializeResponse`.
-    async fn initialize(&self, request: serde_json::Value) -> Result<serde_json::Value, AcpError> {
-        self.send(ControlBody::Initialize { request }).await?;
+    /// value to deserialize into `InitializeResponse`. A `HandshakeFailed`
+    /// is surfaced as the reconstructed crate error so the caller propagates
+    /// the same `AgentStartupError` (with `data.details`) the relay path did.
+    async fn initialize(
+        &self,
+        request: serde_json::Value,
+    ) -> Result<serde_json::Value, agent_client_protocol::Error> {
+        self.send(ControlBody::Initialize { request })
+            .await
+            .map_err(|e| acp_internal_error(format!("control write failed: {e}")))?;
         match self.handshake_rx.lock().await.recv().await {
             Some(ControlBody::Initialized { result }) => Ok(result),
-            Some(ControlBody::HandshakeFailed { message }) => Err(AcpError::Spawn(message)),
-            _ => Err(AcpError::Spawn(
+            Some(ControlBody::HandshakeFailed { error }) => Err(acp_error_from_value(error)),
+            _ => Err(acp_internal_error(
                 "control channel closed during initialize".into(),
             )),
         }
@@ -3240,24 +3247,25 @@ impl DaemonControlClient {
 
     /// Run the session-creation request the runner owns; returns
     /// `(acp_session_id, raw result)` to deserialize into the matching
-    /// session response.
+    /// session response, or the reconstructed crate error on failure.
     async fn establish_session(
         &self,
         method: &str,
         request: serde_json::Value,
-    ) -> Result<(String, serde_json::Value), AcpError> {
+    ) -> Result<(String, serde_json::Value), agent_client_protocol::Error> {
         self.send(ControlBody::EstablishSession {
             method: method.to_string(),
             request,
         })
-        .await?;
+        .await
+        .map_err(|e| acp_internal_error(format!("control write failed: {e}")))?;
         match self.handshake_rx.lock().await.recv().await {
             Some(ControlBody::SessionReady {
                 acp_session_id,
                 result,
             }) => Ok((acp_session_id, result)),
-            Some(ControlBody::HandshakeFailed { message }) => Err(AcpError::Spawn(message)),
-            _ => Err(AcpError::Spawn(
+            Some(ControlBody::HandshakeFailed { error }) => Err(acp_error_from_value(error)),
+            _ => Err(acp_internal_error(
                 "control channel closed during session establishment".into(),
             )),
         }
@@ -3272,7 +3280,15 @@ impl DaemonControlClient {
     ) -> oneshot::Receiver<control_protocol::PromptOutcome> {
         let (tx, rx) = oneshot::channel();
         *self.completion.lock().expect("completion mutex poisoned") = Some(tx);
-        let _ = self.send(ControlBody::Prompt { request }).await;
+        if self.send(ControlBody::Prompt { request }).await.is_err() {
+            // Write failed: drop the parked sender so `rx` resolves to Err ->
+            // Aborted immediately instead of hanging until the cancel /
+            // orphan watchdog eventually unwedges the turn.
+            self.completion
+                .lock()
+                .expect("completion mutex poisoned")
+                .take();
+        }
         rx
     }
 
@@ -3426,10 +3442,21 @@ fn acp_internal_error(message: String) -> agent_client_protocol::Error {
     err
 }
 
+/// Reconstruct a crate `Error` from the raw JSON-RPC error object the
+/// runner forwarded in `HandshakeFailed`. Preserves `code` / `message` /
+/// `data` so the downstream `AgentStartupError` surfaces the same
+/// `data.details` remediation the byte-relay handshake did; falls back to a
+/// generic internal error if the object is malformed.
+fn acp_error_from_value(error: serde_json::Value) -> agent_client_protocol::Error {
+    serde_json::from_value(error.clone())
+        .unwrap_or_else(|_| acp_internal_error(format!("runner handshake failed: {error}")))
+}
+
 /// Drive a session-creation request over the v2 control channel and
 /// deserialize the runner's cached result into the crate response type,
 /// so each `session/new|load|fork` site's `Result<Resp, Error>` matches
-/// the crate `send_request` path it replaces.
+/// the crate `send_request` path it replaces (including the failure path:
+/// the runner-forwarded agent error propagates verbatim).
 async fn establish_session_v2<Resp: serde::de::DeserializeOwned>(
     control: &DaemonControlClient,
     method: &str,
@@ -3437,10 +3464,7 @@ async fn establish_session_v2<Resp: serde::de::DeserializeOwned>(
 ) -> Result<Resp, agent_client_protocol::Error> {
     let params = serde_json::to_value(request)
         .map_err(|e| acp_internal_error(format!("serialize {method} params: {e}")))?;
-    let (_id, result) = control
-        .establish_session(method, params)
-        .await
-        .map_err(|e| acp_internal_error(format!("runner {method}: {e}")))?;
+    let (_id, result) = control.establish_session(method, params).await?;
     serde_json::from_value(result)
         .map_err(|e| acp_internal_error(format!("deserialize {method} result: {e}")))
 }
@@ -5903,10 +5927,7 @@ async fn run_connection_task<W, R>(
             let init: InitializeResponse = if let Some(control) = control_client.as_ref() {
                 let params = serde_json::to_value(build_initialize_request())
                     .map_err(|e| acp_internal_error(format!("serialize initialize params: {e}")))?;
-                let result = control
-                    .initialize(params)
-                    .await
-                    .map_err(|e| acp_internal_error(format!("runner initialize: {e}")))?;
+                let result = control.initialize(params).await?;
                 serde_json::from_value(result)
                     .map_err(|e| acp_internal_error(format!("deserialize initialize result: {e}")))?
             } else {
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -23,9 +23,10 @@ use agent_client_protocol::schema::v1::{
     CreateElicitationRequest, CreateElicitationResponse, CreateTerminalRequest,
     CreateTerminalResponse, ElicitationAction, ElicitationCapabilities,
     ElicitationFormCapabilities, EmbeddedResource, EmbeddedResourceResource,
-    FileSystemCapabilities, ForkSessionRequest, ImageContent, Implementation, InitializeRequest,
-    KillTerminalRequest, KillTerminalResponse, LoadSessionRequest, McpServer, MessageId,
-    NewSessionRequest, PermissionOptionKind, PromptRequest, ReadTextFileRequest,
+    FileSystemCapabilities, ForkSessionRequest, ForkSessionResponse, ImageContent, Implementation,
+    InitializeRequest, InitializeResponse, KillTerminalRequest, KillTerminalResponse,
+    LoadSessionRequest, LoadSessionResponse, McpServer, MessageId, NewSessionRequest,
+    NewSessionResponse, PermissionOptionKind, PromptRequest, PromptResponse, ReadTextFileRequest,
     ReadTextFileResponse, ReleaseTerminalRequest, ReleaseTerminalResponse,
     RequestPermissionOutcome, RequestPermissionRequest, RequestPermissionResponse,
     SelectedPermissionOutcome, SessionConfigId, SessionConfigValueId, SessionId,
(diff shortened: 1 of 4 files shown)


src/acp/acp_client.rs
```diff
@@ -3205,51 +3200,125 @@ async fn wait_for_socket(
...
+    /// assigns the `session/prompt` id and reports `PromptCompleted`.
+    async fn prompt(
+        &self,
+        request: serde_json::Value,
+    ) -> oneshot::Receiver<control_protocol::PromptOutcome> {
+        let (tx, rx) = oneshot::channel();
+        *self.completion.lock().expect("completion mutex poisoned") = Some(tx);
+        let _ = self.send(ControlBody::Prompt { request }).await;
+        rx
+    }
```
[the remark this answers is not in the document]
Seluj78: Good catch, fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/97f6dfb1. The completion sender is now dropped when the control `Prompt` write fails, so `rx` resolves to `Aborted` immediately instead of parking until the watchdog.

src/process/runner.rs
```diff
@@ -689,6 +693,33 @@ struct RunnerShared {
...
+    /// longer issues requests over the relay on the v2 path, so this is the
+    /// sole client-request id space and cannot collide.
+    next_req_id: AtomicI64,
+    /// Responders for runner-issued request/response round-trips awaited
+    /// inline (`initialize`, `session/*`). The stdout fanout routes a
+    /// matching response into the oneshot and swallows it instead of
+    /// forwarding it to the daemon. Prompt responses are NOT here; they use
+    /// the async `prompt_requests` path that fires `PromptCompleted`.
+    pending_client_responses: Mutex<HashMap<i64, tokio::sync::oneshot::Sender<serde_json::Value>>>,
+}
```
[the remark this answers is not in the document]
Seluj78: Valid, fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/97f6dfb1. On the v2 path the daemon still issues `set_mode`/`set_config_option` over the relay via the crate connection (ids from ~1), so the runner now seeds `next_req_id` from a disjoint high base (`1 << 48`). That keeps the two id spaces from overlapping, so an agent response routes to exactly one waiter and the wire never carries two live requests with the same id.

src/acp/control_protocol.rs
```diff
@@ -45,21 +62,81 @@ pub enum ControlBody {
...
+    /// from cache on later attaches.
+    SessionReady {
+        acp_session_id: String,
+        result: serde_json::Value,
+    },
+    /// The runner-owned handshake failed (agent incompatible, `session/new`
+    /// error, transport failure). The daemon surfaces this as an
+    /// `AgentStartupError` instead of hanging on a handshake that will
+    /// never complete.
+    HandshakeFailed { message: String },
```
jerome-benoit: [BLOCKER] `HandshakeFailed` carries only `message: String`, no `data` - unlike `PromptOutcome::Error` below (line 135) which keeps `data: Option<Value>`. So `handshake_result` (runner.rs:1320) drops the agent's error `data` on a `session/new` failure, and the native-binary/arch-mismatch remediation (in `data.details`) is lost. The frontend banner regex `/native binary at .* exists but failed to launch/i` then fails to match, which is why the unmodified `startup-error-banner-native-binary.spec.ts` (Playwright live shard 1/2) is red on this PR but green on the merge-base. Carry `data` here (mirror `PromptOutcome::Error`), reconstruct it into the startup-error path, and add a HandshakeFailed-with-`data` test.

src/acp/acp_client.rs
```diff
@@ -3205,51 +3200,125 @@ async fn wait_for_socket(
...
+    /// Issue a turn: register the completion oneshot, send the `Prompt`
+    /// frame, and return the receiver the prompt loop awaits. The runner
+    /// assigns the `session/prompt` id and reports `PromptCompleted`.
+    async fn prompt(
+        &self,
+        request: serde_json::Value,
+    ) -> oneshot::Receiver<control_protocol::PromptOutcome> {
+        let (tx, rx) = oneshot::channel();
+        *self.completion.lock().expect("completion mutex poisoned") = Some(tx);
+        let _ = self.send(ControlBody::Prompt { request }).await;
```
jerome-benoit: [minor] `tx` is stashed in `completion` (3274), then the `Prompt` send is fire-and-forget (`let _ =`, 3275). If the control write fails, nothing ever resolves that oneshot, so the prompt future parks until the cancel-escalation/orphan watchdog unwedges it - defeating Phase B's deterministic completion exactly when the socket is failing. Drop the slot on send-`Err` so the existing consumer maps a dropped sender to `PromptOutcome::Aborted` and ends the turn promptly. A `Drop`/reader-exit drain of `completion` would additionally close the mid-turn socket-close park.

src/process/runner.rs
```diff
@@ -689,6 +693,33 @@ struct RunnerShared {
...
+    /// Runner-owned ACP handshake cache (#2976 Phase B). Populated the
+    /// first time a v2 daemon drives `initialize` / `session/new|load|fork`
+    /// through the control channel; replayed verbatim on every later
+    /// attach so the agent is handshaken exactly once.
+    handshake: Mutex<RunnerHandshake>,
+    /// Monotonic JSON-RPC id allocator for the requests the runner issues
+    /// to the agent on its own (`initialize`, `session/*`, `session/prompt`)
+    /// now that it owns the client side of the protocol. The daemon no
+    /// longer issues requests over the relay on the v2 path, so this is the
+    /// sole client-request id space and cannot collide.
```
jerome-benoit: [nit] This comment ("the daemon no longer issues requests over the relay on the v2 path, so this is the sole client-request id space and cannot collide") is misleading: `set_mode` / `set_config_option` / `delete_session` still go over the relay to the same agent on the v2 path. There is no real collision, but the reason is different: the crate allocates UUID-string request ids (`RequestId::Str(uuid)`, agent-client-protocol jsonrpc.rs:2381) while this uses `i64`, so the id spaces are disjoint by type. Please state that real reason here (and note it would break if a future crate bump switched to numeric ids). This is also why the "prompt-id overlap" concern does not hold.

jerome-benoit (changes requested):
## Review: Phase B (runner owns the ACP handshake + turn over control v2)

Strong, well-scoped Phase B. I verified the load-bearing invariants and they hold: handshake caching is provably exactly-once (the integration test counts each method reaching the agent once across two attaches), the v1<->v2 compat matrix degrades correctly in all four combos behind a strict `== 2` version gate, the prompt-path `PromptOutcome` mapping is lossless with rate-limit classification preserved, and no Phase-A fix (stale-completion gate, single-mutex TOCTOU, write-timeout, one-shot terminal guard) is regressed. Nice work.

One blocker and a few smaller items.

### BLOCKER - the runner-owned handshake drops the error `data`, and it is turning CI red
`HandshakeFailed { message: String }` (`control_protocol.rs:85`) carries no `data`, unlike `PromptOutcome::Error` which keeps `data: Option<Value>` (`control_protocol.rs:135`). So `handshake_result` (`runner.rs:1320`) reduces an agent `session/new` error envelope to `message` + `code` and discards `err.data`. The native-binary / arch-mismatch remediation string lives in `data.details`, so on the v2 path the frontend banner regex (`/native binary at .* exists but failed to launch/i`) no longer matches. That is why `web/tests/live/acp-stories/startup-error-banner-native-binary.spec.ts` fails on `Playwright (live, shard 1/2)`: the spec is unmodified by this PR and passed on the merge-base, and the received string is now `"...runner session/new: agent spawn failed: session/new failed: Internal error (code -32603)"` (note the new `runner` prefix, and the missing detail). It is a deterministic, PR-caused functional regression, not a flake: a user whose agent binary has an arch mismatch loses the actionable banner. Fix: carry `data` on `HandshakeFailed` (mirror `PromptOutcome::Error`) and reconstruct it into the startup-error path, plus add a HandshakeFailed-with-`data` test - the prompt path has `parse_response_surfaces_error_envelope`, the handshake path has no equivalent, which is exactly why this slipped past the Rust suite.

### MINOR - `DaemonControlClient::prompt` swallows a failed control-write and strands the completion sender
`acp_client.rs:3274-3275` stashes `tx` in `completion`, then `let _ = self.send(ControlBody::Prompt ...)`. If the write fails, nothing ever resolves that oneshot, so the turn parks until the cancel-escalation / orphan watchdog unwedges it, defeating Phase B's deterministic completion exactly when the socket is dying. Drop the slot on send-`Err` so the consumer maps it to `Aborted` immediately. (A `Drop`/reader-exit drain of `completion` would also close the mid-turn socket-close park.)

### MINOR - daemon-side control write has no timeout
Unlike the runner's `CONTROL_WRITE_TIMEOUT`, the daemon's control `write` is unbounded, so a wedged runner read could block `prompt`/`cancel` on the write lock. Consider a bounded write for symmetry.

### MINOR - missing tests
HandshakeFailed-with-`data` (would have caught the blocker), the v2 `Cancel` frame, and the v1-daemon-on-v2-runner intercept-from-cache combo are untested.

### NIT - the `next_req_id` comment is misleading
`runner.rs:705` says the daemon "no longer issues requests over the relay on the v2 path, so this is the sole client-request id space and cannot collide." But `set_mode` / `set_config_option` / `delete_session` DO still ride the relay to the same agent on the v2 path. There is no actual collision - the crate allocates UUID-string request ids (`RequestId::Str(uuid)`, `agent-client-protocol` `jsonrpc.rs:2381`) while the runner uses `i64`, so the two id spaces are disjoint by type - but please correct the comment to state that real reason (and note it would break if a future crate bump switched to numeric ids). This also refutes the "prompt-id overlap" concern raised elsewhere: with UUID vs i64 there is no overlap.

### NIT
`run_or_replay_session` ignores the `method` arg on a cache hit - correct for today's single-session lifecycle, but a debug-assert if the cached method/key diverges would harden it.

Happy to re-review quickly once the handshake `data` is carried through (with a test); the rest is close.

src/acp/control_protocol.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/97f6dfb1 (pushed just before your review landed). `HandshakeFailed` now carries the raw JSON-RPC error object (`error: serde_json::Value`); `handshake_result` forwards the agent's error verbatim and the daemon reconstructs the crate error, so `data.details` (native-binary/arch remediation) reaches the banner. Added a runner-level regression test.

src/acp/acp_client.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/97f6dfb1. The completion sender is dropped on send-`Err`, so `rx` resolves to `Aborted` immediately instead of parking until the watchdog.

src/process/runner.rs
Seluj78: You're right, corrected in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/742335c7. The crate allocates UUID-string ids (`RequestId::Str`, jsonrpc.rs:2381), so there's no real collision with the runner's integer ids; the comment now says so. Kept the high seed only as defense-in-depth against a future crate switch to integer ids.

tests/acp_runner_control.rs
```diff
@@ -156,7 +156,330 @@ fn runner_reports_native_prompt_complete_over_control_socket() {
...
+    assert_eq!(failed["kind"], "handshake_failed");
+    // The remediation detail survives the control channel intact.
+    assert_eq!(
+        failed["error"]["data"]["details"],
+        "native binary failed to launch"
+    );
+    assert_eq!(failed["error"]["code"], -32603);
 
     let _ = child.kill();
     let _ = child.wait();
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3056/commits/0d76daa4. The spawned runner is now wrapped in a KillOnDrop guard so an assertion failure before the manual kill still reaps it; Scratch already removes its dir on drop.

src/process/runner.rs
jerome-benoit: The seed is a fine, safe hardening - keep it. One correction for the record though: I traced the crate and the collision it guards against isn't actually reachable. `agent-client-protocol` 0.15.1 allocates outgoing request ids as UUID strings (`RequestId::Str(uuid::Uuid::new_v4())`, jsonrpc.rs:2381), passed verbatim through the outgoing actor and `protocol_compat` (which only rewrite the body, never the id) and serialized as a JSON string. The only `RequestId::Number` in the crate is a test fixture (outgoing_actor.rs:174, inside `mod tests`). So the daemon's relay `set_mode`/`set_config_option` ids are strings, disjoint from the runner's i64 by type, and the runner's `.as_i64()` matcher ignores them anyway. So it's not "ids from ~1" - the high base is defense-in-depth against a *future* numeric-id peer, not a present collision. Worth softening the comment to say that; the code itself is correct.

jerome-benoit (approved):
## Re-review (now at `97f6dfb`) - clearing to APPROVE

`97f6dfb` addresses all three items from my last review, verified against the code and by running the tests.

- **[resolved] BLOCKER - handshake dropped the error `data`.** `HandshakeFailed` now carries the raw JSON-RPC error object (`error: serde_json::Value`) instead of a formatted string; the runner forwards the agent's error verbatim (`handshake_result`), and the daemon reconstructs the crate `Error` via `acp_error_from_value` and propagates it with `?`. The crate `Error`'s `Display` re-embeds `data` (pretty-printed), so `AgentStartupError.message` again contains the native-binary/arch-mismatch detail and the frontend regex matches - byte-identical to the byte-relay path, and uniformly for `initialize` + `session/new|load|fork`. The new `runner_forwards_session_error_data_in_handshake_failed` test drives a real runner + fake agent over the control socket and asserts `error.data.details` survives; it passes, and the `startup-error-banner-native-binary` live test is green again.

- **[resolved] prompt-send-drop.** On a failed control `Prompt` write, the completion sender is now dropped, so `rx` resolves to `Aborted` immediately instead of parking until the watchdog. No double-take/race with the reader (both `take()` under the same mutex; a failed send means no `PromptCompleted` can arrive for that turn).

- **[resolved, with one note] disjoint runner ids.** The `1 << 48` seed is a safe, correct hardening - keep it. But (per the inline thread) the collision it guards against isn't actually reachable: the crate allocates outgoing request ids as UUID strings (`RequestId::Str(uuid)`, jsonrpc.rs:2381), disjoint from the runner's i64 by type, and there's no numeric-renumbering compat layer. So the rationale comment ("ids from ~1", "could collide") overstates the risk - suggest softening it to "defense-in-depth against a future numeric-id peer." Code is fine.

No regression: the `HandshakeFailed` shape change stays inside the v2-gated variant (compat matrix untouched), handshake caching / exactly-once is intact, and happy-path + cancel are unchanged. Local runs green: `acp_runner_control` 3/3, `control_protocol` 9/9, `acp::` lib 519/0, clippy clean. CI is all green (33 checks).

Two non-blocking follow-ups: a dedicated unit test for the prompt-send-drop branch (currently correct-by-inspection only), and the pre-existing reader-exit-mid-turn case (a successful-send then socket-close still leans on the silent-orphan/cancel watchdog rather than dropping the parked sender) - out of this PR's scope.

Approving. Nice work on Phase B.

njbrake (approved):

Approving. Phase B is clean and the load-bearing correctness facts check out against source, not just the comments:

- Runner/daemon request-id spaces are disjoint by construction: the crate allocates outgoing ids as `RequestId::Str(Uuid::new_v4())` (agent-client-protocol 0.15.1, jsonrpc.rs), and the runner's `parse_response_id` reads `as_i64`, which is `None` for a string. The `1<<48` seed is belt-and-suspenders on top of that.
- The stray `session/prompt` response the runner still forwards over the relay on the v2 path is benign: the crate's incoming actor logs a single warn for an unmatched response id and drops it; it does not tear down the connection, so `session/update` forwarding is unaffected.
- Version gating and both downgrade directions (v2 daemon on v1 runner, v1 daemon on v2 runner via `intercept_handshake`) behave as documented.

All CI is green, docstring coverage is 100%, and CodeRabbit's three line findings were addressed. I disagree with the bot's "Linked Issues" pre-merge warning: retaining the Resume skip-new/load shortcut on the v2 path is a deliberate, correct scope call, since the runner keeps the session alive and the skipped result equals its cache.

One item to track as a Phase C follow-up rather than block on: the daemon control reader routes `PromptCompleted` by the `completion` slot alone and ignores `prompt_req_id`. In the scenario #2899 designs for, a new prompt superseding an adopted in-flight resumed turn, the agent answers the old turn first and that completion would be routed to the new turn's oneshot, resolving it early. The v1 path disambiguates by id and does not have this. I could not reproduce it end-to-end (the runner integration test plays the daemon role directly and never drives the real `DaemonControlClient` reader), and its reachability depends on whether the UI admits a prompt while an adopted turn is still "Working". Worth a look when Phase C moves the server methods over.

Issue #2808, opened 2026-07-13
Structured view: agent-agnostic "summary of conversation so far", reimplement like smart-rename (ACP won't surface it)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Claude Code's CLI periodically emits a short summary of what the agent has been doing when a turn runs for a while. It is genuinely useful: on a long agentic run you can glance at a recap instead of scrolling the whole transcript. aoe's structured view has no equivalent. On a long session the only way to know "what has this agent done so far" is to read every row.

We want the same running summary in aoe's structured view, and it must be agent-agnostic: it should work for `claude`, `codex`, and any other agent, the same way smart-rename already does. That rules out leaning on the agent's own summary output, because only some agents produce one and the transport does not carry it (see below). The mechanism should mirror smart-rename: aoe generates the summary itself via a one-shot agent invocation.

### ACP does not surface a summary, and cannot be relied on cross-agent

There is no summary / compaction variant in the ACP schema. `SessionUpdate` in `agent-client-protocol-schema-0.14.0/src/v2/client.rs:96` enumerates `UserMessageChunk`, `AgentMessageChunk`, `AgentThoughtChunk`, `StateUpdate`, `ToolCallUpdate`, `PlanUpdate`, and friends. Nothing for a conversation summary.

The `claude-agent-acp` adapter (0.55.0) actually has Claude's compaction summary in the SDK stream and deliberately drops it. `~/.nvm/.../@agentclientprotocol/claude-agent-acp/dist/acp-agent.js:724-726`:

```js
//  local-only commands (e.g. `/context`) and compaction (`/compact`, whose only
//  user messages are the generated summary and a `<local-command-stdout>` replay
//  — neither carries the prompt's uuid). Promoting the head settles those.
```

The adapter goes further and explicitly discards the summary content blocks. `acp-agent.js:4069-4074`:

```js
case "compaction":
case "compaction_delta":
case "advisor_tool_result":
case "mid_conv_system":
case "fallback":
    break;
```

These `break` with no `sessionUpdate`. The summary is consumed internally to settle the turn queue; it is never forwarded. So even for `claude`, the ACP wire gives us no summary. For `codex` and others there may be no summary at all. Conclusion: an ACP-driven summary is a dead end for the agent-agnostic requirement.

### aoe already renders a compaction marker, but no summary text

[...]

PR #3058, opened 2026-07-21
feat: allow "Auto-name now" when smart rename is off

### LINKED ISSUE #3039: Allow session auto rename on demand
I'd like to use an agent to rename the session even if it's turned off to do automatically on session start.

Seluj78:

## Description

The manual "Auto-name now" sidebar action re-applied the exact same eligibility gate as the automatic turn-end trigger, including the `session.smart_rename` on/off check. So on a session whose auto-rename-on-start was disabled, right-clicking and picking "Auto-name now" returned `409 "Smart rename is disabled in settings"` and reported an error, instead of naming the session.

This threads a `force` flag through `try_smart_rename` that bypasses only the disabled gate. The manual endpoint passes it in both places it gates (the preflight in `force_smart_rename` and the spawned task), so an on-demand rename now runs even when automatic renaming is off. The automatic listener passes `force = false`, so turn-end auto-rename still honors the setting.

Every other gate still applies on both paths: a non-structured, already-named, sandboxed, no-one-shot, or command-overridden session is still refused. This mirrors the existing precedent where the manual conversation summary bypasses its own `conversation_summary` setting (#2808).

Backend-only. The sidebar button already renders on `default_name` alone (independent of the setting), so no frontend change was needed.

Closes #3039

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] In settings, turn OFF smart rename (`session.smart_rename = false`).
- [ ] Create a structured-view session and leave it on its generated civilization name (no `--title`, no branch name).
- [ ] Send one prompt and let the turn finish; confirm the session is NOT auto-renamed (setting is off).
- [ ] Right-click the session in the sidebar and pick "Auto-name now"; confirm it does not error and the session gets a generated title shortly after.
- [ ] Rename a session by hand, then confirm "Auto-name now" no longer appears for it (still gated on the default name).

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design call (bypass only the disabled gate, mirror the #2808 summary precedent), reviewed each commit, and ran the manual test steps.

**NOTE:**

## Summary

- Made **"Auto-name now"** work even when **automatic smart renaming** is turned off (so users can trigger a rename on demand without changing settings).
- Added a **manual-only "force" path** through smart-rename eligibility checks:
  - It bypasses only the **"smart rename disabled in settings"** gate.
  - All other restrictions remain enforced (structured-session rules, still-default naming state, sandbox limits, one-shot availability, and command overrides).
- Ensured **automatic turn-end renaming** continues to follow the configured smart-rename setting (no forced bypass for background/automatic runs).
- Expanded support beyond the web API:
  - **TUI now exposes "Auto-name now"** for **terminal and structured sessions** via **`v`/`V`** and the **command palette**.
  - **Structured sessions** call the existing daemon endpoint; **terminal sessions** spawn a detached runner that passes the force flag.
  - Improved on-demand behavior and feedback (including retrying after a failed automatic terminal attempt and using transient status messaging).
- Updated **documentation** to describe the on-demand re-trigger behavior and how the agent selection works for one-shot renaming.
- Updated **tests** to confirm forced/manual renaming bypasses only the intended eligibility gate while still blocking other disqualifiers.

## Benefits

Users can reliably generate a session name **on demand**, even if smart renaming is disabled globally—while the system still prevents renaming in unsupported/unsafe situations.

## Technical notes (if needed)

- Eligibility uses **`setting_on || force`** for the disabled-setting skip reason; other skip reasons (e.g., oneshot missing / command overridden / sandboxing / non-default state / structured requirements) remain unchanged.
- Manual flows send **`--force`** (or equivalent handler input), while automatic scheduling and polling use the non-forced path.


src/server/api/sessions.rs
```diff
@@ -2788,13 +2788,16 @@ pub async fn force_smart_rename(
...
+    // overridden command). Without this, the sidebar would show success while
+    // no title job runs. Passes `setting_on = true` because this is the manual
+    // "Auto-name now" action, which runs on demand even when auto-rename-on-
+    // start is disabled (#3039); the spawned try_smart_rename gets `force =
+    // true` below to match.
     let config = crate::session::profile_config::resolve_config_or_warn(&profile);
     if let Err(reason) = crate::session::smart_rename::check_eligible_resolved(
         structured,
-        config.session.smart_rename,
+        true,
```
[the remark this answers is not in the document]
Seluj78: Good catch, fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/f8bf5fd5. The preflight now snapshots `project_path` and resolves with `resolve_config_with_repo_or_warn`, matching the worker, so a repo-local `smart_rename_agent` / `agent_command_override` can no longer split the 202 from the actual job.

Seluj78:
> @coderabbitai (nitpick, src/session/smart_rename.rs): Exercise every non-disabled gate with `force` enabled (NoOneshot, CommandOverridden).

Done in https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/a059b851. `manual_force_bypasses_only_the_disabled_gate` now asserts the forced path still returns `NoOneshot` and `CommandOverridden` (alongside the existing structured / name / sandbox cases), so broadening `force` later cannot silently bypass them.

njbrake:
Should this also exist in the TUI in some form? I'm open to your thoughts about this

Seluj78:
Deferred this at first since `feat(smart-rename):` is a valid conventional-commit prefix, but doing it now: retitled to `feat: allow "Auto-name now" when smart rename is off` so the pre-merge Title check passes.

Seluj78:
Extended to the native TUI in https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/9b0f1943. The web "Auto-name now" only covered structured sessions via the daemon; the TUI had no on-demand trigger at all. Added a TUI action on a still-default-named session (`v` / `V` key, or the command palette "Auto-name now"):

- Terminal sessions run the local one-shot via a detached `aoe __smart-rename --force` child (works with no daemon).
- Structured sessions POST the daemon `/smart-rename` endpoint through the existing acp HttpClient (mirrors the view-switch flow); the client is serve-gated so a TUI-only build no-ops that branch.
- The hidden subcommand now routes by session kind and threads `force` into `run_terminal_rename`; the poller path stays unforced (still honors the setting).

Note: the home keyspace is saturated (see the Fork/Tips palette-only precedent), so this landed on the free `v`/`V` pair rather than a mnemonic key. Easy to move to palette-only if you'd prefer not to spend a key.

**Test (TUI):** with `session.smart_rename` off, select a still-default session, press `v` (or palette -> "Auto-name now"); a terminal session renames from its first turn, a structured session renames once its daemon is up.

njbrake (changes requested):

Thanks for this, the fix is real and the primary (structured) path is well scoped and well tested. Requesting changes on one reachable gap on the terminal path.

## Terminal "Auto-name now" silently no-ops after a failed automatic attempt

The structured path handles the documented re-run case: `force_smart_rename` explicitly clears the attempted gate (`state.smart_rename_attempted.remove(&id)`) before re-spawning. The terminal path does not. In `run_terminal_rename`:

```rust
if already || structured {   // `already` = persisted Instance.smart_rename_attempted
    return Ok(());
}
```

`force` never reaches this line, so a terminal session with `smart_rename_attempted = true` that is still default-named cannot be re-run. That state is reachable: when the automatic one-shot returns unusable output, `sanitize_title` returns `None`, and `apply_terminal_title(None)` still sets `smart_rename_attempted = true` while leaving the civ name in place. The user then presses `v`, `auto_name_selected` shows "Auto-name started / Generating a title...", the detached child bails at `if already`, and nothing happens.

This contradicts the docs sentence the PR relies on: "If the automatic rename never lands (the one-shot timed out or returned unusable output), right-click the session in the sidebar and pick Auto-name now to re-run it." The #3039 core scenario (smart_rename off) is unaffected, because the poller never attempts when the setting is off, so `already` stays false; that is likely why it slipped through.

## Suggested fix

Change the guard so the forced path bypasses the attempted gate, mirroring what the structured endpoint already does:

```rust
if (already && !force) || structured {
    return Ok(());
}
```

This is safe: the `NameNotDefault` gate in `check_eligible_resolved`, plus `title_is_auto_overwritable` in `apply_terminal_title`, plus the TUI's own up-front `is_default_civ_name` check, all still protect an already-renamed session on the forced path.

If instead the intent is for the terminal path to deliberately not support re-run after an attempt, that is a defensible call, but then `auto_name_selected` should not show the "Auto-name started" dialog for that case, since the current message reads as success while nothing runs.

## Minor (non-blocking)

Feedback asymmetry: the terminal action surfaces a modal `InfoDialog` ("Auto-name started"), the structured action surfaces a transient status line. Consistent feedback would be a nice follow-up, not a blocker.

Everything else looks good: repo-aware preflight resolution, `force = false` on the automatic listener, the expanded `manual_force_bypasses_only_the_disabled_gate` test, and the docs update all check out.

Seluj78:
> @njbrake: Terminal "Auto-name now" silently no-ops after a failed automatic attempt.

Good catch, both fixed.

- **Blocker (guard):** ab9220fa changes the guard to `(already && !force) || structured`, exactly as you suggested, so the forced terminal path re-runs after a failed automatic attempt. NameNotDefault + title_is_auto_overwritable + the TUI's up-front `is_default_civ_name` check still guard an already-named session. https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/ab9220fa
- **Minor (feedback asymmetry):** 44972ae6 switches the terminal path to a transient status line (`SetTransientStatus`) instead of the modal InfoDialog, so both on-demand triggers now behave the same. https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/44972ae6

The #3039 core scenario (smart_rename off) was indeed unaffected, since the poller never attempts when the setting is off; the failed-attempt state only arises with auto-rename on, which is why it slipped through.

src/session/smart_rename.rs
```diff
@@ -884,8 +961,12 @@ pub async fn run_terminal_rename(profile: &str, session_id: &str) -> anyhow::Res
...
     drop(instances);
     // Durable double-check: storage may have propagated a completed attempt (or
-    // a manual rename) since the poller observed the edge.
-    if already || structured {
+    // a manual rename) since the poller observed the edge. `force` (the manual
+    // "Auto-name now") bypasses the attempted gate so a session whose automatic
+    // one-shot already ran but produced no usable title can be re-run, mirroring
+    // how the structured endpoint clears its attempted set. The NameNotDefault
+    // gate below and title_is_auto_overwritable still protect a named session.
+    if (already && !force) || structured {
```
jerome-benoit, line 969: njbrake's fix is correct here, but revert-silent: no test drives `run_terminal_rename` (or `run_smart_rename_now`) with `force=true` on a session where `smart_rename_attempted=true` and the title is still a default civ name. Both integration tests pass `force=false` with `attempted=false`, so reverting `(already && !force)` to `already` breaks nothing. Suggest a regression test: seed `attempted=true` + default title, assert `force=true` re-runs while `force=false` no-ops.

src/server/api/sessions.rs
```diff
@@ -2788,19 +2789,28 @@ pub async fn force_smart_rename(
...
-    let config = crate::session::profile_config::resolve_config_or_warn(&profile);
+    // drop (sandboxed, or a resolved rename agent with no one-shot / an
+    // overridden command). Without this, the sidebar would show success while
+    // no title job runs. Resolves with the SAME repo-aware config the worker
+    // uses (resolve_config_with_repo_or_warn), so a repo-local smart_rename_agent
+    // or agent_command_override cannot make the preflight and worker disagree.
+    // Passes `setting_on = true` because this is the manual "Auto-name now"
+    // action, which runs on demand even when auto-rename-on-start is disabled
+    // (#3039); the spawned try_smart_rename gets `force = true` below to match.
+    let resolved = crate::session::repo_config::resolve_config_with_repo_or_warn(
```
jerome-benoit, line 2800: CodeRabbit's repo-aware preflight is correct, but also revert-silent: no test exercises the `force_smart_rename` handler (the repo-override test calls `resolve_config_with_repo_or_warn` directly; the web specs stub this endpoint). Reverting to `resolve_config_or_warn` here would pass CI. Suggest a handler-level test with a repo-local `smart_rename_agent` / `agent_command_override` asserting the preflight verdict matches the worker.

jerome-benoit (commented):
## Additional review (not a gate)

I re-reviewed on top of @njbrake's and CodeRabbit's rounds; those two findings are resolved at HEAD and the force path is well-scoped. Adding two things they didn't cover.

**Confirmed resolved (verified by reading + a local `cargo test --lib smart_rename` = 44/0):**
- njbrake's terminal no-op: `run_terminal_rename` now bypasses the attempted gate under force (`if (already && !force) || structured`), and `NameNotDefault` + `title_is_auto_overwritable` still protect a named session.
- CodeRabbit's config mismatch: `force_smart_rename` preflight now resolves repo-aware, matching the worker.
- Force relaxes only `setting_on`: the oneshot / command-override / sandbox / name gates all still fire, and the terminal (flock) + structured (InflightGuard + `attempted.insert` + instance_lock) paths have no double-run window.

**Net-new: both fixes are revert-silent (no regression test at the fixed layer).** `manual_force_bypasses_only_the_disabled_gate` is a good test but it exercises the pure `check_eligible` helper, not the force plumbing. Reverting either fix breaks no test:
- (A) reverting `(already && !force)` to `already`: the only tests that call `run_terminal_rename` pass `force=false` and seed `smart_rename_attempted=false`, so they're identical under the revert.
- (B) reverting the endpoint to `resolve_config_or_warn`: no test drives the `force_smart_rename` handler (the repo-aware resolver test calls the resolver directly; the web specs stub the endpoint). Given both were reachable bugs worth fixing, each deserves a guard per the repo's "land the test alongside the code" guidance.

**Net-new: CityHall parity.** `POST /api/sessions/{id}/smart-rename` is a mutating route (spawns an agent one-shot, mutates the title) guarded only by `read_only_block` - no CityHall guard. That matches its sibling `summarize_session`, so it's fine today, but it's exactly the class #2853 is trying to lock down. When #2853's default-deny CityHall boundary lands, this route needs an explicit allow/deny decision (it spends tokens + mutates state), so it isn't silently reachable or silently broken.

src/session/smart_rename.rs
Seluj78: Added a regression test in https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/e64a5d59: `forced_rename_bypasses_a_prior_failed_attempt` seeds a civ-named session with `smart_rename_attempted = true`, asserts the unforced path no-ops and the forced path renames. Reverting `(already && !force)` to `already` fails it.

src/server/api/sessions.rs
Seluj78: Added a handler-level test in https://github.com/agent-of-empires/agent-of-empires/pull/3058/commits/56c86eb5: `force_smart_rename_preflight_honors_repo_local_override` drives `force_smart_rename` with a repo-local `agent_command_override` and asserts the CONFLICT body says the command is overridden. Reverting to `resolve_config_or_warn` misses the override and falls through to the "no prompt yet" path, so the test asserts the body message (both are 409).

njbrake (approved):

All addressed, thanks for the quick turnaround. Verified locally on `56c86eb5`:

- The terminal force path now bypasses the persisted `smart_rename_attempted` gate (`ab9220fa`), matching how the structured endpoint clears its attempted set. A still-default-named terminal session whose automatic one-shot ran but produced no usable title can be re-named on demand.
- `44972ae6` swaps the misleading "Auto-name started" dialog for a transient status, so the terminal path no longer claims success when the child bails.
- Both coverage gaps @jerome-benoit flagged are now pinned: `forced_rename_bypasses_a_prior_failed_attempt` (reverting `(already && !force)` fails it) and `force_smart_rename_preflight_honors_repo_local_override` (asserts the CONFLICT body, so reverting to the non-repo-aware resolve is caught).

Ran the three tests locally, all green, including the tmux integration test. LGTM.

PR #3060, opened 2026-07-22
fix(tui): route paste to dialogs opened over live-send instead of the pane

### LINKED ISSUE #3059: TUI: paste goes to the agent pane instead of the rename/new-session dialog opened over live-send (preview mode)
## Description

While attached to a session in preview / live-send mode (select a session in the sidebar and press Enter), the sidebar still accepts right-clicks, so you can open the context menu and its follow-up dialogs (Rename, New Session, ...) on top of live-send.

In that state the dialog looks focused (cursor sits in the title / group / path input, typed characters land in the field), but a clipboard paste does NOT go into the input: it streams straight into the AI CLI prompt in the preview pane behind the dialog.

## Steps to reproduce

1. `aoe` -> select a session in the sidebar -> press Enter (live-send / preview mode).
2. Right-click a sidebar row -> Rename (or New Session).
3. Focus the Title / Group / Path field and paste from the clipboard.

**Expected:** pasted text appears in the focused dialog input.
**Actual:** pasted text is sent to the agent's tmux pane (the AI CLI prompt) in the preview.

## Cause

The key path and the paste path route differently:

- `handle_key` (`src/tui/home/input.rs`) only lets live-send capture keys when no overlay is open: `if self.live_send.is_some() && !self.has_non_live_send_overlay()`. That's why typing works in the dialog.
- `handle_paste` (`src/tui/home/input.rs`) checks `self.live_send` first with **no** overlay guard ("Live-send mode wins above every dialog"), so a paste always streams to the pane while attached, even with a dialog on top.

The `handle_paste` comment was written under the assumption that dialogs can't open during live-send (keyboard hotkeys get swallowed by the capture). The right-click context menu broke that assumption; the key path was fixed for it, the paste path wasn't.

## Fix

Gate the live-send branch of `handle_paste` on `!self.has_non_live_send_overlay()`, mirroring `handle_key`, so pastes fall through to the rename / worktree-name / send-message / new-session dialog branches when an overlay is open.

dongwook-chan:
## Description

Closes #3059

While attached to a session in live-send (preview) mode, the sidebar still accepts right-clicks, so the context menu can open a Rename / New Session dialog on top of live-send. The dialog looked focused and accepted typed keys, but a clipboard paste streamed into the agent's tmux pane behind it.

`handle_key` already suspends live-send capture when a non-live-send overlay is open (`self.live_send.is_some() && !self.has_non_live_send_overlay()`), but `handle_paste` kept the older unconditional "live-send wins above every dialog" rule, written before any click-to-open overlay path existed. This PR gates the live-send branch of `handle_paste` on the same `has_non_live_send_overlay()` check, so keys and pastes route identically: to the overlay when one is open, straight to the pane otherwise.

Also adds a `#[cfg(test)] title_value()` accessor on `RenameDialog` (mirroring `NewSessionDialog::path_value()`) and two regression tests:
- a rename dialog opened over live-send receives the paste,
- with no overlay, a paste in pure live mode is still consumed by the live-send branch (priority pin).

The mosh paste-burst fallback needs no change: `wants_paste_burst()` already returns true for both live-send and the text-input dialogs, and the accumulated burst funnels through the fixed `handle_paste`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt --check`, `cargo test --lib tui::home::tests::live_send_mode` (35 passed), and `cargo clippy --all-targets` (no new warnings; the one `items_after_test_module` warning is pre-existing on main in `src/tui/home/mod.rs`) were run in a `rust:latest` container.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the routing decision is pure view-state logic with no tmux round-trip; it's pinned by two in-module regression tests in `src/tui/home/tests.rs` (dialog-over-live-send and pure-live-mode priority), which exercise the exact `handle_paste` entry point the terminal event loop calls.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Claude Fable 5)

Diagnosis, fix, tests, and this PR were produced with Claude Code under human direction and review.

## Summary

- Fixed how clipboard pastes are routed during live-send: when a Rename or New Session dialog is open over the live-send view, the pasted text now goes into the focused dialog field instead of the underlying tmux pane.
- Updated live-send paste behavior to follow the same “only capture when no overlay is open” rule used for keyboard input, so pastes match what users expect when popups/overlays are on screen.
- Cleared finalized preview selection highlights at the start of paste handling to prevent stale highlight artifacts after a paste redirects.
- Added a test-only `RenameDialog::title_value()` accessor to support assertions about dialog input.
- Added regression tests covering dialog routing under live-send, priority/behavior when no overlay is open, and highlight clearing after paste.

## Benefits

Pastes now behave consistently and predictably—clipboard input lands in the dialog the user is interacting with (not the background live-send target), and UI highlight state stays correct after routing.

### DIFF
--- a/src/tui/dialogs/rename.rs
+++ b/src/tui/dialogs/rename.rs
@@ -76,6 +76,11 @@ impl RenameDialog {
         self.mode
     }
 
+    #[cfg(test)]
+    pub fn title_value(&self) -> &str {
+        self.new_title.value()
+    }
+
     pub fn new(
         current_title: &str,
         current_group: &str,
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -5661,26 +5661,35 @@ impl HomeView {
 
     /// Route a bracketed paste event to the active text input dialog.
     ///
-    /// Live-send mode wins above every dialog: a paste while the user is
-    /// "attached" should stream straight to the agent's pane, not buffer
-    /// in a dialog the user isn't even looking at. Text-input dialogs
-    /// (rename / send_message / new) come next so multi-line dictation
-    /// lands in whichever dialog the user is actively typing into. The
-    /// settings view is checked last; its paste handler strips newlines,
-    /// which would destroy multi-line dictation if we checked it first.
+    /// Live-send mode wins over every dialog as long as no overlay is
+    /// stacked on top of it: a paste while the user is "attached" should
+    /// stream straight to the agent's pane, not buffer in a dialog the
+    /// user isn't even looking at. But once an overlay HAS been opened
+    /// on top of live-send (via a right-click context menu, the
+    /// empty-sidebar click, or any future click-to-open path), the paste
+    /// must go to that overlay, exactly like the key path in
+    /// `handle_key` — otherwise the user sees a focused input field
+    /// while their clipboard silently lands in the pane behind it.
+    /// Text-input dialogs (rename / send_message / new) come next so
+    /// multi-line dictation lands in whichever dialog the user is
+    /// actively typing into. The settings view is checked last; its
+    /// paste handler strips newlines, which would destroy multi-line
+    /// dictation if we checked it first.
     pub fn handle_paste(&mut self, text: &str) {
-        if let Some(state) = self.live_send.clone() {
-            // Mirror the live-send key path: any interaction dismisses
-            // the finalized highlight so it doesn't follow agent output
-            // through subsequent renders.
-            self.clear_preview_selection();
-            if let Some(worker) = &self.live_send_worker {
-                for key in split_paste_for_live_send(text) {
-                    worker.send(key);
+        if !self.has_non_live_send_overlay() {
+            if let Some(state) = self.live_send.clone() {
+                // Mirror the live-send key path: any interaction dismisses
+                // the finalized highlight so it doesn't follow agent output
+                // through subsequent renders.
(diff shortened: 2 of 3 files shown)


njbrake (approved):

Approving. The core fix correctly gates the live-send paste branch on `has_non_live_send_overlay()`, mirroring the `handle_key` guard, so a Rename or New Session dialog opened over live-send now receives the paste instead of the agent's pane. Verified locally: `cargo test --lib tui::home::tests::live_send_mode` passes (35/35) and `cargo fmt --check` is clean.

I pushed one follow-up commit onto this branch. `handle_key` drops a finalized preview-pane highlight unconditionally at the top, but after this change `handle_paste` cleared it only when streaming to the pane, so a mouse driven drag select, right click, paste into dialog sequence stranded the highlight. The follow-up hoists that clear to the top of `handle_paste` and pins it with a regression test.

One optional, non blocking follow-up remains: with a non text overlay (context menu, confirm, help) stacked over live-send, a paste falls through to the compose fallback and can open a SendMessageDialog behind the overlay instead of buffering into pending_paste. It self heals and no text is lost, so it can be addressed separately.

PR #3061, opened 2026-07-22
fix(acp): show real rate-limit reset time and continue the turn on resume

### LINKED ISSUE #3028: Rate limit reset datetime parsed wrong
I just ran into my rate limit and aoe resolved a wrong datetime;

Additionally, after pressing `resume now` after the real reset time (which worked) aoe did not send a continuation message

Seluj78:

## Description

Fixes both halves of #3028.

**Wrong reset time.** claude-agent-acp's rate-limit JSON-RPC error carries only `{ "errorKind": "rate_limit" }` with no reset timestamp, so `classify_rate_limit_error` always fell back to `now + 1h`. A user in Europe/Berlin therefore saw a reset roughly an hour past the real one. The authoritative reset (`resetsAt`, a unix epoch) is forwarded out-of-band on a `usage_update`'s `_meta._claude/rateLimit`; the message text is only a locale rendering of that same value. We now capture that epoch into a per-connection atomic (cleared at each prompt start so a stale window cannot leak) and prefer it, when still in the future, over the fallback at both rate-limit classify sites. A millisecond-scale epoch is normalized back to seconds. No new dependency and no locale/timezone parsing.

**Resume did not continue.** Resuming a rate-limited session (manual "Resume now" or opt-in auto-resume) respawned the worker and restored context via `session/load` but never re-issued the interrupted prompt, so the agent sat idle. A new `EventStore::rate_limited_turn_prompt` recovers the prompt whose turn ended in `Stopped{rate_limited}`, and both resume sites enqueue it as the session's pending initial turn so the existing drain-after-live plumbing delivers it once the worker is running. It is a no-op when the last turn completed normally or was agent-initiated.

Closes #3028

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Rate-limit a structured-view turn (or use the fake ACP agent). Confirm the recovery banner shows the real reset time from the provider, not a time roughly one hour from now.
- [ ] While parked, click "Resume now". Confirm the agent resumes and continues the interrupted prompt instead of sitting idle.
- [ ] With `acp.rate_limit_auto_resume` enabled, let the reset window elapse. Confirm the auto-resume also re-issues the interrupted prompt.
- [ ] Rate-limit, resume, and verify a turn that had already completed normally before the limit is NOT re-sent.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (epoch capture over message parsing, reusing the pending-initial-turn drain for continuation) and reviewed each commit.

## Summary

- Rate-limit handling now uses the provider’s authoritative reset time from the `usage_update` metadata (instead of an estimated fallback), and safely ignores stale/invalid values.
- When a conversation is interrupted by a rate limit, “Resume now” now truly continues the interrupted turn by re-sending the original prompt (including attachment references).
- Auto-resume behavior was updated to queue the interrupted prompt so the resumed worker continues immediately rather than idling.
- If a newer user message arrives, any older queued rate-limit continuation is cleared to prevent replaying the wrong prompt.
- Added durable session support for queued pending-turn continuations (including attachment references) so behavior remains consistent across worker respawns/restarts.
- Expanded test coverage: added unit tests for reset-time priority and interrupted-prompt reconstruction, plus a Playwright story with a stateful fake ACP agent that persists its progress across restarts.

## Benefits

Users see the correct “rate limit reset” time and can reliably resume after rate limiting—manually or automatically—without losing or mis-replaying the interrupted work, even across restarts.

### DIFF
--- a/src/acp/event_store.rs
+++ b/src/acp/event_store.rs
@@ -1139,6 +1139,63 @@ impl EventStore {
         }
     }
 
+    /// Text of the user prompt whose turn was interrupted by a rate limit,
+    /// i.e. the latest `UserPromptSent` whose next terminal event is a
+    /// `Stopped{reason:"rate_limited"}`. Used on resume to re-issue the
+    /// interrupted prompt so the agent continues instead of sitting idle
+    /// (#3028). Returns `None` when the last turn completed normally, was
+    /// agent-initiated (its last user prompt has a non-rate-limit terminal
+    /// after it), or produced no prompt. `has_in_flight_turn` can't answer
+    /// this: the rate-limit park emits a `Stopped`, so the turn no longer
+    /// reads as in-flight; here we specifically match the rate-limit stop.
+    pub fn rate_limited_turn_prompt(&self, session_id: &str) -> Option<String> {
+        let conn = match self.conn.lock() {
+            Ok(g) => g,
+            Err(p) => p.into_inner(),
+        };
+        let prompt: Option<(i64, String)> = conn
+            .query_row(
+                "SELECT seq, event_json FROM acp_events
+                 WHERE session_id = ?1
+                   AND json_extract(event_json, '$.UserPromptSent') IS NOT NULL
+                 ORDER BY seq DESC
+                 LIMIT 1",
+                params![session_id],
+                |row| Ok((row.get::<_, i64>(0)?, row.get::<_, String>(1)?)),
+            )
+            .optional()
+            .unwrap_or_else(|e| {
+                warn!(target: "acp.event_store", "rate_limited_turn_prompt prompt query {session_id}: {e}");
+                None
+            });
+        let (prompt_seq, prompt_json) = prompt?;
+        let text = match serde_json::from_str::<Event>(&prompt_json).ok()? {
+            Event::UserPromptSent { text, .. } => text,
+            _ => return None,
+        };
+        let terminator: Option<String> = conn
+            .query_row(
+                "SELECT event_json FROM acp_events
+                 WHERE session_id = ?1
+                   AND seq > ?2
+                   AND (json_extract(event_json, '$.Stopped') IS NOT NULL
+                     OR json_extract(event_json, '$.AgentStartupError') IS NOT NULL)
+                 ORDER BY seq ASC
+                 LIMIT 1",
+                params![session_id, prompt_seq],
+                |row| row.get::<_, String>(0),
+            )
+            .optional()
+            .unwrap_or_else(|e| {
+                warn!(target: "acp.event_store", "rate_limited_turn_prompt terminator query {session_id}: {e}");
+                None
+            });
+        match terminator.and_then(|s| serde_json::from_str::<Event>(&s).ok()) {
+            Some(Event::Stopped { reason }) if reason == "rate_limited" => Some(text),
+            _ => None,
+        }
+    }
+
     /// The first turn's transcript for smart rename: the earliest
     /// `UserPromptSent` text plus the agent's prose (`AgentMessageChunk`)
     /// emitted before the first `Stopped` that follows it (any reason, so an
@@ -2836,6 +2893,105 @@ mod tests {
             )
             .unwrap();
         assert!(!store.has_in_flight_turn("s-1"));
+    }
+
+    // #3028: the last user turn was rate-limited, so its prompt is recoverable
+    // for a resume continuation.
+    #[test]
+    fn rate_limited_turn_prompt_returns_interrupted_prompt() {
+        let (_tmp, store) = open_store(1000);
+        store
+            .record(
+                "s-1",
+                1,
+                &Event::UserPromptSent {
+                    text: "keep working".into(),
+                    attachments: Vec::new(),
+                },
+            )
+            .unwrap();
+        store
+            .record(
+                "s-1",
+                2,
+                &Event::Stopped {
+                    reason: "rate_limited".into(),
+                },
+            )
+            .unwrap();
+        assert_eq!(
+            store.rate_limited_turn_prompt("s-1"),
+            Some("keep working".to_string())
+        );
+    }
+
+    // A turn that ended normally must not be re-issued on resume.
+    #[test]
+    fn rate_limited_turn_prompt_none_when_completed_normally() {
+        let (_tmp, store) = open_store(1000);
+        store
+            .record(
+                "s-1",
+                1,
+                &Event::UserPromptSent {
+                    text: "go".into(),
+                    attachments: Vec::new(),
+                },
+            )
+            .unwrap();
+        store
+            .record(
+                "s-1",
+                2,
+                &Event::Stopped {
+                    reason: "prompt_complete".into(),
+                },
+            )
+            .unwrap();
+        assert_eq!(store.rate_limited_turn_prompt("s-1"), None);
+    }
+
+    // An agent-initiated turn hitting the limit leaves an old user prompt whose
+    // terminal is a normal Stopped, so we must not re-send it.
+    #[test]
+    fn rate_limited_turn_prompt_none_for_earlier_completed_user_prompt() {
+        let (_tmp, store) = open_store(1000);
+        store
+            .record(
+                "s-1",
+                1,
+                &Event::UserPromptSent {
+                    text: "old prompt".into(),
+                    attachments: Vec::new(),
+                },
+            )
+            .unwrap();
+        store
+            .record(
+                "s-1",
+                2,
+                &Event::Stopped {
+                    reason: "prompt_complete".into(),
+                },
+            )
+            .unwrap();
+        // Later agent-initiated turn rate-limits, but no new UserPromptSent.
+        store
+            .record(
+                "s-1",
+                3,
+                &Event::Stopped {
+                    reason: "rate_limited".into(),
+                },
+            )
+            .unwrap();
+        assert_eq!(store.rate_limited_turn_prompt("s-1"), None);
+    }
+
+    #[test]
+    fn rate_limited_turn_prompt_none_on_empty_store() {
+        let (_tmp, store) = open_store(1000);
+        assert_eq!(store.rate_limited_turn_prompt("s-1"), None);
     }
 
     #[test]
--- a/src/acp/event_store.rs
+++ b/src/acp/event_store.rs
@@ -1139,16 +1139,20 @@ impl EventStore {
         }
     }
 
-    /// Text of the user prompt whose turn was interrupted by a rate limit,
-    /// i.e. the latest `UserPromptSent` whose next terminal event is a
-    /// `Stopped{reason:"rate_limited"}`. Used on resume to re-issue the
-    /// interrupted prompt so the agent continues instead of sitting idle
-    /// (#3028). Returns `None` when the last turn completed normally, was
-    /// agent-initiated (its last user prompt has a non-rate-limit terminal
-    /// after it), or produced no prompt. `has_in_flight_turn` can't answer
-    /// this: the rate-limit park emits a `Stopped`, so the turn no longer
-    /// reads as in-flight; here we specifically match the rate-limit stop.
-    pub fn rate_limited_turn_prompt(&self, session_id: &str) -> Option<String> {
+    /// The user prompt whose turn was interrupted by a rate limit: the text
+    /// plus its attachment refs, from the latest `UserPromptSent` whose next
+    /// terminal event is a `Stopped{reason:"rate_limited"}`. Used on resume to
+    /// re-issue the interrupted prompt (with its images/files) so the agent
+    /// continues instead of sitting idle (#3028). Returns `None` when the last
+    /// turn completed normally, was agent-initiated (its last user prompt has a
+    /// non-rate-limit terminal after it), or produced no prompt.
+    /// `has_in_flight_turn` can't answer this: the rate-limit park emits a
+    /// `Stopped`, so the turn no longer reads as in-flight; here we specifically
+    /// match the rate-limit stop.
+    pub fn rate_limited_turn_prompt(
+        &self,
+        session_id: &str,
+    ) -> Option<(String, Vec<crate::acp::state::PromptAttachmentRef>)> {
         let conn = match self.conn.lock() {
             Ok(g) => g,
             Err(p) => p.into_inner(),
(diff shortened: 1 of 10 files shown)


src/acp/event_store.rs
```diff
@@ -1139,6 +1139,63 @@ impl EventStore {
...
+            .unwrap_or_else(|e| {
+                warn!(target: "acp.event_store", "rate_limited_turn_prompt terminator query {session_id}: {e}");
+                None
+            });
+        match terminator.and_then(|s| serde_json::from_str::<Event>(&s).ok()) {
+            Some(Event::Stopped { reason }) if reason == "rate_limited" => Some(text),
+            _ => None,
+        }
+    }
+
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3061/commits/8128f26ecd5ae824139e5e5ff30021b050a5cfc8. `rate_limited_turn_prompt` now returns the interrupted prompt's attachment refs alongside its text; the pending-turn queue carries them on the instance, and the drain reloads the blobs from the event store before re-sending, so a rate-limit resume replays images/files, not just text.

src/server/acp_reconciler.rs
```diff
@@ -1240,6 +1243,35 @@ async fn resume_one(state: Arc<AppState>, target: ResumeTarget) -> ResumeOutcome
...
+                "rate-limit continuation lookup failed: {e}"
+            );
+            return;
+        }
+    };
+    state
+        .session_service
+        .set_pending_initial_turn(id, text)
+        .await;
+}
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3061/commits/8128f26ecd5ae824139e5e5ff30021b050a5cfc8. A fresh user prompt via `acp_prompt` now clears any queued `pending_initial_turn` before forwarding, so the reconciler can no longer replay an older interrupted prompt after the newer message. The drain path never routes through `acp_prompt`, so this doesn't race the continuation delivery itself.

src/server/api/acp.rs
```diff
@@ -1179,6 +1184,12 @@ pub async fn acp_prompt(
         Ok(a) => a,
         Err((code, msg)) => return (code, msg).into_response(),
     };
+    // A fresh user prompt supersedes any queued rate-limit resume
+    // continuation, so drop it before sending: otherwise the reconciler could
+    // later replay the older interrupted prompt after this newer one (#3028).
+    // The drain path never routes through here, so this cannot race the
+    // continuation delivery itself.
+    state.session_service.clear_pending_initial_turn(&id).await;
```
jerome-benoit: [MAJOR] This clear closes the *sequential* stale-replay case, but it doesn't hold `instance_lock`. The `instance_lock` acquired at line 1095 is inside `touch_and_wake_if_sunk` and dropped when that helper returns, so at this point `acp_prompt` holds no per-session lock. Meanwhile `drain_pending_initial_turn` holds `instance_lock` across its whole snapshot -> reload -> send_turn -> clear. A fresh prompt landing while a drain is mid-await (after the drain snapshotted the old continuation) clears the field the drain already read, and the drain then sends the stale continuation *after* this newer prompt - the exact #3028 replay this guard targets. The comment above ("cannot race the continuation delivery itself") is inaccurate: it addresses handler re-entrancy, not a concurrent drain. Fix: wrap this clear + the `send_turn` below in `state.instance_lock(&id)` (the same lock the drain holds), and correct the comment.

web/tests/live/acp-stories/rate-limit-resume-continuation.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+import { test as base, expect } from "@playwright/test";
+import { spawnAoeServe, listSessions, seedSessionViaAoeAdd } from "../../helpers/aoeServe";
+import {
+  enableStructuredViewAndWait,
+  waitForStructuredView,
+  waitForReplayContains,
+  attachServeDiagnostics,
+} from "../../helpers/acp";
+
+const RESETS_AT = new Date(Date.now() + 60 * 60 * 1000).toISOString();
```
jerome-benoit: [MAJOR, tests] `RESETS_AT` is set to `now + 60min` - the very `now+1h` fallback this PR exists to replace - and the spec never asserts the displayed reset time (only the "Rate-limited" banner + the continuation marker). So the reset-time half (the PR title's first claim) has zero e2e teeth: a regression to the old guess would still pass here. Consider asserting the rendered reset time against a distinctive provider value, and driving it through a `usage_update._meta._claude/rateLimit.resetsAt` rather than the error `data.resets_at` branch.

jerome-benoit (changes requested):
## Review: show real rate-limit reset time and continue the turn on resume

I verified this thoroughly, including building and running the relevant Rust suites in an isolated checkout. The core logic is correct and the unit coverage is genuinely strong, but two items should be addressed before merge - one is a real concurrency gap in the stale-replay fix, the other is a test-adequacy gap on the headline half.

### What's solid (verified)
- **Reset-time classification** is correct: the `resetsAt` epoch is captured from `usage_update._meta._claude/rateLimit`, normalized to seconds at the write site, cleared before each new turn, and preferred over the `now+1h` fallback only when strictly in the future, at both classify sites. Unit tests cover future-preferred, past-ignored, `data.resets_at` (RFC3339) beats captured, and ms->s normalization - all pass locally (8/0).
- **Resume recovery query** (`rate_limited_turn_prompt`) is correct: anchoring on the latest `UserPromptSent` and its *first* terminator, and returning `Some` only for `Stopped{rate_limited}`, correctly no-ops for normally-completed, agent-initiated-earlier, and empty-store cases (all tested, pass, 33/0).
- **CodeRabbit MAJOR #1 (attachments)** is addressed: refs ride the pending-turn flow, blobs are reloaded lazily and re-persisted idempotently (`INSERT OR IGNORE` on `(session_id, attachment_id)`), so a rate-limited turn with attachments replays without dropping or duplicating them (`rate_limited_turn_prompt_preserves_attachments` passes).
- No Phase-A/B double-fire: the parked turn reads `in_flight_turn: false`, so the continuation is a fresh prompt owned by the new connection's own future, not a control-channel adoption. CI is fully green with all heavy suites run.

### MAJOR - the stale-replay fix (CodeRabbit #2) has a concurrency gap
The fresh-prompt guard at `api/acp.rs:1192` clears `pending_initial_turn` before `send_turn`, which closes the *sequential* case. But it does not hold `instance_lock` (the `instance_lock` at `acp.rs:1095` lives inside `touch_and_wake_if_sunk` and is deliberately dropped when that helper returns), whereas `drain_pending_initial_turn` holds `instance_lock` across its whole snapshot -> reload -> `send_turn` -> clear. Since the two paths don't share the lock, a fresh user prompt landing while a drain is mid-`await` (after the drain snapshotted the old continuation) will clear the field the drain already read, and the drain then sends the stale continuation - delivering the older interrupted prompt after the newer one, which is exactly the #3028 replay the guard is meant to prevent. The in-code comment "The drain path never routes through here, so this cannot race the continuation delivery itself" is inaccurate: it reasons about handler re-entrancy, not a concurrent drain. Narrow window (the agent serializes prompts), but real and reachable on a resume immediately followed by a user prompt. Fix: wrap `acp_prompt`'s `clear_pending_initial_turn` + `send_turn` in `state.instance_lock(&id)` (the same lock the drain holds), and correct the comment.

### MAJOR (tests) - the headline reset-time half has no end-to-end coverage
The e2e (`web/tests/live/acp-stories/rate-limit-resume-continuation.spec.ts`) sets `RESETS_AT = now + 60min` (i.e. the very `now+1h` fallback the PR replaces) and never asserts the displayed reset time - so a regression of the reset-time half would still pass. Worse, the fake agent (`web/tests/helpers/fakeAcpAgent.mjs`) injects `resets_at` into the JSON-RPC error `data` (the RFC3339 branch), never into the out-of-band `usage_update._meta._claude/rateLimit.resetsAt` - so the actual #3028 mechanism (the captured-epoch atomic plus its capture/clear sites) is exercised only by Rust units, never through a live agent. Please either extend `fakeAcpAgent.mjs` to emit a `usage_update` with `_meta._claude/rateLimit.resetsAt` and assert the rendered reset time in the spec, or explicitly note that half (a) is unit-tested only.

### MINOR / NIT
- `rate_limited_turn_prompt` matches only `$.UserPromptSent`, so a `UserDiffCommentsPrompt` turn parked on `rate_limited` is not recoverable (agent stays idle for that submission type). Safe (idle, not wrong-replay), but the doc comment overstates coverage.
- The ms/s normalization threshold (`raw > 1e11`) is one-sided: a sub-1e11 milliseconds value would be misread as a far-future seconds epoch, and the `> now` filter doesn't catch a future misread. Not reachable with real near-future data; a lower-bound note would make the domain explicit.
- Auto-resume continuation and the stale-continuation-drop have no dedicated test.
- `load_attachment` returning `None` silently drops that blob via `filter_map`; a `warn!` would aid diagnosis.

Great work on the recovery query and the attachment replay - the unit coverage of the edge cases is exemplary. Happy to re-approve once the `instance_lock` gap is closed (with the comment corrected) and half (a) has some e2e teeth (or an explicit note).

src/server/api/acp.rs
Seluj78: Good catch. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3061/commits/22acf3b888339812cf9ef7ff8b612c169d7baf81. `acp_prompt` now holds the per-session `instance_lock` across the clear + `send_turn`, the same lock the drain holds across its whole snapshot -> reload -> send -> clear, so a drain mid-await can no longer deliver the stale continuation after the newer prompt. `send_turn` doesn't take that lock (the drain already calls it while holding it), so no deadlock.

web/tests/live/acp-stories/rate-limit-resume-continuation.spec.ts
Seluj78: Agreed, no teeth as written. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3061/commits/bb39839bcfbfcd36e6c2521f2691b55d474953ec. The reset now rides only on `usage_update._meta._claude/rateLimit.resetsAt` (the error omits `resets_at`), driving the capture path, and the spec asserts the banner renders that distinctive value (computed in-browser via the same `toLocaleTimeString`), so a regression to the now+1h guess fails.

src/server/api/acp.rs
```diff
@@ -1179,18 +1184,31 @@ pub async fn acp_prompt(
...
+    // pending-turn drain holds that same lock across its whole snapshot ->
+    // reload -> send -> clear; without it a drain mid-await could deliver the
+    // stale continuation *after* this newer prompt. `send_turn` does not take
+    // `instance_lock` (the drain calls it while holding the lock), so this
+    // cannot deadlock. Resume + publish + forward all live in the shared
+    // service so the plugin host delivers turns through the same path (#2897).
+    let outcome = {
+        let inst_lock = state.instance_lock(&id).await;
+        let _serialized = inst_lock.lock().await;
+        state.session_service.clear_pending_initial_turn(&id).await;
```
jerome-benoit, line 1200: F1 confirmed closed here: the clear + `send_turn` now run under `state.instance_lock(&id)`, the same per-session lock `drain_pending_initial_turn` holds across its whole critical section (both handles share one `instance_locks` map). Atomic w.r.t. a concurrent drain, so the mid-await stale-replay window is gone. Verified no re-entrant acquisition in `send_turn`/`clear_pending_initial_turn`, so no self-deadlock, and no reverse lock ordering anywhere, so no ABBA.

web/tests/live/acp-stories/rate-limit-resume-continuation.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // The turn parks on the rate limit.
+    await expect(page.getByText(/Rate-limited/i)).toBeVisible({ timeout: 15_000 });
+
+    // The banner must show the real reset from the usage_update meta, rendered
+    // the same way the UI does (`new Date(resets_at).toLocaleTimeString()`),
+    // computed in-browser so locale/timezone match. A regression to the old
+    // `now + 1h` guess would render a different time and fail this.
+    const expectedReset: string = await page.evaluate((iso) => new Date(iso).toLocaleTimeString(), RESET_ISO);
+    await expect(page.getByText(`resets at ${expectedReset}`)).toBeVisible({ timeout: 15_000 });
```
jerome-benoit, line 104: F2 confirmed: this asserts the rendered reset time against `now + 2h37m`, which is driven exclusively through `usage_update._meta._claude/rateLimit.resetsAt` (the error now omits `resets_at`). It cannot alias the `now + 1h` fallback, so a regression of the reset-time half would fail here. Real teeth.

jerome-benoit (approved):
## Re-review: rate-limit reset time + resume continuation — LGTM

Both findings from my previous review are fully addressed. I verified this the same way as last time: three independent read-only reviewers plus my own build and test run in an isolated checkout, followed by a dedicated adversarial pass that tried hard to break the APPROVE (lock-ordering deadlock, livelock, e2e triviality, incomplete fix). It held.

### F1 (stale-replay race) — RESOLVED
`22acf3b` wraps the `clear_pending_initial_turn` + `send_turn` in the per-session `instance_lock` (`api/acp.rs:1197-1211`), the same lock `drain_pending_initial_turn` holds across its whole snapshot -> reload -> send -> clear. Confirmed both handles resolve to one shared map (a single `instance_locks` is `Arc::clone`d into `SessionService` and `AppState`), so the clear+send is now atomic w.r.t. a concurrent drain, closing the mid-`await` window. No deadlock: neither `send_turn` (documents "deliberately no instance_lock here") nor `clear_pending_initial_turn` re-acquires `instance_lock`. I mapped every lock reachable under the new critical section: the `instance_lock`/`instances` domain and the supervisor `workers` domain are disjoint, and the acquisition order (`instance_lock` first, then `instances` transiently) is uniform across all call sites, so there is no ABBA reversal. The wider hold across `send_turn` merely matches the envelope the drain already imposed, so it introduces no new stall. The misleading comment was corrected too.

### F2 (reset-time had no e2e) — RESOLVED
`bb39839` removes `resets_at` from the error `data` and drives the reset time exclusively through `usage_update._meta._claude/rateLimit.resetsAt` (the real #3028 mechanism), then asserts the rendered time against a distinctive `now + 2h37m` that cannot alias the `now + 1h` fallback. Verified the fake agent forwards that meta verbatim while the error genuinely omits `resets_at`, so the assertion is gated on the captured-epoch path and would fail on a regression to the guess. No subsecond/timezone/collision flake.

### My own verification
- Build clean.
- `cargo test --features serve --lib` for the rate-limit surface: 14/14 pass (9 classify + meta, 5 `rate_limited_turn_prompt`), no hang (a deadlock in the lock fix would have hung these).
- `acp_client.rs` and `event_store.rs` are byte-identical to the previously-reviewed commit; no regression.

Nice work closing both cleanly. Minor, non-blocking: the ms/s normalization threshold is one-sided (a sub-1e11 ms value would misread), harmless with real near-future data; a lower-bound note would make the domain explicit.

PR #3064, opened 2026-07-22
feat: pin favorites to the top in every sort order

bjkim95:
## Description

Favorites already exist, but the star only does anything in the `Attention` sort: it biases
`attention_session_key` / `attention_group_key` within a status tier, the `f` binding is gated on
`Context::AttentionSort`, and the `* ` prefix is gated on `in_attention`. In every other sort order
the flag persists silently, the key is inert, and nothing renders.

That leaves no way to keep a session where you can find it in the sort orders people actually leave
the list in. Attention is a triage view whose order moves on its own as statuses change, and
favorite sits third in its key (behind urgent and the status tier), so it does not pin to the top
there either.

This makes a favorited session, and a group holding one, pin to the top of its sibling scope in
every sort order, behind a new `session.favorites_first` toggle that defaults to on.

No new persisted state: this reuses `favorited_at`, so there is no format change and no migration.
The mechanism is a second stable pass. `sort_by_key` is stable, so each mode sorts exactly as before
and then a `!is_live_favorite` pass partitions favorites to the front, preserving the mode's own
ordering inside both partitions. That is equivalent to prepending the bool to each mode's key tuple,
without touching five key types.

**`Attention` is deliberately untouched.** Both sort helpers return from the Attention arm before the
favorites pass, and `attention_session_key` is unmodified, so a favorited `Running` row still never
leaps above a plain `Waiting` one. A test asserts the flag makes no difference there.

**Snooze outranks the star.** `snooze()` leaves `favorited_at` set (only `archive()` clears it), so
"snoozed favorite" is a reachable state. Sorting, the key gate, and the decoration all go through one
`is_live_favorite` predicate, so a snoozed favorite is not pinned and does not render a star. Without
that shared predicate the row would have worn a star while sorting like an ordinary row.

Commits are split so each is reviewable on its own: a pure refactor, the config field, session
sorting, group sorting, the keybinding, the decoration, then a fix plus copy refresh.

### Decisions worth a reviewer's opinion

- **Default on.** A user who has never favorited anything sees zero change (every row's bool is
  equal, so the pass is a no-op), and a user who has favorited something has already expressed the
  intent. Happy to flip it to opt-in if you would rather ship it dark.
- **TUI only.** The web sidebar keeps favorite as a within-tier Attention signal and has its own Pin
  control, so this does not touch web ordering. The setting is still web-writable, matching
  `strict_hotkeys` / `click_action` / `confirm_before_quit` which are also TUI-only in effect; its
  description says it governs the TUI list so the dashboard toggle is not silently inert.
- **Groups pin without a marker.** A group can float up because a nested member is favorited, with no
  glyph on the group row explaining why. Decoration stayed scoped to sessions; say the word if you
  want a marker on the header too.
- **Perf shape.** The new group pass uses `sort_by_key`, which re-runs the key on every comparison,
  and `has_live_favorite` allocates a `format!("{}/")` and walks the instance slice. That matches the
  existing `max_created_at_in_group` keys around it rather than introducing a new pattern, but it does
  add one more such pass. `sort_by_cached_key` would be a strict improvement for all of them if you
  want that cleaned up here.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

On the last box: I verified the TUI by driving it under tmux and capturing the pane rather than
recording, so I have transcripts instead of a gif. Screens below if useful.

`cargo test --lib`: 4248 passed, up from 4233 on `main`. The 2 failures in my run
(`git::worktree::tests::delete_branch_reaps_lingering_locked_worktree`,
`session::trash::tests::purge_recovers_when_project_path_diverged_and_locked_entry_survives`) also
fail on unmodified `main` here, so they look environmental to my box rather than caused by this
branch. `cargo clippy -- -D warnings` and `cargo fmt -- --check` are clean.

Documentation: `src/cli/session.rs` help for `session favorite` was stale (it still said
Attention-only), so `docs/cli/reference.md` is regenerated via `cargo xtask gen-docs`. The
`tui-triage` tip in `src/tips.rs` and the Pin bullet in `docs/guides/web/dashboard.md` said the same
stale thing and are updated.

Manual TUI check, isolated `HOME`, debug build:

| Step | Result |
|---|---|
| Newest sort, `f` on a session in group `old` | row became `* alpha` and jumped to the top |
| same | its group `old` pinned above sibling group `new` |
| switch to Attention | favorite pins within its Idle tier, order otherwise unchanged |
| `favorites_first = false` written to config | pin and star dropped at runtime, no restart; `f` went inert |
| snooze the favorite | `z alpha 59m`, sunk in Attention; back in Newest: no star, no pin |
| restart | star and pin persisted |

## Test Coverage Analysis

The only web-visible effect is a new schema-derived toggle rendered by the existing generic
`SchemaSection`. No new component, no changed dashboard flow, and web ordering is unchanged, so
`coverage-matrix.json` has nothing new to point at. Tell me if you would rather have an entry anyway.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is covered by 15 unit tests at the layer where it lives,
  and I verified the integrated path by hand under tmux (table above) rather than encoding it. The
  six manual steps map cleanly onto a `TuiTestHarness` spec, so I am happy to add one if you want it
  in this PR rather than as follow-up.

New tests: pinning in Newest and A-Z; order preserved among favorites; snoozed favorite not pinned;
Attention output identical with the flag on and off; group pinning; archived member not pinning its
group; root-group pinning in the all-profiles view; the key gate on and off the flag; the flag's
round trip; and the decoration on both sides of the flag. Two existing tests pinned the old
"favorite is Attention-only" contract: each was re-scoped to the flag-off case where that contract
still holds, with the flag-on case moved to a new test, and the bindings one additionally gained an
assertion that snooze stays gated.

One thing worth calling out because it is easy to miss when reviewing: `flatten_tree_all_profiles`
orders root groups with its own inline `match` instead of calling the shared helper, because each
root carries per-profile instances rather than sharing one slice. My first pass missed it, so the
feature silently did nothing there and behavior depended on how many profiles you had. That path had
no ordering test at all; there is one now.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 4.8 via Claude Code.

The implementation, tests, and this description were written by the agent, working from a design doc
and a task plan reviewed by me before any code was written. Every commit was verified against the
suite, and a separate reviewer pass caught the `flatten_tree_all_profiles` gap plus three stale
strings before this was opened.

## Summary

- Added a new default-on `session.favorites_first` toggle to prioritize “live-favorited” sessions by pinning them to the top across standard TUI session sort orders (Newest, Oldest, Last Activity, and alphabetical), and to pin groups that contain live favorites (including all-profile root groups).
- Defined “live favorites” so snoozed, archived, and trashed favorites are not pinned and do not show the star/pin decoration; snoozing temporarily removes the pin/star until it wakes.
- Kept existing Attention sorting behavior intact, while aligning the pin/star/keybinding behavior with the new toggle: when `favorites_first` is off, favorites are only prioritized within the Attention tier; when on, they’re prioritized across other sorts too.
- Made the behavior runtime-configurable by introducing a process-wide gate that is refreshed on startup and config reload, so UI ordering, decoration, and keybinding/help text update immediately.
- Updated UX copy across the CLI, TUI help/shortcuts, dashboard wording, and tips so “favorite” behavior always matches the setting (including conditional “Attention” labeling).
- Added unit tests covering session/group ordering, live-favorite rules (including nested groups), config/runtime gating, keybinding and decoration eligibility, snoozed favorites, Attention interactions, and all-profile root group handling.
- Reused existing favorite timestamps (`favorited_at`) instead of adding migrations.

## Benefits

- Users get consistent, predictable visibility for favorites (“favorites first”) across most views and sort modes without changing the meaning of Attention results.
- The star/pin visuals and the keybinding hints stay accurate because they’re gated by the same runtime setting and live-favorite rules.

## Potential areas for further enhancement

- Add broader end-to-end/UI tests to confirm pin/star/keybinding consistency across all affected views when toggling `favorites_first` at runtime (not just unit-level coverage).

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -367,7 +367,7 @@ Manage session lifecycle (start, stop, attach, etc.)
 * `set-base` — Set or clear the per-session diff base branch. The diff view compares the worktree against this ref instead of the auto-detected default. Useful when the PR target differs from the project default (stacked PRs, hotfix off `release/*`, renamed default branch). See #970
 * `snooze` — Snooze a session for a duration (temporary archive, auto wakes)
 * `unsnooze` — Wake a snoozed session immediately
-* `favorite` — Mark a session as a favorite. With `session.favorites_first` on (the default), favorited rows pin to the top of their sibling scope in every sort order; with it off, they pin within their status tier in the Attention sort only. Either way the row renders with a leading `* ` glyph plus bold + underline wherever the pin applies. Snoozing a favorite suspends the pin until it wakes
+* `favorite` — Mark a session as a favorite. With `session.favorites_first` on (the default), favorited rows pin to the top of their sibling scope in every sort order; with it off, they pin within their status tier in the Attention sort only. Either way the row renders with a leading `*` marker plus bold and underline wherever the pin applies. Snoozing a favorite suspends the pin until it wakes
 * `unfavorite` — Clear the favorite flag on a session
 * `color` — Set (or clear) a per-session color label, rendered as a colored dot in the web sidebar for at-a-glance status signaling. Intended for a running agent to flag its own state, e.g. `aoe session color $(aoe session current -q) red`. Colors: `red` (needs attention), `amber` (working), `green` (done); `none` clears it
 * `archive` — Archive a session: sink it in the Attention sort and tear down its tmux sessions. Worktree, branch, container preserved. `--no-kill` skips tmux teardown. See #1868
@@ -578,7 +578,7 @@ Wake a snoozed session immediately
 
 ## `aoe session favorite`
 
-Mark a session as a favorite. With `session.favorites_first` on (the default), favorited rows pin to the top of their sibling scope in every sort order; with it off, they pin within their status tier in the Attention sort only. Either way the row renders with a leading `* ` glyph plus bold + underline wherever the pin applies. Snoozing a favorite suspends the pin until it wakes
+Mark a session as a favorite. With `session.favorites_first` on (the default), favorited rows pin to the top of their sibling scope in every sort order; with it off, they pin within their status tier in the Attention sort only. Either way the row renders with a leading `*` marker plus bold and underline wherever the pin applies. Snoozing a favorite suspends the pin until it wakes
 
 **Usage:** `aoe session favorite <IDENTIFIER>`
 
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -367,7 +367,7 @@ Manage session lifecycle (start, stop, attach, etc.)
 * `set-base` — Set or clear the per-session diff base branch. The diff view compares the worktree against this ref instead of the auto-detected default. Useful when the PR target differs from the project default (stacked PRs, hotfix off `release/*`, renamed default branch). See #970
 * `snooze` — Snooze a session for a duration (temporary archive, auto wakes)
 * `unsnooze` — Wake a snoozed session immediately
-* `favorite` — Mark a session as a favorite. Favorited rows pin to the top of their status tier in the Attention sort and render with a leading `* ` glyph plus bold + underline
+* `favorite` — Mark a session as a favorite. With `session.favorites_first` on (the default), favorited rows pin to the top of their sibling scope in every sort order; with it off, they pin within their status tier in the Attention sort only. Either way the row renders with a leading `* ` glyph plus bold + underline wherever the pin applies. Snoozing a favorite suspends the pin until it wakes
 * `unfavorite` — Clear the favorite flag on a session
 * `color` — Set (or clear) a per-session color label, rendered as a colored dot in the web sidebar for at-a-glance status signaling. Intended for a running agent to flag its own state, e.g. `aoe session color $(aoe session current -q) red`. Colors: `red` (needs attention), `amber` (working), `green` (done); `none` clears it
 * `archive` — Archive a session: sink it in the Attention sort and tear down its tmux sessions. Worktree, branch, container preserved. `--no-kill` skips tmux teardown. See #1868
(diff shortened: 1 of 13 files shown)


njbrake (approved):

Approving. This closes a real gap: the favorite key was gated to `Context::AttentionSort`, so `f` did nothing in Newest/AZ/Oldest/LastActivity and the star was inert outside the one sort view people rarely stay in. The PR opens favoriting up in every sort order behind a default-on `session.favorites_first` toggle.

What I verified locally:

- The mechanism is a stable second `sort_by_key(|i| !is_live_favorite(i))` pass, so each mode sorts as before and favorites partition to the front with their in-mode order preserved. Ran the new favorites unit tests; all pass.
- Attention sorting is provably untouched: both sort helpers `return` before the favorites pass, and `test_favorites_first_does_not_change_attention_sort` asserts byte-identical output with the flag on and off.
- The `attention_group_key` refactor to the shared `has_live_favorite` helper keeps the identical net predicate.
- `is_live_favorite` correctly excludes snoozed, archived, and trashed favorites, and that one predicate gates the key, the row decoration, and the footer hint, so all three stay in lockstep.
- `cargo fmt --check` is clean; the lone `dead_code` warning is pre-existing and untouched here.

Non-blocking notes:

- The group and root-group passes re-run `has_live_favorite` per comparison via `sort_by_key`. `sort_by_cached_key` would be a strict improvement for these and the sibling attention keys, but it matches the existing pattern, so deferring is fine.
- Default-on is a defensible behavior change: it is a no-op unless the user has already favorited something.

I pushed one trivial follow-up to the branch: an inverted comment in the footer test where "Fav included" should read "excluded".

PR #3066, opened 2026-07-22
fix(mcp): honor disabled Codex native servers

microHoffman:
## Description

Honor Codex's native `[mcp_servers.<name>] enabled` metadata when AoE builds the effective MCP server set.

Disabled definitions remain present during drift reconciliation, so disabling a server is not mistaken for removal and re-enabling it does not create a stale conflict. They are filtered only from effective CLI output and ACP forwarding. Missing or `true` values remain enabled, while malformed values keep the existing per-entry skip-and-warn behavior.

Testing performed:

- `cargo fmt --check`
- `cargo clippy --features serve --all-targets`
- `cargo test --lib codex_`
- CLI MCP list e2e regression
- ACP forwarding integration regression
- Live Playwright MCP dashboard regression
- Web formatting, linting, typechecking, and coverage-matrix validation

The full local `cargo test --features serve` run reached 5,636 passing tests with 11 environment-sensitive shell-launch failures in an unrelated module. That complete 53-test module passes in isolation with the system shell.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

OpenAI Codex

Codex implemented and verified the fix with user review of the resulting behavior and diff.

**NOTE:**

## Summary

- Fixed Codex MCP server handling to honor the native `enabled` flag when building the effective set.
- If `enabled` is omitted or set to `true`, the server is included for effective use and forwarding.
- If `enabled = false`, the server is kept for drift/change bookkeeping, but excluded from the effective set shown in CLI output and forwarded via ACP.
- Malformed `enabled` values keep the existing behavior: the entry is skipped with a warning, without breaking other valid entries.
- Added/updated coverage across parsing, drift reconciliation, CLI (`aoe mcp list`), ACP forwarding, the dashboard UI, docs, plus e2e and integration tests (including Playwright tests).

## Benefits

Ensures explicitly disabled Codex MCP servers aren’t treated as removed during reconciliation, preventing stale conflicts when re-enabled, and keeps what the user sees/what gets forwarded aligned with the intended effective configuration.

### DIFF
--- a/src/session/mcp_model.rs
+++ b/src/session/mcp_model.rs
@@ -19,7 +19,7 @@
 //! [`ProjectMcpServer::redacted_summary`], so names reach a screen, a log, or
 //! CLI output and values never do.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::io::BufReader;
 use std::path::Path;
 
@@ -407,14 +407,17 @@ fn native_config_for(agent_key: &str) -> Option<NativeMcpConfig> {
     }
 }
 
-/// A native config read: the converted servers plus the names of any entries
-/// that were skipped because they failed to convert. The skipped list gates
-/// drift detection (#1996): a native file with a malformed entry must not make
-/// the drift detector report that entry as "removed" (it is still there, just
-/// unparseable), so callers pause drift for that agent when `skipped` is
-/// non-empty.
+/// A native config read: every converted definition, the names disabled by the
+/// native agent, and the names of entries skipped because they failed to
+/// convert. Disabled definitions remain in `servers` so drift reconciliation
+/// sees them as present; only the effective-set loader filters them out. The
+/// skipped list gates drift detection (#1996): a native file with a malformed
+/// entry must not make the drift detector report that entry as "removed" (it is
+/// still there, just unparseable), so callers pause drift for that agent when
+/// `skipped` is non-empty.
 pub struct NativeRead {
     pub servers: Vec<ProjectMcpServer>,
+    pub disabled_names: BTreeSet<String>,
     pub skipped: Vec<String>,
 }
 
@@ -422,6 +425,7 @@ impl NativeRead {
     fn empty() -> Self {
         NativeRead {
             servers: Vec::new(),
+            disabled_names: BTreeSet::new(),
             skipped: Vec::new(),
         }
     }
@@ -446,10 +450,15 @@ pub fn load_native_mcp_servers_checked(agent_key: &str, home: &Path) -> Result<N
     }
 }
 
-/// Like [`load_native_mcp_servers_checked`] but discards the skipped-entry list.
-/// Used by forwarding, which only needs the parseable servers.
+/// Like [`load_native_mcp_servers_checked`] but returns only definitions enabled
+/// by the native agent. Used by effective-set resolution and forwarding.
 pub fn load_native_mcp_servers(agent_key: &str, home: &Path) -> Result<Vec<ProjectMcpServer>> {
-    Ok(load_native_mcp_servers_checked(agent_key, home)?.servers)
+    let read = load_native_mcp_servers_checked(agent_key, home)?;
+    Ok(read
+        .servers
+        .into_iter()
+        .filter(|server| !read.disabled_names.contains(&server.name))
+        .collect())
 }
 
 /// Convenience wrapper that resolves the real home dir. Kept separate from
@@ -494,6 +503,7 @@ fn convert_tolerant<R>(
     }
     NativeRead {
         servers: out,
+        disabled_names: BTreeSet::new(),
         skipped,
     }
 }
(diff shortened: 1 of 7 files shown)


src/session/mcp_model.rs
```diff
@@ -678,7 +693,15 @@ fn read_codex_toml(path: &Path) -> Result<NativeRead> {
...
         .with_context(|| format!("parsing native MCP config at {}", path.display()))?;
-    Ok(convert_tolerant(parsed.mcp_servers, path, convert_codex))
+    let disabled_names = parsed
+        .mcp_servers
+        .iter()
+        .filter(|(_, raw)| matches!(raw.enabled, Some(toml::Value::Boolean(false))))
+        .map(|(name, _)| name.clone())
+        .collect();
+    let mut read = convert_tolerant(parsed.mcp_servers, path, convert_codex);
+    read.disabled_names = disabled_names;
```
jerome-benoit, line 703: This is the crux and it's right: `disabled_names` is tracked separately (matching only `Boolean(false)`), while the server definition still flows into `read.servers` via `convert_tolerant`. Because `enabled` never enters `ProjectMcpServer`, a pure toggle produces no fingerprint diff - so enabled->disabled isn't a removal and disabled->re-enable isn't a stale conflict. Verified across a persisted-snapshot reload too.

src/session/mcp_model.rs
```diff
@@ -446,10 +450,15 @@ pub fn load_native_mcp_servers_checked(agent_key: &str, home: &Path) -> Result<N
...
-/// Used by forwarding, which only needs the parseable servers.
+/// Like [`load_native_mcp_servers_checked`] but returns only definitions enabled
+/// by the native agent. Used by effective-set resolution and forwarding.
 pub fn load_native_mcp_servers(agent_key: &str, home: &Path) -> Result<Vec<ProjectMcpServer>> {
-    Ok(load_native_mcp_servers_checked(agent_key, home)?.servers)
+    let read = load_native_mcp_servers_checked(agent_key, home)?;
+    Ok(read
+        .servers
+        .into_iter()
+        .filter(|server| !read.disabled_names.contains(&server.name))
```
jerome-benoit, line 460: Confirmed this is the single filter both effective consumers share: CLI via `resolve_effective` and ACP forwarding via `project_servers_to_acp` (same resolver). The unfiltered `_checked` loader feeds only the drift/tracking path. I enumerated every native-set consumer; none leaks a disabled server into an effective/forwarded set.

jerome-benoit (approved):
## Review: honor disabled Codex native servers — LGTM

Verified in depth: three independent read-only reviewers (parser, drift/reconciliation, tests) plus two adversarial waves that tried to break the approval, and my own build + test run in an isolated checkout. It holds.

### What's correct (verified)
- **Parser semantics.** `enabled` is deserialized as `Option<toml::Value>` (not `Option<bool>`), so missing/`true` => enabled, `false` => disabled-but-tracked, and any non-boolean bails into the pre-existing per-entry skip-and-warn path rather than erroring the whole file or silently mis-toggling. No dead field.
- **The elegant crux.** `enabled` is deliberately NOT part of `ProjectMcpServer {name, transport}`, so toggling it never changes the fingerprint. That is what makes both transitions correct at once: enabled->disabled is not seen as a removal (the def stays in `read.servers`), and disabled->re-enable produces no stale conflict.
- **Filtering is complete.** Both effective paths - CLI (`resolve_effective`) and ACP forwarding (`project_servers_to_acp` via the same resolver) - filter disabled entries through `load_native_mcp_servers`/`disabled_names`, while the drift/tracking path uses the unfiltered `_checked` loader. I enumerated every consumer of the native set (CLI, dashboard REST, plugin `mcp.list`/`resolve`, ACP): none leaks a disabled server into an effective/forwarded set.
- **Persistence + concurrency.** The on-disk drift snapshot round-trips correctly across a restart (disabled def retained, effective set still excluded), and all snapshot mutations run under the exclusive `with_locked_state` lock.
- My run: `cargo test --lib codex_` = 106/0, including `codex_enabled_toggle_preserves_drift_and_restores_effective_server`.

### Non-blocking notes
1. **Doc**: a server previously "kept" into the global `mcp.json` keeps forwarding even after you set `enabled=false` in the Codex native config, because the global layer outranks native (existing precedence). That is consistent and intentional, but a one-line note in the MCP docs ("disabling in the native config does not remove a copy you kept to global; remove it from global to stop forwarding") would prevent confusion.
2. **Test gap**: the new tests assert entry-skip for a malformed `enabled`, but none asserts the warn + whole-agent drift-pause that a non-boolean value triggers through the surface reconcile. Worth one `resolve_surface`-level assertion.

Nice, well-scoped fix with genuine e2e/ACP/live coverage. Approving.

njbrake (approved):

Approve. This is a clean, well-scoped bug fix: AoE was silently forwarding Codex MCP servers that the user had disabled with `enabled = false`, because the field was never deserialized. The fix honors it.

Verified the load-bearing fact against the real Codex source (`codex-rs/config/src/mcp_types.rs`): `enabled: bool` with a default of true and the comment "When `false`, Codex skips initializing this MCP server." So omitted/`true` stay enabled, `false` is disabled, and a non-boolean fails Codex's own parse too, which makes the skip-and-warn behavior defensible. The mapping in `convert_codex` matches exactly.

The design choice I like: disabled definitions stay in `NativeRead.servers` so drift reconciliation still sees them as present (a toggle is neither a removal nor a conflict), and only the effective-set loader filters them out. Coverage is thorough at every surface: parser unit tests, a full enable/disable/re-enable drift toggle, CLI `mcp list`, ACP forwarding, and the live dashboard panel.

One non-blocking edge case for a possible follow-up: a disabled server stays in the drift snapshot, so if it is later deleted from `config.toml` it surfaces as keep-on-removal, which could offer to promote (and forward) a server the user had disabled. It always requires an explicit user action, so nothing forwards silently; fine to leave as is.

PR #3077, opened 2026-07-23
fix(session): persist engine swap from the restart dialog

Eric162:
## Description

Switching the AI engine in the restart dialog (`e`) had no effect: the session always respawned with its original tool.

**Root cause.** The restart dialog's tool/command/args swap is applied in `restart_selected_session` to the TUI's in-memory `Instance` only. On launch, `Instance::start_with_size_opts` calls `reconcile_from_disk`, which does a wholesale `*self = disk` reload and reverts `tool`/`command`/`extra_args` to the on-disk values. Those disk values were never updated, because the TUI's `save()` syncs existing rows through `merge_from_tui`, which only copied `status` and the activity timestamps. Net effect: the swap was silently dropped.

**Fix.** `merge_from_tui` now also syncs `tool`, `command`, and `extra_args`. These are TUI-authoritative launch config, set at creation and mutated post-creation only by the restart dialog (no CLI/server writer), so a plain copy is race-safe. With disk carrying the swap, `reconcile_from_disk` reads the new engine instead of clobbering it.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (N/A)
- [ ] For UI changes: included screenshot or recording (N/A, no visual change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated a test: `test_merge_from_tui_syncs_launch_config_swap` in `src/session/instance.rs` covers the persistence layer that was dropping the swap. Ran `cargo test --lib merge_from_tui` and the `restart_selected_session` suite; both green. `cargo fmt` clean.

## AI Usage

- [x] AI was used

## Summary

- Fixed restart-dialog AI engine swaps so the selected tool, command, and arguments are saved and remain effective after a session restart.
- Added coverage to verify launch configuration changes are preserved during TUI-to-disk synchronization.

## Benefits

- Prevents user-selected restart settings from being silently reverted to stale values.
- Improves reliability of session restarts.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -1649,6 +1649,15 @@ impl Instance {
         self.status = src.status;
         self.last_accessed_at = self.last_accessed_at.max(src.last_accessed_at);
         self.idle_entered_at = src.idle_entered_at;
+        // Launch-config fields are TUI-authoritative and only mutated after
+        // creation by the restart dialog (engine / command / args swap). They
+        // have no peer writer, so a plain copy is safe. Syncing them here is
+        // required: `reconcile_from_disk`'s `*self = disk` reload runs on every
+        // launch, so a swap that never reached disk is silently reverted and
+        // the session respawns with its original tool. See #switching-tools.
+        self.tool = src.tool.clone();
+        self.command = src.command.clone();
+        self.extra_args = src.extra_args.clone();
     }
 
     /// Apply a passively-detected status transition to a disk row. Touches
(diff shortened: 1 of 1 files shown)


njbrake (approved):

Approving. This fixes a real bug: the restart dialog's engine/command/args swap was applied only to the TUI's in-memory row, then silently reverted on the next launch when `reconcile_from_disk` reloads the whole instance from disk, because `save()`'s `merge_from_tui` never wrote those fields.

Verified against source:
- The revert mechanism is exactly as described: `merge_from_tui` previously synced only `status` plus the two activity timestamps.
- `restart_selected_session` applies the swaps and then calls `save()`, so the new `merge_from_tui` copies land on disk before the next spawn.
- The race-safety claim holds. The only post-creation writer of `tool` / `command` / `extra_args` is the TUI restart path; every other writer is creation-time (`cli/add.rs`, `session/builder.rs`) or test code. The server reads these fields to spawn but does not persist them onto existing rows, so a plain TUI-to-disk copy has no peer writer to clobber.
- `cargo test --lib merge_from_tui` passes, including the new `test_merge_from_tui_syncs_launch_config_swap` and the existing `does_not_touch_user_action_fields` guard.

Minimal, well-tested, correctly scoped. Two optional nits, neither blocking: the `merge_from_tui` doc-comment header still frames its fields as the old set (the added inline comment covers the why), and `See #switching-tools.` is an informal anchor rather than a real issue reference.

PR #3078, opened 2026-07-23
fix(session): wire OMP session capture so --resume works

njbrake:
## Description

OMP (Oh My Pi) shipped in #3073 with `ResumeStrategy::Flag("--resume")` but no session-capture wiring, so `--resume` never had a session id to resume: every restart began a fresh conversation.

Session capture is dispatched by tool name in two places in `src/session/instance.rs`, the live poller and `try_retroactive_capture`. Neither had an `"omp"` arm, so omp fell through to the `_ => return` / `_ => None` defaults. With no id ever captured, `agent_session_id` stayed `None`, `build_resume_flags` was never fed a valid id, and the resume flag was effectively dead code.

OMP is a pi fork that shares pi's on-disk session format, but its host data dir defaults to `~/.omp/agent` rather than pi's `~/.pi/agent`. Routing omp straight through `capture_pi_session_id` would have scanned the wrong (empty) directory, so the fix also remaps the host default while still honoring `PI_CODING_AGENT_DIR`.

### Changes

- `src/session/capture.rs`: factor `capture_pi_session_id` into a shared `capture_pi_family_session_id` parameterized by host default subdir; add `capture_omp_session_id` (defaults `~/.omp/agent`, honors `PI_CODING_AGENT_DIR`) and `omp_poll_fn`. Sandboxed omp reuses pi's container scan, since the omp container already sets `PI_CODING_AGENT_DIR`, so no container-side change is needed.
- `src/session/instance.rs`: add `"omp"` arms to `try_retroactive_capture` and the live poller (host uses the omp-defaulted capture; sandbox reuses the pi container path).

This targets the tmux/terminal resume path. OMP's ACP structured-view resume goes through the ACP `EventStore` and was already wired in #3073.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (none needed; behavior fix)
- [ ] For UI changes: included screenshot or recording (N/A)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated a test: unit regression tests in `src/session/capture.rs`:
  - `test_capture_omp_defaults_to_omp_agent_dir`: with `PI_CODING_AGENT_DIR` unset, omp finds a session under `~/.omp/agent` while pi capture does not. Guards the path remap; fails without the default-subdir fix.
  - `test_capture_omp_session_id_basic`: omp reads the shared pi format when `PI_CODING_AGENT_DIR` is set.

Verification: `cargo build` clean, `cargo fmt --check` and `cargo clippy` clean, 130 `session::capture` tests plus 266 omp-matching tests pass.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Root-cause investigation and implementation done via Claude Code, reviewed by @njbrake.

## Summary

- Fixed OMP session capture so `--resume` can reuse previous conversations.
- Shared session-capture logic between Pi and OMP while supporting each tool’s default data directory.
- Added OMP support to live polling and retroactive capture, including sandboxed environments.
- Added regression tests for OMP directory handling and shared session formats.

## Benefits

OMP sessions now reliably retain their IDs, improving conversation resumption across standard and sandboxed workflows.

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -508,7 +508,31 @@ pub(crate) fn capture_pi_session_id(
     project_path: &str,
     exclusion: &HashSet<String>,
 ) -> Result<String> {
-    let pi_home = resolve_agent_home(Some("PI_CODING_AGENT_DIR"), ".pi/agent")?;
+    capture_pi_family_session_id(project_path, exclusion, ".pi/agent")
+}
+
+/// Capture an Oh My Pi (omp) session ID.
+///
+/// OMP is a pi fork that shares pi's on-disk session format and the
+/// `PI_CODING_AGENT_DIR` override, but defaults its data dir to `~/.omp/agent`
+/// on the host rather than `~/.pi/agent`. Only the host default differs, so
+/// this delegates to the shared pi scan with the omp default subdir.
+pub(crate) fn capture_omp_session_id(
+    project_path: &str,
+    exclusion: &HashSet<String>,
+) -> Result<String> {
+    capture_pi_family_session_id(project_path, exclusion, ".omp/agent")
+}
+
+/// Shared pi-family session scan. `default_subdir` is the host home directory
+/// used when `PI_CODING_AGENT_DIR` is unset (`.pi/agent` for pi, `.omp/agent`
+/// for omp).
+fn capture_pi_family_session_id(
+    project_path: &str,
+    exclusion: &HashSet<String>,
+    default_subdir: &str,
+) -> Result<String> {
+    let pi_home = resolve_agent_home(Some("PI_CODING_AGENT_DIR"), default_subdir)?;
     let sessions_dir = pi_home.join("sessions");
 
     if !sessions_dir.exists() {
@@ -647,6 +671,26 @@ pub(crate) fn pi_poll_fn(
         capture_pi_session_id(&project_path, &exclusion)
             .map_err(
                 |e| tracing::debug!(target: "session.capture", "Pi poll capture failed: {}", e),
+            )
+            .ok()
+            .and_then(validated_session_id)
+    }
+}
+
+/// Host polling closure for Oh My Pi (omp), mirroring [`pi_poll_fn`] against
+/// omp's `~/.omp/agent` default data dir. Sandboxed omp reuses
+/// [`pi_poll_fn_sandboxed`], since `PI_CODING_AGENT_DIR` is set in the omp
+/// container so the shared container scan already resolves the right dir.
+pub(crate) fn omp_poll_fn(
+    project_path: String,
+    instance_id: String,
+    extra_excludes: HashSet<String>,
+) -> impl Fn() -> Option<String> + Send + 'static {
+    move || {
+        let exclusion = compose_exclusion(&instance_id, &extra_excludes);
+        capture_omp_session_id(&project_path, &exclusion)
+            .map_err(
+                |e| tracing::debug!(target: "session.capture", "OMP poll capture failed: {}", e),
             )
             .ok()
             .and_then(validated_session_id)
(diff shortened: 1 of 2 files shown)


njbrake:
_Note: this comment was drafted by Claude via back-and-forth with @njbrake._

@smitt04 @leowang-prism this PR wires up OMP session capture so `--resume` actually continues a conversation on restart (it was inert before, see #3073). I can't request either of you as a formal reviewer since you're not repo collaborators, and I don't run OMP myself, so I'd really value a real-world test: start an OMP session, restart it, and confirm it resumes rather than starting fresh. Both host and sandbox (Docker) paths would be great to check. Thanks for building and requesting this.

src/session/capture.rs
```diff
@@ -3227,6 +3271,88 @@ mod tests {
         }
     }
 
+    /// omp shares pi's on-disk format: with `PI_CODING_AGENT_DIR` set, the omp
+    /// capture reads the same sessions dir the pi capture would.
+    #[test]
+    #[serial]
+    fn test_capture_omp_session_id_basic() {
```
jerome-benoit, line 3278: Non-blocking: because this sets `PI_CODING_AGENT_DIR`, `resolve_agent_home` returns early and never consults `default_subdir` — so this test would stay green even if the `.omp/agent` remap were reverted. It's a useful format-parity smoke test, but the actual fail-on-revert guard is `test_capture_omp_defaults_to_omp_agent_dir` (which correctly asserts omp finds the `~/.omp/agent` session AND pi does not). No change needed; just flagging that this one isn't the guard.

src/session/instance.rs
```diff
@@ -2485,6 +2485,24 @@ impl Instance {
                     capture_pi_session_id(&self.project_path, &exclusion).ok()
                 }
             }
+            "omp" => {
```
jerome-benoit, line 2488: The arm is correct (host → `capture_omp_session_id`, sandbox → the pi container capture since the omp container sets `PI_CODING_AGENT_DIR`). Non-blocking: neither this arm nor the poller arm (4166) is unit-tested, so a future typo (e.g. calling `capture_pi_session_id` here) would silently re-break omp resume and still pass CI. A cheap dispatch-level guard on `self.tool == "omp"` selecting the omp capture would lock it.

jerome-benoit (approved):
## Review: wire OMP session capture so --resume works — LGTM

Verified in depth: 3 scoped reviewers + an adversarial pass, plus my own build and `cargo test --lib session::capture` (130/0, incl. both new omp tests). The fix is correct and complete for the tmux/terminal path it targets, and I couldn't break it.

**Confirmed:**
- **Tool-name matches.** The `"omp"` arms fire: registry `AgentDef.name = "omp"` with no aliases/`detect_as`, `"omp"` collides with no earlier agent name in `resolve_tool_name`'s substring scan, so a terminal omp session is stored as `self.tool == "omp"`.
- **No dispatch site missed.** The two patched sites (`try_retroactive_capture`, live poller) are the only tool-name capture dispatches; `capture_freshest_session_id` delegates to the former, and `should_attempt_resume` / `build_resume_flags` are registry-driven (`ResumeStrategy::Flag("--resume")`), so they cover omp with no per-tool code. Resume chain intact end-to-end.
- **Pi no-regression.** `capture_pi_family_session_id` is a byte-for-byte extraction; the pi path passes `.pi/agent`, identical to the old literal.
- **Sandbox chain verified.** omp `container_env` sets `PI_CODING_AGENT_DIR=/root/.omp/agent` (agents.rs) → injected in `build_container_config` → consumed by the container scan (`${PI_CODING_AGENT_DIR:-...}/sessions`), so sandboxed omp reuses pi's container capture correctly.

**Non-blocking notes:**
1. `test_capture_omp_session_id_basic` sets `PI_CODING_AGENT_DIR`, which short-circuits `resolve_agent_home` before `default_subdir` is consulted — so it stays green even if the `.omp/agent` remap were reverted. It's a format-parity smoke test; the real fail-on-revert guard is `test_capture_omp_defaults_to_omp_agent_dir` (which is solid). Fine as-is, just noting it isn't the guard.
2. Neither instance.rs `"omp"` dispatch arm is unit-tested. The code is correct (host → `capture_omp_session_id`/`omp_poll_fn`, sandbox → pi container capture), but a future wrong-fn typo there would silently defeat the fix and pass CI — a cheap guard would help.
3. `capture_pi_family_session_id` keeps pi-specific error text ("Pi sessions directory not found") now shared with omp — cosmetically misleading in omp logs; trivial.
4. Optional: the two tests hand-roll `HOME`/`PI_CODING_AGENT_DIR` save-restore; the repo's `EnvGuard` RAII helper is panic-safe and would be more consistent with the newer capture tests.
5. Observation (pre-existing, shared with pi, not this PR): docker env dedup is first-wins and `collect_environment` (user/profile `extra_env`) is prepended before the agent `container_env`, so a user-set `PI_CODING_AGENT_DIR` overrides the agent default in sandbox mode for both pi and omp. Worth a separate hardening pass if that matters; not a regression here.

Clean, well-scoped fix with a genuine regression guard on the load-bearing path remap. Approving.

smitt04:
@njbrake looks like it works great on host with omp resume. I am not able to test sandbox at this time.

Issue #2587, opened 2026-06-30
Dashboard transcript artifact links under /tmp look clickable but cannot open


amanzainal:
## Problem

When an agent shares a generated screenshot or status artifact from `/tmp/codex-agent-views/...`, the structured dashboard renders it as a clickable link. The link appears valid, but clicking it does not open the artifact.

Example transcript link:

```text
/tmp/codex-agent-views/aoe-sidebar-markers/sidebar-markers-aoe-serve-right-aligned-mobile.png
```

When copied from the browser, this becomes something like:

```text
https://snake-wsl.tail931cb3.ts.net/tmp/codex-agent-views/aoe-sidebar-markers/sidebar-markers-aoe-serve-right-aligned-mobile.png
```

Opening that URL routes back to the dashboard. In some cases, clicking directly from the transcript also shows this toast:

```text
Could not open /tmp/codex-agent-views/...png: not inside this session's repo
```

## Why this seems to happen

This looks like an interaction between two intended behaviors:

1. The structured transcript intercepts filesystem-looking links and tries to resolve them as files inside the active session repo/worktree.
2. `aoe serve` does not serve arbitrary host filesystem paths. Unknown routes fall back to the SPA dashboard.

The security boundary makes sense. The confusing part is that the UI presents a temp artifact path as a normal clickable link even though it cannot work in the dashboard.

## Question

Is this intended behavior, or should agent-generated artifacts under `/tmp/codex-agent-views` be handled differently?

From a user expectation standpoint, when I see a screenshot link in the web dashboard, I expect it to open in the browser, not in my local image viewer and not in the repo diff viewer.

## Possible fixes

A few possible designs:

1. Do not auto-link absolute local paths outside the active repo. Render them as copyable path text instead.
2. Keep repo/source-file links opening in AoE's file viewer, but treat generated artifacts separately.
3. Add an authenticated, scoped artifact route, for example:

```text
/artifacts/<session-id>/<artifact-id>/<filename>
```

Then screenshots/status HTML generated by agents could open in the browser, while arbitrary `/tmp` paths remain inaccessible.

My preference would be #3 for agent-generated screenshots/status pages, plus #1 for arbitrary local paths outside the repo. That keeps the security model clean while making generated artifacts behave like web artifacts in the dashboard.

PR #3090, opened 2026-07-24
feat(web): inline Markdown viewer + Files pane

### LINKED ISSUE #3088: Feature Request: Inline Markdown Viewer in the Web Dashboard
First of all, you guys are awesome! AoE has quickly become one of my favorite tools. Thank you for all the work you've put into it.

## Feature Request

It would be great if the **Web Dashboard** (`aoe serve`) could support viewing Markdown (`.md`) files inline.

Currently, `.md` files are not clickable, so whenever an agent generates something like a `plan.md`, I have to switch over to my IDE just to read it. It interrupts the workflow since I'm already using the Web Dashboard to monitor the agent.

### Proposed behavior

- Make `.md` files clickable in the file tree.
- Render Markdown directly within the Web Dashboard (similar to GitHub's Markdown renderer).
- Optionally allow toggling between the rendered view and the raw Markdown source.

I think this would be especially useful since agents frequently generate planning documents, design notes, implementation plans, and other Markdown files.

Thanks again for building such an amazing project!

Seluj78:

## Description

Adds inline Markdown viewing to the web dashboard, in two places, plus a general session file browser.

**1. Markdown in the diff viewer.** `.md` / `.markdown` files in the diff/file list now render as formatted Markdown by default, with a **Rendered** / **Raw** toggle (Raw keeps the syntax-highlighted source / diff). Rendering uses the already-bundled `react-markdown` + `remark-gfm` and escapes content (no `dangerouslySetInnerHTML`, no `rehype-raw`). The choice persists as a client-local `markdownPreview` setting.

**2. Files pane.** A new **Files** activity-bar pane browses the session's working directory, not just its git changes, so it lists files even in a non-git scratch session (the reported gap: scratch sessions have no diff, so their files never appeared). Selecting a file opens it in a viewer that renders Markdown or falls back to the shared shiki full-file view.

**3. Provenance-confined file endpoint.** `GET /api/sessions/{id}/file` reads a file for the viewer, git-agnostic. A read is allowed only when the canonical target is under a session project root, or is a path the agent actually touched this session (Write/Edit/Read/apply_patch/memory-recall), recovered from the ACP event log. Transcript links to files the agent wrote outside the repo (a plan in `/tmp`, a note in an agent config dir) are now openable via this endpoint; the server's provenance check is the gate.

**Dependency note:** the sandboxed read adds `cap-std` (optional, gated behind the `serve` feature). An in-tree `rustix`/`libc` `openat2` + `O_NOFOLLOW` path would avoid the ~9 transitive crates, but it reintroduces the OS-specific branching AGENTS.md wants isolated under `src/process/`, so the capability crate was the deliberate trade-off.

The confinement design was reviewed via `/debate` (two models): it canonicalizes before any check, uses component-aware `Path::starts_with`, opens the canonical path (no TOCTOU symlink swap), rejects non-regular files (FIFO/device/`/proc`), and bounds the read. An earlier ambient `/tmp` + agent-home allowlist was rejected in favor of provenance, because env-controlled roots and a `.md`-only rule were not a real boundary.

Closes #3088

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open a session with a changed `.md` file; click it in the diff list and confirm it renders as formatted Markdown, then click **Raw** to see the diff/source.
- [ ] Reload and open another `.md`; confirm the Rendered/Raw choice persisted.
- [ ] Create a scratch (non-git) session with a `.md` in its directory; open the **Files** pane (folder icon), confirm the file is listed, click it, and confirm it renders.
- [ ] In a session where the agent wrote a Markdown file outside the repo (e.g. `/tmp/plan.md`), click that path in the transcript and confirm it opens rendered.
- [ ] Confirm a non-Markdown file opens in the syntax-highlighted viewer with no Rendered/Raw toggle.

## Checklist

## Test Coverage Analysis

Added a live Playwright spec (`right-panel-files.spec.ts`) exercising the scratch-session repro end to end, Vitest for the Markdown/Files components, and Rust unit tests for the path-confinement + provenance extraction. Coverage matrix updated with `right-panel.markdown-viewer`, `right-panel.files-viewer`, and `right-panel.files-pane-live`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (default to rendered; provenance instead of an ambient allowlist for out-of-project reads) and reviewed each commit; the confinement design was run through a multi-model `/debate`.

**NOTE:**

## Summary

- Added inline Markdown viewing in the web diff/file viewer for `.md`/`.markdown` files, with a Rendered/Raw toggle and persisted preference.
- Introduced a new “Files” pane to browse and open session working-directory files (including non-git/scratch sessions), with Markdown rendering and the same rendered/raw toggle.
- Added a provenance-confined, git-agnostic backend endpoint to serve session files, allowing reads only for project roots or files the agent actually touched; safely handles binary and oversized content.
- Updated transcript link behavior so absolute/out-of-repo file references route to the in-app viewer instead of being blocked as inert text.
- Added/updated tests and coverage (Vitest, React Testing Library, Playwright, and Rust), plus documentation for the Files pane and Markdown behavior.

## Benefits

Users can read agent-generated plans and design notes directly in the dashboard (rendered Markdown by default) without switching to an IDE, while the server continues to restrict file viewing to authorized/provenanced content.

### DIFF
--- a/src/server/api/file_provenance.rs
+++ b/src/server/api/file_provenance.rs
@@ -1,0 +1,419 @@
+//! Provenance-based confinement for the session file-read endpoint (#3088).
+//!
+//! The web dashboard can render a Markdown (or any) file that belongs to a
+//! session. A read is allowed only when the canonicalized target is either
+//!   - under one of the session's project roots (project_path + worktree
+//!     paths), or
+//!   - a path the session's agent actually touched this session (Write / Edit /
+//!     Read / apply_patch / memory-recall), recovered from the ACP event log.
+//!
+//! Provenance, not a directory allowlist, is the boundary: the dashboard can
+//! open exactly what the agent already worked with (whose content was already
+//! in the transcript), and nothing it never touched. See the debate synthesis
+//! on #3088 for why an ambient `/tmp` + agent-home allowlist was rejected.
+
+use std::collections::HashSet;
+use std::io::Read;
+use std::path::{Component, Path, PathBuf};
+
+use axum::http::StatusCode;
+
+use crate::acp::state::Event;
+
+/// Keys under which the various agents stash a file path in a tool call's
+/// `args_preview` JSON. Mirrors the client's `pickStr` order in
+/// `web/src/components/acp/ToolCards.tsx`.
+const PATH_KEYS: [&str; 4] = ["file_path", "path", "filePath", "filename"];
+
+/// Pull a file path out of a tool call's `args_preview` JSON blob. Returns
+/// `None` when the blob does not parse (it is capped at 16 KB at ingest, so a
+/// huge leading argument can truncate the JSON) or carries no path key; callers
+/// treat that as "path unknown", never "denied", and fall back to the
+/// structured `diffs[].path`.
+fn extract_path_from_args_preview(args_preview: &str) -> Option<String> {
+    let value: serde_json::Value = serde_json::from_str(args_preview).ok()?;
+    let obj = value.as_object()?;
+    for key in PATH_KEYS {
+        if let Some(s) = obj.get(key).and_then(|v| v.as_str()) {
+            if !s.is_empty() {
+                return Some(s.to_string());
+            }
+        }
+    }
+    None
+}
+
+/// Build the set of paths a session's agent touched, from its ACP event log.
+///
+/// Folds both `ToolCallStarted` and `ToolCallUpdated`: claude-agent-acp ships
+/// the initial tool call with an empty `raw_input` and fills the path in on a
+/// later update, so the update frames must be scanned too. Paths are collected
+/// exactly as the agent emitted them (absolute for Claude's Read/Write/Edit),
+/// so entries are directly usable as absolute allow-set keys.
+pub fn collect_touched_paths(events: &[(u64, Event)]) -> HashSet<PathBuf> {
+    let mut out = HashSet::new();
+    let mut add = |s: &str| {
+        if !s.is_empty() {
+            out.insert(PathBuf::from(s));
+        }
+    };
+    for (_seq, event) in events {
+        match event {
+            Event::ToolCallStarted { tool_call } => {
+                if let Some(p) = extract_path_from_args_preview(&tool_call.args_preview) {
+                    add(&p);
+                }
+                for d in &tool_call.diffs {
+                    add(&d.path);
+                }
+                if let Some(mr) = &tool_call.memory_recall {
+                    for p in &mr.paths {
+                        add(p);
+                    }
+                }
+            }
+            Event::ToolCallUpdated {
+                args_preview,
+                diffs,
+                ..
+            } => {
+                if let Some(ap) = args_preview {
+                    if let Some(p) = extract_path_from_args_preview(ap) {
+                        add(&p);
+                    }
+                }
+                if let Some(diffs) = diffs {
+                    for d in diffs {
+                        add(&d.path);
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+    out
+}
+
+/// Reject a relative request that tries to climb out of its root. Absolute
+/// requests skip this (they are matched against roots / provenance instead).
+fn has_traversal(requested: &Path) -> bool {
+    requested.components().any(|c| {
+        matches!(
+            c,
+            Component::ParentDir | Component::RootDir | Component::Prefix(_)
+        )
+    })
+}
+
+/// Resolve and confine a requested path.
+///
+/// `project_roots` must already be canonicalized. `touched` is the raw
+/// provenance set (canonicalized here, lazily; entries that no longer resolve
+/// are skipped). Returns the canonical path to read, or an HTTP error.
+///
+/// Security invariants (see #3088 debate): the path is canonicalized (symlinks
+/// resolved, `..` collapsed) before any containment check; containment uses
+/// `Path::starts_with` (component-aware, so `/repo-evil` is not under `/repo`);
+/// the returned canonical path is what the caller must open (never the raw
+/// request), closing the symlink-swap gap between check and read.
+pub fn confine_path(
+    project_roots: &[PathBuf],
+    touched: &HashSet<PathBuf>,
+    requested: &Path,
+) -> Result<PathBuf, (StatusCode, &'static str)> {
+    if requested.as_os_str().is_empty() {
+        return Err((StatusCode::BAD_REQUEST, "empty path"));
+    }
+
+    let canonical = if requested.is_absolute() {
+        requested
+            .canonicalize()
+            .map_err(|_| (StatusCode::NOT_FOUND, "file not found"))?
+    } else {
+        if has_traversal(requested) {
+            return Err((StatusCode::BAD_REQUEST, "path escapes project"));
+        }
+        // A relative request is always resolved against the primary project
+        // root (the first entry). Multi-repo members are reached by their
+        // absolute worktree path via provenance / project-root containment.
+        let root = project_roots
+            .first()
+            .ok_or((StatusCode::INTERNAL_SERVER_ERROR, "no project root"))?;
+        root.join(requested)
+            .canonicalize()
+            .map_err(|_| (StatusCode::NOT_FOUND, "file not found"))?
+    };
+
+    // Under a project root: any file (the agent's own workspace).
+    if project_roots.iter().any(|r| canonical.starts_with(r)) {
+        return Ok(canonical);
+    }
+
+    // Otherwise it must be a path the agent touched this session. Compare
+    // canonical forms so a symlinked-but-touched path still matches.
+    let touched_hit = touched
+        .iter()
+        .any(|t| t.canonicalize().map(|ct| ct == canonical).unwrap_or(false));
+    if touched_hit {
+        return Ok(canonical);
+    }
+
+    Err((StatusCode::FORBIDDEN, "path not readable for this session"))
+}
+
+/// Read a confined, canonical path with a byte cap.
+///
+/// Rejects non-regular files (directories, FIFOs, devices, `/proc` nodes)
+/// before reading, so a blocking or endless special file can't stall or OOM the
+/// server. Reads at most `cap + 1` bytes to detect truncation without
+/// allocating an unbounded buffer. Returns `(content, is_binary, truncated)`;
+/// binary content yields an empty string (the client shows a "binary file"
+/// notice), matching the diff endpoint.
+pub fn read_bounded(
+    path: &Path,
+    cap: usize,
+) -> Result<(String, bool, bool), (StatusCode, &'static str)> {
+    let mut file =
+        std::fs::File::open(path).map_err(|_| (StatusCode::NOT_FOUND, "file not found"))?;
+    let meta = file
+        .metadata()
+        .map_err(|_| (StatusCode::INTERNAL_SERVER_ERROR, "stat failed"))?;
+    if !meta.file_type().is_file() {
+        return Err((StatusCode::BAD_REQUEST, "not a regular file"));
+    }
+
+    let mut bytes = Vec::new();
+    file.by_ref()
+        .take(cap as u64 + 1)
+        .read_to_end(&mut bytes)
+        .map_err(|_| (StatusCode::INTERNAL_SERVER_ERROR, "read failed"))?;
+
+    let truncated = bytes.len() > cap;
+    if truncated {
+        bytes.truncate(cap);
+    }
+    let is_binary = bytes.contains(&0);
+    let content = if is_binary {
+        String::new()
+    } else {
+        String::from_utf8_lossy(&bytes).into_owned()
+    };
+    Ok((content, is_binary, truncated))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::acp::state::{DiffPreview, MemoryRecall, ToolCall};
+    use std::fs;
+
+    fn tool_call(args_preview: &str, diffs: Vec<&str>, recall: Vec<&str>) -> ToolCall {
+        ToolCall {
+            id: "t1".into(),
+            name: "Write".into(),
+            kind: "edit".into(),
+            args_preview: args_preview.into(),
+            started_at: chrono::Utc::now(),
+            parent_tool_call_id: None,
+            memory_recall: if recall.is_empty() {
+                None
+            } else {
+                Some(MemoryRecall {
+                    mode: "recall".into(),
+                    paths: recall.into_iter().map(String::from).collect(),
+                    synthesized_text: None,
+                })
+            },
+            diffs: diffs
+                .into_iter()
+                .map(|p| DiffPreview {
+                    path: p.into(),
+                    old_text: None,
+                    new_text: None,
+                    created_at: chrono::Utc::now(),
+                })
+                .collect(),
+        }
+    }
+
+    #[test]
+    fn collects_paths_from_all_channels() {
+        let events = vec![
+            (
+                1u64,
+                Event::ToolCallStarted {
+                    tool_call: tool_call(r#"{"file_path":"/tmp/plan.md"}"#, vec![], vec![]),
+                },
+            ),
+            (
+                2,
+                Event::ToolCallStarted {
+                    tool_call: tool_call(
+                        "{}",
+                        vec!["/tmp/patched.rs"],
+                        vec!["/home/u/.claude/mem.md"],
+                    ),
+                },
+            ),
+            (
+                3,
+                Event::ToolCallUpdated {
+                    tool_call_id: "t3".into(),
+                    title: None,
+                    args_preview: Some(r#"{"path":"/tmp/late.txt"}"#.into()),
+                    started_at: None,
+                    diffs: Some(vec![DiffPreview {
+                        path: "/tmp/diff-late.rs".into(),
+                        old_text: None,
+                        new_text: None,
+                        created_at: chrono::Utc::now(),
+                    }]),
+                },
+            ),
+        ];
+        let set = collect_touched_paths(&events);
+        for p in [
+            "/tmp/plan.md",
+            "/tmp/patched.rs",
+            "/home/u/.claude/mem.md",
+            "/tmp/late.txt",
+            "/tmp/diff-late.rs",
+        ] {
+            assert!(set.contains(&PathBuf::from(p)), "missing {p}");
+        }
+    }
+
+    #[test]
+    fn unparseable_args_preview_is_ignored_not_denied() {
+        // A 16 KB-truncated args_preview may be invalid JSON; that must not
+        // panic or fabricate a path, and the diffs fallback still lands.
+        let events = vec![(
+            1u64,
+            Event::ToolCallStarted {
+                tool_call: tool_call(r#"{"content":"unterminated"#, vec!["/tmp/x.rs"], vec![]),
+            },
+        )];
+        let set = collect_touched_paths(&events);
+        assert_eq!(set.len(), 1);
+        assert!(set.contains(&PathBuf::from("/tmp/x.rs")));
+    }
+
+    #[test]
+    fn extract_prefers_key_order() {
+        assert_eq!(
+            extract_path_from_args_preview(r#"{"filename":"b","file_path":"a"}"#),
+            Some("a".into())
+        );
+        assert_eq!(extract_path_from_args_preview(r#"{"nope":"x"}"#), None);
+        assert_eq!(extract_path_from_args_preview("not json"), None);
+    }
+
+    #[test]
+    fn confine_allows_relative_in_project_rejects_traversal() {
+        let dir = tempfile::tempdir().unwrap();
+        let root = dir.path().canonicalize().unwrap();
+        fs::write(root.join("a.md"), "# hi").unwrap();
+        let roots = vec![root.clone()];
+        let touched = HashSet::new();
+
+        assert!(confine_path(&roots, &touched, Path::new("a.md")).is_ok());
+        assert_eq!(
+            confine_path(&roots, &touched, Path::new("../etc/passwd"))
+                .unwrap_err()
+                .0,
+            StatusCode::BAD_REQUEST
+        );
+        assert_eq!(
+            confine_path(&roots, &touched, Path::new("")).unwrap_err().0,
+            StatusCode::BAD_REQUEST
+        );
+    }
+
+    #[test]
+    fn confine_absolute_requires_project_or_provenance() {
+        let dir = tempfile::tempdir().unwrap();
+        let root = dir.path().canonicalize().unwrap();
+        let outside = tempfile::tempdir().unwrap();
+        let outside_root = outside.path().canonicalize().unwrap();
+
+        let in_project = root.join("in.md");
+        fs::write(&in_project, "x").unwrap();
+        let touched_file = outside_root.join("plan.md");
+        fs::write(&touched_file, "x").unwrap();
+        let untouched_file = outside_root.join("secret.md");
+        fs::write(&untouched_file, "x").unwrap();
+
+        let roots = vec![root.clone()];
+        let mut touched = HashSet::new();
+        touched.insert(touched_file.clone());
+
+        // Absolute path inside the project root: allowed.
+        assert!(confine_path(&roots, &touched, &in_project).is_ok());
+        // Absolute path the agent touched: allowed.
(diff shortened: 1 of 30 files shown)


Seluj78:
@coderabbitai review

web/src/components/diff/DiffFileViewer.tsx
```diff
@@ -409,56 +417,90 @@ export function DiffFileViewer({
...
+              >
+                Split
+              </button>
+            </div>
+          )}
         </div>
       </div>
 
-      {findOpen && !contents.is_binary && !contents.truncated && (
+      {findOpen && !showRendered && !contents.is_binary && !contents.truncated && (
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/b52da848: Cmd/Ctrl+F is only intercepted in Raw mode now, so the browser's native find works on rendered Markdown.

web/src/components/diff/DiffFileViewer.tsx
```diff
@@ -472,7 +514,9 @@ export function DiffFileViewer({
             <span className="text-xs text-text-dim bg-surface-900/80 rounded px-2 py-1">Loading diff...</span>
           </div>
         )}
-        {contents.is_binary ? (
+        {showRendered ? (
+          <MarkdownFileView content={newContent || oldContent} />
+        ) : contents.is_binary ? (
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/b52da848: a modified-to-empty Markdown file now renders its (empty) new content; only a deleted file falls back to old content.

src/server/api/file_provenance.rs
```diff
@@ -0,0 +1,419 @@
...
+    // canonical forms so a symlinked-but-touched path still matches.
+    let touched_hit = touched
+        .iter()
+        .any(|t| t.canonicalize().map(|ct| ct == canonical).unwrap_or(false));
+    if touched_hit {
+        return Ok(canonical);
+    }
+
+    Err((StatusCode::FORBIDDEN, "path not readable for this session"))
+}
```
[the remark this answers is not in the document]
Seluj78: Hardened in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/037ed2a4: the read now opens beneath a cap-std capability directory (Dir::open_ambient_dir + dir.open), which refuses '..' and symlink escapes, so the open cannot leave the intended root even if a component is swapped after the containment check.

web/src/App.tsx
```diff
@@ -1202,6 +1210,15 @@ function AppContent({
...
       if (!resolved) {
+        // Outside the session's repo roots. The agent may still have touched
+        // it this session (e.g. a plan written to /tmp or ~/.claude); the
+        // provenance-confined /file endpoint is the gate, so open it in the
+        // FileContentViewer and let the server allow or refuse. A relative
+        // path we couldn't resolve has no absolute target to try. See #3088.
+        if (ref.path.startsWith("/")) {
+          setSelectedFile({ path: ref.path, line: ref.line, cited: true, external: true });
+          return;
+        }
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/b52da848: both gates now use a shared isAbsolutePath predicate that also matches C:/ and C:\ drive-letter paths.

web/src/App.tsx
```diff
@@ -1623,6 +1640,9 @@ function AppContent({
       if (id === "agents") {
         return <BackgroundAgentsPanel sessionId={activeSessionId} />;
       }
+      if (id === "files") {
+        return <FilesPane sessionId={activeSessionId} />;
+      }
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/b52da848: 'files' is now in allPaneIds so the activity bar shows the toggle. It stays closed by default (excluded from AUTO_OPEN_PANES, like Sub agents).

web/src/components/FilesPane.tsx
```diff
@@ -0,0 +1,74 @@
...
+  if (!sessionId) {
+    return (
+      <div className="flex-1 flex items-center justify-center bg-surface-900 text-text-dim">
+        <span className="text-sm">No active session</span>
+      </div>
+    );
+  }
+
+  if (selected) {
+    return <FileContentViewer sessionId={sessionId} filePath={selected} onBack={() => setSelected(null)} />;
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/b52da848: the Files pane is remounted with key={activeSessionId}, so the selected file (and any in-flight read) resets on session switch.

web/src/components/diff/DiffFileViewer.tsx
```diff
@@ -409,56 +417,90 @@ export function DiffFileViewer({
...
+                title="Raw Markdown source"
+                className={`px-2 py-0.5 text-[11px] font-mono cursor-pointer transition-colors ${
+                  settings.markdownPreview === "raw"
+                    ? "bg-brand-600 text-white"
+                    : "text-text-dim hover:text-text-secondary"
+                }`}
+              >
+                Raw
+              </button>
+            </div>
```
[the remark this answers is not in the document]
Seluj78: These toggles deliberately mirror the adjacent Unified/Split/Find controls in the same header, which already use bg-brand-600 + text-white for their active state; changing only the Markdown toggle would make it visually inconsistent with its neighbors. Migrating every header toggle control to semantic selected-control tokens is worth doing, but as one separate, focused cleanup rather than diverging a single control here. Leaving this open for your call.

web/src/components/diff/FileContentViewer.tsx
```diff
@@ -0,0 +1,147 @@
...
+              title="Raw Markdown source"
+              className={`px-2 py-0.5 text-[11px] font-mono cursor-pointer transition-colors ${
+                settings.markdownPreview === "raw"
+                  ? "bg-brand-600 text-white"
+                  : "text-text-dim hover:text-text-secondary"
+              }`}
+            >
+              Raw
+            </button>
+          </div>
```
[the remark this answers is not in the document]
Seluj78: Same reasoning as the diff-viewer toggle thread: FileContentViewer's Rendered/Raw toggle mirrors the existing header toggle styling for consistency. I'd rather migrate all of these toggle controls to the semantic text/selected tokens in one dedicated pass than change just this one. Leaving open for your call.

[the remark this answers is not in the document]
Seluj78: Agreed, deferring to a dedicated pass. No follow-up issue needed from you, I'll let @Seluj78 decide whether to file one for the toggle-token migration. Resolving this thread.

Seluj78:
@jerome-benoit can I get your review? Especially security wise

jerome-benoit (commented):
## Full review (global + security)

I did a comprehensive pass on top of the CodeRabbit rounds: security (endpoint confinement + XSS + gating), correctness, performance, architecture, dependency, frontend/UX, tests, docs, and conventions. Method: several scoped read-only reviewers per dimension, an adversarial exploit-attempt pass on the file-read confinement, cross-validation, plus my own build and tests in an isolated checkout (`cargo test --features serve --lib file_provenance` = 7/7; web vitest for the 5 new/changed component suites = 76/76).

**Overall: strong, well-tested, and secure. No blocking issues.** One performance item is worth doing before merge; the rest are minor/low or deferred follow-ups.

### Security (the part you flagged) - SOLID
- **Path confinement is sound.** `confine_path` canonicalizes the target (symlinks resolved, `..` collapsed) then allows it only if it is under a canonicalized project root (component-wise `starts_with`, so `/repo-evil` does not match `/repo`) or an exact canonical match to the session's provenance set. The final read opens beneath a `cap_std` capability `Dir`, which re-validates every component at open time (cap-primitives 4.0.2 uses `openat2 RESOLVE_BENEATH|NO_MAGICLINKS`, with a per-component `O_NOFOLLOW` fallback). I ran an adversarial pass specifically trying to build a TOCTOU / symlink-swap / provenance-injection / encoding bypass and could not: the capability open closes the canonicalize-then-open race, and `touched` is derived exclusively from the agent's ACP tool-call events (no HTTP path lets a client seed it), so arbitrary-file-read is not reachable.
- **No XSS.** Both Markdown renderers use `react-markdown` with no `rehype-raw` (absent from the whole lockfile), so raw HTML in a crafted `.md` is escaped to text; `dangerouslySetInnerHTML` is used only on trusted Shiki output; CSP is hardened. There is a test feeding ``/`<script>` that asserts they are inert.
- **Gating/scope.** The route is behind `auth_middleware`; roots + provenance are server-derived from the session, so no cross-session leak. It is a read (correctly not `read_only`-gated) and passes the CityHall default-deny gate as a read. Windows `#2587` inert-link handling is fixed.
- CodeRabbit's security `:188` (TOCTOU) and stability `:220` Majors are resolved (the cap_std switch + non-regular-file/`cap+1` bounding).
- **One security-adjacent test gap (recommend, not a defect):** the cap_std TOCTOU layer is exercised but never *provoked* - no test swaps a component between `confine_path` and the open, so a revert to a plain `File::open(canonical)` would leave every test green. A symlink-inside-root open test through `read_confined` would lock the marquee defense.

### Performance - the one item to consider before merge
The handler replays the **entire** session event log per request (JSON-deserializing every event) to build the provenance set, *before* trying the project-root fast path. For a repo Markdown file (the common case) the provenance set is never needed, yet the whole transcript is re-parsed on every file open, and the Files pane is a many-request surface. Two cheap wins: (1) reorder so project-root containment is attempted first and provenance is built lazily only on a miss - pure reordering, removes the replay from the hot path; (2) the `ponytail:` cache keyed on `highest_seq`. Not blocking, but a visible latency floor on long sessions.

### Correctness - solid
No off-by-one; truncation (`cap+1` probe), binary (NUL heuristic), empty, deleted/renamed, and non-existent files are all handled with sensible status codes. Frontend loading/error/empty/binary/truncated states render usable messages; the Raw/Rendered toggle, stale-response race guard, and session-change remount (`key={activeSessionId}`) are correct. Both CodeRabbit correctness flags (empty-Markdown staleness, FilesPane selection reset) are verified fixed.

### Architecture / quality - good
Clean module (`file_provenance.rs`) and clean component split (`FilesPane` -> `FileContentViewer` -> `MarkdownFileView`, reused by the diff viewer's rendered branch); the DiffFileViewer refactor is a net win; no dead code; pane system integrated consistently (`files` correctly excluded from auto-open). Minors: the handler returns an inline `serde_json::json!` instead of a typed `SessionFileResponse` struct like its sibling `RichFileContentsResponse`; one orphaned comment in DiffFileViewer; `markdownPreview` isn't normalized in `normalizeSnapshot`.

### Dependency
`cap-std 4.0.2` is justified (capability-safe open), feature-gated behind `serve`, and current. It pulls ~9 transitive crates for one capability-open; a lighter in-tree path exists (`rustix`/`libc` `openat2`/`O_NOFOLLOW`, already used for plugin reads) but that reintroduces the OS-specific branching AGENTS.md wants isolated, so cap-std is a reasonable call - worth a one-line note in the PR that the trade-off was weighed.

### Tests / docs / conventions
Feature behavior is well covered (vitest + meaningful live Playwright that spawns a real `aoe serve` against a non-git scratch session), `docs/guides/web/diff.md` accurately documents the Files pane + toggle + provenance model, the coverage-matrix is compliant, commits are conventional, and there are no emdash/`--`, `any`, `ts-ignore`, or dead-code violations. Low: no dedicated test asserts Files-pane session-change/close (covered structurally by the remount key).

### Low / deferred (non-blocking)
- `text-white` on the toggles is flagged by CodeRabbit but is the established repo-wide active-toggle pattern (FindBar, CommentForm, ...); correct to defer to a repo-wide token migration.
- Focus restoration after opening/closing a file; silent 500-file cap in the Files pane; an empty file list is indistinguishable from a fetch error.

Nice feature and a genuinely careful security design. My only real ask is the performance reorder; everything else is polish or a follow-up test. Happy to re-look once the hot-path reorder lands.

src/server/api/sessions.rs
```diff
@@ -6172,6 +6172,93 @@ pub async fn session_diff_file(
...
+            let advance = page.last_scanned_seq;
+            events.extend(page.events);
+            match (page.has_more, advance) {
+                (true, Some(seq)) => since = seq,
+                _ => break,
+            }
+        }
+        let touched = super::file_provenance::collect_touched_paths(&events);
+
+        let confined = super::file_provenance::confine_path(
```
jerome-benoit: [Performance] The full session event log is replayed and JSON-deserialized here (`replay_page` loop -> `collect_touched_paths`) on *every* file open, before `confine_path` is even called. For a repo file the provenance set is never needed (it resolves via the project-root `starts_with` branch), yet the whole transcript is re-parsed each request, and the Files pane fires one request per click. Cheap win: try project-root containment first and build `touched` lazily only on a miss - pure reordering, removes the replay from the hot path. The `ponytail:` per-session cache keyed on `highest_seq` is the follow-up. Not blocking, but a visible latency floor on long sessions.

src/server/api/file_provenance.rs
```diff
@@ -0,0 +1,462 @@
...
+/// before reading, so a blocking or endless special file can't stall or OOM the
+/// server. Reads at most `cap + 1` bytes to detect truncation without
+/// allocating an unbounded buffer. Returns `(content, is_binary, truncated)`;
+/// binary content yields an empty string (the client shows a "binary file"
+/// notice), matching the diff endpoint.
+pub fn read_confined(
+    confined: &Confined,
+    cap: usize,
+) -> Result<(String, bool, bool), (StatusCode, &'static str)> {
+    let dir = Dir::open_ambient_dir(&confined.root, ambient_authority())
```
jerome-benoit, line 206: [Security - solid, with a test-gap] This cap_std capability open is the TOCTOU defense, and it genuinely works: cap-primitives 4.0.2 re-validates every component at open time (`openat2 RESOLVE_BENEATH|NO_MAGICLINKS`, per-component `O_NOFOLLOW` fallback), so a component swapped after the earlier `canonicalize` cannot escape `root`. I tried to build a swap/symlink/provenance-injection bypass and couldn't. But no test *provokes* the race - all tests reach here with a stable tree - so a revert to a plain `File::open(canonical)` would leave every test green. Recommend a `read_confined` test that opens a symlink-inside-root pointing outside and asserts it errors, so this marquee defense stops being exercised-but-never-proven.

src/server/api/sessions.rs
```diff
@@ -6172,6 +6172,93 @@ pub async fn session_diff_file(
...
+            &roots,
+            &touched,
+            std::path::Path::new(&requested),
+        )?;
+        let (content, is_binary, truncated) =
+            super::file_provenance::read_confined(&confined, MAX_CONTENTS_BYTES)?;
+        Ok::<_, (StatusCode, &'static str)>(serde_json::json!({
```
jerome-benoit: [Minor] The sibling file-content endpoint returns a typed `#[derive(Serialize)] RichFileContentsResponse`; this one hand-rolls an inline `serde_json::json!({...})`, which skips compile-time field checking and diverges from the file's own convention. A named `SessionFileResponse` struct would be more consistent and refactor-safe.

src/server/api/sessions.rs
Seluj78: Done in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/329aa0ea. `confine_path` now takes the provenance set as a `FnOnce` and invokes it only when the target lands outside every project root, so a workspace file never triggers the replay. Added `provenance_is_not_built_for_a_project_file`, which fails if the fast path ever regresses. The per-session cache keyed on `highest_seq` stays the follow-up (`ponytail:` note kept on the miss path).

src/server/api/file_provenance.rs
Seluj78: Good catch, and you were right that nothing pinned it. Added `read_confined_refuses_symlink_escaping_root` in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/329aa0ea: it hands `read_confined` a target that escapes `root` via a symlink (the shape a post-check component swap produces) and asserts the open is refused. I verified it actually goes red, by temporarily swapping `dir.open(rel)` for `File::open(canonical)` and re-running: FAILED with cap-std removed, passes with it. So the defense is now locked rather than merely exercised.

src/server/api/sessions.rs
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/329aa0ea: added `#[derive(Serialize)] SessionFileResponse { content, is_binary, truncated }` and the handler returns it instead of the inline `json!`, matching `RichFileContentsResponse`.

Seluj78:
Thanks for the thorough pass @jerome-benoit, especially the adversarial attempt on the confinement. All of it is addressed:

**Performance (your one ask)** - `329aa0ea`. `confine_path` takes the provenance set as a `FnOnce`, invoked only when the target resolves outside every project root, so a workspace file no longer pays for the event-log replay. A test asserts the closure is never called for an in-project file. The `highest_seq` cache stays the follow-up.

**TOCTOU test gap** - `329aa0ea`. Added a `read_confined` test that opens a symlink-inside-root pointing outside and asserts refusal. I checked it genuinely goes red: swapping `dir.open(rel)` for `File::open(canonical)` makes it FAIL, restoring cap-std makes it pass.

**Typed response** - `329aa0ea`. `SessionFileResponse` replaces the inline `json!`.

**CodeRabbit's outside-diff `fileRef.ts` point** (valid, and the doc comment was the misleading part) - `257c3fe2`. `resolveToRepoRelative` now calls `isAbsolutePath` rather than keeping its own narrower regex, so the two cannot drift.

**Two minors from your body** - `257c3fe2`. `markdownPreview` is now normalized in `normalizeSnapshot` (localStorage is user-editable), and the comment orphaned by the Markdown-flag move is gone.

**Dependency note** added to the PR body, recording that the in-tree `rustix`/`openat2` alternative was weighed and rejected to keep OS-specific branching isolated.

Deferred as you suggested: the toggle-token migration (repo-wide pattern, own pass), the provenance cache, focus restoration, and the silent 500-file cap in the Files pane. Validation: `cargo fmt`/`clippy` clean, 9/9 `file_provenance` tests, full web Vitest 3535 passed.

Seluj78:
@jerome-benoit could you give another review before I merge ? :)

src/server/api/file_provenance.rs
```diff
@@ -0,0 +1,525 @@
...
+        return Ok(Confined {
+            canonical,
+            root: root.clone(),
+        });
+    }
+
+    // Outside every project root, so fall back to provenance: only now is it
+    // worth replaying the event log. Compare canonical forms so a
+    // symlinked-but-touched path still matches; open it beneath its own parent.
+    let touched_hit = touched()
```
jerome-benoit, line 178: Verified the perf reorder is behavior-preserving: the project-root branch above returns without ever calling `touched`, exactly as before, so a workspace file no longer pays for the event-log replay; the provenance set is built (via the `FnOnce`) only here on the miss path and is byte-for-byte the same set the eager code built. No allow/deny decision changes. `provenance_is_not_built_for_a_project_file` locks the skip against regression.

src/server/api/file_provenance.rs
```diff
@@ -0,0 +1,525 @@
...
+    /// The capability open is the TOCTOU defense, so provoke it directly rather
+    /// than only reaching it with a stable tree: hand `read_confined` a target
+    /// that escapes `root` via a symlink, the shape a component swapped after
+    /// `confine_path` validated containment would produce. `cap_std` re-checks
+    /// every component at open time and refuses it; a plain
+    /// `File::open(canonical)` would happily follow the link and leak the
+    /// outside file, so this test fails if that hardening is ever reverted.
+    #[cfg(unix)]
+    #[test]
+    fn read_confined_refuses_symlink_escaping_root() {
```
jerome-benoit, line 476: This closes the G1 gap from my prior review: it provokes the cap_std capability open directly with an in-root symlink escaping `root` and asserts `NOT_FOUND`, so a revert to a plain `File::open(canonical)` (which would follow the link and leak the secret) fails the test. The `ok.md` sanity leg confirms the mechanism isn't rejecting everything. Solid. Minor: the trailing `assert_ne!(secret, "")` is a tautology; the real guarantee is the `unwrap_err()` above.

jerome-benoit (approved):
## Re-review: LGTM, approving

All three findings from my previous pass are addressed, and I re-verified with three scoped reviewers plus my own build and tests in an isolated checkout (`file_provenance` = 9/9 incl. the two new tests; web vitest for the touched suites = 53/53).

### Prior findings -> resolution (verified)
| Finding | Fix | Verification |
|---|---|---|
| **Perf**: full event-log replay ran before the project-root fast path | `975c36c8`: `confine_path(touched: impl FnOnce…)`, provenance built only on a project-root miss | **Semantically identical**: the fast path never consulted provenance before and still returns first without invoking the closure; the miss path builds the same set; `FnOnce` runs at most once. No allow/deny decision changes (checked in-root-not-touched, outside-not-touched, empty-roots, symlink, touched-symlink). |
| **Security test gap (G1)**: the cap_std TOCTOU layer was exercised but never provoked | `read_confined_refuses_symlink_escaping_root` | Provokes `read_confined` directly with an in-root symlink escaping `root`, asserts `NOT_FOUND`; **fails on revert** to `File::open(canonical)`. |
| **Minor**: inline `serde_json::json!` response | typed `SessionFileResponse` | Fields match the web client's `Loaded` shape (`content`/`is_binary`/`truncated`). |
| **Nits**: `isAbsolutePath` reuse, `markdownPreview` unnormalized | `c9f411be` | No #2587 regression (paths are normalized before the predicate); `markdownPreview` now clamped in `normalizeSnapshot`. |

### Also checked
- Rebase onto current main is clean: endpoint routed, pane wired, all deps resolve, no conflict markers.
- `provenance_is_not_built_for_a_project_file` locks the fast-path skip (fails if the reorder regresses).

### Non-blocking (optional follow-ups)
- The new symlink test's final `assert_ne!(secret, "")` (file_provenance.rs:500) is a tautology; the real guarantee is the `unwrap_err()` above it. Assert the returned content never equals the secret if you want it to read truer.
- Pre-existing LOW polish from the prior pass remains open: distinct error state for a failed file-list fetch, focus restoration on file open/close.

Nice, thorough follow-up, especially adding the cap_std provocation test. Approving; good to merge.

src/server/api/file_provenance.rs
Seluj78: Thanks for re-verifying the reorder is decision-preserving, that was the property I most wanted a second pair of eyes on. Nothing further here.

src/server/api/file_provenance.rs
Seluj78: You were right about the tautology. Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3090/commits/87fc2eeb: instead of re-reading the secret from disk (true by construction), the test now asserts the returned content is never the secret's bytes, then pins the NOT_FOUND. Reads truer and still fails on a revert to `File::open(canonical)`.

Seluj78:
Thanks for the approval @jerome-benoit. Took the two optional follow-ups now rather than deferring them, since both were small:

**Failed file-list fetch is now distinct** - `78bca550`. `useFilesIndex` was swallowing a failed fetch into an empty array, so "could not read the list" looked identical to "this session has no files". It now surfaces an `error` flag plus a `reload` callback, and the Files pane renders its own message with a **Retry** button instead of a dead end. Tests cover both branches (error vs genuinely empty) and that Retry calls reload.

**Focus restoration** - `78bca550`. Opening a file moves focus into the viewer; closing it returns focus to the row that was opened, via `flushSync` so the row is remounted before it takes focus (the list is unmounted while the viewer is up). I verified the test is real by removing the focus call and re-running: it fails, so the behavior is pinned.

**Tautological assert** - `87fc2eeb`, per the inline thread.

Validation: `cargo fmt`/`clippy` clean (also fixed three `cloned_ref_to_slice_refs` warnings clippy only surfaces under `--features serve --tests`), `file_provenance` 9/9, full web Vitest 3579 passed.

PR #3095, opened 2026-07-24
test(hooks): cover Linux privdrop ownership guards

### LINKED ISSUE #2229: hooks: add Linux privdrop CI job to exercise alien-uid rejection paths
## Problem

`dir_guard::verify_dir_metadata` rejects the per-user hook base when its
`st_uid` does not match the running euid (issue #1844 AC3). This path is
only reached when another local user has pre-squatted
`/tmp/aoe-hooks-<our-euid>`. We have unit-test coverage of mode-rejection
(`init_rejects_dir_mode_0o755`, `init_rejects_dir_mode_0o770`,
`init_rejects_setgid/setuid/sticky_dir`) but no test that runs the same
init code under a different uid. PR #2221 explicitly deferred the
privdrop suite (4 tests in the plan §8 inventory: alien-uid base,
alien-owned per-instance subdir, dash variant under alien uid, v016
foreign legacy intact) because they need a separate CI mechanism.

## Why this matters

The wrong-uid branch in `verify_dir_metadata` is the line of defence
against an alien co-tenant on a multi-user POSIX host pre-creating the
hook base 0o700 ours-shaped but owned by themselves. Without a test that
actually exercises an alien uid, a future refactor could silently drop
the check and we would not notice until a security report.

## Suggested fix

Add a Linux-only CI job `privdrop-tests` on `ubuntu-latest` that runs
the suite inside `unshare -U` so an alien uid is observable without
sudo. Gate the matching tests with
`#[cfg(target_os = "linux")]` plus an env-var early-return that
auto-skips outside CI. Tests to add (names mirror the existing
`init_rejects_*` shape):

- `init_rejects_alien_uid_base`: pre-create `<base>` 0o700 owned by uid
  N != euid, expect `Err` containing `uid=`
- `instance_subdir_rejects_alien_uid`: same but for the per-instance
  subdir
- `host_shell_in_dash_refuses_wrong_uid`: cross-shell variant of the
  above through the host snippet
- `migration_v016_leaves_foreign_legacy_alone`: an alien-owned
  `/tmp/aoe-hooks/<their-id>` survives the v016 sweep

macOS has no user-namespace equivalent; tests skip via
`cfg(target_os = "linux")`.

## Tracking

Follow-up to #1844 / PR #2221.

shixi-li:
## Description

Closes https://github.com/agent-of-empires/agent-of-empires/issues/2229

Add the Linux ownership regressions deferred from #2221 and wire them into the stable `Cargo Test (linux)` required-check context.

The new coverage verifies that:

- an alien-owned hook base is rejected
- an alien-owned per-instance directory is rejected
- the generated dash host hook refuses an alien-owned base
- the v017 legacy sweep removes owned entries while preserving foreign entries

A dedicated `ubuntu-latest` job builds the library test binary as the normal runner user, asserts that the `privdrop_` filter names exactly four tests, and only then executes that prebuilt binary with `sudo`. The job is explicitly restricted to a disposable GitHub-hosted runner, uses read-only contents permission, and does not persist checkout credentials.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

The new tests pass 4/4. The unchecked test item reflects the disclosed Docker-only full-suite exception below; required upstream CI remains the merge-readiness authority.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## Validation

- `cargo clippy --locked -- -D warnings`
- Linux `cargo test --locked --no-run --lib`
- Linux root fixture run: 4 passed, 0 failed, 0 ignored
- workflow YAML parse and `git diff --check`
- independent adversarial review: approved with no blocking findings

`cargo test --locked` in Docker completed 4,266 tests successfully but still failed two existing locked-worktree cleanup tests: `git::worktree::tests::delete_branch_reaps_lingering_locked_worktree` and `session::trash::tests::purge_recovers_when_project_path_diverged_and_locked_entry_survives`. The same pair failed as root and uid 1000; their implementation files are byte-identical to base and outside this diff. Upstream required CI remains the merge-readiness authority.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** OpenAI GPT-5

OpenAI GPT-5 was used for implementation, validation orchestration, and PR drafting. A separate independent agent reviewed the final committed diff and validation evidence, with no blocking findings.

## Summary
- Added Linux regression coverage for rejecting directories owned by another user.
- Verified host hooks refuse unsafe ownership and legacy cleanup preserves foreign-owned entries.
- Added a dedicated CI job to run these tests with elevated privileges and included it in required Linux checks.

## Benefits
Improves protection against unsafe hook directories and prevents regressions in ownership validation and migration cleanup.

## Technical details
- Added privileged test helpers and four ownership-focused tests.
- Tests run only on Linux under `sudo` with controlled, single-threaded execution.
- CI validates the expected test count before running the suite.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -94,13 +94,63 @@ jobs:
           cargo test --no-default-features --features e2e-tests --test e2e
           cargo check --no-default-features --features default-plugins,e2e-tests --all-targets
 
+  hook-privdrop:
+    name: Hook privdrop tests
+    # The privileged step is safe only on a disposable GitHub-hosted runner.
+    # Do not move this job to a persistent or self-hosted machine.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        with:
+          cache-bin: "true"
+          save-if: true
+      - name: Build the library test binary as the runner user
+        shell: bash
+        run: |
+          set -euo pipefail
+          test_binary="$(
+            cargo test --no-run --lib --message-format=json \
+              | jq -r '
+                  select(
+                    .reason == "compiler-artifact"
+                    and .profile.test == true
+                    and .target.name == "agent_of_empires"
+                    and .executable != null
+                  )
+                  | .executable
+                ' \
+              | tail -n 1
+          )"
+          test -n "$test_binary"
+          printf '%s\n' "$test_binary" > "$RUNNER_TEMP/aoe-lib-test-binary"
+      - name: Exercise alien uid guards
+        shell: bash
+        run: |
+          set -euo pipefail
+          test_binary="$(cat "$RUNNER_TEMP/aoe-lib-test-binary")"
+          test -x "$test_binary"
+          test_count="$(
+            "$test_binary" privdrop_ --list \
+              | awk '/: test$/ { count += 1 } END { print count + 0 }'
+          )"
+          if [ "$test_count" -ne 4 ]; then
+            echo "expected 4 privdrop tests, found $test_count"
+            exit 1
+          fi
+          sudo env AOE_PRIVDROP_TESTS=1 \
+            "$test_binary" privdrop_ --test-threads=1 --nocapture
+
   cargo-linux-ok:
     # Fan-in for the branch ruleset: "Cargo Test (linux)" is a required status
     # check, and matrix jobs report per-leg contexts the ruleset doesn't know
     # about. This job carries the stable context name and fails unless every
     # leg succeeded.
     name: Cargo Test (linux)
-    needs: cargo-linux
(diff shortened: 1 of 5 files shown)


njbrake (approved):

Approving. This closes #2229 (the privdrop suite deferred from #2221) and is cleanly scoped to test and CI code (+200/-1).

I verified the load-bearing pieces rather than reading them:

- Built the lib test binary and reproduced the CI count primitive; the `privdrop_` filter names exactly 4 tests (`privdrop_test_enabled` is a helper, not a `#[test]`, so it is correctly excluded), and only those 4 symbols in the tree match.
- Ran the suite as root with `AOE_PRIVDROP_TESTS=1`: 4 passed, 0 failed.
- Ran it again without the env var (the shape the normal cargo-linux legs hit): all 4 early-return and pass trivially, so there is no double-meaningful execution and no risk to the existing legs.

The CI design is careful about the privilege boundary: build the test binary as the runner user, then run that prebuilt binary under sudo, so the compiler never runs elevated. Read-only contents permission, `persist-credentials: false`, pinned action SHAs, and the inline comment forbidding self-hosted runners all hold up. The hard-coded count of 4 is a deliberate tripwire that forces a conscious bump when a fifth privileged test is added.

One note, not blocking: CodeRabbit's title warning (`test(hooks):` vs `test:`) is a false positive against the repo's real commitlint config; the live Lint PR title check passes and recent main commits use scoped prefixes.

PR #2699, opened 2026-07-06
feat(conductor): experimental orchestrator subcommand (#553)

Talador12:
## Description

First cut of the orchestrator feature @njbrake accepted in #553 (thread [comment 4450622259](https://github.com/agent-of-empires/agent-of-empires/issues/553#issuecomment-4450622259)). Ports the [`agent-of-agent-of-empires`](https://github.com/Talador12/agent-of-agent-of-empires) TypeScript prototype into aoe as a native `aoe conductor` subcommand (with `ao` alias), gated behind `AOE_EXPERIMENTAL_AO_MODE` per the design agreed in the thread.

**Framing** (answering [@jerome-benoit, comment 4470905230](https://github.com/agent-of-empires/agent-of-empires/issues/553#issuecomment-4470905230)): the conductor is convergence-oriented, not a transport. It reads real `Instance` state, ranks sessions by an attention score, and takes non-destructive actions bounded by `ConductorPolicies`. Mechanism lives on `Instance`; policy lives on the orchestrator. Full design writeup in the new `DESIGN.md` "Conductor" section.

## Surfaces

* `aoe conductor status` (also `aoe ao status`) - ranked queue as table or JSON. Excludes archived / trashed / actively snoozed rows.
* `aoe conductor watch [--once] [--live] [--allow-nudge] [--reasoner-binary]` - tick loop calling a pluggable `Reasoner` (subprocess `claude --print`).
* `aoe conductor spawn --repo owner/repo [--limit] [--label] [--max-active]` - one session per open GitHub issue via the `gh` CLI. Dry-runs by default; `--live` to spawn. Respects a `SessionPool` cap and retries with exponential backoff.
* **TUI panel** reachable via the command palette (search "conductor" / "orchestrator" / "attention" / "queue"). Display-only in this first cut; interactive tick + apply are labeled follow-up.
* **`GET /api/conductor/state`** on the daemon (serve-gated). Returns the same ranked queue as JSON; 403 with an opt-in hint when the gate is off.
* **Remote TUI overlay** (`AOE_DAEMON_URL` mode) - press `c` in the remote picker to fetch conductor state from the daemon and see the same ranked table.

## Guardrails

* Whole feature is off unless `AOE_EXPERIMENTAL_AO_MODE=1`. Every surface fails closed with the same hint.
* `ConductorPolicies` defaults to `allow_destructive: false`, `allow_nudge: false`. `--live` alone does not enable nudges.
[...]

Issue #553, opened 2026-03-28
Orchestrator agent


njbrake:
Look into adding a feature where we have an agent that sits above all of the individual sessions and can view the state of all of them and help recommend action actions so that theoretically this agent can kind of be the controller of everything that's happening in aoe could be a nice pro feature

Seluj78:
This would be an amazing feature (and I would pay for it for the right price ;) ) 

I am imagining one or two different agents:

- One for orchestrating and recommending actions across all running sessions
- One for backlog/ticket preparation/creation, like [kanban](https://github.com/cline/kanban) does which could be pretty nice

dirien:
ngl, this would be amazing. I saw that this project uses this approach https://agent-deck.devopstoolkit.ai/ and it would be really cool, if every group could have their own orchestrator agent.

njbrake:
Stumbled onto this project from @Talador12 https://github.com/Talador12/agent-of-agent-of-empires

Incredible name and also might be relevant for adding an orchestrator feature into aoe

Seluj78:
That's amazing !

Would it be kept as a separate tool or merged into `aoe` ?

Talador12:
I would love to help contribute this as a new feature, and you can take the extended name if you want it👍

njbrake:
Before doing any implementation I would be interested in hearing how you would propose adding it as a feature: would it be a command inside AoE, something like 'aoe conductor'? Or would it be a plugin and a separate binary to install? Would it be accessible via the TUI and the web dashboard, or just CLI?

I'm open to whatever way gets this feature available for the AoE community to try and get benefit from. I assume it would best to put it into this repo but if you think some sort of plugin registry system makes more sense where the code stays in your aoaoe repo I'm also open to whatever idea.

Talador12:
1st option: My vote is `aoe conductor` (name TBD?) as a proper subcommand built into this repo. I'm fine with aoaoe going away if the feature lands here natively and that is probably the better outcome for the community anyway. A real "conductor" panel and a dashboard route would make it feel like a first-class feature. I iterated on that section the most on aoaoe, since there is a lot you can show. On the aoe session list, it could appear as a top level item, or something similar so it is visible.

2nd option: wrapper to aoaoe, but I do not think that makes sense based on the above.

Seluj78:
[...]

PR #3108, opened 2026-07-25
feat: add featured aoaoe orchestrator plugin entry (8.0.0)

Talador12:
## Description

Adds the **aoaoe orchestrator** plugin ([`gh:Talador12/agent-of-agent-of-empires`](https://github.com/Talador12/agent-of-agent-of-empires), id `dev.talador12.aoaoe`) to the featured index, pinning release [v8.0.0](https://github.com/Talador12/agent-of-agent-of-empires/releases/tag/v8.0.0) by its source tree hash.

This is the plugin form of the orchestrator from #553, delivered per the decision on #2699 to ship orchestration as an installable plugin living in my repo rather than inside aoe core.

## What the plugin does

- **Attention queue** — ranks every session by who needs eyes next (status + time-in-status escalation): dashboard card, sortable "Attention" column via `row-column`/`sort-key`, row badges, per-session pane with tick/pause actions, status-bar segment, deduplicated notifications for urgent sessions.
- **Spawn from GitHub issues** — one session per open issue in a configured repo via `sessions.create`, seeded with the issue as `initial_turn`, `idempotency_key = "issue:<repo>#<n>"` so reruns never double-spawn. Dry-run by default.
- **Reasoners** — deterministic `rules` engine by default (no LLM, no tokens); `claude-code` / `opencode` opt-in.
- **Guardrails** — `dry_run=true` and `allow_nudge=false` by default; nudges only ever go to plugin-created sessions (host-enforced ownership); 5s tick floor; per-session in-batch action cooldowns; quiet hours. Does **not** request `session.unattended`.

Design writeup (mechanism-in-aoe vs policy-in-plugin boundary, and the host primitives it does *not* work around): [DESIGN.md](https://github.com/Talador12/agent-of-agent-of-empires/blob/v8.0.0/DESIGN.md).

## Hash provenance

```
$ git clone --depth 1 --branch v8.0.0 https://github.com/Talador12/agent-of-agent-of-empires
$ aoe plugin hash agent-of-agent-of-empires
sha256:38c1c6708edea12da6128c5f7e466a6f78909314ec6e4194b9cfadacf3341dd7
```

Generated with `aoe plugin hash` built from current main (1.13.1) on Linux, LF endings, clean checkout. Build outputs go to `.aoe-build/` only (worker: `node .aoe-build/dist/plugin/worker.js`), so the install cannot trip the load-time tree hash the way plugin-github 1.1.0 did.

## Verification done

- Manifest parses against `aoe-plugin-api` (`PluginManifest::from_toml_str`), api_version 11, capabilities: `runtime.worker, session.read, notifications, acp.capabilities.read, session.create, session.prompt, process.spawn, net`.
- `aoe plugin install gh:Talador12/agent-of-agent-of-empires --yes` on aoe 1.13.1: fetches v8.0.0, runs the build, records grants, plugin lists as enabled, commands graft onto the CLI/palette.
- Worker protocol exercised end-to-end against a scripted host (JSON-RPC over stdio): `sessions.list` → ranked queue → `ui.state.set` pushes for all six declared slots + `events.publish("queue.updated")`.
- 5098 tests green in the plugin repo, including a fake-host contract test for the worker.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (N/A — data-only change to `plugins/featured.toml`, no UI code touched)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Fable 5) under my direction.

The plugin implementation, the hash provenance steps, and this PR were prepared with Claude Code; I direct the work and review what it produces. Reviewer discussion will be answered by me (Talador12) directly.

## Summary

- Added `dev.talador12.aoaoe` to the featured plugin index.
- Pinned the plugin to the vetted `v8.0.0` release using its source tree hash.
- Enables installation of the orchestrator as a standalone plugin, including attention ranking, issue-based sessions, notifications, configurable reasoning, and safety guardrails.

## Benefits

- Makes the orchestrator discoverable and installable through the featured registry.
- Ensures reproducible installations from a verified release.
- Keeps orchestrator functionality separate from aoe core.

### DIFF
--- a/plugins/featured.toml
+++ b/plugins/featured.toml
@@ -37,3 +37,7 @@ versions = {"0.1.0" = "sha256:6b506512d4efbc8a70c4d989188e701e547f5c35bd690c2827
 [plugins."agent-of-empires.cron"]
 source = "gh:agent-of-empires/plugin-cron"
 versions = {"1.0.0" = "sha256:b98ab9304a9906bd33acdcc5b4c1f5e9d6eb44db442d7409f82dcd8b0eeac2ac"}
+
+[plugins."dev.talador12.aoaoe"]
+source = "gh:Talador12/agent-of-agent-of-empires"
+versions = {"8.0.0" = "sha256:38c1c6708edea12da6128c5f7e466a6f78909314ec6e4194b9cfadacf3341dd7"}


njbrake:
Copied from a discord conversation:


Thank you for the PR! Yes i'm wondering if maybe instead of hardcoding the reference into AOE itself we can improve the plugin discovery mechanism in the TUI and Web UI so that the plugins show up faster: since there aren't many plugins yet, this one would naturally show up near the top because it's one of the first ones, and then as download counts increase if we automatically sort by 'number of downlaods' then this one would still remain usable.

Aka I think in aoe we should make sure we have a way to help promote and encourage the use of plugins, without the need to put a reference to any specific plugins into the source code.

I think this means that we actually may want to pursue a design where we pull out the existing things that are hard coded into the code, like the cron job.

If we sharpen our plugin design a bit more, the promotion and discoverability of plugins shouldn't require us to hard code references into the source code, and this could apply not only to plugins like aoaoe, but also to the 'official' aoe plugins like github/cron/etc.

Issue #2573, opened 2026-06-30
Cockpit: sessions stuck 'running' and transcript drift around agent-initiated (async Agent) work and worker respawns


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Structured-view / web sessions get stuck green ("running") with no further output, indefinitely, until a human sends another message. The underlying defect is upstream (the adapter never emits a `PromptResponse`, see upstream `agentclientprotocol/claude-agent-acp#825`), but aoe makes it worse than it has to be: when the wedged turn involved any off-protocol / background work, aoe's own silent-orphan watchdog is floored at 30 minutes, so the session does not auto-recover for half an hour. In one case aoe even detects the stall itself and still leaves the session green.

This is the aoe-side mitigation tracking issue while upstream #825 is open. Three live sessions reproduced it; all three share the same end state (open turn, no terminal `Stopped`), with different triggers.

## Evidence from the affected sessions

| Session | Last events | Model conn | Trigger |
|---|---|---|---|
| `d40acff8` | partial `AgentMessageChunk`, cut mid-sentence (seq 1237) | gone | model stream dropped mid-stream |
| `48984ed9` | `BackgroundAgentProgress` → `BackgroundAgentCompleted{status:"completed"}` (seq 421) | alive (`160.79.104.10`) | async `Agent` (`isAsync`) finished, host turn never resolved |
| `7d790b5a` | `BackgroundAgentCompleted{completed}` (538), `BackgroundAgentCompleted{status:"stalled"}` (539) | gone | async `Agent` stalled; prompt loop poll never returned |

Common to all three: the main prompt already got its `prompt_complete` Stopped, then async/background work left the turn with no terminal. The `claude` SDK process is idle in `kevent` (not blocked on a syscall), so no `PromptResponse` ever reaches the daemon, so aoe never synthesizes a `Stopped`. Recovery only happens on the next `session/prompt` (verified on `d40acff8`: a `"continue"` prompt advanced the stream instantly from seq 1237 to 1356, abandoning the stalled turn's tail).

## Root cause (aoe side)

The true fix is upstream #825. aoe's amplifiers:

[...]

Issue #317, opened 2026-02-23
[Feature]: Bookmark/favorite directories for quick session creation


hansonkim:
## Problem

When working with multiple projects, creating a new session requires typing the full directory path every time. For users who frequently switch between the same set of projects, this becomes repetitive and slows down the workflow.

## Proposed Solution

Add a "bookmark" or "favorite directories" feature that allows users to save frequently used project paths and quickly select from them when creating a new session.

### Possible UX

- **TUI**: When pressing `n` to create a session, show a list of bookmarked directories to choose from (in addition to manual path input)
- **CLI**: `aoe bookmark add /path/to/project` / `aoe bookmark list` / `aoe bookmark remove <name>`
- **Shortcut**: A dedicated key (e.g., `b`) in the TUI to manage bookmarks

### Possible Storage

Store bookmarks in `config.toml` or a separate `bookmarks.json` under `~/.agent-of-empires/`:
```toml
[[bookmarks]]
name = "my-api"
path = "/home/user/projects/my-api"

[[bookmarks]]
name = "frontend"
path = "/home/user/projects/frontend-app"
```

## Alternatives Considered

- **Profiles** partially address this by organizing sessions, but don't help with quickly targeting a directory for a new session.
- **Shell aliases** (`alias aoe-api="aoe add /path/to/api"`) work but aren't integrated into the TUI experience.

## Additional Context

I'm willing to contribute a PR if this feature aligns with the project direction. Happy to discuss the implementation approach first.

PR #3132, opened 2026-07-27
fix(acp): drive a real conversation reset for codex /new

### LINKED ISSUE #2979: Structured Codex: /new shows Conversation cleared but context and model memory are not reset
(Note: issue discovered by me, description written by Codex)

## Problem

In a web structured Codex session, sending `/new` appears to be recognized by AoE because the structured view renders the `Conversation cleared` boundary. However, the underlying Codex conversation does not actually appear to reset.

Two concrete signs:

- The context tracker stays at the exact same high value after `/new`.
- The Codex session still remembers context from before `/new`.

So this does not look like the old missing-UI-boundary bug. AoE now shows the clear boundary, but the model memory/context accounting remain unchanged.

## Steps to reproduce

1. Create or open a web structured Codex session.
2. Build up some recognizable context in the conversation.
3. Note the context tracker value.
4. Send `/new`.
5. Observe that the web structured view says `Conversation cleared`.
6. Observe the context tracker value.
7. Ask a follow-up question that should prove whether the old context is gone.

## Expected

`/new` should be a real Codex conversation reset in structured/web mode:

- Codex should no longer remember pre-`/new` context.
- The context tracker should reset or drop to the post-reset baseline.
- AoE's `Conversation cleared` UI should correspond to an actual adapter/session clear, not only a local boundary marker.

## Actual

- AoE renders `Conversation cleared`.
- The context tracker stays at the same high value.
- Codex still remembers prior conversation context.

See screenshot here for evidence https://imgur.com/a/WXz2XlE

## Why this matters

The UI currently communicates that the conversation was cleared, but the model behavior and context tracker indicate that it was not. That makes `/new` unsafe to rely on when the user needs a clean slate in a structured Codex session.

## Notes / likely area

This may mean AoE is emitting or handling `SessionCleared` based on the `/new` prompt text before the adapter has actually reset the Codex session. Structured Codex may need a stronger reset path than passing `/new` through as text, for example an adapter-supported reset, a worker/session restart, or a new ACP session.

This is specifically about Codex structured/web behavior. Terminal Codex behavior may differ.

## Environment

- AoE: `1.11.2`
- Surface: web structured view
- Agent: Codex via `codex-acp`
- `codex-acp --version` is not available in this install.

## Related

- #1134 - Codex parity issue that called out `/new` as Codex's reset shortcut.
- #1101 - `/clear` UI/session-boundary behavior for cockpit.
- #1354 - Related usage/context accounting after clear boundaries.

shixi-li:
## Description

Closes https://github.com/agent-of-empires/agent-of-empires/issues/2979

In web structured Codex sessions, sending `/new` renders the "Conversation cleared" boundary but the underlying Codex conversation never resets: the supervisor emits only the UI-level `SessionCleared` event and forwards the raw `/new` text to codex-acp, which treats it as an unknown command, so context persists. Claude is unaffected because it handles `/clear` natively.

This implements the issue's Proposal A, Codex-scoped. Upstream declined `/new` handling in codex-acp (agentclientprotocol/codex-acp#317: "just have the client create new sessions via session/new"), so the reset is driven client-side:

- `agent_profiles.rs`: new `clear_requires_driven_reset` flag, `true` only for CODEX. Claude stays `false` (native `/clear` text-forward unchanged); OpenCode/OMP/Kimi stay `false` pending the issue's open generalization question.
- `supervisor.rs`: `publish_user_prompt_with_attachments` returns a `PromptDisposition` (`Forward` | `ResetContext`); `reset_session_context` drives the reset then re-asserts the persisted `acp_mode_id`/yolo mode and re-applies the spawn-time `default_effort`/`default_mode` config options on the fresh session — answering the open question about mode/config-option survival.
- `acp_client.rs`: new `ClientCmd::ResetSession` — the connection task opens a fresh `session/new` on the live worker (no restart), swaps the stored ACP session id, and emits `SessionContextReset` → `AcpSessionAssigned` (ordering keeps the fresh id persisted for the #3083 listener) → `Stopped(session_reset)`. Mid-turn resets are refused with a `PromptRejected` frame (parity with the busy-prompt path); failures surface as `PromptRuntimeError` + `Stopped(session_reset_failed)`.
- `session_service.rs`: `send_turn` branches on the disposition, so a clear alias on Codex never reaches `send_prompt`.
- `process/runner.rs`: the reset goes over the byte relay (the v2 control channel replays the cached handshake and would resurrect the old id); the runner tracks relay `session/new` requests and refreshes its handshake cache so `Cancel`/reattach replays follow the fresh session.
- `web/src/lib/acpTypes.ts`: `SessionContextReset` marks the turn as having output so the reset divider doesn't add a spurious "Command produced no output." row.

Failing baseline captured first on the unfixed code: the fake codex agent received one `session/new` plus a `session/prompt` carrying `/new` addressed to the old session id.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt --check`, clippy (`-D warnings` on the CI feature set), `acp::` + `server::` suites (1002 tests), web fmt/lint and the web suite all pass. The full `--lib --bins` run has 7 failures in `git::worktree`/`session::trash` that reproduce identically at the base commit on this machine (localized git output), i.e. pre-existing and unrelated. The visible change is an event-stream boundary already covered by the reducer test rather than a styling change, so no screenshot is attached.

## Test Coverage Analysis

`web/tests/live/acp-codex-new-reset.spec.ts`: real `aoe serve` with a fake codex ACP agent through the real runner — `/new` yields `SessionCleared`, `SessionContextReset`, a second `AcpSessionAssigned` with a new id, and `Stopped(session_reset)`. The vitest reducer test asserts the context tracker drops to the post-reset baseline.

**TUI / CLI (e2e test)**

The reset path is exercised end-to-end by the scripted-agent Rust integration tests (two `session/new` calls on the wire, id swap, alias never forwarded, `default_effort` re-applied after reset, mid-turn refusal with `PromptRejected`) and the live web spec above.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Fable 5 (Claude Code)

Implementation and tests were AI-authored under human direction with an independent adversarial review pass before submission; validation commands and their results are listed above.

## Summary

- Fixed Codex `/new` in structured web sessions to trigger a real ACP “conversation reset” (`session/new`) instead of sending `/new` as plain prompt text.
- Implemented a Codex-only “driven reset” path for clear aliases where the adapter can’t reset natively, so the app drives the reset by resetting the underlying ACP session/context.
- Added safeguards to reject reset attempts when the session is busy (e.g., an in-flight prompt or active tool/background work), returning clear failure outcomes.
- Ensured the reset truly replaces the underlying ACP session id, then reapplies the persisted session mode and default configuration so subsequent turns continue with the expected settings.
- Updated relay/session tracking so the system doesn’t reuse stale session state after a reset (including runner handshake cache refresh).
- Refined the event/UI reset boundary so the timeline and “reset boundary visibility” match an actual reset sequence.
- Added coverage: Rust unit/integration tests, web reducer tests, and a live Playwright test validating `/new` produces the expected reset event order and a new ACP session id (plus a coverage-matrix entry).

## Benefits

Codex “Conversation cleared” in the web UI now corresponds to an actual reset of model conversation state, preserves the user’s configured session settings through the reset, and keeps the web event timeline consistent with what the ACP worker really did.

## Further considerations

The driven-reset behavior is currently focused on Codex; extending the same mechanism to other adapters that can’t reset natively could improve consistency across agents.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -339,6 +339,39 @@ const ACP_SESSION_DELETE_TIMEOUT: std::time::Duration = std::time::Duration::fro
 /// adapter from stalling the prompt path that requested the reset.
 const SESSION_RESET_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 
+/// Inner deadline for the connection task's complete reset RPC sequence.
+/// Kept below `SESSION_RESET_TIMEOUT` so the task can report the specific
+/// failure before the caller's outer guard expires, then resume draining
+/// commands instead of remaining parked on a wedged adapter.
+const SESSION_RESET_IN_TASK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(28);
+
+#[derive(Debug)]
+enum ResetRequestError {
+    Acp(agent_client_protocol::Error),
+    TimedOut,
+}
+
+async fn await_reset_request<T, F, Fut>(
+    deadline: tokio::time::Instant,
+    request: F,
+) -> Result<T, ResetRequestError>
+where
+    F: FnOnce() -> Fut,
+    Fut: std::future::Future<Output = Result<T, agent_client_protocol::Error>>,
+{
+    // `send_request` enqueues the stateful RPC synchronously. Keep it
+    // lazy so a command whose caller-created deadline expired in the
+    // queue cannot send a late session/new or config mutation at all.
+    if tokio::time::Instant::now() >= deadline {
+        return Err(ResetRequestError::TimedOut);
+    }
+    match tokio::time::timeout_at(deadline, request()).await {
+        Ok(Ok(response)) => Ok(response),
+        Ok(Err(error)) => Err(ResetRequestError::Acp(error)),
+        Err(_) => Err(ResetRequestError::TimedOut),
+    }
+}
+
 /// Configuration for spawning an ACP agent.
 #[derive(Debug, Clone)]
 pub struct SpawnConfig {
@@ -448,14 +481,19 @@ enum ClientCmd {
     },
     /// Drive a real conversation reset: issue a fresh `session/new` on
     /// the live connection, swap the task's ACP session id to the new
-    /// one, and emit `SessionContextReset` + `AcpSessionAssigned` + a
-    /// terminal `Stopped` so bookkeeping and the UI follow. Issued by
-    /// the supervisor when a clear command hits a profile whose adapter
-    /// has no native reset (codex `/new`). `text` carries the user's
-    /// original clear invocation, used only by the mid-turn refusal's
-    /// `PromptRejected` so the retry pill shows what was typed. See #2979.
+    /// one, and emit `SessionCleared` + `SessionContextReset` +
+    /// `AcpSessionAssigned` + a terminal `Stopped` so bookkeeping and
+    /// the UI follow. Issued by the supervisor when a clear command hits
+    /// a profile whose adapter has no native reset (codex `/new`). `text`
+    /// carries the user's original clear invocation, used only by the
+    /// mid-turn refusal's `PromptRejected` so the retry pill shows what
+    /// was typed. See #2979.
     ResetSession {
         text: String,
+        /// Absolute deadline created by the caller. Starting it before the
+        /// command is queued prevents a delayed command from resetting the
+        /// session after the caller has already timed out.
+        deadline: tokio::time::Instant,
         respond_to: oneshot::Sender<ResetSessionOutcome>,
     },
     Shutdown,
@@ -2109,6 +2147,43 @@ impl AcpClient {
         (client, event_tx, cmds)
     }
 
+    /// Like `fake_for_test_cmd_recording`, but answers a driven reset
+    /// with a deterministic failure. Used by supervisor tests to assert
+    /// that a busy or otherwise rejected reset never publishes the
+    /// successful `SessionCleared` boundary.
+    #[cfg(test)]
+    pub fn fake_for_test_reset_failure(
+        session_id: AcpSessionId,
+        message: impl Into<String>,
+    ) -> (Self, mpsc::Sender<Event>) {
+        let (event_tx, event_rx) = mpsc::channel(64);
+        let (cmd_tx, mut cmd_rx) = mpsc::channel::<ClientCmd>(16);
+        let message = message.into();
+        tokio::spawn(async move {
+            while let Some(cmd) = cmd_rx.recv().await {
+                match cmd {
+                    ClientCmd::ResetSession { respond_to, .. } => {
+                        let _ = respond_to.send(ResetSessionOutcome::Failed {
+                            message: message.clone(),
+                        });
+                    }
+                    ClientCmd::DeleteSession { respond_to, .. } => {
+                        let _ = respond_to.send(DeleteSessionOutcome::UnsupportedMethod);
+                    }
+                    _ => {}
+                }
+            }
+        });
+        let client = Self {
+            session_id,
+            inbound: Some(event_rx),
+            cmd_tx: Some(cmd_tx),
+            pending_responders: Arc::new(Mutex::new(HashMap::new())),
+            _child: None,
+        };
+        (client, event_tx)
+    }
+
     /// Spawn an ACP agent subprocess, run the handshake + create a
     /// session, and start pumping notifications into the inbound channel.
     pub async fn spawn(config: SpawnConfig, session_id: AcpSessionId) -> Result<Self, AcpError> {
@@ -2773,22 +2848,27 @@ impl AcpClient {
 
     /// Drive a real conversation reset on the live worker: the connection
     /// task issues a fresh `session/new`, swaps its ACP session id, and
-    /// emits `SessionContextReset` + `AcpSessionAssigned` + a terminal
-    /// `Stopped`. Used for clear commands whose adapter has no native
-    /// reset (codex `/new`, #2979). `text` is the user's clear invocation,
-    /// surfaced in the mid-turn refusal's `PromptRejected`. Bounded so a
-    /// wedged adapter cannot stall the prompt path; the timeout surfaces
-    /// as a `Failed` outcome.
+    /// emits `SessionCleared` + `SessionContextReset` +
+    /// `AcpSessionAssigned` + a terminal `Stopped`. Used for clear
+    /// commands whose adapter has no native reset (codex `/new`, #2979).
+    /// `text` is the user's clear invocation, surfaced in the mid-turn
+    /// refusal's `PromptRejected`. Bounded so a wedged adapter cannot
+    /// stall the prompt path; a `session/new` timeout surfaces as a
+    /// `Failed` outcome, while post-reset config re-application remains
+    /// best-effort.
     pub async fn reset_session(&self, text: &str) -> Result<ResetSessionOutcome, AcpError> {
         let cmd_tx = self.cmd_tx.as_ref().ok_or(AcpError::NotRunning)?;
         let (tx, rx) = oneshot::channel();
+        let deadline = tokio::time::Instant::now() + SESSION_RESET_IN_TASK_TIMEOUT;
         // Mirror `delete_session`: the guard wraps BOTH the cmd_tx send
         // and the response wait so a wedged connection task cannot park
-        // the caller indefinitely.
+        // the caller indefinitely. The connection task receives the
+        // caller-created inner deadline so queueing time counts too.
         let request = async {
             if cmd_tx
                 .send(ClientCmd::ResetSession {
                     text: text.to_string(),
+                    deadline,
                     respond_to: tx,
                 })
                 .await
@@ -7639,7 +7719,9 @@ async fn run_connection_task<W, R>(
                                                 break;
                                             }
                                         }
-                                        Some(ClientCmd::ResetSession { text, respond_to }) => {
+                                        Some(ClientCmd::ResetSession {
+                                            text, respond_to, ..
+                                        }) => {
                                             // Resetting under an in-flight
                                             // `session/prompt` would orphan
                                             // the pending turn on the old
@@ -7650,8 +7732,8 @@ async fn run_connection_task<W, R>(
                                             // `PromptRejected` so the caller
                                             // gets a terminal frame (retry
                                             // pill) under the persisted
-                                            // UserPromptSent + divider, not
-                                            // just an HTTP error. See #2979.
+                                            // UserPromptSent, not just an
+                                            // HTTP error. See #2979.
                                             warn!(
                                                 target: "acp.protocol",
                                                 "conversation reset requested during in-flight prompt; refusing"
@@ -7862,7 +7944,11 @@ async fn run_connection_task<W, R>(
                             event_tx_for_block.clone(),
                         );
                     }
-                    Some(ClientCmd::ResetSession { text: _, respond_to }) => {
+                    Some(ClientCmd::ResetSession {
+                        text: _,
+                        deadline: reset_deadline,
+                        respond_to,
+                    }) => {
                         // Driven conversation reset (#2979): a clear command
                         // hit a profile whose adapter has no native reset
                         // (codex `/new`), so open a genuinely fresh session
@@ -7876,6 +7962,10 @@ async fn run_connection_task<W, R>(
                         // runner watches for it and refreshes its own
                         // handshake cache from the response (see
                         // `process/runner.rs`).
+                        // Use the caller-created shared deadline for
+                        // session/new and both config re-application
+                        // requests. Queueing time counts, and per-request
+                        // deadlines cannot accumulate past the outer guard.
                         info!(
                             target: "acp.protocol",
                             session = %session_label,
@@ -7884,25 +7974,48 @@ async fn run_connection_task<W, R>(
                         );
                         let req = NewSessionRequest::new(agent_cwd.clone())
                             .mcp_servers(mcp_servers_for_reset.clone());
-                        match connection.send_request(req).block_task().await {
+                        match await_reset_request(
+                            reset_deadline,
+                            || connection.send_request(req).block_task(),
+                        )
+                        .await
+                        {
                             Ok(new_session)
                                 if new_session.session_id.0 != acp_session_id.0 =>
                             {
                                 let new_id = new_session.session_id.clone();
+                                // session/new is the irreversible reset
+                                // commit. Adopt its id before attempting
+                                // best-effort config restoration so this
+                                // client and the runner cannot disagree
+                                // about which session owns later prompts.
                                 acp_session_id = new_id.clone();
+                                available_mode_ids =
+                                    new_session.modes.as_ref().map(|modes| {
+                                        modes
+                                            .available_modes
+                                            .iter()
+                                            .map(|m| m.id.0.to_string())
+                                            .collect()
+                                    });
+                                mode_config_option_id = new_session
+                                    .config_options
+                                    .as_deref()
+                                    .and_then(mode_config_id)
+                                    .map(|id| id.0.to_string());
                                 info!(
                                     target: "acp.protocol",
                                     session = %session_label,
                                     new_id = %new_id.0,
                                     "conversation reset: session/new succeeded, swapped acp_session_id"
                                 );
-                                // Ordering matters: `SessionContextReset`
-                                // first (listeners null the stored resume id
-                                // and drop fork/import markers), then
-                                // `AcpSessionAssigned` persists the fresh id
-                                // on top, then a terminal `Stopped` ends the
-                                // clear command's synthetic turn so the
-                                // composer unlocks.
+                                // Keep every success boundary on the same
+                                // FIFO. SessionCleared folds the transcript
+                                // only after session/new committed, then
+                                // SessionContextReset clears the old resume
+                                // id before AcpSessionAssigned persists the
+                                // fresh one.
+                                let _ = event_tx_for_block.send(Event::SessionCleared).await;
                                 let _ = event_tx_for_block
                                     .send(Event::SessionContextReset {
                                         reason: "conversation cleared; the agent started a fresh session".into(),
@@ -7918,13 +8031,6 @@ async fn run_connection_task<W, R>(
                                 // the new session starts on adapter defaults,
                                 // so the old session's picker state is stale.
                                 if let Some(modes) = &new_session.modes {
-                                    available_mode_ids = Some(
-                                        modes
-                                            .available_modes
-                                            .iter()
-                                            .map(|m| m.id.0.to_string())
-                                            .collect(),
-                                    );
                                     let infos: Vec<ModeInfo> = modes
                                         .available_modes
                                         .iter()
@@ -7944,11 +8050,6 @@ async fn run_connection_task<W, R>(
                                         })
                                         .await;
                                 }
-                                mode_config_option_id = new_session
-                                    .config_options
-                                    .as_deref()
-                                    .and_then(mode_config_id)
-                                    .map(|id| id.0.to_string());
                                 if let Some(event) =
                                     config_options_event(new_session.config_options.clone())
                                 {
@@ -7982,14 +8083,19 @@ async fn run_connection_task<W, R>(
                                         );
                                         continue;
                                     };
-                                    match connection
-                                        .send_request(SetSessionConfigOptionRequest::new(
-                                            acp_session_id.clone(),
-                                            config_id,
-                                            SessionConfigValueId::new(value.to_string()),
-                                        ))
-                                        .block_task()
-                                        .await
+                                    match await_reset_request(
+                                        reset_deadline,
+                                        || {
+                                            connection
+                                                .send_request(SetSessionConfigOptionRequest::new(
+                                                new_id.clone(),
+                                                config_id,
+                                                SessionConfigValueId::new(value.to_string()),
+                                            ))
+                                                .block_task()
+                                        },
+                                    )
+                                    .await
                                     {
                                         Ok(resp) => {
                                             if let Some(event) = config_options_event(Some(
@@ -7999,13 +8105,23 @@ async fn run_connection_task<W, R>(
                                                     event_tx_for_block.send(event).await;
                                             }
                                         }
-                                        Err(e) => {
+                                        Err(ResetRequestError::Acp(e)) => {
                                             warn!(
                                                 target: "acp.protocol",
                                                 session = %session_label,
                                                 value,
                                                 "re-applying structured view default after reset failed: {e}"
                                             );
+                                        }
+                                        Err(ResetRequestError::TimedOut) => {
+                                            warn!(
+                                                target: "acp.protocol",
+                                                session = %session_label,
+                                                value,
+                                                timeout_secs = SESSION_RESET_IN_TASK_TIMEOUT.as_secs(),
+                                                "post-reset config re-application timed out; skipping remaining defaults"
+                                            );
+                                            break;
                                         }
                                     }
                                 }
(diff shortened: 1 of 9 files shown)


src/acp/acp_client.rs
```diff
@@ -7695,6 +7862,209 @@ async fn run_connection_task<W, R>(
...
+                                        Err(e) => {
+                                            warn!(
+                                                target: "acp.protocol",
+                                                session = %session_label,
+                                                value,
+                                                "re-applying structured view default after reset failed: {e}"
+                                            );
+                                        }
+                                    }
+                                }
```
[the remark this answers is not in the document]
shixi-li: Fixed in 40863d5.

The caller now creates one absolute 28s deadline before enqueueing `ResetSession`; the connection task uses that same deadline for `session/new` and each `SetSessionConfigOptionRequest`. Request creation is lazy, so a command that expired in the queue cannot send a late stateful RPC. This leaves the outer 30s guard time to receive the classified result, and the loop resumes after a stalled adapter.

A `session/new` timeout is a reset failure. A config re-apply timeout after `session/new` committed is intentionally bounded best-effort: the new session/id already exists and cannot be rolled back, so the client warns, skips remaining defaults, adopts the authoritative new id, and returns reset success instead of diverging from the runner.

Added real scripted-adapter regressions for expired commands sending no RPC, stalled `session/new` followed by a successful later reset, and stalled config re-apply followed by a prompt on the fresh id. All focused tests pass.

src/server/session_service.rs
```diff
@@ -440,10 +440,28 @@ impl SessionService {
...
+        let outcome = match disposition {
+            crate::acp::supervisor::PromptDisposition::Forward => {
+                self.acp_supervisor.send_prompt(id, text, attachments).await
+            }
+            crate::acp::supervisor::PromptDisposition::ResetContext => {
+                self.acp_supervisor
+                    .reset_session_context(id, text, acp_mode_id.as_deref(), yolo_mode)
+                    .await
+            }
+        };
```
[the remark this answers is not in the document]
shixi-li: Fixed in 40863d5.

Native clear commands keep their existing immediate boundary. For driven resets, prompt publication now records only `UserPromptSent`; the connection task emits `SessionCleared` only after a fresh `session/new` has succeeded, on the same FIFO immediately before `SessionContextReset` and `AcpSessionAssigned`. Busy, same-id, RPC-error, and `session/new` timeout paths never emit the clear boundary and preserve their rejection/error result.

Coverage includes both the real in-flight busy path and a supervisor-level failed-reset regression asserting that no `SessionCleared` is published. The focused Rust tests and the live Codex `/new` Playwright spec pass.

web/tests/live/acp-codex-new-reset.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+          const res = await fetch(`${serve.baseUrl}/api/sessions/${sessionId}/acp/prompt`, {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ text: "/new" }),
+          });
+          return res.status;
+        },
+        { timeout: 30_000, intervals: [500] },
+      )
+      .toBe(202);
```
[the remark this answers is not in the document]
shixi-li: Fixed in 40863d5. The stateful `/new` prompt is now submitted with exactly one `fetch`, followed by a direct `202` assertion. The existing `AcpSessionAssigned` poll remains the readiness oracle, and the original headers and payload are unchanged. The updated live Playwright spec passes against the rebuilt branch binary.

src/acp/acp_client.rs
```diff
@@ -7695,6 +7944,268 @@ async fn run_connection_task<W, R>(
                             event_tx_for_block.clone(),
                         );
                     }
+                    Some(ClientCmd::ResetSession {
```
Seluj78, line 8004: **Required:** this success arm does not treat in flight background agents or open tools as busy.

The arm proceeds straight to `session/new` and swaps `acp_session_id` without consulting `between_prompt_bg_agents` or `between_prompt_tools`. The between prompt idle watchdog at line 7101 treats a non-empty `between_prompt_bg_agents` as "work in flight just like an open tool" (#2573), and a launched async background agent's tailer (`spawn_tailer`, line 6209) keeps emitting `BackgroundAgentProgress` and `BackgroundAgentCompleted` into the same `event_tx` until it self terminates on completion, hard idle, or a closed `event_tx`. A reset does none of those.

Compounding it: `on_receive_notification` (line 5968 onward) never reads `notification.session_id`, so it cannot distinguish pre reset from post reset traffic. That was safe while one connection meant one ACP session for its lifetime, which this PR changes.

Failure scenario: an async background agent is running with no `session/prompt` in flight, which is the state this tracker exists to represent. The user sends `/new`. The busy refusal at line 7722 only guards the inner `select!` that runs while `prompt_fut` is pinned, so nothing is busy by that check and the reset is accepted. `session/new` succeeds, the id swaps, and the full boundary fires. The old session's tailer keeps pushing events after the boundary, where they are attributed to the fresh conversation.

Same class as the busy prompt case already fixed in `40863d53`; the second category of busy did not get the same guard.

src/acp/acp_client.rs
```diff
@@ -10302,6 +10821,467 @@ mod tests {
...
+        assert_eq!(
+            wire.matches("\"value\":\"high\"").count(),
+            2,
+            "both applications must carry the configured effort value;\nwire capture:\n{wire}"
+        );
+        let _ = client.shutdown().await;
+    }
+
+    /// #2979: a reset requested while a `session/prompt` is in flight must
+    /// be refused — resetting under the turn would orphan it on the old
```
Seluj78: **Required:** em dash on this line and 11208.

Rule: AGENTS.md forbids em dashes as a separator in docs and comments. Use commas, semicolons, or rephrase.

Context for the maintainers rather than a hard gate: `src/` already carries 121 such lines across 39 files at the base commit and no CI job gates on it, so these match surrounding style even though they contradict the written rule.

src/acp/supervisor.rs
```diff
@@ -169,6 +171,22 @@ pub enum SupervisorError {
...
+/// `publish_user_prompt_with_attachments` recorded it. The publish step
+/// owns clear-command detection (it already resolves the session's
+/// `AgentProfile`), so it also owns the routing decision. See #2979.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PromptDisposition {
+    /// Forward the text to the agent as an ordinary `session/prompt`.
+    Forward,
+    /// The text is a clear command for a profile whose adapter has no
+    /// native reset handler (`clear_requires_driven_reset`, e.g. codex's
+    /// `/new`): do NOT forward it — codex-acp would swallow it as an
```
Seluj78: **Required:** em dash on this line and 4479.

Rule: AGENTS.md forbids em dashes as a separator in docs and comments. Note this file already carries 11 pre-existing instances at the base commit, so the added lines match surrounding style; see the summary for the repo wide count.

src/process/runner.rs
```diff
@@ -1331,6 +1354,57 @@ impl RunnerShared {
         bytes.push(b'\n');
         Some(bytes)
     }
+
+    /// Track a daemon-issued relay `session/new` — a driven conversation
```
Seluj78: **Required:** em dash on this line and 1359.

Rule: AGENTS.md forbids em dashes as a separator in docs and comments. Use commas, semicolons, or rephrase.

src/acp/agent_profiles.rs
```diff
@@ -29,6 +29,16 @@ pub struct AgentProfile {
     /// for agents whose reset semantic isn't a slash command (or isn't
     /// known yet).
     pub clear_aliases: &'static [&'static str],
+    /// When true, the adapter has no native handler for this profile's
+    /// clear aliases: forwarding the raw text would be swallowed as an
+    /// unknown command and the conversation would keep its full context
+    /// (codex-acp has no `/new`; upstream declined to add one in
+    /// codex-acp#317). The server must drive the reset itself — a fresh
```
Seluj78: **Required:** em dash on this line, 37, and 408.

Rule: AGENTS.md forbids em dashes as a separator in docs and comments. Use commas, semicolons, or rephrase.

src/server/session_service.rs
```diff
@@ -440,10 +440,28 @@ impl SessionService {
         // to the agent so the replay buffer / on-disk store captures it
         // even if the agent forward fails. The frontend treats UserPromptSent
         // as authoritative and dedupes against its own optimistic row.
-        self.acp_supervisor
+        //
+        // The publish step owns clear-command detection and tells us what to
+        // do with the text: forward it as an ordinary prompt, or — for a
```
Seluj78: **Required:** em dash on this line and 446.

Rule: AGENTS.md forbids em dashes as a separator in docs and comments. Use commas, semicolons, or rephrase.

Seluj78 (changes requested):

## AoE Review

Reviewed PR #3132 against base `main`@`5bac5397`. 8 files changed (acp, runtime, server, web lib, web tests). Existing inline threads: 3, all resolved (CodeRabbit found two real bugs, both fixed in `40863d53`).

The core approach is sound and the engineering is careful: client driven `session/new` over the byte relay rather than the v2 control channel, runner handshake cache refresh so a later `Cancel` or reattach cannot resurrect the old id, and irreversible commit ordering that adopts the fresh id before best effort config restoration. The Codex only scope is correctly justified in `agent_profiles.rs` rather than left as an oversight: OpenCode and OMP also map `/new`, but whether their adapters handle it natively is unverified, so the text forward path stays. One correctness gap, one missing test artifact, and a style cleanup below.

### Blocker

**`Cargo Test (macos)` is failing.**

`tui::home::live_send::tests::worker_flags_lock_loss_and_drops_resize_after_external_steal` panics with `timed out waiting for lock_lost flag` at `src/tui/home/live_send.rs:2798`.

`src/tui/home/live_send.rs` is not touched by this diff (last modified in #3096), so this reads as a pre-existing timing flake rather than a regression from this PR. It is not covered by the baseline the PR body pre-clears (`git::worktree` / `session::trash`), so it needs a rerun to confirm flake versus regression before merge.

### Required

**1. The driven reset does not treat in flight background agents or open tools as busy.**

`src/acp/acp_client.rs:7947`

The `ClientCmd::ResetSession` success arm in the outer idle loop proceeds straight to `session/new` and swaps `acp_session_id` without consulting `between_prompt_bg_agents` or `between_prompt_tools`. The between prompt idle watchdog at `acp_client.rs:7101` treats a non-empty `between_prompt_bg_agents` as "work in flight just like an open tool" (citing #2573), and a launched async background agent's tailer (`spawn_tailer`, wired at `acp_client.rs:6209`) keeps emitting `BackgroundAgentProgress` and `BackgroundAgentCompleted` into the same `event_tx` until it self terminates on completion, hard idle, or a closed `event_tx`. A reset does none of those. `on_receive_notification` (`acp_client.rs:5968` onward) never reads `notification.session_id`, so it cannot distinguish pre reset from post reset traffic; that was safe when one connection meant one ACP session for its whole lifetime, which this PR changes.

Failure scenario: an async background agent is running with no `session/prompt` in flight, which is the state `between_prompt_bg_agents` exists to represent. The user sends `/new`. The existing busy check only guards the inner `select!` that runs while `prompt_fut` is pinned, so nothing is busy by that check and the reset is accepted. `session/new` succeeds, the id swaps, and `SessionCleared` / `SessionContextReset` / `AcpSessionAssigned{new}` / `Stopped{session_reset}` all fire. The old session's tailer keeps pushing progress and completion events after the reset boundary, where they are attributed to the fresh conversation.

This is the same class as the busy prompt case CodeRabbit already had you fix; the second category of "busy" just did not get the same guard.

**2. `web/tests/coverage-matrix.json` was not updated.**

The diff for `web/tests/coverage-matrix.json` between base and head is empty, but the Test Coverage Analysis checkbox claiming the matrix was updated is checked.

This PR changes activity timeline behavior: `SessionContextReset` now marks the turn as having output (`web/src/lib/acpTypes.ts:1892` onward) so the reset divider does not stack a "Command produced no output." row. AGENTS.md mandates a matrix entry for any PR changing a user facing dashboard flow under right panel, diff, or notifications. CI's `validate-coverage-matrix.mjs` only enforces coverage for files under `web/src/components/**`, so it will not catch this (`acpTypes.ts` lives under `web/src/lib/`), but the AGENTS.md mandate is broader than that mechanical check. The new live spec `web/tests/live/acp-codex-new-reset.spec.ts` is the natural entry to register.

**3. Em dashes on 11 added lines.**

AGENTS.md bans em dashes as a separator in docs and comments. Added occurrences:

- `src/acp/acp_client.rs:11207-11208`
- `src/acp/supervisor.rs:184`, `4479`
- `src/process/runner.rs:1358-1359`
- `src/acp/agent_profiles.rs:36-37`, `408`
- `src/server/session_service.rs:445-446`

Context worth flagging to maintainers rather than treating as a hard gate: `src/` already carries 121 such lines across 39 files at the base commit (11 of them in `supervisor.rs` itself), and no CI job gates on it, so the added lines match surrounding style even though they contradict the written rule.

### What I verified as correct

Worth recording so this does not get re-litigated:

- Reset failure paths (`Ok(_)` same id replay, `ResetRequestError::Acp`, `ResetRequestError::TimedOut`) all leave `acp_session_id` untouched; the mutation only happens under the `new_session.session_id.0 != acp_session_id.0` guard, so no half reset state is reachable.
- `SessionCleared`, `SessionContextReset`, `AcpSessionAssigned`, `Stopped` are four sequential awaits on one `event_tx` sender inside a single async block, no spawned tasks, so the ordering the #3083 listener depends on is guaranteed by mpsc FIFO rather than only asserted in a comment.
- `runner.rs` correlates the relay `session/new` response by raw JSON id string, gates on `method == "session/new"`, requires `result.sessionId`, and clears tracking without refreshing on an error response. It cannot collide with `intercept_handshake`, which only matches numeric ids, because the crate issued reset request uses UUID string ids.
- Post reset config re-application lives inside the `ResetSession` handler and completes before `ResetSessionOutcome::Reset` is sent, so the `SetMode` command the supervisor issues afterward is guaranteed to land after it. Same ordering as the spawn path.
- `PromptDisposition` is matched exhaustively at its single production consumer (`session_service.rs`), no wildcard arm, so a third variant would fail to compile rather than silently no op.
- `clear_requires_driven_reset` is read at exactly one site; no second source of truth string matching agent names.
- The `acpTypes.ts` change is correctly placed and cannot regress the no prior prompt path, since `turnActive` implies a prompt row exists. `turnHasOutput` has one write site and one consumer, so no parallel classification logic needed the same update.
- The live spec is a genuine oracle, not a pass either way check: the fake agent mints a fresh random session id per `session/new`, so asserting a second distinct `AcpSessionAssigned` id proves a real `session/new` fired.

### Verdict

**Implementation:** Request Changes, for finding 1 plus the CI rerun.

**Project Fit:** Accept. The bug is real and filed (#2979), the scope is tight, the Codex only limitation is deliberate and documented, and the approach matches existing session lifecycle patterns in the codebase.

shixi-li:
Addressed all requested changes in 39fb7be5 and synchronized the branch with the latest `main`.

- Refuse a driven reset before `session/new` when either an open tool or a tracked async background agent is still active.
- Added independent scripted-adapter regressions for both busy categories; each verifies `PromptRejected(agent_busy)` and exactly one wire-level `session/new`.
- Registered `tests/live/acp-codex-new-reset.spec.ts` in `web/tests/coverage-matrix.json` as a high-risk live Playwright flow.
- Rephrased the newly added prose flagged by review; the added-line scan is clean of em dash and ` -- ` separators.

Validation completed:
- `cargo test --features serve acp::acp_client::tests::reset -- --nocapture` (6 passed)
- `cargo test --features serve --test integration acp_effort_respawn -- --nocapture` (3 passed with the test shim installed)
- `cargo fmt --check`
- `npm run format:check`, `npm run lint`, `npm run build`
- `npm run test:unit -- src/lib/acpTypes.test.ts` (128 passed)
- `node web/tests/validate-coverage-matrix.mjs`
- live `acp-codex-new-reset.spec.ts` against the rebuilt branch binary (1 passed)

All six review threads have replies and are resolved. @Seluj78, could you please take another look when the new checks settle?

src/acp/supervisor.rs
shixi-li: Fixed in 39fb7be5. Rephrased both added comment lines in supervisor.rs to remove the em dash separators. A scan of all lines added by this PR now finds no em dash or ` -- ` prose separator.

src/acp/agent_profiles.rs
```diff
@@ -376,6 +405,31 @@ mod tests {
         assert!(!OMP.is_clear_command("/clear"));
     }
 
+    /// #2979: only codex needs the server-driven reset because codex-acp has no
+    /// native `/new` (upstream codex-acp#317), so the text-forward path
+    /// silently keeps the conversation's context. Claude's `/clear` is
+    /// handled by its adapter and must stay on the forward path; the
+    /// other `/new` mappings (opencode, omp, kimi) are unverified and
+    /// deliberately keep the old behavior until observed.
```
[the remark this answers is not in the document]
shixi-li: Fixed in d353246e. All three newly added references in agent_profiles.rs now use the unambiguous upstream wording with the bare issue reference `#317`. cargo fmt --check, the focused profile test, and cargo clippy --features serve -- -D warnings pass.

src/acp/acp_client.rs
shixi-li: Fixed in 39fb7be5. Rephrased both added comment lines in acp_client.rs to remove the em dash separators. A scan of all lines added by this PR now finds no em dash or ` -- ` prose separator.

src/acp/agent_profiles.rs
shixi-li: Fixed in 39fb7be5. Rephrased all three added comment lines in agent_profiles.rs to remove the em dash separators. A scan of all lines added by this PR now finds no em dash or ` -- ` prose separator.

src/server/session_service.rs
shixi-li: Fixed in 39fb7be5. Rephrased both added comment lines in session_service.rs to remove the em dash separators. A scan of all lines added by this PR now finds no em dash or ` -- ` prose separator.

src/process/runner.rs
shixi-li: Fixed in 39fb7be5. Rephrased the two added comment lines in runner.rs to remove the em dash separators. A scan of all lines added by this PR now finds no em dash or ` -- ` prose separator.

src/acp/acp_client.rs
shixi-li: Fixed in 39fb7be5. The idle-loop ResetSession arm now checks the same between_prompt_tools and between_prompt_bg_agents state used by the between-prompt watchdog before issuing session/new. If either tracker is non-empty, it emits PromptRejected(agent_busy), returns Failed, leaves both trackers and the running work intact, and does not issue session/new or publish reset-success boundaries.

Two independent scripted-adapter regressions cover an open tool and a tracked async background agent. Each waits for the corresponding live event, attempts /new, asserts the refusal, and verifies the wire still contains exactly one session/new. The focused reset group passes 6/6.

shixi-li:
Follow-up d353246e also addresses the new CodeRabbit issue-reference wording comment; that thread is replied to and resolved. The latest head now has 0 unresolved review threads.

The new fork workflow runs are waiting for maintainer approval (`action_required`), not failing:
- CI: https://github.com/agent-of-empires/agent-of-empires/actions/runs/30344785782
- Tests: https://github.com/agent-of-empires/agent-of-empires/actions/runs/30344785944

@Seluj78, please approve the workflow runs when convenient so the macOS rerun and the rest of the checks can execute.

shixi-li:
Re-review update for the current head d353246e:

- all 10 review threads are resolved;
- the requested busy-tool and background-agent guards, regressions, coverage-matrix registration, and prose fixes are present;
- CI run 30344785782 is successful;
- Tests run 30344785944 is successful;
- the PR is mergeable and there have been no further code changes since those successful runs.

@Seluj78 @njbrake, could you re-review the current head when convenient and replace the now-stale CHANGES_REQUESTED state if the requested changes are satisfactory?

Seluj78 (approved):

## AoE Re-Review

Head moved `40863d53` to `d353246e` (3 new commits plus an upstream merge; base is now `a89a9fa2`). All four findings from my previous review are addressed. All 10 review threads resolved. CI: 28 passed, 0 failed.

### Findings resolved

**1. Reset ignored in flight background work.** Fixed in `39fb7be5`, and fixed at the right abstraction level rather than patched at the symptom. The commit extracts `BetweenPromptWorkState` plus `between_prompt_work_state()` and refactors the existing idle watchdog to use the same helper, so the two consumers of "is between prompt work in flight" cannot drift apart. The `ResetSession` arm now checks `is_busy()` before anything else, and on busy it warns with both flags as structured fields under `target: "acp.protocol"`, emits `PromptRejected { reason: "agent_busy", text }`, sends `ResetSessionOutcome::Failed`, then continues, so there is no `session/new`, no id swap, and no `SessionCleared`. Changing `text: _` to `text` means the rejection frame carries the original text. The `"agent_busy"` reason matches the two pre-existing busy sites verbatim (`acp_client.rs:7753`, `7800`), so the parity claim holds.

The tests are genuine oracles rather than shape checks. `reset_between_prompts_with_background_agent_is_refused` builds a real `async_launched` Agent tool response with an output transcript and blocks until `Event::BackgroundAgentLaunched` confirms registration in `between_prompt_bg_agents` before asserting the refusal. The shared assertion helper checks four things: the outcome is `Failed`, a `PromptRejected{agent_busy}` arrives, no `SessionCleared` is emitted, and no second `session/new` appears in the wire capture. That would fail against the pre-fix code.

**2. Coverage matrix.** The `acp.codex-new-reset` entry is added, `kind: live-playwright`, `risk: high`. Listing Rust paths under `components` matches existing convention (7 pre-existing entries do it, for example `acp.startup-error-banner-live`), so there is no schema concern.

**3. Em dashes.** Zero on added lines across all 9 files now. The rewordings preserve meaning. This also cleaned the `agent_profiles.rs` test to go through `resolve("codex")` rather than the `CODEX` const, which incidentally covers the lookup path too.

**4. `Cargo Test (macos)`.** Green. Confirms the earlier `worker_flags_lock_loss_and_drops_resize_after_external_steal` failure was the flake it looked like rather than a regression.

Also picked up since the last pass: one new CodeRabbit thread on issue reference format in `agent_profiles.rs`, resolved by `d353246e`.

Clean on re-run of the mechanical gates: dead code allows, `CLAUDE.md` direct edits, git config mutation, hardcoded namespace strings, untagged tracing calls, `format!` inside tracing macros, secrets in logs, and raw `console.*` in `web/src`.

### Nit, not blocking

**Residual check then act window on the busy guard.** `between_prompt_tools` and `between_prompt_bg_agents` are written by the notification handler task and read by the command loop, so the guard is a check across an await point. After it passes, `session/new` runs under the roughly 28s deadline, and a late arriving notification (an adopted turn per #2899, or a straggler) could register new work from the old session during that window and have it land after the boundary.

Probably not worth fixing: a `std::sync::Mutex` cannot be held across the await, the commit deliberately treats `session/new` as the irreversible commit point so a re-check could not roll back anyway, and closing the window properly would need a reset in progress latch that the notification handler consults. The consequence is now a handful of misattributed events in a narrow race rather than the guaranteed leak the original code had. Recording it here so it is a known and accepted limit rather than an undiscovered one.

### Verdict

**Implementation:** Approve. The one substantive finding was fixed at the right abstraction level with real regression coverage, and the replies matched what the code actually does on every thread I spot checked.

**Project Fit:** Accept. The bug is real and filed (#2979), the scope is tight, the Codex only limitation is deliberate and documented, and the approach matches existing session lifecycle patterns in the codebase.

PR #3140, opened 2026-07-28
feat(acp): route MiniMax models via Anthropic-compatible endpoint

octo-patch:
Reason: The bundled ACP agent's pickModel only routed Claude/Anthropic, OpenAI, and Gemini prefixes; MiniMax models and their global/CN endpoints were absent.

## Changes

- Add a MiniMax provider via `createAnthropic` using the Anthropic-compatible Messages endpoint (`https://api.minimax.io/anthropic` by default).
- Route `MiniMax-` and `minimax:` model-id prefixes to the MiniMax provider in `pickModel`, matching the existing prefix-dispatch pattern.
- Allow operators to override the base URL through `AOE_MINIMAX_BASE_URL` (e.g. to point at the CN endpoint `https://api.minimaxi.com/anthropic`) and supply the API key through `MINIMAX_API_KEY`.
- Export `pickModel` and guard the `main()` entry-point call so the module can be imported by tests without starting the ACP server.

## Checks

- `npx tsx --test test/*.ts` — 8 tests pass (5 transcript + 3 pick-model).

## Summary

- Added MiniMax model support through its Anthropic-compatible API.
- Added configurable MiniMax endpoint and API key handling.
- Updated model routing for `MiniMax-` and `minimax:` identifiers.
- Exported model selection logic and made startup test-friendly.
- Added tests covering MiniMax routing and existing providers.

## Benefits

Enables MiniMax models in the bundled ACP agent while preserving existing provider behavior and improving testability.

### DIFF
--- a/acp-worker/aoe-agent/src/index.ts
+++ b/acp-worker/aoe-agent/src/index.ts
@@ -17,13 +17,22 @@
 import * as acp from "@agentclientprotocol/sdk";
 import { Readable, Writable } from "node:stream";
 import { streamText, tool, stepCountIs, type ModelMessage } from "ai";
-import { anthropic } from "@ai-sdk/anthropic";
+import { anthropic, createAnthropic } from "@ai-sdk/anthropic";
 import { openai } from "@ai-sdk/openai";
 import { google } from "@ai-sdk/google";
 import { z } from "zod";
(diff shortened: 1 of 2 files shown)


Seluj78:
Thanks but no thanks, we don't accept drive by ads, even less without an issue or pre-existing discussion in discord or github.

Issue #488, opened 2026-03-20
Agent marked dead when detaching with a split pane active


patjlm:
## Description

When a user creates a split pane inside an aoe-managed tmux session and detaches with the split pane active (rather than the agent pane), aoe incorrectly marks the session as dead/errored.

## Steps to reproduce

1. Create a new session in aoe
2. Inside the tmux window, create a split pane (`Ctrl+b "` or `Ctrl+b %`)
3. Make the new (non-agent) pane active (e.g. click on it). The agent session remains live on the other pane.
4. Detach (`Ctrl+b d`)

**Expected:** aoe shows the agent session as still running (green)
**Actual:** aoe marks the session as dead (red). Pressing Enter creates a brand new session instead of reattaching -- the original agent panes are lost.

If the agent pane is active when detaching, aoe correctly detects the agent as alive.

## Root cause

The status detection functions in `src/tmux/utils.rs` use the target `{session_name}:^`:

```rust
// src/tmux/utils.rs:63
let target = format!("{session_name}:^");
```

The `^` syntax targets the **first window**, but does not specify a pane. When no pane is specified, tmux resolves to the **active pane** within that window. If the user has split the window and made the new pane active, all three status checks query the wrong pane:

- **`is_pane_dead()`** (line 59) -- checks `#{pane_dead}` on the user's split pane, not the agent pane
- **`pane_current_command()`** (line 73) -- reads `#{pane_current_command}` from the user's split pane, which is typically a shell, causing `is_pane_running_shell()` to return `true`
- **`get_pane_pid()`** in `src/process/mod.rs:16` -- same issue with the same target format

Since the user's split pane is running a shell, `is_pane_running_shell()` returns `true`, and the status logic in `src/session/instance.rs` marks the session as `Status::Error`.

## Suggested fix

Explicitly target pane index 0 of the first window by appending `.0`:

```rust
let target = format!("{session_name}:^.0");
```

This ensures the check always hits the agent's pane regardless of which pane is active. This needs to be updated in three places:

1. `src/tmux/utils.rs:63` (`is_pane_dead`)
2. `src/tmux/utils.rs:75` (`pane_current_command`)
3. `src/process/mod.rs:16` (`get_pane_pid`)

Note: `remain-on-exit` is already correctly set per-pane (`-p` flag in `append_remain_on_exit_args`), so only the agent pane has it -- no changes needed there.

## Related

[...]

Issue #435, opened 2026-03-11
My opencode agent keeps getting nuked every time I jump out and in


from-nibly:
I'm on version 0.15.0 and whenever I use `Ctrl+b d` to jump out of my opencode agent session and then jump back into the agent session it gets killed.

It looks like it doesn't get killed until I try to rejoin.

This is really annoying because it stops my agent and I have to load a previous session.

I'm on macos sonoma.

gdw2vs:
My opencode sessions are also getting nuked, but it seems to happen when I'm in the session and open up a separate window (ctrl-b c) and then dettach while the second window has focus.  If I switch back to the first window, which is hosting opencode, and then dettach, its fine.

gdw2vs:
This sounds plausible.  I can try to test tomorrow...

> Here's the key issue, found at src/tui/app.rs:390-395:
> if !tmux_session.exists()
>     || tmux_session.is_pane_dead()
>     || (!instance.expects_shell() && tmux_session.is_pane_running_shell())
> {
>     if tmux_session.exists() {
>         let _ = tmux_session.kill();  // <-- kills the session
>     }
>     // ... then restarts it
> When you press Enter to attach to a session, attach_session() is called. It runs is_pane_dead() which calls:
> tmux display-message -t <session_name> -p "#{pane_dead}"
> Here's the bug: when you have a second window open and it's the active window when you detach, tmux display-message -t <session_name> resolves the target to the currently active window/pane -- which is your second window. That pane has no remain-on-exit on, so when its shell exits (or if it's just in a different state), pane_dead may return 1. The app then concludes the agent is dead and kills+restarts the session.
> To confirm: next time you try to attach after detaching from the second window, does it restart the agent from scratch rather than resuming where it left off? If so, that's exactly this code path.
> The fix would be to explicitly target pane 0 of window 0 in is_pane_dead():
> // Instead of: tmux display-message -t <session_name> -p "#{pane_dead}"
> // Use:        tmux display-message -t <session_name>:0.0 -p "#{pane_dead}"
> That pins the check to the agent's pane regardless of which window is active. Want me to make that fix?

<img width="667" height="496" alt="Image" src="https://github.com/user-attachments/assets/43e38b0f-0e53-4dfd-bc18-e96edd47f6dc" />

from-nibly:
[...]

PR #3141, opened 2026-07-28
feat(tui): show every tmux pane in the session preview

### LINKED ISSUE #2068: Preview captures only pane 0 — agent-team split-windows render the main pane thin and hide teammate panes
## Summary

When an agent running under aoe spawns a **multi-pane** tmux layout — e.g. a Claude Code **agent team** (`teammateMode: auto`, which splits the current tmux window into one pane per teammate when running inside tmux) — the home-screen **preview** shows only the main agent's pane, rendered **thin**, and the teammate panes are **not visible at all**. The preview area itself stays full width; only the captured content is narrow.

## Repro

1. Start a session in aoe (Claude Code agent).
2. From inside that agent, start an agent team — Claude Code splits the tmux window side-by-side into N panes.
3. Look at the aoe preview for that session.

- **Expected:** preview shows the whole window (all panes).
- **Actual:** preview shows only pane 0, now ~`window_width / N` columns wide → thin output with dead space on the right; the teammate panes are invisible.

## Root cause (line refs at `51fe1bd`)

aoe's preview/status model assumes **1 session = 1 pane (pane 0)**. Every capture/status read hard-targets `<session>:^.0`:

- `src/tmux/session.rs:294` — `capture_pane_with_size` → `tmux capture-pane -t <session>:^.0 …`
- `src/tmux/session.rs:278` (`capture_pane`), `:140`, `:350`
- `src/tui/home/live_send.rs:763` — the capture worker: `format!("{}:^.0", tmux_name)`

This is deliberate, per the comment at `session.rs:292`: *"target the first window's first pane regardless of base-index or which pane is active."*

The preview render resizes the **whole window** to the preview rect (`src/tui/home/render.rs:1503` → `session.rs:437` `resize_window` → `tmux resize-window -x -y`). That sizes the window correctly — but pane 0 inside a split window is only a fraction of it.

Confirmed on plain tmux, window held at 80 cols throughout:

| state | window | pane `^.0` width (captured) | other panes |
|---|---|---|---|
| 1 pane | 80×24 | **80** | — |
| 3 panes (`split-window -h` ×2), window resized to 80 | 80×24 | **26** | `1 @left27`, `2 @left54` |

After the split, `capture-pane -t <session>:^.0` returns the 26-col pane 0 and never sees panes 1–2 — exactly the "thin output + missing panes" symptom. aoe doesn't split panes itself (its only `split-window` calls are test-only — the #435/#488 regression tests); the split comes from the agent inside the session.

## Proposed fix — multi-pane-aware preview

Keep the `^.0` fast path when the window has a single pane (no regression to the common case). When `pane_count > 1`, composite all panes by geometry:

1. `tmux list-panes -t <session>:^ -F '#{pane_id} #{pane_left} #{pane_top} #{pane_width} #{pane_height}'`
2. `tmux capture-pane -t <pane_id> -p -e -S -<lines>` per pane
3. Blit each pane's captured lines into a `window_width × window_height` grid at `(pane_left, pane_top)`, with divider columns at the boundaries, and render that.

Touch points: `Session::capture_pane_with_size` / `capture_pane` (`session.rs:278-312`) and the worker target (`live_send.rs:763`) gain a `pane_count > 1` branch; the resize path already sizes the window correctly.

## Environment

- aoe built from `51fe1bd` (line refs above are at that commit).
- Linux (kernel 6.19), tmux preview substrate.

---

*AI-usage disclosure (per the issue template): the observed symptoms are mine, reproduced while running aoe with a Claude Code agent team. The root-cause walk, the tmux repro table, and the proposed fix were produced with Claude (AI) under my direction.*

Eric162:
## Preview

https://github.com/user-attachments/assets/ae8cce0a-eec1-4585-bc53-3877293f32c9

## Description

Closes #2068.

The session preview captured `^.0` and nothing else, so a pane you split off by hand was invisible: aoe showed the agent's pane and silently dropped everything beside it. No error, no hint, just missing output.

This reads every pane and lays them back out on the window grid with tmux-style borders in the gaps.

```
AGENT PANE                         │LOGS
working...                         │200 ok
                                   │404 nope
                                   │
───────────────────────────────────│
$ tail -f                          │
```

**Input routing is untouched.** Keystrokes still go to `^.0` exclusively, exactly as before (#435, #488). Compositing is read-only, so the mis-targeted-keystroke class of bug that the pin exists to prevent cannot arise here. That asymmetry is the whole reason this is tractable: reading every pane is safe, writing to the wrong one is not.

### Cost

An unsplit window, which is nearly every session, pays nothing new:

- **One fork, as before.** The pane count rides along as a header on the capture that was already happening, and the pane bytes are returned verbatim. There's a test asserting byte-for-byte equality with `capture_pane` on an unsplit window.
- **The VT fast path is preserved.** The capture worker probes the pane count on a lazy 2s cadence (one tiny fork on the single previewed pane), so an unsplit session never leaves its fork-free steady state.

A split window in live mode splits the refresh rates rather than forking per frame. Pane 0 is the only pane receiving input, so it's the only one whose latency can be felt; it keeps coming from its VT grid every frame. The panes beside it are being watched, not driven, so they come from a layout cached for 300ms. That's ~3 forks/sec instead of ~80, with echo latency identical to an unsplit session.

### One nice discovery

tmux puts pane 0 at the window origin, and keeps it there: pane indices follow layout order, and closing pane 0 renumbers whichever pane takes that corner. Verified against a live server. So the VT grid's cursor already indexes the composite with no coordinate translation. What does need restating is the frame it's measured against, since the renderer anchors on `pane_height` against the painted line count, which is now the window rather than one pane.

### Deliberate limits

- **Splits lose scrollback.** Panes have independent histories with no coherent way to stack them, so the composite covers the visible window only. The preview's scroll offset clamps itself to the shorter capture, so this reads as "a split window doesn't scroll back" rather than misbehaving. Close the extra pane and scrollback returns.
- **Wheel and mouse forwarding still go to pane 0** regardless of which pane the pointer is over. Consistent with keystroke routing rather than a new inconsistency, but now that a split window is fully visible, hovering the other pane and scrolling will feel off. Happy to gate it on pane 0's rectangle in a follow-up if reviewers prefer.
- **Border junctions are approximate.** tmux draws tee/cross glyphs where rules meet; this draws `─` and `│` and lets them abut. Deliberate for a preview.

### Also fixes a latent panic

vt100 0.16 underflows whenever content wraps on a one-row grid, and `resize-pane -y 1` is a layout a user can produce. The pane parser now works at two rows minimum and reads back the real height. Found by a unit test, not in the wild.

### Removals

Three methods lost their last caller and are gone per the no-dead-code rule: `Instance::capture_output`, `TerminalSession::capture_pane`, `ContainerTerminalSession::capture_pane`. `Session::capture_pane` stays; the CLI, web API, status detection, and smart-rename all still use it.

## PR Type

- [x] New Feature
- [x] Bug Fix

## Checklist

The block above is real captured output from a 3-pane layout, not a mockup. A screen recording of the TUI can be attached if reviewers want to see it live.

Docs unchanged: this makes the existing preview show more of what's already there, and adds no setting, command, or flag to document.

## Test Coverage Analysis

Rust-only; the web live view has its own relay path and is untouched.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: see below

Covered by 22 unit tests plus 4 integration tests against a real tmux server:

- **Pure splice** (14 tests, no tmux): side-by-side, stacked, a pane shorter than its neighbour, a row missing from a capture, ANSI rows spliced without being cut, the 3-pane layout where a left column's rule must stop at the vertical border, unclaimed columns, zero-width panes, and the first-pane/row-swap contract the live path depends on.
- **Padded row extraction** (5 tests): full-width padding, bare-LF unstaircasing, style reset before padding, the one-row-pane panic, and grid rectangles narrower/taller/equal.
- **Against real tmux** (4 tests): an unsplit window composites byte-identically to `capture_pane`; a split-off pane lands on the same rows as the agent's; the captured layout returns pane 0 first at the origin *with pane 1 selected active*; and both composite transports agree byte-for-byte on a static window.

That last one guards the invariant whose violation produced a real bug during development: the fallback path and the worker path disagreeing made the split visible for one frame after each keystroke and vanish while idle.

**What isn't covered, honestly:** the worker's transport gate itself is a thread over atomics, and the render-side wiring needs a live `HomeView`, so neither is unit-testable. Both were verified by hand (split → composite appears; `Tab` → stays visible while typing; unsplit → unchanged). A TUI e2e that splits a session, enters live mode, types, and waits out an idle cycle would close that gap; glad to add one if reviewers want it before merge.

## AI Usage

- [x] AI was used

Implemented with Claude Code across two commits, with the design argued out first: whether to follow the active pane, composite read-only, or leave splits alone. The read/write asymmetry (#488 was an input bug, so reads can safely see everything) and the pane-0-at-origin property were both established during that discussion, the latter by probing a live tmux server rather than assuming it.

Two bugs in the work were caught by its own tests: the initial gap-fill rule painted border glyphs into the dead corner beside a short pane, and the vt100 one-row panic. A third, the one-frame flicker, was caught by manual testing and is now pinned by the transport-agreement test.

## Summary

- Updated the TUI session preview to build a composited “window snapshot” that captures **all tmux panes** and renders them into a single preview using the **real tmux window layout**, with **approximate borders/junctions** between panes.
- Added a tmux compositing layer that parses/caches the window layout and **rebuilds pane placement** efficiently for split windows, while keeping **previous byte-identical behavior** for unsplit sessions.
- Switched the preview capture APIs across the session/terminal/container/tool layers from **pane-only** to **composited-window** variants, including cursor metadata updates and documentation of the split-window behavior.
- Refined live-mode behavior:
  - Detects whether the tmux window is split (with throttled pane-count probing).
  - Uses a VT-grid-based compositing path when possible, otherwise falls back to a simpler composited capture.
  - Drops stale cached layouts when panes close.
  - In composite/split mode, **scrollback history is unavailable** and **input/mouse handling is restricted to pane 0** (events outside pane 0 are dropped where appropriate).
- Hardened VT/pane parsing to prevent crashes for edge cases (notably **one-row panes**) by enforcing a minimum parser height and then using the actual pane height after parsing.
- Removed several unused capture methods while keeping the core `Session::capture_pane`.

## Benefits

- Split-window previews now look much closer to what users see in tmux: **all panes are visible**, laid out correctly by **window geometry/order**, and visually separated by **border/junctions**.
- More reliable preview rendering and cursor alignment across resizing and tricky pane dimensions.
- Clearer/safer interaction semantics in split mode: **no misleading scrollback**, and mouse/input routing consistently targets **pane 0**.

## Potential Enhancements

- The border/junction rendering is intentionally approximate; further tuning could improve fidelity for unusual/rapidly changing geometries.
- Cached-layout reuse depends on probing and refresh timing; performance/accuracy tradeoffs may be improved with additional tuning for highly dynamic layouts.

### DIFF
--- a/docs/guides/live-mode.md
+++ b/docs/guides/live-mode.md
@@ -91,6 +91,24 @@ agent's history without leaving live mode or forwarding the keys. Bare
 `PageUp` / `PageDown` still pass through to the agent, so agents that
 page their own UI keep working.
 
+## Split windows
+
+If you split the session's tmux window by hand (`Ctrl+B %`, `Ctrl+B "`), or
+an agent splits it for you (a Claude Code agent team spawns one pane per
+teammate), the preview composites every pane onto the window grid with
+borders in the gaps, so nothing is hidden.
+
+Two things change while a window is split:
+
+- **Scrolling history is unavailable.** Panes keep independent scrollbacks
+  with no coherent way to stack them, so the composite covers the visible
+  window only and `Shift+PageUp` has nothing to scroll into. Close the extra
+  pane and history comes back.
+- **Input still goes to the first pane.** Keystrokes, clicks, and the scroll
+  wheel all reach the agent's own pane regardless of which pane you split
+  off, so a wheel or click aimed at a neighbouring pane does nothing. Attach
+  to the session to interact with the other panes.
+
 ## Inserting newlines
 
 `Shift+Enter` inserts a newline into the agent's input box on
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -1162,6 +1162,7 @@ mod tests {
             mouse_sgr: false,
             mouse_all: false,
             position_reliable: true,
+            composite_pane0: None,
         };
         let json = frame_json("hello\nworld", Some(&cursor));
         let v: serde_json::Value = serde_json::from_str(&json).unwrap();
@@ -1190,6 +1191,7 @@ mod tests {
             mouse_sgr: false,
             mouse_all: false,
             position_reliable: true,
+            composite_pane0: None,
         };
         let v: serde_json::Value = serde_json::from_str(&frame_json("x", Some(&cursor))).unwrap();
         assert_eq!(v["altScreen"], true);
@@ -1211,6 +1213,7 @@ mod tests {
             mouse_sgr: false,
             mouse_all: false,
             position_reliable: true,
+            composite_pane0: None,
         };
         let v: serde_json::Value = serde_json::from_str(&frame_json("x", Some(&cursor))).unwrap();
         assert!(v["cursor"].is_null());
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -5185,16 +5185,9 @@ impl Instance {
         self.update_status_with_metadata(None);
     }
 
-    pub fn capture_output(&self, lines: usize) -> Result<String> {
-        // capture-pane has no size parameters: the pane is captured at
-        // the window's own dimensions. (A previous *_with_size variant
-        // accepted width/height and silently ignored them.)
-        self.tmux_session()?.capture_pane(lines)
-    }
-
-    /// Like [`capture_output`](Self::capture_output) but with the window's
-    /// other panes composited in, for the passive preview. Costs the same
-    /// single fork when the session has not been split.
+    /// Capture the session's window for the preview, with any panes the user
+    /// split off composited in. `capture-pane` has no size parameters: the
+    /// window is captured at its own dimensions.
     pub fn capture_output_composited(&self, lines: usize) -> Result<String> {
         self.tmux_session()?.capture_window_composited(lines)
     }
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -5191,6 +5191,13 @@ impl Instance {
         // accepted width/height and silently ignored them.)
         self.tmux_session()?.capture_pane(lines)
     }
+
+    /// Like [`capture_output`](Self::capture_output) but with the window's
+    /// other panes composited in, for the passive preview. Costs the same
+    /// single fork when the session has not been split.
+    pub fn capture_output_composited(&self, lines: usize) -> Result<String> {
+        self.tmux_session()?.capture_window_composited(lines)
+    }
 }
 
 fn generate_id() -> String {
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -55,6 +55,53 @@ pub(crate) struct CapturedPane {
     pub rows: Vec<String>,
 }
 
+/// A window's dimensions and every pane's rectangle plus captured rows.
+///
+/// Held as a unit so the live preview can cache one across frames: the panes
+/// the user is only *watching* refresh on a lazy cadence, while pane 0 (the
+/// one receiving input, whose latency is the only one that can be felt) is
+/// re-rendered every frame from its VT grid.
+pub(crate) struct WindowLayout {
+    pub window_width: u16,
+    pub window_height: u16,
+    pub panes: Vec<CapturedPane>,
+}
+
+impl WindowLayout {
+    pub(crate) fn composite(&self) -> String {
+        composite_window(self.window_width, self.window_height, &self.panes)
+    }
+
+    /// The first pane's rectangle, which tmux guarantees sits at the window
+    /// origin: pane indices follow layout order, so index 0 is the top-left
+    /// pane, and closing it renumbers whichever pane takes that corner. That
+    /// is what lets a VT grid's cursor be painted onto the composite with no
+    /// coordinate translation.
+    pub(crate) fn first_pane(&self) -> Option<PaneGeom> {
+        self.panes.first().map(|p| p.geom)
+    }
+
+    /// Composite with the first pane's rows swapped for `rows`, for the live
+    /// path's fresh VT-grid frame over a cached layout.
+    pub(crate) fn composite_with_first_pane_rows(&self, rows: &[String]) -> String {
+        let Some(first) = self.panes.first() else {
+            return self.composite();
+        };
+        let mut panes: Vec<CapturedPane> = Vec::with_capacity(self.panes.len());
+        panes.push(CapturedPane {
+            geom: first.geom,
+            rows: rows.to_vec(),
+        });
+        for pane in &self.panes[1..] {
+            panes.push(CapturedPane {
+                geom: pane.geom,
+                rows: pane.rows.clone(),
+            });
+        }
+        composite_window(self.window_width, self.window_height, &panes)
+    }
+}
+
 /// Border glyphs. tmux draws proper tee/cross junctions; a preview only needs
 /// the two edges, so a full-width gap row is drawn as an unbroken rule rather
 /// than tracking which columns carry a vertical border through it.
@@ -213,6 +260,55 @@ mod tests {
             pane(4, 0, 3, 2, &["xyz", "uvw"]),
         ];
         assert_eq!(composite_window(7, 2, &panes), "abc│xyz\n   │uvw");
+    }
+
+    fn layout(w: u16, h: u16, panes: Vec<CapturedPane>) -> WindowLayout {
+        WindowLayout {
+            window_width: w,
+            window_height: h,
+            panes,
+        }
+    }
+
+    #[test]
+    fn first_pane_is_the_one_at_the_window_origin() {
+        // tmux orders pane indices by layout, so index 0 is the top-left pane.
+        // The live path relies on this to paint pane 0's cursor onto the
+        // composite without translating its coordinates.
+        let l = layout(
+            9,
+            1,
+            vec![pane(0, 0, 4, 1, &["left"]), pane(5, 0, 4, 1, &["rght"])],
+        );
+        let first = l.first_pane().expect("a first pane");
+        assert_eq!((first.left, first.top), (0, 0));
+    }
+
+    #[test]
+    fn swapping_the_first_pane_rows_leaves_the_others_alone() {
+        // The live path's whole trick: a cached layout re-rendered with only
+        // pane 0 refreshed from its VT grid.
+        let l = layout(
+            9,
+            2,
+            vec![
+                pane(0, 0, 4, 2, &["old1", "old2"]),
+                pane(5, 0, 4, 2, &["keep", "same"]),
+            ],
+        );
+        let fresh = vec!["new1".to_string(), "new2".to_string()];
+        assert_eq!(
+            l.composite_with_first_pane_rows(&fresh),
+            "new1│keep\nnew2│same"
+        );
+        // The cached layout is not consumed: the next frame swaps again.
+        assert_eq!(l.composite(), "old1│keep\nold2│same");
+    }
+
+    #[test]
+    fn swapping_rows_on_an_empty_layout_is_a_no_op() {
+        let l = layout(3, 1, vec![]);
+        assert_eq!(l.composite_with_first_pane_rows(&["x".to_string()]), "");
     }
 
     #[test]
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -43,6 +43,21 @@ impl PaneGeom {
         row >= self.top && row < self.top.saturating_add(self.height)
     }
 
+    /// Whether two pane rectangles share any cell.
+    ///
+    /// Panes in a normal window tile the grid, but a zoomed pane (`C-b z`) is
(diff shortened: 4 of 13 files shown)


njbrake (approved):

Approving. This is careful work, and the design argument holds up: the read/write asymmetry is the right distinction, and it is what makes compositing tractable where following the active pane would not be.

I did not take the empirical claims on faith, because this specific area has bitten the project three times. Notes below on what I reproduced, one issue with a suggested patch, and one thing I would like checked by hand before merge.

## The pane-0 pin lineage, and why this does not reopen it

For context on why I went at this hard: the `^.0` pin exists because of three prior bugs, all the same root cause of a tmux target resolving to the active pane instead of the agent's.

- #435 / #440: a user creates an extra tmux window and detaches from a non-first one. `display-message -t <session>` resolved to the active window and pane, so the health checks declared the agent dead and killed the session on reattach.
- #488 / #489: a user splits a pane and detaches with the split active. The split runs a shell, so `is_pane_running_shell()` returned true and the session went to `Status::Error`. The reporter lost their agent panes.
- #2231 / #2568: hosts with `pane-base-index 1` set globally make `^.0` unresolvable.

#488 is the same scenario this PR targets, which is what made it worth verifying rather than reasoning about. I checked against a live server with pane 1 selected, which is the #488 shape:

```
:^   pane_width=29  pane_left=31   <- the user's split, the wrong pane
:^.0 pane_width=30  pane_left=0    <- the agent's pane
window_panes=2  window_width=60  window_height=12   <- immune to selection
```

The PR gets this right. Every pane-level format field is explicitly addressed (`^.0` for the cursor probe, `^.{i}` for each pane's geometry), and only window-level fields go through `:^`, where active-pane selection cannot reach them. `src/tmux/utils.rs` and `src/process/mod.rs`, the three functions #489 fixed, are untouched and still pinned. Status detection and health checks still go through `capture_pane`. The #2231 lesson is carried forward too, including the retrofit in 3d824fc1 onto the test that was missing the pin.

## Other claims I reproduced rather than read

- **Pane 0 sits at the window origin, and indices follow layout order.** Verified: `split-window -hb` makes the new left-hand pane index 0. The untranslated-cursor argument holds.
- **Pane indices are contiguous from 0.** Verified: 0,1,2 then killing index 1 gives 0,1.
- **Unsplit is byte-identical to `capture_pane`.** Same `capture-pane -t <s>:^.0 -p -e -S -<lines>` invocation, sentinel preamble stripped line by line.
- **A split reads as "no scrollback" rather than misbehaving.** `clamp_scroll_to_capture` gives `real_max = captured_lines - visible_height`, so a window-height composite clamps to roughly zero, and `capture_is_exhausted` already handles a capture that is always precisely the pane height.
- **The `row_last_col` change leaves `row_to_ansi` output unchanged.** Its loop already advanced two columns per wide cell, so the corrected bound terminates in the same place.
- **One-row panes no longer panic.** Ran `resize-pane -y 1` against a real session through the composite path.
- **Nested layouts composite correctly.** Dumped a real 3-pane nested L layout; borders, junction rows, and per-row widths all correct.
- **`capture-pane -p` emits trailing blank rows to full pane height**, which is what makes `capture_rows_padded(body, w, h)` receive precisely h rows and no more.

All five CodeRabbit findings were valid and all five are already addressed in 3298779d and ff60ea87. The cursor-metadata one was the most valuable: the `alternate_on` and `mouse_*` flags feed the wheel-forward path from #2123, so dropping them would have broken scrolling for full-screen agents on the non-VT transport.

## One issue: a failed layout capture leaves a ghost pane and reforks every frame

`capture_window_layout` addresses `^.0..^.{count-1}`, and `pane_count` in the worker is up to `PANE_COUNT_PROBE_MS` (2s) stale. When the user closes a pane, the chain over-addresses and tmux fails the whole invocation. Reproduced:

```
$ tmux display-message ... -t 'e1:^.2' ... ; capture-pane -t 'e1:^.2' ...
EXIT STATUS = 1
stderr: [can't find pane: 2]
```

and through the real code path, `capture_window_layout(5)` on a 3-pane window returns `None`.

Because the cache is stamped only on success, `cache` keeps its old `Some` value and its old timestamp. So for up to two seconds the preview composites the stale rectangles, showing a ghost of the closed pane, while re-forking the failing capture every frame at `LIVE_CAPTURE_INTERVAL_FAST_MS` (25ms). That is roughly 80 forks, and it is the one place where the cost model in the description does not hold.

This matters a little more than it first looks, because #1485 ("use tmux control-mode socket for live-send preview refresh instead of per-frame capture-pane forks") closed as completed, delivered via the `VtChannel` path. The per-frame fork is the specific cost that issue was closed to eliminate, so it is worth keeping this path's fork budget intact.

Suggested patch, verified locally (`cargo test --lib` 4332 passed, `cargo fmt --check` clean, no new clippy warnings):

```diff
 fn capture_composited_over_grid(
     name: &str,
     channel: &crate::tmux::vt::VtChannel,
     cache: &mut Option<(std::time::Instant, crate::tmux::composite::WindowLayout)>,
+    last_pane_probe: &mut Option<std::time::Instant>,
     pane_count: u16,
     lines: usize,
     forward_empty: bool,
 ) -> (Option<String>, Option<crate::tmux::PaneCursor>) {
     let stale = cache.as_ref().is_none_or(|(at, _)| {
         at.elapsed() >= std::time::Duration::from_millis(COMPOSITE_LAYOUT_MS)
     });
     if stale {
-        if let Some(layout) =
-            crate::tmux::Session::from_name(name).capture_window_layout(pane_count)
-        {
-            *cache = Some((std::time::Instant::now(), layout));
+        match crate::tmux::Session::from_name(name).capture_window_layout(pane_count) {
+            Some(layout) => *cache = Some((std::time::Instant::now(), layout)),
+            // The window is no longer the one these rectangles describe. Drop
+            // them rather than compositing a pane that is gone, force a count
+            // re-probe on the next cycle, and let the fallback below carry this
+            // frame: it probes `window_panes` in the same fork as its capture,
+            // so it is correct no matter how stale the count had become.
+            None => {
+                *cache = None;
+                *last_pane_probe = None;
+            }
         }
     }
```

plus passing `&mut last_pane_probe` at the call site. Falling through to `capture_composited` is correct here because it probes `window_panes` in the same fork as its capture, so it cannot be fooled by a stale count. Failure is inherently transient (the count and the panes come from the same tmux state), so this bounds the window to a frame or two. Happy for you to take it as-is or to solve it differently.

## Please check by hand before merge: attached plus split

#3155 landed one commit before this branch's merge base, and the two have never been exercised together. That fix added a `#{session_attached}` guard so the passive preview resize no longer shrinks attached sessions, which means an attached session's window now keeps the client's size, potentially much wider than the preview rect. The composite emits rows at `window_width`, so attached plus split produces rows wider than the preview area.

Reading the painter, `render_scrolled_output` uses `Paragraph::new(...)` with no `.wrap(...)`, so rows should clip at the right edge rather than wrapping, degrading to a cut-off rightmost pane rather than a corrupted layout. That is the same class as before, since `capture-pane` always captured at window dimensions, but the rows are wider now. Worth one manual pass: attach to a split session from a client wider than the preview pane, detach, and confirm the composite clips cleanly instead of staircasing.

## Follow-ups, not blocking

1. **Link #2068.** It is open, P2, area:tui, and describes this precisely, including the "rendered thin" symptom that the full-width composite also fixes. The description never references it, so CodeRabbit's linked-issues check passed only by finding none.
2. **The web dashboard still shows pane 0 only** (`src/server/api/sessions.rs:9052`, `src/server/live_ws.rs:534,554`). After this lands the TUI and the dashboard disagree about what a split session looks like, which is visible for the same agent-team case that motivates #2068. Scoping it out of this PR is the right call; it deserves its own issue.
3. **Splitting a window silently disables preview scrollback** for a non-alternate-screen agent, since #2123 falls back to capture-window scroll and the composite has no history. It is documented and it clamps rather than misbehaving, but it is the most user-visible tradeoff here and worth a line in the docs or a follow-up.
4. **Wheel and mouse still route to pane 0** regardless of pointer position. Right call to defer, and more noticeable now that the other panes are visible. Worth an issue rather than growing this PR.
5. #1952 (extract a `PaneTransport` trait) will have a third capture transport to absorb.

## Verification

`cargo build`, `cargo fmt --check` clean, `cargo clippy --all-targets` with no new warnings (the seven present are identical on `main`), and `cargo test --lib` at 4332 passed, 0 failed, with the patch above applied. `restart_selected_session_surfaces_resume_failed_after_async_restart` fails on this head, but it also fails on the merge base and is fixed by 76a6722c on `main`, so it is a `CLAUDE_CONFIG_DIR` sensitivity unrelated to this work; merging `main` locally gives the clean run.

Worth noting that the substantive CI (`CI` and `Tests`) had not actually run on this PR; only the template and title lints had. I have approved the gated runs, so full CI is going now. Approving on the strength of the local verification above, with the expectation that CI confirms it.

PR #3143, opened 2026-07-28
feat(tmux): experimental shared tmux session with named tabs

Eric162:
Refs #3144 — motivation, tradeoffs, and breaking changes are written up there.

## Description

Experiment behind `AOE_USE_SHARED_TMUX_SESSION`: render an agent and its paired terminals as **tabs of one attached session** instead of separate tmux sessions you switch between, and give every window a real name.

Draft — the flag is deliberate. If we like it, it graduates to a real setting via the single-source settings schema.

### Two problems

1. Attaching a terminal `switch-client`s to a wholly separate session (`aoe_term_*`), so there's no tab strip and no visual relationship to the agent.
2. The tab reads as a bare version string like `2.1.220`. aoe never set a window name — repo-wide grep for `rename-window` / `automatic-rename` / `allow-rename` / `set-titles` / `window-status-*` returned zero production hits — so tmux's default `automatic-rename on` named the window `#{pane_current_command}`. Claude Code installs at `.../claude/versions/2.1.220/claude`, so the executable's **basename is the version**.

### Approach: `link-window`, not a session→window migration

The obvious implementation (make the terminal a window of the agent's session) is expensive: `^` means *first window*, and `format!("{}:^.0", name)` is hardcoded at 16 call sites for every capture, send-keys, respawn, resize, pane-pid and pane-dead check. The session-scoped `@aoe_size_owner` / `@aoe_vt_owner` cross-process locks would collide, and `kill-session` would become `kill-window` with "last window destroys the session" semantics.

Instead **each surface keeps its own tmux session**, and the terminal's window is additionally `link-window`'d into the agent's session. tmux linked windows belong to both sessions at once, so `<terminal>:^.0` keeps resolving:

```
aoe_Smoke_96439cdd      win 0  id=@0  agent (claude)
aoe_Smoke_96439cdd      win 1  id=@2  terminal
aoe_term_Smoke_96439cdd win 0  id=@2  terminal      <- same window id
```

Nothing about pane targeting, the locks, or teardown had to change. There's a test asserting `<terminal>:^.0` still resolves after linking, since that's the property the whole approach rests on.

Attach is redirected: `attach_via_host` selects the linked window in the agent session then hands off to the agent's existing `attach()`, because attaching the terminal's own session would show one window and no tab strip. It falls back to the standalone attach whenever the window isn't actually linked.

### Naming

`automatic-rename off` + `allow-rename off` (ignoring the agent's OSC 2 title escapes entirely) + explicit `rename-window`. Tabs read `agent (claude)` / `agent (pi)` from `instance.tool`, `terminal`, `container`, or the tool name; the web dashboard's extra terminals number from 2 (`terminal 2`) to match its existing labels. `window-status-format` is styled to the theme, since tmux's default `0:name-` clashes with aoe's bar.

### Three lifecycle traps from dual window ownership

Found by testing, each fixed with a test that first asserts the underlying tmux behavior so the reason the code exists is recorded rather than implied:

- **`kill-session` does not destroy a linked window.** It survives in the agent session as a stranded dead-pane tab, and the next terminal start appends a second one beside it. Teardown now kills the window first (`kill_first_window`), which drops the tab from every session holding it; tmux then destroys any session left with no windows.
- **A restart drops the tabs while the terminals live on.** `kill_clean` kills only the agent session, so the terminals' sessions and scrollback survive, but the recreated agent session has an empty window list. `finalize_launch` now re-links them (idempotent, so it runs on every launch).
- **Selecting a terminal tab leaked into the next agent attach.** tmux persists a session's current window and `attach-session` resumes there, so pressing Enter on the agent landed on the terminal. Agent attach sites now call `focus_own_window()` first. Deliberately *not* inside `Session::attach()` — the terminal attach routes through it right after selecting the terminal tab on purpose, and folding it in would break that instead.

### Flag off

Unchanged: separate sessions, one window each, tmux default names, `window-status-*` untouched. Verified on the same binary.

## PR Type

- [x] New Feature

## Checklist

## Test Coverage Analysis

The web reaches all of this through `Instance` (`server/pane.rs:101`/`:162`, `api/sessions.rs:5459`/`:5561`), so it inherits the behavior with no dashboard-flow change of its own.

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: **flagged experiment, and I want reviewer input on the approach before investing in the harness work.** 12 new tests cover the tmux mechanics against a live server, and all three bugs plus the flag-off path were verified end-to-end by driving the real binary through tmux (transcripts in the commit discussion). If this graduates past the flag, the e2e belongs on the `aoe serve` ensure-terminal endpoint, which exercises the same `Instance` path without the attach hijacking the test's tmux client. Happy to add it in this PR if preferred.

## AI Usage

- [x] AI was used

Explored the tmux session model, proposed the `link-window` approach over a session→window migration, implemented, and found/fixed the three lifecycle bugs. One test (`link_window_into_is_idempotent`) caught a real bug during development: tmux happily links the same window twice at two indices, producing duplicate tabs onto one pane — fixed in the code rather than by relaxing the test.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1177,6 +1177,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                 }
 
                 let tmux_session = crate::tmux::Session::new(&instance.id, &instance.title)?;
+                tmux_session.focus_own_window();
                 tmux_session.attach()?;
             }
             Err(e) => {
--- a/src/cli/session.rs
+++ b/src/cli/session.rs
@@ -1418,6 +1418,9 @@ async fn attach_session(profile: &str, args: SessionIdArgs) -> Result<()> {
         );
     }
 
+    // Land on the agent's own tab, not whichever paired-terminal tab was last
+    // viewed under `AOE_USE_SHARED_TMUX_SESSION`.
+    tmux_session.focus_own_window();
     tmux_session.attach()?;
     Ok(())
 }
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -38,6 +38,27 @@ pub struct TerminalInfo {
     pub created: bool,
 }
 
+/// Tab label for a paired host terminal. Index 0 is the only terminal the TUI
+/// uses and stays unnumbered; the web dashboard's additional terminals (#2437)
+/// are numbered from 2 so the strip reads `terminal`, `terminal 2`, ... matching
+/// how the web already labels those tabs.
+pub(crate) fn terminal_window_name(index: u32) -> String {
+    if index == 0 {
+        "terminal".to_string()
+    } else {
+        format!("terminal {}", index + 1)
+    }
+}
+
+/// Container counterpart of [`terminal_window_name`].
+pub(crate) fn container_window_name(index: u32) -> String {
+    if index == 0 {
+        "container".to_string()
+    } else {
+        format!("container {}", index + 1)
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 #[serde(rename_all = "lowercase")]
 pub enum Status {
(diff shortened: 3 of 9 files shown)


njbrake:

Closing to keep the review queue tidy. This has been a draft since July 28, conflicts with main, and I don't have a signal that it's still active.

Shared tmux sessions with named tabs is a big surface change, so if you want to continue, let's start with an issue to agree on the design before the implementation. Reopen or ping me any time. Thanks!

Issue #3071, opened 2026-07-22
Passive preview resize shrinks attached tmux sessions to preview-pane width


nigel-dev:
When I attach to a session (Enter from the dashboard, or tmux switch-client), the window briefly goes full-width then immediately shrinks back to whatever the preview pane width is. It stays cramped until I manually force a resize.

This started in v1.10.0 — v1.9.x didn't have this problem. Still happening on 1.13.1.

## Environment

- aoe 1.13.1 (Homebrew, macOS arm64)
- tmux 3.7b
- Ghostty
- Config: `default_attach_mode = "tmux"`, `click_action = "select_only"`

## Reproduction

1. Have a few sessions running, open the AoE dashboard
2. Select a session and press Enter
3. Watch — full width for a split second, then it snaps back to the preview width

Workaround: `tmux set-option -w -t <session> window-size latest` after attaching.

## What I think is happening

The passive preview resize in `src/tui/home/render.rs` keeps resizing previewed sessions to match the preview pane area every render cycle. When I attach, tmux expands the window to my client size (`window-size latest`), but the TUI's render loop is still running on the other tmux pane and fires `session.resize_window(preview_width, preview_height)` right after — shrinking me back down.

The guard only checks `has_active_size_owner()`, which is the AoE size-owner lock protocol. A plain tmux attach doesn't register as a size owner, so the guard never fires and the passive resize keeps winning.

The commit that introduced this logic is `0f2ce94c` ("keep the agent preview from clipping; single-source the preview geometry", #1621), first released in v1.10.0.

## Possible fix

Add a `#{session_attached}` check so the passive resize skips sessions with an attached client. The pattern already exists in `attached_session_names()` used for idle-reap.

<details>
<summary>AI-suggested code fix</summary>

`src/tmux/session.rs` — add method:

```rust
pub fn is_attached(&self) -> bool {
    let out = crate::tmux::tmux_command()
        .args([
            "display-message",
            "-t",
            &self.name,
            "-p",
            "#{session_attached}",
        ])
        .output();
    match out {
        Ok(o) if o.status.success() => {
            let s = String::from_utf8_lossy(&o.stdout);
            s.trim().parse::<u32>().unwrap_or(0) > 0
        }
        _ => false,
    }
}
```

`src/tui/home/render.rs` ~line 2161 — add guard:

```rust
if !session.has_active_size_owner() && !session.is_attached() {
    session.resize_window(width, height);
    self.preview_pane_synced = Some(want);
[...]

Issue #2766, opened 2026-07-09
Live-session terminal.ws re-asserts pane size every 2s in an unconvergeable loop (want_rows = pane_height + 1) → cursor flicker (regression in 1.12.x)


Daniel-Wiszowaty:
## Summary

Since **1.12.x**, the live-session owner loop in `src/server/live_ws.rs` re-asserts the tmux window size **every ~2s indefinitely** because its wanted grid height is **permanently 1 row greater than the pane height tmux yields** (`want_rows == pane_height + 1`). The mismatch never resolves, so the re-assert (and the pane repaint it triggers) recurs forever. User-visible effect: the hosted agent's cursor **flickers** ~twice a second, snapping to the bottom row and back. No such behavior on 1.11.x.

## Environment

| | |
|---|---|
| aoe | 1.12.1 (Homebrew); also seen on 1.12.0 |
| OS | macOS 26.5.1 / Darwin 25.5.0, Apple Silicon |
| tmux | 3.7b |
| Hosted agent | Claude Code 2.1.205 (live-send session) |
| Driver | `aoe serve`; **no attached tmux client** (`tmux list-clients` empty) |

## Evidence

The daemon logs this on a ~2.0s cadence, unattended, indefinitely:

```
WARN terminal.ws: pane drifted from live owner's grid; re-asserting tmux=aoe_… kind="live" pane_cols=115 pane_rows=66 want_cols=115 want_rows=67
WARN terminal.ws: pane drifted from live owner's grid; re-asserting tmux=aoe_… kind="live" pane_cols=115 pane_rows=66 want_cols=115 want_rows=67
WARN terminal.ws: pane drifted from live owner's grid; re-asserting tmux=aoe_… kind="live" pane_cols=115 pane_rows=66 want_cols=115 want_rows=67
```

**Invariant across every observed state:** `want_rows == pane_rows + 1`, `want_cols == pane_cols`. Confirmed at multiple sizes and on a second terminal simultaneously (`…_t1`, `pane_rows=62 want_rows=63`).

`tmux display`: `window_height=67 pane_height=66`. Directly: `want_rows` = the window height the client requested; `pane_height` = what the pane actually is → always 1 less.

## Not the status bar

[...]

PR #3155, opened 2026-07-29
fix(tui): skip passive preview resize when session has attached client

nigel-dev:
## Description

The passive preview resize in `src/tui/home/render.rs` shrinks attached tmux sessions back to the preview-pane width immediately after the user attaches. This adds a `#{session_attached}` guard so the resize skips sessions with a live tmux client, fixing the "full width for a split second then shrinks" behavior reported in #3071.

Introduced in v1.10.0 via `0f2ce94c` (#1621). The existing `has_active_size_owner()` guard only covers AoE's size-owner lock protocol (web/mobile live clients); a plain `switch-client` attach claims no lock, so the passive resize kept firing.

The dedup (`preview_pane_synced`) is left unset when skipping, so the resize retries automatically once the client detaches.

## PR Type

- [x] Bug Fix

## Checklist

- [ ] Documentation was updated where necessary (no docs affected)
- [ ] For UI changes: included screenshot or recording (behavioral fix, no visual change to capture)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: added a focused tmux-gated unit test (`test_is_attached_false_for_detached_session`) in `src/tmux/session.rs` that pins the detached path. The attached-client path needs a live tmux client which is impractical in CI. Mirrors the existing `test_is_pane_dead_*` pattern.

## AI Usage

- [x] AI was used

AI assisted with root cause analysis, source tracing through the passive resize path, and drafting the `is_attached()` implementation. I found and reproduced the bug, tested the fix locally, and verified the behavior.

## Summary

- Fixed passive preview resizing so tmux sessions that already have an attached/live client are no longer immediately shrunk to the preview-pane width after attachment.
- Added session attachment detection (`Session::is_attached()`) and used it to gate the resize logic before the existing “size owner” checks.
- Extended tmux integration tests to cover both:
  - detached sessions (should report “not attached”)
  - sessions with a live client (should eventually report “attached”)
- Documented and adjusted the guard ordering to reduce unnecessary tmux calls while a client remains attached.

## Benefits

- Attached tmux sessions keep their intended size instead of being forced into the preview-pane width.
- Detached sessions still resize correctly, and the system can retry after detaching when a resize was intentionally skipped.


njbrake (approved):

Approving. The diagnosis holds up end to end, and I reproduced the load-bearing primitive rather than taking the description's word for it.

**Root cause confirmed independently.** `Session::attach()` uses `switch-client` when `$TMUX` is set (`src/tmux/session.rs:399`), which returns immediately instead of blocking. The render loop keeps running after the attach, `clear_preview_pane_sync()` at `src/tui/app.rs:3715` has just dropped the dedup so `passive_resize_step` returns `Fire`, and the only guard was `has_active_size_owner()`, which a `switch-client` attach never trips. `default_attach_mode = "tmux"` is that path.

**Verified against real tmux 3.6**, since the whole fix rests on one format value:

| Case | `display-message -t <sess> -p '#{session_attached}'` |
|---|---|
| Detached, no client anywhere | `0`, exit 0 |
| Live client attached | `1`, exit 0 |
| Sibling session while another is attached | `0`, exit 0 (per session, not per server) |
| Dead pane (`remain-on-exit`) | `0`, exit 0; target still resolves |
| Nonexistent session | empty stdout, exit 0 |

The dead-pane row was my main regression worry, because the `Fire` branch explicitly handles previewing a Stopped session. It resolves fine, so those previews are unaffected.

**Pushed follow-ups to your branch** (one merge commit plus one commit; fast-forward, nothing rewritten):

1. Merged `main`. The branch had a conflict: your test and `test_create_forwards_desktop_env_to_session` from #3079 both landed right after `test_is_pane_dead_after_command_exits`. Kept both.
2. Added `test_is_attached_true_with_live_client`. The detached-only test passes against a hard-coded `false`, which is the inversion that reintroduces this bug, so it could not fail for the reason it exists. The attached path turns out to be testable without a controlling terminal: a `-d` session can host a real tmux client if its command unsets `$TMUX` and attaches to the target. Runs in about 0.1s, and I confirmed it fails when `is_attached` is stubbed to `false`.
3. Reordered the guard to `!is_attached() && !has_active_size_owner()`, and documented the new method. `has_active_size_owner` spawns two `show-options` calls against `is_attached`'s one, and attached is the sticky state here: the skip leaves the pending slot armed, so the block re-runs every poll for as long as the user stays attached. Short-circuiting on the cheap check drops that from three tmux spawns per poll to one.

**On the CodeRabbit findings**, for the record:

- Fail closed when attachment state is unknown: not taking it. I asked for best-effort `false` in #3071, and the matrix above shows the failure branch is close to unreachable, since a bad target exits 0 with empty output rather than erroring. The tension is worth recording though: `attached_session_names()` documents the opposite convention for the same tmux datum, so that is the thread to pull if we ever revisit.
- TOCTOU between the check and the resize: real, but not worth the cost. The AoE event loop is single threaded, so `switch-client` and a render cannot interleave on the attach path. Cross process it is a sub-millisecond window with a self-limiting consequence, and folding it into one `if-shell` would mean restructuring the `pane_chrome_rows` probe that #2766 added.

Local gate on the merged branch: `cargo fmt --check`, `cargo clippy --all-targets` (no new warnings versus main), and `cargo test` all clean, 4300 lib tests plus every integration target.

Thanks for chasing this one down, and for the writeup in #3071. The issue report did most of the work.

Issue #3157, opened 2026-07-29
Lifecycle ops resolve tmux by title-derived name: smart-renamed sessions strand their agent on trash/archive and duplicate on unarchive


sduke1848:
## Summary

Session lifecycle operations resolve a session's tmux session **by a name derived from the session's current title**. `smart_rename` changes the title without renaming the existing tmux session, so from that moment the derived name no longer matches reality. Every lifecycle operation then acts on a name that doesn't exist, and the failures are silent:

- **trash** marks the row but does not stop the agent — the tmux session and its `claude --resume` child keep running indefinitely
- **unarchive** creates a *second* tmux session under the now-current title-derived name, alongside the still-running original — two tmux sessions and two agents on one session id
- **archive** tears down only the name it knows and strands the other
- the status poller reports `Error` ("tmux session is gone") for sessions whose agent is demonstrably alive
- `aoe ps` silently falls back to `pane_current_command`, rendering `AGENT` as `2.1.220` (Claude Code sets its process title to its version) instead of `claude`

Everything below is one defect. I originally chased it as five.

## Environment

- `aoe` 1.13.1 (Homebrew), macOS 15 (Darwin 25.5.0) arm64, tmux 3.7b
- agent: Claude Code 2.1.220, launched by AoE as `claude --dangerously-skip-permissions --session-id <uuid>`
- `smart_rename = true`, `delete_to_trash = true`, `trash_retention_days = 30`, `[worktree] enabled = false`
- 1.13.2 changelog reviewed — no related fix

## Reproduction

```
1. aoe add            # session gets codename title, e.g. "Vikings" -> tmux aoe_Vikings_<id8>
2. Let smart_rename retitle it, e.g. to "Refactor billing module"
   (tmux session is NOT renamed; it stays aoe_Vikings_<id8>)
3. Trash the session from the TUI.
   => row disappears from the sidebar
   => `ps` shows the claude process still running; tmux aoe_Vikings_<id8> still alive
4. Restore it, then unarchive it.
   => AoE creates tmux aoe_Refactor_billing_mod_<id8> with a NEW claude --resume
   => aoe_Vikings_<id8> and its original agent are STILL running
   => two tmux sessions, two agents, one session id
5. Archive it.
   => aoe_Refactor_billing_mod_<id8> is torn down
   => aoe_Vikings_<id8> and its agent survive, now unreachable from the UI
```

[...]

PR #3160, opened 2026-07-30
refactor(tmux): resolve id -> live session name through NameShape

jerome-benoit:
## Description

Follow-up to #3158, addressing the non-blocking maintainability points raised in that review.

### R1: dedup id -> live tmux name resolution through `NameShape`

`tmux_env_session_name_for_instance_id` (`src/session/instance.rs`) re-implemented, inline, the two things `NameShape` already owns: the `_<id8>` suffix computation and the per-kind prefix classification (agent / paired-terminal / container-terminal, with the tool sub-session exclusion). That duplication is exactly the kind of scattered, hand-rolled id-resolution that made the #3157 title-vs-id bug possible in the first place.

- Adds `NameShape::terminal(suffix)` / `NameShape::container(suffix)` constructors alongside the existing `NameShape::agent`.
- Adds a `live_any_kind_name_for_id(live_names, session_id)` helper next to the other `NameShape` resolvers in `src/tmux/mod.rs`, so the name shapes remain the single source of truth.
- Rewrites `tmux_env_session_name_for_instance_id` to keep its own `list-sessions` call and delegate all matching to the helper.

Behavior preserving: same dead-pane filter, same tool exclusion, same `agent > terminal > container` preference, same `_<id8>` suffix semantics. The dead-pane probe (`is_pane_dead`, which shells out to tmux) runs for each kind's candidates until its first live one, so it is called at most as often as the previous inline loop, never more.

### R2: reuse the poller's resolved session name in the status probe

The status poller (`src/tui/status_poller.rs`), `aoe serve`'s poll loop (`src/server/mod.rs`), and `aoe ps` (`src/cli/ps.rs`) each resolve a session's live tmux name against the batch pane-metadata snapshot via `resolve_agent_session_name_in`, then called `update_status_with_metadata`, which resolved the *same id a second time* via `tmux_session()` -> `Session::resolve_name`.

- `update_status_with_metadata` / `update_status_with_metadata_inner` now take `resolved_name: Option<&str>`.
- The three hot paths pass the name they already resolved; the probe acts on it via `Session::from_name` instead of re-resolving.
- Every other caller passes `None` and keeps the title-derived `tmux_session()` fallback.

The `None` path is the original code moved verbatim into the `None` arm (only re-indented). The resolved path is **not** a pure behavioral identity, and intentionally so: the probed name now comes from the batch `list-panes` snapshot the poller already resolved and indexed metadata against, while `existence()` still consults the session cache. This makes the metadata lookup and the existence probe agree on a single resolved name where previously they could disagree within one poll cycle. Both the old and new code carry one narrow race, in opposite directions, when a rename lands in the sub-millisecond gap between the cache refresh and the batch scan: the old code re-resolved the probe name from the cache, so metadata and probe could target different names for one cycle while the session stayed `Present`; the new code binds both to the batch name, so a not-yet-in-cache name reads `Absent` and surfaces a transient `Error` for that instance, self-healing on a later poll once the two snapshots realign. A unit test proves the resolved name is the one actually probed (a live name the title could never derive is confirmed present only when passed in, and absent -> `Error` when it is not).

### R3 (out of scope)

The `truncate_id(id, 8)` collision domain flagged in #3158 is pre-existing across the codebase, not a regression introduced or touched here, and changing the id length is a data-format concern (session-name stability, live tmux sessions carry the `_<id8>` tail) that would need a migration and does not belong in this refactor.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

Added unit tests in `src/tmux/mod.rs` (kind-preference ordering; tool / other-id exclusion for the new helper) and `src/session/instance.rs` (the resolved-name path probes the passed name, not the title-derived one). The neighbouring `resolve_agent_session_name*` and `update_status_with_metadata_inner` suites continue to pass.

Note: `update_status_reconciles_running_hook_to_waiting_on_claude_approval_prompt` fails locally on this machine both with and without this change (a pre-existing, environment-dependent tmux pane-capture / Claude-prompt-detection test); it is unrelated to this PR.

**TUI / CLI (e2e test)**
- [x] N/A: internal refactor of status-resolution helpers; behavior preserved on the fallback path and covered by unit tests, no TUI rendering / CLI subcommand / session-lifecycle change

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Opus) via OpenCode

**Any Additional AI Details you'd like to share:** AI assisted with the refactor and tests; the behavior-preservation analysis (verbatim `None` path, prefix non-nesting, probe-call-count parity, and the R2 snapshot-consistency nuance) was cross-reviewed by three independent agents and reviewed by hand.

## Summary

- Centralized tmux session ID-to-name resolution with shared `NameShape` helpers and consistent preference ordering.
- Added terminal and container session matching, while preserving exclusions and existing suffix behavior.
- Reused resolved session names across status polling, `aoe serve`, and `aoe ps` to avoid duplicate lookups and probe the correct session.
- Updated the instance status API and tests to support explicitly resolved names.

## Benefits

- More reliable status reporting when session titles differ from live tmux names.
- Less repeated tmux name resolution.
- Improved coverage for session matching, exclusions, and preference behavior.

### DIFF
--- a/src/cli/ps.rs
+++ b/src/cli/ps.rs
@@ -400,7 +400,7 @@ fn collect_tmux_states(instances: &mut [Instance]) -> Vec<TmuxState> {
             &inst.id,
             &crate::tmux::Session::generate_name(&inst.id, &inst.title),
         );
-        inst.update_status_with_metadata(meta.get(&name));
+        inst.update_status_with_metadata(meta.get(&name), Some(&name));
         let agent = if inst.tool.is_empty() {
             meta.get(&name)
                 .and_then(|m| m.pane_current_command.clone())
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -4001,7 +4001,7 @@ async fn status_poll_loop(state: Arc<AppState>) {
                     &crate::tmux::Session::generate_name(&inst.id, &inst.title),
                 );
                 let metadata = pane_metadata.get(&session_name);
-                inst.update_status_with_metadata(metadata);
+                inst.update_status_with_metadata(metadata, Some(&session_name));
             }
             (instances, live_structured_worker_records())
         })
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -9,7 +9,6 @@ use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
-use crate::cli::truncate_id;
 use crate::containers::{self, DockerContainer};
 use crate::tmux;
 
@@ -1251,7 +1250,6 @@ fn override_if_distinct(stored: Option<&str>, fresh: String) -> Option<String> {
 }
 
 fn tmux_env_session_name_for_instance_id(instance_id: &str) -> Option<String> {
-    let suffix = format!("_{}", truncate_id(instance_id, 8));
     let output = crate::tmux::tmux_command()
         .args(["list-sessions", "-F", "#{session_name}"])
         .output()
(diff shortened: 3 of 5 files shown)


njbrake (approved):

Clean, tightly-scoped follow-up to the maintainability points from #3158.

R1 is a verified behavior-preserving dedup: the classification is a true partition (agent shape excludes term/cont/tool; term and cont have disjoint prefixes), so check order does not change bucketing, and the `agent > terminal > container` preference plus the `_<id8>` suffix semantics are unchanged. The `is_pane_dead` call-count parity holds; the new path probes strictly fewer or equal names than the old inline loop.

R2 trades one narrow sub-millisecond rename race for its mirror image while making the metadata lookup and the existence probe agree on a single resolved name. In steady state both old and new resolve to the same live name, so common and retitled cases are identical; the residual transient Error is documented and self-heals on the next poll.

Verified by reading both versions rather than trusting the description: built, ran the touched unit tests (9 passed, including both new ones), clippy clean on all five files, and confirmed the removed `truncate_id` import has no other use. CI is green and both Codecov gates pass.

PR #3161, opened 2026-07-30
fix(send): never submit or fuse with a human's unsent composer draft on injected sends

BTForIT:
## Description

When an external caller injects a message into a Claude pane (server `POST /api/sessions/{id}/send`, restart auto-resume, live send), and a human has **unsent text parked in the composer**, two things could go wrong:

1. **Fusion/submission of the draft** — the injected paste landed in the same composer as the parked draft and the submitting Enter sent *both*, publishing text the human never chose to send (or dropping their draft into a fused, garbled message).
2. **Dim autocomplete ghost text misread as a draft** — the inline suggestion Claude renders in dim style after the prompt was counted as composer content, so sends were refused against an actually-empty composer, and deliveries stalled.

This PR makes composer-draft detection style-aware and enforces refusal end-to-end:

- `src/tmux/session.rs`: `send_keys_verified` detects a non-empty composer before pasting and **refuses** instead of typing into it; the restart-to-paste race (pane not yet at a ready composer, folder-trust dialog, resume picker) waits/accepts/refuses instead of pasting blind.
- `src/tmux/status_detection.rs`: composer-content detection distinguishes real typed text from the dim autocomplete ghost (SGR dim run covering the run after the prompt marker), so an empty-composer-with-ghost pane is sendable.
- `src/server/api/sessions.rs`: the send endpoint surfaces the refusal as **423 Locked** with the composer excerpt, instead of reporting `{"sent":true}` for a send that fused with or overwrote a human draft.
- `src/tui/home/live_send.rs`: agent-pane literal runs are framed in bracketed-paste markers so the agent's paste-burst detection can't swallow the submitting Enter.

The two fixes share the same composer-detection substrate (the first two commits), which is why they travel in one PR: the refusal (fix 1) is only correct once the detector stops counting ghost text (fix 2), and vice versa.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: the behaviors are covered by unit tests against captured live pane frames (`src/tmux/status_detection.rs`, `src/tmux/session.rs`, `src/server/api/sessions.rs`, `src/tui/home/tests.rs`); reproducing a human-parked composer draft deterministically inside the e2e tmux harness would require racing keystrokes against the agent boot, which is exactly the nondeterminism the fix removes. Happy to add one if you see a deterministic shape for it.

How tested: `cargo fmt`, `cargo clippy --features serve` (clean), `cargo test --features serve --lib` (5896 passed; 1 pre-existing environment-dependent failure in `hooks_install::tests::test_content_shows_settings_path` under a non-default `CLAUDE_CONFIG_DIR`, unrelated). The fix is also running in production on our ~30-session fleet: phantom-ghost panes deliver, panes with real parked drafts refuse with 423 and the draft survives byte-identical.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Claude Fable 5)

Ported from our production fork where both defects were root-caused from live incident transcripts (a human's parked draft was fused into an injected send; separately, ghost text caused spurious refusals).

## Summary

- Prevents automated sends from overwriting or merging with unsent human drafts in Claude panes.
- Improves composer detection across multiple lines while ignoring autocomplete ghost text and interface elements.
- Adds safe send verification, including readiness checks, trust-dialog handling, retry behavior, and protection against unsafe re-pasting.
- Returns clear `423 Locked` responses when a draft blocks sending, without exposing draft contents.
- Uses bracketed-paste framing for agent-pane messages to prevent keystrokes from being swallowed.
- Adds extensive tests covering pane detection, send behavior, API contracts, restart handling, and TUI flows.

## Benefits

- Protects operator-written content from accidental overwrites.
- Makes automated sends safer and more reliable during restarts and UI transitions.
- Provides actionable, machine-readable refusal responses.
- Improves handling of Claude menus, dialogs, and partially completed submissions.

## Potential follow-up

- Continue validating behavior across different Claude UI layouts and environment-specific terminal behavior.

### DIFF
--- a/src/session/restart.rs
+++ b/src/session/restart.rs
@@ -81,6 +81,58 @@ fn should_send_restart_wake(outcome: &Result<StartOutcome, String>) -> bool {
     )
 }
 
+/// The single action the wake worker should take next, given the current pane
+/// contents. Computed by [`classify_wake_pane`] so the decision is a pure
+/// function we can unit-test without a live tmux pane.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WakeStep {
+    /// The pane is already generating (or for a non-resume start, the agent is
+    /// otherwise live) — the wake landed. Nothing left to do.
+    Done,
+    /// The pane is parked on Claude's resume-from-summary picker. Press Enter to
+    /// select the highlighted (recommended) option, then re-evaluate. A wake
+    /// *message* sent here would type into the menu instead of resuming.
+    DismissPicker,
+    /// The pane is at an idle prompt — or, crucially, showing a *stale*
+    /// prior-account usage-limit banner in scrollback after a cross-account
+    /// relocation. Send the wake message to spool generation. If the new account
+    /// is genuinely capped, generation simply won't start and we retry-then-give
+    /// up; the stale banner never blocks the attempt.
+    SendWake,
+    /// The wake message is sitting unsubmitted in Claude's composer: the paste
+    /// landed but the submitting Enter was swallowed by boot-time terminal-mode
+    /// churn (the post-restart first-message race). Press a bare Enter to
+    /// submit the existing draft. Re-sending the full message here (the old
+    /// behavior, via `SendWake`) doubled the text in the composer.
+    SubmitStuck,
+}
+
+/// Decide the next wake action from a pane capture. Pure + tool-aware so the
+/// retry loop in [`spawn_wake_worker`] stays trivial and the policy is tested.
+///
+/// Order matters: the resume picker is checked *first* and explicitly, because
+/// `detect_status_from_content` already collapses the picker to `Waiting`
+/// (alongside a real approval prompt and a stale limit banner) — and a `Waiting`
+/// pane otherwise routes to `SendWake`, which would dump the wake text into the
+/// menu. A wake message stuck unsubmitted in the composer is checked next
+/// (before the Running collapse: submitting it is correct even if a turn is
+/// already generating). Only an actively-*Running* pane is `Done`; every other
+/// state (Idle, or Waiting-because-stale-banner) sends the wake.
+fn classify_wake_pane(content: &str, tool: &str, wake_message: &str) -> WakeStep {
+    if tool == "claude" {
+        if crate::tmux::status_detection::claude_pane_has_resume_picker(content) {
+            return WakeStep::DismissPicker;
+        }
+        if crate::tmux::status_detection::claude_message_stuck_in_composer(content, wake_message) {
+            return WakeStep::SubmitStuck;
+        }
+    }
+    match crate::tmux::status_detection::detect_status_from_content(content, tool) {
+        crate::session::Status::Running => WakeStep::Done,
+        _ => WakeStep::SendWake,
+    }
+}
+
 /// Wait for the restarted pane to become live and past its boot shell, then
 /// send the wake-up message. Best-effort: a failure to spawn or send is logged,
 /// never fatal.
@@ -108,12 +160,110 @@ fn spawn_wake_worker(session_id: String, title: String, tool: String, wake_messa
                 std::thread::sleep(std::time::Duration::from_millis(50));
             }
 
-            if !tmux_session.exists() {
-                return;
-            }
+            // Phase 2 — guarded auto-resume: classify, act, verify, retry.
+            // After a cross-account boot Claude may still be loading well past the
+            // 3s readiness wait, may present the resume-from-summary picker, or may
+            // sit idle showing the prior account's stale cap banner. A single
+            // keystroke can't survive any of those; this loop re-checks the pane
+            // after each action so a wake that didn't take is resent.
+            const WAKE_CAPTURE_LINES: usize = 200;
+            const MAX_WAKE_SENDS: u32 = 3;
+            const MAX_PICKER_DISMISSALS: u32 = 3;
+            // After Enter on the picker, give Claude a beat to load the summary.
+            const PICKER_SETTLE: std::time::Duration = std::time::Duration::from_millis(700);
+            // After a wake send, give generation a chance to start before re-polling.
+            const WAKE_VERIFY: std::time::Duration = std::time::Duration::from_millis(1500);
+            // Hard wall-clock cap so a permanently-stuck pane can't pin the thread.
+            let overall_deadline =
+                std::time::Instant::now() + std::time::Duration::from_secs(20);
+
             let delay = crate::agents::send_keys_enter_delay(&tool);
-            if let Err(e) = tmux_session.send_keys_with_delay(&wake_message, delay) {
-                tracing::warn!(target: "session.restart", "failed to send wake-up message after restart: {}", e);
+            let mut wake_sends: u32 = 0;
+            let mut picker_dismissals: u32 = 0;
+            let mut stuck_submits: u32 = 0;
+
+            while std::time::Instant::now() < overall_deadline {
+                if !tmux_session.exists() {
+                    return;
+                }
+                let content = tmux_session
+                    .capture_pane(WAKE_CAPTURE_LINES)
+                    .unwrap_or_default();
+                match classify_wake_pane(&content, &tool, &wake_message) {
+                    WakeStep::Done => {
+                        tracing::info!(
+                            target: "session.restart",
+                            session_id = %session_id,
+                            wake_sends,
+                            picker_dismissals,
+                            "restart wake confirmed: pane is generating"
+                        );
+                        return;
+                    }
+                    WakeStep::DismissPicker => {
+                        if picker_dismissals >= MAX_PICKER_DISMISSALS {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: resume picker persisted after {MAX_PICKER_DISMISSALS} dismissals; giving up"
+                            );
+                            return;
+                        }
+                        picker_dismissals += 1;
+                        // Bare Enter selects the highlighted (recommended) option;
+                        // send raw so no message text leaks into the menu.
+                        if let Err(e) = tmux_session.send_raw_bytes(b"\r") {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: failed to dismiss resume picker: {e}"
+                            );
+                        }
+                        std::thread::sleep(PICKER_SETTLE);
+                    }
+                    WakeStep::SubmitStuck => {
+                        if stuck_submits >= MAX_WAKE_SENDS {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: message still stuck in composer after {MAX_WAKE_SENDS} submits; parked"
+                            );
+                            return;
+                        }
+                        stuck_submits += 1;
+                        // The wake text already landed in the composer; only its
+                        // Enter was swallowed by the still-booting TUI. Submit it
+                        // with a bare Enter, never a re-paste, which would double
+                        // the message text.
+                        if let Err(e) = tmux_session.send_raw_bytes(b"\r") {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: failed to submit stuck composer message: {e}"
+                            );
+                        }
+                        std::thread::sleep(WAKE_VERIFY);
+                    }
+                    WakeStep::SendWake => {
+                        if wake_sends >= MAX_WAKE_SENDS {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: pane not generating after {MAX_WAKE_SENDS} wake sends; parked"
+                            );
+                            return;
+                        }
+                        wake_sends += 1;
+                        if let Err(e) = tmux_session.send_keys_with_delay(&wake_message, delay) {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: failed to send wake-up message: {e}"
+                            );
+                        }
+                        std::thread::sleep(WAKE_VERIFY);
+                    }
+                }
             }
         });
     if let Err(err) = spawn_result {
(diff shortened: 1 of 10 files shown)


src/session/restart.rs
```diff
@@ -81,6 +81,58 @@ fn should_send_restart_wake(outcome: &Result<StartOutcome, String>) -> bool {
     )
 }
 
+/// The single action the wake worker should take next, given the current pane
+/// contents. Computed by [`classify_wake_pane`] so the decision is a pure
+/// function we can unit-test without a live tmux pane.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WakeStep {
+    /// The pane is already generating (or for a non-resume start, the agent is
+    /// otherwise live) — the wake landed. Nothing left to do.
```
Seluj78, line 90: Blocker: em dash in doc comment (also at lines 96, 115, 163 in this file). Rule: AGENTS.md Coding Style bans em dashes in docs/comments.

src/session/restart.rs
```diff
@@ -108,12 +160,110 @@ fn spawn_wake_worker(session_id: String, title: String, tool: String, wake_messa
...
+                        if wake_sends >= MAX_WAKE_SENDS {
+                            tracing::warn!(
+                                target: "session.restart",
+                                session_id = %session_id,
+                                "restart wake: pane not generating after {MAX_WAKE_SENDS} wake sends; parked"
+                            );
+                            return;
+                        }
+                        wake_sends += 1;
+                        if let Err(e) = tmux_session.send_keys_with_delay(&wake_message, delay) {
```
Seluj78, line 257: Required: WakeStep::SendWake still calls send_keys_with_delay, not send_keys_verified. An unrelated human draft parked in the composer when this fires can still be fused with the wake message and submitted under the operator's name, the same WO#453 defect this PR fixes elsewhere. classify_wake_pane only guards the resume picker and the wake message's own stuck-composer state, never an unrelated draft. (Corroborates the unresolved CodeRabbit thread at restart.rs:265.)

src/tmux/session.rs
```diff
@@ -1068,6 +1068,151 @@ impl Session {
...
+        const DRAFT_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+        const VERIFY_SETTLE: std::time::Duration = std::time::Duration::from_millis(1000);
+        const MAX_SUBMIT_RETRIES: u32 = 3;
+        const VERIFY_CAPTURE_LINES: usize = 30;
+
+        // Phase 1: wait for the composer. On timeout proceed anyway; a pane in
+        // a state the detector doesn't recognize (dialog, alt screen) must
+        // still get the send, and the verify phase below recovers the Enter.
+        // Exception: the folder-trust dialog (first boot in a never-seen dir)
+        // is modal and never self-dismisses — pasting into it types into the
```
Seluj78, line 1101: Blocker: em dash in comment (also at line 1165 in this file). Rule: AGENTS.md Coding Style bans em dashes in docs/comments.

src/tmux/session.rs
```diff
@@ -1068,6 +1068,151 @@ impl Session {
...
+            );
+            self.send_raw_bytes(b"\r")?;
+        }
+        let content = self.capture_pane(VERIFY_CAPTURE_LINES).unwrap_or_default();
+        if super::status_detection::claude_message_stuck_in_composer(&content, text) {
+            tracing::warn!(target: "tmux.command",
+                "send_keys_verified: message still unsubmitted after {MAX_SUBMIT_RETRIES} Enter resends"
+            );
+        }
+        Ok(())
```
Seluj78, line 1213: Required: after MAX_SUBMIT_RETRIES bare-Enter resends are exhausted and claude_message_stuck_in_composer is still true, this returns Ok(()) anyway. The caller (send_message in src/server/api/sessions.rs) then reports {"sent": true} for a send whose message may still be unsubmitted, the ambiguous-outcome problem this PR's ParkedDraftRefusal/423 contract exists to eliminate. (Corroborates the unresolved CodeRabbit thread at session.rs:1213.)

src/tmux/status_detection.rs
```diff
@@ -555,8 +555,313 @@ fn claude_pane_shows_ready_prompt(
...
+        .filter(|l| !l.trim().is_empty())
+        .collect::<Vec<&str>>()
+        .iter()
+        .rev()
+        .take(CLAUDE_COMPOSER_TAIL)
+        .any(|line| {
+            let trimmed = line.trim();
+            // Claude Code renders the prompt as `❯` + U+00A0 NO-BREAK SPACE
+            // when text follows (captured live, WO#453), so a literal `"❯ "`
+            // match misses any composer holding a draft — accept `❯` followed
```
Seluj78, line 622: Blocker: em dash in comment (also at lines 651, 712, 718-719, 788, 2935, 3040, 3154 in this file). Rule: AGENTS.md Coding Style bans em dashes in docs/comments.

src/tui/home/live_send.rs
```diff
@@ -406,6 +406,55 @@ pub(super) fn coalesce(batch: Vec<WorkerMsg>) -> Vec<TmuxAction> {
...
+/// only milliseconds. Mirrors the daemon send path's
+/// `PASTE_BYTE_THRESHOLD` in `tmux::session` so both injection surfaces
+/// agree on what "paste-shaped" means.
+pub(super) const LIVE_SEND_PASTE_THRESHOLD: usize = 16;
+
+/// Frame machine-speed literal runs in bracketed-paste markers for an
+/// agent pane. An unframed burst trips the agent CLI's own paste-burst
+/// detector (Claude Code buffers rapid input to guess at pastes) and the
+/// user's submitting Enter arriving inside that window is coalesced into
+/// the buffered text as a newline instead of firing submit — the typed
```
Seluj78, line 431: Blocker: em dash in doc comment. Rule: AGENTS.md Coding Style bans em dashes in docs/comments.

Seluj78 (changes requested):

## AoE Review

Reviewed: PR #3161 (base: main@4f4f3b91). Files changed: 10 (tmux, server, session, cli, tui, rust-tests).
Unresolved comments: 2 (2 bot, 0 human).

### Unresolved comments

- coderabbitai[bot], src/session/restart.rs:265: "Route restart wake sends through send_keys_verified when they can hit Claude panes." Corroborates the Required finding below.
- coderabbitai[bot], src/tmux/session.rs:1213: "Document send_keys_verified's exhausted-send fallback." Corroborates the Required finding below.

### Blocker

- Em dash in 16 newly-added doc comments/comments: src/session/restart.rs:90,96,115,163; src/tmux/session.rs:1101,1165; src/tmux/status_detection.rs:622,651,712,718-719,788,2935,3040,3154; src/tui/home/live_send.rs:431.
  Rule: AGENTS.md, Coding Style: "No emdashes or `--` as separators in docs/comments."

### Required

- src/session/restart.rs:257: WakeStep::SendWake still calls send_keys_with_delay, not send_keys_verified. An unrelated human draft parked in the composer when the restart-wake worker fires can still be fused with the wake message and submitted under the operator's name, the same WO#453 defect this PR fixes for the daemon/CLI send paths. classify_wake_pane only guards the resume picker and the wake message's own stuck-composer state, never an unrelated draft.

- src/tmux/session.rs:1213: after MAX_SUBMIT_RETRIES bare-Enter resends are exhausted and claude_message_stuck_in_composer is still true, send_keys_verified logs a warning and returns Ok(()) anyway. The caller (send_message in src/server/api/sessions.rs:8840) then reports {"sent": true} for a send whose message may still be unsubmitted, the ambiguous-outcome problem this PR's ParkedDraftRefusal/423 contract exists to eliminate.

- branch pr/composer-draft-send-refusal does not follow the feature/ / fix/ / docs/ / refactor/ convention (AGENTS.md, Commit & Pull Request Guidelines).

- Em dash in PR body (2 lines) and in two commit subjects/bodies (42a63f6d, 0df79354). Same AGENTS.md rule applies to PR metadata.

- CI: no cargo fmt / clippy / test / build workflow has run for this branch (cross-repo/fork PR, mergeStateStatus BLOCKED). Only check-template and PR-title-lint have run, so cargo fmt/clippy/test cannot be verified yet.

### Followup

- Claude-only composer-draft guard: send_keys_verified (src/tmux/session.rs:920) and classify_wake_pane (src/session/restart.rs) both gate on tool == "claude". Codex, Gemini, and other interactive-composer agents fall through to plain send_keys_with_delay and remain exposed to the same draft-fusion/resume-race bug class this PR fixes for Claude. Suggest a follow-up issue scoped to a per-agent composer detector.

- src/cli/session.rs:1327 (restart_session): the CLI's `aoe session restart` command implements its own separate restart-wake send via plain send_keys_with_delay, bypassing this PR's new classify_wake_pane/spawn_wake_worker guard (used only by the TUI restart poller). Suggest a follow-up issue to route the CLI path through the guarded worker.

### Verdict

Implementation: Request Changes. The two independently-confirmed correctness gaps (unrefused fusion in the restart SendWake path; silent Ok(()) on exhausted-retry non-delivery) undermine the sent/not-sent guarantee this PR is built to establish, and CI has not yet run for this fork PR.

Project Fit: Accept. The bug is real and well-evidenced (production incident transcripts, WO#453/WO#904, already deployed on the author's fork with before/after data), the scope matches the fix, and test coverage is thorough (captured live-pane fixtures for the edge cases).

The implementation direction is correct and the fix is needed, but it should not merge while its own headline guarantee, that the daemon knows definitively whether a send fused with or reported success against a still-parked message, is broken in two of its own new code paths.

njbrake (changes requested):

The bug class is real and the production transcripts are convincing. The architectural core is also better than what is there now and worth landing: a typed refusal that survives the `anyhow` boundary, and a pure `send_error_parts` so the wire contract is testable without a daemon. 423 with `retry_when: "composer_clear"` is an accurate status choice, and withholding the draft text while reporting only size is the right instinct. I agree with all seven of @Seluj78's points and will not repeat them.

Two things mean the PR does not currently hold its own headline guarantee.

**1. `send_keys_verified` reports success for a message it can see is unsent.** After `MAX_SUBMIT_RETRIES` bare-Enter resends, if `claude_message_stuck_in_composer` is still true it logs a warning and returns `Ok(())`, and `send_message` then answers `200 {"sent": true}`. That is the ambiguity this PR exists to remove, and your own `test_only_the_refusal_claims_certainty_that_nothing_was_sent` asserts that only definitive outcomes may claim certainty. A blind retry is also unsafe there, because the text is already in the composer, so a retry doubles it. This needs to be an error rather than a documented fallback.

**2. Two of the three named callers still fuse.** `restart.rs:257` (`WakeStep::SendWake`) calls `send_keys_with_delay`, and `classify_wake_pane` guards only the resume picker and the wake message's own stuck state, so an unrelated operator draft gets fused and submitted; `classify_unrelated_composer_draft_is_not_stuck` documents that as intended. The TUI live-send path at `home/mod.rs:5144` also still uses `send_keys_with_delay`. The description names both as paths this fixes.

**Two findings from reproducing the detectors in isolation.** I extracted the seven functions verbatim and compiled them standalone so I could check them against real pane shapes rather than reasoning about them.

The draft detector's chrome exclusion list is a fail-closed allowlist of size two: dim SGR spans, plus the literal string `"Press up to edit queued messages"`. Any non-dim, non-numbered text in the composer region reads as an operator draft:

```
non-dim colored placeholder hint    -> draft=Some("Try \"how does X work?\"")
italic-only hint (SGR 3, not dim)   -> draft=Some("? for shortcuts")
```

The consequence is not cosmetic. A Claude Code release that renders a composer placeholder with a color rather than dim turns every send to every claude pane into a permanent 423 after a 10 second hang, with no operator remedy since the composer is already clear, and the refusal is undiagnosable from the log because the draft text is deliberately withheld. You already hit one instance of this class and patched it with a string literal, which is evidence the class recurs rather than that it is closed.

At minimum, log `draft_chars` and whether the composer region was `closed` on every refusal, so a chrome misread is distinguishable from a real draft without a live pane. Better would be a bounded consecutive-refusal counter that fails open with a loud warning, so an upstream UI change degrades to today's behavior instead of bricking sends.

Relatedly, the dim strip is fail-open, so the bypass it claims to close can reopen: a pasted-text chip rendered dim returns `draft=None`, and the fixture for that case is written without ANSI. An unclosed dim span, or a tmux `pane-style` that dims an inactive pane, silently disables the whole guarantee. That is the pre-PR behavior so it is not a regression, but the doc comments present the guarantee as absolute when it is conditional.

Separately, `claude_trust_dialog_visible` has a reproducible false positive that causes an unsolicited Enter injection. The option-line branch strips an *optional* `❯` and then matches `starts_with("1. yes, i trust this folder")`, so a bare numbered list in an agent's own prose trips it, and when that pane also has no visible composer the code sends `\r` into it. An agent writing docs about the trust dialog, or reading this PR, produces that line. Requiring the `❯` selection cursor is strictly tightening: I checked that all four existing positive fixtures carry it and the false positive goes away. There is also a principle worth deciding separately, since auto-accepting "Yes, I trust this folder" is a security-relevant grant now made by any holder of the API token, on a send, gated by no setting.

**One unflagged cost.** Phase 1 polls the full 7 second `READY_TIMEOUT` whenever `claude_pane_input_ready` is false, which I reproduced as false for a tool-permission approval prompt, an `AskUserQuestion` menu, a plain shell prompt after the agent exited, and any alt-screen view. Answering an approval prompt is one of the most common reasons an orchestrator calls this endpoint, so it now costs 7 seconds of polling plus the 1 second `VERIFY_SETTLE`. The happy path pays that 1 second unconditionally on every claude send. Compounding it, `send_message` holds `state.instance_lock(&id)` across the whole blocking send, so a session with a permanently parked draft holds its per-id lock for 10 seconds per request while `retry_safe: true` invites the caller to hammer it.

The swallowed-Enter recovery also misses the messages most likely to need it: `claude_message_stuck_in_composer` matches a 32-char prefix of the first line against a prompt row, but `send_keys_with_delay` routes anything 16 bytes or longer, or containing a newline, through `paste-buffer`, and Claude collapses multi-line pastes to `[Pasted text #N +M lines]`. For those the composer never shows the text, so phase 2 returns on its first check regardless of whether the Enter landed. Panes narrower than about 35 columns and messages beginning with a newline also fall through.

**Scope.** Please split the `live_send.rs` bracketed-paste framing into its own PR. It is unrelated to the stated purpose, and its 16-byte threshold decides whether to inject escape sequences the receiving process may not understand, which is a different question from the daemon transport threshold it mirrors. I would also drop `INJECT_DRAFT_DELIMITER`, `reject_forged_boundary`, and their five tests: they guard against a marker upstream AoE has never written, and `reject_forged_boundary` runs before the `tool != "claude"` early return, so it now hard-fails any message from any tool containing that exact string.

**Housekeeping.** 16 em dashes and 12 `--` separators in added comments, both of which AGENTS.md forbids; also in the PR body and in commits `42a63f6d` and `0df79354`. The branch should be `fix/composer-draft-send-refusal`. `docs/api.md:52` documents this endpoint and needs the new 423 and the `parked_draft` code. And the roughly 45 new one-liner tests want consolidating into tables per the test-cost section in AGENTS.md; the `claude_pane_input_ready_*` and `claude_composer_draft_none_on_*` families are about 20 functions that fit in two.

I have approved the workflow runs, so we will have CI shortly. Note the changes touch the `serve` feature, so `--features serve` matters here.

njbrake:

Closing as stale. The July 31 review (my two points plus @Seluj78's seven) is still unaddressed and the branch hasn't moved since July 30.

To be clear about why: the bug class is real, the production transcripts were convincing, and the architecture is better than what's on main today. I'd still like this to land. Push the fixes and reopen, or ping me and I'll reopen it. Thanks.

PR #3165, opened 2026-07-30
fix(session): avoid nested-runtime panic in opencode preassign

jakshi:
## Description

opencode sessions crash on launch when I set `session.opencode_preassign_session_id = true`:

```
thread 'main' panicked at src/session/capture.rs:
Cannot start a runtime from within a runtime. This happens because a function
(like `block_on`) attempted to block the current thread while the thread is
being used to drive asynchronous tasks.
```

`aoe add --launch`, `aoe session start`, `aoe session restart`, and creating a session in the TUI all hit it. claude and copilot launch fine — only opencode breaks. I hit it on v1.13.1, and it's still there on v1.13.2 and `main`.

### Why it happens

`preassign_opencode_session_id_impl` (`src/session/capture.rs`) spawns a throwaway `opencode serve`, builds its own current-thread Tokio runtime, and blocks on a `POST /api/session` to reserve the `ses_` id:

```rust
let rt = tokio::runtime::Builder::new_current_thread().enable_all().build()?;
rt.block_on(async { /* wait for serve, POST the id */ })
```

The comment above it says this "never runs inside a live async task." It does. `src/main.rs` is `#[tokio::main]`, so the CLI is already inside a runtime. The launch path calls `acquire_session_id` → `preassign_opencode_session_id`, builds a second runtime on the same thread, and panics.

Only opencode hits it — `src/session/instance.rs` gates preassign to opencode:

```rust
let preassign = self.tool == "opencode" && self.opencode_preassign_enabled();
```

Setting the flag to `false` avoids the crash, but I use preassign to keep opencode sessions from cross-wiring when several agents share a directory. I want it on.

### The fix

Run the runtime on a dedicated OS thread with `std::thread::scope`. A fresh OS thread has no ambient runtime, so `block_on` is safe no matter who calls it — the CLI, a server `spawn_blocking` worker, or the TUI event loop.

- The scoped thread borrows `id` / `base` / `project_path` — no `'static` clones.
- The `opencode serve` guard stays alive across the join.
- The preassign protocol doesn't change.

### How I tested

- `cargo build --release`, `cargo fmt --check`, `cargo clippy`.
- `aoe session start` on an opencode session with preassign on: no crash. It panicked every time before.
- Added an e2e test, `tests/e2e/opencode_preassign_no_runtime_panic.rs`. It fails on the old code with the same panic and passes with the fix.
- copilot still launches fine.

One note on the diff: it looks big, but it's mostly indentation — the async block moves one level deeper into the closure. The logic is the same.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

**NOTE:**

## Summary

- Fixed OpenCode launch crashes when session ID preassignment runs inside an existing async runtime.
- Moved preassignment work to a dedicated thread while preserving the existing fallback behavior.
- Added a parallel-safe end-to-end test that verifies successful launch and `opencode serve` invocation.

## Benefits

- Prevents nested-runtime panics.
- Improves confidence in OpenCode preassignment across launch paths.
- Keeps launches resilient when the preassignment server is unavailable.

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -1599,58 +1599,74 @@ fn preassign_opencode_session_id_impl(project_path: &str) -> Result<String> {
     let _guard = ServeGuard(Some(child));
 
     let base = format!("http://127.0.0.1:{port}");
-    // acquire_session_id runs on a synchronous launch thread (the CLI process,
-    // a server `spawn_blocking` worker, or the TUI event loop), never inside a
-    // live async task, so a short-lived current-thread runtime is safe here.
-    let rt = tokio::runtime::Builder::new_current_thread()
-        .enable_all()
-        .build()
-        .context("failed to build the preassign runtime")?;
-
-    rt.block_on(async {
-        let client = reqwest::Client::builder()
-            .timeout(Duration::from_secs(2))
+    // `acquire_session_id` runs on a launch thread that may itself be async: on
+    // the CLI it runs under the `#[tokio::main]` entrypoint, i.e. *inside* a live
+    // Tokio runtime. Building a runtime and `block_on`-ing it on that same thread
+    // panics with "Cannot start a runtime from within a runtime". Run the
+    // short-lived current-thread runtime on a dedicated OS thread instead, which
+    // never carries an ambient runtime, so `block_on` is valid regardless of
+    // whether the caller (CLI, a server `spawn_blocking` worker, or the TUI event
+    // loop) is itself async. `thread::scope` lets the worker borrow
+    // `id`/`base`/`project_path` without `'static` clones and keeps the
+    // `opencode serve` `_guard` alive across the join.
+    let preassign = || -> Result<()> {
+        let rt = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
             .build()
-            .context("failed to build the preassign HTTP client")?;
+            .context("failed to build the preassign runtime")?;
 
-        let deadline = Instant::now() + OPENCODE_PREASSIGN_DEADLINE;
-        loop {
-            if let Ok(resp) = client.get(format!("{base}/api/session")).send().await {
-                if resp.status().is_success() {
-                    break;
+        rt.block_on(async {
+            let client = reqwest::Client::builder()
+                .timeout(Duration::from_secs(2))
+                .build()
+                .context("failed to build the preassign HTTP client")?;
+
+            let deadline = Instant::now() + OPENCODE_PREASSIGN_DEADLINE;
+            loop {
+                if let Ok(resp) = client.get(format!("{base}/api/session")).send().await {
+                    if resp.status().is_success() {
+                        break;
+                    }
                 }
+                if Instant::now() >= deadline {
+                    anyhow::bail!("opencode serve did not become ready within the deadline");
+                }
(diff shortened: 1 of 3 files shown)


njbrake (approved):

Approving, and thank you for the clear write-up. The diagnosis is correct and the fix is the most general of the available options.

I verified the load-bearing fact in the tokio 1.52.4 source rather than reasoning about it: `enter_runtime` reads `CONTEXT.with(|c| c.runtime)`, a thread-local, and panics only if it is already `Entered`, so `block_on` on a freshly spawned OS thread is unconditionally safe regardless of runtime flavor or caller. That makes the dedicated thread strictly more general than the repo's existing `Handle::try_current` plus `block_in_place` idiom, which panics on `current_thread` and needs a three-arm match.

The part I want to call out is that this also dodges a second landmine. `Runtime::drop` panics with "Cannot drop a runtime in a context where blocking is not allowed", so wrapping only `rt.block_on(...)` in `block_in_place` would have traded one panic for another once `rt` dropped after `exit_runtime` restored the flag. Putting both the build and the drop on the worker thread avoids it. That is the classic second-order error in this fix category and you got it right.

I also compiled the exact `thread::scope` shape (closure borrowing the three locals, `scope.spawn(...).join().map_err(...)?`, a `Drop` guard declared before the scope, returning the borrowed `String` after) to confirm the guard drops after the join and that a worker panic becomes `Err` rather than aborting the process. That last part is the right fail-closed behavior for a helper documented as returning `None` on any failure.

I checked for other latent sites and found none left unfixed: the remaining `block_on` / `Runtime::new` / `block_in_place` hits in `src/` are `tui/app.rs` (already on `std::thread::spawn` or already `block_in_place`) and `acp/supervisor.rs` (flavor-gated). No `reqwest::blocking` or `futures::executor::block_on` anywhere.

One correction to the description: TUI session *creation* is safe, because it defers tmux start to attach time and `creation_poller.rs` runs on a plain `std::thread::spawn` with no ambient runtime. It is the TUI launch and attach path that panicked. Cosmetic, and the CLI entries you list are all correct.

I pushed the two nits myself in 21c364a5 so this could land: `#[serial]` to `#[parallel]` (this test mutates nothing process-global, and a default-key `#[serial]` is excluded against every `#[parallel]` test in the binary, so it would have stalled the whole 3-thread e2e suite for its ~7s), three `--` prose separators, and the `#[cfg(not(unix))]` harness helper (`new_in_tmp` exists to keep ACP worker socket paths under Darwin's `sun_path` limit; this test binds a TCP port and spawns no worker).

Note that pushing to your fork reset the workflow gate, so I re-approved the runs. This is the first time any Rust job has executed on this PR, and the e2e test in particular had never run on a CI runner before now.

One idea for later, not a request: if the fake `opencode` bound its port and answered `POST /api/session`, the test would drop from ~7s to a couple hundred milliseconds and upgrade from "the panic is gone" to "preassign succeeds and the launch carries `--session ses_...`", which nothing covers today. There is precedent in `web/tests/helpers/fakeAcpAgent.mjs` behind `require_node!`.

Issue #2287, opened 2026-06-19
Structured view: live updates silently stall after WS idle-reset behind a proxy; only resync on user action (apparent regression of #1130)


gilbertl:
**Note:** Investigated with Claude Code assistance against a live 1.11.1 deployment; decisions and report are mine.

Looks like an apparent regression / incomplete coverage of #1130 (which fixed `useCockpit` reconnect for the old cockpit). On the **1.11.1 structured view**, the live WebSocket still dies on idle and the UI does not auto-reconnect — with a subtler symptom than #1130.

## Symptom
On a remote (behind-proxy) deployment, returning to a session tab shows **no new messages** even though the agent has been producing output. There's **no obvious "disconnected" indicator** — the view just silently stops updating. **Sending a message (or any action) triggers a reconnect**, and all buffered events flush in at once. No data is lost (events are persisted and replayed); the live view is simply stale until the user acts.

## Environment
- aoe `1.11.1` (`1.11.1+g9d37e6bee953`), Linux host
- `aoe serve --auth=passphrase --behind-proxy`, fronted by **Tailscale Serve** (HTTPS tailnet → `http://127.0.0.1:8080`)
- Clients: Chrome/Edge 149 (Windows desktop) and Android Chrome (PWA)

## Repro
1. Run `aoe serve --behind-proxy` behind a reverse proxy with a WS idle timeout (Tailscale Serve, Cloudflare, etc.).
2. Open a session in the structured view; leave the tab idle/backgrounded while the agent keeps emitting output.
3. Return to the tab → no new messages.
4. Send a message → view reconnects and all missed messages appear together.

## Evidence
Daemon log shows the browser WS being reset abruptly, repeatedly (33× over ~3 days in one `debug.log`):

```
WARN acp.ws: client recv error: WebSocket protocol error: Connection reset without closing handshake
```

"Reset without closing handshake" = the proxy/network dropped the idle connection (RST), not a clean close — consistent with proxy idle-timeout and/or background-tab socket suspension, the same mechanism as #1130 (there: Cloudflare 100s idle + mobile background; here: Tailscale Serve + desktop/mobile idle).

## Suspected cause
1. The structured-view WS client has no reconnect path on `close`/`error`, and no `visibilitychange` / `online` listener — so it only re-opens when the user switches sessions or hard-refreshes (the same gap #1130 fixed for `useCockpit`, now present on the structured-view hook).
[...]

PR #3171, opened 2026-07-31
fix: four bugs found in a live debug.log (ws heartbeat, phantom status transitions, wedged trash purge, worktree move gate)

njbrake:
_Note: this PR description was drafted by Claude via back-and-forth with @njbrake. The investigation and decisions are his; the prose is Claude's._

## Description

Four independent bugs found by reading a `debug.log` from a live session. Each is a separate commit with its own regression test, so any one can be reverted alone.

**1. `fix(acp)`: the native TUI structured view dropped its WebSocket every 30 seconds.**

The daemon emits `{"kind":"heartbeat"}` on every ping tick (#2287, `PING_INTERVAL` = 30s). The Rust client's `parse_text` only probed for the `lagged` sentinel, so the heartbeat fell through to the `AcpBroadcastFrame` parse and failed on the missing `session_id` field. `tui::structured_view` treats a parse error as a dropped socket, so every quiet session showed an error toast, cleared `in_flight`, and reconnected, every 30 seconds, indefinitely:

```
WARN acp.client.ws: ws frame parse failed error=missing field `session_id` at line 1 column 20
WARN acp.tui.ws: ws disconnect: failed to parse websocket frame
```

`{"kind":"heartbeat"}` is 20 characters, which is where "column 20" comes from. The web client already handled this shape (`useAcpSession.ts`); only the Rust client was never taught about it. Note `heartbeat_frame_shape_is_stable` in `src/server/acp_ws.rs` guards the server's wire shape and its comment predicts this exact failure, but nothing guarded the Rust client's ability to read it. There is now a test on both sides.

**2. `fix(session)`: phantom `Running -> Idle` every 2 seconds on ACP sessions.**

Thousands of identical lines for a single session:

```
INFO session.status_change: aed2f79faaf04082 [claude] Running -> Idle (hook=None hook_age_ms=None pane=empty_capture)
```

These were not transitions. A structured session's status is owned by the ACP overlay, and `update_status_with_metadata_inner` returns early for those rows, so `status` on exit is the value loaded from disk while `live_status_baseline` is the overlay's live value. Diffing the two compares an overlay-derived value against a stale disk read, and `decide_passive_transition` returns `patch: None` for structured rows by design, so nothing reconciles disk and the mismatch is permanent.

Beyond log volume, the phantom `StatusChange` reached `push::handle_status_change`, which re-armed `idle_since` on every tick and raced the 2s `DWELL_TERMINAL_MS` it is meant to enforce. The "session went idle" push therefore fired or not depending on tick jitter. The idle reaper is unaffected; it already filters `is_structured()`.

Fixed at both exits. The `Error -> Idle` heal, the one change `_inner` genuinely makes for these rows, is still observed, and there is a test asserting that so the guard cannot regress into a blanket skip.

**3. `fix(git)`: trash auto-purge wedged in a permanent hourly retry.**

```
WARN http.api.sessions: auto-purge of expired trash failed: Worktree: Git worktree command failed:
  fatal: '.../.aoe-trash/822ece2dad8c46c5' is not a working tree
```

Same session, same failure, every hour, no progress. The checkout is on disk with a dangling `.git` pointer while git's admin entry is gone, so `git worktree remove` refuses and retrying can never help. The error fell through to the generic arm. Now classified by `is_not_a_worktree_error` and recovered the way a missing `.git` already is: remove the directory by hand, then prune.

**4. `fix(session)`: `git worktree move` failing with Permission denied.**

The rename gate tested `probe_running()` alone, but a stopped-but-not-removed container still pins its bind mount. On Docker Desktop for macOS the bind is re-exported through the VM's file-sharing layer, so `rename(2)` fails with `EACCES` rather than the `EBUSY` the gate's docstring predicted. Timeline from the log:

```
23:48:49  stopping container aoe-sandbox-aed2f79f
23:48:59  git worktree move -> fatal: failed to move ...: Permission denied
23:56:39  removing container
23:56:40  "removed stale sandbox container after worktree move" -> the same move succeeded
```

`sandbox_container_holds_worktree` becomes `ensure_sandbox_container_released`, splitting the two conflated cases: a running container still blocks (we cannot yank a live agent's mount), while a stopped one is discarded on the spot, which frees the mount. Renamed rather than given a new parameter so the compiler found every caller; all five gates shared the bug, including the workdir-rename flow in the TUI, CLI and web API.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (N/A, no visual change; see the note below)

On tests: `cargo fmt --check` clean, `cargo clippy --features serve --all-targets` adds no new warnings (the 7 reported all reproduce on a clean `main`), `cargo test --features serve` is 5845 passed / 2 failed. The 2 failures are pre-existing and environmental, not from this branch: `shutdown_and_wait_degrades_when_load_errs_without_peer` and `pid_source_for_falls_back_to_peer_pid_on_load_err` both force a read error with `chmod 0o000`, which is a no-op when the test runs as uid 0. Both fail identically on a clean checkout of `main` in the same container.

Each of the four regression tests was verified to fail without its fix. Two of them reproduce the logged error string verbatim: the heartbeat test panics with `Parse("missing field \`session_id\` at line 1 column 20")` and the trash test with `fatal: '...' is not a working tree`.

Documentation: no user-facing docs changed. The reasoning for each fix lives in the code, on the functions whose contracts changed.

On the screenshot box: nothing renders differently. The one visible effect is the *absence* of a recurring error toast in the TUI structured view, which needs a live daemon and a 30s ping tick to demonstrate, so a recording would be 30 seconds of a session correctly doing nothing. Say the word if you want it anyway.

## Test Coverage Analysis

- [x] Skipped a test, because: no `web/` files change, so the matrix gate does not apply, and the dashboard renders the same status either way (the web session list reads the ACP overlay, which is untouched). What commit 2 does change is push-notification timing, which needs a real push subscription and a dwell window to observe and is not reachable from Playwright. Flagging it rather than claiming N/A because the notifications surface is on the mandate list.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the natural e2e for the heartbeat fix is "attach the structured view to a live daemon and assert no disconnect toast appears", which needs 30+ seconds of wall clock to see the first ping tick. Covered instead from both ends by unit tests pinned to the same literal wire bytes: `heartbeat_frame_shape_is_stable` on the server and `parse_text_heartbeat_is_ignored_not_an_error` on the client, so drift on either side fails one of them. Happy to add the slow e2e if you would rather pay the 30s.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Starting point was a pasted `debug.log`. Every root cause was traced to specific code and confirmed there rather than inferred from the log text, and every fix has a regression test that was checked to fail without it. Worth a close look from a human on two points: whether skipping structured rows in the `status_poll_loop` diff loses a `StatusChange` consumer I did not account for (I traced `apply_status_intent` as the ACP-driven broadcaster that covers those rows), and whether discarding a stopped container inside the rename gate is acceptable in the workdir-rename flows, not just the trash flows, since the rename made it apply to all five call sites.

## Summary

- Ignored ACP WebSocket heartbeat frames without causing parse errors or reconnects.
- Removed orphaned worktree directories during trash cleanup.
- Released stopped sandbox containers only when a worktree move is required.
- Avoided force-removing containers that may restart during cleanup.
- Added regression tests for the fixes.

## Benefits

These changes reduce unnecessary reconnects and cleanup retries. They also make worktree operations safer and more predictable. Formatting, clippy, and documented tests pass. Two permission-test failures remain due to the test environment running as root.

### DIFF
--- a/src/acp/client/ws.rs
+++ b/src/acp/client/ws.rs
@@ -1,13 +1,20 @@
 //! WebSocket client for the structured view broadcast stream.
 //!
 //! Subscribes to `/sessions/{id}/acp/ws?since=N` and yields a
-//! stream of decoded events. The daemon may push two shapes:
+//! stream of decoded events. The daemon may push three shapes:
 //!
 //! - `{"kind":"frame", ...AcpBroadcastFrame}`: the next replayed
 //!   or live event.
 //! - `{"kind":"lagged"}`: the in-memory ring buffer evicted events
 //!   the client hadn't acked yet. The consumer must drop its local
 //!   state and rehydrate via [`super::http::HttpClient::replay`].
+//! - `{"kind":"heartbeat"}`: the app-level keepalive the daemon emits
+//!   on every ping tick (`PING_INTERVAL` in `src/server/acp_ws.rs`).
+//!   Carries no state, so the reader loop drops it without waking the
+//!   consumer. Any new `kind` sentinel the daemon grows must be added
+//!   here too: an unrecognised sentinel falls through to the frame
+//!   parse and surfaces as [`WsError::Parse`], which consumers treat
+//!   as a dropped socket. See #2287 and `parse_text`.
 //!
 //! Auth: the bearer token is sent as a `?token=<>` query string on the
 //! WebSocket URL. Most WS clients do not surface custom headers cleanly,
@@ -155,9 +162,20 @@ async fn reader_loop(
             next = stream.next() => {
                 match next {
                     Some(Ok(Message::Text(text))) => {
-                        let msg = parse_text(&text);
-                        if tx.send(msg).await.is_err() {
-                            return; // consumer dropped
+                        match parse_text(&text) {
+                            // Keepalive: no consumer-visible state, so
+                            // don't wake the consumer at all.
+                            Ok(None) => {}
+                            Ok(Some(msg)) => {
+                                if tx.send(Ok(msg)).await.is_err() {
+                                    return; // consumer dropped
+                                }
+                            }
+                            Err(e) => {
+                                if tx.send(Err(e)).await.is_err() {
+                                    return; // consumer dropped
+                                }
+                            }
                         }
                     }
                     Some(Ok(Message::Binary(_))) => {
@@ -186,24 +204,33 @@ async fn reader_loop(
     }
 }
 
-fn parse_text(raw: &str) -> Result<WsMessage, WsError> {
-    // The daemon sends either a `AcpBroadcastFrame` JSON object or
-    // a `{ "kind": "lagged" }` sentinel. We try the sentinel first
-    // (cheap discriminant probe) and fall back to a full frame parse.
+/// Decode one text frame. `Ok(None)` means the frame was a sentinel with
+/// nothing for the consumer to act on (the daemon's keepalive), which is
+/// distinct from `Err` because consumers escalate a parse error to a
+/// socket teardown and reconnect.
+fn parse_text(raw: &str) -> Result<Option<WsMessage>, WsError> {
+    // The daemon sends an `AcpBroadcastFrame` JSON object or one of the
+    // `{ "kind": ... }` sentinels. We try the sentinels first (cheap
+    // discriminant probe) and fall back to a full frame parse.
     #[derive(serde::Deserialize)]
     struct KindProbe<'a> {
         kind: Option<&'a str>,
     }
     if let Ok(probe) = serde_json::from_str::<KindProbe>(raw) {
-        if probe.kind == Some("lagged") {
-            return Ok(WsMessage::Lagged);
+        match probe.kind {
+            Some("lagged") => return Ok(Some(WsMessage::Lagged)),
+            // App-level keepalive (#2287). A real frame always carries
+            // `session_id`/`seq`/`event` and never a `kind`, so this
+            // cannot shadow one.
+            Some("heartbeat") => return Ok(None),
+            _ => {}
         }
     }
     let frame: AcpBroadcastFrame = serde_json::from_str(raw).map_err(|e| {
         warn!(target: "acp.client.ws", error = %e, "ws frame parse failed");
         WsError::Parse(e.to_string())
     })?;
-    Ok(WsMessage::Frame(Arc::new(frame)))
+    Ok(Some(WsMessage::Frame(Arc::new(frame))))
 }
 
 fn ws_url(endpoint: &DaemonEndpoint, session_id: &str, since: u64) -> String {
@@ -294,7 +321,34 @@ mod tests {
     #[test]
     fn parse_text_lagged_sentinel() {
         let m = parse_text(r#"{"kind":"lagged"}"#).unwrap();
-        assert!(matches!(m, WsMessage::Lagged));
+        assert!(matches!(m, Some(WsMessage::Lagged)));
+    }
+
+    /// The daemon emits `{"kind":"heartbeat"}` every `PING_INTERVAL`
+    /// (30s). Before #2287's client-side follow-up this fell through to
+    /// the `AcpBroadcastFrame` parse, failed on the missing `session_id`
+    /// field, and surfaced as `WsError::Parse`, which
+    /// `tui::structured_view` treats as a dropped socket: an error toast
+    /// plus a full reconnect every 30 seconds on any quiet session.
+    /// Asserted against the server's literal wire bytes (kept stable by
+    /// `heartbeat_frame_shape_is_stable` in `src/server/acp_ws.rs`) so a
+    /// drift on either side fails one of the two tests.
+    #[test]
+    fn parse_text_heartbeat_is_ignored_not_an_error() {
+        let m = parse_text(r#"{"kind":"heartbeat"}"#)
+            .expect("heartbeat must not surface as a parse error");
+        assert!(
+            m.is_none(),
+            "heartbeat carries no consumer-visible state and must not wake the consumer"
+        );
+    }
+
+    /// An unrecognised `kind` sentinel must still surface as a parse
+    /// error rather than being silently swallowed: the client cannot
+    /// know whether it carried state it needed.
+    #[test]
+    fn parse_text_unknown_kind_sentinel_is_an_error() {
+        assert!(parse_text(r#"{"kind":"something_new"}"#).is_err());
     }
 
     #[test]
@@ -307,12 +361,12 @@ mod tests {
         .unwrap();
         let m = parse_text(&raw).unwrap();
         match m {
-            WsMessage::Frame(f) => {
+            Some(WsMessage::Frame(f)) => {
                 assert_eq!(f.session_id, "s-1");
                 assert_eq!(f.seq, 7);
                 assert!(matches!(*f.event, Event::ThinkingStarted));
             }
-            WsMessage::Lagged => panic!("expected frame"),
+            other => panic!("expected frame, got {other:?}"),
         }
     }
 
--- a/src/acp/client/ws.rs
+++ b/src/acp/client/ws.rs
@@ -318,37 +318,53 @@ mod tests {
         assert!(ws_url(&e, "s-1", 0).starts_with("wss://"));
     }
 
-    #[test]
-    fn parse_text_lagged_sentinel() {
-        let m = parse_text(r#"{"kind":"lagged"}"#).unwrap();
-        assert!(matches!(m, Some(WsMessage::Lagged)));
+    /// How each `{"kind":...}` sentinel the daemon can send must classify.
+    ///
+    /// The heartbeat row is the #3171 regression. The daemon emits
+    /// `{"kind":"heartbeat"}` every `PING_INTERVAL` (30s); before this it
+    /// fell through to the `AcpBroadcastFrame` parse, failed on the missing
+    /// `session_id` field, and surfaced as `WsError::Parse`, which
+    /// `tui::structured_view` treats as a dropped socket: an error toast plus
+    /// a full reconnect every 30 seconds on any quiet session. Asserted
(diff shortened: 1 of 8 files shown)


jerome-benoit (commented):
Reviewed all four fixes against `eeabc4d3`. They are root-cause fixes, not band-aids, and the two self-corrections (`0ce4a58c`, `ad125036`) are genuine. **One hard blocker before merge: this collides with #3162** (details below). The rest are non-blocking nits.

## Per fix

**1. ws heartbeat** LGTM. `parse_text` -> `Result<Option<WsMessage>>` with `Some("heartbeat") => Ok(None)` is the right layer, and the two-sided pin (`heartbeat_frame_shape_is_stable` <-> `parse_text_heartbeat_is_ignored_not_an_error`) is a good substitute for the 30s e2e. Nit: an unknown future `kind` now returns `Err`, which consumers escalate to a socket teardown, so a new daemon sentinel shipped before clients learn it re-arms the reconnect loop. It is a defensible fail-loud choice for a replay protocol and it is documented in the module header, so fine as-is; just flagging the forward-compat coupling.

**3. trash purge wedge** LGTM. `is_not_a_worktree_error` classified before the permission/submodule fallbacks is the correct ordering, and the recovery mirrors the missing-`.git` branch. The regression test reproduces the exact `is not a working tree` state (orphaned admin entry, `.git` pointer intact) and asserts idempotency. Nit: the test hard-depends on the `git` binary (panics rather than skips if absent), consistent with the rest of this file.

**4. worktree move EACCES** LGTM on the logic and on `ad125036`, which is a real correctness fix, not polish: before it, the two web gates ran the now-side-effecting release *before* `blocks_worktree_edit()`, so a Starting/Creating/Deleting session with a stopped container had its container discarded and then got a 409. All five gates now order the status check first. Nit: the core `NotRunning -> discard` path is Docker-gated and untested, and (unlike the TUI's pure `worktree_rename_block`) the web-side ordering is inlined in the async handlers, so it is not cheaply unit-testable without a refactor. A branch-level test would be worth it given this touches five call sites with a destructive side effect.

**2. phantom Running -> Idle on structured sessions** Root cause is correct and this is a by-design suppression (disk status for structured rows is never reconciled, `decide_passive_transition` returns `patch: None`), not a mask. The two guards are not redundant: the `instance.rs` guard kills the `log_status_transition` + restamp (the log-spam symptom), the `status_poll_loop` guard kills the `status_tx` broadcast (the push-dwell race). `0ce4a58c` correctly narrows the guard to the broadcast so the unread mark is preserved.

Two things to flag on this commit:
- **Residual (low, narrow):** the preserved `Error -> Idle` heal in `_inner` re-logs and restamps every tick if a structured row's *disk* status is ever `Error` (the heal is never persisted, so disk reloads `Error` next tick). The common `AgentStartupError` case is disk=`Idle`/overlay=`Error`, which does not trigger the heal, so this only bites a converted-from-tmux / pre-fix row. Worth gating the heal-log for structured, or letting #3162's approach subsume it.
- **Unread stuck-on:** this commit knowingly leaves the structured unread indicator re-marked within 2s of a read (documented in the daemon comment). Net effect vs `main` is zero (pre-existing), so not a regression, but #3162's approach happens to fix it as a side effect.

## Author-flagged questions
- **Does suppressing the poll-loop diff for structured rows drop a legitimate `StatusChange` consumer?** No. The only `status_tx.subscribe()` consumer is `push.rs`; genuine ACP transitions are broadcast by `apply_status_intent` (an independent path), and the web live status flows through a separate channel. For a structured row the poll-loop diff is overlay-vs-disk and permanently mismatched by design, i.e. always a phantom. Confirmed safe.
- **Is discarding a stopped container acceptable across all five gates?** Yes. `discard()` force-removes the container but leaves the named volumes intact (recreated with the new bind path on next start), so it is non-destructive, and after `ad125036` every gate short-circuits on status so a request about to be rejected never discards. A running container reports held without discarding.

## The blocker: collision with #3162
This overlaps and conflicts with #3162 (the other structured phantom-status fix). Both rewrite the same `status_poll_loop` region and diverge on the `update_status_with_metadata` signature, so they cannot both merge. They also differ in scope: #3162 is a single-choke-point daemon refactor (sets `status = live` for structured, skips the tmux decision) that is testable and additionally fixes the unread-stuck-on and the disk=`Error` heal edge, but touches only the daemon; this PR's `instance.rs` guard additionally covers the non-daemon callers (`aoe ps`, TUI poller). Suggested reconcile: take #3162's daemon refactor and keep this PR's `instance.rs` guard for the non-daemon paths, rather than merging either alone. Whoever owns both should pick one shape.

## Nits summary (non-blocking)
- Add a `status_poll_loop`-level test for the structured broadcast suppression + unread preservation (currently untested).
- Add a branch-level test for the `ensure_sandbox_container_released` `NotRunning`/`Failed` split.
- Optionally gate the `Error -> Idle` heal-log for structured rows (disk=`Error` edge).

Bundling four independent, individually-revertible fixes is fine; only Fix 2 needs coordination with #3162. Fixes 1/3/4 look mergeable on their own.

_Review assisted by AI and cross-validated across independent passes; conclusions were traced to the code (file:line) rather than the PR prose. I am happy to discuss any of this directly._

njbrake:

Thanks, this was a genuinely useful review. Taking the blocker and acting on all three flags.

## Fix 2 is dropped, #3162 should own it

You are right that these cannot both merge, and on comparing them #3162 is the better fix, so I have dropped fix 2 from this PR entirely (both `71b08f9e` and `0ce4a58c`). The branch is force-pushed and now has zero changes to `src/server/mod.rs` and `src/session/instance.rs`.

#3162 wins on the merits, not just on being first:

- One choke point instead of my two guards in two files.
- It fixes the unread stuck-on, which mine knowingly left broken.
- It fixes the disk=`Error` phantom you identified below, which mine leaves.
- It extracts `apply_tick_status_decisions` and `observed_transitions` as pure functions, so the regression is actually testable. Mine was untested at the loop level, as you noted.

One correction to your reconcile suggestion, though. You proposed keeping my `instance.rs` guard for the non-daemon callers. I checked and it does not earn its place:

- `aoe ps` already does `if inst.is_structured() { continue; }` before the call (`src/cli/ps.rs:395`), so it never reaches the guard for a structured row.
- The TUI poller does reach it, but its baseline and status both come from the same in-memory row and it writes both back each tick, so a heal converges in one tick with or without my guard. No phantom loop there.

So my `instance.rs` change was effectively daemon-only, which is precisely where #3162 does it better. Keeping it would have been two mechanisms for one bug. #3162's own reasoning for leaving `instance.rs` alone (the unconditional `Error` heal is correct for callers with no overlay to re-pin the value) is more precise than the framing in my PR description was.

## Your residual on the `Error` heal was right

The heal re-logging every tick when a structured row's disk status is `Error` is real, and my "still observes the heal" test locked that behavior in rather than catching it. Worth noting it is not something my fix introduced: on `main` the same row logs every tick too, via baseline-vs-healed instead of disk-vs-healed. My guard narrowed the phantom from all structured rows to just disk=`Error` ones without closing it. Moot now that fix 2 is dropped, since #3162's approach never runs the heal in the daemon at all.

## CodeRabbit found a real gap in `ad125036`, now fixed

Your point that the five gates deserved a branch-level test led somewhere: CodeRabbit flagged that all five released the container before establishing the directory would move at all. A name that sanitizes to the current leaf, a no-op edit, or a branch-only edit each destroyed a stopped container while leaving the worktree where it was.

The tell was that `discard_sandbox_container_after_move` already gates on `path != current_path`, and `rename_selected` even has a comment explaining why a branch-only rename leaves the mount valid. The pre-move gate never learned the same rule, so the two sat five lines apart disagreeing.

Fixed in `b83d8e11` with `worktree_move_required`, and both it and `edit_worktree_workdir`'s own `path_changes` now route through a shared `target_worktree_path` so the preflight cannot drift from the operation it predicts. That is also the branch-level test you asked for: table-driven over the no-op, sanitizes-to-current and changed cases, plus one asserting the preflight agrees with the target path it gates.

Worth recording that I got a case wrong on the first attempt and the test caught it: I assumed `"Feature Login"` would sanitize to an existing `feature-login` leaf. Neither sanitizer lowercases, so a case-only change is a real relocation. It is now asserted as one, against the actual chain rather than my assumption.

## The remaining nits

- **Branch-level test for the `NotRunning` / `Failed` split:** not added. The split is Docker-gated and I have no runtime here to exercise it for real. I would rather leave it uncovered and say so than add a test that mocks away the only thing worth testing.
- **Test for the `status_poll_loop` broadcast suppression:** moot, that code is gone with fix 2. #3162 covers the equivalent with its extracted functions.
- **The unknown-`kind` forward-compat coupling on fix 1:** agreed, and I kept the fail-loud behavior deliberately. A sentinel the client cannot interpret might carry state it needs, and silently dropping it in a replay protocol seemed worse than a reconnect. It is called out in the module header so the next person adding a sentinel sees that both sides need updating.
- **The `git`-binary dependency in the trash test:** left as-is, consistent with the rest of that file.

## Still unverified, and I would value your read

The premise under fix 4 is that a stopped-but-not-removed Docker Desktop container keeps pinning its bind mount on macOS. I have not reproduced it. My evidence is two log timelines where the move failed about 10 seconds after `docker stop` and then succeeded immediately after `docker rm`. That rules out root-owned files as the cause, since `docker rm` does not change ownership, but it is inference rather than a reproduction, and no CI job can reach that path.

If stopping alone does release the mount on your setup, fix 4 is treating the wrong cause and should come out. Fixes 1 and 3 stand on their own either way.

Current state: seven commits, rebased onto `c857547c`, all gates green locally including the `cargo doc` job that caught two rustdoc errors I had missed.

Issue #3094, opened 2026-07-24
Web dashboard: suspended session resume stuck until Stop, spurious 503 toast, composer re-fills


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Sending a message to a suspended (idle-dormant) structured view session from the web dashboard gets stuck. The session tries to resume, the prompt POST comes back 503, the message drops into the queue, and it never drains on its own. The user has to press **Stop**, which finally brings the worker online and auto-sends the queued message. Reported in #3087.

Three separate things go wrong in that one flow, visible across the reporter's three screenshots:

1. A red **"Server error 503 from /api/sessions/&lt;id&gt;/acp/prompt"** toast fires, even though the app deliberately treats that exact 503 as a normal "worker still resuming, will retry" state.
2. The queued prompt **never drains** while the worker is resuming; it stays parked until the user presses Stop.
3. After the Stop + auto-resend, the message text **reappears in the composer** (a duplicate of what was already sent).

Net effect: every suspended session needs a "Stop, then it sends, now delete the leftover text" dance before it is usable again.

## What's there today

The prompt path already has a designed-in "worker is waking, this is fine" contract, and the frontend honors it:

`web/src/hooks/useAcpSession.ts:1576-1585` classifies a `503` whose body starts with `worker_not_ready` as a transient and suppresses its own error banner:

```ts
const workerNotReady = res.status === 503 && detail.startsWith("worker_not_ready");
if (rejected) {
  dispatch({ kind: "prompt_send_rejected" });
}
if (!workerNotReady) {
  dispatch({ kind: "error", message: `Could not send prompt (${res.status}). ${detail}`.trim() });
}
```

The server emits exactly that body while a resume is in flight (`src/server/api/acp.rs:1226-1233`, `src/server/session_service.rs:446-455`).

### Layer 1: the toast is not from the prompt hook

The global fetch wrapper toasts **every** 5xx on an `/api/*` path, with no exemption for the suppressed `worker_not_ready` case, at `web/src/lib/fetchInterceptor.ts:133-134`:

```ts
if (isApi && res.status >= 500 && !isServerDown()) {
  reportError(`Server error ${res.status} from ${path}`);
}
```

So the prompt hook says "this is fine, queuing" and the interceptor fires a red toast for the same response. That is the "Server error 503" the reporter sees.

[...]

Issue #2897, opened 2026-07-16
Plugin host: ACP-capability discovery, host-owned session create/prompt, plugin storage, structured settings (cron plugin PR1 of 2)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Plugin-first pivot for cron scheduling (supersedes the in-core approach in agent-of-empires/agent-of-empires#2887). This issue is PR1 of 2: the AoE plugin-host API changes that let a cron PLUGIN work. PR2 is the plugin itself in `agent-of-empires/plugin-cron`.

Decided by a two-model architecture debate. Full design notes on the branch; summary below.

## Goal

Generalize the plugin host so an out-of-process plugin worker can discover ACP capabilities, create a host-owned structured-view session, and deliver a prompt to it, under a tight unattended-execution security model, plus structured settings widgets so a plugin can present a rich form. Cron logic itself moves out of core into the plugin.

## Already landed on `feat/plugin-host-session-api` (off main)

- `refactor(server): extract spawn_structured_session` (the create-path core, to become the shared SessionService).
- `feat(plugin-api): add session/acp capabilities and bump API_VERSION to 9` — new capabilities `acp.capabilities.read`, `session.create`, `session.prompt`, `session.unattended`; API_VERSION 8 -> 9.

## Remaining work (dependency order)

1. **SessionService.** Reshape `spawn_structured_session` (`src/server/session_spawn.rs`) into a host service with `create_structured_session(...)` and `send_turn(...)`. Both the HTTP create handler and the new plugin RPC call it. It owns validation, ACP option validation, repository-trust enforcement, persist, ACP worker spawn, mode selection, initial-turn delivery, caller attribution (user/HTTP vs plugin id), and idempotency. No second create path.
2. **Host security model.** Host classifies each approval mode as interactive / guarded / unattended (the plugin supplies a mode id but cannot self-label it safe; unknown modes are treated as unattended). `session.unattended` grant required for an unattended mode. Repository trust stays a runtime fail-closed check at spawn (canonicalize path immediately before), independent of install grants. Turn-send restricted to sessions whose `creator_plugin_id` matches the caller. Host-side rate + concurrency limits, audit (plugin id, schedule/idempotency key, project, agent, mode, session id, denial reason), and a kill switch.
[...]

PR #3175, opened 2026-07-31
fix(acp): release instance_lock before awaiting the worker, and withhold the publish until it is ready

### LINKED ISSUE #3172: bug(acp): prompt to an idle-dormant session is published but never forwarded (instance_lock deadlock between acp_prompt and build_spawn_request)
## Summary

Since v1.13.2, a prompt sent to an **idle-dormant** session is published to the event store but **never forwarded to the agent**. The worker spawns ~10s late, the session shows a queued prompt forever, and nothing retries. Any cron/scheduler that wakes a dormant session with a prompt silently loses that prompt.

Root cause is a lock-ordering deadlock between `acp_prompt` and the background respawn it triggers.

## The deadlock

`acp_prompt` holds the per-session `instance_lock` across the entire `send_turn` call:

```rust
// src/server/api/acp.rs (acp_prompt)
let outcome = {
    let inst_lock = state.instance_lock(&id).await;
    let _serialized = inst_lock.lock().await;
    state.session_service.clear_pending_initial_turn(&id).await;
    state.session_service.send_turn(&SessionCaller::User, &id, &req.text, &attachments, woke_idle_dormant).await
};
```

For a dormant session, `send_turn` calls `trigger_resume_background`, whose detached spawn task calls `build_spawn_request` — which takes **the same lock**:

```rust
// src/server/acp_reconciler.rs:1319 (build_spawn_request)
let inst_lock = service.instance_lock(&target.id).await;
...
let _guard = inst_lock.lock().await;
```

`AppState::instance_lock` and `SessionService::instance_lock` share one map — the doc comment on the latter says so explicitly ("both operate on the shared map, so a lock taken through either handle excludes the other").

Sequence:

1. `UserPromptSent` is published to the event store (so the UI shows the prompt queued).
2. The spawn task blocks on `instance_lock`.
3. `send_prompt` → `wait_for_worker` waits `WORKER_READY_TIMEOUT` (10s), gives up, returns `UnknownSession` → `SendTurnError::WorkerNotReady`.
4. The handler returns and drops the lock; the spawn task unblocks and spawns the worker **exactly 10s late**.
5. Nothing re-forwards the prompt. The session ends up with a live worker and an undelivered prompt.

Two stale comments encode the now-violated invariant:

- `acp.rs`: "`send_turn` does not take `instance_lock` (the drain calls it while holding the lock), so this cannot deadlock." — it does, transitively via `trigger_resume_background`.
- `acp_reconciler.rs:1438`: "The wake path drops `instance_lock` before calling this" — no longer true for the `acp_prompt` path.

Introduced by b8c2bea8 (#3061), which added the `instance_lock` wrap to serialize the pending-turn clear. Still present on `main` @ eeabc4d3.

## Evidence

Wake→spawn latency measured across 565 historical `waking idle-dormant` events in my worker logs:

| Period | wakes | median latency | ≥9s |
|---|---|---|---|
| before 2026-07-29 | 528 | 0.02s | 2 |
| since 2026-07-29 | 37 | 0.04s | **18** |

My daemon restarted onto v1.13.2 at `2026-07-30T04:29:47Z`. The first prompt-wake after that (`04:35:53Z`) took 10.04s, and every subsequent one pinned at 10.0x s — matching `WORKER_READY_TIMEOUT` exactly. Each 10s stall corresponds 1:1 with a stuck session.

Worker log for one occurrence (`e1096fd3a23c4ef6`):

```
2026-07-31T02:00:01.946274Z INFO acp.supervisor: waking idle-dormant structured view session on prompt; spawning a fresh worker
2026-07-31T02:00:11.972483Z INFO acp.mcp: forwarding MCP servers count=1        <-- +10.03s
2026-07-31T02:00:12.299838Z INFO acp.protocol: resuming session via session/load
2026-07-31T02:00:15.357851Z INFO acp.protocol: session/load succeeded
2026-07-31T02:00:15.357984Z INFO acp.protocol: sending session/set_config_option mode=bypassPermissions
<end of log — no "sending prompt" ever>
```

Event store for the same session confirms `UserPromptSent` at 10:00:01 local, then `AcpSessionAssigned` at 10:00:15, then nothing. The session sat idle for 30 minutes until I manually re-sent the prompt.

## Steps to reproduce

1. Set `[acp] auto_stop_idle_secs` to something short.
2. Let a session go idle-dormant (`Stopped { reason: "idle_auto_stop" }`).
3. `POST /api/acp/sessions/{id}/prompt` (or `aoe acp prompt <id> "hi"`).
4. Observe: 10s gap between the `waking idle-dormant` log line and the spawn; no `sending prompt` line; `UserPromptSent` present in the event store with no agent response.

## Impact

Hits every scheduler-driven wake. On this machine three sessions were silently stuck simultaneously, including a cron-kicked PR-review job that had been dead for 30 minutes. Because `UserPromptSent` *is* recorded, the UI looks like the agent is thinking rather than showing an error — so the failure is invisible until someone notices the session never replied.

## Suggested fix

Drop `instance_lock` before `send_turn` (the clear could keep it, the send does not need it), or have `build_spawn_request` accept an already-held guard / not re-acquire. Either way, a regression test asserting wake→spawn latency stays sub-second, or that a dormant-session prompt reaches `send_prompt`, would catch this class.

A retry would also help defence-in-depth: `WorkerNotReady` currently leaves a published-but-undelivered prompt with no recovery path.

## Workaround

`[acp] auto_stop_idle_secs = 0` — sessions never go dormant, so the buggy wake path is never taken.

## Environment

- aoe 1.13.2 (`1.13.2+g806583f8f6ef`), Linux, `aoe serve --auth=passphrase --behind-proxy`

Seluj78:

## Description

Since v1.13.2, a prompt sent to an idle-dormant structured session gets written to the event store and then never reaches the agent. The worker spawns exactly 10 seconds late, the session renders as "running" forever, and the only way out is to stop it and re-send. It reproduces several times per session on a machine that lets sessions go dormant.

The cause is a lock-ordering stall. `acp_prompt` held the per-session `instance_lock` across the whole `send_turn` call. For a dormant session `send_turn` calls `trigger_resume_background`, which detaches a task that takes that same lock inside `build_spawn_request` (and again inside `ensure_container_for_session`). So the spawn could not start until the handler released the lock, while the handler sat in `wait_for_worker` burning the full 10s `WORKER_READY_TIMEOUT` waiting for precisely that spawn. It then returned 503, dropped the lock, and the worker came up 10s late with nobody left to forward the text. Introduced by b8c2bea8 (#3061), which added the lock wrap so a fresh prompt supersedes a queued rate-limit continuation (#3028).

Two changes:

1. **Scope the handler's guard to `clear_pending_initial_turn`.** Mutual exclusion with the pending-turn drain is all the lock was ever for, and the drain holds it across its whole snapshot-to-publish sequence, so the #3028 ordering survives either way the race lands: if the drain wins it delivers first, and if the handler wins the drain then reads `None` and returns.

2. **Withhold the publish until the worker is ready.** `send_turn` used to publish `UserPromptSent` and only then discover, inside `send_prompt`, that no worker had arrived. A durable prompt with no turn behind it is what the frontend reads as authoritative and renders as a stuck spinner. A readiness wait before the publish makes `WorkerNotReady` a pre-publish failure, so a slow cold resume now leaves nothing behind. The wait is unconditional rather than gated on `needs_resume`, because `is_running` is also true for a resume another caller already reserved, so a false `needs_resume` does not imply a live worker.

No frontend change is needed. `useAcpSession.ts` already treats `503 worker_not_ready` as transient: it rolls the optimistic row back by id and re-queues the prompt for the drain to re-fire on the next `AcpSessionAssigned` (#3094, #1748).

One incidental behavior change worth calling out: a prompt that races a resume another caller already reserved used to publish and then answer 404, which the dashboard classifies as a non-retryable rejection and drops from the queue. It now answers a retryable 503 with nothing published, so that prompt is re-queued instead of lost.

Fixing `drain_pending_initial_turn`, which has the same lock-across-`send_turn` shape since #2897, is deliberately out of scope. It is pre-existing, it self-heals because a failed drain leaves `pending_initial_turn` set for the reconciler to retry, and change 2 already stops it publishing phantom prompts. What is left there is lock duration, not correctness. Happy to file that as a follow-up.

Design was debated across two models. Both rejected threading a "caller already holds the lock" flag into the resume path, since the detached task outlives the caller's guard, and both landed on this shape over a larger `prepare_turn` / `commit_turn` split.

Closes #3172

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## How to test

- [ ] Set `auto_stop_idle_secs` to something short in the profile config that actually applies (note that `~/.agent-of-empires/profiles/<profile>/config.toml` overrides the global `~/.agent-of-empires/config.toml`), then let a structured session go idle-dormant and confirm `Stopped { reason: "idle_auto_stop" }` lands in the transcript.
- [ ] Send a prompt to that dormant session from the dashboard. Confirm the agent picks it up and replies, rather than the session sitting on "running" with no output.
- [ ] Check the worker log at `~/.agent-of-empires/acp-workers/<id>.log`: the gap between `waking idle-dormant structured view session` and `resuming session via session/load` should be well under a second, not the ~10s it was before, and a `sending prompt` line must appear.
- [ ] Repeat the dormant-wake prompt a few times in one session. Previously this failed intermittently and often more than once per session.
- [ ] Confirm a normal prompt to a live, already-running session is unaffected and shows no added latency.
- [ ] Confirm no `UserPromptSent` is left behind on a failed wake: if you do manage to hit a `worker_not_ready` 503, the composer should roll the message back and re-queue it rather than leaving a prompt bubble with a spinner that never resolves.

## Test Coverage Analysis

The change is server-side. The client's `503 worker_not_ready` handling already exists and is already covered by `web/src/hooks/useAcpSession.queue.test.ts`.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: a Rust unit test pins this deterministically in 2.4s with no process, sandbox, or agent involved. `wake_prompt_frees_instance_lock_and_publishes_nothing_without_a_worker` in `src/server/api/acp.rs` holds a `ResumeReservation` to make `wait_for_worker` park exactly as it does mid-respawn, then asserts both halves of the bug: that the handler is not sitting on `instance_lock` while it waits, and that a prompt no worker received leaves nothing in the event store. Verified red on the pre-fix tree (`eeabc4d3`) in a throwaway worktree, on both assertions independently: the lock probe fails its 2s budget because the pre-fix handler holds the mutex for the full 10s, and with that assertion neutralised the response is a 404 with `UserPromptSent` already persisted. An e2e would add process and socket nondeterminism without covering anything more.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via Claude Code, `/aoe-investigate` and `/aoe-implement`), with the fix design debated against gemini-3.1-pro-preview and gpt-5.6-sol.

I hit this bug repeatedly on my own machine and pointed Claude at the session that was stuck. It traced the root cause from the event store and worker logs, found the issue was already filed as #3172, and drafted the fix. I made the scoping calls: A plus the readiness gate over the larger `prepare_turn` / `commit_turn` refactor, and deferring the drain. I reviewed each commit and ran the manual test steps.

## Summary

- Prevents deadlocks when prompts wake idle-dormant sessions.
- Releases `instance_lock` before waiting for the worker.
- Waits for worker readiness before publishing or forwarding prompts.
- Adds `Supervisor::wait_until_ready`.
- Adds regression coverage for lock release and unpublished failed prompts.
- Documents the locking requirements for background resume tasks.

## Benefits

Prompts now reach dormant sessions after the worker starts. Failed wake attempts remain retryable, and the event store does not retain prompts that no worker accepted.

### DIFF
--- a/src/acp/supervisor.rs
+++ b/src/acp/supervisor.rs
@@ -2192,6 +2192,17 @@ impl<S: BroadcastSink> Supervisor<S> {
         self.client_for_session(session_id).await
     }
 
+    /// Await worker readiness without resolving a client, so a caller can
+    /// gate a durable side effect on the worker actually being there. Used
+    /// by `send_turn` to hold back the `UserPromptSent` publish until the
+    /// resume it kicked has landed: publishing first and only then
+    /// discovering the worker never arrived is what leaves a session
+    /// rendering "running" forever with no agent behind it (#3172). Costs a
+    /// single worker-map lookup when the worker is already live.
+    pub async fn wait_until_ready(&self, session_id: &str) -> Result<(), SupervisorError> {
+        self.ready_client(session_id).await.map(|_| ())
+    }
+
     /// Send a user prompt (with optional attachments) to a running
     /// structured view worker.
     pub async fn send_prompt(
--- a/src/server/acp_reconciler.rs
+++ b/src/server/acp_reconciler.rs
@@ -1435,9 +1435,9 @@ async fn resume_target_for_session(
 ) -> Option<ResumeTarget> {
     let instances = service.instances.read().await;
     // Filter the same triage states the reconciler skips everywhere else.
-    // The wake path drops `instance_lock` before calling this, so an archive
-    // or snooze can win the race after dormancy was cleared; resolving to
-    // None (then NotFound) keeps us from respawning a session the reconciler
+    // This runs without `instance_lock` held, so an archive or snooze can
+    // win the race after dormancy was cleared; resolving to None (then
+    // NotFound) keeps us from respawning a session the reconciler
     // intentionally leaves sunk. See #1748.
     let inst = instances.iter().find(|i| {
         i.id == id
@@ -1481,6 +1481,11 @@ pub(crate) enum ResumeTrigger {
 /// reconciler tick sees the reservation via `is_running` and skips the
 /// session, so there is no double-spawn. Returns `Err(CapacityFull)` when
 /// the worker cap is reached so the handler can surface 503. See #1748.
+///
+/// Callers MUST NOT hold the session's `instance_lock` while awaiting the
+/// worker this kicks: the detached task takes that same lock inside
+/// `build_spawn_request`, so a caller that holds it stalls the spawn for
+/// its whole `WORKER_READY_TIMEOUT` wait and then gives up. See #3172.
 pub(crate) async fn trigger_resume_background(
     service: &Arc<SessionService>,
     id: &str,
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -1197,28 +1197,36 @@ pub async fn acp_prompt(
     // A fresh user prompt supersedes any queued rate-limit resume
     // continuation, so drop it before sending: otherwise the reconciler could
     // later replay the older interrupted prompt after this newer one (#3028).
-    // The clear + send run under the per-session `instance_lock` because the
-    // pending-turn drain holds that same lock across its whole snapshot ->
-    // reload -> send -> clear; without it a drain mid-await could deliver the
-    // stale continuation *after* this newer prompt. `send_turn` does not take
-    // `instance_lock` (the drain calls it while holding the lock), so this
-    // cannot deadlock. Resume + publish + forward all live in the shared
-    // service so the plugin host delivers turns through the same path (#2897).
-    let outcome = {
+    // The clear alone runs under the per-session `instance_lock`, and the
+    // guard is dropped before `send_turn`. Mutual exclusion with the
+    // pending-turn drain is enough to keep the #3028 ordering: the drain
+    // holds this same lock across its whole snapshot -> reload -> send ->
+    // clear, so whichever side wins the lock, the stale continuation can
+    // never be published after this newer prompt. If the drain wins it
+    // delivers first; if we win, the drain then reads None and returns.
+    //
+    // Holding the guard across `send_turn` is what broke #3172:
+    // `send_turn` -> `trigger_resume_background` detaches a task that calls
+    // `build_spawn_request`, which takes this very lock, so the spawn could
+    // not start until this handler released it, and the handler was busy
+    // burning `WORKER_READY_TIMEOUT` waiting for that spawn. Resume +
+    // publish + forward still live in the shared service so the plugin host
+    // delivers turns through the same path (#2897).
+    {
         let inst_lock = state.instance_lock(&id).await;
         let _serialized = inst_lock.lock().await;
         state.session_service.clear_pending_initial_turn(&id).await;
-        state
-            .session_service
-            .send_turn(
-                &SessionCaller::User,
-                &id,
-                &req.text,
-                &attachments,
-                woke_idle_dormant,
-            )
-            .await
-    };
+    }
+    let outcome = state
+        .session_service
+        .send_turn(
+            &SessionCaller::User,
+            &id,
+            &req.text,
+            &attachments,
+            woke_idle_dormant,
+        )
+        .await;
     // Smart-rename fires from `acp_event_listener` on the first clean
     // `prompt_complete` `Event::Stopped` (turn-end), so the one-shot never
     // races this handler's live worker for the provider API. See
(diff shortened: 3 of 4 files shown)


Seluj78:
@coderabbitai review

PR #3177, opened 2026-07-31
fix(cli): capture agent_session_id after CLI launch (#3169)

### LINKED ISSUE #3169: CLI-launched sessions never capture agent_session_id without a running daemon or TUI (resume silently breaks for capture-deferred agents)
## Summary

For agents whose session id is only observable *after* launch, a session launched purely through the CLI never gets `agent_session_id` filled or persisted, as long as no `aoe serve` daemon and no TUI are running. This silently breaks resume/recovery for that session.

The only unconditionally exempt agent is `claude` (it pre-mints its UUID at launch). `opencode` is exempt **only** when the opt-in `session.opencode_preassign_session_id` is enabled (it is **off by default**, host-only, and disabled under a command-override or profile host-env), so default-config `opencode` is affected too.

## Affected vs unaffected

- **Affected** (capture-deferred, resume-capable): `codex`, `gemini`, `copilot`, `vibe`, `pi`, `hermes`, `kiro`, `qwen`, `kimi`, `omp`, and default-config `opencode`.
- **Not affected**: `claude` (pre-mint), preassigned `opencode`, and the `ResumeStrategy::Unsupported` agents (`cursor`, `droid`, `settl`, `antigravity`) which start no poller and never resume anyway.

## Repro

1. No `aoe serve` daemon running, no TUI open.
2. `aoe add --cmd <affected agent, e.g. codex / gemini / copilot> --launch` (or `aoe session start <id>`).
3. Inspect the row (`aoe session list-trash` is unrelated; check sessions.json / the resume state): `agent_session_id` is empty.
4. Restart the machine / `aoe session restart`: the agent starts a brand-new conversation instead of resuming, because there is no id to resume from. Re-running `aoe session start`/`restart` does not fix it (start no-ops on a live session; restart kills tmux before acquire).

## Root cause

`agent_session_id` is filled at launch by `Instance::acquire_session_id` -> `fresh_launch_session_id` (`src/session/instance.rs`): `claude` pre-mints a UUID, `opencode` optionally preassigns, **every other agent returns `None`** and is "captured post-launch".

That post-launch capture is done by the per-instance `session_id_poller`, started in `finalize_launch` (`src/session/instance.rs`, gated on `supports_session_poller()` = `resume_strategy != Unsupported`). The poller's `on_change` is log-only; it pushes the observed id onto an in-memory channel. Persisting it requires an external DRAIN step, `drain_and_persist_session_ids` (defined in `src/session/sync.rs`), which is invoked **only** by:
- the TUI event loop (`src/tui/home/mod.rs`), and
- the `aoe serve` daemon (`drain_session_id_updates_in_state`, `src/server/mod.rs`).

The CLI never drains it (no `drain_and_persist_session_ids` call anywhere in `src/cli`). `aoe session start` launches, starts the poller, then returns and exits before the poller's first (~2s) tick. `aoe add --launch` attaches to tmux and blocks while attached, so the poller does run, but nothing reads its channel and the observed id is dropped on detach. The synchronous `try_retroactive_capture` in `acquire_session_id` cannot rescue this either: it is gated on the tmux session already existing, but on a fresh launch the session is created *after* acquire, and on `start`/`restart` of a live session the launch either no-ops (`start_with_size_opts` returns early when the session exists) or kills tmux first (`restart`).

## Impact

`is_recovery_candidate` (`src/session/recovery.rs`) requires `should_attempt_resume(agent_session_id, tool)` (defined in `src/session/instance.rs`), which returns false for an empty/invalid id. So resume-after-reboot / orphan-resume and `--resume` continuity silently never fire for these sessions. A user who drives AoE purely from the CLI loses conversation continuity for every affected agent, with no error surfaced. (Unsupported-resume agents are excluded by the same gate, so this is not a regression for them.)

## Proposed direction (for discussion, not prescriptive)

- Have the CLI launch path run a **bounded post-launch capture + drain** before returning (poll for the agent's id for a short, capped window, then `drain_and_persist_session_ids` once), reusing the existing capture/drain logic rather than adding a parallel one.
- Or, at minimum, surface the limitation: warn on a CLI launch of an affected agent when no daemon is reachable, and document that CLI-only resume needs `aoe serve`.

## Scope note

Bug is in the CLI launch path plus the session-id capture/persist wiring (`src/cli`, `src/session`); no TUI/web dashboard flow changes, so no `web/tests/coverage-matrix.json` entry is required. A CLI e2e test under `tests/e2e/` (fake agent that writes a known id post-launch; assert the launched session persists it) would lock the fix.

jerome-benoit:
## Description

Fixes #3169.

A session launched purely through the CLI never persisted `agent_session_id` for **capture-deferred agents** (every resume-capable agent except `claude` and preassigned-`opencode`) as long as no `aoe serve` daemon and no TUI were running. Resume, resume-after-reboot, and `--resume` continuity then silently broke for those sessions, with no error surfaced.

### Root cause (verified in code)

- `agent_session_id` is set at launch by `Instance::acquire_session_id` -> `fresh_launch_session_id` (`src/session/instance.rs`): `claude` pre-mints a UUID, `opencode` optionally preassigns (off by default), **every other agent returns `None`** and is captured post-launch.
- That post-launch capture is done by the per-instance `session_id_poller`, started in `finalize_launch` (`src/session/instance.rs`). The poller only accumulates the observation on an in-memory mpsc channel; its `on_change` is log-only.
- Persisting the observation requires an external drain, `drain_and_persist_session_ids` (`src/session/sync.rs`), which is invoked **only** by the TUI event loop (`src/tui/home/mod.rs`) and the `aoe serve` daemon (`drain_session_id_updates_in_state`, `src/server/mod.rs`).
- **No CLI path drained it.** Four one-shot CLI launch surfaces start (or restart) the poller and exit without draining: `aoe session start`, `aoe session import --launch`, `aoe add --launch`, and `aoe session restart` (single and `--all`). `session start` returns before the poller's first (~2s) tick; `add --launch` blocks on `tmux attach` while nothing reads the channel; `restart` re-execs the agent, starting a fresh poller, then merges without draining.
- Impact: `is_recovery_candidate` (`src/session/recovery.rs`) requires `should_attempt_resume(agent_session_id, tool)`, which is false for an empty id, so resume/recovery/`--resume` never fired for these sessions.

### Evidence (before/after)

Two e2e regressions in `tests/e2e/cli_session_id_capture.rs`, each with no daemon/TUI and a fake capture-deferred `codex`:

- `session start`: **before**, `sessions.json` keeps `agent_session_id: null` (test asserts `None`, fails); **after**, the id is persisted before the CLI exits (passes).
- `session restart` (with `auto_resume_on_restart = false` forcing a fresh relaunch, and a toggling fake codex minting a second sid): **before**, restart leaves `agent_session_id` unset (`left: None`); **after**, the new sid is persisted (`Some(...)`). Verified locally: red before the restart change, green after.

### Fix

Add `capture_launched_session_id_blocking` in `src/session/sync.rs`: a bounded wait that reuses the **same** `drain_and_persist_session_ids` CAS path the TUI/daemon use (no parallel persistence logic), and call it from every CLI launch site before the process exits:

- `aoe session start` and `aoe session import --launch`: wait after `start_with_size`, bounded by `CLI_SESSION_ID_CAPTURE_TIMEOUT` (8s), exiting as soon as the id lands.
- `aoe add --launch`: drain after `tmux attach` returns, with a short bound (`CLI_ATTACHED_SESSION_ID_CAPTURE_TIMEOUT`, 2s) since the poller has already been observing throughout the attached session.
- `aoe session restart` (single and `--all`): drain after the relaunch (single: after the wake wait; `--all`: inside each worker's `spawn_blocking`, parallel and bounded by the `--all` semaphore), 8s bound. `merge_post_restart` is left untouched; the drain persists directly and phase 3 preserves it.

The helper is a no-op (zero added latency) for sessions that already have an id (`claude`, preassigned-`opencode`, and `restart` outcomes `Resumed`) or that run no poller (`ResumeStrategy::Unsupported`: `cursor`/`droid`/`settl`/`antigravity`; sandboxed agents with no container; `restart` outcome `ResumeFailed`, which clears the poller).

### Risks / limitations

- `aoe session start` and `aoe session restart` for a capture-deferred agent now block up to 8s (typically ~2s, or near-zero on restart since the wake wait already overlaps the poller; exits early on capture). Correctness of resume is worth the bounded wait.
 - `aoe session import --launch` applies the 8s bound **per launched session, serially**, so importing N slow-to-emit sessions can add up to N×8s in the foreground. Bounded and best-effort (a session that never emits within the window is still captured by the next TUI/daemon poller). Parallelization was not retained: it would duplicate the `--all` JoinSet/semaphore machinery for a cold path where the worst case only occurs if the agent writes no transcript at all.
 - `aoe session restart --all` runs the same 8s bounded capture inside each worker's `spawn_blocking`, parallelized under the `--parallel` semaphore (default 3), so M capture-deferred sessions that never emit cost at most ⌈M/parallel⌉×8s in the foreground rather than M×8s. Bounded and best-effort like the import path (each is still captured by the next TUI/daemon drain). Not short-circuited on a dead pane: the cost comes from a freshly relaunched pane that is alive but has not yet written a session id, which within a finite bound is indistinguishable from one about to; a liveness check would not help, and coupling the protocol-agnostic drain helper to tmux was rejected on review. Raise `--parallel` to amortize a large fan-out.
 - The single CLI paths (`session start`, `session restart`) run the bounded capture on the runtime thread, consistent with the already-inline `start_with_size`; acceptable for a one-shot command with no concurrent work. `restart --all` runs it inside `spawn_blocking`.
- `aoe add --launch` can still miss capture if the user kills the terminal without detaching (the process dies before the post-attach drain); it then spins the 2s bound to no effect. Unfixable without a daemon and out of scope; the id is still captured on the next TUI/`aoe serve` drain.
- On timeout the helper logs at `warn` and the message no longer implies a pending in-memory drain: the CLI poller thread dies on exit, so recovery depends on a future TUI/daemon re-observing the id via its own poller.
- No change to the TUI/daemon drain path or to `drain_and_persist_session_ids` itself; concurrent daemon/CLI drains stay race-safe via the existing CAS + flock-scoped `foreign_sid_holder` guard (also true for the parallel `restart --all` workers, which each write a distinct instance under the flock).

Also drops an internal `R2:` label from a test doc comment in `src/session/instance.rs` (separate commit; the label referenced nothing in the code).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added `tests/e2e/cli_session_id_capture.rs` with two e2e tests (fake capture-deferred agent, no daemon/TUI): one asserts `aoe session start` persists `agent_session_id`, the other asserts `aoe session restart` persists the new id after a forced fresh relaunch. Plus 4 in-module unit tests in `src/session/sync.rs` locking the wait/drain/persist behaviour of `capture_launched_session_id_blocking` (persist-to-disk, already-captured fast path, no-poller fast path, late-observation wait).

**Coverage scope, stated honestly:** the e2e exercises the `aoe session start` and `aoe session restart` launch sites end-to-end. The `aoe add --launch` and `aoe session import --launch` sites call the **same** unit-tested helper (only the timeout constant / parallelism differs); their post-`tmux attach` / per-item wiring is not exercised by a dedicated e2e, since driving an interactive `tmux attach` in the harness is impractical and prone to flakiness.

Validated locally: `cargo fmt --check`, `cargo clippy --all-targets --features e2e-tests` (no new warnings; pre-existing `useless_vec`/`dead_code` warnings are outside this diff), `cargo test --lib session::sync` (12 pass), `cargo test --features e2e-tests --test e2e cli_session_id_capture` (2 pass; restart test verified red before the restart change), and `cargo check --features serve`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via opencode)

Root-cause investigation, design, and implementation were AI-assisted under my direction and cross-checked by independent review passes; every claim above is code-referenced and the fix is locked by regression e2e tests that fail before the change and pass after. I have reviewed and understand every line.

## Summary

- Added bounded session ID capture for CLI-launched sessions.
- Applied capture to session start, session import, add, and restart flows.
- Added timeout notices for delayed capture during parallel restarts.
- Sanitized control characters in capture-timeout notes.
- Added unit and end-to-end tests for session ID persistence.
- Removed an obsolete test comment label.

## Benefits

CLI-launched sessions now persist session IDs without `aoe serve` or the TUI. This restores resume and recovery behavior while keeping capture waits bounded.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1179,9 +1179,9 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                 let tmux_session = crate::tmux::Session::new(&instance.id, &instance.title)?;
                 tmux_session.attach()?;
 
-                // #3169: the poller ran throughout the attached session but the
-                // CLI never drained it, dropping the observed id on detach. Drain
-                // it now (short bound: it is almost always already queued).
+                // The poller ran throughout the attached session but the CLI
+                // never drained it, dropping the observed id on detach. Drain it
+                // now (short bound: it is almost always already queued).
                 let file_watch = crate::file_watch::FileWatchService::noop();
                 crate::session::sync::capture_launched_session_id_blocking(
                     &mut instance,
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1178,6 +1178,16 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
 
                 let tmux_session = crate::tmux::Session::new(&instance.id, &instance.title)?;
                 tmux_session.attach()?;
+
+                // #3169: the poller ran throughout the attached session but the
+                // CLI never drained it, dropping the observed id on detach. Drain
+                // it now (short bound: it is almost always already queued).
+                let file_watch = crate::file_watch::FileWatchService::noop();
+                crate::session::sync::capture_launched_session_id_blocking(
+                    &mut instance,
+                    &file_watch,
+                    crate::session::sync::CLI_ATTACHED_SESSION_ID_CAPTURE_TIMEOUT,
+                );
             }
             Err(e) => {
                 if let Err(rollback_err) = storage.update(|all_instances, _groups| {
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -1187,6 +1187,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                     &mut instance,
                     &file_watch,
                     crate::session::sync::CLI_ATTACHED_SESSION_ID_CAPTURE_TIMEOUT,
+                    true,
                 );
             }
             Err(e) => {
--- a/src/cli/session.rs
+++ b/src/cli/session.rs
@@ -752,6 +752,18 @@ async fn start_session(profile: &str, args: SessionIdArgs) -> Result<()> {
     // so a slow agent startup does not block peer mutators on the same
     // profile (daemon poller, sibling CLI invocations).
     working.start_with_size(crate::terminal::get_size())?;
+
+    // #3169: the CLI has no long-lived loop to drain the just-started session-id
+    // poller, so a capture-deferred agent would exit with agent_session_id unset
+    // and silently lose resume. Wait briefly for the poller and persist via the
+    // same drain the TUI/daemon use.
+    let file_watch = crate::file_watch::FileWatchService::noop();
+    crate::session::sync::capture_launched_session_id_blocking(
+        &mut working,
+        &file_watch,
+        crate::session::sync::CLI_SESSION_ID_CAPTURE_TIMEOUT,
+    );
+
     let title = working.title.clone();
     let id = working.id.clone();
 
@@ -1023,6 +1035,7 @@ async fn import_sessions(profile: &str, args: ImportArgs) -> Result<()> {
 /// do not abort the rest.
 fn launch_imported(profile: &str, ids: &[String]) -> Result<()> {
     let storage = Storage::open_unwatched(profile)?;
+    let file_watch = crate::file_watch::FileWatchService::noop();
     for id in ids {
         let (instances, _groups) = storage.load_with_groups()?;
         let Some(inst) = instances.iter().find(|i| &i.id == id) else {
@@ -1034,6 +1047,12 @@ fn launch_imported(profile: &str, ids: &[String]) -> Result<()> {
             eprintln!("Warning: failed to start {}: {e}", working.title);
             continue;
         }
+        // #3169: persist the poller-observed id before exit (see start_session).
+        crate::session::sync::capture_launched_session_id_blocking(
+            &mut working,
+            &file_watch,
+            crate::session::sync::CLI_SESSION_ID_CAPTURE_TIMEOUT,
+        );
         let wid = working.id.clone();
         storage.update(|instances, _groups| {
             if let Some(stored) = instances.iter_mut().find(|i| i.id == wid) {
--- a/src/cli/session.rs
+++ b/src/cli/session.rs
@@ -753,7 +753,7 @@ async fn start_session(profile: &str, args: SessionIdArgs) -> Result<()> {
     // profile (daemon poller, sibling CLI invocations).
     working.start_with_size(crate::terminal::get_size())?;
 
-    // #3169: the CLI has no long-lived loop to drain the just-started session-id
+    // The CLI has no long-lived loop to drain the just-started session-id
     // poller, so a capture-deferred agent would exit with agent_session_id unset
     // and silently lose resume. Wait briefly for the poller and persist via the
     // same drain the TUI/daemon use.
@@ -1047,7 +1047,7 @@ fn launch_imported(profile: &str, ids: &[String]) -> Result<()> {
             eprintln!("Warning: failed to start {}: {e}", working.title);
             continue;
         }
-        // #3169: persist the poller-observed id before exit (see start_session).
+        // Persist the poller-observed id before exit (see start_session).
         crate::session::sync::capture_launched_session_id_blocking(
             &mut working,
             &file_watch,
@@ -1183,6 +1183,18 @@ async fn restart_all_sessions(profile: &str, parallel: usize) -> Result<()> {
             let title = inst.title.clone();
             let res = tokio::task::spawn_blocking(move || {
                 let result = inst.restart_with_size(size);
+                // Drain the fresh poller so a fresh-relaunched capture-deferred
+                // agent persists its new agent_session_id. No-op for Resumed /
+                // ResumeFailed. In spawn_blocking: off the runtime, parallel,
+                // bounded by the semaphore.
+                if result.is_ok() {
+                    let file_watch = crate::file_watch::FileWatchService::noop();
+                    crate::session::sync::capture_launched_session_id_blocking(
+                        &mut inst,
+                        &file_watch,
+                        crate::session::sync::CLI_SESSION_ID_CAPTURE_TIMEOUT,
+                    );
+                }
                 (inst, result)
             })
             .await;
@@ -1353,6 +1365,17 @@ async fn restart_session(profile: &str, args: SessionIdArgs) -> Result<()> {
             }
         }
     }
+
+    // Restart starts a fresh session-id poller; a capture-deferred agent that
+    // relaunches fresh mints a new agent_session_id no CLI loop would drain.
+    // Same drain as `session start`; no-op for Resumed (sid kept) and
+    // ResumeFailed (poller cleared). After the wake wait, so it is usually ready.
+    let file_watch = crate::file_watch::FileWatchService::noop();
+    crate::session::sync::capture_launched_session_id_blocking(
+        &mut working,
+        &file_watch,
+        crate::session::sync::CLI_SESSION_ID_CAPTURE_TIMEOUT,
+    );
 
     // touch_last_accessed runs on `stored`, not `working`: its fields are
     // peer-mutable and do not belong in `merge_post_restart`.
--- a/src/cli/session.rs
+++ b/src/cli/session.rs
@@ -748,10 +748,25 @@ async fn start_session(profile: &str, args: SessionIdArgs) -> Result<()> {
     let mut working = inst.clone();
     working.source_profile = profile.to_string();
 
+    // Snapshot the sid for the same reason `restart_session` does: a persisted
+    // `ResumeIntent::Cleared` (from `aoe session set-session-id <id> ""`) makes
+    // `acquire_session_id` drop it on this launch, but the abandoned rollout
+    // lingers and stays newest-by-mtime, so the fresh poller's immediate first
+    // poll can re-observe it and the drain below would silently revert the
+    // user's clear.
+    let prior_sid = working.agent_session_id.clone();
+
     // Phase 2 (unlocked): tmux work happens outside the cross-process flock
     // so a slow agent startup does not block peer mutators on the same
     // profile (daemon poller, sibling CLI invocations).
     working.start_with_size(crate::terminal::get_size())?;
+
+    // Cleared on this launch, so the sid we came in with is abandoned.
+    if working.agent_session_id.is_none() {
+        if let Some(sid) = prior_sid {
+            working.retroactive_capture_excludes.insert(sid);
+        }
+    }
 
     // The CLI has no long-lived loop to drain the just-started session-id
     // poller, so a capture-deferred agent would exit with agent_session_id unset
@@ -1044,9 +1059,17 @@ fn launch_imported(profile: &str, ids: &[String]) -> Result<()> {
         };
         let mut working = inst.clone();
         working.source_profile = profile.to_string();
+        // See `start_session`: a cleared sid whose rollout is still newest on
+        // disk would be re-adopted by the drain below.
+        let prior_sid = working.agent_session_id.clone();
         if let Err(e) = working.start_with_size(crate::terminal::get_size()) {
             eprintln!("Warning: failed to start {}: {e}", working.title);
             continue;
+        }
+        if working.agent_session_id.is_none() {
+            if let Some(sid) = prior_sid {
+                working.retroactive_capture_excludes.insert(sid);
+            }
         }
         // Persist the poller-observed id before exit (see start_session).
         crate::session::sync::capture_launched_session_id_blocking(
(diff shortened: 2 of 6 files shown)


njbrake (changes requested):

Thanks, and this is aimed at the actual root cause #3169 identified rather than a nearby symptom. Re-driving the existing CAS and flock drain instead of adding a second persistence path is the right architectural call. Coverage of the launch surfaces is complete: `git grep "start_with_size\|restart_with_size" -- src/cli src/main.rs` returns exactly five hits and all five now call the helper.

I checked a few claims rather than taking them. Both new e2e tests actually ran on CI rather than skipping, per the `serve-e2e` log, and both passed. I reproduced the codex parsing primitives against the fixture the test writes: the 64-char stem's last 36 chars parse as a UUID, and `payload.cwd` extracts correctly. The fast-path guard is sound: `maybe_start_poller` really does leave `session_id_poller = None` for `ResumeStrategy::Unsupported`, an exhausted budget, and sandboxed copilot and kimi, so the zero-added-latency claim holds. And the phase-3 merges are safe, since `merge_post_start` copies only `status` and `sandbox_info`, and `merge_post_restart` copies `resume_probe_failed_sid` only when the sids match, so the directly persisted sid is never clobbered.

**The blocker is on the restart path: it can persist the pre-restart id and then stop waiting.** On a forced-fresh restart (`auto_resume_on_restart = false`, or the `FreshAfterFailedResume` loop-breaker), `acquire_session_id` takes the `ResumeIntent::Cleared` arm and sets `agent_session_id = None`, so `maybe_start_poller` snapshots `initial_known = None`. The poller thread then does an immediate first poll *before* entering its 2 second loop. `capture_codex_session_id` returns the newest `rollout-*.jsonl` matching the cwd by mtime, and old rollouts persist forever, so if the freshly relaunched agent has not written its own rollout yet (real `codex` takes a few hundred milliseconds to boot, while the immediate poll fires within tens of milliseconds of `tmux new-session` returning) that poll observes the **previous** sid and enqueues it. `try_drain_poller` pops exactly one item per drain call and the helper returns as soon as `agent_session_id.is_some()`, so the stale sid is CAS-written to disk (`expected_prior = None`, disk `None`, so Applied) and the CLI exits.

Nothing filters it: `build_exclusion_set` only excludes sids owned by other live instances, and `retroactive_capture_excludes` is never populated in production code. A TUI or daemon self-heals on its next tick, because it keeps draining and the poller's 2 second tick reports the new sid; the one-shot CLI does not. So `aoe session restart` on a capture-deferred agent can leave `agent_session_id` pointing at the conversation the user just restarted away from, and the next launch resumes it. Before this PR the field was left `None`, so this is a new failure mode that is worse than what it replaces: a silently wrong id instead of no id.

The existing `retroactive_capture_excludes` filter is the clean fix, since it is honored before the write and is runtime-only. Capture the prior sid before `restart_with_size`, and on a `Fresh` or `FreshAfterFailedResume` outcome insert it into the exclusion set so the drain rejects it and keeps waiting for the real one. Same shape in `restart_all_sessions` inside the `spawn_blocking` closure.

The new restart e2e cannot catch this because `install_toggling_fake_codex` writes the second rollout synchronously inside the script before `exec sleep 300`, so the immediate poll almost always already sees the new file. Delaying that write by a second makes the test red without the fix and green with it, which is the assertion currently missing.

**Second, a convention that landed on main after your base.** Both new e2e tests use `#[serial]`. `0db32ba2` (the test-cost and `#[parallel]` guidance) and `3e027792` (the parallel e2e runner) both postdate `ce799398`, so you could not have known, and it is also why the cost is invisible in this PR's CI numbers: it ran against a still-serial suite at 184s, while current main runs at `--test-threads=3` and 74s. On current main a default-key `#[serial]` test is excluded against every `#[parallel]` test in the binary, and these mutate nothing process-global: `TuiTestHarness::set_env` only appends to `extra_env` for spawned commands, `install_path_command` writes into the per-test temp `$HOME`, and the tmux socket derives from `$HOME`. Worth a rebase so the suite runs under the parallel harness before merge.

**One thing to consider rather than change.** `aoe session start` now blocks up to 8 seconds with no feedback and no opt-out. For any agent whose on-disk session record only materializes after the first user turn (gemini scans `~/.gemini/tmp/<sha>/chats/session-*`, copilot reads a `sessions` row with a `cwd`), a *detached* `aoe session start` can never capture the id inside the window, because the poller dies with the CLI process and the user attaches separately afterwards. Those users pay the full 8 seconds plus a warning on every start for no benefit. I read the capture code rather than timing each agent, so treat that as likely rather than verified, but the description's "typically ~2s" is agent-dependent and probably wrong for that family. A one-line stderr notice after a second, a smaller bound, or a `--no-wait` escape hatch would each address it.

**Minor.** The helper returns on the *first* drained observation, so a long attached session under `add --launch` (a conversation switch, say) can leave newer observations queued and keep the oldest; draining while `touched()` is a two-line change, and it also hardens the restart case above. `disable_auto_resume_on_restart` appends a `[session]` table to a config the harness seeds with `[updates]` and `[app_state]`, which is safe today but becomes a duplicate-key TOML error if the harness ever seeds `[session]`. And the `bool` return is read only by tests, which is borderline against the no-dead-code rule.

njbrake (approved):

Approving. The stale-sid fix is correct and the new test is an honest reproduction, which is the part I most wanted to confirm.

The exclusion works for the right reason: `prior_sid` is snapshotted *before* `restart_with_size` clears it, both forced-fresh outcomes are gated, and `restart_all_sessions` has the same treatment inside its `spawn_blocking` closure. I also checked the interaction with the drain-newest change, since those could have fought each other: the exclusion is applied per observation *inside* `drain_and_persist_session_ids`, while the newest-drain is the loop *around* it, and a filtered observation still sets `filtered_ids` so `touched()` stays true and the loop keeps draining. Draining to newest reinforces the exclusion rather than defeating it. Worth noting this PR is the first production writer of `retroactive_capture_excludes`; that filter was reachable only from test code before now.

The test is a real reproduction rather than a passing assertion. Rollout-1 persists from the first launch, the fake now sleeps before writing rollout-2, and blanking `restart_wake_message` removes the 5s pane wait that used to sit between the relaunch and the drain, so at t=0 the only rollout on disk is the abandoned one. Without the fix the drain writes it and returns; with the fix it is filtered and the t=2s poll picks up the real one, with about 6s of margin against the 8s bound. CI confirms both e2e tests ran rather than skipped. One durability note: that coverage now hinges entirely on `restart_wake_message` staying empty, so if anyone reintroduces a pre-capture wait the test keeps passing while covering nothing. The doc comment calls this out, which is the right mitigation.

I pushed one required fix in 735615b8. The same hole was still open on `start_session` and `launch_imported`: `aoe session set-session-id <id> ""` writes `ResumeIntent::Cleared` but leaves `agent_session_id` on disk, so the next `aoe session start` drops it in `acquire_session_id`, the abandoned rollout is still newest-by-mtime, and the new CLI drain re-adopts it. That silently reverts the user's explicit clear, for exactly the capture-deferred agents this PR targets. Same snapshot-and-exclude treatment on both.

I also dropped "(Ctrl-C to skip)" from the wait notice. Ctrl-C does not skip, it kills the CLI, and in the start and restart paths this wait sits between the tmux launch and the phase-3 storage merge. That gap used to be microseconds and is now up to 8 seconds, so telling the user to interrupt into it invites leaving a running pane with the row never merged. The notice now just says the agent is up and we are waiting on its session id.

Three things worth resolving separately, none blocking:

- The control-char strip added in 79de9e9e handles Cc, which covers the terminal-forging case I raised. But it misses the bidi Cf characters that this repo's own `validate_display_label` explicitly handles for Trojan-Source spoofing (CVE-2021-42574), and that validator is only applied on the server and web paths, so `aoe add -t "<title>"` can put a Cf-bearing title on disk. Meanwhile 34 other `println!`/`eprintln!` sites in `cli/session.rs` and `cli/add.rs` interpolate the title raw, including one about twenty lines below the sanitized note. Either the threat is real and this wants a shared `sanitize_display_label()` used at all 35 sites, or it is not and the inline filter is noise. I would extract the existing validator's char set.
- There is still no opt-out for the 8s wait, and no docs change describing it. For gemini and copilot the on-disk record only materializes after the first user turn, so a detached `aoe session start` can never win inside the window; those users pay the full 8s plus a note every time. On `restart --all` at the default parallelism of 3, that is ceil(N/3) times 8s added to wall clock.
- The `tracing::warn!` beside the timeout note logs `id` and `tool` but not the title that d0d409a4 deliberately added to the user-facing line, so the log is now less useful than the stderr message for working out which session timed out.

PR #3180, opened 2026-07-31
fix(web): guard crypto.randomUUID for insecure-context prompt sends

### LINKED ISSUE #3173: Structured view composer Send silently no-ops when served over plain HTTP on a non-loopback host (crypto.randomUUID insecure-context throw)
## Summary

On `1.13.2`, the structured-view web composer's Send action does nothing — no error toast, no network request, no server-side log entry — when `aoe serve` is bound to a non-loopback, non-HTTPS host (e.g. `--host 192.168.1.x` on a LAN). Sending via `aoe acp prompt` (CLI) works fine, confirming the daemon/agent are healthy; only the browser composer's send path is affected.

## Root cause

`useAcpSession.ts` (added in #3098 / f69379b2, part of 1.13.2):

```ts
const optimisticId = \`user-opt-\${crypto.randomUUID()}\`;
```

This runs unconditionally on every prompt send, before the POST fires. `crypto.randomUUID()` is restricted to [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS, or `localhost`/`127.0.0.1`). `aoe serve --host <LAN-IP>` serves plain HTTP on a non-loopback address, which is not a secure context in any standards-compliant browser — so `crypto.randomUUID` is `undefined` there, and calling it throws a synchronous `TypeError`. That aborts `submitComposer` before the fetch is ever issued, with no visible feedback to the user.

Notably, the codebase already has a guarded fallback for exactly this case elsewhere (`StructuredWidgets.tsx`):

```ts
function newItemId(): string {
  const c = globalThis.crypto;
  if (c && typeof c.randomUUID === "function") return c.randomUUID();
  return \`id-\${Math.random().toString(36).slice(2)}-\${Date.now().toString(36)}\`;
}
```

The new call in `useAcpSession.ts` doesn't use it.

## Repro steps

1. `aoe serve --host <LAN-IP>` (any non-loopback IP, plain HTTP — no `--remote`/tunnel).
2. Open the dashboard from another device via `http://<LAN-IP>:8080/?token=...`.
3. Open a structured-view (ACP) session, type a message, hit Send (or Enter).
4. Nothing happens — no toast, no network request in devtools, no server-side log activity. Confirmed via `aoe acp doctor`/`ps`/`history`: worker healthy, WS connects and replays fine, zero prompt events ever recorded.
5. `aoe acp prompt <session> "text"` from the CLI works and completes a full turn normally.
6. Reproduced identically across Chrome incognito, desktop Safari, and a from-source Vite dev build proxied to a throwaway backend — all served over `http://<LAN-IP>`. Confirmed fixed locally by adding the same `typeof crypto.randomUUID === "function"` guard used in `StructuredWidgets.tsx`.

## Suggested fix

Guard the `crypto.randomUUID()` call in `useAcpSession.ts` the same way `StructuredWidgets.tsx` does, e.g.:

```ts
const optimisticId = \`user-opt-\${
  globalThis.crypto && typeof globalThis.crypto.randomUUID === "function"
    ? globalThis.crypto.randomUUID()
    : \`\${Date.now().toString(36)}-\${Math.random().toString(36).slice(2)}\`
}\`;
```

## Environment

- `aoe 1.13.2` (`1.13.2+g806583f8f6ef`), Linux
- `aoe serve --host <LAN-IP>` (plain HTTP, no reverse proxy/tunnel)
- Reproduced on `1.13.2`; `1.13.1` unaffected (this call site didn't exist before #3098)

Seluj78:

## Description

On `1.13.2`, the structured-view web composer's Send action silently no-oped (no error toast, no network request, no server-side log entry) when `aoe serve` is bound to a non-loopback, non-HTTPS host (e.g. `--host 192.168.1.x` on a LAN). Sending via `aoe acp prompt` (CLI) worked fine, confirming the daemon/agent were healthy; only the browser composer's send path was affected.

Root cause: `useAcpSession.ts`'s `sendPrompt` built its optimistic-row id with an unguarded `crypto.randomUUID()` call. `crypto.randomUUID` is restricted to [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS, or `localhost`/`127.0.0.1`). `aoe serve --host <LAN-IP>` serves plain HTTP on a non-loopback address, which is not a secure context in any standards-compliant browser, so `crypto.randomUUID` is `undefined` there, and calling it threw a synchronous `TypeError` that aborted `submitComposer` before the fetch was ever issued.

The codebase already had a guarded fallback for exactly this case elsewhere (`StructuredWidgets.tsx`'s `newItemId()`). This PR applies the same guard to the `sendPrompt` call site: check `typeof globalThis.crypto.randomUUID === "function"` and fall back to a `Date.now()`/`Math.random()` id when it is not available.

Closes #3173

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] `aoe serve --host <LAN-IP>` (plain HTTP, no `--remote`/tunnel), open the dashboard from another device via `http://<LAN-IP>:8080/?token=...`
- [ ] Open a structured-view (ACP) session, type a message, hit Send: the message now sends (POST fires, agent responds), where before it silently did nothing
- [ ] Confirm the existing `localhost`/HTTPS composer send flow is unaffected

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: this is a browser-API compatibility guard inside a hook, not a new/changed dashboard flow from the coverage-matrix mandate list (auth, wizard, settings, sessions, etc.). Added a Vitest regression test (`web/src/hooks/useAcpSession.queue.test.ts`) that stubs `crypto.randomUUID` away and asserts `sendPrompt` still POSTs instead of throwing; confirmed it fails on the pre-fix code and passes post-fix.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. The issue report already included the root cause and suggested fix; I reviewed the guard placement, the regression test, and confirmed the red-to-green transition before committing.

**NOTE:**

## Summary

- Fixed prompt sending over plain HTTP on non-loopback hosts.
- Added UUID v4 generation with `crypto.getRandomValues` and a timestamp/random fallback.
- Added a regression test for insecure-context behavior.

## Benefits

- Prompts now reach the server in LAN HTTP deployments.
- Existing behavior remains unchanged for HTTPS and localhost contexts.

### DIFF
--- a/web/src/hooks/useAcpSession.queue.test.ts
+++ b/web/src/hooks/useAcpSession.queue.test.ts
@@ -397,6 +397,26 @@ describe("useAcpSession drain race (#1144)", () => {
     expect(reportAcpInteraction).not.toHaveBeenCalled();
   });
 
+  // #3173: aoe serve --host <LAN-IP> is plain HTTP on a non-loopback host,
+  // which is not a secure context, so crypto.randomUUID is undefined there.
+  // sendPrompt must not throw building the optimistic id; it should still
+  // POST the prompt.
+  it("still POSTs when crypto.randomUUID is unavailable (insecure context, #3173)", async () => {
+    vi.stubGlobal("crypto", { ...globalThis.crypto, randomUUID: undefined });
+    const { result } = renderHook(() => useAcpSession("sess-insecure-context"));
+    await flushAsync();
+    const ws = sockets[0]!;
+    act(() => {
+      ws.readyState = FakeWebSocket.OPEN;
+      ws.onopen?.({} as Event);
+    });
+    await flushAsync();
+    await act(async () => {
+      await result.current.sendPrompt("sent over plain http");
+    });
+    expect(promptPostCount).toBe(1);
+  });
+
   it("reports a prompt_queued interaction on the retryable-failure requeue path (#1888)", async () => {
     // Idle-dormant wake: the worker was reaped, so sendPrompt POSTs directly
     // to wake it. When that POST fails retryably (worker_not_ready 503), the
--- a/web/src/hooks/useAcpSession.queue.test.ts
+++ b/web/src/hooks/useAcpSession.queue.test.ts
@@ -402,19 +402,31 @@ describe("useAcpSession drain race (#1144)", () => {
   // sendPrompt must not throw building the optimistic id; it should still
   // POST the prompt.
   it("still POSTs when crypto.randomUUID is unavailable (insecure context, #3173)", async () => {
-    vi.stubGlobal("crypto", { ...globalThis.crypto, randomUUID: undefined });
-    const { result } = renderHook(() => useAcpSession("sess-insecure-context"));
-    await flushAsync();
-    const ws = sockets[0]!;
-    act(() => {
-      ws.readyState = FakeWebSocket.OPEN;
-      ws.onopen?.({} as Event);
-    });
-    await flushAsync();
-    await act(async () => {
-      await result.current.sendPrompt("sent over plain http");
-    });
-    expect(promptPostCount).toBe(1);
+    // Shadow only randomUUID on the real crypto instance; a full stub
+    // object (e.g. `{...globalThis.crypto}`) drops inherited members
+    // like getRandomValues, which would mask this test behind an
+    // unrelated getOrCreateDeviceBindingSecret failure path.
+    const originalRandomUUID = globalThis.crypto.randomUUID;
+    Object.defineProperty(globalThis.crypto, "randomUUID", { value: undefined, configurable: true });
+    try {
+      const { result } = renderHook(() => useAcpSession("sess-insecure-context"));
+      await flushAsync();
+      const ws = sockets[0]!;
+      act(() => {
+        ws.readyState = FakeWebSocket.OPEN;
+        ws.onopen?.({} as Event);
+      });
+      await flushAsync();
+      await act(async () => {
+        await result.current.sendPrompt("sent over plain http");
+      });
+      expect(promptPostCount).toBe(1);
+    } finally {
+      Object.defineProperty(globalThis.crypto, "randomUUID", {
+        value: originalRandomUUID,
+        configurable: true,
+      });
+    }
   });
 
   it("reports a prompt_queued interaction on the retryable-failure requeue path (#1888)", async () => {
--- a/web/src/hooks/useAcpSession.ts
+++ b/web/src/hooks/useAcpSession.ts
@@ -767,6 +767,25 @@ export const ACP_MAX_RETRIES_EXPORT = ACP_MAX_RETRIES;
 const ACP_WS_WATCHDOG_INTERVAL_MS = 15000;
 export const ACP_WS_STALE_MS = 75000;
 
+/** A real UUID v4 for the optimistic prompt id (#3173). `crypto.randomUUID`
+ *  is restricted to secure contexts (HTTPS or localhost); `aoe serve
+ *  --host <LAN-IP>` is plain HTTP on a non-loopback host, where it is
+ *  undefined. `crypto.getRandomValues` has no such restriction, so build
+ *  the UUID from that instead of falling back to a non-UUID string. Only
+ *  when crypto itself is entirely unavailable (older webviews, jsdom) does
+ *  this fall back to a Date.now/Math.random id, same as StructuredWidgets.tsx's
+ *  `newItemId()`. */
+function optimisticPromptId(): string {
+  const c = globalThis.crypto;
+  if (c && typeof c.randomUUID === "function") return c.randomUUID();
+  if (c && typeof c.getRandomValues === "function") {
+    return "10000000-1000-4000-8000-100000000000".replace(/[018]/g, (digit) =>
+      (Number(digit) ^ (c.getRandomValues(new Uint8Array(1))[0]! & (15 >> (Number(digit) / 4)))).toString(16),
+    );
+  }
+  return `${Date.now().toString(36)}-${Math.random().toString(36).slice(2)}`;
+}
+
 export function useAcpSession(
   sessionId: string | null,
   /** Live structured view worker lifecycle from `SessionResponse.acp_worker_state`.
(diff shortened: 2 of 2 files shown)


web/src/hooks/useAcpSession.ts
```diff
@@ -1558,7 +1558,14 @@ export function useAcpSession(
...
       // exact row back (by id) because the prompt is re-queued and the
       // drain would otherwise echo a duplicate. See #3094 / #3087.
-      const optimisticId = `user-opt-${crypto.randomUUID()}`;
+      // crypto.randomUUID is restricted to secure contexts (HTTPS or
+      // localhost); `aoe serve --host <LAN-IP>` is plain HTTP on a
+      // non-loopback host, where it is undefined. See #3173.
+      const optimisticId = `user-opt-${
+        globalThis.crypto && typeof globalThis.crypto.randomUUID === "function"
+          ? globalThis.crypto.randomUUID()
+          : `${Date.now().toString(36)}-${Math.random().toString(36).slice(2)}`
```
jerome-benoit: Better to use a browser compatible UUID v4 generator ...

jerome-benoit (changes requested):

web/src/hooks/useAcpSession.queue.test.ts
```diff
@@ -397,6 +397,26 @@ describe("useAcpSession drain race (#1144)", () => {
     expect(reportAcpInteraction).not.toHaveBeenCalled();
   });
 
+  // #3173: aoe serve --host <LAN-IP> is plain HTTP on a non-loopback host,
+  // which is not a secure context, so crypto.randomUUID is undefined there.
+  // sendPrompt must not throw building the optimistic id; it should still
+  // POST the prompt.
+  it("still POSTs when crypto.randomUUID is unavailable (insecure context, #3173)", async () => {
+    vi.stubGlobal("crypto", { ...globalThis.crypto, randomUUID: undefined });
```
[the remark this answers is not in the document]
Seluj78: Good catch, fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3180/commits/d545c3c6323bc6e0d9fa39845331ac0f48a18580 — spreading crypto dropped inherited getRandomValues since it lives on the prototype, not as an own property. Narrowed the stub to shadow only randomUUID via Object.defineProperty, with restore.

web/src/hooks/useAcpSession.ts
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3180/commits/1aa7a4849f1f60372905200c6fed54b27aee3ade — crypto.getRandomValues (unlike randomUUID) isn't secure-context-gated, so it's still available on plain-HTTP LAN hosts. Built a real UUID v4 from it for the fallback branch instead of the ad-hoc Date.now/Math.random string; that string is now only a last resort if crypto is entirely unavailable.

Seluj78:
Please re-review @jerome-benoit :)

jerome-benoit (commented):
Thanks, the getRandomValues fallback resolves my comment for the composer path: it is a correct browser-compatible RFC 4122 v4. Verified by construction (template `10000000-1000-4000-8000-100000000000` with `/[018]/`: the version nibble stays the literal `4`, the variant nibble is `8 ^ (rand & 3)` in `{8,9,a,b}`, the rest are random nibbles) and empirically (20000/20000 generated ids match `/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/`). `crypto.getRandomValues` is not secure-context-gated (unlike `randomUUID`), so it works on the LAN HTTP path, and the throw before the fetch is gone. The regression test is a valid red-to-green (it shadows only `randomUUID`, keeping `getRandomValues`, so it exercises the real fallback branch). LGTM for the stated scope (#3173).

One thing worth raising, since my comment was really about having a browser-compatible v4 generator in general:

**Sibling call site still unguarded: `web/src/hooks/useDiffComments.ts:160`.** `cryptoRandomId()` is a bare `return globalThis.crypto.randomUUID();` with no guard and no try/catch. It is reached from `addComment` (`useDiffComments.ts:77`) via `DiffFileViewer` draft-save, which is a live dashboard action over the exact same plain-HTTP LAN surface. So adding a diff-review comment throws the identical `TypeError` and silently no-ops in an insecure context, i.e. the same bug class as #3173, just a different call site. It is pre-existing and outside this PR's literal scope, but the fix here is localized rather than shared, so this one is left throwing.

Suggestion: extract a single `browserUuid()` helper (this new `optimisticPromptId` + `StructuredWidgets.tsx` `newItemId` are already near-duplicates) and route `useDiffComments.ts:160` through it too. `fetchInterceptor.ts:107` is fine as-is (its `randomUUID()` is inside a try/catch that falls through), so only `useDiffComments` needs it.

Non-blocking nits:
- The test asserts only that `sendPrompt` POSTs, not that the emitted id is a well-formed v4. A `toMatch(/…-4…-[89ab]…/)` assertion would lock in exactly the property this change is about, so a future regression that silently dropped to the `Date.now()/Math.random()` last-resort would be caught.
- Micro: the fallback calls `getRandomValues(new Uint8Array(1))` once per hex char (~31 times per id); a single `Uint8Array(16)` would do. Only on the rare insecure-context path, so cosmetic.

_Review assisted by AI and cross-validated across independent passes; every claim was traced to the code (file:line) or checked empirically. Happy to discuss._

jerome-benoit (approved):
Approving. The fix is correct and complete for its stated scope (#3173, composer Send): a valid RFC 4122 v4 built from `getRandomValues` (which is not secure-context-gated), so it works on the plain-HTTP LAN path, with a red-to-green regression test. My original comment is resolved.

To be explicit about the sibling I flagged above: `useDiffComments.ts:160`'s unguarded `crypto.randomUUID()` is pre-existing and out of this PR's scope, so it is **non-blocking** and should not hold this merge. It is worth a follow-up (ideally a shared `browserUuid()` helper routing both call sites, plus this one). The two nits (assert the v4 shape in the test; per-char `getRandomValues` allocation) are likewise non-blocking.

PR #3184, opened 2026-07-31
feat(server): expose sleep-inhibit runtime state (enabled/held/backend-available)

### LINKED ISSUE #3032: feat(server): expose sleep-inhibit runtime state (enabled/held/backend-available)
## Problem

Whether the `aoe serve` daemon currently holds the OS "prevent user-idle system sleep" assertion, and whether the backend has latched **unavailable** (missing `caffeinate` / `systemd-inhibit`, or no logind), is invisible to every surface. That state lives in the poll loop's local `sleep_inhibitor: Option<Box<dyn SleepInhibit>>` and the process-global unavailable latch in `src/server/mod.rs`. A user who opts in on a host without the backing tool gets a single one-time `warn!` and no way to see why the machine still sleeps; a remote `aoe` client cannot observe the live state at all.

The **write** side of remote control is already covered: `prevent_sleep_when_active` / `prevent_sleep_idle_grace_minutes` are plain-allow `#[setting]` fields, so `PATCH /api/settings` already toggles them remotely (elevation-gated). What is missing is read-side observability.

## Proposal

Expose a read-only runtime status, e.g. `{ prevent_sleep_enabled, currently_held, backend_available }`, on an existing runtime endpoint (`get_status` in `src/server/push.rs`, or `get_about` in `src/server/api/system.rs`). Read-only; no new write/toggle endpoint.

## Scope (1-PR)

Thread the reconciler's held-state and the unavailable latch into one status struct + one endpoint field + a unit test.

## Out of scope

- A write/"controller" API (redundant with `PATCH /api/settings`).
- Plugin-host RPC exposure (separate follow-up).
- Aging out `Waiting` sessions (separate discussion).

## Anchors

- `reconcile_sleep_inhibit` / `update_sleep_inhibit` + the unavailable latch in `src/server/mod.rs`
- `SleepInhibit` trait + `sleep_inhibitor()` in `src/process/mod.rs`
- `get_status` `src/server/push.rs`; `get_about` `src/server/api/system.rs`

Follow-up to #2916 (closes #2733). Surfacing this was also flagged in the PR #2916 review discussion.

jerome-benoit:
## Description

Closes #3032. Follow-up to #2916.

The `aoe serve` daemon's sleep-inhibit runtime state was invisible to every client: whether it currently holds the OS "prevent user-idle sleep" assertion, and whether the backend has latched **unavailable** (missing `caffeinate` / `systemd-inhibit`, WSL2/container with no logind). That state lived only in the poll loop's local `sleep_inhibitor: Option<Box<dyn SleepInhibit>>` and the process-global unavailable latch in `src/process/mod.rs`. The **write** side is already covered (`prevent_sleep_when_active` / `prevent_sleep_idle_grace_minutes` via `PATCH /api/settings`); this adds the missing **read-side observability**.

This threads the reconciler's held-state and the unavailable latch into a read-only status on the existing `GET /api/about` endpoint:

```json
"sleep_inhibit": {
  "prevent_sleep_enabled": true,
  "currently_held": true,
  "backend_available": true
}
```

### Design (chosen after independent multi-agent analysis + cross-validation)

- **Endpoint**: `ServerAbout` / `GET /api/about` (the daemon's typed runtime-facts document), not `push::get_status` (push-namespaced, untyped).
- **Threading the loop-local slot out**: the inhibitor slot stays loop-local (single-owner, per the existing comment). `update_sleep_inhibit` publishes a packed `AtomicU8` snapshot onto `AppState` (`sleep_inhibit_snapshot`): bit0 = `prevent_sleep_when_active` as the reconciler last read it, bit1 = a slot is retained. Packed into one byte so the two correlated bits are read torn-free. Only this derived scalar reaches `AppState`.
- **Backend availability**: read live via a new `pub fn sleep_inhibit_backend_available()` in `src/process/mod.rs` over the existing monotonic latch (no new global, no snapshot copy). A `target_os` split makes it a single symbol callable unconditionally under `serve`; it returns `false` on platforms with only the no-op backend.
- **Core correctness rule**: `currently_held = slot_present && backend_available`. `is_held_alive()` deliberately reports held under the unavailable latch (to suppress futile respawns), so `slot.is_some()` alone would falsely report held; gating on `backend_available` corrects cases (c) latched-unavailable and (d) no-op platform.
- **Publishing `enabled` from the reconciler snapshot** (not re-read live) keeps `(enabled, held)` mutually consistent: after a toggle-off the assertion is genuinely still held for up to one reconcile interval, so the pair reflects the daemon's actual machinery rather than not-yet-applied intent.
- The logic is isolated in a pure `derive_sleep_inhibit_status(...)` fn, unit-tested with a table (no server stand-up).

### Out of scope (per issue)

- A write/"controller" API (redundant with `PATCH /api/settings`).
- Plugin-host RPC exposure (separate follow-up).
- A web dashboard UI surface for the new field.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Validation on macOS: `cargo fmt` clean, `cargo clippy --features serve --all-targets` clean (7 pre-existing `useless vec!` warnings in `src/session/groups.rs`, unrelated), `cargo test --features serve --lib` 5790 passed / 0 failed / 2 ignored (Docker `#[ignore]`), and non-serve `cargo build` green. The new `derive_sleep_inhibit_status_gates_held_on_backend` table test passes.

## Test Coverage Analysis

No `web/` file is touched. The new field is additive backend-only JSON on `/api/about`; no dashboard code reads it, `ServerAbout` in `web/src/lib/api.ts` is a plain interface (extra JSON fields are ignored at runtime), so the coverage-matrix mandate does not apply.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is pure derivation logic (`slot_present && backend_available`) covered far more cheaply by an in-module table unit test than by standing up a live `aoe serve` daemon; the existing `/api/about` e2e in `tests/e2e/serve.rs` still exercises the endpoint and passes.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Opus) via an autonomous agent workflow.

**Any Additional AI Details you'd like to share:** Design was produced by three independent agent analyses that were cross-validated; implementation and local validation (fmt/clippy/test/build) were run and verified before opening this PR.

**NOTE:**

## Summary

- Added read-only sleep-inhibit status to `GET /api/about`.
- Reported whether sleep prevention is enabled, currently held, and available through the backend.
- Added atomic runtime state tracking in `AppState`.
- Updated TypeScript types, test fixtures, unit tests, and documentation.
- Kept existing settings updates unchanged.

## Benefits

- Improves visibility into daemon sleep-inhibit behavior.
- Separates configuration, active holds, and backend availability.
- Supports clearer runtime diagnostics.


jerome-benoit:
### Self-review note

Ran an initial multi-agent self-review (3 independent analyses, cross-validated) against design/implementation elegance, Rust best practices, AGENTS.md rules (no emdash/`--`, no dead code, conventional commits), terminology harmonization, and root-cause correctness.

**Verdict:** no blocker/major on the code. `currently_held = slot_present && backend_available` was verified correct across all reconciler-reachable states (holding, idle-released, latched-unavailable doomed-slot, no-op platform); atomics/`Relaxed`/cfg-gating/`Serialize`-only all sound; no dead code; commits conventional.

**Addressed:**
- Style nit: added a blank line between the new `derive_sleep_inhibit_status` test and the following test fn.
- (Earlier in the branch) fixed a real CI-breaking finding: `AppState.sleep_inhibit_snapshot` doc used intra-doc links to `pub(crate)` consts, tripping the crate-level `deny(rustdoc::private_intra_doc_links)`; switched to plain backticks. `cargo doc -D warnings` now clean.

**Verified then rejected (non-findings):** a review flag that the TS `currently_held` caveat should mention `Starting`/`Creating` (from #3182) is spurious: #3182 touches `src/session/instance.rs::has_recent_activity` + a docs guide, not this PR's `system.rs`/`process/mod.rs`, and the caveat concerns backing-child death, not session statuses.

Gates green locally: `cargo fmt --check`, `cargo clippy --features serve`, `cargo test --features serve --lib`, `cargo doc -D warnings`, and `cd web && npm run format:check && npm run lint && npx tsc -b`. Branch is a few commits behind `main` but `mergeable: MERGEABLE` (clean merge-tree); left as-is to avoid churn. Kept as a draft.

njbrake (approved):

Approving. I checked the load-bearing claim against the source rather than taking the description's word for it.

**What I verified.** The correctness of this PR rests on one line, `currently_held = slot_present && backend_available`. Reading the latch machinery in `src/process/mod.rs`: `sleep_inhibit_child_held_alive` returns `true` while the latch is set (deliberately, to suppress futile respawns), and `CaffeinateInhibitor::acquire` returns `Ok(())` with `child: None` when the helper binary is missing. So `slot.is_some()` on its own genuinely would report held on a host with no working backend, and gating on `backend_available` is a real correction rather than a redundant one. On platforms with only `NoopInhibitor`, `is_held_alive` returns `true` while the new `sleep_inhibit_backend_available` returns `false`, so those hosts report `currently_held: false` with `backend_available: false`, which is the informative pair.

I also like that `backend_available` documents its own optimism in three places instead of pretending to be a probe. That is the right tradeoff for a status field that must not spawn a subprocess per request.

**Two non-blocking follow-ups**, neither worth holding this up for:

1. `sleep_inhibit_snapshot` is initialized to `0`, so `/api/about` reports `prevent_sleep_enabled: false` until the first reconcile lands, even when the toggle is on. The window is one poll tick, so this is cosmetic. The same staleness lasts a full interval if the `spawn_blocking(Config::load_or_warn)` join errors, since that path returns after `last_reconcile` has already been advanced. A sentence on the field would keep a future reader from trusting the snapshot at t=0.

2. `docs/api.md` is the page aimed at external orchestrators, and per #3032 an external client is the only consumer of this field today (the dashboard surface is deferred to #3189). Two lines there would make the feature discoverable to the audience it was built for.

**On CodeRabbit's failed pre-merge check:** the 0% docstring coverage warning is a false positive. Every new item in this diff carries a doc comment; the tool does not appear to count them on these forms.

Verdict: implementation approved, and the scope matches the issue's stated one-PR boundary (read side only, no controller API, no UI). Merging.

Issue #1651, opened 2026-05-30
Bug - aoe add: repo config default_image is ignored, falls back to global default


Davicittod:
Hi! When debugging for another feature I encountered this problem to switch the default docker image. I let you below the report that an llm helped me draft.

---

When a repo specifies a custom sandbox image via `.agent-of-empires/config.toml`, `aoe add --sandbox` ignores it and always uses the runtime's global default. The user must pass `--sandbox-image` explicitly on every add call to get the repo-configured image.

**Root cause**

In `src/cli/add.rs`, the image is resolved as:

```
let image = args.sandbox_image
    .as_ref()
    .map(|s| s.trim().to_string())
    .unwrap_or_else(|| runtime.effective_default_image());
```
`runtime.effective_default_image() `reads the global config only. The repo config is already loaded earlier in the same function for hook trust checking, but sandbox.default_image from it is never consulted when setting the session image.

**To reproduce**

1. Add to `.agent-of-empires/config.toml`:
```
[sandbox]
default_image = "ghcr.io/agent-of-empires/aoe-dev-sandbox:latest"
```
2. Run `aoe add . --sandbox`
3. `docker inspect aoe-sandbox-<id> --format '{{.Config.Image}}'` → still shows aoe-sandbox:latest

**Workaround**

`aoe add . --sandbox --sandbox-image ghcr.io/agent-of-empires/aoe-dev-sandbox:latest`

**Proposed fix**

After loading the repo config in add.rs, use its `sandbox.default_image` as the fallback before `runtime.effective_default_image()`:

Option A — Apply it directly in add.rs when building the image string, checking the already-loaded repo config before falling back to the global default.

Option B — Move the resolution into `resolve_config_with_repo` so the merged config's `sandbox.default_image `is already correct by the time add.rs reads it, making the fix available to any future callers too.

Option B is cleaner since it centralizes the logic and makes repo config behave consistently for all sandbox settings, not just hooks.

Issue #557, opened 2026-03-31
Repo-level `.aoe/config.toml` sandbox settings not applied to containers


Davicittod:
Hello I've been trying to create a sandbox environment to allow the YOLO mode for my project without much success (I have needed to manipulate the container manually to do so finally), without success doing it through the automatic setup. I leave a report below of the issue with some help from my agent.

**Version:** `aoe 1.0.0`
**Platform:** macOS 15, Apple Silicon (arm64), Docker Desktop with VirtioFS

## Summary

Repo-level `.aoe/config.toml` settings for `[sandbox].environment`, `[sandbox].volume_ignores`, and `[hooks]` are silently ignored when creating sandboxed sessions. The `[sandbox]` block is partially read (sandbox is enabled when `enabled_by_default = true`), but sub-settings are not merged into the container configuration.

---

## Reproduction Steps

1. Create `.aoe/config.toml` at the project root:

```toml
[sandbox]
volume_ignores = [".venv"]
environment = [
  "CI=true",
  "MY_VAR=hello",
]

[hooks]
on_launch = "echo 'hook fired' > /tmp/hook_test"
```

2. Create a sandboxed session:

```bash
aoe add --sandbox --trust-hooks /path/to/project
aoe session start <session-name>
```

3. Inspect the running container:

```bash
docker exec aoe-sandbox-<id> bash -c "echo \$MY_VAR"
docker exec aoe-sandbox-<id> bash -c "cat /tmp/hook_test 2>/dev/null || echo 'hook did not run'"
```

---

## Expected Behavior

- `echo $MY_VAR` prints `hello`
- `cat /tmp/hook_test` prints `hook fired`
- `.venv` inside the container is empty (shadowed by anonymous volume)

## Actual Behavior

- `echo $MY_VAR` prints nothing
- `hook did not run` — `on_launch` hook never executed
- `.venv` inside the container reflects the host directory (no shadowing)

---

## Additional Finding: `volume_ignores` does not shadow existing host directories on macOS Docker Desktop (VirtioFS)

Anonymous volume shadowing works correctly for **empty or non-existent** host directories. However, when the host directory already exists and contains a large number of files (e.g. a populated `.venv` or `node_modules`), VirtioFS on macOS Docker Desktop appears to resolve the bind-mount path before Docker's volume layer can shadow it — the host directory remains visible and writable from inside the container.

[...]

PR #3185, opened 2026-07-31
fix(session): refuse repo-defined session launch commands

### LINKED ISSUE #3154: `.agent-of-empires/config.toml` can execute `session.custom_agents` commands during `aoe add --launch` without a trust prompt
## Summary

AoE's repository config path allows a checked-out repository to commit `.agent-of-empires/config.toml` with `session` overrides such as `default_tool`, `custom_agents`, `agent_command_override`, and `agent_extra_args`. When a user creates and launches a normal terminal-mode session for that repository, AoE merges the repository config, resolves the selected tool to the repository-defined command, and passes that command to `tmux new-session` as the pane command.

This is deterministic command execution from repository-controlled configuration, not prompt injection.

The important trust-boundary failure is that AoE has a trust/fingerprint system for repository hooks and project `.mcp.json`, but not for these repository-controlled `session.*` command-bearing fields. A malicious repository can therefore cause local command execution during normal session startup without a hook trust prompt, MCP trust prompt, or model/tool approval.

This is not a claim that merely opening a repository silently executes code. The trigger is narrower and should be described precisely: the user must create and launch an AoE terminal-mode session for that repository, or create it and later start it.

## Affected Product

- Product: Agent of Empires (`aoe`)
- Repository: `https://github.com/agent-of-empires/agent-of-empires`
- Confirmed source revision: `3f0ee180371796cc47e96ecdd4195fc687f73662`
- Confirmed date: July 29, 2026

## Security Boundary

Repository-shared project configuration should not be able to define host command execution at session launch without being covered by the same trust boundary AoE already applies to repository hooks and project-local MCP servers.

`session.custom_agents`, `session.agent_command_override`, and related session-launch fields are executable configuration. If they are allowed from repository config at all, they should be explicitly trusted, fingerprinted, and re-approved on change before they influence the actual spawn path.

## Reproduction

Create a disposable repository with this file:

```toml
# .agent-of-empires/config.toml
[session]
default_tool = "repo-agent"

[session.custom_agents]
repo-agent = "sh -c 'printf repo-session-command-executed > .aoe-marker; sleep 20'"
```

Then run:

```bash
aoe add --launch /path/to/disposable/repo
```

Expected secure behavior:

- AoE should treat repository-defined session-launch commands as untrusted executable configuration.
- AoE should show a trust prompt, or otherwise refuse to apply repository-defined launch commands until the repository is explicitly trusted.

Actual behavior:

- AoE creates the session and launches the tmux pane command derived from repository config.
- `.aoe-marker` appears in the repository working directory.
- No hook trust prompt appears, because the repository is not using `[hooks]`.
- No MCP trust prompt appears, because the repository is not using `.mcp.json`.
- No model interaction or tool approval is involved.

In a headless reproduction, `aoe add --launch` may print `open terminal failed: not a terminal` after the session is already created. That attach failure is not protective; the command has already been handed to tmux.

### Dynamic verification

I verified this with an isolated local build of the pinned revision under a temporary HOME/XDG/tmux namespace:

- Control case: an otherwise identical repository with no `.agent-of-empires/config.toml` launched successfully and did not create `.aoe-marker`.
- Exploit case: the repository containing the config above launched successfully and did create `.aoe-marker` with contents `repo-session-command-executed`.
- No trust prompt appeared in the exploit case because only `session` overrides were used.

## Impact

An attacker who can convince a user to run `aoe add --launch` against a repository, or otherwise start an AoE terminal-mode session for that repository, can execute arbitrary shell commands with the privileges and environment of the AoE process in the repository working directory.

This can affect local files accessible to that user, inherited credentials, and reachable network resources. The execution happens before any model request or model-selected tool call.

### 1. Repository config is allowed to override `session`

`src/session/repo_config.rs` includes `session` in `REPO_OVERRIDABLE_SECTIONS`, so repository config is allowed to supply those values:

```rust
const REPO_OVERRIDABLE_SECTIONS: &[&str] = &[
    "hooks", "session", "sandbox", "worktree", "updates", "tmux", "sound",
];
```

The add/session-creation path resolves repository-aware config through `resolve_config_with_repo_or_warn(...)` before creating the session.

### 2. The repository-controlled `session` fields are command-bearing

`SessionConfig` includes command-bearing fields such as:

- `default_tool`
- `custom_agents`
- `agent_command_override`
- `agent_extra_args`
- `agent_acp_cmd`

The code comment on `custom_agents` describes it as a shell command run in a tmux pane, and `resolve_tool_command()` returns either `agent_command_override` or `custom_agents` for the selected tool.

### 3. `aoe add` copies the resolved repository command into the session instance

When no explicit `--tool` or `--cmd` is provided, `resolve_tool_for_add()` uses `config.session.default_tool`. If that name matches a configured custom agent, it is accepted directly.

Later, `aoe add` applies config-derived command overrides and writes the result into `instance.command`:

- `src/cli/add.rs`, `resolve_tool_for_add(...)`
- `src/cli/add.rs`, the `instance.command = resolved;` path

### 4. The launch path passes that command to tmux

For host terminal sessions, `build_host_command()` uses `self.command` when non-empty. `Instance::start_with_size_opts()` then calls `session.create_with_size(...)`, and `src/tmux/session.rs` appends the command as the final `tmux new-session` argument.

That is the execution sink.

### 5. The trust gate does not cover this surface

AoE's repository trust logic currently covers only:

- repository lifecycle hooks from `.agent-of-empires/config.toml`
- project-local MCP servers from `.mcp.json`

`check_repo_trust(...)` computes and checks hashes only for those two surfaces. The CLI/TUI trust prompt also displays and records only hooks and project MCP trust state.

I did not find a comparable trust/fingerprint check for repository-controlled `session.custom_agents`, `session.agent_command_override`, `session.agent_extra_args`, or `session.default_tool` before those values are merged, persisted onto the instance, and passed to tmux.

## Suggested Remediation

1. Treat repository-defined session-launch fields that can alter executables, arguments, or ACP commands as executable configuration and move them behind the same repository trust boundary as hooks/MCP.
2. Bind approval to a fingerprint of the effective command-bearing repository `session` config, not just to repository path.
3. Re-check that fingerprint at the actual spawn point, not only at session creation.
4. Consider disallowing repository overrides for `custom_agents`, `agent_command_override`, `agent_extra_args`, `agent_acp_cmd`, and `yolo_mode_default` by default.
5. Show the exact command and arguments that will be launched before first use and after any repository config change.

## Expected Behavior

A repository should not be able to cause local command execution at AoE session launch merely by committing `.agent-of-empires/config.toml` with `session.custom_agents` or related launch-command fields. If AoE wants to support repository-shared custom launchers, that capability should be explicitly trusted and re-prompted on change.

Seluj78:

## Description

A checked-out repo's `.agent-of-empires/config.toml` could override the whole `[session]` config section, and that section holds the fields that decide what AoE hands to `tmux new-session`: `custom_agents` (name to shell command), `agent_command_override` (replaces the agent binary), `agent_extra_args` (argv injection), `agent_acp_cmd` (argv exec'd directly), plus `yolo_mode_default` and `smart_rename_agent`. So a repo could commit a config, a user could run `aoe add --launch <repo>`, and repo-controlled shell ran on the host with no prompt: the trust/fingerprint system in `check_repo_trust` only covers repo `[hooks]` and project `.mcp.json`, so neither of those prompts fires for a repo that only sets `session.*`.

This makes `session` default-deny at repo scope instead of adding a third trust surface. Only `default_tool` and `agent_detect_as` carry over from a repo; both merely name a tool the user already configured and are never executed as commands. Listing the safe fields rather than today's dangerous ones means a field added to `SessionConfig` later stays repo-denied until someone decides otherwise, which is the failure mode that produced this issue: `session` was allowlisted when it looked like preferences, then accumulated command-bearing fields. Same rationale that already keeps `host_hooks` out of `REPO_OVERRIDABLE_SECTIONS`.

The filter runs in `merge_repo_config` (the execution path, so a programmatically built `RepoConfig` cannot bypass it), in both profile conversions, and inside `save_repo_config`. It also runs *before* the override type-check: otherwise a repo could give a denied field a bogus type and weaponize the validator into failing the whole load, silently discarding the repo's legitimate overrides. Rejected paths are logged without their values, since a dropped command can carry secrets or terminal escape sequences. Nothing hard-fails; a hostile config would otherwise be a repo-controlled DoS on opening the repo at all.

The TUI settings Repo scope now filters on the same rule. It already hid `global_only` fields there so an edit could not write an override the read path ignores; repo-denied fields (and fields in sections a repo may not override at all, e.g. `acp`) were still offered and stranded on save.

I consulted two external models on the design. Both independently rejected the issue's suggested remediation of fingerprinting the effective command fields as a third trust surface: the effective command depends on global, profile, repo, CLI and the tool registry, one displayed command cannot represent argv injection or the ACP/rename subprocesses, and it needs a TOCTOU recheck at every spawn site plus an approval path for headless daemon launches. Refusal is both smaller and stronger.

**Sandbox, added in review.** @jerome-benoit pointed out that `[sandbox]` was the same vulnerability class, and sharper than the residual I had noted: `enabled_by_default` lets a repo drop a plain `aoe add` (no `--sandbox`) into a container the user never asked for, `default_image` picks whose code runs in it, and `extra_volumes` / `mount_ssh` hand that code the host filesystem and the user's SSH keys. `add.rs` gates the container path on the repo-merged `config.sandbox.enabled_by_default` and takes the image from `config.sandbox.default_image`, so no prompt is involved. Those four fields are now denied at the same boundary.

The two sections want opposite defaults, so the policy is per-section: `session` is default-deny with an allowlist, `sandbox` is allow-with-a-denylist of the four fields, and the other overridable sections carry over whole. The rest of `[sandbox]` (resource limits, env passthrough, `volume_ignores`, `custom_instruction`) only tunes a container the user already chose to run, which is the point of repo-shared sandbox config, so it stays. `container_runtime` is a bounded enum and needs nothing.

**Breaking changes, called out:**

- A repo config may no longer set `session.custom_agents`, `agent_command_override`, `agent_extra_args`, `agent_acp_cmd`, `yolo_mode_default`, `smart_rename_agent`, `smart_rename_model`, or any other `[session]` key beyond the two allowed. Move them to global or profile config. The #2351 smart-rename test asserted repo-local `smart_rename_agent` and `agent_command_override`; it now pins the opposite boundary while still covering that helper's repo-aware resolution.
- A repo config may no longer set `sandbox.enabled_by_default`, `default_image`, `extra_volumes`, or `mount_ssh`. This removes two deliberate features: repo-level `default_image` (#1651) and repo-level `extra_volumes` / `mount_ssh` (#557), both of which had tests asserting they worked; those tests now assert the opposite. Use global or profile config, or `--sandbox` / `--sandbox-image` per session.
- TUI settings Repo scope shows fewer rows: only fields a repo config can actually set.

**Known residual risk (deliberately not fixed here):** a session created before this fix keeps its resolved launch command in `instance.command`, and starting it after upgrade still runs that command. The store has no provenance for that string, so a migration cannot tell a repo-derived command from one a user typed via `aoe add --cmd` or from their own custom agent; regenerating or clearing would break working sessions. Anyone who launched a hostile repo already executed the payload at creation time. Delete and recreate pre-fix sessions you do not trust. Worth a follow-up to persist a typed launch spec with provenance.

**Also out of scope, worth a separate issue:** a default-deny audit of the remaining repo-overridable sections as a set. @jerome-benoit flagged `worktree.init_submodules` and the `[sound]` file-name fields as the same pattern; both are inert today and neither puts attacker-chosen code or arbitrary host paths in play, so they are better handled by that audit than by a spot fix here.

Closes #3154

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Create a scratch repo containing `.agent-of-empires/config.toml` with `[session] default_tool = "repo-agent"` and `[session.custom_agents] repo-agent = "sh -c 'printf pwned > .aoe-marker; sleep 20'"`, run `aoe add --launch <that repo>`, and confirm no `.aoe-marker` file appears and the pane runs your own default agent.
- [ ] With `AGENT_OF_EMPIRES_DEBUG=1`, launch a session for that repo and confirm `debug.log` has one `Ignoring repo config overrides that are not permitted at repo scope` line naming `session.custom_agents` and no line containing the command text itself.
- [ ] Change that repo's config to only `[session] default_tool = "codex"` (or any agent you have installed), create a session, and confirm it still starts that agent: the documented repo-config use case is intact.
- [ ] In the TUI, open Settings, switch scope to Repo, open the Agents tab, and confirm Custom Agents / Agent Command Override / Agent Extra Args / Agent Acp Command are gone while Default Tool is still editable. Switch to Global scope and confirm all of them are back.
- [ ] Edit something in Repo scope, save, and confirm `.agent-of-empires/config.toml` contains only keys a repo may set.
- [ ] Add `[sandbox] enabled_by_default = true`, `default_image = "attacker/img"`, `extra_volumes = ["/:/host:rw"]` to that repo's config, run a plain `aoe add` with no `--sandbox`, and confirm no container starts. Then add `memory_limit = "16g"`, run `aoe add --sandbox`, and confirm the limit still applies but the image is yours, not the repo's.

## Checklist

## Test Coverage Analysis

The dashboard has no Repo settings scope, and the API surface is unchanged; the merge boundary the server calls is covered by the Rust unit tests below.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is a config-merge boundary, proven directly and deterministically in unit tests (an e2e would need a hostile fixture repo and a real agent launch to assert the same thing). The one rendering change, which rows Repo scope offers, is asserted against the built field list in `src/tui/settings/fields.rs`.

Tests added: `test_repo_config_cannot_inject_launch_command` (the issue's exact repro TOML; verified red on the pre-fix tree, `repo-declared custom_agents must be dropped on merge`), `test_session_section_is_default_deny_at_merge` (every denied field, also red pre-fix, plus the exact rejected-path list), `test_denied_session_field_with_wrong_type_does_not_void_repo_config`, `test_save_repo_config_strips_denied_session_fields`, `test_repo_may_override_field`, and `repo_scope_hides_fields_a_repo_may_not_override`.

Validation run after the review changes and a rebase onto current main: `cargo fmt --all -- --check`, `cargo clippy --all-targets` (no new warnings), `cargo test` (4583 passed), `cargo check --features serve`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via Claude Code, `/aoe-implement` skill), with gemini-3.1-pro and gpt-5.6-sol consulted as design reviewers on the refuse-versus-fingerprint question

Investigation, plan and implementation drafted by Claude under my direction. I made the design calls: refuse instead of adding a trust surface, keep the policy as one allowlist in `repo_config.rs` rather than plumbing a new flag through the settings derive macro, and leave pre-fix `instance.command` alone rather than ship a migration that cannot tell attacker commands from mine. I reviewed each commit and run the manual steps above.

## Summary

- Restricted repository `[session]` configuration to `default_tool` and `agent_detect_as`.
- Restricted selected repository `[sandbox]` fields, including `enabled_by_default`, `default_image`, `extra_volumes`, and `mount_ssh`.
- Filtered rejected fields during loading, merging, profile conversion, and saving.
- Logged rejected field names without exposing their values.
- Hid unsupported fields from the TUI Repo scope.
- Updated documentation and tests for security, persistence, allowed fields, sandbox behavior, and settings visibility.

## Benefits

Repository configuration can no longer inject session commands, launch arguments, or selected sandbox settings. Global and profile settings retain control over session behavior, while safe repository settings remain available.

### DIFF
--- a/docs/guides/repo-config.md
+++ b/docs/guides/repo-config.md
@@ -76,9 +76,12 @@ Container hooks get the same variables (forwarded via `docker exec -e`). Status-
 ```toml
 [session]
 default_tool = "opencode"   # Override the default agent for this repo
+agent_detect_as = { my-agent = "claude" }
 ```
 
 Any supported agent name (run `aoe add --help` to see the list).
+
+`default_tool` and `agent_detect_as` are the only `[session]` keys a repo may set. Everything else in that section is ignored from repo config (with a warning naming the keys), because fields like `custom_agents`, `agent_command_override`, `agent_extra_args`, `agent_acp_cmd` and `yolo_mode_default` decide what command AoE launches or how much it is allowed to do: set them in your global or profile config instead.
 
 ### Sandbox
 
--- a/docs/guides/repo-config.md
+++ b/docs/guides/repo-config.md
@@ -89,18 +89,17 @@ Override sandbox settings for this repo:
 
 ```toml
 [sandbox]
-enabled_by_default = true
-default_image = "ghcr.io/agent-of-empires/aoe-dev-sandbox:latest"
 environment = ["NODE_ENV", "DATABASE_URL", "CUSTOM_KEY=value"]
 volume_ignores = ["node_modules", ".next", "target"]
-extra_volumes = ["/data:/data:ro"]
 cpu_limit = "8"
 memory_limit = "16g"
 auto_cleanup = true
 default_terminal_mode = "host"   # "host" or "container"
 ```
 
-List fields (`environment`, `volume_ignores`, `extra_volumes`, `port_mappings`) accept either an array or a single string:
+`enabled_by_default`, `default_image`, `extra_volumes` and `mount_ssh` are ignored from repo config (with a warning naming the keys): together they let a repo put you in a container you did not ask for, running an image it chose, with your filesystem and SSH keys mounted in. Set them in your global or profile config, or pass `--sandbox` / `--sandbox-image` per session.
+
+List fields (`environment`, `volume_ignores`, `port_mappings`) accept either an array or a single string:
 
 ```toml
 [sandbox]
@@ -156,8 +155,6 @@ on_destroy = ["docker-compose down"]
 default_tool = "claude"
 
 [sandbox]
-enabled_by_default = true
-default_image = "ghcr.io/agent-of-empires/aoe-dev-sandbox:latest"
 environment = ["DATABASE_URL", "REDIS_URL", "NODE_ENV=development"]
 volume_ignores = ["node_modules", ".next"]
 
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -7297,8 +7297,24 @@ mod tests {
     #[cfg(feature = "serve")]
     #[tokio::test]
     #[serial_test::serial]
-    async fn force_smart_rename_preflight_honors_repo_local_override() {
+    async fn force_smart_rename_preflight_sees_command_override_but_not_from_a_repo() {
         use axum::body::to_bytes;
+
+        async fn preflight_message(repo: &std::path::Path) -> String {
+            let mut inst = Instance::new("Vikings", repo.to_str().unwrap());
+            inst.tool = "claude".to_string();
+            inst.source_profile = "default".to_string();
+            inst.view = crate::session::View::Structured;
+            let id = inst.id.clone();
+
+            let state = crate::server::test_support::build_test_app_state(vec![inst]);
+            let resp = force_smart_rename(axum::extract::State(state), axum::extract::Path(id))
+                .await
+                .into_response();
+            assert_eq!(resp.status(), StatusCode::CONFLICT);
+            let body = to_bytes(resp.into_body(), 1024).await.unwrap();
+            String::from_utf8_lossy(&body).to_string()
+        }
 
         let tmp_home = tempfile::tempdir().expect("tempdir HOME");
         let repo = tempfile::tempdir().expect("tempdir repo");
@@ -7307,8 +7323,9 @@ mod tests {
             std::env::set_var("HOME", tmp_home.path());
             std::env::set_var("XDG_CONFIG_HOME", tmp_home.path().join(".config"));
         }
-        // Repo-local override of the session's own agent binary: the profile
-        // config is otherwise eligible, so only the repo-aware resolver sees it.
+
+        // A repo declaring the override changes nothing: command-bearing
+        // session fields are not repo-overridable (#3154).
         let cfg_dir = repo.path().join(".agent-of-empires");
         std::fs::create_dir_all(&cfg_dir).unwrap();
         std::fs::write(
@@ -7316,24 +7333,25 @@ mod tests {
             "[session.agent_command_override]\nclaude = \"wrapper-3058\"\n",
         )
         .unwrap();
+        let msg = preflight_message(repo.path()).await;
+        assert!(
+            !msg.contains("command is overridden"),
+            "a repo must not be able to declare the agent command override; got: {msg}"
+        );
 
-        let mut inst = Instance::new("Vikings", repo.path().to_str().unwrap());
-        inst.tool = "claude".to_string();
-        inst.source_profile = "default".to_string();
-        inst.view = crate::session::View::Structured;
-        let id = inst.id.clone();
-
-        let state = crate::server::test_support::build_test_app_state(vec![inst]);
-        let resp = force_smart_rename(axum::extract::State(state), axum::extract::Path(id))
-            .await
-            .into_response();
-
-        assert_eq!(resp.status(), StatusCode::CONFLICT);
-        let body = to_bytes(resp.into_body(), 1024).await.unwrap();
-        let msg = String::from_utf8_lossy(&body);
+        // The user's own override is still seen through the repo-aware
+        // resolution the preflight routes through (#3058).
+        let app_dir = isolated_app_dir(tmp_home.path());
+        std::fs::create_dir_all(&app_dir).unwrap();
+        std::fs::write(
+            app_dir.join("config.toml"),
+            "[session.agent_command_override]\nclaude = \"wrapper-3058\"\n",
+        )
+        .unwrap();
+        let msg = preflight_message(repo.path()).await;
         assert!(
             msg.contains("command is overridden"),
-            "preflight must see the repo-local override via repo-aware resolution; got: {msg}"
+            "preflight must see the user's override via repo-aware resolution; got: {msg}"
         );
     }
 
@@ -8495,10 +8513,20 @@ mod tests {
 
     #[test]
     #[serial_test::serial]
-    fn session_tool_identity_uses_repo_aware_config_for_request_path() {
+    fn session_tool_identity_uses_repo_aware_config_but_not_repo_custom_agents() {
         let temp_home = tempfile::tempdir().unwrap();
         std::env::set_var("HOME", temp_home.path());
-        let _app_dir = isolated_app_dir(temp_home.path());
+        let app_dir = isolated_app_dir(temp_home.path());
+        std::fs::create_dir_all(&app_dir).unwrap();
+        std::fs::write(
+            app_dir.join("config.toml"),
+            r#"
+                [session.custom_agents]
+                my-agent = "ssh -t lenovo claude"
+            "#,
+        )
+        .unwrap();
+
         let project = tempfile::tempdir().unwrap();
         let repo_config_dir = project.path().join(".agent-of-empires");
         std::fs::create_dir_all(&repo_config_dir).unwrap();
@@ -8511,7 +8539,14 @@ mod tests {
         )
         .unwrap();
 
+        // The user's own custom agent resolves through the repo-aware path.
         assert!(validate_session_tool_identity(
+            "my-agent",
+            "default",
+            project.path()
+        ));
+        // A repo-defined one does not exist as far as AoE is concerned (#3154).
+        assert!(!validate_session_tool_identity(
             "repo-agent",
             "default",
             project.path()
--- a/src/session/container_config.rs
+++ b/src/session/container_config.rs
@@ -3288,6 +3288,7 @@ mod tests {
 environment = ["MY_VAR=hello", "CI=true"]
 volume_ignores = [".venv", "node_modules"]
 extra_volumes = ["/host/data:/container/data:ro"]
+mount_ssh = true
 "#,
         )
         .unwrap();
@@ -3346,15 +3347,16 @@ extra_volumes = ["/host/data:/container/data:ro"]
             config.anonymous_volumes
         );
 
-        // Verify extra_volumes from repo config are present
+        // A repo cannot mount host paths into the container or hand it the
+        // user's SSH keys (#3154); the tuning fields above still apply.
         let volume_pairs: Vec<(&str, &str)> = config
             .volumes
             .iter()
             .map(|v| (v.host_path.as_str(), v.container_path.as_str()))
             .collect();
         assert!(
-            volume_pairs.contains(&("/host/data", "/container/data")),
-            "extra_volumes should include /host/data:/container/data, got: {:?}",
+            !volume_pairs.contains(&("/host/data", "/container/data")),
+            "repo-declared extra_volumes must not mount, got: {:?}",
             volume_pairs
         );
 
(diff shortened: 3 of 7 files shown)


jerome-benoit (commented):
## Review — repo-config sanitization (#3185)

**Verdict: the #3154 launch-command vector is fully closed; but "Closes #3154" over-claims — an equivalent no-prompt host-compromise vector remains open via `[sandbox]`.**

### Confirmed correct (proof)
- Enforcement lives in the single merge choke point: `sanitize_repo_overrides` → `allowed_overrides` → `merge_repo_config`, reached by both `resolve_config_with_repo` and `_or_warn`. Not UI-gated. `repo_config.rs:302-345`.
- Default-deny: for `session` scope only `default_tool` + `agent_detect_as` survive (`REPO_ALLOWED_SESSION_FIELDS`, `repo_config.rs:81`); both are non-executable. Launch-command / alias / wrong-typed / direct-`.overrides` bypasses all rejected.
- `smart_rename.rs` / `sessions.rs` diffs are test-only.

### Blocking-adjacent: sandbox residual is the same vulnerability class as #3154 (HIGH)
`sanitize_repo_overrides` keeps any non-`session` section **verbatim** (`repo_config.rs:313-314`), and `sandbox` is in `REPO_OVERRIDABLE_SECTIONS` (`:60-61`; test `:1820` confirms `sandbox.enabled_by_default=true` survives). A malicious repo's `.agent-of-empires/config.toml` can therefore set:

```toml
[sandbox]
enabled_by_default = true
default_image = "attacker/img"
extra_volumes = ["/:/host:rw"]   # or mount_ssh = true
```

On a plain `aoe add` (no `--sandbox`), `add.rs:855` gates on the repo-merged `config.sandbox.enabled_by_default`, and `:873-876` launches the repo-chosen `default_image` — attacker code runs with the host FS mounted, **no prompt**. `container_runtime` is a bounded enum (Docker/Podman/AppleContainer), so no arbitrary-binary vector there — but image + volumes + mount_ssh compose to host compromise. This is runtime-gated (needs Docker/Podman installed), so not *worse* than #3154, but it is the **same class** the PR claims to close.

**Ask:** either (a) default-deny the impactful sandbox fields (`enabled_by_default`, `default_image`, `extra_volumes`, `mount_ssh`) the way `session.*` is, or trust-gate them; and (b) soften "Closes #3154" to note the deferred sandbox residual so it isn't read as fully closing the repo-config-trust class.

Nits (LOW): `worktree.init_submodules`, `[sound]` file-name fields are repo-settable; inert today but same pattern.

Seluj78:
> either (a) default-deny the impactful sandbox fields (`enabled_by_default`, `default_image`, `extra_volumes`, `mount_ssh`) the way `session.*` is, or trust-gate them

Went with (a), in 82b02cf2. You were right that this is the same class, and sharper than the out-of-scope note I had written: I had framed the sandbox residual as mount escalation *inside* a sandbox the user opted into, and missed that `enabled_by_default` lets the repo force the opt-in itself. Confirmed your trace before changing anything: `add.rs` resolves the repo-merged config, gates the container path on `config.sandbox.enabled_by_default`, and takes the image from `config.sandbox.default_image`, so a plain `aoe add` was enough.

The policy is now per-section rather than one global rule, since the two sections want opposite defaults:

- `session`: default-deny, allowlist of `default_tool` + `agent_detect_as`.
- `sandbox`: allow-with-a-denylist of the four fields you named. The rest (resource limits, env passthrough, `volume_ignores`, `custom_instruction`) only tunes a container the user already chose to run, and is the reason repo-shared sandbox config exists, so it stays.
- Everything else in `REPO_OVERRIDABLE_SECTIONS`: unchanged.

`repo_may_override_field` is the single lookup behind the sanitizer and the TUI's Repo scope, so the two cannot drift.

This is a real breaking change and the PR body now says so: repo-level `default_image` (#1651) and repo-level `extra_volumes` / `mount_ssh` (#557) were deliberate features with tests asserting them; those tests now assert the opposite. Users who relied on them move the settings to global/profile config or pass `--sandbox` / `--sandbox-image`.

On `container_runtime`: agreed, bounded enum, no arbitrary-binary vector, left alone.

On your LOW nits (`worktree.init_submodules`, `[sound]` file-name fields): left as-is deliberately. Both are inert today and neither puts attacker-chosen code or arbitrary host paths in play the way the sandbox four do, so folding them in would widen an already-widened security PR without closing a live vector. Worth a follow-up that audits the remaining repo-overridable sections as a set, rather than a spot fix here.

"Closes #3154" stays, since #3154 is specifically the `session.custom_agents` launch-command vector, and the sandbox class is now closed in the same PR rather than deferred.

Seluj78:
@coderabbitai all five nitpicks applied, in 82b02cf2 and 515e5489.

- **smart_rename fixture not observable** — correct, and the best of the five. The fixture wrote only denied fields, so the test would have passed even if the repo config had never been discovered. It now also sets `default_tool = "codex"` and asserts that lands before the boundary assertions run.
- **Cover the second allowed session key** — done, the Repo-scope test now loops over both `session.default_tool` and `session.agent_detect_as`.
- **Log dropped fields on save** — done, `save_repo_config` warns with the same paths-only shape as the load path.
- **Non-object `session`** — done, and extended to a non-object `sandbox` (`sandbox = 7`) since that section is now field-filtered too.
- **Make `RepoConfig.overrides` private** — done, with one correction: the field was not unreferenced outside the module. `tests/integration/repo_config.rs` read it directly, which only showed up once the compiler complained. That test now goes through serde instead.

PR #3188, opened 2026-07-31
fix(config): write through symlinks in every atomic_write, not just the settings save

### LINKED ISSUE #3186: Globally symlinked config overwritten during updates :-(
As a follow-up to https://github.com/agent-of-empires/agent-of-empires/issues/2784, it seems that the globally symlinked `config.toml` also/now gets overwritten from an update. 

I don't really have a way to reproduce or confirm this other than, after 2784 was released, I symlinked it, updated AoE about 1-2 days ago and now it's a regular file again. 

Please let me know if there are any logs or additional steps I can take to help troubleshoot this.

Seluj78:

## Description

A symlinked `~/.agent-of-empires/config.toml` survived a settings save after #2785, but not an update. Every migration that rewrites the file went through the plain `atomic_write`, whose temp-file plus `rename(2)` dance replaces the symlink with a regular file. So on the first launch after an update carrying a config-touching migration, a dotfiles user's symlink silently became a detached regular file and their repo copy stopped receiving changes.

`v021_split_app_state_to_state_toml` (2026-07-15, fires whenever `[app_state]` is present in `config.toml`, so effectively anyone who used the TUI) and `v022_prune_tuning_settings` (2026-07-17) both landed after the #2785 fix merged on 2026-07-14, which is why this came back as a fresh regression rather than old history.

The direct cause is 16 migration call sites, but the reason it will keep happening is that #2785 fixed one call site instead of the invariant, and two runtime writers were still wrong for the same reason (`save_profile_config`, `save_repo_config`). Rather than swap 20 more call sites and add a guard to police the next one, this PR resolves the symlink chain inside `atomic_write` itself and deletes the opt-in `atomic_write_following_symlinks`.

That is the shape `locked_update` already had: it resolves the chain unconditionally for `sessions.json`, `groups.json`, `state.toml`, and `mcp_state.json`, and its docstring already carried the dotfiles rationale. The codebase was giving two answers to one question, and the wrong answer had the shorter, more discoverable name. Nothing in AoE wants to replace a user's symlink with a regular file, so one behavior for every writer removes the footgun instead of policing it. `src/plugin/host_api.rs` is untouched; it refuses symlinks via its own `symlink_metadata` checks as a separate security boundary.

Non-symlinked installs are byte-identical: `resolve_symlink_chain` returns the path unchanged when the path is not a symlink. Perms do not regress either, because `fs::metadata` already followed symlinks, so `atomic_write` was already reading the target's perms; now the write lands there too.

Closes #3186

### Follow-on from review: the login-session write path

Making `atomic_write` follow symlinks reversed an explicit invariant on one file, caught by @jerome-benoit in review. `check_path_security` refuses a symlinked `login_sessions.toml` and ran at load (`login.rs:795`) and at the startup rewrite, but `persist()` to `write_sessions` to `atomic_write` skipped it. That was only incidentally safe while `rename(2)` destroyed the planted link.

The check now lives inside `write_sessions`, so every writer inherits it instead of one call site remembering to ask, and the outer guard at the startup rewrite is gone (it became a redundant second stat and a second warn path). Not privilege escalation: app-dir perms gate it and the store is re-asserted 0600. The reason to fix it is that `login_sessions.toml` is the one file with an explicit no-symlink invariant, so it should not depend on a write primitive's incidental behavior.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Move `~/.agent-of-empires/config.toml` into a separate dir and symlink it back: `mkdir -p ~/dotfiles && mv ~/.agent-of-empires/config.toml ~/dotfiles/aoe-config.toml && ln -s ~/dotfiles/aoe-config.toml ~/.agent-of-empires/config.toml`.
- [ ] Force a migration to run: `echo 20 > ~/.agent-of-empires/.schema_version` (this replays v021 onward on next launch). Back up `config.toml` first if you care about the contents.
- [ ] Launch `aoe`, quit, then confirm `ls -l ~/.agent-of-empires/config.toml` still shows a symlink, not a regular file.
- [ ] Confirm the migrated content landed in `~/dotfiles/aoe-config.toml` (its `[app_state]` table should be gone, and `~/.agent-of-empires/state.toml` should exist).
- [ ] Symlink a profile config the same way (`~/.agent-of-empires/profiles/<name>/config.toml`), change a profile-scoped setting in the settings TUI, and confirm the link survives with the new value at the target.
- [ ] On a non-symlinked install, change a setting and confirm nothing about saving behaves differently.
- [ ] Symlink `~/.agent-of-empires/login_sessions.toml` at something harmless, run `aoe serve`, log in, and confirm the target is never written and the log carries `refusing to write persisted login sessions to an insecure path`.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is a filesystem write primitive, so a unit test can fail on it deterministically while an e2e test would only exercise it incidentally. Three unit tests cover it instead, each verified red on the pre-fix tree:

- `session::storage::tests::atomic_write_follows_symlinks`: the link survives, the target receives the bytes, and a dangling link materialises its target rather than a regular file at the link path. Fails pre-fix with "the symlink must survive".
- `migrations::v021_split_app_state_to_state_toml::tests::symlinked_config_survives_migration`: end-to-end on the migration that most likely clobbered the reporter's config. Fails pre-fix with "the migration must not replace the symlink with a regular file". This is the one that proves the reported bug dead.
- `server::login::tests::write_sessions_refuses_symlinked_path`: the persist path refuses a symlinked store and never writes the target. Verified red by disabling the new guard, where the write returns `true` and the target receives the session store.

The red direction for the first two was proven in a detached worktree at the pre-fix commit, so the fix line is reverted but the committed test text is unchanged.

One thing the issue asked for that deliberately did not become its own test fn, per the repo's test-cost policy: a source-scanning guard test so a future migration cannot reintroduce the bug. Not needed under this shape, because the unsafe variant no longer exists, so there is no wrong helper left to reach for. A profile-config-level test is likewise skipped: `save_profile_config` writes through `atomic_write` with no logic of its own, so it would assert the primitive a second time; the manual step above covers it.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via Claude Code, agent-of-empires `/aoe-investigate` and `/aoe-implement` skills)

Claude did the investigation and drafted the implementation and prose. I made the design call: collapse the primitive rather than swap 20 call sites and add a guard test, on the grounds that `locked_update` already established that behavior. I reviewed each commit and ran the manual test steps.

**NOTE:**

## Summary

- Updated `atomic_write` to follow symlink chains and write to their targets.
- Removed `atomic_write_following_symlinks`.
- Applied the safer write behavior to configuration, migration, and hook files.
- Added regression tests for symlinked configuration and storage paths.
- Moved session-store security checks into `write_sessions` so all writes reject symlinks.
- Updated the live-frame-compression test to poll for compressed frames, reducing CI timing flakes.

## Benefits

These changes preserve symlinked configuration files and protect the session store from unintended writes. They also improve test reliability.

## Validation

- 6,193 tests passed.
- Formatting and Clippy checks passed.

## Potential follow-up

Additional coverage could verify symlink behavior across all supported configuration formats.

### DIFF
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -691,7 +691,7 @@ pub fn install_hooks(
             std::fs::create_dir_all(parent)?;
         }
         let formatted = serde_json::to_string_pretty(&settings)?;
-        crate::session::atomic_write_following_symlinks(settings_path, formatted.as_bytes())?;
+        crate::session::atomic_write(settings_path, formatted.as_bytes())?;
 
         tracing::info!(target: "hooks.install", "Installed AoE hooks in {}", settings_path.display());
         Ok(())
@@ -844,7 +844,7 @@ fn with_config_lock<T>(
 }
 
 fn write_codex_config(config_path: &Path, config: &toml_edit::DocumentMut) -> Result<()> {
-    crate::session::atomic_write_following_symlinks(config_path, config.to_string().as_bytes())
+    crate::session::atomic_write(config_path, config.to_string().as_bytes())
 }
 
 fn read_codex_config(config_path: &Path) -> Result<toml_edit::DocumentMut> {
@@ -1154,7 +1154,7 @@ pub fn disable_gemini_folder_trust(settings_path: &Path) -> Result<()> {
             std::fs::create_dir_all(parent)?;
         }
         let formatted = serde_json::to_string_pretty(&settings)?;
-        crate::session::atomic_write_following_symlinks(settings_path, formatted.as_bytes())?;
+        crate::session::atomic_write(settings_path, formatted.as_bytes())?;
         tracing::info!(target: "hooks.install",
             "Disabled Gemini folder trust in {}", settings_path.display());
         Ok(())
@@ -1277,7 +1277,7 @@ pub fn uninstall_hooks(settings_path: &Path) -> Result<bool> {
         }
 
         let formatted = serde_json::to_string_pretty(&settings)?;
-        crate::session::atomic_write_following_symlinks(settings_path, formatted.as_bytes())?;
+        crate::session::atomic_write(settings_path, formatted.as_bytes())?;
 
         tracing::info!(target: "hooks.uninstall", "Removed AoE hooks from {}", settings_path.display());
         Ok(true)
@@ -1366,7 +1366,7 @@ pub fn install_settl_hooks_with_events(
             std::fs::create_dir_all(parent)?;
         }
         let formatted = toml::to_string_pretty(&config)?;
-        crate::session::atomic_write_following_symlinks(config_path, formatted.as_bytes())?;
+        crate::session::atomic_write(config_path, formatted.as_bytes())?;
 
         tracing::info!(target: "hooks.install", "Installed AoE hooks in {}", config_path.display());
         Ok(())
@@ -1404,7 +1404,7 @@ pub fn uninstall_settl_hooks(config_path: &Path) -> Result<bool> {
         }
 
         let formatted = toml::to_string_pretty(&config)?;
-        crate::session::atomic_write_following_symlinks(config_path, formatted.as_bytes())?;
+        crate::session::atomic_write(config_path, formatted.as_bytes())?;
         tracing::info!(target: "hooks.uninstall", "Removed AoE hooks from {}", config_path.display());
         Ok(true)
     })
@@ -1479,10 +1479,7 @@ pub fn install_kimi_hooks_with_events(
         if let Some(parent) = config_path.parent() {
             std::fs::create_dir_all(parent)?;
         }
-        crate::session::atomic_write_following_symlinks(
-            config_path,
-            config.to_string().as_bytes(),
-        )?;
+        crate::session::atomic_write(config_path, config.to_string().as_bytes())?;
 
         tracing::info!(target: "hooks.install", "Installed AoE hooks in {}", config_path.display());
         Ok(())
@@ -1579,10 +1576,7 @@ pub fn uninstall_kimi_hooks(config_path: &Path) -> Result<bool> {
             config.as_table_mut().remove("hooks");
         }
 
-        crate::session::atomic_write_following_symlinks(
-            config_path,
-            config.to_string().as_bytes(),
-        )?;
+        crate::session::atomic_write(config_path, config.to_string().as_bytes())?;
         tracing::info!(target: "hooks.uninstall", "Removed AoE hooks from {}", config_path.display());
         Ok(true)
     })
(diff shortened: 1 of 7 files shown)


jerome-benoit (commented):
## Review — atomic_write follows symlinks (#3188)

**Verdict: core change is correct and data-loss-safe. One defense-in-depth asymmetry worth closing before merge.**

### Confirmed correct (proof)
- `atomic_write` resolves the link first (`storage.rs:105`), derives the temp dir from the **resolved target**, `persist` renames onto the target — symlink inode untouched, same-FS (no EXDEV). Durability: `sync_data` before persist + best-effort dir `sync_all` after. Perms preserved via `fs::metadata` (followed links pre-fix too → no widening). Dangling link materializes target; loop capped at 32 hops.
- Pre-fix at `f207dba2` had no resolve → renamed over the link (real regression). Tests are genuinely red-before, hermetic.
- Dedup complete: `git grep atomic_write_following_symlinks` → zero refs. v021 idempotent (existence-guard + `remove` early-return). Web change inert.

### Should fix: `login_sessions.toml` no-symlink policy is now only half-enforced (LOW-MEDIUM)
`check_path_security` explicitly **refuses symlinks** (`login.rs:751-752`, `:765-766`) and runs at load (`:795`) and startup (`:222`). The runtime persist path — `persist()` `:578` → `write_sessions` `:716` → `atomic_write` `:724` — does **not** call it. Pre-fix, `atomic_write` clobbered a planted symlink (safe); post-fix it follows it, writing the login-session secret through to the symlink target. Exploitability is gated by app-dir perms (the startup check itself refuses a group/world-writable dir) and the re-asserted `0600` (`:734`), so this is not privilege escalation — but it silently reverses the one file with an explicit no-symlink invariant.

**Ask:** call `check_path_security(path)` inside the persist path for parity with load/startup (root-level restore of the invariant, not a mask).

Seluj78:
> @jerome-benoit: the runtime persist path does not call `check_path_security`. **Ask:** call it inside the persist path for parity with load/startup.

Agreed, and confirmed before fixing: `check_path_security` ran at load (`login.rs:795`) and at the startup rewrite (`:222`), but `persist()` -> `write_sessions` -> `atomic_write` skipped it. Pre-fix that was only incidentally safe, because `rename(2)` destroyed the planted link; resolving the chain reverses exactly that.

Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3188/commits/6a2ed006891f58adb2f28e52fd0d2b4df1dfb68d

The check now lives inside `write_sessions`, so every writer inherits it rather than one call site remembering to ask. That also made the outer guard at `:222` a redundant second stat and second warn path, so the startup rewrite just calls `write_sessions` now. Three enforcement points collapse to one.

Regression test `write_sessions_refuses_symlinked_path` (`#[cfg(unix)]`) asserts the call returns `false` and the planted target is never written. Verified red by disabling the guard: without it the write returns `true` and the target receives the store.

Kept your severity read rather than inflating it. The commit message says plainly that this is not privilege escalation, app-dir perms gate it and the store is re-asserted 0600; the reason to fix it is that `login_sessions.toml` is the one file with an explicit no-symlink invariant, so it should not depend on a write primitive's incidental behavior.

Also rebased onto main (19 commits) and revalidated: 6193 passed, fmt and clippy clean.

PR #3194, opened 2026-08-01
fix(tui): make Ctrl+P open the worktree branch picker

### LINKED ISSUE #3166: Bug: on new worktree branch cannot pick branch
when I enter the new session modal, add path, title, tool then go on to configure worktree, I expect to be able to see branches in the repo and pick the one I want the agent to work on. 
In the worktree configuration modal I see that 'Ctrl+P' should let me choose a branch however pressing that does not cause any UI change

shixi-li:
## Description

Fixes #3166. The worktree configuration overlay advertises `Ctrl+P  branches`, but pressing it produced no UI change at all. Three separate things combined to swallow the keypress:

1. **The path was never expanded.** The handler passed the raw path field straight to `list_branches`, so a path the user spelled as `~/code/project` failed to open the repo. Every other consumer of that same field expands it first (`expand_tilde` on submit, on create-dir, and on each extra-repo entry) - the branch picker was the one that didn't.
2. **Failures were silent.** `if let Ok(branches) = ...` with an inner `if !branches.is_empty()` returned `Continue` in every failure path, so a non-repo path, a git error, and a branchless repo all looked identical to a keypress that did nothing.
3. **The overlay never rendered `error_message`.** Even where an error was set, the worktree overlay only ever drew its hint row, so nothing could have surfaced.

The reporter's flow (type a path, then open worktree config) hits all three at once.

## Changes

- Route both branch-taking fields (Name and Base) through one `open_branch_picker` that expands the path before listing, matching the rest of the dialog.
- Report failures inline instead of returning silently: a git/listing error, a repo with no branches, and an empty path each get their own message.
- Render a pending `error_message` in the overlay's hint row, and clear it when the overlay is left so nothing leaks into the main dialog.

The two Ctrl+P arms were byte-identical apart from which `Input` they wrote on selection (that dispatch already lives in the `branch_picker` block at the top of the handler), so they collapse into one arm.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (no user-facing docs describe this overlay; the in-app hint was already correct, the behavior was not)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: driving this needs a real git repo plus a tilde-addressable `$HOME` inside the tmux harness, and the behavior is fully determined by dialog state (`branch_picker.is_active()` / `error_message`) that unit tests can assert directly. Two table-driven unit tests in `src/tui/dialogs/new_session/tests.rs` cover it at ~15ms instead of ~2s, per the tiering guidance in AGENTS.md.

Added:
- `branch_picker_opens_for_reachable_repo_paths` - name field, base-branch field, and a `~/...` path. Uses `isolate_home` + `#[serial_test::serial]` so the tilde case resolves against a temp home rather than the developer's real one.
- `branch_picker_surfaces_failures_and_clears_them` - non-repo path and empty path each produce an inline error rather than nothing, and leaving the overlay clears it.

Verification, including a negative control: reverting just the `expand_tilde` call turns the tilde case red with `Cannot list branches in ~/repo: ... No such file or directory`, which is the reported symptom.

- `cargo test -p agent-of-empires --lib new_session` - 121 passed
- `cargo clippy --all-targets` - no new warnings
- `cargo fmt --all -- --check` - clean

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Fable 5

I had the AI trace the keypress rather than guess at it, which is how the tilde gap turned up: the fix started as "the handler is fine, maybe the picker renders wrong", and comparing the branch picker against the dialog's other path consumers is what showed it was the only one skipping expansion. The first version of the tilde test silently skipped on macOS (temp dirs live outside `$HOME`), and the second flaked under the parallel suite against a sibling test that swaps `HOME`; the `isolate_home` + `serial` pattern already in this file is what fixed it. I'll answer review questions myself.

## Summary

- Fixed `Ctrl+P` branch selection in worktree configuration.
- Enabled branch selection from the Name, New Branch, and Base fields.
- Routed keyboard and mouse selections to the focused field.
- Expanded `~` in repository paths before listing branches.
- Added inline errors for missing paths, empty repositories, and listing failures.
- Rendered wrapped errors and cleared them when the overlay closes.
- Added tests for path handling, error states, error clearing, rendering, and mouse selection.

## Benefits

Users can select a worktree branch reliably. The overlay now provides clear feedback when branch loading fails and keeps selections in the correct field.

### DIFF
--- a/src/tui/dialogs/new_session/mod.rs
+++ b/src/tui/dialogs/new_session/mod.rs
@@ -1442,6 +1442,7 @@ impl NewSessionDialog {
             if self.focused_field == worktree_field {
                 self.worktree_config_mode = true;
                 self.worktree_config_focused_field = 0;
+                self.error_message = None;
                 return DialogResult::Continue;
             }
             if self.focused_field == sandbox_field && self.sandbox_enabled {
@@ -1731,6 +1732,32 @@ impl NewSessionDialog {
         }
     }
 
+    /// Build and activate the branch picker (Ctrl+P on the worktree name and
+    /// base-branch fields). The path field holds whatever the user typed, so
+    /// it needs the same tilde expansion the submit and dir-picker paths do;
+    /// without it `~/repo` never opens and the picker silently did nothing.
+    /// Failures surface inline instead of being swallowed. See #3166.
+    fn open_branch_picker(&mut self) {
+        let path_str = self.path.value().trim().to_string();
+        if path_str.is_empty() {
+            self.error_message = Some("Set the project path before picking a branch.".into());
+            return;
+        }
+        let resolved = path_input::expand_tilde(&path_str);
+        match crate::git::diff::list_branches(std::path::Path::new(&resolved)) {
+            Ok(branches) if !branches.is_empty() => {
+                self.error_message = None;
+                self.branch_picker.activate(branches);
+            }
+            Ok(_) => {
+                self.error_message = Some(format!("No branches found in {}.", path_str));
+            }
+            Err(e) => {
+                self.error_message = Some(format!("Cannot list branches in {}: {}", path_str, e));
+            }
+        }
+    }
+
     /// Build and activate the registered-projects picker (Ctrl+R on the
     /// extra-repos field). Filters out the primary repo and any paths already
     /// in the workspace_repos list to avoid the builder's duplicate-name guard.
@@ -1790,39 +1817,22 @@ impl NewSessionDialog {
         match key.code {
             KeyCode::Esc => {
                 self.worktree_config_mode = false;
+                self.error_message = None;
                 DialogResult::Continue
             }
             KeyCode::Char('?') => {
                 self.show_help = true;
                 DialogResult::Continue
             }
-            // Ctrl+P on name field opens branch picker
-            KeyCode::Char('p')
-                if key.modifiers.contains(KeyModifiers::CONTROL)
-                    && self.worktree_config_focused_field == WT_NAME =>
-            {
-                let path = std::path::Path::new(self.path.value().trim());
-                if let Ok(branches) = crate::git::diff::list_branches(path) {
-                    if !branches.is_empty() {
-                        self.branch_picker.activate(branches);
-                    }
-                }
-                DialogResult::Continue
-            }
-            // Ctrl+P on base-branch field opens the branch picker too.
-            // Selection routes through `branch_picker` like WT_NAME, so we
+            // Ctrl+P opens the branch picker for the name and base-branch
+            // fields. Selection routes through `branch_picker` for both, so we
             // disambiguate after selection by checking the focused field.
             // See `branch_picker` handling at the top of this function.
             KeyCode::Char('p')
                 if key.modifiers.contains(KeyModifiers::CONTROL)
-                    && self.worktree_config_focused_field == WT_BASE_BRANCH =>
+                    && matches!(self.worktree_config_focused_field, WT_NAME | WT_BASE_BRANCH) =>
             {
-                let path = std::path::Path::new(self.path.value().trim());
-                if let Ok(branches) = crate::git::diff::list_branches(path) {
-                    if !branches.is_empty() {
-                        self.branch_picker.activate(branches);
-                    }
-                }
+                self.open_branch_picker();
                 DialogResult::Continue
             }
             // Ctrl+R on extra_repos field opens the registered-projects picker.
@@ -1841,6 +1851,7 @@ impl NewSessionDialog {
             }
             KeyCode::Enter => {
                 self.worktree_config_mode = false;
+                self.error_message = None;
                 DialogResult::Continue
             }
             KeyCode::Tab | KeyCode::Down => {
--- a/src/tui/dialogs/new_session/mod.rs
+++ b/src/tui/dialogs/new_session/mod.rs
@@ -32,6 +32,11 @@ pub(super) struct FieldHelp {
 
 pub(super) const HELP_DIALOG_WIDTH: u16 = 85;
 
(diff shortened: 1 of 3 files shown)


shixi-li:
Both findings were real; fixed in the follow-up commit.

**Mouse selections routed to the wrong field.** Confirmed: `handle_click` stored every picker result in `worktree_branch`, so opening the picker from Base and clicking a row overwrote Name and left Base empty. Rather than duplicating the routing condition in a third place, both paths (and the main-dialog arm) now go through one `apply_branch_selection`, with the Base field index hoisted to a module constant so the two handlers cannot drift again.

Added `branch_picker_mouse_selection_routes_to_the_focused_field`, which renders into a `TestBackend` (the picker only learns its clickable area while drawing), finds the branch row in the rendered buffer, and clicks it. Negative control: reverting just the `handle_click` line turns it red with `left: "" right: "master"`.

**Clipped error text.** Also confirmed: `Constraint::Min(1)` gave the hint row one line and `Wrap` cannot grow a chunk, so a long git error lost everything after the first line. The overlay now computes the wrapped height the same way the main dialog does at `render.rs:107`, clamped to 6.

`cargo test -p agent-of-empires --lib new_session` 122 passed, `cargo clippy --all-targets` no errors, `cargo fmt --all -- --check` clean.

shixi-li:
Triaged both CodeRabbit findings against the current head `537a877e`:

**1. Mouse picker selections from Base (Major) — not reproducible on this head.** Every branch-picker selection path routes through `apply_branch_selection`, which keys on the focused field: the mouse path at `src/tui/dialogs/new_session/mod.rs:1064` (`handle_click`), main-mode keyboard at `mod.rs:1329`, and worktree-config keyboard at `mod.rs:1813`. The regression test the review asks for also already exists: `branch_picker_mouse_selection_routes_to_the_focused_field` (`tests.rs:1780`) drives a real mouse click while Base is focused and asserts the value lands in `base_branch`, not `worktree_branch`.

**2. Wrapped error rows (Minor) — the overlay already reserves them.** `hint_height` is computed from the pending error text and clamped to 1..=6 rows (`render.rs:819-824`), feeds `Constraint::Min(hint_height)`, and `dialog_height` sums it in — the same required-height approach the main dialog uses.

No code change needed for either; happy to adjust if I've misread a case.

njbrake (approved):

Approving. Thanks for this, it fixes three real defects rather than one, and the tests are the good kind: I reverted each fix in turn locally and confirmed the matching test fails, so they are load-bearing rather than decorative.

What I verified rather than took on faith:

* Reverting the `apply_branch_selection` call in `handle_click` back to `self.worktree_branch = Input::new(value)` makes the mouse test fail with `left: "" right: "master"`, so the CodeRabbit routing catch was genuine and `537a877` really closes it.
* Replacing `expand_tilde(&path_str)` with the raw string fails only the `~/repo` table row; the absolute-path rows still pass. That is how I established the tilde was the actual gap and that absolute paths already worked on main.
* Rendering a 200 character git error into a `TestBackend` wraps to four rows inside the dialog with the tail intact, so the `hint_height` sizing is right and the pre-fix `Constraint::Min(1)` would have clipped it.
* `cargo fmt`, `cargo clippy --lib --tests --all-features`, and `cargo test --lib new_session` are all clean. The seven remaining clippy warnings are all in files this PR does not touch.

I pushed one follow-up commit to your branch with the review nits, since they were all mechanical:

1. **The reported symptom was still reachable one Tab away.** The hint row advertises `Ctrl+P branches` for Name, New Branch, and Base, but the guard only covered Name and Base, so on the New Branch checkbox row the key fell through to `_ => Continue` and did nothing. That is #3166 verbatim. The guard now covers the checkbox row and the selection lands in Name. Negative control: dropping `WT_NEW_BRANCH` back out of the guard fails the new table row with `field 1 should open the picker, got None`.
2. **No coverage for the third fix.** Both of your state tests would still have passed with the error rendering reverted to the hint only row, which is the invisible error bug. I folded two render assertions into `branch_picker_surfaces_failures_and_clears_them` (the error text is drawn, and the hints give way to it). Verified they fail under that revert.
3. `apply_branch_selection` now also checks `worktree_config_mode`, so the focused field only reads as "Base" while the overlay is actually open. No behavior change today, since `activate` has a single call site inside the overlay, but it stops being load-bearing on that.
4. Housekeeping: unstacked the doc comment that sat above `apply_branch_selection` while describing `open_branch_picker`, dropped the redundant in-function alias for `WT_BASE_BRANCH_FIELD` and made the const private now that only this module reads it, and used it in the tests instead of a bare `2`.
5. Corrected a claim in that doc comment: it said the picker needed "the same tilde expansion the submit and dir-picker paths do", but the dir picker does not expand. `DirPicker::activate` gets the raw field value and does a bare `PathBuf::from`, so `Ctrl+P` browse on the Path field with `~/code` lands on a literal `~` directory. Separate pre-existing bug, worth its own issue, not something to fold in here.

Two things I deliberately left alone, both predating this PR:

* `docs/guides/worktrees.md:71` says the picker covers "local and remote-tracking branches", but `list_branches` is `BranchType::Local` only. `list_branches_with_remotes` exists for this and is what the web wizard uses. Either the call switches or the sentence does, but that is a product call, not a nit.
* The "path is a subdirectory of a repo" case surfaces "could not find repository", because `open_repo_at` uses `NO_SEARCH` on purpose. It reads like a broken repo rather than "point at the repo root". Worth better copy at some point.

One request before this closes #3166 automatically: absolute paths already worked before this PR, so the reporter's instance is explained only by a `~` path, a non-repo path, or focus on the checkbox row. Item 1 covers the last of those, so all three now give feedback, but it would be good to hear back from them to confirm.

I have approved the workflow runs on the new head.

Issue #3204, opened 2026-08-01
Nix build drops embedded ACP adapter manifests from cleanCargoSource


jerome-benoit:
## Summary

`nix build .#default` fails because `src/acp/adapters.rs` embeds six JSON files under `acp-worker/adapters/`, but the Nix source filter removes those files.

## Reproduction

```sh
nix build .#default
```

Rust reports that each adapter's `package.json` and `package-lock.json` cannot be read by `include_bytes!`. A regular `cargo check --lib` succeeds from the checkout.

## Root cause

PR #3101 added `src/acp/adapters.rs` and the six tracked manifests together, but left this source definition unchanged:

```nix
src = craneLib.cleanCargoSource ./.;
```

Crane's Cargo filter retains Cargo, Rust, and TOML sources, but excludes these JSON files. In the evaluated Nix source, `src/acp/adapters.rs` is present while all three `acp-worker/adapters/<adapter>/` directories are empty.

Extending the source filter to include `acp-worker/adapters` restores all six files with hashes matching the checkout.

## Expected

Both `default` and `aoe-with-web` Nix packages should include the embedded adapter manifests and build successfully. CI should compile at least one Rust Nix package so source-filter regressions are caught; the current Nix jobs only evaluate derivations and build the standalone web frontend.

## AI usage

I reproduced and verified the findings. AI assisted with the investigation and drafting this report.

njbrake:
_Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Fixed by #3205, merged as `9ce82e06`. Closing.

Your root-cause analysis was right on both counts: crane's filter keeps `.rs`, `.toml` and `Cargo.lock` and nothing else, so the six `acp-worker/adapters/*/package{,-lock}.json` manifests were absent from the evaluated source while `src/acp/adapters.rs` was present. `commonArgs.src` is now a `lib.fileset` union that keeps them.

Two things landed beyond the reported bug, both from the same class you identified:

- `docker/Dockerfile` was missing too. The `dockerfile_pin_matches_floor` test embeds it with `include_str!`, and while the packages build regardless (`doCheck = false`), `checks.aoe-test` and `checks.aoe-clippy` compile test code, so neither could build from the flake. Added to the union.
[...]

PR #3205, opened 2026-08-01
fix: include acp-worker adapter manifests in Nix build source

jerome-benoit:
## Description

`nix build .#default` (and `.#aoe-with-web`) fail to build the Rust crate:
`src/acp/adapters.rs` embeds six npm manifests with
`include_bytes!("../../acp-worker/adapters/<adapter>/{package.json,package-lock.json}")`,
but `commonArgs.src` was `craneLib.cleanCargoSource ./.`, whose Cargo source
filter keeps `.rs` and every `.toml` but drops the six `.json` adapter
manifests (`package.json`/`package-lock.json`). In the evaluated Nix source the
three `acp-worker/adapters/<adapter>/` directories therefore contained no
manifests, so the `include_bytes!` reads had nothing to resolve against. (The
`.toml` files embedded via `include_str!`, such as `themes/builtin/*.toml` and
`plugins/*.toml`, were unaffected, since crane's filter retains all `.toml`.)

Fix: build `commonArgs.src` from `lib.fileset.toSource` unioning crane's
`commonCargoSources` with `./acp-worker/adapters`, so the six `.json` manifests
are retained while the source stays Cargo-clean (no `target/`, no
`node_modules`). `aoe-with-web` and the `aoe-clippy` / `aoe-fmt` / `aoe-test`
checks all reuse `commonArgs.src`, so the fix flows to every package that
embeds the manifests.

I also added a `Nix Build` CI job that compiles `.#default`. The existing
`Nix Eval` job only evaluates `drvPath` for each system and never builds a
Rust package, so it could not catch a source-filter regression that only
surfaces at compile time; `Nix Build Web` builds just the frontend. The new
job closes that gap. It is path-gated with the same `Detect relevant changes`
step as `Nix Build Web`, firing only when the flake recipe, workspace
manifests, crate sources, `build.rs`/`build_git_watch.rs`, the cargo build
config, the aoe crate's path dependencies, or the assets embedded via
`include_bytes!`/`include_str!` (`acp-worker/adapters`, `themes`, `plugins`)
change, so web-only, docs-only, and website-only PRs skip the expensive
compile on a concurrency-bound workflow.

`Nix Eval` is kept: it still covers all three systems, `aoe-with-web`, and the
devShells, and fast-fails on evaluation errors before the expensive compile.

### How tested

`nix eval` proof that the fixed `commonArgs.src` now contains exactly the six
manifests and nothing extraneous (cache-independent):

```
$ find $(nix eval --raw ...commonArgs.src...)/acp-worker/adapters -type f
.../acp-worker/adapters/claude-agent-acp/package-lock.json
.../acp-worker/adapters/claude-agent-acp/package.json
.../acp-worker/adapters/codex-acp/package-lock.json
.../acp-worker/adapters/codex-acp/package.json
.../acp-worker/adapters/pi-acp/package-lock.json
.../acp-worker/adapters/pi-acp/package.json
# json count in whole source: 6; no target/, no node_modules
```

`default` and `aoe-with-web` resolve to the identical fixed source path.

`nix build .#default` succeeds:

```
agent-of-empires>    Compiling agent-of-empires v1.13.2 (.../source)
agent-of-empires>     Finished `release` profile [optimized] target(s) in 4m 34s
$ echo $?
0
```

`cargo fmt --check` and `cargo clippy -- -D warnings` pass (run by the
cargo-husky pre-commit hook). `aoe-with-web` shares the same `src`; it was
verified by evaluation rather than a full build because the npm frontend build
requires network for `npmDepsHash` fetching, which is not available in this
sandbox.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle. The regression is a Nix packaging issue guarded by the new `Nix Build` CI job.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Opus 4.8) via opencode

**Any Additional AI Details you'd like to share:** AI drafted the flake source-filter fix and CI job; the author reviewed and verified the `nix build` and `nix eval` results.

## Summary

- Updated Nix source filtering to retain adapter manifests and `docker/Dockerfile`.
- Added an embedded-asset checker with self-tests for Rust include paths.
- Added CI validation for embedded assets on every pull request.
- Limited the slower Nix build job to `flake.nix` and `flake.lock` changes.
- Applied the source-filter update to the default, web, clippy, format, and test checks.

## Benefits

- Prevents missing embedded files during Nix builds.
- Detects source-filter regressions earlier.
- Reduces unnecessary Nix builds for unrelated changes.
- Provides clearer CI errors when embedded assets are missing.

## Potential Enhancements

- Extend the checker to cover additional dynamic asset-path patterns.
- Document the source-filter rules for future embedded assets.

## Self-review notes

A multi-agent self-review of this PR surfaced only non-blocking items; the core fix (source-filter union) is proven correct and complete (the evaluated Nix source equals the old Cargo source plus exactly the six adapter manifests, nothing dropped). Addressed in follow-up commits:

- Gated `.cargo/config.toml` (cargo reads it at build time) and `build_git_watch.rs` (it is `include!`d by `build.rs`), which crane keeps in the source but the path gate had missed, and reworded the gate comment so it justifies the non-obvious inclusions, not only the exclusions (`xtask`, `rust-toolchain.toml`).
- Tightened the job comments to the non-obvious rationale.

Known and deliberately not changed: (1) the intermediate commit messages are frozen (no force-push); a squash-merge takes the accurate title and this description. (2) `cachix-push.yml` has a pre-existing, unrelated `paths:` gap of the same class (it omits `acp-worker/adapters/**`, `themes/**`, `plugins/**`, and the path-dep crates), tracked as a follow-up rather than folded into this #3204 fix. (3) A pre-existing `cargo clippy --all-targets --features e2e-tests` dead-code warning in `src/session/poller.rs` is unrelated to this PR (no Rust changed here) and the repo's actual `Clippy` CI job (`cargo clippy -- -D warnings`) is green.


njbrake (commented):

The `flake.nix` fix is correct and I verified it at the pinned crane rev rather than from the description. Commenting rather than approving because of the CI half, which I think deserves a decision from @njbrake before it lands.

## The flake fix: verified good

Reading crane at the rev in `flake.lock` (`3d3de3313e263e04894f284ac18177bd26169bad`):

- `lib/filterCargoSources.nix` keeps only `*.rs`, `*.toml`, `Cargo.lock`, and an extension-less `.cargo/config`. The six adapter `.json` manifests were indeed dropped, so the diagnosis in #3204 is right.
- `lib/fileset/commonCargoSources.nix` exists at that rev and resolves to `cargoTomlAndLock + rust + toml`, so the replacement keeps everything the old filter kept. The only divergence is the extension-less `.cargo/config` form, and this repo ships `.cargo/config.toml`, so nothing regresses there.
- `src/acp/adapters.rs` is reachable in a default (non-serve) build (`pub mod acp;` in `lib.rs`, `pub mod adapters;` in `acp/mod.rs`), so `.#default` really does exercise those `include_bytes!` calls. The fix belongs on `commonArgs.src` where you put it.

## The CI job: worth a second look

Two things give me pause, and both are measurable rather than stylistic.

**1. It is now the longest job on the board.** The `Nix Build` job took 18.1 minutes on this PR. The next longest in the same run was `Cargo Test (linux, serve-rest)` at 6.8 minutes. The gate fires on `^src/`, which is most PRs, and the job installs Nix without a store cache (unlike `cachix-push.yml`, which uses `cachix-action`), so each run recompiles the dependency tree from scratch. That roughly triples the CI critical path for any Rust PR.

**2. The premise that no CI compiles a Rust Nix package is not quite right.** `.github/workflows/cachix-push.yml` builds `.#aoe-with-web` on every push to main matching `src/**`, `flake.nix`, `Cargo.*`, or `build.rs`. It did catch this regression: it has failed on every run since `e6dd11de` (#3101) landed at 2026-08-01T11:51, eight consecutive red runs as of today. So the gap that let #3204 reach users is an unwatched post-merge signal, not a missing build.

That reframes the fix. Cheaper options that close the same class:

- Alert on a `cachix-push` failure, so an existing 20-minute build that already runs stops being invisible.
- Narrow the new job's gate to `flake.nix|flake.lock` and add a seconds-long static guard that every `include_bytes!` / `include_str!` asset path outside `src/` survives the fileset. That covers both regression shapes (a filter change and a new embedded asset) without a compile. I could not run `nix` in my sandbox, so treat the "expose `commonArgs.src` and assert against it" variant as untested rather than a verified recipe.
- Keep the job but add a cachix pull, so the dependency derivation comes from the cache `cachix-push` already populates on main. That should take most of the 18 minutes off.

If you would rather have a hard pre-merge gate at full cost, that is a reasonable call and this job is a correct implementation of it. I just do not want the cost paid silently.

## Two smaller notes

- `set -euo pipefail` plus a `grep` in a pipeline means a `git diff` failure fails the pipeline, which the surrounding `if` reads as "not relevant" and skips the build. It matches the existing `nix-build-web` step, so this is a pre-existing pattern, not something introduced here, but the failure mode is a silent skip rather than a loud error. This is also now the third copy of the `Detect relevant changes` step.

- Adjacent and not this PR's bug: `checks.aoe-test` and `checks.aoe-clippy --all-targets` compile test code, and `dockerfile_pin_matches_floor` in `src/acp/agent_compat.rs` does `include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/docker/Dockerfile"))`. No fileset in the flake keeps `docker/Dockerfile`, so those checks look like they cannot compile today, before or after this change, and the new job does not cover them (`.#default` builds with `doCheck = false`). I could not run `nix` here to confirm. Worth a follow-up issue; the claim that the fix flows to every package holds for the manifests but not for those two checks.

Net: the `flake.nix` change is good to land as-is. Leaving the call on the CI job to @njbrake.

njbrake:

@jerome-benoit I pushed two commits to this branch rather than send it back to you, since the flake fix itself was already right and I only wanted to change how it is guarded. Revert or push over anything you disagree with.

**`f5549b4` fix(nix): keep docker/ in the Nix build source.** While checking your fix I found a second instance of the same class: `dockerfile_pin_matches_floor` in `src/acp/agent_compat.rs` embeds `docker/Dockerfile` with `include_str!`, and the filter drops it. The two packages build anyway because they set `doCheck = false`, but `checks.aoe-test` and `checks.aoe-clippy` compile test code, so neither can build from the flake today. One line in the union.

**`95594c8` ci: check embedded assets on every PR, gate Nix Build on the flake.** This is the substantive change, and it swaps the coverage model rather than reducing it.

`scripts/check-nix-embedded-assets.py` resolves every `include_bytes!` / `include_str!` / `include!` path in the crate (both the plain literal and the `concat!(env!("CARGO_MANIFEST_DIR"), ...)` form), parses the extra paths out of the `fileset.unions` list in `flake.nix`, and fails if any asset is neither a `.rs`/`.toml` nor under one of those paths. It runs before Nix is even installed, so it costs a couple of seconds on every PR. Two deliberate properties: it fails loudly on an include form it cannot resolve, rather than skipping it silently, and it fails loudly if `flake.nix` stops using a crane source filter, since its `.rs`/`.toml` assumption would no longer hold. `--self-test` covers the predicate and the flake parser.

Proof that it catches the bug this PR fixes: with `./acp-worker/adapters` removed from the union, it reports all six manifests with file and line, and exits 1. On the fixed tree it reports 40 embedded assets all surviving.

With that guard on every PR, the compile-based `Nix Build` job no longer needs the broad `^src/` gate, so I narrowed it to `flake.nix` and `flake.lock`. The reason is the number in my review: that job took 18.1 minutes here against 6.8 for the next longest job, with no binary cache, and the old gate fired it on most Rust PRs. What is left is the part a fast checker genuinely cannot cover (a flake recipe or pinned input that compiles nowhere), and that only changes in PRs like this one. Post-merge, `cachix-push.yml` still builds a Rust package on every push to main touching `src/**`, `Cargo.*` or `build.rs`.

Worth knowing separately: that cachix job has been red on main since #3101 landed, which is why #3204 reached users at all. It should go green again once this lands, and I will open a follow-up about alerting on it.

`Nix Build` still fires on this PR (it touches `flake.nix`), so the `./docker` addition and your original fix both get a real compile before merge.

njbrake (approved):

Approving and merging. Thanks for tracking this down, @jerome-benoit; the diagnosis in #3204 was right and the fileset union is the correct shape for the fix.

Recording what the run proved, since the point of the two commits I pushed was to swap the coverage model rather than shrink it:

- `Nix Eval` ran the new checker in 0.6 minutes total for the job: `self-test passed`, then `OK: 40 embedded asset(s) all survive the Nix build source (extra fileset paths: acp-worker/adapters, docker)`. That covers the #3204 failure class on every PR from here on.
- `Nix Build` fired (this PR touches `flake.nix`) and passed in 17.9 minutes, so both your original fix and the `./docker` addition got a real compile. That is also the last time it will run until someone next changes the flake recipe.
- All 28 checks green. `codecov/project` flickered red mid-run at 84.74% while the Playwright live shards were still uploading, then settled green once their LCOVs landed; `codecov/patch` read "coverage not affected" throughout.

Follow-up I will open separately: `cachix-push.yml` has been red on main since #3101, which is the only reason this reached users at all. It should recover with this merge, and it needs an alert so the next one does not sit unnoticed.

PR #3209, opened 2026-08-01
fix(tui): stop TUI from persisting structured-session status, and two related daemon-apply guards (#3201)

### LINKED ISSUE #3201: TUI persists structured-session status to disk, so a stopped daemon leaves rows stuck across restarts

Follow-up to #3170, found in review after it merged. Verified against `main` at cbe7e0b2.

## The problem

#3170's own description rules out persisting structured status:

> The persist-to-disk route was rejected because it reverses the documented #2690 contract and, with no tmux poller able to heal a structured row, a daemon crash mid-turn would leave a session stuck spinning `Running` forever.

The TUI does persist it, through an inherited call rather than a deliberate one. `apply_daemon_status_update` routes into the shared `apply_status_update`, which on any status change calls `persist_passive_status_transition` (`src/tui/home/mod.rs:3060`). That builds a `PassiveStatusPatch::from_instance` (which carries `status`) and applies it via `merge_passive_status_patch`, whose first line is an unconditional `self.status = patch.status` (`src/session/instance.rs:1730`). Neither `persist_passive_status_transition` (`src/tui/home/mod.rs:6297`) nor the merge has an `is_structured()` gate.

The asymmetry is explicit in the code. The daemon's own passive writer refuses the same write:

```rust
// src/server/mod.rs:4023
(!inst.is_structured()).then(|| crate::session::PassiveStatusPatch::from_instance(inst));
```

So the TUI writes to `sessions.json` exactly what the daemon deliberately declines to write.

## Why it does not self-heal

#3170 also moved the tmux poller's structured bail above `update_status_with_metadata` (`src/tui/status_poller.rs:212`), which was correct for the phantom container error it was fixing, but it removed the only local producer that used to heal a structured row's status. Combined with the persist above:

1. Daemon reports `Running` mid-turn, TUI persists `running` to disk.
2. User runs `aoe serve --stop`, or the daemon crashes.
3. `fetch_structured_statuses` returns empty by design, and the tmux poller returns early for structured rows.
4. `polling_tier(Status::Running)` is `TIER_HOT`, so the poller runs and bails. Nothing else touches the row.
5. The pill spins `Running` forever, **and survives a TUI restart**, because disk now says `running`. There is no load-time status reset in `Storage::load` / `load_all_instances`.

Same shape for `Error`: a session that hit `AgentStartupError` when the daemon went away stays red with no producer able to clear it.

`seed_acp_statuses` does recover it once a daemon comes back, so this is not permanent damage. But "start a daemon to unstick your session list" is the same class of problem the v016 and v023 migrations exist to clean up, and #3170 shipped v023 to clear exactly this kind of poisoned row while adding a new generator of it.

## Suggested fix

Gate the TUI's passive writer the same way the daemon's is gated, in `persist_passive_status_transition`:

```rust
pub(super) fn persist_passive_status_transition(&self, id: &str, mark_unread: bool) {
    let Some(inst) = self.instances.get(id) else {
        return;
    };
    // Structured rows are not durable. The daemon's own passive writer gates
    // on the same predicate (`decide_passive_transition` returns `patch: None`
    // for `is_structured()`), per the durability contract on
    // `apply_acp_overlay_inplace`: their status is an in-memory overlay
    // rebuilt from live worker state and re-derived at daemon boot by
    // `seed_acp_statuses`. Writing it here strands a row at `running` or
    // `error` with no producer left to heal it once the daemon is gone, since
    // the tmux poller now bails on structured rows.
    if inst.is_structured() {
        return;
    }
    let Some(storage) = self.storages.get(&inst.source_profile) else {
        return;
    };
    // ...
}
```

This cannot regress what it replaces: the only structured writes reaching this path before #3170 were the healed statuses the new bail no longer produces. Explicit user actions (stop, start, delete) go through `apply_user_action`, not this function, so they still persist.

Worth pairing with a unit test that a structured `Idle -> Running` apply leaves `sessions.json` untouched, since the absence of that test is what let this through.

## Two related things worth deciding at the same time

- **`apply_status_update` mutates on every tick, not only on change.** `should_update` tests the local status against `Deleting|Creating|Stopped` and the incoming against `Stopped`, but does not require `old != new`. At the daemon poll cadence an unchanged reading rewrites `last_error` and `idle_entered_at` every tick, and the daemon reports `last_error: null` for most structured rows because the field is `#[serde(skip)]`. That can wipe a locally-set message, e.g. the delete-failure text from `apply_deletion_results`.
- **Archived and trashed rows are not excluded.** `daemon_status_applies_to` only filters `recovery_in_flight` / `restart_in_flight`, and `/api/sessions` returns archived and trashed rows unfiltered. The `is_archived()` short-circuit that protects them from the tmux producer lives inside `update_status_with_metadata_inner`, which this path never reaches, so a sunk row can be restamped and re-marked unread. That is the #1868 / #2206 regression.

jerome-benoit:
## Description

Follow-up to #3170, fixing the three issues raised in #3201. The daemon-status
apply path the TUI grew in #3170 was grafted onto the tmux-shaped
`apply_status_update`, inheriting its persist call and its `should_update`
predicate while bypassing the two guard points the tmux producer relied on. Each
of the three bugs is a guard that lives on a path this daemon flow never
traverses. All three are fixed here at the root, each with a regression test
that fails before the fix and passes after. The maintainer-review follow-up
below refines the P2 and P3 fixes; every commit is atomic.

### P1: TUI persists structured-session status to disk

**Root cause.** `apply_daemon_status_update` routes into the shared
`apply_status_update`, which on a genuine transition calls
`persist_passive_status_transition` (`src/tui/home/mod.rs`). That writer built a
`PassiveStatusPatch::from_instance` and applied it via
`merge_passive_status_patch`, whose first line is an unconditional
`self.status = patch.status` (`src/session/instance.rs`). Neither had an
`is_structured()` gate. The daemon's own passive writer refuses the same write:
`decide_passive_transition` returns `patch: None` for `is_structured()`
(`src/server/mod.rs`). So the TUI wrote to `sessions.json` exactly what the
daemon deliberately declines to write. It does not self-heal because #3170 moved
the tmux poller's structured bail above `update_status_with_metadata`
(`src/tui/status_poller.rs`), removing the only local producer that used to heal
a structured row, and `Storage::load` has no load-time status reset. Effect: a
row stamped `Running`/`Error` mid-turn spins forever and survives a TUI restart.

**Proof.** The regression test seeds a structured row on disk as `Idle`, applies
a daemon `Running`, and asserts the disk row stays `Idle`. It failed before the
fix (disk read `Running`) and passes after.

**Fix.** Gate the status patch on `is_structured()` in
`persist_passive_status_transition`, mirroring the daemon. The unread mark is
deliberately left ungated: the daemon marks a structured row unread on a
`Running -> Idle` turn regardless of view (`decide_passive_transition`'s
`mark_unread` is not gated on `is_structured`), so this keeps the two producers
symmetric rather than introducing a new divergence.

**Regression risk.** None. The only structured writes reaching this path before
#3170 were the healed statuses the new bail no longer produces. In-memory status
still updates, so the pill still moves live.

> **Maintainer correction (@njbrake).** This gate closes the passive writer, but
> it does not close #3201 on its own. `save` -> `merge_from_tui` is not only the
> explicit-user-action route: `HomeView::save` flushes *every* in-memory row
> through `merge_from_tui`, whose first line is an unconditional
> `self.status = src.status` with no `is_structured()` gate. A daemon-sourced
> status therefore still reaches `sessions.json` on the next `save()`, which runs
> on TUI quit and from roughly forty other call sites under `src/tui`. Reproduced
> on this branch: apply a daemon `Running` to a structured row, call `save()`,
> read the row back from `Storage`, and disk holds `running`. The #3201 sequence
> (quit mid-turn, stop the daemon, relaunch) is still reachable. Tracked in
> #3201, which stays open.

### P2: `apply_status_update` wiped a local `last_error` on every tick

**Root cause.** `should_update` excludes local `Deleting|Creating|Stopped` and an
incoming `Stopped`, but does not require `old != new`. At the daemon poll cadence
an unchanged reading still entered the mutate block and rewrote `last_error`
every tick. The daemon reports `last_error: null` for most structured rows
(the in-memory field is `None` for a row with no ACP error), so an unchanged tick
wiped a locally-set message.

> **Maintainer correction (@njbrake).** The original example here was the
> delete-failure text `apply_deletion_results` leaves on a row. That case is not
> actually protected by this fix: `apply_deletion_results` sets
> `inst.status = Status::Error` together with the message, so the next daemon
> tick reporting the row's live ACP status makes `status_changed` true and clears
> `last_error` anyway. Verified on this branch. The fix is still a real
> improvement for a same-status tick, which is what its regression test locks;
> only the motivating example was wrong. The converse also has to be handled:
gating the entire `last_error` write on a status change is itself lossy, because
the daemon can carry a present `last_error: Some(..)` for an `Error` row, so an
`Error -> Error` tick with a changed message would be dropped and the first error
would freeze on the row.

**Proof.** Two regression tests. The first sets a local `last_error`, applies an
unchanged daemon tick carrying `last_error: None`, and asserts the message
survives; it failed before the fix
(wiped to `None`) and passes after. The second seeds an `Error` row, applies an
`Error -> Error` tick carrying a new `Some(..)`, and asserts the new message
lands; it failed before the fix (the old first message stayed) and passes after.

**Fix.** Handle the two directions separately: apply an incoming `Some`
unconditionally (the daemon's present `last_error` is authoritative for an ACP
error), and clear `last_error` only across a genuine transition. An unchanged
tick carrying `None` is treated as non-authoritative, since the daemon tracks
only ACP errors and cannot distinguish "no error" from a locally-set message, so
it leaves that message in place. `status`, `idle_entered_at`,
`last_accessed_at`, `live_status_baseline`, and `pane_dead_observed` keep their
existing per-tick semantics, which is required: `idle_entered_at` must still
apply on a same-status Stop-hook tick (#872,
`apply_status_update_propagates_idle_entered_at_into_live_instance`) and
`live_status_baseline` must still seed every tick (#2690,
`apply_status_update_propagates_live_status_baseline_from_poller`).

**Regression risk.** Low. A present incoming message always applies, so a real
error is never dropped; a `None` clears only on a transition, which
`daemon_status_clears_a_stale_error_message` locks. A stale local message on a
same-status row is deliberately retained until the next transition, because on an
unchanged tick the daemon's `None` cannot be told apart from a live local
message. Both #872 and #2690 anchor tests still pass.

### P3: archived and trashed rows were not excluded from the daemon apply

**Root cause.** `daemon_status_applies_to` filtered only `recovery_in_flight` /
`restart_in_flight`. `/api/sessions` returns archived and trashed rows
unfiltered, and the `is_archived()` short-circuit that keeps the tmux producer
off a sunk row lives in `update_status_with_metadata_inner`
(`src/session/instance.rs`), a path the daemon overlay never reaches. So a sunk
structured row was restamped and, on a `Running -> Idle` turn, re-marked unread.
That is the #1868 / #2206 regression.

**Proof.** The regression test archives (and, in a second case, trashes) a
structured row, applies a daemon `Running`, and asserts the status stays `Idle`.
It failed before the fix (restamped to `Running`) and passes after.

**Fix.** Exclude `is_archived()` and `is_trashed()` rows in
`daemon_status_applies_to`. (Maintainer correction, @njbrake: this mirrors the
tmux producer only for archived rows. `update_status_with_metadata_inner`
short-circuits on `is_archived()` and has no `is_trashed()` short-circuit.
Excluding trashed rows is still correct, since the ACP reconciler filters both.)
Snoozed
rows are intentionally left live, matching the tmux producer which does not
short-circuit them, so live agents keep updating while snoozed.

**Regression risk.** None found. The daemon never spawns a worker for an
archived/trashed row (the reconciler filters them), so it does not emit a live
status for one; this gate only drops the stale restamp. Unarchive/restore go
through `apply_user_action`, not this path.

## Maintainer review follow-up (njbrake)

The maintainer's early (non-gating) review verified P1 and P3, and raised one
substantive finding plus one nit on P2. Both are addressed:

- **F1, P2 gate breadth.** The original P2 gated the entire `last_error` write on
  a status change, which the review correctly flagged as not lossless: the daemon
  carries `last_error: Some(..)` for an `Error` row, so an `Error -> Error` tick
  with a changed message was dropped and the first error froze on the row. Fixed:
  an incoming `Some` is applied unconditionally and a `None` clears only across a
   genuine transition. The same-status `last_error` policy is locked by
   `daemon_status_reconciles_last_error_on_a_same_status_tick` (a `None` tick
   keeps a local message; a `Some` replaces it), and the healing of a stale
   error on a genuine transition by `daemon_status_clears_a_stale_error_message`.
   The earlier "lossless" wording in this body has been corrected accordingly.

- **F2, double `get_instance` lookup (nit).** `apply_daemon_status_update` did one
  `get_instance` for the `is_structured()` check and `daemon_status_applies_to`
  did a second immediately after. Fixed: a single lookup feeds both guards, and
  `daemon_status_applies_to` now takes `&Instance` instead of `&str`. Behavior is
  unchanged (the guard set is identical), and the scoped borrow drops before the
  later `Stopped`-lift mutation.

Comment and test-rationale wording was tightened in a follow-up docs commit (no
behavior change).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (N/A: internal behavior only)
- [x] For UI changes: included screenshot or recording (N/A: no rendering change; status-apply logic only)

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (the change is TUI-side daemon-status apply logic; no `web/` files touched)

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: each behavior is exercised by a deterministic unit test that drives `apply_daemon_status_update` directly against the real `Storage`-backed test harness (`daemon_status_does_not_persist_a_structured_row_to_disk`, `daemon_status_persists_the_unread_mark_but_not_the_status_for_structured`, `daemon_status_reconciles_last_error_on_a_same_status_tick`, `daemon_status_skips_a_sunk_structured_row`, `daemon_status_applies_to_a_snoozed_structured_row` in `src/tui/home/tests.rs`), each red-before / green-after. An e2e for P1 would need a live daemon mid-turn plus a stop and a TUI restart to observe the persisted row; the unit tests model that exact sequence at a fraction of the cost, per the test-tier guidance in AGENTS.md.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Anthropic), driven via OpenCode.

**Any Additional AI Details you'd like to share:** Root-cause analysis and fix design were cross-validated across multiple independent analysis passes; every fix is confined to its converged root cause and paired with a red-before / green-after regression test.

Refs #3201 (kept open: see the maintainer correction under P1)

## Summary

- Keeps structured-session status in memory without writing it to disk.
- Preserves unread markers and avoids empty persistence writes.
- Applies new `last_error` messages and preserves local errors when ticks are unchanged.
- Excludes archived and trashed rows from daemon updates.
- Keeps snoozed rows active.
- Adds regression tests for persistence, error handling, and row filtering.
- Reuses instance lookups and improves test isolation.

These changes prevent stale session states, lost error messages, and unwanted updates to inactive rows.


jerome-benoit:
## Self-review addendum

Ran an initial self-review of this PR with three independent analyses (correctness/tests, design/Rust, scope/harmonization), cross-validated. Verdict: **approve-with-nits**; all three fixes are correct at the root, behavior-preserving, and clean on scope/AGENTS.md/terminology. One reviewer reverted each fix in isolation and confirmed all four tests fail-before / pass-after.

Three minor findings, all addressed (comment-accuracy and test-locality only, no behavior change):

- **F1 (last_error gate comment):** the comment justified only the *set* direction, while the `old != new` gate also suppresses *clearing* `last_error` on an unchanged tick. Verified still lossless (a `last_error` is only set as a row enters `Error` and only cleared as it leaves `Error`, so both directions coincide with a status change), and extended the comment to say so. In a follow-up docs commit.
- **F2 (passive-writer early return):** the `structured && !mark_unread` return is a flock-avoidance optimization, not a correctness gate; clarified the comment so a future reader does not misread it. In a follow-up docs commit.
- **F3 (test locality):** pinned `unread_enabled` in the structured-persist test so its assertion cannot couple to the process-global under a future edit, matching the sibling parity test. In a follow-up docs commit.

Rejected on proof: adding a set-on-transition `last_error` test (same code line as the already-covered clear-on-transition path); the implicit `#[serial]` justification (pre-existing module convention, not introduced here).

Validation on the fixed tree: `cargo fmt --check` clean; `cargo clippy --features serve --all-targets` clean for the changed files (the 7 remaining warnings are pre-existing and outside this PR's files); `cargo test --features serve --lib` = 5931 passed, 0 failed. All changed code lives in the library. Three `tests/acp_runner_control.rs` integration cases failed in this local run, but they are unrelated to and unaffected by this diff (that binary spawns the compiled `aoe`, which is byte-identical here since these two follow-up commits are comments plus a `#[cfg(test)]` line); they are an environmental flake on this machine and pass in a clean run.

njbrake (commented):

Early feedback while this is still a draft, so none of it is gating. The root-cause writeup checks out everywhere I followed it, and P1 in particular is a real durability bug worth fixing before it strands more rows.

## What I verified

**P1's premise is precise.** `decide_passive_transition` in `src/server/mod.rs` reads:

```rust
let patch = (!inst.is_structured()).then(|| crate::session::PassiveStatusPatch::from_instance(inst));
```

with `mark_unread` computed independently and not gated on `is_structured()`. The fix in `persist_passive_status_transition` mirrors that expression, including leaving the unread mark ungated, so the two producers stay symmetric. The daemon side already has `decide_passive_transition_skips_patch_for_structured_session` (#2697) locking its half, and the new `daemon_status_persists_the_unread_mark_but_not_the_status_for_structured` locks the other. That pairing is a nice touch.

**P3 is safe.** `apply_daemon_status_update` already returns early when `get_instance(&update.id)` is absent or not structured, before `daemon_status_applies_to` is consulted, so requiring the instance to exist inside `daemon_status_applies_to` cannot newly drop an unknown id. Updates were already consumed rather than queued, so nothing is deferred and lost either.

## One finding: P2's gate is broader than the justification

The reasoning given is that `last_error` is only ever set as a row enters `Error` and only ever cleared as it leaves, so gating both directions on `status_changed` is lossless. Two cases fall outside that:

1. **`Error -> Error` with a changed message.** The daemon can carry `last_error: Some(...)` (there is a poller test doing so with `"agent failed to start"`), so a second, different failure on an unchanged status is now dropped and the older text stays on the row.

2. **A stale error on a row that is not in `Error`.** The comment on the existing `daemon_status_clears_a_stale_error_message` test describes the original bug as sandboxed structured rows left "at Idle with a phantom `Container is not running` hanging off them." That test seeds `Error -> Idle`, so it still passes. But a row already sitting at `Idle` carrying a phantom error will no longer be healed by an unchanged tick; it waits for the next genuine transition, which for an idle structured row can be a long time. That is the shape of the bug the earlier fix was written for.

Both are addressed by splitting the two directions instead of gating them together: always apply an incoming `Some`, and gate only the clear on `status_changed`. That still fixes the reported bug, since the daemon's `None` on an unchanged tick no longer wipes a locally-set message, while letting a real message update through.

If you would rather keep the single gate, the PR text is worth softening, since it currently asserts losslessness and these two cases are counterexamples.

## Nit

`daemon_status_applies_to` performs a second `get_instance` lookup immediately after the caller has done one. Threading the instance in would avoid it, though at this cadence it costs nothing measurable.

The test set is otherwise well-shaped: red-before and green-after per commit, one table covering both sunk-row cases rather than two functions, and each test naming the guard it locks.

jerome-benoit:
## Re-review addendum (convergence rounds)

Ran a follow-up re-review with the multi-agent method (three fresh independent analyses per round, cross-validated), looping until findings stopped being actionable.

- **Round 1** found one nit: the `last_error` gate comment overstated a global invariant ("only ever cleared as it leaves `Error`") when restart and recovery also clear it while entering `Starting`. Fixed by scoping the comment to the producers that reach this apply path, then correcting a mechanism misattribution (recovery persists via the recovery cascade, not `apply_user_action`/`save`). In follow-up docs commits.
- **Round 2** found two comment-precision nits (the "every producer pairs a change with a status change" phrasing, and an imprecise "skipped on the wire" note) plus two doc nits (the `daemon_status_applies_to` docstring opener over-attributed exclusions to `pollable_instances`; the PR body's test list omitted the parity test). Fixed: rewrote the `last_error` comment to a minimal message-focused form, generalized the docstring opener to attribute each exclusion class correctly, and added the fourth test to the body. A follow-up docs commit plus this body edit.
- **Round 3**: three independent reviewers all approved. Each re-verified mutation-kill (reverting any fix fails its test, including the archived and trashed halves of the P3 table), confirmed both tightened comments are accurate and appropriately scoped, and confirmed every PR-text claim against the code. The only residual items were contradictory wording preferences on an already-accurate comment and non-blocking style notes, so no further edits were made to avoid churn.

Throughout, the three behavioral fixes (P1/P2/P3) were unchanged and correct at the root; only comments, one test-locality pin, and PR text were refined.

Validation on the final tree: `cargo fmt --check` clean; `cargo clippy --features serve --all-targets` clean for the changed files (the 7 remaining warnings are pre-existing and outside this PR's files); `cargo test --features serve --lib` = 5931 passed, 0 failed. All changed code lives in the library.

jerome-benoit:
## Initial-review round

Ran a fresh initial review with the multi-agent method (three independent analyses on correctness, design/style, and regression/PR-text, cross-validated), then designed and implemented the fixes with a second three-agent pass. Verdict of the review: no blocking or major issues; the three fixes (P1/P2/P3) are correct at the root, behavior-preserving, and scope-clean. Three findings were actionable and are now addressed; three were rejected with proof.

- **F-A (LOW), stale commit hashes in the body.** The "Maintainer review follow-up" section pinned three commit hashes that the rebase onto `main` orphaned (`git merge-base --is-ancestor` returns false for each), so browsing them fails. Fixed by describing each change by its commit subject instead of a hash, which survives future rebases. Also dropped the now-inaccurate "table-driven over `Running` and `Idle`" phrasing (see F-B).

- **F-B (minor), tautological test row.** `daemon_status_unchanged_tick_keeps_a_local_last_error` looped over `[Running, Idle]`, but both drive the identical no-branch path in `apply_status_update` (`status_changed` false, daemon `None`, so neither `last_error` branch fires and the transition block is skipped), so the `Idle` row could not fail independently of `Running`. Reduced to the single `Running` case (the realistic mid-turn scenario) and trimmed the doc-comment. No branch coverage lost.

- **F-E (nit), third `get_instance` in the Stopped-lift.** After the F2 guard consolidation, the Stopped-lift still did a third `get_instance` to read the pre-lift status. Since nothing mutates between the guard and the lift, that status is captured as a `Copy` bool (`was_stopped`) while the guard borrow is live and reused, so a single lookup now feeds both guards and the lift. Behavior is identical and the bool drops before the `mutate_instance` borrow.

Rejected with proof:

- **Comment density** in the passive writer: load-bearing and accurate; tightening it further is wording preference on already-correct prose.
- **Commit-history churn** (several `docs(tui):` iterations): a squash is a history rewrite that is the maintainer's call, not a code defect.
- **`structured_row` hardcoding `tool = "claude"`**: a test helper with no bearing on any assertion.

Gates on the final tree: `cargo fmt --check` clean; `cargo clippy --features serve --all-targets` clean for the changed files (the 7 remaining warnings are pre-existing and outside this PR's changed regions); `cargo test --features serve --lib` = 5985 passed, 0 failed. The three behavioral fixes (P1/P2/P3) are unchanged; this round touched only a lookup refactor, a test simplification, and this body.

jerome-benoit:
## Convergence round

Re-reviewed the PR with three fresh independent agents (correctness, design/style, regression/PR-text), cross-validated. The three prior findings (stale body hashes, the Stopped-lift lookup reuse, the reduced unchanged-tick test) were confirmed fixed with proof; the three behavioral fixes P1/P2/P3 remain correct at the root, and the shared `apply_status_update` path is unaffected on the tmux side. Verdict: approve, no blocking or major issues. One new actionable nit, addressed:

- **Orphaned commit hashes in earlier addendum comments (LOW).** The rebase onto `main` orphaned the SHAs cited in the two earlier review-addendum comments (`git merge-base --is-ancestor` returns false for each), so clicking them 404s, the same defect that was removed from the body. The two comments were edited to describe those changes by their nature (follow-up docs commits) instead of by a volatile hash.

Rejected with proof (no change):

- **`daemon_status_skips_a_sunk_structured_row` uses `if label == "archived"`.** The test is already table-driven over `["archived", "trashed"]` in a single test, which is the house style; the `if` only selects which sunk field the setup writes. Converting it to a closure table adds type annotations for no real gain.
- **The `last_error` gate comment is framed around the daemon.** The daemon is the only producer for which the asymmetry has any effect: the tmux producer carries `inst.last_error` as a self-copy, so the "apply Some always" branch is a no-op there, and structured rows never enter the tmux projection. The framing is accurate.
- **`#[serial]` on the new tests, commit-history churn, `is_trashed` absent from the tmux short-circuit.** Respectively: matches the module-wide convention; a squash is the maintainer's call at merge; the daemon excluding trashed rows is a defensible superset of the tmux archived short-circuit.

Gates unchanged (code tree untouched this round): `cargo fmt --check` clean, `cargo clippy --features serve --all-targets` clean for the changed files, `cargo test --features serve --lib` = 5985 passed, 0 failed.

jerome-benoit:
## CodeRabbit review follow-up

Thanks @coderabbitai. All three nitpicks were validated against the current code (three independent analyses, cross-validated) and all three are addressed.

- **Doc comment on `apply_daemon_status_update` (mod.rs).** Valid: after the P1 fix, `persist_passive_status_transition` skips the `sessions.json` status patch for structured rows, and this path only ever handles structured rows, so the "the `sessions.json` patch behaves exactly as a tmux-derived transition" claim was wrong. Reworded to state that the status patch is the deliberate exception (only the unread mark persists), while sounds, hooks, and unread marking still match the tmux path. Commit `18fc8bb8`.

- **Snoozed row stays live (tests.rs).** Valid coverage gap: `daemon_status_applies_to` excludes archived and trashed but deliberately not snoozed (#3201 keeps snoozed live), and nothing locked that asymmetry. Added `daemon_status_applies_to_a_snoozed_structured_row` (sets `snooze(30)`, applies a daemon `Running`, asserts the status becomes `Running`). Not tautological: injecting a symmetric `&& !inst.is_snoozed()` into the guard makes it fail (`left: Some(Idle)`, `right: Some(Running)`), so it guards the exact regression you flagged. A separate function rather than a row in `daemon_status_skips_a_sunk_structured_row`, since the setup and the assertion are the opposite (live, not sunk). Commit `daa6ffe5`.

- **Consolidate the two same-status `last_error` tests (tests.rs).** Valid: both were same-status ticks (`status_changed == false`) varying only the incoming value (`None` keeps the local message, `Some` replaces it), with identical structural setup, so they are input permutations of one behavior. Merged into `daemon_status_reconciles_last_error_on_a_same_status_tick` as a two-row table; the assertion count is unchanged and per-row comments keep the intent. The clear-on-transition path stays its own test (`daemon_status_clears_a_stale_error_message`). Commit `daa6ffe5`.

Gates on the final tree: `cargo fmt --check` clean; `cargo clippy --features serve --all-targets` clean for the changed files (the 7 remaining warnings are pre-existing and outside this PR's changed regions); `cargo test --features serve --lib` = 5985 passed, 0 failed.

njbrake (approved):

Approving. I re-derived all three fixes from the code rather than from the write-ups on this thread, and found no defect in any of them. P1 mirrors `decide_passive_transition` including leaving `mark_unread` ungated, the P3 single-lookup refactor is behavior-preserving because `was_stopped` is captured before the `Stopped` lift and nothing mutates in between, and the `last_error` split is correct in memory (the field is `#[serde(skip)]`, and the tmux poller echoes the row's own value back, so the only changed case is a locally-set `Some` after the poller snapshot, which the change now preserves). Four of the five new tests fail against `main`, so they are real regression tests rather than lock-ins.

I pushed 942e66a9, a comment only. P3's exclusion means a sunk structured row already in `Error` has no producer left to clear it: the daemon is excluded here, the tmux poller bails on structured rows before probing, and `reload_storage_only` carries `prev.status` forward. It stays on screen because `agent_row_icon` lets `Error` punch through the sunk-row mask deliberately. The tmux producer has the same property, so this is consistent rather than new, but it was not written down.

I also corrected three claims in the description, each marked inline. The important one is P1. Your own risk note cites `save` -> `merge_from_tui` as the separate route for explicit user actions, but `HomeView::save` flushes every in-memory row through `merge_from_tui`, whose first line is an unconditional `self.status = src.status` with no structured gate. So a daemon-sourced status still reaches `sessions.json` at the next `save()`, which runs on quit and from roughly forty other call sites. I reproduced it on this branch. #3201's sequence is still fully reachable, so `Closes` is now `Refs` and the issue stays open.

That is not a reason to hold this. All three fixes stand on their own, P1 makes the TUI symmetric with the daemon writer regardless, and the remaining leak is a design question rather than a patch, since `save()` is also the explicit-action route. I will add my reasoning to #3201.

One request for future PRs: the self-review addenda on this thread assert convergence across independent passes, and neither the `save()` leak nor the P2 example survived checking. Please treat those as notes to yourself rather than as evidence for a reviewer; they raised the cost of reviewing this rather than lowering it. The underlying work here is good and I would rather see it with a shorter description.

Issue #2532, opened 2026-06-28
Trash purge can try to delete branches while preserving their worktrees


jerome-benoit:
## Summary

The web permanent-delete and retention auto-purge paths can pass `delete_branch = true` while `delete_worktree = false`. `perform_deletion()` then attempts branch deletion even though the preserved worktree still has that branch checked out, so purge fails or leaves noisy branch-cleanup errors for a config combination that is allowed by the UI/settings.

## Steps / code path

1. Configure `worktree.auto_cleanup = false` and `worktree.delete_branch_on_cleanup = true`.
2. Permanently delete a trashed worktree/workspace session from the web, or let retention auto-purge an expired trashed session.
3. The web cleanup defaults expose `delete_worktree = cfg.worktree.auto_cleanup` and `delete_branch = cfg.worktree.delete_branch_on_cleanup` independently.
4. Retention auto-purge builds `DeleteSessionBody { delete_worktree: cfg.worktree.auto_cleanup, delete_branch: cfg.worktree.delete_branch_on_cleanup, ... }`.
5. `perform_deletion()` resolves `branch_to_delete` whenever `request.delete_branch` is true, and treats `worktree_ok` as true when `request.delete_worktree` is false.

## Expected

Branch cleanup should be tied to successful worktree cleanup, or disabled when preserving the worktree. Keeping the worktree means the checked-out branch must remain available.

## Actual

The purge path attempts `git branch -d/-D <branch>` while the worktree is intentionally preserved. Git rejects deleting a branch checked out in another worktree, so the purge can fail or report branch cleanup errors even though the user/config chose to keep the worktree.

## Evidence

On `main`, `src/server/api/sessions.rs` sets cleanup defaults and retention purge bodies with independent `delete_worktree` and `delete_branch` fields. In `src/session/deletion.rs`, branch cleanup runs if `delete_branch` is true; `worktree_ok` is `!request.delete_worktree || ...`, so preserving the worktree is considered OK for branch deletion. `GitWorktree::delete_branch()` ultimately runs `git branch -d` and then `-D`, which cannot delete a branch currently checked out by the preserved worktree.

## Deduplication

This is distinct from #2525, which covers CLI `rm --purge --delete-branch` skipping branch cleanup for multi-repo workspaces. This issue covers web/manual purge and retention auto-purge enabling branch cleanup when worktree cleanup is explicitly off.

Seluj78:
you or me ? 👀

PR #3216, opened 2026-08-03
fix(session): never remove a default-branch worktree or delete the branch

### LINKED ISSUE #3215: Deletion paths can remove a default-branch worktree and delete the default branch itself — no guard
## Summary

AoE has no guard against deleting a session whose worktree is the repository's **default-branch checkout**. Every deletion path (TUI `d`, web/API DELETE, trash auto-purge, `empty-trash`) will remove the managed worktree and — with `worktree.delete_branch_on_cleanup = true` — delete the default branch itself, even in a bare-repo layout where that branch is `HEAD`.

This breaks externally-managed repo layouts where the default-branch worktree is permanent infrastructure, not a disposable per-task checkout.

## Version

aoe 1.13.2 (also verified against current `master`)

## Deletion paths and their gates

| Delete path | Worktree removed? | Branch deleted? |
|---|---|---|
| `aoe rm` (CLI) | No — trash-first, worktree relocated to `.aoe-trash/`, restorable | No |
| `aoe rm --purge` | Only with explicit `--delete-worktree` | Only with `--delete-branch` |
| TUI `d` delete dialog | **Yes by default** (`worktree.auto_cleanup` defaults `true`) — `src/tui/dialogs/delete_options.rs:85` | **Yes by default** when `delete_branch_on_cleanup = true` |
| Web/API `DELETE /api/sessions/{id}` | Same defaults — `src/server/api/sessions.rs:727` | Same |
| Trash auto-purge (daemon startup + hourly, `trash_retention_days` default 30) and `aoe session empty-trash` | **Yes, forced** — dirty-worktree protection bypassed (`force_delete: true`, `sessions.rs:4236`, `src/cli/session.rs:626`) | `delete_worktree && delete_branch_on_cleanup` |

`perform_deletion` (`src/session/deletion.rs`) has no "is this the default branch" check anywhere.

## Why the existing safety mechanisms do not help

1. **Worktree lock is not honored.** `GitWorktree::remove_worktree` unlocks before removing (`src/git/worktree.rs:1009`: "aoe locks every worktree it creates; unlock first"). A `git worktree lock` placed by an external tool therefore protects against `git worktree prune` but not against AoE.
2. **Dirty-tree protection is bypassed on the automated paths.** Trash auto-purge and `empty-trash` run with `force_delete: true`, so uncommitted changes in the default-branch worktree do not save it.
3. **Git does not protect the default branch in a bare repo.** Verified experimentally:

   ```
   git init --bare repo.git   # HEAD -> refs/heads/main
   git worktree add wt-main main
   git worktree remove wt-main
   git branch -d main         # succeeds; HEAD now dangles
   ```

   So once AoE removes the worktree, `delete_branch` (`git branch -d`, falling back to `-D` on "not fully merged") deletes `main` itself and leaves the bare repo's `HEAD` pointing at a nonexistent ref.

## Real-world scenario

An external tool manages projects as a bare-repo container: git dir at `<project>/.bare`, default branch checked out at `<project>/main` as a locked worktree, and an AoE session attached to that worktree for tooling/inspection. The worktree is permanent: sibling tooling expects `<project>/main` to exist, and the default branch must never be deleted.

With `delete_branch_on_cleanup = true` in the profile config, a single TUI `d` on that session — or simply trashing it and letting the 30-day auto-purge elapse — destroys both `<project>/main` and the `main` branch. There is no way to mark such a session as "detach AoE, never touch the checkout."

## Proposed fixes (any subset helps)

1. **Default-branch guard (highest value).** When `worktree_info.branch` equals the repo's detected default branch (`detect_default_branch`), refuse worktree removal and branch deletion unless the caller passes an explicit new override (e.g. `--delete-default-branch`). Apply uniformly to CLI, TUI dialog, API body defaults, `empty-trash`, and trash auto-purge.
2. **Per-session protected/pinned flag.** A durable `protected: true` on the instance that blocks purge and worktree/branch teardown (settable via CLI/TUI), so users can mark infrastructure sessions regardless of branch name.
3. **Trash relocation opt-out.** Trashing currently relocates `<project>/main` into `.aoe-trash/<id>` (`src/session/trash.rs`), which already breaks layouts that expect the path to stay stable, even though it is restorable. A per-session or per-repo "do not relocate" option would keep trash-first semantics without moving the checkout.
4. **Docs.** State explicitly that AoE unlocks worktrees before removal, so externally placed `git worktree lock` is not a deletion guard.

## Workarounds in use today

- `worktree.auto_cleanup = false`, `worktree.delete_branch_on_cleanup = false`, `session.trash_retention_days = 0` (0 skips auto-purge), `session.confirm_delete = true` in the profile config.
- Per-repo `<main-repo>/.agent-of-empires/config.toml` with `[worktree] auto_cleanup = false`.
- Discipline: never pass `--delete-worktree` for these sessions; `aoe rm --purge` without it preserves the checkout.

These are opt-out-at-every-level defenses; a single config flip or one `--delete-worktree` re-exposes the default branch. A built-in guard (proposal 1 or 2) would make the safe behavior the default.

Seluj78:

## Description

Every session-deletion path could remove the git worktree holding the repository's default branch and then delete that branch. In a bare-repo layout (git dir at `<project>/.bare`, default branch checked out at `<project>/main` as a linked worktree) that destroys permanent infrastructure and leaves the bare repo's `HEAD` pointing at a ref that no longer exists.

`perform_deletion` gated worktree removal on `managed_by_aoe` alone, and it is the chokepoint for all seven delete paths. The dirty-working-tree skip set was only built when `!force_delete`, so trash auto-purge and `aoe session empty-trash` bypassed the one protection that existed and took uncommitted work with them. The existing safety mechanisms did not help either: AOE unlocks every worktree before removing it, so an externally placed `git worktree lock` is not a guard.

The guard is metadata-only and deliberately conservative, because as a deletion policy a wrong guess is expensive in both directions. It protects a bare repo's own `HEAD`, plus every remote's `refs/remotes/<remote>/HEAD`, and falls back to local `main` / `master` only when neither exists. A non-bare repo's `HEAD` is excluded on purpose: it names whatever that checkout currently has, not the repo's default, so trusting it would protect the user's current feature branch while still missing `main`. The existing `detect_default_branch_info` is not reused, since it picks a branch to base new work on and so scores candidates by commit recency and falls back to the first local branch; used as a guard, those heuristics would refuse to delete ordinary session branches.

Four call sites, not one:

- **Deletion** preserves the worktree, and branch preservation follows from the existing gate that only deletes a branch whose worktree was actually removed (#2532). The refusal is a message, not an error, so the purge still clears the row; an error would make auto-purge retry the same refusal every hour forever and strand the session in the trash. Protected paths join the same set dirty worktrees use, which is what suppresses the in-container `find -delete` preclean and the recursive workspace-dir removal.
- **Trashing** leaves the checkout in place instead of relocating it into `.aoe-trash/<id>`, which broke a layout expecting `<project>/main` to exist and would now strand the checkout in the holding area forever.
- **Reconciliation** moves a checkout that an earlier version already relocated back out of the holding area, so the fix does not convert a recoverable trash row into an orphan on upgrade. It reuses the existing strict restore, so an occupied original leaves the row untouched.
- **`aoe worktree cleanup --force`** treated any worktree with no live session as garbage and removed it with force, so `<project>/main` qualified the moment its session went away. Protected checkouts are now listed as skipped rather than refused outright, since the command still has real orphans to reap.

Reachable without the bare-repo layout too: `aoe add --worktree main` creates a worktree on a pre-existing `main` with `managed_by_aoe: true`, and `WorktreeInfo` has no `branch_preexisting` field, so that branch was treated as AOE's to delete.

Left out on purpose, and worth separate issues: the override flag the issue proposes (with the narrow authority above, the only false-positive source left is the no-metadata `main`/`master` fallback, and the escape hatch is `git worktree remove` plus `git branch -D` then delete the session); `branch_preexisting` on primary `WorktreeInfo`, which needs a tri-state origin and a migration to be correct for existing rows; a per-session `protected` flag; and hardening the workspace-dir `remove_dir_all` against unmanaged descendants.

Closes #3215

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

Set up the reported layout once:

```bash
mkdir -p /tmp/proj && cd /tmp/proj
git init --bare .bare && git --git-dir=.bare symbolic-ref HEAD refs/heads/main
git --git-dir=.bare worktree add /tmp/proj/main main 2>/dev/null \
  || (cd .bare && git commit-tree $(git hash-object -w -t tree /dev/null) -m init | xargs -I{} git update-ref refs/heads/main {} && git worktree add /tmp/proj/main main)
aoe add /tmp/proj/main --cmd claude
```

Then, with `worktree.delete_branch_on_cleanup = true` in your config:

- [ ] Delete that session from the TUI with `d`, cleanup checked. `/tmp/proj/main` still exists, `git --git-dir=/tmp/proj/.bare branch --list main` still prints `main`, and `git --git-dir=/tmp/proj/.bare symbolic-ref HEAD` still resolves. The delete reports "Worktree preserved; 'main' is a default branch of its repository" and the session is gone.
- [ ] Repeat via the web dashboard's `DELETE /api/sessions/{id}` and confirm the same.
- [ ] `aoe rm <session>` (trash, not purge). `/tmp/proj/main` stays exactly where it is; nothing appears under `/tmp/proj/.aoe-trash/`.
- [ ] With the session trashed, run `aoe session empty-trash`. The row leaves the trash, and the checkout plus branch survive.
- [ ] Simulate the upgrade path: `git --git-dir=/tmp/proj/.bare worktree move /tmp/proj/main /tmp/proj/.aoe-trash/<session-id>` on a trashed session, then start the daemon or open the TUI. The checkout moves back to `/tmp/proj/main`.
- [ ] Delete the session, then run `aoe worktree cleanup --force` from `/tmp/proj/main`. It lists `/tmp/proj/main` under "Skipped (default-branch checkouts, never removed)" and does not remove it.
- [ ] Create an ordinary session on a feature branch and delete it with cleanup and branch deletion. The worktree and branch are removed exactly as before, so the guard has not made normal cleanup sticky.
- [ ] In a repo whose `origin/HEAD` names something other than `main` (`git remote set-head origin develop`), a session on a stale `main` is still fully deletable.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the guard lives at the shared chokepoints, and each is covered by a test at the cheapest tier that can actually fail if it breaks. `session::deletion` builds a real bare repo plus linked `main` worktree and asserts a forced delete preserves both and still succeeds; `session::trash` covers relocation and the already-relocated upgrade path; `cli::worktree` covers the orphan split. All four are verified red on the pre-fix tree. An e2e would drive the same functions through tmux for ~2s of permanent serial CI time without being able to fail where these cannot.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. The plan went through a two-model debate, which is what killed the original authority rule (a non-bare repo's `HEAD` is not a statement about the default branch) and surfaced the `aoe worktree cleanup` hole. I made the design calls, including dropping the override flag and keeping the refusal a message rather than an error, reviewed each commit, and ran the manual test steps.

**NOTE:**

## Summary

- Protects default-branch worktrees from deletion, trashing, reconciliation, and forced cleanup.
- Detects protected branches from repository metadata, remote `HEAD` references, or `main`/`master` fallback.
- Preserves protected worktrees while still removing related sessions and trash records.
- Restores default-branch worktrees moved by older versions when possible.
- Adds documentation and tests for bare repositories, deletion flows, cleanup, reconciliation, and feature-branch removal.

## Benefits

This prevents AoE from deleting critical default branches or worktrees. It also keeps cleanup safe while preserving normal feature-branch cleanup behavior.

## Potential Enhancement

Future changes could add more coverage for unusual Git reference layouts and restoration failures.

### DIFF
--- a/docs/guides/worktrees.md
+++ b/docs/guides/worktrees.md
@@ -130,6 +130,16 @@ On the delete side, aoe runs `git submodule deinit -f --all` before `git worktre
 
 Moving a worktree session to the trash (rather than purging it) relocates its worktree out of the active worktree dir into a sibling `.aoe-trash/<session-id>` holding directory via `git worktree move`, so trashed sessions stop cluttering the active checkouts. The worktree stays a live checkout, so previewing a trashed session still works. Restoring the session moves the worktree back to its original path; if that path is now occupied, the restore is refused so nothing is overwritten. Purging a trashed session removes the worktree from the holding dir. Sessions trashed before this behavior existed are relocated the next time the daemon starts or the TUI loads.
 
+### The default branch's checkout is never removed
+
+A bare-repo layout (git dir at `<project>/.bare`, default branch checked out at `<project>/main`) keeps the default branch in a linked worktree that other tooling expects to stay put. AOE therefore refuses to remove that checkout, refuses to delete the branch, and leaves the checkout where it is when the session is trashed. The refusal is reported as a message and the session is still deleted; only the worktree and branch survive.
+
+Detection uses what git itself states: a bare repo's own `HEAD`, plus every remote's `refs/remotes/<remote>/HEAD`. When neither exists (a repo that never ran `git remote set-head`), local `main` and `master` are protected by convention. A repo that explicitly names some other default leaves a branch merely called `main` deletable.
+
+Force does not bypass this, including the forced trash auto-purge and `aoe session empty-trash`. `aoe worktree cleanup` lists such a checkout as skipped rather than removing it. To remove one anyway, do it with git (`git worktree remove`, `git branch -D`) and then delete the session.
+
+Note that an externally placed `git worktree lock` is not a deletion guard: AOE locks every worktree it creates and unlocks before every intentional remove or move, so it unlocks yours too.
+
 ### Template Variables
 
 | Variable | Description |
--- a/src/cli/worktree.rs
+++ b/src/cli/worktree.rs
@@ -5,7 +5,7 @@ use clap::Subcommand;
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 
-use crate::git::GitWorktree;
+use crate::git::{GitWorktree, WorktreeEntry};
 use crate::session::Storage;
 
 #[derive(Subcommand)]
@@ -153,6 +153,33 @@ async fn show_info(profile: &str, identifier: &str) -> Result<()> {
     Ok(())
 }
 
+/// Split a repo's worktrees into the ones `cleanup` may remove and the ones it
+/// must leave alone, dropping the main worktree and anything a session still
+/// points at.
+///
+/// Having no session normally makes a worktree garbage, but not when it holds a
+/// branch git states is the repo's default: in a bare-repo layout that checkout
+/// is the default branch's only working tree, and cleanup removes with force, so
+/// reaping it would destroy infrastructure the moment its session went away
+/// (#3215).
+fn partition_orphaned_worktrees(
+    worktrees: Vec<WorktreeEntry>,
+    main_repo: &Path,
+    tracked_paths: &HashSet<String>,
+    protected_branches: &HashSet<String>,
+) -> (Vec<WorktreeEntry>, Vec<WorktreeEntry>) {
+    worktrees
+        .into_iter()
+        .filter(|wt| {
+            wt.path != main_repo && !tracked_paths.contains(&wt.path.to_string_lossy().to_string())
+        })
+        .partition(|wt| {
+            !wt.branch
+                .as_ref()
+                .is_some_and(|b| protected_branches.contains(b))
+        })
+}
+
 async fn cleanup_orphaned(profile: &str, force: bool) -> Result<()> {
     let storage = Storage::open_unwatched(profile)?;
     let (instances, _groups) = storage.load_with_groups()?;
@@ -177,27 +204,34 @@ async fn cleanup_orphaned(profile: &str, force: bool) -> Result<()> {
     }
 
     // Find worktrees not associated with any session
+    let mut protected_worktrees = Vec::new();
     let current_dir = std::env::current_dir()?;
     if GitWorktree::is_git_repo(&current_dir) {
         let main_repo = GitWorktree::find_main_repo(&current_dir)?;
         let git_wt = GitWorktree::new(main_repo)?;
         let worktrees = git_wt.list_worktrees()?;
+        let tracked: HashSet<String> = instances
+            .iter()
+            .map(|inst| inst.project_path.clone())
+            .collect();
 
-        for wt in worktrees {
-            let is_main = wt.path == git_wt.repo_path;
-            if is_main {
-                continue;
-            }
+        (orphaned_worktrees, protected_worktrees) = partition_orphaned_worktrees(
+            worktrees,
+            &git_wt.repo_path,
+            &tracked,
+            &git_wt.protected_default_branch_names()?,
+        );
+    }
 
-            let wt_path_str = wt.path.to_string_lossy().to_string();
-            let is_tracked = instances
-                .iter()
-                .any(|inst| inst.project_path == wt_path_str);
-
-            if !is_tracked {
-                orphaned_worktrees.push(wt);
-            }
+    if !protected_worktrees.is_empty() {
+        println!("Skipped (default-branch checkouts, never removed):\n");
+        for wt in &protected_worktrees {
+            let unknown = "(unknown)".to_string();
+            let branch = wt.branch.as_ref().unwrap_or(&unknown);
+            println!("  • {}", wt.path.display());
+            println!("    Branch: {}", branch);
         }
+        println!();
     }
 
     if orphaned_sessions.is_empty() && orphaned_worktrees.is_empty() {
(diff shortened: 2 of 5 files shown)


Seluj78:
@hyperlapse122 this implements proposal 1 from your report, plus proposal 3 scoped to the same case and proposal 4's docs note. Would you be able to test it against your actual bare-repo layout?

What landed:

- A default branch's worktree is never removed and the branch is never deleted, on every path: TUI `d`, web `DELETE`, `aoe rm --purge --delete-worktree`, `empty-trash`, and the forced trash auto-purge. `force` does not bypass it.
- Trashing no longer relocates the checkout into `.aoe-trash/<id>`, so `<project>/main` stays where your sibling tooling expects it. If you already have a trashed session whose checkout was relocated by 1.13.2, the next daemon start or TUI load moves it back.
- `aoe worktree cleanup --force` now lists such a checkout as skipped instead of force-removing it. That one was reachable independently of session deletion, so it is worth a look in your setup.
- Docs now state that AOE unlocks worktrees before removal, so your `git worktree lock` is not a deletion guard.

Detection is metadata-only: the bare repo's own `HEAD`, plus every remote's `refs/remotes/<remote>/HEAD`, falling back to local `main`/`master` only when neither exists. Worth confirming that resolves correctly for your `.bare` layout.

Two things I did not do, both easy to revisit if they block you. There is no `--delete-default-branch` override; the escape hatch is `git worktree remove` plus `git branch -D` and then deleting the session. And I did not add the per-session `protected` flag from proposal 2, since it is a durable schema change and the branch-based guard covers the layout you described. If you have infrastructure sessions on non-default branches that also need pinning, say so and that becomes its own issue.

Your workarounds should no longer be load-bearing, but no need to drop them to test.

hyperlapse122:
@Seluj78 Tested `e1417e0` against both a scratch replica of my layout and my actual repos. The guard works everywhere I could reach it.

**Build note (environment, not the PR):** vendored `openssl-src` fails on Fedora because the split `perl` package lacks `FindBin.pm`; built with `OPENSSL_NO_VENDOR=1` against system OpenSSL 3.5.7, rustc 1.97.0. The CodeRabbit finding on `src/git/worktree.rs:439` is a false positive — in git2 0.21 `Reference::symbolic_target()` returns `Result<Option<&str>>`, so `.ok().flatten()` is the correct shape and matches the existing usage at `:1406`. It builds, and the guard tests pass: 83/83 (`default_branch_worktree_survives_a_forced_delete`, the reconcile trio, `partition_orphaned…` included).

**Scratch replica of my layout** (`.bare` with `HEAD -> refs/heads/main`, `origin/HEAD` set, locked `main` worktree — the exact metadata my repos carry):

| Test | Result |
|---|---|
| `aoe rm --purge --delete-worktree --delete-branch --force` on the `main` session | `Worktree preserved; 'main' is a default branch of its repository`, `Branch 'main' kept`, worktree still locked, session row removed |
| `aoe rm` (trash) | checkout stays at `<project>/main`, no `.aoe-trash/` created |
| `aoe session empty-trash` on that row | purged, checkout + branch survive |
| Same worst-case purge on a `feature/x` session | worktree removed, branch deleted — guard is not sticky |
| `aoe worktree cleanup --force` | `<project>/main` listed under "Skipped (default-branch checkouts, never removed)"; a real orphan was still reaped |
| Repo with `HEAD` and `origin/HEAD` → `develop`, session on stale local `main` | fully deletable — the fallback does not over-protect |
| Simulated 1.13.2 upgrade path: trashed row + checkout moved into `.aoe-trash/<id>` by hand | next TUI load moved it back to `<project>/main` and repointed the row |

**Actual repos** (registered in a throwaway debug-build profile, so my real session list was never touched):

- `signet` (bare `HEAD` → `main`): worst-case purge printed the preservation message; worktree, `main` branch, `HEAD`, and the worktree lock all intact. Trash leaves the checkout in place, `aoe session restore` brings the row back, and a purge after restore still preserves everything.
- `telerad-frontend` (bare `HEAD` → `develop`, so the default is not named `main`): cleanup dry-run lists the `develop` worktree under "Skipped" — detection via bare `HEAD` resolves correctly for non-`main` defaults. All my other repos carry the same shape (`HEAD` in `.bare` + `origin/HEAD` both present).

Two notes:

1. To replicate the sessions my bootstrap creates, I had to hand-set `worktree_info.managed_by_aoe: true` in the profile JSON: `aoe add <existing-worktree-path>` records no `worktree_info` at all, so such a session never reaches the guard (it also never reaches any cleanup — fine for my case, since my real sessions are wizard/`-w`-created and carry `managed_by_aoe: true`). Not a PR issue, just how I tested.
2. On the per-session `protected` flag (proposal 2): I don't need it. My infrastructure sessions are exactly the default-branch ones this PR covers; my feature-branch sessions are real work and *should* stay deletable.

LGTM from the reporter's seat — this closes #3215 for my layout. Thanks for the fast turnaround.

PR #3226, opened 2026-08-03
fix(session): self-heal agent_session_id from read-only CLI commands

jerome-benoit:
## Description

Follow-up to #3177 (`fix(cli): capture agent_session_id after CLI launch (#3169)`): closes the residual case that fix could not reach. A capture-deferred agent launched purely through the CLI with **no `aoe serve` daemon and no TUI** can still finish with `agent_session_id = None`, silently breaking `--resume` / resume-after-reboot.

### Root cause (proven)

- **Lazy store write.** opencode writes its session row to its SQLite `opencode.db` **only on the first user turn**, not at process boot. Empirically: launching opencode with no prompt leaves **0 rows** in the `session` table through T+10s; the row appears **within 1s of the first prompt** (`time_updated` is 13-digit ms, same magnitude as `now_ms()`, so no unit/floor mismatch). The per-instance session-id poller started in `finalize_launch` filters opencode sessions by `updated >= launch_time_ms`, so during the bounded launch-time wait added by #3177 (`capture_launched_session_id_blocking`, 8s for `session start`, 2s for `add --launch`) there is nothing to observe. The wait times out with the "did not report a session id in time" note.
- **No CLI self-heal.** `drain_and_persist_session_ids` is driven **only** by the TUI event loop (~500ms) and the `aoe serve` daemon (~2s). No read-only CLI command (`aoe status`, `aoe session show/list`, `aoe ps`) drains the poller or re-runs capture. So once the launch window misses, `None` is permanent until a TUI/daemon happens to observe it. This explains the observed evidence: of 4 CLI-created opencode sessions, 3 were later rescued by a TUI/daemon the user ran, and 1 (`1cc5e7bf`) never was.

Symptom -> cause chain: `finalize_launch` snapshots `launch_time_ms` -> `opencode_poll_fn` -> `filter_agent_sessions` retains `updated >= launch_time_ms` -> nothing matches -> drain finds nothing -> timeout note -> `None` persisted; and no later CLI surface re-drains.

### Fix (targets the cause, no timeout bump)

Add `Instance::self_heal_session_id`, a best-effort backfill called from the read-only CLI commands that already refresh status (`aoe status`, `aoe session show`). It reuses the same `try_retroactive_capture` path the TUI/daemon use at next launch: it reads the tool's store **directly** (needs no live poller) and persists through the guarded `persist_session_to_storage` CAS. So the next time a pure-CLI user inspects the session (after the agent has created its store entry, i.e. after the first prompt) the id is recovered, with no daemon or TUI.

Gated so it can never adopt a wrong id: only a session that owns no id, has a plain `ResumeIntent::Default` (so a user-cleared, pinned, or fork-seeded id is left alone), is not mid-teardown or mid-creation (`Deleting` / `Creating`), is still in the active bucket (an archived or trashed row is a sink a read command must not mutate, and `--no-kill` archiving can leave such a row with a live pane), and has a live tmux session is eligible; a captured id equal to `resume_probe_failed_sid` is rejected. The live-tmux check is the real liveness guard; the `Default` gate is what keeps the #3177 abandoned-sid contract intact **without** the runtime-only `retroactive_capture_excludes` (which is empty on a fresh disk load). Launch-time paths are untouched.

### Tests

- `read_command_backfills_agent_session_id_for_a_lazy_store_write`: a fake codex that writes its rollout only after a first-turn marker reproduces the lazy-store-write gap; `aoe status` must backfill the id the launch window missed. **Fails before this change (`None`), passes after.**
- Three regression locks, each with a live pane and a lingering rollout so only the gate under test stops the write (each verified to go red if its guard is removed):
  - `read_command_self_heal_does_not_readopt_a_cleared_sid`: a `(sid = None, ResumeIntent::Cleared)` row is not re-adopted (the #3177 abandoned-sid contract).
  - `read_command_self_heal_skips_mid_operation_rows`: a `Deleting` row is not touched.
  - `read_command_self_heal_skips_archived_rows`: an archived (sink-bucket) row is not touched.

### Risks

- Read commands do a direct store read only for id-less, live, `Default`-intent, active-bucket sessions; healthy, stopped, mid-op, and archived/trashed sessions pay nothing. Persist goes through the existing `foreign_sid_holder` + CAS + `Use`-pin guards, so a shared-cwd peer's id cannot be stolen. The daemon/TUI path and `--fork-from` are untouched (neither calls this helper).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus (via OpenCode)

**Any Additional AI Details you'd like to share:** AI assisted with the root-cause analysis (multi-agent convergence plus an isolated opencode timing repro), the fix, and the tests, followed by multi-round adversarial review. The author reviewed and validated all changes locally.

## Summary

- Added session ID recovery for CLI-launched agents that create their session store after launch.
- `aoe status` and `aoe session show` now recover and persist missing IDs.
- Recovery avoids ambiguous or ineligible sessions, including cleared, archived, failed, and active-operation sessions.
- Added `agent_session_id` to session JSON output.
- Added shared-directory contention checks and symlink-safe path handling.
- Added end-to-end coverage for delayed stores, session conflicts, lifecycle states, and real OpenCode SQLite recovery.

## Benefits

This improves session tracking without changing launch-time behavior. It also prevents incorrect session attribution when multiple agents share a working directory.

## Potential enhancement

Future work could add user-facing diagnostics when recovery is skipped because session ownership is ambiguous.


njbrake (approved):

Reviewed the fix for lazy session-store backfill (`self_heal_session_id`). Verified the CAS semantics in `persist_session_to_storage` and the canonicalization primitive in `filter_agent_sessions` against source, built the branch, and ran the full unit suite plus the new e2e tests (`cli_session_id_capture`) against a real `aoe` binary, real tmux, and a real opencode SQLite store; all pass. Contention handling (two id-less peers in one cwd) is closed by both the pre-loop snapshot and the `foreign_sid_holder` flock backstop.

Scope is tight and proportional to the residual gap in #3177/#3169, tests are at the right tier, no regressions to existing capture paths.

Issue #3218, opened 2026-08-03
No way to request privileged / cap_add / security_opt for a sandbox container — blocks running a container engine in an agent session


bakeb7j0:
## Summary

There is no way to ask `aoe` to launch a sandbox container with elevated container-creation options — no `privileged`, no `cap_add`, no `security_opt`, no generic passthrough. This blocks running a container engine (rootless podman/buildah) inside an agent session, which is otherwise fully working.

Measured on **aoe 1.11.2**, Linux, docker runtime.

## Reproduction

Add the candidate keys to a profile's `[sandbox]` block, launch a session, and inspect the resulting container:

```bash
# ~/.config/agent-of-empires/profiles/<name>/config.toml
[sandbox]
privileged = true
cap_add = ["SYS_ADMIN"]
security_opt = ["seccomp=unconfined"]

aoe -p <name> add --sandbox --launch "$(mktemp -d)"
docker inspect <container> \
  --format 'Privileged={{.HostConfig.Privileged}} CapAdd={{.HostConfig.CapAdd}} SecurityOpt={{.HostConfig.SecurityOpt}}'
```

Result:

```
Privileged=false  CapAdd=[]  SecurityOpt=[]
```

The keys parse, the file keeps them, and **nothing reaches Docker**. `aoe settings explain sandbox.privileged` reports `not a known setting`, so this reads as "unsupported" rather than "broken" — but a silently-accepted key in the config file is the part worth fixing either way.

## Why it matters

Agents running in aoe sandboxes cannot build or test container images. That is a real functional gap for any project whose deliverable *is* a container image: an agent can edit a Dockerfile but cannot verify it builds, cannot run the result, and cannot scan it.

The capability works perfectly once the container is created with the right options — verified by launching the same image under plain `docker run --privileged` and building end-to-end inside it:

```
podman login docker.io …            Login Succeeded!
podman build -t probe:1 .           <image id>
podman run --rm probe:1 cat /proof  built-in-container
```

Rootless podman, `vfs` storage, agent (non-root) user, no `/dev/fuse`.

Notably, **incremental grants are not enough** — this is not a case where one capability would do:

| flags | outcome |
|---|---|
| default | `cannot clone: Operation not permitted` (seccomp blocks the rootless re-exec) |
| `--security-opt seccomp=unconfined` | `newuidmap: write to uid_map failed` |
[...]

PR #3227, opened 2026-08-03
fix(tmux): honor [tmux] mouse at global and profile scope

### LINKED ISSUE #3207: [tmux] mouse config key is never applied — sessions always get mouse on, and settings explain reports the schema default regardless
## Summary

The `[tmux] mouse` config key is never applied. Sessions are always created with tmux `mouse on`, regardless of what the config says, in any scope. There is no error and the file keeps the value, so it reads exactly like a setting that works.

Measured on **aoe 1.11.2** (`~/.local/bin/aoe-1.11.2`), Linux, tmux 3.x, Wayland/Konsole.

## Reproduction

For each candidate value, set it, create a fresh session, and read tmux's own view of the session option:

```bash
# in ~/.config/agent-of-empires/config.toml
[tmux]
mouse = "off"        # then "disabled", then "never"

aoe -p default add --launch "$(mktemp -d)"
tmux show -t aoe_<Name>_<id> -v mouse
```

| `[tmux] mouse` | scope | new session reports |
|---|---|---|
| `"off"` | global `config.toml` | `mouse=on` |
| `"disabled"` | global `config.toml` | `mouse=on` |
| `"never"` | global `config.toml` | `mouse=on` |
| `"off"` | profile `profiles/<name>/config.toml` | `mouse=on` |

`aoe settings explain tmux.mouse` reports the schema default no matter what the file contains:

```
$ aoe settings explain tmux.mouse
tmux.mouse = "auto"
  source: schema default
  candidates (highest precedence first):
    - schema default: "auto"
```

So the resolver never sees the user value — this looks like `[tmux]` not being wired into the settings registry that `settings explain` reads, rather than a bad enum value on my side. (The binary does contain the string `Failed to apply tmux mouse option:`, so there is an apply path; it just always applies `on` here.)

Existing sessions confirm it too — all 11 on this host report `mouse=on` at the session level, overriding a global `set -g mouse off` in `~/.config/tmux/tmux.conf`, because a tmux session option beats a global one.

## Why it matters

With `mouse on`, tmux captures drag-selection into copy-mode. If the tmux server also has `set-clipboard = external` (a common default), that selection never reaches the system clipboard — so **highlight appears to do something and copy silently goes nowhere**. From the user's side this presents as "copy/paste is broken in aoe sessions", and the config key that looks like it should fix it does nothing.

I spent a considerable amount of time assuming I was using the values wrong before measuring the knob itself.

## Workaround, for anyone who lands here

Neutralise it from tmux instead, on hooks that fire *after* aoe has finished (a `session-created` hook is too early — aoe sets the option after creating the session):

```tmux
set-hook -g client-attached        'set mouse off'
set-hook -g client-session-changed 'set mouse off'
```

## Expected

Either `[tmux] mouse` is honoured, or an unsupported/unwired value is rejected loudly at startup rather than accepted and ignored. Silent acceptance is the expensive failure — it is indistinguishable from a value that works.

## Environment

- aoe 1.11.2
- Linux, Wayland session, Konsole, tmux server with `set-clipboard = external`

njbrake:
## Description

Sessions were always *created* with tmux `mouse on`, regardless of `[tmux] mouse`, with no error and the config file keeping its value, so it read like a setting that worked. Because a tmux *session* option outranks a global one, this silently overrode a user's own `set -g mouse off`; with `set-clipboard = external` on the server that turns drag-selection into copy-mode and copy goes nowhere, which presents as "copy/paste is broken in aoe sessions".

To be precise about what did and did not work before, since the issue title says "never applied": a global `mouse = "disabled"` *did* eventually land on the main agent-session path, because `apply_all_tmux_options` ran the resolver after creation and set `off` (so the session went `on`, then `off`). What genuinely never worked is `auto`'s deferral, since creation's forced `on` outlived a `None`; every profile-scoped value, in any mode; and every mode on the tool-session path in `src/tui/home/mod.rs`, which creates a session with no follow-up `apply_all_tmux_options` at all. That last one is the cleanest argument for moving the decision into creation rather than patching it up afterwards.

Two independent defects were in play. The issue body's own guess (that `[tmux]` is not wired into the settings registry) is not one of them: `TmuxConfig` does derive `SettingsSection`.

**1. Creation forced `mouse on` unconditionally.** `append_mouse_on_args` appended `set-option -t <target> mouse on` with no reference to `config.tmux.mouse`, from all three tmux create paths (`Session`, `PairedTerminal`, `ToolSession`). Meanwhile `should_apply_tmux_mouse()` could return `None` ("do not touch") and `Some(false)`, and nothing ever unset what creation had set. It is now `append_mouse_args(args, target, decision: Option<bool>)`, which emits `mouse on` / `mouse off` / nothing.

**2. Profile-level `[tmux]` overrides were ignored.** `should_apply_tmux_mouse`, `should_apply_tmux_status_bar`, and `should_apply_tmux_clipboard` each called `Config::load_or_warn()` internally, which is the global config only. `merge_configs` supports a `tmux.mouse` override and has tests for it, so the merge worked; no consumer ever read the merged result. The three resolvers now take `&Config`, and every call site resolves through `tmux::tmux_option_config(profile)`, so the session's profile is threaded into the create paths and into `apply_all_tmux_options`.

### The `auto` decision table

`mouse on` is load-bearing for the Web dashboard's touch scroll, which needs tmux copy-mode. So `auto` deliberately does **not** blanket-defer to anyone who merely owns a `tmux.conf`: most such files exist for a prefix key or a theme and never mention `mouse`, and tmux's own default is off, so deferring on mere existence would trade the reported silent breakage for a different silent breakage in a flagship surface. `auto` defers only when the user's config *actually sets* `mouse`:

| `[tmux] mouse` | user's tmux config | result |
|---|---|---|
| `auto` | sets `mouse` | leave the option alone (they win) |
| `auto` | silent on `mouse` | `mouse on` (dashboard scroll preserved) |
| `auto` | no config at all | `mouse on` |
| `enabled` | anything | `mouse on` |
| `disabled` | anything | `mouse off` |

`tmux_config_sets_mouse` is the recognizer: it matches the four `set` aliases (`set`, `set-option`, `setw`, `set-window-option`) with any flag bundle, and requires the option name to match `mouse` and nothing else, so tmux's separate `mouse-select-pane` / `mouse-resize-pane` / `mouse-utf8` family does not count. `tmux_line_commands` does the lexing: one pass per line tracking single and double quotes, ending the line at a token-initial `#` and splitting only on an unquoted `;`. Quote and comment tracking is correctness, not tidiness. Splitting naively on `;` first makes `# set -g status off; set -g mouse on` parse its tail as a live command, so a commented-out line reads as "the user set `mouse`", `auto` steps aside, and the session loses mouse support altogether. That false-positive direction is the damaging one, so it is the direction the recognizer is engineered against and the one the table's most important rows cover.

The recognizer is pure over the file text, so the table test covers realistic lines without touching the filesystem. Its remaining gaps are documented on the function and asserted as table rows so narrowing them stays deliberate: `source-file`, `if-shell '...' 'set -g mouse on'`, a `set` behind a `bind` (the common toggle idiom), and an explicit `-t` target. All four are false *negatives*, so they fail safe to the `mouse on` this PR replaces, and `mouse = "disabled"` is the escape hatch.

### Deferring also clears a stale value

`None` cannot just mean "decline to write". Sessions created by an older aoe carry a session-scoped `mouse on`, and that outranks the global option forever, so the `None` branch of `apply_all_tmux_options` now unsets the option (`set-option -u`). Without it the reporter's 11 existing sessions would look unchanged after upgrading. This mirrors the status-bar unset directly above it, which exists for the same reason.

`status_bar` and `clipboard` still key on mere existence of a tmux config (`user_has_tmux_config`, unchanged). That is a different concern and is not touched here beyond taking the config as a parameter.

### Notes for the reporter

- **`mouse = "off"` is still inert, and so is `"never"`.** The valid values are `auto` / `enabled` / `disabled`; an invalid one fails `Config::load()`. The value you want is `mouse = "disabled"`. This is worth stating plainly because it is the first thing to retry after this lands.
- **Existing sessions need a start or restart** to pick the change up, since `apply_all_tmux_options` runs on session start rather than on attach.

### Deliberately out of scope

Both are pre-existing and both want their own change rather than riding along here:

- **Repo scope.** `tmux` is in `REPO_OVERRIDABLE_SECTIONS` with `AllowAll`, and the settings TUI renders a Repo tab for "Mouse Support", but `tmux_option_config` resolves global + profile via `resolve_config_or_warn`, not `resolve_config_with_repo_or_warn`. So a repo `.agent-of-empires/config.toml` setting `mouse` is still inert, which is the same stranding this PR fixes for profile scope. Wiring the repo layer in moves all three tmux resolvers onto the repo-merged config at once and deserves its own review; hence this PR's title says "global and profile scope" rather than claiming all of them. Filed as #3229.
- **The silent whole-config fallback.** An invalid enum value makes `Config::load()` return `Err`, and `Config::load_or_warn()` swallows it into `Config::default()` behind a `tracing::warn!` that is invisible without debug logging. That is why `aoe settings explain tmux.mouse` reported the schema default: the resolver never saw the file. It also means *any* typo in *any* key silently reverts *every* setting to default, and `settings_schema::resolved::resolve` is separately profile-blind (it reads `Config::load_or_warn()`), so `settings explain` still cannot show a profile override even after this fix. Fixing it properly spans every config consumer (45 `load_or_warn` call sites) and needs a visible TUI surface rather than an `eprintln!` the alt-screen immediately paints over. Overlaps the secondary ask in #3218. Filed as #3228.

### Breaking changes to the lib crate's public API

Calling this out per CLAUDE.md rather than because compatibility is being preserved. Six `pub` items change shape, all internal callers updated in this PR:

- `Session::create`, `Session::create_with_size`, `PairedTerminal::create_with_size`, `ContainerTerminalSession::create_with_size`, and `ToolSession::create_with_size` each gain a required `profile: &str`.
- `tmux::status_bar::apply_all_tmux_options` gains a required `profile: &str`.
- `should_apply_tmux_mouse` / `should_apply_tmux_status_bar` / `should_apply_tmux_clipboard` each gain a required `&Config` and no longer read the global config themselves.
- `TerminalSession::create` is removed. It had no callers, so it is deleted rather than given a new parameter, per the repo's no-dead-code rule.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`docs/guides/configuration.md`'s `mouse` row previously said only "Controls mouse support in aoe tmux sessions"; it now states the `auto` contract, the two paths read, and the detection gaps. `docs/guides/tmux-status-bar.md` also documents `[tmux] mouse` and implied `auto` keyed on tmux.conf existence, so it gets a note on the divergence. The `#[setting]` description on the field (what the TUI and web settings panes render) said "Auto respects your tmux config", which is now misleading, and is reworded. No UI changes, so no screenshot.

On test runs: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, `cargo test --no-fail-fast`, `cargo test --features serve --no-fail-fast`, `cargo doc --no-deps --features serve` with `RUSTDOCFLAGS=-D warnings`, and `cargo xtask gen-docs` (no drift) all pass locally. Two `--features serve` unit tests fail in my container only: `process::worker_registry::tests::pid_source_for_falls_back_to_peer_pid_on_load_err` and `acp::supervisor::tests::shutdown_and_wait_degrades_when_load_errs_without_peer` both force a read error via `chmod 0o000`, which does not stop uid 0, and my sandbox runs as root. Both fail identically on a clean `origin/main` worktree, neither file is in this diff, and `Cargo Test (linux, serve-rest)` was green on CI.

## Test Coverage Analysis

No `web/` files change. `node tests/validate-coverage-matrix.mjs` passes unchanged.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is covered by three cheap unit tests, and an e2e would cost ~2s on the strictly serial critical path forever for no extra failure mode.

Three tests, all unit:

- `tmux::utils::tests::test_mouse_args_follow_configured_mode`: a 6-row table over `{auto, enabled, disabled}` x `{user's config sets mouse, does not}` asserting the emitted tmux args. Verified to fail against the old behavior: making `append_mouse_args` ignore its decision gives `left: [";", "set-option", "-t", "aoe_x", "mouse", "on"], right: []` on the `auto` + user-sets-mouse row.
- `session::config::tests::test_tmux_config_sets_mouse`: a 28-row table over realistic config text. The negative rows are where a sloppy match misfires: `mouse-select-pane`, `mouse-resize-pane`, `mouse-utf8`, a commented-out `# set -g mouse on`, `set -g @plugin 'tmux-better-mouse-mode'`, a `WheelUpPane` bind, the four documented false-negative gaps, and the four `;`-in-a-comment-or-quoted-value cases that a naive split reads as a live `set` (the false-positive direction).
- `session::profile_config::tests::test_tmux_decisions_follow_the_config_they_are_given`: all three resolvers return opposite answers for the global vs profile-merged config. Verified to fail against the old behavior: restoring the internal `Config::load_or_warn()` gives `left: Some(true), right: Some(false)`.

I did write an e2e that measured `tmux show -t <session> -v mouse` end to end across all five scopes; it passed with the fix and failed against each defect separately. Dropped on review, since the unit tests above catch the same regressions and `AGENTS.md` prices a Rust e2e at ~2s on a strictly serial suite.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Claude did the investigation and the implementation. Two things were human calls rather than Claude's:

1. The reconciliation between "the user asked for mouse off" and "the Web dashboard needs mouse on for scrollback". Claude's first pass had `auto` defer whenever a `tmux.conf` merely existed, which @njbrake rejected because it silently breaks mobile scroll for everyone who keeps a tmux config that never mentions `mouse`. The "defer only if their config actually sets `mouse`" rule and the recognizer requirement came from that review.
2. Scoping repo-scope support out rather than expanding this PR to cover it.

The stale-`mouse on`-on-existing-sessions fix, the recognizer's comment and quote handling (a second cold pass caught that a commented-out `; set -g mouse on` read as a live setting, which broke the very invariant the design section claims), the gap documentation, and the stale doc comments all came from review passes by separate agents reading the diff cold rather than from the agent that wrote it.

### On `Fixes #3207`

#3207's title has two halves, and this PR fixes the first (`mouse` is never applied) but not the second (an unsupported value is accepted and `settings explain` reports the schema default). Both follow-ups are now filed, so `Fixes` stands and nothing is dropped when the issue closes:

- #3228 the silent whole-config fallback, plus `settings explain` being profile-blind.
- #3229 repo-level `[tmux]` overrides advertised in settings but never resolved.

Note for the reporter of #3207: the valid values are `auto`, `enabled`, and `disabled`. The literal `mouse = "off"` in that report is not one of them, and until #3228 lands it fails silently, taking the rest of the config file with it.

Fixes #3207

## Summary

- Fixed `[tmux] mouse` handling for global and profile configurations.
- Applied `mouse on`, `mouse off`, or no session override based on the resolved configuration.
- Preserved dashboard scrolling in `auto` mode unless the user explicitly configures tmux mouse behavior.
- Applied profile-specific tmux settings to sessions, status bars, tools, containers, and agents.
- Unified tmux setting resolution for status bar, mouse, and clipboard options.
- Improved tmux configuration detection, including XDG paths and option-specific parsing.
- Added regression tests for configuration parsing, mouse behavior, and profile resolution.
- Updated tmux documentation.
- Removed the unused `TerminalSession::create` convenience method.

## Benefits

Users can control tmux behavior without unwanted session-level overrides. Profile settings now apply consistently across session types. More precise configuration detection reduces false deferrals and stale session settings.


jerome-benoit (commented):
The `mouse` fix here is correct and well-reasoned (the session-scope override of a user's `set -g mouse off` was a real bug, and the `auto` = defer-only-when-the-user-sets-`mouse` semantics is the right call). Approving the fix on its own terms.

One structural point worth raising, because this PR half-closes it: the tmux settings-application logic is **not consolidated**, and after this PR the three `[tmux]` options diverge in two ways.

**1. Deferral granularity is inconsistent** (`src/session/config.rs`):
- `mouse` (auto) now defers **per-option**: `TmuxMouseMode::Auto => (!user_sets_mouse).then_some(true)` (`:2477`) via `user_tmux_config_sets_mouse()` (`:2428`), which parses the user's tmux config for the `mouse` option specifically.
- `clipboard` (auto) still defers **coarsely**: `TmuxClipboardMode::Auto => !user_has_tmux_config()` (`:2491`).
- `status_bar` (auto) same coarse gate: `TmuxStatusBarMode::Auto => !user_has_tmux_config()` (`:2390`).

So a user whose `tmux.conf` sets a prefix key but never touches clipboard still loses aoe's `set-clipboard` passthrough, while the same user now *does* get aoe's `mouse on` (the whole point of this PR). `clipboard` maps to a single tmux option (`set-clipboard`, `append_clipboard_passthrough_args`, `utils.rs:202`), so it is the exact same shape as `mouse` and could use the same per-option probe. This is the same latent inconsistency `mouse` had *before* this PR, now left in place for `clipboard`.

**2. Tri-state modeling diverges**: `should_apply_tmux_mouse` returns `Option<bool>` (on / off / leave-alone), while `should_apply_tmux_status_bar` / `should_apply_tmux_clipboard` return plain `bool` (apply / skip, with no distinct "force off"). Three create paths (`session.rs`, `terminal_session.rs`, `tool_session.rs`) each call the three `should_apply_tmux_*` helpers independently.

**Suggested consolidation (single source of truth).** Rather than three ad-hoc helpers plus three call sites, describe each aoe-managed tmux setting once in a table, and drive resolution + application from it:

```rust
// One row per tmux option aoe manages on its OWN sessions. Single source of truth:
// the config mode, the option(s)+value aoe sets, and the rule that makes `auto` defer.
struct TmuxManagedSetting {
    key: &'static str,                         // for docs/logs
    mode: TmuxTriMode,                          // Auto | Enabled | Disabled (resolved from config)
    apply: &'static [(&'static str, &'static str)], // e.g. [("mouse","on")], [("set-clipboard","on")]
    defer: DeferRule,
}
enum DeferRule {
    WhenUserSets(&'static str),   // fine: defer only if the user's tmux config sets THIS option
    WhenUserHasAnyConfig,         // coarse: for a holistic theme, not a single option
}
```

- Generalize `tmux_config_sets_mouse` into `tmux_config_sets(option: &str)` and back `WhenUserSets` with it.
- `mouse` -> `WhenUserSets("mouse")`, `clipboard` -> `WhenUserSets("set-clipboard")` (this closes gap #1), both with uniform tri-state resolution.
- `status_bar` -> `WhenUserHasAnyConfig` is **defensible and can stay**: it applies a whole theme (`status`, `status-left/right`, `status-style`, ... `status_bar.rs:35-73`), not one option, so a per-option probe is fuzzy. Just document that it is coarse by design.
- One resolver (`mode + defer -> apply / skip / force-off`) and one applier that iterates the table at the shared create path, so adding a future managed option is one row, not a new helper plus three call-site edits.

Not blocking this PR (the mouse fix stands on its own); flagging it as a follow-up so the per-option-defer idea this PR introduces for `mouse` becomes the transverse rule rather than a one-off. Happy to take the consolidation as a separate PR if you prefer.

Issue #2858, opened 2026-07-15
bug(session): same-cwd clobbering still recurs on v1.12.1 — incl. cross-directory inheritance from a worktree session with unnormalized project_path (recurrence of #2708)


simon-spektrum:
**Version:** aoe 1.12.1 (latest as of 2026-07-15) · macOS · concurrent TUI + `aoe serve` daemon · ~16 sessions, several sharing the repo-root `project_path`

#2708 was closed by #2709 (released in v1.12.1), but the clobbering continues on 1.12.1. Two kinds of evidence:

**1. Ongoing store churn.** `debug.log` contains **8,585** `session.store: Session ID changed` events. This morning alone (2026-07-15), two different sessions were assigned the *same* conversation UUID within 3 minutes:

```
2026-07-15T05:56:28.595640Z  INFO session.store: Session ID changed for 275e69fef5454567: d2c54d89-56fb-4f09-a373-4222d0bd4ac0
2026-07-15T05:59:22.675842Z  INFO session.store: Session ID changed for e2787df686d240d7: d2c54d89-56fb-4f09-a373-4222d0bd4ac0
```

**2. New failure shape: cross-directory inheritance.** In #2708 worktree sessions appeared immune — attribution is by cwd. Today a **repo-root** session (`verify-labs-2468-schema-validation`) was found holding the conversation UUID belonging to a **worktree** session ("Journeys not displaying"): opening the root session's tab resumed the worktree session's conversation. Possibly relevant: that worktree session's stored `project_path` is **unnormalized** — `…/Spektrum.nosync/../Spektrum.nosync-worktrees/journeys-not-displaying` — which may defeat cwd-based attribution of status-hook reports.

User-visible impact is the same as #2708: tabs open the wrong conversation's history, and `set-session-id` pins on stopped sessions get overwritten by live same-dir peers.

Both deferred follow-ups called out in #2709 remain plausible causes for the store race:
- no storage flock for concurrent TUI + daemon drains
- the poller's pre-CAS tmux publish

Happy to provide a fuller `debug.log` window or run an instrumented build.

njbrake:
Hi! Would you be willing to investigate and contribute a PR fix?

tomdcc:
Recurrence on **v1.13.1** (i.e. with #2921 in), this time crossing a **profile** boundary.

Two sessions sharing `project_path` `/Users/tom/scratch`:

| profile | session | `CLAUDE_CONFIG_DIR` | own conversation |
[...]

Issue #2653, opened 2026-07-04
git::worktree::branch_exists swallows spawn errors as false (same class as #2596)


jerome-benoit:
Split out from the review of #2652 (fix for #2596). Same swallowing-existence-probe class of bug, on the git surface rather than the containers surface — flagged as follow-up so #2652's scope stays contained.

## The bug

[`src/git/worktree.rs:1069`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/worktree.rs#L1069):

```rust
Err(_) => false,   // spawn failure, git not on PATH, or I/O error → "branch absent"
```

`branch_exists()` shells out to `git show-ref --verify --quiet refs/heads/<branch>` and maps every `run_git` error (git binary missing, I/O error, permission denied) to `false`. Callers use `branch_exists()` as a guard before branch creation/rename operations. If git is transiently unavailable, a branch that exists is reported absent, and a caller could attempt to create a duplicate branch or skip a rename guard.

Same shape as #2596 was for containers (`is_running().unwrap_or(false)` on a probe that gates a mutation). Trigger probability is low (git missing from PATH is unusual) but non-zero, and the class is worth closing everywhere.

## Fix shape

Mirror the pattern #2652 landed for containers: introduce a tri-state (or `Option<bool>` — the caller API decides which) that preserves the difference between "definitively absent" and "check failed". Fail closed at callers — if we can't confirm the branch is absent, refuse the mutation and surface the git error to the user.

`GitError::WorktreeCommandFailed(String)` already exists in `src/git/error.rs` and is the right error variant to propagate.

## Anchors (state at time of filing)

- Target fn: [`pub fn branch_exists(&self, name: &str) -> bool`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/git/worktree.rs#L1069)
- Callers to audit for correctness under the new signature: grep `.branch_exists(` in `src/session/`, `src/git/`.
- Reference precedent: #2652 (`DockerContainer::probe_running`, `RuntimeBase::classify_inspect_failure`).

## Scope note

Explicitly out of scope for #2652 (which fixed the containers-runtime surface of this class). This issue tracks the git-runtime surface. Same class, separate surface, separate PR.

PR #3230, opened 2026-08-03
fix(session): attribute OMP captures to the launching pane

jerome-benoit:
## Description

Attribute each AoE session's OMP (oh-my-pi) conversation to its exact tmux pane, and serialize the session lifecycle that the fix exposed as fragile. Covers concurrent sessions in the same repository, sibling worktrees, host launches, and Docker launches.

**OMP capture attribution**
- Resolve OMP's effective `HOME`, session store, project path, and launch options once, then pin them into the launched process and typed tmux metadata.
- Attribute captures through a per-pane terminal breadcrumb plus a launch marker, instead of the newest file under a shared cwd.
- Require three proofs before adopting a breadcrumb: the marker's launch-id and routing-fingerprint CAS anchors, and post-launch freshness (host: `modified_at_ms > launched_at_ms`; container: the breadcrumb is `-nt` the launch marker). The marker alone is necessary but not sufficient, since a stale pre-launch breadcrumb also differs from the pending sentinel; freshness is what proves this pane authored it.
- Canonicalize worktree and project paths at resolution so sibling worktrees of the same repository no longer cross-attribute; enforce live-instance exclusions, validate UUIDs and headers, reject stale or cross-project candidates; keep legacy panes recoverable through a guarded fallback that cannot cross a newer launch generation.
- Persist SID updates with generation-aware compare-and-set in the CLI, TUI, and daemon; retry after transient write failures; repair a stopped poller after draining its final observation, including concurrent restart races.

**Session lifecycle serialization**
- Unify the two ad-hoc ownership primitives (per-op claim and launch generation) into a single durable `LifecycleReservation` (op + generation + timestamp), acquired under the instance flock with a generation compare-and-set and a TTL self-heal.
- Release the flock while user launch hooks run, since a hook may re-invoke `aoe` for the same session; the durable reservation, not the flock, provides cross-process exclusion during that window.
- Tear down forcibly removed sessions and keep purge, restore, and trash mutually exclusive through the reservation's generation-checked commit.

**Sandbox and shell hardening**
- Transport launch environment (including secrets) through a protected, one-shot, mode-0600 env file consumed by the pane, never through tmux argv or `ps`; the container path reads it via `--env-file /dev/fd/9` so host env vars no longer leak in.
- Bound and `O_NOFOLLOW`-guard host and container file reads, canonicalize-and-recheck the store before opening, validate env keys, and honor the configured Docker runtime.

This fixes the ambiguity identified in #2858. It grew beyond the original narrow capture fix because that fix exposed the session's ownership primitives (per-op claim and launch generation) as unsafe under concurrency, so the lifecycle itself had to be serialized (see Session lifecycle serialization above).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (see Verification)
- [x] Documentation was updated where necessary: no `docs/` contract changed; user-facing purge/restore refusal messages were harmonized across CLI, server, and TUI
- [x] For UI changes: no visual UI change, screenshot not applicable

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A

e2e (files under `tests/e2e/`): `cli_session_id_capture.rs` covers OMP capture for isolated same-cwd panes across a host restart generation, and prelaunch-reject then cross-project reconstruction with legacy backfill; `kiro_launch.rs` covers the chat launch, yolo trust-all-tools, and agent-hook install commands via a recording stub; `purge_restore_race.rs` and `archive_restore.rs` cover reservation-gated purge/restore races and archive teardown. Unit: store resolution, breadcrumb and marker parsing and freshness (host and container), canonical worktree paths, lifecycle-reservation CAS and generation guards, runtime-reload field ownership, CAS persistence and retry delivery, concurrent restart merges, poller repair.

## Verification

- `cargo fmt --all --check`: clean.
- `cargo clippy --features serve --lib --bins -- -D warnings`: clean.
- `cargo test --features serve --lib`: the `session::` tree (which holds the capture, sync, tmux, and tui suites) passes.
- CI `Cargo Test (linux, serve-e2e)` is green on this commit. The serve-e2e regressions this rework first introduced were root-caused and fixed in-branch; the e2e files listed above exercise them (host restart-generation and prelaunch/cross-project OMP attribution, the kiro launch commands, and the purge/restore reservation races).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** OpenAI Codex in Oh My Pi

Used for repository investigation, implementation, test construction, concurrency and security review, and verification. Multiple independent reviewer agents cross-validated race handling, filesystem boundaries, the lifecycle-reservation CAS, and host/container OMP parity.


src/session/environment.rs
```diff
@@ -900,25 +736,16 @@ pub(crate) fn build_docker_env_args(
...
-                docker_flag_parts.push(format!("-e {}", key));
-                exports.push(format!("export {}={}", key, shell_escape(value)));
-            }
-            EnvEntry::Literal { key, value } => {
-                // Non-secret literal values are safe in argv
-                docker_flag_parts.push(format!("-e {}={}", key, shell_escape(value)));
-            }
-        }
+        docker_flag_parts.push(format!("-e {}", entry.key()));
+        env.push((entry.key().to_string(), entry.value().to_string()));
```
jerome-benoit: **[P1] Keep container-only values out of the host wrapper**

`sandbox.environment` is repository-overridable, but this now moves every resolved entry, including literal values, into the pane environment. `Instance::build_launch_command` maps these pairs to `PaneEnvMutation::Set`, and `EphemeralEnvFile` emits the exports before an unqualified host-side `rm` and the Docker/Podman launch. Because the pane cwd is the repository, a repo containing `[sandbox] environment = ["PATH=."]` plus an executable `rm` runs that file on the host before isolation. Loader and runtime variables expose the same trust-boundary problem. I reproduced the generated script shape: exporting `PATH=.` before `rm -f` executed a repo-local `rm` and then continued the launch. Preserve the Literal/Inherit distinction or transport these values directly to the container runtime without installing them in the outer host shell, and add a regression case proving repo-derived `PATH` is visible only in-container.

src/session/instance.rs
```diff
@@ -3207,100 +3788,246 @@ impl Instance {
...
+        let profile = self.effective_profile();
+        let storage = super::storage::Storage::new(&profile, self.resolve_file_watch())
+            .context("failed to open lifecycle lock storage")?;
 
-        let session = self.tmux_session()?;
-
-        if session.exists() {
+        let _lifecycle_lock = storage
+            .acquire_instance_lifecycle_lock(&self.id)
+            .context("failed to acquire instance launch lock")?;
```
jerome-benoit, line 3930: **[P2] Release the lifecycle flock while launch hooks run**

This guard remains live through `prepare_launch`, which synchronously waits for `before_session` and `on_launch` hooks. Those hooks receive `AOE_SESSION_ID`, so a supported hook such as `aoe session set-session-id "$AOE_SESSION_ID" ...` re-enters `acquire_instance_lifecycle_lock`: the child polls this flock forever while the parent waits for the child. The launch stays in `Starting` with its durable reservation. Keep the durable generation reservation, drop the flock for arbitrary user hooks, then reacquire it and validate that exact generation before spawning the pane. The same lock-lifetime defect also exists in `PurgeTransaction::run_hooks`.

src/session/deletion.rs
```diff
@@ -26,12 +30,488 @@ pub struct DeletionRequest {
...
+            was_trashed,
+            generation,
+            lifecycle_lock: Some(lifecycle_lock),
+            active: true,
+        }))
+    }
+
+    /// Run best-effort hooks without a lifecycle or storage flock held.
+    pub fn run_hooks(self) -> Self {
+        run_on_destroy_hooks(&self.request.instance, self.request.detach_hooks);
```
jerome-benoit: **[P2] Do not retain the purge flock across `on_destroy`**

`reserve` stores the acquired flock in `self.lifecycle_lock`, so this method still owns it while `run_on_destroy_hooks` synchronously waits for every hook. A same-session AoE lifecycle command from `on_destroy` then waits for the parent's flock while the parent waits for the hook, producing an unbounded deadlock. `detach_hooks` only detaches the TTY, `run_hooks_best_effort` still calls `Command::output()`. This also contradicts the method comment that says no lifecycle flock is held. Release the flock for the hook phase, then reacquire it and generation-check the Purge reservation before teardown or irreversible commit.

jerome-benoit (commented):
Blocking review findings. GitHub does not allow requesting changes on one's own pull request. The OMP attribution and generation-CAS paths held up under independent review, but the surrounding launch/lifecycle refactor introduces three blockers:

- P1: repository-controlled sandbox environment values are exported into the host pane before container launch, allowing `PATH` or loader/runtime variables to cross the isolation boundary.
- P2: launch holds the non-reentrant per-instance flock while synchronously waiting for `before_session` and `on_launch` hooks.
- P2: purge holds the same flock while synchronously waiting for `on_destroy` hooks.

The two hook findings are the same design hazard on separate code paths and both require an unlocked hook phase followed by flock reacquisition and exact-generation validation. The security path was reproduced with the generated script ordering; the flock block was reproduced independently and cross-validated against each AoE call path. No additional proven defect survived cross-validation.

src/session/instance.rs
```diff
@@ -1974,76 +2125,91 @@ impl Instance {
...
-    /// TTL for an `OpClaim`. Longer than any realistic teardown or worktree
-    /// move so a live operation is never overridden mid-flight, short enough
-    /// that a crash mid-operation self-heals promptly (the next purge/restore
-    /// overrides the expired claim, and the load-time reconcile clears it). See
-    /// #2541.
-    pub const OP_CLAIM_TTL: chrono::Duration = chrono::Duration::minutes(10);
+    /// Longer than any bounded hook, teardown, or worktree move. A crashed
+    /// owner cannot retain the reservation forever; a late owner is still
+    /// harmless because every commit is generation-checked.
+    pub const LIFECYCLE_RESERVATION_TTL: chrono::Duration = chrono::Duration::minutes(10);
```
jerome-benoit, line 2241: [P1] This reservation can expire while a hook is still running. The launch/restart paths deliberately release the flock before `run_pre_launch_hooks` (`3819-3826` and `5896-5908`), and purge does the same in `deletion.rs:189-203`. But ordinary hooks have no deadline: `repo_config.rs:1068-1081` takes `timeout = None` then `command.output()`, the timeout being installed only during startup recovery (`repo_config.rs:27-32`). After 10 minutes another process can therefore acquire a new generation and run the same hook effects concurrently; the first owner only discovers the loss after those effects. We should renew the reservation by CAS during the hook, or enforce and cleanly kill a production timeout strictly below the TTL, then test two processes crossing the TTL.

src/git/worktree.rs
```diff
@@ -1136,66 +1138,86 @@ impl GitWorktree {
...
-
-                if !force_output.status.success() {
-                    let force_stderr = String::from_utf8_lossy(&force_output.stderr);
-                    tracing::debug!(target: "git.worktree",
+            // Resolve worktree ownership from Git's stable porcelain output,
+            // not from localized stderr. A missing checkout with a retained,
+            // locked admin entry is ours to reap; a live checkout must remain
+            // untouched.
+            if let Some(path) = self.worktree_path_for_branch(branch)? {
+                if !used_by_worktree_retried && !path.exists() {
```
jerome-benoit: [P1] `Path::exists()` also returns `false` when `stat` fails, notably on `EACCES` or an I/O error; this check therefore does not prove the checkout is gone. The following path unlocks then prunes the Git entry and retries the branch deletion, which can detach a worktree that is still present but temporarily inaccessible and delete its branch. Use `path.try_exists()` and only run the reap on `Ok(false)`; any error must fail closed and preserve the entry.

src/server/api/sessions.rs
```diff
@@ -3916,156 +3816,152 @@ async fn purge_session_artifacts(
...
-                    target: "acp.supervisor",
-                    session = %id,
-                    "shutdown during purge failed: {e}"
-                );
+    let deletion_result = if transcript_purged {
+        // Commit the row removal before deleting the ACP transcript. A lost
+        // restore/generation race therefore leaves both row and transcript
+        // intact; a successful commit makes later cleanup failures
+        // non-restorable by construction.
+        let committed = tokio::task::spawn_blocking(move || transaction.begin_irreversible())
```
jerome-benoit, line 3895: [P1] This commit already removes the durable row (`deletion.rs:305-325`) before the `await shutdown_and_delete` and before `CommittedPurge::finish`, which alone cleans up the worktree, branch, container, and scratch (`deletion.rs:483-489`). A crash or SIGKILL in that window therefore leaves these sidecars permanently orphaned: the session no longer exists in `sessions.json`, so no CLI/TUI/API rerun can resume the purge. The reconciler recovers the orphaned ACP worker, but no analogous mechanism replays these sidecars. Keep a durable tombstone/outbox carrying the `DeletionRequest` until `finish`, replayed at startup, or keep a `Deleting` row until cleanup completes.

src/session/instance.rs
```diff
@@ -3195,6 +3651,137 @@ impl Instance {
...
+        let committed = storage.update(|instances, _groups| {
+            let Some(stored) = instances.iter_mut().find(|instance| instance.id == self.id) else {
+                return Ok(false);
+            };
+            if !stored.lifecycle_reservation_is_owned(LifecycleOperation::Launch, generation) {
+                return Ok(false);
+            }
+            stored.status = self.status;
+            stored.idle_entered_at = self.idle_entered_at;
+            stored.last_accessed_at = self.last_accessed_at;
```
jerome-benoit, line 3791: [P2] The `Launch` reservation does not grant exclusivity over these fields, yet this commit overwrites them from the snapshot taken before the spawn. As soon as the pane becomes visible, `aoe send --no-revive` can send it a message and then persist `touch_last_accessed()` and `Status::Running` under the storage flock (`src/cli/send.rs:73-89`) without touching the reservation. This block therefore always finds its reservation valid, then rewrites the old timestamp and `Starting`, durably losing the accepted activity. Merge `last_accessed_at` by maximum and only overwrite the presentation fields when their value still matches the baseline read at launch (CAS), rather than treating them as implicitly owned by `Launch`.

jerome-benoit (commented):
Complete initial review, redone from scratch on the exact diff `86ccdd16ee19c91a91bbbd5a01fe061e494903e4...bbd88e66278ed0de0679dbdd062e2066701a572f` (38 files), with four independent analyses followed by local cross-validation.

Technical verdict: **CHANGES_REQUESTED**. GitHub strictly forbids `REQUEST_CHANGES` on one's own PR, so this review is published as `COMMENT`. Four introduced and proven defects are commented inline: 3 P1 and 1 P2. The candidate points on archive, out-of-lifecycle-protocol worktree mutations, and truncated tmux identity were excluded from this review because their windows already existed in the exact base, even though they remain architectural risks.

No tests were rerun: static review only, as requested. The evidence is the execution paths and exact lines cited in each finding.

src/session/capture.rs
```diff
@@ -527,6 +564,13 @@ pub(crate) fn capture_omp_session_id(
 /// Shared pi-family session scan. `default_subdir` is the host home directory
 /// used when `PI_CODING_AGENT_DIR` is unset (`.pi/agent` for pi, `.omp/agent`
 /// for omp).
+///
+/// The primary lookup joins the sessions dir with [`encode_pi_project_path`]
+/// (pi's `--<abspath>--` form); for omp it misses (see
+/// [`capture_omp_session_id`]), so the cwd-matching fallback below locates the
+/// session. The final newest-directory fallback is reached only when no
+/// header's recorded cwd matches the project path, so a well-formed omp session
+/// is resolved by the cwd fallback, not by that project-agnostic heuristic.
```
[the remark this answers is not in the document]
jerome-benoit: Correction to my earlier reply: the fix landed in `dab828d1` ("fix(session): error instead of cross-project resume"), not the review range I first cited. When every CWD-matching candidate is excluded, resolution now returns an error (`anyhow::bail!`) instead of falling through to the project-agnostic newest-directory fallback, so a session from another project can no longer be resumed. Regression test: `test_capture_omp_all_cwd_matches_excluded_errs`.

src/session/capture.rs
```diff
@@ -3278,6 +3389,105 @@ mod tests {
...
+        };
+
+        // omp-shaped directory names: `<scope>-<basename>-<64-hex-sha256>`. The
+        // hash value is irrelevant to the cwd fallback (it scans every dir); the
+        // point is that it never equals `encode_pi_project_path`, so the primary
+        // lookup misses exactly as it does for real omp.
+        let sha = "0".repeat(64);
+        // Target written first (older dir mtime).
+        write_session(&format!("home-target-project-{sha}"), target_cwd, target_id);
+        std::thread::sleep(std::time::Duration::from_millis(50));
```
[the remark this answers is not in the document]
jerome-benoit: Correction to my earlier reply: the fix landed in `dab828d1` ("deterministic test mtimes"). The fixtures now set mtimes explicitly via a `set_mtime` helper (fixed UNIX timestamps) instead of relying on `sleep`, so candidate ordering is deterministic across filesystems.

tests/e2e/cli_session_id_capture.rs
```diff
@@ -635,3 +635,183 @@ fn read_command_backfills_agent_session_id_for_opencode_sqlite_path() {
...
+         terminal_dir={terminals}\n\
+         mkdir -p \"$sessions_dir\" \"$terminal_dir\"\n\
+         tty_path=$(tty) || exit 1\n\
+         terminal_id=$(printf '%s' \"${{tty_path#/dev/}}\" | tr '/' '-')\n\
+         timestamp=$(date -u +%Y-%m-%dT%H-%M-%S-%3NZ)\n\
+         first_path=\"$sessions_dir/${{timestamp}}_{first}.jsonl\"\n\
+         printf '%s\\n%s\\nfresh\\n' {cwd} \"$first_path\" > \"$terminal_dir/$terminal_id\"\n\
+         sleep 4\n\
+         printf '{{\"type\":\"session\",\"id\":\"{second}\",\"cwd\":\"%s\"}}\\n' {cwd} > {historical}\n\
+         printf '%s\\n%s\\n' {cwd} {historical} > \"$terminal_dir/$terminal_id\"\n\
```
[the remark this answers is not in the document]
jerome-benoit: Obsolete at HEAD. `install_switching_fake_omp` and the fixed `sleep 4` no longer exist. The current fakes (`install_toggling_fake_omp`, `install_reconstructing_fake_omp`) synchronize via test-controlled control files (`while [ ! -f "$control/release-$slot" ]; do sleep 0.05; done`, signalled by `ready-$slot` / `switched-$slot`), which is exactly the marker gate this comment proposed.

src/session/environment.rs
jerome-benoit: Fixed by the sandbox env-file rework. On the sandbox path `build_launch_command` builds `LaunchEnvironment { pane: Vec::new(), container: env_info.env }`, so repo `sandbox.environment` values go only into the container channel: a 0600 env-file consumed inside the container via `--env-file /dev/fd/9`, never exported into the host pane shell or tmux argv. The host `rm -f --` runs under the host PATH, not a repo-controlled one. Regression test: `test_container_env_file_does_not_mutate_host_process_environment`.

src/session/instance.rs
jerome-benoit: Fixed. The instance flock is now dropped before `run_pre_launch_hooks` (`instance.rs:3822`) and reacquired afterward, with a post-hook `ensure_reservation_current_or_fail`; cross-process exclusion during the hook window is provided by the durable generation reservation, not the flock, so a hook that re-invokes `aoe` on the same session no longer deadlocks. Locked by `launch_hooks_run_without_the_instance_lifecycle_flock`.

src/session/deletion.rs
jerome-benoit: Fixed. `PurgeTransaction::run_hooks` sets `self.lifecycle_lock = None` before `run_on_destroy_hooks`, then reacquires the flock and generation-checks the Purge reservation before teardown. Locked by `on_destroy_hooks_run_without_the_instance_lifecycle_flock`.

src/session/instance.rs
jerome-benoit: Confirmed and tracked in #3297. The behavior stays fail-closed: if the reservation is stolen after the TTL, the launch fails cleanly at `ensure_reservation_current_or_fail` (`instance.rs:3832`), with no double spawn and no deadlock. So no fix in this PR; the issue proposes either a production timeout strictly below the TTL (configurable, so a legitimately long hook is not killed) or a CAS heartbeat re-stamping `lifecycle_reservation.at`.

src/git/worktree.rs
jerome-benoit: Fixed in b9e04f2c. The reap no longer uses `!path.exists()` (which also returns `false` on EACCES/I/O) but a fail-closed tri-state probe `Self::checkout_confirmed_absent(path.try_exists())`: reap only on `Ok(false)`, fail closed on any `stat` error (a present-but-inaccessible worktree is preserved, its branch not deleted), mirroring the `branch_exists` rule (#2653). Deterministic test added.

src/server/api/sessions.rs
jerome-benoit: Confirmed and tracked in #3298. Real but bounded: structured path only (the non-structured path tears down before removing the row), crash-only, flock held end-to-end within a single process, no user-repo data loss. The fix (durable tombstone/outbox + replay at startup, or a `Deleting` row until `finish`) is a durable-state change, so it goes through `src/migrations/`.

src/session/instance.rs
jerome-benoit: After re-checking, I classify this as a false positive: `commit_lifecycle_launch` (`instance.rs:3660-3675`) is CAS-guarded; it writes `status`/`last_accessed_at` only when `lifecycle_reservation_is_owned(Launch, generation)` holds. A concurrent writer (`aoe send`) bumps the generation, so this commit is skipped (`ensure!(committed, ...)`), and passive pollers go through `merge_passive_status_patch` with a monotonic `last_accessed_at`. Reopen if there is a window I missed.

jerome-benoit:
Re: the earlier CodeRabbit outside-diff comment on `maybe_start_poller_since` ("tolerate startup recovery without an OMP watermark"). Obsolete at HEAD `b9e04f2c`, no change needed:

- `AOE_SESSION_STARTED_AT_MS` no longer exists anywhere in the tree, so the launch-watermark gating that finding assumed is gone.
- `maybe_start_poller_since` does not gate on a startup watermark. It gates on `supports_session_poller()` and, for `omp`, re-resolves the metadata via `omp_capture_metadata(...)` when none is passed in (`src/session/instance.rs:5311`).
- `omp_capture_metadata` reads the persisted hidden tmux env `AOE_OMP_CAPTURE_META_KEY` (`src/session/instance.rs:2589`), which survives an `aoe` restart because it lives with the tmux session, with a legacy fallback via the tmux session `created_at_ms`.

So recovery/startup restarts the poller without any launch watermark. The only case it does not is a dead tmux session, where there is nothing to poll anyway.

njbrake (commented):
> Drafted by Claude via back-and-forth with @njbrake.

The implementation is carefully defended and CI is green after fixing the parallel test hang. The attribution path fails closed, generation checks protect stale lifecycle commits, and the focused tmux, purge, restore, environment transport, and lifecycle tests pass.

This remains a high-blast-radius change across capture, launch, restart, stop, trash, restore, purge, and cleanup. The main residual risks are a live destruction hook outlasting the ten-minute reservation TTL, incomplete external cleanup if a process crashes after the durable session row is removed, and portability gaps in custom container images or runtimes.

Rollout plan: merge this immediately after the weekly release on Wednesday, August 12. That gives us a full week to smoke test it on `main`, watch for stuck lifecycle reservations, duplicate cleanup hooks, orphaned resources, incorrect OMP attribution, and container launch failures before the next weekly release on Wednesday, August 19.

njbrake (approved):
Here we gooo 🚀  thank you for the work on this, it's ok as far as I can tell but of course with a change this big some smoke testing from the build on main should help to smooth out any sharp edges 🙏

PR #3245, opened 2026-08-04
chore(plugins): pin the featured hash for GitHub plugin 2.0.0

Seluj78:
## Description

Adds the featured-index pin for [plugin-github 2.0.0](https://github.com/agent-of-empires/plugin-github/pull/87), the first release built on the `api_version = 12` pane block vocabulary from #3244.

`agent-of-empires.github` is a reserved namespace, so a release is installable only while the featured index vouches for its exact source tree (`reject_reserved_or_builtin` at install, and its load-time twin in `registry::validation_for`). Without this pin, plugin-github 2.0.0 resolves as featured-unverified and the install gate refuses it outright.

The pin does not make 2.0.0 loadable on its own. Its manifest declares `aoe_version = ">=1.14.0, <2.0.0"`, so `host_compat` skips it on 1.13.x with a load error rather than failing startup. **It becomes installable only once the aoe release that carries api_version 12 ships, and that release must be a minor bump** (`open-release-pr.yml` defaults to patch, which would cut 1.13.3 and leave the floor unmet).

### How the hash was produced

```
git archive v2.0.0 | tar -x -C "$tmp"
aoe plugin hash "$tmp"
```

`git archive` is used rather than a working checkout so the tree holds exactly the tracked files, with no venv, no `.aoe-build/`, and no stray artifacts. The procedure was validated by reproducing the already-shipped 1.8.0 pin (`sha256:5f8d13ed…`) byte for byte before being used for this one.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

A pin is data, and the code that consumes it is already covered by `plugin::featured`'s tests plus the install/load gates in `plugin::install` and `plugin::registry`. A test asserting this specific hash string would restate the declaration and fail only when someone deliberately edited both, which `AGENTS.md` calls out as a test not worth adding.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Hash computed locally and cross-checked against the shipped 1.8.0 pin to validate the procedure.

## Summary

- Added the featured-index hash for `agent-of-empires.github` plugin version `2.0.0`.
- Documented its `api_version = 12` requirement and `aoe_version >=1.14.0` compatibility floor.
- Preserved safe loading behavior on older AoE versions.

## Benefits

- Enables installation validation for plugin version `2.0.0`.
- Prevents loading on unsupported AoE versions.
- Introduces no user-facing or CLI/TUI behavior changes.

## Potential follow-up

- Enable the plugin after the AoE release that supports API version 12 becomes available.

### DIFF
--- a/plugins/featured.toml
+++ b/plugins/featured.toml
@@ -32,7 +32,13 @@ source = "gh:agent-of-empires/plugin-github"
 # namespace gate skips it). Dropping the pin downgrades it to community so the
 # reserved-namespace install gate blocks it outright. Fixed in 1.2.0, which
 # builds under `.aoe-build/`.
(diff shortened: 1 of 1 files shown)


Seluj78:
Ordering note, since this PR carries a hash for a tag that does not exist yet.

`v2.0.0` is not tagged: [plugin-github#87](https://github.com/agent-of-empires/plugin-github/pull/87) is the release-staging PR and is green but gated on a review. The hash here was computed from the staging branch tree (`release-staging/v2.0.0` at `4146d26`), whose parent is the current `plugin-github` `main` (`3ceb33c`). Because main has not advanced past that parent, the merge tree is identical to the staging tree under squash, merge-commit, and rebase alike, so the hash is deterministic rather than a guess.

The one way it goes stale: something else lands on `plugin-github` `main` before #87 merges, which changes the merge tree and therefore the hash. So:

1. merge plugin-github#87 (tags `v2.0.0`, publishes the release)
2. re-verify: `git archive v2.0.0 | tar -x -C "$t" && aoe plugin hash "$t"` should print `sha256:f5343edc97e1a7f0892064abc07645f7f4d96b6badbd508be3b5ede53b2ac4e7`
3. merge this PR
4. cut the aoe release as a **minor** bump (`gh workflow run open-release-pr.yml -f bump=minor`), so the version satisfies the plugin's `>=1.14.0` floor

A wrong pin fails quietly: the plugin resolves as featured-unverified, and the reserved-namespace gate refuses to install it. Worth doing step 2 before merging.

PR #3249, opened 2026-08-04
fix(session): quote a shell-active --model value in extra_args at launch

BTForIT:
## What

A model id carrying shell metacharacters aborts the launch line before the agent starts. `extra_args` is spliced into the launch command verbatim, so a context-window suffix does exactly that:

```
aoe add --tool claude      # session extra_args: --model claude-x[1m]
aoe session start <id>
zsh:1: no matches found: claude-x[1m]
```

The pane exits 1 at launch. zsh reads `[1m]` as a character-class glob, matches no file, and aborts the whole line; bash with `failglob` behaves the same way. Nothing in the output points at the model id, so it presents as "the session just dies".

## Why this shape

Only the `--model` / `-m` **value** is quoted, in both the spaced and joined forms. The rest of `extra_args` is the caller's own argv, where shell syntax may well be intended, so it is left untouched. An ordinary model id is not quoted at all, so existing command lines keep their current shape, and an already-quoted value is not re-quoted (that would nest and hand the agent a model id containing literal quotes).

## How it was tested

`cargo test --lib` green (4402 passed). Five unit tests cover both flag forms, the untouched ordinary case, other arguments preserved, no nesting, and a dangling flag.

Also verified end to end against a real pane: with the fix the pane launches and the agent receives the value intact; without it the pane is dead at launch with the glob error above.

## Summary

- Adds shell-safe quoting for model values in `extra_args` during default host and sandboxed launches.
- Supports spaced and joined `--model` and `-m` forms.
- Preserves unrelated arguments, whitespace, ordinary model IDs, and existing quotes.
- Adds regression tests for model quoting and edge cases.
- Prevents shell expansion from interrupting session launches while preserving custom command behavior.

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -2650,6 +2650,113 @@ pub fn get_telemetry_settings() -> TelemetryConfig {
     Config::load_or_warn().telemetry
 }
 
+/// Wrap a value so the launching shell passes it through literally.
+///
+/// Only quotes when the value contains something a shell would interpret, so
+/// ordinary model ids keep their current, unquoted command line.
+fn shell_quote_value(value: &str) -> String {
+    const SAFE: &str = "-_./:=@,+";
+    let plain = !value.is_empty()
+        && value
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || SAFE.contains(c));
+    if plain {
+        return value.to_string();
+    }
+    // Already quoted by whoever wrote it: re-quoting would nest, and the agent
+    // would receive a model id with literal quote characters in it.
+    if value.len() >= 2 && value.starts_with('\'') && value.ends_with('\'') {
+        return value.to_string();
+    }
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Real bug, right fix: an unquoted `--model`/`-m` value with shell metacharacters (e.g. a `[1m]` context-window suffix) aborts the whole launch line under zsh glob expansion.

Two gaps in the original patch, now fixed in the latest commit:

1. The quoting used `split_whitespace()`/`join(" ")`, which isn't shell-safe: it double-nested quotes on an already-quoted value and collapsed unrelated whitespace elsewhere in `extra_args`. Replaced with a byte-preserving scanner that only rewrites the `--model`/`-m` value span.
2. The fix only covered the host launch path; the sandboxed launch path had the identical bug. Now covered too.

CI is green across the board. Approving.

PR #3257, opened 2026-08-05
feat(session): add host_hooks.before_session to mint host agent env at spawn

hairbui76:
### What I am trying to do

I run several Claude accounts and I want to use aoe as the single place I manage work across all of them: a handful of sessions in flight at once, each billed to a different account, all visible in one session list and drivable from the web dashboard on my phone.

The reason is quota. One subscription's 5-hour and weekly windows put a hard ceiling on how much parallel agent work I can have running, and aoe is very good at encouraging parallel work — worktrees, groups, structured view, push notifications when something wants my input. I hit the ceiling long before I run out of things to hand an agent. Spreading sessions across accounts lifts it, and it also keeps client work and personal work billed separately.

aoe is already the right control plane for this. The one thing missing is that **a session cannot say which account it runs as**.

### Why the existing knobs do not get there

The Host Environment docs suggest pinning an account by path, which does work:

```toml
environment = ["CLAUDE_CONFIG_DIR=/Users/me/.claude-accounts/work"]
```

But that string is static and scoped to a whole profile, which means:

- **One aoe profile per account.** With a couple of accounts and a couple of relay providers I am pre-creating and hand-maintaining every combination, and switching means switching profile rather than just picking an account.
- **No per-directory answer.** A repo cannot run as the account that owns it; `environment` has no repo or session layer (deliberately, and I am not proposing to change that).
- **Nothing can refresh a token.** Switchers keep per-account tokens that rotate. A config string is never *called*, so nothing gets a chance to refresh one before the agent starts, and stale-token failures show up as a confusing agent error.

`host_hooks.before_start` is the right shape for this problem, but it only fires when a **sandbox container** comes up — and it cannot help here even in principle, because `CLAUDE_CONFIG_DIR` is a host path that deliberately does not cross the container boundary. Host sessions have no equivalent.

### What this adds

The missing quadrant: a command that runs per **host** launch and whose `KEY=VALUE` stdout becomes the agent's environment.

|  | static | minted at spawn |
| --- | --- | --- |
| sandbox | `sandbox.environment` | `host_hooks.before_start` |
| host | `environment` | `host_hooks.before_session` **(new)** |

A launch runs exactly one of the two, chosen by `is_sandboxed()`, so a key is never minted twice.

With it, an account switcher answers per launch instead of being hardcoded:

```toml
[host_hooks]
before_session = ['cc env --format plain --if-bound']
```

```text
CLAUDE_CONFIG_DIR=/home/me/.claude-accounts/profiles/work
ANTHROPIC_BASE_URL=http://127.0.0.1:8317
```

One line of config now covers every account. The hook runs with cwd set to the session's project and receives `AOE_PROJECT_PATH`, so it can resolve **per directory** — which is what makes "this repo is client work, that one is personal" express itself without a profile per account. And because it runs on every launch, including restart, a switcher gets its chance to refresh a rotated token first; when an account's window is exhausted I rebind and restart, and the session resumes on the other account because sessions are shared.

Nothing about the hook is switcher-specific — it is just a command that prints environment variables. `cc env` above is the switcher I happen to use; a three-line shell script that echoes a path works exactly as well.

Same stdout contract as `before_start`: `KEY=VALUE` lines collected, others ignored, stdout never logged (it is the documented secret channel), non-zero exit aborts the launch. Nothing is persisted between launches, so a short-lived value is refreshed rather than replayed.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No UI change: this is config plumbing with no TUI or dashboard surface.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

`tests/e2e/host_before_session_e2e.rs`, modelled on the existing `acp_host_environment_e2e.rs`: the proof is read off the adapter's own environment, captured by the shim immediately before it execs the fake agent, so it is downstream of the daemon's `env_clear` + allowlist and downstream of the detached runner.

It discriminates three ways on one key. `CODEX_HOME` has a wrong ambient value on the daemon, a wrong value in the static `environment` list, and the expected value from the hook:

- forwarding the daemon's ambient env stays RED;
- applying minted pairs **before** the static list — the way `before_start` does for the first-wins container list — stays RED, which is what pins the opposite precedence rule.

`GIT_CONFIG_GLOBAL` is minted with no static entry and no ambient value, so it doubles as the oracle for *which* shim invocation was the structured worker (only a process that ran the mint can carry it). The `CODEX_HOME` comparison is then a real assertion on that invocation, and names in its failure message which of the two losing values won — so a red test says "precedence is backwards" or "the mint never landed" rather than just "didn't see what I wanted".

The hook also prints `AOE_TOKEN` on purpose, and every capture is swept to prove it never reaches an adapter.

Unit tests (13) cover the resolver, the runner's stdout contract and error labelling, `drop_shadowed_host_entries` (empty-mint identity, all three entry forms, whole-key-only matching so a minted `FOO` does not drop `FOOBAR`), `host_hooks` parsing, and that a repo cannot inject `before_session`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 via Claude Code, used for the implementation, tests, and docs.

Worth flagging two things that surfaced during verification rather than while writing, since they are the parts most worth a reviewer's scrutiny:

1. The terminal-view precedence inversion described above. The first cut passed minted values via `tmux -e` and left the static `environment` prefix intact, which silently let stale config win. `drop_shadowed_host_entries` and the e2e discrimination exist because of that.
2. The sandbox gate was initially `sandbox_info.is_some()` while `build_launch_command` dispatches on `is_sandboxed()`. An instance carrying disabled `SandboxInfo` builds a host command, so it would have silently skipped minting.
3. The mint block initially ran a `spawn_blocking` hop and a config read on every structured spawn, because the `is_empty()` check sat inside the closure. It surfaced as the timing-sensitive `reset_*` tests in `acp_client` failing intermittently — one per run, a different test each time — while `upstream/main` was clean over a full run. Gated on the already-resolved config; two consecutive full runs are green.

Verified on Linux: fmt clean, clippy no new warnings, 6011 lib tests pass over two consecutive full runs, e2e compiles. Not verified on macOS.

## Summary

- Added `host_hooks.before_session` for non-sandboxed launches.
- Runs the hook for each host launch, including restarts.
- Passes generated environment values to terminal, tmux, and structured-view sessions.
- Gives generated values priority over static environment values.
- Blocks repository-level configuration from injecting host commands.
- Keeps hook output private and stops initial launches when hooks fail.
- Keeps sandboxed launches on `before_start`.
- Added documentation, tmux support, unit tests, and end-to-end coverage.

## Benefits

This provides controlled host-session setup across all supported launch types. It refreshes environment values for each launch and protects sandboxed sessions and sensitive hook output.

## Technical notes

- Added `HostHooksConfig::before_session`.
- Added shared hook execution and configuration resolution helpers.
- Added `drop_shadowed_host_entries`.
- Added `Session::create_with_size_env`.
- Added separate logical tool tracking in ACP spawn configuration.
- Added coverage for environment precedence, forwarding, secret filtering, and sandbox gating.

## Potential enhancement

Future changes could add more targeted diagnostics for hook configuration and launch failures without exposing hook stdout.


hairbui76:
Btw, I have skipped the issue-first step in CONTRIBUTING and went straight to a PR, sorry about that. Happy to move the discussion to an issue if you'd rather scope the approach before reviewing code.

PR #3261, opened 2026-08-05
fix(session): ignore Codex subagent rollouts during capture

jona-wilmsmann:

**This implementation and pull request were created with OpenAI Codex. The code has not been human-reviewed.**

We are making it available because it is the local patch we are currently using for a session-mismatch problem, and it may be of interest to the maintainers. We do not claim that this patch is perfect, complete, or the preferred upstream design. Please treat it as a tested candidate fix and reproduction rather than as an assertion that this is necessarily the right final implementation.

## Description

AOE currently discovers a Codex conversation by recursively scanning rollout files, selecting the newest file whose recorded CWD matches the session project, and using the UUID from that rollout filename.

Codex also persists Guardian/subagent rollouts under the same sessions directory. Those child rollouts record the same CWD as their parent and can have a newer mtime, but their filename UUID is a child thread rather than the top-level user conversation. AOE can therefore adopt the child UUID and persist it on an unrelated same-CWD session, causing tabs to resume mismatched history.

This patch:

- keeps the rollout filename UUID authoritative because that is the ID Codex resumes;
- validates every present `payload.session_id` and `payload.id` independently, rejecting the rollout if either is invalid or differs from the filename UUID;
- remains compatible with older rollout metadata and capture fixtures that omit `type` and both ID fields;
- applies the same identity check to host and container capture;
- adds deterministic host and container regressions where a newer child rollout points `session_id` at its parent while sharing the parent's CWD.

Related context: #2858. This is a distinct Codex child-rollout attribution case rather than a claim to address every same-CWD race.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] New and existing tests pass (all affected tests and the 6,003-test library suite pass; one unrelated tmux tool-session e2e timed out, including when rerun alone)
- [x] Documentation was updated where necessary (no user-facing documentation change was needed)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped an e2e test, because: the capture decision is exercised deterministically with host and container unit fixtures; there is no UI rendering or CLI interaction change.

Commands run on current upstream `main` (`d702762`):

- `cargo +1.95.0 test codex --lib` (107 passed)
- `cargo +1.95.0 test --features e2e-tests --test e2e cli_session_ -- --test-threads=1` (4 passed, including the two feature-gated cases identified in review)
- `cargo +1.95.0 clippy -- -D warnings`
- `cargo +1.95.0 test --features serve,e2e-tests --quiet` (6,001 library tests passed, 2 ignored; subsequent integration groups passed until the unrelated `tool_sessions::test_tool_session_full_attach_and_cleanup_roundtrip` timed out waiting for its tmux marker; the same test timed out earlier when rerun alone)

## AI Usage

- [ ] No AI was used
- [x] AI was used

Codex investigated the local failure, produced the patch and regression tests, rebased it onto current upstream `main`, and ran the checks listed above. The patch has not been human-reviewed. It is being submitted because it is the local workaround currently in use and may provide a useful reproduction or starting point. It should not be treated as a claim that the implementation is perfect or that maintainers should merge it unchanged.

## Summary

- Updated Codex session capture to use the rollout filename UUID as the authoritative identity.
- Added validation for `payload.session_id` and `payload.id`.
- Applied the validation to host and container capture.
- Removed subagent marker and `session_meta` type checks to support minimal fixtures.
- Added deterministic regression fixtures and table-driven parser tests.

## Benefits

- Prevents child rollouts from being captured as separate sessions.
- Keeps host and container capture behavior consistent.
- Improves compatibility with different rollout metadata shapes.

## Potential enhancement

- Add coverage for future rollout metadata fields and additional `session_meta` formats.


src/session/capture.rs
```diff
@@ -1815,17 +1821,56 @@ pub(crate) fn capture_codex_session_id(
...
+/// own UUID in the filename, but are not independently resumable user
+/// conversations. Reject them before the host or container scanner ranks
+/// candidates. Prefer Codex's canonical session id for accepted roots, with
+/// the filename UUID as a compatibility fallback for older metadata.
+fn parse_codex_session_meta(line: &str, filename_uuid: &str) -> Option<CodexSessionMeta> {
     let parsed: serde_json::Value = serde_json::from_str(line).ok()?;
-    parsed
-        .get("payload")
-        .and_then(|p| p.get("cwd"))
+    if parsed.get("type")?.as_str()? != "session_meta" {
```
Seluj78: Blocker: this new gate breaks two existing e2e tests.

`tests/e2e/cli_session_id_capture.rs:57` installs a fake `codex` whose rollout first line is `{"payload":{"cwd":"%s"}}`, with no `type` field. The old `parse_codex_cwd_from_json` accepted it; this returns `None`, so capture finds nothing and `agent_session_id` stays `null`.

Failing checks: `Cargo Test (linux, serve-e2e)` and the `Cargo Test (linux)` umbrella leg, both with `left: None`. `cargo test --quiet` does not build the `e2e-tests` feature, which is why it passed locally; the full command is `cargo test --features serve,e2e-tests`.

All 302 real rollouts I surveyed do carry `"type":"session_meta"`, so the gate is accurate about production data. It is the test fixture it breaks, and the gate is not needed once identity is bound to the filename UUID (see the comment on line 1855).

src/session/capture.rs
```diff
@@ -1815,17 +1821,56 @@ pub(crate) fn capture_codex_session_id(
...
+        return None;
+    }
+
+    let cwd = payload
+        .get("cwd")
         .and_then(|v| v.as_str())
-        .map(|s| s.to_string())
+        .filter(|cwd| !cwd.is_empty())?
+        .to_string();
+    let session_id = payload
```
Seluj78: Required: this remaps session identity with no consistency check.

Whatever UUID the payload declares becomes the value persisted to `sessions.json` and later handed to `codex resume`. Across the 302 real rollouts under my `~/.codex/sessions`, `payload.id` equals the filename UUID in 302/302, and `session_id` equals the same UUID in all 225 that carry it. So the remap is a no-op on every observed file, and on a divergent file it returns an id that names no rollout on disk, with no cited evidence that `codex resume` resolves it. It also changes `exclusion` semantics: that set is keyed on ids written by the old filename-UUID path.

Suggested inversion, which also subsumes this PR's goal: when `session_id` or `id` is present, require equality with the filename UUID and reject on mismatch, then return the filename UUID. That rejects this PR's own child fixture (`session_id` = root, filename = child) without depending on any marker field.

src/session/capture.rs
```diff
@@ -1815,17 +1821,56 @@ pub(crate) fn capture_codex_session_id(
...
+        .get("session_id")
+        .or_else(|| payload.get("id"))
```
Seluj78: Required: `or_else` short-circuits before the UUID validation below, so a present-but-malformed `session_id` discards a valid `payload.id`.

Given `{"session_id":"corrupt","id":"<valid uuid>"}` the chain takes `"corrupt"`, fails `.filter`, and falls through to `filename_uuid`; `payload.id` is never consulted. This is the open CodeRabbit thread, and its `find`-over-both-keys form is the right fix.

src/session/capture.rs
```diff
@@ -1815,17 +1821,56 @@ pub(crate) fn capture_codex_session_id(
...
-    parsed
-        .get("payload")
-        .and_then(|p| p.get("cwd"))
+    if parsed.get("type")?.as_str()? != "session_meta" {
+        return None;
+    }
+
+    let payload = parsed.get("payload")?;
+    let has_parent = payload
+        .get("parent_thread_id")
```
Seluj78: Required: this discriminator keys off fields that appear in zero observed rollouts.

I surveyed the 302 rollout files under my `~/.codex/sessions`: 0/302 contain a `parent_thread_id` key at all, `thread_source` is only ever `"user"` (224) or `null` (78), and `source` is only ever `"exec"`, `"vscode"`, `"cli"`, `"unknown"`. No Codex version, changelog, or captured child rollout is cited. If the version producing the bug names these differently, the guard is inert and the newest-mtime child still wins, which is the pre-PR behavior.

Two secondary defects in the same block: the doc comment above claims Guardian rollouts are rejected, but nothing here matches `"guardian"`; and `!value.is_null()` treats `"parent_thread_id": ""`, `false`, or `0` as a parent, false-rejecting a root and leaving its session uncapturable.

If these checks stay as defense in depth, please include a real captured child rollout (Codex version plus first line) so the field names are verifiable.

src/session/capture.rs
```diff
@@ -1815,17 +1821,56 @@ pub(crate) fn capture_codex_session_id(
...
+        .unwrap_or(filename_uuid)
+        .to_string();
+
+    Some(CodexSessionMeta { cwd, session_id })
+}
+
+fn codex_source_is_subagent(source: &serde_json::Value) -> bool {
+    match source {
+        serde_json::Value::String(value) => value.to_ascii_lowercase().starts_with("subagent"),
+        serde_json::Value::Object(fields) => fields
```
Seluj78: Required: this branch matches object *keys* while the string branch above matches *values*.

So `{"type":"subagent"}`, the common internally-tagged serde shape, is accepted as a root, while a root carrying any field named `subagent_*` (for example a future `subagent_capable: false`) is false-rejected and its session becomes uncapturable. No observed Codex version encodes `source` or `thread_source` as an object at all. Either match values in both branches or drop the object branch.

src/session/capture.rs
```diff
@@ -4041,6 +4113,52 @@ mod tests {
...
+            format!(
+                r#"{{"type":"session_meta","payload":{{"id":"{root_uuid}","session_id":"{root_uuid}","cwd":"{}","parent_thread_id":null,"thread_source":"user"}}}}"#,
+                project_dir.display()
+            ),
+        )
+        .unwrap();
+        std::fs::write(
+            &child_file,
+            format!(
+                r#"{{"type":"session_meta","payload":{{"id":"{child_uuid}","session_id":"{root_uuid}","cwd":"{}","parent_thread_id":"{root_uuid}","thread_source":"subagent"}}}}"#,
```
Seluj78: Required: this test passes with the entire marker filter deleted.

The child fixture sets `session_id` to the root UUID, and `parse_codex_session_meta` prefers `session_id`. Remove `has_parent` and `is_subagent_source` and the newer child is still accepted, still resolves to `root_uuid`, and `assert_eq!(result.unwrap(), root_uuid)` still holds. So the test cannot fail if the feature under test is removed.

The production shape that matters is untested: a child with no marker fields, or one whose `session_id` equals its own filename UUID. `test_parse_codex_session_meta_rejects_subagent` does cover the marker path at unit level, so the gap is specific to the scanner tests.

src/session/capture.rs
```diff
@@ -4099,6 +4217,26 @@ mod tests {
...
+    fn test_select_codex_session_in_container_ignores_newer_subagent() {
+        let root_uuid = "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa";
+        let child_uuid = "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb";
+        let stdout = format!(
+            "\
+===CODEX:1700001000:rollout-2025-01-02T00-00-00-{root_uuid}.jsonl===
+{{\"type\":\"session_meta\",\"payload\":{{\"id\":\"{root_uuid}\",\"session_id\":\"{root_uuid}\",\"cwd\":\"/workspace\",\"parent_thread_id\":null,\"thread_source\":\"user\"}}}}
+===END===
+===CODEX:1700002000:rollout-2025-01-02T00-01-00-{child_uuid}.jsonl===
+{{\"type\":\"session_meta\",\"payload\":{{\"id\":\"{child_uuid}\",\"session_id\":\"{root_uuid}\",\"cwd\":\"/workspace\",\"parent_thread_id\":\"{root_uuid}\",\"thread_source\":\"subagent\"}}}}
```
Seluj78: Required: same tautology as the host test. The child block sets `session_id` to the root UUID, so `select_codex_session_in_container` returns `root_uuid` even with the marker filter removed, and the assertion still passes.

src/session/capture.rs
```diff
@@ -1762,6 +1762,12 @@ find "$SESS_DIR" -name '*.jsonl' -type f | while read -r f; do
 done
 "#;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
```
Seluj78: Nit: `Clone` is never used on `CodexSessionMeta`. `AGENTS.md` bans dead code, including unused derives.

src/session/capture.rs
```diff
@@ -1918,19 +1963,20 @@ fn select_codex_session_in_container(
...
+            Some(u) => u,
             _ => continue,
         };
         let json_part = match rest.split_once("\n===END===") {
             Some((j, _)) => j,
             None => rest,
         };
-        let cwd = match parse_codex_cwd_from_json(json_part.trim()) {
-            Some(c) => c,
+        let meta = match parse_codex_session_meta(json_part.trim(), &filename_uuid) {
```
Seluj78: Nit: `Some(meta) if guard => meta, None => continue, Some(_) => continue` is correct as written but reads as a reorder hazard. `Some(m) if !exclusion.contains(&m.session_id) => m, _ => continue` is equivalent and harder to break.

src/session/capture.rs
```diff
@@ -3929,23 +3975,49 @@ mod tests {
     }
 
     #[test]
-    fn test_parse_codex_cwd_from_json() {
-        let line = r#"{"type":"session_meta","payload":{"cwd":"/home/user/myproject"}}"#;
+    fn test_parse_codex_session_meta_root() {
```
Seluj78: Nit: these four `test_parse_codex_session_meta_*` functions are one behavior with four inputs. House style per `AGENTS.md` ("One test per behavior, not per input") is a single test with a case table, since test-fn count drives compile time across six Rust CI jobs.

Seluj78 (changes requested):

## Review summary

Reviewed head `fb023a99` against `main@d702762`. One file changed (`src/session/capture.rs`, +166/-28).

Thanks for the honest framing in the description; treating this as a tested candidate rather than a proposed final design is the right call, and the reproduction is useful. The problem is real. The request below is to shrink the patch rather than to drop it.

Two required CI checks are currently red, and the cause is in the diff.

Grounding note for the findings: I surveyed the 302 real rollout files under `~/.codex/sessions` on my machine. Results:

- 302/302 carry top-level `"type":"session_meta"`.
- 0/302 contain a `parent_thread_id` key at all.
- `thread_source` is only ever `"user"` (224) or `null` (78).
- `source` is only ever `"exec"`, `"vscode"`, `"cli"`, `"unknown"`.
- 302/302 have `payload.id` equal to the filename UUID; all 225 that also carry `session_id` have it equal to the same UUID; 77 have no `session_id` field.

I also ran this diff past five independent models (Claude Opus 4.7, GPT-5.6, Gemini 3.1 Pro, Grok 4.5, GPT-5.5). All five converged on the same two conclusions: the marker sniffing is unsupported by the observed schema, and the `session_id` remap should be an equality check rather than an override. Findings below were each re-verified against the diff before being reported.

### Blocker

**1. The new `type` gate breaks two existing e2e tests.** `src/session/capture.rs:1834`

`tests/e2e/cli_session_id_capture.rs:57` installs a fake `codex` whose rollout first line is `{"payload":{"cwd":"%s"}}`, with no `type` field. The old `parse_codex_cwd_from_json` accepted that; the new parser returns `None`, so capture finds nothing and `agent_session_id` stays `null`. Both `cli_session_start_persists_agent_session_id_without_daemon` and `cli_session_restart_persists_new_agent_session_id_without_daemon` fail with `left: None`.

Failing checks: `Cargo Test (linux, serve-e2e)` and the `Cargo Test (linux)` umbrella leg. The evidence list in the description is `cargo test --quiet`, which does not build the `e2e-tests` feature, so this leg was never exercised locally. Per `AGENTS.md`, the full command is `cargo test --features serve,e2e-tests`.

### Required

**2. Payload-declared session id replaces the filename UUID with no consistency check.** `src/session/capture.rs:1855-1861`

Whatever UUID the payload declares becomes the value persisted to `sessions.json` and later handed to `codex resume`. On all 302 observed rollouts the remap is a no-op, and on a divergent file it returns an id that names no rollout on disk, with no cited evidence that `codex resume` resolves it. It also quietly changes `exclusion` semantics: that set is keyed on ids written by the old filename-UUID path.

**3. `or_else` short-circuits before UUID validation, discarding a valid `payload.id`.** `src/session/capture.rs:1856-1857`

`or_else` fires only when `session_id` is absent. Given `{"session_id":"corrupt","id":"<valid uuid>"}`, the chain takes `"corrupt"`, fails `.filter`, and lands on `filename_uuid`, never consulting the valid `id`. This is the open CodeRabbit thread, and the suggested `find`-over-both-keys form fixes it.

**4. The subagent discriminator keys off fields absent from every observed rollout.** `src/session/capture.rs:1836-1843`

See the survey above: `parent_thread_id` never appears, `thread_source` never says subagent, `source` carries a launch channel rather than a thread kind. No Codex version, changelog, or captured child rollout is cited. If the version producing the bug names these differently, the guard is inert and the newest-mtime child still wins, which is pre-PR behavior. Two secondary defects in the same block: the doc comment at `:1826` claims Guardian rollouts are rejected but nothing matches `"guardian"`, and `!value.is_null()` treats `"parent_thread_id": ""`, `false`, or `0` as a parent, false-rejecting a root.

**5. Both new scanner tests pass with the entire marker filter deleted.** `src/session/capture.rs:4118` and `src/session/capture.rs:4221`

The child fixtures set `session_id` to the root UUID, and the parser prefers `session_id`. Remove `has_parent` and `is_subagent_source` and the newer child is still accepted, still yields `root_uuid`, and both assertions still hold. Neither test can fail if the feature under test is removed. The production shape that matters, a child with no markers or with `session_id` equal to its own filename UUID, is untested. The unit test `test_parse_codex_session_meta_rejects_subagent` does cover the marker path directly, so the gap is specific to the two scanner tests.

**6. `codex_source_is_subagent` matches values for strings but keys for objects.** `src/session/capture.rs:1866-1874`

The asymmetry means `{"type":"subagent"}`, the common internally-tagged serde shape, is accepted as a root, while a root carrying any field named `subagent_*` is false-rejected and its session becomes uncapturable. No observed Codex version encodes these fields as an object at all.

### Nit

- `src/session/capture.rs:1765`: `Clone` on `CodexSessionMeta` is never used (`AGENTS.md` bans dead code).
- `src/session/capture.rs:1974`: `Some(meta) if guard => meta, None => continue, Some(_) => continue` is correct but reads as a reorder hazard; `Some(m) if guard => m, _ => continue` is equivalent.
- `src/session/capture.rs:3978-4020`: four separate `test_parse_codex_session_meta_*` functions where the house style is one test with a case table (`AGENTS.md`, "One test per behavior, not per input").

### Suggested smaller shape

One discriminator covers the reported case without guessing at the schema: when `payload.session_id` or `payload.id` is present, require it to equal the filename UUID and reject on mismatch; keep returning the filename UUID.

- It rejects this PR's own child fixture, whose `session_id` is the root while its filename is the child.
- It holds for 302/302 observed rollouts, including the 77 with no `session_id`.
- It needs no marker fields, so it survives a Codex rename.
- It leaves the resume value bound to a file that exists on disk.
- It lets the `type` gate be dropped, which unbreaks the e2e fixture.

If the marker checks are kept as defense in depth behind that, please cite a real captured child rollout (Codex version plus its first line) so the field names are verifiable, and add a scanner test for a child with no markers.

### Verdict

Implementation: Request Changes. Project fit: Accept with scope reduction. Placement: n/a, this is a bug fix rather than a feature.

jona-wilmsmann:
Thanks for the detailed review. I revised this with Codex to use the smaller filename-consistency design you suggested:

- the rollout filename UUID remains authoritative;
- every present `payload.session_id` and `payload.id` is validated independently and must match that filename UUID;
- the `type == "session_meta"` gate and marker/source sniffing are gone, so the existing minimal fixtures remain supported;
- the host and container scanner regressions omit all subagent markers and fail if the ID-consistency guard is removed;
- the parser cases are table-driven, and the smaller review nits are addressed.

The two feature-gated session tests identified in review now pass. Formatting, the 107 focused Codex tests, Clippy with warnings denied, and the 6,003-test library portion of `cargo test --features serve,e2e-tests --quiet` pass. One later, unrelated tmux tool-session e2e timed out waiting for its own marker; rerunning that test alone also timed out, so I have disclosed that incomplete full-suite result in the PR body rather than representing the entire command as passing.

For completeness, the original marker fields came from a locally observed Codex CLI 0.146.0 child rollout. With IDs sanitized, its relevant metadata was:

```json
{"id":"<child>","session_id":"<root>","parent_thread_id":"<root>","thread_source":"subagent","source":{"subagent":{"other":"guardian"}}}
```

That likely explains the difference from the rollout sample in the review, but this revision no longer depends on those fields.

As stated in the PR description, this revision was produced with OpenAI Codex and has not been human-reviewed. It is offered as the tested local patch/reproduction, not as a claim that the implementation is perfect or necessarily the preferred upstream solution.

njbrake (approved):

Approving. The revision answers @Seluj78's review, and independently of that I checked the mechanism against Codex's own source rather than the description, since a wrong discriminator here silently resumes the wrong conversation.

`rollout/src/recorder.rs` builds the filename as `rollout-{date}-{conversation_id}` and writes `SessionMeta { session_id, id: thread_id, .. }` where `thread_id` is that same id, so `payload.id` is the filename UUID by construction. `core/src/session/session.rs` then sets `session_id` to `agent_control.session_id()` when `session_source.is_non_root_agent()` and to `SessionId::from(thread_id)` otherwise. A subagent rollout therefore carries the root session's UUID while its filename is its own thread id, which is exactly what the equality check keys on. I cross-checked the same `SessionMeta` roster against the `codex-cli` 0.146.0 build I have installed, and traced every `CreateThreadParams` construction to confirm a legitimate top-level rollout cannot false-reject. The three tests are mutation-sensitive: replacing the identity loop with a no-op fails all of them.

The red Playwright check is not yours. `skills-management.spec.ts` tripped strict mode on `getByRole("status")` because the update banner carries that role too; b5855f66 fixed it on main yesterday, and its own message notes the spec fails on any branch cut before the current release. This branch predates it, so squashing onto current main makes it moot.

Two things I am filing as follow-ups rather than holding this for. The guard is inert for legacy self-consistent child rollouts, where `session_id`, `id`, and the filename all agree; upstream filters that case separately. And `collect_codex_sessions` only globs `*.jsonl`, so Codex's compressed `*.jsonl.zst` rollouts are invisible to capture entirely, which is pre-existing and unrelated to this change.

One correction to the earlier review thread, for the record: the marker fields `parent_thread_id` and `thread_source` are real `SessionMeta` fields. They were absent from the 302 surveyed rollouts because that survey held root rollouts, where `parent_thread_id` is skipped when `None` and `thread_source` is `User`. The conclusion to avoid marker sniffing still stands on other grounds.

PR #2764, opened 2026-07-09
feat: respond to agent permission prompts from the sidebar

athal7:
## Description
Adds an `a`/`A` sidebar action that opens an Allow / Allow-Always / Deny dialog for the selected session's pending permission prompt (Claude Code, opencode), sending the exact literal keystrokes each CLI expects via a new `send_key_tokens` tmux primitive — no attach required. Reuses the same tmux keystroke-injection path as the existing "send message without entering session" (`m`/`M`) feature. Static per-agent keystroke tables live on `AgentDef.permission_response`; no ACP/serve/daemon dependency, no pane parsing, no `Status` enum change.

Closes #2747

## PR Type

- [x] New Feature

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary (`docs/development/adding-agents.md`)
- [ ] For UI changes: included screenshot or recording — this is a terminal UI change; happy to add the `needs-recording` CI label if a recording is wanted.

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: this PR doesn't change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (`permission_response_e2e.rs`)

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude (Sonnet) via opencode, multi-agent orchestration

**Any Additional AI Details you'd like to share:** Implemented via an agentic plan → build → verify pipeline, with human review and approval at each phase (issue creation, design decisions, and shipping).

- [x] I am an AI Agent filling out this form

### High-level summary
This PR adds a new “Respond to Permission” quick action (**a/A**) to the TUI session list/sidebar. It lets you approve or deny a session’s pending permission/tool prompt **without attaching to the session**.

Key changes:
- Adds the **Respond to Permission** action (shortcut **a/A**) with an info fallback when the selected session/agent doesn’t support this response
- Introduces an **Allow / Allow-Always / Deny** dialog (quick keys **a/A/d**, plus **Esc** to cancel)
- Adds **per-agent configuration** for the *exact keystrokes* needed to answer the prompt correctly
- Implements a new tmux sending primitive to deliver those keystrokes in the correct order (reusing the existing tmux “injection” behavior pattern used elsewhere)
- Adds gating so the action is a **no-op** for unsupported or **structured** sessions, and only applies to supported non-structured sessions
[...]

PR #3264, opened 2026-08-05
feat(agents): add omp permission_response with optional allow_always

athal7:
## Description
Maps Oh My Pi's (`omp`) terminal tool-approval prompt (`Enter` to allow, `Down`+`Enter` to deny) to the sidebar quick permission-response feature added in #2764. OMP's prompt (`ExtensionToolWrapper` in `packages/coding-agent/src/extensibility/extensions/wrapper.ts`) is a plain two-choice `Approve`/`Deny` list with no in-prompt "always allow" option, so `PermissionResponse.allow_always` is now `Option<&[KeyToken]>` instead of a required field. The respond-to-permission dialog now builds its choice list dynamically: it drops the "Allow Always" choice (and its `Shift+A` shortcut) for any agent whose mapping has none, rather than offering a button that would silently no-op.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the tmux keystroke-injection mechanism (`send_key_tokens`, including `KeyToken::Named` sequences) already has dedicated unit + e2e coverage from #2764, and only `claude` gets a full e2e fixture as the feature's representative (`opencode`'s `Named` tokens aren't e2e-tested either). Added unit coverage instead: `test_omp_agent_definition` (agent detection, untouched from main), `permission_response_tokens_tests::allow_always_none_maps_to_none`, `dialogs::permission_response::tests::allow_always_unsupported_is_skipped_and_shift_a_is_noop` (2-choice dialog, `Shift+A` no-ops), and `agent_without_allow_always_still_opens_dialog`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** anthropic/claude-sonnet-5 via Oh My Pi (omp), orchestrated interactively with human review/redirection at each step.

**Any Additional AI Details you'd like to share:** Keystroke mapping was verified against `omp://approval-mode.md` / `omp://keybindings.md` docs and the shipped `omp` binary's bundled JS (via `strings`), not guessed.

This PR adds Oh My Pi (`omp`) support to the quick permission-response flow.

It maps `omp` approval prompts to allow with `Enter` and deny with `Down` + `Enter`. It makes `allow_always` optional because `omp` does not provide that choice. The dialog hides "Allow Always" and its `Shift+A` shortcut when unsupported.

The flow now safely ignores unavailable permission-response actions. Tests cover `omp`, optional "Allow Always" support, and dialog behavior.

Benefits:
- Supports another agent.
- Prevents unavailable actions from appearing.
- Keeps choices and shortcuts aligned with agent capabilities.

### DIFF
--- a/src/tui/dialogs/permission_response.rs
+++ b/src/tui/dialogs/permission_response.rs
@@ -29,7 +29,7 @@ fn focused_choice_style(theme: &Theme) -> Style {
     Style::default().fg(fg).bg(theme.selection).bold()
 }
 
-/// Which of the three static choices the user picked.
+/// Which choice the user picked; not every agent offers `AllowAlways`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PermissionResponseChoice {
     Allow,
@@ -42,6 +42,9 @@ pub struct PermissionResponseDialog {
     choices: Vec<(&'static str, PermissionResponseChoice)>,
     /// Index of the focused choice within `choices`.
     focused: usize,
+    /// Whether `choices` includes `AllowAlways`; computed once in `new`
+    /// instead of re-scanning `choices` from `handle_key` and `render`.
+    supports_allow_always: bool,
 }
 
 const ALL_CHOICES: [(&str, PermissionResponseChoice); 3] = [
@@ -55,24 +58,22 @@ impl PermissionResponseDialog {
         session_title: &str,
         allow_always: Option<&'static [crate::agents::KeyToken]>,
     ) -> Self {
+        let supports_allow_always = allow_always.is_some();
         let choices = ALL_CHOICES
             .into_iter()
             .filter(|(_, choice)| {
-                allow_always.is_some() || *choice != PermissionResponseChoice::AllowAlways
+                supports_allow_always || *choice != PermissionResponseChoice::AllowAlways
             })
             .collect();
         Self {
             session_title: session_title.to_string(),
             choices,
             focused: 0,
+            supports_allow_always,
         }
     }
 
     pub fn handle_key(&mut self, key: KeyEvent) -> DialogResult<PermissionResponseChoice> {
-        let supports_allow_always = self
-            .choices
-            .iter()
-            .any(|(_, choice)| *choice == PermissionResponseChoice::AllowAlways);
         match key.code {
             KeyCode::Esc => DialogResult::Cancel,
             KeyCode::Enter => DialogResult::Submit(self.choices[self.focused].1),
@@ -84,8 +85,10 @@ impl PermissionResponseDialog {
                 self.focused = (self.focused + 1) % self.choices.len();
                 DialogResult::Continue
             }
+            // a/d mirror structured_view/input.rs; A offered only when the
+            // agent has allow_always.
             KeyCode::Char('a') => DialogResult::Submit(PermissionResponseChoice::Allow),
-            KeyCode::Char('A') if supports_allow_always => {
+            KeyCode::Char('A') if self.supports_allow_always => {
                 DialogResult::Submit(PermissionResponseChoice::AllowAlways)
             }
             KeyCode::Char('d') | KeyCode::Char('D') => {
@@ -150,11 +153,7 @@ impl PermissionResponseDialog {
         );
 
         let mut hint = String::from("a=allow");
-        if self
-            .choices
-            .iter()
-            .any(|(_, choice)| *choice == PermissionResponseChoice::AllowAlways)
-        {
+        if self.supports_allow_always {
             hint.push_str("  A=always");
         }
         hint.push_str("  d=deny  Esc=cancel");
@@ -174,13 +173,19 @@ mod tests {
     use super::*;
     use crossterm::event::KeyModifiers;
 
+    /// Non-empty stand-in for a real agent's `allow_always` mapping. `Some(&[])`
+    /// would also satisfy `.is_some()` but reads as "supported with zero
+    /// keystrokes", a footgun if copy-pasted into a real `AgentDef`.
+    const ALLOW_ALWAYS: Option<&[crate::agents::KeyToken]> =
+        Some(&[crate::agents::KeyToken::Literal("2")]);
+
     fn key(code: KeyCode) -> KeyEvent {
         KeyEvent::new(code, KeyModifiers::NONE)
     }
 
     #[test]
     fn enter_submits_focused_default_allow() {
-        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
+        let mut dialog = PermissionResponseDialog::new("test", ALLOW_ALWAYS);
         let result = dialog.handle_key(key(KeyCode::Enter));
         assert!(matches!(
             result,
@@ -190,7 +195,7 @@ mod tests {
 
     #[test]
     fn a_submits_allow_directly() {
-        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
+        let mut dialog = PermissionResponseDialog::new("test", ALLOW_ALWAYS);
         let result = dialog.handle_key(key(KeyCode::Char('a')));
         assert!(matches!(
             result,
@@ -200,7 +205,7 @@ mod tests {
 
     #[test]
     fn shift_a_submits_allow_always_directly() {
-        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
+        let mut dialog = PermissionResponseDialog::new("test", ALLOW_ALWAYS);
         let result = dialog.handle_key(key(KeyCode::Char('A')));
         assert!(matches!(
             result,
@@ -210,7 +215,7 @@ mod tests {
 
     #[test]
     fn d_submits_deny_directly() {
-        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
+        let mut dialog = PermissionResponseDialog::new("test", ALLOW_ALWAYS);
         let result = dialog.handle_key(key(KeyCode::Char('d')));
         assert!(matches!(
             result,
@@ -220,7 +225,7 @@ mod tests {
 
     #[test]
     fn right_cycles_focus_and_enter_submits_it() {
-        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
+        let mut dialog = PermissionResponseDialog::new("test", ALLOW_ALWAYS);
         dialog.handle_key(key(KeyCode::Right));
         let result = dialog.handle_key(key(KeyCode::Enter));
         assert!(matches!(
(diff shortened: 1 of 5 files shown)


src/tui/dialogs/permission_response.rs
```diff
@@ -39,41 +39,55 @@ pub enum PermissionResponseChoice {
...
-                self.focused = (self.focused + 1) % CHOICES.len();
+                self.focused = (self.focused + 1) % self.choices.len();
                 DialogResult::Continue
             }
-            // Mirrors structured_view's a/A/d mnemonics (src/tui/structured_view/input.rs)
-            // for the same three decisions, so the shortcut is consistent whether the
-            // user is inside the structured view or answering from the sidebar.
             KeyCode::Char('a') => DialogResult::Submit(PermissionResponseChoice::Allow),
-            KeyCode::Char('A') => DialogResult::Submit(PermissionResponseChoice::AllowAlways),
+            KeyCode::Char('A') if supports_allow_always => {
```
jerome-benoit: Minor: the comment that used to sit on this arm pointed out that `a`/`A`/`d` mirror `structured_view/input.rs`. That invariant still holds (structured_view still maps `Shift+A` -> AllowAlways), but the only breadcrumb tying the two mnemonic tables together is now gone, so they can silently drift. Consider restoring a trimmed line, e.g. `// a/d mirror structured_view/input.rs; A offered only when the agent has allow_always.`

src/tui/dialogs/permission_response.rs
```diff
@@ -39,41 +39,55 @@ pub enum PermissionResponseChoice {
 
 pub struct PermissionResponseDialog {
     session_title: String,
-    /// Index of the focused choice: 0=Allow, 1=Allow Always, 2=Deny.
+    choices: Vec<(&'static str, PermissionResponseChoice)>,
```
jerome-benoit, line 42: Nit (optional, non-blocking): `choices` is a <=3 element list whose only variability is whether `AllowAlways` is present. A `supports_allow_always: bool` set once in `new()` would avoid the heap `Vec` and remove the two repeated `.iter().any(|(_, c)| *c == AllowAlways)` scans in `handle_key` and `render`. Defensible as-is (cold path, file is 232 LOC), so treat as taste.

src/tui/dialogs/permission_response.rs
```diff
@@ -157,7 +180,7 @@ mod tests {
 
     #[test]
     fn enter_submits_focused_default_allow() {
-        let mut dialog = PermissionResponseDialog::new("test");
+        let mut dialog = PermissionResponseDialog::new("test", Some(&[]));
```
jerome-benoit: Nit: `Some(&[])` reads as "supports allow_always but with zero keystrokes". It works because the dialog only checks `.is_some()` and never dereferences the slice, but a non-empty sentinel (e.g. `Some(&[KeyToken::Literal("2")])`) would be more self-documenting and avoid a copy-paste footgun into a real `AgentDef`, where an empty slice would make `send_key_tokens` a silent no-op.

jerome-benoit (commented):
## Initial review

**Verdict: approve-with-nits — no blockers, no majors.** The `Option<&'static [KeyToken]>` migration is complete and correct: every reader and the sole `PermissionResponseDialog::new` call site are updated, no unmigrated usage, no positional-index assumption survives, and `AllowAlways` cannot be submitted for an agent that lacks it (three independent gates: filtered out of `choices`, `Char('A')` guarded by `supports_allow_always`, and `permission_response_tokens` returns `None` with a `let-else` bail in `execute_permission_response`).

Verified locally on the PR head:
- `cargo check -p agent-of-empires` clean; `cargo clippy -- -D warnings` zero warnings; `cargo test permission_response` **18 passed / 0 failed**.
- `permission_response.rs` = 232 pure LOC (under the 250 ceiling).
- No positional `0=Allow/1=.../2=Deny` indexing survives; no `--`/emdash added; no dead code.
- New tests genuinely lock the behavior (each fails on revert); `agent_without_allow_always_still_opens_dialog` exercises the real `omp` mapping.

All remaining points are minor/cosmetic.

### Findings
1. **Minor — no e2e for the omp / 2-choice path.** `tests/e2e/permission_response_e2e.rs` covers only the 3-choice `claude` agent; the new 2-choice dialog and the omp `Down+Enter` deny round-trip through tmux are unit-tested only. Per AGENTS.md this is "should consider", not mandatory, so non-blocking. A ~40-line sibling e2e (fake `omp`, press `d`, assert the pane received `Down`+`Enter`, and the hint shows no `A=always`) would close it.
2. **Minor — stale "three" wording.** Two doc comments now over-specify the choice count, which is dynamic (2 or 3) after this PR. Both are context lines so I can't attach inline: `src/agents.rs:343` "The three keystroke sequences" -> drop "three"; `src/tui/dialogs/permission_response.rs:32` "Which of the three static choices" -> reword.
3. **Minor — dropped cross-view breadcrumb** (inline).
4. **Nit — `choices: Vec` + repeated scans** (inline), and `Some(&[])` in migrated tests (inline).

### Out of scope / unverifiable
The actual omp keystroke contract (Enter = allow, Down+Enter = deny) can't be verified without the omp binary; internal wiring is consistent and this rests on the author's stated binary check, which matches the intended validation method for these hand-mapped sequences.

_Review assisted by AI; every claim verified against the checked-out PR head._

athal7:
Pushed fd6fb0f1 addressing all three inline nits (breadcrumb comment restored, `supports_allow_always` hoisted to drop the duplicate scans, `Some(&[])` sentinel swapped for a non-empty stand-in) plus the stale "three" wording in both `agents.rs` and `permission_response.rs`.

On the e2e suggestion (finding 1): agreed it'd be a nice-to-have, but per AGENTS.md this is "should consider" not mandatory, and the tmux keystroke-injection path it'd exercise is already covered by #2764's e2e + this PR's new unit tests. Leaving as-is rather than adding a fixture-only test; happy to pick it up as a follow-up if you feel strongly.

athal7:
Re-requesting review — all nits from the initial review have been addressed:

1. **Breadcrumb comment restored** — `permission_response.rs:88-89` now has `// a/d mirror structured_view/input.rs; A offered only when the agent has allow_always.`
2. **`supports_allow_always` hoisted** — field on `PermissionResponseDialog` computed once in `new()`, eliminating the repeated `.iter().any()` scans in `handle_key` and `render`.
3. **`Some(&[])` sentinel replaced** — tests now use `Some(&[KeyToken::Literal("2")])` as a non-empty stand-in, avoiding the copy-paste footgun.

CI is green (Lint PR title ✅). Ready to merge once approved.

njbrake (commented):

I verified the omp mapping against the shipped build rather than the description, since a wrong `Down`+`Enter` would turn a deny into a grant. In `packages/coding-agent/src/extensibility/extensions/wrapper.ts` the approval is `select(p, ["Approve", "Deny"])`, a literal two-option list and the only one in the binary. `HookSelectorComponent` starts focus at index 0 and its navigation clamps rather than wraps. And `"Always allow"` occurs exactly once in the whole binary, in `acp-permission-gate.ts`, which is the ACP path the structured view uses and which `execute_permission_response` returns early on. So `Enter` is Approve, `Down`+`Enter` is Deny, and there is no always-allow choice to map. The `Option` widening is forced by that rather than opportunistic, and the three gates (filtered choices, the guarded `Char('A')`, and the `None` bail in the executor) each independently keep `AllowAlways` away from agents that lack it.

Two asks before this merges.

**1. The Test Coverage Analysis is inaccurate.** It credits `test_omp_agent_definition` with covering the "exact keystroke mapping". That test lives on `main` at `src/agents.rs:1774`, is untouched by this diff, and asserts only the detection, yolo, resume, and oneshot flags. Nothing in the tree reads omp's real keystroke mapping. I am not asking for a test to close that: pinning a hand-mapped constant is what AGENTS.md rules out. Just correct the description, since as written it invited the earlier review to file the mapping under covered.

**2. `docs/development/adding-agents.md` is now stale.** It is the canonical guide for wiring `permission_response` and still presents `allow_always` as a plain field, so the next contributor writes the pre-PR shape.

```diff
diff --git a/docs/development/adding-agents.md b/docs/development/adding-agents.md
index e8cbffb0..d4bfc375 100644
--- a/docs/development/adding-agents.md
+++ b/docs/development/adding-agents.md
@@ -123,4 +123,4 @@ hooks:
 - **Wrong hook format:** test that hooks fire by sending a message and checking `/tmp/aoe-hooks-$(id -u)/*/status` (host) or `/tmp/aoe-hooks/*/status` (inside the sandbox).
 - **Sandbox hooks are separate:** host installation skips containers; wire into `build_container_config` too.
 - **Waiting status needs a dedicated event:** not all agents expose an approval/permission event. If none exists, document it as a limitation and consider filing upstream.
-- **Sidebar quick permission response:** the TUI's `a`/`A` sidebar action (respond to a pending permission prompt without attaching) needs each agent's exact keystroke sequence, not detection. Set `AgentDef.permission_response` to a `PermissionResponse { allow, allow_always, deny }` of `KeyToken`s (see `claude`/`opencode` in `src/agents.rs`) once you've confirmed, by hand, how the agent's own CLI prompt is answered (bare digit, arrow+Enter, etc., with no assumed trailing Enter). Leave it `None` if you haven't verified the sequence; the action simply no-ops for that agent.
+- **Sidebar quick permission response:** the TUI's `a`/`A` sidebar action (respond to a pending permission prompt without attaching) needs each agent's exact keystroke sequence, not detection. Set `AgentDef.permission_response` to a `PermissionResponse { allow, allow_always, deny }` of `KeyToken`s (see `claude`/`opencode` in `src/agents.rs`) once you've confirmed, by hand, how the agent's own CLI prompt is answered (bare digit, arrow+Enter, etc., with no assumed trailing Enter). `allow_always` is an `Option`: set it to `None` when the agent's prompt offers no "don't ask again" choice (see `omp`), and the dialog drops the "Allow Always" button plus its `Shift+A` shortcut rather than offering one that would silently do nothing. Leave the whole `permission_response` field `None` if you haven't verified the sequences; the action then tells the user the agent isn't supported yet.
```

On @jerome-benoit's points: I agree with you on skipping an e2e here, since `send_key_tokens` ordering is already asserted directly and a Rust e2e is serial forever. The `Vec` is still there after `fd6fb0f1`, which was his actual suggestion, but it is a cold path and I would not hold the PR for it.

I could not push either fix myself because "Allow edits by maintainers" is off on this branch. If you enable it, I can push directly here and on #3269.

njbrake (approved):

Approving, on the strength of the verification in my earlier comment: omp's terminal approval is a literal two-option `["Approve", "Deny"]` select, focus starts at index 0, navigation clamps rather than wraps, and the only "Always allow" string in the binary belongs to the ACP path that `execute_permission_response` returns early on. So the mapping is right and the `Option` widening is forced by it rather than opportunistic.

The `check-template` red was an infrastructure cancellation, killed after seventy minutes with zero steps recorded. I re-ran it and it passes; nothing was wrong with the body's structure.

I am landing the `docs/development/adding-agents.md` update separately since I cannot push to this branch. The Test Coverage Analysis point stands but is a record-keeping matter, not a code one, so it is not worth holding a verified fix for.

Two notes for whoever touches this next. The mapping is pinned to the omp build I inspected; if omp adds an always-allow choice to its terminal prompt, nothing here will notice. And @jerome-benoit's point about the `Vec` is still open, which I agree with and consider not worth a round trip on a cold path.

Thanks for tracking down the keystroke sequences by hand.

PR #3270, opened 2026-08-06
fix(session): reject shell-injection env keys at every ingestion point

jerome-benoit:
## Description

Environment keys from `sandbox.environment`, per-session `extra_env`, `before_session` hook stdout, and `validate_env_entry` were only partially validated before feeding command/docker construction. This gates every ingestion point through the existing `is_valid_env_key`, skipping an invalid key (one carrying `;`, backticks, `$(...)`, etc.) with a warning instead of letting it reach shell construction.

- `environment.rs`: guard `validate_env_entry`, and the `before_start_env` and configured-entry loops in `collect_environment`.
- `repo_config.rs`: warn when a hook emits an invalid key (the skip already existed silently).

Extracted from #3230 as a self-contained, orthogonal slice; it does not depend on the OMP session-id capture work and reproduces that PR's exact guards so #3230 rebases cleanly on top.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary: no user-facing contract changed
- [x] For UI changes: no UI change, screenshot not applicable

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: added a focused unit test (`test_collect_environment_rejects_invalid_config_extra_and_hook_keys`) that asserts injection keys from config/extra/hook channels are dropped while valid keys survive; a full-binary e2e is unwarranted for input validation.

## Verification

- `cargo test --lib session::environment::tests::test_collect_environment_rejects_invalid_config_extra_and_hook_keys`
- `cargo clippy -- -D warnings`, `cargo fmt -- --check` (pre-commit)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Anthropic) in Oh My Pi

**Any Additional AI Details you'd like to share:** Used to analyze #3230 and extract this orthogonal, self-contained slice.

## Summary

- Validates environment keys from all session inputs, including terminal-view host prefixes.
- Skips invalid keys and logs warnings.
- Preserves valid environment entries.
- Adds focused unit tests for configuration, session extras, hooks, and shell prefixes.

## Benefits

- Prevents malformed keys from reaching shell, Docker, tmux, or process environment APIs.
- Improves visibility when invalid configuration is skipped.
- Keeps the existing user-facing behavior unchanged.

### DIFF
--- a/src/session/environment.rs
+++ b/src/session/environment.rs
@@ -331,9 +331,20 @@ pub(crate) fn shell_escape(val: &str) -> String {
 /// Values are passed through `shell_escape` so spaces, quotes, and shell
 /// metacharacters are preserved literally. Returns an empty string when
 /// the entry list is empty so callers can format unconditionally.
+///
+/// Keys are not escapable here (a shell assignment needs a bare name on the
+/// left), so an entry whose key is not a valid environment variable name is
+/// skipped with a warning rather than concatenated into the command line. The
+/// structured-view sibling [`resolve_host_environment_pairs`] applies the same
+/// rule, so both views drop the same entries.
 pub(crate) fn host_environment_prefix(entries: &[String]) -> String {
     let mut out = String::new();
     for entry in entries {
+        let key = entry.split_once('=').map(|(key, _)| key).unwrap_or(entry);
+        if !is_valid_env_key(key) {
+            tracing::warn!(target: "session.create", "invalid host environment key '{}'; skipping", key);
+            continue;
+        }
         if let Some((key, value)) = entry.split_once('=') {
             let resolved = if let Some(rest) = value.strip_prefix("$$") {
                 Some(format!("${}", rest))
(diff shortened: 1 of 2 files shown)


njbrake (approved):

Approving. I reproduced the hole and the fix by execution rather than reading: on `main`, `collect_environment` with `sandbox.environment = ["FOO; touch /tmp/aoe_pwn; X=1"]` returns the injected key verbatim, and `sandbox` is repo-overridable with `environment` not in `REPO_DENIED_SANDBOX_FIELDS`, so a checked-out repo could reach a host shell. On this branch the key is gone.

I pushed one commit (98fab0ca) rather than send it back for a round trip. `host_environment_prefix`, the terminal-view path, was still concatenating the key into the tmux command line as a bare assignment, which is a position no quoting can protect. Its structured-view sibling `resolve_host_environment_pairs` already had the guard and the doc comment claims both use the same grammar, so the two views disagreed about the same config list. That is now closed and the PR title is accurate.

Two things I am taking as follow-ups rather than holding the fix for, tracked separately:

`is_valid_env_key` is stricter than Docker and `Command::env`, so `FOO-BAR=x` is now dropped with only a `tracing::warn!`. That is the right trade, but it is a user-facing narrowing and the checklist says no contract changed. It needs a line in `docs/guides/configuration.md`.

The new warning in `parse_env_kv_lines` fires per line for any hook that prints a diagnostic containing `=`, which the docs describe as silently ignored.

Nice judgment call keeping the key out of the `repo_config` warning, given hook stdout is the documented secrets channel.

PR #3271, opened 2026-08-06
fix(session): remove workspace dir non-recursively to protect user data

jerome-benoit:
## Description

Session deletion ended the workspace stage with `std::fs::remove_dir_all` on a path read straight off the session record, guarded only by a best-effort `workspace_dir_is_aoe_owned` shape check. A corrupt or future-mislaid `workspace_dir` naming an unrelated ancestor could therefore recursively delete user data.

This switches the final workspace-parent teardown to a non-recursive `std::fs::remove_dir`: it succeeds only once the managed worktrees are gone and the directory is empty, so an ancestor holding unrelated content now surfaces a visible `Workspace dir:` error instead of a silent recursive wipe. The shape check is retained as defense-in-depth, and its doc is updated to reflect that it no longer claims to prove ownership on its own.

Extracted from #3230 as a self-contained, orthogonal safety fix independent of the OMP session-id capture work.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary: `workspace_dir_is_aoe_owned` doc updated; no user-facing contract changed
- [x] For UI changes: no UI change, screenshot not applicable

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: added a deterministic in-module regression test (`e2e_workspace_ancestor_with_unrelated_content_is_not_recursively_removed`) that drives `perform_deletion` against a corrupt-ancestor record and asserts unrelated content survives with a visible error; this reproduces the hazard without a full-binary e2e.

## Verification

- `cargo test --lib session::deletion::` (28 passed, incl. the new regression test)
- `cargo clippy -- -D warnings`, `cargo fmt -- --check` (pre-commit)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Anthropic) in Oh My Pi

**Any Additional AI Details you'd like to share:** Used to analyze #3230 and extract this orthogonal, self-contained slice.

## Summary

- Changed workspace cleanup to use non-recursive directory removal.
- Preserved directories that contain unrelated content.
- Clarified workspace ownership checks and cleanup behavior.
- Added regression tests for populated ancestor and non-empty workspace directories.
- Updated the kept-workspace message and related documentation.

## Benefits

This prevents a misplaced workspace path from deleting unrelated user data. Cleanup now preserves unexpected content and avoids repeated deletion attempts.

## Verification

- Session deletion tests
- Clippy checks
- Formatting checks

### DIFF
--- a/src/session/attach_project.rs
+++ b/src/session/attach_project.rs
@@ -30,10 +30,12 @@
 //!
 //! Two invariants shape the code below.
 //!
-//! **`workspace_dir` only ever contains worktrees aoe created.** That is what makes
-//! the delete path's `remove_dir_all` safe, and it is why the in-place case creates
-//! a worktree instead of adopting the user's checkout. `deletion` still verifies it
-//! with `workspace_dir_is_aoe_owned` rather than trusting the record.
+//! **`workspace_dir` only ever contains worktrees aoe created.** That is why the
+//! in-place case creates a worktree instead of adopting the user's checkout.
+//! `deletion` still verifies the layout with `workspace_dir_is_aoe_owned` rather
+//! than trusting the record, and its final removal is non-recursive, so anything
+//! else still sitting under the directory keeps it on disk instead of being
+//! deleted with it.
 //!
 //! **A branch aoe did not create is never touched.** The session's branch name is a
 //! suggestion. If the added repo already has that branch, the attach refuses unless
--- a/src/session/deletion.rs
+++ b/src/session/deletion.rs
@@ -383,10 +383,12 @@ fn perform_deletion_with(
             // preserved; otherwise we'd nuke the user's uncommitted changes,
             // or a default-branch checkout, through the back door.
             //
-            // The ownership guard is the second half of that: this is a
-            // recursive delete of a path read straight off the session
-            // record, so it must prove aoe created that directory rather
-            // than trusting the record. See `workspace_dir_is_aoe_owned`.
+            // The ownership guard is the second half of that: the workspace dir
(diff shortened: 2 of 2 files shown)


njbrake (approved):

Approving. Never recursively deleting a path read off a stored record is the right default for a tool whose auto-purge runs unattended with `force_delete: true`, and downgrading the non-empty case to a message rather than an error is the correct call, since a refusal that can never converge would otherwise strand the row forever. I confirmed `DirectoryNotEmpty` is the kind `remove_dir` actually returns across the non-empty forms, and that it is stable well inside our 1.85 MSRV.

I pushed one commit (b8ea2359) for wording. The kept message claimed the leftover files were "not created by aoe"; for a multi-repo session the workspace root is the agent's own cwd, so in the common case aoe's agent did write them. `attach_project`'s module doc also still credited the invariant with making `remove_dir_all` safe.

One thing worth being explicit about, since the description leads with the corrupt-ancestor case: I could not find a production path that produces `WorkspaceRepo { managed_by_aoe: false }`, so that scenario currently needs a hand-edited `sessions.json`. The everyday effect is different. Anything sitting at a multi-repo workspace root now keeps the directory on disk after deletion, and a single `.DS_Store` is enough to trigger it. That is a better failure than deleting a user's files, but it leaves an orphaned directory with no row to retry from. Filing that as a follow-up rather than holding this.

PR #3275, opened 2026-08-07
feat(web): add collapsible header and composer on mobile conversations

summelon:
First contribution here — I really appreciate the web dashboard of AoE. This is the one I 
have been looking for a long time.It's definitely the current best one in my mind, comapred 
to YepAnywhere or Collie for Herdr. Still, by using it day to day, there are some minor parts 
of the frontend on my phone, which I think could be further improved. That's why I raise this 
PR and some others WIP. I am a newbie to TypeScript and Rust. But I have co-reviewed with 
my agents about the modifications and the PR contents. I would appreciate if anyone can remind 
me that there is any aspect I should take more care of. Happy to reshape any of it if the approach 
doesn't fit where you want the mobile layout to go.

## Description

On a phone the app top bar and the ACP composer together eat roughly a third of
the screen, leaving very little room for the transcript. This gives each of them
its own collapse handle so the transcript can take that height back. The two
states are independent, so all four combinations are reachable.

Both regions collapse via a `1fr` / `0fr` grid row with `min-h-0` on the child.
The row genuinely leaves layout, so the `flex-1` transcript viewport grows into
the freed space with no measured heights and no `vh` math. Clipping is applied
only while collapsed, because the composer's model and permission menus render
`absolute bottom-full`, outside their own box, and a permanent `overflow-hidden`
would cut them off.

Each handle is a sibling of the region it collapses, never a child: a handle
nested inside the collapsing element would disappear with it and strand the user
in the collapsed state. The host is zero-height with an absolutely positioned
button, so the handle costs the layout nothing in either state. Labels swap with
state (Collapse / Expand conversation header, Collapse / Expand message
composer) and the collapsed region is `inert`, so hidden chrome leaves the tab
order and the accessibility tree.

Scope is deliberately narrow. Header state lives in `App` (the top bar is the
shell's own child) and applies only to the mobile structured-view conversation
pane; desktop, the dashboard, settings, the diff pane, and terminal sessions
render the bar exactly as before. Composer state is local to the structured view
and is width-gated on the existing `md` breakpoint via `useIsWideViewport`.

One supporting fix: the transcript viewport now observes itself for resizes as
well. The existing below-viewport observer only sees chrome that sits inside the
structured view, so a bottom-pinned transcript would silently un-pin when the
header expands. Both observers share one `repinOnResize` helper.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

| Both expanded | Composer collapsed | Both collapsed | Header collapsed |
| --- | --- | --- | --- |
|1-both-expanded | 2-composer-collapsed | 3-both-collapsed | 4-header-collapsed |

Checks run locally (web-only change, no Rust touched):

- `cd web && npm run format:check` clean (oxfmt)
- `cd web && npm run lint` clean (ESLint)
- `cd web && npx tsc -b` clean
- `cd web && npx vitest run`: 342 files, 3715 tests pass
- `cd web && npm run build && npx playwright test`: 399 pass, 3 skipped. One pre-existing failure, `acp-history-window.spec.ts:76`, which also fails on unmodified `main` locally under full-suite load and passes when run alone.

Documentation updated: `docs/guides/web/dashboard.md`'s "On mobile" section now covers collapsing the top bar and the composer independently, and `docs/structured-view/interface.md` qualifies its tap-to-focus paragraph, since a collapsed composer cannot be focused until it is restored.

## Test Coverage Analysis

`web/tests/acp-mobile-chrome-collapse.spec.ts` at iPhone 13 walks all four
header / composer combinations in one test (per the repo's "one test per
behavior, not per input" rule) and asserts the released heights actually land in
the transcript viewport. It asserts on measured heights rather than visibility,
because a collapsed region clips a child that still has a box of its own, so
Playwright would report that child as visible. It also checks that both handles
stay on screen and tappable while both regions are collapsed, and that the
composer still sends after a full round trip.

Mocked rather than live because everything asserted is client-side layout; no
backend state is involved.

`web/src/components/__tests__/CollapsibleChrome.test.tsx` covers the `inert` transition, the handle's label / `aria-expanded` / glyph direction across all four (edge, collapsed) combinations in one table, and that `aria-controls` resolves to the region element the handle collapses. `web/src/lib/repinOnResize.test.ts` covers the transcript re-pin path the review flagged as the coverage gap. The Playwright spec adds a hit-area regression test: each handle measures 28×16, paints its own background, is what `elementFromPoint` reports inside that box and not just outside it, and the update banner's dismiss button still takes a real click.

Matrix entry: `app-shell.mobile-chrome-collapse`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Code (Opus 5)

**NOTE:**

## Summary

- Added independent collapse controls for the mobile conversation header and message composer.
- Returned collapsed space to the transcript without viewport calculations.
- Kept collapse controls visible and accessible for restoration.
- Kept desktop layouts and other views unchanged.
- Preserved composer drafts and prevented focus or typing in the collapsed composer.
- Improved transcript repinning when the header or viewport changes size.
- Added unit tests, mobile browser tests, accessibility checks, and coverage tracking.
- Updated mobile behavior documentation.

## Benefits

- Mobile users get more space to read the transcript.
- Users can hide or restore the header and composer independently.
- Existing desktop behavior remains unchanged.
- Transcript position stays stable during layout changes.

## Further enhancement

- Test additional mobile layouts and device sizes.


njbrake (dismissed):

This is careful work for a first contribution. The `1fr`/`0fr` plus `min-h-0` collapse is the right technique, clipping only while collapsed so the composer's `bottom-full` menus survive is a real detail to have caught, keeping `CollapsibleRegion` unconditional on the composer so crossing `md` does not drop the draft is the right call, and the comments explain why rather than what. Coverage matrix entry is present and the test tiers match `web/AGENTS.md`.

I pushed a2f228da rather than send it back. The 32px touch target you added for CodeRabbit sat directly on top of the update banner's dismiss button, which lives in the same corner, so on an iPhone 13 viewport `elementFromPoint` at the dismiss button returned the collapse handle and a real click timed out on interception. The banner was untappable for as long as a release was pending. Moving the handle below the banners fixes it, verified end to end.

**What I would like resolved before this merges: give the handle a real layout box.** The zero-height host means its 32px hit area is always sitting on top of something, so my fix relocated the hazard rather than removing it. Measured on the current head, the composer handle occupies the bottom-right 32x32 of the transcript itself, and the nearest thing under the header handle after my fix is the rate-limit notice's "Resume now" button, which clears it by 0px. It lands exactly at the handle's bottom edge. That holds today and stops holding the moment that notice renders one line shorter.

A slim full-width grab strip of 16 to 24px, where the whole strip is the target, removes the entire class of collision and still nets roughly +24px for the header and +116px for the composer. That is a design change rather than a patch, which is why I have not written it.

Also needed: `codecov/patch` is at 67.39% against a 75% target, with 11 of the 15 missed lines in the new `StructuredView` resize path.

Smaller things, none blocking on their own: `docs/guides/web/dashboard.md`'s "On mobile" section and `docs/structured-view/interface.md` (which still promises tap-to-focus, now a no-op when the composer is `inert`) both want a sentence; `items-center` on the handle button is dead and only harmless because of Tailwind's alphabetical layer order; the `App.tsx` ternary swaps element type so `TopBar` remounts whenever the collapsible boundary flips, which the composer side avoids; and `aria-expanded` wants a matching `aria-controls`.

One question worth settling first, since there is no linked issue: do we want persistent handles at all, or auto-hide-on-scroll? The latter reclaims the same space with no hit target to place. Happy to go either way, but it is worth agreeing before you spend more time on the handle.

summelon:
Thanks for laying out the alternatives. The following is the discussion results with my AI.
I thought through the interaction a bit more, and my preference is currently 
**3A > 3B > 4 > 1 > 2**, although I’m happy to follow whichever direction you think fits the project best.

### 1. Persistent collapse strip

**Description:** Always-visible thin strip (16–24px) that allows collapsing/expanding UI chrome.
**Pros:** Structurally safe, avoids hit-target overlap, predictable behavior.
**Cons:** Permanently consumes vertical space, reducing the original goal of maximizing transcript area.

### 2. Auto-hide based on scrolling

**Description:** UI chrome (header/composer) hides or shows based on scroll direction.
**Pros:** Potentially maximizes reading space without explicit controls.
**Cons:** Ambiguous intent mapping across multiple UI states (header/composer combinations), increases interaction complexity and edge cases (scroll, focus, keyboard behavior), and is least predictable.

### 3. Explicit overlay controls

**Description:** Use dedicated overlay buttons to independently collapse or expand the header and composer. In both variants, **the visible button area and the clickable area are identical, with no hidden hit target**.

The only decision is the size of the control:

**3A — Current compact size (28×16px)**

* **Pros:** Minimizes UI chrome and maximizes transcript space. The control is visually subtle and takes up very little space.
* **Cons:** Smaller than the repository’s standard 32px touch target. It may require an intentional exception to the guideline, even though space efficiency is the main goal here.

**3B — Standard 32×32px size**

* **Pros:** Aligns with the design system and improves tap comfort. The visible and clickable areas still match exactly.
* **Cons:** More visually prominent and slightly increases coverage over the transcript.

Both options preserve the same interaction model: explicit, independent controls for header and composer, without reserving permanent layout space.

### 4. Central transcript tap area

**Description:** Tap on a neutral area (center) of the transcript toggles UI chrome visibility.
**Pros:** Clean UI with minimal visible controls, potentially fastest interaction.
**Cons:** High risk of interaction conflicts with text selection, links, message actions, scrolling, keyboard focus, and future transcript features; requires careful validation of gesture ambiguity.

---

Overall preference:
**explicit overlay controls > transcript tap gesture > persistent strip > scroll-driven auto-hide**

My main goal is to preserve a simple, predictable way for mobile users to maximize transcript space while keeping interactions robust and maintainable. 
Please kindly let me know your comments on this. I will proceed with your direction and handle the remaining blocking items.
Thank you.

njbrake:
@summelon thanks for the effort and work on this, and also for thinking through the pro/cons of the design in your ticket 🙏  I think you present a great case, and I'm good with your recommended 3A! Let me know when you've tested it all out and are happy with it, and i'll get this thing merged and out to everyone. 

Thanks for using and contributing to AoE :)

summelon:
Thanks @njbrake 

The hidden 32×32 hit target is gone (reverted as its own commit); the handle is back to the compact 28×16 tab where the clickable area is exactly the visible one. Your banner-placement fix is untouched — worth noting that with the handle back under the bar, 28×16 clears the dismiss button by 0px, so your fix is still what gives it real separation.

Also in: `aria-controls` on both handles pointing at the regions they collapse, `TopBar` wrapped unconditionally so it no longer remounts when the collapsible boundary flips (the composer side already did this), and the two doc sections you asked for. `items-center` I kept — it was only dead under the 32×32 wrapper, and it centers the glyph again on the compact button.

For coverage, the re-pin logic moved into a small `repinOnResize` helper so it can be tested without mounting the runtime, and there is a browser-level regression test for the hit target itself: it measures the handle, checks it paints its own background, and checks `elementFromPoint` just outside the tab, so a future invisible wrapper fails the test rather than the user. Locally that puts patch coverage around 79%, up from 67.4%.

PR #3293, opened 2026-08-08
fix(cli): harden tmux dispatch for headless add --launch and send

athal7:
## Description
Hardens `aoe add --launch` and `aoe send` for headless dispatch (LaunchAgent, cron, or any caller with no controlling terminal), found while porting a wrapper script off the raw opencode HTTP API:

- `aoe add --launch` no longer fails with exit 1 when stdout isn't a terminal (the interactive attach step needs a TTY it doesn't have); it skips the attach and reports success since the session already started fine.
- `aoe send` now waits for the target pane's content to settle before typing (reusing the poller `aoe session restart`'s wake-message send already relies on, extracted into `tmux::Session`), instead of typing immediately whenever the pane merely exists (`EnsureReadyOutcome::AlreadyAlive` previously got no wait at all).
- A message starting with `/` is padded with a trailing space before the terminating Enter, so an agent's own slash-command autocomplete dropdown doesn't swallow the Enter and leave the command typed-but-unsubmitted.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/`: tightened `kiro_launch.rs` and `permission_response_e2e.rs` to assert `--launch` now succeeds without a TTY (previously they explicitly skipped the exit-code check because of this bug), and added `test_cli_send_waits_for_slow_boot_before_typing` in `cli.rs` proving `aoe send` delivers a message to a session that's still mid-boot. Also added a pure unit test for the slash-command padding decision.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** anthropic/claude-sonnet-5, via an agentic coding harness (OpenCode/omp-based)

**Any Additional AI Details you'd like to share:** Implementation, tests, and this description were produced by the AI agent under my direction and review; I verified the diff, ran the test suite, and did a manual smoke test of `aoe add --launch` from a genuine non-TTY subprocess before opening this PR.

## Summary

- `aoe add --launch` now succeeds without a controlling terminal. It skips interactive attachment and prints a quoted attach command.
- `aoe send` waits for the target pane to become ready before sending input.
- Agents can define readiness markers. Readiness matching is case-insensitive, with content settling used when no marker exists.
- Slash-prefixed messages receive trailing-space padding before submission. This prevents autocomplete from consuming the command.
- Added unit and end-to-end tests for headless launch, slow startup, readiness detection, and slash-command handling.
- Updated tests to refresh stale tmux session state and verify successful headless execution.

## Benefits

These changes improve message delivery during slow startup. They also make tmux launches reliable in CI, automation, and other headless environments.

## Technical notes

- Added `Session::wait_until_ready`.
- Added `AgentDef.ready_marker` and the `ready_marker` lookup function.
- Updated session restart and send flows to use agent-specific readiness checks.
- Added a delayed OpenCode agent scenario to test input delivery during slow startup.

### DIFF
--- a/src/agents.rs
+++ b/src/agents.rs
@@ -319,6 +319,16 @@ pub struct AgentDef {
     /// newlines within a paste rather than as "submit". A delay longer than the
     /// agent's burst window lets the suppression expire before Enter arrives.
     pub send_keys_enter_delay_ms: u64,
+    /// Pane-content substring (matched case-insensitively) that indicates
+    /// this agent's TUI has finished booting and is ready to accept input,
+    /// if known. `None` means no such per-agent signal is known yet;
+    /// callers fall back to a generic content-settle heuristic
+    /// (`tmux::Session::wait_until_content_settles`), which cannot tell a
+    /// genuinely idle prompt from a still-booting pane that merely hasn't
+    /// printed anything new in the last couple of samples. Used to avoid
+    /// typing into a pane before the agent is actually listening (see `aoe
+    /// send`'s pre-send wait).
+    pub ready_marker: Option<&'static str>,
     /// One-line install command shown when the agent is missing from PATH.
     pub install_hint: &'static str,
     /// Static keystroke sequences for answering this agent's own interactive
@@ -714,6 +724,7 @@ pub const AGENTS: &[AgentDef] = &[
         fork_strategy: ForkStrategy::ClaudeFork,
         host_only: false,
         send_keys_enter_delay_ms: 0,
+        ready_marker: None,
         install_hint: "npm install -g @anthropic-ai/claude-code",
         permission_response: Some(PermissionResponse {
             allow: &[KeyToken::Literal("1")],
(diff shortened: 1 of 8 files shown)


njbrake (approved):
Reviewed locally. The remaining findings are minor and suitable for a fast-follow PR.

njbrake (changes requested):
The new `test_cli_send_waits_for_slow_boot_before_typing` e2e test fails in both Linux test jobs. `aoe add --tool opencode --cmd-override ...` still performs the built-in tool availability check, so CI exits with:

> Error: 'opencode' is not installed or not on $PATH.\n\nPlease install an `opencode` shim in the harness PATH (as other agent e2e tests do), or otherwise make the fixture satisfy availability detection, then verify the bare-core and serve-e2e jobs pass.

njbrake (approved):
drafted by Claude via back-and-forth with @njbrake

Reviewed current head 3b781254. All requested fixes are addressed, targeted local validation passed, and the complete CI matrix is green.

Issue #2011, opened 2026-06-06
Mobile: input field hidden behind keyboard in structured view


hoxkss:
On iOS Safari (and likely other mobile browsers), when tapping the message input field in the web dashboard's structured view, the virtual keyboard opens and the viewport shifts, but the input field ends up positioned behind/underneath the on-screen keyboard. The user cannot see what they are typing.

This appears to be a viewport/layout issue where the app does not properly respond to resize events on mobile. When the on-screen keyboard opens, the layout viewport shrinks but the input area doesn't reposition to stay above the keyboard.

## Steps to reproduce
1. Open the aoe web dashboard on iOS Safari
2. Navigate to a structured view session
3. Tap the message input field at the bottom
4. Keyboard opens, input field ends up hidden behind it

## Expected behavior
Input field should remain visible above the virtual keyboard.

## Environment
- iOS Safari (PWA mode)
- aoe v1.10.1

njbrake:
@hoxkss thank you for the report! Helping to smooth out the corner cases with web UI 🙏 .  PR up to fix it.

njbrake:
_Note: this comment was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Hi @hoxkss, thanks for the report. I have a candidate fix on the branch `fix/structured-composer-ios-keyboard` (PR #2014). It reserves space for the soft keyboard so the structured-view input stays above it on iOS.

The fix depends on how your iOS version reports viewport height when the keyboard opens, so I would like to confirm it resolves the issue on your actual device before merging. If you can build from the branch:

1. `git fetch origin && git checkout fix/structured-composer-ios-keyboard`
2. `cargo build --features serve`
3. `./target/debug/aoe serve --host 0.0.0.0` (plus your usual serve flags)
4. On iOS Safari (PWA), open a structured-view session and tap the message input.

Expected: the input stays visible above the on-screen keyboard instead of being hidden behind it.

If you get a chance to try it, a quick "yes, fixed" would be awesome. Thanks!

hoxkss:
Thanks for the quick fix! I'm using a fork of AoE (added omp agent support in #2010) so I'll need to merge the fix branch into mine first. Will test on my iOS device and report back soon.

PR #3299, opened 2026-08-10
feat(web): add conversation font size controls to the structured view

### LINKED ISSUE #3290: feat(web): configurable font size for Structured View conversation (transcript only)
Hi, me again.
While pending your review of the collapsible button PR, I am simultaneously working on other features 
which may further improve AoE's user experience on the mobile webpage, including this font size adjustment for 
structured view settings and another undergoing exploration of structured view support for Codex.

I did some research on the existing issues/PR relevant to structured view font size adjustment. There seems 
to be no good support for this feature. Therefore, I built a prototype and evaluate on my device. It looks good.

The following is the issue content generated by AI. Please have a look. If you also think it's a useful feature, I will 
further polish the implementation and well-test it based on my first PR experience. Thank you.

---
## Summary

The web dashboard currently lets users configure separate mobile and desktop font sizes for terminal sessions, but Structured View conversation content uses a fixed font size.

I would like to add a browser-local font-size preference for the Structured View conversation transcript, with separate mobile and desktop values, similar to the existing terminal font-size settings.

This is especially useful on phones, tablets, and high-DPI displays, where users may want either larger text for readability or smaller text to fit more conversation history on screen.

## Current behavior

Under **Settings → Terminal**, AoE already provides separate **Mobile font size** and **Desktop font size** controls.

Structured View does not currently have an equivalent setting. Its conversation Markdown uses a fixed text size.

The proposed setting should affect the conversation transcript only, rather than scaling the entire dashboard UI such as the sidebar, TopBar, settings pages, or other controls.

## Related work

There have been two previous PRs around dashboard text sizing:

* #2781 — `feat(web): add text size toggle to TopBar`

  * Proposed a global four-step text-size toggle in the TopBar.
  * During review, @Seluj78 suggested that this preference would be better placed in **Settings** rather than the TopBar.
  * The PR was later automatically closed because the required PR checklist was not restored.

* #2812 — `feat(web): add text size toggle to TopBar`

  * A follow-up implementation of the global TopBar approach.
  * It was closed without being merged.
  * Review also identified problems with applying the size at the document/root level, where Tailwind/root `rem` interactions could affect more of the dashboard layout than intended.

For broader context, #675 / #793 established the expanded web Settings UI and the general pattern of exposing web/dashboard preferences through Settings.

## Proposed behavior

Add a **Conversation display** section under:

**Settings → Structured view**

with:

* **Mobile font size**
* **Desktop font size**

The preferences should:

* apply only to Structured View conversation content;
* be independent from terminal font-size preferences;
* update immediately without requiring a page reload;
* automatically select the mobile or desktop value based on the existing `768px` responsive breakpoint;
* persist locally in the browser using the existing `aoe-web-settings` mechanism;
* scale the Markdown hierarchy consistently, including normal prose, headings, lists, tables, inline code, and fenced code blocks.

I think a default of **14px** with a configurable range of **10–24px** is a reasonable starting point, but the exact range could be adjusted based on maintainer preference.

## Reference implementation

I have a working implementation here:

https://github.com/summelon/agent-of-empires/tree/feature/structured-view-font-size

At the time of opening this issue, the branch is based on current upstream `main` (`9cd37b6`) and is 2 commits ahead / 0 behind.

The prototype currently:

* adds `structuredMobileFontSize` and `structuredDesktopFontSize` to the existing browser-local web settings;
* adds **Conversation display** to the existing Structured View settings page;
* extracts the existing slider + px selector pattern into a reusable `FontSizeControl`, also reused by Terminal settings;
* scopes font-size CSS variables to the Structured View conversation root rather than the document root;
* keeps other dashboard UI and other `.acp-markdown` consumers at their existing sizes;
* preserves relative Markdown typography when the base conversation size changes;
* validates/clamps persisted values;
* updates the web settings documentation;
* includes Vitest/component coverage and a Playwright user-story test for mobile/desktop breakpoint switching and proportional Markdown scaling.

The branch is intended as a reference/prototype and I am happy to adjust the implementation, settings placement, range, naming, or mobile/desktop behavior based on maintainer feedback.

## Expected result

Structured View users can tune conversation readability/density in the same way terminal users can tune terminal text, without changing the scale of the rest of the AoE dashboard.

summelon:
Hi, @njbrake and @Seluj78. Thank you for your reply in the issue. I cannot find any option for reviewer selection in the PR page. Sorry that I will just let it go.
Closes #3290.
Implement the structured-view font size controls in the settings page, aligning with the existing terminal font size control. Below are generated by Claude:

## Description

The structured-view transcript was pinned to `text-sm` with a handful of fixed `rem` child sizes, so there was no way to fit more conversation on a phone screen short of zooming the whole dashboard.

This adds a mobile and a desktop conversation font size (6-28px in 1px steps, default 14px) under **Settings > Sessions > Structured view > Conversation display**.

- Both are dashboard preferences stored in the existing `aoe-web-settings` entry via `useWebSettings`, deliberately separate from the terminal font sizes: the transcript is proportional prose, the terminal is a fixed-cell grid, and users size them for different reasons. Like the rest of that entry they ride the existing `webUiSync` mirror to the daemon's web-UI state, so they follow the user across browsers; the agent config schema is untouched, and the values are not per profile.
- The structured-view root publishes both values as scoped custom properties, and CSS picks the active one using the dashboard's canonical mobile classifier: a coarse primary pointer and a viewport narrower than 768px. This keeps a narrow desktop window on the desktop size and switches a narrow touch device to the mobile size, with no reload or JS width probe. The stored integer values are published as `rem` relative to the 16px baseline, preserving browser root-font-size accessibility preferences. The style object is merged on top of the existing `structuredViewRootStyle(keyboardHeight)` reservation rather than replacing it, keeping the iOS keyboard layout behavior from #2011 intact.
- Only the conversation renderer opts into the variable, via a new `.acp-markdown-body` class. The other `.acp-markdown` consumers (tool-card bodies, the diff pane's file viewer) keep their own explicit sizes: those base rules stay in `rem` and the `em` ratios are re-declared under `.acp-markdown-body` only. Markdown child typography (headings, table cells, fenced code, code header) uses `em` so the whole hierarchy scales together; the ratios reproduce the previous `text-sm` appearance at the 14px setting with the normal 16px browser root, while continuing to honor larger browser default font sizes.
- The other caller of the `Markdown` component, the skills manager, renders outside the structured-view root and so resolves the variable's `0.875rem` fallback, which is exactly the `text-sm` it had before.
- Persisted values are clamped on read: localStorage is user-editable and the value reaches CSS directly, so a `null`, stringy, fractional, or absurd size must not render the transcript at NaN or 0px.
- `TerminalSettings` was migrated onto the extracted `FontSizeControl`, removing a hand-rolled duplicate of the same slider/select/description markup. The shared control also gives the terminal sliders and selects the `aria-label`s they previously lacked.
- The px range is declared once in `web/src/lib/fontSizeRange.ts` and consumed by both panels and by the conversation clamp, so the two sliders span identical steps and cannot drift apart.

## Screenshots

Captured on a 390x844 mobile viewport against a real `aoe serve` with a scripted agent turn.

| 6px (minimum) | 22px | 28px (maximum) |
| --- | --- | --- |
| transcript-6px | transcript-22px | transcript-28px |

The new panel, at the 14px default:

settings-conversation-display

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (`docs/guides/web/settings.md`)

## Test Coverage Analysis

Tests added:

- `web/tests/acp-conversation-font-size.spec.ts` (mocked Playwright): stores 11px mobile / 20px desktop and exercises the dashboard's coarse-pointer + narrow-viewport classifier in separate fine- and coarse-pointer browser contexts. A narrow fine-pointer desktop remains at the desktop size, while a narrow touch device uses the mobile size and switches back to desktop when widened. The spec also verifies proportional `h1` scaling and that `rem` output follows a changed browser root font size. jsdom evaluates neither media queries nor custom properties, so Playwright covers the browser-level cascade.
- `web/src/components/settings/__tests__/StructuredViewDisplaySettings.test.tsx` (Vitest): the localStorage contract - defaults, independent per-axis writes, slider/select sync, remount persistence, and a table covering backfill plus junk clamping. It also renders `TerminalSettings` alongside and asserts all four sliders carry the same bounds and step count, so a panel that re-hardcodes its own range fails.
- `web/src/components/acp/__tests__/StructuredViewRoot.test.tsx`: both custom properties are published as `rem` without dropping the keyboard reservation, the 14px default maps to `0.875rem`, and the root carries the scope class used by the responsive media query.
- `web/src/components/acp/Markdown.test.tsx`: the conversation renderer carries `.acp-markdown-body` and no fixed `text-*` utility that would outrank the variable.

Coverage matrix entry: `settings.structured-view-font-size`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Implementation, tests, and docs were drafted with Claude Code and reviewed and verified locally.

## Summary

- Added separate mobile and desktop font-size controls for Structured View conversations.
- Added synchronized sliders and selectors with a 6–28px range and a 14px default.
- Stored settings in `aoe-web-settings`, separate from terminal font sizes.
- Applied changes immediately across responsive layouts.
- Updated Markdown typography to scale consistently while preserving other dashboard Markdown sizes.
- Reused the shared font-size control for terminal settings.
- Added validation that rounds and clamps stored values.
- Added Vitest and Playwright coverage for settings, responsive switching, scaling, and code-block line height.
- Updated settings documentation and the coverage matrix.

## Benefits

Users can adjust conversation readability without changing terminal settings. The settings respond to mobile and desktop layouts without a page reload. Invalid stored values no longer create unsupported font sizes.

## Potential follow-up

Consider adding a reset option if more conversation display settings are added.

### DIFF
--- a/web/src/components/acp/StructuredView.tsx
+++ b/web/src/components/acp/StructuredView.tsx
@@ -62,6 +62,7 @@ import { useIsCoarsePointer } from "../../hooks/useIsCoarsePointer";
 import { useIsWideViewport } from "../../hooks/useIsWideViewport";
 import { useMobileKeyboard } from "../../hooks/useMobileKeyboard";
 import { ChromeCollapseHandle, CollapsibleRegion } from "../CollapsibleChrome";
+import { useWebSettings } from "../../hooks/useWebSettings";
 import { dispatchFocusTerminal } from "../../lib/terminalFocus";
 import { shouldFocusComposerOnThreadTap } from "./threadTapFocus";
 import type {
@@ -223,17 +224,40 @@ export function structuredViewRootStyle(keyboardHeight: number): React.CSSProper
   return keyboardHeight > 0 ? { paddingBottom: keyboardHeight } : undefined;
 }
 
+/** Publishes both conversation font-size preferences as scoped custom
+ *  properties on the structured-view root; `index.css` picks the active one via
+ *  a media query, so crossing the 768px breakpoint reflows without a reload and
+ *  without a JS width probe. Merged on top of the keyboard reservation rather
+ *  than replacing it. Both sizes arrive already clamped from `useWebSettings`,
+ *  which is the single normalization boundary for stored prefs. */
+function structuredViewFontStyle(
+  keyboardHeight: number,
+  mobileFontSize: number,
+  desktopFontSize: number,
+): React.CSSProperties {
+  return {
+    ...structuredViewRootStyle(keyboardHeight),
+    "--acp-conversation-font-size-mobile": `${mobileFontSize}px`,
+    "--acp-conversation-font-size-desktop": `${desktopFontSize}px`,
+  } as React.CSSProperties;
+}
+
 /** Fixed-height flex root for the structured view, owning the mobile-keyboard
  *  reservation (see {@link structuredViewRootStyle}). Exported and kept tiny so
  *  the hook-to-style wiring is testable without mounting the assistant-ui
  *  runtime, mirroring the #1282 rate-limit-recovery extraction. */
 export function StructuredViewRoot({ children }: { children: React.ReactNode }) {
   const { keyboardHeight } = useMobileKeyboard();
+  const { settings } = useWebSettings();
   return (
     <div
       data-testid="structured-view-root"
-      className="flex h-full flex-col bg-surface-900 text-text-primary"
-      style={structuredViewRootStyle(keyboardHeight)}
+      className="acp-conversation-scope flex h-full flex-col bg-surface-900 text-text-primary"
+      style={structuredViewFontStyle(
+        keyboardHeight,
+        settings.structuredMobileFontSize,
+        settings.structuredDesktopFontSize,
+      )}
     >
       {children}
     </div>
--- a/web/src/components/acp/StructuredView.tsx
+++ b/web/src/components/acp/StructuredView.tsx
@@ -63,6 +63,7 @@ import { useIsWideViewport } from "../../hooks/useIsWideViewport";
 import { useMobileKeyboard } from "../../hooks/useMobileKeyboard";
 import { ChromeCollapseHandle, CollapsibleRegion } from "../CollapsibleChrome";
 import { useWebSettings } from "../../hooks/useWebSettings";
+import { conversationFontSizeRem } from "../../lib/conversationFontSize";
 import { dispatchFocusTerminal } from "../../lib/terminalFocus";
 import { shouldFocusComposerOnThreadTap } from "./threadTapFocus";
 import type {
@@ -226,10 +227,13 @@ export function structuredViewRootStyle(keyboardHeight: number): React.CSSProper
 
 /** Publishes both conversation font-size preferences as scoped custom
  *  properties on the structured-view root; `index.css` picks the active one via
- *  a media query, so crossing the 768px breakpoint reflows without a reload and
- *  without a JS width probe. Merged on top of the keyboard reservation rather
- *  than replacing it. Both sizes arrive already clamped from `useWebSettings`,
- *  which is the single normalization boundary for stored prefs. */
+ *  the same coarse-pointer + narrow-viewport rule the rest of the dashboard
+ *  classifies mobile with, so a resize or an orientation change reflows without
+ *  a reload and without a JS width probe. Merged on top of the keyboard
+ *  reservation rather than replacing it. The stored px values are published as
+ *  `rem` (see {@link conversationFontSizeRem}) so the transcript still tracks
+ *  the reader's root font size. Both sizes arrive already clamped from
+ *  `useWebSettings`, the single normalization boundary for stored prefs. */
 function structuredViewFontStyle(
   keyboardHeight: number,
   mobileFontSize: number,
@@ -237,8 +241,8 @@ function structuredViewFontStyle(
 ): React.CSSProperties {
   return {
     ...structuredViewRootStyle(keyboardHeight),
-    "--acp-conversation-font-size-mobile": `${mobileFontSize}px`,
-    "--acp-conversation-font-size-desktop": `${desktopFontSize}px`,
+    "--acp-conversation-font-size-mobile": conversationFontSizeRem(mobileFontSize),
+    "--acp-conversation-font-size-desktop": conversationFontSizeRem(desktopFontSize),
   } as React.CSSProperties;
 }
 
--- a/web/src/components/settings/FontSizeControl.tsx
+++ b/web/src/components/settings/FontSizeControl.tsx
@@ -1,8 +1,8 @@
+import { MAX_FONT_SIZE, MIN_FONT_SIZE } from "../../lib/fontSizeRange";
+
 interface FontSizeControlProps {
   label: string;
   value: number;
-  min: number;
-  max: number;
   onChange: (value: number) => void;
   description?: string;
   /** Prefix for the slider/select `data-testid`s, so a settings panel with two
@@ -10,11 +10,14 @@ interface FontSizeControlProps {
   testIdPrefix: string;
 }
 
+const OPTIONS = Array.from({ length: MAX_FONT_SIZE - MIN_FONT_SIZE + 1 }, (_, i) => MIN_FONT_SIZE + i);
+
 /** Slider + px select pair shared by the terminal and conversation font-size
  *  settings. Both controls write the same value so they stay synchronized; the
- *  select offers every integer step in range. */
-export function FontSizeControl({ label, value, min, max, onChange, description, testIdPrefix }: FontSizeControlProps) {
-  const options = Array.from({ length: max - min + 1 }, (_, i) => min + i);
+ *  select offers every integer step in range. The range is not a prop: every
+ *  font-size control in the dashboard spans {@link MIN_FONT_SIZE}-{@link
+ *  MAX_FONT_SIZE}, and users compare the sliders side by side. */
+export function FontSizeControl({ label, value, onChange, description, testIdPrefix }: FontSizeControlProps) {
   return (
     <div>
       <label className="block text-[13px] text-text-secondary mb-2">{label}</label>
@@ -23,8 +26,8 @@ export function FontSizeControl({ label, value, min, max, onChange, description,
           type="range"
           aria-label={label}
           data-testid={`${testIdPrefix}-slider`}
-          min={min}
-          max={max}
+          min={MIN_FONT_SIZE}
+          max={MAX_FONT_SIZE}
           step={1}
           value={value}
           onChange={(e) => onChange(Number(e.target.value))}
@@ -37,7 +40,7 @@ export function FontSizeControl({ label, value, min, max, onChange, description,
           onChange={(e) => onChange(Number(e.target.value))}
           className="bg-surface-800 border border-surface-700 rounded-md px-2 py-1 text-sm text-text-primary font-mono w-16 text-center"
         >
-          {options.map((s) => (
+          {OPTIONS.map((s) => (
             <option key={s} value={s}>
               {s}px
             </option>
--- a/web/src/components/settings/FontSizeControl.tsx
+++ b/web/src/components/settings/FontSizeControl.tsx
@@ -1,0 +1,50 @@
+interface FontSizeControlProps {
+  label: string;
+  value: number;
+  min: number;
+  max: number;
+  onChange: (value: number) => void;
+  description?: string;
+  /** Prefix for the slider/select `data-testid`s, so a settings panel with two
+   *  of these controls stays addressable in tests. */
+  testIdPrefix: string;
+}
+
+/** Slider + px select pair shared by the terminal and conversation font-size
+ *  settings. Both controls write the same value so they stay synchronized; the
+ *  select offers every integer step in range. */
+export function FontSizeControl({ label, value, min, max, onChange, description, testIdPrefix }: FontSizeControlProps) {
+  const options = Array.from({ length: max - min + 1 }, (_, i) => min + i);
+  return (
+    <div>
+      <label className="block text-[13px] text-text-secondary mb-2">{label}</label>
+      <div className="flex items-center gap-3">
+        <input
+          type="range"
+          aria-label={label}
+          data-testid={`${testIdPrefix}-slider`}
+          min={min}
+          max={max}
+          step={1}
+          value={value}
+          onChange={(e) => onChange(Number(e.target.value))}
+          className="flex-1 accent-brand-600 h-1.5"
+        />
+        <select
+          aria-label={`${label} (px)`}
+          data-testid={`${testIdPrefix}-select`}
+          value={value}
+          onChange={(e) => onChange(Number(e.target.value))}
+          className="bg-surface-800 border border-surface-700 rounded-md px-2 py-1 text-sm text-text-primary font-mono w-16 text-center"
+        >
+          {options.map((s) => (
+            <option key={s} value={s}>
+              {s}px
+            </option>
+          ))}
+        </select>
+      </div>
+      {description && <p className="text-[11px] text-text-muted mt-1">{description}</p>}
+    </div>
+  );
+}
--- a/web/src/index.css
+++ b/web/src/index.css
@@ -229,14 +229,18 @@ body {
    names). Style them via descendant selectors so the prose looks
    right inside the structured-view conversation. */
 /* The structured-view root publishes both conversation font-size preferences as
-   custom properties; the breakpoint picks the active one in CSS so resizing the
-   window across 768px switches sizes without a reload. */
+   custom properties; CSS picks the active one, so a resize or an orientation
+   change switches sizes without a reload. Desktop is the default and mobile is
+   the override, keyed on the same classifier `clientFormFactor()` uses: coarse
+   primary pointer AND a viewport below the 768px `md` breakpoint. A touch
+   laptop and a narrow desktop window both stay on the desktop size, matching
+   how the rest of the dashboard decides what "mobile" means. */
 .acp-conversation-scope {
-  --acp-conversation-font-size: var(--acp-conversation-font-size-mobile, 14px);
+  --acp-conversation-font-size: var(--acp-conversation-font-size-desktop, 0.875rem);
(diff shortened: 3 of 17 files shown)


web/src/index.css
```diff
@@ -228,6 +228,26 @@ body {
    raw <p>/<ul>/<ol>/<h*>/<blockquote>/<code> elements (no class
    names). Style them via descendant selectors so the prose looks
    right inside the structured-view conversation. */
+/* The structured-view root publishes both conversation font-size preferences as
+   custom properties; the breakpoint picks the active one in CSS so resizing the
+   window across 768px switches sizes without a reload. */
+.acp-conversation-scope {
+  --acp-conversation-font-size: var(--acp-conversation-font-size-mobile, 14px);
+}
+@media (min-width: 768px) {
```
Seluj78: Required: this gates the conversation font size on viewport width only, while the identically-labeled terminal control gates on `pointer: coarse` (`web/src/components/MobileLiveTerminal.tsx:444`) and the repo's canonical classifier uses both (`web/src/lib/formFactor.ts:28`). Three definitions of "mobile" for controls that sit in the same Settings view.

Concrete: on an iPad in landscape (coarse pointer, 1024px) the terminal renders at the mobile size and the transcript at the desktop size, so dragging "Mobile font size" here does nothing while the terminal's does. A desktop window tiled to 700px does the inverse.

`StructuredViewDisplaySettings.tsx:27` and `:37` say "on mobile devices" / "on desktop devices", which is not what this evaluates.

Still pure CSS, no JS width probe:

```css
.acp-conversation-scope {
  --acp-conversation-font-size: var(--acp-conversation-font-size-desktop, 0.875rem);
}
@media (pointer: coarse) and (max-width: 767.98px) {
  .acp-conversation-scope {
    --acp-conversation-font-size: var(--acp-conversation-font-size-mobile, 0.875rem);
  }
}
```

Drop the width clause instead to match the terminal exactly. If width is genuinely the right axis for prose, relabel both controls and their descriptions to say viewport width and update `docs/guides/web/settings.md:31`.

web/src/components/acp/StructuredView.tsx
```diff
@@ -223,17 +224,40 @@ export function structuredViewRootStyle(keyboardHeight: number): React.CSSProper
...
+ *  which is the single normalization boundary for stored prefs. */
+function structuredViewFontStyle(
+  keyboardHeight: number,
+  mobileFontSize: number,
+  desktopFontSize: number,
+): React.CSSProperties {
+  return {
+    ...structuredViewRootStyle(keyboardHeight),
+    "--acp-conversation-font-size-mobile": `${mobileFontSize}px`,
+    "--acp-conversation-font-size-desktop": `${desktopFontSize}px`,
```
Seluj78: Required: publishing absolute px means the transcript stops honoring the browser's default text size.

`Markdown.tsx:71` previously carried `text-sm` (`0.875rem`), and `index.css:178` sets no `html { font-size }`, so the base tracked the browser preference. These two variables are published unconditionally, including at the 14px defaults, so `.acp-conversation-scope` always resolves to px and the `0.875rem` fallback at `index.css:249` is unreachable inside the structured view.

Browser default text size 20px, user never opens AoE settings:
- before: base 17.5px, `h1` 25px
- after: base 14px, `h1` 20px

The users who lose out are the ones who raised their browser font size for accessibility. The PR body says the ratios reproduce the current appearance at the default, which holds only at a 16px root.

Keeps the stored integers, the clamp, and the px labels:

```ts
const toRem = (px: number) => `${px / 16}rem`;
"--acp-conversation-font-size-mobile": toRem(mobileFontSize),
"--acp-conversation-font-size-desktop": toRem(desktopFontSize),
```

If absolute px is deliberate (the terminal is px because a cell is px), worth saying so in the PR body and the settings guide rather than leaving it implicit.

web/src/index.css
```diff
@@ -391,6 +411,28 @@ body {
...
 }
 
+/* Child typography for the opted-in conversation renderer only, in `em` so the
+   whole hierarchy scales with the user's base size. These tie on specificity
+   with the `.acp-markdown <type>` rules above and win on source order, so they
+   must stay after them. Ratios reproduce the `rem` values at the 14px default;
+   the other `.acp-markdown` consumers set their own base (tool cards render at
+   11px) and keep the fixed `rem` sizes. */
+.acp-markdown-body h1 {
+  font-size: 1.43em; /* 20px at the 14px default */
```
Seluj78, line 425: Nit: these tie on specificity with the `.acp-markdown h1/h2/h3/th/td` rules above and win only on source order, which is why the comment has to say "they must stay after them". `.acp-markdown.acp-markdown-body h1` wins on specificity and drops the ordering dependency along with the load-bearing comment.

web/src/components/settings/FontSizeControl.tsx
```diff
@@ -0,0 +1,50 @@
+interface FontSizeControlProps {
+  label: string;
+  value: number;
+  min: number;
+  max: number;
```
Seluj78: Nit: all four call sites pass `MIN_FONT_SIZE` / `MAX_FONT_SIZE` from `lib/fontSizeRange.ts`. The component can import them directly; these props are configurability with no second configuration.

Seluj78 (dismissed):
Drafted by Claude via back-and-forth with @Seluj78.

## AoE Review

Base `main`@`86ccdd16e`, head `f91abbf46`. 17 files (web, web-tests, docs).

Two changes needed, both one-liners inside the shape you already built. The CSS-variable-plus-media-query design is the right one and should stay. Details are inline on the cited lines; summarized here so the review is self-contained.

### Required

**1. `web/src/index.css:237`: conversation "mobile" is width-only, and disagrees with the identically-labeled terminal control.**

Three definitions of "mobile" now coexist:

| Surface | "mobile" condition | Source |
| --- | --- | --- |
| Conversation transcript | viewport width < 768px | `web/src/index.css:237` |
| Live terminal | `pointer: coarse` | `web/src/components/MobileLiveTerminal.tsx:444` |
| Canonical classifier | coarse AND width < 768px | `web/src/lib/formFactor.ts:28` |

Concrete failures:

* iPad in landscape (coarse pointer, 1024px): terminal renders at the mobile size, transcript at the desktop size. Dragging "Mobile font size" under Structured view does nothing to the transcript, while the identically-named terminal slider does work.
* Desktop browser tiled to 700px (fine pointer): the inverse. Transcript jumps to the mobile size, terminal stays desktop.

`StructuredViewDisplaySettings.tsx:27` and `:37` compound it: the descriptions say "on mobile devices" / "on desktop devices" while the gate is pure viewport width.

Fix either way, still pure CSS, no JS width probe:

```css
/* match formFactor.ts */
.acp-conversation-scope {
  --acp-conversation-font-size: var(--acp-conversation-font-size-desktop, 0.875rem);
}
@media (pointer: coarse) and (max-width: 767.98px) {
  .acp-conversation-scope {
    --acp-conversation-font-size: var(--acp-conversation-font-size-mobile, 0.875rem);
  }
}
```

Or drop the width clause to match the terminal exactly. If width really is the intended axis for prose, that is defensible, but then relabel both controls and their descriptions to say viewport width and update `docs/guides/web/settings.md:31`.

**2. `web/src/components/acp/StructuredView.tsx:240-241`: publishing absolute px drops the browser's root font-size preference.**

`Markdown.tsx:71` previously carried Tailwind `text-sm` (`0.875rem`), and `index.css:178` sets no `html { font-size }`, so the transcript tracked the browser's configured default text size. `StructuredViewRoot` now always publishes both variables, including at the 14px defaults, so `.acp-conversation-scope` always resolves to an absolute px value and the `0.875rem` fallback at `index.css:249` is unreachable inside the structured view.

Input: browser default text size 20px, user never opens AoE settings.
Before: transcript base 17.5px, `h1` 25px.
After: 14px and 20px.

The users who lose out are the ones who raised their browser font size for accessibility reasons. The PR body describes the change as reproducing the current appearance at the default, which holds only at a 16px root.

One-line fix, keeps the stored integers, the clamp, and the px labels:

```ts
const toRem = (px: number) => `${px / 16}rem`;
"--acp-conversation-font-size-mobile": toRem(mobileFontSize),
"--acp-conversation-font-size-desktop": toRem(desktopFontSize),
```

If absolute px is deliberate (the terminal is px because a cell is px), say so in the PR body and the settings guide rather than leaving it implicit.

**3. CI has not run on this PR.**

`gh pr checks 3299` returns only `Lint PR title`, `check-template`, and a skipped `close-stale`. `ci.yml` and `tests.yml` were never dispatched, which is normal for a fork PR pending maintainer approval. Lint, tsc, Vitest, and Playwright are unverified by CI. Nothing for you to do here; flagging so it is not mistaken for a green board.

### Nits

* `web/src/index.css:419-433`: the `.acp-markdown-body` child overrides tie on specificity with the `.acp-markdown` rules and win only on source order, which the comment above them has to state explicitly. `.acp-markdown.acp-markdown-body h1` removes the ordering dependency and the load-bearing comment.
* `web/src/components/settings/FontSizeControl.tsx:4-5`: `min` and `max` are props, but all four call sites pass `MIN_FONT_SIZE` / `MAX_FONT_SIZE`. The component can import them directly.

### Checked and clear

Recording these so they are not re-litigated:

* Cascade: every `rem` child font-size under `.acp-markdown` (`h1`, `h2`, `h3`, `th`, `td`) has an `em` counterpart under `.acp-markdown-body`. Complete.
* `em` compounding: inline code is scoped `:where(p, li) > code`, so it never multiplies against `td`'s `0.86em` or a heading's `1.43em`. Code blocks sit directly under the body base. No runaway.
* Non-body `.acp-markdown` consumers (`ToolCards.tsx:1942`, `diff/MarkdownFileView.tsx:21`) sit inside `.acp-conversation-scope` but do not match `.acp-markdown-body`, and the scope only declares a custom property. Appearance unchanged, as the PR claims.
* `SkillsManager` renders `<Markdown>` from `SettingsView`, outside the structured-view root, so it resolves the `0.875rem` fallback. Identical to the `text-sm` it had.
* Security: `normalizeConversationFontSize` accepts only finite numbers or non-blank numeric strings, rounds, and clamps to 6..28 before interpolation as `${n}px`. No CSS injection path from localStorage or the `webUiSync` mirror.
* The 6px floor is a deliberate user choice matching the existing terminal range. Not a regression.

### Metadata

Clean. Template fully filled, PR Type set, AI Usage declares the tool, screenshots for all three size extremes plus the new panel, conventional commit subject, branch naming correct, no em dashes anywhere. Coverage matrix entry added, settings guide updated. Test tiers are well chosen: mocked Playwright for the media-query resolution (jsdom evaluates neither media queries nor custom properties), Vitest for the localStorage contract with a clamping table.

### Verdict

**Project fit:** Accept
**Placement:** core

Closes a real issue, reuses the existing `aoe-web-settings` plumbing instead of touching the agent config schema, and nets out negative on hand-rolled markup by extracting `FontSizeControl`. Fix the two Required items and this is good to land.

summelon:
Hi, @Seluj78.
Thank you for your review.
I updated a fix commit according to your comments.

Additionally, I noticed that in both this PR and previous PR #3275, you and @njbrake found some device/browser-specific corner cases that my own tests did not cover.
For example, in this PR you caught differences between the new responsive logic and the existing terminal/form-factor behavior. In #3275, you also found the collapse handle could conflict with other UI elements like the update banner.

I am trying to improve my test workflow before opening future PRs.
May I know do you mainly find these cases from pure experience, or do you use techniques/skills like adversarial-review, repo-aware-edge-case-review, or counterexample-driven exploratory testing?

Seluj78:
Thanks for addressing the comments @summelon !

Sadly most of those come from just using `aoe`, we don't have a good adversarial review system or anything along those lines.

summelon:
Thank you for your reply @Seluj78.
I suppose I need more frontend experience to detect these subtle issues.

BTW, anything else needed from my action for this PR?
Is the 80% docstring coverage mentioned by CodeRabbit a hard blocker?

Seluj78:
No it's fine, don't worry about that.

summelon:
Hi, @Seluj78 and @njbrake.
I am not sure if there is anything I haven't done that is blocking this PR.
The status shows that @Seluj78 still requests changes from me.
But I should have resolved them in the fix commit.
So I tagged you two again and re-requested a review.
Please kindly advise me if I did anything wrong.
Thank you.

njbrake:
Hi @summelon ! I'll take a look for you, thanks for your patience 🙏

njbrake (approved):
Approving. I pushed one commit to your branch (`13b62a4`) and re-approved the workflow runs.

**What I changed.** Swapping `text-xs` for `text-[0.86em]` on the fenced-code branches kept the font size but dropped the line height that came with it. Tailwind registers `--tw-leading` as `inherits: false`, so the old `text-xs` resolved its own 1.3333 ratio rather than the root's `leading-relaxed`; the em size carries none and inherits 1.625. Measured in Chromium against the built bundle: 12px/16px before, 12.04px/19.565px after, so code blocks rendered about 22% looser. The commit pins the ratio and adds a computed-leading assertion to your Playwright spec, which now renders a fenced block. I confirmed the assertion fails without the fix rather than passing vacuously.

**Everything else checked out.** The `rem` publishing and the coarse-pointer classifier both match `clientFormFactor()`. All five `rem` font-size rules under `.acp-markdown` have `em` counterparts, and `index.css` has not drifted on main since your branch point. `webUiSync` mirrors `aoe-web-settings` as a whole key, so the new fields sync without a registry edit, and the settings search index is schema derived, so dashboard prefs are intentionally not searchable there. No stale mirrors anywhere.

**Two follow-ups, neither blocking.** Tool card markdown stays at 11px, so at a 6px conversation setting the prose ends up smaller than the cards around it; the 6px floor came from matching the terminal range rather than from anything about prose. And the new settings-guide section is the longest on that page for one slider pair, with the closing paragraph about `aoe-web-settings` and `config.toml` being implementation detail a reader does not need.

On your earlier question about how these get found: this one came from asking what a utility class stops doing rather than what it starts doing, then measuring both variants in a real browser instead of trusting the ratio arithmetic.

Nice work on the test tiering, and thanks for your patience on the review.

PR #3300, opened 2026-08-10
fix(session): resume each codex session's own conversation, not a co-located peer's

bjkim95:
## Problem
When several aoe sessions share a workdir and run `codex`, they stop resuming their own conversations. Observed live (aoe 1.14.0): 5 sessions in the same workdir were all launched as `codex resume <one-shared-uuid>`, and a `session stop` + `session start` of one session relaunched it with a *different* session's uuid — it adopted a sibling's conversation.

## Root cause
Codex has no per-instance session-id hook (`CODEX_HOOK_EVENTS` all set `session_id_capture: false`), so its resume target is resolved purely by the mtime scan of `~/.codex/sessions/`: newest rollout matching the workdir, minus an exclusion set. `try_retroactive_capture` built that exclusion from live tmux peers only, so a co-located **stopped / pane-less** codex peer is not excluded and its newest rollout wins. This is the Claude #2355/#2858 steal one tool over: that fix, `compose_exclusion_with_stopped_peers`, was hardcoded to `inst.tool == "claude"` and never applied to codex.

## Fix
Generalize `compose_exclusion_with_stopped_peers` to filter co-located peers by the caller's `tool` (no-op for the existing claude call site), and use it in the codex arm of `try_retroactive_capture`. Each codex session now resumes its own conversation. The steady-state `codex_poll_fn` is intentionally untouched: live peers already exclude each other via the tmux env scan, and adding a `sessions.json` read there would put I/O on the hot polling path (same tradeoff the claude design accepts).

## Test
Adds `codex_mtime_fallback_skips_stopped_peer_sid`, the codex analog of the claude `mtime_fallback_skips_stopped_peer_sid`: with an older own rollout and a newer stopped-peer rollout for the same cwd, `acquire_session_id()` must return the session's own id. Fails on `main`, passes with this fix.

## Summary

- Fixed Codex session resumption when multiple host sessions share a work directory.
- Excluded stopped, archived, or pane-less peer sessions from host Codex rollout selection.
- Preserved sandboxed Codex and existing Claude behavior.
- Added a regression test for newer rollouts from stopped peer sessions.

## Benefits

Codex now resumes the correct session instead of using a newer rollout from an inactive peer. This prevents cross-session conversation mix-ups.


njbrake:
Codex posting on behalf of @njbrake.

Maintainer update: I rebased this PR onto current `main` and force-pushed the contributor branch with `maintainer_can_modify`, using a lease against the previous head.

Since this PR was opened, #3317 landed and gave sandboxed Codex sessions an instance-private `CODEX_HOME`. That fixes the shared-rollout problem for sandboxed sessions more fundamentally, so this PR is now scoped to host Codex sessions, which still share `~/.codex/sessions/`.

The updated patch:

- applies stopped-peer session-ID exclusions only to host Codex sessions;
- keeps sandboxed Codex on the existing private-home capture path;
- preserves the existing Claude behavior;
- renames and updates the regression test to make the host-only scope explicit.

Validation completed:

- all three stopped-peer regression tests
- `git diff --check`

New head: `7c9ed417`. No contributor changes newer than the previous PR head were overwritten; the force push was lease-protected.

njbrake (approved):
Drafted by Claude via back-and-forth with @njbrake.

Reviewed the host Codex stopped-peer capture fix after merging current main and resolving it against the newer persisted peer exclusion logic. The resolution preserves parked conversation protection for every tool, applies inactive same-tool protection to Claude and host Codex, and keeps sandboxed Codex on its instance-private transcript store. The focused regression tests and the full fresh CI suite pass.

PR #3311, opened 2026-08-10
fix(cli): error on duplicate session add

### LINKED ISSUE #3224: aoe add with existing (title, path) silently no-ops with exit 0, reviving the old --cmd-override on session start
## Summary

`aoe add` for a (title, path) pair that already has a session record prints `Session already exists with same title and path: <title>` and **exits 0**, keeping the previously stored `command`. A subsequent `aoe session start <title>` then runs the OLD command. Any script that treats `aoe add`'s exit code as success will silently launch a stale configuration.

Observed on 1.10.0 through 1.13.0 (the duplicate check appears to be `is_duplicate_session()` matching title+path in `src/cli/add.rs`, which returns `Ok(())` after printing).

## Repro

```
aoe add /tmp -t demo --tool pi --cmd-override "echo OLD"
aoe session stop demo            # stop, not rm
aoe add /tmp -t demo --tool pi --cmd-override "echo NEW"; echo "exit=$?"
# prints "Session already exists with same title and path: demo", exit=0
aoe session show demo --json | jq .command
# -> "echo OLD"
```

## Why it bites

Our dispatcher re-dispatches hung agents by title (`stop` then `add` with new `--provider/--model` flags in the command). The exit-0 no-op meant a re-dispatch requesting a different model ran the original model for ~45 minutes before the mismatch was noticed — the only signal was inside the pane. Auto-generated titles also collide over time (we accumulated `Shu V`, `Hindustanis IV`, … from the random-title fallback against never-pruned records).

## Suggested behavior (any of these would fix the scripting hazard)

1. Nonzero exit on duplicate (with the current message), or
2. `--replace` / `--force` flag that updates the stored command in place, or
3. Keep exit 0 but update the stored command when `--cmd-override` differs (add-or-update semantics).

We've worked around it by post-add verification (`aoe session show --json` and comparing `.command`), but the exit code is the natural contract.

Environment: Linux (Debian), tarball installs, `aoe` 1.10.0–1.13.2.

shixi-li:
## Description

Fixes #3224.

`aoe add` previously treated an existing `(title, path)` pair as a successful no-op. That made a conflicting command override look accepted even though the persisted session continued using its old command.

This change returns the same actionable error from both duplicate paths:

- the normal preflight check
- the final check under the storage lock, which covers concurrent adds

Both paths still perform their existing partial-resource cleanup before returning the error, and the existing session remains unchanged.

This intentionally changes duplicate adds from exit status 0 to 1. That user-visible compatibility change matches the behavior requested in #3224: callers can now distinguish a collision from a successful add instead of silently continuing with stale session configuration.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (no documentation change is needed for this error-path correction)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The regression test adds a session with `echo OLD`, retries the same title and path with `echo NEW`, and verifies that the second command exits 1 with an actionable stderr message while the single persisted row still contains `echo OLD`. The test fails on `main` because the duplicate command exits 0, and passes with this change.

Validation performed:

- `cargo test`
- `cargo test --features e2e-tests --test e2e cli::test_cli_add_duplicate_errors_and_preserves_original_command -- --exact --nocapture`

## AI Usage

- [ ] No AI was used
- [x] AI was used

Codex assisted with the bounded issue/duplicate preflight, implementation, regression test, and validation. A separate read-only review checked both duplicate paths, cleanup ordering, stored-session preservation, and the final diff.

## Summary

- Duplicate `aoe add` operations now fail with exit status 1 and provide remediation guidance.
- The check runs before resource creation and during final persistence to handle concurrent adds.
- Existing sessions remain unchanged, and partial resources are still cleaned up.
- Added an end-to-end test for the error, exit status, cleanup, and command preservation.

## Benefits

This prevents duplicate adds from appearing successful and helps callers avoid using stale configuration.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -588,10 +588,6 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
     // seed its directory leaf from it); run the path-dependent duplicate check
     // now that `path` points at the final worktree/workspace directory.
     if is_duplicate_session(&instances, &final_title, path.to_str().unwrap_or("")) {
-        println!(
-            "Session already exists with same title and path: {}",
-            final_title
-        );
         cleanup_partial_session(
             &path,
             worktree_info_opt.as_ref(),
@@ -600,7 +596,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
             None,
             None,
         );
-        return Ok(());
+        return Err(duplicate_session_error(&final_title));
     }
 
     let mut instance = Instance::new(&final_title, path.to_str().unwrap_or(""));
@@ -1065,10 +1061,6 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
     match persist_result {
         Ok(true) => {}
         Ok(false) => {
-            println!(
-                "Session already exists with same title and path: {}",
-                instance.title
-            );
             cleanup_partial_session(
                 &path,
                 instance.worktree_info.as_ref(),
@@ -1081,7 +1073,7 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
                 },
                 instance.sandbox_info.as_ref().map(|_| instance.id.as_str()),
             );
-            return Ok(());
+            return Err(duplicate_session_error(&instance.title));
         }
         Err(e) => {
             cleanup_partial_session(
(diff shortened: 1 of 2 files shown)


njbrake (approved):
Prose by Codex; decisions are njbrake.

Approved. The duplicate-add error behavior and regression coverage were verified locally.

Issue #3265, opened 2026-08-05
Fresh/restarted panes spawn with the daemon's launch cwd instead of the session's project_path (worktree sessions)


adRn-s:
## Summary

For worktree-managed sessions, a freshly-spawned tmux pane's working directory does not match the session's stored `project_path`. Instead, it lands wherever the `aoe` daemon process itself was launched from. This causes Claude's own per-project transcript scoping to diverge from what AoE expects, which manifests as:

- `Error: Session ID <uuid> is already in use.` on restart (AoE decides `--session-id` is safe because it can't find a transcript under the *correct* worktree-encoded project dir — the transcript is actually sitting under the daemon's launch-dir-encoded project dir instead).
- `No conversation found with session ID: <uuid>` printed directly by Claude inside the pane, when using `aoe session set-session-id` to pin a specific conversation and then `aoe session start`.

This reproduces with **no reboot involved** — a plain stop/start cycle on an already-running daemon triggers it.

## Environment

- aoe 1.13.2 (homebrew bottle, `agent-of-empires 1.13.2`)
- Linux (Fedora Atomic-based / uBlue "aurora"), tmux 3.7b
- `$HOME` on one LUKS/btrfs volume (`/var/home`), worktrees on a second LUKS/xfs volume (`/var/mnt/data`) — mentioned in case mount ordering is relevant, though see "What I ruled out" below.
- Claude Code 2.1.220 (`claude` CLI)

## Repro steps

1. Create an AoE worktree-managed session, e.g. a git worktree at `/var/mnt/data/.../some-repo-worktrees/branch`, tool `claude`. Let it run long enough to accumulate real conversation history.
2. Launch the `aoe` TUI/daemon from an **unrelated directory** (e.g. `$HOME`) — this is the normal way most people invoke it.
3. `aoe session stop <id>`
4. `aoe session start <id>` (or `aoe session set-session-id <id> <old-sid>` followed by `aoe session start <id>` to force a specific resume target)
5. Observe: the pane dies almost immediately. `aoe session capture <id>` shows either:
   - `Error: Session ID <uuid> is already in use.`, or
   - `No conversation found with session ID: <uuid>`

Running `aoe session start` from *inside* the correct worktree directory makes no difference — the daemon (not the CLI invocation) does the actual spawning over the existing socket, so the CLI process's own cwd is irrelevant.

## Evidence

[...]

PR #3314, opened 2026-08-10
fix: preserve working directory after login shell startup

adRn-s:
## Description

I traced every spawn call site (CLI `session start`, the TUI's restart/auto-recovery path) and found `project_path`/`working_dir` is threaded through correctly everywhere — `reconcile_from_disk` reloads the right path from disk before every spawn, and `tmux new-session -c <working_dir>` always gets the right value. That part of the code was never the bug.

The actual mechanism, found by building this branch and reproducing #3265 live against a throwaway worktree session:

`wrap_command_ignore_suspend` (`src/session/instance.rs`) runs the real agent command via `bash -lc '...'` — a **login shell**, used deliberately so version-manager PATH setup (NVM, etc.) sourced from the user's profile is available to the agent. But a login shell also sources `~/.bash_profile`/`~/.bashrc`, and **anything in those files that changes directory silently overrides tmux's already-correct `-c`** before the real agent process starts. In my case it was a stray bare `cd` in my own `~/.bashrc` (leftover from disabling a conda-init block) — but the same failure mode fires for anyone with a legitimate nvm/pyenv/direnv hook, or any other profile script that does a `cd`. tmux sets the pane's *initial* cwd correctly; the login shell's own startup then quietly moves it before the agent ever runs.

I confirmed this end-to-end:
1. Built this branch, created a fresh worktree session, and watched its pane land in `$HOME` instead of the worktree — the exact #3265 symptom, on a directory that unquestionably existed.
2. Added a debug log at the tmux spawn site (first commit) and confirmed the `-c` value passed to tmux was **already correct** — the divergence happens strictly after that, inside the shell.
3. `bash -x -lc 'true'` on my own machine traced the exact culprit: `~/.bashrc` line ~52, a bare `cd` with no arguments.
4. Fixed my own `.bashrc`, then instead fixed aoe defensively so this class of bug (anyone's profile scripts, not just mine) can't silently break session cwd again.

`wrap_command_ignore_suspend` now takes `working_dir` and re-asserts it with `cd <dir> || exit 1` as the **first statement inside the login shell's own script**, i.e. after profile/rc sourcing has already run and right before `exec`ing the real agent command. This wins regardless of what the profile did in between, while keeping the intentional benefit of `-l` (PATH setup from version managers). Applied to all three call sites (host agent, host custom command, sandboxed/container).

Also included (first commit, still valid defense-in-depth): `create_with_size_env` now checks the working directory exists before invoking tmux and bails loudly instead of letting tmux's own silent `$HOME` fallback (verified separately: `tmux new-session -c <missing-dir>` succeeds silently) mask a genuinely missing/racing project path, plus a diagnostic `tracing::debug!` of the resolved `-c` value at the spawn site. Not the root cause of the reported symptom, but a real gap with no coverage previously.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered instead by targeted unit tests (`test_create_with_size_env_rejects_missing_working_dir`, `test_wrap_command_reasserts_working_dir_after_login_shell`) plus a manual live repro (built the branch, created/stopped/started a real worktree session, inspected `/proc/<pid>/cwd` before and after the fix) — an e2e test would need to fake a user shell profile with a stray `cd`, which felt like more indirection than the unit tests already covering the actual code path

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Claude Sonnet 5), acting on behalf of the repo/issue owner (myself, adRn-s) in an interactive session.

**Any Additional AI Details you'd like to share:** Root cause was found through actual live reproduction (building the branch, spawning real tmux sessions, tracing `/proc/<pid>/cwd`, and `bash -x -lc` tracing my own shell), not just static code reading — an earlier static-only pass had proposed a plausible but ultimately-wrong theory (a missing-directory tmux fallback) before the live repro surfaced the real mechanism (login-shell profile sourcing). Verifying reviewer feedback and responding here directly rather than relaying it through the AI unexamined.

## Summary

- Validate working directories before creating tmux sessions.
- Return clear errors for missing or invalid directories.
- Restore the working directory after login-shell startup.
- Apply this behavior to host, custom, sandboxed, and container sessions.
- Expand tests for invalid paths, shell escaping, directory changes, and resume behavior.
- Use temporary project directories in integration tests.

## Benefits

- Prevents sessions from starting in `$HOME` or another unexpected directory.
- Makes working-directory errors easier to diagnose.
- Provides consistent behavior across session types.
- Confirms that invalid paths do not create sessions.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -5760,8 +5760,8 @@ const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$(
 /// new-session -c <working_dir>` (see `tmux::Session::create_with_size_env`)
 /// only sets the *shell's* initial cwd; the `-l` login flag below sources
 /// `~/.bash_profile`/`~/.bashrc` (needed for version-manager PATH setup, e.g.
-/// NVM), and anything in those files that changes directory — a user's own
-/// stray `cd`, or a legitimate nvm/pyenv/direnv hook — silently overrides
+/// NVM), and anything in those files that changes directory, such as a user's
+/// own stray `cd` or a legitimate nvm/pyenv/direnv hook, silently overrides
 /// tmux's `-c` before the real agent command ever runs, landing the pane
 /// wherever the profile left it (typically `$HOME`) instead of
 /// `project_path`. This was the actual mechanism behind #3265: `-c` was
@@ -5771,17 +5771,15 @@ const CONTAINER_TERMINAL_AUTODETECT_CMD: &str = r#"sh -c 'exec "$(command -v "$(
 fn wrap_command_ignore_suspend(cmd: &str, working_dir: &str) -> String {
     let user = super::environment::user_shell();
     let posix = super::environment::user_posix_shell();
-    let escaped = cmd.replace('\'', "'\\''");
     // Use login shell (-l) so version-manager PATHs (NVM, etc.) are available.
     // Skip -l when falling back to bash for a non-POSIX user shell (fish, nu,
     // pwsh): bash's login scripts won't contain the user's PATH setup and -l
     // may reset the inherited PATH that already has the correct entries.
     let flag = if user == posix { "-lc" } else { "-c" };
     let cd = super::environment::shell_escape(working_dir);
-    format!(
-        "exec {} {} 'cd {} || exit 1; stty susp undef; exec env {}'",
-        posix, flag, cd, escaped
-    )
+    let script = format!("cd {cd} || exit 1; stty susp undef; exec env {cmd}");
+    let escaped = script.replace('\'', "'\\''");
+    format!("exec {posix} {flag} '{escaped}'")
 }
 
 /// Prepend shell `export` statements to an already-wrapped sandbox command.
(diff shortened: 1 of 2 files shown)


adRn-s:
Sorry it all came down to a stray `cd` line at the bottom of my bashrc. Apparently I have had it for like a month, probably because of my own incompetence, unable to leave vim :P 

But since I had already burnt some tokens I felt like keeping the PR with some "hardening" was due. I leave it up to you, to merge or not. It's not really a bugfix that attains to `aoe`. Though it may help in other cases (ie. hooks from direnv/ nvm/ etc.)

Cheers!

njbrake:
@adRn-s no problem! I have done similar things so many times too 😆 I'll review this one for the changes that are left, thanks 🙏

njbrake (approved):
Drafted by Claude via back-and-forth with @njbrake.

The working-directory quoting defect is fixed in dd7de1c2, including an execution regression test covering spaces and apostrophes. Full cargo test passes locally.

Issue #3310, opened 2026-08-10
Web live terminal: forwarded mouse clicks land one row too low (pointerCell measures from the scroller box, not the rendered grid)


os-tohe:
# Me

In short: I need to position the mouse approximately one row above the actual row, if I want to highlight some text or to click an option. I am using the web dashboard with a terminal view. The same problem occurs for my coworker too, who uses a Mac.

# AI

_The repro below is mine; the source-level diagnosis was drafted with Claude after I described the symptom. I have not compiled or tested the suggested patch._

## Summary

In the web dashboard's live terminal, mouse clicks **are** forwarded to the agent, but the reported cell is off by roughly one row vertically: to hit a target I have to put the pointer about one text row **above** it. Columns are fine. The same clicks work correctly in a real terminal / the TUI.

## Repro

1. Start a Claude Code session in AoE.
2. Get it to show a multi-option prompt (any list you can normally click to select).
3. Open the web dashboard on a desktop browser with a physical mouse and click **directly on** an option.

**Expected:** the option under the pointer is selected.
**Actual:** the click registers about one row lower than the pointer. Aiming one row above the option selects it.

## Environment

- AoE version: `1.14.0`
- OS / browser: `Linux Mint 22.3 / Chrome`
- Agent: Claude Code (alt-screen + mouse tracking on, so `forwardMode` is engaged)

## Diagnosis (from reading `main`)

The click path itself is fine — `onPointerDown` → `buttonMouseBytes` in `web/src/lib/liveMouse.ts`. The problem is the cell mapping.

[`pointerCell`, `web/src/components/MobileLiveTerminal.tsx#L915-L927`](https://github.com/agent-of-empires/agent-of-empires/blob/main/web/src/components/MobileLiveTerminal.tsx#L915-L927):

```ts
const r = el.getBoundingClientRect();          // el = scrollerRef
const row = Math.min(rows, Math.max(1, Math.floor((clientY - r.top) / lineH) + 1));
```

This assumes pane row 1 begins at the scroller box's top edge. In `forwardMode` it doesn't, for two reasons that stack:

1. **The grid is bottom-aligned.** `mt-auto` on the content div ([L1804](https://github.com/agent-of-empires/agent-of-empires/blob/main/web/src/components/MobileLiveTerminal.tsx#L1804)) pushes the rendered rows to the bottom of the box.
[...]

PR #3321, opened 2026-08-11
fix: composite split panes in web terminal

### LINKED ISSUE #3312: Web live terminal captures only pane 0 — agent-team split windows hide the teammate pane (web-side #2068)
# Me

On the web dashboard when a Claude session splits the subagent on the right, the web terminal view only shows the main agent and the right side is completely empty. I would expect to see the subagent on the right in web terminal too.

TUI preview (both panes):

Image

Web dashboard (main agent only):

Image

# AI

_The symptom and screenshots are mine. The source walk below was drafted with Claude after I described the problem; I have not compiled or tested any patch._

## Summary

The web dashboard's live terminal captures `<session>:^.0` and nothing else, so when an agent-team split puts a teammate in a second pane, the web view never shows it.

This is #2068 on the web surface. #2068 was fixed for the TUI preview by #3141, which added a general compositing layer but deliberately scoped the change to the TUI. Its own test-coverage note says so:

> Rust-only; the web live view has its own relay path and is untouched.

## Repro

1. Start a Claude Code session in AoE.
2. From inside that agent, start an agent team / teammate so Claude Code splits the tmux window side by side.
3. Look at the session in the **TUI preview**: both panes are composited with a divider (the #3141 behaviour).
4. Look at the same session in the **web dashboard's live terminal**.

- **Expected:** the same composited window the TUI preview shows.
- **Actual:** pane 0 only. The teammate pane is invisible.

## Environment

- AoE version: `1.14.0` (so #3141 is present — the TUI side is working)
- OS / browser: `Linux Mint 22.3 / Chrome`
- Agent: Claude Code (alt-screen, agent-team split)

## Diagnosis (reading `main` at `86ccdd1`)

The web relay has two frame sources and both are pinned to pane 0.

**Fast path — the shared VT grid.** [`live_ws.rs#L557-L559`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L557-L559) samples the `VtChannel`:

```rust
Some(ch) if ch.is_alive() && lines <= crate::tmux::vt::SCROLLBACK_LINES => {
    let ch = ch.clone();
    match tokio::task::spawn_blocking(move || ch.sample(lines)).await {
```

That channel is armed against pane 0 only — [`vt.rs#L1400`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/vt.rs#L1400):

```rust
let target = format!("{name}:^.0");
```

and the file says as much at [`vt.rs#L812-L814`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/vt.rs#L812-L814): *"The grid is pane 0's alone. `capture_composited_over_grid` sets this when it splices that grid into a composited window."*

**Fallback path — the capture fork.** [`live_ws.rs#L574`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L574) and [`#L594`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L594) call `Session::capture_pane_with_cursor`, which hard-targets pane 0 at [`session.rs#L916`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/session.rs#L916).

**The composite exists, but only the TUI calls it.** `Session::capture_window_composited_with_cursor` ([`session.rs#L660`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/session.rs#L660)) and the VT-spliced variant `capture_composited_over_grid` ([`live_send.rs#L905`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tui/home/live_send.rs#L905)) are reachable only from `src/tui/home/`. `live_ws.rs` never mentions either.

### Two things I could not isolate, offered as questions rather than claims

**Narrow rendering.** Alongside the missing pane, the main agent's output renders narrow with a lot of dead space to its right — the same "thin pane" shape #2068 described. I can't tell which of two paths causes it, because my TUI was previewing the same session at the time:

- the split itself (window resized to the browser viewport, pane 0 only a fraction of it), or
- the non-owner path at [`live_ws.rs#L679-L684`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L679-L684) — *"Another writer (most commonly the TUI's preview sync) can resize the window out from under this viewer... Non-owners render best-effort instead (the client hard-wraps drifted frames)"* — which would look the same with or without a split.

Happy to re-test with the TUI parked elsewhere if that distinction is useful.

**Size-owner convergence.** Predicted from reading the code, not confirmed in logs. When the web viewer *is* the size owner it resizes the **window** ([`live_ws.rs#L658`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L658)) but the drift check compares against **pane 0's** dimensions ([`live_ws.rs#L692-L695`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L692-L695):

```rust
let drifted = want_cols > 0
    && want_rows > 0
    && c.pane_width > 0
    && (c.pane_width != want_cols || c.pane_height != want_rows);
```

On a split window `c.pane_width` can never equal `want_cols`, so this reads as permanent drift — the #2766 shape. The `reassert_guard` added for #2766 should stop it spinning, so I'd expect a stuck target and a `pane drifted from live owner's grid; re-asserting` line rather than a loop. Compositing would resolve it either way: on that path the cursor's `pane_width`/`pane_height` are rebased onto the **window** ([`session.rs#L770-L778`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/session.rs#L770-L778)), which is what `resize_window` sets, so the comparison becomes window-vs-window.

## Suggested fix

Point the web relay at the same compositing path the TUI preview uses, and give the client pane 0's extent so input stays correctly targeted.

1. **Frame source.** In the fetch arm of `live_ws.rs`, take the composite when the window has more than one pane: `capture_composited_over_grid` when a `VtChannel` is held, `capture_window_composited_with_cursor` otherwise. Both already fall back to the pane-0 bytes on any failure, and both keep the unsplit case on the existing single-fork path, so the common session pays nothing. This probably means lifting `capture_composited_over_grid` and its layout cache out of `src/tui/home/live_send.rs` into something both surfaces can call — the helper is already transport-shaped and takes its cache by `&mut`.

2. **Wire format.** `frame_json` ([`live_ws.rs#L1145`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/server/live_ws.rs#L1145)) currently drops `PaneCursor::composite_pane0`. That field exists precisely for this — [`session.rs#L775-L778`](https://github.com/agent-of-empires/agent-of-empires/blob/main/src/tmux/session.rs#L775-L778): *"Rebasing the frame onto the window erases the only record of how wide the input pane is, which mouse forwarding maps into. Carry pane 0's extent alongside."* Emitting it (e.g. `"pane0": {"cols":..,"rows":..}`) lets the client clamp mouse forwarding to pane 0's rectangle. Without it, `rows` becomes the window height and the client maps clicks against the wrong frame.

3. **Drift check.** Compare `want_cols`/`want_rows` against the window rather than pane 0 when a composite is in play — which falls out of (1) automatically, since the rebased cursor already reports window dimensions.

4. **Input routing unchanged.** Keystrokes stay pinned to `^.0` exactly as today (#435, #488). Compositing is read-only, so the mis-targeted-keystroke hazard the pin guards against doesn't arise — the same argument #3141 made for the TUI.

### Limits that carry over from #3141

- **Splits lose scrollback.** The composite covers the visible window only, so `history` would be 0 while the window is split. The web client's scrollback affordance would need to handle that the way the TUI preview does (clamp rather than misbehave).
- **A zoomed pane (`C-b z`) is treated as unsplit** and takes the single-pane path.
- **Wheel/mouse forwarding still goes to pane 0** regardless of which pane the pointer is over. On the web that's more visible than in the TUI preview, since the web view is interactive — which is why item (2) matters. It also overlaps with #3310, where the pointer row mapping is already off by one; the two touch the same mapping.

njbrake:
## Description

Fixes #3312.

The web live terminal captured only tmux pane 0, while the native TUI used the existing whole-window compositor. Claude agent-team splits therefore left the teammate pane invisible in the browser.

This changes the web relay to use `capture_window_composited_with_cursor`, sends pane 0's input extent in composite frames, and clamps forwarded browser mouse coordinates to pane 0. Keyboard and mouse input remain routed to pane 0.

### Tradeoffs

- Split windows show only the visible pane contents and have no combined scrollback. Each tmux pane owns independent history, so there is no coherent timeline to merge. This matches the native TUI behavior.
- Split captures cost more than normal captures because tmux must capture and compose multiple panes. The common single-pane path remains the existing low-cost path.
- Secondary panes are view-only. Keyboard input stays pinned to pane 0 for safety, and mouse coordinates are clamped to pane 0.
- Zoomed tmux panes are treated as unsplit because tmux reports overlapping rectangles.
- The wire protocol gains an optional `pane0` field. Older clients ignore it, and newer clients fall back safely when it is absent.
- Window-size drift checks now use the composite window geometry for split frames, avoiding a permanent pane-width mismatch.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

The relevant tests pass. The full Rust suite completed with 6081 passed, 2 ignored, and 2 unrelated fixture failures where the test setup did not force the expected registry load error:

- `acp::supervisor::tests::shutdown_and_wait_degrades_when_load_errs_without_peer`
- `process::worker_registry::tests::pid_source_for_falls_back_to_peer_pid_on_load_err`

No screenshot is included because reproducing this requires a live Claude agent-team split. The issue reporter has been asked to test the draft.

## Test Coverage Analysis

- [x] Skipped a test, because: the capture behavior is covered by the existing Rust tmux compositing tests, while Vitest covers frame parsing and pane-0 mouse clamping without the cost and nondeterminism of launching a real Claude agent team. The coverage matrix was updated.

**TUI / CLI (e2e test)**

## Validation

- `cargo fmt --check`
- Focused Rust frame protocol test
- `cargo clippy --features serve --lib -- -D warnings`
- `cd web && npm run format:check`
- `cd web && npm run lint`
- `cd web && npx tsc -b`
- 33 focused Vitest tests
- Web coverage matrix validation
- Full `cargo test --features serve`, with the two unrelated failures noted above

All-target clippy also reports existing warnings in unrelated test code under `src/acp/`, `src/session/`, and `src/tui/`. Production library clippy is clean.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Codex investigated issue #3312, implemented the fix, and ran the validation listed above.

## Summary

- Web live terminals now show all visible panes in tmux split sessions.
- The protocol now includes optional pane 0 dimensions for input handling.
- Mouse and touch input stay targeted to pane 0.
- Single-pane and zoomed-pane captures keep their existing behavior.
- Cursor handling is more reliable during composited captures.
- Tests and coverage checks now validate split-pane rendering and input behavior.

## Benefits

Claude agent-team sessions are now visible in the web terminal. Pane resizing behaves correctly, while unsplit sessions retain the efficient capture path.

## Follow-up considerations

The main-agent cursor can appear one row too high. Splitter alignment at the bottom also needs further review.


njbrake:
@os-tohe, would you be willing to test this draft against the Claude agent-team split from #3312? It now composites every visible tmux pane in the web terminal while keeping input pinned to pane 0.

The main tradeoffs are documented in the PR: split windows have visible-window content only with no combined scrollback, secondary panes remain view-only, split capture is more expensive than the normal single-pane path, and zoomed panes still use the unsplit capture path. In particular, it would be helpful to confirm both panes render at the expected width and that the web size owner remains stable.

os-tohe:
I tested this, it now displays the sub-agents on the side. It is not perfect, but a lot better than previously. Here is a picture of it:
test
The main agent visible cursor position is one row too high, but it might be related to this that was already fixed: https://github.com/agent-of-empires/agent-of-empires/issues/3310
Also the splitter lines at the bottom are not perfectly aligned. The width is distributed correctly when I change the width of the window and everything else seems stable.

PR #3324, opened 2026-08-11
fix(acp): scope provider credentials by adapter

### LINKED ISSUE #3238: Non-Claude agent keys are stripped before reaching structured-view agents
Structured-view (ACP) agent spawns clear the environment and then forward a fixed allowlist, `ALWAYS_FORWARD_ENV` in `src/acp/acp_client.rs`:

```
PATH, HOME, XDG_CONFIG_HOME, LANG, LC_ALL, TERM, USER, SSH_AUTH_SOCK,
ANTHROPIC_API_KEY, ANTHROPIC_AUTH_TOKEN, CLAUDE_CODE_OAUTH_TOKEN, CLAUDE_CONFIG_DIR
```

So `OPENAI_API_KEY`, `GEMINI_API_KEY`, and `OPENROUTER_API_KEY` set in the environment that runs `aoe serve` reach terminal-view sessions but are dropped from `codex-acp`, `pi-acp`, and any other non-Claude ACP adapter. Only Claude's keys are forwarded.

Surfaced from the CityHall side. CityHall now stores per-user agent credentials and injects them into a workspace container (agent-of-empires/cityhall#16), which works for Claude and is a no-op for the other three in structured view. CityHall marks those entries as terminal-only in its UI, which is honest but not a fix.

Possible directions, in rough order of preference:

1. Add the other provider keys to `ALWAYS_FORWARD_ENV`. Smallest change, though it grows a hardcoded list every time a provider appears.
2. Derive the forwarded set from the agent being spawned, so each adapter declares which variables it needs. More precise than one global list, and it stops an unrelated agent seeing a key it has no use for.
3. Let `Config.environment` reach natively-spawned agents through the CityHall bundle. Today the bundle's `[settings]` path drops that field: `retain_schema_fields` keeps only leaves whose section maps to an object in `schema()`, and root `environment` is an array, so it is discarded. This would give operators a general mechanism rather than a per-key allowlist.

Whichever direction, a companion would help CityHall stop guessing: a way to ask a given aoe build which variables it forwards, so a control plane can report the real capability instead of a static compatibility table that goes stale.

Acceptance: an `OPENAI_API_KEY` present in the environment running `aoe serve` authenticates a structured-view Codex session.

shixi-li:
## Description

Structured-view agents start from an empty subprocess environment and then restore an explicit allowlist. The previous implementation forwarded provider credentials too broadly and also injected ambient provider credentials into sandboxed agents, bypassing the existing `sandbox.environment` authorization boundary.

This revision scopes ambient host credentials to the actual ACP adapter selected by `config.agent_key`:

- Claude adapters receive the existing Anthropic and Claude credentials.
- Codex receives `OPENAI_API_KEY`.
- Gemini receives `GEMINI_API_KEY`.
- OpenCode receives `OPENROUTER_API_KEY`.
- Custom and unrelated adapters receive no ambient provider credentials by default.

Both host spawn paths use the same mapping. Sandboxed agents no longer receive any ambient provider credentials; credentials must instead be authorized through `sandbox.environment` / `extra_env` or supplied explicitly through `provider_env`. Secret values remain outside Docker arguments, and `AOE_TOKEN` remains excluded.

A deterministic table-driven regression covers the positive and negative adapter matrix, custom allowlisting, sandbox runner isolation, explicit sandbox authorization, and secret-in-argv boundaries. The structured-view documentation now describes the same least-privilege contract.

Fixes #3238

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (N/A: no UI changes)

## Test Coverage Analysis

- [x] Skipped a test, because: this is a subprocess environment boundary with no browser interaction; a deterministic Rust regression exercises the host and sandbox authorization policies

**TUI / CLI (e2e test)**

## Validation

- `cargo test --features serve --lib provider_auth_env_is_scoped_to_agent_and_sandbox_authorization -- --nocapture`
- `cargo test --features serve --lib build_sandbox_docker_argv_ -- --nocapture`
- `cargo test --features serve` (6,108 library tests plus integration targets; 0 failures)
- `cargo clippy --all-targets --features serve -- -D warnings -A clippy::collapsible_match -A clippy::type_complexity -A clippy::items_after_test_module -A clippy::useless_vec`
- `git diff --check`

The four Clippy allowances correspond only to Rust 1.95 warnings in files unchanged from `main`; strict `-D warnings` was also used to identify and verify that baseline.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Codex assisted with issue analysis, implementation, regression-test design, and validation. The final diff and evidence were independently reviewed before publication.

**NOTE:**

## Summary

- Forwarded provider-specific API keys to structured-view agents for OpenAI, Gemini, and OpenRouter.
- Applied the same credential rules to host and sandbox agents.
- Kept secret values out of Docker arguments and excluded `AOE_TOKEN`.
- Added regression tests for provider access, custom agent restrictions, sandbox behavior, and secret handling.
- Updated documentation to explain credential forwarding and allowlisting.
- Benefits: Non-Claude structured-view sessions can authenticate while credentials remain protected.

## Technical details

- Added deterministic Rust coverage for all supported providers and spawn paths.
- Preserved explicit environment precedence and parent-environment inheritance.
- Validation included targeted and full Rust tests, formatting, Clippy, and diff checks.

### DIFF
--- a/docs/structured-view.md
+++ b/docs/structured-view.md
@@ -34,7 +34,7 @@ The `npm install -g` commands above are optional: `aoe acp doctor --fix` install
 
 Tools not yet wired into the registry (aider, cursor, copilot, droid, hermes, kiro) always run in the terminal view. A **custom agent** can opt in by setting an ACP launch command via `agent_acp_cmd` (see [Configuration](guides/configuration.md#running-a-custom-agent-in-the-structured-view)).
 
-The structured view always forwards `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, and `CLAUDE_CONFIG_DIR` to the agent. For other agents, set their auth env in the environment that runs `aoe serve` (or the per-session `extra_env` field).
+For host sessions, each built-in adapter receives only its matching ambient credentials: Claude receives `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, and `CLAUDE_CONFIG_DIR`; Codex receives `OPENAI_API_KEY`; Gemini receives `GEMINI_API_KEY`; and OpenCode receives `OPENROUTER_API_KEY`. Custom adapters require an explicit `env_allowlist` or per-session credential. Sandboxed sessions do not inherit ambient credentials; list them under `sandbox.environment` or provide a per-session credential instead.
 
 ### Feature matrix
 
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3806,6 +3806,9 @@ fn build_sandbox_docker_argv(
         "ANTHROPIC_API_KEY",
         "ANTHROPIC_AUTH_TOKEN",
         "CLAUDE_CODE_OAUTH_TOKEN",
+        "OPENAI_API_KEY",
+        "GEMINI_API_KEY",
+        "OPENROUTER_API_KEY",
     ];
     for &key in PROVIDER_AUTH_KEYS {
         if seen_keys.contains(key) {
@@ -3881,6 +3884,9 @@ const ALWAYS_FORWARD_ENV: &[&str] = &[
     "ANTHROPIC_AUTH_TOKEN",
     "CLAUDE_CODE_OAUTH_TOKEN",
     "CLAUDE_CONFIG_DIR",
+    "OPENAI_API_KEY",
+    "GEMINI_API_KEY",
+    "OPENROUTER_API_KEY",
 ];
 
 /// The inherited host environment layer for a structured-view agent, applied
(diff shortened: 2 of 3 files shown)


njbrake (changes requested):
drafted by Claude via back-and-forth with @njbrake

The host-side authentication bug is real, but this implementation introduces an unnecessary credential exposure and should be narrowed before landing.

Blocking findings:

1. Provider credentials are forwarded to unrelated agents. The forwarding loops do not inspect `config.agent_key`. The new test constructs a Claude spawn configuration and then asserts that OpenAI, Gemini, and OpenRouter keys all reach it. An arbitrary custom ACP adapter would receive the same ambient keys.

2. The sandbox path bypasses the documented opt-in boundary. Sandbox internals say that environment values other than terminal/color variables come from explicit `sandbox.environment`, and specifically tell users to list API keys there. This change injects three additional host secrets even when the user did not authorize them for the container. A compromised agent can read and exfiltrate those keys without escaping the container.

3. Avoiding secret values in Docker argv is good, but it only protects process listings. It does not address disclosure to the wrong agent or container.

Please scope built-in adapters to only the provider credentials they require. Custom adapters should require `env_allowlist`, `provider_env`, or configured environment. For sandboxed sessions, preserve `sandbox.environment` as the authorization boundary unless a per-session credential is explicitly supplied. Add negative matrix assertions proving unrelated provider keys and omitted sandbox keys remain absent.

The existing automatic Anthropic sandbox forwarding appears inconsistent with the same documented contract; it should be reconciled rather than used as precedent for expanding the exposure.

shixi-li:
Addressed all three blocking findings in `5258b62d`:

- built-in adapters now receive only their own provider credentials; custom adapters receive no ambient provider credentials unless explicitly allowlisted/configured
- sandboxed sessions preserve `sandbox.environment` as the authorization boundary and do not inherit ambient provider credentials
- table-driven negative coverage verifies unrelated provider keys, custom-agent keys, omitted sandbox keys, `AOE_TOKEN`, and Docker argv remain absent

The same scoping is applied to detached and in-process spawn paths, and `docs/structured-view.md` documents the behavior. The current head is mergeable; its fork workflows still require maintainer approval. @njbrake, could you re-review the scoped follow-up and approve the workflows when convenient?

njbrake (changes requested):
Two things to change, both in `agent_host_env_keys`.

**`aoe-agent` now receives no credentials at all.** It falls to the `_ => &[]` arm, but it is a real registry entry, it is the bundled Vercel AI SDK agent (`pickModel` routes to `@ai-sdk/anthropic` / `openai` / `google`, defaulting to Anthropic), and it is the step 4 fallback `pick_agent_for_tool` returns for any tool without a registry entry. On main it got `ANTHROPIC_API_KEY` from `ALWAYS_FORWARD_ENV`. Adding `("aoe-agent", &["ANTHROPIC_API_KEY"])` to your own test table fails with the whole environment printed: `TERM`, `HOME`, `PATH`, `XDG_*`, and nothing else.

**Add each adapter's provider key, but only drop an inherited one where the adapter provably cannot use it.** My earlier review objected to widening: ambient OpenAI/Gemini/OpenRouter keys reaching arbitrary adapters, and the sandbox bypassing `sandbox.environment`. Narrowing what built-ins already had is a separate change, and it is where these regressions come from. `opencode` is multi-provider and now sees only `OPENROUTER_API_KEY`, so anyone running it against Anthropic from a shell-exported key breaks. `vibe`, `pi`, `omp`, and `kimi` get nothing, and their rows in `docs/structured-view.md` say "provider env". `codex` and `gemini` are fine to scope tightly. For the rest, keep what main forwarded and add their own key on top. Custom and unknown adapters receiving nothing is the boundary worth keeping, and none of this touches it.

Two smaller items:

`docs/structured-view.md` says custom adapters "require an explicit `env_allowlist`". No config surface populates that field: `from_acp_cmd` hardcodes `None`, the registry is only ever `with_defaults()`, and the dashboard never sends `provider_env`. The knob that works is the top-level `environment` list, or `session.inherit_host_environment`. The new test sets `spec.env_allowlist` directly, so it passes for a path users cannot reach.

The sandbox change is right, and it resolves a contradiction with `docs/development/internals/sandbox.md`. It is also breaking for anyone whose sandboxed Claude session authenticated off an ambient `ANTHROPIC_API_KEY`. Please note that in the description.

njbrake:
Closing this in favor of #3360, which fixes #3238 from the other direction. Thank you for the two rounds here, and sorry to land it this way after the back and forth; the analysis was solid and the sandbox point moved my thinking.

The deciding difference is that #3360 is purely additive. It leaves `ALWAYS_FORWARD_ENV` alone and fills the per-adapter `AgentSpec.env_allowlist` that both spawn paths already consume, so nothing an adapter receives today goes away. This PR reaches the same place by narrowing: Claude keys leave the shared list, ambient credentials stop crossing into containers, and `agent_host_env_keys` becomes a second per-adapter mechanism beside `env_allowlist`. Each piece is defensible, but together they change behavior for people running today, including a sandboxed Claude session with an exported `ANTHROPIC_API_KEY`, a custom `agent_acp_cmd` adapter, and opencode against Anthropic. I don't want a fix for missing credentials to carry a credential removal in the same commit.

The least-privilege argument is worth keeping. Scoping the Claude keys to Claude adapters, and making `sandbox.environment` the only ambient path into a container, is a change I would take on its own with its own release note. If you want to open that once #3360 lands, I will review it.

Issue #2708, opened 2026-07-07
bug(session): same-cwd clobbering recurs on v1.12.0 — stopped sessions (incl. explicit set-session-id pins) inherit a live peer's conversation within seconds


simon-spektrum:
## Summary

The same-cwd session-id clobbering family (#1522, #2344) still reproduces on the released **v1.12.0** (Homebrew), which should contain the #2344 fix set (#2345, #2481, #2361 all merged before the v1.12.0 tag). Two distinct symptoms observed over ~24h of fleet use:

1. **Fleet-wide cross-assignment drift.** With ~10 `tool=claude` sessions sharing one `project_path` (a monorepo root), 9 of 17 sessions' `agent_session_id` values collapsed onto just 2 conversation UUIDs over a day of normal use (start/stop/wake cycles). Every affected session resumed the wrong conversation after the morning wake. Worktree sessions (unique cwd) were never affected.

2. **Explicit `set-session-id` pins on stopped sessions are overwritten within seconds.** After repairing the fleet with `aoe session set-session-id`, any session left **stopped** had its freshly pinned id reverted to a *live same-cwd peer's* conversation id within ~3–10 seconds. This reproduces **with the serve daemon stopped** (TUI open), so it is not a daemon race. Sessions that were started immediately after pinning kept the correct id — i.e. launching wins the race, staying stopped loses it.

## Environment

- aoe `1.12.0` (Homebrew bottle), macOS (darwin 25.5.0, Apple Silicon), tmux 3.7b
- agent = `claude` (Claude Code 2.1.201), `session.agent_status_hooks = true`
- TUI open most of the day + `aoe serve --daemon`; 17 sessions, 10 sharing `project_path` = one repo root; the rest in worktrees

## Repro (symptom 2, deterministic on my machine)

1. Have ≥2 claude sessions with the same `project_path`; keep at least one **running** (actively used), leave another **stopped**.
2. `aoe session set-session-id "<stopped-session>" <its-correct-conversation-uuid>` — CLI confirms.
3. Read `profiles/main/sessions.json` a few seconds later: the stopped session's `agent_session_id` now equals the **running** peer's conversation id (or another recently-touched transcript in the shared encoded-cwd dir).
4. Repeat step 2 followed immediately by `aoe session start "<stopped-session>"` — the pin sticks and survives.

## Evidence (debug.log, UTC)

Two different session records assigned the *same* UUID 6 ms apart (the UUID is a third session's live conversation):

```
[...]

PR #3325, opened 2026-08-11
fix(session): scope pi session-id fallback to the target project

athal7:
capture_pi_family_session_id's third fallback — used when every JSONL
header fails to parse, so the id has to come from a filename — picked the
most recently modified session directory in the whole sessions root,
regardless of which project it belonged to. With sessions open across many
repos, "most recently modified anywhere" is almost never the project being
resumed, so a revived session could attach to a different project's
conversation entirely.

Restrict the scan to directories whose encoded name matches the target
project path, keeping the mtime sort within that set (the directory itself
can legitimately be recreated). When nothing matches, return an error
rather than falling back to an arbitrary directory: no session id is a
recoverable condition, the wrong one is not.

`test_capture_pi_session_id_fallback_by_dir_mtime` now pins that a
non-matching directory holding a NEWER session is ignored in favour of the
matching one, and `test_capture_pi_session_id_fallback_no_match_returns_error`
covers the empty case. Both were reverified against the pre-fix production
logic (temporarily reverted locally) to confirm they fail without the fix,
then restored.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: this only touches `capture_pi_family_session_id`'s in-process fallback, not TUI rendering, a CLI subcommand, or session lifecycle; the new unit tests (`test_capture_pi_session_id_fallback_by_dir_mtime`, `test_capture_pi_session_id_fallback_no_match_returns_error`) exercise the fallback directly and cheaper than an e2e test.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Anthropic, Claude Sonnet)

Fix, tests, and review-feedback follow-ups were authored by an AI coding agent. Verification included reverting the pre-fix fallback logic locally to confirm both new tests fail without the scoping fix, then restoring it.

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -615,25 +615,32 @@ fn capture_pi_family_session_id(
         return Ok(id.clone());
     }
 
-    // Third fallback: when all JSONL headers fail to parse, pick the most
-    // recently modified session directory and extract a UUID from its files.
-    let mut dirs_by_mtime: Vec<(PathBuf, std::time::SystemTime)> = Vec::new();
+    // Third fallback: when JSONL headers fail to parse (no `id` field),
+    // extract a UUID from the filename. Only consider directories whose
+    // encoded name matches the target project path, so we never grab a
+    // session from the wrong project.
+    let mut project_dirs: Vec<PathBuf> = Vec::new();
     if let Ok(entries) = resilient_read_dir(&sessions_dir) {
         for entry in entries {
             let path = entry.path();
             if !path.is_dir() {
                 continue;
             }
-            let mtime = entry
-                .metadata()
-                .and_then(|m| m.modified())
-                .unwrap_or(std::time::SystemTime::UNIX_EPOCH);
-            dirs_by_mtime.push((path, mtime));
+            if path.file_name().and_then(|n| n.to_str()) == Some(&encoded_name) {
+                project_dirs.push(path);
+            }
         }
     }
-    dirs_by_mtime.sort_by_key(|c| std::cmp::Reverse(c.1));
+    // Sort by mtime descending so we pick the newest project directory
+    // (handles the case where the directory itself was recently recreated).
+    project_dirs.sort_by_key(|d| {
+        d.metadata()
+            .and_then(|m| m.modified())
+            .unwrap_or(std::time::SystemTime::UNIX_EPOCH)
+    });
+    project_dirs.reverse();
 
-    for (dir, _) in &dirs_by_mtime {
+    for dir in &project_dirs {
         if let Ok(entries) = resilient_read_dir(dir) {
             let mut file_candidates: Vec<(String, std::time::SystemTime)> = Vec::new();
             for entry in entries {
@@ -3494,6 +3501,9 @@ mod tests {
         }
     }
 
+    /// Third fallback: when JSONL headers fail to parse, extract a UUID from
+    /// the filename. Only consider directories whose encoded name matches the
+    /// target project path, so we never grab a session from the wrong project.
     #[test]
     #[serial]
     fn test_capture_pi_session_id_fallback_by_dir_mtime() {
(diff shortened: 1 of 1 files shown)


athal7:
Moving to draft so this doesn't take reviewer time yet — not because the change is wrong, but because I can't currently justify its priority.

I opened it believing it fixed a symptom it does not. It was written to address omp sessions reviving in the wrong directory. After installing a local build carrying this patch, I restarted two omp sessions and both still came up in the wrong working directory (one in an unrelated worktree, one in a parent repo), while two opencode sessions in the same batch resolved correctly. So the wrong-cwd behaviour lives on a different code path — `capture_pi_family_session_id` captures a session id, it doesn't select the pane's working directory.

The change still looks correct in isolation: that third fallback scans the entire sessions root and sorts by mtime, so with sessions open across many projects it can return an id belonging to a different project than the one being resumed. The added tests pin that a non-matching directory holding a newer session is ignored, and that no match returns an error rather than an arbitrary directory.

What I can't yet supply is a reproduction of user-visible breakage traceable to *this* path, as opposed to the reasoning above plus the tests. Rather than have someone review it on my say-so that it fixes something it doesn't, I'd rather find the real cause of the wrong-cwd bug first — that may well land here, or may show this is better folded into a different change. Happy to reopen for review as-is if you'd take it purely as a correctness/robustness fix to the fallback.

athal7:
Retracting my previous comment — I was wrong, and this PR does fix the bug I originally opened it for. Marking it ready for review again.

**Why I thought it didn't.** After installing a build with this patch, omp sessions still came up attached to another project's conversation, so I concluded the behaviour lived on a different code path. What I missed: this patch prevents a *bad capture*, but it does nothing about ids the pre-patch binary had **already persisted**. Every session in my profile was carrying a stale, cross-wired `agent_session_id`, so the patched binary faithfully resumed the wrong conversation it had been told to resume. Code fix and data repair are two separate problems; I conflated them.

**Evidence.** omp writes each conversation to `~/.omp/agent/sessions/<cwd-derived-dir>/<ts>_<sid>.jsonl`, so the directory a jsonl sits in is authoritative proof of the cwd that created it. Cross-referencing every omp session's stored `agent_session_id` against the directory its jsonl actually lives in:

```
11 of 14 omp sessions miswired.
```

The mis-assignments form a near-permutation — session A pinned to B's conversation, B to C's, C back into a shared parent-repo directory — which is the signature of a scan that picks "newest jsonl anywhere" rather than "newest jsonl for this project". Two examples where the same repo has both a parent checkout and worktrees: two distinct worktree sessions both resolved to the *parent* repo's directory, and two others resolved to a third, unrelated worktree.

Critically, the two sessions I had eyeballed manually and used as my "disproof" were both predicted exactly by the stored-id cross-reference: each came up in precisely the directory its stale id pointed at. So the observation that made me draft this PR was never evidence against it — it was the bug's expected output given already-poisoned state.

**Repair, for anyone who hits this.** `aoe session set-session-id <id> <sid>` is the clean fix, and its `Use -> Default` promotion-at-launch semantics make it exactly right for a one-shot correction — no manual `sessions.json` editing, and the pin self-clears. Repointing each session at the newest jsonl in its *own* cwd-derived directory took all 14 to OK.

One note on the existing defenses in `sync.rs`: the `sid_owners` "never adopt an id another instance already owns" guard (#2708) would have rejected most of these assignments, but it only covers poller-reported sids — the capture-time fallback path this PR patches writes straight past it. If it's worth hardening further, applying the same ownership check at the capture boundary would make the two paths consistent.

njbrake (approved):
> Drafted by Claude via back-and-forth with @njbrake.

Reviewed the current head. The filename fallback is now scoped to the target project directory, preserves exclusion filtering, and returns an error instead of adopting an unrelated session. `cargo fmt --all --check` and all 125 `session::capture::tests` passed locally.

Issue #3316, opened 2026-08-10
Heads-up: building an optional "memory pressure" strip in the TUI


MatthewWolff:
# Heads-up: building an optional "memory pressure" strip in the TUI

Opening this to flag what I am about to build and to give anyone a chance to steer it before I put up a PR. I am planning to build it either way, but I would rather hear objections now than in review, especially on the dependency question below.

## What I am building

When I run several agent sessions at once, I sometimes stack up enough of them that the machine runs low on memory and starts to thrash, and by the time I notice it is already too late to back off. `aoe` currently shows nothing about host memory, so I am blind on the exact resource that takes the box down.

So: a small, optional readout inside `aoe` that shows how close memory is to the edge and how many agents/processes are running, so I can decide for myself whether to start another session or stop some. It is a gauge, not an advisor: it shows state, I draw the conclusions.

A toggle-able strip docked at the bottom of the home view, off by default:

```
┌ memory ───────────────────────────────────────────────────┐
│ ▂▃▄▅▅▆▇▇██▇▇▆▅▅▄▃▃  64% 22.7/32G  │  7 agents · 43 procs  │
└───────────────────────────────────────────────────────────┘
```

- **Left half:** a memory-over-time area sparkline, colored green / amber / red.
- **Right half:** the current memory figure as both a percentage and used/total (e.g. `64% 22.7/32G`, easy to read at a glance), plus running agent sessions and total processes in those agent trees.

Toggled by a keybinding and a persisted setting so it survives restarts.

## Scope

One PR, tightly scoped on purpose: Linux + macOS, the strip only, TUI-only (no web dashboard surface), with an e2e test for the toggle. The breakdown/drill-down, a per-process table, and history persistence are deliberately left out (not deferred to a follow-up): the point is a single glanceable gauge, not a mini system monitor.

## The one thing worth flagging before I start: a dependency

For cross-platform memory sampling I am leaning toward the `sysinfo` crate (`default-features = false, features = ["system"]`) for system RAM and per-process memory on Linux and macOS. What it costs, stated honestly: only `libc` is already in the tree unconditionally. On macOS `sysinfo` would add `objc2-core-foundation` and `objc2-io-kit`, which are **not** present today (the `objc2` currently in the lockfile comes in only through the `serve`-gated `webbrowser` dep, so a TUI-only build has none of it); on Windows it adds `ntapi`.

[...]

PR #3326, opened 2026-08-11
feat(tui): add optional memory diagnostics strip

MatthewWolff:
## Description

Adds an optional memory-diagnostics strip to the plain TUI so it is possible to see, at a glance, how close the machine is to a memory thrash and how many agents are driving it, before it becomes a problem. This is aimed at the case where several agent sessions run at once on one box (often watched over SSH) and memory quietly fills until things lock up, with nothing in `aoe` surfacing it.

The strip docks under the session-list column, off by default, toggled by F9 and a persisted setting (`session.show_diagnostics_pane`). It shows two stacked rows: a memory-over-time sparkline on top (colored green/amber/red), and a readout row with the current usage (percent plus used/total) and the running agent and process counts.

The pressure line is memory headroom, `1 - MemAvailable/MemTotal` on Linux (`MemAvailable` already excludes unreclaimable tmpfs/shmem, so it is the number that actually moves when a RAM-backed `/tmp` fills), with PSI (`/proc/pressure/{memory,io}`) folded into the color band only, not the plotted line. macOS uses the native memory-pressure level (`kern.memorystatus_vm_pressure_level`). Sampling runs every second on a background worker (modeled on the existing status poller), never on the render thread, into a fixed ring buffer. The agent and process counts reuse the existing session-liveness check and a single process-tree walk, scoped to AoE-managed sessions.

Memory sampling is hand-rolled from `/proc` and `/sys` on Linux and `sysctl`/`vm_stat` on macOS, so the change adds no new dependency. It is TUI-only; there is no web-dashboard surface in this PR. Discussed on #3316 (build it built-in first, then use it to grow a plugin UI-extension mechanism later).

### Screenshot

Real screen capture from the e2e harness with the strip toggled on (empty test environment, so the counts read zero and the sparkline has no samples yet). It docks under the session-list column on the left; the preview keeps full height:

```
╭ aoe ────────────────────────── « ╭ Preview ────────────────────────── hide info with i ╮
│                                  │            Select a session to preview               │
│          No sessions yet         │                                                      │
│      Press 'n' to create one     │                                                      │
│          or 'aoe add .'          │                                                      │
│                    ...           │                    ...                               │
╰──────────────────── sort: Newest │                                                      │
                                   │                                                      │
 0 agents · 0 procs                ╰──────────────────────────────────────────────────────╯
 t View · g Group · n New · / · D Diff · ^K Cmds · ?
```

With live sessions and memory history, the strip's two rows read like this (sparkline on top, readout below), colored green/amber/red by pressure:

```
 ▂▃▄▅▅▆▇▇██▇▆▅▄▃
 64% 22.7/32G · 7 agents · 43 procs
```

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added `tests/e2e/diagnostics_strip.rs` (F9 toggles the strip on and off through the real binary), plus unit tests for the `/proc` and `vm_stat` parsers, the pressure-band and byte-formatting helpers, and the keybinding wiring. Verified locally: `cargo fmt`, `cargo clippy -- -D warnings` (default and `--features serve`), `cargo test` (default, `--features serve`, and `--no-default-features` bare-core), and the e2e leg all clean.

## AI Usage

- [ ] No AI was used
- [x] AI was used

I used it to help with research (the memory-accounting details, how other terminal monitors present this, and the ratatui widget APIs) and to speed up the implementation and tests. The problem, the design decisions (headroom-not-PSI as the plotted line, sampling off the render thread, scoping counts to AoE-managed sessions), and the review are mine, and I have read and understand all of the diff.

## Summary

- Added an optional diagnostics strip to the plain TUI.
- Added persistent visibility control through `session.show_diagnostics_pane`.
- Added a System Health view with memory, CPU, pressure, timeline, and agent metrics.
- Added background metric sampling for Linux and macOS.
- Added parser, unit, regression, and end-to-end test coverage.

## Benefits

Users can monitor system health and AoE resource usage without leaving the TUI. Background sampling keeps the interface responsive. The feature remains disabled by default.

## Potential enhancement

The diagnostics strip could become a sidebar tab with detailed health timelines and per-agent metrics. Support for additional platforms could also be added.

### DIFF
--- a/src/process/linux.rs
+++ b/src/process/linux.rs
@@ -132,6 +132,71 @@ pub(super) fn sample_memory() -> super::metrics::MemorySample {
     }
 }
 
+pub(super) fn sample_system() -> super::metrics::SystemReading {
+    let stat = fs::read_to_string("/proc/stat").unwrap_or_default();
+    let cpu = stat.lines().next().and_then(|line| {
+        let values: Vec<u64> = line
+            .split_whitespace()
+            .skip(1)
+            .filter_map(|value| value.parse().ok())
+            .collect();
+        (!values.is_empty()).then(|| {
+            (
+                values.iter().sum(),
+                values.get(3).copied().unwrap_or(0) + values.get(4).copied().unwrap_or(0),
+            )
+        })
+    });
+    let load = fs::read_to_string("/proc/loadavg").ok().and_then(|value| {
+        let values: Vec<f64> = value
+            .split_whitespace()
+            .take(3)
+            .filter_map(|part| part.parse().ok())
+            .collect();
+        (values.len() == 3).then(|| [values[0], values[1], values[2]])
+    });
+    let mem = fs::read_to_string("/proc/meminfo").unwrap_or_default();
+    let field = |name: &str| {
+        mem.lines()
+            .find(|line| line.starts_with(name))
+            .and_then(|line| line.split_whitespace().nth(1)?.parse::<u64>().ok())
+            .unwrap_or(0)
+            * 1024
+    };
+    let total = field("SwapTotal:");
+    let free = field("SwapFree:");
+    (cpu, None, load, (total, total.saturating_sub(free)))
+}
+
+pub(super) fn process_snapshot() -> Vec<super::metrics::ProcessRecord> {
+    let hz = unsafe { libc::sysconf(libc::_SC_CLK_TCK) }.max(1) as f64;
+    let page = unsafe { libc::sysconf(libc::_SC_PAGESIZE) }.max(1) as u64;
+    let Ok(entries) = fs::read_dir("/proc") else {
+        return Vec::new();
+    };
+    entries
+        .filter_map(Result::ok)
+        .filter_map(|entry| {
+            let pid = entry.file_name().to_str()?.parse::<u32>().ok()?;
+            let stat = fs::read_to_string(entry.path().join("stat")).ok()?;
+            let end = stat.rfind(')')?;
+            let fields: Vec<&str> = stat[end + 2..].split_whitespace().collect();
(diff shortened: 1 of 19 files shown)


MatthewWolff:
## Screenshot of a stable memory graph 

Screenshot 2026-08-11 at 4 23 50 PM

njbrake:
@MatthewWolff Ok i've chewed on this for a few days and I think I love this, and actually i'm now thinking that this is a perfect fit for a core feature.

I'm building up a few additions now so that actually this can be a 'system health' tab in the sidebar. The tab will display the summary of metrics as you have it now, but then if you click the tab it will populate the preview pane of the aoe TUI with all of the detailed health metrics.

This way we can keep the sidebar slim but have a way for people to dive into details if they need. With this in mind I'm thinking we have the sidebar only display the immediate Mem/CPU stats, but then if they click on the sidebar item it can populate the preview pane with the timeline and per-agent metrics, etc. How does that sound?

MatthewWolff:
@njbrake As I've used the feature a bit more on my side, I agree the memory visualization isn't very useful and generally appears flat. What I _have_ found actually useful about it was the color-coding of that graph (sometimes it would flash yellow to red (even at 40%), which was more indicative of the box thrashing than the actual % level). Very aligned with your proposal—I think seeing a color-coded percentage would totally capture the value for me

njbrake (approved):
The system health MVP is focused and ready: current pressure signals remain, while historical timelines are deferred for a follow-up.

PR #3343, opened 2026-08-12
feat(tui): merge remote aoe serve sessions into the local session list

### LINKED ISSUE #3340: TUI: merge remote daemon sessions into the local home list instead of replacing it
> _Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

**Status: proposal, not a spec.** The code references below were read from `main` at the time of writing, and the design follows from them, but nothing here has been prototyped. Validate the claims (especially that the daemon's status poller and live-ws surface really do cover remote tmux rows end to end) and re-think the shape before implementing. Treat the file and line citations as starting points that may have drifted, not as settled fact.

## Problem

Users running agents on more than one machine cannot get a single session list. From Discord: a user keeps most sessions on their desktop, needs a couple on a remote box, and wants them under one pane of glass. Both current workarounds fail:

- A `[session.custom_agents]` entry like `remote-claude = "ssh -t host claude"`, which is the answer given in #473, ties the agent's lifetime to the SSH connection. A link interruption kills the running agent.
- Nesting a remote `tmux new-session -A` inside the local pane works on first launch but does not reattach cleanly.

## What's there today

Most of the machinery already exists, and the durability property the workarounds lack is already solved: `src/process/runner.rs` and `src/process/worker_registry.rs` keep the agent subprocess alive across daemon restarts, so a dropped connection is a detach rather than a kill.

Three things stop it from being a single pane of glass:

1. `src/tui/mod.rs:231` early-returns into `remote_home::run_standalone(endpoint)` when `discover_env()` finds a URL, bypassing `App::run`. The remote list replaces the local one instead of merging into it.
2. `AOE_DAEMON_URL` is single-valued by construction. `discover_env()` (`src/acp/client/discovery.rs:130`) reads one URL plus one token from env at startup, and `--daemon-url` is a global clap arg (`src/cli/definition.rs:65`) that retargets every `aoe` invocation, not just the TUI (`src/cli/serve.rs:1576`). Once endpoints are plural there is no answer to which daemon a bare `aoe ps` means. Unreachability is also fatal today (`src/cli/serve.rs:1594`), which does not survive one machine being asleep.
3. `src/tui/remote_home/mod.rs:218` filters the list to `View::Structured`, so remote tmux sessions are invisible.

That third one looks like it needs an SSH hop to fix, and it appears not to. The daemon already covers tmux sessions on both axes:

- **Status.** `status_poll_loop` (`src/server/mod.rs:4334`) runs in the daemon on a 2s tick and is authoritative for tmux-backed sessions, with no TUI needed on that host. `GET /api/sessions` returns both kinds already; `RemoteSession` even deserializes `view` with a terminal default (`src/tui/remote_home/mod.rs:47`) before the client discards those rows.
- **Interaction.** `/sessions/{id}/live-ws` (`src/server/mod.rs:1933`, implemented in `src/server/live_ws.rs`) is a bidirectional remote terminal for tmux sessions: the server polls `capture-pane` and pushes ANSI frames, and client input returns as raw bytes delivered through `tmux send-keys -H`. No PTY and no `tmux attach`, plus resize under a size-owner lock and read-only viewers.

`live_ws.rs` describes itself as mirroring the TUI's own live-send architecture (`src/tui/home/live_send.rs`), so both ends already agree on the interaction model.

## Proposal

- Add a persisted registry of remote endpoints (name, url, token, enabled) and union them with local sessions into one home list with a machine column. Keep `AOE_DAEMON_URL` as the existing single-target retarget; it is not the multi-machine primitive.
- Store the registry at 0600 in the app dir rather than in `config.toml`, following the precedent set by `serve.passphrase` and `login_sessions.toml`.
- Drop the `View::Structured` filter. Structured rows keep opening in the structured view; tmux rows reuse `src/tui/home/live_send.rs` with live-ws as the transport instead of local `capture-pane` and `send-keys`.
- Make per-endpoint failure a per-row degrade rather than a startup error. An unreachable machine greys out its rows.
- Keep local-filesystem actions (worktree, diff, file watch) unavailable on remote rows. Those operate on a `worktree_path` that does not exist on the client machine.

## Explicitly not this

Teaching `tmux_command()` (`src/tmux/mod.rs:175`) to prefix `ssh`, and equally an `ssh -t <host> aoe attach <id>` handoff for tmux rows. The first puts an SSH round trip inside the status poller and leaves the diff panel and file watcher pointed at a path that is not there; the second is unnecessary if live-ws already carries frames out and keystrokes back. The remote `aoe` should own its own machine's state. Full-fidelity `tmux attach` stays available by SSHing in and running `aoe` there.

## Open questions

- Does live-ws's capture-snapshot fidelity suffice for a remote tmux row, given it is not a PTY and so has no prefix passthrough or native copy-mode? A row-level "SSH in for a real attach" action may be the right escape hatch.
- How the master authenticates to N daemons, and whether tokens should be a keyring reference rather than a file even at 0600.
- Whether the merged list should be built over local storage plus N HTTP clients as an interim, or wait for the TUI to become a client of its own local daemon so that local and remote collapse into one code path with N+1 endpoints. The latter is #2734's end-state and this should not block on it.

## Related

- #473, closed on the custom-agents workaround. This is the durability gap that workaround leaves.
- #1704, a `SessionSource` trait over `HomeView` and `remote_home`. Closed, and staying closed, but it is the natural enabling refactor for a merged list and worth reading before starting.
- #2734, the full TUI-as-daemon-client inversion. Multi-remote makes it more valuable rather than depending on it.

Eric162:
## Description

Adds `[remotes.<name>]` config entries whose structured-view sessions are mirrored into the local session list as read-only rows in their own pinned bottom-shelf section, sibling of Archived and Trash. Pressing Enter opens the remote session's transcript in the existing embedded structured view.

The motivation: if you run `aoe` locally and also have a Linux box running its own `aoe`, there is currently no way to see both sets of sessions in one place. You either keep two terminals open or SSH over and run a second TUI.

```
▼ ⇄ deadbox (0) unreachable: transport error
▼ ⇄ linuxbox (2)
 ⠿ fix-flaky-test
 ⠛ refactor-parser
```

Managed by `aoe remote add|list|remove|enable|disable`. The subcommands are config-only: nothing contacts the remote, so `add` works while the other box is still booting, and reachability is reported on the row itself, which is where you would look for it.

### The design decision worth reviewing

Remote rows are a new `Item::RemoteSession { remote, id, depth }` variant held in their own `remote_sources` collection, **not** synthesized `Instance`s in `HomeView::instances`.

Every local operation (stop, delete, worktree edit, diff, tmux attach) resolves its target through `instances`, against this machine's filesystem and tmux server. A synthesized `Instance` would make all of those silently addressable and wrong. Keeping remote rows behind a distinct variant means an operation that has not explicitly opted in cannot reach one at all, and adding the variant made the compiler enumerate all 14 decision sites rather than leaving me to find them by inspection. Selection uses a separate `selected_remote` field, so every local keybinding (which all gate on `selected_session`) is disarmed at once.

This is also why `Item::RemoteSession` and `append_remote_sections` are deliberately **not** `serve`-gated: the match arms then compile identically in both build configurations, so a TUI-only build cannot drift into a non-exhaustive match.

### Notes for the reviewer

- Terminal-view remote sessions are filtered out at fetch time, not render time: a remote tmux pane cannot be attached from this machine, so a row for one would advertise an action the TUI has no way to perform.
- The daemon serializes session status with `format!("{:?}", ...)`, not the lowercase serde repr, so `parse_api_status` parses tolerantly and degrades an unknown variant to `Unknown` rather than dropping the row.
- Refresh is 6s per remote with at most one request in flight each. The HTTP timeout is 15s, so without the in-flight guard a remote that black-holes packets would accumulate overlapping requests.
- `aoe remote list` reports `has_token: bool` and never the token, since its output ends up in issue reports and scrollback.
- `Config::remotes` intentionally bypasses the single-source settings schema, following the existing `Config::tools` precedent: the entries are user-named, so there is no fixed field for a widget to bind to.

### Known limitation

A stored token goes stale when the remote daemon restarts and rotates it. Local endpoints handle this by re-reading `serve.token`, but that file is on the other machine. Today you see `unreachable: unauthorized` and re-run `remote add --force`. Fixing it properly needs the daemon to expose a longer-lived credential for configured remotes, which felt like a separate change.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt --check`, `cargo clippy --features serve,e2e-tests --all-targets` (7 warnings, all pre-existing and none in the added lines), `cargo test --features serve` (6142 passed, 0 failed), and `cargo test --features serve,e2e-tests --test e2e remote_cli` (6 passed) are all clean. `cargo xtask gen-docs` produces no diff. TUI-only `cargo check --all-targets` is clean, confirming the feature gating.

Docs: added `docs/guides/remote-sessions.md` and registered it in `website/scripts/sync-docs.mjs` (PAGES + URL_MAP) and `website/src/data/docsNav.ts`.

Screenshot: the sidebar section above is a real capture from a live TUI run against a second `aoe serve` daemon on port 8099, with `deadbox` pointed at a host that does not exist.

## Test Coverage Analysis

No `web/` file is touched; the merged list is TUI-only.

**TUI / CLI (e2e test)**

Added `tests/e2e/remote_cli.rs`: 6 tests driving the real binary with no tmux and no network (every remote points at a `.invalid` host), covering the `config.toml` round-trip, `?token=` lifting, duplicate/`--force`, disable/enable/remove, validation rejections, and that no output path prints a stored token.

Plus 22 unit tests for the TUI half (section layout, shelf placement, selection safety, Enter dispatch, cursor preservation across a refresh, collapse, empty-section visibility, status parsing).

Two bugs found by running the real binary that the unit tests and the compiler both missed, each now locked in by a test:

1. With no local sessions, the "No sessions yet" placeholder painted over the whole sidebar, hiding the remote sections and their connection errors.
2. Enter did nothing on a remote row: the binding gates on `selected_session.is_some()` before dispatching, and remote rows deliberately leave that unset.

The one path not covered end to end is the embedded view actually connecting to a remote session. It reuses `EmbeddedView::connect(endpoint, session_id)` verbatim, identical to the local structured view with only the endpoint differing, and the Action dispatch is unit-tested, but the remote WebSocket handshake itself is not exercised by a test.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Code (Opus 5)

Used for the implementation, the test suite, and this description. The design decision to model remote rows as a distinct `Item` variant rather than synthesized `Instance`s, and the choice of the daemon-client approach over an SSH exec target, were discussed and settled before any code was written.

## Summary

- Added `aoe remote` commands to add, list, remove, enable, and disable remote daemons.
- Added remote session configuration with safe token handling and URL validation.
- Added remote sessions to the TUI in a separate, read-only shelf.
- Added independent selection, search, collapse, status indicators, and structured-view access for remote sessions.
- Added refresh handling, connection-state reporting, terminal-session filtering, and protection against local actions targeting remote sessions.
- Added CLI and TUI tests for configuration, validation, lifecycle commands, token safety, persistence, and session behavior.
- Added CLI and user documentation for remote sessions.

## Benefits

Users can view structured sessions from configured remote daemons in the same TUI as local sessions. Remote sessions remain isolated from local operations, and connection failures do not remove existing session data.

A known limitation is that users must refresh tokens manually after a remote daemon rotates them.

## Technical details

- Added `Source::ConfiguredRemote`, `RemoteConfig`, `Item::RemoteSession`, and remote section helpers.
- Added asynchronous per-remote refreshes with in-flight request limits.
- Added tolerant status parsing with unknown values mapped to `Status::Unknown`.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -67,6 +67,12 @@ This document contains the help content for the `aoe` command-line program.
 * [`aoe project list`↴](#aoe-project-list)
 * [`aoe project add`↴](#aoe-project-add)
 * [`aoe project remove`↴](#aoe-project-remove)
+* [`aoe remote`↴](#aoe-remote)
+* [`aoe remote list`↴](#aoe-remote-list)
+* [`aoe remote add`↴](#aoe-remote-add)
+* [`aoe remote remove`↴](#aoe-remote-remove)
+* [`aoe remote enable`↴](#aoe-remote-enable)
+* [`aoe remote disable`↴](#aoe-remote-disable)
 * [`aoe worktree`↴](#aoe-worktree)
 * [`aoe worktree list`↴](#aoe-worktree-list)
 * [`aoe worktree info`↴](#aoe-worktree-info)
@@ -148,6 +154,7 @@ Run without arguments to launch the TUI dashboard.
 * `plugin` — Manage plugins (list, info, enable, disable, install, update, uninstall)
 * `profile` — Manage profiles (separate workspaces)
 * `project` — Manage the project registry used by multi-repo session pickers
+* `remote` — Manage remote `aoe serve` daemons whose sessions appear in the local list
 * `worktree` — Manage git worktrees for parallel development
 * `tmux` — tmux integration utilities
 * `sounds` — Manage sound effects for agent state transitions
@@ -1087,6 +1094,88 @@ Remove a project from the registry
 
   Possible values: `global`, `profile`
 
+
+
+
+## `aoe remote`
+
+Manage remote `aoe serve` daemons whose sessions appear in the local list
+
+**Usage:** `aoe remote <COMMAND>`
+
+###### **Subcommands:**
+
+* `list` — List configured remotes
+* `add` — Add a remote daemon whose sessions appear in the local list
+* `remove` — Remove a remote
+* `enable` — Enable a remote without re-entering its URL
+* `disable` — Skip a remote's sessions without deleting its entry
+
+
+
+## `aoe remote list`
+
+List configured remotes
+
+**Usage:** `aoe remote list [OPTIONS]`
+
+###### **Options:**
+
+* `--json` — Output as JSON
+
+
+
+## `aoe remote add`
+
+Add a remote daemon whose sessions appear in the local list
+
+**Usage:** `aoe remote add [OPTIONS] <NAME> <URL>`
+
+###### **Arguments:**
+
+* `<NAME>` — Short label for the remote; badged onto each of its rows
+* `<URL>` — Base URL of the remote daemon, e.g. `http://linuxbox:8080`. A `?token=` copied from the remote's `aoe url` output is accepted and lifted into the stored token
+
+###### **Options:**
+
+* `--token <TOKEN>` — Bearer token for the remote daemon. Omit for a `--no-auth` daemon or when the token is already in the URL
+* `--force` — Replace an existing entry with the same name
+
+
+
+## `aoe remote remove`
+
+Remove a remote
+
+**Usage:** `aoe remote remove <NAME>`
+
+###### **Arguments:**
+
+* `<NAME>` — Name of the remote to remove
+
+
+
+## `aoe remote enable`
+
+Enable a remote without re-entering its URL
+
+**Usage:** `aoe remote enable <NAME>`
+
+###### **Arguments:**
+
+* `<NAME>` — Name of the remote
+
+
+
+## `aoe remote disable`
+
+Skip a remote's sessions without deleting its entry
+
+**Usage:** `aoe remote disable <NAME>`
+
+###### **Arguments:**
+
+* `<NAME>` — Name of the remote
 
 
 
--- a/docs/guides/remote-sessions.md
+++ b/docs/guides/remote-sessions.md
@@ -1,0 +1,118 @@
+# Merged Remote Sessions
+
+Show the sessions running on another machine in your local `aoe` TUI, alongside your own.
+
+If you keep sessions on a laptop and also run agents on a Linux box, this puts both lists on one screen. Each remote gets its own section at the bottom of the sidebar, below Archived and Trash, and you can open a remote session's transcript and composer without leaving the TUI.
+
+## What this is (and is not)
+
+The merged list is a **read-only mirror plus a live transcript**. A remote row lets you read the session, watch it work, and send it messages through the structured view. It does not let you run local operations against it.
+
+| | Local sessions | Remote sessions |
+|---|---|---|
+| Read transcript, send messages | yes | yes |
+| Stop, delete, restart, rename | yes | no |
+| Diff view, worktree edits | yes | no |
+| tmux attach | yes | no |
+| Create new sessions | yes | no |
+
+Two limits are worth understanding before you set this up:
+
+- **Structured-view sessions only.** A remote terminal (tmux) session is not listed, because its pane cannot be attached from another machine without SSH'ing into the host first. Showing the row would advertise something the TUI cannot do.
+- **No local filesystem.** A remote session's worktree lives on the other machine, so the diff view, file watching, and worktree editing do not apply.
+
+For anything outside that, use the remote's own web dashboard (`aoe serve`), or SSH in and run `aoe` there.
+
+## Setup
+
+### 1. Run a daemon on the remote machine
+
+On the Linux box:
+
+```bash
+aoe serve --host 0.0.0.0
+```
+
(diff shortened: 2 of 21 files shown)


njbrake:
> Terminal-view remote sessions are filtered out at fetch time, not render time: a remote tmux pane cannot be attached from this machine, so a row for one would advertise an action the TUI has no way to perform.

I think that we need to adjust functionality so that this is possible. Lets think about what options we have

njbrake:
Closing this as stale but feel free to re-open when it's ready for a look, or feel free to discuss via a GH issue. 🙏

PR #3351, opened 2026-08-13
refactor(tmux): single source of truth for aoe-managed tmux setting resolution and application (#3349)

### LINKED ISSUE #3349: refactor(tmux): single source of truth for aoe-managed tmux setting resolution and application
## Problem

The `[tmux]` setting application logic is not consolidated. After #3227 introduced per-option `auto` deferral for `mouse` only, the three managed options diverge on two axes, all in `src/session/config.rs`:

**1. Deferral granularity is inconsistent.**
- `mouse` (auto) defers per option: `TmuxMouseMode::Auto => (!user_sets_mouse).then_some(true)` via `user_tmux_config_sets_mouse()` which parses the user's tmux config for the `mouse` option specifically.
- `clipboard` (auto) defers coarsely: `TmuxClipboardMode::Auto => !user_has_tmux_config()`.
- `status_bar` (auto) uses the same coarse gate: `TmuxStatusBarMode::Auto => !user_has_tmux_config()`.

A user whose `tmux.conf` sets a prefix key but never touches clipboard still loses aoe's `set-clipboard` passthrough, while the same user now gets aoe's `mouse on`. `clipboard` maps to a single tmux option (`set-clipboard`), so it is the same shape as `mouse` and could use the same per-option probe.

**2. Tri-state modeling diverges.** `should_apply_tmux_mouse` returns `Option<bool>` (on / off / leave-alone), while `should_apply_tmux_status_bar` and `should_apply_tmux_clipboard` return plain `bool` (apply / skip, no distinct force-off). Three create paths (`src/tmux/session.rs`, `src/tmux/terminal_session.rs`, `src/tmux/tool_session.rs`) each call the three `should_apply_tmux_*` helpers independently.

## Proposal

Describe each aoe-managed tmux setting once, and drive resolution plus application from it:

```rust
// One row per tmux option aoe manages on its own sessions.
struct TmuxManagedSetting {
    key: &'static str,                              // for docs/logs
    mode: TmuxTriMode,                              // Auto | Enabled | Disabled
    apply: &'static [(&'static str, &'static str)], // e.g. [("mouse", "on")]
    defer: DeferRule,
}

enum DeferRule {
    WhenUserSets(&'static str), // defer only if the user's config sets THIS option
    WhenUserHasAnyConfig,       // coarse, for a holistic theme
}
```

- Generalize `tmux_config_sets_mouse` into `tmux_config_sets(option: &str)` and back `WhenUserSets` with it.
- `mouse` -> `WhenUserSets("mouse")`, `clipboard` -> `WhenUserSets("set-clipboard")` (closes gap 1), both with uniform tri-state resolution.
- `status_bar` -> `WhenUserHasAnyConfig` stays: it applies a whole theme (`status`, `status-left/right`, `status-style`, ...), not one option. Document that it is coarse by design.
- One resolver (`mode + defer -> apply / skip / force-off`) and one applier iterating the table at the shared create path, so a future managed option is one table row, not a new helper plus three call-site edits.

Implementation MUST add tests covering each story; PR is not done without them.

- As an AoE user, when my `tmux.conf` sets a prefix key but never sets clipboard, then `[tmux] clipboard = "auto"` still applies aoe's `set-clipboard` passthrough. Suggested: unit table over `tmux_config_sets(option)` on realistic config text, mirroring the existing `test_tmux_config_sets_mouse`.
- As an AoE user, when I set `[tmux] mouse = "disabled"`, then all three managed options resolve through the same tri-state path with no forced-on value surviving. Suggested: unit test asserting uniform resolution per mode across the three settings.
- As a maintainer, when I add a new managed tmux option, then it is one `TmuxManagedSetting` row, not a new `should_apply_*` helper plus three create-path call sites. Suggested: a test that iterates the table and asserts each row resolves and applies without per-helper branching.

## Out of scope

- Changing the decision semantics of any single option beyond unification (each option keeps its current effective behavior).
- Repo-scope `[tmux]` overrides (#3229) and the silent whole-config fallback (#3228), both already tracked separately.

## Cross-references

- Raised in the review of #3227 (https://github.com/agent-of-empires/agent-of-empires/pull/3227#pullrequestreview-4848027057), which fixed the `mouse` half only.
- #3207: the bug #3227 fixed.
- #3228, #3229: follow-ups already filed from the #3227 review.

jerome-benoit:
## Description

Fixes #3349

The `[tmux]` setting application logic was only half consolidated after #3227: resolution already ran through one table (`TmuxSetting::row` + `resolve_tmux_setting`), but application still branched per setting inside `append_tmux_setting_args`, which called two dedicated helpers (`append_mouse_args`, `append_clipboard_passthrough_args`) with different action handling. Adding a managed option meant a new helper plus an edit at the shared create path.

This PR makes the table the single source of truth for application too:

- `TmuxManagedSetting` rows now carry the writes each action emits: `apply` (for `Apply`) and `force_off` (empty where "off" is not expressible, e.g. clipboard's server- and window-scoped options).
- `TmuxOptionWrite` models the three tmux scopes (session `-t <target>`, server `-s` with no target, window `-w -t <target>`) plus the `-q` flag, so an illegal write (a server option carrying a target) does not compile.
- `append_tmux_setting_args` iterates `TmuxSetting::ALL` and emits whatever the resolved action declares; `LeaveToUser` writes nothing at creation (a fresh session has no session-scoped value to clear).
- The post-creation path (`status_bar::apply_all_tmux_options`) is untouched: it paints the theme and unsets stale values, which the static write lists cannot express (`set-option -u` is not a value write).

No decision semantics change: same `(setting, mode, user tmux config)` to action mapping, same emitted tmux args (mouse before clipboard, `-q` kept for tmux < 3.3, no target on the server scope), same three create paths calling `append_tmux_setting_args` with the same signature. One new per-creation cost: resolving the status-bar row now probes the user's tmux config (`user_has_tmux_config`, three `exists()` calls) and discards the result, the documented price of iterating the table uniformly.

Deviations from the issue's sketch, each documented in code: no `key` field (no consumer; the enum variant is the row's identity), `WhenUserSets` takes a slice (clipboard defers on two options, `set-clipboard` and `allow-passthrough`), and the write type carries scope plus `-q` rather than a bare `(&str, &str)` pair because the emitted args differ per scope.

## PR Type

- [x] Refactor
- [ ] New Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (see test notes below; the only red tests are pre-existing environment-dependent ones, reproduced on a clean `86ccdd16` worktree)
- [x] Documentation was updated where necessary (code docs; `docs/guides/configuration.md` and `docs/guides/tmux-status-bar.md` unchanged because user-visible behavior is identical)
- [x] For UI changes: N/A (no UI change)

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (no `web/` files)

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the three user stories from #3349 are covered by cheap unit tests (table, emitter, isolated-HOME end-to-end), following the precedent set by #3227; an e2e would add roughly 2s to the strictly serial suite for no extra failure mode.

Tests added (all unit, table-driven where the repo style calls for it):

- `tmux::utils::tests::test_tmux_option_write_emission`: token-exact emission per write scope, including `-q` before the scope flags and the absence of a target on the server scope.
- `tmux::utils::tests::test_tmux_setting_writes_table`: the full setting x action to writes matrix straight from the table through the `tmux_setting_writes` seam (US3: adding a managed option is a table row, not a helper plus three call-site edits).
- `tmux::utils::tests::test_append_tmux_setting_args_emits_rows_in_order`: the public entry point with explicit modes, asserting emission order (mouse before clipboard) and that no forced-on write survives `disabled` (US2).
- `tmux::utils::tests::test_user_config_silent_on_option_still_applies_auto`: isolated HOME; a prefix-only tmux.conf still gets the `set-clipboard` passthrough under `clipboard = "auto"`, and each option defers independently (US1); also pins StatusBar's `WhenUserHasAnyConfig` wiring.
- `session::config::tests::test_tmux_setting_defer_matches_apply_options`: invariant pinning `WhenUserSets` options against the row's `apply` writes, compared as sets (a write added without its defer entry would override a user who took that option in hand, the #3207 failure mode for a new option).
- `session::config::tests::test_tmux_setting_all_is_exhaustive`: compiler tripwire pinning `TmuxSetting::ALL` to the enum variants (content and emission order), so a new managed setting cannot be added while silently skipping the unified applier.
- `session::profile_config::tests::test_tmux_decisions_follow_the_config_they_are_given` now iterates `TmuxSetting::ALL` so a new managed option is covered without editing the test.

Replaced: `test_mouse_args_follow_configured_mode` and `test_append_clipboard_passthrough_args` (they tested the deleted helpers; their coverage moves into the tests above with no loss).

Validation run locally:

- `cargo clippy -- -D warnings` (the CI invocation): clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`: clean
- `cargo xtask gen-docs`: no drift (no CLI surface change)
- `cargo test --lib` on every touched or consuming module (filters: `session::config::tests::`, `session::profile_config::tests::`, `tmux::`): 97, 40 and 367 passed respectively (the `tmux::` run has one pre-existing environment failure and one timing-flaky test that passes 3/3 in isolation), all green
- Full `cargo test --lib`: every observed failure is pre-existing and environment-dependent, each reproduced identically on a clean `86ccdd16` worktree: `file_watch` (18, inotify instance limits in this sandbox), `git::worktree`/`git::cleanup` (5), `session::repo_config` hook tests (2), one `session::instance` storage test, one `tmux::terminal_session` env test. The full suite also does not complete within 60 minutes in this sandbox (blocked/hung environment-dependent tests, unrelated to this diff: the diff touches `session/config.rs`, `session/profile_config.rs`, `tmux/utils.rs` only).

## AI Usage

- [x] AI was used

**AI Model/Tool used:** opencode-go/deepseek-v4-flash (agentic; multi-agent design and review per the delivery template)

Design (phase 1) and review (phase 5) ran as independent multi-agent analyses with a convergence loop: 7 design tours (3 fresh-eye agents each, 21 validations; convergence after two consecutive zero-finding rounds, tours 6-7) and 4 code-review tours (3 fresh-eye agents each, 12 validations; convergence after two consecutive zero-finding rounds, tours 3-4). Implementation was a single focused executor. Two judgment calls are encoded in the design rather than delegated: the `TmuxOptionWrite` scope enum (illegal writes do not compile, `-q` is a per-write flag, not derived from scope) and keeping the post-creation unset path out of the static table (it is `set-option -u`, not representable as writes, and `LeaveToUser` means skip at creation but unset on existing sessions).

## Summary

- Consolidated tmux setting resolution and application into one shared `TmuxSetting::ALL` table.
- Added uniform handling for session, server, and window option writes.
- Preserved existing decision behavior, argument ordering, status-bar handling, and configuration deferral rules.
- Removed setting-specific write helpers and replaced them with a shared write serializer.
- Added tests for table coverage, write ordering, scope handling, independent overrides, and invariants.

## Benefits

This change reduces duplicated logic and makes future managed tmux options easier to add. Each new option requires a table entry instead of changes across multiple helpers and creation paths. The tests also protect registration completeness and existing behavior.

## Technical detail

`TmuxOptionWrite` now defines option scope and quiet handling. `append_tmux_setting_args` resolves each managed setting and emits its declared writes. `LeaveToUser` emits no creation-time writes, while status-bar application remains unchanged.


njbrake (approved):
Approving. This is the #3349 follow-up, and it extends the existing `TmuxSetting::row` table from #3227 rather than standing up a second mechanism beside it.

I verified the emitted tmux args are token-identical to main rather than taking the description's word for it: `StatusBar` contributes nothing at creation (both write lists empty), `Mouse` emits the same `set-option -t <target> mouse on|off`, and `Clipboard` emits the same `-q -s set-clipboard on` plus `-q -w -t <target> allow-passthrough on`, in the same mouse-then-clipboard order. Deferral semantics are untouched; `WhenUserSets(["set-clipboard", "allow-passthrough"])` was already on main, so the new US1 test pins existing behavior rather than introducing it. The only real delta is the one you disclosed: resolving the `StatusBar` row now costs up to three `exists()` calls per session creation and discards the result, which is noise next to a `tmux new-session` fork.

Locally: `cargo fmt --all -- --check` and `cargo clippy -- -D warnings` clean; `session::config` 100/100, `session::profile_config` 40/40, `tmux::utils` 26/26. The `tmux::session` / `tmux::tool_session` failures I hit are non-deterministic across runs, live in files this diff does not touch, and are green in CI. Nothing outside the diff went stale either: no references to the deleted helpers remain anywhere in the tree, and the two guides describe the deferral rule, which did not change.

Two nits, neither worth a round trip:

- `TmuxOptionWrite`'s doc says the variant-per-scope means "an illegal write (a server option carrying a target, say) does not compile". No variant holds a target; it is a runtime parameter of the emitter, and all three variants carry the identical `{option, value, quiet}` triple. The shape is fine, the stated rationale is not, and a struct with a `scope` field would collapse the emitter's three near-identical arms into one.
- `append_tmux_setting_writes` is `pub(crate)` with no cross-module caller, so it can be private.

Happy to take either as a follow-up. Thanks for the thorough write-up on the emitted-args equivalence; it made this fast to check.

PR #3284, opened 2026-08-07
feat(sessions): add remote-owner (org) grouping axis to TUI and web sidebar

athal7:
## Description
Adds a remote-owner grouping axis to both the TUI (`GroupByMode::Org`) and the web dashboard sidebar (`org` axis), built on the already-existing `crate::git::get_remote_owner` primitive that today only feeds the decorative sidebar avatar. `get_remote_owner` is pure URL-shape parsing (`host/owner/repo`) with no GitHub API dependency, so it resolves identically for GitHub, GitLab, Bitbucket, and self-hosted git remotes (a test mixes a `github.com` and `gitlab.com` repo in one grouping run to prove it). Sessions/repos with no resolvable owner (scratch sessions, local-only repos, non-hosted remotes) land in a synthetic "No organization" bucket on both surfaces.

See the discussion on #3283: this was originally scoped as "GitHub-org", but since no GitHub-specific data or API is involved, both the issue and this PR were retitled/reworded to the accurate, host-agnostic framing before requesting review.

**Known asymmetry (called out, not fixed here):** the TUI renders org groups flat (`org -> sessions`), while the web dashboard nests repos under each org (`org -> repo -> sessions`). Defensible as-is, since TUI Project mode is already flat and web Repo mode is already flat, so each surface stays internally consistent with its own existing grouping conventions; flagged by @Seluj78's review as worth a line here rather than a silent difference. Not planned as a follow-up unless a concrete need for TUI repo sub-headers under an org comes up.

Resolves #3283

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary (in-app tip copy)
- [x] For UI changes: included screenshot or recording (see comment below)

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] Added or updated a Playwright spec under `web/tests/` and updated `web/tests/coverage-matrix.json`

**TUI / CLI (e2e test)**
[...]

PR #3357, opened 2026-08-13
fix(tui): give the synthetic scratch bucket its own group identity

### LINKED ISSUE #3237: TUI: a real repo named "scratch" shares one project header with the synthetic Scratch bucket

Split out of #3133 / #3222, which fixed the pin gate but deliberately left this.

In the TUI's project view, a real repository whose basename is `scratch` and AoE's synthetic Scratch bucket render as a **single** group header. Reproduced with a probe over one real-repo session plus one scratch session: `headers=[("scratch", 2)]`.

The cause is that in project mode a group's identity *is* its display label. `build_flat_items_by_project` sets `inst.group_path = project_group_name(&inst)`, empty pinned headers are built as `Group::new(&p.label, &p.label)`, and `GroupTree` keys groups by that path. Two different identities that stringify to `scratch` therefore collapse into one node.

#3222 makes the merged case well defined rather than fixing it: the real repo backs the header, so `p` pins the repo and never the app-internal `<app_dir>/scratch/<id>` path. But the two still cannot be collapsed, sorted, or pinned independently, and the session counts are pooled.

The fix is to give the synthetic bucket a sentinel path, the way `ARCHIVED_SECTION_PATH` already works, and then map that sentinel to a display name on every surface that renders or matches a project group label: the row renderer, `project_header_repo_path`, `is_project_label_pinned`, `known_project_group_paths`, `archived_project_sub_path`, the collapse-state map, the context menu, and the command palette. The web dashboard already does it this way (`SCRATCH_GROUP_ID = "__scratch__"` in `web/src/hooks/useRepoGroups.ts`), so the TUI is the outlier here.

Once a sentinel lands, `is_synthetic_scratch_header` in `src/tui/home/mod.rs` (the interim heuristic #3222 added) becomes deletable.

MatthewWolff:
## Description

Closes #3237.

In the TUI's project view a group's identity IS its display-label string: `project_group_key` (formerly `project_group_name`) returned the literal `"scratch"` for every synthetic scratch session, and `GroupTree` keys nodes by that path. A user's own repository whose basename happens to be `scratch` produced the same key, so the two collapsed into one header with a pooled session count and a shared collapse / sort / pin state. The interim `is_synthetic_scratch_header` heuristic from #3222 (label + backing-repo + pin lookup) masked the pin gate but could not fix the merge.

This fix mirrors what org mode already does for `owner@host` keys (#3284): a sentinel identity path plus a seeded `Group` carrying the display name.

- `SCRATCH_GROUP_PATH = "__aoe_scratch_group__"` (identity, no real repo basename can equal it) and `SCRATCH_GROUP_NAME = "scratch"` (display) live beside the existing `ARCHIVED_/TRASH_SECTION_*` constants in `src/session/groups.rs`.
- `project_group_key` returns the sentinel for `inst.scratch`, so a real `scratch` repo keeps a distinct identity and `GroupTree` renders two separate headers.
- `build_flat_items_by_project` seeds `Group::new(SCRATCH_GROUP_NAME, SCRATCH_GROUP_PATH)` only when a live scratch session populates the sentinel (an archived-only scratch group would otherwise create an undeletable phantom header, the same trap the surrounding phantom-header guard already prevents for regular projects).
- A shared `is_synthetic_project_header(path)` predicate (archived || trash || scratch) gates the pin glyph in the header renderer and `project_group_at_cursor`, so those actions never target a header with no backing repo. Keyed strictly on path; a real repo named `scratch` is never caught.
- `project_group_display_name(key)` maps the sentinel back to `"scratch"` at the four sites that render a raw group key to the user: the archived sub-header (`append_archived_section_by_project`), the archive-group prompt (`prompt_archive_selected_group`), the New Session group-field prefill (`open_new_from_selection`), and the command-palette "Jump to group" disambiguator. Real-repo and org keys pass through unchanged.
- `group_repo_path` short-circuits to `None` for any synthetic project header, so "New Session" from the scratch bucket lands on the default cwd rather than another scratch session's throwaway `<app_dir>/scratch/<id>` directory (which would tie the new session's cwd to an unrelated scratch session's lifetime). This matches the "no backing repo to lend" invariant `project_header_repo_path` already documents for the pin action.

`is_synthetic_scratch_header` and the now-dead `!i.scratch` guard in `project_header_repo_path` are removed.

### One-time behavior on upgrade (no migration)

The `project_group_collapsed` map is keyed by group path. Before this change a collapsed synthetic bucket saved key `"scratch"`; after, it saves the sentinel. Two consequences, both self-healing on the first toggle and neither worse than today's shared-key behavior, so no migration is needed:

1. A user who had the synthetic bucket collapsed will see that state land on a real `scratch` repo header once (if one exists); the synthetic bucket starts expanded.
2. The archived-scratch sub-folder collapse key changes from `__aoe_archived_section__/scratch` to `__aoe_archived_section__/__aoe_scratch_group__` once.

### Known limitation left in place

Org mode's `owner@host` key still renders verbatim in the archive-group prompt and archived sub-header (pre-existing, shared with this PR's changed sites). Fixing it needs `&self` and the git-remote cache and is out of scope for #3237.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

TUI grouping change; no visual redesign, no new screen. Screenshot/recording skipped.

## Test Coverage Analysis

The web dashboard already handles this correctly: `useRepoGroups.ts` keys the scratch bucket on `SCRATCH_GROUP_ID = "__scratch__"` with its own display label. Group identity is computed client-side from the `scratch: bool` flag on each session; no group-path string crosses the wire.

TUI / CLI (e2e test)

The change is entirely in the flat-items build for project mode, which is covered by unit tests (see below). Nothing about how the TUI renders, subcommand behavior, or session lifecycle changed.

Tests added / migrated in `src/tui/home/tests.rs`:

- `test_project_group_key_scratch_uses_sentinel_not_label`: asserts `project_group_key` returns the sentinel and `project_group_display_name` maps it back.
- `synthetic_scratch_bucket_is_distinct_from_real_repo`: regression. A real `/repos/scratch` session and a synthetic scratch session render as two headers with distinct paths, `session_count == 1` each (not pooled), the synthetic one is not pinnable, and `selected_group = sentinel` scopes bulk archive to only the scratch session. Models on `build_flat_items_by_org_scopes_same_named_owners_by_host`.
- `scratch_bucket_absent_from_main_flow_when_only_scratch_is_archived`: pitfall guard. When the only scratch session is archived, the seed must not create a phantom `scratch` header in the main flow; the bucket appears only under Archived, and its sub-header displays `"scratch"` (not the sentinel).
- `scratch_bucket_lends_no_repo_path_for_new_session_prefill`: asserts `group_repo_path(SCRATCH_GROUP_PATH)` returns `None`.
- `scratch_label_pin_gate_keys_on_backing_repo_not_label`: rewritten to locate headers by path (the two `"scratch"`-named headers now coexist); each of the five cases still exercises the pin gate on the intended header.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, `cargo test` (default), `cargo test --features serve`, `cargo test --no-default-features` all clean.

## AI Usage

- [ ] No AI was used
- [x] AI was used

I used it to speed up mapping the touch points and drafting the tests, and had subagents adversarially review the design and the diff (correctness, altitude, blast-radius, senior review, then simplification). The identity/display split was the design decision; the sentinel-path direction came from #3237's own description. I reviewed the whole diff, understand it, and answered reviewer feedback in my own words.

## Summary

- Separated the synthetic Scratch bucket from real repositories named `scratch` with a distinct internal identity and shared display name.
- Updated grouping, sorting, collapse state, pinning, selection, archive actions, and session handling to use the correct group identity.
- Created the synthetic group only when live scratch sessions exist.
- Prevented synthetic groups from supplying repository paths for new sessions.
- Replaced the scratch-header heuristic with centralized synthetic-header detection.
- Added tests for group separation, archived-only sessions, pinning, selection, sorting, and new-session path handling.

## Benefits

This prevents session counts and group state from mixing between real and synthetic groups. Real `scratch` repositories can now be managed normally, while synthetic groups remain excluded from repository-specific actions.

### DIFF
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -31,7 +31,12 @@ pub const TRASH_SECTION_NAME: &str = "Trash";
 /// (the way org host-scoped keys are). That keeps a user's own repo named
 /// `scratch` on a distinct identity instead of merging into the bucket (#3237).
 pub const SCRATCH_GROUP_PATH: &str = "__aoe_scratch_group__";
-pub const SCRATCH_GROUP_NAME: &str = "scratch";
+/// Capitalized so the bucket reads as a system group rather than a repo
+/// basename, matching the web sidebar's `Scratch` label. Only safe to differ
+/// from the basename because the identity above no longer rides on the label.
+/// Name sorts lowercase both here and in `sort_by_name`, so the header still
+/// lands at `s`.
+pub const SCRATCH_GROUP_NAME: &str = "Scratch";
 
 #[inline]
 pub fn is_archived_section_path(path: &str) -> bool {
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -1233,10 +1233,18 @@ pub fn append_archived_section_by_project(
 /// uses the min `archived_at` ascending. Attention falls back to
 /// most-recently-archived because archived rows are all tier 99 in the
 /// attention bucket and the tier offers no discriminator inside the shelf.
+///
+/// The bucket key is an identity, not always the label, so AZ/ZA sort on the
+/// display name: the scratch sentinel would otherwise sort under `_` and jump
+/// ahead of every real project instead of landing at `s` (#3237).
 fn sort_archived_project_buckets(buckets: &mut [(String, Vec<&Instance>)], sort_order: SortOrder) {
     match sort_order {
-        SortOrder::AZ => buckets.sort_by_key(|b| b.0.to_lowercase()),
-        SortOrder::ZA => buckets.sort_by_key(|b| Reverse(b.0.to_lowercase())),
+        SortOrder::AZ => {
+            buckets.sort_by_key(|b| project_group_display_name(&b.0).to_lowercase());
+        }
+        SortOrder::ZA => {
+            buckets.sort_by_key(|b| Reverse(project_group_display_name(&b.0).to_lowercase()));
+        }
         SortOrder::Oldest => {
             buckets.sort_by_key(|(_, sessions)| {
                 sessions
@@ -2906,5 +2914,32 @@ mod tests {
         let inst: Instance = serde_json::from_str(legacy).unwrap();
         assert!(!inst.is_archived());
         assert!(inst.archived_at.is_none());
+    }
+
+    /// #3237: the scratch bucket's key is a sentinel, so name-ordering the
+    /// archived sub-folders on the raw key would sort it under `_` and hoist
+    /// it above every real project. It must sort where its label reads.
+    #[test]
+    fn archived_project_buckets_name_sort_uses_display_label() {
+        let myrepo = Instance::new("a", "/repos/myrepo");
+        let throwaway = Instance::new("b", "/app/scratch/x");
+        let zeta = Instance::new("c", "/repos/zeta");
+        let cases = [
+            (SortOrder::AZ, ["myrepo", SCRATCH_GROUP_NAME, "zeta"]),
+            (SortOrder::ZA, ["zeta", SCRATCH_GROUP_NAME, "myrepo"]),
+        ];
+        for (order, expected) in cases {
+            let mut buckets: Vec<(String, Vec<&Instance>)> = vec![
+                ("myrepo".to_string(), vec![&myrepo]),
+                (SCRATCH_GROUP_PATH.to_string(), vec![&throwaway]),
+                ("zeta".to_string(), vec![&zeta]),
+            ];
+            sort_archived_project_buckets(&mut buckets, order);
+            let labels: Vec<&str> = buckets
+                .iter()
+                .map(|b| project_group_display_name(&b.0))
+                .collect();
+            assert_eq!(labels, expected, "{order:?}");
+        }
     }
 }
--- a/src/session/groups.rs
+++ b/src/session/groups.rs
@@ -23,6 +23,16 @@ pub const ARCHIVED_SECTION_NAME: &str = "Archived";
 pub const TRASH_SECTION_PATH: &str = "__aoe_trash_section__";
 pub const TRASH_SECTION_NAME: &str = "Trash";
 
+/// Synthetic group identity for the scratch bucket in project mode. Unlike the
+/// Archived/Trash shelves this IS a real in-flow GroupTree node built from live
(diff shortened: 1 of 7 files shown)


njbrake (approved):
Approving. The identity/display split is the right shape: it extends what org mode already does for `owner@host` keys instead of adding a second mechanism, and it lets the `is_synthetic_scratch_header` heuristic from #3222 go away rather than accumulating another special case. The archived-only seed guard and the "lends no repo path" short-circuit are both good catches that I would have expected to need a follow-up for.

I pushed two commits to your branch, so flagging what and why:

**`fix(tui): sort archived project sub-folders by display label, not group key`.** `sort_archived_project_buckets` name-sorts on the raw bucket key. Once the scratch bucket's key became the sentinel, `_` sorts below every lowercase letter, so the archived `scratch` folder jumped to the top of the Archived shelf in AZ order and to the bottom in ZA instead of landing at `s`. I reproduced it before fixing: AZ over `myrepo` / scratch / `zeta` returned `["scratch", "myrepo", "zeta"]`. The main flow was unaffected because `flatten_tree` passes `g.name` to `sort_groups`, so only the archived shelf sorted on the key. Fix routes AZ/ZA through `project_group_display_name`, plus a table-driven regression test.

**`feat(tui): label the scratch bucket "Scratch"`.** A product call on my side, not a defect in your work. The web sidebar already uses `Scratch` and the guide calls it that, so the TUI was the odd one out; capitalizing was unsafe while the label doubled as identity, and your change is what makes it safe. It also disambiguates the two headers a user with a real `~/scratch` repo now sees. Sort position is unchanged since both sort paths lowercase before comparing.

One thing I left alone: `scratch_bucket_lends_no_repo_path_for_new_session_prefill` is a single assertion whose setup is a subset of `synthetic_scratch_bucket_is_distinct_from_real_repo`, so it could fold in and drop a test fn. Not worth another round trip.

Thanks for picking this one up, and for the writeup on the collapse-key upgrade behavior. That saved me from having to work out whether a migration was needed.

PR #3358, opened 2026-08-13
fix(config): surface config load errors and unrecognized keys

### LINKED ISSUE #3228: config: an invalid value in any key silently reverts the entire config file to defaults
# config: an invalid value in any key silently reverts the entire config file to defaults

## Problem

A single bad value anywhere in `config.toml` discards every setting in the file, with no visible diagnostic. The user keeps a config that looks correct, and aoe runs on built-in defaults.

Found while fixing #3207. The reporter there wrote `[tmux] mouse = "off"`. The valid values are `auto|enabled|disabled`, so `"off"` fails deserialization, and the failure takes the whole file down with it. That is why `aoe settings explain tmux.mouse` reported `source: schema default`: the resolver never saw the file at all. They spent considerable time assuming they were using the values wrong, which is a reasonable conclusion when the tool reports a default and says nothing else.

The reporter of #3218 hit the adjacent shape: unknown keys under `[sandbox]` (`privileged`, `cap_add`, `security_opt`) were accepted, parsed, and discarded with no warning at launch and no error at load. Their words, and they apply equally here: "a warning on unrecognised keys would have turned a multi-step investigation into a one-line answer."

## Root cause

`Config::load` (`src/session/config.rs:2444`) propagates the error correctly:

```rust
let mut config: Config = table.try_into()?;
```

`Config::load_or_warn` (`src/session/config.rs:2462`) then swallows it:

```rust
pub fn load_or_warn() -> Self {
    match Self::load() {
        Ok(config) => config,
        Err(e) => {
            tracing::warn!(target: "session.store", "Failed to load global config, using defaults: {e}");
            Config::default()
        }
    }
}
```

`tracing::warn!` is invisible unless the user has already enabled debug logging, which nobody does before they know something is wrong. There are 35 `load_or_warn()` call sites across `src/`, so this is the default way config is read, not an edge path.

`main.rs:218` does `eprintln!` a warning, but only for the separate `session::load_config()` call that reads the `[logging]` section, and ratatui's alternate screen paints over it a moment later.

Two distinct failure classes, worth one design:

- **Invalid value for a known key.** Hard-fails deserialization, takes the file with it, silent.
- **Unknown key.** Accepted and ignored (no `deny_unknown_fields` on `Config` or its sub-structs), silent. Indistinguishable from a key that works.

## Proposal

Surface both. The plumbing already exists: `main.rs` carries a `debug_log_warning: Option<String>` through to `App::show_startup_warning` (`src/tui/app.rs:700`), and combines multiple messages at `main.rs:458`. Routing a config parse failure into that same channel makes it visible in the TUI rather than under the alt screen.

Things to decide as part of the design rather than piecemeal:

1. **Should a parse error be fatal instead of defaulted?** Running on defaults when the user has a config file is arguably worse than refusing to start: it silently changes behavior they configured. A hard failure with the offending key and line is the most honest outcome for the TUI, but the 35 call sites include non-interactive paths (`aoe serve`, CLI subcommands) where the right answer may differ.
2. **Unknown keys: warn or reject?** `deny_unknown_fields` would turn the #3218 case into a loud error, but it also breaks forward compatibility (an older aoe reading a config written by a newer one) and would fire on keys left behind by a removed feature. A warning listing unrecognised keys is the safer default.
3. **What do the other 34 `load_or_warn` call sites do?** They cannot each print to stderr. Either the parse result is resolved once at startup and shared, or `load_or_warn` gains a "warn once per process" behavior.
4. **`aoe settings explain` should say the file failed to load.** Reporting `source: schema default` when the file exists but did not parse is the specific thing that misled the #3207 reporter. This is the cheapest single improvement in the issue and could land on its own.

## Scope

Global `config.toml` and profile `config.toml`. Both classes above (invalid value, unknown key).

Out of scope: changing any setting's valid values or adding aliases such as accepting `"off"` for `"disabled"`. That is a separate question, and making the error visible is the fix that generalizes.

## Cross-references

- #3207 where this was found; the `[tmux] mouse` fix there deliberately left this alone.
- #3218 secondary complaint, silently ignored `[sandbox]` keys, the unknown-key half of the same problem.

MatthewWolff:
## Description

Closes #3228.

A bad value or a typo in `config.toml` or a profile config used to revert the entire file to built-in defaults with no visible message. `Config::load_or_warn`'s `tracing::warn!` is dropped in one-shot CLI mode (no subscriber), and unknown keys are dropped by serde entirely because `Config` has no `#[serde(deny_unknown_fields)]` (and cannot, without breaking forward-compatibility with older or newer config files).

The extended startup probe `collect_startup_config_warnings` now reports both classes for global config and the active profile:

- The existing parse-failure class (wording unchanged; the two existing tests keep passing).
- Unrecognized keys at any depth via `serde_ignored`, so a nested typo like `[sandbox] privildged = true` prints its dotted path instead of being silently dropped. Map-keyed sections (`agents`, `tools`, `plugins`, `session.custom_agents`, `acp.acp_defaults`, ...) never false-positive because their keys are entries, not struct fields; struct-field typos inside those maps still do.

Two new surfaces route the warning where users watch:

- One-shot CLI (`main.rs::run`): `eprintln!` for user-visible commands (`add`, `list`, `ps`, `status`, `session`, `serve` foreground). Gated on `!should_init` so a subscribed run doesn't duplicate the tracing sink's line, and on `cli::command_name(&cmd).is_some()` so hidden machine-spawned subcommands (`__acp-runner`, `__vt-pipe`, ...) never eprintln into a detached worker's redirected stderr.
- `aoe settings explain`: a top banner when the config failed to parse (in that case every value below is a real schema default; the previous output read as "your file is honored and matches the defaults" — the exact trap the #3207 reporter fell into). A different note lists unrecognized keys when the file parsed cleanly but keys were dropped.

`SettingSource` is untouched: provenance is per-key, parse-failure is per-file. The banner solves the misleading `source: schema default` at the file level without stamping a whole-file condition onto every key.

Adds `serde_ignored = "0.1"` (MIT/Apache; no new transitive deps beyond `serde`, already in the tree). `cargo deny check` on CI covers the license/supply-chain gate.

The profile probe merges overrides onto a default `Config` before running `serde_ignored`, mirroring `validate_overrides_typecheck`. A raw `serde_ignored::deserialize::<ProfileConfig>` would return nothing because `#[serde(flatten)] overrides: serde_json::Map` absorbs every unknown key as valid JSON.

Explicitly not doing:

- Adding `#[serde(deny_unknown_fields)]` to `Config`. Would break forward-compat and removed-feature drift; the issue itself lands on "warn is safer than reject."
- Making `load_or_warn` fatal. Would break repair paths like `aoe settings explain` on a broken file.
- Fixing profile write clobbering (server profile PATCH does `unwrap_or_default` then `save_profile_config`, so editing one field overwrites the rest of a broken profile file). Different defect class (write-path data loss vs read-path visibility). Split into a companion PR.
- Fixing the remote-standalone TUI's dropped startup warning. `remote_home::run_standalone` has no `InfoDialog` analog; needs a real display surface, not a one-liner. Deferred.
- Fixing the same misleading label on `/api/settings/resolved`. Deferred to keep the coverage-matrix mandate off this PR; a companion "add a `parseError` field on the endpoint envelope" PR is the right shape.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

CLI + startup-probe change; no visual redesign, no new screen. Screenshot/recording skipped.

## Test Coverage Analysis

The change surfaces messages via a startup probe. `/api/settings/resolved` and the web dashboard are intentionally out of scope (see the "Explicitly not doing" list above); a companion PR will land the endpoint envelope + Playwright coverage.

TUI / CLI (e2e test)

`main.rs::run` gained one `eprintln!` after migrations; no new subcommand, no new TUI screen, no session-lifecycle change. The startup probe path is fully unit-tested (see below), which is a stronger signal than an e2e assertion on a formatted line.

Tests added in `src/session/mod.rs` (all `#[serial_test::serial]` + `isolate_app_dir()`, alongside the pre-existing bad-global / bad-profile / clean tests):

- `test_collect_startup_config_warnings_flags_nested_unknown_key`: `[sandbox] privildged = true` produces a warning containing the dotted path.
- `test_collect_startup_config_warnings_allows_documented_map_sections`: documented map-key sections (`session.custom_agents`, `agents.<name>`, `tools.<name>`, `plugins."<id>"`) produce zero warnings.
- `test_collect_startup_config_warnings_flags_typo_inside_map_entry`: a struct-field typo INSIDE a map entry (`[agents.claude] status_maap = ...`) still flags. Guards against a reintroduced allow-list that would swallow the nested typos the feature exists to catch.
- `test_collect_startup_config_warnings_flags_unknown_key_in_profile`: profile probe reports unknown keys the same way. Guards the flatten trap (a raw `serde_ignored::deserialize::<ProfileConfig>` would return nothing because `#[serde(flatten)] overrides` absorbs everything).

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, `cargo test --lib` (4494 pass), `cargo test --no-default-features --lib` (4494 pass), `cargo test --features serve --lib` (all scratch/config tests pass).

## AI Usage

- [ ] No AI was used
- [x] AI was used

I used subagents to map the ~35 `load_or_warn` call sites and every user-facing surface where a config error could reach the user, then to adversarially review the design (correctness / altitude / blast-radius / senior), and finally to review the diff (reuse / simplification / efficiency / altitude). The design decisions (surface-not-fatal, `serde_ignored` for nested detection, banner-not-enum-variant for `settings explain`, splitting the profile-write fix into a companion PR) are mine. I reviewed the whole diff, understand it, and answered reviewer feedback in my own words.

## Summary

- Added diagnostics for invalid and unrecognized settings in global and profile `config.toml` files.
- Displayed warnings for applicable CLI commands and notes in `aoe settings explain`.
- Preserved compatibility by ignoring unknown settings.
- Added shared configuration probing and warning helpers.
- Prevented profile checks from creating directories for nonexistent profiles.
- Added tests for nested keys, map sections, typos, parse failures, profile settings, and serialization round-trips.

## Benefits

Users can identify configuration errors instead of silently using built-in defaults. Newer or removed configuration fields remain supported. Nonexistent profiles continue to fail validation without creating directories.

## Technical changes

- Added `serde_ignored` for unknown-key detection.
- Reported ignored keys with nested paths while excluding valid map-entry keys.
- Reused configuration loading and profile override helpers.
- Kept `SettingSource` and valid setting behavior unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,6 +143,7 @@ dependencies = [
  "rust-embed",
  "semver",
  "serde",
+ "serde_ignored",
  "serde_json",
  "serde_yaml",
  "serial_test",
@@ -3915,6 +3916,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.119",
+]
+
+[[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ aoe-plugin-api = { path = "aoe-plugin-api" }
 
 # Serialization
 serde = { version = "1.0", features = ["derive"] }
+serde_ignored = "0.1"
 serde_json = "1.0"
 serde_yaml = "0.9"
 toml = "1.1"
--- a/src/cli/settings.rs
+++ b/src/cli/settings.rs
@@ -34,6 +34,24 @@ fn source_label(source: &SettingSource) -> String {
 }
 
 fn run_explain(key: &str) -> Result<()> {
+    // Surface parse failures / unrecognized keys BEFORE the per-key output.
+    // On a parse failure `resolve` sees `Config::default()`, so every source
+    // legitimately reports `schema default`; without the banner that reads as
+    // "your file is honored and matches the defaults," which was the exact
+    // trap the #3207 reporter fell into (#3228).
+    let probe = crate::session::probe_global_config();
+    if let Some(err) = probe.load_err.as_deref() {
+        eprintln!(
+            "warning: config.toml failed to parse; every value below is a built-in default.\n  file: {}\n  reason: {err}\n",
+            crate::session::config_path_display()
+        );
+    } else if !probe.ignored_keys.is_empty() {
+        eprintln!(
+            "note: some keys in config.toml were not recognized and were ignored: {}\n",
+            probe.ignored_keys.join(", ")
+        );
+    }
+
     let Some(resolved) = resolve(key) else {
         bail!("'{key}' is not a known setting. Use a core `section.field` or a `plugin:<id>.<field>` key.");
     };
--- a/src/main.rs
+++ b/src/main.rs
@@ -424,19 +424,30 @@ async fn run(
         migrations::run_migrations()?;
     }
 
-    // Surface config parse errors and unrecognized keys on stderr for the
-    // one-shot CLI commands that reach this point (add/list/ps/status/session/
-    // remove/send/killall/group/serve-foreground). Skipped when a tracing
-    // subscriber is already running (`should_init`) so the `_or_warn` helpers'
-    // `tracing::warn!` handle it without a duplicate. Gated on
-    // `cli::command_name` so hidden machine-spawned subcommands (`__acp-runner`
-    // etc.) never eprintln into a detached worker's redirected stderr. The TUI
-    // path skips this: `collect_startup_config_warnings` already runs there and
-    // is rendered as a startup dialog by `App::show_startup_warning` (#3228).
-    if !should_init && cli.command.as_ref().and_then(cli::command_name).is_some() {
-        if let Some(warning) = agent_of_empires::session::collect_startup_config_warnings(&profile)
-        {
-            eprintln!("{warning}");
+    // Surface config diagnostics on stderr for user-visible CLI commands
+    // (`add`/`list`/`ps`/`status`/`session`/`remove`/`send`/`killall`/`group`/
+    // `serve` foreground). Two classes with different subscriber overlap:
+    //
+    // - Unrecognized keys: collected only by `serde_ignored` inside the
+    //   startup probe; no other surface reports them. Always emit when a user
+    //   is watching, even when a tracing subscriber is running (`should_init`
+    //   true), because the `_or_warn` helpers cover only parse failures.
+    // - Parse failures: reported by `Config::load_or_warn`'s `tracing::warn!`
+    //   when a subscriber is up. Emit here only when it isn't, so a foreground
+    //   run doesn't duplicate the tracing line.
+    //
+    // Gated on `cli::command_name` so hidden machine-spawned subcommands
+    // (`__acp-runner` etc.) never eprintln into a detached worker's redirected
+    // stderr. The TUI path skips this: `collect_startup_config_warnings`
+    // already runs there and is rendered by `App::show_startup_warning` (#3228).
+    if cli.command.as_ref().and_then(cli::command_name).is_some() {
+        let warning = if should_init {
+            agent_of_empires::session::collect_startup_ignored_key_warnings(&profile)
+        } else {
+            agent_of_empires::session::collect_startup_config_warnings(&profile)
+        };
+        if let Some(w) = warning {
+            eprintln!("{w}");
         }
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -313,6 +313,7 @@ async fn main() -> Result<()> {
     if let Err(e) = run(
         cli,
         is_daemon_child,
+        should_init,
         debug_namespace_drift,
         debug_log_warning,
     )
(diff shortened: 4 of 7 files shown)


njbrake:
Pushed two commits to your branch.

The second is a bug fix rather than cleanup. `collect_startup_config_warnings` builds its display path with `profile_config::get_profile_config_path`, which resolves through the *creating* `get_profile_dir`. That was harmless while only the TUI called it, but the new `main.rs` block runs before dispatch on every one-shot CLI command, so naming an unknown profile creates the directory before the command's own `resolve_existing_profile` can reject it:

```
$ aoe list -p ghost-profile
No sessions found in profile 'ghost-profile'.        # exit 0, should error
$ ls ~/.config/agent-of-empires/profiles/ -> ghost-profile
```

That is the lazy-profile-creation bug `tests/e2e/profile_lazy_creation.rs` guards, and it was the only red check. The fix routes the display path through the non-creating `get_profile_dir_path`. Worth carrying into the companion PRs: the probe is not side-effect-free on the app dir, so any new caller needs the same care.

The first commit collapses the duplicated `collect_startup_ignored_key_warnings` into one collector behind a `WarningClass` enum, and extends the existing clean test with a serializer round-trip, since a Serialize/Deserialize asymmetry anywhere in the schema would warn every user on every command instead of only on a real typo.

Design and deferrals both look right to me.

njbrake (approved):
Verified locally rather than from the description: the ignored-key probe has no false positives (a serializer round-tripped `config.toml`, profile overrides merged onto defaults, `[plugins."<id>".settings]`, `session.custom_agents`, `tools.<name>`, and a stale `[app_state]` block all probe clean), and the `settings explain` banner does not double up with the startup line because `Commands::Settings` returns before the `main.rs` block. The design extends the existing startup probe instead of adding a second warning channel, and the deferrals are the right cuts.

Two non-blocking follow-ups: `docs/guides/configuration.md` still says nothing about invalid values falling back to defaults or unknown keys warning, which is now user-visible behavior; and the three same-shape unknown-key tests could collapse into one table-driven test per AGENTS.md.

PR #3359, opened 2026-08-13
fix(config): stop advertising inert repo [tmux] and [sound] overrides

### LINKED ISSUE #3229: config: repo-level [tmux] overrides are advertised in settings but never resolved

## Problem

A repo-level `[tmux]` block is accepted, is presented as overridable in the settings UI, and never reaches the code that reads it. The settings TUI renders a Repo tab for Mouse Support, so a user can set it there and watch it do nothing.

Found while fixing #3207, which was the same class of bug one layer up (a profile `[tmux]` block was inert because the resolvers read the global config). That fix wired the profile layer and deliberately stopped there, so this is the remaining layer.

## Root cause

`"tmux"` is listed in `REPO_OVERRIDABLE_SECTIONS` (`src/session/repo_config.rs:61`) with `AllowAll`, so `repo_may_override_field` returns true for every `[tmux]` field and `src/tui/settings/fields.rs:752` renders the Repo tab accordingly.

But the resolver the tmux call sites use is profile-scoped:

```rust
// src/tmux/mod.rs
pub(crate) fn tmux_option_config(profile: &str) -> crate::session::Config {
    crate::session::profile_config::resolve_config_or_warn(profile)
}
```

`resolve_config_or_warn` (`src/session/profile_config.rs:124`) merges global and profile and stops. The repo-merged resolver, `resolve_config_with_repo` (`src/session/repo_config.rs:439`), is never called on this path. So `mouse`, `status_bar`, and `clipboard` all ignore the repo layer.

## Why it was scoped out of #3207

Wiring it moves all three tmux resolvers onto repo-merged config at once, and repo config is a different trust domain from profile config: it comes from a file inside the repository, so it is attacker-controlled in a way a profile is not. `REPO_OVERRIDABLE_SECTIONS` exists to bound that, and the comment at `src/session/repo_config.rs:81` shows the section list has already been reasoned about in terms of "fields that name or build the command AoE hands to tmux". Deciding whether tmux *option* application belongs in that trust boundary is a real decision, not a mechanical rewire, and it should not ride along on a bug fix.

Two shapes to weigh:

1. **Honor the repo layer.** Point `tmux_option_config` at `resolve_config_with_repo`. It needs a project path, which the three tmux create paths have and `apply_all_tmux_options` does not obviously have, so that plumbing is the bulk of the work.
2. **Stop advertising it.** Narrow `tmux` in `REPO_OVERRIDABLE_SECTIONS` from `AllowAll` to the fields that actually resolve, so the settings UI stops offering a control that cannot work. Cheaper, and arguably the correct answer if repo-controlled tmux options are not something we want.

Either is an improvement on the current state, where the UI promises something the resolver does not deliver. What should not persist is the third option of leaving it as is.

## Scope

`[tmux]` only (`mouse`, `status_bar`, `clipboard`). Whether other sections in `REPO_OVERRIDABLE_SECTIONS` have the same advertised-but-inert gap is worth a sweep and is not assumed here.

## Cross-references

- #3207 fixed the global and profile layers of this for `[tmux] mouse`; the PR narrowed its own title to "at global and profile scope" rather than claim this layer.
- #3228 the adjacent silent-config-failure class: a bad value or unknown key producing no visible diagnostic.

MatthewWolff:
## Description

Closes #3229. Also fixes the same class of bug for `[sound]` (surfaced during the sweep the issue explicitly asks for) — see below.

The `tmux` and `sound` sections were listed in `REPO_OVERRIDABLE_SECTIONS` at `SectionPolicy::AllowAll`, so the settings UI rendered a Repo tab for each and the sanitizer wrote a `[tmux]` / `[sound]` block to the repo `config.toml` on save. But every consumer of those sections resolves through profile/global-only paths (`resolve_config_or_warn` or `Config::load*`) and never a repo-aware resolver, so the read side never picked up the block: user edits stranded at save time, and hand-written entries silently no-oped.

The fix is Option 2 from the issue body: stop advertising. Remove both sections from `REPO_OVERRIDABLE_SECTIONS`, gate the Tmux and Sound category tabs off Repo scope in `categories_for_scope` (mirroring the existing `StatusHooks` gate), delete the shipped repo template's `# [tmux]` and `# [sound]` example blocks, and add `[sound]` to the "does not support" list in `docs/guides/configuration.md` (docs already called out `[tmux]` there).

Explicitly rejected Option 1 (threading `project_path` through `tmux_option_config` and `apply_all_tmux_options` to make the repo layer live): repo configs are attacker-controlled, and wiring it would force a trust decision on every future tmux field. It also leaves `socket_name` / `vt_live` (which are `global_only`) at risk of drifting into repo-visibility later since the sanitizer keys on `REPO_OVERRIDABLE_SECTIONS` and does not consult the schema's `global_only` flag. Removing the sections produces the honest signal instead: `configuration.md:506` already documents them as unsupported at repo scope; this PR makes code and docs agree.

### Why bundle `[sound]`

The issue's scope note asks for a sweep of every `REPO_OVERRIDABLE_SECTIONS` entry to check for the same defect. `sound` has the identical shape: same `AllowAll` default, no repo-aware consumer, advertised in the shipped template. Bundling is one word deleted from the same array in the same function, one test case added to the same table, one line changed in the same docs paragraph — a stacked pair of PRs would be pure ceremony. The `worktree`, `sandbox`, `session`, and `hooks` sections were verified to route their primary paths through `resolve_config_with_repo*`, so they are unaffected. `updates` is technically inert at the repo layer too, but its host-scoped semantics deserve their own trust discussion; left for a follow-up.

### Latent related bug left for a follow-up (out of scope)

Under Repo scope, `Acp`, `Web`, `Diff`, and `Logging` tabs render editable fields today whose edits silently strand at save time (they behave like tmux/sound did before this PR, minus the sanitizer-writes-to-file half). Same failure class, different fields, requires deliberately deciding which of those sections should become repo-overridable. This PR is scoped to what #3229 names.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Settings UI + docs change; no visual redesign, no new screen. The behavior change is that two tabs no longer appear under the Repo scope selector, mirroring the existing `StatusHooks` gate one line up. Screenshot skipped.

## Test Coverage Analysis

The web has no repo-scope settings editor; it edits global/profile only. No web PATCH path targets repo `[tmux]` / `[sound]`.

TUI / CLI (e2e test)

The change is entirely in the settings-model + sanitizer, covered directly by unit tests. Nothing about how the TUI renders, subcommand behavior, or session lifecycle changed.

Tests added / updated:

- `test_repo_may_override_field` in `src/session/repo_config.rs` is now table-driven (per AGENTS.md "one test per behavior, not per input"), covering every allow / deny case including the new `tmux` and `sound` negatives.
- `test_merge_repo_config_strips_tmux_and_sound` in `src/session/repo_config.rs` mirrors the shape of `test_repo_config_cannot_inject_host_hooks`. Each override differs from its section default so the "merge is a no-op" assertion cannot pass on a false-green.
- `categories_for_scope_tests::repo_scope_drops_non_repo_overridable_categories` in `src/tui/settings/mod.rs` locks the tab-hide contract for `StatusHooks`, `Tmux`, and `Sound` (which had zero test coverage before).
- `repo_scope_yields_no_fields_for_tmux_or_sound` in `src/tui/settings/fields.rs` asserts the field builder returns zero rows for those categories under Repo scope, defense-in-depth against a future direct caller.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, `cargo test --lib` (4493 pass), `cargo test --no-default-features --lib` (4493 pass).

## AI Usage

- [ ] No AI was used
- [x] AI was used

I used subagents to (a) map `REPO_OVERRIDABLE_SECTIONS`, the `SectionPolicy` mechanism, and every `[tmux]` / `[sound]` consumer in the repo; (b) adversarially review the design across correctness / altitude / blast-radius; (c) senior-review the converged plan (which caught a load-bearing rationale error the adversarial pass missed: the TUI never builds fields with `SettingsScope::Repo` because `field_build_inputs` maps `Repo` to `Profile`, so the failure mode is stranded writes at save time, not empty tabs); and (d) run a compensating quality pass over the diff. The design decisions (Option 2 not Option 1; bundle `[sound]`; keep direct `if` gates rather than extracting a generalization that would also hide 6 additional empty tabs) are mine. I reviewed the whole diff, understand it, and answered reviewer feedback in my own words.

## Summary

- Removed repository-level `[tmux]` and `[sound]` overrides.
- Hid their categories and fields from Repo settings.
- Removed their examples from the repository configuration template.
- Documented `[sound]` as unsupported in repository configuration.
- Added tests for permissions, sanitization, field generation, and category visibility.

These changes prevent users from configuring repository settings that have no effect. Future work could add repository-level `tmux` support after defining an explicit trust model.

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -503,6 +503,6 @@ Profile overrides go in `~/.agent-of-empires/profiles/<name>/config.toml` and us
 
 Per-repo settings go in `.agent-of-empires/config.toml` at your project root. Run `aoe init` to generate a template.
 
-Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, and `[worktree]` sections. It does not support `[tmux]`, `[updates]`, `[claude]`, or `[diff]` (those are personal settings).
+Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, and `[worktree]` sections. It does not support `[tmux]`, `[sound]`, `[updates]`, `[claude]`, or `[diff]` (those are personal settings).
 
 See [Repo Config & Hooks](repo-config.md) for details.
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -503,6 +503,6 @@ Profile overrides go in `~/.agent-of-empires/profiles/<name>/config.toml` and us
 
 Per-repo settings go in `.agent-of-empires/config.toml` at your project root. Run `aoe init` to generate a template.
 
-Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, `[worktree]`, and `[updates]` sections. It does not support `[tmux]`, `[sound]`, `[claude]`, or `[diff]` (those are personal settings).
+Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, and `[worktree]` sections. It does not support `[tmux]`, `[sound]`, `[updates]`, `[claude]`, or `[diff]` (those are personal settings).
 
 See [Repo Config & Hooks](repo-config.md) for details.
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -503,6 +503,6 @@ Profile overrides go in `~/.agent-of-empires/profiles/<name>/config.toml` and us
 
 Per-repo settings go in `.agent-of-empires/config.toml` at your project root. Run `aoe init` to generate a template.
 
-Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, and `[worktree]` sections. It does not support `[tmux]`, `[sound]`, `[updates]`, `[claude]`, or `[diff]` (those are personal settings).
+Repo config supports: `[hooks]`, `[session]`, `[sandbox]`, `[worktree]`, and `[updates]` sections. It does not support `[tmux]`, `[sound]`, `[claude]`, or `[diff]` (those are personal settings).
 
 See [Repo Config & Hooks](repo-config.md) for details.
--- a/src/session/repo_config.rs
+++ b/src/session/repo_config.rs
@@ -57,9 +57,13 @@ use super::project_mcp::ProjectMcpServer;
 /// [`HostHooksConfig`]), so honoring it from a repo would let a checked-out
 /// repository execute arbitrary host commands. Host hooks are profile/global
 /// only.
-const REPO_OVERRIDABLE_SECTIONS: &[&str] = &[
-    "hooks", "session", "sandbox", "worktree", "updates", "tmux", "sound",
-];
+///
+/// `tmux` and `sound` are deliberately absent (#3229): every consumer
+/// resolves them through profile/global-only paths (`resolve_config_or_warn`
+/// or `Config::load*`), never a repo-aware resolver, so a repo entry here
+/// could never take effect. Wire those resolvers before making either
+/// section repo-overridable.
+const REPO_OVERRIDABLE_SECTIONS: &[&str] = &["hooks", "session", "sandbox", "worktree", "updates"];
 
 /// What a repo may set inside one overridable section (#3154).
 enum SectionPolicy {
@@ -1622,13 +1626,6 @@ pub const INIT_TEMPLATE: &str = r#"# Agent of Empires - Repository Configuration
 
 # [updates]
 # update_check_mode = "off"
-
-# [tmux]
-# status_bar = "auto"
-# mouse = "auto"
-
-# [sound]
-# enabled = false
 "#;
 
 #[cfg(test)]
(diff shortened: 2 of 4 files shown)


njbrake (approved):
Approving. I verified the load-bearing claim rather than taking the description's word for it: every `[tmux]` consumer resolves through `tmux_option_config` -> `resolve_config_or_warn` or through `Config::load_or_warn`, every `[sound]` consumer through the same two, and none of the `resolve_config_with_repo*` call sites touch either section. The sections really were inert at repo scope.

Design fits. The two new `if scope != SettingsScope::Repo` guards mirror the existing `StatusHooks` gate rather than introducing a visibility table, and `rebuild_categories_for_scope` already falls back to `first_tab_index()`, so switching from Global/Tmux to Repo lands on a real tab.

I pushed one commit to your branch reverting the `[updates]` half of 5aae1b19. CodeRabbit reasoned from `REPO_OVERRIDABLE_SECTIONS` and `INIT_TEMPLATE`, but `updates` has the same defect as `tmux` and `sound`: `get_update_settings` is `Config::load_or_warn().updates`, `auto_update::spawn_if_enabled` gets `Config::load_or_warn()` from both the TUI and `aoe serve`, `get_update_status` uses `resolve_config_or_warn`, and `telemetry::features::active_features` documents itself as deliberately pre-merge global config. A repo `[updates]` block merges and is then read by nobody, and `docs/guides/repo-config.md` lists only Hooks / Session / Sandbox / Worktree. Documenting it as supported asserted the thing this PR exists to stop asserting.

Two non-blocking notes for the follow-up you flagged.

`repo_scope_yields_no_fields_for_tmux_or_sound` covers a branch nothing reaches. `field_build_inputs` maps `Repo` to `Profile` and is the only caller of `build_fields_for_category`, so the `repo_may_override_field` filter never runs in production. The new table row in `test_repo_may_override_field` already catches the regression it guards.

That mapping is the root of the whole class. Theme, Agents, Interaction, Diff, Acp, Web, Logging, and Updates all still render Repo-scope fields whose edits strand at save. The general fix belongs at `field_build_inputs`, not in more per-category gates, and each gate added now makes that slightly harder to land. Scoping it out here was the right call. Worth folding into that follow-up: drop `updates` from the allow list and the template too, so the docs sentence stays true by construction.

PR #3360, opened 2026-08-13
fix(acp): forward non-Claude provider keys to their ACP adapters

### LINKED ISSUE #3238: Non-Claude agent keys are stripped before reaching structured-view agents

MatthewWolff:
## Description

Closes #3238.

Structured-view (ACP) agent spawns clear the environment and forward a fixed allowlist `ALWAYS_FORWARD_ENV` in `src/acp/acp_client.rs` that lists Claude/Anthropic keys only. So `OPENAI_API_KEY`, `GEMINI_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY`, etc. set in the environment running `aoe serve` reach terminal-view sessions but are dropped from `codex-acp`, `pi-acp`, `opencode acp`, `gemini --acp`, and every other non-Claude ACP adapter, including the bundled `aoe-agent`. Auth silently fails in the structured view.

The plumbing was already Direction 2 from the issue body: `AgentSpec.env_allowlist: Option<Vec<String>>` at `src/acp/agent_registry.rs` is a per-adapter extra allowlist consumed by both spawn paths in `src/acp/acp_client.rs` (`apply_env_filter` and `spawn_subprocess`). Every default registration passed `env_allowlist: None`. The mechanism was built; the data was empty.

## What this changes

- New `install_hints::env_allowlist_for(binary: &str) -> &'static [&'static str]`, keyed on the friendly binary token used at registration (`"aoe-agent"`, not `spec.command`, whose `${aoe_data_dir}` placeholder is substituted at spawn time).
- `agent_registry::with_defaults` populates each adapter's `env_allowlist` via `default_env_allowlist(binary)` which reads that slice.
- Two spawn sites (`apply_env_filter` and `spawn_subprocess`) route allowlist entries through `provider_env_denyreason` (already used for `provider_env`) so a malicious or edited spec cannot smuggle `AOE_TOKEN`, `AOE_*`-prefixed daemon carriers, or `LD_*` / `DYLD_*` linker hooks under cover of a provider auth key. Not exploitable today (registry never deserialized from disk; `upsert_agent` has zero production callers), but it removes the asymmetry between the allowlist and `provider_env`.
- Doc comment on `AgentSpec.env_allowlist` was stale (enumerated a subset of `ALWAYS_FORWARD_ENV` that omitted several keys and mis-described "provider auth env" as `ANTHROPIC_API_KEY`); rewrote to reference `ALWAYS_FORWARD_ENV` directly. Also dropped the incorrect "Vercel AI SDK 6" from `aoe-agent`'s description (`acp-worker/aoe-agent/package.json` pins `ai ^7`).

## Verified per-adapter allowlists (only)

| Adapter | Allowlist | Source |
|---|---|---|
| `aoe-agent` | `OPENAI_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY` | Verified from `acp-worker/aoe-agent/src/index.ts:20-22` imports `@ai-sdk/{anthropic,openai,google}`. Anthropic is already in `ALWAYS_FORWARD_ENV`. `@ai-sdk/google` reads `GOOGLE_GENERATIVE_AI_API_KEY`, NOT the CLI-native `GEMINI_API_KEY`. |
| `codex-acp` | `OPENAI_API_KEY`, `OPENAI_BASE_URL`, `CODEX_HOME` | OpenAI SDK env plus `CODEX_HOME` so the adapter can reuse the operator's `codex login` `auth.json`. |
| `opencode` | `OPENAI_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY`, `OPENROUTER_API_KEY`, `OPENCODE_API_KEY` | AI-SDK-based, so the Google env is the AI-SDK convention (distinct from `gemini --acp`'s CLI-native names). |
| `gemini` | `GEMINI_API_KEY`, `GOOGLE_API_KEY`, `GOOGLE_APPLICATION_CREDENTIALS`, `GOOGLE_CLOUD_PROJECT` | Google Gemini CLI-native names plus Vertex / ADC. |

## Explicit deferrals (rather than ship guessed names)

`pi`, `omp`, `kimi`, `vibe`, and both Claude aliases keep `env_allowlist: None`. Shipping a guessed key that never matches would silently no-op the fix; shipping a wrong one would forward a key the adapter doesn't use. Users needing any of those today can flip `session.inherit_host_environment = true` (already forwards arbitrary env, filtered by `passthrough_denyreason`). Follow-up: verify each adapter's real env reads from its own source and add its arm. This is a deliberate correctness-over-completeness call; the issue body explicitly asks for the fix to not silently no-op.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

ACP env-forwarding change; no visual surface. Screenshot skipped.

## Test Coverage Analysis

The change is entirely in the ACP spawn plumbing; no dashboard control flow moves.

TUI / CLI (e2e test)

Covered by two unit tests directly at the mechanism boundary.

Tests added:

- `agent_registry::tests::default_env_allowlists_match_verified_providers`: table-driven, pins all 10 default adapters (4 populated allowlists + 6 explicit `None`s). One row asserts that `aoe-agent` does NOT receive `GEMINI_API_KEY` (that's the CLI-native name; the AI-SDK bundled agent uses `GOOGLE_GENERATIVE_AI_API_KEY` instead). This negative row catches the "keyed on `spec.command`" regression, where the `${aoe_data_dir}` placeholder would never match and every `aoe-agent` provider key would silently drop.
- `acp_client::tests::apply_env_filter_forwards_agent_env_allowlist`: `#[serial]` + `EnvGuard` + `isolate_app_dir_at`. Sets `OPENAI_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY`, `GEMINI_API_KEY`; spawns the real `aoe-agent` spec through `apply_env_filter`; asserts the first two reach the child while `GEMINI_API_KEY` does not.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings` (both featuresets), `cargo test --features serve --lib` (my new tests pass; the four pre-existing serve-only test flakes in `acp::supervisor::spawn_unknown_agent_errors_cleanly`, `acp::acp_client::reset_*`, and `server::api::acp::wake_prompt_frees_instance_lock_and_publishes_nothing_without_a_worker` are unrelated — they fail the same way with my new tests excluded), `cargo test --no-default-features --lib` clean.

## AI Usage

- [ ] No AI was used
- [x] AI was used

I used subagents to (a) map `ALWAYS_FORWARD_ENV`, the two spawn paths, and the `AgentSpec.env_allowlist` field; (b) adversarially review the design across correctness / altitude / blast-radius (which caught that my initial per-adapter key list for `aoe-agent` had the wrong Google key name — I'd assumed `GEMINI_API_KEY` but the source actually reads `GOOGLE_GENERATIVE_AI_API_KEY`); (c) senior-review the converged plan (which caught a second blocking landmine: keying `env_allowlist_for` on `spec.command` would use the templated `${aoe_data_dir}/...` string and silently miss the bundled agent). The design decisions (Option 2 over Option 1; keyed lookup in `install_hints.rs` over inline literals; explicit deferrals over silent no-ops) are mine. I reviewed the whole diff, understand it, and will answer reviewer feedback in my own words.

## Summary

- Added per-adapter environment allowlists for ACP structured-view sessions.
- Forwarded approved provider credentials through host, detached, and sandboxed sessions.
- Preserved request-specific credential precedence.
- Blocked protected, daemon, linker-hook, and host-only path variables where required.
- Added support for `aoe-agent`, `codex-acp`, `opencode`, and `gemini`.
- Kept unverified adapters unchanged. They can use `session.inherit_host_environment = true` for broader forwarding.
- Unified the `aoe-agent` registration and allowlist token.
- Updated documentation and added tests for forwarding, precedence, sandbox behavior, adapter mappings, and filtering.

## Benefits

- Provider credentials such as `OPENAI_API_KEY` reach supported ACP agents.
- Structured-view sessions authenticate consistently across providers.
- Adapter-specific forwarding limits unnecessary environment exposure.
- Sandboxed sessions receive approved credentials without exposing host-only paths.
- Clear mappings and tests simplify future adapter updates.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3924,9 +3924,11 @@ fn apply_env_filter(cmd: &mut std::process::Command, config: &SpawnConfig) {
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
             // Route through the same deny check `provider_env` uses so a
-            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
-            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
-            // or infra keys under the guise of provider auth.
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`,
+            // `LD_*`/`DYLD_*` linker hooks, or infra keys under the guise of
+            // provider auth. It does not cover the daemon->runner carrier
+            // `AOE_ACP_AGENT_ENV`; only `host_environment_denyreason` rejects
+            // that, and the runner clears it before exec anyway.
             if provider_env_denyreason(name).is_some() {
                 continue;
             }
@@ -4399,9 +4401,11 @@ fn spawn_subprocess(config: &SpawnConfig) -> Result<tokio::process::Child, AcpEr
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
             // Route through the same deny check `provider_env` uses so a
-            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
-            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
-            // or infra keys under the guise of provider auth.
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`,
+            // `LD_*`/`DYLD_*` linker hooks, or infra keys under the guise of
+            // provider auth. It does not cover the daemon->runner carrier
+            // `AOE_ACP_AGENT_ENV`; only `host_environment_denyreason` rejects
+            // that, and the runner clears it before exec anyway.
             if let Some(reason) = provider_env_denyreason(name) {
                 warn!(target: "acp", "ignoring env allowlist entry '{name}' ({reason})");
                 continue;
@@ -15206,7 +15210,8 @@ done
     /// #3238: `AgentSpec.env_allowlist` populated by `with_defaults` must
     /// reach the agent via `apply_env_filter`. Uses the `aoe-agent` spec (the
     /// one that would silently no-op if we keyed `env_allowlist_for` on
-    /// `spec.command` — placeholder-templated — instead of the binary token).
+    /// `spec.command`, which is placeholder-templated, instead of the binary
+    /// token).
     /// A negative case rides along: `GEMINI_API_KEY` is set but not in the
     /// AI-SDK-based `aoe-agent`'s allowlist, so it must NOT be forwarded.
     #[test]
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3821,7 +3821,17 @@ fn build_sandbox_docker_argv(
     // provider auth (the #3238 symptom). docker only forwards a name handed
     // to it via `-e`, so each allowlisted host value is set on the runner
     // (`inherit_env`) and named with `-e KEY`.
+    //
+    // Value-typed entries only, for the same reason `PROVIDER_AUTH_KEYS` below
+    // omits `CLAUDE_CONFIG_DIR`: a host path names nothing inside the
+    // container, so forwarding it points the adapter at a directory that does
+    // not exist instead of the one `AGENT_CONFIG_MOUNTS` bind-mounts at the
+    // canonical container path. These keys stay host-only and keep flowing on
+    // the two non-sandboxed spawn paths, where they are the point.
     for (key, value) in allowlisted_env_pairs(config) {
+        if is_host_only_path_env(&key) {
+            continue;
+        }
         if seen_keys.insert(key.clone()) {
             docker_args.push("-e".into());
             docker_args.push(key.clone());
@@ -3928,6 +3938,22 @@ fn inherited_host_env_pairs(config: &SpawnConfig) -> Vec<(String, String)> {
     }
     let profile = config.source_profile.as_deref().unwrap_or_default();
     crate::session::environment::inherited_host_env(profile)
+}
+
+/// Allowlist entries whose value is a host filesystem path rather than a
+/// credential. They are legitimate on the two host spawn paths and must not
+/// cross into a container: the path names nothing there, so forwarding it
+/// points the adapter away from the config dir `AGENT_CONFIG_MOUNTS` mounts
+/// at the canonical container location. `CLAUDE_CONFIG_DIR` is the case that
+/// established the rule (it is deliberately absent from `PROVIDER_AUTH_KEYS`
+/// in `build_sandbox_docker_argv`); the other two arrived with the per-adapter
+/// allowlists in #3238. Adding a path-valued key to `env_allowlist_for`
+/// means adding it here too.
+fn is_host_only_path_env(key: &str) -> bool {
+    matches!(
+        key,
+        "CLAUDE_CONFIG_DIR" | "CODEX_HOME" | "GOOGLE_APPLICATION_CREDENTIALS"
+    )
 }
 
 /// Resolve the adapter's `env_allowlist` (#3238) against the operator's live
@@ -11774,20 +11800,34 @@ mod tests {
         );
     }
 
-    /// `CLAUDE_CONFIG_DIR` is a host filesystem path, not a value, so
-    /// it must NOT be auto-forwarded into the container even when set
-    /// on the host. The agent's config dir is bind-mounted at the
-    /// canonical container path by `AGENT_CONFIG_MOUNTS`.
+    /// A host filesystem path is not a credential, so it must NOT be
+    /// auto-forwarded into the container even when set on the host and even
+    /// when the adapter's `env_allowlist` names it (#3238). The path resolves
+    /// to nothing inside the container, so forwarding it points the adapter
+    /// away from the config dir `AGENT_CONFIG_MOUNTS` bind-mounts at the
+    /// canonical container location. `CLAUDE_CONFIG_DIR` established the rule;
+    /// `CODEX_HOME` and `GOOGLE_APPLICATION_CREDENTIALS` reach the same
+    /// function through the per-adapter allowlists.
     ///
     /// Tagged `#[serial]` because the test mutates the process-wide
     /// env; parallel readers of `std::env::var` would race.
     #[test]
     #[serial_test::serial]
-    fn build_sandbox_docker_argv_drops_host_only_claude_config_dir() {
-        // Set the env var to simulate the host having it; the function
-        // under test must still skip it.
-        let prev = std::env::var("CLAUDE_CONFIG_DIR").ok();
-        std::env::set_var("CLAUDE_CONFIG_DIR", "/Users/operator/.claude");
+    fn build_sandbox_docker_argv_drops_host_only_path_env() {
+        // Set the vars to simulate the host having them; the function under
+        // test must still skip every one. `OPENAI_API_KEY` rides along as the
+        // control: a value-typed allowlist entry alongside them must still
+        // cross, or the assertion below would pass on a function that forwards
+        // nothing at all.
+        let _env = crate::session::test_support::EnvGuard::set(&[
+            ("CLAUDE_CONFIG_DIR", "/Users/operator/.claude"),
+            ("CODEX_HOME", "/Users/operator/.codex"),
+            (
+                "GOOGLE_APPLICATION_CREDENTIALS",
+                "/Users/operator/gcp-key.json",
+            ),
+            ("OPENAI_API_KEY", "sk-test-value"),
+        ]);
         let tmp = tempfile::tempdir().unwrap();
         let sandbox = SandboxInfo {
             enabled: true,
@@ -11800,13 +11840,18 @@ mod tests {
             container_workdir: None,
         };
         let config = SpawnConfig {
-            agent_key: "claude".into(),
-            tool: "claude".into(),
+            agent_key: "codex".into(),
+            tool: "codex".into(),
             spec: AgentSpec {
-                command: "claude-agent-acp".into(),
+                command: "codex-acp".into(),
                 args: vec![],
                 description: "test".into(),
-                env_allowlist: None,
+                env_allowlist: Some(vec![
+                    "CLAUDE_CONFIG_DIR".into(),
+                    "CODEX_HOME".into(),
+                    "GOOGLE_APPLICATION_CREDENTIALS".into(),
+                    "OPENAI_API_KEY".into(),
+                ]),
             },
             cwd: tmp.path().to_path_buf(),
             additional_dirs: vec![],
@@ -11825,27 +11870,35 @@ mod tests {
         };
         let argv = build_sandbox_docker_argv(&config, &sandbox, "/workspace/proj")
             .expect("docker argv built");
-        match prev {
-            Some(v) => std::env::set_var("CLAUDE_CONFIG_DIR", v),
-            None => std::env::remove_var("CLAUDE_CONFIG_DIR"),
+
+        for key in [
+            "CLAUDE_CONFIG_DIR",
+            "CODEX_HOME",
+            "GOOGLE_APPLICATION_CREDENTIALS",
+        ] {
+            assert!(
+                !argv.docker_args.iter().any(|a| a == key),
+                "{key} is a host path and must not be forwarded as `-e KEY`"
+            );
+            assert!(
+                !argv
+                    .docker_args
+                    .iter()
+                    .any(|a| a.starts_with(&format!("{key}="))),
+                "{key} must not appear as a literal `KEY=VALUE` either"
+            );
+            assert!(
+                !argv.inherit_env.iter().any(|(k, _)| k == key),
+                "{key} must not land in inherit_env"
+            );
         }
-        assert!(
-            !argv.docker_args.iter().any(|a| a == "CLAUDE_CONFIG_DIR"),
-            "CLAUDE_CONFIG_DIR is a host path and must not be forwarded as `-e KEY`"
-        );
-        assert!(
-            !argv
-                .docker_args
+        assert_eq!(
+            argv.inherit_env
                 .iter()
-                .any(|a| a.starts_with("CLAUDE_CONFIG_DIR=")),
-            "CLAUDE_CONFIG_DIR must not appear as a literal `KEY=VALUE` either"
-        );
-        assert!(
-            !argv
-                .inherit_env
-                .iter()
-                .any(|(k, _)| k == "CLAUDE_CONFIG_DIR"),
-            "CLAUDE_CONFIG_DIR must not land in inherit_env"
+                .find(|(k, _)| k == "OPENAI_API_KEY")
+                .map(|(_, v)| v.as_str()),
+            Some("sk-test-value"),
+            "the value-typed allowlist entry must still cross the boundary"
         );
     }
 
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -3923,7 +3923,11 @@ fn apply_env_filter(cmd: &mut std::process::Command, config: &SpawnConfig) {
     }
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
-            if name == "AOE_TOKEN" {
+            // Route through the same deny check `provider_env` uses so a
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
+            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
+            // or infra keys under the guise of provider auth.
+            if provider_env_denyreason(name).is_some() {
                 continue;
             }
             if let Ok(value) = std::env::var(name) {
@@ -4394,8 +4398,12 @@ fn spawn_subprocess(config: &SpawnConfig) -> Result<tokio::process::Child, AcpEr
     }
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
-            if name == "AOE_TOKEN" {
-                warn!(target: "acp", "ignoring AOE_TOKEN in agent env allowlist");
+            // Route through the same deny check `provider_env` uses so a
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
+            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
+            // or infra keys under the guise of provider auth.
+            if let Some(reason) = provider_env_denyreason(name) {
+                warn!(target: "acp", "ignoring env allowlist entry '{name}' ({reason})");
                 continue;
             }
             if let Ok(value) = std::env::var(name) {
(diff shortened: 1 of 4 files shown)


src/acp/acp_client.rs
```diff
@@ -15195,6 +15203,56 @@ done
         }
     }
 
+    /// #3238: `AgentSpec.env_allowlist` populated by `with_defaults` must
+    /// reach the agent via `apply_env_filter`. Uses the `aoe-agent` spec (the
+    /// one that would silently no-op if we keyed `env_allowlist_for` on
+    /// `spec.command` — placeholder-templated — instead of the binary token).
```
Seluj78: Blocker: em dash on an added doc-comment line.

`AGENTS.md` Coding Style: "No emdashes or `--` as separators in docs/comments." The rule covers human-authored prose, which this is. Slightly ironic given this PR scrubs four pre-existing em dashes out of the `agent_registry.rs` descriptions.

src/acp/agent_registry.rs
```diff
@@ -129,16 +143,16 @@ impl AgentRegistry {
                 args: vec![],
                 description:
                     "OpenAI Codex CLI via ACP adapter (npm i -g @agentclientprotocol/codex-acp@latest)".into(),
-                env_allowlist: None,
+                env_allowlist: default_env_allowlist("codex-acp"),
```
Seluj78, line 146: Required: this fix silently no-ops for every sandboxed structured-view session.

Sandboxed spawns take `spawn_runner_detached`, which wraps the agent in `docker exec` argv built by `build_sandbox_docker_argv` (`src/acp/acp_client.rs:3505+`). That function forwards a separate hardcoded list, `PROVIDER_AUTH_KEYS = ["ANTHROPIC_API_KEY", "ANTHROPIC_AUTH_TOKEN", "CLAUDE_CODE_OAUTH_TOKEN"]`, plus `collect_environment` and `provider_env`, as `-e KEY` flags. It never reads `config.spec.env_allowlist`. `apply_env_filter` does set the allowlist keys on the runner process env, but `docker exec` only crosses a name it was given via `-e`, so nothing reaches the container.

Failure scenario: operator has `OPENAI_API_KEY` in `aoe serve`'s environment, launches a sandboxed `codex` structured-view session. Post-PR behavior is identical to pre-PR, auth fails, exactly the #3238 symptom. No test, no doc note. This is the failure mode the PR body pledges to avoid.

Note the existing comment at `acp_client.rs:3546`: host filesystem paths must not cross the namespace boundary, so `CODEX_HOME` and `GOOGLE_APPLICATION_CREDENTIALS` should stay out of the container even if the value-typed keys are added there.

src/acp/acp_client.rs
```diff
@@ -3923,7 +3923,11 @@ fn apply_env_filter(cmd: &mut std::process::Command, config: &SpawnConfig) {
     }
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
-            if name == "AOE_TOKEN" {
+            // Route through the same deny check `provider_env` uses so a
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
+            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
```
Seluj78: Required: this comment is factually wrong about what it protects.

`provider_env_denyreason` (`acp_client.rs:3191-3213`) denies exactly the empty key, the literal `AOE_TOKEN`, `INFRA_KEYS`, and `LD_*` / `DYLD_*`. It denies no other `AOE_*` key. The actual daemon-to-runner carrier is `ACP_AGENT_ENV = "AOE_ACP_AGENT_ENV"` (`src/process/runner.rs:141`), guarded only by the separate `host_environment_denyreason`, which this code does not call. An existing test at `acp_client.rs:15088` asserts `provider_env_denyreason("AOE_AGENT_MODEL").is_none()`.

Not exploitable today (no production path writes `env_allowlist`, `from_acp_cmd` hardcodes `None`), but the comment asserts a defense the code does not implement, and the PR body repeats the claim. Either fix the comment or add the carrier check so the claim becomes true.

src/acp/acp_client.rs
```diff
@@ -3923,7 +3923,11 @@ fn apply_env_filter(cmd: &mut std::process::Command, config: &SpawnConfig) {
...
+            // or infra keys under the guise of provider auth.
+            if provider_env_denyreason(name).is_some() {
```
Seluj78: Required: this deny predicate is the security-critical line the PR rewrites, and it has no negative test.

`apply_env_filter_forwards_agent_env_allowlist` covers the positive path and one wrong-key negative, both against `apply_env_filter`. Nothing puts `LD_PRELOAD`, `AOE_TOKEN`, or `PATH` into a `spec.env_allowlist` and asserts it is dropped, and `spawn_subprocess` (the second, near-identical site at `:4405` with different logging) is untested entirely, so the two can drift.

`AGENTS.md` Testing Guidelines: missing error path for a risky behavior. One table-driven test over the deny cases covers it in a few lines.

src/acp/acp_client.rs
```diff
@@ -4394,8 +4398,12 @@ fn spawn_subprocess(config: &SpawnConfig) -> Result<tokio::process::Child, AcpEr
     }
     if let Some(extra_allowlist) = &config.spec.env_allowlist {
         for name in extra_allowlist {
-            if name == "AOE_TOKEN" {
-                warn!(target: "acp", "ignoring AOE_TOKEN in agent env allowlist");
+            // Route through the same deny check `provider_env` uses so a
+            // malicious/edited allowlist can't smuggle `AOE_TOKEN`, an
+            // `AOE_*`-prefixed daemon carrier, `LD_*`/`DYLD_*` linker hooks,
```
Seluj78: Required (same defect as `:3928`): `provider_env_denyreason` does not deny `AOE_*` or the `AOE_ACP_AGENT_ENV` carrier. Fix the comment or add the carrier check at both sites.

src/acp/install_hints.rs
```diff
@@ -39,6 +39,55 @@ pub fn npm_package_for(binary: &str) -> Option<&'static str> {
...
+/// spawn time (`supervisor.rs`), so keying on it here would silently
+/// miss the bundled agent.
+pub fn env_allowlist_for(binary: &str) -> &'static [&'static str] {
+    match binary {
+        // Verified from source: acp-worker/aoe-agent/src/index.ts imports
+        // only @ai-sdk/{anthropic,openai,google}. Anthropic is already in
+        // ALWAYS_FORWARD_ENV. @ai-sdk/google reads GOOGLE_GENERATIVE_AI_API_KEY
+        // (NOT GEMINI_API_KEY, which is the gemini-CLI-native name).
+        "aoe-agent" => &["OPENAI_API_KEY", "GOOGLE_GENERATIVE_AI_API_KEY"],
+        // Verified from canonical provider env names: codex reads OpenAI SDK
```
Seluj78: Required: the provenance claim is stronger in `agent_registry.rs:258` than it is here.

`agent_registry.rs:258` calls `codex`, `opencode`, `gemini`, and `aoe-agent` "verified adapters". Only the `aoe-agent` arm is source-verified (line 59, citing `acp-worker/aoe-agent/src/index.ts`). This arm and the `opencode` / `gemini` arms say "Verified from canonical provider env names" and, for `OPENROUTER_API_KEY`, "a common opencode target". That is convention, not verification.

The PR's entire stated rationale for deferring `pi` / `omp` / `kimi` / `vibe` is refusing to ship guessed names, so three of the four shipped arms are held to a weaker bar than the deferral policy claims. Either source-verify those three or relabel the comment and the PR table.

src/acp/install_hints.rs
```diff
@@ -39,6 +39,55 @@ pub fn npm_package_for(binary: &str) -> Option<&'static str> {
...
+/// #3238 and shipping a guess that never matches would silently no-op
+/// the fix. Follow-up: verify each adapter's real reads from its own
+/// package/binary and add its arm.
+///
+/// The key is the friendly binary token used at registration time (e.g.
+/// `"aoe-agent"`), NOT `AgentSpec.command`. `command` for `aoe-agent`
+/// carries a `${aoe_data_dir}/...` placeholder that is substituted at
+/// spawn time (`supervisor.rs`), so keying on it here would silently
+/// miss the bundled agent.
+pub fn env_allowlist_for(binary: &str) -> &'static [&'static str] {
```
Seluj78, line 65: Nit: no structural link between these arms and the registry entries. `covers_every_default_registry_binary` pins install hints for 8 binaries but omits `aoe-agent`, and nothing catches an adapter added to `with_defaults` with no allowlist arm here. Extending that existing test's binary list covers both.

src/acp/agent_registry.rs
```diff
@@ -89,7 +103,7 @@ impl AgentRegistry {
                 description: format!(
                     "Anthropic Claude via the official ACP adapter ({claude_install})"
                 ),
-                env_allowlist: None,
+                env_allowlist: default_env_allowlist("claude-agent-acp"),
```
Seluj78, line 106: Nit: five of these `default_env_allowlist(...)` calls provably return `None` today (`claude-agent-acp`, `vibe-acp`, `pi-acp`, `omp`, `kimi`). Uniform shape is defensible. The fragile part is the `aoe-agent` token literal living in two modules with no shared `const` tying them together.

Seluj78 (changes requested):

## Review summary

The problem is real and documented (#3238), the mechanism reuse is right, and the `spec.command` versus binary-token trap was correctly avoided and pinned with a negative test. What blocks it is that the fix covers only the unsandboxed path.

Also cross-checked with four external models (`gemini-3.1-pro-preview`, `gpt-5.6-sol`, `claude-opus-4-7`, `gpt-5.4`). All four independently claimed this PR leaks host credentials into the Docker sandbox. That claim does not hold and I am recording it here so nobody re-raises it: `docker exec` only crosses a variable it was given via `-e`, and `build_sandbox_docker_argv` never names the allowlist keys. The real defect is the inverse, finding 1 below.

### Blocker

**1. Em dash on an added line, `src/acp/acp_client.rs:15209`.** `AGENTS.md` bans em dashes in docs and comments. Slightly ironic given this PR scrubs four pre-existing em dashes out of the `agent_registry.rs` descriptions.

### Required

**1. The fix silently no-ops for every sandboxed structured-view session.** Sandboxed spawns take `spawn_runner_detached`, which wraps the agent in `docker exec` argv built by `build_sandbox_docker_argv` (`src/acp/acp_client.rs:3505+`). That function forwards a separate hardcoded list, `PROVIDER_AUTH_KEYS = ["ANTHROPIC_API_KEY", "ANTHROPIC_AUTH_TOKEN", "CLAUDE_CODE_OAUTH_TOKEN"]`, plus `collect_environment` and `provider_env`, as `-e KEY` flags. It never reads `config.spec.env_allowlist`. `apply_env_filter` does set the allowlist keys on the runner process env, but `docker exec` only crosses a name it was given via `-e`, so nothing reaches the container.

Failure scenario: operator has `OPENAI_API_KEY` in `aoe serve`'s environment and launches a sandboxed `codex` structured-view session. Post-PR behavior is identical to pre-PR: auth fails, exactly the #3238 symptom. No test, no doc note. This is precisely the failure mode the PR body pledges to avoid ("shipping a guessed key that never matches would silently no-op the fix").

Note the existing comment at `acp_client.rs:3546`: host filesystem paths must not cross the namespace boundary. So `CODEX_HOME` and `GOOGLE_APPLICATION_CREDENTIALS` should stay out of the container even if the value-typed keys are added there.

**2. The added security comment is factually wrong about what it protects** (`src/acp/acp_client.rs:3928` and `:4403`). `provider_env_denyreason` (`acp_client.rs:3191-3213`) denies exactly the empty key, the literal `AOE_TOKEN`, `INFRA_KEYS`, and `LD_*` / `DYLD_*`. It denies no other `AOE_*` key. The actual daemon-to-runner carrier is `ACP_AGENT_ENV = "AOE_ACP_AGENT_ENV"` (`src/process/runner.rs:141`), guarded only by the separate `host_environment_denyreason`, which this code does not call. An existing test at `acp_client.rs:15088` asserts `provider_env_denyreason("AOE_AGENT_MODEL").is_none()`.

Not exploitable today: no production path writes `env_allowlist`, and `from_acp_cmd` hardcodes `None`. But the comment asserts a defense the code does not implement, and the PR body repeats the claim. Fix the comment, or add the carrier check so the claim becomes true.

**3. `docs/structured-view.md` contradicts the new behavior and is not in the diff.** Line 37 still reads: "The structured view always forwards `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, and `CLAUDE_CONFIG_DIR` to the agent. For other agents, set their auth env in the environment that runs `aoe serve` (or the per-session `extra_env` field)." That second sentence describes the workaround this PR removes, and the first is now an incomplete list. Line 32 also still says `bundled (Vercel AI SDK 6)`, the identical claim the PR corrected in `agent_registry.rs:191`. The docs should state which keys each adapter now receives, and that sandboxed sessions do not get them. `AGENTS.md`: "`docs/` is the canonical source for all documentation."

**4. The deny predicate is the security-critical line this PR rewrites, and it has no negative test.** `apply_env_filter_forwards_agent_env_allowlist` covers the positive path and one wrong-key negative, both against `apply_env_filter`. Nothing puts `LD_PRELOAD`, `AOE_TOKEN`, or `PATH` into a `spec.env_allowlist` and asserts it is dropped, and `spawn_subprocess` (the second, near-identical site with different logging) is untested entirely, so the two can drift. `AGENTS.md` Testing Guidelines: missing error path for a risky behavior. One table-driven test over the deny cases covers it in a few lines.

**5. The provenance claim is stronger in one comment than in the other.** `agent_registry.rs:258` calls `codex`, `opencode`, `gemini`, and `aoe-agent` "verified adapters". Only the `aoe-agent` arm is source-verified (`install_hints.rs:59`, citing `acp-worker/aoe-agent/src/index.ts`). The `codex-acp`, `opencode`, and `gemini` arms say "Verified from canonical provider env names" and, for `OPENROUTER_API_KEY`, "a common opencode target". That is convention, not verification. The PR's entire stated rationale for deferring `pi` / `omp` / `kimi` / `vibe` is refusing to ship guessed names, so three of the four shipped arms are held to a weaker bar than the deferral policy claims. Either source-verify those three or relabel the comment and the PR table.

### Nit

**1.** `env_allowlist_for` arms have no structural link to registry entries (`install_hints.rs:57`). `covers_every_default_registry_binary` pins install hints for 8 binaries but omits `aoe-agent`, and nothing catches an adapter added to `with_defaults` with no allowlist arm. Extending that existing test's binary list covers both.

**2.** Five `default_env_allowlist(...)` calls provably return `None` today (`claude-agent-acp`, `vibe-acp`, `pi-acp`, `omp`, `kimi`). Uniform shape is defensible; the `aoe-agent` token literal living in two modules with no shared `const` is the fragile part.

### PR metadata

**1.** Two em dashes in the PR body, in the "Verified locally" paragraph and the "AI Usage" paragraph. `AGENTS.md` Commit and Pull Request Guidelines. The commit message itself is clean.

**2.** "Documentation was updated where necessary" is ticked, but no docs file is in the diff and `docs/structured-view.md` is stale in two places. Corroborates Required 3.

### Verdict

**Project Fit:** Accept
**Placement:** n/a (bug fix, core ACP substrate)

Fix or explicitly scope the sandbox gap, correct the deny-check comment, add the negative deny test, and update `docs/structured-view.md`, and this lands.

njbrake (changes requested):
The mechanism is right, and it is the one I want over #3324: additive, and it fills the `env_allowlist` field both spawn paths already read rather than adding a second map beside it. What needs another pass is the key lists themselves, since that is what the PR is.

**The per-adapter keys do not match what the adapters read.** I pulled the artifacts:

- codex-acp 1.3.0 `readApiKeyFromEnv` checks `CODEX_API_KEY` first, then `OPENAI_API_KEY`. `CODEX_API_KEY` is missing. `OPENAI_BASE_URL` is not read by the adapter, though it does reach the codex CLI, which supports it.
- `@ai-sdk/openai@4.0.27` reads `OPENAI_BASE_URL` as well as the key (`src/openai-provider.ts:197`), so `aoe-agent` should have it. Today the codex arm has it and the arm that actually reads it does not.
- opencode resolves providers through models.dev, whose `google` entry declares `GOOGLE_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY`, and `GEMINI_API_KEY`. A user with `GEMINI_API_KEY` exported still fails silently.
- `@google/gemini-cli@0.55.1` selects Vertex on `GOOGLE_GENAI_USE_VERTEXAI` and pairs the project with `GOOGLE_CLOUD_LOCATION`. Without both, the Vertex path the comment claims stays dead.

This matters because the stated reason for deferring `pi` / `omp` / `kimi` / `vibe` is refusing to ship a key that silently no-ops, and three of the four shipped arms were labeled verified when they were conventional.

**The `AOE_*` comment is wrong at both sites.** `provider_env_denyreason` denies the empty key, literal `AOE_TOKEN`, `INFRA_KEYS`, and `LD_*` / `DYLD_*`. It denies no other `AOE_*` key; `AOE_ACP_AGENT_ENV` is guarded only by `host_environment_denyreason`. Not exploitable, but the comment and the PR body assert a defense that is not there.

**Docs.** `docs/structured-view.md` line 37 is now false, and it is the paragraph that tells users to do the thing #3238 says does not work. Please update it with the per-adapter set. Also, `Vercel AI SDK 6` appears in six places; changing one leaves the tree disagreeing with itself, so either sweep them or drop that edit.

Sandboxed sessions are still unfixed, since `build_sandbox_docker_argv` never reads `env_allowlist`. I am fine treating that as out of scope, because `sandbox.environment` is the documented container boundary, but say so in the PR and in the docs rather than leaving it silent. If you do forward anything there later, `CODEX_HOME` and `GOOGLE_APPLICATION_CREDENTIALS` are host paths and must not cross.

Also the em dash on the new doc comment at `acp_client.rs:15213`, per AGENTS.md.

I have these fixes locally with fmt, clippy, and the touched tests green. Happy to push them to your branch if that is easier than redoing them.

MatthewWolff:
Thanks for the thorough review. Pushed 4c003926 and e1592039 addressing all of it.

Sandboxed no-op (Required): build_sandbox_docker_argv now forwards spec.env_allowlist. I pulled the allowlist resolution into a shared allowlisted_env_pairs() helper and routed all three spawn paths (detached runner, in-proc stdio, docker-exec wrap) through it, so the sandbox forwards each allowlisted host value as `-e KEY` with the value on inherit_env (never `-e KEY=VALUE`, which would leak it into argv), and the two non-sandbox sites no longer duplicate the filter so they can't drift.

Missing negative coverage (Required): added apply_env_filter_drops_denied_allowlist_entries (AOE_TOKEN and LD_PRELOAD dropped even when allowlisted) and build_sandbox_docker_argv_forwards_env_allowlist (value rides inherit_env, a denied linker hook stays out of both argv and the container).

Precedence: you were right that this was worth more than a note. The sandbox claimed the ambient host PROVIDER_AUTH_KEYS before provider_env, so a session that selected a different Anthropic credential silently ran under the operator's ambient key inside the container, while the host paths (provider_env applied last) honored the request. Reordered the sandbox claims to provider_env, then the per-adapter allowlist, then the ambient host keys, so first-claim-wins matches the host precedence, with a test (build_sandbox_docker_argv_provider_env_beats_ambient_host_key).

Nits: added install_hints::AOE_AGENT_BINARY as the single spelling shared by the registry registration and env_allowlist_for (used as the match arm too), so a rename can't drop the agent's keys; and strengthened default_env_allowlists_match_verified_providers to assert the exact set of adapters carrying an allowlist, so adding or removing an arm without updating the verified set fails.

The comment, em-dash, and provenance items were resolved in e90e8f82; also corrected a doc sentence that still claimed the per-adapter list doesn't cross the container boundary, which this change reverses.

njbrake (approved):
Approving. The per-adapter key sets are now source-verified rather than conventional, the sandbox path forwards them without letting host filesystem paths cross the container boundary, and sandbox auth precedence matches the host paths.

Note for the record that I pushed two commits to this branch myself, `e90e8f82` (corrected key sets, comment fixes, docs table) and `9d080db4` (the host-only path guard), so this approval covers work I had a hand in. @Seluj78's review predates all three of the last commits; his sandbox finding was addressed by `4c003926`, and his shared-token and structural-link nits by `e1592039`. I am merging with admin rather than dismissing his review, so it stays on the record.

Thanks for the turnaround on this, and for catching the sandbox precedence bug where ambient `PROVIDER_AUTH_KEYS` beat a per-session `provider_env` credential. That one predates this PR and was worth fixing here.

One cosmetic item left over, not worth holding the merge: `Vercel AI SDK 6` still appears in five other places (`docs/structured-view.md` lines 31 and 76, `acp-worker/aoe-agent/package.json`, `acp-worker/aoe-agent/src/index.ts`, `src/acp/mod.rs`, `src/acp/agent_profiles.rs`). Happy to take a follow-up that sweeps them.

Issue #3258, opened 2026-08-05
telemetry: session_total and every session census field count trashed sessions


Seluj78:
**Note:** I found this while rebuilding the PostHog usage dashboard and had Claude write it up. The finding and the decision on how to fix it are mine. - @Seluj78

## What

Every session census field in `usage_snapshot` counts **trashed** sessions. `session_total` is the one that matters most, because "average sessions per install" is a headline number on the usage dashboard, and it currently reads high by however much trash each install is sitting on.

## Why

`aggregate_instances` takes the raw instance slice and reports its length with no trash filter:

```rust
// src/telemetry/mod.rs:348
total: instances.len() as u32,
```

Both callers hand it the full list:

- `build_serve_snapshot` uses `state.instances.read().await.clone()` (`src/server/mod.rs:3966`)
- `build_telemetry_snapshot` uses `self.home.instances()` (`src/tui/app.rs:2123`)

Neither list excludes trash. Trashed rows stay in `state.instances` until `trashed_at + session.trash_retention_days` expires; the API filters them per request instead, via `SessionScope` (`src/server/api/sessions.rs:608`), and the TUI filters at render time (`src/tui/home/mod.rs:4975`). Telemetry is the one consumer that never applies that filter.

The same slice feeds `session_running` / `session_idle` / `session_error`, `session_sandboxed`, `session_yolo`, `sessions_by_agent`, `sessions_by_model_bucket`, and `sessions_by_substrate`, so a trashed session keeps voting in the agent and substrate mix too. `peak_concurrent_sessions` inherits it through the aggregator.

Note `session_archived` already exists and is reported separately, so the schema has precedent for naming a triage state, and archived sessions are at least visible in the data. Trash is not.

## Why it can't be fixed downstream

Nothing on the wire distinguishes a trashed session from a live one, so no PostHog query can subtract them. It has to change in the client.

## Options

1. **Filter trash out of the census.** One `.filter(|i| !i.is_trashed())` at the top of `aggregate_instances`, or in both callers. Makes every existing field mean "live sessions", which is what every dashboard tile already assumes. Silently changes the meaning of historical series, so it wants a `SCHEMA_VERSION` bump and a note.
[...]

PR #3361, opened 2026-08-13
feat(cli): filter and expose session state in `aoe list`

### LINKED ISSUE #3350: cli: `aoe list` has no state filter and omits trashed_at, so a trashed session reads as an errored one
## What

`aoe list` has no state filter, and its `--json` records carry no state field — so a
**trashed** session is indistinguishable from a **failed** one to anything consuming the
CLI.

The REST API already solved this. `GET /api/sessions?state=live|trashed|all` landed in
#3187 and is documented in `docs/api.md`, down to rejecting an unrecognised value with
`400` "so a typo surfaces instead of silently returning every session". The CLI is simply
a different consumer that never got the same treatment.

## Measured

aoe 1.14.0 on Linux. Create a session, `aoe remove` it (to trash, not purge), then look at
it from every angle the CLI offers:

```
$ aoe list --json          # still listed; the record is
{"id":"…","title":"zz-trash-probe","path":"…","group":"zztest","tool":"claude",
 "command":"claude --permission-mode auto","profile":"personal",
 "created_at":"…","workspace_repos":[]}      # no state, no trashed_at

$ aoe session show <id>
  Status:  Error                              # same as a genuinely failed session

$ aoe status --json
{"waiting":0,"running":2,"idle":62,"stopped":0,"error":4,"total":68}   # trash counted

$ aoe add <same title+path>
Session already exists with same title and path
```

The state does exist in the store — `profiles/<profile>/sessions.json` carries
`trashed_at` alongside `status` — it just does not reach the CLI's output.

Checked against **v1.14.1**: `src/cli/list.rs` still declares `ListArgs { json, all }`, so
this is current rather than something my pin is behind on.

## Why it bites

A trashed session **keeps its title**, and `aoe add` refuses a duplicate title+path. So an
external tool that provisions sessions sees this: the title it wants is listed as though
live, creating is refused, and attaching would resume something the user threw away —
with no way to tell which situation it is in. `aoe session list-trash` does exist, so a
consumer can in principle shell out a second time and cross-reference by title — but it
prints a table with no `--json`, and there is no archived equivalent at all.

That is the same shape as #3156's item 1 ("an external supervisor keying on 'does my
session still exist / what's its status' will read a trashed session as alive"), which was
accepted and fixed for the API. I hit it building a lane-management tool that shells out
to `aoe add` / `aoe session start` and matches sessions by title.

It also looks like the class already fixed elsewhere, surface by surface: #3258 (telemetry
counting trashed sessions in `session_total`), #2731 (TUI sidebar group count disagreeing
with the list), #2530. I could not find an existing issue covering the CLI listing itself.

## Suggested direction

Mirror what the API already does, rather than inventing a second vocabulary:

1. `aoe list --state <live|trashed|all>`, defaulting to today's unfiltered behaviour so
   nothing breaks, with the same `400`-style rejection of an unrecognised value.
2. Include `trashed_at` (and ideally `archived_at`) in the `--json` record, so a consumer
   that wants to distinguish states without re-running the command can.

(2) alone would unblock my case; (1) is the part that matches the API and the TUI.

Two things I would not want to see chosen by accident: the `status` field is PascalCase on
the wire but lowercase in the CLI and `[status_hooks]`, which `docs/api.md` already warns
about — worth not widening. And `archive`, `snooze` and `favorite` are in the same
position as trash: modelled by the tool, absent from `list --json`.

Happy to put up a PR for either or both if the direction is right — it looks contained to
`ListArgs`, `session_json()` and the filter the API already has. Wanted to check the
approach here first, per CONTRIBUTING.

## Environment

- aoe 1.14.0 (verified against v1.14.1 source), Linux, tmux substrate, `claude` sessions
- 68 sessions in the profile, several trashed — which is how this surfaced

MatthewWolff:
## Description

Closes #3350.

`aoe list --json` records carried no `state` field and no `trashed_at`/`archived_at` timestamps, so a scripted consumer that provisions sessions could not distinguish a trashed session from a genuinely failed one. `aoe list` still shows the trashed row (title, path, id, etc.), so an external tool sees a title-collision on `aoe add` and cannot tell whether the existing row is live, trashed, or archived without a second `aoe session list-trash` shellout — and that command has no `--json` output either.

The REST API already solved this in #3187 with `GET /api/sessions?state=live|trashed|all`, rejecting an unrecognized value with a `400`. This PR mirrors that vocabulary in the CLI.

## Changes

- New `aoe list --state <live|trashed|all>` clap flag, defaulting to `all` (today's unfiltered behavior, preserved verbatim so nothing scripted against v1.14.1 breaks). Clap rejects an unrecognized value at parse time, mirroring the API's `400` behavior.
- New fields on the `SessionJson` record: `state: "live" | "trashed" | "archived"` (always present, one new key on the wire) plus `trashed_at` and `archived_at` (both `skip_if_none`, so a live row's JSON adds only the `state` key relative to v1.14.1).
- Human-readable table gains a `STATE` column only under `--state=all` where rows are mixed. Under `--state=live` or `--state=trashed` every row has the same state, so the column would carry no information and the header stays as it was.
- New `src/session/scope.rs` moves the API's `SessionScope` enum out of `src/server/api/sessions.rs` into a bare-core module so the CLI (bare-core) and the serve daemon share one source of truth and cannot drift. The API's inline `instance_matches_scope` collapses into `SessionScope::matches`.

## Follow-ups the issue mentions but this PR does not cover

The issue reporter listed a few adjacent gaps that would be well-suited to follow-up PRs; this PR is deliberately scoped to the `aoe list` changes it names:

- Archive-only, snooze, and favorite filters on `aoe list` (`--state` covers `trashed` and `archived` as observable state today; distinct filters would let a consumer split them further).
- Matching `--state` / `--json` treatment on `aoe session show` and `aoe session list-trash`.
- Applying the `state` field to `aoe status --json` counts so trashed rows are not double-counted (adjacent to #3258).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

Titled `feat(cli):` because the observable behavior of `aoe list` (new flag, new JSON keys) is additive, but the motivating class of bug is the trashed/failed indistinguishability the issue is filed under. Either prefix would be defensible; happy to rename the commit before merge if the maintainer prefers.

## Checklist

CLI change, no visual surface. `docs/cli/reference.md` regenerated via `cargo xtask gen-docs` and committed.

## Test Coverage Analysis

The REST API's existing behavior is unchanged; the API's `SessionScope` enum moved to a shared module but keeps its serde behavior (round-trip tests preserved on the shared type). No dashboard control flow moves.

TUI / CLI (e2e test)

The change adds an optional flag and additional JSON keys; no lifecycle. Full unit coverage on the state tag, JSON output shape, backward-compat default, and the shared filter predicate.

Tests added (all bare-core, no `#[serial]` needed since none touch `HOME`/env):

- `src/session/scope.rs`: `SessionScope` deserializes lowercase wire names; rejects unrecognized values (mirrors API's #3187); `matches` predicates for no-scope / live / trashed.
- `src/cli/list.rs`: `state_tag` covers all three states; `session_json` exposes state+timestamp for trashed and archived rows; a live row's JSON omits `trashed_at`/`archived_at` (backward-compat with pre-#3350 parsers that don't expect the keys); `default_state_is_all_for_backward_compat` guards against a future refactor flipping the default to `live` and dropping trashed rows for anyone scripted against today's output.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings` (both featuresets), `cargo test --lib` (4500 pass), `cargo test --no-default-features --lib` (4500 pass), `cargo xtask gen-docs` (committed).

## AI Usage

- [ ] No AI was used
- [x] AI was used

Skipped the multi-agent chain for this issue: the issue body pinpoints the exact touch points (`ListArgs`, `session_json`, the API's `SessionScope`) and the design is a small mirror of the already-shipped `GET /api/sessions?state=` from #3187. I read the API's implementation, moved `SessionScope` to a bare-core module, and wired the CLI through it. I reviewed the whole diff, understand it, and will answer reviewer feedback in my own words.

## Summary

- Added `aoe list --state <live|trashed|all>`, with `all` as the default.
- Added session state and optional `trashed_at` and `archived_at` fields to JSON output.
- Added a `STATE` column to human-readable listings when needed.
- Centralized `SessionScope` for shared use by the CLI and REST API.
- Added validation, documentation, and unit and end-to-end test coverage.

## Benefits

- Consumers can identify live, trashed, and archived sessions without extra CLI calls.
- Existing listing behavior remains unchanged by default.
- Shared state handling reduces duplicated CLI and API logic.

## Potential follow-up

- Add state handling for archive-only, snooze, favorite, and related commands.

### DIFF
--- a/contrib/hermes-skill/SKILL.md
+++ b/contrib/hermes-skill/SKILL.md
@@ -68,6 +68,7 @@ aoe add . -t "yolo" -y -l
 aoe list              # human-readable
 aoe list --json       # JSON for parsing
 aoe list --all        # across all profiles
+aoe list --json --state=live   # skip trashed and archived rows
 ```
 
 **JSON shape** (`aoe list --json`):
@@ -81,13 +82,14 @@ aoe list --all        # across all profiles
     "tool": "claude",
     "command": "claude",
     "profile": "default",
+    "state": "live",
     "created_at": "2025-01-01T00:00:00Z",
     "workspace_repos": []
   }
 ]
 ```
 
-`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `list --json` does not include live status: use `aoe status --json` or `aoe session capture --json` for that.
+`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `state` is `live`, `archived`, or `trashed`, with `trashed_at` / `archived_at` present only in those states; a trashed session stays in the listing and keeps its title, so check `state` before treating a row as a live session. `list --json` does not include live status: use `aoe status --json` or `aoe session capture --json` for that.
 
 ## Session Lifecycle
 
--- a/contrib/openclaw-skill/SKILL.md
+++ b/contrib/openclaw-skill/SKILL.md
@@ -69,6 +69,9 @@ aoe list --json
 
 # List across all profiles
 aoe list --all
+
+# Skip trashed and archived rows (default is every persisted session)
+aoe list --json --state=live
 ```
 
 **JSON output shape** (`aoe list --json`):
@@ -82,13 +85,14 @@ aoe list --all
     "tool": "claude",
     "command": "claude",
     "profile": "default",
+    "state": "live",
     "created_at": "2025-01-01T00:00:00Z",
     "workspace_repos": []
   }
 ]
 ```
 
-`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `list --json` does not include live status; use `aoe status --json` or `aoe session capture --json` for that.
+`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `state` is `live`, `archived`, or `trashed`, with `trashed_at` / `archived_at` present only in those states; a trashed session stays in the listing and keeps its title, so check `state` before treating a row as a live session. `list --json` does not include live status; use `aoe status --json` or `aoe session capture --json` for that.
 
 ### Session lifecycle
 
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -242,7 +242,7 @@ List all sessions
 
 * `--json` — Output as JSON
 * `--all` — List sessions from all profiles
-* `--state <STATE>` — Filter by session state. Default is `all` (every persisted session, which is what `aoe list` has always shown). #3350 added the flag so a scripted consumer can pass `--state=live` to skip trashed and archived rows without a second `aoe session list-trash` shellout, mirroring the REST API's `state=live|trashed|all` from #3187. Rejects any other value at parse time
+* `--state <STATE>` — Filter by session state. Defaults to `all`, every persisted session, which is what `aoe list` has always shown. Pass `--state=live` to skip trashed and archived rows; the vocabulary matches the REST API's `GET /api/sessions?state=`
 
   Default value: `all`
 
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -242,6 +242,18 @@ List all sessions
 
 * `--json` — Output as JSON
 * `--all` — List sessions from all profiles
+* `--state <STATE>` — Filter by session state. Default is `all` (every persisted session, which is what `aoe list` has always shown). #3350 added the flag so a scripted consumer can pass `--state=live` to skip trashed and archived rows without a second `aoe session list-trash` shellout, mirroring the REST API's `state=live|trashed|all` from #3187. Rejects any other value at parse time
+
+  Default value: `all`
+
+  Possible values:
+  - `live`:
+    Only sessions that are neither archived nor trashed
+  - `trashed`:
+    Only sessions currently in the trash
+  - `all`:
+    Every persisted session in the profile (default)
+
 
 
 
--- a/src/cli/list.rs
+++ b/src/cli/list.rs
@@ -1,15 +1,43 @@
 //! `agent-of-empires list` command implementation
 
 use anyhow::Result;
-use clap::Args;
+use clap::{Args, ValueEnum};
 use serde::Serialize;
 
-use crate::session::{Instance, Storage};
+use crate::session::{Instance, SessionScope, Storage};
 
 const TABLE_COL_TITLE: usize = 20;
 const TABLE_COL_GROUP: usize = 15;
 const TABLE_COL_PATH: usize = 40;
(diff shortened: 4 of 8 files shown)


njbrake (approved):
Approving. Two follow-ups below are taste calls, not blockers.

I pushed two commits to this branch rather than round-tripping them:

`5026591f` fixes the one thing I would have blocked on. `run` checked `instances.is_empty()` before the `--json` branch, so `aoe list --json --state=live` printed `No sessions found in profile '<p>'.` on stdout when the filter emptied the result. That was reachable only on a genuinely empty profile before this PR; `--state=live` makes it routine with sessions present, which is the scripted consumer #3350 exists for. `aoe ps --json` and this command's own `--all` path already emitted `[]`, so the two branches disagreed with each other. Same commit derives `state_tag` from `Instance::effective_bucket()` instead of re-deriving the `Trashed > Archived > Active` precedence a third time; corrects two contract comments (`SessionScope::Live` excludes archived rather than folding it in, and `trash()` leaves `archived_at` set, so trash is orthogonal to archive rather than a superset); drops four orphaned lines of the old docstring left spliced into `ListSessionsQuery`; and trims the issue numbers out of the `--state` help text, which clap renders into `aoe list --help`. An e2e test walks the whole contract. Thanks for tightening its last assertion in b56f8ca.

`7748a15c` adds `state` to the `aoe list --json` shapes in the OpenClaw and Hermes skills. Both spell the record out field by field and went stale here; `check-skill` passes regardless because it only validates command names. Those skills are written for agents that provision sessions by title, so they are the consumer the issue describes.

Two I left for you:

1. `StateFilter` is a second enum shadowing `SessionScope` with a `From` impl, in a PR whose stated point is one source of truth. The comment justifies it as clap needing its own `PossibleValue` list, but `AuthMode` in `src/cli/serve.rs` carries `ValueEnum` and serde on one type. The real tradeoff is that deriving `ValueEnum` on `SessionScope` puts a clap derive on a `src/session/` core type, which nothing does today.
2. `table_shows_state` keys off the scope rather than the data, so plain `aoe list` now renders a `STATE` column reading `live` on every row for anyone with no trashed or archived sessions.

Mirroring the API vocabulary instead of inventing a second one was the right instinct, and the `SessionScope` hoist is the part I most wanted.

PR #3363, opened 2026-08-14
fix(web): fall back to main-thread highlighting when the diff worker fails to load

### LINKED ISSUE #3362: Web dashboard: diff pane renders blank when the Pierre highlighter worker fails, no fallback to main-thread highlighting

## Problem

Opening any file in the web dashboard's diff pane rendered a blank body. The file header (`Added src/primmo/config.py +92`), the Unified/Split toggle, the Find button, and the 68-file changed-file list all rendered normally; only the diff lines were missing. Toggling between Unified and Split changed nothing. Switching files changed nothing. A full page reload fixed it.

Firefox Developer Edition 154 on macOS. The same session rendered fine in Chrome at the same moment, against the same daemon.

The browser console carried exactly one relevant error, repeated twice:

```
Worker error:
error { target: Worker, isTrusted: true, srcElement: Worker, eventPhase: 0, ... }
Object { worker: Worker, requestId: undefined, pendingSetupRequestId: undefined,
         initialized: false, langs: Set(1), customExtensionsVersion: 0 }
index-DEUx2mUr.js:163:34951
```

That is the Pierre highlighter worker pool failing during setup (`initialized: false`).

## What ruled itself out

**The backend is correct.** Against the live daemon for session `7b954a87cfee423b`:

```
/api/sessions/7b954a87cfee423b/diff/files        -> 68 files, all status "added"
/api/sessions/7b954a87cfee423b/diff/file?path=src/primmo/config.py
  old_content: ""      (0 bytes)
  new_content: 3521 bytes
  patch:       3771 bytes, well formed:
    diff --git a/src/primmo/config.py b/src/primmo/config.py
    index e69de29..e3cc470 100644
    --- a/src/primmo/config.py
    +++ b/src/primmo/config.py
    @@ -0,0 +1,92 @@
```

**Not a stale bundle.** The failing page ran `/assets/index-DEUx2mUr.js`, which is byte-for-byte what the daemon serves right now; `/assets/wasm-BnjxR4X6.js` (the Shiki oniguruma chunk referenced in the console's source-map warning) returns 200 with `cache-control: public, max-age=31536000, immutable`. The `DashboardUpdateBanner` stale-build path (`web/src/components/DashboardUpdateBanner.tsx:23-46`) is not implicated.

**Not the service worker.** `web/public/sw.js` registers no `fetch` handler and precaches nothing (`web/public/sw.js:21-23`); its `activate` handler deletes every cache. A Playwright Firefox run that registered the service worker, reloaded under its control, and opened the same file rendered the diff correctly.

**Not browser-wide.** Playwright Firefox 151 and Chromium, clean profiles, both render this exact session's diff. The failure needs the worker to actually error, which it does not do on a cold profile.

## Root cause

The diff body is rendered only through the Pierre worker pool, and nothing handles a worker that constructs successfully and then errors.

`web/src/components/diff/pierre/DiffWorkerPoolProvider.tsx:22-33`:

```tsx
export function DiffWorkerPoolProvider({ children }: { children: ReactNode }) {
  const { theme } = useShikiTheme();

  if (typeof Worker === "undefined") {
    return <>{children}</>;
  }

  return (
    <WorkerPoolContextProvider key={theme} poolOptions={{ workerFactory, poolSize: 4 }} highlighterOptions={{ theme }}>
      {children}
    </WorkerPoolContextProvider>
  );
}
```

The only degradation path is "`Worker` does not exist at all" (SSR, jsdom). The comment at lines 18-20 states the intent correctly: when the pool is unavailable the children should highlight on the main thread, preserving correctness. That intent is not honored for a runtime worker failure, which is the common case in a real browser.

`web/src/components/diff/DiffFileViewer.tsx:555-569` then renders the diff unconditionally inside that provider, with no error boundary:

```tsx
<div ref={scrollResetRef} className="flex-1 min-h-0 flex flex-col">
  <DiffWorkerPoolProvider>
    <Virtualizer key={viewKey} className="flex-1 overflow-auto">
      {fileDiff && (
        <FileDiff<AnnotationMeta> fileDiff={fileDiff} options={options} ... />
      )}
    </Virtualizer>
  </DiffWorkerPoolProvider>
</div>
```

Two consequences follow:

1. **No fallback.** A pool that never reaches `initialized: true` leaves `FileDiff` with nothing to paint. The patch parse itself already succeeded upstream: `processFile` at `DiffFileViewer.tsx:174-181` is pure text parsing and does not touch the worker, which is why the header, the `+92` counters, and the layout toggles all worked while the body stayed empty.
2. **No signal and no recovery.** The user sees an empty pane with no error text, no retry affordance, and no hint that reloading fixes it. Every subsequent file open in that page reuses the same dead pool, so the pane stays blank for the lifetime of the tab.

The trigger for the worker error itself is secondary and, on this evidence, intermittent and Firefox-specific. The defect worth fixing is that a highlighting failure degrades the diff to nothing instead of to unhighlighted text.

## Proposal

**A. Fall back to main-thread highlighting when the pool fails.** Track worker health in `DiffWorkerPoolProvider`: attach an `error` handler in `workerFactory` (or subscribe to the pool's error surface) and, on failure, flip to the same `return <>{children}</>` branch already used for `typeof Worker === "undefined"`. Worst case becomes a slower diff, never a blank one. This is the minimum fix and the one that makes the symptom unreachable regardless of what made the worker die.

**B. Surface the degradation.** When the fallback engages, render a one-line notice in the diff header area ("syntax highlighting unavailable, showing plain text") so the state is legible instead of silently different. Keeps the pane useful without a reload.

**C. Recreate the pool once before giving up.** `poolSize: 4` with a single shared provider means one bad start poisons the whole session. One retry with a fresh worker before falling back to (A) would recover the transient case, which is what a manual reload does today by accident.

**D. Investigate the Firefox trigger separately.** Worth its own issue if it recurs: Firefox Developer Edition 154, module worker (`new Worker(url, { type: "module" })`, `DiffWorkerPoolProvider.tsx:9-11`) loading a 622 KB Shiki wasm chunk under `script-src 'self' 'wasm-unsafe-eval'`. Not reproducible on Playwright Firefox 151, so a 152+ regression is possible. (A) makes this non-urgent.

## Why now

- The diff pane is the primary review surface in the dashboard; a silent permanent blank there reads as "AoE is broken", not "highlighting failed".
- The fix is confined to one small component with an already-written fallback branch.
- The failure is invisible to the user: no error, no retry, no hint that reloading helps.

- As a dashboard user, when the diff highlighter worker fails to initialize, then the diff still renders with unhighlighted text instead of a blank pane.
  - Suggested test layer: Vitest + RTL, stubbing `workerFactory` to produce a worker that emits `error`.
- As a dashboard user, when the worker pool errors, then I see a visible signal that highlighting is degraded rather than silence.
  - Suggested test layer: Vitest + RTL, asserting the notice renders in the fallback branch.
- As a dashboard user, when I open a newly added file's diff (empty old side, full new side), then its lines render without a page reload.
  - Suggested test layer: mocked Playwright, `page.route()` aborting `**/assets/worker-*.js` to force the failure, then asserting diff lines are present.

## Open questions

- Does `FileDiff` render nothing by design while the pool is uninitialized, or does it throw and get swallowed? I could not force the worker to fail on demand to confirm the exact mechanism; repeated automated sign-ins tripped the dashboard's login rate limit before the experiment landed. The fix in (A) does not depend on the answer, but the test in story 1 does.
- Should the fallback be per-provider (this diff pane) or global for the tab? A single failure is currently tab-wide because the pool is shared.
- Is `poolSize: 4` worth keeping if one worker's failure is fatal to the pool? If the pool degrades on any single worker error, a smaller pool with a retry may be strictly better.

## Out of scope

- Chasing the Firefox 154 worker failure itself (proposal D, separate issue if it recurs).
- Any change to the diff endpoints; the backend output is correct.
- Service worker / PWA caching behavior; ruled out above.

## Cross-references

- #1969 (merged) introduced `@pierre/diffs` with the worker-pool highlighting this issue is about.
- #2152 (closed) covered a different empty-diff case: a genuinely empty diff with no "no changes vs base" signal. This one has 68 files and 9896 additions to render.

Seluj78:

## Description

The diff pane rendered a permanently blank body whenever the syntax-highlighting worker failed to load. The file header, the +/- counters, the unified/split toggle and the changed-file list all rendered normally, so the pane looked alive while showing nothing. Switching files did not help, toggling unified/split did not help, and only a full page reload recovered it. Hit in Firefox Developer Edition 154 while Chrome was fine against the same daemon at the same moment.

The bug is upstream, in `@pierre/diffs@1.2.12`. `WorkerPoolManager.initializeWorkers` resolves each worker's init promise only when that worker posts back an `initialize` success message, and the pool's own worker `error` listener does nothing but `console.error`. A worker that dies at load therefore leaves its init promise pending forever: `initialize()` never settles, `workersFailed` is never set (it is assigned only in the `catch`), and so `isWorkingPool()` keeps reporting the pool healthy. `DiffHunksRenderer` takes the worker branch on that basis and never reaches its main-thread fallback, while `drainQueue` returns early because the manager is not initialized. Nothing is ever painted. Because the pool is a module-level singleton, one poisoned instance blanks every file in the tab, which is why only a reload fixed it.

The fix is downstream and small. `DiffWorkerPoolProvider` now attaches an `error` listener in its worker factory and, on failure, renders its children without the pool provider. Swapping the provider for a fragment is what makes the recovery stick: `useFileDiffInstance` captures the pool reference once when its DOM node attaches, so only a subtree remount rebuilds the renderer pool-free. Unmounting the provider also drops the library's instance count to zero, which terminates the poisoned singleton, so the next fresh mount builds a new pool. That is the retry, and a retry loop in the provider would only re-race the same broken worker.

With no pool the renderer calls `asyncHighlight`, which lazily builds a main-thread highlighter on the JavaScript regex engine, so the fallback still highlights correctly and needs no wasm. The only real cost is main-thread tokenization on large diffs, so the pane also shows one muted line saying highlighting moved off the worker and a reload will retry.

A host-side `error` event is the right trigger rather than a timeout heuristic: the library's worker catches its own per-message failures and reports them as `error` messages instead, so this event fires only when the worker script never ran, which is exactly the fatal case.

One accepted cost, called out for reviewers: the fallback remount replaces the scroll container that `DiffFileViewer`'s scroll effect captured, so target-line positioning is lost for the file that was open at the moment the worker died. The new container starts at the top, which is that effect's own default. Every alternative shape considered replaced that node too, so this is documented in the provider rather than widened into the scroll effect.

Not fixed here: the Firefox-specific trigger that killed the worker in the first place. It is a separate question, and this change makes it non-fatal either way. The upstream defect is filed as pierrecomputer/pierre#1076; once the pool reports its own failure, this workaround can be deleted.

Closes #3362

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open a session with changes in the web dashboard, click a file in the diff pane, confirm the diff renders as before with syntax highlighting.
- [ ] In devtools, block the highlighter worker chunk (Network request blocking on `*/assets/worker-*.js`), reload, then open a file in the diff pane. The diff renders instead of a blank pane, and a muted line reads "Highlighting on the main thread; the worker failed to load. Reload to retry."
- [ ] With the worker still blocked, click a second and third file. Each renders, with no page reload needed.
- [ ] Unblock the worker, reload, confirm the notice is gone and highlighting is back on the worker pool.

## Checklist

The one visual change is a single muted line of text that appears only in the failure path; its exact wording is asserted by both new tests.

Documentation: none warranted. There is no new setting, flag, or user-controllable behavior; the change is a failure-path fallback inside one component.

## Test Coverage Analysis

Both directions were proven rather than asserted green at the end:

- `web/tests/diff-syntax-highlight.spec.ts` aborts `**/assets/worker-*.js` to force the real failure, then asserts the diff content and the notice are visible. On the pre-fix bundle it fails because the diff content never appears at all, which is the reported bug reproduced in a browser.
- `web/src/components/diff/pierre/__tests__/DiffWorkerPoolProvider.test.tsx` drives the same transition at unit level. On the pre-fix provider it fails at the "pool is gone" assertion, since nothing was listening for the worker error.
- `web/tests/coverage-matrix.json` gains the new Vitest spec under the existing `right-panel.diff-render-and-find` surface, which already maps `DiffWorkerPoolProvider.tsx`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude (via Claude Code, the `/aoe-investigate` and `/aoe-implement` skills)

Investigation, plan and implementation were drafted by Claude under my direction, including reading the shipped `@pierre/diffs` dist to find the root cause. I made the design calls, reviewed each commit, and the plan was pressure-tested against two other models before any code was written; their proposed alternatives (relocating the provider inside the virtualizer, and a timeout watchdog using `disableWorkerPool`) were both rejected with reasons recorded in the plan.

**NOTE:**

### DIFF
--- a/web/src/components/diff/pierre/DiffWorkerPoolProvider.tsx
+++ b/web/src/components/diff/pierre/DiffWorkerPoolProvider.tsx
@@ -1,14 +1,6 @@
 import { WorkerPoolContextProvider } from "@pierre/diffs/react";
-import type { ReactNode } from "react";
+import { useCallback, useState, type ReactNode } from "react";
 import { useShikiTheme } from "../../../hooks/useShikiTheme";
-
-/**
- * Creates a Pierre highlighter worker. Vite bundles the worker entry from
- * `@pierre/diffs/worker/worker.js` when referenced via `new URL(..., import.meta.url)`.
- */
-function workerFactory(): Worker {
-  return new Worker(new URL("@pierre/diffs/worker/worker.js", import.meta.url), { type: "module" });
-}
 
 /**
  * Provides a shared off-main-thread highlighter worker pool for the diff
@@ -18,12 +10,54 @@ function workerFactory(): Worker {
  * with the new theme. When `Worker` is unavailable (SSR / jsdom tests) it
  * renders children directly; the diff components then highlight on the main
  * thread (`disableWorkerPool`), preserving correctness.
+ *
+ * The same bare-children fallback also engages when a worker fails to load.
+ * `@pierre/diffs@1.2.12` resolves each worker's init promise only on an
+ * `initialize` success message and its own `error` listener merely logs, so a
+ * worker that dies at load leaves `initialize()` pending forever while
+ * `isWorkingPool()` still reports healthy. The renderer then waits on a pool
+ * that will never drain and paints nothing, tab-wide, because the pool is a
+ * module-level singleton. See #3362.
+ *
+ * Swapping the provider for a fragment is what makes the recovery stick:
+ * `useFileDiffInstance` captures the pool reference once when the DOM node
+ * attaches, so only a subtree remount rebuilds the renderer without a pool.
+ * Unmounting also drops the library's instance count to zero, which
+ * terminates the poisoned singleton, so the next fresh mount builds a new
+ * pool. That is the retry; a retry loop here would just re-race the same
+ * broken worker.
+ *
+ * One known cost: the remount replaces the scroll container that
+ * `DiffFileViewer`'s scroll effect captured, so target-line positioning is
+ * lost for the file that was open when the worker died. The new container
+ * starts at the top, which is that effect's default anyway.
  */
 export function DiffWorkerPoolProvider({ children }: { children: ReactNode }) {
   const { theme } = useShikiTheme();
+  const [workerFailed, setWorkerFailed] = useState(false);
+
+  const workerFactory = useCallback(() => {
+    const worker = new Worker(new URL("@pierre/diffs/worker/worker.js", import.meta.url), { type: "module" });
+    // A host-side `error` here means the worker script never ran: the library's
+    // worker catches its own per-message failures and reports them as `error`
+    // messages instead. So this fires only for the fatal load case.
+    worker.addEventListener("error", () => setWorkerFailed(true), { once: true });
+    return worker;
+  }, []);
 
   if (typeof Worker === "undefined") {
     return <>{children}</>;
+  }
+
+  if (workerFailed) {
+    return (
+      <>
+        <div role="status" className="shrink-0 px-3 py-1 text-[11px] font-mono text-text-dim">
+          Highlighting on the main thread; the worker failed to load. Reload to retry.
+        </div>
+        {children}
+      </>
+    );
   }
 
   return (
--- a/web/src/components/diff/pierre/__tests__/DiffWorkerPoolProvider.test.tsx
+++ b/web/src/components/diff/pierre/__tests__/DiffWorkerPoolProvider.test.tsx
@@ -1,0 +1,82 @@
+// @vitest-environment jsdom
+//
+// A highlighter worker that dies at load used to blank the diff pane forever:
+// `@pierre/diffs@1.2.12` never settles `initialize()` in that case, so the
+// renderer waits on a pool that will never drain. See #3362. The provider must
+// notice the worker error and drop the pool so the renderer falls back to
+// main-thread highlighting.
+//
+// `WorkerPoolContextProvider` is mocked to call the factory the way the real
+// pool does, since the real one only builds workers once a renderer triggers
+// initialization, which does not happen under jsdom.
+
+import { act, render, screen } from "@testing-library/react";
+import { useEffect, useRef } from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { DiffWorkerPoolProvider } from "../DiffWorkerPoolProvider";
+
+class FakeWorker {
+  static instances: FakeWorker[] = [];
+  private listeners: Array<(event: Event) => void> = [];
+
+  constructor() {
+    FakeWorker.instances.push(this);
+  }
+
+  addEventListener(type: string, cb: (event: Event) => void) {
+    if (type === "error") this.listeners.push(cb);
+  }
+  removeEventListener() {}
+  terminate() {}
+
+  /** What the browser does when the worker script fails to load. */
+  failToLoad() {
+    for (const cb of this.listeners) cb(new Event("error"));
+  }
+}
+
+vi.mock("@pierre/diffs/react", () => ({
+  WorkerPoolContextProvider: ({
+    children,
+    poolOptions,
+  }: {
+    children: React.ReactNode;
+    poolOptions: { workerFactory: () => Worker };
+  }) => {
+    const built = useRef(false);
+    useEffect(() => {
+      if (built.current) return;
+      built.current = true;
+      poolOptions.workerFactory();
+    }, [poolOptions]);
+    return <div data-testid="worker-pool">{children}</div>;
+  },
+}));
+
+afterEach(() => {
+  FakeWorker.instances = [];
+  vi.unstubAllGlobals();
+});
+
+describe("DiffWorkerPoolProvider", () => {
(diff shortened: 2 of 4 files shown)


Seluj78:
@coderabbitai review

PR #3381, opened 2026-08-15
fix(session): scope hermes conversation capture to the project

### LINKED ISSUE #3373: hermes conversation capture ignores the project: picks the most recent active conversation regardless of cwd
## Terminology

- **conversation**: a hermes native chat session (a row in hermes' `state.db`), identified on the aoe row by `agent_session_id`.
- **session** (aoe): the aoe row; not what this issue targets.
- **project matching**: scoping a capture to the conversation whose working directory belongs to the aoe session's project, as other agents' captures do.

## Problem

`hermes_poll_fn` reads `~/.hermes/state.db` and returns the most recent active hermes conversation, with no project matching: « capture returns the most recent active CLI session regardless of project » (`src/session/capture.rs:2660-2664`). The sessions table has no cwd column, so the poll cannot tell which conversation belongs to the session's project.

With two hermes sessions on different projects, both capture the same most-recent conversation id unless the cross-instance exclusion set happens to separate them (`retroactive_capture_exclusion_set`, `src/session/instance.rs:3044-3062`); the second session to poll can adopt the first's conversation. Resuming then binds the wrong conversation. Every other capture path validates cwd or project hash (claude encoded-cwd scan, pi header cwd, gemini projectHash, codex rollout first-line cwd, vibe meta.json working_directory) precisely to avoid this.

Comparative context (herdrdev/herdr study): herdr treats hermes as session-identity-only too (`session_identity_only_integration`, `src/detect/mod.rs:310-314`) and does not project-match either, so this is a known-hard area upstream, but aoe's multi-session model makes the mis-attribution observable across sessions.

## Proposal

- Investigate whether hermes persists any project/working-directory signal (transcript file paths, `.hermes` layout, CLI `--resume` behavior) that can scope the capture; if yes, add it to the query in `hermes_poll_fn` (`src/session/capture.rs:2646-2730`).
- If no project signal exists on disk, degrade explicitly: return `None` when more than one active hermes conversation exists (ambiguous), so the agent starts a fresh conversation rather than silently resuming the wrong one, and document the limitation.

- As an AoE user, when two hermes sessions run on different projects, then each resumes its own conversation and never adopts the peer's.
  - Suggested test layer: e2e with a fake hermes state.db seeded with two active conversations (mirror `claude_shared_project_correlation_e2e`), asserting each session's persisted `agent_session_id` is its own.
- As an AoE user, when hermes state is ambiguous (multiple active conversations, no project signal), then the agent starts fresh instead of silently resuming a wrong conversation.
  - Suggested test layer: unit test over `hermes_poll_fn` with a multi-row store asserting `None`.

## Out of scope

- hermes hook sidecar status (separate concern, tracked by its own paths).
- Changing hermes itself to persist cwd.

## Cross-references

- `src/session/capture.rs:2646-2730` (current query), `src/session/sync.rs:89-263` (cross-claim guards that currently mask the mis-pick).
- Study: herdrdev/herdr `src/detect/mod.rs:310-314` (identity-only treatment upstream).

jerome-benoit:
## Description

Fixes #3373. `hermes_poll_fn` read `~/.hermes/state.db` and returned the most recent active CLI conversation regardless of project, so with two hermes sessions on different projects the second session to poll could adopt the first's conversation and resume bound the wrong conversation. Every other capture path validates cwd or project hash (claude encoded-cwd, pi header cwd, gemini projectHash, codex rollout cwd, vibe working_directory); hermes was the exception.

Investigation result: hermes DOES persist a project signal. Its `sessions` table records `cwd` on each local CLI session row (and `git_repo_root` when a gateway/TUI frontend later enriches it), verified against hermes v0.20.1, and hermes' own `-c`/`--resume latest` is workspace-scoped (`_resolve_last_session` + `_workspace_key_clause`, falling back to the global MRU only when the workspace query is empty). The issue's "no cwd column" premise was stale; the proposal's primary branch applies.

## What changed

- `capture_hermes_session_id` scopes the active CLI set to the canonicalized project path: a row matches when its canonicalized `cwd` or `git_repo_root` equals the canonicalized project path. No LIKE/wildcard matching; `canonicalize_or_raw` handles symlinks and deleted-dir `../` spellings (pre-#2858 convention).
- Schema-gated degradation: on legacy databases without `cwd`/`git_repo_root` columns, the sole unclaimed active conversation is returned (unambiguous) and more than one yields `None` (issue story 2: agent starts fresh instead of silently resuming a wrong conversation). With a signal-capable schema, no-signal rows are never returned (a NULL-cwd cli row is foreign by construction; resuming it is the #3373 bug shape).
- Deliberate divergences from `hermes -c` (documented in code): aoe requires exact canonicalized equality, considers active rows only (`ended_at IS NULL`), orders by `started_at` rather than hermes' `last_active` recency, and does NOT fall back to the global most-recent conversation when nothing matches the project; that fallback is the reported bug.
- Container path (`HERMES_CONTAINER_CAPTURE_SCRIPT`): the python one-liner now dumps a `SIGNAL`/`LEGACY` mode line plus TAB-separated `id\tcwd\tgit_repo_root` tuples; all matching, degradation, and exclusion decisions stay in Rust so host-side canonicalization applies on both paths. The previously-unused `container_cwd` parameter is now load-bearing.
- Ordering: `ORDER BY started_at DESC, id DESC` (deterministic tiebreak) and no `LIMIT` (the ambiguity count needs the full active set).

## PR Type

- [x] Bug Fix

## Checklist

- [ ] For UI changes: included screenshot or recording (N/A)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**AI Model/Tool used:** omp harness (deepseek-v4-flash orchestration with multi-agent design/review convergence)

Design and review followed multi-agent convergence loops (design: 7 tours / 21 analyses, two consecutive clean tours; correctness review: 6 tours / 18 analyses, two consecutive clean tours; initial review per the review prompt: 3 independent lenses with tool-source validation against hermes v0.20.1, cross-validated). Tests:

- Unit (62 hermes tests pass): project cwd match, git_repo_root match, monorepo-subdir regression, symlink + `../` spellings, foreign rows never resumed, legacy sole-row rule, full-schema NULL-cwd never resumed, exclusion interplay, container mode-line/TAB parsing (incl. malformed-row residuals), python3-gated literal script test, cwd-only/root-only partial schemas, `hermes_poll_fn` story-1 match and story-2 `None`.
- E2E `hermes_shared_project_correlation` (passes; fails against a reverted fix, teeth verified): two sessions on different projects, seeded fake state.db, TUI poller host, positive per-session `agent_session_id` assertions.
- `cargo fmt` clean; `cargo clippy --features e2e-tests --tests` adds no new warnings; full lib suite 4604 passed / 0 failed; hermes e2e passes.

## Summary

- Scoped Hermes conversation matching to the canonical project path.
- Added deterministic matching by `cwd` and `git_repo_root`.
- Added fail-safe handling for legacy databases and ambiguous sessions.
- Improved container capture with schema mode and session metadata.
- Added regression and end-to-end tests for project isolation, path handling, and container output.

## Benefits

- Prevents conversation IDs from being assigned to the wrong project.
- Preserves existing session data when project signals are unavailable.
- Provides predictable behavior across supported database schemas.
- Confirms correct correlation for concurrent sessions in different projects.

## Potential follow-up

- Monitor legacy database usage in production.
- Add coverage for further schema and path edge cases.


jerome-benoit:
@coderabbitai (CodeRabbit), thanks for the two nits in [your review](https://github.com/agent-of-empires/agent-of-empires/pull/3381#pullrequestreview-4953875661). We cross-checked current HEAD `91357fba`; neither requires a change.

- `docker exec` is invoked without `-t` (`src/session/capture.rs:2949-2956`), and Rust `str::lines()` consumes CRLF terminators (`:2725-2733`), so a CRLF mode line is `SIGNAL`, not `SIGNAL\r`. Trimming would broaden the intentional fail-closed mode protocol.
- The container probe exits before opening an absent store, then already uses default read-write `sqlite3.connect(db, timeout=1.0)`, not `mode=ro` (`:2662-2666`). No code change needed.

Issue #1452, opened 2026-05-23
Mobile right panel: keyboard collapses paired terminal to zero height; convert right panel into a picker that takes over the main view


njbrake:
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Problem

On mobile, the right panel (diff file list + diff viewer + paired terminal) is rendered as an `85vw` slide-in overlay pinned to `bottom-0`. When the soft keyboard opens against the paired terminal inside that overlay, the panel itself does not shrink (it's `position: fixed` anchored to the viewport bottom) but `PairedTerminal` still adds `paddingBottom: keyboardHeight` to push content above the keyboard. The padding eats from a container whose height is already capped, so the terminal collapses to near-zero and the user sees the keyboard sitting over almost nothing.

User proposal: drop the slide-in-sidebar model on mobile entirely. Make the right panel a picker (a list of "Diff", "Paired terminal", maybe "Comments / Notes") that, when an entry is tapped, swaps that view into the main pane in place of the agent session terminal. The left sidebar (sessions / repos) keeps its current drawer behavior; only the right side changes shape. Mobile then has a single full-viewport content area at any moment, which makes the keyboard, scrollbar, and gestures all behave normally.

## Why the current layout breaks under keyboard

Current architecture:

- `web/src/components/ContentSplit.tsx:85-133` mounts the right panel inline on desktop (`md:` flex child with `width: diffWidth`) and as a fixed overlay on mobile (`fixed top-12 bottom-0 right-0 z-40 w-[85vw]`). Backdrop at line 110, panel at line 113.
- Inside the right panel: `web/src/components/RightPanel.tsx:228-406` stacks `DiffFileList` (top), drag handle (middle), `PairedTerminal` (bottom) at a 50/50 ratio controlled by `topRatio` state.
- `PairedTerminal` mount: `RightPanel.tsx:171-176`. Live `keyboardHeight` from `useMobileKeyboard()` (line 85) is applied as `paddingBottom`. Comment at lines 80-84 acknowledges the trap: "applying a ~340px reservation there collapses `data-term=\"paired\"` to 0 height."

So the bug is structural, not a bad height calc:

[...]

PR #3386, opened 2026-08-15
feat(sv): mobile UX fixes, agent inheritance, sandbox sub-agent transcript

njbrake:
## Description

Two tracks on one branch: a self-contained backend feature (server-owned prompt queue) plus a structured-view iteration pass. seluj78 is tagged for the structured-view side.

**Server-owned prompt queue (backend + web):** the ACP follow-up queue moved from client-owned (browser localStorage, client-driven drain) to daemon-owned, so a prompt queued behind a busy turn now survives a closed PWA and delivers server-side. Includes a durable per-session store surfaced on the session response, enqueue / edit / remove / clear HTTP endpoints, drain at turn-end split on `/clear` boundaries, server-side attachments (dedicated table outside the retention prune, per-session byte cap, 24h TTL sweep), and wake-on-drain for idle-stopped workers (respawns without holding the instance lock, so no re-entrant deadlock). The client posts optimistically and reconciles against the server snapshot, with a one-time localStorage migration; the old client-drain machinery is removed.

**Structured view mobile/UX (web):**
- iOS composer hugs the bottom instead of floating above a safe-area gap. The bottom safe-area inset moved off the shared App root onto each bottom-anchored surface (composer, terminal, mobile diff, plugin pane, settings, dashboard).
- Stick-to-bottom reworked: it holds through the keyboard, the hide-input collapse, the composer growing as you type, and streaming, and releases only on a real scroll gesture. Added a centered mobile "jump to latest" button.
- Right-panel picker replaces the slide-in overlay so the paired terminal owns the full viewport and stays tall under the keyboard (#1452).
- Transcript taps no longer open the keyboard; send button greys out when empty; per-row "Send now" on a queued prompt.

**Structured view backend:** sandboxed sub-agent transcripts are read across the container boundary (`docker exec`). Custom agents that map via `agent_detect_as` run in the structured view through the base adapter, with no hand-written `agent_acp_cmd`.

**Auth (behavior change, review finding 13):** `serve.token` now rotates on a 24h *idle* window rather than 24h from creation. The file's mtime is refreshed on every reuse, so a daemon restarted at least once a day keeps the same token indefinitely and only a token untouched for 24h rotates. This is deliberate: a token change forces the rotation-prune path, which silently kills push notifications until every device re-subscribes. Calling it out here because it changes auth-token rotation posture, which a 64-file diff would otherwise hide.

**CLI:** `aoe list` can filter by session state and expose it in JSON.

**Docs:** configuration + structured-view guides; the server-side prompt queue design doc, now implemented.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [x] Documentation

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (`aoe list` state filter; queue drain/store covered by unit + integration tests)

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Opus 4.8 (1M context) via Claude Code

**Any Additional AI Details you'd like to share:** Implemented through back-and-forth with the author, who drove the design decisions and verified behavior on-device.

## Summary

- Improved mobile structured-view layout, safe-area handling, keyboard behavior, and transcript scrolling.
- Added jump-to-latest controls and prevented transcript taps from opening the keyboard.
- Moved queued prompts and attachments from browser storage to durable server storage.
- Added queue APIs, persistence, limits, cleanup, server-side delivery, session wake-up, and client reconciliation.
- Added “Send now” support, including active-turn interruption when required.
- Added sandbox-aware transcript reading through `docker exec`.
- Added ACP adapter inheritance for custom agents through `agent_detect_as`.
- Reset converted terminal sessions to `Idle`.
- Updated push subscription refresh, token rotation, documentation, CI triggers, and test coverage.

## Benefits

- Mobile users get safer layouts and more predictable transcript scrolling.
- Queued prompts survive reloads and device changes.
- Server-side delivery reduces browser coordination and supports dormant sessions.
- Custom agents can reuse supported structured-view adapters.
- Expanded tests protect queue, mobile, sandbox, and session-conversion behavior.

## Technical notes

- Removed client-side queue drainers and drain coordination.
- Added server queue CRUD endpoints and queued prompts to session responses.
- Added durable pending-attachment storage with size limits and TTL cleanup.
- Further improvement could replace transcript polling with `tail -f` after the required tailer-loop and runtime testing work.


src/acp/state.rs
```diff
@@ -595,6 +595,18 @@ impl PromptAttachmentKind {
             PromptAttachmentKind::Resource => "resource",
         }
     }
+
+    /// Parse the lowercase tag written by [`as_str`], for reading the kind
```
Seluj78: Blocker: this breaks the required `Docs` CI check. `[`as_str`]` is an unresolved rustdoc intra-doc link, and `src/lib.rs:4` sets `#![deny(rustdoc::broken_intra_doc_links)]`, so the job ends in `error: could not document agent-of-empires`, exit 101. Use ``[`Self::as_str`]`` or escape the brackets.

docs/development/server-side-prompt-queue.md
```diff
@@ -0,0 +1,282 @@
...
+- Client rewired to the server queue (`useAcpSession`): `sendPrompt`'s enqueue
+  branch POSTs to `/queue` with an optimistic `pending` row and confirms it;
+  edit/remove/clear mirror to the endpoints; the queue reconciles against the
+  server snapshot via `hydrate_server_queue` (on connect and turn-end), keeping
+  local attachment thumbnails and any in-flight enqueue. The old client drain,
+  `AcpQueueDrainer`/`AcpBackgroundDrainers`, and `acpDrainCoordinator` are
+  removed (the daemon drains now). A one-time migration on first connect pushes
+  any localStorage-restored rows to the server.
+
+Still open: the `send-now` endpoint (G3) — today "Send now" sends immediately
```
Seluj78: Blocker: em dash. `AGENTS.md` Coding Style: "No emdashes or `--` as separators in docs/comments; use commas, semicolons, or rephrase."

web/src/hooks/useAcpSession.ts
```diff
@@ -1812,144 +1826,120 @@ export function useAcpSession(
...
+      const steerable = !!state.promptCapabilities?.steering && !state.cancelling && !state.compacting;
+      if (state.turnActive && !steerable) {
+        await cancelPromptRef.current();
+        return;
+      }
+      // Remove server-side first so the turn-end drain can't also deliver this
+      // row, then send it directly. The optimistic dequeue hides it locally.
+      dispatch({ kind: "dequeue_prompt", id: prompt.id });
+      await removeServerQueuedPrompt(sid, prompt.id);
+      const result = await dispatchPromptNow(prompt.text, prompt.attachments);
```
Seluj78, line 1957: Blocker: this permanently destroys a queued prompt with attachments after any reload.

Repro: attach an image while the agent is busy (`Composer.tsx:533` allows an attachment-only send) so it queues server-side, then reload the page. `toPersistedState` (`useAcpSession.ts:182-185`) strips queued rows carrying attachments from localStorage, so `hydrate_server_queue` rebuilds the row from the server refs with `dataB64: ""` (`useAcpSession.ts:714-719`). Tap Send now: line 1926 deletes the server row and its buffered bytes (`session_service.rs:886-889`), then this line POSTs `data: ""` (`useAcpSession.ts:1637`), which `validate_attachments` rejects 400 on `bytes.is_empty()` (`src/server/api/acp.rs:176-178`). A 4xx is not `retryable_failure`, so the re-queue branch below never runs.

Net: prompt, caption, and image gone, nothing sent. Either refuse Send now for a row whose bytes are server-side only, or add a server-side send-now endpoint that reuses the buffered blobs (the G3 the design doc lists as still open).

src/server/session_service.rs
```diff
@@ -732,6 +766,300 @@ impl SessionService {
...
+                    .iter()
+                    .flat_map(|pid| store.load_pending_attachments_for_ref(&session_id, pid))
+                    .collect::<Vec<_>>()
+            })
+            .await
+            .unwrap_or_default()
+        };
+        // An attachment-only batch has empty combined text but real blobs; only
+        // a batch with neither is a no-op.
+        if combined.trim().is_empty() && attachments.is_empty() {
```
Seluj78, line 1109: Blocker (first half; see also `acp_reconciler.rs:837`): the hourly attachment sweep can strand a queued prompt here forever.

An attachment-only prompt has `text == ""`. If it sits past `PENDING_ATTACHMENT_TTL` (a session that stays busy or dormant for 24h), `prune_pending_attachments_older_than` (`src/server/mod.rs:1334`) reclaims the bytes but nothing removes the queue row. This guard then computes `combined == ""` with zero blobs and returns **without retiring the row**, on every tick, forever. A `/clear` row behind it can never drain either, since the batch split stops at it.

The sweep and this guard need to agree: either the sweep drops rows whose blobs it reclaimed, or this branch retires the batch instead of returning.

src/server/acp_reconciler.rs
```diff
@@ -825,6 +829,12 @@ async fn reap_idle_workers(state: &Arc<AppState>) {
                     && !i.is_snoozed()
                     && !i.is_trashed()
                     && !i.is_idle_dormant()
+                    // A session with queued work waiting to drain is not idle:
+                    // reaping it here would fight wake-on-drain, which clears
+                    // dormancy to respawn exactly these sessions. Leave it alive
+                    // until its queue drains; the next idle window (empty queue)
+                    // reaps it normally. See `drain_queued_prompts`.
+                    && i.queued_prompts.is_empty()
```
Seluj78, line 837: Blocker (second half of the `session_service.rs:1013` finding): this reap guard turns an undrainable queue row into a permanently running agent worker.

Any session with a non-empty queue is now never idle-reaped. Combined with the drain's empty-batch early return, one stranded row (TTL-swept attachment-only prompt, or a `PATCH` to empty text) keeps an agent subprocess alive indefinitely with no way for the user to notice why. A liveness bound here (drop the guard once the queue has failed to drain N times, or bound how long the guard defers a reap) would stop a stuck queue from leaking a worker.

src/server/api/queue.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use super::acp::{read_only_block, validate_attachments};
+use crate::acp::protocol::PromptAttachmentUpload;
+use crate::acp::state::PromptAttachmentRef;
+use crate::server::AppState;
+
+/// Cap on total queued-attachment bytes buffered per session. Generous enough
+/// for several image follow-ups (`/acp/prompt` caps one prompt at 20 MiB) while
+/// bounding what an undrained queue can hold on disk.
+const MAX_QUEUED_ATTACHMENT_BYTES_PER_SESSION: u64 = 64 * 1024 * 1024;
```
Seluj78, line 33: Required: attachment bytes are capped per session, but queue depth and per-prompt text are not. `enqueue_prompt` unconditionally pushes (`session_service.rs:812-830`) and every mutation rewrites the whole profile session file through `Storage::update`, so an authenticated client can grow one session's on-disk record without limit and pay a full rewrite per enqueue. That this path was bounded and the other two were not reads as an oversight rather than a decision.

src/server/api/queue.rs
```diff
@@ -0,0 +1,227 @@
...
+    if let Some(resp) = read_only_block(&state) {
+        return resp;
+    }
+    let Json(req) = match req {
+        Ok(j) => j,
+        Err(rej) => return rej.into_response(),
+    };
+    if state
+        .session_service
+        .edit_queued_prompt(&id, prompt_id, req.text)
```
Seluj78, line 239: Required: the edit path is missing the emptiness validation the enqueue path enforces. Line 80 rejects `text.trim().is_empty() && attachments.is_empty()`; this accepts anything, so a `{"text": ""}` PATCH creates exactly the undrainable, worker-pinning row described at `session_service.rs:1013`. The UI happens to guard it (`StructuredView.tsx:2736`), so this is the trust boundary missing its check, not a UI bug.

src/server/session_service.rs
```diff
@@ -732,6 +766,300 @@ impl SessionService {
         }
     }
 
+    /// Apply `mutate` to a session's in-memory `Instance` and persist the same
+    /// change to disk, mirroring the `pending_initial_turn` write path. Returns
+    /// what `mutate` produced on the in-memory instance, or `None` if the
+    /// session is gone. `mutate` runs twice, once in memory and once against the
```
Seluj78: Required: this lockstep claim does not hold under concurrent mutation. The `instances.write()` guard is dropped at the end of the first block, before `spawn_blocking` does the disk write. Two enqueues (two devices, or a double tap) can interleave so memory records A=seq0/B=seq1 while the disk writes land B first, giving B=seq0/A=seq1 on disk. After a daemon restart the queue drains in the opposite order from what the user saw. Either hold the lock across the persist, or assign `seq` once and pass it into the on-disk run instead of re-deriving it.

src/acp/background_agent.rs
```diff
@@ -52,6 +52,92 @@ const WAIT_FILE_FOR: Duration = Duration::from_secs(30);
...
+    async fn exists(&self, path: &str) -> bool {
+        match self {
+            TranscriptSource::Host => tokio::fs::metadata(path).await.is_ok(),
+            TranscriptSource::Container { runtime, container } => {
+                tokio::process::Command::new(runtime)
+                    .args(["exec", container, "test", "-e", path])
+                    .stdin(std::process::Stdio::null())
+                    .stdout(std::process::Stdio::null())
+                    .stderr(std::process::Stdio::null())
+                    .status()
```
Seluj78: Required: this and the `.output()` at line 130 are unbounded and not `kill_on_drop`. Both are awaited **outside** the `tokio::select!` at lines 271 and 354, so a wedged `docker`/`podman` (or a container stuck in `D` state) blocks the await; the tailer never observes `event_tx.closed()` and never reaches `ABORT_AFTER`, and `ActiveGuard` stays held, so the between-prompt idle watchdog reports work in flight for that session forever. `tokio::fs` on a local file did not have this failure mode; the container path does. Wrap both in `tokio::time::timeout` and set `kill_on_drop(true)`. Corroborates the unresolved CodeRabbit thread on this file.

src/acp/background_agent.rs
```diff
@@ -52,6 +52,92 @@ const WAIT_FILE_FOR: Duration = Duration::from_secs(30);
...
+                // `tail -c +N` prints bytes from the 1-based byte offset N to
+                // EOF, so a 0-based `offset` maps to `+(offset + 1)`. On the
+                // first read (offset 0) this is `+1`, i.e. the whole file.
+                let start = format!("+{}", offset.saturating_add(1));
+                let output = tokio::process::Command::new(runtime)
+                    .args(["exec", container, "tail", "-c", &start, path])
+                    .stdin(std::process::Stdio::null())
+                    .stderr(std::process::Stdio::null())
+                    .output()
+                    .await;
```
Seluj78: Required: `POLL_INTERVAL` is 500ms and the loop runs up to `ABORT_AFTER` (300s), so each active sandboxed sub-agent costs 2 `docker exec` spawns per second: a docker CLI fork+exec, a daemon socket round trip, and a new in-container process each time. A turn with several parallel `Task` sub-agents multiplies that. One long-lived `docker exec ... tail -c +1 -f` streaming into the existing line folder removes the polling entirely and fixes the unbounded-exec problem at line 94 in the same change.

web/src/hooks/useAcpSession.ts
```diff
@@ -1812,144 +1826,120 @@ export function useAcpSession(
...
+  //     live prompt path, and remove it from the server queue so the server
+  //     drain doesn't also deliver it (dedup). The optimistic dequeue keeps the
+  //     strip in sync; a hydrate reconciles.
+  //   - A live, non-steerable turn is blocking it: interrupt. Cancel the
+  //     current turn; the server then drains the queue (this row included) on
+  //     turn-end. We deliberately do NOT POST during the cancel: the daemon
+  //     treats a prompt arriving mid-cancel as a wedge and restarts the runner
+  //     (see `sendPrompt`). This is the destructive "interrupt and send" the
+  //     user opted into.
+  const sendQueuedNow = useCallback(
```
Seluj78, line 1919: Required: no test coverage. `grep -rn sendQueuedNow web/` matches only `StructuredView.tsx`, `AcpRuntime.tsx`, and this hook. `QueuedPromptsStrip.test.tsx` covers the button (wiring, disabled state, aria-label) but nothing here: the remove-then-send ordering, the attachment path, or the `retryable_failure` re-queue. This untested function is where the data-loss Blocker at line 1927 lives.

web/src/hooks/useAcpSession.ts
```diff
@@ -1812,144 +1826,120 @@ export function useAcpSession(
...
+      if (!sid) return;
+      const steerable = !!state.promptCapabilities?.steering && !state.cancelling && !state.compacting;
+      if (state.turnActive && !steerable) {
+        await cancelPromptRef.current();
+        return;
+      }
+      // Remove server-side first so the turn-end drain can't also deliver this
+      // row, then send it directly. The optimistic dequeue hides it locally.
+      dispatch({ kind: "dequeue_prompt", id: prompt.id });
+      await removeServerQueuedPrompt(sid, prompt.id);
```
Seluj78: Required: this can double-deliver the prompt. `drain_queued_prompts_once` holds `instance_lock` across `send_turn` (`session_service.rs:958-959`), but `remove_queued_prompt` goes through `mutate_instance_persisted`, which takes only `instances.write()`. A Send now tap landing inside the drain's `send_turn` window removes a row the drain has already snapshotted and is mid-delivery on; the drain's `retain` then no-ops and `dispatchPromptNow` sends the same text again. Taking `instance_lock` for the removal closes it.

src/server/mod.rs
```diff
@@ -2893,32 +2928,51 @@ fn is_valid_token_format(token: &str) -> bool {
...
                 .unwrap_or_default();
-            if age < std::time::Duration::from_secs(24 * 60 * 60) {
+            if age < max_age {
                 if let Ok(token) = tokio::fs::read_to_string(&token_path).await {
                     let token = token.trim().to_string();
                     if !token.is_empty() && is_valid_token_format(&token) {
-                        return Ok(token);
+                        // Refresh the mtime so this reuse resets the idle
+                        // window; the token stays stable across restarts.
+                        write_secret_file(token_path, &token).await;
```
Seluj78, line 3036: Required (PR metadata): this change is not described anywhere in the PR body. Refreshing the mtime on every reuse turns a 24h-from-creation window into a 24h-idle window, so a daemon restarted at least daily keeps the same auth token indefinitely instead of rotating. The doc comment above justifies it well (push subscriptions die on rotation), but a change to auth-token rotation posture belongs in the body of a 64-file PR, not only in a doc comment.

src/server/session_service.rs
```diff
@@ -732,6 +766,300 @@ impl SessionService {
...
+    ///
+    /// Only drains an idle turn: `Status::Idle` means the prior turn emitted its
+    /// terminal `Stopped`, so this starts a fresh turn rather than racing a live
+    /// one. Callers (the reconciler tick) gate on `is_running`, so the worker is
+    /// live and `send_turn` will not spawn, making it safe to hold the instance
+    /// lock across delivery (the #3172 re-entrant-spawn deadlock cannot occur on
+    /// a live worker). The `/clear`-boundary split matches the client
+    /// (`useAcpSession`): a clear-command row fires as its own turn; a leading
+    /// run of non-clear rows combines into one with blank-line separators.
+    /// Attachments are a later increment; text-only rows drain here.
```
Seluj78: Nit: stale comment. Lines 993-1010 load the batch's buffered blobs and line 1021 forwards them to `send_turn`, so attachments are no longer "a later increment".

web/src/components/acp/Composer.tsx
```diff
@@ -173,30 +173,46 @@ export function decideBeforeInputAction(
...
-/** Wrapper class + inline style for the composer's outer <div>. When the
- *  soft keyboard is open we drop the bottom padding and apply a negative
- *  bottom margin equal to the App root's safe-area-inset-bottom so the
- *  composer sits flush with the top of the keyboard instead of leaving a
- *  visible gap (the home-indicator inset is physically occluded by the
- *  keyboard anyway). `accessoryBarPx` adds bottom padding that lifts the footer
- *  above the iOS accessory bar (see IOS_ACCESSORY_BAR_PX); it is 0 off iOS-PWA.
+/** Wrapper class + inline style for the composer's outer <div>.
+ *
+ *  Both keyboard states cancel the App root's `safe-area-inset-bottom` padding
```
Seluj78, line 178: Nit: this paragraph is stale and contradicts the one at line 186. `index.css:648` no longer applies bottom padding on the App root, so there is no reservation to cancel and no negative margin. The comment at line 418 describes the same removed behavior.

src/server/api/queue.rs
```diff
@@ -0,0 +1,227 @@
...
+        .iter()
+        .map(|b| PromptAttachmentRef {
+            id: b.id.clone(),
+            kind: b.kind,
+            mime_type: b.mime_type.clone(),
+            name: b.name.clone(),
+            size: b.data.len() as u64,
+        })
+        .collect();
+    if !req.attachments.is_empty() {
```
Seluj78: Nit: this gate leaks blobs on an attachment-free re-enqueue. Re-posting the same prompt id with no attachments skips the cleanup, so the prior bytes stay buffered and keep consuming the per-session cap until the 24h TTL sweep, with no queue entry referencing them. The cap math above already assumes a re-enqueue replaces the prior blobs. Clearing unconditionally would match.

src/events/mod.rs
```diff
@@ -641,6 +669,163 @@ pub fn load_attachment(
...
+        return false;
+    }
+    true
+}
+
+/// Load every pending attachment buffered under `(topic, ref_id)` as
+/// `(attachment_id, kind, mime_type, name, data)`, in stable insertion order.
+/// Used at drain time to reconstitute a queued prompt's blobs before it is
+/// forwarded to the agent.
+#[allow(clippy::type_complexity)]
```
Seluj78, line 724: Nit: a named struct for the row would read better than `Vec<(String, String, String, Option<String>, Vec<u8>)>` and drops the `allow`.

web/src/components/MobileMainPane.tsx
```diff
@@ -125,23 +125,50 @@ export function MobileMainPane({
...
-            />
+            // Reserve the bottom home-indicator inset on this wrapper (the App
+            // root no longer does; see index.css .safe-area-inset) so the last
+            // terminal row clears it. Kept off the pane root itself, which owns
+            // the keyboard-open lift and must stay inset-free when closed (#1432).
+            // Collapses to 0 with the keyboard open (iOS reports the inset as 0)
+            // and on desktop.
+            <div
+              className="flex-1 flex flex-col min-h-0 overflow-hidden"
+              style={{ paddingBottom: "env(safe-area-inset-bottom)" }}
```
Seluj78, line 136: Nit: only 2 of the 6 surfaces that took over the bottom safe-area reservation have a guard (`mobile-paired-layer`, `mobile-plugin-layer`). This agent-terminal wrapper, the mobile diff at line 170, `SettingsView.tsx:916`, and `Dashboard.tsx:68` have none. This class of miss already happened twice inside this PR: commit da7ef55ba fixes two surfaces the earlier commit missed.

Seluj78 (dismissed):

## AoE Review

Reviewed: PR #3386 (base: `main`@`0e8d263bd`). Files changed: 64, +3881/-2910 (server, acp, events, cli, session, web, web-tests, docs-user).

### Summary

The server-owned queue solves a real gap (a follow-up queued behind a busy turn dying with the tab), the mobile keyboard and safe-area work is grounded in #1432/#1452, and the sandbox sub-agent transcript fix repairs a genuine bug. The design doc and comment density are unusually good.

The defects cluster in the queue's failure paths: two independent ways to create a row that can never drain and now also pins an agent worker forever, and a Send now path that silently destroys a queued image after a reload. Both live in code with no test coverage. Two required CI checks are also red.

Every finding below is anchored inline on its cited line.

### Blocker

1. **CI check `Docs` fails.** Broken rustdoc intra-doc link at `src/acp/state.rs:599` under `#![deny(rustdoc::broken_intra_doc_links)]`; `error: could not document agent-of-empires`, exit 101.
2. **CI check `Lint PR title` fails.** Title has no conventional-commit prefix: `No release type found in pull request title`. `AGENTS.md` requires `feat:` / `fix:` / etc.
3. **Em dash in a new docs page** (`docs/development/server-side-prompt-queue.md:37`). `AGENTS.md` bans em dashes in docs and comments.
4. **`sendQueuedNow` permanently destroys a queued prompt with attachments after any reload.** `toPersistedState` strips attachment-carrying queued rows from localStorage, so a hydrated row's `dataB64` is `""`; Send now deletes the server row and its buffered bytes, then POSTs empty base64, which `validate_attachments` rejects 400. A 4xx is not `retryable_failure`, so nothing re-queues. Prompt, caption, and image are gone with nothing sent.
5. **The 24h attachment sweep strands an attachment-only queued prompt, which then pins the worker forever.** The sweep reclaims the bytes but nothing removes the row; the drain's empty-batch guard returns without retiring it, every tick; and `reap_idle_workers` now skips any session with a non-empty queue, so the agent subprocess is never reaped.

### Required

6. **The server queue has no depth or text bound.** Attachment bytes are capped per session, text and queue depth are not, and every mutation rewrites the whole profile session file.
7. **`PATCH /queue/{promptId}` skips the emptiness validation `POST /queue` enforces**, producing the undrainable row from finding 5 through the API.
8. **`mutate_instance_persisted`'s lockstep claim does not hold under concurrent mutation.** The instances lock is released before the disk write, so two concurrent enqueues can persist a different `seq` order than memory holds.
9. **Container transcript reads are unbounded and not `kill_on_drop`.** A wedged `docker`/`podman` blocks the tailer before it can observe `event_tx.closed()` or its idle timeout, holding `ActiveGuard` indefinitely. Corroborates the unresolved CodeRabbit thread on this file.
10. **500ms `docker exec` polling per sandboxed sub-agent**, up to 300s each. One streaming `tail -f` exec removes the polling and fixes finding 9 at the same time.
11. **No test for `sendQueuedNow`**, the function finding 4 lives in.
12. **`sendQueuedNow` can double-deliver a prompt.** The drain holds `instance_lock` across `send_turn`; `remove_queued_prompt` takes only `instances.write()`, so a tap inside the drain window sends the same text twice.
13. **The `serve.token` lifetime change is not described in the PR body.** Refreshing the mtime on every reuse turns a 24h-from-creation window into a 24h-idle window, so a daemon restarted at least daily keeps the same auth token indefinitely. The code comment justifies it well; a change to auth-token rotation posture belongs in the body of a 64-file PR too.

### Nit

Five inline: stale comments in `session_service.rs` and `Composer.tsx` describing removed behavior, orphaned blobs on an attachment-free re-enqueue, a 5-tuple behind `#[allow(clippy::type_complexity)]`, and safe-area inset guards covering 2 of the 6 surfaces that took over the reservation.

### PR metadata

Body is otherwise complete: PR Type set, checklist engaged, AI Usage filled with model and agent checkbox, screenshot omission justified inline, commits all conventionally prefixed, no em dashes in title or commit messages. Branch name `vietnamese` does not follow the `AGENTS.md` `feature/` / `fix/` convention, noted once since repo practice differs.

### Unresolved comments

8 unresolved CodeRabbit threads, none outdated. The Major one on `src/acp/background_agent.rs` is real and adopted as finding 9. Also worth acting on: `acpScrollState.ts` accepts negative and non-finite `top` values, and the two new send controls do not use the dashboard button tokens.

### Verdict

**Project Fit:** Accept with scope reduction
**Placement:** core (session lifecycle and the web shell, not optional capability)

On scope: the body says "two tracks on one branch," and it is really five (server queue, mobile UX, agent inheritance, sandbox transcript, push/token). At 64 files and 6.8k lines that is not reviewable as one unit, and the undisclosed auth-token change is the kind of thing a diff this size hides. The server-owned queue stands alone cleanly and should be its own PR; the mobile UX pass and the push/token fix are each independently shippable.

njbrake:
@Seluj78 ready for another look. All five blockers and seven of the eight Required findings are fixed; CI is 33/33.

- **1, 2, 3**: rustdoc link, PR title, em dash.
- **4**: Send now no longer takes the destructive path for a row whose bytes live only server-side. It leaves the row alone and the turn-end drain delivers it.
- **5**: the drain retires an undeliverable batch instead of returning. Writing the test narrowed the repro: the wedge needs the husk to be the whole batch, since a deliverable row alongside it supplies the text and the batch sends normally.
- **6**: 100-row depth cap and 256 KiB text cap, on enqueue and edit.
- **7**: edit rejects an empty result, as enqueue does.
- **8**: the closure now runs once against memory and the disk write copies the result, so concurrent enqueues cannot persist an order memory never had.
- **9**: both container reads bounded by a 10s timeout and `kill_on_drop`.
- **11**: three `sendQueuedNow` tests, mutation checked (removing either new guard fails them).
- **12**: remove takes the per-instance lock the drain holds, and the client reads a 404 as "the drain claimed it" and sends nothing.
- **13**: the `serve.token` idle-window change is described in the body.

The five nits and the CodeRabbit majors (`session_spawn` resolved-key downgrade, per-hook `queueMigratedRef`, orphaned blobs, `loadScrollState` validation) are in as well.

**Not done: 10.** Swapping the 500ms poll for a streaming `tail -f` restructures the tailer loop, and there is no container runtime in my environment to exercise it. Finding 9's timeout removes the hang risk; the polling cost stands. Happy to take it as a follow-up.

Your scope-split point is with @njbrake.

PR #3408, opened 2026-08-17
fix(tui): confirm the quick-delete shortcut instead of trashing on one keystroke

### LINKED ISSUE #3364: Suggestion: Make the Command for Deleting an Agent/Group More Complex
I have deleted my agents several times because I switch between agents, forget to enter live mode, start typing, which often includes "d" and my agent deletes.

Making the command to delete an agent or group slightly more complex will help avoid this problem in the future.

klubrake:
## Description

Fixes #3364.

Switching between agents without entering live mode makes a typed `d` trash the selected session outright. It is easy to do by accident and it costs real work.

This turns `session.confirm_delete` **on by default**, so `d` opens the existing confirmation dialog, and lets the delete key itself accept that dialog. The deliberate gesture becomes two taps of one key; a single stray keystroke is now harmless. `Esc` cancels and returns to the session list.

The dialog as it actually renders (captured from `ConfirmDialog::render` against a `TestBackend`, not a mockup):

```
╭ Confirm Delete ────────────────────────────────╮
│                                                │
│ Move 'refactor-auth' to the trash? It can be   │
│ restored from the Trash section.               │
│ Press d again to confirm, Esc to cancel.       │
│                  [Yes]    [No]                 │
│                                                │
│                                                │
╰────────────────────────────────────────────────╯
```

Under `strict_hotkeys` the prompt and the accept key are both `D`, matching the binding.

The buttons sitting flush under a multi-row message is the dialog's existing layout for any wrapped body (the switch-view confirm renders the same way); I left it alone rather than widen this PR into a layout change.

### Notes for reviewers

- **The setting already existed.** `session.confirm_delete` shipped in #2595; it just defaulted off. This PR flips the default rather than adding a second knob, so there is nothing new to configure. Users who prefer the one-keystroke trash set `session.confirm_delete = false`.
- **The confirm key is opt-in per dialog** (`ConfirmDialog::confirmed_by`), not a blanket `d` binding. `ConfirmDialog` is shared with the quit, force-remove, stop-session and switch-view prompts, and binding `d` on all of them would recreate the exact bug this fixes. Cancel keys keep priority over the opt-in key.
- **Groups were never affected.** They already route through `GroupDeleteOptionsDialog` or a confirm dialog, so the session trash path was the only instant-delete hole. That is why the diff is this narrow.
- **Breaking behavior change**, called out per `AGENTS.md`: `d` on a session now asks before trashing.
- Four assertions in `settings/mod.rs` and `config.rs` used `confirm_delete = true` as a stand-in for "a non-default value a peer process wrote". With the default flipped they would have passed tautologically, so they now write `false`.
- Docs were not touched: the keymap tables in `docs/quick-start.md` and `docs/guides/workflow.md` say "Delete session", which is still accurate.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

The web delete flow already confirms and is unchanged. `confirm_delete` is surfaced in the CityHall settings pane, but the schema resolves defaults dynamically via `serde_json::to_value(Config::default())`, so the toggle picks up the new default with no web-side change.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: five unit tests cover this at a tier that can actually fail if the behavior breaks, including the on-disk `trashed_at` write. Per the "What a test costs" guidance, a Rust e2e is strictly serial and adds to the critical path forever; the only thing it would add here is proof that keys route to an open `ConfirmDialog`, which `tests/e2e/force_remove_tmux_teardown_e2e.rs` already exercises on the same code path.

Tests added or changed:

| Test | Covers |
| --- | --- |
| `d_then_d_confirms_the_trash_and_persists_the_marker` | the new default gesture, end to end through storage |
| `d_with_confirm_delete_prompts_before_trashing` | first `d` opens the dialog and does not trash |
| `d_with_confirm_delete_off_trashes_on_the_keystroke` | the opt-out still trashes immediately |
| `confirm_delete_dialog_cancel_leaves_session` | `Esc` cancels and clears the pending target |
| `confirmed_by_accepts_the_opening_hotkey` | either case confirms; cancel keys win; inert without the opt-in |

### Verification

Run locally on Rust 1.97.1:

| Check | Result |
| --- | --- |
| `cargo clippy` | no new warnings |
| `cargo test --lib` | 4587 passed, 0 failed |
| `cargo test` (integration) | all targets pass |
| `cargo test --features serve --lib` | 6233 passed, 2 failed |

The 2 failures are environmental, not from this diff: `pid_source_for_falls_back_to_peer_pid_on_load_err` and `shutdown_and_wait_degrades_when_load_errs_without_peer` `chmod 0o000` a file and assert the read fails, but my container runs as root (uid 0), which bypasses permission checks, so the fixture's own precondition fails before the test body runs. Both fail identically on a clean tree.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Used to locate the existing delete/confirm plumbing, write the patch and tests, and run the checks above. I reviewed the result and will respond to review comments myself.

## Summary

- Session deletion now requires confirmation by default.
- Press `d` to open confirmation, press `d` again to delete, or press `Esc` to cancel.
- Users can restore immediate deletion with `session.confirm_delete = false`.
- The confirmation dialog now supports an optional confirmation key.
- Tests cover confirmation, cancellation, opt-out behavior, key handling, and configuration persistence.

## Benefits

This reduces accidental session deletion while preserving a deliberate deletion workflow. Group deletion and other confirmation dialogs remain unchanged.

## Technical Notes

The trash flow persists the session’s trash marker only after confirmed deletion. Configuration updates also preserve external changes to `session.confirm_delete`.

### DIFF
--- a/src/migrations/mod.rs
+++ b/src/migrations/mod.rs
@@ -32,13 +32,14 @@ mod v021_split_app_state_to_state_toml;
 mod v022_prune_tuning_settings;
 mod v023_clear_structured_container_error;
 mod v024_backfill_detect_as;
+mod v025_reenable_confirm_delete;
 
 use anyhow::Result;
 use std::fs;
 use std::path::PathBuf;
 use tracing::{debug, info};
 
-const CURRENT_VERSION: u32 = 24;
+const CURRENT_VERSION: u32 = 25;
 const VERSION_FILE: &str = ".schema_version";
 
 struct Migration {
@@ -167,6 +168,11 @@ const MIGRATIONS: &[Migration] = &[
         version: 24,
         name: "backfill_detect_as",
         run: v024_backfill_detect_as::run,
+    },
+    Migration {
+        version: 25,
+        name: "reenable_confirm_delete",
+        run: v025_reenable_confirm_delete::run,
     },
 ];
 
--- a/src/migrations/v025_reenable_confirm_delete.rs
+++ b/src/migrations/v025_reenable_confirm_delete.rs
@@ -1,0 +1,159 @@
+//! Migration v025: re-enable the delete confirmation on existing installs.
+//!
+//! `session.confirm_delete` shipped defaulting off (#2595), and `update_config`
+//! re-serializes the whole `Config` on every write, so any install that has
+//! ever saved a setting carries an explicit `confirm_delete = false` in its
+//! `config.toml`. Flipping the compiled default to on (#3364) therefore reaches
+//! nobody but fresh installs: the persisted `false` outranks the serde default
+//! forever, and `d` keeps trashing on one keystroke.
+//!
+//! This rewrites a persisted `false` to `true` in the global `config.toml` so
+//! the guard actually lands. A `false` on disk is indistinguishable from a
+//! deliberate opt-out, but since the old default was off, nearly every one of
+//! them is a serialized default rather than a choice; the rare deliberate
+//! opt-out is one settings toggle (or one tick of the dialog's "don't warn me
+//! again" box) away from being restored.
+//!
+//! Profile configs are deliberately untouched. They are sparse, holding only
+//! the keys a user actually overrode, so a `confirm_delete = false` there is a
+//! real decision rather than a serialized default. Repo configs never carry
+//! this setting at all.
+
+use anyhow::Result;
+use std::fs;
+use std::path::Path;
+use tracing::{debug, info};
+
+pub fn run() -> Result<()> {
+    let app_dir = crate::session::get_app_dir()?;
+    run_in(&app_dir.join("config.toml"))
+}
+
+/// Inner body so the test can drive the migration end-to-end against a temp
+/// file instead of inlining a near-copy of the production logic.
+pub(crate) fn run_in(path: &Path) -> Result<()> {
+    if !path.exists() {
+        debug!("No global config.toml, nothing to re-enable for v025");
+        return Ok(());
+    }
+
+    let content = fs::read_to_string(path)?;
+    // A config that does not parse is skipped rather than aborting startup:
+    // this migration is a default correction, not a load-bearing relocation.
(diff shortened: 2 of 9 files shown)


njbrake (approved):
Approving. Thanks for this, and for the writeup; the reasoning about why the confirm key is opt-in per dialog rather than a blanket `d` binding is right, and it saved the quit and force-remove prompts from inheriting the bug.

One thing the PR as posted could not do, which I've pushed four commits to fix: the default flip reached almost nobody. `update_config` re-serializes the whole `Config` on every write, so any install that has ever saved a setting already carries an explicit `confirm_delete = false`, and that beats the serde default forever. I hit it on my own machine, where `d` still trashed on the first press. Migration v025 rewrites a persisted `false` to `true` in the global config; profile configs are sparse, so a `false` there is a real choice and is left alone.

The other three: the accept key now comes from `bindings::label(ActionId::Delete, ...)` instead of a hardcoded `d` / `D`; the dialog offers the same "don't warn me again" checkbox the quit confirm has, so the opt-out doesn't require finding a setting; and the buttons read `[Delete]` `[Cancel]` instead of `[Yes]` `[No]`. That last one turned up a pre-existing bug in the shared button row: the click hit-rects sat one cell left of the drawn glyphs, because ratatui centers with `width / 2 - line_width / 2` rather than `(width - line_width) / 2`. Every confirm dialog was affected. Fixed with a test that reads the glyphs under each rect.

Worth knowing, not blocking: `d`-then-`d` narrows the accident but doesn't close it. Typing `middle`, `hidden` or `did` into the sidebar still deletes, since a later `d` (or `y`) confirms with no adjacency or timing requirement. I measured it against `main`, where a single `d` was enough, so this is a real improvement; tightening the gesture can be a follow-up.

PR #3409, opened 2026-08-17
chore(deps): keep pre-1.0 web packages out of the grouped minor bumps

njbrake:
## Description

Dependabot classifies `0.14 -> 0.15` as a plain `minor`, so a 0.x package whose minors are breaking by convention lands inside the `web-minor-patch` group and takes the whole group down with it. In #3394 that was `@assistant-ui/react`: its 0.15 bump dropped `useComposerRuntime`, `useThreadRuntime`, and `useMessage`, failed `npm run build`, and blocked 15 unrelated updates behind 11 red jobs.

Excluding the 0.x packages from the group gives each its own PR, so one breaking library cannot hold the routine bumps hostage.

Note that grouping conditions are pattern plus `update-types` only, with no version-range predicate, so an explicit exclusion list is the only way to express this. It needs a line when a new 0.x dep is added or an existing one reaches 1.0. `/web` is the only npm directory with 0.x deps today; `website`, `acp-worker/aoe-agent`, and `acp-worker/test-shim` are all 1.0+.

`oxfmt` and `eslint-plugin-react-refresh` are excluded on the same reasoning: an oxfmt minor can change formatting output and fail the repo-wide `format:check` gate.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] N/A: this PR does not change TUI rendering, a CLI subcommand, or session lifecycle

Dependabot config is not executable by the test suite, so there is nothing to assert beyond the file parsing. I verified it parses and that the group resolves to the intended exclusion list.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 via the Claude Code CLI.

**Any Additional AI Details you would like to share:**

Written while triaging #3394, where the grouping problem surfaced. Reasoning and decisions are @njbrake's; the prose is the model's.

## Summary

- Updated Dependabot to exclude `@assistant-ui/*`, `eslint-plugin-react-refresh`, and `oxfmt` from grouped web dependency updates.
- Dependabot will create separate pull requests for these packages.
- This prevents breaking 0.x updates and repository-wide formatting or check changes from blocking unrelated updates.

## Benefit

Dependency updates remain smaller, more focused, and easier to review.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -97,6 +97,15 @@ updates:
           - "react-dom"
           - "tailwindcss"
           - "@tailwindcss/*"
+          # Pre-1.0 packages: a minor bump is a breaking change by semver
+          # convention, but dependabot still classifies it as "minor" and so
+          # would group it with the routine updates. Excluding them gives each
+          # its own PR, so one breaking library cannot hold the rest hostage.
+          # In #3394, @assistant-ui/react 0.14 to 0.15 dropped three hooks and
+          # blocked 15 unrelated bumps behind a red build.
+          - "@assistant-ui/*"
+          - "eslint-plugin-react-refresh"
+          - "oxfmt"
         update-types: ["minor", "patch"]
     ignore:
       - dependency-name: "react"


.github/dependabot.yml
```diff
@@ -97,6 +97,15 @@ updates:
           - "react-dom"
           - "tailwindcss"
           - "@tailwindcss/*"
+          # Pre-1.0 packages: a minor bump is a breaking change by semver
```
jerome-benoit, line 100: Non-blocking: SemVer 2.0 says a `0.y.z` API is unstable, not that every minor bump is necessarily breaking. Suggested wording: `Pre-1.0 packages may introduce breaking changes in minor releases`. The policy remains valid.

jerome-benoit (approved):
Configuration correct: Dependabot has no dynamic current-version `<1.0.0` group predicate, and these exclusions produce individual PRs without affecting the stable catch-all. All four current direct pre-1.0 `/web` dependencies are covered; checks are green.

Two non-blocking cleanups: narrow or explicitly document the scope-wide `@assistant-ui/*` policy (existing thread), and change the PR-body claim to `/web is the only grouped npm directory with 0.x dependencies` because the Claude and Pi adapter directories also pin 0.x packages.

Issue #3399, opened 2026-08-17
Transcript probe ignores profile CLAUDE_CONFIG_DIR, breaking --resume


tomdcc:
## Problem

The transcript-existence probe in `session.store` resolves conversation files against the
**default** `~/.claude/projects/<encoded-cwd>/`, ignoring the per-profile `CLAUDE_CONFIG_DIR`
that AoE itself injects into the spawned pane.

For any profile pinned to a non-default config dir, the probe therefore always concludes the
stored sid has no transcript, and downgrades the launch from `--resume <sid>` to
`--session-id <sid>`:

```
INFO session.store: stored Claude sid has no transcript on disk; launching fresh with
--session-id instead of --resume to avoid a certain resume failure sid=5fed8d3a-…
```

The transcript exists — under the profile's own config dir. The spawned pane *does* get the
right `CLAUDE_CONFIG_DIR`, so `claude` finds the conversation, sees the id already taken, and
exits immediately:

```
Error: Session ID 5fed8d3a-c6b3-402e-a760-5ffd5c56549c is already in use.
Pane is dead (status 1)
```

With `remain-on-exit on` the pane lingers as a corpse, so `aoe session start` afterwards
reports `✓ Started` but silently no-ops (the tmux name is taken) — the session appears to
simply refuse to restart, with no surfaced error.

The guard intended to *avoid* a certain resume failure is what causes one. It fires on every
session in the profile: 13 of 15 sessions in my `personal` profile were unstartable.

## Context

Discovered on **aoe 1.13.1** (macOS 15.6, tmux 3.7b). Not fixed in 1.14.1 — nothing in the
1.14.0/1.14.1 notes touches the probe.

Profile config, `~/.agent-of-empires/profiles/personal/config.toml`:

```toml
environment = [
  "CLAUDE_CONFIG_DIR=/Users/tom/.claude-personal",
  "AGENT_OF_EMPIRES_PROFILE=personal",
]
```

The env var reaches the pane correctly — from `pane_start_command`:

```
exec env CLAUDE_CONFIG_DIR='/Users/tom/.claude-personal' AGENT_OF_EMPIRES_PROFILE='personal' \
  … claude --remote-control reindex-validate-partitioned \
  --session-id 5fed8d3a-c6b3-402e-a760-5ffd5c56549c
```

So this is purely the **read** path: the launch path honours the profile env, the probe does
not. Both should resolve the projects root the same way.

[...]

PR #3410, opened 2026-08-17
fix(session): resolve Claude's config dir per profile when reading its state

njbrake:
## Description

The launch path injects a profile's `CLAUDE_CONFIG_DIR` into the pane, but every host-side read of Claude's on-disk state resolved AoE's own environment instead, so for any profile pinned to a non-default config dir the two disagreed.

The read that broke is #2700's empty-thread guard. It checks whether `<config>/projects/<encoded-cwd>/<sid>.jsonl` exists and, when confirmed missing, launches `--session-id <sid>` instead of a doomed `--resume <sid>`. Looking in `~/.claude` for a profile whose transcripts live elsewhere made that check report every real conversation absent, so every launch was downgraded. The pane, which does have the right config dir, then found the conversation, saw the id was taken, and exited:

```
Error: Session ID 5fed8d3a-... is already in use.
Pane is dead (status 1)
```

The reporter had 13 of 15 sessions in the profile unstartable. Their configuration is correct: #1042 removed the `[claude] config_dir` field and pointed users at the environment variable, and `docs/guides/configuration.md` calls the per-session config-dir switch the canonical use case for `before_session` hooks.

Two things made it silent rather than merely broken, both fixed here:

1. A launch that pins an already-stored id without resuming it (the #2700 downgrade, or a fork's pre-generated child id) was never probed, so an instant death went unnoticed.
2. `remain-on-exit` left the dead pane owning the tmux name, so `start` read it as a running session and returned early. `aoe session start` printed `Started` and launched nothing.

### What changed

**Read path.** The transcript probe, the project-dir scan (`capture_claude_session_id`, behind retroactive capture and the live poller), and the `.claude.json` `lastSessionId` fallback all resolve through the config dir the launch injects. That last one moves inside the dir when the variable is set but sits next to the default `~/.claude` when it isn't, verified against a live install.

Resolution uses the full launch environment, not just the profile's static `environment` list: profile entries with `before_session`-minted keys dropped, then the minted pairs, which is the precedence `build_launch_command` applies to the pane. `omp_host_environment` already computed this and is generalized to `resolved_host_environment`. Minted pairs are `#[serde(skip)]` by design, so a read outside a launch degrades to the profile list.

**Failure visibility.** `LaunchSidOutcome::Fresh` now carries `pinned_prior_sid`; those launches get the same settle probe a resume gets. A death returns `Err`, which routes through the existing `fail_reserved_launch` to tear down the corpse and record the pane's own error line on the session. Brand-new sessions skip the probe and pay nothing. Separately, `start` no longer treats a dead pane as a running session.

### Notes for reviewers

- The `Existing`-only probe gate from #2465 is preserved in intent: pinned-fresh launches probe but still return `StartOutcome::Fresh`, never `Resumed`.
- The `Skipped` cache-race guard still holds. A stale-cached `exists()` pairs with a live `is_pane_dead()` query that errors on a killed session and returns false, so it still short-circuits.
- The pinned-fresh error path has no `resume_probe_failed_sid` equivalent, so a repeated manual restart repeats the error. Reusing the loop-breaker would force a fresh sid and silently abandon the conversation, which seemed worse than an honest repeat. Startup recovery cannot spin on it: the per-boot ledger attempts each session at most once.
- Out of scope, worth its own issue: `install_json_host_hooks` / `install_codex_host_hooks` still resolve via the profile list alone, so a minted config dir puts hooks in the wrong `settings.json`. That is a write path, so I left it alone here.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered one tier down, where the assertions are stronger and cost about 1.8s total instead of an e2e slot. `start_with_size_opts_relaunches_a_dead_pane` and `restart_surfaces_a_pinned_fresh_launch_that_dies` in `tests/integration/session_id_acquisition.rs` drive real tmux panes through the actual launch path, which is where both lifecycle changes live. The config-dir resolution is covered by two unit tests. All four were confirmed red against the pre-fix tree.

Four tests, each verified to fail without the fix:

| Test | Tier | Covers |
| --- | --- | --- |
| `claude_reads_resolve_the_profile_scoped_config_dir` | unit | both reads resolve per-profile; same cwd, two config dirs |
| `same_cwd_sessions_resume_their_own_profile_scoped_conversation` | unit | the issue's requested regression test, plus the `before_session`-minted override |
| `start_with_size_opts_relaunches_a_dead_pane` | integration | a corpse pane no longer short-circuits start |
| `restart_surfaces_a_pinned_fresh_launch_that_dies` | integration | the pane's error reaches the caller, session lands in `Error`, corpse torn down |

`cargo fmt`, `cargo clippy --all-targets --features serve,e2e-tests`, `cargo test --features serve --tests`, and the e2e suite were run. Remaining failures are identical to clean `main` in my container (running as root defeats two read-only-dir fixtures, and an ambient `CLAUDE_CONFIG_DIR` breaks a few tests that do not pin it). The e2e failure set is byte-identical to `main`. `cargo xtask gen-docs` produces no diff.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, implementation, and tests were done by Claude Opus 5 in back-and-forth with @njbrake. The first pass resolved only the static profile `environment` list; a review pass caught that it missed `before_session`-minted values, which is the documented pattern for exactly this scenario, and the second commit fixes that. Prior decisions in this area (#1042, #2465, #2612, #3230) were read before changing them, and the notes above record which ones are preserved and where this PR knowingly diverges.

## Summary

- Resolve Claude session state from each profile’s complete launch environment, including values created by `before_session` hooks.
- Use the resolved configuration directory for transcript checks, project scans, polling, and `.claude.json` fallback reads.
- Relaunch dead `remain-on-exit` panes instead of treating them as active sessions.
- Report pane errors when fresh launches fail and remove dead panes after failure.
- Strip ANSI escape sequences from persisted pane error details.
- Share configuration-directory override resolution between session launches and hooks.
- Add unit and integration tests for profile isolation, runtime overrides, dead-pane recovery, and failed launches.

## Benefits

- Prevents profiles with the same working directory from selecting the wrong Claude session state.
- Makes launch failures visible and easier to diagnose.
- Improves recovery from dead panes.
- Keeps configuration-directory precedence consistent across session launches and hooks.

## Further enhancement

- Expand coverage to other launch hooks and environment-dependent session state.

### DIFF
--- a/src/hooks/mod.rs
+++ b/src/hooks/mod.rs
@@ -258,7 +258,7 @@ pub(crate) fn agent_settings_path_display_for_host_environment(
 /// session's host environment list and falling back to AoE's own env so a var
 /// exported in the shell that launched `aoe` (and thus inherited by the agent)
 /// is honored too. Empty values are treated as unset.
-fn resolve_config_dir_override(var: &str, host_env: &[String]) -> Option<String> {
+pub(crate) fn resolve_config_dir_override(var: &str, host_env: &[String]) -> Option<String> {
     crate::session::environment::resolve_host_environment_value(host_env, var)
         .or_else(|| std::env::var(var).ok())
         .filter(|v| !v.is_empty())
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -2038,6 +2038,7 @@ pub async fn acp_enable(
                 !crate::session::capture::claude_host_transcript_confirmed_absent(
                     &instance.project_path,
                     sid,
+                    &instance.profile_host_environment(),
                 )
             })
             .unwrap_or(false);
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -2038,7 +2038,7 @@ pub async fn acp_enable(
                 !crate::session::capture::claude_host_transcript_confirmed_absent(
                     &instance.project_path,
                     sid,
-                    &instance.profile_host_environment(),
+                    &instance.resolved_host_environment(),
                 )
             })
             .unwrap_or(false);
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -61,10 +61,11 @@ fn claude_home_for_host_environment(host_env: &[String]) -> Result<PathBuf> {
 
 /// The `CLAUDE_CONFIG_DIR` value the launched pane will see, if any: the
 /// session's host environment wins, then AoE's own env. Empty is unset.
+///
+/// Shares [`crate::hooks::resolve_config_dir_override`] with the hook-install
+/// path so the read side and the write side cannot drift on precedence.
 fn claude_config_dir_override(host_env: &[String]) -> Option<String> {
-    crate::session::environment::resolve_host_environment_value(host_env, "CLAUDE_CONFIG_DIR")
-        .or_else(|| std::env::var("CLAUDE_CONFIG_DIR").ok())
-        .filter(|value| !value.is_empty())
+    crate::hooks::resolve_config_dir_override("CLAUDE_CONFIG_DIR", host_env)
 }
 
 /// Resolve a path to a comparable identity: canonicalize when the directory
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -39,6 +39,34 @@ fn resolve_agent_home(env_var: Option<&str>, default_subdir: &str) -> Result<Pat
         .join(default_subdir))
 }
 
+/// Resolve the Claude config dir the *launched pane* will see.
+///
+/// The launch path injects the session's profile-scoped `environment` entries
+/// into the pane, so a profile pinning `CLAUDE_CONFIG_DIR` makes the agent read
+/// and write a config tree that is not `~/.claude`. Every host-side read of
+/// Claude's on-disk state has to resolve the same way or it inspects a tree the
+/// agent never touches: the transcript probe then reports a real conversation
+/// absent, and the project-dir scan can hand back a conversation belonging to
+/// another profile that happens to share the cwd. See #3399.
+///
+/// Precedence mirrors [`crate::hooks::agent_settings_path_in`]: the session's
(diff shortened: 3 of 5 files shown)


njbrake (commented):
Reviewed by a separate agent with no hand in writing this, working from the diff and repo alone. Verdict was approve; GitHub blocks a self-approval on my own PR, so this is recorded as a comment.

Verified rather than assumed: the `.claude.json` placement branch was checked against the installed Claude Code bundle source, not just the on-disk layout; all four tests were confirmed red by reverting each production hunk individually; and the `Skipped` cache-race guard still short-circuits, since `is_pane_dead` runs a live query and returns false when the session is gone.

The review found one real defect, now fixed in b07c6881: `capture_pane` captures with `-e`, so a styled agent error line was splicing raw SGR sequences into the persisted `last_error`. The original test missed it because its fake agent printed uncoloured text, so the test carried the same blind spot as the code. It now prints in red and asserts the message is plain text. b99f1228 shares the config-dir override resolver with the hook-install path instead of duplicating it.

Two known gaps, both deliberate and neither blocking:

1. `aoe session start` calls `start_with_size_opts` directly, so it never reaches the probe. A re-death on that path is still a silent success. The fourth acceptance criterion in #3399 is met on the restart paths only.
2. `install_json_host_hooks` and `install_codex_host_hooks` still resolve from the profile list alone, so a `before_session`-minted config dir puts status hooks in a `settings.json` the agent never opens. `pending_host_env` is populated at that call site, so this is a scoping choice about blast radius on a write path, not a constraint. Worth a follow-up issue, along with the sibling agents (`CODEX_HOME`, `VIBE_HOME`, and friends) that still resolve from the process environment alone.

PR #3412, opened 2026-08-17
fix(tmux): read Claude's folder-trust prompt as waiting, not idle

blueagledev:
## Description

Claude's first-run folder-trust prompt read `Idle`, so a session parked on it
never reached the waiting count. `claude_has_approval_prompt` matched only the
two stock openers (`do you want to`, `would you like to proceed`), and the trust
prompt phrases its question without either.

**This has grown past its original one-line shape during review** — the
description below describes what the branch actually does now, not what it did
when it was opened.

### What it does

Two conjuncts, both required:

1. **The question**, matched on the whitespace-collapsed window. It is 46
   characters of prose, so Claude word-wraps it in a narrow pane and
   `recent_lower` is a newline join, which defeats a plain `contains`. AoE
   produces the affected widths itself: side-by-side preview is
   `viewport − list_width − 4`, so the question only fits unwrapped from a
   viewport of 107 up; stacked is `viewport − 4`, fitting from 72. The wrapped
   band is therefore `≤71` and `80..=106`.
2. **The option label**, matched over the choice row *and its wrapped
   continuations*, and required to start the option's own text. The row must
   carry no `>` — `claude_line_is_numbered_choice` strips one, so a markdown
   blockquote quoting this prompt would otherwise open a block.

The label conjunct is not decoration. Collapsing the window for the question
alone let ordinary prose match: a generating turn that renders the question
across a line break, with any numbered line on screen, reported `Waiting` with
a live spinner and `esc to interrupt` visible. That is the opposite of the bug
being fixed, and a blocking rule outranks every running signal, so it sticks.

### Known limit, stated rather than claimed away

A turn that reproduces a real menu row verbatim — quoting it, or echoing it via
`cat`/`less`/a diff context line — still reads `Waiting`. `starts_with` also
admits trailing prose after the label. Both need the question phrase in the
same window. Prefixed echoes (`grep -n`, diff `+`, `⎿`) are rejected, but by
the option-text requirement rather than by any echo filter.

Narrowing further wants a position anchor, not another substring rule. Three
substring attempts during this review were each falsified by the next reader,
which is why the limit is documented in the code rather than papered over.

### Not addressed here

CodeRabbit raised that the two halves of the shared guard are matched
independently anywhere in the window, so an assistant's numbered list plus
`do you want to` can read `Waiting`. That reproduces on `main` today through
the pre-existing arm, so it is out of scope here and filed separately.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (none needed; the detector is
      internal and has no user-facing surface)
- [x] For UI changes: N/A

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: an e2e would need a real Claude parked at a
      first-run trust prompt, which the harness cannot reproduce. The eight unit
      tests this branch adds cover the prompt verbatim, re-wrapped at two
      widths, and five prose/echo shapes that must NOT match.

Three fixtures, honestly labelled: one verbatim `tmux capture-pane -p`, one
re-wrap of it, one abridged narrow variant.

```
$ cargo test --lib claude_folder_trust
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 4602 filtered out
```

## AI Usage

- [ ] No AI was used
- [x] AI was used

Every commit here after the first is a response to review — yours, CodeRabbit's,
or an adversarial pass run over the branch before offering it. That pass has
repeatedly found claims in this description and in the commit messages that did
not survive re-derivation; this description has been rewritten because the
previous version described a design the branch abandoned, and stated that the
option label was deliberately *not* matched while the code required it.

### DIFF
--- a/src/tmux/status_detection.rs
+++ b/src/tmux/status_detection.rs
@@ -504,8 +504,15 @@ pub(crate) fn claude_pane_is_ambiguous_typed_prompt(raw_content: &str) -> bool {
 /// menu. Requiring both keeps an assistant-authored numbered list from being
 /// mistaken for a prompt. `recent_lower` is the lowercased join of `recent`.
 fn claude_has_approval_prompt(recent: &[&str], recent_lower: &str) -> bool {
+    // The folder-trust prompt phrases its question without either stock
+    // opener, so a first launch in a new directory read as Idle and never
+    // reached the waiting count. Only the question is matched: its option
+    // label ("yes, i trust this folder", which the Antigravity detector below
+    // does carry) sits on a numbered-choice line, so matching that instead
+    // would let one quoted line satisfy both halves of the guard.
     let has_question = recent_lower.contains("do you want to")
-        || recent_lower.contains("would you like to proceed");
+        || recent_lower.contains("would you like to proceed")
+        || recent_lower.contains("is this a project you created or one you trust");
     has_question
         && recent
             .iter()
@@ -2610,6 +2617,50 @@ enter to select · esc to cancel";
         // turn's live spinner, so the fresh idle must read as Running.
         let pane = "✶ Working… (4s · ↓ 88 tokens)\n  esc to interrupt";
         assert_eq!(reconcile_claude_idle_hook_status(pane), Status::Running);
+    }
+
+    /// Verbatim `tmux capture-pane -p` of a claude pane parked at the
+    /// folder-trust prompt, 2026-08-15. `aoe status` read `0 waiting` while
+    /// this was on screen.
+    const CLAUDE_FOLDER_TRUST_PROMPT: &str = "\
+ Accessing workspace:
+ /tmp/scratch/exp
+ Quick safety check: Is this a project you created or one you trust? (Like your
+ own code, a well-known open source project, or work from your team). If not,
+ take a moment to review what's in this folder first.
+ Claude Code'll be able to read, edit, and execute files here.
+ Security guide
+ \u{276f} 1. Yes, I trust this folder
+   2. No, exit
+";
+
+    /// The trust prompt's option label is menu text, so matching it as the
+    /// question would collapse the two-signal guard: an assistant quoting the
+    /// option while working renders both signals on one line.
+    const CLAUDE_ASSISTANT_QUOTING_THE_TRUST_OPTION: &str = "\
+ I found the folder-trust handling in src/tmux/status_detection.rs. The two
+ menu options Claude renders are:
+   1. Yes, I trust this folder
+   2. No, exit
+ The detector matches those against the numbered-choice helper.
+ \u{2726} Working... (12s \u{b7} \u{2193} 431 tokens)
+   esc to interrupt
+";
+
+    #[test]
+    fn claude_assistant_quoting_the_trust_option_is_not_waiting() {
+        assert_eq!(
+            detect_claude_status(CLAUDE_ASSISTANT_QUOTING_THE_TRUST_OPTION),
+            Status::Running
+        );
+    }
+
+    #[test]
+    fn claude_folder_trust_prompt_is_waiting() {
+        assert_eq!(
+            detect_claude_status(CLAUDE_FOLDER_TRUST_PROMPT),
+            Status::Waiting
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


njbrake (dismissed):
Right shape, but incomplete: below roughly 69 columns the matched phrase wraps and the pane reads `Idle` again, which is the bug this fixes.

`"is this a project you created or one you trust"` is 45 characters, Claude word-wraps the prose, and `recent_lower` is a newline join, so the `contains` misses. Reproduced against Claude Code 2.1.234 launched in an untrusted directory under tmux:

| pane width | main | this PR | with `collapse_ascii_whitespace` |
| --- | --- | --- | --- |
| 60 / 66 / 68 | Idle | **Idle** | Waiting |
| 70 / 80 / 100 | Idle | Waiting | Waiting |

That band is one AoE produces itself. `src/tui/home/render.rs:2239` sizes a detached agent's pane to the visible preview rect, and with the default `list_width` of 35 the side-by-side preview is about viewport minus 37, so any viewport between the 80 column `STACKED_BREAKPOINT` and roughly 106 lands in the dead zone. Below 80 the stacked layout restores full width and the match works again, so the failure window is the ordinary laptop terminal.

Suggested fix, using the idiom already in this file (`claude_pane_has_running_signal` collapses whitespace for the same reason). `||` short-circuits, so the allocation is only paid when the two cheap arms miss:

```rust
let has_question = recent_lower.contains("do you want to")
    || recent_lower.contains("would you like to proceed")
    || collapse_ascii_whitespace(recent_lower)
        .contains("is this a project you created or one you trust");
```

Two smaller items:

- `CLAUDE_ASSISTANT_QUOTING_THE_TRUST_OPTION` renders the spinner as `Working...` with ASCII dots, but `claude_line_is_active_spinner` requires U+2026, so that fixture's `Running` verdict comes from `esc to interrupt` alone and it does not model what its comment claims.
- `claude_has_approval_prompt`'s doc comment still opens "when a tool needs the user's permission", which the folder-trust prompt is not.

On CodeRabbit's independent-signals finding: the mechanism is real, but the same false positive reproduces on `main` today through the existing `"do you want to"` arm. It is a pre-existing limit of the shared two-signal guard and scoping it touches every arm, so it belongs in its own PR and should not gate this one.

The decision not to match the option label is correct and well argued, and the fixture text checks out against the strings in the installed Claude binary. One caveat on my end: the width table comes from a line-for-line mirror of the detector rather than the compiled crate.

njbrake (approved):
Approving. The trust-prompt gap is real and the fix goes through the existing detector idiom rather than beside it. Verified locally rather than trusting the description:

- The matched strings are verbatim in Claude Code 2.1.234 (`Quick safety check: Is this a project you created or one you trust?`, `Yes, I trust this folder`); the 46/24/29 column figures check out.
- The wrap band is real: `render_preview` uses `Borders::ALL` plus `Padding::horizontal(1)`, so the pane is viewport minus list_width minus 4 side-by-side and minus 4 stacked, which puts the question's wrapped band at <=71 and 80..=106 as claimed.
- `claude_line_is_numbered_choice` strips a leading `>`, so the new `claude_trust_choice_option_text` rejecting it is load-bearing, not decoration; the blockquote fixture proves it.
- All 169 status_detection tests pass, plus the full lib suite serially (4603). I also probed beyond the fixtures: a 3-line label wrap matches, the documented residuals (trailing prose after the label, an echoed menu row after a numbered item) reproduce as `Waiting` exactly as the comments state, and a verbatim menu row mentioned in prose without a choice anchor stays `Running`.
- CodeRabbit's independent-signals finding is real but pre-existing on `main` through the `do you want to` arm; scoping the shared guard is its own change and should not gate this one.

One fix pushed on top: three em dashes in the new comments, which AGENTS.md bans (b2ea1b16 dropped one and later commits reintroduced three). c21d3abc swaps them for semicolons, comments only.

Optional, not blocking: the three `..._is_waiting` tests could be one table-driven test over the fixture constants, per the house one-test-per-behavior style.

_Note: this review was drafted by Claude Fable 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #3413, opened 2026-08-17
fix(session): require a transcript on disk before emitting --resume

blueagledev:
## Description

`claude --resume <sid>` fails with "No conversation found" when nothing was ever
written for `<sid>`, and the session lands in the "resume failed for sid ...;
preserved for explicit retry" state, so the pane is dead on every restart from
then on. Two paths on `main` can hand `acquire_session_id` exactly such an id.

**1. The `.claude.json` fallback arm** (`src/session/capture.rs`, in
`capture_claude_session_id`). `lastSessionId` is one slot per directory, and the
freshness gate around it reads the mtime of `.claude.json` itself, which any
live Claude anywhere rewrites. A months-dead id therefore passes the gate, and
every session rooted at that directory adopts it at once. Unlike the directory
scan above it, which can only return ids it found as files, this slot can name a
UUID no transcript was ever written for.

This gates the arm on the transcript being confirmed present, reusing the
existing `claude_host_transcript_confirmed_absent` (`capture.rs:148` on `main`).
Existence only, no mtime window, so an idle but real conversation still resumes.
The companion test asserts the arm still returns the id when the conversation
does exist.

What this costs, stated rather than left for review to find: the arm is
documented as the fallback for `/clear` and `/new`, and in the window where the
new thread has a `lastSessionId` but no jsonl yet, the arm now declines. The dir
scan picks the new id up as soon as content lands, and resuming it during that
window would have failed anyway, so the gap is short and self-healing.

**2. The observed-sid path** (`src/session/instance.rs`, in
`acquire_session_id_with`, not the `acquire_session_id` wrapper above it). The
empty-thread downgrade below it already covers a sid
loaded from storage, but a sid that arrived as a live observation returned early
and skipped it. The SessionStart hook fires before Claude writes any content, so
the sidecar can legitimately name a thread with no transcript. Rebinding rather
than returning early runs the observation through the same downgrade, and the
session launches with `--session-id <sid>` instead of `--resume <sid>`, which
succeeds.

## Reproduction

Measured on this branch, rebased onto `upstream/main` @ `d2b76ff1`, with the
two behaviour hunks withheld and then restored on the same tree:

    # behaviour withheld, tests applied
    cargo test --offline --lib -- --test-threads=1 \
      observed_sid_without_transcript_downgrades_to_fresh \
      claude_json_fallback_rejects_sid_with_no_transcript \
      claude_json_fallback_accepts_sid_with_transcript

    claude_json_fallback_rejects_sid_with_no_transcript
      panicked: a lastSessionId with no transcript must not be captured:
      "11111111-2222-3333-4444-555555555555"
    observed_sid_without_transcript_downgrades_to_fresh
      panicked: an observed sid with no transcript must launch as --session-id,
      never --resume
    test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 4597 filtered out; finished in 0.00s

    # behaviour restored, same tree, same invocation
    test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 4597 filtered out; finished in 0.01s

The one test that passes either way is
`claude_json_fallback_accepts_sid_with_transcript`, the companion asserting the
arm is narrowed rather than removed. It carries a second job: the reject test
alone cannot distinguish "the gate declined the id" from "the id was never read
at all", so if the fixture ever stopped reaching the arm, this companion is what
fails. That is also why its `.claude.json` key is canonicalized the way
`read_claude_json_session_id` reads it.

For an observed sid that *does* have a transcript, hunk 2 is behaviour
identical: same returned tuple, same `agent_session_id` write, same log line.
The rebinding only changes the case the downgrade below was already written for.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (none needed: no CLI surface or
      config change)
- [x] For UI changes: N/A

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: both paths are inside session-id acquisition and
      are asserted by three unit tests that build the `.claude` layout on disk.
      An e2e would have to fabricate the same layout and then read back the
      launched command line, which is what the unit tests already do without a
      real agent.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Carried downstream since
2026-08-15 and re-verified against current `main` before being offered.

## Summary

- Prevents startup from using `--resume` when the session has no transcript.
- Confirms fallback and observed session IDs against profile-specific transcript files.
- Uses `--session-id` for sessions without transcripts.
- Adds regression tests for missing, existing, stale, and profile-specific transcripts.

## Benefits

- Prevents Claude’s “No conversation found” error.
- Allows affected panes to restart correctly.
- Preserves recovery for sessions with existing transcripts.

### DIFF
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -2876,7 +2876,7 @@ mod tests {
     /// jsonl is older than the 5-minute live-capture window, so
     /// `capture_claude_session_id` falls past the dir scan, plus a
     /// `.claude.json` naming `last_session_id` for that directory. Returns the
-    /// project path and the encoded transcript dir.
+    /// encoded transcript dir.
     fn claude_json_fallback_home(
         temp: &tempfile::TempDir,
         project_path: &str,
--- a/src/session/capture.rs
+++ b/src/session/capture.rs
@@ -154,7 +154,23 @@ pub(crate) fn capture_claude_session_id(
         let is_fresh = claude_json
             .and_then(|t| t.elapsed().ok())
             .is_some_and(|age| age <= Duration::from_secs(5 * 60));
-        if is_fresh && Uuid::parse_str(&id).is_ok() {
+        // `is_fresh` is the mtime of `.claude.json` itself, which any live
+        // Claude anywhere rewrites, so it says nothing about when *this*
+        // directory's `lastSessionId` was set and a value months old still
+        // reads as fresh. And unlike the dir scan above, which can only return
+        // ids it found as files, this slot can name a UUID no transcript was
+        // ever written for. Requiring the conversation to exist is what keeps
+        // `--resume` off an id Claude answers with "No conversation found",
+        // which leaves the pane dead on every restart from then on.
+        //
+        // The cost is the short window after `/clear` or `/new` where the slot
+        // names a thread Claude has not written to yet: the arm declines it,
+        // and the dir scan picks the new id up once content lands. Resuming it
+        // in that window would fail anyway.
+        if is_fresh
+            && Uuid::parse_str(&id).is_ok()
+            && !claude_host_transcript_confirmed_absent(&canonical.to_string_lossy(), &id, host_env)
+        {
             return Ok(id);
         }
     }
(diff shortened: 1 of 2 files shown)


njbrake (approved):
Approving. This extends the mechanism the project already built for this failure mode rather than adding a second one beside it, and the second hunk gets there by deleting a special case instead of adding one.

I replicated `capture_claude_session_id` and its helpers against fixtures matching the new tests and ran a six case matrix. The gate flips only the no-transcript cases; an idle but real conversation stays resumable via the `.claude.json` arm, and the dir scan picks the id back up the moment the jsonl lands, so the post-`/clear` decline is self-healing as the description says. I also walked the existing tests that could regress (`supersedes_stale_claude_sid_after_clear`, `no_bascule_when_claude_stored_matches_freshest`, `sidecar_wins_over_fresher_peer_jsonl`, `mtime_fallback_applies_without_sidecar`, and the `resume_fallback` e2e); all of them seed a transcript for the observed id, so none changes verdict.

Two things worth folding in before merge, neither blocking:

1. `capture_claude_session_id`'s doc comment at `src/session/capture.rs:117-124` is now slightly stale. It still ends on the `/clear` and `/new` fallback, which is the one case the gate narrows. One sentence saying the arm now yields only transcript-backed ids would cover it.
2. The two new test fixtures hand-roll a `FileTimes` call while `set_mtime_secs` already exists in the same `mod tests` (defined below the new code, which is probably why it was missed). Reusing it drops about 14 lines.

One note for the commit message rather than the code: hunk 1 also gates `claude_poll_fn` and `self_heal_session_id`, which the description does not mention. I traced both and they are improvements. The poller returns `None` for a tick and retries, and `self_heal_session_id` only runs when `agent_session_id` is empty, which is where adopting a phantom `lastSessionId` would persist it and produce a doomed `--resume` at the next launch. Worth stating so whoever bisects to this commit later sees the full reach.

I approved the held CI runs for `0d9cb4ee` after auditing the diff. My approval assumes those go green, since I reviewed without a local build.

PR #3414, opened 2026-08-17
feat(cli): add state and timestamps to `session show --json`

blueagledev:
## Description

#3350, shipped as #3361, taught `aoe list --json` a `state` tag plus
`trashed_at` and `archived_at`, so a scripted consumer can tell a trashed or
archived row from a live one. The lookup path was left behind:
`SessionDetails` in `src/cli/session.rs` carries none of the three, so
`aoe session show --json <id>` cannot answer the question `aoe list --json` now
answers, and a script that already holds an id has to shell out to a full
listing and filter it back down.

This adds the same three keys, with the same vocabulary and the same source of
truth: `state_tag` is reused from `cli::list` rather than re-derived, so the
`Trashed > Archived > Active` precedence keeps exactly one definition and an
archived row that is then trashed reports `trashed` on both surfaces. `status`
cannot carry this, because a session can be archived and running at the same
time.

Both timestamps keep `skip_serializing_if = "Option::is_none"`, so a live row
gains exactly one key (`state`) over today's output and no `null`s.

The human-readable `session show` gains a `State:` line, printed only when the
session is not live, so its output is unchanged for every live session.

The JSON construction moves out of `show_session` into `session_details`, so it
can be asserted without standing up storage, mirroring `session_json` in
`cli::list`.

## Tests

Three unit tests in `cli::session`, mirroring the ones #3361 added for
`cli::list`: an archived row reports `state: "archived"` with `archived_at` and
no `trashed_at`; a row archived and then trashed reports `trashed` with both
stamps present; a live row serializes with `"state":"live"` and neither
timestamp key.

Two e2e tests in `tests/e2e/cli.rs`, companions to
`test_cli_list_state_filter_and_json_shape` for the lookup path: one drives
`aoe rm` and asserts `live` with no timestamps before, `trashed` with a
`trashed_at` after; the other drives `aoe session archive`, asserts `archived`
with `archived_at` and no `trashed_at`, then trashes the same row and asserts
`state` reports `trashed` with both stamps present, since `trash()` deliberately
keeps `archived_at` so a restore is faithful.

`contrib/hermes-skill/SKILL.md` and `contrib/openclaw-skill/SKILL.md` document
the literal `session show --json` object, and #3361 updated both when it changed
`list --json`; both gain `state` and the sentence describing the vocabulary.
`cargo xtask check-skill` validates frontmatter and `aoe <subcommand>` paths, not
the JSON blocks, so nothing in CI would have caught this.

## PR Type

- [x] New Feature

## Checklist

- [x] Documentation was updated where necessary: both `contrib/*/SKILL.md`
      files that document the `session show --json` shape carry `state`.
      `docs/cli/reference.md` is generated from clap help and no flag changed,
      so it is unaffected
- [x] For UI changes: N/A

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (two)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** A downstream patch had added
`archived_at` to both `list --json` and `session show --json` before #3361
landed. The `list` half is now redundant, and upstream's shipped shape is wider
than ours was, so only the `show` half is offered here and it follows #3361's
vocabulary rather than the downstream one.

## What changed

- Added `state`, `trashed_at`, and `archived_at` to `aoe session show --json`.
- Added lifecycle state to human-readable output for non-live sessions.
- Reused the state vocabulary and `Trashed > Archived > Active` precedence from `aoe list --json`.
- Clarified that archive and trash timestamps are independent and not durable state indicators.
- Added unit and end-to-end tests for live, archived, trashed, and archived-then-trashed sessions.

## Benefits

- Provides consistent lifecycle information across CLI commands.
- Preserves `archived_at` when an archived session is trashed.
- Separates lifecycle state from live pane status.
- Improves confidence through focused test coverage.

## Potential enhancement

- Document state precedence in the main CLI reference.

### DIFF
--- a/contrib/hermes-skill/SKILL.md
+++ b/contrib/hermes-skill/SKILL.md
@@ -138,11 +138,12 @@ aoe status -q   # just the waiting count (for scripting)
   "tool": "claude",
   "command": "claude",
   "status": "running",
+  "state": "live",
   "profile": "default"
 }
 ```
 
-`parent_session_id` is included only for sub-sessions.
+`parent_session_id` is included only for sub-sessions. `state` is `live`, `archived`, or `trashed`, with `trashed_at` / `archived_at` present only in those states, the same vocabulary `aoe list --json` uses; `status` is the pane's live status and does not carry it, since an archived session can still be running.
 
 **JSON shape** (`aoe status --json`):
 ```json
--- a/contrib/hermes-skill/SKILL.md
+++ b/contrib/hermes-skill/SKILL.md
@@ -89,7 +89,7 @@ aoe list --json --state=live   # skip trashed and archived rows
 ]
 ```
 
-`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `state` is `live`, `archived`, or `trashed`, with `trashed_at` / `archived_at` present only in those states; a trashed session stays in the listing and keeps its title, so check `state` before treating a row as a live session. `list --json` does not include live status: use `aoe status --json` or `aoe session capture --json` for that.
+`command` is omitted when empty; `worktree` appears only for worktree-backed sessions. `state` is `live`, `archived`, or `trashed`; `trashed_at` and `archived_at` are set independently and each persists once set, so a session archived and then trashed carries both and reports `trashed`. A trashed session stays in the listing and keeps its title, so check `state` before treating a row as a live session. `list --json` does not include live status: use `aoe status --json` or `aoe session capture --json` for that.
 
 ## Session Lifecycle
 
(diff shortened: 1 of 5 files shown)


njbrake (approved):
Approving. Symmetric completion of what #3361 shipped one command over, built through the existing mechanism: `state_tag` is called rather than copied, so the `Trashed > Archived > Active` precedence keeps one definition and the two surfaces cannot drift.

The CI runs I approved for `aaabbfa5` settle the load-bearing claims. All three new unit tests pass in `bare-core` and both e2e tests pass in `serve-e2e`, so the trashed-over-archived precedence and the "`archived_at` survives trashing" behavior were executed against the real binary rather than reasoned about. The Docs check also passed, and since that job runs `cargo xtask gen-docs` and then `git diff --exit-code docs/cli/reference.md`, it confirms no regeneration is needed here. The two `contrib/*/SKILL.md` files are the only hand-maintained mirrors of this object and both are updated.

Three wording nits in the contrib docs, all non-blocking:

1. `favourite` is the only British spelling in the repo; the command is `aoe session favorite`.
2. The durability sentence still reads as though anything that wakes a session clears `trashed_at`. `touch_last_accessed` (`src/session/instance.rs:1741`) clears `archived_at`, `snoozed_until`, and `idle_dormant_since`, never `trashed_at`, which only `untrash()` and so `aoe session restore` clears. Also, `pin` is not reachable from the CLI at all, so it is an odd example in a CLI skill doc. The same imprecision is in the parallel `list --json` paragraph that predates this PR.
3. `trash()` names an internal Rust method to third-party consumers; "trashing" reads the same without leaking the symbol.

One thing worth deciding rather than changing: `aoe list` prints the STATE column for every row including `live` when the scope is `all`, while `session show` prints `State:` only when the session is not live. Keeping live output byte-identical is a sound reason, but it does mean a human cannot tell "this session is live" from "this build predates the feature".

Also flagging that `codecov/project` currently reads as failing. It was evaluated on a partial upload seconds before `codecov/patch` passed, and this diff touches no `web/src` files, so I expect it to settle once the Playwright uploads land. Worth a recheck rather than a change.

PR #3421, opened 2026-08-18
test(session): stop assuming an FHS layout in the launch-path tests

blueagledev:
## Description

Several unit tests hardcode paths that only exist under a Filesystem Hierarchy
Standard layout, so `cargo test` fails on hosts that do not have them (NixOS,
minimal containers). No production code changes; the diff is entirely inside
`#[cfg(test)]` modules.

Four sites, all the same class:

| Test | Assumed |
|---|---|
| `test_wrap_command_reasserts_working_dir_after_login_shell` | `SHELL=/bin/bash`, then execs it |
| `omp_capture_gate_executes_nested_stdin_scripts` | `PATH=<shim>:/usr/bin:/bin` |
| `omp_routing_fingerprint_accepts_matching_live_env_and_rejects_drift` | `env_clear()` + `PATH=/usr/bin:/bin` |
| `test_container_env_file_does_not_mutate_host_process_environment` | `PATH=/usr/bin:/bin`, runs `/bin/cat` |

On this host `/bin` contains only `sh`, so the first fails on itself:

```
wrapped command failed: bash: line 1: /bin/bash: No such file or directory
```

Each now resolves what it needs instead of assuming a location: `bash` via
`which` (already a regular dependency, 17 call sites before this PR), and the
shim `PATH` prepended to the caller's inherited one rather than replacing it. In
`test_container_env_file_...` the `PATH` is asserted back — that assertion is
the point of the test, proving the env file did not mutate the host
environment — so it is taken from the caller, passed in, and asserted against
the same value.

### The `SHELL` literal was also escaping its serial group

`std::env::set_var` is process-global, so the literal did not stay inside
`#[serial_test::serial(shell_env)]` — it leaked to every test running
concurrently. The `repo_config` hook tests spawn the user's shell
(`repo_config.rs:1019` → `user_shell()`, which reads `$SHELL`), so they
inherited a path that does not resolve and failed with `os error 2`. That is
why they pass under `--test-threads=1` and fail in parallel, which reads like
flakiness but is not.

The override now goes through `EnvGuard::set`, which restores on `Drop`
including on an unwind — the previous manual restore at the end of the test did
not run if an assertion failed first — and serialises against other `EnvGuard`
users.

Being precise about what that does **not** do, since an earlier revision of this
description overstated it: `ENV_LOCK` is acquired only by `EnvGuard`, so it does
not hide the override from tests that take no guard, and 12 of the 14
`repo_config` hook tests take none. They still observe the override. What makes
that harmless is resolving a real `bash`, so the value they see is one they can
actually run. The guard buys panic-safety and guard-mutual-exclusion; the
`which` lookup is what fixes the 14.

### Measured on a NixOS host

Totals move run to run, so a single before/after pair would be misleading.
Three full `cargo test --lib` runs each:

| | run 1 | run 2 | run 3 |
|---|---|---|---|
| `main` @ 26e5fe93, sweep A | 25 | 24 | 26 |
| `main` @ 26e5fe93, sweep B | 27 | 27 | 26 |
| this branch, sweep A | 3 | 5 | 5 |
| this branch, sweep B | 4 | 3 | 2 |

Two independent three-run sweeps, because a single sweep understated the
spread: `main` ranges 24-27 and the branch 2-5. The set difference below was
**identical** in both sweeps, which is why it is the figure to read.

The stable figure is the set difference, not a subtraction of noisy totals:
**22 tests fail in every `main` run and in no run of this branch**, 14 of them
the `repo_config` hook tests.

What still fails on this branch, by how many of the three runs it appeared in:

```
3/3  tui::dialogs::hooks_install::tests::test_content_shows_settings_path
1/3  tmux::session::tests::{a_stacked_split_composites_one_line_per_window_row,
     send_key_tokens_appends_no_implicit_enter, size_owner_lock_claims_...,
     test_capture_pane_targets_first_window_with_multiple_windows,
     test_is_attached_true_with_live_client, test_is_pane_dead_on_running_session,
     vt_owner_lock_claims_rejects_and_releases_independently}
1/3  tmux::terminal_session::tests::test_terminal_session_is_pane_dead_after_command_exits
1/3  tmux::test_helpers::tests::drop_kills_session
```

Neither group is an FHS problem, and neither is addressed here:

- The `tmux::*` ones are **not** thread-parallelism flakes, and an earlier
  revision of this description wrongly said they pass single-threaded. They do
  not: `cargo test --lib -- --test-threads=1` on this branch still fails
  `size_owner_lock_claims_rejects_steals_and_releases` and
  `test_is_attached_true_with_live_client`, and the same reproduces on `main`.
  Run individually they pass, so the cause is in-process state left behind by
  earlier tmux tests in the same process, not concurrency. tmux 3.6a is
  installed; a missing server is not the explanation either. Separate issue,
  not addressed here, and I am not claiming a diagnosis beyond "not this PR".
- `test_content_shows_settings_path` is the only stable remainder and is not
  environment-layout related at all. It renders through
  `agent_settings_path_display_for_host_environment`, which honours
  `$CLAUDE_CONFIG_DIR`; with that variable set the dialog shows that path
  instead of `~/.claude/settings.json`. Verified directly:

  ```
  $ cargo test --lib tui::dialogs::hooks_install::tests::test_content_shows_settings_path
  test result: FAILED. 0 passed; 1 failed
  $ env -u CLAUDE_CONFIG_DIR cargo test --lib ...same test...
  test result: ok. 1 passed; 0 failed
  ```

  It depends on the *value*, not merely on the variable being set:
  `CLAUDE_CONFIG_DIR=$HOME/.claude` makes it pass, since the rendered path then
  ends in `.claude/settings.json` anyway. So it fails wherever the variable
  points somewhere else, on any OS. Happy to fix it in a follow-up if you want
  it.

### Why CI does not see this

The workflows run `ubuntu-latest` and `macos-latest` only, both of which have
`/bin/bash` and FHS coreutils. Worth flagging a near miss: `flake.nix` defines
`checks.aoe-test = craneLib.cargoTest`, and no job runs `nix flake check` — the
Nix Build job runs `nix build` with `doCheck = false`. I ran that check on both
sides rather than guess at it: `main` 27 failed, this branch 8 failed. So it
catches 19 of the 22, not all of them (three want tmux, which the sandbox has
no `nativeBuildInputs` entry for), and it also surfaces 8 failures unrelated to
this PR that persist on the branch — sandboxed network and nosuid cases. The
check would not go green on this branch; it would go from 27 to 8.

I have not run the suite on an FHS host, so I am not claiming it is unchanged
there; the changes are all "resolve instead of assume", which should be a no-op
where the assumed paths exist.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (see the measured remainder above: 2-5 tests
  still fail on my host, all pre-existing and none introduced here)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The change is to the tests themselves; the existing tests are the coverage, and
the evidence is the 22-test stable set difference above.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Worth disclosing how the first revision of this PR was wrong, since it was
public for a while. As opened it claimed "26 → 5", "19 cleared", and described
the remainder as "tmux pane-liveness and dialog-path tests that want a live
tmux server". An adversarial review pass re-derived every number and found:
the totals are unstable so the pair was a best-case cherry-pick; 19 contradicted
the PR's own table (26 − 5 = 21); and the remainder description was simply
false — two of those remaining failures were further instances of the very
`/usr/bin:/bin` bug this PR is about, which the wording hid. Those two are now
fixed here, and the numbers above are the stable set difference across three
runs per side.

Review rounds since: Copilot's `EnvGuard` suggestion and CodeRabbit's
non-UTF-8 `PATH` point were both adopted in `6a30089e` — the first removes the
`SHELL` leak outright instead of neutralising it, the second keeps the
round-tripped `PATH` as bytes. The `to_string_lossy` comments from the first
round were against the original commit and were already fixed in `f92f364b`.

That pass also caught a real regression in the first commit: the new helper was
inserted between an existing `#[cfg(unix)]` and the `#[test]` it guarded,
dropping the gate from the test. Fixed in the follow-up commit.

Stated carefully, because an earlier revision of this description overstated it:
this restores a gate, it does not avert a Windows build break. `mod tests` in
that file carries no `cfg(unix)` and already contains ungated `std::os::unix`
uses on `main` (`instance.rs:13014` and `:16166`), so the Windows test build was
broken before this PR and still is. No workflow runs Windows.

---

**Revision note.** This description has been corrected twice after adversarial
review passes over my own commits. Claims that were published and are now known
to have been wrong: "26 → 5" and "19 cleared" (cherry-picked from unstable
totals; the stable set difference is 22), "the remaining 5 are tmux tests that
want a live tmux server" (two of them fail single-threaded; two more were
further instances of the bug this PR fixes), "would hit exactly these failures
in a nix sandbox" (19 of 22, and it adds 8 unrelated), "18 existing call sites"
(17), "would have broken the Windows test build" (already broken on `main`), and
that `EnvGuard` stops concurrent tests observing the override (it does not).
Flagging them rather than quietly editing, since the earlier text was public.

### DIFF
--- a/src/session/environment.rs
+++ b/src/session/environment.rs
@@ -771,6 +771,7 @@ pub(crate) fn build_docker_env_args_with_managed_codex_home(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::session::test_support::EnvGuard;
 
     fn owned(pairs: &[(&str, &str)]) -> Vec<(String, String)> {
         pairs
@@ -1919,72 +1920,43 @@ environment = ["GH_TOKEN=write_token"]
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_shell_reads_env() {
-        let original = std::env::var("SHELL").ok();
-        std::env::set_var("SHELL", "/bin/zsh");
+        let _shell = EnvGuard::set(&[("SHELL", "/bin/zsh")]);
         assert_eq!(user_shell(), "/bin/zsh");
-        match original {
-            Some(v) => std::env::set_var("SHELL", v),
-            None => std::env::remove_var("SHELL"),
-        }
     }
 
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_shell_fallback() {
-        let original = std::env::var("SHELL").ok();
-        std::env::remove_var("SHELL");
+        let _shell = EnvGuard::unset(&["SHELL"]);
         assert_eq!(user_shell(), "bash");
-        if let Some(v) = original {
-            std::env::set_var("SHELL", v);
-        }
     }
 
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_shell_empty_falls_back() {
-        let original = std::env::var("SHELL").ok();
-        std::env::set_var("SHELL", "  ");
+        let _shell = EnvGuard::set(&[("SHELL", "  ")]);
         assert_eq!(user_shell(), "bash");
-        match original {
-            Some(v) => std::env::set_var("SHELL", v),
-            None => std::env::remove_var("SHELL"),
-        }
     }
 
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_posix_shell_returns_posix() {
-        let original = std::env::var("SHELL").ok();
-        std::env::set_var("SHELL", "/bin/zsh");
+        let _shell = EnvGuard::set(&[("SHELL", "/bin/zsh")]);
         assert_eq!(user_posix_shell(), "/bin/zsh");
-        match original {
-            Some(v) => std::env::set_var("SHELL", v),
-            None => std::env::remove_var("SHELL"),
-        }
     }
 
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_posix_shell_falls_back_for_fish() {
-        let original = std::env::var("SHELL").ok();
-        std::env::set_var("SHELL", "/usr/bin/fish");
+        let _shell = EnvGuard::set(&[("SHELL", "/usr/bin/fish")]);
         assert_eq!(user_posix_shell(), "bash");
-        match original {
-            Some(v) => std::env::set_var("SHELL", v),
-            None => std::env::remove_var("SHELL"),
-        }
     }
 
     #[test]
     #[serial_test::serial(shell_env)]
     fn test_user_posix_shell_falls_back_for_nu() {
-        let original = std::env::var("SHELL").ok();
-        std::env::set_var("SHELL", "/usr/bin/nu");
+        let _shell = EnvGuard::set(&[("SHELL", "/usr/bin/nu")]);
         assert_eq!(user_posix_shell(), "bash");
-        match original {
-            Some(v) => std::env::set_var("SHELL", v),
-            None => std::env::remove_var("SHELL"),
-        }
     }
 
     #[test]
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -11672,7 +11672,10 @@ mod tests {
             command
                 .args(["-c", &script])
                 .env_clear()
-                .env("PATH", "/usr/bin:/bin");
+                // Same reason as `test_path_with_shim`: `env_clear` is here to
+                // control the OMP_STORE_ENV_KEYS the fingerprint folds in, not
+                // to pin a filesystem layout, so inherit the caller's PATH.
+                .env("PATH", std::env::var_os("PATH").unwrap_or_default());
             // Pin the exact routing environment a host launch installs into the
             // pane for this HOME, so the check reproduces the fingerprint's env
             // instead of assuming the ambient OMP_STORE_ENV_KEYS are empty. They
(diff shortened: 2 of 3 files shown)


njbrake (approved):
Approving the FHS work as reviewed at `aca186d3`. It adds zero test functions and no new assertions, which is the cheapest shape this change could take, and it reuses the mechanisms the project already has: `EnvGuard` for env shimming and the `split_paths`/`join_paths` idiom from `src/acp/adapters.rs:307`. The test copy is the only one of the four copies of that idiom that handles the empty-`PATH` case, and I confirmed the trap is real: `split_paths("")` yields one empty entry, so the naive join puts cwd on the child's `PATH`.

Verified on the branch in an isolated clone: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets` at 6 findings all pre-existing in untouched files, `cargo test --lib` at 4594 passed with the single failure being the `CLAUDE_CONFIG_DIR` case the PR itself documents. The motivating claim also holds: `flake.nix:159` defines `checks.aoe-test`, no workflow runs `nix flake check`, and CI's `nix build` sets `doCheck = false`, so the flake ships a test check that cannot pass.

**Disclosure:** I pushed `d84c6349` to this branch, which finishes the sweep. It converts the other nine `set_var("SHELL", ...)` sites and takes `EnvGuard::unset(&[])` purely as a lock in the two tests that now read the inherited `PATH` without a serial annotation. That second half matters because `crate::acp::node` and `crate::acp::acp_client` scrub `PATH` process-globally under `#[serial]`, which excludes annotated peers but not those two, so this PR traded an FHS assumption for a narrow new flake window. `cargo test --lib` goes to 4595 passed / 0 failed with it applied. That commit is my own code and this approval does not cover it; it wants someone else's eyes before merge.

Two things left for you:

1. 33 of the 54 added lines are comments, and four of them narrate the change ("Resolved rather than hardcoded to `/bin/bash`") instead of stating the current invariant. Reworded, the set is roughly 8 lines shorter. The load-bearing ones (the empty-`PATH` trap, `var_os` over `var`, why `env_clear` is there) should stay verbatim.
2. `which::which("bash").expect("bash on PATH")` hard-fails on a container that ships only `dash`. `src/plugin/host.rs:883` skips with a message instead, which fits this PR's own goal better.

njbrake (approved):
Approving through `67e28071`. This covers the four commits added after my earlier approval at `d84c6349`, plus `d84c6349` itself, which was mine and which I said wanted separate eyes. It was reviewed here in a fresh context with no hand in writing it.

Verified on the branch and again on a local merge with current `main`: `cargo fmt` clean, `cargo clippy --all-targets` at 6 warnings, all pre-existing in `src/tui/home/mod.rs` and `src/session/groups.rs`, and `cargo test --lib` at 4595 passed / 0 failed both in parallel and at `--test-threads=1`, plus 4638 / 0 on the merge. This host has an FHS layout, so that also confirms the "no-op where the assumed paths exist" claim you could not test.

Two things I reproduced rather than took on faith:

1. `test_container_env_file_does_not_mutate_host_process_environment` did not become a tautology. Injecting the bug it guards, sourcing the payload env file before printing `$PATH`, fails it; so does a PATH difference that leaves `cat` resolvable, on the byte comparison. The discriminator is still "child PATH is not the env file's PATH".
2. The `#[serial]` reasoning holds. All eleven process-global PATH mutators in `update::install`, `acp::node` and `acp::acp_client` carry the default key, so default-key `#[serial]` on the two new readers buys real mutual exclusion. No `EnvGuard` caller sets PATH, so the guard-versus-annotation gap your comment describes is accurate.

Your `ENV_LOCK` correction is the right one, and it has a consequence worth stating plainly: nine sites still export a hardcoded `/bin/zsh`, `/usr/bin/fish` or `/usr/bin/nu` process-globally, where the unguarded hook tests can read it. Those are microsecond string builders rather than process spawns, so this is strictly better than `main`, but the leak is narrowed rather than closed, and my "finishes the SHELL sweep" overstated it. Two follow-ups worth filing: have the hook tests pin `SHELL` themselves, and move this test off `#[serial(shell_env)]` to the default key, since `which("bash")` now turns a PATH scrub race into a silent skip instead of a failure.

I pushed `67e28071` for the nits: the two em dashes AGENTS.md forbids, a helper-site count that was wrong and will rot again (dropped rather than re-corrected), three comments narrating the old literal instead of the current invariant, and a readable failure message for the byte comparison.

Issue #1754, opened 2026-06-01
Cockpit: aoe update leaves running workers on old code, daemon re-adopts stale runners without a build-version check


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Updating `aoe` to a new version does not propagate the new code to running cockpit workers. After an update, the per-session `aoe __cockpit-runner` processes keep executing the OLD binary's code. They only pick up the new build when they are manually killed and recreated, or when a brand-new session spawns a fresh runner.

This is the direct tradeoff of #1037 (workers now survive `aoe serve` stop/restart). Surviving the daemon is the right behavior for in-flight turns, but it means a daemon that has been swapped to a newer binary happily re-adopts workers that are still running the previous release. The cockpit silently runs mixed versions: a new daemon talking to old runners.

Two distinct gaps stack here:

1. `aoe update` does not restart the daemon at all; it only prints a "restart aoe" hint. So even the daemon stays on the old code until the user manually restarts `aoe serve`.
2. Even once the daemon is restarted on the new binary, the reconciler adopts the old workers without any binary-version check, so the workers stay stale.

## What's there today

The update flow swaps the binary and tells the user to restart, with no daemon restart and no worker signal:

`src/cli/update.rs:107`
```rust
println!(
    "✓ Updated to v{}. Restart `aoe` to use the new version.",
    info.latest_version
);
```

The binary is replaced atomically in place, so a running process keeps its already-loaded code:

`src/update/install.rs:292` (`atomic_replace`)
```rust
std::fs::rename(&source, &target)?;
```

Workers are detached `aoe __cockpit-runner` processes spawned by the daemon (`src/cockpit/acp_client.rs:1856`+, spawn at `acp_client.rs:1934` via `cmd.arg("__cockpit-runner")`). By design they survive daemon restart (#1037).

On daemon startup the reconciler decides per session whether to adopt the existing worker or spawn fresh. Adoption is gated solely on liveness:

`src/cockpit/cockpit_reconciler.rs:537`
```rust
if let Ok(Some(record)) = crate::cockpit::worker_registry::load(&id) {
    if crate::cockpit::worker_registry::is_record_live(&record) {
        // attach to the running worker ...
    }
}
```

`is_record_live` is where a version gate could live, but the only version it checks is the on-disk schema version, not the build version:

`src/cockpit/worker_registry.rs:400`
```rust
[...]

PR #3432, opened 2026-08-18
fix(build): stop watching .git/index for the build-version rerun trigger

cwrau:
## Description

`build.rs` watches `.git/HEAD` and `.git/index` via `cargo:rerun-if-changed` so the embedded `AOE_BUILD_VERSION` (and its dirty-flag suffix) recomputes when the checked-out revision or staged state changes.

The `index` half backfired: git rewrites `index`'s stat cache on a plain `git status` (racily-clean revalidation), not just on real `git add`/staging. Anything that runs `git status` in the background — shell prompts, editor/IDE git integrations — touches `index`'s mtime, so cargo treated the crate as dirty and recompiled the lib + binary on effectively every build, even with zero source changes. Watching `HEAD` alone is a coarser but reliable signal; the dirty-flag suffix now only recomputes on the next real rebuild trigger (checkout/commit) instead of on every restart.

No functional change to the worktree-safe path resolution from #2050 or the `AOE_BUILD_VERSION` embedding logic itself — only which git file triggers a rerun.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- Updated the existing build-script regression test (`tests/build_version_rerun.rs`) in place: it now asserts only `HEAD` is watched instead of `HEAD` + `index`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Sonnet 5 (Claude Code)

Root-caused via `journalctl` timing correlation on a systemd service running `cargo run` on this crate, then confirmed directly with `touch .git/index` reproducing a full rebuild.

## Summary

- Replace `.git/index` with `logs/HEAD` as a build watch path.
- Continue watching `HEAD` to detect checkouts and detached-HEAD changes.
- Add regression coverage for commits and plain `git status`.
- Preserve worktree-safe path resolution and version embedding.

## Benefits

This prevents routine `git status` operations from triggering unnecessary rebuilds. It still detects revision changes after commits, pulls, merges, rebases, and resets.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -35,10 +35,11 @@ fn emit_build_version() {
     use std::process::Command;
 
     // Re-run when the committed revision changes or an override toggles.
-    // Only HEAD is watched (not `index`, which git also rewrites on a plain
-    // `git status` and would otherwise force a rebuild on every restart, not
-    // just on real staging — see `git_watch_paths`'s doc comment). Resolve
-    // the real path via `git rev-parse --git-path` rather than hardcoding
+    // HEAD moves on checkout; logs/HEAD moves on every commit, pull, merge,
+    // rebase, or reset. `index` is not watched: git rewrites it on a plain
+    // `git status`, which forced a full recompile on builds with no source
+    // change (see `git_watch_paths`'s doc comment). Resolve the real paths
+    // via `git rev-parse --git-path` rather than hardcoding
     // `.git/HEAD`: in a git worktree `.git` is a file pointing at
     // `<main>/.git/worktrees/<name>/`, so the literal `.git/HEAD` path does not
     // exist. Cargo treats a missing `rerun-if-changed` input as perpetually
--- a/build.rs
+++ b/build.rs
@@ -35,9 +35,11 @@ fn emit_build_version() {
     use std::process::Command;
 
     // Re-run when the committed revision changes or an override toggles.
-    // HEAD moves on checkout/commit; index moves on stage. Resolve the real
-    // paths via `git rev-parse --git-path` rather than hardcoding `.git/HEAD`:
-    // in a git worktree `.git` is a file pointing at
+    // Only HEAD is watched (not `index`, which git also rewrites on a plain
+    // `git status` and would otherwise force a rebuild on every restart, not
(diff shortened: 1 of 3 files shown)


njbrake (approved):
Approving, with a follow-up commit pushed to this branch (a0f73575).

The root cause is right and I reproduced both halves: a plain `git status` rewrites `.git/index` after any stat-only churn, and `touch .git/index` forces a full lib plus binary recompile even when the build script's emitted value is unchanged.

What the fix missed is that `HEAD` holds `ref: refs/heads/<branch>`, so advancing the branch never rewrites `HEAD` itself:

| operation | .git/HEAD | .git/logs/HEAD | .git/index |
|---|---|---|---|
| plain `git status` | unchanged | unchanged | rewritten |
| `git commit` | unchanged | changed | changed |
| `git pull --ff-only` | unchanged | changed | changed |
| `git reset --hard` | unchanged | changed | changed |
| `git fetch` | unchanged | unchanged | unchanged |
| `git checkout` | changed | changed | changed |

So with `HEAD` alone, `git pull` plus `cargo build` recompiles the sources but does not rerun `build.rs`, and the new binary still reports the old commit. `is_build_current` then matches a worker still running the previous binary, and the #1754 respawn gate never fires. That lands on the run-from-source, restart-the-daemon workflow this PR is meant to help.

The follow-up watches `logs/HEAD` alongside `HEAD`. It moves on commit, pull, merge, rebase, and reset, resolves per worktree through `git rev-parse --git-path`, and stays quiet through `git status` and `git fetch`, so it does not bring the index churn back. It also adds a test pinning both halves of the contract: a commit must disturb a watched path, a plain `git status` must not. That test fails on the previous commit, and fails again if `index` is ever re-added.

The lagging dirty suffix I left alone. It is documented as coarse, its only consumer is the respawn gate, and its apparent freshness before was a side effect of the churn you removed.

`cargo fmt`, `cargo clippy --all-targets`, and `cargo test` are clean locally.

Thanks for tracking this down, and for the journalctl correlation in the description. It made the claim quick to verify.

PR #3451, opened 2026-08-20
test(session): pin SHELL in the hook tests that spawn one

### LINKED ISSUE #3449: test: repo_config hook tests spawn $SHELL unguarded while EnvGuard sites export shells that may not exist
Follow-up from #3421, filed at @njbrake's request in his approval there.

#3421 closed the `PATH` window on the launch-path tests. The `SHELL` one is still
open, and the shape is the same one you corrected in that PR: **the writers hold
a lock the readers never take.**

`EnvGuard::set` acquires `test_support::ENV_LOCK` for the guard's lifetime
(`src/session/test_support.rs:139-145`), so the `SHELL` writers exclude each
other. The readers are the `repo_config` hook tests, and the file contains
exactly two `EnvGuard` uses in total:

    src/session/repo_config.rs:2860   EnvGuard::unset(&["GIT_TERMINAL_PROMPT"])
    src/session/repo_config.rs:2949   EnvGuard::unset(&["GIT_TERMINAL_PROMPT"])

Everything else in that module that spawns a hook runs unguarded, so it reads
whatever `SHELL` a concurrent writer has installed.

**What the reader does with it.** `build_hook_command`'s `HookTarget::Local` arm
resolves `user_shell()` and spawns it (`src/session/repo_config.rs:1019-1020`):

    let shell = super::environment::user_shell();
    let mut command = std::process::Command::new(shell);

`user_shell()` is `std::env::var("SHELL")` with an empty-filter and a `bash`
fallback (`src/session/environment.rs:46-50`). These tests are not stubbed: they
reach `.output()` on a real subprocess. So a hook test racing a writer spawns
`/usr/bin/nu` on a host without `nu` and fails as `Failed to execute hook: <cmd>`
(`repo_config.rs:1160`, `:1272`) or, for the `run_before_*` tests, as
`Failed to execute <label> hook: <cmd>` (`repo_config.rs:1505`). Either way it
names the wrong cause, which is the reason to close it.

**The writers.** Twelve `SHELL` guard sites across three files:

    src/session/environment.rs:1923   /bin/zsh
    src/session/environment.rs:1930   (unset)
    src/session/environment.rs:1937   "  "
    src/session/environment.rs:1944   /bin/zsh
    src/session/environment.rs:1951   /usr/bin/fish
    src/session/environment.rs:1958   /usr/bin/nu
    src/session/instance.rs:10661     loop: /bin/bash, /bin/zsh, /usr/bin/fish, /usr/bin/nu
    src/session/instance.rs:10675     /bin/zsh
    src/session/instance.rs:10686     /usr/bin/fish
    src/session/instance.rs:10699     /usr/bin/nu
    src/session/instance.rs:10725     the resolved `bash` (added by #3421)
    src/tmux/terminal_session.rs:853  /bin/sh, in a multi-pair guard

Eight of them export a shell path that may not exist on the host; that is the
subset that can break a reader. `environment.rs:1930` and `:1937` are harmless
against *this* hazard, since `user_shell()` filters both to the `bash` fallback,
though that fallback is a bare name resolved through `PATH`, so it is not immune
to the `PATH` hazard in the companion issue.

As you noted, these are microsecond string builders rather than process spawns,
so the window is narrow and the current state is strictly better than before
#3421.

**One thing to fix while here.** The comment added at
`src/session/instance.rs:10721-10724` says `ENV_LOCK` "is taken only by
`EnvGuard` and 12 of the 14 `repo_config` hook tests take none". I cannot
reconcile those totals with the file: it has 69 `#[test]` fns and exactly the two
`EnvGuard` uses above, and I count more than 14 that reach a local hook spawn.
The comment's conclusion is right; its numbers do not appear to be. Worth
correcting in the same patch rather than leaving a count that will be quoted.

**Suggested direction**, yours: have the hook tests pin `SHELL` themselves rather
than trusting the ambient value. Local to the readers, and it does not require
touching the writers.

Happy to send the patch.

blueagledev:
## Description

Closes #3449.

`build_hook_command`'s `HookTarget::Local` arm runs `user_shell()`, which reads
the process-global `SHELL`, and spawns it. Sixteen `repo_config` tests reach a
hook runner and fourteen of them took no guard, while twelve `EnvGuard` sites
across `session::environment`, `session::instance` and `tmux::terminal_session`
touch `SHELL` to exercise the launch wrapper (eleven exporting one, eight of
those a path that need not exist on the host). Those writers hold
`test_support::ENV_LOCK`; the readers took nothing, so a concurrent one could
be read here and fail the spawn as a hook error naming the wrong cause.

Your #3421 approval counted nine of those, which was right for the
`set_var("SHELL", ...)` sites `d84c6349` converted; eight is the same set after
the conversion, minus the one that sets whitespace.

The readers now pin it themselves, which is the direction you suggested on
#3421 and is local to them: it needs no change to the writers. The pin goes
through `EnvGuard`, so it also takes `ENV_LOCK` and closes the window from the
writer side as well.

The effect is easiest to see with a real shell rather than a contrived path.
With fish as `$SHELL`, which is a supported login shell and cannot parse the
POSIX hook bodies these tests use:

```
env SHELL=$(command -v fish) cargo test --lib -- session::repo_config
  before (24226821):  64 passed; 5 failed
  after  (bc3d06cc):  69 passed; 0 failed
```

and with a path that does not resolve at all, which is the shape a mid-read
write produces:

```
env SHELL=/nonexistent/nu cargo test --lib -- session::repo_config
  before (24226821):  55 passed; 14 failed
  after  (bc3d06cc):  69 passed;  0 failed
```

(The fourteen that fail there and the fourteen that took no guard are
overlapping but different sets; twelve are in both.)

It costs the sixteen a place in the `ENV_LOCK` queue: the module goes from
about 0.02s to about 0.09s. I judged that worth it, but it is a real cost and
the sibling PR argues partly from serialisation cost, so it should be stated.

Two notes on the shape:

- `sh` is resolved through `which` rather than hardcoded. `/bin/sh` was not
  one of the paths #3421 removed and does exist on the host that motivated it,
  so this is not undoing that work, just not adding a new assumption.
- `pin_host_shell` is `#[must_use]`. Wrapping the guard in an `Option` would
  otherwise discard `EnvGuard`'s own annotation, and a bare call would compile
  clean while pinning nothing.

While here it corrects the comment I added in #3421 at `instance.rs:10721`. It
counted these as "12 of the 14" (it is 14 of 16) and rested on them reading the
override, which after this is no longer true. Per your nit on `67e28071` about
counts that rot, the replacement states the invariant and cites #3449 rather
than carrying numbers.

- **Hardcode `/bin/sh`.** Works, and would not have failed on the host that
  motivated #3421, but it adds an assumption that PR was written to remove.
  Resolving costs one line.
- **Convert the writers instead of the readers.** A wider diff across three
  modules for the same property, and your suggestion was reader-local.
- **Have the readers take a bare `EnvGuard` without pinning**, as the two
  `GIT_TERMINAL_PROMPT` tests already do. This is the one worth recording,
  because it looks sufficient and is not: a guard that does not name `SHELL`
  still leaves the ambient value in play. Both of those tests are in the five
  that fail under fish on `main`. Pinning fixes the race and the ambient value
  together, which is why the fish column moves at all.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

This is a change to test hygiene, so the measurement above is the evidence
rather than a new test: the same suite under a hostile `SHELL` goes from 14
failures to 0. `cargo fmt --check` clean, `cargo clippy --lib --tests` at the
same 6 warnings as `main`, `session::repo_config` 69/69.

The lib suite does not go green on my machine either way (rotating `tmux::*`
failures plus one `$HOME`-dependent dialog test, all present on unpatched
`main` here), so I am comparing against `main` rather than claiming a green
suite.

## AI Usage

- [ ] No AI was used
- [x] AI was used

write-up were done by Claude working from that. The hostile-`SHELL` framing came
out of an adversarial review round rather than the original patch, which argued
from the race alone and was weaker for it.

## Summary

- Updated `repo_config` hook tests to pin `SHELL` with `EnvGuard`.
- Added a clear diagnostic when `sh` is not available on `PATH`.
- Corrected the `session::instance` test comment.

## Benefits

- Prevents concurrent environment changes from causing misleading hook failures.
- Improves test reliability under hostile or changing `SHELL` values.
- Keeps shell setup local to the tests that depend on it.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -10719,9 +10719,9 @@ mod tests {
             return;
         };
         // The guard restores on unwind; the resolved path matters separately,
-        // because `test_support::ENV_LOCK` is taken only by `EnvGuard` and 12
-        // of the 14 `repo_config` hook tests take none, so they read this
-        // override regardless and must not read a path that cannot run.
+        // because `wrap_command_ignore_suspend` execs `$SHELL` below. The
+        // `repo_config` hook tests used to read this override too and now pin
+        // their own (#3449).
         let _shell = EnvGuard::set(&[("SHELL", &bash)]);
         let temp = tempfile::tempdir().unwrap();
         let working_dir = temp.path().join("some project's dir");
--- a/src/session/repo_config.rs
+++ b/src/session/repo_config.rs
@@ -1717,7 +1717,10 @@ mod tests {
     /// `ENV_LOCK`, which closes the window from the writer side. See #3449.
     #[must_use = "bind it to `_shell`; dropped immediately, it unpins SHELL again"]
     fn pin_host_shell() -> Option<crate::session::test_support::EnvGuard> {
-        let sh = which::which("sh").ok()?;
+        let Ok(sh) = which::which("sh") else {
+            eprintln!("not pinning SHELL: sh not found on PATH");
+            return None;
+        };
         Some(crate::session::test_support::EnvGuard::set(&[(
             "SHELL", &sh,
         )]))
--- a/src/session/repo_config.rs
+++ b/src/session/repo_config.rs
@@ -1709,6 +1709,20 @@ mod tests {
     use super::*;
     use tracing_test::traced_test;
 
+    /// Pin `SHELL` for a test that spawns a local hook. `build_hook_command`'s
+    /// local arm runs `user_shell()`, so an ambient `SHELL` this host cannot
+    /// run, or one another test is mid-way through exporting, surfaces here as
+    /// a hook failure naming the wrong cause. Resolved rather than hardcoded so
+    /// the FHS assumption #3421 removed stays removed; the guard also takes
+    /// `ENV_LOCK`, which closes the window from the writer side. See #3449.
+    #[must_use = "bind it to `_shell`; dropped immediately, it unpins SHELL again"]
+    fn pin_host_shell() -> Option<crate::session::test_support::EnvGuard> {
+        let sh = which::which("sh").ok()?;
+        Some(crate::session::test_support::EnvGuard::set(&[(
+            "SHELL", &sh,
+        )]))
+    }
+
     #[test]
     fn test_hooks_config_empty() {
         let hooks = HooksConfig::default();
@@ -1793,7 +1807,8 @@ mod tests {
 
     #[test]
     fn test_run_before_start_hooks_collects_kv() {
-        // Real host shell; multiple commands, later command's keys appended.
+        let _shell = pin_host_shell();
+        // Multiple commands; later command's keys appended.
         let tmp = tempfile::tempdir().unwrap();
         let cmds = vec![
             "echo GH_TOKEN=ghs_abc".to_string(),
@@ -1812,6 +1827,7 @@ mod tests {
 
     #[test]
     fn test_run_before_start_hooks_reads_session_env() {
+        let _shell = pin_host_shell();
         // A per-session value reaches the hook and can scope what it mints.
         let tmp = tempfile::tempdir().unwrap();
         let cmds = vec!["echo \"GH_TOKEN=tok-$TEST_VAR\"".to_string()];
@@ -1826,6 +1842,7 @@ mod tests {
 
     #[test]
     fn test_run_before_start_hooks_hard_fail() {
+        let _shell = pin_host_shell();
         let tmp = tempfile::tempdir().unwrap();
         let cmds = vec!["exit 3".to_string()];
         let err = run_before_start_hooks(&cmds, tmp.path(), &[], &[])
@@ -1835,6 +1852,7 @@ mod tests {
 
     #[test]
     fn test_run_before_start_hooks_error_omits_stdout_secret() {
+        let _shell = pin_host_shell();
         // A failing hook may have already printed secret KEY=VALUE lines; those
         // must never appear in the error (which can be logged/displayed). The
         // secret is supplied via env so it lives only in the hook's stdout, not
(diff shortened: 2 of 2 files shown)


njbrake (approved):
Approving. I re-derived every count and reproduced both measurements rather than taking the description on trust, and they all hold.

Counts: 16 tests reach a local hook spawn (19 name a runner, but two are container-target only and `lifecycle_env_vars_shape` matches on its doc comment), and 14 took no guard. Writers: 12 `EnvGuard` SHELL sites, 11 exporting one, 8 of those a path that need not exist. The "twelve in both" overlap is right too.

Reproductions, run on `main` and on the branch: fish 64/5 to 69/0, and `/nonexistent/nu` 55/14 to 69/0. Both match to the test name.

The part I most wanted to check was whether pinning inside the two tests that already hold an `EnvGuard` could self-deadlock on the non-reentrant `ENV_LOCK`. It cannot. `acquire_env_lock` gates on the `ENV_LOCK_HELD` thread-local, so a nested guard acquires nothing, and reverse drop order restores SHELL before the outer guard releases. Against a writer on another thread it is contention rather than deadlock, since `EnvGuard` only ever takes one lock.

Two things worth recording. `sh` is dash on the CI image, so the pin gets exercised under a stricter shell than most contributors run locally, which is the right direction. And your `#[must_use]` justification is correct in a way I doubted at first: `Option` is not `must_use` at the type level, so a bare call really would have compiled clean without it.

I pushed one nit on top (`faaea2d1`). The helper returned `None` silently when `which("sh")` failed, so a host without `sh` would have run unpinned with no signal. It now says so, matching the loud-skip shape you used for the same lookup at `instance.rs:10717`.

Also confirmed this merges cleanly with #3452 and that the two compose rather than fight, since neither relies on serial keys for SHELL exclusion.

_Note: this review was carried out and drafted by Claude Opus 5 at @njbrake's direction via the Claude Code CLI. The verdict to approve and merge is his._

PR #3452, opened 2026-08-20
test(session): move the launch-path test onto the default serial key

### LINKED ISSUE #3450: test: launch-path test should use the default serial key, not `serial(shell_env)`
Follow-up from #3421, filed at @njbrake's request.

`test_wrap_command_reasserts_working_dir_after_login_shell`
(`src/session/instance.rs:10715`, attribute at `:10714`) carries
`#[serial_test::serial(shell_env)]`. A named serial key only excludes other
holders of that key, and the eleven process-global `PATH` mutators in
`update::install`, `acp::node` and `acp::acp_client` all carry the **default**
key, so `shell_env` buys this test no exclusion against the thing that can
actually break it.

**The window is the whole test, not just the prologue.** `which::which("bash")`
at `:10717` is the read that now skips rather than fails. But `:10740` then does

    std::process::Command::new("bash").args(["-c", &wrapped])

which resolves `bash` through `PATH` again, inside the guard, where `ENV_LOCK`
gives no protection against a raw `set_var` scrubber. So depending on where a
race lands, the outcome is either the silent skip you identified or a spurious
hard failure at the spawn. Both argue the same way.

Worth pairing with a one-token fix: `:10740` can use the already-resolved
`Command::new(&bash)`, which closes the second window outright.

**Why leaving `shell_env` costs nothing.** The key has 49 holders, so the
question is fair. The answer is that the exclusion that matters is `ENV_LOCK`,
not the key: this test takes an `EnvGuard` itself at `:10725`, and 48 of the 49
holders also take one, so both sides serialise on `ENV_LOCK` regardless of which
serial key they carry.

- The 30 holders across `v015_rewrite_hook_strings.rs`,
  `v017_rewrite_hook_strings_for_per_user_base.rs` and
  `v018_strip_codex_config_toml_hooks.rs` all call `unset_agent_home_env()`, an
  `EnvGuard::unset` over `CODEX_HOME`, `CLAUDE_CONFIG_DIR`, `CURSOR_CONFIG_DIR`,
  `GEMINI_CONFIG_DIR` and `QWEN_CONFIG_DIR`. No `PATH`, no `SHELL`.
- Six of the seven in `src/hooks/mod.rs` are `CLAUDE_CONFIG_DIR` guards
  (`:2207-2278`); the seventh (`:5152`) guards `CODEX_HOME`, `HOME` and
  `XDG_CONFIG_HOME`.
- The one exception is
  `test_validate_env_entries_skips_default_terminal_vars_when_unset`
  (`src/session/environment.rs:1717`), which hand-rolls snapshot/restore with raw
  `remove_var`/`set_var` and takes no guard. It touches only `TERM`,
  `COLORTERM`, `FORCE_COLOR` and `NO_COLOR`, none of which this test reads or
  passes to the child, so it is not a hazard here, but it is a genuine
  counterexample to "the `shell_env` group is all guarded", and it relies on the
  key rather than on a lock.

**The cost, stated plainly.** The default key has ~1341 members, so this test's
runtime joins a long serial chain rather than the group of eleven the framing
might suggest. It spawns a process, so that is not free.

The tree already reaches this conclusion for the two tests #3421 put on the
default key: `omp_capture_gate_executes_nested_stdin_scripts`
(`src/session/instance.rs:11871-11877`) and
`test_container_env_file_does_not_mutate_host_process_environment`
(`src/tmux/session.rs:4037-4041`). The sharpest line is `instance.rs:11877`:

> A future PATH mutator outside the default serial group would reopen it.

**Suggested direction**, yours: move this test onto the default key, joining the
group the scrubbers already serialise against.

Happy to send the patch.

blueagledev:
## Description

Closes #3450.

`test_wrap_command_reasserts_working_dir_after_login_shell` carried
`#[serial(shell_env)]`. A named key only excludes other holders of that key,
and every test that mutates `PATH` process-globally carries the default one, so
`shell_env` bought this test no exclusion against them. It resolves `bash`
through the inherited `PATH` at `which::which("bash")`, and since #3421 that
skips with a message instead of hard-failing, so a scrub racing it turns a
failure into a silent skip. That is what makes this worth changing rather than
merely inconsistent.

It also spawned `bash` a second time by bare name, which the same scrub would
break. That now reuses the already-resolved path, which is the one-token half
of #3450 and closes the second window outright.

What leaving the group costs, stated plainly rather than waved off. Of the 49
`shell_env` holders, 48 take an `EnvGuard`, which holds `test_support::ENV_LOCK`
whatever the serial key says, and so does this test, so those stay excluded. The
49th takes no guard at all,
`test_validate_env_entries_skips_default_terminal_vars_when_unset`, and this
does open a window against it. It is narrow and I think acceptable:
`std::process::Command` snapshots the environment at spawn, so the exposure is
that instant rather than the child's lifetime, and that test only removes and
restores `TERM`, `COLORTERM`, `FORCE_COLOR` and `NO_COLOR`. The residual is a
profile file that both writes to stdout and gates on one of those, since the
inner shell is a login shell.

In the other direction it joins the default key's roughly 1350 members, which is
a real cost for a test that spawns a process.

- **Move all 49 `shell_env` holders onto the default key.** It buys them
  nothing: 48 already hold `ENV_LOCK` through an `EnvGuard`, whatever the key
  says. It would put 49 more tests into the roughly 1350-member default chain
  for a property they already have.
- **Rely on the `EnvGuard` this test already holds, and drop the serial key.**
  The tempting one, since the guard is right there and it does have to set
  `SHELL` regardless. It does not cover `PATH`, and `instance.rs:11871` already
  says why: "none of them takes `test_support::ENV_LOCK`, so a guard would
  exclude unrelated guard users and leave this window open."

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

No new test: the change is one attribute and one argument on an existing test,
and the property it buys (exclusion against a process-global `PATH` mutation) is
not observable from inside the suite without staging the race itself. The
rationale is in a comment on the test, matching the pattern at
`instance.rs:11871` and `tmux/session.rs:4037` that #3421 established.

as `main`, the moved test passing. As on the sibling PR, the lib suite does not
go green on my machine either way, so this is a comparison against `main`
rather than a claim of a green suite.

## AI Usage

- [ ] No AI was used
- [x] AI was used

write-up were done by Claude working from that. An adversarial review round
caught two false statements in the first draft of this description, including a
claim that every `shell_env` holder takes an `EnvGuard`, which is not true.

## Summary

- Moved the working-directory wrapper test to the default serial group.
- Reused the resolved `bash` path for shell execution.
- Documented the concurrency trade-offs of the change.

## Benefits

The test now avoids conflicts with tests that modify the process-wide `PATH`. This prevents silent skips and shell launch failures caused by concurrent environment changes.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -10710,6 +10710,7 @@ mod tests {
     /// the agent in the wrong directory. The wrapper must re-assert
     /// `working_dir` inside the login shell's own script, after profile
     /// sourcing, so it wins regardless of what those files did.
+    ///
     /// `#[serial]` on the default key, not `shell_env`: this resolves `bash`
     /// through the inherited `PATH`, and every test that mutates `PATH`
     /// process-globally (`update::install`, `acp::node`, `acp::acp_client`)
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -10710,8 +10710,16 @@ mod tests {
     /// the agent in the wrong directory. The wrapper must re-assert
     /// `working_dir` inside the login shell's own script, after profile
     /// sourcing, so it wins regardless of what those files did.
+    /// `#[serial]` on the default key, not `shell_env`: this resolves `bash`
+    /// through the inherited `PATH`, and every test that mutates `PATH`
+    /// process-globally (`update::install`, `acp::node`, `acp::acp_client`)
+    /// carries the default key, so `shell_env` bought no exclusion against
+    /// them. Since #3421 a scrub racing the `which` is a silent skip rather
+    /// than a failure. The `shell_env` holder this stops excluding touches
+    /// only `TERM`/`COLORTERM`/`FORCE_COLOR`/`NO_COLOR`, and `Command`
+    /// snapshots the environment at spawn, so the exposure is that instant.
     #[test]
-    #[serial_test::serial(shell_env)]
+    #[serial_test::serial]
     fn test_wrap_command_reasserts_working_dir_after_login_shell() {
         // The wrapper execs `$SHELL`, so it has to be a shell that exists here.
         let Ok(bash) = which::which("bash") else {
@@ -10737,7 +10745,7 @@ mod tests {
             wrapped.contains("|| exit 1\nstty susp undef"),
             "the cd must exit-on-failure before disabling suspend: {wrapped}",
         );
-        let output = std::process::Command::new("bash")
+        let output = std::process::Command::new(&bash)
             .args(["-c", &wrapped])
             .output()
             .unwrap();


njbrake (approved):
Verified the load-bearing claim against `serial_test` 3.5.0 rather than taking it on trust. `code_lock.rs` keys locks in a `HashMap<String, UniqueReentrantMutex>`, and the derive pushes an empty string when the attribute has no argument, so a named key and the default key are unrelated mutexes. `shell_env` bought this test nothing against the 11 process-global `PATH` mutators in `update::install`, `acp::node` and `acp::acp_client`, all of which carry the default key and none of which takes an `EnvGuard`.

Reproduced both failure modes. With `PATH` scrubbed the test reports `ok` in 0.00s instead of 0.75s, and the skip line is invisible under libtest's default capture. Feeding the script `wrap_command_ignore_suspend` emits to a login bash with no usable `PATH` gives `exec: env: not found` and exit 127. I also re-derived the 49/48/1 split and confirmed the single unguarded holdout is the one you name.

Two notes on the framing rather than the change.

The AGENTS.md tension is smaller than the description suggests. Lines 150 to 152 sit inside the E2E Tests section, so they govern `tests/e2e`, a separate binary with its own lock map, and `#[parallel]` has zero occurrences under `src/`. The guarantee that reservation protects has no counterpart in the binary you changed, and line 152's closing sentence is your own argument.

There is no third option that removes the `PATH` dependency. `exec env {cmd}` is emitted by the function under test, so exclusion is the only remedy available. That makes this change necessary rather than merely tidier. The option missing from your list is the structural one: move those 11 scrubbers onto `EnvGuard` so `ENV_LOCK` becomes the mechanism, which is what `test_support.rs` argues for and would retire the caveat at `instance.rs:11877`. Worth a follow-up issue.

One optional tightening: replacing the holdout's hand-rolled snapshot with `EnvGuard::unset(DEFAULT_TERMINAL_ENV_VARS)` closes the window this opens and removes the need for the comment's last two lines. `EnvGuard` is already imported in that module.

I pushed one line to your branch adding the missing blank `///` separator between the two rationales.

Thanks for the care in the write-up, and for correcting your own earlier draft.

PR #3453, opened 2026-08-20
docs(session): archive() is not cleared by the attention signal

blueagledev:
## Description

A doc-comment correction on `Instance::archive()`. It currently ends:

> Auto-cleared by the attention-signal hook on Waiting/Error.

Nothing does that, and the hook does the inverse: `is_urgent` returns false
while a session is archived, so archive suppresses the attention signal rather
than the signal clearing archive. Both were added in the same commit (#1084),
and `git log -G "archived_at = None"` shows no clearer has since been removed.

What holds the property is that no status path clears the flag, rather than any
guard doing so. The replacement says that, names the wake paths, and names the
one edge that is easy to miss: `merge_user_action_diff` mirrors a wake onto disk
on any advance of `last_accessed_at` whoever wrote it, and `apply_status_intent`
advances that on a status transition without checking `is_archived`.

It deliberately does not enumerate every clearer. I tried that and got it wrong
twice, most recently by omitting the arm the TUI's own unarchive reaches disk
through, and #3421's review already nitted a count off a comment for the same
reason.

Doc only; no behaviour change.

### One wording worth recording

"Cleared only by `unarchive` and `touch_last_accessed`" and "never cleared by a
status change" are both natural things to write here and both are wrong, in the
same direction as the line being replaced. The second is the subtle one: a
status transition on the row cannot clear the flag, but `apply_status_intent`
stamps `last_accessed_at` on a transition without an archived guard, and
`merge_user_action_diff` mirrors any advance of that field.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

No test: the diff is a doc comment, and it now describes an absence (no status
path clears the flag) rather than a mechanism, so there is nothing new to pin.
`test_attention_tier_archived_returns_99` pins `unarchive` and only that, which
is why I stopped citing it as the guarantee.

as `main`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

The direction and decisions are mine; the investigation and this write-up were
done by Claude working from that. Relevant here because it is a doc PR whose
whole value is one sentence's accuracy: earlier drafts of that sentence were
wrong twice, each time by claiming a completeness the code does not support, so
the version above deliberately describes an absence and enumerates nothing.

### Summary
- Updated `Instance::archive()` documentation to reflect actual archive behavior.
- Clarified that attention signals do not clear the archive flag.
- Documented relevant archive-clearing paths without claiming the list is complete.
- Kept the change documentation-only.

### Benefits
- Improves documentation accuracy.
- Reduces confusion about archived session handling.
- Introduces no behavior changes.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -2238,8 +2238,15 @@ impl Instance {
     }
 
     /// Mark the session archived. Archived sessions sink to the bottom of
-    /// the Attention sort and render in italic+dim style, but remain
-    /// visible. Auto-cleared by the attention-signal hook on Waiting/Error.
+    /// the Attention sort and render in italic+dim style, but remain visible.
+    /// Archive suppresses the attention signal rather than the signal
+    /// clearing archive: `is_urgent` returns false while archived, and the
+    /// attention sort short-circuits the row to its bottom tier.
+    ///
+    /// Cleared by `unarchive`, by `touch_last_accessed`, and by `favorite`
+    /// and `pin` through the mutual exclusion below; not by `snooze`.
+    /// `merge_user_action_diff` mirrors those onto disk, and #3465 tracks a
+    /// status transition reaching that mirror without a user gesture.
     ///
     /// Mutual exclusion with `favorite`, `snooze`, and `pin`: archiving
     /// clears `favorited_at`, `snoozed_until`, and `pinned_at`. Archive


njbrake (changes requested):
The correction is right and worth landing. Three things in the replacement text need fixing first.

**`snooze` is not a clearer.** "the mutual-exclusion clears below" points at the paragraph describing archive clearing `favorited_at`/`snoozed_until`/`pinned_at`, which is the opposite direction. The clears *of* `archived_at` live on `favorite()` and `pin()`. `snooze()` sets `snoozed_until` and `pinned_at` only (`instance.rs:2477`), so a reader who takes that pairing as symmetric concludes that snoozing unarchives. That is a new wrong inference, in the same direction as the line being replaced.

**The named stamp is the less reachable of the two.** `update_status_with_metadata` (`instance.rs:6699`) makes the identical unguarded stamp for tmux rows, and unlike `apply_status_intent` it is durably persisted by the daemon's passive writer. Structured rows' `last_accessed_at` is deliberately never persisted by the daemon (the durability contract on `apply_acp_overlay_inplace`) and reaches disk only via the TUI's own fold-and-persist, which keeps memory and disk in lockstep, so `touched` stays false there. The edge that can actually bite is the one not named.

**This reads as contract but it is a bug.** Reproduced locally: with disk's `last_accessed_at` advanced by a status transition and a stale `pre`, `merge_user_action_diff` splices the user's `archived_at` in at 2139 and then the `touched` arm at 2223 wipes it. The TUI archive key is silently undone on disk and the row comes back on the next reload. That arm's own comment assumes an advance of `last_accessed_at` means a user interaction; both status stamps break the premise. Worth an issue and an `is_archived` guard rather than a sentence in the doc.

Suggested shape: keep the first half (`is_urgent` returns false while archived, `attention_tier` short-circuits the row to 99), drop the mechanism narration, and say that `snooze` is not among the clearers.

Separately, CodeRabbit's title nit is a false positive. `docs(session):` matches the house convention.

njbrake (approved):
All three points from my earlier review are addressed, and I retitled this: the old title re-asserted the "no status signal clears it" claim you removed from the body in the last force-push, and it would have landed as the squash commit message. It now reads `archive() is not cleared by the attention signal`.

Re-verified the replacement text against the tree. {`unarchive`, `touch_last_accessed`, `favorite`, `pin`} is the complete set of mutators that clear `archived_at`; `snooze()` (`instance.rs:2478`) touches `snoozed_until` and `pinned_at` only; `is_urgent` (`2450`) and the tier short-circuit (`groups.rs:623`) support the suppression claim. Pointing at #3465 instead of describing the merge path is the right call.

Your correction on the fix location is right and worth keeping: an `is_archived` guard on the `touched` arm fails `test_merge_diff_peer_touch_clears_tui_archive` (`9283-9300`), which asserts the current behavior on purpose.

Not blocking, for whenever you are next in here: "through the mutual exclusion below" points at a paragraph documenting only the archive-clears-others direction, so a reader following the pointer will not find the clears of `archived_at` there. The explicit "not by `snooze`" already prevents the wrong inference.

PR #3468, opened 2026-08-21
feat(web): persistent structured-view sessions (#3466)

### LINKED ISSUE #3466: [Feature] Persistent structured view sessions for instant chat switching
## Problem

When a user with many structured-view chats jumps between them in the web dashboard, every chat switch shows a "Starting structured view…" banner and takes a few seconds before the transcript is interactive. The backend idle-stop setting is already disabled, so the delay is not the worker being respawned; it is the frontend tearing down and remounting the entire `<StructuredView>` subtree on every switch.

## Evidence

- `<StructuredView>` is mounted with `key={activeSessionId}` in both places:
  - `web/src/App.tsx:1885`
  - `web/src/components/MobileMainPane.tsx:114`
- This causes React to unmount the previous chat surface and mount a fresh one, which opens a new WebSocket and replays state.
- The banner text is rendered at `web/src/components/acp/StructuredView.tsx:1817` while `status === "connecting"`.
- The connection/setup path is in `useAcpSession` (`web/src/hooks/useAcpSession.ts:948`), which on mount calls `fetchReplay` and then opens a WebSocket to `/sessions/{id}/acp/ws`.
- The user's local `config.toml` already has `[acp].auto_stop_idle_secs = 0` and `[session].auto_stop_idle_secs = 0`, confirming the backend is not the bottleneck.

## Artifacts & Reproduction

1. Open the web dashboard with `acp.auto_stop_idle_secs = 0`.
2. Open two or more structured-view sessions.
3. Click back and forth in the sidebar.
4. Observe the "Starting structured view…" / "Reconnecting to structured view…" banner and the ~1–5s wait before the chat is fully live.

## How It Works (moving parts)

- **Per-session state cache.** `useAcpSession` stores hydrated ACP state in a module-level `stateCache: Map<string, AcpState>` and publishes updates via `stateListeners: Map<string, Set<() => void>>` (`web/src/hooks/useAcpSession.ts:163-169`). This already supports multiple concurrent sessions.
- **Component isolation.** Each `<StructuredView>` wraps its own `AcpSessionContext`, `AgentProfileProvider`, `ToolDisplayModeProvider`, `BackgroundAgentsContext`, and `<AcpRuntime>` → `<AssistantRuntimeProvider>` tree (`web/src/components/acp/StructuredView.tsx:161-224`, `web/src/components/acp/AcpRuntime.tsx:132-170`). So multiple instances can coexist without sharing React state.
- **Precedent.** `web/src/components/TerminalSessionStack.tsx` already implements the desired pattern: keep N recent sessions mounted but invisible (`inert`, `aria-hidden`, `pointer-events-none`) so switching back is instant. It is gated by the client-side `persistentTerminals` / `maxPersistentTerminals` settings.
- **Lifecycle props.** `StructuredView` receives `acpWorkerState`, `archivedAt`, `snoozedUntil`, `trashedAt`, `fileRefSession`, etc. per session. A stack would need to pass each mounted session its own values from the `sessions` array.
- **Edge cases that become active concerns with a stack:**
  - Mobile keyboard height reservation in `StructuredViewRoot` (`web/src/components/acp/StructuredView.tsx:270-286`) must apply only to the active instance.
  - `AttentionChime`, approval modals, and keyboard shortcuts must not fire for inactive sessions.
  - Scroll restoration (`web/src/lib/acpScrollState.ts`) must still work when a session becomes active again.
  - Composer drafts are already keyed by `sessionId` (`web/src/lib/acpDrafts.ts`).
  - Visibility/online reconnect effects in `useAcpSession` (`web/src/hooks/useAcpSession.ts:1143-1204`) will fire for every mounted instance simultaneously when the page wakes.

## Definition of Done

- A design is accepted for keeping a configurable number of recent structured-view sessions mounted off-screen, default-off, similar to the existing terminal persistence feature.
- The feature must not regress focus, accessibility, mobile keyboard handling, scroll restoration, or approval/elicitation flows.
- There is a clear cap/limit setting to bound memory and WebSocket usage.

## Watch Out For

- Assistant-ui runtime and DOM overhead scale with the number of mounted sessions; a low default limit is important.
- All N mounted WebSockets will reconnect together after sleep/offline/visibility changes, creating a reconnect burst.
- Inactive sessions still receive WebSocket heartbeats and any live transcript events, so they are not truly idle.
- File viewer overlays and plugin panes must remain scoped to the active session.
- The change touches both desktop (`App.tsx`) and mobile (`MobileMainPane.tsx`) entry points.

karljuhlthefool:
## Description

Adds an opt-in, default-off dashboard setting that keeps the N most recently viewed structured-view sessions mounted (but inert and idled) so switching back to a recent structured chat is instant and avoids the "Starting structured view…" reconnect banner.

- New web-settings fields: `persistentStructuredViews` (default false) and `maxPersistentStructuredViews` (default 2, clamped 1–5).
- New `StructuredViewStack` component renders the active session plus recent inactive structured sessions behind an `inert` / `aria-hidden` / `pointer-events-none` layer.
- Threads `active: boolean` through `StructuredView` → `AcpChrome` → `Composer`, `AttentionChime`, and `StructuredViewRoot` to gate autofocus, audio, mobile keyboard reservation, auto-scroll, scroll observers, and reconnect behavior.
- Inactive `useAcpSession` instances skip `localStorage` cache writes and defer aggressive wake/online reconnects.
- Saves/restores scroll position across active ↔ inactive transitions.
- Wires the stack into `App.tsx` (desktop) and `MobileMainPane.tsx` (mobile).

Closes #3466

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: the new behavior is covered by unit tests for `StructuredViewStack` and the settings component. A live Playwright story would need to assert across multiple WebSocket reconnects, which is better validated by the existing structured-view suite once the feature is enabled.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Kimi Code CLI / Claude Code assisted with implementation planning and code changes.

Implementation followed the plan in `.plan-persistent-structured-view.md` and the two-iteration generic review loop the user requested.

## Summary

- Added an opt-in setting to keep recent structured-view sessions mounted.
- Added a configurable limit of one to five sessions, with a default of two.
- Added `StructuredViewStack` for desktop and mobile session handling.
- Preserved session state and scroll positions during view changes.
- Disabled inactive-session focus, keyboard handling, audio alerts, auto-scroll, cache writes, and reconnect retries.
- Prevented inactive plugin drafts from moving focus to the composer.
- Added session filtering, terminal-session pruning, accessibility labels, and conditional loading.
- Added regression tests for persistence, limits, visibility, reconnects, keyboard behavior, draft focus, and attention chimes.
- The full web unit suite passes with 3,705 tests.

## Benefits

- Recent chats reopen without the “Starting structured view…” delay.
- Inactive sessions do not interfere with the active session.
- The session limit helps control memory and WebSocket usage.

## Further enhancement

- Monitor resource usage with higher session limits to validate the default cap in typical workloads.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1877,21 +1877,25 @@ function AppContent({
             left={
               <div className="flex-1 flex flex-col min-h-0 overflow-hidden relative">
                 <div className={selectedFilePath ? "hidden" : "relative flex-1 flex flex-col min-h-0 overflow-hidden"}>
-                  <Suspense fallback={<AcpLoadingFallback />}>
-                    <StructuredViewStack
-                      activeSessionId={activeSessionId}
-                      sessions={sessions}
-                      persistent={webSettings.persistentStructuredViews}
-                      maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
-                      visible={activeSession?.view === "structured"}
-                      onOpenFileRef={handleOpenFileRef}
-                      onOpenAgentsPane={openAgentsPane}
-                      onRestoreSession={async (sessionId) => {
-                        const restored = await handleRestoreSession(trashedWorkspaceRestoreIds(workspaces, sessionId));
-                        return restored;
-                      }}
-                    />
-                  </Suspense>
+                  {(activeSession?.view === "structured" || webSettings.persistentStructuredViews) && (
+                    <Suspense fallback={<AcpLoadingFallback />}>
+                      <StructuredViewStack
+                        activeSessionId={activeSessionId}
+                        sessions={sessions}
+                        persistent={webSettings.persistentStructuredViews}
+                        maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
+                        visible={activeSession?.view === "structured"}
+                        onOpenFileRef={handleOpenFileRef}
+                        onOpenAgentsPane={openAgentsPane}
+                        onRestoreSession={async (sessionId) => {
+                          const restored = await handleRestoreSession(
+                            trashedWorkspaceRestoreIds(workspaces, sessionId),
+                          );
+                          return restored;
+                        }}
+                      />
+                    </Suspense>
+                  )}
                   {activeSession?.view !== "structured" && (
                     <TerminalSessionStack
                       activeSessionId={activeSessionId!}
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -114,10 +114,8 @@ import { TerminalSessionStack } from "./components/TerminalSessionStack";
 // dependency tree. Cuts ~hundreds of KB off the cold-start bundle
 // for the (currently default) tmux-only flow. The Suspense fallback
 // below covers the brief load while the chunk arrives.
-const StructuredView = lazy(() =>
-  import("./components/acp/StructuredView").then((m) => ({
-    default: m.StructuredView,
-  })),
+const StructuredViewStack = lazy(() =>
+  import("./components/StructuredViewStack").then((m) => ({ default: m.StructuredViewStack })),
 );
 import { type PaneDisplay } from "./components/Dock";
 import { DockGroups, type DockGroupView } from "./components/DockGroups";
@@ -1878,31 +1876,23 @@ function AppContent({
             onToggleCollapse={toggleRightDock}
             left={
               <div className="flex-1 flex flex-col min-h-0 overflow-hidden relative">
-                <div className={selectedFilePath ? "hidden" : "flex-1 flex flex-col min-h-0 overflow-hidden"}>
-                  {activeSession?.view === "structured" ? (
-                    <Suspense fallback={<AcpLoadingFallback />}>
-                      <StructuredView
-                        key={activeSessionId}
-                        sessionId={activeSessionId!}
-                        acpWorkerState={activeSession.acp_worker_state ?? "absent"}
-                        tool={activeSession.tool}
-                        acpAgent={activeSession.acp_agent ?? null}
-                        clearAliases={activeSession.clear_aliases}
-                        archivedAt={activeSession.archived_at ?? null}
-                        snoozedUntil={activeSession.snoozed_until ?? null}
-                        trashedAt={activeSession.trashed_at ?? null}
-                        onRestore={
-                          activeSession.trashed_at
-                            ? () => handleRestoreSession(trashedWorkspaceRestoreIds(workspaces, activeSessionId!))
-                            : undefined
-                        }
-                        onOpenFileRef={handleOpenFileRef}
-                        fileRefSession={activeSession}
-                        onOpenAgentsPane={openAgentsPane}
-                        isSandboxed={activeSession.is_sandboxed}
-                      />
-                    </Suspense>
-                  ) : (
+                <div className={selectedFilePath ? "hidden" : "relative flex-1 flex flex-col min-h-0 overflow-hidden"}>
+                  <Suspense fallback={<AcpLoadingFallback />}>
+                    <StructuredViewStack
+                      activeSessionId={activeSessionId}
+                      sessions={sessions}
+                      persistent={webSettings.persistentStructuredViews}
+                      maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
+                      visible={activeSession?.view === "structured"}
+                      onOpenFileRef={handleOpenFileRef}
+                      onOpenAgentsPane={openAgentsPane}
+                      onRestoreSession={async (sessionId) => {
+                        const restored = await handleRestoreSession(trashedWorkspaceRestoreIds(workspaces, sessionId));
+                        return restored;
+                      }}
+                    />
+                  </Suspense>
+                  {activeSession?.view !== "structured" && (
                     <TerminalSessionStack
                       activeSessionId={activeSessionId!}
                       sessions={sessions.filter((session) => session.view !== "structured")}
--- a/web/src/components/MobileMainPane.tsx
+++ b/web/src/components/MobileMainPane.tsx
@@ -12,8 +12,11 @@ import { isPluginPaneId, type PluginPane } from "../lib/pluginPanes";
 import type { RepoBase, RichDiffFile, SessionResponse } from "../lib/types";
 import type { useDiffComments } from "../hooks/useDiffComments";
 import type { FileRef } from "../lib/fileRef";
+import type { WebSettings } from "../hooks/useWebSettings";
 
-const StructuredView = lazy(() => import("./acp/StructuredView").then((m) => ({ default: m.StructuredView })));
+const StructuredViewStack = lazy(() =>
+  import("./StructuredViewStack").then((m) => ({ default: m.StructuredViewStack })),
+);
 
 interface Props {
   view: RightPanelView;
@@ -23,7 +26,12 @@ interface Props {
   activeSession: SessionResponse | null;
   activeSessionId: string | null;
   sessions: SessionResponse[];
-  webSettings: { persistentTerminals: boolean; maxPersistentTerminals: number };
+  webSettings: Pick<
+    WebSettings,
+    "persistentTerminals" | "maxPersistentTerminals" | "persistentStructuredViews" | "maxPersistentStructuredViews"
+  >;
+  onOpenAgentsPane?: () => void;
+  onRestoreSession?: (sessionId: string) => Promise<boolean> | void;
   selectedFilePath: string | null;
   selectedRepoName: string | undefined;
   selectedFileLine: number | undefined;
@@ -66,6 +74,8 @@ export function MobileMainPane({
   activeSessionId,
   sessions,
   webSettings,
+  onOpenAgentsPane,
+  onRestoreSession,
   selectedFilePath,
   selectedRepoName,
   selectedFileLine,
@@ -108,24 +118,19 @@ export function MobileMainPane({
       )}
       <div className="relative flex-1 flex flex-col min-h-0 overflow-hidden">
         <div className={layerClass(view === "agent")} inert={view !== "agent"}>
-          {activeSession?.view === "structured" ? (
-            <Suspense fallback={null}>
-              <StructuredView
-                key={activeSessionId}
-                sessionId={activeSessionId!}
-                acpWorkerState={activeSession.acp_worker_state ?? "absent"}
-                tool={activeSession.tool}
-                acpAgent={activeSession.acp_agent ?? null}
-                clearAliases={activeSession.clear_aliases}
-                archivedAt={activeSession.archived_at ?? null}
-                snoozedUntil={activeSession.snoozed_until ?? null}
-                trashedAt={activeSession.trashed_at ?? null}
-                onOpenFileRef={onOpenFileRef}
-                fileRefSession={activeSession}
-                isSandboxed={activeSession.is_sandboxed}
-              />
-            </Suspense>
-          ) : (
+          <Suspense fallback={null}>
+            <StructuredViewStack
+              activeSessionId={activeSessionId}
+              sessions={sessions}
+              persistent={webSettings.persistentStructuredViews}
+              maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
+              visible={view === "agent"}
+              onOpenFileRef={onOpenFileRef}
+              onOpenAgentsPane={onOpenAgentsPane}
+              onRestoreSession={onRestoreSession}
+            />
+          </Suspense>
+          {activeSession?.view !== "structured" && (
             // Reserve the bottom home-indicator inset on this wrapper (the App
             // root no longer does; see index.css .safe-area-inset) so the last
             // terminal row clears it. Kept off the pane root itself, which owns
--- a/web/src/components/MobileMainPane.tsx
+++ b/web/src/components/MobileMainPane.tsx
@@ -118,18 +118,20 @@ export function MobileMainPane({
       )}
       <div className="relative flex-1 flex flex-col min-h-0 overflow-hidden">
         <div className={layerClass(view === "agent")} inert={view !== "agent"}>
-          <Suspense fallback={null}>
-            <StructuredViewStack
-              activeSessionId={activeSessionId}
-              sessions={sessions}
-              persistent={webSettings.persistentStructuredViews}
-              maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
-              visible={view === "agent"}
-              onOpenFileRef={onOpenFileRef}
-              onOpenAgentsPane={onOpenAgentsPane}
-              onRestoreSession={onRestoreSession}
-            />
-          </Suspense>
+          {(activeSession?.view === "structured" || webSettings.persistentStructuredViews) && (
+            <Suspense fallback={null}>
+              <StructuredViewStack
+                activeSessionId={activeSessionId}
+                sessions={sessions}
+                persistent={webSettings.persistentStructuredViews}
+                maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
+                visible={view === "agent"}
+                onOpenFileRef={onOpenFileRef}
+                onOpenAgentsPane={onOpenAgentsPane}
+                onRestoreSession={onRestoreSession}
+              />
+            </Suspense>
+          )}
           {activeSession?.view !== "structured" && (
             // Reserve the bottom home-indicator inset on this wrapper (the App
             // root no longer does; see index.css .safe-area-inset) so the last
--- a/web/src/components/StructuredViewStack.tsx
+++ b/web/src/components/StructuredViewStack.tsx
@@ -1,0 +1,115 @@
+import { useEffect, useMemo, useState } from "react";
+import type { SessionResponse } from "../lib/types";
+import type { FileRef } from "../lib/fileRef";
+import { normalizePersistentStructuredViewLimit } from "../lib/persistentStructuredViews";
+import { StructuredView } from "./acp/StructuredView";
+
+interface Props {
+  activeSessionId: string | null;
+  sessions: SessionResponse[];
+  persistent: boolean;
+  maxPersistentStructuredViews?: number;
+  /** On desktop this is true whenever the active session is structured.
+   *  On mobile it is true only when `view === "agent"`. The stack uses it to
+   *  decide whether the active structured session is actually interactive. */
+  visible: boolean;
+  onOpenFileRef?: (ref: FileRef) => void;
+  onOpenAgentsPane?: () => void;
+  /** Restore a trashed session by id. If absent, trashed sessions render
+   *  their banner read-only. */
+  onRestoreSession?: (sessionId: string) => Promise<boolean> | void;
+}
+
+export function StructuredViewStack({
+  activeSessionId,
+  sessions,
+  persistent,
+  maxPersistentStructuredViews,
+  visible,
+  onOpenFileRef,
+  onOpenAgentsPane,
+  onRestoreSession,
+}: Props) {
+  const [recentIds, setRecentIds] = useState<string[]>([]);
+  const limit = normalizePersistentStructuredViewLimit(maxPersistentStructuredViews);
+  const sessionsById = useMemo(() => new Map(sessions.map((session) => [session.id, session])), [sessions]);
+
+  useEffect(() => {
+    let cancelled = false;
+    queueMicrotask(() => {
+      if (cancelled) return;
+      if (!persistent) {
+        setRecentIds([]);
+        return;
+      }
+      if (!activeSessionId) {
+        setRecentIds((ids) => ids.filter((id) => sessionsById.has(id)));
+        return;
+      }
+      const activeSession = sessionsById.get(activeSessionId);
+      if (!activeSession || activeSession.view !== "structured") {
+        // Active session is not structured; keep recent structured sessions warm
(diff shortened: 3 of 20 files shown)


web/src/components/StructuredViewStack.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (!persistent) {
+      return activeStructuredId ? [activeStructuredId] : [];
+    }
+    return activeStructuredId ? [activeStructuredId, ...retained].slice(0, limit) : retained.slice(0, limit);
+  }, [activeStructuredId, limit, persistent, recentIds, sessionsById]);
+
+  if (visibleIds.length === 0) return null;
+
+  return (
+    <div className="relative flex-1 min-h-0 overflow-hidden">
```
Seluj78, line 84: Blocker: this root is `flex: 1 1 0%`. When persistence is on, the active session is a terminal session, and at least one structured session is retained, `visibleIds` is non-empty (`:78`), so this div renders as a sibling of `TerminalSessionStack`'s root (`flex-1 min-h-0 overflow-hidden relative`, TerminalSessionStack.tsx:54) inside a `flex flex-col` parent (App.tsx:1879, and the mobile `layerClass` base at MobileMainPane.tsx:59). The two flex items split the column, so the terminal pane collapses to roughly half height and the other half is an invisible structured layer.

Your own test `"does not add terminal sessions to the recent list"` (StructuredViewStack.test.tsx:117) sets up exactly this state.

Fix: `absolute inset-0 overflow-hidden`. Both parents already establish a containing block, and the children here are already `absolute inset-0`, so nothing else changes.

web/src/App.tsx
```diff
@@ -1878,31 +1876,27 @@ function AppContent({
             onToggleCollapse={toggleRightDock}
             left={
               <div className="flex-1 flex flex-col min-h-0 overflow-hidden relative">
-                <div className={selectedFilePath ? "hidden" : "flex-1 flex flex-col min-h-0 overflow-hidden"}>
-                  {activeSession?.view === "structured" ? (
+                <div className={selectedFilePath ? "hidden" : "relative flex-1 flex flex-col min-h-0 overflow-hidden"}>
+                  {(activeSession?.view === "structured" || webSettings.persistentStructuredViews) && (
```
Seluj78, line 1880: Blocker (same as StructuredViewStack.tsx:84): this mount condition keeps the structured stack rendered while the active session is a terminal session, and the sibling `TerminalSessionStack` below is also `flex-1`. Two `flex: 1 1 0%` children of a `flex flex-col` parent split the height.

Also Required, separate issue: this file's `<MobileMainPane>` call site at 1791-1822 never passes the two new props `onRestoreSession` or `onOpenAgentsPane`, which breaks the mobile trashed-session banner (see MobileMainPane.tsx:131) and leaves two props nothing supplies (AGENTS.md, no dead code).

web/src/components/MobileMainPane.tsx
```diff
@@ -108,24 +118,21 @@ export function MobileMainPane({
...
+                activeSessionId={activeSessionId}
+                sessions={sessions}
+                persistent={webSettings.persistentStructuredViews}
+                maxPersistentStructuredViews={webSettings.maxPersistentStructuredViews}
+                visible={view === "agent"}
                 onOpenFileRef={onOpenFileRef}
-                fileRefSession={activeSession}
-                isSandboxed={activeSession.is_sandboxed}
+                onOpenAgentsPane={onOpenAgentsPane}
+                onRestoreSession={onRestoreSession}
```
Seluj78, line 131: Required: `onRestoreSession` is `undefined` here, because App.tsx:1791-1822 never passes it. But StructuredViewStack.tsx:110 builds the callback unconditionally (`onRestore={session.trashed_at ? () => onRestoreSession?.(sessionId) : undefined}`), so it is always truthy. `TrashedWorkerStoppedBanner` therefore renders the Restore button (StructuredView.tsx:2240), `handleRestore` sets `restoring = true`, and `Promise.resolve(undefined)` resolves to `undefined`, which fails the `ok === false` reset at StructuredView.tsx:2221. The button is stuck disabled at "Restoring…" until reload and nothing restores.

Before this PR mobile passed no `onRestore` at all, so the banner was correctly read-only. Fix either end: pass the handler from App, or guard on `onRestoreSession` in the stack.

web/src/components/StructuredViewStack.tsx
```diff
@@ -0,0 +1,121 @@
...
+              sessionId={sessionId}
+              active={active}
+              acpWorkerState={session.acp_worker_state ?? "absent"}
+              tool={session.tool}
+              acpAgent={session.acp_agent ?? null}
+              clearAliases={session.clear_aliases}
+              archivedAt={session.archived_at ?? null}
+              snoozedUntil={session.snoozed_until ?? null}
+              trashedAt={session.trashed_at ?? null}
+              onRestore={session.trashed_at ? () => onRestoreSession?.(sessionId) : undefined}
```
Seluj78, line 110: Required: this is truthy whenever the session is trashed, even when `onRestoreSession` is absent, which contradicts the prop doc at :18-20 ("If absent, trashed sessions render their banner read-only"). On mobile `onRestoreSession` is always absent today, so the Restore button renders and hangs at "Restoring…" (see MobileMainPane.tsx:131).

```tsx
onRestore={session.trashed_at && onRestoreSession ? () => onRestoreSession(sessionId) : undefined}
```

web/src/components/StructuredViewStack.tsx
```diff
@@ -0,0 +1,121 @@
...
+    let cancelled = false;
+    queueMicrotask(() => {
+      if (cancelled) return;
+      if (!persistent) {
+        setRecentIds([]);
+        return;
+      }
+      setRecentIds((ids) => {
+        // Drop sessions whose view is no longer structured and remove duplicates.
+        const structuredIds = ids.filter((id) => sessionsById.get(id)?.view === "structured");
```
Seluj78, line 47: Required: this effect depends on `sessionsById`, a `useMemo` over the `sessions` prop, whose identity changes on every session-list refresh or WS session update. Two of the three `setRecentIds` branches (`:49` and `:55`) return a freshly allocated array unconditionally, with no identity short-circuit; only `:59` has one. Each new reference commits a state change, so the stack and all N mounted `<StructuredView>` trees re-render on every unrelated session update.

Apply the same `next.join("\0") === ids.join("\0") ? ids : next` short-circuit to all three branches, or move the "is this session still structured" filter out of state and into the `visibleIds` memo where it already runs.

web/src/components/acp/StructuredView.tsx
```diff
@@ -706,7 +718,58 @@ function AcpChrome({
...
+  // reactivated by the persistent structured-view stack. Save on
+  // active -> inactive; restore on inactive -> active once the
+  // transcript has stabilized. A ref latch prevents rapid toggles from
+  // interleaving save and restore.
+  const activeTransitionLatchRef = useRef(false);
+  const prevActiveRef = useRef(active);
+  /* eslint-disable react-you-might-not-need-an-effect/no-event-handler */
+  useEffect(() => {
+    const prev = prevActiveRef.current;
+    prevActiveRef.current = active;
```
Seluj78, line 733: Required: `prevActiveRef.current` is advanced before both guards on the next line, so the transition is consumed even when the effect does no work. Two paths lose the edge permanently:

1. A deactivate then activate pair inside one `requestAnimationFrame` hits the activation run with `activeTransitionLatchRef.current` still set. It returns having already recorded `prevActiveRef.current = true`; no later dependency change can re-detect the transition, so the reader's scroll offset is neither saved nor restored.
2. The readiness gate at `:745` (`if (!hasEverOpened && status !== "open" && state.activity.length === 0) return;`) has the same shape, so an activation that arrives before the transcript is ready never retries once it becomes ready.

Keep an explicit armed-until-handled ref, and only advance `prevActiveRef` once the save or restore actually runs.

web/src/components/acp/AttentionChime.tsx
```diff
@@ -15,7 +18,7 @@ interface Props {
 // backgrounded) and the in-app toast (when foregrounded). A question
 // arriving while an approval is already pending does not re-chime, but its
 // OS push still fires on the live event edge regardless. See #1038, #2146.
-export function AttentionChime({ approvals, elicitations }: Props): null {
-  useApprovalSound(approvals + elicitations);
+export function AttentionChime({ approvals, elicitations, active = true }: Props): null {
+  useApprovalSound(active ? approvals + elicitations : 0);
```
Seluj78, line 22: Required: `useApprovalSound` is a rising-edge detector, documented and implemented as "Watch `pendingCount` for a 0 -> >=1 edge and play the configured approval sound". Feeding it `active ? count : 0` drives N to 0 on deactivation and 0 to N on reactivation, which is exactly that edge, so switching back to a warm session with a pending approval re-rings for an approval the user already saw. `REPLAY_QUIET_MS` only swallows the post-mount replay, not a reactivation.

Gate the playback instead of the count: track which approval the chime already fired for, or pass `active` into `useApprovalSound` and have it suppress playback while keeping its tracked count intact.

web/src/hooks/useMobileKeyboard.ts
```diff
@@ -47,16 +47,22 @@ function createKeyboardStore() {
 
 type KeyboardStore = ReturnType<typeof createKeyboardStore>;
 
-export function useMobileKeyboard() {
+export function useMobileKeyboard(enabled = true) {
   const [store] = useState<KeyboardStore>(() => createKeyboardStore());
-  const state = useSyncExternalStore(store.subscribe, store.getSnapshot);
+  const state = useSyncExternalStore(enabled ? store.subscribe : () => () => {}, store.getSnapshot);
```
Seluj78, line 52: Nit: the inline `() => () => {}` has a fresh identity on every render, so `useSyncExternalStore` tears down and re-establishes the subscription each render for every inactive session. Hoist it:

```ts
const NOOP_SUBSCRIBE = () => () => {};
```

web/src/components/acp/Composer.tsx
```diff
@@ -344,6 +344,10 @@ interface Props {
    *  when the user submits while browsing the queue, so an edit does not
    *  enqueue a duplicate. */
   editQueuedPrompt: (id: string, text: string) => void;
+  /** When false the session is mounted but inactive (kept warm by the
+   *  persistent structured-view stack). The composer avoids autofocus,
+   *  imperative focus, and draft persistence work. */
```
Seluj78, line 349: Nit: the doc says the composer "avoids autofocus, imperative focus, and draft persistence work", but the draft persistence effects at `:834` and `:879` are not gated on `active`. Inactive composers still hydrate drafts, register `beforeunload` / `pagehide` / `visibilitychange` listeners, and run the debounced `setDraft` write. Per-session keys mean no cross-contamination, so this is a doc accuracy issue, not a bug: drop the clause or gate the effects.

web/src/components/StructuredViewStack.tsx
```diff
@@ -0,0 +1,121 @@
...
+  useEffect(() => {
+    let cancelled = false;
+    queueMicrotask(() => {
+      if (cancelled) return;
+      if (!persistent) {
+        setRecentIds([]);
+        return;
+      }
+      setRecentIds((ids) => {
+        // Drop sessions whose view is no longer structured and remove duplicates.
```
Seluj78, line 46: Nit: the comment says "remove duplicates", but the `filter` on the next line only drops non-structured ids. If the invariant matters, use a `Set`; otherwise drop the claim.

web/src/components/acp/StructuredView.tsx
```diff
@@ -1730,19 +1799,24 @@ export function RateLimitRecoverySection({
...
   sessionId: string;
   currentAgent: string | null;
   onPrefill: (text: string) => void;
+  active?: boolean;
   children: (renderProps: { onSwitchAgent: () => void }) => React.ReactNode;
 }) {
   const [open, setOpen] = useState(false);
+  // Close the modal when this session becomes inactive; derive the effective
+  // open state so no setState in an effect is needed.
+  const effectiveOpen = active && open;
```
Seluj78, line 1814: Nit: the comment above says "Close the modal when this session becomes inactive", but this only hides it. `open` survives, so reactivating the session pops the modal back with no user action. Either reset `open` on the transition or reword the comment to say the modal is suppressed, not closed.

web/src/components/settings/StructuredViewDisplaySettings.tsx
```diff
@@ -31,6 +37,72 @@ export function StructuredViewDisplaySettings() {
...
+                <input
+                  type="range"
+                  aria-labelledby="max-persistent-structured-views-label"
+                  min={MIN_PERSISTENT_STRUCTURED_VIEWS}
+                  max={MAX_PERSISTENT_STRUCTURED_VIEWS}
+                  step={1}
+                  value={maxPersistentStructuredViews}
+                  onChange={(e) =>
+                    update({
+                      maxPersistentStructuredViews: normalizePersistentStructuredViewLimit(Number(e.target.value)),
```
Seluj78, line 77: Nit: normalizing on every keystroke means clearing the field snaps the value to `1`, because `Number("") === 0` and `normalizePersistentStructuredViewLimit` clamps to `MIN`. Normalize on blur or commit and keep the raw string while editing.

Seluj78 (changes requested):

## AoE Review

Reviewed: PR #3468 (base `main`@6500195). Files changed: 20 (web, web-tests).
Unresolved comments: none (4 inline threads, all resolved).
Cross-model consult: `gemini-3.1-pro-preview`, `gpt-5.6-sol`, `claude-opus-4-7`, `grok-4.5` (18 claims, 5 became findings).

**CI note:** the main workflows never ran on head `2858bb3d`. Only `PR Title Check` and `PR Template Check` completed. No Rust build, no web lint/typecheck/unit, no coverage-matrix validation, no Playwright. So a green merge box here is not evidence; finding 1 below is currently invisible.

### Blocker

**1. `web/src/components/StructuredViewStack.tsx` is not in the coverage matrix; the validator fails.**

`node web/tests/validate-coverage-matrix.mjs` on this head:

```
- Component 'web/src/components/StructuredViewStack.tsx' is not listed in any coverage-matrix surface or coverage-matrix.exempt.json. Add it to a surface's components[] or to the exempt list with a one-line reason.
```

Rule: `AGENTS.md` § Definition of done, "CI fails on a missing matrix entry." The sibling `TerminalSessionStack.tsx` is listed at `web/tests/coverage-matrix.json:636`. The new Vitest spec `src/components/__tests__/StructuredViewStack.test.tsx` exists, but no surface points at it.

**2. The terminal pane collapses to roughly half height whenever the setting is on.**

`App.tsx:1880` and `MobileMainPane.tsx:121` mount the stack on `(activeSession?.view === "structured" || webSettings.persistentStructuredViews)`, and `TerminalSessionStack` renders under the sibling condition `activeSession?.view !== "structured"`. With persistence on, an active terminal session, and at least one retained structured session, `StructuredViewStack` does not return `null` (`visibleIds = retained.slice(0, limit)`, `StructuredViewStack.tsx:78`), so its root `<div className="relative flex-1 min-h-0 overflow-hidden">` (`:84`) renders as a sibling of `TerminalSessionStack`'s root `flex-1 min-h-0 overflow-hidden relative` (`TerminalSessionStack.tsx:54`). Both are `flex: 1 1 0%` inside a `flex flex-col` parent (`App.tsx:1879`, and the mobile `layerClass` base at `MobileMainPane.tsx:59`), so they split the column. Half the pane becomes an invisible structured layer.

This is reachable through the exact path your own test sets up: `"does not add terminal sessions to the recent list"` (`StructuredViewStack.test.tsx:117`) switches from a structured session to a terminal session with persistence on, and asserts the structured layer stays mounted.

Suggested fix: the stack root should be `absolute inset-0 overflow-hidden`, not `flex-1`. Both parents already establish a containing block, and the children are already `absolute inset-0`.

### Required

**3. On mobile, a trashed session's Restore button sticks at "Restoring…" forever.**

`App.tsx:1791-1822` (the `<MobileMainPane>` call site, outside this diff) never passes the two new props `onRestoreSession` and `onOpenAgentsPane`. `MobileMainPane.tsx:131` forwards `undefined`, but `StructuredViewStack.tsx:110` builds the callback unconditionally:

```tsx
onRestore={session.trashed_at ? () => onRestoreSession?.(sessionId) : undefined}
```

That is always a truthy function, so `TrashedWorkerStoppedBanner` renders the button (`StructuredView.tsx:2240`), `handleRestore` sets `restoring = true`, and `Promise.resolve(undefined)` resolves to `undefined`, which fails the `ok === false` reset at `StructuredView.tsx:2221`. The button is disabled at "Restoring…" until reload, and nothing restores. Before this PR mobile passed no `onRestore` at all, so the banner was correctly read-only.

This also contradicts the component's own documented contract at `StructuredViewStack.tsx:18-20` ("If absent, trashed sessions render their banner read-only"), and leaves two props that nothing supplies (`AGENTS.md` § Coding Style, no dead code).

**4. Every mounted structured session re-renders on every session-list update.**

The retention effect depends on `sessionsById`, a `useMemo` over the `sessions` prop, whose identity changes on every session-list refresh or WS session update. Two of the three `setRecentIds` branches (`StructuredViewStack.tsx:49` and `:55`) return a freshly allocated array unconditionally, with no identity short-circuit; only the third branch has one (`:59`). Each new reference commits a state change, re-rendering the stack plus all N mounted `<StructuredView>` trees on every unrelated session update.

**5. The scroll save/restore effect burns its transition and never retries.**

`StructuredView.tsx:733-735`:

```tsx
const prev = prevActiveRef.current;
prevActiveRef.current = active;
if (active === prev || activeTransitionLatchRef.current) return;
```

`prevActiveRef.current` is advanced before both guards, so two paths lose the edge permanently:

- A deactivate then activate pair inside one `requestAnimationFrame` hits the activation run with the latch still set. It returns, having already recorded `prevActiveRef.current = true`; no later dependency change can re-detect the transition, so the reader's offset is neither saved nor restored.
- The readiness gate at `:745` (`if (!hasEverOpened && status !== "open" && state.activity.length === 0) return;`) has the same shape, so an activation arriving before the transcript is ready never retries once it becomes ready.

**6. The approval chime replays on every switch back to a warm session.**

`AttentionChime.tsx:22` passes `active ? approvals + elicitations : 0` into `useApprovalSound`, which is documented and implemented as "Watch `pendingCount` for a 0 -> >=1 edge and play the configured approval sound". Gating the count rather than the playback drives N to 0 on deactivation and 0 to N on reactivation, which is precisely that edge. `REPLAY_QUIET_MS` only swallows the post-mount replay, not a reactivation, so returning to a warm session with a pending approval re-rings for an approval the user already saw. Gate the playback on a "have I already chimed for this approval" key instead of on `active`.

### Nit

7. `useMobileKeyboard.ts:52`: the inline `() => () => {}` has a fresh identity every render, so every inactive session unsubscribes and resubscribes on each render. Hoist a module-level `const NOOP_SUBSCRIBE = () => () => {};`.
8. `Composer.tsx:347-350`: the doc says the composer "avoids autofocus, imperative focus, and draft persistence work", but the draft effects at `:834` and `:879` are not gated on `active`. Drop the clause or gate the effects.
9. `StructuredViewStack.tsx:46`: the comment says "remove duplicates"; the `filter` below does not dedupe.
10. `StructuredView.tsx:1811-1814`: the comment says "Close the modal when this session becomes inactive", but `effectiveOpen = active && open` only hides it. `open` survives, so reactivating pops the modal back.
11. `StructuredViewDisplaySettings.tsx:70-83`: the number input normalizes on every keystroke, so clearing the field snaps it to `1` (`Number("") === 0`, clamped to `MIN`). Normalize on blur or commit.

### PR metadata

`PR body: Checklist`: "For UI changes: included screenshot or recording" is unchecked and nothing is attached, for a PR that adds a settings control and changes desktop and mobile pane layout. `.github/pull_request_template.md` and `AGENTS.md` § Commit & Pull Request Guidelines both ask for it.

### Followup, not blocking

Three of the four consulted models independently proposed the same lighter shape: keep the module-level `stateCache` plus a ref-counted per-session WebSocket supervisor outside React, and mount exactly one `<StructuredView>`. The current shape keeps N live `useAcpSession` instances, because the connect effect depends on `[sessionId, ...]` and not on `active`. Each retained session dials, subscribes, reduces every frame, and registers its own `pagehide` / `visibilitychange` / `online` / `orientationchange` listeners. `inert` suppresses paint and interaction, not frame processing. On mobile at limit 5 with a chatty agent that is real main-thread and memory cost the user attributes to the active session. Three of the six findings above are omissions in the `active` threading, which is the argument for collapsing that threading entirely. Worth an issue rather than a change here.

Separately, worth confirming the daemon's `[acp]` limits before raising the default or the cap: a tab can now hold 5x the authenticated ACP sockets and nothing was added server side.

### Verdict

**Project Fit:** Accept
**Placement:** core (structured-view session lifecycle is substrate; `docs/plugins.md` scopes plugins to optional third-party capability, and the plugin API cannot host pane mounting)

The idea is right and this should land. It cannot land as-is: the flex-sibling layout bug halves the terminal pane the moment the setting is enabled and the user visits a terminal session, and the mobile Restore button is permanently stuck because `App.tsx` never wires the two new `MobileMainPane` props. Both are small fixes. Findings 4 through 6 are contained edits too. Note that none of the real CI ran on this head, so finding 1 will only surface once the workflows are approved.

Issue #2276, opened 2026-06-18
feat(structured): Import existing Claude Code sessions


victorhooi:
Is there some way of importing existing Claude Code sessions?

As in, doing a claude --resume, and pulling from the existing session list - and making that a session in Agent of Empires?

njbrake:
Hi @victorhooi ! If you open a  new aoe claude code session and then run /resume, that should load up the session and have it be resumed and managed in aoe.

Let me know if that doesn't make sense 🙏

jerome-benoit:
The CLI also allows the set the session id.

njbrake:
I think we've resolved the question so I'll close this out

Seluj78:
I think it'd be nice to be able to do that in the structured view as well 👀 so I'd lean towards keeping it open for this ? + Documentation needed?

njbrake:
@Seluj78 oooh good point. I reopened and updated the title.

victorhooi:
When you say structured view - does that mean the dialog that appears when you hit "n" or "N" (or a similar dialog)?

That would be really awesome, actually - some kind of picker within aoe - that reads from your existing sessions on disk, allows you pick them, and resume them (and then get them imported cleanly into Agent of Empires 😀)

Is that possible at all?

Seluj78:
> When you say structured view - does that mean the dialog that appears when you hit "n" or "N" (or a similar dialog)?

Structured view is the web dashboard native rendering mode :) you can try it with `aoe serve` and then starting a new session and turning on the structured view for it!

> That would be really awesome, actually - some kind of picker within aoe - that reads from your existing sessions on disk, allows you pick them, and resume them (and then get them imported cleanly into Agent of Empires 😀)
> 
> Is that possible at all?

I don't actually know right now, but it certainly might be. I need to look into the capabilities of Claude's ACP as well as other ACPs (like codex open code etc...) to see how their sessions are stored and to see how possible it might be to import things. But I really like your idea of scanning existing sessions, and I think it's doable if the ACP permits it!

I'm still on vacation until tomorrow evening so I can't check properly but I'll investigate!

Necmttn:
I would split this into read-only import and live resume.

Read-only import can scan `~/.claude/projects/<encoded-cwd>/*.jsonl`, build structured sessions, and never invoke Claude. Live resume should be a separate explicit action, because `claude --resume <sessionId>` changes harness state and can send messages.

[...]

Issue #1921, opened 2026-06-03
Cockpit: orphaned __cockpit-runner + agent processes leak forever when their daemon dies without an explicit stop


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

`htop` filtered on `claude` shows far more processes than live sessions justify. With 5 cockpit sessions running I expect roughly 5 `aoe __cockpit-runner` shims plus 5 `claude-agent-acp` node agents. Instead there were hundreds: ~300 `__cockpit-runner` processes, 67 `fakeAcpAgent.mjs` (the E2E fake agent), 15 `aoe_e2e` tmux servers, with the oldest runners 21+ hours old and ~826 MB total RSS across the runner pile.

The 5 real sessions are fine. The hundreds of extras are orphaned children of daemons that have since died, mostly from test runs. They never get reaped because nothing is left alive to reap them.

This is two problems stacked:

1. A product gap: a cockpit runner (and its node ACP agent and `claude` grandchild) has no way to self-terminate when the daemon that owns it goes away and never comes back. It leaks forever.
2. A test-harness multiplier: both the live-Playwright and e2e harnesses tear down by killing only the daemon process, never the runners, then delete the isolated temp `$HOME`. Every cockpit session a test spawns orphans one runner plus one node agent permanently. Hundreds of test runs produce the pile.

## What's there today

Runners are deliberately built to outlive the daemon so an `aoe serve --stop` followed by a fresh `aoe serve` can reattach mid-turn. The runner is spawned via `setsid` into its own process group (`src/cockpit/runner.rs:4-5`) and documents itself as "the per-worker shim that owns the agent subprocess and outlives `aoe serve`."

Graceful daemon shutdown detaches, it does not kill:

```rust
// src/server/mod.rs:1087
// Detach (but do NOT kill) every cockpit ACP worker. The per-session
// `aoe __cockpit-runner` shims outlive this daemon ...
cockpit_supervisor.detach_all().await;
```

`detach_all` (`src/cockpit/supervisor.rs:1978`) closes the socket and aborts the drain task; the runner observes EOF and goes back to accepting. It does not signal the runner. A forced exit (grace window exceeded) skips even the detach block (`src/server/mod.rs:1069-1076`), and a `SIGKILL` or crash skips everything.

The runner's main loop only ever exits on agent-child death or an explicit signal:

```rust
[...]

PR #3476, opened 2026-08-22
fix: cancel timed-out ACP runner handshakes

### LINKED ISSUE #3474: ACP socket runner survives handshake timeout and can strand resumed sessions
## Problem

When an ACP socket runner takes longer than the daemon's initialize timeout, the daemon returns a timeout but leaves that runner and its connection task alive. The abandoned task can later complete `initialize` and `session/load`, retaining the provider's thread writer while a replacement runner is started.

The replacement then fails `session/load` with an active-writer error. AoE can subsequently reattach to a runner whose adapter never installed the session in memory, so every prompt fails with `Session <id> not found` even though the stored session ID, transcript, and provider catalog row are valid.

## Observed sequence

1. Detached socket runner starts.
2. Daemon waits 30 seconds for `initialize` and times out.
3. The same connection completes late and successfully runs `session/load`.
4. A concurrent attach receives `Already initialized` and falls back to another spawn.
5. Replacement loads fail because the late connection still owns the thread writer.
6. A later attach reuses a runner in resume mode, but prompts fail because that adapter never loaded the session.

## Root cause

- The socket attach path has no direct child handle, so handshake failure/timeout does not terminate the registered runner.
- The reconciler's attach timeout deletes the registry record without terminating the runner.
- Concurrent current-v2 control attaches can both miss the handshake cache before either response is published.

## Expected behavior

A handshake timeout/failure must revoke and terminate that exact runner generation before returning. Late work from the timed-out generation must not run `session/load`, retain a writer, or become attachable. Only one current-v2 control handshake may reach the ACP writer per runner generation.

## Proposed regression

Use the real `__acp-runner` entrypoint with a fake ACP agent that delays `initialize` past a bounded test timeout:

- first attach times out;
- assert its PID, registry record, and socket generation are gone;
- start a replacement runner and resume the exact stored session;
- assert the fake agent sees one initialize for each runner and exactly one successful `session/load`, with no duplicate load/new calls.

## Proposed fix

- Terminate and wait for the exact registered socket-runner generation on connect handshake error/timeout.
- Terminate, rather than only delete, that generation on reconciler attach timeout.
- Serialize current-v2 runner control handshakes through response and cache publication, replaying the cached result to overlapping attaches.

aaiyer:
Fixes #3474.

## Summary

- close the exact runner control socket when daemon attach is cancelled
- keep one task as the control-frame reader so cancelling handshake work cannot discard a partial frame
- terminate the runner only when control is lost before its ACP session is established
- keep agent-returned handshake errors recoverable so `session/load` can fall back to `session/new`
- retain the existing registry cleanup for timeout paths with no usable v2 control channel

The broader lifecycle redesign raised during review remains deferred to #3487.

## Verification

- `cargo test --test acp_runner_control --features serve` (5/5)
- `AOE_WEB_DIST="$PWD/target/ci-web-dist" cargo clippy --all-features -- -D warnings -A clippy::duplicated-attributes`

The narrow Clippy allowance is for one unchanged current-main lint emitted by the local Rust 1.93.1 toolchain in `src/tui/dialogs/serve.rs`; current upstream-main CI is green.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing affected tests pass
- [x] Documentation was updated where necessary: N/A; no documented interface changed
- [x] For UI changes: N/A; no UI changes

## Test Coverage Analysis

**Web dashboard / structured view**
- [x] N/A: backend runner lifecycle only

**TUI / CLI**
- [x] N/A: no TUI or CLI behavior changed
- [ ] Added or updated an e2e test

## AI Usage

- [ ] No AI was used
- [x] AI was used

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -12,6 +12,7 @@
 //! mpsc channel into ACP requests until shutdown.
 
 use std::collections::{HashMap, VecDeque};
+use std::os::fd::{AsRawFd, RawFd};
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::atomic::{AtomicU64, Ordering as AtomicOrdering};
@@ -2767,6 +2768,7 @@ impl AcpClient {
         .await;
         let external_terminal_guard = control_client.as_ref().map(|_| guard);
         let external_prompt_in_flight = control_client.as_ref().map(|_| prompt_in_flight);
+        let mut handshake_control = ShutdownControlOnDrop(control_client.clone());
 
         let (ready_tx, ready_rx) = oneshot::channel::<Result<(), AcpError>>();
 
@@ -2799,8 +2801,8 @@ impl AcpClient {
             )
             .instrument(conn_span),
         );
-
         wait_for_handshake(&session_label, ready_rx, None, &install_binary).await?;
+        handshake_control.0.take();
 
         Ok(Self {
             session_id,
@@ -3200,6 +3202,19 @@ impl AcpClient {
     /// mutex (which would deadlock send_prompt).
     pub fn take_inbound(&mut self) -> Option<mpsc::Receiver<Event>> {
         self.inbound.take()
+    }
+}
+
+/// Cancel a socket handshake if its constructor is dropped before completion.
+/// Closing the exact runner control channel cancels only that runner.
+struct ShutdownControlOnDrop(Option<Arc<DaemonControlClient>>);
+
+impl Drop for ShutdownControlOnDrop {
+    fn drop(&mut self) {
+        if let Some(control) = self.0.take() {
+            // SAFETY: `control` keeps this exact socket alive for the call.
+            unsafe { libc::shutdown(control.raw_fd, libc::SHUT_RDWR) };
+        }
     }
 }
 
@@ -4043,6 +4058,7 @@ struct DaemonControlClient {
     write: Mutex<tokio::net::unix::OwnedWriteHalf>,
     handshake_rx: Mutex<mpsc::Receiver<ControlBody>>,
     completion: Arc<std::sync::Mutex<Option<oneshot::Sender<control_protocol::PromptOutcome>>>>,
+    raw_fd: RawFd,
 }
 
 impl DaemonControlClient {
@@ -4286,10 +4302,12 @@ async fn connect_runner_control_v2(
         }
     });
 
+    let raw_fd = write_half.as_ref().as_raw_fd();
     Some(Arc::new(DaemonControlClient {
         write: Mutex::new(write_half),
         handshake_rx: Mutex::new(hs_rx),
         completion,
+        raw_fd,
     }))
 }
 
--- a/src/process/runner.rs
+++ b/src/process/runner.rs
@@ -55,7 +55,7 @@ use serde::Deserialize;
 use tokio::io::{AsyncBufRead, AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::net::{UnixListener, UnixStream};
 use tokio::process::{Child, Command};
-use tokio::sync::Mutex;
+use tokio::sync::{mpsc, watch, Mutex};
 use tracing::{debug, info, warn};
 
 use super::worker_registry::{self, WorkerRecord};
@@ -360,7 +360,7 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
     let control_shared = Arc::clone(&shared);
     let control_session = args.session_id.clone();
     let control_stdin = Arc::clone(&agent_stdin);
-    let control_accept_task = tokio::spawn(async move {
+    let mut control_accept_task = tokio::spawn(async move {
         loop {
             match control_listener.accept().await {
                 Ok((stream, _addr)) => {
@@ -369,13 +369,16 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
                         session = %control_session,
                         "daemon connected (control channel)"
                     );
-                    handle_control_connection(
+                    if handle_control_connection(
                         stream,
                         Arc::clone(&control_shared),
                         Arc::clone(&control_stdin),
                         control_session.clone(),
                     )
-                    .await;
+                    .await
+                    {
+                        return;
+                    }
                     info!(
                         target: "acp.runner",
                         session = %control_session,
@@ -517,6 +520,21 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
         }
         _ = accept_loop => {
             warn!(target: "acp.runner", session = %session_id, "accept loop exited unexpectedly");
+        }
+        result = &mut control_accept_task => {
+            match result {
+                Ok(()) => {
+                    // The daemon owns path cleanup after cancelling an
+                    // incomplete handshake and may already have spawned a
+                    // replacement. Never unlink that replacement here.
+                    preserve_registry = true;
+                }
+                Err(error) => {
+                    warn!(target: "acp.runner", session = %session_id, "control accept task failed: {error}");
+                }
+            }
+            let _ = agent_child.start_kill();
+            let _ = agent_child.wait().await;
         }
     }
 
@@ -1704,6 +1722,27 @@ async fn handle_connection(
     shared.clear_outbound().await;
 }
 
+enum ControlRead {
+    Frame(ControlBody),
+    Closed,
+    Failed(String),
+}
+
+async fn await_handshake_or_control_loss<T>(
+    control_closed: &mut watch::Receiver<bool>,
+    handshake: impl std::future::Future<Output = T>,
+) -> Option<T> {
+    if *control_closed.borrow() {
+        return None;
+    }
+    tokio::pin!(handshake);
+    tokio::select! {
+        biased;
+        _ = control_closed.changed() => None,
+        result = &mut handshake => Some(result),
+    }
+}
+
 /// Handle one control-channel connection (#2976 Phase B). Greets with
 /// `Hello`, installs the persistent write half (draining any completion
 /// buffered across a no-daemon gap), then drives the runner-owned protocol:
@@ -1724,7 +1763,7 @@ async fn handle_control_connection(
     shared: Arc<RunnerShared>,
     agent_stdin: Arc<Mutex<tokio::process::ChildStdin>>,
     session_id: String,
-) {
+) -> bool {
     let (mut read_half, write_half) = stream.into_split();
     let mut write_half = Some(write_half);
     if !shared
@@ -1732,25 +1771,60 @@ async fn handle_control_connection(
         .await
     {
         shared.clear_control_outbound().await;
-        return;
+        return false;
     }
-    loop {
-        let body = match control_protocol::read_frame(&mut read_half).await {
-            Ok(Some(body)) => body,
-            Ok(None) => break, // clean EOF: daemon closed the control socket.
-            Err(e) => {
-                warn!(
-                    target: "acp.runner",
-                    session = %session_id,
-                    "control read error: {e}"
-                );
-                break;
+
+    let (frame_tx, mut frame_rx) = mpsc::channel(8);
+    let (control_closed_tx, mut control_closed_rx) = watch::channel(false);
+    let reader_session = session_id.clone();
+    let frame_reader = tokio::spawn(async move {
+        loop {
+            match control_protocol::read_frame(&mut read_half).await {
+                Ok(Some(frame)) => match frame_tx.try_send(ControlRead::Frame(frame)) {
+                    Ok(()) => {}
+                    Err(mpsc::error::TrySendError::Closed(_)) => return,
+                    Err(mpsc::error::TrySendError::Full(_)) => {
+                        warn!(target: "acp.runner", session = %reader_session, "control command queue exceeded capacity");
+                        let _ = control_closed_tx.send(true);
+                        return;
+                    }
+                },
+                Ok(None) => {
+                    let _ = frame_tx.try_send(ControlRead::Closed);
+                    let _ = control_closed_tx.send(true);
+                    return;
+                }
+                Err(error) => {
+                    let _ = frame_tx.try_send(ControlRead::Failed(error.to_string()));
+                    let _ = control_closed_tx.send(true);
+                    return;
+                }
+            }
+        }
+    });
+
+    let mut handshake_complete = shared.acp_session_id().await.is_some();
+    let terminate_runner = 'connection: loop {
+        let body = match frame_rx.recv().await {
+            Some(ControlRead::Frame(frame)) => frame,
+            Some(ControlRead::Closed) | None => break 'connection !handshake_complete,
+            Some(ControlRead::Failed(error)) => {
+                warn!(target: "acp.runner", session = %session_id, "control read error: {error}");
+                break 'connection !handshake_complete;
             }
         };
         match body {
             ControlBody::Attach { .. } => {}
             ControlBody::Initialize { request } => {
-                let frame = match shared.run_or_replay_initialize(&agent_stdin, request).await {
+                let Some(result) = await_handshake_or_control_loss(
+                    &mut control_closed_rx,
+                    shared.run_or_replay_initialize(&agent_stdin, request),
+                )
+                .await
+                else {
+                    break 'connection true;
+                };
+                let frame = match result {
                     Ok(result) => ControlBody::Initialized { result },
                     Err(error) => {
                         warn!(target: "acp.runner", session = %session_id, "initialize failed: {error}");
@@ -1760,14 +1834,22 @@ async fn handle_control_connection(
                 shared.emit_control(frame).await;
             }
             ControlBody::EstablishSession { method, request } => {
-                let frame = match shared
-                    .run_or_replay_session(&agent_stdin, &method, request)
-                    .await
-                {
-                    Ok((acp_session_id, result)) => ControlBody::SessionReady {
-                        acp_session_id,
-                        result,
-                    },
+                let Some(result) = await_handshake_or_control_loss(
+                    &mut control_closed_rx,
+                    shared.run_or_replay_session(&agent_stdin, &method, request),
+                )
+                .await
+                else {
+                    break 'connection true;
+                };
+                let frame = match result {
+                    Ok((acp_session_id, result)) => {
+                        handshake_complete = true;
+                        ControlBody::SessionReady {
+                            acp_session_id,
+                            result,
+                        }
+                    }
                     Err(error) => {
                         warn!(target: "acp.runner", session = %session_id, "{method} failed: {error}");
                         ControlBody::HandshakeFailed { error }
@@ -1792,8 +1874,10 @@ async fn handle_control_connection(
             // Runner -> daemon frames should never arrive here; ignore.
             _ => {}
         }
-    }
+    };
+    frame_reader.abort();
     shared.clear_control_outbound().await;
+    terminate_runner
 }
 
 fn spawn_agent(
--- a/tests/acp_runner_control.rs
+++ b/tests/acp_runner_control.rs
@@ -84,7 +84,7 @@ fn wait_for_u32(path: &Path, what: &str) -> u32 {
     let deadline = Instant::now() + Duration::from_secs(10);
     loop {
         if let Ok(value) = std::fs::read_to_string(path) {
-            if let Ok(value) = value.parse() {
+            if let Ok(value) = value.trim().parse() {
                 return value;
             }
         }
--- a/tests/acp_runner_control.rs
+++ b/tests/acp_runner_control.rs
@@ -22,6 +22,9 @@ use std::path::{Path, PathBuf};
 use std::process::{Child, Command};
 use std::time::{Duration, Instant};
 
+use agent_of_empires::acp::acp_client::AcpClient;
+use agent_of_empires::acp::state::AcpSessionId;
+
 /// App data dir for the debug binary under this test's env, mirroring the
 /// XDG resolution the runner uses.
 fn app_dir(home: &Path, xdg: &Path) -> PathBuf {
@@ -73,6 +76,51 @@ fn wait_for(path: &Path, what: &str) {
         if Instant::now() > deadline {
             panic!("{what} never appeared at {}", path.display());
         }
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
+
+fn wait_for_u32(path: &Path, what: &str) -> u32 {
+    let deadline = Instant::now() + Duration::from_secs(10);
+    loop {
+        if let Ok(value) = std::fs::read_to_string(path) {
+            if let Ok(value) = value.parse() {
+                return value;
+            }
+        }
+        if Instant::now() > deadline {
+            panic!("{what} never became a u32 at {}", path.display());
+        }
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
+
+fn wait_for_runner_exit(child: &mut Child) {
+    let deadline = Instant::now() + Duration::from_secs(10);
+    loop {
+        if child.try_wait().expect("inspect runner").is_some() {
+            return;
+        }
+        assert!(Instant::now() < deadline, "timed-out runner stayed live");
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
+
+fn wait_for_record_pid(path: &Path, pid: u32) {
+    let deadline = Instant::now() + Duration::from_secs(10);
+    loop {
+        if std::fs::read(path)
+            .ok()
+            .and_then(|bytes| serde_json::from_slice::<serde_json::Value>(&bytes).ok())
+            .and_then(|record| record["pid"].as_u64())
+            == Some(u64::from(pid))
+        {
+            return;
+        }
+        assert!(
+            Instant::now() < deadline,
+            "replacement record never appeared"
+        );
         std::thread::sleep(Duration::from_millis(50));
     }
 }
@@ -177,11 +225,14 @@ fn runner_reports_native_prompt_complete_over_control_socket() {
 
 /// Write a length-prefixed control frame (4-byte big-endian length, then
 /// the JSON body).
-fn write_frame(stream: &mut UnixStream, body: &serde_json::Value) {
+fn encode_frame(body: &serde_json::Value) -> Vec<u8> {
     let json = serde_json::to_vec(body).expect("serialize frame");
     let len = (json.len() as u32).to_be_bytes();
-    stream.write_all(&len).expect("write frame length");
-    stream.write_all(&json).expect("write frame body");
+    [len.as_slice(), &json].concat()
+}
+
+fn write_frame(stream: &mut UnixStream, body: &serde_json::Value) {
+    stream.write_all(&encode_frame(body)).expect("write frame");
     stream.flush().expect("flush frame");
 }
 
@@ -198,6 +249,217 @@ fn find_python3() -> Option<PathBuf> {
         }
     }
     None
+}
+
+#[tokio::test]
+async fn cancelled_attach_reaps_runner_and_replacement_survives_load_fallback() {
+    if cfg!(not(unix)) {
+        return;
+    }
+    let Some(python3) = find_python3() else {
+        eprintln!("skipping: python3 not found for fake ACP agent");
+        return;
+    };
+
+    let scratch = Scratch::new("latehs");
+    let home = scratch.0.join("home");
+    let xdg = scratch.0.join("xdg");
+    std::fs::create_dir_all(&home).unwrap();
+    std::fs::create_dir_all(&xdg).unwrap();
+
+    let agent_log = scratch.0.join("agent-methods.log");
+    let agent_pid_file = scratch.0.join("agent.pid");
+    let agent_py = scratch.0.join("delayed_agent.py");
+    std::fs::write(
+        &agent_py,
+        r#"
+import json, os, sys, time
+with open(os.environ["AOE_FAKE_AGENT_PID"], "w") as f:
+    f.write(str(os.getpid()))
+for line in sys.stdin:
+    try:
+        msg = json.loads(line)
+    except Exception:
+        continue
+    method = msg.get("method")
+    mid = msg.get("id")
+    if method is None or mid is None:
+        continue
+    with open(os.environ["AOE_FAKE_AGENT_LOG"], "a") as f:
+        f.write(method + "\n")
+    if method == "initialize":
+        time.sleep(int(os.environ["AOE_FAKE_INIT_DELAY_MS"]) / 1000)
+        result = {"protocolVersion": 1, "agentCapabilities": {"loadSession": True, "promptCapabilities": {}}}
+    elif method == "session/load" and os.environ["AOE_FAKE_LOAD_ERROR"] == "1":
+        error = {"code": -32000, "message": "stored session unavailable"}
+        sys.stdout.write(json.dumps({"jsonrpc": "2.0", "id": mid, "error": error}) + "\n")
+        sys.stdout.flush()
+        continue
+    elif method == "session/new":
+        result = {"sessionId": "fresh-thread"}
+    else:
+        result = {}
+    sys.stdout.write(json.dumps({"jsonrpc": "2.0", "id": mid, "result": result}) + "\n")
+    sys.stdout.flush()
+"#,
+    )
+    .unwrap();
+
+    let session_id = "slate001";
+    let workers = app_dir(&home, &xdg).join("acp-workers");
+    let socket = workers.join(format!("{session_id}.sock"));
+    let control = workers.join(format!("{session_id}.control.sock"));
+    let record = workers.join(format!("{session_id}.json"));
+    let bin = env!("CARGO_BIN_EXE_aoe");
+    let spawn_runner = |delay: &str, fail_load: bool| {
+        Command::new(bin)
+            .args([
+                "__acp-runner",
+                "--socket",
+                socket.to_str().unwrap(),
+                "--session-id",
+                session_id,
+                "--agent-name",
+                "fake-agent",
+                "--cwd",
+                home.to_str().unwrap(),
+                "--",
+                python3.to_str().unwrap(),
+                agent_py.to_str().unwrap(),
+            ])
+            .env("HOME", &home)
+            .env("XDG_CONFIG_HOME", &xdg)
+            .env("AOE_FAKE_AGENT_LOG", &agent_log)
+            .env("AOE_FAKE_AGENT_PID", &agent_pid_file)
+            .env("AOE_FAKE_INIT_DELAY_MS", delay)
+            .env("AOE_FAKE_LOAD_ERROR", if fail_load { "1" } else { "0" })
+            .env("AOE_ACP_WATCHDOG_POLL_MS", "5000")
+            .spawn()
+            .expect("spawn acp runner")
+    };
+
+    let mut old = KillOnDrop(spawn_runner("2000", false));
+    wait_for(&record, "old registry record");
+    wait_for(&control, "old control socket");
+    wait_for(&socket, "old relay socket");
+    let old_agent_pid = wait_for_u32(&agent_pid_file, "old agent pid");
+
+    let attach = async {
+        tokio::time::timeout(
+            Duration::from_millis(500),
+            AcpClient::attach(
+                socket.clone(),
+                home.clone(),
+                vec![],
+                "stored-codex-thread".into(),
+                false,
+                AcpSessionId(session_id.into()),
+                None,
+                "fake-agent".into(),
+                None,
+            ),
+        )
+        .await
+    };
+    let spawn_replacement = async {
+        let deadline = tokio::time::Instant::now() + Duration::from_secs(10);
+        while !std::fs::read_to_string(&agent_log)
+            .unwrap_or_default()
+            .lines()
+            .any(|method| method == "initialize")
+        {
+            assert!(
+                tokio::time::Instant::now() < deadline,
+                "old runner never began initialize"
+            );
+            tokio::time::sleep(Duration::from_millis(20)).await;
+        }
+        let replacement = KillOnDrop(spawn_runner("750", true));
+        wait_for_record_pid(&record, replacement.0.id());
+        replacement
+    };
+    let (attach, mut replacement) = tokio::join!(attach, spawn_replacement);
+    assert!(
+        attach.is_err(),
+        "delayed initialize must exceed the attach budget"
+    );
+    wait_for_runner_exit(&mut old.0);
+    assert!(
+        !agent_of_empires::process::worker_registry::is_pid_alive(old_agent_pid),
+        "timed-out agent {old_agent_pid} stayed live"
+    );
+    assert!(
+        replacement
+            .0
+            .try_wait()
+            .expect("inspect replacement")
+            .is_none()
+            && record.exists()
+            && socket.exists()
+            && control.exists(),
+        "old runner teardown removed the replacement runner's files"
+    );
+
+    let baseline = std::fs::read_to_string(&agent_log)
+        .unwrap_or_default()
+        .lines()
+        .filter(|method| *method == "initialize")
+        .count();
+    let mut ctl = UnixStream::connect(&control).expect("connect replacement control");
+    ctl.set_read_timeout(Some(Duration::from_secs(10))).unwrap();
+    assert_eq!(read_frame(&mut ctl)["kind"], "hello");
+    write_frame(
+        &mut ctl,
+        &serde_json::json!({"kind": "attach", "control_protocol_version": 2}),
+    );
+    write_frame(
+        &mut ctl,
+        &serde_json::json!({"kind": "initialize", "request": {"protocolVersion": 1}}),
+    );
+    let load = encode_frame(&serde_json::json!({
+        "kind": "establish_session",
+        "method": "session/load",
+        "request": {"sessionId": "stored-codex-thread", "cwd": home.to_str().unwrap()}
+    }));
+    ctl.write_all(&load[..2]).expect("write partial frame");
+    ctl.flush().expect("flush partial frame");
+    std::thread::sleep(Duration::from_millis(150));
+    assert_eq!(read_frame(&mut ctl)["kind"], "initialized");
+    ctl.write_all(&load[2..]).expect("finish partial frame");
+    ctl.flush().expect("flush completed frame");
+    assert_eq!(read_frame(&mut ctl)["kind"], "handshake_failed");
+    assert!(
+        replacement
+            .0
+            .try_wait()
+            .expect("inspect replacement")
+            .is_none()
+            && record.exists()
+            && socket.exists()
+            && control.exists(),
+        "recoverable session/load error tore down the replacement runner"
+    );
+    write_frame(
+        &mut ctl,
+        &serde_json::json!({
+            "kind": "establish_session",
+            "method": "session/new",
+            "request": {"cwd": home.to_str().unwrap()}
+        }),
+    );
+    let ready = read_frame(&mut ctl);
+    assert_eq!(ready["kind"], "session_ready");
+    assert_eq!(ready["acp_session_id"], "fresh-thread");
+
+    let methods = std::fs::read_to_string(&agent_log).unwrap_or_default();
+    assert_eq!(
+        methods.lines().filter(|m| *m == "initialize").count(),
+        baseline + 1
+    );
+    assert_eq!(methods.lines().filter(|m| *m == "session/load").count(), 1);
+    assert_eq!(methods.lines().filter(|m| *m == "session/new").count(), 1);
+
+    drop(replacement);
 }
 
 /// #2976 Phase B: the runner owns the ACP handshake. Drive it as a v2


src/acp/acp_client.rs
```diff
@@ -9905,6 +10419,552 @@ mod tests {
...
+        crate::process::worker_registry::save(&record).unwrap();
+        let generation = RunnerGeneration::from_record(&record, session_id, &socket).unwrap();
+
+        let started = std::time::Instant::now();
+        assert!(
+            !terminate_runner_generation_with_deadline(&generation, Duration::from_millis(100))
+                .await,
+            "a hard deadline must report NotProvenDead until the child is reaped"
+        );
+        assert!(started.elapsed() < Duration::from_secs(1));
```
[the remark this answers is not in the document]
aaiyer: Fixed in 35b189b8. Once the hard deadline is exhausted, teardown no longer performs a zero-budget reap that could report success after the deadline; it returns not-proven-dead and preserves the exact-generation fence. The focused deadline regression passes.

tests/acp_runner_control.rs
```diff
@@ -200,6 +236,197 @@ fn find_python3() -> Option<PathBuf> {
...
+    let old = KillOnDrop(spawn_runner("500"));
+    let old_pid = old.0.id();
+    wait_for(&record, "old registry record");
+    wait_for(&control, "old control socket");
+    wait_for(&socket, "old relay socket");
+    wait_for(&agent_pid_file, "old agent pid");
+    let old_agent_pid = std::fs::read_to_string(&agent_pid_file)
+        .expect("read old agent pid")
+        .parse::<u32>()
+        .expect("parse old agent pid");
```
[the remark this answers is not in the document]
aaiyer: Fixed in 35b189b8. The integration test now polls until the PID file contains a parseable u32, with the existing bounded timeout.

src/acp/supervisor.rs
```diff
@@ -2823,33 +2878,22 @@ impl<S: BroadcastSink> Supervisor<S> {
...
+                .collect();
+        session_ids.extend(self.workers.lock().await.keys().cloned());
+        for session_id in session_ids {
+            debug!(target: "acp.supervisor", session = %session_id, "shutting down");
+            self.shutdown_with_reason(&session_id, "user_stopped", false, false, None)
+                .await?;
         }
-        #[cfg(not(unix))]
-        let _ = registry_pids;
+        Ok(())
```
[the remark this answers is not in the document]
aaiyer: Fixed in 35b189b8. shutdown_all now processes sessions in deterministic order, continues after each failure, logs failures, and returns an error after the full batch. A regression proves a later worker is still stopped after an earlier failure.

src/server/acp_reconciler.rs
```diff
@@ -1941,31 +2136,47 @@ async fn sweep_orphan_workers(state: &Arc<AppState>, live: &HashSet<&String>) {
...
+        if let Some(generation) = generation {
+            if !crate::acp::acp_client::terminate_runner_generation(&generation).await {
+                tracing::warn!(
+                    target: "acp.supervisor",
+                    session = %record.session_id,
+                    "orphan runner teardown deferred because its generation changed or did not exit"
+                );
+            }
+        }
     }
```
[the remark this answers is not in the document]
aaiyer: Fixed in 35b189b8. The orphan sweep now warns when a live orphan generation cannot be authenticated and explicitly leaves it untouched.

src/acp/supervisor.rs
```diff
@@ -2775,11 +2807,11 @@ impl<S: BroadcastSink> Supervisor<S> {
...
-        if crate::process::worker_registry::load(session_id)
-            .ok()
-            .flatten()
-            .is_some()
-        {
+        let disk_record = crate::process::worker_registry::load_existing_strict(session_id)
+            .map_err(|error| {
+                SupervisorError::Acp(AcpError::Spawn(format!("registry load: {error}")))
+            })?;
+        if let Some(record) = disk_record {
```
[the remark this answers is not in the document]
aaiyer: Fixed in 35b189b8. Both strict registry reads now run via spawn_blocking outside the workers mutex. restart_decision snapshots the observed client and revalidates it after the read so a replacement cannot be mistaken for the original generation.

src/acp/acp_client.rs
```diff
@@ -1893,13 +1893,441 @@ fn runner_socket_deadline() -> std::time::Duration {
...
+                && runner_process_identity_is_current(generation)
+        }
+        Ok(None) if generation.started_at.is_none() => {
+            runner_process_identity_is_current(generation)
+        }
+        Ok(None) if allow_missing_after_term => {
+            match crate::process::worker_registry::process_start_identity_for(generation.pid) {
+                Some(identity) => Some(identity) == generation.process_start_identity,
+                None => false,
+            }
```
jerome-benoit: **P1: escalation loses authority after the group leader exits.** After the first authenticated SIGTERM, the `allow_missing_after_term` path still requires `process_start_identity_for(generation.pid)` to find the leader. For a reattached runner, the leader may already have been reaped while a group descendant ignores TERM: `process_start_identity_for` then returns `None`, while `killpg(pgid, 0)` still confirms a live group. A live Linux process-group probe performed during this review produced `runner_identity_after_term=false`, `descendant_alive_after_term=true`, and `process_group_after_term=alive`. This branch therefore rejects precisely the SIGKILL needed to remove the surviving ACP writer. Preserve escalation authority after the initial authentication under the generation lock until the entire group is gone, without requiring the leader identity again after exit. Add a regression with a reattached runner and a grandchild that ignores TERM.

src/acp/supervisor.rs
```diff
@@ -2743,25 +2749,49 @@ impl<S: BroadcastSink> Supervisor<S> {
...
+                Ok(())
+            };
+            if let Err(error) = teardown {
+                warn!(target: "acp.supervisor", session = %session_id, %error,
+                    "shutdown could not confirm exact runner exit; retaining worker handle to block replacement");
+                return Err(error);
+            }
+            let removed = {
+                let mut workers = self.workers.lock().await;
+                match workers.get(session_id) {
```
jerome-benoit: **P1: the drain can remove this fence before this check runs.** `shutdown_with_reason` releases `workers` before awaiting runner termination. Closing the runner then wakes the drain; with the registry now absent and no restart marker, `restart_decision` returns `UserStopped`, and the drain removes the same `WorkerHandle` at lines 2177-2179. This block no longer finds its `Arc` and returns `worker generation changed` even though teardown succeeded. A concurrent spawn can also reserve a replacement between those lock acquisitions, effectively undoing the requested stop. Both independent reviews found and cross-validated this interleaving. The current shutdown tests keep their fake sender alive, so they do not close the inbound channel and cannot cover the race. Introduce a `Stopping` state or tombstone shared by the drain and spawn paths, so only the shutdown owner removes the handle and every spawn remains fenced until atomic completion.

src/server/acp_reconciler.rs
```diff
@@ -1319,17 +1435,63 @@ async fn resume_one(state: Arc<AppState>, target: ResumeTarget) -> ResumeOutcome
...
+        let legacy_macos_generation = pid_alive && record.process_start_identity.is_none();
+        #[cfg(not(target_os = "macos"))]
+        let legacy_macos_generation = false;
+        if pid_alive && runner_generation.is_none() && !legacy_macos_generation {
+            tracing::warn!(
+                target: "acp.supervisor",
+                session = %id,
+                "runner generation did not match its canonical identity; deferring recovery"
+            );
+            return ResumeOutcome::RetryAfterAttachTimeout;
```
jerome-benoit: **P1: this guard makes `FreshSpawn` unreachable for `live PID + missing socket`.** `is_record_live` explicitly treats a missing socket as non-live, so `adopt_decision` selects `FreshSpawn`; the focused `dead_record_fresh_spawns` test passes and confirms the `live=false -> FreshSpawn` contract. However, `RunnerGeneration::from_record` is evaluated before that branch, and `authenticate_generation_record` requires a peer on the same missing socket. With a modern live PID, `runner_generation` is therefore always `None`, and this return retries until the session is parked without killing the runner or removing its record. Authenticate this missing-socket case under the generation lock using the persisted native process identity, without requiring an unavailable peer, then terminate the exact generation before fresh spawn. If authority is insufficient, publish an explicit terminal state instead of entering a retry loop.

jerome-benoit (changes requested):
Initial review of HEAD `a9cfb098729a3a06e3b0e18d572594778e17ddec`, using two independent analyses followed by targeted cross-validation.

Verdict: changes requested. The three inline comments are P1 findings: escalation loses authority when the group leader exits before its descendants; the drain can remove the replacement fence during `shutdown`; and a live PID with a missing socket can never reach `FreshSpawn`. All three directly violate the teardown and recovery invariants claimed by this PR.

I also independently confirmed three P2 findings already covered by existing threads, so I did not add duplicate inline comments: unbounded `lock_exclusive` can outlive the teardown deadline indefinitely; `runner_generation_has_drifted` performs flock and synchronous I/O directly from three async contexts; and `mark_restart_pending` is written before the generation check, allowing a later stop to be misclassified as a restart during that marker's lifetime.

Explicitly rechecked and rejected on this HEAD: an unbounded handshake gate, late validation of `session/*` methods, `shutdown_all` aborting on the first failure, stale-generation metadata overwrites, and loss of arbitrary JSON-RPC IDs.

Targeted evidence executed: `cargo test --features serve dead_record_fresh_spawns -- --nocapture` and `cargo test --features serve restart_decision_returns_user_stopped_when_registry_deleted -- --nocapture`, both 1/1; Linux reproductions for the orphaned process group and unbounded flock. I did not rerun the full suite.

src/process/runner.rs
```diff
@@ -1774,6 +1810,9 @@ async fn handle_control_connection(
                     }
                 };
                 shared.emit_control(frame).await;
+                if failed {
```
Seluj78: Blocker: this kills the runner on any `EstablishSession` error, which destroys the `session/load` to `session/new` fallback.

`run_or_replay_session` returns `Err` whenever the agent answers `session/load` with a JSON-RPC error (`handshake_result(&response)?`, runner.rs:1315). Returning `true` here makes the arm at runner.rs:524 do `agent_child.start_kill()`. But `src/acp/acp_client.rs:7495` treats that error as recoverable: it logs "session/load failed, falling back to session/new" and calls `establish_session_v2::<NewSessionResponse>(control, "session/new", &req)` on the same `DaemonControlClient` (acp_client.rs:7524). That write hits a dead runner, the spawn fails, and `stored_acp_session_id` is never replaced, so the next respawn repeats the same failure until the #1945 crash budget parks the session.

Rule: correctness. Only daemon cancellation justifies teardown, not an agent-returned handshake error. `Initialize` at :1784 has the same conflation but no fallback depends on it.

src/process/runner.rs
```diff
@@ -518,6 +521,13 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
         _ = accept_loop => {
             warn!(target: "acp.runner", session = %session_id, "accept loop exited unexpectedly");
         }
+        result = &mut control_accept_task => {
+            if !matches!(result, Ok(true)) {
+                warn!(target: "acp.runner", session = %session_id, "control accept loop exited unexpectedly");
+            }
+            let _ = agent_child.start_kill();
```
Seluj78, line 536: Blocker: this arm does not set `preserve_registry`, so the post-loop `worker_registry::delete(&session_id)` at :536 runs unconditionally.

`worker_registry::delete` (`src/process/worker_registry.rs:274`) unlinks `<id>.json`, `<id>.sock`, and `<id>.control.sock` with no pid or generation ownership check. `preserve_registry` exists precisely to stop a superseded runner from doing this.

Combined with the `if failed { return true }` blocker at :1813 the overlap becomes routine: on a `session/load` error the reconciler takes the `Ok(Err(e))` branch at `src/server/acp_reconciler.rs:1419`, deletes the record, then falls straight through to the fresh-spawn block in the same call. R2 binds the sockets and writes the record while R1 is still inside `agent_child.wait().await`; R1 then deletes R2 files. R2 is alive but undiscoverable and unconnectable, so the next tick spawns R3.

This PR removed the timeout-side delete at acp_reconciler.rs:1427 but left the `Ok(Err)` delete at 1419, so cleanup ownership is split rather than moved.

src/process/runner.rs
```diff
@@ -1750,20 +1760,46 @@ async fn handle_control_connection(
         match body {
             ControlBody::Attach { .. } => {}
             ControlBody::Initialize { request } => {
-                let frame = match shared.run_or_replay_initialize(&agent_stdin, request).await {
+                let result = tokio::select! {
+                    result = shared.run_or_replay_initialize(&agent_stdin, request) => result,
+                    closed = control_protocol::read_frame(&mut read_half) => {
```
Seluj78: Required: `control_protocol::read_frame` is not cancel-safe, so it must not be the losing branch of a `tokio::select!`.

`read_frame` (`src/acp/control_protocol.rs:170-185`) is two sequential `read_exact` calls, 4-byte length then body. When the handshake future wins, this future is dropped and any bytes `read_exact` already accumulated are lost. The next `read_frame` in the outer loop starts mid-frame and reads body bytes as a length prefix, tripping `control frame length N exceeds cap` or a serde parse error. `handle_control_connection` then breaks and returns `false`, so the connection dies with no runner teardown.

Reachable today through the `Ok(Some(_))` arm this code itself anticipates. Fix: one task owns `read_half` and forwards decoded frames over an `mpsc`; race the handshake against the cancel-safe `recv()`. Same as CodeRabbit unresolved thread.

src/process/runner.rs
```diff
@@ -1750,20 +1760,46 @@ async fn handle_control_connection(
...
+                }
             }
             ControlBody::EstablishSession { method, request } => {
-                let frame = match shared
-                    .run_or_replay_session(&agent_stdin, &method, request)
-                    .await
-                {
+                let result = tokio::select! {
+                    result = shared.run_or_replay_session(&agent_stdin, &method, request) => result,
+                    closed = control_protocol::read_frame(&mut read_half) => {
```
Seluj78: Required: the reap only happens when cancellation lands inside one of these two `select!` arms.

`ShutdownControlOnDrop` fires whenever the attach future is dropped, but the runner reaction depends on where it is. If the EOF arrives while the outer loop is in `read_frame` at :1750 (before `Initialize` is sent, between `Initialized` and `EstablishSession`, or after `SessionReady` was emitted but before the daemon `wait_for_handshake` resolves), the loop breaks, `handle_control_connection` returns `false`, and the runner plus its agent stay alive.

That last window is #3474 itself: the runner completes `session/load`, emits `SessionReady`, the 3s timeout fires before the response is consumed, and the runner survives holding the provider thread writer. The replacement `session/load` then fails with the active-writer error the issue reports.

The invariant worth establishing is "EOF before the handshake completed means abandoned, terminate", tracked as connection-local state, not "EOF observed inside one of two specific select!s".

tests/acp_runner_control.rs
```diff
@@ -200,6 +232,160 @@ fn find_python3() -> Option<PathBuf> {
     None
 }
 
+#[tokio::test]
```
Seluj78, line 254: Required: this test cannot fail on any of the blockers.

The fake agent delays only `initialize` and returns `{}` success for every other method, so the `session/load`-error fallback is never exercised. `wait_for_runner_exit(&mut old.0, &record, &socket)` blocks until the old runner is fully gone before `spawn_runner("0")`, which serialises away the delete race by construction. And the timeout is on `initialize`, not `session/load`, so the writer-retention scenario is not modelled.

Missing coverage for behavior this PR introduces:
- `session/load` returns an error, daemon must reach `session/new` on the same runner.
- Replacement spawned while the old runner is mid-teardown; both generations files must survive.
- Cancellation during `session/load` rather than `initialize`.

Minor: `wait_for_runner_exit` checks `record` and `<id>.sock` but not `<id>.control.sock`, so a stale control socket would not fail the test.

src/acp/acp_client.rs
```diff
@@ -12,6 +12,7 @@
 //! mpsc channel into ACP requests until shutdown.
 
 use std::collections::{HashMap, VecDeque};
+use std::os::fd::{AsRawFd, RawFd};
```
Seluj78, line 15: Nit: this import is unconditional while every other unix-only import in this file is `#[cfg(unix)]`-gated (3729, 12105, 12124, ...), and `libc` sits under `[target.(cfg(unix)).dependencies]` in Cargo.toml:192. Nothing breaks today (README:147 says native Windows is unsupported, CI is ubuntu/macos only), so this is convention drift rather than a defect.

src/process/runner.rs
```diff
@@ -518,6 +521,13 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
         _ = accept_loop => {
             warn!(target: "acp.runner", session = %session_id, "accept loop exited unexpectedly");
         }
+        result = &mut control_accept_task => {
+            if !matches!(result, Ok(true)) {
```
Seluj78: Nit: near-unreachable. The accept loop only ever exits via `return true`, so `Ok(false)` cannot occur and this warn fires only on a `JoinError`. Worth rewording the message or matching on the join error instead.

Seluj78 (changes requested):

## AoE Review

Reviewed at `af9ae468`, base `main@748137290`. 4 files (acp, runtime, server, rust-tests).

The scope decision is right: #3474 is a real, reproducible stranding bug, and shrinking this from the `a9cfb098` generation-registry redesign to a 4-file fix with the rest deferred to #3487 was the correct call. The direction (revoke the exact runner generation on attach cancellation) is also right, and the RAII fd-shutdown guard is sound as written: the guard holds the `Arc<DaemonControlClient>` that owns the `OwnedWriteHalf`, so the descriptor cannot be closed or reused during the `libc::shutdown` call.

The problem is on the runner side. One `bool` now carries two unrelated meanings, so "the agent said no" is treated the same as "the daemon abandoned me".

### Blockers

**1. Every `EstablishSession` error kills the runner, destroying the `session/load` to `session/new` fallback.** `src/process/runner.rs:1813`

`run_or_replay_session` returns `Err` whenever the agent answers `session/load` with a JSON-RPC error (`handshake_result(&response)?`, runner.rs:1315). The runner emits `HandshakeFailed`, returns `true`, and the new arm at runner.rs:524 does `agent_child.start_kill()`. But the daemon's `Err` handler at `src/acp/acp_client.rs:7495` logs "session/load failed, falling back to session/new" and then calls `establish_session_v2::<NewSessionResponse>(control, "session/new", &req)` on the same `DaemonControlClient` (acp_client.rs:7524). That write hits a runner that has already killed its agent, so the whole spawn errors out and `stored_acp_session_id` is never replaced.

Concrete case: a resumed session whose stored ACP id the agent no longer knows (agent restarted, transcript GC'd). Before this PR, `session/load` fails, `session/new` succeeds on the live runner, session recovers. After, the spawn fails, the reconciler respawns, the same stale id produces the same failure, and it repeats until the #1945 crash budget parks the session.

`Initialize` (runner.rs:1784) has the same conflation, but no fallback depends on it, so it is defensible there.

**2. The runner's unconditional self-delete can unlink a replacement runner's registry record and both sockets.** `src/process/runner.rs:528`

`worker_registry::delete` (`src/process/worker_registry.rs:274`) unlinks `<id>.json`, `<id>.sock`, and `<id>.control.sock` with no pid or generation ownership check. The `preserve_registry` flag exists to stop a superseded runner from doing exactly this, and the new arm never sets it.

The overlap is not theoretical once blocker 1 fires. On a `session/load` error, `resume_one` takes the `Ok(Err(e))` branch at `src/server/acp_reconciler.rs:1419`, calls `worker_registry::delete(&id)`, then falls straight through to the fresh-spawn block in the same call. R2 binds the sockets and writes the record while R1 is still inside `agent_child.wait().await`; R1 then deletes R2's files. R2 is alive but has no registry record and no socket on disk, so nothing can `connect()` to it. The next reconciler tick sees a missing record and spawns R3.

This PR removed the reconciler's timeout-side delete (acp_reconciler.rs:1427) but left the `Ok(Err)` delete at 1419, so ownership of cleanup is now split rather than moved.

**3. CI: `Cargo Test (linux, serve-rest)` is red.**

`server::tests::load_or_generate_token_is_stable_across_restarts_but_rotates_when_idle` panicked at `src/server/mod.rs:9726`, `assertion left == right failed: a restart within the window must reuse the token`. Result: `6362 passed; 1 failed`.

Unrelated to this diff and mtime-backdate sensitive, so this looks like flaky infrastructure rather than a defect here. The merge box is still blocked; please rerun the job. Note the PR's verification block lists `cargo test --test acp_runner_control` but not the full suite, and the full suite is what is failing.

### Required

**4. `read_frame` is not cancel-safe and is used as the losing branch of a `tokio::select!`.** `src/process/runner.rs:1765` and `:1791`

`control_protocol::read_frame` (`src/acp/control_protocol.rs:170-185`) is two sequential `read_exact` calls, one for the 4-byte big-endian length and one for the body. When the handshake future wins the race, the `read_frame` future is dropped and any bytes `read_exact` already accumulated are lost. The next `read_frame` in the outer loop starts mid-frame and reads body bytes as a length prefix, tripping either `control frame length N exceeds cap` or a serde parse error. `handle_control_connection` then breaks and returns `false`, so the connection dies with no runner teardown.

Reachable today via the `Ok(Some(_))` arm the code itself anticipates ("control command arrived before initialize completed"). Fix: one task owns `read_half` and forwards decoded frames over an `mpsc`; race the handshake against the cancel-safe `recv()`. This matches CodeRabbit's still-unresolved thread.

**5. The fix only reaps the runner when cancellation lands inside one of the two handshake `select!` arms.** `src/process/runner.rs:1750`

`ShutdownControlOnDrop` fires whenever the attach future is dropped, but the runner's reaction depends on where it happens to be. If the EOF arrives while the outer loop is in `read_frame` (before `Initialize` is sent, between `Initialized` and `EstablishSession`, or after `SessionReady` was emitted but before the daemon's `wait_for_handshake` resolves), the loop `break`s, `handle_control_connection` returns `false`, the accept loop iterates, and the runner plus its agent stay alive.

That last window is #3474's own description: the runner completes `session/load`, emits `SessionReady`, the daemon's 3s timeout fires before the response is consumed, and the runner survives holding the provider's thread writer. The replacement's `session/load` then fails with the active-writer error the issue reports.

The invariant worth establishing is "EOF before the handshake completed means abandoned, terminate", tracked as connection-local state, rather than "EOF observed inside one of two specific `select!`s".

**6. The new test cannot fail on any of the above.** `tests/acp_runner_control.rs:235`

The fake agent delays only `initialize` and returns `{}` success for every other method, so the `session/load`-error fallback is never exercised. `wait_for_runner_exit(&mut old.0, &record, &socket)` blocks until the old runner is fully gone before `spawn_runner("0")`, which serialises away the delete race by construction. And the timeout is on `initialize`, not `session/load`, so the writer-retention scenario is not modelled.

Behaviors introduced with no coverage:
- `session/load` returns an error, daemon must reach `session/new` on the same runner.
- Replacement spawned while the old runner is mid-teardown; both generations' files must survive correctly.
- Cancellation during `session/load` rather than `initialize`.

Minor: `wait_for_runner_exit` checks `record` and `<id>.sock` but not `<id>.control.sock`, so a stale control socket would not fail the test.

### Nits

- `src/acp/acp_client.rs:15`: `use std::os::fd::{AsRawFd, RawFd};` is unconditional while every other unix-only import in this file is `#[cfg(unix)]`-gated (3729, 12105, 12124, ...), and `libc` sits under `[target.'cfg(unix)'.dependencies]` in Cargo.toml:192. Nothing breaks today (README:147 says native Windows is unsupported, CI is ubuntu/macos only), so this is convention drift rather than a defect.
- `src/process/runner.rs:525`: `if !matches!(result, Ok(true))` is near-unreachable. The accept loop only ever exits via `return true`, so `Ok(false)` cannot occur and the warn fires only on a `JoinError`.

### Two claims checked and dropped

For completeness, two plausible-sounding concerns did not survive verification and are not findings:

- `emit_control` buffering the `HandshakeFailed` frame past process exit. It awaits `write_control_frame` synchronously (runner.rs:1116-1131), so the frame is written before the return.
- `start_kill()` leaking agent grandchildren. True, but the pre-existing `shutdown_signal` arm does the same, so this PR does not introduce it.

### Housekeeping

Six review threads from the pre-rewrite `a9cfb098` version are marked outdated and point at `worker_registry.rs` / `supervisor.rs` / generation-registry code the rewrite deleted. Worth resolving so the PR reads clean.

### Verdict

**Project fit:** Accept. Real bug, right scope, correct deferral to #3487.
**Placement:** core. Runner and session lifecycle is the substrate `docs/plugins.md` names as staying in core.

Separating cancellation from handshake failure and pid-guarding the delete keeps this PR small and fixes both blockers.

PR metadata is otherwise clean: conventional-commit title, no em dashes, PR Type set, Test Coverage Analysis and AI Usage filled, branch naming correct, `Fixes #3474` present.

src/process/runner.rs
```diff
@@ -518,6 +521,13 @@ pub async fn run(args: AcpRunnerArgs) -> Result<()> {
         _ = accept_loop => {
             warn!(target: "acp.runner", session = %session_id, "accept loop exited unexpectedly");
         }
+        result = &mut control_accept_task => {
```
Seluj78, line 524: **Blocker:** this arm does not set `preserve_registry`, so teardown falls through to the unguarded `worker_registry::delete(&session_id)` at lines 536-538 and can unlink a *replacement* runner's record and both sockets.

`preserve_registry` (line 466) exists precisely to stop a superseded runner from deleting a replacement's files, and it is set only by the watchdog `Superseded` reason at line 516.

Interleaving, reachable on any handshake error:

1. `run_or_replay_*` returns `Err`, this runner emits `HandshakeFailed` and returns `true` (line 1784).
2. The daemon's `attach` errors, the reconciler enters `Ok(Err(e))` at `src/server/acp_reconciler.rs:1414-1421`, deletes the record, and falls through **inline in the same call** to `build_spawn_request` plus `acp_supervisor.spawn`.
3. The replacement runner writes `<id>.json` and binds `<id>.sock` and `<id>.control.sock`.
4. This runner finishes `start_kill()` plus `wait()` on its Node agent and only then calls `delete(&session_id)`, unlinking the replacement's files.

`worker_registry::delete` unlinks all three paths by session id with no owner check, so both processes are path-based deleters with no ownership token. The replacement's abandonment watchdog (#1921) then sees its record missing and self-destructs, so the session burns the 5-in-60s reconciler budget and parks with an `AgentStartupError`, which is the failure class #3474 set out to remove.

Dropping the `if failed` branch at line 1784 removes the reachable path into this. A longer-term fix is an ownership-checked `delete_if_owner(&id, our_pid)` that retires the `preserve_registry` special case, but that belongs in #3487.

src/process/runner.rs
Seluj78: **Required:** `read_frame` is not cancellation safe, so using it as a `select!` branch desyncs the framing on `read_half`. Same issue in the `EstablishSession` block at line 1791.

`control_protocol::read_frame` (`src/acp/control_protocol.rs:170`) is two sequential `AsyncReadExt::read_exact` calls into local buffers, and tokio documents `read_exact` as not cancellation safe. When `run_or_replay_initialize` wins, `select!` drops this future and any bytes it already consumed are lost. With 1 to 3 bytes of a length prefix consumed, or the full prefix plus part of a body, the outer-loop `read_frame` at line 1745 reads body bytes as a length: either a bogus length that trips `MAX_CONTROL_FRAME_BYTES`, or a misframed command. Both exit via `break` to `return false`, which keeps the runner **alive** rather than reaping it, so a desync lands on the non-teardown path.

`DaemonControlClient::initialize` / `establish_session` (`src/acp/acp_client.rs:4077`, `4097`) serialize on `handshake_rx.recv()` today, so a conforming daemon does not pipeline. Nothing in the wire protocol enforces that, and the `Ok(Some(_))` arm three lines down exists precisely because an early frame is considered possible.

Fix: let one task own `read_half`, decode frames, and forward them over an `mpsc`; race the handshake against cancel-safe `recv()`. This matches the live CodeRabbit thread on this file.

src/process/runner.rs
```diff
@@ -1750,20 +1760,46 @@ async fn handle_control_connection(
...
+                let failed = result.is_err();
+                let frame = match result {
                     Ok(result) => ControlBody::Initialized { result },
                     Err(error) => {
                         warn!(target: "acp.runner", session = %session_id, "initialize failed: {error}");
                         ControlBody::HandshakeFailed { error }
                     }
                 };
                 shared.emit_control(frame).await;
+                if failed {
```
Seluj78: **Required:** SIGKILLing the agent on a handshake *error* is behavior the PR body does not claim, and it is what makes the line 524 unlink race reachable.

Before this change, a failed `initialize` or `session/load` emitted `HandshakeFailed` and the loop continued, leaving the runner and its warm agent (plus the `run_or_replay_*` replay cache in `RunnerShared`) available for the daemon to re-drive. Now every `Err` returns `true`, which resolves the accept task, SIGKILLs the agent at lines 528-529, and deletes the registry. That fires on transient request-scoped errors: a provider 401 or 429 during `initialize`, an invalid stored `session/load` id, a version-skew request.

The PR body scopes this to "make the runner terminate its agent when control disappears during initialize or session establishment". A handshake error is not control disappearing. Tie teardown to control-socket loss only, meaning the `Ok(None)` and `Err(e)` arms of the `select!` above. Same branch at line 1814 for `EstablishSession`.

src/server/acp_reconciler.rs
```diff
@@ -1424,7 +1424,6 @@ async fn resume_one(state: Arc<AppState>, target: ResumeTarget) -> ResumeOutcome
                         session = %id,
                         "attach timed out after 3s; falling back to fresh spawn"
                     );
-                    crate::process::worker_registry::delete(&id).ok();
                     return ResumeOutcome::RetryAfterAttachTimeout;
```
Seluj78, line 1428: **Required:** dropping `worker_registry::delete(&id).ok();` from this arm makes recovery conditional on `control_client` being `Some`, and nothing checks that precondition.

`ShutdownControlOnDrop(None)` is a no-op (`src/acp/acp_client.rs:3206-3213`), and `connect_runner_control_v2` returns `None` for "an absent or older (v1) runner" per its own doc comment at `src/acp/acp_client.rs:4159`, plus a failed dial, a Hello timeout, or a version mismatch. In that case:

- nothing reaps the runner;
- `AcpClient::attach` has already spawned `run_connection_task` at `src/acp/acp_client.rs:2776` *before* `wait_for_handshake`, and the outer 3s timeout does not abort it;
- the record survives, so `attempted.remove(&id)` at line 528 re-arms an attach against the same wedged runner every 2s tick until the 5-in-60s budget parks the session with a user-visible `AgentStartupError`.

The deleted line was the only forcing function turning a repeatedly timing-out attach into a clean fresh spawn. "Leave timed-out registry cleanup to the runner that owns it" holds only when a v2 control channel exists.

Either restore the delete, or make it an ownership-checked `delete_if_owner(&id, record.pid)` on both sides so the daemon can sweep without racing the runner.

tests/acp_runner_control.rs
```diff
@@ -200,6 +232,160 @@ fn find_python3() -> Option<PathBuf> {
...
+            "fake-agent".into(),
+            None,
+        ),
+    )
+    .await;
+    assert!(
+        attach.is_err(),
+        "delayed initialize must exceed the attach budget"
+    );
+    wait_for_runner_exit(&mut old.0, &record, &socket);
```
Seluj78: **Required:** this wait serializes away the race the test name asserts, so the test cannot fail if the line 524 unlink race regresses.

`wait_for_runner_exit` blocks until the old runner has exited *and* its record and socket are gone, before `spawn_runner("0")` starts the replacement. Production never waits: the reconciler's `Ok(Err(e))` arm fresh-spawns inline while the old runner is still in `start_kill()` plus `wait()`.

The test also delays only `initialize`, while #3474's stated root cause is a late `session/load` retaining the provider thread writer, and the Python fake agent has no writer concept, so the active-writer failure mode cannot be reproduced.

What the test does prove (timeout plus v2 control means the old runner and agent die, and the replacement's `session/load` is the only one) is worth keeping. Missing cases:

- a replacement spawned concurrently with old-runner teardown, asserting the replacement's record and sockets survive;
- a `control_client == None` timeout, asserting recovery;
- a timeout while `session/load` rather than `initialize` is pending.

src/process/runner.rs
Seluj78: **Nit:** `Ok(false)` is unreachable, so this check only ever fires on a panic and "control accept loop exited unexpectedly" misdescribes it.

The accept loop at lines 363-394 has no `break` and no `return false`; a `false` from `handle_control_connection` continues the loop, so the task can only resolve via `return true` or a panic. A two-variant enum (`Detached` / `TerminateRunner`) would make the lifecycle contract legible and drop the impossible branch.

src/acp/acp_client.rs
```diff
@@ -4040,6 +4055,7 @@ struct DaemonControlClient {
     write: Mutex<tokio::net::unix::OwnedWriteHalf>,
     handshake_rx: Mutex<mpsc::Receiver<ControlBody>>,
     completion: Arc<std::sync::Mutex<Option<oneshot::Sender<control_protocol::PromptOutcome>>>>,
+    raw_fd: RawFd,
```
Seluj78, line 4061: **Nit:** `raw_fd` duplicates state the adjacent `Mutex<OwnedWriteHalf>` already owns, and the `libc::shutdown` return at line 3210 is discarded, so `EBADF` and `ENOTCONN` are invisible.

The invariant is sound today: the guard holds the `Arc` across the `unsafe` call, which keeps the write half and therefore the fd alive, and `shutdown` is not `close`, so no descriptor recycling is possible. It is non-local, though: a later refactor that replaces or takes the write half silently invalidates the stored fd with nothing in the type system objecting.

Seluj78 (changes requested):

## Review summary

Reviewed `af9ae468` against `main@7481372`. 4 files, 255 additions. Surfaces: acp, runtime, server, rust-tests.

The problem is real and well evidenced, the diagnosis in #3474 is right, and the cancel signal shape (an attach-scoped guard that closes that exact runner's control socket) is the correct primitive. Shrinking the change from `a9cfb098` down to 255 lines was the right call. Two of the three production changes overreach past the stated fix, though, and each one opens a new lifecycle hole. Both are one-line reversions.

Findings are anchored inline. Full reasoning below.

### Blocker: the new teardown path can unlink a replacement runner's registry record and sockets

`src/process/runner.rs:524-538`

`preserve_registry` (declared at `src/process/runner.rs:466`) exists specifically to stop a superseded runner from deleting a replacement's files, and it is set only by the watchdog `Superseded` reason at `src/process/runner.rs:516`. The new `control_accept_task` select arm does not set it, so teardown falls through to `worker_registry::delete(&session_id)` at `src/process/runner.rs:536-538`, which unlinks `<id>.json`, `<id>.sock` and `<id>.control.sock` by path with no owner check.

Interleaving, reachable on any handshake error:

1. Old runner's `run_or_replay_*` returns `Err`, emits `HandshakeFailed`, returns `true` (`src/process/runner.rs:1784`).
2. The daemon's `attach` errors, the reconciler enters the `Ok(Err(e))` arm at `src/server/acp_reconciler.rs:1414-1421`, deletes the record, and falls through **inline in the same call** to `build_spawn_request` plus `acp_supervisor.spawn`.
3. The replacement runner writes `<id>.json` and binds both sockets.
4. The old runner finishes `start_kill()` plus `wait()` on its Node agent and only then calls `delete(&session_id)`, unlinking the replacement's files.

The replacement's own abandonment watchdog (#1921) then sees its record missing and self-destructs, so the session burns the 5-in-60s reconciler budget and parks with an `AgentStartupError`. That is the failure class #3474 set out to remove.

Both processes are path-based deleters with no ownership token, and the existing `preserve_registry` guard covers only one of the two paths that now reach the delete.

### Blocker: required check red

`Cargo Test (linux, serve-rest)` fails on `server::tests::load_or_generate_token_is_stable_across_restarts_but_rotates_when_idle`:

```
assertion `left == right` failed: a restart within the window must reuse the token
  at src/server/mod.rs:9726
test result: FAILED. 6362 passed; 1 failed
```

Unrelated to this diff (auth token rotation window; the PR touches no server auth code), so it reads as a pre-existing flake rather than something you introduced. It still blocks merge under `AGENTS.md` Definition of done item 1. Needs a re-run, not a code change.

### Required: `read_frame` as a `select!` branch is not cancellation safe

`src/process/runner.rs:1765` and `src/process/runner.rs:1791`

`control_protocol::read_frame` (`src/acp/control_protocol.rs:170`) is two sequential `AsyncReadExt::read_exact` calls into local buffers, and tokio documents `read_exact` as not cancellation safe. When `run_or_replay_initialize` / `run_or_replay_session` wins, `select!` drops the read future and any bytes it already consumed are gone. With 1 to 3 bytes of a length prefix consumed, or the full prefix plus part of a body, the outer-loop `read_frame` at `src/process/runner.rs:1745` reads body bytes as a length: either a bogus length that trips the `MAX_CONTROL_FRAME_BYTES` cap, or a misframed command. Both exit via `break` to `return false`, which keeps the runner alive rather than reaping it, so a desync lands on the non-teardown path.

Today's `DaemonControlClient::initialize` / `establish_session` (`src/acp/acp_client.rs:4077`, `4097`) serialize on `handshake_rx.recv()`, so a conforming daemon does not pipeline. Nothing in the wire protocol enforces that, and the code's own `Ok(Some(_))` arm exists precisely because an early frame is considered possible. A dedicated reader task feeding an `mpsc`, with `select!` on cancel-safe `recv()`, fixes it. This corroborates the one live CodeRabbit thread on the PR.

### Required: SIGKILLing the agent on a handshake error is behavior the PR does not claim

`src/process/runner.rs:1784`

Before this change, a failed `initialize` or `session/load` emitted `HandshakeFailed` and the loop continued, leaving the runner and its warm agent (plus the `run_or_replay_*` replay cache in `RunnerShared`) available for the daemon to re-drive. Now every `Err` returns `true`, which resolves the accept task, SIGKILLs the agent at `src/process/runner.rs:528-529`, and deletes the registry. That fires on transient request-scoped errors: a provider 401 or 429 during `initialize`, an invalid stored `session/load` id, a version-skew request.

The PR body scopes the change to "make the runner terminate its agent when control disappears during initialize or session establishment". A handshake error is not control disappearing. Tie teardown to control-socket loss only (the `Ok(None)` and `Err(e)` select arms). Dropping this one branch also removes the reachable path into the Blocker above.

### Required: removing the reconciler timeout sweep makes recovery conditional on `control_client` being `Some`

`src/server/acp_reconciler.rs:1424` (deleted line)

`ShutdownControlOnDrop(None)` is a no-op (`src/acp/acp_client.rs:3206-3213`), and `connect_runner_control_v2` returns `None` for "an absent or older (v1) runner" per its own doc comment at `src/acp/acp_client.rs:4159`, plus a failed dial, a Hello timeout, or a version mismatch. In that case:

- nothing reaps the runner;
- `AcpClient::attach` has already spawned `run_connection_task` at `src/acp/acp_client.rs:2776` before `wait_for_handshake`, and the outer 3s timeout does not abort it;
- the record survives, so `attempted.remove(&id)` (`src/server/acp_reconciler.rs:528`) re-arms an attach against the same wedged runner every 2s tick until the 5-in-60s budget parks the session with a user-visible `AgentStartupError`.

The deleted line was the only forcing function turning a repeatedly timing-out attach into a clean fresh spawn. The PR body's rationale ("leave timed-out registry cleanup to the runner that owns it") holds only when a v2 control channel exists, and that precondition is never checked. Either keep the delete, or make it an ownership-checked `delete_if_owner(&id, record.pid)` on both sides so the daemon can sweep without racing the runner.

### Required: the new test serializes away the race its name asserts

`tests/acp_runner_control.rs:339`

`cancelled_attach_reaps_runner_before_replacement_loads_session` blocks until the old runner has exited and its record and socket are gone before it calls `spawn_runner("0")` for the replacement. Production never waits: the reconciler's `Ok(Err(e))` arm fresh-spawns inline while the old runner is still in `start_kill()` plus `wait()`. The test also delays only `initialize`, while #3474's stated root cause is a late `session/load` retaining the provider thread writer, and the Python fake agent has no writer concept, so the active-writer failure mode cannot be reproduced.

What it does prove (timeout plus v2 control means the old runner and agent die, and the replacement's `session/load` is the only one) is worth keeping. Missing cases:

- a replacement spawned concurrently with old-runner teardown, asserting the replacement's record and sockets survive;
- a `control_client == None` timeout, asserting recovery;
- a timeout while `session/load` rather than `initialize` is pending.

### Security: no finding

Three concerns were raised and checked against the code; all three are refuted.

- **The `RawFd` plus `libc::shutdown` cannot hit a recycled descriptor.** `ShutdownControlOnDrop` takes the `Option<Arc<DaemonControlClient>>` and holds that `Arc` across the `unsafe` call (`src/acp/acp_client.rs:3208-3211`). The `Arc` keeps `Mutex<OwnedWriteHalf>` alive, which keeps the socket fd open. `shutdown` is not `close`, so no recycling is possible, and the separately spawned reader task owning the `OwnedReadHalf` of the same `into_split` socket cannot close it independently. `SHUT_RDWR` rather than `SHUT_WR` is the right choice: it is what unblocks that reader task. The SAFETY comment states the correct invariant.
- **The widened local DoS is same-UID only.** A peer that connects to `<id>.control.sock`, sends `Initialize`, then disconnects can now kill the runner and agent. The worker directory is `0o700` (`src/process/worker.rs:255`) and records are `0o600` (`src/process/worker_registry.rs:181`), so the boundary is same-UID, which the ACP control channel already trusts unauthenticated.
- **`HandshakeFailed` does reach the daemon.** `emit_control` awaits `write_control_frame` to completion before `return true`, so the bytes are in the kernel socket buffer and survive process exit for an AF_UNIX stream whose peer is still reading, which is the case on the `Ok(Err(e))` path.

### PR metadata: clean

Template filled, PR Type set to Bug Fix, Test Coverage Analysis N/A boxes justified with reasons, AI Usage complete, branch `fix/acp-runner-handshake-timeout` matches the `fix/` convention, conventional-commit title, zero em dashes in title, body, commit message, or diff. No dead-code allows, no untagged `tracing!` calls, no `format!` inside a tracing macro, no secrets logged, no hardcoded namespace strings, no docs or settings surfaces touched.

Six of the seven inline threads on this PR are outdated: they target the generation-registry redesign that `af9ae468` removed. Your note at issuecomment-5383831654 is right that they are obsolete; they just need resolving on GitHub so the live CodeRabbit cancel-safety thread is the only one left.

### Verdict

**Project fit:** Accept with scope reduction
**Placement:** n/a (bug fix, core runner lifecycle)

Suggested landing shape: keep the `ShutdownControlOnDrop` guard and the control-loss teardown, revert the `if failed` teardown and the reconciler delete removal, and move the `select!` off `read_frame` onto a channel-fed reader. That is the minimal #3474 fix without the new races. An ownership-checked `worker_registry::delete_if_owner(&id, pid)` would let the daemon and runner both clean up safely and would retire the `preserve_registry` special case, but that belongs in #3487.

njbrake (commented):
Not adding a third changes-requested, the box is already blocked. Three things the existing reviews do not cover.

**The red CI is one failure, not two, and your new test has never run.** `Cargo Test (linux)` is a roll-up gate whose entire script is `echo "linux legs result: failure"; exit 1`, so it mirrors the leg rather than testing anything. The real failure is `server::tests::load_or_generate_token_is_stable_across_restarts_but_rotates_when_idle` at `src/server/mod.rs:9726`, a `--lib` unit test over a tempdir in auth token loading, and this diff touches no server auth code. `load_or_generate_token_at` returns the same token on every path that reaches its reuse branch, so the failure means the branch was skipped, which is mtime sensitive; the test's own comment concedes it uses a generous window for that reason. More useful than the red itself: the step runs bare `cargo test --features serve`, `--lib` failed, and cargo aborted before any integration target, so `tests/acp_runner_control.rs` has never executed in CI on `af9ae468`. I have re-run the leg.

**Both blockers reproduce independently.** I traced them before reading @Seluj78's review, so they are not one reader's interpretation. On the `session/load` error path: `handshake_result` returns Err at runner.rs:1415, `run_or_replay_session` propagates at :1315, the new arm returns true and kills the agent, and the daemon then calls `establish_session_v2` for `session/new` on that same control client at acp_client.rs:7528.

**Blocker 2 is sharper than a missing flag.** The unconditional delete does not merely omit `preserve_registry`; it bypasses the mechanism that makes main's delete safe. The watchdog's `RecordMissing` branch is unconditional too, but it is deliberately slow, N consecutive misses at the poll interval, so a replacement has time to claim the record and the state reads `Superseded` instead. The new arm deletes immediately with no ownership re-check. Gating the post-loop delete on `inspect_record_for_runner(...) == Matches` fixes it without a flag a future path can forget.

**Housekeeping:** `RunnerGeneration`, `terminate_runner_generation` and `runner_generation_has_drifted` exist neither on main nor in this diff, so the earlier P1s are moot rather than unanswered and those six threads can be resolved.

src/process/runner.rs
aaiyer: Fixed in dd831e57. Normal control-task completion now sets preserve_registry before agent teardown, so the canceled generation never performs the post-loop path deletion. The regression starts the replacement while the old initialize is still pending, waits for the old runner to exit, and verifies the replacement record plus both sockets remain.

src/acp/acp_client.rs
aaiyer: Leaving this scoped nit unchanged. Drop cannot await the write mutex; the held Arc keeps OwnedWriteHalf and the fd alive across shutdown, and EBADF or ENOTCONN are already terminal for this best-effort cancellation. A broader typed lifetime redesign remains in #3487.

tests/acp_runner_control.rs
aaiyer: Addressed in dd831e57: the replacement is spawned before the old runner exits; its record, relay socket, and control socket are asserted after old teardown. The replacement returns a real session/load error, stays on the same runner, and succeeds through session/new; a split control frame covers partial reads. The constructor timeout is intentionally initialize-only: AcpClient::attach uses ConnectMode::Resume and never sends session/load, while fresh session/load is deliberately outside the ready timeout for large history replay under the #2276 contract. The test also proves the timed-out old runner contributes no extra session/load, covering the #3474 late-writer failure.

src/acp/acp_client.rs
aaiyer: Leaving this scoped nit unchanged. This module already unconditionally uses tokio UnixStream and OwnedWriteHalf throughout the production control path, and native Windows is unsupported, so gating only the fd import would not create a supported non-Unix build boundary.

src/process/runner.rs
aaiyer: Fixed in dd831e57. Normal control-task completion now sets preserve_registry before agent teardown, so the canceled generation never performs the post-loop path deletion. The regression starts the replacement while the old initialize is still pending, waits for the old runner to exit, and verifies the replacement record plus both sockets remain.

njbrake (approved):
Approving at `efccee4a`. All 26 checks are green, including `Cargo Test (linux, serve-rest)`, the leg that actually runs `tests/acp_runner_control.rs`. On `af9ae468` that leg never reached the integration targets, so this is the first run where the new regression test has executed in CI.

I verified both previously blocking findings against the source rather than the PR description:

- The `session/load` to `session/new` fallback works. `run_or_replay_session` caches only on success (`src/process/runner.rs:1316`), so a failed load leaves the cache empty and the fallback runs fresh against the agent. The new test asserts `handshake_failed` and then `session_ready`.
- `preserve_registry` is now set on the control termination arm (`src/process/runner.rs:527`), so a cancelled generation cannot unlink a replacement's record or sockets.

I also reproduced the load-bearing primitive in isolation: `libc::shutdown(fd, SHUT_RDWR)` on one end of an AF_UNIX stream does return EOF to the peer's read. Locally the new test passed 10 of 10 runs, so its timing constants look stable.

I pushed one commit, `efccee4a`, trimming the pid file before parsing in `wait_for_u32`. Test only.

Residual risk is the concurrent handshake serialization deferred to #3487: the original symptom stays reachable in a narrower window, and nothing here surfaces it proactively. Worth keeping #3487 staffed.

Issue #1057, opened 2026-05-12
Cockpit: group consecutive tool calls (between agent text) into one collapsible block — matches Claude Code CLI density


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

A turn with lots of tool calls blows the chat scroll way out. Every `Read`, `Grep`, `Bash`, etc. call renders as its own card via the kind-dispatch switch at `web/src/components/cockpit/ToolCards.tsx:34-53`, so a turn that does e.g. "read 14 files, grep 6 patterns, run 3 bash checks" before saying anything surfaces as **23 stacked cards**, each with its own header and (collapsed) body. Scrolling past it to find the next thing the assistant said is genuinely painful, and it's the most common shape of agent work (long silent investigation phases > brief explanatory text).

The Claude Code CLI handles this nicely: a run of tool calls that aren't broken up by any assistant text folds into one summary block. The boundary is **text**, not kind — as soon as the agent says something (even one sentence), the next tool call starts a fresh group. The information is intact; the visual density is right. The web cockpit could do the same.

## What's there today

- `ToolCard` in `web/src/components/cockpit/ToolCards.tsx:34-53` dispatches per-kind: `execute` / `read` / `edit` / `delete` / `search` / `fetch` / `think` / default → one component per call, no awareness of siblings.
- `activityToThreadMessages` in `web/src/components/cockpit/CockpitRuntime.tsx:124-203` builds the linear message tree. Each `tool_start` row becomes its own `tool-call` content part on the current assistant message. Tool completes are merged into their matching start (lines 180-185) — but no grouping across distinct tool calls.
- Kinds arrive from the backend mapper `tool_kind_str` at `src/cockpit/acp_client.rs:881-893`, which projects ACP's `ToolKind` enum to the lowercase strings the renderer dispatches on. Consecutive same-kind tool starts already share this label.

So everything needed for grouping (a stable per-tool kind, a known order of arrival, the completion linkage) is in place. There's just no consolidation pass between the activity log and the message tree.

## Proposal

Three layers; A is the high-value piece.

### A. Collapse runs of tool calls between text boundaries into one group

Insert a grouping pass inside `AssistantBuilder.build()` where tool-call parts are emitted (cleaner than a post-pass on the message tree). The pass:

1. Walks the part list of each assistant message in order.
[...]

PR #3478, opened 2026-08-22
fix(web): cap tool-call group size and memoize group children

### LINKED ISSUE #3477: Web structured view: Kimi turns collapse all tool calls into one slow-loading group card
## Problem

When using Kimi Code in the web dashboard's structured view, every action the agent performs in a single turn is folded into one collapsible "N actions" card. Loading or re-rendering the view is slow because that single card carries the entire turn's tool-call payload, which is eagerly re-parsed on every render.

Other agents (e.g. Claude) usually interleave small text chunks between tool calls, so consecutive runs stay short and produce small groups. Kimi's `kimi acp` adapter appears to stream all tool calls in one batch before emitting the turn's response text, so the whole turn satisfies the grouping threshold and becomes one giant group.

## Evidence

- Observable in the web dashboard (`/sessions/<id>`) when Kimi runs a turn with many tool calls (reads/edits/executes). The transcript shows a single "N actions" card instead of several smaller groups or individual cards.
- The native TUI structured view (`aoe acp attach`) does **not** group tool calls; the same session renders each tool call as its own card.

## Artifacts & Reproduction

1. Start a session with Kimi (`aoe add --cmd kimi` or the web wizard).
2. Send a prompt that causes many tool calls in one turn, e.g. a multi-file refactor or broad search/read/edit cycle.
3. Open the web structured view and note the single "N actions" card.
4. Expand the card; the browser pauses while React mounts all child tool cards at once.

A unit-level repro: feed `activityToThreadMessages` a sequence of >=10 consecutive `tool_start` rows with no `message` row between them. The output contains a single `_aoe_tool_group` part whose `argsText` JSON holds every child.

## How It Works (moving parts)

- `web/src/components/acp/AcpRuntime.tsx:882` defines `const TOOL_GROUP_MIN_RUN = 3;`.
- `web/src/components/acp/AcpRuntime.tsx:938` `collapseToolRuns()` walks the assistant message's parts and folds any run of >=3 consecutive `tool-call` parts into one synthetic `_aoe_tool_group` part.
- `web/src/components/acp/StructuredView.tsx:1296` `AssistantToolGroup` parses the synthetic part's `{ children: [...] }` payload and hands the reconstructed items to `ToolGroupCard`.
- `web/src/components/acp/ToolCards.tsx:1441` `ToolGroupCard` renders the "N actions" header and expands to show every child.
- `AssistantToolGroup` currently parses and reconstructs every child on every render, even when the group is collapsed.

The grouping threshold is intentional (#1057) for "silent investigation" shapes, but it has no upper bound, so a long batch from Kimi becomes one monolithic card and payload.

## Definition of Done

- A turn with many consecutive tool calls does not produce a single group card containing all of them. Long runs should be split into bounded-size groups (or fall back to inline cards once a run exceeds a reasonable size).
- Re-renders of the structured view do not re-parse the full group children JSON. Child reconstruction should be memoized per payload.
- Existing grouping behavior for short runs (3-10 consecutive tool calls) is preserved.
- Tests cover long runs and the memoization/performance path.

## Watch Out For

- Don't regress #2802: group part ids are anchored to the first child so appending a new tool call doesn't remount the card and re-collapse it.
- Don't change the native TUI; it intentionally renders each tool row separately.
- Keep the change agent-agnostic if possible. The issue is the run length, not Kimi specifically.
- Existing `AcpRuntime.test.ts` grouping tests must still pass.

karljuhlthefool:
## Description

Fixes #3477.

When Kimi streams many tool calls in one turn without interleaving text, the web structured view folds the entire turn into a single "N actions" group card. That card carries a huge JSON payload and re-parses all children on every render, making the view slow to load and re-render.

This change:

1. Caps generic tool-call grouping in `collapseToolRuns` at `TOOL_GROUP_MAX_RUN = 10`. Runs longer than that are split into multiple bounded groups (any trailing chunk smaller than `TOOL_GROUP_MIN_RUN` falls back to inline cards).
2. Memoizes child reconstruction in `AssistantToolGroup` and `AssistantTodoGroup` with `useMemo` keyed by `argsText`, so re-renders don't re-parse the full children payload.

Short "silent investigation" runs (3–10 tools) still group exactly as before. The native TUI is unaffected.

## PR Type

- [x] Bug Fix
- [ ] New Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Added or updated a Vitest spec (`src/components/acp/AcpRuntime.test.ts`) for the new chunking behavior. The existing `coverage-matrix.json` entries already cover `AcpRuntime.tsx` and `StructuredView.tsx`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Kimi Code CLI / Agent of Empires session.

## Summary

The web structured view now splits long consecutive tool-call runs into groups of up to 10 items. Short runs retain their existing grouping, and small trailing runs remain inline.

Tool-group child reconstruction now uses memoization to reduce repeated parsing during re-renders. Group IDs remain anchored to the first child, and native TUI rendering is unchanged.

Tests verify tool-call order, completeness, and IDs for runs of 10, 11, and 25 calls. This prevents oversized group cards and improves rendering performance.

### DIFF
--- a/web/src/components/acp/AcpRuntime.test.ts
+++ b/web/src/components/acp/AcpRuntime.test.ts
@@ -106,6 +106,44 @@ describe("activityToThreadMessages; tool-call grouping (#1057)", () => {
     expect(toolParts[0]!.toolName).toBe(TOOL_GROUP_NAME);
   });
 
+  it("chunks long runs into bounded groups (#3477)", () => {
+    const runOf = (n: number): ActivityRow[] => [
+      userRow("go"),
+      ...Array.from({ length: n }, (_, i) => toolStart(`t${i}`)),
+    ];
+    const groupsIn = (n: number) => {
+      const messages = activityToThreadMessages(runOf(n), false);
+      const assistant = messages.find((m) => m.role === "assistant")!;
+      const parts = assistant.content as Array<{
+        type: string;
+        toolName?: string;
+        argsText?: string;
+      }>;
+      return {
+        groups: parts.filter((p) => p.type === "tool-call" && p.toolName === TOOL_GROUP_NAME),
+        inline: parts.filter((p) => p.type === "tool-call" && p.toolName !== TOOL_GROUP_NAME),
+      };
+    };
+
+    // At the max chunk size, still one group.
+    const ten = groupsIn(10);
+    expect(ten.groups).toHaveLength(1);
+    expect(ten.inline).toHaveLength(0);
+    expect(JSON.parse(ten.groups[0]!.argsText!).children).toHaveLength(10);
+
+    // Just over the max: first chunk groups, tail is too short to group.
+    const eleven = groupsIn(11);
+    expect(eleven.groups).toHaveLength(1);
+    expect(eleven.inline).toHaveLength(1);
+
+    // Long enough to fill multiple chunks.
+    const twentyFive = groupsIn(25);
+    expect(twentyFive.groups).toHaveLength(3);
+    expect(twentyFive.inline).toHaveLength(0);
+    const grouped = twentyFive.groups.reduce((sum, g) => sum + JSON.parse(g.argsText!).children.length, 0);
+    expect(grouped).toBe(25);
+  });
+
   it("does not group runs of 1 or 2 tool calls", () => {
     const messages = activityToThreadMessages([userRow("go"), toolStart("t1"), toolStart("t2")], false);
     const assistant = messages.find((m) => m.role === "assistant")!;
--- a/web/src/components/acp/AcpRuntime.test.ts
+++ b/web/src/components/acp/AcpRuntime.test.ts
@@ -129,19 +129,40 @@ describe("activityToThreadMessages; tool-call grouping (#1057)", () => {
     const ten = groupsIn(10);
     expect(ten.groups).toHaveLength(1);
     expect(ten.inline).toHaveLength(0);
-    expect(JSON.parse(ten.groups[0]!.argsText!).children).toHaveLength(10);
+    const tenChildren = JSON.parse(ten.groups[0]!.argsText!).children as Array<{ toolCallId: string }>;
+    expect(tenChildren.map((c) => c.toolCallId)).toEqual(["t0", "t1", "t2", "t3", "t4", "t5", "t6", "t7", "t8", "t9"]);
 
     // Just over the max: first chunk groups, tail is too short to group.
     const eleven = groupsIn(11);
     expect(eleven.groups).toHaveLength(1);
     expect(eleven.inline).toHaveLength(1);
+    const elevenChildren = JSON.parse(eleven.groups[0]!.argsText!).children as Array<{ toolCallId: string }>;
+    expect(elevenChildren.map((c) => c.toolCallId)).toEqual([
+      "t0",
+      "t1",
+      "t2",
+      "t3",
+      "t4",
+      "t5",
+      "t6",
+      "t7",
+      "t8",
+      "t9",
+    ]);
+    expect((eleven.inline[0] as { toolCallId?: string }).toolCallId).toBe("t10");
 
     // Long enough to fill multiple chunks.
     const twentyFive = groupsIn(25);
     expect(twentyFive.groups).toHaveLength(3);
     expect(twentyFive.inline).toHaveLength(0);
-    const grouped = twentyFive.groups.reduce((sum, g) => sum + JSON.parse(g.argsText!).children.length, 0);
-    expect(grouped).toBe(25);
+    const chunkSizes = twentyFive.groups.map(
+      (g) => (JSON.parse(g.argsText!).children as Array<{ toolCallId: string }>).length,
+    );
+    expect(chunkSizes).toEqual([10, 10, 5]);
+    const allIds = twentyFive.groups.flatMap((g) =>
+      (JSON.parse(g.argsText!).children as Array<{ toolCallId: string }>).map((c) => c.toolCallId),
+    );
+    expect(allIds).toEqual(Array.from({ length: 25 }, (_, i) => `t${i}`));
   });
 
   it("does not group runs of 1 or 2 tool calls", () => {
--- a/web/src/components/acp/AcpRuntime.tsx
+++ b/web/src/components/acp/AcpRuntime.tsx
@@ -881,6 +881,12 @@ function collapseSubagents(parts: DraftPart[], profile: AgentProfile): DraftPart
  *  block (#1057). */
 const TOOL_GROUP_MIN_RUN = 3;
 
+/** Maximum run length that stays in a single group card. Longer runs are
+ *  chunked so one agent turn (e.g. Kimi streaming many tool calls in a
+ *  single batch) doesn't collapse the whole turn into one enormous,
+ *  slow-loading card. See #3477. */
+const TOOL_GROUP_MAX_RUN = 10;
+
 /** Synthetic toolName used for the folded group card. Namespaced with
  *  the `_aoe_` prefix so it can't collide with a real ACP tool kind. */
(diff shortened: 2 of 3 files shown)


web/src/components/acp/AcpRuntime.test.ts
```diff
@@ -106,6 +106,65 @@ describe("activityToThreadMessages; tool-call grouping (#1057)", () => {
     expect(toolParts[0]!.toolName).toBe(TOOL_GROUP_NAME);
   });
 
+  it("chunks long runs into bounded groups (#3477)", () => {
```
Seluj78, line 109: **Required:** chunked-group key stability (the #2802 invariant) is untested.

The existing #2802 test at `AcpRuntime.test.ts:304` only asserts `group-t1` stays stable as a run grows 3 to 4, which is chunk 0. The new loop introduces a second class of key, `group-t10` / `group-t20`, whose stability rests on a new and unstated invariant: chunk boundaries are index-aligned from the run start, so chunk N's first child never changes on append. Nothing asserts it.

A future tweak to the chunk loop (balancing 11 into 6+5 instead of 10+1, or seeding `i` off the last flush) silently reintroduces the #2802 remount-and-recollapse bug for every chunk past the first, and the whole suite stays green. #3477's "Watch Out For" names that regression explicitly.

This test already builds runs at 10 / 11 / 25, so it is two lines:

```ts
// Chunk ids anchor to each chunk's own first child and stay put as the
// run grows, so an expanded second group survives an append (#2802).
expect(groupsIn(13).groups.map((g) => g.toolCallId)).toEqual(["group-t0", "group-t10"]);
expect(groupsIn(25).groups.map((g) => g.toolCallId)).toEqual(["group-t0", "group-t10", "group-t20"]);
```

`toolCallId` needs adding to the inline part type at lines 117-121 for that to typecheck.

web/src/components/acp/AcpRuntime.tsx
```diff
@@ -881,6 +881,12 @@ function collapseSubagents(parts: DraftPart[], profile: AgentProfile): DraftPart
  *  block (#1057). */
 const TOOL_GROUP_MIN_RUN = 3;
 
+/** Maximum run length that stays in a single group card. Longer runs are
```
Seluj78, line 884: **Nit:** this comment claims broader scope than the code has.

The cap applies only to the generic `_aoe_tool_group` path. The `run.every(isTodo)` branch at line 975 returns before the chunk loop, so an all-TodoWrite run of any length still folds into one unbounded `_aoe_todo_group` payload.

Leaving Todo unchunked is the right call, since #1468's card shows the latest list once and five cards each showing "latest list" would be wrong. Worth saying so here, though: all four models I consulted read this comment as covering both group kinds.

Seluj78 (changes requested):

## Review

Reviewed at `eeee94a`. Surfaces: web (structured view / ACP runtime), web-tests. CI has no failures; 12 checks still pending at review time. No unresolved review threads; CodeRabbit reported no actionable comments.

Pressure-tested the design against four independent models (Gemini 3.1 Pro, GPT-5.6, Claude Opus 4.7, Grok 4.5). No correctness or security defect survived verification. Two items below are worth landing before merge, plus one comment nit.

### The logic is sound

Chunk boundaries are index-aligned from the run start, so `group-t0` / `group-t10` / `group-t20` freeze the moment they form and #2802 does not regress: appending a call never mutates an existing chunk's first child, so an expanded group card stays mounted and stays expanded.

The `useMemo` deps work as intended. `argsText` is a string primitive and React compares deps with `Object.is`, which is value equality for strings, so a freshly built but identical JSON payload is a cache hit rather than a miss.

Chunking and memoization compose rather than overlap. Sealed chunks stop re-parsing entirely once their payload stops changing, and per-append parse work is capped at 10 children instead of growing with the turn. Leaving grouping unbounded and only memoizing would not achieve that: the tail `argsText` would grow on every append, making the parse quadratic over the stream. Keying by chunk index instead of first-child id would collide across two text-split runs in one message, which is the case the existing test at `AcpRuntime.test.ts:337` pins.

No security surface. The group envelope is built locally, agent-controlled strings ride through `JSON.stringify` / `JSON.parse` as data, and no new HTML or URL sink is introduced.

### Required

1. **Chunked-group key stability is untested.** See the inline comment on `AcpRuntime.test.ts`.

2. **`PR body: Checklist`, "For UI changes: included screenshot or recording" is unchecked** on a change that alters what the user sees. A 25-tool turn now renders as three stacked "10 actions" / "10 actions" / "5 actions" cards instead of one, and a 22-tool turn renders as two group cards plus two loose inline cards. Whether that stack reads as one logical run or as visual noise is not judgeable from the diff. Per `.github/pull_request_template.md` and `AGENTS.md` § Commit & Pull Request Guidelines.

### Nit

3. `TOOL_GROUP_MAX_RUN` doc comment scope. See the inline comment on `AcpRuntime.tsx`.

### Noted, not a finding

The tail flips from inline cards to a group at run lengths 13, 23, 33, discarding expand state on those one or two loose cards. That is the same transition the timeline already makes at 2 to 3, given `TOOL_GROUP_MIN_RUN = 3`, so it is not a new class of behavior.

### Verdict

Implementation: request changes, two cheap items, no defect in the logic.
Project fit: accept. Matches every line of #3477's Definition of Done.

njbrake (changes requested):
Seconding the request for changes at `eeee94a`, with one item to add and one to retire.

**Add: the memoization half has no test, and #3477 asks for one.** Codecov puts `StructuredView.tsx` patch coverage at 0%, and the two uncovered lines are the two `useMemo` calls. The Definition of Done line is "Re-renders of the structured view do not re-parse the full group children JSON" plus "Tests cover long runs and the memoization/performance path". Either add a Vitest render of `AssistantToolGroup`, which would also be the first coverage of `groupChildToItem`, or say in the description why the memo needs none. AGENTS.md asks for the latter explicitly: "if a change genuinely does not warrant a new test, say so in the PR description".

**Retire: the `useMemo` key is sound.** `argsText` is the only prop either group component takes, and the whole reconstruction chain (`parseGroupChildren`, `groupChildToItem`, `pickStartedAt`, `pickEndedAt`, `prettifyToolName`, `resultRowKind`) is pure in its arguments. The one impure call, `toolCallTimestamp`, is a module-level Map keyed by `toolCallId` and idempotent after the first mint. Stale children are not reachable, so this does not need revisiting.

**On the chunk-key ask:** I enumerated run lengths 1 through 26 through `activityToThreadMessages`. Group ids come out as `group-t0` / `group-t10` / `group-t20`, constant at every length, so the invariant holds today and the two-line assertion is a pin rather than a fix.

That enumeration is also the case for the screenshot. 11 tools render as one "10 actions" card plus one loose inline card, 12 as the card plus two loose cards, and at 13 those loose cards fold into a second group and lose their expand state, repeating at 23 and 33. Merging a sub-`TOOL_GROUP_MIN_RUN` tail into the previous chunk rather than spilling it inline removes both shapes, caps a group at 12 instead of 10, and leaves `group-t0` stable. Optional, and it is one line.

**Nit:** there is no coverage-matrix surface for the generic #1057 tool-group fold at all, while `acp.todo-group-fold` covers the sibling #1468 behavior. An `acp.tool-group-chunking` entry pointing at `AcpRuntime.test.ts` would match. CI cannot catch this either way: `validate-coverage-matrix.mjs` checks only that spec paths exist and that every component file is listed somewhere, never that a changed surface has an entry.

Checked on a clean rebase onto `2d1fa8e0`: oxfmt, ESLint, `tsc -b` and Vitest all clean (one unrelated `useFileContents` load flake that passes in isolation and on main).

Issue #3465, opened 2026-08-21
A status transition on disk can wipe a concurrent archive through merge_user_action_diff's touch arm


blueagledev:
@njbrake reproduced this while reviewing #3453 and asked for it as an issue.
Quoting his finding rather than restating it:

> Reproduced locally: with disk's `last_accessed_at` advanced by a status
> transition and a stale `pre`, `merge_user_action_diff` splices the user's
> `archived_at` in at 2139 and then the `touched` arm at 2223 wipes it. The TUI
> archive key is silently undone on disk and the row comes back on the next
> reload.

### The arm

`src/session/instance.rs`, in `merge_user_action_diff`:

```rust
// Touch is an event invariant: any advance of last_accessed_at
// (TUI-side or peer-side) dethrones a concurrent archive.
let touched = self.last_accessed_at > pre.last_accessed_at;
...
// touch_last_accessed(): clears archived + snoozed + idle-dormant.
if touched {
    self.archived_at = None;
    self.snoozed_until = None;
    self.idle_dormant_since = None;
}
```

The comment names the premise it needs: *any advance of `last_accessed_at`* is
treated as a user interaction. **One** status path advances it without being one
and reaches disk, which is what makes this bite:

`update_status_with_metadata` (`instance.rs:6682`) stamps
`last_accessed_at = Some(now)` whenever `prev != self.status`, with no
`is_archived` check at that point. `PassiveStatusPatch::from_instance`
(`instance.rs:1572`) carries that field, and `decide_passive_transition`
(`server/mod.rs:4317`) builds the patch for every non-structured row, so it
lands on disk. `touched` compares against the disk row, so that is the advance
it sees.

**Correcting my own first version of this issue on two counts.** It also listed
`apply_status_intent` (`server/mod.rs:5777`) as a second such path. That stamp
is in-memory only: the surrounding block persists through `storage.update` ->
`apply_acp_session_change`, which writes `acp_session_id` and not
`last_accessed_at`, so the merge never sees it. And it said structured rows are
protected by `apply_acp_overlay_inplace`'s durability contract. They are
protected, but by two plain persistence gates rather than by that contract:
`decide_passive_transition` and `persist_passive_status_transition`
(`tui/home/mod.rs:6630`) both build no patch when `is_structured()`. The
structured branch does still take the stamp in memory.

### Why the existing guard does not cover it

`update_status_with_metadata_inner` already short-circuits archived rows
[...]

PR #3481, opened 2026-08-22
fix(session): stop passive status stamps from wiping concurrent archives

jerome-benoit:
## Description

A passive status transition could erase a concurrent user archive through `merge_user_action_diff`'s `touched` arm (#3465, maintainer-reproduced). This PR removes the lying stamp rather than touching the deliberate arm.

### RCA (all links verified on this branch)

1. `Instance::update_status_with_metadata` (`src/session/instance.rs`, transition block) wrote `last_accessed_at = Some(now)` on every detected status change, with no user gesture behind it. Introduced by 16bdfad3 (#762) to power a "last activity" column/sort treating transitions as activity.
2. The stamp reaches disk through `PassiveStatusPatch::from_instance` (`src/session/instance.rs:1574`) via both producers: the daemon's `decide_passive_transition` (`src/server/mod.rs:4312`, patch for every non-structured row) and the TUI's `persist_passive_status_transition` (`src/tui/home/mod.rs:6892`).
3. A writer surface acting on a stale in-memory snapshot hits the wipe: `apply_user_action` captures `before` from TUI memory (`src/tui/home/mod.rs:6963`); if the daemon stamped disk in between, `merge_user_action_diff` computes `touched = self.last_accessed_at > pre.last_accessed_at` (`src/session/instance.rs:2205`) and clears `archived_at` / `snoozed_until` / `idle_dormant_since` (`instance.rs:2232`). The arm is deliberate (messaging-unarchives rule, pinned by `test_merge_diff_peer_touch_clears_tui_archive`), so the fix targets the stamp.

Competing hypotheses checked and refuted:

- "The passive patch itself clears triage fields": `merge_passive_status_patch` writes only `status`, `idle_entered_at`, monotone `last_accessed_at` and lifecycle generation (`instance.rs:2100`).
- "`update_status_with_metadata_inner`'s `is_archived()` short-circuit prevents stamping archived rows": it only tests the polling process's own copy, so the cross-process window (daemon stamps disk while the TUI holds an older snapshot) bypasses it entirely.
- "Consumers need the transition stamp": idle-reap anchors on `max(idle_entered_at, last_accessed_at)` and `idle_entered_at` still re-anchors on every Idle entry (`idle_reap.rs:90`, set in the same block). Display already prefers `idle_entered_at` for Idle rows since #872, which called `last_accessed_at` a liar after attach/send-keys. Freshness/Attention sorts document the field as user engagement.

### Fix

`update_status_with_metadata` no longer writes `last_accessed_at`; it keeps the `idle_entered_at` bookkeeping and baseline seeding untouched. Passive patches now carry the loaded value, which the monotone guard turns into a disk no-op. `merge_user_action_diff` and its pinned test are unchanged. `apply_status_intent` (defined at `src/server/mod.rs:5734`, stamping at :5777) stamps daemon memory for structured rows too; left untouched here. Its value can still reach disk indirectly through the TUI relay (`GET /api/sessions`, parsed by `DaemonStatusPoller`, written into TUI memory by `apply_status_update`, persisted by `save`'s monotone max), keeping a narrow #3465-shaped window alive for structured rows only; flagged as the follow-up.

### Behavior changes to be aware of

- `SortOrder::LastActivity` ("Recent") now ranks by last user interaction instead of last status flip, matching every doc comment on the field. Surfacing "just did something" stays the job of Attention sort, unread marks, and the fresh-idle pulse.
- Smart-attach fallback ranking and recent-projects recency now reflect user interaction rather than poller flips.
- Sidebar activity column: non-Idle rows now show time since the last user interaction rather than time since the last status flip, and a never-touched row shows an empty slot instead of an age.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the bug is a cross-process timestamp race; it is covered deterministically at the unit layer by four regression tests (red before the fix, verified by re-introducing the stamp): `test_merge_diff_passive_transition_stamp_does_not_wipe_concurrent_sink_state`, `test_merge_diff_passive_transition_stamp_does_not_wake_dormant_row`, and the rewritten `test_update_status_with_metadata_*` pair. An e2e reproduction would need to interleave a daemon tick between snapshot and action, which is not deterministic.

Regression suite:

- `test_merge_diff_*` family: 13 tests; full merge filter (`cargo test --lib merge`, all merge surfaces): 124 passed
- `*_status_with_metadata*`: 4 passed
- serve-gated `decide_passive_transition_*` / `flush_passive_transition_*` filters: 7 passed
- `cargo clippy -- -D warnings`: clean

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** ox-alpha (via Oh My Pi harness)

**Any Additional AI Details you'd like to share:** Full cycle executed autonomously: repro, RCA, design, fix, validation, two review cycles (initial + fresh-eyes) with convergence loops described in the comments below.

**NOTE:**

Ref #3465

## Summary

- Passive status changes no longer update `last_accessed_at`.
- User activity timestamps now reflect user interaction instead of background polling.
- Idle tracking still updates `idle_entered_at`, so idle-reap timing remains unchanged.
- Added regression tests for archive, snooze, dormancy, and status metadata preservation.
- Updated documentation to clarify timestamp ownership and passive patch behavior.
- Structured rows still retain a narrow related window through `apply_status_intent`; this remains follow-up work.

## Benefits

This prevents stale background updates from clearing archived, snoozed, or dormant state. It also improves activity sorting, smart-attach ranking, recent-project ordering, and sidebar timestamps.

### DIFF
--- a/docs/development/internals/sessions.md
+++ b/docs/development/internals/sessions.md
@@ -33,7 +33,7 @@ The canonical glossary lives on `PassiveStatusPatch` in [`src/session/instance.r
 
 Authority rule:
 
-- **Plain tmux sessions**: the poller is authoritative on `status` / `idle_entered_at` / `last_accessed_at`. `Instance::update_status_with_metadata` reads pane metadata + tmux state, compares against `live_status_baseline`, and mutates in place. `merge_passive_status_patch` on disk applies the same fields per patch.
+- **Plain tmux sessions**: the poller is authoritative on `status` / `idle_entered_at`. `last_accessed_at` is user-gesture-only since #3465: passive patches carry it through unchanged and the monotone guard turns that into a disk no-op. `Instance::update_status_with_metadata` reads pane metadata + tmux state, compares against `live_status_baseline`, and mutates in place. `merge_passive_status_patch` on disk applies the same fields per patch.
 - **Structured / ACP sessions**: `apply_acp_overlay_inplace` in `src/server/mod.rs` is the sole authority. `decide_passive_transition` returns `patch: None` for `is_structured()` rows, so the passive-status writers deliberately skip them. Post-daemon-restart the row's `idle_entered_at` resets to the last durable value (either creation-time, or the last explicit user action); ACP event handlers re-emit as they observe new state.
 
 Writer shapes:
--- a/docs/development/internals/sessions.md
+++ b/docs/development/internals/sessions.md
@@ -33,7 +33,7 @@ The canonical glossary lives on `PassiveStatusPatch` in [`src/session/instance.r
 
 Authority rule:
 
-- **Plain tmux sessions**: the poller is authoritative on `status` / `idle_entered_at`. `last_accessed_at` is user-gesture-only since #3465: passive patches carry it through unchanged and the monotone guard turns that into a disk no-op. `Instance::update_status_with_metadata` reads pane metadata + tmux state, compares against `live_status_baseline`, and mutates in place. `merge_passive_status_patch` on disk applies the same fields per patch.
+- **Plain tmux sessions**: the poller is authoritative on `status` / `idle_entered_at`. `last_accessed_at` is user-gesture-only since #3465: passive patches carry it through unchanged and the monotone guard turns that into a disk no-op. `Instance::update_status_with_metadata` reads pane metadata + tmux state, compares against `live_status_baseline`, and mutates in place. `merge_passive_status_patch` on disk applies the patch's `status` and `idle_entered_at` writes, the monotone `last_accessed_at` carry-through, and the lifecycle generation.
 - **Structured / ACP sessions**: `apply_acp_overlay_inplace` in `src/server/mod.rs` is the sole authority. `decide_passive_transition` returns `patch: None` for `is_structured()` rows, so the passive-status writers deliberately skip them. Post-daemon-restart the row's `idle_entered_at` resets to the last durable value (either creation-time, or the last explicit user action); ACP event handlers re-emit as they observe new state.
 
 Writer shapes:
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -4561,7 +4561,7 @@ async fn status_poll_loop(state: Arc<AppState>) {
         // that finds disk out of sync with live reality (the common case,
         // since nothing persists a passive transition until the patch below
         // lands) misreads that mismatch as a brand new transition and
-        // restamps idle_entered_at/last_accessed_at. See #2690.
+        // restamps idle_entered_at. See #2690.
         let prev_for_poll = prev.clone();
         // Invariant 8: read before `load_all_instances()` below. The tmux
         // scrape that follows it can block for seconds when the tmux server is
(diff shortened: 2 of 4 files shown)


njbrake (commented):
Approving the approach; two things to fix in the body before this goes in, neither of them code.

**Add the Attention sort to the behavior-change list.** `attention_session_key` (`groups.rs:682-740`) sorts within a tier by `(last_accessed_at.is_none(), ..., last_accessed_at)` ascending, so untouched rows bucket last. Before this change, an untouched Waiting row carried a flip timestamp and sorted near the top of its tier; now it is `None` and sorts to the bottom. That inverts the sort's own stated intent, "a Waiting session that has been sitting untouched for 2 days should rank above one that was just bumped a minute ago", for precisely the rows it exists to surface. Low impact in practice, since every realistic start path touches the field (attach-return at `app.rs:3809`, send at `home/mod.rs:5975`, API send at `api/sessions.rs:11059`, `cli/send.rs:101`), so the residual set is "created and started but never attached, prompted, or edited". Worth naming alongside the three surfaces you already list. `groups.rs:675` also still says "never polled", which is no longer what `None` means.

**Get the `apply_status_intent` residual into #3465.** Your read of it is right and the issue's is now stale: the stamp at `server/mod.rs:5777` does reach disk, through the relay you describe. `DaemonStatusPoller` carries the field on the wire (`daemon_status_poller.rs:80`), `apply_status_update` writes it into TUI memory unconditionally (`home/mod.rs:3273`, `3331`), and while `persist_passive_status_transition` gates on `is_structured()`, `save()`'s `merge_from_tui` monotone max does not. A second writer surface with a stale `pre` then hits the same arm. Since this is data loss, I do not want the remaining half tracked only in a merged PR body. Either drop that stamp here (structured rows already touch on every user prompt via `api/acp.rs:1202`, so nothing depends on it) or add it to #3465 as the open half.

Verified the two claims the RCA rests on. `16bdfad3` does introduce the stamp specifically for the last-activity column and the `LastActivity` sort, so the field you are changing is load-bearing for #762 by construction. And the `touched` arm is deliberate, pinned by `test_merge_diff_peer_touch_clears_tui_archive`, so targeting the stamp instead of the arm is the right call.

Also checked the consumer that would have made this blocking: idle reap is bit-identical. `should_auto_stop_session` requires `idle_entered_at.is_some()` and anchors on `max(idle_entered_at, last_accessed_at)`, and `accessed > entered` was only ever true for a real user touch even before this change, because entering Idle stamped both to the same instant and any non-Idle transition cleared `entered`. Auto-stop timing does not move.

Nit, non-blocking: `instance.rs:2118` "dropped passive status patch's `last_accessed_at` as a no-op" was written for the exceptional stale-peer case and now fires on essentially every passive transition. `trace!` would fit better. Note that `drop_log_count` at `instance.rs:10055` matches the message by substring and two `traced_test` tests assert its count, so check those if you touch either the level or the text.

I have a one-line patch for the `groups.rs:675` wording if you want me to push it to your branch; say the word.

Issue #1581, opened 2026-05-28
Web dashboard: surface snooze, archive, and favorite controls (parity with TUI/CLI triage)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

The web dashboard cannot snooze, archive, or favorite a session. All three are first-class triage primitives in the TUI (single-key bindings, with archive killing the tmux pane and snooze opening a duration picker) and in the CLI (`aoe session archive|unarchive|snooze|unsnooze|favorite|unfavorite`). The web sidebar currently reads `favorited` as a render-only field to pin rows, but offers no toggle for any of the three, and renders nothing for `archived_at` / `snoozed_until` because those fields are not even exposed on the API.

Net effect: a user driving AoE primarily from the web dashboard (phone, remote, read-only viewer attached to a remote daemon) has no way to triage their session list. They can only delete or rename. Sessions accumulate at the top of the sidebar with no way to sink or temporarily defer them.

## What's there today

**Data model is complete.** `Instance` already carries the three fields with their mutual-exclusion semantics and peer-merge invariants:

```rust
// src/session/instance.rs:304-326
pub archived_at: Option<DateTime<Utc>>,
pub favorited_at: Option<DateTime<Utc>>,
pub snoozed_until: Option<DateTime<Utc>>,
```

Predicates and mutators at `src/session/instance.rs:812-895`:

- `archive()` / `unarchive()` / `is_archived()`
- `snooze(minutes)` / `unsnooze()` / `is_snoozed()` / `snooze_remaining()`
- `favorite()` / `unfavorite()` / `is_favorited()`

Mutual exclusion lives in the mutators (`favorite()` clears `archived_at` + `snoozed_until`; `archive()` clears `favorited_at`; `touch_last_accessed()` clears both sinks). Peer-merge invariants in `merge_diff` at `src/session/instance.rs:737-805` enforce these across concurrent TUI / CLI / API writes.

**CLI is complete.** `src/cli/session.rs:53-95, 238-353`:

```rust
SessionCommands::Snooze(args)     => snooze_session(profile, args).await,
SessionCommands::Unsnooze(args)   => unsnooze_session(profile, args).await,
SessionCommands::Archive(args)    => archive_session(profile, args).await,
SessionCommands::Unarchive(args)  => unarchive_session(profile, args).await,
SessionCommands::Favorite(args)   => favorite_session(profile, args).await,
[...]

PR #3489, opened 2026-08-23
fix(cli): expose snoozed_until and pinned_at in list and session JSON

### LINKED ISSUE #3415: cli: session JSON exposes archived and trashed but not snoozed or pinned, so a CLI consumer sees two of the four states the API has
`GET /api/sessions` has carried the full set of session-state timestamps since
#1581: `pinned_at` (`src/server/api/sessions.rs:97`), `snoozed_until` (`:110`),
plus `archived_at` and `trashed_at`, each `skip_serializing_if = "Option::is_none"`.
A web client can therefore tell any of the four apart.

The CLI cannot. #3350, closed by #3361, gave `aoe list --json` a `state` tag with
`trashed_at` and `archived_at`. #3414 gives `aoe session show --json` the same three. Neither surface carries `snoozed_until` or
`pinned_at`, and there is no `aoe session list-snoozed` to cross-reference
against, so a scripted consumer has no route to either state at all.

The asymmetry matters most for snooze, because a snoozed session is
indistinguishable from an idle one in every CLI output: `status` reports `Idle`,
the row lists normally, and `aoe add` on the same title and path is refused. That
is the same trap #3350 was filed for, one state over.

Two things worth deciding together rather than one at a time:

1. **Whether the CLI should mirror the API's field set** rather than growing one
   field per report. `Instance` already holds all five (`archived_at:715`,
   `favorited_at:724`, `snoozed_until:737`, `pinned_at:774`, `trashed_at:791`);
   only the two JSON projections choose what to expose.
2. **Whether `state` should stay a single tag.** It currently answers
   live/archived/trashed, derived from `effective_bucket`, and snooze and pin are
   orthogonal to all three: a session can be snoozed *and* archived. If the set
   is exposed as timestamps, `state` stays as it is and the two new keys are
   independent, which is what the API already does.

I am happy to put up the PR if the direction is right. The one design question I
cannot answer from outside is whether `snoozed_until` should be surfaced
unconditionally or only while the snooze is live, as the API does
(`sessions.rs:414` gates it on `inst.is_snoozed()`); the CLI has no such
precedent to follow.

jerome-benoit:
## Description

Fixes #3415. `GET /api/sessions` has exposed four session-state timestamps since #1581 (`pinned_at`, `archived_at`, `snoozed_until`, `trashed_at`), but the CLI JSON projections behind `aoe list --json` (#3361) and `aoe session show --json` (#3414) only carried `state`, `archived_at`, and `trashed_at`. A snoozed session was therefore indistinguishable from an idle one in every CLI output: same `"state": "live"`, no key to cross-reference, and no `list-snoozed` subcommand to fall back on.

Both CLI projections now mirror the API field set:

- `aoe list --json` (`SessionJson`) and `aoe session show --json` (`SessionDetails`) gain `snoozed_until` and `pinned_at`.
- `snoozed_until` is gated on `Instance::is_snoozed()` exactly like the API (`src/server/api/sessions.rs:417`): an expired deadline stays persisted until the next mutation rewrites it, but the row has woken, so the key disappears instead of advertising a stale snooze.
- The gate lives in one place, `cli::list::active_snoozed_until` next to `state_tag`, and the `show` projection reuses it so the two surfaces cannot drift (the same pattern as `super::list::state_tag`).
- `pinned_at` is a plain presence mirror of `SessionResponse::from_instance` (the client derives the boolean as `pinned_at != null`).
- `state` stays the single `effective_bucket` tag; both new keys are independent of it.

## RCA

Symptom: a snoozed session serialized identically to an idle one on both surfaces (captured pre-fix failure: `{"command":"","group":"","id":"...","path":"/repo","profile":"p","state":"live","status":"idle","title":"z","tool":"claude"}`).

Cause chain: `Instance` carries everything (`snoozed_until` at `src/session/instance.rs:751`, `pinned_at` at `:788`), but both projections hand-pick fields: `SessionJson` (`src/cli/list.rs`) and `SessionDetails` (`src/cli/session.rs`) declared neither key and their constructors copy only declared fields. Loss by projection omission, not data loss.

Competing hypothesis considered and refuted: "`state` could carry snooze as a fourth tag". `effective_bucket()` defines exactly three buckets (`Trashed > Archived > Active`), snooze is orthogonal to all three (a session can be snoozed and archived at once on the store, e.g. after a restore from trash preserves sibling timestamps), and #3361/#3414 established `state` as the bucket vocabulary shared by both surfaces.

## Design decisions (justified by the API precedent)

- Mirror the API's field set rather than grow a per-report field (#3415 proposal 1).
- Gate `snoozed_until` on `is_snoozed()` like the API; expose unconditionally would resurrect the "snoozed 0m" chip problem the API comment warns about.
- Keep `state` untouched (#3415 proposal 2); the new keys are independent timestamps.
- Types follow the existing CLI convention (`Option<chrono::DateTime<Utc>>`, RFC3339 on the wire) rather than the API's pre-formatted strings; same semantics.
- Out of scope for this PR: `favorited_at`, which is not part of the reported four-state gap (the REST API surfaces favorite as a boolean, not a timestamp), and no new `list-snoozed` style subcommand.

## Tests

RED before GREEN: the two table-driven unit tests were written first against the unfixed code and failed exactly on the active-snooze case, then passed after the fix.

Per surface, one table-driven test covers: active snooze (key present), expired snooze (key absent via the gate), pinned (`pinned_at` present), plain row (both keys absent), snoozed+archived combination (keys independent of `state`), pinned+snoozed combination (neither key suppresses the other), and markers on non-live buckets: a snoozed or pinned row later trashed still reports its key with `state: "trashed"`, and an archived row keeps a peer-written pin. Existing narrow-output tests still pass, so rows that carry neither marker serialize byte-identical to before.

New e2e `test_cli_snoozed_row_exposes_snoozed_until` drives the real binary through `aoe session snooze` and asserts both surfaces.

Validation: targeted lib tests green, e2e green, `cargo fmt --all -- --check` clean, `cargo clippy -- -D warnings` clean.

## Risks

Additive-only wire change: consumers parsing strictly see new keys only on snoozed/pinned rows; plain rows are unchanged (guarded by the narrow-output unit tests and the live-row assertions in the new e2e). Note for reviewers: running clippy with `--all-targets` on current stable additionally flags pre-existing lints in `src/session/groups.rs` tests and `src/tui/home/mod.rs`; those are untouched by this diff and outside the CI invocation.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

(`pinned_at` has no CLI producer, pin is set from the web sidebar, so the e2e covers the snooze path end to end and the unit tables cover pin.)

## AI Usage

- [x] AI was used

**AI Model/Tool used:** ox-alpha (Oh My Pi coding agent)

**Any Additional AI Details you'd like to share:** Automated issue cycle: repro, RCA, implementation, validation, adversarial self-review.

## Summary

- Added `snoozed_until` and `pinned_at` to `aoe list --json` and `aoe session show --json`.
- Show `snoozed_until` only while the session is actively snoozed.
- Preserve both fields independently of session state, including archived and trashed sessions.
- Added shared projection logic, documentation, unit tests, and an end-to-end snooze test.
- Validated the changes with targeted tests, end-to-end tests, formatting, and Clippy.

## Benefits

CLI JSON output now matches `GET /api/sessions`. Consumers can identify active snoozes and pinned sessions directly.

## Further enhancement

Add consumer-facing examples for sessions that are both pinned and snoozed.


njbrake (commented):

---

## Summary

This closes a real gap: `GET /api/sessions` has surfaced `pinned_at` / `archived_at` / `snoozed_until` / `trashed_at` since #1581, while `SessionJson` (`src/cli/list.rs:94`) and `SessionDetails` (`src/cli/session.rs:333`) hand-picked only `state`, `archived_at`, and `trashed_at`, so a snoozed row and an idle row serialized identically on both CLI surfaces. The fix adds the two missing keys to both projections, gates `snoozed_until` behind a shared `active_snoozed_until` helper, and leaves `state` alone. I rebased it onto `origin/main`, reproduced the load-bearing claims, and it is sound. The conflict is trivial. Every claim in the PR body that I checked held up, with one wording caveat noted below.

I verified rather than assumed the two decisive facts:

- **The gate is genuinely shared.** I mutated `active_snoozed_until` (`src/cli/list.rs:85-91`) to return `inst.snoozed_until` unconditionally and re-ran the suite. *Both* `cli::list::tests::session_json_mirrors_the_api_snooze_and_pin_keys` and `cli::session::show_json_tests::show_json_mirrors_the_api_snooze_and_pin_keys` failed on the "expired snooze" row. A single-site mutation breaking the other module's test is proof the helper is not reimplemented.
- **The gate matches the API.** `src/server/api/sessions.rs:417-421` is `if inst.is_snoozed() { inst.snoozed_until.map(to_rfc3339) } else { None }`; `Instance::is_snoozed` (`src/session/instance.rs:2537-2539`) is `snoozed_until.map(|t| t > Utc::now())`. The CLI helper is the same predicate over the same field. Same semantics, different wire type (`DateTime` vs pre-formatted string), which matches each surface's existing convention.

## Design alignment

**Aligned, with one deviation worth naming.**

The precedent is `state_tag` (`src/cli/list.rs:71`), declared in `list.rs` and reused by `session.rs:376` — this PR follows it exactly, including placing the new helper directly beside it. That is the right instinct.

But look at *why* `state_tag` works: the rule itself lives on the model (`Instance::effective_bucket`), and `state_tag` is only a CLI-specific string mapping over it. `active_snoozed_until` is different — the rule it encodes is not CLI-specific at all. The API needs the identical rule and has its own inline copy at `src/server/api/sessions.rs:417`. So after this PR there are **three** consumers of one rule and only two of them share code. The PR body says "the two surfaces cannot drift," which is true and also narrower than the problem. The `state_tag`-faithful version puts `active_snoozed_until()` on `Instance` next to `is_snoozed()` (`src/session/instance.rs:2537`) and has the API call `inst.active_snoozed_until().map(|t| t.to_rfc3339())`, leaving one definition for all three. Non-blocking, but it is the same drift class the PR names as its own rationale, so it is worth doing while you are here.

## Blocking findings

None.

## Non-blocking nits

**1. The merge conflict is purely mechanical, but resolve it carefully.** `tests/e2e/cli.rs` conflicts only because #3484 (`ec6b6d04`, the `acp doctor` version-floor tests) appended four tests at EOF while this branch appended one at the same spot. No overlap in intent, and nothing on main touched `src/cli/list.rs`, `src/cli/session.rs`, or the CLI JSON projections — `git merge-tree origin/main HEAD` reports exactly one conflicted path. There is a trap in it: git factors the shared trailing `}` out of the conflict region, so a naive "keep both sides, delete the markers" resolution silently swallows one closing brace and nests the new `test_cli_snoozed_row_exposes_snoozed_until` inside `test_cli_acp_doctor_bundle_only_judged_by_pinned_copy`. It fails to compile so it cannot ship, but it will cost a cycle. Keep both blocks and re-add the `}` that closes the `acp doctor` fn. After that resolution the branch is clean: `cargo fmt --all -- --check` passes, `cargo clippy --all-targets --features e2e-tests` produces only the pre-existing `src/process/macos.rs:266` and `src/tui/home/mod.rs:6044` warnings already called out in the PR body, all 105 `cli::` lib tests pass, and the new e2e passes in 1.34s.

**2. Neither table asserts the emitted timestamp's *value*.** Every row checks only `value.get("snoozed_until").is_some()` (`src/cli/list.rs`, `src/cli/session.rs`, in the `for (label, inst, want_snooze, want_pin, want_state)` loops). A regression that emitted `Some(Utc::now())` instead of `inst.snoozed_until`, or that swapped `snoozed_until` and `pinned_at`, passes all nine rows on both surfaces and passes the e2e too (which asserts only `.is_string()`). One extra assertion on the active-snooze row comparing the parsed value to the fixture deadline would close that, at a cost of about three lines.

**3. Nine rows is more table than the invariants need.** House style is right that the cost is code volume, and this is duplicated verbatim across two modules, so it is ~100 lines for what is really four distinct invariants: the gate (active vs expired), plain-row absence, pin presence, and marker/bucket independence. `snoozed and archived`, `trashed and snoozed`, `trashed and pinned`, and `pinned and archived` are four rows pinning one invariant. Not blocking — the rows are one line each, and the commit history (`14adb848`, `d5f88cb9`) shows each was added in response to a specific mutation that survived the suite, which is exactly the justification the guideline asks for. Flagging only that two of the four could go without losing a mutant.

**4. The new skill-doc paragraphs are asymmetric with the ones they sit beside.** In `contrib/hermes-skill/SKILL.md:94` and `contrib/openclaw-skill/SKILL.md:97`, the existing paragraph is meticulous about durability: which command clears `archived_at`, which clears `trashed_at`, and that a consumer should key off `state` rather than cache a timestamp. The new paragraph says nothing about what clears `snoozed_until`, and it is cleared by more things than `archived_at` is: `touch_last_accessed` (`src/session/instance.rs:1746`, so any `aoe send`), `archive()` (`:2281`), `favorite()` (`:2441`), `unsnooze()` (`:2505`), and `pin()` (`:2575`). A skill consumer reading only this paragraph would reasonably cache the deadline. One clause listing the clearers would match the surrounding voice.

Two related notes on those files, both fine as-is: the JSON sample blocks above the prose do not enumerate `archived_at`/`trashed_at` either, so leaving the new keys out of the samples is consistent with existing style; and `cargo xtask check-skill` (`xtask/src/main.rs:404-540`) validates only frontmatter `version:` rules and that referenced `aoe …` commands exist in the clap tree — it does **not** validate JSON field lists. So these two files are a hand-maintained mirror with no CI gate on the part changed here. That is a reason to be precise in them, not a reason to relax.

## On the specific claims checked

- **"Additive-only."** True, and stronger than the PR argues. `serde_json` serializes struct fields into a `BTreeMap`, so key order is alphabetical rather than declaration order; combined with `skip_serializing_if = "Option::is_none"` on both new fields, a row carrying neither marker is byte-identical. This showed up directly in the mutation-test failure dump, where the keys came out sorted. Consumers using `serde(deny_unknown_fields)` on rows that *do* carry a marker would break, but nothing in-repo deserializes either struct — both are `Serialize`-only.
- **`docs/cli/reference.md` did not need regenerating.** It is generated from clap help, and this PR adds no flag, subcommand, or help text. Grepping it for `snoozed_until` / `pinned_at` / `archived_at` gives zero hits; the file documents no JSON shapes at all. The only JSON-shape mirrors in the tree are the two `contrib/` skill files, which the PR did update.
- **Test structure.** Genuinely one `#[test]` per surface with a `cases` array inside, not one test per case. The e2e uses `#[parallel]`, correctly — it needs no process-global state, only the harness's per-test `$HOME` and tmux socket.
- **Breadth.** Checked the other `#[derive(Serialize)]` CLI structs for a missed surface. `RowJson` / `AcpRowJson` (`src/cli/ps.rs:314`, `:508`) and `StatusJson` (`src/cli/status.rs:35`) are process/pane views carrying no session-state timestamps, so `SessionJson` and `SessionDetails` really are the complete set. No gap.
- **One wording correction to the PR body.** It says "the gate lives in one place… so the two surfaces cannot drift." Accurate for the two CLI surfaces, but the API is a third surface with its own copy of the same rule, and the PR does not mention it. Worth restating precisely, since the sentence currently reads as a stronger guarantee than the diff delivers.

## Merge risk

**Low.** Reverts cleanly, no schema or migration involved, blast radius is one optional key on two JSON surfaces. The most plausible way this bites is nit 2: the tests pin key *presence*, not value, so a future refactor that emits the wrong instant is invisible to the whole suite, and you would hear about it from a scripted consumer rather than from CI.

## Verdict

**MERGE WITH NITS.**

Real bug, right scope, correct fix, and the tests were built by mutation rather than by coverage-chasing, which shows in the commit history. Resolve the `tests/e2e/cli.rs` conflict (mechanical, watch the swallowed brace); nits 2 and 4 are a few lines each and would be worth taking in the same push. The design point about hoisting `active_snoozed_until` onto `Instance` is worth a follow-up but should not hold this up.

PR #3503, opened 2026-08-24
feat(web): search box for the new-session wizard's project picker

### LINKED ISSUE #3461: Web dashboard: new-session wizard needs a search box for projects, not just a capped Recent/Saved list
## Problem

The new-session wizard's Project step (`web/src/components/session-wizard/steps/ProjectStep.tsx`) shows "Saved projects" and "Recent" sections on its Recent tab, but there's no way to search/filter them by typing. With more than a handful of projects, the useful ones scroll past or get cut off entirely (recents are sliced to 6), and finding a specific one means scanning a list or falling back to the Browse tab to re-navigate the filesystem by hand.

For a user with many projects (multiple orgs, many repos checked out), neither the capped Recent/Saved list nor manual filesystem browsing scales. What's needed is a fast, type-to-filter search over **all** known projects (saved projects plus everything session history knows about), independent of the arbitrary recency cap.

## What's there today

- `ProjectStep.tsx` already computes `filteredRecent`/`filteredSaved` by substring-matching `data.path`, but that's only driven by a path chosen via click/Browse/Clone — there's no rendered `<input>` for the user to type a query on the Recent tab.
- The saved-projects registry already exists as a first-class concept, separate from session-derived recents: `GET /api/projects` (`src/server/api/projects.rs`, backed by `src/session/projects.rs`), persisted per-scope in `<app_dir>/projects.json` or `<app_dir>/profiles/{profile}/projects.json`.
- Recents are separately persisted at `<app_dir>/recent-projects.json` (`src/session/storage.rs`) and merged with saved projects via `mergeRecentProjects`.

## Proposal

Add a visible search input to the wizard's project picker that filters across the full combined list (saved + recent, not just the first 6), live as you type — similar in spirit to a command palette. Consider whether it should also reach into the filesystem browser results for paths not yet known to either list.

## Why

Recents-only / Browse-only doesn't scale past a handful of projects. This makes starting a session in an existing, less-recently-used project fast instead of a scroll-and-hunt or a manual directory walk.

Seluj78:

## Description

The new-session wizard's Recent tab capped its recents at 6 rows and had no way to type a filter. With more than a handful of checkouts, a project either scrolled past that cap or never rendered at all, so picking a less-recently-used one meant re-walking the filesystem in the Browse tab.

`ProjectStep.tsx` already had `filteredRecent` / `filteredSaved` memos, but they matched against `data.path`, the path the user had *already selected*. Nothing ever fed a query into them, so the only visible effect was that clicking a project silently collapsed the list down to rows matching it. This repoints both memos at a real `query` state driven by a new search input, keeps the full merged recents list in component state instead of slicing it at fetch time, and applies the 6-row cap only when the query is empty. A search therefore reaches every project the dashboard knows about, saved registry entries and session-derived plus persisted recents alike, while an untouched tab still reads as a short "jump back in" list.

Two consequences worth calling out for review. First, selecting a project no longer collapses the list, so you can change your mind without clearing the field; that is the behavior the old `data.path` filter was accidentally preventing. Second, because the cap now applies after `splitSavedAndRecent` dedupes saved paths out of recents, a saved project no longer eats a recent slot, so you may see 6 recents where you previously saw 5.

Scoped deliberately: plain case-insensitive substring matching over name and path, matching the memos that were already there. No fuzzy ranking, no keyboard navigation, and no reach into the filesystem browser for paths unknown to both lists. The issue floats that last one as a "consider whether"; it needs a debounced backend search and the Browse tab already covers genuinely unknown paths, so it is left out.

## Also in this branch: two unrelated CI fixes

CI was red on this branch for reasons that have nothing to do with the search box, so two pre-existing flakes on `main` are fixed here as drive-bys. `WorkspaceSidebar.tsx` swaps a `useEffect` for a `useLayoutEffect` so the Escape-key listener re-subscribes in the same commit that paints the result, closing a window where RTL's `waitFor` returned before the passive-effect flush and the keypress was swallowed (`AddProjectAction.test.tsx:339`). `web/tests/live/mobile-live-no-flutter.spec.ts` tightens its settle gate from one equal sample pair at 80ms to four consecutive equal samples at 90ms, so it no longer starts sampling before the layout has converged. The 5000ms budget and the `jitter < 3` assertion are both untouched, so the gate only got stricter.

One test edit in `web/tests/wizard-extra-repos.spec.ts:99` looks like a loosened assertion and is the opposite: the picker chip concatenates name and scope with no separator, so `/^primary$/` matched nothing and its `toHaveCount(0)` was passing vacuously. Dropping the anchor to `/^primary/` gives that assertion teeth for the first time.

Closes #3461

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Open the dashboard with more than 6 known projects, press `n`, and confirm the Recent tab shows a "Search projects by name or path" box above the lists.
- [ ] With the box empty, confirm at most 6 recent rows render (plus any Saved projects section), same as before.
- [ ] Type part of the name of a project you know sits below the 6-row cap; confirm it appears.
- [ ] Type part of a *path* rather than a name (for example `/repo/`), confirm matching rows render.
- [ ] Confirm the query also narrows the Saved projects section.
- [ ] Type something that matches nothing; confirm "No projects match that search. Try the Browse tab." instead of a blank tab.
- [ ] Clear the box; confirm the capped list comes back.
- [ ] Click a project row, then confirm the other rows stay visible so you can pick a different one, and that launching the session still uses the clicked path.

## Checklist

## Test Coverage Analysis

New Vitest spec `web/src/components/session-wizard/__tests__/ProjectStep.search.test.tsx` covers the cap-with-empty-query, reaching past the cap by typing, path matching, saved-section filtering, and the no-match empty state. `ProjectStep.saved-projects.test.tsx` had a case pinning the old `data.path`-as-query behavior; it now drives the search box instead. `web/tests/wizard-extra-repos.spec.ts` needed its two ambiguous locators scoped to a new `data-testid="extra-repos-picker"`, because the project list no longer collapses on selection and the same project names now render in both places. `ProjectStep.tsx` and `ExtraReposPicker.tsx` were already mapped by the existing `wizard` entry in `coverage-matrix.json`, so no matrix edit was needed and `validate-coverage-matrix.mjs` passes.

**TUI / CLI (e2e test)**

No docs change: the search box is discoverable UI, so there is nothing a user cannot work out by opening the wizard.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I made the design calls (substring over fuzzy, no filesystem reach, cap only when the query is empty), reviewed each commit, and will run the manual test steps above before marking this ready.

## Summary

- Added project search to the Recent project picker.
- Search matches saved and recent projects by name or path.
- The six-project limit applies only when the search field is empty.
- Added an empty-results message.
- Expanded test coverage and improved project-picker locator scoping.
- Added a test ID for the extra repositories picker.
- Improved Escape-key handling and mobile viewport settling.

## Benefits

Users can find projects beyond the visible recent-project limit. Saved projects and recent projects now share one search experience.

## Potential enhancement

Filesystem search for paths unknown to the dashboard remains outside this change.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1610,6 +1610,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1627,6 +1630,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1644,6 +1650,9 @@
         "ppc64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1661,6 +1670,9 @@
         "riscv64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1678,6 +1690,9 @@
         "riscv64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1695,6 +1710,9 @@
         "s390x"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1712,6 +1730,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1729,6 +1750,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1610,9 +1610,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1630,9 +1627,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1650,9 +1644,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1670,9 +1661,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1690,9 +1678,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1710,9 +1695,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1730,9 +1712,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1750,9 +1729,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/web/src/components/WorkspaceSidebar.tsx
+++ b/web/src/components/WorkspaceSidebar.tsx
@@ -2245,7 +2245,14 @@ export function AddProjectModal({
   // letting Escape close the modal mid-request would throw away the result,
   // the warnings, and the "the agent did not restart" notice this component
   // exists to surface.
-  useEffect(() => {
+  //
+  // The subscription closes over `busy`, so it has to be re-established in a
+  // layout effect, not a passive one. A keydown on `document` never reaches
+  // React's root listener, so nothing flushes pending passive effects before
+  // this handler runs. As a passive effect it left a window right after the
+  // response landed where the DOM already showed the result but the listener
+  // still held `busy === true`, and Escape was silently swallowed.
+  useLayoutEffect(() => {
     const onKey = (e: KeyboardEvent) => {
       if (e.key === "Escape" && !busy) onCancel();
     };
--- a/web/src/components/session-wizard/__tests__/ProjectStep.saved-projects.test.tsx
+++ b/web/src/components/session-wizard/__tests__/ProjectStep.saved-projects.test.tsx
@@ -175,15 +175,17 @@ describe("ProjectStep saved-projects render (#2140)", () => {
     expect(queryByText("Recent", asHeader)).toBeNull();
   });
 
-  it("narrows the Saved projects section to entries matching the path filter", async () => {
+  it("narrows the Saved projects section to entries matching the search query", async () => {
     vi.mocked(fetchProjects).mockResolvedValue([
       savedProject({ name: "alpha", path: "/repo/alpha" }),
       savedProject({ name: "zeta", path: "/repo/zeta" }),
     ]);
     vi.mocked(fetchSessions).mockResolvedValue({ sessions: [], workspace_ordering: [] });
 
-    // data.path acts as the filter query against both sections.
-    const { findByText, queryByText } = renderStep("zeta");
+    // The search box drives the filter over both sections (#3461); before it
+    // existed the query came from the already-selected data.path.
+    const { findByLabelText, findByText, queryByText } = renderStep();
+    fireEvent.change(await findByLabelText("Search projects"), { target: { value: "zeta" } });
     expect(await findByText("/repo/zeta")).toBeTruthy();
     expect(queryByText("/repo/alpha")).toBeNull();
   });
--- a/web/src/components/session-wizard/__tests__/ProjectStep.search.test.tsx
+++ b/web/src/components/session-wizard/__tests__/ProjectStep.search.test.tsx
@@ -1,0 +1,120 @@
+// @vitest-environment jsdom
+//
+// #3461: the Recent tab has a search box that filters across the whole
+// saved + recent list. With an empty query the recents stay capped at
+// RECENT_CAP (6) so the tab reads as a short "jump back in" list; typing
+// searches the full merged list, so a project below the cap is reachable
+// without falling back to the Browse tab.
+
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+
+import { ProjectStep } from "../steps/ProjectStep";
+import { initialData } from "../wizardReducer";
+import type { ProjectInfo } from "../../../lib/types";
+import type { RecentProjectEntry } from "../../../lib/api";
+
+vi.mock("../../../lib/api", () => ({
+  fetchSessions: vi.fn().mockResolvedValue({ sessions: [], workspace_ordering: [] }),
+  fetchRecentProjects: vi.fn(),
+  fetchProjects: vi.fn(),
+  cloneRepo: vi.fn(),
+}));
+
+import { fetchRecentProjects, fetchProjects } from "../../../lib/api";
+
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
+});
+
+// Eight recents, newest first, so `zebra` lands at index 7 and is below the
+// six-row cap. Names are distinct enough that a substring hits exactly one.
+const NAMES = ["alpha", "bravo", "charlie", "delta", "echo", "foxtrot", "golf", "zebra"];
+
+function recents(): RecentProjectEntry[] {
+  return NAMES.map((name, i) => ({
+    path: `/repo/${name}`,
+    display_name: name,
+    tool: "claude",
+    // Descending timestamps keep the array order after the merge sort.
+    last_used_at: `2025-09-${String(20 - i).padStart(2, "0")}T00:00:00+00:00`,
+  }));
+}
+
+function savedProject(): ProjectInfo {
+  return { name: "saved-yankee", path: "/repo/saved-yankee", scope: "global" } as ProjectInfo;
+}
+
+function renderStep() {
+  return render(
+    <ProjectStep data={{ ...initialData, path: "", extraRepoPaths: [], scratch: false }} onChange={vi.fn()} />,
+  );
+}
+
+describe("ProjectStep project search (#3461)", () => {
+  it("caps recents with an empty query and finds a below-the-cap project by typing", async () => {
+    vi.mocked(fetchRecentProjects).mockResolvedValue({ projects: recents() });
+    vi.mocked(fetchProjects).mockResolvedValue([]);
+
+    renderStep();
+    const box = await screen.findByLabelText("Search projects");
+
+    // Only the first six recents render before any query.
+    expect(await screen.findByText("alpha")).toBeTruthy();
+    expect(screen.getByText("foxtrot")).toBeTruthy();
+    expect(screen.queryByText("golf")).toBeNull();
+    expect(screen.queryByText("zebra")).toBeNull();
+
+    // Typing reaches past the cap.
+    fireEvent.change(box, { target: { value: "zeb" } });
+    expect(await screen.findByText("zebra")).toBeTruthy();
+    expect(screen.queryByText("alpha")).toBeNull();
+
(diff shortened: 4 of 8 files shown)


njbrake (commented):

_Reviewed at `6cf2b694`. Push new commits and it will be re-reviewed._

---

## Summary

Adds a type-to-filter search box to the wizard's Recent tab. The 6-row recents cap moves from fetch time to the render memo and applies only when the query is empty, so a search reaches the whole merged saved + recent list. Also carries two unrelated CI-flake fixes that the PR body never mentions.

### Design Alignment: **Aligned**

Compared against the code the PR replaces and its siblings:

- The feature extends the two existing `filteredRecent` / `filteredSaved` memos rather than adding a parallel filter path. No new abstraction, no new state machine.
- The magic `6` at `ProjectStep.tsx:191` becomes a named `RECENT_CAP` at `ProjectStep.tsx:57`.
- The new input's `className` at `ProjectStep.tsx:375` is byte-identical to the existing clone-URL input at `ProjectStep.tsx:475` (minus `font-mono`), and matches `ExtraReposPicker.tsx:61`. `focus:border-brand-600` is the brand-accent focus convention `DESIGN.md:257` calls for. I diffed the class strings rather than eyeballing the screenshot; they match.
- The `useLayoutEffect` swap in `WorkspaceSidebar.tsx:2266` has an in-file precedent at `WorkspaceSidebar.tsx:3543`, which uses a layout effect for exactly the same "keep the closure fresh at commit time" reason.

### Verifying the body's five claims

**1. "The memos matched `data.path`; nothing ever fed a query into them." — TRUE.** Confirmed against the pre-PR lines in the diff, and issue #3461 independently documents the same thing. The user-visible symptom was real: clicking a project collapsed both lists to rows matching the clicked path. The diff repoints both memos at `query` (`ProjectStep.tsx:206-219`).

**2. "Cap applied only when the query is empty." — TRUE, and the uncapped case is a non-issue.** The slice moved from `ProjectStep.tsx:191` to `ProjectStep.tsx:209` (`recent.slice(0, RECENT_CAP)` on the empty-query branch). A query does render an uncapped list, but the wizard body is `flex-1 overflow-y-auto` (`SessionWizard.tsx:418`), so a long result set just scrolls, and the reachable set is bounded anyway: saved registry + projects with sessions + at most 20 persisted recents (`RECENT_PROJECTS_CAP`, `src/session/storage.rs:839`). I raised this as a concern and then found the evidence against it, so I am dropping it.

**3. "A saved project no longer eats a recent slot." — TRUE, and it is the only ordering side effect I found.** Before: `slice(0,6)` then dedupe, so recents could render 5. After: dedupe then `slice(0,6)`, so up to 6. The two other consumers of the pre-slice list are unaffected: `hasPicks` (`ProjectStep.tsx:222`) and the browse-tab-default check (`ProjectStep.tsx:198`) both only test emptiness, which the slice never changed.

**4. "Selecting a project no longer collapses the list." — TRUE, nothing downstream depended on the collapse.** I walked each consumer: row highlighting is `data.path === r.path` (still correct), `ExtraReposPicker` is gated on `data.path && activeTab !== "browse"` (`ProjectStep.tsx:618`, unchanged), and `LaunchFooter` sits outside the scroll container (`SessionWizard.tsx:490`) so the launch button never moves. One minor UX consequence the body does not spell out: because the list no longer collapses, the Extra repos section now sits roughly six rows further down the scroll after you pick a project, which on a short mobile viewport means scrolling to reach it. Acceptable, but worth knowing.

**5. "No matrix edit was needed." — TRUE, verified by running the validator.** `node web/tests/validate-coverage-matrix.mjs` passes. Reading `web/tests/validate-coverage-matrix.mjs:1-16`, the only component-side rule is that every `web/src/components/**/*.tsx` appears in some surface's `components[]`; all three touched components already do (`coverage-matrix.json:833`, `:832`, `:709`). There is no rule requiring a new spec to be registered, and 5 of the 7 existing `ProjectStep.*.test.tsx` specs are likewise unregistered, so leaving the new one out is within existing practice. `WorkspaceSidebar.tsx` is already mapped, so the sessions/sidebar surface needs nothing either.

One cosmetic mismatch: the body ticks "Added or updated a Playwright spec under `web/tests/` **and updated** `web/tests/coverage-matrix.json`". Neither half happened (no new Playwright spec, no matrix edit). The prose two lines below explains the real situation correctly, so this is a checkbox that contradicts its own paragraph.

### The two undocumented commits — the main finding

`WorkspaceSidebar.tsx` (+8/-1) and `mobile-live-no-flutter.spec.ts` (+12/-9) are unrelated to #3461 and appear nowhere in the PR body. CodeRabbit's "Out of Scope Changes" pre-merge check flagged this and I agree with it.

Both are legitimate. I pulled the CI log for the pre-fix SHA `3202482` (run `32696406066`) and both failures are real and match their commit messages precisely:

- Vitest: `AddProjectAction.test.tsx:339` — "ignores Escape and backdrop clicks while the attach is in flight".
- Playwright live: `mobile-live-no-flutter.spec.ts:110` — `expect(jitter).toBeLessThan(3)`, `Received: 19`.

Neither is caused by this PR's feature; both are pre-existing flakes on `main` that this branch happened to hit.

**Why this matters more than "CodeRabbit says split it":** this repo squash-merges (`(#3409)`-style commit titles on `main`). The two commit messages that *do* explain these fixes get collapsed at merge, so the PR body becomes the permanent record, and it says nothing. Six months from now, `git log -S useLayoutEffect` lands on a commit titled "search box for the project picker" with no explanation.

**Ask (non-blocking): add two sentences to the PR body describing both drive-by fixes**, or split them into their own PR. Splitting has a real cost here (both fix currently-red CI), so I would take the body edit.

#### The Escape fix is correctly diagnosed

The mechanism in the commit message checks out. RTL's `waitFor` sets `IS_REACT_ACT_ENVIRONMENT=false` for its duration and resolves off a MutationObserver microtask fired right after React's commit, before the separately scheduled passive-effect flush. So `await waitFor(() => result is showing)` at `AddProjectAction.test.tsx:335` can return while the `document` keydown listener still closes over `busy === true`, and the Escape at line 338 is swallowed with no second chance (`waitFor` retries the assertion, not the keypress). `useLayoutEffect` re-subscribes during the same commit that paints the result, closing the window.

I ran that spec three times with `useEffect` restored on macOS and it passed 3/3, so this is load-dependent and not locally reproducible. The CI log is the evidence, not my local run.

Nit: a `busyRef` read inside a subscribe-once effect would avoid re-subscribing on every `busy` toggle. The layout-effect version is fine and has the in-file precedent; mentioning it only as an alternative.

#### The flutter spec change does NOT weaken the guard

This is the edit that deserved the hardest look, and it survives it. Concretely:

- The assertion is untouched: 25 samples at 100ms, `expect(jitter).toBeLessThan(3)`, plus the pre-existing `expect(ys.length).toBeGreaterThan(0)` false-green guard (`mobile-live-no-flutter.spec.ts:104-110`).
- The settle budget is still 5000ms and still cannot be exceeded. On a genuine regression the loop burns its budget and falls through to the sampling, exactly as before. Nothing was made skippable.
- The gate got **stricter**, not looser: 4 consecutive equal samples at 90ms (270ms of stability) replaces 1 equal pair at 80ms.
- Can a real flutter fake a 4-sample run? No. The fake agent toggles the bottom row every 120ms (`spec:36`, `sleep 0.12`), period ~240ms. Four consecutive equal samples at ~90-100ms requires ~270-300ms of samples to land inside a single 120ms half-cycle. Arithmetically impossible. I verified this by working the phase arithmetic against the fake agent script at `spec:30-38`; I did **not** run the live test (it needs a built `aoe` binary plus tmux).

Honest residual: the diagnosis "sampling started mid-move" is plausible but not proven. A 19px spread across 2.5s of sampling could equally be a single late relayout, which a 4-sample stable run would not prevent either. So this is a strictly-better probabilistic mitigation of a flake whose root cause is inferred rather than pinned. Fine to land; do not treat it as the last word on that spec.

#### One rescoping edit is a strengthening, and the body should say so

`wizard-extra-repos.spec.ts:99`: `/^primary$/` becomes `/^primary/`. A reviewer scanning the diff reads a dropped `$` anchor as a loosened assertion. It is the opposite. The picker chip's `textContent` is name and scope concatenated with no separator (`ExtraReposPicker.tsx:210-213` renders `<span>{p.name}</span><span>{p.scope}</span>`, so `"primaryglobal"`), which means `/^primary$/` matched nothing anywhere and `toHaveCount(0)` was passing **vacuously**. The test named "the primary path is filtered out" was not testing anything. Dropping the anchor gives it teeth. Good catch, but it is buried and unexplained.

The rest of the rescoping is necessary, not cosmetic: `openProjectStepWithPath` (`spec:73-82`) clicks the `/tmp/example` row, so pre-PR `data.path` collapsed `filteredSaved` down to "primary" and `shared-lib`/`docs` existed only inside the picker. Post-PR they render in both places, and a wizard-scoped `getByRole("button").filter({hasText:/^shared-lib/})` matches two elements and trips Playwright strict mode. Reasoned statically from the markup; I did not run Playwright.

### The new Vitest spec: earns most of its 120 lines, not all

Right tier (Vitest, not Playwright: a filter memo needs no real browser, and the house rule prices a mocked Playwright test at ~200x). Nothing restates a constant or a default; no "renders without crashing" smoke test. Good judgment overall.

But four test fns is one too many under the table rule. `"matches on path as well as display name"` (`ProjectStep.search.test.tsx:89-99`) has setup **identical** to test 1 and adds exactly one substring assertion. That is an input permutation, and it belongs as two lines inside test 1 after the `"zeb"` block. Saves ~12 lines and one test fn/symbol. Tests 1, 3, and 4 do have genuinely different setup (`fetchProjects` mock differs; the empty-state assertion differs), so three fns is defensible.

Smaller nit: the mock-and-render preamble is copy-pasted into all four; a `setup({ saved })` helper trims another ~10 lines.

### Compliance with project guidelines

| Gate | Result |
|---|---|
| Diff is web-only | Confirmed, 7 files, all under `web/`. No Rust implicated. |
| `npm run format:check` (oxfmt) | **Ran it. Clean**, 931 files. |
| `npm run lint` (ESLint) | **Ran it. Clean.** |
| Vitest | **Ran** the 17 `session-wizard/__tests__` specs + `AddProjectAction.test.tsx`: 18 files / 122 tests pass, tsc typecheck no errors. |
| Coverage matrix | **Ran** the validator. Passes. |
| codecov/patch (75%) | Green per the bot: "All modified and coverable lines are covered by tests." |
| No em-dashes in added prose | Confirmed by grepping added lines. Clean. |
| Conventional commits | All five commits conform. |

Not run: Playwright (mocked or live), the full Vitest suite, anything Rust.

### Automated review check-in

- **CodeRabbit "Out of Scope Changes" warning** — **agree**, and it is my main finding. See above.
- **CodeRabbit "Title check" (not imperative mood)** — **disagree as blocking.** `main`'s own recent history is full of non-imperative titles (`fix(tmux): omp provider errors and retries no longer read Idle`, `chore(deps): keep pre-1.0 web packages out of...`). The bot is applying a rule this repo does not follow.
- **CodeRabbit line-level findings** — none generated.
- **Codecov** — green on both gates.

### Nits (none blocking)

1. `ProjectStep.tsx:125` — `mergeRecentProjects`'s doc says "the caller still slices to the visible cap". Still technically true, but a reader following it to line 191 now finds no slice there. One-word tweak keeps it honest.
2. `ProjectStep.tsx:411` — the "Recent" section header renders only when `filteredSaved.length > 0`. A query matching recents but no saved projects renders the recents unlabeled. Pre-existing, but the search box makes that state reachable far more often than before. Cheap fix: also render it when a query is active.
3. No debounce on `query`. Correct call at these list sizes (pure in-memory filter over 50-odd rows); noting it only so it is a decision rather than an oversight.
4. #3461 asked for search "independent of the arbitrary recency cap", and one arbitrary cap survives: `RECENT_PROJECTS_CAP = 20` (`src/session/storage.rs:839`) bounds what a search can reach for a project with no live session. The body's "reaches every project the dashboard knows about" is literally true but reads stronger than it is. Worth a sentence so the issue is not closed as more complete than it is.

### Merge risk: **Low**

Blast radius is one wizard tab; a wrong filter shows too few rows, which is visible immediately and has an obvious workaround (clear the box, or use Browse). Reverts cleanly: no schema, no migration, no persisted state.

The one path with no test coverage is the settle-loop change itself: the live flutter spec cannot assert that its own settle gate is correct, only that the jitter bound holds afterward. If the 19px spread was a late relayout rather than a slow initial converge, that spec will flake again and we will hear about it from a red CI run, not from a user. That is the same place we would have heard about it before, so this is not a new exposure.

---

### Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** PR body needs the two drive-by fixes described; no tree docs needed
**Merge risk:** Low

## **LGTM**

The feature does what it claims, all five body claims hold up under checking, and every gate that could be run locally is green. The only thing to ask for before merge is two sentences in the PR body covering the `WorkspaceSidebar` Escape fix and the flutter-settle change (and ideally a note that the dropped `$` anchor in `wizard-extra-repos.spec.ts:99` fixes a vacuous assertion). Because this repo squash-merges, the commit messages that currently explain them will not survive, and the body is what remains. The flutter edit specifically does not weaken the guard; the arithmetic was checked and the assertion is untouched.

PR #3507, opened 2026-08-24
fix(session): keep co-located Claude panes on their own conversations

jona-wilmsmann:

**This implementation and pull request were created with Claude Code. The code has not been line-reviewed by a human.**

We are making it available because it is the local patch we are currently running against a session-mismatch problem we hit in production, and it may be of interest to the maintainers. We do not claim it is the preferred upstream design. Please treat it as a tested candidate fix plus a reproduction, not as an assertion that this is the right final implementation.

## Description

When several AoE sessions share one working directory, they share one `<claude-config>/projects/<encoded-cwd>/` transcript directory. Attribution of a transcript to a pane currently falls back to "most recently modified `*.jsonl` in that directory", which cannot distinguish this pane's conversation from a peer's.

We hit this after a machine restart. Ten sessions on one cwd recovered at once; every hook sidecar had expired (they are age-gated at 5 minutes, and a reboot expires all of them simultaneously), so every pane fell through to the mtime scan. Each resume touched a transcript, so "newest jsonl" changed every few hundred milliseconds and the panes cross-assigned. Observed outcome, from `ps` on the running panes:

```
aoe session A -> claude --resume <uuid-1>   (session E's conversation)
aoe session B -> claude --resume <uuid-1>   (the same one again)
aoe session C -> claude --resume <uuid-2>   (session F's conversation)
aoe session D -> claude --resume <uuid-2>   (the same one again)
```

Two conversations were each open in two panes, appending to one transcript; four other sessions had their stored `agent_session_id` overwritten with a peer's. The logs show the swap directly: `Session ID observed for <instance>: <sid>` for three different sids on one instance within four seconds, and `Ignoring poller-reported sid already owned by another instance` as peers raced for the same file.

This is not restart-only. In steady state the same fallback re-targets an idle session every time its sidecar ages past five minutes: our log has one instance flip-flopping between seven different sids over a single day.

This patch:

- **stops the mtime scan from re-targeting a session whose transcript directory it shares.** An anchored caller in a shared directory gets its own transcript or nothing. A newer sibling jsonl is not evidence that *this* pane changed conversations; with several panes on one namespace it usually just means a peer is busy.
- **stops age-gating the anchor.** A pane's own conversation going quiet for five minutes does not make a peer's transcript the better resume target.
- **restricts `.claude.json`'s `lastSessionId` to anchorless callers and to sole writers.** It is per-project, not per-pane: whichever Claude ran last in that cwd owns it, so a pane sharing the directory with a peer, holding a pinned `--session-id` that has not written its first transcript, would otherwise adopt that peer's conversation.
- **removes the 5-minute expiry on the per-instance hook sidecar.** It is written by this pane's own `SessionStart`/`UserPromptSubmit` hooks, so it names the conversation Claude last ran here; idling does not make it wrong, and the only alternative on expiry is the scan that cannot attribute at all.
- **keeps the #2291 mtime re-target for the single-session-per-namespace case**, on the host, in a container, at acquisition and in the poller. That is where the scan is sound and where it is what promotes a stored id after `/clear`. It is disabled only where another AoE session writes into the same directory, which is exactly when it becomes a lottery.

**Second round** (@jerome-benoit), closing two gaps that both let the mtime retarget back in:

- **The sandbox path.** I had scoped this to the host and argued on the CodeRabbit thread that container transcript directories are private. That was wrong, and I have retracted it there. `sandbox_dir_for` appends the instance id only for Codex, so every Claude container binds the shared `$HOME/.claude/sandbox` RW and co-located sandboxed panes share one transcript namespace. The per-instance hook directory *is* bind-mounted into the container, so the sidecar is readable host-side; the doc comment on `claude_poll_fn_sandboxed` claiming otherwise was stale and is corrected. `claude_poll_fn_sandboxed` now reads the sidecar first, and the container listing emits a held anchor regardless of mtime, since the freshness gate was removing it outright.
- **The sharing predicate's namespace.** It was keyed to the AoE profile while the transcript namespace is keyed by the effective Claude home, so two profiles resolving to the same home and cwd bypassed the guard. It now walks every profile, and fails closed: unreadable profile list, unreadable store, unresolvable home all return "shared", because not being able to prove sole ownership must not license an mtime retarget.

**Third round** (@jerome-benoit at `fe153a39`, plus CodeRabbit's still-open symlink finding), four findings, all in the predicate and the anchor it gates. Every one of them was correct:

- **Peers were resolved through the wrong profile.** `source_profile` is runtime-only and `Storage::load` does not stamp it, so every peer the walk loaded resolved its Claude home through the *default* profile rather than the profile it is stored in. The round-two commit said it compared "each peer on its own resolved home"; it did not. Two sessions in one non-default profile, or in two profiles pointing at one home, were judged isolated and the retarget came straight back. The walk now stamps the profile it loaded from.
- **Canonical host cwd equality was the wrong comparison, in both directions.** `encode_claude_project_path` maps every non-alphanumeric character to `-`, so `/tmp/a_b` and `/tmp/a/b` land in one directory while comparing unequal; and a sandboxed session's transcripts are keyed by its `container_workdir` against the shared sandbox store, so `/repos/a/foo` and `/repos/b/foo` both become `/workspace/foo`. Comparison is now on a `ClaudeTranscriptNamespace`: the canonicalized backing store, which also settles CodeRabbit's symlinked-home finding, plus the encoded cwd the agent actually sees.
- **#2291 had been removed from the poller, not just from the shared case.** This is the one I am most glad was caught, because the PR description above asserted a guarantee the code did not honour. `claude_poll_fn` passes its `last_known` on every tick, so the unconditional anchor hold left an isolated single-session cwd unable to follow `/clear` or `/new` whenever the hook sidecar is unavailable (`agent_status_hooks = false`, or hooks that do not fire). The anchor now carries its policy: `ClaudeAnchor::Held` for a shared namespace, `MtimeMayRetarget` for a proven sole writer, which restores the pre-PR ordering, the container listing's pure `ls -tL` order, and the `.claude.json` arm.
- **The doc comment was orphaned again.** Inserting the predicate above `compose_exclusion_with_persisted_peers` left that function's documentation attached to the new one, the same slip this PR already fixed once in `instance.rs`. The predicate now sits after it.

**One residual worth naming rather than burying.** The sole-writer verdict is resolved once, where the poller is started, because the predicate walks every profile's `sessions.json` and that is far too expensive for a poll loop. A peer created on this cwd *after* a poller is already running therefore does not flip that poller's verdict. What covers the new peer is the per-tick exclusion set, which drops its sid as soon as it publishes one to tmux env; the residual window is a peer that is writing but has not yet captured. That is narrower than the pre-PR behaviour, but it is not nothing, and I would rather have it in the description than discovered. Happy to make the verdict lazy (resolved only when a scan actually proposes a promotion, then latched) if you want it closed here rather than as a follow-up.

The wider class you identified (OpenCode, Vibe, host Codex, Gemini, Pi, Hermes; MRU selectors in host Copilot and Kimi; OMP and sandboxed Codex unaffected) is filed as #3516 rather than folded in here.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (4,672 library tests pass; one unrelated pre-existing failure, see below)
- [x] Documentation was updated where necessary (no user-facing documentation change needed; the behavioural contracts are documented at the call sites)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped an e2e test, because: the decision under test is a pure capture/attribution choice, exercised deterministically by unit fixtures that stage a shared project directory with staggered mtimes. Reproducing it end-to-end would require several concurrent live agent panes on one cwd.

Added regressions:

- `capture::tests::test_capture_claude_session_mass_recovery_keeps_panes_apart` — the storm itself: three anchors in one project dir, all sidecars expired, one peer written during recovery; every pane must still get its own conversation back.
- `instance::tests::…::mtime_fallback_declines_when_project_dir_is_shared` — a peer sharing the cwd whose own stored id is already stale, so the freshest transcript is in no exclusion set.
- `capture::tests::shared_check_resolves_each_peer_through_its_own_profile` — the profile-stamping fix, both directions, with the default profile's home pinned so neither assertion can pass on profile ordering.
- `capture::tests::transcript_namespace_compares_encoded_cwds_not_host_paths` — the encoder collision and the container-workdir collision.
- `capture::tests::sole_writer_lets_a_newer_transcript_retarget_the_anchor` and `capture::tests::container_selection_retargets_only_where_the_namespace_is_isolated` — the no-hooks `/clear` case on the host and in a container, which is what #2291 asks for.
- `tmux`-free unit coverage for the anchor rules (`…_keeps_stale_anchor`, `…_anchor_wins_over_newer_sibling`, `claude_poll_fn_holds_anchor_across_polls`, `claude_poll_fn_uses_sidecar_regardless_of_age`).

Each of the four new regressions was run against the code with its own fix reverted, and fails there. The profile-stamping one is worth calling out: my first version of it passed either way, because an unstamped peer resolves through whichever profile `resolve_default_profile` picks and in that fixture it happened to be the right one. It only became a real test once the fixture pinned `default_profile` to a third profile with its own Claude home.

Five existing tests encoded the behaviour this patch inverts and were updated with their rationale rather than deleted.

Commands run on this branch (based on `9e970918`):

- `cargo clippy -- -D warnings` (the CI form)
- `cargo clippy --lib --tests -- -D warnings` — clean except the pre-existing `items_after_test_module` in `src/process/macos.rs`, identical on `main`
- `cargo test --lib` — 4,672 passed, 1 failed (below), 2 ignored
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve` — clean

One note on running these locally, unrelated to this PR: `rust-toolchain.toml` pins `channel = "stable"`, and since #3499 the `serial_test 4.0.1` dev-dependency requires rustc 1.93.1, so any `stable` older than that cannot build the test targets at all. I ran the above on 1.95.0.

`git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` fails on this machine, and fails identically on an unmodified checkout of the same commit — it compares `/var/...` against macOS's `/private/var/...`. Unrelated to this change.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Claude investigated the live incident from the AoE debug log, `ps` output of the running panes and the on-disk transcripts, reconstructed the correct session-to-conversation mapping, wrote the patch and its regressions, and ran the checks above. Before narrowing the change it found the existing tests that asserted the old mtime behaviour and treated them as a deliberate contract (#2291) to preserve where it is sound, rather than as failures to delete.

## Summary

- Fixed Claude session attribution when multiple sessions share a directory or transcript namespace.
- Preserved valid session anchors without age-based expiry.
- Limited fallback retargeting to namespaces confirmed as sole-writer.
- Added shared-directory detection across profiles, sandbox environments, symlinks, and encoded path collisions.
- Updated polling and container capture flows to apply the same anchor rules.
- Added regression tests for stale anchors, namespace collisions, profile resolution, sandbox sessions, and sole-writer retargeting.
- Reported checks include 4,672 passing library tests. One unrelated macOS path canonicalization test remains failing.

## Benefits

- Reduces incorrect session attribution.
- Prevents stale or shared transcripts from claiming another session.
- Preserves recovery behavior for isolated Claude sessions.
- Improves fail-closed behavior when ownership is uncertain.


src/session/capture.rs
```diff
@@ -140,12 +140,30 @@ pub(crate) fn capture_claude_session_id(
...
+    // `lastSessionId` is per-project, not per-pane: whichever Claude ran last in
+    // this cwd owns it. A pane that already has a session of its own must not
+    // consult it, or it adopts a peer's conversation the moment its own
+    // transcript is missing (a pinned `--session-id` before the first prompt).
+    if known_session_id.is_some() {
+        anyhow::bail!(
+            "No transcript yet for the session anchored to {}",
+            project_path
+        );
+    }
```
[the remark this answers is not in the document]
jona-wilmsmann: Checked this against the container path and I don't think it applies — declining, but here's the reasoning so you can push back if I've got it wrong.

The host fix is specifically about a *shared* transcript directory: several AoE sessions on one cwd all write into one `~/.claude/projects/<encoded-cwd>/`, which is what makes mtime unable to attribute a file to a pane. That premise doesn't hold in a container. The container name is derived from the session id (`format!("aoe-sandbox-{}", truncate_id(session_id, 8))`, `src/containers/mod.rs`), so it's one container per session, and nothing bind-mounts the Claude home from the host by default. Every transcript in that directory therefore belongs to the same pane — its current conversation and its own earlier ones. There is no peer conversation there to retarget to.

Worse, "newest wins" is the *only* adoption mechanism the sandboxed path has. `claude_poll_fn_sandboxed` deliberately does not read the host hook sidecar (its doc comment explains why: the in-container hook writes to the container's own `/tmp/aoe-hooks/`, which the host poller can't see). On the host, `/clear` promotion arrives on that sidecar, which is what makes it safe to stop the mtime scan from retargeting. Take the mtime retarget away in the container and there's no fallback left — sandboxed `/clear` and `--fork-session` would stop being adopted at all.

The one case where your concern would be real is a user who bind-mounts a shared Claude home into several containers via `sandbox.extra_volumes`. That's non-default and, if it's worth handling, I'd rather do it as its own change with the sharing detection made explicit, rather than by making the container path unconditionally prefer the anchor.

src/session/capture.rs
```diff
@@ -981,6 +995,35 @@ pub(crate) fn compose_exclusion(
...
+    let Ok(instances) = storage.load() else {
+        return false;
+    };
+    let canonical_current = canonicalize_or_raw(current_project_path);
+    instances.iter().any(|inst| {
+        inst.id != current_instance_id
+            && inst.tool == "claude"
+            && !inst.is_trashed()
+            && canonicalize_or_raw(&inst.project_path) == canonical_current
+    })
```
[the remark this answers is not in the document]
jona-wilmsmann: Declining this one, because a liveness condition would reintroduce the bug in exactly the scenario the PR is about.

The failure is a *concurrent start*. On the recovery this came from, ten sessions on one cwd were started within about fifteen seconds. At the moment the first of them runs `acquire_session_id_with`, none of its peers has a tmux pane yet — they haven't been started. So `has_live_tmux_pane()` is false for all of them, the predicate returns `false`, and that session takes the single-session mtime path and can adopt a peer's transcript. Which is what happened: in our log the very first session recovered picked up a wrong sid two seconds in, before any peer had claimed anything. Liveness is precisely the property that is racing during the storm, so it can't be the thing that decides whether the storm is possible.

`Status::Stopped` has the same problem from the other direction — a session recovered from a stopped state is stopped right up until it starts, which is when this predicate runs.

The cost of the conservative answer is bounded: when the directory is shared we require the hook sidecar to retarget, and the sidecar covers `/clear` whenever hooks are installed (`session.agent_status_hooks`, on by default). A stale `true` from a stopped peer means that cwd relies on the sidecar rather than mtime — it does not block capture, and it never resumes the wrong conversation. The reverse error does.

I did exclude trashed sessions for this reason, since those need explicit user action to come back. Excluding archived ones too would be defensible on the same grounds — `startup_recovery` skips them, so they can't be part of a concurrent start. Happy to add that if you'd like the predicate tightened that far; I left it out because it's one more state to keep in sync for a small gain.

src/session/instance.rs
```diff
@@ -3123,6 +3137,16 @@ impl Instance {
...
+    /// which makes transcript mtime useless for attributing a conversation.
+    fn claude_project_dir_is_shared(&self) -> bool {
+        super::capture::claude_project_dir_is_shared(
+            &self.id,
+            &self.project_path,
+            &self.effective_profile(),
+        )
+    }
+
     fn retroactive_capture_exclusion_set(&self) -> HashSet<String> {
```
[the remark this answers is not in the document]
jona-wilmsmann: Good catch, and correct — that was mine. Inserting the new function between the doc comment and `retroactive_capture_exclusion_set` left the exclusion-set description sitting above `claude_project_dir_is_shared`, and the function it belonged to with nothing.

Fixed in 39c066cb, arranged as you suggested: the one-liner stays with `claude_project_dir_is_shared` and the original comment moves back down to `retroactive_capture_exclusion_set`.

src/session/instance.rs
```diff
@@ -3007,7 +3007,21 @@ impl Instance {
...
+            // Claude's non-hook capture is an mtime scan of
+            // `<config>/projects/<encoded-cwd>/`. With one session per cwd that
+            // scan is sound and it is what re-targets a stored id after `/clear`
+            // (#2291). With several sessions on one cwd it is a lottery: they
+            // share the directory, so on a mass recovery every pane sees the
+            // same "newest" transcript and resumes it, swapping conversations
+            // between panes and leaving one conversation open in two. Where the
+            // directory is shared, only the per-instance hook sidecar may
+            // retarget an id we already hold.
+            let observed = if self.tool == "claude" && self.claude_project_dir_is_shared() {
```
jerome-benoit: **[P1] This guard does not cover the sandboxed Claude poller.** `sandbox_dir_for` only gives Codex an instance-suffixed directory; Claude containers bind the shared `$HOME/.claude/sandbox` directory RW. `claude_poll_fn_sandboxed` then skips the per-instance sidecar and `select_claude_session_in_container` explicitly returns a newer unexcluded candidate before the known anchor. As a result, this acquisition-time branch can preserve the correct ID and the later sandbox poller can still retarget it to a peer. Please make the sandbox poller consume the authoritative per-instance sidecar (the hook directory is already bind-mounted), and otherwise hold a known anchor instead of applying MRU promotion. A regression test should cover two sandboxed Claude instances sharing one cwd with a newer peer transcript.

src/session/capture.rs
```diff
@@ -981,6 +995,35 @@ pub(crate) fn compose_exclusion(
...
+/// identifying a pane — during a mass recovery every pane sees the same newest
+/// transcript, and the peer-exclusion sets cannot help because no peer has
+/// claimed anything yet. Callers use this to refuse an mtime guess that would
+/// retarget a session already holding a conversation of its own.
+pub(crate) fn claude_project_dir_is_shared(
+    current_instance_id: &str,
+    current_project_path: &str,
+    profile: &str,
+) -> bool {
+    let Ok(storage) = crate::session::storage::Storage::new_unwatched(profile) else {
```
jerome-benoit: **[P1] The safety decision is scoped to an AoE profile, not to the actual Claude transcript namespace.** This loads only `sessions.json` for `profile`, while the comment immediately above acknowledges that host transcript directories are keyed by `$HOME`. Two profiles using the same effective Claude home and cwd therefore return `false` here and `acquire_session_id` falls back to the MRU scan again. Please derive sharing from the effective Claude store across profiles, or fail closed and refuse MRU retargeting whenever ownership cannot be established globally.

jerome-benoit (changes requested):
The host-side anchor change is sound, but two safety gaps remain before this fixes co-located Claude recovery end to end.

1. Sandboxed Claude instances still share the same RW `$HOME/.claude/sandbox` store. Their poller does not read the per-instance hook sidecar and `select_claude_session_in_container` still intentionally promotes a newer candidate over the known anchor. That preserves the original cross-adoption failure on the sandbox path.
2. Shared-store detection only loads the caller's AoE profile even though the transcript namespace is keyed by the effective Claude home. Co-located sessions in different profiles therefore bypass the new authoritative-only branch.

I also checked the equivalent recovery paths for the other agents. OpenCode, Vibe, host Codex, Gemini, Pi, and Hermes still inherit stored-ID to freshest-observation replacement, with analogous MRU-by-cwd selectors also present for host Copilot and Kimi. This predates this PR, so I am not making it a blocker here, but it needs a follow-up rather than assuming the class is Claude-only. OMP is not affected because it uses pane-generation breadcrumbs; sandboxed Codex is instance-private.

I did not duplicate the existing documentation and `EnvGuard` cleanup comments.

Validated at this exact head: `git diff --check`, `cargo fmt --all -- --check`, `cargo clippy --lib -- -D warnings`, and `cargo test --lib` (`4662 passed; 0 failed; 2 ignored`).

src/session/capture.rs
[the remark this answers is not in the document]
jona-wilmsmann: Retracting my previous reply — it was wrong on both counts, and @jerome-benoit's review has the correct reading.

I claimed the container transcript directory is private per session. It isn't: `sandbox_dir_for` appends the instance id only for `codex` (`src/session/container_config.rs:727`), so every Claude container binds the shared `$HOME/.claude/sandbox` RW. Sandboxed sessions on one cwd share a transcript namespace exactly like host sessions do.

I also claimed there is no sidecar to fall back on in a container, quoting the doc comment on `claude_poll_fn_sandboxed`. That comment is stale: the per-instance hook directory *is* bind-mounted into the container (`container_config.rs:1801`, `read_only: false`), so the host poller can read a sandboxed pane's `session_id` sidecar. The `/clear` adoption route I said didn't exist is available.

The method error is worth naming, since it is what produced a confident wrong answer: I grepped `src/containers/` for a `.claude` mount, found none, and treated absence in a narrow search as evidence of absence. The mount is configured in `src/session/container_config.rs`.

So your original finding stands. I'm fixing the sandbox path in this PR: the sandboxed poller will consume the per-instance sidecar, and the container selection will hold a known anchor instead of promoting a newer candidate, with a regression test covering two sandboxed instances sharing one cwd.

src/session/capture.rs
```diff
@@ -1001,6 +1047,59 @@ pub(crate) fn compose_exclusion(
...
+            return true;
+        };
+        for inst in instances {
+            if inst.id == current_instance_id || inst.tool != "claude" || inst.is_trashed() {
+                continue;
+            }
+            if canonicalize_or_raw(&inst.project_path) != canonical_current {
+                continue;
+            }
+            let Ok(peer_home) = claude_home_for_host_environment(&inst.resolved_host_environment())
```
jerome-benoit: [P1] Resolve each peer using the profile currently being iterated. `source_profile` is runtime-only (`#[serde(skip)]`), and `Storage::load()` does not stamp it, so every `inst` loaded here has an empty profile. `resolved_host_environment()` therefore falls back to the global default instead of `profile`. Two sessions stored in the same non-default profile, or in non-default profiles sharing one Claude home, can be compared against the wrong home and incorrectly classified as isolated; acquisition then re-enables the anchorless MRU scan this PR is meant to block. Stamp `inst.source_profile = profile.clone()` before resolution, or resolve the environment directly from `profile`, and add a regression with default home A and two non-default peers sharing home B.

src/session/capture.rs
```diff
@@ -1001,6 +1047,59 @@ pub(crate) fn compose_exclusion(
...
+            return true;
+        };
+        let Ok(instances) = storage.load() else {
+            return true;
+        };
+        for inst in instances {
+            if inst.id == current_instance_id || inst.tool != "claude" || inst.is_trashed() {
+                continue;
+            }
+            if canonicalize_or_raw(&inst.project_path) != canonical_current {
```
jerome-benoit: [P1] Compare the actual transcript namespace, not canonical host cwd equality. Sandboxed Claude instances share the same host `~/.claude/sandbox` backing store, while their transcript key is derived from `container_workdir`; distinct host paths such as `/repos/a/foo` and `/repos/b/foo` both become `/workspace/foo` and therefore share one transcript directory even though this guard skips the peer. The host encoder is also non-injective (`/tmp/a_b` and `/tmp/a/b` both encode to `-tmp-a-b`). In both cases this false negative licenses an anchorless MRU retarget and can resume a peer's conversation. Compare an identity made from the effective backing store plus the encoded effective cwd, with a sandbox-specific branch, and cover both collision classes.

src/session/capture.rs
```diff
@@ -144,12 +144,30 @@ pub(crate) fn capture_claude_session_id(
     if let Some((id, modified)) =
         scan_claude_project_dir(&claude_home, &canonical, known_session_id, exclusion)?
     {
+        // The anchor is this pane's own conversation, so its age says nothing
+        // about whether it is the right target: an idle session is still ours.
+        // The freshness gate belongs to the anchorless guess below, which is
+        // only a guess.
+        if known_session_id == Some(id.as_str()) {
```
jerome-benoit: [P1] Preserve the retarget path for a namespace proven to be isolated. `claude_poll_fn` passes its `last_known` value on every poll, so this unconditional anchor return means that with supported `agent_status_hooks = false`, or when hooks are unavailable, `/clear` or `/new` can create a newer transcript but the poller keeps returning the old UUID indefinitely. This contradicts the PR's stated guarantee that the #2291 single-session fallback remains available. Gate anchor-only selection on proven sharing, or give the poller an explicit isolated mode, and add a no-hooks single-session regression covering `/clear`/`/new`.

jerome-benoit (changes requested):
Re-review at `fe153a39`. The previous sandbox-poller issue is addressed: the per-instance sidecar is read first and an existing container anchor is no longer displaced by a newer sibling transcript. However, the acquisition-time sharing predicate still resolves non-default peers through the wrong profile and compares host paths rather than the effective transcript namespace. The change also removes the supported isolated-session retarget path when hooks are disabled or unavailable. Please address the three inline findings.

The existing CodeRabbit symlink-home finding remains valid and is not duplicated here. `git diff --check` and `cargo fmt --all -- --check` are clean at this head; GitHub has not run the Rust CI jobs on it.

src/session/capture.rs
jona-wilmsmann: Correct, and the round-two commit's description was wrong about it: it claimed to compare "each peer on its own resolved home" while every peer resolved through `resolve_default_profile()`. Fixed in `f4fd43a2` by stamping `inst.source_profile = profile.clone()` before resolving, with a comment saying why the field is empty in the first place.

One detail worth recording for the next reader: the fallback is `resolve_default_profile()`, not a fixed "default" profile, so an unstamped peer resolves through the configured default or, failing that, the first profile listed. That is why my first attempt at the regression passed with and without the fix: in the fixture the fallback happened to land on the right home. The committed test pins `default_profile` to a third profile with its own Claude home, so both assertions now fail without the stamp.

`shared_check_resolves_each_peer_through_its_own_profile` covers both directions: a peer in a different non-default profile resolving to the same home must read as shared, and a peer whose home really is the default's must not make an unrelated profile read as shared.

src/session/capture.rs
jona-wilmsmann: Agreed on both collision classes, fixed in `f4fd43a2`. The comparison is now on a `ClaudeTranscriptNamespace` rather than on canonical host paths:

- **Host**: the resolved Claude home, canonicalized, plus `encode_claude_project_path` of the canonical cwd. Canonicalizing the home also settles CodeRabbit's symlinked-home finding, and comparing the *encoded* cwd rather than the path is what closes the non-injective encoder: `/tmp/a_b` and `/tmp/a/b` now compare equal, in the conservative direction.
- **Sandbox**: the encoded `container_workdir` alone, since `sandbox_dir_for` suffixes only Codex and every sandboxed Claude binds the one shared store. `/repos/a/foo` and `/repos/b/foo` both reduce to `/workspace/foo` and now read as shared.

A host and a sandboxed session are never equal, which is the other half of "the sandbox store is not the host's".

`transcript_namespace_compares_encoded_cwds_not_host_paths` covers all three.

src/session/capture.rs
jona-wilmsmann: You are right, and this one was the worst of the three because the PR description asserted the opposite: it claimed the #2291 fallback was kept for the single-session case while `claude_poll_fn` passed `last_known` unconditionally, so an anchored poller could never retarget, shared or not. With `agent_status_hooks = false` that leaves `/clear` and `/new` stuck on the old UUID indefinitely.

Fixed in `f4fd43a2` by giving the anchor its policy rather than gating at the call site, so the two modes cannot drift apart across the four places that consult it:

```rust
enum ClaudeAnchor<'a> { None, MtimeMayRetarget(&'a str), Held(&'a str) }
```

`MtimeMayRetarget` restores the pre-PR behaviour exactly: the `scan_claude_project_dir` ordering, the container snippet's pure `ls -tL` listing with no anchor line prepended, `select_claude_session_in_container`'s position tie-break, and the `.claude.json` arm, which a sole writer may consult again because "whichever Claude ran last in this cwd" is then this pane. `Held` keeps everything this PR added.

Regressions on both sides: `sole_writer_lets_a_newer_transcript_retarget_the_anchor` for the host and `container_selection_retargets_only_where_the_namespace_is_isolated` for the container, each asserting both modes against one fixture.

One residual I would rather state than have you find: the verdict is resolved once, where the poller is started, because the predicate walks every profile's `sessions.json` and that cannot go in a poll loop. A peer created on the cwd after a poller is already running does not flip that poller's verdict; what covers it is the per-tick exclusion set, once the peer publishes a sid to tmux env. The window is a peer that is writing but has not captured yet. If you want that closed here rather than as a follow-up, the shape I would use is a lazily resolved verdict, evaluated only on a tick where the scan actually proposes a promotion and then latched to "shared", so the walk costs nothing in steady state. Say the word and I will add it.

src/session/capture.rs
```diff
@@ -1001,6 +1047,59 @@ pub(crate) fn compose_exclusion(
...
+            }
+            if canonicalize_or_raw(&inst.project_path) != canonical_current {
+                continue;
+            }
+            let Ok(peer_home) = claude_home_for_host_environment(&inst.resolved_host_environment())
+            else {
+                return true;
+            };
+            if peer_home == current_home {
+                return true;
```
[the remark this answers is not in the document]
jona-wilmsmann: Valid, and fixed in `f4fd43a2`, though not at this line: the comparison moved to a `ClaudeTranscriptNamespace` whose host variant holds `canonicalize_or_raw(home)`, so a symlinked `CLAUDE_CONFIG_DIR` and its target now compare equal. The same change covers a second collision class @jerome-benoit raised, where the encoded cwd matters rather than the path.

Cross-profile coverage is `shared_check_resolves_each_peer_through_its_own_profile`, which stages two non-default profiles resolving to one Claude home.

src/session/capture/mod.rs
```diff
@@ -295,27 +376,40 @@ fn scan_claude_project_dir(
...
-    if let Some((kid, kmt)) = known_hit {
-        let known_fresh = kmt
+    let Some((kid, kmt)) = known_hit else {
+        // A held anchor gets its own transcript or nothing. Reaching here means
+        // it is missing (a pinned `--session-id` before its first prompt) or
+        // excluded (claimed by a peer); `best` is then just whichever peer
+        // wrote last, which is never a safe substitute for a session this pane
+        // owns. The sole writer of the directory has no such peer, so `best`
+        // there is its own.
+        if anchor.is_held() {
```
jerome-benoit, line 388: **[P1] Distinguish a provisional launch UUID from a confirmed anchor**

This `Held` return breaks the existing shared-project correlation path when the stored ID is only Claude's launch-minted UUID and has no transcript. In the live `Cargo Test (linux, serve-e2e)` job ([98316042458](https://github.com/agent-of-empires/agent-of-empires/actions/runs/32953013107/job/98316042458)), `claude_shared_project_correlation_variant2_filesystem_scan_exclusion` fails with 194 tests passing and one failing: session A remains on random ID `91ee0daf-...` instead of correlating to its own `aaaaaaaa-...` transcript after B's established ID is excluded. Model provisional IDs separately from confirmed anchors and allow correlation only after complete peer-ownership elimination; also keep a negative case where a peer is not established so mass recovery still fails closed.

src/session/capture/mod.rs
```diff
@@ -120,6 +120,54 @@ pub(crate) fn encode_claude_project_path(project_path: &str) -> String {
...
+    /// Held, and sharing the directory with another AoE session: this pane
+    /// resumes its own conversation or none. Promotion after `/clear` arrives
+    /// on the per-instance hook sidecar, which is attributable.
+    Held(&'a str),
+}
+
+impl<'a> ClaudeAnchor<'a> {
+    pub(crate) fn new(id: Option<&'a str>, project_dir_is_shared: bool) -> Self {
+        match (id, project_dir_is_shared) {
+            (None, _) => Self::None,
```
jerome-benoit, line 154: **[P1] Preserve shared-namespace policy when there is no anchor**

`(None, _)` discards whether the namespace is shared, so `try_retroactive_capture()` can still use the MRU transcript with `ClaudeAnchor::None`. Its persisted exclusions only inspect the caller's profile and compare canonical host paths, which misses stopped cross-profile peers sharing one Claude home and sandbox peers whose distinct host paths map to one `container_workdir`. A restored live session without `agent_session_id` can therefore adopt a peer's transcript. Represent the shared-unanchored state explicitly and require an attributable sidecar or complete ownership proof instead of an MRU guess.

src/session/instance.rs
```diff
@@ -5691,6 +5726,13 @@ impl Instance {
 
         let poll_fn: Box<dyn Fn() -> Option<String> + Send + 'static> = match tool {
             "claude" => {
+                // Resolved once here, not per tick: the predicate walks every
+                // profile's `sessions.json`, which is far too expensive for a
+                // poll loop. A peer created on this cwd after the poller starts
+                // therefore does not flip an already-running "sole writer"
+                // verdict; what covers that peer is the per-tick exclusion set,
+                // which drops its sid as soon as it publishes one to tmux env.
+                let project_dir_is_shared = self.claude_project_dir_is_shared();
```
jerome-benoit: **[P2] Revalidate sharing before accepting an mtime retarget**

This snapshot is not a durable property of the poller. If a peer enters the namespace later, the old poller keeps `MtimeMayRetarget`; during the interval between that peer creating a new transcript and its new SID passing the sync guards, CAS, and tmux-env publication, the old poller can adopt the still-unowned SID. The per-tick exclusion set cannot cover an ID that has not been published yet. Recheck the namespace only when the scan proposes replacing the anchor, then latch or invalidate the policy, so the expensive profile walk is not paid on stable ticks.

jerome-benoit (changes requested):
Re-review at `f4fd43a2`. The previous findings around the sandbox sidecar, profile stamping, effective transcript namespaces, the isolated #2291 retarget path, and the misplaced documentation are addressed. The old sandbox thread is no longer actionable, and I did not duplicate CodeRabbit's existing `EnvGuard` note.

Three correctness gaps remain, including a deterministic failure in the live `Cargo Test (linux, serve-e2e)` job. See the inline comments.

Non-blocking documentation follow-up: `src/session/capture.rs:450-453` and `src/session/instance.rs:3566-3569` still describe the sidecar as expiring after five minutes, but the reader no longer applies an age limit. Please update those comments with the behavioral fix.

PR #3508, opened 2026-08-24
perf(tmux): take one liveness snapshot per pass, not per instance

jona-wilmsmann:

**This implementation and pull request were created with Claude Code. The code has not been line-reviewed by a human.**

It is the local patch we are currently running. Please treat it as a measured candidate fix rather than an assertion that this is the right final implementation.

## Description

`Instance::has_live_tmux_pane()` forks a full `tmux list-sessions` on every call, via `tmux_env_session_name_for_instance_id`. Several passes call it once per instance — notably `compose_exclusion_with_persisted_peers`, which walks **every stored session sharing the project path, trashed ones included**. On a store with a few hundred sessions that turns one pass into a few hundred `fork`+`exec` round-trips.

This is visible as input lag in the TUI: keystrokes land on a thread that is blocked waiting on those subprocesses.

Measured on macOS with 169 stored sessions (11 live, 158 trashed) on one project path, by shimming `tmux` to log its argv:

```
568 tmux invocations in 6s — of which 520 were the identical
    list-sessions -F '#{session_name}'
```

One such call costs 4.99 ms here (mean of 100), so ≈ 87 calls/s ≈ **470 ms of subprocess round-trip per wall-clock second**. `sample` on the TUI agrees independently: the main thread sits in `poll` on a child pipe for 45% of samples. The idle TUI drew 9.5% CPU.

**Reshaped after review** (@njbrake, @jerome-benoit). The first attempt made `has_live_tmux_pane()` itself cache-backed. That was wrong: it trusted a cached *miss*, which this module's own doctrine on `session_exists()` rejects ("a HIT proves the session existed as of the last scan, but a MISS is unreliable"). The decision drives both peer exclusion and tmux env publication, so a false negative and a false positive are each harmful, and one consumer (`instance.rs:5173`) is a one-shot cleanup that does not self-heal. My safety argument — "both creation paths already refresh the cache" — also only held in-process: the TUI, the daemon and each CLI invocation carry separate caches.

The current shape takes **one fresh, fallible observation per pass** and threads it through the per-item lookups, as @njbrake proposed. Four passes share one snapshot: the peer-exclusion walk, the two `HomeView::new` walks, poller repair, and the daemon's equivalent behind `serve`. The per-item probe underneath (`tmux_env_session_name`) is untouched and still live at the two correctness-sensitive sites @jerome-benoit named, the post-CAS env publication in `sync.rs` and the one-shot holder cleanup in `instance.rs`. "Server unreachable" stays distinguishable on the snapshot (`names() == None`) instead of collapsing into "absent": the exclusion walk maps it to "not live", which is exactly what the per-item probe did when its own `list-sessions` failed, and that walk re-runs.

**Third round** (@njbrake at `10ea6a0a`, @jerome-benoit at `10ea6a0a`), three items, one commit:

- **Unknown must not read as Absent in a one-shot pass.** `HomeView::new`'s hidden-env publication is not retried: no reload re-runs it, and a poller does not re-emit an unchanged sid. A single transient `list-sessions` failure at startup therefore skipped every row and left `AOE_INSTANCE_ID` / `AOE_CAPTURED_SESSION_ID` missing or stale until an unrelated sid change or a relaunch — which is exactly what peer exclusion reads. That caller now falls back to a fresh per-item probe when the snapshot could not reach the server. Startup recovery in the same file already draws this distinction the other way (skip the whole pass rather than read it as "all panes dead"), so the rule is not new to the file, only to this pass.
- **The snapshot is lazy, which closes the "adds forks to passes it does not fix" finding.** `LiveSessionSnapshot` now observes on first use rather than at construction, and observes `list-sessions` and `list-panes -a` independently. `build_exclusion_set` also reads the pass's snapshot instead of forking its own `list-sessions`, so the duplicate call @njbrake and CodeRabbit both flagged is gone as well. Measured with a `tmux` argv shim, against `10ea6a0a`:

  | | before | after |
  |---|---|---|
  | snapshot taken but never queried | 2 forks | **0** |
  | pass with no live same-tool peer | 3 forks | **1** |

  The one remaining fork is `build_exclusion_set`'s own `list-sessions`, which is what `main` pays for that pass too. So no pass is now more expensive than it was before this PR, which was the substance of the finding.
- **`has_live_tmux_pane()` is deleted.** Once every production pass takes a snapshot it had no callers left, surviving only in one test assert and in two `recovery.rs` doc comments. It is `pub fn`, so nothing warned. Its rationale moved to `has_live_tmux_pane_in`; the doc comments are reworded.

**On the numbers.** The reshape initially did not work, and only re-measuring caught it. Batching `compose_exclusion_with_persisted_peers` alone left the fork rate unchanged, because that is not the hot caller — I had inferred it from reading the code. Profiling an unstripped build named the real one:

```
App::run
  -> HomeView::repair_session_id_pollers          4389 of ~4700 main-thread samples
    -> Instance::repair_session_id_poller_if_needed
      -> has_live_tmux_pane -> tmux_env_session_name -> Command::output
```

Poller repair walks every instance on the `App::run` tick, on the thread that also serves keystrokes. @njbrake reproduced this independently and sized it larger still: 338 forks per `HomeView::new` and 169 per repair pass on a 169-instance store, against 2 each after.

End-to-end numbers, measured on the same machine and store (182 sessions, 9 live), 8-second windows, `tmux` shimmed to log argv with per-pid attribution so two TUIs cannot contaminate each other. **These were taken at `10ea6a0a`**, before the laziness commit, which only reduces the counts further; I have not re-run the 8-second windows since, and the per-pass shim numbers above are what covers that commit:

| | `list-sessions` / 8s | total tmux calls / 8s | idle TUI CPU |
|---|---|---|---|
| before | ~693 | ~760 | 9.5% |
| at `10ea6a0a` | **32** | **209** | **3.7%** |

For reference, the rejected cache-the-accessor version measured 16 and 182 on the same store. The batched version makes more calls than the cache and matches it on CPU — that is the intended trade: a fresh observation per pass instead of a possibly-`CACHE_TTL`-stale one.

One caveat on the totals: `capture-pane` volume depends on whether the previewed pane is streaming, so only the `list-sessions` row is a clean like-for-like. Separately, and unrelated to this PR: previewing a session that is *not* in the session cache forks a `has-session` per rendered frame (~22/s) via `refresh_preview_cache_if_needed` -> `Session::exists`. That is pre-existing on `main`; happy to file it separately if useful.

Also addressed from review: `cached_session_names()` is gone, so the rustdoc failure @jerome-benoit found (`public documentation ... links to private item`) goes with it — I verified `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve` is clean. The per-call `Vec<String>` clone he and @njbrake both flagged is gone too. The test helper I had "fixed" is restored to its original form: @njbrake is right that its failing assertion *was* the out-of-band-creation coverage, so refreshing inside the helper deleted the signal rather than fixing it. The four em dashes in added comments are gone; I left the ones in surrounding pre-existing prose alone.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (4,666 library tests pass; one unrelated pre-existing failure, see below)
- [x] Documentation was updated where necessary (rationale documented on the new accessor and at the call sites)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped an e2e test, because: this is a subprocess-count change behind an existing accessor with no behavioural difference to assert end-to-end. The observable effect is a fork rate, measured above with a `tmux` argv shim; the correctness-relevant edges (fresh snapshot, unreachable server, one-shot fallback) are covered by unit tests.

Added:

- `tmux::tests::snapshot_lookup_matches_the_per_item_probe` — a live name resolves from the snapshot, and a dead pane does not.
- `tmux::tests::snapshot_lookup_reports_not_live_when_server_unreachable` — pins the "server unreachable" behaviour for the re-running walks to what the per-caller `list-sessions` did on failure.
- `instance::tests::one_shot_name_probes_when_the_snapshot_missed_tmux` — the regression @jerome-benoit asked for: with a shimmed `tmux` on `PATH`, an unreachable snapshot must still resolve the row through a per-item probe, while a snapshot that did reach the server stays authoritative. I checked it fails with the fallback removed, so it is pinning the behaviour rather than passing vacuously.

Commands run on this branch (based on `9e970918`):

- `cargo clippy -- -D warnings` (the CI form)
- `cargo check --features serve` (a released binary is built with it; CI's plain `cargo clippy` does not cover the `server/mod.rs` call site)
- `cargo test --lib` — 4,666 passed, 1 failed (below), 2 ignored
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve` — clean

Two notes on running these locally, neither caused by this PR:

- `rust-toolchain.toml` pins `channel = "stable"`, and since #3499 the dev-dependency `serial_test 4.0.1` requires rustc 1.93.1, so any `stable` older than that cannot build the test targets at all. I ran the above on 1.95.0.
- `cargo clippy --features serve -- -D warnings` on 1.95.0 reports two `collapsible_match` errors, in `src/acp/event_store.rs:1275` and `src/process/runner.rs:1779`. Neither file is in this diff; this is a newer clippy than the one CI currently runs.

`git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` fails on this machine, and fails identically on an unmodified checkout of the same commit — it compares `/var/...` against macOS's `/private/var/...`. Unrelated to this change.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Claude profiled the running TUI (`sample`), identified the fork storm by shimming `tmux` to log argv, traced it to the accessor and its per-instance callers, wrote the patch, and re-ran the same measurement to confirm the drop. The before/after numbers above are from that instrumentation, not estimates; where a number predates the current head, the body says so rather than restating it as current.

## Summary

- Added one lazy, fresh `tmux` snapshot for each liveness scan.
- Reused the snapshot across session checks, exclusions, TUI updates, and poller repairs.
- Added fallback probing when `tmux` is temporarily unreachable.
- Removed the obsolete liveness accessor.
- Updated documentation and tests for snapshot freshness, unreachable servers, and pane visibility.

## Benefits

- Reduces repeated `tmux` process calls and idle CPU usage.
- Prevents cached misses from hiding live sessions.
- Preserves correct behavior when `tmux` is unreachable.
- Keeps session and poller state more accurate.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -564,6 +564,110 @@ pub fn agent_session_belongs_to(tmux_name: &str, session_id: &str) -> bool {
     NameShape::agent(&id_suffix(session_id)).matches(tmux_name)
 }
 
+/// One tmux observation shared by a batch of per-instance liveness lookups.
+///
+/// A pass that asks "is this instance's pane live?" once per stored session
+/// otherwise pays a `list-sessions` fork per instance, plus a `pane_dead`
+/// fork per match. `compose_exclusion_with_persisted_peers` walks every
+/// session sharing the project path — trashed ones included — so on a store
+/// of a few hundred that is a few hundred `fork`+`exec` round-trips per pass,
+/// on the thread that also serves input.
+///
+/// This is a *fresh* observation taken once per pass, not a cached one: the
+/// session cache's answers are asymmetric (a hit proves existence, a miss only
+/// means "not seen at the last scan"), and liveness here decides both
+/// exclusion and env publication, where a false negative and a false positive
+/// are each harmful. Taking one live snapshot keeps the decision exactly as
+/// authoritative as the per-item probe it replaces.
+///
+/// Server-unreachable is preserved rather than collapsed into "absent":
+/// [`Self::names`] returns `None`, so a caller that needs Unknown distinct
+/// from Absent can still tell them apart.
+pub(crate) struct LiveSessionSnapshot {
+    names: Option<Vec<String>>,
+    panes: Option<HashMap<String, PaneMetadata>>,
+}
+
+impl LiveSessionSnapshot {
+    /// Take one fresh snapshot: a single `list-sessions` plus a single
+    /// `list-panes -a`, regardless of how many instances are then looked up.
+    pub(crate) fn take() -> Self {
+        let names = tmux_query_command()
+            .args(["list-sessions", "-F", "#{session_name}"])
+            .output()
+            .ok()
+            .filter(|out| out.status.success())
+            .map(|out| {
+                String::from_utf8_lossy(&out.stdout)
+                    .lines()
+                    .map(str::trim)
+                    .filter(|l| !l.is_empty())
+                    .map(String::from)
+                    .collect()
+            });
+        Self {
+            names,
+            panes: batch_pane_metadata().ok(),
+        }
+    }
+
+    /// Build a snapshot from already-known parts, for tests that must not
+    /// depend on a live tmux server.
+    #[cfg(test)]
+    pub(crate) fn from_parts(
+        names: Option<Vec<String>>,
+        panes: Option<HashMap<String, PaneMetadata>>,
+    ) -> Self {
+        Self { names, panes }
+    }
+
+    /// Live session names, or `None` when the tmux server could not be reached.
+    pub(crate) fn names(&self) -> Option<&[String]> {
+        self.names.as_deref()
+    }
+
+    /// Whether `name`'s first pane is dead, mirroring `utils::is_pane_dead`
+    /// but read from the batched metadata. An absent entry is reported alive,
+    /// matching the per-item probe's `unwrap_or(false)` on a failed query.
+    pub(crate) fn pane_dead(&self, name: &str) -> bool {
+        self.panes
+            .as_ref()
+            .and_then(|panes| panes.get(name))
+            .map(|meta| meta.pane_dead)
+            .unwrap_or(false)
+    }
+}
+
+/// [`live_any_kind_name_for_id`] against an already-taken snapshot, so a batch
+/// of lookups costs one observation instead of one per instance.
+pub(crate) fn live_any_kind_name_for_id_in(
+    snapshot: &LiveSessionSnapshot,
+    session_id: &str,
+) -> Option<String> {
+    let names = snapshot.names()?;
+    let suffix = id_suffix(session_id);
+    let agent = NameShape::agent(&suffix);
+    let terminal = NameShape::terminal(&suffix);
+    let container = NameShape::container(&suffix);
+    let (mut agent_hit, mut terminal_hit, mut container_hit) = (None, None, None);
+    for name in names {
+        let name = name.as_str();
+        let bucket = if agent.matches(name) {
+            &mut agent_hit
+        } else if terminal.matches(name) {
+            &mut terminal_hit
+        } else if container.matches(name) {
+            &mut container_hit
+        } else {
+            continue;
+        };
+        if bucket.is_none() && !snapshot.pane_dead(name) {
+            *bucket = Some(name.to_string());
+        }
+    }
+    agent_hit.or(terminal_hit).or(container_hit)
+}
+
 /// The live tmux session name carrying `session_id`'s `_<id8>` tail, preferring
 /// the agent pane, then a paired terminal, then a container terminal, skipping
 /// dead panes (tool sub-sessions never match any of these shapes). Unlike
@@ -997,47 +1101,6 @@ impl Drop for SessionCacheGuard {
 /// force a fresh `refresh_session_cache()` call.
 const CACHE_TTL: Duration = Duration::from_secs(2);
 
-/// Live tmux session names from the shared [`SESSION_CACHE`], refreshing the
-/// snapshot only when it is older than [`CACHE_TTL`].
-///
-/// Any caller that needs the name list *per instance* must use this rather than
-/// its own `list-sessions`. A store with N sessions turns one such caller into N
-/// forks per pass, and a fork+exec round-trip is ~5ms on macOS, so a few hundred
-/// sessions is enough to keep a thread blocked for a large fraction of every
-/// second — including the thread serving keystrokes.
-///
-/// An empty list is returned both for "no sessions" and for "the last refresh
-/// could not reach the server". That matches what the per-caller
-/// `list-sessions` did on failure, and deliberately does not re-probe: during a
-/// real outage the retry fails the same way and just burns a fork per lookup.
-pub fn cached_session_names() -> Vec<String> {
-    if let Some(names) = session_names_from_cache() {
-        return names;
-    }
-    refresh_session_cache();
-    session_names_from_cache().unwrap_or_default()
-}
-
-/// Names from the current snapshot, or `None` when it is stale (older than
-/// [`CACHE_TTL`]) or the lock is poisoned, meaning the caller must refresh.
-fn session_names_from_cache() -> Option<Vec<String>> {
-    let cache = SESSION_CACHE.read().ok()?;
-    let fresh = cache
-        .time
-        .map(|t| t.elapsed() <= CACHE_TTL)
-        .unwrap_or(false);
-    if !fresh {
-        return None;
-    }
-    Some(
-        cache
-            .data
-            .as_ref()
-            .map(|m| m.keys().cloned().collect())
-            .unwrap_or_default(),
-    )
-}
-
 pub fn session_exists_from_cache(name: &str) -> Option<bool> {
     let cache = SESSION_CACHE.read().ok()?;
 
@@ -1453,32 +1516,6 @@ mod tests {
 
     #[test]
     #[serial_test::serial]
-    fn cached_session_names_serves_a_fresh_snapshot() {
-        let guard = SessionCacheGuard::capture();
-        let a = format!("{P}cached_names_one_abc12345");
-        let b = format!("{P}cached_names_two_abc12345");
-        guard.force_present(&[&a, &b]);
-        let mut got = cached_session_names();
-        got.sort();
-        let mut want = vec![a, b];
-        want.sort();
-        assert_eq!(got, want);
-    }
-
-    #[test]
-    #[serial_test::serial]
-    fn cached_session_names_is_empty_when_server_unreachable() {
-        let guard = SessionCacheGuard::capture();
-        // Fresh snapshot, no data: the last `list-sessions` failed. Callers
-        // treated that as "no live sessions" when they forked it themselves, so
-        // it stays that way — and must not re-probe, which during an outage
-        // would burn one fork per lookup.
-        guard.force_unreachable();
-        assert!(cached_session_names().is_empty());
-    }
-
-    #[test]
-    #[serial_test::serial]
     fn probe_session_existence_returns_unknown_when_server_unreachable() {
         let guard = SessionCacheGuard::capture();
         let name = format!("{P}probe_unknown_abc12345");
@@ -1676,6 +1713,47 @@ mod tests {
             ID
         ));
         assert!(!agent_session_belongs_to("vim", ID));
+    }
+
+    fn dead_pane_meta(dead: bool) -> PaneMetadata {
+        PaneMetadata {
+            pane_dead: dead,
+            pane_current_command: None,
+            pane_start_command_is_protected: false,
+        }
+    }
+
+    #[test]
+    fn snapshot_lookup_matches_the_per_item_probe() {
+        let agent = format!("{P}Refactor_{ID8}");
+        let snapshot = LiveSessionSnapshot::from_parts(
+            Some(vec![agent.clone()]),
+            Some(HashMap::from([(agent.clone(), dead_pane_meta(false))])),
+        );
+        assert_eq!(
+            live_any_kind_name_for_id_in(&snapshot, ID).as_deref(),
+            Some(agent.as_str())
+        );
+    }
+
+    #[test]
+    fn snapshot_lookup_skips_a_dead_pane() {
+        let agent = format!("{P}Refactor_{ID8}");
+        let snapshot = LiveSessionSnapshot::from_parts(
+            Some(vec![agent.clone()]),
+            Some(HashMap::from([(agent.clone(), dead_pane_meta(true))])),
+        );
+        assert_eq!(live_any_kind_name_for_id_in(&snapshot, ID), None);
+    }
+
+    #[test]
+    fn snapshot_lookup_reports_not_live_when_server_unreachable() {
+        // `names() == None` is Unknown, preserved on the snapshot so a caller
+        // can distinguish it; the exclusion walk collapses it to "not live",
+        // which is what the per-item probe did when its `list-sessions` failed.
+        let snapshot = LiveSessionSnapshot::from_parts(None, None);
+        assert!(snapshot.names().is_none());
+        assert_eq!(live_any_kind_name_for_id_in(&snapshot, ID), None);
     }
 
     #[test]
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -569,46 +569,37 @@ pub fn agent_session_belongs_to(tmux_name: &str, session_id: &str) -> bool {
 /// A pass that asks "is this instance's pane live?" once per stored session
 /// otherwise pays a `list-sessions` fork per instance, plus a `pane_dead`
 /// fork per match. `compose_exclusion_with_persisted_peers` walks every
-/// session sharing the project path — trashed ones included — so on a store
-/// of a few hundred that is a few hundred `fork`+`exec` round-trips per pass,
-/// on the thread that also serves input.
+/// session sharing the project path, trashed ones included, so on a store of
+/// a few hundred that is a few hundred `fork`+`exec` round-trips per pass, on
+/// the thread that also serves input.
 ///
-/// This is a *fresh* observation taken once per pass, not a cached one: the
-/// session cache's answers are asymmetric (a hit proves existence, a miss only
-/// means "not seen at the last scan"), and liveness here decides both
-/// exclusion and env publication, where a false negative and a false positive
-/// are each harmful. Taking one live snapshot keeps the decision exactly as
-/// authoritative as the per-item probe it replaces.
+/// The observations are *fresh*, not cached: the session cache's answers are
+/// asymmetric (a hit proves existence, a miss only means "not seen at the last
+/// scan"), and liveness here decides both peer exclusion and env publication,
+/// where a false negative and a false positive are each harmful. One live
+/// observation per pass leaves the decision exactly as authoritative as the
+/// per-item probe it replaces.
 ///
-/// Server-unreachable is preserved rather than collapsed into "absent":
-/// [`Self::names`] returns `None`, so a caller that needs Unknown distinct
-/// from Absent can still tell them apart.
+/// Each observation is taken on first use, and only if used: a pass that ends
+/// up asking nothing, because no stored peer shares the project path or every
+/// row short-circuits before the liveness clause, forks nothing, and a caller
+/// that only needs session names never forks `list-panes`.
+///
+/// An unreachable server is preserved rather than collapsed into "absent":
+/// [`Self::names`] returns `None`, so a one-shot caller that cannot retry can
+/// tell Unknown from Absent and probe per row instead (see
+/// `Instance::tmux_env_session_name_in_or_probe`).
+#[derive(Default)]
 pub(crate) struct LiveSessionSnapshot {
-    names: Option<Vec<String>>,
-    panes: Option<HashMap<String, PaneMetadata>>,
+    names: OnceLock<Option<Vec<String>>>,
+    panes: OnceLock<Option<HashMap<String, PaneMetadata>>>,
 }
 
 impl LiveSessionSnapshot {
-    /// Take one fresh snapshot: a single `list-sessions` plus a single
-    /// `list-panes -a`, regardless of how many instances are then looked up.
-    pub(crate) fn take() -> Self {
-        let names = tmux_query_command()
-            .args(["list-sessions", "-F", "#{session_name}"])
-            .output()
-            .ok()
-            .filter(|out| out.status.success())
-            .map(|out| {
-                String::from_utf8_lossy(&out.stdout)
-                    .lines()
-                    .map(str::trim)
-                    .filter(|l| !l.is_empty())
-                    .map(String::from)
-                    .collect()
-            });
-        Self {
-            names,
-            panes: batch_pane_metadata().ok(),
-        }
+    /// A snapshot for one pass: at most one `list-sessions` and at most one
+    /// `list-panes -a`, however many instances are then looked up.
+    pub(crate) fn new() -> Self {
+        Self::default()
     }
 
     /// Build a snapshot from already-known parts, for tests that must not
@@ -618,19 +609,42 @@ impl LiveSessionSnapshot {
         names: Option<Vec<String>>,
         panes: Option<HashMap<String, PaneMetadata>>,
     ) -> Self {
-        Self { names, panes }
+        let snapshot = Self::new();
+        let _ = snapshot.names.set(names);
+        let _ = snapshot.panes.set(panes);
+        snapshot
     }
 
     /// Live session names, or `None` when the tmux server could not be reached.
     pub(crate) fn names(&self) -> Option<&[String]> {
-        self.names.as_deref()
+        self.names
+            .get_or_init(|| {
+                tmux_query_command()
+                    .args(["list-sessions", "-F", "#{session_name}"])
+                    .output()
+                    .ok()
+                    .filter(|out| out.status.success())
+                    .map(|out| {
+                        String::from_utf8_lossy(&out.stdout)
+                            .lines()
+                            .map(str::trim)
+                            .filter(|l| !l.is_empty())
+                            .map(String::from)
+                            .collect()
+                    })
+            })
+            .as_deref()
     }
 
     /// Whether `name`'s first pane is dead, mirroring `utils::is_pane_dead`
     /// but read from the batched metadata. An absent entry is reported alive,
     /// matching the per-item probe's `unwrap_or(false)` on a failed query.
+    ///
+    /// Only reachable once [`Self::names`] has produced a candidate, so an
+    /// unreachable server never pays for this observation.
     pub(crate) fn pane_dead(&self, name: &str) -> bool {
         self.panes
+            .get_or_init(|| batch_pane_metadata().ok())
             .as_ref()
             .and_then(|panes| panes.get(name))
             .map(|meta| meta.pane_dead)
@@ -1726,33 +1740,28 @@ mod tests {
     #[test]
     fn snapshot_lookup_matches_the_per_item_probe() {
         let agent = format!("{P}Refactor_{ID8}");
-        let snapshot = LiveSessionSnapshot::from_parts(
-            Some(vec![agent.clone()]),
-            Some(HashMap::from([(agent.clone(), dead_pane_meta(false))])),
-        );
-        assert_eq!(
-            live_any_kind_name_for_id_in(&snapshot, ID).as_deref(),
-            Some(agent.as_str())
-        );
-    }
-
-    #[test]
-    fn snapshot_lookup_skips_a_dead_pane() {
-        let agent = format!("{P}Refactor_{ID8}");
-        let snapshot = LiveSessionSnapshot::from_parts(
-            Some(vec![agent.clone()]),
-            Some(HashMap::from([(agent.clone(), dead_pane_meta(true))])),
-        );
-        assert_eq!(live_any_kind_name_for_id_in(&snapshot, ID), None);
+        let cases = [(false, Some(agent.as_str())), (true, None)];
+        for (pane_dead, expected) in cases {
+            let snapshot = LiveSessionSnapshot::from_parts(
+                Some(vec![agent.clone()]),
+                Some(HashMap::from([(agent.clone(), dead_pane_meta(pane_dead))])),
+            );
+            assert_eq!(
+                live_any_kind_name_for_id_in(&snapshot, ID).as_deref(),
+                expected,
+                "pane_dead = {pane_dead}"
+            );
+        }
     }
 
     #[test]
     fn snapshot_lookup_reports_not_live_when_server_unreachable() {
-        // `names() == None` is Unknown, preserved on the snapshot so a caller
-        // can distinguish it; the exclusion walk collapses it to "not live",
-        // which is what the per-item probe did when its `list-sessions` failed.
+        // Unknown collapses to "not live" for the exclusion walk, which is what
+        // the per-item probe did when its own `list-sessions` failed, and the
+        // walk re-runs. A one-shot caller must not collapse it; that rule is
+        // covered by
+        // `instance::tests::one_shot_name_probes_when_the_snapshot_missed_tmux`.
         let snapshot = LiveSessionSnapshot::from_parts(None, None);
-        assert!(snapshot.names().is_none());
         assert_eq!(live_any_kind_name_for_id_in(&snapshot, ID), None);
     }
 
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -997,6 +997,47 @@ impl Drop for SessionCacheGuard {
 /// force a fresh `refresh_session_cache()` call.
 const CACHE_TTL: Duration = Duration::from_secs(2);
 
+/// Live tmux session names from the shared [`SESSION_CACHE`], refreshing the
+/// snapshot only when it is older than [`CACHE_TTL`].
+///
+/// Any caller that needs the name list *per instance* must use this rather than
+/// its own `list-sessions`. A store with N sessions turns one such caller into N
+/// forks per pass, and a fork+exec round-trip is ~5ms on macOS, so a few hundred
+/// sessions is enough to keep a thread blocked for a large fraction of every
+/// second — including the thread serving keystrokes.
+///
+/// An empty list is returned both for "no sessions" and for "the last refresh
+/// could not reach the server". That matches what the per-caller
+/// `list-sessions` did on failure, and deliberately does not re-probe: during a
+/// real outage the retry fails the same way and just burns a fork per lookup.
+pub fn cached_session_names() -> Vec<String> {
+    if let Some(names) = session_names_from_cache() {
+        return names;
+    }
+    refresh_session_cache();
+    session_names_from_cache().unwrap_or_default()
+}
+
+/// Names from the current snapshot, or `None` when it is stale (older than
+/// [`CACHE_TTL`]) or the lock is poisoned, meaning the caller must refresh.
+fn session_names_from_cache() -> Option<Vec<String>> {
+    let cache = SESSION_CACHE.read().ok()?;
+    let fresh = cache
+        .time
+        .map(|t| t.elapsed() <= CACHE_TTL)
+        .unwrap_or(false);
+    if !fresh {
+        return None;
+    }
+    Some(
+        cache
+            .data
+            .as_ref()
+            .map(|m| m.keys().cloned().collect())
+            .unwrap_or_default(),
+    )
+}
+
 pub fn session_exists_from_cache(name: &str) -> Option<bool> {
     let cache = SESSION_CACHE.read().ok()?;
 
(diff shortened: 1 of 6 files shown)


njbrake (commented):
The problem is real and the measurement is the strongest part of this PR. Please reshape the fix: batch the snapshot per pass instead of making the per-item accessor cache-backed.

Your own module states the doctrine this crosses. `session_exists` says "a HIT proves the session existed as of the last scan (trust it), but a MISS is unreliable ... Trusting a cached miss is what made teardown and drift decisions racy." `cached_session_names()` plus `live_any_kind_name_for_id` trusts a cached miss with no live fallback, and trusts a cached hit for a decision where a false positive is destructive, so it takes on both risk directions where `probe_session_existence` deliberately splits them into Present, Absent and Unknown.

Both directions land on the same invariant. A false "live" (peer killed under 2s ago) makes `should_exclude` false in `compose_exclusion_with_persisted_peers`, so a dead peer's `agent_session_id` stays available for the current instance to capture, which is the #2355 steal class. A false "not live" makes `tui/home/mod.rs:2468` and `session/sync.rs:583` skip publishing `AOE_INSTANCE_ID` and `AOE_CAPTURED_SESSION_ID` into tmux env, which is how other instances attribute sids. Both self-heal on the next pass except `instance.rs:5173`, a one-shot cleanup whose comment explains why skipping it reopens the exclusion bug. Note too that "both creation paths already refresh" holds only in-process: the TUI, the daemon and each CLI invocation carry separate caches, so a session created by one is invisible to another for up to the TTL.

The batching pattern is already here: `recovery::orphaned_agents_alive` ("N candidates cost one /proc walk") and `tmux::batch_pane_metadata`. Hoisting one refresh into `compose_exclusion_with_persisted_peers` and threading the snapshot down gets the same 169 forks down to 1, with no change to what `has_live_tmux_pane()` means at its other five call sites, several of which are not in the hot loop.

Two smaller notes. `cached_session_names()` clones the list per call, so the optimized pass trades 169 forks for roughly 169 squared string clones; doing the lookup under the read lock avoids that. And the test helper you adjusted was real signal: its failing assertion was the out-of-band creation case, so refreshing inside the helper removes the only coverage of it.

Worth coordinating with #3507, which adds 35 lines inside `compose_exclusion` hardening the invariant this loosens for a TTL window. They merge clean, so nothing will flag the interaction.

I approved the pending workflow runs, so the Rust legs are executing now.

src/tmux/mod.rs
```diff
@@ -997,6 +997,47 @@ impl Drop for SessionCacheGuard {
 /// force a fresh `refresh_session_cache()` call.
 const CACHE_TTL: Duration = Duration::from_secs(2);
 
+/// Live tmux session names from the shared [`SESSION_CACHE`], refreshing the
+/// snapshot only when it is older than [`CACHE_TTL`].
```
jerome-benoit: **[P1] Restore the required rustdoc build**

The required docs command fails here because this public documentation links to the private `SESSION_CACHE` and `CACHE_TTL` items. Live CI runs `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve` and reports `public documentation for cached_session_names links to private item` for both links. All current call sites are internal (`src/session/instance.rs` plus tests), so please make `cached_session_names` `pub(crate)`; alternatively, render these names as plain code spans instead of intra-doc links.

jerome-benoit (dismissed):
I agree with @njbrake’s blocking liveness finding and will not repeat it here. I independently reproduced the same stale-name failure mode and traced one additional consequence: a post-CAS tmux environment publication that targets a stale name is not guaranteed to heal on the next poll. The observation is acknowledged before `publish_tmux_env` (`src/session/sync.rs:325`, publication at `:375`), the sequential fallback logs and swallows target failures (`src/tmux/env.rs:241-269`), and the poller emits only changed observations (`src/session/poller.rs:312-319`).

Please keep liveness batch-scoped: take one fresh, fallible tmux snapshot per pass and pass it through the per-item lookups, preserving `Present` / `Absent` / `Unknown`. Add regression coverage for out-of-process create/kill between passes and for post-CAS environment publication, proving that the current live name receives `AOE_INSTANCE_ID` / `AOE_CAPTURED_SESSION_ID` or that a failed target is explicitly retried.

The durable flock guard prevents this evidence from proving durable SID duplication, so I classify the functional defect as P2. It still requires a change before merge.

njbrake (commented):

_Reviewed at `b4155849`. Push new commits and it will be re-reviewed._

---

## What changed since the prescription

The reshape landed. Commit `b415584` replaces the `SESSION_CACHE`-backed accessor with `LiveSessionSnapshot` (`src/tmux/mod.rs:586-641`), a fresh per-pass observation (`list-sessions` + `list-panes -a`), hoisted once into `compose_exclusion_with_persisted_peers` (`src/session/capture.rs:1011`) and threaded down through a new `has_live_tmux_pane_in(&snapshot)` (`src/session/instance.rs:3738`). This is the batching that was asked for, built on the pattern already in the file (`resolve_agent_session_name_in` against a `batch_pane_metadata()` map, `src/tmux/mod.rs:753`), not a parallel mechanism.

Every objection in the earlier review is answered by construction. I verified each rather than taking the diff's word for it:

- **Cached-miss doctrine.** Gone. The diff to `src/tmux/mod.rs` is purely additive (+145/-0) and touches nothing in `SESSION_CACHE`; the new path never consults it. The asymmetry `session_exists` warns about (`src/tmux/mod.rs:1191-1198`) is no longer reachable from liveness.
- **The one-shot cleanup.** `tmux_env_session_name_for_instance_id` (`src/session/instance.rs:1515`) is byte-identical to `main`, and the `cleared_holder_ids` loop that calls it (`src/session/instance.rs:5184` on this head) is untouched. That site still does its own live probe.
- **Cross-process staleness.** No cache, so no TTL window. `src/tui/home/mod.rs:2512`, `src/session/sync.rs:583` and `src/session/instance.rs:5883` all still use the live per-item probe; the meaning of `has_live_tmux_pane()` is unchanged at every site the PR did not touch.
- **`cached_session_names()` clone cost.** The function does not exist on this head.
- **Removed test coverage.** Restored. The `src/session/instance.rs` diff is +11/-0, all of it the new accessor; the test helper is back to what `main` has.
- **The P1 rustdoc failure** (public docs linking private `SESSION_CACHE`/`CACHE_TTL`) is moot with `cached_session_names` gone. `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps` on the rebased head gives one error, the pre-existing `try_smart_rename` intra-doc link at `src/session/smart_rename.rs:243`, byte-identical on `origin/main` and resolved under `--features serve`, which is what CI runs.

I also checked that `live_any_kind_name_for_id_in` is semantically identical to the probe it replaces, since that is the load-bearing fact. All three name shapes (`aoe_`, `aoe_term_`, `aoe_cterm_`) begin with `SESSION_PREFIX`, so `parse_pane_metadata`'s prefix filter (`src/tmux/mod.rs:1001`) drops no candidate; it keys on the first window's pane 0, which is what `utils::is_pane_dead`'s `:^.0` target resolves to; an absent map entry reporting "alive" matches `is_pane_dead`'s `unwrap_or(false)`; and `names() == None` collapsing to "not live" matches the old `.ok()?` / `!status.success()` early return. Four for four.

**I re-verified the perf claim rather than trusting the body.** Scratch unit test in a worktree: a `tmux` shim on `PATH` logging argv, a temp `$HOME` store seeded with 169 Claude instances on one project path, one call to `compose_exclusion_with_persisted_peers`.

```
pre-PR behavior (two production hunks reverted): 170 invocations, 170 list-sessions
head b4155849:                                     3 invocations,   2 list-sessions + 1 list-panes -a
```

Identical 169-entry exclusion set both ways. The mechanism holds.

## Blocking

**1. The PR body and title describe the implementation you replaced.** This is the biggest thing to fix, because the description becomes the squashed commit message. The title says "serve session liveness from the cache"; the Description says "`SESSION_CACHE` already exists for exactly this data… The fix routes the lookup through that cache"; the "On staleness" paragraph argues a freshness contract this head does not use; the two tests it names (`cached_session_names_serves_a_fresh_snapshot`, `cached_session_names_is_empty_when_server_unreachable`) do not exist on `b4155849`; and the "One test helper in `session::instance::tests`" paragraph describes a change not in the diff. The after-measurement is the sharpest instance: "100 tmux invocations in 6s, of which **0 uncached list-sessions**" is structurally impossible on this head, which forks `list-sessions` twice per pass by design. Those numbers were taken on `f47ac5cc`. Please rewrite the description around `LiveSessionSnapshot` and re-measure, or state plainly that the after-numbers are from the superseded commit.

**2. `LiveSessionSnapshot::take()` is unconditional, so this fork-reduction PR adds forks to the paths it does not fix.** `src/session/capture.rs:1011` takes the snapshot before any branch that decides whether liveness is ever consulted. `include_inactive_same_tool` is `tool == "claude" || (tool == "codex" && !is_sandboxed())` (`src/session/instance.rs:3131`), and the loop also `continue`s on `inst.tool != current_tool`. For a gemini, opencode, omp, vibe or cursor session, and for any pass where no stored peer shares the project path, `has_live_tmux_pane_in` is never called, yet the pass now pays two forks it previously did not. It fires even when the store is empty. Make it lazy: a `OnceCell<LiveSessionSnapshot>` populated on first use, or hoist the `take()` inside a guard on `include_inactive_same_tool`. One line either way, and it removes the only case where this PR makes something slower.

## Nits

**3. Duplicate `list-sessions` in the same pass.** `compose_exclusion` → `build_exclusion_set` already forks `list-sessions` at `src/session/capture.rs:1060`, then `take()` forks it again; my shim run counted 2 per pass. `build_exclusion_set` is otherwise already batched (`get_hidden_env_batch`), so threading one name list through gets the pass to two forks total. Not blocking on its own, but it pairs naturally with fix 2 since both are edits to the same three lines.

**4. `take()` runs `batch_pane_metadata()` even when `names` is `None`** (`src/tmux/mod.rs:594-611`). `live_any_kind_name_for_id_in` returns on `names()?` before ever consulting `panes`, so during a tmux outage that is a guaranteed-wasted fork at the moment forks are least likely to be cheap.

**5. Test shape.** `snapshot_lookup_matches_the_per_item_probe` (`src/tmux/mod.rs:1727`) and `snapshot_lookup_skips_a_dead_pane` (`:1740`) are the same setup with one bool flipped, which is AGENTS.md's table case: `let cases = [(false, Some(agent)), (true, None)]`. The unreachable-server test has different setup and should stay separate, though `assert!(snapshot.names().is_none())` at `:1755` just restates `from_parts(None, None)`. Worth saying: these three are pure and need no tmux and no `#[serial]`, which is a genuine improvement over the `#[serial_test::serial]` live-tmux test right below them.

**6. Two em dashes in added doc comments**, `src/tmux/mod.rs:572` and `src/session/instance.rs:3736`. AGENTS.md forbids them in prose. The surrounding files already violate this heavily so it is low priority, but the rule is the rule.

**7. `names()` is documented for a caller that does not exist.** The doc block promises "a caller that needs Unknown distinct from Absent can still tell them apart" (`src/tmux/mod.rs:583-585`), but the only consumer, `live_any_kind_name_for_id_in`, collapses it immediately. Fine as written since the accessor is genuinely used; just do not let it grow speculative surface.

## On the remaining review request

The ask was regression coverage of "out-of-process create/kill between passes" and of post-CAS environment publication. The reshape moots most of that: the post-CAS publication path (`src/session/sync.rs:325`/`:375`) and the one-shot holder cleanup are untouched and still probe live, so there is no new failure mode there to pin.

The one residual semantic change worth naming explicitly, so nobody has to re-derive it: within a single pass, item *i* is now judged against an observation taken at *t₀* rather than at *t_i*. Under the old code that per-item freshness was zero, but the pass itself spanned ~850ms at 169 peers, so its result set was already internally inconsistent across a far wider window. Under the new code the whole pass is ~3 forks, so max intra-pass staleness is single-digit milliseconds. Strictly better in aggregate, and three orders of magnitude tighter than the 2s TTL the first attempt would have introduced.

## Interaction with #3507

They merge clean with each other; I verified `git merge --no-commit --no-ff pull/3507/head` onto `b4155849` succeeds with no conflict. #3507 *does* conflict with current `main` on its own, in `resolve_resume_target` against #3413 (`e75abee`), but that is #3507's rebase to do and nothing to do with this PR. Semantically the coordination concern evaporates with the reshape, since there is no longer a TTL window for #3507's hardening to fight.

One thing to watch if both land: #3507 adds `claude_project_dir_is_shared`, which does its own `Storage::new_unwatched` + `load()` walk of the same store on a per-instance path. That is a fresh O(N) walk of exactly the thing this PR just stopped forking over. Different cost shape, cheaper than forks, but worth watching.

## Definition of done, run locally on the head rebased onto current `main`

- `cargo fmt --all -- --check`: clean.
- `cargo clippy -- -D warnings` (the CI form, `.github/workflows/ci.yml:40`): clean.
- `cargo clippy --all-targets -- -D warnings`: one failure, `items_after_test_module` at `src/process/macos.rs:266`, identical on `main`. Pre-existing.
- `cargo test --lib`: 4665 passed, 2 failed — `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` and `tui::dialogs::hooks_install::tests::test_content_shows_settings_path`. Both reproduce on an unmodified `origin/main` checkout; both are macOS-environmental. Note the second is *not* the one your body lists, so your local run and mine differ on which pre-existing failures surface; neither is related to this change.
- Rust-only diff, so no `--features serve` needed and patch coverage is N/A.

## CI

The Rust legs have **not** run on `b4155849`. The earlier approval covered `f47ac5cc` only, so runs `32838138944` (CI) and `32838139065` (Tests) are sitting at `action_required`. "No failing checks" on this PR currently means "no checks". My audit of the head is clean: three Rust source files, no workflow or CI config change, no install/build/test hook, no dependency or lockfile churn, no obfuscated or binary content, nothing unrelated smuggled in.

## Verdict

**NEEDS WORK**, on items 1 and 2 only.

The prescription was followed and the result holds up under scrutiny; that is worth saying plainly, especially given the patch is AI-generated and you flagged it as unreviewed. `LiveSessionSnapshot` extends the existing batched-observation idiom rather than adding a mechanism beside it, leaves the per-item probe intact for the five call sites that need it, and the 170 → 3 fork reduction reproduces independently. Merge risk is low: reverts cleanly, no persisted state, decision semantics provably identical to what they replace.

What is left is two mechanical fixes and a description that has to catch up to the code it describes. Fix the body, make the snapshot lazy, optionally fold in `build_exclusion_set`'s `list-sessions` while you are in those three lines, and this is ready.

njbrake (commented):

_Reviewed at `f69341f9`. Push new commits and it will be re-reviewed._

---

## Delta since `b4155849`

**The push is a pure rebase. No code changed.** I diffed the PR's own three-dot diff at both heads (`ec6b6d04...b4155849` vs `origin/main...f69341f9`): byte-identical apart from hunk offsets and blob hashes. The `capture.rs` / `instance.rs` movement is main arriving underneath — #3413's `claude_host_transcript_confirmed_absent` gate in `capture_claude_session_id`, #3509's `detect_as` re-resolve in the tool swap, #3465's doc rewording. None of it yours.

So everything established in the previous review stands, and I did not re-run the fork measurement: `compose_exclusion_with_persisted_peers` is untouched, so 170 → 3 still holds by construction.

## Blocking item 2 — still open

`let live = crate::tmux::LiveSessionSnapshot::take();` at `src/session/capture.rs:1031`; the `if !include_inactive_same_tool || inst.tool != current_tool` guard is 20 lines below at `:1051`, with no branch between them. Nothing else in the file makes it moot.

Every gemini / opencode / omp / vibe / cursor pass, and every pass over an empty or non-matching store, still pays two forks it did not pay before. A `OnceCell<LiveSessionSnapshot>` populated on first use, or hoisting the `take()` inside a guard on `include_inactive_same_tool`, closes it in one line.

## Blocking item 1 — mostly resolved, a few lines short

The Description is genuinely fixed, and better than what I asked for: it names the reshape, credits the review, and explicitly declines to quote the after-numbers because they were measured on the superseded commit rather than passing them off as current. That is the right call and worth saying.

What still describes the old implementation:

- **The title.** `perf(tmux): serve session liveness from the cache instead of forking list-sessions` becomes the squash subject and names a design that is not in the branch. Commit 2's own subject, "batch the liveness snapshot instead of caching the per-item probe", is already the correct one.
- **The "Added:" test list.** Still names `tmux::tests::cached_session_names_serves_a_fresh_snapshot` and `cached_session_names_is_empty_when_server_unreachable`. Neither exists; `cached_session_names` appears zero times in `src/tmux/mod.rs` at this head. The real names are `snapshot_lookup_matches_the_per_item_probe` (`:1727`), `snapshot_lookup_skips_a_dead_pane` (`:1740`), and `snapshot_lookup_reports_not_live_when_server_unreachable` (`:1750`).
- **"Commands run on this branch (based on `ec6b6d04`)"** — the head is now on `9e970918`. Counts drifted too: the checklist says 4,652, the commands block says 4,653, and the pre-existing-failure note lists one test where current main has three.

## Nits 3 and 4 — both untaken

`build_exclusion_set` still forks its own `list-sessions` at `src/session/capture.rs:1081` alongside the snapshot's. `src/tmux/mod.rs` is byte-identical, so `take()` still calls `batch_pane_metadata()` when `names` is `None`.

## Definition of done on `f69341f9`

Run fresh in a clean worktree: `cargo fmt --all -- --check` clean; `cargo clippy -- -D warnings` (CI form) clean; `cargo test --lib` 4664 passed / 3 failed. The three are `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents`, `tui::dialogs::hooks_install::tests::test_content_shows_settings_path`, and one the rebase pulled in, `cli::graft::tests::augmented_command_keeps_core_subcommands` (it is reading local plugin config and seeing `refresh` / `open_pr` / `open` subcommands). All three reproduce on unmodified `origin/main`. None related to this change.

The Rust CI legs still have not run on `f69341f9` — the earlier approval covered `f47ac5cc`. My audit of this head is clean: three Rust source files, no workflow or CI config, no install/build/test hook, no dependency or lockfile churn, nothing binary or obfuscated.

## Verdict

**NEEDS WORK, but only just.** One code change (make the snapshot lazy), one title edit, and three stale lines in the body's test and commands blocks. The design question is settled and I would not re-litigate it.

njbrake (commented):

_Reviewed at `10ea6a0a`. Push new commits and it will be re-reviewed._

---

## Re-review at `10ea6a0a`

### On the "make `take()` lazy" finding: **partially addressed, and the earlier finding sized it wrong**

Sharing is not a substitute for laziness, but it turns out to be the far more valuable change, and the author found that by measuring rather than by arguing. Taking the correction: the finding named a real cost but sized it wrong, and would have had you optimize the cheap half.

All three passes were measured directly, with a `tmux` argv-logging shim and a seeded 169-instance store, calling the real entry points:

| pass | `origin/main` | `10ea6a0a` |
|---|---|---|
| `HomeView::new` (two per-instance walks) | **338 forks** | **2** |
| `repair_session_id_pollers`, no pollers running | **169** | **2** |
| `repair_session_id_pollers`, every row short-circuits | **0** | **2** |

Row 2 is the one that matters. `repair_session_id_pollers` is on a 500ms timer (`src/tui/app.rs:1813`, `STATUS_REFRESH_INTERVAL` at `:868`) and walks *every* instance in the view, and `HomeView::new` loads all rows including trashed (`src/tui/home/mod.rs:2080-2100`). So on the author's store `main` is doing ~338 `list-sessions` forks **per second** on the thread that serves keystrokes. That is a bigger storm than the 87/s originally measured, it dwarfs the `compose_exclusion_with_persisted_peers` path this PR started with, and commit `61db415c` is what actually kills it. The author's own profile (`4389 of ~4700 main-thread samples` in `repair_session_id_pollers`) says the same thing, reproduced independently here.

Row 3 is the original finding, surviving intact: `has_live_tmux_pane_in(snapshot)` is the **last** clause in `repair_session_id_poller_if_needed` (`src/session/instance.rs:5938-5942`), after three in-memory short-circuits, so the eager `take()` at each caller pays for a probe the callee often never performs. Cost: 4 forks/sec in the TUI, 1 fork/sec in the daemon (2s tick, `src/server/mod.rs:6044`), and 2 per pass at `src/session/capture.rs:1031` regardless of tool. **The author has already quantified and accepted exactly this**: "about four calls per second... that is the intended trade: a fresh observation per pass instead of a possibly-`CACHE_TTL`-stale one." That is the right call and correctly argued. **Downgrading the finding from blocking to a nit**: a `OnceCell` at those four sites buys both properties for a few lines, but it is a follow-up, not a merge gate.

### No new staleness or correctness problem

Each of the three specific risks was checked:

- **Lifetime.** Nothing long-lived. `HomeView::new`'s snapshot spans two adjacent loops in one constructor; the other two span a single loop each. No snapshot outlives its pass or is cached across ticks.
- **Acting on a pre-create or pre-kill observation.** Possible in principle, strictly narrower in practice. Under `main` the same pass took ~850ms to walk 169 rows, so its result set was *already* internally inconsistent across a far wider window; the batched pass completes in single-digit milliseconds. Both error directions are non-destructive and self-heal: a false "not live" skips a repair and retries in 500ms; a false "live" starts a poller on a dead pane, `session_id_poller_is_running()` goes false, and it retries. Nothing here writes a durable sid; that still goes through the untouched flock/CAS path.
- **TUI vs daemon disagreement.** No change. They were always separate processes holding independently-timed observations; neither now trusts a *cache*, only its own fresh probe. And `src/session/sync.rs:583`, the post-CAS env publication jerome-benoit specifically flagged, is still on the live per-item probe, as is the one-shot holder cleanup at `src/session/instance.rs:5238`. Both correct.

### Scope note (new, not a blocker)

These commits go beyond the original prescription's "no change to what `has_live_tmux_pane()` means at its other five call sites." Every production call site has now migrated, which the measurement justifies, but it leaves **`has_live_tmux_pane()` (`src/session/instance.rs:3775`) with zero production callers**, surviving only in one test assert (`:14604`) and in doc comments at `src/session/recovery.rs:116,200`. It is `pub fn`, so the compiler will not say so. AGENTS.md's no-dead-code rule applies: either delete it and have the test use a snapshot, or keep it deliberately and say why. Check the `recovery.rs` doc comments either way, since they name it as the mechanism.

One remaining un-shared nesting, worth a follow-up issue rather than a change here: `repair_session_id_poller_if_needed` -> `maybe_start_poller` -> `retroactive_capture_exclusion_set()` (`src/session/instance.rs:5677`) -> `compose_exclusion_with_persisted_peers`, which takes its *own* snapshot plus a full `Storage::load()`. Only fires for rows actually being repaired, so it is bounded, but it is where **#3507** lands: its `claude_project_dir_is_shared` adds another `Storage::load()` on that same Claude path. #3507 now merges clean against both current `main` and this head (both tested; its earlier conflict with `main` is gone), so there is no merge-order problem, just a compounding-cost note for whoever lands the second one.

### The title and body: **still open, and now self-contradicting**

The Description is otherwise excellent and that deserves saying plainly: the author re-profiled, found their own earlier code-reading inference was wrong, said so in writing, re-measured with per-pid attribution, and published the trade against the rejected design instead of hiding it. That is better than what was asked for. What is still stale:

- **The title**, unchanged: `perf(tmux): serve session liveness from the cache instead of forking list-sessions`. This is the squash subject and it names the design the body now calls "the rejected cache-the-accessor version."
- **Line 24**: "`has_live_tmux_pane()` is unchanged at its other call sites" is no longer true, and the very next paragraph contradicts it with "Four call sites now share one snapshot."
- **Lines 78-79**: still lists `tmux::tests::cached_session_names_serves_a_fresh_snapshot` / `..._is_empty_when_server_unreachable`. The real names are `snapshot_lookup_matches_the_per_item_probe`, `snapshot_lookup_skips_a_dead_pane`, `snapshot_lookup_reports_not_live_when_server_unreachable`.
- **Line 81** "based on `ec6b6d04`" and **line 60** "4,652 library tests" are both stale, and the commands block predates `src/server/mod.rs` entering the diff, so it should now show a `--features serve` run.

### Definition of done, run fresh on the head rebased onto current `origin/main`

`cargo fmt --all -- --check` clean. `cargo clippy --features serve -- -D warnings` clean (the serve variant, because the diff now reaches `src/server/mod.rs`). `cargo test --features serve --lib`: 6379 passed, 2 failed, `worktree_move_observation_canonicalizes_symlinked_parents` and `test_content_shows_settings_path`, both reproduced on unmodified `origin/main`, both macOS-environmental.

CI note: the Rust legs still have not run on this head. Runs `32849766391` (CI) and `32849766420` (Tests) sit at `action_required`. An audit of the diff is clean (five Rust files, no workflow, hook, dependency or binary changes), so the gate is safe to release when the signal is wanted.

### Verdict

**NEEDS WORK** — the code is right and the blocking objection to it is withdrawn. What remains is a title that names the rejected design, one now-false sentence and three stale lines in the body, and a `has_live_tmux_pane()` left with no callers. All editorial, none of it design.

src/tui/home/mod.rs
```diff
@@ -2461,11 +2461,16 @@ impl HomeView {
...
+        // instance in the view, so a per-item `list-sessions` fork scales with
+        // the whole store — measured as the dominant tmux cost of this pass on
+        // a store of a few hundred sessions.
+        let live = crate::tmux::LiveSessionSnapshot::take();
         {
             let mut set_batch: Vec<(String, String, String)> = Vec::new();
             let mut unset_batch: Vec<(String, String)> = Vec::new();
             for inst in view.instances.values() {
-                let Some(tmux_name) = inst.tmux_env_session_name() else {
+                let Some(tmux_name) = inst.tmux_env_session_name_in(&live) else {
```
jerome-benoit: [P2] Retry this one-shot synchronization when the snapshot is `Unknown`. `LiveSessionSnapshot::take()` preserves a transient `list-sessions` failure as `names: None`, but `tmux_env_session_name_in()` collapses that to `None`, so this loop skips every row. Neither `reload()` nor `reload_storage_only()` reruns the bulk env sync, and an existing poller does not report an observation equal to its `initial_known` value. A single startup probe failure can therefore leave all `AOE_INSTANCE_ID` / `AOE_CAPTURED_SESSION_ID` entries missing or stale until an unrelated SID change or relaunch, weakening the ownership exclusion this block is meant to restore. Preserve `Unknown` through this caller and schedule a retry, or fall back to fresh per-item probes, with a regression where the first snapshot fails and the next succeeds.

src/tui/home/mod.rs
```diff
@@ -2461,11 +2461,16 @@ impl HomeView {
 
         // Batch-sync instance IDs and captured session IDs to tmux hidden env
         // so that build_exclusion_set() on other AoE instances can see them.
+        // One observation for both per-instance walks below. They visit every
+        // instance in the view, so a per-item `list-sessions` fork scales with
+        // the whole store — measured as the dominant tmux cost of this pass on
```
jerome-benoit: [P3] Replace the em dash in this added comment. The repository rules explicitly prohibit em dashes in human-authored comments and docs. The same issue is present in the added prose at `src/tui/home/mod.rs:3653`, `src/session/instance.rs:3790`, and `src/tmux/mod.rs:572`; please fix all four together.

jerome-benoit (dismissed):
Re-review at exact head `10ea6a0a`. The stale shared-cache design is gone, and the new pass-scoped snapshot preserves fresh per-item probes at the correctness-sensitive post-CAS publication and one-shot cleanup sites.

One correctness issue remains: `HomeView::new` collapses a transient `list-sessions` failure into "no live session" for every row during its one-shot hidden-env synchronization. Reloads do not retry that synchronization, and a poller does not emit an unchanged initial SID. The hidden ownership variables can therefore remain missing or stale until an unrelated SID change or relaunch.

The maintainer's existing dead-code and PR-metadata findings remain open, as does CodeRabbit's duplicate `list-sessions` comment; I have not duplicated them. `git diff --check` and `cargo fmt --all -- --check` are clean, but GitHub has not run the Rust CI jobs on this head.

src/tui/home/mod.rs
jona-wilmsmann: Fixed in `e894680f`, and this was the right thing to hold the PR on: it is the only place in the batch where a lost observation does not come back.

`HomeView::new`'s publication now resolves each row through `Instance::tmux_env_session_name_in_or_probe`, which falls back to a fresh per-item probe when the snapshot could not reach the server rather than reading `names() == None` as "no live pane". I took the fallback rather than the retry you also offered, because it restores exactly what the pass did before this PR without introducing retry state, and because the same file already draws this distinction the other way: startup recovery skips its whole pass on a failed `batch_pane_metadata` rather than treating it as "every pane is dead".

The walks that do re-run (peer exclusion, poller repair, the second `HomeView::new` walk) still collapse Unknown to "not live", which is what the per-item probe did when its own `list-sessions` failed.

Regression: `instance::tests::one_shot_name_probes_when_the_snapshot_missed_tmux`. With a `tmux` shim on `PATH` standing in for the probe that succeeds after a failed snapshot, an unreachable snapshot must still resolve the row, and a snapshot that did reach the server stays authoritative with no probe behind it. I checked it fails with the fallback removed, so it is pinning the behaviour rather than passing vacuously.

src/tui/home/mod.rs
jona-wilmsmann: Done in `e894680f`, all four: `src/tui/home/mod.rs:2466` and `:3653`, `src/session/instance.rs:3790`, `src/tmux/mod.rs:572`.

I left the em dashes in the surrounding pre-existing prose alone rather than widening the diff. `git diff origin/main...HEAD | grep '^+' | grep '—'` is empty at this head.

njbrake (commented):

_Reviewed at `e894680f`. Push new commits and it will be re-reviewed._

---

## Re-review at `e894680f`

Both blocking findings are resolved, the two nits carried since the first round are taken, and the laziness commit does what its subject says. The harness was re-run rather than anything carried forward.

### Laziness landed, verified, row 3 is 0

Same shim, same seeded 169-instance store, but this time the shim answers `list-sessions` with the rows' real generated names so the live paths are actually exercised (the earlier table used an empty-name shim and undercounted both sides):

| pass, 169 rows | `origin/main` | `e894680f` |
|---|---|---|
| `HomeView::new` | **62,180** forks (30,512 `list-sessions`) | **1,671** (565) |
| `repair_session_id_pollers`, no pollers running | **41,340** (20,610) | **858** (328) |
| `repair_session_id_pollers`, **every row short-circuits** | 0 | **0** |
| `compose_exclusion...`, gemini (never consults liveness) | 2 (1) | **2** (1) |
| `compose_exclusion...`, claude, 169 peers | **341** (170) | **8** (5) |

Wall clock for the identical test body: **493s on `main`, 19.9s at the head.**

Row 3 is the one that had to move and it moved to zero. Row 4 is the blocking finding, closed: the gemini pass now costs exactly what `main` costs, so no pass is more expensive than before the PR, which was the whole substance of the objection. That is better than the `OnceCell` retrofit suggested last round, because folding `build_exclusion_set` onto the same snapshot (`src/session/capture.rs:1096-1103`) also removed the duplicate `list-sessions` raised as nit 3.

### Mechanism, and the `OnceLock` error path

`names: OnceLock<Option<Vec<String>>>` / `panes: OnceLock<Option<HashMap<..>>>` (`src/tmux/mod.rs:607-608`), each filled by `get_or_init` in its accessor. The failure mode worth checking:

**A failed `list-sessions` is memoized as `None` = Unknown, not as "no sessions live."** That distinction is load-bearing and every consumer handles it deliberately:

- `build_exclusion_set` uses `let Some(names) = live.names() else { return HashSet::new() }`, identical to `main`'s `_ => return HashSet::new()`.
- `live_any_kind_name_for_id_in` uses `names()?`, giving "not live". In the exclusion walk that sets `should_exclude = true`, the *safe* direction (more exclusion, no steal). In poller repair it skips and retries in 500ms.
- The one non-self-healing consumer, `HomeView::new`'s startup env publication, now routes through `tmux_env_session_name_in_or_probe` (`src/session/instance.rs:3782-3800`), which falls back to a fresh per-row probe when `names()` is `None`. That is precisely the "one-shot pass that cannot retry" hole worth going looking for, found and closed before this review got there, with a real regression test (`one_shot_name_probes_when_the_snapshot_missed_tmux`) that the author says he mutation-checked by removing the fallback.

Worth noting the memoization is not new: the eager `take()` captured the same `Option`s once per pass. Laziness changed *when*, not *whether*.

**Staleness, re-derived rather than carried forward.** One thing genuinely changes: `names` and `panes` are now initialized at different moments instead of back to back. A session in `names` but killed before `panes` is fetched is absent from the pane map, so `pane_dead` returns `false` and it is reported alive, which is exactly what `is_pane_dead`'s `unwrap_or(false)` did per item on `main`. Same direction, and the window is *narrower* than the eager version's, since both calls now happen inside the same loop iteration rather than across two sequential forks. No snapshot outlives its pass; `LiveSessionSnapshot::new()` is constructed inside each of the five passes and dropped with it. The only new cross-consumer skew is in the tmux-unreachable case, where `HomeView::new`'s loop 1 probes per row while loop 2 reads the snapshot's `None`, so a row can get its env published and its poller repair deferred to the next 500ms tick. Harmless and self-healing.

### `src/tmux/mod.rs` (+154) is confined to the snapshot path

Diffed: the entire change is `LiveSessionSnapshot` plus `live_any_kind_name_for_id_in` plus their tests. `take()` becomes `new()` plus `get_or_init` accessors, doc rewrite, and the two happy-path tests folded into one table (nit 5, taken). Nothing else in the module is touched; `session_exists`, `SESSION_CACHE`, `batch_pane_metadata`, `resolve_session_name` are all untouched. `pane_dead` now lazily inits, so an unreachable server never pays for `list-panes` at all, which was nit 4, also taken.

### Dead code: resolved

`has_live_tmux_pane()` is **deleted**; zero occurrences of the bare name remain anywhere in `src/`. And the two `src/session/recovery.rs` doc comments (`:116`, `:200`) that named it as the mechanism were reworded rather than left dangling, which is the stale-mirror case this review flags as a matter of habit and rarely sees anyone actually chase.

### Title and body: resolved

Title is now `perf(tmux): take one liveness snapshot per pass, not per instance`, which is accurate. The body's test list now names the three tests that exist; "unchanged at its other call sites" is gone; the base is `9e970918`; the emdashes in added comments were cleaned up while leaving surrounding prose alone. The one remaining nitpick, that the 8-second end-to-end table was taken at `10ea6a0a`, is explicitly labelled as such with a note on why the per-pass shim numbers cover the newer commit. That is the right way to report it.

### Definition of done, on the head rebased onto current `origin/main`

Run with `env -u CLAUDE_CONFIG_DIR`:

- `cargo clippy -- -D warnings`: clean
- `cargo clippy --features serve -- -D warnings`: clean
- `cargo test --features serve --lib`: **6411 passed, 0 failed, 2 ignored.** No failures at all, environmental or otherwise; #3535 did make it green and this PR keeps it there.

#3507 merges clean against this head (tested). jerome-benoit's `CHANGES_REQUESTED` asked for a fresh fallible per-pass snapshot preserving Present/Absent/Unknown, with regression coverage for the post-CAS env publication. The Unknown distinction is now actually load-bearing via `_or_probe`, `sync.rs:583` and the one-shot holder cleanup still use the live probe, and the regression test exists; his objection is addressed on the substance, though the review state is his to clear.

CI's Rust legs are still gated on fork approval for this head (runs `32948878671`, `32948878664`); the diff is six Rust files with no workflow, hook, dependency or binary changes.

### Verdict

**LGTM** — every finding raised across four rounds is closed with evidence, the change now makes no pass more expensive than `main` while cutting the hot ones by 25-40x, and the suite is fully green.

PR #3509, opened 2026-08-24
fix(session): resolve the detect_as alias at launch, and re-resolve it on tool swap

njbrake:
## Description

A non-sandboxed session running a custom agent (`[session.custom_agents]` + `[session.agent_detect_as]`) could launch with **no `AOE_PROFILE`/`AOE_INSTANCE_ID` on its command line**. Every status hook aoe installs opens with `[ -n "$AOE_INSTANCE_ID" ] || exit 0`, so all of them silently no-op: `/tmp/aoe-hooks-<euid>/<id>/status` is never written, the session reports **Idle forever**, and nothing is logged.

Found on a live session that had been frozen at Idle for four days. Its agent process carried neither variable, while every other session's did; its hook sidecar's mtime predated the pane it was supposed to belong to.

### Root cause

Both symptoms come from reading `Instance::detect_as` raw. #3398 established that the persisted field is a *cache* rather than the authority and routed three read sites through `status_rules::effective_detect_as` — but the launch path kept resolving by hand:

```rust
let agent = get_agent(&self.tool).or_else(|| get_agent(&self.detect_as));
```

With an empty alias that yields `None`, and `status_hook_env_prefix` emits nothing for `None`. The sandbox branch appends `-e AOE_PROFILE -e AOE_INSTANCE_ID` unconditionally, which is why only host sessions could hit this — 1 of ~24 custom-agent sessions on the reporting machine.

The sidecar is also where a custom-agent session's `agent_session_id` comes from (`aoe __extract-session-id`), so a row in this state loses track of its conversation as well — the reported row's `agent_session_id` was absent and the id of its live 5.9 MB transcript was known only to Claude Code.

`swap_tool` is how the row got an empty alias in the first place: it reassigns `self.tool` and carefully migrates `agent_session_id`, `acp_session_id`, `resume_intent`, `acp_effort` and `agent_model`, but left `detect_as` untouched. That is not merely stale — a `codex-personal` -> `claude-personal` swap kept aiming `resolved_agent` at codex. The reported row was created on plain `claude` (nothing to store) and swapped onto `claude-personal` two days *after* v024 backfilled every existing row, so the migration could not have covered it.

### Changes

- `Instance::resolved_agent` / `Instance::effective_detect_as` resolve through the live registry, and every launch-time reader now goes through them: hook install, the launch-command build, container agent selection, managed `CODEX_HOME`, the container config refresh, and yolo trust config. Per v024's own note, leaving the field wrong on disk means every future reader has to know to distrust it, so this converts the remaining hand-rolled readers rather than fixing only the one that was reported.
- `swap_tool` re-resolves `detect_as` for the incoming tool, off the same process-global registry `effective_detect_as` reads (a lookup, not a config load), so the row ends up exactly as if it had been built on the new tool.

The fallback is a strict no-regression: it only fires when the stored field is empty, which today is already the broken state.

### Deliberately not fixed here

`apply_session_flags` -> `append_resume_flags` -> `build_resume_flags` key on the raw `self.tool` rather than the alias, and `get_agent` only searches the static `AGENTS` table. So **no custom-agent session ever emits `--session-id`/`--resume`**, and `aoe session set-session-id` is accepted and then silently dropped at launch. Same class of raw read, but a wider behavior change — it would start splicing resume subcommands into wrapper command lines (`codex resume <sid> --profile ...`) — so it wants its own PR rather than riding along here.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt`, `cargo clippy --all-targets` (no new warnings), `cargo test` — 4651 passed. Two pre-existing failures are unrelated and reproduce on the parent commit: `tui::dialogs::hooks_install::tests::test_content_shows_settings_path` fails whenever `CLAUDE_CONFIG_DIR` is set in the environment (passes under `env -u CLAUDE_CONFIG_DIR`), and `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` fails on this symlinked checkout.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: an e2e test would have to stand up a custom-agent wrapper on `PATH`, launch it non-sandboxed, and assert on the launched process's environment — roughly 10-28s of the serial e2e budget for a defect that is fully determined by pure resolution logic. Two unit tests pin it instead, both of which fail on the parent commit:

- `empty_detect_as_still_resolves_the_launch_agent` — a custom-agent row with an empty stored alias must resolve its built-in agent and keep the `AOE_PROFILE`/`AOE_INSTANCE_ID` launch prefix. Fails on `main` with `resolved_agent()` returning `None`, exactly as the reported session did.
- `swap_tool_reresolves_detect_as` — table over built-in -> custom (the reported row), custom -> custom (the actively-wrong pin), and custom -> built-in.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 (1M context), via Claude Code.

Used to diagnose the live frozen session (agent process env vs. hook sidecar mtimes vs. `session.status_change` history) and to write the fix and tests. Diagnosis and patch reviewed by me before pushing.

## Summary

- Resolve session agent aliases from the active profile registry during launch, status hooks, container setup, managed `CODEX_HOME` configuration, and status detection.
- Re-resolve aliases when tools change to prevent stale agent mappings.
- Restore the session profile before tool swaps so profile-specific aliases resolve correctly.
- Add regression tests for empty aliases, tool swaps, and profile-specific session restarts.

These changes prevent custom-agent sessions from remaining idle because required launch environment variables are missing.

## Potential enhancement

- Add resume flag support for custom agents in a separate change, as planned.

### DIFF
--- a/src/session/instance.rs
+++ b/src/session/instance.rs
@@ -2036,6 +2036,18 @@ impl Instance {
                 .insert(self.tool.clone(), outgoing);
         }
         self.tool = new_tool.to_string();
+        // The alias is resolved per-tool, so the outgoing tool's answer cannot
+        // survive: kept, it points `resolved_agent` at the wrong built-in
+        // outright (a `codex-personal` -> `claude-personal` swap would keep
+        // detecting as codex); cleared, the row lands in the same
+        // empty-`detect_as` state a session built before its tool joined
+        // `[session.agent_detect_as]` does. Re-resolve against the same
+        // process-global registry `effective_detect_as` reads, so this stays a
+        // lookup rather than a config load, and the row ends up exactly as if
+        // it had been built on the new tool.
+        self.detect_as =
+            tmux::status_rules::effective_detect_as(&self.source_profile, new_tool, "")
+                .into_owned();
         // Consumed, not copied: the row owns exactly one live conversation per
         // agent, and leaving the entry behind would let a later swap restore an
         // id this session has since replaced.
@@ -2609,6 +2621,32 @@ impl Instance {
     /// when `source_profile` was never populated (e.g. legacy callers).
     pub fn effective_profile(&self) -> String {
         super::config::effective_profile(&self.source_profile)
+    }
+
+    /// The `agent_detect_as` alias that actually applies to this session.
+    ///
+    /// `detect_as` is resolved once at session build and persisted, so it is
+    /// empty on a row created before its tool gained an
+    /// `[session.agent_detect_as]` entry. Treat the stored field as a cache
+    /// and let [`tmux::status_rules::effective_detect_as`] consult the live
+    /// registry when it is empty, the same way the pane detector, hook
+    /// reconciliation, and the status-change log line already do (#3398).
+    fn effective_detect_as(&self) -> std::borrow::Cow<'_, str> {
+        tmux::status_rules::effective_detect_as(&self.source_profile, &self.tool, &self.detect_as)
+    }
+
+    /// The built-in agent backing this session: its own tool when that names
+    /// one, else the agent its `agent_detect_as` alias points at.
+    ///
+    /// Every launch-time consumer resolves through here rather than reading
+    /// `detect_as` raw, because a miss is silent and permanent. `None` drops
+    /// the `AOE_PROFILE`/`AOE_INSTANCE_ID` prefix from the launch line
+    /// ([`status_hook_env_prefix`]) and skips hook install, so every hook the
+    /// agent does have bails on `[ -n "$AOE_INSTANCE_ID" ]` and the session
+    /// reports Idle forever with nothing logged.
+    fn resolved_agent(&self) -> Option<&'static crate::agents::AgentDef> {
+        crate::agents::get_agent(&self.tool)
+            .or_else(|| crate::agents::get_agent(&self.effective_detect_as()))
     }
 
     /// Resolve the effective `environment` list for this session's profile,
@@ -3750,9 +3788,10 @@ impl Instance {
             .as_ref()
             .ok_or_else(|| anyhow::anyhow!("sandbox_info missing for sandboxed session"))?;
 
+        let detect_as = self.effective_detect_as().into_owned();
         let managed_codex_home = container_config::managed_codex_home(
             &self.tool,
-            Some(&self.detect_as),
+            Some(detect_as.as_str()),
             &self.source_profile,
             &self.id,
         )?;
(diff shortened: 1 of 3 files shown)


njbrake (commented):
Approving. Correct diagnosis and the right mechanism: this extends #3398's `status_rules::effective_detect_as` rather than adding a second resolver beside it, and `swap_tool`'s new write mirrors what `builder.rs:751` already does at create time.

I verified the launch path actually primes the registry it now depends on. `run_pre_launch_hooks` calls `mint_host_session_env` first, which reaches `repo_config::resolve_before_session_hooks` -> `resolve_config_or_warn` -> `install_from_config` before `install_agent_status_hooks` runs. That early-returns for sandboxed sessions, but the sandbox path was never the broken one.

One defect found and fixed in 2ae85a1: `persist_tool_swap` calls `swap_tool` on a storage-loaded row, and `source_profile` is `skip_serializing`, so that row always comes back blank. The new alias re-resolution therefore keyed the default profile's `agent_detect_as` entries rather than the session's own. It sticks, because `detect_as` is not in `reconcile_from_disk`'s carry set, so the next launch reads the disk row rather than the correct in-memory one written alongside it at `operations.rs:490`. Where a tool name is aliased differently per profile the row lands pinned to the wrong built-in, which is the failure mode this PR exists to remove. Restoring `source_profile` before the swap, the way `reconcile_from_disk` already does, is the fix; the regression test fails without it.

Worth knowing when reading the diff: six of the touched call sites were already self-healing. `container_config::resolve_active_agent` has a third fallback that loads `session.agent_detect_as` itself, so `managed_codex_home`, `ensure_yolo_trust_config_for_active_agent`, and `ContainerAgentSelection` were never broken. The behavioral fix is `run_launch_hooks`, `build_launch_command`, the `build_container_config` sidecar lookup, and `swap_tool`. Good consistency work either way, and it narrows the risk.

Two follow-ups, neither blocking:

Because `effective_detect_as` gives a non-empty stored value absolute precedence, writing a resolved alias in `swap_tool` converts a self-healing row into a pinned one. Before this, a row created on a built-in and swapped onto a custom tool kept an empty `detect_as` and tracked config live; now it is pinned to whatever the registry said at swap time, and a later retarget of that alias will not reach it until the next swap. Consistent with create-time behavior, so defensible, but it is a real narrowing of the healing the first half of this PR relies on and nothing would surface it except a user report.

`telemetry/mod.rs:256` and `tui/status_poller.rs:184` still read `detect_as` raw. The poller one is harmless (it self-heals through `resolve_active_agent`); the telemetry one buckets a custom-agent row under its wrapper name. Cosmetic.

Local verification: `cargo fmt` clean, `cargo clippy --all-targets` clean apart from six pre-existing warnings, `cargo test --lib` 4663 passed with only the known `CLAUDE_CONFIG_DIR` failure, which passes under `env -u CLAUDE_CONFIG_DIR`.

PR #3511, opened 2026-08-24
fix(session): drop the apply_status_intent stamp that wipes concurrent archives

jerome-benoit:
## Description

Follow-up to #3481 (its merged fix removed the plain-tmux passive stamp). This drops the SECOND last_accessed_at stamp, the one inside `apply_status_intent` for structured rows, which reaches disk through the TUI relay and reproduces the same concurrent-archive wipe. Requested explicitly in the post-merge review of #3481.

### Maintainer request being implemented (review on #3481)

> Get the `apply_status_intent` residual into #3465. Your read of it is right and the issue's is now stale: the stamp at `server/mod.rs:5777` does reach disk, through the relay you describe. `DaemonStatusPoller` carries the field on the wire (`daemon_status_poller.rs:80`), `apply_status_update` writes it into TUI memory unconditionally (`home/mod.rs:3273`, `3331`), and while `persist_passive_status_transition` gates on `is_structured()`, `save()`'s `merge_from_tui` monotone max does not. A second writer surface with a stale `pre` then hits the same arm. Since this is data loss, I do not want the remaining half tracked only in a merged PR body. Either drop that stamp here (structured rows already touch on every user prompt via `api/acp.rs:1202`, so nothing...)

### RCA (all links verified on this branch, org main bbc45320 lineage)

1. `apply_status_intent` stamped `last_accessed_at = Some(now)` on every worker-event transition (`src/server/mod.rs:5777`), daemon memory only.
2. Relay: `GET /api/sessions` serializes the field (`src/server/api/sessions.rs:390`) -> `DaemonStatusPoller` parses it (`src/tui/daemon_status_poller.rs:80`) -> `apply_status_update` writes it into TUI memory unconditionally when `Some` (`src/tui/home/mod.rs:3273`, `3331`).
3. Persistence gate missing: `HomeView::save` folds every TUI row into disk via `disk_inst.merge_from_tui(tui_inst)` (`src/tui/home/mod.rs:6509`), whose `last_accessed_at` arm is an ungated monotone max (`src/session/instance.rs:1991`). Disk now carries the phantom stamp.
4. Wipe: a second writer surface whose `pre` snapshot predates that advance trips `touched = self.last_accessed_at > pre.last_accessed_at` (`instance.rs:2217`) and loses `archived_at` / `snoozed_until` / `idle_dormant_since` (`:2244-2247`) through `merge_user_action_diff`. Same shape as #3481, narrower surface (structured rows, requires a second writer racing within one relay interval).

"Never reaches disk" is refuted by construction: `merge_from_tui` has no `is_structured()` gate and the new regression test drives the real `merge_from_tui` + real `merge_user_action_diff`.

### Fix

Delete the stamp line in `apply_status_intent`; keep `now` for `idle_entered_at` bookkeeping and the `StatusChange.at` timestamp. Dropping it exposed that prompts on live structured rows never refreshed the field either: `touch_and_wake_if_sunk` only fired for sunk rows, and the stamp had been masking that at every turn boundary. The function is now `touch_on_prompt_and_wake_if_sunk`: it touches on every prompt in daemon memory, then persists in two tiers via `apply_prompt_persist_to_disk`: a waking prompt keeps touch semantics on disk (lifts sink state), while any other prompt advances recency monotonically without clearing sinks, so the #3465 wipe cannot return through this save path. Restores the premise the review quoted. Message/send endpoints still touch as well. The deliberate `touched` arm and its pinned messaging-unarchives test are untouched. Idle-reap is unaffected: it excludes structured rows outright (`src/session/idle_reap.rs:47-49`) and its anchor math is untouched for plain rows.

### Tests

- `relayed_intent_stamp_wipes_concurrent_archive`: full residual chain through real `apply_status_intent` -> `merge_from_tui` -> `merge_user_action_diff`; red before the fix, green after (captured live both ways by temporarily restoring the stamp).
- `plain_prompt_recency_never_clears_a_peer_archive` / `wake_persist_keeps_touch_semantics_for_sunk_rows`: pin both tiers of the new disk-side persist (recency-only never clears sinks; wake keeps touch semantics).
- `status_intent_transitions_preserve_last_accessed_at`: unit pin across Set(Running) / Set(Idle) with an exact-equality sentinel.
- Regression: merge family 124 passed; idle_reap 15 passed (anchor invariant unchanged); decide_passive_transition + flush_passive_transition + intent/heal/trailing tests 15 passed under `--features serve`.
- `cargo fmt --all -- --check` clean; `cargo clippy -- -D warnings` clean.

### Behavior changes to be aware of

- Structured-row recency now advances on each prompt instead of at worker-event turn boundaries; the attention sort's aging heuristic no longer inverts for actively-conversed structured sessions.
- Queueing a follow-up now advances recency too (`SessionService::enqueue_prompt`), as a monotone max that never lifts a sink. Both enqueue routes land there, so the web composer's `POST /queue` and the TUI's `/acp/prompt` Tier 3 `Queued` stay symmetric. Without it, dropping the intent stamp would have left an actively-queued session ageing into the top of the attention sort as the most neglected one.
- Write volume: a prompt on a live structured row now does a `spawn_blocking` read-modify-write of the profile session file, where `main` wrote nothing unless the row was sunk. Prompts are human-paced so this is not expected to matter, but it is a real change. The queue-side advance adds no write of its own; it rides the `Storage::update` that `enqueue_prompt` already performed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the wipe needs a daemon-to-TUI relay interleaved with a concurrent writer holding a stale snapshot; the unit pair locks both ends deterministically (source stamp removal pinned directly, end-to-end wipe reproduced through the real `apply_status_intent` -> `merge_from_tui` -> `merge_user_action_diff` sequence).

Considered but not changed (documented per review thread): `merge_passive_status_patch`'s dropped-patch `debug!` now fires once per passive transition. Lowering it to `trace!` would keep `drop_log_count` and both `traced_test` counts passing unchanged (`tracing-test` runs with `no-env-filter`, so TRACE is captured); left as-is to keep this PR single-purpose.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** ox-alpha (via Oh My Pi harness)

**Any Additional AI Details you'd like to share:** Follow-up cycle executed autonomously on top of the merged #3481 work; adversarial self-review loop ran until two consecutive zero-finding tours.

**NOTE:**

Ref #3465

## Summary

- Removed recency updates from status transitions.
- Updated prompts and queued prompts to advance recency monotonically.
- Preserved archive, snooze, and dormancy state during ordinary recency updates.
- Kept wake behavior for prompts that wake sunk sessions.
- Added regression tests for relay persistence, status transitions, merge behavior, and state preservation.
- Updated related documentation and comments.

## Benefits

This prevents stale daemon or TUI state from overwriting newer session metadata. It keeps session recency accurate while preserving user-controlled session state.


njbrake (commented):

_Reviewed at `a10ec9e8`. Push new commits and it will be re-reviewed._

---

## Relationship to #3481, and the file list

Two things worth stating up front, because both look like problems from outside the diff and neither is one.

**The title and file list agree.** `apply_status_intent` lives in `src/server/mod.rs:5764`, which is in the changed set. `fix(session):` is a subsystem scope (session state), matching the merged sibling #3481, which used the identical scope with an overlapping file set. The two `web/` files are pure comment renames (`touch_and_wake_if_sunk` → `touch_on_prompt_and_wake_if_sunk` in a JSDoc block and a code comment); zero behavior.

**This is a strict follow-up to #3481, not a competing attempt.** #3481 merged at `bbc45320` before this branch was cut, and `git merge-base` for this PR is exactly `bbc45320`. #3481 removed the plain-tmux passive stamp (`update_status_with_metadata` / `PassiveStatusPatch`); this removes the structured-row stamp in `apply_status_intent`. No ordering question, nothing to close.

## The bug is real, chain verified on this branch

Each link, read rather than restated:

1. `src/server/mod.rs:5777` (pre-diff) — `inst.last_accessed_at = Some(now)` on every intent transition, no gesture behind it.
2. `src/server/api/sessions.rs:390` serializes it → `src/tui/daemon_status_poller.rs:80` parses it → `src/tui/home/mod.rs:3273`, `3331` write it into TUI memory unconditionally.
3. `src/session/instance.rs:1991` — `merge_from_tui` does `self.last_accessed_at = self.last_accessed_at.max(src.last_accessed_at)` with **no `is_structured()` gate**. The "never reaches disk" reading is genuinely refuted here.
4. `src/session/instance.rs:2217` — `let touched = self.last_accessed_at > pre.last_accessed_at;` → `:2244-2247` clears `archived_at` / `snoozed_until` / `idle_dormant_since`.

The claim that idle-reap is unaffected also holds: `src/session/idle_reap.rs:47-49` `continue`s on `is_structured()` before the anchor math, and the ACP reaper in `acp_reconciler.rs:884-900` anchors on `last_event_at_for_sessions` from the event store, not on `last_accessed_at`. No reap behavior changes.

Every remaining writer I could find is either a real gesture (`cli/send.rs:101`, `sessions.rs:11059`/`11082`, `groups.rs:2765`, `cli/session.rs:1493`) or a carry-forward rather than a fabrication (`server/mod.rs:3417`, `3468`). No passive fabricator left.

## Blocking findings

### 1. `apply_prompt_persist_to_disk`'s doc comment and its test name claim more than the code delivers

`src/server/api/acp.rs:1258-1261` says the recency-only tier "must NOT clear sinks, or #3465's wipe returns through this save path for archives a peer process wrote after this process's snapshot." That is the load-bearing claim of the whole two-tier design, so I reproduced it rather than reasoned about it. Scratch test added to your `tests` mod:

```
peer archives on disk → writer snapshots `pre` (pre-prompt) → apply_prompt_persist_to_disk(&mut disk, false)
  → archived_at still Some  ✅ (direct clear avoided)
→ later unrelated user action: disk.merge_user_action_diff(&pre, &post)
  → archived_at == None     ❌
```

```
test server::api::acp::tests::SCRATCH_recency_only_still_arms_touched_arm_for_a_later_writer ... FAILED
  panicked: recency-only persist still armed the touched arm and lost the archive
```

Tier 2 avoids *this process's direct clear*; it does not stop the wipe. It advances `last_accessed_at` on disk, which is precisely the signal `instance.rs:2217`'s `touched` arm keys on, so any later writer holding a pre-prompt `pre` still loses the archive one hop later. And this is a *new* disk-side advance: on `bbc45320` the non-sunk branch wrote nothing to disk at all, so the trap was not armed.

The *behavior* is defensible — `instance.rs:2236-2238` states the invariant outright ("any advance of `last_accessed_at` dethrones a concurrent archive"), and a prompt is a genuine touch, so a prompt beating an archive is a reasonable outcome. The problem is that the comment and `plain_prompt_recency_never_clears_a_peer_archive` both assert an absolute the code does not provide, on the exact invariant #3465 is about. The next person to read this will trust it.

What tier 2 *actually* buys is narrower and worth saying plainly: it stops a **stale daemon-memory view** from clearing a sink it has not observed yet (`wake` is computed from `state.instances`, which may not have picked up a peer's archive). That is a good property. Please say that instead, and rename the test to match what it proves (e.g. `recency_only_persist_does_not_itself_clear_a_peer_sink`).

### 2. The server-owned prompt queue is not covered, and this PR removes the mask that hid it

The justification is "structured rows take real touches from user prompts instead" (`server/mod.rs:5781`). That holds for `/acp/prompt` and `/acp/prompt/diff-comments`. It does not hold for the web composer's follow-up path:

- `web/src/hooks/useAcpSession.ts:1896-1913` — `enqueueServer` dispatches to `enqueueServerPrompt`, which POSTs `/api/sessions/{id}/queue` (`web/src/lib/api.ts:1563-1580`).
- `src/server/api/queue.rs:76` `queue_enqueue` — no `touch_last_accessed`, no `touch_on_prompt_and_wake_if_sunk`.
- The drain (`src/server/session_service.rs:1076-1128` → `send_turn`) never touches either; `grep -c touch_last_accessed src/server/session_service.rs` returns `0`.

Typing a follow-up while the agent is working is a first-class gesture in the structured view, and on this branch it advances no recency at all. Before this PR the queued turn's Running/Idle edges hit `apply_status_intent` and restamped, so the gap was invisible. The consumers that now see a stale value are real: `src/tui/home/render.rs:1857-1859` (last-activity column), `src/session/groups.rs:738-740` (attention sort, ASC — "longest ignored first", so an actively-queued session bubbles *toward the top* as if forgotten), `web/src/components/Dashboard.tsx:265`, and `web/src/App.tsx:1269`.

The TUI structured view is unaffected: `src/tui/structured_view/mod.rs:1591` always POSTs `/acp/prompt` and lets the server return `Queued`, so it touches. Only the web path bypasses. That asymmetry between the two clients is itself worth not shipping.

A `touch_last_accessed` (or the same two-tier persist) in `queue_enqueue` looks like the whole fix. If you would rather scope it out, please say so in the body — right now the PR reads as though the prompt path fully replaces what the stamp was doing, and it does not.

## Nits

**3. `docs/development/internals/sessions.md` should get the structured half.** #3481 added "`last_accessed_at` is user-gesture-only since #3465" to the **Plain tmux sessions** bullet (line 36). The **Structured / ACP sessions** bullet directly below (line 37) documents authority but says nothing about `last_accessed_at` — and this PR is what makes that half true. The checklist says docs were updated where necessary; the sibling PR set the precedent for exactly this file. One sentence.

**4. Divergence from the send path it was written to mirror.** The old doc comment said "matching the tmux send path (`/api/sessions/{id}/send`)", and the diff drops that phrase. `src/server/api/sessions.rs:11082` still does an unconditional `disk_inst.touch_last_accessed()` on the disk row for every send, which *will* clear a peer sink the daemon has not seen. So the two sibling gesture paths now have different disk semantics. The body notes "message/send endpoints still touch as well," which acknowledges it, but if tier 2 is the right shape it is the right shape there too. Worth a follow-up issue rather than scope creep here.

**5. New per-prompt `Storage::update`.** Every prompt on a live structured row now does a `spawn_blocking` full read-modify-write of the sessions file where main did none. Prompts are human-paced so this probably does not bite, but it is a real change in write volume the body does not mention.

**6. Two `now`s.** `inst.touch_last_accessed()` at `acp.rs:1202` and `chrono::Utc::now()` at `acp.rs:1266` produce different timestamps, so disk lands slightly ahead of daemon memory. The monotone max makes it harmless; worth knowing only because it means the daemon row is momentarily the older of the two.

## Coverage matrix: mandate does not apply, and CI is right to be green

Checked properly rather than inferred from the green tick. `web/tests/validate-coverage-matrix.mjs` enforces matrix-or-exempt membership only for `.tsx` files under `web/src/components/**` (lines 4-11, `walkTsx` rooted at `componentsRoot`). `web/src/App.tsx` is not under `components/`, and `acpTypes.ts` is `.ts`. Beyond the mechanics, both edits are comment-only, so the AGENTS.md mandate genuinely does not attach. `node web/tests/validate-coverage-matrix.mjs` → `Coverage matrix validation passed.` The `N/A` checkbox is correct.

## Definition of done, verified locally

| Gate | Result |
|---|---|
| `cargo clippy --features serve --all-targets` | 11 warnings, **all** in files this PR does not touch (`acp/supervisor.rs`, `session/groups.rs`, `session/smart_rename.rs`, `tui/home/mod.rs`, `acp/background_agent.rs`, `process/macos.rs`). CI runs `cargo clippy -- -D warnings` (`.github/workflows/ci.yml:40`), lib-only, clean. Not yours. |
| `cargo test --features serve --lib` | 6369 passed, 2 failed. Both pre-existing and host-environmental: `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` (macOS `/private` symlink) and `tui::dialogs::hooks_install::tests::test_content_shows_settings_path` (local `.claude` layout). Neither file is in the diff. All four new tests pass. |
| `cd web && npm run format:check` | clean (oxfmt, 928 files) |
| `cd web && npm run lint` | clean |

Test shape follows house style — four focused tests, no constant/default/derive restatements, and `status_intent_transitions_preserve_last_accessed_at` correctly folds two transitions into one test rather than splitting.

## Bot findings

CodeRabbit posted no actionable comments and rated merge risk minimal. Its one warning is a 79-char title against a ~70-char guideline — cosmetic, and I would leave it; the title is doing real work. I disagree with the "minimal risk" framing given finding 2, which a diff-local reviewer would not catch because the gap is in a file the PR does not touch.

## Merge risk

**Low-moderate.** Reverts cleanly, no schema or migration. The dominant residual is finding 2: recency silently stops advancing for web-composer queued prompts, degrading the attention sort and the TUI activity column. Nothing in the PR would surface it — no test covers `queue_enqueue`'s recency behavior, and the failure mode is a mis-ordered list rather than an error. You would hear about it from a user saying the sidebar ranks their busiest session as the most neglected one. Finding 1 is documentation risk rather than runtime risk, but it sits on a data-loss invariant, which raises its weight.

## Verdict

**MERGE WITH NITS**, conditional on two cheap changes:

1. Correct the `apply_prompt_persist_to_disk` comment and the `plain_prompt_recency_never_clears_a_peer_archive` test name to what the code actually guarantees (finding 1).
2. Either touch in `queue_enqueue`, or explicitly scope the queue path out in the body (finding 2).

The core removal at `server/mod.rs:5777` is correct, the chain is exactly as described, and the regression test drives real `merge_from_tui` plus real `merge_user_action_diff` rather than mocking the interesting part — the right call, and the reason the tier-2 claim was checkable at all. The one substantive disagreement is that the two-tier persist is presented as closing the wipe when it closes a narrower and different hole. That hole is still worth closing; it just needs to be described accurately.

PR #3518, opened 2026-08-25
fix(tui): stop forking tmux per row in Terminal and Tool views

njbrake:
## Description

Scrolling the Terminals list was visibly laggy while the agents list stayed snappy, and the lag persisted with a live terminal view open. Both surfaces asked tmux for pane liveness at paint time: `exists()` routes to `tmux::session_exists`, which trusts the shared session cache only on a HIT and falls through to a live `has-session` on a miss. Most sessions have no paired terminal, so nearly every row missed and forked a subprocess. Tool view forked twice per row, adding an uncached `is_pane_dead()`.

Two render paths were affected:

- **List rows** (`render_item_line`): one fork per visible row per frame.
- **Preview pane** (`render_preview`): one fork per frame in Terminal view, two in Tool view, for the selected session. The preview capture itself is worker-driven and otherwise fork free in steady state, so this was the last fork in a steady state frame.

Measured over 30 rows, before and after:

| View | Before | After |
|---|---|---|
| Structured (agents) | 46.6us/frame | 46.5us/frame |
| Terminal | 52.2ms/frame | 89.2us/frame |
| Tool | ~104ms/frame | 110.2us/frame |

1.74ms/row, matching a 1.86ms `has-session` fork on the same box. Live view targets a 25ms cadence, so the list render alone capped it near 19fps and every keystroke queued behind it.

**Fix.** Add `session_exists_for_display` and `pane_dead_for_display`, both snapshot backed, and seed both render paths from them. `session_exists` keeps its live fallback; teardown and drift decisions still must not act on a cached false negative. `PANE_META_CACHE` mirrors `SESSION_CACHE`, including the rule that a fresh snapshot carrying no data is an answer rather than a stale one, which is what bounds a tmux outage to one fork per TTL instead of one per row per frame.

Trade: a pane created behind this process's back reads idle for up to the 2s TTL. Every start and kill path already calls `refresh_session_cache()`, so anything done through aoe flips the glyph immediately.

**Refactor.** The three view arms of `render_item_line` each carried their own copy of the archive/snooze/urgent/favorite block. Structured and Terminal had byte identical title blocks; the Tool arm had none, so an archived or snoozed session there kept painting its running glyph with no `z ` prefix. `decorate_row` owns it for all three now, with a `SunkRow` enum naming the one policy that legitimately differs: Error and Deleting punch through the sink mask in Structured, where the seed carries `ICON_ERROR` and `theme.error`, but not in the pane views, whose seeds describe terminal liveness and carry no error affordance.

**Behavior change:** Tool view now honors archive, trash, and snooze. That is the only visible difference; everything else renders as before.

**Regression origin:** #2884 centralized existence checks on the authoritative helper to fix a real pane leak, where `kill()` silently no-oped on a stale cached miss. Before it, `TerminalSession::exists()` trusted a cached negative and forked only when the snapshot was stale. That change was right for teardown. The render path was an unnoticed caller.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No screenshot: the only visible change is the Tool view decoration noted above, which the new test covers. The user facing change is frame time.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: both behaviors are pinned at the unit tier, which can actually fail on them, and each assertion was mutation verified rather than just observed green. `display_liveness_answers_from_the_snapshot_instead_of_probing_per_name` forces a snapshot that disagrees with a live pane, so it fails the moment the display path probes per name again. `a_failed_pane_snapshot_is_an_answer_so_rows_do_not_re_fork` pins the invariant that bounds a tmux outage. `every_view_mode_paints_the_same_sunk_row_decoration` tables 3 view modes against the sunk states, seeding the pane views live so the assertions are not satisfied by the seed alone (`ICON_IDLE` and `ICON_STOPPED` are the same glyph), plus an anti vacuity check. An e2e test would cost ~2s on the serial critical path to assert glyphs a unit test already covers, and could not observe fork counts at all.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Root cause, measurements, and patch produced by Claude Opus 5 in back and forth with @njbrake. The before/after numbers come from a throwaway timing test run against both trees, not an estimate. The decision to unify the row arms, and the call to keep the delete-op exception Structured only, are his.

## Summary

- Added cached session and pane liveness checks for Terminal and Tool rows.
- Reduced repeated tmux queries during row rendering.
- Preserved live checks for teardown and state-drift decisions.
- Centralized row decoration for archive, snooze, urgent, and favorite states.
- Applied consistent sunk-row styling across view modes.
- Added tests for cached liveness, failed snapshots, live-row rendering, and view-specific decoration.

## Benefits

- Improves Terminal and Tool view performance.
- Keeps row appearance consistent across view modes.
- Handles unavailable snapshot data safely.
- Preserves authoritative checks when live state matters.

## Further enhancement

- Add end-to-end coverage for rendering performance and tmux state changes.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -252,6 +252,19 @@ struct SessionCache {
     data: Option<HashMap<String, i64>>,
     time: Option<Instant>,
 }
+
+/// Shared `tmux list-panes -a` snapshot behind [`pane_dead_for_display`],
+/// mirroring [`SESSION_CACHE`]'s TTL and its `data: None` "the server could
+/// not answer" state.
+static PANE_META_CACHE: RwLock<PaneMetaCache> = RwLock::new(PaneMetaCache {
+    data: None,
+    time: None,
+});
+
+struct PaneMetaCache {
+    data: Option<HashMap<String, PaneMetadata>>,
+    time: Option<Instant>,
+}
 const TMUX_COMMAND_TIMEOUT: Duration = Duration::from_secs(10);
 
 fn run_tmux_command_with_timeout_inner(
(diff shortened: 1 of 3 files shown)


src/tmux/mod.rs
```diff
@@ -1104,6 +1156,79 @@ pub fn session_exists(name: &str) -> bool {
...
+/// Returns `false` ("not known to be dead") for a session missing from the
+/// snapshot and whenever the snapshot could not be produced, matching
+/// [`batch_pane_metadata`]'s contract that an `Err` means "don't know" rather
+/// than "everything is dead". Callers gate on existence first, so a missing
+/// key is an absent session rather than a live pane.
+pub fn pane_dead_for_display(name: &str) -> bool {
+    if let Some(dead) = pane_dead_from_cache(name) {
+        return dead;
+    }
+    refresh_pane_meta_cache();
```
jerome-benoit, line 1198: [P1] This helper still refreshes `PANE_META_CACHE` synchronously when the snapshot expires. `refresh_pane_meta_cache()` then calls `batch_pane_metadata()`, whose `tmux list-panes -a` uses `Command::output()` without an application-level timeout. The session path has the same issue with a timeout of up to 10 seconds, and the Terminal/Tool wrappers constructed during paint also go through `live_session_name()`, which can refresh the session cache. This removes per-row forks while snapshots are fresh, but it still leaves tmux I/O on the render thread at startup and on cache expiry. Please publish both snapshots from the poller/background path, keep the entire display path read-only, and add an expired-cache test proving that rendering launches no command.

src/tui/home/render.rs
```diff
@@ -1645,85 +1682,58 @@ impl HomeView {
...
+                                },
+                                ICON_STOPPED,
+                            )
                         }
                         ViewMode::Tool(ref tool_name) => {
                             let tool_session =
                                 crate::tmux::ToolSession::new(&inst.id, &inst.title, tool_name);
-                            let tool_running =
-                                tool_session.exists() && !tool_session.is_pane_dead();
+                            let tool_running = crate::tmux::session_exists_for_display(
```
jerome-benoit, line 1716: [P2] `decorate_row` deliberately lets `Status::Error | Status::Deleting` bypass the archive/trash mask, but the Tool seed here, like the Terminal seed, derives its icon and color solely from pane liveness. A failed or in-progress deletion can therefore render as idle/running in those views instead of showing Error/Deleting. Please make these two statuses override pane liveness as they do in Structured view, and add them to the table covering all three modes.

jerome-benoit (changes requested):
Verdict: changes requested on exact head `4b56882d`.

Three independent review passes cross-validated the two findings below.

- P1: the Terminal/Tool paint path is still not cache-only. The display helpers synchronously refresh expired snapshots, and the session wrappers constructed during rendering can also refresh through `live_session_name`. Reviewers differed between P1 and P2; I am keeping P1 because `list-panes` has no application-level timeout on the render thread, so a stuck tmux process can freeze the TUI indefinitely.
- P2: the shared decorator's `Error`/`Deleting` exception is ineffective in Terminal and Tool views because their seeds encode only pane liveness. One reviewer rated this P3; I am keeping P2 because it hides the exact failed/in-progress deletion state this refactor claims to harmonize across views.

I am not duplicating CodeRabbit's still-open thread about restoring the process-global tmux caches after the new test; that thread should also be addressed.

Checked and rejected: the rustdoc failure is fixed and the Docs job is green. The targeted tests `every_view_mode_paints_the_same_sunk_row_decoration` and `a_failed_pane_snapshot_is_an_answer_so_rows_do_not_re_fork` pass. They validate the added fixes, but do not prove that cache renewal performs no I/O on the paint path. `git diff --check` is clean.

PR #3522, opened 2026-08-25
fix: keep new-session lifecycle badge visible

jerome-benoit:
## Description

Follow-up to #3495. The New Session dialog appended configuration metadata before the lifecycle badge, so a configured Gemini row clipped `⚠ deprecated` to `⚠ dep` on terminals narrower than roughly 76 columns.

This change centralizes lifecycle and configuration suffix ordering for both New Session and Restart. Lifecycle status now renders first, and the focused configuration hint uses the existing compact form whenever a lifecycle badge is present.

64-column TUI capture after the fix:

```text
│ Tool: ← ● gemini  [3/3]  → ⚠ deprecated  (configured) Ctrl+P │
```

Review finding: https://github.com/agent-of-empires/agent-of-empires/pull/3495#pullrequestreview-5021984424

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the deterministic Ratatui TestBackend contract exercises the exact 64-column configured row and failed before the fix. The actual TUI was also launched in isolated tmux at 64 columns and verified with the Tool row focused.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** openai-codex/gpt-5.6-sol

The agent reproduced the clipping, implemented the shared suffix ordering, ran the regression and full Rust suites, and drove the actual TUI in isolated tmux.

## Summary

- Centralized lifecycle and configuration suffix rendering for New Session and Restart dialogs.
- Rendered lifecycle badges before configuration hints to prevent clipping in narrow terminals.
- Added compact configuration hints when a lifecycle badge is present.
- Added regression coverage for narrow, configured rows and expanded deprecated badge tests.

## Benefits

- Prevents the `⚠ deprecated` badge from being clipped.
- Keeps suffix behavior consistent across dialogs.
- Preserves useful configuration hints in constrained layouts.

### DIFF
--- a/src/tui/components/mod.rs
+++ b/src/tui/components/mod.rs
@@ -14,7 +14,7 @@ mod text;
 mod text_input;
 mod tool_config;
 
-pub use cycler::{profile_cycler_spans, tool_cycler_spans, tool_lifecycle_spans};
+pub use cycler::{profile_cycler_spans, tool_cycler_spans};
 pub use dir_picker::{DirPicker, DirPickerResult};
 pub use help::HelpOverlay;
 pub use list_picker::{ListPicker, ListPickerResult};
@@ -26,5 +26,5 @@ pub use text_input::{
     set_input_cursor_position, set_prefixed_input_cursor_position, GroupGhostCompletion,
 };
 pub use tool_config::{
-    handle_tool_config_key, render_tool_config_overlay, tool_config_suffix_spans, ToolConfigOutcome,
+    handle_tool_config_key, render_tool_config_overlay, tool_row_suffix_spans, ToolConfigOutcome,
 };
--- a/src/tui/components/tool_config.rs
+++ b/src/tui/components/tool_config.rs
@@ -12,7 +12,7 @@ use ratatui::widgets::*;
 use tui_input::backend::crossterm::EventHandler;
 use tui_input::Input;
 
-use super::render_text_field;
+use super::{cycler::tool_lifecycle_spans, render_text_field};
 use crate::tui::styles::Theme;
 
 /// Tool-config overlay fields: command override, then extra args.
@@ -20,12 +20,28 @@ pub const TOOL_CONFIG_CMD: usize = 0;
 pub const TOOL_CONFIG_ARGS: usize = 1;
 const TOOL_CONFIG_FIELD_COUNT: usize = 2;
 
-/// Trailing spans appended to the Tool row: a dimmed (configured) marker
-/// when an override is set, plus the Ctrl+P hint while the row is focused.
-/// has_config is true when either the command override or extra args is
-/// non-empty. compact shortens the focused configured hint so a higher-
-/// priority status badge can share the fixed-width row without clipping.
-pub fn tool_config_suffix_spans(
+/// Trailing lifecycle and configuration spans appended to the Tool row.
+/// Lifecycle status comes first so lower-priority configuration metadata cannot
+/// clip it. The focused configuration hint is compact when a status is present.
+pub fn tool_row_suffix_spans(
+    tool: &str,
+    has_config: bool,
+    focused: bool,
+    theme: &Theme,
+) -> Vec<Span<'static>> {
+    let mut spans = tool_lifecycle_spans(tool, theme);
+    let compact_config = !spans.is_empty();
+    spans.extend(tool_config_suffix_spans(
+        has_config,
+        focused,
+        compact_config,
+        theme,
+    ));
+    spans
+}
+
+/// Configuration spans appended after higher-priority Tool row status.
+fn tool_config_suffix_spans(
     has_config: bool,
     focused: bool,
     compact: bool,
--- a/src/tui/dialogs/new_session/render.rs
+++ b/src/tui/dialogs/new_session/render.rs
@@ -10,7 +10,7 @@ use super::{NewSessionDialog, FIELD_HELP, HELP_DIALOG_WIDTH};
 use crate::tui::components::{
     focused_input_spans, input_scroll, profile_cycler_spans, render_text_field,
     render_text_field_with_ghost, render_tool_config_overlay, set_prefixed_input_cursor_position,
-    tool_config_suffix_spans, tool_cycler_spans, tool_lifecycle_spans, visible_slice,
+    tool_cycler_spans, tool_row_suffix_spans, visible_slice,
 };
 use crate::tui::styles::Theme;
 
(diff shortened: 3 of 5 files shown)


njbrake (commented):

_Reviewed at `88b23e95`. Push new commits and it will be re-reviewed._

---

Reviewed at head `88b23e95`, rebased onto `origin/main` (`95a18227`, with #3495 landed as `bcf4a578`). This is a direct follow-up to a finding raised on #3495, so the same measurement harness was re-run rather than reading the diff and inferring.

**It does exactly what it says, and it took the durable form rather than the local one. Recommend merging.**

### The fix, measured

The width-sweep probe was rebuilt at this head and run against **both** dialogs, with the same configuration that triggered the bug (3 tools, gemini selected, Tool row focused, `command_override` plus `extra_args` set):

| terminal width | New, before (#3495) | New, after | Restart, after |
|---|---|---|---|
| 40 | clipped | clipped | clipped |
| 50 | | **intact** | intact |
| 56 | | **intact** | intact |
| 64 | **clipped** | **intact** | intact |
| 70 | **clipped** | **intact** | intact |
| 76 | intact | intact | intact |
| 80 | intact | intact | intact |
| 100 | intact | intact | intact |

The two clipping widths reported in the original finding are fixed, and the fix reaches further down than the finding asked for: the badge now survives to 50 columns, not just 76.

**Restart did not regress.** Its call site shrank by 10 lines because it delegates, and the delegation is faithful: at every width in the sweep the two dialogs now emit a byte-identical Tool row. At 64 columns both produce

```
Tool: ← ● gemini  [3/3]  → ⚠ deprecated  (configured) Ctrl+P
```

which matches the capture in the PR body character for character.

On the 40-column row: both dialogs clip there, and that is a width floor, not a regression. `Tool: ← ● gemini  [3/3]  → ⚠ deprecated` is itself about 40 columns, so nothing fits. The relevant improvement is that the badge is now the *last* element to be lost instead of the first, which is the correct priority ordering. Not a finding.

### The centralization is real, and drift is structurally impossible

This was the substance of the original suggestion, so it was checked rather than taking the "centralizes" claim at face value. Three things landed together, and it is the combination that matters:

1. `tool_row_suffix_spans(tool, has_config, focused, theme)` in `src/tui/components/tool_config.rs:26` is the single public entry point. It computes the lifecycle spans first, derives `compact_config = !spans.is_empty()`, and appends the config suffix after.
2. `tool_config_suffix_spans` was demoted from `pub fn` to a module-private `fn` (`tool_config.rs:44`).
3. `tool_lifecycle_spans` was **removed from the `pub use cycler::{...}` list** in `src/tui/components/mod.rs:17`, and `mod cycler;` is a private module.

Point 3 is the load-bearing one. Because `cycler` is private and the symbol is no longer re-exported, `tool_lifecycle_spans` is simply unreachable from `src/tui/dialogs/**`. Neither dialog can reconstruct the wrong ordering even if someone wanted to; there is no path to the two halves. That is a compiler-enforced guarantee, not two call sites that happen to agree today.

Confirmed by grep that nothing else in the tree consumes any of the three symbols. The only callers are the two dialogs (via `tool_row_suffix_spans`) and the in-module tests, so the signature change reaches nothing else.

### The bool-parameter smell got better, not worse

The prior review flagged that `tool_config_suffix_spans` had grown to three consecutive `bool`s. The new public signature is:

```rust
pub fn tool_row_suffix_spans(tool: &str, has_config: bool, focused: bool, theme: &Theme)
```

Two bools, separated at the front by `&str`, and the third (`compact`) is now *derived* inside the helper rather than passed by callers. So the public surface went from three bools to two and lost the parameter most likely to be passed wrong. No fourth bool was added. This is the right direction.

### Tests

`src/tui/dialogs/new_session/tests.rs` shows +43/-21, but the **net `#[test]` fn delta is exactly zero**: the old `test_deprecated_tool_badge_renders_on_single_tool_read_only_row` was rewritten in place as `test_deprecated_tool_badge_remains_visible_on_tool_rows`, and the new 64-column constrained case is a **table row inside that one fn**, not a second fn. That is the house rule followed precisely. Combined with #3495 dropping `test_gemini_deprecated_fields`, the feature's total test-fn count is now one *below* where #3495 started.

The tests are load-bearing, verified by reverting the ordering inside the shared helper back to its pre-fix shape and re-running. Both dialogs' tests fail together:

```
tui::dialogs::restart::tests::deprecated_tool_badge_renders_on_constrained_configured_restart_row ... FAILED
tui::dialogs::new_session::tests::test_deprecated_tool_badge_remains_visible_on_tool_rows ... FAILED
```

with the exact pre-fix rendering in the panic output: `Tool: ← ● gemini  [3/3]  →  (configured)  Ctrl+P: edit ⚠ dep`. One edit to one function fails two dialogs' regression tests, which is itself the clearest evidence the consolidation is genuine.

The second case also asserts `(configured) Ctrl+P` is present, so it pins the compact form rather than only the badge. Before this PR the New dialog had no test that could catch this at all.

### Definition of done

- `cargo fmt --all -- --check`: clean.
- `cargo clippy --all-targets --features serve`: 5 warnings, all in pre-existing files (`acp/background_agent.rs`, `acp/supervisor.rs`, `process/macos.rs`, `session/smart_rename.rs`). None in files this PR touches.
- `cargo test --features serve --lib`: 6410 passed, 2 failed, exactly the two known macOS-environmental failures (`worktree_move_observation_canonicalizes_symlinked_parents`, `test_content_shows_settings_path`). Nothing else.
- No emdashes in the added prose.
- GitHub CI: 27 checks pass, 1 skipping, none failing.

**On the `BLOCKED` state:** it has since cleared to `MERGEABLE / CLEAN`. It was transient pending CI; `reviewDecision` is still empty, so the only thing outstanding is an approving review.

### Nits (neither worth blocking on)

- `tool_lifecycle_spans` is still declared `pub fn` in `cycler.rs`. It is already unreachable outside `components` because the module is private, so this is correct as written; `pub(super)` would just make the intent legible at the declaration.
- In the consolidated test, the cases are built as an array and then mutated by index (`cases[1].1.tool_index = 2;`), and the compact-hint assertion is dispatched on a string compare (`if *name == "constrained configured"`). The sibling test in `restart.rs` uses a cleaner `let cases = [(tool, expected)]` tuple form; carrying an `expect_compact: bool` in the tuple would match it. Purely cosmetic.

### Merge risk

**Low.** Pure presentation change to one span-assembly function, no persisted state, reverts cleanly. Both changed render paths now have regression tests confirmed to fail without the fix, and the tightened visibility means a future contributor cannot reintroduce the ordering bug without deliberately editing the shared helper, where both tests are watching.

### Verdict

**Design alignment:** Aligned; replaces two parallel span-assembly sites with one, and closes the door behind it by making the pieces private
**Implementation:** Approve
**Project fit:** Accept; small, scoped, fixes a real narrow-terminal defect on a documented mobile-SSH use case
**Documentation:** Proportionate (comment only, and the stale "the New dialog appends its own config summary" comment was correctly updated)
**Merge risk:** Low; none identified beyond the two cosmetic nits above

**LGTM** — it fixes the clipping at every width measured, does not regress Restart, and takes the shared-helper form that makes the two dialogs structurally incapable of drifting again, with regression tests verified to be load-bearing.

PR #3525, opened 2026-08-25
test(acp): serialize serve-gated registry writers

jerome-benoit:
## Description

Follow-up to #3510 after its final review found a serve-feature gap in the registry exclusion audit.

Six #[tokio::test] functions in src/acp/acp_client.rs call the structured-view spawn path with no source profile. Host environment resolution then resolves the configured default profile and calls install_from_config, which writes both process-global status registries. Those tests did not share serial_test's default key, so they could interleave with a ProfileRegistryGuard::take("default") snapshot or restore.

Validation used an isolated XDG config with default_profile = "default" plus temporary writer instrumentation. The five reset tests and spawn_with_nonexistent_command_errors_cleanly each logged a write to default. With the workstation's normal default_profile = "beta", the same tests wrote beta, confirming the key follows process config rather than a dedicated test profile.

The fix adds #[serial_test::serial] to exactly those six tests. It also removes #3510's now-obsolete documentation note naming them as the known unexcluded cohort. No production code changes.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording, N/A because there is no UI change

## Test Coverage Analysis

- [ ] Added or updated a Playwright spec under web/tests/ and updated web/tests/coverage-matrix.json

**TUI / CLI (e2e test)**
- [ ] Added or updated an e2e test under tests/e2e/

## Validation

- cargo test --lib --features serve acp::acp_client::tests::reset_ -- --test-threads=8: 6 passed
- cargo test --lib --features serve acp::acp_client::tests::spawn_with_nonexistent_command_errors_cleanly -- --test-threads=8: 1 passed
- cargo test --lib --features serve tmux::status_rules::tests:: -- --test-threads=8: 12 passed
- cargo test --features serve: 6695 passed across 24 suites, 8 ignored
- cargo fmt --all -- --check: clean
- cargo clippy --features serve -- -D warnings: clean

The additional local cargo clippy --lib --tests --features serve -- -D warnings probe reaches test-only code and reports five current-main warnings outside this diff. The standard CI clippy invocation above is clean.

No new test function is warranted: the change corrects execution exclusion on six existing behavior tests, and the registry restoration contract remains covered by #3510's sensitivity-tested probes.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** OpenAI Codex coding harness

Review findings were treated as untrusted input, traced through inherited_host_env and resolve_config_or_warn, and checked with isolated-config writer instrumentation before correction.

**NOTE:**

## Summary

- Serialized six ACP tests that modify process-global default-profile registries.
- Removed obsolete documentation about a registry race condition.
- Improved test isolation without changing production behavior.

## Validation

- Ran targeted tests, the full `serve` feature test suite, formatting, and Clippy checks.

### DIFF
--- a/src/acp/acp_client.rs
+++ b/src/acp/acp_client.rs
@@ -12316,6 +12316,7 @@ done
     /// old-session updates to the fresh conversation.
     #[cfg(unix)]
     #[tokio::test]
+    #[serial_test::serial]
     async fn reset_between_prompts_with_open_tool_is_refused() {
         let tmp = tempfile::TempDir::new().expect("tempdir");
         let initial_update = serde_json::json!({
@@ -12357,6 +12358,7 @@ done
     /// tailer removes the agent from the between-prompt in-flight set.
     #[cfg(unix)]
     #[tokio::test]
+    #[serial_test::serial]
     async fn reset_between_prompts_with_background_agent_is_refused() {
         let tmp = tempfile::TempDir::new().expect("tempdir");
         let transcript = tmp.path().join("background-agent.jsonl");
@@ -12410,6 +12412,7 @@ done
     /// abandoned request.
     #[cfg(unix)]
     #[tokio::test]
+    #[serial_test::serial]
     async fn reset_session_new_timeout_releases_the_connection_loop() {
         let tmp = tempfile::TempDir::new().expect("tempdir");
         let (script, _capture) = write_reset_fake_agent(tmp.path(), 0, 1, 0);
@@ -12455,6 +12458,7 @@ done
     /// client and runner would disagree about the live session.
     #[cfg(unix)]
     #[tokio::test]
+    #[serial_test::serial]
     async fn reset_config_timeout_commits_and_releases_the_connection_loop() {
         let tmp = tempfile::TempDir::new().expect("tempdir");
         let (script, capture) = write_reset_fake_agent(tmp.path(), 0, 0, 1);
@@ -12642,6 +12646,7 @@ done
     /// The success-only `SessionCleared` boundary must remain absent.
     #[cfg(unix)]
     #[tokio::test]
+    #[serial_test::serial]
     async fn reset_during_in_flight_prompt_is_refused_with_prompt_rejected() {
         let tmp = tempfile::TempDir::new().expect("tempdir");
         // 3s prompt delay: long enough to land the reset mid-turn, short
@@ -12844,6 +12849,7 @@ done
     }
 
     #[tokio::test]
+    #[serial_test::serial]
     async fn spawn_with_nonexistent_command_errors_cleanly() {
         let config = SpawnConfig {
             wrapper_substitution: None,
--- a/src/tmux/status_rules.rs
+++ b/src/tmux/status_rules.rs
@@ -277,11 +277,7 @@ impl ProfileRegistryGuard {
     /// between the phases can leave a half-swapped pair standing. Guarding a
     /// profile nothing else writes therefore needs nothing more, since the
     /// keys never overlap; guarding a profile other tests write needs those
-    /// writers held off, and a config resolve counts as a write. One known
-    /// case that is not held off: the serve-gated `#[tokio::test]` functions
-    /// in `src/acp/acp_client.rs` reach this registry through
-    /// `resolved_acp_config` and write `default` outside serial_test's default
-    /// key, so a `take("default")` can be raced under `--features serve`.
+    /// writers held off, and a config resolve counts as a write.
     /// Audit the writers of the profile you guard rather than assuming a group
     /// already covers them. The guard covers exactly these two registries and
     /// only its caller's writes; a future process-global needs its own restore


njbrake (commented):

_Reviewed at `c9841006`. Push new commits and it will be re-reviewed._

---

## Summary

Adds `#[serial_test::serial]` to the six `#[tokio::test]` functions in `src/acp/acp_client.rs` that write a status-registry profile outside serial_test's default key, and removes the now-false example sentence from `ProfileRegistryGuard::take`'s doc in `src/tmux/status_rules.rs`. Stacks on #3510 (`a0137269`). +7 / -5, no production code.

## The claim that failed last time now holds

The `install_from_config` write log was rebuilt at this head, rebased on `a0137269`, and the full lib suite run in both configurations. Not reasoned from the diff, because the whole reason this finding existed is that a completeness claim inherited a blind spot from its evidence.

| Configuration | Tests | Writers of `default` | Non-default-key writers |
| --- | --- | --- | --- |
| `cargo test --lib` | 4693 passed, 2 known-environmental failures | 72 | **0** |
| `cargo test --lib --features serve` | 6413 passed, same 2 | 105 | **0** |

Extending the audit to every profile a `ProfileRegistryGuard::take` is ever called on, under `--features serve`:

```
default                     writers= 105  ALL DEFAULT-KEY
test                        writers= 569  ALL DEFAULT-KEY
other                       writers=   3  ALL DEFAULT-KEY
sandbox-wrapped-codex ..    writers=   1  SOLE WRITER (disjoint)   [9 profiles]
INVARIANT HOLDS
```

Four profiles still carry mixed exclusion domains (`guards-owned`, `main`, `personal`, `work`), and none of them is ever guarded. That was checked rather than assumed: all 25 `take()` call sites pass a non-empty string literal or a non-empty const, so no guard ever resolves an empty profile and none can land on `main`. Writer-versus-writer collisions on an unguarded profile tear no snapshot and nobody asserts on those registries, so they are not exposure.

## The reproduction no longer reproduces

Same setup that failed at the previous head: widen `contains_is_case_insensitive_and_regex_is_as_written` by 2.5 s between install and assert, run alongside the ACP reset tests at `--test-threads=8` with `--features serve`.

```
test tmux::status_rules::tests::contains_is_case_insensitive_and_regex_is_as_written ... ok
test result: ok. 7 passed; 0 failed; ... finished in 8.24s
```

Previously: `left: None, right: Some(Running)`, finished in 3.17 s. The run also got slower in exactly the way serialization predicts, which is worth noting as evidence in itself: it proves the attribute is *effective*, not merely present. An inert `#[serial]` would have left the timing alone and the assertion red.

## The doc comment, read in full

Five lines came out, and what remains is:

```
/// The aliases and rules phases each lock their own map, so this guard is
/// rollback, not mutual exclusion: a writer touching the same profile
/// between the phases can leave a half-swapped pair standing. Guarding a
/// profile nothing else writes therefore needs nothing more, since the
/// keys never overlap; guarding a profile other tests write needs those
/// writers held off, and a config resolve counts as a write.
/// Audit the writers of the profile you guard rather than assuming a group
/// already covers them. ...
```

**No completeness claim, in any form.** Every remaining statement is either a mechanical property of the code or a conditional; there is no census of who writes what, so there is nothing to go stale. The sentence that matters most, "Audit the writers of the profile you guard rather than assuming a group already covers them", survives the trim, and that is precisely the instruction that prevents a third recurrence.

The trim was also *required*, not optional. The removed sentence said those six run outside the default key. After this PR that is false, so leaving it would have made the comment wrong for a third time, just in the opposite direction. Removing a now-false example while keeping the method is the right call.

One cosmetic note only: the paragraph is not reflowed after the deletion, so "counts as a write." ends a short line. rustfmt does not touch doc comments and nothing depends on it.

## The profile is not fixed, and that makes `#[serial]` the better fix

The body claims the registry key follows process config rather than a dedicated test profile. Verified directly, and it is true. The writes come from `inherited_host_env_pairs` (`src/acp/acp_client.rs:3931`), which passes `config.source_profile.as_deref().unwrap_or_default()`, so an unset `source_profile` becomes `""` and `effective_profile("")` resolves through `resolve_default_profile()` (`src/session/config.rs:3154`).

Instrumented, one test, isolated `$HOME` with the debug-build app dir:

| `default_profile` | Profile actually written |
| --- | --- |
| `beta` | `beta` |
| `default` | `default` |
| (absent, virgin home) | `main`, via `ensure_bootstrap_profile()` |

And in the full-suite logs the `acp_client` module wrote **both** `default` and `main` within a single run, depending on which `$HOME` a previously-run test had leaked at that moment.

This changes one conclusion from the earlier rounds, in the PR's favour. Any statement of the form "profile X's writers all share one domain" is only valid for the environment the audit ran in, for any writer that resolves an empty profile. `#[serial]` is immune to that: it excludes these tests from the default-key group whatever profile they land on. The scoping fix recommended in the previous round would have hard-coded a profile name that is not stable across machines, so this is the better fix, not merely the more expensive one.

## A correction to the prior review

The #3510 doc sentence written in the last round named `resolved_acp_config` as the path these tests take into the registry. That was wrong. With `source_profile: None` that function takes its `load_config()` branch (`src/acp/acp_client.rs:1875`), which does not install. The real writer is `inherited_host_env` (`src/session/environment.rs:121`). The author's PR body attributes it correctly, and this PR's deletion incidentally removes that error along with the stale example.

## Cost, measured

The previous round declined to recommend this fix because the six measured 5.76 s single-threaded and the tradeoff was the maintainer's call. Here is what it actually costs, serve lib suite, three runs each:

| | Runs | Median |
| --- | --- | --- |
| With the six `#[serial]` (as shipped) | 79.77 s, 80.34 s, 80.46 s | 80.34 s |
| Without them (attributes stripped) | 74.89 s, 75.15 s, 77.90 s | 75.15 s |

**About +4 to +5 s, roughly 5.5% of the serve lib suite.** The groups do not overlap (slowest baseline run is faster than the fastest serialized run), so unlike the 37 dialog tests in #3510 this is a real, visible cost rather than something lost in run-to-run noise. Against the 7m21s `Cargo Test (linux, serve-rest)` CI job it is about 1%. Reported because the number was promised, not as an objection: buying a reproducible race for 1% of one CI job is a good trade.

## Definition of done

| Check | Result |
| --- | --- |
| `cargo clippy --lib --tests -- -D warnings` | only `clippy::items_after_test_module` at `src/process/macos.rs:266`, pre-existing |
| `cargo clippy --lib --tests --features serve -- -D warnings` | exactly 5 errors, all pre-existing and outside the diff: `src/acp/background_agent.rs:627` and `:745`, `src/acp/supervisor.rs:4030`, `src/process/macos.rs:266`, `src/session/smart_rename.rs:1546`. The body's claim of "five current-main warnings outside this diff" is accurate. |
| `cargo test --lib` | 4693 passed, 2 failed (the two known macOS-environmental) |
| `cargo test --lib --features serve` | 6413 passed, same 2 |
| Three sensitivity probes | All pass as shipped; the two guard-usage probes still fail when their guards are stripped; the contract test correctly passes either way |
| Prose rules | no emdashes in added lines |
| CI and bots | 28 checks green, `MERGEABLE` / `CLEAN`; CodeRabbit "no actionable comments"; Codecov green, all modified lines covered |

Nothing failed that is not on the known-environmental list.

## Design alignment

**Aligned.** `#[serial]` placed after the test attribute matches both this repo's convention (the same ordering #3510 corrected in `builder.rs`) and serial_test's documented form for async runtimes. The fully-qualified `#[serial_test::serial]` avoids adding an import to a file that had no `serial` usage, and `src/acp/` already mixes both spellings.

No new test functions, which is right and the body argues it correctly: the change corrects execution exclusion on six existing behavior tests, and a test asserting "these six are serialized" would be testing an attribute. The restoration contract stays covered by #3510's three probes, re-verified as still sensitive. Six attribute lines add no compile weight, so the AGENTS.md test-cost calculus is untouched. Conventional commit prefix, one commit, no production code.

## Merge risk

**Low.** Test-only, reverts cleanly, and the change is six attributes plus a deletion of prose that had become false. The residual is a 4 to 5 second wall-clock cost on the serve suite, which is a known and accepted tradeoff rather than a risk. Nothing here reaches a release build.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Proportionate; the trim was required and what remains is true in every configuration
**Merge risk:** Low; a measured ~5.5% wall-clock cost on the serve lib suite, already weighed

**LGTM** — fresh instrumentation in both feature configurations shows zero non-default-key writers of `default`, the reproduction no longer reproduces, and the surviving doc text makes no completeness claim while keeping the audit-the-writers instruction that prevents this from happening a third time.

PR #3526, opened 2026-08-25
fix(web): one-row cursor/mouse offset in split live terminal windows

### LINKED ISSUE #3515: Web live terminal: in a split window the cursor and forwarded mouse rows are one row too high, because pane 0's `cursor_y` is painted onto the composite without adding `pane_top`
# Me

Currently in Web live terminal whenever Claude starts a sub agent which splits the terminal to multiple panels, the cursor gets a one row offset. Everything needs to be clicked an highlighted one row higher than where the content actually is.

Image

The top arrow demonstrates the position where the mouse cursor has to be when highlighting that text. The bottom cursor points to the flashing input cursor which is one row too high.

# AI

**Version**: aoe dev build from source, `1.15.0` (`5b16d21a`), debug namespace (`aoe_dev_` prefix, port 8081)
**Server**: Linux Mint 22.3, tmux 3.4, agent Claude Code v2.1.245
**Client**: Chrome
**Mode**: `aoe serve` <local / tunnel / tailscale>

## Summary

When Claude Code spawns a subagent it splits the tmux window side by side, and from that moment the web dashboard terminal view is one row off vertically:

- the blinking cursor is drawn one row **above** the input line (it lands on the input box's top border),
- clicks and other mouse events have to be aimed one row **above** the thing I want to hit.

The rendered pane content itself is correct, only the cursor row and the pointer mapping are
shifted. **Unsplit the same view is perfectly correct**, cursor and clicks both, which is the
clean split between the two capture paths described below.

Root cause: the compositor lays the panes out on the **window** grid, but the cursor row it ships alongside is **pane 0 relative**, and nothing translates between the two. The two agree only while pane 0 sits at `pane_top == 0`. When the window has a pane border status row, pane 0 starts at `pane_top == 1`, so every pane row is painted one row lower than the cursor row that indexes it.

## Evidence from the live session

Probing the exact session from the screenshot (`aoe_dev_Test_web_split_pane__e2ecb795`) while the split was up:

```
$ tmux -S ~/.config/agent-of-empires-dev/tmux.sock display-message -p -t <sess>:^ \
    -F 'panes=#{window_panes} win_w=#{window_width} win_h=#{window_height}'
panes=2 win_w=235 win_h=72

$ tmux ... list-panes -t <sess>:^ -F 'pane#{pane_index}: left=#{pane_left} top=#{pane_top} w=#{pane_width} h=#{pane_height} curx=#{cursor_x} cury=#{cursor_y}'
pane0: left=0   top=1 w=164 h=71 curx=2 cury=63
pane1: left=165 top=1 w=70  h=71 curx=2 cury=66

$ tmux ... show-options -w -t <sess>:^ | grep pane-border
pane-border-status top
pane-border-style fg=blue
```

`window_height` (72) is `pane_height` (71) plus the one border status row, and both panes sit at `top=1`.

Pane 0's own capture, 1-based line numbers, so 0-based rows are one less:

```
63  ────────────────────────────────  <- input box top border   (0-based row 62)
64  ❯ give it a longer task so I ...  <- input line              (0-based row 63)
65  ────────────────────────────────
```

tmux reports `cursor_y = 63`, which is the input line. Correct, and pane relative.

The composite paints pane 0's row 63 at composite row `63 + pane_top` = **64**, while the client maps the cursor with `Math.max(0, lines.length - screenRows) + cursor.y` = `max(0, 72 - 72) + 63` = **63**, i.e. onto the input box's top border. That is exactly where the screenshot shows it.

The spurious full width horizontal rule above both panes in the screenshot is the same defect showing itself: composite row 0 is covered by no pane, so `gap_fill` draws a border rule across the whole window there.

## Why it only shows up after a subagent split

Two conditions have to hold at once, and the subagent split is what makes both true:

1. **`window_panes >= 2`**. Below that, `capture_window_composited_with_cursor` early returns
   pane 0's raw capture, where the rows and `cursor_y` are both pane relative and agree for any
   `pane_top`. Compositing is the only place a pane relative row is indexed into window relative
   content, which is why an unsplit window is never affected even with `pane-border-status` on.
2. **`pane_top != 0` for pane 0**, which here comes from `pane-border-status top`. `grep -rn pane-border-status .` finds nothing in this repo and it is not in my `~/.tmux.conf` nor in the tmux globals on that socket, so the splitter sets it as a window option; Claude Code presumably does it to label the panes. My unsplit session on the same socket has no `pane-border-status` and `pane_top=0`, and shows no offset.

So the general condition is: **any composited window whose pane 0 is not at the window origin**. `pane-border-status` is just the common way to get there.

## Reproduction

1. Open a session in the web dashboard terminal view, Claude Code running in it.
2. Ask Claude Code to spawn a subagent that splits the window to the side (the window ends up with 2 panes and `pane-border-status top` set).
3. Confirm the trigger: `tmux -S <socket> list-panes -t <sess>:^ -F '#{pane_index} top=#{pane_top}'` reports `top=1`.
4. Watch the blinking cursor in the terminal view: it sits on the input box's top border, one row above the input text.
5. Try to click anything in the pane: the click registers one row below where you clicked, so you have to aim one row high.
6. Negative control: any session whose window has a single pane renders the cursor exactly right
   in the same dashboard, and clicks land where I aim them. The offset appears and disappears
   with the split.

Step 5 is my own observation from using it; the code path that explains it is `pointerCell()` returning `visualRow - screenTopVisual + 1`, which for composite row `i` yields pane row `i + 1` while the true pane row is `i`.

## Where it goes wrong

- `src/tmux/session.rs`, `capture_window_composited_with_cursor()`: the rebase sets `c.pane_height = layout.window_height`, `c.pane_width = layout.window_width`, `c.history_size = 0` and records `composite_pane0 = (width, height)`, but leaves `c.x` / `c.y` in pane 0's coordinate space. The doc comment above it states the assumption that makes this safe ("tmux puts it at the window origin, so its coordinates index the composite untranslated"), and `pane-border-status` breaks it.
- `web/src/components/MobileLiveTerminal.tsx`: the `live` memo (`const lineIdx = Math.max(0, lines.length - screenRows) + cursor.y`) and `pointerCell()` share that mapping, which is why the drawn cursor and the forwarded mouse rows are off by the same single row.
- Likely the same in the TUI live preview, **not verified in this repro**: `src/tui/home/render.rs:144` consumes the same untranslated `cursor.y`, and `mouse_pane_rect()` in `src/tui/home/input.rs` documents the same false assumption ("Pane 0 sits at the window origin, so the sub-rectangle shares the preview's top-left corner and only its extent differs").

## Note for whoever fixes this

The frame's `pane0` payload (`src/server/live_ws.rs`, `frame_json`) carries only `{cols, rows}`, so pane 0's **origin** is not on the wire at all. The cursor row (composite space) and `pointerCell`'s reverse mapping (back into pane 0 space) therefore cannot both be corrected without carrying `pane_top` / `pane_left` through `PaneCursor::composite_pane0` and the frame.

Renormalising the composite grid onto the panes' bounding box instead looks tempting and would also drop the phantom border rule, but it only fixes the topmost pane: with `pane-border-status` on a stacked split every pane gets its own border row, so the lower pane's `pane_top` still is not its offset within a bounding box. Translating by pane 0's actual `pane_top` handles every layout.

jerome-benoit:
## Description

In the web live terminal, any composited (split) window whose pane 0 is not at the window origin rendered pane 0's cursor one row too high and mapped forwarded mouse rows one row low. The trigger in the wild is Claude Code's subagent split, which sets `pane-border-status top` and pushes every pane down to `pane_top == 1` (closes #3515).

Root cause: the compositor lays panes out on the window grid, but `capture_window_composited_with_cursor` shipped the cursor still in pane 0's coordinate space, and the frame's `pane0` payload carried only `{cols, rows}`, so neither the server nor the client had pane 0's origin to translate with.

Fix: transport pane 0's full rectangle (`left`, `top`, `width`, `height`) through `PaneCursor::composite_pane0`, emit it on the wire as `pane0.{left,top}` next to the existing `cols`/`rows`, translate the cursor onto the window grid once at emission (`frame_json`), and subtract it again in the client's inverse pointer projection. The client projection is extracted into pure functions (`cursorLineIndex`, `pointerPaneCell`) in `web/src/lib/liveMouse.ts`.

Protocol compatibility: both new fields are optional additions; frames without them (single-pane windows) are byte-identical to before. A client that ignores `left`/`top` still paints the cursor correctly at any origin, because translation happens at emission; its mouse forwarding is what degrades to the one-row shift whenever pane 0's origin is not `(0, 0)`.

Known residual (documented in code, not hacked around): the border-status row itself is covered by no pane, so it is still drawn as a full-width gap-fill rule; dropping it would render a grid shorter than `window_height` and rebase every consumer. The TUI preview shares this geometry and its mouse mapping still assumes a shared top-left corner; that side was never reproduced, so this PR only corrects its now-false doc comments.

## PR Type

- [x] Bug Fix

## Checklist

No screenshot: I could not stand up a real split session with `pane-border-status top` plus a dashboard browser against it in this environment; the behavioral guarantees rest on the real-tmux geometric test below plus the real-browser Playwright test.

## Test Coverage Analysis

Added a mocked-playwright case in `tests/live-click-forward.spec.ts`: clicking composite row 1 with `pane0.top = 1` must forward SGR row 1, not 2. The touched flows (`terminal.live-split-composite`, `terminal.mobile-live-click-forward*`) were already registered in the coverage matrix, so no matrix entry changed; `node tests/validate-coverage-matrix.mjs` passes.

New tests:
- Rust, `frame_json_includes_geometry_and_cursor` (table): origin-carrying frame serializes cursor `{x+left, y+top}` and `pane0.{cols,rows,left,top}`; the origin-less case pins the unchanged unsplit wire shape.
- Rust, tmux-gated `composited_cursor_and_pane0_origin_track_pane_border_offset`: for horizontal, vertical, and stacked splits with `pane-border-status top`, `composite_pane0` matches tmux-reported geometry and the composite paints the pane's cursor row at `cursor.y + top` (with an anti-vacuity guard that the untranslated row does not contain the marker).
- Vitest: `cursorLineIndex` (including the `screenRows < lines.length` scrollback case) and `pointerPaneCell` (identity, shifted origin, clamping).
- Playwright mocked: the offset click case above.

Mutation checks (both reverted translations made the new tests fail, identity cases stayed green):
- Reverting the `frame_json` translation fails `frame_json_includes_geometry_and_cursor` (got x 3, expected 5).
- Reverting `pointerPaneCell`'s origin subtraction fails the new Vitest case and the new Playwright case in a real Chromium bundle rebuild.

**TUI / CLI (e2e test)**

The TUI consumes the same struct mechanically (type change), but no TUI behavior was deliberately altered and none was reproduced.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** Claude Code (Opus 5) produced the diagnosis in the issue; the fix, tests, and this PR were implemented by ox-alpha.

**Any Additional AI Details you'd like to share:** Verification ran `cargo fmt --all -- --check`, `cargo clippy --features serve -- -D warnings`, `cargo test --lib --features serve` (6404 passed), targeted tmux-gated tests, web `format:check`/`lint`/full Vitest (2 pre-existing locale-dependent failures in `Composer.usageHint.test.tsx` also failing on the base commit), `validate-coverage-matrix.mjs`, and the mocked Playwright suite file (7/7).

## Summary

- Fixes cursor and mouse-coordinate offsets in split live-terminal windows when pane 0 starts away from the window origin.
- Sends pane 0’s complete position and size from the server to the web client.
- Translates cursor and mouse coordinates between window space and pane space.
- Reuses `PaneGeom` across the Rust code and removes the older `Pane0Rect` type.
- Keeps unsplit frames and older metadata compatible.
- Adds Rust, Vitest, and Playwright coverage for offsets, clamping, and borderless splits.
- Preserves the full-width gap-fill row for `pane-border-status top`.
- Documents the remaining TUI preview offset behavior.

## Benefits

Users receive more accurate cursor positions and mouse input in split web live-terminal windows. The shared geometry type and tested coordinate helpers make the behavior easier to maintain and extend.

### DIFF
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -21,6 +21,13 @@
 //!   `#{mouse_sgr_flag}`: when the pane is a full-screen mouse app the
 //!   client forwards the wheel to it (as input bytes) instead of widening
 //!   the capture window, since the alternate screen has no scrollback.
+//!   On a composited (split) window the frame also carries `"pane0"`:
+//!   `{"cols":..,"rows":..,"left":..,"top":..}`, pane 0's rectangle within
+//!   the window grid. The cursor is translated onto that grid before it is
+//!   sent, and clients map pointer cells back by subtracting
+//!   `(left, top)` (#3515). Absent on single-pane frames; older clients
+//!   that ignore `left`/`top` render correctly whenever pane 0 sits at the
+//!   window origin.
 //!   `{"type":"size_owner","is_owner":bool}`: whether this client holds
 //!     the session's size-owner lock. Only the owner resizes the shared
 //!     tmux window and may type; a non-owner renders best-effort at the
@@ -1027,10 +1034,19 @@ async fn handle_live_ws(
 /// virtual scroll spacer off `history` and slices the live screen off
 /// the content's last `rows` lines, independent of cursor visibility.
 fn frame_json(content: &str, cursor: Option<&crate::tmux::PaneCursor>) -> String {
+    // On a composited frame the content grid is the whole WINDOW while
+    // `x`/`y` are pane 0 relative; the client paints the cursor by indexing
+    // that grid directly, so translate onto the composite here (#3515).
+    // `composite_pane0` is `None` (or at the origin) on every other frame,
+    // where this is the identity. The origin also rides on `pane0` so the
+    // client's inverse mapping, pointer cell -> pane coordinate, can
+    // subtract it again.
+    let pane0 = cursor.and_then(|c| c.composite_pane0);
+    let (origin_x, origin_y) = pane0.map_or((0, 0), |p| (p.left, p.top));
     let cursor_value = match cursor {
         Some(c) if c.visible => serde_json::json!({
-            "x": c.x,
-            "y": c.y,
+            "x": c.x.saturating_add(origin_x),
+            "y": c.y.saturating_add(origin_y),
         }),
         _ => serde_json::Value::Null,
     };
@@ -1046,8 +1062,13 @@ fn frame_json(content: &str, cursor: Option<&crate::tmux::PaneCursor>) -> String
         "altScreen": cursor.map(|c| c.alternate_on).unwrap_or(false),
         "mouse": cursor.map(|c| c.mouse_tracking).unwrap_or(false),
         "mouseSgr": cursor.map(|c| c.mouse_sgr).unwrap_or(false),
-        "pane0": cursor.and_then(|c| c.composite_pane0).map(|(cols, rows)| {
-            serde_json::json!({ "cols": cols, "rows": rows })
+        "pane0": pane0.map(|p| {
+            serde_json::json!({
+                "cols": p.width,
+                "rows": p.height,
+                "left": p.left,
+                "top": p.top,
+            })
         }),
     })
     .to_string()
@@ -1139,33 +1160,56 @@ mod tests {
 
     #[test]
     fn frame_json_includes_geometry_and_cursor() {
-        let cursor = crate::tmux::PaneCursor {
-            x: 3,
-            y: 7,
-            visible: true,
-            pane_height: 46,
-            history_size: 1200,
-            pane_width: 74,
-            alternate_on: false,
-            mouse_tracking: false,
-            mouse_sgr: false,
-            mouse_all: false,
-            position_reliable: true,
-            composite_pane0: Some((37, 46)),
-        };
-        let json = frame_json("hello\nworld", Some(&cursor));
-        let v: serde_json::Value = serde_json::from_str(&json).unwrap();
-        assert_eq!(v["type"], "frame");
-        assert_eq!(v["content"], "hello\nworld");
-        assert_eq!(v["rows"], 46);
-        assert_eq!(v["history"], 1200);
-        assert_eq!(v["cursor"]["x"], 3);
-        assert_eq!(v["cursor"]["y"], 7);
-        assert_eq!(v["altScreen"], false);
-        assert_eq!(v["mouse"], false);
-        assert_eq!(v["mouseSgr"], false);
-        assert_eq!(v["pane0"]["cols"], 37);
-        assert_eq!(v["pane0"]["rows"], 46);
+        let cases = [
+            // Unsplit (or pane 0 at the window origin): the translation is
+            // the identity and `pane0` is absent.
+            (None, (3, 7), serde_json::Value::Null),
+            // Composited with pane-border-status top: the wire cursor moves
+            // onto the window grid by pane 0's origin (#3515).
+            (
+                Some(crate::tmux::Pane0Rect {
+                    left: 2,
+                    top: 1,
+                    width: 37,
+                    height: 46,
+                }),
+                (5, 8),
+                serde_json::json!({
+                    "cols": 37,
+                    "rows": 46,
+                    "left": 2,
+                    "top": 1,
+                }),
+            ),
+        ];
+        for (pane0, want_cursor, want_pane0) in cases {
+            let cursor = crate::tmux::PaneCursor {
+                x: 3,
+                y: 7,
+                visible: true,
+                pane_height: 46,
+                history_size: 1200,
+                pane_width: 74,
+                alternate_on: false,
+                mouse_tracking: false,
+                mouse_sgr: false,
+                mouse_all: false,
+                position_reliable: true,
+                composite_pane0: pane0,
+            };
+            let json = frame_json("hello\nworld", Some(&cursor));
+            let v: serde_json::Value = serde_json::from_str(&json).unwrap();
+            assert_eq!(v["type"], "frame");
+            assert_eq!(v["content"], "hello\nworld");
+            assert_eq!(v["rows"], 46);
+            assert_eq!(v["history"], 1200);
+            assert_eq!(v["cursor"]["x"], want_cursor.0, "{pane0:?}");
+            assert_eq!(v["cursor"]["y"], want_cursor.1, "{pane0:?}");
+            assert_eq!(v["altScreen"], false);
+            assert_eq!(v["mouse"], false);
+            assert_eq!(v["mouseSgr"], false);
+            assert_eq!(v["pane0"], want_pane0, "{pane0:?}");
+        }
     }
 
     #[test]
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -25,9 +25,11 @@
 //!   `{"cols":..,"rows":..,"left":..,"top":..}`, pane 0's rectangle within
 //!   the window grid. The cursor is translated onto that grid before it is
 //!   sent, and clients map pointer cells back by subtracting
-//!   `(left, top)` (#3515). Absent on single-pane frames; older clients
-//!   that ignore `left`/`top` render correctly whenever pane 0 sits at the
-//!   window origin.
+//!   `(left, top)` (#3515). Absent on single-pane frames. Clients that
+//!   ignore `left`/`top` still paint the cursor correctly at any origin,
+//!   because the translation already happened here; their mouse forwarding
+//!   is what degrades, back to the one-row shift of #3515 whenever the
+//!   origin is not `(0, 0)`.
 //!   `{"type":"size_owner","is_owner":bool}`: whether this client holds
 //!     the session's size-owner lock. Only the owner resizes the shared
 //!     tmux window and may type; a non-owner renders best-effort at the
@@ -1167,7 +1169,7 @@ mod tests {
             // Composited with pane-border-status top: the wire cursor moves
             // onto the window grid by pane 0's origin (#3515).
             (
-                Some(crate::tmux::Pane0Rect {
+                Some(crate::tmux::PaneGeom {
                     left: 2,
                     top: 1,
                     width: 37,
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -1163,9 +1163,25 @@ mod tests {
     #[test]
     fn frame_json_includes_geometry_and_cursor() {
         let cases = [
-            // Unsplit (or pane 0 at the window origin): the translation is
-            // the identity and `pane0` is absent.
+            // Unsplit: `pane0` is absent and the cursor is untouched.
             (None, (3, 7), serde_json::Value::Null),
+            // Composited with pane 0 at the corner (a borderless split):
+            // identity translation, but `pane0` rides with zero origin.
+            (
+                Some(crate::tmux::PaneGeom {
+                    left: 0,
+                    top: 0,
+                    width: 37,
+                    height: 46,
+                }),
+                (3, 7),
+                serde_json::json!({
+                    "cols": 37,
+                    "rows": 46,
+                    "left": 0,
+                    "top": 0,
+                }),
+            ),
             // Composited with pane-border-status top: the wire cursor moves
             // onto the window grid by pane 0's origin (#3515).
             (
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -18,9 +18,12 @@
 //! `window_height` and rebasing every consumer, so the rule stays.
 
 /// One pane's rectangle within its window, from
-/// `#{pane_left} #{pane_top} #{pane_width} #{pane_height}`.
+/// `#{pane_left} #{pane_top} #{pane_width} #{pane_height}`. Pub because
+/// pane 0's copy rides on the public [`crate::tmux::PaneCursor::
+/// composite_pane0`] so composited frames can translate between pane
+/// coordinates and the window grid (#3515).
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) struct PaneGeom {
+pub struct PaneGeom {
     pub left: u16,
     pub top: u16,
     pub width: u16,
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -366,9 +366,10 @@ mod tests {
 
     #[test]
     fn first_pane_is_the_one_at_the_window_origin() {
-        // tmux orders pane indices by layout, so index 0 is the top-left pane.
-        // The live path relies on this to paint pane 0's cursor onto the
-        // composite without translating its coordinates.
+        // tmux orders pane indices by layout, so index 0 is the top-left
+        // pane. Consumers no longer assume that corner paints untranslated:
+        // they translate the cursor by the origin carried on
+        // `composite_pane0` (#3515).
         let l = layout(
             9,
             1,
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -10,6 +10,12 @@
 //!
 //! Compositing is read-only and changes nothing about input routing, which
 //! stays pinned to `^.0` (#435, #488).
+//!
+//! Known residual: a window row covered by NO pane (a `pane-border-status
+//! top` status line, which tmux draws itself and no capture sees) is filled
+//! with a full-width border rule. Translating the cursor (#3515) does not
+//! remove that row; dropping it would mean rendering a grid shorter than
+//! `window_height` and rebasing every consumer, so the rule stays.
 
 /// One pane's rectangle within its window, from
 /// `#{pane_left} #{pane_top} #{pane_width} #{pane_height}`.
@@ -87,11 +93,16 @@ impl WindowLayout {
         composite_window(self.window_width, self.window_height, &self.panes)
     }
 
-    /// The first pane's rectangle, which tmux guarantees sits at the window
-    /// origin: pane indices follow layout order, so index 0 is the top-left
-    /// pane, and closing it renumbers whichever pane takes that corner. That
-    /// is what lets a VT grid's cursor be painted onto the composite with no
-    /// coordinate translation.
+    /// The first pane's rectangle. Pane indices follow layout order, so
+    /// index 0 is the top-left pane and closing it renumbers whichever pane
+    /// takes that corner; the composite therefore paints pane 0's rows at
+    /// `geom.top` and its columns at `geom.left`, which is `(0, 0)` for the
+    /// common layouts but NOT always: a window option like
+    /// `pane-border-status top` reserves a border row above every pane,
+    /// shifting pane 0 down to `top == 1` (#3515). Consumers of the cursor
+    /// must translate by the origin carried on
+    /// [`crate::tmux::PaneCursor::composite_pane0`] rather than assuming the
+    /// window origin.
     pub(crate) fn first_pane(&self) -> Option<PaneGeom> {
         self.panes.first().map(|p| p.geom)
     }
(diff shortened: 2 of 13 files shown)


njbrake (commented):

_Reviewed at `7088cbb2`. Push new commits and it will be re-reviewed._

---

## Summary

On a composited (split) tmux window the live preview paints the whole **window** grid while pane 0's cursor stays **pane-relative**. The two agree only while `pane_top == 0`. Claude Code's subagent split sets `pane-border-status top`, which pushes pane 0 to `top == 1`, so the web dashboard drew the caret one row above the input line and forwarded clicks one row too far down (#3515).

The fix carries pane 0's full rectangle (`left`/`top`/`width`/`height`) on `PaneCursor::composite_pane0` instead of only its extent, emits `pane0.{left,top}` on the wire, translates the cursor onto the window grid once at emission (`frame_json`), and subtracts the origin again in the client's inverse pointer projection, extracted as pure `cursorLineIndex` / `pointerPaneCell` in `web/src/lib/liveMouse.ts`.

Net production code is roughly 20 lines; the rest is tests and comment corrections.

## Design alignment

**Aligned.**

Compared against the compositor itself, `composite_window` in `src/tmux/composite.rs`, which is the authority on where a pane lands:

```rust
if let Some(text) = pane.rows.get((row - pane.geom.top) as usize)
```

plus the column walk that emits a pane's whole row when `p.geom.left == col`.

So the compositor's own contract is "pane cell `(x, y)` paints at composite `(x + left, y + top)`", and the PR's `c.x.saturating_add(origin_x)` / `c.y.saturating_add(origin_y)` in `src/server/live_ws.rs:1049` is exactly that mapping, not an empirically tuned constant. This is the load-bearing fact and it was read from the source rather than taken from the description.

It extends the existing mechanism rather than adding a parallel one: `composite_pane0` already existed as the channel for "what the client needs to know about pane 0 that the rebase erased," and the PR widens its payload from `(u16, u16)` to the `PaneGeom` the layout already produces. Dropping the intermediate `Pane0Rect` (commit 2) in favour of the existing `PaneGeom` removes a duplicate type rather than introducing one. No new abstraction, no new layer, no second source of truth.

The one asymmetry worth naming: the cursor is translated **server-side** while the pointer is translated **client-side**. That is deliberate and argued in the wire doc (`src/server/live_ws.rs:24-30`): a pre-fix client that ignores `left`/`top` still paints the cursor correctly, because the translation already happened. The trade is right, and `cursorLineIndex`'s signature (no `pane0` parameter at all) is a decent structural guard against someone later "helpfully" subtracting the origin a second time on the client.

## Code review

| File | Assessment |
| --- | --- |
| `src/tmux/composite.rs` | `PaneGeom` `pub(crate)` to `pub`. Doc corrections only; the `first_pane()` doc previously asserted pane 0 sits at the window origin, which is the false claim that caused the bug. Correcting it at the site is right. |
| `src/tmux/session.rs` | `composite_pane0: Option<(u16,u16)>` to `Option<PaneGeom>`; `c.composite_pane0 = layout.first_pane()`. Plus the new tmux-gated geometry test (~110 lines). Correct. |
| `src/tmux/mod.rs` | `pub use composite::PaneGeom;`. Widens the lib crate's public surface by one plain data struct; unavoidable given `PaneCursor` is already `pub`. |
| `src/server/live_ws.rs` | The actual fix. `pane0` hoisted once, origin applied to the cursor, `left`/`top` added to the payload. Single emission point (`live_ws.rs:736` is the only caller), so no chance of a partially-translated frame. |
| `src/tui/home/{input,live_send,render}.rs` | Mechanical tuple to struct plus doc comments. **No TUI behavior change** (see issue 1). |
| `web/src/lib/liveMouse.ts` | `pointerPaneCell` inverse: `col = clamp(compositeCol - left, 1, cols)`, `row = clamp(compositeRow - top + 1, 1, rows)`. The arithmetic was checked against the compositor's layout: pane 0 owns composite 1-based columns `left+1..left+width` and 0-based rows `top..top+height-1`, so both are right. |
| `web/src/components/MobileLiveTerminal.tsx` | Sole `pane0` consumer in the web tree (grepped; `LiveTerminalView` delegates to it). The per-field `??` fallbacks collapse into one object-level fallback, which is equivalent because the server always sends both `cols` and `rows`. |

**Two symptoms, one root cause: confirmed, and the fix moves them in opposite directions without over-correcting.**

- Cursor render needs **+top**: client does `lineIdx = max(0, lines.length - screenRows) + cursor.y`, indexing the composite. With `top = 1` and `cursor.y = 63` it landed on composite row 63, the input box's top border, while pane 0's row 63 paints at composite 64. Server now sends 64.
- Mouse forward needs **-top**: client did `compositeRow + 1` and reported pane row `i + 1` for composite row `i`. Now `compositeRow - top + 1`.

These are genuinely one defect (pane 0's origin never crossing the boundary), not two that mask each other. A single naive "add the offset everywhere" would have doubled the mouse error; the PR does not do that.

**Unsplit case: pinned, and it does not regress.** `composite_pane0` is `None` on any window with `count <= 1` (early return in `capture_window_composited_with_cursor`), so `origin` is `(0, 0)` and both translations are the identity. Three tests pin it: the `None` row of `frame_json_includes_geometry_and_cursor`, the identity rows of the `pointerPaneCell` Vitest, and the pre-existing Playwright "bottom-aligned row" case (no `pane0`, still expects row 2).

**Other split shapes generalize.** `pane-border-status bottom` puts the gap row below, leaving `top == 0` and the translation an identity. A horizontal split leaves pane 0 at `left == 0` (it is the top-left pane by tmux's index ordering), so the `left` term is defensive rather than load-bearing, but it is symmetric on both sides so it stays consistent if a layout ever produces one. The new tmux test covers horizontal, vertical, and stacked. Only pane 0 receives input (#435 / #488) so "a pane that is not pane 0" does not arise for the cursor.

**Verification run** (rebased on `origin/main` at `adc053c2`, own worktree, `env -u CLAUDE_CONFIG_DIR`):

- `cargo fmt --all -- --check` clean.
- `cargo clippy --features serve -- -D warnings` clean. (`--all-targets` is red, but on `src/process/macos.rs` and `src/session/smart_rename.rs`, both untouched by this PR and pre-existing. CI's Clippy job is green.)
- `cargo test --features serve` **fully green, exit 0**, no failures at all.
- `npm run format:check && npm run lint` clean; full Vitest **3713 passed / 353 files, 0 failures**. The PR body's "2 pre-existing locale-dependent failures in `Composer.usageHint.test.tsx`" did not reproduce here, so that appears to be environment-local, not a real gap.
- Mocked Playwright `tests/live-click-forward.spec.ts`: 7/7 pass.

**Mutation checks (reverted, not assumed):**

- Reverted the `frame_json` translation, and `frame_json_includes_geometry_and_cursor` fails with `left: Number(3), right: 5`, matching the PR body's claim character for character.
- Forced `composite_pane0` to `top: 0`, and the tmux test fails with `left: (0, 0, 40, 23), right: (0, 1, 40, 23)`, against real tmux 3.7c geometry. This also proves the tmux test genuinely runs and is not silently taking its `tmux_available()` skip branch.
- Reverted `pointerPaneCell`'s origin subtraction, and the new Vitest case fails while the identity cases stay green.
- Same revert **with a `npm run build` rebuild**, and the new Playwright case fails while the other six, including the unsplit control, stay green. (Worth flagging for anyone repeating this: `playwright.config.ts` serves `dist/` via `vite preview`, so a source-only revert silently tests the stale bundle and passes.)

- Conventional commit prefixes, `fix/...` branch name, template followed including the honest "no screenshot, here is why" note.
- No emdashes in any added prose (scanned the added lines).
- No dead code: every `PaneGeom` field is read on the web path.
- `web/tests/coverage-matrix.json` **correctly needs no change**. Checked: `terminal.mobile-live-click-forward` already lists `src/lib/__tests__/liveMouse.test.ts` and `terminal.mobile-live-click-forward-e2e` already lists `tests/live-click-forward.spec.ts`, so both touched specs are registered. The PR's claim here is accurate.
- No workflow, dependency, lockfile, or build-hook changes anywhere in the diff.
- Doc-link check: the intra-doc link split across two lines in `composite.rs` does resolve. Reproduced in an isolated crate with `#![deny(rustdoc::broken_intra_doc_links)]` plus a control that correctly errors. Ugly but valid, and the Docs CI job is green.

## Automated review check-in

- **CodeRabbit, "no actionable comments":** agree, and no correctness defect was found here either.
- **CodeRabbit title check ("noun phrase, not imperative mood"):** disagree. The repo's own `Lint PR title` job passes, and this is CodeRabbit's convention rather than the project's. Not worth a rename.
- **Codecov:** all modified lines covered, project +0.04%. Consistent with the local run.

## Issues found

**Nothing blocking.** All non-blocking.

**1. The TUI has the same bug and this PR does not fix it.** `src/tui/home/render.rs:141`'s `map_live_preview_cursor` indexes the composited grid with untranslated `cursor.y`, and `mouse_pane_rect` at `src/tui/home/input.rs:242` still builds pane 0's sub-rectangle at `pane.x, pane.y` rather than `pane.x + rect.left, pane.y + rect.top`. Concretely: a TUI user previewing a session with `pane-border-status top` and a 71-row pane at `top == 1` sees the caret on the input box's top border and clamps clicks against a row that is one off, exactly the #3515 symptom.

This is disclosed in the PR body and marked with residual comments at both sites, and the data needed to fix it now rides on `composite_pane0`, so it is a two-line follow-up. But the `fix(web):` scope understates it: the root cause is shared and one of the three consumers is left broken. Either take the two lines here (the geometry is already carried and tested) or open a tracking issue, rather than leaving the codebase with the same struct interpreted two different ways.

**2. The new mocked-Playwright case did not need a browser.** `web/AGENTS.md` and the repo's cost table put a mocked Playwright test at ~200x a Vitest test. Whether jsdom can see this was checked: `MobileLiveTerminal.forward.test.tsx`'s own comment says only the **column** axis is unstable in jsdom, and the existing "desktop wheels keep the pointer cell" test already asserts a concrete forwarded **row**. A throwaway probe in that file asserting `offRow === baseRow - 1` for `pane0.top = 1` versus no origin passes in jsdom in ~15ms. So the offset wiring (`frame.pane0` to `pointerCell` to `forwardButton`) is fully assertable one tier down.

That said, the new Playwright case does hold up as an end-to-end guard and it demonstrably fails on revert, so this is a cost observation rather than a defect.

**3. The two adjacent Playwright cases should be one table.** `tests/live-click-forward.spec.ts:65` and `:84` are the same test with `pane0` added and the expected row changed from `2` to `1`. Per the house "one test per behavior, permutations as rows" rule, folding them saves a whole `setup()` including a page load (~1.9s of the ~2.5s measured per case).

**4. The #3515 explanation is restated in roughly fifteen places** across `composite.rs` (module doc, `PaneGeom` doc, `first_pane` doc, a test comment), `session.rs` (field doc, method doc, test doc, test comment), `live_ws.rs` (module doc, `frame_json` comment), `live_send.rs`, `render.rs`, `input.rs`, `liveMouse.ts` (twice), and `useLiveTerminal.ts`. The house rule says pick one canonical home because copies diverge.

In fairness, most of these sites previously carried their own **wrong** claim ("tmux puts pane 0 at the window origin"), so leaving them silent would leave the lie in place; the duplication is inherited from the original code, not invented here. Still, several could shrink to "see `PaneCursor::composite_pane0`" with the full story living only on the field doc.

**5. Nit:** the `frame_json` test comment says "Unsplit: `pane0` is absent", but the field serializes as JSON `null`, not omitted (which is also what the assertion checks). One word.

**6. Pre-existing, worth knowing, not introduced here:** `first_pane()` is really "first *surviving* pane." If `PaneGeom::parse` fails on pane 0's segment it is dropped and pane 1's geometry becomes the carried rectangle. Before this PR that only mis-sized the mouse clamp; now it would also shift the painted cursor. tmux's format output is machine-generated so this is close to unreachable, and `capture_window_layout`'s overlap filter explicitly keeps pane 0 first, so not worth acting on.

## Documentation

Proportionate on balance, with one caveat.

The doc changes are not narration of the change; they are corrections to comments that asserted something false, which is the highest-value kind of prose in a diff, and the residual markers on `render.rs` / `input.rs` are exactly the "non-obvious why" the guidelines want. No standalone markdown file was added, and nothing in the tree went stale: `docs/guides/live-mode.md`'s "Split windows" section describes user-visible behavior only and says nothing this fix falsifies, and grepping `docs/` and `website/` for any mirror of the live-ws frame protocol finds none (the protocol is documented in the `live_ws.rs` module doc, which was updated).

Caveat is item 4 above: the same paragraph appears about fifteen times. Worth trimming maybe 20-25 lines of restatement, not the corrections themselves.

## Suggestions

- Take the two-line TUI fix (`render.rs` cursor `+ top`, `input.rs` rect origin `+ left / + top`) here, or file the follow-up. The geometry is already carried and already tested.
- Fold the two Playwright click cases into one table; consider demoting the offset assertion to a row in `MobileLiveTerminal.forward.test.tsx`, which was confirmed able to catch it.
- Optional: add `pane0?: LivePaneRect | null` to `MockHandle["pushLiveFrame"]` in `tests/helpers/terminal-mocks.ts` so the new spec does not need the `as Parameters<...>[0]` cast. The cast follows the file's own precedent, so this is cosmetic.
- Optional: the composite still paints a full-width rule on the border-status row. The PR calls this a residual, but it is closer to correct than "phantom" as the issue framed it: tmux really does draw a titled border line there, and the composite is just rendering it untitled. Leaving it is the right call and the rationale (dropping it would render a grid shorter than `window_height` and rebase every consumer) is sound.

## Merge risk

**Low.**

*Blast radius:* confined to composited previews, and only ones whose pane 0 is off the window origin. A single-pane window returns before the compositor runs, so the overwhelmingly common path is byte-identical, which three separate tests pin. If the translation were wrong it would be loudly and immediately visible (a visibly misplaced caret), not a silent wrong answer in a data path.

*Reversibility:* clean `git revert`. No schema, no migration, no persisted state. The two new wire fields are additive and optional, and pre-fix clients degrade to exactly today's behavior rather than to something new.

*Coverage:* the untested seam is the round trip itself. The Rust side pins "server emits `y + top`," the TS side pins "client subtracts `top`," but no single test spans both, because the forward translation is in Rust and the inverse is in TypeScript. If someone later moved the cursor translation to the client, the cursor would double-shift and every existing test would stay green. The mitigations are decent (the wire contract is documented at the emission point, and `cursorLineIndex` takes no `pane0` argument so there is nothing to subtract with), and a live Playwright spec spanning it would cost far more than it is worth. Not worth holding the PR for.

*Most likely way it bites:* a user on the TUI preview hits the identical one-row offset and reports it as "you said you fixed #3515." Nothing in the PR surfaces that; it would arrive as a user report. That is issue 1 above, and it is the single argument for taking the two extra lines now.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Proportionate (trim ~20 lines of restatement if you like)
**Merge risk:** Low; residual TUI copy of the same bug, disclosed and commented but unfixed

This is a precisely diagnosed off-by-one with a named real-world trigger, a root cause verified against the compositor's own indexing rather than against the PR's description, and mutation checks reproduced in all four directions including a real Chromium rebuild. Both symptoms provably move in opposite directions from one origin value, the unsplit path is pinned by three independent tests, and the whole suite is green locally. The one thing worth pushing back on is scope rather than correctness: the same defect is still live in the TUI preview, documented but not fixed, and the two lines that would close it are now trivially available.

**LGTM** — merge it, and either take those two lines or file the follow-up before closing #3515.

Issue #2755, opened 2026-07-09
session: defer `mark_unread` until `persist_session_update` returns Ok — #2729 follow-up


jerome-benoit:
### Summary

The daemon's [`status_poll_loop`](src/server/mod.rs) applies `inst.mark_unread()` to the local instances vec at [`src/server/mod.rs:3386`](src/server/mod.rs) *before* awaiting [`api::persist_session_update`](src/server/api/sessions.rs) at `:3398`. On persist failure the mark survives in memory until [`reload_state_instances_from_disk`](src/server/mod.rs) overwrites it, and the next tick's `prev == inst.status` short-circuit prevents retry via the [`PassiveTransitionDecision`](src/server/mod.rs) filter. A Running→Idle transition whose flock write failed loses its unread indicator with no user-visible recovery path.

### Proposed direction

Two options:

1. **Defer `mark_unread` until persist Ok**: reorder so `inst.mark_unread()` is only applied after `persist_session_update` returns `Ok`. The local `instances` vec remains mutable through the persist await; the current code shape supports this.
2. **Retry on persist failure**: keep the in-memory mark, but on persist `Err` push the `(profile, session_id)` pair into a bounded retry queue drained on the next tick. Bound the queue to avoid pathological flock-contention loops.

Option 1 has lower complexity and no memory footprint. Option 2 tolerates transient flock loss but adds a queue and a bound.

### Acceptance

- Integration test: simulate a persist failure (readonly filesystem, revoked permission, or forced `Storage::update` error injection) and prove the unread indicator either persists (Option 1) or is retried on the next successful tick (Option 2).
- `cargo test --features serve --lib` green.

### Context

Filed from the review of merged PR #2729. Related: #2690, #2697.

jerome-benoit:
Status at `d9fa4188`: partially addressed by the #2729 refactor (`334f50c2`), not fully resolved. Keeping open.

- The persist closure in `status_poll_loop` now correctly calls `inst.mark_unread()` on the disk rows inside `persist_session_update`'s `Storage::update` (`src/server/mod.rs` ~3439), so the durable half is right.
[...]

PR #3530, opened 2026-08-25
fix(serve): mark structured sessions unread when an ACP turn ends

### LINKED ISSUE #3181: Structured sessions never get the automatic unread mark when a turn finishes

## Problem

`session.unread_indicator` is on by default and promises, at `src/session/config.rs:1200`:

> Show an unread indicator on sessions. When on (default), a session whose turn just finished is painted in the theme's unread color until you open it.

Structured sessions never get that. No code path marks one unread when its turn finishes. The row stays read, so a turn that completes while you are looking at another session leaves nothing behind to tell you.

Manual marking still works: `U` calls `Instance::toggle_unread` regardless of view, and the sidebar renders the marker for any row that carries it. Only the automatic mark on turn completion is missing.

Raised by @jerome-benoit reviewing #3162, which is what made the gap visible.

## What's there today

Exactly two places compute an automatic unread mark, and neither can fire for a structured row.

**1. The daemon poll loop.** `decide_passive_transition` at `src/server/mod.rs:3830` is not gated on `is_structured`:

```rust
let mark_unread = unread_enabled
    && old_status == Status::Running
    && inst.status == Status::Idle
    && !inst.unread;
```

It is only reached for a row that `observed_transitions` reported. After #3162 a structured row always compares equal to `prev` there, so it is never reported and this never runs for one. That is correct: the poll loop has no authority over a paneless row, and the transitions it used to report for them were phantoms.

**2. The TUI.** `src/tui/home/mod.rs:2897`, also ungated:

```rust
let should_mark_unread = crate::session::unread_enabled()
    && old == Status::Running
    && new_status == Status::Idle
    && !is_live_target
    && !already_unread;
```

This needs a `Running -> Idle` edge inside a `StatusUpdate`. For a structured row the producer cannot make one: `status_poller.rs:227` calls `update_status_with_metadata`, which short-circuits at `src/session/instance.rs:4965` for `is_structured()` without touching `status`, then `status_poller.rs:237` reports that unchanged value. `old == new`, so the guard at `:2879` skips the whole block.

Nothing else fills the gap:

- The path that *does* move a structured session's status, `apply_status_intent` at `src/server/mod.rs:5222`, never calls `mark_unread`. It sets status and broadcasts a `StatusChange`, nothing more.
- The TUI has no ACP status path at all. `derive_acp_status`, `apply_status_intent`, and `StatusIntent` have zero occurrences under `src/tui/`.
- `src/acp/` has no `unread` logic of any kind.

## Root cause

Automatic unread is wired to the tmux status pipeline, on both the daemon and TUI sides, and a structured session does not travel through it. Its lifecycle runs on ACP events into `apply_status_intent`, which was never taught the unread concern.

Before #3162 the daemon appeared to mark structured sessions unread, but only via the phantom transition: a row whose live status was `Running` while disk said `Idle` got `mark_unread` on every 2s tick. That fired at an arbitrary moment unrelated to any turn ending, and re-marked the row seconds after the user read it (228 such writes in one log). So the behavior was never correct; #3162 removed a wrong implementation rather than a working one.

## Proposal

**A. Mark unread on the ACP `Stopped` edge.** `derive_acp_status` at `src/server/mod.rs:5436` already maps `Event::Stopped` to `StatusIntent::Set(Status::Idle)`, which is precisely "the turn just finished". Have `apply_status_intent` apply the same predicate the other two sites use when it takes a `Running -> Idle` transition:

```rust
if crate::session::unread_enabled()
    && prev == Status::Running
    && target == Status::Idle
    && !inst.unread
{
    inst.mark_unread();
}
```

Two things to settle in the implementation:

- **Durability.** `apply_status_intent` mutates `state.instances` only. `unread` is a persisted field, and `decide_passive_transition` returns `patch: None` for structured rows so the poll loop will not carry it to disk. The mark needs its own write, or it dies with the daemon.
- **Suppression while viewing.** Both existing sites skip the mark when the user is looking at the session (`!is_live_target` in the TUI, and the daemon relies on the user not being attached). The web dashboard has an equivalent: `AppState::web_active_within` is already used for push suppression in `src/server/push.rs:551`. Whatever is chosen should not mark a session the user is actively reading, or the fix reintroduces the annoyance #3162 removed by a different route.

**B. Gate the two existing sites on `!is_structured()`.** Once (A) owns the concern, the ungated predicates in `decide_passive_transition` and `src/tui/home/mod.rs:2897` are reachable for structured rows only through a bug. Making the ownership explicit stops a future change to either pipeline from re-marking them from two places at once. Cheap, and it documents which path owns the mark.

- As a user with a structured session, when its turn finishes while I am looking at another session, then the row is marked unread until I open it.
  - Suggested test layer: Rust unit test on `apply_status_intent` in `src/server/mod.rs`, asserting `Running -> Idle` from an ACP `Stopped` sets `unread`.
- As a user, when I read that structured session, then it stays read and no later poll tick re-marks it.
  - Suggested test layer: Rust unit test alongside `tick_reports_no_transition_for_a_structured_phantom` in `src/server/mod.rs`, so the #3162 fix and this one are locked together.
- As a user with `session.unread_indicator = false`, when a structured turn finishes, then nothing is marked unread.
  - Suggested test layer: Rust unit test on the same predicate with `unread_enabled` off.
- As a user watching a structured session in the dashboard, when its turn finishes in front of me, then it is not marked unread.
  - Suggested test layer: live Playwright spec under `web/tests/live/`, or a Rust unit test on the suppression predicate if the decision is made server-side.

## Open questions

- Is `Stopped -> Idle` the right edge, or should a `Waiting` turn that the user answers and then completes also count? `derive_acp_status` maps several events to `Idle`; only `Event::Stopped` is unambiguously "turn finished".
- Should the mark persist across a daemon restart? Doing it right means a durable write from `apply_status_intent`, which currently touches only in-memory state. Accepting in-memory-only would make the marker vanish on restart, which is inconsistent with terminal sessions.
- Which "user is looking at it" signal should suppress the mark for a structured session: `web_active_within`, TUI presence, per-session focus, or all three?

## Out of scope

- The phantom transition loop itself. Fixed in #3162; this issue is only the gap that fix exposed.
- `seed_acp_statuses` re-deriving a stale `AgentStartupError` on every daemon boot. Separate defect, worth its own issue.
- Any change to how the unread marker is rendered or sorted.

## Cross-references

- #3162 removed the phantom transition that was the only thing (incorrectly) marking structured sessions unread. This gap was found reviewing it.
- #2571 (closed) covered the render side of the unread indicator, painting on archived and snoozed rows. Adjacent surface, different layer.

Seluj78:

## Description

`session.unread_indicator` is on by default and promises, at `src/session/config.rs`, that "a session whose turn just finished is painted in the theme's unread color until you open it." Structured (ACP) sessions largely did not get that.

To be precise, because #3181's title and an earlier draft of this description both overstated it: a TUI home view running next to a daemon *did* mark these rows, via `apply_daemon_status_update`. What got nothing was every other case, which is most of them. A dashboard-only user, since that path is inside the TUI's home loop. Anyone with no TUI running. Any turn shorter than the 1s `DAEMON_STATUS_REFRESH_INTERVAL`, since the poller has to catch both sides of the edge. And `/api/sessions` never reported the mark at all, because it lived in TUI memory rather than the daemon's.

Both existing producers of the automatic mark hang off the tmux status pipeline, and a structured session does not travel through it. The daemon's `decide_passive_transition` is only reached for a row that `observed_transitions` reported, and since #3162 a structured row always compares equal to `prev` there. That is correct rather than a regression: the poll loop has no authority over a paneless row, and the transitions it used to report for one were phantoms that re-marked a row seconds after the user read it. A structured session's lifecycle runs on ACP events into `apply_status_intent`, which was never taught the unread concern.

So the live ACP listener learns it. It snapshots the row's status around `apply_status_intent` and marks on `Running -> Idle`, the same predicate the two tmux-side sites already use. Snapshotting is exactly "the transition that was actually applied": `apply_status_intent` assigns `status` in one place and every rejected or no-op path leaves it untouched, and the write lock is held across both reads. That avoids changing its signature, which matters because `seed_acp_statuses` shares the function at daemon boot to replay the latest on-disk lifecycle event; marking there would re-mark from history on every restart.

The write is durable rather than in-memory. `reload_state_instances_from_disk` rebases every row on the freshly loaded disk row on each 2s poll tick, and `merge_runtime_fields` does not carry `unread`, so a memory-only mark would be gone within two seconds. Memory is mirrored only after the write lands, the #2755 ordering, so a mark disk never accepted is not served over `/api/sessions`, mirrored into the TUI, and then silently dropped by the next reload.

Ownership is then made single-writer. `decide_passive_transition` is gated on `!is_structured()`; it is unreachable for a structured row today, but the gate stops a later change to the poll loop from quietly re-marking one from a second daemon path. The TUI's predicate is gated too: `apply_daemon_status_update` (#3170, #3201) folds a structured row's daemon-sourced status through the same `apply_status_update` path a tmux transition takes, so the TUI's mark has in fact been firing for structured rows, and the daemon writing the mark itself makes that a second writer of one boolean.

Nothing is lost on the suppression or clear side, which is what settles the issue's three open questions:

- **`Running -> Idle` is the whole edge.** An approval or elicitation excursion comes back through `Set(Running)` (`ApprovalResolved` / `ElicitationResolved`) before the turn's `Stopped`, so an answered-then-completed turn still ends on this edge and needs no case of its own. The only edge missed is a direct `Waiting -> Idle`, which means the turn stopped while still blocked on a user who is by construction present for it. Keeping the predicate byte-identical across all three sites means "a turn just finished" means one thing everywhere.
- **The mark must be durable**, per the 2s disk rebase above. Not a preference.
- **No server-side suppression.** `AppState::web_active_within` is a *global* any-dashboard-focused boolean with no per-session dimension, so using it would suppress session B's mark because a dashboard is focused on session A. Both clients already clear reactively: the dashboard's active-session effect in `web/src/App.tsx` (whose own comment covers "having it open when its turn finishes"), and the TUI's `tick_unread_dwell`, which re-reads `is_unread()` every tick and so picks up a daemon-written mark on the row under the cursor. The TUI exemption that might have argued for keeping its own mark, `!is_live_target`, is inert on a structured row regardless, because `start_live_send` returns early on `is_structured()` (`src/tui/home/input.rs:6158`, with a matching guard at `src/tui/app.rs:3616`). The absence of a pane is not the reason: a structured row can still have a paired terminal or tool pane.

The write and the mirror happen under the same `instance_lock` that `PATCH /api/sessions/:id/unread` takes, and only a write that actually mutated the owning row is mirrored. An earlier revision of this PR did neither, on the reasoning that the mark was invisible until mirrored. Review showed that reasoning was wrong in both directions: `persist_session_update` lands the disk write *before* the mirror, and the TUI reads `unread` from disk on its watcher-driven reload, so the mark is visible in that window; and `persist_session_update` reports `Ok` for a write whose closure matched nothing, so a stale profile could mark memory off a successful no-op.

A dropped broadcast frame is also recovered. `acp_events_tx` holds 256 frames and tells a lagged receiver only how many it missed, never which. Status tolerated that by being level-triggered, but the unread mark is edge-triggered, so a dropped `Stopped` lost it permanently. Since ACP events are recorded before they are broadcast, the `Lagged` arm now replays the structured rows from the durable log.

Closes #3181

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo test --features serve` is 6378 passed / 1 failed locally, and the one failure is
`git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents`, a macOS
`/private/var` vs `/var` tmpdir-symlink assertion in a file this PR does not touch. Confirmed
pre-existing by running it against the untouched pre-fix tree, where it fails identically.
`cargo fmt --all -- --check`, `cargo clippy -- -D warnings` (CI's exact invocation),
`cargo clippy --features serve -- -D warnings`, and
`cargo check --no-default-features --lib --bins --tests` are all clean.

No `docs/` change: `grep -rn unread docs/` is empty, and the promise this makes true already lives in the `session.unread_indicator` doc comment. No new config field, so no settings parity. No stored-data change, so no migration.

## How to test

- [ ] With `aoe serve` running and `session.unread_indicator` on (the default), open the dashboard, start a turn in a structured session, then switch to a different session. When the turn finishes, the first session's sidebar row paints unread.
- [ ] Open that row. It clears, and stays clear across several 2s poll ticks (this is what the old phantom-transition behavior got wrong).
- [ ] Restart `aoe serve` while a structured row is unread. The marker is still there, so it is genuinely on disk rather than in daemon memory.
- [ ] Sit *on* a structured session in the dashboard while its turn finishes. No unread dot appears.
- [ ] With the TUI home view open next to the daemon, let a structured turn finish on a row you are not sitting on. The row paints unread there too, and dwelling on it clears it.
- [ ] Set `session.unread_indicator = false`, **restart `aoe serve`**, finish a structured turn, and confirm nothing is marked. The restart is required and is pre-existing behavior, not something this PR changes: the daemon reads the flag once at startup (`src/server/mod.rs`, "read once at startup; a config change needs a restart to take effect"). Without it the daemon still writes the mark, though both render surfaces gate on their own live copy of the flag so nothing paints.
- [ ] Confirm terminal (tmux) sessions still mark unread on turn end exactly as before.

## Test Coverage Analysis

Zero `web/` files change. The dashboard-side behavior the issue asks about (a session open in front of you is not left unread) is the pre-existing clear-on-view effect in `App.tsx`, and its render half already has Vitest coverage in `SessionRowTriage.test.tsx` ("suppresses the unread dot on the active row").

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: driving this end to end needs a live `aoe serve` plus the fake ACP agent to run a turn to completion, the 10-28s e2e tier, to assert one boolean that the unit tests already pin from both sides. Covered instead by:

  - `should_mark_acp_unread_only_on_a_structured_running_to_idle_turn_end`, a table over the daemon's predicate (8 cases: the turn-end edge, `Waiting -> Idle`, `Running -> Error`, turn start, no-transition-applied, feature off, already-unread, and a terminal row).
  - `tui_persists_neither_status_nor_unread_for_a_structured_turn_end`, which locks the TUI ownership gate on both the in-memory and the on-disk half. This is the retargeted `daemon_status_persists_the_unread_mark_but_not_the_status_for_structured`; see the third commit for why its old assertion was resting on a premise that #3162 had already invalidated.
  - the existing `decide_passive_transition_marks_unread_only_on_running_to_idle` table, which gains the structured case, deliberately alongside `tick_reports_no_transition_for_a_structured_phantom` so the #3162 fix and this one are locked together.

  - `acp_event_listener_marks_a_finished_turn_unread_on_disk_and_in_memory`, added after review. Both reviewers independently flagged that nothing exercised the listener, so the fix's own wiring was the one untested path and a future refactor of it would reintroduce the bug against a green suite. This one drives a real `Event::Stopped` frame through the broadcast and asserts the mark on disk and in memory.
  - `persist_and_mirror_unread_mirrors_only_a_committed_mutation` and `..._skips_the_mirror_on_a_failed_write`, covering the profile-move no-op and the failed write.
  - `recover_structured_unread_after_lag_marks_only_a_missed_turn_end`, covering a lagged `Running -> Stopped` against an already-reconciled row and a terminal row.

  Each new assertion was checked for the red direction rather than assumed. `tui_persists_neither_status_nor_unread_for_a_structured_turn_end` fails on the pre-fix tree on the mark assertion itself, not on setup. The listener test was verified against the exact hypothetical raised in review: moving the `old_status` snapshot to after `apply_status_intent` leaves the predicate table green and fails only that test.

  One residual gap, stated rather than papered over: the clear interleaving is now unrepresentable rather than tested. Holding `instance_lock` across the write and the mirror means a `PATCH /unread` clear cannot land between them, so a test for it would be asserting `tokio::sync::Mutex` rather than this code. A stale-profile write is also not retried; the mark is dropped rather than half-applied, which is a deliberate call given how rare a move mid-turn-end is.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (via Claude Code, `/aoe-implement`)

Investigation, plan, and implementation drafted by Claude under my direction, with a two-round design debate between two external models to pressure-test the three open questions in the issue. I made the design calls, reviewed each commit, and ran the manual test steps. Two of the issue's own proposals were revised against what the code actually does: the issue says the TUI has no ACP status path, which #3170/#3201 made stale, and the durability question resolves to "required" rather than "preferred" because of the 2s disk rebase.

## Summary

- ACP sessions now become unread when a turn changes from `Running` to `Idle`.
- The unread state is saved to disk and updated in memory only after a successful save.
- Event-log recovery restores unread updates when live ACP events are delayed or missed.
- Tmux and TUI paths no longer duplicate unread marking for structured sessions.
- Tests cover persistence, recovery, locking, listener behavior, and disabled-feature cases.

## Benefits

This keeps unread status consistent across restarts and event delays. It also prevents duplicate updates from competing session owners.

## Further enhancement

The behavior for turns that change from `Waiting` to `Idle` remains open for future clarification.

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -4299,6 +4299,9 @@ struct PassiveTransitionDecision {
     /// intra-doc link that would degrade to literal text under
     /// `cargo doc`).
     patch: Option<crate::session::PassiveStatusPatch>,
+    /// Always `false` for structured / ACP sessions: `should_mark_acp_unread`,
+    /// driven off the live ACP turn-end event, is the sole producer of their
+    /// automatic mark. See the gate in `decide_passive_transition`.
     mark_unread: bool,
 }
 
@@ -4306,9 +4309,9 @@ struct PassiveTransitionDecision {
 /// `status` differs from the tick's `prev` snapshot. The full
 /// contract lives on the return type at [`PassiveTransitionDecision`]:
 /// `patch: None` for structured / ACP sessions (the ACP overlay is the
-/// sole authority), and `mark_unread: true` only on genuine
-/// Running -> Idle when unread is enabled and the row is not already
-/// unread.
+/// sole authority), and `mark_unread: true` only on a genuine
+/// Running -> Idle for a *terminal* session when unread is enabled and the
+/// row is not already unread.
 fn decide_passive_transition(
     inst: &Instance,
     old_status: Status,
@@ -4316,7 +4319,14 @@ fn decide_passive_transition(
 ) -> PassiveTransitionDecision {
     let patch =
         (!inst.is_structured()).then(|| crate::session::PassiveStatusPatch::from_instance(inst));
+    // Structured rows are excluded for the same reason as the patch: the poll
+    // loop has no authority over a paneless row, and since #3162 one never
+    // reaches here anyway (it compares equal to `prev`, so `observed_transitions`
+    // does not report it). `should_mark_acp_unread`, driven off the live ACP
+    // `Stopped` event, is the sole producer for them; the gate is what stops a
+    // later change to this loop from quietly re-marking from two daemon paths.
     let mark_unread = unread_enabled
+        && !inst.is_structured()
         && old_status == Status::Running
         && inst.status == Status::Idle
         && !inst.unread;
@@ -5621,7 +5631,7 @@ async fn acp_event_listener(state: Arc<AppState>) {
 
         // Acquire `instances` once for both branches. Releases before
         // the (potentially blocking) sessions.json save.
-        let profile_to_save = {
+        let (profile_to_save, unread_profile) = {
             let mut instances = state.instances.write().await;
             let Some(inst) = instances.iter_mut().find(|i| i.id == frame.session_id) else {
                 continue;
@@ -5630,9 +5640,74 @@ async fn acp_event_listener(state: Arc<AppState>) {
                 continue;
             }
 
+            // Snapshotting around the call is exactly "the transition
+            // `apply_status_intent` actually applied": it assigns `status` in
+            // one place and every rejected or no-op path (the trashed /
+            // Deleting / Creating guard, the Stopped guard, the ineligible
+            // HealError guard, `status == target`) leaves it untouched. We hold
+            // the write lock across both reads, so nothing else can move it in
+            // between. A future refactor that makes `apply_status_intent`
+            // assign `status` more than once has to revisit this.
+            let old_status = inst.status;
             apply_status_intent(inst, status_intent, &state.status_tx);
-            apply_acp_session_change(inst, &frame.session_id, acp_change.as_ref())
+            let unread_profile =
+                should_mark_acp_unread(inst, old_status, crate::session::unread_enabled())
+                    .then(|| inst.source_profile.clone());
+
+            (
+                apply_acp_session_change(inst, &frame.session_id, acp_change.as_ref()),
+                unread_profile,
+            )
         };
+
+        // The turn just finished, so the row takes the automatic unread mark.
+        // This is the sole producer of it for a structured row. The tmux poll
+        // loop has no authority over a paneless one, which is why #3162 stopped
+        // it reporting phantom transitions for them and so left this gap; the
+        // TUI's passive path is gated off them to keep the boolean single-writer.
+        //
+        // The write has to be durable. `reload_state_instances_from_disk` rebases
+        // every row on the disk row on each 2s tick and `merge_runtime_fields`
+        // does not carry `unread`, so an in-memory-only mark is gone within two
+        // seconds. Memory is mirrored only after the write lands, the same
+        // ordering `flush_passive_transition_writes` uses (#2755) and for the
+        // same reason: a mark that exists only in daemon memory is served over
+        // `/api/sessions`, mirrored into the TUI, and then silently dropped by
+        // the next reload.
+        //
+        // Deliberately not folded into the `profile_to_save` save below:
+        // `derive_acp_session_change` yields nothing for `Event::Stopped`, so an
+        // identity change and a turn-end can never arrive on the same event and
+        // there is no atomicity to win.
+        //
+        // No per-instance lock. A concurrent `PATCH /api/sessions/:id/unread`
+        // clear cannot be lost to this write: `should_mark_acp_unread` requires
+        // the row to be read at decision time, and until the mirror below runs
+        // it still reads as read everywhere, so the reactive clears (the
+        // dashboard's active-session effect, the TUI's unread dwell) cannot have
+        // fired yet. A clear racing this window is therefore a no-op on an
+        // already-read row, and marking after it is the chronologically correct
+        // outcome: the turn did finish afterwards.
+        if let Some(profile) = unread_profile {
+            let persist_id = frame.session_id.clone();
+            let persisted = api::persist_session_update(
+                profile,
+                "acp turn-end unread",
+                state.file_watch.clone(),
+                move |instances| {
+                    if let Some(inst) = instances.iter_mut().find(|i| i.id == persist_id) {
+                        inst.mark_unread();
+                    }
+                },
+            )
+            .await;
+            if persisted.is_ok() {
+                let mut instances = state.instances.write().await;
+                if let Some(inst) = instances.iter_mut().find(|i| i.id == frame.session_id) {
+                    inst.mark_unread();
+                }
+            }
+        }
 
         // Persist `acp_session_id` to disk if the field changed.
         // Sync FS (file copy + JSON write) goes through spawn_blocking
@@ -5791,6 +5866,32 @@ pub(crate) fn apply_status_intent(
         new: target,
         at: now,
     });
+}
+
+/// Whether a structured row whose ACP status just moved should take the
+/// automatic unread mark, i.e. whether its turn just finished.
+///
+/// `inst` is the row *after* [`apply_status_intent`] ran and `old_status` is
+/// the snapshot taken before it. The predicate is deliberately byte-identical
+/// to the one in [`decide_passive_transition`] and in the TUI's
+/// `apply_status_update`, so "a turn just finished" means the same thing on
+/// every surface.
+///
+/// `Running -> Idle` is the whole edge. An approval or elicitation excursion
+/// comes back through `Set(Running)` (`ApprovalResolved` /
+/// `ElicitationResolved`) before the turn's `Stopped`, so an answered-then-
+/// completed turn still ends on this edge and needs no case of its own. A
+/// direct `Waiting -> Idle` means the turn stopped while still blocked on the
+/// user, who is by construction present for it. Every `Stopped` reason maps to
+/// `Idle`, so a rate-limit park marks unread too; that is the same policy
+/// terminal sessions get, and a parked session does want attention.
+#[cfg(feature = "serve")]
+fn should_mark_acp_unread(inst: &Instance, old_status: Status, unread_enabled: bool) -> bool {
+    unread_enabled
+        && inst.is_structured()
+        && old_status == Status::Running
+        && inst.status == Status::Idle
+        && !inst.unread
 }
 
 /// Fold a derived `AcpSessionChange` into an `Instance`. Returns the
@@ -7888,6 +7989,127 @@ mod tests {
             !decision.mark_unread,
             "already-unread sessions must not re-mark"
         );
+
+        // #3181: a structured row's turn-end mark belongs to the live ACP
+        // listener (`should_mark_acp_unread`), so the poll loop must not also
+        // produce it. Paired with
+        // `tick_reports_no_transition_for_a_structured_phantom` above, which
+        // covers the other half: the tick never even reports such a row, so a
+        // read structured session cannot be re-marked seconds after the user
+        // read it (the #3162 defect).
+        let mut structured = Instance::new("acp-session", "/tmp/test");
+        structured.view = crate::session::View::Structured;
+        structured.status = Status::Idle;
+        let decision = decide_passive_transition(&structured, Status::Running, true);
+        assert!(
+            !decision.mark_unread,
+            "structured turn-end unread is owned by the acp event listener"
+        );
+    }
+
+    /// #3181: the automatic mark's predicate for a structured row, driven off
+    /// the live ACP turn-end event. One table rather than a test per case, per
+    /// the repo's compile-cost rule.
+    #[cfg(feature = "serve")]
+    #[test]
+    fn should_mark_acp_unread_only_on_a_structured_running_to_idle_turn_end() {
+        // (name, structured, old_status, new_status, unread_enabled, already_unread, expected)
+        let cases = [
+            (
+                "turn finished",
+                true,
+                Status::Running,
+                Status::Idle,
+                true,
+                false,
+                true,
+            ),
+            // The turn stopped while still blocked on the user, who is by
+            // construction present for it; an answered approval comes back
+            // through Running first, so this is not the answered-then-completed
+            // path.
+            (
+                "still blocked on the user",
+                true,
+                Status::Waiting,
+                Status::Idle,
+                true,
+                false,
+                false,
+            ),
+            (
+                "crashed, not finished",
+                true,
+                Status::Running,
+                Status::Error,
+                true,
+                false,
+                false,
+            ),
+            (
+                "turn starting",
+                true,
+                Status::Idle,
+                Status::Running,
+                true,
+                false,
+                false,
+            ),
+            (
+                "no transition applied",
+                true,
+                Status::Running,
+                Status::Running,
+                true,
+                false,
+                false,
+            ),
+            (
+                "feature off",
+                true,
+                Status::Running,
+                Status::Idle,
+                false,
+                false,
+                false,
+            ),
+            // Re-marking would churn the flock once per turn, and would undo a
+            // read the user has not been given a new turn to earn.
+            (
+                "already unread",
+                true,
+                Status::Running,
+                Status::Idle,
+                true,
+                true,
+                false,
+            ),
+            // Terminal rows stay with the tmux poll loop's
+            // `decide_passive_transition`.
+            (
+                "terminal row, owned elsewhere",
+                false,
+                Status::Running,
+                Status::Idle,
+                true,
+                false,
+                false,
+            ),
+        ];
+        for (name, structured, old, new, enabled, already_unread, expected) in cases {
+            let mut inst = Instance::new(name, "/tmp/test");
+            if structured {
+                inst.view = crate::session::View::Structured;
+            }
+            // The helper reads the row *after* `apply_status_intent` ran.
+            inst.status = new;
+            inst.unread = already_unread;
+            assert_eq!(
+                should_mark_acp_unread(&inst, old, enabled),
+                expected,
+                "{name}"
+            );
+        }
     }
 
     // #2755 (follow-up to #2729): the poller must not strand an in-memory
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -595,17 +595,7 @@ impl AppState {
     /// `RwLock` is only held long enough to insert/lookup the `Arc<Mutex>`;
     /// the caller awaits the inner mutex without holding the map lock.
     pub async fn instance_lock(&self, id: &str) -> Arc<tokio::sync::Mutex<()>> {
-        {
-            let guard = self.instance_locks.read().await;
-            if let Some(lock) = guard.get(id) {
-                return lock.clone();
-            }
-        }
-        let mut guard = self.instance_locks.write().await;
-        guard
-            .entry(id.to_string())
-            .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
-            .clone()
+        instance_lock_in(&self.instance_locks, id).await
     }
 
     /// Get or create the per-idempotency-key serialization mutex. Same
@@ -5343,12 +5333,34 @@ async fn acp_event_listener(state: Arc<AppState>) {
             // reconcile on the next event; a missed acp_session_id
             // means at most one restart loses context. Far better to
             // continue than to exit the listener entirely.
+            //
+            // The unread mark does NOT self-heal like status does: it is
+            // edge-triggered on `Running -> Idle`, so a dropped `Stopped` would
+            // lose it for good and no later event would reproduce it. The
+            // events are durable, recorded before broadcast, so replay the
+            // structured rows from the event log before continuing.
             Err(tokio::sync::broadcast::error::RecvError::Lagged(skipped)) => {
                 tracing::warn!(
                     target: "acp.event_listener",
                     skipped,
                     "broadcast lagged; status and acp_session_id may briefly desync"
                 );
+                let recovered = recover_structured_unread_after_lag(
+                    &state.instances,
+                    &state.acp_event_store,
+                    &state.instance_locks,
+                    state.file_watch.clone(),
+                    &state.status_tx,
+                )
+                .await;
+                if recovered > 0 {
+                    tracing::info!(
+                        target: "acp.event_listener",
+                        skipped,
+                        recovered,
+                        "replayed turn-end unread marks missed by the lagged frames"
+                    );
+                }
                 continue;
             }
             // Closed: AppState dropped (shutdown). Exit cleanly.
@@ -5680,33 +5692,18 @@ async fn acp_event_listener(state: Arc<AppState>) {
         // identity change and a turn-end can never arrive on the same event and
         // there is no atomicity to win.
         //
-        // No per-instance lock. A concurrent `PATCH /api/sessions/:id/unread`
-        // clear cannot be lost to this write: `should_mark_acp_unread` requires
-        // the row to be read at decision time, and until the mirror below runs
-        // it still reads as read everywhere, so the reactive clears (the
-        // dashboard's active-session effect, the TUI's unread dwell) cannot have
-        // fired yet. A clear racing this window is therefore a no-op on an
-        // already-read row, and marking after it is the chronologically correct
-        // outcome: the turn did finish afterwards.
+        // `persist_and_mirror_unread` owns the lock and commit-check ordering;
+        // see its docstring for why both are load-bearing.
         if let Some(profile) = unread_profile {
-            let persist_id = frame.session_id.clone();
-            let persisted = api::persist_session_update(
-                profile,
-                "acp turn-end unread",
+            let lock = state.instance_lock(&frame.session_id).await;
+            persist_and_mirror_unread(
+                &state.instances,
+                &lock,
                 state.file_watch.clone(),
-                move |instances| {
-                    if let Some(inst) = instances.iter_mut().find(|i| i.id == persist_id) {
-                        inst.mark_unread();
-                    }
-                },
+                &frame.session_id,
+                profile,
             )
             .await;
-            if persisted.is_ok() {
-                let mut instances = state.instances.write().await;
-                if let Some(inst) = instances.iter_mut().find(|i| i.id == frame.session_id) {
-                    inst.mark_unread();
-                }
-            }
         }
 
         // Persist `acp_session_id` to disk if the field changed.
@@ -5868,6 +5865,27 @@ pub(crate) fn apply_status_intent(
     });
 }
 
+/// Get or create the per-instance serialization mutex in `locks`. Free function
+/// rather than only an [`AppState`] method so the ACP turn-end unread writers
+/// can take the *same* lock a REST handler takes without needing an `AppState`
+/// (which has no test constructor). See [`AppState::instance_lock`].
+async fn instance_lock_in(
+    locks: &RwLock<std::collections::HashMap<String, Arc<tokio::sync::Mutex<()>>>>,
+    id: &str,
+) -> Arc<tokio::sync::Mutex<()>> {
+    {
+        let guard = locks.read().await;
+        if let Some(lock) = guard.get(id) {
+            return lock.clone();
+        }
+    }
+    let mut guard = locks.write().await;
+    guard
+        .entry(id.to_string())
+        .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
+        .clone()
+}
+
 /// Whether a structured row whose ACP status just moved should take the
 /// automatic unread mark, i.e. whether its turn just finished.
 ///
@@ -5892,6 +5910,149 @@ fn should_mark_acp_unread(inst: &Instance, old_status: Status, unread_enabled: b
         && old_status == Status::Running
         && inst.status == Status::Idle
         && !inst.unread
+}
+
+/// Write the automatic unread mark for `id` to its profile store, then mirror it
+/// into daemon memory. Returns whether the mark actually landed.
+///
+/// Takes primitives rather than [`AppState`] so it is reachable from tests
+/// (`AppState` has no test constructor).
+///
+/// Ordering rules, both of which cost correctness if dropped:
+///
+/// 1. **Under `instance_lock`.** The same mutex `PATCH /api/sessions/:id/unread`
+///    takes, held across both the write and the mirror. Without it a clear can
+///    land between them and leave disk read while memory says unread, and the
+///    user's explicit mark-read loses to a mark it happened after. Holding it
+///    makes the two orderings the only ones possible, and both are correct: a
+///    clear before this marks (the turn genuinely finished afterwards), a clear
+///    after this wins (the user read it afterwards).
+/// 2. **Only mirror a committed mutation.** `persist_session_update` reports
+///    `Ok` for a write whose closure matched no row, so `profile` going stale
+///    (a concurrent profile move) would otherwise mark memory off a successful
+///    no-op on the *old* profile, and the next reload would drop the
+///    notification. The flag reports whether the owning row was really mutated.
+///
+/// A stale-profile write is not retried. The row is left read rather than
+/// half-marked, the turn's mark is simply lost, and the move is rare enough that
+/// a re-resolve loop is not worth the added failure surface here.
+#[cfg(feature = "serve")]
+async fn persist_and_mirror_unread(
+    instances: &RwLock<Vec<Instance>>,
+    instance_lock: &tokio::sync::Mutex<()>,
+    file_watch: Arc<crate::file_watch::FileWatchService>,
+    id: &str,
+    profile: String,
+) -> bool {
+    let _guard = instance_lock.lock().await;
+    let marked = Arc::new(std::sync::atomic::AtomicBool::new(false));
+    let marked_in_closure = marked.clone();
+    let persist_id = id.to_string();
+    let persisted = api::persist_session_update(
+        profile.clone(),
+        "acp turn-end unread",
+        file_watch,
+        move |instances| {
+            if let Some(inst) = instances.iter_mut().find(|i| i.id == persist_id) {
+                inst.mark_unread();
+                marked_in_closure.store(true, std::sync::atomic::Ordering::SeqCst);
+            }
+        },
+    )
+    .await;
+    if persisted.is_err() {
+        return false;
+    }
+    if !marked.load(std::sync::atomic::Ordering::SeqCst) {
+        tracing::debug!(
+            target: "acp.event_listener",
+            session = %id,
+            profile = %profile,
+            "turn-end unread write found no row in that profile store (moved?); \
+             not mirroring so memory cannot disagree with disk"
+        );
+        return false;
+    }
+    let mut instances = instances.write().await;
+    if let Some(inst) = instances.iter_mut().find(|i| i.id == id) {
+        inst.mark_unread();
+    }
+    true
+}
+
+/// Re-derive every structured row's status from the durable event log after the
+/// ACP broadcast dropped frames, marking any row whose turn ended while we were
+/// not listening. Returns the number of rows marked.
+///
+/// `acp_events_tx` is a `broadcast` of [`ACP_CHANNEL_CAPACITY`], and a lagged
+/// receiver is told only *how many* frames it missed, never which. Status
+/// tolerated that because it is level-triggered: any later event re-derives the
+/// right value. The unread mark is edge-triggered, so a dropped `Stopped` loses
+/// it permanently, and nothing else would ever produce it.
+///
+/// The events themselves are durable (recorded before broadcast), so the log is
+/// the recovery source. `latest_seed_status_event` is the same query
+/// `seed_acp_statuses` uses at boot, and for the same reason: it returns the
+/// most recent *lifecycle* event, which is exactly the frame whose loss matters.
+///
+/// This deliberately does NOT reuse `seed_acp_statuses`, despite the shared
+/// shape, because the two differ on both points that matter:
+///
+/// - **Boot must not mark.** Its replay re-reads history, so a `Stopped` from
+///   before the restart would re-mark a row the user has already read, on every
+///   restart. Here a `Stopped` under a still-`Running` memory status is evidence
+///   of a turn that ended during this daemon's life and was missed.
+/// - **Boot lifts a stale `Stopped`**, because a persisted `Stopped` from a
+///   daemon that died mid-turn would otherwise trap the dot grey. Mid-run a
+///   `Stopped` in memory is a deliberate stop, so `apply_status_intent`'s guard
+///   should keep it.
+#[cfg(feature = "serve")]
+async fn recover_structured_unread_after_lag(
+    instances: &RwLock<Vec<Instance>>,
+    event_store: &crate::acp::event_store::EventStore,
+    instance_locks: &RwLock<std::collections::HashMap<String, Arc<tokio::sync::Mutex<()>>>>,
+    file_watch: Arc<crate::file_watch::FileWatchService>,
+    status_tx: &broadcast::Sender<StatusChange>,
+) -> usize {
+    let ids: Vec<String> = instances
+        .read()
+        .await
+        .iter()
+        .filter(|i| i.is_structured())
+        .map(|i| i.id.clone())
+        .collect();
+    let unread_enabled = crate::session::unread_enabled();
+    let mut marked = 0usize;
+    for id in ids {
+        let Some(event) = event_store.latest_seed_status_event(&id) else {
+            continue;
+        };
+        let Some(intent) = derive_acp_status(&event) else {
+            continue;
+        };
+        // Same snapshot-around-apply as the live path, so "the transition that
+        // was actually applied" means the same thing in both.
+        let unread_profile = {
+            let mut guard = instances.write().await;
+            let Some(inst) = guard.iter_mut().find(|i| i.id == id) else {
+                continue;
+            };
+            if !inst.is_structured() {
+                continue;
+            }
+            let old_status = inst.status;
+            apply_status_intent(inst, Some(intent), status_tx);
+            should_mark_acp_unread(inst, old_status, unread_enabled)
+                .then(|| inst.source_profile.clone())
+        };
+        if let Some(profile) = unread_profile {
+            let lock = instance_lock_in(instance_locks, &id).await;
+            if persist_and_mirror_unread(instances, &lock, file_watch.clone(), &id, profile).await {
+                marked += 1;
+            }
+        }
+    }
+    marked
 }
 
 /// Fold a derived `AcpSessionChange` into an `Instance`. Returns the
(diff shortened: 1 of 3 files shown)


njbrake (commented):

_Reviewed at `a6c218bb`. Push new commits and it will be re-reviewed._

---

## Summary

Moves the automatic "turn finished" unread mark for structured/ACP rows out of the TUI and into the daemon's live ACP event listener, where it is written durably to `sessions.json` before being mirrored into daemon memory. Adds a `!is_structured()` gate to the two pre-existing producers so the boolean has a single writer. The 227 added lines in `src/server/mod.rs` decompose as 85 comment lines, ~110 lines of one table-driven test, and **~27 lines of production code**. This is one hook plus tests, not a new event path with its own state.

## Verifying the premise before judging the fix

The gap was checked rather than taken on the body's word, and it is real but **narrower than the first sentence of the PR body and the title of #3181 claim**.

The pre-PR test being retargeted here, `daemon_status_persists_the_unread_mark_but_not_the_status_for_structured`, asserted `disk.is_unread()` after a structured `Running -> Idle` through `apply_daemon_status_update`, and it is green on `main`. So a TUI home view running next to a daemon **did** mark structured rows unread. #3181's "No code path marks one unread when its turn finishes" is stale, and so is its "The TUI has no ACP status path at all."

The body says this outright further down ("the TUI's mark has in fact been firing for structured rows"), so the PR is not misleading, but its opening line and the linked issue's framing are. **Please correct the first paragraph before merge** so the commit history does not record a wrong root cause.

The residual gap is genuine and worth fixing:

- Dashboard-only users get nothing (`apply_daemon_status_update` is `#[cfg(feature = "serve")]` at `src/tui/home/mod.rs:3144` and only runs inside the TUI's home loop).
- Nothing at all when no TUI is running.
- Sub-second turns, since `DAEMON_STATUS_REFRESH_INTERVAL` is 1s (`src/tui/app.rs:874`) and the TUI needs to poll both sides of the edge.
- The mark was in TUI memory, not the daemon's, so `/api/sessions` never reported it.

The poll loop genuinely cannot reach `decide_passive_transition` for a structured row: `skip_tmux_decision_for_structured` sets `inst.status = live_status_baseline` (which *is* `prev`) at `src/server/mod.rs:3253-3255`, so `observed_transitions` (`:3199`) filters it out. The new `!inst.is_structured()` gate at `src/server/mod.rs:4327` is defensive rather than load-bearing, which is what #3181's Proposal B asked for. Not dead code by the AGENTS.md rule, since it is a condition in a live function rather than an unused item.

## Design alignment

**Aligned.**

The precedent is `PATCH /api/sessions/:id/unread` at `src/server/api/sessions.rs:4167-4202`: gate on `unread_enabled()`, call `persist_session_update`, then take the `instances` write lock and mirror. The new block at `src/server/mod.rs:5688-5710` is structurally the same code. That helper's own docstring (`src/server/api/sessions.rs:2348-2358`, #1589) states the persist-first rule the PR follows, and `flush_passive_transition_writes` (#2755) is the same ordering on the tmux side. It extends the existing mechanism rather than adding one beside it.

The durability argument is not just asserted, it checks out: `merge_runtime_fields` (`src/server/mod.rs:3101-3115`) carries no `unread`, and `reload_state_instances_from_disk` rebases every row on the fresh disk row (`:3388-3395`), so an in-memory-only mark would be gone inside one 2s tick. Likewise `merge_from_tui` (`src/session/instance.rs:1987-2003`) and `merge_runtime_from_reload` (`:1961-1976`) have no `unread` arm, so the TUI can neither clobber the daemon's write nor keep a stale one.

The `old_status` snapshot is sound: `apply_status_intent` assigns `inst.status` in exactly one place (`src/server/mod.rs:5846`) and every guard returns before it, with the write lock held across both reads.

The one other live caller of `apply_status_intent` (`src/server/acp_reconciler.rs:1402`) is a re-derive-from-history seed on attach, gated on `in_flight_turn`, and is correctly excluded for the same reason `seed_acp_statuses` is.

| Check | Result |
| --- | --- |
| `cargo clippy --features serve --all-targets -- -D warnings` | clean |
| `cargo test --features serve --lib` (rebased on `a0137269`) | 6414 passed, 2 failed, both the known macOS-environmental ones (`worktree_move_observation_canonicalizes_symlinked_parents`, `test_content_shows_settings_path`). Nothing else. |
| The four PR-relevant tests | all pass individually |
| Test-cost rules | followed. One new `#[test]` fn with an 8-row table; the `decide_passive_transition` case appended to the existing test rather than split out |
| Weakened tests? | no. `tui_persists_neither_status_nor_unread_for_a_structured_turn_end` (`src/tui/home/tests.rs:20685`) went from 2 assertions to 4 and asserts the new contract on both memory and disk |
| Coverage matrix | correctly N/A. `web/tests/validate-coverage-matrix.mjs` only requires an entry per `web/src/components/**/*.tsx`; zero `web/` files change and the dashboard's render and clear behavior is untouched |
| Docs | correctly none. `grep -rni unread docs/ website/` is empty; the promise lives in the `#[setting]` doc at `src/session/config.rs:1345-1352` and is now true |
| Emdashes in added prose | none |
| Migrations / settings parity | not applicable, no stored-data or schema change |

## Automated review check-in

CodeRabbit: "No actionable comments were generated." Nothing to reconcile. Its walkthrough matches what the diff does.

## Issues found

None blocking. All of the following are non-blocking.

**1. `session.unread_indicator` now answers to the daemon's startup-frozen copy.** `src/server/mod.rs:802-805` sets the flag once and its own comment says "read once at startup; a config change needs a restart to take effect." Nothing else in `src/server/` calls `set_unread_enabled`. Before this PR the structured mark was gated by the TUI's live-refreshed copy (`src/tui/home/mod.rs:2159`, `:7654`).

Concrete: user turns `session.unread_indicator` off in TUI settings while `aoe serve` is running; a structured turn ends; the daemon still writes `unread = true` into `sessions.json`. Both render surfaces gate on their own flag (`src/tui/home/render.rs:1651`, `useUnreadIndicatorEnabled` in `App.tsx`) so nothing paints, and `tick_unread_dwell` returns early when disabled (`src/tui/home/mod.rs:7046`) so nothing clears it either. Re-enable the setting later and stale unread rows appear. The dashboard's `PATCH .../unread` already has this property, so it is consistent rather than novel, but manual test step 6 in the PR body ("confirm nothing is marked") only holds after a daemon restart, and passes visually even when it should not. Worth a sentence in the body.

**2. The "No per-instance lock" comment's reasoning is wrong, though its conclusion is harmless.** `src/server/mod.rs:5682-5686` claims "until the mirror below runs it still reads as read everywhere." It does not: `persist_session_update` lands the disk write *before* the mirror, and the TUI reads `unread` from disk on its watcher-driven `reload_storage_only`.

Concrete: daemon writes `unread=true` to disk, TUI's file watcher reloads and picks it up, `tick_unread_dwell` (cursor already dwelled on that row) clears it and writes `unread=false`, then the daemon's mirror sets memory `unread=true` against a disk that says read. `/api/sessions` serves a spurious dot for up to one 2s reload tick, then self-heals. Cosmetic, but the comment as written is what would let a future reader conclude this cannot happen.

**3. The inverse race can silently drop a mark.** `should_mark_acp_unread` reads `inst.unread` from daemon *memory*, which trails disk by up to one 2s reload tick. Concrete: a row is unread from a previous turn; the user dwells on it in the TUI and it clears on disk; daemon memory still says `unread = true`; a turn ends inside that window; `!inst.unread` is false and the mark is skipped, so the row stays read with nothing to re-trigger it. Same staleness class `decide_passive_transition` already carries for terminal rows, so consistent, but it now exists on the structured path too.

**4. The wiring that actually fixes the bug has no test at any tier.** `should_mark_acp_unread` is tested as a pure predicate and the TUI ownership gate is tested from both sides, but nothing exercises the `old_status` snapshot, the `persist_session_update` call, or the memory mirror in `acp_event_listener` (`src/server/mod.rs:5646-5710`). No test in the repo drives `acp_event_listener` at all. If the snapshot were taken *after* `apply_status_intent`, or the persist dropped, every test in this PR still passes and #3181 is unfixed.

The skip rationale ("driving this end to end needs a live `aoe serve` plus the fake ACP agent") is fair for the e2e tier, but a unit test looks reachable: `build_test_app_state` (`src/server/mod.rs:6203+`) already hands you an `AppState` carrying `acp_events_tx`, so spawning the listener, publishing one `Event::Stopped` frame, and asserting disk plus memory is roughly 40 lines. It needs a temp `$HOME` for the `Storage` open, so `#[serial]`. This is the one test that could fail if the fix regresses, and it is worth more than the eighth row of the predicate table.

**5. Cadence change on the ACP event loop.** The persist is `.await`ed inline (`src/server/mod.rs:5690-5701`) in the loop that also carries approval requests and streaming frames for every structured session. Its sibling `profile_to_save` write below has the same shape but fires roughly once per session lifetime; this one fires on every turn end, and the dashboard's reactive clear (`web/src/App.tsx:814-817`) makes "watching a session while its turns finish" cost two whole-file flock'd `sessions.json` rewrites per turn. Probably fine at human turn cadence, but `tokio::spawn`ing the persist-plus-mirror pair would keep the listener off the flock entirely.

**6. A comment states the right conclusion for the wrong reason.** `src/tui/home/mod.rs:3313-3317` says `is_live_target` is inert on a structured row because "live-send needs a tmux pane (`LiveSendState.tmux_name`) and a structured row has none." A structured row *can* have a paired terminal, container-terminal, or tool pane (`LiveSendTarget::Terminal` and siblings, `src/tui/home/mod.rs:6274-6284`). The real reason is the explicit `if inst.is_structured() { return None; }` in `start_live_send` (`src/tui/home/input.rs:6158`) and the matching guard at `src/tui/app.rs:3616-3623`. Conclusion verified correct; cite the guard instead.

**7. Maintainer sign-off item, not a defect.** `Waiting -> Idle` is excluded on the stated grounds that "the turn stopped while still blocked on the user, who is by construction present for it." That is not always true: an agent that crashes or is cancelled with an approval still pending goes `Waiting -> Idle` with the user away, and the row drops from yellow to grey with no unread mark. Narrow, and #3181 lists this as an open question, so it needs a maintainer call rather than a code change.

The double-marking and unmark-race failures were looked for specifically and not found. Marking under the user's cursor resolves correctly in every surface traced: the dashboard effect depends on `activeSession?.unread` so it re-fires when the flag flips while the session stays active; `has_dialog()` (`src/tui/home/mod.rs:4875-4916`) does not include the embedded structured view, so `tick_unread_dwell` keeps running and clearing while the TUI agent view is open. The design is "mark, then reactively clear," and #3181's story 4 ("watching in the dashboard, not marked unread") is satisfied by suppression at render plus a clear, not by never marking. That is a defensible reading of the story, and the PR argues for it explicitly; worth confirming you agree.

## Documentation

Slightly heavy in one place, proportionate elsewhere.

- **Trim:** the retargeted test's docstring at `src/tui/home/tests.rs:20657-20684`, ~25 lines, is mostly narration of the change ("This test previously asserted the opposite, on the rationale that..."). That is commit-message material and the commit already carries it. Five lines stating the current contract would do.
- **Keep:** the ~40 comment lines in `acp_event_listener`. They explain the durability requirement, the persist-then-mirror ordering, and why the write is not folded into the sibling save. All non-obvious, all things that otherwise require reading three other files to confirm, and all correct except item 2 above.
- **Missing:** nothing. No user-facing doc mirrors this behavior.

## Merge risk

**Low to moderate.** Blast radius is one boolean on the sidebar; a wrong answer is a missing or spurious dot, never data loss. Reverts cleanly: three commits, no schema, no migration, no released artifact.

The dominant residual risk is issue 4. The path that actually closes #3181 is untested at every tier, so a future refactor of `acp_event_listener` (moving the snapshot, dropping the persist, reordering the mirror) reintroduces the exact bug this PR fixes with a fully green suite. That would arrive from a user, not from CI. The predicate table gives real but misleading confidence here, because the predicate is not the part that was broken.

## CI and merge state

`MERGEABLE/BLOCKED` is caused by exactly one thing: `Playwright (live)` is a required context in the `main` ruleset and it is red. Its shard 2/2 failed on `wizard-scratch-launch.spec.ts::scratch happy path`, a spec this PR cannot have touched (zero `web/` files), and `Playwright (live, shard 1/2)` fails on `main` itself (run `32886028761`). A re-run should clear it. `codecov/project` also shows red but is **not** in the ruleset's required contexts, so it does not gate the merge box; it is the expected artifact of a Rust-only diff against a `web/`-scoped coverage config. No review requirement in the ruleset.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Trim suggested (one ~25-line test docstring)
**Merge risk:** Low to moderate; the fix's own wiring is the one untested path

A real gap, closed the way the codebase already closes this class of problem, in ~27 lines of production code with the ownership question settled rather than papered over. The load-bearing facts were verified rather than taken from the body: the durability requirement (`merge_runtime_fields` carries no `unread`), the single-assignment snapshot in `apply_status_intent`, the unreachability of the poll loop for structured rows, and the reactive-clear paths on both clients. The one thing to ask for before merge is a correction to the body's opening claim, since the TUI demonstrably did mark these rows and the PR itself says so three paragraphs later. A test over `acp_event_listener` is the strongest optional addition.

src/server/mod.rs
```diff
@@ -5630,10 +5640,75 @@ async fn acp_event_listener(state: Arc<AppState>) {
...
+        //
+        // No per-instance lock. A concurrent `PATCH /api/sessions/:id/unread`
+        // clear cannot be lost to this write: `should_mark_acp_unread` requires
+        // the row to be read at decision time, and until the mirror below runs
+        // it still reads as read everywhere, so the reactive clears (the
+        // dashboard's active-session effect, the TUI's unread dwell) cannot have
+        // fired yet. A clear racing this window is therefore a no-op on an
+        // already-read row, and marking after it is the chronologically correct
+        // outcome: the turn did finish afterwards.
+        if let Some(profile) = unread_profile {
```
jerome-benoit, line 5697: [P1] Do not derive this durable mark solely from the lossy broadcast stream. `ACP_CHANNEL_CAPACITY` is 256, and `acp_event_listener` explicitly handles `RecvError::Lagged` by logging and skipping the missing frames. ACP events are persisted before broadcast, but this consumer never replays the missed sequence; `seed_acp_statuses` only restores status at startup. If a `Stopped` frame is among the skipped events, this block never runs and the unread mark is lost permanently. Replay gaps from the durable event store before continuing, or move this projection to a sequenced durable consumer, and cover a lagged `Running -> Stopped` case.

src/server/mod.rs
```diff
@@ -5630,10 +5640,75 @@ async fn acp_event_listener(state: Arc<AppState>) {
...
+                "acp turn-end unread",
+                state.file_watch.clone(),
+                move |instances| {
+                    if let Some(inst) = instances.iter_mut().find(|i| i.id == persist_id) {
+                        inst.mark_unread();
+                    }
+                },
+            )
+            .await;
+            if persisted.is_ok() {
```
jerome-benoit: [P1] Do not mirror based only on `persisted.is_ok()`. `unread_profile` was captured before the await, and `persist_session_update` returns `Ok` even when its closure finds no matching row. A concurrent profile move can therefore make the old-profile write a successful no-op, after which this code marks only RAM and the next reload drops the notification. There is also an inverse clear race: `PATCH /unread` can persist and mirror `false` after this disk write but before this late mirror, leaving disk read and RAM unread. Serialize with the same per-instance/ownership authority, or return whether the owning row was actually mutated, re-resolve/retry when necessary, and only mirror the committed mutation. Please add deterministic barrier-based coverage for the move/no-op and clear interleavings.

jerome-benoit (dismissed):
Requesting changes. Moving structured-session unread ownership into the live ACP path is directionally correct, and I verified that the tmux/TUI writers are gated consistently. However, the new durable projection is built on a lossy broadcast consumer and its persist-then-mirror sequence is not linearizable with the existing clear/profile-move writers. The predicate-only and negative TUI tests do not exercise either failure mode. Please address the two inline findings and add deterministic listener-level coverage.

Independent CI note: the Rust gates are green at this head, while the overall PR is currently blocked by live Playwright and `codecov/project` failures.

src/server/mod.rs
Seluj78: Confirmed and fixed in 74e1826 (tests in 28dc7ec).

You're right, and the asymmetry you're pointing at is the crux: status tolerated a dropped frame because it is level-triggered, so any later event re-derives it. The unread mark is edge-triggered, so a dropped `Stopped` lost it with nothing left to reproduce it. `seed_acp_statuses` only covers the restart case, not a mid-run gap.

The `Lagged` arm now replays. Since ACP events are recorded before they are broadcast, the log is the recovery source: `recover_structured_unread_after_lag` re-derives every structured row from `latest_seed_status_event`, the same query the boot seed uses, and marks any row whose turn ended unobserved.

It deliberately does not reuse `seed_acp_statuses`, because the two differ on both points that matter here:

- Boot must not mark at all. Replaying history would re-mark a row the user had already read, on every restart.
- Boot lifts a stale `Stopped`, since one persisted by a daemon that died mid-turn would otherwise trap the dot grey. Mid-run a `Stopped` in memory is a deliberate stop, so `apply_status_intent`'s guard should keep it.

Covered by `recover_structured_unread_after_lag_marks_only_a_missed_turn_end`: three rows, all with a durable `Stopped` as their latest lifecycle event, so the row is the discriminator rather than the event. Only the structured row still sitting at `Running` is a missed turn-end; the already-reconciled row has no transition left, and the terminal row is never touched.

One thing worth naming since you raised the lag window: my own change widened it slightly, because the persist is a flock'd write on the listener's hot path at turn-end. I left it inline rather than spawning it, since it now has to stay ordered under `instance_lock`, but if you'd rather that write move off the listener entirely I'm happy to do it.

src/server/mod.rs
Seluj78: Conceded on both, fixed in 74e1826 (tests in 28dc7ec).

I had argued this lock was unnecessary, including against the same objection raised during design. That reasoning rested on the mark being invisible until mirrored, which only closes the race where a *clear* is lost to this write, not the reverse one you describe. And it was wrong on its own terms anyway: `persist_session_update` lands the disk write before the mirror, and the TUI reads `unread` from disk on its watcher-driven reload, so the mark is visible inside that window.

Both halves now:

- **Under `instance_lock`**, the same mutex the `PATCH` handler takes, held across the write and the mirror. That leaves only two orderings, and both are correct: a clear before this marks, since the turn genuinely finished afterwards; a clear after this wins, since the user read it afterwards.
- **Only a committed mutation is mirrored.** The write reports whether it actually mutated the owning row, so a stale profile can no longer mark memory off a successful no-op.

Two places I did less than you asked, both deliberate, both arguable:

**No retry on a stale-profile write.** The mark is dropped rather than half-applied. A profile move landing inside a turn-end is rare enough that I did not think a re-resolve loop paid for its own failure surface. Say the word and I'll add it.

**No barrier test for the clear interleaving.** With the lock held across write and mirror, a clear cannot land between them, so the interleaving is unrepresentable rather than merely untested, and a test would be asserting `tokio::sync::Mutex` rather than this code. I did add deterministic coverage for the other one you named: `persist_and_mirror_unread_mirrors_only_a_committed_mutation` drives the profile-move no-op (stale profile seeded empty, so the write succeeds matching nothing) and asserts memory is not marked and the owning row is untouched, then repeats against the real profile. Plus `..._skips_the_mirror_on_a_failed_write` for the failed-write case.

Leaving this thread open rather than resolving it, since those two are my calls rather than yours and you may not agree with either.

Separately, on your listener-level coverage point, which @njbrake's review landed on independently: there is now `acp_event_listener_marks_a_finished_turn_unread_on_disk_and_in_memory`, driving a real `Event::Stopped` frame through the broadcast. I checked it guards something rather than assuming: moving the `old_status` snapshot to after `apply_status_intent` leaves the predicate table green and fails only that test.

CI is green on the current head. The earlier live-Playwright and `codecov/project` reds were a flaky `wizard-scratch-launch` spec plus its knock-on missing LCOV upload; a re-run cleared both without a code change.

njbrake (commented):

_Reviewed at `28dc7ecd`. Push new commits and it will be re-reviewed._

---

## Re-review at `28dc7ecd`

Previously reviewed `a6c218bb`. Rebased onto `origin/main` (`f226e216`).

### Bottom line up front

Every finding raised is either closed or consciously answered, and the two that mattered most, the untested wiring and the wrong root cause in the body, are closed properly rather than nominally. The load-bearing claim was verified by mutation rather than by reading the commit message. Recommend merge.

### What the two new commits do

`30bb08db` answers @jerome-benoit's two P1s: hold `instance_lock` across the write and the mirror, mirror only a write that actually mutated the owning row, and replay the durable event log on `RecvError::Lagged`. `29723d8d` adds the end-to-end listener test and two prose corrections from the last review.

Line accounting for `src/server/mod.rs`, since +566 is a large answer to three findings. Since the previous head: **128 lines of production code**, 64 production comments, 250 test code, 55 test comments, 35 blank. Cumulative against `main`: **151 production code lines**, 121 production comments, 349 test code, 75 test comments. So the production surface grew from ~27 lines to ~151, and about two thirds of the total diff is tests. The 128 new production lines are four things: `instance_lock_in` (~17, mostly the body of `AppState::instance_lock` hoisted verbatim so a non-`AppState` caller can take the same mutex), `persist_and_mirror_unread` (~40), `recover_structured_unread_after_lag` (~45), and the `Lagged` arm hook plus the call-site rewrite (~26).

**No new bookkeeping to keep in sync.** The commit-check is a local `AtomicBool` scoped to one call, not persisted state. `recover_structured_unread_after_lag` derives everything from the existing event log via `latest_seed_status_event`, the same query `seed_acp_statuses` already uses, and holds nothing between calls. Nothing was added that a future writer has to remember to update.

### Finding 4 (untested wiring): closed, and proven

This was the finding that mattered, so `29723d8d`'s subject was not taken at its word. `acp_event_listener_marks_a_finished_turn_unread_on_disk_and_in_memory` (`src/server/mod.rs:8560+`) spawns the real `acp_event_listener` against `build_test_app_state`, waits for `acp_events_tx.receiver_count() > 0` rather than racing the subscription, publishes one real `Event::Stopped` frame, and polls until the **mirror** lands so it cannot abort the listener between the disk write and the mirror and then blame the mirror. Then it asserts disk, memory, and status.

Two mutations were run against the fix.

**Mutation 1, move the `old_status` snapshot to after `apply_status_intent`** (the exact hypothetical raised last round):

```
should_mark_acp_unread_only_on_a_structured_running_to_idle_turn_end ... ok
persist_and_mirror_unread_mirrors_only_a_committed_mutation ... ok
tui_persists_neither_status_nor_unread_for_a_structured_turn_end ... ok
recover_structured_unread_after_lag_marks_only_a_missed_turn_end ... ok
persist_and_mirror_unread_skips_the_mirror_on_a_failed_write ... ok
acp_event_listener_marks_a_finished_turn_unread_on_disk_and_in_memory ... FAILED
  "daemon memory must mirror the mark, so /api/sessions reports it"
```

Five green, one red, and the red one is the new test. The author's claim is accurate.

**Mutation 2, drop `persist_session_update` entirely and mirror memory only** (the pre-#2755 shape):

```
acp_event_listener_marks_a_finished_turn_unread_on_disk_and_in_memory ... FAILED
  "and the mark must be durable, which is the #3181 fix; a memory-only mark is dropped by the next reload"
persist_and_mirror_unread_mirrors_only_a_committed_mutation ... FAILED
persist_and_mirror_unread_skips_the_mirror_on_a_failed_write ... FAILED
recover_structured_unread_after_lag_marks_only_a_missed_turn_end ... FAILED
```

The predicate table and the TUI ownership test stayed green under both mutations, which is exactly the blind spot described last round. It is now covered. Source restored and re-verified clean afterwards.

### "Linearizable and lag-safe": mechanism worked out, not accepted

**The mechanism** is not a disk re-read and not a reordering. It is two things. First, `persist_and_mirror_unread` (`src/server/mod.rs:5936+`) takes `instance_lock` as its first act and holds it across both the write and the mirror. This is genuinely *the same* mutex `PATCH /api/sessions/:id/unread` takes; that handler already did `state.instance_lock(&id).await` at `src/server/api/sessions.rs:4153` on `main`, so the claim is not aspirational. Second, `persist_session_update`'s closure sets an `AtomicBool` when it finds the row, and the mirror is skipped unless that flag is set, so a stale profile can no longer mark memory off a successful no-op.

Re-deriving both races against the new code:

**Finding 2 (persist visible before the mirror, so a clear can land between them and leave memory unread against a read disk) is closed for the realistic path.** The dashboard `PATCH` clear now serializes on the same mutex, so the two writers cannot interleave and the divergence is unreachable. The theoretical remainder is cross-process: the TUI's `tick_unread_dwell` writes `sessions.json` directly and knows nothing about a tokio mutex in the daemon. But that window is now the microseconds between `persist_session_update` returning and a memory write on the next line, and no in-daemon lock can close a cross-process race anyway. This is as closed as it gets. The comment whose reasoning was wrong is deleted; the author conceded it explicitly in the thread and named the same mechanism.

One nuance worth stating rather than filing: the *decision* (`should_mark_acp_unread`, `src/server/mod.rs:5686`) is still made under the `instances` write lock **before** `instance_lock` is taken, so a clear can still land between the decision and the write. The docstring calls that ordering correct because "the turn genuinely finished afterwards." The other reading is defensible (the user read the session after the turn finished, so the mark arguably should not survive), but the window is microseconds and, crucially, disk and memory now always agree either way. That consistency is what "linearizable" was buying, and it is delivered.

**Finding 3 (the memory-trails-disk window silently dropping a mark) is NOT closed**, and that deserves saying plainly rather than letting the commit subject cover it. `should_mark_acp_unread` still reads `inst.unread` from daemon memory, which is rebased from disk only every 2s by `reload_state_instances_from_disk`. Concrete: a row is unread from turn N; the user dwells on it in the TUI, which clears it on disk directly; daemon memory still says `unread = true`; turn N+1 ends inside that window; `!inst.unread` is false and the mark is dropped with nothing left to reproduce it. The `instance_lock` cannot help, because the TUI's clear never passes through the daemon. Note this is a *different* race from the "inverse clear race" @jerome-benoit raised (disk/RAM divergence), which is the one that got fixed. It was non-blocking last round and stays non-blocking: it is the same staleness `decide_passive_transition` has carried for terminal rows since forever, and it needs a turn to complete within 2s of the user reading the row.

### The lag-replay path, reviewed on its own merits

This is new production code that did not exist at the previous head, so it was checked rather than assumed to inherit the earlier review.

The failure worth hunting was replay re-deriving a *stale* status and producing a false mark, and it is not reachable: `latest_seed_status_event` (`src/acp/event_store.rs:1086-1123`) includes `AgentMessageChunk`, `ToolCallStarted`, `ThinkingStarted`, and both approval/elicitation pairs, not just `Stopped`. So a turn resumed by a `ScheduleWakeup` after a `Stopped` has a chunk as its latest lifecycle event and replays as `Running`, and a row blocked on an approval replays as `Waiting`. There is even an existing test for it, `latest_seed_status_event_sees_activity_after_stopped`. The mark then additionally requires an actual `Running -> Idle` transition to have been applied in that pass, so a row already reconciled to `Idle` no-ops. The three-row test covers exactly that discriminator.

Lock ordering checked for the new call graph: `persist_and_mirror_unread` takes `instance_lock` then `instances.write()`, matching `update_session_unread`. The live listener drops its `instances` guard at the end of the decision block before taking `instance_lock`, so there is no hold-and-wait and no inversion. `recover_structured_unread_after_lag` does the same per row.

Two honest scope limits on the replay, neither a request for changes:

- The docstring says it "marks any row whose turn ended unobserved." More precisely it marks any row whose turn **start** was observed and whose end was not. If a lag window swallows both the `UserPromptSent` and the `Stopped`, memory never left `Idle`, `apply_status_intent` no-ops, and the turn stays invisible. A single turn can easily emit more than the 256-frame channel capacity in chunks, so this is not exotic. It is still strictly better than the previous behavior of recovering nothing, and full recovery would need a sequenced durable consumer, which is what @jerome-benoit offered as the alternative and is clearly a separate change.
- `recover_structured_unread_after_lag` is unit-tested as a function, but nothing exercises the one line in the `Lagged` match arm that calls it. Same class of gap as finding 4, several orders of magnitude smaller (one call in a degraded path, plainly visible in the diff). Noted for completeness, not a request.

### Status of the other findings

| # | Finding | Status |
| --- | --- | --- |
| 1 | Daemon's `unread_indicator` flag is startup-frozen | **Answered in the body**, not in code. Manual test step 6 now says "restart `aoe serve`", cites `src/server/mod.rs`'s own "read once at startup" comment, notes it is pre-existing, and states the exact symptom described (the daemon still writes; both render surfaces gate on their own live copy so nothing paints). Documenting rather than fixing the daemon's config refresh is the right scope call. |
| 2 | Persist-before-mirror divergence | **Closed** for the daemon-internal path; see above. |
| 3 | Memory-trails-disk drops a mark | **Open, unchanged.** Non-blocking, and pre-existing in the same form for terminal rows. |
| 4 | Untested listener wiring | **Closed**, mutation-verified. |
| 5 | Persist `.await`ed inline on the ACP event loop | **Unchanged**, and the `Lagged` path adds an inline pass over every structured row (one SQLite point query each, plus a flock'd write per qualifying row) at precisely the moment the listener is already behind. Still non-blocking: point queries are microseconds and qualifying rows are few, but if this ever shows up in a latency trace, this is where to look. |
| 6 | `is_live_target` comment right for the wrong reason | **Closed** in the code comment (`src/tui/home/mod.rs:3313`), which now cites `start_live_send`'s explicit `is_structured()` return and says outright that the absence of a pane is *not* the reason. |
| 7 | `Waiting -> Idle` exclusion | **Unchanged**, still a maintainer sign-off. An agent that crashes or is cancelled with an approval pending drops from yellow to grey with no mark, and the "user is by construction present" justification does not hold in that case. #3181 lists it as an open question. |
| | Body's wrong root cause | **Closed, and well.** The body now names the overstatement, corrects #3181's title, and enumerates the four cases that actually got nothing. This was the one pre-merge ask. |
| | Docstring narrating the change | **Closed.** Trimmed from ~25 lines to 6 stating the contract. |

One stale leftover: the third bullet in the body's "Nothing is lost on the suppression or clear side" list still says "`LiveSendState` needs a tmux pane name and target, so live-send is unreachable for a paneless session", the same wrong reason that was corrected in the code comment. Body-only, cosmetic.

### Test churn check

`src/tui/home/tests.rs` shows -31/+31, checked the same way as last round: filtering the diff to non-comment lines returns **nothing**. The churn is 100% docstring; not one assertion, setup line, or helper changed. Same for the 17 lines in `src/tui/home/mod.rs`. No weakening.

The new tests follow the file's established pattern (`tempfile::tempdir` plus `unsafe set_var("HOME")` plus `#[serial_test::serial]`, verbatim from the neighbouring `flush_passive_transition_*` tests, including the SAFETY comment), so they introduce no new process-global hazard. Four new `#[test]` fns for four genuinely distinct setups; the failed-write one needs `sessions.json` deliberately made a directory, so it cannot be a table row. That is the right side of the house rule.

### Definition of done

| Check | Result |
| --- | --- |
| `cargo clippy --features serve -- -D warnings` (CI's shape) | clean |
| `env -u CLAUDE_CONFIG_DIR cargo test --features serve` | **6423 passed, 0 failed, 2 ignored**, plus 24 green integration targets. No failures at all, including the two that were environmental last round |
| Emdashes in added prose | none |
| `#[allow(dead_code)]` | none |
| Coverage matrix | still correctly N/A; zero `web/` files |
| Docs | still correctly none |

One thing to report so it is not mistaken for a regression: `cargo clippy --features serve --all-targets` (broader than CI, which runs `--all-features` without `--all-targets`) surfaces 5 test-code lints, all in `src/acp/background_agent.rs`, `src/acp/supervisor.rs`, `src/process/macos.rs`, and `src/session/smart_rename.rs`. None of those files is in this PR's three-file diff, so they came in with `main`. Not this PR's problem, and outside CI's gate either way.

### CI and merge state

Every check is green, including `Playwright (live)` (both shards) and `codecov/project`, which were the two red ones last round. `mergeStateStatus: CLEAN`.

The only remaining gate is human: `reviewDecision` is still `CHANGES_REQUESTED` from @jerome-benoit's `2026-08-25T22:42Z` review. The author has replied to both P1s in-thread with what changed and where. That review needs to be resolved or dismissed before the merge box opens; nothing else blocks.

### Merge risk

**Low.** Reverts cleanly: five commits, no schema, no migration, no data backfill, no released artifact. Blast radius is one boolean on a sidebar row.

What changed since the last assessment: the dominant residual risk named then (a future refactor of `acp_event_listener` silently reintroducing #3181 against a green suite) is gone, confirmed by breaking the fix two different ways and watching the new test catch both. What is left is smaller and named: a mark can still be dropped when daemon memory trails a TUI-side clear (finding 3, up to a 2s window), and a lag window that swallows an entire turn is still unrecovered. Both fail *closed*, a missing dot rather than a wrong one, which is the right direction for this feature. Neither has an alert or a log line, so either would arrive as a user report; but "occasionally missed a dot" is a complaint, not an incident.

The new code itself is unreviewed-elsewhere production logic written in one pass, which normally counts as residual risk. Here it is offset: 349 of the 566 new lines are tests, the replay path has a three-row discriminator test, and the one failure mode that would have made replay actively harmful (a stale re-derive producing a false mark mid-turn) was independently checked and is ruled out by `latest_seed_status_event`'s event set.

### Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Merge risk:** Low; the remaining gaps fail closed and are named in the PR body rather than hidden

**LGTM.** The author closed the finding that mattered and proved it the way this review would have, the proof was re-run here and it holds, corrected the root-cause claim in the body without being asked twice, and where he declined to go further (no stale-profile retry, no barrier test for an interleaving the lock made unrepresentable, no daemon config-refresh fix) he said so in the PR rather than quietly. Two items are maintainer-side rather than author-side: dismiss or resolve @jerome-benoit's `CHANGES_REQUESTED` review, and sign off on the `Waiting -> Idle` exclusion, which #3181 raised as an open question and which this PR answers by argument rather than by code.

Issue #3070, opened 2026-07-22
Structured view: opencode subagent (task tool) renders as a bare think card showing only the title, drops description/prompt/result


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

In an opencode structured-view session, when the agent spawns a subagent through opencode's native `task` tool, the dashboard renders it as a bare italic think card showing only the task title (e.g. `Trace clear session resets`) and nothing else. The subagent's `description`, `prompt`, result, and any work it did are all invisible. There is no subagent framing, no card body, no result.

Reported from session `8ffdf4f7e89345d1` (switched codex to opencode, then ran `/aoe-investigate`).

## Evidence from the affected session

From `~/.agent-of-empires/acp_events.db`, the subagent launch (seq 76 to 78):

```
seq 76  ToolCallStarted   { name: "task", kind: "think", args_preview: "{}" }
seq 77  ToolCallUpdated   title: "Trace clear session resets"
                          args_preview: { "description": "Trace clear session resets",
                                          "prompt": "Research only, do not edit files. In the
                                          agent-of-empires repo, trace all code paths related
                                          to `/clear` ..." }
```

The `task` tool call carries a rich `description` + `prompt`, but it never completes on this stream (no `ToolCallCompleted` for its id) and it carries no parent linkage: a full scan of the session's events shows zero `parentToolUseId` / `_meta` linkage in the opencode stream. The agent switch is recorded at seq 10 (`{"AgentSwitched":{"from":"codex","to":"opencode","reason":"manual"}}`).

## Root cause

Two layers, both on the frontend/profile side (opencode's native adapter already emits `kind: "think"` correctly, unlike the aoe-agent case in #1904 where the adapter emits `kind: "other"`).

**1. Opencode `task` routes to `ThinkToolCard`, which shows only the title.**

The opencode profile maps `task` to the think card:

```ts
// web/src/lib/agentProfiles.ts:136-143  (OPENCODE.aliases)
think: ["task"],
```

The wire `kind` is already `"think"`, so `resolveEffectiveKind` returns `"think"` and the dispatcher lands on `ThinkToolCard` (`web/src/components/acp/ToolCards.tsx:307-308`). That card renders the name and nothing else:

```tsx
// web/src/components/acp/ToolCards.tsx:1067-1074
function ThinkToolCard({ tool }: Props) {
  return (
[...]

Issue #1904, opened 2026-06-03
Structured view: aoe-agent subagent Task renders as generic 'other' card, classifyKind misses 'task', no _meta linkage emitted


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

When using a structured-view session with the bundled `aoe-agent` (the multi-provider Vercel AI SDK ACP server), `task` tool calls for subagent spawning show up as generic `"other"` tool cards in the web dashboard instead of being rendered as proper subagent cards with indented child tool calls. The children (bash, read, etc. that the subagent executes) also appear as flat top-level entries with no visual link to their parent Task.

Reported from session `2934dc8c87804072` which ran `aoe-agent` with model `opencode/big-pickle`. Two separate `task` tool calls (at DB seq 1387 and 7004) both landed with `kind: "other"`.

## Evidence from the affected session

From the structured-view events DB (`cockpit_events.db`):

```
seq=1387  ToolCallStarted  name="task"  kind="other"
seq=1390  ToolCallStarted  name="bash"  kind="execute"
  (child Bash has no parent_tool_call_id — no _meta linkage)
...
seq=1398  ToolCallStarted  name="task"  kind="other"  completed
```

The parent Task tool call carries `kind: "other"`, and the child Bash tool call has no `parent_tool_call_id`. Both failures trace back to the same adapter code.

## Root cause

Two independent defects in the `aoe-agent` adapter at `cockpit-worker/aoe-agent/src/index.ts`:

**1. `classifyKind` at line 292-303 only recognizes "Read", "Write", "Bash". Everything else, including "task", falls through to `"other"`:**

```typescript
function classifyKind(toolName: string): acp.ToolKind {
  switch (toolName) {
    case "Read":    return "read";
    case "Write":   return "edit";
    case "Bash":    return "execute";
    default:        return "other";
  }
}
```

claude-agent-acp at `tools.js:18-34` classifies Task as `kind: "think"`, so the AOE frontend dispatches it to `ThinkToolCard`. The `aoe-agent` emits `kind: "other"` instead, which lands in `GenericToolCard`.

[...]

PR #3533, opened 2026-08-25
fix(acp): classify aoe-agent task calls as think and drop AOE_AGENT's false claude claims

Seluj78:

## Description

Partial progress on #1904. Two of its three slices, A and C; slice B stays out because it is `blocked:upstream`.

`aoe-agent`'s `classifyKind` only recognised `Read`, `Write` and `Bash`, so every other tool name fell through to `kind: "other"` and landed on the generic tool card. Models trained against harnesses that ship a subagent tool call `task` anyway, which is how two of them landed in session `2934dc8c87804072`. `claude-agent-acp` and `opencode` >= 1.16.0 both report `think` for that name, so `aoe-agent` now matches them and the call renders on the think card.

Worth being precise about what that does and does not fix: `buildTools` registers exactly Read, Write and Bash, so the AI SDK cannot execute a `task` call at all. No subagent runs, and no child tool calls stream. This is a rendering fix. Real subagent support needs a `task` tool in the palette plus slice B's `_meta` linkage, and B is gated on [anomalyco/opencode#30557](https://github.com/anomalyco/opencode/issues/30557) per the issue thread.

Slice C is the larger half. Both `AOE_AGENT` profiles, server and frontend, inherited Claude's profile wholesale (`..CLAUDE` / `...CLAUDE`). Read against the adapter source, almost every inherited field was false: it emits no `_meta` on any notification, has no ExitPlanMode or ScheduleWakeup tool, sends no heartbeat keepalives, and its `session/set_mode` is a stub that returns `{}` and gates nothing. So the structured view was advertising subagent indentation it could never render, the Claude specialised cards, and a Default/Plan/AcceptEdits/Yolo mode picker whose selections the adapter silently discarded. Every field is now stated explicitly with the source fact behind it, so a later `..CLAUDE` edit cannot quietly re-inherit a claim.

Two decisions inside slice C worth flagging for review:

- **`clear_requires_driven_reset` flips to `true`.** The issue left this field "tbd". The adapter handles no slash command at all, so a forwarded `/clear` reached the model as an ordinary prompt: the conversation kept its whole history while AoE drew a clear divider. That is codex-acp's defect exactly, so it takes codex-acp's remedy. One hole remains, commented at the field rather than fixed here: `session/load` reseeds from `${AOE_ARTIFACT_DIR}/transcript.jsonl`, which still holds the pre-clear turns, so a worker restart after a clear resurrects the cleared context. Fixing that means truncating the transcript on reset, which is a separate change.
- **`yolo_mode_id` drops to `None`, and this is not a security loosening.** `bypassPermissions` is still in `TRUSTED_MODE_TABLE` in `src/plugin/automation_policy.rs`, so it classifies `Unattended` regardless of the profile. The profile field only stopped claiming a bypass mode the adapter does not advertise.

`subagentToolNames` stays empty on purpose. That field (#3070) renders a childless subagent card for an off-protocol subagent that really ran, which is opencode's case. `aoe-agent` has no `task` tool, so nothing runs and the card would be inventing a subagent.

**Separate follow-up I did not touch, but noticed while reading:** `is_reviewed("aoe-agent")` is `true`, which classifies an omitted mode as `Interactive` on the assumption the adapter prompts for approval. `aoe-agent` never sends `session/request_permission`; AoE's own `fs/*` and `terminal/*` handlers gate it server-side instead. That is a security-policy call in #2897's territory, not a rendering fix, so it wants its own issue.

Refs #1904

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view session with `--agent aoe-agent` and a non-Claude model, prompt it into something multi-step, and confirm any `task` call the model emits renders as a think card rather than a generic "other" card.
- [ ] In the same session, open the mode picker: it should now be absent rather than offering Default / Plan / Accept-edits / Yolo, since the adapter advertises no modes and ignored every selection.
- [ ] Send `/clear` in an `aoe-agent` session, then ask the agent about something from before the clear. It should not remember it. Before this change it did, while the timeline showed a clear divider.
- [ ] Known limitation to confirm, not a regression: after that `/clear`, stop the worker (`aoe acp stop <session>`) and prompt again. The pre-clear context comes back, because the adapter reseeds from its on-disk transcript. Commented at the profile field.
- [ ] Confirm a Claude session is untouched: subagent indentation, TodoWrite / Skill cards, and the mode picker all still work as before.

## Checklist

`cargo fmt`, `cargo clippy --features serve --all-targets`, and `cargo test --features serve` are clean, except for one pre-existing macOS-only failure unrelated to this diff: `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` compares a `tempfile::tempdir()` path against a canonicalized one, and macOS resolves `/var` to `/private/var`. Verified failing at the base commit `fb0e7af` in a detached worktree, and this diff does not touch `src/git/`. Web side: `npx vitest run` (3716 pass), `npm run format:check`, `npm run lint`, `npx tsc -b` all clean.

No `docs/` change. The feature matrix in `docs/structured-view.md` puts `aoe-agent` in the "Other ACP" column, which is already marked as either adapter-dependent or unsupported for every capability this PR turns off, so nothing there becomes wrong.

## Test Coverage Analysis

- [x] Skipped a Playwright test, because: the change is in a pure data table (`web/src/lib/agentProfiles.ts`), covered by a Vitest regression test in `agentProfiles.test.ts` that asserts the whole capability set and fails on a re-spread of `CLAUDE`. `coverage-matrix.json` already maps that file to that spec, and `validate-coverage-matrix.mjs` passes. A live Playwright spec proving the aoe-agent mode picker is gone would need a fake adapter advertising no modes, which is worth having but is a harness addition rather than part of this fix.

**TUI / CLI (e2e test)**

Regression tests, each verified red before the fix and green after:

- `acp-worker/aoe-agent/test/toolKind.test.ts`: table over the tool palette plus `task` / `Task`. `classifyKind` moved into its own `src/toolKind.ts` to get a test seam at all, since `index.ts` calls `main()` at module scope and importing it from a test would connect to stdio. Note this package's `npm test` is not wired into CI today (neither is its existing `transcript.test.ts`); run it with `cd acp-worker/aoe-agent && npm test`.
- `src/acp/agent_profiles.rs`: three existing tests pinned the old claim and now pin the new one (`capability_flags_only_set_for_claude_family`, `yolo_mode_id_is_adapter_specific`, `clear_requires_driven_reset_for_codex_and_claude`). The last one carried a doc comment arguing aoe-agent must *not* inherit Claude's answer here, so it is rewritten to state the new reason rather than just moved. Added an assertion that Claude's own `_meta` namespace does not resolve for aoe-agent.
- `web/src/lib/agentProfiles.test.ts`: asserts the full capability object, `parentMetaNamespaces`, `specialTitles`, and that `isSubagentToolName("task", …)` is false.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction. I scoped the PR to slices A and C and made the two judgment calls it surfaced (extracting a test seam for `classifyKind`, and driving the clear reset rather than dropping the alias). I reviewed each commit and ran the manual test steps.

## Summary

- Fixed `aoe-agent` rendering for `task` calls, regardless of letter case.
- Failed think calls now display their error messages.
- Removed unsupported Claude capabilities from the server and web `AOE_AGENT` profiles.
- Added server-driven `/clear` reset handling.
- Added regression tests and CI coverage for the adapter.

## Benefits

- Tool activity renders correctly.
- Unavailable tools no longer fail silently.
- The profile exposes only supported capabilities.
- Tests reduce the risk of future regressions.

## Technical details

- `task` calls map to the `think` kind for rendering only.
- `subagentToolNames` remains empty because `aoe-agent` does not execute subagents.
- `/clear` uses a driven reset and documents transcript-reload limitations.
- Slice B remains blocked upstream.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -272,6 +272,14 @@ jobs:
       - name: Run Vitest with coverage
         working-directory: web
         run: npm run test:unit -- --coverage --reporter=default --reporter=junit --outputFile.junit=test-report.junit.xml
+      # The bundled aoe-agent ACP adapter is a separate npm package with its
+      # own node:test suite (transcript round-tripping, tool-kind mapping).
+      # Nothing ran it before, so both of its test files could rot unnoticed;
+      # it rides this job because Node is already set up here. Seconds to run,
+      # and it is deliberately not part of the coverage upload below.
+      - name: aoe-agent adapter tests
+        working-directory: acp-worker/aoe-agent
+        run: npm ci && npm test
       # Vitest's v8 coverage reporter emits LCOV paths relative to the
       # vitest cwd (web/), so SF: lines read "SF:src/...". Codecov
       # matches file paths against the PR diff, which uses repo-root
--- a/acp-worker/aoe-agent/src/index.ts
+++ b/acp-worker/aoe-agent/src/index.ts
@@ -22,6 +22,7 @@ import { openai } from "@ai-sdk/openai";
 import { google } from "@ai-sdk/google";
 import { z } from "zod";
 import { appendTurn, loadTranscript } from "./transcript.ts";
+import { classifyKind } from "./toolKind.ts";
 
 const DEFAULT_MODEL = "claude-opus-4-7";
 
@@ -266,19 +267,6 @@ function buildTools(sessionId: string, client: acp.AgentContext) {
       },
     }),
   };
-}
-
-function classifyKind(toolName: string): acp.ToolKind {
-  switch (toolName) {
-    case "Read":
-      return "read";
-    case "Write":
-      return "edit";
-    case "Bash":
-      return "execute";
-    default:
-      return "other";
-  }
 }
 
 function serialiseToolOutput(output: unknown): Record<string, unknown> {
--- a/acp-worker/aoe-agent/src/toolKind.ts
+++ b/acp-worker/aoe-agent/src/toolKind.ts
@@ -6,22 +6,25 @@
 
 import type * as acp from "@agentclientprotocol/sdk";
 
+// Matched case-insensitively, the way opencode's mapper normalises before
+// its switch. The names below are how `buildTools` registers them, but a
+// model picks the casing, and a mismatch here only costs a wrong card.
 export function classifyKind(toolName: string): acp.ToolKind {
-  switch (toolName) {
-    case "Read":
+  switch (toolName.toLowerCase()) {
+    case "read":
       return "read";
-    case "Write":
+    case "write":
       return "edit";
-    case "Bash":
+    case "bash":
       return "execute";
     // `task` is not in `buildTools`, but models trained against harnesses
     // that do have a subagent tool call it anyway; two such calls landed in
-    // the wild (#1904). The AI SDK cannot execute it, so no subagent runs.
-    // This only fixes how the call renders: `think` matches what
-    // claude-agent-acp and opencode >=1.16.0 report for the same name,
-    // which puts it on the think card instead of the generic one.
+    // the wild (#1904). The AI SDK answers each one with a NoSuchToolError,
+    // so no subagent runs and the call always fails. `think` matches what
+    // claude-agent-acp and opencode >=1.16.0 report for the same name; the
+    // think card falls through to the error body on a failure, so the
+    // NoSuchToolError text stays visible.
     case "task":
-    case "Task":
       return "think";
     default:
       return "other";
--- a/acp-worker/aoe-agent/src/toolKind.ts
+++ b/acp-worker/aoe-agent/src/toolKind.ts
@@ -1,0 +1,29 @@
+/**
+ * Tool-name to ACP `ToolKind` mapping for the structured view's card
+ * dispatch. Its own module so it is testable: `index.ts` calls `main()` at
+ * module scope, so importing it from a test would connect to stdio.
+ */
+
+import type * as acp from "@agentclientprotocol/sdk";
+
+export function classifyKind(toolName: string): acp.ToolKind {
+  switch (toolName) {
+    case "Read":
+      return "read";
+    case "Write":
+      return "edit";
+    case "Bash":
+      return "execute";
+    // `task` is not in `buildTools`, but models trained against harnesses
+    // that do have a subagent tool call it anyway; two such calls landed in
+    // the wild (#1904). The AI SDK cannot execute it, so no subagent runs.
+    // This only fixes how the call renders: `think` matches what
+    // claude-agent-acp and opencode >=1.16.0 report for the same name,
+    // which puts it on the think card instead of the generic one.
+    case "task":
+    case "Task":
+      return "think";
+    default:
+      return "other";
+  }
+}
--- a/acp-worker/aoe-agent/test/toolKind.test.ts
+++ b/acp-worker/aoe-agent/test/toolKind.test.ts
@@ -10,7 +10,10 @@ test("maps the tool palette and the unregistered task tool", () => {
     // Regression for #1904: `task` used to fall through to "other", which
     // rendered a generic card instead of the think card.
     ["task", "think"],
+    // The model picks the casing, so every arm is matched case-insensitively.
     ["Task", "think"],
+    ["TASK", "think"],
+    ["bash", "execute"],
     // Anything the adapter does not know still falls through.
     ["WebFetch", "other"],
     ["", "other"],
--- a/acp-worker/aoe-agent/test/toolKind.test.ts
+++ b/acp-worker/aoe-agent/test/toolKind.test.ts
@@ -1,0 +1,21 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { classifyKind } from "../src/toolKind.ts";
+
+test("maps the tool palette and the unregistered task tool", () => {
+  const cases: [string, string][] = [
+    ["Read", "read"],
+    ["Write", "edit"],
+    ["Bash", "execute"],
+    // Regression for #1904: `task` used to fall through to "other", which
+    // rendered a generic card instead of the think card.
+    ["task", "think"],
+    ["Task", "think"],
+    // Anything the adapter does not know still falls through.
+    ["WebFetch", "other"],
+    ["", "other"],
+  ];
+  for (const [name, expected] of cases) {
+    assert.equal(classifyKind(name), expected, name);
+  }
+});
--- a/src/acp/agent_profiles.rs
+++ b/src/acp/agent_profiles.rs
@@ -293,19 +293,43 @@ pub const PRIME_AGENT: AgentProfile = AgentProfile {
     yolo_mode_id: None,
 };
 
-/// aoe's own multi-provider agent. Treated as Claude-equivalent
-/// for now (Vercel AI SDK 7 with Claude as one of the providers); the
-/// claude_capabilities subset is the safest reference until aoe-agent
-/// has its own conventions.
+/// aoe's own multi-provider agent (Vercel AI SDK 7), at
+/// `acp-worker/aoe-agent/src/index.ts`. Every field below is read off that
+/// source, which is a much smaller surface than Claude's: three tools
+/// (Read, Write, Bash), no `_meta` on any event, no plan mode, no
+/// ScheduleWakeup, no heartbeats, and a `session/set_mode` that is a stub
+/// returning `{}`. It used to inherit `..CLAUDE`, which claimed all of
+/// those; the structured view then advertised subagent indentation and a
+/// mode picker the adapter cannot honor. See #1904.
 pub const AOE_AGENT: AgentProfile = AgentProfile {
     key: "aoe-agent",
-    // Deliberately NOT inherited from CLAUDE: aoe-agent is multi-provider, and
-    // nobody has checked whether its clear rotates an observable session id the
-    // way claude-agent-acp fails to. Driving a reset on an adapter that already
-    // handles the alias natively would double-reset it, so it stays on the
-    // forward path until someone verifies it.
-    clear_requires_driven_reset: false,
-    ..CLAUDE
+    // The adapter sets no `_meta` on its tool_call notifications at all, so
+    // there is no namespace to look up and child tool calls cannot be linked
+    // to a parent. Populate this only once the adapter emits linkage.
+    parent_meta_namespaces: &[],
+    clear_aliases: &["/clear"],
+    // Same defect class as codex-acp: the adapter has no clear handler of any
+    // kind, so a forwarded `/clear` is answered as an ordinary prompt and the
+    // model keeps its whole history while AoE draws a clear divider. Driving
+    // the reset via `session/new` mints an id we can persist and does reset
+    // the in-memory history.
+    //
+    // Known hole: the adapter reseeds from `${AOE_ARTIFACT_DIR}/transcript.jsonl`
+    // on `session/load`, and that file still holds the pre-clear turns, so a
+    // worker restart after a clear resurrects the cleared context. Fixing that
+    // means truncating the transcript on reset, tracked separately.
+    clear_requires_driven_reset: true,
+    // No ExitPlanMode tool and no mode channel; no ScheduleWakeup or cron
+    // tools. Synthesising Plan or WakeupScheduled events here would fire on
+    // nothing.
+    supports_exit_plan_mode: false,
+    supports_wakeup_tools: false,
+    emits_heartbeat_keepalives: false,
+    // `session/set_mode` is `() => ({})`: it accepts any id and changes
+    // nothing, and the adapter advertises no modes. Claiming Claude's
+    // `bypassPermissions` made YOLO look applied when nothing was gated by
+    // the adapter at all, so keep it a documented no-op.
+    yolo_mode_id: None,
 };
 
 /// Permissive default for unknown registry keys: no claude-specific
@@ -422,7 +446,9 @@ mod tests {
             resolve("claude-code").yolo_mode_id,
             Some("bypassPermissions")
         );
-        assert_eq!(resolve("aoe-agent").yolo_mode_id, Some("bypassPermissions"));
+        // aoe-agent's `session/set_mode` is a stub that accepts any id and
+        // gates nothing, so it advertises no bypass mode (#1904).
+        assert_eq!(resolve("aoe-agent").yolo_mode_id, None);
         // Regression for #1142 and the @agentclientprotocol/codex-acp
         // migration: codex's bypass preset is `agent-full-access`, not
         // Claude's `bypassPermissions` or the old Zed adapter's `full-access`.
@@ -470,19 +496,18 @@ mod tests {
     /// text-forward path leaves the new conversation unresumable across a
     /// worker restart. Both need AoE to mint the id itself.
     ///
-    /// `aoe-agent` must NOT inherit claude's answer here even though it
-    /// inherits the rest of the profile via `..CLAUDE`: it is a different
-    /// multi-provider adapter whose clear semantics nobody has verified. The
-    /// assertion below is what stops a future `..CLAUDE` edit from silently
-    /// flipping it. Same standard for the unverified `/new` mappings
-    /// (opencode, omp, kimi), which keep the old behavior until observed.
+    /// `aoe-agent` lands in the same bucket as codex, for codex's reason
+    /// rather than claude's: its adapter source handles no slash command at
+    /// all, so a forwarded `/clear` reaches the model as an ordinary prompt
+    /// and the conversation keeps its full history (#1904). Same standard for
+    /// the unverified `/new` mappings (opencode, omp, kimi), which keep the
+    /// old behavior until observed.
     #[test]
     fn clear_requires_driven_reset_for_codex_and_claude() {
-        for profile in [&CODEX, &CLAUDE, &CLAUDE_CODE] {
+        for profile in [&CODEX, &CLAUDE, &CLAUDE_CODE, &AOE_AGENT] {
             assert!(profile.clear_requires_driven_reset, "{}", profile.key);
         }
         for profile in [
-            &AOE_AGENT,
             &OPENCODE,
             &GEMINI,
             &VIBE,
@@ -527,6 +552,18 @@ mod tests {
         // Opencode's parent linkage convention is unverified; even if the
         // wire carries the value, we don't claim it until observed.
         assert!(OPENCODE.parent_tool_use_id_from_meta(&Some(meta)).is_none());
+
+        // aoe-agent emits no `_meta` on any tool_call notification, so even
+        // claude's own namespace must not resolve for it (#1904). It used to,
+        // via `..CLAUDE`, which promised linkage the adapter never sends.
+        let mut claude_meta = serde_json::Map::new();
+        claude_meta.insert(
+            "claudeCode".to_string(),
+            serde_json::json!({ "parentToolUseId": "tc-parent-7" }),
+        );
+        assert!(AOE_AGENT
+            .parent_tool_use_id_from_meta(&Some(claude_meta))
+            .is_none());
     }
 
     #[test]
@@ -571,10 +608,13 @@ mod tests {
 
     #[test]
     fn capability_flags_only_set_for_claude_family() {
-        for profile in [&CLAUDE, &CLAUDE_CODE, &AOE_AGENT] {
+        for profile in [&CLAUDE, &CLAUDE_CODE] {
             assert!(profile.supports_exit_plan_mode);
             assert!(profile.supports_wakeup_tools);
         }
+        // `aoe-agent` sits with the others despite bundling Claude as one of
+        // its providers: the adapter's own tool palette is Read/Write/Bash,
+        // with no ExitPlanMode and no ScheduleWakeup to synthesise from (#1904).
         for profile in [
             &CODEX,
             &OPENCODE,
@@ -584,6 +624,7 @@ mod tests {
             &OMP,
             &KIMI,
             &PRIME_AGENT,
+            &AOE_AGENT,
             &DEFAULT,
         ] {
             assert!(!profile.supports_exit_plan_mode, "{}", profile.key);
--- a/src/acp/agent_profiles.rs
+++ b/src/acp/agent_profiles.rs
@@ -612,8 +612,8 @@ mod tests {
             assert!(profile.supports_exit_plan_mode);
             assert!(profile.supports_wakeup_tools);
         }
-        // `aoe-agent` sits with the others despite bundling Claude as one of
-        // its providers: the adapter's own tool palette is Read/Write/Bash,
+        // `aoe-agent` sits with the non-Claude group despite bundling Claude
+        // as one of its providers: the adapter's tool palette is Read/Write/Bash,
         // with no ExitPlanMode and no ScheduleWakeup to synthesise from (#1904).
         for profile in [
             &CODEX,
--- a/web/src/components/acp/ToolCards.tsx
+++ b/web/src/components/acp/ToolCards.tsx
@@ -1103,7 +1103,21 @@ function FetchToolCard({ tool, result }: Props) {
 
 /* ── think ──────────────────────────────────────────────────────── */
 
-function ThinkToolCard({ tool }: Props) {
+function ThinkToolCard({ tool, result }: Props) {
+  // A think call's substance normally lives somewhere other than the card:
+  // in streamed child tool calls, or in a returned report. The quiet
+  // one-line card is right for that. A *failed* one inverts it, the error
+  // is all there is, and this card has nowhere to put it. Fall through to
+  // the generic card, which auto-expands on error and prints the text.
+  //
+  // aoe-agent makes this the common case rather than an edge: `task` is not
+  // in its palette, so the AI SDK answers every call with a NoSuchToolError
+  // naming the tools that do exist, which is the one message a user needs
+  // to see. A claude or opencode Task can fail too, and was equally silent
+  // before this branch. See #1904.
+  if (statusFor(result) === "err") {
+    return <GenericToolCard tool={tool} result={result} />;
+  }
   return (
     <div className="my-1 flex items-center gap-2 px-3 py-1 text-xs italic text-text-muted">
       <Sparkles className="h-3 w-3 text-text-dim" />
--- a/web/src/components/acp/ToolCards.tsx
+++ b/web/src/components/acp/ToolCards.tsx
@@ -311,7 +311,22 @@ function renderToolCard(tool: ToolCall, result: ActivityRow | undefined, profile
     case "fetch":
       return <FetchToolCard tool={tool} result={result} />;
     case "think":
-      return <ThinkToolCard tool={tool} result={result} />;
+      // A think call's substance normally lives somewhere other than the card:
+      // in streamed child tool calls, or in a returned report. The quiet
+      // one-line ThinkToolCard is right for that. A *failed* one inverts it,
+      // the error is all there is, and that card renders no error body, so
+      // route failures to the generic card instead.
+      //
+      // aoe-agent makes this the common case rather than an edge: `task` is
+      // not in its palette, so the AI SDK answers every call with a
+      // NoSuchToolError naming the tools that do exist, which is the one
+      // message a user needs to see. A claude or opencode Task can fail too,
+      // and was equally silent before this branch. See #1904.
+      return statusFor(result) === "err" ? (
+        <GenericToolCard tool={tool} result={result} />
+      ) : (
+        <ThinkToolCard tool={tool} result={result} />
+      );
     default:
       return <GenericToolCard tool={tool} result={result} />;
   }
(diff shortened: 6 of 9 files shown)


njbrake (commented):

_Reviewed at `0283883b`. Push new commits and it will be re-reviewed._

---

## Summary

Two independent slices of #1904. **Slice C** (the larger half) replaces `AOE_AGENT`'s `..CLAUDE` / `...CLAUDE` spread with explicit fields in both `src/acp/agent_profiles.rs` and `web/src/lib/agentProfiles.ts`, each annotated with the adapter fact behind it. **Slice A** adds `case "task": case "Task": return "think"` to a newly extracted `acp-worker/aoe-agent/src/toolKind.ts`.

Slice C is careful, well-evidenced work and is ready as-is. Slice A ships a concrete regression, detailed below.

## Design alignment

**Aligned.** Compared against the precedent set by `CODEX` (`src/acp/agent_profiles.rs:157-175`) and `OMP` (`:245-256`), which are the two profiles that state every field explicitly with a per-field justification comment rather than inheriting. Slice C reshapes `AOE_AGENT` into exactly that shape, and the `clear_requires_driven_reset: true` flip reuses codex's existing mechanism (`PromptDisposition::ResetContext` into `Supervisor::reset_session_context`, `src/acp/supervisor.rs:1359` and `:2645`) rather than adding a second reset path. No new abstraction; `toolKind.ts` has two real consumers (`index.ts` and the test).

## What was verified rather than taken from the PR body

Every load-bearing claim in the description checks out:

- **claude-agent-acp reports `think` for `Task`.** `@agentclientprotocol/claude-agent-acp@0.59.0` `dist/tools.js:19-24`: `case "Agent": case "Task": ... kind: "think"`. Note it matches **capital `Task` only**; lowercase `task` falls to the default `other` there.
- **opencode reports `think` for `task` from >= 1.16.0, and the boundary is exact.** `packages/opencode/src/acp/tool.ts:63-64` at tag `v1.16.0` has `case "task": return "think"` (case-insensitive via `toLocaleLowerCase()`); the same file at `v1.15.13` does not, and the minified mapper in a local `opencode 1.14.25` binary has no `task` arm either. Landed in `fc62b3dc` "fix(acp): classify task as think (#30565)". `docs/structured-view.md:69` already documents "opencode >=1.16.0 ... classifies ... `task` as `think`", so the PR is consistent with the repo's own record. Users on older opencode get `other`, unchanged by this PR, since this only touches `aoe-agent`.
- **Every slice-C claim about the adapter is true**, read off `acp-worker/aoe-agent/src/index.ts`: `session/set_mode` is literally `() => ({})` (`:352`), no `_meta` on any `session/update`, no ExitPlanMode or ScheduleWakeup in `buildTools` (`:199-266`), no heartbeats, no slash-command handling anywhere, and `session/load` reseeds from `${AOE_ARTIFACT_DIR}/transcript.jsonl` (`:330-350`), so the documented "worker restart after a clear resurrects the context" hole is real and correctly scoped.
- **`yolo_mode_id: None` is not a security loosening**, and for a stronger reason than the PR gives: `bypassPermissions` is in `TRUSTED_MODE_TABLE` as `Unattended` regardless of profile (`src/plugin/automation_policy.rs:59`, and the `*class == Unattended || reviewed` branch at `:114` bypasses the reviewed gate), *and* aoe-agent's tools are gated by `res.fs_policy` path roots in `handle_write_text_file` / `handle_create_terminal` (`src/acp/acp_client.rs:9427`, `:9452`), never by an approval mode. There was nothing for a bypass mode to bypass.
- **The mode picker really does disappear.** `resolveModeChannel` (`web/src/lib/modeChannel.ts:105-140`) returns `null` when there is no `category:"mode"` config option, no `availableModes`, and `allowLegacyFallback` is false. aoe-agent advertises none of the three. Correct, and user-visible.
- **Rust/TS mirror.** The two files agree on every overlapping field (`parent_meta_namespaces`/`parentMetaNamespaces` both `[]`, `supports_wakeup_tools`/`capabilities.wakeup` both false, `emits_heartbeat_keepalives`/`heartbeatKeepalives` both false). Nothing enforces this: the only cross-file test is `src/agents.rs:2033 test_lifecycle_facts_stay_synced_with_ts_mirror`, which reads the TS file but only checks gemini *lifecycle strings*, not capability data. Pre-existing gap; the two edits here are consistent, so not a finding against this PR.
- **The extraction is faithful.** `toolKind.ts` reproduces the three original arms verbatim, the import is wired at `index.ts:25`, and `npm test` in `acp-worker/aoe-agent` passes 6/6 (1 new plus 5 existing).

## Blocking: slice A routes an always-failing call onto a card that renders nothing

The PR's own investigation established that `buildTools` registers only Read/Write/Bash, so a `task` call cannot execute. What it did not follow through on is **what the user sees instead**.

The adapter's actual wire output was reproduced against the pinned `ai@7.0.66`, with `task` unregistered:

```
PART: tool-call  {"toolName":"task","toolCallId":"call-1",
                  "error":"AI_NoSuchToolError: Model tried to call unavailable tool 'task'. Available tools: Read."}
PART: tool-error {"toolName":"task", ... same error}
```

So the adapter emits `tool_call` (which is why `classifyKind` runs at all) and then `tool_call_update { status: "failed", rawOutput: { error: "AI_NoSuchToolError: ..." } }` via `index.ts:126-138`. Every `task` call is a failed call, always.

`ThinkToolCard` (`web/src/components/acp/ToolCards.tsx:1108-1115`) destructures only `tool` and drops `result` on the floor: it renders a sparkle icon and `{tool.name || "thinking…"}`, nothing more.

Both paths were rendered through the real `ToolCard` dispatcher under `toolKey="aoe-agent"` with a `tool_error` result carrying the NoSuchToolError text. Captured `container.textContent`:

- **before this PR** (`kind: "other"`, so `GenericToolCard`, auto-expanded because `useToolCardExpansion` at `:388` forces `open` on `status === "err"`):
  `"othertask1.0sfailedtool failedAI_NoSuchToolError: Model tried to call unavailable tool 'task'. Available tools: Read, Write, Bash.Show attempted actioninput{\"description\":\"delegate\"}"`
- **after this PR** (`kind: "think"`):
  `"task"`

Concrete failure scenario: a user on `aoe-agent` with a non-Claude model asks for something multi-step. The model reaches for `task`, gets a hard error, flails, and the turn goes sideways. Today they see a red card naming the exact cause and the three tools that *are* available, which is precisely the signal that would lead someone to file the real fix. After this PR they see a grey italic sparkle that says `task` and nothing else. If the call lands inside a `ToolGroupCard`, the header still shows "· 1 error" (`ToolCards.tsx:1447`), but expanding the group renders the same empty think card, so the error text is unreachable either way.

The consistency-with-upstream argument is real but does not transfer: for claude-agent-acp and opencode >=1.16.0, the `Task`/`task` call *runs* and its substance lives in streamed children or a returned report, so a quiet parent card is right. For `aoe-agent` the substance **is** the error, and `think` erases it. Worth noting the branch name is `fix/aoe-agent-subagent-task-generic-other-card`: the change is doing what #1904 literally asked for, but #1904 was written believing a subagent actually ran (`seq=1390 ToolCallStarted name="bash"` was read as a streamed child; it is far more likely the model's fallback after the NoSuchToolError). This PR is the one that disproved that premise, and slice A is the part that did not get re-derived from it.

Two ways out, either is fine:

1. **Preferred, and generally useful:** make `ThinkToolCard` fall through to `GenericToolCard` (or render `CardChrome` plus `ToolErrorBody`) when `statusFor(result) === "err"`. That closes a latent hole for *every* agent's think card, since a failed claude `Task` is silent today too, and it keeps slice A's consistency argument intact.
2. Drop slice A, ship slice C alone, and let the `task` card change ride with slice B when a real `task` tool exists.

No fix was staged, since which of the two to take is a maintainer call.

## Non-blocking

- `classifyKind` enumerates `"task"` and `"Task"`; opencode's equivalent normalises with `toLocaleLowerCase()` first. A `switch (toolName.toLowerCase())` would cover `TASK` and any other casing a model emits, at zero cost. `toolKind.ts:9`.
- `acp-worker/aoe-agent`'s `npm test` still is not in CI: `.github/workflows/tests.yml:56-63` only runs `acp-worker/test-shim`. The PR flags this honestly, but it is now adding a *second* test file to a package whose suite nothing runs. A four-line job addition belongs in this PR rather than a follow-up, or both files rot.
- The new Rust test comment "`aoe-agent` sits with the others" (`agent_profiles.rs`, in `capability_flags_only_set_for_claude_family`) reads ambiguously; "sits with the non-Claude group" would be clearer.
- Aside, **unverified and pre-existing, not a finding against this PR**: nothing in the repo appears to produce `${aoe_data_dir}/acp-worker/dist/aoe-agent` (the registry command at `src/acp/agent_registry.rs:189`), no build script, no xtask, no release or flake step, and `command_present` (`src/cli/acp.rs:772`) returns `true` unconditionally for any `${...}` command, so `aoe acp doctor` reports it present regardless. It was absent from the reviewing environment's app dir. If that is genuinely unpackaged it deserves its own issue, and it lowers the practical blast radius of the finding above. The packaging step may simply not have been found.

- **Test shape is exemplary.** `toolKind.test.ts` is one test with a 7-row table (`Read`/`Write`/`Bash`/`task`/`Task`/`WebFetch`/empty), including the `#1904` regression comment on its row. The Rust side *modified three existing tests* rather than adding new `#[test]` fns, which is exactly what AGENTS.md asks for. `web/src/lib/agentProfiles.test.ts` asserts the whole capability object, so a re-spread of `CLAUDE` fails it loudly.
- **Coverage matrix:** `web/tests/validate-coverage-matrix.mjs` passes; `acp.omp-profile` and `acp.opencode-todos` already map `web/src/lib/agentProfiles.ts` to `src/lib/agentProfiles.test.ts`, so no new entry is needed. The Playwright skip justification holds: a live spec proving the picker is gone needs a fake adapter advertising no modes, which is harness work.
- **No emdashes in any added prose.** No dead code. `scripts/check-nix-embedded-assets.py` passes (41 assets OK).
- **Docs:** the PR's "nothing becomes wrong" claim holds. `docs/structured-view.md:51-69` puts `aoe-agent` in the "Other ACP" column, already showing dashes or "depends" for every capability this turns off, and it already documents the opencode 1.16.0 boundary. `docs/development/internals/structured-view.md#agent-profiles` describes the mechanism generically and enumerates no per-agent facts.

## Local results (rebased onto `origin/main` at `a0137269`)

- `cargo fmt --check` clean.
- `cargo test --features serve`: **6413 passed, 2 failed**, exactly the two known macOS-environmental failures (`git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents`, `tui::dialogs::hooks_install::tests::test_content_shows_settings_path`). Nothing else.
- `cargo clippy --features serve --all-targets -- -D warnings`: 5 errors, **all in files this PR does not touch** (`src/acp/background_agent.rs:627,745`, `src/acp/supervisor.rs:4030`, `src/process/macos.rs:266`, `src/session/smart_rename.rs:1546`), local clippy 1.96.0 outrunning the pinned CI toolchain, not caused by this diff.
- `web`: `format:check` clean, `lint` clean, `npx vitest run` **3726/3726 pass**, no type errors.
- `acp-worker/aoe-agent`: `npm test` 6/6 pass.

## Automated review check-in

- **CodeRabbit:** "No actionable comments were generated." Two soft warnings: title 89 chars against a ~70 guideline (**disagree** that it matters, `PR Title Check` is green), and 71% docstring coverage (**disagree**, the diff is dense with genuine *why* comments; adding docstrings to satisfy a percentage is exactly what AGENTS.md tells reviewers to push back on).
- **Codecov:** patch 100% covered, project -0.04%. Fine.

## What blocks the merge box right now

**Infrastructure, not code.** The `Clippy` check on `0283883b` failed at *Set up job*: the runner could not download `dtolnay/rust-toolchain@29eef33` (`Name or service not known (internal-api.service.iad.github.net:443)`) after 3 retries. A re-run should clear it.

## Merge risk

**Moderate, and entirely concentrated in slice A.** Slice C reverts cleanly and its riskiest piece, `clear_requires_driven_reset: true`, rides an existing codex-proven code path; the worst case there is the documented transcript-reseed hole, which is strictly better than the status quo where `/clear` never cleared anything. Slice A is also trivially revertible, but nothing in the PR would surface its cost: no test asserts what a *failed* think call renders, and a green suite is exactly what you get today. It would arrive as a user saying "aoe-agent just stops doing what I asked and the transcript shows nothing."

Second-order: the mode picker vanishing for aoe-agent is correct but is a visible control disappearing. Worth a line in the release notes so it does not read as a bug.

## Verdict

**Design alignment:** Aligned
**Implementation:** Request changes (slice A only)
**Project fit:** Accept with scope reduction
**Merge risk:** Moderate; a guaranteed-failing `task` call renders as a silent one-line card with the NoSuchToolError hidden

Slice C is some of the better profile work in this repo: every claim spot-checked against the adapter source, the two upstream adapters, and the automation policy held up, and the tests are shaped exactly the way AGENTS.md asks. Slice A is the problem: the PR's own investigation proved `task` can never run on `aoe-agent`, and `think` is the one card that cannot show the user *why*. Fix `ThinkToolCard` to fall through on error (which fixes the same latent hole for claude and opencode), or drop slice A and ship slice C now.

**NEEDS WORK** — slice C is ready to merge as-is; slice A trades a self-explaining error card for silence in exactly the scenario it was written to fix.

njbrake (commented):

_Reviewed at `8da30f44`. Push new commits and it will be re-reviewed._

---

_Re-reviewed at `8da30f44`, against the earlier pass at `0283883b`. Push new commits and it will be re-reviewed._

---

## Summary

Four commits answering the previous review. Three of them land cleanly. The fourth, `6ba4bc3c`, implements the recommended fix correctly and it genuinely works for claude and opencode, but it does not deliver its stated payoff for `aoe-agent` itself, because the error text never reaches AoE in the first place. That is a new finding, one layer below where the last review looked.

## Disposition of every finding from the previous review

| Finding | Status | Commit |
| --- | --- | --- |
| **Blocking:** failed `task` renders a silent think card with the error hidden | **Addressed, mechanism correct, motivating case still not fixed** | `6ba4bc3c` |
| `classifyKind` should normalise casing rather than enumerate `task` / `Task` | **Addressed correctly** | `63d38e6e` |
| `acp-worker/aoe-agent`'s `npm test` is not in CI | **Addressed correctly** | `db062095` |
| Rust test comment "sits with the others" reads ambiguously | **Addressed correctly** | `8da30f44` |

Everything from the previous pass that was already approved (slice C, design alignment, docs, coverage matrix, test shape) re-verified unchanged after rebase.

## `db062095`: the CI job is real, verified two ways

This was flagged as the highest-scrutiny item, so it got the most direct check. It is not a no-op.

**In this PR's own CI run** (job `98451679249`, step 6 `aoe-agent adapter tests`, conclusion `success`):

```
08:12:36.19  ##[group]Run npm ci && npm test
08:12:37.50  added 16 packages, and audited 17 packages in 1s
08:12:37.82  # tests 6
08:12:37.82  # pass 6
08:12:37.82  # fail 0
```

**Locally, same command in that package:** `npm ci` resolves against the committed `package-lock.json` (17 entries, in sync with `package.json`), `npm test` runs `node --experimental-strip-types --test test/*.ts` and executes 6 real tests across both files.

**Failure propagates.** Deleting the `task` arm from `toolKind.ts` and re-running gave `not ok 1`, `# fail 1`, and **raw exit code 1**, so the step and the `Vitest` check both go red. It cannot pass silently.

On the rest of the checklist: it is a *step* inside the existing `vitest` job, not a new job, so there is no `needs:` to wire, no new required-check context to add to the branch ruleset, and no extra runner. Node is already set up by that job. Triggers are inherited and correct (`push: [main]` plus all `pull_request`). **Total added wall-clock: 1.6 seconds**, on a job that already takes 3m54s. Placement is right too: it sits before the codecov steps, which carry `if: ${{ !cancelled() }}`, so an adapter-test failure does not swallow the coverage upload. Deliberately excluded from the LCOV merge, which the comment states.

One cosmetic nit, non-blocking: an adapter-test failure will surface as a red check named "Vitest". Worth a sentence in the step name, or leaving as is.

## `63d38e6e`: the widening is real but correctly bounded

The `refactor:` prefix does undersell it, but the check the previous review asked for comes back clean:

- **The match is anchored, not a substring test.** It is `switch (toolName.toLowerCase())` over whole-string cases. `TaskList` becomes `"tasklist"`, hits `default`, returns `"other"`. `MultiTask` becomes `"multitask"`, likewise. No collision.
- **No other agent can be affected.** `classifyKind` lives in `acp-worker/aoe-agent/src/toolKind.ts` and has exactly two importers: that package's own `index.ts:25` and its test. It is not on any shared path.
- What it newly matches is casing variants of the adapter's own three tools plus `task`: `read`, `WRITE`, `TASK`, and so on. Since `buildTools` registers `Read` / `Write` / `Bash` case-sensitively, any lowercase variant a model emits still hits a `NoSuchToolError`; the widening only ever changes which card an already-failing call lands on. Harmless, and the comment says as much honestly.

## `6ba4bc3c`: correct fix, but the aoe-agent error text is dropped a layer earlier

The `ThinkToolCard` change is exactly option 1 from the last review and it is implemented correctly. Mutation-tested: removing the fallthrough turns the new spec red with `expected 'task' to contain 'failed'`.

**It genuinely closes the latent cross-agent hole.** claude-agent-acp does ship tool errors as `ContentBlock::Text` entries (`web/src/lib/toolErrorParse.ts` exists precisely to peel its `<tool_use_error>` wrapper out of `result.text`), so a failed claude `Task` now renders its message where it was silent before. Same for opencode. That part is delivered.

**For `aoe-agent` it is not**, and the reason is one layer below where the last review stopped. Traced through the real code path and reproduced against it:

1. The adapter's `tool-error` arm (`acp-worker/aoe-agent/src/index.ts:129-138`) sets **only** `rawOutput: { error: String(part.error) }`. It never sets `content`.
2. AoE's server builds the row's text exclusively from `update.fields.content`: `content_text` at `src/acp/acp_client.rs:5242-5247` via `extract_tool_content_text` (`:5936`), which walks `ToolCallContent::Content` text blocks. **`rawOutput` / `raw_output` appears nowhere in `src/`.** The field exists on the schema (`agent-client-protocol-schema-1.5.0/src/v1/tool_call.rs:65`); AoE simply does not read it.
3. So `content` is empty, and `transcript.rs:575-582` falls back to the literal `"tool failed"`.
4. `GenericToolCard` renders the error solely from `result.text` (`ToolErrorBody status={status} errorText={result?.text}`); the structured `output` blocks are not consulted on the error path.

Reproduced by feeding the adapter's exact wire payload through `map_update_to_events` with the real `aoe-agent` profile:

```
TODAY  -> is_error=true content=""
FIXED  -> is_error=true content="AI_NoSuchToolError: Model tried to call unavailable tool 'task'. Available tools: Read, Write, Bash."
```

where FIXED is the identical payload with one `content` text block added. And rendered through the real `ToolCard` dispatcher with the production row shape, `container.textContent` is:

```
"thinktask5.0sfailedtool failedtool failedShow attempted actioninput{"description":"delegate"}"
```

So the user does get a red failed card, a duration, and the attempted input under "Show attempted action". That is a real improvement on `0283883b`'s bare `"task"`. But the message naming the three tools that *do* exist, the one thing the whole slice is justified by, is not there.

**Three comments in the diff assert the opposite**, and all three are wrong end to end:

- `toolKind.ts`: "the think card falls through to the error body on a failure, so the NoSuchToolError text stays visible."
- `ToolCards.tsx`: "the AI SDK answers every call with a NoSuchToolError naming the tools that do exist, which is the one message a user needs to see."
- `ToolCards.test.tsx`: "every call fails with a NoSuchToolError naming the tools that do exist, and that text is the whole point."

**And the new test arranges a shape production never produces for this adapter.** `errorRow(text)` puts the error string directly into `row.text`; the pipeline puts `"tool failed"` there. The spec is a valid test of `ThinkToolCard` in general (and it does fail when the fix is reverted), it just is not evidence for the aoe-agent claim it is annotated with.

**The fix is one line, in the package this PR is already editing.** In the `tool-error` arm of `acp-worker/aoe-agent/src/index.ts`, alongside the existing `rawOutput`:

```ts
status: "failed",
content: [{ type: "content", content: { type: "text", text: String(part.error) } }],
rawOutput: { error: String(part.error) },
```

Verified above that this is the shape AoE parses. Worth checking it against the TS SDK's `ToolCallUpdate` type before committing.

**Related, pre-existing, not this PR's fault, and probably its own issue:** the `tool-result` arm (`:117-126`) has the same shape, `rawOutput` only, no `content`. So *no* `aoe-agent` tool output reaches the transcript either. A successful `Bash` call renders as `"completed"` with no stdout. Same root cause, same one-line remedy, and it makes the case that this belongs in the adapter rather than being papered over in the card.

## Rust and TS profile tables agree

Diffed directly, since a fix on one side only would reintroduce the bug on the other surface.

| Fact | `src/acp/agent_profiles.rs` | `web/src/lib/agentProfiles.ts` |
| --- | --- | --- |
| parent meta namespaces | `parent_meta_namespaces: &[]` | `parentMetaNamespaces: []` |
| wakeup tools | `supports_wakeup_tools: false` | `capabilities.wakeup: false` |
| heartbeat keepalives | `emits_heartbeat_keepalives: false` | `capabilities.heartbeatKeepalives: false` |
| mode picker | `yolo_mode_id: None`, `supports_exit_plan_mode: false` | `capabilities.legacyModeFallback: false` |
| claude specials | n/a (server has no card gates) | `todos` / `skills` / `subagents` all `false`, `specialTitles` all empty |

Consistent on every overlapping field. Nothing enforces this mechanically: `src/agents.rs:2034 test_lifecycle_facts_stay_synced_with_ts_mirror` reads the TS file but only cross-checks gemini *lifecycle strings*, not capability data. Pre-existing gap, correctly not this PR's job, but it is why the manual diff was worth doing.

`mcpPrefixes: ["mcp__"]` on the TS side is inert (the adapter has no MCP surface at all; `grep -i mcp` over `acp-worker/aoe-agent/src/` is empty), but every one of the ten profiles carries it, so keeping it is the house shape rather than a claim.

## Effect on already-recorded sessions

Asked explicitly, so stated plainly:

- **`aoe-agent`'s old `task` cards do not change.** `classifyKind` runs adapter-side at emission, and `renderToolCall` dispatches on the persisted wire `tool.kind` plus the profile's alias table, which for `aoe-agent` is `{}`. Calls already in the event log keep `kind: "other"` and keep rendering the generic card. No replay surprise there.
- **The `ThinkToolCard` change does alter replay, for every agent.** Any previously-recorded think-kind call that failed now renders as an auto-expanded generic error card instead of a one-line sparkle. This is the desirable direction, but it is a visible change to old transcripts and should be a stated consequence rather than a surprise.
- **`clear_requires_driven_reset: true` takes effect immediately for live `aoe-agent` sessions**, not just new ones. Also desirable (today `/clear` clears nothing), also worth a release-note line.
- The mode picker disappearing for `aoe-agent` is the same category, and was already flagged last time.

## Test quality, mutation-tested

Every central assertion was reverted and confirmed red. Actual output:

| Mutation | Result |
| --- | --- |
| Remove `case "task": return "think"` from `toolKind.ts` | `not ok 1 - maps the tool palette and the unregistered task tool`, `# fail 1`, exit 1 |
| Revert `toolName.toLowerCase()` to `toolName` | `not ok 1 - maps the tool palette and the unregistered task tool`, `# fail 1` |
| Remove the `ThinkToolCard` err fallthrough | `falls through to the error body for a failed think card` fails: `AssertionError: expected 'task' to contain 'failed'` |
| Restore `AOE_AGENT = { ...CLAUDE, key: "aoe-agent" }` | `aoe-agent claims no claude specials` fails: `expected { todos: true, skills: true, ...(4) } to deeply equal { todos: false, skills: false, ...(4) }` |

None of them is a test that cannot fail. Shape is house style throughout: `toolKind.test.ts` is still one test with a 9-row table (the four new rows are four lines, not four test fns), the `ToolCards` addition is one `it`, and the Rust side continues to modify existing tests rather than add new `#[test]` fns. No Playwright test was added, correctly, since nothing here needs a real browser.

The one caveat is the annotation problem above: the `ToolCards` spec is a sound test of the component and a poor citation for the aoe-agent claim.

- **Coverage matrix: no entry needed, nothing missing.** `web/src/components/acp/ToolCards.tsx` is already mapped to `src/components/acp/__tests__/ToolCards.test.tsx` in existing entries (`coverage-matrix.json` around lines 1872-1899 and 2369). `node tests/validate-coverage-matrix.mjs` reports `Coverage matrix validation passed`. Not a blocking item.
- No emdashes or `--` separators in any added prose (grepped the three-dot diff).
- No dead code, no `#[allow(dead_code)]`.
- `scripts/check-nix-embedded-assets.py`: `OK: 42 embedded asset(s)`. Nothing here is compile-time embedded, so `flake.nix` needs no change.
- All four commit subjects use conventional prefixes. `refactor(acp):` on `63d38e6e` is arguably `fix:`, given it is a behavior widening, but the scope is right and it is a nit.

## PR body does not match what is pushed

Flagging these together because there are four, and one of them matters.

1. **The description does not mention the `ToolCards.tsx` change at all.** The "Description" section covers slices A and C only. `6ba4bc3c` changes how a failed think card renders **for every agent, including claude**, and touches replay of existing transcripts. The only place it appears is the auto-generated CodeRabbit block at the very bottom. A cross-agent UI behavior change needs to be in the hand-written body, and it needs a "How to test" step.
2. **"Note this package's `npm test` is not wired into CI today"** is now false; `db062095` wires it in. Stale.
3. **The Playwright-skip justification** says "the change is in a pure data table (`web/src/lib/agentProfiles.ts`)". The PR now also changes `ToolCards.tsx`, which is not a pure data table. The conclusion still holds (the Vitest spec is the right tier), but the stated reason no longer covers the diff.
4. **"table over the tool palette plus `task` / `Task`"**: the table is now 9 rows including `TASK` and `bash`.

Also, the auto-generated summary's **"Failed think calls now show their error messages"** is an overstatement for `aoe-agent` specifically, per the finding above.

Minor doc note, pre-existing and not made worse here: `docs/structured-view.md:66` shows "none" for the "Other ACP" `/clear` boundary divider, but `aoe-agent` has had `clear_aliases: &["/clear"]` all along and now drives a real reset. The row was already stale for this agent; the PR's "nothing there becomes wrong" claim is fine as far as it goes.

## Automated review check-in

- **CodeRabbit** at `8da30f44`: "No actionable comments were generated." Pre-merge checks 4 passed, 1 warning: title is 89 chars against a ~70 guideline. **Disagree** that it matters; the `Lint PR title` check is green and the title is accurate. Docstring coverage now passes at 87.5%.
- **Codecov:** patch covered, project delta negligible.

## Local results, rebased onto `origin/main` (`ecffb860`), rebase clean with no conflicts

All cargo commands run as `env -u CLAUDE_CONFIG_DIR cargo ...`.

- `cargo fmt --check`: clean.
- `cargo clippy --features serve -- -D warnings`: **exit 0, clean.**
- `cargo test --features serve`: **6721 passed, 0 failed**, 24 suites. Both `git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` and `tui::dialogs::hooks_install::tests::test_content_shows_settings_path` pass.
- `cd web && npm run format:check`: clean. `npm run lint`: clean. `npx vitest run`: **3715/3715 pass, no type errors.**
- `acp-worker/aoe-agent && npm ci && npm test`: **6/6 pass.**
- `node web/tests/validate-coverage-matrix.mjs`: passed.
- `scripts/check-nix-embedded-assets.py`: OK, 42 assets.

All 30 GitHub checks are green. The infrastructure flake that blocked the box last time is gone.

## Merge risk

**Low, and lower than last time.** Slice C is unchanged and was already judged ready. The three infrastructure and hygiene commits are strictly additive. The `ThinkToolCard` fallthrough is a small, well-tested, trivially revertible change that strictly increases what a failed card shows for every agent.

The residual risk is a documentation risk rather than a behavior risk: three code comments and a test annotation state that `aoe-agent` users will see the `NoSuchToolError`, and they will not. Nothing in the suite would catch that, because the assertion is written against a hand-built row rather than the pipeline's. The way it bites is someone reading `toolKind.ts` in six months, believing the comment, and not fixing the adapter. Cost is low; the fix is one line in the same package.

Second-order, unchanged from last time and worth a release-note line each: the `aoe-agent` mode picker disappears, and `/clear` starts actually clearing.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve with a follow-up (the one-line adapter `content` block), or Request changes if you want it in this PR
**Project fit:** Accept
**Documentation:** Trim not needed; three comments need correcting, and the PR body needs the `ToolCards.tsx` change added
**Merge risk:** Low; the stated user-visible payoff for `aoe-agent` does not arrive, though the card is still strictly better than before

@Seluj78 turned around four commits and every one of them does what it says. The CI job is the standout: it was the item most likely to be a silent no-op and it is verifiably not, at 1.6 seconds of added wall-clock and with failures that propagate. The case-insensitive match is anchored and cannot touch another agent. The Rust and TS tables agree field for field.

The one thing to fix is that `6ba4bc3c` inherited the previous review's model of where the error text lives, and that model was wrong: `aoe-agent` ships the message in `rawOutput`, which AoE never reads. The card fix is still correct and still worth keeping, since it delivers for claude and opencode today. It just needs the adapter to put the error where AoE looks, plus three comments corrected to stop asserting a fact that is not true yet.

**READY, with one follow-up.** I would take this as is and land the adapter `content` block either here or immediately after, rather than hold slice C any longer.

jerome-benoit (approved):
Approved at `b894c3121c1d21200e92c894a33ddc83f915cfe2`.

I independently revalidated the force-pushed head. The Rust and TypeScript capability profiles agree, tool-kind classification is exact and case-insensitive, failed think cards correctly use the generic error rendering path, and the added aoe-agent test step executes successfully. All current GitHub checks pass.

The previously reported `rawOutput` versus `content` limitation remains valid: aoe-agent still exposes only the generic `tool failed` text through AoE. This behavior is already present in the base revision and #1904 remains open, so I do not consider it a regression or blocker for this scoped partial fix. I am not duplicating the existing review finding inline.

Issue #2948, opened 2026-07-17
TUI: render the row-column plugin slot in the daemon-connected session picker (remote-home)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Part of #268. Follow-up to #2366 / #2402. Sibling of the `row-badge` (companion issue) and `sort-key` TUI-parity issues.

## Problem

A plugin that fills the `row-column` slot (the GitHub plugin's `github_pr_status` cell, today) renders a per-session status word on the web dashboard sidebar but shows nothing on the native TUI. The `row-column` is the one-word scannable summary of a session's most-urgent PR signal ("changes requested", "CI failing", "unresolved comments", "awaiting review", "healthy"), so without it the native session list gives no PR context per row even when attached to a daemon.

## What's there today

`#2366` shipped the host backend + web rendering; `#2402` rendered only `status-bar` + `detail-badge` + `notification` in the native structured view and left the session-row slots out. The exclusion is documented (`docs/development/internals/plugin-system.md:790-798`) and enforced in code (`src/tui/plugin_ui.rs:1-7`; `src/tui/structured_view/render.rs:500-512` iterates only `StatusBar` / `DetailBadge`).

`row-column` is a per-session slot (`aoe-plugin-api/src/manifest.rs:337-346`). Its natural TUI home is the daemon-connected remote-home picker (`src/tui/remote_home/`), whose rows render title / status / project-path today (`src/tui/remote_home/render.rs:82-119`) and which lists sessions over HTTP (`src/tui/remote_home/mod.rs:206`) but pulls no plugin UI state. The web equivalent is `PluginRowColumn` (`web/src/components/plugin/PluginSlots.tsx:177-203`), a right-anchored tone-colored text cell, mounted per row at `web/src/components/WorkspaceSidebar.tsx:1538`.

The host already stores and serves the payload: `RowColumnPayload { text, tone, tooltip, sort_value?, filter_values? }` at `src/plugin/ui_state.rs:155-170`. The `sort_value` scalar on this payload is what the `sort-key` sibling issue consumes.

## Proposal

Render `row-column` text on each remote-home picker row, reusing the snapshot the TUI can already fetch.

[...]

Issue #2528, opened 2026-06-28
Plugin commands: TUI keybind/palette executor + fire-and-forget worker-command path (follow-up to #2526)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Follow-up to #2526 (client-executed plugin command actions + `open-ui-link` in the web cmd+k palette and keymap). That PR shipped the web surface and the generic `ClientAction::OpenUiLink` machinery. Two pieces were deliberately deferred and are tracked here.

## 1. TUI keybind + palette executor for plugin commands

The native TUI does not yet execute plugin commands. Two gaps:

- Home view (`src/tui/home/input.rs:2369`): a resolved plugin keybind still pops a "plugin runtime not available yet" `InfoDialog`. The home view has no plugin UI-state snapshot, so it cannot resolve an `open-ui-link` href.
- Structured view (`src/tui/structured_view/mod.rs`): holds the per-session plugin UI snapshot (`state.plugin_ui`) and already has `webbrowser::open` (the `Intent::OpenInBrowser` arm), but its `Intent` key pipeline does not consult plugin keybinds at all.

Target: in the structured view, consult plugin keybinds (`bindings::resolve_action` / `plugin_bindings`) and, for a command whose manifest `action` is `OpenUiLink { slot, id }`, read the active session's `(plugin_id, slot, id)` href from the snapshot, validate the scheme (http/https only), and `webbrowser::open` it. Decide whether the home view also gains an executor (it would need the snapshot plumbed in) or whether plugin command execution is structured-view-only for now.

This needs a live-daemon e2e (`tests/e2e/`, the `acp_focus_isolation_e2e` harness pattern) since it depends on a running plugin host pushing UI state, which is why it was split out of #2526.

## 2. Fire-and-forget worker-command path (action-less commands)

#2526 surfaces only commands that carry a client `action`. A command without an `action` (e.g. the GitHub plugin's `status` / `refresh`) cannot be invoked from the palette or a keybind yet. The design from the #2526 debate:

- A standard host->worker notification `plugin.command.invoke` with `{ command, session_id, source, invocation }` (the host validates the command exists, has no client action, and the session exists), dispatched through the existing fire-and-forget `notify_worker`.
- `POST /api/plugins/commands/{fqid}/invoke` for the web to dispatch these (no result expected; success means dispatched).
[...]

PR #3532, opened 2026-08-25
refactor(plugin): remove the dead plugin-API surface from the abandoned MCP/Skills-plugin approach

Seluj78:
**Note:** Claude handled the implementation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Description

Four PRs (#2992, #2993, #2989, #2991) added plugin-API surface so MCP and Skills management could ship as external plugins. That plan is reversed: both move back into core (#3043, #3044), so the plugin-facing surface has no consumer left. No shipped plugin (`aoe-web`, `plugin-github`, `plugin-cron`, `plugin-template`) declares any of it; the only callers were unit tests and one Playwright fixture.

Removed:

- **Host RPCs** in `src/plugin/host_api.rs`: `config.read`, `config.write`, the eight `mcp.*` methods, `fs.read` / `fs.write`, and the seven `skills.*` methods, plus their handlers and the 1 MiB fs size cap.
- **`UiSlot::SettingsPage`** and its full stack: the manifest variant and validation, `SettingsPagePayload`, `PluginSettingsPage` / `PluginSettingsPageBody`, and the parametric `/settings/plugin-page:<...>` nav route in `SettingsView.tsx`.
- **`UiSlot::ToolCardBadge`** and its full stack: the manifest variant and validation, `ToolCardTarget` / `ToolCardBadge` / `ToolCardBadgePayload`, `PluginToolCardBadges`, the two `meta=` overlays on the transcript MCP and skill cards, and `AcpSessionContext` / `useAcpSessionId`, which existed only to reach those cards.
- **Their tests and fixtures**, including `web/tests/fixtures/tool-card-badge-plugin/` and the two dead `coverage-matrix.json` entries.

Kept deliberately, because the still-open core epics name them as the backend they wire into:

- `src/session/mcp_overrides.rs` write primitives (`remove_global_server`, `insert_global_server_if_absent`, `replace_global_server_if_present`), which #3045 calls out as "added for the abandoned plugin RPCs, now repurposed".
- `src/session/mcp_state.rs` conflict-token hardening, already live via `src/server/api/mcp.rs`.
- `src/session/skills_model.rs`. The issue expected this to be orphaned; it is not. `src/cli/skill.rs`, `src/server/api/skills.rs`, `src/tui/dialogs/skills_manager.rs`, `src/session/instance.rs`, and `src/session/container_config.rs` all use it.
[...]

PR #3534, opened 2026-08-25
feat(tui): render the row-badge plugin slot in the remote-home picker

### LINKED ISSUE #2947: TUI: render the row-badge plugin slot in the daemon-connected session picker (remote-home)

Part of #268. Follow-up to #2366 / #2402. Sibling of the `row-column` (companion issue) and `sort-key` TUI-parity issues.

## Problem

A plugin that fills the `row-badge` slot (the GitHub plugin's PR badge, `github_pr_badge`, today) renders its per-session badge chips on the web dashboard sidebar but shows nothing on the native TUI. Attach the GitHub plugin to a daemon, open the native session list, and the PR state that the web sidebar shows as tone-colored chips is simply absent. The badge is the plugin's primary at-a-glance session-row indicator, so its absence is the most visible half of the "GitHub plugin isn't shown in the TUI" gap.

## What's there today

`#2366` shipped the full host backend for all slots (`ui.state.*` RPCs, the `UiStore`, the `GET /api/plugins/ui-state` snapshot) and web rendering. `#2402` then rendered a narrow subset in the native structured view (`status-bar`, `detail-badge`, `notification`) and explicitly left the session-row slots out.

The exclusion is documented and enforced in code:

- `docs/development/internals/plugin-system.md:790-798`: the structured-view TUI renders text/tone for `status-bar` + `detail-badge` + `notification` only, and "`card`, `pane`, `row-badge`, `row-column`, `sort-key`, and `filter-facet` have no structured-view surface."
- `src/tui/plugin_ui.rs:1-7` (the only TUI consumer of the snapshot) states the same ignore list in its module doc.
- `src/tui/structured_view/render.rs:500-512` only iterates `UiSlot::StatusBar` and `UiSlot::DetailBadge`.

`row-badge` is a per-session slot (`aoe-plugin-api/src/manifest.rs:337-346`), so its natural TUI home is a per-session-row surface. The daemon-connected session list is the remote-home picker (`src/tui/remote_home/`), whose rows render title / status / project-path today (`src/tui/remote_home/render.rs:82-119`) and which already lists sessions over HTTP (`src/tui/remote_home/mod.rs:206`, `client.list_sessions`). It does not fetch the plugin UI snapshot at all. The web equivalent is `PluginRowBadges` (`web/src/components/plugin/PluginSlots.tsx:145-170`), mounted per row via `PluginRowLine` at `web/src/components/WorkspaceSidebar.tsx:1538`.

The payload the host already stores and serves is `RowBadgePayload` (single `{ text, tone, icon, href, tooltip }` or `items: BadgeItem[]`) at `src/plugin/ui_state.rs:138-153`.

## Proposal

Render `row-badge` entries on each remote-home picker row, reusing the snapshot the TUI can already fetch.

1. Fetch the plugin UI snapshot in `remote_home`, mirroring the structured view's poller (`src/tui/structured_view/mod.rs:254-262`, `:386`) and reusing `HttpClient::plugin_ui_state` (`src/acp/client/http.rs:193`). Store it on `RemoteHomeState`.
2. Extend the TUI selectors in `src/tui/plugin_ui.rs` beyond `StatusBar`/`DetailBadge` to read per-session `RowBadge` entries by `session_id`, mirroring `web/src/lib/pluginUi.ts`.
3. In `render_list` (`src/tui/remote_home/render.rs:82-119`) append a tone-colored badge span per row for the matching session. Terminal fidelity: render `text` (falling back to a compact marker when a badge item is icon-only), tone-colored via `plugin_ui::tone_style`; `icon`, `tooltip`, and `href` are dropped, matching how the existing TUI slots already degrade. Opening the PR from the badge (the `href`) is the `open-ui-link` executor tracked in #2528.
4. Unit tests for the badge-entry-to-`Line` selector (present / empty / malformed).

## Why now

- The badge is the GitHub plugin's single most visible surface; its absence is the core of the reported "nothing shows in the TUI" gap.
- The data is already in the snapshot the TUI fetches elsewhere; this is rendering plus one fetch, no new host or endpoint work.
- Pairs directly with the `row-column` companion issue (same render path) and the `sort-key` issue (same picker).

- As an AoE user attached to a daemon in the native TUI, when a session's branch has an open PR the GitHub plugin reports, then that session's row in the remote-home picker shows a tone-colored PR badge indicator matching what the web sidebar shows.
  - Suggested test layer: `tests/e2e/` (the `acp_focus_isolation_e2e` live-daemon pattern, a fake worker pushing a `row-badge` entry).
- As a user with no GitHub token (or a session with no open PR), when the plugin pushes no `row-badge` for that session, then the row renders unchanged with no empty artifact or layout shift.
  - Suggested test layer: Rust unit on the badge selector in `src/tui/plugin_ui.rs` with an empty / non-matching snapshot.

## Open questions

- Column budget: the badge competes with title / status / path for a narrow terminal width. Truncate the path, cap the badge to N chars, or right-anchor it?
- Should `items: []` (the documented "clear the row" form) be a no-op here, matching the web?

## Out of scope

- The standalone (non-daemon) `home` list, which has no daemon link (the #2402 deferral).
- `href` / open-in-browser from the badge (#2528).
- `pane` (#2467), `sort-key` and `row-column` (their own sibling issues).

## Cross-references

- #2366 — landed the host backend + web rendering for the slots.
- #2402 — rendered the text/badge/toast subset in the structured-view TUI; deferred the session-row slots.
- #2467 — renders the `pane` slot in the TUI (sibling gap).
- #2528 — TUI keybind/palette executor for plugin commands incl. `open-ui-link` (the badge `href`).
- #2948 — `row-column` TUI-parity sibling (shares this render path; land together).
- #2949 — `sort-key` TUI-parity sibling.
- #268 — plugin system epic.

Seluj78:

## Description

A plugin filling the `row-badge` slot (the GitHub plugin's PR badge) rendered tone-colored chips on the web sidebar and nothing at all on the native session picker. That was the most visible half of the "the GitHub plugin isn't shown in the TUI" gap. This renders those badges on each row of the daemon-connected remote-home picker.

No host or fetch work was needed: #2948 already pulls the plugin UI snapshot alongside the session list, so this is a selector plus rendering. `plugin_ui::row_badge_cells` reads the slot's two payload forms (a single badge, or an `items` list), and the picker paints the result as a tone-colored group left of the existing `row-column` group, matching the web's left-to-right order.

The one design call worth flagging: **badges share a single 24-column plugin region with `row-column` rather than claiming a second column of their own.** The row already spends 41 columns on the highlight symbol, title, and status before any plugin content, and `row-column` may take 24 more. A second independent 24-column group would have put the project path at column 85, off the right edge of an 80-column terminal, rather than merely squeezing it. Sharing the region means badges take their cells from the column, never from the path, and a picker with no badges anywhere still gives `row-column` the full 24 and renders byte-identical to today. There is a regression test for each half of that.

Two smaller calls, both deliberate deviations from the issue body, both argued out with a `/debate` consult:

- **Icon-only badges render nothing.** The issue suggested a compact marker. A lucide icon name is what carries the meaning (success, draft, warning); a bullet preserves only "something exists" while discarding which state it was, so three icon-only badges would paint `•••` that can be read as the wrong status. The web renders nothing when it cannot resolve an icon, and the TUI already drops unsupported icons elsewhere. A plugin that wants terminal visibility should send text.
- **No terminal-width-responsive budgeting.** Deriving the region from `area.width` would also require rewriting the title and status columns to be terminal-cell-aware (they count chars today), which is a layout change unrelated to this slot. The shared region already guarantees the path never sits further right than it does on `main`, so responsive budgeting is a separate hardening concern.

`items: []` clears the row as documented, and a malformed `items` clears it too rather than silently falling back to the entry's own `text`. `tooltip` and `href` have no terminal surface; opening a badge's `href` is #2528.

Closes #2947

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] With a daemon running and the GitHub plugin attached, open the native picker (`AOE_DAEMON_URL=... aoe`) and confirm a session whose branch has an open PR shows a tone-colored badge between the status and the project path, matching what the web sidebar shows for that session.
- [ ] Confirm a session with no open PR renders unchanged: no empty artifact, and its project path starts at the same screen column as the badged row's.
- [ ] Detach the plugin (or run against a daemon with no plugins) and confirm the picker rows look exactly as they did before this PR, with no reserved blank column.
- [ ] Narrow the terminal toward 80 columns with both a badge and a `row-column` entry present and confirm the project path is still visible; the badge should eat into the column text, not the path.
- [ ] Stop the daemon and press `r` and confirm the badges clear while the session list itself stays intact.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is fully reachable from the cheaper tier that #2948 established for this exact render path. 6 ratatui `TestBackend` tests paint the real picker and assert the badge text, its position left of the column, path alignment across badged and unbadged rows, the region-sharing invariant, and the zero-width cases; 2 table-driven selector tests cover all 9 payload shapes. A live-daemon e2e would add roughly 10 to 28s to a strictly serial suite to re-assert what those already fail on. Per AGENTS.md, cheapest tier that can actually fail.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, plan, and implementation drafted by Claude under my direction, with a `/debate` consult between two models on the width budgeting and the icon-only question. I made the design calls, reviewed each commit, and ran the manual test steps.

## Summary
- Added `row-badge` rendering to the daemon-connected remote-home picker.
- Rendered badge text with tone colors and shared a 24-cell region with `row-column`.
- Preserved project-path alignment and avoided layout shifts when badges are absent.
- Added handling for empty, malformed, icon-only, and oversized badge content.
- Added selector, rendering, alignment, and width regression tests.
- Documented TUI support and badge limitations.

## Benefits
Users can now see session badges, such as GitHub pull request status, in the native TUI. The picker keeps its layout stable across sessions.

## Further enhancement
Icon rendering, tooltips, links, and browser actions remain unsupported in the TUI.

### DIFF
--- a/docs/plugin-api.md
+++ b/docs/plugin-api.md
@@ -366,6 +366,11 @@ id = "my_pane"
 | `composer-action` | per-session | A button beside the ACP composer controls (requires `api_version >= 8`). |
 | `notification` | n/a | A transient notification pushed via `ui.notify`; gated by the `notifications` capability, not a slot declaration. |
 
+The native TUI's daemon-connected session picker renders `row-badge` and
+`row-column` as tone-colored text, sharing one bounded column between them. A
+badge carrying only an `icon` renders nothing there, since a lucide name has no
+terminal glyph: send `text` for a badge that should be visible in a terminal.
+
 ### Pane payload
 
 A `pane` entry renders a dockable tool-window. The worker pushes it with
--- a/src/tui/plugin_ui.rs
+++ b/src/tui/plugin_ui.rs
@@ -963,6 +963,79 @@ mod tests {
     }
 
     #[test]
+    fn row_badge_cells_read_both_payload_forms() {
+        // One badge payload and the cells it must yield. Covers the single
+        // badge form, the `items` list, and every shape that renders nothing.
+        let cases: [(serde_json::Value, &[(&str, Option<Tone>)]); 9] = [
+            // Single badge: text and tone kept, the web-only fields dropped.
+            (
+                json!({"text": "open", "tone": "success", "icon": "git-pull-request",
+                       "href": "https://example.test/1", "tooltip": "dropped"}),
+                &[("open", Some(Tone::Success))],
+            ),
+            // Items keep snapshot order and carry their own tones.
+            (
+                json!({"items": [{"text": "draft"}, {"text": "2 checks", "tone": "warn"}]}),
+                &[("draft", None), ("2 checks", Some(Tone::Warn))],
+            ),
+            // The documented "clear the row" form.
+            (json!({"items": []}), &[]),
+            // `items` present selects the list form, so the entry's own text
+            // is not a fallback: a plugin clearing the row means it.
+            (json!({"text": "stale", "items": []}), &[]),
+            (
+                json!({"text": "outer", "items": [{"text": "inner"}]}),
+                &[("inner", None)],
+            ),
+            // Malformed `items` clears the row rather than quietly rendering
+            // the other form.
+            (json!({"text": "stale", "items": 7}), &[]),
+            // A lucide name has no terminal glyph, so an icon-only badge
+            // yields nothing; its text-carrying neighbours still render.
+            (json!({"icon": "git-pull-request"}), &[]),
+            (
+                json!({"items": [{"icon": "check"}, {"text": "kept"}, {"text": "  "}, {"text": 7}]}),
+                &[("kept", None)],
+            ),
+            // An unreadable tone degrades to neutral rather than dropping text.
+            (
+                json!({"text": "kept", "tone": "chartreuse"}),
+                &[("kept", None)],
+            ),
+        ];
+        for (payload, expected) in cases {
+            let snap = pane_snapshot(json!([
+                {"plugin_id": "gh", "slot": "row-badge", "id": "pr",
+                 "session_id": "s1", "payload": payload}
+            ]));
+            let expected: Vec<(String, Option<Tone>)> = expected
+                .iter()
+                .map(|(t, tone)| ((*t).to_string(), *tone))
+                .collect();
+            assert_eq!(row_badge_cells(&snap, "s1"), expected);
+        }
+    }
+
+    #[test]
+    fn row_badge_cells_read_this_session_and_slot_only() {
+        let snap = pane_snapshot(json!([
+            {"plugin_id": "gh", "slot": "row-badge", "id": "pr", "session_id": "s1",
+             "payload": {"text": "mine"}},
+            {"plugin_id": "gh", "slot": "row-badge", "id": "pr", "session_id": "s2",
+             "payload": {"text": "theirs"}},
+            {"plugin_id": "gh", "slot": "row-badge", "id": "pr",
+             "payload": {"text": "global"}},
+            {"plugin_id": "gh", "slot": "row-column", "id": "col", "session_id": "s1",
+             "payload": {"text": "column"}}
+        ]));
+        assert_eq!(
+            row_badge_cells(&snap, "s1"),
+            vec![("mine".to_string(), None)]
+        );
+        assert!(row_badge_cells(&snap, "s3").is_empty());
+    }
+
+    #[test]
     fn row_column_cells_ignore_other_slots() {
         let snap = pane_snapshot(json!([
             {"plugin_id": "a", "slot": "row-badge", "id": "b", "session_id": "s1",
--- a/src/tui/plugin_ui.rs
+++ b/src/tui/plugin_ui.rs
@@ -4,9 +4,9 @@
 //! `StatusBar` (global) and `DetailBadge` (per-session) text, tone-colored,
 //! plus `Notification` toasts, and `Pane` (per session) and `HomePane` (global)
 //! blocks in a toggleable overlay (#2467); the remote-home picker shows
-//! `RowColumn` text per session row
-//! (#2948). Icons, tooltips, hrefs, and the
-//! `Card`/`RowBadge`/`SortKey`/`FilterFacet` slots have no TUI surface here and
+//! `RowBadge` (#2947) and `RowColumn` (#2948) text per session row.
+//! Icons, tooltips, hrefs, and the
+//! `Card`/`SortKey`/`FilterFacet` slots have no TUI surface here and
 //! are ignored.
 //!
 //! Kept side-effect-free so the render layer can borrow the snapshot and so the
@@ -72,6 +72,35 @@ pub fn row_column_cells(snapshot: &UiSnapshot, session_id: &str) -> Vec<(String,
     session_entries(snapshot, UiSlot::RowColumn, session_id)
         .filter_map(|e| entry_text(e).map(|t| (t.to_string(), entry_tone(e))))
         .collect()
+}
+
+/// This session's `RowBadge` chips as `(text, tone)`, in snapshot order
+/// (#2947). One entry is either a single badge or a list of them under `items`,
+/// so the picker row can show several chips the way the web sidebar does.
+///
+/// The presence of `items` selects the list form: the documented `items: []`
+/// clears the row rather than falling back to the entry's own `text`, and a
+/// malformed `items` clears it too rather than silently rendering the other
+/// form. A badge carrying only an `icon` yields nothing, since a lucide name
+/// has no terminal glyph and a generic stand-in would keep the badge's presence
+/// while dropping which state it meant. `tooltip` and `href` have no terminal
+/// surface either; opening the `href` is tracked as #2528.
+pub fn row_badge_cells(snapshot: &UiSnapshot, session_id: &str) -> Vec<(String, Option<Tone>)> {
+    session_entries(snapshot, UiSlot::RowBadge, session_id)
+        .flat_map(|e| match e.payload.get("items") {
+            Some(items) => items
+                .as_array()
+                .map(|items| items.iter().filter_map(badge_cell).collect())
+                .unwrap_or_default(),
+            None => badge_cell(&e.payload).into_iter().collect::<Vec<_>>(),
+        })
+        .collect()
+}
+
+/// One badge's renderable `(text, tone)`, from either a whole `row-badge`
+/// payload or one element of its `items` list: the two carry the same fields.
+fn badge_cell(badge: &Value) -> Option<(String, Option<Tone>)> {
+    Some((block_str(badge, "text")?.to_string(), block_tone(badge)))
 }
 
 /// Map a tone to a foreground style against the active theme. `None` (no tone)
@@ -697,7 +726,8 @@ fn section_lines(block: &Value, indent: usize, theme: &Theme) -> Vec<Line<'stati
     out
 }
 
-/// The block's tone, if it carries a valid one.
+/// The tone of a pane block or a `row-badge` payload, if it carries a valid
+/// one. Both spell the field the same way.
 fn block_tone(block: &Value) -> Option<Tone> {
     block
         .get("tone")
@@ -711,8 +741,9 @@ fn value_tone(block: &Value) -> Option<Tone> {
         .and_then(|v| serde_json::from_value::<Tone>(v.clone()).ok())
 }
 
-/// A trimmed, non-empty string field, or `None`. Defensive against a malformed
-/// or schema-skewed block so the renderer never panics on plugin data.
+/// A trimmed, non-empty string field of a plugin-authored object, or `None`.
+/// Defensive against a malformed or schema-skewed payload so the renderer never
+/// panics on plugin data.
 fn block_str<'a>(block: &'a Value, key: &str) -> Option<&'a str> {
     block
         .get(key)
--- a/src/tui/remote_home/render.rs
+++ b/src/tui/remote_home/render.rs
@@ -342,6 +342,11 @@ mod tests {
                "session_id": session_id, "payload": {"text": text}})
     }
 
+    fn row_badge(session_id: &str, payload: serde_json::Value) -> serde_json::Value {
+        json!({"plugin_id": "gh", "slot": "row-badge", "id": "pr",
+               "session_id": session_id, "payload": payload})
+    }
+
     /// Screen column where `needle` starts on the first row carrying it. Indexes
     /// the per-cell strings from `rows`, one character per painted cell, so it is
     /// a real column: a byte offset would be skewed by the multi-byte `▸ `
@@ -490,6 +495,85 @@ mod tests {
         let (cells, width) = row_column_cells(&state, "s1", PLUGIN_REGION_MAX_WIDTH);
         assert_eq!(cells.len(), 1);
         assert_eq!(width, PLUGIN_REGION_MAX_WIDTH);
+    }
+
+    #[test]
+    fn row_shows_the_plugin_row_badge_text_before_the_column() {
+        let state = state_with(
+            &["s1"],
+            json!([
+                row_column("s1", "CI failing"),
+                row_badge("s1", json!({"text": "draft", "tone": "warn"}))
+            ]),
+        );
+        let painted = rows(&state);
+        assert!(painted.iter().any(|l| l.contains("draft")), "{painted:?}");
+        // Badges sit left of the column, matching the web's row line.
+        assert!(column_of(&painted, "draft") < column_of(&painted, "CI failing"));
+    }
+
+    #[test]
+    fn badges_share_the_plugin_region_instead_of_widening_it() {
+        // A badge plus a column wordy enough to have filled the region alone.
+        let long = "x".repeat(PLUGIN_REGION_MAX_WIDTH + 10);
+        let state = state_with(
+            &["s1"],
+            json!([
+                row_badge("s1", json!({"text": "draft"})),
+                row_column("s1", &long)
+            ]),
+        );
+        let painted = rows(&state);
+        assert!(painted.iter().any(|l| l.contains("draft")), "{painted:?}");
+        // The 41-cell prefix, the whole region, and the gap before the path:
+        // the badge took its cells from the column, not from the path, which
+        // starts exactly where a badge-free capped column leaves it.
+        assert_eq!(
+            column_of(&painted, "/tmp/s1"),
+            41 + PLUGIN_REGION_MAX_WIDTH + 2
+        );
+    }
+
+    #[test]
+    fn badge_free_rows_leave_the_whole_region_to_the_column() {
+        let long = "x".repeat(PLUGIN_REGION_MAX_WIDTH + 10);
+        let state = state_with(&["s1"], json!([row_column("s1", &long)]));
+        let (_, width) = row_column_cells(&state, "s1", PLUGIN_REGION_MAX_WIDTH);
+        assert_eq!(width, PLUGIN_REGION_MAX_WIDTH);
+    }
+
+    #[test]
+    fn badge_column_pads_so_the_path_stays_aligned() {
+        // s2 has no badge; both rows must start the path at the same column.
+        let state = state_with(
+            &["s1", "s2"],
+            json!([row_badge("s1", json!({"text": "changes requested"}))]),
+        );
+        let painted = rows(&state);
+        assert_eq!(
+            column_of(&painted, "/tmp/s1"),
+            column_of(&painted, "/tmp/s2")
+        );
+    }
+
+    #[test]
+    fn cleared_and_icon_only_badges_reserve_no_width() {
+        // Both render nothing, so the path sits where it does with no entries.
+        for payload in [json!({"items": []}), json!({"icon": "git-pull-request"})] {
+            let state = state_with(&["s1"], json!([row_badge("s1", payload.clone())]));
+            assert_eq!(column_of(&rows(&state), "/tmp/s1"), 41, "{payload}");
+        }
+    }
+
+    #[test]
+    fn long_badge_text_is_capped_to_the_badge_budget() {
+        let long = "b".repeat(ROW_BADGE_MAX_WIDTH + 20);
+        let state = state_with(&["s1"], json!([row_badge("s1", json!({"text": long}))]));
+        let (cells, width) = row_badge_cells(&state, "s1");
+        assert_eq!(width, ROW_BADGE_MAX_WIDTH);
+        assert!(cells[0].0.ends_with('\u{2026}'));
+        let painted = rows(&state);
+        assert!(painted.iter().any(|l| l.contains("/tmp/s1")), "{painted:?}");
     }
 
     #[test]
--- a/src/tui/remote_home/render.rs
+++ b/src/tui/remote_home/render.rs
@@ -16,40 +16,47 @@ use crate::tui::styles::{has_min_contrast, Theme};
 
 const SELECTED_ROW_CONTRAST_RATIO: f32 = 3.0;
 
-/// Widest a plugin's `row-column` cell may grow, in terminal cells, so a chatty
-/// plugin cannot push the project path off the row. Matches the title column's
-/// budget.
-const ROW_COLUMN_MAX_WIDTH: usize = 24;
+/// Widest the whole plugin region may grow, in terminal cells: the `row-badge`
+/// chips, the `row-column` cells, and the gap between the two groups. One shared
+/// cap rather than one per slot because the row already spends 41 cells on the
+/// highlight symbol, title, and status before any plugin content, so a second
+/// independent 24-cell group would push the project path past the right edge of
+/// an 80-column terminal instead of merely squeezing it.
+const PLUGIN_REGION_MAX_WIDTH: usize = 24;
 
-/// Gap between two plugins' cells inside the same column.
-const ROW_COLUMN_GAP: &str = "  ";
+/// Slice of the region the `row-badge` chips may claim, leaving the rest to the
+/// wordier `row-column` text. Badges are terse state markers ("draft",
+/// "approved"), so they need far less room than a status sentence.
+const ROW_BADGE_MAX_WIDTH: usize = 10;
 
-/// One plugin's `row-column` text plus the tone to paint it in.
-type RowColumnCell = (String, Option<Tone>);
+/// Gap between two plugins' cells inside a group, and between the two groups.
+const PLUGIN_COLUMN_GAP: &str = "  ";
+
+/// One plugin's cell text plus the tone to paint it in.
+type PluginCell = (String, Option<Tone>);
 
 /// A session's cells and the display width they occupy together.
-type MeasuredCells = (Vec<RowColumnCell>, usize);
+type MeasuredCells = (Vec<PluginCell>, usize);
 
-/// One session's `row-column` cells, truncated to the shared budget, plus the
-/// display width they occupy. Separate from rendering so the column width can be
-/// computed across every listed session before any row is painted: every row
-/// pads to that one width, so a session with no cell leaves a blank of the same
-/// size instead of shifting the path column (#2948). Returns width 0 when no
-/// plugin pushed anything, which keeps the row identical to before this slot
-/// rendered at all.
+/// One session's cells for a slot, truncated to `budget`, plus the display width
+/// they occupy. Separate from rendering so a group's width can be computed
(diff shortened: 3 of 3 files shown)


Labeeb2339 (commented):
I reviewed the current head `e11417fd39be908675bfca6c62d984b3632bfe31` across the three changed files. The session-scoped badge filtering in `src/tui/plugin_ui.rs` is covered for single/list payloads, malformed entries, and cross-session isolation; the render tests also cover badge/column ordering, shared-width budgeting, wide glyphs, and empty badges. The shared `PLUGIN_REGION_MAX_WIDTH` calculation preserves the existing path column when badges are absent, and the current checks show formatting, Clippy, static analysis, and Codecov passing while the Cargo test jobs are still running. I found no additional blocking defect in the reviewed scope; I’d recheck the final test results before merge.

njbrake (commented):

_Reviewed at `e11417fd`. Push new commits and it will be re-reviewed._

---

_Reviewed at `e11417fd`. Push new commits and it will be re-reviewed._

---

## Summary

Renders the `row-badge` plugin slot on each row of the daemon-connected remote-home picker, left of the existing `row-column` group, matching the web sidebar's left-to-right order. No host or fetch work: #2948 already pulls the plugin UI snapshot alongside the session list, so this is a new selector (`plugin_ui::row_badge_cells`) plus a generalized measure-and-pad step in `render_list`. The former `ROW_COLUMN_MAX_WIDTH` becomes a shared `PLUGIN_REGION_MAX_WIDTH` of 24 cells that both groups draw from, with badges capped at 10 of those.

## Design alignment

**Aligned.**

I compared against the sibling that shipped this exact render path, `row_column_cells` in `src/tui/plugin_ui.rs` and the measure/pad/paint pipeline in `src/tui/remote_home/render.rs` (#2948). The PR extends that mechanism rather than adding a second one beside it: `row_column_cells(state, id)` becomes `measured_cells(cells, budget)` with the budget parameterized, and both slots call through it. There is one measurement path, one padding path, one gap constant.

Three more precedents checked:

- Icon-only badges yielding nothing already exists in this file. `row_badges` (`src/tui/plugin_ui.rs:627`) does `filter_map(|b| Some((block_str(b, "text")?.to_string(), block_tone(b))))` for pane row and section badges, with the comment "an icon-only badge (the web's glyph-only signal) contributes nothing rather than an empty span". The new `badge_cell` is character-for-character that closure, so the behavior is precedented even where I question its product consequence below.
- Tone styling routes through `plugin_ui::tone_style` and `selected_row_style`, so it reads only semantic `Theme` fields (`dimmed`, `accent`, `running`, `waiting`, `error`) and keeps the 3.0 contrast guard against `theme.session_selection`. DESIGN.md's "no hardcoded chrome colors, use semantic tokens" rule is satisfied; there is no new color surface to approve.
- `aoe-plugin-api/` correctly untouched. `UiSlot::RowBadge` and the `BadgeItem` shape already exist; this PR adds no manifest or capability type, and none is needed.

No new abstraction with a single caller. `push_group` has two callers; `measured_cells` has two callers.

## Code review

| File | Assessment |
| --- | --- |
| `src/tui/plugin_ui.rs` (+110/-6) | `row_badge_cells` picks the list form on the presence of `items`, else the single-badge form. Reuses `session_entries`, `block_str`, `block_tone`. Two table-driven tests, 9 payload shapes plus a session-isolation test. Correct, with one doc-comment inaccuracy noted below. |
| `src/tui/remote_home/render.rs` (+215/-60) | Constant rename, `measured_cells` extraction, `push_group`, and the two-phase budget in `render_list`. The clamp comment on `.min(budget - gap)` is the good kind of "why" comment. One test is redundant. |
| `docs/development/internals/plugin-system.md` (+18/-7) | Accurate on the shared region, the 24/10 split, and the icon-only and `items: []` behavior, with one exception noted below. |

Definition of done, run locally in a clean worktree at `e11417fd` rebased on `origin/main` (`ecffb860`), all with `env -u CLAUDE_CONFIG_DIR`:

- `cargo fmt --check`: clean.
- `cargo test --features serve`: exit 0, zero failures across every target including the doc tests.

House rules: three commits, all with conventional prefixes (`feat(tui):`, `test(tui):`, `docs(plugins):`). No emdashes or double hyphens in any added prose. No `#[allow(dead_code)]`; every new symbol (`PLUGIN_REGION_MAX_WIDTH`, `ROW_BADGE_MAX_WIDTH`, `PLUGIN_COLUMN_GAP`, `measured_cells`, `push_group`, `badge_cell`, both `row_badge_cells`) has a live caller. No OS-specific logic. Permutations live in tables inside one test rather than one test per case, which is the house style.

**Test tier: I agree with the PR's decision to skip an e2e test.** The `TestBackend` tests paint the real picker through the real `render()` and assert screen columns, so they fail on any layout regression this PR could cause. The only thing a live-daemon e2e would add is the snapshot fetch, which #2948 established and this PR does not touch, at 10 to 28 seconds on a strictly serial suite. That is the AGENTS.md cost calculus applied correctly.

## Automated review check-in

**CodeRabbit, "Out of Scope Changes" warning** ("the documentation also removes unrelated host APIs and UI-slot documentation for global configuration, MCP, filesystem, skill management, settings-page, and tool-card-badge"): **disagree, false positive.** CodeRabbit reviewed the range `81e9d3f3..e11417fd`, which spans the rebase. #3532 ("remove the dead plugin-API surface from the abandoned MCP/Skills-plugin approach") landed on main inside that range and is what removed those sections. The authored diff, `git diff origin/main...e11417fd`, touches only the row-badge paragraphs in that file, 18 added and 7 removed. Nothing to fix.

**CodeRabbit merge-risk note** ("the documented plugin contract should still clarify that malformed `items` values clear the row without using the outer text"): **agree in spirit, but the underlying premise needs correcting first.** See finding 2; the host cannot deliver a malformed `items` value to a client at all, and the interesting case is the empty one, which does not behave as the comment says.

**Codecov:** all modified lines covered, project 85.43%, +0.05%. No action.

## Reconciling with @Labeeb2339's review

I checked the two claims that would hurt most if wrong and I agree with both.

**Cross-session isolation holds.** `session_entries` filters on `e.session_id.as_deref() == Some(session_id)`, a full string equality, so two sessions sharing a prefix cannot cross over, a global entry (`session_id: None`) never matches any session, and an empty id matches only an empty id. That comparison is the only place a session id enters the badge path, so there is no second route to bypass: neither payload form, a malformed `items`, nor a list carrying entries for several sessions changes which entries are selected, because the filter runs before any payload is read. On the render side, `badge_cells` and `column_cells` are built by mapping over `state.sessions` in order and indexed by the same `enumerate` index in the item loop, so there is no row-to-session skew either. I could not construct a bypass.

**The badge-free path preserves the existing path column.** With no badges anywhere, `badge_width` is 0, so `column_budget` is the full `PLUGIN_REGION_MAX_WIDTH` and the badge `push_group` returns before pushing a single span. The remaining span sequence is byte-identical to main's: cells, inter-cell gaps, then `format!("{:pad$}{GAP}", "", pad = width - group_width)`. `no_plugin_entries_reserve_no_width` (main's test, unchanged here) asserts the path lands at column 41, and every other pre-existing alignment test in the file (`plugin_column_pads_so_the_path_stays_aligned`, `wide_glyphs_are_budgeted_by_terminal_cells`, `emoji_presentation_status_text_stays_within_budget`, `second_plugin_cell_is_dropped_when_the_budget_is_spent`) keeps its original assertions and passes.

Beyond that pass, three things I verified that are worth recording, and one finding I think it missed.

## Issues found

**1. Icon-only badges render nothing, and that appears to be the shape the GitHub plugin actually sends. Worth resolving before merge.**

This is the deliberate deviation the PR body flags, and the reasoning given for it is sound in isolation: a bullet preserves "something exists" while discarding which state it meant. My concern is not the reasoning, it is the consequence.

The canonical `row-badge` payload in this repository is icon-only. The host's own test at `src/plugin/ui_state.rs:1292` stores:

```json
{"items": [
  {"icon": "git-pull-request-arrow", "tone": "success", "href": "https://x/pr/1", "tooltip": "PR #1"},
  {"icon": "git-pull-request-draft", "tone": "warn"}
]}
```

with no `text` on either item. `web/tests/plugin-row-slots-mobile.spec.ts:42` models the same thing under the id `github_pr_badge`, twelve items of `{icon, tone, tooltip}`, and its header comment names it "the github plugin's row-badge (icon chips)". `BadgeItem.text` is `Option<String>` in `src/plugin/ui_state.rs:122`, and the plugin-system doc at line 709 describes an item as rendering "as a compact, tone-tinted icon (falling back to `text`)", so icon-first is the documented primary form.

I ran that exact host payload through the real `UiStore` and then through the new selector:

```
stored payload = {"items":[{"icon":"git-pull-request-arrow","tone":"success","href":"https://x/pr/1","tooltip":"PR #1"},{"icon":"git-pull-request-draft","tone":"warn"}]}
row_badge_cells  = []
```

So for the payload the codebase treats as canonical, the picker renders nothing at all. That is the first user story of #2947 ("when a session's branch has an open PR the GitHub plugin reports, then that session's row in the remote-home picker shows a tone-colored PR badge") and the PR's own manual test step 1, whose checkbox is unchecked, as is "For UI changes: included screenshot or recording".

I want to be careful about what I am and am not asserting. I verified that icon-only yields nothing. I could not verify what the real GitHub plugin sends, because it lives outside this repository; the two fixtures above are the best evidence available here and they both point the same way. So the ask is evidence, not a redesign:

- Cheapest resolution: attach the screenshot AGENTS.md asks for on a UI change, with the real GitHub plugin attached, showing a badge on a row. That settles it in one image and closes the checklist gap at the same time.
- If the plugin does send icon-only items, the smallest fix that satisfies the user story is a `tooltip` fallback for an item with no `text`, since the GitHub fixture carries `tooltip: "PR #1"`, which is genuinely useful terminal text:

  ```rust
  fn badge_cell(badge: &Value) -> Option<(String, Option<Tone>)> {
      let text = block_str(badge, "text").or_else(|| block_str(badge, "tooltip"))?;
      Some((text.to_string(), block_tone(badge)))
  }
  ```

  That does cut against the file's standing "tooltips have no terminal surface" convention, so it is a call for @njbrake rather than something to apply unilaterally. The alternative, a small lucide-name to short-label map, is more work for the same outcome.

The issue body itself anticipated this and asked for "a compact marker" for icon-only items. The PR argues that specific suggestion down convincingly. What it does not address is that rejecting the marker leaves the primary case rendering nothing at all.

**2. The `items: []` doc comment describes behavior the host cannot produce, and one test case asserts a shape that cannot reach the client.**

`src/tui/plugin_ui.rs:82` says "the documented `items: []` clears the row rather than falling back to the entry's own `text`", and the test table encodes `(json!({"text": "stale", "items": []}), &[])`.

`RowBadgePayload.items` is declared `#[serde(default, skip_serializing_if = "Vec::is_empty")]` (`src/plugin/ui_state.rs:151`), and `validate_payload` normalizes by deserialize-then-reserialize, so an empty `items` key is dropped before storage. Through the real store:

```
input  = {"text": "stale", "items": []}
stored = {"text":"stale"}
cells  = [("stale", None)]
```

The text is the fallback in production, which is the opposite of what the comment claims, and it matches the web (`objectList` returns undefined for a missing key, so the web also falls back to the single-badge form). The user-visible contract still holds for the pure `{"items": []}` case, which normalizes to `{}` and renders nothing, so this is a comment and test accuracy issue rather than a bug. Same reasoning applies to the malformed case: `{"items": 7}` is rejected by `validate_payload` with a `BadRequest` and never reaches a client, so the "malformed `items` clears the row" branch is defensive only. Worth saying that in the comment instead of describing it as an observable divergence from the web. This also supersedes CodeRabbit's suggestion to document the malformed behavior in the plugin contract: documenting a state the host rejects would make the contract less accurate, not more.

**3. `badge_free_rows_leave_the_whole_region_to_the_column` cannot fail independently, and does not test what its name says.**

```rust
let long = "x".repeat(PLUGIN_REGION_MAX_WIDTH + 10);
let state = state_with(&["s1"], json!([row_column("s1", &long)]));
let (_, width) = row_column_cells(&state, "s1", PLUGIN_REGION_MAX_WIDTH);
assert_eq!(width, PLUGIN_REGION_MAX_WIDTH);
```

It passes the budget in by hand, so it never exercises `render_list`'s badge-free budget selection, which is the behavior the name promises. And it is a strict subset of `long_plugin_text_is_capped_so_the_path_survives` twelve lines above it: same helper, same explicit budget, a proper subset of the assertions. Per AGENTS.md, either drop it or make it earn its place by going through the painted picker, which would also make it the true counterpart of `badges_share_the_plugin_region_instead_of_widening_it`:

```rust
let state = state_with(&["s1"], json!([row_column("s1", &long)]));
assert_eq!(column_of(&rows(&state), "/tmp/s1"), 41 + PLUGIN_REGION_MAX_WIDTH + 2);
```

Related and minor: the PR body says "6 ratatui `TestBackend` tests paint the real picker". Five of the six do; this one paints nothing.

**4. Nit: `badge_cell` duplicates the closure inside `row_badges`.** `src/tui/plugin_ui.rs:635` is `filter_map(|b| Some((block_str(b, "text")?.to_string(), block_tone(b))))`, which is now the body of `badge_cell`. `filter_map(badge_cell)` there would delete the duplicate and make the shared icon-only rule live in one place.

**5. Nit: the rebase left two ragged rewraps.** The module doc now breaks after "Icons, tooltips, hrefs, and the", and the doc paragraph after "project path. The snapshot". Cosmetic, purely from adapting to #3532's deletions.

## What was verified, and how

Recorded because these are the claims that would flip the verdict if wrong.

**Plugin-supplied text cannot inject escape sequences into the terminal.** Two independent checks. First, the pinned source: `ratatui-core 0.1.2` filters control characters on both paths the picker can take, `Span::styled_graphemes` at `src/text/span.rs:314` (`.filter(|g| !g.contains(char::is_control))`) and `Buffer::set_stringn` at `src/buffer/buffer.rs:351`. Second, an actual render: I pushed `{"text": "a\u{1b}[2Jb\u{1b}[31mc"}` through the real `UiStore` into the real `render()` against a `TestBackend` and dumped every cell symbol:

```
painted row = "▸  session s1                idle        a[2Jb[3…    /tmp/s1"
contains ESC = false
measured width = 10, path column = 53
```

The ESC bytes are dropped at the buffer, and because `UnicodeWidthStr::width` scores them 0 in the measurement too, the accounting stays consistent: 41 + 10 (the badge cap) + 2 = 53, precisely where the model says the path should land. A plugin cannot move the cursor, clear the screen, or desynchronize the layout through a badge. One residual caveat, not introduced here: `char::is_control` covers C0 and C1 but not Unicode format characters such as U+202E, which could visually reorder a row. That exposure is shared with `row-column`, session titles, and project paths and is pre-existing.

**Width accounting is in display columns and cannot split a cluster.** `truncate_to_width` steps by grapheme cluster and measures the accumulated prefix rather than summing per-piece widths, so a VS16 emoji, a combining mark, and CJK are all handled; `measured_cells` then clamps with `.min(budget - gap)` so a misbehaving helper cannot underflow the budget subtraction. The region total is bounded: with badges present, `badge_width + 2 + column_width` is at most 24 by construction, so the path never sits further right than main's worst case of 41 + 24 + 2. The 40-column case degrades rather than corrupts: a `Line` with no explicit alignment renders left-aligned, so the right side clips and the project path is what falls off, the same as on main, where the 41-cell prefix alone already exceeds 40.

**The rebase from `81e9d3f3` to `e11417fd` dropped nothing.** I fetched the pre-rebase head, computed its merge base (`fb0e7afb`), and diffed the two authored diffs against each other. The only substantive differences are the two places where the branch's edits adapt to #3532 removing the `settings-page` and `ToolCardBadge` slots from main. Every other difference is a line number or a blob hash. Nothing from main was reverted and nothing from the branch was lost.

## Documentation

Proportionate. Eighteen lines added, seven removed, all describing user-visible rendering behavior in the file that already documents the sibling `row-column` slot. No standalone markdown file, no narration of the change, no duplicated prose. The one fix is the accuracy issue in finding 2, which is a comment in `plugin_ui.rs` rather than the markdown.

## Suggestions

- The PR is right that responsive budgeting is a separate concern, since deriving the region from `area.width` would require making the title and status columns terminal-cell-aware first (they count chars via `truncate`). Worth a follow-up issue rather than a silent deferral, so the reasoning outlives the PR description.
- Consider a table row in `row_badge_cells_read_both_payload_forms` for a badge whose text is a lone combining mark or a zero-width joiner sequence. `truncate_to_width` handles it and `emoji_presentation_status_text_stays_within_budget` covers the width side, so this is optional.

## Merge risk

**Low for regression, moderate for the feature being inert.**

Blast radius on the regression side is small and well fenced. Every path that could shift the existing layout is asserted at the screen-column level by tests that predate this PR and still pass unchanged, and the shared 24-cell region means the project path is never further right than it is on main. Revert is clean: three commits, one crate, no schema, no stored data, no released artifact.

The residual risk is the one in finding 1. If the GitHub plugin sends icon-only items, this merges as a no-op for the plugin it names, the picker looks byte-identical to today for every real user, and nothing in the PR would surface that: the unit tests assert the icon-only case renders nothing, so they pass on the broken outcome. We would hear it from a user reporting that the TUI still shows no badge, and it would read as the same bug #2947 already describes. A screenshot on this PR is the cheapest thing that would surface it before merge.

Second-order: users running both the GitHub plugin and a `row-column` plugin will see their column text truncate 12 cells earlier than on main, because a badge anywhere in the list shrinks the column budget for every row. That is the documented and correct tradeoff, but it is a visible change for that group and no test names it as intentional.

## Verdict

**Design alignment:** Aligned
**Implementation:** Needs discussion
**Project fit:** Accept
**Documentation:** Proportionate, one accuracy fix
**Merge risk:** Low for regression, moderate that the feature is inert for the plugin it targets

The implementation is genuinely good work, @Seluj78. It extends the #2948 mechanism instead of building beside it, the width accounting is right in a place where it is usually wrong, plugin text cannot reach the terminal as escapes, session scoping is a single exact-equality filter with no second route around it, and the badge-free row is preserved down to the screen column. Clippy, fmt, and the full `--features serve` suite are clean locally.

I put it at Needs discussion rather than Approve on one fact I could not verify from this repository: whether the GitHub plugin sends `text` on its badge items. Everything in the tree, the host's own test fixture, the web Playwright fixture, and the plugin-system doc's description of `items`, says it sends icons, and for that shape this renders nothing. A screenshot of the real plugin in the picker, which AGENTS.md asks for on a UI change anyway, would move this to Approve; findings 2 through 5 are small enough to fold in alongside it.

njbrake (commented):

_Reviewed at `e11417fd`. Push new commits and it will be re-reviewed._

---

**Note on this one: the direction at the bottom is @njbrake's call, not just an agent's suggestion.** He read the icon-only question and decided how he wants it resolved. The evidence above it was gathered to answer a question the previous review left open.

## The open question is now settled: the real plugin sends no `text`

The previous review flagged that icon-only badge items render nothing, and was explicit that it could not verify what the real GitHub plugin sends, because the plugin lives outside this repository. I went and checked.

`plugins/featured.toml` names the plugin as `gh:agent-of-empires/plugin-github`. In that repository, `src/aoe_github_plugin/uistate.py:223`:

```python
def _chip(icon: str, tone: str, tooltip: str, href: str | None) -> dict[str, Any]:
    chip: dict[str, Any] = {"icon": icon, "tone": tone, "tooltip": tooltip}
    if href:
        chip["href"] = href
    return chip
```

`_chip` is the only constructor for `row-badge` items, and it never sets `text`. Every chip reaching the slot is `{icon, tone, tooltip}` plus an optional `href`: the PR affordance, the review-state chip, the CI-rollup chip, the unresolved-comments chip, the conflicts chip, and the per-repo error marker. All of them.

So with this PR as written, `badge_cell` returns `None` for every item the flagship plugin emits, and the row-badge region in the picker renders nothing at all. That is the first user story of #2947, and it is the PR's own manual-test step 1.

This is not an objection to the mechanism. The measure, pad, and paint work is correct, it reuses the #2948 pipeline rather than forking it, and the tests are the right shape and the right tier. It is a question about what the slot displays.

## Why the tooltip fallback is not the answer

The previous review floated falling back to `tooltip` when an item has no `text`. Having read the plugin, that would be worse than blank. Its tooltips are sentences, not labels. From `_pr_badge_chips`:

```python
head = f"{repo_name}: PR #{pull.get('number', '?')} {pull.get('title', '')}".rstrip()
chips.append(_chip(icon, tone, f"{head} (CI {label})", href))
```

which produces strings on the order of `agent-of-empires: PR #3534 feat(tui): render the row-badge plugin slot (CI failing)`. Four of those concatenated into `ROW_BADGE_MAX_WIDTH` gives truncated noise. Please disregard that suggestion.

## Why changing the plugin instead is the expensive path

The natural reading is "add `text` to `_chip` and this fixes itself." That would regress the web. `BadgeChip` in `web/src/components/plugin/PluginSlots.tsx:98` renders both fields, not one or the other:

```tsx
{iconComp && createElement(iconComp, { className: "size-3 shrink-0", "aria-hidden": true })}
{text && <span className="truncate">{text}</span>}
```

The guard above it is `if (!iconComp && !text) return null`, so an item carrying both shows an icon followed by a word. The plugin emits up to four chips per PR, so adding `text` turns four compact icon pills into four worded pills on every session row in the dashboard. That is a visible change to a surface this PR is not trying to change.

Making the plugin route work properly therefore means a plugin change, plus a web change so an icon suppresses the text, plus a plugin release and a pin update in `plugins/featured.toml` before this PR can be demonstrated. Three repositories and a release cycle to light up one terminal region.

**Related documentation bug, worth fixing whatever else happens here.** `docs/development/internals/plugin-system.md:709` says each item renders as "a compact, tone-tinted icon (falling back to `text`)". There is no fallback; the code above renders both. That sentence is inaccurate on `main` today, and it is precisely the sentence that would send someone down the plugin-first path.

## What @njbrake would like done

Keep it in the host, in this PR:

1. **Add a small icon-name to short-label map** so the badge renders something for icon-only items. The names are a closed set on the plugin side (`_ICON_OPEN`, `_ICON_DRAFT`, `_ICON_ERROR`, and the values in `_REVIEW_VISUAL` and `_CHECK_VISUAL`), and the terminal budget is a host constraint, so the mapping belongs where the column arithmetic already lives rather than in every plugin that wants a terminal surface. An unmapped icon with no `text` can keep today's behavior and contribute nothing.
2. **Fix the `plugin-system.md:709` fallback sentence** while you are in that file.
3. **Attach the screenshot** AGENTS.md asks for on a UI change, with the real plugin loaded and a badge visible on a row. That closes the unchecked box and demonstrates step 1 at the same time.

That keeps this to a single repository, with no plugin release in the dependency chain and no change to the web's appearance.

**One alternative, if you would rather argue for it.** The `row-column` slot from #2948 already renders this plugin's compact word labels in the picker, since its payload is `{"text": label, "tone", "tooltip", "sort_value"}` built from `_ATTENTION_VISUAL`. A reasonable position is that icons are a web affordance, the terminal's channel is `row-column`, and the badge slot should not exist in the picker at all. That would make this PR smaller rather than larger. If you think that is the better design, say so rather than implementing the map; it is a live option and not a rejection of your work.

Everything else in the previous review still stands: definition of done is clean, the CodeRabbit out-of-scope warning is a false positive from a rebase-spanning range, and the decision to skip an e2e test is the AGENTS.md cost calculus applied correctly.

jerome-benoit (changes requested):
Requesting changes at `25d5ae6b746a4644dcf3697f77efedeb7ad263d7` because the previously identified product-level issue remains unresolved.

**P1: render the payload emitted by the featured GitHub plugin**

`row_badge_cells` only retains badges carrying `text`, and the new test explicitly asserts that an icon-only badge renders nothing. However, the featured `agent-of-empires/plugin-github` v2.0.0 worker constructs every badge item as `{icon, tone, tooltip}` plus an optional `href`, with no `text`, and emits those items directly into `row-badge`.

Consequently, every badge from the integration targeted by #2947 is filtered out, so the native picker still shows no PR badge. Please provide a meaningful terminal rendering for the actual icon-only payload and add a regression test using that payload shape. A screenshot or recording with the featured plugin visible would validate the acceptance story.

The shared-width layout, terminal-cell accounting, session isolation, and semantic tone styling otherwise look sound. All current CI checks pass, but the current test suite encodes the invisible result rather than detecting it.

PR #3535, opened 2026-08-25
fix(test): unbreak the macOS lib suite and de-race the live session-list reads

Seluj78:
## Description

Main has been red on the **Tests** workflow for every commit for at least two days. Two distinct problems, both in test code:

**1. `Cargo Test (macos)` fails 100% of runs (the actual blocker).**
`git::worktree::tests::worktree_move_observation_canonicalizes_symlinked_parents` (added in #3457) compares the output of `canonicalize_move_endpoint` against a raw `tempfile::tempdir()` path. On macOS that temp dir lives under `/var`, which is itself a symlink to `/private/var`, so the two sides can never be equal:

```
left:  "/private/var/folders/df/.../T/.tmpkg7IXg/real/old"
right: "/var/folders/df/.../T/.tmpkg7IXg/real/old"
```

Linux CI is unaffected because `/tmp` is real there. Fix: canonicalize `real_parent` before building the expectation, so both sides are canonical on either platform. Production code is untouched, `canonicalize_move_endpoint` was behaving correctly.

**2. `Playwright (live)` flakes on the session list (~8% of runs) — and the server race behind it.**

`wizard-scratch-launch` failed with `expect(sessions).toHaveLength(1)` receiving `[]`, on the line *after* an `expect.poll` had already observed a session.

The first version of this PR treated that as a test-side read-ordering issue. @jerome-benoit correctly pushed back: it is a real server-side race, and the helper was hiding it. It is now fixed properly.

`reload_state_instances_from_disk` builds `merged` exclusively from the caller's `fresh` disk snapshot and then does `*current = merged` (`src/server/mod.rs:3432`), so a reload carries no memory of rows it never saw. `create_session` persists the new row to `sessions.json` **before** upserting it in-memory (documented on `upsert_instance`, `src/server/api/sessions.rs:5262`). So a `status_poll_loop` tick whose disk read started before that persist has a `fresh` without the new row, and the wholesale replace drops the session the create just inserted. `GET /api/sessions` loses it until the next tick re-reads disk 2s later. `delete_epoch` did not cover this: no delete occurred.

The fix reuses the existing mechanism rather than adding a parallel one:

- `delete_epoch` becomes `mutation_epoch`; it guards session-set **membership** in both directions, not deletes only. Field doc and invariant 8 updated.
- The create path bumps it (`src/server/session_spawn.rs:488`) under the `instances` write lock the upsert already holds, so the insert and the bump land as one step against a concurrent reload, exactly as the delete path does at `sessions.rs:4424`.
- Threaded through `SessionService` as a shared `Arc`, same pattern as `telemetry_session_creates`.

On the test side, `waitForSessions()` returns the snapshot the poll settled on (removing the second racy fetch) across five specs, and `wizard-scratch-launch` now asserts the row **stays** listed across more than one 2s poll tick rather than only that it appeared.

New coverage, both mutation-checked:

| Test | Fails when |
| --- | --- |
| `a_reload_predating_a_create_does_not_drop_the_new_row` | the epoch comparison is disabled |
| `the_create_bumps_the_mutation_epoch_under_the_instances_lock` | the bump moves after `drop(instances)` |

The first also pins the pre-fix outcome, so the bump cannot later be removed as redundant.

### Known remaining flake, not fixed here

`acp-stories/queue-follow-up-with-nav` failed once in the last 25 main runs. That one is **not** a test-side race: the attached `debug.log` stops dead at `18:54:56.9` while the test ran until `18:55:22`, the fake agent log shows only a single `session/prompt`, and the page snapshot shows `Network error sending prompt: Failed to fetch` plus `Structured view worker stopped`. The serve daemon went away mid-test. That deserves its own investigation rather than a timeout bump here, so it is deliberately left alone.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

Verified locally on macOS with the exact command the failing job runs:

```
cargo test --features serve,e2e-tests --lib --bins
# 6417 passed, 2 ignored   <- was 6414 passed; 1 failed on main
cargo fmt --check      # clean
cargo clippy --all-features  # clean
cd web && npx tsc --noEmit && npm run format:check && npm run lint  # all clean
```

The four live specs are left for CI to run.

## Test Coverage Analysis

- [x] Added or updated a Playwright spec under `web/tests/`

`wizard-scratch-launch.spec.ts` gained a real regression assertion for the stale-create race (the session must stay listed across more than one poll tick). No `coverage-matrix.json` entry: the matrix tracks dashboard flows, and no flow changed. The server fix's primary coverage is the two Rust tests above.

**TUI / CLI (e2e test)**

Both changes are fixes to existing tests. The macOS fix restores a regression test that was already asserting the right behavior; the live-spec change removes a racy read.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Failure triage across 25 main runs, root-causing, and the fixes were done with Claude Code. Playwright trace artifacts were downloaded and inspected to separate the test-side race from the daemon-death flake.

## Summary

- Fixed a macOS worktree test that failed because temporary paths can use `/var` or `/private/var`.
- Added `waitForSessions()` to centralize session polling and improve timeout diagnostics.
- Updated live tests to use settled session snapshots and reduce duplicated polling logic.
- Fixed a server race that could restore deleted sessions or remove newly created sessions during stale disk reloads.
- Improved test coverage for session creation, deletion, and reload ordering.

## Technical details

- Replaced deletion-only tracking with shared `mutation_epoch` protection for session creation and deletion.
- Routed session removals through a shared helper that updates the mutation epoch.
- The separate `queue-follow-up-with-nav` daemon failure remains unresolved for future investigation.

### DIFF
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -4422,7 +4422,7 @@ async fn purge_session_artifacts(
         // reload cannot slip between them. See invariant 8 on
         // `reload_state_instances_from_disk`.
         state
-            .delete_epoch
+            .mutation_epoch
             .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
     }
     state.instance_locks.write().await.remove(id);
--- a/src/server/api/sessions.rs
+++ b/src/server/api/sessions.rs
@@ -4296,7 +4296,11 @@ async fn purge_session_artifacts(
         crate::session::deletion::PurgeReservation::Rejected(result) => {
             return match result.disposition {
                 crate::session::deletion::DeletionDisposition::AlreadyGone => {
-                    state.instances.write().await.retain(|row| row.id != id);
+                    remove_instance(
+                        &mut *state.instances.write().await,
+                        id,
+                        &state.mutation_epoch,
+                    );
                     state.instance_locks.write().await.remove(id);
                     Ok((true, result.messages))
                 }
@@ -4337,8 +4341,14 @@ async fn purge_session_artifacts(
             Err(result) => *result,
             Ok(committed) => {
                 // Remove the local mirror before awaiting ACP so the reconciler
-                // cannot surface a durable row that no longer exists.
-                state.instances.write().await.retain(|row| row.id != id);
+                // cannot surface a durable row that no longer exists. Bumps the
+                // epoch under the same lock: the ACP teardown below is slow, and
+                // a reload landing inside it would otherwise restore the row.
+                remove_instance(
+                    &mut *state.instances.write().await,
+                    id,
+                    &state.mutation_epoch,
+                );
 
                 // The worker may still use the worktree, so ACP teardown stays
                 // ahead of sidecar cleanup. The durable row is already gone.
@@ -4412,18 +4422,15 @@ async fn purge_session_artifacts(
     }
 
     {
-        let mut instances = state.instances.write().await;
-        instances.retain(|i| i.id != id);
         // The row is now gone from both disk and memory, so any reloader still
         // carrying a `sessions.json` snapshot that predates either removal must
-        // drop it rather than fold the deleted row back in. Bump while still
-        // holding the `instances` write lock: a reloader checks the epoch under
-        // that same lock, so the removal and the bump land as one step and a
-        // reload cannot slip between them. See invariant 8 on
-        // `reload_state_instances_from_disk`.
-        state
-            .mutation_epoch
-            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        // drop it rather than fold the deleted row back in. `remove_instance`
+        // bumps while still holding the `instances` write lock: a reloader
+        // checks the epoch under that same lock, so the removal and the bump
+        // land as one step and a reload cannot slip between them. See
+        // invariant 8 on `reload_state_instances_from_disk`.
+        let mut instances = state.instances.write().await;
+        remove_instance(&mut instances, id, &state.mutation_epoch);
     }
     state.instance_locks.write().await.remove(id);
     if let Some(entry) = recent_entry {
@@ -5275,6 +5282,35 @@ pub(crate) fn upsert_instance(
         *existing = instance;
     } else {
         instances.push(instance);
+    }
+}
+
+/// Remove `id` from the live registry, bumping `mutation_epoch` when a row was
+/// actually removed.
+///
+/// The delete path removes a row from `state.instances` in three places: the
+/// `AlreadyGone` short-circuit, the structured purge's early mirror removal
+/// (which then awaits ACP teardown before the handler finishes), and the final
+/// commit block. Every one of them has to bump, and has to bump while the
+/// caller still holds the `instances` write lock, because a reloader compares
+/// the epoch under that same lock. A removal that skips the bump leaves a
+/// window where a disk reload carrying a pre-delete snapshot rebuilds
+/// `state.instances` from it and puts the deleted row back, so
+/// `GET /api/sessions` lists a session the user just deleted.
+///
+/// Bumping only on an actual removal keeps the final commit block from
+/// spending an epoch when the early removal already took the row; if a stale
+/// reload DID restore it in between, the retain here finds it, removes it
+/// again, and bumps as it should.
+pub(crate) fn remove_instance(
+    instances: &mut Vec<crate::session::Instance>,
+    id: &str,
+    mutation_epoch: &std::sync::atomic::AtomicU64,
+) {
+    let before = instances.len();
+    instances.retain(|i| i.id != id);
+    if instances.len() != before {
+        mutation_epoch.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
     }
 }
 
@@ -7667,6 +7703,43 @@ pub async fn serve_session_artifact(Path((id, path)): Path<(String, String)>) ->
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// `remove_instance` is the only way a row leaves `state.instances` on the
+    /// delete path, so the epoch bump has to be tied to an actual removal
+    /// rather than to reaching the call. Bumping unconditionally would spend
+    /// an epoch on the final commit block after the structured purge's early
+    /// removal already took the row, dropping a reload that was perfectly
+    /// valid; not bumping at all leaves the window a stale reload uses to put
+    /// a deleted row back.
+    #[test]
+    fn remove_instance_bumps_the_epoch_only_when_it_removes_a_row() {
+        let epoch = std::sync::atomic::AtomicU64::new(0);
+        let read = || epoch.load(std::sync::atomic::Ordering::SeqCst);
+        let mut instances = vec![
+            Instance::new("keep", "/tmp/keep"),
+            Instance::new("doomed", "/tmp/doomed"),
+        ];
+        let doomed_id = instances[1].id.clone();
+
+        remove_instance(&mut instances, &doomed_id, &epoch);
+        assert_eq!(read(), 1, "a real removal bumps");
+        assert_eq!(
+            instances
+                .iter()
+                .map(|i| i.title.as_str())
+                .collect::<Vec<_>>(),
+            vec!["keep"]
+        );
+
+        // The structured purge reaches the final commit block after its early
+        // removal already took the row. Nothing left to remove, nothing to
+        // invalidate, so no epoch is spent.
+        remove_instance(&mut instances, &doomed_id, &epoch);
+        assert_eq!(read(), 1, "a no-op removal does not bump");
+
+        remove_instance(&mut instances, "never-existed", &epoch);
+        assert_eq!(read(), 1, "an unknown id does not bump");
+    }
     fn build_rename_test_state(
         persisted: Vec<Instance>,
         cached: Vec<Instance>,
--- a/web/tests/helpers/aoeServe.ts
+++ b/web/tests/helpers/aoeServe.ts
@@ -186,6 +186,31 @@ export async function listSessions(
 }
 
 /**
+ * Poll `GET /api/sessions` until at least one session is present, and
+ * return the snapshot the poll settled on. The list is a cache the
+ * daemon reconciles on a 2s tick (see `waitForView` below), so a fresh
+ * `listSessions()` issued right after a poll that already saw the
+ * session can come back empty; reading the array from inside the poll
+ * removes that second, racy fetch.
+ */
+export async function waitForSessions(
+  baseUrl: string,
+  timeout = 15_000,
+): Promise<Awaited<ReturnType<typeof listSessions>>> {
+  let settled: Awaited<ReturnType<typeof listSessions>> = [];
+  await expect
+    .poll(
+      async () => {
+        settled = await listSessions(baseUrl);
+        return settled.length;
+      },
+      { timeout },
+    )
+    .toBeGreaterThan(0);
+  return settled;
+}
+
+/**
  * Poll `GET /api/sessions` until the given session's `view` reaches
  * `expected`. The sessions list is a cache the daemon reconciles on a
  * 2s tick, so a disk snapshot taken just before an endpoint's write
--- a/web/tests/helpers/aoeServe.ts
+++ b/web/tests/helpers/aoeServe.ts
@@ -204,7 +204,11 @@ export async function waitForSessions(
         settled = await listSessions(baseUrl);
         return settled.length;
       },
-      { timeout },
+      {
+        timeout,
+        intervals: [100, 200, 400],
+        message: `at least one session should appear in GET /api/sessions within ${timeout}ms`,
+      },
     )
     .toBeGreaterThan(0);
   return settled;
--- a/src/git/worktree.rs
+++ b/src/git/worktree.rs
@@ -1871,6 +1871,9 @@ mod tests {
         std::fs::create_dir(&real_parent).unwrap();
         let linked_parent = dir.path().join("linked");
         std::os::unix::fs::symlink(&real_parent, &linked_parent).unwrap();
+        // macOS hands out temp dirs under `/var`, itself a symlink to
+        // `/private/var`, so the expectation has to be canonical too.
+        let real_parent = real_parent.canonicalize().unwrap();
         let from = linked_parent.join("old");
         std::fs::create_dir(&from).unwrap();
         let to = linked_parent.join("new");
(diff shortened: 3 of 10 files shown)


web/tests/helpers/aoeServe.ts
```diff
@@ -185,6 +185,31 @@ export async function listSessions(
...
+  await expect
+    .poll(
+      async () => {
+        settled = await listSessions(baseUrl);
+        return settled.length;
+      },
+      { timeout },
+    )
+    .toBeGreaterThan(0);
+  return settled;
```
jerome-benoit, line 214: **[P2] Keep the regression signal for the stale-create race**

Returning the first non-empty snapshot makes these tests accept a real server-visible `1 -> 0` regression. A reload can read `sessions.json` before creation persists the new session; creation then persists and upserts it into `state.instances`, after which the older reload drains the current state and rebuilds it exclusively from its stale snapshot. `delete_epoch` protects concurrent deletions, but there is no equivalent protection for creation.

This is precisely why the previous second `listSessions()` could return an empty list. Please fix the create/reload ordering, for example with a mutation generation or by preserving rows created after the reload began, and retain an assertion that the session remains present across a subsequent read. Otherwise this helper removes the test signal while leaving the user-visible cache regression in place.

jerome-benoit (changes requested):
The macOS path canonicalization change looks correct, but I am requesting changes because the Playwright adjustment masks the underlying session-cache race instead of fixing it.

A stale disk reload can overwrite the in-memory row after a successful create. Returning the first non-empty response only makes the tests stop observing that regression.

njbrake (commented):

_Reviewed at `e6553b57`. Push new commits and it will be re-reviewed._

---

## Summary

Two test-only repairs. The macOS `Cargo Test` job has failed on **every commit on main since `d91186f8` (#3457, 2026-08-21)**: 23 consecutive runs over four days, not two. This PR fixes the single test responsible, and separately removes a redundant post-poll `listSessions()` from four live specs.

Everything below marked **verified** was executed locally against head `e6553b57` rebased on `origin/main` (`a0137269`); the branch is already on top of main, so no rebase was needed.

## Design alignment

**Aligned.**

`waitForSessions` sits directly beside `waitForView` in `web/tests/helpers/aoeServe.ts:198`, the precedent it names. Same shape (exported async `expect.poll` wrapper), same 2s-reconcile-tick rationale, and it cross-references its sibling in the docstring. This extends the existing mechanism rather than adding a parallel one. The Rust change is three lines inside an existing test.

## Verification

**The Rust diagnosis is exact.** Reverting only the 3 added lines reproduces the failure verbatim:

```
left:  "/private/var/folders/.../T/.tmpf90kaO/real/old"
right: "/var/folders/.../T/.tmpf90kaO/real/old"
```

That matches main's own CI log character for character. On main's macOS run (job 97926584478): `test result: FAILED. 6414 passed; 1 failed; 2 ignored`, and `worktree_move_observation_canonicalizes_symlinked_parents` is the **only** `FAILED` line in the whole run.

**The fix does not turn the test into a tautology.** This was the specific failure mode worth ruling out, so the production function was mutated twice and the test goes red both times:

1. Short-circuit `canonicalize_move_endpoint` to `path.to_path_buf()`: red, `left: ".../linked/old"` vs `right: ".../real/old"`. The symlink resolution is still asserted.
2. Break only the not-yet-existing-path fallback (`parent.canonicalize().map(|_| path.to_path_buf())`): red on the `to` endpoint. Both branches of `src/git/worktree.rs:132-144` remain load-bearing.

On Linux the added line is a no-op (`/tmp` is real), so the assertion's strength there is literally unchanged.

**Production code is genuinely untouched.** `#[cfg(test)]` is at `src/git/worktree.rs:1788`, `mod tests` at 1789; the added lines land at ~1874. The body's claim holds.

**The Playwright helper cannot mask a genuine failure.** A throwaway spec was written to drive the real `waitForSessions` against a stub HTTP server, then deleted. Three properties, all confirmed:

- Serving `[]`, `[]`, `[]`, `[one session]`, then `[]` forever: the helper returns the one-session snapshot. This is precisely the CI flake, and the old code would have failed here.
- Sessions never appear, `timeout: 2_000`: throws in 1.9s. It does not hang, and it does not return `[]`.
- Endpoint returns 500: propagates `GET /api/sessions failed: 500` after exactly **1** call, no 15s retry loop.

**No assertion was dropped.** Each spec's removed 8 lines were the poll (`timeout: 15_000`, `.toBeGreaterThan(0)`) plus the second fetch. The helper's default timeout is 15_000 and it uses the same matcher; every `expect(sessions).toHaveLength(1)` and downstream assertion is preserved.

**The Playwright flake is real, not hypothetical.** Main run 32859507230, shard 2/2: `tests/live/wizard-scratch-launch.spec.ts:40` fails `expect(received).toHaveLength(expected)` with `Received array: []`. That is the exact line the second fetch fed.

## Checks executed

| Check | Result |
| --- | --- |
| `cargo test --features serve --lib` | **6415 passed; 0 failed; 2 ignored** (no caveat) |
| `cargo clippy --all-features -- -D warnings` (CI's exact invocation, `ci.yml:51`) | clean |
| `npm run format:check` | clean (oxfmt, 932 files) |
| `npm run lint` | clean |
| `npx tsc --noEmit` | clean |

Main's macOS CI was 6414/1; head is 6415/0. Changing that count was the point, and it changed.

The live Playwright suite was **not** run (needs a built binary and real daemons). The four specs and the helper were read carefully and the helper was exercised in isolation as described above.

`cargo clippy --features serve --all-targets -- -D warnings` surfaces 5 lints, all in files this PR does not touch (`src/acp/background_agent.rs:627,745`, `src/acp/supervisor.rs:4030`, `src/process/macos.rs:266`, `src/session/smart_rename.rs:1546`). Pre-existing, and CI does not pass `--all-targets`, so not this PR's problem.

## Two things reviewers should update their priors on

**1. The "two known macOS failures" advice was wrong, and not in the direction anyone assumed.**

`tui::dialogs::hooks_install::tests::test_content_shows_settings_path` **passes on macOS CI**, visible in main's own run log alongside the failure. It fails only in a local dev shell that exports `CLAUDE_CONFIG_DIR`; both sides reproduce (fails with `CLAUDE_CONFIG_DIR` set to a custom path, passes under `env -u CLAUDE_CONFIG_DIR`). It was never a macOS problem and never a CI problem.

So the standing advice becomes: **after this merges, zero known failures.** If your shell sets `CLAUDE_CONFIG_DIR`, that one test fails locally and always did. The title "unbreak the macOS lib suite" is accurate, not an overstatement.

Separately worth an issue, not this PR: `test_content_shows_settings_path` reads process env with no isolation, so it is environment-dependent by construction.

**2. This PR's own green `Cargo Test (macos)` check proves nothing.**

`.github/workflows/tests.yml:171-193`: every step in that job carries `if: github.event_name != 'pull_request'`. It runs on push to main and reports success in ~5s on PRs purely to satisfy a required-check ruleset. That is also exactly how #3457 landed a macOS-breaking test. The fix will only be validated on the merge commit, which is why red-then-green was reproduced locally rather than trusting the check.

## Issues found

**Non-blocking, one missed sibling, in a file the PR already edits.**

`web/tests/live/wizard-acp-create-session.spec.ts:80`, the *second* test in that file, has the same exposure:

```ts
await waitForStructuredView(page);
await expect(page.getByRole("button", { name: /Agent \(full access\)/ }).first()).toBeVisible({ timeout: 15_000 });

const sessions = await listSessions(serve.baseUrl);
expect(sessions).toHaveLength(1);
```

A bare, unretried `listSessions()` asserting `toHaveLength(1)` immediately after a UI-side confirmation of a freshly launched session. The stale-snapshot window does not care whether an `expect.poll` preceded it. One-line fix, and `waitForSessions` is already imported at line 7:

```ts
const sessions = await waitForSessions(serve.baseUrl);
```

The other 60-odd `listSessions` call sites were swept for the same shape with no further misses:

- `acp-stories/wizard-launch-button.spec.ts:36` polls but then asserts on the sidebar locator, no refetch.
- `wizard-recent-projects-persist.spec.ts:31` and `wizard-scratch-delete-cleans-tempdir.spec.ts:53` poll toward 0 (deletion), the safe direction.
- Everything else reads a session seeded via `seedSessionViaAoeAdd` before the daemon starts, so the cache is warm.

## Documentation

Proportionate. Two lines of "why" comment on the Rust test and an 8-line docstring on `waitForSessions` that names the mechanism and points at `waitForView`. Both are the earned kind: non-obvious reasoning invisible from the code. No emdashes in the added prose. No standalone markdown file, and correctly no `coverage-matrix.json` entry, since no dashboard flow changed.

## Suggestions

1. `waitForSessions` omits the `message` and `intervals` options its sibling passes (`aoeServe.ts:238-241`), so a timeout produces a generic `expect.poll` failure rather than "no session appeared within 15s". Worth a `message` for parity.
2. Follow-up issue for `test_content_shows_settings_path` reading process env unisolated.
3. Follow-up issue for the `Cargo Test (macos)` PR blind spot. The workflow comment already anticipates this tradeoff, but #3457 is now a concrete instance of the cost, and it may be worth revisiting whether `--lib` alone could run on PRs.

## Merge risk

**Low.** Test-only; the three Rust lines are inside `#[cfg(test)]`, and the mutation testing proves the assertion is still load-bearing rather than self-satisfying. Reverts cleanly.

The one residual: the four Playwright specs were not executed here, only read and reasoned about, with the helper exercised in isolation. If `waitForSessions` were wrong it would surface as a loud, bounded test failure on the next live run rather than a silent false green, because the timeout path throws and the returned array is non-empty by construction.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Merge risk:** Low; no residual concern beyond the live specs not being run locally

This is a small, unglamorous, high-value PR and it does what it says. It ends four days of red main, it is honest about the flake it deliberately does not fix (`queue-follow-up-with-nav`, confirmed to be the remaining Playwright failure on main's latest run and a daemon-death problem rather than a test-side race), and the Rust fix survives mutation testing on both branches of the production function. The one missed sibling at `wizard-acp-create-session.spec.ts:80` is a one-line follow-up, not a reason to hold the PR that unblocks the branch.

**LGTM** — merge it. The macOS suite goes 6414/1 to 6415/0, the fix is not a tautology, the Playwright helper is bounded and loud, and the only gap is one more call site of the same shape that can land separately.

web/tests/helpers/aoeServe.ts
Seluj78: You are right, and thank you for pushing back rather than letting the helper land as-is. Fixed in https://github.com/agent-of-empires/agent-of-empires/commit/78ba30711ff982c8ac0f905347504088b0387e0e.

Confirmed the race exists exactly as you described. `reload_state_instances_from_disk` builds `merged` exclusively from the caller's `fresh` snapshot and then does `*current = merged` (`src/server/mod.rs:3432`), so a reload carries no memory of rows it never saw. `upsert_instance` already documents that `create_session` persists to `sessions.json` **before** pushing in-memory (`src/server/api/sessions.rs:5262`). So a `status_poll_loop` tick whose disk read starts before that persist has a `fresh` without the new row, and the wholesale replace drops what the create just inserted. `delete_epoch` did not fire because no delete happened.

The fix follows the mechanism you suggested rather than adding a parallel one:

- `delete_epoch` is now `mutation_epoch`, since it guards session-set **membership** in both directions rather than deletes only. Field doc and invariant 8 updated to say so.
- The create path bumps it at `src/server/session_spawn.rs:488`, under the `instances` write lock the upsert already holds, so the insert and the bump land as one step against a concurrent reload, exactly as the delete path does at `sessions.rs:4424`.
- Threaded through `SessionService` as a shared `Arc`, same pattern as `telemetry_session_creates`.

On retaining the signal, there are now three checks rather than the accidental one the second `listSessions()` provided:

1. `a_reload_predating_a_create_does_not_drop_the_new_row` covers the reload behaviour, and also pins the pre-fix outcome so the bump cannot later be deleted as redundant.
2. `the_create_bumps_the_mutation_epoch_under_the_instances_lock` is a source-level guard that the bump stays inside the lock scope. Behavioural coverage there would need a real spawn; the failure mode worth catching is a future edit moving the bump after `drop(instances)`.
3. `wizard-scratch-launch.spec.ts` now asserts the row stays listed across more than one 2s poll tick, instead of only that it appeared once.

Both Rust tests were mutation-checked: disabling the epoch comparison fails (1), and moving the bump after `drop(instances)` fails (2). `cargo test --features serve,e2e-tests --lib --bins` is 6417 passed / 0 failed.

src/server/api/sessions.rs
```diff
@@ -4422,7 +4422,7 @@ async fn purge_session_artifacts(
         // reload cannot slip between them. See invariant 8 on
         // `reload_state_instances_from_disk`.
         state
-            .delete_epoch
+            .mutation_epoch
```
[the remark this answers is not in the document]
Seluj78: Good catch, both sites confirmed. Fixed in https://github.com/agent-of-empires/agent-of-empires/commit/57ace7e8e9ebf5d897da536819cd1d301fc840ff.

Verified against the current code rather than taken on trust:

- `AlreadyGone` (`sessions.rs:4299`) removed the row and returned `Ok` immediately, never bumping.
- The structured purge (`sessions.rs:4341`) removed the local mirror and then awaited `shutdown_and_delete`, so the only bump was in the final commit block well after a genuinely slow teardown.

Implemented as you suggested: a `remove_instance` helper that bumps under the `instances` write lock the caller already holds, mirroring `upsert_instance` on the create side, with all three removal sites routed through it. `grep` confirms no raw `retain(|row| row.id != id)` survives outside the helper.

On "increment only when the helper removed a row", that is what it does, and it matters in both directions. It keeps the final commit block from spending an epoch when the early removal already took the row, which would drop a perfectly valid reload; and if a stale reload DID restore the row in between, the retain there finds it, removes it again, and bumps as it should. The new test pins all three cases and was mutation-checked: making the bump unconditional fails on `a no-op removal does not bump`.

One scoping note for the record. Both windows you found are transient in a way the create bug was not. In the structured-purge case the final retain+bump still runs, so the end state was already correct and the exposure was a `GET /api/sessions` listing a deleted session during teardown. In the `AlreadyGone` case the row is already absent from storage, so the next disk read does not contain it and the resurrection self-corrects within one 2s tick. Fixed anyway, since the epoch now claims to cover session-set membership in both directions and a partial guarantee there is worse than none.

`cargo test --features serve,e2e-tests --lib --bins` is 6418 passed / 0 failed; `clippy --all-features` and `fmt` clean.

PR #3537, opened 2026-08-26
fix(tui): remove every synchronous tmux command from the paint path

jerome-benoit:
## Description

The TUI paint path could synchronously invoke tmux when display snapshots expired, names were resolved, preview capture fell back, or passive resize fired. A stalled tmux server could therefore freeze rendering.

This change makes paint cache-only:

- a background snapshot poller refreshes session and pane metadata;
- display name and liveness lookups never invoke tmux;
- `LiveCaptureWorker` is the single preview capture source and fences retargeted generations;
- passive resize runs in the dedicated `aoe-passive-resize` worker;
- VT ownership, pipe ownership, resync, and cleanup share one bounded command deadline;
- failed authoritative refreshes retain a healthy VT source and retry at the rearm cadence.

The ownership and capture changes remain in this PR because they enforce one end-to-end invariant: moving capture and resize off paint requires the worker lifecycle, fencing, cleanup, and deadlines to agree. Splitting them would create an intermediate path with mismatched ownership semantics.

## Reproduction

A thread-local probe at `tmux_command()` counts forks by thread. Rendering Structured, Terminal, and Tool views across cold, absent, and expired cache states previously produced four tmux forks on the paint thread over two frames. The regression test now asserts zero.

## Risks

- VT snapshot installation races with live pipe chunks. Generation and received/settled sequence fences preserve the newer grid.
- Multiple viewers can compete for one `pipe-pane`. Lease and pipe-owner tokens fence stale teardown.
- A tmux timeout can occur during ownership recovery. Every command in one capture cycle consumes the same deadline.
- Display snapshots can become stale during refresh failure. Display keeps the last good snapshot, while lifecycle readers preserve authoritative `Unknown` semantics.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Test Coverage Analysis

- Unit tests cover zero paint-thread forks, stale display snapshots, VT generation and settlement races, deadline exhaustion, ownership fencing, resize queueing, and authoritative refresh cadence.
- Real tmux coverage exercises owner values containing literal `{` and `:`.
- Existing takeover and non-bracketed paste end-to-end scenarios cover the affected live paths.

## Checklist

- [x] No pixel-visible layout change is intended

## AI Usage

- [x] AI was used


src/tui/home/tests.rs
```diff
@@ -21066,56 +21142,296 @@ mod profile_duplicate_reconciliation {
...
+
+    assert_ne!(
+        env.view.preview_cache.content, "previous pane bytes",
+        "a stale-generation frame must be dropped, never applied"
+    );
+}
+
+/// Regression: entering live-send while a blocking capture is in flight must
+/// revalidate the empty-frame policy on the consumer. An empty frame captured
+/// just before the transition cannot blank the agent/tool pane under the
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in `4f95f96b`. `stop_for_test` now owns and joins the capture thread before the test stages cycle 17, so the liveness assertion no longer depends on a fixed sleep or a racing worker update. The same commit also applies the review-summary nit by distinguishing `forward_empty_policy` from the later live-state-expanded forwarding value. Targeted test, formatting, production/serve clippy, test clippy, and the 5,005-test default suite pass.

src/tui/home/live_send.rs
```diff
@@ -806,13 +936,14 @@ pub(in crate::tui) struct LiveCaptureWorker {
...
-    /// the worker dedups unchanged frames, so an idle pane publishes nothing
-    /// for minutes while still cycling.
+    /// Cycle counter, bumped before every deadline-bounded sample. Changed
+    /// frames cannot serve as a heartbeat because idle content is deduplicated;
+    /// render observes this counter and replaces a worker that stops advancing.
     cycles: std::sync::Arc<std::sync::atomic::AtomicU64>,
+    #[cfg(test)]
+    test_id: u64,
+    #[cfg(test)]
+    thread: Option<std::thread::JoinHandle<()>>,
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in `475d43bf`. Test builds now join the capture worker from `Drop`; `stop_for_test` still takes and joins the same handle first, leaving `Drop` with `None`. Cleanup preserves an already-active test panic to avoid a double-panic abort. The targeted liveness test, test clippy, formatting, and all 5,005 default tests pass.

src/tmux/session.rs
```diff
@@ -1515,100 +1602,554 @@ impl Session {
...
+            (&observed_owner, &observed_heartbeat),
+            owner_id,
+            heartbeat,
+            deadline,
+        );
+        if matches!(replaced, Ok(true)) {
+            return true;
+        }
+
+        let verify_deadline = crate::tmux::TmuxCommandDeadline::new();
```
jerome-benoit: **[P2] Keep owner recovery inside the caller's deadline**

This method receives the capture cycle's shared deadline, but after the owner-pair replacement fails it starts fresh 10-second budgets for verification and cleanup. If the initial snapshot and replacement exhaust the original budget, one claim can therefore remain active for roughly 30 seconds while the render watchdog replaces the worker after 12 seconds, causing overlapping workers. The same budget renewal appears in `resize_window_if_owner` and in VT/OSC52 arm-failure cleanup.

Please reuse the supplied deadline across snapshot, replacement, verification, and cleanup. If cleanup needs a separate grace period, make it explicitly bounded and include it in the watchdog contract. Add a regression where the budget expires after the initial snapshot rather than before entry.

jerome-benoit (commented):
Re-review of `4b183f4f`. The earlier CodeRabbit findings are resolved and the CI checks are green. One regression remains in the shared-deadline invariant; details are inline.

The PR description also needs a small architecture update: passive resize now runs in the dedicated `aoe-passive-resize` worker and is woken when work is enqueued, rather than being executed by the snapshot poller on an approximately one-second cadence.

njbrake (commented):
**Trim the description before merge.** It becomes the squash commit body, and most of it is review-cycle logs. Keep the summary, reproduction, and risks.

**`tmux_format_literal` mis-escapes `{` and `:`** (`session.rs:1867`). A `#`-prefixed `{` opens a nested format; `#:` stays two characters. On tmux 3.6 an owner containing either never compares equal, so every guard built on it fails closed and silently. Unreachable with today's `live-N` / `tui-PID-N` ids, and untested. I have a patch plus a mutation-checked real-tmux test.

**Your `[P2]` on owner recovery is still open.** `vt.rs:1486` and `:1492` open fresh 10s budgets inside the cycle deadline, so a cycle can outrun the 12s stall detector and overlap workers.

**Can the ownership rework land first?** `session.rs` and `vt.rs` carry most of the risk and are only partly forced by moving resize off paint. Split, and each half gets a review proportional to its risk.

jerome-benoit (commented):
Re-reviewed and cross-validated at `8328236080ab9714f42384c9fab68f064f3d66fa`. The main shared-deadline owner recovery is fixed, the tmux format handling for literal `{` and `:` is corrected with real-tmux coverage, and the PR description is now appropriately scoped and concise. One residual P2 is inline.

GitHub reports the PR mergeable with 31 checks and no failures. I did not run local builds or tests.

src/tmux/session.rs
```diff
@@ -1515,100 +1602,543 @@ impl Session {
...
+                SIZE_OWNER_OPT,
+                SIZE_OWNER_HB_OPT,
+                owner_id,
+                &heartbeat.to_string(),
+            );
+            if self.resize_window_if_format_with_deadline(&condition, cols, rows, &deadline) {
+                return true;
+            }
+
+            let verify_deadline = crate::tmux::TmuxCommandDeadline::new();
```
jerome-benoit: **[P2] Keep live-send resize within one command deadline.** `resize_window_if_owner()` creates one deadline for the resize attempt, then fresh full deadlines here for verification and again for cleanup. A stalled resize can therefore hold the ordered live-send worker for roughly 30 seconds while subsequent input waits behind it. Please thread one deadline through the owner read, guarded resize, verification, and cleanup, and add a regression where the first command exhausts the budget mid-operation rather than before entry.

jerome-benoit: Addressed in 15fca6de85098e8b456857c77f80454c60a47bea. `resize_window_if_owner` now creates one deadline and threads it through owner reads, guarded resize, verification, cleanup, and the retry. The real-tmux regression starts with a live 200 ms budget, exhausts it inside the guarded resize via a blocking hook, proves the resize entered, and verifies recovery returns without a fresh budget.

PR #3538, opened 2026-08-26
feat(acp): runner terminates the ACP protocol and the relay socket is retired (#1054 Phase C)

### LINKED ISSUE #2977: ACP runner termination Phase C: server-method proxying + retire relay socket (#1054)

Seluj78:

## Description

Phase C of runner-side ACP protocol termination (#1054), following PR #2975 (Phase A) and PR #3056 (Phase B). The `aoe __acp-runner` shim becomes the sole ACP protocol terminator: it owns the handshake, the turn, and every JSON-RPC id on the agent's stdin, and the raw `<id>.sock` byte relay is retired. Only `<id>.control.sock` remains.

**Control protocol v3.** Two generic correlated lanes carry opaque JSON-RPC payloads. Reverse (agent to daemon): `ServerCall` for the nine agent-to-client requests, answered by `ServerResult` / `ServerError`, plus `Notify` for the event stream. Forward (daemon to agent): `AgentCall`, answered by `AgentResult` / `AgentError`, carries `session/set_mode`, `session/set_config_option`, `session/delete`, `_session/steering`, and conversation-reset `session/new`. Errors retain their structured envelopes. Cache-only `ResumeSession` waits for resets sent by the previous attachment to commit, then returns the authoritative session identity without issuing ACP load/new.

**One ordered channel fixes the watermark hazard for free.** Notifications and turn completion now share a single FIFO in agent-stdout order, so a `Stopped` can no longer overtake the chunks that preceded it. That was a real cross-socket race Phase B had to defer, and it closes here by construction rather than by adding machinery.

**The runner stops writing from its stdout task.** `emit_control` used to hold the control mutex across a bounded socket write while running on the sole agent-stdout task. That was survivable when it carried the occasional completion; carrying the entire event stream, a daemon that stopped reading (suspended laptop, overloaded UI) would stall the drain of the agent's stdout pipe and then the agent itself. It is now an append under an uncontended lock, with a dedicated writer task owning the socket. A failed write returns the frame to the front of the queue, which is what keeps ordering intact across a reattach.

**Disconnect contract: replay output, cancel attachment-scoped requests.** Buffered notifications and turn completion survive detach. Reverse calls are not replayed: pending permission requests are cancelled and other pending reverse requests receive a JSON-RPC error. Late forward replies belonging to an old attachment are discarded. A sent reset keeps runner-side bookkeeping until its new identity is committed, even if its reply can no longer be delivered. The outstanding-call cap refuses new calls instead of evicting unanswered ones.

**Prompt completion correlation.** Attachment-scoped `PromptStarted` announces the runner's canonical request ID before the agent receives a prompt. Only that ID can resolve the local waiter. Local activation supersedes adoption before terminal state is rearmed; older completions remain ignored even after local delivery. Protocol and runner generation remain 3.

**Daemon side is a transport shim, not a second driver.** The daemon presents the ACP crate with an in-process duplex and translates at the control-channel boundary. A `ConnectionTo<Agent>` remains on the runner path over this shim, reusing request handlers with the direct-stdio path rather than duplicating the connection driver.

**`RUNNER_VERSION` 3 and compatibility.** Liveness remains separate from build currency and protocol generation. A compatible runner from an older build can finish its current turn before replacement at idle. An incompatible generation cannot be adopted: its process tree is terminated and the instance returns to Terminal, including interruption of an active turn. Structured-row repair holds the session transition lock through persistence, skips active transitions, and revalidates the sampled worker identity so it cannot undo teardown.

Also fixed, found while working through the above:

- `delete()` on a `FreshSpawn` could strand a live runner and its whole agent tree. `terminate()` resolves the PID by re-reading the record, so delete-then-terminate silently no-ops. Now terminates first.
- Both supervisor cleanup paths unlinked only the relay path, which is no longer the socket a runner binds.
- `terminal/release` was not replay-safe; it now distinguishes a duplicate release from an id that never existed rather than masking both.
- Stop reasons `cancelled` / `max_tokens` / `refusal` / `max_turn_requests` reached the UI as `prompt_complete`. They are preserved now, resolving the in-code TODO that named this issue.

**Integration fixtures use the real runner.** Install shim dependencies with `npm ci` in `acp-worker/test-shim`; otherwise the Node-backed cases skip. The midturn and silent-orphan tests spawn `aoe __acp-runner` with that shim and establish the runner session before reattach. The reset regression attaches an `AcpClient` with a stale snapshot while reset is pending, then checks the published identity and the session addressed by its next prompt.

Relates to #1054
Closes #2977

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- [ ] Start a structured-view ACP session and send a prompt that makes the agent read files and run a shell command. Confirm tool calls, approvals, and terminal output all still work.
- [ ] Confirm `ls ~/.agent-of-empires*/acp-workers/` shows a `.control.sock` for the session and **no** `.sock`.
- [ ] Approve a permission prompt while the agent has other work in flight; confirm the approval does not stall the rest of the turn.
- [ ] Mid-turn, `aoe serve --stop` then `aoe serve`. Confirm the session picks the turn back up and clears "thinking" when the agent's response lands, with buffered output appearing before the turn ends rather than after.
- [ ] While detached mid-turn, confirm a pending approval is cancelled rather than re-presented twice on reattach, and the agent is not wedged.
- [ ] With a compatible runner on an older build, restart the daemon during a turn and confirm continuity followed by replacement at idle. With an incompatible runner generation, confirm explicit termination and a return to Terminal, not attempted adoption.
- [ ] Switch a session to the structured view and back; confirm `/clear` (conversation reset) still starts a fresh conversation.
- [ ] Run a direct-stdio agent (no runner) and confirm nothing about it changed.

## Checklist

## Test Coverage Analysis

**Runner / server integration coverage**
- [x] Real-runner control, midturn-resume, and silent-orphan cases under `tests/integration/`.
- [x] Reload repair regressions cover active teardown, retired registry samples, and repair persistence before a following transition.
- [x] Reset/reattach regression exercises the production runner and `AcpClient`.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Investigation, a three-model design debate, the plan, and the implementation were drafted by Claude under my direction. I made the scope and design calls (full bundled Phase C rather than a C1/C2 split, the cancel-not-replay disconnect contract, and the transport shim over a second driver), reviewed each commit, and ran the manual test steps.


docs/development/internals/structured-view.md
```diff
@@ -6,14 +6,21 @@ aoe is the ACP *client*; each agent (Claude Code, Gemini, `aoe-agent`, etc.) is
...
 
-Each runner registers at `<app_dir>/acp-workers/<session_id>.json` (PID, socket path, cached ACP session id, `build_version`); the same dir holds the per-session `.sock` and `.log` (runner stderr drain). `aoe ps --acp --dead` lists them.
+Each runner registers at `<app_dir>/acp-workers/<session_id>.json` (PID, socket path, cached ACP session id, `build_version`, `runner_version`); the same dir holds the per-session `.control.sock` and `.log` (runner stderr drain). `aoe ps --acp --dead` lists them.
+
+The runner is the ACP protocol terminator (#1054). It owns the handshake, the turn, and every JSON-RPC id on the agent's stdin; the daemon speaks a typed, length-framed control protocol to it over `<session_id>.control.sock` and never touches the agent's stdio. Agent-to-daemon calls (permission, elicitation, fs, terminal) are proxied over that channel as opaque JSON under a runner-assigned call id, and the agent's event stream rides the same socket. The raw `<session_id>.sock` byte relay was retired in #2977; `socket_path` in the record remains only as the derivation base for the control sibling's path.
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3538/commits/ce9eec152 — you're right, `expected_socket` still returns `rec.socket_path` for `runner_version < 2`, so the flat "was retired" reading was wrong. Reworded to scope it to `runner_version >= 2` and to say a v1 record keeps using its relay socket for attach and resume until the worker is replaced at an idle boundary.

docs/development/internals/structured-view.md
```diff
@@ -6,14 +6,21 @@ aoe is the ACP *client*; each agent (Claude Code, Gemini, `aoe-agent`, etc.) is
...
+
+Because notifications and turn completion now share one ordered channel, a turn's terminal event can no longer overtake the chunks that preceded it, which was possible while the two travelled over separate sockets.
+
+`runner_version` is the runner's protocol generation. It is deliberately not part of the liveness check: a live worker of the wrong generation is still live, and treating it as dead would orphan its in-flight turn. It behaves exactly like `build_version` staleness instead (see the upgrade section below), so a daemon that inherits an older runner mid-turn adopts it, lets the turn finish, then reaps and respawns it.
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3538/commits/acb495c7d, though in the opposite direction to the suggestion.

You found a real inconsistency, but the stale side is the code comment, not the docs. This PR deliberately changed what `runner_version` means: v2 changes the socket topology (a v2 runner binds only `<id>.control.sock`), and the field now gates protocol compatibility rather than record parsing. The record's *shape* did not change. What I missed was the cross-reference in `WorkerRecord::build_version`, which still called `runner_version` "the on-disk SCHEMA version" — that is what I corrected, so the two now agree that it is a protocol/topology generation tracked as a separate staleness axis from `build_version`.

src/acp/acp_client.rs
```diff
@@ -4286,11 +4411,99 @@ async fn connect_runner_control_v2(
...
+                match value.get("error") {
+                    Some(err) if !err.is_null() => ControlBody::ServerError {
+                        call_id,
+                        error: serde_json::from_value(err.clone()).unwrap_or_else(|_| {
+                            control_protocol::JsonRpcError::new(
+                                control_protocol::METHOD_NOT_FOUND,
+                                "handler produced a malformed error",
+                            )
+                        }),
+                    },
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3538/commits/ceaf158f9 — agreed, and the reason it matters beyond spec-correctness: an agent that special-cases -32601 would conclude the method is unsupported and stop calling it, when in fact a handler ran and only its error envelope was unparseable. Added `INTERNAL_ERROR = -32603` next to the other codes and used it at that fallback.

tests/acp_runner_control.rs
```diff
@@ -129,52 +138,102 @@ fn runner_reports_native_prompt_complete_over_control_socket() {
...
+    // the JSON value it arrived as rather than a rendering of it.
+    write_frame(
+        &mut ctl,
+        &serde_json::json!({
+            "kind": "agent_call",
+            "call_id": 1,
+            "method": "fs/read_text_file",
+            "params": {"path": "/tmp/x"},
+        }),
+    );
```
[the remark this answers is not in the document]
Seluj78: Fixed in https://github.com/agent-of-empires/agent-of-empires/pull/3538/commits/437fcb281 — both halves were leftovers from an earlier draft of this test, thanks for catching it. Rewrote the comment to describe the `fs/read_text_file` round trip it actually drives, and pointed at where the string-id round-trip really is covered (the `parse_request_value_id` table in `src/process/runner.rs`).

njbrake (commented):

_Reviewed at `bf31c2b0`. Push new commits and it will be re-reviewed._

---

Reviewed at `bf31c2b0`, rebased onto `origin/main` (`f226e216`). Ran `cargo fmt`, `cargo clippy --features serve --all-targets`, `cargo test --features serve`, `cargo test --features serve,e2e-tests --test e2e acp`, plus two scratch reproductions against a real `aoe __acp-runner`.

## Summary

Phase C: the runner becomes the sole ACP protocol terminator, `<id>.sock` is retired, control protocol goes to v3 with generic correlated reverse and forward lanes, and `RUNNER_VERSION` becomes a second staleness axis beside build currency. The engineering is careful and the reasoning in the comments is unusually good. Three things do not hold up, and one of them reproduces.

## Design alignment: **Aligned**

The transport shim keeps `ByteStreams` as the crate's transport exactly as `connect_via_stdio` (`acp_client.rs:2605`) and the old `connect_via_socket` did, swapping the unix socket for a `tokio::io::duplex`. That extends the existing mechanism rather than adding a second driver, and the nine `on_receive_request` handlers are untouched apart from a mechanical `Responder` to `Result` split behind `reply()`. `RespawnReason` / `is_runner_current` mirror `is_build_current` and the #1754 drain-then-respawn machinery instead of inventing a parallel one. `remove_runner_sockets` collapses three copies of the same two-line teardown into one helper. The `ConnectionTo<Agent>` deviation from #2977's wording is argued in the body and the call is right: forking a 2,878-line function for no behavioral difference is the worse trade.

## The upgrade-path answer first

**A v1 worker that is idle when the new daemon starts is handled correctly.** `is_record_live` (`worker_registry.rs:364`) now probes `expected_socket`, which returns `socket_path` for `runner_version < 2`, so a live v1 runner reads as live rather than dead; `is_runner_current` returns false; `adopt_decision(true, _, false, false)` gives `RespawnStaleIdle`, then `terminate()`, then a clean fresh spawn. Keeping the generation out of `is_record_live` is the right call and the comment explaining why is correct.

**A v1 worker that is mid-turn strands its whole process tree.** This is the common upgrade case the PR's own manual test step 6 describes, and it does not do what the body says.

1. `adopt_decision(true, false, false, true)` gives `AdoptStaleForDrain` (asserted by the new truth table, `acp_reconciler.rs:2231`).
2. The daemon dials `<id>.control.sock`. A v1 runner *has* one (Phase A bound it) and greets with `control_protocol_version: 2`. `connect_runner_control_v3` (`acp_client.rs:4216-4222`) requires `== 3`, returns `None`, and `connect_via_socket` turns that into `AcpError::Spawn("does not speak control protocol v3")`. There is no fallback; attach against a v1 runner **can never succeed**.
3. `resume_one` lands in `Ok(Err(e))` at `acp_reconciler.rs:1460` and calls **`worker_registry::delete(&id).ok()`** (`:1462`; the 3s-timeout branch at `:1470` is identical).
4. `delete` removes the record and unlinks the sockets but signals nothing. The v1 runner and its agent subtree keep running with no registry entry, so every reaper that resolves a PID by re-reading the record can no longer find them. `aoe ps --acp` will not list it either.

This is precisely the hazard this PR fixes 100 lines earlier: the `FreshSpawn` branch was changed from `delete` to `terminate` with the comment "deleting first would leave it with nothing to signal and strand the runner plus its whole agent tree with no daemon left to reap it" (`acp_reconciler.rs:1348-1358`). The same argument applies verbatim at `:1462` and `:1470`, and `supervisor.rs:3070-3076` makes the same argument again for the allowlist path ("Refusing to attach is not enough on its own... So terminate it"). The attach-failure `delete` is pre-existing, but this PR is what routes every mid-turn upgrade into it.

The draining that `AdoptStaleForDrain` exists to permit is impossible for runner-generation staleness, so the honest fix is to treat `!runner_current` as `RespawnStaleIdle` regardless of `in_flight_turn` (terminate kills the turn, but so does today's path, it just also leaks the process), or at minimum to make `:1462` and `:1470` call `terminate` rather than `delete`.

Two claims in the tree assert the opposite and should go either way:

- `acp_client.rs:2756`: "the reconciler's runner-version migration is what keeps a v1 worker on its legacy path until it drains". There is no legacy path left.
- `docs/development/internals/structured-view.md:13`: "A legacy `runner_version < 2` record still names a real relay socket there, and **keeps using it for attach and resume** until the worker is replaced at an idle boundary". It is used only for the liveness probe, never for attach. This sentence was added in response to a CodeRabbit comment and is more wrong than what it replaced.
- The same claim again in the `runner_control_absent_socket_leaves_guard_unclaimed` doc comment (`acp_client.rs:~16094`).

**Downgrade** behaves as it always has: an old daemon requires `runner_version == 1` inside `is_record_live`, sees a v2 worker as dead, and `delete`s it, stranding it. Not fixable from this side, but worth knowing that `aoe update` rollbacks will leak v2 runners.

## Blocking

### 1. A reverse-lane call issued while the runner is servicing a handshake frame deadlocks the runner permanently

`handle_control_connection` (`runner.rs:1705`) processes control frames on a single loop and `await`s the handler inline. For `Initialize` (`:1759`) and `EstablishSession` (`:1769`) that handler is `run_or_replay_initialize` / `run_or_replay_session`, which call `agent_request` then `rx.await` (`runner.rs:1231`) with **no timeout**. While that is parked, no `ServerResult` can be read. If the agent issues any client request while answering `session/new` or `session/load`, the agent waits on the runner, the runner waits on the agent, and neither moves.

The doc comment at `runner.rs:1701` states "The read loop never awaits a handler... Nothing here can park for the minutes a permission request takes." That is not true of the two handshake arms, and the merged lanes are what made it matter: on `main` the reverse traffic rode the relay's own accept loop (`origin/main:src/process/runner.rs:429`), a separate task, so this could not happen.

Reproduced against the built binary with a Python agent that issues `fs/read_text_file` while handling `session/new`:

```
hello: {'kind': 'hello', 'control_protocol_version': 3, 'session_id': 'dlrepro'}
frame: {'kind': 'initialized', 'result': {'protocolVersion': 1}}
frame: {'kind': 'server_call', 'call_id': 1, 'method': 'fs/read_text_file', ...}
answered server_call; now awaiting session_ready...
RESULT: DEADLOCK - no session_ready within 8s
```

And it is unrecoverable, because the accept loop `await`s `handle_control_connection` inline (`runner.rs:389`), so no later daemon can attach:

```
SECOND DAEMON: no Hello within 5s (accept loop wedged)
```

The runner survives with a live agent until the 48h retention watchdog or an explicit `aoe acp kill`. `aoe-agent` does not trip this (its `fs/*` calls only come from tools during `session/prompt`, and `Prompt` does not await), so no shipped adapter demonstrates it, but an agent using the client's fs capability during session setup is legal ACP, and third-party adapter support is the whole point of this layer. Both handshake arms already deliver their answer via `emit_control`, so nothing needs to return to the loop; moving them off it (with the `run_or_replay_*` cache guarded so two concurrent `Initialize` frames cannot double-send) looks like a small change.

### 2. `silent_orphan_fires_on_cost_then_silence` becomes flaky

`tests/integration/acp_silent_orphan.rs:88`. On macOS, `cargo test --features serve`, shim deps installed via `npm ci`:

| branch | result |
| --- | --- |
| `origin/main` (`f226e216`) | **8 pass / 0 fail** |
| this PR | **5 pass / 3 fail**, plus one more failure during the full-suite run |

Failure is always `left: None, right: Some("prompt_orphaned")`: the watchdog never synthesizes the stop. The test was already tuned once for CI headroom (15s drain budget), and it does not exhaust that budget when it fails, so this reads as a behavioral difference rather than pure scheduling noise. The silent-orphan escalation now routes `session/cancel` through `control.cancel()` (`acp_client.rs:7939`) instead of a crate notification, and the fixture now has a real runner process in the loop. CI passed once on this PR, so it will land as an intermittent red rather than a hard one. Worth diagnosing before merge rather than after.

### 3. `server_call_results` is never read, so the advertised results-replay contract does not exist

`runner.rs:663` declares it; `:1019` inserts; `:1029` removes on successful delivery. Nothing anywhere reads it. `forward_server_call` does not consult it, and could not usefully: it allocates a fresh `call_id` per agent request, so a "replayed" call never shares a key.

That makes four statements untrue:

- PR body: "The runner replays buffered notifications and completed-but-undelivered results (memoized by call id, **which is why a daemon restart during a `terminal/create` does not run the command twice**)."
- `structured-view.md`: "Results the runner computed but could not deliver ARE replayed, keyed by call id, which is why a daemon restart during a `terminal/create` does not run the command twice."
- `runner.rs:656-663` and `:1050`, same claim.
- `acp_client.rs:9867-9871` justifies the new `terminal/release` idempotency by "#2977 replays a completed-but-undelivered result across a control reattach."

The forward lane's `AgentResult` frames *do* replay via the outbound queue, which is presumably where the idea came from, but the reverse lane's memo is inert. The `terminal/release` `was_released` hardening is harmless and arguably good on its own, but its stated rationale is fictional. Under the actual contract, `cancel_server_calls` answers a live `terminal/create` with a `DAEMON_GONE` error and the agent may simply retry, which is a *different* double-execution story that nothing here addresses. Either implement the memo lookup or delete the field and correct all four statements. AGENTS.md's no-dead-code rule applies, and a `pub`-adjacent struct field will not be flagged by the compiler.

## Non-blocking

**`METHOD_NOT_FOUND` is dead.** `control_protocol.rs:214` is referenced only by its own round-trip test. Its doc comment claims it "Answers a `ServerCall` for a method this daemon has no handler for," but the shim never constructs it; the crate's own dispatch answers unknown methods. Delete it or wire it.

**`cancel_server_calls` has a drain/purge race.** `runner.rs:1060` drains `pending_server_calls`, then `:1078` purges every queued `ServerCall` frame. The stdout fanout runs on a different task and can insert a new pending entry plus its frame between those two steps; the retain then deletes that frame while its pending entry survives, so the agent parks on a request no daemon will ever see, until the *next* control disconnect. Narrow window, but the fix is to purge only the ids actually drained rather than all `ServerCall` frames.

**Forward `call_id` space resets per daemon connection.** `ShimCorrelation::next_forward` starts at 0 for each `connect_runner_control_v3`, while the runner's `pending_agent_calls` persists across attaches. A stale `AgentResult { call_id: 1 }` queued from a previous daemon could resolve the new daemon's first forward call. Socket ordering makes this safe today (the backlog flushes before the new daemon's handshake completes, let alone its first `set_mode`), but that is accidental. The reverse lane's `next_call_id` is monotonic per runner; the forward lane should be too, or the runner should tag results with a connection generation.

**Hello's `session_id` is still ignored** (`acp_client.rs:4218`). Pre-existing, but the stakes rose: that channel now carries the entire event stream and every reverse call, not the occasional completion. One `==` against the expected session id is cheap insurance.

**The dial retry loop swallows every connect error** (`acp_client.rs:4207-4213`). Previously `wait_for_socket` returned a typed `connect {path}: {e}` for anything other than `ConnectionRefused`. Now `ENAMETOOLONG`, `EACCES`, or a path that is not a socket all spin for the full 10s deadline and then report "does not speak control protocol v3," which is misleading. It also makes `runner_control_absent_socket_leaves_guard_unclaimed` burn 10s where it used to return instantly.

**`TerminalManagerInner.released` grows for the session's lifetime** (`terminal_handler.rs:73`). "Bounded by the agent's tool calls" is a real bound but not a small one for a long-lived session. Minor.

**Stale comments the rewrite left behind**, all in `runner.rs`: `:764` points at `cancel_outstanding_requests`, renamed to `cancel_server_calls`; `:809` documents a "soft cap on `outstanding_requests`" and a shed-the-non-permission-entries policy, but the field is gone and the code now refuses rather than sheds; `:783` says `emit_control` "holds the `control` mutex across the write," which this PR is specifically about no longer doing; `:777` says the crate "still sends over the relay on the v2 path"; `:1376` references the removed `relay_session_news`.

**Environment note, not the PR's:** `cargo clippy --all-targets` fails locally on rust 1.96 with `items_after_test_module` in `src/process/macos.rs:266` and `useless_vec` in `src/session/smart_rename.rs:1546`. Neither file is in this diff and CI's Clippy job is green, so this is local toolchain skew.

## Test coverage

`tests/integration/*` **did not lose assertions.** The -101/-49 is entirely fixture code: three hand-rolled byte-proxy spawners deleted, all ten tests kept with their assertions intact, now pointed at `spawn_runner_with_shim`. That helper in `common/mod.rs` is used by both files, so it is genuinely shared, and driving the real runner instead of mocking the peer under test is better coverage. Aside from finding 2, this is a straight improvement.

The new runner unit tests are table-driven and meaningful (`control_disconnect_cancels_reverse_calls_rather_than_replaying` walks permission versus non-permission methods and asserts both the echoed id and the outcome shape). `adopt_decision_truth_table` correctly consolidates four `#[test]` fns into one table with more rows than before. `tests/acp_runner_control.rs` asserting `!socket.exists()` is exactly the right way to pin the retirement.

What is not covered: the upgrade path itself. `adopt_decision` is tested in isolation, but nothing exercises "v1 record plus live PID plus new daemon" end to end, which is why the consequence at `acp_reconciler.rs:1462` slipped through. Given #1054's whole premise is surviving daemon restarts, one integration test that plants a `runner_version: 1` record over a live process and asserts the process is signalled rather than orphaned would be worth its cost.

## Test results

| Suite | Result |
| --- | --- |
| `cargo test --features serve --lib` | 6416 passed, 0 failed |
| `cargo test --features serve --test acp_runner_control` | 4 passed |
| `cargo test --features serve --test integration` | 167 passed, **1 failed** (finding 2) |
| `cargo test --features serve,e2e-tests --test e2e acp` | 9 passed, including `acp_focus_isolation_e2e::tui_acp_modal_approval_with_live_daemon` |
| GitHub CI on `bf31c2b0` | all green |

The live-daemon ACP e2e suite passes, so the happy path through the new lanes is genuinely exercised end to end.

## Automated review check-in

CodeRabbit's four line-level findings were all addressed by `ce9eec1..bf31c2b` and the resolutions look right, except that its `structured-view.md:13` suggestion introduced the false "keeps using it for attach and resume" sentence; that one is worse than what it replaced. Its walkthrough "Merge Risk: High" cites three things: "mismatched sessions may receive traffic" (real but pre-existing, see the Hello note), "resumed turns may complete incorrectly" (could not be substantiated; the ordered-FIFO argument holds and `read_typed_frame` in the tests demonstrates it), and "stopping a worker can leave its process running when registry data is unavailable" (the concrete instance is real, but at `acp_reconciler.rs:1462`, not where it looked). Its "Linked Issues check" objection about cancel-not-replay and the retained `ConnectionTo<Agent>` is answered by the PR body; those are argued deviations, not gaps.

## Documentation

Proportionate for a protocol version bump: 11 lines in the one file that describes this subsystem, no new standalone markdown, no narration of the change. No other doc mentions the relay socket, so nothing else went stale. Two corrections needed, both above: the "keeps using it for attach and resume" sentence and the results-replay sentence.

## Merge risk

**High.** Blast radius is every structured-view session on every daemon restart, and the two worst outcomes are both silent: a mid-turn upgrade that leaves an `aoe __acp-runner` plus its agent tree running with no record and no reaper (the user sees a working new session and pays for a leaked one, discovering it via `ps`, not via aoe), and a wedged runner that no reattach can recover. `git revert` is clean at the code level, but a runner already orphaned by finding 1 outlives the revert, and a `RUNNER_VERSION` bump means a revert re-crosses the generation boundary in the other direction. Nothing in the PR would surface either: no test covers the v1-record path, and the deadlock produces silence rather than an error. The flake in finding 2 is the one thing that *will* be noticed, and it will be noticed as CI noise rather than as a signal.

No fixes were staged. Findings 1 and 3 need the author's design judgment (where to move the handshake off the read loop; whether to implement the memo or drop it), and finding 2 needs diagnosis rather than a patch.

## Verdict

**Design alignment:** Aligned
**Project fit:** Accept
**Documentation:** Missing updates (two corrections)
**Merge risk:** High; a mid-turn upgrade orphans the old runner's process tree, silently

This is strong, well-staged work and the design is the right one; the shim, the single ordered channel, and the runner-generation-as-staleness-axis decisions are all correct and well argued. It should land, but not at `bf31c2b0`: the mid-turn upgrade path leaks a process tree on the exact scenario the PR body says it handles, a reverse call during the handshake wedges the runner beyond recovery (reproduced), the replay-memo the disconnect contract is built on is not wired up, and the test suite goes from 8/8 to 5/8 on a test this PR rewrote.

njbrake (commented):

_Reviewed at `41fdb54d`. Push new commits and it will be re-reviewed._

---

Rebased onto `origin/main` (`522a7d2e`, which now includes #3551 and #3540). Ran `cargo fmt --check`, `cargo clippy --features serve -- -D warnings`, `cargo clippy --features serve --all-targets`, `cargo test --features serve`, `cargo test --features serve,e2e-tests --test e2e`, `scripts/check-nix-embedded-assets.py`, three mutation tests against the new regression guards, and 32 repeat runs of the ACP integration suite plus a 12-run baseline on `origin/main`.

## Headline

**All three blocking findings from `bf31c2b0` are genuinely fixed, and I verified each by reverting the fix and watching the new test go red.** The mid-turn upgrade no longer orphans a process tree, the handshake deadlock is closed, and the phantom results memo is gone rather than papered over. The four follow-up commits are careful, well-argued, and each names the review comment it answers.

What is left is smaller but real: one narrow orphan path this PR opens and has not closed, a doc that now contradicts itself four lines apart, a PR body that three of the five new commits invalidated, and a test this PR rewrote that fails roughly one run in four on macOS.

## The CodeQL failure: false positive, do not gate on it

**Merge should not be gated on this, and the alert is dismissible with high confidence.** Four independent pieces of evidence:

**1. Both sinks are `panic!` in test code, not a log file.** CodeQL's Rust cleartext-logging query treats `panic!` / `println!` as logging sinks.

- `src/process/runner.rs:2235` sits inside `#[cfg(test)] mod tests` (the module opens at `:2066`). The sink is `panic!("{method}: got {actual:?}, expected {expected:?}")`, the failure arm of the `established_session_id` truth table. The tainted values are the table's own hardcoded literals (`"existing-id"`, `42`).
- `tests/acp_runner_control.rs:74` is `panic!("{what} never appeared at {}", path.display())` inside the `wait_for` fixture helper. The taint is a test session id (`"sdl00001"`, `"midturn-true"`) reaching the path it prints.

**2. Both flagged lines are byte-identical and unmodified on `main`.** `wait_for` is at `main:70` (this PR shifted it to 74 by adding a doc comment above `KillOnDrop`), and the `established_session_id` match arm is at `main:2148` (shifted to 2235). This PR did not touch either hunk; it added new *callers* of `wait_for`, which is what gave the dataflow query a new taint path to report. The check-run summary says as much itself: "Alerts not introduced by this pull request might have been detected because the code changes were too large."

**3. An ACP session id is not a credential.** It is a conversation handle minted by the agent's `session/new` response (`established_session_id`), meaningful only to the one local agent subprocess the runner already owns stdin on. It authorizes nothing across a trust boundary: reaching that agent requires being the runner, and reaching the runner requires connecting to a 0600 unix socket owned by the same uid. The aoe session id is likewise a filesystem-visible name (`acp-workers/<id>.json`, `<id>.control.sock`).

**4. `debug.log` is 0600 and already carries session ids everywhere.** `src/logging.rs:733` sets `0o600` on the log file on creation, and `origin/main` has roughly 300 `session = %session_id` tracing fields across `src/acp/`, `src/server/`, `src/process/`. If logging a session id were a finding, it would be a repo-wide policy question, not this PR's.

**5. CodeQL is not a required check.** The ruleset on `main` requires: Playwright (live), Playwright (mocked), Format, Clippy, Supply Chain, Docs, Skill Check, Nix Eval, Website Build, Vitest, Cargo Test (linux), Cargo Test (macos). All twelve are green. `mergeable_state: unstable` rather than `blocked` reflects that.

Recommendation: dismiss both alerts as "used in tests" in the GitHub UI. The `tests/` one is the clearer of the two; the `src/process/runner.rs` one is equally clear once you see it is inside `mod tests`.

## Prior findings, one by one

### 1. Mid-turn upgrade orphans a process tree: **addressed correctly** (`782df9e9`)

The fix is the honest one I hoped for rather than the minimal one. `adopt_decision` now sends any generation-stale worker to `RespawnStaleIdle` regardless of `in_flight_turn`, with a comment that states the reason plainly: draining requires attaching, the control-version gate refuses that runner, so adopting only routed the session into the attach-failure path. Both attach-failure branches (`acp_reconciler.rs:1480` and `:1488`) now call `terminate` rather than `delete`, with the same argument the `FreshSpawn` branch already made 100 lines earlier. All three false "legacy path" claims are corrected.

The new coverage is the right coverage: `live_v1_record_is_live_but_stale_and_terminate_reaps_it` plants a `runner_version: 1` record over a genuinely live process in its own process group and asserts all three links in the chain (reads live, reads generation-stale, is actually signalled). Mutation-verified: folding the generation check back into `is_record_live` fails it at `worker_registry.rs:983`. The truth-table change is mutation-verified too:

```
adopt_decision_truth_table ... FAILED
  live=true build_current=true runner_current=false in_flight=true
  left: AdoptStaleForDrain
 right: RespawnStaleIdle
```

The composition itself (attach failure reaching `terminate`) has no test, but with `RespawnStaleIdle` firing first that path is now belt-and-braces rather than the main line, so I would not hold the PR for it.

### 2. Reverse call during the handshake wedges the runner: **addressed correctly** (`3da94863`)

Both handshake arms are spawned off the read loop, `handshake_gate` (a separate lock from `handshake`, for the reason the comment gives) serializes them so two frames cannot both miss the cache and double-send, and the doc comment at `runner.rs:1706` is now specific about which frames await the agent instead of claiming none do. `agent_request_during_session_new_does_not_deadlock_the_runner` drives a real runner with a python agent that requests a file mid-`session/new`. Mutation-verified by restoring the inline `await` on both arms:

```
agent_request_during_session_new_does_not_deadlock_the_runner ... FAILED
  panicked at tests/acp_runner_control.rs:84:
  read frame length: Os { code: 35, kind: WouldBlock }
```

### 3. `server_call_results` never read: **addressed correctly** (`0c1a03bb`)

Deleted rather than implemented, which is the right call and the commit message explains why implementing it could not have worked (a `call_id` is minted per agent request, so a retry never shares a key). All four statements that leaned on it are corrected, and `structured-view.md` now says plainly that an agent retrying a `terminal/create` after a disconnect runs the command again and names that as the Phase D gap. The `was_released` hardening is kept with an honest justification.

### 4. Non-blocking items: **all addressed** (`2a6ad8e9`)

`METHOD_NOT_FOUND` removed, the `cancel_server_calls` drain/purge race closed by purging only the ids actually drained, the forward `call_id` space made process-monotonic via `NEXT_FORWARD_CALL_ID`, `Hello`'s `session_id` checked against `session_label`, the dial retry loop narrowed to `NotFound | ConnectionRefused` with everything else surfaced, and the five stale comments corrected. The absent-socket test now shrinks its deadline through `AOE_ACP_RUNNER_SOCKET_TIMEOUT_MS` with a `RestoreEnvOnDrop` guard under `#[serial]` rather than burning 10s.

The session-id check paying for itself immediately is a nice signal: `41fdb54d` exists because turning it on caught `spawn_runner_with_shim` hardcoding `"midturn1"` while each test attached under a different id, which had been quietly misrepresenting the fixture in all ten attach tests.

### 5. `silent_orphan_fires_on_cost_then_silence` flake: **not addressed**, and I have sharper data now

| tree | fixture | result |
| --- | --- | --- |
| `origin/main` (`522a7d2e`) | byte proxy | **0 failures / 12 runs** |
| this PR, whole `acp_` set | real runner | **5 failures / 20 runs** |
| this PR, that test alone | real runner | **2 failures / 12 runs** |

Always the same assertion: `left: None, right: Some("prompt_orphaned")`, meaning the watchdog never synthesized a stop inside the 15s budget. Running it in isolation reproduces it, so this is not contention with the other `#[serial]` tests.

The important new detail, and it cuts in the PR's favour on one axis and against it on another: **on `main` this test never exercised the runner path at all.** The old `spawn_shim_socket_bridge_with_preseed` gave `AcpClient::attach` a raw byte proxy, so `control_client` was `None` and `send_session_cancel!` took the `connection.send_notification` branch. With a real runner in the fixture it takes `control.cancel()` and the cancel travels as a control frame. So the 0/12 on `main` is not a fair baseline: the rewritten test covers strictly more, and covers what production actually does.

That is why I would not call this a regression, and also why I would not wave it through. The failure is not "slightly slow" (the test does not exhaust its 15s budget), it is a stop that never arrives. Because the newly covered path is the one every real structured-view session takes, the flake may be a genuine defect in the runner cancel path rather than test timing. CI on linux has passed at both `bf31c2b0` and `41fdb54d`, so this may be macOS-specific, but "we do not know why" is the state, and this test guards a user-visible wedge.

## Still open

### A. `pid_source_for` still probes the retired relay socket

`worker_registry.rs:489`:

```rust
let sock = socket_path_for(session_id).ok()?;
let pid = crate::process::worker::peer_pid_from_socket(&sock);
```

That is `<id>.sock`, which a v2 runner never binds (`tests/acp_runner_control.rs` asserts `!socket.exists()` precisely to pin that). So when the registry JSON is unreadable, the #2102 SO_PEERCRED backstop returns `None` for every v2 worker and `terminate` signals nothing: the runner and its agent subtree survive with no daemon able to find them. This is the same orphan class the PR spent four commits closing, reached through a different door.

Narrow trigger (corrupt or unreadable record), one-line fix, and the same helper `expected_socket` already encodes the answer. CodeRabbit flagged this at round 2 with the right patch shape:

```rust
let pid = crate::process::worker::peer_pid_from_socket(
    &crate::process::worker::control_socket_sibling(&sock),
)
.or_else(|| crate::process::worker::peer_pid_from_socket(&sock));
```

I would take that.

### B. `structured-view.md:17` contradicts `:13`

Line 13, corrected in `782df9e9`: "the reconciler replaces such a worker rather than attaching to it."

Line 17, not updated: "It behaves exactly like `build_version` staleness instead, so a daemon that inherits an older runner mid-turn adopts it, lets the turn finish, then reaps and respawns it."

The code does the former. Line 17 is the last statement of the old design left standing, and it is the paragraph a reader goes to for what `runner_version` means. It needs the same treatment the `RUNNER_VERSION` doc comment and the `adopt_decision_truth_table` comment already got: the two axes are not interchangeable, a build-stale worker drains and a generation-stale one does not.

### C. The PR body no longer describes what is pushed

Three claims that the last four commits invalidated. At +2492/-1354 across an architectural boundary this matters more than usual, because the body is what a reviewer calibrates against.

1. **"The runner replays buffered notifications and completed-but-undelivered results (memoized by call id, which is why a daemon restart during a `terminal/create` does not run the command twice)."** `0c1a03bb` deleted that memo and corrected every in-tree statement of it; the body is now the only place the claim survives. Under the shipped contract the command **is** run twice, and `structured-view.md` says so.
2. **"Runner generation becomes a second staleness axis beside build currency, both adopted for the duration of a turn and replaced at the next idle boundary."** `782df9e9` made generation staleness replace immediately, in flight or not.
3. **Manual test step 6:** "With a session running on the old build, `aoe update` and restart the daemon. Confirm the in-flight turn keeps streaming and the worker is replaced once it goes idle." For a v1 to v2 upgrade, which is what shipping this means for every existing user, the in-flight turn does not keep streaming. It is terminated and the session gets a synthetic stop. That is the correct behavior (the alternative was the leak), but the checklist as written will read as a failure to whoever runs it.

### D. `tests/acp_runner_control.rs` module doc describes the socket this PR retires

Lines 1-15 still say the test drives "the main relay socket" and that "before this change the control socket did not exist." The file's own test at `:144` now asserts the relay socket must **not** exist, and the control socket has existed since Phase A. Pre-existing text, but this PR is what made it false. CodeRabbit raised it at round 1 and it is still open.

### E. `respawn_pending` is not cleared on the attach-failure path

`resume_one` calls `mark_respawn_pending` before attaching (`acp_reconciler.rs:1412`). If the attach then fails, the marker survives into the fresh spawn, and the next `respawn_drained_stale_workers` tick can `mark_restart_pending` plus `terminate` the brand new replacement. Self-healing (the resume pass spawns again) but it costs the user a turn.

Pre-existing: `main` has the same shape with `mark_build_respawn_pending` and `delete`. This PR narrows it rather than widening it, because the case where attach was *guaranteed* to fail (generation-stale) no longer reaches `AdoptStaleForDrain` at all. Worth a follow-up issue rather than a change here, but CodeRabbit is right that it is real.

### F. One in-flight frame can be lost at disconnect

`handle_control_connection` calls `writer.abort()` when the read loop ends. If the writer has popped a frame from the queue and is inside `write_control_frame`, that frame is dropped rather than requeued. The queue itself is retained correctly (`clear_control_outbound` only clears `ready`), and the blast radius is small: a lost `ServerCall` frame is immediately followed by `cancel_server_calls` answering the agent, and a lost `PromptCompleted` is covered by the resume-idle watchdog. On a full socket close the writer would usually fail and requeue anyway. Noting it as residual, not asking for a change.

### G. Rollback still leaks v2 runners

An older daemon's `is_record_live` requires `runner_version == 1`, sees a v2 worker as dead, and `delete`s it, stranding the process tree. Not fixable from this side, and the forward path is now correct, but it means an `aoe update` rollback across this boundary leaks. Worth knowing before recommending a downgrade.

## CodeRabbit reconciliation

Seven review objects from `coderabbitai[bot]`, but only three carry a walkthrough (`99754772`, `bf31c2b0`, `41fdb54d`); the other four are empty-bodied replies to @Seluj78's responses on inline threads. So this is three substantive rounds plus four acknowledgements, not six independent passes. The signal-to-noise is decent, with one clear miss.

| Round | Finding | My position |
| --- | --- | --- |
| 1 | `remove_runner_sockets` should live in `worker_registry` | **Agree, addressed.** It is now `worker_registry.rs:393`, used by `delete` and both supervisor sites. |
| 1 | `INTERNAL_ERROR` constant instead of borrowing `METHOD_NOT_FOUND` | **Agree, addressed**, and `2a6ad8e9` went further and deleted `METHOD_NOT_FOUND`. |
| 1 | `acp_runner_control.rs:162-175` comment describes the wrong call | **Agree, addressed.** |
| 1 | `structured-view.md:17` should call `runner_version` an on-disk schema version | **Disagree.** This PR deliberately redefines it as the protocol generation, and the `RUNNER_VERSION` doc comment argues why. Following this would have documented the old meaning. |
| 1 (nit) | `acp_runner_control.rs` module doc still describes the relay socket | **Agree, still open.** Finding D above. |
| 2 | `Hello`'s `session_id` ignored | **Agree, addressed** in `2a6ad8e9`, and it caught a real fixture bug on the way. |
| 2 | `pid_source_for` should probe the control socket | **Agree, still open, and the most valuable thing CodeRabbit found on this PR.** Finding A above. |
| 2 | Prompt completion correlation across resumed turns | **Agree in principle, pre-existing.** `main:4222` has the same `PromptCompleted { outcome, .. }` destructure. Reachable only if the resume-idle watchdog synthesizes a stop while the old prompt is still live and the user then prompts again; the consequence is an early turn end plus a spurious waiterless stop, not corruption. Follow-up, not this PR. |
| 3 | `touch_live_socket` needs `#[cfg(test)]` | **Disagree, false positive.** `#[cfg(test)]` is at `worker_registry.rs:405` at the exact head CodeRabbit reviewed, directly above the function it quoted. |
| 3 | `structured-view.md:13` should restate that mid-turn workers are adopted | **Right instinct, wrong target.** The doc genuinely is inconsistent, but it is line 17 that is stale and line 13 that is correct. Applying this suggestion would document the bug `782df9e9` fixed. |
| 3 | `respawn_pending` not cleared on attach failure | **Agree, pre-existing, narrowed by this PR.** Finding E above. |
| 3 | RAII guard around the `victim` child in the new registry test | **Agree, low value.** If the `is_record_live` assert panics first, a `sleep 60` leaks for at most 60s. Fair nit, not worth a round trip. |
| 3 | `find_python3` should consult `PATH` | **Agree, minor, pre-existing.** The helper hardcodes three absolute paths, so on a pyenv/asdf/nix machine the deadlock regression test silently skips. It guards the highest-severity bug this PR fixes, so a silent skip is worth avoiding, but CI's linux runner has `/usr/bin/python3`. |

The "six rounds" pattern here is mostly CodeRabbit re-raising the same doc sentence from three angles while getting its direction backwards each time, plus one genuinely useful catch (`pid_source_for`) that has now been open across two rounds. Worth taking that one and letting the doc thread go, since the fix for it is the opposite of what was suggested.

## Design alignment: **Aligned**

Unchanged from the last pass, and the new commits stay inside the grain. `adopt_decision`'s new arm extends the existing #1754 staleness machinery rather than adding a second policy beside it, and it keeps the pure-function-plus-truth-table shape the file already had. `remove_runner_sockets` is now the single answer to "which sockets does a runner of generation N own," which is what CodeRabbit asked for and what `expected_socket` pairs with. `RespawnReason` earns its place: it is read in `respawn_drained_stale_workers`'s log line, and the doc comment explains why a bare set was insufficient once two causes could fire together. The transport shim keeps `ByteStreams` as the crate's transport exactly as `connect_via_stdio` does, so the nine `on_receive_request` handlers and the direct-stdio path provably cannot regress.

## Test results

| Suite | Result |
| --- | --- |
| `cargo clippy --features serve --all-targets` | 5 warnings, all the known `main` set (`background_agent.rs`, `supervisor.rs:4054`, `macos.rs`, `smart_rename.rs`); none in this diff |
| `cargo test --features serve` | **all green**, 6450 lib + 168 integration + the rest |
| `cargo test --features serve,e2e-tests --test e2e` | **195 passed**, 0 failed |
| `scripts/check-nix-embedded-assets.py` | OK, 42 assets |
| ACP integration, 20 repeats | 5 failures, all `silent_orphan_fires_on_cost_then_silence` |
| Same test alone, 12 repeats | 2 failures |
| `origin/main`, 12 repeats | 0 failures (different code path, see finding 5) |
| GitHub CI on `41fdb54d` | all 12 required checks green; CodeQL red (false positive) |

Note that the full-suite green above ran without `acp-worker/test-shim/node_modules` present for part of it. With `npm ci` run there, the ten attach tests execute for real, which is how the flake surfaces. That gap between "green locally" and "green with the shim installed" is worth a line in the file's module doc.

**Mutation testing.** Three assertions carrying the most weight, each verified red when its fix is reverted: `adopt_decision_truth_table` (the generation-stale-in-flight row), `live_v1_record_is_live_but_stale_and_terminate_reaps_it` (folding the generation check into `is_record_live`), and `agent_request_during_session_new_does_not_deadlock_the_runner` (restoring the inline handshake await). Output for each is quoted above. These are not tests that cannot fail.

No commit-level house-rule problems: conventional prefixes throughout, no em dashes or double-hyphen separators in added prose, no `#[allow(dead_code)]`, no OS-specific logic outside `src/process/{macos,linux}.rs`, tables rather than one test per case.

## Migration

**No `src/migrations/` entry is needed, and none should be added.** `WorkerRecord` gains no fields; `runner_version` already existed on `main` with the same serde shape, and `RUNNER_VERSION` is a value bump. The breaking change is not on disk, it is a live process from the previous generation, which a startup file migration cannot touch. The precedent is `is_build_current` and the #1754 drain-then-respawn machinery, and this extends it. Stale `<id>.sock` files left by a v1 runner are swept by `terminate` to `delete` to `remove_runner_sockets`, which removes both paths, so nothing is left behind on disk either.

## Documentation

Proportionate: 15 lines in the one file that describes this subsystem, no new standalone markdown, no narration of the change. Two corrections needed, both above: `structured-view.md:17` (finding B) and the `tests/acp_runner_control.rs` module doc (finding D). Everything else that went stale in the last four commits was corrected in the same commit that caused it, which is the right discipline.

## Merge risk

**Moderate**, down from High at `bf31c2b0`. The two silent, unrecoverable outcomes are gone and both are pinned by tests I confirmed can fail.

What remains: `pid_source_for` is a real orphan path, but it needs an unreadable registry record to reach, and it fails the way the old code did rather than a new way. The flaky test is the thing most likely to be noticed, and it will be noticed as CI noise rather than as the signal it might be. Reversibility is the residual worth naming: `git revert` is clean at the code level, but the `RUNNER_VERSION` bump means a revert re-crosses the generation boundary in the direction the old daemon handles badly (finding G), so a rollback after this ships will leak v2 runners. Nothing in the PR would surface that; a user would find it with `ps`.

Coverage gap worth naming specifically: nothing exercises `resume_one`'s attach-failure to `terminate` composition, and nothing exercises `pid_source_for` against a v2 record.

No fixes staged. Findings A, B, C, and D are all mechanical and better done by the author in one commit; finding 5 needs diagnosis rather than a patch.

## Verdict

**Design alignment:** Aligned
**Project fit:** Accept
**Documentation:** Missing updates (two corrections)
**CI workflow approval:** Not applicable (not a fork)
**Merge risk:** Moderate; a rollback across the generation boundary leaks v2 runners, and the flaky test may be masking a real defect in the newly covered cancel path

This turned around three hard blocking findings in four commits, each fix is the correct one rather than the minimal one, and I confirmed all three by mutation rather than by reading. The design was right at `bf31c2b0` and it is still right. The remaining list is short and mostly mechanical: take CodeRabbit's `pid_source_for` one-liner, fix the doc sentence at `structured-view.md:17` that now contradicts the one four lines above it, refresh the three stale claims in the PR body, and refresh the `acp_runner_control.rs` module doc. The one item that needs thought rather than typing is the `silent_orphan` flake, which fails about one run in four locally and is doing so on a code path the old fixture never reached. I would want an explanation for that before this lands, even if the explanation turns out to be scheduling.

src/process/runner.rs
```diff
@@ -902,133 +870,238 @@ impl RunnerShared {
...
+            pending.insert(
+                call_id,
+                PendingServerCall {
+                    agent_id,
+                    method: method.clone(),
+                },
+            );
+            call_id
+        };
+        self.emit_control(ControlBody::ServerCall {
```
jerome-benoit: [P1] Reject reverse calls created while detached

The disconnect sweep only drains calls that are present when `handle_control_connection` exits. The stdout fanout keeps running afterward, and `forward_server_call` unconditionally inserts and queues a new `ServerCall` even while `control.ready` is false. `emit_control` then retains that frame until the next `mark_control_ready`, so a permission or elicitation issued after disconnect is replayed on reattach, exactly the stale-card and parked-agent case the contract in `cancel_server_calls` says must be cancelled. Coordinate attachment state with insertion and the sweep, answer calls immediately while detached, and add a regression test that emits a reverse call after disconnect.

src/process/runner.rs
```diff
@@ -1613,127 +1638,105 @@ fn parse_response(line: &[u8]) -> Option<(i64, PromptOutcome)> {
...
         return;
     }
+    // Writer owns the write half from here; everything outbound goes through
+    // the queue so ordering is one FIFO.
+    let writer = tokio::spawn(run_control_writer(
+        write_half,
+        Arc::clone(&shared),
+        session_id.clone(),
+    ));
+    shared.mark_control_ready().await;
```
jerome-benoit: [P2] Validate `Attach` before opening the outbound queue

`mark_control_ready` runs before the first frame is read. This lets the writer pop and successfully write queued frames before the peer proves v3 with `Attach`; non-`Attach` frames are also processed. A mismatched older daemon closes after reading `Hello` without sending `Attach`, so writes accepted by the socket can remove backlog that the peer never accepted. Require a matching `Attach` as the first frame, with a deadline, before starting or readying the writer and accepting commands.

src/server/acp_reconciler.rs
```diff
@@ -1354,11 +1394,22 @@ async fn resume_one(state: Arc<AppState>, target: ResumeTarget) -> ResumeOutcome
...
+                    new_runner_version = crate::process::worker_registry::RUNNER_VERSION,
+                    "adopting stale structured view worker to drain in-flight turn before respawn"
                 );
-                state.acp_supervisor.mark_build_respawn_pending(&id);
+                let mut reasons = Vec::new();
+                if !build_current {
+                    reasons.push(RespawnReason::BuildStale);
+                }
+                if !runner_current {
+                    reasons.push(RespawnReason::RunnerStale);
```
jerome-benoit: [P3] Remove the unreachable `RunnerStale` reason

This block can only execute for `AdoptStaleForDrain`, but `adopt_decision` returns `RespawnStaleIdle` whenever `runner_current` is false. Therefore `!runner_current` is impossible here, `RunnerStale` is never constructed in production, and the comments claiming build and runner staleness can combine are false. Simplify this path to `BuildStale` and update the comments, or change the decision policy if cross-generation draining is intended.

jerome-benoit (changes requested):
Reviewed exact head `41fdb54d0283c71636b81e2f326e8e8ee36fe59a`. I did not repeat the issues already covered by the existing njbrake and CodeRabbit reviews. Two additional protocol-correctness issues remain blocking: reverse calls created after detach can survive until reattach, and the outbound queue is opened before `Attach` validates the peer. I also found an unreachable `RunnerStale` path that contradicts the current reconciliation policy. CI is green at this SHA, but GitHub currently reports the PR as conflicting.

src/acp/acp_client/terminal_handlers.rs
```diff
@@ -6,17 +6,16 @@ use agent_client_protocol::schema::v1::{
...
 
 use super::fs_handlers::enter_timestamp_ns;
 use super::SessionResources;
 
 pub(super) async fn handle_create_terminal(
     request: CreateTerminalRequest,
-    responder: Responder<CreateTerminalResponse>,
     res: SessionResources,
-) -> agent_client_protocol::Result<()> {
+) -> Result<CreateTerminalResponse, agent_client_protocol::Error> {
```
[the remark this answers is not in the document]
jerome-benoit: Addressed at the transport invariant in f707c600 rather than with a handler-local cache. The replay condition came from the detached reverse-call race: a post-disconnect `ServerCall` could enter the retained queue. `forward_server_call` and `disconnect_control` now serialize attachment state, pending insertion, and queue purging under the same lock. Consequently `terminal/create` is either delivered once to the currently validated daemon or answered with `DAEMON_GONE`; it cannot be queued for a later daemon. The reverse-call regression includes `terminal/create` specifically and verifies a call emitted after disconnect is neither pending nor queued.

src/server/acp_reconciler.rs
jerome-benoit: Fixed in f707c600. `RunnerStale` and the `RespawnReason` abstraction were removed. Runner-generation mismatch still goes directly to `RespawnStaleIdle`; only a build-stale, current-generation worker can reach `AdoptStaleForDrain`. The pending set is now a `HashSet<String>`, and the comments/log fields describe only build-stale draining.

src/process/runner.rs
jerome-benoit: Fixed in f707c600. `forward_server_call` now holds the control-state lock across the attached check, pending insertion, and queue insertion. `disconnect_control` takes the same lock while marking detached, draining pending calls, and purging their queued frames. Calls observed after detach are answered immediately (`cancelled` for permission, `DAEMON_GONE` otherwise) and never enter the replay queue. The regression covers both an in-flight call and a call emitted after disconnect.

tests/integration/acp_runner_control.rs
```diff
@@ -185,6 +246,148 @@ fn write_frame(stream: &mut UnixStream, body: &serde_json::Value) {
...
+/// Asserts the session completes, which it cannot do if the loop parks.
+#[test]
+fn agent_request_during_session_new_does_not_deadlock_the_runner() {
+    if cfg!(not(unix)) {
+        return;
+    }
+    let Some(python3) = find_python3() else {
+        eprintln!("skipping: python3 not found for fake ACP agent");
+        return;
+    };
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in f707c600. `find_python3()` now resolves `python3` through `PATH` with `which::which` before consulting the existing absolute fallbacks. The runner-control integration suite executed all seven scenarios successfully.

src/process/runner.rs
jerome-benoit: Fixed in f707c600. After writing `Hello`, the runner now requires a matching v3 `Attach` as the first inbound frame under a two-second deadline. Only then does it start the writer and mark the outbound queue ready. The new integration regression proves a non-`Attach` peer cannot consume buffered traffic and that the next valid peer receives the intact backlog.

src/acp/acp_client/control.rs
```diff
@@ -530,11 +837,11 @@ mod tests {
 
         assert!(
             client.is_none(),
-            "absent control socket must fall back to the watchdog"
+            "absent control socket must not yield a control client"
         );
         assert!(
             !guard.claimed(),
-            "absent control socket must fall back to the watchdog"
+            "absent control socket must not yield a control client"
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in f707c600. The guard assertion now reports `absent control socket must not claim the terminal`. The control test module passes.

src/process/worker_registry.rs
```diff
@@ -857,6 +935,80 @@ mod tests {
...
+                }
+                std::thread::sleep(std::time::Duration::from_millis(50));
+                false
+            });
+            let _ = victim.kill();
+            let _ = victim.wait();
+            assert!(
+                reaped,
+                "terminate must signal the live v1 runner, not orphan it"
+            );
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in f707c600. The spawned `sleep` is now wrapped in a `KillOnDrop` guard that always kills and waits in `Drop`. The explicit post-assertion cleanup is gone, so an assertion failure cannot leave the stand-in runner alive. The focused worker-registry regression and the full Rust suite pass.

jerome-benoit (approved):
Re-reviewed exact head `f707c60019ef0877abfde33b112fa8761ae22ca6` through two independent blind analyses and code-level cross-validation.

The three findings from my previous `REQUEST_CHANGES` are fixed: reverse calls created after detach are rejected instead of replayed, `Attach` is validated before the outbound queue opens, and the unreachable runner-staleness branch is gone.

I also checked the three current CodeRabbit concerns. The socket-cleanup race and the stale `respawn_pending` marker are real P2 follow-up defects, but both already exist in the exact base; this PR does not introduce the underlying interleavings and narrows the latter to current-generation runners. The fixed 150 ms sleep is a real P3 determinism gap in the new regression test, already covered by CodeRabbit's inline comment, but it is not a production blocker.

All exact-head CI checks are green. I found no remaining in-scope blocker.

src/acp/acp_client/control.rs
```diff
@@ -230,51 +326,113 @@ pub(super) async fn connect_runner_control_v2(
...
+                        // after the daemon already committed the terminal;
+                        // that case must be ignored rather than published
+                        // twice. Clearing the flag first hands idle ownership
+                        // back before any terminal event is emitted.
                         let was_in_flight =
                             reader_prompt_in_flight.swap(false, AtomicOrdering::Relaxed);
-                        if claimed {
-                            // The expected shape: a turn adopted at reattach,
-                            // whose completion this connection has to surface.
+                        if !was_in_flight {
```
jerome-benoit: [P1] Initialize the adopted-turn completion state before reading control frames. Production creates `prompt_in_flight` as `false` in `acp_client/mod.rs:578`; resume only sets the separate `adopted_turn_active` flag in `connection.rs:864-874`, while this flag becomes true only for a newly issued prompt at `connection.rs:1672`. Therefore every waiterless `PromptCompleted` for a real adopted turn takes this branch and is discarded, including its native stop reason or rate-limit metadata. The unit tests mask this by constructing the atomic as `true`. Derive an explicit adopted-turn state from `ConnectMode::Resume { in_flight_turn: true }` before the reader starts, and cover an immediately replayed completion plus an already-durable terminal through the production attach path.

src/acp/acp_client/control.rs
```diff
@@ -230,51 +326,113 @@ pub(super) async fn connect_runner_control_v2(
...
+                        let line = serde_json::json!({
+                            "jsonrpc": "2.0", "id": id, "error": error,
+                        });
+                        if !shim_write_line(&reader_shim_write, &line).await {
+                            return;
+                        }
+                    }
+                }
                 Ok(Some(_)) => {}
                 Ok(None) => return,
```
jerome-benoit, line 476: [P1] Propagate control EOF to the synthetic ACP transport. Returning here leaves the outbound pump blocked on `shim_read` while it retains the duplex halves (`455-459`), and the completion sender remains in `DaemonControlClient`. The crate-side connection therefore need not observe EOF, pending prompt/forward requests can remain parked, and `on_close`/supervisor recovery may never run. Dropping one Tokio split half is insufficient because the halves share the underlying `DuplexStream`. Supervise both bridge tasks as one lifetime: on reader termination, clear the completion waiter, stop the pump, and close both shim halves. Restore coverage for runner EOF with outstanding prompt and forward calls.

src/process/runner.rs
```diff
@@ -839,404 +818,478 @@ async fn write_control_frame(
...
+            let mut pending = self.pending_server_calls.lock().await;
+            let ids: Vec<u64> = pending
+                .iter()
+                .filter_map(|(id, call)| (call.attachment_id == attachment_id).then_some(*id))
+                .collect();
+            let drained: Vec<_> = ids
+                .into_iter()
+                .filter_map(|id| pending.remove(&id).map(|call| (id, call)))
+                .collect();
+            self.pending_agent_calls
```
jerome-benoit: [P2] Preserve durable reset bookkeeping after the requesting daemon disconnects. If `session/new` was already written to the agent, this purge removes the only mapping that records the method. A later successful response then bypasses `refresh_session_from_reset` (`917-923`, `1477-1504`), so the agent has committed a new conversation while the runner record and handshake cache retain the old session ID. Do not replay the stale `AgentResult` to another daemon, but retain enough internal state to persist/cache a successful reset result. Add the sequence `session/new` sent -> control disconnect -> agent response -> reattach`.

src/server/acp_reconciler.rs
```diff
@@ -1483,15 +1497,39 @@ async fn resume_one(state: Arc<AppState>, target: ResumeTarget) -> ResumeOutcome
...
+            crate::process::worker_registry::terminate_and_wait(&id).await;
+        } else if decision == AdoptDecision::ReplaceIncompatibleRunner {
+            tracing::info!(
+                target: "acp.supervisor",
+                session = %id,
+                old_runner_version = record.runner_version,
+                new_runner_version = crate::process::worker_registry::RUNNER_VERSION,
+                "replacing incompatible structured view runner"
+            );
+            if in_flight_turn {
```
jerome-benoit: [P2] Emit only one terminal event for this replacement. With an in-flight turn this publishes `runner_protocol_upgraded`, then falls through to the fresh-spawn block at `1651-1654`, which publishes a second `Stopped` with `orphaned_at_restart`. `synthesize_stopped_for_orphan` appends unconditionally, so the event log contains two terminal events and the generic reason becomes the latest one. Carry a `terminal_already_published` state or select the reason once at the fallback, and test that this path records exactly one `Stopped` with `runner_protocol_upgraded`.

jerome-benoit (changes requested):
Re-reviewed exact head `921213ec11debcdd26212afcf90ea1a48bc0f312` through two independent blind analyses and code-level cross-validation. The previous reverse-call, pre-`Attach`, runner-staleness, and replay-buffer-test findings are fixed, and all existing threads are resolved. All 32 exact-head checks are green.

Four lifecycle issues remain below: two P1 blockers and two P2 correctness defects. This was a static exact-SHA review; I did not run an additional local build or runtime reproduction.

src/acp/acp_client/control.rs
jerome-benoit: Fixed in 7174d2a3. The production socket path now initializes the shared completion state directly from `ConnectMode::Resume { in_flight_turn: true }` before `connect_runner_control_v3` starts its reader. `cached_completion_obeys_durable_in_flight_state_on_attach` primes a real runner, completes while detached, and verifies that production `AcpClient::attach` preserves `max_tokens` for an adopted turn while suppressing the same cached completion when durable state is already terminal.

src/server/acp_reconciler.rs
jerome-benoit: Fixed in 7174d2a3. The adoption decision is carried to the single fresh-spawn terminal publication; `ReplaceIncompatibleRunner` selects `runner_protocol_upgraded` there instead of publishing early and falling through to a second generic event. The regression uses a recording sink and asserts exactly one `Stopped` with the specific reason.

src/acp/acp_client/control.rs
jerome-benoit: Fixed on the current head. Commit 7135b10a makes reader termination clear the completion waiter, abort the outbound pump, and shut down both synthetic duplex halves; the restored EOF regressions pass. Commit 7174d2a3 also closes the control socket on a clean connection-task exit, and the production reattach regression verifies that the runner accepts the next daemon.

src/process/runner.rs
jerome-benoit: Fixed in 7174d2a3. Disconnect now retains only sent `session/new` bookkeeping from the dead attachment. Its late response still runs `refresh_session_from_reset`, while the stale attachment scope rejects the `AgentResult` instead of replaying it. The regression exercises send, disconnect, replacement attach, late response, updated cache, and empty outbound queue.

jerome-benoit (approved):
Approved after the fresh review/design/correction cycle. The four requested lifecycle fixes and two additional lifecycle races are covered at 7174d2a3.

src/server/api/acp.rs
```diff
@@ -2287,15 +2308,57 @@ pub async fn acp_disable(
...
+                Ok(storage
+                    .load()?
+                    .into_iter()
+                    .find(|candidate| candidate.id == id_for_load))
+            },
+        )
+        .await;
+        match persisted {
+            Ok(Ok(Some(durable))) if durable.is_structured() => {
+                instance = super::super::reload::merge_runtime_fields(instance, durable);
```
jerome-benoit: [P2] Preserve the source profile when adopting the durable row

When the cache says terminal but disk says structured, `Storage::load()` deserializes `source_profile` as empty (`instance/mod.rs:481`, `storage.rs:1126-1129`), and `merge_runtime_fields` does not restore it (`reload.rs:60-75`). The captured `profile` still targets the correct disable write, but the later `instance.start()` resolves its now-empty profile to the global default (`instance/start.rs:93-95`, `config/mod.rs:3402-3407`). If the session exists only in its original nondefault profile, lifecycle reservation fails (`instance/lifecycle.rs:199-223`) after the ACP worker has been stopped, leaving terminal view without a launched tmux session while the endpoint returns 200. Restore `source_profile` from the known `profile` before/after this merge and add a nondefault-profile stale-cache regression test. This consequence is source-traced, not an executed end-to-end reproduction.

jerome-benoit (changes requested):
Re-reviewed `7174d2a3..756a342e3649770377beb9718b22885079770bdf` with two blind sequential analyses and final cross-validation. The earlier runner lifecycle fixes remain outside this new finding.

**Request changes for one new P2:** the durable fallback in `acp_disable` loses the session's source profile before terminal launch; details inline. Both reviewers confirmed the narrowed consequence: launch failure in the stale-cache/nondefault-profile case, not demonstrated data duplication or wrong-profile execution.

**Nonblocking P3:** `acp_enable` now creates the per-instance lock before checking existence (`acp.rs:1990-1995`, `state.rs:429`). A missing-session response retains an unnecessary map entry. Preflight existence before allocation, retain the post-lock lookup, and mirror the existing missing-session tests for spawn/disable. This is a bookkeeping regression, not a claim of unauthenticated exploitation.

**Verified/rejected:** the deferred enable task uses the locked sandbox helper, avoiding recursive acquisition; the new epoch checks run under the instances write lock. No new lock inversion established. Existing persistence-error continuation and the broader reconciler stale-target issue were not reclassified as regressions from this commit.

Evidence: exact-object `git diff`, `git show` and `git grep`; `git diff --check` clean. Live CI, Tests and CodeQL are green at this SHA. Findings are source-traced, not local runtime reproductions; the changed Playwright file only updates comments and does not add a regression assertion for this fallback.

src/server/api/acp.rs
jerome-benoit: Fixed in `0e282e95`. The durable structured row is now adopted through `adopt_persisted_structured_instance`, which restores the cached source profile before runtime fields are merged. The regression asserts that a durable row deserialized without `source_profile` still selects the non-default launch profile.

jerome-benoit (approved):
Fresh review complete. All actionable findings are resolved.

jerome-benoit (approved):
Re-reviewed `0e282e95..6d11a8adec758a5c4bc8fe12f7be73482c2d9c4b` with two blind, sequential analyses and final code-level arbitration. No new in-scope blocker established.

Verified the added fixes: attachment-scoped prompt IDs and local/adopted terminal ownership (`src/process/runner.rs:1345`, `src/acp/acp_client/control.rs:350`); resume waits for the committed reset identity (`runner.rs:919,1446`); repair revalidates the runner under the transition lock and retains that lock through persistence (`src/server/reload.rs:363`, `src/server/structured_repair.rs:121`). The new regression assertions and protocol/documentation consistency were inspected.

The delta includes main imports: VT/web and migration production paths received targeted inspection, not a fresh line-by-line audit of every inherited test. Existing follow-up #3811 remains nonblocking; I have taken the macOS diagnostic and its scope clarification into account, rather than repeating the historical flake finding.

Evidence: exact-object `git diff`/`git show`; `git diff --check` clean; live discussions and CI checked. All 33 current checks succeeded. No new local tests were run in this pass. Both independent analyses agree on this disposition.

Issue #3417, opened 2026-08-18
Structured view / web: composer stuck on Stop + spinner after a finished turn, steered prompts bump the turn counter that a single Stopped cannot retire


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

A session that has finished keeps a live-looking composer in the web dashboard: the `Stop` button is rendered instead of `Send`, and the working spinner sits under the transcript. The sidebar correctly shows the session as finished. No further output ever arrives, and the state never heals; only a page reload clears it.

Trigger is mid-turn steering (#2805). Send one prompt, then type follow-ups while the turn is still running. Each follow-up gets injected into the running turn via `_session/steering`, and the turn ends with a single terminal `Stopped`. The web client counts prompts and stops, so N steered prompts leave N-1 phantom turns pending forever.

Every steering-capable adapter is affected, which today means `claude-agent-acp` at or above the steering floor, so this is the default path.

## Evidence from the affected session

Session `031472e08a1045cd`, from `~/.agent-of-empires/acp_events.db`:

```
seq 13   UserPromptSent            08:09:45
seq 79   CancelRequested           08:10:42
seq 81   Stopped{cancelled}        08:10:42
seq 82   UserPromptSent            08:11:34   <- fresh turn
seq 318  UserPromptSent            08:14:59   <- steered into the seq-82 turn
seq 370  UserPromptSent            08:15:40   <- steered
seq 516  UserPromptSent            08:17:44   <- steered
seq 1039 Stopped{prompt_complete}  08:29:27
```

Histogram for the session: `UserPromptSent` 5, `Stopped` 2.

Steering was live:

```json
{"PromptCapabilities":{"image":true,"audio":false,"embedded_context":true,"steering":true}}
```

Nothing else could have retired the extra pending turns. A query for `Stopped` / `PromptRejected` / `PromptRuntimeError` / `AgentStartupError` with `seq > 82 AND seq < 1039` returns zero rows.

## Root cause

**The daemon is correct.** A steered prompt deliberately emits no extra terminal event, `src/acp/acp_client.rs:8392`:

```rust
// No event: the prompt handler
// already published this text as
// `UserPromptSent` before it
// reached the daemon, and the
// running turn's own
// `PromptResponse` still owns the
// terminal Stopped.
```

[...]

Issue #3401, opened 2026-08-17
serve: user-initiated force-stop during an in-flight prompt returns HTTP 500


njbrake:
Every HTTP 500 in 10 days of my log is the same sequence, two occurrences, both identical. Pressing stop while a prompt is in flight is a normal user action, but it surfaces as a server error.

```
02:03:11.915277  WARN acp.protocol: force-stop requested during in-flight prompt; ending turn and restarting worker
02:03:11.915594  INFO acp.protocol: ACP connection task ended cleanly session=3ff5edb5a1f043c4
02:03:11.916317  WARN acp.supervisor: drain channel closed (agent connection task ended); evaluating respawn session=3ff5edb5a1f043c4 agent_unresponsive=true
02:03:11.916710  INFO acp.supervisor: SIGTERM wedged runner process group before respawn (agent_unresponsive) session=3ff5edb5a1f043c4 pid=1112
02:03:11.952062 ERROR http.request: completed status=500 latency_ms=0
02:03:12.019768  INFO acp.supervisor: worker registry deleted by user (`aoe acp stop|kill`); dropping WorkerHandle without respawn
```

The recovery itself works: the worker is torn down and respawned cleanly ~5s later, and the session continues. So this is a wrong status code on a path that otherwise behaves correctly, not a broken feature.

`latency_ms=0` suggests the request fails immediately rather than timing out, most likely because the handler's channel or worker handle is already gone by the time it tries to respond. The force-stop is raised at `src/acp/acp_client.rs:8481`.

Expected: a user-initiated stop returns 2xx, or a 409 if genuinely racing, and 500 is reserved for faults.

Related but distinct from the closed stop-button issues (#1908, #2237, #2805): those are about the stop not working; here it works and reports failure.

The companion logging issue (no method or path on `http.request` events) is why I cannot name the route.

<sub>Found by auditing 10 days of my own <code>debug.log</code> (v1.14.0/1.14.1, 25,416 lines). Investigation and draft text: Claude (Opus 5); reviewed by me before posting.</sub>

Issue #1281, opened 2026-05-20
Cockpit: rate_limit on session/prompt triggers respawn loop and missing rate-limit history signal


Seluj78:
**Note:** codex handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

When Claude Code hits account usage limits, `claude-agent-acp` returns a structured ACP error on `session/prompt` with `errorKind: "rate_limit"` and a reset time. Cockpit currently treats that as a worker crash, then immediately respawns the worker.

This creates a bad loop:
1. User sends prompt while limited.
2. Cockpit records `AgentStartupError`.
3. Supervisor respawns after 500ms.
4. Same ACP session is reattached with `session/load`, then the next prompt fails the same way until reset time.

The user sees churn, websocket reconnect noise, and repeated startup-error behavior for a condition that is not a process crash.

## Evidence from the affected sessions

From `~/.agent-of-empires/cockpit_events.db`:

```sql
SELECT seq, 'UserPromptSent' AS kind, substr(json_extract(event_json,'$.UserPromptSent.text'),1,120) AS detail
FROM cockpit_events WHERE session_id='98cc1bb786b5471a' AND json_extract(event_json,'$.UserPromptSent') IS NOT NULL
UNION ALL
SELECT seq, 'AgentStartupError', substr(json_extract(event_json,'$.AgentStartupError.message'),1,180)
FROM cockpit_events WHERE session_id='98cc1bb786b5471a' AND json_extract(event_json,'$.AgentStartupError') IS NOT NULL
UNION ALL
SELECT seq, 'AcpSessionAssigned', json_extract(event_json,'$.AcpSessionAssigned.acp_session_id')
FROM cockpit_events WHERE session_id='98cc1bb786b5471a' AND json_extract(event_json,'$.AcpSessionAssigned') IS NOT NULL
ORDER BY seq;
```

Output:

```text
2|AcpSessionAssigned|4f40f97d-f347-49f7-a51c-f56f6b3fefbe
4|UserPromptSent|/aoe-investigate https://github.com/njbrake/agent-of-empires/issues/1223
14|AgentStartupError|ACP connection failed: Internal error: You've hit your limit · resets 12:10pm (Europe/Paris): {
  "errorKind": "rate_limit"
}
15|AcpSessionAssigned|4f40f97d-f347-49f7-a51c-f56f6b3fefbe
```

Same pattern repeats in two other sessions from the same incident:

[...]

PR #3539, opened 2026-08-26
fix(acp): order a cancel behind an in-flight prompt submission

Seluj78:
## Description

> **Rewritten after a rebase.** This PR started as three live-spec de-flakes plus a standalone dispatch reservation. While it sat in review, #3618 fixed one of those specs independently (with better evidence) and deleted the other two, and #3639 / #3673 built the per-session prompt-submission guard my reservation was a hand-rolled precursor to. What is left is one focused change on top of that guard. Earlier review comments describe the old diff; the reasoning trail is preserved in the thread and the old branch is at `backup/pre-rebase-3539`.

`POST /acp/prompt` answers 202 and *then* works: it wakes a dormant session, folds control state, and only then hands the prompt to the agent. `POST /acp/cancel` reaches the command channel almost immediately.

So a Stop pressed right after Enter lands **first**. The connection loop takes its no-prompt-in-flight branch (`acp_client.rs`), and the agent clears its cancel state when the prompt finally opens a turn. The turn then runs to completion and the composer never returns to "Send a message".

This is user-reachable, not a test artifact: the Stop affordance rides the client's optimistic `running` flag, so it is clickable before the daemon owns a turn at all. CI caught it on two live specs with the two POSTs **78ms apart**:

```
POST /api/sessions/<id>/acp/prompt -> 202   48.782
POST /api/sessions/<id>/acp/cancel -> 202   48.860
```

### The fix

`acp_cancel` waits for the per-session prompt-submission guard before forwarding, and holds it across the forward.

That guard already exists and is already the single authority: #3639 and #3673 made every turn-starting surface decide and dispatch under it, including both prompt endpoints, the plugin host's `sessions.turn.send`, and both queue drains. Waiting for it therefore orders the cancel behind the prompt on **every producer at once**, without introducing a second mechanism beside it.

**Bounded and best-effort.** On timeout, or for an unknown session, the cancel is forwarded unguarded, which is exactly today's behaviour and never worse. 2s is sized for the case that matters:

| Worker state | Guard held for | Stop waits |
| --- | --- | --- |
| live | a decide plus a channel enqueue | effectively not at all |
| cold / dormant | up to `WORKER_READY_TIMEOUT` | bounded at 2s, and no turn is running to cancel |

## Testing

`cancel_waits_for_an_in_flight_prompt_submission` drives `acp_cancel` in-process, holds the real submission guard to stand in for a prompt mid-dispatch, and asserts the endpoint does not sail past it.

Two things about that test worth a reviewer's attention, because the obvious version of it is wrong:

- **It installs a live fake worker on purpose.** Without one, `cancel_prompt` parks on `wait_for_worker` for `WORKER_READY_TIMEOUT` whether or not this fix is present, so "still running" proves nothing. My first version of this test **passed with the fix reverted** for exactly that reason.
- **It asserts the unguarded baseline is fast** before the real assertion, so if that fixture ever stops being a fast path the test fails loudly instead of quietly going tautological again.

Mutation-checked: deleting the wait fails it with *"acp_cancel forwarded while a prompt submission held the guard"*.

| Check | Result |
| --- | --- |
| `cargo test --features serve,e2e-tests --lib --bins` | 6570 passed, 0 failed |
| `cargo clippy --all-features` | clean |

**Not verified:** the underlying race does not reproduce on my machine, with or without the fix, in either the old or the new implementation. The prompt wins the dispatch locally every time; the 78ms inversion needs CI's parallel load. So this rests on the captured traces plus the mutation-checked test, not on a local red-to-green demonstration. Weigh it on the argument and the code.

Also note `Cargo Test (macos)` on a PR is the ~5s no-op shim (`tests.yml`), as always.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No UI change. The invariant is documented in `docs/development/internals/structured-view.md`, next to the prompt-dispatch paragraph it constrains.

## Test Coverage Analysis

No web files are touched any more. The live specs that originally motivated this were fixed or merged upstream by #3618; `composer-stop.spec.ts` still covers Stop end to end.

**TUI / CLI (e2e test)**

Covered by the in-process handler test above. Asserting "the agent observed prompt-then-cancel" end to end needs a real worker plus a deterministic interleaving of two in-flight requests; the orderable window is what this change owns.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Triage from CI trace artifacts (unpacking the Playwright report and trace zips to read request timings and attached daemon/agent logs), root-causing, and the fix were done with Claude Code. Both the earlier "not verified" note and the tautological-test finding came from deliberately reverting each half of the fix to check whether anything actually failed.

## Summary

- Prevents `acp_cancel` from overtaking an in-progress prompt submission.
- Adds a bounded wait for the shared per-session submission guard.
- Holds the guard while forwarding the cancel request.
- Uses the existing behavior if the session is unknown or the wait times out.
- Adds regression coverage and documents the prompt-dispatch ordering.

## Benefits

This reduces lost Stop requests during prompt dispatch. It keeps prompt producers consistent without adding a second dispatch mechanism.

## Further consideration

The race was not reproduced locally. Captured CI traces and mutation-checked regression coverage support the fix.


src/server/api/acp.rs
```diff
@@ -1319,6 +1319,13 @@ pub async fn acp_prompt(
         Ok(j) => j,
         Err(rej) => return rej.into_response(),
     };
+    // Reserved before any of the work below, because that work is the window a
+    // Stop pressed right after Enter races: the cancel reaches the command
+    // channel while this handler is still waking a dormant session, folding
+    // control state, or waiting on worker readiness. Held to the end of the
+    // handler so every early return releases it. See
+    // `Supervisor::begin_prompt_dispatch`.
+    let dispatch_guard = state.acp_supervisor.begin_prompt_dispatch(&id);
```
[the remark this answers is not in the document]
Seluj78: Correct, and worse than "Minor" — thank you. Fixed in https://github.com/agent-of-empires/agent-of-empires/commit/a5e93c309a03547ab80140342e1a89ddb13fef3e.

`PromptDispatch::Queued` is reached precisely BECAUSE a turn is already running, so the discarded latch was leaving a live turn running after the user pressed Stop. And before I added the reservation the cancel was forwarded and did stop that turn, so this was a regression I introduced, not merely an unhandled case.

I did not take the literal suggestion of consuming the latch on each non-dispatch exit. That keeps the invariant "every early return in `acp_prompt` must remember to forward", which is the same assumption that produced the bug, and `acp_prompt` has a lot of early returns. Instead I removed the deferral: `acp_cancel` now always forwards the cancel, and the latch only adds the re-send after an in-dispatch prompt is away. Losing a cancel is now structurally impossible, and no exit needs per-path handling.

Trade accepted: a duplicate `session/cancel` inside the race window, which the in-flight branch resends harmlessly, plus the synthetic Stopped the no-turn-in-flight branch already publishes on main today. I originally deferred to avoid that transient; avoiding a cosmetic artifact was not worth risking a Stop that does nothing.

On the regression test: the guard-level case is in (`a_cancel_during_prompt_dispatch_is_resent_after_the_prompt`, renamed since "deferred" no longer describes it), pinning that a latch dropped on a non-dispatch exit is gone and therefore that the forward at cancel time is what stops the running turn — so a future move back to deferring fails the test. I did not add the full HTTP-level "queue a follow-up and cancel during enqueue" test: it needs a real spawn plus a deterministic interleaving of two in-flight requests, and the part a future edit can realistically break is the reservation logic, which is now covered directly.

Verification: `cargo test --features serve,e2e-tests --lib --bins` 6419 passed / 0 failed, `clippy --all-features` and `fmt` clean, and the three live acp stories pass locally (18.0s).

njbrake (commented):

_Reviewed at `a5e93c30`. Push new commits and it will be re-reviewed._

---

## Summary

Two separable halves. The server half reserves the accept-to-dispatch window in `acp_prompt` so a `POST /acp/cancel` landing inside it latches onto the reservation instead of being forwarded into the connection loop's no-prompt-in-flight branch; the dispatch re-sends it once the prompt is away. The test half replaces three live specs' reliance on optimistic client state with server-driven signals, and adds a `waitForQueuedPrompts` helper.

The triage is good work. Diagnosing a real ordering bug out of CI trace timings, on a tree the author correctly established had not regressed, is exactly the right instinct, and the write-up separating "verified" from "not verified" is unusually honest. The mechanism checks out end to end: `acp_client.rs:8786` is the no-prompt-in-flight branch, and `web/tests/helpers/fakeAcpAgent.mjs:827` does reset `cancelFlags` on `session/prompt`, so a cancel that arrives first is provably destroyed in the fake agent. A real adapter has no better answer for a cancel with nothing to cancel, so "user-reachable, not a test artifact" is a fair claim.

But the fix as pushed introduces a new way to lose a cancel, in a flow that is at least as common as the one it fixes. That is blocking.

## Design alignment

**Aligned.**

The precedent is right next door in the same file: `Supervisor::begin_resume` (`src/acp/supervisor.rs:1580`) hands back a `ResumeReservation` RAII guard backed by `Arc<std::sync::Mutex<HashMap<..>>>` on the Supervisor, with `cancelled_resumes` as a companion breadcrumb set consumed by a later operation. `begin_prompt_dispatch` / `PromptDispatchGuard` / `prompt_dispatches` is that same shape with a generation counter added. This extends the existing mechanism rather than standing a second one beside it, and `PromptDispatch` / `QueueReason` in `src/acp/dispatch.rs` stay the single source of dispatch truth.

`data-thinking` on `StructuredView.tsx:1590` is not a test affordance leaking into production. `web/src/` already carries `data-checked`, `data-selected`, `data-pane-focused`, `data-sort-mode` and a dozen others as state mirrors, and the attribute mirrors `state.thinking`, which is `reduced.thinking != null` from the server-owned `reduced_state` frame (`web/src/lib/acpTypes.ts:1806`). It is genuinely server-driven, which is the whole claim the spec rests on. 33 bytes of bundle. Fine.

One deviation: the file has a `lock_recover` helper (`src/acp/supervisor.rs:3648`) used at 25 call sites for exactly this kind of `std::sync::Mutex`, including on the sibling `pending_resumes` and `cancelled_resumes` fields. The three new locks use `.lock().expect("prompt dispatch mutex")` instead, so a poisoned lock panics a request handler rather than recovering. Non-blocking, but it should match its neighbours.

## Issues found

### 1. Blocking: the `Queued` path swallows a cancel aimed at the turn that is actually running

`src/server/api/acp.rs:1381` takes the reservation at the very top of the handler. `src/server/api/acp.rs:1381-1417` is the `Queued` early return, and it drops `dispatch_guard` without consuming it, so `Drop` discards any latched cancel.

Read `src/acp/dispatch.rs:76-99`: three of the four queue reasons (`TurnActive`, `Compacting`, `Cancelling`) exist precisely *because a turn is already running*. So the reachable scenario is:

1. A long turn is running on a non-steerable agent.
2. The user types a follow-up and hits the queue button. `POST /acp/prompt` takes the reservation and starts working.
3. Mid-handler, the user changes their mind and presses Stop.
4. `acp_cancel` (`acp.rs:1656`) sees the reservation, latches, returns 202, forwards nothing.
5. `acp_prompt` reaches the `Queued` branch, returns, `dispatch_guard` drops, latch discarded.
6. The running turn is never cancelled. The composer sits in "Stopping..." until the turn ends on its own.

On `main` that same Stop is forwarded to `cancel_prompt` and correctly ends the running turn. This is a regression, not merely an uncovered case, and it is the flow `queue-follow-up-with-nav` exercises.

The window is not negligible. It spans `touch_on_prompt_and_wake_if_sunk`, the `instances` read, `validate_attachments`, `state.instance_lock(&id).await` plus `clear_pending_initial_turn`, `fold_control_state`, and `buffer_and_enqueue`'s storage write. The instance lock is the one to worry about: the comment at `acp.rs:1352-1360` says the turn-end drain holds that same lock across its whole snapshot-to-reload-to-send-to-clear, so the handler can park there for a while with the reservation held.

CodeRabbit flagged this independently at `acp.rs:1328`. **Agree.** It is the one finding on the PR and it is correct.

Suggested fix (staged locally, verified compiling, `cargo test --features serve` green):

```rust
if let crate::acp::dispatch::PromptDispatch::Queued { reason } = dispatch {
    // Every queue reason but `WorkerDown` means a turn is already running, so a
    // Stop latched onto this dispatch was aimed at THAT turn, not at the prompt
    // we are about to park. Forward it before the reservation goes out of scope:
    // letting the guard drop here would discard the cancel, which is strictly
    // worse than the pre-reservation behavior of forwarding it straight through.
    // `WorkerDown` is excluded because there is no turn to cancel and
    // `cancel_prompt` would burn the whole `WORKER_READY_TIMEOUT` before failing,
    // stalling this enqueue behind it.
    if reason == crate::acp::dispatch::QueueReason::WorkerDown {
        drop(dispatch_guard);
    } else {
        forward_deferred_cancel(&state, &id, dispatch_guard).await;
    }
    // ... existing enqueue
```

with the existing `Ok(())` block extracted into a shared `forward_deferred_cancel(&state, &id, guard)` helper so both sites use one implementation. The `WorkerDown` exclusion matters: `cancel_prompt` goes through `ready_client` to `wait_for_worker(WORKER_READY_TIMEOUT)`, so forwarding unconditionally would stall the enqueue response by up to 10s when no worker exists.

The deeper point is that `Drop` currently defaults to *discard*. For a reservation whose entire purpose is "do not lose this cancel", the safe default is the opposite. Every naked early return is a potential repeat of this bug, and there are five more of them.

### 2. Non-blocking: two more exits with the same shape

- `acp.rs:1341` (`validate_attachments` error). A turn can be running here, and the dispatch decision has not been made yet. Rarer, but the same loss.
- The `send_turn` `Err` arms. When `dispatch` was `Steered` a turn *is* active, so a latched cancel is lost there too. `Sent` is fine (no turn exists). Not staged, because it needs the `Steered`-only condition threaded through the match and the failure mode (a steering send that errors) is rare enough that the shape is the author's call.

### 3. Non-blocking: the reservation is a single slot per session, and a second prompt silently evicts the first

`begin_prompt_dispatch` does an unconditional `insert`, so two concurrent `POST /acp/prompt` for the same session leave only the newer generation in the map with `cancel_requested` reset to `false`. Any cancel latched onto the first is destroyed by the second prompt's insert, and the first guard's `remove_if_current` then returns `None` and forwards nothing. The generation counter correctly prevents the *stale guard clobbering a live slot* case (and the test pins that), but it does not preserve the latched flag across the eviction. Narrow, and the ejected prompt is not the common case, but worth a comment at minimum saying the single-slot design is deliberate.

### 4. Non-blocking: two doc comments got reattached to the wrong items

The new items were spliced between an existing doc block and the item it documented, in two places:

- `src/acp/supervisor.rs:2754-2764`: the `cancel_prompt` doc comment (the #3401 force-stop-window explanation) now prefixes `begin_prompt_dispatch`, and `cancel_prompt` is left undocumented. Reading the new function's rustdoc now leads with a paragraph about a dead `WorkerHandle` that has nothing to do with it.
- `src/acp/supervisor.rs:5224-5230`: the #1281 rate-limit drain doc block now prefixes `a_cancel_during_prompt_dispatch_is_deferred_to_the_dispatch`, and `drain_skips_restart_when_stopped_rate_limited` lost its explanation.

Pure motion to fix: move the two new items below the block that was already there.

### 5. Non-blocking: dead branch in `waitForQueuedPrompts`

`web/tests/helpers/aoeServe.ts:211`: `Array.isArray(body) ? body.length : (body?.queued_prompts?.length ?? 0)`. `queue_list` (`src/server/api/queue.rs:237`) returns `Json(queued_prompts_snapshot(..))`, a bare array, so the second arm is unreachable. Worse, if the shape ever does change, `?? 0` converts that into a 15s timeout with a message blaming the daemon. `listSessions` twelve lines above does the right thing and throws on an unexpected shape. `expect.poll` propagates a thrown callback immediately (the codebase documents this at `aoeServe.ts:264-267`), so throwing here fails fast with the real reason. AGENTS.md's no-dead-code rule applies.

## Do the tests still prove the fix?

This is the half worth being sceptical about, and the answer splits cleanly.

**`queue-follow-up-with-nav.spec.ts` (+8/-1): genuine fix, and it settles the open question.** The abort explanation is self-consistent and the trace evidence (`POST /acp/prompt` status `-1` seventeen milliseconds before the `/settings` request) is the right kind of evidence. `waitForQueuedPrompts` is a bounded `expect.poll` on the daemon's own `GET /queue` snapshot, throws loudly at 15s with a specific message, drops no assertion, and matches `waitForSessions` from #3535 line for line. It also *discriminates* the two hypotheses the earlier triage was stuck between: if the daemon had actually died, the fetch throws immediately rather than polling out. That is a better test than what it replaces, and it is not papering over anything.

**The two composer-stop specs (+13, +8): the waits are real signals, but they retire the coverage that caught the bug.** `data-thinking="true"` really is `ThinkingStarted` from the server fold, and `Task: investigate` really is the agent's own `tool_call` row arriving over the event stream. Neither is a sleep, a retry, or a widened timeout, and both fail loudly at a bounded 15s. On the narrow question, the tests were not relaxed.

But once those waits are in place, the prompt has demonstrably reached the agent, the dispatch window is closed, and those specs can no longer reach the race in either direction. That is defensible on its own terms (their stated subject is "Stop cancels a thinking turn", not "Stop races the dispatch"), and the author is upfront that this is deliberate. The consequence, though, is that **nothing in the tree exercises the server fix's wiring**. The Rust test pins the supervisor bookkeeping: latch, no-latch, abandon-before-dispatch, stale generation. It never touches `acp_prompt` or `acp_cancel`. Delete the `begin_prompt_dispatch` call from `acp_prompt`, or the `take_pending_cancel` on the `Ok` arm, and the entire suite stays green. That is precisely the gap that let issue 1 through: a handler-level bug in code whose only test is one layer below the handler.

The PR argues a behavioural test "would need a real spawn plus deterministic scheduler interleaving". That is stronger than warranted. `wake_prompt_frees_instance_lock_and_publishes_nothing_without_a_worker` (`src/server/api/acp.rs:3376`) already drives `acp_prompt` in-process against `build_test_app_state`, and parks it mid-handler by holding a `ResumeReservation`. The same recipe pins at least two things worth pinning without any agent process:

- `acp_cancel` returns 202 and does **not** forward while a reservation is live.
- The `Queued` path drains the latch instead of dropping it (issue 1), which is the regression a future edit is most likely to reintroduce.

The full "the agent observed prompt-then-cancel" assertion does need a worker; those two do not.

Test *shape* is otherwise house style: one test fn, four scenarios as commented blocks inside it, mutation-checked against the generation branch. Correct.

| Check | Result |
| --- | --- |
| `cargo fmt --check` | clean (pristine head and with the staged fixes) |
| `cargo clippy --features serve --all-targets -- -D warnings` | 5 errors, **all pre-existing and unrelated** (`src/process/macos.rs` `items_after_test_module`, `src/session/smart_rename.rs` `useless_vec`); neither file is in the diff, and CI's Clippy job passes, so this is local toolchain drift on rust 1.96 |
| `cargo test --features serve` (`env -u CLAUDE_CONFIG_DIR`) | **green**: 6409 plus 24 integration targets, 0 failed |
| `npm run format:check` (oxfmt) | clean |
| `npm run lint` (eslint) | clean |
| `npx tsc --noEmit` | clean |
| Vitest | 353 files, 3706 tests, 0 failed |
| `node tests/validate-coverage-matrix.mjs` | passed |
| Full PR CI | all 34 checks green, including both live Playwright shards |

**Coverage matrix**: no entry needed, and the author's reasoning is right. All three specs already have rows (`acp.story.queue-follow-up-with-nav` at `coverage-matrix.json:3179`, plus 3147 and 3189); no new flow was added, and the validator passes on the tree as pushed.

Everything above was run. The live Playwright specs were **not** run (they need a built binary and real daemons); those three were judged by reading them and their helper, and by confirming that `data-thinking` and the `Task: investigate` row are both server-driven rather than client-optimistic.

## Documentation

Proportionate, with one thing missing. The prose in the diff is the "why" kind AGENTS.md wants: the `prompt_dispatches` field comment is the single canonical explanation of the race with the measured 78ms, and the spec comments explain a non-obvious wait rather than restating the code. No standalone markdown file, no narration of the change. No emdashes in added prose.

Missing: `docs/development/server-owned-prompt-dispatch.md` is the named home for the Tier 3 dispatch contract and is cited from `src/acp/dispatch.rs:2`. The accept-to-dispatch reservation is a new invariant on that contract (a cancel arriving mid-handler is deferred, not forwarded) and belongs there in two or three lines, especially since issue 1 shows the invariant is easy to get wrong at a new exit.

## Automated review check-in

- **CodeRabbit, `acp.rs:1328`, forward a latched cancel on the non-dispatch exits including `Queued`**: **agree**, and it is the blocking issue. This review reached the same place independently. Their suggested regression test is also right, and the harness for it already exists.
- **CodeRabbit merge risk "Moderate"**: agree with the level and the reason.
- **CodeRabbit title check (94 chars)**: cosmetic. The title is accurate and does two things because the PR does two things.
- **CodeRabbit docstring coverage 64%**: **disagree**, false positive. AGENTS.md explicitly does not want docstrings that restate signatures, and the functions here that lack rustdoc are the ones that need none. Do not act on this.
- **Codecov**: patch and project green; all modified web lines covered.

## Merge risk

**Moderate**, and it is the residual from issue 1 rather than anything speculative.

- **Blast radius**: silent. A lost Stop leaves a turn running and the composer wedged in "Stopping..." with no error anywhere. The user's recourse is Force stop or a page reload, and nothing in the logs says a cancel was dropped, because the `deferring until the prompt is away` info line fires on the path that *works* as well as the one that does not.
- **Reversibility**: clean `git revert`. No schema, no migration, no persisted state. The `data-thinking` attribute is additive and the spec waits are inert without it.
- **Coverage**: the specific untested path is `acp_prompt`'s `Queued` early return with a latched cancel, and more broadly every exit from `acp_prompt` other than `Ok(())`. Nothing at any tier exercises the reservation through the HTTP handlers.

Most likely way this bites in production: a user queues a follow-up on a running turn, presses Stop within the handler's window, and the turn keeps going. CI would not catch it, because the three specs that used to stumble into this family of bug now wait past the window. It would arrive as a user filing "Stop doesn't work sometimes", which is roughly how #3417 and this one both surfaced.

Apply the issue-1 fix and this drops to Low. The fix is small and mechanical; the staged fixes for issues 4 and 5 are pure motion and a thrown error, so they add no risk of their own.

## Suggestions

1. Fold the `Ok(())` forward block into a `forward_deferred_cancel` helper and call it from every exit that can coexist with an active turn, so the next early return added to `acp_prompt` does not silently reintroduce this.
2. Add the handler-level regression test using the `build_test_app_state` plus held-`ResumeReservation` recipe already in the file. Two assertions is enough.
3. Swap the three `.lock().expect("prompt dispatch mutex")` calls for `lock_recover`, matching the other 25 sites and the sibling reservation maps.
4. Consider `pub(crate)` on `begin_prompt_dispatch` and `cancel_during_prompt_dispatch`, matching `begin_resume`. There is one caller module.
5. In `composer-stop-during-sub-agent.spec.ts:79`, `page.getByText("Task: investigate")` is unscoped and will trip strict mode if that title ever renders twice. Scoping it to a testid would be sturdier for the same cost.
6. Three or four lines in `docs/development/server-owned-prompt-dispatch.md` recording the reservation invariant.

## Verdict

**Design alignment:** Aligned
**Project fit:** Accept
**Documentation:** Proportionate, one missing update (`server-owned-prompt-dispatch.md`)
**Merge risk:** Moderate, dominated by the `Queued`-path cancel loss being silent and untested

The diagnosis is sound, the design fits the codebase's existing reservation idiom, and the test changes wait on real server signals rather than waiting out the race, so the de-racing half is honest work and `queue-follow-up-with-nav` is strictly improved. But the reservation is taken at the top of a handler with six exits and consumed at exactly one of them, which converts a correct Stop into a lost one on the queue-a-follow-up-then-Stop flow. That is a regression against `main` in a path at least as common as the one being fixed, and it slipped through precisely because the fix's only test sits one layer below the handler. Fix the `Queued` exit, add the handler-level test, and this is a clear merge.

njbrake (commented):

_Reviewed at `a5e93c30`. Push new commits and it will be re-reviewed._

---

> **This supersedes the earlier automatic review on this PR, which was written against `57ab8abe`.** The head moved while that review was running. Judging **`a5e93c30`** here (rebased onto `origin/main` at `7641fd3b`), which adds one commit: `fix(acp): forward every cancel, and latch it only to re-send after dispatch`.

## The blocker is closed, and the new shape is better than the fix proposed last round

The previous review's issue 1 was that `acp_cancel` withheld the cancel whenever a reservation existed, and `acp_prompt`'s `Queued` early return then dropped the latch, losing a Stop aimed at a turn that was genuinely running. The recommendation was to forward the latched cancel at that one exit. The author did something better: inverted the default.

The scenario was re-derived against the new code line by line rather than trusting the commit subject.

**Setup:** a long turn is running on a non-steerable agent; the user types a follow-up and hits queue; mid-handler they press Stop.

1. `acp_prompt` takes the reservation at `src/server/api/acp.rs:1328` and starts working.
2. Stop arrives. `acp_cancel` at `acp.rs:1622` calls `cancel_during_prompt_dispatch`, which now sets the latch and returns a bool used **only for the log line**. The early `return` that used to sit under it is gone.
3. `acp.rs:1629` forwards `cancel_prompt` unconditionally. A turn *is* in flight (the long one), so the connection loop takes its prompt-in-flight branch and cancels it properly. **The running turn dies. This is the case that was broken.**
4. `acp_prompt` reaches `Queued` at `acp.rs:1381`, returns, `dispatch_guard` drops, latch discarded. Correct: no prompt was dispatched, so there is nothing to reorder against.

**Issue 1 is closed**, and closed structurally rather than patched. Losing a cancel no longer depends on someone enumerating every early return correctly, now or when the next exit gets added. That is the right property to want here, and the `Drop` comment at `supervisor.rs:390-398` and the `cancel_during_prompt_dispatch` doc at `supervisor.rs:2793-2800` both now state the invariant that makes the drop safe, which is what a future editor needs to read before changing it back.

## The trade this shape makes, checked

Forwarding always means the cancel goes out *before* the prompt in the race window, which is the original bug. The re-send is what saves it, so both halves were checked: that it fires everywhere it must, and that the doubling is harmless.

**Does the re-send cover every path that needs it?** Exactly one path puts a prompt in front of the agent: `send_turn` returning `Ok(())`, and that arm re-sends at `acp.rs:1440`. `Queued` never reaches the agent (the turn-end drain delivers it later, as a new turn the user still wants). Every `SendTurnError` variant is documented pre-publish except `Send`, which is "the forward to the agent failed", so no turn started there either. `Steered` goes through `send_turn` and gets the re-send. One path needs it, one path has it.

**Is the doubled `session/cancel` harmless?** Three things had to hold, and all three do:

- The first cancel hits `acp_client.rs:8787`'s no-prompt-in-flight branch, which calls `terminal_claim.claim()`. `TerminalClaim` (`acp_client.rs:1652`) is epoch-based and `begin_turn()` bumps the epoch when the prompt starts, so a claim taken before the turn cannot disarm that turn's terminal. No cross-turn leak.
- That branch also publishes a synthetic `Stopped { reason: "cancelled" }`. This is restored behavior versus main, not new. It cannot flicker the composer: `deriveTurnActive` (`web/src/lib/acpTypes.ts:1926`) is `serverTurnActive || inflightPromptIds.length > 0`, so a false frame arriving while the POST is unacknowledged does not flip `turnActive`. That is #3417's guard, and this window is exactly what it was built for.
- The second cancel could in principle land on the *next* turn, but only if the turn completes between `send_turn` returning `Ok` and `cmd_tx.send(ClientCmd::Cancel)` reaching the loop. `ready_client` is a worker-map lookup here (the worker is live by construction), so that is sub-millisecond, and the same hazard exists on main for any Stop pressed near turn end. Not a new risk of substance.

The one real behavioral change versus the previous head is that `acp_cancel` no longer short-circuits, so it awaits `cancel_prompt` and can therefore block on `wait_for_worker(WORKER_READY_TIMEOUT)` and can return a 5xx instead of 202. Both are main's behavior today. Restoration, not regression.

## Re-check of the other findings

| Prior finding | Status at `a5e93c30` |
| --- | --- |
| **1.** `Queued` exit swallows a cancel | **Closed** (above) |
| **2.** `validate_attachments` / `send_turn` Err exits, same shape | **Closed** by the same inversion. The cancel was already forwarded; dropping the latch on those exits only skips a re-send that no dispatched prompt needs. |
| **3.** Single-slot eviction (concurrent prompts on one session) | **Downgraded, still live.** `begin_prompt_dispatch` (`supervisor.rs:2776`) still does an unconditional `insert`, so prompt B's reservation resets `cancel_requested` to `false` and prompt A's re-send never fires. Under the old shape that lost the cancel outright; now it only loses the *re-send*, degrading to main's pre-PR behavior in that narrow window. Not a regression against anything. Worth one comment saying the single slot is deliberate; not worth code. |
| **4.** Two doc comments reattached to the wrong items | **Still live.** `supervisor.rs:2767` still glues `cancel_prompt`'s #3401 doc block onto `begin_prompt_dispatch`, leaving `cancel_prompt` undocumented; `supervisor.rs:5236` still glues the #1281 rate-limit drain doc onto the new test. Pure motion to fix: move the two new items below the block that was already there. |
| **5.** Dead `queued_prompts` branch | **Still live** at `web/tests/helpers/aoeServe.ts:211`. `queue_list` (`src/server/api/queue.rs:237`) returns a bare array, so the second arm is unreachable, and `?? 0` would turn a future shape change into a 15s timeout blaming the daemon. `listSessions` twelve lines above throws instead, and `expect.poll` propagates a thrown callback immediately (documented at `aoeServe.ts:264-267`). AGENTS.md no-dead-code. |

Also unchanged and still true from the first review: **design alignment is good** (the reservation mirrors `begin_resume` / `ResumeReservation` at `supervisor.rs:1580` in shape and storage); **`data-thinking` is not a test affordance leaking into production** (it mirrors `reduced.thinking != null` from the server fold at `acpTypes.ts:1806`, and `web/src/` already carries a dozen `data-*` state mirrors); and the **coverage matrix needs no entry** (all three specs already have rows at `coverage-matrix.json:3147/3179/3189`; validator passes).

## The test-coverage gap is now the main thing left, and the check was re-run

The new commit added one block to the existing supervisor unit test (`s5`, pinning that the queued-path drop happens) and renamed it to `a_cancel_during_prompt_dispatch_is_resent_after_the_prompt`. Good rename, and the new block is genuinely load-bearing documentation of why the drop is safe. Test shape still matches house style: one fn, scenarios as commented blocks, no test-per-case.

But it is still a supervisor-level test. No handler-level test was added: `begin_prompt_dispatch` and `cancel_during_prompt_dispatch` appear in `src/server/api/acp.rs` only at their two call sites (1328, 1622), never in its `mod tests`.

The mutation was re-run at this head. Deleting the entire re-send block at `acp.rs:1440-1454` and replacing it with `let _ = dispatch_guard;` leaves **6409 passed; 0 failed**. The one thing the server fix actually does at the handler level is unprotected by any test at any tier.

And the code it guards just changed shape underneath it. The `Queued`-path bug got into a pushed commit and was caught by a bot rather than by the suite, precisely because of this gap. The #3535 precedent still applies: this is the sort of thing worth pinning.

The PR argues such a test needs "a real spawn plus deterministic scheduler interleaving". That remains stronger than warranted. `wake_prompt_frees_instance_lock_and_publishes_nothing_without_a_worker` (`src/server/api/acp.rs:3376`) already drives `acp_prompt` in-process against `build_test_app_state` and parks it mid-handler by holding a `ResumeReservation`. The same recipe pins the two things a future edit is most likely to break, with no agent process:

- `acp_cancel` latches **and still forwards** while a reservation is live (the exact property this commit just established).
- The `Queued` path returns without consuming the latch.

The full "the agent observed prompt-then-cancel" assertion does need a worker. Those two do not.

## Do the tests still prove the fix?

Unchanged from the last review, and worth restating since it is the central question:

- **`queue-follow-up-with-nav.spec.ts`**: genuinely fixed. `waitForQueuedPrompts` is a bounded `expect.poll` on the daemon's own `GET /queue`, throws loudly at 15s with a specific message, drops no assertion, and mirrors `waitForSessions` from #3535 line for line. It also *discriminates* the daemon-death hypothesis from the abort hypothesis: a dead daemon makes the fetch throw immediately rather than poll out. Not papering over.
- **The two composer-stop specs**: the added waits are real server signals (`ThinkingStarted` via `data-thinking`, and the agent's own `tool_call` row), bounded and loud, not sleeps or widened timeouts. So the tests were not relaxed. But once those waits are in place, neither spec can reach the racing window in either direction, so neither exercises the server fix. That is defensible on its own terms and the author says so plainly. It just means the mutation result above is the whole story on coverage.

## Checks executed at this head

All run in a scratch worktree, rebased onto `origin/main` (`7641fd3b`), with `env -u CLAUDE_CONFIG_DIR`.

| Check | Result |
| --- | --- |
| `cargo test --features serve` | **green**: 6409 lib plus 24 integration targets, 0 failed |
| `cargo clippy --features serve --all-targets -- -D warnings` | 5 errors, all pre-existing and outside the diff (`src/process/macos.rs:266`, `src/session/smart_rename.rs:1546`, `src/acp/background_agent.rs:627/745`, `src/acp/supervisor.rs:4144`, a pre-existing test table between the PR's hunks at 2768-2815 and 5237). CI Clippy is green, so this is local rust 1.96 drift. |
| `npm run format:check` (oxfmt) | clean |
| `npm run lint` (eslint) | clean |
| `npx tsc --noEmit` | clean |
| Vitest | 353 files, 3706 tests, 0 failed |
| `node tests/validate-coverage-matrix.mjs` | passed |
| Mutation: delete the `acp.rs:1440` re-send | **6409 passed, 0 failed** (the gap) |
| PR CI at `a5e93c30` | 34 SUCCESS, 1 skipped, including both live Playwright shards |

Not run: the live Playwright specs (they need a built binary and real daemons). Those three were judged by reading them and their helper, and by confirming both new waits are server-driven rather than client-optimistic.

CodeRabbit's only line-level finding is the `acp.rs:1328` one from 08:07, which predates this commit and is exactly what it fixes. No new bot findings at this head.

## Merge risk

**Low.** The remaining items are two misplaced doc comments and one unreachable ternary branch in a test helper; none can produce a wrong runtime answer. Reverts cleanly, no schema or persisted state. The residual worth naming is the coverage gap rather than a defect: if someone later removes or reorders the re-send at `acp.rs:1440`, nothing fails, and it would arrive as "Stop doesn't work sometimes" from a user, which is how both this bug and #3417 surfaced. The `Drop` comment and the `s5` test block are decent mitigations, and better than nothing.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Proportionate, two misplaced blocks to fix and one optional note in `docs/development/server-owned-prompt-dispatch.md`
**Merge risk:** Low, dominated by the untested handler-level re-send rather than by any defect

**Recommend merging.** The author took the blocker seriously and answered it with a better fix than the one proposed here: inverting the default so a cancel is never withheld, and demoting the latch to a re-send, makes losing a Stop structurally impossible instead of contingent on enumerating early returns. The queued-path scenario was verified end to end against the new code, the re-send was confirmed to cover the only path that needs it, and the doubled cancel was confirmed inert on all three fronts it could have bitten. Diagnosing this bug from CI traces on a tree that had not regressed, then taking the review correction and generalizing it rather than patching the one reported case, is good work.

Non-blocking follow-ups, in priority order: the handler-level test at `acp.rs:3376`'s recipe (highest value, since the mutation shows nothing guards the re-send); the two doc comment moves at `supervisor.rs:2767` and `5236`; the `aoeServe.ts:211` dead branch; `lock_recover` instead of `.expect()` on the three new locks, matching the 25 other sites and the sibling reservation maps; and a line in `server-owned-prompt-dispatch.md` recording the reservation invariant.

**LGTM** — the blocking regression is closed structurally rather than patched, all checks are green at `a5e93c30`, and everything left is cosmetic plus one test worth having but not worth holding the PR for.

jerome-benoit (changes requested):
**[P2] Make prompt/cancel ordering a shared delivery invariant**

The new reservation only covers the ordinary composer handler. `acp_prompt_diff_comments` publishes `UserDiffCommentsPrompt` and then calls `send_prompt` without a reservation, while plugin turns, pending initial turns, and queue drains enter through `SessionService::send_turn`, which likewise publishes the turn-active event before forwarding the prompt.

A connected UI can therefore expose Stop and enqueue Cancel before Prompt reaches the ACP command channel on those paths, with no post-dispatch re-send. Please enforce prompt/cancel ordering in the shared delivery layer while retaining the earlier reservation needed for the optimistic HTTP composer path, and add coverage for at least one non-composer producer.

I am not repeating the overlapping-dispatch finding already posted by CodeRabbit at this head.

njbrake (commented):

_Reviewed at `3e83119d`. Push new commits and it will be re-reviewed._

---

_Re-reviewed at `3e83119d`, rebased onto `origin/main` (clean). Judged specifically against @jerome-benoit's standing CHANGES_REQUESTED. Push new commits and it will be re-reviewed._

---

## Summary

Since `a5e93c30` the branch has gained one commit, `3e83119d`, which moves the prompt-dispatch reservation from the composer handler alone to three places: `acp_prompt` (outer), `SessionService::send_turn` (inner, shared by four producers), and `acp_prompt_diff_comments`. The slot became re-entrant and refcounted so the composer's outer reservation nests with `send_turn`'s inner one instead of evicting it. `acp_cancel` still forwards every cancel unconditionally; the latch is purely an additive re-send after dispatch.

## Verdict on @jerome-benoit's P2, up front

**Half resolved. Do not dismiss the review yet.**

The producer enumeration ask is met. The coverage ask is not met, and the shared-layer ask surfaced a new hole of the same family.

| Producer | Reaches channel via | Reservation now? | Where |
| --- | --- | --- | --- |
| Composer `POST /acp/prompt` | `send_turn` | Yes, outer + inner | `api/acp.rs:1328`, `session_service.rs:489` |
| `acp_prompt_diff_comments` | `supervisor.send_prompt` directly | Yes | `api/acp.rs:1538` |
| Plugin turn | `session_api.rs:561` then `send_turn` | Yes, inner | `session_service.rs:489` |
| Pending initial turn | `session_service.rs:707` then `send_turn` | Yes, inner | same |
| Queue drain | `session_service.rs:1207` then `send_turn` | Yes, inner | same |

I traced this myself rather than taking the list: `grep '\.send_prompt('` gives exactly two production call sites (`session_service.rs:583` inside `send_turn`, `api/acp.rs:1555` in diff-comments), and `send_turn(` gives exactly four production callers. Every one is now covered. **No producer publishes turn-active before the prompt reaches the channel without a reservation open.** That part of jerome's finding is genuinely fixed, and the fix is a real refactor, not a copy-paste to a second place.

## [P1] Two overlapping dispatches on one session still lose the cancel

This is CodeRabbit's overlapping-dispatch finding, which jerome deliberately did not repeat. **It stands.** The refcount change fixed slot *eviction* and replaced it with slot *sharing*, which reproduces the identical symptom on the identical sequence CodeRabbit described.

`PromptDispatchGuard::release` returns `None` whenever `refs > 0` after decrement, and `take_pending_cancel` is `release().unwrap_or(false)`. That is correct for a nesting relation (outer outlives inner). It is wrong for a *sibling* relation, and two concurrent `POST /acp/prompt` on the same session are siblings, not nested.

I reproduced it as a unit test in the PR's own tree:

```rust
let a = sup.begin_prompt_dispatch("sib");   // composer send
let b = sup.begin_prompt_dispatch("sib");   // second send, will be queued
assert!(sup.cancel_during_prompt_dispatch("sib"));  // user presses Stop
let a_owes = a.take_pending_cancel();       // A actually dispatched its prompt
drop(b);                                    // B returns Queued, no dispatch
assert!(a_owes);
```

```
thread 'acp::supervisor::tests::scratch_sibling_dispatches_lose_the_cancel' panicked at
src/acp/supervisor.rs:5290:9:
the request that ACTUALLY dispatched a prompt must owe the re-send
```

A gets `false` because B still holds a ref; B then drops, `release()` returns `Some(true)`, and `Drop` throws it away. The prompt is on the wire, the cancel is not re-sent, the agent cleared it on turn start. That is verbatim the bug this PR exists to fix.

Reachability is not theoretical: send a prompt, immediately queue a follow-up, press Stop, all inside prompt 1's dispatch window (which spans dormant wake, `trigger_resume_background`, and `wait_until_ready`, so it is seconds wide on a cold session, not the 78ms of the CI trace). `queue-follow-up-with-nav` already drives the send-then-queue half of that sequence.

Severity is bounded: `acp_cancel` still forwards immediately, so this is a *failure to improve* on the pre-PR behavior for that interleaving, not a regression versus `main`. But it is precisely the case jerome's "shared delivery invariant" was meant to close, and it is the one CodeRabbit named.

Shape of a fix: the latch has to belong to whichever holder actually dispatched, not to whichever one happens to release last. Either track `cancel_requested` per guard rather than per slot, or serialize dispatch per session so siblings cannot overlap. The current `generation` field cannot help, see below.

## [P2] Neither reservation is pinned by any test, and the non-composer coverage ask is unmet

I mutation-tested each reservation independently by pointing it at a session id that never matches.

**Mutant 1, `SessionService::send_turn`'s reservation removed** (`begin_prompt_dispatch(id)` becomes `begin_prompt_dispatch("__mutant__")`):

```
test result: ok. 6452 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 88.88s
```

The entire lib suite stays green. **There is no coverage of any non-composer producer.** jerome asked for "at least one"; the commit does not add one.

The commit *appears* to, via this case in `a_cancel_during_prompt_dispatch_is_resent_after_the_prompt`:

```rust
// A non-composer producer (plugin turn, pending initial turn, queue
// drain) reaches `send_turn` with no outer holder, so its reservation
// is the outermost and it owes the re-send itself.
let only = sup.begin_prompt_dispatch("s7");
assert!(sup.cancel_during_prompt_dispatch("s7"));
assert!(only.take_pending_cancel());
```

That is assertion-for-assertion identical to the `s1` case fifteen lines above it. It never calls `send_turn`, so it cannot observe whether `send_turn` takes a reservation. It is a relabeled duplicate table row, and under the house "one behavior per test, permutations in a table" rule it should be deleted rather than kept: the label is doing the work the assertions do not.

**Mutant 2, `acp_prompt`'s outer reservation removed** (`api/acp.rs:1328`):

```
test result: ok. 2 passed; 0 failed
```

Also green, including the new handler-level test.

**Mutant 3, both removed:**

```
---- server::api::acp::tests::cancel_during_the_prompt_window_is_latched_and_still_forwarded stdout ----
panicked at src/server/api/acp.rs:3438:9:
the prompt handler must hold a reservation while it works
```

So `cancel_during_the_prompt_window_is_latched_and_still_forwarded` proves only that *some* reservation is open at the 300ms mark, not which. When `acp_prompt`'s outer one is removed, `send_turn`'s inner one satisfies the assertion; when `send_turn`'s is removed, `acp_prompt`'s does. The test does not pin the invariant it is named for.

What would actually meet the ask: drive `SessionService::send_turn` for a plugin or drain caller, assert `cancel_during_prompt_dispatch` returns `true` mid-flight, and assert the re-send fires. That is a Rust unit/integration test, the cheap tier, and it is the one thing standing between this and jerome dismissing his review.

`acp_prompt_diff_comments`'s reservation has no coverage at all.

## [P3] CodeRabbit's newest finding on `acp_prompt_diff_comments` is factually right, severity overstated

CodeRabbit at this head flags that `begin_prompt_dispatch` sits at line 1538, after `touch_on_prompt_and_wake_if_sunk` (1499) and after the `woke_idle_dormant` / `trigger_resume_background` block (1509-1533), whereas `acp_prompt` reserves at 1328 *before* its equivalent work and `send_turn` reserves at 489 before `trigger_resume_background`. The line numbers check out.

I disagree with "Major". The exposure that makes the composer window dangerous is the client's optimistic `running` flag, and `SendCommentsDialog.send` has no equivalent: it `await`s the fetch inside a modal, calls `onSent()` only on 2xx, and blocks Esc while in flight. Nothing publishes turn-active before line 1538 in this handler either. So there is no path I can find where Stop is exposed during the un-reserved window for *this* producer specifically.

It is still worth taking as a consistency fix. Moving one line costs nothing and removes the "why is this one different" question for the next reader; leaving it means the invariant reads as three ad-hoc placements rather than one rule.

## [P3] `DispatchSlot::generation` is now unreachable, and the PR body still justifies it

The `generation != self.generation` early return in `release()` cannot fire under the current API:

- A guard holds a ref for its whole life and releases exactly once (`settled` blocks the double-release through `Drop`).
- The slot is removed only when `refs` hits 0.
- Therefore while any guard is unreleased, `refs >= 1` and the slot still carries that guard's generation.
- `prompt_dispatches` is mutated in exactly three places (`begin_prompt_dispatch`, `cancel_during_prompt_dispatch`, `release`), none of which can recreate the slot underneath a live guard.

Under AGENTS.md's "no dead code, if a field isn't used yet don't add it" this should go. It is also why the `s4` case in the supervisor test is mislabeled: `first` and `second` are two refs on one slot with the same generation, so "a stale guard must not clear the reservation a later prompt made" describes the `55cb17d3` overwrite design, not this one. The case tests refcounting.

The deeper point is that the generation was the mechanism protecting overlapping dispatches in the original design, and it was removed in substance without anything replacing it. That is the P1 above.

## [P2] The PR body describes a design that is no longer what is pushed

Three claims in the description are stale, all from the first head, and a reviewer reading the body before the diff will form the wrong model:

1. > **Deliberately not applied to `acp_prompt_diff_comments`.** ... I had it there in a first pass and pulled it back out.

   The head commit applies it there (`api/acp.rs:1538`). This is now the exact opposite of what the code does.

2. > A cancel arriving inside it latches onto the reservation **instead of being forwarded**

   Since `a5e93c30`, `acp_cancel` always forwards and additionally latches. The deferral design this sentence describes was abandoned, correctly.

3. > **Generation counter**, so a stale guard cannot clear a slot a later prompt installed.

   See above; refcounting replaced it and the generation check is now unreachable.

The Test Coverage Analysis section also names `a_cancel_during_prompt_dispatch_is_deferred_to_the_dispatch`, renamed to `..._is_resent_after_the_prompt`. Worth a body pass before merge; the diff has moved a long way from what it says.

## Interleavings and leak analysis

Asked to reason about this rather than line-read it, so, explicitly:

- **Cancel between reservation and dispatch.** Latched *and* forwarded. If the forward loses the race, the re-send after dispatch fixes the ordering. This is the intended path and it works, for the single-dispatch case.
- **Dispatch fails, or the handler takes an early return.** `Drop` releases; the latch is discarded; the cancel already went out at `acp_cancel` time, so nothing is lost. The `a5e93c30` change is what makes a "leaked" latch harmless, and that is the right call. Well reasoned in the code comments.
- **Task panic or tokio abort.** `Drop` runs during unwind and on abort. There is no `mem::forget` or leak path, so a permanently-open slot is not reachable. **No wedge risk, and no timeout or drop guard is needed.** A reservation that leaks would not stick the composer, because cancels never depend on it.
- **Duplicate cancel.** Deliberate and documented. Worst case is one extra `session/cancel` inside the window plus a synthetic `Stopped`, which the no-turn branch already publishes on `main`. Correct trade.
- **Late re-send hitting the next turn.** `forward_deferred_cancel` runs immediately after `send_prompt` returns, in the same handler, so turn A cannot realistically have ended and a queued turn B started in between. Theoretically possible under a stalled await; not worth guarding.
- **Poisoned mutex.** Uses the existing `lock_recover` helper. Consistent.

The RAII shape is genuinely good. The failure mode of the design is not leaks, it is the sibling case in P1.

## Is the invariant shared, or enforced N times?

Enforced three times, and a future producer can silently opt out.

`SessionService::send_turn` genuinely is a shared layer, and putting the reservation there covers four of the five producers with one edit. That is a real improvement over the composer-only version and is what jerome asked for in spirit. But the actual bottom of the stack, `Supervisor::send_prompt` (`supervisor.rs:2697`), has no reservation, no `debug_assert`, and no type-level requirement that a guard be held. `acp_prompt_diff_comments` calls it directly today, and the next caller that does the same gets no warning.

I do not think that is easily fixable at `send_prompt`: the window that matters starts at the *publish*, not the send, so enforcement inside `send_prompt` would cover the wrong span. The honest structural answer is a type that owns publish-then-forward as one operation, which is out of scope for a flake fix. What is in scope and cheap: a `debug_assert!` in `send_prompt` that a slot exists for the session, which turns "the next producer silently opts out" into "the next producer's test fails". Worth considering, given the file's precedent for exactly this kind of belt-and-braces guard (`supervisor.rs:2565`, the "never fires today" respawn-inbound branch).

## The optimistic composer path

Retained and unchanged, as jerome asked. `acp_prompt` still reserves at the very top of the handler (before `touch_on_prompt_and_wake_if_sunk`, before the instance lock, before the control-state fold), and `forward_deferred_cancel` on the `Ok(())` arm releases that outer level. The `202`-then-work behavior, the queued disposition, and the response shape are all byte-identical to before. No frontend behavior changed.

## The three live specs

These are the good half of the PR. None of the three widened a timeout; all three added a *precondition wait on a server-confirmed signal* before the action that was racing:

| Spec | Added gate | Signal source |
| --- | --- | --- |
| `composer-stop-during-thinking` | `[data-testid="acp-working-spinner"][data-thinking="true"]` | reducer's `ThinkingStarted` (`acpTypes.ts:1379`) |
| `composer-stop-during-sub-agent` | `Task: investigate` row visible | agent's own `tool_call` update |
| `queue-follow-up-with-nav` | `waitForQueuedPrompts(...)` | daemon's `GET /api/sessions/{id}/queue` |

`data-thinking` is a genuine distinction: the spinner itself renders on the optimistic `running` flag, so `acp-working-spinner` alone cannot tell client-optimism from daemon-truth. I verified `state.thinking` traces back to the `ThinkingStarted` reducer branch, so the comment is accurate. `waitForQueuedPrompts` matches the existing `listSessions` idiom in the same file (bare fetch, throw on `!res.ok`, throw on shape mismatch), and I confirmed against Playwright's own source that a throw from the `expect.poll` callback propagates immediately rather than burning the timeout (`const value = await actual()` sits outside the try/catch at `matchers/expect.js:13388`), so the fail-fast comment is correct too.

One consequence nobody has raised: the two composer-stop specs now deliberately wait *past* the window the server fix addresses. That is right for a user story, but it means that after this PR **no test at any tier exercises the HTTP-level ordering race**. The PR body is upfront that the race does not reproduce locally, so this is an accepted position rather than an oversight, but it is worth stating that the server half of this PR ships with only structural coverage of the primitive, and that structural coverage does not pin two of the three call sites (see P2).

## Design alignment

**Aligned.** Compared against the existing per-session reservation machinery in the same file: `pending_resumes` / `begin_resume` / `ResumeReservationOutcome` (`supervisor.rs:353`) and the `cancelled_resumes` breadcrumb set. `DispatchSlot` sits directly alongside them, keyed the same way, guarded by the same `lock_recover` helper, released through the same RAII shape. No parallel mechanism, no new abstraction with one caller, no layering inversion: the guard type lives in the supervisor and the handlers depend downward on it. `SessionService::send_turn` was already the shared publish-then-forward chokepoint for four producers, and the PR extends it rather than routing around it. The doc section lands in the existing `docs/development/server-owned-prompt-dispatch.md` rather than a new file.

| Check | Result |
| --- | --- |
| `cargo test --features serve` (full, not just `--lib`) | 6452 + all integration targets, 0 failed |
| `cargo clippy --features serve --all-targets` | 5 warnings, all pre-existing (`supervisor.rs:4181`, `smart_rename.rs:1546`, `background_agent.rs:627/745`, `process/macos.rs:266`), none in this PR's added regions |
| `cd web && npm run format:check` | clean, 926 files |
| `cd web && npm run lint` | clean |
| Vitest | 353 files, 3713 tests, 0 failed, no type errors |
| `node tests/validate-coverage-matrix.mjs` | passed |

**Coverage matrix:** correctly unchanged. `StructuredView.tsx` already appears in six entries, the validator's requirement is that every component file appear at least once, and no new component or dashboard flow was added. The author's "no matrix change needed" call is right.

Conventional commit prefix on all four commits. No emdashes or `--` separators in any added prose (checked mechanically over the added lines). No `#[allow(dead_code)]`.

## Issues found

One real defect beyond the review findings above:

**Misplaced doc comment, `src/acp/supervisor.rs:5267-5290`.** The new test was inserted between `drain_skips_restart_when_stopped_rate_limited`'s doc comment and its `#[tokio::test]`. The result is that the #1281 rate-limit-drain doc comment ("...does NOT emit a synthetic AgentStartupError, which would flip the sidebar to Error") is now the *opening* of the new cancel test's doc comment, and the drain test is left undocumented:

```rust
    /// handle so the next `/acp/spawn` or `/acp/switch-agent`
    /// doesn't hit AlreadyRunning, and does NOT emit a synthetic
    /// AgentStartupError (which would flip the sidebar to Error).
    /// A Stop pressed while `acp_prompt` is still working latches onto the
    /// pending dispatch, so the dispatch re-sends it once the prompt has
    /// reached the agent. ...
    #[test]
    fn a_cancel_during_prompt_dispatch_is_resent_after_the_prompt() {
```

One-line fix: move the three new `///` lines above a blank line and restore the original comment to the drain test.

## Documentation

**Proportionate, and the new section is earned.** The 34 lines added to `docs/development/server-owned-prompt-dispatch.md` document a non-obvious ordering invariant with two load-bearing rules and an explicit warning about the trap (`PromptDispatchGuard::drop` discards the latch, safe only because `acp_cancel` already forwarded). That is exactly the "why invisible from the code" case worth committing. I verified the one checkable factual claim in it: "three of the four `QueueReason` values mean a turn is already running" is correct against `src/acp/dispatch.rs:31` (`TurnActive`, `Cancelling`, `Compacting` yes; `WorkerDown` no).

Nothing was made stale that I could find, with one exception: the doc's rule 1 says "**Every producer** reserves the window", which is true for producers but overstates the guarantee given the sibling hole in P1. If the P1 fix lands, the sentence becomes accurate; if it is deferred, the doc should say the reservation is per-session and describe what happens when two are open.

## Proactive fixes

Not staged. The doc-comment misplacement is a one-line fix I would normally stage, but the P1 sibling hole needs a design decision from the author (per-guard latch versus serialized dispatch) rather than a patch written by a reviewer under time pressure, and shipping the cosmetic fix without it would just add a commit to a branch that needs another one anyway.

## Merge risk

**Moderate.**

- **Blast radius.** A cancel silently doing nothing, leaving the composer wedged with a turn running to completion. Loud to the user, invisible to the daemon: nothing logs, nothing alerts. We hear about it from a user or from a CI flake, which is exactly how this PR started.
- **Reversibility.** Clean `git revert`. No migration, no schema, no persisted state, no released artifact.
- **Coverage.** Named specifically: `SessionService::send_turn`'s reservation has zero test coverage (proven by mutation), `acp_prompt`'s outer reservation has zero (proven by mutation), `acp_prompt_diff_comments`'s has zero, and the sibling-dispatch interleaving has zero. The handler test that looks like it covers the first two covers neither individually.
- **Most likely way this bites us.** Somebody refactors `send_turn` in six months, drops or moves the `begin_prompt_dispatch` line, and every test stays green. Nothing surfaces it: no assertion, no debug assert, no log that fires on absence. The next signal is a flaky live spec, which is precisely the loop this PR is trying to exit.

Net versus `main`: this is an improvement on every path it covers and a regression on none. The residual is that it stops one step short of the invariant it claims, and ships without the coverage that would keep the invariant from silently eroding.

## Verdict

**Design alignment:** Aligned
**Project fit:** Accept
**Merge risk:** Moderate; the ordering invariant is unpinned by any test, so it can silently erode

The producer enumeration in @jerome-benoit's P2 is genuinely resolved, @Seluj78, and the `send_turn` placement is the right shared-layer answer rather than a copy-paste. That was the hard part and you got it right.

But his second ask, coverage for at least one non-composer producer, is not met: I removed `send_turn`'s reservation entirely and all 6452 tests stayed green, and the test case that claims to cover a non-composer producer is an assertion-identical duplicate of the case above it. Separately, the refcount rewrite that made the reservation shareable reintroduced CodeRabbit's overlapping-dispatch finding in a new shape, which I reproduced as a failing test in the PR's own tree. Both are small, well-understood fixes on a change that is otherwise good work and should land.

**@jerome-benoit, the enumeration half of your P2 is fixed; the coverage half is not, so the CHANGES_REQUESTED is still doing useful work.**

Prominent body/diff mismatch, repeated here because it is easy to miss: the PR description still says the reservation is **deliberately not applied** to `acp_prompt_diff_comments` and that a cancel latches **instead of being forwarded**. The head commit does the opposite on both counts.

src/server/session_service.rs
```diff
@@ -580,7 +589,27 @@ impl SessionService {
             }
         };
         match outcome {
-            Ok(()) => Ok(()),
+            Ok(()) => {
+                // Prompt is away. If a Stop landed while it was in dispatch and
+                // no outer holder will handle it (a non-composer producer), the
+                // agent already dropped that cancel on turn start; re-send it.
+                if dispatch_guard.take_pending_cancel() {
```
jerome-benoit: **[P2] Keep the cancel owed until the resend is durably queued**

`take_pending_cancel()` removes the per-session slot before `cancel_prompt().await`. That await ultimately uses `mpsc::Sender::send`; Tokio 1.53.1 documents cancellation as dropping the unsent message, and plugin shutdown can abort the task executing `sessions.turn.send` (`plugin/host.rs:350`, `plugin/session_api.rs:560-571`). In that interleaving, the initial cancel may precede the prompt, the prompt is then accepted, the latch is removed, and aborting the resend recreates the lost Stop. Please retain or transfer this debt until the cancel is successfully enqueued, and add a saturated-channel/abort regression test.

docs/development/server-owned-prompt-dispatch.md
```diff
@@ -130,6 +130,40 @@ lock over the POST round trip, not as a term in any dispatch decision. And the
 wake PATCH for an archived or snoozed session (#1581) stays, because it is a
 session-lifecycle action the user takes before the prompt exists.
 
+## Cancel ordering across the dispatch window
+
+Dispatch is not instantaneous, and the turn is advertised as active before the
+prompt reaches the agent: `SessionService::send_turn` publishes the user prompt
+and only then forwards it, and `acp_prompt` answers 202 before doing its
```
jerome-benoit: **[P3] Correct the 202 timing description**

`acp_prompt` does not return 202 before its pre-dispatch work: it reserves at `api/acp.rs:1328`, awaits dispatch and `send_turn` through `:1429`, and only constructs the response at `:1442`. The race exists because the client exposes optimistic running state before the POST completes. Please document that actual ordering.

jerome-benoit (changes requested):
Reviewed at exact head `3e83119d6836718c481497e946cabbba031fa9b7`. Existing discussions already cover the sibling-dispatch ownership bug, the missing producer-level coverage, and the late diff-comments reservation, so I have not duplicated them. One additional blocking cancellation window remains: the resend debt is consumed before an abortable asynchronous enqueue. The documentation timing correction is non-blocking. CI is green at this SHA, but it does not exercise the cancellation window described inline.

src/server/api/acp.rs
```diff
@@ -1616,13 +1616,58 @@ pub async fn acp_attachment(
...
+/// and forwarding anyway.
+///
+/// Sized for the case that matters rather than the worst case. With a live
+/// worker, a submission holds the guard for a decide plus a channel enqueue,
+/// which is milliseconds, so Stop almost never waits at all. The long holds
+/// belong to a cold or dormant worker, where the submission parks on
+/// `WORKER_READY_TIMEOUT` waiting for a spawn; there is no turn running then,
+/// so there is nothing for the cancel to order itself behind and no reason to
+/// make the user watch a dead Stop button for ten seconds.
+const CANCEL_SUBMISSION_WAIT: std::time::Duration = std::time::Duration::from_secs(2);
```
jerome-benoit, line 1563: **[P1] Preserve ordering when worker startup exceeds two seconds**

This timeout reopens the exact race the patch is meant to close. A prompt submission keeps this guard while `send_turn` waits for worker readiness, whose normal ACP handshake is documented as 2–3 seconds and whose actual timeout is 10 seconds (`src/acp/supervisor.rs:54,2550-2553`; `src/server/session_service.rs:533-554`). Once this two-second timeout expires, `acp_cancel` and the prompt both wait on `ready_client`; `notify_waiters` wakes them without preserving prompt-before-cancel order. Cancel can therefore reach the newly started worker first and be cleared when Prompt opens the turn. Please keep a cancellation debt or otherwise preserve ordering until the accepted submission is either delivered or fails, and cover the >2 s startup path with an assertion on the actual `prompt`, `cancel` command order.

jerome-benoit (changes requested):
Reviewed at exact head `163c14c19dd0978cb4888c88d9163977d34d5d80`. CodeRabbit already covers the pre-guard wake window, so I have not duplicated it. One separate blocking window remains in the timeout fallback: after two seconds the cancel is forwarded without ordering even though worker readiness is documented as normally taking 2–3 seconds and is allowed to wait for ten. CI is green at this SHA, but the added test releases the guard after 500 ms and therefore does not cover this fallback.

src/server/api/acp.rs
```diff
@@ -1605,6 +1617,39 @@ pub async fn acp_cancel(
...
+    // reservation of its own: every turn-starting surface already decides and
+    // dispatches under this one guard (#3639, #3673), so waiting for it means
+    // waiting for the prompt to be away, on every producer at once. The guard
+    // is held across the forward so a new submission cannot start in the gap.
+    //
+    // Best-effort by design: on timeout, or for an unknown session, forward
+    // unguarded, which is exactly the pre-existing behaviour and never worse.
+    let _submission = tokio::time::timeout(
+        CANCEL_SUBMISSION_WAIT,
+        state.session_service.prompt_submission_for_session(&id),
```
jerome-benoit, line 1642: [P2] Remove the second acquisition of the submission guard

The merged handler already holds this session's `OwnedMutexGuard` from lines 1606–1609. Shadowing `_submission` does not drop that first binding, and `prompt_submission_for_session` acquires the same non-reentrant mutex. For an admitted session, this second acquisition therefore waits until the two-second timeout even when no prompt is in flight. Every normal Stop is delayed, and the `forwarding unordered` log is misleading because the first guard still protects the forward. Keep the single acquisition already provided by main and remove this redundant wait; retain coverage that a cancel with an available guard forwards promptly. This is established by source tracing, not a local runtime reproduction.

jerome-benoit (changes requested):
Re-reviewed `52de566ad43bfc0ab1a5d58edcbb7f64e41cba54` against current main and the existing discussions.

Main already provides the submission ordering this PR intended to add. My previous startup-timeout P1 no longer applies: the first guard remains held through cancellation. The old wake/reservation findings likewise describe superseded code.

One introduced P2 blocks approval, detailed inline: the merged handler acquires the same guard twice and adds two seconds to a normal Stop. Remove the second wait and its obsolete constant/documentation, preserving main's first guard. Closing this PR as superseded is also reasonable.

The existing documentation issue remains P3: `acp_prompt` awaits `send_turn` before constructing its 202 response (`acp.rs:1333-1350`); optimistic client state opens the original race window.

Two independent sequential analyses agreed on the double acquisition. I classify it as P2 because the extra delay is bounded; no permanent deadlock or lost cancellation is established. Static review only, no local tests run. The failed CI jobs were rerun after the unchanged terminal test reported `docker: Text file busy`; bare-core was still running at the final check.

jerome-benoit:
Superseded by #3844 on main; see the disposition comment.

Issue #3337, opened 2026-08-12
Regression (#3335): an absent tmux socket makes batch_pane_metadata() return Err, so poll_statuses_once() skips every refresh and two unit tests fail


jerome-benoit:
## Summary

#3335 made `batch_pane_metadata() == Err` load-bearing: `poll_statuses_once()` now early-returns `Vec::new()` on `Err` (`src/tui/status_poller.rs:154`) instead of the previous `unwrap_or_default()`. But `batch_pane_metadata()` returns `Err` in a case that is actually authoritative "no panes": when the tmux **socket file is absent**. The classifier `tmux_no_server_running()` only matches the string `"no server running"` (`src/tmux/mod.rs:253`), and an absent socket produces a *different* message.

## Root cause

`tmux_no_server_running()` misses the socket-absent variant. Empirically, when the socket file does not exist, tmux exits non-zero with:

```
error connecting to <path> (No such file or directory)
```

for every socket mode (`-S <path>`, `-L <name>`, and the default socket). None of these contain `"no server running"` (which tmux emits only when the socket file *exists* but no server is listening). So `batch_pane_metadata()` takes the non-zero branch and returns `Err` (`src/tmux/mod.rs:631`).

Before #3335 this was harmless (`unwrap_or_default()` -> empty map -> poll continued). After #3335 the `Err` is treated as "skip this pass", so a socket-absent condition, which is really "zero panes", freezes the poller instead.

## Impact

1. **Runtime**: when the resolved tmux socket file is absent but sessions are tracked (e.g. server never started / socket swept from `/tmp` / server crash removing the socket), the status poller silently skips *every* refresh. Statuses freeze at their last value instead of the sessions being flagged. This partially re-opens the direction of #3327/#3328 for the socket-absent case.

2. **Tests**: two unit tests fail on a clean checkout whenever the resolved socket has no live server:
   - `tui::status_poller::tests::poll_statuses_once_never_emits_idle_intent_keep`
   - `tui::status_poller::tests::terminal_sandboxed_rows_still_get_a_container_error`

   Both call `poll_statuses_once` expecting an update; the early return yields zero. They pass only when a tmux server happens to exist on the resolved socket, so this is a hidden environment dependency, not a deterministic pass.

## Reproduction (verified)

Bisected on the same machine, same environment:

| Test | parent before #3335 | pristine `f2b81a62` (#3335) |
| --- | --- | --- |
[...]

Issue #3327, opened 2026-08-12
All sessions flip to Error (empty_capture) at once when a new session is spawned, under load


edbror:
## Summary

Spawning a **new session** while several sessions are already running causes **every other session to flip `Idle -> Error` simultaneously**, all with `pane=empty_capture`. The sessions then recover on their own over the next ~1–2 minutes. No agent process actually dies — it is a health-tracking false positive — but in the web dashboard it looks like a total outage (every row goes red at once), and with `notify_on_error = true` it fires a notification storm. Severity scales with system load and session count.

## Environment

- aoe **1.14.0** (confirmed latest via update check)
- macOS 26.5.2, arm64 (Apple Silicon)
- ~28 concurrent sessions (mostly `claude`, plus `opencode`, `codex`, `pi`, `antigravity`)
- tmux 3.7b, `focus-events on`

## Repro

1. Run aoe with a couple dozen sessions attached.
2. Open one new session (any agent — reproduced by opening a `codex` session).
3. Watch the other sessions: within ~1s they all go `Idle -> Error`.

## Evidence (debug.log)

```
06:27:22  [codex]  Starting -> Waiting        # new session aoe_Codex_98264364, vt channel armed, pid=3519
06:27:23  [claude] Idle -> Error   x28        # every other session, all within ~0.2s
          ...all with pane=empty_capture
06:27:44+ [*]      Error -> Idle              # recoveries begin, vt channels re-arm
06:29:xx  back to normal (Running/Idle)
```

Same signature (`pane=empty_capture` → mass `Idle -> Error`) reproduces on **any** new-session spawn, and also appeared transiently when the daemon restarted.

## The processes do NOT die

During the storm, `tmux list-panes -a -F '#{pane_dead}'` reports **`dead=0` for every pane** — the underlying agent processes are alive the whole time. There are **no** `pane dead` / `SIGKILL` / respawn entries in the log for the affected sessions. It is purely aoe's capture-based status tracker briefly reading every pane as empty and defaulting them to `Error`.

## Load correlation

The red-out is worse and slower to recover under load. Observed load average of ~24 during the dramatic events vs ~12 during milder ones. With ~28 sessions, spawning one appears to trigger a burst of `capture-pane` across all panes; under load enough of them return empty at once to flip the whole set to `Error`.

## What did NOT help

- `set -g focus-events on` in `~/.tmux.conf` (aoe recommends it) — enabled, no effect on the storm.
[...]

PR #3540, opened 2026-08-26
fix(tmux): preserve UTF-8 session names

### LINKED ISSUE #3536: 1.15.0: sessions with non-ASCII (CJK) titles are unreachable — attach targets a width-substituted name, `exists=true` but exits 1
## Summary

On 1.15.0, sessions whose **title contains non-ASCII characters** (CJK — Korean in my case) can no longer be reached. TUI attach fails, and `aoe session capture <id>` returns empty. Sessions with ASCII-only titles on the same machine work fine.

The tmux session itself is healthy the whole time — `tmux ls` lists it, the agent inside keeps running, and attaching by hand with `tmux attach -t <real name>` works. Only aoe cannot reach it.

## Environment

- aoe 1.15.0 (Homebrew), macOS 15.7.1 arm64, tmux 3.7c
- `LANG=ko_KR.UTF-8`, `LC_ALL` unset (the tmux server inherited the same `LANG`)

## Reproduction

1. Create a session with a non-ASCII title, e.g. `QA 협업 정리`.
2. `tmux ls` → the tmux session is named `aoe_QA_협업_정리_9f4e4c6a` (Hangul preserved).
3. `aoe session capture 9f4e4c6a` → **empty output**.
4. Try to attach from the TUI → fails, and `debug.log` records:

```
WARN tui.input: tmux attach returned error: Failed to attach to tmux session
'aoe_QA___________9f4e4c6a' (exit 1): exists=true, pane_dead=false, tmux_info=
```

Note the name in that error: `aoe_QA___________9f4e4c6a`. That is **not** the name the session actually has. Every non-ASCII character has been replaced by underscores — and by *two* underscores per character, matching each character's display width (Hangul syllables are width 2, the spaces became one each).

## What is inconsistent

Two things stand out to me:

**1. The attach target does not match the created name.** The session is created with the Hangul intact but attached to by the width-substituted name. This is consistent across all my sessions — I have 6 live ones, and the only one that worked was the single ASCII-titled one (`ongoing`):

```
9f4e4c6a  QA 협업 정리    capture → empty
418b8685  Relux연동      capture → empty
b18d94a3  agent개선      capture → empty
32c5397d  thread 관리    capture → empty
dcaf9a30  결제정보정리     capture → empty
776105cf  ongoing       capture → OK      ← ASCII only
```

**2. `exists=true` while attach exits 1.** The existence probe accepts the substituted name but the actual `attach-session -t` with that same name fails, so the two paths disagree about what the session is called.

I could not pin down where the width-based substitution enters. What I did find while looking (may or may not be relevant):

- `Session::generate_name` (`src/tmux/session.rs:330`) builds the name via `sanitize_session_name`, which keeps `c.is_alphanumeric()`. That is the *Unicode* predicate, so Hangul passes through unchanged — which matches the name the tmux session actually gets. Nothing there produces underscores.
- There are two identical copies of `sanitize_session_name`, at `src/tmux/utils.rs:69` and `src/tmux/session.rs:1785`.
- `NameShape` / `agent_session_belongs_to` (`src/tmux/mod.rs:433`) match on the `_<id8>` suffix rather than the title, which may explain why the existence side is satisfied while the literal attach target is not.
- `tmux_name` is not persisted in `sessions.json`, so the name is re-derived on each use.

## Regression

This worked before. In my `debug.log` from 2026-07-23 on **1.13.2**, the same kind of Hangul-titled session was driven normally, with the Hangul intact in the target:

```
INFO agent_of_empires::tmux::vt: vt channel armed (pipe-pane -IO <-> vt100 grid)
target=aoe_소스코드_디렉토리_이관_0b86be62:^.0 cols=228 rows=55
```

The failures start after the 1.15.0 upgrade. I cannot say whether 1.14.1 was already affected — my log has no attach attempt in that window, though it does show two `start: resume with sid ... crashed pane within probe; preserving sid and marking resume failure` warnings on 1.14.1 that I have not investigated.

## Workaround

Renaming the tmux session to the width-substituted form makes it reachable again, immediately and without disturbing the agent running inside:

```bash
tmux rename-session -t 'aoe_QA_협업_정리_9f4e4c6a' 'aoe_QA___________9f4e4c6a'
```

The rule that worked for all five of my broken sessions: replace each non-ASCII character with `_` repeated by its display width (`unicodedata.east_asian_width` in `W`/`F` → 2, otherwise 1). After renaming all five, `aoe ps` also started reporting their PIDs, which had been blank.

Titling new sessions in ASCII avoids the problem entirely.

---

<sub>Written with Claude Code. The investigation, log excerpts, and the workaround are from my own machine and I verified them; the prose and the code-reading notes above are AI-drafted and reviewed by me.</sub>

atirna:
## Description

`tmux_query_command` used `LC_ALL=C` to keep tmux connection errors parseable. That also forced `LC_CTYPE=C`, so `list-sessions` replaced wide characters in CJK session names with underscores. The cached name then no longer matched the UTF-8 session name used by attach and capture.

Narrowing the override to `LC_MESSAGES=C` and removing inherited `LC_ALL` is not enough on its own: when `LC_CTYPE=C` overrides a UTF-8 `LANG`, or when `LC_ALL` was the caller's only UTF-8 locale, query output still width-substitutes. This adds tmux's global `-u` so query clients emit UTF-8 independently of those locales, while `LC_MESSAGES=C` keeps diagnostics stable English.

Fixes #3536

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## How to test

- Create a tmux session named `aoe_QA_협업_정리_9f4e4c6a` under a UTF-8 locale.
- `tmux -u list-sessions -F '#{session_name}'` with `LC_CTYPE=C` (or with `LC_ALL` unset and no other UTF-8 locale) returns the Hangul name intact.
- The same command without `-u` emits `aoe_QA___________9f4e4c6a`.

## Checklist

## Test Coverage Analysis

- [ ] Added or updated a Playwright spec under `web/tests/`

**TUI / CLI (e2e test)**
- [x] N/A: this PR changes tmux status-query locale handling, not a rendered TUI or CLI flow

`tmux_query_command_preserves_ctype_for_session_names` covers the command environment at the owning seam: `-u` is present, `LC_MESSAGES=C` stays explicit, and inherited `LC_ALL` is removed.

Verified locally:

- `cargo fmt --check`
- `cargo test tmux_query_command_preserves_ctype --lib`

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** gpt-5.6-terra via Codex

**Any Additional AI Details you'd like to share:** The locale behavior was reproduced against tmux with a CJK session name, and the resulting patch was independently reviewed.

## Summary

- Updated tmux queries to preserve UTF-8 session names.
- Set `LC_MESSAGES=C` and removed inherited `LC_ALL` to keep diagnostics stable.
- Added tmux’s global `-u` option for UTF-8 query output.
- Added session liveness snapshots to reduce repeated tmux lookups.
- Added regression tests for UTF-8 handling and liveness behavior.
- Re-exported `PaneGeom` for shared use.

## Benefits

- Prevents non-ASCII session names from becoming unreachable.
- Improves liveness checks.
- Keeps tmux output reliable for parsing.

## Potential enhancement

- Continue monitoring snapshot behavior across tmux server and session state changes.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -203,7 +203,10 @@ pub(crate) fn tmux_command() -> Command {
 /// removed so it cannot override that. Global `-u` forces UTF-8 session names
 /// even when the caller has `LC_CTYPE=C` or when `LC_ALL` was the only UTF-8
 /// locale source. Used by the status-query callers (which classify via
-/// [`tmux_no_server_running`]) and by `kill_session_if_present`.
+/// [`tmux_no_server_running`]) and by `kill_session_if_present`. Not folded into
+/// [`tmux_command`]: the interactive attach/switch-client/capture-pane paths must
+/// keep the user's locale, and `-u` would assert UTF-8 to a terminal that may not
+/// be.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
     cmd.arg("-u");
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -193,19 +193,20 @@ pub(crate) fn tmux_command() -> Command {
     cmd
 }
 
-/// Like [`tmux_command`], but pins `LC_ALL=C` so tmux's connection-failure
+/// Like [`tmux_command`], but pins `LC_MESSAGES=C` so tmux's connection-failure
 /// messages on stderr stay stable English for callers that match them. tmux's
 /// `client.c` prints `error connecting to <socket> (strerror(errno))` for a
 /// non-`ECONNREFUSED` connect failure, and glibc localizes `strerror` by
 /// `LC_MESSAGES`, so on a non-English host the `(No such file or directory)`
-/// ENOENT marker for an absent socket (#3337) would not match. Used by the
-/// status-query callers (which classify via [`tmux_no_server_running`]) and by
-/// `kill_session_if_present`. NOT folded into [`tmux_command`]: the
-/// interactive attach/switch-client/capture-pane paths must keep the user's
-/// locale for UTF-8 and status-bar rendering.
+/// ENOENT marker for an absent socket (#3337) would not match. Keeping the
+/// caller's `LC_CTYPE` matters because `list-sessions` returns names later
+/// passed to locale-preserving tmux commands; `LC_ALL=C` substitutes wide
+/// characters with underscores in those names. Used by the status-query
+/// callers (which classify via [`tmux_no_server_running`]) and by
+/// `kill_session_if_present`.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
-    cmd.env("LC_ALL", "C");
+    cmd.env("LC_MESSAGES", "C");
     cmd
 }
 
@@ -1463,6 +1464,22 @@ mod tests {
         assert_eq!(args.first().map(|a| a.to_str().unwrap()), Some("-S"));
         assert!(args.get(1).is_some(), "socket path arg present");
         assert_eq!(cmd.get_program().to_str(), Some("tmux"));
+    }
+
+    #[test]
+    fn tmux_query_command_preserves_ctype_for_session_names() {
+        let command = tmux_query_command();
+        let message_locale = command
+            .get_envs()
+            .find(|(key, _)| key.to_str() == Some("LC_MESSAGES"))
+            .and_then(|(_, value)| value.and_then(|value| value.to_str()));
+        assert_eq!(message_locale, Some("C"));
+        assert!(
+            command
+                .get_envs()
+                .all(|(key, _)| key.to_str() != Some("LC_ALL")),
+            "LC_ALL changes LC_CTYPE and corrupts non-ASCII tmux session names"
+        );
     }
 
     #[cfg(unix)]
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -206,6 +206,7 @@ pub(crate) fn tmux_command() -> Command {
 /// `kill_session_if_present`.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
+    cmd.env_remove("LC_ALL");
     cmd.env("LC_MESSAGES", "C");
     cmd
 }
@@ -1477,8 +1478,9 @@ mod tests {
         assert!(
             command
                 .get_envs()
-                .all(|(key, _)| key.to_str() != Some("LC_ALL")),
-            "LC_ALL changes LC_CTYPE and corrupts non-ASCII tmux session names"
+                .find(|(key, _)| key.to_str() == Some("LC_ALL"))
+                .is_some_and(|(_, value)| value.is_none()),
+            "LC_ALL must not override the caller's LC_CTYPE"
         );
     }
 
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -198,14 +198,14 @@ pub(crate) fn tmux_command() -> Command {
 /// `client.c` prints `error connecting to <socket> (strerror(errno))` for a
 /// non-`ECONNREFUSED` connect failure, and glibc localizes `strerror` by
 /// `LC_MESSAGES`, so on a non-English host the `(No such file or directory)`
-/// ENOENT marker for an absent socket (#3337) would not match. Keeping the
-/// caller's `LC_CTYPE` matters because `list-sessions` returns names later
-/// passed to locale-preserving tmux commands; `LC_ALL=C` substitutes wide
-/// characters with underscores in those names. Used by the status-query
-/// callers (which classify via [`tmux_no_server_running`]) and by
-/// `kill_session_if_present`.
+/// ENOENT marker for an absent socket (#3337) would not match. `LC_ALL` is
+/// removed so it cannot override that. Global `-u` forces UTF-8 session names
+/// even when the caller has `LC_CTYPE=C` or when `LC_ALL` was the only UTF-8
+/// locale source. Used by the status-query callers (which classify via
+/// [`tmux_no_server_running`]) and by `kill_session_if_present`.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
+    cmd.arg("-u");
     cmd.env_remove("LC_ALL");
     cmd.env("LC_MESSAGES", "C");
     cmd
@@ -1470,6 +1470,11 @@ mod tests {
     #[test]
     fn tmux_query_command_preserves_ctype_for_session_names() {
         let command = tmux_query_command();
+        let args: Vec<_> = command.get_args().map(|a| a.to_owned()).collect();
+        assert!(
+            args.iter().any(|a| a.to_str() == Some("-u")),
+            "tmux -u forces UTF-8 names independently of inherited LC_CTYPE"
+        );
         let message_locale = command
             .get_envs()
             .find(|(key, _)| key.to_str() == Some("LC_MESSAGES"))
@@ -1480,7 +1485,7 @@ mod tests {
                 .get_envs()
                 .find(|(key, _)| key.to_str() == Some("LC_ALL"))
                 .is_some_and(|(_, value)| value.is_none()),
-            "LC_ALL must not override the caller's LC_CTYPE"
+            "LC_ALL must not override LC_MESSAGES=C"
         );
     }
 
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -203,10 +203,10 @@ pub(crate) fn tmux_command() -> Command {
 /// removed so it cannot override that. Global `-u` forces UTF-8 session names
 /// even when the caller has `LC_CTYPE=C` or when `LC_ALL` was the only UTF-8
 /// locale source. Used by the status-query callers (which classify via
-/// [`tmux_no_server_running`]) and by `kill_session_if_present`. Not folded into
+/// [`tmux_no_server_running`]) and by `kill_session_if_present`. NOT folded into
 /// [`tmux_command`]: the interactive attach/switch-client/capture-pane paths must
-/// keep the user's locale, and `-u` would assert UTF-8 to a terminal that may not
-/// be.
+/// keep the user's locale for UTF-8 and status-bar rendering, and `-u` would
+/// assert UTF-8 to a terminal that may not be.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
     cmd.arg("-u");


src/tmux/mod.rs
```diff
@@ -193,19 +193,20 @@ pub(crate) fn tmux_command() -> Command {
...
+/// ENOENT marker for an absent socket (#3337) would not match. Keeping the
+/// caller's `LC_CTYPE` matters because `list-sessions` returns names later
+/// passed to locale-preserving tmux commands; `LC_ALL=C` substitutes wide
+/// characters with underscores in those names. Used by the status-query
+/// callers (which classify via [`tmux_no_server_running`]) and by
+/// `kill_session_if_present`.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
-    cmd.env("LC_ALL", "C");
+    cmd.env("LC_MESSAGES", "C");
```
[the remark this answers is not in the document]
atirna: reworked in 8fb5b64b: tmux_query_command now env_remove()s LC_ALL before setting LC_MESSAGES=C, so an inherited LC_ALL can no longer override the caller locale. LC_MESSAGES=C stays for the #3337 ENOENT matching.

confirmed live with tmux: with LC_ALL=C forced on the child, list-sessions returns a Hangul session name with every wide char substituted by underscores; with LC_ALL removed and only LC_MESSAGES=C set the name round-trips intact, so removing LC_ALL is what preserves the names.

the get_envs() test now asserts LC_ALL is explicitly removed (present with value None), not just absent from the explicit overrides, which was the gap in the first version.

src/tmux/mod.rs
```diff
@@ -193,19 +193,21 @@ pub(crate) fn tmux_command() -> Command {
...
+/// ENOENT marker for an absent socket (#3337) would not match. Keeping the
+/// caller's `LC_CTYPE` matters because `list-sessions` returns names later
+/// passed to locale-preserving tmux commands; `LC_ALL=C` substitutes wide
+/// characters with underscores in those names. Used by the status-query
+/// callers (which classify via [`tmux_no_server_running`]) and by
+/// `kill_session_if_present`.
 pub(crate) fn tmux_query_command() -> Command {
     let mut cmd = tmux_command();
-    cmd.env("LC_ALL", "C");
+    cmd.env_remove("LC_ALL");
```
jerome-benoit, line 213: **Force query clients to emit UTF-8 independently of the inherited character locale.**

Removing `LC_ALL` fixes the resolved precedence issue, but it still does not guarantee UTF-8 output. With tmux 3.7c, I created `aoe_QA_협업_정리_9f4e4c6a` under `LC_ALL=C.UTF-8`; the current query environment returned `aoe_QA___________9f4e4c6a` both when `LC_CTYPE=C` overrides a UTF-8 `LANG` and when `LC_ALL` was the caller's only UTF-8 locale source. Adding tmux's global `-u` flag returned the exact name in both cases.

Please add `-u` to `tmux_query_command()` while retaining `env_remove("LC_ALL")` and `LC_MESSAGES=C`, and assert the flag in the owning unit test. tmux has documented `-u` since at least 3.0a.

jerome-benoit (changes requested):
Thanks for addressing the inherited `LC_ALL` precedence issue. One remaining locale configuration still corrupts UTF-8 session names, so I'm requesting changes. See the inline reproduction and recommended `tmux -u` fix.

src/tmux/mod.rs
atirna: reworked in 97f5f06: tmux_query_command now passes -u, still env_remove()s LC_ALL and sets LC_MESSAGES=C.

reproduced both cases on tmux 3.6b with session aoe_QA_협업_정리_9f4e4c6a:

- LC_CTYPE=C over LANG=C.UTF-8, LC_MESSAGES=C, no -u → aoe_QA___________9f4e4c6a
- LC_ALL removed and no LANG/LC_CTYPE left (LC_ALL was the only UTF-8 source) → same substitution
- both with -u → aoe_QA_협업_정리_9f4e4c6a

the owning test now asserts get_args() contains -u.

njbrake (commented):

_Reviewed at `72ece483`. Push new commits and it will be re-reviewed._

---

## Summary

`tmux_query_command()` pinned `LC_ALL=C` so tmux's ENOENT connection error stayed parseable (#3337). That also cleared tmux's `CLIENT_UTF8` flag, so every query client ran its stdout and stderr through `utf8_sanitize()`, replacing each wide character with one `_` per display column. The cached name then diverged from the real session name, and `resolve_session_name` adopted the mangled one as the attach/capture target. This PR narrows the override to `LC_MESSAGES=C`, removes inherited `LC_ALL`, and adds tmux's global `-u` so query clients emit UTF-8 regardless of the caller's character locale.

First off: this is a genuinely well-diagnosed bug. Locale bugs are the ones where the repro depends on facts nobody writes down, and the PR body correctly identified that narrowing to `LC_MESSAGES=C` alone is *not* sufficient. Everything below was reproduced rather than read, and the reasoning holds up on every path that could be found.

## Design alignment

**Aligned.**

The codebase already has exactly one mechanism for this class of problem: the two-tier split between `tmux_command()` (`src/tmux/mod.rs:176`, interactive paths, user's locale) and `tmux_query_command()` (`src/tmux/mod.rs:207`, machine-parsed paths, pinned locale), introduced for #3327 and #3337. This change stays entirely inside the tier that already owns the concern. No parallel mechanism, no new abstraction, no caller-side locale handling sprinkled around. Every tmux invocation under `src/` was checked and all route through one of those two constructors, so the seam is still the single source of truth.

## What was verified, by reproducing

Created `aoe_test_QA_협업_정리_9f4e4c6a` on an isolated socket (tmux 3.7c, the reporter's version) and hex-dumped the round trip through the actual `tmux_query_command()` in a scratch test, on both main's implementation and this PR's.

| Environment | main (`LC_ALL=C`) | this PR (`-u`, `LC_MESSAGES=C`, no `LC_ALL`) |
| --- | --- | --- |
| `LANG=ko_KR.UTF-8`, `TMUX` unset (the reported setup) | `aoe_test_QA___________9f4e4c6a` | correct UTF-8 bytes |
| `LC_CTYPE=C` over a UTF-8 `LANG` | mangled | correct |
| bare `env -i` (minimal container) | mangled | correct |
| `TMUX` set | correct | correct |

The main-branch result is byte for byte the shape in #3536, 11 underscores and all. So the divergence is real, and this closes it.

**The mechanism, confirmed against tmux source, not inferred.** In `tmux.c:503-521`, `CLIENT_UTF8` is set if `TMUX` is set, else by substring-matching `"UTF-8"` / `"UTF8"` against the *first* non-empty of `LC_ALL`, `LC_CTYPE`, `LANG`. `LC_ALL=C` therefore short-circuits the whole sniff to a non-UTF-8 answer even when `LANG` is `ko_KR.UTF-8`. When `CLIENT_UTF8` is clear, `server-client.c:3204` sanitizes stdout and `cmd-queue.c:866-869` sanitizes stderr, both via `utf8_sanitize()` (`utf8.c:784`), which emits one `_` per display column. `-u` is handled at `tmux.c:477-480` as an unconditional `flags |= CLIENT_UTF8` set *before* the env sniff, so it is genuinely independent of every locale variable. That is the right lever, and it is the only one that covers the bare-environment case.

**The error parsing that `LC_ALL=C` existed to protect still works.** Both markers `tmux_no_server_running` matches (`src/tmux/mod.rs:328-341`) were checked under the new environment: `no server running on <sock>` and `error connecting to <sock> (No such file or directory)` both come through verbatim, including under `LANG=ko_KR.UTF-8`. `LC_MESSAGES` is the only category glibc localizes `strerror` by, and `env_remove("LC_ALL")` is what keeps it from being overridden, so the narrowing is complete for that purpose. Nothing else the query callers parse is locale-sensitive: the formats are `#{session_activity}`, `#{session_attached}`, `#{pane_index}`, `#{pane_dead}` and command names, all raw integers or ASCII, so dropping `LC_ALL` cannot leak `LC_NUMERIC` or `LC_TIME` formatting into a parsed field.

**Non-ASCII is not the only thing that has to survive.** ASCII names round-trip byte-identical before and after (md5 match against the literal), which is the regression nobody would have noticed.

**Removing `LC_ALL` does not risk a missing-locale startup failure.** tmux tries `setlocale(LC_CTYPE, "en_US.UTF-8")` then `"C.UTF-8"` before ever consulting the environment (`tmux.c:403-409`), and only `errx`es out if both fail *and* the environment resolves to a non-UTF-8 charmap. Dropping `LC_ALL=C` strictly widens what can satisfy that fallback, so if anything this makes the loud-failure case less likely, not more.

**Why no existing test caught this, which is worth knowing.** `TMUX` being set forces `CLIENT_UTF8` on unconditionally. Every tmux test in this repo runs under tmux, so the bug is structurally invisible to all of them. It is also why this only bites users who launch `aoe` from outside tmux, which fits the "worked on 1.13.2, broken on 1.15.0" report shape.

- `cargo fmt --check`: clean.
- `cargo test --features serve`: fully green, 6413 unit tests plus all integration suites, 0 failures (run with `env -u CLAUDE_CONFIG_DIR`).
- `cargo clippy --all-targets --features serve`: warnings are exactly the four known-red pre-existing spots (`acp/supervisor.rs`, `acp/background_agent.rs` twice, `session/smart_rename.rs`, `process/macos.rs`). Nothing new from this PR.
- Branch is level with `origin/main` (`5988e958`), 0 commits behind.
- One test for one behavior; not tmux-gated, so it actually runs rather than taking a skip branch (verified, 0.00s). No dead code. No emdashes in the added prose. Conventional commit prefix, correct scope.

## Automated review check-in

- **CodeRabbit, "Remove inherited `LC_ALL`" (major):** agree, and already addressed in `8fb5b64`. `env_remove` is the right call, and the bot is right that a `get_envs()` assertion on explicit overrides alone would not have detected inheritance.
- **CodeRabbit, "Out of Scope Changes" pre-merge check (`LiveSessionSnapshot`, liveness selection, `PaneGeom` re-export):** **disagree, false positive.** None of those are in this PR. They landed on `main` in #3508 (`adc053c2`) and only appear in CodeRabbit's walkthrough because the author merged `main` into the branch. The three-dot diff against `origin/main` is one file, two hunks, +31/-7. Nothing needs splitting out.
- **CodeRabbit, outside-diff note on `LiveSessionSnapshot::names()` not using `run_tmux_command_with_timeout` (`mod.rs:637-640`):** correct observation, but it is main's code from #3508 and out of scope here.

## Issues found

None blocking. Everything below is a suggestion.

## Documentation

The diff is proportionate, but one sentence was dropped that is worth asking back. The old docstring ended:

> NOT folded into `tmux_command`: the interactive attach/switch-client/capture-pane paths must keep the user's locale for UTF-8 and status-bar rendering.

That rationale is *more* load-bearing after this change, not less. A reader who sees `-u` on the query constructor will reasonably ask why it is not simply on every tmux call, and the answer is no longer written down: `-u` asserts to tmux that the terminal is UTF-8, so putting it on an attach or switch-client would garble output on a genuinely non-UTF-8 terminal. AGENTS.md singles out non-obvious "why" as the most valuable prose in a diff, and this is a textbook case. Something like this restores it without re-lengthening the docstring much:

> Not folded into [`tmux_command`]: the interactive attach/switch-client/capture-pane paths must keep the user's locale, and `-u` would assert UTF-8 to a terminal that may not be.

## Suggestions

**1. The fix is partial by one hop, in a pre-existing way.** Two other name-producing calls still go through `tmux_command()`:

- `src/tmux/mod.rs:872`, `stop_all_sessions()`, which reads `list-sessions -F '#{session_name}'` and then feeds each line to `get_pane_pid(line)` and `kill-session -t line`.
- `src/tmux/mod.rs:1354`, `get_current_session_name()`, `display-message -p '#{session_name}'`.

Neither is a regression from this PR, and both are fine in the reporter's environment, because `LANG=ko_KR.UTF-8` sets `CLIENT_UTF8` on its own. They break when aoe's *own* environment has no UTF-8 in `LC_ALL` / `LC_CTYPE` / `LANG` and `TMUX` is unset, which is the normal state for a systemd unit with a scrubbed environment or a minimal container. `stop_all_sessions()` is the one that would hurt: the mangled name still passes `is_aoe_session` (the prefix is ASCII, `mod.rs:499-501`), so `get_pane_pid` returns `None`, the agent process tree is never killed, `kill-session -t <mangled>` fails silently into `let _ =`, and `killed += 1` increments anyway. The panic button would report "stopped N" while every agent kept running. Worth a follow-up issue rather than expanding this PR.

**2. The new test pins the implementation rather than the behavior.** `tmux_query_command_preserves_ctype_for_session_names` asserts the argv and env shape, which is the right cheap seam and worth keeping. But it cannot fail if `-u` ever stops having the effect it has today, and it would pass just as happily against a tmux where the flag did nothing. The test that would actually have caught #3536 is a tmux-gated one that creates a CJK-named session on the harness socket and asserts `refresh_session_cache()` returns the exact UTF-8 bytes, which is essentially the scratch test used to verify this. If you add one, the non-obvious part is worth writing into the test itself: it must clear `TMUX` from the child, or it passes vacuously everywhere this suite runs, since the suite runs under tmux. Entirely optional; the unit test plus the reproduction recorded in the PR body is a defensible line to draw for a three-line fix, and the test-cost rules in AGENTS.md argue against reaching for a live-tmux test without a reason.

## Merge risk

**Low.** One file, three added lines of behavior, reverts cleanly with `git revert`, no persisted state, no schema or migration, nothing that outlives the commit.

The blast radius is bounded by which tmux invocations changed: only the query tier, and everything it parses is either locale-insensitive or was verified above to survive. ASCII names, which is essentially all of them, are byte-identical.

The most likely way this bites is the case that could not be tested here: a build of tmux where `-u` behaves differently, or an old enough tmux that rejects the flag. `-u` has been documented since 3.0a and this is a global option accepted before the subcommand (`tmux -S <sock> -u list-sessions`), and it parses fine on 3.7c, but if some tmux ever refused it, *every* query would fail rather than degrade, which means the whole status surface would go quiet at once. That would be heard about loudly and immediately, which is the good failure mode. The coverage gap is that no test exercises `-u` against a live tmux, so the unit test would stay green through it. That is a small residual, and it is the reason for suggestion 2 rather than a reason to hold the PR.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Proportionate, with one dropped rationale sentence worth restoring
**Merge risk:** Low; residual is the untested-against-live-tmux `-u` flag, which would fail loudly rather than silently

Correct fix for a real, well-diagnosed bug, at the right seam, with the reasoning verified rather than asserted. The failure was reproduced on main and the fix on this branch, the mechanism confirmed against tmux's own source, and the error parsing the old override protected confirmed intact. The suite is fully green and clippy adds nothing new. The two suggestions are follow-ups, not conditions.

**One line on the merge block:** the PR is `MERGEABLE` but `BLOCKED` by `reviewDecision: CHANGES_REQUESTED`, from @jerome-benoit's review at 13:51 that asked for precisely the `-u` flag, which the author then added in `97f5f06` at 14:26. That review is stale and needs re-review or dismissal; nothing in the code is blocking it. (The `Lint PR title` check also shows as failed, but the run was `CANCELLED`, which reads as the superseded push rather than a real title problem, since the title is a valid `fix(tmux):` conventional commit.)

**LGTM** — thanks for chasing this one down, @atirna. Diagnosing a locale bug through to tmux's own `CLIENT_UTF8` sniff, and recognising that `LC_MESSAGES=C` alone would not have been enough, is the hard part and you got it right.

jerome-benoit (approved):
Re-reviewed at `3109143550362c5b7b52030672918de8d4d911f3`. The requested fix is correctly implemented: query clients use tmux's global `-u` option while inherited `LC_ALL` remains removed and `LC_MESSAGES=C` preserves stable diagnostics.

An independent tmux 3.7c reproduction confirmed exact Unicode session-name output under `LC_CTYPE=C` and the expected English missing-socket diagnostic. The latest commit only restores the rationale for keeping `-u` off interactive tmux paths. I found no remaining blocking issue in the PR diff. The cancelled title check should still be rerun separately.

njbrake (commented):

_Reviewed at `0a2e8045`. Push new commits and it will be re-reviewed._

---

_Re-reviewed at `0a2e8045`, following the earlier pass at `72ece483`. Push new commits and it will be re-reviewed._

---

## Summary

This is a re-review. The earlier pass at `72ece483` reached LGTM and asked for exactly one thing: restore the docstring sentence explaining why `-u` is not folded into `tmux_command()`. That has been done correctly, and nothing else moved.

The delta since `72ece483` is **four added doc-comment lines and zero lines of code**:

```
 /// [`tmux_no_server_running`]) and by `kill_session_if_present`. NOT folded into
 /// [`tmux_command`]: the interactive attach/switch-client/capture-pane paths must
 /// keep the user's locale for UTF-8 and status-bar rendering, and `-u` would
 /// assert UTF-8 to a terminal that may not be.
```

Both halves survived, as `fe2a1c09` claimed. Main's original reason is restored **verbatim**, including the "for UTF-8 and status-bar rendering" clause that the first restore attempt had paraphrased away, and the `-u` rationale is appended rather than substituted. No emdashes or `--` separators in the added prose.

## The authored diff is one file. The PR body's summary is not.

`git diff origin/main...0a2e8045` is **one file, `src/tmux/mod.rs`, +34/-7**, matching GitHub's own file list.

Worth stating precisely, because it settles the merge-resolution question completely: the author's merge of main means `origin/main` is now an *ancestor* of the head, so the two-dot and three-dot diffs are identical. The head tree is therefore provably `main` plus exactly those 41 lines. There is no conflict resolution to get wrong, and no main content was dropped or altered in either merge.

**The CodeRabbit release-notes block appended to the PR description is wrong** and should be disregarded. It claims this PR "Added session liveness snapshots" and "Re-exported `PaneGeom`". Neither string appears anywhere in the authored diff (0 occurrences). Traced to their real sources:

- `PaneGeom` re-export comes from **#3526** (`5988e958`), on main.
- The liveness snapshot comes from **#3508** (`adc053c2`), on main.

This is the same misattribution the earlier pass caught in CodeRabbit's "Out of Scope Changes" flag, now recurring in the body summary. CodeRabbit has had reviews **paused** on this branch since `8fb5b64b`, so that summary is stale as well as wrong. Nothing for @atirna to fix; the body text is bot-generated.

## Did anything change after jerome's approval?

Yes, but it is strictly a prose improvement, not a behavior change. @jerome-benoit approved at `31091435`; `fe2a1c09` landed after. The entire diff between them:

```
-/// ... and by `kill_session_if_present`. Not folded into
+/// ... and by `kill_session_if_present`. NOT folded into
 /// [`tmux_command`]: the interactive attach/switch-client/capture-pane paths must
-/// keep the user's locale, and `-u` would assert UTF-8 to a terminal that may not
-/// be.
+/// keep the user's locale for UTF-8 and status-bar rendering, and `-u` would
+/// assert UTF-8 to a terminal that may not be.
```

`src/tmux/mod.rs` is byte-identical between `fe2a1c09` and the head. His approval body remains accurate.

## Verification performed

I reproduced rather than re-derived the prior pass's central claim, on tmux 3.7c against an isolated socket (created and torn down; the real server was never touched):

| Case | Result |
| --- | --- |
| **PR path**: `-u`, `LC_ALL` removed, `LC_CTYPE=C`, `LC_MESSAGES=C` | `aoe_test_QA_협업_정리_9f4e4c6a`, byte-identical to input |
| Same but **without** `-u` | `aoe_test_QA___________9f4e4c6a` |
| **Pre-PR main**: `LC_ALL=C`, no `-u` | `aoe_test_QA___________9f4e4c6a` (reproduces #3536) |
| `LC_ALL` as the only UTF-8 source, `-u` present | name intact |

So `-u` is load-bearing, not decorative, and the docstring's specific claim about the `LC_ALL`-only case is true rather than merely plausible. tmux's own man page states the mechanism in the same terms the docstring uses: "`-u` Write UTF-8 output to the terminal even if the first environment variable of LC_ALL, LC_CTYPE, or LANG that is set does not contain UTF-8". `-u` has been a global flag since tmux 1.x, and the repo pins no minimum tmux version, so there is no compatibility floor to worry about.

The ENOENT marker that `LC_ALL=C` used to protect still matches under the new environment: `error connecting to /tmp/... (No such file or directory)`, which satisfies the anchored `line.ends_with("(No such file or directory)")` test in `tmux_no_server_running`. I could not reproduce *localization* of that string on macOS, which is expected, since Darwin libc does not localize `strerror`; the docstring correctly scopes the concern to glibc.

Argument ordering is correct. `tmux_command()` emits `-S <path>` or `-L <name>`, `tmux_query_command()` appends `-u`, and callers append the subcommand afterwards, so every global flag precedes the command.

I also checked the other four query call sites for locale sensitivity that removing `LC_ALL` could newly expose. All parse pipe-separated strings and integers (`#{session_activity}` as `i64`, `#{session_attached}` compared as a string). No float or time parsing, so `LC_NUMERIC` and `LC_TIME` are not in play.

**On the new test**, the earlier pass noted it pins argv/env shape rather than live behavior. I mutation-tested it: deleting `cmd.arg("-u")` fails it, and deleting `cmd.env_remove("LC_ALL")` fails it. It is load-bearing rather than tautological, and it satisfies the house rule against tests that restate a declaration.

## Definition of done

| Check | Result |
| --- | --- |
| `cargo test --features serve` | 6451 passed, 0 failed, 2 ignored (all integration targets green) |

Codecov reports all modified lines covered, project 85.41%.

## Design alignment

**Aligned.** The change extends the existing `tmux_command()` / `tmux_query_command()` split rather than adding a parallel path: the query variant remains the single seam where locale is scrubbed for parseability, and `-u` is applied there and only there. No new abstraction, no second mechanism, and the AGENTS.md rule about keeping OS-specific logic in `src/process/{macos,linux}.rs` is not implicated since this is locale handling, not platform branching.

Conventional commit prefix on the PR title. One test for one behavior, no per-input test explosion. No dead code, no `#[allow(dead_code)]`. No emdashes in the added prose. Coverage matrix is N/A (no web surface touched). The two merge commits lack conventional prefixes, but they are merges and the squash title is what lands.

## Automated review check-in

- **CodeRabbit body summary** ("session liveness snapshots", "Re-exported `PaneGeom`"): **disagree, false positive.** Merge-commit artifact from #3508 and #3526, proven absent from the authored diff above.
- **CodeRabbit line comment on `src/tmux/mod.rs:214`** (inherited `LC_ALL` precedence, at `fd3c9585`): **already addressed** by `env_remove("LC_ALL")` in `8fb5b64b`, and CodeRabbit itself confirmed this.
- **@jerome-benoit's CHANGES_REQUESTED** (asking for `-u`): **already addressed** in `97f5f06c`; he has since approved.

## Issues found

None.

## Documentation

Proportionate. The docstring change is the entire documentation delta, it explains non-obvious "why" that the code cannot convey, and it is now more complete than what it replaced.

One optional nano-nit, not worth a round trip on its own: `kill_session_if_present` in `src/tmux/utils.rs` still says its markers are "swallowed in the C locale", which this PR narrowed to the C *messages* locale. The matching is unaffected and the sentence is not wrong in substance. Fold it in only if you are touching that file anyway.

## Suggestions

Both carried over from the earlier pass, both still pre-existing and correctly out of scope here. Verified unchanged at the head: `stop_all_sessions()` (`src/tmux/mod.rs:874`) and `get_current_session_name()` (`src/tmux/mod.rs:1356`) still go through `tmux_command()` and can mangle names in a scrubbed environment. Worth a follow-up issue, not a change to this PR.

## CI workflow approval

**Withheld deliberately, and not because anything failed the audit.** The PR is `MERGEABLE` but `BLOCKED`: workflow runs are pending maintainer approval, so only "Lint PR title" has run. This agent does not approve workflow runs, so they remain pending for @njbrake. For what it is worth, the audit itself is clean: no workflow or CI config touched, no install/build/test hook touched, no dependency or lockfile change, no obfuscated or binary content, and the diff does exactly what the PR says.

## Merge risk

**Low.** Residual risk is close to nil, and lower than at the previous pass rather than merely unchanged.

- **Blast radius**: three lines behind one function, whose every caller was enumerated and checked.
- **Reversibility**: `git revert` fully undoes it. No migration, no schema, no persisted state, no released artifact.
- **Coverage**: the unit test is mutation-proven against both meaningful mutations, and the live behavior is reproduced above on the current tmux. No untested path was introduced.

The one thing a unit test still cannot catch is a future tmux dropping or changing `-u`, which would be a silent regression back to underscore substitution that we would first hear about from a user. That is a hypothetical about upstream, not a defect in this PR, and `-u` has been stable in tmux for well over a decade.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Merge risk:** Low, with no residual concern identified

The one change the previous review asked for was made, and made correctly: main's rationale is restored word for word rather than paraphrased, with the `-u` clause appended to it. The only commit that landed after @jerome-benoit's approval is that prose refinement, so his approval still describes what is here. The authored diff is one file at +34/-7; the extra files visible in the PR are main's, and the body's bot-written summary attributing #3508 and #3526 work to this PR should be ignored.

**LGTM.** Nice follow-through, @atirna, and good instinct on `fe2a1c09` to recover upstream's exact sentence instead of settling for the paraphrase that had already been approved.

PR #3542, opened 2026-08-26
feat(tui): configure new session open mode

### LINKED ISSUE #3541: feat(tui): configure new-session open mode independently
Add a Session settings select named `New Session Mode`.

Options: `Match default attach`, `Live mode`, and `Tmux`.

Default to `Match default attach` to preserve existing behavior.

Use this setting only after session creation. Keep `Attach Mode` for Enter and double-click on existing session rows.

This enables Enter = tmux attach, Tab = live-send, and new sessions = live-send.

athal7:
## Description

Adds an independent New Session Mode setting. It keeps existing-session activation unchanged and lets new sessions enter live mode. Resolves #3541

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: the schema-derived select uses the existing dashboard renderer. A Vitest PATCH contract test covers its dashboard behavior.

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**AI Model/Tool used:**
openai-codex/gpt-5.6-terra

Implemented and verified the settings split, the dashboard PATCH contract, and the new-session live-mode flow.

## Summary

- Added a **New Session Mode** setting with Match default attach, Live mode, and Tmux options.
- Kept **Attach Mode** for existing-session actions, including Enter and double-click.
- Applied the new setting only after session creation.
- Updated the TUI, documentation, demo setup, and tests.
- Added dashboard PATCH contract coverage and TUI end-to-end coverage.

## Benefits

Users can open new sessions in live mode without changing existing-session behavior. The default preserves current behavior.


njbrake (commented):

_Reviewed at `8071bd0d`. Push new commits and it will be re-reviewed._

---

## Summary

Splits new-session open behavior out of the single `Attach Mode` setting into a new `session.new_session_mode` select with three options (`match_default` / `tmux` / `live_send`), defaulting to `match_default`. Existing-session activation (Enter, double-click) keeps reading `default_attach_mode`. Resolves #3541.

First, the good news, because there is a lot of it: this is a well-shaped contribution. The setting is declared exactly the way this repo wants settings declared, the default is chosen to preserve every existing user's behavior, both creation paths are covered, and it arrives with a unit test, an e2e test, a docs update, and a coverage-matrix entry. Several things that looked like risks going in turned out to be fine on closer reading.

## Design alignment

**Aligned with minor deviations.**

Compared against the three sibling resolvers in `src/tui/home/mod.rs` (`click_action` at 7531, `default_attach_mode` at 7541, `live_send_on_view_switch` at 7568) and the sibling enums in `src/session/config.rs` (`AttachMode`, `ClickAction`, `RowTagMode`):

- The setting is declared once with `#[setting(...)]` on `SessionConfig` (`src/session/config.rs:1299-1308`) and nothing else. No `FieldKey`, no `build_*_fields`, no `apply_field_*`, no `*ConfigOverride`. That is exactly the single-source path in `docs/development/adding-settings.md`, and it is why a three-surface feature is only 15 lines of non-test Rust. `settings_schema/registry.rs:23` calls `SessionConfig::settings_descriptors()`, which is derive-generated, so the field reaches the TUI Interaction tab, `GET /api/settings/schema`, PATCH validation, and profile overrides with no further edits.
- `new_session_attach_mode` (`src/tui/home/mod.rs:7552`) sits directly beside `default_attach_mode`, uses the same `resolve_session_config_for` plus `.map()` shape, and has the same `pub(super)` visibility. Correct placement and idiom.
- The issue asked for this on the **Session** tab; the PR put it on **Interaction**. That is the right call, not a miss: `src/tui/settings/fields.rs:1633-1643` asserts `session.default_attach_mode` must *not* be on the Session tab, so putting its sibling anywhere else would have been inconsistent.

**One deviation worth a maintainer decision, not a code change.** This setting existed before and was deliberately removed. `src/migrations/v022_prune_tuning_settings.rs:6-8` and `:87` record it:

> `session.new_session_attach_mode` merged into `session.default_attach_mode` (one "Attach Mode" setting covering both the post-create attach and Enter/double-click activation).

That was #2914, "prune internal tuning knobs and consolidate overlapping settings", merged 2026-07-17. So this PR re-splits a setting the project consolidated about six weeks ago.

That is not automatically wrong, and the PR is arguably a *better* design than what #2914 removed: the old `new_session_attach_mode` was a plain `AttachMode`, so users had two independent knobs to keep in sync, which is presumably why it got merged away. The new `MatchDefault` default makes the two settings a hierarchy instead of a duplicate pair, which removes that failure mode. But it is an unacknowledged reversal of a recent explicit decision, and @njbrake is the one who gets to say whether the split comes back. **Ask:** please mention #2914 in the PR description and say why `MatchDefault` addresses what the consolidation was solving. That is a description edit, not a code edit.

## Verified claims

Rebased onto `origin/main` (`5988e958`) and ran everything. Recording what was actually reproduced versus what was only read:

| Claim | Status |
|---|---|
| `cargo fmt --check` | Clean |
| `cargo clippy --features serve --all-targets` | Exit 0, 5 warnings, **all pre-existing** (`acp/background_agent.rs:745`, `acp/supervisor.rs:4030`, `process/macos.rs:266`, `session/smart_rename.rs:1546`, plus one `useless_format`). Nothing new from this PR. |
| `cargo test --features serve` | **Fully green**, exit 0 |
| `cargo test --features serve,e2e-tests --test e2e new_session` | **13/13 pass**, including `test_new_session_enters_live_mode_when_configured` |
| `npm run format:check` (oxfmt) | Clean |
| `npm run lint` (ESLint) | Clean |
| `node web/tests/validate-coverage-matrix.mjs` | Passes |
| Vitest `NewSessionModePatch` plus `SchemaFieldPatches` | 7/7 pass, no type errors |
| "Keeps existing-session activation unchanged" | **Verified.** `new_session_attach_mode` has exactly two callers: `dispatch_new_session_attach` (`src/tui/app.rs:3625`) and the unit test. Every existing-row path still reads `default_attach_mode`: `input.rs:2765,4222,4238`, `render.rs:3815`, `mod.rs:7581`, and the help overlay via `profile_default_attach_mode`. |
| Both creation paths routed | **Verified.** Sync path `input.rs:6942` emits `AttachAfterCreate` into `app.rs:3444`; async path `app.rs:1889` calls `dispatch_new_session_attach` directly. Both land on the new resolver. |
| No migration needed | **Verified, and this was the thing most expected to be wrong.** The field is additive with `#[serde(default)]` and `NewSessionMode::default() == MatchDefault`, which resolves to `default_attach_mode`. An existing `config.toml` with no `new_session_mode` key behaves identically to before. A *pre*-v022 config with `new_session_attach_mode = "live_send"` is still merged into `default_attach_mode` by v022 and then honored through `MatchDefault`. No `.schema_version` bump required. |
| No emdashes in added prose | Verified, none |
| `docs/cli/reference.md` regeneration | Not needed, no clap changes |
| `flake.nix` embedded assets | Not needed, no new `include_*!` |

## On the two things that looked like risks and are not

**`src/tui/home/tests.rs` -42 is not lost coverage.** Checked whether the deleted `defaults_to_tmux_when_no_config_present` and `returns_live_send_when_globally_configured` took real assertions with them. They did not: there is a sibling `mod default_attach_mode` immediately below (starting `src/tui/home/tests.rs:17182`) that already owns both, at `:17201` and `:17226` respectively, plus three more Enter/Terminal-view cases. The two deleted tests were duplicates of tests in the adjacent module, and the replacement table test raises the assertion count in that module from 2 to 4. This is a good deletion.

**`tests/e2e/new_session.rs` -30 deletes no tests.** The one test in that area is preserved and still passes; the -30 is the helper rewrite plus comment churn. `#[parallel]` is retained, correctly (no process-global state mutated).

**The e2e config-seeding rewrite is correct, and that deserves saying explicitly because it looks scarier than it is.** Writing `state.toml` directly rather than `[app_state]` in `config.toml` is right: `app_state` moved to `state.toml` in migration v021, `Config::load` strips `app_state` from `config.toml` (`src/session/config.rs:2990-3000`), and v021 treats a pre-existing `state.toml` as authoritative (`v021_split_app_state_to_state_toml.rs:50-55`). Appending `[session]` to the harness config is safe because `TuiTestHarness::new` seeds only `[updates]` and `[app_state]` (`tests/e2e/harness.rs:351-361`), so there is no duplicate-table parse error. And the e2e run confirms it. This also resolves CodeRabbit's first finding properly.

## Issues found

None blocking. Everything below is small and cheap.

**1. The Vitest test cannot fail if the setting is deleted, and it does not test the app's PATCH construction.** `web/src/components/settings/__tests__/NewSessionModePatch.test.tsx:49-54` renders `SchemaSection` with a hand-written descriptor passed as a prop, and supplies its own `onSaveField` lambda that builds `{ session: { [field]: value } }` inline. So the assertion at :62 checks a payload the *test* constructed, not the one the dashboard constructs. Delete `session.new_session_mode` from `config.rs` and this test still passes.

The sibling `web/src/components/settings/__tests__/SchemaFieldPatches.test.tsx` does the stronger thing: `renderTab()` (:120) mounts the real `SettingsView`, so its assertions pin the actual `saveField` into `updateProfileSettings(profile, patch)` wiring. Its header comment (:3-21) describes it as exactly the home for "payload-permutation coverage for the schema-driven settings tabs".

**Suggestion:** move this case into `SchemaFieldPatches.test.tsx` as one more `SCHEMA` entry plus one `it`, which is also what AGENTS.md's "one test per behavior, permutations as rows" asks for. That drops a 64-line file and the duplicate `story.settings-new-session-mode` coverage-matrix entry.

**This one is genuinely not the contributor's fault, and it is worth flagging to the maintainer instead.** CodeRabbit told you to use MSW, citing `web/AGENTS.md:27`, which says request-payload tests belong in "Vitest + RTL + MSW" and names `web/src/components/settings/__tests__/SoundSettings.test.tsx` as the canonical example. **That file does not exist**, and after this PR your test is the only file in the entire `web/src` tree that uses `setupServer`. Everything else uses `vi.mock("../../../lib/api")`. So the written guideline and the actual house pattern have diverged, you followed the written one, and the result is an outlier. @njbrake: either `web/AGENTS.md:27` needs updating or the existing tests need migrating, but a contributor should not have to discover that by reading the whole test directory.

**2. Two tests in the renamed module now exercise the wrong function.** `src/tui/home/tests.rs:17099` (`returns_none_for_missing_instance`) and `:17119` (`returns_none_for_acp_session`) both still call `env.view.default_attach_mode(...)`, but their comments describe `dispatch_new_session_attach`, which now calls `new_session_attach_mode(...)`. They pass because both wrap the same `resolve_session_config_for`, and the existing-row resolver already has its own module covering it (`mod default_attach_mode` below). Change both to `new_session_attach_mode` so the post-create module actually guards the post-create function. Two one-word edits.

**3. `AttachMode::Tmux` lost its doc comment as collateral.** `src/session/config.rs:1555-1556`: `LiveSend` kept its variant doc, `Tmux` lost `/// Attach to the session's tmux pane (the historical behavior; the user lands inside tmux with the agent running).` Nothing in the PR required that deletion, and it leaves the enum asymmetric. Restore it. Relatedly, `NewSessionMode`'s three variants (`:1544-1549`) have no docs, where `AttachMode`, `ClickAction`, and `RowTagMode` all document theirs.

**4. Three "why" comments went missing from the tests.** In `src/tui/home/tests.rs`, the old docs on `add_session` explaining that `source_profile = "test"` is what makes the resolver read the test profile's config, and on the config-writing helper. Per AGENTS.md, comments that explain non-obvious *why* are the valuable kind, and "why is this line here" is exactly what `source_profile = "test"` needs. The module doc also shrank enough that it no longer mentions the sync-create-path invariant, which `sync_create_path_emits_attach_after_create_not_attach_session` (:17155) still tests right below it. This is also what CodeRabbit's docstring-coverage warning (77.78% against 80%) is pointing at.

**5. The setting's description is one line, and because settings are single-source, that one line is the user-facing text on three surfaces.** `src/session/config.rs:1298` reads "How the TUI opens a terminal-mode session after it is created." A user looking at a dropdown whose first option is "Match default attach" gets no explanation of what it matches. The old combined description also carried "Terminal/Tool views and acp sessions ignore this setting", which is still true (`dispatch_new_session_attach` short-circuits structured sessions at `app.rs:3616-3623`) and is now documented nowhere. Two extra sentences here pay off in the TUI row, the web field, and settings search.

**6. The e2e docstring overclaims slightly.** `tests/e2e/new_session.rs:408`: "New-session mode must remain independent from the setting that controls Enter and double-click." The test only sets `new_session_mode = "live_send"` and asserts the LIVE banner; it never activates an existing row, so it does not demonstrate independence. The unit table does cover independence. Just reword to what the test proves.

**7. Two cosmetic bits in the new Vitest file, if it survives finding 1.** The `http.get("*/api/settings/schema", ...)` handler at :41 is dead, since `SchemaSection` takes `schema` as a prop (`SchemaSection.tsx:20`). And the fixture says `category: "Session"` (:14) where the real annotation is `category = "Interaction"` (`config.rs:1303`).

**8. Optional, and arguably against house rules, so your call.** Nothing asserts the resolver's behavior with *no* `config.toml` at all, which is the actual upgrade path (absent key, serde default, `MatchDefault`). The table writes an explicit value on every row. `create_test_env_empty()` gives you a config-free home, so it is three lines before the loop. That said, AGENTS.md explicitly discourages tests that only restate a default, and the mechanism here is `#[serde(default)]` plus `#[default]`, so skipping it is defensible.

## Test coverage assessment

One more worth adding, and it is the one that would actually protect this feature. There is an established pattern for pinning a setting's presence in the single-source schema: `src/session/config.rs:4266-4281`, `test_vt_live_in_settings_schema`, which looks the field up in `settings_schema::schema()` and cites `docs/development/adding-settings.md`. A four-line equivalent for `session.new_session_mode` (or adding it to the Interaction allowlist in `src/tui/settings/fields.rs:1624-1630`) is the assertion that fails if the setting is ever accidentally dropped. Right now, with finding 1 outstanding, nothing in the repo fails if `new_session_mode` disappears from `config.rs`.

## Automated review check-in

CodeRabbit's latest pass found no actionable comments. Its two earlier line findings:

- **e2e config overwrite dropping harness `app_state` flags**: valid, and **already addressed** in `8071bd0d`. The author's fix goes further than the suggested diff by writing `state.toml` directly, which is more correct given migration v021. Agree with the finding, agree with the resolution.
- **"Use MSW for the PATCH payload test"**: **partially disagree**. The finding correctly quotes `web/AGENTS.md:27`, but that guideline cites a file that does not exist, and no other test in `web/src` uses MSW. Following it produced a test that verifies the HTTP boundary while losing the `SettingsView` wiring the sibling test covers. Net, the change made the test weaker, not stronger. See finding 1.
- **Docstring coverage 77.78%**: agree in substance; see finding 4.

## What is blocking, in one line

CI has never run. The `CI` workflow sits at `action_required` awaiting fork approval, and the `PR Title Check` workflow (a `pull_request_target` job) has one `startup_failure` and one `failure` where the job itself was `cancelled`, which looks like infrastructure rather than a real title problem, since the title is a valid scoped conventional commit. Approving the run should clear it. The full suite was run locally in the meantime and is green.

For whoever approves it, the audit is clean: head SHA `8071bd0d`, no changes to `.github/workflows/**` or any CI config, no `package.json` / lockfile / dependency changes, no install, build, or test hooks (`postinstall`, `build.rs`, `Makefile`, `conftest.py`, husky), no binary or minified or obfuscated content, no network egress. The only script touched, `web/scripts/record-tui-demo.mjs`, is a comment-only edit.

## Documentation

**Proportionate**, with one gap.

`docs/development.md:172` and `docs/guides/live-mode.md:16-19` are both accurate and correctly placed in `docs/` (canonical) with nothing compensating in `website/`. The `record-tui-demo.mjs` recipe fix is right, and it removes a stale reference to the long-gone `new_session_attach_mode`. The `palette-scratch-launch.spec.ts` comment fix is a nice catch: that comment described a setting matching `/New session/i` that has not existed since #2914, so the PR makes the test's stated rationale true again.

The gap is in `docs/guides/live-mode.md:16-19`. The old text pointed at the settings by name; the new text says "set live mode as your default so that Enter (and clicking a row) drops you straight into it" without naming **Attach Mode**, so a reader is told the behavior exists but not what to change. Worth naming both settings in that bullet. Separately and not this PR's doing, line 40 still says "**Default Attach Mode**" where the label is now just "Attach Mode".

See also finding 5: the setting's own description is the user-facing docs for three surfaces and is currently thinner than its neighbors.

## Merge risk

**Low.**

Blast radius is a TUI-only preference that defaults to a no-op. Reversibility is clean: `git revert` fully undoes it, since the field is additive with a behavior-preserving default and no migration or stored-data change is involved. An orphaned `new_session_mode` key in someone's `config.toml` after a revert is ignored by serde.

Coverage of the changed paths is real: both creation dispatch sites route through the new resolver, the unit table covers all four mode combinations, and the e2e test exercises the sync create path end to end through tmux.

The residual gap is finding 1. Because the only test asserting this setting exists on the dashboard uses a hand-written fixture, a future refactor that drops or renames `session.new_session_mode` would leave the whole suite green while the setting silently vanished from the TUI and web settings. That is the most likely way this bites, and today nothing would surface it. The `test_vt_live_in_settings_schema`-style test in the coverage section closes it for about four lines.

## Verdict

**Design alignment:** Aligned with minor deviations
**Implementation:** Approve
**Project fit:** Accept, pending @njbrake's call on the #2914 reversal
**Merge risk:** Low, with one gap: nothing fails if the setting is later dropped from the schema

This is a clean, correctly-shaped implementation of a small feature, built through the settings mechanism the project actually uses rather than beside it, with a default chosen so no existing user notices anything. Everything flagged is a polish item worth a follow-up commit; none of it is a defect. The one thing that is not a reviewer's to decide is whether the new-session setting should come back at all, six weeks after #2914 deliberately merged it away, and that is a description edit plus a maintainer nod rather than any change to the code.

Thanks for the thorough job here, @athal7. Filing the issue first, picking a default that preserves existing behavior, and shipping unit, e2e, and dashboard tests together with a coverage-matrix entry is exactly the shape this repo wants.

jerome-benoit (approved):
Reviewed the current head after the `main` synchronization. I found no remaining functional issue.

The synchronous and asynchronous creation paths both use the new-session resolver, while Enter and double-click on existing sessions remain governed by `default_attach_mode`. Structured sessions bypass the terminal modes, profile-specific resolution uses the created session's source profile, and the additive `MatchDefault` value preserves existing and v022-migrated configurations.

I did not duplicate the non-blocking test and documentation improvements already covered in the existing review.

Approval is for the code at `19b8ab24399876d8c05782ee840b5b132f7071cd`. The full Rust/web CI suites remain `action_required`; only the title and template checks have run on this head.

njbrake (approved):
CI green on ce1da2cd. Pushed a guard for the `new_session_mode` select options and its Interaction tab placement, pointed the two post-create tests at `new_session_attach_mode`, and restored the `AttachMode::Tmux` variant doc.

_Note: this review was drafted and submitted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #2237, opened 2026-06-17
Structured view: Stop button is a no-op when the daemon has no prompt in flight, stuck spinner only escapes via aoe acp restart


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Structured-view sessions get stuck "thinking" / mid-tool with a spinner that never clears, and the composer Stop button does nothing. The only escape is `aoe acp restart <id>`. This is happening more and more and makes long-lived sessions unusable.

The Stop button is not just slow here, it is a hard no-op: clicking it once, twice, or ten times changes nothing. The agent appears frozen in a random state (thinking, or partway through writing a file), and the turn never ends.

## Root cause

This is the acknowledged residual of #1216, now reachable through the everyday Stop button.

The daemon's per-turn prompt loop owns the terminal `Stopped` and the cancel-escalation path. When a worker reattaches or respawns and adopts an already in-flight turn, that turn has no owning `prompt_fut` on the new client. The resume-idle watchdog that would synthesize a `Stopped` for an orphaned turn disarms permanently the moment the adopted turn emits its first event:

`src/acp/acp_client.rs:4634-4650`
```rust
if first_event_after_attach.load(Ordering::Relaxed) {
    // ... any further silence is normal mid-turn reasoning ... Disarm
    // permanently. The narrow residual (the turn completes after attach
    // and its PromptResponse is lost, leaving a stale spinner) is rare
    // and recoverable via force-end-turn / a new prompt. See #1216.
    return;
}
```

So when the adopted turn finishes (or wedges) and its `PromptResponse` is lost, no terminal `Stopped` is ever emitted. The UI keeps `turnActive = true` while the daemon has no prompt in flight. This is the "rare" residual, except it is no longer rare: every build-stale respawn adopts the in-flight turn, and someone developing `aoe` rebuilds constantly. Session `768e497afa9e4415`'s worker log shows the exact sequence (build-stale respawns adopting an in-flight turn, then a Stop that finds no prompt, then the manual restart):

```
05:43:57  adopting build-stale structured view worker to drain in-flight turn before respawn
06:01:59  sending session/cancel (no prompt in flight)
06:02:09  registry entry gone while worker handle live; tearing down reason=restart_pending
```

Now the second failure, which is what makes it unrecoverable from the UI: with no prompt in flight, the Stop button is structurally a no-op.

[...]

PR #3544, opened 2026-08-26
feat(tui): resolve structured approvals from home

athal7:
## Description

Lets you answer a pending structured (ACP) permission request from the home
session list, without entering the structured view. Pressing `a` on a
structured row opens the existing permission-response dialog; its choice is
resolved through ACP on a background worker.

`/api/sessions` gains a `pending_approvals` projection (nonce + tool name +
target + destructive flag) for **running** structured workers; the TUI's 1 Hz
daemon-status poller carries it into a `structured_pending_approvals` map, and
the dialog resolves the captured nonce through `POST /api/sessions/{id}/acp/approvals/{nonce}`.

Hardened per review (@njbrake) before merge:
- **Only live workers project approvals.** Gated on `acp_worker_state == Running`,
  and `spawn` now runs `cancel_orphaned_approvals` like `attach` does, so the
  durable log has no orphans by the time a worker is running. Closes the
  "phantom approval that can only 404" case (stopped session / respawned worker).
- **The dialog shows what is being approved** (tool + target, `· destructive` in
  red) and drops the terminal-only "raw keystrokes" guidance, which is wrong on
  the ACP path.
- **Indexed query.** The pending-approval lookups filter on the `discriminant`
  column so they hit `idx_acp_events_session_discriminant_seq` instead of a full
  `json_extract` scan on the 1 Hz path.
- `ApprovalGone` now shows a distinct "already resolved" dialog; a transient
  failure is left to the next poll to restore rather than manual re-insert.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] N/A: this PR doesn't change a user-facing dashboard flow (Rust-only; the field is deliberately not consumed by the dashboard)

**TUI / CLI (e2e test)**
- [x] Added or updated an e2e test under `tests/e2e/` (`acp_home_approval_e2e.rs`: live daemon + fake ACP agent, resolves an approval from the home list and asserts the turn completes). Plus unit tests for the projection gate + fields, the choice->wire mapping, the Gone/Failed apply paths, and the structured dialog body.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** anthropic/claude-opus-4 (Oh My Pi harness)

**Any Additional AI Details you'd like to share:** AI drafted the implementation and tests under my direction; I reviewed every change.

## Summary

- Adds ACP permission approval to the home TUI.
- Shows the requested tool, target, and destructive status.
- Resolves approval choices in the background without blocking the TUI.
- Exposes pending approvals through the sessions API.
- Handles resolved, stale, and failed approvals with clear feedback.
- Cancels orphaned approvals when workers start or restart.
- Improves approval target summaries and prompt handling.
- Adds unit tests and an end-to-end test.

## Benefits

Users can review and approve ACP requests from the home session list. The TUI remains responsive while approvals resolve. Stale approvals do not reappear after worker recovery.


njbrake (commented):

_Reviewed at `5013d308`. Push new commits and it will be re-reviewed._

---

## Summary

Wires the existing `a` / `A` ("Respond to permission prompt") binding into structured (ACP) sessions on the home screen. `/api/sessions` gains a `pending_approval_nonces` projection, the TUI's 1 Hz daemon-status poller carries it into a new `structured_pending_approvals` map, and `a` opens the existing `PermissionResponseDialog` whose choice is resolved through `POST /api/sessions/{id}/approvals/{nonce}` on a new background `Worker`.

This is a real gap being closed and the plumbing is done with care. Three things worth calling out as genuinely well done before the findings, because they are the parts most likely to have gone wrong:

- **The nonce is captured at dialog-open time**, not re-read at keypress (`input.rs:6539`, `PermissionResponseTarget::Structured { session_id, nonce }`). That is the correct choice, and it means a poll tick landing between render and keypress cannot retarget the answer at a different approval. A stale nonce 404s instead of resolving something else. This was the failure looked for hardest and it is not there.
- **`ORDER BY seq ASC`** (`event_store.rs:1320`) makes `.first()` deterministic, and it happens to match the structured view's own default selection (`structured_view/state.rs:584` also takes `pending_approvals.first()`, which is push order). Home and the structured view will agree on which approval is "the" approval.
- **The worker pattern is textbook-aligned.** `StructuredApprovalPoller` is request/response over `tui::worker::Worker`, exactly like `DeletionPoller` / `StopPoller` / `DaemonStatusPoller`. It adds **no new poll cadence** and does no work until the user submits a decision. The `list-sessions`-fork hazard from #3508 does not recur here. One parked thread per TUI launch, same as its four siblings.

Server-side replay safety is also fine, verified end to end: `resolve_approval` into `Supervisor::resolve_permission` into `AcpClient::resolve_permission` (`acp_client.rs:2995`) takes the `pending_responders` mutex and `remove`s the entry, so a double-answer (TUI and dashboard racing) is a compare-and-set by construction. The second one gets `UnknownNonce`, then 404, then `ApprovalGone`. The nonce is **not** client-trusted.

On nonce exposure specifically, since the PR body highlights it: the nonce is an id, not a bearer secret. Every client that can `GET /api/sessions` can already read `ApprovalRequested` events (which carry the nonce) off the ACP event stream, and `read_only_block` still refuses the resolve POST. Exposing it here does not widen the trust boundary.

## Design alignment

**Diverges, unacknowledged.**

The precedent is `src/tui/structured_view/`, which is how this codebase already resolves ACP approvals from a TUI: `pending_approvals` comes from the live transcript reducer fed by the daemon's event stream (`reducer.rs:164`, carrying `nonce`, `title`, `kind`, `args`, `destructive`), the shelf renders `Approval N/M · <title> · <target>` with `· destructive` in `theme.error` (`render.rs:284-320`), and `input.rs:878-940` branches `Ok` / `ApprovalGone` / `Err` into three distinct toasts.

This PR introduces a **second source of truth** for "what is pending": a projection of the durable SQLite event log. Home genuinely cannot reuse the reducer (it holds no per-session SSE connection), so *a* projection through `/api/sessions` is defensible and the mechanism is not the objection. What diverges is what is projected and where it is read from:

1. **It reads the durable log, not the live worker.** `supervisor.rs:2586-2591` states the invariant in so many words: *"a pending nonce only exists on a live worker."* `unresolved_approval_nonces` is documented (`event_store.rs:1301-1307`) as a *reattach recovery* query, and it is used that way by its only three prior callers. Repurposing it as "currently pending" breaks the invariant. Details in issue 1.
2. **It projects a bare `Vec<String>`.** Every other approval surface in the product renders what is being approved. The home dialog renders the *session* title and three buttons. Details in issue 2.
3. **`ApprovalGone` is silent** where the precedent shows an "approval already resolved" toast.

None of this is argued in the PR body, which is why this reads as unacknowledged rather than a deliberate departure. The fix is not a rewrite: widen `pending_approval_nonces: Vec<String>` into a small struct and gate the projection.

## Code review

| File | Assessment |
| --- | --- |
| `src/acp/event_store.rs` (+2/-1) | Just `ORDER BY seq ASC` on the outer SELECT (the `)` above it closes the subquery). Correct, and no regression for the three existing callers: two use `is_empty()`, the third iterates order-independently. Storage-layer hazards checked: retention prunes strictly `seq <= cutoff` with only `NON_SUBSTANTIVE_EVENT_DISCRIMINANTS` pinned, and `ApprovalRequested` always precedes its `ApprovalResolved`, so pruning cannot resurrect a resolved approval. The `NOT IN` NULL trap fails safe (returns zero rows). Nothing about what is persisted changed. |
| `src/server/api/sessions.rs` (+58/-2) | New field plus projection. Correctly `#[serde(default, skip_serializing_if)]` so old clients are unaffected. Two problems: the gate is wrong (issue 1) and the query is unindexed on a 1 Hz path (issue 3). |
| `src/tui/approval_poller.rs` (+100, new) | Clean, idiomatic, mirrors `daemon_status_poller.rs` closely. Runtime built once and reused; build failure degrades to a `Failed` result rather than a panic. No complaints. |
| `src/tui/daemon_status_poller.rs` (+7) | Pure field passthrough with `#[serde(default)]`. Fine. |
| `src/tui/dialogs/permission_response.rs` (+10/-1) | `structured()` constructor forcing `supports_allow_always = true`. Checked whether this over-promises: it does not, because `pick_option_id` (`acp_client.rs:4603`) degrades `AllowAlways` to `AllowOnce`, and the structured view's own shelf offers `A` unconditionally too (`render.rs:340`). Aligned. But see issue 2 about the dialog *body*. |
| `src/tui/home/input.rs` (+45/-15) | Specifically checked whether the deletions changed anything for terminal rows: they do not. `title` / `tool` just moved above the `is_structured()` branch so the `&Instance` borrow ends before the mutations. `bindings.rs:611` is untouched (`a` non-strict, `A` strict, `Context::Always`). The only removed behavior is "structured = silent no-op". Non-`serve` builds keep the silent early return. |
| `src/tui/home/mod.rs` (+125/-13) | State, resolve, and result-apply. See issue 4 and suggestions. |
| `src/tui/app.rs` (+4) | A `try_recv` per loop iteration inside the existing `#[cfg(feature = "serve")]` block. Free. |
| `src/tui/home/tests.rs` (+47/-6) | Proportionate in count; gaps noted under issue 5. |

`cargo fmt --check` clean. `cargo clippy --features serve -- -D warnings` clean. `cargo clippy --features serve --all-targets -- -D warnings` fails with exactly the 5 pre-existing errors on `main` (`background_agent.rs:627`, `:745`, `supervisor.rs:4030`, `process/macos.rs:266`, `smart_rename.rs:1546`) and **nothing new**. No emdashes in added prose. No dead code, no `#[allow(dead_code)]`. Conventional commits on both commits. No settings schema, migration, or `gen-docs` surface touched. Rust-only, so no `web/tests/coverage-matrix.json` entry is owed (the field is deliberately not consumed by the dashboard).

`cargo test --features serve`: **6414 plus 187 passed, 1 failed**, `terminal_smart_rename::forced_rename_bypasses_a_prior_failed_attempt` (`tests/integration/terminal_smart_rename.rs:205`, got `"Britons"`, wanted `"Retry title wins"`). **Not attributable to this PR**: it passes on rerun in isolation, it is `#[serial]` and depends on a fake `claude` on `PATH` plus real tmux pane content, and the PR touches no smart-rename, tmux, or `PATH` code. Reported because any failure is worth reporting; it looks like a load-dependent flake worth its own issue.

Two process notes from CI, both easy: the repo's `check-template` bot has posted a **"PR will be automatically closed in 24 hours"** warning because the description dropped the required checklist. Worth restoring the template. CodeRabbit also flags docstring coverage at 51.85% against an 80% threshold.

## Automated review check-in

CodeRabbit posted one actionable finding: `self.pending_daemon_status_refresh = false;` in the `Disconnected` arm of `apply_structured_approval_results`, clearing a different poller's debounce flag. **Agree, and already addressed** in `5013d30` ("fix(tui): isolate approval poller recovery"). The current code no longer touches that flag. CodeRabbit's latest pass reports no remaining actionable comments; its "Merge Risk: Minimal" verdict is the one place this review parts company with it, for the reasons below.

## Issues found

### Blocking

**1. `structured_live` is the wrong gate: it manufactures approvals that can never be answered.** `sessions.rs:715` defines `structured_live = is_structured() && !is_archived() && !is_trashed()`, with **no check that a worker exists**. But the codebase's own invariant (`supervisor.rs:2586-2591`) is that a pending nonce only exists on a live worker, and the sweep that reconciles the durable log against reality (`cancel_orphaned_approvals`, `supervisor.rs:3121`) has exactly one call site, `supervisor.rs:3098`, inside `pub async fn attach` (`:2897`). It never runs from `spawn()` and never from `shutdown()`.

Two concrete reachable states:

*Stopped session, loud failure.* Agent requests permission, `ApprovalRequested` is recorded, user presses `x` to stop the session. `shutdown_with_reason` (`supervisor.rs:2816`) publishes `Stopped` and never an `ApprovalResolved`. `daemon_status_applies_to` (`home/mod.rs:3121`) does not exclude `Stopped` rows, so the nonce lands in `structured_pending_approvals`. User presses `a`, dialog, Allow, `client_for_session`, `UnknownSession`, 404 *"session has no running structured view"*, `classify_resolve_error` returns `SessionNotFound` (not `ApprovalGone`), `Failed`, and the result is a **"Respond Failed" info dialog, the nonce re-inserted, and the next 1 Hz poll putting it right back.** The affordance is permanently broken and shouts about it every time.

*Freshly spawned worker, silent failure.* Daemon dies with an approval on screen. On restart the runner is gone, so a later prompt goes through `spawn()`, which does **not** sweep. The worker is `Running`, its `pending_responders` is empty, but the log still says unresolved. User presses `a`, Allow, `UnknownNonce`, 404 with nonce echo, `ApprovalGone`, and the TUI **silently** drops the nonce with no feedback at all, with the next poll restoring it. A dialog that looks live whose answer goes nowhere, quietly.

Gating on `acp_worker_state == Running` fixes the first case cheaply (the value is already in scope at `sessions.rs:729`) and would also skip the SQLite query for every non-running structured row. It does not fix the second. The durable fix is to source the pending set from the supervisor's live `pending_responders`, and there is no public accessor today so that is new plumbing; or, if the log projection stays, to also run `cancel_orphaned_approvals` from `spawn()`, which is arguably a standalone bug fix worth its own PR.

**2. The dialog never says what is being approved, and its one line of guidance is wrong on this path.** `PermissionResponseDialog::render` (`permission_response.rs:136`) hard-codes: *"AoE sends these as raw keystrokes; make sure the prompt is on screen."* On the ACP path nothing is sent as keystrokes, and "make sure the prompt is on screen" is precisely the safety premise that no longer holds.

Being fair about the scope, because this was initially over-called: home **does** auto-mount an `EmbeddedView` for a selected structured row (`home/mod.rs:758`, `render.rs:2768`), and `EmbeddedView::render` delegates to the full structured renderer including the approval shelf. So in the common case the user *does* see `Approval 1/1 · <title> · <target> [· destructive]` in the preview above the dialog.

But the shelf is gated on `matches!(self.view_mode, ViewMode::Structured)` (`render.rs:2729`), while the `a` binding is `Context::Always`. So in **Terminal or Tool view mode**, while `structured_preview_pending` is still connecting, or after a failed connect, `a` opens a content-free approve/deny dialog on an unseen tool call, and the approval the codebase itself flags `destructive: true` (`permissions.rs:20`) looks identical to a `read`.

Everything needed is already there: the `ApprovalRequested` event carries `tool_call.name`, `args_preview`, and `destructive`, and `approval_target()` (`structured_view/render.rs:363`) is the existing helper that turns those into the one-line summary. Widening the projection from `Vec<String>` to `Vec<{nonce, title, target, destructive}>` and rendering it in the dialog would close this and issue 3's "which one am I answering" ambiguity in the same change. At minimum, the keystroke sentence needs to become mode-aware.

Related: the deleted doc comment on `open_permission_response_dialog` (`input.rs:6396`) said *"the user has already visually confirmed the prompt is showing, and AoE never parses pane content to verify it."* That was the documented justification for the whole feature having no `Status::Waiting` gate, and this PR is the change that stops it from being true for structured rows. Losing it removes the reader's only chance to notice.

**3. An unindexed full JSON scan lands on a 1 Hz path.** `DAEMON_STATUS_REFRESH_INTERVAL` is `Duration::from_secs(1)` (`app.rs:874`), so the home TUI now calls `unresolved_approval_nonces` **once per second per structured session**, plus 0.33 Hz per open dashboard tab (`useSessions.ts:6`). The query filters on `json_extract(...) IS NOT NULL` rather than the indexed `discriminant` column, so it cannot use `idx_acp_events_session_discriminant_seq` and scans plus JSON-parses the session's entire log. `replay_events` defaults to **0 = unlimited** (`config.rs:470`), so that log is unbounded.

Reproduced rather than asserted. Synthetic DB matching the real schema, 40k events for the session, warm cache:

```
CURRENT            SEARCH ... USING INDEX idx_acp_events_session_seq (session_id=?)
                   10.1 ms/query
+ discriminant     SEARCH ... USING INDEX idx_acp_events_session_discriminant_seq (session_id=? AND discriminant=?)
                    0.1 ms/query
```

100x, on the exact query, and it runs synchronously on the axum worker thread inside the `.map()` closure (the reconciler wraps the same call in `spawn_blocking` at `acp_reconciler.rs:715`, which is the tell). Ten structured sessions is ~100 ms of blocking SQLite per second. The fix is two clauses: `AND discriminant = 'ApprovalRequested'` on the outer query and `AND discriminant = 'ApprovalResolved'` on the subquery, replacing the `json_extract(...) IS NOT NULL` predicates. Stacking that with the worker gate from issue 1 makes this a non-event.

### Non-blocking

**4. `ApprovalGone` is silent where the precedent toasts.** `apply_structured_approval_result` (`home/mod.rs:3097`) treats `Resolved` and `Gone` identically. The structured view deliberately distinguishes them (`structured_view/mod.rs:922`): *"approval already resolved"*, with a comment explaining why the user needs to know. Worth matching, especially since issue 1 makes `Gone` reachable in the confusing case.

**5. Test gaps.** Counts and structure are proportionate to the AGENTS.md cost rules, no complaints there. But the highest-value test is missing: **`resolve_structured_approval`'s choice to `ApprovalDecisionWire` mapping** (`home/mod.rs:3023-3030`) is pure, three-line-table-shaped, and a swap there is the silent "right prompt, wrong answer" failure this whole feature has to not have. It is correct today and untested. One table test with three rows.

Also untested: the `Failed` path (nonce restored plus info dialog), the `Gone` path, and the `Disconnected` recovery, which is notable given the second commit is titled "isolate approval poller recovery".

**An e2e is owed here**, and the harness already exists. `tests/e2e/acp_focus_isolation_e2e.rs` stands up a real `aoe serve --daemon`, drives the shared Node fake ACP agent whose `permission_request` turn gates until the client decides, and attaches the native TUI over tmux. Its `APPROVAL_SCRIPT` produces exactly one deterministic pending approval. Staying on the home list instead of entering the structured view, pressing `a` then `a`, and asserting the fake's turn completes would cover the entire seam this PR adds (poller, map, dialog, HTTP, supervisor) which is the seam every finding above lives in and the only one no current test crosses. `#[parallel]`, `require_node!`, ~10-28s. That looks like well under 50 lines on top of the existing file.

## Documentation

Proportionate; nothing to trim. The doc comments in the diff all explain non-obvious *why* and the new `pending_approval_nonces` field is documented on the struct.

Two small regressions, both deletions rather than additions:

- The `open_permission_response_dialog` rationale discussed under issue 2 (`input.rs:6396`, ~8 lines) should come back, updated to say what the structured path's premise actually is.
- `// Structured rows have no pane, so the Attention sort's dead-pane tier never applies to them.` above `pane_dead: false` (`home/mod.rs:3009`) is an unrelated drive-by deletion of a "why" comment. Worth restoring.

Not owed: `docs/api.md` does not enumerate `SessionResponse` fields, and `web/src/lib/types.ts:129` is a partial hand mirror that can stay partial while nothing in the dashboard reads the field.

## Suggestions

- `structured_pending_approvals` is never GC'd against `self.instances`. A deleted structured session leaves its entry behind for the life of the TUI process. Tiny, but the codebase is fastidious about this.
- `apply_structured_approval_result`'s `Failed` arm does `nonces.insert(0, ...)`. That restores request order only because `.first()` guaranteed the answered nonce was the oldest. Correct today, coupled implicitly; one comment would pin it.
- The same arm overwrites `self.info_dialog` unconditionally, which can stomp an unrelated dialog the user is mid-read on.
- Restoring the PR template will stop the auto-close bot; it has already posted the 24-hour warning.

## Merge risk

**Moderate**, and the dominant residual is issue 1: a permanently stuck phantom approval on any structured session that was stopped, or whose worker was respawned via `spawn()` rather than `attach()`, after an approval was pending.

- **Blast radius.** Small population, but the failure is a *security affordance that looks functional*. In the `spawn()` variant it fails silently, and silent is the bad kind here. It is not a wrong-approval-approved bug, since the server's compare-and-set makes that unreachable; it is answer-goes-nowhere, which is worse than no dialog.
- **Reversibility.** Clean. Pure code, no schema change, no migration, no data backfill. `git revert` fully undoes it.
- **Coverage.** The untested path that would surface all three blocking findings is the same one: poller, `structured_pending_approvals`, dialog, HTTP, supervisor, against a live daemon. Nothing in the suite crosses it. The `Failed` and `Gone` branches, and the choice-to-wire mapping, have no test either.

Most likely way this bites in production: a user stops a structured session that had an approval on screen, then presses `a` from home and gets "Respond Failed" every single time with no path to clearing it. Nothing in the PR would surface that ahead of a user report, and CI cannot reach it. The perf regression in issue 3 would similarly show up only as a vaguely sluggish home view.

No fixes were staged: issues 1 and 2 are design calls that are the maintainer's to make (how much of the approval to project, and where the pending set is sourced from), and issue 3's one-line query change is best landed alongside whichever gate is picked.

## Verdict

**Design alignment:** Diverges, unacknowledged
**Project fit:** Accept
**Documentation:** Proportionate (two deleted "why" comments worth restoring)
**Merge risk:** Moderate; a stopped or respawned session can show a permanently unanswerable approval, silently in one of the two variants

**NEEDS WORK.** The feature belongs in the product, the async plumbing is genuinely well built, and the two details most likely to produce a wrong approval (nonce captured at open, deterministic `ORDER BY seq`) are both handled correctly. What holds it back is that the pending set is read from the durable event log rather than from the live worker, which the supervisor's own comment says is not a valid source, and that the dialog carries a bare nonce where every other approval surface in the codebase carries the tool, the target, and the destructive flag. Both come from the same decision, and widening the projection past `Vec<String>` plus gating it on a live worker fixes them together. Add the three-row decision-mapping table test and the home-view e2e on the existing `acp_focus_isolation` harness, and this is a strong contribution.

Thanks for taking this on, @athal7 — closing the "can't answer an approval without entering the session" gap is worth doing, and the nonce-capture and ordering decisions here are the two easiest things to get wrong in this area.

njbrake (commented):

_Reviewed at `2e82cfbf`. Push new commits and it will be re-reviewed._

---

Re-review at head `2e82cfbf`, rebased state verified (head is already a merge of `origin/main`; zero commits behind). All five prior findings were checked against the code, not the commit message.

### Prior findings: verdict

**1. `structured_live` gate: addressed correctly (with one residual hole, pre-existing).**
`sessions.rs:759` now reads `if structured_live && acp_worker_state == AcpWorkerState::Running`, and `spawn_inner` (`supervisor.rs:1953`) calls `cancel_orphaned_approvals` / `cancel_orphaned_elicitations` after the `workers.insert`, matching `attach`'s call site at `3212`.

Mutation-verified: reverting the gate to bare `structured_live` turns `list_sessions_projects_pending_approvals_only_for_running_workers` red ("a pending approval on a non-running worker must not be projected"). The test bites.

On the `shutdown()` path: it needs no sweep. Shutdown removes the `WorkerHandle`, `worker_states_snapshot` reports `Absent`, and the gate blocks the projection; the next `spawn`/`attach` sweeps the log. Covered.

Two residual gaps worth naming, both non-blocking:

- **The in-place crash respawn is not covered.** `restart_decision` keeps the `WorkerHandle` in the `workers` map and the respawn only swaps `handle.client` (`supervisor.rs:2515-2521`). `pending_responders` is per-`AcpClient` (`acp_client.rs:2188`), so the new client starts empty while `worker_state` stays `Running` continuously and nothing sweeps. An approval pending at the moment the agent crashes therefore survives as a projectable phantom: `a` posts, `acp.rs:2636` returns "no pending approval with nonce ...", `classify_resolve_error` mints `ApprovalGone`, the TUI shows "Already Resolved", and the next 1 Hz poll re-adds the card, forever. This is pre-existing at the durable-log level (the structured view's shelf has the same hole today), so it is not a regression this PR introduces, only a hole its new surface inherits. Follow-up, not a blocker.
- **The spawn sweep has no test at any tier.** Deleting both `cancel_orphaned_*` lines from `spawn_inner` and running the full `acp::` suite: `675 passed; 0 failed`. Given the `Running` gate is what actually prevents the user-visible symptom, the sweep is belt-and-braces log hygiene, so I would not hold the PR for a test here. Worth knowing it is unguarded.
- Micro-nit: the sweep runs *after* `workers.insert`, so there is a sub-millisecond window where the state reads `Running` with the orphan still in the log. Sync code with no `.await` between, self-healing on the next tick. Moving both calls above the `workers.insert` would close it for free, in both `spawn_inner` and `attach`.

**2. Dialog content: addressed, but the destructive marker gets clipped on ordinary inputs.**
The keystroke line is correctly branch-gated: `detail: None` (terminal) still renders "AoE sends these as raw keystrokes...", `detail: Some` renders `tool: target` plus `· destructive` in `theme.error`. Mutation-verified: deleting the `push_str("  · destructive")` turns `structured_dialog_shows_target_and_destructive_not_keystroke_hint` red.

But the header chunk is `Constraint::Length(3)` at 52 inner columns, and the structured path pushes exactly 3 lines, so it has zero slack. Any target that wraps eats the rest. Rendered probe, a `Write` to a file in this repo:

```
+- Respond to Permission Prompt -----------------------+
| aoe                                                  |
| Write:                                               |
| /Users/nbrake/scm/agent-of-empires/src/acp/superviso |
|           [Allow]   [Allow Always]   [Deny]          |
|         a=allow  A=always  d=deny  Esc=cancel        |
+------------------------------------------------------+
```

versus the short case, which is fine:

```
| Bash: rm -rf node_modules  · destructive             |
| Resolves the agent's pending permission request.     |
```

`destructive: true` in both. Ratatui wraps on whitespace, so a path longer than 52 chars burns row 2 on "Write:" and hard-breaks across row 3, pushing `· destructive` (and the guidance line) off entirely. The path is also truncated mid-token. This lands in a dialog whose default focus is `[Allow]` and where a single `a` submits Allow, and it fails precisely on the long-path and long-command inputs most likely to be genuinely destructive. The red text color is the only surviving signal.

CodeRabbit flagged a version of this ("long targets may still clip the destructive-action warning"). I agree, and it is worse than its summary suggests, since the guidance line and the tail of the path go too.

Suggested fix, aligned with the structured view's own shelf (`structured_view/render.rs:298-305`, which paints the whole card border `theme.error` when destructive):

```rust
Some(detail) => {
    if detail.destructive {
        header_lines.push(Line::from(Span::styled(
            "· destructive",
            Style::default().fg(theme.error).bold(),
        )));
    }
    header_lines.push(Line::from(Span::styled(
        detail.summary.clone(),
        Style::default().fg(summary_color),
    )));
    if !detail.destructive {
        header_lines.push(Line::from(Span::styled(
            "Resolves the agent's pending permission request.",
            Style::default().fg(theme.dimmed),
        )));
    }
}
```

Marker first means it can never be clipped; the filler guidance line yields its row to the summary in the case that needs it. Optionally also set `border_style` to `theme.error` when destructive, matching the shelf. A regression test with a target longer than 52 characters would pin it.

**3. Indexed query: addressed correctly.** Verified with `EXPLAIN QUERY PLAN` against the real schema (`events/mod.rs:878`), not by reading the SQL:

```
=== NEW (discriminant) ===
|--SEARCH acp_events USING INDEX idx_acp_events_session_discriminant_seq (session_id=? AND discriminant=?)
`--LIST SUBQUERY 1
   `--SEARCH acp_events USING INDEX idx_acp_events_session_discriminant_seq (session_id=? AND discriminant=?)

=== OLD (json_extract IS NOT NULL) ===
|--SEARCH acp_events USING INDEX idx_acp_events_session_discriminant_seq (session_id=?)
`--LIST SUBQUERY 1
   `--SEARCH acp_events USING INDEX idx_acp_events_session_discriminant_seq (session_id=?)
```

The seek narrows from `(session_id)` to `(session_id, discriminant)`, and the added `ORDER BY seq ASC` is satisfied by the index (no `USE TEMP B-TREE FOR ORDER BY`). **No migration is needed**: `discriminant` and `idx_*_session_discriminant_seq` already exist on `main` via `ensure_discriminant_column` (`events/mod.rs:190`), which is idempotent, backfills historical rows with the same first-quoted-token derivation as `discriminant_of`, and runs on every open. The PR adds no index and no compat shim, correctly.

**4. `ApprovalGone`: addressed correctly.** Now clears the card and raises an "Already Resolved" info dialog, guarded by `info_dialog.is_none()`. The message is honest: `classify_resolve_error` (`client/http.rs:725`) mints `ApprovalGone` only for a 404 whose body names *this* nonce.

**5. Test gaps: addressed correctly, both halves.**

The choice to wire mapping is now extracted to a pure `HomeView::approval_decision_wire` with a table test. Mutation-verified: swapping `Allow`/`AllowAlways` fails with `left: AllowAlways / right: Allow`.

The e2e (`tests/e2e/acp_home_approval_e2e.rs`) is real, and it does cross the seam rather than stubbing it. `#[parallel]` yes, `require_node!` plus `require_tmux!` yes, live `aoe serve --daemon` yes, shared Node fake agent whose `permission_request` genuinely blocks on `session/request_permission` (`fakeAcpAgent.mjs:382`) yes, `stop_daemon_on_drop` plus `new_in_tmp` yes. It passes in 12.56s.

Mutation-verified end to end: mangling the nonce inside `resolve_structured_approval` (`format!("{nonce}-MUTANT")`) makes it fail at `wait_for_resolved` with the durable log showing `ApprovalRequested` still unresolved. So `wait_for_resolved` proves the resolve round-tripped daemon to supervisor to worker, not just that the TUI cleared local state.

### The `a` shortcut question

**Confirmed: no, this PR adds no global TUI shortcut.** Grepping every added line for `Char('`, `KeyCode::`, `shortcut`, `keybind`, `hotkey`, `non_strict`, `strict:` across the whole diff returns exactly one hit, and it is a `cargo test -- acp_home_approval` invocation in a doc comment. `bindings.rs` is not in the 14 changed files. `ActionId::RespondToPermission` with `non_strict: &[k('a')]`, `strict: &[k('A')]`, `context: Context::Always` is verified present on `origin/main` at `bindings.rs:611`. The PR replaced the `if inst.is_structured() { return; }` no-op inside `open_permission_response_dialog` with a real branch, exactly as established at the prior head.

**On the UX consequence: it is right.** Pressing `a` on a structured row with no pending approval now shows "No Pending Approval / The daemon has no pending approval for this session." A silent no-op reads as a broken keybinding; more importantly the terminal path *already* raises an info dialog in the analogous case (agent with no `permission_response` mapping), so the new dialog makes the two paths consistent rather than introducing a new pattern. Note that this does deliberately reverse an explicit prior intent: the deleted test `structured_session_is_a_no_op` asserted "structured session no-op must be silent, not surface an info dialog". The reversal is argued in code and I agree with it, but it is a behavior reversal and should be a conscious call. In `--no-serve` TUI-only builds the silent no-op is retained via `#[cfg(not(feature = "serve"))] return;`, which is fine since structured view requires serve anyway.

### Design Alignment

**Aligned with minor deviations.**

The precedent is the native TUI structured view's own approval shelf, and the PR matches it on the decisive points. `structured_view/input.rs` has no destructive-specific handling: `a` allows in one keystroke there too (`render.rs::approval_actions_line`, "a allow once"), and destructive is signalled by color only. The 800ms long-press with the confirmation ring documented in `docs/structured-view/controls.md:27` is **web dashboard only** (`web/src/components/acp/ApprovalCard.tsx`); `destructive_require_double_confirm` has zero references in `src/`. So the home dialog is not weakening a policy the native TUI enforces. The only place it *is* weaker than the shelf is the destructive signal itself: the shelf paints the entire card border `theme.error`, the dialog colors one text line that can be clipped. That is the finding in 2 above, and coloring the dialog border is the aligned fix.

Everything else extends rather than parallels: `StructuredApprovalPoller` is built on the existing `tui::worker::Worker` substrate and mirrors `DeletionPoller` / `StopPoller` / `TrashPoller` / `DaemonStatusPoller` down to the `impl Default`; the dialog is the existing `PermissionResponseDialog` with a `detail: Option<...>` discriminator rather than a second dialog type; `pending_approval_requests` sits next to `unresolved_approval_nonces` in the same store and uses the same index.

One genuine deviation: `acp::approvals::summarize_target` duplicates `structured_view/render.rs::approval_target` (line 364), including verbatim copies of `PATH_KEYS` and `CMD_KEYS` (`render.rs:1531,1534`) and the identical `edit|write|read|delete|move` / `execute` match. The doc comment acknowledges it and explains the divergence (no path-root relativization server-side), which is the honest way to do it, but the arg-key lists are now maintained in two places and an adapter that starts emitting `"file"` or `"paths"` will be fixed in one and forgotten in the other. The clean shape is to keep the kind-to-key mapping in `acp/approvals.rs` and have `approval_target` call it and then relativize. Non-blocking.

### Compliance with Project Guidelines

- `cargo fmt --check`: clean.
- `cargo clippy --features serve,e2e-tests --all-targets`: 5 warnings, all the known-on-`main` set (`background_agent.rs`, `supervisor.rs:4042`, shifted from 4054 by this PR's +12 lines, same lint, plus `macos.rs`, `smart_rename.rs`). Nothing attributable to this PR.
- `cargo clippy --all-targets` (TUI-only, no serve): no errors, no new warnings. The `#[cfg]` split compiles both ways.
- `cargo test --features serve --lib`: **6458 passed, 0 failed**.
- `cargo test --features serve,e2e-tests --test e2e -- acp_home_approval`: **1 passed** in 12.56s.
- Test-cost rules: table tests used for the permutation cases (`summarize_target_maps_kind_to_path_or_command`, `approval_choice_maps_to_the_correct_wire_decision`); no constant/default/derive-restating tests added; the new e2e is `#[parallel]`.
- No `#[allow(dead_code)]`, no emdash or double-hyphen prose separators in authored lines, no new `include_bytes!`/`include_str!` so `flake.nix` needs nothing, no migration required.
- Rust-only diff, so the `web/tests/coverage-matrix.json` mandate and `codecov/patch` are correctly N/A.

Nits: `PendingApproval` in `sessions.rs` derives `Deserialize`, which nothing uses (the TUI deserializes into its own mirror type in `daemon_status_poller.rs`); `#[serde(default)]` on `pending_approvals` is likewise inert on a Serialize-only `SessionResponse`. Harmless, and arguably right for a wire type.

### Automated Review Check-in

CodeRabbit generated no actionable comments on the latest review and rates merge risk Low. Its one substantive note, "long targets may still clip the destructive-action warning", **I agree with, and it understates the case**: as reproduced above, the guidance line and the tail of the target go with it. That is my one blocking item. Its docstring-coverage pre-merge warning (67.44% versus 80%) I disagree with; the new code is well commented where the "why" is non-obvious and AGENTS.md explicitly discourages docstrings that restate signatures.

### CI Workflow Approval

Not acted on. The audit itself passes cleanly: the diff touches only `src/` and `tests/e2e/`, with no change to `.github/workflows/**`, `Cargo.toml`/`Cargo.lock`, `package.json`, `flake.nix`, `build.rs`, or any install/build hook, no new or bumped dependency, and no binary or obfuscated content. The fork's runs sit at `action_required` and were left there for @njbrake.

### Documentation

Proportionate in the diff; one earned update is missing. `docs/structured-view/controls.md`'s "Approvals" section (line 22) enumerates where an approval can be answered and now omits the home list. One sentence there, that pressing `a` on a structured row in the home list answers without entering the structured view, is the earned doc for a user-facing capability the PR adds. `docs/development/adding-agents.md:126` describes the `a`/`A` action as requiring per-agent keystrokes; that stays accurate for its own subject (terminal `AgentDef`s) so I would leave it, though a clause noting the ACP path needs no mapping would not hurt.

No unnecessary prose added: no standalone markdown file, no change-narration, and the comments that were added are load-bearing "why" (the `Running`-gate rationale, the nonce-captured-at-open-time note, the index rationale on the SQL). That is the good kind.

### Suggestions

1. The destructive-clipping fix above, plus a render test with a target longer than 52 characters.
2. Consider `theme.error` for the dialog border when destructive, matching the shelf.
3. Hoist `PATH_KEYS`/`CMD_KEYS` and the kind match into `acp/approvals.rs` and have `structured_view/render.rs::approval_target` call it.
4. Move the `cancel_orphaned_*` pair above `workers.insert` in both `spawn_inner` and `attach`.
5. File a follow-up for the in-place crash-respawn orphan (needs `&self` inside the drain task, so not a one-liner).

### Merge Risk

**Low to moderate.** Blast radius is confined behind `#[cfg(feature = "serve")]` and one previously-dead branch of an existing keybinding; nothing that previously worked changes path. It reverts cleanly (no schema change, no migration, no released artifact). Coverage is genuinely good for a change of this size, and I confirmed by mutation that the three load-bearing guards fail loudly when broken and that the e2e exercises the full HTTP round trip rather than the local state clear.

The one way this bites in production: a user presses `a` on a structured row, gets a dialog whose destructive warning has been clipped off by an ordinary file path, and single-keys Allow on something they would have denied. Nothing in the PR would surface that; no test asserts the destructive marker survives a long target, and the failure is silent by construction. We would hear about it from a user, if at all. The residual crash-respawn phantom is the second, much rarer one, and it is at least loud (a repeating "Already Resolved" dialog) rather than silent.

### Verdict

**Design Alignment:** Aligned with minor deviations
**Implementation:** Request Changes, narrowly, for the destructive-marker clipping
**Project Fit:** Accept
**Documentation:** Missing updates (one sentence in `docs/structured-view/controls.md`)
**CI Workflow Approval:** Withheld; audit passes, runs left at `action_required`
**Merge Risk:** Low to moderate; a clipped destructive warning on an ordinary long path, in a dialog where `a` is one keystroke to Allow

The author did what was asked. All five prior findings are genuinely addressed, and each was verified by reproduction rather than by reading the commit message: the gate by mutation, the index by `EXPLAIN QUERY PLAN`, the mapping by mutation, the e2e by nonce mutation. **The PR body matches what is pushed; no stale claim was found.** The only thing standing between this and an approve is that finding 2's fix, which was about making the dialog say what it is approving, silently drops the most safety-relevant token it added on inputs as ordinary as a path into this repo. That is a small, contained change to the render arm plus one test. With it, this is a clear ship.

njbrake (commented):

_Reviewed at `2e82cfbf`. Push new commits and it will be re-reviewed._

---

**Short follow-up to the review above: the two red Playwright (live) checks are not caused by this PR. Do not chase them.**

`composer-stop-during-thinking` and `composer-stop-during-tool-call` failed in run 33070114064, shard 1/2. I suspected this PR at first, because `main` has been green on Playwright (live) for seven consecutive runs and #3539, #3534, and #3531 all pass all 33 checks right now. That reasoning was wrong: it is a load-dependent race, so a green history is exactly what it looks like.

**Reproduced on `origin/main` with the PR absent.** A probe that fires `POST /acp/prompt` and `POST /acp/cancel` without awaiting the prompt response, reproducing the CI trace ordering at the HTTP level, then polls the durable log for a `Stopped` newer than the `UserPromptSent`:

```
PR   2e82cfbf : PROMPT=202 CANCEL=202  TURN_CLOSED_WITHIN_10S=false  promptSeq=5 stopSeqs=[4]  FAILED
main 522a7d2e : PROMPT=202 CANCEL=202  TURN_CLOSED_WITHIN_10S=false  promptSeq=5 stopSeqs=[4]  FAILED
```

Byte-identical on both binaries. The event log shows the whole thing:

```
seq 4  Stopped { reason: "cancelled" }
seq 5  UserPromptSent { text: "think about this" }
```

The daemon records `Stopped{cancelled}` for a turn that has not opened yet, then `UserPromptSent` opens the turn (`state.rs:1331` sets `turn_active = true`) and nothing ever closes it.

**Why CI hit it.** From the traces, the cancel was issued 102 to 105 ms after the prompt POST began, while that POST was still in flight (it took 125.6 ms and 258.9 ms on a loaded runner, against a few ms normally). The Stop button renders optimistically on send, so Playwright clicks it the instant it appears and beats the prompt to the agent. `fakeAcpAgent.mjs`'s `session/prompt` handler then does `cancelFlags.set(sessionId, false)` to reset any prior flag, wiping the early cancel, and the scripted `wait_ms: 30_000` runs to completion.

**Why it is not this PR specifically.** The only PR code that executes in these two specs is `pending_approval_requests` on `GET /api/sessions`, which the dashboard polls at 3000 ms, so roughly seven extra zero-row index seeks over a 20 second test. `pending_approvals` carries `skip_serializing_if = "Vec::is_empty"`, so with no pending approvals the HTTP response is byte-identical to main and the client cannot observe a difference. The `spawn_inner` sweep is a strict no-op here: both `cancel_orphaned_*` early-return on `stale_nonces.is_empty()`, and neither script emits a permission request or an elicitation. And `composer-stop-mid-turn` and `composer-stop-during-sub-agent`, near-identical specs in the same shard, both passed in the same run.

Locally the full CI-faithful shard 1 passes on this PR head (74 passed, 3 skipped, 0 failed), as does `--repeat-each=8` at 10 workers under concurrent build load.

A re-run of shard 1 should clear these. My verdict is unchanged: Request Changes solely for the destructive-marker clipping described above.

**Two things worth a separate issue, @njbrake, not this PR's problem:**

1. **Product.** `POST /acp/cancel` returns 202 and writes `Stopped{cancelled}` even with no turn in flight, so a cancel that beats its prompt is silently swallowed and the next turn cannot be stopped until it ends on its own. That is #2237 still live, now reachable through the optimistic Stop button rather than a fast finger. The durable fix is to either reject a cancel with no turn in flight so the client can retry, or latch it so it applies to the turn that opens next.
2. **Test.** These specs treat "Stop is visible" as "the daemon has a turn", which the optimistic button does not guarantee. Gating the click on the turn's first server-driven artifact, or on `page.waitForResponse` for the prompt's 202, closes the window without weakening the assertion. `playwright.live.config.ts` already documents this failure class rotating across specs run to run, and the suite runs `retries: 0`.

src/acp/acp_client/connection.rs
```diff
@@ -1911,6 +1911,10 @@ pub(super) async fn run_connection_task<W, R>(
 
                         loop {
                             tokio::select! {
+                                // Send the prompt before a queued Cancel. A
+                                // notification delivered first is ignored by agents
+                                // that have not yet started the prompt it should stop.
+                                biased;
```
jerome-benoit, line 1922: [P2] Preserve cancellation fairness

`biased;` prioritizes lifecycle notifications (line 1976) over Cancel/ForceStop (line 2207) and the cancellation deadline (line 2087) on every iteration. Sustained notification traffic can therefore delay those branches while the prompt remains pending, matching Tokio's documented fairness caveat for streams ahead of shutdown. Preserve prompt dispatch ordering without imposing permanent priority over cancellation, and cover cancellation while notifications remain queued. This is a static finding, not a reproduced hang.

src/tui/home/status.rs
```diff
@@ -252,6 +252,14 @@ impl HomeView {
...
                     self.apply_daemon_status_update(update);
                 }
+                // Drop pending-approval entries for sessions that no longer
+                // exist locally (deleted/trashed rows the daemon stops
+                // listing, so `apply_daemon_status_update` never revisits
+                // them). Otherwise a removed session leaks its entry for the
+                // life of the TUI process.
+                let instances = &self.instances;
+                self.structured_pending_approvals
+                    .retain(|id, _| instances.contains_key(id));
```
jerome-benoit: [P2] Clear cached approvals for archived and trashed sessions

Archived and trashed sessions remain in `instances` (`operations.rs:1933,2040`), so this predicate retains their cached approvals. Subsequent daemon updates return at lines 311-312 before clearing that cache because `daemon_status_applies_to` excludes both states. Selecting such a row and invoking the permission action can still open the obsolete approval because `open_permission_response_dialog` only guards Creating/Deleting. Clear cached approvals when a session becomes archived or trashed and guard the action for those states; add a regression case that applies an empty daemon update after the transition.

tests/e2e/acp_home_approval_e2e.rs
```diff
@@ -0,0 +1,251 @@
...
+//! resolves it, and the turn only completes (recording `ApprovalResolved`)
+//! once the resolve round-trips through the daemon.
+//!
+//! Compiled only with the `web` feature (structured view +
+//! `aoe add --structured-view` don't exist otherwise). Run via:
+//!
+//! ```sh
+//! cargo test --features e2e-tests --test e2e -- acp_home_approval
+//! ```
+#![cfg(feature = "web")]
```
jerome-benoit, line 25: [P2] Run the approval E2E without the dashboard feature

This gate excludes the entire module from the documented `cargo test --features e2e-tests --test e2e -- acp_home_approval` invocation. `Cargo.toml:197-221` now makes ACP and the structured TUI core functionality; `web` only adds the dashboard. Remove this gate and correct the preceding explanation so the documented command actually exercises the new test without requiring an unrelated dashboard build.

jerome-benoit (changes requested):
Reviewed at 72fc38ca3e2aab0fc15115f27c447d05c15a3435 with two independent reviews and cross-validation. Please address the three P2 findings below. The previously reported destructive-marker clipping and crash-respawn orphan handling remain relevant; I have not duplicated those comments.

Validation: static inspection of the exact diff and affected call paths, plus git diff --check (clean). No local build or tests were run because of host memory pressure. The only successful check currently exposed for this SHA is PR-title lint; earlier test results do not validate this revision.

src/tui/approval_poller.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub resolution: ApprovalResolution,
+}
+
+async fn resolve(request: ApprovalRequest) -> ApprovalResult {
+    let session_id = request.session_id;
+    let nonce = request.nonce;
+    let resolution = match require_daemon().await {
+        Ok(endpoint) => match HttpClient::new(endpoint) {
+            Ok(client) => match client
+                .resolve_approval(&session_id, &nonce, request.decision, None)
```
jerome-benoit, line 37: [P2] Keep answer lists out of the generic Allow dialog

The home projection drops `options`/`choice`, but main now supports questions such as pi's `ask_user_question`, represented by several `allow_once` options (`src/acp/approvals.rs:94–110`). This call always sends `option_id: None`; `resolve_permission` forwards it without a choice-list guard (`acp_client/mod.rs:859–880`), and `pick_option_id` selects the first matching kind (`permission_handlers.rs:90–92`). Consequently, pressing Allow answers the question with its first option even though the home dialog never displayed the choices.

Please exclude choice lists from this generic approval flow and direct the user to the structured view, or carry the labels and explicitly selected option ID through it. A projection/routing regression test with two same-kind options should ensure Allow cannot silently pick the first. This is an integration defect in the new home consumer, established by exact-head source tracing, not a runtime reproduction.

jerome-benoit (changes requested):
Re-reviewed `cc8d7a0d` against `8f0cece8`, including existing discussions and two sequential independent analyses. The wire-type relocation is structurally consistent, but the new home path does not preserve the choice-list contract now present in main; one new P2 is inline and was cross-validated against the resolver.

The existing P2 requests remain required, without duplicate annotations: [cancellation fairness](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256378), [archived/trashed approval-cache invalidation](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256384), and [the E2E feature gate](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256388). The previously raised destructive-marker clipping and crash-respawn orphan handling also remain unresolved. Fixed nonce strings in test fixtures are not cryptographic-secret findings.

Static inspection only; `git diff --check 8f0cece8 cc8d7a0d` is clean. No local build/test or runtime reproduction. CI/Tests/CodeQL were awaiting approval, not validated by the green title check; I approved those workflows.

jerome-benoit (changes requested):
Re-reviewed `1623b4d8f1529a72edd8c9011ff2a56764dac4de`, including the main merge and existing discussions. The session-feed integration carries approvals through, but these existing P2 requests still require changes:

- [Choice lists](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3962054240): show/select the offered answer or route to the structured view; the home resolver still sends no option ID.
- [Archived/trashed cache](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256384): clear these entries and guard the action. The thread is outdated, but the defect remains at `home/status.rs:301-302,355-362` and `home/input.rs:6517`.
- [Cancellation fairness](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256378): remove the notification-starvation risk from the added `biased` select.
- [E2E feature gate](https://github.com/agent-of-empires/agent-of-empires/pull/3544#discussion_r3937256388): exercise this core-TUI feature without requiring `web`; this is a coverage defect, not a demonstrated runtime failure.
- [Destructive-marker clipping](https://github.com/agent-of-empires/agent-of-empires/pull/3544#pullrequestreview-5042319049): keep the warning visible for long targets (`dialogs/permission_response.rs:183,203-225`).

Scope clarification: the crash-respawn orphan issue also exists on the current base. It remains relevant follow-up work, but is not an independent blocker for this PR. The poller-flag defect is fixed; constant test nonces are not cryptographic-secret findings. No duplicate inline comments.

Two independent sequential static reviews were cross-validated against exact-SHA source and the base. CI is green at this SHA. No local tests or builds run.

src/acp/supervisor.rs
```diff
@@ -1770,6 +1770,18 @@ impl<S: BroadcastSink> Supervisor<S> {
...
+        // Runner starts with an empty `pending_responders`, so any
+        // `ApprovalRequested` in the log with no matching `ApprovalResolved`
+        // is orphaned: its responder oneshot died with the previous daemon or
+        // worker. Publish the synthetic `ApprovalResolved { Cancelled }` now so
+        // it never resurfaces as a dead 404 card, and so the home TUI's
+        // `running`-gated pending-approval projection only ever sees approvals
+        // this live worker can actually resolve. See #1099.
+        self.cancel_orphaned_approvals(&session_id);
+        self.cancel_orphaned_elicitations(&session_id);
+
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in #3867: `start_drain_task` runs both orphan sweeps immediately after the respawn installs the replacement client. The sweep bodies moved to free helpers (`cancel_orphaned_approvals_on` / `cancel_orphaned_elicitations_on`) compatible with the task's owned state and shared with the spawn/attach paths, so stale approval and elicitation cards now receive cancellation events after a crash respawn.

src/tui/home/tests/permission_response_dialog.rs
```diff
@@ -68,19 +79,116 @@
...
-fn structured_session_is_a_no_op() {
+fn structured_session_reuses_the_permission_dialog() {
     let mut env = create_test_env_empty();
     let id = add_session_with_tool(&mut env.view, "session-one", "claude");
     env.view
         .mutate_instance(&id, |inst| inst.view = crate::session::View::Structured);
-    env.view.selected_session = Some(id);
+    env.view
+        .structured_pending_approvals
+        .insert(id.clone(), pending("approval-1"));
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in #3867: fixture nonces are now generated at runtime (uuid), so no string literal flows into a `nonce` field and rust/hard-coded-cryptographic-value no longer fires on this file.

src/tui/home/tests/permission_response_dialog.rs
```diff
@@ -68,19 +79,116 @@
...
+}
+
+#[test]
+#[serial]
+fn resolved_structured_approval_clears_a_nonce_reintroduced_by_polling() {
+    let mut env = create_test_env_empty();
+    let id = add_session_with_tool(&mut env.view, "session-one", "claude");
+    env.view
+        .structured_pending_approvals
+        .insert(id.clone(), pending("approval-1"));
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in #3867: same runtime-derived fixture nonce; the literal no longer reaches the `ApprovalResult.nonce` field.

src/tui/home/tests/permission_response_dialog.rs
```diff
@@ -68,19 +79,116 @@
...
+}
+
+#[test]
+#[serial]
+fn gone_structured_approval_clears_and_informs() {
+    let mut env = create_test_env_empty();
+    let id = add_session_with_tool(&mut env.view, "session-one", "claude");
+    env.view
+        .structured_pending_approvals
+        .insert(id.clone(), pending("approval-1"));
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in #3867: same runtime-derived fixture nonce across the file, covering all three alert sites (89, 112, 154).

src/acp/acp_client/connection.rs
jerome-benoit: The production select priority is fixed in #3867: prompt completion first, then commands and escalation, then lifecycle notifications. Correction to my earlier test disposition: cancel_reaches_the_agent_while_notifications_remain_queued proves eventual command delivery during a flood, not the requested first-selection ordering; it was inaccurate to say no assertion could expose the old order. #3881 at 2f2fd798 now gates one connection with both queues ready and observes its first selected arm through a task-local probe, exercising two concurrent connections across 32 rounds. Removing biased; failed locally; the restored ACP client suite passed 235 tests. The unbiased mutation check remains probabilistic, while readiness and observer ownership are controlled.

src/tui/home/status.rs
jerome-benoit: Fixed in #3867: the daemon refresh returns early for archived/trashed rows, so the transition itself now drops the cached approvals (`apply_daemon_status_update`), and `open_permission_response_dialog` guards both states. Regression: `daemon_update_clears_cached_approvals_when_a_row_is_sunk` applies an empty daemon update after the transition and asserts the cache is gone.

tests/e2e/acp_home_approval_e2e.rs
jerome-benoit: Fixed in #3867: the `#![cfg(feature = \"web\")]" gate is removed. ACP and the structured TUI are core (`Cargo.toml` features); the fake agent is a Node script read from `web/tests/helpers/` at runtime, not a build dependency. The module doc now states the plain e2e invocation and the auto-skip behavior.

src/tui/approval_poller.rs
jerome-benoit: Fixed in #3867 (first remedy): the resolver cancels a choice list answered without a picked option id — by-kind selection would send the agent's first `allow_once` as the user's answer — and the daemon projection carries the `choice` flag so the home dialog routes choice-list requests to the structured view with an info dialog instead of the generic dialog. Regressions: handler-level `choice_list_with_no_picked_option_cancels_instead_of_picking_the_first` (two same-kind options, Allow without an id → Cancelled, never the first pick) and home routing `choice_list_approval_is_routed_to_the_structured_view`.

PR #3547, opened 2026-08-26
fix(tui): align split preview cursor with pane origin

jerome-benoit:
## Description

Follow-up to merged PR #3526 and issue #3515.

The native TUI preview consumes the same composited `PaneCursor` geometry as the web surface, but `map_live_preview_cursor` still indexed the window grid with pane-relative `x`/`y`. With `pane-border-status top`, the bottom-anchor delta clipped the reserved row while the missing pane origin left the hardware cursor one row above its painted prompt.

Add `composite_pane0.left/top` after the bottom-anchor delta. The unsplit case remains an identity because `composite_pane0` is `None`. Mouse geometry is intentionally unchanged: the TUI viewport clips the reserved top row before input mapping, so adding `top` there would move input below the displayed pane.

Also move the pane-offset forwarding permutation from a second mocked Playwright page load into the existing component Vitest, retain the browser baseline for DOM-to-WebSocket wiring, make `PaneCursor::composite_pane0` the canonical coordinate contract, and correct the `pane0: null` protocol wording.

Actual TUI verification after the fix:

- tmux source: `source_cursor=32,4 pane0=0,1,41,33 window=81,34`
- native TUI: hardware cursor row 9; painted prompt row 9
- before the fix, the same geometry placed the hardware cursor on row 8

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

No screenshot or recording was produced. The actual TUI was launched in an isolated outer tmux session and verified by exact tmux source, hardware-cursor, and captured prompt coordinates above; the isolated session and scratch state were then removed.

## Test Coverage Analysis

The web production behavior is unchanged. Its pane-offset wiring assertion moved to the existing component Vitest; the existing mocked Playwright baseline still covers real DOM row measurement through WebSocket bytes. Existing coverage-matrix entries already list both surfaces, and matrix validation passes.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the pure cursor mapper is the minimal tier that directly fails when the origin addition is removed; an actual TUI smoke test verifies the integrated surface.

Validation:

- `cargo test --features serve`: 6692 passed, 8 ignored
- focused TUI cursor tests: 4 passed
- `npm run format:check` and `npm run lint`
- targeted Vitest: 53 passed
- full Vitest with the suite's expected `en_US.UTF-8` locale: 3713 passed
- rebuilt mocked Playwright click-forward spec: 6 passed

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** openai-codex/gpt-5.6-sol via Oh My Pi

**Any Additional AI Details you'd like to share:** Two independent validators reconstructed every published #3526 review finding. Both reproduced the TUI cursor offset on the actual surface; they independently disproved the proposed simple mouse-origin change because viewport clipping cancels that offset. A separate reviewer cross-validated the completed diff before commit.

## Summary

- Fixed cursor alignment in split native TUI previews.
- Preserved correct behavior for single-pane previews.
- Clarified pane origins and coordinate handling across Rust and web components.
- Added component test coverage for pane-origin and vertical-offset behavior.
- Removed a redundant Playwright regression test.
- Updated protocol and geometry documentation for compatibility and clarity.

## Benefits

The hardware cursor now matches the painted prompt when the top pane border is enabled. Mouse behavior remains unchanged. The updated tests and documentation make coordinate handling easier to maintain.

### DIFF
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -21,15 +21,12 @@
 //!   `#{mouse_sgr_flag}`: when the pane is a full-screen mouse app the
 //!   client forwards the wheel to it (as input bytes) instead of widening
 //!   the capture window, since the alternate screen has no scrollback.
-//!   On a composited (split) window the frame also carries `"pane0"`:
-//!   `{"cols":..,"rows":..,"left":..,"top":..}`, pane 0's rectangle within
-//!   the window grid. The cursor is translated onto that grid before it is
-//!   sent, and clients map pointer cells back by subtracting
-//!   `(left, top)` (#3515). Absent on single-pane frames. Clients that
-//!   ignore `left`/`top` still paint the cursor correctly at any origin,
-//!   because the translation already happened here; their mouse forwarding
-//!   is what degrades, back to the one-row shift of #3515 whenever the
-//!   origin is not `(0, 0)`.
+//!   A composited frame also carries `"pane0"` as
+//!   `{"cols":..,"rows":..,"left":..,"top":..}`. Cursor coordinates are
+//!   translated onto the window grid before emission; clients subtract the
+//!   same origin when forwarding pointer cells. Single-pane frames serialize
+//!   `pane0` as `null`. The origin fields are optional for older clients and
+//!   default to zero for frames from older servers.
 //!   `{"type":"size_owner","is_owner":bool}`: whether this client holds
 //!     the session's size-owner lock. Only the owner resizes the shared
 //!     tmux window and may type; a non-owner renders best-effort at the
@@ -1036,13 +1033,9 @@ async fn handle_live_ws(
 /// virtual scroll spacer off `history` and slices the live screen off
 /// the content's last `rows` lines, independent of cursor visibility.
 fn frame_json(content: &str, cursor: Option<&crate::tmux::PaneCursor>) -> String {
-    // On a composited frame the content grid is the whole WINDOW while
-    // `x`/`y` are pane 0 relative; the client paints the cursor by indexing
-    // that grid directly, so translate onto the composite here (#3515).
-    // `composite_pane0` is `None` (or at the origin) on every other frame,
-    // where this is the identity. The origin also rides on `pane0` so the
-    // client's inverse mapping, pointer cell -> pane coordinate, can
-    // subtract it again.
+    // The cursor is pane relative while composited content uses the window
+    // grid. Emit window-relative coordinates and carry the same origin for
+    // the client's inverse pointer mapping. No pane rectangle means identity.
     let pane0 = cursor.and_then(|c| c.composite_pane0);
     let (origin_x, origin_y) = pane0.map_or((0, 0), |p| (p.left, p.top));
     let cursor_value = match cursor {
@@ -1163,7 +1156,7 @@ mod tests {
     #[test]
     fn frame_json_includes_geometry_and_cursor() {
         let cases = [
-            // Unsplit: `pane0` is absent and the cursor is untouched.
+            // Unsplit: `pane0` is null and the cursor is untouched.
             (None, (3, 7), serde_json::Value::Null),
             // Composited with pane 0 at the corner (a borderless split):
             // identity translation, but `pane0` rides with zero origin.
@@ -1182,8 +1175,8 @@ mod tests {
                     "top": 0,
                 }),
             ),
-            // Composited with pane-border-status top: the wire cursor moves
-            // onto the window grid by pane 0's origin (#3515).
+            // Composited with pane-border-status top: move the wire cursor
+            // onto the window grid by pane 0's origin.
             (
                 Some(crate::tmux::PaneGeom {
                     left: 2,
--- a/src/tmux/composite.rs
+++ b/src/tmux/composite.rs
@@ -11,17 +11,15 @@
 //! Compositing is read-only and changes nothing about input routing, which
 //! stays pinned to `^.0` (#435, #488).
 //!
-//! Known residual: a window row covered by NO pane (a `pane-border-status
+//! Known residual: a window row covered by no pane (a `pane-border-status
 //! top` status line, which tmux draws itself and no capture sees) is filled
-//! with a full-width border rule. Translating the cursor (#3515) does not
-//! remove that row; dropping it would mean rendering a grid shorter than
-//! `window_height` and rebasing every consumer, so the rule stays.
+//! with a full-width border rule. Cursor translation does not remove that row;
+//! dropping it would render fewer than `window_height` rows and rebase every
+//! consumer, so the rule stays.
 
 /// One pane's rectangle within its window, from
-/// `#{pane_left} #{pane_top} #{pane_width} #{pane_height}`. Pub because
-/// pane 0's copy rides on the public [`crate::tmux::PaneCursor::
-/// composite_pane0`] so composited frames can translate between pane
-/// coordinates and the window grid (#3515).
+/// `#{pane_left} #{pane_top} #{pane_width} #{pane_height}`. Public because
+/// [`crate::tmux::PaneCursor::composite_pane0`] exposes it.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct PaneGeom {
     pub left: u16,
@@ -96,16 +94,10 @@ impl WindowLayout {
         composite_window(self.window_width, self.window_height, &self.panes)
     }
 
-    /// The first pane's rectangle. Pane indices follow layout order, so
-    /// index 0 is the top-left pane and closing it renumbers whichever pane
-    /// takes that corner; the composite therefore paints pane 0's rows at
-    /// `geom.top` and its columns at `geom.left`, which is `(0, 0)` for the
-    /// common layouts but NOT always: a window option like
-    /// `pane-border-status top` reserves a border row above every pane,
-    /// shifting pane 0 down to `top == 1` (#3515). Consumers of the cursor
-    /// must translate by the origin carried on
-    /// [`crate::tmux::PaneCursor::composite_pane0`] rather than assuming the
-    /// window origin.
+    /// The first pane's rectangle. Pane indices follow layout order, but the
+    /// top-left pane can still have a non-zero origin when window chrome
+    /// reserves rows or columns. Keep that origin as geometry, not an assumed
+    /// `(0, 0)`.
     pub(crate) fn first_pane(&self) -> Option<PaneGeom> {
         self.panes.first().map(|p| p.geom)
     }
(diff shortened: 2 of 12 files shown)


njbrake (commented):

_Reviewed at `3a235c95`. Push new commits and it will be re-reviewed._

---

## Summary

Adds `composite_pane0`'s origin to `map_live_preview_cursor` so the native TUI preview paints pane 0's caret on the right row of a composited window: the TUI half of #3515 that #3526 disclosed but left unfixed. It also consolidates the #3515 prose onto `PaneCursor::composite_pane0` as the canonical contract, and moves the web pane-offset forwarding assertion out of mocked Playwright into the existing component Vitest.

**Disclosure:** this reviewer also reviewed #3526 and wrote #3543, which njbrake closed in favour of this PR. This has been held to the same standard as any other change, and where the two differ the difference was measured rather than argued. One of those measurements went against #3543, and it leads.

## Design alignment

**Aligned.**

The cursor fix extends the existing mechanism exactly as #3526 set it up: the origin was already carried on `composite_pane0`, and this reads it at the one consumer that was still ignoring it. No new type, no new field, no parallel path.

More than that, the formula is the *general* one, which took verifying. `map_live_preview_cursor` computes `(anchor - cursor.pane_height)` where `anchor = min(line_count, visible_rows)` and, on a composite, `cursor.pane_height` is rebased to `window_height == line_count`. That term is therefore identically `-max(0, line_count - visible_rows)`, which is exactly `-compute_scroll(...)`, i.e. minus `first_line`. `src/tui/home/mod.rs:135` states the paint contract as "screen row `pane.y + k` shows content line `first_line + k`". So the new formula reduces to `pane.y + (cursor.y + top) - first_line`, which is the correct mapping for **every** layout, not just the one that was reproduced. That is a better result than a fix tuned to the reported geometry.

## The load-bearing question: does this undo #3526?

**No. Not one line of executable code outside `src/tui/` changed.**

Every hunk in `src/server/live_ws.rs`, `src/tmux/composite.rs`, `src/tmux/session.rs`, `web/src/lib/liveMouse.ts`, `web/src/hooks/useLiveTerminal.ts`, and `web/src/components/MobileLiveTerminal.tsx` was read. Every deletion in those files is a doc comment, an inline comment, a test comment, or an assertion message. `pointerPaneCell` and `cursorLineIndex` have identical bodies to `main`; `frame_json`'s translation is untouched. The net-negative line count is prose, not logic.

All four #3526 mutation checks were re-run against this head. All still fail as they should:

| Mutation | Result on `3a235c95` |
| --- | --- |
| `frame_json` cursor translation reverted | `frame_json_includes_geometry_and_cursor` fails, `left: Number(3), right: 5` |
| `composite_pane0` forced to `top: 0` | tmux test fails against real tmux, `left: (0,0,40,23), right: (0,1,40,23)` |
| `pointerPaneCell` origin subtraction reverted | `liveMouse.test.ts` fails, `{col:10,row:6}` vs `{col:10,row:5}` |
| Same revert, jsdom component test | **also** fails, `expected 6 to be 5` |

The unsplit path still produces an identity translation and is still pinned: the `None` row of `frame_json_includes_geometry_and_cursor`, `pointerPaneCell`'s identity rows, the retained Playwright "bottom-aligned row" case (`;2M`), and the single-pane assertion kept as the first statement of the rewritten render test. The server-translates / client-subtracts asymmetry is preserved unchanged.

**The Playwright `-22` is a demotion, not a deletion.** The case moved into `MobileLiveTerminal.forward.test.tsx` as a `baseRow` vs `offsetRow` comparison asserting `offsetRow === baseRow - 1`. That is the exact pattern probed and recommended in the #3526 review, and mutation row 4 above confirms it genuinely catches the revert. The browser baseline for real DOM row measurement through to WebSocket bytes is retained in the sibling test. Trading a ~2.5s browser case for a ~10ms jsdom one while keeping the assertion is a straight win, and it is responsive to a review comment rather than a quiet removal.

## The TUI mouse path: #3543 was wrong and this PR is right

This is the finding this reviewer was uniquely positioned to check, and it went against #3543.

#3543 changed `mouse_pane_rect` to start at `pane.y + rect.top`. #3547 deliberately does not, and its comment says viewport clipping already cancels that offset. Rather than take that on trust, the sizing path was read and then measured.

`finalize_live_send_resize` (`src/tui/home/mod.rs:6443`) calls `session.resize_window(width, preview_pane_area.height)`, and `resize_window` (`src/tmux/session.rs:1450`) sizes the **window** to `rows + chrome_rows(...)`, where `chrome_rows = window_height - pane0_height` when that delta is `<= MAX_CHROME_ROWS` (5). With `pane-border-status top` on a side-by-side split the delta is exactly 1, so the window is sized to `visible_rows + 1`, the composite is one row taller than the viewport, and `compute_scroll` clips the border row off the top. `first_line` therefore equals `top`.

A round-trip probe was built on this head using the reporter's own geometry from #3515 (`pane0=0,1,41,33`, `window=81,34`): paint the caret with `map_live_preview_cursor`, then feed that same screen row back through `mouse_pane_rect` plus `map_pane_cell` and check it reports `cursor.y + 1`.

```
first_line = 1  (pane0.top = 1)
ground-truth row = 8, painted row = 8
mouse rect = Rect { x: 10, y: 4, width: 41, height: 33 }, app_row = 5, want = 5
```

Round-trips correctly with the mouse path untouched. Then #3543's change was applied on top of this head and the identical probe re-run:

```
mouse rect = Rect { x: 10, y: 5, width: 41, height: 32 }, app_row = 4, want = 5
panicked: clicking the caret's own cell must report the app its own row
  left: 4
 right: 5
```

So #3543's mouse change would have introduced a one-row click offset in exactly the geometry #3515 reports. The author's two validators reached the same conclusion independently, and they were right. It follows that #3543's "extent clamp" finding was moot: the overflow only existed as a consequence of moving the origin, which should not have been moved. Nothing pre-existing needs that clamp.

Closing #3543 in favour of this PR was the correct call.

## Issues found

Nothing blocking. Two non-blocking items, both about the mouse decision rather than the code.

**1. The most important decision in this PR is documented but not pinned by a test.** Leaving `mouse_pane_rect` alone is correct, and it is correct for a non-obvious reason (`first_line` cancels `top`). No committed test asserts that. A future contributor reading `composite_pane0`'s doc, which now says consumers "subtract it from pointer coordinates", will reasonably try the change #3543 tried, and the full suite will stay green while TUI clicks land a row off. The round-trip probe above is about 25 lines and fails loudly on exactly that edit; it belongs in `input.rs`'s test module. This is the single highest-value thing to add.

**2. The cancellation is not universal, and the PR's own comment says so without pinning it.** `chrome_rows` resolves a delta above `MAX_CHROME_ROWS` to 0, which is what a stacked split produces. The same probe with that geometry (`first_line = 0`, `top = 1`):

```
first_line = 0  (pane0.top = 1)
ground-truth row = 9, painted row = 9
mouse rect = Rect { x: 10, y: 4, width: 81, height: 15 }, app_row = 6, want = 5
```

The cursor is correct (this fix is general), but the click reports one row too far down. This is pre-existing and unchanged by the PR, and the comment at `input.rs:205` names the missing ingredient precisely: "Mapping a non-bottom-aligned slice would additionally need that slice's first visible row." Not worth expanding this PR to fix. It deserves a follow-up issue, because #3547 narrows the residual (cursor now right everywhere) without closing it, and the invariant is now load-bearing for a decision rather than incidental.

Same shape horizontally: the cursor adds `left` but the mouse never subtracts it, and there is no horizontal clip to cancel it. Latent only, since pane 0 is the top-left pane so `pane_left` is structurally 0. Worth one clause in the comment.

## Documentation

**Proportionate, and a net improvement.** The #3526 review counted roughly fifteen restatements of the #3515 story and recommended one canonical home with the satellites reduced to the contract. That is what this does, and the canonical home gained a sentence that is genuinely new and load-bearing: "A cropped preview must also account for rows removed before mapping." That is the `first_line` insight, recorded where a consumer will find it. It also fixes nit 5 from the #3526 review, the test comment that said `pane0` was "absent" when the wire carries `null`.

One thing worth keeping. The old `live_ws.rs` module doc explained *why* the cursor is translated server-side: a client that ignores `left` / `top` still paints correctly because translation already happened at emission. The new wording states the contract but drops that rationale, and it is the one genuinely non-obvious "why" in the design. Losing it makes it easier for someone to later "simplify" by moving translation clientward and silently break older clients. One sentence, restored at the canonical home rather than in the module doc, would cover it.

## Automated review check-in

- **CodeRabbit, no actionable comments, 5/5 pre-merge checks:** agree, this read found no defect either.
- All CI checks green on `3a235c95`.

## Verification run

Rebased on `origin/main` (`5988e958`), own worktree, `env -u CLAUDE_CONFIG_DIR`.

- `cargo fmt --all -- --check` clean.
- `cargo clippy --features serve -- -D warnings` clean. `--all-targets` reports the same five pre-existing sites (`src/acp/supervisor.rs`, `src/acp/background_agent.rs` twice, `src/session/smart_rename.rs`, `src/process/macos.rs`), none in `src/tui/`, so nothing new.
- `cargo test --features serve`: **exit 0, 6412 lib tests, all 24 result lines ok, zero failures anywhere.** Net test-fn count unchanged from main, since the new assertions were folded into the existing test rather than added as a new fn.
- `npm run format:check`, `npm run lint` clean. Full Vitest **3713 passed / 353 files**, identical to main.
- `node tests/validate-coverage-matrix.mjs` passes. No matrix change is needed and none was made: the spec file still exists with one fewer case, so `terminal.mobile-live-click-forward-e2e` still resolves, and the Vitest side was already registered under `terminal.mobile-live-click-forward`. The PR's claim here is accurate.
- Playwright `tests/live-click-forward.spec.ts` **6/6 pass after a `npm run build`** (rebuilt deliberately; `vite preview` serves `dist/`, so a source-only check tests a stale bundle).
- Mutation check on this PR's own fix: reverting `+ top` / `+ left` fails `live_cursor_maps_single_and_composited_origins` on both axes (`Position { x: 43, y: 6 }` vs `{ x: 44, y: 7 }`), while the three other `live_cursor_*` anchoring guards stay green.

## Merge risk

**Low.** The change is one added term in one pure function, plus comments and a test move. Blast radius is the composited TUI preview only; unsplit previews carry `composite_pane0: None` and are provably identity. It reverts cleanly, ships no schema or persisted state, and the web surface is untouched so #3526 cannot regress.

The residual risk is item 1: the decision not to touch the mouse path is correct but unguarded, so the most likely way this bites is a well-intentioned future PR making the change #3543 made, with a green suite. That would arrive as a user report, not from CI.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Documentation:** Proportionate (restore one sentence of "why" on the server-side translation)
**Merge risk:** Low; the correct decision not to touch the mouse path is documented but not test-pinned

**LGTM.** This is the stronger change: the cursor formula it lands is general rather than tuned to the reproduced geometry, the web fix is fully intact with all four #3526 mutation guards still firing, the Playwright deletion is a verified demotion to jsdom rather than lost coverage, and the author's refusal to touch the TUI mouse path is correct for a reason #3543 got wrong and which was confirmed here by measurement. Merge it, and file a follow-up for the `first_line != top` residual plus the round-trip test that would pin the mouse decision.

Issue #3550, opened 2026-08-26
TUI composited preview: project pointer input through the visible pane slice


jerome-benoit:
## Problem

PR #3547 correctly leaves `mouse_pane_rect` at the visible output origin for the reproduced side-by-side `pane-border-status top` layout. In that layout the preview bottom-follows a composite one row taller than its viewport, so `first_line == pane0.top == 1`; clipping cancels the pane origin before mouse mapping.

That cancellation is not universal:

- A stacked split can report `pane0.top == 1` while the pane/window chrome delta exceeds `MAX_CHROME_ROWS`, leaving `first_line == 0`. The current mouse rect then forwards a displayed row one too large.
- tmux pane indexes follow creation order, while `rotate-window` and `swap-pane` can move pane 0 away from the left edge. With `pane0.left > 0` and no horizontal clip, the current mouse rect forwards the wrong column.

This is the pointer-side residual of #3515, identified and reproduced during review of #3547. It predates that PR; #3547 fixes cursor painting and adds a guard proving that the common `top == first_line` layout must keep its current mouse origin.

## Required design

Project pane 0 through the actual visible composite slice rather than adding `left`/`top` directly:

```text
screen_x = output.x + pane0.left - first_visible_col
screen_y = output.y + pane0.top  - first_visible_row
```

Intersect the resulting rectangle with the preview viewport before using it both as the origin for `map_pane_cell` and as the clamp extent. Preserve the existing bottom-follow cancellation when `pane0.top == first_visible_row`.

## Acceptance criteria

- Extend the existing composited mouse mapping test as table cases, without adding one test function per layout.
- Cover side-by-side top-border cancellation, stacked `top != first_line`, and rotated/swapped `left != first_visible_col`.
- A click on the painted cursor cell round-trips to the cursor's 1-based app cell in each case.
- Verify the affected layouts on the actual native TUI.

References: #3515, #3526, #3547.

PR #3551, opened 2026-08-26
test(tui): guard crop-cancelled mouse projection

jerome-benoit:
## Description

Post-merge follow-up to the maintainer review on #3547 and the TUI pointer residual from #3515.

The common side-by-side `pane-border-status top` layout bottom-follows a composite one row taller than the preview. Its `first_line` therefore equals `pane0.top`, so clipping already cancels the pane origin before mouse mapping. #3547 correctly left `mouse_pane_rect` at the visible output origin, but no committed test protected that non-obvious decision.

Extend the existing composited mouse test with a render-to-input round trip: paint a pane-relative cursor through `map_live_preview_cursor`, feed that screen cell through `mouse_pane_rect` and `map_pane_cell`, and require the same 1-based app cell. A naive `pane.y + rect.top` mutation now fails `(11, 4)` versus `(11, 5)`.

`map_live_preview_cursor` becomes `pub(super)` only so the sibling input test can exercise the real mapper. The WebSocket emission comment also restores why cursor translation happens server-side: older clients that ignore the origin still paint the cursor correctly; only their pointer mapping degrades.

The cancellation is not universal for stacked, non-bottom-aligned, or horizontally moved pane 0 layouts. Those require slice-aware pointer projection and are tracked separately in #3550 rather than expanding this regression-guard PR.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

No user-visible behavior changes in this PR; it adds a regression guard and restores design rationale.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The regression is a pure coordinate round trip and is folded into the existing Rust test function, matching the repository's one-test-per-behavior guidance.

Validation:

- mutation: naive mouse `+top` fails `(11, 4)` versus `(11, 5)`
- focused composited mouse test passes after restore
- focused live cursor tests: 4 passed
- focused `frame_json` contract test passes
- `cargo test --features serve`: 6730 passed, 8 ignored on rerun; the first run's unrelated ACP test flake passed in isolation before the clean full rerun

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** openai-codex/gpt-5.6-sol via Oh My Pi

**Any Additional AI Details you'd like to share:** Two independent validators reconstructed each maintainer review item. A separate reviewer cross-validated the completed patch and the split of the broader pointer residual into #3550.

## Summary

- Added a regression test for cursor mapping in bottom-follow, side-by-side previews.
- Exposed `map_live_preview_cursor` to sibling modules for testing.
- Restored documentation for server-side cursor translation and older WebSocket clients.
- Prevented incorrect pane offset handling during mouse coordinate mapping.

## Benefits

- Keeps painted cursor positions aligned with application cells.
- Protects against regressions in composited TUI layouts.
- Preserves compatibility with older WebSocket clients.

Broader pointer projection cases remain tracked separately in `#3550`.

### DIFF
--- a/src/server/live_ws.rs
+++ b/src/server/live_ws.rs
@@ -1035,7 +1035,9 @@ async fn handle_live_ws(
 fn frame_json(content: &str, cursor: Option<&crate::tmux::PaneCursor>) -> String {
     // The cursor is pane relative while composited content uses the window
     // grid. Emit window-relative coordinates and carry the same origin for
-    // the client's inverse pointer mapping. No pane rectangle means identity.
+    // the client's inverse pointer mapping. Translating before emission also
+    // keeps cursor painting correct in older clients that ignore the origin;
+    // only their pointer mapping degrades. No pane rectangle means identity.
     let pane0 = cursor.and_then(|c| c.composite_pane0);
     let (origin_x, origin_y) = pane0.map_or((0, 0), |p| (p.left, p.top));
     let cursor_value = match cursor {
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -7110,6 +7110,8 @@ mod tests {
     /// goes to pane 0 alone, so a pointer over a neighbouring pane must not be
     /// mapped against the full rect: that reported a column past pane 0's right
     /// edge to the agent as though its own pane were window-wide.
+    /// It also round-trips a painted composite cursor cell through mouse mapping,
+    /// pinning the bottom-follow clipping semantics.
     #[test]
     fn composited_preview_maps_the_mouse_into_pane_zero_only() {
         use ratatui::layout::Rect;
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -7137,6 +7137,34 @@ mod tests {
         assert!(hover_forward_bytes(&split, pane, 39, 5).is_some());
         assert_eq!(hover_forward_bytes(&split, pane, 40, 5), None);
 
+        // In the bottom-follow layout, the composite's top border row is
+        // clipped before painting: `first_line == pane0.top == 1`. The cursor
+        // mapper adds `top` after its anchor delta, while the mouse rect stays
+        // at the visible output origin. A click on the painted cursor must
+        // round-trip to that cursor's 1-based app cell.
+        let mut bottom_follow = cursor_for(true, true, true);
+        bottom_follow.x = 10;
+        bottom_follow.y = 4;
+        bottom_follow.pane_height = 25;
+        bottom_follow.composite_pane0 = Some(crate::tmux::PaneGeom {
+            left: 0,
+            top: 1,
+            width: 40,
+            height: 24,
+        });
+        let painted = crate::tui::home::render::map_live_preview_cursor(
+            pane,
+            usize::from(pane.height),
+            25,
+            bottom_follow,
+        )
+        .expect("visible pane cursor");
+        assert_eq!(
+            map_pane_cell(mouse_pane_rect(&bottom_follow, pane), painted.x, painted.y,),
+            (bottom_follow.x + 1, bottom_follow.y + 1),
+            "clicking the painted cursor must report the same app cell"
+        );
+
         // A no-mouse full-screen agent gets no page key from a wheel aimed at
         // the neighbour either, but keeps it over pane 0.
         let mut no_mouse = cursor_for(true, false, false);
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -123,7 +123,7 @@ const READING_CAPTURE_LINES: u16 = 2000;
 /// cursor.y`; a short capture instead anchors at the top. This keeps the
 /// cursor on the same text row for both the status-row offset (#3515) and the
 /// shorter-pane case (#2742). A hidden or out-of-bounds cursor yields `None`.
-fn map_live_preview_cursor(
+pub(super) fn map_live_preview_cursor(
     output: Rect,
     visible_rows: usize,
     line_count: usize,


njbrake (commented):

_Reviewed at `79e8414d`. Push new commits and it will be re-reviewed._

---

_Reviewed at `79e8414d`. Push new commits and it will be re-reviewed._

---

## Summary

Third PR in the #3515 coordinate chain, and the one that pins the decision the second one made. #3547 fixed the split-preview cursor by adding pane 0's origin on the render side while deliberately leaving `mouse_pane_rect` at the visible output origin, because in the common `pane-border-status top` bottom-follow layout the composite clip has already cancelled that origin. Nothing in the tree protected that decision. This PR adds a render-to-input round-trip assertion to the existing composited mouse test, widens `map_live_preview_cursor` to `pub(super)` so the sibling test can call the real mapper, and restores the older-client rationale to the `frame_json` comment. It changes no runtime behavior.

## Design alignment

**Aligned.**

Compared against the precedent this PR extends directly: `composited_preview_maps_the_mouse_into_pane_zero_only` in `src/tui/home/input.rs` (the existing home for pane-0 pointer mapping assertions, grown across #435 / #488 / #3515) and the `live_cursor_*` family in `src/tui/home/render.rs`. The PR extends both mechanisms rather than standing up a parallel one: no new helper, no new test module, no new abstraction. `pub(super) fn` is the established idiom in this module (8 occurrences in `render.rs`, 46 in `input.rs`), and there is no `#[cfg(test)]`-conditional-visibility precedent anywhere in the tree that it should have used instead.

Placement is also the minimal-surface choice, which is worth stating explicitly since the visibility widening is the one thing here that adds API surface. `map_live_preview_cursor` is private in `render.rs`; `mouse_pane_rect` and `map_pane_cell` are both private in `input.rs`. Putting the round trip in `render.rs` instead would require widening **two** functions rather than one. `pub(super)` scopes to `crate::tui::home`, which is exactly the module that owns both halves. And #3550's stated design has the future table cases living in this same `input.rs` test, so this is where the coverage accretes. I would keep it as written.

## Code review

| File | Change | Assessment |
|---|---|---|
| `src/tui/home/render.rs` | `fn map_live_preview_cursor` to `pub(super) fn` | Minimal widening, established idiom, one real reader. Body unchanged. |
| `src/tui/home/input.rs` | +28 lines of assertion inside the existing composited mouse test | The load-bearing change. Verified below. |
| `src/server/live_ws.rs` | Comment only, re-expanding the older-client rationale | Claim verified true against the current wire format. |

## The mutation test (the core of this review)

I reproduced the PR body's claim rather than taking it. Applying #3543's actual mutation to `mouse_pane_rect` (I pulled #3543's diff and copied its exact edit: `x: pane.x.saturating_add(rect.left)`, `y: pane.y.saturating_add(rect.top)`):

```
thread '...::composited_preview_maps_the_mouse_into_pane_zero_only' panicked at src/tui/home/input.rs:7162:9:
assertion `left == right` failed: clicking the painted cursor must report the same app cell
  left: (11, 4)
 right: (11, 5)
```

`(11, 4)` versus `(11, 5)`, exactly as the body claims. Reverted, GREEN.

I then ran the **entire** lib suite under that mutation, because "the guard fires" and "the guard is the only thing that fires" are different facts:

```
test result: FAILED. 4736 passed; 1 failed; 2 ignored
    tui::home::input::tests::composited_preview_maps_the_mouse_into_pane_zero_only
```

One failure out of 4,737. Before this PR, the #3543 mistake would have passed the whole suite. That is precisely the gap the PR claims to close, and it closes it.

The guard is bidirectional as a bonus. Reverting #3547's render-side fix (dropping `+ top` from the row formula) also turns it RED with the same `(11, 4)` vs `(11, 5)`, so it pins both halves of the decision, not just the mouse half.

**Is the geometry the real one?** Yes, and this is the part I most expected to find arranged. I traced it end to end:

- `finalize_live_send_resize` (`src/tui/home/mod.rs:6527`) calls `session.resize_window(pane.width, pane.height)`, which (`src/tmux/session.rs:1443`) sizes the **window** to `rows + chrome` so pane 0 lands at exactly `output.height`. For side-by-side with `pane-border-status top`, chrome is 1.
- So `window_height = 25`, `pane0 = {top: 1, height: 24}` when `output.height = 24`.
- `capture_window_composited_with_cursor` (`src/tmux/session.rs:1024-1029`) sets `c.pane_height = layout.window_height` and `c.composite_pane0 = layout.first_pane()`. So `pane_height = 25` in the test is the window height, not pane 0's, which is what the render formula needs.
- The composite is built from each pane's visible rows, so `total_lines == window_height == 25`. `set_preview_text_view` calls `compute_scroll(25, 24, 0)` = `25 - 24` = **1** = `pane0.top`. The premise holds in the real code path, not just in the test's arranged values.

The test's numbers are more faithful than the sibling `live_cursor_maps_single_and_composited_origins` in `render.rs`, which passes `line_count = 200` for a composite. A composite never carries scrollback, so 200 is not physically realizable there. It happens to be harmless because `anchor` clamps at `visible_rows`, but the new test is the one with real numbers.

**Is the round trip real?** Yes. Both sides are the production functions. `map_live_preview_cursor` is what `live_preview_cursor_pos` calls at `render.rs:3137` before `frame.set_cursor_position(pos)`; `mouse_pane_rect` + `map_pane_cell` are what `forward_mouse_to_preview` calls at `input.rs:4647`. No reimplementation, and the only production step between them is `frame.set_cursor_position` and crossterm handing the screen cell back, which is an identity on coordinates. The one argument the test collapses is that production passes `self.preview_pane_area` on the render side and `self.preview_text_view.pane` on the input side, while the test passes one `pane` to both. I checked all three view-mode branches (`render.rs` 2866/2885, 2949/2979, 3045/3059) and they are set from the same `pane_area` in the same block, so the collapse is faithful today. See Merge risk for the caveat.

- **One test per behavior:** folding into the existing test is the right call. AGENTS.md reserves splitting for cases that need *different setup*, and the setup here is identical (same `cursor_for` helper, same `pane` rect, same composited subject). #3550 additionally specifies that the remaining layouts arrive as table cases in this same function, so a separate `#[test]` now would be a symbol to merge away later. The rule genuinely cuts both ways here and I looked at the other side, but this lands on the fold side.
- **No dead code:** the `pub(super)` has a real reader; no `#[allow]`.
- **No emdashes or `--`:** clean. I grepped the added lines.
- **Conventional commit:** `fix(tui):` is valid per `pr-title-check.yml`, but see Issues.
- Coverage matrix / `#[parallel]` e2e: N/A, no web or e2e surface touched.

## Definition of done, run locally

All with `env -u CLAUDE_CONFIG_DIR`, rebased state (merge-base is `ecffb860`, current `origin/main` tip, so no rebase was needed):

- `cargo fmt --check`: clean
- `cargo clippy --features serve -- -D warnings`: clean
- `cargo test --features serve`: **6721 passed, 0 failed, 8 ignored**, exit 0

On the body's claim of "6730 passed, 8 ignored": the ignored count matches exactly and the passed count is 9 lower on macOS. That delta is platform-gated tests, not a discrepancy worth chasing. The claimed ACP flake did not reproduce here; that run was clean first try.

## Automated review check-in

CodeRabbit: "No actionable comments were generated." No line-level bot findings on the pulls endpoint. Nothing to reconcile.

## Issues found

Nothing blocking. Three things worth a line each.

1. **`fix:` prefix on a PR that fixes nothing at runtime.** `cliff.toml:85` maps `^fix` into the "Bug Fixes" changelog group; `^test` is skipped (`cliff.toml:100`). The PR's own body says "No user-visible behavior changes in this PR", and I verified that against the diff: `render.rs` is a visibility keyword, `live_ws.rs` is a comment, `input.rs` is inside `mod tests`. So this will publish a "Bug Fixes" release-note entry for a change no user can observe. `test(tui): guard crop-cancelled mouse projection` is in the allowed type list and is the accurate prefix. Easiest fix is retitling in the squash-merge box.

2. **The test's doc comment and name no longer describe everything it covers.** `/// On a composited preview the rect is the whole window while input still goes to pane 0 alone...` describes only the pane-zero-only behavior; the round trip is documented in an inline comment inside the body. Since the fold-into-one-test decision is the right one, the docstring is where the second behavior should be named. One line.

3. **Minor style.** The trailing comma inside the single-line call at `input.rs:7163` (`painted.y,`) and the fully-qualified `crate::tui::home::render::` path rather than `super::super::render::`. `cargo fmt` accepts both; take them or leave them.

## The deferral to #3550

I checked, and the deferral is honest and correctly scoped, not a dodge.

#3550 is OPEN, labeled `P2` / `area:tui`, and describes the residual precisely, including the projection formula and acceptance criteria that mandate extending *this* test as table rows. I independently derived the residual rather than reading it off the issue, and it is real:

The mouse side is correct exactly when `first_line == pane0.top` (and `pane0.left == 0`). The residual offset is `first_line - pane0.top` rows and `pane0.left` columns. Concretely, for a **stacked** split where pane 1 exceeds `MAX_CHROME_ROWS = 5` (`src/tmux/session.rs:272`), `chrome_rows` deliberately returns 0, so the window is resized to `output.height`, `first_line` becomes 0 while `pane0.top` stays 1, and the forwarded row is **one too large**. That matches #3550's description. The render side stays correct in that case, because `anchor - pane_height` is algebraically `-first_line`, so the row formula reduces to "window row minus the clip" in general; it is specifically the pointer side that lacks the clip.

Notably `self.preview_text_view.first_line` is already in hand at the `mouse_pane_rect` call site, so the real fix is small. Deferring it is a scoping choice, not a capability gap. I still think deferring is right: this PR is a regression guard, the residual predates #3547, and #3550's acceptance criteria include manual verification on the native TUI, which is not something to bolt onto a test-only PR.

## Documentation

Proportionate, and the restored comment earns its place. I verified the claim before endorsing it, since a restored "why" that is stale is worse than the short version:

`frame_json` emits `cursor.x/y` already translated by `(pane0.left, pane0.top)` and ships the origin separately under `pane0`. On the web side, `MobileLiveTerminal.tsx:862-875` paints the cursor from `cursorLineIndex(lines.length, screenRows, cursor.y)` and `cursor.x % cols` with no reference to `pane0`, while `pointerPaneCell` (`web/src/lib/liveMouse.ts:106-119`) is the only place `left`/`top` are subtracted. So a client that ignores the origin does paint the cursor correctly and only degrades in pointer mapping. The comment is accurate.

It also does not duplicate: #3547 shortened both the module doc and the `frame_json` comment; this restores the rationale only to the `frame_json` comment, next to the translation it explains, and the module doc (lines 24-29) still carries the short form. One canonical home.

## Merge risk

**Low.** Zero runtime behavior change, verified against the diff line by line. Reverts cleanly. Full `--features serve` suite green locally and in CI.

Two residuals, neither introduced by this PR:

- The test collapses `preview_pane_area` and `preview_text_view.pane` into one rect. True in all three view-mode branches today, but not itself pinned by any test. If those ever diverged, this guard would silently stop modelling production while still passing. Cheap to note, not worth a test.
- The round trip asserts *consistency* between the two sides, not the absolute correctness of either. A same-sign shift applied to both would pass it. That is covered from the other direction by `live_cursor_maps_single_and_composited_origins`, which pins the render side absolutely.

The most likely way this bites us is not this PR at all: it is the #3550 residual, where a user on a stacked split clicks and the agent receives a row one off. Nothing here surfaces that, and we would hear it from a user. Which is the case for keeping #3550 warm, not for holding this.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Merge risk:** Low; no runtime change, and the residual pointer offset for stacked layouts stays open under #3550

The guard holds. I reproduced the mutation and got the exact values claimed, and confirmed it is the only test in 4,737 that catches the #3543 mistake, which is the strongest form of the claim this PR could make. The geometry is the real production geometry, traced from `resize_window` through `chrome_rows` and `compute_scroll` rather than assumed. Both round-trip endpoints are the production functions. The `pub(super)` widening is the minimal option and matches the module's idiom.

The only thing worth changing before merge is the commit type: this belongs under `test(tui):`, not `fix(tui):`, so it does not land in the "Bug Fixes" release notes for a change with no user-visible effect. Retitle at squash and merge.

**LGTM.** Thanks for coming back to pin this one, @jerome-benoit. The decision you made in #3547 was the non-obvious half of that fix, and it was the half nothing in the tree defended; a round trip through both production mappers is exactly the right shape of guard for it.

njbrake (commented):

_Reviewed at `cf5e18e5`. Push new commits and it will be re-reviewed._

---

_Re-reviewed at `cf5e18e5`, following the pass at `79e8414d`. Push new commits and it will be re-reviewed._

---

## Both suggestions taken. Nothing else moved.

This is a short pass on purpose: the delta since the last review is **two lines of doc comment and zero lines of code**, so re-running the full analysis would be theatre rather than diligence. What was checked:

- **Suggestion 1, the commit type.** Title is now `test(tui): guard crop-cancelled mouse projection`. `cliff.toml:100` skips `^test`, so this no longer publishes a "Bug Fixes" release-note entry for a change with no user-visible effect. That was the only thing standing between the last review and a merge recommendation.
- **Suggestion 2, the test docstring.** `cf5e18e5` adds the two lines naming the second behavior the test now covers:

  ```
  /// It also round-trips a painted composite cursor cell through mouse mapping,
  /// pinning the bottom-follow clipping semantics.
  ```

  Accurate to what the assertion does, and it names the non-obvious part (the bottom-follow clipping) rather than restating the mechanics. No emdashes or `--` separators.
- **Suggestion 3** was explicitly optional style and is untouched, which is fine.

**Nothing else changed.** `origin/main` has not moved (`ecffb860` is still both the tip and the merge base), and the authored diff is the same three files at +34/-2 that the last pass verified: `src/server/live_ws.rs` comment only, `src/tui/home/render.rs` one visibility keyword, `src/tui/home/input.rs` test-only. The two added lines are inside `mod tests`, so this remains a zero-runtime-behavior PR. All 27 CI checks pass at this head, including `codecov/patch` and `codecov/project`.

The mutation evidence from the last pass stands unchanged, since neither the assertion nor either mapper was touched: the guard fails with `(11, 4)` versus `(11, 5)` under #3543's exact mutation, and it is the only test in 4,737 that catches it.

## Verdict

**Design alignment:** Aligned
**Implementation:** Approve
**Project fit:** Accept
**Merge risk:** Low; no runtime change, and the residual pointer offset for stacked layouts stays open under #3550

**LGTM, and now with the right commit type.** Thanks for turning both of those around, @jerome-benoit.

njbrake (approved):
Approving on the strength of the mutation evidence: the new assertion goes red under #3543's exact naive `+top` mutation with the claimed `(11, 4)` versus `(11, 5)`, and it is the only test in 4,737 that catches it. The geometry was traced through `resize_window`, `chrome_rows`, and `compute_scroll` rather than assumed, and both round-trip endpoints are the production mappers. Thanks for pinning the #3547 decision, and for the `test(tui):` retitle so this stays out of the Bug Fixes release notes.

PR #3556, opened 2026-08-27
fix(session): keep opencode conversations pinned when project_path is shared

### LINKED ISSUE #3555: bug(session): opencode launch-time verify-on-resume adopts another same-cwd session's conversation (survives exclusions; pins are one-shot)
### Environment

- agent-of-empires `1.15.0` (reproduces identically on current `main` @ `ecffb86`)
- Fedora 42, tmux 3.6a, tool = **opencode** (shared global SQLite store)
- ~33 AoE sessions, **20 sharing one `project_path`** (repo root) — the worst-case supported layout (AoE's own worktree/group model)

### Summary

Two independent holes let an opencode session walk away from its own stored conversation and adopt the directory's newest one whenever more than one AoE session lives in the same working directory:

1. **Missing inactive-peer exclusion for opencode.**
   `Instance::retroactive_capture_exclusion_set()` passes `include_inactive_same_tool = true` only for claude/host-codex/kimi. opencode gets `false`, so stopped/archived/pane-less same-cwd peers are *not* excluded from the launch-time MRU scan over the shared SQLite store. During boot recovery every peer is pane-less by definition → nothing protects anyone (#2344/#2708 shape).

2. **verify-on-resume supersedes regardless of ownership (the surviving path).**
   Even with all peers excluded, `acquire_session_id()`'s freshest-scan replaces the stored sid with "freshest live observation" in the shared directory. With >1 session per cwd that conversation is frequently *another session's* — "freshest in dir" ≠ "mine". `set-session-id` pins cannot defend against this path long-term because they are consumed at first launch (`promote_one_shot`, comment on `persist_session_id_with_storage`) — so each repaired fleet re-breaks at the next restart/reboot. We hit exactly this recurrence loop across reboots on Jun→Aug versions despite #2345/#2481/#2709/#2921 fixes.

### Production evidence (real fleet)

Debug log, three later sessions superseded to the *same* foreign sid within milliseconds of each other at post-reboot startup recovery:

```
2026-08-27T07:45:37.863310Z INFO session.store: Replacing stored session id … fresh=ses_17466c3b…
2026-08-27T07:45:37.863315Z INFO session.store: Replacing stored session id … fresh=ses_17466c3b…
2026-08-27T07:45:37.863306Z INFO session.store: Replacing stored session id … fresh=ses_17466c3b…
```

`sessions.json` afterwards: conversation UUIDs claimed by multiple session rows; panes attached to other sessions' chats while the flock guard kept the store itself mostly correct — user-visible state worse than either half alone (pane ↔ store divergence).

A deterministic live A/B on production hardware (same harness flow as startup recovery; PID-verified pane cmdlines, not just store reads):

| binary | pane actually ran | verdict |
|---|---|---|
| upstream `1.15.0` | `opencode --session <peer's sid>` | ❌ stolen |
| patched | `opencode --session <own sid>` | ✅ |

### Proposed fix (PR linked below)

1. Add opencode to the inactive-peer exclusion arm (claude/codex/kimi already have it).
2. When another instance shares the instance's `project_path`, skip opencode `try_retroactive_capture()` entirely: resume must trust the stored sid. Fresh-launch sid assignment (preassign/fresh arms) and post-prompt poller binding are untouched; the drain filters keep doing the bookkeeping.

Both mitigations together were validated live on a 20-sessions-per-directory fleet: 4/4 adversarial relaunches resumed their own conversations (including pin-consumed `Default` intents with older stored sids than peers/orphans present), zero improper `Replacing stored session id` lines afterward.

### Regression tests included

- `exclusion_set_includes_inactive_same_cwd_peer_for_opencode`
- `retroactive_capture_is_skipped_when_project_path_is_shared`

Happy to provide fuller debug.log windows or test any maintainer-proposed variant.

Chiyo-no-sake:
## Description

`opencode` sessions walk away from their own stored conversation whenever more than one AoE session shares a working directory, the same family as #2344 → #2708 → #2858, through two remaining gaps plus two ownership surfaces the earlier fixes did not cover:

1. **Inactive-peer exclusion misses opencode.** `retroactive_capture_exclusion_set()` passed `include_inactive_same_tool = true` only for claude/host-codex/kimi. opencode uses one shared SQLite store keyed by directory, structurally identical to Claude's shared encoded-cwd dir, so stopped/archived/pane-less same-cwd peers were never excluded from the launch-time MRU scan.
2. **verify-on-resume supersedes regardless of ownership.** Even with a complete exclusion set, `acquire_session_id()` replaces the stored sid with the "freshest live observation" in the directory. With >1 session per cwd, "freshest in dir" ≠ "this session's conversation" *by construction*, and `set-session-id` pins cannot defend long-term because they are consumed at first launch (`promote_one_shot`).

### Changes

- `retroactive_capture_exclusion_set()`: add host opencode to the inactive-peer arm (gated on `!is_sandboxed()`, consistent with codex/kimi: sandboxed stores are container-private, #3317).
- New `capture::opencode_shared_store_peers()`: walks every AoE profile, resolves each profile's effective OpenCode store identity with opencode's own rules (`OPENCODE_DB` override, `XDG_DATA_HOME`/`HOME` data dir, mtime-picked `opencode*.db`), and reports whether another host-opencode instance shares this session's canonicalized `project_path`, collecting the conversation ids those peers own (current `agent_session_id` *and* opencode conversations parked in `prior_tool_session_ids`, regardless of bucket or liveness).
- `try_retroactive_capture()` (host opencode arm): when the store+cwd is shared, return `None` early; resume trusts the stored sid. The own side resolves from the **process environment** because that is the store `opencode_db_path()` actually opens for the scan; peer sides resolve their profile `environment` through `resolve_host_environment_value` (the `$VAR` / bare-key / last-wins grammar launch applies) layered over the process env, and every in-scope read failure reports shared.
- Post-launch poller: after every SQLite candidate, the host-opencode poller re-reads cross-profile ownership. A current id-less peer already marks attribution ambiguous, and once that peer reports its sid the sid becomes an owner claim; either condition rejects the candidate. This closes the staggered race where the peer creates its SQLite row after the poller was constructed, while the scan itself keeps running for fresh-session post-prompt binding.

### User-visible trade-off, stated plainly

A session sharing a store+cwd with any other opencode row (current or parked, any profile, including archived or trashed) permanently gives up retroactive MRU capture while that peer exists: a missing sid now yields a fresh conversation instead of a possibly-wrong adopted one. Fresh sessions are unaffected in practice because `preassign_opencode` mints the sid up front and the poller binds the first prompt.

Fixes #3555

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (compose_exclusion_with_persisted_peers agent enumeration, restored opencode_data_dir rustdoc, inline rationale next to the guard)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The guard is now covered by tests that fail when it is removed, against a real seeded SQLite store (`opencode_same_cwd_clobbering_tests`):

- `retroactive_capture_is_skipped_for_every_same_store_peer_shape`: table-driven over five peer shapes (same profile; cross-profile with an unrelated env entry; own profile redirecting `OPENCODE_DB`; peer reaching this store through `$VAR`; cross-profile parked owner). The seeded store holds an untracked freshest row, so deleting the guard turns the test red with `Some(<foreign sid>)` instead of `None`.
- `retroactive_capture_still_runs_without_a_same_cwd_peer`: the sole-session case keeps its retroactive capture (guards against making the guard unconditional).
- `opencode_poller_revalidates_idless_peer_after_observation`: constructs the poller while a cross-profile peer has no sid, inserts the peer's SQLite row afterward, proves the raw MRU scan sees it, then proves the poller abstains after dynamic ownership revalidation.
- `exclusion_set_includes_inactive_same_cwd_peer_for_opencode`: unchanged, mutation-verified in review.

(`cargo test --lib session::` → 1809 passed on the merged current `main` layout; clippy `-D warnings` clean; production-fleet A/B validation from the original report stands, PID-verified via `/proc/<pid>/cmdline`.)

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** opencode CLI pairing sessions with GLM (Z.ai).

**Any Additional AI Notes you'd like to share:** Root-cause analysis and code were developed jointly against this machine's real debug logs; every review round was implemented and re-verified locally by me (including reproducing the reviewers' findings before fixing them). Per your note: reviewer questions are answered by me directly, not piped through the model.

**NOTE:**

---

<details>
<summary>Appendix — deterministic A/B repro (production hardware, original report)</summary>

Two AoE/opencode sessions in one cwd, each owning its own conversation; peer pane fully dead; restart the older session exactly like startup recovery does:

| binary | pane actually ran | verdict |
|---|---|---|
| upstream `1.15.0` | `opencode --session <peer's sid>` | ❌ stolen |
| patched | `opencode --session <own sid>` | ✅ |

</details>

## Summary

- Prevents OpenCode sessions from adopting conversations from other sessions that share the same project directory.
- Detects shared OpenCode stores across profiles, redirected paths, and environment-based configuration.
- Skips retroactive conversation capture when another session owns the same project and store.
- Excludes active and parked peer sessions from poller ownership checks.
- Preserves fresh-session assignment, sandboxed paths, and post-prompt binding.
- Removes the former profile-specific store resolver and centralizes shared-store detection.
- Adds regression tests for shared paths, inactive peers, redirected stores, profile environments, parked owners, and poller behavior.

## Benefits

- Improves resume reliability.
- Prevents conversation mix-ups between concurrent OpenCode sessions.
- Keeps fresh-session and sandboxed-session behavior unchanged.
- Fails closed when ownership information is unavailable.


njbrake (commented):

_Reviewed at `d8cef91b`. Push new commits and it will be re-reviewed._

---

## Summary

Two changes to opencode's launch-time session-id attribution. First, `Instance::retroactive_capture_exclusion_set()` now passes `include_inactive_same_tool = true` for opencode, so stopped, archived, and pane-less same-cwd opencode peers are excluded from the MRU scan. Second, `try_retroactive_capture()`'s opencode arm returns `None` outright when `opencode_shares_store_with_peer()` finds another non-sandboxed opencode instance on the same canonicalized `project_path` resolving to the same OpenCode store, so resume trusts the stored sid instead of re-deriving one from the directory's freshest row.

**The bug is real and I reproduced it in-tree.** I seeded a genuine opencode SQLite store with two rows in one shared directory (a peer's recorded sid, plus a fresher untracked one) and ran `try_retroactive_capture()` with and without the guard hunk. Without the guard it returns `Some("untracked-sid-FRESH")`, a conversation this session has no claim to; with the guard it returns `None`. The exclusion-set hunk alone does not close this, because the exclusion set can only contain sids AoE has recorded, and the freshest row in a shared directory is frequently one it has not (the peer's older thread, or a conversation started with a bare `opencode` in that cwd). Both hunks are load-bearing.

## Design alignment: aligned

The precedent is Kimi, and it is a close one. `capture::kimi_store_is_shared` already does exactly this: canonicalize the project path, resolve the agent's store identity, walk every profile from `list_profiles()`, skip profiles whose store differs, load each same-store profile's `sessions.json`, and return true on the first same-cwd peer. `try_retroactive_capture`'s kimi arm refuses the MRU scan on that basis, and `capture_freshest_session_id` carries a comment saying so in as many words ("a shared store refuses the MRU scan entirely inside try_retroactive_capture and surfaces here as None"). This PR extends that mechanism to opencode rather than inventing a parallel one, and it correctly extracts `opencode_db_path`'s resolution rules into `opencode_store_identity(lookup)` so the same rules serve both the real capture and the cross-profile comparison. That extraction is the right shape; `pub(crate)` on it is genuinely needed, since `instance.rs` is a different module.

Three deviations from that precedent, all worth noting:

- Kimi passes **both** `resolved_host_environment()` and `profile_host_environment()` for its own side, specifically so a `before_session`-minted `KIMI_CODE_HOME` still counts static-profile siblings as sharing. The opencode version reads only the static profile list. `resolved_host_environment`'s own doc says "anything that has to agree with the launched agent about a variable's value must read it here rather than from `profile_host_environment` alone."
- Kimi fails **closed** on an unresolvable peer store (`let Some(peer_home) = ... else { return true }`); this PR fails open (`Err(_) => continue`). See non-blocking item 4.
- Kimi also treats a peer that *parked* a kimi conversation (`prior_tool_session_ids`) as an owner; the opencode version only matches `i.tool == "opencode"`. In practice `compose_exclusion_with_persisted_peers` already puts parked sids in the exclusion set, so this is narrow, but it is an asymmetry.

## Compliance

`cargo fmt --check` clean. `cargo clippy --all-targets` produces exactly one warning, `clippy::items_after_test_module` in `src/process/macos.rs`, which is pre-existing on main and not attributable to this PR. `cargo test --lib` green at 4742 passed. All checks on `d8cef91b` are now green (28 success, 1 skipped) after the `CI` and `Tests` runs were approved; before that, neither had ever run on this branch at any head, so the body's "1776 passed, clippy clean" had not been independently confirmed until now. It holds up.

No `web/` changes, so the body's N/A coverage-matrix checkbox is correct, and the Rust-only diff is N/A for `codecov/patch` as documented. No dead code, no `#[allow(dead_code)]`, no new dependencies, no CI or build-hook changes, no breaking changes to stored data or config schema.

One house-rule violation: an emdash in the new doc comment on `opencode_shares_store_with_peer` ("skip the MRU capture ... a skipped rescan costs one resume"). AGENTS.md forbids emdashes in human-authored prose. The new doc blocks are otherwise long but earn their space, explaining why rather than restating code; the one exception is the inline comment above the guard call, which restates the function name it is about to call.

## On the four CodeRabbit rounds

1. **Restrict the guard to same-store opencode peers, handle sandboxing.** Agree, addressed in `cea5f57`. The original `opencode_shares_project_path` matched any same-profile peer regardless of tool, which really would have suppressed valid capture for a Claude or Codex neighbour.
2. **Add `#[serial]`, reuse `EnvGuard`.** Agree, addressed. These tests mutate process-global `HOME`/`XDG_CONFIG_HOME`, and AGENTS.md reserves `#[serial]` for exactly that.
3. **Walk cross-profile peers, fail closed on storage errors.** Agree, addressed in `d0bfab5`. Profiles with no overrides do share the default store, so a single-profile check was genuinely insufficient.
4. **Compare store identity rather than environment-list emptiness.** Agree on the diagnosis, and this is where I part company with the remedy. The env-emptiness proxy was wrong. But the fix it steered toward, resolving the *own* side's store from the own profile, opened a new hole in the exact direction this PR exists to close (blocking item 1). A cheaper and strictly safer remedy was available: keep the plain canonicalized-cwd comparison across all profiles, which over-skips and never under-skips.
5. **`opencode_profile_env_lookup` returns literal `$VAR` instead of resolving it.** Agree, and it is unaddressed on the pushed head. I confirmed this is a reachable steal, not a theoretical one (blocking item 2).

Where CodeRabbit is wrong-adjacent: its round-4 remark about bare `KEY` entries implies the doc is inaccurate. The doc claim that bare `KEY` entries inherit the process value is in fact true, not by handling the case but by falling through to `std::env::var(key)`. The doc is accurate; the mechanism is just implicit.

## Blocking

**1. The guard's own-store resolution reopens the steal for any profile that redirects the OpenCode store.** `opencode_shares_store_with_peer` resolves *this* session's store identity from its own profile's `environment`, but `try_capture_opencode_session_id` takes no environment at all: it calls `opencode_db_path()`, which reads the **process** env unconditionally. The guard therefore compares against a store the scan will never open.

Reproduced: with a profile carrying `environment = ["OPENCODE_DB=..."]`, the guard resolves its own store to that override, classifies every default-store peer as a different store, returns `false`, and `try_retroactive_capture()` then returns `Some("peer-conv-in-process-store")`, the peer's conversation.

`opencode_launch_mirrorable_by_ambient_serve` already documents this exact configuration as a real one, and it is the configuration for which `preassign_opencode` is skipped, so it is precisely the population that relies on retroactive capture most. This is a regression introduced by the round-3 change; the `opencode_shares_project_path` design described in the PR body would not have it.

**2. `opencode_profile_env_lookup` does not implement the project's environment grammar.** It extracts values with a bare `strip_prefix("KEY=")`, so `XDG_DATA_HOME=$SOME_VAR` yields the literal string `"$SOME_VAR"`, and the peer resolves to a store path that exists nowhere. Reproduced: a peer profile pointing at this session's store through `$VAR` is classified as a different store, the guard returns `false`, and the scan adopts the peer's directory-freshest conversation.

It also takes the **first** matching entry, while `resolve_host_environment_value` and `resolve_host_environment_pairs` both take the **last** ("later entries replace earlier entries, matching historical assignment behavior"), so a duplicated key makes the guard disagree with what launch actually applied. `super::environment::resolve_host_environment_value` already handles `$VAR`, `$$`, bare-key, and last-wins; the `kimi_var_form_homes_still_detect_shared_store` test proves this configuration shape is expected in this codebase.

**3. Two of the three new tests cannot fail.** I mutation-tested each production hunk:

- Deleting `|| self.tool == "opencode"` from the exclusion arm turns `exclusion_set_includes_inactive_same_cwd_peer_for_opencode` red: `panicked at src/session/instance.rs:17565: stopped same-cwd opencode peer sid must be excluded from retroactive capture`. That test is real.
- Deleting the entire guard call from `try_retroactive_capture` leaves **all three tests green**. `retroactive_capture_is_skipped_when_project_path_is_shared` and `cross_profile_peer_with_unrelated_env_is_still_same_store` both assert `try_retroactive_capture() == None`, and in a temp `$HOME` with no opencode SQLite store the capture path returns `None` regardless. They assert the default answer.
- Additionally, `cross_profile_peer_with_unrelated_env_is_still_same_store` writes its `config.toml` to `$XDG_CONFIG_HOME/agent-of-empires/profiles/...`, but debug builds use the `agent-of-empires-dev` namespace (AGENTS.md, Build and Test Commands). I confirmed by printing `load_profile_config`: the file is never read, `overrides` comes back empty, and the "unrelated env entry" the test is named for is never applied. That test would pass with its config write deleted entirely.

The load-bearing half of this PR ships with no test that can detect its removal.

## Non-blocking

**4. The fail-closed doc overstates the code.** The doc says every unreadable path returns `true`, but the peer-store arm returns `continue` on `Err`. I traced the reachable failures: `opencode_store_identity` errors only on `OPENCODE_DB=:memory:` or an unresolvable `HOME`. In the first case the peer provably writes to a different (in-memory) store, so `continue` is correct; in the second, this session's own resolution would have failed first and returned `true`. The asymmetry is defensible, but it contradicts both the doc paragraph directly above it and the kimi precedent, which returns `true` in the same spot. Either fix the doc or match the precedent.

**5. Cost.** The body's "extra storage read occurs once per launch on the resume path only" is roughly right but understated. Measured in a debug build: about 2ms when it finds a peer early (it returns on first match), about 0.8ms per profile when it does not, so roughly 40ms at 50 profiles with 30 sessions each. Not a hot path. `try_retroactive_capture` is reached from `acquire_session_id` (up to twice per launch, once through `capture_freshest_session_id` and once through the `tmux_exists` fallback), from `self_heal_session_id` on read-only CLI commands, and from `capture_omp_before_restart` which opencode never reaches.

One avoidable multiplier: `opencode_profile_env_lookup` calls `resolve_config_or_warn` once **per key**, so up to three full config resolves per profile, each re-reading the global config and reinstalling that profile's status rules via `install_from_config`. Kimi resolves once per profile. Hoisting the resolve out of the closure is a 3x saving for free. Log spam is a non-issue: a non-existent profile returns defaults without warning; only a malformed `config.toml` warns, and then three times per profile.

**6. Sandboxed opencode is added to the exclusion arm unconditionally.** Codex and kimi are gated on `!self.is_sandboxed()`; opencode is not. `DockerContainer::generate_name(&instance.id)` makes containers per-instance, so a sandboxed opencode store is instance-private and cannot contain a peer's sid, which makes the widening harmless but also pointless and inconsistent with the two tools it sits next to. Claude is also unconditional, so this is arguably a coin flip; worth a sentence of rationale either way.

**7. What a shared-cwd user gives up, stated plainly.** The guard does not filter by bucket, and `storage.load()` returns archived and trashed rows (the sibling walk in `compose_exclusion_with_persisted_peers` inspects `inst.is_archived()` on that same list). So an opencode session sharing a directory with *any* other opencode row, including one the user trashed but has not purged, permanently loses retroactive capture. Concretely: a session that has no stored sid can never adopt one from the store, and `self_heal_session_id` becomes a permanent no-op for it (its own `contended_capture_cwds` gate is narrower, covering only id-less peers).

In practice `preassign_opencode` mints the sid up front in the common case and the poller binds post-prompt, so retroactive capture is the third fallback, and trading "adopts the wrong conversation" for "starts a fresh one" is the right trade. I verified the single-session case is untouched: with no same-cwd peer, capture still runs and returns the session's own sid. But this trade is not mentioned in the PR body and should be, because it is the change's real user-visible cost.

**8. The production-fleet evidence is unverifiable here.** Twenty sessions, `/proc/<pid>/cmdline` PID verification, "4/4 own sids". I do not dispute any of it, and it is Linux-only and not reproducible from this repo. It should not substitute for the in-tree regression test, which is what blocking item 3 is about.

## Documentation: two misses, both caused by this diff

- **The `opencode_data_dir` doc block is orphaned.** Confirmed on the rebased tree in `src/session/capture.rs`. The block was lifted off `opencode_data_dir` when it became `opencode_data_dir_with`, and re-added below the new function, immediately above `read_opencode_sessions_from_sqlite_at`'s doc. The rendered rustdoc for `read_opencode_sessions_from_sqlite_at` now reads, verbatim: "Resolve opencode's data directory. Uses `XDG_DATA_HOME` if set (same `xdg-basedir` npm package opencode uses), otherwise `$HOME/.local/share`. Both Linux and macOS use this path. Load opencode's session rows from its SQLite store at `db_path`. ..." `opencode_data_dir_with` itself is left undocumented. Mechanical misplacement, not a judgment call.
- **`compose_exclusion_with_persisted_peers`'s doc is now stale.** It still says "Claude, host Codex, and host Kimi additionally need inactive same-tool protection", which is the enumeration this PR just changed. This is the hand-maintained mirror the diff invalidated without touching.

## Suggested fixes

I staged and mutation-tested all of these; they are described rather than attached since this review is read-only.

- Resolve the guard's **own** store from the process env (`opencode_store_identity(&|k| std::env::var(k).ok())`), matching what `try_capture_opencode_session_id` actually opens, with a comment explaining why. Peer profiles keep their per-profile resolution. Closes item 1.
- Replace `opencode_profile_env_lookup(profile, key)` with `opencode_env_lookup(entries, key)` delegating to `super::environment::resolve_host_environment_value(entries, key).or_else(|| std::env::var(key).ok())`, and hoist the `resolve_config_or_warn` call to once per profile in the loop. Closes item 2 and the 3x resolve from item 5.
- Replace the two vacuous tests with one table-driven test, `retroactive_capture_is_skipped_for_every_same_store_peer_shape`, seeding a real opencode SQLite store (`rusqlite` is already a dependency) holding an untracked freshest row, walking four cases: same profile; cross-profile with an unrelated env entry; own profile redirecting `OPENCODE_DB`; peer profile pointing at this store through `$VAR`. Plus a companion `retroactive_capture_still_runs_without_a_same_cwd_peer` proving the sole-session case is untouched. Three test functions in, three out, so no net compile-weight cost, and the config writes go through `get_profile_dir_path` so they land in the namespace the code reads.
- Restore the `opencode_data_dir` doc block onto `opencode_data_dir_with`, update the `compose_exclusion_with_persisted_peers` enumeration to include opencode with a one-line reason, and replace the emdash.

Mutation results on the fixed tree, each hunk reverted in turn: removing the guard turns the shape test red (`left: Some("...oc0000000003"), right: None`, on the same-profile case); reverting own-store to profile-resolved turns it red on the OPENCODE_DB-redirect case; reverting the env lookup to `strip_prefix` turns it red on the `$VAR` case; removing the exclusion arm turns the exclusion test red. With all fixes applied: `cargo fmt --check` clean, clippy unchanged from main, `cargo test --lib` green at 4742 passed.

Two things I deliberately did not change, because they are the author's or maintainer's call: whether sandboxed opencode belongs in the exclusion arm (item 6), and whether the guard should also match peers that parked an opencode conversation via `prior_tool_session_ids` the way kimi does.

## Merge risk: moderate as pushed, low with the fixes

Blast radius is a silent wrong answer in a data path, which is the bad kind. When the guard mis-classifies, the user does not get an error; they get someone else's conversation attached to their pane, and `debug.log` records it as a routine "Replacing stored session id with fresher live observation". That is how this bug family (#2344, #2708, #2858) has kept coming back.

Reversibility is good: pure logic, no migration, no stored-data change, `git revert` fully undoes it.

Coverage is the actual problem as pushed. The guard, which is the half of this PR that fixes the reported symptom, has no test that fails when it is deleted, and the two configurations where it currently fails open have no test either. Nothing in the PR would surface either case: no assertion, no distinguishable log line, and the `tracing::debug!` fires only on the path that works. We would hear about it from a user filing #3555 again, and the report would look identical to the current one, which is why this has recurred across four releases.

The most likely way it bites: a user with an `OPENCODE_DB` or `XDG_DATA_HOME` entry in a profile, which is exactly the population `opencode_launch_mirrorable_by_ambient_serve` excludes from preassignment and therefore the population most dependent on this code path, upgrades, sees the fix announced, and keeps losing conversations.

## Verdict: request changes, and I would take it once they land

This is good work on a real, recurring bug, built through the existing kimi mechanism rather than beside it, which is the thing most likely to go wrong in a PR of this shape and did not. The author responded to four rounds of bot review carefully and the code reads like someone who understood the problem.

What holds it back is narrow and fixable. The round-3 move to per-profile store identity was the right diagnosis with a remedy that overshot: resolving the own side per profile compares against a store the capture never opens, which reopens the exact steal for profiles that redirect `OPENCODE_DB`, and the hand-rolled env parser misses the `$VAR` grammar that the sibling kimi code has a locked-in regression test for. I reproduced both as live steals. And the guard half of the fix currently ships with two tests that pass identically when it is deleted, one of which also writes its fixture to the wrong app-dir namespace. Fix the own-store resolution, route the env lookup through `resolve_host_environment_value`, and land one test that goes red without the guard, and this is an approve.

## One flag on the PR body

**The body describes a different implementation than what is pushed.** The Changes section says the fix adds `opencode_shares_project_path()`, "(new helper, canonicalized comparison like the #2858 walk)". No such function exists on the branch; it was replaced across three follow-up commits by `opencode_shares_store_with_peer()`, which resolves per-profile store identity and iterates every profile's `Storage`.

That matters beyond bookkeeping: the simpler design the body still describes would have been both cheaper and *strictly safer*, because a plain canonicalized-cwd comparison over-skips and can never under-skip, whereas the shipped store-identity comparison is what introduces both fail-open paths. The body also still claims "the two unit tests below lock both behaviors deterministically", which the mutation testing shows is not true of the second behavior. Worth updating before this lands, with a sentence on why store identity was judged worth the added surface.

src/session/instance.rs
```diff
@@ -3217,12 +3217,92 @@ impl Instance {
...
+            }
+            let Ok(storage) = super::storage::Storage::new_unwatched(&profile) else {
+                return true;
+            };
+            let Ok(instances) = storage.load() else {
+                return true;
+            };
+            if instances.iter().any(|i| {
+                i.id != self.id
+                    && i.tool == "opencode"
```
jerome-benoit: [P1] Count parked OpenCode owners across profiles

This guard only recognizes peers whose current tool is `opencode`. `swap_tool` moves the outgoing OpenCode SID into `prior_tool_session_ids["opencode"]` and changes `tool`. `compose_exclusion_with_persisted_peers` only loads the current profile, so a parked owner in another profile is absent from both checks. With the same store and canonical cwd, resume can therefore MRU-adopt that parked SID, reproducing #3555. Match the existing Kimi ownership rule: count either a current OpenCode peer or a non-empty parked OpenCode SID, regardless of bucket or liveness, and add a cross-profile parked-owner regression.

src/session/instance.rs
```diff
@@ -3217,12 +3217,92 @@ impl Instance {
             &self.project_path,
             &self.tool,
             self.tool == "claude"
+                || self.tool == "opencode"
```
jerome-benoit: [P2] Keep shared-store pollers from making cross-profile MRU claims

This exclusion is snapshotted into the OpenCode poller, but persisted peers are loaded only from the current profile. The new cross-profile guard runs only in `try_retroactive_capture`; `opencode_poll_fn` still selects the cwd MRU after launch. The drain checks `sid_owners` and duplicate claims only in its loaded snapshot, and the final CAS opens only the update profile’s `sessions.json`, so two profiles sharing one store cannot see each other’s ownership. With skewed polling, the non-owner can persist a newly observed SID first. Propagate shared-store ambiguity into the poller or replace MRU selection with a pane-attributable signal, and add a regression with staggered cross-profile observations.

jerome-benoit (changes requested):
Reviewed exact head `d8cef91b27a4c59f50d35edb9929f4ad41671d42`. I independently verified the blocking own-store, environment-resolution, and vacuous-test findings already covered by the existing njbrake and CodeRabbit reviews, so I am not duplicating them.

Two additional cross-profile ownership gaps remain: parked OpenCode owners are omitted from the new guard, and the post-launch MRU poller still bypasses that guard while its collision and CAS protections remain profile-local.

CI is green at this SHA and GitHub reports the PR as mergeable, but these ownership paths can still reproduce the conversation misattribution the PR is intended to prevent.

src/session/capture.rs
```diff
@@ -1979,6 +1990,102 @@ pub(crate) fn opencode_poll_fn_sandboxed(
...
+            }
+            if let Some(parked) = inst
+                .prior_tool_session_ids
+                .get("opencode")
+                .and_then(|p| p.agent_session_id.as_deref())
+                .filter(|s| !s.is_empty())
+            {
+                owned.push(parked.to_string());
+            }
+            if owned.is_empty() {
```
jerome-benoit: **[P1] Treat id-less peers as shared and revalidate poller ownership.** `shared` is set only after a current or parked SID is found, so a same-store, same-cwd OpenCode peer whose SID is not known yet is skipped here. That state is expected because OpenCode creates its SQLite row only after the first prompt. The host poller then snapshots `owner_sids` only once at construction, so a peer SID created afterward cannot enter the exclusion set and another profile's poller can still MRU-adopt and persist that row first. Please determine shared-store ambiguity independently of whether a SID is already known, and re-evaluate ownership before accepting each poll result (or abstain while an id-less peer exists). Add a regression that constructs the poller with an id-less cross-profile peer, inserts the SQLite row afterward, and verifies that the non-owner cannot claim it.

jerome-benoit (changes requested):
Re-reviewed and cross-validated at `06b676f4885814e709f2b3e5bf24edf634eb7ac4`. The previous parked-owner, environment-resolution, and already-known cross-profile poller ownership findings are fixed, and the replacement SQLite tests now exercise those paths. One remaining P1 is inline.

The existing CodeRabbit sandbox finding also remains valid as a **P2** and should be fixed; I am not duplicating it: https://github.com/agent-of-empires/agent-of-empires/pull/3556#discussion_r3896342883.

The PR is currently conflicting. GitHub exposes only title/template checks for this head; I did not run local builds or tests.

PR #3558, opened 2026-08-27
fix(session): reap session-id pollers whose tmux session is gone

Eric162:
## Description

A session-id poller whose tmux session was killed outside AoE never stopped, holding one of the 50 budget slots for the life of the process.

`is_pane_dead` cannot tell a dead pane from a session tmux cannot find. tmux answers a `display-message` against a missing session with **exit status 0, an empty stdout, and nothing on stderr**, so the old `stdout.trim() == "1"` mapped "no such session" to "not dead":

```
missing session : exit 0, stdout ""     -> "" == "1" -> false   (poller never stops)
live pane       : exit 0, stdout "0"
dead pane       : exit 0, stdout "1"
```

Measured on a daemon up for three days: **50/50 poller threads, 32 of them polling sessions with no live tmux session at all**. Live sessions were left with no poller, so their agent session ids stopped refreshing, and the log filled with `Session-id poller budget exhausted (50/50)` (37,388 lines, 23 MB).

Two changes:

**Rebased onto current `main`.** `src/session/instance.rs` was split into `src/session/instance/` upstream (#3595 and siblings), so the poller changes now live in `instance/polling.rs` and `instance/tmux_session.rs`. The two commits are squashed into one, since the second no longer applies to the split module and they were always one logical change.

**Review round 2 (@jerome-benoit) added a third change**, and it is the same leak in a case the first pass missed: poller repair and the poller's own seed both went through the *any-kind* pane lookup (`live_any_kind_name_for_id`, which falls back `agent -> terminal -> container`). With the agent gone and a paired terminal alive, the replacement poller was seeded with the terminal's name, `live_agent_session_name`'s cache fast path returned it before `NameShape::agent` was applied, and `probe_pane` read `Alive`. The slot was held for an agent that was gone. Both now resolve the agent pane specifically; `has_live_tmux_pane_in` keeps the any-kind lookup for peer exclusion and TUI reload, where any live pane really is the question.

**1. A four-state `probe_pane`** (`Alive` / `Dead` / `Missing` / `Unknown`). `is_pane_dead` keeps folding `Missing` into `false`, which is what its callers want: every one of them gates on `exists()` first, so treating an absent session as a dead pane there would be wrong. The poller reads the probe directly and treats `Missing` as terminal, guarded twice:

- the existing two-tick debounce, since a tmux rename between resolution and probe makes the old name look absent for one tick;
- a 60s grace while the pane has never been seen alive, so a poller started microseconds before its tmux session registers is not reaped during that race.

`Unknown` (tmux could not be run at all) is never terminal. Stopping is otherwise the safe direction: both the TUI and the daemon run a repair pass that restarts a poller for any instance with a live tmux pane, so a poller stopped in error comes back within a refresh cycle, while a leaked slot never comes back.

**2. An advisory budget pre-check before the spawn setup.** That setup builds the retroactive capture-exclusion set (a `sessions.json` load plus a `canonicalize` syscall per stored peer) and only then reached the budget acquire, so with the budget full each repair pass burned ~100ms per instance on a spawn that was always going to be rejected. The repair pass runs twice a second, on the thread that also serves keystrokes. The budget-exhausted warning is now throttled to one line per minute: it is a standing state rather than an event.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No UI change.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the bug is a three-day accumulation in a long-lived process, which an e2e test cannot reach at a sane cost. Covered instead at the two tiers that can fail:
  - `missing_target_is_terminal_only_after_the_pane_was_seen_or_the_grace_passed` (unit, pure): tables the whole termination policy, including the never-seen-alive grace, the debounce, `Unknown`, and the recovered-tick reset.
  - `test_is_pane_dead_on_running_session` (existing tmux-backed test, extended): pins the actual `Alive` / `Missing` distinction against a real tmux server, and that `is_pane_dead` still reports `false` for a missing session so its `exists()`-gating callers are unaffected.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Found by sampling thread names in a live daemon (50 `aoe-poller/*` threads against 23 live tmux sessions), then cross-referencing each poller's instance id against `tmux ls`. The empty-stdout behavior was then confirmed directly against the local tmux server.

## Summary

- Stops orphaned session-ID pollers when their tmux sessions disappear.
- Tracks pane health with startup grace periods and debounce handling.
- Prevents poller setup when the shared budget is full.
- Throttles repeated budget warnings.
- Limits poller recovery and session lookup to live agent panes.
- Preserves existing `is_pane_dead` behavior.
- Adds tests for liveness, poller termination, budget handling, and agent-pane selection.

## Benefits

Active sessions can refresh their agent session IDs reliably. The change reduces unnecessary setup work and avoids treating unrelated terminal panes as agent panes.

## Potential follow-up

Consider adding monitoring for poller budget usage and more tests for unusual tmux failures.


njbrake (commented):

_Reviewed at `7f79aa58`. Push new commits and it will be re-reviewed._

---

### Summary

A session-id poller whose tmux session was killed outside AoE never terminated, holding one of 50 budget slots for the life of the process. The PR adds a four-state `probe_pane` so the poller can distinguish "pane exists and is dead" from "tmux cannot find this session", makes the latter terminal behind two guards, and adds an advisory budget pre-check so a full budget stops costing ~100ms of spawn setup per instance on a repair pass that runs twice a second.

### The central factual claim: verified against real tmux

I ran this on a dedicated socket against tmux 3.7c, with `aoe_test_*` session names, cleaned up afterward. The PR body's table is exactly right:

```
target=aoe_test_probe_live        exit=0 stdout='0'  stderr=''
target=aoe_test_probe_dead        exit=0 stdout='1'  stderr=''
target=aoe_test_probe_missing     exit=0 stdout=''   stderr=''
```

A `display-message` against a session tmux cannot find really does exit 0 with empty stdout and nothing on stderr, so the old `stdout.trim() == "1"` really did map "no such session" to "not dead". The bug is real and the diagnosis is correct.

Two things I measured that the body does not mention, and that matter for judging the change:

```
no server on socket               exit=1 stdout=''   stderr='no server running on /tmp/aoe_test_probe.sock'
target=aoe_test_pbi:99.0          exit=0 stdout='0'  stderr=''   # out-of-range window index
target=aoe_test_pbi:^.99          exit=0 stdout='0'  stderr=''   # out-of-range pane index
```

The second pair is the reassuring one. Every target tmux can resolve *to a session* prints a digit, including bogus window and pane indices, which it falls back from. Empty stdout is produced only when the session name itself does not resolve. So `Missing` is a precise signal, not a catch-all, and the "user has `pane-base-index 1` so `^.0` stops resolving" churn scenario I went looking for does not occur.

Budget is `SESSION_ID_POLLER_MAX_THREADS: u32 = 50` (`src/session/poller.rs:17`), consistent with the 50/50 measurement. And on `main` a poller in the missing-session state genuinely never exits: `poll_resolved_target` had no exit path other than `is_pane_dead` returning true twice, which it cannot for a missing session.

### Design Alignment

**Aligned.**

`is_pane_dead` was a bare helper in `src/tmux/utils.rs` with a single string comparison; the PR pushes the classification down into `probe_pane` and re-expresses `is_pane_dead` on top of it, rather than adding a second probe beside it. That is the right direction, and I confirmed it is behavior-preserving bit for bit: old code was `output().ok().and_then(utf8).map(|s| s.trim() == "1").unwrap_or(false)`; new code is `probe_pane(..) == Dead`, where `Dead` is exactly `trim() == "1"` and both failure modes (`.output()` fails, invalid UTF-8) become `Unknown`, which is not `Dead`. So all 40-odd `is_pane_dead` call sites are untouched in behavior.

The termination policy moves out of loose locals into a `TargetLiveness` struct that owns it. `poll_resolved_target` keeps the same injected-closure shape it already had, so the existing test seam survives. `TargetLiveness` has exactly one caller, but it is a state machine, not an abstraction layer, and it is what makes the policy unit-testable without a clock; that earns its place.

### On the four guards

**`is_pane_dead` callers gating on `exists()` first.** I enumerated all of them rather than taking the body's word. Substantively true; two are worth naming as not literally `exists()`-gated:

- `src/tmux/tool_session.rs:156` (`wait_until_ready`) has no `exists()` gate. Folding `Missing` into `false` is still what it wants: a not-yet-registered session keeps it waiting out its 200ms budget instead of bailing with "pane died before becoming ready".
- `src/tmux/mod.rs:730` iterates `live_names`, which came from `list-sessions`, so it is effectively gated.

The rest (`app.rs:3671`, `instance.rs:4225/6158/6260/6631/6672/7133`, `restart.rs:100`, `session.rs:558`, the `sessions.rs` trio) all gate explicitly. But this whole question is less load-bearing than the body implies, because `is_pane_dead` is behavior-identical pre- and post-PR. Nothing can regress here.

**The repair pass.** Verified, and the body understates it. `repair_session_id_poller_if_needed` (`instance.rs:6033`) is called from `HomeView::repair_session_id_pollers` on the `App::run` tick behind `STATUS_REFRESH_INTERVAL = 500ms` (`tui/app.rs:868,1813`), from `tui/home/mod.rs:2581` on reload, and from `drain_session_id_updates_in_state` on the daemon's 2s tick (`server/mod.rs:6334`). A poller stopped in error for a live instance comes back in 500ms or less under the TUI, 2s or less under the daemon. "Within a refresh cycle" is accurate.

**The two-tick debounce.** The race it covers is real: `resolve_target` resolves a name, tmux renames the session, and the probe then hits a name that no longer exists. The existing test `rename_between_resolution_and_liveness_retries_the_new_name` drives exactly that. Two ticks is enough because the rename completes between ticks and the next tick re-resolves; a rename that straddled two consecutive ticks would need the poll interval (2s minimum) to land inside the rename window, which is not a thing. And the recovered-tick reset means an intermittent target never accumulates toward a stop.

**The 60s grace.** A session that legitimately takes longer than 60s to register would be reaped, after the second gone tick, so realistically at ~124s given the 1.5x backoff from the 2s base. But this cannot strand anyone: pollers are only started from `finalize_launch` (after the tmux session is created) or from the repair pass (which requires `has_live_tmux_pane_in`), and the repair pass restarts within 500ms of the pane appearing. The grace is belt and braces on a race that is already covered.

**`Unknown` never terminal.** Confirmed at `poller.rs:256`, and mutation-tested below. It is reachable in production (spawn failure, invalid UTF-8), so it is not dead code.

### Naming: it is not tri-state

The body says "a tri-state `probe_pane`" and then lists four variants. The code has four: `Alive` / `Dead` / `Missing` / `Unknown` (`src/tmux/utils.rs:227-238`). Body wording only, no code impact.

### The advisory budget pre-check: the TOCTOU is benign

I traced both directions.

- **Pre-check true, acquire fails.** Worst case is one wasted setup, which is exactly what `main` pays unconditionally. In `repair_session_id_poller_if_needed` the pre-check is placed *before* `self.session_id_poller = None`, so a rejected spawn is a no-op rather than a downgrade; on `main` that same clear ran unconditionally. The PR strictly narrows the window.
- **Pre-check false, slot actually free.** Returns false, and the repair pass retries 500ms (TUI) or 2s (daemon) later. Transient, self-healing.

It cannot drop a poller that should have been spawned. I also checked that `maybe_start_poller_since` has no side effect other than starting the poller, so its early return is safe for its other caller (`finalize_launch`, `instance.rs:5056`).

**The throttle has no clock bug.** `Instant` is monotonic by construction, and `Instant::duration_since` saturates rather than panicking, so no overflow and no backwards-clock hazard. Two nits: a poisoned `LAST_BUDGET_WARN` silences the warning permanently (`budget_warn_due` returns `false` on `Err`), and the throttle is global rather than per-instance, so with 50 starved sessions you get one line naming one arbitrary instance per minute. Both acceptable.

### Issues Found

**1. The budget-exhausted warning becomes effectively unreachable. This is the one thing I would not merge without.**

`SessionPoller::start` has exactly one production caller: `maybe_start_poller_since` at `instance.rs:6010`. The new pre-check returns before that call whenever the budget is full. So once the budget is steadily full, nothing reaches the acquire, and the throttled warning fires only on the TOCTOU race, which is to say almost never. The log line that diagnosed this very bug ("Session-id poller budget exhausted (50/50)") disappears from a saturated daemon. If starvation recurs for a legitimate reason (50 genuinely live sessions), the user gets no signal at all that session ids stopped refreshing.

The throttle already solves the flood problem the body describes; the pre-check does not need to also delete the signal. Suggested shape, verified locally at +23/-11:

- Extract `warn_budget_exhausted(instance_id)` in `poller.rs`, same message, same 60s `BUDGET_WARN_INTERVAL` throttle.
- Call it from the acquire failure path as before, and also from both new pre-check early returns in `instance.rs` (`maybe_start_poller_since`, `repair_session_id_poller_if_needed`).

**2. Worth a doc line on the `Missing` variant.** Record that both the exit-0 and the exit-1 ("no server running") empty-stdout shapes mean the session is absent, and why a nonzero exit is deliberately not classified `Unknown`. See the CodeRabbit section for the reasoning; the point is to stop the next reader re-litigating it.

### Automated Review Check-in

CodeRabbit posted one actionable comment with two parts.

**"Check `output.status.success()` and return `Unknown` on a nonzero exit": disagree, on measurement.** I chased this empirically because it is the one place where the classification could reap a live poller. The only nonzero-exit shape I could produce is "no server running on `<socket>`" (exit 1, empty stdout). If the tmux server is gone, every session on it is gone too, so `Missing` is the correct answer and adopting the suggestion would reintroduce a poller leak in exactly that case: `Unknown` is never terminal, so all 50 pollers would hang until a server came back. Every target that resolves to a live session exits 0 and prints a digit, including out-of-range window and pane indices, so there is no live-session-with-nonzero-exit case to protect.

**"Use plain code formatting for the private `probe_pane` reference in the public `is_pane_dead` docs": false positive.** `src/tmux/utils.rs` is `pub(crate) mod utils` (`src/tmux/mod.rs:13`), so `is_pane_dead` is not publicly documented and rustdoc emits no `private_intra_doc_links` warning. I reproduced the actual CI gate (`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`, `ci.yml:121-127`) and it is clean.

**"Docstring coverage 62.50% below 80%" pre-merge check: disregard.** This is a CodeRabbit-side check, not a repo status gate, and AGENTS.md explicitly discourages docstrings that restate a signature. The undocumented functions are `probe_pane` (whose semantics live on the `PaneProbe` enum right above it) and `TargetLiveness::new`. Adding prose to satisfy this would make the diff worse.

### Mutation testing

I reverted each load-bearing guard one at a time and confirmed a test goes red. All five, plus the classification itself:

| Mutation | Test that caught it |
|---|---|
| `_ => PaneProbe::Alive` (the original bug restored) | `test_is_pane_dead_on_running_session`: `left: Alive, right: Missing` at `session.rs:3112` |
| `Missing => false` (never terminal) | `missing_target_...`: `left: (false, false)` at `poller.rs:799` |
| `Missing => true` (grace removed) | `missing_target_...`: `left: (true, false)` at `poller.rs:794` |
| `Unknown => true` | `missing_target_...`: `left: (true, false)` at `poller.rs:821` |
| `should_stop = true` (debounce removed) | `missing_target_...`: `left: (true, true)` at `poller.rs:802` |
| recovery no longer clears `gone_candidate` | `missing_target_...`: `left: (true, true)` at `poller.rs:831` |

Every guard is genuinely covered. Notably the two tiers split cleanly: the pure unit table covers the policy, and the tmux-backed test covers the `Alive`/`Missing` classification that the unit test injects as a given. Neither is redundant, and neither can pass with the bug present.

### Compliance with Project Guidelines

`cargo fmt --check` clean. `cargo clippy --all-targets` produces one warning, `items_after_test_module` in `src/process/macos.rs`, which is pre-existing on `main` and unrelated. `cargo test --lib`: 4739 passed, 0 failed, 73.6s. The CI rustdoc gate passes.

Test-cost rules followed well. The whole termination policy is one `#[test]` with a table of scenarios, not a test per variant, which is the house style. The two added assertions inside the existing `test_thread_budget_cap` reuse a `#[serial]` test that already had the `CountRestorer` guard, so no new serial test and no leaked global. No new e2e, no `#[allow(dead_code)]`, no emdashes or double-hyphen separators in the new prose (I grepped the added lines).

**On the e2e question:** AGENTS.md says session-lifecycle changes should *consider* an e2e, and the author's skip is justified here. The behavior is "a poller thread stops"; there is no CLI or TUI surface that observes thread count, so an e2e would have to assert on log output or timing, at ~2s and strictly serial forever, to cover something two cheaper tiers already cover with mutation-proven tests. I agree with the decision and with how it is argued in the PR body's coverage section.

One coverage gap worth naming: nothing tests the *wiring*, that the poller loop actually passes `probe_pane` into `poll_resolved_target` (`poller.rs:439`). The type change from `impl FnOnce(&str) -> bool` to `-> PaneProbe` makes a wrong wiring a compile error, so this is close to compiler-proven, but a closure returning a constant would still typecheck. Not worth a test.

### Documentation

Proportionate. No new standalone markdown, no narration of the change. The added rustdoc is all non-obvious "why": the tmux exit-status behavior nobody would guess, why `Missing` folds to `false` for `is_pane_dead` but not for the poller, why the pre-check is advisory. That is the most valuable kind of prose in a diff. Nothing outside the diff went stale; I grepped `docs/` and `website/` for the budget constant and `is_pane_dead` and found no hand-maintained mirror.

### Merge Risk

**Low.**

Blast radius is bounded to session-id refresh. The failure mode of a wrong reap is that one session's agent session id stops updating for at most one refresh cycle, and the repair pass at 500ms/2s puts it back. Reverts cleanly: no schema, no migration, no persisted state.

The one residual behavior worth knowing about, which I could not make into a bug but which is the most likely way this shows up in a log: the repair gate and the poller probe resolve *different* names. `has_live_tmux_pane_in` goes through `live_any_kind_name_for_id_in` (`tmux/mod.rs:676`), which accepts an agent, terminal, *or* container name for the id; the poller probes `live_agent_session_name`, the agent name specifically. An instance whose agent session is gone but whose terminal or container session is still live therefore passes the repair gate and gets a poller whose probe returns `Missing` forever. Post-PR that becomes a spawn/reap cycle roughly every two minutes instead of `main`'s single permanently-leaked slot. Budget occupancy is the same in that sub-case; the new cost is ~100ms of setup and a thread create/destroy every ~124s per such instance, which is a ~0.1% duty cycle. Not a regression and not worth blocking on, but if you see `aoe-poller/<id>` threads recycling in a steady loop, this is why, and narrowing the repair gate to the agent shape would be the follow-up.

Untested path: the loop wiring at `poller.rs:439`, discussed above, effectively compiler-proven.

### CI Workflow Approval

Not acted on. The `CI` and `Tests` runs for `7f79aa58` sit at `action_required` because this is a fork PR. For what it is worth, the exposure audit passes cleanly: the diff is four `.rs` files, no workflow or CI config change, no install/build hook, no dependency or lockfile churn, no binary or obfuscated content, and every changed path was read in full.

### Verdict

**Design Alignment:** Aligned
**Implementation:** Approve
**Project Fit:** Accept
**CI Workflow Approval:** Not acted on (fork run pending at `7f79aa58`; audit clean)
**Merge Risk:** Low; a terminal-alive/agent-dead instance trades one leaked slot for a ~2min spawn/reap cycle, which is not worse than today

This is a well-diagnosed bug with a fix that matches the diagnosis. The central factual claim about tmux is true, I reproduced it, and I also reproduced the cases the body did not mention, which turned out to strengthen the change rather than undermine it: `Missing` is produced only by an unresolvable session name, not by any resolvable-but-odd target. Every guard is mutation-covered, and I could not construct a case where a live session loses its poller for longer than a repair cycle. The one thing I would not merge without is the observability fix in Issue 1: the pre-check makes budget exhaustion silent, and that is the signal that found this bug in the first place.

src/tmux/utils.rs
```diff
@@ -218,19 +218,60 @@ pub fn append_window_size_args(args: &mut Vec<String>, target: &str) {
...
+        return PaneProbe::Unknown;
+    };
+    let Ok(stdout) = String::from_utf8(output.stdout) else {
+        return PaneProbe::Unknown;
+    };
+    match stdout.trim() {
+        "1" => PaneProbe::Dead,
+        "0" => PaneProbe::Alive,
+        _ => PaneProbe::Missing,
+    }
```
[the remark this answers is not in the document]
Eric162: Not taking either part. @njbrake reproduced both against tmux 3.7c in his review and I agree with his reasoning.

**`output.status.success()` -> `Unknown` on a nonzero exit.** The only nonzero-exit shape either of us could produce is `no server running on <socket>` (exit 1, empty stdout). If the tmux server is gone then every session on it is gone too, so `Missing` is the correct classification. Adopting this would reintroduce the leak the PR exists to close in exactly that case: nothing terminates on `Unknown`, so all 50 pollers would hang until a server came back.

The precision worry behind the suggestion also does not hold. Every target tmux can resolve *to a session* expands `#{pane_dead}` to a digit, including out-of-range window and pane indices (`:99.0`, `:^.99`), which it falls back from. Only an unresolvable session name yields empty stdout, so `Missing` is a precise signal rather than a catch-all, and a user's `pane-base-index` cannot make a live pane read as `Missing`.

I have written all of that onto the `PaneProbe::Missing` variant in `ee79d5d`, so the next reader gets the reasoning instead of re-litigating it.

**Plain code formatting for the `probe_pane` reference.** False positive. `src/tmux/utils.rs` is `pub(crate) mod utils` (`src/tmux/mod.rs:13`), so `is_pane_dead` is not publicly documented and rustdoc emits no `private_intra_doc_links` warning. njbrake ran the actual CI gate (`RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`, `ci.yml:121-127`) and it is clean.

njbrake (commented):

_Reviewed at `ee79d5d8`. Push new commits and it will be re-reviewed._

---

Reviewed the new head rebased onto current `origin/main` (`fdb46c3f`), which rebased cleanly with no conflicts. One new commit since my last pass, "fix(session): keep the budget-exhausted warning reachable from the pre-checks".

## 1. The blocking finding is fixed

I blocked because the advisory pre-check returns before `start_observations`, so once the budget was steadily full nothing reached the acquire and its throttled warning fired only on the pre-check race. A saturated process would go silent about starving its sessions of session-id refresh, and that log line is what found the orphaned-poller bug.

The new commit extracts `warn_budget_exhausted(instance_id)` and calls it from three places: the acquire failure inside `start_observations`, the pre-check early return in `maybe_start_poller_since`, and the pre-check early return in `repair_session_id_poller_if_needed`. Same message, same 60 second throttle.

I re-enumerated the paths rather than trusting the commit message. `SessionPoller::start` still has exactly one production caller, and the three warn sites are now the complete set of places a session can go without a poller for budget reasons. There is no fourth path. The placement in the repair pass is also right: it sits after the `is_structured` / `supports_session_poller` / `session_id_poller_is_running` / `has_live_tmux_pane_in` gates, so it only reports instances that genuinely want a poller and cannot have one, which is exactly the starved set rather than every row in the store.

**This finding is cleared.** The fix does what I asked and I verified it by reading every path, not by reading the commit message.

**One nit, and I want to be precise that it is a nit rather than a reopening.** The commit adds a test, `budget_warn_reports_once_per_window`, but that test does not cover the thing that was blocking. I mutation-tested all three pieces:

- Removing both pre-check calls to `warn_budget_exhausted`, which is a byte-for-byte revert of the fix, leaves the whole suite green: `test result: ok. 4741 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out`.
- Gutting `warn_budget_exhausted` into a no-op also leaves it green.
- Only breaking the throttle itself goes red: `test session::poller::tests::budget_warn_reports_once_per_window ... FAILED`, panicking at `poller.rs:925`.

So the new test pins `budget_warn_due`'s window mechanics, which were already present at the previous head and were not the problem, and nothing defends the reachability that was. I am not asking for a test here. The fix is three call sites at early returns, visible in the diff and verifiable by reading, and the original omission was a design gap that no test would have caught either. A `#[traced_test]` assertion on the message would cover the function but still not the wiring, and the wiring test needs a full `Instance`, which is the expensive tier. Worth knowing when reading the coverage claim, not worth paying for.

The test itself is well behaved: it is correctly `#[serial]` because `LAST_BUDGET_WARN` is process-global state, it snapshots and restores the previous value, and it backdates the stamp instead of sleeping out the 60 second interval, so it stays instant. It does not restate a constant.

## 2. Disposition of every other finding from my last review

**"tri-state `probe_pane`" wording in the PR body.** Fixed. The body now reads "A four-state `probe_pane`", matching the enum.

**CodeRabbit's `status.success()` suggestion.** Resolved the way I recommended, and better than I recommended. The author declined it, replied on the thread with the reasoning, and wrote the reasoning onto the `PaneProbe::Missing` variant so the next reader gets it instead of relitigating. I re-measured every claim in that new doc comment against tmux 3.7c on a dedicated socket with `aoe_test_*` names, cleaned up after:

```
live pane                exit=0 stdout='0'  stderr=''
dead pane                exit=0 stdout='1'  stderr=''
unknown session name     exit=0 stdout=''   stderr=''
existing session :99.0   exit=0 stdout='0'  stderr=''
existing session :^.99   exit=0 stdout='0'  stderr=''
no server on socket      exit=1 stdout=''   stderr='no server running on <socket>'
```

Every sentence in the doc comment is accurate, including the out-of-range index fallback and the exact stderr string. Cleared.

**CodeRabbit's private intra-doc link.** Still a false positive, and still confirmed: I re-ran the actual CI gate on this head with `RUSTDOCFLAGS="-D warnings"` and `cargo doc --no-deps --features serve` and it is clean, and the `Docs` job in CI passed on this head too.

**The five load-bearing guards.** Re-verified on the rebased tree, since both a second commit and a new `main` sit underneath now. Spot-checking the three that matter most, each still goes red:

- reverting the classification to `_ => PaneProbe::Alive`, which is the original bug: `test_is_pane_dead_on_running_session ... FAILED`, `left: Alive`, `right: Missing`.
- removing the 60 second startup grace: `missing_target_is_terminal_only_after_the_pane_was_seen_or_the_grace_passed ... FAILED`, `left: (true, false)`.
- making `Unknown` terminal: same test `FAILED`, `left: (true, false)`.

**The repair gate versus the poller probe target.** Still open, still not a regression, still not blocking. `has_live_tmux_pane_in` goes through `live_any_kind_name_for_id_in`, which accepts an agent, terminal, or container name for the id, while the poller probes the agent name specifically. An instance whose agent session is gone but whose terminal or container session is still live passes the repair gate and gets a poller that reads `Missing` forever, so post-merge that becomes a spawn and reap cycle roughly every two minutes instead of today's single permanently leaked slot. Budget occupancy is unchanged in that sub-case; the new cost is a thread create and destroy plus the setup, about a 0.1 percent duty cycle. Neither commit touched either resolution function and the new `main` commit did not either. Follow-up material, not merge material.

**The poisoned-lock case silencing the warning permanently.** Not changed, but now documented on the throttle: "A poisoned lock silences the warning rather than risking a flood." That is the right call written down, which is what I wanted.

**The untested wiring between the poller loop and `probe_pane`.** Still true, still effectively compiler-proven by the closure signature change from returning `bool` to returning `PaneProbe`. Unchanged.

**Nothing newly regressed.** I diffed the full PR against the new `main` and re-read both commits.

## 3. On the `silent_orphan_fires_on_cost_then_silence` flake

Two things to separate here, and the first one matters.

**That item did not come from my review of this PR.** My last pass never mentioned `silent_orphan_fires_on_cost_then_silence`, and there is no byte-proxy fixture in this diff. The test lives in `tests/integration/acp_silent_orphan.rs`, this PR does not touch it, and it has not changed on `main` since `cae24e55` (PR #2860). I think the item belongs to a different PR's review. I did the work anyway rather than dismiss it on provenance, and the result is worth having.

**I could not reproduce any flake, and I found why a baseline for it can be false.** My first attempt returned a clean 16 out of 16, which I did not believe, so I re-ran with `--nocapture` and got:

```
skipping: shim deps not installed; run `cd acp-worker/test-shim && npm ci` first
test acp_silent_orphan::silent_orphan_fires_on_cost_then_silence ... ok
```

The test early-returns on a missing Node shim and still reports `ok`. `acp-worker/test-shim/node_modules` is gitignored, so **every fresh worktree silently skips this test and reports a pass**, in 0.01 seconds rather than the real 5 seconds. I confirmed the same on a fresh `main` worktree. That is a trap worth knowing about generally: any pass rate measured for this test in a fresh worktree is vacuous. It cannot manufacture failures, so it cannot explain a 5 in 20 failure rate, but it can absolutely manufacture a clean baseline, and a real rate on one side compared against a vacuous 0 for 12 on the other would produce exactly the pattern described.

After running `npm ci` in `acp-worker/test-shim` on both trees, so the test genuinely executes at about 5 seconds a run:

- PR head, the single test in isolation: 20 passed, 0 failed of 20.
- PR head, the whole `acp_silent_orphan` module: 8 rounds, 6 tests each, 48 executions, 0 failures, deterministic at 13.7 seconds a round.
- PR head, the whole `integration` target, which is the CI condition and exercises the serial ordering and env leakage the file's own comments warn about: 3 rounds, `171 passed; 0 failed` each time.
- `main`, the single test in isolation: 20 passed, 0 failed of 20.

That is 88 genuine executions of the test with zero failures, plus 513 executions of the surrounding target. CI agrees: `Cargo Test (macos)` passed on this head. I cannot reproduce the reported rate and I have no evidence of a defect in the runner cancel path. If the original measurement was taken in a worktree where one side had the shim installed and the other did not, that would explain the asymmetry, and it is cheap to check before spending more on it.

## 4. The CI failure is not this PR's fault, and here is direct evidence rather than an argument

First, a correction on the shape of it. The two failing checks are one failure. `Playwright (live)` is an aggregator job whose entire body is a shard-result check; it exists because the branch ruleset targets a stable check name rather than per-shard matrix contexts. `Playwright (live, shard 2/2)` passed. So it is one spec failing in one shard, reported twice.

Everything else on the head is green: all 12 `CI` jobs including `Clippy`, `Format`, `Docs`, `Nix Build`, and `Supply Chain`, and 9 of 11 `Tests` jobs including all four `Cargo Test` matrix legs on linux and macos, `Playwright (mocked)`, and `Vitest`.

I did not reason from "main is usually green". I built both trees and ran the suite locally, interleaved so both sides shared the same machine conditions.

**The spec in isolation.** 32 runs on the PR head and 32 on `main`. The PR head failed once, on the very first run immediately after a cold build, and passed the other 31. `main` passed 32 of 32. On its own that is nowhere near significant, which is why I kept going.

**The suite under the CI condition.** The live config runs three workers with `fullyParallel` and `retries: 0`, and the CI failure happened alongside 82 passing tests, so contention is the condition to reproduce. Running the whole `acp-stories` directory at three workers, interleaved between the two binaries:

- PR head: 6 rounds, 4 of them had at least one failing spec, 7 test failures total.
- `main`: 9 rounds, 8 of them had at least one failing spec, 8 test failures total.

`main` flaked slightly more often than the PR head. In four rounds where I captured names, `main` failed `acp-stories/queue-follow-up-with-nav.spec.ts:54` three separate times with no PR code anywhere in the process.

**CI history at spec granularity.** Sweeping the recent failing `Tests` runs and pulling the annotation for each failed Playwright job, the victim rotates across branches, `main` included:

```
2026-08-27  fix/reap-orphaned-session-id-pollers   rate-limit-reset-across-turns
2026-08-27  structured-approval                    composer-stop-during-thinking, composer-stop-during-tool-call
2026-08-26  fix/tmux-paint-zero-fork               queue-follow-up-with-nav
2026-08-26  fix/tmux-paint-zero-fork               composer-stop-during-tool-call
2026-08-26  main                                   composer-stop-during-tool-call
2026-08-26  main                                   composer-stop-during-sub-agent, composer-stop-during-thinking, queue-follow-up-with-nav
2026-08-26  fix/remove-the-serve-cargo-feature     composer-stop-during-thinking
2026-08-25  fix/worktree-reconcile-stale-...       composer-stop-during-sub-agent, queue-follow-up-with-nav
2026-08-25  main                                   queue-follow-up-with-nav
2026-08-25  fix/omp-custom-spinner-frames          wizard-scratch-delete-cleans-tempdir
```

**And the repo already documents this exact mode**, in `web/playwright.live.config.ts`, next to the `workers: 3` setting:

> "The failing spec rotated across runs (queue follow up, stop-during-sub-agent, profile-switch-view, etc.) because the contention bit a different worker each time, not because the spec itself was broken: all of those passed locally under the CI-matching debug+AOE_COVERAGE config at 4 workers in isolation but flaked when the full suite ran on the GH Actions runner."

**Mechanism check, separately from the statistics.** The failing spec seeds a session with `aoe add`, switches it to structured view, and drives a fake ACP agent. Structured sessions are excluded from `repair_session_id_poller_if_needed` by its `is_structured` gate before any of this PR's code runs, and `maybe_start_poller_since` is only reached from `finalize_launch`, which a seeded-then-structured session never goes through. So the PR's changed code is not executed in that test at all, and the only thing the PR does to the daemon loop is remove work.

**Conclusion: the failure is a known contention flake in the live suite, present on `main` at the same or higher rate, and not attributable to this PR.** A re-run should clear it. Separately, and not this PR's problem, a suite that fails a merge gate roughly half the time it runs is worth its own issue; `retries: 0` was a deliberate choice to surface flakes rather than mask them, and this is the surfacing working.

## 5. House rules on the new commit

`cargo fmt --check` clean. `cargo clippy --all-targets` produces one warning, `items_after_test_module` in `src/process/macos.rs`, which is pre-existing on `main` and untouched here; I diffed before attributing. `cargo test --lib`: 4741 passed, 0 failed. `cargo test --features serve --test integration`: 171 passed, three times. The CI rustdoc gate passes.

No emdashes and no double hyphens as separators anywhere in the added prose, checked across both the whole PR diff and the new commit alone. No `#[allow(dead_code)]` and no dead code: `warn_budget_exhausted` has three callers and `PaneProbe::Unknown` is still reachable in production through a spawn failure or invalid UTF-8. No new e2e, so the `#[parallel]` rule does not apply; the one new `#[serial]` unit test is correctly serial because it mutates a process-global static, which is exactly the carve-out AGENTS.md names. The new comments explain non-obvious why rather than restating code, and the `PaneProbe::Missing` doc in particular is the kind of prose that saves the next reader a tmux session. No breaking change to call out: `is_pane_dead` remains behavior-identical to its pre-PR form for all of its callers.

On the e2e question, my previous assessment stands. The behavior is "a poller thread stops", which has no CLI or TUI surface to assert on, and it is covered at two cheaper tiers that I have now mutation-proved twice. Skipping the e2e is the right call and is argued honestly in the body.

## 6. PR body versus what is pushed

Nothing in the body is false, and the one wording error I flagged last time was corrected. There are two omissions, both from the second commit:

- Section 2 still says only "The budget-exhausted warning is now throttled to one line per minute: it is a standing state rather than an event." That remains true, but it does not mention that the warning is now also emitted from both pre-check paths, which is the change that made the throttle safe to keep.
- The Test Coverage Analysis lists two tests and does not mention `budget_warn_reports_once_per_window`.

Worth a two-line body edit so a future reader of the merged PR sees the design that actually shipped, but nothing here misrepresents the code.

## 7. Verdict

**Design Alignment:** Aligned
**Implementation:** Approve
**Project Fit:** Accept
**CI:** One failure, reproduced on `main` locally at the same rate, not this PR's
**Merge Risk:** Low

**Blocking findings: none. The one I raised is fixed and I verified it by tracing every path, not by reading the commit message.**

**Nits, none of which should hold the merge:** the new test covers the throttle rather than the reachability the commit is about, so a revert of the fix stays green; the PR body does not describe the second commit; and the terminal-alive with agent-dead churn remains as follow-up material.

Eric162 did exactly what was asked, wrote the reasoning into the code where the next reader will find it rather than leaving it in a PR thread, and correctly declined the bot suggestion that would have reintroduced the leak. The tmux measurements the whole change rests on reproduce exactly on tmux 3.7c, every load-bearing guard goes red under mutation, and the only red check is a suite that fails on `main` just as often. I recommend merging it, with a re-run of the live shard.

src/tmux/utils.rs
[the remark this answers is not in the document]
jerome-benoit: The non-zero no-server case should indeed remain `Missing`; simply mapping every unsuccessful status to `Unknown` would reintroduce the leak. However, the current resolution still collapses every other empty-output failure into `Missing`.

This repository already encodes the required distinction in `tmux_no_server_running`: no-server and ENOENT are empty states, while EACCES, ENOTSOCK, empty stderr, and other unexpected failures remain errors. After a pane has been seen alive, two such failures here stop its poller.

Please use `tmux_query_command`, classify recognized no-server/ENOENT and successful empty output as `Missing`, and classify unexpected non-zero or malformed output as `Unknown`. A pure table-driven classifier can cover both sides without weakening orphan cleanup.

src/session/instance.rs
```diff
@@ -6037,6 +6046,15 @@ impl Instance {
...
+        // after paying for its setup. Checked before the handle is cleared so
+        // an exhausted budget is a no-op rather than a downgrade. Warns here
+        // because this early return is what the acquire's own warning would
+        // otherwise have reported.
+        if !crate::session::poller::session_id_poller_budget_available() {
+            crate::session::poller::warn_budget_exhausted(&self.id);
+            return false;
+        }
         self.session_id_poller = None;
         self.maybe_start_poller();
```
jerome-benoit: **[P2] Keep poller repair scoped to the agent pane**

This repair accepts any live pane through `has_live_tmux_pane_in`, and `maybe_start_poller` seeds the replacement with the same any-kind lookup. If the agent pane is gone but a paired terminal remains, `live_agent_session_name` sees that derived terminal name in the cache and returns it through the fast path before applying `NameShape::agent`. The replacement therefore probes the terminal as `Alive` and keeps the budget slot without a live agent. Use an agent-only lookup for both the repair gate and initial poller name, and cover `agent missing + paired terminal alive`.

src/session/instance.rs
```diff
@@ -5740,6 +5740,15 @@ impl Instance {
         if !self.supports_session_poller() {
             return;
         }
+        // The setup below loads `sessions.json` and canonicalizes a path per
+        // stored peer to build the capture-exclusion set. Skip all of it when
+        // the poller budget is already full, since `start_observations` would
+        // reject the spawn anyway. The warning moves here with the decision:
+        // the acquire this replaces is no longer reached.
+        if !crate::session::poller::session_id_poller_budget_available() {
```
jerome-benoit: **[P3] Check actual poller eligibility before consuming the warning window**

`supports_session_poller` only checks the resume strategy, but this function later returns without constructing a poller for Kiro/Qwen and for sandboxed Copilot/Kimi. At full capacity, one of those instances can therefore consume the global once-per-minute warning even though it would never have requested a slot, masking the identifier of a genuinely starved instance. Apply a cheap exact eligibility check before the budget pre-check, while keeping the expensive exclusion-set setup afterward.

jerome-benoit (changes requested):
Reviewed at `305ee0bcc2b33c1164cccb71ea8158a716c3c418`.

The exact-head CI is green, but two lifecycle defects remain and one diagnostic path is inaccurate.

Most importantly, the existing review’s conclusion that an agent-dead/terminal-live repair merely cycles every two minutes does not hold. The repaired poller can retain the terminal name through `session_name_from_cache`’s derived-name fast path and therefore keep probing the terminal indefinitely.

I replied to the existing `probe_pane` thread rather than duplicating it inline: the repository already distinguishes recognized no-server conditions from EACCES, ENOTSOCK, and other unexpected tmux failures, while this probe currently collapses all empty or malformed output into `Missing`.

src/session/instance.rs
Eric162: Confirmed and fixed. This is the same leak the PR is about, in a case I missed.

Traced it: `tmux_env_session_name` goes through `live_any_kind_name_for_id`, which returns `agent_hit.or(terminal_hit).or(container_hit)`. So with the agent gone and a paired terminal up, the replacement poller is seeded with the terminal's name. In the loop, `live_agent_session_name(id, derived = <terminal name>)` hits `session_name_from_cache`'s fast path (`names.contains_key(derived)` is true, the terminal really is live) and returns it before `NameShape::agent` is ever applied. `probe_pane` then reads `Alive` and the slot is held for an agent that is gone.

Worth noting this also corrects a prediction in @njbrake's review: he expected this sub-case to become a ~2 minute spawn/reap cycle post-PR. It does not, because the probe never returns `Missing`; it is a permanently held slot, same as `main`.

Fixed in `4ae3569`:

- `live_agent_name_for_id` / `live_agent_name_for_id_in` in `src/tmux/mod.rs`, agent shape only.
- `Instance::agent_tmux_session_name` and `Instance::has_live_agent_pane_in`.
- `repair_session_id_poller_if_needed` gates on `has_live_agent_pane_in`, and `maybe_start_poller_since` seeds from `agent_tmux_session_name`.

`has_live_tmux_pane_in` keeps the any-kind lookup for its other callers (peer exclusion in `capture.rs`, TUI reload), where any live pane for the id genuinely is the question.

Covered by `live_agent_name_for_id_ignores_terminal_and_container_panes`, which asserts the `agent missing + paired terminal alive` case returns `None` and that the any-kind lookup still answers it, so the two do not drift together.

src/session/instance.rs
Eric162: Correct. `supports_session_poller` only rules out `ResumeStrategy::Unsupported`, so kiro (`Flag`) and qwen (`FlagPair`) reach the poll-fn match and hit `_ => return`, and sandboxed copilot/kimi return from inside their arms. All of those are after the budget pre-check, so at capacity any of them could burn the global once-per-minute window and hide the id of an instance that actually wanted a slot.

Fixed in `4ae3569` with `constructs_session_poller`, checked alongside `supports_session_poller` before the budget pre-check. The expensive exclusion-set build stays where it was.

On the obvious risk with that fix, a mirrored predicate drifting from the match it mirrors: `poller_construction_matches_eligibility` tables every case, and then asserts every entry in `crate::agents::AGENTS` appears in the table. Adding an agent fails the test until someone decides which side it belongs on, rather than silently defaulting to "eligible".

I did consider restructuring so no predicate is needed (compute `extra_excludes` lazily, or have the match yield `Option<SessionIdPollFn>` so the budget check could sit after it). Both are a wider change to the spawn path than this observability fix justifies, so I went with the mirrored predicate plus the drift test. Say the word if you would rather have the restructure.

src/session/instance/polling.rs
```diff
@@ -330,10 +357,18 @@ impl Instance {
         if self.is_structured()
             || !self.supports_session_poller()
             || self.session_id_poller_is_running()
-            || !self.has_live_tmux_pane_in(snapshot)
+            // Agent pane, not any pane: a terminal outliving the agent is not
+            // something a session-id poller can follow.
+            || !self.has_live_agent_pane_in(snapshot)
```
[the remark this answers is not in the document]
Eric162: Right, and it was an incomplete fix on my part: I added `constructs_session_poller` to the spawn path but not to the repair gate, which calls `warn_budget_exhausted` too. Same finding as jerome-benoit's P3, in the other of the two call sites.

Fixed in `c9d3535`. Repair now gates on it alongside `supports_session_poller`.

On the regression test: the return value cannot carry it, since `repair_session_id_poller_if_needed` returns `false` either way (a kiro spawn builds no poller regardless). `repair_declines_a_resume_capable_tool_the_match_cannot_build_for` asserts through the handle instead, with a snapshot that *does* contain a live agent pane so the pane lookup cannot be what declines: proceeding past the gate clears `session_id_poller` before the spawn that was never going to happen, so the handle surviving is the observable.

Verified it catches the mutation (eligibility term removed from the gate):

```
kiro must be declined before the handle is cleared
```

src/tmux/mod.rs
```diff
@@ -671,6 +671,37 @@ impl LiveSessionSnapshot {
...
+    snapshot: &LiveSessionSnapshot,
+    session_id: &str,
+) -> Option<String> {
+    let names = snapshot.names()?;
+    let suffix = id_suffix(session_id);
+    let agent = NameShape::agent(&suffix);
+    names
+        .iter()
+        .map(String::as_str)
+        .find(|name| agent.matches(name) && !snapshot.pane_dead(name))
```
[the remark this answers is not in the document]
Eric162: Verified and fixed. This one is a regression my change introduced, so thanks for catching it.

`NameShape::agent` excludes `AGENT_EXCLUDED_PREFIXES`, and its own doc says why: "A name whose sanitized title pushes it under an excluded prefix fails here ... mistaking a paired terminal for the agent pane would be worse than not resolving at all." That trade is right for a name-only match. It is wrong for an id -> live-name lookup, where the caller knows which name is its own.

Concretely, a session titled so it sanitizes to `term_...` gets the agent name `aoe_term_rewriting_<id8>`. Before this PR, `has_live_tmux_pane_in` used the any-kind lookup, matched that as a *terminal*, returned `Some`, and repair proceeded. After my agent-only change it returned `None`, so that session's poller would never be repaired again.

Fixed in `c9d3535`: both helpers take the instance's title-derived agent name and accept it verbatim, still requiring the id suffix, and fall back to `NameShape::agent` for every other name. `Instance::derived_agent_session_name` supplies it at both call sites.

Covered by four cases added to `live_agent_name_for_id_ignores_terminal_and_container_panes`, direct and snapshot: the shadowed name resolves when its derived name is supplied and not otherwise, a derived name carrying a different id suffix is still rejected, and a genuine paired terminal stays excluded even when some other derived name was supplied. Verified the mutation is caught:

```
assertion failed: the instance's own derived name is accepted verbatim
```

Also note the PR is rebased onto current `main`, which split `src/session/instance.rs` into `src/session/instance/`, so these live in `instance/polling.rs` and `instance/tmux_session.rs` now.

src/session/instance/polling.rs
```diff
@@ -28,14 +28,41 @@ impl Instance {
...
+    /// no arm, and copilot and kimi read a host store a sandbox cannot reach.
+    ///
+    /// Checked before the budget pre-check so an instance that would never
+    /// have asked for a slot cannot consume the warning window and mask a
+    /// genuinely starved one. Kept beside the match it mirrors; the table in
+    /// `poller_construction_matches_eligibility` pins the two together.
+    fn constructs_session_poller(&self) -> bool {
+        match self.tool.as_str() {
+            "copilot" | "kimi" => !self.is_sandboxed(),
+            "claude" | "opencode" | "vibe" | "pi" | "codex" | "gemini" | "hermes" | "omp" => true,
```
jerome-benoit: [P1] Keep Prime Agent on the session-id poller path

This new allowlist omits `prime-agent`, so both initial startup and repair return before the still-present `prime-agent` match arm can construct `prime_agent_poll_fn`. `AgentDef` declares `ResumeStrategy::Flag("--resume")`, and the base code reached this arm for non-sandboxed sessions. As written, Prime Agent no longer refreshes its captured conversation ID and can lose resume persistence after restart. Add the same host-only eligibility used by the match arm and change the table to cover host `true` and sandbox `false`.

src/tmux/mod.rs
```diff
@@ -712,6 +712,64 @@ impl LiveSessionSnapshot {
...
+/// (mistaking a paired terminal for the agent pane is worse than not
+/// resolving), but wrong for an id -> live-name lookup, where the caller knows
+/// which name is its own and would otherwise never see a live agent at all.
+fn is_agent_name_for(
+    name: &str,
+    agent: &NameShape<'_>,
+    derived: Option<&str>,
+    suffix: &str,
+) -> bool {
+    if derived == Some(name) {
```
jerome-benoit: [P2] Do not let an ambiguous derived name authenticate a terminal pane

The exact-name exception can still recreate the leak this PR is fixing. An agent initially titled `Foo` can have a paired terminal named `aoe_term_Foo_<id>`. If a smart rename persists `term_Foo`, the new agent name is exactly that terminal name; `rekey_session` then fails on the collision while the new title remains durable. Once the agent disappears, `derived == Some(name)` accepts the surviving terminal as the agent and repair seeds a poller from it. Require durable pane-kind evidence or fail closed for this ambiguous collision, and cover old title `Foo`, persisted title `term_Foo`, agent absent, terminal live.

jerome-benoit (changes requested):
Re-review at exact head `c9d353551cec334030e52419bd29d53facce5902`, cross-validated through two independent sequential analyses and a final code-level arbitration.

The previous agent-pane scoping and pre-budget eligibility findings are fixed for the ordinary cases. Two additional blocking issues remain inline.

I did not duplicate two existing findings that remain valid:

- The resolved `probe_pane` thread still correctly requires recognized no-server/ENOENT results to map to `Missing`, while EACCES, ENOTSOCK, malformed output, and other unexpected failures remain `Unknown`: https://github.com/agent-of-empires/agent-of-empires/pull/3558#discussion_r3887220694
- The active `tmux_query_command()` finding is valid because this new session-name query otherwise loses the repository's UTF-8 and stable-locale guarantees: https://github.com/agent-of-empires/agent-of-empires/pull/3558#discussion_r3900149338

`git diff --check 51ee1e81e3411f27077e2303bab8824d429374d0 c9d353551cec334030e52419bd29d53facce5902` is clean. No local build or test was run because the host is already under load. GitHub currently reports only title/template checks for this head, not Rust CI.

src/session/instance/polling.rs
```diff
@@ -28,14 +28,41 @@ impl Instance {
...
+            _ => false,
```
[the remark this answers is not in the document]
Eric162: Confirmed, and this was the worst thing in the PR. Fixed in `246368f6`.

`prime-agent` has a live arm at `polling.rs:312` building `prime_agent_poll_fn`, host-only like copilot and kimi, and `ResumeStrategy::Flag("--resume")`. My allowlist put it in the ineligible group, so both startup and repair returned before that arm: no session-id capture, and resume lost across a restart.

Root cause worth naming, because it is the interesting part: `prime-agent` landed on `main` between when I wrote the allowlist and when I rebased, and I added it to the table mechanically as ineligible without checking for an arm. My drift test forced a decision and I made the wrong one, so the test cannot be the safeguard.

So the predicate is now a **denylist**:

```rust
match self.tool.as_str() {
    "kiro" | "qwen" => false,                                  // no arm in the match
    "copilot" | "kimi" | "prime-agent" => !self.is_sandboxed(), // host store
    _ => true,
}
```

The two failure directions are not symmetric, and the fix should reflect that: a wrongly-eligible instance spends one throttled log line, while a wrongly-ineligible one silently stops capturing its session id. A new agent now defaults to eligible.

The test asserts that property directly rather than restating my beliefs: every arm in the match must be eligible, the sandbox split for the three host-only ones, `kiro`/`qwen` ineligible, and then every other entry in `AGENTS` must default to eligible. Verified it catches the regression:

```
prime-agent has an arm in the match and must not be declined
```

src/session/instance/polling.rs
Eric162: Confirmed, and this was the worst thing in the PR. Fixed in `246368f6`.

`prime-agent` has a live arm at `polling.rs:312` building `prime_agent_poll_fn`, host-only like copilot and kimi, and `ResumeStrategy::Flag("--resume")`. My allowlist put it in the ineligible group, so both startup and repair returned before that arm: no session-id capture, and resume lost across a restart.

Root cause worth naming, because it is the interesting part: `prime-agent` landed on `main` between when I wrote the allowlist and when I rebased, and I added it to the table mechanically as ineligible without checking for an arm. My drift test forced a decision and I made the wrong one, so the test cannot be the safeguard.

So the predicate is now a **denylist**:

```rust
match self.tool.as_str() {
    "kiro" | "qwen" => false,                                  // no arm in the match
    "copilot" | "kimi" | "prime-agent" => !self.is_sandboxed(), // host store
    _ => true,
}
```

The two failure directions are not symmetric, and the fix should reflect that: a wrongly-eligible instance spends one throttled log line, while a wrongly-ineligible one silently stops capturing its session id. A new agent now defaults to eligible.

The test asserts that property directly rather than restating my beliefs: every arm in the match must be eligible, the sandbox split for the three host-only ones, `kiro`/`qwen` ineligible, and then every other entry in `AGENTS` must default to eligible. Verified it catches the regression:

```
prime-agent has an arm in the match and must not be declined
```

src/tmux/mod.rs
Eric162: You are right, and the exception is gone in `246368f6`. I could not find a way to make it sound.

Working the collision through: for title `Foo` the paired terminal is `aoe_term_Foo_<id8>`, and for title `term_Foo` the agent name is `SESSION_PREFIX + "term_Foo" + suffix`, which is the same string. So the two are equal, not merely similar. And the general form of the problem is worse than the rename case: a derived agent name is only refused by `NameShape::agent` **because** it sits under an auxiliary prefix, which is exactly the condition that makes it indistinguishable from that auxiliary kind. So "accept the derived name" and "the derived name is ambiguous" are the same predicate; there is no subset of the exception that is safe.

I checked for durable pane-kind evidence before giving up on it. The tmux hidden env carries `AOE_INSTANCE_ID` and the capture keys, nothing that records which kind of pane it is, so there is no cheap authoritative disambiguation available today.

So it fails closed, which is what `NameShape::agent`'s own doc already chose: "mistaking a paired terminal for the agent pane would be worse than not resolving at all."

Being explicit about the resulting gap, since it is a real one and it also reopens what CodeRabbit raised earlier in this PR: **a session whose title sanitizes under `term_` / `cterm_` / `tool_` gets no poller repair.** Against `main` that trades "repair happens, seeded from a name that may be the terminal" for "no repair", which is the right direction for a change whose purpose is closing that leak, but it is a gap rather than a fix. If it is worth closing, it wants a durable pane-kind marker, and that is its own change.

Test updated to pin the fail-closed behavior in both the direct and snapshot paths.

src/session/instance/tmux_session.rs
```diff
@@ -14,6 +14,24 @@ pub(super) fn tmux_env_session_name_for_instance_id(instance_id: &str) -> Option
...
 }
 
+/// [`tmux_env_session_name_for_instance_id`] restricted to the agent pane.
+/// `derived` is this instance's title-derived agent name, accepted verbatim so
+/// a title under an auxiliary prefix still resolves.
+pub(super) fn agent_tmux_session_name_for_instance_id(
+    instance_id: &str,
+    derived: Option<&str>,
+) -> Option<String> {
+    let output = crate::tmux::tmux_command()
```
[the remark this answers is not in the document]
Eric162: Fixed in `246368f6`, one word: `tmux_query_command()`.

You are right that the sibling four lines up already uses it, and `LiveSessionSnapshot::names()` does too. Matching them is the point; the `-u` and pinned `LC_MESSAGES=C` are what keep a non-ASCII session name readable regardless of an inherited `LC_CTYPE`, and `sanitize_session_name` keeps Unicode alphanumerics, so those names are real here.

Agreed on the blast radius: the derived-name fallback meant this degraded rather than broke. It should still match its siblings.

src/session/instance.rs
Eric162: Closing the loop on my question here, since events answered it: the mirrored allowlist was as fragile as it looked, and it broke `prime-agent` (your P1 on the newer revision). So no restructure, but the predicate is now a **denylist** rather than an allowlist:

```rust
"kiro" | "qwen" => false,                                   // no arm in the match
"copilot" | "kimi" | "prime-agent" => !self.is_sandboxed(),  // host store
_ => true,
```

The reasoning is the asymmetry your P1 exposed: a wrongly-eligible instance spends one throttled log line, a wrongly-ineligible one silently stops capturing its session id and loses resume. Failing open is the only defensible default for a predicate that has to track a match arm by hand, and it means a newly added agent cannot lose function through this gate.

Details and the test in `246368f6` and the P1 thread. Still happy to do the restructure (lazy exclusion-set, or the match yielding `Option<SessionIdPollFn>`) if you would rather not carry a hand-maintained list at all.

jerome-benoit (changes requested):
Re-review at exact head `246368f6fbb6458bda315ef8f3d8acc8765c86ce`, independently cross-validated through two sequential analyses and final code-level arbitration.

The previous agent-pane repair, construction eligibility, Prime Agent host/sandbox, and query-locale findings are fixed. One previously raised blocker remains valid on this head: [`probe_pane` still collapses unexpected tmux failures into `Missing`](https://github.com/agent-of-empires/agent-of-empires/pull/3558#discussion_r3887220694). It ignores exit status and stderr, so after a target has been seen alive, two EACCES, ENOTSOCK, or malformed-output probes can terminate a poller for a pane that may still exist. Please use the query-safe tmux command and classify only successful empty output and recognized no-server/ENOENT results as `Missing`; unexpected failures must remain `Unknown`. I am not duplicating the existing inline.

I also rechecked [the current title-prefix fallback finding](https://github.com/agent-of-empires/agent-of-empires/pull/3558#discussion_r3905462658) and do not consider it blocking as written. Initial startup invokes the fallback only after creating the agent pane, while repair is gated by `has_live_agent_pane_in` and a running poller is seeded from the agent-only lookup. The proposed terminal-only restart path therefore does not reach a harmful poller start.

Exact-head CI has one real failure plus its aggregate: `tmux::session::tests::size_owner_lock_claims_rejects_steals_and_releases` fails at `assert!(out.status.success())` after 6,566 passing tests. Attribution to this PR is not established, but the required gate must be rerun or fixed before merge. Format, Clippy, CodeQL, macOS, bare-core, serve-e2e, web, Nix, and coverage checks are green. `git diff --check 51ee1e81..246368f6` is clean.

src/session/instance/polling.rs
```diff
@@ -28,14 +28,48 @@ impl Instance {
...
+            crate::session::poller::warn_budget_exhausted(&self.id);
             return;
         }
         let tool = self.tool.as_str();
 
+        // Agent pane only: seeded with a paired terminal's name the poller
+        // probes that pane as alive and holds its slot for a dead agent.
         let tmux_session_name = self
-            .tmux_env_session_name()
+            .agent_tmux_session_name()
```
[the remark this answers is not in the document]
Eric162: Correct, and this is the third revision of the same hole. Fixed in `ae9980c4`.

Making the lookup fail closed was not enough because the fallback undid it: `Session::resolve_name` goes through `live_agent_session_name`, whose cache fast path is `if names.contains_key(derived) { return derived }`, so it returns a live exact derived name **before** `NameShape::agent` is applied. With `agent_tmux_session_name()` now returning `None` for a terminal-shaped name, the `.or_else` fell straight through to that name.

`poller_target_name` filters the fallback through `agent_session_belongs_to` and returns `None` when nothing agent-shaped resolves, so `maybe_start_poller_since` declines rather than seeding from a terminal.

I took your second suggestion as well, since it costs nothing here: the fallback survives for agent-shaped names, which preserves the distinction between a missing agent pane and a failed tmux lookup. A failed `list-sessions` still starts a poller on the derived name; a genuinely absent agent gets one that reads `Missing` and reaps itself; only the ambiguous collision declines.

`poller_target_declines_a_terminal_shaped_derived_name` covers the collision directly, with no tmux dependency: title `Foo` yields an agent-shaped target, title `term_Foo` yields `None`. Verified it catches the mutation (filter removed):

```
assertion failed: a terminal-shaped derived name must not become the poll target
```

Separately, on the two red checks on the previous revision: `Cargo Test (linux, serve-rest)` failed `tmux::session::tests::size_owner_lock_claims_rejects_steals_and_releases` on `assert!(out.status.success())`, and `Cargo Test (linux)` is the aggregator reporting it. That is not this PR. `main` is red on its last three runs with the same size-owner-lock family (`resize_window_if_owner_retries_same_owner_heartbeat_until_deadline`, `resize_window_if_owner_keeps_timeout_recovery_in_one_deadline`), and this diff contains no reference to size-owner code at all.

src/session/instance/polling.rs
```diff
@@ -28,16 +28,70 @@ impl Instance {
...
+    /// Keeping the fallback for agent-shaped names preserves the distinction
+    /// that matters: a failed `list-sessions` still starts a poller on the
+    /// derived name, while a genuinely absent agent gets one that reads
+    /// `Missing` and reaps itself.
+    fn poller_target_name(&self) -> Option<String> {
+        self.agent_tmux_session_name().or_else(|| {
+            self.tmux_session()
+                .ok()
+                .map(|s| s.name().to_string())
+                .filter(|name| crate::tmux::agent_session_belongs_to(name, &self.id))
```
jerome-benoit: [P2] Preserve polling for valid auxiliary-prefixed agent titles

This filter fixes the terminal collision, but it also rejects the real agent session for any otherwise valid title that sanitizes to `term_*`, `cterm_*`, or `tool_*`. For example, title `term_Foo` produces `aoe_term_Foo_<id8>`; both `agent_tmux_session_name()` and this fallback reject it, so `maybe_start_poller_since()` returns without session-ID capture. That weakens or loses resume across restart for a valid user title. Please record durable pane-kind evidence when the agent pane is created, or carry equivalent launch-time proof into startup, so confirmed agent panes keep polling while ambiguous restart repair remains fail-closed.

jerome-benoit (changes requested):
Re-review at exact head `ae9980c4fa9b0446fefbe3b3a99fac8ccaa2f0fe`, cross-validated through two independent sequential analyses and final code-level arbitration.

The new `poller_target_name()` filter correctly closes the terminal-as-agent fallback leak. Two P2 correctness blockers remain:

- The existing [`probe_pane` finding](https://github.com/agent-of-empires/agent-of-empires/pull/3558#discussion_r3887220694) is still unchanged. `src/tmux/utils.rs:268-281` ignores exit status and stderr, so an unexpected non-zero tmux failure with empty output becomes `Missing`. After the target has been seen alive, two such probes stop the poller. Please preserve the repository's existing distinction: successful absence and recognized no-server/ENOENT are `Missing`; EACCES, ENOTSOCK, malformed output, and other unexpected failures are `Unknown`. I am not duplicating that inline.
- The new filter has a separate feature-loss case, noted inline: valid agent titles sanitizing under `term_`, `cterm_`, or `tool_` now get no session-ID poller at all.

Rejected after arbitration: I am not raising the budget-warning ordering as a finding. Resolving the tmux target before the budget guard would reintroduce avoidable lookup work when the global budget is already full, and the claimed diagnostic impact is not established strongly enough to require a change.

Evidence at this SHA: `git diff --check 51ee1e81e3411f27077e2303bab8824d429374d0 ae9980c4fa9b0446fefbe3b3a99fac8ccaa2f0fe` produced no output, and `cargo fmt --all -- --check` exited successfully. GitHub currently reports only the successful `Lint PR title` check for this head; no exact-head Rust CI result is available.

Issue #3554, opened 2026-08-27
aoe 1.5.0 is *extremely* slow


anpr:
I just updated to aoe 1.5.0, and it's not usable. When I type a scroll button, it doesn't react. 
I downgraded to aoe 1.4.1 and it works again. 

Maybe the problem is that I have many sessions, especially >300 archived, >300 trash. Not sure.

Tryfciu:
same here unfortunately, newest version is not usable in current state, lag on typing/jumping between items is huge

njbrake:
Thanks all! I've got a patch release up it will get merged today.

njbrake:
Btw Could you upgrade the latest? 1.5 is pretty old

Issue #2903, opened 2026-07-16
TUI live mode: paired terminal flashes and shifts under chatty output (vite dev server)


njbrake:
> _Note: this issue was drafted by Claude via back-and-forth with @njbrake. The diagnosis and decisions are his; the prose is Claude's._

## Symptom

Viewing a session's paired bash terminal in TUI live mode while a chatty process runs in it (e.g. a vite dev server), the pane visibly flashes and the output shifts/jumps during redraws.

## Root cause

Two stacked mechanisms in the live capture pipeline, plus compounders:

1. **No repaint-quiescence gate.** The `pipe-pane` reader thread (`VtChannel::run_reader`, `src/tmux/vt.rs`) pokes the capture worker's change-wakeup after every read chunk. A full-screen redraw (clear region, then reprint) frequently spans multiple chunks, so `LiveCaptureWorker` (`src/tui/home/live_send.rs`) samples and publishes half-repainted grids. The 25ms publish floor paces frames but does not align them to logical repaint boundaries.
2. **Terminal panes forward blank frames.** `ViewMode::Terminal` sets `forward_empty = true` (`src/tui/home/render.rs`), so a blank or half-cleared intermediate frame is published; agent panes suppress those (the #1501 kill switch). This is why the paired terminal flashes where the agent pane does not.

Compounders, out of scope here but noted for context: `reconcile_size` re-seeds the parser (full reset plus `capture-pane` replay) on any pane-size change, which can re-anchor rows (the #2742/#2766 desync family); and entering live resizes the tmux window, SIGWINCHing the dev server into a full redraw.

## Fix direction

Add a short quiescence debounce to the sample path: when the reader is actively streaming, wait a few ms (roughly 5 to 10) after the last chunk before sampling, with a hard latency cap (roughly 25 to 50ms) so sustained output still renders and a stalled stream cannot delay frames indefinitely.

Constraints:

- Must not regress the keystroke echo latency work from #2822 (19ms median end-to-end on Linux). A single small chunk (echo of a keystroke) should still sample immediately or near-immediately; the debounce should only kick in when chunks are arriving back-to-back.
- Keep `forward_empty` semantics for genuine clears (user runs `clear` in the terminal and the preview should blank). If the debounce alone eliminates the mid-repaint blanks, leave `forward_empty` untouched.
- Keep the change scoped to the reader/wakeup/sample pacing; do not touch the resize or size-owner machinery.

[...]

PR #3559, opened 2026-08-27
perf(tui): stop re-forking capture-pane for an idle preview pane

Eric162:
## Description

The home view re-forked `tmux capture-pane` **on the render thread** every 250ms while previewing an idle session, to re-read bytes it had already been told were unchanged.

The preview's off-thread `LiveCaptureWorker` exists precisely so that fork stays off the render thread, but it dedups unchanged frames. An idle pane therefore hands over nothing, `apply_worker_capture` returns `false`, and the render path falls through to the synchronous fallback in `refresh_preview_cache_core` , whose 250ms idle timer then forks. The existing comment there already noted the fallback "runs on every idle refresh"; what wasn't visible is what that costs on the render thread.

Measured from the `tui.render` trace on a 429x113 terminal, home view, `live=false`:

```
frame_ms=65  capture_us=51086   <- worst observed
frame_ms=45  capture_us=33392
frame_ms=35  capture_us=21494
frame_ms=20  capture_us=0       <- same view, no fallback fork
```

`capture_us` was up to **80% of the frame**, so every ~250ms spent *dwelling* on a selection cost one to three whole 16ms budgets re-reading identical output.

**What this does not fix, to be precise about the scope:** arrowing to a different session changes `selected_session`, which trips the `cache.session_id != Some(&id)` clause , deliberately left forking. So every arrow-key press still pays a full synchronous `capture-pane`. This fixes dwelling, not moving between selections.

The worker cannot prove liveness by publishing , an idle pane publishes nothing for minutes while its loop keeps cycling , so it now bumps a cycle counter every iteration. The render path suppresses the **idle poll only** while that counter advances, and falls back to the fork if it stalls for 1.5s (a wedged or stopped worker must not freeze the preview). Cold start, session change, resize, and a scroll past the cache all still fork, since none of those are answered by "the worker had no new frame".

A 10s trust ceiling forces the fork regardless of pulse. The 250ms poll was incidentally a self-healing authoritative refresh, and a cycle counter cannot detect a worker that pulses at full rate while being *wrong* rather than wedged; the ceiling restores that self-heal at one fortieth of the old fork rate.

`Refs #3554`, but I'd now say it is **unlikely to be that reporter's regression**: an archived or trashed selection never reaches the capture path at all (`render_preview` returns into the archived/trashed placeholder first), their 600 rows are mostly those, and their bisect window (1.14.1 → 1.15.0) is far newer than this 250ms poll. Keep #3554 open.

### Remaining, not addressed here

The same trace shows a ~20ms floor with `capture_us=0` , widget build plus ratatui diff at 429x113 (48k cells). That caps the view near 45fps and wants its own investigation.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No screenshot: the rendered output is byte-identical, only its cost changes. Verified by hand against a real store (23 live sessions, 61 stored) with the `tui.render` trace on: `capture_us` drops to ~0 on home frames and frame times settle at the ~20ms floor.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: what changed is whether a fork happens, and asserting "no subprocess was spawned this frame" from the e2e harness is not something it can observe. The decision itself is extracted into the pure `worker_pulse_step` helper (same pattern as the existing `preview_frozen` / `passive_resize_step` helpers in this file) and tabled by `worker_pulse_step_trusts_a_running_worker_and_gives_up_on_a_wedged_one`, which covers cold start, an advancing counter, an unmoved-but-recent counter (and that the observation is *not* re-stamped, or a wedged worker would never be caught), and the timeout. The rendering paths it gates are already covered by the existing `tui::home::render` suite (64 tests, passing unchanged).

## AI Usage

- [ ] No AI was used
- [x] AI was used

Found by enabling the existing `tui.render` trace (`[logging.targets] "tui.render" = "trace"`) against a real store and reading the `capture_us` breakdown on over-budget frames. Worth noting the trace paid for itself here: the frame accounting was already in the code, just filtered out at the default `info` level.

## Summary

- Prevents repeated idle `tmux capture-pane` calls on the TUI render thread.
- Tracks `LiveCaptureWorker` progress with a cycle counter.
- Uses synchronous capture when the worker stalls, changes target, or requires an immediate refresh.
- Resets worker progress tracking after retargeting.
- Adds tests for worker progress and idle capture decisions.

## Benefits

- Reduces render-thread work during idle previews.
- Keeps preview output unchanged.
- Preserves reliable updates for cold starts, session changes, resizes, scrolling, stalled workers, and forced refreshes.

## Technical details

`HomeView` records the latest worker pulse. Idle capture stays suppressed while the worker progresses. A 1.5-second timeout detects stalled workers. A 10-second trust ceiling provides a self-healing fallback. Synchronous capture remains enabled for required refreshes.

### DIFF
--- a/src/tui/home/live_send.rs
+++ b/src/tui/home/live_send.rs
@@ -806,6 +806,13 @@ pub(in crate::tui) struct LiveCaptureWorker {
     /// capture-pane. Mirrors the Clipboard Pass-through setting, so disabled
     /// mode does not keep a second pipe-pane connection open.
     clipboard_capture_enabled: std::sync::Arc<std::sync::atomic::AtomicBool>,
+    /// Cycle counter, bumped at the top of every worker iteration. The render
+    /// thread reads it to tell "the worker is running and has nothing new"
+    /// (an idle pane: no fork needed) from "the worker is wedged" (fall back
+    /// to the synchronous fork). Publishes cannot answer that on their own:
+    /// the worker dedups unchanged frames, so an idle pane publishes nothing
+    /// for minutes while still cycling.
+    cycles: std::sync::Arc<std::sync::atomic::AtomicU64>,
 }
 
 impl Drop for LiveCaptureWorker {
@@ -1000,6 +1007,8 @@ impl LiveCaptureWorker {
         // Pass-through setting, avoiding an observer before configuration is
         // available for a newly created worker.
         let clipboard_capture_enabled = Arc::new(AtomicBool::new(false));
+        let cycles = Arc::new(AtomicU64::new(0));
+        let cycles_cell = cycles.clone();
         let lines_cell = capture_lines.clone();
         let target_cell = target.clone();
         let slot = latest.clone();
@@ -1070,6 +1079,7 @@ impl LiveCaptureWorker {
                 crate::tmux::composite::WindowLayout,
             )> = None;
             while !stop_flag.load(Ordering::Relaxed) {
+                cycles_cell.fetch_add(1, Ordering::Relaxed);
                 let lines = lines_cell.load(Ordering::Relaxed);
                 // Read the target without holding the lock across the fork:
                 // `set_target` must never wait on a `capture-pane`.
@@ -1407,6 +1417,7 @@ impl LiveCaptureWorker {
             clipboard,
             vt_enabled,
             clipboard_capture_enabled,
+            cycles,
         }
     }
 
@@ -1534,6 +1545,15 @@ impl LiveCaptureWorker {
     /// Take the newest capture the worker has produced since the last
     /// call, if any. Returns `None` when nothing new has arrived (the
     /// render loop then keeps the current preview).
+    /// Snapshot of the worker's cycle counter. Two reads a moment apart that
+    /// differ mean the capture loop is running; identical reads across more
+    /// than a cycle's worth of time mean it is wedged (or stopped). The render
+    /// thread uses this to decide whether "no new frame" can be trusted as
+    /// "nothing changed".
+    pub(in crate::tui) fn cycles(&self) -> u64 {
+        self.cycles.load(std::sync::atomic::Ordering::Relaxed)
+    }
+
     pub(in crate::tui) fn take_latest(&self) -> Option<String> {
         self.latest.lock().ok().and_then(|mut guard| guard.take())
     }
--- a/src/tui/home/live_send.rs
+++ b/src/tui/home/live_send.rs
@@ -1545,6 +1545,10 @@ impl LiveCaptureWorker {
     /// Take the newest capture the worker has produced since the last
     /// call, if any. Returns `None` when nothing new has arrived (the
     /// render loop then keeps the current preview).
+    pub(in crate::tui) fn take_latest(&self) -> Option<String> {
+        self.latest.lock().ok().and_then(|mut guard| guard.take())
+    }
+
     /// Snapshot of the worker's cycle counter. Two reads a moment apart that
     /// differ mean the capture loop is running; identical reads across more
     /// than a cycle's worth of time mean it is wedged (or stopped). The render
@@ -1552,10 +1556,6 @@ impl LiveCaptureWorker {
     /// "nothing changed".
     pub(in crate::tui) fn cycles(&self) -> u64 {
         self.cycles.load(std::sync::atomic::Ordering::Relaxed)
-    }
-
-    pub(in crate::tui) fn take_latest(&self) -> Option<String> {
-        self.latest.lock().ok().and_then(|mut guard| guard.take())
     }
 
     /// The newest cursor for the worker's CURRENT target pane, or `None`
--- a/src/tui/home/mod.rs
+++ b/src/tui/home/mod.rs
@@ -668,6 +668,17 @@ pub struct HomeView {
     /// at, so the reconcile can tell when the displayed pane changed and
     /// retarget. `None` before the first preview or when nothing is selected.
     pub(super) preview_capture_target: Option<String>,
+    /// Last observed `LiveCaptureWorker::cycles` value and when it was seen,
+    /// so the render thread can tell a running worker with nothing new (an
+    /// idle pane) from a wedged one. `None` before the first observation and
+    /// after every retarget.
+    ///
+    /// This gates the synchronous fallback fork. Without it the 250ms idle
+    /// poll re-forked `capture-pane` on the render thread to re-read content
+    /// the worker had already reported unchanged: measured at 10-50ms per
+    /// fork against a 429x113 terminal, one to three whole frame budgets
+    /// burned every 250ms of simply looking at an idle session.
+    pub(super) preview_worker_pulse: Option<(u64, std::time::Instant)>,
     /// Notified by the capture worker thread when it has fresh, changed
     /// content. The event loop selects on this to repaint without
     /// busy-polling; an idle pane (no new content) never wakes it.
@@ -2344,6 +2355,7 @@ impl HomeView {
             live_send_worker: None,
             preview_capture_worker: None,
             preview_capture_target: None,
+            preview_worker_pulse: None,
             preview_wake: std::sync::Arc::new(tokio::sync::Notify::new()),
             live_send_last_resize: None,
             live_send_pending_leader: false,
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -2300,8 +2300,9 @@ impl HomeView {
         // `sync_preview_capture_worker` in `render_preview`) keeps fresh
         // content flowing on its own thread; `apply_worker_capture` below
         // just applies the newest it has produced. The synchronous fork via
-        // `refresh_preview_cache_core` remains only as the cold-start /
-        // worker-empty fallback (its 250ms gate still applies there). This
+        // `refresh_preview_cache_core` remains as the cold-start fallback,
+        // plus the wedged-worker and trust-ceiling paths in `idle_poll_due`;
+        // a worker that is merely idle no longer trips it. This
         // moves the per-frame capture cost (~8.5ms on macOS, ~90% of a
         // frame; the `tui.render` `capture_us` trace measures it) off the
         // render thread for every view, not just agent live-send.
@@ -2426,8 +2427,8 @@ impl HomeView {
                 // "session looks gone" signal the non-live path uses).
                 let same_session = s.preview_cache.session_id.as_deref() == Some(id);
                 // Composited in BOTH modes, matching what the worker publishes.
-                // This fallback also runs on every idle refresh (the worker
-                // dedups unchanged frames, so it has nothing to hand over), so a
+                // This fallback still runs behind a live worker (the trust
+                // ceiling, and any stall past `WORKER_PULSE_TIMEOUT`), so a
                 // pane-0-only capture here would clobber the worker's composite
                 // a beat after each keystroke and leave the split visible for
                 // only one frame at a time. On an unsplit window this returns
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -181,6 +181,46 @@ fn capture_lines_for(height: u16, scroll_offset: u16) -> usize {
 /// worker's bottom-anchored live frames would shift the content out from under
 /// the user: the read position gets yanked toward the tail as output streams,
 /// or the drag anchors slide off their text. Frozen, wheel-scroll stays smooth
+/// and a drag-select tracks exactly the cells under the pointer.
+fn preview_frozen(scroll_offset: u16, has_selection: bool) -> bool {
+    scroll_offset > 0 || has_selection
+}
+
+/// Idle-poll cadence for the preview cache, in ms. The synchronous
+/// `capture-pane` fallback fires no more often than this.
+const PREVIEW_REFRESH_MS: u128 = 250;
+
+/// Ceiling on how long the capture worker may be the sole source of truth.
+///
+/// The 250ms poll was incidentally a self-healing authoritative refresh: it
+/// overwrote the cache with real `capture-pane` output four times a second, so
+/// any divergence between the worker's picture of the pane and tmux's was
+/// repaired before anyone saw it. Suppressing it removes that net for the one
+/// failure a cycle counter cannot see, a worker that cycles at full rate while
+/// being confidently wrong (its VT grid stably diverged, so `last_captured`
+/// matches every cycle and nothing publishes). This restores the self-heal at
+/// one fortieth of the old fork rate, which keeps essentially all of the win.
+const WORKER_TRUST_CEILING_MS: u128 = 10_000;
+
+/// Whether the idle poll should pay for a synchronous `capture-pane`.
+///
+/// Three ways it does not:
+/// - inside the cadence, which is the pre-existing throttle;
+/// - while frozen (reading scrollback or holding a selection), because a fresh
+///   bottom-anchored snapshot would shift the held content out from under the
+///   reader or the drag. Session change, resize, and the reading grow refresh
+///   through their own clauses regardless;
+/// - while the capture worker is pulsing, because it publishes every change,
+///   so a fork here could only re-read identical bytes at 10-50ms on the
+///   render thread. That suppression expires at [`WORKER_TRUST_CEILING_MS`],
+///   so a worker that is wrong rather than wedged still gets corrected.
+fn idle_poll_due(frozen: bool, idle_elapsed_ms: u128, worker_covers_idle: bool) -> bool {
+    if frozen || idle_elapsed_ms <= PREVIEW_REFRESH_MS {
+        return false;
+    }
+    !worker_covers_idle || idle_elapsed_ms > WORKER_TRUST_CEILING_MS
+}
+
 /// How an observed capture-worker cycle counter folds into a liveness verdict,
 /// given the previous observation. Returns `(is_pulsing, new_observation)`.
 ///
@@ -213,11 +253,6 @@ fn worker_pulse_step(
 /// capture fork inside a cycle can add tens of ms, so one slow cycle must not
 /// trip this.
 const WORKER_PULSE_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(1500);
-
-/// and a drag-select tracks exactly the cells under the pointer.
-fn preview_frozen(scroll_offset: u16, has_selection: bool) -> bool {
-    scroll_offset > 0 || has_selection
-}
 
 /// Decide whether the cached capture window still covers the requested scroll.
 /// Returns true when the cache must be re-captured because the visible window
@@ -2001,9 +2036,12 @@ impl HomeView {
     /// (#1501); the terminal/tool wrappers use it when the instance has gone
     /// away, matching the original "only write inside `if let Some(inst)`".
     ///
-    /// `force` bypasses the idle throttle; the agent passes `in_live` here so
-    /// every render refreshes the preview in live-send mode (see the throttle
-    /// note in `refresh_preview_cache_if_needed`), the others pass `false`.
+    /// `force` bypasses the idle throttle. All four wrappers pass `false`
+    /// today, so it is inert defensive code; the doc used to claim the agent
+    /// passed `in_live` here to refresh every render in live-send mode, which
+    /// has not been true at any call site. Live-send does not need it: on the
+    /// fast cadence `apply_worker_capture` has a fresh frame nearly every
+    /// render and returns before this is reached.
     fn refresh_preview_cache_core(
         &mut self,
         width: u16,
@@ -2012,35 +2050,35 @@ impl HomeView {
         select: fn(&mut Self) -> &mut super::PreviewCache,
         capture: impl FnOnce(&Self, &str, usize) -> Option<String>,
     ) {
-        const PREVIEW_REFRESH_MS: u128 = 250;
         let Some(id) = self.selected_session.clone() else {
             return;
         };
         let scroll_offset = self.preview_scroll_offset;
         let frozen = self.preview_is_frozen();
         let visible_rows = self.preview_visible_rows;
-        // Only the idle timer is suppressed by a healthy worker; a live-send
-        // `force`, a session change, a resize, and a scroll that outgrew the
-        // cache all still fork, because none of those are answered by "the
-        // worker had no new frame".
-        let worker_covers_idle = !force && self.worker_is_pulsing();
+        // Only the idle timer is suppressed by a pulsing worker. A session
+        // change, a resize, and a scroll that outgrew the cache all still
+        // fork, because none of those are answered by "the worker had no new
+        // frame". `force` is inert defensive code: every caller passes
+        // `false`, so live-send is covered by the suppression rather than
+        // exempt from it, and it does not need an exemption because on the
+        // fast cadence `apply_worker_capture` almost always has a fresh frame
+        // to hand over and returns before this.
+        let worker_covers_idle = !force && self.observe_worker_pulse();
 
         let cache = select(self);
+        let idle_elapsed = cache.last_refresh.elapsed().as_millis();
         let needs_refresh = force
             || cache.session_id.as_ref() != Some(&id)
             || cache.dimensions != (width, height)
-            || capture_window_stale(frozen, cache.captured_lines, height, visible_rows, scroll_offset)
-            // While frozen (reading scrollback or holding a selection) the idle
-            // poll must not re-capture: a fresh bottom-anchored snapshot would
-            // shift the held content out from under the reader or the drag.
-            // Session change, resize, and the reading grow still refresh.
-            //
-            // A pulsing worker also suppresses it: the worker publishes every
-            // change, so re-forking `capture-pane` here could only re-read
-            // identical bytes, at 10-50ms on the render thread.
-            || (!frozen
-                && !worker_covers_idle
-                && cache.last_refresh.elapsed().as_millis() > PREVIEW_REFRESH_MS);
+            || capture_window_stale(
+                frozen,
+                cache.captured_lines,
+                height,
+                visible_rows,
+                scroll_offset,
+            )
+            || idle_poll_due(frozen, idle_elapsed, worker_covers_idle);
         if !needs_refresh {
             return;
         }
@@ -2164,16 +2202,18 @@ impl HomeView {
         preview_frozen(self.preview_scroll_offset, self.preview_selection.is_some())
     }
 
-    /// Whether the capture worker is demonstrably running for the pane on
+    /// Fold this frame's observation of the capture worker's cycle counter in,
+    /// and report whether the worker is demonstrably running for the pane on
     /// screen, so "the worker handed over nothing" can be read as "nothing
-    /// changed" rather than "nobody is capturing".
+    /// changed" rather than "nobody is capturing". Takes `&mut self` because
+    /// the observation it compares against is the state it updates.
     ///
     /// Liveness comes from the worker's cycle counter, not from its publishes:
     /// it dedups unchanged frames, so an idle pane publishes nothing for
     /// minutes while the loop keeps running. Decision lives in the pure
     /// [`worker_pulse_step`] helper so it is unit-tested away from a live
     /// worker thread.
-    fn worker_is_pulsing(&mut self) -> bool {
+    pub(super) fn observe_worker_pulse(&mut self) -> bool {
         let Some(worker) = self.preview_capture_worker.as_ref() else {
             self.preview_worker_pulse = None;
             return false;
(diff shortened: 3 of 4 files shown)


njbrake (commented):

_Reviewed at `c4f3fc83`. Push new commits and it will be re-reviewed._

---

### Summary

The home preview's off-thread `LiveCaptureWorker` dedups unchanged frames, so an idle pane publishes nothing, `apply_worker_capture` returns `false`, and the render path falls through to `refresh_preview_cache_core`, whose 250ms idle timer forks `tmux capture-pane` on the render thread to re-read bytes already known to be identical. This PR gives the worker an `AtomicU64` cycle counter bumped at the top of every loop iteration, and suppresses only that idle-timer clause while the counter is observed advancing, with a 1.5s stall fallback.

Verified locally on `c4f3fc83`, already on top of `origin/main` (`522a7d2e`), no rebase needed. `cargo fmt --check` clean, `cargo test --lib` 4739 passed / 0 failed, `cargo clippy --all-targets` shows only the pre-existing `macos.rs` `items_after_test_module` warning and nothing in the three changed files.

### Design Alignment

**Aligned.** The change extends the existing mechanism rather than adding one beside it. Concretely:

- The suppression is a fourth term in the *same* `needs_refresh` disjunction in `refresh_preview_cache_core`, next to `capture_window_stale` and the `!frozen` guard, not a new gate wrapped around the call.
- `worker_pulse_step` is a free function taking its inputs and returning `(verdict, new_state)`, tested by table, sitting immediately beside `preview_frozen`, `passive_resize_step`, `capture_window_stale`, and `capture_is_exhausted` in the same file. That is precisely the local idiom for "pull the decision out of `&mut self` so it can be unit-tested," and `worker_is_pulsing` mirrors `preview_is_frozen`'s one-line delegation, comment and all.
- The counter is an `Arc<AtomicU64>` cloned into the worker closure and stored on the struct, matching `capture_lines` (`AtomicUsize`), `interval_ms` (`AtomicU64`), `stop`/`forward_empty`/`vt_enabled` (`AtomicBool`) exactly.
- Retarget invalidation goes in `sync_preview_capture_worker`'s `if self.preview_capture_target != desired` block, alongside the existing `hover_forward_cell = None` reset. Same block, same reason, same shape.

No new abstraction, no parallel path, no layering change.

### Verification of the body's claims

| Claim | Verdict |
| --- | --- |
| Worker dedups unchanged frames so an idle pane publishes nothing | **Confirmed.** `if (forward_empty \|\| !content.is_empty()) && changed` in the loop; `changed` compares against `last_captured`. |
| Only the idle clause is suppressed | **Confirmed.** Read the full `needs_refresh`; session change, dimension change, and `capture_window_stale` are untouched, and `!frozen` still gates the idle clause independently. |
| Cold start still forks | **Confirmed.** `worker_pulse_step(_, None, _)` returns `(false, Some(..))`. |
| Session change still forks | **Confirmed**, by two independent mechanisms: `cache.session_id.as_ref() != Some(&id)`, and `preview_worker_pulse = None` on retarget. |
| Resize still forks | **Confirmed.** `cache.dimensions != (width, height)`. |
| Scroll past the cache still forks | **Confirmed.** `capture_window_stale(frozen, ...)` sits ahead of the suppressed clause. Also checked the `apply_worker_capture` undersized-capture path that deliberately returns `false` after `store_capture`: it lands on `capture_window_stale`, not the idle clause, so the catch-up still runs. |
| "the live-send `force` all still fork" | **Wrong, though harmlessly so.** All four wrappers pass `false` for `force`, on `main` and in this PR. No caller has ever set it. So `!force &&` is inert defensive code and live-send is *not* exempted from the suppression; it is covered by it. Not a bug, because `apply_worker_capture` returns `true` whenever the worker has anything fresh, and during live-send on the fast cadence it almost always does. But the body claims an exemption that does not exist. (`refresh_preview_cache_core`'s own doc comment, "the agent passes `in_live` here", is stale on `main` for the same reason. Not this PR's fault, worth fixing while nearby.) |
| Rendered output is byte-identical | **True as far as the worker's frames are correct.** See Merge Risk. |

### Memory ordering, wraparound, clock

All three check out.

- `fetch_add(1, Relaxed)` on the worker, `load(Relaxed)` on the render thread. Relaxed is *sufficient and correct* here: the counter is a pure heartbeat that orders no other memory. Nothing is being released alongside it. The frames it gates on travel through `latest: Arc<Mutex<Option<String>>>`, which carries its own acquire/release. A stronger ordering would buy nothing and cost a fence on every worker iteration. Single-location relaxed atomics are modification-order coherent, so the render thread cannot observe the counter going backwards, and the worst case is observing a value one or two cycles behind, which lands in the "unmoved but recent" branch and is already handled.
- Wraparound: `u64`, and the comparison is `!=`, not `<`. At the fast cadence the loop is nowhere near a rate that matters; and even at the wrap the `!=` still reads as "moved."
- `std::time::Instant` is monotonic, and `saturating_duration_since` guards a `now` earlier than `at`. Correct.

### The 1.5s stall fallback

It does what it says for a *stopped* loop, and the code is honest that this is what it detects. Two properties worth stating explicitly:

**It cannot be defeated by repeated observation.** This is the subtle part and the author got it right: the unmoved branch returns `Some((seen, at))`, preserving the *original* timestamp rather than re-stamping. Had it re-stamped, every render would push the deadline out by a frame and a wedged worker would never be caught. The test asserts this specifically, with a comment saying why. Mutating `Some((seen, at))` to `Some((seen, now))` fails the test.

**It cannot detect a worker that cycles fast but never publishes.** The author does not claim otherwise (the doc says "wedged"), but this is the load-bearing assumption of the whole change and deserves to be named: `cycles_cell.fetch_add` sits at the *very top* of the loop, before the target read, before the `lines > 0 && !name.is_empty()` guard, before any capture. A worker whose capture path fails on every iteration, or whose VT grid has stably diverged from tmux truth, pulses at full rate and suppresses the fallback forever. See Merge Risk.

**What the user sees during a real wedge:** the last-good frame, held. Worst-case staleness for a wedged worker goes from ~250ms (today) to ~1.75s (one observation frame plus the timeout). Bounded, and reasonable for the cost saved.

### Stale-preview walk

I went looking for a pane that stays on screen while the worker pulses and publishes nothing. Every one I found is already short-circuited *ahead of* the cache, by a check that predates this PR:

- **Archived / trashed / stopped (`TMUX_SESSION_GONE_ERROR`) / structured-ACP**, Structured view: `render_preview` computes `desired = None`, calls `sync_preview_capture_worker(None)` (which retargets and therefore resets the pulse), then `return`s into a placeholder before `refresh_preview_cache_if_needed` is ever reached.
- **Terminal / container view, session gone:** `terminal_running` comes from `session_exists_for_display`, independent of the cache, and renders "Press Enter to start terminal". Also, Terminal panes run `forward_empty = true`, so the worker publishes empties and clears stale text on its own.
- **Tool view, session gone:** same, via `tool_running` = `session_exists_for_display && !pane_dead_for_display`.

The one residual transient: between a pane dying and the poller setting `last_error`, an agent preview now holds its last frame where `main` would have written `Some("")` within 250ms and flashed "No output available" before the Stopped placeholder took over. That is arguably an improvement, and it is bounded by one poller interval. Not a finding.

I also checked that the worker is never replaced (both assignment sites are `is_none()`-guarded), so `cycles` never restarts at 0 under a live observation; and that `cache.last_refresh` has no consumer outside the gate it feeds, so freezing it in the idle case is inert.

### Issues Found

**1. Two doc comments were spliced into, hijacking documentation off existing functions.** This is the only thing I would call a defect in the diff, and it is mechanical.

`src/tui/home/render.rs`: `worker_pulse_step` and `WORKER_PULSE_TIMEOUT` were inserted **into the middle of `preview_frozen`'s doc comment**. Six lines of `preview_frozen`'s prose now document `worker_pulse_step`, and `preview_frozen` is left with the orphaned fragment:

```rust
/// and a drag-select tracks exactly the cells under the pointer.
fn preview_frozen(scroll_offset: u16, has_selection: bool) -> bool {
```

`src/tui/home/live_send.rs`: identical splice. `take_latest`'s three-line doc now sits above `cycles()`, and `take_latest` is left undocumented:

```rust
    /// Take the newest capture the worker has produced since the last
    /// call, if any. Returns `None` when nothing new has arrived (the
    /// render loop then keeps the current preview).
    /// Snapshot of the worker's cycle counter. ...
    pub(in crate::tui) fn cycles(&self) -> u64 {
```

Fix is to move each new item *below* the function whose doc it split: `worker_pulse_step` plus `WORKER_PULSE_TIMEOUT` after `preview_frozen`'s body, and `cycles()` after `take_latest()`'s body. Nothing else changes. (Note this exact pattern already exists on `main` between `apply_worker_capture` and `preview_is_frozen`; that one is pre-existing and not attributable here, but it does suggest the file is prone to it.)

**2. The body's "navigating the list" framing overclaims what this fixes.** The body says "Navigating the list means landing on those frames continuously, which is what the slowness feels like." Arrowing to a new session changes `selected_session`, which trips `cache.session_id.as_ref() != Some(&id)`, a clause this PR deliberately and correctly leaves forking. So **every arrow-key press still pays a full synchronous `capture-pane` on the render thread.** What this PR fixes is *dwelling* on a selection, not moving between selections. That is still a real and worthwhile fix, and the 51ms measurement is real, but the two cases should not be conflated, especially since the second one is what the #3554 reporter describes.

**3. Minor: `worker_is_pulsing` is an `is_*` query with a `&mut self` side effect** (it writes `preview_worker_pulse`). Correct as written, and unavoidable without threading state through the caller, but the name reads as pure. Not blocking.

### Test Coverage

The one new test is real and load-bearing. I mutation-tested it:

Dropping the timeout (`now.saturating_duration_since(at) < WORKER_PULSE_TIMEOUT` to `true`):

```
test tui::home::render::tests::worker_pulse_step_trusts_a_running_worker_and_gives_up_on_a_wedged_one ... FAILED
assertion `left == right` failed
  left: (true, Some((7, Instant { tv_sec: 3729162, tv_nsec: 218612041 })))
 right: (false, Some((7, Instant { tv_sec: 3729162, tv_nsec: 218612041 })))
```

Re-stamping the observation (`Some((seen, at))` to `Some((seen, now))`):

```
test tui::home::render::tests::worker_pulse_step_trusts_a_running_worker_and_gives_up_on_a_wedged_one ... FAILED
assertion `left == right` failed
  left: (true, Some((7, Instant { tv_sec: 3729187, tv_nsec: 473822416 })))
 right: (true, Some((7, Instant { tv_sec: 3729185, tv_nsec: 974822416 })))
```

Both caught. Good test, correct house style: one test fn, four cases in sequence, each with a comment saying what it protects, no assertion count lost to consolidation.

**But the wiring is untested.** Two more mutations, both against the full 808-test `tui::home` suite:

- Removing `&& !worker_covers_idle` from the `needs_refresh` gate, that is, reverting the entire perf fix, leaves **808 passed; 0 failed**.
- Removing `self.preview_worker_pulse = None;` from the retarget block, the guard that stops a retargeted worker's pulse from suppressing the new pane's cold-start fork, also leaves **808 passed; 0 failed**.

So nothing fails if someone later deletes the feature or the retarget reset. The author's coverage note is honest about the first (the helper is what is tested), but the retarget reset is the one place where a missed path really would produce a stale preview, and it is the cheapest thing in the diff to pin.

**On whether an e2e test is owed: no.** The author's reasoning is right and I agree with it. "No subprocess was spawned this frame" is not observable from `TuiTestHarness`, a Rust e2e is ~2s and strictly serial, and the rendered output is unchanged by design so there is nothing for a screen assertion to catch. Applying the cost table, this does not earn an e2e slot.

**What I would add instead is one Rust unit test, not an e2e**, exercising `HomeView::worker_is_pulsing` against a real spawned worker (`live_send.rs` already spawns real `LiveCaptureWorker`s in its own tests, e.g. around the `set_capture_lines` cases): assert `false` on the first call, `true` once the counter has moved, and `false` again immediately after `sync_preview_capture_worker` retargets. That is ~15 lines, ~20ms, in the tier the house rules prefer, and it kills both surviving mutants.

### Documentation

Proportionate in volume and genuinely good in content: the comments explain *why* (why publishes cannot prove liveness, why the timestamp must not be re-stamped, why the timeout is 1500ms against a 250ms cadence) rather than restating the code. No new markdown files, no narration of the change in the tree. The only documentation problem is the splice damage in Issue 1, which is a placement bug, not excess prose.

No hand-maintained mirror is made stale: `cycles()` is `pub(in crate::tui)` with one caller, `preview_worker_pulse` is `pub(super)`, nothing is generated from these, and `docs/cli/reference.md` is unaffected.

### DESIGN.md

Read it. Nothing visual changes. No color, typography, spacing, layout, icon, or theme surface is touched; the diff adds one atomic, one `Option<(u64, Instant)>` field, and one boolean term in a refresh predicate. The "no screenshot, output is byte-identical" claim is correct as a design matter. The only way pixels differ is the staleness case in Merge Risk, which is a correctness question rather than a design one.

### On #3554

`Refs #3554`, not `Closes`, and I would go slightly further than the body's hedge. The reporter says "when I type a scroll button, it doesn't react" and volunteers "300+ archived, 300+ trash" as the likely cause. Two things point away from this PR:

1. An archived or trashed selection never reaches the preview capture path at all. `render_preview` computes `desired = None` and returns into `render_archived_preview` / `render_trashed_preview` before `refresh_preview_cache_if_needed` runs. If the reporter is scrolling through those 600 rows, this change is a no-op for them.
2. Even on live rows, scrolling is the `session_id` clause, which still forks every keypress (Issue 2 above).

The reporter's own bisect is 1.4.1 to 1.5.0 (almost certainly 1.14.1 to 1.15.0 given `Cargo.toml` is at 1.15.0), which is a narrow window worth bisecting directly. The 250ms idle poll being fixed here is much older than that window, so it is unlikely to be the regression they hit. The hedge in the body is appropriately honest; I would just make it a bit stronger and keep #3554 open.

### On the deferred ~20ms floor

Yes, worth its own issue. `capture_us=0` with `frame_ms=20` at 429x113 (48k cells) is widget construction plus the ratatui diff, capping the view near 45fps, and it is a different subsystem from anything touched here. Filing it separately is the right call rather than growing this PR. Given point 1 above, a combined issue covering "the list itself is expensive at 600+ rows" would probably be the more valuable one for #3554.

### CI Workflow Approval

Not acted on. Two runs (`Tests` id 33078892546, `CI` id 33078892576) sit at `action_required` for `c4f3fc83`. For the record, the exposure audit passes cleanly: three `.rs` files, no workflow or CI config, no install/build/test hook, no dependency or lockfile change, no binary or obfuscated content, nothing unrelated smuggled in.

### Merge Risk

**Low to moderate.** Blast radius is one preview pane; reversibility is a clean `git revert` with no migration, artifact, or config residue.

The residual risk is structural and worth naming rather than filing under suggestions: **the 250ms idle fork was, incidentally, a self-healing authoritative poll.** Every 250ms it overwrote the preview cache with real `capture-pane` output, so any divergence between the worker's picture of the pane and tmux's truth was silently repaired within a quarter second and nobody ever saw it. After this PR, in steady state the worker's in-process vt100 grid is the sole source of truth for as long as it keeps cycling.

That matters because the grid is fed by a `pipe-pane -IO` channel with a seed handoff, and this file's own history is full of the ways that goes wrong: #2903's mid-repaint publishes, #2822's echo path, the PROMPT_EOL_MARK false frame that `sync_preview_capture_worker` disables VT for shell terminals to avoid. Those were all *caught* bugs; the concern is the uncaught kind, where the grid stably diverges, `last_captured` matches on every cycle, nothing publishes, the counter pulses at full rate, and the fallback never fires. The 1.5s timeout does not cover this, because the worker is not wedged, it is confidently wrong. CodeRabbit flagged the same shape ("a worker that keeps running while capture fails could leave the displayed pane state stale indefinitely"), and I agree with it; worker liveness and capture success are genuinely different signals and only the first is being measured.

I could not construct a concrete reproduction, and I looked: every "pane is gone" case turns out to be short-circuited ahead of the cache by an existence check. So this is a removed safety net rather than a demonstrated bug, and I would not block on it.

**Most likely way it bites us:** a user reports "the preview is showing something stale and pressing a key fixes it," on a setup where the VT channel has a problem we do not currently know about. Nothing in this PR would surface it: no assertion, no log line, no test. We would hear it from a user, and it would be hard to reproduce because any interaction that changes selection, size, or scroll repairs it instantly.

**The cheap insurance, if you want it:** a much longer floor on the suppression, forcing a fork if `cache.last_refresh.elapsed()` exceeds, say, 10 seconds, regardless of pulse. That restores the self-heal at one fortieth of the current fork rate, keeps essentially all of the measured win, and costs about four lines. Worth offering to the author, not worth blocking on.

Coverage gap to name specifically: nothing exercises `refresh_preview_cache_core`'s new `!worker_covers_idle` term or the retarget `preview_worker_pulse = None` reset. Both survive deletion with a green suite.

### Verdict

**Design Alignment:** Aligned
**Implementation:** Approve, with the doc-comment splices fixed before merge
**Project Fit:** Accept
**Documentation:** Trim not needed; two placement fixes required (see Issue 1)
**CI Workflow Approval:** Not acted on; audit passes for `c4f3fc83`
**Merge Risk:** Low to moderate; the idle poll was an incidental self-healing authoritative refresh, and nothing replaces it for a worker that pulses while being wrong

This is a well-built change: it found a real 51ms-per-250ms cost on the render thread with the project's own existing trace, fixed it by extending the mechanism that was already there rather than adding one beside it, got the two hard details right (relaxed ordering is the correct choice for a heartbeat, and the observation timestamp is deliberately not re-stamped so the timeout cannot be pushed out by repeated renders), and tested the decision at the cheapest tier that can actually fail. The author did what was asked. I would merge it.

Three things to settle first, in descending order: fix the two spliced doc comments, which are the only defect in the diff; soften the body's "navigating the list" claim and the "live-send `force` still forks" claim, since `force` is never `true` at any call site and list navigation still forks on the untouched `session_id` clause; and consider the ~15-line `worker_is_pulsing` unit test, because the retarget reset is the one guard whose deletion produces a genuinely stale preview and it currently survives deletion with 808 tests green.

njbrake (commented):

_Reviewed at `796b936c`. Push new commits and it will be re-reviewed._

---

Re-reviewed at head `796b936c`, one commit on top of the previously reviewed `c4f3fc83`. Rebased onto current `origin/main` (`fdb46c3f`); the rebase applied cleanly with no conflicts and I judged the rebased tree.

Short version: every finding from the last round is fixed, the fixes are correct, and the new commit goes further than what I asked for. Both mutants that previously survived a green suite now die. I recommend merging.

## Verification environment

`cargo fmt --check` clean. `cargo clippy --all-targets` produces exactly one warning, the pre-existing `items_after_test_module` in `src/process/macos.rs`, which is on the known-unclean list and is untouched by this diff; zero warnings reference any of the four changed files. `cargo test --lib` is 4742 passed, 0 failed, 2 ignored.

## Disposition of the previous findings

**1. Spliced doc comments. Fixed, and verified in the post-rebase file rather than taken from the commit headline.**

In `src/tui/home/render.rs`, `preview_frozen` now carries its complete original doc comment, ending on "and a drag-select tracks exactly the cells under the pointer", with the function body directly under it. `worker_pulse_step` and `WORKER_PULSE_TIMEOUT` were moved below it and now carry only their own prose. In `src/tui/home/live_send.rs`, `take_latest` has its three-line doc back and `cycles()` sits after it with its own. The commit message claims "the diff against main is now additions only in `live_send.rs`", and that is literally true: the diff of that file against main has zero deleted lines. Both restored comments describe the functions they sit on. Closed.

**2. The `force` claim. Fixed, in both the body and the code, and fixed more honestly than I expected.**

The body's enumeration no longer lists live-send `force` as a path that still forks; it now reads "Cold start, session change, resize, and a scroll past the cache all still fork". The stale doc on `refresh_preview_cache_core` that claimed "the agent passes `in_live` here" has been replaced with a statement of what is actually true: all four wrappers pass `false`, the parameter is inert defensive code, and live-send does not need an exemption because on the fast cadence `apply_worker_capture` almost always has a fresh frame and returns before this point is reached. The inline comment on `worker_covers_idle` says the same. That is the right correction: rather than quietly dropping the claim, the author documented the actual state of a parameter that has been misdescribed for some time.

One residual, and I am calling it a nit rather than a blocker. AGENTS.md says do not write things nothing reads, and `force` is a parameter no caller has ever set to `true`. The compiler cannot see that, so nothing flags it, but the honest resolution under the house rule is deletion: the signature, four `false` arguments, and two clauses, well under ten lines. It predates this PR, so removing it here is arguably scope creep and leaving it documented is defensible. If you would rather see the rule applied strictly, this is small enough that the author could fold it in; otherwise a one-line follow-up issue covers it.

**3. List navigation. Fixed, and given its own prominent paragraph.**

The body now carries a bolded scope caveat: arrowing to a different session trips the `cache.session_id != Some(&id)` clause, which is deliberately left forking, so every arrow-key press still pays a full synchronous `capture-pane`, and this PR fixes dwelling rather than moving between selections. The measurement paragraph was reworded from "navigating the list means landing on those frames continuously" to "every ~250ms spent dwelling on a selection". That is the accurate framing and it is stated where a reader will see it.

## The new trust ceiling

The commit adds a module-level `WORKER_TRUST_CEILING_MS` of 10 seconds and extracts the whole idle decision into a pure `idle_poll_due(frozen, idle_elapsed_ms, worker_covers_idle)`. It returns false while frozen or inside the 250ms cadence, and otherwise forks when either the worker is not trustworthy or the elapsed time has passed the ceiling.

**What it does when hit, and whether it can strand a user with a stale pane.** The premise that a ceiling which stops re-forking also stops recovering does not apply here, because this ceiling runs the other way: it is a ceiling on how long the worker is allowed to be the sole source of truth, and hitting it *forces* a synchronous fork regardless of pulse. It strictly increases recovery relative to the previous head, and it directly answers the residual risk I raised last time and that CodeRabbit independently flagged, which is a worker that cycles at full rate while being confidently wrong rather than wedged. The cycle counter cannot see that failure; a periodic authoritative capture can.

**The arithmetic checks out.** `cache.last_refresh` is stamped by `PreviewCache::store_capture`, and both the worker path (`apply_worker_capture`) and the fork path (`refresh_preview_cache_core`) call it. So any pane that publishes at all keeps the timer fresh and never reaches the ceiling; only a pane that has genuinely produced nothing for ten seconds does, and that fork immediately restamps the timer. Steady-state idle fork rate goes from four per second to one per ten seconds, which is the "one fortieth" the body claims.

**The honest cost, which the body does not spell out.** An idle preview now pays one synchronous capture roughly every ten seconds, and at the worst measured 51ms that is one dropped frame every ten seconds. On main it was four such hitches per second. This is clearly the right trade and I would not change it, but it is worth knowing that the ceiling is not free and that "capture_us drops to ~0 on home frames" is now true of 39 frames out of 40 rather than all of them.

**One case the ceiling deliberately does not cover, correctly.** `idle_poll_due` returns false when frozen, ahead of the ceiling check, so a user reading scrollback or holding a selection gets no ceiling fork no matter how long they sit there. That is the entire point of the frozen state, and it matches main, where the idle poll was also gated on `!frozen`. Not a gap.

**And one pathological case the ceiling makes better rather than worse.** In agent live-send, when the fork returns empty for the same session, the capture closure returns `None` and leaves the cache untouched including `last_refresh`, so elapsed keeps growing and the idle clause stays true. On main that means forking every frame. With the ceiling it means forking every frame only after ten seconds of suppression. Pre-existing behavior, slightly improved, not a regression.

**Mutation testing of the ceiling.** Removing the ceiling disjunct so the function reads `!worker_covers_idle` alone:

```
---- tui::home::render::tests::idle_poll_due_suppresses_a_pulsing_worker_only_until_the_trust_ceiling stdout ----
assertion `left == right` failed: frozen=false elapsed=10001 covered=true
  left: false
 right: true
test result: FAILED. 0 passed; 1 failed
```

Covered.

## The two mutants that survived last time

Both now die. This was the substantive gap in the previous round and the author closed it directly.

Reverting the suppression entirely, so `idle_poll_due` ignores `worker_covers_idle`:

```
---- tui::home::render::tests::idle_poll_due_suppresses_a_pulsing_worker_only_until_the_trust_ceiling stdout ----
assertion `left == right` failed: frozen=false elapsed=251 covered=true
  left: true
 right: false
test result: FAILED. 0 passed; 1 failed
```

Deleting `self.preview_worker_pulse = None` from the retarget block in `sync_preview_capture_worker`, which is the one guard whose absence leaves a genuinely stale preview on a newly selected pane:

```
---- tui::home::tests::observe_worker_pulse_needs_a_moving_counter_and_resets_on_retarget stdout ----
panicked at src/tui/home/tests.rs:21113:5:
a retarget must clear the pulse observation
test result: FAILED. 0 passed; 1 failed
```

Both previously passed 808 green tests. The `worker_pulse_step` timeout and no-restamp mutants I ran last round still die as before.

## Test shape against the house rules

Three test functions total for a change of this size, and each one kills a real mutant. No tautological additions.

`idle_poll_due_suppresses_a_pulsing_worker_only_until_the_trust_ceiling` is the house style done properly: one function, a seven-row table of `(frozen, elapsed, covered, expected)`, a comment on each row explaining what it protects, and a formatted assertion message carrying the inputs. Permutations are rows, not functions.

`observe_worker_pulse_needs_a_moving_counter_and_resets_on_retarget` is a sequential state machine rather than a set of independent inputs, so a table would be the wrong shape and the author correctly did not use one. It drives a real spawned `LiveCaptureWorker`, which is the right call because the surviving mutants were in the wiring rather than in the pure helper. Checks worth noting: the worker's target and capture-line budget are never set before the pulse loop, so the loop cycles at the 250ms idle interval without ever forking tmux; `LiveCaptureWorker`'s `Drop` sets the stop flag and nudges, so no thread leaks; the retarget name follows the `aoe_test_*` convention AGENTS.md asks for; and `create_test_env_empty` roots everything in a `TempDir`, so no real user state is touched. I ran it five times consecutively and it completed in 0.01s or less every time, so the hundred-iteration polling loop exits on the first or second pass and there is no meaningful flake surface. Total added runtime is negligible, which is exactly the tier the cost table wants.

Nothing here restates a constant, a default, or a derived impl, and there is no "renders without crashing" smoke test.

**On whether an e2e is owed: no, and more clearly no than last round.** "No subprocess was spawned this frame" is still not observable from `TuiTestHarness`, the rendered output is unchanged by design so a screen assertion has nothing to catch, and a Rust e2e is roughly 2s and strictly serial. The wiring that a heavier test would have existed to cover is now covered by a real-worker unit test costing effectively zero. Demanding an e2e here would be reflexive.

## House rules

No emdashes and no `--` used as a prose separator anywhere in the added lines; I grepped the added side of the diff for both. No `#[allow(dead_code)]`. No leftover references to the old `worker_is_pulsing` name. Both new constants are referenced by production code, not only by tests. `PREVIEW_REFRESH_MS` was hoisted from a function-local const to module scope so `idle_poll_due` can read it; that is necessary rather than gratuitous, it collides with nothing, and it incidentally makes the two existing prose references to it in `live_send.rs` point at something real. The rename of `worker_is_pulsing` to `observe_worker_pulse`, with the doc now saying it takes `&mut self` because the observation it compares against is the state it updates, addresses the naming nit I raised. Its visibility widened to `pub(super)` so the test in `home/tests.rs` can reach it, which matches how `sync_preview_capture_worker` and `refresh_preview_cache_if_needed` are already exposed in that module. Comments throughout explain non-obvious why rather than restating code. No breaking change to call out.

## CI, now that it has actually run

This is the first head at which `CI` and `Tests` executed; every earlier head sat at `action_required` and was never approved, so the previously self-reported green suite was unverified. It is verified now. All 28 named check runs on `796b936c` conclude `success`, including Clippy, Format, Docs, Nix Build, Nix Eval, Supply Chain, Skill Check, the four Cargo Test jobs across linux bare-core, serve-rest, serve-e2e and macos, Vitest, Playwright mocked and both live shards, and both Codecov gates.

The three `skipped` conclusions are all the same job, `close-stale`, appearing once per PR event. That is the repo's stale-issue housekeeping workflow, which skips on every pull request regardless of content. **No skip is attributable to this PR**, and nothing is failing or pending.

CodeRabbit re-reviewed the new head and generated no actionable comments. Its Docstring Coverage pre-merge check still warns, at 73.33% against its own 80% threshold, up from 61.54%; that is CodeRabbit's own bar counting the new test functions, not a repo gate, and none of the merge-box checks reflect it. I would ignore it. Worth noting that CodeRabbit's earlier merge-risk note, that a worker which keeps running while capture fails could leave the pane stale indefinitely, is precisely what the trust ceiling now fixes.

## Does this still plausibly fix #3554?

No, and the body now says so itself, which is the change I wanted. It was downgraded from a hedge to an explicit statement: `Refs #3554`, unlikely to be that reporter's regression, an archived or trashed selection never reaches the capture path at all because `render_preview` returns into the archived or trashed placeholder first, the reporter's 600 rows are mostly those, their bisect window is far newer than this 250ms poll, and #3554 should stay open.

The new commit does not change my assessment, and it was not meant to. The reporter's symptom is unresponsive scrolling over a list dominated by archived and trashed rows, which never touches this code path, and even on live rows scrolling trips the session-change clause that this PR deliberately leaves forking. That list-render cost, and the ~20ms floor with `capture_us=0` that the body defers, both still want their own investigation, and a combined issue covering list cost at 600-plus rows would probably serve #3554 better than either alone.

## Body versus what is pushed

The Description section now matches the code accurately on every point I checked. One stale spot, and it is minor: the **Test Coverage Analysis** section still describes only `worker_pulse_step_trusts_a_running_worker_and_gives_up_on_a_wedged_one` and says "the rendering paths it gates are already covered by the existing `tui::home::render` suite (64 tests, passing unchanged)". Two more tests have landed since that text was written, including the real-worker wiring test that is the strongest coverage in the diff. The section understates the PR rather than overstating it, so nothing misleading follows from it, but it should be refreshed before merge so the checklist reflects what shipped.

Cosmetically, the automated emdash strip in the body left dangling spaces before several commas, for example "in `refresh_preview_cache_core` , whose" and "`capture_us=0` , widget build". That is PR-body typography, not tree content, and the house prose rule does not reach it; worth a pass only because the body is the artifact reviewers read.

## Design and visuals

Design alignment is unchanged and remains aligned. `idle_poll_due` joins `preview_frozen`, `passive_resize_step`, `capture_window_stale`, `capture_is_exhausted` and `worker_pulse_step` as another pure decision helper in the same file, tested by table, which is precisely the local idiom; the ceiling is a term in the existing disjunction rather than a new gate wrapped around it. Nothing visual changes: no color, typography, spacing, layout, icon, or theme surface is touched, and DESIGN.md has no bearing on the diff. The "output is byte-identical, only its cost changes" claim holds, with the ceiling now making it true even in the divergence case where it previously would not have been.

## Blocking versus nits

**Blocking: none.** Everything I raised last round is resolved, and the one substantive risk I named is now mitigated by the ceiling and covered by a test.

**Nits, in descending order:**

- Refresh the Test Coverage Analysis section of the body to mention `idle_poll_due_suppresses_a_pulsing_worker_only_until_the_trust_ceiling` and `observe_worker_pulse_needs_a_moving_counter_and_resets_on_retarget`. It currently undersells the coverage that shipped.
- `force` is now correctly documented as inert but still exists. Under a strict reading of the no-dead-code rule the resolution is deletion, which is under ten lines; either fold it in here or open a follow-up.
- The body could mention that the ceiling costs one synchronous capture per ten seconds on an idle preview, so nobody is surprised by a periodic frame at the old cost.
- Cosmetic dangling spaces before commas in the PR body, left by the emdash strip.

## Verdict

**Design alignment:** Aligned.
**Implementation:** Approve.
**Project fit:** Accept.
**Documentation:** Proportionate, with one stale body section to refresh.
**CI:** Independently green for the first time; 28 successes, 3 skips all from the unrelated `close-stale` housekeeping job.
**Merge risk:** Low. The residual risk from the previous round, a worker that pulses while being wrong, is now bounded at ten seconds by the trust ceiling and pinned by a test. Reverts cleanly with no migration, artifact, or config residue.

Eric162 fixed all three findings, and rather than doing the minimum he took the residual architectural risk I flagged as a merge-risk note, not as a blocking item, and closed it with a ten-line mechanism that is well argued in the code, correctly bounded, and mutation-tested. He also killed both mutants that had survived a green suite, using a real spawned worker for the one that mattered, and corrected the two overclaims in the body rather than deleting them. The commit message is specific about what was wrong and why. This is what a good response to review looks like. Merge it, ideally after a quick refresh of the Test Coverage Analysis section.

PR #3561, opened 2026-08-27
fix: recover stale ACP sessions and ignore unknown WS frames

### LINKED ISSUE #3560: Structured-session ACP: unknown WS control frame causes a reconnect storm; a rejected resumed session terminates the runner
Two independent resiliency defects in the structured-session ACP path, both observed on a live session.

### 1. WS reconnect storm on an unrecognized control-frame `kind`

The native structured view connects `frames=0`, so the daemon sends only `kind`-tagged control frames (`reduced_state`, `transcript_snapshot`, `transcript_delta`, `heartbeat`, `lagged`). In `src/acp/client/ws.rs::parse_text` the catch-all arm sent both a kind-less object *and* an unrecognized `kind` to the `AcpBroadcastFrame` parser, which fails with `missing field ` + "`event`" + `. That `WsError::Parse` reads as a dropped socket, so `acp.tui.ws` reconnects, gets the same connect-snapshot control frame, and loops. Any `kind` a client predates (version skew or a future sentinel) triggers it, the same class as the #3171 heartbeat regression.

Logs: `acp.client.ws ... ws frame parse failed ... missing field event` followed by repeated `acp.tui.ws ws disconnect ... reconnecting`.

### 2. A rejected resumed session terminates the runner

On resume/respawn the worker reuses the stored `acp_session_id`. Verified sequence: `session/load` succeeds and emits `AcpSessionAssigned` for `01a040bb-...`, then the next `session/prompt` is rejected with `Unsupported ACP session`. The prompt-error path returned the error unconditionally, so the runner terminated (`AgentStartupError`) with no recovery, even though the established `session/load`-failure pattern (clear the stored id, `session/new`, `SessionContextReset`) already exists for the handshake path.

### Proposed fix

1. Treat any unrecognized `kind` as an ignorable sentinel; only a kind-less object is parsed as a raw event frame.
2. Recognize the "session unknown/unsupported" prompt rejection and emit `SessionContextReset` before returning the error, so the supervisor clears the stored id and the respawn opens a fresh `session/new` (SQLite transcript preserved for replay).

athal7:
## Description

Two structured-session ACP resiliency fixes (Fixes #3560):

1. `parse_text` skips WS control frames whose `kind` this build does not recognize instead of dropping the socket, ending a tight `acp.tui.ws` reconnect loop (misreported as a missing `event` field). A real event frame never carries `kind`, so only a kind-less object is parsed as a raw event frame.
2. A `session/prompt` rejected because the agent no longer holds the resumed session (`Unsupported ACP session`) emits `SessionContextReset` before returning the error, so the supervisor clears the stored id and the respawn creates a fresh `session/new` with the transcript preserved, instead of terminating the runner.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (code doc-comments only)
- [x] For UI changes: N/A

## Test Coverage Analysis

- [x] N/A: no user-facing dashboard flow changed (client parser + connection-task recovery)

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered by cheaper tiers, `parse_text` sentinel classification (`src/acp/client/ws.rs`), a classifier table plus a real-prompt-round-trip behavior test `unsupported_session_prompt_rejection_emits_context_reset_before_error` (`src/acp/acp_client.rs`). A live-daemon e2e for these paths would be disproportionate per the repo's test-cost guidance.

## AI Usage

- [x] AI was used

**AI Model/Tool used:** anthropic/claude-opus-4-8 via the Oh My Pi (omp) agent harness

AI investigated, implemented, and tested both fixes; a human is reviewing before merge.

## Summary

- Ignore unknown WebSocket control frames and parse only kind-less objects as event frames.
- Recover unsupported resumed ACP sessions by emitting `SessionContextReset` before the error.
- Move session-error detection to the shared rate-limit module.
- Add tests for frame classification and session recovery ordering.

## Benefits

These changes prevent reconnect loops and allow fresh sessions without losing transcript data.


jerome-benoit (changes requested):
Reviewed exact head `f98026195b0c3a36157a911dc56cd26b5949d240`.

The stale-session finding already raised by CodeRabbit remains valid, so I am not duplicating it inline. `is_unsupported_session_error` combines generic words and is applied to every `session/prompt` failure. An unrelated error such as “unsupported session mode” can therefore emit `SessionContextReset`, clear the persisted ACP session ID, and restart with `session/new`.

The regression test currently establishes a fresh session: its `SpawnConfig` has no stored ID and the fake agent handles `session/new`. It therefore proves the reset ordering, but does not reproduce the reported successful `session/load` or attach/resume path.

Please track whether the active session was actually loaded or reattached, restrict recovery to that provenance, tighten the documented error fingerprints or structured error data, and add negative generic-message cases plus a real loaded/resumed-session test.

The WebSocket control-frame fix is sound and covers unknown, null, and non-string `kind` values. CI is green at this SHA. The branch currently conflicts with `main`.

jerome-benoit (changes requested):
Re-reviewed `2cb1971f29055cfa0d87d1a3a2fad13a7751db2e`, including the latest reply and changes against main, with two independent sequential analyses and code-level cross-validation.

**P2, still blocking: the stale-session classifier is broad again.** `src/acp/acp_client/rate_limit.rs:57-62` accepts independent words rather than the narrower phrases removed from `errors.rs`. Both `Unsupported content block in session/prompt` and `Method not found: session/prompt`, previously explicit negative test cases, now match. For a stored session, `connection.rs:1992-2009` then emits a context reset; `supervisor.rs:2265-2276` clears the cached session ID. An unrelated prompt error can therefore discard the agent context. Please restore narrow fingerprints and the negative regression cases. This is the existing review concern, not a new duplicate inline.

The consolidated reset/flag handling and reset-before-terminal assertion are fixed. The test now supplies a stored ID and exercises `session/load`, so the previous fresh-session objection no longer applies. A separate socket-backed `attach` test is a coverage limitation, not an independent blocker. No remaining issue identified in the WebSocket frame classification.

Evidence: exact-SHA `git show` and base/head `git diff`; `git diff --check` is clean. No local tests run. I approved the waiting exact-head workflows; CI validation is still in progress.

jerome-benoit:
Correction 2 (2026-09-10 17:1x): this PR itself was closed at 15:56:44 unmerged, same instant as #3873. Since its three own commits (skip unknown WS control-frame kinds, route WS frames on kind presence, recover a resumed session the agent rejects + the consolidated flagged branch) are not on main, the companion #3882 now carries the complete set, not only the classifier fix - its description has been updated to state that it closes this PR's content. Awaiting review there.

PR #3562, opened 2026-08-27
fix: route structured messages to composer

athal7:
## Description

Routes the TUI home `m` shortcut for structured sessions to the active structured composer. Adds focused coverage for the routing.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
openai-codex/gpt-5.6-terra

Used to implement and verify the TUI action change.

## Summary

- Fixed the TUI home `m` shortcut for structured sessions.
- Routed messages to the active structured composer instead of the send-message dialog.
- Preserved buffered text during composer activation.
- Merged buffered text for the same session and replaced it for a different session.
- Prevented buffered text from leaking between sessions.
- Completed buffered pastes through the asynchronous file-loading path.
- Added regression tests for routing and buffered paste handling.

This improves message entry reliability without changing dashboard, CLI, session lifecycle, or documentation behavior.


src/tui/home/input.rs
```diff
@@ -2859,7 +2859,14 @@ impl HomeView {
                     }
                 }
             }
-            ActionId::SendMessage => self.open_send_message_dialog(),
+            ActionId::SendMessage => {
+                #[cfg(feature = "serve")]
+                if let Some(id) = self.selected_structured_session() {
+                    self.exit_live_send_if_active();
+                    return Some(Action::OpenStructuredView(id));
```
jerome-benoit, line 2882: **[P2] Forward pending input into the structured composer**

This early return bypasses `open_send_message_dialog()`, which is currently the only path that drains `pending_paste`. Paste or strict-mode typing captured while no legacy target is available therefore opens the structured composer empty, remains buffered, and can later surface in another session's legacy dialog. Please transfer and clear the pending input only after the structured composer is successfully activated, preserving it if activation fails, and add a regression test covering the buffered-paste path.

jerome-benoit (changes requested):
The structured routing itself is consistent with the existing action flow, but the new early return leaves buffered user input behind. Please address the inline finding before merging.

src/tui/home/mod.rs
```diff
@@ -661,6 +661,12 @@ pub struct HomeView {
     /// into the next compose dialog the user opens, so voice/dictation never
     /// gets thrown on the floor with a scolding info dialog.
     pub(super) pending_paste: Option<String>,
+    /// Pasted text captured at the home view that we couldn't immediately
+    /// route into the structured composer. Drained by
+    /// `open_structured_view` after the view activates, preserving the
+    /// text if activation fails so the next 'm' press can still drain it.
+    #[cfg(feature = "serve")]
+    pub(super) pending_paste_for_structured_view: Option<String>,
```
jerome-benoit: [P2] Bind buffered text to the intended structured session

This global `Option<String>` loses the target selected when `m` moved the pending input. If opening session A is cancelled or fails, selecting B and opening its structured view drains A's text into whichever preview is mounted because `drain_pending_paste_for_structured_view` does not check a session id. A failed connection can also leave an older preview mounted and receive the buffer. Store `(session_id, text)`, preserve or merge only entries for that target, and drain only when the mounted view has the same session id.

src/tui/structured_view/embedded.rs
```diff
@@ -226,4 +226,10 @@ impl EmbeddedView {
     pub fn selection_text(&self, width: u16) -> ratatui::text::Text<'static> {
         render::wrapped_transcript(&self.state, &crate::tui::styles::Theme::default(), width)
     }
+
+    /// Paste text into the composer, focusing it if needed.
+    /// Used to forward buffered paste into the structured composer.
+    pub fn paste_text(&mut self, text: &str) {
+        super::paste_into_composer(&mut self.state, text);
```
jerome-benoit, line 236: [P2] Reuse the complete async paste pipeline

The normal structured-view paste path calls `paste_into_composer` and then `ensure_files_loaded`. This synchronous handoff only performs the first step. Buffered text ending in an `@` mention therefore opens the mention state while `FileIndex` remains `Unloaded`, leaving the picker empty until a later keystroke happens to trigger the load. Route this through the same async paste handler, or expose a shared async helper, and test a transferred `@` mention.

jerome-benoit (changes requested):
Re-review at exact head `4171d7546a31ff9f59169907b08171a01244dbcc`, cross-validated through two independent sequential analyses and final code-level arbitration.

The previous buffered-input finding is fixed when the target preview is already mounted or a daemon is already reachable, but the correction is not complete. Two additional P2 findings remain inline.

I did not duplicate the existing CodeRabbit findings, both of which are valid P2s:

- Repeated attempts overwrite `pending_paste_for_structured_view` instead of preserving both buffered inputs: https://github.com/agent-of-empires/agent-of-empires/pull/3562#discussion_r3900077557
- The accepted "start a local daemon" path connects and activates without draining the buffered text: https://github.com/agent-of-empires/agent-of-empires/pull/3562#pullrequestreview-5073234670

The PR is currently `DIRTY` against `main`. `git diff --check 3a3504c6044878f494a38b09d92beb53dfd10de8 4171d7546a31ff9f59169907b08171a01244dbcc` is clean. No local build or test was run because the host is already under load. GitHub currently reports only title/template checks for this head.

src/tui/home/input.rs
```diff
@@ -2859,7 +2859,30 @@ impl HomeView {
...
+                    // bound to the selected session: a second paste for the
+                    // same target appends instead of silently replacing the
+                    // earlier buffered text, and a different target takes
+                    // over only because its predecessor can no longer drain.
+                    if let Some(buf) = self.pending_paste.take() {
+                        match self.pending_paste_for_structured_view.as_mut() {
+                            Some((existing_id, existing)) if *existing_id == id => {
+                                existing.push_str(&buf);
+                            }
+                            _ => self.pending_paste_for_structured_view = Some((id.clone(), buf)),
```
jerome-benoit, line 2878: [P2] Preserve the previous target's unsent draft

If opening A fails or is cancelled, its text is retained as `(A, draft)`. A new paste followed by `m` on B reaches this replacement arm and permanently drops A's draft. Returning to A remains possible, contrary to the comment's rationale. Preserve the unsent text per target, or offer equivalent explicit recovery instead of silently overwriting it. Cover A failing to open, B receiving new text, and returning to A; the current cross-target test asserts the destructive replacement.

jerome-benoit (changes requested):
Re-reviewed `58407ad2c372b62df3e5768139357ba7704b00be`, including the latest reply. Three P2 changes remain required. The first two are already covered, so I am not duplicating them inline:

- **Consume a successfully forwarded buffer** ([existing finding](https://github.com/agent-of-empires/agent-of-empires/pull/3562#discussion_r3972730283)). `app.rs:3160-3175` clones and inserts the text but never clears the stored entry. Returning home and reopening the same session inserts it again, including after sending the original draft. Clear it after insertion, retaining it only when the target view cannot receive it.
- **Drain after starting the daemon** ([existing finding](https://github.com/agent-of-empires/agent-of-empires/pull/3562#pullrequestreview-5073234670)). `app.rs:3225-3226` connects and activates without the drain. Pasting, pressing `m`, then accepting daemon startup opens an empty composer until another activation. Use the same target-checked handoff on this continuation.

- **Preserve an earlier target's unsent draft** (new inline). `home/input.rs:2878` replaces A's retained text when a subsequent paste targets B. Returning to A is possible, so failed activation does not make its draft disposable. Preserve it per target or provide equivalent explicit recovery.

Same-target merging, session binding and the async mention-file loading path are fixed. The added tests cover HomeView routing/buffering, not successful App-level consumption or the daemon-start continuation; please cover those behaviors and preservation across target changes. The new cross-target test currently codifies the loss.

Cross-validated through two independent sequential analyses and exact-SHA source inspection. `git diff --check` is clean; no local tests run. I approved the waiting exact-head CI workflows; their results are still pending.

src/tui/app.rs
```diff
@@ -3148,6 +3150,31 @@ impl App {
...
+            .structured_preview
+            .as_ref()
+            .is_some_and(|v| v.session_id() == target);
+        if !matches_mounted {
+            // Activation failed or the mounted view is another session's:
+            // keep the buffer for the captured session.
+            return;
+        }
+        if let Some(view) = self.home.structured_preview.as_mut() {
+            view.paste_text_with_file_load(&buf).await;
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in #3868: the mounted view now consumes its own entry on success (`pending_paste_for_structured_view.remove(&mounted)` before `paste_text_with_file_load`), so a later `m` activation cannot re-insert the same text.

src/tui/home/input.rs
jerome-benoit: #3868 moved drafts to a session-keyed map, preserving A when B receives new text. Correction to my previous evidence claim: send_message_keeps_buffered_paste_per_session proves HomeView routing/retention, not draining on return. The later App-level drain test exercises the real composer handoff and entry consumption. Follow-up #3906 (73c68762) extends that App boundary to reject a mismatched requested target even if an older view remains mounted, and verifies a repeated matching drain does not duplicate text. The daemon-start caller passes the requested ID too; the test does not launch a daemon or simulate a construction failure.

PR #3579, opened 2026-08-29
fix(session): make Pi's conversation authoritative instead of guessed

### LINKED ISSUE #3576: Wrong pi sessions restored after quit
I repeatedly had the case that I quit aoe, upgrade it, launch it again, just to see that my >20 sessions were still listed in the left pane (TUI), but sometimes simply had the wrong session. This is annoying and confusing.

Currently on aoe 1.15.1, but this is not a recent issue. I don't think the upgrade step is necessary for reproduction, I just happen to quit aoe only to upgrade it.

Edit: corrected aoe version

njbrake:
## Description

Sessions sharing a project directory adopted each other's Pi conversation after a quit (#3576): capture picked the newest file in a store keyed by cwd, which names no pane.

Pi is now asked rather than inferred. A launch pins the conversation with `--session-id`, and an AoE extension loaded with `-e` publishes the pane's conversation and transcript path on every `session_start`, covering `/new` and a moved worktree, host and sandbox alike. The store scan goes, along with the launch floor, exclusion set, drain refusal and poller retirement that existed to bound it.

A binary AoE cannot vouch for gets no capture rather than a guess.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: N/A, no UI changes

## Test Coverage Analysis

**TUI / CLI (e2e test)**

`pi_pinned_session_id_e2e` drives the real binary through tmux. The extension paths need a real pi, which CI lacks, so their decision points are unit tested (`pi_resumes_by_published_path_only_for_its_own_transcript`, `pi_accepts_a_sidecar_retarget_and_keeps_its_poller`, `pi_stop_persists_the_conversation_the_extension_published`, `pi_polls_only_what_names_a_pane`, `sandboxed_pi_launch_carries_the_extension_and_container_paths`, `sandboxed_pi_mounts_its_instance_dir_and_extension`), each failing when its rule is reverted.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Every claim about pi here was reproduced against real binaries, 0.75.5 through 0.84.4: flag semantics, that the session file lands on the first message, that `session_start` fires on `/new`, and that a moved worktree resumes by path.

Fixes #3576


src/session/capture.rs
```diff
@@ -1008,8 +1060,11 @@ fn compose_exclusion_in(
 /// another session can capture the parked conversation before its owner swaps
 /// back. Claude, host Codex, and host Kimi additionally need inactive
 /// same-tool protection because their shared-store MRU scans can select a
-/// conversation after the owning pane disappears. Sandboxed Codex and Kimi
-/// omit that protection because their stores are instance-private or are not
+/// conversation after the owning pane disappears. Host Pi is included for its
+/// poller alone: acquisition no longer scans its store at all (#3576), but a
+/// peer that goes inactive after this pane launched can still leave the
```
jerome-benoit, line 713: **[P1] Prevent a pinned Pi session from adopting a stopped cross-profile peer**

This protection is scoped to the current profile. Consider profile A with a pinned Pi pane, then profile B starts another Pi pane in the same cwd and shared Pi home. While B is alive, tmux ownership excludes its ID. Once B stops, that ownership disappears, A's snapshotted persisted exclusions do not contain B, and the floored MRU scan can return B's newer JSONL. The profile-A sync pass has no owner for that ID, so it persists the reassignment; A's next restart then emits `--session-id` with B's conversation.

Please either make stopped-peer ownership global across profiles, keep an already pinned host-Pi ID immutable under heuristic polling, or use a pane-attributable source for `/new`. The E2E should cover two profiles sharing one Pi home, real shim-created JSONLs, stopping the newer peer, and verifying that the older session's stored and restarted ID never changes.

jerome-benoit (changes requested):
The launch-time pinning is sound, but the remaining host-store poller can still overwrite that pin with a stopped peer's conversation when the peer belongs to another profile. This leaves a concrete cross-profile form of #3576 unresolved.

src/session/instance.rs
```diff
@@ -2994,6 +3004,20 @@ impl Instance {
...
+        // it could make is a guess against a store shared with every session
+        // on this path, this profile's and other profiles' alike. Live tmux
+        // ownership hides that while a peer runs, but a peer that writes after
+        // this pane launched and then stops leaves the freshest file inside the
+        // floor and outside the (single-profile) persisted exclusion, and the
+        // scan would hand its conversation to this pane (#3576). A pinned pane
+        // simply does not poll. The trade is that a `/new` started inside the
+        // pane is no longer adopted, which needs a pane-attributable signal
+        // (pi's extension API, blocked on #3380) rather than a better guess.
+        if self.pi_session_id_pinnable() {
```
jerome-benoit: **[P1] Persist whether this live Pi pane was actually pinned**

This recomputes Pi's current pinnability, but poller repair also runs for already-live tmux panes after an AoE restart. The original poller and `before_session` environment are not persisted, and `pi --help` may now fail or resolve differently because of a timeout, binary/PATH change, or profile change. A pane originally launched with `--session-id` can therefore be reclassified as unpinned and receive the shared-store scanner again; any later foreign write can then retarget the pinned conversation and reproduce #3576.

Please record the actual per-launch mode in durable pane-generation metadata and use that fact during repair. The capability probe should only decide how a new pane is launched. Add coverage that keeps the pane alive, restarts AoE with a changed probe result, and verifies that a foreign session cannot be adopted.

jerome-benoit (changes requested):
The stable launch path now keeps a pinned Pi conversation immutable, but poller repair still derives that decision from the current capability probe instead of the mode used by the live pane. This can restore the same shared-store attribution bug after an AoE restart.

src/session/instance.rs
```diff
@@ -2993,7 +3003,26 @@ impl Instance {
...
+    /// a peer while it runs, and `session::sync` refuses an id another row
+    /// claims, but neither covers a conversation no row claims at all, which
+    /// is what a `pi` run by hand in the same directory leaves behind: that
+    /// one the scan would adopt over the pane's own (#3576). The trade is that
+    /// a `/new` started inside such a pane is not adopted; that needs a
+    /// pane-attributable signal (pi's extension API, blocked on #3380) rather
+    /// than a better guess. Read from memory only: this runs per session on
+    /// every TUI refresh through `repair_session_id_poller_if_needed`.
     pub fn supports_session_poller(&self) -> bool {
+        if self.tool == "pi" && !self.is_sandboxed() && self.agent_session_id.is_some() {
```
jerome-benoit: **[P1] Stop an unpinned host-Pi poller after its first successful capture**

This predicate prevents creating or repairing a poller once `agent_session_id` is `Some`, but it does not stop the poller that performs the `None -> Some` transition. Old Pi binaries, command overrides, and custom-PATH launches start without an ID, so their floored poller captures the first conversation and then keeps scanning indefinitely. The sync path only assigns the new ID; the existing worker remains live and reports every later distinct observation. A newer unowned JSONL from a manually launched Pi can therefore replace the now-authoritative ID and reproduce #3576.

Please make host-Pi capture one-shot, or explicitly stop and discard its poller immediately after the first ID is successfully persisted. Cover this with an unpinnable Pi that captures its own ID, followed by a newer foreign conversation that must not replace it.

jerome-benoit (changes requested):
The reattach decision is now independent of the current Pi probe, but the state-based gate does not terminate the poller that creates that state. An unpinnable host Pi can therefore still be retargeted after its first successful capture.

src/session/instance.rs
```diff
@@ -3289,18 +3376,33 @@ impl Instance {
...
+                // recovered sessions a co-located peer's conversation (#3576).
+                // A sandboxed pane still scans, because AoE cannot probe the
+                // container's pi and so never pins one: the scan is that
+                // pane's only source of identity. It is not a private store,
+                // though. `sandbox_dir_for` bind-mounts one `~/.pi/sandbox`
+                // into every sandboxed Pi container, so co-located sandboxed
+                // peers keep #3576's exposure here. Recovery passes no floor
+                // because resuming a session older than this launch is the
+                // point.
+                if !self.is_sandboxed() {
```
jerome-benoit: **[P1] Do not retroactively guess a sandboxed Pi conversation**

The post-launch poller guard now covers sandboxed Pi, but this retroactive path still scans the same shared store without a launch floor. `sandbox_dir_for` only gives Codex an instance-specific directory, so every Pi container bind-mounts the same `.pi/sandbox`. The selector then chooses the newest same-CWD JSONL.

For a row with an existing `agent_session_id`, `acquire_session_id()` calls this path through `capture_freshest_session_id()` and can replace the stored ID before relaunch. For an id-less live pane, self-heal can adopt a newer manual conversation that no ownership exclusion can identify.

Please refuse retroactive Pi scans in sandbox as well and rely on the launch-floored poller, or on a future pane-attributable signal. Add coverage proving that a stored sandboxed Pi ID survives a newer foreign JSONL.

src/session/instance.rs
```diff
@@ -3157,21 +3194,68 @@ impl Instance {
...
+        session_id: Option<&str>,
+    ) -> bool {
+        let takes_pinning_arm = self.tool == "pi"
+            && pi_pinnable
+            // `--session-id` needs an exact project-local id and creates an
+            // empty conversation under anything else, while `--session` takes
+            // a partial and searches. A user can pin a partial through
+            // `aoe session set-session-id`, so the creating arm is only for an
+            // id of the shape AoE mints and captures.
+            && session_id.is_some_and(|sid| uuid::Uuid::parse_str(sid).is_ok());
```
jerome-benoit: **[P2] Preserve provenance for explicit full-UUID pins**

UUID syntax does not prove that AoE minted the ID. `set-session-id` stores both partial and full values as `ResumeIntent::Use`, but this predicate routes every parseable UUID to `--session-id`. Pi only searches the current project for that flag and creates an empty conversation when the ID is absent; `--session` is the arm that performs the broader ID search.

Please base this decision on provenance: explicit `ResumeIntent::Use` values should keep `--session`, while `--session-id` should be reserved for IDs AoE actually minted or anchored itself. Cover both manual partial and manual full-UUID values in the existing table.

jerome-benoit (changes requested):
The post-launch immutability rule now covers host and sandboxed Pi sessions, but the sandbox retroactive path can still replace a known ID before relaunch. Two additional gaps remain around explicit full-UUID provenance and retiring the first-capture poller; the latter is followed up in the existing thread rather than duplicated inline. CodeQL is also failing at this head.

src/session/instance.rs
jerome-benoit: The replacement is now rejected, but the poller is still not retired when the first `None -> Some` capture is successfully applied. `SidWrite::Applied` only queues the assignment, and the later `to_apply` loop sets `agent_session_id` without stopping the worker. Unless a second distinct JSONL appears, the poller remains alive indefinitely and consumes one of the global 50 slots.

Please retire a Pi poller immediately after its first successful capture and extend the existing id-less Pi test to assert that the worker is no longer running.

src/session/sync.rs
```diff
@@ -362,6 +385,19 @@ fn drain_and_persist_session_ids_inner(
         if let Some(inst) = instances.iter_mut().find(|i| i.id == *id) {
             inst.agent_session_id = Some(sid.clone());
             inst.resume_probe_failed_sid = None;
+            // A Pi capture is the pane's whole identity and its last: the
+            // store is keyed by cwd alone, so a later observation could only
+            // propose someone else's conversation. Retire the worker here
+            // rather than waiting for it to offer one, which would hold a
+            // poller slot indefinitely.
+            if inst.tool == "pi" {
```
jerome-benoit: **[P2] Retire Pi when a skipped CAS converges to a known ID**

This retires Pi only from `to_apply`. A concurrent drainer can persist the first `None -> Some` capture before this snapshot reaches the CAS; `SidWrite::Skipped` then reloads that durable `disk_sid` into `to_rollback`, acknowledges the same observation, and leaves the poller running. The TUI reinserts that rolled-back snapshot, while the daemon path likewise keeps the live poller; once the ID is known, `supports_session_poller()` returns false, so repair will not clean it up. The losing process can therefore retain one of the global 50 poller slots indefinitely and keep issuing Pi scans, including sandbox `docker exec` calls.

Please retire a Pi poller whenever rollback converges it to a durable non-empty ID (at least when it equals the observed SID), and cover the two-drainer CAS race.

src/session/instance.rs
```diff
@@ -12041,13 +12166,127 @@ mod tests {
...
+            enabled: true,
+            container_id: None,
+            image: "test-image".to_string(),
+            container_name: "aoe-pi-retro".to_string(),
+            extra_env: None,
+            custom_instruction: None,
+            container_workdir: None,
+            before_start_env: Vec::new(),
+        });
+        assert_eq!(inst.try_retroactive_capture(), None);
```
jerome-benoit: **[P2] Make the sandbox regression test fail on the parent**

This assertion does not exercise the removed sandbox scan. The fixture names a container that does not exist and seeds no Pi conversation, so the parent implementation's `docker exec` fails and `.ok()` also returns `None`; the test passes before and after this fix. That gives no regression protection for the exact foreign-JSONL path being removed.

Please use an injected capture seam or a fixture where the parent path actually returns a sandbox Pi SID, then assert that the new path refuses it.

jerome-benoit (changes requested):
The three previously reported logic defects are addressed on this head: Pi no longer scans shared stores retroactively, explicit `ResumeIntent::Use` values keep `--session`, and an `Applied` first capture retires its poller. The lifecycle is still incomplete when the first capture loses the persistence CAS and converges through `SidWrite::Skipped`, and the new sandbox regression test cannot distinguish this head from its parent. CodeQL is also still failing with an existing bot finding, which is not duplicated here.

src/session/instance.rs
```diff
@@ -5883,6 +5965,12 @@ impl Instance {
                 }
             }
             "pi" => {
+                // Only host launches with no id yet and sandboxed panes reach
+                // here; `supports_session_poller` refuses the rest. Floored (see
+                // `capture_pi_session_id`), which is how an unpinned launch
+                // gets its id at all: pi writes the session file on the first
+                // message, necessarily after launch.
+                let launch_time_ms = crate::util::now_ms() as f64;
```
jerome-benoit: [P1] The first unpinned Pi capture is still not attributable to this pane. This floor is sampled after `create_with_size_env_and_container_env` has already started the agent, while the selector accepts any same-CWD file newer than `floor - 2s` and chooses the newest. An unowned manual or peer Pi session written shortly before or concurrently with this launch can therefore be returned by the immediate first poll before this pane writes its own file; `session::sync` then persists that ID and retires the Pi poller, making the wrong assignment durable. Moving the timestamp before spawn alone would not close the two-second ambiguity window. Please use an instance-attributable signal or isolated store, or fail closed for launch modes where neither is available, and add a deterministic regression with a foreign file inside this boundary before the pane's own file appears.

jerome-benoit (changes requested):
Re-review of `bf67de32`. The previously reported CAS-loss cleanup and sandbox regression-test gaps are fixed, and the CI checks are green at this head. One root-cause gap remains for unpinnable Pi launches: their first ID is still selected heuristically from a shared store, so it is not attributable to this pane. See the inline comment.

src/session/instance/session_id.rs
```diff
@@ -347,6 +356,47 @@ impl Instance {
...
+    /// message, so a pane pinned and never prompted holds an id `--session`
+    /// exits 1 on, and `--session-id` recreates it.
+    fn resume_flag_arm_is_existing(
+        &self,
+        is_existing: bool,
+        pi_pinnable: bool,
+        session_id: Option<&str>,
+        explicitly_pinned: bool,
+    ) -> bool {
+        // `--session-id` searches this project only and creates the
```
jerome-benoit, line 585: **[P1] Preserve the existing transcript across a worktree move**

This assumes every non-explicit UUID is still an unwritten AoE pin, but AoE also retains `agent_session_id` when a managed worktree is renamed or moved. The transcript remains indexed under the old cwd while the next launch runs here with the new cwd. Pi's `--session-id` lookup is project-local and creates the ID when it is absent, whereas `--session` performs the global lookup. The restart therefore creates an empty conversation with the same UUID instead of reopening the existing transcript.

Please distinguish an unwritten pin from an ID whose transcript already exists, then resolve the latter globally (or persist its exact provenance/path). Add a regression that writes a Pi transcript, moves the managed worktree, restarts the session, and verifies that the original history is resumed.

src/session/sync.rs
```diff
@@ -184,6 +186,21 @@ fn drain_and_persist_session_ids_inner(
             acknowledge_poller_observation(inst, &observation);
```
jerome-benoit, line 184: **[P2] Retire a Pi poller when reload already imported its ID**

`merge_runtime_from_reload` can import the durable `agent_session_id` while preserving the previous runtime poller. If that poller then reports the same ID, this branch acknowledges the observation and returns without adding the instance to `to_retire`. Since a known Pi ID also makes `supports_session_poller()` false, repair will not reap this worker later; it can keep one of the global 50 poller slots and continue scanning until the pane exits.

Please retire the Pi poller in this equality branch, or drop a preserved poller as soon as reload imports a known Pi ID. Cover the reload-before-drain interleaving where the poller reports the same durable ID.

jerome-benoit (changes requested):
Re-review of `837e9aee`. The CI checks are green, but the previously reported first-capture blocker still applies at its new location: an unpinnable Pi launch still selects its initial ID heuristically from a shared store. The existing thread covers that issue, so I am not duplicating it inline. Although #3592 appears to remove that scan, it is stacked on this head and is not part of this PR.

I found one additional resume regression in the current pinning arm: after a managed worktree moves, AoE retains the UUID but `--session-id` only searches the new project directory, so Pi can silently create an empty conversation under the same UUID instead of reopening the existing transcript. A lower-severity reload path can also retain a Pi poller after its durable ID has already been imported. Details are inline.

src/session/instance/kill.rs
```diff
@@ -3,6 +3,58 @@
...
+            }
+            Ok(())
+        });
+    }
+
+    fn flush_pi_sidecar_conversation(&self, storage: &crate::session::storage::Storage) {
+        if !self.uses_pi_session_sidecar() {
+            return;
+        }
+        let Some(published) = crate::hooks::read_hook_session_id(&self.id) else {
```
jerome-benoit: **[P1] Do not expire the authoritative Pi ID before the stop flush**

The extension writes this sidecar on `session_start`, including `/new`, but `read_hook_session_id()` rejects it after five minutes. In the CLI-only lifecycle the initial poller is already gone, so `/new`, waiting more than five minutes, then stopping reaches this return, skips the published path, and subsequently deletes the instance sidecar directory. The next launch therefore keeps the old conversation instead of the one the pane published. Please use a validated no-age read for this final authoritative flush, persist the ID/path pair together, and add a stale-mtime `/new` regression.

src/session/instance/session_id.rs
```diff
@@ -347,6 +365,131 @@ impl Instance {
...
+    /// still exists and belongs to the conversation we hold.
+    ///
+    /// Pi names its files `<timestamp>_<uuid>.jsonl`, so the id check is what
+    /// keeps a path left over from a previous conversation (a `/new` the drain
+    /// recorded but no stop flushed) from resuming the wrong one.
+    fn pi_resumable_transcript(&self) -> Option<String> {
+        let path = self.pi_session_path.as_deref()?;
+        let id = self.agent_session_id.as_deref()?;
+        let name = std::path::Path::new(path).file_name()?.to_str()?;
+        (name.contains(id) && std::path::Path::new(path).is_file()).then(|| path.to_string())
```
jerome-benoit: **[P1] Validate sandbox transcript paths in the host namespace**

For sandboxed Pi, `getSessionFile()` runs inside the container and publishes a path under `/root/.pi`, while the corresponding host storage is `$HOME/.pi/sandbox` bind-mounted at `/root/.pi`. This host-side `is_file()` checks the container path verbatim, so it rejects a valid sandbox transcript and falls back to project-local `--session-id` after a worktree move. Please validate through the corresponding host bind source while retaining the container path for Pi's argv, or perform the validation inside the container. Add a sandbox move/restart regression that resumes the original history.

src/session/container_config.rs
```diff
@@ -1799,6 +1834,13 @@ pub(crate) fn build_container_config(
         }
     }
 
+    // Pi declares no hooks, so the block below skips it, but its extension
+    // publishes into the same per-instance dir and needs both that dir and the
+    // extension file inside the container.
+    if active_agent.is_some_and(|a| a.name == "pi") {
+        volumes.extend(pi_extension_mounts(instance_id));
```
jerome-benoit: **[P1] Migrate existing Pi containers before relying on these mounts**

Docker cannot add bind mounts to an existing container, and `get_container_for_instance()` reuses or starts an existing named container without calling `build_container_config()`. After upgrading, an existing Pi sandbox is still launched with `-e /tmp/aoe-pi-session-id.js` and the sidecar environment, but it has neither of these new mounts, so the extension cannot publish the conversation or transcript. Please recreate/version-migrate affected containers (or fail with an actionable recreation path) before enabling this launch arm, and cover an existing-container upgrade.

src/session/instance/session_id.rs
```diff
@@ -377,8 +533,25 @@ impl Instance {
...
             session_id = None;
         }
-        let emitted =
-            append_resume_flags(&self.tool, session_id.as_deref(), is_existing, cmd, context);
+        // A transcript the pane published outranks its id: `--session <path>`
+        // resolves the conversation wherever it was started, while
+        // `--session-id` looks only in the current project and would create an
+        // empty one under the same uuid after a worktree move.
+        if is_existing && session_id.is_some() {
+            if let Some(path) = self.pi_resumable_transcript() {
```
jerome-benoit, line 646: **[P2] Do not let a stored path override an explicit partial pin**

This optimization runs even when `ResumeIntent::Use` made `explicitly_pinned` true, while `pi_resumable_transcript()` only requires `filename.contains(id)`. `set-session-id` intentionally accepts partial IDs, so a stale transcript can match that short value in its timestamp or another filename segment and silently replace the user's explicit target with the old path. Skip the stored-path arm for explicit pins, or require an exact parsed conversation ID and clear/reassociate the stored path when the pin changes.

jerome-benoit (changes requested):
Reviewed the exact current head `754ae227ac1193ada27b8de273989d7392a4a358`, the existing review threads, and Pi 0.84.4's resume behavior. Removing the shared-store scan closes the original cross-pane attribution bug, and resuming by the published transcript path fixes the host worktree-move case. CI is green. However, the implementation still loses CLI-only `/new` updates after the sidecar age cutoff, cannot validate the container-namespace transcript path on the host, does not migrate existing Pi sandbox containers to the newly required mounts, and can let a stale path override an explicit partial pin. These are correctness gaps in the feature this PR is meant to restore, so I am keeping `REQUEST_CHANGES`. No local build or test was run; this rereview used static exact-SHA inspection plus the completed live CI results.

src/session/instance/polling.rs
```diff
@@ -151,24 +162,28 @@ impl Instance {
...
-                        extra_excludes.clone(),
-                    ))
+                // Sidecar or nothing. Pi's store is keyed by cwd and names no
+                // pane, so a scan of it can only guess, and a guess is what
+                // #3576 cost. A binary that cannot load the extension gets no
+                // poller and, absent a pin, no resume.
+                if !self.uses_pi_session_sidecar() {
+                    return;
                 }
+                let inner = crate::session::capture::pi_sidecar_poll_fn(self.id.clone());
```
jerome-benoit: **[P1] Read the sandbox Pi sidecar from its bind-backed path**

This routes host and sandboxed Pi through `pi_sidecar_poll_fn`, but that closure always reads `hooks::read_hook_session_id()` from the host hook directory. A sandboxed Pi publishes under `/root/.pi/aoe-session/<instance>/session_id`, backed by `pi_sandbox_sidecar()`, so the live poller never persists its first ID or a later `/new`; only a later clean stop can recover it. Please select the reader by namespace (or reuse a sandbox-aware sidecar reader) and add a sandbox live-capture regression.

src/session/instance/kill.rs
```diff
@@ -3,6 +3,60 @@
...
+        }
+        let _ = storage.update(|instances, _| {
+            if let Some(inst) = instances.iter_mut().find(|i| i.id == self.id) {
+                inst.pi_session_path = Some(path.clone());
+            }
+            Ok(())
+        });
+    }
+
+    fn flush_pi_sidecar_conversation(&self, storage: &crate::session::storage::Storage) {
```
jerome-benoit: **[P1] Apply the any-age Pi flush to restart as well as stop**

This fixes an idle `/new` for `stop()`, but the restart path only stops and drains an existing poller before killing the pane; it never calls this any-age flush. For a host Pi with no live poller, `/new` followed by more than the sidecar freshness window and `aoe session restart` makes acquisition use the age-limited read and retain the old durable ID. Please invoke the same flush under the restart lifecycle lock before teardown and cover the stale-mtime restart case.

jerome-benoit (changes requested):
Re-review of exact head `0863a7d3`. The recent commits address the previously reported stale stop flush, container-path mapping, existing-container mount, and explicit-pin issues. Two lifecycle blockers remain: sandboxed Pi polls the host hook sidecar instead of the sandbox bind-backed sidecar, and restart bypasses the any-age Pi flush that stop now performs. Both were independently challenged and revalidated against this head. The Linux CI rerun for the latest test-serialization commit is still in progress. No local build or test was run because the host is under load; this review uses exact-SHA static inspection and live GitHub evidence.

src/session/instance/session_id.rs
```diff
@@ -347,6 +369,202 @@ impl Instance {
...
+            return;
+        }
+        if let Some(path) = self.pi_published_session_path() {
+            self.pi_session_path = Some(path);
+        }
+    }
+
+    pub(crate) fn uses_pi_session_sidecar(&self) -> bool {
+        self.tool == "pi"
+            && (self.pi_extension_launched || crate::hooks::session_id_sidecar_exists(&self.id))
```
jerome-benoit: **[P1] Rediscover sandbox Pi sidecars after AoE reload**

`pi_extension_launched` is runtime-only (`#[serde(skip)]`), so it resets to `false` when the daemon or TUI reloads this still-live session. This fallback checks only the host hook directory, but a sandboxed Pi publishes under the bind-backed `$HOME/.pi/sandbox/aoe-session/<instance>/session_id`. The reloaded instance therefore fails `uses_pi_session_sidecar()`: poller repair is disabled, and `flush_pi_sidecar_conversation()` also returns before reading the sandbox sidecar on stop/restart. A later `/new` can consequently remain unpersisted. Please include the sandbox sidecar path in this existence check and add a serialized/reloaded sandbox regression that proves both poller repair and final flush still observe the published ID/path.

jerome-benoit (changes requested):
Re-review of exact head `6b6d6184`. The two blockers reported at `0863a7d3` are fixed on their direct paths: the live poller now reads the bind-backed sandbox sidecar, and restart teardown invokes the any-age Pi flush. One sandbox reload lifecycle blocker remains inline. All GitHub checks are green at this head; no local build or test was run.

src/session/instance/session_id.rs
```diff
@@ -347,6 +369,216 @@ impl Instance {
...
+    /// hook directory for a host pane.
+    ///
+    /// This is the reload path. `pi_extension_launched` is runtime state, so a
+    /// daemon or TUI that reloads a still-live session has only the file to go
+    /// on, and looking in the wrong place makes a publishing pane read as a
+    /// silent one: no poller repair, and a final flush that returns early.
+    fn pi_sidecar_exists(&self) -> bool {
+        match self.pi_sandbox_sidecar_dir() {
+            Some(dir) => dir.join("session_id").is_file(),
+            None => crate::hooks::session_id_sidecar_exists(&self.id),
```
jerome-benoit: **[P2] Fail closed when the sandbox sidecar path cannot be resolved**

`None` here is ambiguous: it means either a host pane or a sandboxed pane whose bind-backed path could not be resolved, for example because instance-id validation or home-directory resolution failed. In the latter case this falls back to the host hook directory; poller creation then also passes `None` to `pi_sidecar_poll_fn`, whose `None` arm reads that host sidecar. A sandbox can therefore adopt a host conversation from the wrong namespace. Please branch on `self.is_sandboxed()` first and return `false` when a sandbox path cannot be resolved, matching `pi_published_session_id()`'s fail-closed behavior.

jerome-benoit (changes requested):
Re-review of exact head `ab54a955`. The previous sandbox-reload blocker is fixed for the normal bind-backed path, and all current CI checks are green. One fail-open namespace edge remains inline: a sandbox path-resolution failure is treated as a host session, which can reintroduce cross-source attribution. No local build or test was run.

jerome-benoit (approved):
Re-review of exact head `cd520c72`. The 12 previously reviewed commits are patch-equivalent after the rebase. The two new commits correctly isolate the affected `HOME`-mutating tests and replace the ambiguous optional sandbox path with an explicit `PiSidecarSource`: an unresolved sandbox source now fails closed across sidecar detection, ID/path reads, and poller creation, with no host fallback. The previous blocker is resolved, all review threads are closed, and all current CI checks are green. No local build or test was run.

Issue #3553, opened 2026-08-27
aoe-agent is the default ACP agent but nothing appears to build or install it, and `aoe acp agents` reports it present regardless


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

`aoe-agent` is the documented default ACP agent and the wizard's multi-provider fallback, but nothing in the repo appears to build or install it, and `aoe acp agents` reports it present regardless.

Surfaced as an aside during review of #3533 (agent review on @njbrake's behalf) and verified separately.

## Evidence

**Nothing produces the registry's command path.** `src/acp/agent_registry.rs:189` spawns `aoe-agent` as:

```rust
command: "${aoe_data_dir}/acp-worker/dist/aoe-agent".into(),
```

No step anywhere writes that path: not `build.rs`, not `xtask`, not `.github/workflows/release.yml`, not `flake.nix`. The only code that writes into that directory is `src/acp/adapters.rs:92`:

```rust
app_dir.join("acp-worker").join("adapters")
```

which owns `acp-worker/adapters`, the third-party adapters `aoe acp doctor --fix` installs via npm. It never touches `dist`. `acp-worker/aoe-agent/package.json` has a `build` script (`tsc`, `outDir: dist`), but nothing in the repo invokes it, and its output would land in `acp-worker/aoe-agent/dist/` in the source tree rather than in the app dir.

**`aoe acp agents` reports `[OK]` unconditionally.** `command_present` at `src/cli/acp.rs:763`:

```rust
if command.contains("${") {
    true
}
```

The comment explains the intent (a literal `stat` of a path containing `${...}` would report missing for every placeholder-based agent), but the effect is that `aoe-agent` prints `[OK]` whether or not the binary exists. `resolve_agent_command` is honest about it, and `src/acp/acp_client.rs:13391 resolve_agent_command_returns_none_for_placeholder` pins that it returns `None` for exactly this path, so the two surfaces disagree.

**Not present on a real machine.** Neither `~/.agent-of-empires/acp-worker/` nor `~/.agent-of-empires-dev/acp-worker/` exists here, on a host that has run release and dev builds.

**Docs claim it ships.** `docs/structured-view.md:31` lists `aoe-agent` as "bundled (Vercel AI SDK 6) | ships with `aoe`", and `docs/development/internals/structured-view.md:114` shows `default_agent = "aoe-agent"` as the `[acp]` example.

[...]

PR #3583, opened 2026-08-29
fix(acp): default the structured-view agent to claude-code

njbrake:
## Description

`acp.default_agent` defaulted to `aoe-agent`, which nothing builds or installs (#3553): the release ships only the `aoe` binary, `adapters.rs` embeds just the three npm adapters, and `acp-worker/.gitignore` ignores `dist/`.

The setting also had no runtime reader. `Supervisor::pick_agent_for_tool` and its sync mirror `pick_acp_agent_name` both hard-coded `aoe-agent` as the non-claude fallback, so flipping the serde default alone would not have changed any spawn.

This flips the default to `claude-code`, threads the resolved setting through both selectors, and corrects the docs and `--agent` help that claim `aoe-agent` ships with the binary.

Wiring the setting needs a migration: v005 seeded `default_agent = "aoe-agent"` into existing installs' configs (renamed to `[acp]` by v012), and a persisted value outranks a serde default. v026 repoints only that seeded value in the global config. Profile configs are sparse, so a value there is a real choice and is left alone; `[acp]` is not repo-overridable.

Packaging and the `command_present` fix stay on #3553.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the selector is covered by a table-driven unit test on `pick_agent_for_tool`, and the migration by unit tests in `v026`. Both are cheaper than a full-binary test and catch the regression.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Fable 5 and Claude Opus 5 via Claude Code

The packaging investigation and the original default flip were drafted by Claude Fable 5. The selector wiring, migration, and this description were drafted by Claude Opus 5 during review. The reasoning and decisions are @njbrake's.

_Note: the code and this description were drafted by Claude Fable 5 and Claude Opus 5 via back-and-forth with @njbrake using the Claude Code CLI. The reasoning and decisions are his; the prose is the models'._

## Summary

- Changed the ACP structured-view default agent from `aoe-agent` to `claude-code`.
- Updated agent selection to use the configured `acp.default_agent`.
- Added migration v026 for existing global configurations using `aoe-agent`.
- Preserved explicit profile settings and handled blank defaults safely.
- Corrected ACP capability checks before enabling structured view.
- Updated CLI help, documentation, and troubleshooting guidance.
- Added tests for fallback behavior, capability checks, and migration safety.

## Benefits

Users receive a working default agent without manual configuration. Existing configurations migrate safely, explicit agent choices remain unchanged, and terminal-only sessions are no longer treated as ACP-capable.

### DIFF
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -423,7 +423,8 @@ pub struct AcpConfig {
     )]
     pub offer_structured_in_new_session: bool,
     /// Acp agent used when --agent is not specified (e.g. claude-code,
-    /// codex, aoe-agent).
+    /// codex). Must name an agent that can start: `aoe-agent` is not
+    /// packaged yet (#3553).
     #[serde(default = "default_agent")]
     #[setting(label = "Default agent", widget = "text", validate = "nonempty")]
     pub default_agent: String,
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -422,8 +422,8 @@ pub struct AcpConfig {
         widget = "toggle"
     )]
     pub offer_structured_in_new_session: bool,
-    /// Acp agent used when --agent is not specified (e.g. aoe-agent,
-    /// claude-code, gemini).
+    /// Acp agent used when --agent is not specified (e.g. claude-code,
+    /// codex, aoe-agent).
     #[serde(default = "default_agent")]
     #[setting(label = "Default agent", widget = "text", validate = "nonempty")]
     pub default_agent: String,
@@ -646,7 +646,7 @@ impl Default for AcpConfig {
 }
 
 fn default_agent() -> String {
-    "aoe-agent".to_string()
+    "claude-code".to_string()
 }
 fn default_max_workers() -> u32 {
     100
--- a/src/session/config.rs
+++ b/src/session/config.rs
@@ -645,8 +645,12 @@ impl Default for AcpConfig {
     }
 }
 
+/// Built-in `acp.default_agent`. `aoe-agent` is not packaged yet (#3553), so
+/// the default has to be an adapter `aoe acp doctor --fix` can install.
+pub const DEFAULT_ACP_AGENT: &str = "claude-code";
+
 fn default_agent() -> String {
-    "claude-code".to_string()
+    DEFAULT_ACP_AGENT.to_string()
 }
 fn default_max_workers() -> u32 {
     100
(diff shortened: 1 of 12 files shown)


src/session/config.rs
```diff
@@ -646,7 +646,7 @@ impl Default for AcpConfig {
 }
 
 fn default_agent() -> String {
-    "aoe-agent".to_string()
+    "claude-code".to_string()
```
jerome-benoit: **[P1] Make the configured default participate in agent selection**

Changing this serde default cannot alter any ACP worker launch. At this head, `default_agent` has no runtime reader in the create/spawn paths: `/acp/spawn` derives the override only from the request or stored `Instance.agent_name`, then `pick_agent_for_tool` still returns `aoe-agent` for its unmatched non-Claude fallback. The CLI mirror retains the same hard-coded fallback. Therefore every path that selected the missing `aoe-agent` before this PR still selects it afterward; this change only affects settings serialization/display.

Thread the resolved `AcpConfig.default_agent` through the actual selection paths, or replace their real fallbacks consistently, and cover the selector behavior with one table-driven test.

jerome-benoit (changes requested):
Blocking issue: the changed serde default is not connected to the ACP worker selection paths, which still retain the aoe-agent fallback. I did not duplicate the existing CodeRabbit documentation thread.

jerome-benoit (changes requested):
The original selector-wiring blocker is fixed, but two functional inconsistencies remain:

1. `acp_enable` validates the post-fallback agent rather than the session tool or explicit agent key. With this PR, an unregistered terminal-only tool falls back to `claude-code`, passes the registry check, and the endpoint proceeds to kill the tmux session and switch it to structured view. Please gate this transition on the pre-fallback ACP capability key, consistently with the create and TUI paths, and resolve the configured default only after that check passes.

2. The explicit-adapter error path still recommends `aoe-agent` as a "bundled fallback". This contradicts the premise of this PR: `aoe-agent` is not packaged, and following that hint can create a structured session that later fails at worker spawn. Please remove that suggestion or point users to an installed/doctor-fixable adapter.

jerome-benoit (approved):
Re-review of exact head `54edbe3d70c1a8f2c386fbd38a2e2b4bc1187024`.

Both previous blockers are resolved. `acp_enable` now validates the pre-fallback capability key before resolving the configured default agent and before any tmux teardown. The explicit-adapter error path no longer recommends the unpackaged `aoe-agent` fallback. Independent cross-validation found no remaining functional issue.

Non-blocking documentation debt: a few internal comments still describe `aoe-agent` as the fixed or bundled fallback (`src/acp/supervisor.rs:508`, `src/session/builder.rs:1048`, `src/cli/acp.rs:771`), and `src/session/instance.rs:1051` still says the model is forwarded specifically to `aoe-agent`. These should be generalized to the configured or selected ACP agent.

The full CI matrix is green at this SHA. No local build or test was run.

PR #3584, opened 2026-08-29
fix: preserve utf-8 names when stopping sessions

### LINKED ISSUE #3548: stop_all_sessions reports killed sessions it never matched when the environment has no UTF-8 locale
`stop_all_sessions()` and `get_current_session_name()` read session names through `tmux_command()`, which does not force UTF-8. When aoe's own environment has no UTF-8 in `LC_ALL` / `LC_CTYPE` / `LANG` and `TMUX` is unset (a systemd unit with a scrubbed environment, or a minimal container), tmux sanitizes each wide character to one `_` per display column. Same mechanism as #3536, just a different call site.

`stop_all_sessions()` (src/tmux/mod.rs:872) is the one that hurts. `SESSION_PREFIX` is `aoe_`, pure ASCII, so a mangled name still passes `is_aoe_session`:

```rust
if is_aoe_session(line) {
    if let Some(pid) = crate::process::get_pane_pid(line) {
        crate::process::kill_process_tree(pid);
    }
    let _ = tmux_command().args(["kill-session", "-t", line]).output();
    killed += 1;
}
```

So with a mangled name:

- `get_pane_pid(line)` returns `None`, and the agent process tree is never killed
- `kill-session -t <mangled>` fails into the `let _ =`
- `killed += 1` runs anyway

The panic button reports "stopped N" while every agent keeps running.

`get_current_session_name()` (src/tmux/mod.rs:1354) has the same read but a smaller blast radius.

Not a regression from #3540, which only narrowed the query tier. Raised there by @njbrake as a follow-up.

xianjianlf2:
## Description
Use the UTF-8-safe tmux query command when listing sessions for the panic button and when reading the current session name. The stop count now includes only successful `kill-session` calls, and a focused unit test covers Unicode names, ownership filtering, and failed kills.

Closes #3548

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] Documentation was updated where necessary (N/A: no public workflow changed)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the regression is covered by the focused pure unit test and existing `tmux_query_command` environment test; tmux is unavailable on this development host.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you would like to share:** Used for codebase navigation, implementation, and test execution.

## Testing

- `cargo fmt --check`
- `cargo test stop_aoe_sessions_counts_only_successful_kills`
- `cargo test tmux_query_command_preserves_ctype_for_session_names`
- `cargo clippy --lib -- -D warnings`
- `cargo test --lib` reached 4,751 passing / 2 ignored; three unrelated tmux-dependent status tests failed because tmux is not installed on this host.
- `cargo clippy --all-targets -- -D warnings` is blocked by the pre-existing `items_after_test_module` warning in `src/process/macos.rs`.

## Summary

- Preserves Unicode session names by using UTF-8-safe tmux queries.
- Applies this handling to session stopping, current-session lookup, terminal sessions, tool sessions, and instance sessions.
- Counts only sessions that stop successfully.
- Refreshes the session cache after matching AOE sessions, including race conditions.
- Adds tests for Unicode names, ownership filtering, and failed session kills.

## Benefits

This prevents incorrect session matching and false success counts in environments without a UTF-8 locale. It also improves cleanup reliability during teardown races.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -877,22 +877,21 @@ pub fn stop_all_sessions() -> anyhow::Result<usize> {
         .output()
         .map_err(|e| anyhow::anyhow!("tmux list-sessions spawn failed: {e}"))?;
 
+    let mut matched = false;
     let killed = if output.status.success() {
         let stdout = String::from_utf8_lossy(&output.stdout);
         stop_aoe_sessions(stdout.lines(), |name| {
+            matched = true;
             if let Some(pid) = crate::process::get_pane_pid(name) {
                 crate::process::kill_process_tree(pid);
             }
-            tmux_command()
-                .args(["kill-session", "-t", name])
-                .output()
-                .is_ok_and(|result| result.status.success())
+            utils::kill_session_if_present(name).is_ok()
         })
     } else {
         0
     };
 
-    if killed > 0 {
+    if matched {
         refresh_session_cache();
     }
     Ok(killed)
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -872,29 +872,40 @@ fn session_name_from_cache(derived: &str, shape: &NameShape) -> Option<String> {
 /// for a panic button with a handful of sessions; if counts grow, batch the
 /// SIGTERM across all pids, wait once, then SIGKILL survivors.
 pub fn stop_all_sessions() -> anyhow::Result<usize> {
-    let output = tmux_command()
+    let output = tmux_query_command()
         .args(["list-sessions", "-F", "#{session_name}"])
         .output()
         .map_err(|e| anyhow::anyhow!("tmux list-sessions spawn failed: {e}"))?;
 
-    let mut killed = 0;
-    if output.status.success() {
+    let killed = if output.status.success() {
         let stdout = String::from_utf8_lossy(&output.stdout);
-        for line in stdout.lines() {
-            if is_aoe_session(line) {
-                if let Some(pid) = crate::process::get_pane_pid(line) {
-                    crate::process::kill_process_tree(pid);
-                }
-                let _ = tmux_command().args(["kill-session", "-t", line]).output();
-                killed += 1;
+        stop_aoe_sessions(stdout.lines(), |name| {
+            if let Some(pid) = crate::process::get_pane_pid(name) {
+                crate::process::kill_process_tree(pid);
             }
-        }
-    }
+            tmux_command()
+                .args(["kill-session", "-t", name])
+                .output()
+                .is_ok_and(|result| result.status.success())
+        })
+    } else {
+        0
+    };
 
     if killed > 0 {
         refresh_session_cache();
     }
     Ok(killed)
+}
+
+fn stop_aoe_sessions<'a>(
+    names: impl Iterator<Item = &'a str>,
+    mut stop: impl FnMut(&str) -> bool,
+) -> usize {
+    names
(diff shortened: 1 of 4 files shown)


src/tmux/mod.rs
```diff
@@ -872,31 +872,42 @@ fn session_name_from_cache(derived: &str, shape: &NameShape) -> Option<String> {
 /// for a panic button with a handful of sessions; if counts grow, batch the
 /// SIGTERM across all pids, wait once, then SIGKILL survivors.
 pub fn stop_all_sessions() -> anyhow::Result<usize> {
-    let output = tmux_command()
+    let output = tmux_query_command()
```
jerome-benoit, line 875: **[P2] Convert the remaining session-name scans to the UTF-8 query helper**

The same locale-dependent failure remains in `kill_all_tool_sessions_for_id`, `kill_all_terminals_for_id`, and `tmux_env_session_name_for_instance_id`: each still reads `#{session_name}` through `tmux_command()`. Tool and paired-terminal names include the sanitized title, and the sanitizer preserves alphanumeric Unicode. Under `LC_ALL=C`, a tool session named `aoe_tool_git_会话_abc12345` is returned as `aoe_tool_git______abc12345`; killing that returned name fails, while listing with `-u` returns the real name and the kill succeeds. Consequently per-instance teardown can still orphan ancillary sessions under the exact conditions fixed here. Please route every programmatic session-name scan through `tmux_query_command()`.

src/tmux/mod.rs
```diff
@@ -872,31 +872,42 @@ fn session_name_from_cache(derived: &str, shape: &NameShape) -> Option<String> {
...
+        stop_aoe_sessions(stdout.lines(), |name| {
+            if let Some(pid) = crate::process::get_pane_pid(name) {
+                crate::process::kill_process_tree(pid);
             }
-        }
-    }
+            tmux_command()
+                .args(["kill-session", "-t", name])
+                .output()
+                .is_ok_and(|result| result.status.success())
```
jerome-benoit: **[P2] Treat a session that disappears during process teardown as stopped**

`kill_process_tree` runs before `kill-session`, so it can remove a session that does not have `remain-on-exit` before the second command executes. The existing `kill_session_if_present` helper explicitly documents and handles this race by treating missing-session and no-server responses as success. The new `status.success()` check instead counts zero and skips the cache refresh even though the requested end state was achieved. Please reuse that helper, or apply the same absent-is-success classification here, and refresh after a matched session was processed.

jerome-benoit (changes requested):
Two blocking gaps remain: analogous programmatic session-name scans still use the locale-sensitive tmux path, and the new success-only count misses sessions removed by the preceding process-tree teardown. See the inline comments.

jerome-benoit (approved):
Re-review of exact head `e92895e9658cd23c1087d2e527333109ac85864f`. The previously reviewed patch is patch-equivalent after the rebase. The new commit routes the remaining session-name scans through `tmux_query_command()` and reuses `kill_session_if_present()`, so UTF-8 names are preserved and teardown races where the tmux session has already disappeared count as successful. The cache is also refreshed whenever an AoE session was processed.

Both previous blockers are resolved, and independent cross-validation found no new actionable issue. Full CI is still running at the time of this review; no local build or test was run.

PR #3586, opened 2026-08-30
fix(tmux): detect OMP 18.0.10 active turns

jerome-benoit:
## Description

Fix OMP 18.0.10 active-turn detection in tmux panes. OMP moved the live spinner and elapsed timer into its status brand segment while leaving a static interrupt row above the composer, so AoE&#39;s existing manifest rules reported active turns as Idle.

This revision supersedes the original bespoke Rust detector with the manifest architecture now on main:

- adds a high-priority `active_brand` manifest rule for the strict whole-unit timer/status band plus a nearby built-in interrupt row
- accepts the shipped Unicode, Nerd Font, and ASCII variants while excluding persistent clock segments, malformed durations, stale pushed bands, and unrelated indented prose
- leaves composer-only Idle unwitnessed so a Running session requires the normal confirming poll before it settles
- prevents Ask footer prose inside a live composer from overriding the real prompt state
- keeps `detect_omp_status` as the thin manifest dispatcher; no detector DSL extension was required

The branch normally merged the manifest-detector main tip and then refreshed to current organization main `4f6356f7`; no rebase or force-push was used. The ACP queue changes from #3618 remain unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No documentation or changelog edit is needed: this changes internal pane classification, and the root changelog is generated from merged PRs.

## Test Coverage Analysis

- [x] N/A: this PR doesn&#39;t change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [ ] N/A: this PR doesn&#39;t change TUI rendering, a CLI subcommand, or session lifecycle
- [x] Skipped a test, because: the regression is pane-text classification covered by focused unit matrices and an installed OMP 18.0.10 end-to-end smoke through the public CLI surfaces

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** OpenAI Codex / GPT-5.6

**Any Additional AI Details you&#39;d like to share:** Root-cause analysis, implementation, adversarial review, and verification support.

## Verification

- Exact-main historical matrix: 211/261 old captures covered versus 171/261 on main, with zero regressions; retry/error 74/74 and near-miss 12/12
- Inverse mutation: disabling only `active_brand` returns the real OMP 18.0.10 status-band capture to Idle
- `cargo test test_detect_omp_status_ -- --nocapture`: 11 passed
- Latest `cargo test`: 4,809 passed, 2 ignored; only unrelated real-tmux environment case `test_batch_output_frames_records_against_a_real_tmux` failed and reproduced in isolation
- `cargo clippy -- -D warnings`: passed
- `nix develop --command cargo build --features serve`: passed
- `cargo fmt -- --check` and `git diff --check`: passed
- Official OMP 18.0.10 binary smoke: Idle -&gt; Running -&gt; Idle through `aoe session show` and `aoe ps`

## Review Cycles

The original manifest redesign received one fresh double adversarial review cycle. It reproduced and fixed an overly broad custom interrupt-row false positive, then independently validated Unicode, Nerd Font, ASCII, separator, bordered-composer, strict-duration, and persistent-clock cases.

A second requested review/correction cycle used two new targeted prompts:

- architecture/state review traced priority arbitration, positional matching, visibility, hook precedence, and parked-composer transitions
- adversarial regex review required an exact pane, expected/actual status, source evidence, and a minimal correction for every finding
- every report was reproduced before editing production code
- two proposed composer-priority findings were rejected: raising Idle to the live-rule priority regressed real retry/error and legacy loader contracts
- one finding was accepted: OMP&#39;s narrow overflow can drop every trailing status segment, leaving only spinner, whole-unit timer, and border; `active_brand` now recognizes that shipped shape while retaining the interrupt-row and clock exclusions

The accepted case has a permanent regression row. No additional review pass was run in this cycle.

### Latest requested review/correction cycle

One fresh cycle used two new source-grounded prompts:

- exact Rust matcher semantics, nested regions, positional arbitration, end-cap boundaries, and winning-rule calculation
- OMP 18.0.10 source-to-detector completeness across presets, overflow, separators, and adversarial pane shapes

All three reports were independently reproduced and corrected:

- timer-only powerline bands now accept the emitted `` end cap
- Nerd Font composers suppress retained Ask footer evidence like Unicode and ASCII composers
- presets without the `pi` segment use a singleton bottommost built-in `Working` interrupt row as live Running evidence

Each failure has a permanent table case. Focused OMP tests, Clippy, formatting, and the serve build pass. No additional review pass was run in this cycle.

### Maintainer review

The approving maintainer review identified four edge cases. All four were reproduced and corrected:

- a real Ask panel remains Waiting when its composer is still visible; draft footer prose is suppressed only without an Ask border
- composer-only Idle is no longer marked visible, so a live loader missed during redraw cannot bypass the confirming poll
- the bare interrupt fallback requires one bottommost row and rejects duplicated stale rows around completed output
- lowercase duration prose such as `x 30m saved per run` no longer matches the active status band

Each case has a focused regression assertion.


njbrake (approved):
Approving. Four things to take or leave.

**`panel_prompt`'s new `not` clause is the one I'd want settled.** It disables the rule on any composer border in the bottom 8 lines, in either direction. The existing `╭─ Ask ─╮ … Enter select · n note` fixture with `╭── π  > draft ─╮` below it reads Waiting on main and Idle on this branch; same for the `current prompt to answer` guard footer. That footer exists because a composer draft exists (7391acaa7), so the question is whether the composer is on screen in that state. If it is, a session blocked on an ask dialog reports Idle and the user gets no needs-input signal. A capture of that state would settle it either way. `approve_deny_panel` and `plan_panel` are unaffected.

**`composer_idle` changes no verdict.** `detect_omp_status` passes `hook: None` and omp has no `hook_config`, so nothing sits between priority 950 and `idle_by_default`; every pane reaching the rule already resolved to Idle. Its only effect is `visible = true`, which skips the Running→Idle confirming poll in `confirm_detection`. A live turn whose loader scrolls one line past the 3-line window now flips to Idle on the first frame instead of the second, the flap the existing fixture comment calls bounded. Either state that as the rule's purpose or drop it. Note also that `docs/development/adding-agents.md` scopes `visible` to chrome that reads live state, and omp's composer box is on screen mid-turn too.

**`active_interrupt`'s only fixture is a one-line pane** (`"  ⎋ Working…"`). Its guard against a stale interrupt row is untested against a realistic shape: `⎋ Working…` / `Done. Wrote 3 files.` / `esc Working...` reads Running here, Idle on main.

**`active_brand`'s `\S` frame token matches prose.** `x 30m saved per run` under a line opening `esc ` reads Running. Low, since the bottom rows are normally chrome, but it's the loosest part of an otherwise tightly bounded regex.

CodeRabbit's `is_box_bottom` line comment is stale; it targets a helper this revision no longer carries.

Issue #472, opened 2026-03-18
disable trust folder confirmation for YOLO-mode sandboxed sessions


lgmars:
When we start a session in a container in YOLO-mode, `gemini` and `codex` will still ask for trust folder confirmation. It's inconvenient and unintuitive.

for gemini, add to the settings.json
```
{
  "security": {
    "folderTrust": {
      "enabled": false
    }
  }
}
```

for codex, set per project in config.toml
```
[projects."/absolute/path/to/project"]
trust_level = "trusted"
```

PR #3589, opened 2026-08-30
fix(session): stop sessions opening on an agent folder-trust prompt

njbrake:
## Description

Claude Code, Codex, and Gemini each refuse to start in a directory they have not been told to trust, keyed on the resolved directory. Every new worktree and every container workspace is a fresh key, so the session opens on a blocking prompt instead of the agent.

#472 covered Codex and Gemini in sandboxes. Claude took the `_ => Ok(())` arm of that same registry, and host sessions were never covered for any of the three. This fills the claude arm, moves the YOLO decision inside the registry (Claude's dialog is a startup gate, not an approval gate, so it applies in non-YOLO sessions too), drops the create path's duplicate copy of the per-agent match so create and attach cannot drift, and adds `session.pre_trust_agent_folders` (default off) for host sessions.

Host writes land in the user's real config and outlive the session, so that setting stays opt-in and always scopes to one directory. Gemini gets a per-path `trustedFolders.json` entry instead of the global `security.folderTrust.enabled = false` the sandbox path uses, which would disable the feature for all of the user's work.

A repo's `.mcp.json` still asks per server and permission modes are unchanged, but trust is also what activates the repo's own `.claude/settings.json`: an untrusted workspace drops its `permissions.allow` rules, and because the prompt is what holds a session before startup, a pre-trusted workspace runs that file's hooks unprompted. Confirmed against 2.1.251 with a marker-file `SessionStart` hook: untrusted did not run it, pre-trusted did. Both guides say so and the setting's own description carries the warning.

**Rejected alternative:** `CLAUDE_CODE_SANDBOXED=1`. Against Claude Code 2.1.251 it skips the dialog but leaves the repo's `.claude/settings.json` permission rules ignored ("this workspace has not been trusted"), does not auto-approve `.mcp.json` servers, and is baked at `docker run`, so it never reaches an already-created container.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (N/A, the setting renders from the schema)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered by unit tests at the layer the bug lives in. `test_build_container_config_seeds_claude_folder_trust` asserts the seed lands under the in-container working dir in both YOLO and non-YOLO and merges with existing onboarding state; `trust_claude_project` / `trust_gemini_project` / `trust_host_project` have merge, malformed-file, idempotence, per-agent routing, and default-config-path tests, and `ensure_host_folder_trust` has a table test over the setting gate and the host-only gate. An e2e would need a Docker daemon plus a real agent login to observe the prompt.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

The behavioral claims were reproduced against the installed Claude Code 2.1.251 with a pty harness rather than taken from its docs, including the two negative results above. An earlier revision asserted the env var also lifted the `.mcp.json` and project-permission gates; that was wrong and motivated the mechanism change. A second review then caught that the host arm wrote Claude's trust record to `~/.claude/.claude.json` when `CLAUDE_CONFIG_DIR` is unset, where Claude Code never reads it, making the setting a silent no-op on a default install; the container sets that variable, which is what hid it.

🤖 Generated with [Claude Code](https://claude.com/claude-code), running Claude Opus 5 (1M context)

https://claude.ai/code/session_018Aj6BCfub1s1kPHf7ZxcEG

## Summary

- Prevents Claude Code, Codex, and Gemini from opening sessions with folder-trust prompts.
- Pre-trusts sandbox workspaces through each agent’s native configuration.
- Adds optional host-session pre-trust through `session.pre_trust_agent_folders`.
- Preserves per-server `.mcp.json` prompts and existing permission modes.
- Centralizes trust handling for create, attach, and restart flows.
- Adds safe config merging, malformed-file recovery, idempotence, and focused tests.
- Documents folder-trust behavior and configuration.

## Benefits

Sessions start without unnecessary trust prompts while preserving agent security controls. Host-session pre-trust remains opt-in and limited to the session worktree.


src/session/container_config.rs
```diff
@@ -1274,29 +1274,44 @@ fn refresh_codex_sandbox_hooks(
...
             container_workspace_path,
         ),
-        ("gemini", ".gemini") => {
+        ("gemini", ".gemini") if is_yolo_mode => {
             crate::hooks::disable_gemini_folder_trust(&sandbox_dir.join("settings.json"))
         }
+        // The same host file is bind-mounted at both `$CLAUDE_CONFIG_DIR/.claude.json`
+        // (via the config dir) and `~/.claude.json` (via `home_seed_files`), so one
+        // write covers whichever path this Claude Code build reads.
+        ("claude", ".claude") => crate::hooks::trust_claude_project(
```
jerome-benoit, line 1300: **[P1] Keep repository hooks behind explicit workspace consent**

This accepts Claude workspace trust for every sandbox, including non-YOLO sessions with no explicit trust decision. Claude's hooks documentation states that interactive sessions hold back settings-file hooks until workspace trust is accepted, and repository-provided `.claude/settings.json` hooks can execute commands automatically. AoE also copies the agent's authentication state into the sandbox. Please require explicit consent before persisting `hasTrustDialogAccepted`, or otherwise suppress the startup prompt while keeping project settings and hooks untrusted.

src/hooks/mod.rs
```diff
@@ -1219,6 +1219,155 @@ pub fn trust_codex_project(config_path: &Path, project_path: &str) -> Result<()>
...
+    host_env: &[String],
+    project_path: &str,
+) -> Result<()> {
+    let config_dir = |var: &str, default: &str| {
+        resolve_config_dir_override(var, host_env)
+            .map(PathBuf::from)
+            .unwrap_or_else(|| home.join(default))
+    };
+    match agent_name {
+        "claude" => trust_claude_project(
```
jerome-benoit, line 1359: **[P1] Write the default Claude trust entry to `~/.claude.json`**

Without `CLAUDE_CONFIG_DIR`, this resolves to `~/.claude/.claude.json`. The existing `claude_json_path` contract in `src/session/capture.rs` correctly distinguishes the layouts: the file is inside an explicit `CLAUDE_CONFIG_DIR`, but is `~/.claude.json` for the default tree. As written, the advertised host opt-in does not affect normal Claude launches. Please share that resolver and test both layouts.

src/hooks/mod.rs
```diff
@@ -1219,6 +1219,155 @@ pub fn trust_codex_project(config_path: &Path, project_path: &str) -> Result<()>
...
+            &config_dir("CLAUDE_CONFIG_DIR", ".claude").join(".claude.json"),
+            project_path,
+        ),
+        "codex" => trust_codex_project(
+            &config_dir("CODEX_HOME", ".codex").join("config.toml"),
+            project_path,
+        ),
+        "gemini" => trust_gemini_project(
```
jerome-benoit, line 1369: **[P2] Honor `GEMINI_CLI_HOME`**

AoE already treats `GEMINI_CLI_HOME` as Gemini's config root in the capture paths, but this writer always targets `~/.gemini`. A profile or process using a custom Gemini home will still prompt while AoE mutates an unused default tree. Resolve the override from `host_env` and add default, override, and empty-value cases.

src/hooks/mod.rs
```diff
@@ -1219,6 +1219,155 @@ pub fn trust_codex_project(config_path: &Path, project_path: &str) -> Result<()>
...
+/// per-server prompt.
+///
+/// The file doubles as Claude Code's onboarding state, so every other key is
+/// preserved and a malformed file is treated as empty rather than propagated,
+/// mirroring the installers above.
+pub fn trust_claude_project(config_path: &Path, project_path: &str) -> Result<()> {
+    with_config_lock(config_path, "json.lock", || {
+        let mut config: Value = if config_path.exists() {
+            let content = std::fs::read_to_string(config_path)?;
+            serde_json::from_str(&content).unwrap_or_else(|e| {
```
jerome-benoit, line 1241: **[P2] Do not erase malformed user state**

A parse failure is converted to `{}` and later atomically replaces the original file. For the host path this is user-owned state containing sign-in and per-project data, so a partial or temporarily invalid write can become permanent data loss during launch. Please return the parse error and leave the bytes untouched; the caller already treats trust installation as best-effort. The malformed-file test should assert preservation rather than replacement.

jerome-benoit (changes requested):
The CI suite is green, but this implementation crosses the Claude workspace-trust boundary without consent and has three host-config correctness issues. In particular, normal host Claude sessions write the trust entry to the wrong file, Gemini ignores its configured home, and malformed user state is replaced instead of preserved.

Issue #3573, opened 2026-08-28
refactor: split the session/acp/server/tui fix-hotspot files


njbrake:
**Proposal: incrementally split the largest fix-hotspot files along ownership seams, with no behavior change per step.**

Four files absorb a disproportionate share of fixes and are large enough that reviewers cannot hold their invariants in view:

| File | Lines | Fix commits (last 4 months) |
| --- | --- | --- |
| `src/session/instance.rs` | 17,555 | 54 |
| `src/acp/acp_client.rs` | 15,904 | 33 |
| `src/server/mod.rs` | 10,680 | 34 |
| `src/tui/home/mod.rs` | 7,998 | 35 |

This is an amplifier rather than a root cause (the root-cause design work is #2734, making the TUI a thin client of the daemon's session-lifecycle API): concurrent PRs conflict in the same file, review misses cross-cutting invariants, and fixes regress neighboring fixes. Splitting is also a prerequisite for doing #2734 safely, since the direct-storage call sites to migrate are buried in `instance.rs` and `tui/home/mod.rs` (which #2734 measures at 157 fields on `HomeView`).

Suggested approach: extract one cohesive module at a time (pure moves, `pub(crate)` seams, no logic changes), starting with `instance.rs`, and land each extraction as its own PR so blame stays useful. A checklist of target seams can be drafted alongside #2734's call-site inventory, since the two carve along the same lines.

_Note: this issue was drafted by Claude Fable 5 via back-and-forth with @njbrake, based on analysis of the last four months of commit history. The reasoning and decisions are his; the prose is Claude's._

PR #3594, opened 2026-08-30
refactor(server): split server/mod.rs into per-concern modules

njbrake:
## Description

One file held almost everything the web dashboard's server does: bringing the daemon up, deciding who is allowed to connect, serving the web bundle, watching the session files on disk, the loops that poll session status and reap idle sessions, and the tests for all of it. At roughly 10,700 lines nobody could review it whole, and separate pull requests kept colliding in it. This change moves that code, unchanged, into files that each do one job. Nothing behaves differently; only the location changed.

`src/server/mod.rs` is now 71 lines: module declarations and the public surface. Eighteen new sibling modules take one concern each (app state, token, startup, router, access policy, embedded assets, address discovery, disk reload, structured-row repair, disk watches, serve telemetry snapshot, status poll loop, idle reap, sleep inhibit, startup recovery, the ACP event listener, session identity). The inline `test_support` module becomes its own file. Tests move with the code they cover; the two fixtures shared across submodules live in `test_helpers`.

The move is pure. `self::` paths are absolute, private items used across the new boundaries are `pub(super)`, and the names the rest of the crate reaches for by their old `crate::server::` path are re-exported from `mod.rs`. No item's visibility narrowed, none widened past `pub(super)`, and no call site outside the module changed. A token-level comparison of the old file against the union of the new ones loses nothing: the only additions are the per-module `#[cfg(test)] mod tests` wrappers and rustfmt trailing commas. The one behavioral detail worth naming is the router source-slicing helper, which now reads `router.rs` instead of `mod.rs`.

Comments elsewhere that sent a reader to `server/mod.rs` for a specific symbol now name the file that holds it.

One visible detail does change. The TUI serve dialog's log tail renders the last segment of a line's tracing module path, so daemon lines that read `INFO server: ...` now read `INFO startup: ...`, `INFO access: ...`, and so on. Nothing matches on the module path (the dialog's filtering is all on message content) and explicit `target:` strings are unchanged, so this is cosmetic, and arguably more useful than before.

Two follow-ups this PR does not take. `.github/CODEOWNERS` routes `/src/server/acp_*.rs` to the ACP team, which now picks up `acp_events.rs` on its own; `access.rs` and `token.rs` are plausible additions to the security block, but that is a routing decision rather than part of a move. And this commit wants a `.git-blame-ignore-revs` entry, the way #3587's landed separately in #3590.

Part of #3573, which asks for these hotspot files to be split one at a time, each as its own PR. `instance.rs` was #3587; `acp_client.rs` is #3593.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo test --features serve --lib` reports 6481 passing and 3 failing; the same three fail on `main` in this sandbox (they assume a non-root user). `serve_disk_reload_helper_equivalence` and `serve_cityhall_lockdown` are green. `cargo fmt`, `cargo clippy` and `cargo doc --no-deps` are clean.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

No tests were added or removed. Every existing test moved next to the code it covers. Two `coverage-matrix.json` component paths were repointed at the file that now holds the code each spec exercises.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 (1M context), via Claude Code.

The split was scripted rather than hand-edited: a parser cut the original file at top-level item boundaries, a hand-written map assigned each item and each test to a module, and the generator emitted the files. Visibility widening and import pruning were driven from rustc diagnostics.

_Note: this PR description was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Summary

- Split the large server module into 18 focused modules.
- Moved routing, startup, state, access control, authentication, session handling, monitoring, telemetry, and recovery logic into separate files.
- Added shared test support and kept tests with the code they cover.
- Preserved existing `crate::server::` paths and runtime behavior.
- Updated documentation, coverage references, and route-source tooling.

## Benefits

- Makes the server easier to navigate and review.
- Keeps related code and tests together.
- Reduces merge conflicts.
- Makes future changes easier to isolate.

## Validation

- `cargo fmt`, `cargo clippy`, and `cargo doc --no-deps` pass.
- `cargo test --features serve --lib` reports 6,481 passing tests.
- Three failures remain on `main` because they assume a root user.


Seluj78 (approved):
Giving you approval after reading the reasoning, not the code. But the principle is very sound!

PR #3598, opened 2026-08-30
fix(tmux): stop a parallel refresh clobbering a forced session cache

njbrake:
## Description

Fixes the macOS-only `Cargo Test (macos)` failure in `session::instance::status_update::tests::update_status_probes_the_resolved_name_not_the_title` (red on runs 33319514778, 33320567444, 33323320022).

`SessionCacheGuard` forces `SESSION_CACHE` into a known state, but `refresh_session_cache` is not atomic: it checks staleness, forks `list-sessions`, then writes. A parallel test already blocked on that fork lands its write inside the guarded window. `#[serial]` does not prevent this; it only orders serial tests against each other. On a host with no tmux server that write is `data: None`, which `session_existence_from_cache` reads back as `SessionExistence::Unknown`.

That is the observed failure. The second probe got `Unknown` instead of `Absent`, so the instance stayed `Running` inside the grace window instead of latching `Error`:

```
DEBUG session.store: status 'resolve-r2': tmux server unreachable for 958ns
      (< 4s window, ever_confirmed_present=false), retaining status Running
```

A live guard now suppresses the refresh write. The check happens under the same lock `capture` takes, so guard registration and a concurrent write cannot interleave: either the guard registers first and the refresh skips, or the write lands first and the guard captures it as the state to restore.

Non-test builds compile `forced_session_cache_active` to `false`, so production behavior is unchanged.

## Tests

Added `a_forced_cache_snapshot_survives_a_concurrent_refresh`, which reproduces the clobber deterministically (calls `refresh_session_cache` directly inside a guarded window rather than racing). It fails on `main` with the same `Unknown` vs `Present` mismatch seen on macOS, and passes with the fix.

`cargo test --lib` run three times: 4770 pass. `cargo fmt`, `cargo clippy --all-features`, and `cargo check --features serve` clean.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Diagnosis and fix drafted by Claude Opus 5 through back-and-forth with @njbrake.

## Summary

- Prevents concurrent session-cache refreshes from overwriting a forced `SessionCacheGuard` snapshot.
- Coordinates guard operations and cache writes under the same lock.
- Keeps the lock held during guard deregistration to prevent refresh races.
- Adds a deterministic regression test for the concurrency issue.
- Preserves production behavior outside test builds.

## Benefits

- Makes macOS test runs more reliable.
- Keeps forced cache snapshots stable during concurrent refreshes.
- Improves synchronization without changing runtime behavior.

## Potential follow-up

- Monitor related cache tests for similar concurrency issues.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -1154,7 +1154,11 @@ impl SessionCacheGuard {
 #[cfg(test)]
 impl Drop for SessionCacheGuard {
     fn drop(&mut self) {
-        if let Ok(mut cache) = SESSION_CACHE.write() {
+        // Deregistered under the same lock the restore takes, mirroring
+        // `capture`: a refresh arriving mid-drop is either suppressed or lands
+        // on top of the restored snapshot, never dropped on the floor.
+        let mut cache = SESSION_CACHE.write();
+        if let Ok(cache) = cache.as_mut() {
             cache.data = self.prev_data.take();
             cache.time = self.prev_time;
         }
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -39,6 +39,8 @@ pub mod test_support {
 use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 use std::process::{Command, Output, Stdio};
+#[cfg(test)]
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{OnceLock, RwLock};
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


jerome-benoit (approved):
Reviewed at `70d06ea235508824d87c4ff3d53bdb1821e05797`.

The guard registration, refresh-write suppression, restoration, and deregistration are now serialized by the same `SESSION_CACHE` write lock. The follow-up commit also closes the drop-time window identified in the existing review.

The remaining deterministic-interleaving test coverage note is already captured by CodeRabbit and is non-blocking. Required CI checks are green. No additional findings.

PR #3599, opened 2026-08-30
fix(session): seed folder trust in the config dir a custom agent reads

njbrake:
## Description

#3589 seeds folder trust into the config AoE stages for an agent. A custom agent whose wrapper points the CLI at a different config directory (the usual way to run one CLI against two accounts: export `CLAUDE_CONFIG_DIR` and exec) reads a file AoE never writes, so the seed is a silent no-op and the session still opens on the trust dialog. AoE cannot see the redirect: the wrapper sets the variable after AoE has already chosen a file, which is the same ordering `agent_status_hooks` already documents.

Adds `session.agent_config_dir`, a per-agent map from the agent name a session runs to the config directory it actually reads. The value is a host path in both contexts: host sessions write the trust record into the directory itself, sandboxed sessions into its `sandbox` subdirectory, mirroring the `~/.claude` and `~/.claude/sandbox` split AoE already makes for the built-ins. The container has to see that subdirectory through a `sandbox.extra_volumes` mount, and AoE warns when nothing mounts it rather than writing where no agent will look.

Per-agent rather than per-profile because a profile routinely runs both the wrapper and the plain CLI; a `CLAUDE_CONFIG_DIR` entry in `sandbox.environment` would move both to the wrapper's account.

Gemini's sandbox arm turns folder trust off wholesale, which is fine against a staged config that reaches one container. In a directory the user named it would outlive the session in a config they also run on the host, so that arm falls back to the per-path `trustedFolders.json` entry the host path already uses.

The setting is `local_only` (a config dir carries a `settings.json` that can run commands) and repo-denied by the existing session allow-list. Only folder trust reads it; status hooks keep resolving the agent's own env var.

**Rejected alternatives.** Honoring `CLAUDE_CONFIG_DIR` from `sandbox.environment` and mapping it back through `extra_volumes`: no new setting, but the declaration is profile-wide, so it also redirects plain `claude` sessions in that profile to the wrapper's account. Seeding every bind-mounted directory that looks like an agent config: zero configuration, but it writes to real host configs AoE was never pointed at, the elevation #3589 deliberately kept opt-in.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (N/A, the setting renders from the schema)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: covered by unit tests at the layer the bug lives in. `test_build_container_config_seeds_declared_agent_config_dir` runs the create path for a custom agent and asserts the record lands in the declared directory and not in the built-in one; `test_trust_host_project_config_dir_wins_over_env_var` covers the host arm for all three agents; `test_apply_folder_trust_config_gemini_disable_is_staged_only`, `test_is_mounted_into_container_covers_ancestors_only` and `test_agent_config_dir_for_resolves_only_usable_paths` cover the rest. An e2e would need a Docker daemon plus a real second agent login to observe the prompt.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Found from a live reproduction rather than from reading #3589: a sandboxed session running a `CLAUDE_CONFIG_DIR` wrapper prompted for trust, and the two config files on disk showed why. AoE's staged `~/.claude/sandbox/.claude.json` held a bare `hasTrustDialogAccepted` entry for the workspace, while the wrapper's mounted dir held the same workspace with the full per-project state Claude Code writes itself, which is also what confirms the agent reads and writes trust under `$CLAUDE_CONFIG_DIR` rather than at container `~/.claude.json`.

## Summary

- Added `session.agent_config_dir` for custom agent configuration directories.
- Routed folder-trust records to the correct host or sandbox directory.
- Added validation and warnings for invalid paths and missing sandbox mounts.
- Preserved Gemini trust behavior with per-path entries.
- Added safer Pi sandbox sidecar handling and symlink-resistant file replacement.
- Updated settings access, tests, and documentation.

## Benefits

Custom-agent sessions now use the correct folder-trust configuration and avoid repeated trust prompts. Users receive clear warnings when configured paths are invalid or unavailable in a sandbox. Pi session files are updated safely without following symlink redirects.

## Technical Notes

The setting is per-agent, local-only, and unavailable to repository configuration. Sandboxed sessions use the configured directory’s `sandbox` subdirectory through `sandbox.extra_volumes`. Pi sessions with custom configuration directories use store polling instead of the conversation sidecar.


jerome-benoit (changes requested):
Reviewed at 665fccb9b8d8ff5956ce8254844d6e59af49dbe6. CI is green, but the following existing findings must be resolved before approval:

- **[P1 · BLOCKER] Sandbox-controlled symlink can redirect a host write.** The Pi extension installer writes through a predictable path inside a sandbox-writable bind using symlink-following semantics. Use exclusive/no-follow creation and safe atomic replacement.
- **[P2 · BLOCKER] Pi sidecar ignores the effective agent_config_dir.** Custom Pi configurations therefore lose extension discovery and session-ID publication. All host and container sidecar paths must derive from the session effective configuration directory.
- **[P2 · BLOCKER] Concurrent Pi installations share one temporary filename.** A concurrent rename can make another launch silently start without session-ID reporting. Use a unique temporary path per installation.
- **[P2 · BLOCKER] Accepted configuration paths do not match the resolver.** Restrict tilde paths to ~ and ~/..., and use platform-aware absolute-path validation so Windows paths remain supported.

These findings are already covered by the [existing CodeRabbit review](https://github.com/agent-of-empires/agent-of-empires/pull/3599#pullrequestreview-5065578668) and [validation thread](https://github.com/agent-of-empires/agent-of-empires/pull/3599#discussion_r3890139465), so I am not duplicating them inline.

**Verdict: REQUEST_CHANGES.**

src/session/instance/session_id.rs
```diff
@@ -1126,6 +1151,48 @@ mod tests {
...
+    /// A Pi session pointed at a config dir the user named gets no sidecar.
+    /// Both halves of the path (the discovered extension, the published file)
+    /// live in the bind AoE mounts itself; the user's own dir reaches the
+    /// container through their `extra_volumes` entry, at a path AoE cannot
+    /// see, so publishing there would report into a directory the pane in the
+    /// container never opens.
+    #[test]
+    #[serial_test::serial]
+    fn sandboxed_pi_with_its_own_config_dir_declines_the_sidecar() {
+        let _guard = crate::session::test_support::isolate_app_dir();
```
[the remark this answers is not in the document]
njbrake: No change. `isolate_app_dir` already points `HOME` at a fresh tempdir: `test_support::install_env_vars` sets HOME plus XDG_CONFIG_HOME and XDG_DATA_HOME for the guard's lifetime, so the extension install lands in that tempdir's `.pi` and never in a real one.

_Note: this reply was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/config.rs
```diff
@@ -4728,6 +4785,40 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_agent_config_dir_for_resolves_only_usable_paths() {
+        let home = std::path::Path::new("/home/me");
+        let mut config = SessionConfig::default();
+        for (value, expected) in [
+            ("~/.claude-personal", Some("/home/me/.claude-personal")),
+            ("~", Some("/home/me")),
+            ("/opt/claude", Some("/opt/claude")),
```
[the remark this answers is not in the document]
njbrake: No change. The crate does not build on native Windows: `nix` is an unconditional dependency, and `test_support` documents Windows as WSL2 only. The table has assumed POSIX paths since before this PR. `is_resolvable_agent_config_dir` itself calls `Path::is_absolute`, so a native Windows path would resolve if that target were ever added.

_Note: this reply was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/container_config.rs
```diff
@@ -1354,15 +1407,24 @@ pub(crate) fn ensure_folder_trust_config_for_active_agent(
...
+    // below is mounted at the built-in default, which that agent never opens.
+    let agent_config_dir = session_config.agent_config_dir_for(tool, &home);
 
     for mount in AGENT_CONFIG_MOUNTS
         .iter()
         .filter(|m| m.tool_name == config_tool)
     {
-        let sandbox_dir = match sandbox_dir_for(mount, &home, Some(instance_id)) {
+        let sandbox_dir = match agent_config_dir.as_ref() {
+            Some(dir) => sandbox_dir_in(dir, mount, Some(instance_id)),
```
jerome-benoit: **[P1] Keep the declared Codex config at the mounted root**

For a custom Codex wrapper, the documented setup mounts `<agent_config_dir>/sandbox` as the wrapper's `CODEX_HOME`. `sandbox_dir_in` appends `instance_id`, so this writes trust to `<agent_config_dir>/sandbox/<instance_id>/config.toml` while Codex reads `<CODEX_HOME>/config.toml`. Because the ancestor mount satisfies `warn_unless_mounted`, this fails silently and leaves the trust prompt in place. Either do not append the instance ID for declared directories, or point both the mount and `CODEX_HOME` at that same per-instance directory. Please add a custom Codex regression alongside the Claude case.

src/session/container_config.rs
```diff
@@ -1379,6 +1444,7 @@ pub(crate) fn ensure_folder_trust_config_for_active_agent(
                     &sandbox_dir,
                     container_workspace_path,
                     is_yolo_mode,
+                    agent_config_dir.is_none(),
```
jerome-benoit, line 1449: **[P1] Make custom sandbox trust updates no-follow**

When `agent_config_dir` is set, its `sandbox` directory is expected to be mounted read-write into the container. On a later attach, that container can replace the trust file or an ancestor with a symlink before this host-side update. The `trust_*` helpers then perform path-based reads and call `atomic_write`, which deliberately resolves destination symlinks, so the sandbox can redirect AoE's write outside the bind. Use a descriptor-relative, no-follow read/modify/write rooted at `sandbox_dir`, including the lock and ancestor traversal, or fail closed on links. Keep symlink-following behavior only for host-owned configuration.

jerome-benoit (changes requested):
Re-reviewed current head 35c61237366661d89f2e16844b1ed8271344a31f.

The four blockers from my review at 665fccb9 are materially addressed. However, two current-head blockers remain:

- Custom Codex sandbox trust is written below the directory exposed as the wrapper's `CODEX_HOME`, so Codex still reads an untrusted config.
- Trust updates inside a container-writable custom config bind still use symlink-following read/modify/write operations, allowing a reused sandbox to redirect a host write outside that bind.

Live CI is green. I did not run local build, test, or clippy commands under the review resource constraints.

Verdict: REQUEST_CHANGES.

src/session/container_config.rs
njbrake: Fixed in `1b513b00`. A declared directory now stops at `sandbox`, with no per-instance segment for Codex, so the record lands at the level the user's `extra_volumes` entry mounts. Pointing the mount at the per-instance directory was the other option and is not usable: the instance id changes per session, so no static entry can name it.

Regression test added alongside the Claude case (`test_declared_codex_config_dir_trusts_at_the_mounted_level`), asserting the record is at `<dir>/sandbox/config.toml` and that no per-instance level exists. Confirmed it fails against the previous behavior.

_Note: this reply was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/container_config.rs
njbrake: Fixed in `1b513b00` and `9fc9f67d`. Folder-trust writes now carry a `SymlinkPolicy`. The sandbox arm passes `Never`, the host arm keeps `Follow` so a config symlinked into a dotfiles repo still works (#2784, #3186).

Under `Never`: the read ignores anything that is not a regular file, so a planted link's target is never merged into a config the container reads; the write goes through `replace_file_no_follow`, which opens the bind root, creates the temp file `O_EXCL` and renames through that descriptor, replacing a link at the destination instead of writing through it; the lock path is not resolved through a link, and the lock sidecar itself is opened `O_NOFOLLOW`, so a link planted there fails the open and takes the guarded write with it.

Ancestor traversal below the root is covered by the same walk, though for these three files the path below the bind root is a single component. The root is the mount point, which the container cannot replace from inside.

One gap left deliberately: the Codex hook installers write the same staged directory and still use `Follow`, unchanged from before this PR. Same exposure, different call path, and threading the policy through them widens this change further. Happy to file it as a follow-up.

_Note: this reply was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #3600, opened 2026-08-30
refactor(status): decide pane status from a rule table, not a chain

njbrake:
## Description

Sessions sat on `Running` for hours. From `debug.log`, with the panes captured live: one session held a `running` hook write nobody had refreshed for 1h51m, another for 9h55m, and both flipped between Idle and Running every couple of seconds in between. The pane that produced it:

```
✻ Cooked for 1m 58s · done 7:17 PM
                    ✔ Update installed · Restart to update
────────────────────────────────────────────────────────────
❯ it works now, images are coming through
────────────────────────────────────────────────────────────
  ⏵⏵ bypass permissions on (shift+tab to cycle) · ← for a…
```

The turn is over. `reconcile_claude_hook_status` could only disbelieve the write if the pane yielded a clean ready prompt, `claude_line_above_input_box` skipped a fixed list of chrome to find the completion line, and Claude's own update banner was not on that list. So the banner stood in for the status slot, the parked evidence was unreachable, and the write won for as long as the session lived. The flapping came from the same capture being re-read with no confirmation: mid-redraw frames alternate between an empty box and a typed one.

Both are shape-of-the-code problems rather than missing cases. Two thousand eight hundred lines of imperative detection for three agents, with the hook file as an authority the pane could only override through hand-written reconcilers, gives every new case somewhere different to go.

**Detection is now a table.** Each agent's rules live in `src/tmux/detect/manifests/<agent>.toml` as `{id, state, priority, region, matcher}`; the highest-priority match wins. The hook file is a rule like the others (`region = "hook"`), so the ordering that used to live in reconcilers is declared: a blocking prompt outranks a `running` write, a write younger than 30s outranks parked evidence (a turn that has just started still shows the last completion line), and a write older than 15 minutes outranks nothing. Rules name the region they read rather than each detector re-deriving its own window.

Every agent whose pane carries state moves to a manifest: claude, codex, cursor, opencode, omp, vibe, droid, gemini, qwen, copilot, antigravity, hermes and pi. All the hand-written matchers and every reconciler are deleted, and the four hook-only agents share one stub.

**Detection production goes from 2,812 lines to 303.** The engine and thirteen rule files come to 2,250, so production is a net **−259**; what remains in `status_detection.rs` is the fixture corpus and three entry points. The PR's overall +281 is that corpus, which grew rather than shrank: the reconciler tests were rewritten against the detection entry point instead of deleted, and the incident that started this is now a fixture.

Three agents needed the table to learn something before they would fit, and each addition is general rather than agent-shaped:

- **Named markers** and the regions that resolve through them, `after(<name>)` and `above(<name>, n)`. A marker can be the nth occurrence from the bottom (pi's input box is its *second* rule line), can refuse to resolve past a depth (further up, the same glyph run is transcript prose), and can span wrapped lines (codex's interruption banner runs to four).
- **Per-clause regions**, so a rule whose evidence sits in two places can say so. Codex's plan dialog needs its choice rows below the question and its running check below the submit hint; that was three index comparisons and is now two `region` clauses.
- **Positional arbitration**. omp stacks its state markers, so the current one is whichever sits lowest on screen, not whichever ranks highest. Rules marked `positional` at one priority compete by position, ties falling to declaration order, which is what its `consider` chain did.

**Two signals we were not using:**

- **The terminal title.** tmux already carries it, and it costs one more field on a `list-panes` batch we already run. Claude publishes a spinner frame there while a turn streams (`◐`/`◑`, cycling), so it outranks everything. The static `✳` ranks near the bottom: measured against 2.1.251 it also shows during a long tool call, so it is weak evidence of idle, never proof.
- **`#{window_activity}`**, tmux's last-output stamp. A pane that has drawn nothing since the last look cannot have changed, so the capture is skipped and the previous verdict stands. A parked session with no hook file then costs a batched format read per poll rather than a subprocess. Sessions that do have one still capture every poll, since a hook write can change the verdict without the pane drawing anything.

**Confirmation.** A change no rule read off the agent's own chrome waits for a second agreeing poll, and only running-to-idle, which is the direction a mid-redraw frame produces. Everything else publishes on sight, so a turn starting or a prompt appearing is never a poll late.

**Prior art.** The manifest model, the region vocabulary, the title priority asymmetry and the confirmation hold follow [herdr](https://github.com/herdrdev/herdr) (Apache-2.0), which carries this design across 22 agents; its transcript-viewer and model-picker rules are ported directly. Where we diverge: herdr's working authority is live PTY byte activity, which a tmux poller cannot see, and it does not treat Claude's hooks as authoritative at all (`full_lifecycle_hook_authority` lists six agents, Claude not among them). Keeping hooks as ranked evidence rather than dropping them is deliberate: they are the fastest `waiting` signal, they cover sessions whose title tmux reports as the hostname, and the same mechanism captures the agent's session id, which is what makes resume work.

**Verified live**, against every pane on the author's machine with a two-hour-old `running` write injected: the two stuck sessions resolve `Idle` on `completed_turn`, an actively streaming one reads `Running`, and a session parked on the folder-trust dialog reads `Waiting` where the old detector read `Idle`, because that dialog's option rows carry no numbers and the old matcher required them.

**Three behaviours changed on purpose**, both where the old code's reach was an artefact of where a check lived rather than of what it knew. A live prompt on screen now outranks an `idle` hook write for codex, as it already did for claude: the reconciler only consulted the pane for a `running` write, so an idle one bypassed the screen entirely. And codex's generic approval phrases (`(y/n)`, `run command?`) rank below a `running` write from the last thirty seconds, since they appear in ordinary turn output, which is why the reconciler ignored them; the ranking says the same thing without a second code path. And a session with configured `[[agents.<name>.status_rules]]` now has those rules decide even when a hook file exists; previously a present hook meant they were never consulted.

`aoe session capture` loses its own copy of the codex reconciliation and runs the same table as the poller.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (N/A, no rendering change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the regression is a pure function of one capture plus one hook observation, and the whole existing fixture corpus already exercises it: every `detect_claude_status` and `detect_cursor_status` test now runs through the manifest unchanged, and the 35 reconciler tests were rewritten to call the detection entry point rather than deleted. `test_claude_stuck_running_pane_recovers` is the captured pane from this incident. The poller half (confirmation hold, activity gate) has unit tests in `status_update.rs`. An e2e would need a real agent driven to a silent tool stop to reproduce what a fixture already pins.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Started from `debug.log` rather than from the code: the `session.status_change` trail added in #3489 carries the hook age and pane markers at each flip, which is what identified stale writes and mid-redraw flapping as two separate faults before any code was read. The fixture in `test_claude_stuck_running_pane_recovers` is the real pane, with the typed text replaced. Every claim about which glyph Claude puts in its title, and about how a parked pane still redraws, was measured on live panes rather than assumed.

## Summary

- Replaced custom status checks with TOML manifests for 13 agents.
- Added shared rules for priorities, regions, markers, terminal titles, hook freshness, and activity signals.
- Added safeguards for stale `Running` states, redraw flapping, and unchanged panes.
- Updated `aoe session capture` to use the same detection path as session polling.
- Removed agent-specific matchers and Codex reconciliation logic.
- Added manifests, fixtures, tests, and documentation for the new detection flow.

## Benefits

- Centralizes detection in clear, reviewable rule tables.
- Allows agent updates without Rust code changes.
- Keeps polling and manual capture behavior consistent.
- Improves handling of stale and ambiguous pane content.
- Reduces detection code from 2,812 to 303 lines.

## Potential follow-up

- Add manifest validation tooling or CI checks.
- Continue expanding fixtures as agent terminal interfaces change.


njbrake (commented):
Reviewed independently; two blocking issues found, both fixed in ecadf282.

`hook_waiting` outranked every parked-idle rule, so a `waiting` write that
nothing clears (an Esc-cancelled prompt fires no clearing hook, #2937) pinned
claude, codex, cursor, qwen and gemini yellow until the next submit. The write
now speaks only for an empty capture, the contract `reconcile_waiting_hook` had
before this branch retired it.

The unchanged-screen guard skipped the poll that would have confirmed a held
Idle, leaving hookless sessions on Running until new output arrived. It is now a
named function with a test per condition.

Two findings deferred, neither a regression from this branch: `status_rules` are
looked up by the `agent_detect_as` alias rather than the session's tool, so a
custom agent with its own rules loses them; and `#{window_activity}` is
second-resolution, so a redraw inside the same second is invisible to the
capture guard.

_Note: this comment was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #2522, opened 2026-06-28
Trash: relocate a session's worktree out of the active dirs when trashed, restore moves it back (keep structured-view preview working)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Trashing a session leaves its git worktree exactly where it was created. The worktree base directory (e.g. `<repo>-worktrees/`) keeps filling up with checkouts for sessions the user already sent to the trash, sometimes for the full `trash_retention_days` window (default keeps them around, `0` keeps them forever). The trash is supposed to be "out of the way," but on disk the workspace clutter is identical to a live session.

The ask: when a session moves to the trash, also move its worktree directory into a separate trash area so it stops polluting the active worktree dirs. Restore moves it back. This must keep working for the structured view, because we allow previewing trashed sessions and the preview reads files from the session's working directory.

## What's there today

`Instance::trash()` is intentionally additive: it only stamps `trashed_at` and touches nothing on disk.

`src/session/instance.rs:1335`
```rust
pub fn trash(&mut self) {
    if self.trashed_at.is_none() {
        self.trashed_at = Some(Utc::now());
    }
}
```

The doc comment on it confirms the worktree is deliberately preserved in place: "keeps every durable artifact so `untrash` can bring it back intact. Intentionally additive: only `trashed_at` is set."

The session's worktree directory is its `project_path` (`src/session/instance.rs:387`). All three trash entrypoints set `trashed_at` and stop the agent, but never move the directory:

- TUI: `trash_session_by_id` (`src/tui/home/operations.rs:1450`) calls `inst.trash()`, then `kill_all_tmux_sessions()`.
- Web: `trash_session` (`src/server/api/sessions.rs:2177`) calls `inst.trash()` and shuts the structured-view supervisor down.
- CLI: `aoe rm` routes through `stored.trash()` (`src/cli/remove.rs:82`) when `delete_to_trash` is on.

The structured-view preview of a trashed session reads files from `project_path`. The ACP worker's `cwd` is resolved from `project_path` (`src/acp/acp_client.rs:1648`, container path mapping at `:1659`), so wherever `project_path` points is what preview shows.

[...]

Issue #2002, opened 2026-06-05
Worktree: reconcile a stale session project_path when the directory is moved outside aoe


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

aoe tracks each worktree session by the `project_path` it recorded at creation. It never watches the filesystem or syncs from `git worktree list`. If someone moves or renames the worktree directory outside aoe (a plain `mv`, or `git worktree move` from another shell), the stored `project_path` goes stale and aoe does not detect or repair it. Attach, status detection, and the tied rename all operate on the now-wrong path until the user manually fixes it.

Raised by @njbrake on #1997 (the tie-title-and-worktree-dir PR): "what happens if someone outside aoe manually changes the worktree dir/name?"

The current answer is "nothing good, but nothing destructive": operations on that session fail rather than corrupting anything. This issue tracks adding self-healing.

## What's there today

The tied rename (and the standalone workdir edit from #1723) guards against acting on a missing directory: `edit_worktree_workdir` checks the recorded path exists before any git move.

`src/session/worktree_edit.rs:130`
```rust
if !req.current_path.exists() {
    return Err(WorktreeEditError::SourceMissing(
        req.current_path.to_path_buf(),
    ));
}
```

The server maps that to `409 "The worktree directory no longer exists on disk"`; the CLI bails with the same message. So an external move produces a clear error, not a botched `git worktree move`.

What is missing is any reconciliation. There is no code path that, on load or on the failure, consults `git worktree list` to find where the directory actually went and updates `project_path`. git already knows the new location: `list_worktrees` returns it keyed by branch.

`src/git/worktree.rs:59`
```rust
pub struct WorktreeEntry {
    pub path: PathBuf,
    pub branch: Option<String>,
    pub is_detached: bool,
}
```

`src/git/worktree.rs:869` `pub fn list_worktrees(&self) -> Result<Vec<WorktreeEntry>>`.

A managed session carries the branch in `worktree_info.branch` and the main repo in `worktree_info.main_repo_path`, so a lookup of "which worktree currently checks out this branch" is available without any new bookkeeping.

## Proposal

[...]

PR #3615, opened 2026-08-31
perf(tui): move trashed-row reconciliation off the first-frame path

### LINKED ISSUE #3611: TUI startup: trashed-row reconciliation runs unbounded external work before the first frame
`HomeView::new` reconciles every trashed row serially before the first paint ([`src/tui/home/lifecycle.rs:29`](src/tui/home/lifecycle.rs)). The cost is per trashed row and unconditional, even when nothing changes: `Storage::open_unwatched`, a lifecycle flock, and two `Storage::update` calls, each a full read-parse-serialize-write of the profile's `sessions.json` ([`src/session/storage.rs:1005`](src/session/storage.rs)). Managed worktrees add a container remove plus up to three git spawns via `reconcile_trashed_location` ([`src/session/trash.rs:448`](src/session/trash.rs)).

First-frame time by profile on one host, 3 runs each, idle machine:

| profile | rows | trashed | first frame |
|---|---|---|---|
| personal | 11 | 0 | 0.15s |
| mzai | 17 | 2 | 0.30s |
| aoe | 18 | 12 | 0.41s |
| all | 46 | 14 | 0.64s |

Alt-screen enters at 0.05s to 0.10s in every case, so the cost sits entirely in `HomeView::new`. It scales with trashed rows, not row count.

The path is spawn-latency-bound, so it degrades non-linearly. On the same host with `syspolicyd` saturated (17 `which` spawns costing 1.5s against a 70ms baseline), first frame went to 13.9s, 15.2s and 17.5s across three runs. #3554 reported an unusable TUI from a user with >300 trashed sessions.

Second defect in the same path: a relocation that can never succeed has no terminal state. An orphaned managed worktree (directory present, git admin dir pruned) fails `git worktree move` with `fatal: not a working tree`, persists nothing, and retries on every launch and every trash-poller tick. Observed firing every 27s indefinitely until the row was purged.

Asks:

1. Move reconciliation off the pre-first-frame path. It is healing work, not render input.
2. Batch the per-row storage writes into one `Storage::update` per profile.
3. Give a permanently-failing relocation a terminal state so it stops retrying.

Related: #2754 (same `Storage::update`-on-the-async-path mechanism, scoped to passive-status persist), #2522 (introduced relocate-on-trash), #3554.

njbrake:
## Description

When you launched the TUI, it did a round of repair work on your trashed sessions before it would draw anything. For each trashed session it opened the session store, took a lock, rewrote the store twice, and sometimes ran git to move a leftover folder around. It did all of that every single time, even when nothing was actually broken, so the more trashed sessions you had, the longer you stared at a blank terminal. People with a lot of trash were waiting seconds; someone with over three hundred trashed sessions found the app effectively unusable. On top of that, one particular broken folder could never be repaired, and the app kept trying to repair it on every launch forever. Now the repair work happens in the background after the screen is already up, it first checks cheaply whether there is anything to fix and does nothing if there is not, it fixes several sessions in one write instead of two writes each, and a repair that can never succeed is recorded once so it stops being retried. The app paints immediately and updates itself if a repair changes something.

Three asks from #3611:

**1. Reconciliation off the pre-first-frame path.** New `ReconcilePoller` runs both load-time healing sweeps (trashed-worktree relocation #2522, worktree paths moved outside aoe #2002) on a worker. `HomeView::new` kicks it after the view is built; `apply_reconcile_results` reloads storage when a sweep repairs something.

**2. Batched writes.** `reconcile_trashed_location` splits into `plan_trashed_reconcile` (decided from recorded paths plus the filesystem) and an apply step. A profile whose trash is already consistent now costs one `sessions.json` read and no lock, git spawn, or write. Rows that need work reserve, move, and commit in batches of 8, and only one lifecycle flock is held at a time so no row waits behind another row's git call. The `HomeView::new` expired-reservation clear is one `Storage::update` per profile instead of one per row, and the daemon's `reconcile_trashed_transition` takes the same pre-filter.

**3. Terminal state.** Derived from the filesystem, not recorded on the row. A linked worktree carries a `.git` file naming its admin dir under the main repo; once that admin dir is pruned the checkout is stranded and `git worktree move` can only ever answer "not a working tree". `is_orphaned_checkout` follows that link, and the plan step treats such a row as terminal, so the retry loop stops. No schema change, a repaired repo becomes relocatable again on its own, and no transient git failure can be frozen into `sessions.json`.

An earlier revision of this PR persisted the verdict on the row instead. Review caught that it classified *every* `git worktree move` error as permanent, including the 30s timeout and the failed spawn, which are likeliest on exactly the spawn-latency-bound machines #3611 describes; a row marked that way would never relocate again, and the marker outlived a revert. Note also that the predicate has to follow the `gitdir:` link rather than test for `.git`: in the shape #3611 reports the file survives and only its target is gone.

One behavior trade: a row needing repair can show its stale path on the first frame until the sweep lands and the reload publishes it.

Four behaviours were added during review, none of them in the original three asks, and they are where most of the risk now sits:

- **Startup-recovery gate.** Recovery launches from `project_path` and records each attempt in a boot-scoped ledger it never retries, so running it before the sweep could spend a row's only attempt on a path the sweep was about to fix. `HomeView::new` arms a gate; `apply_reconcile_results` releases it once the sweep lands.
- **Gate deadline.** `Storage::update` blocks on a contended profile flock with no timeout, so a stuck sweep could hold the gate for the whole boot. It opens after 30s regardless, and the repair is applied before it opens.
- **Revalidation before reserving.** The plan comes from an unlocked snapshot, so the batch compares and sets on the plan-relevant fields before taking a reservation; otherwise a peer restore between scan and reserve pinned a Trash reservation on a live session.
- **Pending-reload latch, throttled.** A failed reload keeps the repair pending and the gate shut so it retries, at the same 5s cadence as the heartbeat reload beside it rather than on every 33ms tick.

Batches are 8 rows, not larger: a batch holds its reservations from the first write to the second and `git worktree move` is bounded at 30s per row, so the batch size sets how long that window can get and has to stay well inside the 10-minute reservation TTL. Each row takes its own lifecycle flock only across its own filesystem work, so the batch size is not a flock-contention multiplier.

Fixes #3611

## Measured

Two binaries from the same tree, a detached worktree at the branch point vs. this branch, each driven through a sized pty. Timings are alt-screen entry vs. first painted frame, the two markers the issue uses. Profile restored to pristine between runs. Store: 300 trashed rows plus 20 live, spread over 12 git repos. Debug builds, so read the ratio rather than the absolute numbers.

| | baseline | this branch |
|---|---|---|
| alt-screen enter | 0.26s to 0.40s | 0.30s to 0.31s |
| **first frame** | **15.2s / 17.2s / 15.2s** | **0.88s / 0.89s / 0.85s** |
| control, 0 trashed rows | 0.44s | 0.37s |

Alt-screen entry is unchanged, so the cost sat entirely after it. The zero-trash control being a wash confirms it scales with trashed rows, not row count. Under load the same baseline store reached 36.0s and 40.7s, which is the non-linear degradation the issue describes.

The unconditional-write claim is visible in `lifecycle_generation`, which only moves when a row takes a reservation. Starting from a store where every row is already consistent and nothing needs repair:

- baseline: launch 1 leaves all 300 rows at generation 1, launch 2 leaves all 300 at 2. Every launch reserves and rewrites every row.
- this branch: `{0: 270, 1: 30}` after the repairing launch, unchanged after a second launch. Rows needing nothing are never reserved.

Driven under tmux against a store with 30 rows deliberately left needing repair, the first frame painted while 286 of 300 were repaired, so the frame arrives before the healing finishes. The background sweep then completed all 300, and the repaired rows carry the right durable state: `project_path` repointed, `pre_trash_project_path` restored, reservation released.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Documentation: none needed. The new field is not on the API surface and `docs/` has no page describing trash reconciliation. No screenshot: nothing about the rendered UI changes, only when the first frame appears.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: an e2e test cannot assert the property that matters here without being a timing test, and the behaviours that regressed in review are ordering and error paths an e2e run would not reach deterministically. Unit tests cover them instead, each one checked to fail with its fix reverted before it was trusted: `profile_sweep_leaves_a_consistent_profile_untouched` (a consistent profile takes no reservation, asserted on `lifecycle_generation`), `profile_sweep_heals_every_row_that_needs_it`, `trashed_row_healing_lands_through_the_reconcile_poller` (the repair reaches the view through `apply_reconcile_results`, not `HomeView::new`), `reconcile_never_retries_a_checkout_the_repo_no_longer_registers` and `a_relative_gitdir_link_is_not_mistaken_for_a_stranded_checkout` and `a_stat_failure_on_the_admin_dir_is_not_a_stranded_checkout` (the three ways a checkout is or is not stranded), `a_move_failure_over_a_live_checkout_stays_retriable`, `a_default_branch_checkout_is_never_planned_for_relocation`, `a_row_restored_after_the_scan_is_not_reserved` (the scan/reserve race), and five covering the startup-recovery gate: it is armed at construction, released by the sweep, expires on its deadline, survives live-send, and holds shut while a repair is unapplied or its reload is failing.

Checks run: `cargo fmt`, `cargo clippy --all-targets`, `cargo clippy --features serve --all-targets`, `cargo test` (4773 passed, 0 failed), `cargo test --features serve` filtered to trash / poller / home / server-sessions, and the Docs job's `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features serve`. The clippy warnings that remain are pre-existing on main in files this PR does not touch.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 5 via the Claude Code CLI

**Any Additional AI Details you'd like to share:** Implemented from the issue's three asks through back-and-forth with @njbrake. The design decisions (splitting plan from apply so the pre-filter cannot drift, keeping the terminal state narrow to the git move alone, bounding the batch size) were reviewed by him.

## Summary

- Moved trash and worktree reconciliation to a background `ReconcilePoller`.
- Added batched repairs with fewer locks, writes, and repeated lookups.
- Added filesystem-based detection for orphaned managed worktrees.
- Kept transient relocation failures retryable.
- Batched expired-reservation cleanup into one profile update.
- Deferred startup recovery until reconciliation completes.
- Reloaded repaired data after the first frame.
- Added protection for reservation races and active paste operations.
- Added regression tests for repairs, retries, idempotency, relative Git links, and recovery ordering.

## Benefits

- The TUI renders its first frame without waiting for reconciliation.
- Startup performs less storage and filesystem work.
- Recovery uses repaired worktree paths.
- Permanently failed relocations do not retry indefinitely.
- Relative worktree links resolve correctly.


src/tui/app.rs
```diff
@@ -1857,6 +1857,11 @@ impl App {
                 needs_full_refresh = true;
             }
 
+            if self.home.apply_reconcile_results() {
```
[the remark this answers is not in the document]
njbrake: Fixed, though gated inside `apply_reconcile_results` rather than at the call site. The check has to run before `try_recv_result`, or the short-circuit drains the worker verdict and drops it; keeping it next to the reload it protects also stops the tick loop growing a special case. Covered by `reconcile_reload_waits_for_live_send_to_finish`, which asserts both halves: skipped while live-send is active, still delivered once it ends.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

projects/-tmp--tmp0UGHJO-project/33333333-3333-3333-3333-333333333333.jsonl
```diff
@@ -0,0 +1 @@
+seed
```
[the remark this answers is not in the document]
njbrake: These files should never have been in the tree, so they are deleted rather than repaired. The resume tests seed Claude transcripts under the Claude config dir; a local run with `CLAUDE_CONFIG_DIR` set to the empty string resolved that to the repo root, and a `git add -A` swept the output into the previous commit. All 15 are gone.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

projects/-tmp--tmp9Wovff-project/66666666-6666-6666-6666-666666666666.jsonl
```diff
@@ -0,0 +1 @@
+seed
```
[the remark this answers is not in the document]
njbrake: These files should never have been in the tree, so they are deleted rather than repaired. The resume tests seed Claude transcripts under the Claude config dir; a local run with `CLAUDE_CONFIG_DIR` set to the empty string resolved that to the repo root, and a `git add -A` swept the output into the previous commit. All 15 are gone.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/trash.rs
```diff
@@ -91,12 +92,61 @@ fn is_managed_single_worktree(inst: &Instance) -> bool {
...
+    };
+    if metadata.is_dir() {
+        // A repo of its own, not a linked worktree; nothing to strand.
+        return false;
+    }
+    let Ok(contents) = std::fs::read_to_string(&link) else {
+        return false;
+    };
+    match contents.split_once("gitdir:") {
+        Some((_, target)) => !Path::new(target.trim()).exists(),
```
jerome-benoit: [P1] Resolve relative `gitdir` targets from the worktree directory. AoE converts every managed worktree's `.git` file to a relative target in `GitWorktree::create_worktree` (`src/git/worktree.rs:906-910`), but this check resolves that target against the aoe process's current directory. I reproduced a registered worktree with `.git = gitdir: ../main/.git/worktrees/wt`: the target is missing from the process CWD, exists relative to the worktree, and `git worktree list` still reports the checkout. This therefore classifies healthy legacy trashed worktrees as terminal and permanently skips relocation. Resolve a relative target against the `.git` file's parent, as `read_linked_worktree_gitdir` already does, and cover the relative-link form in the regression test.

src/session/trash.rs
```diff
@@ -490,77 +542,266 @@ pub fn reconcile_trashed_location(inst: &mut Instance) -> bool {
...
+    batch: &[String],
+) -> anyhow::Result<Vec<Instance>> {
+    let now = Utc::now();
+    let reserved = storage.update(|instances, _groups| {
+        let mut reserved: Vec<(u64, Instance)> = Vec::new();
+        for id in batch {
+            let Some(stored) = instances.iter_mut().find(|candidate| &candidate.id == id) else {
+                continue;
+            };
+            match stored.try_acquire_lifecycle_reservation(
```
jerome-benoit, line 769: [P2] Revalidate the durable row before reserving it. Candidate discovery stores only the ID at lines 679-684; this later update reserves whichever current row has that ID without checking that it is still trashed or still matches the snapshot that produced the plan. If a peer restores the row between those operations, this call successfully puts a `Trash` reservation on the now-live session. `reconcile_trashed_location` then returns `Nothing`, but the reservation is released only by the final update after the rest of the batch, so launch/restore/purge operations can report `Busy` behind unrelated worktree moves. Use a CAS-style revalidation of `is_trashed()` and the plan-relevant snapshot fields before acquiring the reservation, and add a restore-between-scan-and-reserve regression test.

src/tui/home/lifecycle.rs
```diff
@@ -544,6 +529,12 @@ impl HomeView {
         let mut initial_disk_profiles: Vec<String> = view.storages.keys().cloned().collect();
         initial_disk_profiles.sort();
         view.rewire_disk_subscriptions(&initial_disk_profiles);
+        // Trashed-worktree relocation (#2522) and the repair of a worktree
+        // moved outside aoe (#2002) are healing work, not render input, and
+        // cost a git spawn and a storage write per broken row. They sweep the
+        // loaded profiles on a worker so they never delay the first frame
+        // (#3611).
+        view.reconcile_poller.request(initial_disk_profiles.clone());
```
jerome-benoit, line 532: [P2] Gate startup recovery on completion of the initial reconciliation sweep. `maybe_start_startup_recovery()` runs at line 518, before this request. It selects candidates from the stale in-memory rows and records them in the boot-scoped attempted ledger (`src/tui/home/pollers.rs:650-710`) before launching; the launch then uses `project_path`. For the #2002 case, recovery can therefore start with the missing recorded path before this worker repairs it. Once the repair lands, recovery is not retried during this boot because the attempt was already recorded. Keep reconciliation off the first-frame path, but trigger startup recovery after the initial poller result has completed, for both changed and unchanged results, and add a moved-worktree recovery regression test.

jerome-benoit (changes requested):
I reviewed this exact head with two independent, blind passes and cross-validated each result against the current code and live CI. The existing CodeRabbit threads are resolved and not duplicated here. All 35 reported checks completed with 34 successes and one skipped check.

One non-blocking documentation cleanup: the PR description first says batches contain 32 rows, then correctly says 8; the implementation uses 8.

src/session/trash.rs
njbrake: Right, and worse than the relative-paths config: `create_worktree` runs `convert_git_file_to_relative` on every managed worktree, so this would have read essentially every managed trashed worktree as stranded and skipped relocation for good, not just repos opting into `worktree.useRelativePaths`.

Now routed through `read_linked_worktree_gitdir` rather than a second copy of the resolution in `trash.rs`; it already resolves against the worktree and parses the pointer line-wise, which is also stricter than the `split_once` it replaces. Made it `pub(crate)`. The regression test covers the relative form and still asserts the genuinely stranded case; it fails on the previous code with "a live checkout with a relative gitdir link must not read as stranded".

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/trash.rs
njbrake: Agreed, and the Busy-behind-unrelated-moves consequence is the part that makes it worth fixing rather than tolerating: the work itself was already safe, since `reconcile_trashed_location` re-plans and returns Nothing, but the reservation outlived that and pinned a live session for the rest of the batch.

Candidates now carry their snapshot instead of just the id, and the reserving update compares and sets on the plan-relevant fields (still trashed, plus `project_path`, `pre_trash_project_path`, `worktree_info`, `scratch`) before acquiring. A row that moved under the scan is dropped from the batch rather than reserved. `a_row_restored_after_the_scan_is_not_reserved` restores between the scan and the batch and asserts the row is left at generation 0 with no reservation.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/tui/home/lifecycle.rs
njbrake: This one is a regression this PR introduced: reconciliation used to run inline in `HomeView::new` before recovery, and moving it to a worker inverted that order. The boot-scoped ledger makes it a single wasted attempt per row per boot rather than a retry, which is why it is not self-correcting.

`HomeView::new` now arms `startup_recovery_pending` instead of starting recovery, and `apply_reconcile_results` releases it exactly once when the sweep lands. Released on the unchanged path too, as you asked, and also on `Disconnected` so a dead worker cannot strand recovery for the whole boot. `startup_recovery_waits_for_the_first_reconcile_sweep` covers the gate being armed at construction, released by an unchanged sweep, and not re-firing.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/tui/home/pollers.rs
```diff
@@ -222,6 +222,79 @@ impl HomeView {
...
+        self.maybe_start_startup_recovery();
+    }
+
+    pub fn apply_reconcile_results(&mut self) -> bool {
+        use std::sync::mpsc::TryRecvError;
+
+        // Deadline release runs before the live-send guard below. Starting
+        // recovery spawns workers and never touches the terminal, so a long
+        // paste must not strand it any more than a blocked sweep can.
+        self.release_startup_recovery_gate(false);
```
jerome-benoit: **[P1] Apply a completed reconcile result before releasing startup recovery**

This function can release the gate from stale in-memory state in two ways. The deadline path runs before probing the result channel, so an already queued repair can be ignored while recovery starts. On the normal `changed=true` path, line 284 releases the gate and synchronously calls `maybe_start_startup_recovery()` before `reload_storage_only()` refreshes `self.instances` at line 288.

Recovery can therefore clone the old `project_path` and record the boot-scoped attempt even though the sweep has already repaired durable storage, recreating the lost-attempt failure this gate was added to prevent.

Probe the result before applying the deadline. When it reports a change, reload repaired storage before releasing the gate; if live-send postpones the reload, preserve the result and keep the gate armed. Please cover a queued changed result at the deadline, including the live-send case.

jerome-benoit (changes requested):
Re-review at `206fe1f1`. The previous findings around relative `gitdir` resolution, revalidation before reservation, and startup recovery ordering are addressed. One blocking ordering gap remains in the new deadline follow-up; see the inline comment. All checks are green at this head.

Non-blocking: the PR description still says batches contain 32 rows, while `RECONCILE_BATCH` is 8.

src/tui/home/pollers.rs
njbrake: Both paths fixed in `f941bc8a`. The channel is probed before the deadline, the repair is applied via `reload_storage_only`, and only then is the gate released. When live-send postpones the reload, `pending_reconcile_reload` preserves the result and holds the gate shut past its deadline; with nothing to apply the deadline still runs, so a paste alone cannot strand recovery.

Worth flagging one thing about the tests you asked for. The deadline case initially passed with the bug reinstated: the end state is identical either way, and the difference is only what recovery observes mid-call, which is unobservable in a test since `maybe_start_startup_recovery` returns early without a runtime. So `release_startup_recovery_gate` now asserts the invariant it depends on instead of relying on statement order. That assert is what makes the test discriminate; both tests fail on the reinstated bug now, and a future reorder trips it rather than passing quietly.

On the description: there is no "32" in it, both mentions already said 8. The stale part was the rationale, which still described every row holding its flock for the whole batch pass; that stopped being true when the batch moved to one flock at a time, and the real bound is now the reservation TTL. Corrected both sentences.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/tui/home/pollers.rs
```diff
@@ -222,6 +222,99 @@ impl HomeView {
...
+        if self.live_send.is_some() {
+            if !self.pending_reconcile_reload {
+                self.release_startup_recovery_gate(sweep_landed);
+            }
+            return false;
+        }
+
+        let mut reloaded = false;
+        if self.pending_reconcile_reload {
+            self.pending_reconcile_reload = false;
```
jerome-benoit: **[P1] Keep the repaired-state reload pending when it fails**

`pending_reconcile_reload` is cleared before the fallible `reload_storage_only()` call. If that reload fails, the error arm only logs and execution still releases the startup-recovery gate with `sweep_landed = true`. Recovery then reads the stale in-memory `project_path` and consumes the row's boot-scoped attempt even though reconciliation already repaired durable storage, recreating the lost-attempt failure this gate is meant to prevent. Clear the pending flag only after a successful reload, keep the gate armed while the repair remains unapplied, and cover a failed-then-successful reload.

src/session/trash.rs
```diff
@@ -91,12 +91,69 @@ fn is_managed_single_worktree(inst: &Instance) -> bool {
...
+    let metadata = match std::fs::symlink_metadata(&link) {
+        Ok(metadata) => metadata,
+        Err(error) => return error.kind() == std::io::ErrorKind::NotFound,
+    };
+    if metadata.is_dir() {
+        // A repo of its own, not a linked worktree; nothing to strand.
+        return false;
+    }
+    match crate::git::cleanup::read_linked_worktree_gitdir(worktree) {
+        Some(admin) => !admin.exists(),
```
jerome-benoit: **[P2] Distinguish an absent admin directory from a stat failure**

`Path::exists()` returns `false` for every metadata error, not only `NotFound`, so an `EACCES` or `EIO` while inspecting a live worktree admin directory is classified as an orphaned checkout. The new conservative handling above protects the `.git` entry but not its target; because this reconciliation sweep is requested only at load time, a transient error suppresses relocation until another application launch. Use `try_exists()` or equivalent and treat only `Ok(false)` as orphaned, while keeping other errors retriable.

jerome-benoit (changes requested):
Re-reviewed current head `6c5ac1c80c5a8c224051bc120c72279809b9b205`.

The prior reconcile/recovery ordering issue is fixed on the successful path, and the relative-gitdir, reservation CAS, and protected-default-branch findings are addressed. Two current-head blockers remain:

- A failed storage reload clears the pending-repair flag and still releases startup recovery, so recovery can consume its boot-scoped attempt from stale in-memory state.
- The orphan check treats every metadata error on the linked-worktree admin directory as absence, so transient access or I/O errors suppress reconciliation until another application launch.

Live CI is green. I did not run local build, test, or clippy commands under the review resource constraints.

Verdict: REQUEST_CHANGES.

src/tui/home/pollers.rs
njbrake: Fixed in `5d90b4a2`. The flag is cleared only after a successful reload and the call returns early on failure, so the gate stays shut and the next tick retries rather than releasing recovery onto rows the sweep has already superseded on disk.

This was the trade I flagged and left to you, so thank you for calling it: an unreadable store is a broken TUI whichever way it goes, while a silently spent recovery attempt is invisible to the user. `a_failed_reload_keeps_the_repair_pending_and_the_gate_shut` covers the failed-then-successful sequence you asked for, and fails on the previous code with "a dropped repair must not open the gate onto stale rows".

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

src/session/trash.rs
njbrake: Fixed in `5d90b4a2` with `try_exists`, only `Ok(false)` terminal.

Fair hit, and worth naming: this is the same defect the previous commit fixed one line earlier for the `.git` entry, under a doc comment explaining why guessing "stranded" from a transient error is the expensive mistake. I fixed the entry and left its target.

The test uses a symlink loop rather than a permission denial, since the suite runs as root and cannot provoke EACCES; it asserts the loop actually yields a stat error first, so it cannot quietly degrade into testing nothing. It fails with `Path::exists` restored.

_Note: this reply was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #3581, opened 2026-08-29
Sandbox: run-args policy (privileged, cap_add/cap_drop, security_opt, user, extra_run_args)


njbrake:
Every sandbox flag request funnels into `RuntimeInfo::build_create_args` (`src/containers/runtime_base.rs`), which emits a closed set today: `run -d --name -w`, volumes, `--network`, `-p`, `--cpus`, `-m`, env, image. #3218, #2704, #1016 and step 1 of #2268 are all "add fields to `SandboxConfig` and thread them through that function", so this issue consolidates them.

**Proposed:** a `SandboxRunPolicy` with `privileged`, `cap_add`, `cap_drop`, `security_opt`, `no_new_privileges`, `user`, and `extra_run_args` as the escape hatch. Rules:

- All new fields join `REPO_DENIED_SANDBOX_FIELDS` (the mechanism that already drops repo-defined `extra_volumes` and `mount_ssh`).
- `privileged`, `cap_add` and `security_opt` are profile-or-global only.
- Warn on unknown `[sandbox]` keys. `#[serde(flatten)] overrides` silently swallows them today, which is how #3218 was found.

Closes #3218, #2704 and #1016 on landing. #2268's `/root` hardcoding and the arbitrary-uid problem stay separate.

_Note: this issue was drafted by Claude Fable 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

njbrake:
Closing: implemented by #3623 (7da0c5a). `[sandbox]` accepts `privileged`, `cap_add`, `cap_drop`, `security_opt` and `extra_run_args`, all repo-denied; unrecognized repo config keys are logged (#3218). Not added: a `user` key, since container `$HOME` is still `/root` (arbitrary uids stay in #2268; `extra_run_args = ["--user", "1000:1000"]` covers matching images), and `no_new_privileges`, which `security_opt = ["no-new-privileges"]` already emits.

_Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #2704, opened 2026-07-07
Sandbox: no way to add container security hardening flags (--cap-drop ALL, no-new-privileges)


franroa:
## Problem

Sandboxed sessions run with the container runtime's **default capability set** and without `no-new-privileges`. The `docker/podman run` command is built from a closed set of flags in `RuntimeInfo::build_create_args` (`src/containers/runtime_base.rs:383`): `run -d --name … -w …` + volumes + env + ports + optional `--cpus`/`-m` + image. There is no way — via config, CLI, or env — to add security-hardening flags such as:

- `--cap-drop ALL` (drop Linux capabilities)
- `--security-opt no-new-privileges:true`
- `--pids-limit`, `--read-only`, custom seccomp profiles, etc.

For a tool whose sandbox feature exists specifically to contain high-autonomy agents (often running with permission checks disabled), running those agents as root with full default capabilities leaves a lot of hardening on the table.

## Proposal

Either (or both):

1. **First-class settings** on `SandboxConfig` (`src/session/config.rs:2006`), e.g.:
   ```toml
   [sandbox]
   cap_drop = ["ALL"]
   no_new_privileges = true
   ```
2. **A generic escape hatch** for advanced users:
   ```toml
   [sandbox]
   extra_run_args = ["--cap-drop", "ALL", "--security-opt", "no-new-privileges:true"]
   ```

Option 2 is a small, localized change (thread one `Vec<String>` through `SandboxConfig` → `ContainerConfig` → `build_create_args`) and would also unblock other requests (network mode, uid mapping) without waiting for first-class support for each flag. Option 1 gives safer defaults discoverability. Happy to discuss which shape you'd prefer — I can send a PR.

## Environment

- aoe v1.12.0, Docker runtime on Linux (applies to Podman as well)

---
*Disclosure per your template: the codebase research and drafting were done with an AI assistant (Claude Code); the request and review are mine. I'm comparing aoe's sandbox against an internal hardened agent-sandbox launcher and would like to converge on aoe.*

njbrake:
Consolidated into #3581, one run-args policy covering privileged, cap_add/cap_drop, security_opt, user and an extra_run_args escape hatch. The specifics from this issue are carried there.

_Note: this comment was drafted by Claude Fable 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #2268, opened 2026-06-17
Add opt-in run_as_host_user sandbox option (--user) + home-relative mounts for safe RW host mounts


franroa:
## Summary

Add an opt-in sandbox setting (e.g. `run_as_host_user`) that runs the container as the **host user's uid/gid** (`docker run --user $(id -u):$(id -g) ...`) and repoints the agent-config mounts to that user's home, instead of always running as root with config at `/root/.claude`.

## Motivation

Today sandboxes run as **root** and AOE mounts agent config into `/root/.claude`. This forces the copy-into-`sandbox/` design and makes it unsafe to bind-mount host directories read-write (root-owned writes would land on the host, and the cleanup/ownership story gets messy). A concrete consequence: anything Claude Code *rewrites at runtime* cannot be shared via a read-only mount.

**Example that motivated this:** RO-mounting `~/.claude/plugins/marketplaces` into a sandbox fails because Claude Code refreshes marketplaces on load (`rm` + re-clone):

```
Failed to load marketplace "claude-plugins-official" from source (github):
EROFS: read-only file system, rm '/root/.claude/plugins/marketplaces/anthropics-claude-plugins-official'
```

A RO mount can't allow the refresh; a RW mount as root would write root-owned files into the host's `~/.claude`. Running the container as the host user resolves both: writes are owned by the host user, so host dirs can be RW-mounted safely.

## The image already supports it

The sandbox image (`ghcr.io/agent-of-empires/aoe-sandbox:latest`) ships a uid-1000 `ubuntu` user with a writable home:

```
$ docker image inspect ... --format 'User="{{.Config.User}}"'
User=""                       # defaults to root, HOME=/root

$ docker run --rm --user 1000:1000 <image> sh -c 'id; echo $HOME; touch $HOME/.writetest && echo writable'
uid=1000(ubuntu) gid=1000(ubuntu)
HOME=/home/ubuntu
writable                      # /home/ubuntu is writable; /root is NOT writable by uid 1000
```

So `--user` works at the OS level out of the box.

## What needs to change

1. **Add `--user <uid>:<gid>`** in `build_create_args` (`src/containers/runtime_base.rs`). There is currently no `--user` handling anywhere in the source.
[...]

PR #3623, opened 2026-08-31
feat(sandbox): add container run-policy settings (#3581)

unbrice:
## Description

Implements #3581: `[sandbox]` gains a run policy: `privileged`, `cap_add`, `cap_drop`, `security_opt`, and `extra_run_args` to override some runtime options.

FTR, my personal interest is in `extra_run_args`, I'm looking to wire gVisor.
During implementation, I noticed that I would be adding quite a few settings that repo shouldn't be able to override, so I added a schema descriptor, `#[setting(repo = "allow" | "deny")]` to replace `REPO_DENIED_SANDBOX_FIELDS` / `REPO_ALLOWED_SESSION_FIELDS` lists.
As a drive-by fix: `sandbox.container_runtime` becomes `global_only`: it is resolved through `Config::load` and snapshotted per session at creation, so a profile or repo value could never take effect, and it picks which host binary executes, the same trust surface as `default_image`.

## Deviations from #3581

- **No `user` key.** Container `$HOME` is hardcoded to `/root` and the agent's state is bind-mounted under it, so a non-root uid cannot read its own configuration: the key would ship a setting that cannot work. Arbitrary uids are #2268, which #3581 holds separate. `extra_run_args = ["--user", "1000:1000"]` remains, for an image built to match.
- **No `no_new_privileges` key.** `security_opt` already writes `--security-opt`, so a dedicated key is a second spelling of one flag, with Docker taking the last as an undocumented precedence between two settings. #2704's profile is one `security_opt` entry.

## PR Type

- [X] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (not a UI change)

## Test Coverage Analysis

- [X] N/A: this PR doesn't change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [X] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle

## AI Usage

- [ ] No AI was used
- [X] AI was used

**AI Model/Tool used:**
Kimi CLI+Kimi K3, claude CLI+Claude Opus 5

Wrote bulk of impl with K3 in 2-3 iteration, switched to Claude for an automated review session then went back to Kimi for additional human review fixes.

**NOTE:**

## Summary

- Adds sandbox run-policy settings for privileged mode, capabilities, security options, and extra runtime arguments.
- Supports runtime-specific behavior for Docker, Podman, and Apple Container.
- Replaces repository field allowlists with per-field `repo = "allow"` or `repo = "deny"` policies.
- Makes `container_runtime` global-only and snapshots it when a session starts.
- Adds validation for capability and security option lists.
- Improves network setting checks and repository override warnings.
- Updates configuration and sandbox documentation.

## Benefits

- Enables runtime integrations such as gVisor.
- Prevents repository configuration from changing sensitive sandbox options.
- Provides clearer schema-based control for repository overrides.
- Adds stronger validation for container security settings.

## Follow-up

- Dashboard, TUI, CLI, and session-lifecycle test coverage remain outside this change.
- Related pre-existing security concerns are tracked separately.

### DIFF
--- a/src/containers/runtime_base.rs
+++ b/src/containers/runtime_base.rs
@@ -1,4 +1,4 @@
-use super::container_interface::{docker_env_args, ContainerConfig};
+use super::container_interface::{docker_env_args, ContainerConfig, RunFlag};
 use super::error::{sanitize_stderr, DockerError, Result};
 use std::process::{Command, Stdio};
 use std::time::{Duration, Instant};
@@ -59,6 +59,8 @@ pub(crate) struct RuntimeBase {
     pub supports_network_mode: bool,
     /// Whether `run --label key=value` and label inspection are supported.
     pub supports_labels: bool,
+    /// Which Docker-style `run` flags this runtime accepts (see [`RunFlag`]).
+    pub supported_run_flags: &'static [RunFlag],
     /// Case-insensitive stderr substrings that identify a "container does not
     /// exist" error for this runtime. Each runtime words it differently (Docker
     /// "No such container", Apple Container "notFound … not found"), so the
@@ -87,6 +89,14 @@ pub(crate) struct RuntimeBase {
     pub permission_denied_markers: &'static [&'static str],
 }
 
+/// Docker-style run flags supported by Docker and Podman.
+const ALL_RUN_FLAGS: &[RunFlag] = &[
+    RunFlag::Privileged,
+    RunFlag::CapAdd,
+    RunFlag::CapDrop,
+    RunFlag::SecurityOpt,
+];
+
 impl RuntimeBase {
     pub const DOCKER: Self = Self {
         binary: "docker",
@@ -100,6 +110,7 @@ impl RuntimeBase {
         supports_selinux_relabel: true,
         supports_network_mode: true,
         supports_labels: true,
+        supported_run_flags: ALL_RUN_FLAGS,
         not_found_markers: &["no such container"],
         // moby/moby client/errors.go connectionFailed() is the single source
         // of this message across every Docker OS variant (macOS Desktop, Linux
@@ -128,6 +139,9 @@ impl RuntimeBase {
         supports_selinux_relabel: false,
         supports_network_mode: false,
         supports_labels: true,
+        // `--cap-add`/`--cap-drop` exist since container 0.12; there is no
+        // `--privileged` or `--security-opt`.
+        supported_run_flags: &[RunFlag::CapAdd, RunFlag::CapDrop],
         // Apple Container surfaces a missing container with two distinct
         // wordings depending on the subcommand:
         // - `container delete`/`container logs`: `notFound: "container with
@@ -173,6 +187,7 @@ impl RuntimeBase {
         supports_selinux_relabel: true,
         supports_network_mode: true,
         supports_labels: true,
+        supported_run_flags: ALL_RUN_FLAGS,
         not_found_markers: &["no such container"],
         // Two distinct daemon-down wordings observed in real Podman output:
         // - "connect to Podman socket" fires on Linux socket mode
@@ -552,6 +567,51 @@ impl RuntimeBase {
             args.push(mem.clone());
         }
 
+        let policy = &config.run_policy;
+        let pairs = |flag: &str, values: &[String]| -> Vec<String> {
+            values
+                .iter()
+                .flat_map(|v| [flag.to_string(), v.clone()])
+                .collect()
+        };
+        for (flag, config_key, argv) in [
+            (
+                RunFlag::Privileged,
+                "privileged",
+                Vec::from_iter(policy.privileged.then(|| "--privileged".to_string())),
+            ),
+            (
+                RunFlag::CapAdd,
+                "cap_add",
+                pairs("--cap-add", &policy.cap_add),
+            ),
+            (
+                RunFlag::CapDrop,
+                "cap_drop",
+                pairs("--cap-drop", &policy.cap_drop),
+            ),
+            (
+                RunFlag::SecurityOpt,
+                "security_opt",
+                pairs("--security-opt", &policy.security_opt),
+            ),
+        ] {
+            if argv.is_empty() {
+                continue;
+            }
+            if self.supported_run_flags.contains(&flag) {
+                args.extend(argv);
+            } else {
+                tracing::warn!(
+                    target: "containers.runtime",
+                    "ignoring sandbox.{config_key}: {} does not support it",
+                    self.name
+                );
+            }
+        }
+
+        args.extend(policy.extra_run_args.iter().cloned());
+
         args.push(image.to_string());
         args.push("sleep".to_string());
         args.push("infinity".to_string());
@@ -750,7 +810,7 @@ impl RuntimeBase {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::containers::container_interface::{EnvEntry, VolumeMount};
+    use crate::containers::container_interface::{EnvEntry, RunPolicy, VolumeMount};
 
     // Real stderr captured from `<runtime> rm/delete <missing>` on 2026-07-01.
     // These pin the per-runtime not-found classification that `remove()` and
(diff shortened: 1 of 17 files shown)


src/session/config/mod.rs
```diff
@@ -2364,6 +2371,86 @@ pub struct SandboxConfig {
     )]
     pub selinux_relabel: bool,
```
jerome-benoit, line 2460: **[P2] Keep host SELinux relabeling out of repo scope.** This field has no `repo = "deny"`, so it inherits the sandbox section's repo-allow default. A repository can therefore make AoE append `:z` to every bind mount, including automatically mounted host configuration and, when enabled globally, credentials. Docker documents that `:z` changes SELinux labels on the host and can affect files outside Docker: https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label. Please deny this field at repo scope and add it to the policy regression cases.

src/containers/runtime_base.rs
```diff
@@ -124,6 +135,7 @@ impl RuntimeBase {
         supports_named_volumes: false,
         supports_selinux_relabel: false,
         supports_network_mode: false,
+        supported_run_flags: &[],
```
jerome-benoit: **[P2] Preserve Apple Container's supported capability controls.** Current Apple Container supports `--cap-add` and `--cap-drop`, but an empty `supported_run_flags` drops both requests and continues after only a warning. In particular, `cap_drop = ["ALL"]` silently loses requested hardening. Please enable the supported subset (`CapAdd` and `CapDrop`), keep only genuinely unsupported flags disabled, and update the test/docs accordingly: https://github.com/apple/container/blob/main/docs/how-to.md#control-linux-capabilities.

jerome-benoit (dismissed):
I cross-validated this at `6d2a925eddc267b45e3d4cd816855bc0943515b6`. The new run-policy propagation and the repo-deny policy on the five new fields are coherent, but three issues remain.

**[P1] Enforce network value validation for repo/profile TOML.** `sandbox.network` remains repo-overridable, while repo/profile TOML is only type-checked. `sanitize_network` rejects only the exact `host` value and passes every other string through to `--network`, so a repo can set `network = "container:<name-or-id>"` even though the schema validator rejects namespace-sharing forms. Docker documents that this mode shares the target container's networking stack, including its loopback interface: https://docs.docker.com/engine/network/#container-networks. Please apply `validate_network_format` in the runtime sanitization path (or otherwise value-validate TOML overrides) and add repo/profile regression cases for `container:` and Podman's `ns:` form.

The two P2 findings are inline. I did not duplicate CodeRabbit's existing documentation finding. GitHub currently exposes only the PR title/template checks for this head; I did not run local builds or tests.

src/containers/runtime_base.rs
unbrice: Thanks, I'm not familiar at all with Apple Container. Done.

src/session/config/mod.rs
unbrice: Indeed, there seem to be pre-existing security issues. I opened #3679 to prioritize them first.

jerome-benoit (dismissed):
Re-reviewed and independently cross-validated at `b2240300073806ca7f354750fd5cfb10bd9b088f`. The three findings from my previous review are fixed: repo/profile network values are revalidated at container-build time (`src/session/container_config.rs:2131-2166`, regression cases at `:2276-2285` and `:3929-3985`), `selinux_relabel` is repo-denied (`src/session/config.rs:2362-2373`), and Apple Container now preserves `cap_add`/`cap_drop` while rejecting only unsupported flags (`src/containers/runtime_base.rs:127-140,1433-1455`).

One blocker remains, already covered by the unresolved CodeRabbit thread, so I am not duplicating it inline:

- **[P1] A repo can still select a user-defined host command through `session.default_tool`.** `default_tool` is explicitly repo-allowed at `src/session/config.rs:901-911`; default resolution prefers an exact `custom_agents` match at `src/cli/add.rs:1534-1549`; and session construction resolves and launches that command at `src/session/builder.rs:750-780` and `src/session/config.rs:1708-1718`. The regression test at `src/session/repo_config.rs:2015-2038` currently asserts that this selection survives repo sanitization. Although the old allowlist already permitted it, this PR is the change that codifies repo trust policy next to each field, and the claim that the value is safe because it only names a tool does not hold at the execution boundary. Please either repo-deny `default_tool`, or allow only built-in agent names from repo scope, with a regression test proving that a repo cannot select a user `custom_agents` command. Existing thread: https://github.com/agent-of-empires/agent-of-empires/pull/3623#discussion_r3906224300

Also confirmed, but non-blocking by itself:

- **[P3]** The canonical validation list at `docs/development/adding-settings.md:61-66` still omits the newly supported `capability_list` and `security_opt_list` rules implemented at `aoe-settings-derive/src/lib.rs:334-338` and `src/session/settings_schema/validate.rs:112-122`. This is already covered by CodeRabbit's earlier outside-diff finding.

Verification: `git diff --check d90e5531c45f04742824d596d8acb34eac5a7133..b2240300073806ca7f354750fd5cfb10bd9b088f` produced no output; targeted `rg`/`sed` inspection produced the paths above; the final live `gh pr view`/GraphQL query still reported this exact head and the `default_tool` thread unresolved. Exact-head Rust, format, clippy, CodeQL, Nix, Vitest, mocked Playwright, live shard 1, and Codecov checks pass. Live Playwright shard 2 and its fan-in are red because the runner's Microsoft apt repositories returned HTTP 403 before tests ran; this is infrastructure evidence, not a demonstrated product regression, but the check still needs a successful rerun.

jerome-benoit (dismissed):
Correction after reading and incorporating the author's split response: the network P1, SELinux P2, and the shared repo-policy commit `3af85574479529ebb119b596f640b62d2f1d60e1` now belong to #3679. I have transferred the remaining `default_tool` trust-boundary finding there. The run-policy commit specific to this PR (`b2240300073806ca7f354750fd5cfb10bd9b088f`) fixes the Apple Container capability issue and has no remaining blocking implementation finding. The documentation omission already raised by CodeRabbit remains P3/non-blocking.

This exact head must still not merge yet: #3623 and #3679 both target `main`, and #3623 still physically contains the shared `3af8557` commit. Merging it now would bypass the separate review/fix in #3679. After #3679 resolves the transferred blocker and lands, please rebase #3623 onto the updated `main` so its diff contains only the run-policy change; I will then verify the new head.

This review supersedes the attribution in my previous re-review; the blocker is owned by #3679, not by the run-policy delta.

jerome-benoit (commented):
Status correction: I am not requesting any change to #3623's run-policy implementation. The Apple Container issue is fixed, and the remaining repository-policy security finding is owned and tracked by #3679. The existing validation-documentation omission is P3/non-blocking.

I dismissed my previous `REQUEST_CHANGES` reviews on this PR because their findings were either fixed or incorrectly attributed after the author's split. Merge ordering or a later rebase is not a code-quality finding against this PR and should not have been represented as one.

jerome-benoit (approved):
Reviewed at a7a1df86de62cc7357274f677ad46ce18d64e9a5, including the latest rebase, with two independent reviews and cross-validation. No actionable findings retained. The previous Apple Container and validation-documentation concerns are addressed; the security findings split into #3679 are outside this PR’s scope.

Validation is limited to static inspection: only title lint is reported as passing, and no local tests were run.

njbrake (approved):
Approving. Run policy is threaded through the existing `RuntimeBase` capability gating rather than a parallel path, defaults are empty so unconfigured installs produce identical argv, and all five keys are repo-denied.

Verified Apple Container's `--cap-add` / `--cap-drop` support against upstream `docs/runtime-configuration.md`; `--privileged` and `--security-opt` are absent there, so the `supported_run_flags` split is right. Locally: fmt, clippy, `cargo test --lib` (two unrelated failures, fixtures that need a non-root uid), and the config integration tests.

Pushed one commit: `network` was missing from the validate-rule list in the authoring guide, and the new descriptor test repeated an assertion already covered by `field_repo_policy_resolves_from_field_then_section`.

Two non-blocking notes for later. These five keys are the only `local_only` fields in `[sandbox]`, so unlike `extra_volumes` or `mount_ssh` they are invisible in the dashboard rather than elevation-gated; `extra_run_args` earns that, the other four are arguable. And `validate_capability_format` rejects lowercase names that Docker and Apple Container both accept.

PR #3629, opened 2026-08-31
fix(web): render every theme's code pane in its own Shiki palette

matthewpwatkins:
## Description

The dashboard has two code renderers and they disagreed with each other:

* The file viewer honored the built-in 8 default themes, but fell back to GitHub Dark or GitHub Light for everything else.
* The diff pane honored all syntax palettes requested by themes-- even custom themes (like those in my [VS Code themes plugin](https://github.com/matthewpwatkins/aoe-vscode-themes)).

So if you opened one file in both panes, you saw it painted two different ways, with a potentially jarring experience on the file viewer pane.

The file viewer now resolves palettes the way the diff pane already did, by reading Shiki's own registry of the themes it bundles instead of a hand-maintained subset of it. Any of the 65 palettes Shiki ships can be named by any theme-- built-in or not-- and the two panes now agree. 

Palettes still load only when a theme asks for one, so the initial download is unchanged; actually the entry chunk comes out 6 KB smaller. A nonexistent Shiki them keeps the appearance-matched GitHub Light/Dark fallback and now logs a one-time warning instead of substituting silently.

The screenshots below show the same file open in the files pane and the diff pane at once, under the `vscode-red` theme, which asks for Shiki's Red palette, making the old discrepancy especially clear:

### Before (files pane uses GitHub Dark, diff pane uses custom theme)

image

### After (files pane and diff pane both respect the custom theme)

image

TODO (#3913): The two panes now agree on which palettes exist, but I noticed they still load them by different routes. Collapsing them is a bigger change than this fix and reviews better on its own, so this PR only drops a pointer to #3913 in the comments of both entry points.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Covered by Vitest unit tests because this is palette resolution in `web/src/lib/`, not a browser behavior (focus, keyboard, drag/drop, touch). The coverage matrix tracks `web/src/components/**`, so it needs no entry and `node tests/validate-coverage-matrix.mjs` agrees.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code

**Any Additional AI Details you'd like to share:** Claude found the allowlist, wrote the fix and tests, and prepared this PR; a second model reviewed it and its findings are folded in. I reviewed the diff and ran both sides in local debug builds against my own plugin themes to take the screenshots above.

## How I tested

- The theme tests now run against Shiki's real registry instead of a mock of it, so an id Shiki drops upstream fails the suite rather than passing against a stand-in. The two failure paths (module resolves without a usable theme, import rejects) swap one registry entry for the duration of their own test.
- Added a case asserting every builtin theme's declared palette is one Shiki can actually load. The Rust-side check only asserted the field was present.
- Built debug binaries from `main` and from this branch, ran each against an isolated home seeded with the nineteen VS Code plugin themes, and compared the files and diff panes side by side under `vscode-red` (see screenshots). Sampled the rendered token colors rather than eyeballing them: `export` comes back as GitHub Dark's `#f97583` before and Red's `#f12727` after.
- `npm run test:unit`, `npm run format:check`, `npm run lint`, `npx tsc -b`, and `node tests/validate-coverage-matrix.mjs` all clean.

## Summary

- Shiki now provides the theme registry for syntax highlighting.
- All bundled Shiki themes and custom theme selections now work in the file viewer and diff pane.
- Unknown themes use the existing GitHub Light/Dark fallback and log one warning.
- Theme loading remains lazy, with a smaller entry bundle.
- Tests now use Shiki’s real theme registry and validate configured builtin themes.
- Documentation and comments explain the separate highlighting paths.

## Benefits

Users see consistent syntax highlighting across file and diff views. Theme support is broader and easier to maintain.

## Further enhancement

The file viewer and diff pane still use separate highlighting engines. Consolidating them could reduce duplicated work in the future.

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -87,7 +87,7 @@ aoe theme list                      # show all available themes
 aoe theme dir                       # print the custom themes directory
 ```
 
-The schema is flat and every field is optional. Missing color fields fall back to the Empire baseline; an omitted `appearance` or `[syntax].shiki_theme` is derived from the theme's background luminance rather than copied from Empire. Color fields cover background, borders, text, status semantics, diff colors, branch/sandbox chips, and accent. `appearance = "dark" | "light"` and `[syntax].shiki_theme` control the web dashboard's surface ramp and code-block syntax theme.
+The schema is flat and every field is optional. Missing color fields fall back to the Empire baseline; an omitted `appearance` or `[syntax].shiki_theme` is derived from the theme's background luminance rather than copied from Empire. Color fields cover background, borders, text, status semantics, diff colors, branch/sandbox chips, and accent. `appearance = "dark" | "light"` and `[syntax].shiki_theme` control the web dashboard's surface ramp and code-block syntax theme. `shiki_theme` accepts any theme id [Shiki bundles](https://shiki.style/themes); a name it does not ship falls back to `github-dark` or `github-light` by appearance.
 
 ## Session
 
--- a/web/src/components/diff/pierre/DiffWorkerPoolProvider.tsx
+++ b/web/src/components/diff/pierre/DiffWorkerPoolProvider.tsx
@@ -6,6 +6,9 @@ import { useShikiTheme } from "../../../hooks/useShikiTheme";
  * Provides a shared off-main-thread highlighter worker pool for the diff
  * renderer, so syntax highlighting of large diffs doesn't block the UI.
  *
+ * The non-diff panes highlight through `lib/highlighter` on the main thread
+ * instead, so two engines run side by side. See #3913.
+ *
  * Keyed by the active Shiki theme so a theme switch re-initializes the pool
  * with the new theme. When `Worker` is unavailable (SSR / jsdom tests) it
  * renders children directly; the diff components then highlight on the main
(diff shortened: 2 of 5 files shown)


jerome-benoit (approved):
Reviewed a4b666a8 against 03452cff. The registry change removes the artificial palette restriction while keeping lazy imports and guarding arbitrary IDs with Object.hasOwn (`highlighter.ts:49–83`). Tests use the real registry, restore failure stubs in finally, and check builtin palette declarations (`highlighter.test.ts:45–60,329–368`). Consolidating the two engines is appropriately deferred to #3913.

No actionable finding after two independent sequential analyses and arbitration. Evidence: exact-object five-file diff and source inspection, matching installed/locked Shiki 4.4.3 registry, clean git diff --check. No local tests, visual comparison or bundle-size measurement repeated.

CI qualification: only metadata checks had completed when review began. I authorized the waiting CI, Tests and CodeQL workflows; they are now queued/running, not yet validated. This approval concerns the reviewed code, not a claim that pending CI passed.

PR #3632, opened 2026-08-31
refactor(server): make the daemon core, gate only the web bundle behind `web`

njbrake:
## Description

Building from source without Node used to give you a terminal-only program. The background service that powers the phone and browser dashboard, the structured agent view, and the whole API were all left out of that build, which made it far less useful than it sounds and blocked planned work that needs the service to always be present. This change puts the background service in every build. The only optional piece now is the dashboard web page, which is the part that genuinely needs Node to compile. A plain build gives you a working service with no Node involved; one flag adds the dashboard on top. Build without the dashboard and the service still starts and answers requests, you just get a 404 if you point a browser at it.

Concretely: the daemon becomes unconditional and only the embedded bundle stays behind a cargo feature, renamed `serve` to `web`. About 700 `cfg(feature = "serve")` sites and their paired `not(serve)` fallback stubs are gone. `web` gates `src/server/assets.rs`, the `/assets/*` and SPA-fallback routes, `build_frontend()`, the `aoe.web` builtin, and a QR seam.

Three deviations from the issue, all forced by real call sites:

- **`aoe serve` is not hidden in a `web`-less build** (acceptance criterion 4). Per maintainer decision it starts a headless daemon instead, because hiding it would leave no way to start the daemon and would break the TUI structured view in exactly the build this refactor exists to enable. No new gating was needed: `web_disabled()` is already false when the `aoe.web` builtin is absent.
- **`mime_guess` stays core.** It backs the session file-read endpoint's Content-Type, not just embedded assets.
- **`webbrowser` stays core.** It backs the plugin host's `ui.open_url` and structured-view link activation.

A `web`-less build now says so where it used to imply a dashboard: the Remote Access
screen is titled `Remote API Access`, the space the QR occupied explains that the URL
serves the API only, and the home badge reads `Serving API`.

`serve` remains a deprecated alias for `web` for one release. `build.rs` warns when it is set, and a CI guard rejects any new `cfg(feature = "serve")` so the alias cannot silently diverge from `web`.

CI keeps a Node-free leg and a `web` leg: the rest leg drops to default features as `core-rest`, and a new bare-core step builds with `node`, `npm` and `npx` shadowed by failing stubs, which is what actually enforces the Node-free build.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
[...]

PR #3647, opened 2026-08-31
feat(server,tui): expose context resume availability

jerome-benoit:
## Description

Expose request-invariant context-resume availability on each session returned by `GET /api/sessions`.

- Centralize static terminal resume constraints and reuse them for launch flag emission and API projection.
- Keep runtime-dependent availability conservative without blocking the async sessions endpoint on tmux.
- Use the shared authenticated HTTP client in the remote TUI.
- Preserve compatibility with older daemons by treating absent `status`, `view`, and `context_resume` fields as defaults or not reported.
- Keep the remote picker structured-only, title-sorted, and directly openable.
- Do not negotiate attach metadata: callers already know whether they can open structured rows.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Local targeted tests and Clippy pass. The complete local suite still exposes two failures outside this PR's 14-file diff: the main-branch real-tmux framing test on this tmux version and one parallel storage-fixture isolation failure. The storage test passes in isolation. CI is the authoritative clean environment check.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Verification:

- `cargo clippy --all-targets --features web,e2e-tests -- -D warnings`
- `cargo test --features web,e2e-tests context_resume_projects_structured_and_terminal_states`
- `cargo test --features web,e2e-tests remote_home`
- `cargo test --features web,e2e-tests session_id`
- `cargo test --features web,e2e-tests --test e2e remote_home_accepts_old_daemon_rows_and_keeps_structured_title_order`

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

OpenAI Codex GPT-5.6 through Oh My Pi.

Independent reviewer and scout passes were cross-validated against the current code and live command output.

**NOTE:**


src/session/instance/session_id.rs
```diff
@@ -634,6 +634,71 @@ impl Instance {
...
+        }
+        match &self.resume_intent {
+            ResumeIntent::Cleared => TerminalContextResume::ForcedFresh,
+            ResumeIntent::Fork { .. } => TerminalContextResume::ForkPending,
+            ResumeIntent::Use(target) => {
+                if is_valid_session_id(target) {
+                    TerminalContextResume::Available
+                } else {
+                    TerminalContextResume::InvalidTarget
+                }
```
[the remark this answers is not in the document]
jerome-benoit: Addressed in `29f886c9`. `terminal_resume_explicit_target_invalid()` is now the shared decision used by both the client-context projection and `apply_session_flags()`. Invalid explicit targets remain `unavailable/invalid_target`, but no resume flag is emitted, so launch falls back to fresh behavior. The existing projection test now also asserts that the launch command stays unchanged. Targeted Rust test, formatting, and Clippy pass.

web/src/lib/api.coverage.test.ts
```diff
@@ -107,12 +107,96 @@ function bodyOf(init: RequestInit | undefined): unknown {
...
     fetchSpy.mockResolvedValueOnce(jsonResponse(env));
-    const result = await fetchSessions();
-    expect(result).toEqual(env);
-    expect(lastCall()[0]).toBe("/api/sessions");
+    expect(await fetchSessions()).toEqual(env);
+    const [url, init] = lastCall();
+    expect(url).toBe("/api/sessions");
+    expect(init?.headers).toEqual({ "X-Aoe-Client-Capabilities": "acp_ws_v1,terminal_ws_v1" });
+  });
+  it.each([
```
jerome-benoit:
```suggestion

  it.each([
```

njbrake (changes requested):
Three blocking items.

**1. The merged result does not compile.** #3632 landed on main after this PR's last CI run. It demoted `serve` to a deprecated one-directional alias (`serve = ["web"]`), moved `src/server` into the core build, and added a CI job that fails on any `feature = "serve"` under `src/`. This PR adds five such gates. Merging `origin/main` into the head is textually clean, then `cargo check --lib` fails:

```
error[E0432]: unresolved import `crate::session::TerminalContextResume`
  --> src/server/api/sessions.rs:18:42
  note: the item is gated behind the `serve` feature
error[E0599]: no method named `terminal_context_resume` found for reference `&instance::Instance`
  --> src/server/api/sessions.rs:778:16
```

`cargo build` and `cargo build --features web` both fail; only `--all-features` still compiles, which is why the Clippy job would not catch it. The gates cannot become `feature = "web"` either, since `src/server` now compiles unconditionally. Drop all five: `src/session/mod.rs:76`, `src/session/instance/mod.rs:94`, `src/session/instance/session_id.rs:660,669,3290`. That last one also means `terminal_context_resume_matches_emission_constraints` never runs under a normal `cargo test` today.

**2. `GET /api/sessions` can stall on tmux.** `context_resume_for` reaches `probe_session_existence`, which falls through to a synchronous `tmux list-sessions` (10s timeout) whenever the shared snapshot is past its 2s TTL. That runs on the tokio worker thread, holding `state.instances.read()`, on an endpoint each dashboard tab polls every 3s. `status_poll.rs:298` wraps the same call in `spawn_blocking` for this reason, and #3537 moved the TUI the same direction. Either run the projection under `spawn_blocking`, or read the snapshot with no synchronous fallback and let a stale snapshot answer `RuntimeCheckRequired`.

**3. The web strictly validates a field nothing reads.** `session_interactions` appears in `web/src/` only inside `api.ts`. No component consumes it, yet any schema mismatch makes `decodeSessionsEnvelope` return null, which `useSessions` turns into `setServerDown(true)`. Skew or a future server-side enum variant therefore presents as the daemon being offline, and `rejects malformed interaction contracts` locks that in by asserting rejection of unknown states and reasons. Drop the web half until a component renders it, or make the field non-fatal.

Non-blocking:

- `docs/api.md` documents this endpoint for external orchestrators and needs the new request header and its 400. A caller sending no header silently gets `attach: client_missing_transport` on every session.
- The deleted comment explaining why sandboxed Copilot, Kimi, and Prime Agent must start fresh (container-local session stores) has no replacement home. `terminal_resume_static_unavailable` is the place for it.
- `remote_home` went from structured-only sorted by title to all sessions in server order. That is a visible TUI change under a `feat(server)` title with the rendering box marked skipped and no recording.

On design: `session_interactions` is per-session data in a side map, so both clients re-join by id and reimplement a completeness contract that `SessionResponse` gives for free. The request-scoped `attach` half justifies separation; `context_resume` is request-invariant and belongs on the row.

The daemon projection and the resume-classification unification are solid, and the fix CodeRabbit prompted in 29f886c is the right one. The blockers are the stale feature gate, the probe placement, and the web coupling.

web/src/lib/api.coverage.test.ts
jerome-benoit: Applied in `86c63fcc`. The blank line now separates the preceding test from the table-driven `it.each` block in `web/src/lib/api.coverage.test.ts`. No behavioral change.

njbrake (changes requested):
Take `context_resume` and the `session_id.rs` consolidation; drop the attach negotiation.

`context_resume` is the right shape and matches `acp_capable` on the same struct: the daemon answers what can be done with a session so clients stop holding their own copy of the rule. Sharing `terminal_resume_static_unavailable` between `apply_session_flags` and the projection is the durable part of this PR.

`session_attach` is not daemon state. `attach_for` reduces to `view == Structured && caller sent acp_ws_v1`, both already known to the caller. The remote picker filters to structured rows and always advertises ACP, so `can_open()` is true for every row it lists; the `view-only` cell, the `Enter=unavailable` footer, and the Enter filter are unreachable in production. The cost is a required envelope field, a per-request map, and a `Vary` that makes the response uncacheable. If a client later needs an answer it cannot derive, such as a dead ACP worker or a CityHall restriction, that is a fact about the session and belongs on the row.

Separately blocking: `RemoteSessionWire` makes `context_resume` required and `SessionsAttachEnvelope` makes `session_attach` required, so an older daemon fails the parse and the picker renders `missing field context_resume` over an empty list. `daemon_status_poller.rs:36` states the rule this breaks, and the type being replaced carried it too.

Also: split out the four unrelated test fixes; revert the 49 spec files adding `context_resume` to mocks the dashboard never reads; `Composer.usageHint.test.tsx:90` now asserts `toLocaleString()` against `toLocaleString()`.

njbrake (approved):
Rework addresses the standing review. The attach negotiation is gone, `context_resume` compat is fixed with `#[serde(default)]` on the wire type and covered end to end against a daemon that omits the field, the web half and the unrelated test fixes are out, and the picker is back to structured-only and title-sorted.

Pushed four fixes on top: restored the resume diagnostics `build_resume_flags` emits, reverted the `list_sessions_state_filter` JSON round-trip, documented the `context_resume` wire shape, and recorded which states the picker can receive.

Two non-blocking follow-ups. The `env.rs` tmux locale fix dropped from this PR is unlanded. And the structured branch cannot return `available` because nothing records ACP `session/load` capability; `structured_fork_capable` is the pattern for that.

PR #3656, opened 2026-09-01
fix(acp): align aoe-agent node, sdk and model contracts

### LINKED ISSUE #3653: Align aoe-agent runtime and model contracts
## Problem

The `aoe-agent` runtime and model contract has three inconsistent sources:

- `src/acp/node.rs` accepts Node 20 and says it matches the package, while `acp-worker/aoe-agent/package.json` requires Node 22 or newer.
- The package depends on `ai` 7, while its description and `src/acp/agent_profiles.rs` still identify the implementation as Vercel AI SDK 6.
- `Instance.agent_model` advertises `llama3.3:ollama`, but `pickModel` has no Ollama or OpenAI-compatible branch. Unknown model IDs fall through to the Anthropic provider.

## Impact

- A host Node version can pass the AoE prerequisite check while violating the package's declared runtime contract.
- The SDK documentation is misleading during maintenance.
- Copying the advertised Ollama example selects the wrong provider instead of producing a valid local-model configuration.

The runtime mismatch and misleading model example make this **P2**.

## Expected fix

- Align `MIN_NODE_MAJOR`, its documentation, and `engines.node` on one supported minimum, with a boundary test for the prerequisite check.
- Update the SDK-major references to match the installed dependency.
- Either replace the Ollama example with a supported model ID or implement and test explicit provider parsing for the intended Ollama/OpenAI-compatible path.

Found during the cross-validated post-merge review of #3637 at `bb2134b11b7daacf1075ffcfcd2ec76e78a96842`. The separate packaging/presence problem is already tracked by #3553.

zerone0x:
## Description

Fixes #3653.

`aoe-agent`'s runtime and model contract was stated in several places that disagreed with each other. This aligns all of them and adds the drift guards.

**Node minimum (20 vs 22).** `MIN_NODE_MAJOR` was 20 and its doc claimed it matched the package, but `acp-worker/aoe-agent/package.json` declares `engines.node: >=22.0.0`. Node 22 is the real floor, not a preference: `aoe-agent` ships TypeScript sources and its `start` script is `node --experimental-strip-types`, a flag that does not exist before Node 22.6. So a Node 20 host could read `[OK] Node runtime` from `aoe acp doctor` and then fail at spawn. Raised the constant to 22 rather than relaxing `engines.node`, which would have made the doctor bless a runtime that cannot start the agent. Every other Node reference in the repo (`docker/Dockerfile`, `release.yml`, CI) is already on 22, so nothing else moves.

`aoe acp doctor` also kept its own `parse_node_major` plus a hardcoded `m >= 20`, a second copy of the boundary that had already drifted. `check_node` now calls the new `node::meets_minimum`, so the doctor's verdict and the spawn path's gate cannot diverge again. `meets_minimum` returns `None` for unreadable `node --version` output, which the call site's `unwrap_or(false)` treats as "not proven compatible" rather than as a pass.

**AI SDK major (6 vs 7).** The package depends on `ai` `^7.0.77` while the description, `src/acp/agent_profiles.rs`, `src/acp/mod.rs`, the agent's own file header and `docs/structured-view.md` all still said SDK 6. Updated to 7.

**`llama3.3:ollama`.** Took the smaller change: replaced the example with `gemini-2.5-pro`. The scope allowed implementing a provider branch instead "if the code clearly already supports an OpenAI-compatible provider path", and it does not. `@ai-sdk/openai-compatible` is a declared dependency but is never imported, and `pickModel` has no branch for it, so the advertised id falls through to `return anthropic(modelId)` and silently selects the wrong provider. Making it work would mean a new import, a base-URL setting to point at the Ollama endpoint, and settings/env plumbing to carry it, which is a feature, not a docs correction. I left the unused dependency in place so that feature is not blocked, and instead documented the routed set at both ends: on `Instance.agent_model` and above `pickModel`, which now states explicitly that anything outside Anthropic/OpenAI/Google hits the Anthropic fallback. `gemini-2.5-pro` is the example `aoe add --model` and `docs/cli/reference.md` already use.

`flake.nix`: the new test embeds `acp-worker/aoe-agent/package.json` with `include_str!`, and the Cargo source filter keeps only `*.rs`/`*.toml`, so the path is unioned into `commonArgs.src`. This is the same treatment `docker/Dockerfile` already gets for the `agent_compat` pin test. `scripts/check-nix-embedded-assets.py` passes.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

Two tests in `src/acp/node.rs`, plus the removal of the now-dead `parse_node_major` test in `src/cli/acp.rs` (its coverage moved to the shared helper):

- `meets_minimum_is_inclusive_at_the_boundary` is the boundary test the issue asks for. Table-driven over the three majors around `MIN_NODE_MAJOR` (below, exactly at, above), written against the constant rather than a literal so a future bump keeps testing the new boundary, plus the unparseable and empty cases that must stay `None`.
- `package_engines_matches_min_node_major` parses the real `engines.node` out of the manifest and asserts it equals `MIN_NODE_MAJOR`, so bumping one side alone fails CI instead of shipping the drift this issue is about. It also asserts `PINNED_NODE_VERSION` clears the same bar, since otherwise `download` would hand `verify_path` a runtime it rejects.

`acp-worker/aoe-agent`: `node --experimental-strip-types --test test/*.ts` passes (5/5) on Node 22.19.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the `aoe acp doctor` change is a version-comparison boundary, and an e2e test would assert against whatever Node the runner happens to have rather than against the boundary. The unit test above covers it directly and cheaply, which is what AGENTS.md asks for.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Note on local verification: this environment has no C toolchain (`cc` is absent and the sandbox cannot install one), so `cargo build`/`test`/`clippy` could not be run locally; `cargo fmt` and `scripts/check-nix-embedded-assets.py` were run and pass, and the Node test suite was run. Please lean on CI for the Rust checks.

## Summary

- Raised the minimum Node.js version to 22 across runtime checks, package metadata, and documentation.
- Added shared Node.js validation and boundary tests.
- Updated AI SDK references from version 6 to version 7.
- Replaced the unsupported Ollama example with `gemini-2.5-pro`.
- Documented supported model routing and fallback behavior.
- Updated Nix asset handling to include `package.json`.

## Benefits

- Keeps runtime checks, package requirements, and documentation consistent.
- Prevents unsupported model examples from selecting the wrong provider.
- Improves maintenance through shared validation and manifest-based tests.

Node tests, formatting, and embedded-asset checks pass. Rust checks remain unverified because the local environment lacks a C toolchain.

## Further enhancement

Add explicit Ollama or OpenAI-compatible provider support if local-model routing is required in the future.

### DIFF
--- a/acp-worker/aoe-agent/package.json
+++ b/acp-worker/aoe-agent/package.json
@@ -1,7 +1,7 @@
 {
   "name": "aoe-agent",
   "version": "0.0.1",
-  "description": "Multi-provider ACP agent for the Agent of Empires structured view. Wraps Vercel AI SDK 6 in an ACP server.",
+  "description": "Multi-provider ACP agent for the Agent of Empires structured view. Wraps Vercel AI SDK 7 in an ACP server.",
   "type": "module",
   "main": "src/index.ts",
   "bin": {
--- a/acp-worker/aoe-agent/src/index.ts
+++ b/acp-worker/aoe-agent/src/index.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 /**
- * aoe-agent: ACP server wrapping Vercel AI SDK 6.
+ * aoe-agent: ACP server wrapping Vercel AI SDK 7.
  *
  * One Node process per structured-view session. Accepts ACP requests from aoe
  * (the Rust ACP client) on stdin/stdout, drives a Vercel AI SDK loop
@@ -288,6 +288,12 @@ function serialiseToolOutput(output: unknown): Record<string, unknown> {
   return { value: output };
 }
 
+/**
+ * Map an `AOE_AGENT_MODEL` id onto a provider. Anthropic, OpenAI and
+ * Google only, by bare prefix or an explicit `provider:` prefix.
+ * Anything else hits the Anthropic fallback, so local-model ids are not
+ * a valid configuration until an openai-compatible branch exists.
+ */
 function pickModel(modelId: string) {
   if (modelId.startsWith("claude-") || modelId.startsWith("anthropic:")) {
     return anthropic(modelId.replace(/^anthropic:/, ""));
--- a/docs/structured-view.md
+++ b/docs/structured-view.md
@@ -86,7 +86,7 @@ aoe add . --agent codex --model gpt-5       # pick an ACP agent + model (implies
 ## Requirements
 
 - aoe built with `--features serve`.
-- Node.js 22+ on `PATH` (the structured view spawns an ACP agent subprocess; `aoe-agent` declares `engines.node: >=22.0.0` and is started with `node --experimental-strip-types`).
+- Node.js 22+ on `PATH` (the structured view spawns an ACP agent subprocess; `aoe-agent` declares `engines.node: >=22.0.0`).
 - For Claude Code, a `claude login` session.
 
 If Node is missing or too old, the session falls back to the terminal view with an actionable warning. Verify with `aoe acp doctor`:
--- a/docs/structured-view.md
+++ b/docs/structured-view.md
@@ -29,7 +29,7 @@ aoe ships an ACP registry entry for each tool whose ACP server we've verified. F
 | `omp` | `omp acp` (native) | `curl -fsSL https://omp.sh/install \| sh` | provider environment or OMP login |
 | `kimi` | `kimi acp` (native) | `curl -fsSL https://code.kimi.com/kimi-code/install.sh \| bash` | `kimi login`, or provider env |
 | `prime-agent` | `prime-agent --mode acp` (native) | `curl -fsSL https://app.primeintellect.ai/prime-agent/install.sh \| sh` | `/login` once, or provider env |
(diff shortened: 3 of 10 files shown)


njbrake (approved):
Approved. Pushed a3c7e21 to this branch: the `MIN_NODE_MAJOR` comment and the structured-view requirement both attributed the Node 22 floor to `node --experimental-strip-types`, but the registry spawns the compiled `dist/aoe-agent` and that flag landed in 22.6.0, not 22.0. Both now point at `package_engines_matches_min_node_major`, which is the pin the gate tracks.

Two follow-ups, neither blocking:

Slash-qualified ids (`openai/gpt-5.5`) reach `pickModel` unvalidated and come back as an Anthropic model with no error. Same class as the `llama3.3:ollama` case, one layer down. Worth its own issue.

The floor is global to the structured view, not aoe-agent only. A Node 20 host now sees `aoe acp doctor` drop from `ok` to `partial` even when it only runs claude-code-acp. Harmless, since PATH is forwarded and Node 20 is EOL, but it belongs in the description.

Issue #3624, opened 2026-08-31
Pane status can remain Running when multiple frames share a tmux activity second


jerome-benoit:
## Problem

PR #3600 skips pane capture when tmux's `#{window_activity}` equals the value saved after the previous capture.

At exact head `a7c40f4e68112c6bff21068908199f52f115d159`:

- `src/session/instance/status_update.rs:423-438` treats an equal activity value as proof that the screen is unchanged.
- `src/session/instance/status_update.rs:489` stores that value after the capture.
- `src/tmux/mod.rs:934` reads `#{window_activity}`, an epoch value with one-second granularity.
- `src/tui/app.rs:868` polls status every 500 ms.

If a running frame is captured and the final idle frame is rendered later in the same second, both reads have the same activity value. The final frame is skipped, and no later output advances the value, so a hookless manifest agent can remain `Running` indefinitely.

This gate did not exist at the base of #3600 (`90bc4aa6ef4efa359692c402f4d6c8aeebf4ef99`).

## Proposed direction

Do not use equality of a one-second timestamp as the sole proof that no output arrived after a capture. For example, force a trailing capture after the activity second has elapsed before allowing reuse, while preserving the existing pending-detection and hook guards.

Add a regression that models two pane updates sharing one activity value and verifies that the final idle frame is observed and confirmed.

## AI usage

This issue was prepared from two independent static reviews and a final exact-SHA cross-validation. No local build or test was run; all CI checks on #3600 were green.

PR #3672, opened 2026-09-01
fix(status): carry detection state across every poll boundary

njbrake:
## Description

#3633 made a non-visible `Running -> Idle` wait for a second agreeing poll, and kept the state that waiting needs on `Instance` as `#[serde(skip)]` runtime fields. No poller owns the row it decides on, so that state was thrown away every cycle:

- The TUI poller mutates a *clone* and projects the result into `StatusUpdate`, which carried none of the detection fields.
- The daemon rebuilds every row from disk each tick; `seed_unknown_tracking` seeded only the Unknown-escalation pair, and `merge_runtime_fields` never saw these fields either.
- The TUI's own `merge_runtime_from_reload` carries the sibling Unknown-tracking fields but not these, so its 5s heartbeat reload could drop a proposal raised by the 500ms poller.

So every cycle started with no proposal on record. A hookless manifest agent proposed Idle, held Running awaiting confirmation, lost the proposal, and stayed Running for the life of the session, which is the failure #3624 was closed on. The capture-skip gate from #3600 was also dead for the same reason, since its activity stamp reset every cycle too.

The four fields are now one `DetectionState` value, so each boundary carries one field instead of four: `StatusUpdate` (with `None` meaning "this producer never detected", matching `IdleIntent::Keep`), the attached-session watcher, the daemon's per-tick `PriorTickTracking` seed plus the `DiskOnly` merge arm, and `merge_runtime_from_reload`.

Each regression was confirmed red on the unpatched carry: the daemon and TUI tick tests published `[Running, Running]` instead of `[Running, Idle]`, and the reload merge test lost its pending proposal.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No docs change: the detection state is internal, and the status behavior it restores is the one already documented.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: both regressions already drive a live tmux pane through the real production path, one per boundary, and neither needs the full binary. `poll_cycles_confirm_an_unwitnessed_idle_through_the_status_update` runs two real `poll_statuses_once` cycles against `HomeView` and applies each `StatusUpdate` back; `a_proposal_survives_the_tick_that_reloads_its_row_from_disk` runs four daemon ticks, each rebuilding the row from disk as `load_all_instances` does, the last two covering a tick that skips the capture. The reload-merge arms are covered in `tests/serve_disk_reload_helper_equivalence.rs`, and `merge_runtime_from_reload` in its own module.

## AI Usage

- [ ] No AI was used
- [x] AI was used

The issue was drafted by Codex from a static review; the claim was checked against both boundaries and confirmed by making each regression fail on the unpatched carry before landing it. Diagnosis, fix shape, and tests were reviewed by @njbrake.

## Summary

- Consolidates detection data into a single `DetectionState`.
- Preserves detection state across TUI updates, daemon polling, disk reloads, and runtime merges.
- Prevents pending `Running -> Idle` confirmations and capture activity from being lost between polls.
- Updates attached-session handling and structured daemon status updates.
- Adds regression tests for poll cycles, reload behavior, and capture skipping.

## Benefits

- Makes detection state easier to manage.
- Prevents incorrect status transitions after reloads.
- Improves reliability across TUI and daemon workflows.
- Provides coverage for key state-preservation paths.


jerome-benoit (approved):
Reviewed at `a4e3eecb37f651d95c0142ba6c6861cf1a426805`.

The fix consistently carries `DetectionState` across the TUI poller clone, attached-session watcher, daemon tick/disk-reload, and TUI reload boundaries. The `DiskOnly`/`TmuxApplied` split preserves the correct source of authority, and the added regressions cover both production paths that previously dropped a pending confirmation.

I found no blocking issues. Remote CI is green at this SHA; I did not rerun the suite locally.

Issue #1101, opened 2026-05-13
Cockpit: /clear empties agent context but UI keeps full transcript and stale capabilities (sibling to /compact #1050)


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Running `/clear` in cockpit resets the agent's context (claude's internal conversation history is wiped) but the cockpit UI keeps every prior message, tool call, and approval visible. The user reads the chat and sees a full conversation; the model on the other end remembers none of it.

This is the same shape of bug as #1050 (`/compact` has no visible effect), but worse: `/compact` summarises (the model still has *something* about earlier turns), `/clear` truly forgets. Any reply from the user that references the visible transcript is now talking past the model.

Closed-issue #1050's open questions even called this out: *"Does `/clear` (full reset, not summarise) need the same treatment? It probably arrives as a different system message subtype; worth a quick adapter scan. Likely yes, same UX gap, same shape of fix."* This issue follows up on that.

### Adapter passes `/clear` through as a plain user prompt

`/clear` is one of claude-agent-acp's exposed slash commands (`~/.../claude-agent-acp/dist/acp-agent.js:1796-1826` — `getAvailableSlashCommands` filters out `cost`, `keybindings-help`, `login`, `logout`, `output-style:new`, `release-notes`, `todos`, but **not** `clear`). When the user types `/clear`, it goes to the agent as a regular text prompt; the claude SDK handles it internally by ending its current internal session and starting a fresh one.

The adapter emits **no dedicated signal** when this happens. Searching `acp-agent.js` for clear / session-id-rotation handling yields nothing structured. The signals that do exist for similar boundaries (`compact_boundary` system message, `session_state_changed`) are not produced for `/clear`. From aoe's side it looks like a normal user prompt followed by a normal `Stopped`.

### aoe ingests it as a normal turn

- `src/cockpit/acp_client.rs` — no detection of `/clear` user prompts; the prompt text is published via `Supervisor::publish_user_prompt` (`src/cockpit/supervisor.rs:374`) and forwarded to the agent like any other.
- `src/cockpit/event_store.rs` — every prior `Event::*` for the session stays on disk. No row deletion, no boundary marker.
- `web/src/lib/cockpitTypes.ts:582-633` — the `UserPromptSent` reducer arm appends a new user row, clears `assistantMessage` and errors, sets `turnActive=true`. No special-case for `/clear` content.
[...]

PR #3676, opened 2026-09-01
fix(sv): rebuild the runtime on a fold change

### LINKED ISSUE #3640: Structured view: the /clear command stops the conversation view; the fault from #1119 is back
## Problem

The `/clear` command makes the conversation view stop. The view then shows the error boundary message: "Something went wrong. The error has been reported. Reload the page to retry."

The fault happened 3 times in 3 tries on v1.15.0. Release v1.15.1 has the same code and the same library versions, so the fault is also in v1.15.1.

The session does not stop. Only the display stops. These are the facts:

- The agent does the reset each time. The log shows `conversation reset: session/new succeeded, swapped acp_session_id`.
- The worker continues with the same PID. The command `aoe ps --acp` shows the state `attached`.
- A page reload repairs the view. The agent context is empty, as it must be.

So it breaks only the user interface. But the user must reload the page after each `/clear` command.

This is the error in `debug.log`. The browser shows only the boundary message. The boundary is at `web/src/components/ErrorBoundary.tsx:36-37`.

```
ERROR web.client: useClientLookup: index 7 out of bounds (length: 1) (ignore if recovered)
  client_target=react.errorboundary path=/session/<session-id>
  stack=Error: useClientLookup: index 7 out of bounds (length: 1) (ignore if recovered)
    at Object.c [as get] (/assets/StructuredView-<hash>.js)
    at Object.jo [as useSyncExternalStore] (/assets/index-<hash>.js)
    at e.useSyncExternalStore (/assets/react-<hash>.js)
```

Three `/clear` commands gave these values: index 7 with length 1, then index 3 and index 5 with length 2.

## This fault is not new

Issue #1119 is the same type of fault. At that time the name of the function was `tapClientLookup`, in the package `@assistant-ui/tap`.

Pull request #1400 closed issue #1119 on 2026-05-22. That pull request changed `@assistant-ui/react` from 0.12.28 to 0.14.7. The new version included two corrections from the library project:

- assistant-ui/assistant-ui#4069, for the re-key of derived scopes.
- assistant-ui/assistant-ui#4077, for the key of part fibers by absolute index.

The test list in pull request #1400 included the `/clear` command. So the command worked at that time.

There is no open issue for this fault now. The text `useClientLookup` gives no results in the issues and the pull requests of this project.

## The versions

The correction for issue #1119 came in `@assistant-ui/react` 0.14.7. All versions in the 0.14 group have it.

| Release | `@assistant-ui/react` | `@assistant-ui/store` |
| --- | --- | --- |
| v1.13.2 | 0.14.27 | 0.2.x |
| v1.14.0 | 0.14.28 | 0.2.21 |
| v1.14.1 | 0.14.28 | 0.2.21 |
| v1.15.0 | 0.15.14 | 0.3.9 |
| v1.15.1 | 0.15.14 | 0.3.9 |

Release v1.15.0 moved to the 0.15 group. Pull request #3434 made the last part of this move, from 0.15.1 to 0.15.14.

Release v1.15.1 does not correct the fault. The library versions in v1.15.1 are the same as in v1.15.0. The file `AcpRuntime.tsx` has no change between the two releases. The file `ErrorBoundary.tsx` also has no change.

The fault is in the 0.15 group. So v1.14.1 is the last release without the fault. This conclusion comes from the versions only. Nobody has done a test with v1.14.1 to confirm it.

## The current code

The function `activityToThreadMessages` hides the turns above the last `session_cleared` divider. See `web/src/components/acp/AcpRuntime.tsx:323`. The purpose is to stop a reply on top of a transcript that the model forgot. See issue #1101.

The function makes the list of rows shorter:

```ts
if (lastClearIndex >= 0) {
  effectiveRows = rows.slice(lastClearIndex);
}
```

See `web/src/components/acp/AcpRuntime.tsx:345` in v1.15.1. The default value of `showClearedTurns` is false. See the comment at `AcpRuntime.tsx:68-74`. Then a `/clear` command changes the message list from N items to 1 item.

## Cause

The message list becomes shorter. But the assistant-ui child components are still on the display. These components use the old index values.

These are the steps:

1. The `/clear` command adds a `session_cleared` row. The agent does the reset.
2. The fold removes the rows above the divider. The list changes from N items to 1 item.
3. The child components read their position again with `useClientLookup.get({ index })`.
4. This function makes an error if `index` is more than `resources.length`, or equal to it. See `@assistant-ui/store/src/useClientLookup.ts:46`.
5. The error occurs in `useSyncExternalStore`. React cannot recover from an error at this position. The error goes to the error boundary. The boundary then removes the conversation view.

The index values agree with this cause. Index 7 is message 8 of the old transcript. The program read this index after the list became 1 item long.

Issue #1119 is different in one way. That fault did not occur frequently, because it needed a specific order of events. This new fault occurs each time. The fold changes the list from N items to 1 item in one step.

## The state of the correction in the library

The library project corrected one part of this fault. Report assistant-ui/assistant-ui#5704 shows the same message text. Pull request assistant-ui/assistant-ui#5707 corrects by-index scopes when a collection becomes shorter. A shorter collection is our condition. The library project merged that pull request on 2026-08-07.

Our version has this correction. The library published store 0.3.7 on 2026-08-08. Our version is store 0.3.9, from 2026-08-12. So a new library version does not give us this correction, because we have it.

One part of the fault is still open. Report assistant-ui/assistant-ui#5708 is open. Its title is "tap: useSyncExternalStore swallows snapshots that throw on a notification, diverging from React".

That report names our exact path. It says the errors include "the transients the store itself labels `(ignore if recovered)` (`packages/store/src/useClientLookup.ts`), in paths that have no absorber". It also says a first correction went in and then came out again, because the effect was too large.

So the library has no correction for us at this time. This is not the same as issue #1119. At that time the library had a correction, and a new version was enough.

## Possible corrections

**Option A. Remove the old child components. This is the best option.**

Give the thread component a key value from the clear generation. Use the id of the last `session_cleared` row.

React then removes the old child components when the fold point changes. React does not show them again with a shorter list. This option removes the cause in this project. The change is only in the structured view code.

The library project moves in the same direction. Pull request assistant-ui/assistant-ui#6495 keys thread messages by message id, not by index.

**Option B. Let the error boundary recover.** The library message includes the words "ignore if recovered". Let the boundary try the display one more time for this type of error. Then the user does not reload the page. This option is useful with option A, or without it.

**Option C. Do not make the list shorter.** Show all the messages, and hide the old rows in the display only. The list is then never shorter than before. This change is large. It also has an effect on the activity limit. Option A is better.

**A new library version is not sufficient.** The correction for a shorter collection is already in our version. The other part, report #5708, is open. So this project must make the correction.

## Temporary method for the user

In the `ClearedTurnsBanner`, set "show cleared turns" to on. The program then does not do the fold, because of the test `if (!showClearedTurns)`. The list becomes longer by one item. A longer list cannot make this error.

I did not test this method in a browser. I know it only from the code.

## Tests

The file `web/AGENTS.md` gives a rule for a change to the structured view. A correction needs a new entry in `web/tests/coverage-matrix.json`. A correction also needs a test.

Two tests are possible:

- A Vitest test of the function `activityToThreadMessages`. The test makes sure that the fold does not make the list shorter while child components are on the display.
- A live Playwright test. The test sends a `/clear` command to the fake ACP agent.

Both tests fail with the current code.

## Environment

- aoe v1.15.0. Build from source with `--features serve`.
- The command `aoe serve` on Linux. Default port. Loopback address.
- Chrome 151 on macOS.
- Agent: `claude-agent-acp`.

## Procedure to repeat the fault

1. Open a structured view session.
2. Send prompts to the agent, until the transcript has more than 3 messages.
3. Send the `/clear` command.
4. Look at the conversation view. The view shows "Something went wrong".
5. Look at `debug.log`. The log has the message `web.client: useClientLookup: index N out of bounds`.
6. Do the command `aoe ps --acp`. The worker is still in the state `attached`, with the same PID.

Suggested labels: P1, area:web

xande:
## Description

Fixes #3640. A `/clear` put the conversation view into the ErrorBoundary, and the view showed "Something went wrong". The session was never harmed: the reset succeeded and a page reload repaired the view.

The cause is at part level. The fold hides the turns above the last `session_cleared` row (#1101). That makes the part list of a message the fold keeps shorter. assistant-ui reads parts by absolute index, and its store keeps the committed list when a snapshot throws (upstream assistant-ui#5708). A stale part child then reads past the end, inside `useSyncExternalStore`, where React cannot recover. The error then goes to the ErrorBoundary.

The stale state is in the store, not in the children. `AcpRuntime` now holds the external-store hook in a small `RuntimeHost` component, and keys it on the id of the last `session_cleared` row. A move of the fold point builds a fresh store. An ordinary turn does not change the key, so it does not rebuild.

`lastClearIndex` is now one exported helper in `web/src/lib/acpHistoryWindow.ts`. The scan had existed in three places: that module already held a private copy, and `StructuredView` inlined a third for the banner's hidden-turn count. The message tree, the runtime key, and the banner now all read the one helper, so the fold cannot be measured two ways.

I confirmed the cause and the correction on a real session. The steps are in [this comment](https://github.com/agent-of-empires/agent-of-empires/issues/3640#issuecomment-5494429009). The transcript needs a message with more parts than the message that survives the fold. Each `/clear` also makes the transcript shorter, so the next `/clear` has no high part index left. That is why the fault looks intermittent.

The failure and the correction, from the same start shape:

```
crash: gen=cleared-3236 messages=4 parts=[1,1,1,20] -> messages=1 parts=[1]
       -> ERROR web.client: useClientLookup: index 5 out of bounds (length: 1)
fixed: gen=cleared-3236 messages=4 parts=[1,1,1,20] -> gen=cleared-3570 messages=2 parts=[1,1]
       -> no error, divider renders, composer works
```

The second commit carries the consolidation above plus three tidy-ups: assistant-ui's public `ExternalStoreAdapter<T>` in place of a hand-rolled `Parameters<typeof useExternalStoreRuntime<T>>[0]` alias, a session stub in the spec cut to the `state` the runtime actually reads (I re-checked that the rebuild assertion still fails when the key is removed), and one copy of the mechanism comment on `RuntimeHost` instead of three.

Notes for the reviewer:

- I first keyed only the message list. That variant is not enough. I measured it against a real session and it still failed, because the stale state is in the store. The history of that mistake is in the PR comments.
- A rebuild remounts the subtree under the provider, so the transcript loses its scroll position on a `/clear`. Text drafts persist to storage, so they survive.
- The fault occurs in the production build only. The React development build recovers from the throw, so `cargo xtask dev` does not show it.
- The library fault is still open: assistant-ui/assistant-ui#5708. This PR removes our trigger. It does not repair the library.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: I added a Vitest spec and a `coverage-matrix.json` entry (`acp.clear-fold-remount`), but no Playwright spec. I wrote five live Playwright specs and ran each against a build with the correction removed. All five passed, so none of them reproduced the fault, and I removed all five. The fake ACP agent does not render the same parts as the real agent. jsdom is also blind to the failure, because it swallows the throw and keeps the stale list (assistant-ui#5708). The Vitest spec therefore covers the key: its value, and the fact that a clear moves it and an ordinary turn does not. I verified that the spec fails if the code does not move the key. The browser check is the manual procedure on #3640, which I ran before and after the correction. A reviewer may want more, and I can continue the search for a harness repro.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Code, model Opus 5.

Claude read the logs, measured the transcript shapes with temporary instrumentation, ran the experiments on a real session, and wrote the code and this text. I chose what to test, what to keep and what to file. Claude also wrote five Playwright specs that did not reproduce the fault; I had it delete them rather than keep a test that cannot fail.

### DIFF
--- a/web/src/components/acp/AcpRuntime.tsx
+++ b/web/src/components/acp/AcpRuntime.tsx
@@ -27,7 +27,12 @@
 // rattle spinner, ApprovalCard) and slot them into assistant-ui's
 // component-injection points.
 
-import { AssistantRuntimeProvider, useExternalStoreRuntime, type ThreadMessageLike } from "@assistant-ui/react";
+import {
+  AssistantRuntimeProvider,
+  useExternalStoreRuntime,
+  type ExternalStoreAdapter,
+  type ThreadMessageLike,
+} from "@assistant-ui/react";
 import { useCallback, useEffect, useMemo, useRef, useState, type ReactNode } from "react";
 
 import { useAcpSession } from "../../hooks/useAcpSession";
@@ -42,6 +47,7 @@ import type {
 import { hasTodoArrayArgsText, parseJsonObject } from "../../lib/acpArgs";
 import { clearDraft, getDraftAttachments, setDraftAttachments } from "../../lib/acpDrafts";
 import { useHistoryWindow } from "../../hooks/useHistoryWindow";
+import { lastClearIndex } from "../../lib/acpHistoryWindow";
 import { canOfferEarlier, earlierAction } from "../../lib/historyScroll";
 import { useAgentProfile } from "../../lib/agentProfileContext";
 import { type AgentProfile, DEFAULT_AGENT_PROFILE, isSubagentToolName } from "../../lib/agentProfiles";
@@ -123,19 +129,15 @@ export interface AcpContext {
   loadingEarlierHistory: boolean;
 }
 
-type ExternalStoreAdapter = Parameters<typeof useExternalStoreRuntime<ThreadMessageLike>>[0];
-
-/** Owns the external-store hook, so a change of `key` rebuilds the runtime.
+/** Owns the external-store hook so a change of `key` builds a fresh runtime.
  *
- *  The fold shortens the part list of a message it keeps. assistant-ui reads
- *  parts by absolute index, and its store keeps the committed list when a
- *  snapshot throws (upstream assistant-ui#5708). A stale part child then reads
- *  past the end, inside useSyncExternalStore, where React cannot recover. The
- *  ErrorBoundary then replaces the view.
- *
- *  The stale state is in the store, not in the children, so a remount of the
- *  message list is not enough. Only a new store fixes it. See #3640. */
-function RuntimeHost({ adapter, children }: { adapter: ExternalStoreAdapter; children: ReactNode }) {
+ *  A `/clear` shortens the part list of a message the fold keeps, and
+ *  assistant-ui reads parts by absolute index while keeping the committed list
+ *  when a snapshot throws (assistant-ui#5708). The stale part child then reads
+ *  past the end inside useSyncExternalStore, where React cannot recover, and
+ *  the ErrorBoundary replaces the view. The stale state lives in the store, so
+ *  remounting the message list is not enough. See #3640. */
+function RuntimeHost({ adapter, children }: { adapter: ExternalStoreAdapter<ThreadMessageLike>; children: ReactNode }) {
   const runtime = useExternalStoreRuntime<ThreadMessageLike>(adapter);
   return <AssistantRuntimeProvider runtime={runtime}>{children}</AssistantRuntimeProvider>;
 }
@@ -253,7 +255,7 @@ export function AcpRuntime({
     [displayActivity, showClearedTurns],
   );
 
-  const adapter: ExternalStoreAdapter = {
+  const adapter: ExternalStoreAdapter<ThreadMessageLike> = {
     messages,
     isRunning: acp.state.turnActive,
     convertMessage: (m) => m,
@@ -332,21 +334,9 @@ export function AcpRuntime({
   );
 }
 
-/** Index of the row the fold starts at, or -1 if no `/clear` happened yet.
- *  The fold pins to the LAST clear, so repeated /clears collapse
- *  cumulatively. See #1101. */
-function lastClearIndex(rows: readonly ActivityRow[]): number {
-  for (let i = rows.length - 1; i >= 0; i -= 1) {
-    if (rows[i]!.kind === "session_cleared") return i;
-  }
-  return -1;
-}
-
-/** Identity of the fold point, for use as a React `key`.
- *
- *  The value changes only when the fold point moves, so an ordinary turn does
- *  not remount. `"all"` means nothing is folded; `"none"` means no `/clear`
- *  happened yet. Both are stable for the life of the session. See #3640. */
+/** React `key` identifying the current fold point: the id of the last `/clear`,
+ *  `"none"` before the first one, `"all"` when nothing is folded. Only a moved
+ *  fold changes it, so an ordinary turn does not rebuild the runtime. */
 export function clearFoldGeneration(rows: readonly ActivityRow[], showClearedTurns: boolean): string {
   if (showClearedTurns) return "all";
   const i = lastClearIndex(rows);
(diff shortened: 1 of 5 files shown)


njbrake (approved):
Approving.

Two notes, neither blocking.

The key reads the history window, not the full activity, so it also moves when the `/clear` divider enters or leaves that window: a "Load earlier" crossing the divider rebuilds the runtime, and so does the next user prompt when the window cut lands on the divider. Both are real fold moves, so the rebuilds are correct. The comment on `clearFoldGeneration` promising that only a `/clear` moves the key is what needs correcting.

The same part-list collapse is reachable with no `/clear`: when a tail turn exceeds the 150-row window, `historyWindowStart` snaps past the cut on the next prompt and the rendered shape goes from one message of 150 parts to one of 1, key unchanged. Pre-existing and out of scope here, but worth its own issue.

PR #3679, opened 2026-09-01
fix(sandbox)!: harden repo-scoped sandbox config

unbrice:
## Description
Security fix, for issues discovered during the review of #3623

As flagged there by @jerome-benoit , `selinux_relabel` shouldn't be overridable on a per-repo basis, and `sandbox.network`  was only validating settings on interactive edits, allowing for escapes. Also before this PR, the TUI would offer to edit `container_runtime` but that would be ignored. And finally, a repo could run a host command by overriding `default_tool`.

There remain potential issues in related code:
- `sandbox.environment` allows inheriting env variables from outer context, arguably the per-repo overrides should be limited to inline assignments.
- `worktree.path_template` / `bare_repo_path_template` / `workspace_path_template` allow path escape.

I'm not knowledgeable enough to say if either or both of these things are violation of the security model of AOE, but flagging them as fishy.

## PR Type

- [ ] New Feature
- [X] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [X] N/A: this PR doesn't change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [X] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle

## AI Usage

- [ ] No AI was used
- [X] AI was used

**AI Model/Tool used:**
I forked some of these changes from #3623 which was authored by Kimi and Claude under my supervision.

**NOTE:**

## Summary

- Hardened repository configuration by blocking unsafe overrides of sandbox and session settings.
- Prevented repositories from selecting user-defined host commands through `session.default_tool`.
- Validated `sandbox.network` values and rejected namespace-sharing and invalid formats.
- Added schema metadata for repository override policies, with safer defaults for new settings.
- Clarified that `container_runtime` and `selinux_relabel` are global-only settings.
- Updated documentation, templates, and regression tests.
- Cached the settings schema for simpler and more efficient access.

## Benefits

These changes strengthen the repository trust boundary and prevent unsafe configuration from reaching sandbox creation. Future work could address inherited environment variables and path-template escape risks.

### DIFF
--- a/src/session/config/repo_config.rs
+++ b/src/session/config/repo_config.rs
@@ -79,6 +79,10 @@ const REPO_OVERRIDABLE_SECTIONS: &[&str] = &["hooks", "session", "sandbox", "wor
 /// default means a field added to `SessionConfig` later stays repo-denied
 /// until someone marks it `repo = "allow"`.
 ///
+/// Even `default_tool` is denied despite only naming a tool: the launch path
+/// exact-matches the user's `custom_agents` before built-ins, so a repo could
+/// select a user-defined host command (see the field's doc comment).
+///
 /// Every other overridable section is a genuine team-shared surface, so its
 /// fields default to allowed and dangerous ones opt out with `repo = "deny"`
 /// on the field (e.g. the sandbox image and mount fields).
@@ -1688,7 +1692,7 @@ pub const INIT_TEMPLATE: &str = r#"# Agent of Empires - Repository Configuration
 # on_destroy = ["docker-compose down"]
 
 # [session]
-# default_tool = "claude"
+# agent_detect_as = { my-agent = "claude" }
 
 # [sandbox]
 # enabled_by_default = true
@@ -2035,9 +2039,36 @@ mod tests {
             merged.session.custom_agents.is_empty(),
             "repo-declared custom_agents must be dropped on merge"
         );
-        // `default_tool` stays repo-overridable (the documented use case); it
-        // can only name a tool the user configured, never a command.
-        assert_eq!(merged.session.default_tool.as_deref(), Some("repo-agent"));
+        assert_eq!(
+            merged.session.default_tool, None,
+            "repo-declared default_tool must be dropped on merge"
+        );
+    }
+
+    #[test]
+    fn test_repo_config_cannot_select_a_user_custom_agent() {
+        // The launch path exact-matches the user's `custom_agents` before
+        // built-ins when resolving `default_tool` (cli/add.rs), so a repo
+        // pointing it at a user-defined custom agent would launch that host
+        // command. Validating the value against built-in names would not
+        // close the hole because a custom agent may shadow a built-in name,
+        // so the field is repo-denied outright.
+        let mut global = Config::default();
+        global
+            .session
+            .custom_agents
+            .insert("deploy-helper".to_string(), "do-deploy --prod".to_string());
+        let repo: RepoConfig =
+            toml::from_str("[session]\ndefault_tool = \"deploy-helper\"\n").unwrap();
+        let merged = merge_repo_config(global, &repo);
+        assert_eq!(
+            merged.session.default_tool, None,
+            "repo must not select the user's custom agent"
+        );
+        assert!(
+            merged.session.custom_agents.contains_key("deploy-helper"),
+            "the user's own custom_agents are untouched"
+        );
     }
 
     #[test]
@@ -2071,8 +2102,8 @@ mod tests {
         assert!(merged.session.smart_rename_agent.is_empty());
         assert!(merged.session.smart_rename_model.is_empty());
         assert!(!merged.session.yolo_mode_default);
-        // The two allowed fields still come through.
-        assert_eq!(merged.session.default_tool.as_deref(), Some("codex"));
+        assert_eq!(merged.session.default_tool, None);
+        // The one allowed field still comes through.
         assert_eq!(
             merged.session.agent_detect_as.get("x").map(String::as_str),
             Some("claude")
@@ -2087,6 +2118,7 @@ mod tests {
                 "session.agent_config_dir",
                 "session.agent_extra_args",
                 "session.custom_agents",
+                "session.default_tool",
                 "session.smart_rename_agent",
                 "session.smart_rename_model",
                 "session.yolo_mode_default",
@@ -2104,7 +2136,7 @@ mod tests {
         fs::create_dir_all(&config_dir).unwrap();
         fs::write(
             config_dir.join("config.toml"),
-            "[session]\ncustom_agents = 7\ndefault_tool = \"codex\"\n",
+            "[session]\ncustom_agents = 7\nagent_detect_as = { my-agent = \"claude\" }\n",
         )
         .unwrap();
 
@@ -2112,7 +2144,14 @@ mod tests {
             .expect("a wrongly-typed denied field must not fail the load")
             .expect("config exists");
         let merged = merge_repo_config(Config::default(), &loaded);
-        assert_eq!(merged.session.default_tool.as_deref(), Some("codex"));
+        assert_eq!(
+            merged
+                .session
+                .agent_detect_as
+                .get("my-agent")
+                .map(String::as_str),
+            Some("claude")
+        );
         assert!(merged.session.custom_agents.is_empty());
     }
 
@@ -2124,7 +2163,11 @@ mod tests {
         let profile = ProfileConfig {
             description: None,
             overrides: serde_json::from_value(serde_json::json!({
-                "session": { "custom_agents": { "x": "sh -c pwned" }, "default_tool": "codex" },
+                "session": {
+                    "custom_agents": { "x": "sh -c pwned" },
+                    "default_tool": "codex",
+                    "agent_detect_as": { "my-agent": "claude" },
+                },
                 "acp": { "auto_approve": true },
             }))
             .unwrap(),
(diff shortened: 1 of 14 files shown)


src/session/config.rs
```diff
@@ -899,12 +899,14 @@ pub struct AppStateConfig {
...
-    /// unavailable, falls back to the first available tool.
+    /// unavailable, falls back to the first available tool. Only names a tool
+    /// the user configured (built-in or from their own `custom_agents`).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[setting(
         label = "Default Tool",
         widget = "custom:default-tool",
-        category = "Agents"
+        category = "Agents",
+        repo = "allow"
```
jerome-benoit: [P1] Keep repo defaults from selecting user-defined host commands. Although a repo cannot define `custom_agents`, `repo = "allow"` lets it choose the name of an existing user-configured custom agent; `aoe add` then prefers that exact match and the builder launches its host command. A checked-out repository should not be able to select that command implicitly. Please either deny `default_tool` at repo scope, or preserve the built-in-tool feature by validating repo values against built-in agents only, and add a regression test proving that a repo cannot select a local `custom_agents` entry.

jerome-benoit (dismissed):
Reviewed after the author split the pre-existing repo-policy hardening out of #3623. The network namespace validation and SELinux repo-scope fixes are valid at this head. One trust-boundary issue remains in the policy migration and belongs to this PR; I have placed it inline here rather than continuing to attribute it to #3623.

Evidence at `94771e559f46e832b59fc367688d12d8c7bc3cee`: `git/gh diff` shows this PR adds the explicit `repo = "allow"` declaration at `src/session/config.rs:909`; repo sanitization preserves it at `src/session/repo_config.rs:462-497`; default selection prefers an exact local `custom_agents` name at `src/cli/add.rs:1534-1549`; and session construction resolves and launches that command at `src/session/builder.rs:750-780` and `src/session/config.rs:1708-1718`. The behavior predates this PR, but this is now the PR that owns and codifies the repo trust policy.

The full CI suite is not present yet for this head; GitHub currently reports only the successful PR-title check.

jerome-benoit (dismissed):
Re-review at exact head `508d0a0e2e1a8df5385a62ff9880abb0971e6fca`.

The previous `default_tool` blocker is fixed: `src/session/config.rs:901-914` now marks `session.default_tool` as `repo = "deny"`, and the repo-policy tests now expect `session.default_tool` to be rejected. That part should not keep blocking this PR.

Two repo trust-boundary blockers remain:

- [P1] `sandbox.environment` is still repo-overridable. CodeRabbit already placed this inline at https://github.com/agent-of-empires/agent-of-empires/pull/3679#discussion_r3907155798, and I am not duplicating the inline comment. The code confirms the issue: `src/session/config.rs:2244-2259` documents and exposes host passthrough forms (`KEY`, `KEY=$VAR`) without `repo = "deny"`; `src/session/repo_config.rs:487-501` falls back to the section default; `section_default_allows("sandbox")` is true via `src/session/repo_config.rs:86-90`; and `collect_environment()` resolves both `KEY=$VAR` and bare `KEY` from the host at `src/session/environment.rs:601-606` and `src/session/environment.rs:617-625`. A repo-controlled config can therefore request selected host secrets inside the sandbox. Fix: deny `sandbox.environment` at repo scope, or split the type so repo config accepts only literal `KEY=value` entries and rejects every host passthrough form, with regression tests on the merge/build path.

- [P1] Worktree filesystem controls are still repo-overridable by default. `src/session/config.rs:2084-2142` exposes `worktree.enabled`, `path_template`, `bare_repo_path_template`, and `workspace_path_template` without `repo = "deny"`; `src/session/repo_config.rs:86-90` allows non-session sections by default; and the current test still asserts `("worktree", "path_template", true)` at `src/session/repo_config.rs:2180-2187`. That lets a checked-out repo both enable worktree mode by default and choose the host checkout path. The path renderer accepts absolute templates and lexical `..` escapes (`src/git/template.rs:61-75`), then `aoe add` passes the resulting path directly to `git worktree add` (`src/cli/add.rs:546-565`, `src/git/worktree.rs:840-846`). For multi-repo attach/workspace conversion, `workspace_path_template` is resolved the same way (`src/session/attach_project.rs:319-324`) and then used as the parent for new/moved worktrees (`src/session/attach_project.rs:637-664`). Impact: an untrusted repo can make the next worktree session create a checkout under an attacker-chosen absolute or escaped host path, including a previously non-existing directory. Since checkout also materializes repo-controlled files and may run post-checkout hooks/submodule initialization on that path (`src/git/worktree.rs:863-918`), this should not be repo policy. Fix: make worktree filesystem fields repo-denied by default, and only allow narrowly safe team-shared fields if there is a proven containment rule. Add regression tests for repo config with `enabled = true`, absolute templates, and `../..` templates being dropped.

Checked and rejected: `sandbox.network` namespace-sharing and `selinux_relabel` are fixed at this head. `sandbox.enabled_by_default`, `default_image`, `extra_volumes`, `mount_ssh`, `selinux_relabel`, and `container_runtime` are now repo-denied/global-only where expected.

Evidence: `git diff --check e51297c4e7d9d5dedf0e362525c9bb2dc1263afe..508d0a0e2e1a8df5385a62ff9880abb0971e6fca` produced no output, and `cargo fmt --all -- --check` exited successfully. GitHub currently reports PR-template/title checks only for this head, and no exact-head Rust test result. GitHub also reports `mergeStateStatus=DIRTY`.

src/session/config/repo_config.rs
```diff
@@ -2132,22 +2172,23 @@ mod tests {
...
-        // across every scope the sanitizer / TUI query at runtime. tmux and
-        // sound rows guard #3229: their sections are not repo-overridable, so
-        // every field on them must answer false.
+        // (section, field, expected): field-declared repo policy and section defaults.
         let cases = [
-            ("session", "default_tool", true),
+            ("session", "default_tool", false),
             ("session", "agent_detect_as", true),
             ("sandbox", "memory_limit", true),
             ("worktree", "path_template", true),
```
jerome-benoit, line 2166: [P1] Keep repo templates inside a user-approved worktree root

Leaving `worktree.path_template` repo-allowed gives checked-in config an unprompted host-filesystem write primitive. Repo config can also set `worktree.enabled = true`; the TUI adopts that default, `resolve_template()` accepts absolute paths, and `git worktree add` creates a destination that does not exist. A safe reproduction with a tracked `payload.desktop` and an absent `$XDG_CONFIG_HOME/autostart` produced `.git` and `payload.desktop` directly in that directory. The XDG autostart specification states that such desktop files are launched after login: https://specifications.freedesktop.org/autostart/0.5/

Destination-exists checks do not prevent this case because the sensitive directory is intentionally absent before creation. Please repo-deny `path_template`, `bare_repo_path_template`, and `workspace_path_template`, or enforce that every repo-sourced resolved path stays under a user-approved canonical root before `create_dir_all` or `git worktree add`. Add a regression test for both absolute paths and `..` escape.

jerome-benoit (dismissed):
Re-review at exact head `508d0a0e2e1a8df5385a62ff9880abb0971e6fca`, cross-validated through two independent sequential analyses and final code-level arbitration.

The previous `default_tool` P1 is fixed: repo scope now denies the field, the TUI hides it, and the regression test preserves the user's custom agent without letting the repo select it. The network sanitizer also correctly rejects host and namespace-sharing modes.

Two P1 trust-boundary blockers remain:

- I independently confirmed the existing [`sandbox.environment` finding](https://github.com/agent-of-empires/agent-of-empires/pull/3679#discussion_r3907155798). Repo-provided bare keys and `$VAR` references are resolved from the host by `collect_environment()` and delivered to code running in the sandbox. Please either deny this field at repo scope or permit only explicit literal assignments from repo config. I am not duplicating that inline.
- Repo-controlled worktree path templates permit checkout into an arbitrary absent user-writable host directory. This is detailed inline.

The branch is currently conflicting with `main`. At this exact head GitHub exposes only successful title/template checks, not Rust CI. `git diff --check d90e5531c45f04742824d596d8acb34eac5a7133..508d0a0e2e1a8df5385a62ff9880abb0971e6fca` produced no output, and `cargo fmt --all -- --check` exited successfully.

jerome-benoit (commented):
Scope correction at exact head `508d0a0e2e1a8df5385a62ff9880abb0971e6fca`.

After re-reading the author's response and the documented repository-config contract, I agree that the remaining `sandbox.environment` and worktree-path concerns should not block this targeted security fix. Both behaviors predate this PR, are explicitly documented, and are neither introduced nor worsened here. They remain valid P2 security follow-ups, now tracked in #3710 and #3711.

The targeted `selinux_relabel`, network-namespace, and `default_tool` fixes remain valid at this head. One wording correction: the new policy is default-deny for `session`, but fields with `RepoPolicy::Unspecified` in other repository-overridable sections still default to allowed.

I am dismissing my earlier `REQUEST_CHANGES` reviews as over-scoped. This is not an approval yet: the branch currently conflicts with `main`, and GitHub has no substantive Rust CI result for this exact head. Please resolve the conflicts and run the full CI; the resulting head should then be re-reviewed.

src/session/config/repo_config.rs
```diff
@@ -488,16 +479,26 @@ fn sanitize_repo_overrides(
...
+    let Some(desc) = super::settings_schema::descriptor(section, field) else {
+        return section_default_allows(section);
+    };
+    if !desc.profile_overridable {
+        return false;
+    }
+    match desc.repo_policy {
+        RepoPolicy::Allow => true,
+        RepoPolicy::Deny => false,
+        RepoPolicy::Unspecified => section_default_allows(section),
```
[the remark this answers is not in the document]
jerome-benoit: Valid concern, but pre-existing and outside the blocking scope of this targeted PR. Tracked as P2 follow-up #3710.

src/session/config/repo_config.rs
jerome-benoit: Valid concern, but pre-existing and outside the blocking scope of this targeted PR. Tracked as P2 follow-up #3711.

src/session/config.rs
jerome-benoit: Fixed and verified at `508d0a0e2e1a8df5385a62ff9880abb0971e6fca`: repository scope now denies `session.default_tool`, with regression coverage. Resolving.

njbrake (approved):
Approving.

The repo-override policy now lives on the field and its section instead of two hardcoded lists, which is the mechanism AGENTS.md already mandates, and each new denial carries a regression test.

Rebased onto main and added four commits: a stale `smart_rename` fixture that still asserted a repo-set `default_tool` (the CI failure), two container-runtime guides offering a profile override the field never honored, and a refactor moving the section default onto `#[setting_section(repo_default = ...)]` so `section_default_allows` could go. Undescribed keys in schema-backed sections now fail closed; `hooks` stays open.

Title carries `!` for the `default_tool` removal.

The pre-existing `sandbox.environment` and worktree path-template gaps stay tracked in #3710 and #3711.

PR #3680, opened 2026-09-01
fix: keep short lead segments whole in profile short codes

BTForIT:
## Problem

With `row_tag = "profile"`, the short code takes one initial per `-`/`_` segment, so sibling profiles that share an initial collapse to the same tag: `bsc-main` / `bso-main` → `bm` / `bm`, `gna-main` → `gm`, `p9-main` → `pm`. On a registry with many short-prefixed profiles the tag stops distinguishing them.

## Change

Keep a lead segment of up to three characters whole, then take initials of the remaining segments, capped at four characters (lower-cased as before):

| profile | before | after |
|---|---|---|
| `gna-main` | `gm` | `gnam` |
| `bsc-main` / `bso-main` | `bm` / `bm` | `bscm` / `bsom` |
| `RAS-Main` | `rm` | `rasm` |
| `p9-main` | `pm` | `p9m` |
| `forit-backup` | `fb` | `fb` (unchanged — lead longer than 3) |

Single-segment names are unchanged. One function in `src/tui/home/render.rs`; unit tests cover the table above.

## Testing

`cargo test --lib profile_short_code` passes; verified live on a 17-profile registry (tags render as `[gnam] [caym] [p9m ] [xcem] [rasw]`).

## Summary

- Improved TUI profile short-code generation to preserve short lead segments and reduce collisions.
- Lowercased short codes and limited them to four display cells.
- Updated row-tag padding to handle wide and combining characters correctly.
- Added the public `prefix_within_width` helper for grapheme-safe text limits.
- Added tests for Unicode width, grapheme clusters, case changes, and short-code behavior.
- Single-segment profile names remain unchanged.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the change is one pure function (`profile_short_code`) with unit tests covering the before/after table; no session lifecycle or rendering-path change for an e2e to exercise

## AI Usage

- [ ] No AI was used
- [x] AI was used

### DIFF
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -709,11 +709,23 @@ pub(crate) fn profile_short_code(profile: &str) -> String {
     let code: String = match segments.as_slice() {
         [] => String::new(),
         [single] => single.chars().take(3).collect(),
-        many => many
-            .iter()
-            .filter_map(|s| s.chars().next())
-            .take(4)
-            .collect(),
+        [lead, rest @ ..] => {
+            // Keep a short lead segment whole (`gna-main` -> `gnam`,
+            // `bsc-main`/`bso-main` -> `bscm`/`bsom`) so sibling profiles
+            // that share an initial stay distinguishable at a glance; a
+            // longer lead still contributes its initial only
+            // (`forit-backup` -> `fb`).
+            let lead_part: String = if lead.chars().count() <= 3 {
+                (*lead).to_string()
+            } else {
+                lead.chars().take(1).collect()
+            };
+            lead_part
+                .chars()
+                .chain(rest.iter().filter_map(|s| s.chars().next()))
+                .take(4)
+                .collect()
+        }
     };
     code.to_lowercase()
 }
@@ -4913,7 +4925,19 @@ mod tests {
     fn profile_short_code_multi_segment_takes_initials() {
         assert_eq!(profile_short_code("forit-backup"), "fb");
         assert_eq!(profile_short_code("pivot-main"), "pm");
-        assert_eq!(profile_short_code("wma-work"), "ww");
+        assert_eq!(profile_short_code("connect_airlines-work"), "caw");
+    }
+
+    #[test]
+    fn profile_short_code_keeps_short_lead_segment_whole() {
+        assert_eq!(profile_short_code("gna-main"), "gnam");
+        assert_eq!(profile_short_code("bsc-main"), "bscm");
+        assert_eq!(profile_short_code("bso-main"), "bsom");
+        assert_eq!(profile_short_code("RAS-Main"), "rasm");
+        assert_eq!(profile_short_code("aoe-fiw"), "aoef");
+        assert_eq!(profile_short_code("wma-work"), "wmaw");
+        assert_eq!(profile_short_code("p9-main"), "p9m");
+        assert_eq!(profile_short_code("bp-main"), "bpm");
     }
 
     #[test]
--- a/src/tui/home/render.rs
+++ b/src/tui/home/render.rs
@@ -18,8 +18,8 @@ use crate::session::config::{GroupByMode, RowTagMode, SortOrder};
 use crate::session::{Item, Status};
 use crate::tui::components::preview::{self, CachedPreview};
 use crate::tui::components::{
-    format_scroll_indicator, set_prefixed_input_cursor_position, truncate_to_width, HelpOverlay,
-    Preview,
+    format_scroll_indicator, prefix_within_width, set_prefixed_input_cursor_position,
+    truncate_to_width, HelpOverlay, Preview,
 };
 use crate::tui::responsive;
 use crate::tui::styles::{has_min_contrast, Theme};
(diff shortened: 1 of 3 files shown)


jerome-benoit (changes requested):
Requesting changes at `aa68e9dc67a3e22710128a758a80abfda355ebc0`.

- **[P2] Keep profile tags within four terminal cells.** The existing [CodeRabbit thread](https://github.com/agent-of-empires/agent-of-empires/pull/3680#discussion_r3907082736) is valid, but lowercasing before `take(4)` is not sufficient. Profile names created through the CLI are not ASCII-only, and `profile_short_code("界界界-main")` now returns `界界界m`: four Unicode scalars but seven terminal cells. `RowTag::rendered()` only applies a minimum formatting width; it does not truncate the content, so this breaks the fixed-width tag invariant and can shift or suppress the activity column. Enforce the cap using display width (and make padding display-width-aware), or consistently restrict the accepted profile-name grammar. Please cover both case expansion such as `İ` and wide glyphs such as `界`.
- **[P3] Update the function documentation.** The comment above `profile_short_code` still says hyphen/underscore-delimited names collapse to segment initials, but this PR intentionally changes short leading segments to remain whole (`wma-work` is now `wmaw`, not `ww`). Please document the new rule where the function's contract is defined.

The intended ASCII cases and the long-lead fallback are otherwise consistent. `git diff --check` and `cargo fmt --all -- --check` pass. The PR currently has no full Rust CI result at this SHA; I did not treat the author's targeted-test statement as independently reproduced.

src/tui/home/render.rs
```diff
@@ -701,21 +698,66 @@ fn branch_tag_content(branch: &str, max_width: usize) -> Option<String> {
...
+    truncate_to_display_width(&code.to_lowercase(), MAX_CELLS)
+}
+
+/// The longest prefix of `s` that fits in `max_cells` terminal cells
+/// (`unicode-width`), so a tag cap holds on screen and not just in scalars.
+fn truncate_to_display_width(s: &str, max_cells: usize) -> String {
+    let mut out = String::new();
+    let mut used = 0;
+    for c in s.chars() {
+        let w = unicode_width::UnicodeWidthChar::width(c).unwrap_or(0);
```
jerome-benoit: **[P2] Measure contextual Unicode width across grapheme-aligned prefixes**

`UnicodeWidthChar` is not additive for emoji sequences, while `RowTag::rendered()` measures the resulting string with `UnicodeWidthStr`.

With the current implementation, the valid profile `♥️界-a` produces `♥️界a`: the loop counts four cells, but `UnicodeWidthStr` reports five, so the rendered tag still exceeds its fixed-width contract. Conversely, `🤝🏽-main` loses the `m` even though `🤝🏽m` fits in three cells.

Reproduced against locked `unicode-width 0.2.2`:

```
input="♥\u{fe0f}界a" output="♥\u{fe0f}界a" scalar_sum=4 string_width=5
input="🤝🏽m" output="🤝🏽" scalar_sum=4 string_width=2
```

Iterate over grapheme-aligned prefixes and measure each accumulated candidate with `UnicodeWidthStr`, following the existing approach in `src/tui/components/text.rs`. Add both sequences as regressions.

jerome-benoit (changes requested):
Rereviewing `cba4484bed6c3c2542abfb85de2beb810478c329`.

The previous findings are fixed for the covered cases: lowercasing precedes truncation, CJK and case-expansion inputs are capped, padding uses display width, and the function documentation now describes the implemented rule.

One contextual Unicode-width case still prevents approval; see the inline finding.

CodeRabbit's current branch-tag finding is valid but predates this PR: branch tags already used scalar-count truncation and could exceed their display budget at the merge base. It should be handled separately rather than block this focused fix.

Non-blocking: the new test comments at lines 4862 and 5091 use em dashes, contrary to `AGENTS.md`.

`git diff --check` passes. Only title/template CI checks exist at this SHA; no full Rust CI result is available.

jerome-benoit (approved):
Re-reviewed at `691945123bf1f8df68bf49ceed25e6f47b388dee`, including your point-by-point response. The contextual Unicode-width blocker is resolved: grapheme-aligned prefixes and rendering now use the same string-width measurement. The earlier padding, case-expansion and documentation findings remain fixed.

Two independent reviews found no new blocker. An isolated harness using the exact changed helpers and locked Unicode dependencies passed 8 tests, including both reported sequences; `git diff --check` also passed. This was not a full project test run.

The branch-tag overflow remains preexisting and nonblocking for this PR. I approved the pending CI workflows; full CI is still running, so this approval does not claim it is green.

jerome-benoit (changes requested):
Re-reviewed at `f957020b7dc064db8f928f3deaf2df316d872077`, including the previous replies. The rebase preserves the approved production changes, but cross-validation found one profile-tag width case that my previous approval missed. This is not caused by the rebase.

**[P2] Match the renderer's cell-width metric.** For profile `ｶﾞｻﾞﾊﾟ`, `profile_short_code` retains three grapheme clusters and `RowTag::rendered()` adds one space because `UnicodeWidthStr` measures the content as three cells. Locked Ratatui 0.1.2 renders those clusters as six cells, making the bracketed tag nine cells instead of six and breaking row alignment. The base's scalar-limited code did not overflow for this input.

Use the renderer's per-grapheme `CellWidth` metric consistently for the cap and padding, preserving grapheme boundaries, and assert the result through a Ratatui buffer rather than only `UnicodeWidthStr`. Evidence: `src/tui/home/render.rs:599-601,727-738`; `src/tui/components/text.rs:64-71,131-139`; locked `ratatui-core-0.1.2/src/buffer/cell_width.rs` explicitly adds cells for U+FF9E/U+FF9F. These were inspected with `git show` and the local locked dependency source.

Both independent reviewers agree after arbitration. This is static verification, not an executed reproduction. The earlier Unicode fixes remain valid; the preexisting branch-tag issue remains outside this PR. `git diff --check` passes. CI workflow approvals are cleared and the current-head runs are queued; no local tests were run.

jerome-benoit (changes requested):
Re-reviewed `ffb96c4a66456b3291fa844f8c0412b2551be322`, including the latest thread-disposition comment. The existing **P2 remains blocking**: the cap and padding still use `UnicodeWidthStr`, not the renderer's per-grapheme `CellWidth` metric.

For `ｶﾞｻﾞﾊﾟ`, the code retains three clusters and adds one padding space. Locked `ratatui-core 0.1.2` adds a cell for each U+FF9E/U+FF9F, so the bracketed tag occupies nine cells rather than six. Please use the renderer's metric consistently and assert the result through a Ratatui buffer, as requested previously.

The latest comment does not match this remote head: `RowTag::rendered()` still does not cap its content, and neither named `row_tag_rendered_caps_wide_content_to_max_width` nor `row_tag_rendered_keeps_grapheme_sequences_whole` test exists here. The relevant source files are unchanged from reviewed `f957020b`. The separate branch-tag overflow remains pre-existing and nonblocking; fixing that alone would not resolve this profile-tag P2.

Evidence: exact-SHA `git show`, `git grep` and restricted `git diff f957020b ffb96c4a`; `render.rs:599-601,727-738`, `components/text.rs:131-139`, and the locked dependency's `buffer/cell_width.rs:34-43`. Two independent sequential analyses and final cross-validation; no local runtime reproduction. `git diff --check` is clean and exact-head CI is green.

PR #3682, opened 2026-09-01
feat(tui): lock the restart dialog's engine cycler to the profile's pinned tool

BTForIT:
## Problem

The restart dialog's Tool cycler always offers every detected engine. A profile pinned to one engine via `[session] default_tool` (e.g. a `codex` profile) can still be cycled onto a different engine at restart and relaunch the wrong AI — and a one-engine profile is rendered as if it had several. The profile→engine association that exists at create time is lost at restart time.

## Change

Lock the cycler to the profile's pinned engine when that engine is available:

- `cyclable_tools()` returns just the pin; `cycle_tool()` steps only within the allowed set (a no-op at one engine).
- `render_tool_selector()` drives its count off `cyclable_indices()`, so `tool_cycler_spans` drops the arrows/badge and the row shows the single engine.
- No pin, or a pin naming an engine that is not installed, keeps the full upstream behaviour unchanged.
- The per-profile selection is cached (`cyclable_cache`) so the disk-backed config lookup runs at most once per selection, cleared on profile change.

One file: `src/tui/dialogs/restart.rs`.

## Testing

6 new tests (lock-on-cycle, lock-on-click, unconstrained-still-cycles, + 3 pure `cyclable_tools` cases); the restart-dialog suite is 36/36 green. Rebased onto current `main`; `cargo check --features serve` clean.

https://claude.ai/code/session_01Sz4RseJVCpKgfG3oKjNe4t

## PR Type

- [x] New Feature

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the lock is pure dialog-state logic covered by 6 new unit tests in the restart-dialog suite (36/36 green); unpinned profiles keep upstream behaviour byte-for-byte

## AI Usage

- [ ] No AI was used
- [x] AI was used

### DIFF
--- a/src/tui/dialogs/restart.rs
+++ b/src/tui/dialogs/restart.rs
@@ -64,6 +64,24 @@ pub struct RestartDialog {
     /// Which selector row the mouse is over, for the hover highlight.
     /// Visual only; never moves keyboard `focused_field`.
     hover: HoverState,
+    /// The engines the Tool cycler may select for the currently selected
+    /// profile, cached so the config lookup (disk-backed) runs at most once per
+    /// profile selection rather than every keystroke/frame. `None` until first
+    /// primed; cleared whenever the profile changes. See `cyclable_tools`.
+    cyclable_cache: Option<Vec<String>>,
+}
+
+/// The engines the Tool cycler may land on for a profile. When the profile
+/// pins `session.default_tool` to an engine that is actually available, the
+/// cycler is locked to exactly that engine (one-engine-per-profile), so a
+/// restart cannot launch a different model than the profile is configured for.
+/// No pin, or a pin naming an unavailable engine, leaves every detected engine
+/// selectable (unchanged upstream behavior).
+fn cyclable_tools(available: &[String], pinned: Option<&str>) -> Vec<String> {
+    match pinned {
+        Some(p) if available.iter().any(|t| t == p) => vec![p.to_string()],
+        _ => available.to_vec(),
+    }
 }
 
 impl RestartDialog {
@@ -104,6 +122,7 @@ impl RestartDialog {
             profile_selector_area: Rect::default(),
             tool_selector_area: Rect::default(),
             hover: HoverState::default(),
+            cyclable_cache: None,
         }
     }
 
@@ -135,10 +154,8 @@ impl RestartDialog {
         }
         if self.tool_selector_area.contains(pos) {
             self.focused_field = 1;
-            if !self.available_tools.is_empty() {
-                self.tool_index = (self.tool_index + 1) % self.available_tools.len();
-                self.reload_tool_config();
-            }
+            // Cycle within the profile's locked engine set (no-op when pinned).
+            self.cycle_tool(true);
             return Some(DialogResult::Continue);
         }
         None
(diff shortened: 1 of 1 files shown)


src/tui/dialogs/restart.rs
```diff
@@ -64,6 +64,24 @@ pub struct RestartDialog {
...
+    cyclable_cache: Option<Vec<String>>,
+}
+
+/// The engines the Tool cycler may land on for a profile. When the profile
+/// pins `session.default_tool` to an engine that is actually available, the
+/// cycler is locked to exactly that engine (one-engine-per-profile), so a
+/// restart cannot launch a different model than the profile is configured for.
+/// No pin, or a pin naming an unavailable engine, leaves every detected engine
+/// selectable (unchanged upstream behavior).
+fn cyclable_tools(available: &[String], pinned: Option<&str>) -> Vec<String> {
```
jerome-benoit, line 80: [P1] `default_tool` is an initial selection, not an allowed-tool policy

This turns every available `session.default_tool` into an exclusive one-item allowlist. That contradicts the setting's documented "Default agent for new sessions" contract and this dialog's existing promise that users can keep the profile while swapping only the engine. Please retain the normal cycler, or introduce a separate, documented restriction setting and enforce it consistently across New Session and Restart.

src/tui/dialogs/restart.rs
```diff
@@ -249,6 +270,62 @@ impl RestartDialog {
         self.command_override = Input::new(config.session.resolve_tool_command(&tool));
     }
 
+    /// Prime `cyclable_cache` for the selected profile if it isn't already.
+    /// Reads `session.default_tool` once (the config lookup is disk-backed);
+    /// the cache is invalidated whenever the profile changes, so the read runs
+    /// at most once per profile selection rather than every keystroke/frame.
+    fn ensure_cyclable(&mut self) {
```
jerome-benoit, line 277: [P1] Normalize the initial tool before claiming this lock

The dialog initializes `tool_index` from `current_tool`, while `ensure_cyclable()` only caches the configured default. If the session currently uses `codex` and the profile now defaults to `claude`, the row displays `codex` as the sole non-cyclable choice and Enter submits no tool change, so the restart still launches `codex`. The tests preload a matching cache and do not cover this case.

src/tui/dialogs/restart.rs
```diff
@@ -249,6 +270,62 @@ impl RestartDialog {
...
+        if self.cyclable_cache.is_some() {
+            return;
+        }
+        let pinned = self
+            .selected_profile()
+            .and_then(|p| resolve_config_or_warn(p).session.default_tool.clone());
```
jerome-benoit, line 283: [P2] Resolve the repository-level default too

Repository configuration explicitly supports overriding `session.default_tool`, and New Session uses `resolve_config_with_repo_or_warn(profile, path)`. Restart receives neither the instance project path nor repository-aware configuration, so it may lock to the profile default or remain unlocked instead of honoring the repository default.

jerome-benoit (changes requested):
The current approach changes the established meaning of `session.default_tool`: it is documented as the initial tool for new sessions, not as an exclusive per-profile restriction. The New Session dialog also continues to allow explicit tool selection after applying that default.

Additionally, the proposed lock is incomplete: an existing session whose current tool differs from the configured default still renders and submits that current tool unchanged. Repository-level `default_tool` overrides are also ignored because the restart dialog resolves profile configuration without the session's project path.

Please preserve the default-only contract, or introduce an explicit restriction setting and apply it consistently across creation and restart flows. If a hard lock remains intended, the initial selection, command/argument state, repository-aware resolution, documentation, and regression tests must all be updated.

PR #3683, opened 2026-09-01
feat(acp): pin a structured-view model per agent and enforce it at session creation

BTForIT:
## Problem

The structured-view (ACP) model picker offered every catalog model even for a profile whose operator wants one model only, and nothing at session creation enforced that intent. Review (#3683) showed the original change was only presentation, and that its `agent_extra_args` leg was wrong at the root: those args never reach an ACP spawn, so a `--model` there never pinned anything.

## Change

An explicit, opt-in pin, enforced where sessions are created:

- New config field `acp.acp_defaults.<agent>.pin_model = true` (serde default `false`, omitted when false; existing configs unchanged). `acp_defaults.<agent>.model` keeps its documented meaning as a default.
- `resolve_spawn_model_effort` (`src/session/config/mod.rs`), the single resolver behind CLI `add`, the TUI wizard, web create and the ACP supervisor's process launch, lets the pin win over an explicit request and over the default. The supervisor re-resolves on every spawn, so a stored pre-pin `agent_model` cannot escape on respawn.
- The worker's own respawn (`Supervisor` drain task) re-runs that resolution on its cached `SpawnConfig` (`refresh_spawn_model_effort`), so a pin changed between launches applies to the next automatic restart. Every creation surface resolves the pin through the agent the session spawns as (`crate::acp::pick_acp_agent_name` / `pinned_model_for_tool`), so a wrapper mapped through `agent_detect_as` reads its base agent's pin.
- Plugin `sessions.create` (`src/plugin/session_api.rs::admit_and_create`) refuses a `model_id` off the pin with `INVALID_PARAMS`, `data.kind = "model_pinned"` plus `agent_id` / `model_id` / `pinned_model`.
- The picker (`src/server/api/plugin_settings.rs`) collapses only on an explicit pin, using the catalog label when known. `parse_model_flag` and the `agent_extra_args` leg are deleted.
- `docs/guides/configuration.md` documents `pin_model`.

Out of scope: a running session's live model switch (`aoe session set-model`, the composer) is not gated here.

## Testing

- Review round 1 (eee8c489): RED `test result: FAILED. 183 passed; 3 failed` (the three enforcement tests), GREEN `186 passed; 0 failed` on the `pin` filter.
- Review round 2 (a248caaa): RED `0 passed; 2 failed` on the two new behaviour tests (wrapper creation returned `unknown_model` instead of `model_pinned`; picker returned `None`), then `E0425` on the three helper tests added before the helpers; GREEN `6 passed; 0 failed` on the pin filter (four new tests plus the two existing pin tests).
- `cargo test --features web,e2e-tests --lib`: 6668 passed, 0 failed, 2 ignored.
- `cargo fmt --all -- --check` and `cargo clippy --all-features -- -D warnings`: clean.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: server-side collapse of the AcpModels list, covered by unit tests for the pin lookup (`pinned_model_for_tool`, including an `agent_detect_as` wrapper) and the collapse behaviour; the picker's flow/markup is unchanged

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

This PR adds optional ACP model pinning per agent through `acp.acp_defaults.<agent>.pin_model`.

Pinned models apply during session creation and process respawn, including agents mapped through `agent_detect_as`. Conflicting model requests return a typed `model_pinned` error. The model picker shows only the pinned model, while unpinned agents keep existing behavior.

The change centralizes agent and pin resolution, removes unused `agent_extra_args` parsing, adds documentation and tests, and prevents unknown profiles from causing filesystem changes.

Benefit: ACP sessions stay aligned with configured models without changing unpinned behavior.


jerome-benoit (changes requested):
Requesting changes because the picker restriction is not aligned with the authoritative creation contract.

- **[P1] Enforce the pin at the authoritative session-creation boundary.** The resolver only removes catalog options. The generic widget deliberately preserves a previously stored off-pin value as `(unavailable)` (`web/src/components/settings/StructuredWidgets.tsx:26-29,62-69`), while plugin `sessions.create` still accepts any advertised `model_id` and forwards it as the explicit session model (`src/plugin/session_api.rs:344-357,426-469`). Explicit requests then win over defaults (`src/session/config.rs:1545-1565`). Consequently an existing off-pin plugin configuration can continue spawning on that model. Use a single authoritative pin resolver in `sessions.create`, rejecting or replacing conflicting values, and cover the stored-value case.

- **[P2] Do not reinterpret `acp_defaults` as an immutable pin.** `pinned_model_for_agent` collapses the picker whenever `acp_defaults.<agent>.model` is set, but the repository explicitly defines this field as a default: an explicit request wins, and the default only fills an absent request (`src/session/config.rs:1545-1565`; `docs/guides/configuration.md:145`). Preserve the full catalog and preselect the default, or introduce and enforce an explicit pin contract across all creation surfaces.

- **[P2] Parse the launched argv semantics rather than whitespace tokens.** `parse_model_flag` retains shell quoting as part of the value. For example, `--model 'claude-x[1m]'` produces the fabricated model id `'claude-x[1m]'`, while the terminal launch path intentionally preserves those quotes so the shell removes them (`src/session/config.rs:3307-3315,3421-3424`). Use shell-aware/shared parsing and test quoted spaced and joined values.

The previous CodeRabbit finding is resolved and outdated. `git diff --check` is clean. GitHub currently reports only the title/template checks; no Rust build, test, or Clippy check ran for this head.

src/acp/supervisor.rs
```diff
@@ -1751,9 +1751,11 @@ impl<S: BroadcastSink> Supervisor<S> {
 
         // Resolve the per-agent structured-view defaults once, at this single
         // spawn choke point, so CLI create, reconciler respawn, and web create
-        // all honor the same model/effort/mode defaults. An explicit
-        // per-request model or effort wins; otherwise the configured default
-        // fills in. Mode has no per-request override today.
+        // all honor the same model/effort/mode defaults and the same pin. A
```
jerome-benoit: **[P2] Refresh the pin on automatic watchdog respawns**

This resolver is reached by `spawn_inner`, but not every respawn. `restart_decision` clones the saved `SpawnConfig` (`supervisor.rs:3642`), and the retry path calls `AcpClient::spawn` directly (`:2486`). It refreshes MCP and hooks, but retains the original `provider_env` and its `AOE_AGENT_MODEL` (`:1843-1845,1884-1892`).

If a worker launched on A and the profile pin is changed to B, its next automatic crash/watchdog restart still receives A. That contradicts the stated creation-and-respawn contract, independently of live model switching.

Apply the same current model/effort resolution before retrying, replacing the cached model environment, and add a regression covering a changed pin between launches. Alternatively, explicitly narrow the supported contract instead of promising every respawn. Verified by tracing these paths at this SHA; no runtime reproduction claimed.

jerome-benoit (changes requested):
Re-reviewed at `eee8c48945ec20ca6eeb92237e4abbdc70cc6ea4`, including your response and the explicit creation/respawn scope. Plain defaults now retain their semantics, direct-agent creation enforces the opt-in pin, and deleting the irrelevant shell parser resolves that finding. I am not requesting a live `set-model`/composer restriction.

Changes are still requested for two P2s:

- The automatic watchdog-respawn path does not refresh the pin; see the inline comment. Both independent reviews found this, and source-level arbitration retains P2 rather than P1.
- The [existing agent-identity finding](https://github.com/agent-of-empires/agent-of-empires/pull/3683#discussion_r3943153804) also needs resolution. In web creation, model selection/persistence uses the logical tool before `pick_agent_for_tool` can select its backend; the supervisor then reads the backend pin. Use one chosen identity throughout. The plugin example is conditional on registry/catalog admission, not a claim that every arbitrary wrapper is accepted. No duplicate inline added.

Nonblocking: update the stale PR coverage text referring to parser tests, and shorten the new narrative comments.

`git diff --check` passes. CI was approved and is running. These findings are based on traced code paths, not a claimed end-to-end reproduction.

src/acp/supervisor.rs
BTForIT: Fixed. The worker's own respawn now re-runs the spawn resolution on its cached `SpawnConfig` before `AcpClient::spawn`: `refresh_spawn_model_effort` (src/acp/supervisor.rs) takes the cached `AOE_AGENT_MODEL` and `default_effort` as the request, runs `resolve_spawn_model_effort` against the agent's current `acp_defaults` entry (re-read through `resolve_config_with_repo_or_warn` on a blocking task, right after the MCP re-resolution), and replaces the model env entry and the effort. A pin changed from A to B between launches lands on the respawn; without a pin the cached values stand. One nuance, stated in the doc comment: the cached effort is a resolved value, so when the pin moves the model, `effort_by_model` for the new model wins over it.

Regression: `respawn_refreshes_a_changed_pin_on_the_cached_config` covers a moved pin, an unchanged pin, no entry, and a plain default, at the helper level. The existing respawn e2e and integration tests drive `AcpClient::spawn` directly and no shim records the model env, so a watchdog-driven end-to-end assertion needs a new harness; I can add one if you want it in this PR.

Reply written by an AI agent operating the BTForIT account on behalf of its owner; the change was tested as described above.

src/plugin/session_api.rs
```diff
@@ -341,6 +341,45 @@ async fn admit_and_create(
...
+        .as_deref()
+        .map(str::trim)
+        .filter(|value| !value.is_empty());
+    if let Some(model) = requested_model {
+        let profile = deps.profile.clone();
+        let agent_id = req.agent_id.clone();
+        let pinned = tokio::task::spawn_blocking(move || {
+            crate::session::config::profile_config::resolve_config_or_warn(&profile)
+                .acp
+                .pinned_model_for(&agent_id)
```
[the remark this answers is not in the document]
BTForIT: Fixed. There is now one identity: the agent `pick_agent_for_tool` resolves. `pick_acp_agent_name`, the sync mirror `aoe add` already used, moved from src/cli/add.rs to src/acp/agent_registry.rs (re-exported from `crate::acp`), and `pinned_model_for_tool(config, tool, override)` reads the pin keyed by it. Web create (src/server/session_spawn.rs) resolves `agent_key` through the mirror before reading the defaults, so the model it persists and the pin it applies are the backend's; the plugin `sessions.create` gate and the settings picker read the pin through `pinned_model_for_tool`. A wrapper mapped through `agent_detect_as` reads its base agent's pin on every surface; a pin on the wrapper id itself is never consulted, because the spawn never reads it.

Regressions: `create_reads_the_pin_of_the_agent_a_wrapper_spawns` (plugin create with a catalog-admitted wrapper refuses a model off the base pin with `model_pinned`), the wrapper case in `picker_collapses_on_a_pin_but_not_on_a_default`, and unit tests on `pick_acp_agent_name` and `pinned_model_for_tool`. Respawn keys the refresh on the `SpawnConfig`'s `agent_key`, which is already the resolved backend; covered by the refresh test in the other thread.

Reply written by an AI agent operating the BTForIT account on behalf of its owner; the change was tested as described above.

src/server/api/plugin_settings.rs
```diff
@@ -239,6 +250,54 @@ async fn group_options(state: &Arc<AppState>) -> Vec<SelectOption> {
...
+        if let Some(v) = tok
+            .strip_prefix("--model=")
+            .or_else(|| tok.strip_prefix("-m="))
+        {
+            if !v.is_empty() {
+                return Some(v.to_string());
+            }
+        } else if *tok == "--model" || *tok == "-m" {
+            if let Some(v) = toks.get(i + 1).filter(|v| !v.is_empty()) {
+                return Some(v.to_string());
```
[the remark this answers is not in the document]
BTForIT: Resolved by deletion at eee8c489: `parse_model_flag` and the `agent_extra_args` leg are gone, the picker reads the pin from `acp_defaults` only, so there is no flag token to reject.

Reply written by an AI agent operating the BTForIT account on behalf of its owner; the change was tested as described above.

src/acp/supervisor.rs
```diff
@@ -754,6 +754,43 @@ fn log_wrapper_substitution(session_id: &str, tool: &str, wrapper: &str, base: &
...
+        cached_model.clone(),
+        config.default_effort.take(),
+    );
+    if model != cached_model {
+        if let Some(keyed) = defaults
+            .zip(model.as_deref())
+            .and_then(|(defaults, model)| defaults.effort_by_model.get(model))
+            .filter(|value| !value.is_empty())
+        {
+            effort = Some(keyed.clone());
```
jerome-benoit: **[P2] Preserve explicit effort when refreshing a changed model pin**

A session created with explicit effort `low` stores it in `Instance.acp_effort` (`server/session_spawn.rs:290-308`); the initial resolver preserves it (`session/config/mod.rs:1616-1619`) and caches it in `SpawnConfig.default_effort` (`supervisor.rs:1928`). If the pin later changes from A to B with `effort_by_model[B] = "high"`, this assignment overwrites that explicit `low`. The respawn passes the result to the adapter (`src/acp/acp_client/connection.rs:1399-1414`), although the contract says the pin affects only the model.

The cached value can be explicit or inherited; preserve that provenance (or reload the persisted explicit effort) so only inherited effort is re-resolved for B. Add separate explicit-effort and inherited-effort cases. The current helper test treats every cached `low` as inherited and therefore misses this distinction. Verified by exact-head `git show`/data-flow inspection, not a claimed runtime reproduction.

jerome-benoit (changes requested):
Re-reviewed `a248caaa08929175d88e3f9dcf86301d28234212`, including your replies. The watchdog now refreshes the model pin, and valid wrappers consistently resolve the backend identity. Those previous P2s are addressed.

**One new P2 still requires correction:** the refresh overwrites explicitly requested effort when the model pin moves. Both independent sequential reviews found it; the inline explains the request/default distinction needed. I considered your cached-effort explanation, but it does not preserve the documented model-only pin contract.

**Existing nonblocking P3:** [malformed `agent_acp_cmd` identity mismatch](https://github.com/agent-of-empires/agent-of-empires/pull/3683#discussion_r3951745365). The mirror checks key presence while the supervisor parses the command (`agent_registry.rs:279`; `supervisor.rs:1128`). The analyses differed on severity; I retain P3 because this requires invalid configuration and the authoritative spawn still applies the actual backend pin. No duplicate inline.

Live model switching remains out of scope. `git diff --check eee8c489 a248caaa` is clean; CI has no failed check, with `core-rest` still running at the latest read. No local build/test or runtime reproduction was performed.

src/acp/supervisor.rs
```diff
@@ -1760,14 +1797,21 @@ impl<S: BroadcastSink> Supervisor<S> {
...
+        // on its cached config, see `refresh_spawn_model_effort`.
         // ponytail: resolve here instead of threading model/effort/mode through
         // every SpawnRequest site; revisit if explicit per-request values land.
         let acp_defaults = resolved_cfg.acp.acp_defaults_for(&agent);
+        // Provenance of the effort now going into the SpawnConfig: an
+        // explicit request effort is a session pin and must survive a later
+        // model-pin change on respawn; only inherited effort re-resolves.
+        let effort_explicit = effort
+            .as_deref()
+            .is_some_and(|value| !value.trim().is_empty());
```
jerome-benoit: [P2] Preserve request provenance before resolving inherited effort

A nonempty `effort` here is not necessarily user-selected. `server/session_spawn.rs:301-308` resolves the default while leaving `Instance.acp_effort = None`, then forwards that resolved effort through lines 435, 449 and 556. Creating without an effort override on pinned A (`low`), then changing the pin to B (`effort_by_model[B] = "high"`), sets this flag to true and prevents the watchdog from selecting B's inherited effort. Pass the actual explicit-effort value/provenance across the boundary and resolve inherited effort for the new model, including the ordinary fallback when no keyed entry exists. Add a regression through the creation handoff; manually setting this boolean in the helper test misses the bug.

jerome-benoit (changes requested):
Re-reviewed `9b2bd8480c884cc22466999ae6b127e2eb232a6a`, including the latest response and previous scope decisions.

The explicit-effort case in the helper is fixed, but **one P2 remains in the production handoff**: `SpawnRequest.effort` is not always an explicit request. Web creation passes the already resolved default (`server/session_spawn.rs:301-308,435,449,556`), so the new nonempty check in `supervisor.rs:1812-1814` marks inherited effort as explicit. Creating without an effort override on model A with inherited `low`, then changing the pin to B with `effort_by_model[B] = "high"`, leaves the watchdog respawn on B with `low`. Preserve the original provenance through this boundary, or pass only the persisted explicit effort and resolve defaults in the supervisor. When the model changes, inherited effort should use the complete `effort_for_model` resolution, including its fallback when B has no keyed entry, rather than treating the old resolved value as a request. The helper tests set the boolean directly and miss this caller mismatch; add a creation-to-respawn regression.

The watchdog model refresh, valid-wrapper identity and malformed-command fallback are fixed. Plain defaults remain defaults; the obsolete argv parser is gone. Live model switching remains outside the agreed scope. No duplicate of the old explicit-effort inline is added; the new inline identifies the faulty provenance check.

Two independent sequential analyses and final code-level cross-validation. Evidence: exact-SHA `git show`/base-head `git diff`; `git diff --check` is clean. No local tests run. I approved the waiting exact-head workflows; CI is still running.

src/acp/supervisor.rs
jerome-benoit: The production handoff was corrected at a691a5ed: SpawnRequest carries explicit effort provenance, producers derive it from the persisted user effort, and refresh_spawn_model_effort preserves explicit effort while re-resolving inherited effort. Correction to my earlier test claim: creation_handoff_keeps_resolved_default_effort_inherited manually assigned req.effort_explicit=false and inspected the copied internal flag. It was a supervisor-boundary check, not an end-to-end creation handoff and not “no hand-set flags”. Follow-up #3903 removes that misleading implementation-only test and shortens the accompanying comments; the explicit/inherited respawn-resolution regressions remain. #3903 also fixes the remaining whitespace-only pin finding with a failing-before/passing-after resolver regression.

jerome-benoit (approved):
Re-reviewed `a691a5edeb0dcfca373595077bec39d9b0da41d3`, including the latest response and previous scope decisions. The remaining P2 is fixed: creation forwards persisted effort provenance (`src/server/session_spawn.rs:450,559`), the supervisor retains it (`src/acp/supervisor.rs:1961`), and automatic respawn preserves explicit effort while fully re-resolving inherited effort, including fallback/removal (`:752-761,2510-2525`). No remaining blocking finding in the agreed creation/respawn scope; live model switching stays out of scope.

Nonblocking qualifications: the new test sets `req.effort_explicit = false` (`supervisor.rs:4429`), so it covers the supervisor boundary, not the complete creation handoff. Please describe that accurately. Shorten the new narrative comments and remove their em dashes/review-history reference per AGENTS.md. The previously reported whitespace-only configured pin remains P3 configuration hardening, not a duplicate blocker.

Two independent sequential analyses, followed by source-level cross-validation, agree on the verdict. Evidence: exact-SHA `git show` and `git diff 9b2bd848 a691a5ed`; `git diff --check` exits 0 with no output. No local tests run. CI workflows were authorized; at the latest read, bare-core and Rust CodeQL were still running, with no failed check. Approval concerns the code, not an assertion that pending CI passed.

PR #3692, opened 2026-09-02
fix(web): stop leaking intermediate Korean IME syllables to the PTY

orientpine:
## Description

Typing Korean in the web terminal from iOS leaks every intermediate syllable to the PTY: typing `문제` produces `ㅁ무문묹문제`.

iOS WebKit fires no composition events for the Korean keyboard ([WebKit bug 274700](https://bugs.webkit.org/show_bug.cgi?id=274700), same behavior reported in [xterm.js#6084](https://github.com/xtermjs/xterm.js/issues/6084)). Each keystroke instead rewrites the trailing syllable through the plain editing path: `deleteContentBackward` for the previous state, then `insertText` with the new one ("ㅎ" -> "하" -> "한"). The live terminal's hidden textarea was kept permanently empty (`preventDefault` on `insertText`), and WebKit dispatches no `beforeinput` at all for a delete with nothing before the caret, so the deletes vanished and only the inserts reached the PTY.

The fix keeps forwarding `insertText` / `deleteContentBackward` to the terminal but no longer prevents their default action: typed text stays in the hidden textarea as IME context, so each rewrite's delete becomes observable and maps to one DEL (the same path soft-keyboard Backspace autorepeat uses, #1450 / #1615). Enter is the safe point to drop the accumulated text; no IME re-edits a syllable across a line break. Specifics:

- New `forwardTerminalBeforeInput` in `web/src/lib/mobileKeyboardProxy.ts`; the App-level keyboard proxy and `MobileLiveTerminal`'s own input now share it instead of duplicating the switch.
- `compositionend` no longer clears the textarea: IMEs that re-edit a committed syllable (delete + reinsert) need the text present for the delete to surface.
- The persistent keyboard proxy textarea disables autocorrect/autocapitalize/autocomplete/spellcheck, matching the live terminal's own hidden input, since it now retains text.

Verified on a real iPhone (iOS Chrome, WebKit) with the Korean keyboard: previously garbled input now arrives fully composed, and PTY bytes for the ㅎ/ㅏ/ㄴ keystrokes are exactly `ㅎ DEL 하 DEL 한`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (input handling only, no visual change)

## Test Coverage Analysis

New `web/tests/live-terminal-ime-rewrite.spec.ts` covers the delete + reinsert sequence reaching the PTY as DEL + text and Enter dropping the retained IME context. New Vitest cases cover `forwardTerminalBeforeInput`; the `mobile-keyboard.spec.ts` proxy assertion now expects `insertText` to land in the textarea. `npm run format:check`, `npm run lint`, `npx tsc -b`, `npx vitest run`, and the touched mocked Playwright specs (`mobile-keyboard`, `backspace-autorepeat`, `live-terminal-ime-rewrite`) all pass.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude (Anthropic), running as the omo coding agent.

Diagnosed from a live reproduction (iPhone iOS Chrome against `aoe serve` v1.15.1), implemented, and verified end to end: unit + mocked Playwright tests plus manual Korean typing on the affected device against a patched build. The account owner reviewed the change and requested this PR.

## Summary

- Fixed Korean IME input on iOS web terminals.
- Preserved hidden textarea content during IME rewrites.
- Shared input handling between the keyboard proxy and live terminal.
- Cleared retained IME input after Enter, session changes, control input, and toolbar or paste actions.
- Disabled browser corrections on the keyboard proxy textarea.
- Added unit and browser tests for input forwarding, IME rewrites, session switching, Enter handling, and out-of-band input.

## Benefits

- Prevents intermediate Korean syllables from reaching the PTY.
- Preserves the final composed text.
- Prevents stale IME input from leaking between sessions or input paths.
- Keeps keyboard-proxy and live-terminal behavior consistent.


web/src/lib/mobileKeyboardProxy.ts
```diff
@@ -36,3 +36,40 @@ export function clearMobileKeyboardProxyInput() {
...
+ * the soft-keyboard Backspace path this shares.
+ *
+ * Line breaks and pastes never reach the textarea. A line break is also the
+ * safe point to drop the accumulated text: no IME re-edits a syllable across
+ * Enter. */
+export function forwardTerminalBeforeInput(ev: InputEvent, deliver: (input: MobileKeyboardProxyInput) => void) {
+  switch (ev.inputType) {
+    case "insertText":
+    case "deleteContentBackward":
+      deliver({ inputType: ev.inputType, data: ev.data, isComposing: ev.isComposing });
```
jerome-benoit: [P1] Invalidate retained context after out-of-band terminal input

This textarea is now a shadow of the current PTY edit, but navigation, Ctrl+C, Tab, Escape, and toolbar paste are sent directly to the PTY without updating or clearing that shadow. For example, after iOS inserts `ㅎ`, toolbar Ctrl+C resets the PTY line while the textarea still contains `ㅎ`; the next Korean keystroke rewrites that stale value as `DEL + 하` into the new prompt. Please reset or finalize the retained IME context whenever terminal input bypasses native textarea editing, and add a regression covering retained text, Ctrl+C or navigation, then another rewrite.

jerome-benoit (changes requested):
Requesting changes for two retained-IME-context lifetime gaps.

The existing session-boundary thread is valid: changing sessions clears the receiver and queue, but not the persistent textarea value. I am not duplicating that finding inline.

The retained context can also become stale within one session whenever an action bypasses native textarea editing. This needs a defined invalidation boundary and regression coverage before the shadow textarea can safely drive PTY rewrites.

The separate suggestion to run the mocked spec under WebKit is not required as written. The test deliberately synthesizes the relevant event sequence under the repository's Chromium-only mocked/coverage setup; it is handler coverage, not an automated reproduction of the iOS keyboard.

jerome-benoit (changes requested):
Re-reviewed `5dccc789` against `8f0cece8` and the previous head, with two independent sequential analyses and cross-validation. The merge preserves main's Android retroactive-composition handling; no additional integration defect was established.

The existing blockers still require correction, without duplicate inline comments:
- **P1:** [invalidate retained IME context after out-of-band terminal input](https://github.com/agent-of-empires/agent-of-empires/pull/3692#discussion_r3934865594). Ctrl/navigation/paste paths still bypass the retained textarea context.
- **P2:** [clear the persistent proxy textarea across session changes](https://github.com/agent-of-empires/agent-of-empires/pull/3692#discussion_r3910368801). Clearing the receiver and queue alone does not clear its value.

The mocked Chromium event-sequence test remains useful handler coverage, not native WebKit reproduction; changing that test's browser is not an additional requirement.

Evidence: exact-head source/diff inspection; `git diff --check 8f0cece8 5dccc789` clean. No local build/test or device reproduction. CI/Tests/CodeQL required approval despite green metadata checks; I approved those workflows.

web/tests/live-terminal-ime-rewrite.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      else ta.setRangeText("", Math.max(0, end - 1), end, "end");
+    },
+    { selector: INPUT, inputType, data },
+  );
+}
+
+const { defaultBrowserType: _iphoneBrowser, ...iPhone13 } = devices["iPhone 13"];
+
+test.describe("Live terminal IME syllable rewrite", () => {
+  test.use(iPhone13);
```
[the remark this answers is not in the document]
jerome-benoit: Not applying: the mocked configs (`playwright.config.ts` and siblings) pin a single chromium project, and Playwright forbids a describe-level `browserName` override (it would force a webkit worker that has no project, skipping the whole file). Dropping `defaultBrowserType` is the documented convention across the sibling device specs, e.g. tests/backspace-autorepeat.spec.ts:49-52.

web/src/lib/mobileKeyboardProxy.ts
jerome-benoit: Correcting my earlier reply: #3862 did not merge. The toolbar invalidation landed with #3692 itself. Tab, Escape, Ctrl and repeated arrow sends invalidate the relevant retained context before bypassing the textarea. The corrected browser model emits no backward-delete event from an empty shadow, so the Tab scenario expects 한\t하, not 한\t\x7f하. The remaining refusal/upload/replay paths and the composition/surface ownership gaps are handled by #3893; the maintainer’s synchronous paste fix is #3895.

jerome-benoit (dismissed):
Thanks for the session-boundary fix and for recording the remaining work in #3877. Re-reviewed the eight-file diff at `d7d3096a28f66001a06256768f52201adacb79cc`, including your reply and the resolved threads.

- **Fixed:** `App.tsx:882–888` clears the persistent textarea; the regression at `live-terminal-ime-rewrite.spec.ts:108–121` tests that persistent element rather than an unmounted input. The previous P2 is satisfied.
- **P2, still required before approval:** the retained textarea must not record rejected/transformed input or survive edits that bypass it. `mobileKeyboardProxy.ts:56–61` allows the native edit regardless of delivery outcome; `MobileLiveTerminal.tsx:1709–1714,1733` can send Ctrl+C instead of `c` and return false, but that result never reaches the forwarding helper. The textarea therefore records text absent from the PTY. Likewise, out-of-band input leaves its previous text retained. Please propagate delivery/retention semantics and invalidate both textarea entry points at bypass boundaries, with a regression. This is the existing finding, now precisely tracked by #3877, not an additional inline comment.

I classify the remaining problem as P2 rather than the earlier P1: its demonstrated scope is conditional mobile-input corruption, not universal failure. However, the retention behavior is introduced by this PR, so moving it into an issue does not remove the correctness dependency. The fix may be separate, but should land with or before this change.

Cross-validation: both independent sequential reviewers confirmed the remaining gap; one accepted follow-up treatment, the other considered it blocking. Parent arbitration retains the requirement based on the concrete delivery-result mismatch above. No native-device reproduction was performed. The mocked Chromium test is appropriate handler coverage; the withdrawn WebKit suggestion remains rejected.

`git diff --check 79ab4e40...d7d3096` is clean. CI is not fully green: [core-rest](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34489510925/job/102913358342) failed with `docker: Text file busy` in an unchanged Rust fixture (terminal.rs has identical base/head blobs). This is separate from the web finding. GitHub currently refuses its rerun while the parent workflow is running. No local build/test was run.

Issue #3152, opened 2026-07-29
Rate limit reset datetime still parsed wrong ("reopen" #3028)


cwrau:
I am running AoE on commit a89a9fa28e03b6c2e62499ca9bc1bfb6cf5eb820, but this still happens;

I just got the limit with a reset at 2:40pm Europe/Berlin but AoE thinks the limit resets at 12:08:39. The 12 could be my 2 hours time zone difference in UTC (but AoE should display UTC in the browsers' timezone), but the 8 minutes past is definitely wrong.

cwrau:
Now I triggered something new; I sent a message, am still in the same session limit until 2:40pm, but now AoE parsed 12:51 as the reset time.

Now that I see this, it looks like the "parsed" time is just `$now + 1h`??

Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

Confirmed, and @cwrau's read is exactly right: the displayed reset is `now + 1h`, not a parsed value. Nothing is being "parsed wrong"; nothing is being parsed at all. The `now + 1h` fallback #3061 was supposed to retire is still what fires in the real world.

The timezone half is a red herring. The banner renders with `toLocaleTimeString()`, so the browser's own timezone is already applied:

`web/src/components/acp/StructuredView.tsx:1529`

```ts
const reset = new Date(rateLimit.resets_at).toLocaleTimeString();
```

Cross-checking the second report: the comment landed at `09:52:48Z`, which is 11:52 in Europe/Berlin, and AoE showed `12:51`. That is `now + 1h` rendered in Berlin local time. The displayed clock is correct; the value handed to it is fabricated.

## Evidence

A real `RateLimit` event from my own event store (pre-#3061 build, so it shows the fallback shape unambiguously):

```
sqlite3 ~/.agent-of-empires/acp_events.db \
  "SELECT created_at, event_json FROM acp_events
   WHERE json_extract(event_json,'$.RateLimit') IS NOT NULL;"

1782809245943|{"RateLimit":{"info":{"status":"Internal error: You've hit your weekly limit · resets 4am (Europe/Paris)","resets_at":"2026-06-30T09:47:25.943264Z","kind":"rate_limit"}}}
```

[...]

PR #3693, opened 2026-09-02
fix(acp): cap rate-limit auto-resume redeliveries

### LINKED ISSUE #3688: Rate-limit auto-resume redelivers the interrupted prompt with no retry cap, causing dozens of duplicate turns
## Problem

`sessions.turn.send`'s rate-limit auto-resume path (#1722, #3028) re-delivers the *interrupted
prompt text* as a fresh turn every time a parked session is auto-resumed, with no cap on the
number of times this can happen for one original prompt.

At merged head `507db0cb855e1bee4484b233c1579eb8b1c42751`:

- On a rate-limit park, the interrupted turn's text is captured via
  [`set_pending_initial_turn`](https://github.com/agent-of-empires/agent-of-empires/blob/507db0cb855e1bee4484b233c1579eb8b1c42751/src/server/session_service.rs#L754) ("Used to continue a rate-limit-interrupted turn on resume (#3028)").
- [`reap_rate_limit_resumes`](https://github.com/agent-of-empires/agent-of-empires/blob/507db0cb855e1bee4484b233c1579eb8b1c42751/src/server/acp_reconciler.rs#L1165) publishes `RateLimitAutoResumed` and fresh-spawns the worker once
  the resume window elapses (`RATE_LIMIT_MIN_PARK_SECS`, [`src/server/acp_reconciler.rs#L1103`](https://github.com/agent-of-empires/agent-of-empires/blob/507db0cb855e1bee4484b233c1579eb8b1c42751/src/server/acp_reconciler.rs#L1103); or, when the adapter reported no reset time at all,
  a flat [`RATE_LIMIT_UNKNOWN_RESET_RETRY_SECS = 3600`](https://github.com/agent-of-empires/agent-of-empires/blob/507db0cb855e1bee4484b233c1579eb8b1c42751/src/server/acp_reconciler.rs#L1112) retry interval).
- Once the worker is live, [`drain_pending_initial_turn`](https://github.com/agent-of-empires/agent-of-empires/blob/507db0cb855e1bee4484b233c1579eb8b1c42751/src/server/session_service.rs#L629) calls `send_turn` with the stored
  text, which appears on the transcript/event bus as a brand-new `UserPromptSent`, not a resume of
  the prior turn.

This is the right behavior for a single rate-limit interruption. But when the true rate-limit
window is longer than the adapter reports (or the adapter reports no `resets_at` at all and the
account is actually blocked for days, not the 1-hour fallback interval), there is nothing that
stops the cycle: park → auto-resume → immediately re-hit the rate limit before any tool call
completes → re-park → wait → repeat. Each cycle redelivers the identical original prompt as a new
visible turn.

## Impact

Observed in production plugin-created (cron) sessions over the course of a week:

- One structured/plugin session had the identical prompt redelivered **74 times** over several
  days, one per hourly `RATE_LIMIT_UNKNOWN_RESET_RETRY_SECS` cycle, without a single tool call ever
  completing.
- Two other plugin sessions hit **43 identical redeliveries each** over ~42 hours (same 1-hour
  cadence), zero tool calls, and were only stopped by manual intervention (archiving the session).
- Two more sessions saw **20 redeliveries** each over ~19–20 hours before eventually getting
  through.

Besides the wasted spend (dozens of redundant top-of-conversation turns, most immediately
re-rejected), this silently prevents job completion: a plugin cron job whose trailing cleanup step
(e.g. "archive/trash old sessions of this job") never runs because the job body never gets past its
first tool call. That in turn starves the plugin's active-session concurrency cap (see the
companion issue in `agent-of-empires/plugin-cron`), compounding the outage.

**Priority: P2** (silent, unbounded resource waste and job non-completion; not a crash/data-loss
bug, but genuinely blocks unattended automation from ever finishing).

## Expected fix

- Cap the number of times a single `pending_initial_turn` can be auto-redelivered before the
  reconciler gives up and leaves the session parked with a terminal, visible error (e.g.
  `rate_limit_exhausted_retries`) instead of retrying forever.
- Consider exponential backoff instead of a flat hourly retry for the unknown-reset-time case
  (`RATE_LIMIT_UNKNOWN_RESET_RETRY_SECS`), since a week-scale quota exhaustion currently gets
  retried at the same 1-hour cadence as a genuinely-short park.
- Surface the redelivery count/history somewhere visible (status line, session detail) so a human
  glancing at the session can tell "this has retried 40+ times" without reading raw events.

Found via a routine weekly review of AoE session logs (multiple independently-affected cron-plugin
sessions across the week), not a targeted repro — happy to share sanitized `acp_events.db` excerpts
if useful for a regression test.

Possibly related to the very recently filed #3685/#3686/#3687 (same subsystem area — plugin prompt
admission / dormant-session turn delivery) — flagging in case the audit that produced those already
covers this redelivery-cap gap; if so feel free to close as a duplicate.

atirna:
## Description

The rate-limit auto-resume pass re-delivered the interrupted prompt on every retry with no bound, so a limit that never cleared re-sent the same prompt once per retry interval indefinitely (74 duplicate turns observed on one plugin cron session, zero tool calls completing; two more sessions at 43 each).

I capped the redeliveries per streak:

- the streak is counted off the persisted event log (`RateLimitAutoResumed` breadcrumbs since the last completed turn or agent switch), so it survives daemon restarts. Manual RESUME NOW resumes do not count toward it, and neither do resumes whose spawn failed before the redelivery landed
- at 5 redeliveries (matching the existing crash-loop respawn budget) the reconciler stops re-enqueueing the continuation, clears the pending turn, and parks the session on a terminal `Stopped { reason: "rate_limit_exhausted_retries" }` (seq-guarded publish, same as terminal repair)
- both park predicates (main resume loop + the manual `/acp/spawn` resume marker) hold the session parked on that reason; RESUME NOW still delivers the interrupted turn and a fresh prompt still works, either one starting a new streak
- the structured view banner shows the give-up instead of silently sitting parked

The unknown-reset retry now backs off with the streak (1h, 2h, 4h, 8h, 16h) instead of staying flat. A flat hour and a five-attempt cap give up after five hours, and the issue reports two sessions that only got through at ~20 redeliveries over ~19-20 hours; backing off spends the same five attempts across 31 hours, so those recoveries are still covered.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [ ] For UI changes: included screenshot or recording (text-only addition to the existing rate-limit error banner; covered by the SystemNotices vitest instead)

## Test Coverage Analysis

- [x] Skipped a test, because:

`acp.rate-limit-recovery` already lists every file this touches, so the matrix needs no edit. The banner copy and reducer latch are covered by that existing vitest surface (`web/src/components/acp/SystemNotices.test.tsx` gains the exhausted-park case; `web/src/lib/acpTypes.test.ts` covers the reducer transitions). The daemon-side park decision is a reconciler/event-store contract, not a dashboard flow, so no new Playwright story.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** GLM 5.3 coding agent

**Any Additional AI Details you'd like to share:** Used for investigation and drafting the patch; I reviewed and own the change.

- [] I am an AI Agent filling out this form (check box if true)

Fixes #3688

## Why

A cron/plugin session whose provider quota is exhausted for days burns the same prompt every retry interval and never finishes its job body, which also starves the plugin's session concurrency cap until someone archives the session by hand.

## Verification

- before, on current `main`: a parked session's redelivery count is unbounded (issue reports 74 identical `UserPromptSent` turns, one per hourly `RATE_LIMIT_UNKNOWN_RESET_RETRY_SECS` cycle)
- after: `cargo test --features serve --lib` (6571 passed) including `rate_limit_redelivery_streak_counts_resumes_and_resets_organically`, which walks the reporter's exact park → resume → redeliver → re-park sequence and asserts the streak hits 5, resets on an organic turn end, and ignores resumes past a worker startup failure
- `cargo clippy --all-features -- -D warnings`, `cargo fmt --check`, `cargo check --no-default-features --lib` clean
- web: `vitest run` on the touched specs (105 passed), `npm run format:check`, `npm run lint`, `npx tsc -b` clean

## Maintainer follow-up

Rebased onto main and applied two rounds of review feedback in the last three commits.

**Round one.** The fresh-prompt recovery moved out of `acp_prompt` into `dispatch::decide`, so all three prompt-admission sites get it: `acp_prompt_diff_comments` was refusing an exhausted park outright and losing the review. Settling it under the submission guard also closes the window where a park landing mid-handler stranded the prompt. The cap takes `instance_lock` with `try_lock` rather than blocking the reconciler tick behind a manual resume's handshake. Streak resets moved to the boundaries the reducer already used, and `RateLimitAutoResumed` carries a defaulted `manual` flag so RESUME NOW does not spend the automatic budget.

**Round two**, from re-reviewing those fixes. A prompt POSTed while the session was still on the adapter park sat on the server queue and was stranded when the cap fired. The cap counted resumes rather than redeliveries, so a session whose rate-limited turns are agent-initiated got parked with nothing ever re-sent. The context primer dropped the burned prompt, because `ended_non_success` is reassigned on every `Stopped` and the cap's lands on top of the adapter's, which broke the banner's own handoff CTA. Two round-one tests were vacuous: a held `ResumeReservation` makes `is_running` true and short-circuits the park probe, so both passed with the recovery gate deleted.

**Round three**, same again. Round two's queued-prompt release never ran in a live daemon: it sat below the resume loop's `attempted` skip, and the cap only fires for a session already in `attempted` whose slot the park then keeps. It now takes both halves, each mutation-covered separately. The client had the mirror gap: `sendPrompt` re-enqueues a retryable 503 only for `workerIdleStopped`, so a recovery prompt whose resume missed the readiness window was dropped silently.

**Round four** was a two-model cross-review plus a re-read of the standing review. It found that the cap could not actually be reached in the case the issue documents: five attempts an hour apart cover five hours, against reported recoveries at ~19-20. Backoff (above) closes that. The redelivery budget also moved out of the pruned transcript into a per-session row, because `acp.replay_events` at any supported small value evicted the breadcrumbs the streak was counted from and silently restored the original unbounded loop. Smaller items: the park reason moved from `server::acp_reconciler` to `acp::state` (an `acp` module was importing from `server`), the constant's comment claimed a parity with the #1945 respawn budget that does not hold (that one is a self-forgiving rate limiter, this is a lifetime budget), a manual RESUME NOW rendered in the transcript as "auto-resumed", and `SystemNotices` held two unrelated retry-exhaustion values.

At `cdd259d6e`: `cargo fmt`, `cargo clippy --features web --all-targets`, and `cargo test --features web` (6904 passed) clean, bar three sandbox-only failures in files this PR does not touch. Web `format:check`, `lint`, `tsc -b` clean, vitest 3726 passed.

Not addressed here, and worth a follow-up issue: the visibility half of #3688. `derive_acp_status` maps every `Stopped` to `Idle`, so an exhausted session is indistinguishable from a healthy idle one outside the structured view, and the redelivery count is not surfaced anywhere.

_Note: this section was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Summary

- Caps rate-limit auto-resume redeliveries at five attempts per streak.
- Parks exhausted sessions with the visible `rate_limit_exhausted_retries` state.
- Preserves retry streaks across restarts and resets them after normal progress, new prompts, agent changes, or startup failures.
- Adds structured-view warnings, manual recovery support, documentation, and regression tests.

## Benefits

- Prevents duplicate turns and wasted resources.
- Keeps affected sessions visible and available for recovery.
- Preserves manual resume and fresh-prompt workflows.

## Further enhancement

- Consider exponential backoff when the rate-limit reset time is unknown.


src/server/acp_reconciler.rs
```diff
@@ -1111,6 +1118,22 @@ const RATE_LIMIT_MIN_PARK_SECS: i64 = 30;
...
+/// #1945 respawn budget so both bounded-retry mechanisms stop at the same
+/// attempt count. Every `RateLimitAutoResumed` breadcrumb is one re-delivery;
+/// any organic turn end (a `Stopped` with another reason) resets the count.
+const RATE_LIMIT_AUTO_RESUME_MAX_REDELIVERIES: i64 = 5;
+
+/// Terminal park reason the reconciler publishes when a rate-limit streak
+/// exhausts `RATE_LIMIT_AUTO_RESUME_MAX_REDELIVERIES`. Distinct from
+/// the agent-reported `rate_limited` so the park predicates can hold the
+/// session parked without treating it as a fresh adapter park (no reset
+/// schedule applies), while the manual `/acp/spawn` resume and a new prompt
```
jerome-benoit: [P1] Make the fresh-prompt recovery contract real. An exhausted session has no worker and is not idle-dormant, so `dispatch::decide` returns `Queued { WorkerDown }` before `send_turn` can trigger a resume (`src/acp/dispatch.rs:72-75`, `src/server/api/acp.rs:1389-1426`). The queue drain only wakes idle-dormant sessions (`src/server/acp_reconciler.rs:1656-1707`), while this park remains in `attempted`, so the promised fresh prompt stays queued indefinitely. Explicitly wake this park before dispatch, persist a boundary that distinguishes that true user submission from an auto-redelivered `UserPromptSent`, and add a server-level exhausted-park -> fresh-prompt -> delivered-turn regression test.

src/server/acp_reconciler.rs
```diff
@@ -1281,6 +1304,58 @@ async fn reap_rate_limit_resumes(state: &Arc<AppState>, attempted: &mut HashSet<
...
+                        session = %id,
+                        error = %e,
+                        "rate-limit redelivery streak probe failed"
+                    );
+                    continue;
+                }
+            }
+        };
+        if streak >= RATE_LIMIT_AUTO_RESUME_MAX_REDELIVERIES {
+            state.session_service.clear_pending_initial_turn(&id).await;
```
jerome-benoit: [P1] Do not mutate the continuation before the sequence CAS. After `latest_seq` is sampled, a concurrent `RESUME NOW` can finish spawning and enqueue the interrupted prompt (`src/server/api/acp.rs:493-501`); this call then clears it, and `publish_stopped_if_seq` can return false because the manual `RateLimitAutoResumed` advanced the sequence. Its boolean is ignored, leaving a live worker with no continuation. Serialize cap finalization with manual recovery, or condition the clear on the same session generation/epoch as the successful park, and cover the controlled interleaving. Merely moving the clear after the CAS is insufficient because the manual enqueue can land between them.

jerome-benoit (changes requested):
Requesting changes for two recovery-path correctness issues: a fresh prompt neither wakes an exhausted park nor has a durable fresh-streak identity, and the cap path can erase a concurrent manual-resume continuation before its sequence CAS.

I verified the five-attempt boundary, the agent-switch reset, the reducer transitions, the warning banner, and the existing coverage-matrix mapping. `git diff --check` and `cargo fmt --all -- --check` pass at `7995c95ac660928f285991ff7255c33772e47aa9`. I did not rerun the full Rust/web suites; the current GitHub checks cover only the PR template and title.

njbrake (changes requested):
**Request changes.** One blocker: the fresh-prompt recovery the banner promises does not work.

### CI

No workflow run has executed on this branch; the runs sit unapproved behind the fork gate. Local results at `a414622`:

`cargo fmt --all -- --check`, `cargo clippy --features serve --all-targets`, `cargo check --no-default-features --lib` all clean. `cargo test --features web`: 6580 passed, 2 failed. Both failures are `chmod 0o000` fixtures in `supervisor.rs` and `worker_registry.rs` that cannot deny a root reader; neither file is in the diff. Web: `format:check`, `lint`, `tsc -b` clean, vitest 3721 passing. Playwright not run.

### Blocker: an exhausted session swallows a fresh prompt

`Stopped { rate_limit_exhausted_retries }` leaves no live worker and no idle-dormant marker, so `dispatch::decide` returns `Queued { WorkerDown }` (`dispatch.rs:72`) and `acp_prompt` buffers the prompt before reaching `send_turn` (`api/acp.rs:1389`). `drain_queued_prompts` wakes only the dormant branch (`acp_reconciler.rs:1656`) and its fall-through defers to the resume pass, which this PR taught to hold the park forever. A local repro asserting all three (dispatch is `WorkerDown`, queue still length 1 after three full reconcile ticks, no worker spawned) passes.

Not the existing `rate_limited` behavior: with auto-resume on, that park clears itself and the next tick's drain delivers the queue. The exhausted park never clears, and auto-resume on is the only configuration where the cap fires.

The banner tells the user to send a new prompt, and that is the path that fails. Route this park through `send_turn` the way `plugin/session_api.rs:582` already routes `WorkerDown`; `trigger_resume_background` resumes independently of the park predicate. A persisted fresh-submission boundary is not needed, since publishing the park already resets the streak.

### Ordering bug at `acp_reconciler.rs:1341`

`clear_pending_initial_turn` runs before the CAS and the CAS result is discarded, so a refusal destroys the continuation and publishes no park. `publish_stopped_if_seq` compares the in-memory `next_seqs` counter against a `highest_seq` read from the log, which its own doc notes trails the counter. A manual `/acp/spawn` landing between the probe and the clear loses its continuation and leaves a live idle worker. Condition the clear on the CAS returning true. Narrower than a lost prompt, but the ordering is wrong independent of the race.

`reap_rate_limit_resumes` has no test coverage and both findings live in it. A backdating seam on `EventStore::record`'s `created_at` would make the `now < resume_at` gate drivable.

### CodeRabbit at `event_store.rs:1245`: not reproducible

The exhausted park is itself a `Stopped` whose reason is not `rate_limited`, so it becomes the streak's own boundary and the count returns to zero the moment it lands. A fresh prompt rate-limited before completing starts from zero with a full budget.

### The #3688 design question

atirna's account holds. The scheduler reads only `info.resets_at` (`acp_reconciler.rs:1281`); banner text is never parsed, by design in #3152 (`acp_client/rate_limit.rs` module doc). With an epoch present, `rate_limit_resume_at` returns `max(resets_at + 15s, recorded_at + 30s)`, so cwrau's (B) already ships. The reported hourly cadence is `RATE_LIMIT_UNKNOWN_RESET_RETRY_SECS`, not a daily window being ignored.

On (A): `agent-client-protocol-schema` 1.5.0 has no turn-continuation primitive. `session/resume` reattaches a session and still requires a following `session/prompt`. RESUME NOW is itself the re-send (`api/acp.rs:493`), so dropping redelivery leaves that button inert.

The cap is the right shape and 5 matches #1945. It does not address the starvation half of #3688: an exhausted cron session sits `Idle` holding the plugin's concurrency slot and still needs manual archiving. `derive_acp_status` maps every `Stopped` reason to `Idle` (`acp_events.rs:899`), so nothing outside the structured-view banner shows the state, which is the visibility the issue asked for.

### Smaller

- `docs/structured-view/troubleshooting.md:186` still documents unbounded retry ("the next retry is another hour out"). User-facing and stale as of this change.
- The new `SystemNotices` case mounts `rateLimit: null` and asserts the recovery buttons are absent. `Stopped` never clears `rate_limit` in the fold, so both buttons render in the real park and the reachable banner is untested.
- `AgentStartupError` resets the whole streak rather than excluding the resume that failed to spawn, granting a fresh five.
- Manual RESUME NOW publishes `RateLimitAutoResumed`, so five clicks spend the automatic budget.
- The daemon does not reset the streak on `AgentSwitched`, so the new backend inherits the old one's spent budget. The second commit clears this in the reducer only.
- The coverage-matrix box is checked but nothing under `web/tests/` changed. `acp.rate-limit-recovery` already lists both touched files, so no edit is needed and the "skipped" box fits.
- The event-store test writes its terminal stops at `seq + 15`, so it does not walk the reported ordering. Its assertions still hold.

src/acp/event_store.rs
```diff
@@ -1219,6 +1219,66 @@ impl EventStore {
...
+            Ok(g) => g,
+            Err(p) => p.into_inner(),
+        };
+        conn.query_row(
+            "SELECT COUNT(*) FROM acp_events r
+             WHERE r.session_id = ?1
+               AND r.discriminant = 'RateLimitAutoResumed'
+               AND IFNULL(
+                     json_extract(r.event_json, '$.RateLimitAutoResumed.manual'), 0) = 0
+               AND r.seq > (
```
jerome-benoit, line 1293: [P1] Keep the retry budget valid under configured retention. `record()` prunes every insert to `max_events_per_session`, while this query counts only retained `RateLimitAutoResumed` rows. With the supported `replay_events = 3`, reproducing two retry cycles and pruning after each insert leaves `UserPromptSent, RateLimit, Stopped`; this query returns `0`, so the five-attempt cap can never be reached. Persist the budget outside the pruned transcript, or otherwise retain enough durable state for every supported history cap, and add a small-retention regression test.

jerome-benoit (changes requested):
Requesting changes at `a58a2be21`.

Two blockers remain:

1. The redelivery budget is derived exclusively from retained events. With a small valid `replay_events` value, retention removes earlier `RateLimitAutoResumed` breadcrumbs and `rate_limit_redelivery_streak()` returns zero again, making the five-attempt cap ineffective. I left an inline comment with the reproduction and recommended fix.

2. The existing fresh-prompt race remains unresolved. Prompt admission owns `prompt_submission`, while cap finalization owns the distinct `instance_lock`. A prompt can observe the old `Stopped { rate_limited }` state before the reconciler publishes `rate_limit_exhausted_retries`, then select `Queued { WorkerDown }` after the terminal park. A fresh plugin turn that wins this race can also inherit the previous prompt's spent budget because organic and auto-redelivered `UserPromptSent` events remain indistinguishable. Please serialize these transitions, persist an explicit fresh-submission boundary, and add a controlled interleaving regression test. I am not duplicating the existing unresolved CodeRabbit thread for this blocker.

One additional recovery defect should be fixed: `context_primer` treats only `Stopped { rate_limited }` as unsuccessful (`src/acp/context_primer.rs:278-320`). After `rate_limit_exhausted_retries`, **Continue in another agent** creates a handoff with an empty `[MY NEXT REQUEST]`, so the interrupted request is not continued. Give the exhausted terminal the same unprocessed-prompt semantics and cover that switch flow.

The earlier manual-resume/CAS continuation race and the reducer state transitions are fixed. `git diff --check` passes. I did not rerun the complete Rust or web suites; current GitHub checks cover only the PR template and title.

src/acp/event_store.rs
atirna: reproduced on the current head, replay_events=3: after two park -> resume -> redeliver -> re-park cycles the retained rows are exactly `UserPromptSent, RateLimit, Stopped` and the streak query returns 0, so the cap never fires. pinning `RateLimitAutoResumed` would not save it either, since the count also depends on the next-of-`UserPromptSent` lookup over the same pruned table.

so the budget needs to live outside the pruned log (side table updated on `record`, or a per-session counter) as you said. @njbrake documented the caveat in the internals note in a58a2be instead of fixing it, and the boundary/budget distinction also decides whether a fresh prompt inherits the previous one's spent budget. happy to put up the durable-budget fix on this branch if that's the shape you two want; otherwise a `min` on `replay_events` would also close it.

src/server/api/acp.rs
```diff
@@ -1337,6 +1346,23 @@ pub async fn acp_prompt(
         Err(rej) => return rej.into_response(),
     };
     let woke_idle_dormant = touch_on_prompt_and_wake_if_sunk(&state, &id).await;
+    let exhausted_rate_limit_park = {
```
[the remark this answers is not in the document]
atirna: settled in 8bdc8c7: `begin_prompt_submission` resolves the park under the same guard as the dispatch it feeds, so a park landing mid-handler can no longer strand the prompt; covered by `exhausted_rate_limit_park_is_sendable_at_the_shared_decision_point`.

src/acp/event_store.rs
atirna: landed in 610bb55: the budget is now a per-session row in a new acp_rate_limit_budgets table, updated inside record() in the same transaction as the event insert, so prune_retention can no longer evict it. reads fall back to the breadcrumb query only for sessions with no row yet (daemon upgraded mid-park), and the first relevant write seeds that row from the log so the two derivations agree.

replay_events=3 regression covers the full reported cycle: five park -> resume -> redeliver -> re-park rounds keep the streak at 5 with only the last cycle's three rows retained (the pruned-log query returns 0 there). also added a cap=1 variant, and the seed path.

fail-before-fix on both retention tests: left: 0, right: 5 exactly as you described.

src/server/acp_reconciler.rs
```diff
@@ -1290,7 +1420,7 @@ async fn reap_rate_limit_resumes(state: &Arc<AppState>, attempted: &mut HashSet<
         enqueue_rate_limit_continuation(state, &id).await;
         state
             .acp_supervisor
-            .publish_rate_limit_auto_resumed(&id, resume_at);
+            .publish_rate_limit_auto_resumed(&id, resume_at, false);
```
jerome-benoit, line 1455: [P2] Publish an armed retry only when a continuation was actually queued. `enqueue_rate_limit_continuation` returns early on `Ok(None)`, but this unconditional automatic breadcrumb still makes `update_rate_limit_budget` set `armed = 1`. The next unrelated fresh `UserPromptSent` then increments `spent`, contradicting the documented rule that agent-initiated wakeup/monitor/loop resumes with no prompt must not consume the budget. Return whether the continuation was queued, persist that distinction, and cover `auto-resume with no continuation -> fresh prompt` without spending a retry.

src/events/mod.rs
```diff
@@ -860,6 +876,15 @@ pub fn delete_topic(conn: &Connection, schema: &Schema, topic: &str) -> usize {
...
     }
+    if let Err(e) = conn.execute(
+        &format!(
+            "DELETE FROM {} WHERE session_id = ?1",
+            schema.rate_limit_budgets_table()
+        ),
+        params![topic],
+    ) {
+        warn!(target: "events", "delete rate-limit budget {topic}: {e}");
+    }
```
jerome-benoit, line 887: [P2] Keep permanent CLI purges in sync with this new persisted table. `EventStore::delete_session` reaches this helper, but `purge_acp_transcript_rows` in `src/cli/mod.rs:151-164` still deletes only `acp_events` and `acp_attachments`; both `aoe rm --purge` and `aoe session empty-trash` therefore leave the session's budget row behind. Include `schema.rate_limit_budgets_table()` in that purge transaction and extend its regression test. A focused SQLite reproduction leaves counts `events=0`, `attachments=0`, `budgets=1`.

src/acp/transcript.rs
```diff
@@ -469,7 +469,7 @@ impl TranscriptModel {
                     format!("mode switch to \"{mode_id}\" failed: {reason}"),
                 ))]
             }
-            Event::RateLimitAutoResumed { resets_at } => {
+            Event::RateLimitAutoResumed { resets_at, .. } => {
```
jerome-benoit: [P2] Surface the new terminal park in the native structured view as well. `TranscriptModel` still drops every `Stopped` reason (`src/acp/transcript.rs:347-365`), while only the web reducer consumes `rateLimitRetriesExhausted`; the primary TUI therefore presents an exhausted session as ordinary idle silence. This arm also ignores `manual` and labels RESUME NOW as `auto-resumed`. Add a terminal notice/recovery hint and render manual versus automatic recovery accurately, or keep the visibility portion of #3688 open with an explicit follow-up instead of claiming it is fixed here.

docs/development/internals/structured-view.md
```diff
@@ -86,7 +86,7 @@ Three layers recover a turn that stops progressing, in increasing depth:
...
...a misbehaving adapter cannot drive a respawn loop. Each resume re-delivers the interrupted prompt once (`pending_initial_turn`, #3028); that redelivery is capped at 5 per streak, so a limit that keeps rejecting parks the session on a terminal `Stopped { reason: "rate_limit_exhausted_retries" }` instead of re-sending the same prompt forever (#3688). The streak is counted from the persisted `RateLimitAutoResumed` breadcrumbs since the last organic turn end or agent switch, skipping manual RESUME NOW resumes and any whose spawn failed before the redelivery landed; see `EventStore::rate_limit_redelivery_streak`. The park has no schedule, so nothing un-parks it on a timer: `dispatch::decide` treats it like idle dormancy, which is what makes the banner's "send a new prompt" true, and a prompt already on the server queue when the cap fires releases the park through `reap_rate_limit_resumes` instead. Because the streak is read back off the log, a profile that sets `acp.replay_events` low enough to evict a whole park cycle (roughly 30 events) loses the count and stops capping; the default of 0 retains everything. The banner's "Continue in another agent" CTA runs the agent-switch path below.
```
jerome-benoit: [P3] Update this retention caveat. The new durable `acp_rate_limit_budgets` row is specifically outside transcript pruning, and the added `replay_events = 1/3` tests assert that the cap remains effective. Keeping the old statement now documents behavior that this head no longer has; replace it with the durable-budget behavior and, if useful, the bounded upgrade-time limitation.

jerome-benoit (changes requested):
Requesting changes at `610bb55640e94f02d2fd19bc65ac81299ed6446e`. The previously reported retention, fresh-prompt admission, continuation-CAS, and context-primer defects are fixed. Three remaining correctness and lifecycle gaps are annotated below, plus one stale documentation statement.

Validation: the head SHA was rechecked immediately before submission, all prior review threads are resolved, and `git diff --check` passes. I used focused static traces and SQLite state-transition reproductions; I did not rerun the full Rust or web suites. The only current GitHub check on this head is the successful PR-title lint.

PR #3709, opened 2026-09-02
feat(tui): pre-size every open session's pane to its preview rect

njbrake:
## Description

Selecting a session or entering live view waited out a passive resize round-trip because only the selected session's pane was kept at preview geometry. Each session's preview rect is predictable (shared preview area plus its own header height), so render now reconciles every open session (Running, Waiting, or Idle; not archived, trashed, or structured) through the existing passive-resize worker and its atomic detached/no-owner guard.

Bounds: resizes fire only after two consecutive refreshes want the same fleet-wide geometry (a key independent of the selection, so cursor movement never re-arms); a declined resize (attached client, size owner, missing session) is parked until that geometry changes or 30s pass. The viewed session's resize is front-inserted and the worker drains one intent at a time, so it never waits behind the fleet. The shared list-panes snapshot now carries observed window size (stamped pre-fork), and a synced pane resized behind our back (external attach, web live view) is detected and re-asserted. Fleet reconciliation runs only while a single aoe TUI is alive; with more than one, only the selected-session sync runs, so concurrent TUIs cannot fight over background panes. Resizes target the first window explicitly. No visual change; panes resize off-screen.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is a background tmux resize with no rendered output; the debounce, dedup, epoch, priority, and observed-size invalidation logic is covered by unit tests, which are the cheapest tests that catch the regressions.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Fable 5 via Claude Code CLI

Design discussed and directed by @njbrake; implementation, tests, and this description drafted by the model. Cross-reviewed pre-merge by independent Claude Opus and Claude Sonnet review runs; their confirmed findings are fixed in the follow-up commits.

_Note: this PR was drafted by Claude Fable 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

https://claude.ai/code/session_01XtCkD2LW9GMQV4gjEWfM18

## Summary

- Adds fleet-wide pre-sizing for eligible open session panes.
- Waits for stable geometry before sending resize requests.
- Processes one resize request at a time and prioritizes the viewed session.
- Detects external pane size changes and re-applies the expected size.
- Retries declined requests after geometry changes or 30 seconds.
- Tracks resize state per session instead of globally.
- Targets the first tmux window consistently.
- Updates pane observation timestamps and test fixtures.
- Adds unit tests for coordination, retries, invalidation, and single-TUI behavior.

## Benefits

- Selection and live-view transitions avoid passive resize delays.
- Resize work remains off-screen.
- Tracking reduces duplicate requests and resize oscillation.
- Archived, trashed, and structured sessions remain unchanged.

## Further considerations

- Add an end-to-end TUI test for fleet resizing.
- Continue testing unusual session-state and geometry transitions.


src/tui/home/render.rs
```diff
@@ -2247,6 +2292,192 @@ impl HomeView {
...
+        self.invalidate_externally_resized_panes();
+        if inner.width == 0 || inner.height == 0 {
+            return;
+        }
+        // The excluded (selected) and live sessions are skipped at the firing
+        // loop below, NOT while building `wants`: the armed epoch key must be
+        // pure geometry, or moving the selection would re-arm the fleet and
+        // retry every declined session on each cursor stop.
+        let mut wants: Vec<(String, u16, u16)> = Vec::new();
+        for (id, inst) in &self.instances {
```
jerome-benoit, line 2406: **[P1] Coordinate fleet sizing across concurrent TUIs**

Each `HomeView` keeps these sync maps locally, but every TUI now walks and resizes every open session. With two TUIs using different preview rectangles, each process observes the other process's size at `invalidate_externally_resized_panes`, drops its local synced entry, and re-applies its own geometry. The final tmux guard only checks `session_attached == 0` and the size-owner lease; passive resize takes no inter-process owner. The project already tracks this supported case through `active_tui_count` and explicitly notes that multiple TUIs clash over pane sizes (`render.rs:4001-4006`). This change expands that clash from viewed panes to the whole fleet, producing repeated SIGWINCH/reflow and tmux work. Please elect a single passive fleet owner, or disable fleet-wide reconciliation when more than one TUI is active while retaining only the prior selected-session behavior. Add coverage with two home views requesting different rectangles for the same sessions.

src/tmux/mod.rs
```diff
@@ -1288,13 +1303,24 @@ fn parse_pane_metadata(output: &str) -> HashMap<String, PaneMetadata> {
                     .contains(utils::PANE_ENV_FILE_PREFIX),
                 pane_title: (!pane_title.is_empty()).then(|| pane_title.to_string()),
                 window_activity,
+                window_size,
```
jerome-benoit, line 1306: **[P2] Observe and resize the same tmux window**

This metadata intentionally keeps the first window's pane 0, matching the preview/capture target `session:^.0`. However, `resize_window_if_format_with_deadline` computes chrome for `:^.0` but runs `resize-window -t <session>` (`src/tmux/session.rs:1835-1848`). Per tmux target-window semantics, omitting the window selects that session's current window. If a detached session's current window is not its first, the worker resizes a different window while recording this first-window observation as the verification signal; the preview remains unsized and the new invalidation loop can keep retrying. Please target the first window explicitly throughout resize, guard, chrome calculation, and observation, and add a multi-window regression test.

src/tmux/mod.rs
```diff
@@ -1288,13 +1303,24 @@ fn parse_pane_metadata(output: &str) -> HashMap<String, PaneMetadata> {
...
 
     map
 }
 
+/// Observed window geometry for `session_name` from the shared list-panes
+/// snapshot, with the snapshot's publish time. `None` when the snapshot is
+/// absent, failed, or does not include the session.
+pub(crate) fn observed_window_size_from_cache(session_name: &str) -> Option<((u16, u16), Instant)> {
+    let cache = PANE_META_CACHE.read().ok()?;
+    let time = cache.time?;
```
jerome-benoit, line 1320: **[P2] Preserve the snapshot's observation time**

`cache.time` is stamped in `publish_pane_meta_cache` after `batch_pane_metadata` returns (`src/tmux/mod.rs:1715-1739`), but `passive_synced_contradicted` treats it as the time the sizes were observed. A `list-panes` call can therefore read the old size, pause while the resize completes and is adopted, then publish that old value with a timestamp newer than `adopted_at`. Render misclassifies it as an external resize and sends another SIGWINCH. Capture the snapshot timestamp before issuing `list-panes` and carry it with the result; add a test where a pre-resize observation is published after adoption.

jerome-benoit (changes requested):
Reviewed at exact head `9896a2b0051cc4daee90bc5847cc849b8447bb4b` after two independent passes and manual cross-validation. CodeRabbit already covers the order-sensitive fleet key, so I have not duplicated it. CI is green at this SHA. The fleet-wide design still introduces one cross-process resize fight and two correctness issues in the observation loop that should be addressed before merge.

src/tui/home/render.rs
jerome-benoit: Resolved at `091e1a82aa0babdacb853d40602eda66a6b7edc3`: fleet reconciliation and external-size invalidation now stop when `active_tui_count > 1`, while the pre-existing selected-session sync remains available. The new unit test covers disabling and resuming the fleet pass.

src/tmux/mod.rs
jerome-benoit: Resolved at `091e1a82aa0babdacb853d40602eda66a6b7edc3`: `resize-window` now explicitly targets `session:^`, matching the chrome probe, capture path, and observed-size source. The tmux-backed test covers a later current window.

src/tmux/mod.rs
jerome-benoit: The production ordering is fixed at `091e1a82aa0babdacb853d40602eda66a6b7edc3`: `taken_at` is captured before `list-panes` and preserved on publication. One requested part remains: there is still no cache-boundary regression proving that a pre-resize observation published after adoption cannot invalidate the adopted size. Reverting `cache.time` to publication time would leave the current tests green. Also update `observed_window_size_from_cache` at `src/tmux/mod.rs:1314-1316`, which still describes this value as the snapshot's publish time.

jerome-benoit (changes requested):
Rereviewed at exact head `091e1a82aa0babdacb853d40602eda66a6b7edc3` after two independent passes and manual cross-validation. The concurrent-TUI and first-window findings are resolved, and the timestamp implementation is correct. I resolved those two threads. The timestamp thread remains blocking only for the requested cache-boundary regression test; its helper documentation must also be corrected from publication time to observation time. CI is green at this SHA.

PR #3715, opened 2026-09-02
fix(session): resolve a custom agent's built-in before capture and resume

### LINKED ISSUE #3638: Custom agents never capture a conversation id and never emit --resume: every restart starts fresh
## Summary

A custom agent that wraps a supported one (`[session.custom_agents]` + `[session.agent_detect_as]`) never captures `agent_session_id` and never receives a resume flag. Restarting the session drops the conversation, and `/resume` inside the pane is the only recovery.

Capture and resume key off `Instance::tool` raw. For a custom agent that is the config key (`claude-personal`), which has no `AgentDef`, so every `get_agent()` lookup on that path misses. `resolved_agent()` (`src/session/instance/accessors.rs:189`) already exists to fall through `agent_detect_as` for this reason, and the launch and hook paths use it. Capture and resume do not.

## Environment

aoe 1.15.1, macOS 15.7.4, tmux 3.7c, Docker sandbox, wrapper sets `CLAUDE_CONFIG_DIR` then execs `claude`.

```toml
[session.custom_agents]
claude-personal = "claude-personal"

[session.agent_detect_as]
claude-personal = "claude"
```

## Evidence

Across five profiles: 18 of 20 built-in `claude`/`codex` sessions carry an `agent_session_id`; 0 of 50 custom-agent sessions do. In one profile on a single day, 5 `claude` sessions captured a sid and 21 `claude-personal` sessions did not. `debug.log` holds 418 `Session ID observed` lines, none for a custom-agent session. Not poller-budget exhaustion: 1094 hits, all on one day, one session.

## Three layers

**1. No poller starts.** `supports_session_poller()` (`src/session/instance/polling.rs:19`) returns false on the `get_agent(&self.tool)` miss. Past that, the poll-fn dispatch at `polling.rs:35` matches literal built-in names and falls to `_ => return` at `polling.rs:301`.

**2. No resume flag is emitted.** `build_resume_flags` (`src/session/instance/launch_command.rs:81`) bails on the same miss and returns an empty string, so even a manual `aoe session set-session-id` pin produces no `--resume`.

**3. Sandbox scan reads the wrong directory.** `claude_container_list_snippet` (`src/session/capture.rs:429`) resolves `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` from the `docker exec` environment, which aoe sets to `/root/.claude` via `container_env`. The wrapper exports `/root/.claude-personal` inside the pane, after aoe has decided. The transcript is real and on disk at `~/.claude-personal-sandbox/projects/<encoded-cwd>/<uuid>.jsonl`; the scan looks elsewhere. #3399 fixed the equivalent host probe for a profile `environment` entry, not the container read.

Fixing only layer 1 leaves the sid unusable, and fixing 1 and 2 leaves sandboxed wrappers with nothing to capture. All three need to move together.

## Repro

```bash
aoe add /path/to/repo --title wrapper-resume --tool claude-personal --sandbox
aoe session start wrapper-resume
# converse, then:
aoe session stop wrapper-resume && aoe session start wrapper-resume
```

The pane comes back on an empty conversation. `aoe session show wrapper-resume` reports no session id at every point.

## Acceptance

- A sandboxed and a host custom-agent session each persist their own conversation id with no manual pin.
- Restart continues the same conversation.
- Two custom-agent sessions in one project never cross-claim ids.
- A custom agent whose `agent_detect_as` base has `ResumeStrategy::Unsupported` still emits no resume argument.

## Related

#3138 refactors the same dispatch blocks but puts `supports_session_poller()` out of scope. #3376 covers agents whose native resume flag is unverified, which Claude's is not. #3422 is the structured-view half of the same wrapper problem.

njbrake:
## Description

If you point AoE at a custom agent that is really just a wrapper around a
supported one, a script that sets an environment variable and then runs
Claude, AoE lost the conversation every time you stopped and restarted the
session. It never learned which conversation the session was having, and it
never asked the agent to continue one, so restarting always dropped you into
an empty chat. The only way back was typing `/resume` inside the pane
yourself. This changes AoE to look up the built-in agent a wrapper stands in
for before deciding any of that, so a wrapped session now gets a conversation
id at launch, keeps it, and continues the same conversation across restarts.

The lookups were keyed on the session's own tool name. A wrapper's name is
its own, not a built-in's, so every lookup missed and each miss failed
closed: no session-id poller, no id pre-minted at launch, no resume flag.
`resolved_agent()` already exists to fall through `agent_detect_as`, and the
launch and hook paths already use it. Capture and resume now do too.

Two places deliberately keep reading the raw tool:

- The empty-thread transcript probe. A wrapper is free to redirect
  `CLAUDE_CONFIG_DIR` itself, which reads there as "no transcript on disk"
  and would downgrade a live conversation to a `--session-id` Claude rejects,
  killing the pane (#3399). Keeping it raw is what makes the reported setup
  work at all.
- The peer-exclusion tool argument, since peers are matched on their own
  stored `tool` and resolving one side only would stop two sessions of the
  same wrapper from seeing each other.

Inside a container, `docker exec` inherits the environment AoE injected at
`docker run`, not the pane's, so a wrapper exporting its own
`CLAUDE_CONFIG_DIR` wrote where the transcript scan never looked. The scan
now tries each config dir a live process in the container declares before
falling back to the injected value. A declared `agent_config_dir` cannot
serve here: the user mounts that directory through their own
`extra_volumes`, and AoE never sees its container path.

Fixes #3638

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

`tests/e2e/custom_agent_resume_e2e.rs` drives the full binary with a wrapper
shim that records its argv and writes no transcript, so nothing on disk can
supply an identity. Two wrapper sessions share one project directory: each
launches with its own `--session-id`, persists it, and the restart emits
`--resume` for the same conversation without moving the peer's.

Unit coverage alongside it: the launch line for a wrapper (pin, then resume,
and silence when the base agent's `ResumeStrategy` is `Unsupported`), the
poller gate, and the container scan preferring a running process's config
dir over the injected one.

Local runs: `cargo fmt`, `cargo clippy --all-targets` clean. `cargo test`
6578 pass / 2 fail and the e2e suite 181 pass / 9 fail; every one of those
failures reproduces on `origin/main` in the same tree (two force an error
with `chmod 0o000`, which does not stop root, and the rest are pre-existing
in this sandbox).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 via the Claude Code CLI.

Diagnosis and patch drafted by the model through back-and-forth with
@njbrake; the scoping calls, including which two lookups stay keyed on the
raw tool, are his.

## Summary

- Fixed conversation ID capture and resume for custom agents that wrap supported built-in agents.
- Applied resolved-agent handling to polling, recovery, resume, transcript capture, sidecar reconciliation, and peer isolation.
- Improved container transcript discovery for wrapper-defined configuration directories, including paths with spaces.
- Added safeguards for unsupported resume strategies and launcher commands that cannot accept resume flags.
- Added documentation, unit tests, regression tests, and end-to-end coverage for wrapper launch, resume, recovery, poller behavior, and peer isolation.

## Benefits

Custom-agent sessions now preserve conversations across host and sandboxed restarts. Sessions no longer claim conversation IDs from related agents. Resume behavior remains disabled when the wrapped agent does not support it.


src/session/instance/session_id.rs
```diff
@@ -692,7 +701,7 @@ impl Instance {
             }
         }
         let emitted = append_resume_flags(
-            &self.tool,
+            resume_tool,
```
jerome-benoit, line 768: [P1] Preserve the wrapped binary position for subcommand resumes. `resume_tool` can now resolve a custom agent to Codex, but `append_resume_flags` delegates to `splice_subcommand_or_append`, which inserts the subcommand after the first whitespace-delimited token. A command such as `ssh host codex` therefore becomes `ssh resume <sid> host codex`, while `env CODEX_HOME=/x codex` becomes `env resume <sid> CODEX_HOME=/x codex`; restart targets the wrong command or fails. Require an explicit command-aware insertion point, or inherit Codex resume only when the command is proven to be a direct Codex invocation, and add a multi-token wrapper regression test.

src/session/capture.rs
```diff
@@ -769,7 +804,13 @@ pub(crate) fn compose_exclusion_with_persisted_peers(
         {
             set.insert(parked.to_string());
         }
-        if !include_inactive_same_tool || inst.tool != current_tool {
+        // Matched on the built-in each side resolves to, not on the raw
```
jerome-benoit, line 822: [P1] Resolve parked ownership across aliases. The inactive-current-SID comparison below uses the resolved capture agent, but the parked lookup immediately above still does `prior_tool_session_ids.get(current_tool)`. `swap_tool` stores that map under the raw outgoing tool, so a peer that parked `claude-personal` is invisible to `claude` or to another Claude alias after the SID has moved out of `agent_session_id`. A fresh/MRU scan can then adopt the parked conversation. Match parked entries by their resolved capture/store identity and add base-to-wrapper plus wrapper-to-wrapper swap tests.

src/session/instance/retroactive_capture.rs
```diff
@@ -33,8 +39,8 @@ impl Instance {
     }
 
     pub(crate) fn try_retroactive_capture(&self) -> Option<String> {
-        let result: Option<String> = match self.tool.as_str() {
-            "claude" => {
+        let result: Option<String> = match self.capture_agent_name() {
```
jerome-benoit, line 40: [P1] Keep the shared-CWD contention guard on the same identity as this dispatch. This now lets an alias scan its built-in agent store, but `contended_capture_key` still returns the raw `(self.tool, cwd)`. A live id-less `codex` session and `codex-personal` wrapper in one cwd are therefore not marked contended, so `aoe status` or `session show` may let the first self-heal persist the other pane conversation from the MRU store. `foreign_sid_holder` prevents duplicate persistence only after a SID is selected; it cannot prove that the first attribution was correct. Key contention by the resolved capture/store identity and cover mixed aliases.

jerome-benoit (changes requested):
Reviewed at exact head `bbcec21db0b9d2e2095d90b2feff152a665b436e` after two independent blind passes and manual cross-validation. CI is green at this SHA (35 successful, 1 skipped).

Three new blocking ownership/command-construction findings are inline.

One additional P2 is outside the diff: startup recovery still calls `should_attempt_resume(..., &inst.tool)` in `src/session/recovery.rs:125-132`, so a custom wrapper with a valid stored ID is not a recovery candidate after an unexpected pane loss. `orphan_needles` also determines hook capability from the raw tool at lines 145-165. Please use the resolved capture agent for these recovery predicates and cover startup recovery, not only explicit stop/start.

I did not duplicate CodeRabbit comments. Its Pi finding is blocking, but changing only `pi_session_id_pinnable` is insufficient: custom-agent commands are command overrides, while the extension, sidecar, poller, absorb, and teardown-flush gates still require raw `self.tool == "pi"`. Its whitespace-splitting finding in the container transcript scan is a valid P2 and should also be fixed in this PR.

jerome-benoit (commented):
Post-merge review of exact head `e79aaf5bda80547c344c76f70635207fd70efc7b`, cross-validated through two independent sequential passes and manual arbitration.

@njbrake, one process note: patience is a virtue in FOSS. There was no stated deadline that required merging while the requested-changes rereview and CI were still in progress. Please wait for outstanding review work to finish before merging; a short delay is cheaper than post-merge corrective work.

The earlier ownership, recovery, path-with-spaces, and Kimi alias findings are fixed. The following remain:

- **[P1] Codex wrappers do not preserve their store identity.** `polling.rs:196-213` resolves the wrapper to Codex, but `capture.rs:1769-1774,1984-1998` reads only AoE's ambient `CODEX_HOME`. A wrapper that exports its own `CODEX_HOME` writes to a store this poller never reads, so capture/resume can miss the wrapper conversation or inspect an unrelated ambient store. Carry an authoritative store identity into capture, or use an attributable sidecar, and test with a decoy ambient store.
- **[P1] Pi wrappers still receive neither an attributable capture channel nor a resumable pin.** The existing CodeRabbit thread remains valid: `session_id.rs:620-626` requires raw `self.tool == "pi"` and rejects command overrides, while the Pi poller requires its sidecar. This needs an end-to-end capability design, not only a raw-name substitution.
- **[P1 security] The fallback warnings log the full custom command.** The existing CodeRabbit thread remains valid at `session_id.rs:701-704,745-749`; arbitrary command strings may contain credentials. Log a non-sensitive classification instead.
- **[P2] Validation and insertion disagree on whitespace.** `session_id.rs:665-676` uses `split_whitespace()`, while `launch_command.rs:150-158` searches only for `' '`. The deterministic check produced `codex\t--model resume SID o3` from `codex\t--model o3`. Use one parser/offset for both operations and cover tabs and leading whitespace.
- **[P2] The fail-closed fresh launch is still accounted as pinned.** `session_id.rs:739-750` clears only the local SID; `launch_command.rs:312-325` and `start.rs:182-195,280-283` retain it as `pinned_prior_sid`, causing `resume.rs:340-364` to run the pinned-launch probe although no SID was emitted. Represent "resume omitted" separately so it does not incur the probe or produce a false pinned-SID error.

CI eventually completed with 34 successful and 2 skipped checks and no failures. That does not cover the deterministic gaps above.

PR #3724, opened 2026-09-02
fix(server): stop background drains recreating a deleted session's prompt lock

### LINKED ISSUE #3687: Prevent background drains from recreating prompt locks after deletion
## Problem

Delayed background drain tasks can recreate a deleted session's `prompt_locks` entry after permanent deletion has already removed it.

At merged head `576dd7a950e7f5636784d86ba848816cdff68dd2`:

- pending-initial and queued drains call raw [`prompt_submission`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/session_service.rs#L629-L644) ([queued path](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/session_service.rs#L1191-L1207));
- the reconciler spawns them from earlier session snapshots ([pending](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/acp_reconciler.rs#L1542-L1568), [queued](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/acp_reconciler.rs#L1587-L1615));
- deletion removes the row and calls [`forget_prompt_lock`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/api/sessions.rs#L4489-L4493).

If deletion completes after the reconciler snapshot but before the spawned drain reaches `prompt_submission`, the drain creates a fresh mutex, observes the missing row, and returns without removing that entry.

## Impact

The daemon can retain one prompt mutex per deleted session that races a delayed internal drain. This is bounded by real session IDs and is not the random-ID memory-growth issue fixed by #3651, but it persists for the daemon lifetime.

**Priority: P3**

## Expected fix

Use `prompt_submission_for_session` in both drain entry points, including its post-acquisition existence check, or add an equivalent non-vivifying admission primitive. Add a deterministic deletion-race test that asserts the prompt-lock registry returns to its prior size.

This is separate from #3673's explicitly accepted `queue_enqueue` race: background drain tasks alone can recreate the entry.

Found during a cross-validated post-merge audit of #3673. This issue was drafted with Codex under @jerome-benoit's direction and reviewed before submission.

njbrake:
## Description

When you permanently delete a session, the daemon is supposed to forget everything it was holding for it. One of those things is a small per-session lock that makes sure two prompts are never sent to the same agent at once. Background jobs that deliver a session's leftover prompts are started from a slightly stale list of sessions, so one of them could start running just after the session was deleted. When that happened, the job created a brand new lock for the session that no longer existed, noticed the session was gone, and gave up without cleaning the lock up again. The daemon kept that leftover lock until it was restarted, one for every deleted session unlucky enough to hit the timing.

Both delivery jobs now ask for the lock through the variant that refuses to hand one out for a session that does not exist, and that also throws the lock away if the session disappears while the job was waiting its turn. Nothing changes for a session that still exists; a deleted one no longer leaves anything behind.

Both drain entry points (`drain_pending_initial_turn`, `drain_queued_prompts_once`) took the raw `prompt_submission` guard, which auto-vivifies a `prompt_locks` entry for any id. They now take `prompt_submission_for_session`, whose pre-check declines an unknown id and whose post-acquisition re-check retires an entry vivified after the delete's own `forget_prompt_lock` ran.

Separate from #3673's explicitly accepted `queue_enqueue` race; that call site is unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the regression is daemon-internal registry bookkeeping with no observable surface, so a unit test that inspects `prompt_locks` is the cheapest test that can catch it. `drains_leave_no_prompt_lock_for_a_deleted_session` covers both race windows and fails on `main` in each.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Implementation and test written by Claude Opus 5 under @njbrake's direction, then reviewed by a separate fresh-context review pass before this PR left draft.

Fixes #3687

https://claude.ai/code/session_016gmsyoz33u23NWFj55yGGx

## Summary

- Updated both background prompt-drain paths to use `prompt_submission_for_session`.
- Prevented deleted sessions from recreating per-session prompt locks.
- Added deterministic test synchronization to cover both deletion race windows.
- Added a test-only submission-claim notification hook.

## Benefits

- Deleted sessions no longer leave stale prompt locks.
- Existing sessions keep their current behavior.
- The regression test reliably verifies prompt-lock cleanup.

### DIFF
--- a/src/server/session_service.rs
+++ b/src/server/session_service.rs
@@ -584,11 +584,11 @@ impl SessionService {
     ///
     /// Single drain owner: callers (the create fast path and the reconciler
     /// tick) race through the `pending_drains` claim, and the delivery runs
-    /// under the session's [`Self::prompt_submission`] guard, so the turn
-    /// cannot be published twice concurrently and a direct prompt cannot
-    /// decide its own disposition mid-delivery. A delivery failure leaves the
-    /// field set; the reconciler tick retries once the worker is live.
-    /// Clearing writes memory first,
+    /// under the session's [`Self::prompt_submission_for_session`] guard, so
+    /// the turn cannot be published twice concurrently and a direct prompt
+    /// cannot decide its own disposition mid-delivery. A delivery failure
+    /// leaves the field set; the reconciler tick retries once the worker is
+    /// live. Clearing writes memory first,
     /// then disk: a crash (or failed persist) between the forward and the
     /// disk clear re-delivers after restart, which is the documented
     /// at-least-once contract.
@@ -1129,9 +1129,10 @@ impl SessionService {
     /// Drain the leading batch of a session's server-owned queue into the live
     /// worker once the current turn has ended. Mirrors
     /// `drain_pending_initial_turn`'s single-owner `pending_drains` claim +
-    /// [`Self::prompt_submission`] delivery, so a batch is never sent twice
-    /// concurrently and the idle check below cannot be invalidated by a direct
-    /// prompt deciding its own disposition before this one reaches the agent.
+    /// [`Self::prompt_submission_for_session`] delivery, so a batch is never
+    /// sent twice concurrently and the idle check below cannot be invalidated
+    /// by a direct prompt deciding its own disposition before this one reaches
+    /// the agent.
     ///
     /// Only drains an idle turn, and asks the live control fold rather than
     /// `Instance.status`: dispatch parks a prompt on that fold, so gating the
--- a/src/server/session_service.rs
+++ b/src/server/session_service.rs
@@ -194,6 +194,10 @@ pub struct SessionService {
     /// Per-session prompt-submission locks. See
     /// [`SessionService::prompt_submission`].
     prompt_locks: RwLock<HashMap<String, Arc<tokio::sync::Mutex<()>>>>,
+    /// Test-only tap on [`SessionService::prompt_submission`], fired before it
+    /// reaches `prompt_locks`. See [`SessionService::watch_submission_claims`].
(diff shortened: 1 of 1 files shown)


jerome-benoit (changes requested):
Reviewed exact head `ce5220b3f74470968e1152e04f2c4dbb6cfa163c` through two independent passes and manual cross-validation. CI is green.

Requesting changes for the existing CodeRabbit P2, which I independently verified and will not duplicate inline. The second race case uses a fixed delay plus `!drain.is_finished()`, which proves only that the task has not completed, not that it passed the initial `session_exists()` check before deletion. It can therefore return through the initial check and pass without exercising the post-acquisition cleanup this regression test is meant to protect.

Please synchronize on an observable checkpoint immediately after the initial existence check, then delete the session and release the registry lock.

The production change itself looks sound: both drain paths use the non-vivifying helper, guards are released in the correct order, and I found no production regression. The separate `queue_enqueue` race is unchanged and explicitly outside this PR's scope.

PR #3726, opened 2026-09-02
fix(tui): reject preview drag start below rendered content

### LINKED ISSUE #3660: Reject preview drag selection below rendered content
## Problem

`PreviewTextView::contains` accepts every cell inside the preview pane whenever `total_lines > 0`, even when the content occupies fewer rows than the pane. `screen_to_content` then clamps a click in the empty lower area to `total_lines - 1`.

For example, with a 20-row pane and 3 content lines, starting a drag on row 10 is accepted and anchors the selection on line 3.

## Impact

Mouse selection can start on or extend to text the user did not click. This is a user-visible **P2 (minor)** interaction bug.

## Expected fix

- Make hit-testing require the pane-relative row to be below `total_lines - first_line`, in addition to the existing pane and non-empty checks.
- Add focused cases for short non-empty content, including a scrolled view, while preserving clamping for an already-active drag that leaves the pane.

This behavior predates #3595, but its [unresolved review thread](https://github.com/agent-of-empires/agent-of-empires/pull/3595#discussion_r3889418238) surfaced it. It was independently cross-validated at merged head `e8c47e4fcdc0072cdefcf2b0cafaef58b3f6b65d`; no split-induced behavior change was found.

njbrake:
## Description

Clicking in the empty space below short preview output used to start a text
selection anyway, and that selection silently latched onto the last line of the
actual content. So if a session had printed three lines into a twenty-row pane
and you pressed the mouse down somewhere in the blank area underneath, aoe
behaved as though you had clicked on the third line and began dragging from
there. Now a press in that blank area does nothing, and selection only starts on
a row that is really showing text. Dragging itself is unchanged: once a
selection is under way, moving the cursor past the end of the content still
extends it to the last line, the way it always did.

`PreviewTextView::contains` accepted any cell inside the pane rect whenever
`total_lines > 0`; `screen_to_content` then clamped the row to
`total_lines - 1`. Hit-testing now also requires the pane-relative row to fall
within the painted rows (`total_lines - first_line`), which covers both the
terminal preview (`compute_scroll`) and the structured transcript
(`TranscriptGeometry`), since both guarantee `first_line + height <= total_lines`
when the content overflows and `first_line == 0` when it does not. `contains` has
a single caller, `handle_drag_start`, so drag moves and auto-scroll keep their
existing clamping.

Second commit fixes a repaint gap the first one exposes. A rejected press falls
through to the app loop's catch-all click branch, which dismissed a finalized
highlight without drawing; nothing downstream repaints for a bare `Down(Left)`,
so the highlight stayed painted until an unrelated wake-up. That branch now
draws when the clear changed something.

Fixes #3660

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Interaction-only change with no visual difference to capture: the fix is that a
press below the content no longer paints a highlight. The two unit tests assert
that directly.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the change is pure coordinate math on the drag-start hit test, which unit tests in `src/tui/home/tests.rs` cover directly (short content at first_line 0 and scrolled into history, plus the drag-move clamp). A full-binary test would need synthetic mouse input for no extra signal.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Fix and tests drafted by the agent, reviewed by @njbrake.

https://claude.ai/code/session_01Nn6GHG48PAbRQ81L5YWB4E

## Summary

- Prevented preview drag selection from starting in blank space below rendered content.
- Preserved clamping when an active drag moves outside the preview.
- Redraws the TUI when a rejected click clears a finalized highlight.
- Added tests for short content, scrolled views, drag clamping, and highlight clearing.

## Benefits

Users can select only visible content, and cleared highlights no longer remain on screen.

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -1450,8 +1450,14 @@ impl App {
                                     self.home.handle_diff_click(mouse.column, mouse.row);
                                     self.draw(terminal)?;
                                     None
+                                } else if self.home.clear_preview_selection() {
+                                    // A click on no surface at all still
+                                    // dismisses a finalized highlight, and
+                                    // nothing below repaints for a bare
+                                    // Down(Left), so draw the clear here.
+                                    self.draw(terminal)?;
+                                    None
                                 } else {
-                                    let _ = self.home.clear_preview_selection();
                                     None
                                 }
                             } else {
--- a/src/tui/home/input.rs
+++ b/src/tui/home/input.rs
@@ -716,8 +716,9 @@ impl HomeView {
         }
         // Seed the selection in content coords so it survives a scroll
         // and can span more than one page. `contains` also requires the
-        // pane to hold real scrollback, so a drag over an empty / not-yet
-        // -captured pane is a no-op rather than a phantom selection.
+        // cell to sit on a painted content row, so a drag over an empty
+        // pane or the blank area below short content is a no-op rather
+        // than a phantom selection.
         let view = self.preview_text_view;
         if view.contains(col, row) {
             let cell = view.screen_to_content(col, row);
--- a/src/tui/home/preview.rs
+++ b/src/tui/home/preview.rs
@@ -21,12 +21,19 @@ pub(in crate::tui) struct PreviewTextView {
 
 impl PreviewTextView {
     /// True when `(col, row)` lands on a row/col that maps to real
-    /// content. Used to gate drag-select start.
+    /// content. Used to gate drag-select start. Rows in the pane below
+    /// the last painted line are rejected: `screen_to_content` clamps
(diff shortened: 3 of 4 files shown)


jerome-benoit (approved):
Reviewed at d75805b4bc6b757f211a9ff53b4475b2ba51d558.

The painted-row hit test is consistent with both preview geometries, preserves active-drag clamping, and safely handles empty or partially painted panes. The catch-all click path now repaints only when it actually clears a finalized selection. The regression coverage includes empty, short, scrolled, and out-of-content cases.

No actionable findings.

PR #3736, opened 2026-09-03
fix(tmux): advertise the key that returns to aoe after an attach

BTForIT:
## Description

The intro dialog (`intro.rs`) and the aoe session status bar (`status_bar.rs`) promised that `prefix d` comes back to aoe. That holds only when aoe attaches by `attach-session`. When aoe itself runs inside tmux (a `tmux new-session … aoe` wrapper, a mosh mirror session, a tmux window running `aoe`), attach is a `switch-client`, and `prefix d` drops the whole client while the TUI keeps running unseen. The key that returns such a client is the stock `prefix L` (`switch-client -l`).

This PR changes the two hints, not the keymap (an earlier revision installed a `bind-key`; dropped per review):

- `status-left` now carries `#{?client_last_session,L back to aoe,d to detach}`, so tmux paints the right key per client: `Ctrl+b L back to aoe` for a client that arrived by `switch-client`, `Ctrl+b d to detach` for one that arrived by `attach-session`. The bar is painted once at session creation by whichever process created the session, while the attach path is chosen later by the attaching process, and one session can be entered both ways, so a per-client format is right in every case. When the TUI's session is destroyed tmux clears `client_last_session` and the hint falls back to `d`, exactly when `switch-client -l` would fail.
- The intro's tmux-mode blurb names `L` or `d` from `attach_return_key()`, which reads the same `$TMUX` test the three attach paths now share as `inside_tmux()`.
- `docs/quick-start.md` gets one sentence for the inside-tmux case.

Verified on an isolated tmux 3.4 server with real clients: two clients on one aoe session, one switched in from a `tui` session and one attached directly, expand the same `status-left` to `Ctrl+b L back to aoe` and `Ctrl+b d to detach`; killing `tui` flips the first to `d to detach`. A client whose last session is an aoe paired terminal (`aoe_term_*`) also gets `L`, and `switch-client -l` returns it there.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the change is the text of two hints. The emitted `status-left` format, conditional included, is pinned by the unit test `status_left_hint_follows_the_clients_attach_path` in `src/tmux/status_bar.rs`; which branch the conditional takes is tmux's own `client_last_session` behaviour, verified live on an isolated tmux 3.4 server as described above. An e2e test would need a second pty client switched into the session, which the harness does not model.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Authored, unit-tested and verified on an isolated tmux server by an AI agent operating the BTForIT account on behalf of its owner; the review rework followed the maintainers' direction on the PR.

## Summary

- Replaced custom tmux keybinding logic with tmux’s built-in session switching.
- Updated status-bar hints, intro text, and quick-start documentation to show `prefix L` or `prefix d` as appropriate.
- Simplified tmux detection across session attachment paths.
- Removed binding installation, ownership checks, tokenization helpers, and related tests.
- Updated environment scanning to ignore single-letter constants.

This avoids global keybinding changes and keeps return guidance accurate for each tmux client.


src/tmux/utils.rs
```diff
@@ -380,6 +380,108 @@ pub(crate) fn kill_session_if_present(name: &str) -> Result<()> {
...
+    }
+}
+
+/// Install the `prefix d` binding described on [`detach_return_condition`].
+/// Runs on every attach — one cheap, idempotent tmux call — so a restarted
+/// tmux server picks it up again. Never fails the attach: a tmux error is
+/// logged and the stock behaviour stands.
+pub(crate) fn install_detach_return_binding() {
+    let listing = match crate::tmux::tmux_command()
+        .args(["list-keys", "-T", "prefix", "d"])
```
jerome-benoit: **[P1] Query the full key table before treating `d` as unbound**

On tmux 3.7c, `list-keys -T prefix d` exits successfully with empty stdout even when `prefix d` has a custom binding. `should_install_detach_return("")` consequently returns `true`, and every attach replaces the user's server-global binding.

Reproduced on an isolated server: the filtered command returned 0 bytes while `list-keys -T prefix` reported `bind-key -T prefix d display-message user-custom-d`.

Query the full `prefix` table, select the `d` entry structurally, check the command status, and only replace the stock binding or an exact match for AoE's own binding. Please add a live tmux regression test preserving a custom binding.

src/tmux/utils.rs
```diff
@@ -380,6 +380,108 @@ pub(crate) fn kill_session_if_present(name: &str) -> Result<()> {
...
+/// `switch-client -l`; everywhere else it stays `detach-client`.
+///
+/// tmux evaluates the condition itself (`if-shell -F`, no shell fork) and
+/// clears `client_last_session` when that session is destroyed, so a vanished
+/// TUI session degrades to a plain detach rather than a stuck client.
+fn detach_return_condition(managed_glob: &str) -> String {
+    format!(
+        "#{{&&:#{{m:{glob},#{{session_name}}}},\
+         #{{&&:#{{!=:#{{client_last_session}},}},\
+         #{{==:#{{m:{glob},#{{client_last_session}}}},0}}}}}}",
```
jerome-benoit: **[P2] Preserve return to a TUI launched from an AoE terminal**

The condition rejects every managed `client_last_session`. If the TUI is launched from an AoE paired terminal and then switches to an agent, the previous session is the still-running TUI but its name starts with `aoe_`, so `prefix d` detaches instead of returning to it.

The exact condition evaluates to `0` for `current=aoe_agent_target, last=aoe_term_source`, versus `1` for a plain source session. Track or mark the actual return target instead of inferring that every managed previous session is invalid, and cover this nesting path.

jerome-benoit (changes requested):
Requesting changes at `c3bf0097500bb5e51a9c1cd33a7a7a7ba8666a7d`.

The existing CodeRabbit finding about accepting any binding containing `switch-client -l` is valid and should also be addressed; I have not duplicated it inline. The first inline finding is more fundamental: on tmux 3.7c, the query returns no binding at all, so every custom `prefix d` binding is treated as absent.

The three relevant attachment implementations do install the helper, the format expression works for a non-managed TUI source, and binding-installation failures remain non-fatal. The new comments should also be shortened and have their em dashes removed per `AGENTS.md`.

`git diff --check` passes. No full Rust CI result exists at this SHA.

jerome-benoit (approved):
Re-reviewed at `0ed71fa2a7383a5e6746dd908b9c8d4e19fd7649`, following the maintainer's direction and your response. Removing the binding machinery resolves the previous P1/P2 and ownership findings without altering users' keymaps.

Two independent reviews found no new blocker. On an isolated tmux 3.7c server, two clients on the same target correctly expanded the hint to `L` after switching and `d` after direct attachment; `switch-client -l` returned to `aoe_term_source`, and deleting that source changed the hint to `d`. The test server was stopped afterward. `git diff --check` passes.

[CodeRabbit's intro-fallback finding](https://github.com/agent-of-empires/agent-of-empires/pull/3736#discussion_r3943103884) is valid but nonblocking: an inherited `TMUX` can make the intro predict `L` before attachment falls back to `d`. Wording that describes both paths would avoid that prediction. No duplicate inline added.

CI was approved and is still running; the earlier template failure was rerun against the updated description. No full local Rust test run is claimed.

src/tui/dialogs/intro.rs
```diff
@@ -748,11 +748,11 @@ impl IntroDialog {
...
         // how to come back out. Title indent is 2 columns (marker + space);
         // body indent is 6 columns so the body reads as nested under the
-        // title. Tmux's detach key uses the user's actual prefix
-        // (`tmux_prefix_display()`), so the hint is correct on a
-        // remapped-prefix setup.
+        // title. The tmux hint uses the user's actual prefix and the key
+        // that undoes this process's attach path (`attach_return_key()`).
         let prefix = crate::tmux::tmux_prefix_display();
-        let tmux_back = format!("      tmux pane. {prefix} then d comes back to aoe.");
+        let key = crate::tmux::attach_return_key();
```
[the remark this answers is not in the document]
BTForIT: Done in 4c7013ac. From inside tmux the intro now says `L (or d)`, since the switch can fall back to a fresh attach when there is no client to switch; from outside it still says `d`. The choice lives in a small pure helper with a unit test for both cases.

Reply written by an AI agent operating the BTForIT account on behalf of its owner; the change was tested as described above.

jerome-benoit (approved):
Re-reviewed `4c7013ac9e871932c77cc6beeeccd696952e75e6`, including your response and the resolved intro-fallback thread. Both independent sequential analyses and final arbitration found no blocking issue.

The intro now covers switch-client and its attach-session fallback (`tmux/utils.rs:355-360`). The scanner change removes the `L` key false positive without dropping any current environment-variable constant; direct environment-read patterns remain unchanged (`isolatedEnv.test.ts:215-220`). No keybinding changes were reintroduced.

[Vitest logs](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34147873907) confirm all 3 `isolatedEnv` tests pass, with 3738 tests passing overall. The initial `core-rest` failure is `docker: Text file busy` in the unchanged `container_terminal_command_preserves_runtime_boundaries` test (`terminal.rs:561`). I reran only that job and its dependents; `core-rest` is still running. Code approval does not establish that this CI failure is resolved.

`git diff --check 0ed71fa2 4c7013ac` is clean. No local build, tests or new tmux reproduction were run.

PR #3740, opened 2026-09-04
fix(session): configurable session-id poller budget + backoff on repair instead of a warning per session per tick

BTForIT:
A fleet daemon registering more live sessions than the poller-thread ceiling
(50) re-probed every over-budget session on every 2 s status tick: each tick
paid a failed `maybe_start_poller` and logged a "budget exhausted" +
"Failed to start session poller" pair per session (~2.5 lines/s from two
sessions on 2026-09-04, 70k lines per session in debug.log), and the ceiling
itself was a hard-coded constant, so a larger fleet could not raise it.

- `[session] session_id_poller_max_threads` (default 50, 0 = default): the
  process-wide ceiling, applied once at startup from the global config for the
  daemon, every TUI and the CLI. `session_id_poller_budget()` /
  `session_id_poller_budget_exhausted()` expose the live counters.
- `PollerRepairBackoff` on `Instance` (`#[serde(skip)]`, carried through
  `merge_runtime_fields` and `merge_runtime_from_reload` like the poller
  handle): `repair_session_id_poller_if_needed` checks the budget BEFORE
  attempting a start and, on a budget miss or a failed start, schedules the
  next attempt at 5 s doubling to 60 s, logging on the first miss, each
  escalation and every ~10 min at the cap. A successful start resets it.
- The daemon's repair walk runs on a clone; deferrals recorded there are now
  copied back to the live row, otherwise the next tick would re-probe as if
  nothing had been scheduled.

Tests: poller budget under a configured ceiling (serial), zero keeps the
default, backoff schedule / reminder cadence / reset, instance repair defers
then recovers once the budget frees, config default + parse.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01V9iLV4JtVZrNReuaXRpj4B

https://claude.ai/code/session_01V9iLV4JtVZrNReuaXRpj4B

## Summary

- Added a configurable session-ID poller thread limit. The default is 50, and `0` keeps the default.
- Added retry backoff for failed or resource-limited poller repairs. Delays increase to 60 seconds and reset after success.
- Applied the limit at startup and exposed poller budget information.
- Preserved repair state across reloads, merges, and repair walks.
- Added tests for configuration, budget handling, backoff behavior, and recovery.

## Benefits

This reduces repeated repair attempts and warning logs when resources are unavailable. Operators can tune the poller limit through global session configuration.


src/main.rs
```diff
@@ -397,6 +397,16 @@ async fn run(
         migrations::run_migrations()?;
     }
 
+    // Process-wide poller budget: the daemon and every TUI each run their own
+    // set of session-id poller threads, capped per process. Read from the
+    // global config (the daemon serves all profiles) before any session is
+    // loaded, so the first repair walk already sees the configured ceiling.
+    agent_of_empires::session::poller::configure_session_id_poller_max_threads(
```
jerome-benoit, line 404: **[P1] Preserve this reintroduced setting through migration v022.** `src/migrations/v022_prune_tuning_settings.rs:58` still removes `session_id_poller_max_threads`, and CLI migrations run at line 397 before this value is loaded. An installation upgrading from a schema below v22 therefore silently loses its tuned ceiling and starts with 50, recreating the fleet failure this PR is meant to fix. Remove the key from v022's dropped-key list, migration documentation, and removal assertion, then add an upgrade regression proving the value survives.

src/session/instance/polling.rs
```diff
@@ -357,9 +357,43 @@ impl Instance {
...
+            return false;
+        }
         self.session_id_poller = None;
         self.maybe_start_poller();
-        self.session_id_poller_is_running()
+        if self.session_id_poller_is_running() {
+            self.poller_repair.reset();
+            true
+        } else {
+            self.defer_poller_repair(now, "start failed");
```
jerome-benoit, line 390: **[P2] Do not report every no-start outcome as a failed repair.** `maybe_start_poller_since` intentionally returns without a poller for sandboxed Copilot (lines 256-263), Kimi (270-278), and Prime Agent (288-295), while their resume strategies make `supports_session_poller()` true. OMP can also return when capture metadata is unavailable. These live sessions will now defer forever and emit misleading `start failed` warnings suggesting a larger budget. Return a typed outcome such as `Started`, `NotApplicable`, `BudgetExhausted`, or `SpawnFailed`, and only defer/warn for actual failures. Cover the intentionally pollerless cases.

src/session/poller.rs
```diff
@@ -505,6 +631,40 @@ impl Default for SessionPoller {
...
+        prev_count: u32,
+        prev_max: u32,
+    }
+
+    impl BudgetExhausted {
+        pub(crate) fn pin() -> Self {
+            let prev_count = ACTIVE_POLLER_COUNT.load(Ordering::SeqCst);
+            let prev_max = session_id_poller_max_threads();
+            SESSION_ID_POLLER_MAX_THREADS.store(1, Ordering::SeqCst);
+            ACTIVE_POLLER_COUNT.store(1, Ordering::SeqCst);
```
jerome-benoit, line 652: **[P2] Isolate these process-global budget mutations from every poller test.** `#[serial]` only coordinates tests using the same serial-test lock, but unannotated tests such as `test_poller_detects_change` and `test_poller_starts_polling_immediately` start real pollers concurrently. They can consume a slot while this fixture is asserting exact counts, fail under the pinned ceiling, or have their guard decrement a restored count. Inject an isolated budget for tests, or make every poller-starting test participate in the same lock before directly storing these atomics.

src/session/config/mod.rs
```diff
@@ -1188,6 +1188,21 @@ pub struct SessionConfig {
...
+    /// is read from the global config only. 0 keeps the default (50).
+    #[serde(default = "default_session_id_poller_max_threads")]
+    #[setting(
+        label = "Session-id poller threads (restart req.)",
+        widget = "number",
+        min = 0,
+        global_only,
+        advanced
+    )]
+    pub session_id_poller_max_threads: u32,
```
jerome-benoit, line 1211: **[P2] Add the required web settings coverage.** This schema-derived, user-facing setting is exposed by the web Settings UI, but this PR has no `web/` coverage change. `web/AGENTS.md:35-39` requires every user-facing settings change to update `web/tests/coverage-matrix.json` and its applicable test. Add the matrix entry/update and assert the relevant schema and global-only persistence behavior.

jerome-benoit (changes requested):
The configurable budget and bounded retry schedule are directionally sound, but the upgrade path currently deletes the reintroduced setting and the repair path misclassifies intentionally pollerless sessions. I also found test isolation and required web coverage gaps.

I independently verified the existing CodeRabbit thread about resetting `poller_repair` after every successful direct start; it remains valid and is not duplicated below.

No Rust CI result is available for this exact head; GitHub currently reports only the template and title checks.

PR #3747, opened 2026-09-04
fix: add scroll to model dropdown when option list overflows

Istar-Eldritch:
## Description
The model dropdown menu in `SessionConfigControls.tsx` had no `max-height`, so with enough models the menu could grow past the viewport with no way to reach the remaining options. This caps the menu height and makes the option list scroll internally.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

Screenshot From 2026-09-02 22-27-01

## Test Coverage Analysis

- [x] Skipped a test, because: pure styling tweak (caps a menu's `max-height` and adds `overflow-y-auto`); no new behavior to assert beyond what CSS already guarantees

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** claude-sonnet-5 for exploration and changes, claude-opus-5 for review.

Used `claude-sonnet-5` to identify the missing `max-height`/scroll on the model dropdown menu and apply the fix. Automated review process by `claude-opus-5`.

Initially drafted by AI, Manually reviewed by a Human.

## Summary
- Updated the model dropdown to choose the direction with more available space.
- Limited the dropdown height and added internal scrolling for long model lists.
- Recalculates layout when the menu opens, the window changes, or the visual viewport scrolls or resizes.
- Accounts for visual viewport offsets, including on-screen keyboard behavior.
- Added regression tests for short viewports, layout direction, height limits, scrolling, viewport offsets, and test-environment restoration.

Users can view all model options without the dropdown extending beyond the visible viewport.


web/src/components/acp/SessionConfigControls.tsx
```diff
@@ -129,47 +129,49 @@ function ModelDropdown({ option, pending, onSelect }: SubProps) {
       {open && (
         <div
           id={menuId}
-          className="absolute bottom-full left-0 z-30 mb-1 w-64 overflow-hidden rounded-md border border-surface-700 bg-surface-850 shadow-xl"
+          className="absolute bottom-full left-0 z-30 mb-1 flex w-64 max-h-72 flex-col overflow-hidden rounded-md border border-surface-700 bg-surface-850 shadow-xl"
```
jerome-benoit: **[P2] Bound the menu by the available viewport space, not a fixed 18rem.** The menu always opens upward with `bottom-full`, while the app root is `h-dvh overflow-hidden`. If the trigger is less than the menu height plus the 4px gap from the viewport top, as in a short landscape or soft-keyboard viewport, the menu begins above the viewport and its header/first options remain unreachable; scrolling this inner child cannot recover content clipped outside the outer box. Use collision/available-space-aware positioning or derive the maximum height from the trigger and visual viewport. Add a long-list mocked Playwright case with a short viewport that asserts the menu stays in bounds and the first and last options are reachable, then update `web/tests/coverage-matrix.json` as required by `web/AGENTS.md`.

jerome-benoit (changes requested):
The internal scroller works when the fixed 18rem menu fits, but the original reachability problem remains on shorter visual viewports. Please make the menu respect the actual available space and cover that geometry in the existing ACP picker browser story.

web/src/components/acp/SessionConfigControls.tsx
```diff
@@ -106,6 +158,33 @@ function ModelDropdown({ option, pending, onSelect }: SubProps) {
...
+  // `visualViewport` so a software keyboard raising/lowering while the
+  // menu is open (which does not fire `window`'s `resize`) still
+  // recomputes. See `computeMenuLayout`.
+  useLayoutEffect(() => {
+    if (!open) return;
+    const vv = typeof window !== "undefined" ? window.visualViewport : null;
+    const recompute = () => {
+      const rect = ref.current?.getBoundingClientRect();
+      if (!rect) return;
+      setMenuLayout(computeMenuLayout(rect, vv?.height ?? window.innerHeight));
```
jerome-benoit: [P2] Account for the visual viewport origin, not only its height. `getBoundingClientRect()` is layout-viewport-relative, while the visible vertical interval is `visualViewport.offsetTop .. offsetTop + height` ([CSSOM View](https://www.w3.org/TR/cssom-view-1/#dom-visualviewport-offsettop)). With a non-zero `offsetTop`, this can choose the wrong direction and size the menu beyond the visible area. The offset can also change independently through the VisualViewport `scroll` event, which is not observed here. Pass the visual bounds into the calculation, subscribe to `vv.scroll` as well as `resize`, and add a regression with `offsetTop !== 0`.

jerome-benoit (changes requested):
Re-review at exact head `e751c1e9273c398aad40568c52bfbbfef12bdae6`. The previous fixed-height/upward-only P2 is resolved for an unshifted visual viewport, and the short-viewport regression now verifies bounds and real internal overflow. One mobile VisualViewport correctness issue remains inline.

The existing [CodeRabbit descriptor-restoration finding](https://github.com/agent-of-empires/agent-of-empires/pull/3747#discussion_r3940611004) is valid as a P3 test-isolation issue, but it is non-blocking and is not duplicated. The coverage matrix is updated and exact-head CI is green.

jerome-benoit (approved):
Re-reviewed at exact head `84151a8e822c594f58cda005f9bd4ed0ab2a2521`. The VisualViewport origin/scroll correction and the property-descriptor restoration both address the remaining findings. Two independent reviews and a source-level geometry/listener probe found no blocking regression. The required CI workflows were explicitly started and are still running; this approval does not claim a completed CI result.

Issue #3519, opened 2026-08-25
Web terminal view renders OSC 8 hyperlink escapes as text, leaving the link unclickable


os-tohe:
# Me

In the web view all the links are not clickable and their formatting is strange. In terminal I can open the links with ctrl + click, but not in web. Regular click in web does not open it either. Also highlighting the links is usually very buggy, it highlights from a completely different char position than where my mouse is.

Web:

<img width="1648" height="153" alt="Image" src="https://github.com/user-attachments/assets/8fbde1d6-8a64-46fd-8051-b0a97bfa06de" />

Terminal:

<img width="1472" height="142" alt="Image" src="https://github.com/user-attachments/assets/979a7137-0b68-4265-b129-dd396f700f90" />

# AI

## What

In the web dashboard's terminal view, OSC 8 hyperlinks coming out of the agent
are not handled at all. The raw escape bytes are rendered as visible text, the
row ends up much wider than the terminal grid, and the resulting link is not
clickable.

The same session, attached in a real terminal, shows a clean `Google` link that
ctrl+click opens in the browser.

## Repro

1. Start a Claude Code session and open it in the web dashboard's terminal view.
2. Prompt: `Render a clickable link for me, for example to Google.`
3. The dashboard renders the escape wrapper as text (`]8;id=...;https://www.google.com`,
   then `Google`, then `]8;;`) instead of a single `Google` link, and clicking it
   does nothing.
4. Attach the same session in a terminal emulator: one `Google` link, ctrl+click
   opens it.

Environment: tmux 3.4 on Linux, dashboard in a desktop browser.

## Where it comes from

tmux passes OSC 8 through verbatim, terminated with ST (`ESC` `\`), and
`capture-pane -e` output is exactly what `src/server/live_ws.rs` publishes to the
browser as a frame's `content`. Verified on tmux 3.4 with `capture-pane -p -e`
piped through `od -c`: the captured row carries `033 ] 8 ; i d = a b c ;
h t t p s : / / w w w . g o o g l e . c o m 033 \ G o o g l e 033 ] 8 ; ; 033 \`.

The browser-side parser only understands CSI. `web/src/lib/ansi.ts` matches
`ESC [ params final-byte` and nothing else: `ANY_CSI` (`ansi.ts:18`),
`NON_SGR_CSI` (`ansi.ts:25`), `HAS_ANSI` (`ansi.ts:47`). There is no OSC handling
[...]

PR #3749, opened 2026-09-04
fix(tui): open links in the preview

### LINKED ISSUE #3735: TUI: OSC 8 hyperlink target is discarded, so a link whose visible text is not the URL is inert
# Me

I am the submitter and I hit this in daily use. Everything in this section is
quoted verbatim from the session where I ran into it. The prose framing and the
whole `# AI` section below were written by an AI, not me.

> How do I click on the links that you present to me? They seem to be clickable
> but clicking on them doesn't do anything.

After being told to hold Shift:

> shift + click works!

Then, testing three links, where link 1 was a markdown link whose visible text
hid the target and links 2 and 3 had the URL visible as text:

> Link 1 does not work. Links 2 and 3 do. Link 1 doesn't even highlight (i.e.
> underline) when hovered with the mouse.

Link 1 is the bug being raised here. Shift+click itself is not the complaint.

# AI

## What

In the TUI session view, an OSC 8 hyperlink loses its target. The visible text
survives, the URI does not. So a link renders as plain, inert text unless its
visible text happens to be a URL that the host terminal's own URL regex can
match independently.

This is the follow-on to #1181. That fix stopped OSC 8 sequences from eating the
surrounding text, which was the right call, but it did so by stripping the
sequence outright and therefore throwing the target away.

## Repro

1. Start a Claude Code session under `aoe` with default settings.
2. Prompt the agent to print two links, one whose visible text hides the target
   and one whose visible text is the URL:

   ```
   Print these two markdown links and nothing else:
   [the AoE repo](https://github.com/agent-of-empires/agent-of-empires)
   [https://example.com](https://example.com)
   ```

3. In the TUI session view, hover each one, then Shift+click each one (Shift so
   the host terminal handles the click rather than the AoE TUI, which captures
   the mouse by default).

Result:

| Link | Hover underline | Shift+click |
| --- | --- | --- |
| `the AoE repo` (target hidden) | none | nothing happens |
| `https://example.com` (text is the URL) | underlines | opens |

Expected: both open. The first one carries a perfectly good OSC 8 target.

## Evidence that the target reaches tmux and dies after

tmux has the URI. With the session detached and the offending line on screen:

```console
$ tmux capture-pane -p -H -S -60 | sort -u
https://github.com/agent-of-empires/agent-of-empires
https://example.com

$ tmux capture-pane -p -F -S -60 | grep -c H
20
```

`-H` prints the hyperlink URIs held in the grid and the `H` line flag marks rows
carrying hyperlinks, so tmux parsed and stored both targets. The host terminal
never receives the first one: no hover underline means it has no hyperlink there
at all, only characters. The second link works purely because the host
terminal's own URL-matching rules fire on the visible text, not because OSC 8
survived.

## Where it comes from

`strip_osc_st()` in `src/tmux/utils.rs` (added by #1182 to fix #1181) removes
ST-terminated OSC sequences and keeps nothing from the payload:

```rust
(_, Some(s)) => {
    remaining = &payload[s + ST.len()..];
}
```

Nothing is pushed to `result`, so the `8;;<uri>` payload is gone. The function's
own test asserts the loss:

```rust
assert_eq!(
    strip_osc_st("\x1b]8;;https://example.com\x1b\\Click Here\x1b]8;;\x1b\\"),
    "Click Here"
);
```

It is called from `parse_output_text()` in `src/tui/components/preview.rs`:

```rust
pub fn parse_output_text(content: &str) -> Text<'static> {
    let cleaned = crate::tmux::utils::strip_osc_st(content);
    cleaned.into_text().unwrap_or_else(|_| Text::from(cleaned))
```

and also from `src/tui/home/preview.rs`.

Not verified: whether the live VT render path reaches the terminal through this
same function or through a separate path with its own stripping. Both TUI
preview callers are confirmed; the live path was not traced.

## Why stripping was reasonable and is now the blocker

`ansi_to_tui` drops the visible text around ST-terminated OSC, which is what
#1181 was about, and ratatui has no hyperlink cell attribute to carry a target
into anyway. So there is no way to round-trip OSC 8 through a ratatui `Text`.
That is the real constraint, and it means the fix probably cannot live inside
`strip_osc_st`.

## Suggested directions

Rough sketches, not a preference:

1. **Re-emit on the way out.** Keep the grid model free of hyperlinks, but have
   whatever writes the composed frame to the host terminal reinsert `OSC 8` around
   the spans that had one. Needs the target retained alongside the span rather
   than discarded at parse time.
2. **Rewrite rather than strip.** Where a hyperlink's visible text differs from
   its target, render it so the URL is visible, for example `the AoE repo
   (https://github.com/...)`. Ugly, but it makes every link work with the host
   terminal's URL matching and needs no new plumbing.
3. **Give the TUI its own open-link action.** `session.click_action` currently
   takes `live_send` or `none`. A third value that opens the hyperlink under the
   cursor would sidestep the host terminal entirely, and `AOE_OPEN_URL_TO`
   suggests the URL-opening machinery already exists.

Option 2 is the cheap one, option 1 is the correct one.

## Environment

| | |
| --- | --- |
| aoe | 1.15.3 (Homebrew) |
| tmux | 3.7c |
| Host terminal | WezTerm 20260901-002820-4fbd6b8e |
| OS | macOS, Darwin 25.6.0, Apple Silicon |
| Agent | Claude Code |

Relevant settings, all at their schema defaults: `session.mouse_capture = true`,
`session.click_action = "live_send"`, `tmux.vt_live = true`. No tmux client is
attached; the TUI renders the detached pane.

## Related

- #1181 and #1182: made hyperlink text visible in the preview by stripping the
  sequence. This issue is the target that fix discarded.
- #3519: the same target loss in the web terminal view. Different surface, and
  there the escape bytes render as visible text rather than being stripped, but
  a shared fix that keeps the target attached to the span could cover both.

njbrake:
## Description

When an agent printed a link, clicking it did nothing. A link like `the AoE repo` looked like ordinary output and behaved like it, and even a plain URL sitting in the output could only be opened through your terminal's own gesture, never by clicking it in AoE.

For a marked-up link the address was being thrown away long before anything could use it. AoE reads each session's screen through a terminal emulator that understands the "this text is a link" marker but keeps no room to store where the link points, so it dropped the address and kept only the words. The drawing library underneath has the same gap. By the time the text reached your screen there was nothing left to click.

This keeps the addresses to one side, paired with the words they were wrapped around, and finds those words again on the row where they end up. It also spots plain URLs sitting in the output, so both kinds of link behave the same. Links are underlined, so they look like links, and clicking one opens it. AoE passes the link along to your own terminal too, so hovering highlights it and your terminal's usual link gesture works. Over SSH, where the machine running AoE has no browser, a click copies the address to your clipboard instead of failing silently.

Fixes #3735

### How it works

- `src/tmux/osc8.rs`: an incremental scanner over the pane's raw byte stream, modelled on the existing `Osc52Scanner` clipboard tap, pairing each URI with the text it wrapped.
- `src/tmux/vt.rs`: the VT reader feeds that scanner, and `capture-pane -e` seeds are scanned the same way so a link already on screen when a channel arms is not missed. Each channel keeps a bounded, deduped table.
- `src/tui/home/preview.rs`: the capture fallback reads its pairs straight off the frame, before `strip_osc_st` removes them.
- `src/tui/links.rs`: resolves a row to its links from two sources, advertised targets first, then plain URLs in the row's own text. Overlaps go to the advertised target: that is what the pane said the text points at, while a bare match is inference. herdr reached the same two-tier design independently.
- `src/tui/hyperlink.rs`: a ratatui `Backend` that splits a frame into runs by target and wraps each in OSC 8. `Backend::draw` supplies `(x, y)` per cell, which is enough to do this without a hyperlink cell attribute.

Matching on text rather than position is deliberate: it survives the grid scrolling and being reseeded from `capture-pane`, neither of which preserves positions. Known limit: a link whose text wraps across two rows matches on neither.

Only `http`/`https` targets are kept, so pane output cannot hand `file:` or `javascript:` to the browser opener.

`open_url` refuses when no browser the user could see is reachable, rather than reporting an open that never happens. `BROWSER` is honoured unconditionally; an SSH session counts as unreachable unless `DISPLAY` is forwarded; a bare Linux host counts as unreachable without a display server. The click then copies the address instead, which reaches the user's machine over OSC 52.

**This changes pre-existing callers**, not just the new click. The ACP "Open in browser" action and the plugin `open_link` path now toast `no browser you could see from here` on a headless or SSH host where they previously reported success for an open that never happened; the plugin-notification auto-open (`structured_view/mod.rs:655`) discards the `Result` and so is silent both before and after. That is the same bug on those surfaces, so the gate is deliberately not scoped to the new caller. `BROWSER` is the escape hatch for anyone with a working forwarder, and it is documented.

`aoe serve --open` now routes through the same seam. It previously carried its own copy of the reachability rules that refused every SSH session and ignored `BROWSER`, so it and the preview disagreed about the same host; `docs/guides/web-dashboard.md` and the `--open` help text are updated, and `docs/cli/reference.md` regenerated.

Hovering a link shows its target in the status bar. A link's visible text is chosen by whatever runs in the pane, and a plain click opens without confirmation, so the address has to be visible before the click rather than only in the flash afterwards.

Release tarballs now carry `LICENSE`, `THIRD_PARTY_NOTICES.md` and `licenses/`. They previously held only the binary, so the Apache-2.0 §4(a) obligation this PR introduces would not have reached anyone installing a release. `install.sh` moves the binary out by name, so the extra files are inert for installers.

### Consolidation

Two duplications this touched, both collapsed rather than added to:

- The http/https predicate had two copies (`plugin/ui_state.rs`, and the one I first wrote in `tmux/osc8.rs`), now `util::is_http_url`. `looks_like_git_url` in `server/api/git.rs` is a genuinely different predicate and is left alone.
- The scratch-buffer trick that maps a rendered row to display columns had two copies, now `components::text::line_columns`. The drag-select copy reads the same rows for a different gesture, so having the two disagree about column width was a live risk.

### Third-party code

`src/tui/hyperlink.rs` derives its hyperlink state machine from [herdr](https://github.com/herdrdev/herdr) (Apache-2.0). Added `THIRD_PARTY_NOTICES.md` and `licenses/Apache-2.0.txt`, and the module header records what was taken and what changed.

This is the repo's first vendored third-party code, so the declarations that a packager or an SPDX scan reads were updated to match: `Cargo.toml` is now `MIT AND Apache-2.0`, `flake.nix` lists both, and `README.md` names the one file. `deny.toml` already allowed both. Release tarballs previously shipped only the binary, so they now carry `LICENSE`, `THIRD_PARTY_NOTICES.md` and `licenses/` as Apache-2.0 §4(a) requires; `install.sh` and the self-updater both pick the binary out by name, so the extra files are inert for installers.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

UI change is an underline on link text plus a 3-second status flash on click. Both are asserted in tests rather than by screenshot: `link_columns_are_underlined_so_the_text_reads_as_a_link` checks exactly which columns get the underline, and the e2e asserts the underline SGR and the OSC 8 in a capture of the running TUI's own pane.

## Test Coverage Analysis

The web terminal loses the same targets (#3519), but that surface parses ANSI in the browser and is untouched here.

**TUI / CLI (e2e test)**

`tests/e2e/preview_hyperlink_e2e.rs` has three: Terminal view for the `capture-pane` fallback, Agent view for the VT grid (where a trace assertion proves the target came from the channel's stream tap rather than a silent fall back to capture), and a bare URL carrying no sequence at all. Each asserts the underline, the OSC 8 reaching the host terminal, and the URL a real click resolves through the `AOE_OPEN_URL_TO` seam.

Every assertion was verified to fail with its own piece of the fix removed. Unit tests additionally pin bytes captured from a real Claude Code pane, and assert `capture-pane -e` round-trips both hyperlink shapes through a real tmux.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Implementation, tests and this description were drafted by the model in a back-and-forth with @njbrake, who directed the approach, chose the click gesture, asked for the herdr comparison, and did the manual verification.

Two corrections came out of that loop and are worth flagging for reviewers, because the first commit was wrong about both:

1. `capture-pane -e` seeds bypass the reader thread, so a link already on screen when a channel armed lost its target. Fixed, with `seed_records_links_already_on_screen`.
2. The click reported through `UpdateStatus`, which renders a dismissable notice. It is now a self-expiring status flash.
3. The painter and the hit-test both early-returned when no OSC 8 target had been advertised, so a row holding only a plain URL was never scanned. Caught by the bare-URL e2e.
4. The span set could overlap itself, so the click hit-test (first match) and the painter (last write wins) could disagree: clicking `docs` could open a different URL than the terminal reported on hover. Precedence is now resolved once into a non-overlapping set. Found by review.
5. Short or mid-word link texts armed spurious spans, which then stole clicks from a mouse-tracking agent because the link handler runs ahead of the forward. Now needs 3+ characters and word-boundary alignment. Found by review.
6. The click reported `opened <url>` on a host with no reachable browser. `webbrowser::open` returns as soon as it can spawn a helper, so it answered `Ok` while the spawned `xdg-open` failed out of band, and over SSH a browser that did start opened on a screen the user is not at. `open_url` now checks reachability first and the click falls back to the clipboard. Found by @njbrake testing over SSH.

https://claude.ai/code/session_01RVP8dd1Y4Yqbr6wMWbyWrN

## Summary

- Fixed TUI links that use hidden OSC 8 targets.
- Added support for plain HTTP(S) URLs.
- Added link underlines, hover details, click-to-open behavior, and OSC 8 terminal output.
- Added clipboard fallback when no reachable browser is available.
- Added link parsing, rendering, unit tests, and end-to-end tests.
- Improved browser detection and shared it with `aoe serve --open`.
- Updated documentation, licensing notices, and release archives.

## Benefits

Users can open links even when the visible text does not show the URL. Headless and remote sessions can copy links when browser access is unavailable. The added coverage helps protect link handling across wide characters, changing targets, and multiple rendering paths.

## Further Considerations

Future changes should preserve consistent behavior across TUI, live VT, and web preview surfaces.


src/tui/hyperlink.rs
```diff
@@ -0,0 +1,324 @@
...
+        // Start from a known state: an interrupted write, or an active
+        // hyperlink in the host terminal, must not bleed onto unlinked cells.
+        self.inner.write_all(OSC8_CLOSE)?;
+
+        // `CrosstermBackend::draw` keeps no state between calls (it re-derives
+        // colors and cursor position each time and resets SGR at the end), so
+        // a frame can be handed to it one run at a time.
+        let mut run: Vec<(u16, u16, &Cell)> = Vec::new();
+        let mut active: Option<usize> = None;
+        for (x, y, cell) in content {
```
jerome-benoit, line 139: [P2] Repaint when the OSC 8 target changes even if the cell does not. Ratatui calls this backend with only `previous_buffer.diff_iter(current_buffer)`; the URI lives only in `FrameHyperlinks`, so identical text/style with target A -> B produces no cell update. The terminal keeps A while AoE's hit-test/status resolves B. Please include target changes in invalidation (for example, force linked cells to update) and add a two-frame `Terminal::draw` regression with identical cells and different URIs.

src/tui/components/text.rs
```diff
@@ -1,5 +1,58 @@
...
+}
+
+/// Resolve `line` into display columns at `width`.
+pub(crate) fn line_columns(line: &ratatui::text::Line, width: u16) -> LineColumns {
+    let mut buf = ratatui::buffer::Buffer::empty(ratatui::layout::Rect::new(0, 0, width.max(1), 1));
+    buf.set_line(0, 0, line, width.max(1));
+    let mut text = String::with_capacity(width as usize);
+    let mut column_of = Vec::with_capacity(width as usize);
+    for col in 0..width {
+        let symbol = buf[(col, 0)].symbol();
```
jerome-benoit: [P2] Do not append continuation cells as spaces. In the locked ratatui-core 0.1.2, a wide grapheme resets its continuation cell and `Cell::symbol()` returns `" "` for an empty cell. This turns labels such as `日本語` into `日 本 語 ` and prevents OSC 8 matching; a bare URL with a wide path can be truncated to the wrong prefix. Skip continuation columns while retaining their display-column span, and test wide characters inside both a label and a URL.

src/tui/open_url.rs
```diff
@@ -23,14 +32,123 @@ pub fn open_url(url: &str) -> std::io::Result<()> {
...
     webbrowser::open(url)
 }
 
+/// Whether launching a browser here would put it in front of the user.
+///
+/// `BROWSER` is honored first and unconditionally: setting it is a deliberate
+/// choice, and on a remote host it is usually a script that forwards the URL
+/// somewhere useful.
+fn browser_reachable() -> bool {
+    if std::env::var_os("BROWSER").is_some() {
```
jerome-benoit: [P2] Do not treat `BROWSER` as an unconditional escape hatch on macOS. The locked webbrowser 1.2.4 macOS backend calls `NSWorkspace` directly and never reads `BROWSER`. Over SSH, `BROWSER=my-forwarder` therefore passes this guard but does not run the forwarder; a remote AppKit browser may open and this caller reports success instead of copying the URL. Gate this exception to platforms where the launcher honors it, or invoke a documented forwarder explicitly, and add macOS-specific coverage. `DISPLAY` has the same platform caveat here.

jerome-benoit (changes requested):
Requesting changes for three correctness issues in hyperlink metadata propagation, wide-cell mapping, and browser reachability. I did not duplicate CodeRabbit's valid minor comments about selection state and the remaining `AOE_OPEN_URL_TO` test-isolation gap; its URI/text deduplication finding is fixed at this head. CI is not fully green because the live Playwright suite fails in an ACP composer test that appears unrelated to this diff, so I am not treating it as a product finding here.

src/tui/home/preview.rs
```diff
@@ -211,15 +217,78 @@ impl PreviewCache {
...
         if self.content.is_empty() {
             self.parsed_text = None;
+            self.links.clear();
             return;
         }
         if self.parsed_text.is_none() {
             self.parsed_text = Some(crate::tui::components::preview::parse_output_text(
                 &self.content,
             ));
+            self.links = self.collect_links();
```
jerome-benoit, line 237: [P2] Publish OSC 8 metadata changes even when the rendered grid is unchanged. `collect_links()` only runs when `parsed_text` is invalidated, while the live worker publishes only content, cursor, or budget changes. If a pane redraws the same label and final cursor with URI B instead of URI A, vt100 strips both OSC 8 sequences, the sampled content deduplicates, and this cache keeps A; `AlwaysUpdate` then faithfully re-emits the stale target. Carry a link generation/snapshot through `CaptureFrame` and invalidate this cache when it changes. Add an A-to-B regression whose visible cells and cursor remain identical.

src/tui/components/text.rs
```diff
@@ -1,5 +1,71 @@
...
+pub(crate) fn line_columns(line: &ratatui::text::Line, width: u16) -> LineColumns {
+    use unicode_segmentation::UnicodeSegmentation;
+    use unicode_width::UnicodeWidthStr;
+
+    let mut text = String::with_capacity(width as usize);
+    let mut column_of = Vec::with_capacity(width as usize);
+    let mut col = 0u16;
+    for span in &line.spans {
+        for grapheme in span.content.graphemes(true) {
+            let cells = UnicodeWidthStr::width(grapheme) as u16;
```
jerome-benoit: [P2] Match ratatui’s actual cell-width rules. The locked ratatui-core 0.1.2 adds one cell for halfwidth dakuten/handakuten, but `UnicodeWidthStr::width` does not. For example, ratatui paints `ｶﾞ` in two cells while this mapper advances one, shifting every following link underline, OSC 8 span, hit target, and shared selection left. Reuse or mirror ratatui’s `CellWidth` semantics and add a halfwidth-katakana regression.

src/tui/links.rs
```diff
@@ -0,0 +1,512 @@
...
+    for (start, end) in bare_url_ranges(&columns.text) {
+        let end_column = columns.column_at(end);
+        // `columns.text` is the row as painted, so it is cut at the pane's
+        // right edge. A URL running to that edge may be a prefix of the real
+        // one, and opening `https://example.com/log` for a link that reads
+        // `https://example.com/logout` is a wrong action, not a missing one.
+        // An advertised target cannot hit this: a truncated link text simply
+        // fails to match. Rows wider than the pane show up when scrollback was
+        // captured at an older geometry, or mid-resize.
+        if end_column >= width {
```
jerome-benoit: [P2] Reject a URL truncated before a non-fitting wide grapheme. `line_columns` can stop one column short of the pane edge, but this guard rejects only `end_column >= width`. At width 25, `see https://example.com/日本 now` keeps the 24-column prefix through the slash, drops `日`, and accepts `https://example.com/` as a complete URL. Preserve an explicit truncation flag (or equivalent) and reject a bare URL that reaches the retained text boundary when more source text was clipped; cover this wide-edge case.

src/tui/hyperlink.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// renderer marking linked cells `AlwaysUpdate`, the second frame here
+    /// would emit nothing and the terminal would keep pointing at A.
+    #[test]
+    fn a_retargeted_cell_is_re_emitted_on_the_next_frame() {
+        use ratatui::buffer::CellDiffOption;
+
+        let shared: SharedHyperlinks = SharedHyperlinks::default();
+        let tap = Tap::default();
+        let mut terminal =
+            ratatui::Terminal::new(HyperlinkBackend::new(tap.clone(), shared.clone()))
```
jerome-benoit: [P2] Keep the retarget regression independent of the process terminal. `Terminal::new` asks the wrapped `CrosstermBackend` for the host terminal size; on the exact head, both Linux legs failed here with `WouldBlock` before the assertion ran, and the new rerun has already failed `core-rest` again. Construct the terminal with a fixed viewport via `Terminal::with_options`, or use a fully controlled backend, so this unit test exercises retargeting rather than CI terminal availability.

jerome-benoit (changes requested):
Re-reviewed at `258c6f2de53cba44fe788f8dcca8bb55f58aae7c` after reading the author response and all current threads. The three previous findings are addressed in their narrow form: linked cells are forced through ratatui diffing, ordinary wide graphemes no longer gain phantom spaces, and the production macOS path no longer trusts `BROWSER` or `DISPLAY`.

Approval is still blocked by four new P2 findings below and two already-covered issues that I am not duplicating:

- [CodeRabbit’s platform-test finding](https://github.com/agent-of-empires/agent-of-empires/pull/3749#discussion_r3941415093) is valid: the generic test still asserts Unix-launcher behavior on macOS and Windows. The previous green macOS check was only a skipped placeholder; I added the repository’s `ci-macos` label and the real job is running.
- [CodeRabbit’s retained-link finding](https://github.com/agent-of-empires/agent-of-empires/pull/3749#pullrequestreview-5122189273) is valid: historical and rejected-seed metadata is not reconciled with the accepted frame, so text-only matching can attach a stale target.

The DEC 2026 pre-seed hold issue in that bot review is also technically valid, but the affected branch is byte-identical in this PR’s base, so I am not treating it as a #3749 blocker.

CI is not green. The exact-head rerun has already failed `core-rest` again; the preceding run failed both `core-rest` and `bare-core` at the new retarget test with `WouldBlock`. The real macOS job was still running when this review was submitted.

One non-blocking consistency fix is also needed: `src/cli/serve.rs` and the generated/user documentation still say that `BROWSER` always overrides the reachability check, contradicting the intentional macOS behavior.

jerome-benoit (changes requested):
Re-reviewed `d765f8e0817dc1a0023d40620c6a6958c5a229a7` with two independent sequential analyses and final arbitration. The cell-width, clipped-URL, platform-test and terminal-independent backend-test corrections are addressed. CI, Tests and CodeQL are green at this head.

One existing P2 still blocks approval: [retained hyperlinks must match the accepted frame](https://github.com/agent-of-empires/agent-of-empires/pull/3749#pullrequestreview-5122189273). Rejected seeds no longer publish links, but an accepted reseed still merges rather than reconciles them (`src/tmux/vt.rs:2102–2104`, `1413–1414`). When the accepted snapshot contains the old label as plain text without OSC 8, `record_links` returns on an empty result (`1382–1384`), leaving the old URI available to the text matcher (`src/tui/links.rs:84–103`). The label therefore remains actionable with an obsolete target. This is introduced by this PR, not a pre-existing base issue. Please remove obsolete metadata when committing the accepted frame and keep frame/link updates consistently ordered. No duplicate inline finding added.

Please also exercise the actual generation-refresh path in its regression test: `preview_links.rs:219` assigns the expected target directly to the cache, bypassing both collection and `ensure_parsed`. Update the source table, drive the consumer, and assert unchanged visible cells with the new target; also cover removal after an accepted plain-text snapshot. This belongs to the same fix, not a separate blocker.

Evidence: exact-head source and existing discussions inspected with `gh api`. Static source tracing, not a local reproduction; no local tests were run. Reviewer scope disagreement was resolved by distinguishing an issue already present in this PR from one present in its base.

src/tmux/vt.rs
```diff
@@ -800,6 +806,10 @@ fn swap_seeded_parser(
     *p = vt100::Parser::new(rows, cols, SCROLLBACK_LINES);
     p.process(stream);
     app_cursor.store(p.screen().application_cursor(), Ordering::Relaxed);
+    // Under the parser lock, so the grid and the targets that describe it are
+    // installed together: a sampler cannot catch the new frame beside the old
+    // frame's links, or the reverse.
+    reconcile_links(links, crate::tmux::osc8::extract_links(stream));
```
jerome-benoit, line 812: [P2] Serialize stream link updates with the accepted grid swap

After this swap acquires `parser` and passes the counter checks (794-803), the reader can receive another chunk, increment `chunk_seq` (1540), and call `record_links` (1556) before blocking on `parser` (1557). This replacement then removes the new link because it is absent from the older snapshot. When the swap releases the lock, the reader applies the chunk's bytes but never re-records its URI: the grid contains the new label without its hyperlink. The earlier merge did not erase that arrival. Move the reader's link-table mutation under the parser lock alongside processing its bytes, preserving parser-to-table lock order. Add a barrier-driven regression for an arrival after the swap guard but before reconciliation. The existing rejected-swap test exercises an already-stale guard, not this interleaving.

jerome-benoit (changes requested):
Re-reviewed `d765f8e0..598498df` with two blind sequential analyses and code-level arbitration. The previous plain-text reseed defect is fixed: accepted snapshots replace the table, including with an empty result; rejected swaps leave it unchanged. The consumer regression now exercises `ensure_parsed`/`collect_links` instead of assigning the expected URI.

One P2 remains in the integration: the reader publishes new link metadata outside the parser lock, so the new authoritative replacement can erase a concurrently received link. Details inline. Please serialize both producers with the same lock order and add a deterministic interleaving regression.

Nonblocking: shorten the test helper comment at `src/tmux/vt.rs:1427-1434`; it still claims rejected swaps record links.

Evidence: exact-object `git diff`/`git show`, clean `git diff --check`, existing reviews and replies. This is source-traced, not a runtime reproduction; no local tests were run. At submission preflight, 32 checks succeeded and Nix Build was still running.

PR #3751, opened 2026-09-04
fix(web): stop Android IMEs from re-sending a word they retroactively compose

### LINKED ISSUE #3746: SwiftKey duplicates words in mobile terminal after pressing space
I'm using AoE 1.15.1 on Android with Microsoft SwiftKey. When I type a word in the web terminal and press space, the word gets duplicated.
Example:
type: test
become: testtest
I tested it on Firefox and Chrome - the same bahaviour.
I tried to disable autocorrect, predictions and turn on incognito mode but it didn't help.

AI said this: "Looks like an Android IME/composition handling issue."

njbrake:
## Description

On Android, some keyboards (SwiftKey is the reported one) type a word out one
letter at a time and then, the moment you press space, tell the page a second
time "the word you just finished was `test`". The mobile terminal took that
second message at face value and typed the word again, so every word came out
doubled: you typed `test`, the pane showed `testtest`. The terminal now
remembers the word you are part-way through typing, so when the keyboard hands
the finished word back it only passes along whatever is genuinely new. Typing on
Android produces one copy of each word again, and keyboards that do not behave
this way are untouched.

The reporter's device trace on #3746 is what this is built on:

```
beforeinput "t"/"e"/"s"/"t"  insertText            isComposing:false
compositionstart ""
compositionupdate "test"
compositionend "test"
beforeinput " "              insertText            isComposing:false
```

The word arrives as plain `insertText` edits *before* any composition exists;
SwiftKey then wraps it in a composition after the fact, and `compositionend`
carries the whole word again.

`typedWordRef` holds the word under the caret. It lives in `useLiveTerminal`
beside `sendData`, the one funnel every writer uses, so nothing (including
`MobileTerminalToolbar`, which writes straight to `live.sendData`) can leave a
stale word behind. `sendData` now reports whether the pane will actually receive
the bytes, so a confirmed non-owner's dropped keystrokes are never recorded as
delivered. `compositionend` sends only the part of its data the run has not
already delivered.

Suppression is gated on evidence rather than on a bare prefix match. A
retroactive composition announces itself on its first `compositionupdate` by
carrying the whole typed word, as the trace above shows; a composition that
genuinely starts at the caret builds from its own first character, as the
Samsung trace on the same issue does. A composition with no update to settle it,
and one that stood on its own, are both sent whole.

Two limits worth stating rather than burying:

Autocorrect that *replaces* letters is not fixed. `ted` corrected to `test`
leaves the pane with `tedtest`, because undoing it means erasing what the pane
already showed. That is #3812, and behaviour there is unchanged.

The prefix test cannot be right in both directions. A composition adopting the
typed word and one starting fresh beside it are indistinguishable at the DOM
level once `insertText` is `preventDefault`ed, so a fresh composition sharing
the typed prefix loses that prefix. An equality test instead of a prefix test
trades that for re-doubling a mid-word suggestion tap. Adopt-and-replay is the
only behaviour any device trace shows, so the prefix test is the default here.
The failure direction throughout is a duplicate, never dropped input.

Fixes #3746

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No screenshot: the change is what bytes reach the pane, not pixels. The traced
event sequence in the first test case is the evidence.

## Test Coverage Analysis

Vitest rather than Playwright: the behavior is the component's own event
handling, which RTL drives directly and more cheaply than a browser.
`web/src/components/__tests__/MobileLiveTerminal.composition.test.tsx` replays
the traced SwiftKey sequence plus the surrounding cases (multi-word sentences, a
composition that extends the run, an unrelated composition, plain repeated
characters, Enter, backspace). Five of the eight fail on unfixed HEAD.
`terminal.mobile-ime-word-commit` added to the matrix.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Diagnosis came from the event trace the reporter ran on their device. An earlier
revision of this branch guarded the opposite ordering and the trace ruled it
out; that approach is gone.

https://claude.ai/code/session_01MdNyjG6kyGSfoVxxSPjzLq

## Summary

- Fixed duplicate words from Android keyboards such as SwiftKey.
- Shared typed-word tracking between terminal input and the mobile terminal.
- Suppressed only composition text that repeats accepted plain-text input.
- Preserved standalone composition behavior.
- Improved state handling for backspace, Enter, dropped input, and toolbar actions.
- Added regression tests for SwiftKey-like input, long words, Unicode deletion, and related cases.
- Added the test coverage entry.

## Benefits

Mobile users can type with Android keyboards without duplicated words. Accepted input is tracked more reliably, and broader test coverage reduces regression risk.

### DIFF
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -1132,8 +1132,8 @@ export function MobileLiveTerminal({
   // composition after the fact, so `compositionend` carries the whole word a
   // second time and "test" reached the pane as "testtest"; whatever the run
   // already delivered is dropped from that word. Sending clears it below, so it
-  // stays a suffix of what the pane received and the strip can never swallow
-  // bytes the pane never saw; the two paths that continue a word re-arm it.
+  // only ever holds bytes the pane received and the strip can never swallow
+  // bytes it never saw; the paths that continue a word re-arm it.
   const plainRunRef = useRef("");
   // Typed input interrupts the coast. Without this, keystrokes sent in the
   // coast's 1-2s tail interleave with the wheel storm and the app is busy
@@ -1712,6 +1712,10 @@ export function MobileLiveTerminal({
       // edits; only the part the pane has not seen yet is new.
       const rest = run && data.startsWith(run) ? data.slice(run.length) : data;
       if (rest) sendKeys(rest);
+      // The composed word is still the one under the caret, so a second
+      // composition over it (a suggestion tap, then the space commit) has to
+      // be stripped against everything the pane has of that word.
+      plainRunRef.current = plainRunAfter(run, rest);
       if (e.currentTarget instanceof HTMLTextAreaElement) e.currentTarget.value = "";
     },
     [sendKeys],
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -126,7 +126,8 @@ export interface MobileLiveTerminalProps {
   setCadence: (fast: boolean) => void;
   enterReading: (rows: number) => void;
   returnToLive: (rows: number) => void;
-  sendData: (data: string) => void;
+  /** Returns whether the pane will receive the data; see useLiveTerminal. */
+  sendData: (data: string) => boolean;
   /** Upload a clipboard image pasted into the pane and resolve to the path
    *  the tmux pane can read (host path, or the container mount for sandboxed
    *  sessions), or null on failure. See #2678. */
@@ -323,12 +324,19 @@ function navigationKeySequence(e: TerminalKeyLike): string | null {
   }
 }
 
-const PLAIN_RUN_MAX = 64;
-
-/** Longest run of non-whitespace at the end of `run + typed`, capped: only the
- *  word still under the caret can be swallowed by a retroactive composition. */
+/** The word under the caret: the run of non-whitespace ending `run + typed`.
+ *  Uncapped, because a shorter run than the IME's own word fails the prefix
+ *  test and lets the whole word through twice. */
 function plainRunAfter(run: string, typed: string): string {
-  return (/\S*$/.exec(run + typed)?.[0] ?? "").slice(-PLAIN_RUN_MAX);
+  return /\S*$/.exec(run + typed)?.[0] ?? "";
+}
+
+/** Drop the last code point, so backspacing a surrogate pair or an emoji does
+ *  not leave half of it in the run and break the prefix test. */
+function dropLastCodePoint(run: string): string {
+  const points = Array.from(run);
+  points.pop();
+  return points.join("");
 }
 
 function specialKeySequence(e: TerminalKeyLike): string | null {
@@ -1146,7 +1154,7 @@ export function MobileLiveTerminal({
       stopMomentum();
       cancelTouchWheelQueue();
       plainRunRef.current = "";
-      sendDataRaw(data);
+      return sendDataRaw(data);
     },
     [sendDataRaw, stopMomentum, cancelTouchWheelQueue],
   );
@@ -1528,6 +1536,9 @@ export function MobileLiveTerminal({
 
   // --- keyboard input -----------------------------------------------------------
   const composingRef = useRef(false);
+  // Returns whether `data` itself reached the pane, so a caller can tell
+  // whether the run may record it. A Ctrl chord sends a control code instead,
+  // and a non-owner's keystrokes are dropped outright.
   const sendKeys = useCallback(
     (data: string) => {
       if (ctrlActiveRef.current && data.length === 1) {
@@ -1535,10 +1546,10 @@ export function MobileLiveTerminal({
         if (code >= 65 && code <= 90) {
           sendData(String.fromCharCode(code - 64));
           clearCtrl();
-          return;
+          return false;
         }
       }
-      sendData(data);
+      return sendData(data);
     },
     [sendData, ctrlActiveRef, clearCtrl],
   );
@@ -1554,7 +1565,7 @@ export function MobileLiveTerminal({
       switch (input.inputType) {
         case "insertText": {
           const data = input.data ?? "";
-          if (data) sendKeys(data);
+          if (data && !sendKeys(data)) break;
           plainRunRef.current = plainRunAfter(run, data);
           break;
         }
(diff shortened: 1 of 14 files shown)


web/src/components/MobileLiveTerminal.tsx
```diff
@@ -1680,7 +1705,13 @@ export function MobileLiveTerminal({
   const handleCompositionEnd = useCallback(
     (e: CompositionEvent) => {
       composingRef.current = false;
-      if (e.data) sendKeys(e.data);
+      const run = plainRunRef.current;
+      plainRunRef.current = "";
+      const data = e.data ?? "";
+      // The composition may have taken over a word already spelled out as plain
+      // edits; only the part the pane has not seen yet is new.
+      const rest = run && data.startsWith(run) ? data.slice(run.length) : data;
```
jerome-benoit: [P2] Do not treat every prefix-sharing composition as retroactive. `compositionend.data` is the final result of the current composition session; it is not guaranteed to replace earlier plain `insertText`. A valid `insertText("a")` followed by a distinct composition yielding `"android"` intends `aandroid`, but this strips the first `a` and produces `android`. The device trace only establishes an exact replay of `test`; the new `tes` -> `test` test broadens the heuristic without evidence. Please gate deduplication on the observed SwiftKey signature and add a negative regression for an ordinary composition following plain text. See https://www.w3.org/TR/uievents/#event-type-compositionend.

web/src/components/MobileLiveTerminal.tsx
```diff
@@ -1532,16 +1549,24 @@ export function MobileLiveTerminal({
...
       if (composingRef.current || input.isComposing) return;
+      const run = plainRunRef.current;
+      plainRunRef.current = "";
       switch (input.inputType) {
-        case "insertText":
-          if (input.data) sendKeys(input.data);
+        case "insertText": {
+          const data = input.data ?? "";
+          if (data) sendKeys(data);
+          plainRunRef.current = plainRunAfter(run, data);
```
jerome-benoit: [P2] Only retain plain input that was accepted for delivery. `useLiveTerminal.sendData` silently drops keystrokes from a confirmed non-owner, but this line still records them as text the pane received. After ownership changes without a remount, a matching composition can then be suppressed entirely, or the stale run can make the next word duplicate. Please propagate an accepted/sent-or-queued result and re-arm this state only on acceptance; also clear it at ownership boundaries. Add a non-owner -> claim regression.

web/src/components/MobileLiveTerminal.tsx
```diff
@@ -1532,16 +1549,24 @@ export function MobileLiveTerminal({
...
+        }
         case "insertLineBreak":
         case "insertParagraph":
           sendKeys("\r");
           break;
         case "deleteContentBackward":
+          // One character, so the IME's word loses its last one too;
+          // `deleteWordBackward` is a separate input type and not forwarded.
           sendKeys("\x7f");
+          plainRunRef.current = run.slice(0, -1);
```
jerome-benoit: [P2] Do not remove one UTF-16 code unit for a backward deletion. For a non-BMP character such as `🙂`, this leaves an isolated surrogate in the tracked run; the next retroactive composition no longer matches and its text is sent twice. `deleteContentBackward` describes deleting the content before the caret, not one JavaScript code unit. Update the run by the same Unicode unit the terminal deletion represents, or invalidate the heuristic when that cannot be reconstructed, and cover a non-BMP/graphheme case.

jerome-benoit (changes requested):
Requesting changes because the new heuristic can still duplicate long input and can also suppress or duplicate valid input outside the recorded SwiftKey sequence. The existing CodeRabbit finding about the 64-code-unit cap is valid and is not duplicated inline. The three additional issues are annotated below. Please constrain deduplication to evidence that the composition is retroactive, retain only text actually accepted for delivery, and update the tracked run with Unicode-safe deletion semantics.

web/src/components/MobileLiveTerminal.tsx
```diff
@@ -1129,7 +1153,8 @@ export function MobileLiveTerminal({
     (data: string) => {
       stopMomentum();
       cancelTouchWheelQueue();
-      sendDataRaw(data);
+      plainRunRef.current = "";
```
jerome-benoit: [P2] Reset IME tracking for toolbar writes. This wrapper is the only place that clears `plainRunRef`, but `LiveTerminalView` passes `live.sendData` directly to `MobileTerminalToolbar`. After typing `test`, tapping the toolbar `^C`, and typing `test` again, the tracked run is `testtest`; a following SwiftKey composition of `test` no longer matches and the word is sent twice. Route toolbar writes through the same invalidation path (or expose a shared invalidation callback), and add an integrated toolbar regression.

jerome-benoit (changes requested):
Re-reviewed at `3b0f524a849f8d678c99e8043ff85954b226cb2f` after reading the author response and all current threads. The 64-code-unit cap, Ctrl conversion, accepted/dropped propagation, and non-BMP deletion issues are fixed, and CI is green.

Two P2 issues still block approval:

- The existing prefix-sharing thread remains valid. The concern is not that SwiftKey was proven to emit the specific `a` / `android` sequence; it is that the implementation treats any later composition whose result shares the current plain-text prefix as retroactive replacement, although `compositionend.data` only describes that composition session. Please scope suppression to the observed retroactive signature and add a negative regression for a distinct composition following plain input.
- The new inline finding below shows that toolbar writes bypass the only invalidation point for this new IME state, so a normal toolbar interrupt can make the original duplication recur.

I also verified the current CodeRabbit ownership finding. The queued-input replay mechanism predates this PR; the new accepted/dropped return does not create it. I am therefore not reclassifying that pre-existing behavior as a new P1 blocker here. The accepted-input part of the earlier human thread is fixed; its ownership-boundary concern should follow the underlying queue lifecycle.

jerome-benoit (changes requested):
Re-reviewed `6429945b` with two independent sequential analyses and cross-validation. The toolbar P2 is fixed: both writers now use the hook's `sendData`, which clears the shared `typedWordRef`.

One P2 still blocks approval: the existing [prefix-sharing finding](https://github.com/agent-of-empires/agent-of-empires/pull/3751#discussion_r3940605349), also covered by the [new CodeRabbit thread](https://github.com/agent-of-empires/agent-of-empires/pull/3751#discussion_r3952555022). `MobileLiveTerminal.tsx:1723` still accepts any first update starting with the previous run. Plain `a`, followed by a fresh first update and completion of `android`, therefore sends `ndroid` at line 1737, producing `android` instead of `aandroid`. The new negative test at lines 200-205 starts with `n`, so it misses this case. A first update need not be a single character: the [UI Events composition example](https://www.w3.org/TR/uievents/#events-composition-event-order) starts with a whole word. At minimum, constrain detection to the observed exact initial replay and cover this fresh whole-word case; prefix similarity alone is insufficient. No duplicate inline comment added.

The pre-existing ownership/queue issue remains separate, not an additional blocker introduced by this correction. Evidence: exact-head `git show`, `git diff 3b0f524a 6429945b`, and `git diff --check` (exit 0). CI is green. This is a source-level counterexample, not a device reproduction; no local tests were run.

Issue #2414, opened 2026-06-25
BUG: attaching to docker container of paired aoe session is broken


njbrake:
I just see

```
runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/workspace/<name_of_dir>") set in config.json failed: no such file or directory

What's next:
    Try Docker Debug for seamless, persistent debugging tools in any container or image → docker debug aoe-sandbox-3033d139
    Learn more at https://docs.docker.com/go/debug-cli/


```

PR #3754, opened 2026-09-04
fix(sandbox): reclaim named ignore volumes stranded by a worktree move

### LINKED ISSUE #3742: Named ignore volumes orphaned by worktree move
aoe 1.15.3, Docker Desktop 29.6.1, macOS. `volume_ignores_strategy = "named"`.

`ContainerSpec::discard` preserves `aoe-vi-{session_id}-*` volumes on the premise that "the recreated container re-attaches them on next start" (`src/containers/mod.rs:255`). But the volume name is `aoe-vi-{session_id}-{slug(container_path)}-{hash12(container_path)}` (`src/session/config/container_config.rs:1707`). A worktree move changes `container_path`, so the recreated container computes a different name and mounts a fresh volume. The pre-move volume is orphaned at that moment. The two discard call sites on the move path are `worktree_edit.rs:179` and `:240`.

`deletion.rs:597` sweeps by `aoe-vi-{session_id}-` prefix and is correct; it just never runs for a session that is moved and later purged from the trash.

Seven sessions in one profile, each holding both a pre-move and a post-move target volume:

```
403c47ed740c490e  workspace-otari-target-8ec07926d6b0
                  workspace-otari-worktrees-905-target-31ddd0322290
f6e1256028f64cff  workspace-otari-target-8ec07926d6b0
                  workspace-otari-worktrees-rev-912-target-873cf2685e47
(5 more, same shape)
```

Three orphans from sessions no longer in `aoe list` held 64.8 GB (35.92 + 25.71 + 3.18). Rust target dirs run 20 to 35 GB per session here, so a move leaks that much per occurrence.

`docker volume prune` skips named volumes, so these are invisible to the usual cleanup and report 0B reclaimed while the orphans sit there. `-a` is required.

Suggested fix: sweep the old prefix once the move completes, or key the volume name on session id plus mount role instead of the path so a move keeps the same name.

Secondary: a reconcile pass removing `aoe-vi-*` volumes whose session id is absent from the session store would catch orphans from any path, including a delete that ran while the runtime was unreachable. (`remove_named_ignore_volumes` swallows both the list and rm failures and returns `Ok`, so a transient runtime outage during deletion orphans silently.)

njbrake:
## Description

When you use a sandbox session on macOS with the "named" volume strategy, AoE keeps your build caches (`target/`, `node_modules/`, `.venv/`) inside Docker volumes instead of on disk, and it names each volume after the folder it belongs to. Moving a session's worktree changes that folder, so the next time the container starts it makes brand new empty volumes under the new names. The old ones were left behind with nothing to clean them up until the session was deleted, and Docker's usual `docker volume prune` skips them, so they were invisible while sitting on tens of gigabytes each. This change makes AoE throw away the leftovers at the moment it rebuilds the container, keeping only the volumes the new container actually uses, and only when it can prove which ones those are.

A named ignore volume is called `aoe-vi-{session_id}-{slug}-{hash}` over its container mount path (`container_config.rs`). The move path discards the container while preserving the session's volumes on the premise that they are re-attached on next start; that holds only for mounts whose container path did not move. In the reporter's case the main-repo `/workspace/otari/target` survives and the worktree's `/workspace/otari-worktrees/905/target` is orphaned on the spot. `deletion.rs` sweeps the whole `aoe-vi-{session_id}-` prefix correctly, but never runs for a session that is moved and later purged from the trash.

The fix names the volumes the move stranded and removes those, at the create that follows the move, when no container holds them.

### Naming the strand rather than inferring it

For each ignore path under the new `working_dir`, remap it onto the workdir the previous container was created with and compute the volume name it had there. That is exactly the set the move orphaned. A mount that kept its container path is never named, even when this run's config fails to resolve it, which is what a glob whose directory is absent from the host looks like. The runtime primitive takes an allowlist, so a volume the caller did not name cannot be reached from the reclaim at all.

The evidence of a move is `SandboxInfo::container_workdir`, already pinned at create for #2414, so there is no new persisted state and no migration. Nothing is removed unless it changed, and one further gate applies: `ContainerConfig::named_ignore_volumes_authoritative` is false when the mount resolve was degraded. `compute_volume_paths` falls through to `/workspace/{basename}` when `find_main_repo` fails, which the repo has a regression test for (`container_workdir_stays_pinned_when_worktree_linkage_breaks`, the #2414 fix), and there the workdir is provisional too, so the apparent move may be nothing but that failure. `compute_volume_paths_with_resolve` reports which branch it took rather than letting the gate re-derive it later, so a linkage repaired between the two reads cannot make a collapsed resolve look authoritative.

**What this deliberately does not reclaim.** A volume orphaned without a move: an edited `volume_ignores`, a switch back to `volume_ignores_strategy = "anonymous"`, an `extra_volumes` entry that shadows an ignore path, or a multi-repo attach, which clears the workdir pin the reclaim reads. Nor anything already orphaned: the pin advances at every create, so a session that has moved and restarted is out of reach, and the volumes in the issue belong to sessions that no longer exist. Upgrading does not shrink them.

**What it costs.** A move now loses the cache in both directions. Before this, moving A to B stranded A's volumes but left them intact, so moving back re-attached a warm cache; now they are gone at the B create and the return trip rebuilds cold. That is the price of reclaiming at all, and it is why the reclaim refuses to fire without proof of a move. Those keep today's behavior, so no user is worse off than before this PR. `docs/guides/sandbox.md` says so and gives the `docker volume ls` incantation to find them, and the multi-repo guide no longer claims an attach cleans up after itself. Reclaiming those safely needs each volume's container path recorded on `SandboxInfo` plus a migration; that is a separate change.

**Reclaim happens at the session's next start, not at the move.** Sweeping at move time needs the post-move volume name set, which means building a full container config at all nine `discard_sandbox_container_after_move` call sites.

### Why not the issue's first suggestion

#3742 offered two fixes, and this PR takes the second. The first, keying the volume name on session id plus mount role so a move keeps the same name, is the better end state: it removes the strand rather than collecting it, and it keeps the cache warm across a move instead of paying the cost above. It is not in this PR because there is no stable "role" to key on. The mount bases are themselves path-derived, so two ignore entries under different bases (`target/` in the main repo and in the worktree) collide on any role name short enough to be stable, and Docker cannot rename a volume, so every existing volume would need a migration to reach the new scheme.

That migration is worth doing, and it is also what a stable hash needs: `named_volume_for` uses `DefaultHasher`, which std documents as unspecified across releases. Two tests pin literal hashes as a canary for exactly that. Both belong in one follow-up.

The issue's secondary suggestion (a store-wide reconcile for `aoe-vi-*` volumes whose session id is absent from the session store, plus the swallowed failures in `remove_named_ignore_volumes`) is likewise a separate concern and is not in this PR.

Fixes #3742

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: reaching the reclaim needs a live Docker daemon, a named-strategy sandbox session, a worktree move and a restart. `tests/e2e/sandbox.rs` does have the `#[ignore = "requires Docker daemon"]` convention for that, but no CI job runs those and no Docker was available to run one even once while writing this, so it would land unexecuted. Unit tests cover the decisions instead: `stranded_named_ignore_volumes` producing the reporter's own orphaned volume name for their layout, and producing it unchanged when the main repo's volume is missing from the config (the case the previous approach deleted); every branch that names nothing (unmoved workdir, no pin, degraded resolve); `selected_named_ignore_volumes` honoring the allowlist and never reaching a longer session id under the same prefix; `compute_volume_paths_with_resolve` reporting `Fallthrough` for every arrival at the collapsed `/workspace/{basename}`; and the flag end to end through `build_container_config`. What is left uncovered is the call at the create site.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Implementation and tests were written by Claude Opus 5 in back-and-forth with @njbrake; the scope calls above are his. The first draft removed anything the new config did not mount. Three independent review passes found successively narrower ways that deletes a live cache; commits 2 through 4 are those fixes.

https://claude.ai/code/session_01RCYiZhmr6riURz45y3eE9i

## Summary

- Reclaims stranded named Docker ignore volumes after a recognized sandbox worktree move.
- Preserves active volumes and skips uncertain or unsupported cases.
- Adds safeguards, logging, unit tests, and documentation updates.

## Benefits

This reduces disk usage without removing valid volumes. Previously orphaned volumes and Docker end-to-end coverage remain future improvements.

### DIFF
--- a/docs/guides/multi-repo-workspaces.md
+++ b/docs/guides/multi-repo-workspaces.md
@@ -124,9 +124,11 @@ ancestor of the workspace and every repo. Build caches (`target/`,
 `node_modules/`) go with the container under the default
 [volume ignores strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs). Under
 `"named"` they are keyed on their container path, so they survive an attach that
-leaves that path alone; attaching a repo from outside the current common ancestor
-moves every mount, so those caches start empty and the volumes they leave behind
-are reclaimed at the next start.
+leaves that path alone. Attaching a repo from outside the current common ancestor
+moves every mount, so those caches start empty, and the volumes they leave behind
+are not reclaimed: the attach clears the pin AoE compares against to recognize a
+move. Remove them by hand with the `docker volume ls` command under
+[Volume Ignores Strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs).
 
 ## The Project Registry
 
--- a/docs/guides/multi-repo-workspaces.md
+++ b/docs/guides/multi-repo-workspaces.md
@@ -121,11 +121,12 @@ throwaway directory that deletion removes.
 Sandboxed sessions have their container removed and recreated, since bind mounts
 are fixed when the container is created and the container mounts the common
 ancestor of the workspace and every repo. Build caches (`target/`,
-`node_modules/`) are named volumes keyed on their container path, so they
-survive an attach that leaves that path alone. Attaching a repo from outside the
-current common ancestor moves every mount, so those caches start empty; see
-[Volume Ignores Strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs) for
-what happens to the volumes they leave behind.
+`node_modules/`) go with the container under the default
+[volume ignores strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs). Under
+`"named"` they are keyed on their container path, so they survive an attach that
+leaves that path alone; attaching a repo from outside the current common ancestor
+moves every mount, so those caches start empty and the volumes they leave behind
+are reclaimed at the next start.
 
 ## The Project Registry
 
--- a/docs/guides/multi-repo-workspaces.md
+++ b/docs/guides/multi-repo-workspaces.md
@@ -121,7 +121,11 @@ throwaway directory that deletion removes.
 Sandboxed sessions have their container removed and recreated, since bind mounts
 are fixed when the container is created and the container mounts the common
 ancestor of the workspace and every repo. Build caches (`target/`,
-`node_modules/`) are named volumes and survive.
+`node_modules/`) are named volumes keyed on their container path, so they
+survive an attach that leaves that path alone. Attaching a repo from outside the
+current common ancestor moves every mount, so those caches start empty; see
+[Volume Ignores Strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs) for
+what happens to the volumes they leave behind.
 
 ## The Project Registry
 
--- a/docs/guides/multi-repo-workspaces.md
+++ b/docs/guides/multi-repo-workspaces.md
@@ -125,10 +125,8 @@ ancestor of the workspace and every repo. Build caches (`target/`,
 [volume ignores strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs). Under
 `"named"` they are keyed on their container path, so they survive an attach that
 leaves that path alone. Attaching a repo from outside the current common ancestor
-moves every mount, so those caches start empty, and the volumes they leave behind
-are not reclaimed: the attach clears the pin AoE compares against to recognize a
-move. Remove them by hand with the `docker volume ls` command under
-[Volume Ignores Strategy](sandbox.md#volume-ignores-strategy-macosvirtiofs).
+moves every mount, so those caches start empty and the volumes they leave behind
+have to be removed by hand.
 
 ## The Project Registry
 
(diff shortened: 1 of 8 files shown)


jerome-benoit (approved):
Reviewed at `ccb8a93b6d4a8c47f169b8ff100485c292705206`.

The authoritative mount-resolution guard, previous/current worktree remapping, explicit deletion allowlist, and preservation of still-live volume names keep the cleanup narrowly scoped. This addresses the duplicate named volumes reported in #3742 without turning the change into a store-wide reconciliation mechanism.

The remaining purge-without-restart and general orphan-reconciliation cases are pre-existing and explicitly outside this PR’s scope. All checks are green.

Issue #3545, opened 2026-08-26
Custom agent wrapping an agent over ssh never returns to Idle on turn completion


jakecelentano:
## Summary

A custom agent that wraps a supported agent behind `ssh` starts fine and is detected as running, but **never returns to `Idle` when the turn completes**. It sits in `Running` indefinitely while the pane visibly shows an idle prompt.

`Starting -> Idle` and `Idle -> Running` both fire correctly. Only turn-completion detection fails, and it fails only when the wrapper is an `ssh` invocation — the identical custom-agent + `agent_detect_as` config with a local command works perfectly (control below).

This matters because `Running` is the attention signal: a stuck session never sorts up in the Attention sort, never clears the activity indicator, and never fires an idle/waiting notification. Running an agent on another machine is the use case #473 was closed in favour of custom agents for, so the wrapper path is the sanctioned one here.

## Environment

- aoe 1.15.0, Linux (Debian), tmux substrate, terminal view (not SV)
- Wrapped agent: `pi` 0.84.x on both hosts
- Remote host reached over plain OpenSSH, key auth, no jump host

## Repro

```toml
# ~/.config/agent-of-empires/config.toml
[session.custom_agents]
pi-umbra = "ssh -t root@10.89.97.11 'pi -ne'"

[session.agent_detect_as]
pi-umbra = "pi"
```

```bash
aoe add /root --title RemoteSpike2 --tool pi-umbra
aoe session start RemoteSpike2
aoe ps            # -> idle          (correct)

aoe send RemoteSpike2 "reply with exactly the word: ok"
aoe ps            # -> running       (correct)

# agent answers within ~5s; pane shows the answer and an idle input box
aoe ps            # -> running, still, 5+ minutes later   (WRONG)
```

Observed status transitions — note the third one never arrives:

```
14:55:08 INFO session.status_change: bc3e888144804710 [pi] Starting -> Idle    (hook=None hook_age_ms=None)
14:55:22 INFO session.status_change: bc3e888144804710 [pi] Idle    -> Running  (hook=None hook_age_ms=None)
14:55:22 INFO agent_of_empires::tmux::vt: vt channel armed (pipe-pane -IO <-> vt100 grid) target=aoe_RemoteSpike2_bc3e8881:^.0 cols=194 rows=45
(nothing further; polled every 5s for 4 minutes)
```

[...]

Issue #3203, opened 2026-08-01
Structured view: the stranded prompt loop behind #3190 is only half treated, it still rejects prompts as agent_busy


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

#3190 was a structured session pinned at `Running` for an hour after the agent finished, because an agent-initiated turn got no terminal `Stopped`. PR #3192 bounds that symptom to about 90 seconds and closes one reachable cause (a connection-scoped terminal guard, now per-turn), but the leading hypothesis for the original occurrences is still unproven and only half-treated.

The hypothesis: the daemon's prompt loop is stranded. A `PromptCompleted` frame arrives from the runner with no completion waiter installed, the prompt future it was going to resolve never resolves, and the loop stays parked inside its prompt arm. Everything gated on `!prompt_active` then stays disabled, which is the entire between-prompt watchdog lane.

## What #3192 already did

- The reader's no-waiter branch clears the now-stale `prompt_in_flight`, so the between-prompt lane re-arms and the status path heals in-machine (`src/acp/acp_client.rs`, the `ControlBody::PromptCompleted` arm).
- The reconciler repairs a missing terminal from the event log within 60s, so the badge is never wrong for an hour again.
- That branch, and the waiter-install and prompt-drain boundaries, now log, so the next occurrence is diagnosable instead of inferred.

## What is left

**1. The prompt loop is still parked.** Clearing the flag re-arms the lane but cannot resolve a future the reader does not hold. The loop stays in its prompt arm, whose mid-turn `Prompt` handler rejects new work as `agent_busy` (`src/acp/acp_client.rs`, the in-flight `ClientCmd::Prompt` arm). So a stranded session heals its status and still refuses prompts until something else ends the turn. Options, none obviously right yet:

- Have the reader signal the prompt loop to abandon the parked future (a shared oneshot or a typed command), so the arm exits through its normal terminal path.
- Give the prompt arm a bound: if the runner reported completion for this turn while our own future is unresolved, stop waiting.
- Hold the completion sender somewhere the reader can always find, so "no waiter" becomes impossible.

[...]

PR #3758, opened 2026-09-04
fix(acp): give each runner generation one lifecycle owner

### LINKED ISSUE #3487: ACP runner lifecycle needs unified ownership across shutdown, respawn, reaping, and recovery
## Context

Review of #3476 exposed a broader lifecycle problem beyond the handshake-timeout defect in #3474. ACP runner ownership is currently coordinated across several independent mechanisms: the in-memory worker map, pending and cancelled resume sets, the on-disk worker record, session-level restart markers, drain tasks, the reaper, and reconciler-local retry state.

No single authority token currently spans a runner generation from admission through spawn, respawn, shutdown, reaping, and recovery.

## Confirmed race boundaries

The current control flow permits these reviewed interleavings:

- Shutdown can release the worker-map lock while exact teardown is pending; the drain or reaper may remove the old handle and a concurrent spawn may install a replacement before shutdown settles.
- A drain respawn can finish after shutdown begins. If its result is rejected without exact teardown, the new socket runner and registry record remain live and reattachable.
- A reaper can snapshot a missing registry entry, then remove by session ID after a newer handle has been installed.
- Detached-runner teardown failure can release a stop fence even though process-group exit was not proven, allowing resume while descendants may remain alive.
- Session-scoped restart markers can outlive the generation that created them, including across daemon restart, and later authorize an unrelated respawn.
- Reconciliation performs preparation before fresh-spawn reservation, so shutdown can complete before a stale reconciliation attempt reaches spawn admission.

These are variants of one ownership problem, not independent timeout bugs.

## Proposed direction

Introduce one per-session lifecycle controller with an exact runner-generation lease. A minimal state machine should cover at least:

- absent
- starting or attaching
- running
- respawning
- stopping
- retryable teardown

Every terminal effect should consume the same lease or reject drift:

- runner spawn and installation
- drain-driven respawn
- shutdown and delete
- process-group signaling
- registry deletion
- reaper removal
- reconciler adoption or fresh spawn
- restart authorization

Restart intent should be bound to the exact generation or a monotonic lifecycle epoch, rather than represented by an unversioned session-level marker.

## Acceptance criteria

- Concurrent shutdown, spawn, attach, drain respawn, reaper, and reconciler work cannot install or remove a runner without the current lifecycle lease.
- A late or cancelled spawn is exactly torn down before its ownership is released.
- Teardown failure retains retryable ownership until exact process-group exit and registry settlement are proven.
- Reaper removal revalidates the exact in-memory client and persisted generation.
- Restart authorization cannot be consumed by a later unrelated generation or survive as stale authority after explicit stop.
- Crash and daemon-restart recovery has one deterministic transition for every persisted state.
- Deterministic tests exercise the production entrypoints for shutdown-versus-respawn, reaper-versus-replacement, stop-during-reconciliation, late spawn cancellation, stale restart intent, and teardown retry.

## Non-goals

- Do not fold this redesign into the narrow handshake-timeout fix for #3474.
- No public UI lifecycle state is required unless a later product requirement needs it.
- No registry compatibility layer or new dependency is implied.

### LINKED ISSUE #3514: Session doesn't get resumed after rate limit
Very often, I'd say about at least 80% of the time, sessions don't get resumed after they were rate-limited.

Even though after a couple of fixes now the parsed time seems to be correct.

Just this night I had a session be rate-limited and should've resumed at 3:30, correct time, but I checked after I woke up and it still just says that it should resume at 3:30.

### LINKED ISSUE #3553: aoe-agent is the default ACP agent but nothing appears to build or install it, and `aoe acp agents` reports it present regardless

### LINKED ISSUE #3560: Structured-session ACP: unknown WS control frame causes a reconnect storm; a rejected resumed session terminates the runner

### LINKED ISSUE #3686: Handle dormant and orphaned plugin sessions before sending turns
## Problem

`sessions.turn.send` treats every `WorkerDown` disposition as sendable, but two distinct persisted states require opposite handling.

At merged head `576dd7a950e7f5636784d86ba848816cdff68dd2`:

1. **Idle-dormant session:** the plugin path passes `false` to both [`begin_prompt_submission` and `send_turn`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/plugin/session_api.rs#L566-L600). The resume snapshot rejects [`is_idle_dormant()`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/acp_reconciler.rs#L1762-L1779), so an idle-reaped plugin session returns `session_not_found` instead of waking. This contradicts the new plugin API contract that dormant sessions are resumed.
2. **Orphaned active turn:** [`decide`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/acp/dispatch.rs#L71-L79) returns `WorkerDown` before checking `turn_active`. The plugin then resumes through a target that hardcodes [`in_flight_turn: false`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/acp_reconciler.rs#L1757-L1789), bypassing the normal restart path's synthetic [`Stopped { orphaned_at_restart }`](https://github.com/agent-of-empires/agent-of-empires/blob/576dd7a950e7f5636784d86ba848816cdff68dd2/src/server/acp_reconciler.rs#L1434-L1443). A new prompt can be published while the prior turn remains active in the event fold.

Both runtime gaps predate #3673; the PR's new `WorkerDown` exception and public documentation preserve or expose them rather than completing the behavior.

## Impact

- Scheduled plugin sessions stop working after idle auto-reaping.
- A dead worker with an orphaned active turn can receive a second prompt without a terminal event for the first, leaving transcript and control state inconsistent.

**Priority: P2**

## Expected fix

Make plugin cold-worker admission distinguish persisted session state:

- clear and persist idle dormancy through the same serialized wake path used by direct prompts, then resume and wait;
- if the control fold is still active, close the orphan through the normal restart semantics before publishing a new prompt, or return a retryable error until it is closed;
- decide explicitly whether manually stopped sessions should auto-resume, and align the documentation with that policy.

Add deterministic tests for idle-dormant, manually stopped, dead-idle, and dead-with-active-turn plugin sessions.

Found during a cross-validated post-merge audit of #3673. This issue was drafted with Codex under @jerome-benoit's direction and reviewed before submission.

### LINKED ISSUE #3707: Structured view: handle history-window part-list collapse without /clear
## Problem

PR #3676 rebuilds the assistant-ui runtime when the latest `/clear` divider changes. A similar pre-existing transition remains possible without `/clear`.

When a single assistant turn exceeds `DEFAULT_HISTORY_WINDOW` (150 rows), `historyWindowStart` can cut inside that turn. Once the next prompt arrives, the start snaps forward to that user boundary and the rendered assistant part list can shrink sharply, while `clearFoldGeneration` and therefore the runtime key remain unchanged. This recreates the stale absolute-part-index risk addressed by #3676.

This path was identified during the review of #3676 but was outside that PR's scope. It has not yet been reproduced independently.

## Expected behavior

A history-window boundary change must not leave assistant-ui's runtime referencing the previous rendered part list. Avoid unnecessary runtime rebuilds when the effective boundary does not change.

## Verification

Add a regression scenario with no `/clear` that:

1. creates one assistant turn exceeding 150 rows;
2. appends the next prompt so the window start moves forward;
3. verifies that no `useClientLookup` out-of-bounds error or ErrorBoundary occurs, the new turn renders, and the composer remains usable.

If the production build does not reproduce the failure, document why this transition is safe rather than adding a speculative fix.

Related: #3640, #3676, assistant-ui/assistant-ui#5708.

### LINKED ISSUE #3734: Use the session's effective Claude config directory for MCP discovery
## Problem

#3699 makes native MCP discovery honor `session.agent_config_dir`, but it still
ignores `CLAUDE_CONFIG_DIR` from the environment actually injected into a host
session.

At `10b65978ecfecd7f32509b2f4304850e0bdac42e`,
[`native_config_dir_for()`](https://github.com/agent-of-empires/agent-of-empires/blob/10b65978ecfecd7f32509b2f4304850e0bdac42e/src/session/mcp/mcp_model.rs#L524-L540)
checks:

1. `session.agent_config_dir`;
2. `CLAUDE_CONFIG_DIR` from the AoE daemon process;
3. the default home directory.

It does not inspect the profile environment or values produced by
`host_hooks.before_session`.

The ACP spawn path separately resolves the profile environment, applies
`before_session` values with last-write precedence, and injects the result into
the agent process. MCP discovery therefore can read a different `.claude.json`
from the one Claude reads.

## Reproduction

Start AoE without `CLAUDE_CONFIG_DIR` and configure a host profile:

```toml
environment = ["CLAUDE_CONFIG_DIR=/tmp/claude-work"]
```

Create different native MCP definitions:

```text
~/.claude.json
  server: home-account

/tmp/claude-work/.claude.json
  server: work-account
```

Then start a structured Claude session with that profile.

Verified code path:

1. `resolve_host_environment_pairs()` produces
   `CLAUDE_CONFIG_DIR=/tmp/claude-work`.
2. The ACP spawn applies that value to the Claude process.
3. `resolve_mcp_layers()` does not receive the effective environment.
4. `native_config_dir_for()` sees neither `session.agent_config_dir` nor a
   daemon-level variable and reads `~/.claude.json`.
5. AoE forwards `home-account` while Claude itself uses
   `/tmp/claude-work/.claude.json`.

The same mismatch occurs when `host_hooks.before_session` mints or replaces
`CLAUDE_CONFIG_DIR`.

Native MCP definitions can contain stdio environment values and HTTP/SSE
headers, including authorization credentials. Reading the wrong account’s file
can therefore forward the wrong MCP servers and credentials to a session.

## Expected behavior

Native MCP discovery, drift reconciliation, management surfaces, and the
launched agent must resolve the same effective Claude config directory.

## Acceptance criteria

- Resolve native MCP configuration from the session’s final environment after
  profile environment and `before_session` values have been applied.
- Preserve explicit `session.agent_config_dir` support for wrappers whose
  internal environment cannot be observed by AoE.
- Document and test precedence between `agent_config_dir`, effective session
  environment, daemon environment, and the home fallback.
- Add regression coverage for:
  - profile-provided `CLAUDE_CONFIG_DIR`;
  - a `before_session` value overriding the static profile value;
  - daemon-environment fallback;
  - default-home fallback.
- Ensure MCP forwarding and drift reconciliation use the same resolved file.
- Correct `docs/guides/configuration.md`, which currently says that only folder
  trust reads `agent_config_dir`, although native MCP discovery now reads it too.

Related: #3699 and [the unresolved review thread](https://github.com/agent-of-empires/agent-of-empires/pull/3699#discussion_r3914729378).

_AI disclosure: drafted with Codex from a cross-validated post-merge audit
directed by @jerome-benoit. The code path was independently verified; the scope
and publication decision remain his._

njbrake:
## Description

When a structured session's agent process was started, restarted or stopped, several independent bits of bookkeeping could disagree about who owned it. A stop could land while a start was still in flight and the late start would win; a crash-restart could leave a process alive that nothing tracked; a request to restart one process could be picked up by a different one. This change gives each process generation a single owner: every start, restart, stop and cleanup goes through one lease, a stop is not finished until the process is proven gone, and anything acting on an older generation is refused. The same branch closes the recovery gaps that sat beside that ownership problem, and a few structured-view issues in the same area.

**Lifecycle lease (#3487).** `acp::runner_lifecycle` mints an epoch per spawn, attach and respawn; the epoch is stamped on the runner's registry record as `generation`. Shutdown escalates SIGTERM, SIGKILL, then holds the session in a new `stopping` state (refuses prompts and resumes, retried each reconciler tick) until the process is proven dead. Restart markers carry the generation they stopped. The reconciler admits before it prepares and re-checks eligibility under the lease. A connection that fails before it establishes a session is no longer respawned by the drain task (which replaced the real error with the restart-budget message within two seconds); the reconciler retries it on its own budget.

**Recovery gaps (#3560, #3514, #3686).** Unknown WS control frames are dropped instead of reconnecting in a loop; a rejected resumed session resets context so the respawn opens fresh. The rate-limit park is a durable event-store query that a failed resume's startup error cannot mask; a limit hit during the handshake parks the session; the session payload carries the park and the auto-resume flag for the sidebar and banner. Plugin turns wake parked sessions (idle-dormant, archived or snoozed, the same clearing a user prompt does) and close orphaned turns.

**Same area (#3553, #3734, #3707).** `aoe acp approve --option <id>` answers an option-list request from the CLI, on top of #3779. `aoe-agent` is bundled and installed like the npm adapters, and `aoe acp agents` stops reporting a phantom. Native MCP discovery resolves the same `CLAUDE_CONFIG_DIR` the agent launches with. The history window pins its top row so a cut inside a long turn cannot snap forward and shrink a rendered part list.

Supersedes #3561. Rebased onto #3538 (Phase C): the reconciler's pre-spawn cleanup goes through its wait-for-exit registry helper and handles the incompatible-runner replacement, and its v3 control channel already correlates completions by request id, which is what #3203 asked for, so this branch no longer touches the control channel. #3779 landed the option-list approvals first; this branch keeps only the CLI flag. #3748's worker generation is the lease epoch rather than a second counter.

Left as they were, on purpose: the reconciler's build-stale respawn and the attach policy refusal still use the registry's pid-only `terminate` before a fresh spawn on the same socket, where the superseded runner self-terminates; folding those into the proven teardown is a follow-up. So is routing `aoe acp stop|kill|restart` through it (they signal from outside the daemon, and the reaper releases the lease on the record's disappearance), and proving the whole process group rather than the runner pid, which needs a zombie-aware membership probe in the OS layer. Redelivering the prompt a rejected stored session bounced is another (the reset callout asks for it again today).

#3545 was investigated and left out: the pi manifest has no running rule that the reported footer can match, so the stuck `Running` most likely comes from a spinner line the remote pi failed to clear through `ssh -t` (a terminfo or TERM mismatch). That needs the full pane capture from the reporter before any code change.

Fixes #3487
Fixes #3560
Fixes #3514
Fixes #3686
Fixes #3553
Fixes #3734
Fixes #3707

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

- [x] Skipped a test, because: every web change is covered by Vitest at the component or hook level (stopping banner, auto-resume notice, payload-driven badge summary, history-window pin). The #3514 badge is covered by the summary function test and a sidebar row test for the payload-driven badge and the stopping chip; the live story `sidebar-rate-limit` still exercises the park end to end.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the lifecycle races are driven through the production supervisor and reconciler entry points with a fake launcher and fake process control, which is the deterministic layer the issue asked for; a full-binary test cannot interleave a shutdown with a mid-handshake spawn.

## AI Usage

- [ ] No AI was used
- [x] AI was used

_Note: this PR was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

https://claude.ai/code/session_015vrmCC2wgaq2baqK28ms7R

## Summary

- Added per-session lifecycle leases and generation tracking for ACP runners.
- Hardened shutdown, respawn, reaping, recovery, registry updates, and restart authorization against stale operations.
- Added reliable teardown tracking and a visible `stopping` worker state.
- Improved recovery for unknown WebSocket frames, rejected resumed sessions, rate-limit parking, dormant plugin sessions, and orphaned turns.
- Bundled `aoe-agent` with installation checks and Node.js compatibility validation.
- Added option-list approval support across the CLI, daemon, and web interface.
- Aligned MCP discovery with the session’s effective `CLAUDE_CONFIG_DIR`.
- Stabilized history-window boundaries during transcript updates.
- Updated documentation and expanded lifecycle, recovery, approval, rate-limit, MCP, and UI tests.

## Benefits

The changes reduce runner race conditions and prevent stale processes from being reused. They improve recovery after failures and rate limits. They provide clearer worker status and more reliable plugin behavior. They also make `aoe-agent` installation more predictable and keep MCP discovery aligned with the launched agent.


src/acp/supervisor.rs
```diff
@@ -3493,67 +3686,224 @@ impl<S: BroadcastSink> Supervisor<S> {
...
+    });
+    let Some(identity) = identity else {
+        return Settlement::Proven;
+    };
+    let pid = identity.pid;
+    if !killed_before && control.is_alive(pid) {
+        control.terminate_group(pid);
+        wait_for_exit(control, pid, TEARDOWN_TERM_GRACE).await;
+    }
+    if control.is_alive(pid) {
```
jerome-benoit, line 3799: [P2] Keep the lease until the process group, not only its leader, has exited

`SystemProcessControl::is_alive` probes only the runner PID (`runner_lifecycle.rs:143-144`). If that leader exits after group SIGTERM while a grandchild ignores TERM, this branch skips group SIGKILL and the helper reaches `Settlement::Proven`, releasing ownership with an old descendant still alive. The runner's shutdown arm only kills/waits its direct child (`process/runner.rs:505-512`), so it does not establish the missing group proof.

Make group liveness and post-escalation settlement explicit, retaining `Stopping` until the owned group is gone. Add a deterministic process-control test where the leader is dead but a group member remains alive; the current PID-only fake cannot express it. This is a statically verified incomplete fix of #3487's process-group-exit requirement, not a claim of a new runtime reproduction.

jerome-benoit (changes requested):
Reviewed `cb767aa39fccd7d322b5aaa61247794c89d076f8`, including the linked issues, existing discussion and explicit follow-up boundaries. The lease design improves admission and stale-generation handling, but one central #3487 guarantee is still incomplete.

**[P2, blocking] Prove process-group exit before releasing ownership.** See the inline finding. This leak predates the PR; it is blocking here because the new teardown explicitly promises to fix it and uses that incomplete proof to release the lifecycle lease. It is not the separately deferred build-stale or attach-policy path.

Also route externally observed CLI stops through that proof: `src/cli/acp.rs:893-894,931-939` removes the registry before signalling and checks only the leader. The daemon then releases the lease through `drop_handle`/`reap_candidate` (`src/acp/supervisor.rs:2350,2429,3730`) without entering the teardown state. Registry disappearance must not substitute for process-group exit.

Existing CodeRabbit threads, without duplicate annotations:
- **[P2, blocking]** [Consume restart markers atomically](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875931). `reap_candidate` releases the lease at `supervisor.rs:3730` before consuming the marker at 3738. Another request can install generation B; a B restart marker written between the old consumer's read and delete (`worker_registry.rs:225-228`) is erased. Preserve newer restart authority through atomic publication/consumption and a generation-aware interleaving test.
- **[P2, blocking]** [Release the session-list read lock before the new blocking batch](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875948). `list.rs:38,46,200-226,254` retains borrowed instances while serial configuration/SQLite probes run, delaying session writers. Snapshot owned inputs and release the lock before awaiting.
- **Nonblocking:** [stale build-respawn flag](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875940), which predates this PR; [destructive-approval retry](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875954); [stopping-state controls](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875961), where backend rejection still protects ownership; [terminal-first sidebar chip](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958875970). These are not additional approval blockers.

Nonblocking: generic two-choice, same-kind lists without a `pi-ui-` ID still fall back to the first allow option (`src/acp/approvals.rs:107-117`). This follows #3741's proposed heuristic and should remain an explicit limitation, not another blocker. The documented resolved-answer transcript and rejected-prompt redelivery follow-ups are not blocking this review.

Two independent static reviews were arbitrated: the teardown findings are merged into one P2, not P1. Checked option-ID validation, unknown WS control-frame handling, durable rate-limit park queries, plugin wake, bundled-agent assets and history pinning. Evidence came from exact-object `git diff f85619aa cb767aa3` and numbered `git show` output; `git diff --check` passes. No local tests or runtime reproduction were executed. Only the PR-title check has succeeded at this head; there is no full current-head CI result or corresponding run to rerun.

jerome-benoit (changes requested):
Re-reviewed `91629b5e` against the previous review and the post-rebase fixes, with two independent sequential analyses and arbitration. Thanks for fixing the list-lock scope and stale-build flag, and for explicitly narrowing the teardown contract.

**Remaining merge blockers (P2), using the existing threads:**

- [Group signalling](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3958953386): defer full group-exit proof, but preserve the old unconditional initial group SIGTERM. `supervisor.rs:3778` now skips it when the leader is already dead, leaving descendants that the base would signal. This narrower regression does not require a new group-membership probe.
- [Registry ownership](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3959291991): hold the ownership read and deletion under the same registry lock, using strict loading, so cleanup cannot remove a replacement record/socket (`worker_registry.rs:665-677`).
- [Restart publication](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3959606786): rename-based claiming fixes the old read/delete window, but `fs::write` can still race the claim. Publish a fully written marker atomically and test that interleaving (`worker_registry.rs:197,228-230`).
- [Adapter publication](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3959291985): two processes can both start installation before either registers a runner; the slower publisher can then replace the faster one's active install. Coordinate publication with runner admission across processes, or defer automatic reinstall (`acp_client/mod.rs:356-381`, `adapters.rs:236,399`).

**Not blockers for this delta:** stopping-state UI affordances/sidebar selection, escalation-fake coverage and persistence-error logging. The duplicate-terminal emission predates this change; the ignored claim result remains worth fixing but is not a new regression. The ApprovalCard report concerns imported main code. Empty legacy markers should not force an unrequested compatibility layer; keep the documented support claim accurate. Full group-exit proof and the explicitly deferred lifecycle paths remain separate scope, not additional requirements here.

Evidence: exact-object base/head and caller inspection, `git diff --check b94eb987 91629b5e` (exit 0), live description and all review threads. Static validation only, no new runtime reproduction or local tests. CI is still running. No duplicate inline comments added.

jerome-benoit (changes requested):
Re-reviewed `91629b5e..6fc25e56`, including the existing scope decisions and new review threads. The four previous blockers are addressed: unconditional initial group TERM (`supervisor.rs:3778–3782`), strict ownership check/delete under one lock (`worker_registry.rs:664–678`), rename-based marker publication (`193–210`), and removal of automatic adapter reinstall.

**One P2 remains**, already reported in [this thread](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3959773566): `acp_reconciler.rs:1079–1084` invents generation zero when the registry is missing. For an adopted build-stale runner of generation `g > 0`, CLI restart writes `g` and deletes the record (`cli/acp.rs:1011–1012`). The next tick runs stale-worker handling before the reaper (`acp_reconciler.rs:218,230`), overwrites that valid marker with `0`, and the reaper classifies it as `user_stopped` instead of authorizing respawn (`supervisor.rs:3699–3704`). Preserve the attached identity or leave the existing marker untouched when generation is unavailable; add this interleaving to the regression coverage. This is a defect in the new generation binding, not a request to expand the deferred teardown/group-proof work.

**Rejected:** [the missing session/load event report](https://github.com/agent-of-empires/agent-of-empires/pull/3758#discussion_r3959773558). Successful load emits `AcpSessionAssigned` at `connection.rs:1202–1206`; reattach also emits it at `965–969`. The state and durable park consume that event. Previously classified nonblocking/preexisting points remain so.

Two independent sequential analyses and parent arbitration; exact-object numbered `git show`, caller tracing, and `git diff --check` (clean). Static evidence, no runtime reproduction or local tests. CI results are separate from this remaining code finding. No duplicate inline comments.

jerome-benoit (approved):
Re-reviewed `6fc25e56..6cd3341b` with two independent sequential analyses and arbitration. The last P2 is addressed: `acp_reconciler.rs:1079–1093` uses the attached runner's generation and preserves an existing marker when neither identity nor registry is available. The regression at `3056–3082` covers both missing-registry cases. Caller inspection confirms that the reaper consumes that same generation (`supervisor.rs:3714–3719`). No new blocker found; the previously documented nonblocking/deferred scope remains unchanged.

Evidence: exact-object diff and numbered caller inspection; `git diff --check 6fc25e56 6cd3341b` passes. Static review, no local tests. CI is separate: the mocked Playwright job reports 423 passed / 3 skipped, then fails artifact finalization with HTTP 403; other jobs are still running. This approval is not a claim that CI is globally green.

jerome-benoit (approved):
Renewing approval after independently reviewing the rebase to `8bcbf28f`, with two sequential analyses and arbitration.

`git range-diff b94eb987..6cd3341b b4098e09..8bcbf28f` marks all four PR commits unchanged. Exact-tree comparison of the supervisor, registry, reconciler, lifecycle and connection code also returns no differences. The previous fixes, including preservation of the attached generation (`acp_reconciler.rs:1079–1093`) and its regression coverage (`3056–3082`), remain intact. Imported main changes were separated from the PR-owned patch; no new blocker found. Existing nonblocking/deferred scope remains unchanged.

Evidence: the range-diff, `git diff --quiet` for those five files (exit 0), and `git diff --check b4098e09 8bcbf28f` (clean). Static rereview only, no local tests. Codecov patch now passes at 93.61% against the 75% target; the latest CI snapshot still has Nix Build running, so this is not a global CI-pass claim.

PR #3759, opened 2026-09-05
fix(tmux): prevent unread vt pipe replay after reseed

### LINKED ISSUE #3737: Prevent unread VT pipe output from being replayed after snapshot reseeds
## Problem

A VT snapshot can include pane output that is still unread in the pipe socket.

The current fence advances only after `run_reader` calls `read()`. A reseed can therefore install a snapshot containing a chunk while `chunk_seq` remains unchanged. When the reader subsequently drains the same chunk, it applies it again.

## Reproduction

A deterministic unit reproduction:

1. Queue `UNREAD-MARKER` in the pipe socket without starting the reader.
2. Install a snapshot containing `UNREAD-MARKER`.
3. Start `run_reader`.
4. Observe `UNREAD-MARKERUNREAD-MARKER` in the parser.

This reproduces on current `main` at `1e1ef1e187feeaf2f8c89e7e45ecd7bf01807827`.

## Impact

The in-process VT grid can duplicate terminal output after initial or authoritative snapshot reseeds. Shell output may remain visibly corrupted until a later successful resync.

Priority: P2, timing-dependent.

## Expected fix

Establish an ordering boundary that also covers bytes not yet read from the pipe, rather than only chunks already claimed by `run_reader`. Add the deterministic unread-socket regression described above.

## AI usage

Drafted with Codex from a cross-validated post-merge review of #3636. The reproducer was executed locally against the current implementation.

atirna:
## Description

Prevent VT snapshot replay when pane output is still in flight.

Installing a `capture-pane` snapshot used to race the pane bytes feeding the
parser, so output already in the snapshot could be applied a second time when it
arrived. Three parts close that:

- a control socket beside the data socket, over which the `pipe-pane` forwarder
  acknowledges a generation-tagged drain probe only after flushing its whole
  stdin backlog;
- a `FIONREAD` check on the reader socket at install time;
- a snapshot mutex held by the reader across receive and parse, and by the seed
  from the drain through parser replacement, which makes the first two atomic
  against the reader.

Every seed installs through that fence, including the reseeds behind an
authoritative refresh, a resize, and a drift repair. A resize repaints the pane,
so its reseed retries on the arm cadence until the reader drains rather than
giving up on the first `Busy`.

The window this does not cover is tmux's own pipe output buffer: `capture-pane`
can show a byte that `bufferevent_write` has queued but not yet written, which no
`FIONREAD` on either side can see.

Fixes #3737

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the ordering boundary is covered by deterministic Unix-socket regressions in `src/tmux/vt.rs`; a tmux end-to-end race would not give a stable assertion.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
OpenAI Codex (gpt-5.6-terra) for the original implementation. Claude Opus 5 via Claude Code for the later merge with main and the reseed rework, driven by @njbrake.

Used for investigation, implementation, and regression coverage; the author reviewed the original patch and verification.


jerome-benoit (changes requested):
Requesting changes because the receive fence is still not atomic with snapshot installation.

The existing CodeRabbit thread is valid: `run_reader` can drain the socket and wait on `parser` after the counter checks, leaving `FIONREAD` at zero while the same bytes are already present in the captured snapshot.

There is also an earlier unobserved window in `run_pipe`: bytes can remain between `stdin.read()` and `sock_w.write_all()`. tmux queues pane output to `pipe-pane` before parsing it into the pane screen, so a capture can already contain those bytes while neither `chunk_seq` nor `FIONREAD` can observe them.

Please establish one end-to-end ordering boundary covering forwarding, reader receipt, and snapshot commit. Add deterministic regressions that pause both after the forwarder read and after the socket read. Reordering or repeating the current checks is insufficient.

The substantive CI, Tests, and CodeQL workflows have not run; their current conclusion is `action_required`.

src/tmux/vt.rs
```diff
@@ -222,6 +243,138 @@ pub(crate) fn run_pipe(socket: &str) -> std::io::Result<()> {
...
+    };
+    if control
+        .set_read_timeout(Some(Duration::from_millis(100)))
+        .is_err()
+        || control.write_all(&[DRAIN_PROBE]).is_err()
+    {
+        return false;
+    }
+    let mut ack = [0u8; 1];
+    control.read_exact(&mut ack).is_ok() && ack == [DRAIN_ACK]
```
jerome-benoit: **[P1] Correlate drain acknowledgements across retries**

If this `read_exact` times out, the function returns `Busy` but leaves the same control stream usable. The forwarder can finish later and write the old `DRAIN_ACK`; the next reseed then writes a new probe and may immediately consume that stale byte as its acknowledgement. If the forwarder is holding a new pane byte at that point, the data socket is still empty and the snapshot can install before the byte is forwarded, reopening the duplicate-replay race. Please make a timeout terminal for this control connection or correlate each probe/ACK with a generation, and add a deterministic timeout -> late ACK -> retry regression with a newly held byte.

jerome-benoit (changes requested):
Re-review of `66a174f8a4662e3331e7bcc9e99d4e330ba44963`. The new snapshot mutex and forwarder drain address the two previously reported single-attempt races, and the full CI is green. One retry-specific ordering hole remains before this is safe to merge.

src/tmux/vt.rs
```diff
@@ -363,6 +384,146 @@ pub(crate) fn run_pipe(socket: &str) -> std::io::Result<()> {
...
+fn drain_forwarder(control: &Mutex<Option<UnixStream>>) -> bool {
+    use std::io::Write;
+
+    let Ok(mut control) = control.lock() else {
+        return false;
+    };
+    // A timeout leaves an acknowledgement in flight. Keep the stream out of
+    // the channel until its own probe completes, so that late byte cannot
+    // become the acknowledgement for a later snapshot attempt.
+    let Some(mut stream) = control.take() else {
```
jerome-benoit: [P1] Do not permanently disable reseeding after one drain timeout. `control.take()` removes the only control stream and restores it only after a valid ACK. On timeout or I/O failure it is dropped; `run_drain_listener` accepts only once and the forwarder never reconnects. Every later `drain_forwarder` therefore returns false and every reseed remains `Busy`, although `is_alive()` stays true, permanently disabling authoritative repair and resize reconciliation for this live channel. Preserve stale-ACK isolation with correlated probe generations, or retire and re-arm the whole channel, and add a timeout-then-successful-reseed regression.

jerome-benoit (changes requested):
Re-review of exact head `7e56a32819166958dda5dbb9456d934a3cfc4657`, with two independent blind analyses followed by code-level arbitration.

The stale-ACK retry race is fixed: after an unsuccessful probe, a late ACK can no longer certify a later attempt. The replacement safety mechanism nevertheless introduces a blocking recovery defect: one failed drain permanently removes the channel's only control stream while the data channel remains alive, so every later authoritative refresh or geometry reseed returns `Busy`. This is annotated inline as P1.

The existing CodeRabbit stream-mutex finding is also valid and remains a P2; I have not duplicated it inline: https://github.com/agent-of-empires/agent-of-empires/pull/3759#pullrequestreview-5122192345. `reseed` holds `self.stream` across the drain wait and parser replacement, so `write_input` can stall for the 100 ms timeout plus the swap.

The previous reader-receipt and forwarder-held-byte races are addressed, and the missing `ReaderCtx.snapshot` initializer is fixed.

CI at this exact head is red: live Playwright reports five failures, including both VT-grid transport cases failing to arm the grid. This does not prove the inline defect caused those failures, but the affected path is in scope and the failures must be resolved before approval. `Cargo Test (linux, bare-core)` was still pending at the final check. No local build or heavy test was run.

src/tmux/vt.rs
```diff
@@ -1394,6 +1680,12 @@ fn run_reader(listener: UnixListener, ctx: ReaderCtx) {
     let mut osc52 = Osc52Scanner::new();
     let mut sync = SyncOutputScanner::new();
     while !ctx.stop.load(Ordering::Relaxed) {
+        // A snapshot holds this same mutex from its forwarder drain through
+        // parser replacement. A reader that has received a chunk therefore
+        // settles it before the snapshot can inspect the socket queue.
+        let Ok(_snapshot) = ctx.snapshot.lock() else {
```
jerome-benoit, line 1983: **[P1] Do not hold the snapshot fence while waiting for input**

This mutex is held across `conn.read`, which can block for 200 ms, and the timeout path immediately loops and reacquires the same non-fair mutex. `seed_parser` and `reseed` wait on this mutex without a bound, so an idle reader can starve initial grid arming, authoritative refresh, and resize reconciliation indefinitely. A std-only reproducer mirroring this loop consistently reported `snapshot still blocked after 3s on an idle data socket`; the waiter completed only after the reader was stopped.

Wait for socket readiness outside this mutex, then acquire the fence and perform a nonblocking read plus chunk settlement. This keeps unread bytes visible to the seed's `FIONREAD` check when the seed wins, and keeps receipt through parser settlement atomic when the reader wins. Please add an idle-socket regression proving a live reader does not block snapshot acquisition, while retaining the received-before-parse race coverage.

jerome-benoit (changes requested):
Re-reviewed exact head `44485598bf558831b6bd0b8d8ddc29d58ad7c59d` with two independent analyses followed by code-level arbitration.

The generation-matched drain protocol fixes the stale-ACK retry and permanent timeout-retirement findings. Cloning the pipe socket also removes the input-mutex stall. One new blocking issue remains: the reader holds the snapshot mutex across an idle blocking read and can starve every seed/reseed waiter indefinitely. This is annotated inline as P1 and was reproduced with the same mutex/read/timeout control flow.

The exact-head Rust tests, Clippy, Format, and CodeQL pass. Playwright live remains red after one rerun: six tests failed on the rerun, including both live-grid transport scenarios. That does not by itself prove causation, but the failures are in scope and the run is not green.

src/tmux/vt.rs
```diff
@@ -1389,11 +1675,32 @@ fn run_reader(listener: UnixListener, ctx: ReaderCtx) {
     // single-writer. `acquire` is blocked until this flips.
     ctx.alive.store(true, Ordering::Relaxed);
     let mut conn = conn;
-    let _ = conn.set_read_timeout(Some(Duration::from_millis(200)));
+    let _ = conn.set_nonblocking(true);
```
jerome-benoit: **[P2] Keep the published input socket blocking**

`try_clone()` above shares socket options with this descriptor, so `set_nonblocking(true)` also makes the clone stored in `ctx.stream` nonblocking. `write_input()` still uses `write_all()` and treats any error as a total failure. Under backpressure, `write_all()` can accept a prefix and then return `WouldBlock`; `live_ws.rs:1239-1244` subsequently resends the entire payload through `send_raw_bytes`, duplicating that prefix and mixing two input writers while the VT forwarder is still alive. A socket-pair probe reproduced this boundary: `clone_nonblocking=true`, `errno=11`, `prefix_sent=219264`, `peer_received=219264`, for a 1 MiB payload. Please preserve blocking mode for the published clone and make only this receive operation nonblocking, for example with `recv(..., MSG_DONTWAIT)` after `poll`. Add a backpressure regression proving an input payload is delivered exactly once.

jerome-benoit (changes requested):
Re-reviewed exact head `f7239ef8101cc30e7329f5a0319b2f2351647d03` through two independent blind analyses followed by code-level cross-validation.

The previous snapshot-fence starvation P1 is fixed: readiness is now awaited outside the fence, and the new idle-reader regression covers that boundary. The receive/install ordering, generation-matched drain protocol, and released input mutex remain intact.

One new P2 input-corruption regression blocks approval and is annotated inline. Setting the reader socket nonblocking also changes the published input clone, invalidating the fallback's blocking-write invariant. A minimal socket-pair reproduction reached `WouldBlock` after accepting a 219,264-byte prefix; the fallback would then resend the complete payload through tmux.

All exact-head CI checks are green, including Rust tests, Clippy, Format, CodeQL, and Playwright live. `git diff --check` is clean.

src/tmux/vt.rs
```diff
@@ -363,6 +402,160 @@ pub(crate) fn run_pipe(socket: &str) -> std::io::Result<()> {
...
+                    // proven empty, stay silent and let the channel time
+                    // out into Busy.
+                    let mut ack = true;
+                    loop {
+                        let mut pending: libc::c_int = 0;
+                        if unsafe { libc::ioctl(stdin_fd, libc::FIONREAD, &mut pending) } != 0 {
+                            ack = false;
+                            break;
+                        }
+                        if pending <= 0 {
```
jerome-benoit, line 588: **[P2] Fence tmux's pending pipe output before acknowledging**

This `FIONREAD` check covers only bytes already delivered to the forwarder's stdin. tmux 3.5a calls `bufferevent_write(wp->pipe_event, ...)` and then `input_parse_pane(wp)`, so the captured screen can already include a byte that is still queued in tmux's userspace `bufferevent` and is invisible here. The ACK and downstream `FIONREAD` can therefore both pass, the snapshot can install, and tmux can deliver that same byte afterward for a second parse. A libevent/socketpair probe confirmed the missing boundary: the output queue held 18 bytes while forwarder `FIONREAD` was zero, becoming readable only after the event-loop flush. Please establish a producer-side ordering boundary that covers tmux's pending pipe queue (repeating this check or sleeping is not sufficient), and add a regression for `screen parsed -> pipe output still buffered -> snapshot commit`.

jerome-benoit (changes requested):
Re-reviewed exact head `229dd1966731d47c18a64d8aa2cbeba4bd67e3d0` through two independent analyses followed by targeted cross-validation.

The previous input-corruption finding is fixed correctly: the published socket remains blocking, while only the reader uses `recv(..., MSG_DONTWAIT)` after `poll`; the backpressure regression covers a complete 1 MiB write. The earlier snapshot-starvation, received-chunk, stale-ACK, control-recovery, and input-mutex findings also remain fixed.

One P2 ordering gap still blocks approval and is annotated inline. The drain protocol observes the forwarder's stdin and the downstream Unix socket, but not tmux's own `bufferevent` output queue. tmux parses pane bytes into the captured screen immediately after queuing them for asynchronous pipe delivery, so a snapshot can contain bytes that neither `FIONREAD` can see yet; those bytes can arrive after the swap and be applied again. This is the same timing-dependent visible-corruption class as #3737.

Evidence: [tmux 3.5a `window.c`](https://github.com/tmux/tmux/blob/3.5a/window.c#L969-L983) queues with `bufferevent_write` before `input_parse_pane`; [libevent's contract](https://libevent.org/libevent-book/Ref6_bufferevent.html#_manipulating_data_in_a_bufferevent) makes that write asynchronous. A lightweight libevent/socketpair probe observed `tmux_output_queue=18` while `forwarder_FIONREAD=0`, then `FIONREAD=18` only after the event-loop flush. The complete capture/ACK interleaving is a code-level inference, not an end-to-end tmux reproduction.

All six previous inline threads are resolved. Exact-head CI is currently green except `Cargo Test (linux, bare-core)` and Rust CodeQL analysis, which are still running after workflow approval; no failure is currently reported. No local heavy tests were run.

src/tmux/vt.rs
```diff
@@ -1969,53 +2292,25 @@ impl VtChannel {
...
-    /// Rebuild from an authoritative tmux snapshot even when cursor and
-    /// geometry probes agree, healing cell drift that those probes cannot see.
+    /// Tmux can have output represented by `capture-pane` while it is still in
+    /// its own pipe queue, beyond the forwarder's ordering fence. Replacing the
+    /// parser would replay that output when tmux flushes it, so consumers must
+    /// use the capture-only fallback instead.
+    fn retire_live_channel(&self, deadline: &crate::tmux::TmuxCommandDeadline) -> VtRefreshResult {
+        self.alive.store(false, Ordering::Relaxed);
+        self.shutdown_with_deadline(deadline);
+        VtRefreshResult::Busy
```
jerome-benoit: **[P2] Keep intentional capture fallback out of automatic VT recovery**

Returning `Busy` after setting `alive=false` does not keep the TUI on capture-only. `live_send.rs:1597-1599` drops this channel, then `1577-1588` automatically calls `acquire_with_deadline` again. The three-second re-arm throttle is already expired when the normal ten-second quiet refresh retires a healthy channel (`759`, `1459-1464`, `1612`). Each successful replacement resets that refresh clock, producing a recurring retire/re-arm cycle on an unchanged pane. Besides needless pipe/process churn, each re-arm again starts the pipe before installing its initial snapshot (`vt.rs:2071`, `2113-2145`), where the existing tmux-queue ordering limitation remains; visible corruption was not reproduced end to end here.

Please distinguish intentional capture-only retirement from a recoverable disconnect and suppress automatic re-arming for that retirement until an explicit lifecycle boundary. Add a worker-lifecycle regression covering refresh, fallback and subsequent arm decisions; the new `alive` assertion alone does not exercise this path.

src/tmux/vt.rs
```diff
@@ -2183,13 +2476,7 @@ impl VtChannel {
...
             return VtRefreshResult::Refreshed;
         }
-        let result = self.reseed(cols, rows, false, deadline);
-        if refresh_commits_geometry(result) {
-            self.cols.store(cols, Ordering::Relaxed);
-            self.rows.store(rows, Ordering::Relaxed);
-            self.signals.bump_changed();
-        }
-        result
+        self.retire_live_channel(deadline)
```
jerome-benoit: **[P2] Preserve web clipboard observation when retiring the grid**

With clipboard forwarding enabled, open a Grid connection, resize it to different dimensions, then have the pane emit an OSC 52 clipboard update. This call now stops the VT reader. Rendering correctly switches to capture, but `live_ws.rs:675-679` created the raw OSC52 observer only when the connection initially had no VT. The connection retains `capture_osc52=None` and the old `capture_vt` (`695-708`), so clipboard selection at `978-985` continues reading the stopped channel and cannot receive subsequent updates. This is a source-traced regression, not an executed browser reproduction.

Please keep capture-only rendering and transfer clipboard observation to an OSC52-only channel when VT is intentionally retired, initializing its per-connection sequence appropriately. Add a regression that verifies a new clipboard event is forwarded after a real resize-triggered retirement on the same WebSocket. Re-arming the VT grid is not required.

jerome-benoit (changes requested):
Thanks for implementing the capture fallback. Re-reviewed the delta at `5fda24f81452ff025b1fdc616b1c960de2a1373b`, including previous replies and TUI/web consumers, through two independent sequential analyses and code-level arbitration.

The intentional switch from Grid to capture-only is accepted; restoring live reseeding is not requested. Two P2 integration regressions still block approval, both annotated inline: the TUI automatically re-arms intentionally retired channels, and the web fallback loses subsequent OSC 52 clipboard events. Please address both. The direct live-parser replacement is removed. A transient stale frame was considered but not retained as a separate blocker.

Evidence is exact-SHA control-flow inspection plus a standalone Rust probe of the TUI scheduling predicates, not an end-to-end tmux corruption reproduction. No full local build/test suite was run.

The Tests workflow has now completed: Rust tests, Vitest and mocked Playwright pass. Live shard 2 has one failure, `synchronized-output brackets publish whole frames only`, with `Expected: grid; Received: snapshot` (81 passed, 1 failed; [logs](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34047810310/job/101528002300)). The [test](https://github.com/agent-of-empires/agent-of-empires/blob/5fda24f81452ff025b1fdc616b1c960de2a1373b/web/tests/live/live-grid-transport.spec.ts#L74-L120) fails at the transport assertion, before checking frame synchronization. The logs do not establish whether this particular fallback came from retirement after resize or an initial arming failure, so it cannot yet be classified as transient.

Please keep meaningful grid-only synchronization coverage under controlled geometry and separately cover the agreed capture-only transition. Do not merely remove the transport assertion or rerun until green. This CI failure neither proves the two P2 findings nor adds a demand to restore live reseeding.

src/tmux/vt.rs
atirna: changed this so intentional grid retirement is distinct from a dead forwarder. the worker stays on capture for this target until retarget or a VT setting reset, while normal disconnects keep the re-arm path.

src/tmux/vt.rs
atirna: changed the existing connection to start an OSC52-only observer after its grid retires, starting at that observer current sequence. capture rendering stays in place and later clipboard updates keep forwarding.

jerome-benoit (approved):
Re-reviewed `5610c157a50b5ac4626a65a1212b64ef6634ef1b`, including the previous replies, with two independent sequential analyses and cross-validation.

Both remaining P2 findings are addressed: shared lifecycle state prevents the TUI from re-arming intentionally retired channels, and the web capture fallback preserves subsequent OSC 52 clipboard events. The agreed capture-only fallback remains intentional; restoring live reseeding is not required.

All three live grid-transport tests passed locally, including controlled-geometry synchronization and resize-to-capture clipboard coverage. Current-head CI is green. The two ACP cases that initially failed also passed locally and on the targeted CI rerun; their original failure cause is not established.

No remaining actionable findings from this re-review. Thanks for working through the fixes.

jerome-benoit (approved):
Re-reviewed `5610c157..b886d6af`, including the main integration and the existing capture-only decision. Two independent sequential analyses and arbitration found no new blocking defect.

The merged seed path carries the link table through `SeedSink`; its shared snapshot lock covers reader receipt, link recording and parser application (`src/tmux/vt.rs:1104,1948,2009-2017`). This excludes #3818's documented interleaving. Retired/failed channels now stop serving link metadata (`1870-1893`), and the preview's generation check refreshes its link list (`src/tui/home/preview.rs:228-243`).

#3817 is not evidence of a new regression here: its persistent-after-Busy scenario is avoided by retirement on a changed size (`vt.rs:2844-2852`). The imported resize guards remain, so I am not claiming that the mechanism or every possible pre-retirement timing window has been removed. Keep the issue's scope distinction explicit.

Nonblocking P3: update the inherited comments describing parser rebuilding and reseed retries (`live_ws.rs:329-338,372-381`; `vt.rs:2855-2858`) to describe retirement/capture fallback.

Evidence: exact-object `git diff`, conflict-resolution inspection and numbered `git show`; `git diff --check` passed. The five-file PR-owned delta was separated from imported main changes. CI approvals were cleared and checks are still running; this is not a full-CI success claim. No local tests or runtime reproduction were run.

jerome-benoit (changes requested):
Re-reviewed `b886d6af..2b0f8edf`, including the earlier responses, with two independent sequential analyses and evidence-based arbitration. Two P2 regressions prevent renewing the approval:

1. **Active reseeding restores the producer-side ordering gap** ([existing thread](https://github.com/agent-of-empires/agent-of-empires/pull/3759#discussion_r3942234608)). `vt.rs:2688–2708` reinstates snapshot replacement, but capture occurs before the child drain (`1099`, `1207–1210`). The child ACK (`583–605`) and downstream checks (`1177–1182`) cannot see tmux's own output buffer. In [stable tmux 3.7c](https://github.com/tmux/tmux/blob/3.7c/window.c#L1149-L1162), bytes are queued before the screen is parsed; [bufferevent_write](https://github.com/libevent/libevent/blob/release-2.1.12-stable/bufferevent.c#L451-L458) only appends to that buffer. A capture can therefore contain bytes delivered after the swap, which are parsed again. Please retain the approved capture fallback, including its lifecycle handling, until replacement has a producer-side ordering boundary. This blocks the newly restored repeated reseeds, not the unchanged initial-arm limitation.

2. **Web clipboard recovery is removed** ([related existing thread](https://github.com/agent-of-empires/agent-of-empires/pull/3759#discussion_r3944727057)). At `live_ws.rs:781–810`, a connection starting with VT has no OSC52 observer. After a forwarder failure, rendering excludes that dead channel (`849`), but clipboard selection still uses it (`1127–1133`); the dynamic observer recovery was deleted. Please preserve recovery after genuine VT failure and test a subsequent OSC52 update on the same WebSocket. This is not dependent on intentional retirement.

Verified: geometry is committed only after `Refreshed`. Nonblocking: the description and retirement comments need updating; the resize test's already-true grid assertion does not establish resize completion.

Evidence: numbered `git show 2b0f8edf:<path>`, the exact-head delta, and the upstream sources linked above. These are source-traced failure paths, not runtime reproductions. No local build/tests were run. CI was authorized and remains in progress; no CI pass is claimed.

src/tmux/vt.rs
```diff
@@ -835,20 +1080,47 @@ struct SeedSink<'a> {
...
+        return swap_seeded_parser(sink, since, &stream, size, guard);
+    };
+    let Ok(_snapshot) = snapshot.lock() else {
+        return VtRefreshResult::Failed;
+    };
+    let Ok(socket) = socket.lock() else {
+        return VtRefreshResult::Failed;
+    };
+    let guard = SeedGuard {
+        pipe: socket.as_ref(),
```
jerome-benoit: [P2] Release the input socket mutex before waiting for the drain ACK

The `socket` guard remains borrowed through `swap_drained_seeded_parser`, which waits in `drain_forwarder` (`1207`, with a 100 ms ACK deadline at `634`). Live reseeding passes `self.stream` here (`2717`), and `write_input` needs that same mutex (`3118`). A delayed ACK therefore blocks keyboard input during the wait; resize retries can repeat it. The actual base reseed did not hold this input mutex.

Clone the socket under the mutex and release the guard before draining, while retaining the snapshot fence and using the clone for `FIONREAD`. Preserve lock/clone error handling. Add a deterministic delayed-ACK test showing input can acquire its mutex while the reseed waits. This also preserves the intent of the earlier socket-cloning fix discussed in this PR. Static lock/caller tracing confirms the contention; no new runtime reproduction claimed.

jerome-benoit (changes requested):
Re-reviewed `8622e4d8`, including the author replies, revised partial-fix scope and the actual merge base (`5673763b`), with two independent sequential analyses and arbitration.

**One P2 blocks approval:** live reseeding holds the input socket mutex while waiting for the drain ACK (inline below). This is introduced relative to the actual base, not by the latest formatting commit, and was missed in my previous review.

**Scope correction:** I withdraw the blocking classification of the two earlier limitations. The upstream tmux buffer window already exists in the base and is now explicitly acknowledged; the clipboard-recovery difference was against the abandoned `b886` candidate, not the actual base (`src/server/live_ws.rs` is unchanged). Neither requires expanding this partial fix. The upstream limitation still needs follow-up; this PR does not completely eliminate that replay window.

The resize retry refreshes its sequence guard per attempt; the latest test-only delta introduces no additional finding. Evidence: exact-object diff and numbered caller inspection, `git diff --check c13e7073 8622e4d8` (clean). Static validation, no runtime reproduction or local tests. Tests, CI and CodeQL for this SHA were awaiting permission and have now been authorized; no claim of green CI.

jerome-benoit (approved):
Re-reviewed `dd8d5ea5`, including the earlier replies and the merge from `b4098e09`, with two independent sequential analyses and arbitration.

The last P2 is fixed: `src/tmux/vt.rs:1129–1138` clones the socket and drops the input mutex before the drain wait (`1143`). The snapshot guard (`1120`) remains held, and the receive-queue check uses the cloned descriptor. No new blocking regression found. The previously acknowledged upstream tmux-buffer limitation remains outside this partial fix; the earlier scope correction stands.

Nonblocking P3 test hardening: at `vt.rs:4764–4778`, scheduling past the production 100 ms drain timeout could let the old locking implementation falsely pass `try_lock`; the seed join result is also ignored. A controllable drain boundary and an asserted worker result would make this regression deterministic. This does not invalidate the inspected production fix.

Evidence: exact-object `git diff 8622e4d8 dd8d5ea5 -- src/tmux/vt.rs`, numbered `git show`, and `git diff --check b4098e09 dd8d5ea5` (clean). Static review only; no local tests or runtime reproduction. CI is still in progress, not reported as green.

Issue #2864, opened 2026-07-15
test: named serial group (shell_env) does not exclude default #[serial], racing CLAUDE_CONFIG_DIR (8/10 fail on main)


njbrake:
## Summary

`serial_test` only serializes tests **within the same group**. A named group and the default group do not exclude each other. The repo mixes the two on tests that mutate the same process-global env vars, so they run concurrently and yank each other's environment mid-test.

Concretely: the migration tests are annotated `#[serial_test::serial(shell_env)]` (named group) while the `session::instance` tests are plain `#[serial]` (default group). The migrations call `remove_var("CLAUDE_CONFIG_DIR")` while an instance test is mid-scan.

## Reproduction

On a clean checkout of `origin/main`, running the instance tests in-process alongside the migration tests:

```
cargo test --lib
```

`session::instance::tests::resume_fallback::verify_on_resume::mtime_fallback_applies_without_sidecar` fails **8 of 10 runs**.

Under `--test-threads=1` it passes **3 of 3**. That confirms parallelism, not test logic.

This is pre-existing on `main` and is not caused by any open PR. It was found while rebasing #2860, and initially suspected to be that PR's fault; verifying against a clean `main` checkout ruled that out.

## Why it matters beyond the one test

Any PR that shifts test timing can make this fire more often without changing anything relevant. #2860 (which adds four unrelated tests to `test_support.rs`) hits it 10/10 purely through timing perturbation, so it risks a red CI run on a bug it did not introduce. That is the worst kind of flake: it blames the wrong change.

## Suggested fixes

1. Put every test that mutates shell/agent-home env into **one** serial group, rather than a mix of `#[serial]` and `#[serial_test::serial(shell_env)]`. Either move all of them into the named `shell_env` group, or drop the group name so they share the default. The important thing is that they actually exclude one another.
2. Better long term: stop mutating process-global env in tests at all where feasible, so serial grouping is not load-bearing. `EnvGuard` (landing in #2860) is most of the infrastructure for this already.

## Related

This is the fourth instance of one pattern: tests mutating process-global env under parallel `cargo test`.

- #2600, flaky file_watch tests, global HOME env race under parallel cargo test
- #2861, chmod-based write-failure injection gives false failures when tests run as root
- #2863, `AllowFileTransport`'s process-global env mutation races with parallel git-spawning tests, hanging macOS CI for 6h
- #2716 / #2751, env guard consolidation behind `EnvGuard`, snapshotting `Option<OsString>`

[...]

PR #3760, opened 2026-09-05
fix(tests): stop HOME-reading tests racing the suite's HOME writers

njbrake:
## Description

Some tests were reading the `HOME` environment variable while other tests were changing it, so they occasionally saw two different values in the middle of a single check and failed for no real reason. This made the test suite fail at random about one run in three when the machine was busy, with nothing actually wrong with the code. The fix gives those tests their own throwaway home directory and stops them running alongside the tests that change it, so they now check something fixed instead of whatever the rest of the suite happened to leave behind.

`expand_tilde_expands_home` reads `$HOME` twice: once directly via `dirs::home_dir()`, then again inside `expand_tilde`. 163 call sites across 25 modules run `set_var("HOME", ...)`. Those writers are `#[serial]`, but `#[serial]` only excludes other `#[serial]` tests, so a parallel reader races every one of them and a write landing between the two reads makes them disagree.

I reproduced the mechanism standalone rather than inferring it from the failure text. The same two-read pattern against a concurrent writer:

```
two-read disagreements: 55 / 200000
```

Observed locally at roughly 1 failure in 3 full-suite runs while the machine was loaded, and 0 in 5 runs while idle, which is why it reads as random. It surfaced during work on #3754 and is unrelated to that change.

The fix routes `$HOME` through `test_support::isolate_home`, which holds the process-global `ENV_LOCK` for the guard's lifetime and restores the prior value on Drop, and keeps `#[serial]` to match every existing call site. `dirs-sys::home_dir()` reads `$HOME` first on both Linux and macOS (falling back to `getpwuid` only when unset), so pinning it steers resolution on both CI platforms.

`#[serial]` alone would not have been enough, which is the point `src/session/test_support.rs` already makes: `ENV_LOCK` exists because serial groups were "the discipline that broke" in #2864 and #2600, and a bare `set_var` is named in that file as the case that still races. Concretely, `src/session/storage.rs` has 22 tests that move `$HOME` through `isolate_app_dir_at` with only 2 carrying `#[serial]`; they are safe because the guard takes the lock, which a bare `set_var` does not. A bare `set_var` also never restores `$HOME`, leaving later tests pointing at a deleted tempdir, which is the #2600 failure mode.

### Why this touches a second file

The four `shorten_path` tests in `tui::components::preview` have the identical two-read shape against the same writers: the test reads `$HOME`, then `shorten_path` reads it again. They are the same defect, so fixing them here rather than filing a follow-up for identical code. They shared setup and differed only in input and expectation, so per AGENTS.md they become one table case.

Pinning `$HOME` also gives those four something to assert. Each was wrapped in `if let Some(home) = dirs::home_dir()`, so an unset `$HOME` made all four pass without ever running an assertion.

A scan for tests that spell `set_var("HOME"` in their own body and lack `#[serial]` returns these five. That scan does not see guard-based writers, which is why the first version of this PR reached for `#[serial]` instead of the guard.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

This change is itself test-only: the production code is untouched, and the change is what makes five existing assertions deterministic. Verified by 3 full-suite runs with no `expand_tilde_expands_home` failure, against a pre-fix baseline that failed 2 of 6.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Found while investigating an unexplained third test failure during #3754. Diagnosis, the standalone race reproduction, and the fix were done by Claude Opus 5 in back-and-forth with @njbrake.

https://claude.ai/code/session_01RCYiZhmr6riURz45y3eE9i

## Summary

- Isolated `HOME` for `expand_tilde_expands_home` and `shorten_path` tests.
- Added environment locking and automatic restoration of `HOME`.
- Consolidated four `shorten_path` tests into one table-driven test.
- Removed the inaccurate comment about skipped assertions.
- Left production code unchanged.

## Benefits

- Prevents flaky tests caused by concurrent `HOME` access.
- Makes path assertions consistent across environments.
- Improves test coverage and readability.
- Full-suite runs passed after the changes.

### DIFF
--- a/src/tui/components/preview.rs
+++ b/src/tui/components/preview.rs
@@ -609,14 +609,13 @@ fn shorten_path(path: &str) -> String {
 mod tests {
     use super::*;
 
-    /// Each case reads `$HOME` twice, once here and once inside `shorten_path`,
-    /// so a concurrent writer between them makes the two disagree. `isolate_home`
-    /// holds the process-global env lock for the guard's lifetime and restores
-    /// `$HOME` on Drop, before the tempdir is deleted.
+    /// Pins `$HOME` so the read inside `shorten_path` cannot see a value another
+    /// test set. `isolate_home` holds the process-global env lock for the guard's
+    /// lifetime and restores `$HOME` on Drop, before the tempdir is deleted.
     ///
-    /// Pinning `$HOME` also gives these cases something to assert: each was
-    /// wrapped in `if let Some(home)`, so an unset `$HOME` passed all four
-    /// without running an assertion.
+    /// Pinning also gives these cases something to assert: each was wrapped in
+    /// `if let Some(home)`, so an unset `$HOME` passed all four without running
+    /// an assertion.
     #[test]
     #[serial_test::serial]
     fn shorten_path_abbreviates_home() {
--- a/src/tui/components/preview.rs
+++ b/src/tui/components/preview.rs
@@ -609,18 +609,19 @@ fn shorten_path(path: &str) -> String {
 mod tests {
     use super::*;
 
-    /// `#[serial]` because these read `$HOME`, which the suite's many
-    /// `set_var("HOME", ...)` tests mutate. `#[serial]` only excludes other
-    /// `#[serial]` tests, so a parallel reader races every one of those writers:
-    /// the test reads `$HOME` once and `shorten_path` reads it again, and a write
-    /// landing between the two makes them disagree. Pinning `$HOME` also gives
-    /// these cases something to assert; previously an unset `$HOME` made all four
-    /// pass without running an assertion.
+    /// Each case reads `$HOME` twice, once here and once inside `shorten_path`,
+    /// so a concurrent writer between them makes the two disagree. `isolate_home`
+    /// holds the process-global env lock for the guard's lifetime and restores
+    /// `$HOME` on Drop, before the tempdir is deleted.
+    ///
+    /// Pinning `$HOME` also gives these cases something to assert: each was
+    /// wrapped in `if let Some(home)`, so an unset `$HOME` passed all four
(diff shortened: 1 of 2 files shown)


src/tui/components/preview.rs
```diff
@@ -609,14 +609,40 @@ fn shorten_path(path: &str) -> String {
 mod tests {
     use super::*;
 
+    /// Pins `$HOME` so the read inside `shorten_path` cannot see a value another
+    /// test set. `isolate_home` holds the process-global env lock for the guard's
+    /// lifetime and restores `$HOME` on Drop, before the tempdir is deleted.
+    ///
+    /// Pinning also gives these cases something to assert: each was wrapped in
```
jerome-benoit: [P3] Please remove this historical paragraph. On the supported Unix platforms, `dirs::home_dir()` falls back to the passwd database when `HOME` is unset, so an unset variable alone did not make these tests skip every assertion. The claim is therefore inaccurate, and the repository rules also ask test comments not to preserve implementation history. This is non-blocking.

jerome-benoit (approved):
Approved at exact head `39fef8618a2122cea818832217b9c9e699bd53d5`. The test-only change uses the repository's canonical `isolate_home` guard, keeps it alive through the assertions, restores the environment before the temporary directory is dropped, and preserves all four `shorten_path` cases. Exact-head CI is green.

CodeRabbit's similar-prefix finding is technically valid product behavior, but it is pre-existing and the same expectation already existed before this PR; it should not block this HOME-race fix. I left one non-blocking P3 comment about newly added explanatory text.

PR #3762, opened 2026-09-05
feat(web): render the agent live view from the VT grid with whole frames and row patches

njbrake:
## Description

The agent terminal polled `tmux capture-pane` every 50ms and forked `send-keys` per keystroke. Claude Code's fullscreen renderer brackets each repaint in DEC 2026 synchronized output, so the sampler caught half-drawn screens, and the client had grown heuristics to hide the tearing.

The agent surface now renders from the shared VT grid, publishing when the grid changes and withholding frames while a repaint is in flight. Only changed rows go on the wire after the first frame. On the client, the alternate screen renders as a fixed grid, forwarded scroll steps are sized to the backlog, and pinch commits one resize at gesture end.

Paired shells stay on capture-pane snapshots, which is where the #3315 prompt glitch appeared.

### Server

- The agent route acquires the shared `VtChannel` again and publishes on its change signal instead of a 50ms timer. Input takes the socket, so cursor keys are encoded for the pane's DECCKM state rather than passed through `send-keys` verbatim.
- The channel's reader tracks DEC 2026 brackets and holds viewer wakeups until a repaint completes, with a 200ms cap so an app that never closes a bracket cannot freeze a viewer. A freshly seeded grid holds its first publish until output settles.
- Clients advertising `caps.patch` receive row patches carrying the scroll shift and the changed rows; `resync` recovers a lost baseline. Frames still at the pre-resize geometry are withheld briefly after an owner resize.
- A split window or a window deeper than the grid's scrollback falls back to the composited capture path.

### Client

- Rows are patched in place. The alternate screen renders every grid row, so a full-screen agent's layout stops settling after each repaint.
- Forwarded touch scrolling is paced to frame arrivals, with each release sized to how far the finger has outrun the app: a slow drag moves a line at a time with no wait, a fast one clears in larger steps.
- `?livedebug=1` reports frame rate, arrival-to-paint latency, and the share of updates that arrived as patches.

### Side effect worth noting

The TUI's live preview shares the same channel, so it also stops sampling mid-repaint. Its own capture cadence is unchanged.

## PR Type

- [x] New Feature
- [x] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

`tests/live/live-grid-transport.spec.ts` drives a real `aoe serve` against a synthetic full-screen app that brackets its repaints, samples the rendered grid 150 times, and asserts no sample shows a half-drawn frame; a second case asserts a streaming agent arrives as row patches with no resyncs. Vitest covers the patch protocol, the alt-screen grid, and the scroll pacing.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the transport change is covered by Rust unit tests in `src/tmux/vt.rs` and `src/server/live_ws.rs` plus the live Playwright spec above. The TUI's preview inherits the synchronized-output hold through the shared channel, and its own render path is untouched.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Fable 5.1 and Claude Opus 5, via Claude Code

**Any Additional AI Details you'd like to share:** The design was reached by reading the existing transport, comparing against herdr's server-owned emulator, and measuring the tmux fork cost. The synchronized-output tearing was reproduced against tmux 3.6 before the fix. Scroll pacing was tuned against feedback from a real device.

**NOTE:**

_Note: this PR description was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

https://claude.ai/code/session_01He5usTXecgyWa1WrD87j6Z

## Summary

- Added VT-grid rendering for the web agent live view.
- Added synchronized-output handling to prevent partial repaints.
- Added row patches, resynchronization, resize settling, and snapshot fallbacks.
- Improved mobile scrolling, pinch resizing, alternate-screen rendering, and live diagnostics.
- Added Rust, Vitest, and Playwright coverage for the new transport and rendering behavior.
- Updated documentation and the coverage matrix.

## Benefits

The live view uses less data and renders terminal updates more smoothly. It avoids showing incomplete output during synchronized repaints. Unsupported and paired-shell environments continue to use reliable snapshot fallbacks.


src/tmux/vt.rs
```diff
@@ -1234,10 +1364,20 @@ fn run_reader(listener: UnixListener, ctx: ReaderCtx) {
...
         match conn.read(&mut buf) {
             Ok(0) => break,
             Ok(n) => {
+                // Track the app's synchronized-output bracket before anything
+                // can publish this chunk: a frame is published when the
+                // bracket closes (or the hold expires), never in the middle.
+                let sync_event = sync.feed(&buf[..n]);
+                match sync_event {
+                    Some(true) => ctx.signals.begin_hold(),
+                    Some(false) => ctx.signals.end_hold(),
```
jerome-benoit: [P1] Keep the synchronized-output hold active until this closing chunk has been applied to the parser. `end_hold()` currently runs before acquiring `ctx.parser`, so a concurrent sampler can observe the released hold, acquire the parser first, and publish the partial grid left by the preceding chunk. Commit the closing transition together with the corresponding grid update so sampling observes them consistently. The failing live whole-frame test independently confirms that the end-to-end invariant is not met, although it does not prove this race is the only cause.

src/server/live_ws.rs
```diff
@@ -1031,7 +1409,98 @@ async fn handle_live_ws(
...
+        CAPTURE_INTERVAL_IDLE_MS
+    };
+    let since = sample_started.elapsed();
+    let floor = Duration::from_millis(FRAME_MIN_INTERVAL_MS);
+    if since < floor {
+        tokio::time::sleep(floor - since).await;
+    }
+    #[cfg(unix)]
+    {
+        let grid_arm = grid_driven && small_window;
```
jerome-benoit: [P2] Honor `settings.fast` when arming grid-change wakeups. The client sends `fast: false` when the surface is inactive, hidden, or reading scrollback, but a screen-sized VT viewer still wakes on every grid update here. Hidden mounted terminals therefore continue sampling, serializing, and transmitting at the frame floor instead of the idle cadence. Gate grid wakeups on the requested cadence while retaining control-message nudges.

src/server/live_ws.rs
```diff
@@ -1,12 +1,16 @@
...
+//! Live terminal view for the web dashboard.
 //!
-//! Mirrors the TUI's live-send architecture instead of the PTY attach
-//! relay: the server polls `tmux capture-pane` (cursor folded into the
-//! same fork) and pushes ANSI snapshot frames over the WebSocket;
-//! browser input comes back as raw bytes and is delivered via
-//! `tmux send-keys -H`. No PTY, no `tmux attach`, no SIGSTOP pause:
-//! scrollback is just a bigger capture window the client renders and
-//! scrolls natively, and the agent keeps running while the user reads.
+//! The agent surface renders from the shared [`crate::tmux::vt::VtChannel`]
```
jerome-benoit: [P1] This public module documentation links to the crate-private `VtChannel`, and the required Docs job fails here under `rustdoc::private_intra_doc_links`. Use a plain code span or a public documentation target; do not broaden the type's visibility solely to satisfy this comment.

jerome-benoit (changes requested):
Reviewed at exact head `8a625a3fd15c4fcf1890396865f7d6586fc73dd9`, with two independent blind analyses followed by code-level arbitration.

The live end-to-end test currently falsifies this PR's central whole-frame invariant: `live-grid-transport.spec.ts` observed partial `FRAME-A` states without their matching `FRAME-B` rows on the real tmux path. The new synchronization race called out inline makes such partial publication possible; the CI failure does not prove it is the only cause, so the live path still needs to be traced until the regression passes. The required Docs job also fails at the inline rustdoc link.

The existing CodeRabbit pane-count finding remains valid and should be addressed; I have not duplicated it inline: https://github.com/agent-of-empires/agent-of-empires/pull/3762#discussion_r3940954590

I do not consider the CodeQL `new Array(k)` alert a demonstrated resource-exhaustion defect: `k` is bounded by the previous frame length and the server bounds the window. That alert should be dismissed with the corresponding rationale rather than driving an unrelated code change.

No local build or heavy test was run. CI at this exact head is green for formatting, lint, Rust tests, Vitest and mocked Playwright; Docs and live Playwright are failing.

PR #3778, opened 2026-09-05
fix(vt): three review follow-ups on the VT grid live view (#3770-#3772)

### LINKED ISSUE #3770: Prevent partial VT frames when the sample cache misses during synchronized output
## Problem

The VT sampler can return a partially applied synchronized-output frame when its single complete-frame cache does not match the requested window.

At `a5615e8a`, `sample_with_deadline()` reads `frame_incomplete()` while holding the parser lock, but it reuses the cached complete frame only when `(max_lines, cols, rows)` matches. On a cache miss it serializes the current parser grid even when `incomplete` is true. The WebSocket caller later checks `sync_hold_active()` independently: that gate expires after 200 ms while `frame_incomplete()` remains active for 2 seconds, and it can also observe a bracket that closed after sampling. In either case, a payload assembled from a half-applied repaint can be published.

Relevant code:

- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/tmux/vt.rs#L2031-L2078
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/server/live_ws.rs#L1029-L1045

## Proposed deterministic regression case

1. Seed a `VtChannel` and populate its complete-frame cache for window A.
2. Open a DEC 2026 synchronized-output bracket and apply only the first part of a repaint.
3. Sample window B so the cache key does not match.
4. With a controllable test clock, advance beyond `SYNC_HOLD_MAX_MS` while remaining within `SYNC_BRACKET_ABANDON_MS`. In a separate case, use a test-only barrier between sampling and the publication decision to close the bracket at that exact interleaving.
5. Assert that the half-applied payload is never returned as publishable or sent to the client.

The same case should cover two viewers requesting different `window_lines` values. Extract the clock-dependent publication decision if necessary; the regression must not rely on wall-clock sleeps or scheduler timing.

## Expected behavior

No frame assembled while `frame_incomplete()` is true may be published. Sampling and the decision that its result is publishable must describe the same parser state.

## Suggested direction

Return completeness or publishability with the sample while still under the parser lock. On an incomplete cache miss, defer publication or use a compatible last-complete representation instead of serializing the current grid. Do not infer the sampled payload's completeness from a later clock/state check.

## Context

Found during the post-merge audit of #3762. The referenced code is unchanged in merged `main` at `eb0f3829db5e247e93fcbd2d8ec7a88d29d44ad2`. No open issue was found covering this cache-key/publication race; #3737 concerns unread pipe bytes being replayed during reseeding and is a different mechanism.

Investigation and issue drafting were performed with Codex under @jerome-benoit's direction.

### LINKED ISSUE #3771: Track DEC 2026 transitions across VT seeding and back-to-back brackets
## Problem

The VT reader can carry a stale DEC 2026 hold timestamp across discarded pre-seed output or across two brackets represented in one socket read. A later repaint can then outlive that stale timestamp immediately and become publishable before it is complete.

At `a5615e8a`:

- An opener calls `begin_hold()` before the seed check, but a closer received in a later pre-seed read is discarded by the `!seeded` branch before `end_hold()` runs.
- `SyncOutputScanner::feed()` returns only the last transition in a chunk. A chunk containing close then open reports only the opener, while `begin_hold()` deliberately preserves any existing nonzero timestamp.

Relevant code:

- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/tmux/vt.rs#L218-L246
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/tmux/vt.rs#L285-L306
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/tmux/vt.rs#L1400-L1465

### Pre-seed close

1. Deliver an opener in one pre-seed read.
2. Deliver its closer in a second pre-seed read.
3. Complete seeding and set `sync_hold_since_ms` to an already-expired value through the module-test seam.
4. Open a new bracket and apply only the first part of its repaint.
5. Assert that the new bracket receives a fresh hold and that no partial frame is publishable.

### Close then open in one chunk

1. Open a bracket and set its hold timestamp to an observably old value through the module-test seam.
2. Deliver the previous close and the next open in one chunk.
3. Assert that the new bracket starts with a fresh timestamp rather than inheriting the previous bracket's age. Do not use a wall-clock sleep for either case.

## Expected behavior

Each real bracket has its own hold lifetime. Closing transitions must clear the corresponding hold even while pre-seed payload bytes are intentionally discarded, and a close→open sequence must end the old hold before starting the new one.

## Suggested direction

Preserve ordered synchronized-output transitions from each scanned chunk instead of only the last boolean. Apply close/open state transitions in order, while continuing to delay the closing publication signal until the associated bytes have been applied to the parser.

## Context

Found during the post-merge audit of #3762. The referenced code is unchanged in merged `main` at `eb0f3829db5e247e93fcbd2d8ec7a88d29d44ad2`. The pre-seed case was also reported in an unresolved review thread: https://github.com/agent-of-empires/agent-of-empires/pull/3762#discussion_r3941232205. No open issue was found covering it.

Investigation and issue drafting were performed with Codex under @jerome-benoit's direction.

### LINKED ISSUE #3772: Keep VT grid transport suspended until resize reseeding succeeds
## Problem

The live WebSocket path resumes VT-grid publication after a successful tmux resize even when reseeding the VT parser to the new geometry returned `Busy` or `Failed`.

At `a5615e8a`, the direct resize, ownership takeover, and drift reassertion paths discard `set_grid_size_with_deadline()`'s `VtRefreshResult`. They then arm or complete the 300 ms resize-settle window based only on tmux ownership/resize success. `Busy` and `Failed` leave the channel's stored rows and columns unchanged, so the old grid becomes publishable again when that fixed settle interval expires. Periodic reconciliation may repair it later, but it does not prevent the stale-geometry frame from being sent and can itself continue returning `Busy` during sustained output.

Relevant code:

- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/server/live_ws.rs#L940-L961
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/server/live_ws.rs#L1277-L1297
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/server/live_ws.rs#L1348-L1374
- https://github.com/agent-of-empires/agent-of-empires/blob/a5615e8ae38576359d817139a5d576c80ea91472/src/tmux/vt.rs#L2165-L2192

## Proposed deterministic regression case

1. Start grid transport at geometry A.
2. Through an injected refresh operation or an extracted pure resize-result policy, make the tmux resize to geometry B succeed while the VT refresh returns `Busy`.
3. Advance beyond `RESIZE_SETTLE_MS` without a successful reseed.
4. Assert that grid transport remains ineligible and that no stale frame at geometry A is published after tmux has resized to geometry B.
5. Make a later reseed return `Refreshed` and assert that grid transport resumes at geometry B.

Cover direct resize, ownership takeover, and drift reassertion because all three currently discard the refresh result. The test must use an explicit seam rather than depending on tmux timing to produce `Busy`.

## Expected behavior

After tmux changes geometry, VT-grid transport must remain suspended until the parser has been refreshed to that geometry. A bounded snapshot fallback is preferable to publishing a stale grid.

## Suggested direction

Propagate `VtRefreshResult` from the blocking resize operations. Arm or clear the settle state only after `Refreshed`; on `Busy` or `Failed`, keep grid publication disabled and use snapshot transport or a bounded retry until the channel is synchronized.

## Context

Found during the post-merge audit of #3762. The referenced code is unchanged in merged `main` at `eb0f3829db5e247e93fcbd2d8ec7a88d29d44ad2`. No open issue was found covering ignored VT refresh outcomes after resize.

Investigation and issue drafting were performed with Codex under @jerome-benoit's direction.

njbrake:
## Description

Three post-merge review follow-ups on #3762, one commit each. They share the VT
grid transport, so they land together.

**#3770, partial frames on a sample cache miss.** The sampler read
`frame_incomplete()` under the parser lock but reused its single complete-frame
cache only for a matching window; on a miss mid-bracket it serialized the
half-drawn grid, and the WebSocket decided publishability afterwards from
`sync_hold_active()` (expires at 200ms while a frame stays incomplete for 2s,
and can observe a bracket that closed after the sample). Both samplers now
return their publishability, decided under the lock that assembled the payload:
`VtSample` for the scrollback window, `VtRowsSample` for the composite rows.
The web loop holds such a sample whatever the hold says now; the TUI preview
skips it, whole-window and composited alike. Publishing an agent's OSC 52 copy
moved out of the frame block, since the read that fetches it is consuming and it
must reach the host clipboard whether or not this cycle has a frame.

**#3771, DEC 2026 transitions across seeding and back-to-back brackets.** The
scanner reported only the last transition per read, and the pre-seed branch
discarded a chunk before its closer could release the hold. `feed` now appends
transitions in order and `SyncHoldPlan` maps a chunk's transitions to the hold
moves around applying it. The bracket hold and the incomplete run are separate
deadlines: a close that the same read reopens over restarts the bracket, but not
the run, because that frame was never in the grid on its own for a viewer to
sample. Refreshing both would let an app whose repaints straddle every socket
read extend the abandon window forever and freeze the view.

**#3772, grid transport after a failed resize reseed.** Direct resize, ownership
takeover, drift re-assert and auto-reclaim (which the issue did not list,
because it never reseeded the grid at all) discarded `VtRefreshResult` and armed
the settle window on tmux ownership alone, so `Busy`/`Failed` republished the
old geometry once the window expired. The channel is now told the geometry
before tmux is asked for it and holds that expectation until the parser is
rebuilt at it. It lives on the channel because the parser is shared: a viewer
that did not drive the resize renders the same stale cells. Frames are withheld
while it stands rather than moved to another transport.

Whether the expectation is still owed is settled against tmux rather than a
clock. `reconcile_grid` re-reads the pane and either drops one the pane never
took (a resize tmux refused or clamped, which no reseed can satisfy) or re-aims
a real divergence at the geometry tmux does report; it never opens one, since
ordinary drift is what the reseed in that same pass is for. A probe that
straddles a resize cannot retire it, because what it read may describe either
side of it. The withhold stops the loop sampling, so the capture loop drives
that reconcile itself and the size owner re-drives the reseed on a throttle.

Nothing on the list turned out wrong, fixed, or in conflict.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

Rust unit tests, no wall-clock sleeps for the timing cases:

- `sample_reports_a_mid_bracket_cache_miss_as_incomplete`: two viewers on
  different windows, and a bracket closing after the sample.
- `sample_rows_padded_renders_the_visible_grid_at_the_requested_rectangle`:
  extended for the composite path's mid-bracket case.
- `sync_hold_plan_gives_each_bracket_its_own_lifetime` and the extended scanner
  test: ordered transitions, close-then-open in one chunk, fresh timestamp with
  no released window between brackets.
- `repeated_close_open_reads_cannot_freeze_the_view`: fifty straddling reads
  never extend the abandon window, and a clean close still ends the run.
- `reader_releases_a_bracket_closed_before_the_grid_is_seeded`: the pre-seed
  close, through the real reader loop.
- `a_restart_over_a_settled_grid_starts_the_incomplete_run`: a read that opens,
  closes and reopens over a grid that was not mid-repaint.
- `a_resize_holds_every_viewer_off_the_grid_until_the_parser_catches_up`: the
  expectation declared before the resize, cleared only by reaching the geometry,
  withdrawn when the resize was not ours, dropped when tmux shows the pane never
  took it, held across ten failed reseeds while the pane really does differ, and
  never opened by reconcile on its own.
- `a_geometry_probe_that_straddles_a_resize_cannot_retire_it`: the two-viewer
  interleaving, plus the mid-resize and quiescent probes.
- `a_resize_whose_reseed_missed_withholds_frames_until_it_lands` and
  `grid_transport_needs_a_single_pane_within_the_grids_scrollback`: the settle
  window and transport eligibility.

Two assertions the issues asked for are not tested as written. The
sample/publish interleaving needs no barrier once publishability travels with
the payload, so the test asserts the flag survives the bracket closing instead.
The hold expiring while a frame stays incomplete is covered by
`viewer_signals_hold_opens_and_closes`, which injects the clock. The four resize
call sites are covered at the channel state they all record into, not
individually: they are inline in one long async block with no seam to drive them
from a test.

- [x] Skipped a test, because: `live/live-grid-transport.spec.ts` already covers
  this flow end to end and is the spec that caught the transport regression in
  the previous push; it passes here along with the rest of the live suite.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the paths need a real agent repainting inside a
  synchronized-output bracket and a tmux resize racing a reseed, neither of
  which an e2e harness can stage deterministically.

`cargo fmt`, `cargo clippy --all-targets` and `cargo test` are clean, and each
commit builds and tests green on its own. The live Playwright suite is 160
passed / 2 failed locally: `triage-snooze` passes on re-run, and
`acp-view-switch` "switch to terminal resumes the claude conversation" fails the
same way on a clean `main` worktree here, so neither is from this branch.
On the live suite I measured this branch against `main` rather than eyeballing
one run, because an earlier revision of this work made
`live-size-owner-takeover` "ownership ping-pong" flaky: run together with
`live-grid-transport` and `mobile-live-no-flutter` it failed about 4 runs in 14
while `main` failed 0 in 26. It was the runtime transport switch, not a logic
error: the cursor overlay lands rows off its prompt when the view moves between
two serializations of the same screen. Withholding frames instead of switching
transports takes it to 1 in 14, which I cannot separate from `main` at that
sample size but have not proven to be zero either. The spec passes 6 of 6 on its
own; it only shows up under the three-spec load.
`acp::supervisor::tests::shutdown_and_wait_degrades_when_load_errs_without_peer`
and `process::worker_registry::tests::pid_source_for_falls_back_to_peer_pid_on_load_err`
also fail on `main` in my sandbox: both chmod a file to 0o000 and assert the
read fails, which uid 0 bypasses.

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Issues read and fixed by the
model under my direction, then revised against this PR's review. I reviewed the
diff.

Fixes #3770
Fixes #3771
Fixes #3772

## Summary

- Prevents incomplete VT frames from reaching viewers during synchronized output.
- Tracks DEC 2026 transitions across pre-seed reads and consecutive brackets.
- Suspends grid transport after resize until VT reseeding succeeds.
- Uses bounded `capture-pane` polling while the grid catches up.
- Adds documentation and Rust tests for sampling, synchronization, resize recovery, and transport eligibility.

## Benefits

Viewers no longer receive partial or stale frames after synchronized updates or geometry changes. Live view recovery is more reliable when VT reseeding is delayed or fails.

### DIFF
--- a/src/tmux/vt.rs
+++ b/src/tmux/vt.rs
@@ -1587,6 +1587,37 @@ struct SampleCache {
     cursor: PaneCursor,
 }
 
+/// One [`VtChannel::sample`] result and whether it may be published.
+pub(crate) struct VtSample {
+    pub(crate) content: String,
+    pub(crate) cursor: Option<PaneCursor>,
+    /// True when `content` was serialized from a grid inside an unclosed
+    /// synchronized-output bracket, i.e. a half-drawn frame. Decided under the
+    /// same parser lock that assembled `content`, so a caller's publish
+    /// decision describes the state the payload came from; a later
+    /// [`VtChannel::sync_hold_active`] call can see an expired hold or an
+    /// entirely different bracket.
+    pub(crate) incomplete: bool,
+}
+
+impl VtSample {
+    fn whole(content: String, cursor: Option<PaneCursor>) -> Self {
+        Self {
+            content,
+            cursor,
+            incomplete: false,
+        }
+    }
+}
+
+/// One [`VtChannel::sample_rows_padded_with_deadline`] result: the visible grid
+/// as display rows, plus the same publishability [`VtSample`] carries.
+pub(crate) struct VtRowsSample {
+    pub(crate) rows: Vec<String>,
+    pub(crate) cursor: PaneCursor,
+    pub(crate) incomplete: bool,
+}
+
 impl VtChannel {
     /// Get the shared channel for `session`, arming a new one if none is live.
     /// Returns `None` if tmux is too old or the pane is gone or any tmux/socket
@@ -2023,7 +2054,7 @@ impl VtChannel {
     /// the TUI scroll and the web's virtual scroll spacer need real history
     /// here, not just the visible screen.
     #[cfg(test)]
-    pub(crate) fn sample(&self, max_lines: usize) -> (String, Option<PaneCursor>) {
+    pub(crate) fn sample(&self, max_lines: usize) -> VtSample {
         let deadline = crate::tmux::TmuxCommandDeadline::new();
         self.sample_with_deadline(max_lines, &deadline)
     }
@@ -2032,7 +2063,7 @@ impl VtChannel {
         &self,
         max_lines: usize,
         deadline: &crate::tmux::TmuxCommandDeadline,
-    ) -> (String, Option<PaneCursor>) {
+    ) -> VtSample {
         // Both fork tmux and take the parser lock themselves, so they run
         // before this sampler takes it.
         self.reconcile_grid(deadline);
@@ -2041,7 +2072,7 @@ impl VtChannel {
         let rows = self.rows.load(Ordering::Relaxed);
         let mut p = match self.parser.lock() {
             Ok(p) => p,
-            Err(_) => return (String::new(), None),
+            Err(_) => return VtSample::whole(String::new(), None),
         };
         // Read both under the parser lock, which is where the reader applies a
         // chunk and bumps the generation, and where it releases a bracket. The
@@ -2055,7 +2086,7 @@ impl VtChannel {
                 // Mid-bracket the grid is a half-drawn frame: serve the last
                 // complete one instead. The reader wakes viewers on close.
                 if same_window && (c.grid_gen == grid_gen || incomplete) {
-                    return (c.content.clone(), Some(c.cursor));
+                    return VtSample::whole(c.content.clone(), Some(c.cursor));
                 }
             }
         }
@@ -2077,7 +2108,11 @@ impl VtChannel {
                 });
             }
         }
-        (content, Some(cursor))
+        VtSample {
+            content,
+            cursor: Some(cursor),
+            incomplete,
+        }
     }
 
     /// Sample the VISIBLE grid as `want_rows` rows padded to `want_cols`
@@ -2096,7 +2131,7 @@ impl VtChannel {
         &self,
         want_cols: u16,
         want_rows: u16,
-    ) -> Option<(Vec<String>, PaneCursor)> {
+    ) -> Option<VtRowsSample> {
         let deadline = crate::tmux::TmuxCommandDeadline::new();
         self.sample_rows_padded_with_deadline(want_cols, want_rows, &deadline)
     }
@@ -2106,7 +2141,7 @@ impl VtChannel {
         want_cols: u16,
         want_rows: u16,
         deadline: &crate::tmux::TmuxCommandDeadline,
-    ) -> Option<(Vec<String>, PaneCursor)> {
+    ) -> Option<VtRowsSample> {
         self.reconcile_grid(deadline);
         self.refresh_owner_heartbeat(deadline);
         let cols = self.cols.load(Ordering::Relaxed);
@@ -2115,6 +2150,10 @@ impl VtChannel {
         let want_rows = want_rows.max(1);
 
         let p = self.parser.lock().ok()?;
+        // Read under the lock that renders these rows, like the scrollback
+        // sampler: a composite spliced from a half-drawn pane 0 tears the same
+        // way a whole-window frame does.
+        let incomplete = self.signals.frame_incomplete();
         let screen = p.screen();
         let readable_cols = cols.min(want_cols);
         let out = (0..want_rows)
@@ -2135,7 +2174,11 @@ impl VtChannel {
             .collect();
         let cursor = cursor_from_screen(screen, rows, cols);
         drop(p);
-        Some((out, cursor))
+        Some(VtRowsSample {
+            rows: out,
+            cursor,
+            incomplete,
+        })
     }
 
     /// A receiver that fires on every publishable grid change, OSC 52 write, and
@@ -3093,7 +3136,9 @@ mod tests {
             .process(b"hello\r\nworld\r\n\x1b[41mfilled");
 
         // Exact rectangle.
-        let (rows, cursor) = ch.sample_rows_padded(20, 4).expect("sample");
+        let sample = ch.sample_rows_padded(20, 4).expect("sample");
+        let (rows, cursor) = (sample.rows, sample.cursor);
+        assert!(!sample.incomplete, "no bracket open: publishable");
         assert_eq!(rows.len(), 4);
         for (i, r) in rows.iter().enumerate() {
             assert_eq!(
@@ -3108,7 +3153,7 @@ mod tests {
         assert!(cursor.position_reliable);
 
         // Narrower and shorter than the grid: truncate, never overflow.
-        let (rows, _) = ch.sample_rows_padded(6, 2).expect("sample");
+        let rows = ch.sample_rows_padded(6, 2).expect("sample").rows;
         assert_eq!(rows.len(), 2);
         for r in &rows {
             assert_eq!(crate::tmux::utils::strip_ansi(r).chars().count(), 6);
@@ -3117,7 +3162,7 @@ mod tests {
         // Taller than the grid (tmux says the pane grew before the grid caught
         // up): the extra rows are blank filler at the right width, not rows
         // borrowed from elsewhere.
-        let (rows, _) = ch.sample_rows_padded(10, 6).expect("sample");
+        let rows = ch.sample_rows_padded(10, 6).expect("sample").rows;
         assert_eq!(rows.len(), 6);
         for (i, r) in rows.iter().enumerate() {
             let plain = crate::tmux::utils::strip_ansi(r);
@@ -3126,6 +3171,15 @@ mod tests {
                 assert!(plain.trim().is_empty(), "row {i} should be filler: {r:?}");
             }
         }
+
+        // Mid-bracket the rows are a half-drawn repaint. A composite splices
+        // them into the window next to panes captured whole, so the sample says
+        // so and the preview keeps the frame it has.
+        ch.signals.begin_hold();
+        let held = ch.sample_rows_padded(20, 4).expect("sample");
+        assert!(held.incomplete, "mid-bracket rows are not publishable");
+        ch.signals.end_hold();
+        assert!(!ch.sample_rows_padded(20, 4).expect("sample").incomplete);
     }
 
     #[test]
@@ -3204,13 +3258,13 @@ mod tests {
 
         ch.parser.lock().unwrap().process(b"one");
         ch.grid_gen.fetch_add(1, Ordering::Relaxed);
-        let (first, _) = ch.sample(4);
+        let first = ch.sample(4).content;
         assert!(first.contains("one"), "fresh assembly:\n{first:?}");
 
         // Advance the parser WITHOUT bumping gen: the cache must still serve
         // the old frame (this is what makes an idle pane's cadence cheap).
         ch.parser.lock().unwrap().process(b" two");
-        let (cached, _) = ch.sample(4);
+        let cached = ch.sample(4).content;
         assert!(
             !cached.contains("two"),
             "same generation must serve the cached assembly:\n{cached:?}"
@@ -3218,14 +3272,14 @@ mod tests {
 
         // Bump gen (what the reader does per chunk): fresh assembly.
         ch.grid_gen.fetch_add(1, Ordering::Relaxed);
-        let (fresh, _) = ch.sample(4);
+        let fresh = ch.sample(4).content;
         assert!(
             fresh.contains("two"),
             "bumped generation must reassemble:\n{fresh:?}"
         );
 
         // A different window size also misses the cache.
-        let (wider, _) = ch.sample(3);
+        let wider = ch.sample(3).content;
         assert!(wider.contains("two"), "window change must reassemble");
     }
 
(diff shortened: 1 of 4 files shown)


src/tmux/vt.rs
```diff
@@ -235,15 +236,61 @@ impl SyncOutputScanner {
...
+    /// The chunk ends outside any bracket.
+    close: bool,
+}
+
+impl SyncHoldPlan {
+    fn from_events(events: &[bool]) -> Self {
+        let last_open = events.iter().rposition(|&open| open);
+        Self {
+            open: last_open.is_some(),
+            restart: last_open.is_some_and(|i| events[..i].contains(&false)),
```
jerome-benoit, line 276: [P1] Preserve completed frames across close/open chunks

For every read containing `close(A), open(B)`, this sets `restart = true` while `close` remains false. If a continuously repainting app repeatedly delivers `tail(A) + close(A) + open(B) + head(B)` in one socket read, `restart_hold()` refreshes the incomplete-frame deadline on every read, `end()` never releases it, and the reader never notifies viewers. `sample_with_deadline()` therefore keeps serving the previous complete cache indefinitely even though A, B, and later frames each completed. The base eventually escaped through the existing abandon deadline because it did not refresh the timestamp; this change turns that bounded stale period into an unbounded frozen view. Preserve/cache the completed grid at each close boundary before applying the following opener, and add a deterministic repeated close/open test that asserts publication progresses.

jerome-benoit (changes requested):
Reviewed exact head `a75049f0e68b4505a231320e67e93b10757da474` through two independent blind analyses and code-level cross-validation. One new P1 is inline. I also independently confirmed CodeRabbit's stale-owner retry finding: the retry can reseed the shared channel to an obsolete geometry, so that existing thread is blocking and I am not duplicating it. The composite-sampling gap is valid but predates this branch; the per-connection resync gap is explicitly disclosed as a follow-up limitation. The provisional-gate window also predates the persistent stale-grid fix. Those should remain tracked, but they are not additional regressions introduced here. Exact-head Rust, lint, format, and CodeQL checks pass; Playwright live is still red because `live-grid-transport.spec.ts` expected `grid` and received `snapshot`. That failure also needs resolution or a clean rerun before merge.

src/tmux/vt.rs
```diff
@@ -283,35 +348,72 @@ impl ViewerSignals {
...
+    /// the previous bracket's close. One store, so no sampler observes a gap
+    /// where the previous repaint's still half-drawn grid reads as whole.
+    ///
+    /// The incomplete run deliberately keeps running: the frame that closed
+    /// mid-read was never in the grid on its own (this same read already
+    /// applied the next repaint's opening bytes over it), so counting it as
+    /// shown would let a continuously repainting app hold the view forever.
+    fn restart_hold(&self) {
+        self.sync_hold_since_ms
+            .store(chunk_now_ms().max(1), Ordering::Relaxed);
```
jerome-benoit: [P2] Initialize an absent incomplete run when restarting

Starting outside a bracket, a single read containing `[open, close, open]` selects `restart_hold()` (`SyncHoldPlan::from_events`, lines 279–294). This writes only the bracket timer, leaving `incomplete_since_ms == 0`. Both `frame_incomplete()` and `hold_active()` consequently return false, so the reader notifies viewers and samplers mark the unfinished second repaint as publishable. The base's final opener initialized its hold. Initialize the incomplete-run timestamp here when it is absent, while preserving an existing run's timestamp to avoid restoring the previous freeze. Add a deterministic initially-closed `[true, false, true]` case alongside the existing repeated close/open test, which starts with `begin_hold()` and misses this condition. Verified by exact-source tracing, not runtime reproduction.

jerome-benoit (changes requested):
Re-reviewed `a1af8f26140ec790bfbc91a8a4c74ac9cc69540a` with two independent sequential analyses and final arbitration. The previous unbounded freeze is fixed: repeated close/open reads no longer extend the incomplete-run deadline. Ownership revalidation, shared pre-resize gating and composite sample publishability are also present.

Two P2 correctness issues remain:

- New inline: restarting from an initially closed state can leave the incomplete timer unset and publish a partial repaint.
- The existing [resize-expectation finding](https://github.com/agent-of-empires/agent-of-empires/pull/3778#discussion_r3949536301) is valid and blocking, not duplicated. An identical target keeps its original timestamp (`vt.rs:2369–2374`), and expiry clears the gate despite failed reseeds (`2406–2431`). Web eligibility is decided before sampling, so reconciliation during that sample does not prevent the already-selected old-geometry frame from being returned (`live_ws.rs:884–930`; `vt.rs:2194`). Please distinguish an unconfirmed resize request from a tmux-confirmed divergence and keep publication gated while the latter remains unresolved. **Do not simply renew on every retry:** `set_grid_size_with_deadline` also calls `expect_grid_size`, which could then indefinitely retain a size tmux refused. Cover both confirmed divergence with repeated failed reseeds and an unfulfilled request that must expire.

Arbitration: both issues are P2 rendering-correctness defects; no evidence justified the initially proposed P1 for resize expiry. The earlier freeze is not being reclassified as still unfixed.

Validation: exact-head/base source, description and existing discussions read via `gh api`; CI, Tests and CodeQL are green. Findings are source-traced, not runtime reproductions. No local tests were run.

src/tmux/vt.rs
```diff
@@ -2192,6 +2351,92 @@ impl VtChannel {
...
+    /// stays gated for as long as it takes a reseed to land rather than for a
+    /// fixed window that a slow one could outlive.
+    fn observe_pane_geometry(&self, pane: (u16, u16)) {
+        if pane
+            == (
+                self.cols.load(Ordering::Relaxed),
+                self.rows.load(Ordering::Relaxed),
+            )
+        {
+            self.clear_resync_target();
```
jerome-benoit: [P2] Do not retire an in-flight resize from an older geometry probe

The observation comes from `pane_size_cursor` at line 2059, before this unconditional clear. With two viewers, the owner can declare 100x30 (`live_ws.rs:358`), a follower's reconciliation can read the still-current 80x24, then the owner can complete the resize while reseeding returns Busy/Failed. If reconciliation resumes now, the parser is still 80x24, so this clears the valid 100x30 gate. The follower then becomes grid-eligible (`live_ws.rs:891-927`) and can publish the stale layout; it has no owner settle window, and the one-second reconcile throttle prevents an immediate recheck.

Order observations against resize start/completion, for example with an in-flight state and generation, and only withdraw an expectation from a probe known not to straddle a resize. Add a deterministic interleaving test alongside the sequential `observe_pane_geometry` cases. Matching dimensions alone does not prove that the resize was refused. Verified by exact-head `git show` of the two call paths; no runtime reproduction claimed.

jerome-benoit (changes requested):
Re-reviewed `d27d2bb8` with two independent sequential analyses and final arbitration. The initially-closed DEC restart now initializes the incomplete-run timer without extending an existing run. The previous unconditional resize-expiry defect is also removed, and snapshot viewers now drive reconciliation.

One P2 remains in the new reconciliation path: an observation taken before a concurrent resize completes can clear that resize's publication gate. Details inline. This is a bounded stale-layout regression, not the previously fixed unbounded freeze and not a P1.

Evidence: exact-head `git show` and the relevant `git diff a1af8f26 d27d2bb8`; `git diff --check` exits 0. CI, Tests and CodeQL are green at this SHA. The interleaving is source-traced, not a runtime reproduction; no local tests were run.

PR #3779, opened 2026-09-05
fix(acp): answer option-list permission requests with the picked option

### LINKED ISSUE #3741: ACP: option-list permission requests (e.g. pi ask_user_question) render as Allow/Always/Deny; Allow silently picks the first option
## Summary

When an ACP agent sends `session/request_permission` carrying a custom option list, AoE renders the fixed Allow / Always / Deny approval card and never shows the option labels. Resolving with **Allow silently picks the first option**, so the agent receives an answer the user never chose.

Concrete case: pi (`@earendil-works/pi-coding-agent`) via `pi-acp`. pi's `ask_user_question` select dialogs reach AoE as permission requests whose options are the question's choices.

## Steps to reproduce

1. Configure pi as an ACP agent via `pi-acp` 0.0.33.
2. In a session, have the agent call `ask_user_question` with four options (e.g. `Option Alpha`, `Option Bravo`, `Option Charlie`, `Option Delta`).
3. In the AoE web UI the approval card shows only Allow / Always / Deny. No option labels anywhere.
4. Click Allow. The agent receives `Option Alpha` regardless of intent.

## Evidence

Live session, four unrelated multiple-choice questions in a row. Every answer that came back to the agent was exactly the first option of each list. The user never saw the labels.

Screenshot of the card as rendered (the four options were Alpha/Bravo/Charlie/Delta; none are visible):

ask_user_question rendered as Allow/Always/Deny

## Root cause

1. pi-acp maps each select option to a `PermissionOption` with kind `allow_once` (`handleExtensionSelect` in `src/acp/session.ts`). ACP allows option names to carry arbitrary labels, so a choice list arrives looking like N allow-buttons.
2. AoE server drops the option list at ingestion: `Approval` (`src/acp/approvals.rs`) stores only `tool_call` + `destructive`; the options are not persisted or forwarded to the web client.
3. Web `ApprovalCard` (`web/src/components/acp/ApprovalCard.tsx`) renders the fixed Allow / Always / Deny trio for every approval.
4. On Allow, `pick_option_id` (`src/acp/acp_client/permission_handlers.rs`) returns the first option matching `allow_once` — which for these requests is simply `options[0]`.

Net effect: the UI asks a different question than the agent asked, and resolves it with a default.

## Suggested direction

- Carry the request's options on `Approval` (e.g. `options: Vec<{ option_id, name, kind }>`) and record the selected `option_id` in `ResolvedApproval`. New serde field, old clients ignore it.
- In `ApprovalCard`, when a request has multiple same-kind options with custom names, render a single-select list of the option labels instead of the Allow/Always/Deny trio. Keep the trio for genuine allow/deny vocabularies.
- Heuristic for "this is a choice list, not an approval": all options share one kind and count > 2, or the tool call id carries pi-acp's `pi-ui-` prefix.

## Environment

- AoE: current main (worktree `duplicate-thinkg-level-dropdown`, v1.14.x line)
- pi-acp 0.0.33, pi 0.84.2 (`@earendil-works/pi-coding-agent`)
- Linux, web UI in Chrome

---

Disclosure: diagnosed and drafted with AI assistance (pi). The repro, the answer-equals-first-option evidence, and the code references above were verified against the sources linked.

njbrake:
Some coding agents ask you a multiple-choice question by sending it down the same channel they use to ask permission to run a tool. aoe did not notice, so it showed its usual Allow / Always / Deny buttons and never displayed the actual answers. Clicking Allow then sent the agent whichever answer happened to be first on the list, so it looked like you had chosen something when you had not. Now aoe recognises those requests, shows the agent's own answers as buttons, and sends back the one you clicked. The terminal UI gets the same treatment: `a` opens a picker of the answers instead of silently picking the first.

## Description

Reported in #3741 with a reproduction against pi (`pi-acp` 0.0.33), whose `ask_user_question` reaches aoe as a `session/request_permission` carrying one `allow_once` option per answer. Confirmed against `pi-acp`'s published bundle: `handleExtensionSelect` maps every choice to `kind: "allow_once"` and offers no reject option, so `pick_option_id` resolved Allow to `options[0]` every time.

**Detecting it.** `Approval` now carries the options the agent offered, plus a `choice` flag the daemon computes: at least two options, all sharing one kind. A list with no kind to tell its entries apart carries no permission semantics at all, so the names are the whole content. Vendor-neutral, and it covers a two-answer question that a "more than two options" rule would miss.

Checked against published bundles rather than inferred: pi's select sends N `allow_once` (classified), pi's `confirm` sends allow_once + reject_once (trio), `claude-code-acp` always uses distinct kinds (trio), and gemini sends **two** `allow_always` options for an MCP tool ("all server tools" and "this tool") beside allow_once and reject_once (trio). That last one is why the rule is "all share one kind" rather than "a kind repeats": gemini's MCP prompt repeats a kind but is a real approval, and misclassifying it would relabel it a question and record a rejection through it as an allow.

**Dismissing it.** Dismiss resolves as `Cancelled`, not `Deny`. A deny with no `option_id` is resolved by kind, so on an answer list whose options are all reject-kind it would have picked the first one and sent it as the user's answer, which is this PR's own bug pointed the other way. Cancelled takes the resolver's cancellation path and is never mapped onto an option. Picking a reject-kind option explicitly still answers with it and closes the tool card as denied.

**Holding it.** `destructive` and `choice` are independent, so a destructive tool can ask its question through an option list. Every allowing option carries the 800ms hold and its own ring; reject-kind options stay a single tap, matching the trio reserving single-tap for Deny. Only one hold runs at a time: starting one cancels any hold in flight, so two fingers on two options cannot leave the first timer armed with the option the user moved away from, and a generation counter makes a timer that fired just before release submit nothing.

**Counting it.** Telemetry tallies off `Event::ApprovalResolved`, which carries the decision that reached the agent. The endpoint only knows what the client submitted, which for an answered option list is allow-shaped whatever the option means.

**Answering it.** `POST /acp/approvals/{nonce}` takes an optional `option_id`. It is authoritative: an id matching no live option cancels rather than falling back to a kind guess, so a stale card cannot send an answer the user did not pick. Omitting it keeps the old kind-based behaviour, so older clients are unaffected. When the client names an option, the recorded decision is derived from *that option's* kind rather than the allow-shaped decision sent beside it, so a reject-kind pick is broadcast as a Deny and closes the tool card.

Out of scope but worth noting for whoever picks it up: gemini's Always button already picks the first `allow_always`, which is "allow all server tools for this session" rather than "allow this tool". Pre-existing, untouched here.

**Rendering it.** The web card shows one button per option under a Question header, with Dismiss for the deny path. The TUI shelf reuses the existing choice picker: `a` opens it, `d` still denies, and the allow-always hint (shelf and status line) is dropped since it has no meaning here.

`Approval.options` / `Approval.choice` are `#[serde(default)]`, so approvals replayed from an older event log deserialize and render as before. `ResolvedApproval` is unchanged: nothing in production writes it today, and adding a field there would be dead code.

Not covered: `aoe acp approve` still sends kind-based decisions only. It cannot list the options either, so answering a question from the CLI needs more than a flag; the TUI and dashboard both cover it.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

The card is captured by the new live story spec rather than a committed image: `docs/assets/` is generated by `scripts/dev/capture-web-screenshots.sh` and its contract is hero shots only, so a second approval variant does not belong there. The rendered card reads:

```
 (?) QUESTION  Pi select                                    ^
 MESSAGE  Which option?
 [ Option Alpha                                              ]
 [ Option Bravo                                              ]
 [ Option Charlie                                            ]
 [ Option Delta                                              ]
 [ x Dismiss ]
```

## Test Coverage Analysis

`web/tests/live/acp-stories/approval-question.spec.ts`: renders the four labels, asserts Allow / Always are absent, clicks the third, and asserts the agent received `choice-2`. The fake ACP agent gained an opt-in `echoDecision` flag so the option that actually reached the agent is observable in the transcript. `ApprovalCard.test.tsx` covers the card in isolation: the fall back to the trio when the flag arrives with no options, dismissal cancelling on a reject-kind list without selecting one of its options, and a destructive answer list refusing a quick click while resolving on a sustained hold. The destructive cases fail if the hold is removed, and the overlapping-hold case fails without the cancel-before-arm.

**TUI / CLI (e2e test)**

`tests/e2e/acp_option_list_approval_e2e.rs`: real daemon, real permission request, keys through the production input loop; `a` opens the picker, two `j`s and Enter answer, and the agent echoes `choice-2`. `tests/integration/acp_smoke.rs` covers the same round trip without the TUI (it fails with `choice-0` when the fix is reverted) plus a dismissal reaching the agent as a cancellation. The `d`-means-dismiss mapping is unit-tested at `approval_key_outcome`, and the telemetry tally at `record_approval_decision`.

Ran: `cargo fmt`, `cargo clippy --all-targets --features e2e-tests`, `cargo test`, `cargo test --features e2e-tests --test e2e`, and in `web/`: `format:check`, `lint`, `tsc -b`, `test:unit`, plus the live approval specs. Two pre-existing lib tests fail in my container only (`shutdown_and_wait_degrades_when_load_errs_without_peer`, `pid_source_for_falls_back_to_peer_pid_on_load_err`); both assert that `chmod 000` makes a read fail, which does not hold when the test runs as root.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Diagnosis in the issue was verified against `pi-acp`'s published bundle before implementing; the option-kind classification rule and the stale-`option_id` cancel are this PR's, not the issue's suggested heuristic. A fresh-context review pass caught that the first version of the rule ("a kind repeats") would have reclassified gemini's MCP approvals; the bundles of gemini-cli and claude-code-acp were then read directly to settle the boundary.

Fixes #3741

## Summary

- Preserves ACP option lists and identifies question-style requests.
- Shows agent-provided choices in the web UI and TUI.
- Sends the selected option ID to the agent.
- Cancels dismissed, invalid, or stale requests without selecting an option.
- Preserves standard approval behavior and compatibility with older event logs.
- Hardens destructive option selection against stale hold timers.
- Adds web, TUI, integration, and end-to-end coverage.

## Benefits

Users see the actual question and selected answer. ACP agents no longer receive an unintended first option from a generic approval action.

## Potential Follow-up

The CLI approval command remains kind-based and does not support selecting listed options.


web/src/components/acp/ApprovalCard.tsx
```diff
@@ -146,8 +161,29 @@ export function ApprovalCard({ approval, onResolve }: Props) {
...
+              key={option.option_id}
+              type="button"
+              className={[
+                "w-full rounded-md border border-surface-700 bg-surface-800",
+                "px-3 py-2 text-left text-xs font-medium text-text-primary break-words",
+                "hover:border-brand-600/60 hover:bg-brand-700/10",
+                phase === "submitting" ? "cursor-wait opacity-60" : "",
+              ].join(" ")}
+              disabled={offline || phase === "submitting"}
+              onClick={() => void submit("Allow", option.option_id)}
```
jerome-benoit: [P1] Preserve destructive confirmation for choice lists

`destructive` and `choice` are computed independently, so both can be true. In that case this branch submits any option on a single click and the `isChoice` branch below removes the 800 ms hold entirely. A destructive ACP permission encoded as multiple same-kind options therefore bypasses the documented accidental-action safeguard. Keep the explicit `option_id`, but require the hold/confirmation on each allowing option when `approval.destructive` is true, and add a `destructive: true, choice: true` regression test.

jerome-benoit (changes requested):
Reviewed at `bef087a0092486e49ce3a3d35d72327737fcd6bb` after two independent reviews and cross-validation.

Changes are required for two correctness/safety blockers:

- **P1:** a `destructive && choice` approval bypasses the documented hold-to-confirm protection. New inline comment below.
- **P1:** Dismiss can submit the first reject-kind option even though the user selected no answer. This is already covered precisely in the existing [TUI](https://github.com/agent-of-empires/agent-of-empires/pull/3779#discussion_r3941927868) and [web](https://github.com/agent-of-empires/agent-of-empires/pull/3779#discussion_r3941927874) threads; I independently confirmed both paths and am not duplicating them inline. Dismiss must resolve as cancellation while explicit reject-option selection must remain supported.

Also confirmed, non-blocking by itself: the [telemetry thread](https://github.com/agent-of-empires/agent-of-empires/pull/3779#discussion_r3941927863) is valid. The endpoint counts the submitted decision before option mapping, so reject choices can increment allow counters and stale choices can count despite cancellation. Record only the effective resolution and exclude cancellations.

Verified and rejected as additional findings: explicit option IDs are authoritative and stale IDs cancel; old event logs are covered by serde defaults; mixed-kind Gemini permission vocabularies keep the normal approval UI; explicit reject options correctly resolve the tool as denied.

All current CI checks are green for this exact SHA. No local heavy tests were run.

jerome-benoit (changes requested):
Re-reviewed `a160407fa8f612c77be9cd2f8095eb98fc0e0960` with two independent sequential analyses and final arbitration. The previous single-click bypass and web/TUI dismissal findings are addressed; effective-decision telemetry is also corrected.

One P1 remains: the existing [overlapping-hold finding](https://github.com/agent-of-empires/agent-of-empires/pull/3779#discussion_r3949271511) is valid and blocks approval. `ApprovalCard.tsx:84–105` replaces shared timer references without cancelling an earlier hold, while `108–119` cancels only the latest references. With overlapping touches on different options, an abandoned timeout can still submit its captured destructive choice. The new per-option controls expose this cross-option path. Please cancel the previous hold before arming another, ensure cancelled callbacks cannot submit, and add overlapping-touch cancellation coverage. No duplicate inline comment added.

Verified and retained: explicit reject options still answer with the selected option, stale IDs cancel, and Dismiss no longer guesses a reject option. Both reviewers independently confirmed the remaining timer issue; no substantive disagreement remains.

Validation: exact-head source, description and existing discussions read via `gh api`; CI, Tests and CodeQL are green. This is a source-level finding, not a browser reproduction; no local tests were run.

jerome-benoit (approved):
Re-reviewed `62830da2` with two independent sequential analyses and final cross-validation. The remaining overlapping-hold P1 is addressed: `ApprovalCard.tsx:105-137` cancels previous timers before arming a new hold and invalidates cancelled generations; `ApprovalCard.test.tsx:457-494` covers replacement and cancellation. No new blocking finding in this delta.

The earlier dismissal and effective-decision telemetry fixes remain intact. Their outdated threads do not represent outstanding blockers. Pre-existing retry/lifecycle limitations are not regressions introduced by this correction.

Evidence: exact-head `git show`, `git diff a160407f 62830da2`, and `git diff --check` (exit 0). Current CI, Tests and CodeQL are green, verified with `gh pr checks 3779`. No local tests or browser reproduction were run.

PR #3782, opened 2026-09-05
fix(tui): refresh storage during live send

### LINKED ISSUE #3631: TUI: external session changes stay hidden until live-send exits
## Summary

When another process creates a session while the TUI is in live-send mode, the new session is persisted but does not appear in the sidebar. It appears immediately after leaving live-send with Ctrl+Q.

## Steps to reproduce

1. Start the TUI with `aoe`.
2. Select a session and enter live-send mode.
3. From another terminal, create a session with `aoe add /path/to/repo -t live-refresh-repro`.
4. Observe that the new session is absent from the sidebar.
5. Press Ctrl+Q.

## Expected

The sidebar updates while live-send remains active. The current session, input routing, paste state, capture worker, and tmux sizing remain unchanged.

## Actual

The sidebar remains stale for the entire live-send period. Ctrl+Q exits live-send, then the pending storage refresh runs and the session appears.

## Cause

The CLI correctly persists the new row through `storage.update` in `src/cli/add.rs:1144-1160`. The TUI watcher subscribes to `sessions.json` and `groups.json` and sets its dirty latch in `src/tui/home/watchers.rs:269-294`.

The tick loop then gates every disk refresh on `live_send.is_none()` in `src/tui/app.rs:1914-1953`. `decide_disk_refresh` returns `None` whenever live-send is active, including when the watcher is dirty or the heartbeat is due, in `src/tui/app.rs:2745-2755`. The dirty latch is retained until Ctrl+Q clears `live_send`; the next eligible tick calls `reload_storage_only` in `src/tui/app.rs:1974-1978`.

This behavior is explicitly covered by `live_send_blocks_every_decision_branch` in `src/tui/app.rs:4515-4529`. The existing policy protects live input from disruptive reloads, but it also suppresses the storage-only sidebar refresh.

## Scope

The same delay can affect any external `sessions.json` or `groups.json` mutation, including create, rename, group changes, archive, trash, or restore. The event is not lost and the durable state is correct.

## Fix direction

Allow the storage mirror to consume watcher changes during live-send without rebuilding or replacing live-send-owned state. Preserve the active session, selection, input and paste routing, capture worker, and tmux sizing. Keep config or other disruptive full reloads deferred if required.

Add a regression case where an external session insert becomes visible while live-send remains active and verify that the active live session is unchanged.

## Verification

`cargo test --lib live_send_blocks_every_decision_branch -- --nocapture`: 1 passed, 4766 filtered.

Related architectural direction: #2734.

jerome-benoit:
## Description

Fixes #3631.

Allow the TUI storage mirror to consume `sessions.json` and `groups.json` changes while live-send remains active. Watcher notifications trigger the storage-only reload immediately, and the five-second heartbeat uses the same storage-only path during live-send so newly created or temporarily unwatched profiles are still discovered. Full heartbeat and config reloads remain deferred; an overdue full heartbeat is latched across live-send storage refreshes and runs on the first eligible tick after live-send exits.

The live-send target, logical selection, paste buffer, capture worker, input worker, and pane sizing remain pinned while the storage projection rebuilds. If an external mutation hides the active row under a collapsed shelf, the sidebar does not visibly select a peer row. If the refreshed storage removes or renames the active target, the existing drift path ends live-send immediately, reports the reason, and reconciles selection, including an empty sidebar. Reload failures stay latched but their modal is deferred until live-send exits, so input and paste routing are not intercepted.

Regression coverage includes an external `aoe add` becoming visible within the watcher budget, a previously unwatched profile appearing through the periodic fallback, an active row moving to trash without retargeting selection or capture, active-row removal ending live-send, deferred full-heartbeat catch-up, deferred reload-failure presentation, rendered preview text matching copied text, and input remaining routed to the original live pane.

Validation:
- `cargo fmt --check`
- `cargo test`
- `cargo test --lib tui::app::tests` (41 passed)
- `cargo test --lib tui::home::file_watch_tests` (37 passed)
- `cargo test --lib tui::home::tests::live_send_mode` (35 passed)
- `cargo test --lib tui::home::tests::preview_drag_select` (31 passed)
- `cargo test --features e2e-tests --test e2e filewatch_tui_reload` (2 passed)

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

The changed behavior has deterministic TUI E2E coverage. No visual styling changed, so no screenshot or recording was added.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

OpenAI Codex GPT-5.6 Sol via Oh My Pi.

The agent reproduced the issue against the full TUI binary, ran independent fresh code and design reviews from the current base-to-head diff, incorporated every GitHub review and general comment, corrected every validated finding, and reran the targeted unit, lifecycle, E2E, full-suite, format, and lint checks listed above.

**NOTE:**
When responding to reviewer questions, please respond yourself rather than copy/pasting reviewer comments into an AI and pasting back its answer. We want to discuss with you, not our AI :)

## Summary

- Updated the TUI sidebar during live-send when external sessions or groups change.
- Added watcher and periodic storage refreshes without interrupting live-send.
- Preserved the active session, selection, input routing, preview state, capture target, and pane sizing.
- Deferred full reloads and reload-error dialogs until live-send ends.
- Ended live-send and repaired selection when the active session changed or was removed.
- Added regression and end-to-end tests for refresh behavior, selection, input routing, deferred reloads, and preview capture.

## Benefits

The sidebar stays current during live-send. External changes appear promptly, while disruptive updates remain safely deferred.

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -852,6 +852,7 @@ impl App {
         let mut last_metrics_sample = std::time::Instant::now();
         let mut last_daemon_status_refresh = std::time::Instant::now();
         let mut last_disk_refresh = std::time::Instant::now();
+        let mut full_heartbeat_deferred = false;
         let mut last_spinner_redraw = std::time::Instant::now();
         let mut last_heartbeat = std::time::Instant::now();
         let mut last_presence_refresh = std::time::Instant::now();
@@ -1951,20 +1952,9 @@ impl App {
                 needs_full_refresh = true;
             }
 
-            // Disk reload: heartbeat (defense-in-depth) plus the
-            // file-watch-driven kick. Both gate on `live_send.is_none()`
-            // so reloads never interrupt a paste-in-progress; the dirty
-            // flag stays latched (Acquire pairs with the forwarder/adapter
-            // Release) until the next eligible tick. The watcher is scoped
-            // to `sessions.json` / `groups.json`, so the watcher path calls
-            // `reload_storage_only` (storage + profile rediscovery only);
-            // the heartbeat path calls full `reload()` to refresh the
-            // status-hook config cache and mouse-capture toggle.
-            //
-            // Config kick runs before the storage-mirror block:
-            // `refresh_from_config` invalidates profile-derived state that
-            // the block reads. Same `live_idle` gate; recomputing
-            // `tool_hotkey_cache` mid live-send disrupts input.
+            // Full/config reloads stay deferred during live-send to preserve input
+            // policy and mouse-capture state. Storage-only reloads preserve the live
+            // target unless it drifts, in which case normal teardown restores sizing.
             let live_idle = self.home.live_send.is_none();
             let config_kick = take_config_refresh_kick(live_idle, &self.home.config_watch.dirty);
             if config_kick {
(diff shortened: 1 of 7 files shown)


njbrake (approved):
Approve.

Checked the refresh state machine exhaustively rather than by inspection: no watcher kick is dropped now that the latch is consumed unconditionally, no full reload runs while live-send is active, and the deferred heartbeat survives an arbitrary run of live-send ticks and clears itself on the first idle one. Both timer resets are correct.

The e2e test is the strongest part. It drives the real binary through the reproduction and also asserts keystrokes do not reroute to the peer row, which is the failure mode that would have made this worse than the bug.

Pushed one commit guarding `selected_stopped` and `selected_structured` on live-send. Live storage means both can now flip mid-session, and either one zeroed the capture target and pointed the worker off the live pane while input still routed to it. Narrow and self-healing once drift fires, but the three checks beside them already carry the guard.

Residual concern, not blocking: live-send previously did no disk work and now rebuilds the tree at least every five seconds on the render loop. Nothing measures that under a large paste.

PR #3792, opened 2026-09-07
fix(tmux): keep pipe-pane input off tmux that crashes on a dead pane

hairbui76:
## Description

**Problem.** Users on any released tmux (3.4 through 3.7a) can lose every session at once. The VT live transport arms `tmux pipe-pane -IO` on the previewed pane and pushes keystrokes through the `-I` side. When that pane's agent exits under `remain-on-exit` (for example "No, exit" at Claude's folder-trust prompt), tmux frees the pane's pty event but keeps the pipe alive, so the next byte delivered on the pipe calls `bufferevent_write(NULL)` and the tmux server segfaults. Every aoe session lives in that one server, so they all die; the TUI then shows every other row flipping `Idle -> Error` in the same second. It presents as "starting a new session killed all my idle sessions".

`tmux_supports_pipe_pane_io` assumed this crash was fixed in tmux 3.4 and enabled pipe input from 3.4 on. It is not: the upstream fix is tmux commit f751d3f ("When destroying pane, also close pipe and destroy its event", Jul 27 2026), and `server_destroy_pane` in tags 3.4, 3.5, 3.5a, 3.6, 3.7 and 3.7a all lack it.

image

**Fix.**
- Add `tmux_supports_pipe_pane_input`, gated on tmux >= 3.8 (the first release that can carry f751d3f). Below that the channel arms `pipe-pane -O` only; output still streams into the vt100 grid, and keystrokes take the existing `send-keys` fallback.
- `VtChannel` records whether it was armed with input. `input_mode` reports `None` for output-only channels, so both the TUI (`dispatch_via_fork`) and the web terminal (`live_ws`) route input through `send-keys`; `write_input` refuses to write on such channels as a second guard.
- The `vt channel armed` log line now carries a `flags` field (`-IO` or `-O`).
- Docs for the VT live transport and the `vt_live` setting say keystrokes ride the socket only on tmux 3.8+.

Behavior on tmux 3.8+ is unchanged.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the regression is a tmux server crash. Provoking it in e2e would kill the harness's shared tmux server and leave an apport crash dump on every run. Instead, two unit tests in `src/tmux/vt.rs` cover the fix: `pipe_input_requires_a_tmux_that_survives_a_dead_pane_write` pins the version gate (3.4, 3.5a, 3.7a off; 3.8, next-3.8, 4.0 on), and `output_only_channel_never_writes_to_the_pane` proves an output-only channel reports no input mode and never writes to a connected socket, while a full channel delivers. Both were watched failing before the fix.

Verification run on tmux 3.4-1build1 (Ubuntu):
- `cargo test --lib`: 6653 passed (5 `server::login` failures are pre-existing on `main` under `umask 002`, pass with `umask 022`)
- `cargo test --test integration tui_attach_detach`: 8 passed
- `cargo test --features e2e-tests --test e2e -- live_send live_takeover new_session`: 15 passed (on tmux 3.4 this now exercises the `send-keys` input path)
- `cargo clippy --lib --tests`, `cargo fmt --check`: clean

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Claude Fable 5.1)

**Any Additional AI Details you'd like to share:** Claude Code did the log and core-dump investigation, the tmux reproduction, the fix, tests and this description; the author reviewed and ran the checks.

## Summary

- Prevents tmux crashes on versions 3.4–3.7a.
- Uses `send-keys` on older versions and socket input on tmux 3.8+.
- Protects output-only channels from invalid writes.
- Improves cursor handling, resize synchronization, pane capture, and hyperlink updates.
- Adds tests for version handling, input behavior, cursor placement, resize races, and link updates.
- Updates live transport and configuration documentation.

These changes improve session stability and keep keyboard input reliable across supported tmux versions.


src/tmux/vt.rs
```diff
@@ -457,14 +458,15 @@ fn lookup_osc52(session: &str) -> Option<Arc<Osc52Channel>> {
...
+/// normal (`ESC [ A`). `None` means no channel is armed, it is output-only, or
+/// its forwarder has disconnected. Presence of `Some` is the single-writer
+/// signal: while live, ALL pane input must go through [`try_send_input`]
+/// (never `send-keys`), so the two writers don't interleave. Gating on
+/// liveness means a dead channel reports `None` and input falls back to
+/// `send-keys` rather than vanishing.
 pub(crate) fn input_mode(session: &str) -> Option<bool> {
     lookup(session)
-        .filter(|c| c.is_alive())
+        .filter(|c| c.input && c.is_alive())
```
jerome-benoit, line 769: [P2] Preserve cursor-mode translation for output-only web channels

On tmux 3.4–3.7c, this filter makes a live output-only channel return `None`, so [the web input handler](https://github.com/agent-of-empires/agent-of-empires/blob/7cf4b6c87d6f49f900b6efe936c49623925e7d2b/src/server/live_ws.rs#L1235-L1245) skips `translate_cursor_keys`. The browser sends normal-mode sequences such as `ESC [ A`; the fallback [`send_raw_bytes`](https://github.com/agent-of-empires/agent-of-empires/blob/7cf4b6c87d6f49f900b6efe936c49623925e7d2b/src/tmux/session.rs#L1299-L1317) uses `send-keys -H`, which transmits literal bytes rather than translating named keys. Applications requesting DECCKM therefore receive `ESC [ A` instead of `ESC O A`, regressing arrows/Home/End that were translated on the previous live-channel path. Keep read access to the live cursor mode separate from socket-input capability, translate web bytes before choosing either raw-byte transport, and add consumer coverage with an output-only channel and `app_cursor=true`. This is established by source tracing, not an end-to-end UI reproduction.

jerome-benoit (changes requested):
Reviewed exact head `7cf4b6c87d6f49f900b6efe936c49623925e7d2b` with two independent sequential analyses and final arbitration.

The output-only mitigation is sound in scope, but one introduced P2 blocks approval: web cursor-key translation is lost when socket input is disabled. Please address the inline finding without re-enabling unsafe pipe input.

The existing [split-window documentation finding](https://github.com/agent-of-empires/agent-of-empires/pull/3792#discussion_r3946485874) is valid, pre-existing and non-blocking (P3); I have not duplicated it. TUI named-key fallback was checked and does not have the web raw-byte translation problem. The initial reviewer disagreement was resolved by tracing `send-keys -H` through tmux's literal-byte path.

Validation: exact-head/base source reads via `gh api`, plus verified tmux 3.7c source. No local build, test or crash reproduction was run. CI, Tests and CodeQL were awaiting authorization; I approved those workflows, and their results are still pending.

src/tmux/vt.rs
hairbui76: Fixed in b3879c3d. The cursor mode is now read separately from socket-input capability: `vt::cursor_mode` reports the live grid's DECCKM state for any live channel, while `input_mode` still means "socket input is live". The web handler re-encodes browser bytes through `pane_input_bytes` before choosing either transport, so a DECCKM app receives `ESC O A` over `send-keys -H` as it did over the socket. Pipe input stays off below tmux 3.8.

Coverage: `output_only_channel_still_reports_the_pane_cursor_mode` (vt.rs) and `pane_input_bytes_translates_cursor_keys_for_an_output_only_live_grid` (live_ws.rs), both with an output-only live channel and app_cursor=true; both failed before the change.

docs/guides/configuration.md
```diff
@@ -473,7 +473,7 @@ vt_live = true
...
+| `vt_live` | `true` | Render agent previews and the web dashboard's agent terminal from a persistent VT channel: `tmux pipe-pane` streams the pane into an in-process terminal grid, and on tmux 3.8 or newer keystrokes go back over the same socket (older tmux keeps `send-keys`). The paired host and container shells, split windows, and every fallback keep tmux's rendered `capture-pane` snapshots to avoid prompt-paint races; there a raw observer retains OSC 52 clipboard forwarding without affecting rendering. Disabling this setting puts every surface on the slower `capture-pane` / `send-keys` path. Applies in place on the next TUI capture cycle. |
```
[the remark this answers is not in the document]
hairbui76: Adjusted in b3879c3d. configuration.md and live-mode.md now say a split window is composited from capture-pane with the TUI preview keeping pane 0 on its VT grid (`composite_with_first_pane_rows` in live_send). web/terminal.md is unchanged on purpose: live_ws sends any split window to `capture_window_composited_with_cursor` for every pane, so listing it among the capture-pane cases is accurate for the web dashboard.

jerome-benoit (approved):
Re-reviewed `b3879c3db8910944533eb72e8872a53b2cd6ab26` with two independent, sequential analyses and final arbitration. No remaining blocking finding.

- The P2 is fixed: `cursor_mode` remains available for output-only live grids (`src/tmux/vt.rs:463`), and browser input is translated before either transport (`src/server/live_ws.rs:1249`). The new consumer regression test covers this case. Pipe input remains disabled below tmux 3.8.
- Your split-window clarification is correct. My previous blanket endorsement of the documentation finding was too broad: TUI composition retains pane 0's VT rows, whereas the web path captures all panes (`live_ws.rs:775–794`). Keeping the web documentation unchanged is justified.

Validation: exact-head source and replies inspected through `gh api`; both reviewers agreed after independent analysis. No local build or reproduction was run. Current test jobs are green; Rust CodeQL analysis is still running, so this is code approval, not a claim that every CI check has finished.

src/tmux/vt.rs
```diff
@@ -3449,6 +3494,71 @@ impl Drop for Osc52Channel {
...
+    let alive = Arc::new(AtomicBool::new(false));
+    let ch = Arc::new(VtChannel {
+        name: name.to_string(),
+        input,
+        owner_id: new_pipe_owner_id(),
+        target: format!("{name}:^.0"),
+        parser: Arc::new(Mutex::new(vt100::Parser::new(4, 20, SCROLLBACK_LINES))),
+        stream: Arc::new(Mutex::new(None)),
+        app_cursor: Arc::new(AtomicBool::new(false)),
+        alive: alive.clone(),
```
jerome-benoit: [P1] Port the shared test constructor to the merged VtChannel

`VtChannel` now has `lifecycle: Arc<AtomicU8>` (line 2152), not `alive`, but `dummy_channel_with_input` still returns `Arc<AtomicBool>` and initializes this removed field. It also omits the new `links`, `snapshot`, `drain`, `resync_target` and `resize_seq` fields. The wrapper at 4055–4056 promises `Arc<AtomicU8>` while delegating to that boolean constructor. These are incompatible Rust types/fields, so the Unix lib-test target cannot compile.

Please port the full shared constructor and its callers to the lifecycle initialization already present in base `8f0cece8:src/tmux/vt.rs:3921–3956`, retaining the `input` parameter and output-only assertions; update the remaining boolean/alive accesses too. The combined merge diff only shows the wrapper and misses this stale constructor. Established from exact-head source inspection, not a local compiler run.

jerome-benoit (changes requested):
Re-reviewed the merge at `debfa695` against `8f0cece8` and the previously approved head, including the existing replies. The production input gate, output-only cursor-mode translation, and split-window documentation remain correct. One merge-introduced test-compilation blocker remains inline.

Evidence: exact-head/base `git show` inspection and clean `git diff --check 8f0cece8 debfa695`. No local build or test was run. The green title/template checks were not Rust validation: CI, Tests and CodeQL required approval. I have approved all three; they are queued.

src/tmux/vt.rs
hairbui76: Thanks — you were right, and the merge had moved further than the review could see.

**Ported the shared constructor.** `dummy_channel_with_input` now returns `Arc<AtomicU8>`, initialises `lifecycle` from `VtLifecycle::Starting`, and carries the fields it was missing. The `input` parameter and the output-only assertions are unchanged. Callers updated with it: `register_live_for_test` and `output_only_channel_never_writes_to_the_pane` now use `VtLifecycle::Live.store(&lifecycle)`, and the cursor-mode test's stand-down became `VtLifecycle::fail(&channel.lifecycle)` — `is_alive()` is `lifecycle() == Live`, so `Failed` preserves exactly what `alive.store(false)` asserted.

**One thing to flag about the fix as specified.** Porting `8f0cece8:src/tmux/vt.rs:3921-3956` verbatim would still not have compiled: `main` has moved past that base, and `resync_target: AtomicU64` / `resize_seq: AtomicU64` are gone, replaced by a single `resize: Mutex<ResizeState>` (#3827). I merged current `main` first and ported against that, so the constructor now matches the 31-field struct — the 30 upstream fields plus this PR's `input`. That merge also cleared the conflict the PR was carrying; it is `MERGEABLE` again.

**Compiled and ran it.** `cargo fmt --check` clean; the Unix lib-test target builds. Full lib suite: 6881 passed, 0 failed, 2 ignored. `tmux::vt` on its own: 78 passed. The three `output_only_*` tests pass, including the `live_ws` cursor-key translation case.

Thanks for approving the CI runs.

jerome-benoit (approved):
Re-reviewed `02302dab` against `0bd6dd53`, including your reply and the merge integration. Two independent sequential analyses and final arbitration found no remaining blocker.

The P1 is fixed: the shared constructor and callers use `AtomicU8` lifecycle state and the current `ResizeState` (`src/tmux/vt.rs:3617-3656,4262-4311`). Your correction about #3827 is right: restoring the old resize atomics would have been wrong against the new base. The output-only guard and web cursor-key translation remain intact (`vt.rs:753-771,3303`; `src/server/live_ws.rs:1435-1447`).

Evidence: exact-head/base source and all thread replies inspected; `git diff --check 0bd6dd53...02302dab` clean. No local tests or crash reproduction run. CI, Tests and CodeQL have been authorized and are queued/running; the author's reported test results are not independent CI evidence.

njbrake (approved):
Approving. Verified the load-bearing claim against upstream tmux rather than the description: `cmd_pipe_pane_read_callback` at tag 3.7c calls `bufferevent_write(wp->event, ...)` with no NULL check and fires only when `-I` enables `EV_READ`; `server_destroy_pane` at 3.7a/3.7b/3.7c NULLs `wp->event` and leaves `pipe_event` armed; 3.8-rc adds the `wp->pipe_fd` teardown. `window_pane_destroy` does free `pipe_event`, so the crash window is the `remain-on-exit` gap and dropping `-I` closes the class, not just the common trigger.

The version list in the description omits 3.7b and 3.7c. Both also lack the fix, so the `>= 3.8` gate covers them; only the description understates the range.

The gate extends the existing mechanism (`cached_tmux_support`, `parse_tmux_pipe_support`, the `flags` parameter already used for the OSC 52 observer's `-O` arming) rather than adding a second one, and the single-writer invariant is preserved by leaving one writer instead of two.

Pushed one commit for the surfaces the change made stale: the `vt_live` doc comment is the field's description on the TUI and web settings pages, so it still promised socket keystrokes and still carried the split-window claim corrected in the guides; the `live_ws` module header said the same; the OSC 8 test helper's comment referred to `-IO` for a threshold that gates `-O`.

Two non-blocking nits left alone. Arming forks `tmux -V` twice, once per gate, under one shared `TmuxCommandDeadline` budget; one `OnceLock` holding the parsed `(major, minor)` would serve both. And the new module-level `tmux_version` shares a name with the test module's `tmux_version(&str) -> (u32, u32)`, which shadows it through `use super::*`.

Residual risk is coverage, not correctness. tmux 3.8 is not GA, so this disables socket input for every user and every CI job, and the `-I` path plus `write_input`'s success branch lose end-to-end coverage until a 3.8 runner exists. Worth a follow-up.

Issue #1858, opened 2026-06-02
Cockpit: end-of-turn stall hangs for 30 min, silent-orphan watchdog over-suppressed by off-protocol grace floor after cost-resolved usage


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

A cockpit session wedges at end-of-turn and looks permanently stuck. The agent streams its entire final message, the daemon records the cost-resolved usage, then the turn never closes: no `Stopped` frame is ever written. The UI keeps the running spinner and a live Stop button indefinitely, and a user-queued follow-up sits in `QUEUED` forever because it is gated on "sent when the current turn ends".

Live incident on session `db04343cfab648b8` (a `/aoe-implement` run, claude-agent-acp). Observed:

- `aoe cockpit status` showed `highest_seq` frozen. Last frames were `AgentMessageChunk` then two `UsageUpdated`. No `TurnEnded` / `Stopped` was emitted.
- `aoe cockpit cancel` returned "cancel sent" but had no visible effect; `highest_seq` stayed frozen after 3s and again after 5s more.
- `aoe cockpit restart` fixed it: stopped the runner PID, the reconciler respawned a fresh worker, rebound the ACP session, restored config, and the queued prompt fired.
- `aoe cockpit logs --session <id>` showed only the runner startup marker, nothing about the turn, the cancel, or the stall.

The headline is not "the watchdog is missing". A silent-orphan watchdog exists (#1240, #1248). The headline is that the watchdog's own anti-false-fire suppression (#1360, #1401, #1406) over-suppresses at genuine end-of-turn, so a long agentic turn that ran any background work hangs for the full 30-minute off-protocol grace floor before anything self-heals. Three smaller gaps make a self-healing-in-30-min wedge look permanent: cancel surfaces no escalation countdown on the CLI, the natural log surface is empty, and a build-stale worker pinned by a never-ending turn is never auto-replaced.

## Evidence from the affected session

`~/.agent-of-empires/cockpit_events.db`, the wedge boundary for `db04343cfab648b8` (seq 1371+ are from the manual restart, not a natural turn end):

```
1368 | {"AgentMessageChunk":{"text":"ak / hold?"}}                          <- final message
1369 | {"UsageUpdated":{"usage":{"used":223339, ... "cost":null}}}           <- mid-turn usage
[...]

PR #3814, opened 2026-09-08
fix(tests): repair the ACP test fixtures behind #3790 and #3811

### LINKED ISSUE #3790: ACP reset test fails with “Text file busy” when spawning its shell fixture
*AI-generated report based on CI logs and source inspection.*

## Problem

On Linux, the following test fails while spawning its shell fixture, before the ACP handshake or assertions:

```text
acp::acp_client::reset::tests::reset_between_prompts_with_background_agent_is_refused
```

```text
spawn failed: Text file busy (os error 26)
command=/tmp/.tmpxn7Ayl/fake-reset-agent.sh
```

Failure location: `src/acp/acp_client/reset.rs:351`.

**Expected:** the fixture starts and the test verifies that reset is refused while a background agent is active.

## Evidence

Command:

```sh
cargo test --no-default-features --lib --bins
```

[CI logs](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34059057068/job/101556171509): **6657 passed, 1 failed, 2 ignored**.

## Suspected cause

The helper writes the script synchronously and closes its file before spawning. A concurrent `fork` could inherit the writable descriptor and retain it after the parent closes its copy.

An isolated experiment demonstrates that this mechanism produces `ETXTBSY`, but **the descriptor holder and actual interleaving on the runner remain unknown**.

## Investigation needed

Capture file and process syscalls during a failing run to identify the descriptor holder. A reliable reproduction of the suite failure and a confirmed fix are still missing.

### LINKED ISSUE #3811: Tests: malformed ACP usage fixtures mask silent-orphan recovery coverage
## Problem

`acp-worker/test-shim/shim.mjs` emits `usage_update` with `input_tokens`/`output_tokens` instead of mandatory `used`/`size`. The Rust schema rejects it with `missing field used`. Consequently, `tests/integration/acp_silent_orphan.rs::silent_orphan_fires_on_cost_then_silence` passes through the **no-cost, 5 s cancellation path**, not the advertised 300 ms cost-bearing path. Its expected `prompt_orphaned` is also obsolete for valid cost-bearing completion, which currently produces `prompt_complete` (#2237).

The malformed positive payload dates to `fe6b95e9` (#1248), before #3538. This follow-up is separate from #3538’s runner-session-cache fixture correction and resolved historical deadline race.

## Reproduction / evidence

At `6d11a8ad`, change only the positive shim payload to `used: 300, size: 200000`, keeping its cost. Run:

```sh
cargo test --features web --test integration \
  acp_silent_orphan::silent_orphan_fires_on_cost_then_silence \
  -- --exact --nocapture
```

Real macOS control: **5/5 receive `UsageUpdated`, then `Stopped { reason: "prompt_complete" }`; the unchanged assertion fails**. [Run and artifacts](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34160645934), [durable diagnostic details](https://github.com/agent-of-empires/agent-of-empires/pull/3538#issuecomment-5575528943).

## Acceptance

- Correct malformed usage payloads in the silent-orphan shim scenarios, including the background-Bash and scheduled-wakeup fixtures; align affected assertions with their actual contracts.
- Cover valid cost-bearing completion and genuine **no-cost** orphan cancellation distinctly. Assert receipt of the cost-bearing `UsageUpdated` before accepting its terminal outcome, so silently rejected fixtures cannot pass.
- Update misleading test names/comments and verify the affected integration scenarios without increasing deadlines to conceal failures.

**Scope: test fixtures and coverage only.** No runner/protocol redesign or change to runtime completion policy. #3806 concerns timing-wait practices, not this malformed-payload/obsolete-assertion defect.

njbrake:
## Description

Two ACP test-infrastructure defects, one commit each. Kept together because both are the same surface: the fixtures the ACP tests spawn.

**#3790.** `execve` on a file another process still holds open for writing fails with `ETXTBSY`. The fixture writer closes its descriptor before spawning, but a concurrent spawn elsewhere in the test binary forks between that `open` and `close` and the child keeps the writable copy until it execs. Reset fixtures now run as `/bin/sh <script>`; `sh` opens the script for reading only, so the window cannot exist rather than being retried past.

**#3811.** `shim.mjs` emitted `usage_update` with `input_tokens` / `output_tokens` instead of the mandatory `used` / `size`, so the schema rejected every frame and the daemon never saw a usage at all. Three scenarios advertised a cost-bearing path and ran the no-cost one. Correcting the payloads moves two outcomes, both to behavior the runtime means to have:

- A turn that wraps up its accounting and never returns the `PromptResponse` ends as `prompt_complete` through the #2237 finished-but-unacked recovery, not `prompt_orphaned`.
- A backgrounded Bash turn ends the same way once its accounting frame drops the off-protocol floor (#1858), rather than staying suppressed.

So the assertions move and the runtime does not. The cost-bearing and no-cost shapes are now separate scenarios that can only reach their own outcome (`COST_THEN_SILENCE` against `SILENCE_NO_COST`, `BACKGROUND_BASH_ORPHAN` with or without `WRAP_UP`), and every scenario depending on a usage frame asserts the daemon received it.

`docs/development/internals/structured-view.md` still described the pre-#2237 cancel-and-respawn as the watchdog's only outcome, which is the contract these tests just stopped asserting. Corrected here rather than left disagreeing with the suite.

Fixes #3790
Fixes #3811

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**AI Model/Tool used:**

Both defects were reproduced locally before the fix: the ETXTBSY mechanism from the issue's analysis, and the silent-orphan path change by correcting only the shim payload and rerunning the unchanged assertions.

https://claude.ai/code/session_01Ebc3TdovhRpJzArpTUsJNr

## Summary

- Start ACP reset fixtures through `/bin/sh` to prevent Linux `ETXTBSY` failures.
- Fix usage fixtures to send valid `used` and `size` fields.
- Separate cost-bearing completion from no-cost orphan cancellation.
- Verify that usage updates reach the daemon.
- Update test names, comments, terminal expectations, and structured-view documentation.

These changes improve test reliability and ensure the tests validate the intended runtime behavior.


jerome-benoit (approved):
Reviewed all six changed files at `eb8f66c6` with two independent sequential analyses and final arbitration. No blocking finding or substantive disagreement.

The fixtures now deliver valid usage and distinguish cost-bearing completion, no-cost orphan cancellation and off-protocol suppression. These expectations match the existing runtime (`connection.rs:2005-2035`, `watchdog.rs:194-198`), rather than changing its policy. The Unix fixture callers consistently pass their POSIX scripts to `/bin/sh`.

The existing [CodeRabbit timing concern](https://github.com/agent-of-empires/agent-of-empires/pull/3814#discussion_r3958570156) is valid but non-blocking: the 15-second drains do not independently prove the 300ms fast grace rather than the 5-second base grace. They do prove #3811's usage-delivery and terminal-outcome distinction; timestamp-injected unit tests at `watchdog.rs:483` and `:603` cover grace selection. Please correct the comments claiming that only the fast path fits inside the drain. No duplicate inline finding added.

Also non-blocking: shorten fixture narration, replace the obsolete `SILENT_ORPHAN` comment reference at `shim.mjs:246`, and qualify the inherited-descriptor explanation as a possible cause, as #3790 does.

Evidence: `git diff 59aec660 eb8f66c6`, numbered source/schema/caller inspection, live issues and threads, `git diff --check` (exit 0). No local tests/builds or new runtime reproduction. Exact-head CI: 32 successes, 1 skipped; only title/template checks remain queued.

PR #3820, opened 2026-09-08
fix(sandbox): reclaim per-session agent stores that no session owns

### LINKED ISSUE #3816: Per-session agent stores are never reclaimed: 5.68 GB of sandbox-v2 orphans from purged sessions
Per-session agent stores under `<config dir>/sandbox-v2/<instance-id>` are never removed. Deleting, trashing or purging a session leaves its store on disk, and nothing reclaims it later.

Measured on one machine today: 22 stores in `~/.codex/sandbox-v2` belong to no session in any profile, totalling 5.68 GB. Codex stores carry SQLite state, so they run about 260 MB each. Claude stores are smaller but leak the same way: three sessions trashed within the last hour still hold 13.7 MB, 11.3 MB and 14.1 MB.

`sandbox_store_dir` and `legacy_sandbox_store_dir` are the only accessors, and both construct paths. No delete path in `src/session/` references the directory, so purge does not reach it.

Two consequences beyond disk. A store holds the agent's `.credentials.json`, so an orphan is a copy of a live credential sitting in a directory nothing will ever open again. And an orphan is named by an instance id that no longer resolves, so the only way to identify one is to diff the directory against every profile's session registry.

Purge should remove the session's stores. A reclaim pass for stores whose instance id resolves in no profile would cover the ones already stranded, including those created before this lands. #3756 asks for the same thing for `aoe-vi` volumes, which leaked the same way and for the same reason.

njbrake:
## Description

When you permanently delete a sandboxed session, AoE was leaving behind the private folder that session used as its agent's home. Every sandboxed session gets one, and it holds that agent's saved login, its history and its config. Nothing ever cleaned these up, so they piled up: one machine had 22 of them, 5.68 GB, all belonging to sessions that no longer exist. Worse than the disk, each one is a working copy of a credential sitting in a folder nothing will ever open again, and since the folder is named after a session id that no longer resolves, there was no way to even find them.

Two things change. Deleting a session now removes its folder along with its container. And a new command finds the ones already stranded and offers to remove them, which covers everything that leaked before this landed.

Because these folders hold live credentials, the sweep is deliberately timid. It only calls one an orphan if its session is missing from every profile, not just one. If it cannot read a profile's session list for any reason, it stops and removes nothing rather than assuming those sessions do not exist. It removes a folder only when no container for it exists at all, under any runtime installed, not merely when the container is stopped, and if it cannot reach a runtime to ask, it assumes one exists. And it reports before it removes: the bare command prints what would go and how much space that frees, and only `--delete` actually removes anything.

```
$ aoe sandbox reclaim
Scanned 1 store root(s) against 1 session(s) across all profiles.
  orphan     ~/.claude/sandbox-v2/2222222222222222  (244 KB)
1 orphaned store(s), 244 KB total.

Run `aoe sandbox reclaim --delete` to remove them.
```

A session still on the old shared store is skipped: it owns no private folder yet, and the migration may be publishing the one it will own. The sweep picks it up later if it ends up stranded.

Container evidence goes through v027's own probe, including its fail-closed answer for a runtime that cannot be reached, and is re-taken per candidate rather than once for the pass. Requiring absence rather than quiescence is what makes it safe: a stopped container can be started between the check and the removal, and no lock a reclaim can hold is observed by `docker start`.

The pass holds v027's transition lock **shared**, not exclusive. Exclusive would also block `Storage::update`, which is what publishes the row for a store being created, so it would have turned the creation race into a guaranteed loss. Shared still excludes v027's own planning and publishing. The pass also cannot see a half-copied store: v027 stages under `.v027-stage-<id>` and renames, and only a bare 16-hex name is ever a candidate, so a store appears whole or not at all.

Ownership spans both build namespaces. Debug and release builds keep separate app dirs but share `$HOME`, and so share these store roots; reading only one registry would make `cargo run -- sandbox reclaim --delete` delete the installed build's credentials.

One deliberate gap: purge removes stores only when the delete also removes the container, since the store is bind mounted into it. A delete that keeps the container leaves the store, and the sweep catches it later. Documented in `docs/guides/sandbox.md`.

### Known limitation

Store seeding happens before the owning row is published (`cli::add` runs `on_create` hooks, and so container preparation, before it persists). The pass mitigates that with a fifteen-minute mtime grace period and an ownership re-read before deletion, but a root-mtime window is not a reservation through registry publication. Closing it properly belongs on the creation path, not here.

### Review round

A fresh-context review found three defects in the first commit, all fixed in `291c85d`:

- A purge whose container teardown **failed** still deleted the store. `PurgeTransaction::complete_inner` rolls the purge back on teardown failure, so the session survived, logged out and with its history gone. Removal now runs only once the container is provably gone.
- `aoe sandbox reclaim` **refused forever** on any install holding an archived or trashed pre-v027 sandboxed session, because parked rows stay below the current store generation by design. `aoe migrate` said "migrations complete" and the next reclaim still claimed a move was in flight, which killed the feature on exactly the long-lived installs that accumulate stranded stores. The guard now keys on a published transition plan.
- The dev/release namespace hazard above.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

Copy-only web change: the delete dialog said the sandbox checkbox removes only the container, which now understates it. Covered by a new assertion in `DeleteSessionDialog.test.tsx`, which the existing `session.delete-dialog-keyboard` matrix entry already maps to that component, so no new matrix entry is owed. `node tests/validate-coverage-matrix.mjs` passes.

**TUI / CLI (e2e test)**

`tests/e2e/sandbox.rs::sandbox_reclaim_reports_before_it_removes` creates a real session, plants one claimed and one orphaned store, and asserts the bare report names only the orphan and deletes nothing, then that `--delete` removes the orphan and leaves the claimed store.

Unit coverage: 10 tests in `src/session/sandbox_store_reclaim.rs` and 3 in `src/session/deletion.rs`, including the failed-teardown and parked-session regressions, each confirmed to fail without its fix.

`cargo fmt`, `cargo clippy --all-targets`, `cargo test` (6799 + 286), the e2e test, and the web `format:check` / `lint` / `tsc -b` / `test:unit` all pass locally.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Implementation and tests written by the model from the issue plus direction on the three judgement calls (what counts as an orphan, reusing v027's liveness rule instead of inventing one, and requiring a report mode before anything deletes). Reviewed by a separate fresh-context agent; its three findings are the review round above.

Fixes #3816

## Summary

- Removes private sandbox stores after successful session container deletion.
- Adds `aoe sandbox reclaim` to report or delete orphaned stores.
- Checks all profiles, runtimes, and debug/release namespaces before reclaiming stores.
- Preserves stores linked to active, stopped, recent, or uncertain sessions.
- Fails safely when registry, migration, or container runtime data is unavailable.
- Updates documentation, deletion-dialog text, and test coverage.

## Benefits

This reduces disk usage and limits exposure of stranded credentials and session data. Conservative checks reduce the risk of deleting stores that may still be needed.

### DIFF
--- a/src/session/sandbox_store_reclaim.rs
+++ b/src/session/sandbox_store_reclaim.rs
@@ -1,0 +1,573 @@
+//! Reclaim per-instance agent stores whose session is gone.
+//!
+//! A sandboxed session mounts `<agent config>/sandbox-v2/<instance id>` as the
+//! agent's config directory, so the store holds that agent's credentials.
+//! Purging a session removes its own stores; this pass covers the ones already
+//! stranded, including stores left by versions that removed nothing.
+//!
+//! Both are deliberately narrow, because the thing being deleted is a copy of a
+//! live credential:
+//!
+//! - A store is an orphan only when its id resolves in *no* profile. A registry
+//!   that cannot be read is never "a profile with no sessions"; the pass fails
+//!   and deletes nothing.
+//! - Liveness is v027's judgement, not a second one: a store whose container is
+//!   running, or whose path is not a plain directory, is preserved, and a
+//!   container runtime that cannot answer reads as live.
+//! - The pass runs under v027's transition lock and refuses while that
+//!   migration has work outstanding, so it cannot delete a store `aoe migrate`
+//!   is mid-copy of.
+
+use crate::migrations::v027_isolate_sandbox_stores as v027;
+use anyhow::{bail, Context, Result};
+use serde_json::Value;
+use std::collections::BTreeSet;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+/// Why a store that resolves in no profile was kept anyway.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum Preserved {
+    /// Its container is live. A runtime that cannot answer takes this arm
+    /// too, which is the whole point of v027's probe.
+    Live,
+    /// Not a plain directory, so what it is cannot be established.
+    Ambiguous,
+}
+
+impl Preserved {
+    pub fn label(self) -> &'static str {
+        match self {
+            Self::Live => "container is running, or the runtime could not be asked",
+            Self::Ambiguous => "not a plain directory",
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Orphan {
+    pub id: String,
+    pub path: PathBuf,
+    pub bytes: u64,
+}
+
+#[derive(Debug, Default)]
+pub struct Plan {
+    /// Store roots scanned, in path order.
+    pub roots: Vec<PathBuf>,
+    /// Stores that resolve in no profile and are safe to remove.
+    pub orphans: Vec<Orphan>,
+    /// Stores that resolve in no profile but were kept.
+    pub preserved: Vec<(PathBuf, Preserved)>,
+    /// Session rows that claim a store id, across every profile.
+    pub owners: usize,
+}
+
+impl Plan {
+    pub fn bytes(&self) -> u64 {
+        self.orphans.iter().map(|orphan| orphan.bytes).sum()
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct Outcome {
+    pub plan: Plan,
+    /// Stores actually removed.
+    pub removed: Vec<Orphan>,
+    pub failures: Vec<(PathBuf, String)>,
+}
+
+impl Outcome {
+    pub fn freed(&self) -> u64 {
+        self.removed.iter().map(|orphan| orphan.bytes).sum()
+    }
+}
+
+/// What would be reclaimed, without removing anything.
+pub fn report() -> Result<Plan> {
+    let app_dir = crate::session::get_app_dir()?;
+    let home = dirs::home_dir().context("home directory unavailable for store reclaim")?;
+    let _lock = guard(&app_dir)?;
+    plan_in(&app_dir, &home, &v027::batched_running_probe(true))
+}
+
+/// Remove every store [`report`] would name.
+pub fn reclaim() -> Result<Outcome> {
+    let app_dir = crate::session::get_app_dir()?;
+    let home = dirs::home_dir().context("home directory unavailable for store reclaim")?;
+    let _lock = guard(&app_dir)?;
+    reclaim_in(&app_dir, &home, &v027::batched_running_probe(true))
+}
+
+/// Hold v027's transition lock for the pass and refuse while that migration
+/// still has work. The lock is exclusive, and `Storage::update` takes it
+/// shared, so no session row can appear or vanish underneath a pass either.
+fn guard(app_dir: &Path) -> Result<crate::session::StorageFlock> {
+    fs::create_dir_all(app_dir)?;
+    let lock = crate::session::acquire_storage_flock(app_dir, v027::LOCK)?;
+    if v027::transition_may_be_pending(app_dir)? {
+        bail!(
+            "the sandbox store migration is still moving stores; run `aoe migrate` and try again"
+        );
+    }
+    Ok(lock)
+}
+
+/// The store ids every profile's registry claims.
+///
+/// Fails rather than answering short. A missing registry file is a profile
+/// with no sessions; a registry that exists but cannot be read, parsed, or
+/// understood is a profile whose sessions we cannot see, and treating its
+/// stores as unowned would delete them.
+fn owned_ids(app_dir: &Path) -> Result<BTreeSet<String>> {
+    let paths = registry_paths(app_dir)?;
+    if paths.is_empty() {
+        bail!(
+            "no session registry under {}; refusing to treat every agent store as unowned",
+            app_dir.display()
+        );
+    }
+    let mut ids = BTreeSet::new();
+    for path in paths {
+        let bytes = fs::read(&path).with_context(|| format!("reading {}", path.display()))?;
+        let value: Value = serde_json::from_slice(&bytes)
+            .with_context(|| format!("parsing {}", path.display()))?;
+        let rows = value
+            .as_array()
+            .with_context(|| format!("{} is not a session array", path.display()))?;
+        for row in rows {
+            let id = row
+                .get("id")
+                .and_then(Value::as_str)
+                .with_context(|| format!("session row without an id in {}", path.display()))?;
+            ids.insert(id.to_string());
+        }
+    }
+    Ok(ids)
+}
+
+/// Every profile's registry, plus the default one. A `sessions.json` that is
+/// present but not a regular file is a registry we cannot read, so it fails
+/// the pass rather than being skipped.
+fn registry_paths(app_dir: &Path) -> Result<Vec<PathBuf>> {
+    let mut dirs = vec![app_dir.to_path_buf()];
+    let profiles = app_dir.join("profiles");
+    match fs::read_dir(&profiles) {
+        Ok(entries) => {
+            for entry in entries {
+                let entry = entry?;
+                if entry.file_type()?.is_dir() {
+                    dirs.push(entry.path());
+                }
+            }
+        }
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+        Err(error) => return Err(error).with_context(|| format!("reading {}", profiles.display())),
+    }
+    let mut paths = Vec::new();
+    for dir in dirs {
+        let path = dir.join("sessions.json");
+        match fs::symlink_metadata(&path) {
+            Ok(metadata) if metadata.is_file() => paths.push(path),
+            Ok(_) => bail!(
+                "{} is not a regular file; refusing to reclaim stores without reading it",
+                path.display()
+            ),
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+            Err(error) => {
+                return Err(error).with_context(|| format!("inspecting {}", path.display()))
+            }
+        }
+    }
+    paths.sort();
+    Ok(paths)
+}
+
+/// Every `sandbox-v2` root any profile can place a store under: the built-in
+/// root per agent config mount, plus the root of each `agent_config_dir` a
+/// profile declares.
+fn store_roots(app_dir: &Path, home: &Path) -> Result<Vec<PathBuf>> {
+    let mut roots: BTreeSet<PathBuf> = BTreeSet::new();
+    let tools = crate::session::config::container_config::agent_config_mount_tools();
+    for tool in &tools {
+        roots.extend(
+            crate::session::config::container_config::sandbox_store_roots(tool, home, None),
+        );
+    }
+    for path in registry_paths(app_dir)? {
+        let profile = profile_of(app_dir, &path);
+        let config = crate::session::config::profile_config::resolve_config_or_warn(&profile);
+        for tool in &tools {
+            let declared = config.session.agent_config_dir_for(tool, home);
+            if declared.is_some() {
+                roots.extend(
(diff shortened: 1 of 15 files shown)


src/session/sandbox_store_reclaim.rs
```diff
@@ -0,0 +1,657 @@
...
+/// Remove every store [`report`] would name.
+pub fn reclaim() -> Result<Outcome> {
+    let app_dir = crate::session::get_app_dir()?;
+    let home = dirs::home_dir().context("home directory unavailable for store reclaim")?;
+    let _lock = guard(&app_dir)?;
+    reclaim_in(
+        &app_dir,
+        &also_owned(&app_dir),
+        &home,
+        &v027::batched_running_probe(true),
```
jerome-benoit: **[P1] Check liveness across both namespaces' runtimes**

Ownership includes debug and release, but this probe uses only the invoking namespace's runtime: `containers/mod.rs:31–39,62,163–176` calls `Config::load()`, whose path is the current app directory (`config/mod.rs:3178–3179`). With release configured for Docker and debug for Podman, deleting a debug session while keeping its container leaves a live Podman store that release reclaim sees as unowned and absent from Docker. Line 333 then removes the mounted credentials/history. Probe every applicable configured runtime, preserving the store if any is live or cannot establish absence. Add a two-namespace, two-runtime regression. This is source-traced, not a runtime reproduction.

src/session/sandbox_store_reclaim.rs
```diff
@@ -0,0 +1,657 @@
...
+/// so no session row can appear or vanish underneath a pass either.
+///
+/// A row that is merely still on the shared store does not block the pass: it
+/// owns no private store to reclaim yet, and the one it will own carries its
+/// id, which this pass reads as claimed. Blocking on that instead would refuse
+/// forever on any machine holding an archived or trashed pre-transition
+/// session, since those keep their shared store until they are started again.
+fn guard(app_dir: &Path) -> Result<crate::session::StorageFlock> {
+    fs::create_dir_all(app_dir)?;
+    let lock = crate::session::acquire_storage_flock(app_dir, v027::LOCK)?;
```
jerome-benoit: **[P2] Protect stores being created before registry insertion**

This lock does not exclude current-generation store preparation. New instances already use the current generation (`instance/accessors.rs:51–52`), so `instance/container.rs:141–149` skips its shared lock. A sandboxed `on_create` calls container preparation before session insertion (`cli/add.rs:1079–1080` versus `1137`); configuration seeds the store before the container exists (`instance/container.rs:241–252`, `container_config.rs:2033–2040`). A concurrent sweep can therefore see that directory with neither a registered owner nor a container and delete it during creation. Establish a shared exclusion/reservation from preparation through ownership publication, including every namespace scanned. Add a barrier-based creation-versus-reclaim regression. Merely rechecking liveness does not cover the pre-container window.

src/session/sandbox_store_reclaim.rs
```diff
@@ -0,0 +1,657 @@
...
+    let plan = plan_in(app_dir, also_owned, home, is_running)?;
+    let mut outcome = Outcome {
+        plan,
+        ..Outcome::default()
+    };
+    for orphan in &outcome.plan.orphans {
+        // The classification above is re-run against the path as it is now:
+        // cheap next to the removal, and the only thing standing between a
+        // swapped directory and `remove_dir_all`.
+        match classify(&orphan.path, &orphan.id, is_running) {
```
jerome-benoit: **[P2] Do not reuse the planning-time container snapshot for deletion**

This reclassification receives the same `batched_running_probe` used by `plan_in`. Its cached listing persists until `LIVENESS_EPOCH` changes (`v027_isolate_sandbox_stores.rs:104–126`), which reclaim never triggers. A retained orphan container listed as stopped can be restarted while other stores are sized; this final check still returns the cached stopped state and deletes its now-mounted store. v027 explicitly documents that limitation at lines 60–64 and relies on non-forced container removal to prevent data loss, a safeguard absent here. Use fresh deletion-time evidence plus an exclusion strategy against restart, or conservatively preserve stores belonging to retained containers. Cover stopped-at-plan, running-before-delete with an injected probe regression.

jerome-benoit (changes requested):
Reviewed all 14 owned files at `a928edd0` against merge base `5673763b`, the description and #3816, with two blind sequential analyses followed by arbitration. Please address the three inline findings: P1 cross-runtime liveness, P2 unregistered store creation, and P2 cached deletion-time liveness.

Verified: failed container teardown preserves the store; parked pre-v027 rows no longer permanently block reclaim; existing sibling-owned IDs are retained. Rejected: the proposed symlink-sizing issue, because [DirEntry::metadata does not follow symlinks](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.metadata). The creation race is rated P2 rather than the proposed P1 because it requires concurrent preparation and affects a newly seeded store.

Evidence: exact-object numbered `git show` and `git diff 5673763b..a928edd0`; `git diff --check` was clean. Failure paths below are source-derived, not executed reproductions. No local builds/tests were run. CI core-rest failed in `managed_capture_lease_serializes_the_physical_store` at `polling.rs:899` (6803 passed, 1 failed); that file is unchanged, but causality is not established. A rerun request was refused because the workflow is still running. The code findings do not depend on that CI failure.

src/session/sandbox_store_reclaim.rs
```diff
@@ -0,0 +1,935 @@
...
+    let mut paths = Vec::new();
+    for dir in dirs {
+        let path = dir.join("sessions.json");
+        // Presence is decided without following, resolution with: a registry
+        // that is there but cannot be resolved to a regular file (a dangling
+        // symlink, a directory) is one we cannot read, and reading none of it
+        // would call its sessions orphans.
+        if fs::symlink_metadata(&path).is_err() {
+            continue;
+        }
```
jerome-benoit: [P1] Keep registry metadata failures fail-closed

This now skips every `symlink_metadata` error, whereas the previous version skipped only `NotFound`. An existing profile whose directory cannot be traversed can therefore lose all its IDs from the ownership inventory without aborting the pass; its accessible, old stores are then eligible for deletion despite still having registered owners. The profile-directory check at 293 similarly drops metadata-resolution errors. This is a new fail-open regression, distinct from the readable-symlink case already fixed in the earlier thread.

Only treat a genuinely absent registry as absent; propagate access and I/O failures with path context. An unresolved profile entry must also prevent an incomplete ownership inventory from authorizing deletion. Add deterministic metadata-error coverage asserting the pass aborts before removing anything. Evidence: `git diff a928edd0 02a45327`, the former NotFound-only match, and the ownership/classification/deletion call chain at 249–276, 362–390 and 455–481. Static finding, no runtime reproduction claimed.

jerome-benoit (changes requested):
Re-reviewed `a928edd0..02a45327`, including all existing threads and the seven-file delta.

**Remaining blockers:**

- **P1, new regression:** registry discovery now suppresses metadata errors instead of failing closed (inline below). Readable symlinked profiles are handled, but an unreadable profile must not become an empty ownership set.
- **P2, partial fix of [the existing liveness thread](https://github.com/agent-of-empires/agent-of-empires/pull/3820#discussion_r3959952945):** `refresh_liveness()` runs once before the whole deletion loop (`451–466`). The first candidate populates the runtime snapshots, and later candidates reuse them while preceding recursive removals run. A retained container that starts during those removals is still read as stopped. Use per-candidate fresh evidence and retain the requested restart exclusion, or conservatively preserve stores attached to retained containers. Cover a later candidate becoming live while an earlier candidate is removed.

**Previous findings:** multi-runtime liveness is addressed; legacy-store wording is corrected. Creation protection is improved by the grace period and ownership reread, but a root-mtime window is not a reservation through registry publication. I am treating that remaining limitation as nonblocking here rather than claiming the invariant is proven or an unusually long preparation was reproduced. The PR description still describes an exclusive transition lock although the implementation now takes it shared; please align that text.

Validation: two independent sequential analyses and arbitration. One reviewer initially accepted the liveness fix, then confirmed the later-candidate gap; the creation limitation was arbitrated as nonblocking without claiming a complete guarantee. Exact-object diff/caller tracing and `git diff --check a928edd0 02a45327` (clean). Static evidence only; no local tests or runtime reproductions. CI status is separate from these findings.

jerome-benoit (changes requested):
Re-reviewed `337502bc`, including the fixes pushed during this round, with two independent sequential analyses and arbitration.

**P2, still blocking: the batch fast path bypasses the new absence-only policy.** This is now also covered by [CodeRabbit's thread](https://github.com/agent-of-empires/agent-of-empires/pull/3820#discussion_r3960951044), so no duplicate inline is added. `sandbox_store_reclaim.rs:160–162` still uses `batched_running_probe_with`; that helper returns `state.is_live()` before calling the existence fallback (`v027_isolate_sandbox_stores.rs:123–133`). `Created`, `Exited` and `Dead` yield `Some(false)` (`containers/runtime.rs:634–638`). Thus an ordinary stopped container is still treated as absent and its store can reach removal (`sandbox_store_reclaim.rs:489–504`), including if it restarts after the check. Make the reclaim batch semantics presence-based for every listed state, without changing migration's liveness semantics. Cover listed stopped containers through the actual composed probe, not only injected final booleans.

**Fixed:** the previous P1 registry metadata failure now propagates errors (`295–307`, `323–333`); evidence is refreshed per candidate (`479`). Cross-runtime coverage and qualified legacy wording remain addressed. The creation grace remains the previously accepted nonblocking limitation, not a complete reservation guarantee.

**P3, nonblocking:** [malformed transition metadata](https://github.com/agent-of-empires/agent-of-empires/pull/3820#discussion_r3960783601) is genuinely swallowed at `v027:381` in PR-owned code. Propagate the parse error and test it. I am not classifying this as a demonstrated data-loss blocker: `owned_ids` still retains the malformed row's ID (`269–274`), excluding its store.

Evidence: exact-object delta `b5f2bf6b..337502bc`, numbered head/base inspection and caller tracing; `git diff --check` is clean. Static findings, no local tests or runtime reproduction. CI is still in progress. The remaining request for changes is the P2 above, not the nonblocking points.

jerome-benoit (approved):
Re-reviewed `337502bc..1f2015ad` with two independent sequential analyses and arbitration. The last P2 is fixed: reclaim now uses `presence_probe`, where every listed container state means present (`sandbox_store_reclaim.rs:160–162,179–190`). Missing entries still reach existence inspection; migration retains its separate `ContainerState::is_live` interpretation (`v027_isolate_sandbox_stores.rs:104–115`). Cache invalidation and fail-closed fallback are preserved.

The regression exercises the composed probe across all seven listed states and absent/unavailable fallback (`sandbox_store_reclaim.rs:917–964`). The P3 is also fixed: malformed transition metadata propagates its error (`v027:405–407`), with coverage at `2168`. No new blocker found. The documented creation-grace limitation remains nonblocking; the other guard's existing `.ok()` is also present in base `b4098e09:357`, not an introduced regression.

Evidence: complete exact-object two-file diff, numbered caller/test inspection, and `git diff --check 337502bc 1f2015ad` (clean). Static review only; no local tests or runtime reproduction. No duplicate inline comments.

CI is not green: [Playwright live shard 2](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34264551087/job/102190538559) fails `live-size-owner-takeover.spec.ts:115`, cursor delta `76.75` versus `< 2`, the same assertion/value observed on #3782. Other jobs are still running. This approval covers the reviewed code, not a claim that the CI failure is transient or resolved.

PR #3821, opened 2026-09-08
fix(migrations): stop the v027 store move expanding and re-flushing everything

### LINKED ISSUE #3819: v027 store move is fsync-bound and copies physically: 8.6 MB/s, and N sessions multiply a 3.3 GB store into 13 GB
The v027 sandbox-store move copies every file physically and calls `sync_all()` on each one (`v027_isolate_sandbox_stores.rs:1789`). On macOS `sync_all()` is `F_FULLFSYNC`, a full drive cache flush. Measured on an M-series APFS machine:

```
no sync                200 files in 0.12s  ->  0.58 ms/file
os.fsync (data)        200 files in 0.12s  ->  0.58 ms/file
F_FULLFSYNC            200 files in 1.46s  ->  7.29 ms/file
```

Observed `aoe migrate` throughput on that machine: 170 files/s, 8.6 MB/s, process pinned in uninterruptible I/O wait at 15% CPU. At 7.29 ms each, 170 files/s is about 1.24 seconds of flushing per wall-clock second, so the barrier is the whole runtime. The run passed 35 minutes and 290,000 files without finishing.

Separately, each session gets a full physical copy. A 3.3 GB shared Codex store across seven sessions became 13 GB of `sandbox-v2` mid-run and was still growing; free space on the volume went from 158 GB to 95Gi during one migration. The root filesystem is APFS and nothing in the module calls `clonefile`, `FICLONE` or `copy_file_range`.

The cost lands on session start, not just on `aoe migrate`: a sandboxed session moves its own store when it starts, so the first Codex sandbox launch after upgrading blocks on gigabytes at 8.6 MB/s with no indication beyond a session that will not come up.

The per-file barrier also looks stronger than the design needs. This module already recovers from a torn copy, covered by `recovers_publication_before_registry_commit` and `recovers_legacy_quarantine_before_generation_commit`. Durability before the atomic publish is the requirement, which one `sync_data` pass plus a directory fsync ahead of the rename satisfies.

Suggested, in order of payoff: clone on APFS with `clonefile`, which removes both the time and the space multiplication; `F_BARRIERFSYNC` or `sync_data` in place of `sync_all`; and fsync once before publishing rather than per file. A reflink path would apply to Linux btrfs and XFS through `FICLONE` as well.

njbrake:
## Description

When you upgrade, AoE gives every sandboxed session its own copy of the agent's files instead of one shared folder. That move was doing far more work than it needed to. It handed each session a full copy of the whole shared folder, including other sessions' conversation history, caches, logs and plugin trees, and it did that even for sessions deleted long ago. It also asked the disk to fully flush itself after every single file, which on a Mac is the slowest thing you can ask a drive to do. The reported run took 43 minutes, wrote 16 GB, and produced 63 folders belonging to no session at all.

This changes what gets copied before it changes how fast copying is. A folder no session claims is moved rather than duplicated. A session that already had its own folder gets the shared folder's loose files, its login and settings, and not its subfolders. Nothing skipped is deleted: the old shared folder stays put and AoE tells you where it is. Then files are shared on disk rather than copied where the filesystem allows, and the per-file flush becomes one write barrier before each folder is published.

Fixes #3819

### Scope, first

- **A store no registry row names is renamed into the private layout, not copied.** All 63 stores the reported run created resolved to no session in any profile. Renaming preserves one exactly for one syscall, and nothing will start it, so the shared agent home is not folded in.
- **The overlay carries top-level files, not directories.** It folds a shared agent home into a session that already had a private one. That session needs the credentials, config and state kept at that home's root. Its directories are the home's own accumulation: 109 MB of other sessions' history, 26 MB of plugins, 16 MB of cache, 5.9 MB of shell snapshots. None of it is the SQLite single-instance lock this migration exists to unshare, and it turned a 17 MB legacy store into a 252 MB private one.
- **What no private store received is not deleted.** Retirement removes the per-instance children and the top-level files every private store did receive; the legacy root keeps the rest and the migration names it, since it has become the only copy. A `~/.gemini/sandbox`-style root, where the shared store genuinely was each session's working state, is still retired whole.

### Then cost

- **Copy-on-write clones** through `clonefile` on APFS and `FICLONE` on btrfs and XFS. Every clone failure falls back to writing the bytes, not only the filesystem's refusal, and one refusal turns clones off for the rest of the move rather than paying a failing syscall per file.
- **One barrier per publish instead of a full flush per file.** `File::sync_all` is `F_FULLFSYNC` on Apple platforms, measured in the issue at 7.29 ms/file against 0.58 ms for plain `fsync`. Each entry is pushed to the drive with `fsync` as it is created; one `F_BARRIERFSYNC` orders the tree ahead of the publishing rename.

### Durability is unchanged

The requirement is that the staged tree reaches the media before the rename that publishes it. `fsync` moves bytes host-to-drive, the barrier forbids the drive committing anything after it before everything issued before it, and the existing `sync_all` on the parent after the rename makes the publish durable. A crash can lose a publish; it cannot expose a published store whose bytes never reached the media. A clone is metadata only, and its `fsync` persists the new inode's extent map; the extents it shares are the source's, which the same barrier covers.

`recovers_publication_before_registry_commit` and `recovers_legacy_quarantine_before_generation_commit` still run through the same staging protocol: `Publication::prepare` is the old prologue, extracted so the rename path clears the debris the copy path did. Removing a legacy child still renames it aside first, because a half-deleted legacy store is one a still-pending row would copy back over the published one.

### Measured

`aoe migrate` over a store built from a real tree (10,867 files, 163.8 MB, median file 3.8 KB, a slice of `~/.cargo/registry/src`) as the shared agent home, plus nine per-instance children of which one is registered. Debug build, Linux aarch64 container, overlayfs over ext4, medians of three paired runs.

| | before | after |
|---|---|---|
| elapsed | 50.7 s | 3.4 s |
| published | 100,818 files / 1495.5 MB | 2,999 files / 4.7 MB |
| files/s | 1989 | 874 |
| MB/s | 29.5 | 1.4 |

Throughput falls because almost nothing is left to write: 318x fewer bytes, 15x shorter wall clock. The remaining 3.4 s is directory scanning and the one registered store.

The copy path alone, same tree as a `~/.gemini/sandbox` shared store that must be copied whole, five paired runs, medians: before 6.32 s / 1720 files/s / 25.9 MB/s, after 6.44 s / 1688 files/s / 25.5 MB/s. No measurable change, which is all this machine can show: `fsync` and `sync_all` cost the same on Linux, and `ioctl(FICLONE)` here returns `EOPNOTSUPP (95)`, so the fallback is what ran for all 10,867 files and produced trees identical to the old code's. **The clone win is argued here, not measured.**

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt`, `cargo clippy --all-targets`, `cargo test` (6798 + 289 pass). `docs/guides/sandbox.md` said the shared store is deleted once every session has moved; updated.

New tests: `an_unregistered_store_moves_without_being_expanded`, `a_fully_replicated_shared_root_is_still_retired`, `retiring_a_root_keeps_only_what_no_store_received`, `an_interrupted_retirement_keeps_the_published_store`, `a_retired_root_does_not_fail_the_next_row`, `an_interrupted_stage_is_cleared_when_an_unregistered_store_moves` (the reported leftover `.v027-stage-<id>`, on the rename path that never opens it), `an_unclonable_source_falls_back_instead_of_failing` and `a_refused_clone_is_not_retried` (a source no clone implementation accepts, so the fallback is covered on a reflink-capable filesystem too).

`ci-macos` is labelled on this PR: the macOS job is gated behind it, and the `clonefile` and `F_BARRIERFSYNC` paths are the point of the change.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the in-module tests drive the whole pass (`run_in`) with the container probes injected, which is the cheapest test that catches these regressions. A full-binary run would reach nothing they do not.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Investigation, implementation and benchmarking by the model in discussion with @njbrake; the scope decision, that the fix is what gets copied rather than how fast it copies, is his.

https://claude.ai/code/session_01CvnVeqAwMg7iVZc2rRpKWZ

## Summary

- Updated v027 migration to rename unregistered stores instead of copying them.
- Copies only required top-level files into existing private stores.
- Preserves unrelated caches, logs, histories, and plugin data.
- Added copy-on-write cloning with a byte-copy fallback.
- Replaced repeated full-drive flushes with staged syncing and one publish barrier.
- Improved recovery for interrupted staging and retirement.
- Updated documentation and migration tests.

## Benefits

- Reduces migration time and disk usage.
- Preserves data that does not require migration.
- Improves safety during interrupted migrations.

## Technical follow-up

- Continue validating clone support and recovery behavior across supported filesystems.

### DIFF
--- a/docs/guides/sandbox.md
+++ b/docs/guides/sandbox.md
@@ -319,8 +319,9 @@ being replicated into every session, so if any are left the shared store is
 kept rather than deleted and AoE names it when the move finishes; remove it
 yourself once you no longer want it. A store with nothing left in it is
 deleted as before. A large store takes a while, so the first start of a
-session is slower than usual; the TUI shows the copy's progress on its status line and opens the
-session once it is done, and a plain `aoe` start says how many sessions still
+session is slower than usual; the TUI shows the copy's progress on its status
+line and opens the session once it is done, and a plain `aoe` start says how
+many sessions still
 have the move ahead of them. A session whose container is still running is
 skipped and moved on a later start, after it stops. Trashed and archived sessions stay on the shared
 store. Starting one moves it; restoring or unarchiving alone does not, so run
--- a/docs/guides/sandbox.md
+++ b/docs/guides/sandbox.md
@@ -311,8 +311,14 @@ example `~/.claude/sandbox`). Each one moves when you start it: AoE copies the
 shared store into that session's private directory, removes its stopped
 container so the next launch mounts the copy, and deletes the shared store once
 every session that used it has moved (the private copies are the data from then
-on). A large store takes a while, so the first start of a session is slower
-than usual; the TUI shows the copy's progress on its status line and opens the
+on). For an agent whose sessions already had a private store of their own, the
+copy takes the shared store's top-level files, its credentials, config and
+state, and not its directories: caches, logs, plugin trees and conversation
+history belonging to no one session. Those stay where they are rather than
+being replicated into every session, so that shared store is kept instead of
+deleted and AoE names it when the move finishes; remove it yourself once you
+are sure you no longer want it. A large store takes a while, so the first
+start of a session is slower than usual; the TUI shows the copy's progress on its status line and opens the
 session once it is done, and a plain `aoe` start says how many sessions still
 have the move ahead of them. A session whose container is still running is
 skipped and moved on a later start, after it stops. Trashed and archived sessions stay on the shared
--- a/docs/guides/sandbox.md
+++ b/docs/guides/sandbox.md
@@ -315,10 +315,11 @@ on). For an agent whose sessions already had a private store of their own, the
 copy takes the shared store's top-level files, its credentials, config and
 state, and not its directories: caches, logs, plugin trees and conversation
 history belonging to no one session. Those stay where they are rather than
-being replicated into every session, so that shared store is kept instead of
-deleted and AoE names it when the move finishes; remove it yourself once you
-are sure you no longer want it. A large store takes a while, so the first
-start of a session is slower than usual; the TUI shows the copy's progress on its status line and opens the
+being replicated into every session, so if any are left the shared store is
+kept rather than deleted and AoE names it when the move finishes; remove it
+yourself once you no longer want it. A store with nothing left in it is
+deleted as before. A large store takes a while, so the first start of a
+session is slower than usual; the TUI shows the copy's progress on its status line and opens the
 session once it is done, and a plain `aoe` start says how many sessions still
 have the move ahead of them. A session whose container is still running is
 skipped and moved on a later start, after it stops. Trashed and archived sessions stay on the shared
--- a/src/migrations/mod.rs
+++ b/src/migrations/mod.rs
@@ -9,6 +9,7 @@
 //! 3. Add it to the `MIGRATIONS` array below
 
 pub mod progress;
+mod store_fs;
 mod v001_xdg_linux;
 mod v002_seed_sandbox_from_volumes;
 mod v003_yolo_mode_config;
--- a/src/migrations/store_fs.rs
+++ b/src/migrations/store_fs.rs
@@ -1,0 +1,235 @@
+//! Filesystem primitives for the v027 sandbox store move.
+//!
+//! The move stages a tree, makes it durable, then publishes it with one
+//! rename. Two costs dominated that: it wrote every byte of every file, and it
+//! forced a full drive cache flush per file. Both have cheaper answers that
+//! keep the guarantee the publish actually needs, which is that the staged
+//! tree is on the media before the rename is.
+
+use std::fs;
+use std::io;
+use std::path::Path;
+
+/// Whether this move may still try a copy-on-write clone.
+///
+/// A filesystem pair answers the same way for every file in one tree, so the
+/// first refusal turns clones off for the rest of the move instead of paying
+/// a failing syscall per file.
+#[cfg(unix)]
+#[derive(Default)]
+pub(super) struct CloneSupport {
+    refused: bool,
+}
+
+#[cfg(unix)]
+impl CloneSupport {
+    /// Create `target` as a copy-on-write clone of the open regular file
+    /// `source`, described by `stat`. `target` must not exist.
+    ///
+    /// `None` means the caller must copy the bytes itself; nothing is left at
+    /// `target`. Every failure answers that way, not only a filesystem's
+    /// refusal: `clonefile` fails across filesystems and on non-APFS volumes,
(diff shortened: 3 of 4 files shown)


jerome-benoit (approved):
Reviewed the full four-file change and revalidated `ac621394..8947cc6e` with two independent sequential analyses and arbitration. The missing-root recovery P2 found during review was fixed before publication: `retire_legacy_children` now handles `NotFound` through the existing quarantine cleanup (`v027_isolate_sandbox_stores.rs:2075–2085`), and the new Codex test covers another pending row after root retirement (`3686–3717`). No remaining blocker found.

Checked the selective overlay/retention scope against #3819, orphan relocation and copy fallback, interrupted publication, and symlink handling. The new fallback restores a quarantined destination before copying. Speculative clone-metadata failure scenarios were not promoted to findings. Nonblocking cleanup: the platform implementations in `store_fs.rs` should follow the project's `src/process/{macos,linux}.rs` placement rule; the remaining implementation-history/benchmark comments can also be shortened.

Evidence: complete exact-object diff and numbered caller inspection; `git diff --check ac621394 8947cc6e` passes. Static review only, no local tests or runtime reproduction. CI for this new SHA is still running; the previous SHA's failures are not results for this revision.

jerome-benoit (approved):
Re-reviewed `8d75652b` with two independent sequential analyses and arbitration. `git range-diff 4f083edd..8947cc6e b4098e09..8d75652b` confirms the three previously reviewed commits are patch-identical after rebase.

The additional portable-copy change (`v027_isolate_sandbox_stores.rs:2015–2028`) correctly refuses a conflicting destination instead of silently skipping the overlay file. The missing-root recovery and its regression test remain intact (`2082–2093`, `3693–3724`). No new blocker found.

Explicit classification of [CodeRabbit's missing-source overlay finding](https://github.com/agent-of-empires/agent-of-empires/pull/3821#pullrequestreview-5145277972): valid pre-existing data-loss path, nonblocking for this PR, not dismissed as invalid. Actual base `b4098e09` already skips the overlay when the source is absent (`1587–1590`) and retires the shared root (`1190–1192`). Follow-up should preserve/copy the overlay before retirement and cover a missing private source with shared credentials. No duplicate inline comment added.

Evidence: numbered head/base `git show`, the range-diff above, and `git diff --check b4098e09 8d75652b` (clean). Static review only; no local tests or reproduction. CI remains in progress, with no green-suite claim.

PR #3823, opened 2026-09-08
fix: get main green on the sun_path and capture-lease regressions

njbrake:
## Description

Main has been red for 16 commits on two unrelated regressions.

**`runner.rs:3114`, macOS only.** `watchdog_teardown_preserves_replacement_registry_owner` binds the runner control socket under `isolate_app_dir()`, which roots the app dir at `$TMPDIR`. macOS resolves that to `/var/folders/<uid hash>/T/`, so `<app_dir>/acp-workers/<id>.control.sock` runs ~127 bytes against the 104-byte `sun_path` cap and `bind` fails with `path must be shorter than SUN_LEN`. Rooted under `/tmp` instead, matching the convention already used in `worker_registry` and `acp::supervisor` tests. Reproduced on Linux by pointing `TMPDIR` at a macOS-length path, which fails identically before this change and passes after.

**`polling.rs:899`, macOS and bare-core, roughly one run in three.** `/proc/locks` at the failure shows the exclusive flock still held by our own pid after the owning `File` was dropped, while `/proc/self/fd` holds no descriptor for it. A process forked elsewhere in the suite inherited the descriptor before its `exec` cleared it. A flock belongs to the open file description, not to one descriptor, so closing the owner's copy releases nothing while an inherited copy is alive. `Drop` now unlocks the description itself, which releases every inherited copy with it. This is why the failure always landed on the final assertion, the only one that re-acquires after a release.

That path also collapsed "could not resolve the app dir", "could not canonicalize the store" and "another owner holds this" into one `None` through an `.ok()?` chain, so a transient local failure was reported to the user as another process owning their store. It now returns a typed refusal and logs the two cases apart. The test asserts the specific refusal, so a future failure names its cause instead of reporting a bare `is_none()`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Both regressions are covered by the existing unit tests they broke. The lease fix was verified by looping the test body 3000 times inside the full parallel suite: it failed within ~100 iterations before the change and passes 3000 after. Local runs: `cargo fmt`, `cargo clippy --all-targets` clean, `cargo test` green (6790 lib, 289 integration).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Root causes were reproduced locally before either fix was written, not inferred from CI logs.

https://claude.ai/code/session_0176k3KdBmw1AfEiGvz4DVqt

## Summary

- Shortened the macOS watchdog test app path to keep its control socket within system limits.
- Improved capture-lease cleanup so inherited file locks are released correctly.
- Separated path and store errors from genuine ownership conflicts.
- Added clearer logging and test coverage for lease handling.

These changes prevent test failures and allow resources to be reacquired reliably after release. Validation passed with formatting, Clippy, and the full test suite.

### DIFF
--- a/src/process/runner.rs
+++ b/src/process/runner.rs
@@ -3107,7 +3107,11 @@ mod tests {
     #[tokio::test]
     #[serial_test::serial]
     async fn watchdog_teardown_preserves_replacement_registry_owner() {
-        let _app_dir = crate::session::test_support::isolate_app_dir();
+        // Rooted under /tmp, not $TMPDIR: macOS resolves the latter to
+        // /var/folders/<uid hash>/T/, which leaves the control socket past the
+        // 104-byte sun_path limit once <app_dir>/acp-workers/<id> is appended.
+        let app_root = tempfile::TempDir::with_prefix_in("aoe-wd-", "/tmp").unwrap();
+        let _app_dir = crate::session::test_support::isolate_app_dir_at(app_root.path());
         let session_id = "watchdog-replacement";
         let socket = worker_registry::socket_path_for(session_id).unwrap();
         let control_socket = crate::process::worker::control_socket_sibling(&socket);
--- a/src/session/instance/polling.rs
+++ b/src/session/instance/polling.rs
@@ -29,20 +29,55 @@ pub enum PollerStart {
     SpawnFailed,
 }
 
+/// An exclusive claim on one physical capture store, held for as long as the
+/// poller that owns it runs.
+///
+/// `Drop` unlocks explicitly rather than letting the descriptor close do it. A
+/// `flock` belongs to the open file description, not to one descriptor, so a
+/// process forked while the lease was open holds the lock alive through its
+/// inherited copy until `exec` clears it. Releasing by close alone leaves the
+/// store reading as owned by someone else for that window; `unlock` releases
+/// the description itself, and every inherited copy with it.
+#[derive(Debug)]
+struct ManagedCaptureLease(std::fs::File);
+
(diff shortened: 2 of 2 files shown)


jerome-benoit (approved):
Reviewed the two-file change at `3579f0ce` with two independent sequential analyses and arbitration. No introduced blocker found.

The lease wrapper explicitly unlocks on drop (`polling.rs:42–48`) and stays owned by the poller closure (`422–429`); I found no premature release. This matches the documented shared-open-description semantics of [flock](https://man7.org/linux/man-pages/man2/flock.2.html). The Unix-only watchdog test now uses a short, caller-owned temporary root (`runner.rs:3113–3118`), whose lifetime encloses the app-dir guard.

**Nonblocking diagnostic follow-up:** [the existing bot comment](https://github.com/agent-of-empires/agent-of-empires/pull/3823#discussion_r3960466055) is valid: not every lock-system-call error means contention. One reviewer initially rated it blocking; comparison with the base confirmed that the false warning and retry behavior already existed, so arbitration retained it as nonblocking. It does not invalidate the two root-cause fixes. No duplicate inline comment.

Evidence: complete exact-object diff, numbered lifetime/caller inspection, and `git diff --check 4f083edd...3579f0ce` (clean). Static review only; the reproduction/stress-loop results in the description are the author's, not mine. CI is still running, so this approval does not claim globally green checks.

PR #3825, opened 2026-09-08
fix(sandbox): keep the agent store through any purge that gets rolled back

njbrake:
## Description

Deleting a sandboxed session for good was supposed to remove the private folder that session used as its agent's home, since it holds that agent's saved login and history. #3820 added that, and guarded it so a session whose container could not be torn down keeps its folder.

That guard covered only the container step. Deleting a session is a sequence: kill tmux, remove the container, remove the worktree, delete the branch, clean up scratch. If *any* of those later steps fails, the whole delete is rolled back and the session stays. The folder had already been removed by then, so the session came back alive but logged out, with its history gone, and every retry did the same thing again.

The ordinary way to hit this is to delete the repo directory a session lives in and then purge the session: worktree removal fails with "Path is not in a git repository", the delete rolls back, and the credentials are already gone. A failed `git branch -D` does it too.

Removing the folder now happens last, and only when nothing else has failed. Anything stranded that way is picked up by `aoe sandbox reclaim`, which exists for precisely that.

A failure to remove the folder is also no longer treated as an error in its own right. Reporting one rolled the delete back and kept a session whose container was already gone, and it would have failed the same way forever.

The dashboard path was already safe: it commits the row before teardown and never rolls back. This is the CLI and TUI path.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Covered by a unit test rather than an e2e one: `a_purge_that_fails_after_teardown_leaves_the_store` drives `perform_deletion_core` through its real stage sequence with a worktree that fails, and asserts the store survives. It fails without the reordering. An e2e test would need a real container runtime to reach the same stage.

`cargo fmt`, `cargo clippy --all-targets` and `cargo test` all pass.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Found by a fresh-context review of #3820 after it was approved, and reproduced with a failing test before the fix was written.

Follow-up to #3820.

## Summary

- Preserve an agent’s private store when any sandbox purge stage fails.
- Remove the store only after teardown and cleanup complete successfully.
- Report store-removal failures without rolling back the purge.
- Document recovery through `aoe sandbox reclaim`.
- Add a regression test for worktree-removal failure.
- Verify formatting, Clippy, and tests.

## Benefits

This prevents credential and history loss after partial purge failures. Operators can reclaim leftover stores after resolving the underlying issue.

### DIFF
--- a/src/session/deletion.rs
+++ b/src/session/deletion.rs
@@ -720,18 +720,15 @@ fn perform_deletion_core(
     // Stage 3: container removal. Releases the bind mount on the
     // worktree so the host can finish cleanup without racing in-
     // container processes.
+    let mut container_gone = false;
     if request.delete_sandbox && is_sandboxed {
         tracing::debug!(target: "session.delete", session_id = %request.session_id, stage = "container_remove", "perform_deletion: stage");
         let outcome = teardown(&request.instance.id);
         // A failed teardown can leave the container live with the store still
-        // bind mounted, and aborts the purge, so the session keeps running on
-        // the store: only remove it once the container is provably gone. One
-        // stranded by a failed purge is the reclaim pass's job.
-        let container_gone = !matches!(outcome, crate::containers::Teardown::Failed(_));
+        // bind mounted, so the store may only go once the container is
+        // provably gone.
+        container_gone = !matches!(outcome, crate::containers::Teardown::Failed(_));
         deletion_messages_for(outcome, &mut messages, &mut errors);
-        if container_gone {
-            stage_remove_agent_stores(request, &mut messages, &mut errors);
-        }
     }
 
     stage_remove_worktrees_and_branches(
@@ -744,6 +741,14 @@ fn perform_deletion_core(
     );
 
     stage_cleanup_scratch(request, &mut errors, &mut messages);
+
+    // Last, and only when nothing else failed: any error here rolls the purge
+    // back (`PurgeTransaction::complete_inner`), and a session that survives
+    // its own purge must survive with the store holding its login and history.
+    // One stranded by a rolled-back purge is the reclaim pass's job.
+    if container_gone && errors.is_empty() {
+        stage_remove_agent_stores(request, &mut messages);
+    }
 
     // Stage 6: hook status cleanup
     tracing::debug!(target: "session.delete", session_id = %request.session_id, stage = "hook_status_cleanup", "perform_deletion: stage");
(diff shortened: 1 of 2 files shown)


src/session/deletion.rs
```diff
@@ -745,6 +742,14 @@ fn perform_deletion_core(
 
     stage_cleanup_scratch(request, &mut errors, &mut messages);
 
+    // Last, and only when nothing else failed: any error here rolls the purge
+    // back (`PurgeTransaction::complete_inner`), and a session that survives
+    // its own purge must survive with the store holding its login and history.
+    // One stranded by a rolled-back purge is the reclaim pass's job.
+    if container_gone && errors.is_empty() {
+        stage_remove_agent_stores(request, &mut messages);
```
jerome-benoit, line 750: [P2, nonblocking follow-up] The outer transaction can still retain the row after this removes the store. `complete_inner` calls this teardown at line 403, then `after_teardown` at 410; a callback error releases the reservation and keeps the row (411–415). Both CLI purge paths use that callback for `purge_acp_transcript` (`src/cli/remove.rs:247–253`, `src/cli/session.rs:695–698`), which returns database errors even for terminal sessions (`src/cli/mod.rs:119–168`). The subsequent durable commit can also fail (421–452). In those cases the retained session has already lost its private store.

This is pre-existing in `8f0cece8`, not a regression introduced here, so it does not block the useful teardown-stage fix. A follow-up should perform best-effort store removal only after confirmed durable row removal, retaining the lifecycle lock, and test callback/commit failure at the transaction boundary. The new core-level test does not cover that boundary. Verified by inspecting the exact head/base with `git show`; not runtime-reproduced.

jerome-benoit (approved):
Reviewed `0316ef95` against `8f0cece8`, with two independent sequential analyses and cross-validation. The worktree/branch/scratch failure fix is sound; keeping store-cleanup failures nonfatal is appropriate.

One residual P2 is noted inline, explicitly **nonblocking**: it already exists in the base, and this patch reduces rather than introduces the risk. Please scope the description to teardown-stage failures, not every possible purge rollback. CodeRabbit's [documentation correction](https://github.com/agent-of-empires/agent-of-empires/pull/3825#discussion_r3961496300) is also valid, nonblocking P3; no duplicate annotation.

Static inspection only; `git diff --check 8f0cece8 0316ef95` passed. No local tests or reproduction run. CI has no failures at this check, but bare-core tests and Rust analysis are still running.

PR #3827, opened 2026-09-08
fix(vt): serialize concurrent resize declarations and pin the reseed link ordering

### LINKED ISSUE #3817: Concurrent resizes can clear another viewer's VT resync gate
# Concurrent resizes can clear another viewer's VT resync gate

## Problem

After merged #3778, overlapping resize declarations can clear the shared grid-publication gate before the parser catches up with the pane. This is a P2 stale-layout defect, not the previously fixed unbounded DEC 2026 freeze.

Two cases affect the same resize-state contract:

- `begin_resize` increments an odd/even sequence without serializing writers. Two live guards make it even, so an overlapping geometry probe can be accepted as quiescent.
- Declarations use packed geometry as their identity. An older rejected resize to the same dimensions can abandon a newer owner's still-pending resize. This is already reported in [the CodeRabbit thread](https://github.com/agent-of-empires/agent-of-empires/pull/3778#discussion_r3958458803).

## Deterministic regression cases

These are source-derived test cases, not tests executed during this review.

1. Start with pane and parser at A. The old owner opens a guard for B and pauses before ownership verification. A new owner takes over and opens its own B guard. A follower records the now-even sequence and reads pane A. Let the new owner resize to B and return `Busy` from its reseed, but pause before its guard drops. Complete the follower's observation with A and the recorded sequence. The pending expectation must remain; the current equality/parity check clears it.
2. Repeat the ownership takeover and identical B declarations. Let the new owner's resize succeed and its reseed return `Busy`; its guard may now drop. Resume the old caller, which fails ownership verification and abandons its older guard. The new owner's B expectation must remain; the current geometry-based CAS clears it.

Use barriers and injected probe/resize/reseed results, not sleeps or live tmux races. Assert that a follower remains gated after the failed reseed, then resumes after a successful reseed.

## Expected behavior and fix direction

Make declaration identity, active operations and expectation retirement one coherent transaction. A rejected declaration must not erase another owner's outstanding work. Probes must not retire state that changed while they were in flight. Merely checking a separate generation and then unconditionally clearing the target leaves another check/clear race.

## Evidence and scope

Reviewed PR head: `796e6c516a119e993a14570981312551975c64a1`.

- [Resize declaration and observation](https://github.com/agent-of-empires/agent-of-empires/blob/796e6c516a119e993a14570981312551975c64a1/src/tmux/vt.rs#L2389-L2466).
- [Guard begins before the ownership check](https://github.com/agent-of-empires/agent-of-empires/blob/796e6c516a119e993a14570981312551975c64a1/src/server/live_ws.rs#L347-L355).
- [Follower publication gate](https://github.com/agent-of-empires/agent-of-empires/blob/796e6c516a119e993a14570981312551975c64a1/src/server/live_ws.rs#L1192-L1206).

Rechecked on `main` at `5673763b8210c053ce30cfda2e4d20051bcf548f`: the same parity and token logic remains (`vt.rs:2528,2578,2589`). Numbered `git show` confirms the ordering; `git diff 99e93bf6 5673763b -- src/tmux/vt.rs src/server/live_ws.rs` shows only an unrelated test-fixture field addition. No local build or runtime reproduction was performed.

Follow-up to closed #3772, limited to the residual concurrent-declaration cases. #3737 concerns replay of unread pipe output, not resize-gate ownership.

## Coordination with #3759

Update after #3759 integrated main at `b886d6af`: the resize guards are retained (`vt.rs:2870-2932`), contrary to the earlier pre-integration description. A changed-size operation now retires the channel (`2844-2852`), neutralizing the persistent-after-`Busy` scenarios above. Do not demand a parallel gate redesign merely to fix an obsolete reseed path. Before closure, distinguish that resolved scenario from any demonstrable pre-retirement publication window; retention of the counters alone is not proof of a remaining user-visible defect. The PR is still open. [Cross-validated rereview](https://github.com/agent-of-empires/agent-of-empires/pull/3759#pullrequestreview-5143319234).

## AI usage

Analysis and text prepared with Codex under @jerome-benoit's direction. Publication authorized by @jerome-benoit; no human-run reproduction is claimed.

### LINKED ISSUE #3818: VT snapshot reseeding can erase a newly received hyperlink target
# VT snapshot reseeding can erase a newly received hyperlink target

## Problem

The reseed path replaces the hyperlink table while holding the parser lock, but the pipe reader records new OSC 8 targets before taking that lock. A reader arriving after the reseed's one-time counter check can therefore have its new URI erased before its linked text is applied.

Impact: P2, a newly received link label can remain visible but lose its clickable target until the application re-emits the OSC 8 sequence or a later accepted authoritative reseed captures it. An ordinary preview refresh cannot restore the entry. No text-data loss or unbounded freeze is established.

## Proposed deterministic regression test

This is a source-traced interleaving and proposed test, not an executed runtime reproduction.

1. Begin a reseed from a snapshot captured before the new `docs` chunk arrives.
2. Pause the reseed after it acquires the parser lock and accepts its generation/chunk guards, before replacing the link table.
3. Let the reader receive an OSC 8 chunk with label `docs` and URI `https://example.com/new`. It increments the received counter and records the URI, then blocks on the parser lock.
4. Resume the reseed. Its table replacement removes that URI because the old snapshot does not contain it.
5. Release the parser lock and let the reader apply its chunk. Sample the grid and query `pane_links`: assert both the visible label and its URI are present. Current code applies the label without recording the erased URI again.

Use barriers around the real reader and swap paths, not sleeps. Also retain coverage that rejected seeds leave links untouched and accepted plain-text snapshots remove old links.

## Expected fix

Order the reader's link metadata update with its parser-byte application under the same parser lock, using the same parser-to-link-table lock order as the reseed. A seed-table replacement must not interleave between recording a chunk's URI and applying its bytes.

## Evidence and scope

Merged PR #3749 was reviewed at `598498df3e5a9455bcb753d919c4a366dea2f74d`; see the [existing P2 review](https://github.com/agent-of-empires/agent-of-empires/pull/3749#pullrequestreview-5141954250). Rechecked: the code remains affected in `main` at `5673763b8210c053ce30cfda2e4d20051bcf548f`:

- [Parser lock, one-time guards and link-table replacement](https://github.com/agent-of-empires/agent-of-empires/blob/5673763b8210c053ce30cfda2e4d20051bcf548f/src/tmux/vt.rs#L894-L914).
- [Reader records targets before acquiring the parser lock](https://github.com/agent-of-empires/agent-of-empires/blob/5673763b8210c053ce30cfda2e4d20051bcf548f/src/tmux/vt.rs#L1636-L1665).

Evidence: numbered `git show 5673763b:src/tmux/vt.rs` confirms the stated ordering; `git diff 99e93bf6 5673763b -- src/tmux/vt.rs` shows only an unrelated test-fixture field addition. Two independent sequential reviews and parent arbitration previously confirmed this unchanged source interleaving. No local build or runtime reproduction was performed in this cycle.

This is not #3519's missing browser-side OSC 8 parsing, nor #3737's replay of unread pipe output. It is a follow-up to the reseed-table replacement introduced while fixing #3735.

## Coordination with #3759

Update after #3759 integrated main at `b886d6af`: the shared snapshot mutex is acquired by initial seeding (`vt.rs:1104`) and by the reader before receipt (`1948`), and retained through link recording and parsing (`2009-2017`). The documented interleaving is therefore excluded; active-grid reseeds retire the channel instead. This is a candidate for closure when that integration merges, not a request for a parallel parser-lock fix. [Cross-validated rereview](https://github.com/agent-of-empires/agent-of-empires/pull/3759#pullrequestreview-5143319234). No issue-specific runtime test was executed. #3737 tracks unread-byte replay, not hyperlink metadata consistency.

## AI usage

Analysis and text prepared with Codex under @jerome-benoit's direction. Publication authorized by @jerome-benoit; no human-run reproduction is claimed.

njbrake:
## Description

Two P2 reports against the VT reseed path, filed as fallout from #3778 and #3759. Both were source-derived and unexecuted, so each was reproduced (or not) against `main` first.

**#3817 is real; both cases reproduce.** `resync_target` and `resize_seq` were separate atomics, so two viewers resizing the same pane could each retire the other's work:

- The seqlock counted resizes by parity. Two overlapping declarations make it even again, and a follower's geometry probe reads that as quiescent and clears the publication gate with both resizes still running.
- A declaration's identity was its packed geometry. In an ownership handover both callers declare the same size, so the loser's withdrawal CASes away the winner's still-pending expectation.

A third case is the mirror of the second and turned up in review: both viewers declare before either learns who owns the pane size, so the one that declared second is as likely to be the one that turns out not to own it, and naming the declaration alone does not stop its withdrawal dropping an older, still-running resize's gate. Either way the gate drops while the parser still holds the layout the pane has left, and every viewer of the shared channel publishes it. Fixed by folding the three concerns into one `ResizeState` behind a mutex: the geometry owed, a monotonic token naming the declaration that owes it, a count (not a parity) of resizes in flight, and an epoch bumped by every declaration and every resize that finishes. A probe retires an expectation only when nothing moved across it and nothing is moving now; a withdrawal names its own declaration and takes effect only when no other resize is still in flight; the reached-geometry check reads and clears under one lock.

**#3818 does not reproduce on `main`.** Its step 3, the reader having recorded a chunk's URI and blocked on the parser lock while a seed sits between its guards and the table replacement, is not a reachable state. `run_reader` holds the snapshot mutex from before `recv` through the parse, and `install_seeded_parser` holds the same mutex from before its drain through `reconcile_links`, so the two never overlap. #3759 closed it when it extended that mutex over the reader's receipt; the issue's own coordination note anticipated this. A second, independent guard also covers it: the seed's received/settled baseline turns any swap that raced a newly received chunk into `Busy` before it touches the table.

Nothing pinned either, though, and that is the unresolved collision between #3749 (link reconciliation atomic inside the seed swap) and #3759 (which restructured that path). The whole guarantee rests on where two `let` bindings sit, one per side, and the chunk guard hides a regression in either. So the second commit lands the ordering as tests rather than a fix.

### Defined behavior for pane links across a reseed

Written down in `reconcile_links` and at the reader's record site, since it was previously only implied:

- The link table describes what the current grid can show. An **accepted** seed **replaces** it: the capture covers the whole scrollback the grid keeps, so it is the complete set the pane is offering, including links the reader never saw (seed bytes replay into a fresh parser, not through `run_reader`). Merging would keep a label actionable against a URI the pane has stopped advertising.
- A **rejected** seed (`Busy`/`Failed`) leaves it untouched: it describes a frame that was never installed.
- The replacement runs under the parser lock beside the grid it describes, so no sampler sees one without the other, and the reader's record-and-apply is atomic against it under the snapshot mutex. A seed can therefore never land between a target being recorded and the label that needs it reaching the grid.

`record_seed_links` moves into the tests. Its doc had outlived the behavior it described: "recorded even when the swap loses its race" has not been true since the reconcile moved inside the accepted branch in #3749.

Closes #3817. Closes #3818, as verified-not-present plus the coverage and contract that keep it that way.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Unit tests in `src/tmux/vt.rs`, per the issues' own request for barriers and injected results rather than sleeps or live tmux races.

The #3817 cases are single-threaded interleavings, which is stricter than a barrier: the resize guards are values, so an ownership handover is expressed by holding two and ordering the calls, with no wall clock in those tests at all. Both new assertions were confirmed red against `main` before the fix.

#3818 gets one test per side of the ordering. The reader's holds the fence a real `install_seeded_parser` occupies, delivers the OSC 8 chunk into that window, and asserts neither side entered: the install never reaches its drain and the reader never claims the chunk. Releasing it, the label and its URI arrive together whichever side wins. The install's probes the fence at both ends of its own hold: the forwarder probes during the drain, which cannot complete until it answers, and the link table is then held so the install parks inside the swap, with the parser lock going away as the signal it arrived. Both #3818 tests do use bounded polling and a `recv_timeout`, but only in the safe direction: every timing assertion is a negative held open by a lock this test owns, so a slow machine cannot turn one red.

Verified not green-by-accident, one ablation per binding. Removing the reader's fence turns its test red at the `chunk_seq` assertion. Releasing the install's fence before its drain, and releasing it after the drain but before the table replacement, each turn the install-side test red at their own assertion; the second of those left the whole module green until this PR's third commit, which is the gap that commit closes. Removing the reader's fence *and* the seed's chunk/pipe guards reproduces the reported erase exactly, the label on the grid and an empty link table; note that this last one is only observable with the two white-box assertions relaxed, since `chunk_seq` fires first. Those white-box assertions are the regression value here, not the tail.

`cargo fmt`, `cargo clippy --all-targets` and `cargo test` are clean (6858 unit + 289 integration). `main` is separately red on `tmux::session::tests::wait_until_ready_blocks_until_the_marker_appears` (macOS), which is not from this branch.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Both reproductions, the ablations behind them, and the fix were driven by @njbrake through Claude Code. The finding that #3818 is already closed on `main` is the model's, verified by ablating each guard and observing the reported failure return.

https://claude.ai/code/session_01JB7tMd8JJkXwavaLadMyyH

## Summary

- Coordinated concurrent resize declarations with shared state.
- Prevented stale or rejected resizes from clearing pending VT resync work.
- Improved snapshot capture by validating pane geometry before installation.
- Preserved cursor alignment when pane and grid heights differ.
- Added regression tests for resize races, geometry mismatches, cursor placement, and hyperlink handling.

## Benefits

- Reduces stale layouts during resize races.
- Keeps seeded cursor positions accurate.
- Preserves hyperlink targets during snapshot reseeding.
- Makes resize and snapshot behavior easier to verify.


src/tmux/vt.rs
```diff
@@ -5424,6 +5484,264 @@ mod tests {
...
+
+        let (_data_reader, data_forwarder) = UnixStream::pair().expect("data pair");
+        let (parent_control, mut forwarder_control) = UnixStream::pair().expect("control pair");
+        let snapshot = Arc::new(Mutex::new(()));
+        let probed_fence = snapshot.clone();
+        let forwarder = std::thread::spawn(move || {
+            let (kind, generation) =
+                read_drain_frame(&mut forwarder_control).expect("receive drain probe");
+            assert_eq!(kind, DRAIN_PROBE);
+            let fenced = probed_fence.try_lock().is_err();
```
jerome-benoit, line 5729: [P3] Qualify the post-drain coverage claim

This probe runs before the forwarder sends its ACK, so it verifies that the snapshot lock is held during the drain. It does not distinguish the current implementation from one that releases the lock immediately after draining but before replacing the links. Production still holds the guard through the swap (`install_seeded_parser:1120-1149`), so this is non-blocking test hardening, not a runtime defect. Please narrow the comment/PR claim, or add an observation at the post-drain, pre-swap boundary to pin that part of the lifetime too.

jerome-benoit (approved):
Reviewed d6e480f against the current base, including #3817, #3818 and the existing review. No blocking finding after two independent sequential analyses and arbitration.

The resize bookkeeping now keeps declaration identity, active-operation count and retirement under one mutex. The tests cover both reported overlap cases. For #3818, the shared snapshot fence already excludes the reported reader/install interleaving in production; this PR adds coverage and documentation rather than another runtime fix.

One non-blocking P3 below qualifies the install-side test's coverage. CodeRabbit's misplaced helper documentation nit is also valid and non-blocking; no duplicate comment.

Static inspection and clean `git diff --check` only. I did not rerun the author's reproductions or ablations. CI, Tests and CodeQL remain queued at this SHA.

jerome-benoit (approved):
Re-reviewed d6e480f..e2914fa with two independent sequential analyses and cross-validation. No new blocking finding.

The withdrawal now preserves an older resize still in flight (`vt.rs:2242-2247`). The previous P3 is addressed: the install test holds the link table, waits for the parser lock and checks the snapshot fence inside the swap (`5538-5575`), not only before the drain ACK. The misplaced helper documentation is also corrected. The bot's encapsulation suggestion remains optional.

Non-blocking wording: the added test uses bounded polling, so please qualify the description's absolute no-wall-clock claim; its assertion still checks the actual lock lifetime.

Static inspection and git diff --check passed; no local project build/tests or author ablations rerun. Workflow Tests 34285298588 failed on `container_terminal_command_preserves_runtime_boundaries` with `docker: Text file busy`, and `composer-stop.spec.ts:138` with a 15-second timeout. I started one failed-job rerun. Approval covers the code review, not a claim that the rerun has passed.

jerome-benoit (approved):
Re-reviewed e2914fa..06d0a42 with two independent sequential analyses and cross-validation. No finding.

The updated `abandon` documentation (`src/tmux/vt.rs:2314-2318`) matches both withdrawal conditions in `finish` (`2242-2247`) and the deferred call through `Drop` (`2324-2328`). This delta changes no executable code.

Exact-SHA `git show` inspection and `git diff --check e2914fa 06d0a42` passed. No local build/tests rerun. The current head has successful title/template checks; I am not claiming a new test run at this documentation-only SHA.

jerome-benoit (approved):
Renewing approval at f1cf409e after resolving the conflict with main (2e31f0a). Two independent sequential checks and cross-validation confirmed that the merge preserves both the resize/snapshot ordering fixes and main's target-geometry capture and cursor mapping. No new finding.

All 75 VT tests passed with an isolated tmux server and no user configuration; the initial run with local tmux configuration failed two real-tmux tests, which both passed in the isolated run. `cargo clippy -- -D warnings`, `cargo fmt --check` and the merge diff checks passed. GitHub CI is still running; this is not a claim that it has completed.

PR #3828, opened 2026-09-08
refactor(tui): read the sidebar's session list through the daemon client

njbrake:
## Description

Slice E of #2734: the TUI sidebar reads the daemon's session list through the shared Rust client instead of a private wire view.

- `SessionFeed` (was `daemon_status_poller`) fetches `GET /api/sessions` through `DaemonClient` and hands the home view the rows; the home view projects daemon-owned fields onto the sidebar. Today that set is the runtime status of structured rows; terminal rows keep the local tmux poller because the daemon does not own tmux yet (DF-F), and its copy would add a second producer with multi-second lag behind the 500 ms local probe.
- Daemon absent: explicit `SidebarSource::Storage`, logged on transition. The local store serves the sidebar alone and daemon-owned rows keep their last value.
- Feature flag: `session.daemon_sidebar` (default on). Off never fetches and drops an in-flight result.
- Older daemons: `SessionResponse` decodes with only `id` required and `SessionsEnvelope` tolerates a missing `workspace_ordering`, so a field an older daemon omits cannot fail the whole parse. Serialization is unchanged.
- Per #2734 F: `HttpClient::list_sessions`, `remote_home::RemoteSessionWire`, and the poller's private row are deleted; `remote_home` and the structured view's session info read the shared type.

Open decision, kept as #3575 left it: `DaemonClient` refuses a bearer token over non-loopback plaintext HTTP. Two undocumented setups lose the migrated reads (structured status, structured-view header, remote picker list) while `HttpClient` still sends the same token on every other call: a local `aoe serve --host <lan-ip>` (a specific non-loopback bind; `0.0.0.0` and loopback are unaffected) and `AOE_DAEMON_URL=http://<host>` with a token. Both surface as a logged fallback or a picker error, documented in `docs/structured-view.md`. Relaxing the client for same-host endpoints or extending the refusal to `HttpClient` is a policy call for #2734 F, not this slice.

Not moved: the row set and durable fields still come from storage. The TUI is still their writer, so reading them from a 1 s poll would revert each local edit until the daemon catches up. That follows the write-path slices.

Refs #2734

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No rendering changed.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behavior is a projection of one wire payload onto rows plus source transitions, covered by unit tests on the feed and the home view (seeded feed results, setting off, daemon unavailable) and a wire decode test for the minimal older-daemon row.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Fable 5.1 via Claude Code, driven by @njbrake.

Design per #2734 (approved 2026-08-28) and its constraint F from #3575.

https://claude.ai/code/session_012TrdcWW9BR7AMEyLjz4e47

## Summary

- Replaced the private TUI daemon poller with the shared `DaemonClient` and `SessionFeed`.
- Updated structured-session sidebar rows to use daemon data.
- Kept terminal rows on the local tmux poller.
- Added storage fallback when the daemon is unavailable.
- Preserved the last daemon-provided values during failures.
- Added the enabled-by-default `session.daemon_sidebar` setting.
- Reused shared session types and removed duplicate models.
- Improved compatibility with older daemon responses.
- Added tests for feed updates, fallback behavior, configuration, and wire decoding.
- Documented bearer-token restrictions for remote plaintext HTTP.

## Benefits

The sidebar now uses a consistent daemon session source. Failure handling is predictable, and older daemons remain usable.

## Further enhancement

Apply the plaintext HTTP bearer-token restriction consistently to the remaining daemon requests.


jerome-benoit (changes requested):
Reviewed ea429213, including #2734 and the existing discussion.

**P2, required before merge:** please address the existing [private rustdoc link finding](https://github.com/agent-of-empires/agent-of-empires/pull/3828#discussion_r3962162485). `discovery.rs:67` links a public method's documentation to a `pub(crate)` method, while `src/lib.rs:3` denies `rustdoc::private_intra_doc_links`. Plain code formatting fixes this without widening visibility. No duplicate inline comment.

No additional functional finding after two independent sequential reviews and cross-validation: the structured/terminal ownership boundary, disabled-feed guard, shared response projections and missing-field defaults are preserved. The documented non-loopback HTTP bearer restriction is treated as the explicit transport-policy boundary, not an additional blocker.

Static inspection and `git diff --check` only; no local build/tests/rustdoc run. CI, Tests and CodeQL are still queued at this SHA, so this does not assert a passing build.

jerome-benoit (approved):
Re-reviewed ea429213..7b93b41 with two independent sequential analyses and cross-validation. The sole P2 blocker is resolved: discovery.rs:67 now uses plain code formatting without widening visibility. The bearer note accurately distinguishes session-list requests from the other calls and links the open policy decision in #3839. No additional finding in this two-line documentation delta.

Static inspection and git diff --check passed; no local build or tests run. CI, Tests and CodeQL have completed successfully at 7b93b41 (runs 34280717986, 34280717865 and 34280718004).

Issue #3806, opened 2026-09-07
Tests: rule out wall-clock parks in favor of condition-based waits (mock clocks don't cover the flake class)


jerome-benoit:
## Summary

Timing-sensitive tests keep failing on wall-clock assumptions no mock can fix. Latest instance: `server::startup::tests::shutdown_deadline_runs_from_the_cancel_not_the_watchdog_poll` (`src/server/startup.rs:1327`) failed once on macOS main CI (6648 passed, 1 failed) by asserting a watchdog fires within a fixed 1 ms tokio park, while the same test blocks its single-thread worker with a real `std::thread::sleep(GRACE * 2)`. Rerun went green: flake, not regression.

## Why mocks are not the answer here

- Tokio's virtual clock (`start_paused` + `advance`, available: tokio full + dev `test-util`) controls only tokio time. It does not control `std::thread::sleep` or `std::time::Instant`, which the test suite uses pervasively (75 `tokio::time::sleep` sites plus dozens of `std::thread::sleep(10–80ms)` parks inside test modules, e.g. `src/session/recovery.rs`, `src/session/storage.rs`, `src/process/mod.rs`).
- `mock_instant` is not vendored and would still leave every blocking sleep on real time, with added global-state risk under parallel tests. Only one test uses `start_paused` today (`src/server/startup.rs:1202`).
- Web side is the same shape: `vi.useFakeTimers` exists but is used in 2 test files; live specs use zero `page.clock` and instead converge via `expect.poll` (65 files) — the pattern that actually killed flakes in #2775 (RAII guards, Barrier-driven peers, poll-with-deadline, never mocked clocks).

## Proposal

1. Rule for new timing-sensitive tests, Rust and web: condition-based waits (poll-with-deadline, Barrier, `expect.poll`) instead of fixed parks, especially any park ≤100 ms or any park asserting another task ran (the exact macOS failure shape).
2. Where the clock under test is fully tokio-owned, prefer `start_paused` + `advance` over real sleeps.
3. Audit the existing tiny-park sites for the mixed-clock trap (real blocking sleep + assumed scheduling quantum); convert the ones backing CI flakes.
4. Explicitly not adopt a global instant mock: it covers neither blocking sleeps nor multi-thread scheduling, the two actual flake sources.

## Prior art / instances (reference, not duplicates)

- #2775 (deterministic flake fixes without mocks — house precedent to follow).
- #3513 (open, specific tmux timing flake — instance of this class).
[...]

PR #3829, opened 2026-09-08
fix(tests): release the ready-marker pane causally, not on a sleep

njbrake:
## Description

`tmux::session::tests::wait_until_ready_blocks_until_the_marker_appears` is intermittently red on macOS. The pane printed the ready marker 700ms in and the test asserted the wait took at least 600ms. That floor is a proxy for "it did not return on the static screen", and it holds only while the waiter's polls stay in phase with the pane's sleep.

They fall out of phase on macOS. Each poll forks `capture-pane`, roughly 80ms there against a few ms on Linux, so the effective poll interval is nearer 280ms than 200ms. `refresh_session_cache` alone measured 160-183ms on the failing runners, shifting the waiter's start past the pane's. The second poll then lands around 720-740ms, just past the marker, and the wait returns at 561ms. Correct behavior, failed assertion.

This is not a regression from #3759, which changed only `src/tmux/vt.rs`; `capture_pane` forks tmux directly and never touches the VT path. The same assertion failed on macOS on 2026-09-04 (`a5bdb7bca`, 567.8ms) and 2026-09-05 (`8876585ee`, 568.2ms), days before that merge. The two consecutive failures at 18:30 were `bf875fecb` and `e012f15e0`, and only the second one failed on this test; `bf875fecb`'s Tests failure was Playwright (live), with its macOS cargo job green.

The old floor was also blind to the regression it was named for, and worse than blind: its pane printed `booting`, 7 characters, below `wait_until_ready`'s own `now.trim().len() > 20` gate. The settle fallback it claimed to guard was unreachable by construction. Deleting the `continue` that keeps the marker path off that fallback leaves the old test passing; the rewrite fails on it.

The test now drives the pane instead of racing it: the pane holds a screen the generic settle fallback would accept and prints the marker only once the test creates a release file. Observe the settled markerless screen across several of the waiter's polls, assert it has not returned, then release and require the return. The surviving wall-clock values are deadlines rather than floors: a 2s promptness bound after release, and a waiter budget that only has to outlast the window.

**Relation to #3806 / #3807.** #3806 covers this class repo-wide and @jerome-benoit's #3807 converts 21 tests, but not this one: this test's body is byte-identical between `main` and `fix/3806-flaky-wall-clock`, so there is no overlap to collide. This PR touches only this test and no production code, no helper, and no shared fixture, so it should merge cleanly either way.

**macOS is not exercised by PR CI.** `Cargo Test (macos)` skips its real steps on pull requests, so this fix first runs on its target platform after merge, on a main push. Pre-merge evidence is Linux-only.

**Review follow-ups.** The test is now `#[serial_test::serial]`: it was the only one in the module forking tmux from the parallel pool, and the rewrite doubled its overlap with the 35 serialized tmux tests, which is the contention the diagnosis rests on. Observations are collected and asserted after `thread::scope` joins, so a failed assertion no longer detaches a waiter that forks against the killed session for the rest of its budget; the mutated run now fails in 3.3s. The observation window counts samples instead of wall time, so a host slower than ~650ms per iteration lengthens the window rather than starving the count-based assertion and false-failing for the same reason as the original bug. A drop guard releases the pane on every path out of the scope, since `expect` on the capture panics before the release write; an injected panic mid-window unwinds in 0.44s instead of idling to the deadline. The waiter budget is 60s: it is only a failure deadline and must dominate the sample-bounded window rather than race it.

**Evidence.** Injecting a setup latency into the unchanged body reproduces the failure deterministically: pass at 0/100/180/250ms, fail from 300ms, where elapsed drops 600ms to 400ms as a poll moves off the fourth interval onto the third. The rewritten test passes 20/20 idle and 15/15 with every core saturated, and fails on the `continue` mutation the old one missed. The macOS phase itself is not reproducible on this Linux box, where the unchanged test returns at a stable 820-833ms (15/15 under the same load).

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No user-facing change; screenshots are not applicable.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

This PR is itself a test correction. `wait_until_ready`'s production behavior is unchanged, and the rewritten test covers it more tightly than the version it replaces.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Investigation, latency sweep, mutation check and repetition runs were executed and reviewed by @njbrake.

**NOTE:**

## Summary

- Reworked the readiness test to use a test-controlled release file instead of timing delays.
- Added cleanup that releases the waiter on every exit path, including assertion failures.
- Extended the failure deadline to 60 seconds and clarified capture failure handling.
- Kept production code, helpers, and shared fixtures unchanged.

## Benefits

- Reduces intermittent macOS failures caused by timing assumptions.
- Makes the test more deterministic and easier to understand.
- Prevents a failed test from leaving the waiter blocked unnecessarily.

### DIFF
--- a/src/tmux/session.rs
+++ b/src/tmux/session.rs
@@ -2770,17 +2770,32 @@ mod tests {
         assert!(status.success());
         refresh_session_cache();
 
+        /// Releases the pane on the way out, so an unwind inside the scope
+        /// below cannot leave the waiter forking against a killed session
+        /// until its budget expires.
+        struct ReleaseOnDrop<'a>(&'a std::path::Path);
+        impl Drop for ReleaseOnDrop<'_> {
+            fn drop(&mut self) {
+                let _ = std::fs::write(self.0, b"");
+            }
+        }
+
         let (returned_tx, returned_rx) = std::sync::mpsc::channel();
         let session = Session::from_name(&name);
-        // Observations are collected, not asserted, inside the scope: the
-        // release below then always runs, so no failure path leaves the waiter
-        // forking against a killed session and the scope's join stays prompt.
+        // Observations are collected, not asserted, inside the scope, and the
+        // release runs on every path out of it, so no failure leaves the
+        // waiter forking against a killed session and the join stays prompt.
         let (settled, last, premature, early, released) = std::thread::scope(|scope| {
             scope.spawn(|| {
+                // Only a failure deadline, and it has to dominate the
+                // sample-bounded window below: a budget the window can reach
+                // on a slow host would expire mid-window and read as the
+                // early return this test exists to catch.
                 Session::from_name(&name)
-                    .wait_until_ready(std::time::Duration::from_secs(15), Some("ask anything"));
+                    .wait_until_ready(std::time::Duration::from_secs(60), Some("ask anything"));
                 let _ = returned_tx.send(());
             });
+            let _release_on_unwind = ReleaseOnDrop(&release);
             // Negative claim, so it needs a window: hold the settled markerless
             // screen across several of the waiter's 200ms polls, which is the
             // state an early return would key on. Bounded by sample count, not
@@ -2823,6 +2838,7 @@ mod tests {
         );
         released.expect("did not return promptly once the marker appeared");
     }
+
     #[test]
     fn chrome_rows_accounts_for_status_bar_and_ignores_splits() {
         // #2766: the reporter's tmux yields a pane one row shorter than the
--- a/src/tmux/session.rs
+++ b/src/tmux/session.rs
@@ -2722,6 +2722,7 @@ mod tests {
     /// (a "still loading" screen can look "settled" long before the agent is
     /// really listening).
     #[test]
+    #[serial_test::serial]
     fn wait_until_ready_blocks_until_the_marker_appears() {
         if !tmux_available() {
             eprintln!("Skipping test: tmux not available");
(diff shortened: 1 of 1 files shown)


jerome-benoit (approved):
Reviewed fcbdccb after two independent sequential reviews and cross-validation. No blocking finding in this test-only change. The release file controls marker arrival; the test observes a stable markerless screen, checks for premature completion, then requires completion after release. Production behavior is unchanged.

**P3, non-blocking:** the existing [unwind-release comment](https://github.com/agent-of-empires/agent-of-empires/pull/3829#discussion_r3962185511) is valid. A capture panic precedes release, so the scoped join can wait for the waiter timeout. The thread remains joined rather than detached; a best-effort release guard would improve failure cleanup. No duplicate inline.

Please also align the description with the code: the waiter budget is **15s**, not 20s, and a **2s** post-release deadline remains. The marker release is causal, but the negative observation window still depends on waiter scheduling; this is not complete scheduler independence.

`git diff --check` is clean. Static inspection only; I did not rerun the author's mutation/stress tests. CI, Tests and CodeQL remain queued at this SHA.

jerome-benoit (approved):
Re-reviewed fcbdccb..e0e3cfb with two independent sequential analyses and cross-validation. No new blocking finding; the previous P3 is addressed.

The scope-local release guard runs before the implicit thread join on unwind (`src/tmux/session.rs:2776-2798`). The explicit release and before/after assertions remain intact (`2818-2839`); increasing the waiter budget to 60s does not turn premature completion into a passing result. Production code is unchanged.

Both reviewers agree on approval. One wording caveat: cleanup is best-effort, not a guarantee of a prompt join under persistent I/O failure. The negative observation window still depends on scheduling; this is not complete scheduler independence. Neither is a new blocker.

Exact-SHA static inspection and `git diff --check fcbdccb e0e3cfb` passed. No local build/tests or author-reported panic/stress experiments rerun. CI remains in progress; macOS execution is not established by this review.

PR #3831, opened 2026-09-08
fix(vt): map the seeded cursor onto the grid the capture actually filled

### LINKED ISSUE #3824: Flaky live spec: live-size-owner-takeover resolves a stale READY> row (3 candidate VT/capture causes)
## Summary

`live-size-owner-takeover` ("... cursor on the prompt row", chromium) is flaky: 2 failures in 6 executions against `main` at `b4098e09` (delta 76.75 px vs threshold 2 at `spec.ts:115`), while the failing PR CI (`34260214272`) went green on attempt 3 with no code change. The failure and the rerun evidence are already documented in the RCA comments on #3782 ([RCA](https://github.com/agent-of-empires/agent-of-empires/pull/3782#issuecomment-5590214102), [green rerun](https://github.com/agent-of-empires/agent-of-empires/pull/3782#issuecomment-5590289592)); this issue tracks the three candidate causes the RCA's sub-agents surfaced but did not demonstrate.

Observed failure shape: the `rows.find` oracle at `spec.ts:36` selects the first row containing `READY>`, while the CI trace shows `READY>` on DOM rows 30 and 38 with the cursor in the second, so the oracle can resolve to a stale row while the live cursor sits on a duplicate.

## Candidate causes (readings, not demonstrated bugs)

1. **Potentially incoherent VT seed**: `src/tmux/vt.rs:1003-1017` returns the last snapshot after three discordant probes; `:2195-2203` reconciles geometry/cursor but not cells. [INFERENCE] Stale content could persist for a silent agent when cursor and geometry agree. No reproduction attributes the duplicate to this path.
2. **Non-atomic geometry/parser publication**: `src/tmux/vt.rs:2342-2368` reads dimensions before the parser lock; resize swaps the parser at `:881-914` then publishes dimensions at `:2502-2507`. [INFERENCE] A transient snapshot can mix old geometry with new cells; this alone does not explain the 10-second timeout.
3. **Incomplete capture probes**: `src/tmux/session.rs:229-240`, `:929-965`, `:1018-1028` reconcile on `history_size` and `pane_height` only, with no width or y comparison. [INFERENCE] Pre-resize content can survive paired with a post-capture cursor.

## Discrimination protocol for the RCA

Capture together every `READY>` occurrence: DOM line, cursor, raw WS frame, and tmux content at equal geometry. That separates "duplicate is fixture/tmux-side" from "capture corruption". Do not blindly replace `find` with `findLast`; the duplicate row's origin is the open question.

## Caveats

Sample is short (2/6 and 0/3 under different worker counts), so the flake rate is not estimated. Main's `Tests` run `34257778029` was green at merge time.

_Source: internal RCA session on #3782 (2026-09-08); the three candidate causes were its unpublished internal findings, persisted here before its workspace was retired. Line refs measured at `b4098e09`/`f3358931`._

njbrake:
## Description

When you have the live terminal open on your phone and hand control back and forth between two devices, the app sometimes drew the agent's input prompt twice, with the blinking cursor sitting on the lower, phantom copy instead of the real one. The real terminal only ever had one prompt; the app's own copy of the screen had drifted.

It happened because of a timing gap. Each time a device takes control it resizes the terminal, and the app rebuilds its picture of the screen by asking tmux for a fresh snapshot. If the snapshot came back describing the terminal at its old, taller size, the extra rows scrolled off the top of the smaller picture and pulled the content up with them, but the cursor was left behind at the row number tmux had reported. The next time the agent redrew its prompt it printed it wherever the cursor was stranded, and that stray copy is what you saw. Nothing noticed, because every check the app runs compares the screen's size and the cursor position, and by then those matched again; only the text was wrong.

This change places the cursor by counting up from the bottom of the snapshot rather than down from the top, so it moves with the content whenever the two sizes disagree. When they agree, which is the normal case, nothing changes. It also asks tmux again, briefly, for a snapshot taken at the size being rebuilt, so the picture matches the screen rather than merely being self-consistent. The test that was supposed to catch this now checks that only one prompt is on screen, which is the part that had stopped being checked.

Fixes #3824

### Root cause

`assemble_seed_stream` emitted tmux's `#{cursor_y}` as the seeded grid's row directly. That is correct only while the grid is as tall as the pane the `capture-pane` body came from. A reseed racing `resize-window` breaks it: the capture reads the pane at its old height, the surplus rows scroll off the top of the shorter grid and carry the content up, and the cursor stays parked that many rows below where the content landed. The agent's SIGWINCH redraw then prints its prompt at the stranded cursor, and the grid holds two prompt rows where the pane has one.

The divergence is invisible to every existing guard. `reconcile_step` compares geometry and cursor; once the resize settles, both agree and only the cells differ.

`seeded_cursor_row` maps the row from the pane's bottom instead. It reduces to `cursor_y` whenever the heights agree, and a probe carrying no geometry keeps the old mapping.

### Evidence

Instrumenting the seed across 36 runs of the spec: the height mismatch fired on 19 of 248 seeds, 15 of them shifting the cursor by 8 rows. At the spec's 9.6px line height that is 76.8px, matching the 76.75px delta the CI failures reported.

A capture at a failing checkpoint, showing the duplicate is grid-side and not fixture- or tmux-side:

```
tmux:    78x57  cur=7,30  hist=0   -> one READY> row, at 30
grid:    rows=57 hist=8 lines=65   -> READY> at DOM rows 30 AND 38
cursor:  cur=7,30 -> row=38
```

tmux has no scrollback and one prompt; the grid claims 8 lines of history and two.

### Supersedes #3826

#3826 landed on main eight minutes before this branch's first commit and changed the same oracle in the opposite direction, to `promptRows[promptRows.length - 1]`, on the reading that the second prompt row is the pre-resize prompt left behind in the scrollback. Issue #3824 had explicitly asked that this swap not be made until the duplicate's origin was known.

It is not scrollback. At the failing checkpoint tmux reports `history_size=0` and a single prompt row while the grid shows two, and the fixture's WINCH handler is `printf '\r READY> '`, a bare carriage return with no newline, so the pane cannot produce a second row on its own.

#3826 made this spec pass with the grid-side duplicate still live in the product. This PR restores the count assertion and fixes the cause, so rebasing onto #3826 reverts its oracle change deliberately. That is recorded in the spec's own comment so the next reader is not left with two merged changes asserting opposite things about the same three lines.

The before/after flake numbers below were first measured against the pre-#3826 merge base and have been re-measured on the rebased branch.

### Candidate causes from the issue

All three were leads, and all three are ruled out by the capture above.

1. **Incoherent VT seed** (three discordant probes returning the last snapshot). Ruled out: the probes agreed. Every seed logged in the failing runs came back on a probe pair that matched, and the body/probe pair is self-consistent. The corruption is downstream of the snapshot, in how its cursor row is mapped onto a grid of a different height.
2. **Non-atomic geometry/parser publication** (a snapshot mixing old geometry with new cells). Ruled out as the cause of this failure: the published frame is internally consistent (57 rows of cells, geometry 57), and the failure persists past the 10s poll rather than resolving on the next frame, which a transient mixed publication would.
3. **Incomplete capture probes** (`session.rs` reconciling on `history_size`/`pane_height` only, no width or y). Ruled out for this failure: the failing frames are served by the VT grid transport, not `capture_pane_with_cursor`. `transport=grid` in the debug overlay on every captured failure. The legacy path's probe coverage is untouched by this bug.

### Still latent after this

Candidate cause 3 named a real gap that is not this bug and is not fixed here: `merge_cursor_probes` (`src/tmux/session.rs:229`) reconciles on `history_size` and `pane_height` but never `pane_width`, though `CURSOR_FMT` already collects it. A width-only resize between the probes rewraps the pane and still reports `position_reliable: true`. That is the legacy `capture_pane_with_cursor` path, not the grid transport these failures came through.

The width axis of the seed itself is now handled: `capture_seed_snapshot` prefers a snapshot taken at the geometry being seeded, so an over-wide body is retried rather than remapped. `seeded_cursor_row` remains the fallback when the pane will not settle there, and it counts logical rows, so an over-wide body still undercounts.

Correcting an earlier version of this description: a width mismatch confined to the capture does NOT trip `reconcile_step`'s geometry branch. That compares tmux's geometry against the grid's stored geometry, and `set_grid_size_with_deadline` stores the resize target, so once the resize settles both equal the target. It heals through the slower cursor-drift branch, which by design does not fire while output is flowing.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

`live-size-owner-takeover.spec.ts` now asserts the prompt row count alongside the cursor alignment. Its `terminal.live-size-owner` matrix entry already lists the spec, so no matrix edit was needed. Two unit tests cover the mapping: one reproducing the taller-pane seed (fails without the fix), one pinning the identity cases.

**TUI / CLI (e2e test)**

### Verification

The flake rate on this branch's base was 2/6, so repeated runs, not a single green one:

- Before (pre-#3826 base, where the oracle still detected the duplicate): 1 failure in 6 runs of the spec, then 2 in 6 on a rerun; 2 failures in 9 runs of an instrumented copy at the same checkpoints.
- After, on the rebased branch: 36/36 runs of the spec green, plus 12/12 on a re-run after the last comment change, all at 3 workers. Instrumentation confirms the race still fired 19 times across 36 green runs and was corrected rather than avoided.
- The mapping is covered end to end, not only by hand-built state: a real-tmux test seeds a live 40-row pane into a 24-row grid and fails without the fix (cursor at row 20 rather than 4).
- 30/30 green across the neighbouring live specs (`live-grid-transport`, `live-frame-compression`, `mobile-live-no-flutter`, `mobile-live-scrollback-buffer`, `theme-switch-live`).
- `cargo test --features web`: 7153 passed, 0 failed. CI's exact rustdoc gate (`RUSTDOCFLAGS=-D warnings cargo doc --no-deps --features web`) clean. `cargo fmt --check`, `cargo clippy --features web --all-targets` clean. Web `format:check`, `lint`, `tsc -b` clean.

## AI Usage

- [ ] No AI was used
- [x] AI was used

Root cause was found by instrumenting the seed path and capturing DOM, cursor, WS frames and tmux state together at the failing checkpoint, per the discrimination protocol in #3824.

https://claude.ai/code/session_012Pa5NEKgMq7uByVajxEFgW

## Summary

- Fixed a terminal resize race that could place the cursor on the wrong prompt row and create duplicate `READY>` prompts.
- Updated cursor mapping to account for height changes during snapshot capture.
- Requested snapshots at the target terminal size before rebuilding the grid.
- Strengthened the live-size takeover tests to require one prompt and correct cursor alignment.
- Added unit tests for resize races, matching heights, and shorter grids.

## Benefits

- Improves prompt and cursor consistency during terminal resizing.
- Reduces flaky live tests and prevents stale prompt rows.
- Keeps existing behavior when terminal heights match.

### DIFF
--- a/src/tmux/vt.rs
+++ b/src/tmux/vt.rs
@@ -1387,13 +1387,11 @@ fn split_seed_capture(raw: &[u8]) -> (&[u8], &str) {
 /// the full pane height, so the parser's visible screen is a pixel-for-pixel
 /// replica of the pane. Only the single trailing line terminator is dropped:
 /// with it, the final `\n` would push the whole screen up one row (the top row
-/// scrolls into history) and misplace every cell. Because the visible screen is
-/// faithful, the CUP is a plain 1-based `#{cursor_y}` / `#{cursor_x}`, which
-/// addresses the visible screen regardless of how much scrollback sits behind
-/// it (that is the coordinate space tmux reports the cursor in). Without this,
-/// the parser's cursor lands after the last replayed glyph, bottom-right for a
-/// full-screen app, until the first live chunk carries the app's own escapes
-/// (issue #2902).
+/// scrolls into history) and misplace every cell. The CUP that follows carries
+/// `#{cursor_x}` and the row [`seeded_cursor_row`] maps `#{cursor_y}` onto.
+/// Without it the parser's cursor lands after the last replayed glyph,
+/// bottom-right for a full-screen app, until the first live chunk carries the
+/// app's own escapes (issue #2902).
 fn assemble_seed_stream(body: &[u8], state: &PaneSeedState, rows: u16) -> Vec<u8> {
     let mut out: Vec<u8> = Vec::with_capacity(body.len() + 32);
     if state.alt {
@@ -1418,10 +1416,9 @@ fn assemble_seed_stream(body: &[u8], state: &PaneSeedState, rows: u16) -> Vec<u8
         out.extend_from_slice(b"\x1b[?1h");
     }
     out.extend_from_slice(&lf_to_crlf(strip_trailing_row_terminator(body)));
-    // 1-based CUP in visible-screen coordinates, clamped to the grid so a stale
-    // query (the pane moved between the state read and this seed) can't push the
-    // cursor off-screen; the first live chunk re-syncs it either way.
-    let cy = state.cursor_y.min(rows.saturating_sub(1)) + 1;
+    // 1-based CUP, clamped to the grid so a state read this far off can't push
+    // the cursor off-screen; the first live chunk re-syncs it either way.
+    let cy = seeded_cursor_row(body, state, rows).min(rows.saturating_sub(1)) + 1;
     let cx = state.cursor_x + 1;
     out.extend_from_slice(format!("\x1b[{cy};{cx}H").as_bytes());
     out.extend_from_slice(if state.cursor_visible {
@@ -1430,6 +1427,37 @@ fn assemble_seed_stream(body: &[u8], state: &PaneSeedState, rows: u16) -> Vec<u8
         b"\x1b[?25l"
     });
     out
+}
+
+/// The seeded grid's own row for the pane cursor tmux reported at
+/// `state.cursor_y`.
+///
+/// tmux counts `cursor_y` from the top of the pane's visible screen, so that is
+/// the grid's row only while the grid is exactly as tall as the pane the body
+/// came from. A reseed racing a `resize-window` breaks that: the capture reads
+/// the pane at its old, taller height, the surplus rows scroll off the top of
+/// the shorter grid and carry the pane's content up with them, and a bare
+/// `cursor_y` leaves the cursor parked that many rows BELOW the content. The
+/// app's next redraw prints its prompt there, so the grid ends up holding two
+/// prompt rows where the pane has one, and no reconcile can see it: grid and
+/// pane still agree on geometry and on the cursor, only the cells differ.
+///
+/// Counting from the pane's bottom row instead survives a mismatch either way,
+/// and reduces to `cursor_y` whenever the two heights agree. A `pane_height` of
+/// 0 means the probe carried no geometry, so keep the plain mapping there.
+fn seeded_cursor_row(body: &[u8], state: &PaneSeedState, rows: u16) -> u16 {
+    if state.pane_height == 0 {
+        return state.cursor_y;
+    }
+    let fed = strip_trailing_row_terminator(body);
+    let body_rows = if fed.is_empty() {
+        0
+    } else {
+        u16::try_from(fed.iter().filter(|&&b| b == b'\n').count() + 1).unwrap_or(u16::MAX)
+    };
+    // Rows of the body the grid still shows; the rest scrolled into history.
+    let visible = body_rows.min(rows);
+    visible.saturating_sub(state.pane_height.saturating_sub(state.cursor_y))
 }
 
 /// Drop the single trailing line terminator (`\n` or `\r\n`) from a
@@ -3749,6 +3777,7 @@ mod tests {
             cursor_x: 3,
             cursor_y: 1,
             cursor_visible: true,
+            pane_height: rows,
             ..Default::default()
         };
         let mut p = vt100::Parser::new(rows, cols, SCROLLBACK_LINES);
@@ -3809,6 +3838,7 @@ mod tests {
             cursor_x: 2,
             cursor_y: 1,
             cursor_visible: true,
+            pane_height: rows,
             ..Default::default()
         };
         let mut p = vt100::Parser::new(rows, cols, SCROLLBACK_LINES);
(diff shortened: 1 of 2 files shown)


jerome-benoit (approved):
Reviewed 784dbe3, including #3824 and the stated scope. No blocking code finding after two independent sequential reviews and cross-validation.

The bottom-relative cursor mapping follows the captured content when pane and grid heights differ, while preserving the equal-height and missing-geometry cases. The parser-level regression covers the taller-pane case, and the live oracle now requires exactly one prompt plus cursor alignment rather than selecting around a duplicate. The already-declared width/probe limitations are not treated as new blockers.

**Integration remains pending:** GitHub reports a merge conflict, and no CI/Tests/CodeQL run exists for this head; only PR metadata workflows are queued. Please resolve the conflict while preserving the one-prompt assertion. The resulting integration delta needs rereview and CI before merging.

Static inspection and clean `git diff --check` only. I did not rerun the author's tmux/Playwright repetitions or confirm their reported pass counts.

jerome-benoit (approved):
Re-reviewed e954c002 after the rebase and additional seed-capture changes, with two independent sequential analyses and cross-validation. The conflict is resolved, the one-prompt oracle is preserved, and target-geometry retries are bounded. No new blocking finding.

The existing [off-geometry finding](https://github.com/agent-of-empires/agent-of-empires/pull/3831#discussion_r3962666650) is a valid P2 follow-up, non-blocking for this scoped fix: the final fallback can still accept an off-target snapshot, but that behavior predates this delta; the new retries reduce its window rather than introducing it. This is not a claim that width mismatches are fully solved. No duplicate inline.

A proposed pane-index test defect was rejected after an isolated tmux check: with pane-base-index=1, the test's `^.0` target still resolved successfully. Please keep the spec comments focused on the invariant rather than the implementation history of #3826.

Static delta inspection and git diff --check passed. CI, Tests and CodeQL pass at this SHA (34283203790, 34283203704, 34283203700). No local project build or full test suite; the author's repeated live-test results were not rerun.

Issue #3784, opened 2026-09-06
Sandbox containers fail to restart after host reboot due to `/tmp` clearing


clefru:
(Disclosure: The issue was diagnosed with codex, and codex wrote the issue description and patch)

## Summary

AoE sandbox containers created with agent status hooks can no longer be started after the host removes their per-instance hook directory. A host reboot commonly clears `/tmp` and this erases the bind-mount for hooks under `/tmp`. AoE reuses the stopped container without recreating that source directory, so Podman and `crun` reject the start.

## Environment

- AoE 1.15.3
- Linux
- Podman with `crun`
- Sandboxed Codex session with agent status hooks enabled

## Reproduction

1. Create and start a sandboxed session whose agent supports status hooks.
2. Stop the session while preserving its sandbox container.
3. Reboot the host, or otherwise remove `/tmp/aoe-hooks-$(id -u)/<instance-id>`.
4. Resume the session so AoE tries to start the existing container.

## Actual behavior

The container fails to start with an error like:

```text
Failed to start container: Error: unable to start container "<container-id>":
crun: cannot stat `/tmp/aoe-hooks-1000/<instance-id>`:
No such file or directory: OCI runtime attempted to invoke a command that was not found
```

I reproduced this with one current sandbox and found the same failure for another sandbox in an earlier AoE debug log after a host reboot.

## Expected behavior

Before starting an existing sandbox container, AoE should recreate the per-instance hook directory when the selected agent supports status hooks. The existing container should then start normally.

## Root cause

`build_container_config` creates the hook directory and records it as a bind mount when the container is first created. The stopped-container path in `Instance::get_container_for_instance` refreshes agent configuration and calls `container.start()`, but it does not restore the ephemeral host-side mount source first.

## Workaround

Recreate the missing directory before resuming the session:

```sh
install -d -m 700 /tmp/aoe-hooks-$(id -u)/<instance-id>
```

## Suggested fix

Call the existing hardened `crate::hooks::ensure_instance_dir_path` helper before starting a stopped sandbox container for a hook-capable agent. If restoration is unavailable, preserve the current warning and fallback behavior for optional status hooks.

I will prepare a patch for this issue.

PR #3833, opened 2026-09-08
fix(pi): keep a Pi conversation when its transcript is gone

### LINKED ISSUE #3809: Some pi sessions lost after reboot
When I quit aoe, then close ghostty, then reboot my macos, and then restart aoe, I discover that some (but not all) sessions are just gone. They show as red X, and when I enter them it's simply a new session. I'm using pi. I currently use aoe 1.15.3.

njbrake:
## Description

A Pi session's transcript path is recorded in memory when the pane publishes it, but only written to disk at stop. A reboot skips that write, so the next launch has an id and no path. It falls back to `--session <sid>`, which pi treats as a hard-failing selector: it exits 1 on an id that resolves to nothing. `remain-on-exit` holds the dead pane, the resume probe reads it as a failed resume, and the retry starts a new conversation under the same row. That is the "red X, and entering it gives a new session" in #3809.

Pi writes a transcript lazily, so a conversation that was never prompted has no file, and that is the common case for a session sitting idle when the machine goes down.

Two changes:

- persist the published transcript path when the pane publishes it, not only at stop
- skip the `--session` arm when the recorded transcript for this row's conversation is confirmed absent, keeping the id for a later launch that can pin it

The evidence is the pane's own published path; pi's store layout is never reconstructed to guess. `--session-id` creates the conversation rather than failing and is exempt, as is an explicit user pin.

Not the same path as #3784 / #3785. That was a sandbox container failing to start on a missing `/tmp` hook bind mount. This is a host Pi launch whose selector resolves to nothing.

Verified against a real `pi` 0.85.1: reboot simulation (tmux server killed, `/tmp/aoe-hooks-*` removed) left a dead pane before the change and a live pane after, and a session with a real transcript still resumes it by path. Also verified for a session launched under a live `aoe serve` and never stopped, which is the case teardown-only persistence missed.

Closes #3809

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: the decision is in the launch line `apply_session_flags` builds, and the unit test drives that directly. An e2e test would need a real `pi` on the runner plus a lazily-written transcript, and the harness auto-skips when the tool is absent, so it would assert nothing in CI.

## Verification

- `cargo fmt --check`
- `cargo clippy`
- `cargo test --lib`: 6859 passed, 2 ignored, over 5 consecutive runs
- `cargo test --bins --tests`: 12 integration failures in `acp_runner_control`, `acp_runner_orphan`, and `plugin_install`, all reproducing on clean `8f0cece81` in this sandbox

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 5

**Any Additional AI Details you'd like to share:** Claude reproduced the failure against a real pi binary, wrote the patch and the regression test, and ran validation. Reviewed by @njbrake.

https://claude.ai/code/session_0154qwGbN5tRHKpwiPoWqZqu

## Summary

- Persist published Pi transcript paths before shutdown.
- Recover Pi conversations after reboot when the transcript exists.
- Start a fresh unpinned conversation when the recorded transcript is missing.
- Preserve the conversation ID for future launches.
- Keep explicit pins and `--session-id` launches unchanged.
- Distinguish missing transcripts from unreadable storage.
- Carry `pi_session_path` through daemon reapplication.
- Add documentation and unit tests for the new behavior.

These changes prevent recoverable Pi conversations from appearing lost after restart.

### DIFF
--- a/src/session/instance/session_id.rs
+++ b/src/session/instance/session_id.rs
@@ -4,6 +4,16 @@ use super::*;
 
 const PI_SIDECAR_MAX_BYTES: usize = 4096;
 
+/// What a recorded Pi transcript path resolves to. `Unreadable` means the
+/// store could not be inspected, which is a statement about our view rather
+/// than about the conversation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PiTranscriptState {
+    Present,
+    Absent,
+    Unreadable,
+}
+
 impl Instance {
     /// Acquire a pre-launch session ID for the agent.
     ///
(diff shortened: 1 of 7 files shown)


src/session/instance/session_id.rs
```diff
@@ -467,30 +477,120 @@ impl Instance {
...
+        let relative = if self.is_sandboxed() {
+            match path.strip_prefix("/root/.pi/") {
+                Some(relative) => Path::new(relative),
+                None => return PiTranscriptState::Unreadable,
+            }
+        } else {
+            match self.pi_host_view_of(path) {
+                Some(host_path) => {
+                    return match host_path.parent().map(std::path::Path::is_dir) {
+                        Some(true) if host_path.is_file() => PiTranscriptState::Present,
```
jerome-benoit: [P2] Preserve metadata errors as Unreadable

`parent.is_dir()` can succeed while `host_path.is_file()` fails with `PermissionDenied`; `is_file()` converts that error to false, so this branch returns `Absent`. The new gate at lines 920-934 then removes the resume selector instead of preserving the conversation on an inconclusive inspection.

Reproduction on Linux as an unprivileged user: create a transcript in a directory, change that directory to mode 0600, then evaluate this match. I compiled and ran a standalone Rust reproduction of the exact predicate: `PR predicate=Absent; metadata error=PermissionDenied; transcript still exists=true` (existence checked after restoring mode 0700). No Pi execution is needed to reproduce the classification defect.

Use a fallible metadata lookup and distinguish genuine absence from permission/I/O errors; the latter must remain `Unreadable`. Apply the same error-preserving rule to the sandbox lookup rather than relying on its boolean `regular_exists`. Add an access-error regression alongside the missing-file case.

src/session/sync.rs
```diff
@@ -400,6 +400,11 @@ fn drain_and_persist_session_ids_inner(
             } else {
                 inst.resume_probe_failed_sid = None;
             }
+            // The transcript path belongs with the id it names. Recording it
+            // here is what makes it durable while the pane is still running:
+            // the sidecar it comes from lives in the host temp directory, and
+            // the only other writer is teardown, which a reboot never reaches.
+            inst.absorb_published_pi_session();
```
jerome-benoit, line 420: [P3] Cover path publication when the SID has not changed

This call is reached only through `to_apply`. An eligible Pi launch preassigns its SID (`session_id.rs:35-39,152-156`), so a matching poller observation takes the earlier return at `sync.rs:191-194` and never persists its newly published path here. This is a remaining pre-existing coverage gap, not an additional merge blocker or proof of ordinary reboot failure with `--session-id`. Please handle validated same-SID path publication too, or narrow the persistence claim, with a regression asserting the durable path before teardown.

jerome-benoit (changes requested):
Reviewed fc27ae60 against #3809, the description and existing discussion, with two independent sequential analyses and cross-validation.

**Required: P2 below.** An access error can now be treated as a confirmed missing transcript and suppress the resume selector. Reproduced locally with the exact Rust predicate, not a full Pi launch.

**Non-blocking:** the same-SID persistence gap below and the existing memory-before-write retry limitation are not additional introduced blockers. In particular, the same-SID gap alone does not establish failure after an ordinary reboot: the eligible `--session-id` path still applies. Explicit pins and the create-session arm remain exempt from the new gate.

`git diff --check d7fb3c8...fc27ae60` passed. No project build/tests or Pi reproduction run locally; the author's runtime results are not my own.

jerome-benoit (approved):
Re-reviewed fc27ae60..78db89bd with two independent sequential analyses and cross-validation. The required P2 is fixed; no new blocking finding demonstrated.

Host and anchored sandbox lookups now distinguish absence from inspection errors (`session_id.rs:17-35,512-536`). Same-SID publication is also persisted before teardown (`sync.rs:240-243`).

**P3, non-blocking:** the existing [reapply finding](https://github.com/agent-of-empires/agent-of-empires/pull/3833#discussion_r3963258918) is valid, including the TUI: same-SID metadata is omitted from `SessionIdSyncOutcome`, so callers can retain stale in-memory paths (`sync.rs:246-247`, `server/session_identity.rs:115-136`, `tui/home/pollers.rs:379-390`). Please report those metadata updates in the outcome and cover both callers. No duplicate inline.

Both reviewers agree this is not a demonstrated new conversation-loss blocker: the path is durable for reboot, and launch reabsorbs the sidecar before selecting flags (`session_id.rs:875-879`). This does not claim every stale-state scenario is covered. Please also align the description's remaining 'keeps the ID' wording with the corrected guide.

Exact-SHA static inspection and `git diff --check fc27ae60 78db89bd` passed. CI 34291467768, Tests 34291467752 and CodeQL 34291467759 succeeded at this head. No local project tests or Pi reproduction rerun.

PR #3834, opened 2026-09-08
fix(sandbox): share one Claude credential file across sandboxes

njbrake:
## Description

Claude Code rotates its refresh token on every refresh and the old one stops working. The v027 store isolation gave each sandboxed session its own copy of `.credentials.json`, so every container became its own chain: the first to refresh logged all the others out, a login inside one container fixed only that one, and a restart never re-read the host credential because the store sync skips top-level files once a store has prior data.

Every Claude sandbox now bind-mounts one `sandbox-v2/.credentials.json` over its store's path. Each come-up folds the freshest of the Keychain entry, the host file and any legacy per-store copy into it by `expiresAt`. Claude Code 2.1.263 writes the file in place when a rename hits the bind mount, refuses symlinks and reloads on mtime. Containers created before the file was shared are told apart by a create-time label and recreated at their next start.

The fold runs at come-up rather than as a migration: a container built before the file was shared keeps refreshing its store copy while it runs, so which copy is freshest is only known when a session starts. The host stays a separate chain from the sandboxes, as before. Injecting `CLAUDE_CODE_OAUTH_TOKEN` from the host instead would leave containers unable to refresh, so a mounted file it is.

Untested here: Apple Container with a file mount nested in a directory mount. Codex and Gemini keep per-store copies until their writers are checked the same way.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the behaviour is in the store sync and the mount list, both covered by unit tests; the container recreation needs a runtime the e2e harness skips without.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Fable 5.1 via Claude Code

Diagnosis, design and code drafted by the model in discussion with @njbrake, who owns the decisions.

_Note: this PR description was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

https://claude.ai/code/session_01SHEbGF3z4uAJxvY3KX18Fa

## Summary

- Shared Claude credentials across all sandboxes through one synchronized file.
- Selected the freshest valid credential and protected unreadable files from overwrite.
- Recreated older sandboxes when credential mounts did not match.
- Kept Codex and Gemini credentials per store.
- Updated cleanup behavior, tests, and documentation.

## Benefits

- Claude sessions can reuse rotating refresh tokens reliably.
- Concurrent starts preserve credential consistency.
- Existing sandboxes migrate safely.

Runtime recreation and nested Apple Container mounts remain untested.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -1117,13 +1117,13 @@ Inspect and reclaim per-session sandbox agent stores
 
 ###### **Subcommands:**
 
-* `reclaim` — Report per-session agent stores whose session no longer exists in any profile, and how much disk they hold. Reports only unless `--delete` is given: each store holds a copy of that agent's credentials
+* `reclaim` — Report per-session agent stores whose session no longer exists in any profile, and how much disk they hold. Reports only unless `--delete` is given: a store can hold a copy of that agent's credentials
 
 
 
 ## `aoe sandbox reclaim`
 
-Report per-session agent stores whose session no longer exists in any profile, and how much disk they hold. Reports only unless `--delete` is given: each store holds a copy of that agent's credentials
+Report per-session agent stores whose session no longer exists in any profile, and how much disk they hold. Reports only unless `--delete` is given: a store can hold a copy of that agent's credentials
 
 **Usage:** `aoe sandbox reclaim [OPTIONS]`
 
--- a/docs/guides/sandbox.md
+++ b/docs/guides/sandbox.md
@@ -16,7 +16,8 @@ Docker sandboxing runs your AI coding agents (Claude Code, OpenCode, Mistral Vib
 
 Agent credentials are seeded from the host config into a private per-session
 sandbox directory, so agents authenticate without re-login. Containers do not
-share a writable agent store.
+share a writable agent store; Claude Code's credential file is the one
+exception, see [Shared credentials](#shared-credentials).
 
 ## CLI vs TUI Behavior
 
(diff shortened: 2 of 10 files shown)


jerome-benoit (changes requested):
Reviewed 730bbccc with two independent sequential analyses and cross-validation.

**P2, required before merge:** address the existing [read-error finding](https://github.com/agent-of-empires/agent-of-empires/pull/3834#discussion_r3962601031), without duplicating its inline. I reproduced the failure mechanism in a standalone Rust program using a synthetic file with mode 0200: `read=PermissionDenied; existing=None; final={}`. Propagate non-NotFound errors through `read_credential_file` (`container_config.rs:767-772`); do not merely add an `exists()` guard, since the merge arm also needs to preserve unreadable existing content.

**Non-blocking design/validation questions:**
- The host lock (`hooks/mod.rs:941`) is not among the shared mounts (`container_config.rs:2244-2261`). A container write between the host's read and write can therefore escape that lock. Please validate coordination with Claude's actual refresh writer; I have not reproduced a Claude token loss and am not claiming concurrency is safe.
- Please reconcile the persistent path transition with AGENTS.md and `docs/development/adding-a-migration.md`, which allow deferred versioned migrations. Continuous credential refresh can remain a startup operation; the layout transition is a separate concern.

The reviewers disagreed on treating those questions as additional blockers; only the demonstrated overwrite is classified as blocking. Apple Container's nested mount behavior remains unverified as disclosed.

`git diff --check` passed. CI, Tests and CodeQL pass at this SHA (34282277784, 34282277760, 34282277705). No local project build/tests or container/Claude execution; the reproduction used no real credentials.

jerome-benoit (approved):
Re-reviewed 730bbccc..a99484f with two independent sequential analyses and cross-validation. The required P2 is addressed; no new blocking finding.

`read_credential_file` now rejects an unreadable regular file (`container_config.rs:769-780`), including failures hidden by the underlying no-follow reader. The `?` at line 849 aborts the fold before either write arm. Unreadable candidates remain optional and do not bypass this check. The added regression test checks preservation through the actual preparation path; its permission case explicitly skips root.

Both reviewers agree. Existing questions about writer coordination, the layout transition and Apple Container remain non-blocking and are not claimed resolved by this delta.

Verified with exact-SHA `git show` and `git diff --check 730bbccc a99484f`. No local build/tests rerun. CI still has Linux bare-core and Rust CodeQL running; this approval is the code-review verdict, not a claim that all checks have passed.

PR #3844, opened 2026-09-08
fix(acp): stop a cancel overtaking the prompt it is meant to stop

njbrake:
## Description

The live `composer-stop` spec fails on `Stop while reasoning` because a cancel can overtake the prompt it is meant to stop. It is a product bug, not a test problem.

`/acp/prompt` and `/acp/cancel` are separate concurrent requests reaching the agent by different routes. The prompt path wakes the session (a disk write), folds control state, publishes `UserPromptSent` and awaits worker readiness before handing `ClientCmd::Prompt` to the connection task; cancel goes almost straight to the command channel. All of that is window. The agent then sees `session/cancel` with nothing in flight followed by the very prompt it named, and since an ACP cancel names no prompt, the turn runs to completion with the composer still showing Stop.

Evidence from the failing run's own artifacts (`Tests` 34287159752, shard 1/2), fake-agent log:

```
457147 session/prompt   (turn 1)
457828 session/cancel   (turn 1 Stop, during the prompt)
458032 session/cancel   (turn 2 Stop)   <- 118ms before its prompt
458150 session/prompt   (turn 2)
```

Turn 2's prompt POST left the browser 62ms before its cancel POST; the daemon took 182ms to reach `sending prompt` and 2ms to reach the cancel. The window is the prompt handler's own duration, so it is load-scaled: ~1ms idle, ~200ms on a loaded runner. `Stop while reasoning` is the only case that clicks Stop without first waiting for rendered output, which is why it is the one that lands inside the window.

Both handlers now claim the session's prompt-submission authority, and the prompt handlers claim it on entry rather than after the wake. That also puts them on the submission-before-`instance_lock` order every other holder already documents and takes. `force_end_turn` stays unguarded: it is the escape hatch for a wedged UI and must never queue.

Second commit removes eight `dashboard.spec.ts` cases that assert rendering or design-token constants with no asserted behavior and rewrites a ninth as a behavior test. It is independent of the fix and revertible on its own.

Third commit is review follow-up: the plugin turn path now claims on entry too, the now-callerless `begin_prompt_submission` is gone, a cancel that queues behind a non-prompt holder warns past 500ms (it publishes nothing while it waits, so the composer shows no progress and no Force stop hatch), and `prompt_submission`'s doc no longer states the pre-fix lock order.

Not fixed here: `Tests` run 34285833527 on `a604ed390` failed on `acp-compaction-phase` and `queue-follow-up-with-nav`, unrelated symptoms. Live-shard instability is broader than this one race.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

The existing `web/tests/live/acp-stories/composer-stop.spec.ts` is the user story and already fails on the bug; the durable regression test is `cancel_and_prompt_serialize_on_the_submission_guard` in `src/server/api/acp.rs`, which pins both halves of the ordering deterministically without a live agent. Reverting either half fails it (re-verified after the review changes). `coverage-matrix.json` needs no change: it names `tests/dashboard.spec.ts` rather than case titles, and every behavior it maps there still has a surviving case.

Rate measurement, via a throwaway probe that POSTs a prompt, waits N ms, POSTs a cancel and asserts the agent saw them in that order, 20 runs per delay:

| delay | before | after |
|---|---|---|
| 0ms | 20/20 reordered | 0/20 |
| 1ms | 20/20 | 0/20 |
| 2ms | 19/20 | 0/20 |
| 4ms | 2/20 | 0/20 |
| 8ms | 0/20 | 0/20 |

Also run: full live suite (163 passed; the one failure, `acp-view-switch` keep-context, reproduces 3/3 on clean `main` in the same sandbox and is a local tmux/shim issue), `composer-stop` 40x under CPU load, `cargo test --features web`, `cargo fmt`, `cargo clippy --features web --all-targets`, and web `format:check` / `lint` / `tsc -b`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Diagnosis and fix drafted in a back-and-forth session; the failing run's Playwright trace and fake-agent log supplied the ordering evidence.

https://claude.ai/code/session_01EguYWwG3aDaa26yv9sJLv2

## Summary

- Serialized prompt and cancellation requests to prevent `session/cancel` from overtaking its prompt.
- Applied the same ordering guard to plugin turns.
- Added warnings when cancellation waits longer than 500 ms.
- Simplified the prompt submission API and updated related documentation.
- Added a deterministic regression test for prompt and cancellation ordering.
- Removed dashboard tests that checked only visibility or styling constants.
- Reworked one mobile dashboard test to verify user behavior.

## Benefits

This prevents premature cancellation during “Stop while reasoning” and makes session behavior more reliable. The new warning helps identify slow cancellation waits.

### DIFF
--- a/src/plugin/session_api.rs
+++ b/src/plugin/session_api.rs
@@ -570,22 +570,24 @@ async fn sessions_turn_send(
         // it cannot answer `agent_busy` for a session the caller may not see
         // (#3685).
         // Ownership is checked before the wake so a foreign session is never
-        // touched. The wake clears an idle-dormant (or manually stopped)
-        // park the same way a user prompt does: a turn is intent to continue,
-        // so the host resumes rather than refusing (#3686).
-        deps.session_service
-            .admits_turn(&caller, &req.session_id)
+        // touched, and the guard is claimed before it too, as the user prompt
+        // handlers do, so a Stop cannot overtake a turn the host just started
+        // (see `acp_cancel`). The wake clears an idle-dormant (or manually
+        // stopped) park the same way a user prompt does: a turn is intent to
+        // continue, so the host resumes rather than refusing (#3686).
+        let _submission = deps
+            .session_service
+            .admit_prompt_submission(&caller, &req.session_id)
             .await
             .map_err(|e| map_send_error(e.into()))?;
         let woke_idle_dormant = deps
             .session_service
             .touch_and_wake_on_prompt(&req.session_id)
             .await;
-        let (_submission, dispatch) = deps
+        let dispatch = deps
             .session_service
-            .begin_prompt_submission(&caller, &req.session_id, woke_idle_dormant)
-            .await
-            .map_err(|e| map_send_error(e.into()))?;
+            .prompt_dispatch_under_submission(&req.session_id, woke_idle_dormant)
+            .await;
         // A cold worker is not a refusal on this path: `send_turn` resumes it
         // and waits, which is how a scheduler wakes a session it created. The
         // turn gates are, because a second prompt at a busy non-steerable
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -34,6 +34,12 @@ const MAX_ATTACHMENT_BYTES: usize = 10 * 1024 * 1024;
 /// Maximum decoded size of all attachments on one prompt (20 MiB).
 const MAX_TOTAL_ATTACHMENT_BYTES: usize = 20 * 1024 * 1024;
 
+/// How long `acp_cancel` may wait on the session's prompt submission before it
+/// says so. Well over the sub-millisecond claim on an idle daemon and over a
+/// same-gesture prompt's own handler on a loaded one, so this fires for a Stop
+/// queued behind something else. See `acp_cancel`.
+const CANCEL_SUBMISSION_WAIT_WARN: std::time::Duration = std::time::Duration::from_millis(500);
+
 /// Startup-error banner text for a failed detached structured-view spawn,
 /// shared by the create-path (`create_session`) and enable-path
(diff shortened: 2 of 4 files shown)


jerome-benoit (approved):
Reviewed e6007bf against 2e31f0a, including the existing discussion and two independent, sequential analyses. No blocking finding after cross-validation.

- HTTP and plugin submission now acquire admission before waking the session; cancellation holds the same guard through forwarding. This closes the pre-guard window without the timeout bypass discussed in #3539 (`src/server/api/acp.rs:1256,1436,1593-1609`; `src/plugin/session_api.rs:578`).
- The submission/instance-lock ordering remains consistent, and Force stop stays outside the submission guard (`src/server/session_service.rs:480-526,1495-1511`; `src/server/api/acp.rs:1619-1627`). The documented cancellation delay behind other guard holders remains a non-blocking limitation, not a demonstrated deadlock.
- The regression test checks guard placement and absence of an early wake, not end-to-end browser gesture ordering (`src/server/api/acp.rs:3543-3605`). The workerless-fixture objection does not apply here: readiness fails immediately outside `Resuming` (`src/acp/supervisor.rs:2749-2753`).

Validation: full four-file diff inspected; `git diff --check 2e31f0a..e6007bf` clean. No local tests run. Web checks, Clippy and both live Playwright shards passed on this head. The Linux bare-core job is still rerunning after `container_terminal_command_preserves_runtime_boundaries` failed with `docker: /bin/sh: 1: docker: Text file busy`; this approval is not a claim that the complete CI run is green.

PR #3830, opened 2026-09-08
fix(web): keep terminal selection through streamed frames and bracket toolbar pastes

njbrake:
## Description

Copy and paste in the mobile web terminal, seen on an iOS PWA with Claude Code.

**Copy.** Mounted rows were wrapped in a `Fragment` keyed by the visible range, so each appended line or scrolled row remounted every row and collapsed the selection. On iOS this dismisses the Copy callout mid-gesture. Rows and pads now render as one flat list; rows are keyed by pane line plus wrap offset, which is invariant as the window slides. jsdom tests assert the row nodes survive both an append and a slide past wrapped lines.

**Paste.** The toolbar paste button sent clipboard text raw, so a multi-line paste submitted each line as its own prompt. It now shares `bracketedPaste()` with the keyboard paths. On a plain-HTTP origin it tries `execCommand("paste")` into the focused input, which WebKit permits behind its Paste prompt; other engines get the HTTPS toast. The old fallback's listener juggling is gone.

Known tradeoff: web pastes hand-roll the bracketed-paste markers, so a pane that never enabled DECSET 2004 shows `00~`/`01~` debris. Pre-existing for the keyboard paths; the fix is routing web pastes through tmux `paste-buffer` server-side, as the TUI does.

Clipboard MIME normalisation moved to `web/src/lib/clipboard.ts` as `readClipboardText()`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording (no visual change)

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [ ] N/A: this PR doesn't change a user-facing dashboard flow
- [x] Added or updated a Playwright spec under `web/tests/` and updated `web/tests/coverage-matrix.json` (Vitest: `MobileLiveTerminal.rowIdentity.test.tsx`, `MobileTerminalToolbar.test.tsx`; matrix entry `terminal.clipboard-chords`)
- [ ] Skipped a test, because:

**TUI / CLI (e2e test)**
- [x] N/A: this PR doesn't change TUI rendering, a CLI subcommand, or session lifecycle
- [ ] Added or updated an e2e test under `tests/e2e/`
- [ ] Skipped a test, because:

## AI Usage

- [ ] No AI was used
- [x] AI was used

[...]

PR #3847, opened 2026-09-09
fix(web): hold the painted frame while a terminal selection is live

njbrake:
## Description

Follow-up to #3830, which fixed row identity but not row content.

React rewrites a changed row's existing text node in place, and the DOM's replace-data steps collapse any range whose endpoints sit inside it. Node identity never enters into it, so keeping rows mounted was necessary but not sufficient. A full-screen agent repaints every row of every frame, so a selection there never survived long enough to copy. `altScreen` is not the load-bearing condition, though: a spinner on the selected row collapses a normal-buffer selection the same way. The fix keys off the selection, not the mode, and needs nothing new on the wire.

A selection touching the grid now pins the painted frame until the user lets go, the tmux copy-mode convention. Newly captured scrollback still loads in above the held rows, so dragging upward past the top edge reaches older text rather than the blank spacer. A full-screen mouse app also gives up its grip on touch gestures while a selection is live: its `touch-action: none` and non-passive `preventDefault` are what WebKit needs left alone to drag the selection handles, so with them on the callout came up and its handles would not move. A held pane surfaces the existing **Back to live** control rather than looking hung.

Confirmed working on iOS Safari and the PWA against a build of this branch.

## PR Type

- [x] Bug Fix

## Checklist

No visual change to capture: the rendered difference is that the grid stops repainting while a selection is held. The tests below assert that directly; the iOS callout it exists for is a device check, done.

## Test Coverage Analysis

`MobileLiveTerminal.selectionHold.test.tsx` is the regression test: jsdom models the replace-data collapse faithfully, so it goes red on this bug, and it also covers the release, the back-to-live control, a selection outside the terminal, and scrollback populating under a live selection. `live-selection-hold.spec.ts` drives the story over the production bundle with a real Selection, and asserts the gesture release that jsdom cannot see. Every case was confirmed red without its fix.

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Diagnosed from a live report. The reproduction, the minimisation that ruled out `altScreen` as the cause, and every regression test were driven from a browser loop before any fix was written.

https://claude.ai/code/session_01Tj5Khfu9ux54CTxc7uXdKE

## Summary

- Keeps selected terminal text stable while live output continues.
- Adds a **Back to live** control and restores live updates when the selection ends.
- Loads new scrollback without moving or changing selected rows.
- Preserves native touch selection gestures during active selections.
- Prevents incomplete scrollback updates from duplicating rows or breaking rendering.
- Adds component and Chromium regression tests, plus documentation and coverage updates.

This prevents selected terminal text from changing or disappearing during live output.

### DIFF
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -18,6 +18,7 @@ import type { LiveFrame, LiveStats } from "../hooks/useLiveTerminal";
 import { useWebSettings } from "../hooks/useWebSettings";
 import { useIsCoarsePointer } from "../hooks/useIsCoarsePointer";
 import { useTerminalGestureBoundary } from "../hooks/useTerminalGestureBoundary";
+import { useSelectionHold } from "../hooks/useSelectionHold";
 
 // Mobile rendering of a tmux agent pane, mirroring the TUI's live mode:
 // the server streams `capture-pane` snapshots (src/server/live_ws.rs)
@@ -591,7 +592,7 @@ export const Row = memo(function Row({
 });
 
 export function MobileLiveTerminal({
-  frame,
+  frame: streamFrame,
   liveStats,
   transport,
   armAgentClipboard,
@@ -644,6 +645,11 @@ export function MobileLiveTerminal({
     setFontSize(configuredFontSize);
   }
   const scrollerRef = useRef<HTMLDivElement>(null);
+  // A selection anywhere in the grid pins the painted frame until the user
+  // lets go, so no row is rewritten out from under the range. Everything
+  // below renders that held frame; only the stream acknowledgements read
+  // `streamFrame`.
+  const frame = useSelectionHold(streamFrame, scrollerRef);
   const measureRef = useRef<HTMLSpanElement>(null);
   const keyboardLayoutRef = useRef<KeyboardLayoutReader | null>(null);
   useEffect(() => {
@@ -1167,8 +1173,8 @@ export function MobileLiveTerminal({
   // is a transport event, not derived state, so an effect is the right hook.
   useEffect(() => {
     // eslint-disable-next-line react-you-might-not-need-an-effect/no-event-handler
-    if (frame) notchPacer.onFrame();
-  }, [frame, notchPacer]);
+    if (streamFrame) notchPacer.onFrame();
+  }, [streamFrame, notchPacer]);
 
   const onWheel = useCallback(
     (e: React.WheelEvent) => {
@@ -1605,8 +1611,8 @@ export function MobileLiveTerminal({
 
   const [frameTiming] = useState(() => new FrameTimingProbe());
   useLayoutEffect(() => {
-    if (LIVE_DEBUG && frame) frameTiming.record(performance.now(), frame.receivedAt);
-  }, [frame, frameTiming]);
+    if (LIVE_DEBUG && streamFrame) frameTiming.record(performance.now(), streamFrame.receivedAt);
+  }, [streamFrame, frameTiming]);
 
   // --- bottom pinning ---------------------------------------------------------
   useLayoutEffect(() => {
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -645,11 +645,11 @@ export function MobileLiveTerminal({
     setFontSize(configuredFontSize);
   }
   const scrollerRef = useRef<HTMLDivElement>(null);
-  // A selection anywhere in the grid pins the painted frame until the user
-  // lets go, so no row is rewritten out from under the range. Everything
-  // below renders that held frame; only the stream acknowledgements read
-  // `streamFrame`.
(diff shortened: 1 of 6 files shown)


web/src/components/MobileLiveTerminal.tsx
```diff
@@ -644,6 +645,11 @@ export function MobileLiveTerminal({
     setFontSize(configuredFontSize);
   }
   const scrollerRef = useRef<HTMLDivElement>(null);
+  // A selection touching the grid pins the painted frame until the user lets
+  // go, so no row is rewritten out from under the range (see the hook).
+  // Everything below renders that held frame; only the stream
```
jerome-benoit, line 659: [P2] Allow uncaptured scrollback to populate during selection

Start a selection at the live edge with history longer than the small live capture window, then scroll upward to extend it. `onScroll` calls `enterReading` (1059-1060), which requests the full history (`useLiveTerminal.ts:592-599`), but this hold ignores that response too. `lines` and the blank history spacer still come from the held frame (844,878-881), so older text remains unavailable until the selection is cleared. Preserve the selected rows while allowing previously uncaptured history to populate, and cover this transition with nonzero history and a selection started before entering reading mode. The current tests only exercise `history: 0`.

jerome-benoit (changes requested):
Reviewed `3e6ae8a4` against `0bd6dd53`, including existing discussion, with two independent sequential analyses and arbitration. One introduced P2 blocks approval: freezing the entire frame also prevents loading older scrollback into an ongoing selection.

The stream acknowledgements, release/back-to-live path and session-key isolation were checked without another finding. Current CI is green; the new tests use `history: 0` and do not cover the inline scenario. Evidence: exact-SHA `git show`/diff inspection and clean `git diff --check`. This is a source-traced regression, not a browser reproduction or an iOS callout validation; no local tests were run.

web/src/components/MobileLiveTerminal.tsx
```diff
@@ -644,6 +654,31 @@ export function MobileLiveTerminal({
...
+    // Reading mode is the only thing that widens the window, and it is also
+    // what mounts every row: outside it the debounced row count lags a
+    // sudden jump in height and virtualization would unmount the selected
+    // row, which is the collapse this whole change exists to prevent.
+    if (!selectionHeld || !reading || !heldFrame || !streamFrame || heldFrame === streamFrame) return heldFrame;
+    const held = frameLines(heldFrame);
+    const next = frameLines(streamFrame);
+    const older = heldFrame.history - held.length - (streamFrame.history - next.length);
+    if (older <= 0) return heldFrame;
+    return { ...heldFrame, lines: next.slice(0, Math.min(older, next.length)).concat(held) };
```
jerome-benoit: [P2] Hold the newly exposed history once it becomes selectable

With the VT history full (`SCROLLBACK_LINES = 2000`), start a selection in the small live window, scroll up to load history, and extend the selection into that added prefix. Another output line evicts the oldest history line: `history` and capture length remain constant (`vt.rs:1769-1773`), so `older` stays constant but this slice shifts its text under the same row keys. The prefix is not part of `heldFrame`, so the selected text is rewritten despite the hold. Preserve the prefix once exposed for the lifetime of the selection, and test a subsequent shifted frame with unchanged history depth/capture length and an endpoint inside the prefix. The new test covers only the initial expansion. This is a static code-path finding, not a browser reproduction.

jerome-benoit (changes requested):
The initial scrollback-loading issue is fixed, and touch forwarding now yields during selection without changing row layout. One P2 remains in the newly added history prefix, detailed inline.

Reviewed the delta independently twice and cross-checked the remaining case against the capped VT history contract. Static validation only; no local tests run. The new selection tests passed in CI, but the live shard failed on `session-group-edit.spec.ts:71` (expected `team/alpha`, received `[]`). That failure is not established as caused by this change; GitHub currently refuses a rerun while the workflow is still active.

jerome-benoit (changes requested):
The previous P2 is fixed: exposed scrollback is now folded into the held frame, and the new capped-history test covers the eviction case.

One P2 remains, already described accurately by CodeRabbit at `MobileLiveTerminal.tsx:683`, so I am not duplicating the inline annotation. If a received frame carries less than the prefix calculated as `older`, the partial fold repeats on the render path and duplicates the held text after a sharp history reduction. Fold only a complete prefix, and add a short-frame/history-reduction regression test.

Two independent sequential reviews cross-validated this conclusion. Exact-SHA static review only; no local tests run. CI is green at `cad3713`.

jerome-benoit (approved):
Re-reviewed `2c528754` after the previous request changes.

The partial-history guard now avoids folding an incomplete prefix, and the regression test covers the sharp `history: 2000 -> 0` reduction while selection is held. The prior P2 is resolved; I found no remaining actionable finding.

Two independent sequential static reviews were cross-validated against the current diff, existing discussions, and the held-frame invariant. CI is still completing two non-web checks; all relevant web checks, including Vitest and both live Playwright shards, are green at this SHA. No local tests were run.

Issue #3648, opened 2026-08-31
Multiline tmux environment test depends on locale behavior


jerome-benoit:
## Problem

`tmux::env::tests::test_batch_output_frames_records_against_a_real_tmux` assumes a `LC_ALL=C` tmux client flattens multiline hidden-environment values. On Linux with tmux 3.7c, the C-locale leg preserves the newline, so the test fails before checking the parser:

```text
assertion left == right failed: C
left: true
right: false
```

The raw block correctly contains both the real `AOE_INSTANCE_ID` record and the spoofing continuation. Locale behavior is tmux/platform-dependent; parser correctness does not require one locale to flatten it.

## Acceptance criteria

- Do not assert a fixed flatten/preserve result for `LC_ALL=C`.
- Always assert that `parse_batch_output` returns `real-id` for every raw form produced by an available locale.
- Preserve coverage of both flattened and multiline forms with deterministic pure parser fixtures.
- Keep one real-tmux smoke path without encoding locale-specific output behavior.

Observed while validating #3570; unrelated to that PR's Qwen/Kiro resume scope. Related behavior fix: #3616.

njbrake:
Fixed by #3678 (`b2569c75e`), incidentally rather than by name. The `("C", false)` leg and its `assert_eq!(spoofed, expect_continuation)` are gone; `test_batch_output_frames_records_against_a_real_tmux` now asserts only that `parse_batch_output` returns `real-id` for whatever raw form each available locale produces. That covers acceptance criteria 1, 2 and 4.

Criterion 3 is partly open: `test_parse_batch_output_ignores_multiline_continuations` has pure fixtures for the multiline form but none for the flattened one. Worth a one-line addition if you want that shape pinned too.

Not reproducible here to confirm against 3.7c: this box has tmux 3.6, where the C locale does flatten.

_Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #3513, opened 2026-08-24
test_export_exec_compound_command_passes_env is timing-flaky (tmux bare core)


jerome-benoit:
Observed on PR #3488's CI: `test_export_exec_compound_command_passes_env` (`src/tmux/mod.rs`, bare-core tmux export path) went red while the branch was green locally, then passed on retry.

Reproduction notes:
- Locally at PR HEAD: **3/3 green** runs (`cargo test --lib test_export_exec_compound_command_passes_env`, ~2.2s each).
- The test drives a real tmux session for the compound-command export path, so it inherits scheduler/PTY timing jitter; nothing in the failing run pointed at the export logic itself.

Suggested directions: assert on exported state rather than wall-clock ordering, add a retry tolerance for the pane-poll step, or tag it like the other timing-sensitive tmux tests.

(Filed from the #3488 review cycle; the PR body will reference this issue for its currently-red CI job.)

njbrake:
Fixed by #3750. The test no longer sleeps 1000 ms: it polls the pane capture to a 15 second deadline, and `remain-on-exit on` is chained into the session create, so waiting past the process exit cannot lose the output. 5/5 green locally on tmux 3.6.

_Note: this comment was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

Issue #3479, opened 2026-08-22
xtask check-skill: the "command does not exist" arm is unreachable, so the check has never been able to fail


blueagledev:
Your correction on #3470 holds one level deeper than you put it: the error arm
in `check-skill` is not just narrow, it is unreachable, and has been since #442.
Line numbers against `74813729`.

`skill_commands` is only inserted into after `cli_commands.contains(...)` has
already accepted the value, in both the longest-prefix loop (`:526-535`) and the
first-word fallback (`:537-539`). The arm at `:546-559` then asks
`!cli_commands.contains(skill_cmd)`, which is unsatisfiable for a set built that
way. A bad subcommand under a good prefix truncates to the prefix; a wholly
unknown command records nothing at all.

Your two cases, plus the third shape:

```
aoe session bogusverb   Skill check passed.  exit=0
aoe session pin         Skill check passed.  exit=0
aoe totallybogus        Skill check passed.  exit=0
```

`aoe totallybogus` is the quiet one: `best` stays empty, nothing is recorded, so
unlike your two it does not even move the advisory.

Control: replacing the insert so it records what was read rather than what
matched makes the arm fire with its exact text and exit 1; restoring only that
guard, bad line still in place, returns exit 0 with an empty stderr. The
neutered run emits 15 errors, of which 1 is the injected line and 14 are
pre-existing false positives on the untouched files, which is the next section.

No live instances: no `aoe ...` reference in either skill file names a command
that does not exist, and `check-skill` exits 0 on the tree as it stands.

## What a fix has to clear

I tried four shapes; each ran into one of three constraints, all reproduced:

- **`aoe` is an ordinary English word in these files.** Resolving to the longest
  real prefix and flagging trailing words is clean on the current files and does
  catch your two cases, but it also flags
  `Use aoe session to manage agents.`
- **Requiring an inline backtick costs coverage.** Almost every command sits in
  a fenced block, so 16 paths stop being extracted and 3 more (`group`,
  `profile`, `worktree`) lose coverage through the advisory's prefix rule. The
  advisory goes from 89 entries to 108.
- **Scanning code context fixes both, then breaks on aliases and on comments.**
  `collect_subcommand_paths` records `sub.get_name()` only, so the 12
  `#[command(alias = ...)]` in `src/cli/` are invisible to it and `aoe group ls`
  reads as nonexistent. They are hidden aliases, absent from
[...]

PR #3848, opened 2026-09-09
test: repair four tmux/xtask tests that could not fail, and make ENV_LOCK the PATH exclusion

njbrake:
## Description

Three test-infrastructure issues whose common shape is a test that could not fail. Two more from the same cluster turned out to be already fixed on `main`; evidence at the bottom, no code here for either.

**#3368, tmux targeting.** Setting `base-index 1` on a live session does not renumber the window it was created with, so the three tests justified by "with base-index 1, window 0 does not exist" ran against a session that still had a window 0. The first window now moves to index 1 after the option is set, and the helper asserts window 0 is gone. tmux resolves a missing window index to the session's current window, so `:0.0` then reads the active window rather than failing.

Same file, three more gaps. `capture_pane`'s `.expect(...)` could not fire (the call returns `Ok(String::new())` on a failed tmux invocation) and the content was discarded; the first window's pane now prints a marker and the test asserts the capture contains it. The dead-pane test slept 300 ms and set `remain-on-exit` at a scope the second window never inherited, so that window closed and the active window was the agent's own; its pane now blocks on a release file until `remain-on-exit` is on it, and the test polls `#{pane_dead}`. `test_status_checks_with_split_panes_and_pane_base_index_1` set `pane-base-index 1` nowhere and duplicated the test above it; it now sets the server-global option, which is what a user actually sets, and builds the session through the production `append_*_args` helpers.

**#3479, `xtask check-skill`.** `skill_commands` was only inserted into after `cli_commands` had accepted the value, so the arm asking `!cli_commands.contains(...)` was unsatisfiable. Worse than inert: `aoe session bogusverb` truncated to `session` and, through the advisory's prefix rule, suppressed the twelve `aoe session *` entries. Scoped down per your comment on the issue, following @blueagledev's design: read code context only (fenced blocks and paired inline spans, line-local), record what was read rather than what matched, treat a trailing word as a subcommand only where the resolved path takes subcommands, and collect `get_all_aliases()`. Output on the current tree is byte-identical to before, exit 0, advisory unchanged at 92.

xtask's tests also now run in CI. The workspace declares no `default-members`, so every `cargo test` in the workflows was scoped to the root package and the xtask tests ran only on a contributor's machine. Clippy for xtask is deliberately left out: it needs a second lib build under different features.

**#3469, PATH scrubbers onto `EnvGuard`.** The seven save/restore pairs excluded their readers by `#[serial]` key rather than by `ENV_LOCK`. `git grep 'set_var("PATH"' -- src/` now returns nothing. Three comments documented that as a reason not to reach for the guard; all three are gone, and `#[serial]` comes off the three readers they justified. `path_prepended` also drops an empty inherited `PATH` rather than splitting it: `split_paths("")` yields one empty entry, which means the current directory, so the `format!("{prefix}:{prev}")` it replaces put the caller's cwd on the test's `PATH`.

### How each new assertion was shown to fail

| mutation | result |
| --- | --- |
| `^.0` back to `:0.0` in `is_pane_dead`, `pane_current_command`, `pane_start_command_is_protected`, `Session::capture_pane` | 3 of the 4 targeting tests fail; all 4 passed before this PR |
| `append_pane_base_index_args` appends nothing | `test_status_checks_with_split_panes_and_pane_base_index_1` fails |
| `check-skill` records nothing for an unresolvable word | the extractor test fails |
| trailing word after a parent is not checked | the extractor test fails |
| aliases are not collected | the alias test fails |
| whole file scanned instead of its code spans | the extractor test fails |
| a flag after a parent command is read as a subcommand | the extractor test fails |
| a `#` comment inside a fence is read as an invocation | the extractor test fails |
| the reader takes no `ENV_LOCK` | the exclusion test fails on the visible mutation |
| `path_prepended` restores with a raw `set_var` | the exclusion test fails |
| `path_prepended` reads `PATH` before taking the lock | the derivation test fails |

The one test that stays green under the `^.0` → `:0.0` mutation is `test_status_checks_with_split_panes_and_pane_base_index_1`, correctly: it covers the `.0` half, which that mutation does not touch. It has its own mutation above.

Injected into a fenced block in `contrib/openclaw-skill/SKILL.md`, before and after:

| line | before | after |
| --- | --- | --- |
| `aoe totallybogus` | exit 0, advisory 92 | exit 1, advisory 92 |
| `aoe session bogusverb` | exit 0, advisory 80 | exit 1, advisory 92 |
| `aoe session pin` | exit 0, advisory 80 | exit 1, advisory 92 |
| `aoe ls`, `aoe rm x`, `aoe profile mv a b`, `aoe group create g`, `aoe group ls`, `aoe session capture id` | exit 0 | exit 0 |

The first version of the `ENV_LOCK` exclusion test passed both of its mutations, because the reader's 500 lock/read iterations finished before the peer thread had scrubbed anything. It was rewritten to hold the peer's guard across a real window.

### Review round (commit 4)

A review caught three isolation defects, all reproduced before being fixed. The largest was mine and made things worse than `main`: moving the `omp` capture-gate test from a child-scoped `PATH` to a process-global one and dropping `#[serial]` put a fake `tmux` on every concurrent test's `PATH`. `tmux_available()` reads the shim's exit 0 as tmux being present, so tmux tests ran against the shim rather than skipping. `cargo test --lib -- tmux::session::tests::a_zoomed_pane session::instance::omp::tests::omp_capture_gate --test-threads=8` failed 3/3, and in full-suite runs it also tripped this branch's own new assertion. The override is child-scoped again and the test holds `ENV_LOCK` only for the read that builds it; 3/3 green after.

`path_prepended` also read `PATH` before taking the lock, so the helper written to close that hole had it. `EnvGuard::and_set` chains onto `read_lock` so the derivation happens inside the lock. `GlobalPaneBaseIndex` could leak the global on a failed `show-options`, and `check-skill` had gained two false positives (`aoe group --help`, and a `#` comment inside a fence). Each is now a table case that fails without its fix.

### One thing I could not settle

`session::smart_rename::tests::resolve_smart_rename_config_reads_repo_aware_config_but_not_repo_commands` went red once during the review, on the pre-fix code. It is 0/8 red on the current branch. Its state is `HOME` / `XDG_CONFIG_HOME`, set at `src/session/smart_rename.rs:2547` with a raw unguarded `set_var` behind only `#[serial]` and never restored, while fifteen non-serial tests mutate the same state through `AppDirGuard` under `ENV_LOCK`. That is a pre-existing instance of this PR's own bug class in a file this PR does not touch, and none of the three tests de-serialized here reads or writes `HOME`. I could not build a `main` baseline to prove attribution: the box is at 100% disk with 1.4 GB free and a second debug target directory does not fit. Flagging it rather than claiming it is unrelated.

### #3513 and #3648: already fixed, nothing changed here

**#3513** was fixed by #3750 (`src/tmux/mod.rs`): the 1000 ms sleep is a poll to a 15 second deadline, and `remain-on-exit on` is chained into the create so waiting past the exit cannot lose the output. 5/5 green locally on tmux 3.6. Recommend closing.

**#3648** was fixed by #3678 (`b2569c75e`), incidentally rather than by name. The assertion the issue reports, `assert_eq!(spoofed, expect_continuation)` with `("C", false)`, is the one that commit deleted; `test_batch_output_frames_records_against_a_real_tmux` now asserts only that `parse_batch_output` returns `real-id` for whatever raw form each available locale produced, which is acceptance criteria 1, 2 and 4. Criterion 3's multiline fixtures are in `test_parse_batch_output_ignores_multiline_continuations`; the flattened form has no fixture of its own, which is a one-line gap I have not filled. I could not reproduce the failure: this box has tmux 3.6, where `LC_ALL=C` does flatten the newline to `_`, so even the pre-#3678 assertion holds here. The tmux 3.7c behavior the issue reports is real, and the code that depended on it is gone. Recommend closing.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [x] Infrastructure / CI

## Checklist

  - `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo clippy -p xtask --all-targets -- -D warnings`, `cargo test`: 6872 lib + 289 integration passed, 4/4 repeat runs green. `cargo test -p xtask`: 6 passed, 3 of them new.
  - No user-facing behavior changes. The three stale `EnvGuard` caveats are corrected where they sit.
  - No UI changes.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Every change here is to test code or to the check that gates it. No production behavior changes; the mutation table above is the coverage evidence.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** The mutation runs, the before/after `check-skill` matrix and the tmux target-resolution probes were executed and reviewed by @njbrake.

https://claude.ai/code/session_01YWwvnDwmKNLDnVB1kmcjWi

## Summary

- Strengthened tmux tests for window targeting, pane output, dead-pane handling, and user index settings.
- Improved `xtask check-skill` validation for aliases, subcommands, flags, comments, and code spans.
- Added xtask and `aoe-plugin-api` test coverage to CI.
- Centralized test `PATH` isolation with `EnvGuard` and `path_prepended`.
- Replaced manual environment cleanup and obsolete test serialization.
- Added tests for environment isolation and fixed review-discovered test leaks and false positives.

## Benefits

These changes make regression tests more reliable and improve CI coverage. They also prevent test environment leaks without changing production behavior.

### DIFF
--- a/src/session/test_support.rs
+++ b/src/session/test_support.rs
@@ -240,7 +240,6 @@ pub(crate) fn path_prepended(dir: &Path) -> EnvGuard {
 /// Install a PATH command that remains first after the test pane starts its
 /// login shell.
 pub(crate) fn install_login_shell_path_command(root: &Path, name: &str, script: &str) -> EnvGuard {
-    let guard = EnvGuard::read_lock();
     let bin = root.join("bin");
     std::fs::create_dir_all(&bin).expect("create test bin directory");
     let executable = bin.join(name);
@@ -261,17 +260,7 @@ pub(crate) fn install_login_shell_path_command(root: &Path, name: &str, script: 
         ),
     )
     .expect("write test login profile");
-    let path = std::env::join_paths(
-        std::iter::once(bin).chain(
-            std::env::var_os("PATH")
-                .iter()
-                .flat_map(|value| std::env::split_paths(value)),
-        ),
-    )
-    .expect("join test PATH");
-    guard
-        .and_set("PATH", path)
-        .and_set("SHELL", OsString::from("/bin/sh"))
+    path_prepended(&bin).and_set("SHELL", OsString::from("/bin/sh"))
 }
 
 /// Restores the process-global tied-worktree setting even when a test panics.
--- a/src/session/test_support.rs
+++ b/src/session/test_support.rs
@@ -168,6 +168,18 @@ impl EnvGuard {
         guard
     }
 
+    /// Take [`ENV_LOCK`] without mutating anything, for a test that only
+    /// *reads* the process environment and must not observe another guard's
+    /// mutation mid-read. `#[serial]` cannot do that job: it excludes only
+    /// tests sharing its key, which is what left the reads in
+    /// `session::instance::omp` and `tmux::session` open (#3469).
+    pub(crate) fn read_lock() -> Self {
+        Self {
+            prev: Vec::new(),
+            _lock: acquire_env_lock(),
+        }
+    }
+
     fn snapshot(&mut self, key: &'static str) {
         self.prev.push((key, std::env::var_os(key)));
     }
@@ -190,6 +202,25 @@ impl Drop for EnvGuard {
             ENV_LOCK_HELD.with(|held| held.set(false));
         }
     }
+}
+
+/// Put `dir` first on `PATH` for the rest of the scope.
+///
+/// The hand-rolled save/restore pairs this replaces excluded each other by
+/// `#[serial]` key rather than through [`ENV_LOCK`], so a guard user carrying
+/// a different key could read a scrubbed `PATH` mid-test (#3469). They also
+/// restored with `set_var` unconditionally, which leaves an empty `PATH`
+/// behind on a host that inherited none; [`EnvGuard`] removes it instead.
+pub(crate) fn path_prepended(dir: &Path) -> EnvGuard {
+    // An empty `PATH` is dropped rather than split: `split_paths("")` yields
+    // one empty entry, and an empty entry means the current directory, so
+    // keeping it would put the caller's cwd on the test's `PATH`.
+    let inherited = std::env::var_os("PATH").filter(|value| !value.is_empty());
+    let path = std::env::join_paths(
+        std::iter::once(dir.to_path_buf()).chain(inherited.iter().flat_map(std::env::split_paths)),
+    )
+    .expect("join test PATH");
+    EnvGuard::set(&[("PATH", path)])
 }
 
 /// Install a PATH command that remains first after the test pane starts its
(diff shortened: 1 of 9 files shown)


src/session/test_support.rs
```diff
@@ -519,6 +567,117 @@ mod tests {
...
+            let _read = EnvGuard::read_lock();
+            std::env::var_os("PATH")
+        };
+        let shim = TempDir::new().unwrap();
+        let scrubbed = AtomicBool::new(false);
+        let reader_waiting = AtomicBool::new(false);
+
+        let built = std::thread::scope(|scope| {
+            scope.spawn(|| {
+                let scrub = EnvGuard::unset(&["PATH"]);
```
jerome-benoit, line 643: [P2] Exclude existing tmux tests while unsetting process PATH

This new test has no `#[serial_test::serial]`, so its worker removes process-wide PATH for at least 50 ms while existing tmux tests can run. Those tests use the default serial key, not `ENV_LOCK` (`src/tmux/session.rs:4136-4142`), and `tmux_command()` still uses `Command::new("tmux")` (`src/tmux/mod.rs:211`). With tmux installed only through a custom PATH, such as Homebrew/Nix, overlap can make `tmux_available()` silently skip coverage or make a subsequent command fail. Retain the default serial key on this test until those readers share the environment lock. That does not mask this test's inter-thread oracle: its spawned worker and reader still need `ENV_LOCK` to exclude each other.

jerome-benoit (changes requested):
Reviewed `9c86ce02` against `0bd6dd53`, including the linked issues and existing review. Two independent sequential analyses and arbitration identified one introduced P2: the new PATH-unsetting test can run alongside existing unguarded tmux readers.

The tmux targeting assertions, command-check error path and xtask CI step were checked without another blocker. CodeRabbit's [empty-PATH finding](https://github.com/agent-of-empires/agent-of-empires/pull/3848#discussion_r3964146611) is valid but pre-existing, P3/non-blocking; reusing `path_prepended` would remove that remaining duplication.

Evidence: exact-SHA diff/source inspection and clean `git diff --check`. CI is green at this head. No local tests or mutation runs were performed; the inline race is established from the competing code paths, not claimed as locally reproduced.

jerome-benoit (approved):
Re-reviewed the changes since `9c86ce02` with two independent sequential analyses and cross-validation. The P2 is fixed: the default serial key excludes the existing tmux readers while PATH is unset, without weakening the worker/reader ENV_LOCK oracle. CodeRabbit's PATH helper finding is also addressed.

The alias traversal, shell-fence filtering, tmux option restoration and additional workspace-member CI coverage introduce no further finding in this delta. Exact-SHA source inspection and `git diff --check` only; no local tests run. CI is still queued, so this approves the code review, not a claim of passing checks.

PR #3858, opened 2026-09-09
fix(sandbox): keep the host and sandbox Claude credential chains apart

njbrake:
## Description

Four standalone commits on the shared Claude credential file (#3834, #3846).

1. A running container created before the file was shared is no longer folded. Its store copy is a chain it keeps rotating; folding it logged every sandbox on the shared file out until it stopped.
2. A store copy is emptied only when the shared file holds a credential at least as fresh, compared at placement since a poll can write the file after the fold. A failed or skipped fold used to empty the only fresh copy.
3. The 30-minute TUI refresh and a relaunch of a running container seed only. The guide promises a host login carries over on start or restart; the poll put every live sandbox back on the host's refresh token mid-session.
4. **Open question.** The host file and Keychain never replace a credential the shared file holds, even at start. Ends the host/sandbox coupling (both files hold one refresh token on the reporting machine today) at the cost of "log in on the host, restart a sandbox": log in inside any sandbox instead, or delete `sandbox-v2/.credentials.json` and start one. Droppable without touching 1 to 3.

Not done: rejecting expired candidates. `expiresAt` is the access token's expiry; an idle sandbox's copy is expired with a live refresh token, and rejecting it would let a staler copy over it.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the fold rules are unit tested through the store preparation path (`off_a_come_up_the_fold_only_seeds`, `shared_credential_follows_the_freshest_copy_across_starts`, `shadowed_credential_copy_is_emptied_only_in_a_store`); the label skip reads a container label, which needs a runtime the harness skips without.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Audit, diagnosis and implementation by the model in discussion with @njbrake, who owns the decisions.

_Note: this PR description was drafted by Claude Fable 5.1 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

https://claude.ai/code/session_014GEwxHbRR8Tg8fYn74Z6Yd

## Summary

- Separate host Claude credentials from sandbox credential chains.
- Preserve existing shared credentials during refreshes and relaunches.
- Prevent older running containers from receiving unintended credential updates.
- Keep legacy stopped-container behavior.
- Add container-label detection and credential-folding rules.
- Update sandbox credential documentation.
- Add unit coverage for credential folding and placement.

## Benefits

This prevents host or Keychain credentials from replacing sandbox credentials. It makes credential refresh behavior safer and more predictable.

## Potential follow-up

Add an end-to-end test for the container-label scenario when the test harness supports it.

### DIFF
--- a/docs/guides/sandbox.md
+++ b/docs/guides/sandbox.md
@@ -359,14 +359,21 @@ macOS, `~/.claude/.credentials.json` elsewhere, and any copy left in the
 session's store by an earlier layout, each taken only when its expiry is later
 than what the file holds. Logging in on the host and starting or restarting
 one sandboxed session re-authenticates all of them; a login made inside a
-container is never overwritten by a staler host copy. The host itself stays a
-separate chain: a refresh inside a sandbox still rotates the token the host
-holds, as it did before. A container created before this layout mounts only
-its store, so a stopped one is recreated at its next start, as it was for the
-store move, and a running one refuses to relaunch until it is stopped. Running
-`/logout` inside a sandbox revokes the token for every sandbox but cannot
-remove the mounted file, so the revoked token stays there until the next
-login.
+container is never overwritten by a staler host copy. Between starts the file
+belongs to the containers: the TUI's periodic refresh only seeds a file that
+holds no credential, so a host refresh never replaces a token the sandboxes
+are rotating mid-session.
+
+The host and the sandboxes are not separate chains once a start has copied
+the host's token in: both then hold the same refresh token, and whichever
+refreshes first logs the other out, as it did before this layout. A login
+made inside a container starts a chain of its own, which the next start
+replaces again if the host's token is fresher by then. A container created
+before this layout mounts only its store, so a stopped one is recreated at its
+next start, as it was for the store move, and a running one refuses to
+relaunch until it is stopped. Running `/logout` inside a sandbox revokes the
+token for every sandbox but cannot remove the mounted file, so the revoked
+token stays there until the next login.
 
 ### Reclaiming stores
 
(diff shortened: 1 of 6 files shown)


jerome-benoit (approved):
Reviewed `288d268115ea70d4aa29a413b53795d0e5725d0e`. No blocking findings after two independent sequential analyses and arbitration against the base and existing discussion.

Verified refresh/relaunch versus create/start fold policies, freshness rechecking before clearing the private copy, and exclusion of shared credentials from generic config sync. Other agents do not declare shared credential files.

On the explicit policy question: keeping host/Keychain credentials seed-only is consistent with avoiding repeated interference with the sandbox refresh chain. The loss of automatic host-login propagation is deliberate and documented, including recovery. This approval covers that behavior as implemented.

Non-blocking residual limitation: the [cached stopped state](https://github.com/agent-of-empires/agent-of-empires/blob/288d268115ea70d4aa29a413b53795d0e5725d0e/src/tui/status_poller.rs#L198-L205) is not a synchronization guarantee against external container starts. One analysis proposed blocking on this; comparison with the base shows a pre-existing path narrowed by this patch, while normal AoE startup recreates legacy containers. I do not require that separate hardening here.

CI is green at this SHA. Static review only: no local tests, container runs or live credential-rotation experiments were performed.

PR #3868, opened 2026-09-09
fix(tui): keep structured paste drafts per target and consume them on drain

jerome-benoit:
## Description

Closes #3562. Review findings on #3562 (head `athal7:fix/structured-message-shortcut` is not pushable from here) are fixed on this branch, built on that PR's head; merging this one supersedes it.

The buffered structured paste was a single `(target, text)` slot with two defects:

- forwarding it into the mounted composer left the slot populated, so every later `m` activation for that session re-inserted the same text (CodeRabbit, Major);
- a paste captured for another session replaced the slot outright, permanently destroying the previous target's unsent draft even though returning to that session remains possible (P2).

Drafts now live in a map keyed by their target session (`pending_paste_for_structured_view: HashMap<String, String>`). The mounted view consumes its own entry on success; entries for other targets stay put, so a failed activation is recovered by pressing `m` on that session again.

## PR Type

- [x] Bug Fix

## Checklist

- [x] Documentation was updated where necessary (field and function doc comments updated to the per-target semantics)
- [x] For UI changes: included screenshot or recording (no visual change; paste buffering behavior only)

## Test Coverage Analysis

- [x] N/A: TUI-side behavior covered by Rust unit tests

**TUI / CLI (e2e test)**
- [x] N/A: covered by the home unit tests for the 'm' routing

## AI Usage

- [x] AI was used

**AI Model/Tool used:** omp (OpenCode) with Omen Alpha

**Any Additional AI Details you'd like to share:** Fixes authored and verified in an AI session under Jérôme Benoit's review workflow.

## Summary

- Routes the home `m` shortcut to the structured composer for structured sessions.
- Stores pending paste text per session.
- Preserves drafts when activation fails and prevents cross-session leakage.
- Loads pasted file mentions through the asynchronous file-loading path.
- Adds regression tests for routing, draft merging, session isolation, and daemon-start handoff.

These changes make structured message entry more reliable and prevent repeated or misplaced paste insertion.


jerome-benoit (commented):
Reviewed `67c53ad3c853d0dfe59079f13b06d8075479cefd`, including the complete #3562 feature and previous replies. COMMENT because this is a self-authored PR.

**One P2 still needs correction before merge:** [drain after daemon-start activation](https://github.com/agent-of-empires/agent-of-empires/pull/3868#pullrequestreview-5160871298). `src/tui/app.rs:3222-3223` connects and activates without calling the drain. With the daemon stopped, capture a paste at home, press `m`, and accept startup: the composer opens empty while the text remains in the map until another activation. Invoke the same target-checked handoff on this success continuation and test insertion exactly once. This is an existing finding, not a new inline.

Verified fixes: `app.rs:3167-3171` consumes only the mounted session's entry; `home/input.rs:2870-2875` merges same-target text and preserves other targets; `structured_view/embedded.rs:235-237` shares insertion and mention-file loading. The home tests verify storage/routing, not the App-level continuation or consumption. Nonblocking: shorten the repeated field documentation in `home/mod.rs:443-455` and remove its obsolete tuple description.

Evidence: exact-SHA `git show`, base/head and original-head diffs; `git diff --check dd355b8...67c53ad3` exits 0 with no output. CI has no failed or pending check. No local tests/builds run.

Two independent sequential analyses and parent cross-validation agree on this single remaining P2; no substantive disagreement.

jerome-benoit (commented):
Re-reviewed `a2471f1bff32aca912e8d9a891a51a882af36aa9`, including the #3562 history. COMMENT because this is my own PR.

The previous P2 is fixed: `src/tui/app.rs:3224-3230` now drains after daemon-start activation. The App-level test verifies insertion, consumption and preservation of another target; it does not run the daemon-start continuation itself.

The existing stale-preview concern needs narrower scope. `setup_view` (`src/tui/structured_view/mod.rs:265-277,325-343`) converts WebSocket/replay failures into a mounted view with a toast. Ordinary connection failure therefore does not establish the reported stale-view path; the propagated error is HTTP-client construction (`src/acp/client/http.rs:132-137`). Passing the requested ID to the drain remains sensible defensive hardening, but I have not established a normal-runtime P2 reproduction. Do not describe this as cross-session text leakage: the map lookup sends A's draft to A.

No demonstrated merge blocker remains in this bounded static pass. Evidence: exact-SHA source inspection, `git diff --check dd355b8...a2471f1b` clean, live CI with no failed/pending checks. No local tests or runtime reproduction performed.

Cross-validation: two independent analyses initially retained the existing P2. Parent arbitration checked the actual error boundary above; the first reviewer then withdrew that classification. The unreproduced client-construction failure remains an explicit limitation, not a proven connection-failure regression.

PR #3878, opened 2026-09-10
docs(tui): describe live-send's socket transport as the default

njbrake:
## Description

`live_send.rs`'s module doc describes the pre-VT design. It says each translation "produces a `tmux send-keys` call", and a trade-off bullet costs every coalesced keystroke run at one subprocess. Both have been the fallback since the VT channel landed.

While a channel is armed for input (`[tmux] vt_live`, tmux 3.8 or newer, unix) the action is encoded to raw terminal bytes and written into the pane socket, bypassing tmux's key translation (`live_send.rs:2236`, `encode_action_bytes`). The single-writer invariant means no keystroke takes the `send-keys` path at all while that channel is live. `send-keys` covers tmux older than 3.8, a forwarder that has not connected or has died, and `vt_live` off.

Doc comment only, no code change. Prompted by #3340, which proposes reusing this transport for remote tmux rows: the stale text sends a reader looking for a key-name protocol that is no longer the one in use.

## PR Type

- [x] Documentation

## Checklist

- [x] For UI changes: included screenshot or recording (N/A, no UI change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Found while validating #3340's live-ws claims against `main`. The staleness was what led the reviewing agent to the wrong conclusion first time through, which is the argument for fixing it. `cargo fmt`, `cargo clippy` and `cargo doc` clean.

https://claude.ai/code/session_01PzjnMgi5yGTRGiwEVz3ye5

## Summary

- Updated `live_send.rs` documentation to describe the VT socket transport as the default for `vt_live` on tmux 3.8+ Unix systems.
- Documented `tmux send-keys` as the fallback for older tmux versions, unavailable forwarders, or disabled `vt_live`.
- Clarified raw terminal byte handling and noted that the socket path avoids per-keystroke process creation.

This documentation-only change improves accuracy and reduces confusion about live mode’s current transport.

### DIFF
--- a/src/tui/home/live_send.rs
+++ b/src/tui/home/live_send.rs
@@ -2,11 +2,26 @@
 //!
 //! When a user presses `Tab` on a runnable session, the home view installs
 //! a `LiveSendState` and routes every subsequent key event through this
-//! module's translator. Each translation produces a `tmux send-keys` call
-//! against the target pane: plain characters go literally, every other
-//! key (arrows, Esc, Tab, modifier combos) goes by tmux key name with
-//! `C-` / `M-` prefixes. The user exits with one of the configured
-//! exit chords (default: `Ctrl+q`).
+//! module's translator. Each translation produces a `TmuxAction`, delivered
+//! by whichever transport the pane has armed.
+//!
+//! While a VT channel carries input (`[tmux] vt_live`, tmux 3.8 or newer,
+//! unix) the action is encoded to raw terminal bytes and written into the
+//! pane socket, bypassing tmux's own key translation, so the encoder honors
+//! the pane's DECCKM state itself. That is the default. Presence of a live
+//! input channel is a single-writer signal: every keystroke then goes over
+//! the socket and none through `send-keys`, since the two writers would
+//! interleave on one pty input stream.
+//!
+//! Otherwise each action forks `tmux send-keys`: plain characters go
+//! literally, every other key (arrows, Esc, Tab, modifier combos) by tmux
+//! key name with `C-` / `M-` prefixes. This is the path for tmux older than
+//! 3.8 (through 3.7a, writing to a dead pane's input takes the whole tmux
+//! server down), a pane whose forwarder has not connected or has died, and
+//! `vt_live` turned off.
+//!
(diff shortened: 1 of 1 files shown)


src/tui/home/live_send.rs
```diff
@@ -2,11 +2,26 @@
...
+//! module's translator. Each translation produces a `TmuxAction`, delivered
+//! by whichever transport the pane has armed.
+//!
+//! While a VT channel carries input (`[tmux] vt_live`, tmux 3.8 or newer,
+//! unix) the action is encoded to raw terminal bytes and written into the
+//! pane socket, bypassing tmux's own key translation, so the encoder honors
+//! the pane's DECCKM state itself. That is the default. Presence of a live
+//! input channel is a single-writer signal: every keystroke then goes over
+//! the socket and none through `send-keys`, since the two writers would
+//! interleave on one pty input stream.
```
jerome-benoit, line 14: [P3] Mention paste-containing batches as an exception

A live channel does not force every keystroke onto the socket: `dispatch_batch` (2191–2194) sets `force_tmux` for the whole batch when any action is `Paste`, and the socket branch excludes those batches at 2252. Please qualify this paragraph so it does not imply that the current paste path violates the single-writer design. This is a documentation clarification, not a runtime regression.

jerome-benoit (approved):
Reviewed `c0eba5e1dd7b40a9559aae9f82311c75b29cf3b4` against `79ab4e40`, with two independent sequential reviews and parent cross-validation. No blocking issue in this documentation-only change.

One additional P3 clarification is suggested inline. The existing batching wording comment is also nonblocking and is not duplicated. The default setting and >=3.8 input gate match `config/mod.rs:2760` and `vt.rs:1561`; DECCKM encoding and disconnected/failed-write fallback match `live_send.rs:2252–2266`.

Evidence: exact-head `git show` and full diff inspection; `git diff --check 79ab4e40...c0eba5e1` clean. No local build/test or independent verification of the historical upstream tmux crash claim. CI was still running at review time; this approval is not a claim that all checks passed.

PR #3884, opened 2026-09-10
fix(session): seed the session-id poller only from a live agent pane

### LINKED ISSUE #3880: Session-id poller can still be seeded with a paired terminal's tmux name
`Instance::maybe_start_poller` seeds the session-id poller from `tmux_env_session_name()` (`src/session/instance/polling.rs:309-317`), which resolves through `live_any_kind_name_for_id` and falls back agent, then terminal, then container. #3872 filtered the derived-name arm with `agent_session_belongs_to`. The primary arm is unfiltered.

Seeded with a paired terminal's name, the poller never terminates. `resolve_session_name` returns `derived` unchanged while that name is live, so every tick re-resolves to the terminal, `probe_pane` reads `Alive`, and `TargetLiveness` never reaches a terminal state. The budget slot is held for the life of the process, and `poll_fn` reads session-id state off the wrong pane.

Repair gates on `has_live_agent_pane_in(snapshot)` (`polling.rs:497`), then `maybe_start_poller` re-queries tmux live. An agent pane that dies between the snapshot and that query, with its paired terminal still up, lands in this state.

Fix: apply `agent_session_belongs_to` to the primary arm as well, and decide what a poller with no resolvable agent name should do. `PollerStart::NotApplicable` leaves repair free to retry on a later tick.

Raised during review of #3872 and deliberately not blocking it: that PR closes the common case, where a tmux session killed outside AoE leaks every slot. This one needs a race to reach.

njbrake:
## Description

Fixes #3880.

The session-id poller was seeded from `tmux_env_session_name()`, which falls back agent, then paired terminal, then container terminal. #3872 filtered only the derived-name arm. Seeded with a terminal's name the poller re-resolves to that live name every tick, `probe_pane` reads `Alive`, and it never terminates: one budget slot held for the life of the process while `poll_fn` reads session-id state off the wrong pane. Reached when the agent pane dies between the snapshot repair gates on and the live re-query.

Both arms now go through one agent-shape filter. With no agent name resolvable the start returns `NotApplicable`: no slot taken, no "budget exhausted" warning, and repair retries on the next tick that sees an agent pane.

Fails closed for a title sanitizing to `term_`, `cterm_`, or `tool_`, whose own agent name is indistinguishable from a paired terminal's: such a row got a working poller from `finalize_launch` before this change and now gets none. Accepting the name back would reopen this bug for any session smart-renamed across that boundary, so the narrowing is deliberate; tracked in #3888.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the race is between an in-process snapshot and a live tmux query microseconds later, which a full-binary test cannot stage deterministically. Two unit tests in `src/session/instance/polling.rs` cover it: a table over seed resolution (agent, paired terminal, container terminal, nothing live, aux-shaped title), and a repair pass whose snapshot shows a live agent pane while the re-query resolves none, asserting `NotApplicable`, no stored poller, no budget slot, no backoff. The second fails on the pre-fix code.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Diagnosis and fix drafted through back-and-forth with @njbrake; the reasoning and decisions are his.

https://claude.ai/code/session_01HiJcnooM1swcM35Q36XTuV

## Summary

- Updated poller startup to accept only live agent session names.
- Return `NotApplicable` when no valid agent name exists.
- Prevent budget usage and allow repair to retry later.
- Added tests for name filtering and the race where an agent pane disappears during startup.

This prevents pollers from attaching to paired or container terminal panes and reading session state from the wrong pane.

### DIFF
--- a/src/session/instance/polling.rs
+++ b/src/session/instance/polling.rs
@@ -117,10 +117,11 @@ fn try_acquire_managed_capture_lease(
 /// is only reached when the agent pane is absent or dead, and the derived name
 /// would then name a session that is not running.
 ///
-/// This fails closed for a title sanitizing under the agent shape's excluded
-/// prefixes, whose own agent name is unrecognizable as one; see
-/// `tmux::live_agent_name_for_id_in`, which already gates poller repair the
-/// same way.
+/// Fails closed for a title sanitizing under the agent shape's excluded
+/// prefixes: its own agent name reads as a paired terminal's, so such a
+/// session runs no session-id poller at all. See
+/// `tmux::live_agent_name_for_id_in` for why that ambiguity has no cheaper
+/// answer.
 fn poller_seed_name(
     live_any_kind: Option<String>,
     derived: impl FnOnce() -> Option<String>,
@@ -335,6 +336,12 @@ impl Instance {
         };
         self.session_id_poller_retry_after = None;
 
+        // Unlike the eligibility checks above, this forks `tmux list-sessions`,
+        // so it stays behind the budget gate rather than joining them: an
+        // over-budget process would otherwise pay a fork per deferred repair.
+        // A session that is both over budget and without an agent pane is
+        // reported as over budget, and the next repair tick stops looking once
+        // its own snapshot agrees the agent pane is gone.
         let Some(tmux_session_name) = poller_seed_name(
             self.tmux_env_session_name(),
             || self.tmux_session().ok().map(|s| s.name().to_string()),
--- a/src/session/instance/polling.rs
+++ b/src/session/instance/polling.rs
@@ -102,6 +102,35 @@ fn try_acquire_managed_capture_lease(
     Ok(ManagedCaptureLease(lease))
 }
 
+/// The tmux session name to seed a session-id poller with, or `None` when
+/// this instance has no agent pane for one to follow.
+///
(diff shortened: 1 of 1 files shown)


jerome-benoit (approved):
Reviewed f87b44fcae8530432dea49bef8e4d34db3e0340e with two independent sequential analyses and parent arbitration. No additional blocker.

The shared filter rejects terminal/container seeds before poller construction (polling.rs:125–132,345–364); NotApplicable clears repair backoff (:552–554). The auxiliary-title regression is explicitly acknowledged and tracked in #3888 as a nonblocking P2, not duplicated here.

I do not consider the existing budget-ordering thread blocking: the gate avoids a live tmux fork for deferred sessions, and a later fresh snapshot skips a missing agent. P3: qualify the description’s unconditional “no budget exhausted warning”; an exhausted budget returns before seed resolution (:297–299). The new repair test uses an auxiliary-title surrogate, not an executed live-tmux race.

Evidence: exact-head git show/diff and git diff --check 4b56cbe9 f87b44f (clean). No local tests run; CI still running at review time.

PR #3883, opened 2026-09-10
fix(tui): measure cell widths the way the renderer does

njbrake:
## Description

Fixes #3879: two regressions #3876 put on `main`. Neither is in v1.16.0.

**Control characters violated the renderer's width contract.** `rendered_width` and `prefix_within_width` measured every grapheme with `CellWidth`, where ratatui filters control-bearing clusters first. So `truncate_to_width("a\tb", 2)` panicked in debug (locked ratatui-core 0.1.2 debug-asserts on a single-byte ASCII control) and charged a cell the renderer never paints in release. Plugin row-column text can carry internal tabs and reaches this path through `src/tui/remote_home/render.rs:62`. Both helpers now apply the renderer's control predicate; `prefix_within_width` keeps control clusters inside the slice, so its result is still a borrowed prefix.

**Wide branch names hid the workspace repository count.** `workspace_branch_row_tag` reserved `+N` in characters while the whole-tag cap in `RowTag::rendered` works in cells, so ten `界` with two repositories kept six wide characters and dropped `+2`. The suffix is reserved, and the branch cut, in cells; that case now renders `[界界界界界+2]`.

The profile short-code behavior #3876 fixed is unchanged, halfwidth-katakana coverage included.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

`cargo fmt`, `cargo clippy --all-targets` and `cargo test` are clean. No screenshot: the visible change is two cells of a row tag, and the new Buffer-backed assertions state it more precisely than a capture would.

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: this PR doesn't change a user-facing dashboard flow

**TUI / CLI (e2e test)**
- [ ] N/A
- [ ] Added or updated an e2e test under `tests/e2e/`
- [x] Skipped a test, because: both cases are pure width arithmetic reachable from a unit test. `control_characters_cost_no_cells` asserts against a real `Buffer`, and `workspace_branch_row_tag_keeps_the_repository_count` asserts the rendered tag. A full-binary test would cost minutes to cover the same two cells.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 (1M context) via Claude Code.

[...]

PR #3887, opened 2026-09-10
fix(tui): align remote-home columns by the metric that places spans

njbrake:
## Description

Follow-up to #3883, found while reviewing it. The remote-home session picker pads its title, status and context columns with `{:<N$}` and cuts titles with a local char-based `truncate`. Chars are not cells: an 8-char CJK title paints 16, so the column runs long and pushes the project path right on that row alone. Measured against a `TestBackend`, the path lands at column 62 instead of 54, and at 77 for a 40-char wide title.

Padding by `rendered_width` does not fix it. ratatui lays a `Line` out with two metrics that disagree: `Span::render` advances a cell per `CellWidth`, while `render_spans` starts the next span at `Span::width`, which is `UnicodeWidthStr`. They part only on halfwidth katakana dakuten/handakuten (U+FF9E, U+FF9F), where a `CellWidth` pad starts the next column one cell short per mark.

`fixed_width` therefore cuts with `truncate_to_width`, keeping painted glyphs inside the column, and pads with `UnicodeWidthStr`, landing the next span on the boundary.

The plugin row-column measurement at `remote_home/render.rs:66` is deliberately unchanged. `UnicodeWidthStr` is what decides where the path span starts, so it is correct there.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`cargo fmt`, `cargo clippy --lib --tests` and `cargo test --lib` (6930 passed) are clean. No screenshot: the change is a column offset, and the new `TestBackend` assertions state it more precisely than a capture would.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [ ] N/A
- [x] Skipped a test, because: the three unit tests render the real picker through a `TestBackend` and assert painted column offsets, which is the whole behavior. A full-binary test would cost minutes for the same assertion. Each was confirmed to fail against the previous code.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Found during a review of #3883. The agent first proposed padding by `rendered_width` and reverted it after the halfwidth-katakana test failed, then read `ratatui-core` 0.1.2's `Span::render` and `render_spans` to establish which metric governs span placement. Two adjacent changes it had proposed, to the plugin-column measurement and to `diagnostics::format_agent_title`, were dropped once that showed the existing code was already correct.

https://claude.ai/code/session_01PodUXso6Y16zpy8Py1FmK3

## Summary

- Fixed remote-home session picker column alignment for wide and halfwidth characters.
- Added `fixed_width` to truncate and pad text by rendered terminal width.
- Replaced character-count formatting in session rows with width-aware formatting.
- Added tests for CJK, halfwidth Katakana, truncation, padding, and column offsets.

This keeps project paths aligned and prevents titles or statuses from shifting later columns.

### DIFF
--- a/src/tui/components/mod.rs
+++ b/src/tui/components/mod.rs
@@ -19,7 +19,7 @@ pub use dir_picker::{DirPicker, DirPickerResult};
 pub use help::HelpOverlay;
 pub use list_picker::{ListPicker, ListPickerResult};
 pub use preview::{format_scroll_indicator, Preview};
-pub use text::{prefix_within_width, rendered_width, truncate_to_width};
+pub use text::{fixed_width, prefix_within_width, rendered_width, truncate_to_width};
 pub(crate) use text_input::{focused_input_spans, input_scroll, visible_slice};
 pub use text_input::{
     longest_common_prefix, render_text_field, render_text_field_with_ghost,
--- a/src/tui/components/text.rs
+++ b/src/tui/components/text.rs
@@ -161,9 +161,32 @@ pub fn prefix_within_width(text: &str, max_width: usize) -> &str {
     &text[..end]
 }
 
+/// `text` fitted to a `max_width` column of a multi-span [`ratatui::text::Line`].
+///
+/// Cut and padded by different metrics, because ratatui lays a `Line` out with
+/// two that disagree: `Span::render` advances a cell per `CellWidth`, while
+/// `render_spans` starts the next span at `Span::width`, which is
+/// `UnicodeWidthStr`. They part only on halfwidth katakana
+/// dakuten/handakuten (U+FF9E, U+FF9F), where `CellWidth` charges the cell the
+/// terminal spends and `UnicodeWidthStr` scores zero.
+///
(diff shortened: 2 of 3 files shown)


jerome-benoit (approved):
Reviewed 93166fae003b153f05011fe82d283008996976d9 with two independent sequential analyses and parent cross-validation. No actionable finding or substantive disagreement.

The mixed metrics are justified by locked ratatui-core 0.1.2: Span renders with CellWidth, but Line advances subsequent spans using Span::width (UnicodeWidthStr). fixed_width (text.rs:178–182) therefore caps painted glyphs and aligns the next column correctly. The unchanged plugin-column measurement is appropriate. TestBackend cases (remote_home/render.rs:462–504) verify actual path columns for CJK, halfwidth katakana, wide status and truncation.

Evidence: exact-head git show/diff, locked dependency source (line.rs:759–779; span.rs:271–273,424–470), and git diff --check 3ca14eb8 93166fae (clean). No local build/test run. CI was still completing at review time.

PR #3889, opened 2026-09-10
test(acp): pin claim-before-wake on both prompt handlers

### LINKED ISSUE #3859: acp: a Stop during the prompt handlers' wake/resume window is not queued behind the prompt
## Summary

Post-#3844, both turn-starting prompt handlers still run `touch_on_prompt_and_wake_if_sunk` **before** claiming the session's prompt-submission guard, so a Stop landing inside the wake/resume window is forwarded immediately instead of queueing behind the prompt it names.

## Evidence

- `src/server/api/acp.rs:1354` (`acp_prompt`): wake at :1354, claim at :1380 (`begin_prompt_submission`), attachments validation and resume triggering in between.
- `src/server/api/acp.rs:1532` (`acp_prompt_diff_comments`): wake at :1532, claim at :1538.
- The cancel handler (`src/server/api/acp.rs:2273`) claims the same guard (`prompt_submission_for_session`) with **no timeout** (`src/server/session_service.rs:1410-1429`, plain `lock_owned().await`), so once the guard is held, cancel-vs-prompt ordering is preserved by queueing (the #3844 mechanism). The gap is exactly the window before the claim.

## Race shape

A Stop pressed while the handler is waking a sunk/dormant session (or validating attachments, or the background resume trigger is in flight) finds no open submission guard, is forwarded straight to the command channel, and can reach the agent before the prompt it names. Same observable failure as #3844's composer-stop race, restricted to this producer's wake path (milliseconds when awake; longer when a sunk session needs a resume).

## Suggested direction

Mirror the other holders: claim the submission guard before `touch_on_prompt_and_wake_if_sunk` in both handlers (the wake's `woke_idle_dormant` flag is passed into `begin_prompt_submission`, so the claim may need to move to a claim-then-wake ordering with the flag read from the wake's return). #3539's review thread documents the same gap for the diff-comments handler; its other findings are moot post-#3844.

_Filed from the #3539 review (CodeRabbit finding + source tracing on main `02398ddb`); the PR itself is superseded by #3844._

njbrake:
## Description

Closes #3859.

The reorder #3859 asks for is already in the tree. `acp_prompt`, `acp_prompt_diff_comments` and the plugin turn path all claim the session's prompt-submission guard before `touch_and_wake_on_prompt`, so a Stop landing in the wake/resume window queues behind the prompt it names. #3639 moved the first, #3673 the second, #3844 finished the set. The issue traced an older tree: `begin_prompt_submission` and `touch_on_prompt_and_wake_if_sunk` are both pre-#3758 names, and its line numbers match neither main at `02398ddb` nor #3844's parent.

What was missing is the pin. The ordering was asserted for `acp_prompt` alone, as a second assertion inside the cancel test, behind a 500ms sleep. A diff-comments handler that drifted back to waking first failed nothing.

`prompt_handlers_claim_the_submission_guard_before_they_wake` covers both endpoints with no sleep. The submission-claim tap fires after the guard's pre-acquisition existence check and before `prompt_locks` is read, so a held guard parks each handler there with nothing else run yet, and `last_accessed_at` reads the ordering at that checkpoint. Letting the handler finish afterwards keeps it honest: a handler that never woke would pass the same assertion.

The cancel test keeps the other half, that a Stop waits an in-flight submission out, and loses the now-duplicated assertion and the prompt task it was there for.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

Test-only: the behavior #3859 reports is already correct, so this lands the regression coverage rather than a second fix.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Mutation-checked: moving the wake above the claim fails the new test for each handler independently.

| Check | Result |
| --- | --- |
| `cargo test` | 6926 lib + 289 integration passed, 0 failed |
| `cargo clippy --all-targets` | clean |
| `cargo fmt` | clean |

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 5 (1M context) via Claude Code

Issue triage, the source tracing that found the reorder already landed, and the test were done in one session. The mutation check was run for each handler separately.

https://claude.ai/code/session_012q6ATxCt3Dfv2k5ux3Ykvt

## Summary

- Added regression coverage for `acp_prompt` and `acp_prompt_diff_comments`.
- Verified that both handlers claim the submission guard before waking the session.
- Simplified the cancel test so it only checks that Stop waits for an active prompt submission.
- Replaced timing-based checks with deterministic guard-claim monitoring.

This prevents future changes from reintroducing the prompt/Stop ordering race.

### DIFF
--- a/src/server/api/acp.rs
+++ b/src/server/api/acp.rs
@@ -3537,27 +3537,23 @@ mod tests {
     /// stretched it from ~1ms to hundreds and the live `composer-stop` spec
     /// failed on whichever case clicked Stop without waiting for output first.
     ///
-    /// Both halves of the ordering are pinned here, against a held guard
-    /// rather than a live agent: cancel must wait for it, and the prompt
-    /// handler must claim it before its first side effect (the wake, which is
-    /// what `last_accessed_at` records). A handler that claims later leaves
-    /// exactly that much window for a cancel to pass through.
+    /// This pins the cancel half of the ordering, against a held guard rather
+    /// than a live agent: a Stop must wait the submission out instead of
+    /// racing ahead of it. The producer half, that a prompt handler claims the
+    /// guard before its first side effect and so leaves no window to race at
+    /// all, is `prompt_handlers_claim_the_submission_guard_before_they_wake`.
     #[tokio::test]
-    async fn cancel_and_prompt_serialize_on_the_submission_guard() {
+    async fn cancel_waits_for_an_in_flight_prompt_submission() {
         use std::time::Duration;
 
         let mut inst = crate::session::Instance::new("stop-order", "/tmp/aoe-stop-order");
         inst.id = "sess-stop-order".to_string();
         inst.view = crate::session::View::Structured;
-        assert!(
-            inst.last_accessed_at.is_none(),
-            "fresh instance is untouched"
-        );
         let id = inst.id.clone();
         let state = crate::server::test_support::build_test_app_state(vec![inst]);
 
         // Stand in for a submission already in flight. There is no worker, so
-        // every path below fails fast once it gets past the guard.
+        // cancel fails fast once it gets past the guard.
         let submission = state
             .session_service
             .prompt_submission_for_session(&id)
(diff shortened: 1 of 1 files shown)


jerome-benoit (approved):
Reviewed fa0a6ed6e666b35daa72441168d03ba79f889a1e with two independent sequential analyses and parent cross-validation. No actionable finding or substantive disagreement.

Both handlers already claim before waking (acp.rs:1258–1265,1438–1445). This PR correctly adds regression coverage, rather than claiming a new runtime fix for #3859. The service-local tap fires before acquisition (session_service.rs:1461–1480); holding that exact mutex makes the pre-release None assertion causal for both endpoints. The post-release Some assertion rejects a handler that never wakes (acp.rs:3671–3690). The separate cancel test retains its blocking/completion checks; its existing 500 ms observation is not newly presented as causal.

Evidence: exact-head git show/diff and git diff --check 3ca14eb8 fa0a6ed6 (clean). CI is green at this SHA. No local test or mutation run was performed; the author’s mutation results are not claimed as my own.

PR #3882, opened 2026-09-10
fix(acp): restore the narrow stale-session fingerprints

jerome-benoit:
## Description

**Update 2026-09-10 17:0x**: #3561 itself was closed by the maintainer (no disposition comment), unmerged. Its three own commits (skip unknown WS control-frame kinds, route WS frames on kind presence, recover a resumed session the agent rejects) are not on main, so this PR carries the complete set: everything #3561 shipped plus the blocking P2 fix. The Description below was written while #3561 was still open; the delta this PR delivers is the union.

Closes #3561 (superseded-unmerged; this PR delivers its content). The blocking P2 from #3561's 2026-09-09 review is fixed here (head `athal7:fix/acp-frame-parser` was not pushable from here). Built on that PR's head (`2cb1971f`); merging this one supersedes the closed #3873, which carried the same commit (`aae64c03`).
Closes the blocking P2 from #3561's 2026-09-09 review (head `athal7:fix/acp-frame-parser` is not pushable from here). Built on that PR's head (`2cb1971f`); merging this one supersedes the closed #3873, which carried the same commit (`aae64c03`).

The move to `rate_limit.rs` broadened `is_unsupported_session_error` to independent words, so ordinary prompt failures carrying a phrase word next to `session` (`Unsupported content block in session/prompt`, `Method not found: session/prompt`) matched and triggered the `SessionContextReset` path, clearing the cached agent session id. This PR restores whole-phrase matching (the narrower fingerprints the review's earlier lineage established) and the two negative regression cases.

## PR Type

- [x] Bug Fix

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary (doc comment states the whole-phrase rationale)
- [x] For UI changes: included screenshot or recording (no UI change)

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: Rust unit test only

**TUI / CLI (e2e test)**
- [x] N/A: covered by the classifier table test

## AI Usage

- [x] AI was used

**AI Model/Tool used:** omp (OpenCode) with Omen Alpha

**Any Additional AI Details you'd like to share:** Authored in an AI session under Jérôme Benoit's review workflow.

- [x] I am an AI Agent filling out this form (check box if true)

## Summary

- Restored narrow matching for stale ACP session errors.
- Prevented ordinary prompt errors from clearing valid cached session IDs.
- Extended recovery to unsupported sessions rejected during later prompts.
- Ensured `SessionContextReset` occurs before the terminal error.
[...]

Issue #2603, opened 2026-07-01
refactor(sidebar): share per-request Config resolution across list_sessions overlays (follow-up to #2602)


jerome-benoit:
**Note:** Claude (claude-opus-4-7 via OhMyOpenCode) handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @jerome-benoit

Follow-up to #2602 (repo-aware smart-rename resolution). Surfaced by CodeRabbit on the PR review and pre-disclosed in the PR body's "Known limitation and follow-up" section; tracked here per contribution rules to keep the #2351 fix tight.

## Problem

`list_sessions` in `src/server/api/sessions.rs` runs two per-request overlays that both call `resolve_config_with_repo_or_warn` (`src/session/repo_config.rs`) for the same `(profile, project_path)` key:

- `acp_cmd_cache` block (around L548-567, `#[cfg(feature = "serve")]`): reads `.session.agent_acp_cmd` for the ACP-capability overlay.
- `cfg_cache` block introduced by #2602 (around L622-673, unconditional): reads `.session.{smart_rename, smart_rename_agent, agent_command_override}` for the sidebar indicator overlay.

`resolve_config_with_repo_or_warn` reads `<repo>/.agent-of-empires/config.toml` from disk, parses it, and merges the sparse overrides onto the profile config. Both overlays cache their result per-request, but the underlying resolver is called twice per unique `(profile, project_path)` pair.

The sidebar polls every 3s (`web/src/hooks/useSessions.ts:6`, `POLL_INTERVAL = 3000`), so the duplication doubles the filesystem work on this hot path.

Not a regression: the duplication predates #2602 (only the ACP side existed before). #2602 introduced the second call site to fix a correctness bug and deliberately deferred this consolidation to keep scope tight.

## Proposal

Fold both overlays behind one per-request cache of the resolved session sub-struct, keyed by `(profile, project_path)`. Derive `agent_acp_cmd`, `smart_rename`, `smart_rename_agent`, and `agent_command_override` from that single lookup. Rough shape:

```rust
let mut session_cfg_cache: HashMap<(String, String), SessionSubConfig> = HashMap::new();
let mut resolve_session_cfg = |profile: &str, project_path: &str| {
    session_cfg_cache
        .entry((profile.to_string(), project_path.to_string()))
        .or_insert_with(|| {
            crate::session::repo_config::resolve_config_with_repo_or_warn(
                profile,
[...]

PR #3892, opened 2026-09-10
fix(server): scope the list_sessions resolver counter to AppState

njbrake:
## Description

`list_sessions_shares_config_resolution_across_overlays` asserted against a process-global `LIST_SESSIONS_RESOLVER_MISSES` that every `list_sessions` call in the process bumped. Any concurrent caller landing between the counter's reset and its load leaked misses into the assertion. The static's own doc comment named the hazard and pushed the burden onto readers via `#[serial]`, which cannot cover unmarked tests. It failed `Cargo Test (linux, bare-core)` during #3882 with 3 misses instead of 2, and passed on re-run.

The counter now lives in the per-request cache, promoted to a `SessionCfgCache` owning its entries and its miss count, and `list_sessions` publishes that count on `AppState` once both overlays have run. Each test builds its own `AppState`, so the assertion cannot observe another request's resolutions. The count also reaches a `debug!` line, so the sharing invariant (#2603) is visible in the request log instead of only under `cfg(test)`.

The integration-level assertion is kept deliberately: a unit test on the cache alone would still pass if an overlay were given its own cache, which is the regression #2603 guards against.

Verified by hammering `list_sessions` on a second `AppState` from a concurrent task, 200 iterations: the old global read 98 where 2 was expected; the per-state counter read 2 every time. That scratch test is not committed.

I did not identify which concurrent test triggered the CI failure. Both `list_sessions` callers in the suite are `#[serial]`, and the router tests that reach `/api/sessions` are denied before the handler with empty state. The fix removes the class rather than the one trigger.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: included screenshot or recording (no UI change)

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The existing test is the coverage; this change makes it deterministic. No behavior change to the endpoint.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Diagnosed and written in a Claude Code session with @njbrake, including the concurrency experiment that pinned the mechanism.

https://claude.ai/code/session_01JAR986qooEi4ev6gg1c7aw

## Summary

- Replaced process-global miss tracking with request-scoped tracking in `SessionCfgCache`.
- Shared the miss counter across both session-config overlays.
- Added per-request miss reporting and debug logging through `AppState`.
- Updated tests and state initialization for the new counter.

## Benefits

- Prevents concurrent requests from affecting cache-invariant tests.
- Preserves detection of split caches across overlays.
- Keeps endpoint behavior unchanged.

### DIFF
--- a/src/server/api/sessions/list.rs
+++ b/src/server/api/sessions/list.rs
@@ -118,7 +118,12 @@ pub async fn list_sessions(
     // Share resolved config between the ACP-capability and smart-rename
     // overlays, halving disk reads when a profile/project pair repeats in the
     // 3s sidebar poll. See #2603.
-    let mut session_cfg_cache = SessionCfgCache::default();
+    // Monotonic, so the delta below is this request's own count and no reset
+    // can race a concurrent request on the same state.
+    let misses_before = state
+        .list_sessions_resolver_misses
+        .load(std::sync::atomic::Ordering::Relaxed);
+    let mut session_cfg_cache = SessionCfgCache::new(&state.list_sessions_resolver_misses);
 
     // Overlay custom-agent ACP capability (built-ins were resolved in the
     // constructor). Distinct `(profile, project_path)` pairs each resolve
@@ -254,10 +259,10 @@ pub async fn list_sessions(
     }
 
     // Both overlays have run, so the count is final for this request.
-    let resolver_misses = session_cfg_cache.misses();
-    state
+    let resolver_misses = state
         .list_sessions_resolver_misses
-        .store(resolver_misses, std::sync::atomic::Ordering::Relaxed);
+        .load(std::sync::atomic::Ordering::Relaxed)
+        .saturating_sub(misses_before);
     tracing::debug!(
         target: "http.api.sessions",
         rows = sessions.len(),
--- a/src/server/api/sessions/list.rs
+++ b/src/server/api/sessions/list.rs
@@ -118,7 +118,7 @@ pub async fn list_sessions(
     // Share resolved config between the ACP-capability and smart-rename
     // overlays, halving disk reads when a profile/project pair repeats in the
     // 3s sidebar poll. See #2603.
-    let mut session_cfg_cache: HashMap<(String, String), SessionConfig> = HashMap::new();
+    let mut session_cfg_cache = SessionCfgCache::default();
 
     // Overlay custom-agent ACP capability (built-ins were resolved in the
     // constructor). Distinct `(profile, project_path)` pairs each resolve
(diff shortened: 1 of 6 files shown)


src/server/api/sessions/list.rs
```diff
@@ -261,6 +253,18 @@ pub async fn list_sessions(
         }
     }
 
+    // Both overlays have run, so the count is final for this request.
+    let resolver_misses = session_cfg_cache.misses();
```
jerome-benoit: [P2] Count resolutions independently of the cache being tested

This reads only one cache’s misses. If smart rename gains a separate SessionCfgCache inside its block and uses it at :237, the first overlay still records 2, the second performs 2 more resolutions, and this publication still reports 2. The assertion at tests.rs:650–655 passes despite breaking the cross-overlay sharing invariant; the previous instrumentation would report 4. Keep the isolation, but observe underlying resolver calls through request-local instrumentation shared independently of the cache instances, and verify that splitting the caches makes the integration test fail. This is a source-derived mutation, not a claimed local failing run.

jerome-benoit (changes requested):
The cross-test isolation fix is sound, but please preserve the cross-overlay sharing regression oracle before merging: one P2 test-contract finding below, not a current production cache failure.

Two independent sequential reviews and parent arbitration: the second reviewer initially accepted the assertion, then confirmed the split-cache false-positive after checking the concrete mutation. No duplicate existing finding.

Evidence: exact-head git show and three-dot diff against d208effb; git diff --check clean, CI green. The mutation is source-derived, not executed locally. No local build/test run.

jerome-benoit (approved):
The previous P2 is fixed: caches now require an external counter (`model.rs:316–340`), so splitting the two overlay caches would report four resolutions instead of the two asserted at `tests.rs:650–655`. Each test owns a separate AppState; the original cross-test interference remains removed.

The existing [counter-attribution comment](https://github.com/agent-of-empires/agent-of-empires/pull/3892#discussion_r3982776183) is valid P3, non-blocking: the AppState-wide delta at `list.rs:123–126,261–270` can include another request's misses, but only affects debug reporting, not the API response or this isolated test. Use a request-local counter shared across caches and aggregate separately if needed; align the comments and PR description with the actual ownership. No duplicate inline comment.

Two independent sequential reviews agree; cross-validation found no remaining blocker. Evidence: exact-object git show/diff and `git diff --check d208effb4571802edcd6e1925fcf3cba80446b48...993cd60406e97c77e40a01cb760c9edecaf4c1e8` (clean); CI, Tests and CodeQL succeeded at this SHA. The split-cache case is source-derived, not an executed local mutation. No local tests/builds run.

PR #3894, opened 2026-09-10
fix(tmux): record each session's kind on the session itself

### LINKED ISSUE #3888: Sessions whose title sanitizes to term_/cterm_/tool_ get no session-id poller
A title sanitizing to `term_`, `cterm_`, or `tool_` gives an agent tmux name that `NameShape::agent` refuses (`AGENT_EXCLUDED_PREFIXES`, `src/tmux/mod.rs:695`), so nothing recognizes that session's own agent pane by name. #3872 declines to repair its session-id poller; #3884 declines to start one. No poller runs for the row, so its conversation id is not observed while the pane is live and resume can silently fall back to a stale or absent id. The only signal is a debug line on `session.create`.

These rows did get a working poller before #3884, from `finalize_launch` via the live-name arm that #3884 narrowed to close #3880.

Accepting the name back is not the fix. `aoe_term_Foo_<id8>` is both the agent name for title `term_Foo` and the paired-terminal name for title `Foo`, and nothing durable on the pane records which kind it is, so a smart rename across that boundary makes a surviving terminal indistinguishable from the agent. A per-pane kind marker written at creation and read by the resolver would settle this and #3880's ambiguity together.

Repro: create a session titled `term rewriting`; no session-id poller starts for it.

_Note: this issue was drafted by Claude Opus 5 (1M context) via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

njbrake:
## Description

Fixes #3888.

A tmux session's kind was inferred from its name, which cannot always say: `aoe_term_Foo_<id8>` is both the agent name for title `term Foo` and the paired-terminal name for title `Foo`. #3872 and #3884 resolved that ambiguity by refusing the name, which cost those rows their session-id poller. No poller runs, the conversation id is not observed while the pane is live, and resume can fall back to a stale or absent id.

Sessions now carry `@aoe_kind` (`agent`, `term`, `cterm`, `tool`), chained onto the `new-session` that creates them and read back by the shared `list-sessions` scan. `SessionKind::of` is the one classifier: the marker when the session has one, the old name shape when it does not, so a session started by an earlier aoe keeps the behavior it has today rather than becoming unclassifiable.

That closes both directions. An agent titled `term rewriting` is found as its row's agent pane again, and a paired terminal holding what a rename has since made the agent's derived name is no longer adopted as the agent, which is the collision #3880 was about.

A marker is trusted only at session scope. `#{@aoe_kind}` falls through to the global option, so the scan reads the global in the same invocation and subtracts it: a value a user's own `set -g @aoe_kind` could have produced is treated as unmarked, which is the name-shape fallback rather than a wrong answer.

Bounds worth knowing. A session already running when you upgrade carries no marker until it is restarted, and keeps the name-shape answer. `resolve_agent_session_name_in`, which resolves from `list-panes` keys for the status poller and `aoe ps`, has no markers to read and is unchanged. `@aoe_kind` is documented alongside the other `@aoe_*` options users reference in their own status bars.

Carries one unrelated one-liner: `list_sessions_projects_pending_approvals_only_for_running_workers` is the only non-serial `list_sessions` caller, and its resolver-miss bumps leak into the counter a serial sibling resets and reads. It failed CI on this branch; the fix is its own commit.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the facts an e2e would add over the unit tests are already covered against real tmux and the real start path. `tmux::session::tests::every_kind_is_marked_at_creation_and_keeps_its_mark_across_a_rename` creates an agent, a paired terminal and a tool sub-session through their own production paths and reads all three back through the scan command and parser the session cache uses, including across a rename, so a typo in the `-F` string or a missing chain on one kind fails there; `polling::tests::an_aux_shaped_title_polls_its_marked_agent_pane` drives `maybe_start_poller` for a `term rewriting` row against a scan that answers with a marked agent. The rest is classification, which is pure. Reusing the Claude e2e fixture would mean duplicating its two-session correlation setup for one session.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Diagnosis and fix drafted through back-and-forth with @njbrake; the reasoning and decisions are his.

https://claude.ai/code/session_01HiJcnooM1swcM35Q36XTuV

## Summary

- Added a durable `@aoe_kind` marker for agent, terminal, container terminal, and tool sessions.
- Updated session discovery to use the marker and retain name-based fallback for older sessions.
- Restored session-ID polling for agents with auxiliary-looking names.
- Prevented renamed paired terminals from being classified as agents.
- Updated session polling, tmux integration, documentation, and tests.
- Serialized a `list_sessions` caller to prevent resolver counter interference.

## Benefits

- Session classification remains correct after renames.
- Agent conversation IDs are reliably observed.
- Existing sessions remain compatible through fallback classification.
- User-defined global tmux options cannot incorrectly mark all sessions.

### DIFF
--- a/docs/guides/tmux-status-bar.md
+++ b/docs/guides/tmux-status-bar.md
@@ -131,10 +131,15 @@ cannot: a title such as `term notes` gives an agent session a name shaped like
 a paired terminal's. A session started by an earlier aoe carries no value until
 it is restarted.
 
-Read it, but do not set it. aoe sets it per session, and ignores any value it
-could have read from a global `set -g @aoe_kind` of yours, since a global one
-applies to every session on the server and would tell aoe that panes are
-something they are not.
+Read it, but do not set it. A value of yours at server (`set -s`), window
+(`set -w`) or global-window (`set -gw`) scope does not sit behind the one aoe
+writes, it replaces it for every session on the server, and a global-session
+one (`set -g`) is read by every session that has no value of its own. aoe
+discards any value the server-wide scopes could have produced rather than
+believe a pane is something it is not, so setting one costs you the marker
+entirely: those sessions fall back to being classified by name, which is the
+ambiguity `@aoe_kind` exists to remove. Window and pane scope vary per pane
+and cannot be discarded this way at all.
 
 ```tmux
 # Only decorate the agent pane
--- a/docs/guides/tmux-status-bar.md
+++ b/docs/guides/tmux-status-bar.md
@@ -128,8 +128,13 @@ set -g status-right "#{@aoe_title} #{@aoe_branch} #{@aoe_sandbox} | %H:%M"
 `cterm` (container terminal), or `tool`. It is written when the session is
 created and survives renames, so it stays accurate where the session name
 cannot: a title such as `term notes` gives an agent session a name shaped like
-a paired terminal's. A session started by an earlier aoe carries no
-value until it is restarted.
+a paired terminal's. A session started by an earlier aoe carries no value until
+it is restarted.
+
+Read it, but do not set it. aoe sets it per session, and ignores any value it
+could have read from a global `set -g @aoe_kind` of yours, since a global one
+applies to every session on the server and would tell aoe that panes are
+something they are not.
 
 ```tmux
 # Only decorate the agent pane
--- a/docs/guides/tmux-status-bar.md
+++ b/docs/guides/tmux-status-bar.md
@@ -118,10 +118,22 @@ set -g status-right "#{?#{==:#(aoe tmux status),},,%#(aoe tmux status) | }%H:%M"
 
 ## tmux User Options
 
-aoe sets `@aoe_title`, `@aoe_branch` (worktree sessions), and `@aoe_sandbox` (sandboxed sessions) on each session, which you can reference in your own config:
+aoe sets `@aoe_title`, `@aoe_branch` (worktree sessions), `@aoe_sandbox` (sandboxed sessions), and `@aoe_kind` on each session, which you can reference in your own config:
 
 ```tmux
 set -g status-right "#{@aoe_title} #{@aoe_branch} #{@aoe_sandbox} | %H:%M"
+```
+
+`@aoe_kind` is what kind of session it is: `agent`, `term` (paired terminal),
+`cterm` (container terminal), or `tool`. It is written when the session is
+created and survives renames, so it stays accurate where the session name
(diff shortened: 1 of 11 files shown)


jerome-benoit (approved):
The marker-aware lookup and poller seed address #3888. No blocking regression established for supported configurations. Existing threads remain valid, with this scope classification rather than duplicate inline comments:

- **P2, pre-existing, non-blocking here:** [wrong-kind collision](https://github.com/agent-of-empires/agent-of-empires/pull/3894#discussion_r3983444368). Base `1284397a` already returns the derived name without an adopted candidate and uses name-only existence. The residual creation/lifecycle problem is real (`mod.rs:1156–1158,723–724`; `session.rs:453–455`), but is not introduced here. Narrow the description's claim that both directions are fully closed.
- **P2, conditional robustness, non-blocking here:** [scope-record ambiguity](https://github.com/agent-of-empires/agent-of-empires/pull/3894#discussion_r3983542838). `note|0` ends scope collection early (`mod.rs:601–606,627–651`), potentially accepting a later inherited `agent` value. This is introduced, but requires manually configured `@aoe_kind` values, explicitly unsupported by `docs/guides/tmux-status-bar.md:134–142`. Explicit record framing would be the proper hardening.
- **P3:** remove or restrict `mod.rs:1021–1028` to unmarked sessions: it still references the deleted `AGENT_EXCLUDED_PREFIXES` and says no durable marker exists. Also remove the description's obsolete serialization paragraph; that change was dropped.

Checked: creation wiring for agent/paired/container/tool paths, marker precedence, old-session fallback, and agent/other-kind/absent seed separation. No separate missing-e2e finding: production-path tmux tests and the poller-start test cover the relevant seams.

Two independent sequential analyses plus arbitration. An initial blocking verdict was rejected after comparing the collision with the base and checking the parser scenario's unsupported precondition. Evidence: exact-object `git show`, full `git diff 1284397a88e495bccddf985972a795bad6079d45...eec079fdf8b1d8bb57e66704d3f3e75326308961`, and `git diff --check` (success, no output). No local test/build/runtime reproduction.

Code approval only: CI workflow 34530230942 passed; Tests 34530230867 remains incomplete with a Playwright timeout at `acp-compaction-phase.spec.ts:120`. No causal link to this diff established, and this approval does not assert that CI is green.

PR #3893, opened 2026-09-10
fix(web): keep IME edits on the active terminal surface

jerome-benoit:
## Description

Closes #3885. Keep retained IME text consistent with accepted terminal edits, including queued replay, refused composition commits, asynchronous image paste and mobile agent/paired surface switches.

- Preserve independently accepted queued edits and reconstruct their accepted shadow suffix after refusal. The real-component regression replaces the one displaced by the earlier force-push; 9b2c7b6a remains preserved.
- Skip upload completion after unmount. A mounted background terminal can finish its initiated paste without clearing the foreground proxy.
- Propagate the visible mobile surface's active state through TerminalSessionStack and PairedShellPane. Keep terminal views mounted, but register proxy listeners only on the visible surface. Clear buffered and retained proxy text on a real session/surface boundary, not when reselecting the same target.
- Clear the event textarea and shared proxy when compositionend cannot deliver its commit. Accepted text still supplies the next IME rewrite's context.

Main is merged without rewriting history. The maintainer's synchronous paste correction from #3895 and both branches' regression tests are retained. The hidden agent/paired ownership defect previously described as separate is now addressed here, not deferred.

## PR Type

- [x] Bug Fix

## Checklist

- [x] New and existing tests pass locally as detailed below
- [x] Documentation updated where necessary: component ownership comments and coverage-matrix mapping; no new controls or settings
- [x] UI verification: actual Chromium mobile surface inspected; no styling/layout change requiring a comparison screenshot

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] Covered by live-terminal-ime-rewrite.spec.ts, mobile-right-panel-picker.spec.ts, mobile-keyboard.spec.ts and terminal-focus-shortcut.spec.ts; coverage-matrix validator passes.
- All three added regressions failed before correction: refused composition retained c; a surface switch retained the previous syllable; the hidden paired terminal received the agent's keystroke during its pending upload.
- 3824 unit tests in 362 files passed. The four browser specs passed 39 cases; 2 existing native-proxy cases are skipped by the suite. After refining assertions to compare complete byte streams per destination, all 11 IME cases passed again at 09c50a0c.
- Format, ESLint, TypeScript, production build, matrix validation and commit-hook Cargo formatter/Clippy passed.
[...]

Issue #3885, opened 2026-09-10
Mobile terminal: the retained IME shadow still drifts from the PTY line after ba4e5bcd (Ctrl-with-text, async image upload, refused drain)


jerome-benoit:
## Context

Post-merge review of #3692's corrections in `f1c27658` (the `ba4e5bcd` merge of #3692) identified three remaining cases where the hidden textarea retains text that no longer matches the PTY line, so a later Korean rewrite sends `\x7f` against text the user never typed or loses text the user did type. Same contract as #3877, which was closed as fixed by `ba4e5bcd`; these are the parts it did not cover. Split out as its own issue rather than reopening the maintainer's closed one.

## 1. Ctrl modifier with existing text (P2)

Type `한`, toggle the toolbar's **Ctrl**, then type `c`. `MobileLiveTerminal.tsx:1710-1715` (main: the `sendKeys` chord branch) sends `\x03` and returns `false` without invalidating the existing shadow; `mobileKeyboardProxy.ts:80` prevents the new `c`, but the previous `한` stays retained. Invalidate when an input is **transformed into a control code**, not merely when its default edit is cancelled. The regression added in #3692 starts from an empty textarea and misses this case.

## 2. Image-upload completion (P2)

Paste an image through the local terminal textarea, type a syllable while upload is pending, then let the upload complete. `MobileLiveTerminal.tsx:1865` calls `invalidateRetainedImeContext()` with no target; the helper (`mobileKeyboardProxy.ts:49-52`) clears only the proxy in that case, not the local input that acquired new text. Pass the current local input reference when the asynchronous send completes, and test typing during a deferred upload.

## 3. Refused queued drain (P2)

`mobileKeyboardProxy.ts:17-21` accepts queued edits into the shadow while unregistered, but the drain at :29-33 ignores a later `false` return from the receiver. Cover delayed rejection too: invalidate the proxy when a queued edit is refused.

## Related

- #3877 (closed as fixed by ba4e5bcd) - the parts it covered are fixed and verified.
- #3812 - a different neighbouring case (Android IME replaces instead of repeating).

_Source validation: the three transitions were read on main at `4b56cbe`'s successor `3492c9e`-lineage (`sendKeys` chord branch, completion invalidation call, proxy drain); no local browser/device reproduction was run._

_Note: drafted by Claude Opus 5 in the review-fix session under @jerome-benoit's direction; the analysis is from the session's post-merge review._

jerome-benoit:
[...]

PR #3895, opened 2026-09-10
fix(web): clear the local IME shadow when a paste bypasses the textarea

njbrake:
## Description

`forwardTerminalBeforeInput` swallows `insertFromPaste` so the paste never enters the textarea, but the receiver at `MobileLiveTerminal.tsx:1753` invalidated with no target, and `invalidateRetainedImeContext` clears only App's proxy in that case. A syllable typed before the paste stayed in the live terminal's own textarea while the line showed the pasted text, so the next Korean rewrite opened with a delete against text the user did not type.

Found while reviewing #3893. That PR applies the identical correction to the async image-upload path at line 1865; this is its synchronous sibling, reached whenever a paste arrives through the terminal's textarea rather than an upload. Deliberately separate: it is a fourth case, outside the three #3885 enumerates, and independent of #3893's other changes.

Reproduced before fixing: with the old call, a textarea holding `한` still held it after the paste while the proxy cleared.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

`npm run format:check`, `npm run lint`, `npx tsc -b` and `npm run test:unit` (3798 passed, 362 files) are clean, as is `node tests/validate-coverage-matrix.mjs`. No screenshot: nothing visual changes. The behavior is the value of a hidden textarea, which the new assertion states more precisely than a capture could.

## Test Coverage Analysis

- [x] Covered: `MobileLiveTerminal.paste.test.tsx` renders the real component and drives a native `beforeinput`, asserting both hidden inputs drop the retained syllable. The test was confirmed to fail against the previous call (`expected '한' to be ''`). `coverage-matrix.json` already maps this spec to `MobileLiveTerminal.tsx` under `terminal.clipboard-chords`, so no matrix change is needed and the validator passes.

**TUI / CLI (e2e test)**
- [x] N/A: web-only change

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Found during a review of #3893, by grepping the other call sites of the helper that PR was fixing rather than from the issue text. The agent reproduced the stale shadow before writing the fix, then confirmed the test fails without it.

https://claude.ai/code/session_01PodUXso6Y16zpy8Py1FmK3

## Summary

- Fixed paste handling for web IME input.
- Clears both the terminal textarea and keyboard-proxy textarea after paste.
- Added Playwright coverage for native `beforeinput` paste events.
- Prevents previously typed IME syllables from remaining after paste.

## Technical details

`MobileLiveTerminal` now passes the local textarea to `invalidateRetainedImeContext` and updates the callback dependencies.

## Benefit

Paste operations now leave both hidden inputs synchronized and clean.

### DIFF
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -1749,8 +1749,9 @@ export function MobileLiveTerminal({
           return true;
         case "insertFromPaste": {
           // The paste lands on the line without passing through the
-          // textarea, so the retained syllable stops mirroring it.
-          invalidateRetainedImeContext();
+          // textarea, so the retained syllable stops mirroring it. Name the
+          // local input: with no target the helper clears only the proxy.
+          invalidateRetainedImeContext(inputRef.current);
           if (input.data) sendData(bracketedPaste(input.data));
           return true;
         }
@@ -1758,7 +1759,7 @@ export function MobileLiveTerminal({
           return true;
       }
     },
-    [sendKeys, sendData, typedWordRef],
+    [sendKeys, sendData, typedWordRef, inputRef],
   );
   const handleBeforeInput = useCallback(
     (ev: InputEvent) => forwardTerminalBeforeInput(ev, handleMobileKeyboardProxyInput),
--- a/web/src/components/__tests__/MobileLiveTerminal.paste.test.tsx
+++ b/web/src/components/__tests__/MobileLiveTerminal.paste.test.tsx
@@ -102,6 +102,39 @@ function stubKeyboardLayout(entries: [string, string][]) {
 }
 
 describe("MobileLiveTerminal paste", () => {
+  // The paste is swallowed by forwardTerminalBeforeInput and never enters the
+  // textarea, so a syllable typed before it stays retained while the line now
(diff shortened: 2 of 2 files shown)


jerome-benoit (approved):
No blocking code finding after two independent sequential reviews and arbitration. `MobileLiveTerminal.tsx:1750–1762` now clears the local textarea and persistent proxy before forwarding the paste; the added callback dependency is present. The component regression at `MobileLiveTerminal.paste.test.tsx:110–136` exercises the real beforeinput handler and checks both fields, and the existing coverage-matrix entry already references it. The async upload issue tracked in #3893 is separate and pre-existing, not duplicated here.

Non-blocking editorial note: this is Vitest/jsdom coverage, not Playwright as the generated summary says; the added test comment can also omit the implementation history.

Evidence: exact-SHA `git show`, full two-file `git diff 4db652f...5fac94d`, and `git diff --check` (success, no output). Both reviewers agree; no local tests or browser reproduction executed.

Code approval does not waive CI: `codecov/project` currently fails at 83.82% (-2.14%), while patch coverage passes. Tests/CI workflows are still queued or incomplete; the cause of the global coverage change has not been established.

PR #3903, opened 2026-09-10
fix(acp): ignore whitespace-only configured model pins

jerome-benoit:
## Description

Follow-up to the remaining whitespace-pin review in #3683. A configured model containing only whitespace was returned as a pin and overrode valid requests. Filter blank values before cloning in the shared configured-model accessor, so neither the pin nor the fallback-default path can reintroduce the invalid value. Nonblank model IDs are unchanged.

Also address the review's provenance qualifications: shorten the effort comments and remove the test that only checked a copied internal flag while claiming a full creation handoff. Existing explicit/inherited respawn-resolution tests remain; this PR does not claim end-to-end creation-to-respawn coverage.

## PR Type

- [x] Bug Fix

## Checklist

- [x] For UI changes: included screenshot or recording (not applicable)

## Test Coverage Analysis

- [x] N/A: shared Rust configuration resolution; no frontend changes

**TUI / CLI (e2e test)**
- [x] Skipped a test, because the pure resolver regression directly proves requested-model precedence and absent-model behavior without launching an agent.

Regression failed before the fix: actual (Some(" \t\n"), None), expected (Some("requested"), Some("high")). After the fix: all 8 resolve_spawn_model_effort tests and 9 respawn_ tests pass. cargo fmt and cargo clippy --lib -- -D warnings pass; commit hook cargo clippy -- -D warnings also passes. No native UI or full creation-to-respawn run is claimed.

## AI Usage

- [x] AI was used

**Any Additional AI Details:** Implemented from the original review finding, checked against current main, with a failing-before/passing-after regression.

## Summary

- Treats empty and whitespace-only ACP model values as unset.
- Prevents blank configured models from overriding requested models or triggering invalid fallback behavior.
- Adds documentation and coverage for the new behavior.
- Removes a misleading respawn handoff test and shortens related comments.

This keeps valid model selection unchanged and makes ACP configuration more predictable.

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -110,6 +110,7 @@ prevent_sleep_idle_grace_minutes = 15  # release once every session has been idl
 [acp.acp_defaults.opencode]
 model = "openai/gpt-5.5"
 # pin_model = true        # make `model` a pin: creation refuses any other model
+# Empty or whitespace-only model values are unset, including with pin_model.
 effort = "high"           # default thinking level
 mode = "plan"             # default mode, applied when the agent advertises one
 
--- a/src/acp/supervisor.rs
+++ b/src/acp/supervisor.rs
@@ -508,13 +508,8 @@ pub struct SpawnRequest {
     pub provider_env: Vec<(String, String)>,
     pub model: Option<String>,
     pub effort: Option<String>,
-    /// Provenance of `effort`: true only when the user explicitly set it
-    /// (persisted in `Instance.acp_effort`), making it a session pin that
-    /// survives a later model-pin change. A nonempty `effort` alone proves
-    /// nothing — the creation path forwards the daemon-resolved default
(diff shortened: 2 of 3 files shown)


jerome-benoit (commented):
No actionable finding after two independent sequential reviews and arbitration. The shared accessor (`src/session/config/mod.rs:1557–1573`) excludes whitespace-only values without changing nonblank IDs. Both pin and default resolution use it (:1642–1651), preserving request precedence and effort lookup. The regression (:4974–4988) covers both a valid request and no request, so the rejected pin cannot return through the fallback.

This addresses the remaining #3683 hardening point. The removed test did not exercise the full creation handoff; the description now states that limitation, and explicit/inherited respawn-resolution coverage remains. Existing reviews and replies were checked; no duplicate finding is added.

Evidence: exact-SHA source/caller inspection and `git diff --check 4db652f cbfee661` (success, no output). CI is green. No local tests/builds were run for this review. COMMENT because this is my own PR.

PR #3905, opened 2026-09-10
fix(tmux): refuse ambiguous live agent poller seeds

jerome-benoit:
## Description

Follow up the ambiguity finding on #3558 / #3872 (discussion_r3978734237). The selected live name now supplies the actual poller seed, so the earlier is_some-only dismissal no longer describes the caller. Both lookup entrypoints now share one unique-match resolver: two live agent matches for the same id8 return None, while a dead duplicate does not disqualify the only live agent. Allocate the returned name only after uniqueness is established.

Preserve the durable SessionKind markers from #3894 and legacy unmarked name filtering; remove the stale comment claiming no durable kind exists. No exact-derived-name exception or fallback to an auxiliary pane is introduced.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

- [x] N/A: Rust-only correction

**TUI / CLI (e2e test)**

The regression failed before the fix: a two-agent snapshot returned aoe_dev_first_abc12345 instead of None. After the fix, all 7 live_agent_ tests pass, including marker-over-name semantics, the single-live/dead-duplicate boundary, poller seed selection and repair refusal. cargo fmt and cargo clippy -- -D warnings pass. The separate agent_seed filter matched zero tests and is not counted as coverage.

## AI Usage

- [x] AI was used

## Summary

- Centralized tmux live-agent lookup through a shared unique-match resolver.
- Rejects ambiguous results when multiple live agents share the same `id8`.
- Ignores dead duplicates when one live agent remains.
- Uses the verified live-agent name as the poller seed.
- Renamed `AgentSeed::OtherKindOnly` to `AgentSeed::NoUniqueAgent` to clarify ambiguous or unavailable matches.
- Preserves existing session markers and legacy name filtering.
- Added regression coverage for duplicate live and dead agents.

This prevents incorrect poller selection in tmux while preserving existing behavior.

### DIFF
--- a/src/session/instance/polling.rs
+++ b/src/session/instance/polling.rs
@@ -102,27 +102,9 @@ fn try_acquire_managed_capture_lease(
     Ok(ManagedCaptureLease(lease))
 }
 
-/// The tmux session name to seed a session-id poller with, or `None` when
-/// this instance has no agent pane for one to follow.
-///
-/// `live_agent` is the live agent session for the id, which a fresh scan
-/// classifies by the kind each session was stamped with at creation. It is
-/// decisive: a row whose agent pane is gone has nothing a session-id poller
-/// can follow, however many paired terminals outlived it. Seeded with one of
-/// those, a poller re-resolves to that same live name every tick, probes
-/// `Alive`, and so never terminates, holding a budget slot for an agent that
-/// is gone while reading session-id state off the wrong pane (#3880).
-///
-/// `derived` is the title-derived agent name, asked only when the scan found
-/// nothing live for the id: a poller started alongside its own tmux session
-/// still needs a target, and `MISSING_TARGET_GRACE` covers the race. It keeps
-/// the name-shape filter because there is no live session to read a kind
-/// marker from, so a title sanitizing under an auxiliary prefix fails closed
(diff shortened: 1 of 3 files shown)


jerome-benoit (commented):
The ambiguous-seed finding is fixed. Both entrypoints share the unique-live-match resolver (`src/tmux/mod.rs:1019–1046`), retaining marker semantics and ignoring dead duplicates. The caller also fails closed: an ambiguous live set reaches `OtherKindOnly` (`src/session/instance/tmux_session.rs:36–50`), and `polling.rs:132–135` refuses a derived-name fallback. Periodic repair likewise stops without a unique candidate (:539–541).

P3, nonblocking terminology: `tmux_session.rs:26–27` still describes `OtherKindOnly` as having no agent, although it now includes multiple live agents. Update that variant's name/documentation to express “live panes, no unique eligible agent” and keep the caller comments consistent. Behavior is correct; no runtime blocker is implied.

Two independent sequential reviews and arbitration; previous #3872 findings and current conversations checked. Evidence: exact-SHA caller inspection and `git diff --check 4db652f b244760b` (success, no output). CI is green. No local tests/tmux reproduction performed. COMMENT because this is my own PR.

PR #3906, opened 2026-09-10
fix(tui): drain drafts only for the requested session

jerome-benoit:
## Description

Address the bounded hardening finding from #3562 / #3868 (review 5166540622). If constructing B fails while A remains mounted, attempting B must not consume A’s own pending draft. This was not a B-to-A text leak, and ordinary absorbed WebSocket/replay errors are not the construction-failure trigger.

Require the requested session ID at the drain boundary and verify the mounted view matches before removing any entry. All three activation callers pass their requested ID, including the daemon-start continuation. Reuse the mutable view directly instead of allocating a copy of its ID. Shorten the overly broad test and continuation comments.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

- [x] N/A: Rust-only correction

**TUI / CLI (e2e test)**

Extended the existing App-level regression: a mismatched requested target leaves the old composer and draft untouched; the matching target drains once, a repeated drain does not duplicate it, and the other session’s draft survives. The pre-fix drain inserted “buffered draft” where the mismatch scenario expected an unchanged empty composer. The corrected test passes; cargo fmt and cargo clippy -- -D warnings pass. This models the retained-old-mount state directly; it does not inject an EmbeddedView construction error or launch the local-daemon confirmation flow.

## AI Usage

- [x] AI was used

## Summary

- Prevented failed session activation from consuming a draft belonging to another mounted session.
- Updated activation paths to pass and verify the requested session ID before draining drafts.
- Added regression coverage for mismatched targets, one-time draining, repeated drains, and preserving other sessions’ drafts.
- Rust formatting and Clippy checks pass.

## Benefit

Users retain draft content when session activation fails or targets a different session.

### DIFF
--- a/src/tui/app.rs
+++ b/src/tui/app.rs
@@ -3216,14 +3216,16 @@ impl App {
             .is_some_and(|v| v.session_id() == session_id)
         {
             self.activate_embedded();
-            self.drain_pending_paste_for_structured_view().await;
+            self.drain_pending_paste_for_structured_view(session_id)
+                .await;
             return Ok(());
         }
         match require_daemon().await {
             Ok(endpoint) => {
                 self.connect_embedded_structured(endpoint, session_id).await;
                 self.activate_embedded();
-                self.drain_pending_paste_for_structured_view().await;
+                self.drain_pending_paste_for_structured_view(session_id)
+                    .await;
             }
             Err(ManagerError::NoDaemonRunning(_)) => {
                 self.home.prompt_start_daemon_for_structured(session_id);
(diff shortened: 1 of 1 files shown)


jerome-benoit (commented):
No actionable finding after two independent sequential reviews and arbitration. `src/tui/app.rs:3250–3264` checks the requested ID against the mounted view before removing its draft, then inserts through that same mutable view. All three activation paths pass the requested ID (:3219,3227,3318).

The regression (:4384–4419) preserves the old view's composer and pending draft on mismatch, inserts once on a matching target, and preserves the other target's draft. This addresses the bounded #3868 hardening point, not a cross-session text leak. It models the retained-old-mount state; it does not inject a constructor failure or run the daemon-start dialog. The pre-existing activation of the old view on construction failure is outside this draft-consumption fix.

Evidence: exact-SHA source/caller inspection, prior replies, and `git diff --check 4db652f 73c68762` (success, no output). CI is green. No local tests/builds were run. COMMENT because this is my own PR.

PR #3909, opened 2026-09-10
fix(tmux): bound agent probes and verify queued callers

### LINKED ISSUE #3886: Agent availability probes can stall callers behind the shared mutex
## Status

Implemented and verified in #3909, pending merge. This issue records the retained findings from merged #3874 and [review 5168401721](https://github.com/agent-of-empires/agent-of-empires/pull/3874#pullrequestreview-5168401721).

## P2: unbounded probes hold the process-wide mutex

At the original baseline `3ca14eb8e9764dad89bb569af60b23042993bf8a`, `probe_agents_available` held `AGENT_PROBE_LOCK` across blocking direct commands and a login shell with no deadline. Later uncached/expired queries and Recheck shared the stall; fresh memo hits bypassed it.

#3909 uses the existing process-group timeout helper with null stdin for all three subprocess paths. The direct deadline is five seconds; the batched login shell has a separate fifteen-second ceiling for slow profiles. These are not a global agent-list budget. Timeouts emit WARN diagnostics with the executable and deadline. Timed-out groups are terminated on Linux/macOS and the direct child is reaped.

## P2: exercise the actual queued-caller contract

The prior contract test made cold, warm and expired calls sequentially. Neither it nor the helper-only partition test depended on the production post-lock memo reread.

#3909 gates a real caller after its initial miss, publishes through another real probe call, resumes it, and asserts no redundant subprocess launches. Positive and negative TTL behavior and one-shell batching are exercised through controlled subprocesses rather than a process-global counter or script-text assertions.

## Maintainer review 5179158671

Addressed in 3f092820: separate shell budget, timeout WARN diagnostics, and the measured startup rationale for one-shell batching restored. A six-second profile fails at the old deadline and succeeds after correction. The ceiling remains finite and the existing negative TTL policy is unchanged. Maintainer re-review is pending.

## Verification

- Before the timeout fix, the controlled stalled-version scenario failed after 12 seconds: the holder, another uncached request and invalidation did not finish. The fixture release guard cleaned up and joined its workers.
- The final regression covers stalled version and login-shell commands, waiting callers/invalidation, and terminated/reaped fixture processes. The fixtures run in isolated test processes to avoid leaking PATH changes into unrelated tests.
- Removing the production post-lock reread fails with a duplicated probe trace. Confirmed through the final isolation wrapper, then restored.
- Full library suite: 6943 passed, 2 ignored. Final tmux suite: 71 passed. Formatter and Clippy with warnings denied pass.

CI additionally caught the Rust test-only subprocess marker missing from the web harness inventory. fce5bcd1 classifies it in the existing inheritance contract with its cfg(test)-only reason. No daemon path or credential isolation was relaxed. All 3 isolation tests and web format/lint/type checks pass after correction.

## AI disclosure

Analyzed, implemented and verified by Codex at @jerome-benoit’s request. No claim of native UI testing is made for this process-level correction.

jerome-benoit:
## Description

Closes #3886. Completes the retained review findings from #3557 / merged #3874.

Direct `which`/version commands use a five-second deadline; the single batched login shell has a separate fifteen-second ceiling to accommodate slow profiles such as nvm. Both use the existing process-group timeout helper with null stdin. Each timeout logs a WARN with the executable and deadline, without logging command arguments. Timed-out groups are terminated on Linux/macOS and the child is reaped. This is not a five-second budget for the entire agent list: an inconclusive direct probe can still use the batched shell fallback. Positive/negative TTL caching, one in-flight probe and the post-lock memo reread remain intact; invalidation is crate-private.

The real queued-caller test pauses after its first cache miss, lets another real call publish, then verifies that release starts no redundant external probes. Script-trace tests replace the process-global counter and helper-only/source-text assertions. Probe fixtures run in isolated test processes so their temporary PATH cannot interfere with unrelated tests.

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

- [x] N/A: no browser behavior change; the existing environment-isolation test classifies the Rust-only subprocess marker

**TUI / CLI (e2e test)**

- Before the timeout fix, the controlled stalled-version scenario failed after 12 seconds while the holder, another uncached request and invalidation waited. Its release guard cleaned up and joined the fixture.
- The timeout regression covers both stalled command types, concurrent requests/invalidation, terminated/reaped fixture processes, and WARN diagnostics for both deadlines.
- The batched-shell regression now includes a six-second profile: it failed at the old five-second deadline with an empty result and passes with the separate shell budget, still proving one shell and continuation after a miss.
- Removing the production post-lock memo reread failed the queued-caller regression with a duplicated `which:claude / version / login` trace, including through the final subprocess-isolation wrapper. Restored afterward.
- `cargo test --lib`: **6943 passed, 2 ignored**.
- After restoring the mutation, `cargo test --lib tmux::tests::`: **71 passed**.
- `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings`: pass. Commit hooks also pass.
- No UI rendering or user-facing configuration change; source comments updated.

CI follow-up fce5bcd1 classifies AOE_AGENT_PROBE_TEST_CHILD in the existing inheritance contract. It is read only inside the Rust cfg(test) module and is absent from aoe serve; no production path or credential isolation rule was relaxed. The 3 isolation tests, web formatter, ESLint and TypeScript checks pass after the correction. The full library suite was rerun at 3f092820: 6943 passed, 2 ignored; all 71 tmux tests, formatter and Clippy also pass. Review 5179158671 is addressed by the split deadline, timeout diagnostics and restored batching rationale. CI at 3f092820 is green: 36 successful checks, 2 skipped, none pending. Test workflow 34605633306 passed on attempt 2 at the same SHA after the initial Docker-fixture ETXTBSY and artifact-upload 403. Maintainer re-review is requested; CHANGES_REQUESTED remains until that decision.

## AI Usage

- [x] AI was used

## Summary

- Added time limits to tmux agent availability probes.
- Terminated and cleaned up stalled probe processes.
- Preserved fallback behavior, caching, invalidation, and single-flight probing.
- Expanded tests for concurrent callers, stalled probes, process cleanup, TTL behavior, and batched shell probing.
- Isolated probe fixtures to prevent environment changes from affecting other tests.
- Updated test isolation rules for the Rust probe test process.

These changes prevent stalled probes from blocking callers while preserving existing availability behavior.

### DIFF
--- a/src/tmux/mod.rs
+++ b/src/tmux/mod.rs
@@ -2547,12 +2547,18 @@ pub(crate) fn is_binary_on_path(binary: &str) -> bool {
         .unwrap_or(false)
 }
 
-/// Cheap availability probe without a login shell. `Some(_)` is definitive:
-/// an explicit path either exists or it doesn't, and a direct `which` /
-/// version-run hit proves the agent is present. `None` means "not found on
-/// the inherited PATH", which is inconclusive because version-manager PATHs
-/// (NVM, etc.) only materialize inside a login shell; the caller decides
-/// whether to pay for that fallback.
+const AGENT_PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+
+fn agent_probe_output(command: &mut Command) -> Option<std::process::Output> {
+    crate::process::run_with_timeout_process_group(
+        command.stdin(std::process::Stdio::null()),
+        AGENT_PROBE_TIMEOUT,
+    )
+    .ok()
+    .flatten()
+}
+
+/// A direct miss or timeout permits a login-shell fallback; a missing explicit path does not.
 fn agent_available_direct(agent: &crate::agents::AgentDef) -> Option<bool> {
     use crate::agents::DetectionMethod;
     match &agent.detection {
@@ -2560,11 +2566,8 @@ fn agent_available_direct(agent: &crate::agents::AgentDef) -> Option<bool> {
             if binary.contains('/') || binary.contains('\\') {
                 return Some(std::path::Path::new(binary).exists());
             }
-            let found = Command::new("which")
-                .arg(binary)
-                .output()
-                .map(|o| o.status.success())
-                .unwrap_or(false);
+            let found = agent_probe_output(Command::new("which").arg(binary))
+                .is_some_and(|output| output.status.success());
             if found {
                 Some(true)
             } else {
@@ -2572,11 +2575,8 @@ fn agent_available_direct(agent: &crate::agents::AgentDef) -> Option<bool> {
             }
         }
         DetectionMethod::RunWithArg(binary, arg) => {
-            let ok = Command::new(binary)
-                .arg(arg)
-                .output()
-                .map(|o| o.status.success())
-                .unwrap_or(false);
+            let ok = agent_probe_output(Command::new(binary).arg(arg))
+                .is_some_and(|output| output.status.success());
             if ok {
                 Some(true)
             } else {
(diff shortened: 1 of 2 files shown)


jerome-benoit (commented):
Reviewed fce5bcd103403c10be8f4910efa79e547d39f9a3 against merge base fbb5bd4c, including #3886 and the retained #3874 review. Both P2 requirements are addressed; no additional actionable finding after two independent sequential analyses and arbitration.

- Direct and batched-shell probes use the existing timeout helper with null stdin (`src/tmux/mod.rs:2550–2559,2569,2578,2634`). The helper handles group termination on Linux/macOS and direct-child reaping (`src/process/mod.rs:79–101,123–148`). This is a per-command deadline, not a five-second budget for the entire roster.
- The real queued-caller regression checks external probe traces after another real caller publishes (`src/tmux/mod.rs:2930–2963`); the production post-lock cache reread remains intact (:2707–2711). Isolated subprocess fixtures replace the global counter, and the new environment marker is cfg(test)-only, not a daemon isolation exception.

Evidence: exact-object source inspection and complete merge-base diff; `git diff --check fbb5bd4c...fce5bcd1` returned 0 with no output. [CI run 34546828070](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34546828070) completed successfully at this exact SHA; current checks are green. No local tests, mutation run or stalled-process reproduction performed during this review. COMMENT because this is my own PR.

njbrake (changes requested):
**Split the login-shell deadline from the direct-probe one.** `AGENT_PROBE_TIMEOUT` covers both a `which` and a full `$SHELL -lc`, and the comment this PR removes put login shells at 0.5–2.5s typical. Past 5s the timeout is cached as a definitive negative for the whole TTL, so `aoe add --tool claude` bails with `'claude' is not installed or not on $PATH` for a merely slow profile. nvm users have both the slowest profiles and the reason the fallback exists. `src/plugin/discover.rs:58-64` already splits a per-probe deadline from a longer ceiling.

**Log the timeout.** `src/git/command.rs` warns on every killed child since one is never routine. Otherwise a false "not installed" carries no diagnostic.

**Keep the batching rationale on `login_shell_probe`.** The 5–10s startup-hang measurement is why the one-shell invariant exists; the replacement states only the what.

Both regression guards fail on the mutation they target.

njbrake (approved):
All three points resolved at 3f092820. Reverting the 15s deadline fails the batching regression and deleting the `warn!` fails the timeout regression, so both are guarded rather than asserted.

Two non-blocking notes. The four probe tests went from 10.7s to 26.5s and the `tmux::` module from 30s to 46s, since the timeout regression waits out a real 15s deadline and the batching fixture sleeps 6s. Deadlines injectable under `cfg(test)` would keep the same mutation sensitivity at a fraction of the runtime; `probe_agents_available` already carries a `cfg(test)` seam to follow.

`src/tmux/mod.rs:4832` has a real newline inside `printf 'login\n'` where its sibling at `:2929` escapes it. Output is identical, so it passes; looks like accidental un-escaping.

PR #3910, opened 2026-09-11
test(acp): guard effort provenance at the spawn boundary

njbrake:
## Description

Restores the guard #3903 removed. `SpawnRequest.effort_explicit` must reach `SpawnConfig.default_effort_explicit` unchanged. Rederiving it from `effort` is the #3683 bug: the create path forwards a daemon-resolved default while `Instance.acp_effort` is `None`, so a nonempty effort read as a session pin makes a later pin move refuse the new model's effort.

Rederive the flag from the value and the lib suite passes without this test, and fails only with it.

The #3683 review asked for an accurate description of the test's scope, not its removal. The name and docstring now claim the supervisor boundary alone; resolution stays covered by the `respawn_` tests.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

Test-only: it restores coverage, not a second fix.

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

`cargo test --lib` 6945 pass, `cargo clippy --lib --tests` and `cargo fmt --check` clean. Injecting `default_effort_explicit: effort.is_some()` fails this test and nothing else in the suite.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Raised during review of #3903; the coverage claim was measured by injecting the regression rather than inferred.

https://claude.ai/code/session_01Mx7gD3WEsjEP6e313WPyWK

## Summary

- Restored a regression test in `src/acp/supervisor.rs`.
- Verified that `SpawnRequest.effort_explicit` reaches the supervisor spawn boundary unchanged.
- Prevented resolved default effort from being incorrectly treated as explicitly pinned effort.
- Added no production code changes.

## Benefits

This test protects against a recurrence of the `#3683` regression and preserves effort provenance during worker spawning. Library tests, Clippy, and formatting checks pass.

### DIFF
--- a/src/acp/supervisor.rs
+++ b/src/acp/supervisor.rs
@@ -4436,6 +4436,61 @@ mod tests {
         }
     }
 
+    /// `effort_explicit` crosses the spawn boundary as its own field rather
+    /// than being rederived from `effort`. The create path forwards a
+    /// daemon-resolved default while `Instance.acp_effort` is `None`, so a
+    /// nonempty effort is not a pin, and reading it as one makes a later pin
+    /// move refuse the new model's effort.
+    ///
+    /// Covers that boundary only. The resolution it feeds is covered by the
+    /// `respawn_` tests.
+    #[tokio::test]
+    #[serial_test::serial]
+    async fn spawn_does_not_rederive_effort_provenance_from_the_value() {
+        let _home = isolate_home();
+        let control = Arc::new(FakeProcessControl::default());
+        control.alive(4343);
+        let entered = Arc::new(tokio::sync::Notify::new());
+        let gate = Arc::new(tokio::sync::Notify::new());
+        let sup = Arc::new(
+            Supervisor::new(VecSink::new())
+                .with_process_control(control)
+                .with_launcher(gated_launcher(entered.clone(), gate.clone(), 4343)),
+        );
+
+        // What the create path sends: an effort resolved from the pinned
+        // model's defaults, with no user selection behind it.
+        let mut req = spawn_request("s-prov");
+        req.effort = Some("low".into());
+        req.effort_explicit = false;
+
+        // The launcher parks on the gate, so spawn runs beside this task the
+        // way the create path's detached spawn does.
+        let spawner = {
+            let sup = Arc::clone(&sup);
+            tokio::spawn(async move { sup.spawn(req).await })
+        };
+        entered.notified().await;
+        gate.notify_one();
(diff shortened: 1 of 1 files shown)


src/acp/supervisor.rs
```diff
@@ -4436,6 +4436,61 @@ mod tests {
...
+    /// daemon-resolved default while `Instance.acp_effort` is `None`, so a
+    /// nonempty effort is not a pin, and reading it as one makes a later pin
+    /// move refuse the new model's effort.
+    ///
+    /// Covers that boundary only. The resolution it feeds is covered by the
+    /// `respawn_` tests.
+    #[tokio::test]
+    #[serial_test::serial]
+    async fn spawn_does_not_rederive_effort_provenance_from_the_value() {
+        let _home = isolate_home();
```
jerome-benoit, line 4450: P3, nonblocking: please use the restoring test-home guard for this added call. The local isolate_home helper (:6882–6889) sets HOME/XDG_CONFIG_HOME but returns only a TempDir; dropping it deletes the directory without restoring those variables, leaving process-global state for later tests. #[serial] does not restore it. A local TempDir followed by crate::session::test_support::isolate_home(tmp.path()) preserves the same setup and restores the environment before directory deletion. The helper debt is preexisting; no suite failure was reproduced and no global refactor is requested.

jerome-benoit (approved):
Thanks for restoring this guard. The earlier #3683 review asked to narrow its coverage claim, not remove it; the #3903 review should have caught that loss.

At df265c0a, the test traverses real Supervisor::spawn and checks the installed SpawnConfig (src/acp/supervisor.rs:4462–4492). It protects the separate provenance mapping at :1948–1949, which the respawn tests bypass by constructing their config directly. Its stated boundary-only scope is accurate. No production blocker found.

One nonblocking P3 inline concerns environment restoration. Two independent sequential analyses agreed on the regression's value; arbitration downgrades the isolation concern from P2 to P3 because the leaking helper predates this PR and no resulting CI failure was established. No global fixture refactor is requested.

Evidence: exact-object source/helper inspection; git diff --check 9283dc49...df265c0a returned 0 with no output. [Tests](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34602452533) succeeded at this exact SHA; current checks are green. No local tests or mutation run were performed; the full-suite mutation result remains the author's reported verification.

PR #3917, opened 2026-09-11
fix(web): frame the open session's sidebar row in the theme accent

### LINKED ISSUE #3912: Web: make the active sidebar session unmistakable across dark themes
> **AI disclosure:** This issue was drafted with OpenAI Codex from behavior I observed. I will review the final text before submitting.

### Problem

In the web dashboard, the currently active session is difficult to distinguish from other sidebar rows, especially when many sessions are visible.

This affects all dark built-in themes: `zinc`, `empire`, `phosphor`, `tokyo-night-storm`, `dracula`, `rose-pine`, and `deep-ocean`.

The active row currently uses:

`bg-surface-850 border-l-2 border-brand-600`

For dark themes, `surface-850` is derived by mixing only 5% white into the background. The resulting contrast against `surface-900` is approximately 1.12–1.17:1, so the background change is barely perceptible. The single left border is also easy to miss.

AoE already exposes and projects a semantic `session_selection` theme value, but the web sidebar does not currently use it for the active row.

### Proposed behavior

Give the active session a prominent indicator around the complete row, rather than only along its left edge.

One simple option is:

`ring-2 ring-inset ring-brand-500`

This produces a full two-pixel frame using the selected theme's accent color. Across the seven dark built-in themes, the corresponding accent-to-background contrast ranges from approximately 5.60:1 to 13.70:1.

Alternatively, the sidebar could use `session_selection` as its active-row background, provided the focus remains equally obvious and text/status colors retain sufficient contrast.

### Acceptance criteria

- The active session is immediately recognizable in every dark built-in theme.
- The indicator surrounds the complete row instead of relying only on a subtle background or left edge.
- It remains visible in the compact and regular sidebars.
- Hover styling does not override the active-session indicator.
- Sidebar multi-selection remains visually distinct from the active session.
- Session status colors remain readable.

### Reproduction

1. Open the AoE web dashboard with several sessions.
2. Select any dark built-in theme.
3. Open one session.
4. Compare its sidebar row with the surrounding rows.

Observed on AoE 1.15.3 and confirmed unchanged in AoE 1.16.0/current `main` at the time of filing. The relevant code is in [`WorkspaceSidebar.tsx`](https://github.com/agent-of-empires/agent-of-empires/blob/v1.16.0/web/src/components/WorkspaceSidebar.tsx#L1547-L1554).

njbrake:
## Description

The active row leaned on `bg-surface-850` plus a 2px left border. On the dark
builtins surface-850 is the background with 5% white mixed in: 1.12:1 to 1.17:1
against surface-900, so the open session reads as just another row.

New `session-active` token, derived in the theme projection: the accent, lifted
toward the appearance's readable pole only as far as it takes to clear WCAG's
3:1 non-text floor against both the row fill and the sidebar background. All
seven dark builtins keep their accent (4.5:1 to 11.2:1 against the fill);
catppuccin-latte darkens its orange from 2.6:1 to 3.2:1. Plugin and user themes
get the same floor. The sidebar draws it as a full inset ring, and
active/selected/hover chrome now resolves in one helper instead of leaving a
simultaneous `ring-1` / `ring-2` to stylesheet order.

Using `session_selection` as the row background was the other option in the
issue. It is not viable without porting the TUI's contrast fallback: against
that token dracula's idle, dimmed and hint all sit at 1.00:1, deep-ocean idle at
1.20:1, tokyo-night-storm idle at 1.44:1.

Fixes #3912

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

**Any Additional AI Details you'd like to share:** Contrast measurements and the
token derivation were checked against the real projection output for every
builtin theme; the Rust test asserts the floor on all of them.

https://claude.ai/code/session_012gTpUFni8gGUu5AHi55WYq

## Summary

- Added a theme-aware `session-active` indicator with sufficient contrast across built-in and custom themes.
- Replaced the subtle background and left border with a full inset ring around the active sidebar row.
- Applied the indicator to collapsed group headers.
- Centralized active, selected, and hover styling.
- Added coverage for theme contrast, row styling, ring rendering, collapsed groups, and coverage tracking.
- Updated design guidance for the active-session indicator.

## Benefits

The active session is easier to identify across themes, sidebar sizes, hover states, collapsed groups, and multi-selection states. The indicator remains visible without reducing session status readability.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -86,11 +86,10 @@ content. It is denser and quieter than the marketing site.
 - Use semantic theme tokens such as `bg-surface-900`,
   `text-status-running`, and `border-surface-700`.
 - All body text must meet WCAG AA contrast against its surface.
-- Selection and focus indicators must meet WCAG's 3:1 non-text contrast
-  against both the surface they sit on and the one behind it. A surface-ramp
-  step alone does not reach it; the sidebar's open session uses the
-  `session-active` token, which the projection lifts from the theme accent
-  until it clears that floor.
+- The open session's sidebar row must meet WCAG's 3:1 non-text contrast
+  against every surface its indicator can sit on. A surface-ramp step alone
+  does not reach it, so the row uses the `session-active` token, which the
+  projection lifts from the theme accent until it clears that floor.
 - Prefer instant state changes or `transition-colors`. Named motion is limited
   to existing fade, slide, and terminal-cursor behavior.
 - Fixed palette colors are allowed only when hue carries meaning, such as
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -86,6 +86,11 @@ content. It is denser and quieter than the marketing site.
 - Use semantic theme tokens such as `bg-surface-900`,
   `text-status-running`, and `border-surface-700`.
 - All body text must meet WCAG AA contrast against its surface.
+- Selection and focus indicators must meet WCAG's 3:1 non-text contrast
+  against both the surface they sit on and the one behind it. A surface-ramp
+  step alone does not reach it; the sidebar's open session uses the
+  `session-active` token, which the projection lifts from the theme accent
+  until it clears that floor.
 - Prefer instant state changes or `transition-colors`. Named motion is limited
   to existing fade, slide, and terminal-cursor behavior.
 - Fixed palette colors are allowed only when hue carries meaning, such as
--- a/src/tui/styles/resolved.rs
+++ b/src/tui/styles/resolved.rs
@@ -229,6 +229,16 @@ fn web_projection(theme: &Theme, appearance: ThemeAppearance) -> CssVarProjectio
         hex(readable_on(brand_ramp[5])),
     );
 
+    // Frame around the open session's sidebar row. It is a hairline over
+    // the row fill, so it has to clear the WCAG 1.4.11 non-text floor
+    // against both that fill and the surrounding background. Most accents
+    // already do; the rest lift toward the appearance's readable pole
+    // until they clear it.
+    css.insert(
+        "--color-session-active".into(),
+        hex(active_frame(accent, bg, elevated_2, appearance)),
+    );
+
     // Accent ramp anchored on theme.terminal_border (the existing
     // teal-style anchor used by the TUI's accent surface), so secondary
     // affordances like branch chips still read as the theme's secondary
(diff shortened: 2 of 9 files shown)


jerome-benoit (changes requested):
Reviewed 398a10d4 against 004ea029, including #3912 and the existing discussion. The full-frame approach and explicit active/selected precedence are sound. Two P2 corrections remain before merge, both already covered by CodeRabbit; no duplicate inline comments.

- **P2: choose a contrasting pole from the actual surfaces** ([existing thread](https://github.com/agent-of-empires/agent-of-empires/pull/3917#discussion_r3991724707)). `resolved.rs:145–147` accepts explicit appearance; `:412–423` can therefore return white for a dark-declared custom theme with white background/accent, producing 1:1. Choose a pole that satisfies both surfaces and add a regression for inconsistent metadata.
- **P2: cover the active-selected background** ([existing thread](https://github.com/agent-of-empires/agent-of-empires/pull/3917#discussion_r3991724715)). This also affects builtin Catppuccin Latte, not only custom themes. `sessionRowChrome.ts:20` tints the actual `surface-800` sidebar (`WorkspaceSidebar.tsx:3772`), whereas `resolved.rs:238` validates the untinted fill. Recomputing the source formulas gives frame RGB(229,90,10), fill RGB(232,234,238), and selected composite RGB(235.3,213.9,203.95): contrast falls from 3.01:1 to about 2.59:1. Keep the validated fill for active rows or validate the actual composite; test active+selected too.

Evidence: complete eight-file immutable diff, exact-SHA source/theme inspection, clean `git diff --check`, and independent `node -e` sRGB contrast calculations. These are numerical reproductions, not browser measurements. Two independent sequential reviews confirmed both findings; arbitration corrected the composite calculations to use the actual sidebar background. No additional finding: token wiring, deterministic ring precedence, hover handling and coverage-matrix integration are present. CI, Tests and CodeQL succeeded at this SHA; no local build or browser/test suite was run.

jerome-benoit (approved):
Re-reviewed bc60ef35, including the follow-up commit and existing threads. Both previous P2s are addressed: `resolved.rs:418–428` chooses the pole from the actual background and validates the selected composite as well as both opaque surfaces. The extended regressions at :601–649 cover builtin selected fills and inconsistent appearance metadata. Independent numerical checks give approximately 3.20:1 for selected Latte and 3.95:1 for the white/dark counterexample.

The remaining [test assertion comment](https://github.com/agent-of-empires/agent-of-empires/pull/3917#discussion_r3992118308) is valid P3/nonblocking: `sidebar-multi-session.spec.ts:184–189` checks presence, not equality with the projected token. Comparing normalized colors would strengthen the regression test; no incorrect production color is established. No duplicate inline or additional blocker.

Two independent sequential analyses and arbitration agree. Evidence: exact-SHA source and complete delta inspection, clean `git diff --check`, and lightweight numerical calculations, not a browser reproduction. CI, Tests and CodeQL passed at this SHA. No local build or test suite was rerun.

jerome-benoit (approved):
Re-reviewed c46bea18, including the delta after bc60ef35 and current discussions. No blocking finding after two independent sequential analyses and arbitration.

The normal and rename headers consistently use the semantic active token (`WorkspaceSidebar.tsx:2861,2905`), without changing activation or focus handlers. The previous P3 is fixed: `sidebar-multi-session.spec.ts:184–192` now compares the computed ring color with the projected token and still checks that a shadow paints.

I do not consider the new [full-header-frame request](https://github.com/agent-of-empires/agent-of-empires/pull/3917#discussion_r3992349113) blocking: #3912 and DESIGN require a full frame for the active session row, not its aggregate group header. The autogenerated release-note claim about a full group-header frame should be corrected, rather than treated as a new requirement. This review does not claim exhaustive contrast coverage for custom-tinted group headers.

Evidence: exact-SHA two-file delta and surrounding styles, live issue/threads, clean `git diff --check`. The initially incomplete coverage report has updated: patch/project checks now pass at c46bea18. CI passed; Tests and CodeQL are still pending, not claimed successful. No local build, test suite or browser run.

jerome-benoit (approved):
Re-reviewed c5280012, the test-only follow-up after c46bea18. No new finding after two independent sequential analyses and arbitration.

`sidebar-grouping.spec.ts:177–200` explicitly opens s-a, checks that only its collapsed group receives the marker, and compares the painted border with the normalized theme token. The awaited CSS assertion accommodates the existing color transition; the color probe is removed (:83–90). The coverage-matrix entry now identifies the mocked browser layer while retaining the related specs.

Evidence: exact two-file delta, fixture/consumer/matrix inspection and clean `git diff --check`. This test verifies token application, not every theme's contrast; the existing Rust tests cover the latter separately. No production change, local build or test-suite execution in this pass. Approval is for the reviewed code; the new SHA's CI results must still be checked before merge.

Issue #3920, opened 2026-09-11
Web dashboard terminal: a wrapped URL linkifies only its first row, truncating the href


njbrake:
Split out of #3918, which reported it alongside the inert-link bug. The click fix does not address this.

A URL long enough to wrap onto the next terminal row linkifies only its first row. `splitUrls` runs per visual row, so the anchor ends at the wrap column and the href is truncated:

```
row 1: (https://github.example.com/trusted-environments/control    <- anchor ends here
row 2: -plane/issues/1375#issuecomment-287842). Per the spec's     <- plain text
```

```json
{ "href": "https://github.example.com/trusted-environments/control", "target": "_blank" }
```

Hyphens are valid URL characters, so a per-row scanner cannot recover the continuation heuristically. The correct fix joins wrapped rows into the logical line before URL detection, then maps the match back onto the visual rows.

The blocker is that the wire format carries no wrap information, on either transport:

- `vt100::Screen::row_wrapped()` exists and is reachable in `grid_content` (`src/tmux/vt.rs:1795`), but is never called, and `VtSample` / `frame_meta` carry no wrap data.
- It would also be wrong for scrollback. The grid is seeded from `capture-pane -e -N` output with LF promoted to CRLF (`lf_to_crlf`, `src/tmux/vt.rs:999`), so seeded rows enter the parser as hard breaks and read `row_wrapped == false` even where the original output autowrapped. Reseeds on every resize/resync reset that.
- The `capture-pane` fallback (`src/tmux/session.rs:917`, split windows / non-unix / `vt_live` off) has no wrap information at all. tmux offers only `-J`, which joins wrapped rows and loses the row boundary rather than flagging it.
- A single-pane frame does not carry pane width either (`frame_meta` emits `pane0: null` unless composited), so the client cannot fall back on a "row is exactly `cols` wide" heuristic.

Two client-side cases are fixable without any of that plumbing, and may be worth doing first:

- A URL split across ANSI style segments within one row. `splitUrls` runs per segment, so a color change mid-URL truncates the href the same way.
- Client-side rewrap. When `renderCols` is narrower than the pane, `wrapLine` splits a logical line into visual rows and each row linkifies independently.

Both go away if URL detection moves to the logical line and annotates `seg.url`, the way OSC 8 hyperlinks already work, before `wrapLine` runs.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #3921, opened 2026-09-11
fix(web): let terminal link clicks reach the anchor

### LINKED ISSUE #3918: Web dashboard terminal: links are inert — pointer capture on the terminal container swallows every click on an anchor
## Environment

- aoe 1.16.0 (Homebrew bottle), macOS (Darwin 25.6.0)
- Web dashboard served by `aoe serve`, viewed in Chromium
- Terminal pane of a session view (`/session/<id>`)

## Summary

Detected URLs in the live terminal render as `<a target="_blank">` anchors, but a click on them does nothing.
No new tab opens, no navigation happens, and no console error appears.

The cause is the terminal's drag-to-select handler.
On `pointerdown`, the terminal container calls `preventDefault()` and `setPointerCapture()` on itself.
Pointer capture retargets `pointerup` — and therefore the `click` event — to the container `div`, not the anchor.
The anchor never receives the click, so its default navigation never runs.

## Steps to reproduce

1. Run `aoe serve` and open the web dashboard.
2. Open a session whose agent printed a URL. The URL renders underlined with `cursor: pointer`.
3. Click the link.

## Expected behavior

The URL opens in a new tab (the anchor has `target="_blank"`).

## Actual behavior

Nothing happens. No tab opens and no error is logged.

## Evidence

Event trace captured with window-level listeners during a real click on the anchor:

```json
[
  { "type": "pointerdown", "target": "SPAN (inside <a>)", "defaultPrevented_after_bubble": true },
  { "type": "pointerup",   "target": "DIV.absolute.inset-x-0.top-0.bottom-[8px] (terminal container)" },
  { "type": "click",       "target": "DIV.absolute.inset-x-0.top-0.bottom-[8px] (terminal container)" }
]
```

Notes on the trace:

- `mousedown` / `mouseup` never fire, because a handler calls `preventDefault()` on `pointerdown` (this suppresses the compatibility mouse events).
- Instrumenting `Element.prototype.setPointerCapture` shows the terminal container captures the pointer on every `pointerdown`:

```json
{ "tag": "DIV", "cls": "absolute inset-x-0 top-0 bottom-[8px] font-mono flex ...", "pointerId": 1 }
```

- Because the pointer is captured, `pointerup` and `click` retarget to the container. The `click` target is not inside the anchor, so the browser performs no navigation.
- Control test: dispatching a synthetic `click` event directly on the anchor (bypassing the pointerdown/pointerup sequence) opens the tab as expected. This isolates the fault to the pointer-capture gesture handling, not the anchor itself.

## Suggested fix

Options, roughly in order of preference:

1. On `pointerdown`, skip `preventDefault()` + `setPointerCapture()` when the event target is inside an `<a>` (or defer capture until the pointer moves past a small drag threshold, which also preserves drag-to-select starting on a link).
2. Keep the capture, but on `pointerup` with no selection drag, resolve the element under the pointer (`document.elementFromPoint`) and open the anchor's href manually.

## Related bug found while testing: wrapped URLs are linkified per visual row, truncating the href

Even after the click is fixed, a long URL that wraps onto the next terminal row produces a wrong link.
URL detection runs on each visual row independently, so the anchor stops at the wrap column:

- Row 1: `(https://github.example.com/trusted-environments/control` ← anchor ends here
- Row 2: `-plane/issues/1375#issuecomment-287842). Per the spec's` ← plain text

DOM evidence — href and text both stop at the wrap point:

```json
{
  "href": "https://github.example.com/trusted-environments/control",
  "text": "https://github.example.com/trusted-environments/control",
  "target": "_blank",
  "rel": "noopener noreferrer"
}
```

Fix direction: join wrapped rows into the logical line (using the terminal buffer's wrap flag) before URL detection, then map the match back onto the visual rows — the same approach xterm.js link providers use. Hyphens are valid URL characters, so a per-row scanner cannot recover the continuation heuristically.

Happy to split this second item into its own issue if you prefer.

## Duplicate check

- #3519 (closed) covered OSC 8 escapes rendered as text in the web terminal — different mechanism.
- #3735 (closed) covered OSC 8 targets in the TUI — different surface.

njbrake:
## Description

When an agent prints a URL in the web dashboard's terminal, the dashboard turns it into a link. Clicking that link did nothing at all: no tab opened, no error appeared. The reason is that while a full-screen agent is running, the terminal grabs the mouse so it can pass presses and drags through to the agent, and that grab was strong enough that the browser stopped treating a press on a link as a click on the link. This change makes the terminal let go for ordinary left-clicks that land on a link, so the link opens in a new tab like any other link on the page. Right-clicks and middle-clicks still go to the agent as before, so nothing the agent does with the mouse changes.

Mechanically: in forward mode (`altScreen && frame.mouse`) `onPointerDown` calls `preventDefault()` plus `setPointerCapture()` on the scroller. Capture retargets `pointerup`, and therefore `click`, to the scroller, so the anchor never received the click and its default navigation never ran. The fix skips forwarding for a plain primary press whose target is inside an `a[href]`.

Not fixed here: the issue also reports that a URL wrapping onto the next terminal row linkifies only its first row, truncating the href. That needs a per-row wrap flag the wire format does not carry. `vt100::Screen::row_wrapped()` exists but is never called and would read `false` for every pre-seed row, since the grid is seeded from `capture-pane` output with LF promoted to CRLF; the `capture-pane` fallback transport has no wrap information at all, and a single-pane frame does not even carry pane width. Tracked separately in #3920.

Verified in real Chromium, since jsdom implements neither pointer capture nor the retargeting that caused the bug. `web/tests/live-link-click.spec.ts` drives the built bundle with real mouse and touch input against a full-screen SGR-mouse frame. Desktop: the click opens a tab at the anchor's href and forwards no mouse bytes; a click beside the link still forwards `ESC[<0;..M`/`m`; a right-click on the link still forwards `ESC[<2;..M`. Mobile (iPhone 13): the same tap opens the link. Reverting the one-line guard and rebuilding fails the desktop link case and nothing else.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

Diagnosis, fix and tests were produced through back-and-forth with @njbrake.

Fixes #3918

https://claude.ai/code/session_01DhLtwFoJdAXtoVUb3H2581

## Summary

- Fixed terminal URL links in the web dashboard so primary clicks and mobile taps open the link.
- Preserved forwarding for clicks outside links and for right-click actions.
- Added unit and Playwright coverage for desktop and mobile behavior.
- Updated the coverage matrix.

## Benefit

Users can open detected terminal URLs normally without breaking terminal mouse controls.

Wrapped URLs that span terminal rows remain a separate follow-up issue.

### DIFF
--- a/web/src/components/MobileLiveTerminal.tsx
+++ b/web/src/components/MobileLiveTerminal.tsx
@@ -1320,6 +1320,11 @@ export function MobileLiveTerminal({
       if (e.pointerType !== "mouse" || !forwardModeRef.current || e.shiftKey) return;
       const base = e.button === 1 ? 1 : e.button === 2 ? 2 : e.button === 0 ? 0 : -1;
       if (base < 0) return;
+      // A primary press that lands on a linkified URL belongs to the browser.
+      // The preventDefault and pointer capture below would retarget the click
+      // to this container, so the anchor would never navigate (#3918). Other
+      // buttons still reach the app, which keeps its context menu suppression.
+      if (base === 0 && (e.target as Element | null)?.closest?.("a[href]")) return;
       e.preventDefault();
       // Keep the hidden input focused so the physical keyboard still types
       // even though we suppressed the click's default focus.
--- a/web/src/components/__tests__/MobileLiveTerminal.forward.test.tsx
+++ b/web/src/components/__tests__/MobileLiveTerminal.forward.test.tsx
@@ -434,3 +434,37 @@ describe("MobileLiveTerminal forward-mode flick momentum", () => {
     }
   });
 });
+
+describe("MobileLiveTerminal link clicks in forward mode", () => {
+  const linkFrame = () =>
+    frame({ content: "see https://example.com/x\n", altScreen: true, mouse: true, mouseSgr: true });
+
+  it("lets a primary press on a linkified URL reach the anchor", () => {
+    // Forwarding a press calls preventDefault and captures the pointer on the
+    // scroller, which retargets the click away from the anchor and leaves the
+    // link inert (#3918).
+    const { container, forwardButton } = renderTerm(linkFrame());
+    const anchor = container.querySelector("a[href='https://example.com/x']") as HTMLElement;
+    expect(anchor).not.toBeNull();
+    // A real press lands on one of the anchor's cell spans, not the anchor.
+    const target = anchor.querySelector("span") as HTMLElement;
+    expect(target).not.toBeNull();
+    const down = fireEvent.pointerDown(target, { pointerType: "mouse", button: 0, clientX: 10, clientY: 10 });
+    expect(down).toBe(true); // not defaultPrevented
+    fireEvent.pointerUp(anchor, { pointerType: "mouse", button: 0, clientX: 10, clientY: 10 });
+    expect(forwardButton).not.toHaveBeenCalled();
+  });
+
+  it("still forwards a press on plain output", () => {
+    const { scroller, forwardButton } = renderTerm(linkFrame());
+    fireEvent.pointerDown(scroller, { pointerType: "mouse", button: 0, clientX: 10, clientY: 10 });
+    expect(forwardButton).toHaveBeenCalled();
+  });
+
+  it("still forwards a right-click on a link so the app keeps its own menu", () => {
+    const { container, forwardButton } = renderTerm(linkFrame());
+    const anchor = container.querySelector("a[href='https://example.com/x']") as HTMLElement;
+    fireEvent.pointerDown(anchor, { pointerType: "mouse", button: 2, clientX: 10, clientY: 10 });
+    expect(forwardButton.mock.calls[0]![0]).toBe(2);
+  });
+});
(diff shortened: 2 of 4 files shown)


jerome-benoit (approved):
Reviewed 2fd57b30 with two independent sequential analyses and cross-validation. No actionable findings.

The guard at web/src/components/MobileLiveTerminal.tsx:1327 leaves primary link presses to the browser before capture or forwarding; ordinary output and right/middle buttons retain their existing path. The real-input Chromium tests cover navigation and preserved forwarding (web/tests/live-link-click.spec.ts:97-140), rather than relying on synthetic clicks. This matches the pointer-capture targeting rules: https://www.w3.org/TR/pointerevents3/#event-dispatch

Wrapped-URL truncation is correctly tracked separately in #3920 and does not block this focused fix. CI and coverage checks are green on this SHA; the mocked Playwright job reports 438 passed. Static inspection and git diff --check completed locally; no local browser or test-suite run.

PR #3922, opened 2026-09-11
ci: run the core-rest leg under a symlinked temp root

njbrake:
## Description

macOS reaches its per-user temp root through a symlink, so a test comparing `TempDir::path()` against a path that has been through `canonicalize()` or a reported cwd passes on Linux and fails on macOS. The macOS leg runs post-merge only, so this class lands on main before anyone sees it. Twice so far: #3535 compared a canonicalized observation against a raw `tempfile::tempdir()` path, #3919 compared a store path against a cwd Node had already resolved.

Point `TMPDIR` at a symlink for the core-rest leg, reproducing the difference where the unit tests run.

Verified against #3919: its failing test fails identically under a symlinked root on Linux, and passes once fixed.

**Scope.** core-rest only (`cargo test` = lib + bins + integration), which is the surface measured clean. The e2e legs are left alone; they could not be measured cleanly and they are the ones binding sockets under `temp_dir()`, where `sun_path` headroom matters. The symlink path is short for the same reason.

**Measurements.**

| suite | control | symlinked |
| --- | --- | --- |
| `cargo test --lib` x3 (6953 tests) | 0, 0, 2 failures | 0, 0, 1 failure |
| `cargo test --test integration` | not run | 289 passed, 0 failed |
| `cargo test --test e2e` | 11 failures | 12 failures |

The lib failures on both sides come from one population of pre-existing flaky tests (`acp_client::steer`, `acp_client::spawn`, `session::instance::terminal`, `session::smart_rename`), each passing in isolation. No runtime difference: 153s symlinked vs 151s control.

The e2e numbers are from a sandbox with 11 baseline failures (missing agent stubs, container process handling), so that row proves nothing either way. The two sets are identical except `live_send_paste_e2e::test_live_send_paste_into_non_bracketed_pane_has_no_marker_debris`, which fails 2/2 in isolation under both conditions.

Known gap: a path bug that only shows up in the e2e suite still reaches main.

## PR Type

- [x] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [x] AI was used

Measurements run by the model, reviewed by @njbrake.

https://claude.ai/code/session_014ivbF2Ag256wWuUk2Vwe8s

## Summary

- Run `core-rest` tests with a symlinked `TMPDIR` to reproduce macOS-style paths on Linux.
- Log the effective temporary directory for easier CI verification.
- Keep e2e jobs unchanged because temporary socket behavior produced unreliable results.
- Validate 289 integration tests successfully. Library test failures remain pre-existing and flaky.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,6 +102,25 @@ jobs:
             --test integration -- daemon_core_web_optional plugin_install
           cargo test --features web,e2e-tests --lib -- \
             server::assets:: tui::dialogs::serve::qr_seam
+      # macOS reaches its per-user temp root through a symlink, so a test that
+      # compares `TempDir::path()` against a path that has been through
+      # `canonicalize()` or a reported cwd passes on Linux and fails there.
+      # That asymmetry has broken main twice (#3535, #3919), and the macOS
+      # leg only runs post-merge, so nothing catches it on the PR. Pointing
+      # TMPDIR at a symlink reproduces the difference on the leg that runs the
+      # unit tests, where this class lives.
+      #
+      # Scoped to core-rest deliberately: lib + bins + integration were measured
+      # clean under a symlinked root, the e2e legs were not, and those are the
+      # ones binding sockets under `temp_dir()`. Keep the path short for the
+      # same reason (`sun_path` is 104 bytes on macOS, and src/tmux/vt.rs binds
+      # inside `temp_dir()`).
+      - name: Symlinked temp root
+        if: matrix.leg == 'core-rest'
+        run: |
+          mkdir -p /tmp/aoe-real
+          ln -sfn /tmp/aoe-real /tmp/aoe-link
+          echo "TMPDIR=/tmp/aoe-link" >> "$GITHUB_ENV"
       # core-rest: the full default-feature suite minus e2e (lib + bins +
       # the consolidated integration binary). The daemon is core, so this
       # leg covers it end to end with no dashboard bundle and no npm in the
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -102,14 +102,37 @@ jobs:
             --test integration -- daemon_core_web_optional plugin_install
           cargo test --features web,e2e-tests --lib -- \
             server::assets:: tui::dialogs::serve::qr_seam
+      # macOS reaches its per-user temp root through a symlink, so a test that
+      # compares `TempDir::path()` against a path that has been through
+      # `canonicalize()` or a reported cwd passes on Linux and fails there.
+      # That asymmetry has broken main twice (#3535, #3919), and the macOS
+      # leg only runs post-merge, so nothing catches it on the PR. Pointing
```
jerome-benoit, line 109: P3 (non-blocking): macOS tests can already run on a PR with the `ci-macos` label (see the conditions at lines 289 and 297). Please say they do not run on PRs **by default**, rather than only post-merge. That preserves the rationale without hiding the existing on-demand validation path. No workflow behavior change is needed.

jerome-benoit (approved):
Reviewed 9bf1e8e1 with two independent sequential analyses and cross-validation. No blocking findings; one nonblocking wording correction below.

The setup and cargo-test steps share the core-rest condition, and GITHUB_ENV propagates TMPDIR before the test command (.github/workflows/tests.yml:118-135). The other matrix legs run on separate hosted runners.

Verified with `gh run view 34639477967 --job 103395376081 --log`: TMPDIR=/tmp/aoe-link, 6,956 library tests passed and 289 integration tests passed. CI and coverage checks are green for this head. This is Linux symlink-path coverage, not native macOS or e2e symlink coverage; the macOS test steps were skipped on this unlabeled PR.

No local build or test suite was run; git diff --check was clean.

PR #3915, opened 2026-09-11
refactor(session): simplify Prime capture follow-ups

jerome-benoit:
## Description

Address the four code/documentation follow-ups requested in [the final review of #3776](https://github.com/agent-of-empires/agent-of-empires/pull/3776#pullrequestreview-5180564449). Rebased onto its merge commit `b438e91d`; this PR contains only the follow-up delta.

- Remove the Prime-only write to `identity_publisher_launched`. Prime uses the `ManagedExclusiveStore` branch before that field is read; keep the field and its other publisher users intact.
- Remove the single-caller, test-only `prime_agent_poll_fn_sandboxed_store` wrapper. The fallback test now exercises the production factory with an absent preferred publication.
- Use `PRIME_AGENT_DIR_IN_CONTAINER` as the canonical container root for agent environment metadata, settings/store resolution, extension arguments, and sidecar paths. Remove redundant root/default-session/extension path constants; derive suffixes directly in the existing output strings, without an extra path registry, dependency, or intermediate path allocation. Keep Pi on its existing constant.
- Expand RLM as “recursive language model” in the resume guide and explain that these are child sessions.

No change to root-only attribution, Ready/Pending semantics, exclusion handling, the pre-budget argv check, the post-budget plan, or the retry deadline. No persisted paths change, so no migration is needed.

The fifth follow-up, intermittent `plugin_settings` fixture contamination, is tracked separately in **#3914**. A disposable isolated custom-agent configuration reproduced the same registry-membership assertion; the issue contains the reproducer and proposed isolation fix. The temporary test prologue was removed, and this PR does not modify plugin-settings code.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

## Test Coverage Analysis

**Web dashboard / structured view (Playwright user story)**
- [x] N/A: this PR does not change a user-facing dashboard flow
- [ ] Added or updated a Playwright spec under `web/tests/` and updated `web/tests/coverage-matrix.json`
- [ ] Skipped a test, because:

**TUI / CLI (e2e test)**
- [x] N/A: this PR does not change TUI rendering, a CLI subcommand, or session lifecycle behavior
- [ ] Added or updated an e2e test under `tests/e2e/`
- [ ] Skipped a test, because:

[...]

PR #3924, opened 2026-09-11
refactor(session): quote the whole Prime extension path

njbrake:
## Description

Follow-up to #3915. That PR replaced the full-path `PRIME_AGENT_EXTENSION_IN_CONTAINER` constant with a root constant plus an inline suffix, which left `shell_escape` wrapping only the directory:

```
-e '/root/.prime/agent'/extensions/aoe-session-id.js
```

Not a live bug. `PRIME_AGENT_DIR_IN_CONTAINER` has no shell-special characters, and the concatenation parses to the same argv word under bash and dash either way. The problem is that the call reads as though it escapes the path it builds, which turns into a real bug the day the constant changes. Escape the joined path so the guarantee matches the appearance.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the emitted argv word is unchanged, so a regression test would assert the same string before and after. Adding one would need a container launch to be worth anything.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (6,956 + 289 passed, 0 failed) all pass. Rust only, no web build.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Found during review of #3915, raised there as non-blocking, split out at the maintainer's request.

**NOTE:**

https://claude.ai/code/session_01Xk8X6bw944fd8v3HHkxrbB

## Summary

- Updated Prime extension path handling to escape the complete path, including the filename.
- Preserved the current emitted argument.
- Improved safety if path constants change in the future.
- Formatting, linting, and all 7,245 tests pass.

### DIFF
--- a/src/session/instance/session_id.rs
+++ b/src/session/instance/session_id.rs
@@ -725,8 +725,10 @@ impl Instance {
                     (
                         Path::new(PRIME_AGENT_DIR_IN_CONTAINER),
                         format!(
-                            " -e {}/extensions/aoe-session-id.js",
-                            shell_escape(PRIME_AGENT_DIR_IN_CONTAINER),
+                            " -e {}",
+                            shell_escape(&format!(
+                                "{PRIME_AGENT_DIR_IN_CONTAINER}/extensions/aoe-session-id.js"
+                            ))
                         ),
                     )
                 }


jerome-benoit (approved):
Reviewed 058d0ad3, including the #3915 discussion, with two independent sequential reviews and cross-validation. No code findings.

The complete extension path is now quoted at src/session/instance/session_id.rs:727-731, using the existing helper (environment.rs:236-239), and still matches the installed path. Bash/dash probes confirmed identical argv before and after, including roots containing spaces, apostrophes and shell metacharacters.

P3, non-blocking description correction: changing the root constant alone would not introduce a quoting bug, because that root was already escaped. Please describe this as making the complete-path quoting boundary explicit, rather than preventing a bug whenever the constant changes. No additional container test is required for this behavior-preserving refactor.

CI, Tests and CodeQL passed for this head; git diff --check was clean. Local validation was limited to source inspection and shell probes, with no Rust build, test-suite or container run.

Issue #3925, opened 2026-09-11
Discuss: dedicated Status::Background variant for background sub-agent state


Istar-Eldritch:
## Context

While fixing #3926  (session status stays "Idle" while a background sub-agent is running — see the linked PR), we ran into a recurring shape of bug: `Status`/`turn_active`-style fields get asked to answer two different questions at once — "is the main turn active" (an action gate: cancel, queue dispatch) and "is there anything worth a lit indicator" (a display signal: busy dot, banner, spinner) — and fixes keep having to re-split them per surface (daemon `AcpState`, TUI `structured_view`, web `AcpRuntime`/`Composer`).

The PR linked below fixes the current instance of this (background sub-agents specifically) without touching the `Status` enum, by adding separate boolean signals (`has_active_background_agent()`, `is_visibly_busy()`) alongside `turn_active`, scoped per consumer.

## The open question

Is it worth adding a dedicated `Status::Background` variant (or equivalent) to the daemon's session status model, so "a background sub-agent is running but the main turn is idle" is a first-class state instead of a derived combination of flags recomputed at each display boundary?

Arguments for:
- One canonical state instead of N ad-hoc `turn_active || has_active_background_agent()` call sites that can drift out of sync (this PR already had to fix that drift twice across TUI/web).
- Matches the existing precedent in `detect/manifests/claude.toml`'s `background_agent_wait` rule (tmux/PTY sessions, #2909), which already treats "background agent running" as tracked state; the ACP/structured path just never got the equivalent.

Arguments against / why we didn't do it in the linked PR:
- Large blast radius: unread-marking, idle-reap timers, TUI status filters, web's `STATUS_DOT_CLASS`, and both wire serialization forms (daemon JSON, ACP broadcast frames) all read/write `Status` today.
- Risk of scope creep turning a bugfix into a session-status redesign.

Not blocking the linked PR — flagging for discussion on whether it's worth doing as a followup, and if so, roughly what shape.

## Related

- PR: (fix/background-agent-status-signal)

njbrake:
Thanks for the suggestion! Would you have cycles to maybe build a prototype and record a video of how you think it should look? I'm not opposed to the idea if we can get the UI UX right

Istar-Eldritch:
Sounds good. I will create a draft PR with the proposal and update this issue with the video once we close #3926 as its a requirement for this work.

Issue #3900, opened 2026-09-10
refactor(tui): make daemon authoritative for terminal session status


jerome-benoit:
## Scope

Implementation slice **I5** of #2734, defined by the [proposed decision record](https://github.com/agent-of-empires/agent-of-empires/issues/2734#issuecomment-5626207757). The record is the canonical design contract; this ticket carries its execution scope and acceptance.

**Status: blocked on the applicable decision ratification and dependencies. No runtime implementation is delivered by filing this ticket. No assignee is inferred.**

## Dependencies and decision gates

#3899

See D1–D3 in the decision record for the applicable ratification boundary.

## Contract and acceptance

Terminal status authority (#3840): needs I4, separately establishes and verifies
freshness contract before local status-writer removal. Test known terminal state
transitions against daemon and both clients, status events/reloads ordering,
reconnect epochs, and persistence with legacy writers. Do not infer SLO from #3841.

## Verification

The acceptance above is future runtime verification, not a passing-test claim. Use the repository’s smallest behavioral tests, real daemon/client smoke scenarios where needed, and relevant build/lint checks. Preserve existing tests of genuine behavior. The source baseline and observed current behavior are linked in the decision record.

## Exclusions

Do not implement unrelated #2735 tiers, revive #1952/#1960, claim CLI is already migrated, or remove crash-atomicity journals covered by #3582. Reuse existing #3835–#3837 work instead of duplicating its endpoints. #3841’s measurement is complete; its real TUI E2E is separate and must pass before row-set cutover.

_AI-assisted drafting: OpenAI Codex GPT-6-Astra in Oh My Pi, with independent source-grounded design reviews. No maintainer ratification is implied._

PR #3926, opened 2026-09-11
fix(acp): keep session status lit while a background sub-agent runs

Istar-Eldritch:
## Description

Session status (web dashboard, TUI, sidebar) showed **Idle** while a Claude background sub-agent (an async `Task` tool call) was still running, because `AcpState.turn_active` and the various display-side "is anything busy" checks only tracked the *main* ACP turn. A background sub-agent starting after the main turn ends (or outliving it) left every busy/working indicator dark even though real work was in flight.

This PR splits the two concerns that had been conflated:
- `turn_active` now tracks **only** the main turn, since it gates real actions (`dispatch::decide`, `session_service::drain_queued_prompts_once`, the TUI's Esc-to-cancel and queue resync, and assistant-ui's own Enter-swallowing `isRunning` check).
- A new, independent `AcpState::has_active_background_agent()` flag (keyed on `ended_at.is_none()`, not `status`, so a `Stalled`-terminal record doesn't latch forever) feeds the **display-only** signals: the TUI's status banner (`render.rs`), the sidebar's `derive_acp_status`, and the web `AcpRuntime`'s message-decoration busy check.

A `StatusIntent::SetUnlessHeld` intent was added so background-agent lifecycle events (Launched/Progress/Completed) can light the busy dot without clobbering a pending approval/elicitation's `Status::Waiting`.

Round 1 of review caught a real regression from an earlier version of this fix: feeding the combined busy signal into assistant-ui's `isRunning` swallowed Enter entirely whenever a background agent ran with the main turn idle, because assistant-ui's own `ComposerPrimitive.Input` hard-disables Enter under `isRunning && !capabilities.queue`, and this adapter declares no queue capability. `isRunning` was reverted to `turnActive` only; `Composer.tsx`'s own `turnActive`-driven enqueue bypass (a pre-existing, deliberate workaround for that same assistant-ui behavior, #1031) still runs unaffected. A Playwright spec now pins this down explicitly (see Test Coverage Analysis).

Scoped out of this PR, filed as a followup discussion instead: whether background-sub-agent state deserves a first-class `Status::Background` variant instead of derived flags recomputed per surface — agent-of-empires/agent-of-empires#3925. Not blocking.

Maintainer review (2026-09-14) surfaced two lifecycle holes that the original scope had excluded as crash-only; both are now fixed in this PR:
- **Stopping a session mid-sub-agent left it Running forever**: worker teardown (`shutdown_with_reason`) now scans the durable event log for launches without a completion and publishes a synthetic `BackgroundAgentCompleted { Detached }` per outstanding agent before its own `Stopped`, so the record closes durably and replay/restart agree. Known residual: crash/respawn paths that never reach worker teardown still leave orphans; a scan-vs-log race shared with `unresolved_approval_nonces`; historical orphans are detached in one burst on the next teardown (no staleness bound); and a malformed-JSON row carrying a matching discriminant would abort the scan the same way it aborts replay (see the `unresolved_background_agent_ids` tests).
- **Cold control cache after a daemon restart**: a mid-turn `BackgroundAgentCompleted` derived Idle from a cold cache and marked an unfinished turn unread. The live listener now hydrates the cache from the event log on a miss, gated to the two events that actually read the activity flags (`Stopped`, `BackgroundAgentCompleted`); boot seeding keeps the documented conservative `(false, false)`.

Still scoped out: Web's `applyReducedState` doesn't heal `background_agents` on a lag gap the way the TUI's equivalent path does; and the `Status::Background` variant design discussion (#3925).

## PR Type

- [x] Bug Fix

## Checklist

## Test Coverage Analysis

Added `web/tests/acp-background-agent-composer.spec.ts`: launches a background agent, ends the main turn, then asserts Enter still POSTs a prompt (pins down the Enter-swallow regression found in round 1 review). Verified it actually catches the regression by temporarily reverting the `isRunning` fix and confirming the new spec fails, then restoring.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the TUI-side behavior (status banner lighting for a background agent with an idle main turn) is pure rendering logic already covered by an in-module unit test (`src/tui/structured_view/render.rs::status_line_shows_working_banner_for_a_background_agent_alone`); no tmux/PTY-level behavior is involved, so the e2e harness wouldn't add signal over the unit test per AGENTS.md's "cheapest test that can catch the regression."

## AI Usage

- [x] AI was used

Implementation and review were done in an AI-assisted, human-directed loop: I set scope and made the split-vs-overload architecture call, reviewed and approved each round's findings before further changes, and made the call on what to scope out vs fix now. Three rounds of fresh-context AI review (correctness/regressions, tests/simplicity, and a final targeted re-check) were run against the diff before this PR; round 2 caught the Enter-swallowing regression described above.

## Summary

- Separates main-turn activity from background-agent activity.
- Keeps TUI and web status indicators accurate while background agents run.
- Allows prompt submission after the main turn ends.
- Preserves approval, elicitation, error, and stopped states during background-agent updates.
- Prevents terminal stalled agents from reopening after late progress events.
- Improves lag recovery and unread status handling.
- Adds Rust, web, reducer, unit, and Playwright regression tests.

## Benefits

Users receive accurate status indicators without losing input or overwriting important states. Turn and unread status remain correct during background-agent activity.

## Further enhancement

Future work could model background-agent status as a first-class state. Recovery could also improve when state updates stop unexpectedly.

### DIFF
--- a/src/server/acp_events.rs
+++ b/src/server/acp_events.rs
@@ -323,7 +323,11 @@ pub(super) async fn acp_event_listener(state: Arc<AppState>) {
             });
         }
 
-        let status_intent = derive_acp_status(frame.event.as_ref());
+        // `publish_persisted` folds this event into the control cache before
+        // broadcasting the frame, so the cached `turn_active` already
+        // reflects it (#3900).
+        let turn_active_after = state.acp_control_cache.turn_active(&frame.session_id);
+        let status_intent = derive_acp_status(frame.event.as_ref(), turn_active_after);
         let acp_change = derive_acp_session_change(frame.event.as_ref());
         let load_session_capability = match (frame.event.as_ref(), frame.worker_generation) {
             (
@@ -513,7 +517,11 @@ pub(crate) async fn seed_acp_statuses(state: Arc<AppState>) {
         let Some(event) = state.acp_event_store.latest_seed_status_event(&id) else {
             continue;
         };
-        let Some(intent) = derive_acp_status(&event) else {
+        // No live control-state fold to consult at boot (the cache is
+        // cold), so a background sub-agent outstanding across a restart
+        // reads as Idle here same as before #3900; the reconciler and the
+        // next live event correct it once the tailer resumes.
+        let Some(intent) = derive_acp_status(&event, false) else {
             continue;
         };
         let mut instances = state.instances.write().await;
@@ -745,7 +753,8 @@ pub(super) async fn recover_structured_unread_after_lag(
         let Some(event) = event_store.latest_seed_status_event(&id) else {
             continue;
         };
-        let Some(intent) = derive_acp_status(&event) else {
+        // Same cold-cache caveat as `seed_acp_statuses` (#3900).
+        let Some(intent) = derive_acp_status(&event, false) else {
             continue;
         };
         // Same snapshot-around-apply as the live path, so "the transition that
@@ -924,7 +933,17 @@ pub(crate) enum StatusIntent {
     HealError,
 }
 
-pub(crate) fn derive_acp_status(event: &crate::acp::Event) -> Option<StatusIntent> {
+/// `turn_active_after` is the session's folded `AcpState::turn_active`
+/// (checked by the caller) right after this event was applied to it.
+/// `AcpState::apply_event` already keeps `turn_active` true across a
+/// `Stopped` while a background sub-agent the turn spawned is still
+/// running, so this is the one place that verdict needs a second input
+/// beyond the event itself: `Stopped` and `BackgroundAgentCompleted` both
+/// resolve to Idle only once it reads false. See #3900.
+pub(crate) fn derive_acp_status(
+    event: &crate::acp::Event,
+    turn_active_after: bool,
+) -> Option<StatusIntent> {
     use crate::acp::Event;
     match event {
         Event::UserPromptSent { .. }
@@ -939,6 +958,13 @@ pub(crate) fn derive_acp_status(event: &crate::acp::Event) -> Option<StatusInten
         Event::ThinkingStarted
         | Event::AgentMessageChunk { .. }
         | Event::ToolCallStarted { .. } => Some(StatusIntent::Set(Status::Running)),
+        // A launched or still-working background sub-agent keeps the sidebar
+        // dot lit even while the main turn is between its own events.
+        Event::BackgroundAgentLaunched { .. } => Some(StatusIntent::Set(Status::Running)),
+        Event::BackgroundAgentProgress {
+            status: crate::acp::state::BackgroundAgentStatus::Running,
+            ..
+        } => Some(StatusIntent::Set(Status::Running)),
         // A pending approval or elicitation both block the turn on the
         // user, so the sidebar dot goes yellow either way.
         Event::ApprovalRequested { .. } | Event::ElicitationRequested { .. } => {
@@ -950,7 +976,23 @@ pub(crate) fn derive_acp_status(event: &crate::acp::Event) -> Option<StatusInten
         // (or for the user to switch to another ACP backend). The
         // dedicated RateLimit banner carries the reset time, so the
         // sidebar pill staying grey is the right signal. See #1281.
-        Event::Stopped { .. } => Some(StatusIntent::Set(Status::Idle)),
+        //
+        // Unless a background sub-agent the main turn spawned is still
+        // running: it keeps working past its parent's `Stopped`, and the
+        // sidebar dot must stay lit until that agent's own
+        // `BackgroundAgentCompleted` clears it (#3900).
+        Event::Stopped { .. } => Some(StatusIntent::Set(if turn_active_after {
+            Status::Running
+        } else {
+            Status::Idle
+        })),
+        // The last outstanding background agent finished. Only drops to Idle
+        // once none remain; a still-running sibling keeps the dot lit.
+        Event::BackgroundAgentCompleted { .. } => Some(StatusIntent::Set(if turn_active_after {
+            Status::Running
+        } else {
+            Status::Idle
+        })),
         Event::AgentStartupError { .. } => Some(StatusIntent::Set(Status::Error)),
         // A successful session/new or session/load means the agent
         // is alive. Heal a sticky Error banner so the sidebar dot
@@ -1824,24 +1866,33 @@ mod tests {
             diffs: Vec::new(),
         };
         assert_eq!(
-            derive_acp_status(&Event::UserPromptSent {
-                prompt_id: None,
-                text: "hi".into(),
-                attachments: Vec::new(),
-            }),
+            derive_acp_status(
+                &Event::UserPromptSent {
+                    prompt_id: None,
+                    text: "hi".into(),
+                    attachments: Vec::new(),
+                },
+                false
+            ),
             Some(StatusIntent::Set(Status::Running))
         );
         assert_eq!(
-            derive_acp_status(&Event::ApprovalRequested {
-                approval: build_approval(tool_call.clone(), Vec::new()),
-            }),
+            derive_acp_status(
+                &Event::ApprovalRequested {
+                    approval: build_approval(tool_call.clone(), Vec::new()),
+                },
+                false
+            ),
             Some(StatusIntent::Set(Status::Waiting))
         );
         assert_eq!(
-            derive_acp_status(&Event::ApprovalResolved {
-                nonce: Nonce("x".into()),
-                decision: ApprovalDecision::Allow,
-            }),
+            derive_acp_status(
+                &Event::ApprovalResolved {
+                    nonce: Nonce("x".into()),
+                    decision: ApprovalDecision::Allow,
+                },
+                false
+            ),
             Some(StatusIntent::Set(Status::Running))
         );
         // A pending elicitation blocks the turn on the user just like an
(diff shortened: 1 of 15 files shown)


src/server/acp_events.rs
```diff
@@ -745,7 +769,8 @@ pub(super) async fn recover_structured_unread_after_lag(
         let Some(event) = event_store.latest_seed_status_event(&id) else {
             continue;
         };
-        let Some(intent) = derive_acp_status(&event) else {
+        // Same cold-cache caveat as `seed_acp_statuses` (#3900).
+        let Some(intent) = derive_acp_status(&event, false, false) else {
```
jerome-benoit: [P2] Preserve live background activity during lag recovery

This is a live-daemon recovery path, not cold boot. After `UserPromptSent -> BackgroundAgentLaunched -> Stopped`, the normal listener correctly keeps Running. If the global broadcast subsequently lags, recovery visits all structured sessions; `latest_seed_status_event` excludes background events (`event_store.rs:1174-1187`) and returns Stopped. These hardcoded false values then force Idle and can persist a premature unread mark at line 788. A quiet tailer emits no immediate corrective progress event (`background_agent.rs:343-360`).

Use authoritative current folded activity during live recovery, and add a recovery-level regression asserting Running and no unread mark for that sequence. Verified by tracing these functions with `git show 82a55bee:<path>`; this is not a claim of an executed runtime reproduction.

src/server/acp_events.rs
```diff
@@ -564,6 +579,15 @@ pub(crate) fn apply_status_intent(
             }
             s
         }
+        // A background sub-agent's own progress does not speak to whatever
+        // the main turn is blocked on; leave a pending approval/elicitation
+        // showing Waiting until the main turn's own event resolves it.
+        StatusIntent::SetUnlessWaiting(s) => {
+            if matches!(inst.status, Status::Stopped | Status::Waiting) {
```
jerome-benoit: [P2] Do not let background updates clear a main-agent error

Please also preserve Error here. The background tailer owns a cloned event sender (`acp/acp_client/connection.rs:666-671`) and an independent task. An established connection failure emits AgentStartupError (`connection.rs:3047`), but the supervisor continues draining tailer events before evaluating respawn (`supervisor.rs:2272-2332`). A subsequent background Progress or Completed therefore replaces the main connection's Error with Running or Idle through this new arm, without the main connection recovering. Previously those background events did not update sidebar status.

Preserve Error for these background-only intents and cover Error followed by progress/completion. This does not require changing genuine main-agent recovery events. Verified through exact-head source tracing with `git show 82a55bee:<path>`; no runtime reproduction was executed.

jerome-benoit (changes requested):
Reviewed 82a55be after the author response and CodeRabbit fixes. The mid-turn completion and terminal-stalled regressions are addressed. Two remaining P2 status regressions need correction below; neither requires the broader Background-status redesign in #3925 or the deferred web snapshot work.

Cross-validated through two independent reviews and source tracing. No local test suite was run. CI authorized for this head: format, Clippy, web checks and Playwright have passed; remaining Rust/CodeQL jobs were still running at review time.

src/server/acp_events.rs
```diff
@@ -564,6 +580,22 @@ pub(crate) fn apply_status_intent(
             }
             s
         }
+        // A background sub-agent's own progress does not speak to whatever
+        // the main turn is blocked on; leave a pending approval/elicitation
+        // showing Waiting until the main turn's own event resolves it. The
+        // same holds for Error: the tailer runs on a cloned sender and keeps
+        // draining while the supervisor evaluates a respawn, so its events
+        // must not clear the main connection's error banner; only HealError
```
jerome-benoit: [P3] Limit this explanation to background events

Update at 268970c4: the production explanation is corrected. The remaining occurrence is the test comment at src/server/acp_events.rs:2551-2552, which still says only HealError resolves Error. Genuine main-agent Set events can also clear Error. Please remove that absolute claim from the test comment too.

Documentation-only and non-blocking; no runtime change requested. Verified with git show 268970c4:src/server/acp_events.rs. This updates the existing finding rather than adding a duplicate.

jerome-benoit (approved):
Re-reviewed 82a55bee..e191d502, including your response, with two independent sequential analyses and final arbitration. Both P2s are addressed; no remaining blocker.

Background intents now preserve Error (acp_events.rs:590-595). Live lag recovery receives the control cache and uses its activity flags (:789-792); the new regression invokes actual recovery over the prompt/launch/stop history and checks Running with no unread mark. The existing no-background recovery case remains covered. Cold-boot behavior and the separately scoped design work need not change here.

One non-blocking P3 wording correction is inline. Evidence: exact-head git show/diff and clean git diff --check. No local tests or runtime reproduction performed. CI workflow has passed; Tests and CodeQL are still running, so this approval does not claim full CI success. Both independent analyses agree on the fixes.

jerome-benoit (approved):
Re-reviewed e191d502..268970c4, including your response, with two independent sequential analyses and final arbitration. No new blocking finding; approval renewed for this head.

The Stopped arm now preserves a newer active turn (acp_events.rs:1033-1038). The added recovery test covers the stale seed followed by UserDiffCommentsPrompt and asserts Running without unread (:1567-1675). The reducer and prompt-dispatch hydration support that scenario; cold-boot behavior remains unchanged.

The existing P3 is partially addressed: production wording is corrected, but the test comment still says only HealError resolves Error (:2551-2552). This remains documentation-only and non-blocking; no duplicate finding is added.

Evidence: exact-head git show/diff and clean git diff --check. No local tests or runtime reproduction executed. CI workflow passed; Tests and CodeQL remain in progress. Both independent analyses agree on the code change; final source arbitration retains the residual wording correction.

njbrake (changes requested):
**Request changes: stopping a session mid sub-agent leaves it showing Running permanently.**

`POST /api/sessions/{id}/stop` calls `shutdown`, which aborts the drain task (`src/acp/supervisor.rs:3137`) and drops the event receiver. The tailer then exits via `event_tx.closed()` without emitting a terminal event. Only `BackgroundAgentCompleted` sets `ended_at` (`src/acp/state.rs:1639`), and `SessionCleared`/`AgentSwitched` leave `background_agents` untouched. From then on every `Stopped` derives Running, the case `acp_events.rs:2231` asserts: no Idle, no unread mark, and the TUI and web busy indicators stay lit. The cache replays the event log, so this survives daemon restarts. A normal Stop triggers it, not just a crash. Suggested fix: on worker teardown, emit `BackgroundAgentCompleted { status: Detached }` for outstanding agents, and add a test covering Stop → prompt → `Stopped`.

**Cold cache regresses mid-turn completion.** The cache hydrates only through `fold_control_state` on dispatch (`src/server/session_service.rs:1288, 1577`), and a miss reads `false`. After a daemon restart with a reattached worker, a mid-turn `BackgroundAgentCompleted` now derives Idle and marks unread; base returned `None`. Hydrate on a miss, or return `None` when cold.

**Wrong issue references.** 30 added comments cite #3900, which is the TUI terminal-status refactor. #3925 is the `Status::Background` discussion but is cited on fixes.

**Nits:**
- Em dashes at `src/acp/state.rs:1881` and `src/server/acp_events.rs:2693`.
- `backgroundAgentCompleted` (`web/tests/helpers/acpMock.ts:397`) is unused; `AcpState::is_visibly_busy` has no production caller.
- `SetUnlessWaiting` also preserves `Stopped` and `Error`; the name should say so.
- `src/tui/structured_view/reducer.rs:79-81` says to combine the flag for Esc-to-cancel; `mod.rs:692-697` says not to.
- The test comment at `src/server/acp_events.rs:2551` contradicts `apply_status_intent`.
- The `turnActive` parameter of `activityToThreadMessages` now receives `visiblyBusy`.

jerome-benoit (changes requested):
Reassessed `268970c446839a92018eaa346dc20a04c4c068c6` against the maintainer's [review](https://github.com/agent-of-empires/agent-of-empires/pull/3926#pullrequestreview-5197873878). This supersedes my earlier approval: I missed two P2 correctness blockers already described there, so I am not duplicating inline findings.

- **Outstanding background work survives worker teardown.** `shutdown_with_reason` aborts the drain and publishes `Stopped` (`supervisor.rs:3137-3155`), without terminalizing the background records. The activity predicate still sees their missing `ended_at` (`state.rs:1177-1179,1639`), and later `Stopped` events derive Running (`acp_events.rs:1033-1038`). One qualification: the stop endpoint first sets `Instance.status=Stopped`, which `apply_status_intent` preserves (`lifecycle.rs:1082`; `acp_events.rs:578`). The sidebar need not become Running immediately on Stop; the durable stale activity persists and affects the resumed session. Please cover Stop → resume/prompt → completion and replay. Close outstanding activity durably at actual worker teardown, not on ordinary turn-end `Stopped` or daemon detachment that keeps the agent alive.
- **Cold-cache completion can falsely report Idle/unread.** `apply_if_cached` ignores an unhydrated slot, and both activity readers return false (`control_cache.rs:80-85,112-128`). `BackgroundAgentCompleted` consequently derives Idle (`acp_events.rs:1047-1053`) even if the reattached main turn is still active. Preserve unknown versus idle, hydrating authoritative state or declining the background-only status transition when unknown. Cover an unhydrated, reattached session before any new prompt dispatch, not only warm-cache sequences.

The description's “crash-only” exclusion is therefore too narrow. The maintainer's terminology/reference/dead-helper observations and my existing P3 remain non-blocking cleanup; the request is to fix these two lifecycle/status behaviors, not introduce `Status::Background` or redesign the feature.

Evidence: exact-SHA `git show`/`git grep`, baseline comparison, and two independent sequential analyses with parent arbitration. Current CI is green; no local runtime reproduction or test suite was executed during this reassessment. Green CI does not cover the missing lifecycle scenarios above.

jerome-benoit (dismissed):
Re-reviewed `268970c4..3a4c7301`; both of my CHANGES_REQUESTED blockers are resolved, approving (the PR's overall decision stays CHANGES_REQUESTED until @njbrake re-evaluates his own review).
- Outstanding background work at teardown: `shutdown_with_reason` now, inside the `StopDecision::TearDown` arm only, scans the durable log via `unresolved_background_agent_ids` and publishes a `BackgroundAgentCompleted{Detached}` per orphan before its own `Stopped` (supervisor.rs:3189-3212; event_store.rs:1594+). `state.rs:1603` guards on `ended_at.is_none() && !matches!(Completed|Detached|Error)`, so Detached closes the record and resists a late Progress. It fires only at real teardown (no-op on clean teardown / NotOwned / CancelRequested), with regression tests (shutdown detaches before Stopped; derive_acp_status resolves Idle once teardown detaches the last agent).
- Cold-cache completion: the live listener hydrates a cold cache via `fold_control_state` before reading activity flags, gated on `reads_activity_flags` (= Stopped|BackgroundAgentCompleted) with a coupling property test so a new flag-reading arm cannot skip hydration (acp_events.rs:335-341, 998-1002; `is_hydrated`). Regression tests reattach a mid-turn completion off a cold cache and assert Running with no unread.
- P3 wording corrected (main-turn Set and HealError resolve Error; only SetUnlessWaiting preserves it).
Evidence: exact-SHA git show of the PR commits (2f7ffc8d, b1717b00, c81d00c2, 6e72a113, f36022b4) and the two consumer dispatch points; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI is pending/not yet run on this head; this approval concerns the code review only.

jerome-benoit (dismissed):
Re-reviewed `3a4c7301..f6f4fb6d` (commit f6f4fb6d); re-approving (the PR's overall decision stays CHANGES_REQUESTED until @njbrake re-evaluates his own review).
The detach scan now survives a corrupt/unextractable `BackgroundAgentCompleted` row and is proven through the real SQL path. The load-bearing fix is the subquery guard `AND json_extract(event_json,'$.BackgroundAgentCompleted.agent_id') IS NOT NULL` (event_store.rs:1624): SQLite `X NOT IN (subquery)` collapses to NULL (never true) for every outer row once the subquery yields one NULL, so a single `{"BackgroundAgentCompleted":{}}` row previously blanked the whole orphan scan and silently no-opped the detach. I reproduced this deterministically (a NULL in the NOT IN subquery drops the real orphan; the guard restores it). The launched-side guard (:1618) is defensive/harmless (a NULL id is dropped by `row.get::<String>` + filter_map anyway).
Verified/rejected: the new `unresolved_background_agent_ids_survives_an_unextractable_completed_row` (event_store.rs:4599) is a genuine oracle (fails pre-guard, passes post; own tmp DB, no env mutation). The new `shutdown_detaches_through_a_real_channel_sink_and_event_store` (supervisor.rs:6083) genuinely closes the VecSink gap (VecSink returns a canned list, never touching the real json path/session key/SQL); it wires a real disk EventStore + ChannelSink, and `hydrate_seqs([(s,1)])` is essential to avoid the Detached completion colliding with the pre-recorded launch at seq 1 under INSERT OR IGNORE (Detached=2, Stopped=3). Deterministic/isolated (`#[serial_test::serial]` + TempDir; the `unsafe set_var(HOME/XDG)` matches the sibling teardown test's convention). `dummy_runner_kind` is a behavior-preserving field-for-field extraction; the `id`->`idx` rename is cosmetic. `src/server/acp_events.rs` is strictly doc/comment-only (the `reads_activity_flags` body is unchanged). No regression against my prior approval (teardown terminalization, detach scan, cold-cache direct-log hydration all unchanged).
Non-blocking (pre-existing, out of scope): the sibling `has_in_flight_turn` background-agent subquery guards `$.BackgroundAgentCompleted IS NOT NULL` (object presence) rather than `.agent_id IS NOT NULL`, so it keeps the same latent NULL-in-subquery pattern this query just fixed; worth a follow-up but not introduced by this delta.
Evidence: exact-SHA source tracing of the delta and the real files at f6f4fb6d (file:line above); a deterministic SQLite repro of the NOT-IN/NULL three-valued semantics; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI is green on this head per `gh pr checks`; this approval concerns the code review only.

jerome-benoit (dismissed):
Re-reviewed `f6f4fb6d..41833ef8` (commit 41833ef8, test-only); re-approving (the PR's overall decision stays CHANGES_REQUESTED until @njbrake re-evaluates his own review).
This closes a real env leak: the previous commit's raw `unsafe { set_var("HOME"...); set_var("XDG_CONFIG_HOME"...) }` blocks did not restore, leaking HOME/XDG into later serial tests. They are replaced by RAII guards from `crate::session::test_support`: `isolate_home` (supervisor.rs, 5 sites) sets HOME + XDG_CONFIG_HOME identically to the removed pair and restores on Drop; `isolate_app_dir_at` (acp_events.rs, 4 sites) is a strict superset (HOME + XDG_CONFIG_HOME + XDG_DATA_HOME on linux/macos). Both are bound to named `_home`/`_app_dir` (the guards are `#[must_use]`; a `let _ =` would drop on the same line and run the body against the real env), so each guard holds `ENV_LOCK` and its overrides to end of scope.
Verified/rejected: restore-on-Drop is real (`EnvGuard::drop` -> `restore_or_remove`, `set_var(prev)` or `remove_var` if previously unset, panic-safe under ENV_LOCK); `AppDirGuard` sequences env-restore before tempdir deletion and does not own the caller tempdir. The added precondition `assert_eq!(event_store.unresolved_background_agent_ids("s-real-teardown"), vec!["bg-real"])` before teardown is non-tautological and deterministic (a single synchronously-seeded launch, no completion yet), and makes the later detach-completion assertions non-vacuous. No production logic changed (all hunks inside `mod tests`). No regression vs my prior approval; determinism/isolation preserved (`#[serial_test::serial]` + tempdir).
Evidence: exact-SHA source tracing of the delta and the real helper bodies in src/session/test_support.rs (file:line); two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. On this head only `Lint PR title` had run (passed); I authorized the gated CI/Tests/CodeQL workflows and they are now running (not yet green). This approval concerns the code review only.

Istar-Eldritch:
Sorry I just realised I left the agent doing the first pass  of fixes and pushed and created an issue on this repo without my explicit consent. I will be doing a review myself before asking you to review this.

PLEASE DO NOT REVIEW THE NEW CHANGES YET

I will do a manual check before I make you spend time on this.

jerome-benoit (approved):
Re-reviewed `41833ef8..ad7887b5` (2 commits); re-approving (the PR's overall decision stays CHANGES_REQUESTED until @njbrake re-evaluates his own review).
The delta is a pure refactor + cosmetic rename + test-isolation, with no observable behavior change. `is_visibly_busy()` was dead production code (0 callers tree-wide, verified at head; the real TUI display busy signal in `structured_view/render.rs` already combines `turn_active || has_active_background_agent()` directly and is unchanged), so its removal loses no display path. `StatusIntent::SetUnlessWaiting` -> `SetUnlessHeld` is a complete tree-wide rename (0 orphan references at head) with a byte-identical guard (`matches!(status, Stopped | Waiting | Error)`), so it is purely nominal. The 4 submodule tests in `git/worktree.rs` now take `#[serial_test::serial]` + `isolate_home` (RAII EnvGuard), correctly isolating them from an ambient `core.excludesFile` that could ignore the `.claude` fixture.
Verified/rejected: the `state.rs` test assertions swapped `is_visibly_busy()` -> `has_active_background_agent()` are equivalent in scenario (each asserts `turn_active == false` immediately before, so the disjunction reduces to `has_active_background_agent()`); no coverage loss. No production logic changed outside the rename. No regression vs my prior approval.
Non-blocking (P3): (1) the migration left the `Stopped`-arm doc comment in `acp/state.rs` (~:1381) mangled (a dangling "Use" and an orphan "// signal." line); collapse to one sentence. (2) the three `set_unless_waiting_*` unit tests (`acp_events.rs` ~:3017/:3036/:3049) keep the old name while exercising `SetUnlessHeld`; rename for consistency.
Evidence: exact-SHA source tracing of the delta at the pinned head (fetched git blobs; tree-wide grep for orphan references); two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI is green on this head per `gh pr checks` (incl. the Cargo Test shards that exercise the newly-serialized submodule tests); this approval concerns the code review only.

PR #3929, opened 2026-09-12
fix(fork): decide fork eligibility on the row and the user's own config

blueagledev:
## Description

`aoe add --fork-from <parent> --tool <wrapper>` is refused whenever the
wrapper's name differs from the parent row's stored tool, even when
`session.agent_detect_as` declares the wrapper to be that same built-in:

    [session.custom_agents]
    "claude-work" = "claude-work"
    [session.agent_detect_as]
    "claude-work" = "claude"

    $ aoe add . --fork-from parent --tool claude-work
    Error: Cannot fork session 'parent' (agent 'claude') as agent
    'claude-work': a fork must use the parent's agent.

The same fork spelled as a command is allowed, because `--cmd` resolves
through `detect_tool` and `agents::resolve_tool_name` (src/agents.rs:1940),
a substring match that collapses `claude-work` to `claude`.

Two comparisons in `add` key on the raw name: the gate (src/cli/add.rs:392)
and the tool handed to `terminal_fork_seed` (src/cli/add.rs:406), which looks
that name up with `get_agent` and so answers `AgentCannotFork` for every
custom agent. What runs afterwards already resolves the alias: the launch path
picks the fork strategy through `Instance::resolved_agent`
(src/session/instance/session_id.rs:1277-1279), and
`launch_can_carry_resume_selector` (src/session/instance/accessors.rs:289-302)
states the rule this commit applies to the gate, in its own words:
`agent_detect_as` is the user declaring what a wrapper wraps (#3638).

Both sides of that comparison are inputs an attacker may reach, so both are
narrowed rather than resolved the obvious way.

The parent side answers from what its row records: its tool when that names a
built-in, else the alias the row itself carries, and `None` otherwise.
`Instance::resolved_agent` is not used, because `effective_detect_as`
(src/tmux/status_rules.rs:220-231) returns the stored alias **only when it is
non-empty** and otherwise consults the live map, and rows with an empty one
keep being written (`add.rs:681-686` stores `unwrap_or_default()` for a tool
with no entry). A row that records nothing is evidence of nothing.

The chosen side is resolved from the **profile** config, not the repo-merged
one `add` otherwise uses. `agent_detect_as` is `repo = "allow"`
(settings_schema/registry.rs:231) in a section whose `repo_default` is `deny`,
and that exception was granted while the field only picked status heuristics.
Deciding a fork on it lets a checked-out `.agent-of-empires/config.toml`
authorize handing one agent's session id to another, with the user's own
config saying the opposite and nothing of theirs edited.

Resolving only the authorizing side from profile config is not enough, and the
test `fork_gate_refuses_when_a_repo_repoints_the_chosen_tool` pins why: the row
this call creates stores the **merged** alias (add.rs:681-686) and launches
under it, so a repo that re-points a tool the user did name would be authorized
against one agent and run as another. `fork_agent_for` therefore refuses
outright when the two layers disagree about the chosen tool.

The alternative was to mark the field `repo = "deny"`. That is the wider fix
and it would close the same hole for `resolved_agent`'s other consumers, but it
reverses a deliberate exception (#3154), changes behaviour for repos that use
it for status detection today, and rewrites the repo-override tests that assert
it merges. It belongs in its own change, with its own argument; this commit
stops fork authorization from depending on it and says so rather than quietly
relying on the exception standing.

Measured at c008671d with three debug binaries, distinct md5s. `nm -C` finds
`fork_agent_for` and `pinned_capture_agent` in both patched binaries and in
neither stock one, and `terminal_fork_seed` in all three as a control that
`nm` reads symbols at all; that check separates patched from stock, not the
two patched builds from each other, for which the md5s and the behaviour below
are the discriminator. Isolated XDG_CONFIG_HOME and tmux socket, parent row
carrying a Claude UUID, wrappers on fake binaries that record argv. Profile
config says `claude-work = claude` and `codex-work = codex` throughout.

    repo file declares codex-work = claude; fork as codex-work
      stock    refused
      earlier  created; ARGV[codex-work]: --resume 11111111-... --fork-session
      this     refused
    no repo file; fork as claude-work                        [control]
      stock    refused
      earlier  created and forks
      this     created and forks

The control differs from the first arm only in the repo file, so the file is
the whole difference. The other direction (a repo re-pointing a tool the user
did name) is refused by every build here and is pinned by a test rather than
by this table: the design it discriminates against is one no binary ships.

Seven tests drive `add`'s gate and seed through `run()`. Resolving the chosen
side from the merged config fails the repo test; dropping the layer-agreement
check fails the other repo test; reverting both call sites to the raw name
fails the two that expect a fork; resolving the parent through
`Instance::capture_agent_name` and the seed through the live map fails the two
that expect a refusal; dropping the `resolved_tool != source.tool` clause fails
the same-name test. Each mutation fails one test and disturbs no other.

Not covered. The alias a row carries is not pinned at capture: it is written at
session build and can be rewritten later by the `v024` backfill from whatever
the map says then. Measured: a row with an empty alias and `.schema_version`
below 24 has it rewritten by a read-only `aoe list`. So a map retargeted before
that backfill records the wrong agent, and this gate believes it. The backfill
is one-shot per app dir, which leaves the window open for a restored backup or
a copied profile directory. Closing it needs provenance the row does not have.

A row that records no alias is refused even where the live map agrees, which is
stock's behaviour. The two spellings still diverge for a `custom_agents`
wrapper with no `agent_detect_as` entry at all. Two names sharing a base need
not share the agent's conversation store. A wrapper whose configured command
carries flags cannot carry a resume selector (accessors.rs:303-323), so a fork
intent is stored and dropped at launch with no message; `--cmd` reaches that on
stock today and it is reported separately. The TUI
(src/tui/home/input.rs:3056, 3269) and the server create path
(src/server/api/sessions/create.rs:163) decide eligibility from the raw name
and are untouched; the server path performs no same-agent check at all.

**Related issues:** none that I could find. I searched this repo's issues and PRs
for `fork agent_detect_as`, `fork must use the parent` and `agent_detect_as` and
got nothing covering this; #3154 and #3638 are cited above as the prior context
this reasons from, not as issues it closes.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

**Documentation is deliberately left unticked, and this is the one thing in this
PR I would like a maintainer's call on.** `docs/guides/session-fork.md` states
the rule this commit relaxes:

> If you do pass one, it must match the parent's agent: a captured conversation
> is agent-specific, so forking a Claude session as Codex is rejected rather
> than run against the wrong agent.

After this change a wrapper the user mapped to the parent's built-in with
`session.agent_detect_as` is accepted, so that sentence is no longer the whole
rule. I have not edited it here because this PR's commit is reviewed and I did
not want to move it; say the word and I will add the doc change to this PR or
open it separately.

```

fails in all three runs

```

```
this commit e9c95717, cargo test --lib, three runs
  run1  6961 passed; 4 failed; 2 ignored   (6967 tests)
  run2  6960 passed; 5 failed; 2 ignored   (6967 tests)
  run3  6960 passed; 5 failed; 2 ignored   (6967 tests)

    base 0/3, here 1/3   server::startup::tests::shutdown_deadline_runs_from_the_cancel_not_the_watchdog_poll
    base 1/3, here 1/3   session::projects::tests::set_pinned_toggles_without_removing_entry
    base 1/3, here 0/3   session::projects::tests::update_base_branch_sets_clears_and_reports_not_found

  this commit's own tests matching cli::add::tests::fork_gate::*:
    passed [7, 7, 7] across the three runs, failed [0, 0, 0] -- checked by name, not by count

```

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: **no test was added under `tests/e2e/`; the regression coverage landed as in-crate tests instead, and I would rather say so than tick the box above.** This does change a CLI subcommand, so the N/A box is not true either.

Eight tests were added in `src/cli/add.rs`. Seven of them drive `add`'s gate and
fork seed through `run()` against a real parent row and a real config, and what
they discriminate is **which refusal you get**, not whether a fork happens: the
fixture parent's `agent_session_id` is null (`add.rs:1969`), so `fork_as` asserts
an error in every case and none of the seven reaches session creation or argv
capture. The argv-recording fake binaries are the separate manual probe against
three built binaries described in the AI section below, not these tests.

That distinction is corrected here — an earlier version of this paragraph
attributed the argv capture to these tests, which was wrong, and the review was
right to ask for the two to be separated. The tests:

    fork_gate_reads_the_row_and_the_configured_alias
    fork_gate_refuses_a_parent_row_with_no_pinned_alias
    fork_seed_without_tool_ignores_the_rows_pinned_alias
    fork_gate_resolves_the_chosen_tool_from_the_users_own_config
    fork_gate_refuses_when_a_repo_repoints_the_chosen_tool
    fork_gate_leaves_a_same_name_fork_to_the_seed
    fork_seed_without_tool_resolves_a_pinned_alias_that_still_agrees
    capture_agent_for_tool_resolves_detect_as_wrappers

They were checked for discrimination rather than only for passing: each of six
mutations of the fix fails exactly one of them and disturbs no other — resolving
the chosen side from the merged config fails
`fork_gate_resolves_the_chosen_tool_from_the_users_own_config`, dropping the
layer-agreement check fails `fork_gate_refuses_when_a_repo_repoints_the_chosen_tool`,
reverting both call sites to the raw name fails the two that expect a fork,
resolving the parent through `Instance::capture_agent_name` fails the two that
expect a refusal, and dropping the `resolved_tool != source.tool` clause fails
`fork_gate_leaves_a_same_name_fork_to_the_seed`.

`AGENTS.md` asks for the cheapest test that can catch the regression, prefers
Rust unit tests, and reserves full-binary tests for behaviour that needs that
environment, so I read in-crate tests as following that rule rather than dodging
it — but the template's box asks specifically about `tests/e2e/`, so it stays
unticked and this is the reason.

If you would rather have it pinned under `tests/e2e/` too, say so and I will add it.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

not see the earlier ones. Each round found a real defect the previous fix had not
closed, and the description's account of that is not tidied up: rounds 1-3 each
fixed the comparison and left a different input open.

The behavioural claims were driven rather than argued. Three debug binaries with
distinct md5s — stock, the round-3 build and this one — run against fixtures with
their own `XDG_CONFIG_HOME` and tmux socket and fake agent binaries that record
their argv, with `nm -C` used to confirm which build is which and one symbol
present in all three as a control that `nm` reads symbols at all. The two arms in
the table differ only in the repo config file. The tests were then checked for
discrimination by mutating the fix six ways, not just for passing.

---

### Unrelated to this patch: a defect in this repo's own PR template check

I hit this while restoring this PR's body, so it is here rather than in an issue.
It has nothing to do with the fork gate, and if you would rather have it as its
own issue, say so and I will move it — that is probably where it belongs.

`.github/workflows/pr-template-check.yml` reports `check-template: success` on a
PR whose body has no checklist at all. On the `!hasChecklist` path it lists the
PR's labels and, when `missing-template` is already there:

```js
console.log('PR already labeled, skipping comment');
return;
```

That `return` comes before `core.setFailed`, so **only the first run of the check
can ever fail.** The `opened` run fails and applies the label; every later
`edited` run takes the early return and reports green. It does not need the
author to do anything either — a bot appending release notes to the body fires
`edited` on its own, which is what happened here.

Both runs on this PR, against the same unchanged body:

```
check-template  failure  job 103594039151  started 2026-09-12T17:40:49Z
check-template  success  job 103594191632  started 2026-09-12T17:42:01Z
```

and the successful job's own log says why it was green:

```
2026-09-12T17:42:03.8656797Z PR already labeled, skipping comment
```

The label is not removed on that path, and `close-stale` keys on the label alone,
so the PR still gets closed 24 hours later while its check reads green. That is
the harmful half: **`gh pr checks` cannot be used to confirm a template fix**, and
the `missing-template` label is the only signal that reflects what the closer will
do. Moving the `core.setFailed` call above the early `return` would be enough.

## Summary

- Updated `aoe add --fork-from ... --tool ...` to compare resolved agents instead of raw tool names.
- Added support for custom-agent wrappers and tool aliases that resolve to the parent’s built-in agent.
- Forks now use the parent row’s recorded tool metadata and the selected tool from the user’s profile configuration.
- Forks are rejected when agent metadata is missing or when profile and repository configuration resolve the tool differently.
- Updated fork seeding to use the resolved agent.
- Added tests for fork eligibility, aliases, missing metadata, wrapper resolution, and configuration differences.
- TUI and server-session fork behavior remain unchanged.

### DIFF
--- a/src/cli/add.rs
+++ b/src/cli/add.rs
@@ -384,12 +384,27 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
         }
         // The child must fork the SAME agent as the parent: a captured id is
         // agent-shaped (a Claude UUID resumes only under Claude, etc.), so
-        // handing it to a different agent's `--resume` fails or resumes garbage.
-        // When the user did not explicitly choose a tool (`--tool`/`--cmd`),
-        // inherit the parent's; when they did and it differs, reject rather than
-        // launch a cross-agent fork.
+        // handing it to a different agent's `--resume` fails or resumes
+        // garbage. Compared on the built-in each name resolves to, so an
+        // `agent_detect_as` wrapper and its base are one agent. Without an
+        // explicit `--tool`/`--cmd` the parent's tool is inherited instead.
         let user_chose_tool = args.tool.is_some() || args.command.is_some();
-        if user_chose_tool && resolved_tool != source.tool {
+        let parent_agent = pinned_capture_agent(source);
+        // Authorization may not read the checked-out repo. `agent_detect_as`
+        // is `repo = "allow"`, so `.agent-of-empires/config.toml` can re-point
+        // an alias; the profile-only config is the user's own answer.
+        let profile_config =
+            crate::session::config::profile_config::resolve_config_or_warn(profile);
+        if user_chose_tool
+            && resolved_tool != source.tool
+            && fork_agent_for(
+                &resolved_tool,
+                parent_agent,
+                &profile_config.session,
+                &config.session,
+            )
+            .is_none()
+        {
             bail!(
                 "Cannot fork session '{}' (agent '{}') as agent '{}': a fork must use the parent's \
                  agent. Drop --tool/--cmd to inherit it, or fork a session created with '{}'.",
@@ -403,8 +418,18 @@ pub async fn run(profile: &str, args: AddArgs) -> Result<()> {
             resolved_tool = source.tool.clone();
         }
         let parent_agent_session_id = source.agent_session_id.clone();
-        let seed = crate::session::fork::terminal_fork_seed(
+        // The alias is resolved for the seed only where the row itself proves
+        // the child runs the parent's built-in. Without that proof the raw
+        // name goes through, which is the refusal this path had before.
+        let fork_agent = fork_agent_for(
             &resolved_tool,
+            parent_agent,
+            &profile_config.session,
+            &config.session,
+        )
+        .unwrap_or(resolved_tool.as_str());
+        let seed = crate::session::fork::terminal_fork_seed(
+            fork_agent,
             parent_agent_session_id.as_deref(),
             crate::session::capture::generate_session_uuid(),
         )
@@ -1551,6 +1576,75 @@ fn override_launch_binary(
     shell_words::split(&command).ok()?.into_iter().next()
 }
 
+/// The built-in a tool name being chosen now resolves to: the built-in itself,
+/// or the base a `session.agent_detect_as` wrapper declares. Reads the live
+/// map, which is also what the new session's own `detect_as` is written from.
+fn capture_agent_for_tool(
+    tool: &str,
+    session: &crate::session::config::SessionConfig,
+) -> Option<&'static str> {
+    crate::agents::get_agent(tool)
+        .or_else(|| {
+            session
+                .agent_detect_as
+                .get(tool)
+                .and_then(|base| crate::agents::get_agent(base))
+        })
+        .map(|agent| agent.name)
+}
+
+/// The built-in that captured a parent row's session id, from the row and the
+/// agent registry alone: its tool when that names a built-in (the registry
+/// wins there, as it does everywhere), else the `agent_detect_as` alias the
+/// row recorded.
+///
+/// Deliberately narrower than `Instance::resolved_agent`, which falls back to
+/// the live alias map when the stored value is empty: `effective_detect_as`
+/// returns the stored alias only when it is non-empty, and rows with an empty
+/// one keep being written (`add.rs` stores `unwrap_or_default()` for a tool
+/// with no entry). Such a row records no evidence of what shaped its id, and
+/// the live map may have been retargeted since, so `None` is the honest
+/// answer. The recorded alias is only as good as the map when it was written,
+/// which is session build or the `v024` backfill, not the capture itself.
+fn pinned_capture_agent(parent: &Instance) -> Option<&'static str> {
+    crate::agents::get_agent(&parent.tool)
+        .or_else(|| {
+            let pinned = parent.detect_as.trim();
+            (!pinned.is_empty())
+                .then(|| crate::agents::get_agent(pinned))
+                .flatten()
+        })
+        .map(|agent| agent.name)
+}
+
+/// The built-in a fork would run under, when the parent row proves the child
+/// resolves to that same one. `None` whenever the two sides cannot be shown to
+/// match, including when the row recorded nothing: the caller then falls back
+/// to the raw name, which is what this path did before the alias was resolved
+/// at all. So the permitted set only ever grows where the row itself is
+/// evidence.
+///
+/// `chosen` is resolved from `profile`, the user's own config, and the fork is
+/// refused outright when `merged` (the same config with the project repo's
+/// `.agent-of-empires/config.toml` layered on) resolves it differently. A
+/// repo-supplied `agent_detect_as` entry may steer status detection, which is
+/// what it was marked `repo = "allow"` for; it may not decide whose session id
+/// a fork hands to which agent, and it cannot be read only on the authorizing
+/// side because the row this call creates stores the merged alias and launches
+/// under it.
+fn fork_agent_for(
+    chosen: &str,
+    parent: Option<&'static str>,
+    profile: &crate::session::config::SessionConfig,
+    merged: &crate::session::config::SessionConfig,
+) -> Option<&'static str> {
+    let chosen_agent = capture_agent_for_tool(chosen, profile)?;
+    if capture_agent_for_tool(chosen, merged) != Some(chosen_agent) {
+        return None;
+    }
+    (Some(chosen_agent) == parent).then_some(chosen_agent)
+}
+
 enum NamedToolSelection {
     Custom(String),
     BuiltIn(String),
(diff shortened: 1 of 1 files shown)


jerome-benoit (changes requested):
Reviewed `e9c95717` against `bdf21c25`, including the description and existing discussion, with two independent sequential analyses and source arbitration.

**P2, blocking:** please address the existing [custom-agent shadowing finding](https://github.com/agent-of-empires/agent-of-empires/pull/3929#discussion_r3997043180). A Claude-pinned wrapper parent can newly fork as a custom `claude = "codex"`: authorization identifies Claude at `add.rs:1586`, command selection uses the custom command at `:743-745`, and launch still resolves the built-in at `accessors.rs:529`. This passes Claude fork flags to Codex (`session_id.rs:1273-1292`). Reject inconsistent shadowing or make authorization agree with actual launch semantics; merely changing helper precedence can leave the latter mismatch. Please cover that scenario without duplicating the existing thread.

Non-blocking: update the CLI paragraph in `docs/guides/session-fork.md` in this PR to document the accepted wrapper form. Also distinguish the seven gate tests from external native probes in the description: their parent ID is null (`add.rs:1969`), so they stop before creation/argv capture. In-crate tests are appropriate; no duplicate e2e suite is requested.

The broader v024 provenance and other UI/server fork paths are not merge conditions here. Evidence: exact-head git diff/show, clean diff --check; no local builds/tests or author-reported probes rerun. CI was authorized for this SHA and is running.

PR #3930, opened 2026-09-12
fix(hooks): report read-only agent settings once, not on every launch

blueagledev:
## Description

Where an agent's `settings.json` is managed outside the agent (a Nix
store symlink, a read-only mount, a managed image), AoE cannot install
its status hooks and says so for every session it touches:

    WARN session.store: Failed to install agent hooks: Read-only file
    system (os error 30) at path "/nix/store/.tmpiaOTBO"

Population on one such machine, `grep -c` over the two debug logs
present, measured 2026-09-12T12:06 +0200: 24 lines in `debug.log`, whose
span begins 2026-09-10T11:39, and 224 in `debug.log.1`, spanning
2026-07-13T20:26 to 2026-09-10T11:39. `debug.log` is the live log, so its
end moves; every one of the 248 is the read-only variety. 17 of the 24
fall inside twelve seconds, one per session in a batch launch: the line
two above the first of them reads `INFO session.startup_recovery:
starting daemon recovery for missing tmux sessions count=17`. The
condition behind them cannot change while the process runs.

The message also names the wrong path. `/nix/store/.tmpiaOTBO` is the
temp file `atomic_write` opened next to the resolved target, so the line
identifies neither the agent nor the settings file to fix.

Warn once per settings target, naming the settings path, and log later
read-only failures for that target at debug. The memo is keyed on the
canonicalized target rather than the path, so a replaced symlink (a Nix
profile switch points it at a new store path) is reported once for the
new file; where the two differ the message names both, because two
settings paths can share one target and naming only the first would
silence the second without saying which file it was. The install itself
still runs, so a path that becomes writable installs with no special
case. Failures that are not read-only keep the existing message.

Measured against a genuinely read-only `settings.json`, one `aoe serve`
recovering three sessions: before, three `Failed to install agent hooks`
warnings; after, one warning naming the settings file and two debug
lines.

The detection reads `ErrorKind::ReadOnlyFilesystem` off the error chain
rather than a raw errno, because the temp-file layer rebuilds the error:
`session::atomic_write` into /nix/store surfaces
`kind: ReadOnlyFilesystem` with `raw_os_error: None`, so an errno check
would have matched nothing.

`a_real_install_hooks_error_keeps_its_io_error` drives a real
`install_hooks` failure and asserts the chain still carries a
downcastable `io::Error`, which is what the classification rests on and
what an error built inside the test cannot show. It uses a failure any
machine can produce, because a read-only filesystem is not portable to
make; that the read-only kind in particular survives that chain is the
/nix/store measurement above, not a test.

Not covered. The memo is process-local, which the warning now says: a
one-shot `aoe add` warns once per invocation, so a shell loop over N
sessions still prints N warnings. Collapsing those needs durable state
and is not attempted here. That the install keeps running after a
read-only failure is true by reading (`install_hooks` is called
unconditionally; only the reporting branches) and is driven by no test.
Only the JSON-settings host install is touched; the Codex, Cursor
sidecar and standalone sidecar installs in this file have the same shape
and carry their own messages, and sandboxed installs are untouched.

**Related issues:** none that I could find — searched for
`Failed to install agent hooks` and `read-only settings hooks`.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

```

fails in all three runs

```

```
this commit 686986a8, cargo test --lib, three runs
  run1  6955 passed; 4 failed; 2 ignored   (6961 tests)
  run2  6954 passed; 5 failed; 2 ignored   (6961 tests)
  run3  6954 passed; 5 failed; 2 ignored   (6961 tests)

    base 0/3, here 1/3   session::instance::terminal::tests::container_terminal_command_preserves_runtime_boundaries
    base 0/3, here 1/3   session::mcp::mcp_state::tests::resolve_conflict_aoe_wins_promotes_to_global_and_clears
    base 1/3, here 0/3   session::projects::tests::set_pinned_toggles_without_removing_entry
    base 1/3, here 0/3   session::projects::tests::update_base_branch_sets_clears_and_reports_not_found

  this commit's own tests, each checked by name in each run:
    ok, ok, ok   session::instance::hooks::tests::read_only_settings_are_reported_once_per_path
    ok, ok, ok   session::instance::hooks::tests::a_real_install_hooks_error_keeps_its_io_error

```

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: **no test was added under `tests/e2e/`; the coverage landed as in-crate tests, and I would rather say so than tick the box above.** This is on the session-build path, so the N/A box is not true either.

Two tests were added in `src/session/instance/hooks.rs`:

    read_only_settings_are_reported_once_per_path
    a_real_install_hooks_error_keeps_its_io_error

The first drives the memo: two reports against the same settings target, second
one suppressed, and a second target still reported, which is the part a
per-process flag would get wrong. The second exists because the obvious way to
write the first — matching on the error text — would also swallow a genuine
install failure, so it asserts a non-read-only error still arrives intact.

`AGENTS.md` asks for the cheapest test that can catch the regression, prefers
Rust unit tests, and reserves full-binary tests for behaviour that needs that
environment, so I read in-crate tests as following that rule rather than dodging
it — but the template's box asks specifically about `tests/e2e/`, so it stays
unticked and this is the reason.

A `tests/e2e/` case is also awkward here rather than unwanted: reproducing the
real condition needs a settings path that is genuinely read-only to the test
process, which on the machine where this was found is a Nix store symlink. If
you want it pinned end-to-end, I will take a suggestion on how you would rather
fake that.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

not see the earlier ones. The re-review verified the read-only detection against
the real error chain rather than taking the commit's word for it, which is why
`a_real_install_hooks_error_keeps_its_io_error` exists: an error constructed
inside a test cannot show the downcastable `io::Error` the classification rests
on. The log population quoted above (24 lines in `debug.log`, 224 in
`debug.log.1`, every one of them the read-only variety) was reproduced
independently by a second agent rather than carried over from the first.

The "Not covered" paragraph above was written by the same process and is not
padding: the memo is process-local, and that the install keeps running after a
read-only failure is true by reading and driven by no test.

## Summary

- Improved read-only settings installation reporting.
- Canonicalized missing settings paths through their parent directories.
- Reported each canonical target once and logged repeated failures at debug level.
- Added coverage for symlinked paths, target changes, error handling, and preserved I/O errors.

## Benefits

This reduces duplicate warnings and makes read-only settings issues easier to diagnose without changing installation behavior.

### DIFF
--- a/src/session/instance/hooks.rs
+++ b/src/session/instance/hooks.rs
@@ -596,11 +596,66 @@ impl Instance {
         ) {
             Ok(()) => true,
             Err(error) => {
-                tracing::warn!(target: "session.store", "Failed to install agent hooks: {}", error);
+                if is_read_only_filesystem(&error) {
+                    match first_read_only_report(&settings_path) {
+                        // The memo is keyed on the resolved target, so name it
+                        // whenever it differs: two settings paths can share one
+                        // target, and only saying the first would silence the
+                        // second without ever telling the user which file it is.
+                        Some(target) if target != settings_path => {
+                            tracing::warn!(target: "session.store",
+                                "Agent settings at {} resolve to {}, which is on a read-only filesystem, so AoE status hooks cannot be installed there; not reported again for that target while this process runs.",
+                                settings_path.display(), target.display());
+                        }
+                        Some(_) => {
+                            tracing::warn!(target: "session.store",
+                                "Agent settings at {} are on a read-only filesystem, so AoE status hooks cannot be installed there; not reported again for that file while this process runs.",
+                                settings_path.display());
+                        }
+                        None => {
+                            tracing::debug!(target: "session.store",
+                                "Agent settings at {} are still read-only; hook install skipped again.",
+                                settings_path.display());
+                        }
+                    }
+                } else {
+                    tracing::warn!(target: "session.store", "Failed to install agent hooks: {}", error);
+                }
                 false
             }
         }
     }
+}
+
+/// Settings files already reported as unwritable because they sit on a
+/// read-only filesystem. Externally managed settings (a Nix store symlink, a
+/// read-only mount) never become writable, so the identical warning would
+/// otherwise repeat for every session this process touches. Process-local by
+/// design: a one-shot `aoe add` still warns once, which is the first time that
+/// invocation says anything about it.
+static READ_ONLY_SETTINGS: std::sync::LazyLock<
+    std::sync::Mutex<std::collections::HashSet<std::path::PathBuf>>,
+> = std::sync::LazyLock::new(|| std::sync::Mutex::new(std::collections::HashSet::new()));
+
+fn is_read_only_filesystem(error: &anyhow::Error) -> bool {
+    error
+        .chain()
+        .filter_map(|cause| cause.downcast_ref::<std::io::Error>())
+        .any(|io| io.kind() == std::io::ErrorKind::ReadOnlyFilesystem)
+}
+
+/// The resolved target to report, or `None` when it has already been reported.
+/// Keyed on the target rather than the path, so a replaced symlink (a Nix
+/// profile switch points it at a new store path) is reported once for the new
+/// file. The install itself still runs, so a path that becomes writable
+/// installs with no special case.
+fn first_read_only_report(path: &std::path::Path) -> Option<std::path::PathBuf> {
+    let key = std::fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
+    READ_ONLY_SETTINGS
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner())
+        .insert(key.clone())
+        .then_some(key)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


jerome-benoit (dismissed):
Reviewed the complete PR diff at `686986a8` with two independent sequential analyses and source arbitration. No blocking finding.

The memo is consulted only after `install_hooks` fails (`hooks.rs:592-624`), so retries still occur and recovery to writable storage is not suppressed. The error-chain predicate only handles ReadOnlyFilesystem (`:640-644`); other failures keep their warning. Mutex-protected insertion makes concurrent deduplication coherent.

The existing [missing-leaf canonicalization comment](https://github.com/agent-of-empires/agent-of-empires/pull/3930#discussion_r3997035788) is valid but P3/non-blocking: alias paths may still produce duplicate warnings, without changing installation or error handling. Canonicalizing the existing parent and adding the proposed regression would close that edge case. One independent analysis rated it P2; final arbitration retains P3 because the remaining effect is duplicate diagnostics for absent-leaf aliases, not failed installation or lost errors. No duplicate inline finding.

Tests cover classification, existing targets/aliases and a real installer read error. They do not reproduce an EROFS write or assert tracing counts; the Nix probe remains author-reported. Process-local lifetime and JSON-host-only scope are intentional, not reasons to broaden this PR.

Evidence: exact-head git diff/show and clean diff --check. No local builds/tests run. CI was authorized for this SHA and is running; this approval does not claim full CI success.

PR #3931, opened 2026-09-12
docs(cli): say what `aoe rm --purge` actually removes

blueagledev:
## Description

`aoe rm --purge --help` promises "the irreversible teardown
(worktree/branch/container cleanup per the other flags, plus transcript
removal)". A reader takes "transcript" to mean the conversation, and for
a host session it is not that.

What `aoe rm` removes under that word is the session's rows in AoE's own
structured-view event store: `purge_acp_transcript`
(src/cli/mod.rs:121-129) deletes from `acp_events.db` and nothing else,
and its own doc comment calls the target "the local UI transcript".
`remove.rs:247-254` reaches it through `complete_with` and never touches
the supervisor. For a terminal session it deletes zero transcript rows,
which the implementation states in as many words: "deleting zero rows
for a terminal session is harmless".

The purge does remove one more thing the help never mentions, and it is
the half that can lose a conversation: `stage_remove_agent_stores`
(src/session/deletion.rs:1220) calls
`sandbox_store_reclaim::remove_stores_for`, deleting the per-instance
agent store. For a sandboxed session that store is the containerized
agent's own config home, so its history goes with it. Saying only that
the conversation survives would have pointed the other dangerous way.

So the help now names both, and scopes what survives to a host agent's
own history outside them. `docs/cli/reference.md` is regenerated with
`cargo run -p xtask -- gen-docs`, which leaves no other diff.

Scoped to the CLI on purpose. The web delete dialog makes the same claim
in the same words, but the web "Delete permanently" path is not this one.
For a structured session (`delete.rs:115`, gated on
`instance.is_structured()`) it reaches `shutdown_and_delete`, which calls
`try_session_delete` under `delete_adapter_state`
(supervisor.rs:3130-3134, whose comment reads "Only permanent removal
deletes the agent-side transcript"), and
`docs/development/internals/structured-view.md:37` records that adapters
implementing that RPC clear their on-disk session record. The CLI has no
running worker to send it, which `purge_acp_transcript`'s own doc comment
says (`mod.rs:112-114`). The dialog therefore needs its own sentence and
is left alone here.

Established by reading the deletion stages and the mount map, not by
purging a sandboxed session end to end: this seat has no container
runtime.

**Updated after review, at `79a98263`.** The description above is the first
commit's, and one of its claims did not survive: the per-instance agent store is
**not** removed unconditionally — it goes with the sandbox container, so keeping
the container keeps the stores. The shipped help text is the reviewer's own
suggested wording, and `docs/cli/reference.md` was regenerated with
`cargo xtask gen-docs` rather than hand-edited, so the source and the reference
cannot drift. The paragraph beginning "The purge does remove one more thing"
should be read against that correction.

**Related issues:** none that I could find — searched for `purge transcript`.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [x] Documentation
- [ ] Infrastructure / CI

## Checklist

`docs/cli/reference.md` is generated, so it was regenerated rather than hand
edited: `cargo run -q -p xtask -- gen-docs` exits 0 and leaves the tree clean at
this commit, which is the check that the doc comment and the reference cannot
drift apart here.

```

fails in all three runs

```

```
this commit 18c5295e, cargo test --lib, three runs
  run2  6952 passed; 5 failed; 2 ignored   (6959 tests)
  run3  6952 passed; 5 failed; 2 ignored   (6959 tests)

    base 0/3, here 1/3   acp::acp_client::resolve_command::tests::probe_version_bounded_reads_version_output
    base 1/3, here 1/3   session::projects::tests::set_pinned_toggles_without_removing_entry
    base 1/3, here 1/3   session::projects::tests::update_base_branch_sets_clears_and_reports_not_found

```

This commit adds no tests, so the population here is the base's exactly —
6959 both sides — and nothing in `src/` or `tests/` asserts the `--purge` help
text, which I checked with a positive control rather than an empty grep.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: copy only. The change is the `--purge` doc comment and the reference line generated from it; no code path, flag behaviour or output other than the help text changes, which is the "copy change" case the template names.

The claim the wording makes is about `--purge`'s actual scope, which is why the
review's correction mattered more here than it would for an ordinary copy change:
**the first commit, and the first version of this paragraph, said the agent store
goes unconditionally. It does not.** `--keep-container` or
`sandbox.auto_cleanup=false` sets `delete_sandbox=false` (`remove.rs:220-222`),
`container_gone` then stays false (`deletion.rs:723-731`) and
`stage_remove_agent_stores` is skipped (`:749-750`). A *failed* teardown leaves
`container_gone` false as well, even where `delete_sandbox` was true, so the
reviewer's "attempts to remove" covers a case the original report did not.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

not see the earlier ones.

Worth being explicit about for a wording change, since the wording is the whole
claim: what `--purge` removes was established by reading the code paths —
`purge_acp_transcript` (`src/cli/mod.rs:121`) for the structured-view transcript
and `stage_remove_agent_stores` (`src/session/deletion.rs:1220`) for the
per-instance agent store — and **not** by purging a sandboxed session and looking
at what disappeared. If you would rather the sentence rest on an observation than
on a reading, that is a fair thing to ask for and I will go and get it.

## Summary

- Clarified `aoe rm --purge --help` and the generated CLI reference.
- Documented which transcript rows and sandbox agent stores CLI purge removes.
- Clarified that sandbox stores are removed only when sandbox teardown removes the container.
- Confirmed that host agent history outside those stores remains unaffected.
- Documented that CLI purge does not invoke the `session/delete` adapter flow.

This improves purge transparency and reduces the risk of unexpected data loss.

### DIFF
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -327,7 +327,7 @@ Remove a session
 * `--force` — Force worktree removal even with untracked/modified files
 * `--keep-container` — Keep container instead of deleting it (default: delete per config)
 * `--keep-scratch` — For scratch sessions, keep the scratch directory on disk instead of removing it. The session record is still deleted; the kept path is logged so you can find the files later. No effect on non-scratch sessions
-* `--purge` — Permanently delete instead of moving to trash. By default `rm` moves the session to the trash (when `session.delete_to_trash` is enabled, the default) so it can be restored; `--purge` forces the irreversible teardown (worktree/branch/container cleanup per the other flags, plus the session's structured-view transcript and its own agent store, which for a sandboxed session is that agent's config home). A host agent's conversation history outside those is not removed
+* `--purge` — Permanently delete instead of moving to trash. By default `rm` moves the session to the trash (when `session.delete_to_trash` is enabled, the default) so it can be restored; `--purge` forces the irreversible teardown (worktree/branch/container cleanup per the other flags) and removes the session's structured-view transcript. Removing the sandbox container also attempts to remove its private agent stores, including its config home; keeping the container keeps those stores. Host agent conversation history outside these stores is not removed
 
 
 
--- a/docs/cli/reference.md
+++ b/docs/cli/reference.md
@@ -327,7 +327,7 @@ Remove a session
 * `--force` — Force worktree removal even with untracked/modified files
 * `--keep-container` — Keep container instead of deleting it (default: delete per config)
 * `--keep-scratch` — For scratch sessions, keep the scratch directory on disk instead of removing it. The session record is still deleted; the kept path is logged so you can find the files later. No effect on non-scratch sessions
-* `--purge` — Permanently delete instead of moving to trash. By default `rm` moves the session to the trash (when `session.delete_to_trash` is enabled, the default) so it can be restored; `--purge` forces the irreversible teardown (worktree/branch/container cleanup per the other flags, plus transcript removal)
+* `--purge` — Permanently delete instead of moving to trash. By default `rm` moves the session to the trash (when `session.delete_to_trash` is enabled, the default) so it can be restored; `--purge` forces the irreversible teardown (worktree/branch/container cleanup per the other flags, plus the session's structured-view transcript and its own agent store, which for a sandboxed session is that agent's config home). A host agent's conversation history outside those is not removed
 
 
 
--- a/src/cli/remove.rs
+++ b/src/cli/remove.rs
@@ -37,10 +37,11 @@ pub struct RemoveArgs {
     /// Permanently delete instead of moving to trash. By default `rm` moves
     /// the session to the trash (when `session.delete_to_trash` is enabled,
     /// the default) so it can be restored; `--purge` forces the irreversible
-    /// teardown (worktree/branch/container cleanup per the other flags, plus
-    /// the session's structured-view transcript and its own agent store, which
-    /// for a sandboxed session is that agent's config home). A host agent's
-    /// conversation history outside those is not removed.
+    /// teardown (worktree/branch/container cleanup per the other flags) and
+    /// removes the session's structured-view transcript. Removing the sandbox
+    /// container also attempts to remove its private agent stores, including
+    /// its config home; keeping the container keeps those stores. Host agent
+    /// conversation history outside these stores is not removed.
     #[arg(long)]
     purge: bool,
 }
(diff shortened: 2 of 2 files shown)


jerome-benoit (changes requested):
Reviewed the two-file documentation diff at `18c5295e` with two independent sequential analyses and source arbitration.

**P2, blocking for this documentation fix:** please address the existing [conditional agent-store removal finding](https://github.com/agent-of-empires/agent-of-empires/pull/3931#discussion_r3997039280). `--keep-container` or `sandbox.auto_cleanup=false` sets `delete_sandbox=false` (`remove.rs:220-222`); `container_gone` then stays false (`deletion.rs:723-731`), and store removal is skipped (`:749-750`). The new wording currently promises that removal unconditionally.

Suggested scope: “removes the session's structured-view transcript. Removing the sandbox container also attempts to remove its private agent stores, including its config home; keeping the container keeps those stores. Host agent conversation history outside these stores is not removed.” Update the CLI source and regenerate the reference together. No duplicate inline comment.

Verified/rejected: the CLI deletes local ACP event-store rows (`cli/mod.rs:108-128`), not the adapter's transcript via RPC; the web deletion path is different and need not be changed here. No runtime behavior changes or new tests are required for this copy-only correction.

Evidence: exact-head git diff/show and clean diff --check. No purge reproduction, local build or author-reported test rerun. CI was authorized at this SHA and is running.

jerome-benoit (dismissed):
Re-reviewed `79a9826353c8ba18a3f3ff97d239532ebe86f088`, including the updated description and resolved thread. The prior P2 is addressed: `src/cli/remove.rs:37-44` and `docs/cli/reference.md:330` consistently qualify private-store cleanup by container removal. This matches the `delete_sandbox` gate (`remove.rs:221-223`) and the guarded, best-effort cleanup (`deletion.rs:723-750,1220-1233`).

Two independent sequential readings agree that the previous blocker is fixed. One remaining P3 wording nuance, non-blocking after arbitration: another teardown error can preserve stores even after container removal (`deletion.rs:749`). “After a successful teardown” would make that condition explicit; the current “attempts” wording no longer promises deletion. The CLI transcript remains distinct from adapter-side history; no web-path changes or additional runtime tests are required for this documentation-only correction. Evidence: exact-head `git diff`/`git show`, clean `git diff --check`; no local build or purge reproduction. CI, Tests and CodeQL were awaiting authorization at this new SHA and have now been authorized, not claimed as passed.

PR #3936, opened 2026-09-13
fix: report ACP context reset when native resume is unavailable

jerome-benoit:
## Description

When a restarted ACP adapter cannot load a stored session, AoE could silently create an empty native context while retaining the displayed conversation. Emit the existing `SessionContextReset` boundary after successful replacement and before assigning its identity.

Ordinary load-error fallback now uses that same successful-new boundary. Failed replacement, initial creation, successful load, and live-worker reattachment do not introduce an ordinary fallback reset. Existing fork-control resets retain their separate marker-clearing semantics and are not duplicated. Earlier transcript rows remain; no migration or historical backfill is attempted.

**Reproduction and causality:** A controlled ACP child advertised `loadSession=false` with a stored identity. The actual client sent `session/new`, not `session/load`; the reopened journal retained prior turns but contained zero resets instead of one. Two independent RCA reviewers identified the skipped capability branch. A temporary post-new notification made the same reproduction pass through storage and transcript folding, then was restored. The wire refutes an attempted-load failure; retained rows refute replay loss; the diagnostic round-trip refutes downstream filtering of a published boundary.

This is a controlled protocol reproduction, not a demonstrated native-provider memory incident. The design passed two consecutive clean rounds with two fresh independent reviewers per round.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass
- [x] Documentation was updated where necessary
- [ ] For UI changes: included screenshot or recording

Historical verification and final review gates completed on 8520d2db8cebf6d3cbc5301b5a860b0ef89c53cb (not current-head full-suite or review results):

- Complete default-feature library suite: 6970 passed, 2 ignored; standalone integration target: 1 passed; combined integration suite: 289 passed, 5 ignored; doctests: 0 run, 1 ignored. `cargo build` passed.
- GitHub checks: 34 passed, 1 skipped.
- Final code review: two consecutive clean rounds, two fresh independent agents per round, no mandatory findings. Reviewers read the entire change, consumers, actual regression, and immutable runtime evidence; they did not run separate builds or claim human approval.
- Local full-suite execution used non-terminal stdin and process-only `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=commit.gpgsign GIT_CONFIG_VALUE_0=false` for temporary Git fixtures. No persistent Git configuration changed.

Additional historical focused verification on the original implementation:

[...]

PR #3937, opened 2026-09-13
fix: fence ACP events and callbacks by native session identity

jerome-benoit:
## Description

Enforce the lifecycle-selected native ACP session identity before inbound updates or session-scoped callbacks can affect AoE state. Foreign notifications cannot alter transcript/tool/watchdog state, and foreign callbacks cannot perform filesystem, terminal, permission or elicitation effects. Notification and callback metadata never establish session authority.

Fence identity transition, assignment and buffered replay; preserve matching updates on failed or same-ID resets. Publish identity in ordered SDK response handling and announce runner SessionReady before admitting callbacks. Runner attachment generation 4 uses the existing owner-checked incompatible-runner retirement path. This PR stacks on #3936 so unavailable native resume clears application context only after successful replacement establishment.

Keep the user-selected 4096-frame / 128 MiB pending limits. For control transport, account the original validated runner frame bytes (including the four-byte prefix), not a typed JSON reserialization that can expand a valid producer backlog. Trusted byte provenance crosses the private in-process SDK bridge without extra nesting; native marker fields cannot override it. Direct stdio preserves existing object/positional decoding and bounded typed accounting. Overflow fails the connection rather than publishing partial replay.

No storage/path/schema migration or history deletion. Stored native IDs remain available for supported replacement loads. Existing owned runner registry/socket retirement is unchanged; no old records are retagged.

**Limits:** this does not repair or establish global ordering for same-native-ID load replay or valid replies delivered after Stopped. Unknown pre-response callback identities fail closed. Encoded replay bounds are not RSS/CPU/decoder-allocation guarantees. Identity comparison cannot detect semantically wrong content carrying the correct ID. Scripted native peers were used; no real-provider leakage or native-memory-continuity claim is made.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

The scoped checks below pass; project-wide validation and final published-head reviews are reported separately. No visual changes were made.

## Test Coverage Analysis

Final-parent default-feature verification:
- `cargo test --lib acp::`: 754 passed.
- `cargo test --lib server::acp`: 84 passed.
- `cargo test --lib process::runner::tests`: 24 passed.
- `cargo test --lib process::worker_registry::tests`: 27 passed.
- `cargo test --test integration acp_runner_control`: 10 passed.
- `cargo clippy -- -D warnings` and `cargo fmt -- --check`: passed.

Before-fix identity reproduction admitted FOREIGN/LATE chunks despite outgoing prompt IDs changing correctly; the retained regression now passes. The byte-accounting reproduction used three individually valid frames totaling 134,217,426 raw bytes but 134,217,855 typed bytes: an original producer-valid backlog was rejected. An equal-raw-size diagnostic changing numeric representation removed the failure, then was restored. The actual AcpClient/control/SDK regression failed before the byte fix and passes after it.

On `74e477a5` (before the positional-callback correction below), an actual production runner + AcpClient spawn/detach/reattach smoke passed: all three near-cap tool calls arrived, authoritative identity replaced stale stored identity, a foreign callback returned -32602, an admitted own write completed, live and Stopped events were collected, and the runner registry was retired. A detached -32001 callback response served as an explicit producer barrier. This used an isolated scripted native peer, not a real provider or full daemon reconciler. The throwaway executable was removed.

Permanent regressions cover foreign effects, transition/rollback/overflow behavior, the admission-fence race, exact original-wire budget boundaries, actual SDK expansion, positional/depth decoding compatibility, and untrusted marker handling. A final-parent test-fixture mismatch initially caused one ACP failure because upstream changed a shell variable; moving injection to the stable method case fixed the fixture, without changing production or weakening assertions. All 754 ACP tests then passed.

The final amendment preserves the user's merge `0aa25ca05f345743274055386818d1ffa1f1f261` and both fixture corrections. The original foreign-update regression now requires a valid own-ID control from inside the injection block for each session. An intentionally nonmatching anchor failed with missing controls; restoring the user's anchor passed with exactly `[working, OWN-CONTROL, working, OWN-CONTROL]`. This guards against silently testing no foreign traffic. The actual runner smoke and all scoped checks were rerun on this adapted source.

Final review also reproduced a positional-callback ordering omission on an immutable actual runner: an equivalent named callback waited for SessionReady while the valid positional callback overtook it. The cached-reattach gate now recognizes positional forms only for the eight session-scoped permission/filesystem/terminal methods, without cloning or rewriting parameters. Request-scoped elicitation remains ungated. The extended existing actual-runner lifecycle regression failed before correction and passes afterward; the same runtime probe now observes neither callback shape before SessionReady and receives both afterward.

- [x] Skipped a test, because: the changed admission contract is below browser rendering and is exercised through actual ACP/SDK/socket regressions and the production runner smoke; no dashboard controls or rendering changed.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: lifecycle/transport behavior is covered by scoped Rust regressions, `tests/integration/acp_runner_control`, and actual runner smoke rather than a TUI rendering scenario. No web/e2e feature switch was used.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Oh My Pi / OpenAI Codex (gpt-6-astra).

**Any Additional AI Details you'd like to share:** AI-assisted implementation with independent read-only RCA, design and implementation reviews. The byte-accounting amendment passed two consecutive clean design rounds after earlier rounds rejected nesting and positional-compatibility regressions. Protocol reproductions and verification ran in the isolated worktree. Publication and final GitHub reviews are separate; no merge is authorized without the user's GO.

**NOTE:**

Current published head: `acb7cf1175687c1a940086860505f93a44e0209c`. Two consecutive fresh published-head review pairs (R2 and R3) are complete, all four CLEAN. CI, Tests, and CodeQL succeeded on this exact head. Ready for human review; no merge requested or performed. [Review and CI record](https://github.com/agent-of-empires/agent-of-empires/pull/3937#issuecomment-5670204714). 

## What changed

- Added native session identity checks for ACP notifications and callbacks.
- Blocked foreign sessions from changing session state or using files, terminals, permissions, and elicitation callbacks.
- Added bounded buffering during runner reattachment.
- Preserved original wire-frame sizes for accurate replay limits.
- Enforced ordered session handshakes and callback delivery.
- Updated runner generation to 4.
- Replaced `parking_lot` with the standard mutex.
- Added regression and integration tests for identity checks, replay limits, resets, ordering, and reattachment.

## Benefits

- Prevents cross-session state changes.
- Makes runner reattachment safer.
- Prevents unbounded replay memory use.
- Preserves valid updates across failed or repeated session transitions.

Validation results were reported for focused tests, CI, formatting, clippy, and CodeQL.

## Further enhancement

Continue monitoring replay limits and reattachment behavior under sustained high-volume traffic.


jerome-benoit (commented):
Final independent review gate complete at ab0f6cbcd6e86ca2403b1fb7fe14b99b5891903f, based on #3936. Two consecutive clean rounds, four fresh reviewers: R1 admission and lifecycle; R2 failure boundaries and continuity. Each reviewed the complete 13-file diff plus affected callers/tests/docs. No actionable findings remain. These are source/evidence reviews, not independently executed runtime tests.

Full local source verification: library6980 pass/2ignored; standalone integration1 pass; combined integration290 pass/5ignored; doc0run/1ignored; build/clippy/fmt passed. Final commit is documentation-only. CI34 pass/1skip.

The inspected SDK was independently confirmed through locked Cargo metadata as unpatched crates.io agent-client-protocol2.0.0. An apparent callback-serialization concern was withdrawn because dependency dispatch already awaits callbacks to completion in the base.

Controlled actual runner/client/registry helper smoke covers upgrade and Reset<Assigned<early matching update, preserving prior journal rows. It is not whole-reconciler, browser or real-provider proof. Existing same-ID replay and replies afterStopped in the asynchronous control-to-SDK bridge remain outside this patch's guarantee.

Ready for human review. No merge performed or requested.

njbrake (dismissed):
Fix before merge or as a fast follow: reattaching to a live runner fails once more than 1,024 notifications were buffered while detached, and the reconciler then kills the agent.

On runner resume, `ingress.begin()` runs at connect (`connection.rs:199-201`). The runner flushes its detach queue (up to 4,096 frames / 128 MiB, `runner.rs:707-708`) ahead of `SessionReady`, but the pending cap is 1,024 (`session_identity.rs:12`). Overflow fails the attach and `acp_reconciler.rs:1748-1754` calls `terminate_and_wait`. A daemon restart during a long streaming turn is enough to hit it.

Repro with the real `aoe __acp-runner` and a scripted agent emitting N updates while detached: N=1000 reattaches; N=1100 and N=4000 fail 3/3 with `pending native session updates exceed replay capacity`. The #3936 base already hangs at 1,000, so this is not a regression, but the limit is undisclosed.

Size the cap to the runner detach queue, or skip buffering updates that match the stored id on resume. At minimum, document the limit.

jerome-benoit (commented):
Rechecked `9affe64b5412abbfc58cb619b188fd8e8c39625b`, including the maintainer's review and the [author's response](https://github.com/agent-of-empires/agent-of-empires/pull/3937#issuecomment-5666607217). Shared limits fix the reported capacity mismatch for a fixed detached backlog. Keeping identity validation rather than trusting the stored ID is appropriate.

**P2 residual, nonblocking for this bounded correction:** equal queue capacities do not guarantee every live reattachment succeeds. With a previous `session/new` still pending, `resume_session` waits (`src/process/runner.rs:1515-1532`) while the new control writer already drains notifications (`2035-2045`). Drained frames free runner capacity (`1327-1344`), but ingress accumulates them until identity commitment (`src/acp/acp_client/session_identity.rs:144-165`). Consequently, 4,000 small updates followed by 97 more before the reset response can exceed ingress capacity without ever exceeding runner queue capacity. Attach failure terminates the worker (`src/server/acp_reconciler.rs:1747-1753`).

This is a source-derived scheduling counterexample, not an executed runtime reproduction or a new regression from the limit increase. Please qualify the new whole-backlog guarantee. A complete follow-up needs an explicit bounded pre-identity delivery contract, preserving authoritative identity and control-response progress, plus a delayed-reset/continuing-notifications runner regression. Raising another constant alone cannot establish that contract.

The existing new tests cover fixed-cap admission and overflow, not this concurrent reattach sequence. Two independent sequential analyses and parent arbitration; exact-SHA source/diff inspection; CI: 34 successful, one skipped. No local tests rerun. COMMENT on my own PR, not a human approval.

jerome-benoit (commented):
Reassessed `acb7cf1175687c1a940086860505f93a44e0209c` after the [maintainer's streaming-reattach reproduction](https://github.com/agent-of-empires/agent-of-empires/pull/3937#pullrequestreview-5208527279). **Correction to my previous verdict: the unresolved replay overflow should be fixed before merge.** Calling it a nonblocking residual was too permissive for the live-reattach requirement.

The source supports the failure: ingress accumulates updates until identity commitment (`connection.rs:202`, `session_identity.rs:220-246`), while runner delivery frees queue capacity (`runner.rs:1350-1367`). One instantaneous queue bound therefore does not bound all traffic admitted before `SessionReady`. The reconciler terminates the worker after attach failure (`acp_reconciler.rs:1747-1753`). The reported 3/3 streaming failures and quiet/smaller-backlog controls are the maintainer's runtime evidence, not a reproduction I executed.

Establish a bounded pre-identity delivery barrier, rather than increasing another limit. Announcing committed cached identity before replay is appropriate for stable reattach, but must not expose a stale ID when a previous reset is pending (`runner.rs:1538-1555`). Preserve handshake order, attachment ownership, callback admission and control-response progress. Add an actual-runner full-backlog-plus-live-stream regression with explicit synchronization, alongside quiet backlog and pending-reset controls.

The raw-byte accounting and positional-callback fixes remain valid. The changelog hunk and unnecessary direct synchronization dependency are separate maintainer cleanup requests, not the cause of the runtime failure. Both independent sequential reassessments confirm the blocker and recommend P1; my arbitration retains P2 because it requires a near-capacity backlog plus continued traffic, without demonstrated persistent-history loss. It nevertheless blocks acceptance of live reattach. No local tests rerun. COMMENT on my own PR, not an approval.

njbrake (changes requested):
Request changes: the agent is still killed when a daemon reattaches while a turn is still streaming, the most common form of the original report.

With a full detach queue, the runner flushes the backlog, then `Initialized`, then `SessionReady`. While it flushes, `make_room` frees space, so live `session/update` frames keep entering the queue ahead of `SessionReady`. The daemon buffers all of them from `ingress.begin()` (`connection.rs:202`), so any live traffic on top of a full backlog goes past `MAX_PENDING_UPDATES`. The attach then fails and the reconciler terminates the worker.

Repro: real `aoe __acp-runner` with a scripted agent. The agent emits 5000 updates while detached, then keeps streaming one every 2 ms across `AcpClient::attach`. It fails 3/3 with `pending native session updates exceed replay capacity`. The same backlog with a quiet agent reattaches 3/3, and 2000 plus streaming reattaches 3/3.

A larger limit cannot fix this, because pre-identity traffic is unbounded. Pre-identity buffering has to end before the flush: for example, announce the cached identity ahead of the backlog on cached reattach. Add a runner-level regression test for streaming during reattach.

Nits:
- Drop the `CHANGELOG.md` hunk; git-cliff regenerates the file at release.
- Use `std::sync::Mutex` instead of adding `parking_lot` as a direct dependency.

jerome-benoit (commented):
Re-reviewed `500aca0fbd96002237cca88ba850aeabf8e027e4`. Prioritizing the authoritative handshake addresses the reported streaming-backlog mechanism, but the new writer gate introduces a **P2 blocker**.

`src/process/runner.rs:1348-1352` holds every non-Handshake frame until `session_announced`, including `ServerCall`, not just replay notifications. During `session/new`, an agent can wait for a client filesystem callback before returning its session ID: the callback now waits for SessionReady, while SessionReady waits for that callback. The existing `agent_request_during_session_new_does_not_deadlock_the_runner` establishes precisely this cycle (`tests/integration/acp_runner_control.rs:453-480,546-550`). The pending-reset exception at `runner.rs:1034-1036` cannot help if the writer still withholds the admitted callback.

[CI at this head](https://github.com/agent-of-empires/agent-of-empires/actions/runs/34980889527/job/104420607215) fails that test and three other control-channel tests with `read frame length: WouldBlock`. This is consistent with the source regression, not evidence for dismissing the failures as transient. Preserve handshake/reset control progress while withholding the notification backlog, and keep these regressions passing alongside the new streaming case.

Also replace the new test's 200 ms wait (`acp_runner_control.rs:1572`) with a runner-consumption barrier: the Python marker proves emission, not that the detached queue is full. Avoid claiming that observing one assigned ID and one chunk proves the entire backlog was exercised.

Two independent sequential reviews and parent source/CI arbitration. No local suite rerun. COMMENT on my own PR; not approved for merge.

jerome-benoit (commented):
Re-reviewed `9c4aadba0a8a2eca20f7771f3f596fe8a6fec0d8` against the previously reviewed `500aca0f`. **The callback deadlock is fixed.** `next_outbound` now lets attachment-scoped callbacks progress before `SessionReady`, while retaining the persistent backlog behind the identity handshake (`runner.rs:1342-1358`). CI, Tests and CodeQL all passed on this head.

**P2 validation requirement still open from my previous review:** `acp_runner_control.rs:1592-1594` still uses an agent-emission marker followed by a 200 ms sleep. Neither establishes that the runner consumed the full detached backlog before reattachment. The assertion loop exits after one assigned identity and one chunk (`1613-1628`). Please replace that timing assumption with an observable runner-consumption barrier and verify progress through the intended backlog while streaming. This is not a claim that the production overflow still occurs; it is the missing causal regression coverage for the exact failure the maintainer reproduced.

The revised large-frame test correctly establishes identity before stalling the reader and verifies complete delivery of the requeued frame. Two independent sequential delta reviews were arbitrated against exact-SHA `git diff`/`git show`; no new production defect was established. No local tests rerun. COMMENT because this is my own PR, not self-approval.

njbrake (approved):
Streaming-reattach overflow resolved: the backlog is held until SessionReady. The regression test now gates on runner consumption instead of a fixed sleep.

Follow-up, not blocking: session/load history replay now drains after SessionReady, so a prompt sent while an idle-dormant session wakes turns off replay suppression before the drain finishes and replays the full history into the transcript. Main already leaks part of it.

PR #3952, opened 2026-09-14
feat(acp): add model gateway discovery + per-agent model routing

dkattan:
## What

Ports nodeterm's model gateway to AOE end-to-end: one gateway root configured once (config.toml `[acp] gateway_base_url` / `gateway_api_key` / `gateway_discovery_path`), serving every supported harness through its own routing env, plus the discovery endpoint and the web model picker.

- **`src/acp/model_gateway.rs`** — one root + non-secret credential reference (`${env:VAR}` or `${secret:model-gateway-api-key}`), per-agent route derivation (claude → `ANTHROPIC_BASE_URL`/`AUTH_TOKEN`, codex → `OPENAI_BASE_URL`/`API_KEY`, copilot BYOK requires an explicit model), OpenAI-convention discovery parsing with `context_length`/`max_context_length`/`context_window`, and the Claude-specific large-context/subagent machinery (`[1m]` + autocompact pair, `CLAUDE_CODE_SUBAGENT_MODEL` trio). Fail-closed everywhere: an unresolvable credential emits nothing, never a partial credential. A credential never rides argv — values travel by host_environment only.
- **`GET /api/acp/models`** — daemon-side discovery (browser CORS + credential isolation); the ONLY fetchable URL is the one persisted in config.toml, so the credential-exfiltration oracle is closed structurally. Sends the standard bearer header plus Bifrost's legacy `x-bf-vk`. Publishes successful results into a config-digest-scoped in-memory cache for the spawn path.
- **Supervisor spawn** — the gateway routing env + `claude_adapter_model_env` (ANTHROPIC_MODEL pin, CLAUDE_MODEL_CONFIG picker seed, [1m]/autocompact, subagent routing) applied last-wins in `host_environment`, never argv.
- **Launch command (terminal view)** — same gateway routing into tmux pane env mutations, fail-closed.
- **Web** — the chat's Model dropdown merges a Gateway section (fetched lazily from `/api/acp/models`, module-level cache) under the adapter's rows; each row shows the model id + context size. Selection rides the existing `set_config_option` endpoint and persists via `agent_model`, so a worker respawn re-asserts it.

## Verification

- 35 unit tests in `model_gateway.rs` (routes, credential refs, discovery parsing, `[1m]` pairing invariant, subagent routing, catalogue cache scoping) + 773 ACP-module tests green; clippy/fmt clean.
- Live against Bifrost: 5-model discovery; a 400k-context model spawned as `vllm/reasoning[1m]` with `CLAUDE_CODE_AUTO_COMPACT_WINDOW=400000`; a turn completed through the gateway route (1M context window reported by usage); a `set_config_option` switch to another gateway model survived a worker respawn as the new `ANTHROPIC_MODEL`.

## Not included / follow-ups

- The daemon-side gateway fields are `local_only` in the settings schema (credential references resolve only on the daemon host), so configuration is via config.toml — the dashboard PATCH strips them by design.
- Codex/Copilot pickers surface gateway models only where the harness's own config options carry them; the Gateway section currently renders on the model-category dropdown, which all current adapters share.


dkattan:
Follow-up on this branch: the three gateway fields moved out of `[acp]` into their own top-level `[model_gateway]` config section (`188f7a7e`), so the settings UI gives them their own tab naturally instead of inheriting Structured View's.

- `ModelGatewayConfig` is a `SettingsSection` registered in the schema registry like every other section; the three read sites (ACP spawn env in the supervisor, launch-command env, the discovery handler) now read `config.model_gateway`.
- TUI gets a new "Model Gateway" category under Sessions; the web UI gets a "Model gateway" tab beside Structured View. Both render schema-driven, so no per-field row building was needed — and all write paths (TUI apply, web `saveSubField`, the elevation-gated PATCH policy) are generic and unchanged.
- No back-compat aliases (greenfield branch): a PATCH to `acp.gateway_*` is now an unknown-field rejection, pinned by a new test; the elevation pin moved to the new section.

The live config's values were migrated, and the whole loop was verified against a restarted daemon: discovery returns the bifrost model list, PATCH to the new section is 200, non-elevated PATCH is refused.

PR #3963, opened 2026-09-14
fix(status): keep Claude's waiting hook when the question scrolls away

### LINKED ISSUE #3441: A correct AskUserQuestion "waiting" hook is discarded when the prompt scrolls out of the 30-line window
`aoe` already works around Claude having no Waiting-mapped hook for
`AskUserQuestion` — `waiting_tools` on `PreToolUse` writes `waiting` when the
payload's `tool_name` matches, and the `PostToolUse` matcher restores `running`
when the answer lands. That part works and is a neat solution to a real gap
(`src/agents.rs:406` documents the gap exactly).

The problem is what happens next. `reconcile_waiting_hook` treats the pane as
authoritative over that hook:

```rust
match detect_status_from_content(raw_content, agent) {
    Status::Waiting => Status::Waiting,
    // Prompt gone (Esc-cancelled, or answered with a missed clearing hook):
    // the detector's fresh read of the pane wins over the stale hook.
    other => other,
}
```

But "prompt gone" is decided by `with_claude_recent_pane`, which keeps only the
**last 30 non-empty lines**. A prompt that is still on screen and still blocking,
merely pushed above that boundary by later output, is indistinguishable from one
that was cancelled. Measured:

```
prompt alone,      hook=waiting -> Waiting
prompt pushed out, hook=waiting -> Idle
```

The second pane is the identical prompt with 31 further lines appended. Nothing
was answered or cancelled.

## Why this is reachable rather than theoretical

Anything that writes to the pane after the question does it. In my case it is
cross-session notifications from other agents landing in the same pane, but a
long tool result, a verbose hook, or a chatty MCP server would do the same. The
prompt does not need to leave the screen — it only needs 30 non-empty lines
after it.

The consequence is the bad direction: a session genuinely blocked on the user
reports `Idle`, drops out of the waiting count, and sits there until someone
notices by hand. That is the failure the `waiting_tools` hook was added to
prevent, reintroduced one layer down.

## Suggested direction

Not a patch, because the right answer depends on how you want the two sources
ranked, and I would rather not guess:

- **Trust the hook until a clearing hook arrives.** `PostToolUse` on
  `AskUserQuestion` already restores `running`, so an answered question has a
  positive signal. Esc-cancellation is the case that lacks one, which is
  presumably why the pane veto exists.
- **Require positive evidence to clear.** Only downgrade a hook `waiting` when
  the pane shows something that positively contradicts it (a live spinner, a
  ready prompt) rather than on the mere absence of a menu — the same
  positive-evidence discipline `reconcile_claude_idle_hook_status` already
  documents for the Idle direction.
- **Age-gate it**, the way `IDLE_RECONCILE_MIN_RUNNING_AGE` gates the other
  reconciler, so a freshly written `waiting` is not overridden by one poll.

## On the underlying gap

For what it is worth: `AskUserQuestion` firing no `Notification` hook looks like
an upstream Claude Code limitation rather than anything `aoe` can fix, and your
`waiting_tools` workaround is the right shape given that. I looked for an
official "is this session blocked" query as an alternative and could not
substantiate one: on Claude Code **2.1.232**, `claude agents --json` returns
`[cwd, kind, name, pid, sessionId, startedAt, status]` with `status` only ever
`busy` or `idle` — no state or waiting-reason field. If a newer version exposes
one, that would be a cleaner source than either the hook or the pane.

## Repro

The two panes above through `reconcile_waiting_hook("claude", ...)`, second one
being the first with 31 lines appended.

njbrake:
## Description

When Claude asks the user a question, aoe marks the session as waiting. If more output then appears below the question (for example notifications from other sessions), the question scrolls out of the part of the screen aoe reads, and the session wrongly shows as idle even though it is still blocked on the user. Now a waiting session stays waiting until the screen shows that it moved on: the input box returned, the cancel banner appeared, or work resumed.

- Claude declares its own `hook_waiting` rule, ranked below every positive idle/running shape and above the static title glyph (which also shows while `AskUserQuestion` blocks).
- A hook-inferred Waiting now gets the same stale-shell check as Idle, so an agent that exits from an open prompt does not stay Waiting.

Fixes #3441

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: detection rule change; covered by unit tests over pane fixtures (`test_waiting_hook_claude_survives_question_scrolled_out_of_window`, extended `test_resolve_detected_status_shell_stale_bare_prompt_is_error`).

## AI Usage

- [ ] No AI was used
- [x] AI was used

https://claude.ai/code/session_0154XBBff1XPPvpwsAN2mo9T

## Summary

- Preserve Claude’s `Waiting` status when an `AskUserQuestion` prompt scrolls beyond the recent pane window.
- Clear `Waiting` only when the screen shows positive contradictory evidence, such as resumed work or a returned prompt.
- Apply stale-shell checks to inferred `Waiting` states.
- Add coverage for active, stale, and title-bearing Claude panes.

## Benefits

Claude sessions remain correctly marked as waiting until the user answers or cancels the prompt. Exited sessions do not remain stuck in the waiting state.


jerome-benoit (approved):
Reviewed `6e8870f49ebcb3b68c298423c7b3596029dc3492`, including #3441 and the existing review, with two independent sequential analyses and source arbitration. No blocking correctness finding.

The Claude-only hook override preserves a scrolled-away question while positive resume/cancel/prompt evidence still wins (`claude.toml:190-194`, `manifest.rs:552-570`). Other agents retain their shared hook rule; dead panes and command overrides retain their existing guards (`status_update.rs:363-370,564-575`).

**P3, non-blocking:** CodeRabbit's caller-level test request is valid. The resolver test injects `shell_stale=true` (`pane_status.rs:312-327`), so it would not catch removal of the new caller condition. Please cover inferred Waiting on an exited-agent shell through the status-update path. No duplicate inline comment added.

Evidence: exact-head `git diff`/`git show`, clean `git diff --check`; CI 35 successful checks, one skipped. No local tests or live Claude reproduction rerun; #3441 explicitly qualifies its hook observation as source-derived.

PR #3965, opened 2026-09-14
fix(cli): use the command profile when gating worktree edits

### LINKED ISSUE #3739: Preserve the session profile when checking worktree edit safety
## Problem

`set_worktree_name` refreshes a session's live status before deciding whether its worktree may be moved, but it does not restore the session's runtime-only `source_profile`.

Instances loaded from storage have an empty `source_profile`. Status-rule lookup consequently falls back to the default profile instead of using the profile passed to the command. `rename_session` already restores the profile before performing the equivalent safety check.

## Impact

A running custom agent from a non-default profile can be classified as `Idle` when its profile-local status rules are not consulted. `aoe session set-worktree-name` may then move or rename the worktree while the agent is active.

Priority: P1, narrow but safety-critical.

## Expected fix

Set `live.source_profile = profile.to_string()` before calling `update_status_with_metadata`, matching `rename_session`.

Add a regression using a non-default profile whose local status rule classifies the pane as `Running`, and verify that the worktree edit is rejected.

## References

- #3717
- Existing review thread: https://github.com/agent-of-empires/agent-of-empires/pull/3717#discussion_r3917620834

## AI usage

Drafted with Codex from a cross-validated post-merge review of #3717.

njbrake:
## Description

Before moving a session's worktree, the `set-worktree-name` command checks whether the agent is still working. That check ignored which profile the session belongs to, so status rules defined in a non-default profile were never applied. A custom agent that only those rules recognize as busy looked idle, and its worktree could be moved while it was running. The command now passes its profile into the check, as `rename` already did, so the edit is refused for a busy agent.

Regression test runs a real tmux pane that a profile-local rule classifies as Running and asserts the edit is rejected with the worktree untouched; it fails without the fix.

Fixes #3739

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the unit test in `src/cli/session.rs` drives the subcommand handler against a real tmux pane, which covers the gate more cheaply than a full-binary test.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

Fix, test, and description drafted by the model; reviewed by @njbrake before opening.

https://claude.ai/code/session_011cwq23RLVnyyw2a5Svf3do

_Note: this PR was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

## Summary

- Restores the session’s `source_profile` before refreshing live status in `set_worktree_name`.
- Applies profile-specific status rules when checking whether an agent is running.
- Adds a regression test with a real tmux pane.
- Rejects worktree changes for running custom agents and keeps the worktree unchanged.

## Benefits

This prevents active worktrees from being moved or renamed when non-default profile rules identify the agent as running.


jerome-benoit (approved):
Reviewed `9d3e2a6937fd1a3155ac52dc7a5b2c6ef96e2f39` and #3739 with two independent sequential analyses and source arbitration. No actionable finding.

The profile is restored before the live status check (`src/cli/session.rs:2719-2721`), matching `rename_session`. The regression uses a distinct non-default profile, waits for observable pane output, requires the specific active-session rejection and checks the original path remains intact (`:2532-2629`). The fake worktree is sufficient for this refusal-path test; an unrelated Git error cannot satisfy its assertion.

Verified through exact-head `git diff`/`git show` and clean `git diff --check`. CI: 35 successful checks, one skipped. No local builds/tests rerun.

Issue #3973, opened 2026-09-14
Reused code snippets can keep displaying previous highlighted content


jerome-benoit:
# Reused code snippets can keep displaying previous highlighted content

## Problem

Existing P2 followed through #3958, still present at `f7d0a4ed21ac2c9dfca460c30423f7861966e7e9`. This is not a new regression from the final language-lookup change.

The HTML consumers retain their previous result when highlighting returns `null` or rejects: [Markdown:242](https://github.com/agent-of-empires/agent-of-empires/blob/f7d0a4ed21ac2c9dfca460c30423f7861966e7e9/web/src/components/acp/Markdown.tsx#L242), [ToolCards:684](https://github.com/agent-of-empires/agent-of-empires/blob/f7d0a4ed21ac2c9dfca460c30423f7861966e7e9/web/src/components/acp/ToolCards.tsx#L684), and [DiffCommentsUserCard:67](https://github.com/agent-of-empires/agent-of-empires/blob/f7d0a4ed21ac2c9dfca460c30423f7861966e7e9/web/src/components/diff/comments/DiffCommentsUserCard.tsx#L67). They update state only for a truthy result, and their error paths do not clear it.

Impact: a reused component can display old source/output instead of the current plain-text fallback. No underlying file mutation or data loss is claimed.

## Regression scenario

In Vitest/RTL, retain the same component instance:

1. Render snippet A and resolve its highlighting successfully; await A's highlighted output.
2. Rerender with distinct snippet B and an unsupported hint, returning `null` from the highlighter.
3. Await completion; assert B is displayed as plain text and A is absent.

Current source retains A. A concrete same-instance entry point is a truncated tool-output card: await successful highlighting, click **Show more lines**, then reject the next highlight. The expansion updates `shown`, but old truncated HTML remains and the expansion button disappears ([ToolCards:708-715](https://github.com/agent-of-empires/agent-of-empires/blob/f7d0a4ed21ac2c9dfca460c30423f7861966e7e9/web/src/components/acp/ToolCards.tsx#L708-L715)). A focused RTL test can inject that rejection deterministically. Also cover stale asynchronous completions. These are source-established scenarios; no browser or component test was executed for this report.

## Correction and acceptance

Associate cached HTML with its input identity, invalidate it when that identity changes, apply nullable results and handle errors without allowing obsolete asynchronous results to overwrite newer content. Use a lint-compliant reset pattern and preserve the ANSI fast path.

[...]

PR #3974, opened 2026-09-14
fix(web): clear stale highlighted HTML when a reused snippet fails to resolve

matthewpwatkins:
## Description

Follow-up to #3958. CodeRabbit flagged that `ShikiSyntaxHighlighter` (`Markdown.tsx`), `HighlightedBlock` (`ToolCards.tsx`), and `HighlightedSnippet` (`DiffCommentsUserCard.tsx`) only called `setHtml` on a successful highlight, so a component instance reused for a new snippet whose language fails to resolve keeps painting the previous snippet's highlighted markup instead of falling back to plain text.

Those threads were marked fixed and resolved on #3958, but the fix commit landed after njbrake had already merged it, so it never reached `main` (flagged in jerome-benoit's post-merge review). This PR is that fix, plus a second gap jerome-benoit's review here caught: the fix only cleared `html` on a failed or unresolved highlight, not on the no-language early return (a reused block whose language becomes absent entirely, e.g. an extensionless Read card, or unfenced text after a fenced block).

- `Markdown.tsx` / `ToolCards.tsx`: call `setHtml(out)` unconditionally instead of only on truthy `out`, clear `html` in the catch branch, and clear it at render time via a `syncKey` (matching `FullFileViewer`'s existing pattern) so the no-language early return can't leave stale markup either.
- `DiffCommentsUserCard.tsx`: already used the `syncKey` pattern from an earlier review round; unchanged here.
- Added transition regression tests to all three components' specs: highlight A, reuse the instance with B (null, rejected, or no-language), assert B renders and A's markup is gone.
- Wired `DiffCommentsUserCard.test.tsx` into `coverage-matrix.json` — it already covered the component but was never listed.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Added stale-content transition tests to `Markdown.test.tsx`, `ToolCards.render.test.tsx`, and `DiffCommentsUserCard.test.tsx` (null, rejected, and no-language cases). Added `DiffCommentsUserCard.test.tsx` to the `diff-comments.typed-event` coverage-matrix entry, which already listed the component but not its own render spec.

`npx tsc -b`, `npm run lint`, `npm run format:check`: clean. `npm run test:unit`: 3,835 passed.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Sonnet 5 (Claude Code)

**Any Additional AI Details you'd like to share:** Fix and regression tests drafted in response to jerome-benoit's review, verified locally before pushing.

## Summary

- Prevent stale highlighted HTML when reused snippets receive new, missing, unsupported, or failed language input.
- Ignore late results from superseded highlighting requests.
- Add regression tests for Markdown, tool cards, and diff comments.
- Add the diff-comments test to the web coverage matrix.
- Add a shared test helper for deterministic late-result transitions.

## Benefits

Users see current snippet content instead of outdated markup. Highlighting transitions are safer and more predictable.


jerome-benoit (changes requested):
Reviewed `5a5a30cedbfa60c78411ad27595fd6dd1fa9a423` against #3973. The nullable-result and rejection paths are fixed, and cancellation guards are preserved.

**P2, incomplete in-scope fix:** please address the existing [CodeRabbit no-language finding](https://github.com/agent-of-empires/agent-of-empires/pull/3974#pullrequestreview-5203079463), not a duplicate inline. `ToolCards.tsx:675` returns before clearing cached HTML; `:699-702` still renders that cache. An extensionless Read card (`:872,898`) has no fallback language: if a reused result changes from a language-tagged fence to unfenced text, the old highlighted content remains. Synchronize cached output with the rendered input before the early return. This is preexisting, not a new regression, but remains within the stale-content issue this PR closes. Apply the same invariant to `Markdown.tsx:233`.

Please add focused transition regressions: first obtain highlighted A, then reuse the instance with B and test null, rejection and absent language; assert B is visible and A absent. Preserve cancellation behavior. The current ToolCards mock always returns null (`ToolCards.render.test.tsx:20`), while the diff-card test only checks initial success (`DiffCommentsUserCard.test.tsx:158-167`), so those tests do not establish this fix. Update the applicable web coverage-matrix entry and the PR's dashboard N/A declaration.

No additional blocker from ANSI rendering (it takes priority over HTML), theme-only reset omission (styling rather than changed content), or a theoretical delimiter collision without a practical caller case. Two independent sequential source reviews and parent arbitration; exact-SHA `git show`/`git grep` and clean `git diff --check`. No local tests or runtime reproduction executed.

CI follow-up at this same head: `codecov/patch` reports **36.36% against 75% required**, consistent with the missing regression coverage. Playwright live shard 1 failed at `rate-limit-reset-across-turns.spec.ts:105` waiting for the reset-time banner; this alone does not establish a highlighting regression. Approved the initially blocked workflows and requested one targeted rerun of that failed job to check reproducibility. No passing rerun is claimed.

matthewpwatkins:
@jerome-benoit good catch, and you're right that the earlier fix was incomplete. Pushed 902662d2:

- `Markdown.tsx`/`ToolCards.tsx` now use the same render-time `syncKey` reset `DiffCommentsUserCard` already had, so the no-language early return (extensionless Read card, unfenced text after a fenced block) clears `html` instead of leaving it untouched.
- Added transition regression tests to all three specs (highlight A, reuse with B — null, rejected, and no-language — assert B renders and A's markup is gone).
- Added `DiffCommentsUserCard.test.tsx` to the `diff-comments.typed-event` coverage-matrix entry and fixed the PR's dashboard checklist, which was wrongly marked N/A.

`codecov/patch` should clear now that the DiffCommentsUserCard lines have a direct test. The `rate-limit-reset-across-turns.spec.ts` failure looks unrelated to this change (rate-limit banner timing, not highlighting) — let me know if the rerun says otherwise.

jerome-benoit (changes requested):
Re-reviewed `902662d2e60592045243030e5786be2eecaf6991`, including your reply. Thanks: the original no-language reset is fixed, and the transition tests and coverage-matrix update address the earlier feedback.

**P2, still blocking completion of #3973:** I reproduced [CodeRabbit's remaining stale-completion concern](https://github.com/agent-of-empires/agent-of-empires/pull/3974#pullrequestreview-5210234384), so no duplicate inline comment.

Reproduction: extract the exact `ShikiSyntaxHighlighter` from this SHA, transpile its TSX, and mount it with React 19.2.8 in jsdom. Mock `highlightSnippet` with a deferred promise; mount A with `language="rust"`, let its effect start, then use `startTransition` to render B with no language. Resolve A to `<pre>OLD_A</pre>` in the parent's layout effect for B, after B commits but before passive cleanup. **3/3 runs: DOM A → B → OLD_A.** Resolving A from the parent's passive effect instead leaves B displayed throughout the 2-second control window. This is a component reproduction with a mocked highlighter, not a full browser or real-Shiki run.

`Markdown.tsx:254-255` admits A before `cancelled` is set at :262. The input key already equals B, so the render-time reset does not run again; B's no-language effect cannot repair the stale HTML. Source inspection finds the same mechanism in `ToolCards.tsx:695-704` and `DiffCommentsUserCard.tsx:78-86`; those two were not individually reproduced. The added tests finish A before changing inputs, so they miss this ordering.

Associate cached HTML with the input identity that produced it and render it only while that identity matches current inputs, or provide an equivalent commit-safe invalidation. Retain cleanup and cover pending A → committed B → late A with deterministic deferred-promise tests. A plain RTL rerender followed by resolution may already have flushed cleanup and will not establish this window.

Two independent sequential reviews plus parent reproduction/arbitration. One reviewer initially withheld the finding for lack of a repro, then accepted this evidence. The previously reported Playwright failure passed on its old-head rerun. I approved the new-head workflows: CI has passed; Tests and CodeQL are still running. No full local suite rerun and no claim yet about new-head coverage.

matthewpwatkins:
@jerome-benoit thanks for the reproduction — confirmed, and fixed in e6e41c1d.

The fix is the commit-safe invalidation you suggested: each component now caches the highlighted HTML together with the input key that produced it and renders it only while that key matches the current inputs, so a late resolution from a superseded request writes an entry that's inert at render time regardless of whether its cleanup ran. Theme stays out of the key deliberately (a theme switch keeps the old palette until the re-highlight lands, per your earlier note that theme-only staleness is styling, not content).

On the tests: you were right that a plain RTL rerender can't establish the window — `act` flushes the passive cleanup before the resolution's microtask every time (I verified: with that recipe the old code passes). The new tests replicate your setup inside vitest: raw root outside the act environment, B rendered via `startTransition`, A resolved in a layout effect at B's commit. Verified all three fail against the previous commit (stale HTML visible) and pass with the keyed cache.

jerome-benoit (approved):
Re-reviewed `902662d2..2aaad4f4`; my CHANGES_REQUESTED is resolved, approving.
The stale-completion P2 is fixed: Markdown.tsx, ToolCards.tsx and DiffCommentsUserCard.tsx now store `result:{key,html}` keyed by the producing inputs and render the html only while `result.key === current inputKey`, so a late write from a superseded request is inert regardless of effect-cleanup timing; the cancelled guard is retained. The requested deterministic transition tests (pending A -> committed B -> late A) were added via a shared `web/src/__tests__/lateResolution.tsx` harness across the Markdown/ToolCards/DiffCommentsUserCard specs; the coverage-matrix already registers them.
Non-blocking (P3): Markdown.tsx's inputKey `${language ?? ''}${code}` dropped the space separator the sibling components keep; a delimiter (NUL/newline) removes a purely theoretical collision, same class I previously deemed non-blocking.
Evidence: exact-SHA `compare 902662d2...2aaad4f4` (2 commits) plus reading the three components and the harness; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI reports checks green on this head (incl. codecov/patch); this approval concerns the code review only.

matthewpwatkins:
@jerome-benoit applied your P3 in fc5b5d41, with one twist worth knowing: the markdown key hadn't lost its separator -- 2aaad4f4 delimited it with a *raw NUL byte*, which is invisible in review tooling (hence the quote in your review showing no separator) and made git treat `Markdown.tsx` as a binary file, hiding its diffs from GitHub review -- the same trap this PR's description flags on `FullFileViewer.tsx`. All three components now use the `\u0000` escape sequence as the delimiter instead: same unambiguous key at runtime, and the source stays a text file.

matthewpwatkins:
Correction to my comment above: #3974 was merged 19 minutes before fc5b5d41 was pushed, so that commit never landed here. The delimiter fix is now #4007, cherry-picked onto current main.

PR #3977, opened 2026-09-15
fix(tests): send Escape as CSI-u so the e2e input fence survives

njbrake:
## Description

Fixes the `new_session::test_new_session_dialog_escape_cancels` flake on main.

The harness follows each fenced key with an F12 fence (`ESC [ 2 4 ~`). A bare Escape is the single byte `ESC`. When the TUI reads both in one read, crossterm decodes `ESC ESC` as one Esc and the fence arrives as literal `[24~`, so the ack never comes. Introduced with the fence in #3807; all 20 fenced `send_keys("Escape")` calls are exposed.

`send_keys` now sends Escape as CSI-u (`ESC [ 2 7 u`), which is self-delimiting and decodes to the same key. `harness::tui_input_barrier_survives_escape_in_same_read` forces the coalesced read: 5/5 fail with raw `ESC`, 5/5 pass with the fix. Stress, 600 runs each: 2 failures before, 0 after.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

Diagnosis, fix, and stress runs by Claude Opus 5 via back-and-forth with @njbrake.

**NOTE:**

🤖 Generated with [Claude Code](https://claude.com/claude-code) using Claude Opus 5

## Summary

- Encode `Escape` and `C-[` with self-delimiting CSI-u sequences.
- Prevent flaky decoding when Escape is combined with the F12 input fence.
- Add a regression test for coalesced input reads.
- Reuse `send_hex_keys` for paste input.
- Improve end-to-end harness reliability. Stress testing reported zero failures after the change.


jerome-benoit (approved):
Reviewed `8c58d888f6b3e77ac36f479da933c8a77059418b`, including the `C-[` and paste-helper follow-up. No blocking finding.

`tests/e2e/harness.rs:724-730` preserves the input fence while sending Escape/C-[ as `ESC[27u`. The locked crossterm 0.29.0 parser and its `test_parse_basic_csi_u_encoded_key_code_special_keys` explicitly map this to an unmodified Esc press. The paste refactor preserves the bytes and tmux arguments. The regression establishes the open dialog and waits for the rendered F12 acknowledgement before asserting dismissal (`:1285-1294`).

**P3, non-blocking:** soften the comment at `:1289`: grouping bytes in one tmux command does not guarantee one terminal read ([read(2)](https://man7.org/linux/man-pages/man2/read.2.html)). This test checks grouped input and acknowledgement, not the read boundary itself. Also, it calls `send_hex_keys` directly, so the existing public `send_keys` tests remain important. No extra timing-based assertion is needed.

Two independent sequential source reviews and parent arbitration; exact-SHA `git show`/`git diff --check` and locked dependency inspection. No local build or test run. The 5/5 and 600-run results are the author's evidence, not mine. New-head CI is still running; this approval does not certify its eventual result.

Issue #3376, opened 2026-08-14
track native conversation resume integrations blocked on automatic id capture


jerome-benoit:
## Contract

AoE supports conversation resume for an agent only when all of these are true:

- the current native resume invocation is verified against the real CLI;
- the active agent conversation ID is captured automatically and persisted as `agent_session_id`;
- capture is correlated to the exact AoE session, including concurrent sessions for the same project;
- restart passes the captured ID back to the agent and demonstrably continues the same conversation.

A manual `aoe session set-session-id` pin does not satisfy this contract. Until the complete path exists, the agent must use `ResumeStrategy::Unsupported` and AoE must not advertise resume or start a resume poller.

## Affected agents

| Agent | Native resume surface | AoE state while capture is missing | Missing integration |
|---|---|---|---|
| Cursor | `--resume {id}` | `Unsupported` | Verify the current CLI and implement automatic capture |
| Droid | `--resume {id}` | `Unsupported` | Verify the current CLI and implement automatic capture |
| Antigravity | `--conversation {id}` | `Unsupported` | Verify the current CLI and implement automatic capture |
| Qwen | `--resume {id}` | Move to `Unsupported` under #3370 | Implement automatic and retroactive capture with safe correlation |
| Kiro | `--resume-id {id}` | Move to `Unsupported` under #3370 | Implement automatic and retroactive capture with safe correlation |

The native flags establish upstream capability, not AoE support.

## Required work

For each agent:

1. Verify the resume form against a current real installation.
2. Locate a stable hook or conversation store and implement project-agnostic ID capture.
3. Use the existing ownership and exclusion mechanisms so sessions cannot cross-claim conversation IDs.
4. Wire live and retroactive capture where applicable.
5. Replace `Unsupported` only when the complete path is implemented.
6. Update every derived recovery, poller, warning, test, and documentation surface in the same change.

## Acceptance criteria

- A new session automatically persists its own conversation ID without a manual pin.
- Restart uses that ID and continues the same native conversation.
- Two sessions using the same agent in the same project never capture each other's IDs.
- A session without a captured ID starts fresh and emits no resume argument.
- Table-driven unit tests cover argument construction and unsupported behavior.
- A deterministic end-to-end test covers capture, persistence, restart, and same-project isolation.

## Out of scope

- Enabling resume only for manually pinned IDs.
- Treating a known CLI flag as sufficient support.
- Agents outside the table above.
- Sandbox-specific resume gaps.

## Related work

[...]

PR #3978, opened 2026-09-15
test(cli): cover the structured-import claim in already_imported

blueagledev:
## Description

`already_imported` decides whether a re-import would create a second row claiming a conversation
that is already claimed. A row can claim one in three places, and the function checks all three:
a terminal import pins `resume_intent`, the poller writes `agent_session_id`, and a structured
import seeds `acp_session_id` while leaving `resume_intent` at `Default`.

`already_imported_matches_resume_and_observed_ids` covered the first two. **The `acp_session_id`
arm could be deleted with the whole suite still green**, which is how I found it: I disabled that
arm before touching the test, and the existing case stayed green.

Folded the third case into a table rather than adding a test fn, so the test-fn count for this
behaviour is unchanged. The name changes with it, since the old one says "resume and observed"
and it now covers all three.

`+20/-7` in one file, entirely inside `mod import_tests`. No production change.

Related to #3376: `already_imported` is the ownership check that Required-work item 3 leans on,
though this is a test-coverage gap rather than a capture gap and moves nothing off
`session_support: None`.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass. Six pre-existing environment failures, unchanged by this
      commit and detailed under Verification.
- [x] Documentation was updated where necessary. No doc surface changes; this is test-only.
- [ ] For UI changes: included screenshot or recording. N/A, no UI change.

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## Verification

Cut against `main` re-fetched immediately before the cut rather than against an inherited ref:
`507b2386`, from `git ls-remote` at 2026-09-15T13:49:49Z. `main` moved three times while this was
being prepared, so any earlier reading is not what this was built on.

Against `8074f586` on base `507b2386`, clean working tree, cargo 1.95.0:

```
cargo fmt -- --check          clean
cargo clippy --all-targets    2 warnings, both pre-existing and in files this commit does
                              not touch: src/acp/event_store.rs:1438 collapsible_match,
                              src/session/instance/accessors.rs:330 nonminimal_bool

cargo test -p agent-of-empires --no-fail-fast          branch 8074f586      base 507b2386
  lib                                                  6968 passed;  6 fail  6968 passed;  6 fail
  tests/branch_exists_spawn_failure                        1 passed;  0 fail
  tests/filewatch_degradation                              1 passed;  0 fail
  tests/integration                                      289 passed;  1 fail   289 passed;  1 fail
```

`--no-fail-fast` matters here: plain `cargo test` stops at the first failing target, so on a tree
where the lib target has ambient failures it never reaches `tests/integration` at all.

The integration failure is `acp_runner_control::runner_proxies_agent_requests_over_the_control_channel`,
identical on base and unrelated to this commit.

The lib failures are not this sandbox lacking a runtime: `tmux 3.6a` and a container runtime are
both installed here. At least one of them,
`tmux::tests::timed_out_agent_probes_release_waiters_and_invalidation`, panics at
`src/tmux/mod.rs:3110` on `Command::new("/bin/ps")`, and NixOS has no `/bin/ps`. That is the same
class of FHS assumption as #3421, so I suspect the others are too; I have not chased each one and
am not claiming I have.

One caution on those lib numbers, because I would rather you had it than not. **The failing set is
not stable between runs of the same tree.** This work has been observed at `6968/5`, `6967/6` and
`6968/6` across rebases, the difference being
`tmux::session::tests::test_status_checks_target_pane_zero_with_split_panes`, which passes in
isolation on both trees and lives in a file this commit does not touch. So the pairing above is one
observation of a flapping quantity rather than a stable equality, and I am not going to present it
as one. What does hold is that this commit adds no test functions, so the passed count should not
move.

**The control, which is the reason I think this is worth your time.** With the `acp_session_id` arm
of `already_imported` disabled, the widened case fails on `id-3`; with it restored, it passes.
Before the case was widened, that same control passed, which is what established the arm was
unguarded.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Code, running Claude Opus 5.

Drafted by an agent, then attacked in fresh context by separate agents that did not see the
earlier passes. Four rounds, and the reason this PR is one test is that each round found the
previous round still claiming more than it could support. It began as a contested-binding guard
plus 21 assertions. Review established that the guard's in-memory clear does not survive
`reconcile_from_disk`, so it was withdrawn; that a draft credited this project with independently
arriving at the `live_idle` gating of `decide_disk_refresh` when #1740 had it two months earlier;
and finally, by breaking each test's subject and requiring the test to notice, that 11 of the 14
remaining assertions could not fail. This is the one that could. I review the result and decide
what ships, and I will answer review questions here myself rather than pasting them into a model.

## Summary

- Expanded the `already_imported` CLI test to cover all three conversation-claim sources.
- Converted the test to a table-driven case without adding a new test function.
- Confirmed that unclaimed IDs still return `false`.
- No production code or documentation changed.

## Benefit

The test now provides broader coverage for conversation imports and helps prevent regressions in claim detection.

## Verification

Formatting passed. Existing Clippy warnings and environment-related test failures were unchanged. A control confirmed that the new `acp_session_id` case fails when its check is disabled.


jerome-benoit (approved):
Reviewed `8074f58614022c3270bb88bfa5887936b73e4cfe`: no blocking finding.

The added `id-3` case isolates the structured claim (`src/cli/session.rs:3715-3726`): neither resume intent nor observed ID can satisfy it, so removing the `acp_session_id` branch at :926 would fail the assertion. Existing positive and negative cases remain covered in one synchronous table test. This tests the import exclusion predicate, not native capture or the broader guarantees in #3376.

Two independent sequential reviews and parent arbitration; exact-SHA `git show`/`git diff --check`. No local tests rerun. CI workflow passed; Tests and CodeQL are still running after authorization.

PR #3980, opened 2026-09-15
fix(web): show a compact usage hint on mobile

### LINKED ISSUE #3916: Web: the structured-view usage hint is hidden below `sm`, leaving mobile with no usage surface
Split from #1719, which stays on the upstream-blocked `/usage`-style quota meter. This half is not blocked: context-window usage already flows end to end.

## Problem

`UsageHint` is the only usage surface in the structured view, and it is `hidden sm:inline-flex` (`web/src/components/acp/Composer.tsx:1655`). Below the `sm` breakpoint it renders nothing, so mobile users have no way to see context-window pressure or session spend at all.

## What's there today

`sessionUsage` (used, size, optional rebased cost) is applied by the reducer and passed to `Composer`, which renders it as text in the composer actions row. The data is present on mobile; only the element is hidden.

## Ask

Keep a usage surface visible at mobile widths. A compact variant of the existing hint is enough; the progress-bar affordance from #1719 is optional here and can follow with the quota meter.

Keep the math as is: no changes to `applyEvent`, and preserve the #1354 rebased-cost behavior after `/clear` and `/compact`.

## Tests

- Vitest + RTL: renders with and without cost, and carries its accessible label.
- The mocked Playwright mobile spec: the surface is present and readable at a small viewport when usage exists.

## Out of scope

- The quota meter (5-hour and weekly limits, reset times), which needs limit data that does not flow over ACP. Stays in #1719, blocked on agentclientprotocol/claude-agent-acp#625.

njbrake:
## Description

On a phone, the chat composer showed no information about how full the agent's context window was or how much the session had cost, because that small readout was hidden on narrow screens. It now stays visible at every width. On narrow screens it shows only the percentage and the cost (for example "60% · $0.42"). When the controls row is full, the readout moves onto its own line rather than pushing the Stop and Send buttons off screen. Wide screens keep the full token counts.

Usage math is unchanged; no reducer changes.

Screenshot (360px): _to be attached_

Fixes #3916

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used


jerome-benoit (commented):
Post-merge audit of ba9a119, cross-validated in two independent passes:

- **[P2] Keep the mobile usage values above AA text contrast.** The newly visible percent and spend retain `opacity-70` in [ComposerControls.tsx:307-311](https://github.com/agent-of-empires/agent-of-empires/blob/ba9a119baeb4f3730298f517ff7c8dc7a6151eb9/web/src/components/acp/ComposerControls.tsx#L303-L312). On the default Zinc theme, `text-text-dim` is `#8b8b94` and the composer background is `#27272a`; 70% opacity composites to `#6d6d74`, giving **2.90:1** for 11px text versus [WCAG AA 4.5:1](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html). Below `sm`, token counts are hidden, so these are the only visible usage values. Use a mobile foreground with at least 4.5:1 contrast against the composer surface across themes and add a contrast-aware mobile assertion. Merely removing opacity is insufficient for Zinc on this surface (4.41:1).

The separate ARIA-label concern is already tracked in [CodeRabbit's thread](https://github.com/agent-of-empires/agent-of-empires/pull/3980#discussion_r4085224870); I am not reposting it. The affected code remains on main at ccbe389. This contrast result comes from the checked-in colors and opacity calculation; I did not run a local browser or test suite.

PR #3984, opened 2026-09-15
fix(hooks): ignore session-id hooks from nested agent processes

### LINKED ISSUE #3773: session.store: "fresher live observation" heuristic overwrites a session's tracked sid with an unrelated Claude Code subagent transcript
## Symptom
An aoe-managed Claude Code session's tracked transcript (`agent_session_id`) can get silently overwritten with a *different* conversation's transcript — specifically, a leftover transcript from an in-process `Agent`-tool subagent that the session itself had dispatched earlier. No error is surfaced; the session just starts resolving to the wrong conversation.

## Confirmed mechanism
`session.store` logs the exact swap:
```
session.store: Replacing stored session id with fresher live observation stale=<uuid> fresh=<uuid> tool=claude
```
Reproduced against a real session: the "fresher" transcript picked was one whose first line is `{"type":"agent-setting","agentSetting":"general-purpose","sessionId":"<uuid>"}` — Claude Code's own marker for a subagent transcript that happens to sit as a plain sibling file in the same project directory as the real, top-level transcript (both are keyed by the same launch cwd). This happened during a `session.startup_recovery` pass, though separately-confirmed instances of the same log message show the underlying "pick freshest, no ownership check" comparison also runs continuously outside of restart/recovery, not only during it.

## Why this looks like a fixable gap, not a new capability request
aoe already has a similar validity check elsewhere in the same subsystem, at launch time:
```
session.store: stored Claude sid has no transcript on disk; launching fresh with --session-id instead of --resume to avoid a certain resume failure
```
That's an existence check (does a transcript file exist for this sid at all). The "freshest observation" comparison that causes this bug doesn't appear to do the cheaper, more targeted check of peeking at a candidate transcript's first line for the `{"type":"agent-setting",...}` marker before adopting it — a signal Claude Code already writes for exactly this purpose.

## Suggested fix
When comparing candidate transcripts for "freshest observation," skip (or deprioritize) any whose first line is `{"type":"agent-setting",...}` — this loses nothing for genuine top-level sessions and would have prevented the swap described above.

## Note on reproducibility with the current Claude Code CLI
As of Claude Code CLI 2.1.261 (2026-09), subagent transcripts appear to have moved to a nested `<parent-session-id>/subagents/agent-<id>.jsonl` layout rather than sitting as flat siblings in the project directory — so this exact flat-sibling shape may no longer arise on current CLI versions. The underlying "no ownership check" gap in the freshness comparison itself is still real and worth fixing regardless, since a future Claude Code layout change, or an older CLI version, could reintroduce the flat-sibling shape at any time.

## Environment
- aoe 1.15.3
- Claude Code CLI 2.1.261 at time of the nested-layout observation (the collision itself was observed against whatever CLI version was installed 2026-09-03–04, not independently pinned)

njbrake:
## Description

When a program inside an aoe session started its own copy of Claude Code (from Claude's shell tool, a script, or `claude -p`), that inner copy reported its conversation as the session's own, and aoe then tracked and resumed the wrong conversation. aoe now remembers which process it launched and ignores identity reports from another copy of that agent running beneath it.

- Host launch also exports `AOE_AGENT_PID=$$` (the agent keeps that pid through `exec`) and `AOE_AGENT_BIN`.
- `__extract-session-id` walks its ancestors to `AOE_AGENT_PID` and skips the write if the path holds more than one `AOE_AGENT_BIN` process or never reaches that pid.
- The sandbox `jq` hook applies the same count through `/proc` to the container root.
- Not caught: a nested agent of a different kind, a renamed agent binary, or an agent whose process shows as `node`. Panes launched before this change keep the old behavior until restarted.
- If the process table cannot be read at all, the hook writes as before.

Fixes #3773

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: ownership depends on the process tree, not the TUI. A table test covers the host walk and a shell test runs the sandbox hook under a fake nested agent; both fail without the check.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**
Claude Opus 5 via Claude Code

**NOTE:**

## Summary

- Prevents nested Claude Code processes from replacing the parent session’s tracked conversation ID.
- Adds process-tree checks for host and sandbox hooks.
- Adds `AOE_AGENT_PID` and `AOE_AGENT_BIN` to launch environments.
- Preserves the previous behavior when process information cannot be read.
- Adds process inspection helpers and coverage for nested-agent scenarios.
- Updates environment-isolation tests for the new variables.

## Benefits

This prevents session recovery from selecting an unrelated nested-agent transcript. Existing panes require a restart to use the new behavior.

## Technical notes

The change currently targets agents launched with the expected binary name. It does not cover renamed binaries, `node`-based agents, or nested agents of other types. Test execution results were not provided.


jerome-benoit (approved):
Reviewed `0f1186bc5a186c85ea06f91d459fea0153028ad3`: no blocking finding in the stated same-agent, identifiable-binary scope.

The host walk is anchored to the launched PID; the sandbox hook counts matching ancestors before writing. Direct/wrapped and nested cases are covered by the table test and the real shell-hook fixture. The declared limitations (including `node`/renamed binaries) remain limitations, not claims of universal conversation ownership. The current Claude freshness override reads the hook sidecar (`session_id.rs:684-694`), so the issue's old transcript-scan hypothesis is not the current path.

**P3, non-blocking residual:** `process/macos.rs:372-374` splits `ps args` on whitespace. If argv0 contains a path such as `/Applications/Agent Tools/claude`, it is truncated before the basename check and same-agent nesting can remain undetected. Prefer an unambiguous executable/argv lookup and cover paths containing spaces. This is a static parser limitation, not a reproduced macOS regression; it does not block the narrower improvement here.

Evidence: exact-SHA `git diff`/`git show`, `extract_session_id.rs:51-98`, `instance/hooks.rs:16-22`, `hooks/mod.rs:449-455` and its nested-agent fixture. Two independent sequential analyses and parent arbitration; CI, Tests and CodeQL passed on this head. No local tests rerun.

PR #3987, opened 2026-09-15
fix(sandbox): recreate a reused container built for a different tool

### LINKED ISSUE #3976: fix(sandbox): recreate a reused container built for a different tool
A sandbox container reused after a tool swap keeps the previous tool's agent config mounts. `get_container_for_instance` (`src/session/instance/container.rs:113`) checks the store generation and shared-credential mounts before reuse, but not which tool the container was built for.

Structured-view sessions hit this. The TUI restart dialog allows a tool swap, but #3975 skips container removal for them, because they never take the Launch reservation that removal now waits for. The next time their runner starts, it reuses the stale container.

Proposed: stamp the tool on the container as a label at creation, the same way `sandbox_store_generation_matches` checks the store generation, and recreate the container on a mismatch. Removal would then happen at relaunch rather than at the restart request.

Related: #3959, #3972, #3975.

njbrake:
## Description

When a sandboxed session switched to a different agent tool, its next launch could reuse the old container, which still carried the previous tool's configuration. Restarts from the terminal view already removed that container, but structured-view sessions skipped that step. Containers now record which tool they were built for, and a launch that finds a container built for another tool removes it and builds a fresh one.

- Label the container with its tool at create.
- `get_container_for_instance` force-removes a container labelled for another tool before reuse. An unlabelled container predates the label and is reused.

Fixes #3976

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: a unit test with a fake runtime covers the reuse decision (other tool removed, same tool and unlabelled kept); an e2e test would need a real container runtime.

## AI Usage

- [ ] No AI was used
- [x] AI was used

## Summary

- Containers now record the tool used to create them.
- The system recreates containers when the recorded tool differs from the requested tool.
- Unlabeled containers remain reusable for backward compatibility.
- The system reloads container state before removal to avoid deleting a replacement container.
- Tests cover matching, mismatched, and unlabeled containers.

## Benefits

This prevents stale agent configuration mounts after a tool swap. It also supports structured-view sessions that previously reused outdated containers.

## Further enhancement

Add runtime-level coverage for label behavior across all supported container runtimes.


src/session/instance/container.rs
```diff
@@ -149,6 +149,21 @@ impl Instance {
                 None
             };
 
+        // A container built for another tool mounts that tool's agent config.
+        // Decide on the disk row: a stale in-memory copy would remove the
+        // container a peer just recreated for the swapped tool.
+        if container.exists()? && container.agent_tool_matches(&self.tool)? == Some(false) {
+            self.reconcile_from_disk();
+            if container.agent_tool_matches(&self.tool)? == Some(false) {
```
jerome-benoit, line 157: [P2] Require a successful disk refresh before forced removal

`reconcile_from_disk()` silently retains `self` on storage-open/load failure (`reconcile.rs:13–33`). If this copy says Claude but a peer has already persisted Codex and recreated its running container, a failed refresh leaves the second comparison using Claude and reaches `remove(true)`, killing the valid container. Require a successful authoritative refresh before this destructive decision; propagate the error or defer removal. Add a fake-runtime case with stale memory, a peer's matching container and a failed storage read, asserting no removal. This is source-confirmed, not a locally executed reproduction.

src/session/instance/container.rs
```diff
@@ -149,6 +149,21 @@ impl Instance {
                 None
             };
 
+        // A container built for another tool mounts that tool's agent config.
+        // Decide on the disk row: a stale in-memory copy would remove the
+        // container a peer just recreated for the swapped tool.
+        if container.exists()? && container.agent_tool_matches(&self.tool)? == Some(false) {
+            self.reconcile_from_disk();
```
jerome-benoit, line 156: [P2] Recompute agent-derived values after replacing the instance

`detect_as` is captured at line 114, before this refresh replaces `self`. With stale alias A detecting as Claude and persisted alias B detecting as Codex, a container labelled B is correctly retained, but the subsequent config/trust calls receive B with the old Claude detection value. `resolve_active_agent` (`container_config.rs:1839–1856`) prioritizes that explicit value over B's configured mapping, so reuse refreshes the wrong agent's configuration. Recompute the derived detection value after reconciliation, and cover an alias-to-alias swap through successful reuse. Built-in-to-built-in swaps do not expose this because the built-in tool lookup wins.

jerome-benoit (changes requested):
Reviewed a2c3fac with two independent sequential reviews and source-level cross-validation. Two P2 corrections are needed in the new reconciliation path (inline). Missing labels remaining reusable is intentional, not a finding. No local tests were run; CI was still running at review time.

njbrake:
Sorry about the auto merge, some hook ran and merged over Jerome's request changes review. Addressing that problem now.

njbrake:
Follow up PR on the way to address @jerome-benoit 's findings. 🙏

njbrake:
Both P2s are fixed in #3989, since #3987 merged before they were addressed.

- **Failed disk refresh:** `try_reconcile_from_disk` returns the storage error, and the launch fails instead of removing the container. A corrupt `sessions.json` case asserts no removal.
- **Stale `detect_as`:** now resolved after the reload. An alias-to-alias swap through reuse asserts only the new agent's store is refreshed.

#3989 also labels aliases with their resolved agent (`tool:agent`), per CodeRabbit's finding on the same check.

PR #3989, opened 2026-09-15
fix(sandbox): close the gaps in tool-mismatch container recreation

njbrake:
## Description

#3987 recreates a sandbox container when the session's tool no longer matches the tool it was built for. Review after merge found three gaps. If the session could not be re-read from disk, the launch still removed the container using an out-of-date copy, which could remove a container another process had just rebuilt. After that re-read, part of the setup still used the old tool's detection, so a reused container could have the wrong agent's config refreshed. And the container label recorded only the tool name, so remapping a custom agent alias to a different agent kept reusing a container with the old agent's mounts. A failed re-read now fails the launch instead, detection is computed after the re-read, and the label records the agent an alias resolves to.

- `try_reconcile_from_disk` reports a storage failure; `reconcile_from_disk` keeps its silent behavior for existing callers.
- The removal check refuses to remove on a failed reload, and `detect_as` is resolved after it.
- The `agent-tool` label is `tool`, or `tool:agent` for an alias. Built-in tools keep their existing label; alias containers labelled by #3987 are recreated once.

Follow-up to #3987, addressing its review.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the fake-runtime unit test covers each case (failed reload, alias swap through reuse, alias label); an e2e test would need a real container runtime.

## AI Usage

- [ ] No AI was used
- [x] AI was used

## Summary

- Improved sandbox container reuse when a session resolves to a different agent or alias.
- Reloaded sessions before detecting the active agent.
- Stopped launches when session reload or agent resolution fails.
- Added agent identity labels for aliases while preserving built-in tool label formats.
- Recreated containers that use the previous alias label format.
- Added tests for reload failures, alias remapping, and alias labels.

## Benefits

This prevents reuse of containers with incorrect mounts and avoids decisions based on stale session state. It also preserves clear errors when configuration resolution fails.

Test results and current review findings were not provided.


jerome-benoit (changes requested):
Reviewed `d8cc243f`. The storage-error propagation and post-reload detection address my two #3987 findings. Recording both alias and resolved agent also addresses its label-identity gap.

**P2, blocking:** [CodeRabbit's profile-resolution finding](https://github.com/agent-of-empires/agent-of-empires/pull/3989#discussion_r4019157228) is valid; no duplicate inline comment. With a readable session whose persisted `detect_as` is empty, an unavailable alias registry and an invalid profile, `resolve_config_or_warn` substitutes defaults. The new identity becomes `alias` rather than the existing container's `alias:claude`. Reloading the session succeeds but does not restore the alias; the second mismatch reaches `remove(true)`, including for a running container. A configuration-read failure must not authorize its destruction.

Make identity resolution fallible and propagate the profile error before removal. Extend the fake-runtime test with a configuration-only alias, empty persisted detection, readable session and invalid profile; assert an error and zero removals. The existing corrupt-sessions case does not exercise this path.

Evidence: exact-SHA `git show` inspection of `container_config.rs:1861-1878`, `profile_config.rs:169-185`, `status_rules.rs:129-138,220-230`, and `instance/container.rs:154-166`. Two independent sequential reviews initially found no new defect; targeted parent arbitration and one reviewer's recheck confirmed this error path. Source-level validation only, no executed reproduction or local tests. CI is green. Missing-row behavior and the existing reconciliation/removal race are not new findings in this review.

jerome-benoit (approved):
Re-reviewed `2297f740` against `d8cc243f`, including the full PR scope. My blocking P2 is fixed: profile resolution now returns an error before removal, and the added regression starts with an empty alias registry, a readable session and an invalid profile, asserting no removal or store refresh.

**P2, non-blocking pre-existing limitation:** [the stale matching-instance case](https://github.com/agent-of-empires/agent-of-empires/pull/3989#discussion_r4019444445) is valid, but the same precheck already bypasses reconciliation in base `8c8a3f61`. It is not introduced by this correction. Follow up by refreshing authoritative state before reuse decisions and testing stale memory/container agreement against a changed disk row; no duplicate inline comment added.

Two independent sequential reviews plus parent arbitration; evidence from exact-SHA `git diff`/`git show` of the identity helper, removal path and regression table. No new blocking regression found. No local tests run. CI is still in progress, not yet a verified green run.

Issue #3968, opened 2026-09-14
test: follow-ups from #3807 (epoch test, duplicated lock observer)


njbrake:
**Epoch test cannot detect an early unlock.** `the_create_bumps_the_mutation_epoch_under_the_instances_lock` (`src/server/session_spawn.rs:625`) runs on a current-thread runtime, and `publish_created_instance` (`:606`) has no await between `drop(instances)` and the `mutation_epoch` bump. Moving the drop above the bump still passes, so a stale reload can drop a new session from `GET /api/sessions` undetected. Use `flavor = "multi_thread"` with a gate after the unlock, or restore the source-order assertion #3807 removed.

**One lock-contention observer, six copies.** try_lock, report `WouldBlock`, then lock: `session/recovery.rs:501`, `session/storage.rs:1329`, `session/test_support.rs:599,657`, `tmux/mod.rs:2679`, `tmux/vt.rs:2062`. Replace with one helper in `session/test_support.rs`.

**Smaller items**
- `src/acp/acp_client/connection.rs:2238`: `FlushForTest` arm misindented; rustfmt skips `select!` bodies.
- `Cargo.toml:206`: the `test-support` comment still says it only re-exports tmux helpers.
- `structured_view/mod.rs:433` resets the e2e input barrier sequence to 0 when entered from remote home, so the harness (`tests/e2e/harness.rs`) would wait for a sequence that never comes. Untested today.
- `tests/integration/acp_runner_orphan.rs:177`: fixed 500 ms sleep before a negative assertion, which the new AGENTS.md rule prohibits.
- `web/AGENTS.md` has no counterpart to that rule; 41 `waitForTimeout` calls remain.

_Note: this issue was drafted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

PR #3991, opened 2026-09-15
fix: six quick-win fixes (#3810, #3630, #3702, #3472, #3731, #3968)

### LINKED ISSUE #3472: TUI/CLI session list: parent/child relationship (-P) is tracked but never displayed
**Note:** This issue was investigated and drafted by Claude Code (an AI agent) at my direction, based on the direct testing described below, which I reviewed before opening this. — @fransonsr

## Problem

`aoe add -P <parent>` creates a real parent/child link between sessions, and it's stored and queryable (`aoe session show <id> --json` returns `parent_session_id`). But nothing in the TUI or web dashboard's session list surfaces it — a child session looks identical to any unrelated top-level session unless you already know to go check `--json` output for a specific id.

## Evidence

Created a sub-session for a real task:

```
$ aoe add <path> -t "my-task" -P 4f7f06cb813749b3 --tool claude --launch
✓ Added session: my-task
  ...
  ID:      a075f20fb8e447d9
  Parent:  4f7f06cb813749b3

$ aoe session show a075f20fb8e447d9 --json
{
  "id": "a075f20fb8e447d9",
  ...
  "parent_session_id": "4f7f06cb813749b3",
  ...
}
```

`aoe add`'s own stdout even echoes `Parent: <id>` at creation time — so the relationship is treated as meaningful information right up until the session lands in the list/dashboard, where it disappears.

Confirmed via GitHub code search (`parent_session_id`, `parent_id`) that both fields appear only in the CLI layer (`src/cli/add.rs`, `src/cli/session.rs`, `src/session/capture.rs`, `src/session/fork.rs`) — zero hits in any TUI or web rendering code. This isn't a setting I'm missing; the display layer genuinely never reads the field.

`aoe list` and the TUI's session table show sessions as a flat list (optionally grouped by repo/group), with no indication that some are children of others.

## Why this matters

`-P` is the documented mechanism for a supervisor/orchestrator pattern where a parent needs to track several dispatched children at a glance — I'm using it for exactly that: an orchestrating Claude Code session dispatching implementation work to a live sub-session. Right now that tracking happens entirely outside aoe; I have to remember which ids I created and match them up myself.

## Proposal (small, TUI/CLI-scoped)

- In `aoe list` / the TUI session table: indent or otherwise visually nest a session under its parent (or add a "Parent: <title>" column/badge) when `parent_session_id` is set.
- `aoe session show <id>` (non-JSON output) could also list any known children, not just the parent.
- Not asking for anything from the web dashboard, cost aggregation, async dispatch, or the broader persistent-hierarchy system proposed in #2124 — that's a much larger, separate proposal. This is scoped to: the parent/child link that already exists in the data model should be visible somewhere in the existing session list, in either the CLI or TUI.

## Related

#2124 proposes a much larger persistent multi-agent hierarchy system (new dispatch primitives, DB schema, web UI tree) where visibility is one small piece of a multi-week roadmap. This issue is narrower: the link aoe *already* creates and stores today isn't shown anywhere in the UI yet.

### LINKED ISSUE #3630: TUI restart runs the sandbox image pull on the UI thread
Restarting a sandboxed session from the TUI blocks the event loop inside `docker pull`. With no local copy of the sandbox image, the UI is frozen until the pull finishes or `PULL_TIMEOUT` (600s) fires. Pressing enter on a session during that window leaves the TUI unresponsive.

Call chain, all synchronous on the TUI thread:

```
src/tui/app.rs:3751                    restart_instance_with_size_opts
src/tui/home/panes.rs:103              -> inst.restart_with_size_opts  (&mut self, no offload)
src/session/instance/resume.rs:61      -> restart_with_resume_policy -> orchestrate_resume_launch
src/session/instance/hooks.rs:46       -> get_container_for_instance  (is_sandboxed branch)
src/session/instance/container.rs:104  -> ensure_image
src/containers/runtime_base.rs:406     -> pull_image, bounded only by PULL_TIMEOUT = 600s
```

Every other caller of `get_container_for_instance` already offloads:

- `src/tui/creation_poller.rs:86` (`thread::spawn`), module doc: "handles the potentially slow Docker operations (image pull, container creation) in a background thread so the UI remains responsive"
- `src/acp/sandbox.rs:78` (`spawn_blocking`), commented "may pull an image ... can take many seconds"
- `src/tui/app.rs:2437`, the banner-triggered pull

**Repro:** `docker image rm ghcr.io/agent-of-empires/aoe-dev-sandbox:latest`, then press `e` on a sandboxed session. That image is 4.21 GiB compressed across 30 layers, so the freeze lasts minutes.

Hit on 1.15.1 after Docker Desktop came up with an empty image store. Six sandbox sessions each fired their own pull of the same image.

Two related points from the same incident, splittable if you'd rather:

- `ensure_image` (`runtime_base.rs:399`) has no single-flight guard, so N sessions restarting means N concurrent pulls of one image. All of mine failed instead of sharing a download.
- Those failures logged `stderr_summary=` empty (`runtime_base.rs:371`). `docker pull` writes progress and some errors to stdout, so the log carried no reason. Exit status plus a stdout tail would make it diagnosable.

### LINKED ISSUE #3702: prime-agent ACP drops source-verified provider credentials
## Problem

Structured-view spawns clear the environment and forward only each adapter's allowlist. `env_allowlist_for("prime-agent")` currently falls through to `&[]`, so `prime-agent --mode acp` cannot use provider credentials exported to the process running `aoe serve`.

Prime was explicitly deferred in #3238 because its variable names had not been source-verified. Upstream now defines them in [`packages/ai/src/env-api-keys.ts`](https://github.com/PrimeIntellect-ai/prime-agent/blob/main/packages/ai/src/env-api-keys.ts), including `PRIME_API_KEY`, the provider-specific API keys, Vertex and Bedrock credential inputs, and `PRIME_TEAM_ID`.

## Required behavior

- Add a Prime-specific allowlist derived from those upstream reads.
- Include companion variables required by Vertex and Bedrock authentication, not only their primary key names.
- Keep unrelated environment variables stripped.
- Update the registry's table-driven allowlist test.

## Acceptance criteria

- `PRIME_API_KEY` present in the `aoe serve` environment reaches a structured-view Prime Agent process.
- An unlisted sentinel variable does not reach that process.
- Tests pin the complete source-verified Prime provider-auth allowlist.

Related: #3238

### LINKED ISSUE #3731: Align remote token cleanup with the actual grace deadline
## Problem

#3722 made remote token cleanup use the configured grace duration, but the cleanup timer still starts later than token validation's grace window.

`TokenManager::rotate()` sets `grace_expires` immediately. `remote_rotation_loop()` then rewrites `serve.url` and awaits the first push-subscription prune before starting `sleep(grace)`. Consequently, the previous token is rejected at `rotation + grace`, while its state and subscriptions are removed at `rotation + asynchronous work + grace`.

During that excess window, push delivery snapshots every stored subscription without revalidating its owner hash. A subscription owned only by the expired token can therefore still receive session status and title metadata. The normal delay is likely small, but filesystem or push-store I/O can extend it.

## Expected behavior

Previous-token state and subscriptions must be removed at the same absolute deadline used by token validation, regardless of work performed after rotation.

## Acceptance criteria

- Derive cleanup from the absolute grace deadline established by `rotate()`, for example with `sleep_until(deadline)`, instead of sleeping for a fresh grace duration after asynchronous work.
- Add a deterministic paused-clock test that delays a post-rotation operation and asserts cleanup at the validation deadline.
- Preserve current-token subscriptions throughout rotation and previous-token subscriptions only until that deadline.
- Keep shutdown cancellation behavior unchanged.

Related: #3662, #3722

### LINKED ISSUE #3810: shorten_path abbreviates paths that only share a prefix with $HOME
`shorten_path` abbreviates any path whose string starts with `$HOME`, not just paths under it, so a sibling directory whose name merely begins with the home directory's name is displayed as if it lived inside home.

With `HOME=/tmp/xyz`, the path `/tmp/xyzextra/not/home` renders as `~extra/not/home`. It is not under `/tmp/xyz`; the names simply share a prefix.

Both branches of `shorten_path` (`src/tui/components/preview.rs`) use a bare prefix match with no separator check:

```rust
if let Some(stripped) = path_str.strip_prefix(home_str) {   // canonicalized branch
    return format!("~{}", stripped);
}
...
if let Some(stripped) = path.strip_prefix(home_str) {       // raw fallback
    return format!("~{}", stripped);
}
```

Abbreviating should require either an exact match with `$HOME` or a separator immediately after it.

This is long-standing, and the existing test asserts the current output rather than catching it: `test_shorten_path_similar_prefix_not_home` expected `~extra/not/home`. The name says the path is not home while the expectation says it is abbreviated as if it were, so it is worth deciding which was intended. PR #3760 moved that case into a table without changing the expectation, since it is test-only.

Whoever picks this up: the fix changes what the TUI displays, so it wants a look at whether anything downstream parses the `~` form.

Found by CodeRabbit on #3760.

njbrake:
## Description

Six small fixes, one commit each. Folders beside the home folder are no longer shown as inside it. Opening a stopped sandboxed session no longer freezes the terminal UI during an image download. Prime Agent in the structured view can use provider credentials from the `aoe serve` environment. `aoe list` and `aoe session show` display parent and child sessions. A rotated remote token is cleaned up when it stops working. Two weak tests are tightened.

- Fixes #3810
- Fixes #3630
- Fixes #3702
- Fixes #3472
- Fixes #3731
- Part of #3968: the epoch test and the duplicated lock-contention observer; its smaller items remain open.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: each fix is covered by a unit test that fails without it (path abbreviation, Enter restart queueing, allowlist and spawn env, list nesting and show lines, rotation deadline on a paused clock, epoch ordering).

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Opus 5 (1M context) via Claude Code

**Any Additional AI Details you'd like to share:** Implemented from the issues under njbrake's direction.

## Summary

- Corrected `~` path shortening so it only applies to `$HOME` and its descendants.
- Moved sandbox restart and image-pull work off the TUI thread.
- Added a source-verified Prime Agent environment allowlist, while keeping host-only credential paths out of sandboxes.
- Added parent/child session details to `aoe list`, JSON output, and `aoe session show`.
- Aligned rotated-token cleanup with the exact validation grace deadline.
- Strengthened epoch-ordering and lock-contention test coverage.
- Added restart-flow handling for asynchronous attach operations and clearer worker failure dialogs.

## Benefits

- The TUI stays responsive during sandbox restarts.
- Session relationships are easier to understand.
- Prime Agent receives required credentials without exposing unrelated variables.
- Token cleanup occurs at the expected security deadline.
- Path displays avoid misleading abbreviations.

## Technical notes

- Added shared `collapse_home` and `collapse_tilde` helpers.
- Added Prime Agent’s 41-key environment allowlist.
- Added cycle-safe, depth-first child-session nesting.
- Changed token rotation to return an absolute deadline and use `sleep_until`.
- Added shared test helpers for reporting lock contention.


jerome-benoit (approved):
Reviewed `62afafbf` with two independent sequential analyses and final arbitration. No blocking finding. Six independent fixes, one commit each, cleanly separated by file with no undue coupling; each traced against head.

- #3810 collapse_home (src/util.rs): separator-aware, strips only when path==home or a '/' follows, fixing the '/home/userextra' -> '~extra' bug; trailing-slash and sibling cases covered.
- #3630 Enter restart off the UI thread: restart_then_attach queues on the existing serial restart worker, attach deferred to the tick loop via take_restarted_attaches gated by launched_agent(); failures surface through the worker's existing InfoDialog; ResumeFailed correctly excluded.
- #3702 prime-agent env allowlist: single source in install_hints::env_allowlist_for, whole list pinned by test, spawn test proves PRIME_API_KEY forwarded and an unlisted sentinel is not; the four AWS *_FILE path vars are in both the allowlist and is_host_only_path_env, while inline/endpoint AWS vars are correctly not host-only.
- #3472 parent/child nesting: nest_children is cycle-, self-parent- and orphan-safe (placed set + two passes); relationship_lines names parent/children by title.
- #3731 token cleanup: rotate() returns grace_expires and both loops sleep_until that absolute instant, aligning cleanup exactly with the deadline validate() enforces; the rewritten oracle stalls the post-rotation prune and asserts cleanup by the validation deadline (the old fresh-grace behavior would miss it).
- #3968 test tightening: shared lock/write_reporting_contention helpers preserve prior semantics.

Verified/rejected: nest_children caller types line up (Vec<&Instance>); grace() removed with no dangling callers; the epoch-order test's drop concern is outside the sliced body.

Non-blocking (P3): (1) #3630 keeps the Enter path's launch hooks unbounded on the shared serial restart worker (src/tui/home/panes.rs:144-150), so a hung hook can stall subsequent restarts; documented and strictly better than the prior inline-on-UI-thread freeze, flagged for awareness. (2) #3968's publication_bumps_the_epoch_before_releasing_the_instances_lock (src/server/session_spawn.rs:730-756) asserts source text via include_str! rather than runtime ordering; brittle to refactor but a deliberate choice given the current-thread runtime has no await at the unlock. Consider a fragility comment.

Evidence: exact-SHA source tracing of the head diff and per-fix touched files; two independent sequential analyses with final arbitration. One env-var-name check against prime-agent's upstream package was not executed. No local build or test run; source-confirmed, not executed. No existing reviews or threads to arbitrate. CI reports all required checks green on this head; this approval concerns the code review only.

PR #3992, opened 2026-09-15
fix(sandbox): reload the session before deciding to reuse its container

njbrake:
## Description

Before reusing a sandbox container, the launch compared the container with its own copy of the session, which can be a few seconds out of date. If that old copy still named the previous tool and the container had been built for it, the launch reused the container without checking the saved session, so the new tool ran against the old tool's configuration. The launch now re-reads the saved session before deciding whether to reuse or recreate the container. If that re-read fails, an existing container can still be reused, but it is never removed.

- `get_container_for_instance` reloads the row whenever the container exists, then makes one identity decision; removal still requires a successful reload.
- A failed reload with a matching container logs a warning and reuses it.

Follow-up to #3989, from its review.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the fake-runtime unit test covers both cases (stale copy agreeing with the container while the saved row changed; failed reload with a matching container); an e2e test would need a real container runtime.

## AI Usage

- [ ] No AI was used
- [x] AI was used

## Summary

- Reloads saved session data before reusing an existing sandbox container.
- Prevents stale session data from selecting the wrong tool configuration.
- Reuses matching containers with a warning when reload fails.
- Preserves containers when reload fails.
- Adds tests for stale-session and failed-reload cases.

## Benefits

This improves container reuse safety and avoids unnecessary container removal.


jerome-benoit (approved):
Reviewed `3dd3902f` with two independent sequential analyses and final arbitration. No blocking finding.

The fix reloads the disk row unconditionally when the container exists, then makes one identity decision against the reloaded state (src/session/instance/container.rs:150-178). Base only reconciled when the in-memory copy already disagreed with the container label, so a peer-persisted tool change that still matched the stale in-memory tool was reused with the wrong tool's config. The fail-closed-on-removal invariant is preserved: container.remove(true) is reached only after reloaded.context(...)?, so a failed reload can never remove a container; reuse on failed reload is the deliberate, documented, non-destructive fallback.

Verified/rejected: try_reconcile_from_disk carries the same serde-skip/peer-writable field set used before every launch, so no runtime field is wrongly reset; Ok(false) (row gone) is treated as success as in base; identity-resolution and inspect errors propagate and bail before any remove/reuse (fail-closed); legacy unlabeled containers stay reused, never removed; the reload->decision->reuse window is inherent and strictly narrower than base. The new row (("claude",""),"claude",Disk::Row("codex",""),1,None) is a genuine oracle: in-memory matches the label but the disk row is codex, so it fails pre-fix (0 removals) and passes post-fix (1). The corrupt-disk matching row covers the fail-open reuse contract.

Evidence: exact-SHA source tracing of the head diff across container.rs / reconcile.rs / runtime.rs; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. No existing reviews or threads to arbitrate. CI reports all required checks green on this head; this approval concerns the code review only.

PR #3995, opened 2026-09-15
fix(web): re-engage stick-to-bottom when a prompt is submitted

### LINKED ISSUE #3993: Structured view: submitting a prompt does not re-engage stick-to-bottom after typing unpinned the transcript
## What happens

In the web dashboard's structured view, submitting a prompt does not bring the transcript back to the bottom, so the reply streams below the fold and I have to scroll (or tap the mobile jump button) to see it.

## Why (from reading `StructuredView.tsx`)

Stick-to-bottom is the manual sampler in `StructuredView` (`wasAtBottomRef`, `autoScroll={false}` on the viewport). On a fine pointer the sampler runs on every `scroll` event. Typing into the composer auto-grows the textarea; the viewport shrinks and the browser emits an interim scroll, which the sampler reads as "the user scrolled up" and drops the pinned intent (`repinOnResize` then correctly declines to re-pin because `wasAtBottom()` is now false). Nothing re-engages the pin when the prompt is actually sent, so the freshly submitted user message and the streamed reply land under the composer.

The coarse-pointer path already gates sampling on a touch/wheel gesture (the "composer creep" fix), which is why this shows on desktop rather than on phones.

## Expected

The Claude Code CLI re-pins on submit: whatever the scroll position while typing, sending a prompt snaps the transcript to the bottom and the reply follows. The structured view should do the same.

## Proposed fix

Key a re-pin on the newest user-authored row (`user_prompt` / `user_diff_comments`): the optimistic overlay carries it from the click and the server echo reconciles to the same id, so it is one signal every submit path shares (composer send, mid-turn steer, queued delivery, EmptyState pick, retry). On change, set the pinned intent and `scrollTo` the bottom, the same way the mobile jump-to-bottom button does. The mount pass is skipped so a reader who reopened scrolled up stays put.

A follow-up could also make the composer-grow re-pin tolerate the interim scroll on fine pointers (the keyboard-transition effect already uses a "recently at bottom" window for this), but the submit re-pin is the behaviour that matters and is the smaller change.

PR follows.

## Environment

aoe 1.16.0 (Homebrew, macOS), Chrome desktop, structured view with the Claude and Codex ACP adapters.

bcstrawn:
## Description

Submitting a prompt in the structured view now re-engages stick-to-bottom, the way the Claude Code CLI does.

**Problem.** Stick-to-bottom is the manual sampler in `StructuredView` (`autoScroll={false}` on the viewport, `wasAtBottomRef`). On a fine pointer the sampler runs on every `scroll` event. Typing auto-grows the composer, the viewport shrinks, the browser emits an interim scroll, and the sampler reads it as "the user scrolled up" and drops the pinned intent; `repinOnResize` then (correctly) declines to re-pin. Nothing re-pins when the prompt is actually sent, so the new user message and the streamed reply land below the fold. The coarse-pointer path already gates sampling on gestures, which is why this shows on desktop.

**Fix.** `latestUserPromptId(rows)` (pure, in `historyScroll.ts`) returns the newest `user_prompt` / `user_diff_comments` row id. `StructuredView` reads it from the optimistic overlay first (the row exists from the click; the server echo reconciles to the same id, so the echo is not a second change) and from `activity` second, and on change sets the pinned intent and scrolls to the bottom through the same `pinToBottom` the mobile jump-to-bottom button now uses. The mount pass is skipped: the scroll-state restore owns the first pin, and a reader who reopened scrolled up stays there. Keying on the user row makes every submit path behave the same (composer send, mid-turn steer, queued delivery, EmptyState pick, retry).

Closes #3993.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`npm run test:unit`: 364 files / 3834 tests; `npx tsc -b`, `npm run lint`, `npm run format:check` clean)
- [x] Documentation was updated where necessary (inline comments; no user docs describe this behaviour)
- [ ] For UI changes: included screenshot or recording — not included: the change is scroll behaviour with no new visual element. I can add a recording if you want one.

## Test Coverage Analysis

- [ ] N/A
- [x] Skipped a test, because: the decision is the pure helper, covered by four new Vitest cases in `src/lib/historyScroll.test.ts` (newest user row, diff comments count, none, changes only when a prompt lands); the DOM effect is a `scrollTo` on that change, which jsdom cannot observe and a mocked Playwright spec would need a scroll-capable fixture with a growing composer. Added a `vitest` entry `acp.submit-repins-transcript` to `tests/coverage-matrix.json`.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Fable 5.1) in Claude Code, driven by @bcstrawn, who reproduced the behaviour and specified the fix (leave the composer-grow behaviour alone; re-engage on submit as the CLI does).

**Any Additional AI Details you'd like to share:** The diagnosis and the patch were written by the agent from reading `StructuredView.tsx`; the operator reviewed the approach before it was opened. Reviewer questions will be answered by the operator.

## Summary

- Re-pins the structured transcript when a prompt is submitted.
- Keeps submitted messages and streamed replies visible after composer resizing.
- Supports optimistic and live prompts across sends, steering, queued delivery, selections, and retries.
- Preserves restored scroll position when reopening a transcript.
- Adds decision-logic tests, a mocked browser test, coverage entries, and updated `promptSeq` documentation.

## Benefits

Users can submit prompts from an unpinned transcript and view responses without manual scrolling. Reopened transcripts retain their previous scroll position.

## Further enhancement

Verify consistent re-pinning across desktop and mobile pointer interactions.


jerome-benoit (dismissed):
Reviewed `8087fc1f` with two independent sequential analyses and final arbitration. No blocking finding.

The re-pin is keyed on the newest user-prompt row id (latestUserPromptId over optimisticRows then activity), gated by a mount-skip ref and a null guard (web/src/components/acp/StructuredView.tsx:448-458). Optimistic and echoed rows share the client-minted id (useAcpSession.ts:685-705, dropReconciledOptimistic:547-551), so reconciliation does not double-fire; the mount guard preserves a reopened-scrolled-up reader, and load-earlier while scrolled up does not re-pin.

Verified/rejected: the optimistic->echo transition reuses the same id (no second pin); the null guard prevents a clear from pinning; the from-the-end scan is O(rows-since-last-prompt).

Non-blocking (P3): (1) I reproduce CodeRabbit's one actionable point by source tracing rather than a duplicate inline: the coverage-matrix entry acp.submit-repins-transcript lists StructuredView.tsx but its only spec drives the pure latestUserPromptId helper (historyScroll.test.ts), so the effect's own branches (mount guard, optimistic/activity precedence, null guard, the pinToBottom call) are untested; add a component test that submits a prompt while scrolled up and asserts the return to bottom, plus mount-scrolled-up-stays-put and clear-does-not-pin. (2) the re-pin could key on the existing monotonic promptSeq counter (acpTypes.ts:1572-1588, already consumed at useCancelEscalation.ts:32) instead of a third copy of the user-turn predicate (historyScroll.ts:108-118 duplicates isUserTurnBoundary in acpHistoryWindow.ts:16-17); confirm first whether a user_prompt row can enter activity without a UserPromptSent control event (e.g. a prompt authored in another tab), since that path would justify the activity-leg design.

Evidence: exact-SHA source tracing of the head diff and touched files; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI reports the required checks green on this head; this approval concerns the code review only.

jerome-benoit (commented):
Re-reviewed `8087fc1f..5a75fe28` with two independent sequential analyses and final arbitration, after my earlier approval was dismissed by the redesign. No blocking finding; the delta is correct and materially better than what I approved.
The re-pin no longer keys on a third copy of the user-turn predicate; it keys on the reducer's monotonic `promptSeq` via the DOM-free `promptRepinDecision` helper (unit-tested in `promptRepin.test.ts`), gated by `hasEverOpened || localInflight`. Traced edge cases: exactly one `promptSeq` bump per prompt across optimistic/echo/other-device/diff-comments; pre-open submit pins via the `localInflight` leg; hydration/replay and manual scroll do not pin; the mount guard keeps a reopened-scrolled-up reader put; a reset lowers `seen` without a spurious pin. Both of my earlier P3s (drop the duplicate predicate; add component coverage) are resolved.
Not approving only because the sole red check is `codecov/patch`: the pin-firing wiring in `StructuredView.tsx` (the `pinToBottom` body and the `if (d.pin)` branch) executes only in the mocked-Playwright spec (`acp-submit-repin.spec.ts`), whose coverage codecov does not merge, so patch coverage falls below target. This is a coverage-source artifact, not a code defect (`scrollTo` is not observable in jsdom). Please resolve or explain the gate before merge.
Non-blocking (P3): `acp-submit-repin.spec.ts` asserts `isPinned` only after the reply streams; add `expect.poll(isPinned).toBe(true)` right after the optimistic prompt appears to isolate submit-time re-pin (CodeRabbit's open thread on that spec).
Evidence: exact-SHA source tracing of the head diff, the reducer (`acpTypes.ts`), and the helper/tests; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. Functional CI is green on this head; `codecov/patch` is the only red.

PR #3996, opened 2026-09-15
fix(web): open a structured session on its whole last turn

### LINKED ISSUE #3994: Structured view: opening a session hides most of its last turn behind Load earlier messages
## What happens

Switching to (or reopening) a structured-view session shows only the tail of its last turn. For a tool-heavy turn I have to click "Load earlier messages" several times before my own prompt, and the start of the agent's reply, come into view.

## Why (from `acpHistoryWindow.ts` / `useHistoryWindow.ts`)

The transcript opens on a flat `DEFAULT_HISTORY_WINDOW` (150 rows) and `historyWindowStart` treats that cap as primary: it snaps forward to a user boundary when one sits inside the cap, otherwise it hard-cuts mid-turn (`"hard-cuts at the cap when one huge turn has no boundary after it"`). A single turn with a few hundred tool rows is common with Claude Code, so the cut routinely lands inside the last turn. `acp.replay_events = 0` (unlimited) on the server does not change this; the recent-first tail page already holds the rows, they are just windowed out client-side.

## Expected

Whatever the row count, the session should open on at least the person's most recent prompt and the whole reply to it: that context is always relevant. Older turns can keep the default windowing.

## Proposed fix

Size the initial window from the transcript that actually arrived: `max(DEFAULT_HISTORY_WINDOW, rows since the newest user boundary)`, applied on first populate and on a session switch. The snap-forward and the sub-agent snap-back are unchanged; `historyWindowStart` then lands exactly on the last prompt. Backlog before that prompt is still behind "Load earlier", so first paint on a long session is unaffected except for the one turn that is now shown whole. A prompt older than the loaded tail page is the existing server-fetch path and out of scope here.

PR follows.

## Environment

aoe 1.16.0 (Homebrew, macOS), Chrome desktop, Claude and Codex ACP adapters.

bcstrawn:
## Description

A structured-view session now opens on at least its whole last turn.

**Problem.** The transcript opened on a flat `DEFAULT_HISTORY_WINDOW` (150 rows) with the cap primary: `historyWindowStart` snaps forward to a user boundary inside the cap, otherwise it hard-cuts mid-turn. A single tool-heavy turn (common with Claude Code) is often longer than 150 rows, so switching to a session showed the tail of its last reply and the person's own prompt sat behind several "Load earlier messages" clicks. The rows were already loaded (recent-first tail page); they were just windowed out client-side.

**Fix.** `initialHistoryWindow(rows)` returns `max(DEFAULT_HISTORY_WINDOW, rows since the newest user boundary)`; `useHistoryWindow` uses it on first populate and on a session switch instead of the flat default. The snap-forward and the sub-agent snap-back are unchanged, so with that window `historyWindowStart` lands exactly on the last prompt. Older turns keep the default windowing, so first paint on a long backlog only grows by the one turn now shown whole. A prompt older than the loaded tail page is the existing server-fetch path and out of scope.

The #3707 regression test is re-expressed: its precondition (a cut inside the last turn) can no longer occur at open, so it now asserts the new invariant, that a long last turn opens whole and the start does not snap forward when the next prompt lands.

Closes #3994.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`npm run test:unit`: 364 files / 3836 tests; `npx tsc -b`, `npm run lint`, `npm run format:check` clean)
- [x] Documentation was updated where necessary (doc comments on the helpers; no user docs describe the window)
- [ ] For UI changes: included screenshot or recording — not included: same components, more rows visible on open. Happy to add a before/after if useful.

## Test Coverage Analysis

- [ ] N/A
- [x] Skipped a test, because: the decision is pure and is covered by new Vitest cases (`acpHistoryWindow.test.ts`: default kept for a short last turn, widened to a long one, diff comments as boundary, no-boundary fallback; `useHistoryWindow.test.tsx`: opens on the whole long last turn, re-sizes on session switch, #3707 re-expressed). The existing `tests/acp-history-window.spec.ts` stories seed 2-row turns, so their expectations are unchanged by this rule; I did not run Playwright locally. `acp.history-window` in `tests/coverage-matrix.json` already lists every touched file.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Fable 5.1) in Claude Code, driven by @bcstrawn, who reported the behaviour and asked that a session always open on at least the last prompt.

**Any Additional AI Details you'd like to share:** The agent wrote the diagnosis, the patch and the tests from the source; the operator reviewed the approach. Reviewer questions will be answered by the operator.

## Summary

- Structured-view sessions now open with the complete latest user turn, including tool-heavy replies.
- The initial history window expands only when the latest turn exceeds the default size.
- Older turns retain the default window behavior.
- Session switches apply the same sizing logic.
- Added helpers and tests for user boundaries, empty histories, oversized turns, and session switches.
- Prompts outside the loaded history retain the existing server-fetch path.

## Benefits

Users can see their latest prompt and full agent response without repeatedly selecting “Load earlier messages.”


jerome-benoit (dismissed):
Reviewed `faee7a4a` with two independent sequential analyses and final arbitration. No blocking finding.

initialHistoryWindow widens the window to max(default, rows.length - lastUserBoundaryIndex), so historyWindowStart snaps exactly to the last user prompt with no mid-turn cut and falls back to the flat default when no boundary is loaded (web/src/lib/acpHistoryWindow.ts). The hook re-sizes on session switch and on the null->populate transition (web/src/hooks/useHistoryWindow.ts); growth/pin logic is unchanged. AcpRuntime passes sessionId and activity from the same per-session runtime, so no switch path produces a mis-sized window; the added Vitest cases defend the new contract.

Verified/rejected: opening on the whole last turn is the requested feature (Closes #3994), bounded to the loaded recent-first tail already reduced into state, with older prompts remaining the existing server-fetch path (out of scope); the null->populate setState fires once on the anchorLen transition and cannot re-enter; the session-switch stale-activity variant is covered by the empty->populate recompute.

Non-blocking (P3): (1) two new doc comments say "the person"/"The person's" whereas both files' established wording is "the user" (acpHistoryWindow.ts:4,58-59); please harmonize. (2) the re-expressed #3707 test seeds a boundary at index 0, so after this change no test forces computed.start > pinned and the pinnedWindowStart clamp (useHistoryWindow.ts:83-88) loses its oracle even though it stays live for the no-boundary/older-turn paths; consider a case where the initial window is the flat default and start lands mid-turn, then a later prompt appends, asserting windowedActivity[0] does not snap forward.

Evidence: exact-SHA source tracing of the head diff and touched files; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. No existing reviews or threads to arbitrate. CI reports the required checks green on this head; this approval concerns the code review only.

jerome-benoit:
Post-merge follow-up: re-reviewed the delta `faee7a4a..6edf7428` that landed after my (dismissed) approval and merged. It resolves both non-blocking P3s I raised; confirmed:
- Terminology: the doc comments now say "the user" instead of "the person" (`useHistoryWindow.ts`, `acpHistoryWindow.ts`), matching the surrounding files.
- Coverage: the new test `holds a mid-turn cut when a prompt lands after a no-boundary open (pinnedWindowStart)` restores the #3707 forward-snap oracle I flagged as lost, and it is genuine: a no-boundary tail opens on the flat default with the cut mid-turn (top `m-100`, `canLoadEarlier` true), then a later user_prompt+reply is appended and the top stays `m-100` (the start does not snap forward). Removing the `pinnedWindowStart` clamp (`useHistoryWindow.ts:83-88`) would let the start snap to the new boundary and fail this test.
No new finding on the delta. The deliberately out-of-scope case (opening on the whole last turn when it predates the loaded transcript page) is tracked in #4002.
Evidence: exact-SHA `compare faee7a4a...6edf7428` (1 commit, 3 files) plus reading the added test. No local build or test run; source-confirmed, not executed.

PR #3997, opened 2026-09-15
fix: stop two flaky live Playwright tests on main

njbrake:
## Description
Two live Playwright flakes on main.

- `mobile-live-no-flutter`: a timing assumption. After a
  stalled first frame the bottom-align grows once by design;
  the settle check passed first and read that as jitter.
  Replaced with a Vitest case for the shrink debounce
  (fails with `SHRINK_DELAY_MS = 0`).
- `live-grid-transport` synchronized-output: a product race.
  VT seeding sampled its fence before probing the pane, so a
  repainting pane exhausted every install attempt and the
  connection stayed on snapshot. The fence now starts at the
  capture fork. With each tmux call delayed 40ms: before 6/10
  failed, after 10/10 passed. A capture slower than the pane's
  repaint gap can still exhaust the attempts; the web live view
  does not re-arm.

## PR Type
- [x] Bug Fix

## Checklist

## Test Coverage Analysis
**Web dashboard**
- [x] Skipped a test, because: the live flutter spec is
  replaced by a Vitest case; coverage-matrix.json updated.

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: seed ordering is covered by a
  deterministic Unix-socket test in src/tmux/vt.rs.

## AI Usage
- [x] AI was used

## Summary

- Replaced the flaky live mobile test with a deterministic Vitest test for delayed row trimming.
- Added coverage for row stability during bottom-row changes and trimming after the 1.5-second delay.
- Updated the coverage matrix and removed the obsolete live test.
- Improved tmux VT capture fencing to accept pre-capture output and retry when output arrives after capture.
- Preserved existing drain acknowledgments and `FIONREAD` checks.

## Benefits

- Tests rely less on fragile live timing.
- VT capture handles concurrent output more reliably.
- The tests directly verify the intended debounce and capture behavior.


jerome-benoit (approved):
Reviewed `3673cbdb` with two independent sequential analyses and final arbitration. No blocking finding.

Both flakes are fixed at the source, not masked. The Rust VT change samples the chunk/generation fence immediately before the capture fork (src/tmux/vt.rs:1406-1408, paired with the returned snapshot at 1169-1194/1246-1254) instead of before the multi-attempt probe, so a repainting pane no longer forces every install attempt to Busy and strands the connection on the snapshot. The swap still commits only when received==expected && settled==expected && !FIONREAD (vt.rs:1290-1296), and chunk_seq is bumped on receipt before the parser lock (vt.rs:2210/2220/2259), so a pre-fork chunk is both counted in `expected` and present in the capture while any post-fork chunk trips Busy: no applied chunk is dropped and the trusted window is strictly smaller than before. The deleted live flutter spec had an unsound oracle (its stability run could read the one-time legitimate bottom-align growth as jitter); it is replaced by a deterministic Vitest debounce oracle (MobileLiveTerminal.rowIdentity.test.tsx) guarding SHRINK_DELAY_MS (MobileLiveTerminal.tsx:997-1016), the actual #2087 mechanism.

Verified/rejected: the fence still stands down on genuine races (new socket test seed_counts_output_from_before_the_capture_fork_as_captured, vt.rs:5185-5296: pre-fork settled -> Refreshed+installed, post-fork -> Busy and the raced chunk survives); no data-safety regression (the swap leaves the old parser holding the chunks when it stands down); live-grid-transport.spec.ts is not deleted (fixed at product level); coverage-matrix.json drops the removed spec and registers the Vitest file under terminal.mobile-live-align.

Non-blocking (P3): the deleted live spec exercised real serve+tmux+pixel geometry the jsdom replacement cannot; the tradeoff is honestly declared in the checklist. A CodeRabbit docstring-coverage advisory (78.57% < 80%) is not a CI gate.

Evidence: exact-SHA source tracing of the head diff and touched files (refs above); two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI reports all required checks green on this head including both live Playwright shards; this approval concerns the code review and does not itself claim de-flaking beyond the source reasoning and the author-reported 10/10-vs-6/10 delayed-tmux repro.

Issue #3843, opened 2026-09-08
`--auth passphrase` serves the API with no authentication (auth bypass, 1.15.3)


prefect13:
This report was written with AI assistance.

**Version:** aoe 1.15.3 (latest available; `aoe update --check` reports `available: false` as of 2026-09-08)
**Platform:** Debian GNU/Linux 13 (trixie), x86_64, Linux 7.0.14-14-pve (LXC container)

## Summary

Starting `aoe serve` with `--auth passphrase` installs no authentication at all. The
API is reachable without any credential, and the server prints:

```
WARNING: Running without authentication. Anyone with network access to this port can control your agent sessions.
```

The passphrase is accepted on the command line (and via `AOE_SERVE_PASSPHRASE`), and
`--auth=passphrase` without a passphrase is correctly rejected — so the option appears
wired up, but the resulting server has no gate.

Requesting the stronger auth mode yields a weaker one than the default. `--auth token`
(the default) behaves correctly.

## Impact

Unauthenticated callers can enumerate and control agent sessions. `/api/sessions`
returns session ids, titles, filesystem paths, tool, and status:

```json
{"sessions":[{"id":"<redacted>","title":"demo","project_path":"/home/<user>/work/demo",
"artifact_dir":"/home/<user>/.config/agent-of-empires/artifacts/<redacted>",
"tool":"claude","status":"Idle","dormant":false,"yolo_mode":false,
"created_at":"2026-09-08T00:46:45..."}]}
```

Severity is raised by two factors:

1. The failure is **silent and inverted** — an operator who deliberately chooses the
   stronger mode gets no auth. The only signal is one warning line, which is easy to
   miss when the daemon is started by a service manager and the line lands in a journal.
2. `--auth=passphrase` is the mode an operator reaches for precisely when exposing the
   dashboard beyond loopback, which is when the gate matters most.

## Reproduction

```
$ aoe serve --host 127.0.0.1 --port 8095 --auth passphrase --passphrase correct-horse-battery
WARNING: Running without authentication. Anyone with network access to this port can control your agent sessions.
aoe web dashboard running at:
  http://127.0.0.1:8095/

$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8095/api/sessions
200

$ curl -s http://127.0.0.1:8095/api/sessions | head -c 80
{"sessions":[{"id":"...
```

Control, same binary, same host, default auth mode:

```
[...]

PR #3999, opened 2026-09-16
fix(acp-client): authenticate against passphrase daemons via login fallback

cwrau:
## Description

Local CLI commands (`aoe acp prompt`, `history`, `status`, `approve`, `cancel`, `tail`, `switch-agent`, `attach`) couldn't authenticate against a daemon running `aoe serve --auth=passphrase`. `HttpClient` (`src/acp/client/http.rs`) and the WebSocket client (`src/acp/client/ws.rs`) only ever sent `Authorization: Bearer <token>`, sourced exclusively from `serve.token` / `AOE_DAEMON_TOKEN` — a file/var that a passphrase daemon never writes.

The daemon does have a loopback auth bypass for passphrase mode, but it's deliberately withdrawn under `--behind-proxy` (#3843): a request arriving on the proxy's own loopback socket is otherwise indistinguishable from one it forwarded on behalf of a remote caller. That's exactly the `agent-ui.service` deployment shape (passphrase-gated, reverse-proxied to the public internet), so even the same local user who configured the daemon had no way to authenticate short of restarting it in `--auth=token` mode, which kills the live web dashboard/tunnel session.

This adds a genuine passphrase-login fallback (`src/acp/client/passphrase_session.rs`, new module):
- `DaemonEndpoint::resolved_passphrase()`: `AOE_DAEMON_PASSPHRASE` env override (gated to `https://` or loopback transport — see review fixes below), or — for a loopback local daemon — the daemon's own `serve.passphrase` file (the same file `aoe serve --restart` already reads; same trust boundary the client already relies on for `serve.token`).
- When no bearer token resolves but a passphrase does, both clients perform the same `POST /api/login` handshake the web dashboard uses, cache the resulting `aoe_session` cookie + device-binding secret (persisted under the app dir, owner-only `0600`) so repeat CLI invocations reuse one login instead of minting a new device session every process, and retry once after a fresh login if a *cached* session comes back `401` (rotated passphrase, expired/evicted session).
- Extends to `ws.rs` too (`aoe acp tail`, and the TUI structured view's live stream), not just the REST client.

Doesn't touch the server's loopback-bypass / `--behind-proxy` logic at all — this is a pure client-side addition, so the security posture (`--behind-proxy` still requires a real credential) is unchanged.

### Review fixes folded in

Two independent reviews (CodeRabbit + a maintainer) caught real issues, all fixed here:
- **Cleartext transport (CWE-319):** `resolved_passphrase()` now refuses `AOE_DAEMON_PASSPHRASE` for a plaintext non-loopback `http://` endpoint (only `https://` or loopback).
- **Broken 401 retry (CWE-613):** `PassphraseSessionCache::clear()` → `invalidate()` now also deletes the persisted `cli_login_session` file, not just the in-memory copy. Previously the retry-after-401 path silently resent the same rejected cookie forever instead of ever logging in again.
- **WS handshake hard-fails in passphrase mode (CWE-287):** the daemon only echoes a `Sec-WebSocket-Protocol` response header when the client offers the literal `aoe-auth` subprotocol; tungstenite fails the handshake outright if the response omits it while the request carried one. The WS client now offers `aoe-auth, aoe-device.<secret>` together, matching the web client. Without this, every passphrase-mode `aoe acp tail` / TUI live-stream connection failed immediately, before a single frame or even the retry logic ran.
- **`parse_session_id` hardening:** the new e2e test's helper now rejects an empty `ID:` value instead of retrying it for the full 30s timeout.

### Bonus fix (found via the new `tail` e2e test)

Adding the first-ever e2e coverage of `aoe acp tail` surfaced an unrelated, pre-existing bug: `AcpState::recent_diffs` is one of the server's "cold" fields (`COLD_STATE_FIELDS` in `src/server/acp_ws.rs`, omitted from the wire when unchanged), but unlike its three siblings it was missing `#[serde(default)]` — so any `reduced_state` frame that omitted it failed the whole parse. This affects the TUI structured view too, not just this PR's surface; fixed with a one-line `#[serde(default)]`.

### Known remaining gap (not fixed here, flagged by both reviews)

`session_view_info` (`src/acp/client/http.rs`) still goes through `DaemonEndpoint::daemon_client()`, a separate bearer-only client (`src/daemon/`), so it doesn't get the passphrase-session treatment. Both reviewers flagged this as non-blocking / a heavier lift (extending `DaemonClient` itself). Left as a follow-up.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Two new e2e tests in `tests/e2e/serve.rs`:
- `cli_acp_prompt_authenticates_against_behind_proxy_passphrase_daemon`: starts a real `--auth=passphrase --behind-proxy` daemon and runs an actual `aoe acp prompt` against it, which previously 401'd.
- `cli_acp_tail_connects_to_behind_proxy_passphrase_daemon`: same daemon shape, spawns `aoe acp tail` and confirms the WS handshake succeeds (process stays alive streaming) instead of failing immediately — this is what caught both the missing `aoe-auth` subprotocol bug and the `recent_diffs` deserialization bug.

All 3 pre-existing passphrase e2e tests still pass. Also added 24 unit tests across `discovery.rs`, `http.rs`, `passphrase_session.rs`, and `ws.rs` covering passphrase resolution precedence (incl. the transport gate), cookie/header construction (incl. the WS subprotocol list), device-binding persistence, session invalidation, and the login/retry flow.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude Code (Sonnet 5)

Root cause was pre-identified by the reporter. Claude Code designed and implemented the passphrase-login fallback (including extending it to the WS client after flagging that gap mid-implementation), verified and fixed all review feedback from CodeRabbit and the maintainer review against current source (two of the CodeRabbit findings were re-confirmed as already fixed by the time of the second pass; the WS subprotocol finding was verified against actual server/tungstenite source before fixing), found and fixed the unrelated `recent_diffs` bug via the new e2e test, and updated docs.

## Summary

- Added passphrase authentication for REST and WebSocket `aoe acp` commands.
- Added passphrase lookup from `AOE_DAEMON_PASSPHRASE` and local daemon files.
- Added secure session and device-binding storage with session reuse.
- Added one retry when a cached session expires.
- Updated CLI and user documentation.
- Added unit tests and an end-to-end test for passphrase authentication behind a proxy.
- Made older ACP state load correctly when `recent_diffs` is missing.

## Benefits

Users can use `aoe acp`, `tail`, and structured views with passphrase-authenticated daemons. Local clients can authenticate without manual token handling. Cached sessions reduce repeated logins.

## Potential follow-up

Continue monitoring session expiry and authentication behavior across proxy configurations.


njbrake (changes requested):
**Stale cached session is never replaced.** `PassphraseSessionCache::clear` only empties memory; the retry's `get()` reloads the rejected `cli_login_session` from disk, so no fresh login happens. After a passphrase change or session eviction every `aoe acp` call returns 401 until the file is deleted by hand. `clear` must also remove the persisted cookie (keep the binding secret).

**WS upgrade fails against a passphrase daemon.** The client offers only `aoe-device.<secret>`; the daemon selects only `aoe-auth`, and tungstenite rejects a handshake with no echoed subprotocol. `aoe acp tail` exits with `Server sent no subprotocol`. Offer `aoe-auth, aoe-device.<secret>` as the web client does, and cover `tail` in the e2e test.

**Passphrase sent over plaintext HTTP.** `AOE_DAEMON_PASSPHRASE` is posted to any URL. Refuse non-loopback `http://`, matching `DaemonClient::new`'s `InsecureBearerTransport`.

**Remote use evicts browser sessions.** Without a disk cache, each `HttpClient` / WS connection logs in again: one `aoe acp status` creates two server sessions and two new-login pushes. With `MAX_SESSIONS = 50`, about 25 CLI calls sign out every other device. Cache remote sessions (keyed by URL) or share one per process.

**`--daemon-url` help is inaccurate.** The remote TUI session list goes through `DaemonClient`, which has no passphrase support, so `AOE_DAEMON_PASSPHRASE` does not work there yet.

Minor: `write_owner_only` chmods after writing; open with `mode(0o600)` like `write_secret_file`. Login should not follow redirects (307/308 re-post the passphrase). Trim the history narration in the new doc comments and e2e test header.

jerome-benoit (changes requested):
Reviewed `174201db` with two independent sequential analyses (including a security pass) and final arbitration, and reassessed against @njbrake's standing review and the CodeRabbit threads. Changes requested; I independently confirm the existing blockers rather than duplicating inline findings.
The REST prompt path (the only surface the new e2e test exercises) works and the passphrase-fallback idea is sound, but the fix is incomplete and unsafe on the rest of the auth surface. Confirmed by tracing client+server+web source at this head:
- Cleartext passphrase over non-loopback HTTP (CWE-319): `resolved_passphrase` (discovery.rs) returns the env passphrase before any transport check, so `passphrase_session::login` POSTs the passphrase and device-binding secret in cleartext to a remote `http://` daemon. The bearer path already guards this (`InsecureBearerTransport`); the passphrase path needs the same https/loopback gate.
- 401 retry never recovers (CWE-613): on 401, `clear()` nulls only the in-memory session; the rebuild reloads the same rejected cookie from `cli_login_session` on disk, so the retry resends the stale credential. After passphrase rotation/eviction every `aoe acp` call 401s until the file is deleted by hand. Invalidate the persisted cookie (keep the device binding) on 401, for both HTTP and WS.
- WS omits the mandatory `aoe-auth` subprotocol (CWE-287): `build_authed_request` offers only `aoe-device.<secret>`; the daemon (`acp_ws.rs:109`) echoes a subprotocol only if `aoe-auth` was offered, so tungstenite tears the handshake down (NoSubProtocol) and `aoe acp tail` / the TUI live stream cannot connect to a passphrase daemon. Offer `aoe-auth` first like the web client, and cover `tail` in e2e.
Non-blocking (P3): remote endpoints re-login on every HTTP client and every WS connect (no shared/URL-keyed cache), churning server sessions and risking eviction under `MAX_SESSIONS`; and `attach`'s `session_view_info` still bypasses passphrase auth.
Verified/rejected: secret-in-logs, the on-disk-secret trust boundary, and fail-closed behavior all check out; nothing new survived beyond the maintainer's and CodeRabbit's set.
Evidence: exact-SHA source tracing of `src/acp/client/{discovery,http,ws,passphrase_session}.rs`, `src/server/{login,acp_ws}.rs`, and `web/src/hooks/useAcpSession.ts`; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. This confirms the standing CHANGES_REQUESTED; the code bar is not met.

jerome-benoit (approved):
**Verdict: APPROVE** — the new revision resolves the last blocker; all three findings from my `CHANGES_REQUESTED` are now fixed. Re-reviewed the delta `bc5093e3..5af3dd55` at `5af3dd55` (two independent passes incl. security, plus my own trace against the tungstenite 0.30 source); source-confirmed, not executed.

- **CWE-287 WS `aoe-auth` subprotocol — FIXED.** `build_authed_request` now offers `Sec-WebSocket-Protocol: aoe-auth, aoe-device.<secret>` (`ws.rs:250-256`). The daemon's `ws.protocols(["aoe-auth"])` (`acp_ws.rs:109`) now echoes `aoe-auth`, so tungstenite's `verify_response` accepts (returned subprotocol ∈ offered set) instead of failing with `NoSubProtocol`; the auth middleware still reads the binding position-independently from the `aoe-device.<secret>` entry (`auth.rs:466-471`). `aoe acp tail` and the TUI live stream can now complete the upgrade against a passphrase daemon.
- **CWE-319 cleartext passphrase — FIXED** (`transport_is_safe_for_passphrase` gate in `resolved_passphrase`).
- **CWE-613 401 never recovers — FIXED** (persisted-cookie `invalidate()` on 401 for both HTTP and WS, device binding kept).

The new e2e `cli_acp_tail_connects_to_behind_proxy_passphrase_daemon` is a genuine regression test: `--behind-proxy` drops the loopback bypass, so `tail` performs a real passphrase login + WS upgrade, and it asserts the process keeps streaming rather than exiting early on `NoSubProtocol` (fails pre-fix, passes post-fix). Nicely covers the exact gap.

Non-blocking (P3, your call, not new): `attach`'s `session_view_info` remains bearer-only (401s against a remote passphrase daemon); remote endpoints re-login per HTTP client / WS connect (no URL-keyed on-disk cache). Neither blocks this PR.

Evidence: exact-SHA trace of `src/acp/client/{discovery,http,ws,passphrase_session}.rs`, `src/server/{auth,acp_ws}.rs`, the new e2e, and the tungstenite 0.30 client handshake. No local build/test run.

cwrau:
@njbrake Addressed everything from your review in the latest commit:

- **Stale cached session**: fixed — \`clear()\` (now \`invalidate()\`) deletes the persisted \`cli_login_session\` file too, not just the in-memory copy.
- **WS upgrade fails**: fixed — now offers \`aoe-auth, aoe-device.<secret>\` together, matching the web client.
- **Passphrase over plaintext HTTP**: fixed — refused unless \`https://\` or loopback.
- **Remote use evicts browser sessions**: fixed — remote endpoints now get the same on-disk session cache as local ones, keyed by base URL under the app dir, so repeated \`aoe acp <verb>\` calls against the same remote daemon reuse one login instead of minting a new device session each time.
- **`--daemon-url` help inaccurate**: fixed — corrected to say the remote session list goes through the bearer-only \`DaemonClient\` and doesn't get passphrase support yet; \`aoe acp <verb>\` against the same \`AOE_DAEMON_URL\` does.
- **`write_owner_only` chmod-after-write**: fixed — now opens with mode \`0600\` at creation instead of chmod'ing after, closing the brief world-readable window.
- **Login following redirects**: fixed — the login POST now uses a dedicated client with \`redirect::Policy::none()\`.
- Trimmed the narration in the new doc comments and the e2e test header.

Also caught (via the new \`tail\` e2e test, which exercises \`reduced_state\` parsing for the first time in e2e) an unrelated pre-existing bug: \`AcpState::recent_diffs\` was missing \`#[serde(default)]\` even though the server can omit it as a "cold" field — fixed that too.

The one item I'm leaving as a follow-up: \`session_view_info\` still goes through the bearer-only \`DaemonClient\`. Both you and jerome-benoit flagged it as a heavier lift / non-blocking, and it's a genuinely separate client (\`src/daemon/\`) that would need its own passphrase-session support.

cwrau:
The CodeQL check flagged 18 new "high" alerts on this commit (all "Cleartext transmission of sensitive information" / one variant "writes ... to a log file"). I looked into all 18 before assuming they're real:

- 16 of them are on lines like \`self.send_authed(|| self.http.get(&url)).await?\` across pre-existing methods (\`replay\`, \`files\`, \`cancel\`, the \`queue_*\` methods, etc.), flagging that \`session_id\` flows into the request URL. This exact URL construction (\`session_id\` as a path segment) is unchanged from before this PR — only the send call's wrapper changed (\`self.auth(...).send()\` → \`self.send_authed(...)\`, added so passphrase auth can retry on 401). \`session_id\` is a routine ACP session identifier, not a credential — it's already displayed in \`aoe list\`, stored in plaintext session files, and sent in URLs throughout the existing codebase. I believe this is CodeQL's sensitive-data heuristic matching on the substring "session" in the identifier name, re-triggering because the diff touched every send-call line in the file, not a new data-flow issue.
- The other 2 are on \`request.send().await?\` inside the new \`send_authed\` helper, flagged as writing \`ensure_passphrase_session()\`'s result to "a log file." That return value (the \`aoe_session\` cookie + device-binding secret) is sent exactly once, as the \`Cookie\`/\`X-Aoe-Device-Binding\` HTTP headers on the authenticated request — by design, the same way the existing bearer-token path sends \`Authorization\`. It is never logged, and never placed in a URL. I think the "log file" wording is this query's generic sink description misapplied to a plain HTTP send.

I don't have permission to view the alert details or dismiss code-scanning alerts on this repo (pull-only access), so I can't fully confirm the rule ID or dismiss these myself if you agree they're false positives. Flagging so someone with repo access can take a look at https://github.com/agent-of-empires/agent-of-empires/security/code-scanning?query=pr%3A3999 rather than have it silently block on a check I can't act on.

jerome-benoit (dismissed):
**Verdict: APPROVE** — re-reviewed the delta `5af3dd55..68afdf9c` at `68afdf9c`; **no regression**. Two independent analyses (incl. a security pass) + source trace.

My three prior blockers stay fixed: CWE-319 transport gate (`transport_is_safe_for_passphrase`, https-or-loopback only), CWE-613 401 recovery (persisted-cookie `invalidate()` on 401 for HTTP and WS), CWE-287 WS `aoe-auth` subprotocol (`aoe-auth, aoe-device.<secret>`). The delta is additive: a remote passphrase session cache (SHA-256-keyed under the app dir, which resolves the re-login-churn P3 I noted) + a 15s `login_client` timeout + empty-session-id e2e hardening. Both CodeRabbit actionable items (SHA-256 cache key, login timeout) are already applied; the `cli_acp_tail_connects_to_behind_proxy_passphrase_daemon` WS e2e is present and meaningful.

**CI + Tests authorized and green on `68afdf9c`** (fork substantive workflows are gated; I authorized and confirmed them, not just the rollup). No local build/test run beyond that.

Non-blocking (P3, pre-existing): `session_view_info` is still bearer-only, so it 401s against a remote passphrase daemon behind a proxy — a heavy-lift follow-up, out of this delta's scope.

jerome-benoit (approved):
Re-reviewed `208458de` against base `ec34932b`, including the rebase onto the simplified HTTP client, with two independent sequential passes. **Code approval renewed:** no blocking regression established; the prior transport, persisted-cookie invalidation and WS-subprotocol fixes remain intact. The separate bearer-only `session_view_info` limitation remains nonblocking as agreed.

Validation: CI and Tests workflows passed. [Linux web/e2e](https://github.com/agent-of-empires/agent-of-empires/actions/runs/35837012948/job/107179741506) tested merge `49c330c` of this exact head/base; both passphrase CLI tests passed. No local build/tests were run. The macOS job skips its test steps on PRs and is not macOS test evidence.

**Not an all-green merge verdict:** the code-scanning check still flags [216](https://github.com/agent-of-empires/agent-of-empires/security/code-scanning/216) and [250](https://github.com/agent-of-empires/agent-of-empires/security/code-scanning/250) at `request.send()` (`http.rs:597,603`). These are request dispatches, not logging calls in this module; maintainer triage of the reported sink remains outstanding. I have not dismissed the alerts.

**Nonblocking test gap:** `serve.rs:542-544` waits two seconds then checks that `tail` is alive; that does not establish a completed handshake. Both reviewers recommend awaiting an actual received frame with a deadline. This is unchanged since the previous approval.

PR #4000, opened 2026-09-16
test(web): revoke a runner record saved back during teardown

njbrake:
## Description
acp-force-end-turn failed in teardown on two PRs with
"runner groups did not exit", 25s after a test body that
finished in under a second.

Cause: the harness revokes a runner's registry record by
renaming it, without the per-session lock every other writer
takes. A runner that loaded the record just before the rename
(a late mark_detached after the daemon exits) saves it back
just after, so its watchdog keeps matching the record and the
runner never exits.

stopOrphanRunners now re-revokes a record that reappears for
the same pid and generation while it waits. With the race
forced, runs hung 4/4 before and passed 6/6 after;
acp-force-end-turn passed 30/30 under CPU load.

## PR Type
- [x] Bug Fix

## Checklist

## Test Coverage Analysis
**Web dashboard**
- [x] Skipped a test, because: this is a live-harness teardown
  fix; the Vitest regression in tests/helpers/aoeServe.test.ts
  drives serve.stop() against a runner that saves its record
  back, and fails without the fix.

**TUI / CLI (e2e test)**
- [x] N/A: no TUI, CLI, or session-lifecycle change.

## AI Usage
- [x] AI was used

## Summary

- Fixed a teardown race in `stopOrphanRunners`.
- Re-revokes matching runner records that reappear during teardown.
- Extracted fake `aoe` setup into `writeFakeAoe`.
- Added a Linux-only regression test for this race.

## Benefits

- Prevents the watchdog from matching revoked records.
- Allows runners to exit reliably.
- Removes the serve home after teardown.

Tests pass: forced race 6/6 and `acp-force-end-turn` 30/30 under CPU load.


jerome-benoit (approved):
Reviewed `7cd11c00` with two independent sequential analyses and final arbitration. No blocking finding.
The teardown fix is correct: `stopOrphanRunners` records revoked runners (keyed by record name with matching pid and decimal generation) and re-revokes via `revokeAgain` on each 50ms wait-loop tick, so a record a late `mark_detached` saves back after the initial revocation is renamed again; the runner then sees it gone for >=2 ticks and exits, and `serve.home` is removed. The Linux-only regression test is a genuine oracle (pre-fix the resaved record persists, the runner never exits, the 10s race throws), not tautological.
Verified/rejected: statically-imported fs (no leftover dynamic import); string-vs-string generation compare consistent with the main loop; `revokeAgain` only re-revokes a record the same awaited runner saved back (pid+generation guard), so it cannot clobber an unrelated record; the normal path is a cheap ENOENT no-op.
Non-blocking (P3): the test synchronizes only on `runner-ready` (title), not on the runner having loaded its record before `serve.stop()`; losing that ~20ms race yields a spurious FAILURE (never a false pass). Per the repo's causal-sync direction (main d7baaedb), write a `runner-record-loaded` marker and `expect.poll` it before `serve.stop()`. Validates CodeRabbit's open thread on `aoeServe.test.ts`; not duplicating it.
Evidence: exact-SHA source tracing of the head diff (`web/tests/helpers/`), the fake-runner watchdog, and axum/tungstenite handshake semantics; two independent sequential analyses with final arbitration. No local build or test run (the new test is Linux-only; host is macOS); source-confirmed, not executed. CI reports all checks green on this head; this approval concerns the code review only.

PR #4006, opened 2026-09-16
test(web): wait for the fake runner to load its record

njbrake:
## Description
Follow-up to #4000. The resaved-record regression called
stop() right after writing the runner's record. If teardown
revoked it before the fake runner's first 20ms read, the
runner had nothing to save back and never exited, so the
test failed although the harness fix was present.

The fake runner now writes a marker after its first read and
the test waits for it before stopping. Without the harness
re-revoke the test still fails as intended.

## PR Type
- [x] Bug Fix

## Checklist

## Test Coverage Analysis
**Web dashboard**
- [x] N/A: test-only change to tests/helpers/aoeServe.test.ts.

**TUI / CLI (e2e test)**
- [x] N/A: no TUI, CLI, or session-lifecycle change.

## AI Usage
- [x] AI was used

## Summary

- Updated the runner teardown test to wait until the fake runner loads its record.
- Added a marker to synchronize the first record read.
- Preserved the loaded record to verify that the runner can resave it after teardown revokes it.

## Benefits

This removes a timing-dependent failure and makes the teardown behavior deterministic.

No test execution results were provided.


jerome-benoit (approved):
Reviewed `13c3448d` with two independent sequential analyses and final arbitration. No blocking finding.
The added `runner-record-loaded` marker + `expect.poll` before `serve.stop()` makes the #4000 regression test deterministic: it forces the fake runner to prove it read the record (one-shot marker on first read, `saved` goes undefined->Buffer so it is written exactly once, web/tests/helpers/aoeServe.test.ts:181-192) before teardown revokes it, exercising the resave-after-revoke path instead of racing it. The test still fails without the harness re-revoke as intended: the resaved record persists, the runner never hits `++missing>=2`, `stop` never resolves and the `delay(10_000)` race throws.
Verified/rejected: marker isolation (`root` is a per-test `mkdtempSync`, cleaned in `finally` :158/:237, so no stale marker across runs); `existsSync`/`join` are imported (:4-14); `if (!saved)` is equivalent to the old `saved ??=` for the only possible values (undefined | truthy Buffer); `delay` is real here (test 1's `vi.restoreAllMocks` in finally; the it-blocks are independent). Follows web/AGENTS.md causal-sync (observable predicate + deadline, no fixed readiness sleep); coverage-matrix N/A (test-only helper).
Non-blocking (P3): the three serial deadlines (runner-ready poll 10s :201, new runner-record-loaded poll 10s :208, stop-race `delay` 10s :220) sum to 30s worst-case while the `it()` timeout is 20s (:240); on a pathologically slow runner a genuine breakage would surface as an opaque 20s Vitest timeout instead of the intended poll message. The marker lands within a few 20ms ticks, so optionally shorten the new poll timeout (fails faster with its message) or raise the `it()` timeout to >=30s. Harmless in the normal path.
Evidence: exact-SHA source tracing of the head diff and touched file (file:line above); two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI reports all required checks green on `13c3448d` per `gh pr checks`; this approval concerns the code review only.

PR #4007, opened 2026-09-16
fix(web): restore the hidden diffs on the highlight components, and guard them

matthewpwatkins:
## Description

Four highlight components delimited a render cache key with a literal NUL byte. Two distinct problems follow from that, and I described them imprecisely in earlier revisions of this PR; corrected here.

**The full-file viewer's diff has been invisible.** Its NUL sits at byte 1186, inside the roughly 8000-byte window git sniffs to decide binary-ness, so git calls the blob binary and GitHub review shows no diff for that file at all. #3974's description flagged it as unreviewable without identifying the cause.

**The markdown renderer's diff has always rendered.** Its NUL sits at byte 9910, past that window, so git never treated it as binary. The review cost there was different but real: a NUL renders as nothing, so a reviewer of #3974 quoting that key reported it had no field separator when the source in fact had one, and raised a collision concern that was already handled. Same byte, different failure.

Neither shows up in CI, because the code compiles and the tests pass either way.

## What this changes

The delimiter becomes a six-character escape sequence rather than the byte (`u0000` preceded by a backslash) in all four components. The key is byte-identical at runtime; the sources become ordinary text.

Two layers stop it recurring:

- `scripts/check-no-nul-in-source.py` fails when any tracked text file carries a raw NUL. It runs in the Format job ahead of the toolchain, so it gates every PR in about a second with no build. Verified both directions: against `main` it names exactly the two offenders, against this branch it is silent. Python to match the four existing repo-wide checks.
- A root `.gitattributes` marks source extensions diffable, so a stray NUL cannot hide a diff again once this lands.

## Reviewing the one file whose diff cannot render

`.gitattributes` does **not** rescue this PR's own diff, and it is worth being plain about why: git decides binary-ness per diff *pair*, and GitHub evaluates attributes from the base side, which is `main` and has neither the attribute nor the fix. So the file that most needs review is the one file GitHub will not show. That is inherent to removing a NUL, not something this PR can work around.

The change to `web/src/components/diff/FullFileViewer.tsx` is one line plus a comment, reproduced here in full:

```diff
   // Drop stale highlighted markup when the rendered input changes, so a switch
   // to an unknown-language or load-failing file can't keep painting the
   // previous file's html. Synced at render time (not in an effect) to satisfy
   // the set-state-in-effect lint, mirroring DiffFileViewer's syncKey pattern.
-  const inputKey = `${filePath}[NUL-BYTE]${content.length}`;
+  // NUL-delimited (as the escape sequence: a raw NUL byte in source makes
+  // git treat the file as binary) so field concatenations cannot collide.
+  const inputKey = `${filePath}[BACKSLASH-u0000]${content.length}`;
```

Both placeholders are substitutions, because GitHub rewrites the real escape sequence in this text field even inside a code fence (it decodes it and then sanitises the resulting control character, which is how an earlier revision of this description ended up displaying caret notation). `[NUL-BYTE]` is the literal control byte being removed. `[BACKSLASH-u0000]` is the six ASCII characters replacing it: a backslash, then `u0000`.

Two ways to read the real thing rather than my transcription:

- [`FullFileViewer.tsx` at this branch's head](https://github.com/matthewpwatkins/agent-of-empires/blob/fix/nul-key-delim/web/src/components/diff/FullFileViewer.tsx) renders as normal source, because the proposed version contains no NUL. The blob view works where the diff view cannot.
- `git show HEAD -- web/src/components/diff/FullFileViewer.tsx` after checking out this branch, which renders as a text diff locally since the branch carries the attribute. The other six files render normally in the GitHub diff.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**

The component change is runtime-identical, so it carries no new test; the stale-content and late-resolution suites from #3974 cover these paths and pass. The guard ships with its own `--self-test`, wired into CI alongside the live run, following `scripts/check-nix-embedded-assets.py`.

`npx tsc -b`, `npm run lint`, `npm run format:check`: clean. Affected specs: 116 tests pass.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** The per-diff-pair behaviour and the 8000-byte sniff window were established by isolating the blob pair outside any repo, after a reviewer of this PR pointed out that a force-push changes the pairing and the pre-push state is not evidence for the post-push one.


jerome-benoit (approved):
Reviewed `3af66d04` with two independent sequential analyses and final arbitration. No blocking finding.
The invisible-diff fix is correct where it matters: on the two files main renders as binary (`Markdown.tsx`, `FullFileViewer.tsx`) the raw NUL delimiter is replaced by the `\u0000` escape, which is byte-for-byte the same runtime string (`"\u0000" === String.fromCharCode(0)`), so the render cache key value is unchanged (no memoization/collision regression) and the source becomes reviewable text again. The `.gitattributes` `diff` attribute and the `scripts/check-no-nul-in-source.py` guard (run as a fast pre-toolchain CI step) prevent the failure class from recurring; I ran the guard in a throwaway repo (self-test exit 0, clean tree exit 0, a `.ts` with a NUL -> exit 1 with the offender, a binary png with NULs correctly ignored, a space-in-path detected via `git ls-files -z`).
Verified/rejected: `ToolCards.tsx` and `DiffCommentsUserCard.tsx` were never binary (normal diffs on main); their change is a space -> `\u0000` delimiter switch, i.e. collision-hardening (the key is an in-memory `useState` value compared for equality only; worst case a one-time re-highlight on deploy), not a regression.
Non-blocking (P3): (1) the delimiter switch on `ToolCards`/`DiffCommentsUserCard` is beyond the title's "restore the hidden diffs"; worth a line in the description. (2) coverage gap: the guard's `TEXT_SUFFIXES`/`TEXT_NAMES` miss `.astro` (17 tracked files under `website/src/components`) and `Dockerfile.dev`, though the guard advertises "any source file"; add `.astro` (and glob `Dockerfile*`). (3) `.gitattributes` (17 exts) is a strict subset of the guard's `TEXT_SUFFIXES` (which adds .lock/.scss/.svg/.mdx/.txt/.sql/.snap/.patch/.diff); align the two so the display-forcing and prevention layers cover the same set. (4) `self_test()` only asserts path classification, not the detection/exit-code path, so the comment claiming it enforces coverage overstates it; and `tracked_files()` decodes the `ls-files -z` names as UTF-8 (fine for this repo, would spuriously fail on a non-UTF8 tracked path).
Evidence: exact-SHA source tracing of the head diff and touched files (byte-level NUL counts at base vs head; a functional run of the guard script); two independent sequential analyses with final arbitration. No local build of the app or test run; source-confirmed, not executed. I authorized the gated CI/Tests/CodeQL workflows on this head and they are now running (not yet green). This approval concerns the code review only.

Issue #4004, opened 2026-09-16
Web dashboard has no system health strip


matthewpwatkins:
## What

The TUI has an optional system health strip (setting: "Show system health strip") showing host memory/CPU headroom, pressure signals, and per-agent resource usage. The web dashboard has no equivalent: the server exposes no system/agent metrics endpoint and the frontend has no component for it, so remote users get no visibility into host health.

## Why it matters

The web dashboard is the surface used away from the machine, which is exactly where you cannot open the TUI to check whether the host is running out of memory or an agent is eating CPU. Headroom problems (the strip's Warn/Critical states) currently go unseen until sessions start failing.

## Scope

- Expose the existing host + per-agent metrics sampling over the web API.
- Render a health strip in the dashboard with the same Warn/Critical semantics as the TUI.

Found while comparing TUI and web feature coverage; no existing issue or PR covers this.

PR #4013, opened 2026-09-16
refactor(web): read the session toggles the dashboard honours from one gate

matthewpwatkins:
## Description

Three dashboard settings gates were the same twenty lines three times: a context holding the resolved value, a parser that reads the field out of an `/api/settings` payload and treats only the opposite of the default as a deliberate choice, and a hook. A fourth copy was about to land with the web system-health strip (#4004), which is what surfaced this.

`web/src/lib/sessionFlagGate.ts` is now the one implementation, taking the setting key and its default. `sessionColors.ts` and `unreadIndicator.ts` are three lines each on top of it, and their exported names are unchanged, so no call site moves.

Each gate keeps its own default, which is the part that matters for correctness: a daemon older than the field it is asked about still gets the documented fallback rather than a silent `false`. Session colours and the unread indicator default on; the health strip defaults off.

Split out of the #4004 work at review request, since it touches code that feature does not own. The strip stacks on top of this branch.

## Reviewer notes / open questions

No visual change, so nothing to screenshot. Open question from me: the factory takes a default per gate, which keeps the two existing gates' behaviour exactly. If you would rather every gate default off and the default live only on the server, say so and I will invert it.

## PR Type

- [ ] New Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] For UI changes: no visual change, so no screenshot

## Test Coverage Analysis

**TUI / CLI (e2e test)**

Behaviour-preserving, and the existing specs for both gates are the regression test: they already cover the on default, an explicit `false`, a missing section, and a malformed payload, and they pass unchanged against the factory. No new tests, since a test of the factory alone would restate those.

`npx tsc -b`, `npm run lint`, `npm run format:check`: clean. `npm run test:unit`: 367 files, 3,865 passed.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**Any Additional AI Details you'd like to share:** Extracted during a duplication review of #4004, then verified against the two existing gates' own specs.


matthewpwatkins:
Folding this into #4014 at the author's request: one request, rebased on trunk, with the gate refactor as its own commit inside it.

PR #4020, opened 2026-09-17
fix(acp): hold post-load replay suppression until the replay drains

### LINKED ISSUE #4016: Post-load history replay leaks into the transcript when a prompt wakes an idle-dormant session (follow-up to #3937)
**A prompt that wakes an idle-dormant session ends history-replay suppression before the post-`SessionReady` replay finishes draining, duplicating the prior transcript.** Suppression is armed on `session/load` and cleared on the *first* prompt via `suppress_for_block.swap(false, …)` (`src/acp/acp_client/connection.rs:1780`); the notification handler only drops replayed transcript events while `suppressing` is true (`:687`). If the wake prompt arrives before the replay drain completes, the remaining replayed events pass through and are re-published into the SQLite event store, so every prior assistant bubble renders twice. The setup comment (`:285-295`) frames the window as "between session/load success and the first user prompt", which is exactly the too-early close.

**Repro:** multi-turn ACP session -> let it go idle-dormant (next interaction triggers native `session/load`) -> send a prompt to wake it -> prior turns appear duplicated. @njbrake (approving #3937): "Main already leaks part of it."

**Fix direction:** gate the suppression-off transition on observed drain completion (a drain-complete barrier) rather than on the first prompt.

_Note: drafted by an AI on behalf of @jerome-benoit; the analysis and decision are his, the prose is the AI's._

njbrake:
## Description

When a sleeping chat session wakes up, the agent replays the whole earlier conversation so it can pick up where it left off. We already have that history saved, so we hide the replay. The hiding stopped at the first new message, and a message sent right away could stop it before the replay finished, so earlier replies showed up twice. Now the session waits until the whole replay has been handled before it accepts the new message.

After a successful `EstablishSession`, the runner queues a `_aoe/session_replayed` Notify right behind `SessionReady`. Handshake frames go out first, so the marker follows the replayed updates. Ordered SDK dispatch reaches it only after they apply, and the `session/load` path waits for it. The runner drops agent-sent markers. Older daemons ignore the notification.

Fixes #4016

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the race needs a stalled dispatcher; `established_session_barrier_follows_its_replay` checks the runner's wire order and forged-marker drop; `session_replayed_waits_for_replay_sent_after_the_reply` gates a replayed update and asserts the daemon keeps waiting.

## AI Usage

- [ ] No AI was used
- [x] AI was used

## Summary

This PR adds a runner-controlled replay-completion marker for ACP sessions.

After `EstablishSession` succeeds, the runner sends `_aoe/session_replayed` after the session replay updates. The daemon waits for this marker before it accepts new prompts. Agent-sent markers are ignored, and older daemons remain compatible.

This keeps replay suppression active until all replayed updates are processed. It prevents old transcript entries from being written to the event store a second time when an idle session wakes.

## Changes

- Added the `SessionReplayed` control notification.
- Added replay tracking and acknowledgment handling in `DaemonControlClient`.
- Added runner-side ordering for the replay marker.
- Prevented agents from forging the marker.
- Added regression and integration test coverage.

## Benefits

- Prevents duplicated transcript history after session reload.
- Ensures prompt processing starts only after replay completes.
- Preserves compatibility with older daemons.

## Further enhancement

No current review severity counts or test execution results were provided.


jerome-benoit (changes requested):
Reviewed `944cd25f` with two independent sequential analyses and final arbitration. One blocking finding.

The barrier itself is a correct, well-tested fix for #4016: the runner queues a non-sheddable `_aoe/session_replayed` marker right behind `SessionReady`, and `session/load` awaits `session_replayed()` (a monotonic `sessions_replayed >= sessions_established` watch barrier) before the command loop can take a prompt, so suppression is lifted only after the replay drains. The unit test (`control.rs`, `session_replayed_waits_for_replay_sent_after_the_reply`) and the two integration tests prove the barrier stays closed until the replay update is applied. Counter semantics (monotonic, per-connection, `>=` snapshot), the non-sheddable classification, forged-marker rejection, and the non-v3 path are all correct.

**Blocking: the mandatory marker is added to control protocol v3 but `CONTROL_PROTOCOL_VERSION` stays `3`** (`src/acp/control_protocol.rs:27`), whose doc comment says to bump it "whenever a wire-incompatible body or semantic contract changes" — and this is a semantic-contract change. The shipped v1.16.0 speaks v3 *without* this marker (`git show v1.16.0:src/acp/control_protocol.rs` = 3, zero `session_replayed`). Detached v3 runners are designed to survive daemon restarts/upgrades, so a daemon built from this PR that reattaches to a still-alive v1.16.0 runner passes the `Hello` version gate (3 == 3, `src/acp/acp_client/control.rs:302`), reaches `session/load`, increments `sessions_established`, then blocks in `session_replayed().await` (`src/acp/acp_client/connection.rs:1297`) indefinitely: the old runner never emits the marker, `sessions_replayed` stays `0`, and there is no timeout. Because that runner is alive (not EOF'd), connection teardown does not bound the wait either. The version gate is the project's designed safeguard for exactly this case, and leaving the constant at 3 bypasses it, turning a mixed-version reattach into a permanent `session/load` hang.

Fix: bump `CONTROL_PROTOCOL_VERSION` to 4 so a mixed-version reattach is rejected cleanly at `Hello` instead of hanging. As defense-in-depth, consider a bounded wait/timeout on `session_replayed()`.

Evidence: exact-SHA source tracing at the pinned head plus protocol-version archaeology across v1.15.0 / v1.16.0 / main (v2 -> v3 -> 3 unchanged; marker absent at v1.16.0); two independent sequential analyses with final arbitration (one approved — I hold change-requested because that approval assumed every v3 runner emits the marker, which the shipped v1.16.0 runner does not). No local build or test run; source-confirmed, not executed. CI is green on this head per `gh pr checks`; this concerns the code review only.

njbrake:
@jerome-benoit No bump needed; the hang path is unreachable. The wait only follows `session/load` in `ConnectMode::Fresh`, which is built only in `AcpClient::spawn` (`mod.rs:347`). That spawn terminates any prior runner and starts a new one from the daemon's own `current_exe()`, so it never talks to an older runner. Reattach uses `ConnectMode::Resume` and never loads. v1.16.0 runners record `RUNNER_VERSION = 3` against main's 4, so the reconciler replaces them rather than attaching. Per `worker_registry.rs:35-38`, `RUNNER_VERSION` gates generations and the wire version stays 3. Bumping only `CONTROL_PROTOCOL_VERSION` would fail `Hello` for live generation-4 runners that the reconciler still treats as compatible.

jerome-benoit (approved):
Re-reviewed `944cd25f` after @njbrake's response; withdrawing my change request and approving. My blocking concern does not hold, and the rebuttal is correct on the evidence.

I claimed a mixed-version reattach could hang on `session_replayed()`. It cannot:
- The wait runs only on the `session/load` branch (`src/acp/acp_client/connection.rs:1297`), and that branch is reached only on a Fresh spawn: `AcpClient::spawn` builds `ConnectMode::Fresh` and `terminate_and_wait`s any prior runner (`src/acp/acp_client/mod.rs:384`), then starts a new one from the daemon's own `current_exe`. So the runner answering `session/load` is always the same build as the daemon and always emits `_aoe/session_replayed`.
- Reattach uses `ConnectMode::Resume`, which reuses the stored ACP session id without sending session/new or session/load (`mod.rs:765-766`), so it never reaches the wait.
- Generation skew is handled by `RUNNER_VERSION` (`src/process/worker_registry.rs:44` = 4), not the wire version: a v1.16.0 runner records `RUNNER_VERSION` 3 and is reaped/replaced by the reconciler rather than attached. Bumping `CONTROL_PROTOCOL_VERSION` would instead break `Hello` for live generation-4 runners the reconciler treats as compatible.

So no version bump is needed; my earlier reachability inference was wrong. The barrier fix itself is correct and well-tested (non-sheddable marker, monotonic per-connection counters, forged-marker rejection, non-v3 path gated). Approving.

Evidence: source-traced the wait call site and its `ConnectMode` gating (`connection.rs`, `mod.rs`) plus `RUNNER_VERSION`/reconciler (`worker_registry.rs`); source-confirmed, not executed.

Issue #2614, opened 2026-07-02
Able to restore lastest session settings


hwanhoon-getvocalai:
There's a text that I always use for agent instruction. Whenever I create a new session I have to always write again the instruction so it's a bit bothering. But for sure it's not blocking anything just a minor suggestion of UX.

<img width="535" height="740" alt="Image" src="https://github.com/user-attachments/assets/d3d57bb0-930e-4183-bb30-58fbcccea365" />

PR #4022, opened 2026-09-17
feat(web): open a plain New session on the last launched project

### LINKED ISSUE #4021: Web wizard: a plain New session should open on the last launched project
## What happens

Every plain New session in the web dashboard starts with no project selected, so the project has to be picked by hand on every open, even for someone with a single registered project. The wizard already remembers the last tool (`aoe-acp-last-tool`) and the last agent instruction (#2614) per browser; the project is the one field it forgets.

## Expected

A plain New session opens on the project the user launched last, the same per-browser way the tool and instruction are remembered. Prefilled opens (the sidebar's per-project **+**, scratch, clone) keep bringing their own path or none.

## Notes from building it

Seeding the path at mount satisfies the launch gate before the chained profiles/settings fetch has applied the profile's defaults, so a one-action launch (click or Cmd/Ctrl+Enter) in that window would submit `initialData`'s sandbox/worktree/yolo values instead of the profile's; the server trusts those booleans (`create.rs` forwards them into the spec). That weakness pre-exists for prefilled opens (the sidebar **+** with a profile that enables the sandbox) but a remembered path makes it the ordinary path, so the PR gates Launch on the defaults having settled, on every outcome of the chain including a failed fetch.

Two further states the seed exposed, also handled in the PR: the profile-defaults seeder re-enabling a worktree on a path the repo probe already rejected, and a remembered path with no saved or recent rows landing the picker on Browse with Launch enabled and no target on screen. The CityHall name-only wizard never reads the memory.

PR follows.

## Environment

aoe 1.16.0 (Homebrew, macOS), Chrome desktop.

bcstrawn:
## Description

A plain New session now opens on the project you launched last.

**Problem.** The wizard remembers, per browser, the last tool you picked (`aoe-acp-last-tool`) and the last agent instruction you sent (#2614), but the project was still picked by hand on every open, even for someone with a single registered project.

**Change.** `SessionWizard` remembers the path of the last launched session under `aoe-new-session-last-project` and seeds a plain New session with it. A prefill (the sidebar's per-project **+**, scratch, clone) keeps bringing its own path or none, so nothing changes for those flows; a scratch launch leaves the memory alone (the saved value is the request body's `path`, empty for scratch); a stored value that is not an absolute path is ignored. The Recent tab shows the remembered project as the selected row, and picking another project replaces it as usual.

**Launch waits for the profile defaults.** Seeding the path at mount satisfied the launch gate before the chained profiles/settings fetch had applied the profile's defaults, so a one-action launch (click or Cmd/Ctrl+Enter) in that window would have submitted `initialData`'s sandbox/worktree/yolo values instead of the profile's; the server trusts those booleans. The wizard now tracks that the chain has settled, on every outcome including a failed fetch, and `LaunchFooter` gates the button and the shortcut on it. The tests that clicked Launch synchronously after render were asserting that unsafe timing and now wait for the enabled button, as a real click and the browser stories do. Two further states the seed exposed are handled too: the defaults seeder no longer re-enables a worktree on a path the repo probe rejected, and a remembered path with no saved or recent rows keeps the Recent tab so its selection is visible. The CityHall name-only wizard never reads the memory.

Closes #4021.

## PR Type

- [x] New Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

- [x] New and existing tests pass (`npm run test:unit`: 366 files / 3864 tests; full mocked Playwright run, the only failures being stories that also fail on a clean `main` on this machine; `npx tsc -b`, `npm run lint`, `npm run format:check` clean)
- [x] Documentation was updated where necessary (`docs/guides/web/dashboard.md`, one sentence)
- [ ] For UI changes: included screenshot or recording (the wizard opens with the Recent row already selected and Launch enabled; happy to add a recording)

## Test Coverage Analysis

- [ ] N/A
- [x] Skipped a new browser story, because: the behaviour is the wizard's initial state and its post-launch write, exercised end to end in Vitest with the harness the last-instruction memory uses (`SessionWizard.lastProject.test.tsx`: remembered path reaches the create payload; a launch writes the path; a prefill path wins; a scratch launch neither seeds nor overwrites; a non-absolute stored value is ignored; name-only mode never seeds; the selection stays visible with no picker rows; Launch holds until the profile defaults land and then submits them; the form recovers from a failed settings fetch). A reducer case covers the worktree-on-non-repo guard. `wizard.remember-last-project` added to `tests/coverage-matrix.json` as a `vitest` entry; the existing `wizard-scratch-ui.spec.ts` shortcut story now asserts Launch is enabled before Cmd+Enter.

**TUI / CLI (e2e test)**

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:** Claude (Fable 5.1) in Claude Code, driven by @bcstrawn, who chose this design over a keyboard shortcut. Two independent agent reviews preceded the PR (one Claude, one Codex); the launch-before-defaults gate and the three seeded-path states above are their findings, each verified with a test.

**Any Additional AI Details you'd like to share:** The agent wrote the change and tests from the source, mirroring the existing last-instruction memory; the operator answers reviewer questions.

## Summary

- Plain New sessions now remember and preselect the last launched absolute project path per browser.
- Prefilled, scratch, clone, and CityHall sessions keep their existing path behavior.
- Scratch and name-only sessions do not update project memory.
- The Recent tab stays visible for remembered paths, even without saved or recent project rows.
- Launch waits for profile defaults and settings to settle, including after fetch failures.
- Rejected repositories cannot regain worktree mode.
- Tests and dashboard documentation cover the new behavior.

## Benefits

This reduces repeated project selection while preserving existing session flows. It also prevents launches from using stale profile settings.

## Further enhancement

No test execution results or current review findings were provided.


jerome-benoit (dismissed):
Reviewed `c2c6137a` with two independent sequential analyses and final arbitration. No blocking finding.
The feature is correct: a plain New session seeds its project path from `aoe-new-session-last-project` only when there is no prefill; a prefill (sidebar +, scratch, clone) always overrides via `prefillData.path = prefill.scratch ? "" : prefill.path || ""`, `nameOnly` (CityHall) forces empty, and the memory is written only on a successful create (`saveLastUsedProject(body.path)` in the ok branch, guarded to drop the empty scratch path). The PR also proactively closes two races its own seeding exposes: `defaultsReady` gates Launch so a one-action launch on a seeded path cannot submit initial sandbox/worktree/YOLO before `/api/settings` applies, and the reducer forces `useWorktree` off when the mount-time repo probe rejects the path — both with tests.
CodeRabbit's Major (`fetchProfiles()` reject -> `defaultsReady` set without settings) is a valid observation but PRE-EXISTING, not introduced: on main the same profiles->settings coupling already skips `fetchSettings` on a profiles rejection and, being ungated, already launched with `initialData`. Those defaults are the safe-side values (`yoloMode:false`, `sandboxEnabled:false`, `useWorktree:false`), so there is no data-safety regression; the PR only widens exposure (a seeded path makes the pre-existing gap one-action-reachable in the narrow case where /api/profiles fails but /api/settings would succeed while online). Reclassified to non-blocking.
Non-blocking (P3): (1) worth adopting CodeRabbit's cheap quick-win `.catch(() => [])` before the `.then()` so a profiles-only failure still loads explicit-profile/global settings — it addresses the gap this PR's seeding widens. (2) `loadLastUsedProject` validates only an absolute-path prefix, not existence, so a since-deleted last project seeds a picker whose Launch is enabled and fails server-side (a recoverable `SUBMIT_ERROR` banner), unlike the Recent list which drops vanished dirs; this is consistent with the existing last-tool/last-instruction memories. (3) the `fetchProfiles`-reject branch itself is untested (the added test covers only the `fetchSettings` rejection).
Evidence: exact-SHA source tracing at the pinned head plus a base comparison against main for introduced-vs-preexisting; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. I authorized the gated CI/Tests/CodeQL workflows on this head and they are now running. This approval concerns the code review only.

jerome-benoit (approved):
Re-reviewed `c2c6137a..ba790cad` (commit ba790cad); re-approving.
The delta resolves both P3s from my prior review: it adds `.catch(() => [] as Awaited<ReturnType<typeof fetchProfiles>>)` between `fetchProfiles()` and the first `.then` (`SessionWizard.tsx`), so a failed profiles fetch now recovers to `[]` and the chain still calls `fetchSettings(effectiveProfile || undefined)` — loading the explicit-prefill/global settings and their sandbox default instead of launching on `initialData`. This is strictly better than before (the profiles error was already swallowed by the terminal `.catch(() => {})`, so there is no new silent hazard; only the fallback is upgraded). The trailing `.catch(() => {}).finally(setDefaultsReady(true))` is unchanged and still covered by the existing settings-reject test.
Verified/rejected: the new test `still fetches and applies settings when the profiles fetch fails` is a genuine, non-tautological oracle — pre-fix (the flattened chain) a profiles rejection skipped `.then`, so `fetchSettings` was never called and the `toHaveBeenCalledWith(undefined)` wait would time out; post-fix it passes and `createSession` carries `{path:'/tmp/remembered', sandbox:true}`, proving the profile's sandbox default lands even when the profiles fetch fails. The `[] as Awaited<ReturnType<typeof fetchProfiles>>` cast is sound (resolves to `ProfileInfo[]`). No last-project regression; `coverage-matrix.json` is correctly untouched (the test file is modified, not new, and `SessionWizard.tsx` is already listed).
Evidence: exact-SHA source tracing of the delta at the pinned head; two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. I authorized the gated CI/Tests/CodeQL workflows on this head and they are now running. This approval concerns the code review only.

PR #4023, opened 2026-09-17
fix(session): compare hook-path consent by target, not by path string

hairbui76:
## Description

The hook-path consent gate from #3678 compares the disclosed and resolved paths as strings. A `before_session` hook that points the agent's config dir at a per-account directory whose settings file is a **symlink back to the shared one** therefore trips the gate, even though both paths name the same bytes on disk:

```
before_session changed the agent hook path from
  /home/me/.claude/settings.json to
  /home/me/.claude-accounts/profiles/work/settings.json;
declare the override in the profile environment before consenting
```

This is the shape every account switcher I know of uses: only the credential files are per-account, and everything else in `~/.claude` (or `~/.codex`) is linked back so sessions, settings and history stay shared. `settings.json` and `config.toml` are on the linked side, so the two paths differ as strings and resolve to one file.

Compare by resolved target instead. `same_hook_target` canonicalizes both sides; a file that does not exist yet resolves through its parent directory, reusing the idiom already in this file for `first_read_only_report`.

## What this does not weaken

The gate exists so a hook cannot silently redirect the hook install into a file the user never approved. That property is unchanged: a redirect into a genuinely different directory still canonicalizes differently and is still refused. Only "two names for one file" now passes, which is what the consent already covered — consent is about which file gets written, not which string names it.

The fast path is still a plain `==`, so the common case where nothing overrides the config dir does no filesystem work.

## Why not just use `session.agent_config_dir`

It is the right answer for a static setup and I would not change that. Declaring it makes both sides of the comparison resolve through the same branch, so the gate passes — and its doc comment says it exists for "wrappers that export the config-dir env var themselves", which is exactly this class of tool.

What it cannot express is a **per-directory** account, because it is global/profile-only and keyed by agent, and it deliberately wins over the env var including one a hook minted. So `agent_config_dir` covers "this profile is always account X"; it does not cover "this repo is client work, that one is personal", which is the case `before_session` was added for in #3257 and which [`docs/guides/configuration.md`](docs/guides/configuration.md) still documents:

```toml
[host_hooks]
before_session = ['my-account-switcher env --profile "$AOE_PROFILE"']
```

Both mechanisms stay useful after this change; today the dynamic one is unusable for its documented purpose on any switcher that links its settings file.

## PR Type

- [ ] New Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation
- [ ] Infrastructure / CI

## Checklist

No docs change: this restores the behaviour the Host Hooks section already describes. No UI surface.

## Test Coverage Analysis

**TUI / CLI (e2e test)**
- [x] Skipped a test, because: the whole behaviour is a pure path comparison, and a unit test can cover the refusal cases directly; an e2e would mostly be re-testing the harness's ability to build a symlink farm.

`same_hook_target_follows_symlinks_but_not_distinct_files` covers both directions, and the refusal half is the point:

| case | expected |
| --- | --- |
| identical paths | passes, no filesystem touch |
| per-account `settings.json` symlinked to the shared file | passes |
| symlinked config *directory* | passes |
| file not written yet, same real directory | passes |
| two distinct real files | **refused** |
| redirect into an unrelated directory | **refused** |

The existing `ensure_disclosed_host_hook_path` tests still pass unchanged.

## AI Usage

- [ ] No AI was used
- [x] AI was used

**AI Model/Tool used:**

Claude Opus 5 via Claude Code, used for the implementation and tests.

Worth a reviewer's attention: `hook_target_identity` compares canonical *paths*, not `(device, inode)`. That is deliberate — it keeps the missing-file case working, since a file that does not exist has no inode — but it means two hardlinks to one file still compare as different. On Linux the switchers use symlinks, so this does not bite in practice; if you would rather the gate be exact for hardlinks too, say so and I will compare identity for the existing-file case and fall back to the path form only when the file is absent.

Verified on Linux. Not verified on macOS.

## Summary

- Updated the session hook consent check to compare canonicalized target paths.
- Preserved the fast path for identical path strings.
- Added support for missing files by resolving paths through their parent directories.
- Accepted symlinked files and directories that resolve to the same target.
- Continued to refuse redirects to different files or unrelated directories.
- Added unit tests for identical, missing, symlinked, and distinct paths.

## Benefits

This restores support for per-directory account switchers that share configuration files through symlinks without weakening consent checks. Linux verification passed; macOS verification was not performed.


jerome-benoit (approved):
Reviewed `f4fb185b` with two independent sequential analyses (one security-focused) and final arbitration. No blocking finding.
The fix correctly narrows the #3678 hook-path consent gate from a brittle string comparison to a resolved-target comparison, closing the account-switcher false positive without opening a bypass. `same_hook_target` returns true only when both paths canonicalize to the same real file (or, for a not-yet-written file, the same canonicalized parent + file name), so an identity match provably means the install writes the very bytes the user consented to; a genuinely different destination canonicalizes differently and is still refused (the test's unrelated-dir case). Critically, the gate-checked `resolved` path and the actual write path both derive from the same `generic_host_config_path_for(...)` over the resolved environment (gate ~:288-292 vs `install_json_host_hooks` ~:592 / `install_codex_host_hooks` ~:560), so there is no check-vs-write divergence to exploit.
Verified/rejected: edge cases (broken symlink, absent file/parent, symlink loop, non-existent parent) all fall through `canonicalize` errors to conservative parent/literal fallbacks that fail toward refusal; the `disclosed == resolved` fast path means identical paths never touch the filesystem. `canonicalize` is cross-platform (the production path is not unix-gated; only the test is `#[cfg(unix)]`). The unit test is thorough and non-tautological, asserting both the false-positive fix (symlinked file, symlinked dir, missing file via symlinked parent) and the anti-regression refusal (distinct files existing/absent, unrelated-dir redirect).
Non-blocking (P3): the two new functions were inserted between `first_read_only_report`'s doc comment (~:642-644, "The resolved target to report, or `None` ...") and that function with no blank line, so Rust now attaches those three `///` lines to `hook_target_identity` (which always returns a `PathBuf`, not an `Option`), leaving `first_read_only_report` undocumented. Move the three lines back down to immediately above `fn first_read_only_report`.
Evidence: exact-SHA source tracing at the pinned head (gate call site, `same_hook_target`/`hook_target_identity`, and the write-path derivation); two independent sequential analyses with final arbitration. No local build or test run; source-confirmed, not executed. CI is green on this head per `gh pr checks`. This approval concerns the code review only.

njbrake (approved):
Approve.

The gate now compares which file gets written rather than which string names it, which is what the consent already covered. Checked the new comparison in isolation against a real filesystem: dangling symlink, symlink loop, `..` traversal onto a different file, missing parent, and hardlink all still refuse, and every `canonicalize` failure path falls through to a narrower comparison. Gate-time and write-time resolution agree, since `install_hooks` writes through `atomic_write`, which resolves the same symlink chain.

P3, not blocking: the two new functions landed between `first_read_only_report`'s doc comment and that function, so those three lines now document `hook_target_identity` and describe an `Option` it never returns. `hook_target_identity` also repeats `first_read_only_report`'s body verbatim; the latter can just call it. Follow-up material.

Rebased onto main for #4024, which fixes the ACP deadline failure this branch predated.

_Note: this review was submitted by Claude Opus 5 via back-and-forth with @njbrake. The reasoning and decisions are his; the prose is Claude's._

